PR #12, opened 2016-06-28
feat(isomorphic): Refactored the backend-only exports to ./backend.ts…

zakhenry:
… to allow isomorphic apps to load typeorm without the sql drivers attempting to load.

### DIFF
--- a/package.json
+++ b/package.json
@@ -42,7 +42,7 @@
     "ts-node": "^0.8.0",
     "tslint": "next",
     "tslint-stylish": "^2.1.0-beta",
-    "typescript": "next",
+    "typescript": "1.9.0-dev.20160624-1.0",
     "typings": "^1.0.4"
   },
   "dependencies": {
@@ -57,8 +57,8 @@
     "sha1": "^1.1.1"
   },
   "scripts": {
-    "postversion": "./node_modules/.bin/gulp package",
-    "postinstall": "./node_modules/.bin/typings install",
-    "test": "node_modules/.bin/gulp tests"
+    "postversion": "gulp package",
+    "postinstall": "typings install",
+    "test": "gulp tests"
   }
 }
--- a/sample/sample1-simple-entity/app.ts
+++ b/sample/sample1-simple-entity/app.ts
@@ -1,5 +1,5 @@
 import "reflect-metadata";
-import {createConnection, CreateConnectionOptions} from "../../src/index";
+import {createConnection, CreateConnectionOptions} from "../../src/backend";
 import {Post} from "./entity/Post";
 
 const options: CreateConnectionOptions = {
--- a/sample/sample10-mixed/app.ts
+++ b/sample/sample10-mixed/app.ts
@@ -1,5 +1,5 @@
 import "reflect-metadata";
-import {createConnection, CreateConnectionOptions} from "../../src/index";
+import {createConnection, CreateConnectionOptions} from "../../src/backend";
 import {Post} from "./entity/Post";
 import {PostDetails} from "./entity/PostDetails";
 import {Image} from "./entity/Image";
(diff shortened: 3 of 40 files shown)


zakhenry:
@pleerock This PR separates out the backend-only initialisation parts into a separate `backend.ts` file.

This is require to be able to load the typeorm library in the frontend without the Mysql driver complaining. 

I'm not sure if this is something that makes sense to have in the core typeorm repository, also unsure if this is the best strategy. Feel free to either close the PR if it's not something you see being useful in the core, or let me know any changes. I'm happy to maintain my own fork at @ubiquits/typeorm` as I realise I have a niche requirement of full stack orm.

pleerock:
1. Why do you need to "load the typeorm library in the frontend"? Do you mean load typeorm in the browser?
2. Can you tell me why mysql driver is complaining? Because from code I don't see why it should complian in the frontend
3. I'll try to provide as much as possible extensibility features, so you can extend orm for your needs, instead of overriding it and support that hell in your own branch. Also it will be good if we can organize pluggable architecture, so you can unplug/plug your overrides easily with lets say `@ubiquits/typeorm-extensions` in the case if you need lot of customized functions. But overall I don't recommend to go much out of original typeorm nature, because I think orm and framework should exist in separate (but work together like best friends). Also in the future there will be two separate documentations and user may confuse if implementations will be too different.

zakhenry:
1. Yes, typeorm in the browser. I'm starting to think I may be better off keeping typeorm in the backend and adding a layer of interface so the common layer is controlled by the framework. It's hard to work out where to 'draw the line' as there are a number of features of typeorm that make sense on the frontend (relationships, column types, query building etc), but there are parts that don't make sense like the drivers, connection, migrations etc.
2. Specifically there was two issues with the driver - webpack can't cope with the `if (!mysql && require) {` syntax - and the postgres driver was attempting to load the `pg` peer dep which didn't exist. The issue is that webpack statically & eagerly loads modules regardless of whether they are conditionally loaded. I don't think there is a way around this.
3. Yes agreed on all points - I'll revisit my implementation and see if I can keep typeorm completely isolated to the backend with an isomorphic inserted in between. This does mean that there will need to be separate documentation on my part however.

pleerock:
Can you explain your idea? I can't imagine how frontend can use relationships, column types or query building? Query building on the frontend looks sounds strange. How are you going to send the resulted query to the backend to perform a query? Or maybe I missunderstood your definition of the frontend? Because for me frontend is a client side of the app which communicates with backend within http requests

zakhenry:
Your definition is the same as mine - the frontend in my case is an Angular 2 application, communicating to the backend with `@angular/http`.

The frontend can use relationships to handle retrieval of related entities, or bulk creation of entities etc. This of course needs a custom repository definition to interface with the rest api.

The same goes for query building - query building generates an object (not SQL yet), that is sent as json as a query param over REST so the API can effectively blindly take that object and run a query. In practice of course there would be access checks otherwise it would effectively be an sql injection vector.

pleerock:
> The frontend can use relationships to handle retrieval of related entities
are you talking about lazy loading of relations? If yes, then its most likely will be unusable, since frontend may send lot of queries for each relation, and another queries for their subrelations, etc. And lazy loading relations with http is a performance bottleneck 
> or bulk creation of entities etc. 
creation is not related to relations
> - query building generates an object (not SQL yet), that is sent as json as a query param over REST 
this will be uncontrolled thing. query builder will generate you a complex object, which will be hard to control for user on the backend side. Because of its complex nature it will be both hard to control and maintain, and it most likely will be schemaless and non type-safe. But we are creating a backend to have a control of things, right?

zakhenry:
I have managed to refactor my architecture to keep typeorm isolated to the backend, interfacing with typeorm via my own decorators and repository interfaces. This does have the unfortunate side effect that documentation will diverge significantly from typeorm, but it does give me the required flexibility to develop full stack patterns that are unbound from database implementation.

Closed by https://github.com/ubiquits/ubiquits/pull/90

PR #52, opened 2016-10-03
quick fix for multi primary columns (and fix another bug)

yannisgu:
by moving the call to `firstPrimaryColumn` under the condition `isManyToMany` wecan enable working inserts with multi primary columns for simple relations.

### DIFF
--- a/config/parameters.json
+++ b/config/parameters.json
@@ -4,14 +4,14 @@
       "host": "localhost",
       "port": 3306,
       "username": "root",
-      "password": "admin",
+      "password": "root",
       "database": "test"
     },
     "mysqlSecondary": {
       "host": "localhost",
       "port": 3306,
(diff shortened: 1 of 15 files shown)


config/parameters.json
```diff
@@ -4,14 +4,14 @@
       "host": "localhost",
       "port": 3306,
       "username": "root",
-      "password": "admin",
+      "password": "root",
```
pleerock: oh, i know test environment is not perfect for contributing, but next time please discard these changes

pleerock:
Thank you for your PR, can you please discard all unrelated changes and re-submit the PR? I the near future I'll try to setup better testing environment better, so anyone can use their changes without committing it

pleerock:
okay I'll merge it and revert some changes by myself

PR #59, opened 2016-10-18
resolve entity and schema dir paths

strictd:
I know very little about the inter-working of this system. This fixed the issue of loading up entities from ormconfig.json using relative path of the root directory.


pleerock:
thanks ;)

PR #64, opened 2016-10-27
fix dealing with blank mysql date fields

strictd:
currently typeorm returns back a string 'Invalid Date' when a date or datetime or timestamp is set to 0000-00-00 00:00:00 in the database.. Also when persisting a date of 0000-00-00, automatically rewrites it to start of epoch, 1970-01-01 00:00:00 .. It's preferred to have the blank mysql dates as zeros and not an arbitrary date the makes no sense for anything in todays world.

### DIFF
--- a/src/driver/mysql/MysqlDriver.ts
+++ b/src/driver/mysql/MysqlDriver.ts
@@ -217,11 +217,17 @@ export class MysqlDriver implements Driver {
             case ColumnTypes.BOOLEAN:
                 return value === true ? 1 : 0;
             case ColumnTypes.DATE:
-                return moment(value).format("YYYY-MM-DD");
+                if (moment(value).isValid())
+                  return moment(value).format("YYYY-MM-DD");
+                else return '0000-00-00';
             case ColumnTypes.TIME:
-                return moment(value).format("HH:mm:ss");
+                if (moment(value).isValid())
+                    return moment(value).format("HH:mm:ss");
+                else return '00:00:00';
             case ColumnTypes.DATETIME:
-                return moment(value).format("YYYY-MM-DD HH:mm:ss");
+                if (moment(value).isValid())
+                  return moment(value).format("YYYY-MM-DD HH:mm:ss");
+                else return '0000-00-00 00:00:00';
             case ColumnTypes.JSON:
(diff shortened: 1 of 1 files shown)


src/driver/mysql/MysqlDriver.ts
```diff
@@ -217,11 +217,17 @@ export class MysqlDriver implements Driver {
             case ColumnTypes.BOOLEAN:
                 return value === true ? 1 : 0;
             case ColumnTypes.DATE:
-                return moment(value).format("YYYY-MM-DD");
+                if (moment(value).isValid())
+                  return moment(value).format("YYYY-MM-DD");
+                else return '0000-00-00';
```
pleerock: @strictd are you sure `'0000-00-00'` should be persisted here? Maybe `null` should be instead? Because if you want your column to be nullable with this logic it can never be nullable

strictd: No.. you bring up a good point.. would this be better?

```javascript
 if (value === null)
  return null;
else if (moment(value).isValid())
  return moment(value).format("YYYY-MM-DD");
else return '0000-00-00';
```

pleerock: Also I think '0000-00-00' should be never persisted, because if it will be, then after such value will be hydrated new Date('0000-00-00') will create an "invalidate date".

strictd: it seems to be a developer preference on using 0000-00-00 vs null. something to consider, that I just tested, is moment(null) returns NaN .. while Date(null) returns the current timestamp. Wouldn't both Null and 0000-00-00 both hydrate as invalid date anyways?

pleerock: the problem in new Date('0000-00-00').  See [here](https://github.com/typeorm/typeorm/pull/73/commits/146d1499d1cb55dc8c3323af2071cd6c5529daa0) my latest changes in this area

strictd: I like your new logic using native Date much better, including the UTC column options. Handles all situations I can currently think of.  Great update!

PR #65, opened 2016-10-27
fixes mysql datetime issues when persisting records

strictd:
current behavior attempts to insert 'Invalid Date' string into database when moment cannot parse date string. typically due to the default mysql datatime value of 0000-00-00 00:00:00 creates the invalid date issue. This fix allows invalid dates to go into the database as 0000-00-00 00:00:00 and not 01-01-1970 00:00:00 ..

### DIFF
--- a/src/driver/mysql/MysqlDriver.ts
+++ b/src/driver/mysql/MysqlDriver.ts
@@ -217,11 +217,17 @@ export class MysqlDriver implements Driver {
             case ColumnTypes.BOOLEAN:
                 return value === true ? 1 : 0;
             case ColumnTypes.DATE:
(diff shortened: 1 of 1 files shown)


pleerock:
please provide tests too because each uncovered line of code can be removed during refactorings without knowledge of why it is for

strictd:
to be honest, I have no experience or knowledge in writing test libraries. I've managed to update existing tests, however writing a new test isn't something i'm very good at. My mind works mostly in problem solving through root cause and analysis. I look forward to wrapping my head around jasmine, karma, protractor, chai and mocha. At this point I have no idea where to begin with adding a test for this update.

pleerock:
okay

PR #74, opened 2016-11-04
SQLite driver insert fix

f-wrobel:
Error: when attempting to insert an empty row to database using a sqlite driver, it causes an error. Fix: use "insert into default values".


pleerock:
thanks

pleerock:
would be great if you'll add some tests to your issue, so next time if someone change these lines of code and accidental remove them, your tests will warn a changer. Also don't forget to leave comment in the code why you have made this change and link to the github issue if its exist. Also if you want further contributions you can add same driver-specific code for other databases too. Thank you again

Issue #72, opened 2016-11-04
Feature Request: Unique keys combination


AmirTugi:
Hi,
I have a scenario where I want to use 2 fields as a unique key (so the combination of them both would be unique). Just like [here](http://dba.stackexchange.com/questions/17761/how-to-use-unique-key-via-combinations-of-table-fields).
Currently, I see no implementation of this in TypeORM, and I think it'd be a nice-to-have feature.

pleerock:
Hi,
you can use [indices](https://typeorm.github.io/indices.html):


```typescript
import {Table, Index, PrimaryGeneratedColumn, Column} from "typeorm";

@Table()
@Index("my_index_with_id_and_text", ["id", "text"], { unique: true }) // first style
@Index("my_index_with_id_and_title", (post: Post) => [post.id, post.title], { unique: true }) // alternative, type-safe style
export class Post {

    @PrimaryGeneratedColumn()
    id: number;

    @Column()
    title: string;

    @Column()
    text: string;

    @Column()
    likesCount: number;

}
```

AmirTugi:
Mmm, nope, not working

```
import {Table, Column, PrimaryGeneratedColumn, ManyToOne, CreateDateColumn, UpdateDateColumn, Index} from "typeorm";
import {Item} from "./item";

@Table()
@Index("index_item_sequence", ['item', 'sequence'], { unique: true })
export class Attachment {
    @PrimaryGeneratedColumn()
    id: number;
    @Column()
    path: string = '';
    @Column()
    isPublic: boolean = true;
    @Column()
    type: string = 'attachment';
    @Column()
    size: number = 0;
    @Column()
    link: string = '';
    @Column()
    sequence: number = 1;
    // Automatically insert, update, and remove attachments attached to the item.s
    @ManyToOne(type => Item, item => item.attachments, { cascadeAll: true })
    item: Item;
    @Column()
    comments: string = '';
    @CreateDateColumn()
    createdAt: string;
    @UpdateDateColumn()
    updatedAt: string;

}
```

I can still create an attachment with the same sequence for a specific item.

pleerock:
hmm it should work, can you enable sql logging and check what sql it generates? it should create an index with unique properties item+sequence

AmirTugi:
Ok, I did a small research, and after figuring that the the index was created, but not uniquely, I decided to dive into the code, and look for the problem.

I found that in `/decorator/index.js` there's a bug in line 9:
[...]

PR #76, opened 2016-11-05
Unique Issue fixed

AmirTugi:
Fixed an issue when maybeFieldsOrOptions is an array, options would still go for maybeFieldsOrOptions instead of maybeOptions.
Reference to issue #72


pleerock:
thanks

Issue #86, opened 2016-11-15
Map to view and schema autoSchemaSync


apezel:
If i want to map an entity to a view in my db, is there a way to prevent autoSchemaSync to try to create the table ?

pleerock:
don't think so... But this can be easily implemented as extra option to table decorator, like skipSchemaSync: true

pleerock:
closing this because its already implemented thanks to @apezel

PR #92, opened 2016-11-16
added a new table option skipSchemaSync to prevent schema-builder to …

apezel:
Added a new skipSchemaSync option to @Table.
Useful to prevent schema builder to sync a table, for example if it's a view.

Example of usage : 

```javascript
@Table("view", {skipSchemaSync: true})
export class View {

    @PrimaryColumn()
    id: number;

    @Column()
    title: string;

}
```

*Added a test case in functional/connection/connection.ts with a new entity named View* #86

### DIFF
--- a/src/decorator/tables/ClassTableChild.ts
+++ b/src/decorator/tables/ClassTableChild.ts
@@ -11,7 +11,8 @@ export function ClassTableChild(tableName?: string, options?: TableOptions) {
             target: target,
             name: tableName,
             type: "class-table-child",
-            orderBy: options && options.orderBy ? options.orderBy : undefined
+            orderBy: options && options.orderBy ? options.orderBy : undefined,
+            skipSchemaSync: (options && options.skipSchemaSync === true) || false
         };
         getMetadataArgsStorage().tables.add(args);
     };
--- a/src/decorator/tables/ClassTableChild.ts
+++ b/src/decorator/tables/ClassTableChild.ts
@@ -12,7 +12,7 @@ export function ClassTableChild(tableName?: string, options?: TableOptions) {
             name: tableName,
             type: "class-table-child",
             orderBy: options && options.orderBy ? options.orderBy : undefined,
-            skipSchemaSync: (options && options.skipSchemaSync === true) || false
+            skipSchemaSync: !!(options && options.skipSchemaSync === true)
         };
         getMetadataArgsStorage().tables.add(args);
     };
--- a/src/schema-builder/SchemaBuilder.ts
+++ b/src/schema-builder/SchemaBuilder.ts
@@ -96,7 +96,7 @@ export class SchemaBuilder {
     // -------------------------------------------------------------------------
 
     protected get entityToSyncMetadatas(): EntityMetadataCollection {
-        return <EntityMetadataCollection>this.entityMetadatas.filter( metadata => !metadata.table.skipSchemaSync );
+        return this.entityMetadatas.filter(metadata => !metadata.table.skipSchemaSync);
     }
 
     /**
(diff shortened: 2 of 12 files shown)


src/decorator/tables/ClassTableChild.ts
```diff
@@ -11,7 +11,8 @@ export function ClassTableChild(tableName?: string, options?: TableOptions) {
             target: target,
             name: tableName,
             type: "class-table-child",
-            orderBy: options && options.orderBy ? options.orderBy : undefined
+            orderBy: options && options.orderBy ? options.orderBy : undefined,
+            skipSchemaSync: (options && options.skipSchemaSync === true) || false
```
pleerock: `options && options.skipSchemaSync === true` already returns boolean, so `|| false` is redundant here and only brings more confusion. But its not important.

src/schema-builder/SchemaBuilder.ts
```diff
@@ -95,19 +95,23 @@ export class SchemaBuilder {
     // Private Methods
     // -------------------------------------------------------------------------
 
+    protected get entityToSyncMetadatas(): EntityMetadataCollection {
+        return <EntityMetadataCollection>this.entityMetadatas.filter( metadata => !metadata.table.skipSchemaSync );
```
pleerock: can you reformat it this way: `<EntityMetadataCollection> this.entityMetadatas.filter(metadata => !metadata.table.skipSchemaSync)`

pleerock: also your code is not correct. `.filter` does not return instance `EntityMetadataCollection`, it returns you `Array`.  So you are lying to compiler by casting it to `EntityMetadataCollection` will bring you errors once there will be a call to a specific method.

src/decorator/tables/ClassTableChild.ts
apezel: Ok, i can change it but ```options && options.skipSchemaSync === true```returns a boolean only if options isn't undefined. If options is undefined it returns undefined, not false

src/schema-builder/SchemaBuilder.ts
apezel: Btw, the linter doesn't complain about spaces. But i'll change it too.

test/functional/connection/connection.ts
```diff
@@ -351,5 +353,19 @@ describe("Connection", () => {
...
+    describe("skip schema generation when skipSchemaSync option is used", function() {
+
+        let connections: Connection[];
+        beforeEach(() => setupTestingConnections({ entities: [View] }).then(all => connections = all));
+        afterEach(() => closeConnections(connections));
+        it("database should be empty after schema sync", () => Promise.all(connections.map(async connection => {
+            await connection.syncSchema(true);
+            const queryRunner = await connection.driver.createQueryRunner();
+            let schema = await queryRunner.loadSchemaTables(["view"], new DefaultNamingStrategy());
+            expect(!schema.some( table => table.name === "view" )).to.be.true;
```
pleerock: I think `!schema.some( table => table.name === "view" )` expression can be written in more human readable way (because I cant easily understand what this code does)

pleerock:
thank you for your PR. can you please review comments once?

src/decorator/tables/ClassTableChild.ts
pleerock: oh then maybe simply do `!!(options && options.skipSchemaSync === true)` ?

apezel:
I added a filter function directly on EntityMetadataCollection, removed the spaces arround block functions and changed ```options && options.skipSchemaSync```

pleerock:
thanks ;)

PR #100, opened 2016-11-23
Added a transaction function that accepts multiple entity parameters

mingyang91:
Hi.
I wrote a transaction function, use the scene is operating multiple tables.
Please review

Use like the following code
```
connection.transaction([A, B, C...], ([ARepo, BRepo, CRepo...]) => {
  ARepo.persist...
  BRepo.reload...
  CRepo.remove...
})
```


pleerock:
Hi. There is no need in such function.

Transaction already accepts an entity manager:

```typescript
async transaction(runInTransaction: (entityManger: EntityManager) => Promise<any>): Promise<any>;
```

You can get access to any repository using entity manager.

mingyang91:
This ```Connection::transaction``` function copy by ```Repository::transaction``` function

I noticed that the transaction begin / commit / rollback Calling by the QueryRunner. If I were calling different Repository of the transaction, they were operating separately, they can be submitted in the same transaction?

pleerock:
Im sorry but I didnt understand what do you mean. Please try to use `connection.entityManager.transaction` method. Using it you can achieve same thing as you are proposing here.

mingyang91:
Oh, my mistake. Did not notice your reply. ```Transaction``` function from the ```EntityManager```

Issue #102, opened 2016-11-28
persist with cascade false try to save sub relationship 


apezel:
Hi,

It seems there's a problem when saving an object A that has one B that has many C : despite the cascades set to false, typeorm tries to cascade on C (but not on properties of B).

Example : 

```javascript
@Table("stat")
class Stat {
  @PrimaryColumn("int", { generated: true }) id: number;
  @ManyToOne(type => User)
  user: User;
};

@Table("user")
class User {
  @PrimaryColumn("int", { generated: true }) id: number;
  @ManyToMany(type => LexRight) @JoinTable({ name: "user_right" }) roles:LexRight[];
}
```

When i do : 

```javascript
  stat.user = myConnectedUser; // <-- loaded from db, with typeorm
  return await conn.getRepository(Model).persist(stat);
```
It fails with : 

```
executing query: START TRANSACTION
executing query: INSERT INTO stat(user) VALUES (?)
executing query: INSERT INTO user_right(userId, lexRightId) VALUES (?,?) -- PARAMETERS: [1,2]
executing query: INSERT INTO user_right(userId, lexRightId) VALUES (?,?) -- PARAMETERS: [1,3]
executing query: INSERT INTO user_right(userId, lexRightId) VALUES (?,?) -- PARAMETERS: [1,4]
query failed: INSERT INTO user_right(userId, lexRightId) VALUES (?,?) -- PARAMETERS: [1,2]
executing query: ROLLBACK
```

because this relations already exist. The problem exist with delete too : it doesn't delete stat.user but deletes every stat.user.roles. This is a blocking issue.

pleerock:
Hi,

Dont use PrimaryColumn/Column decorator in conjunction with relational decorators. If you want to make relation a primary you need to do:

```typescript
@Table("stat")
class Stat {
  @ManyToOne(type => User, { primary: true })
  user: User;
}
```

Regarding to your issue, can you try to reproduce it with `npm i typeorm@next` ? Because I have significantly refactored persistent mechanizm and save by cascades  there

apezel:
Thank. I think you misendurstood my example : the primary column of the stat table is ```id:number```

Ok i'll try with @next but can you confirm first that my skipSchemaSync option will be available in this release ? Because for now i'm using a custom build based on my commit, i need this option for my code to run.

pleerock:
> I think you misendurstood my example : the primary column of the stat table is id:number

[...]

PR #104, opened 2016-11-30
resolving problem when hydrating entity with date

apezel:
Problem referenced by #102

### DIFF
--- a/src/driver/mysql/MysqlDriver.ts
+++ b/src/driver/mysql/MysqlDriver.ts
@@ -252,6 +252,9 @@ export class MysqlDriver implements Driver {
                 return value ? true : false;
 
             case ColumnTypes.DATETIME:
+                if (value instanceof Date)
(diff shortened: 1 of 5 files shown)


src/driver/mysql/MysqlDriver.ts
```diff
@@ -252,6 +252,9 @@ export class MysqlDriver implements Driver {
                 return value ? true : false;
 
             case ColumnTypes.DATETIME:
+                if (value instanceof Date)
+                    return value;
+                    
                 if (columnMetadata.loadInLocalTimezone) {
```
apezel: It does make sense if you compare with old code : 

```javascript
if (value instanceof Date)
  return value;

return moment(value, "YYYY-MM-DD HH:mm:ss").toDate();
```

Can i help ?

pleerock:
I'll merge it and apply small changes...

Issue #41, opened 2016-09-24
json column type not supported


paistipoikka:
Hi,

I am trying to save an object into a column. Although the `json` column type is supported the `jsonb` column type in (i.e. postgres) is not. Can this please be added?

Cheers

JT

plantain-00:
Other than postgres, SQL Server 2016 and Mysql 5.7.8 support json data type now.
https://msdn.microsoft.com/en-us/library/dn921897.aspx
https://dev.mysql.com/doc/refman/5.7/en/json.html

pleerock:
It should be easy feature. Probably I'll implement it this week.

CreepGin:
Oh just saw the new comments. My PR was just a simple patch for Postgres to support `jsonb` type. If you need to do additional changes to MySql, etc... please have your go at it. :stuck_out_tongue_closed_eyes:

jcrben:
`json` and `jsonb` are different things: see https://www.postgresql.org/docs/9.4/static/datatype-json.html

I don't think SQL Server and MySQL support `jsonb`. Did a quick Google search and couldn't find mention of it. So this can probably be closed.

KerryRitter:
Any roadmap for this feature?

pleerock:
@KerryRitter json is already supported with postgres driver.

PR #126, opened 2016-12-14
Added support for Postgres jsonb type

CreepGin:
#41 

Please give me some pointers on how you would like me to write the tests for this. I just found out about typeorm today and am not sure about your test structure/requirements.

### DIFF
--- a/src/driver/postgres/PostgresDriver.ts
+++ b/src/driver/postgres/PostgresDriver.ts
@@ -206,6 +206,7 @@ export class PostgresDriver implements Driver {
                 }
 
             case ColumnTypes.JSON:
+            case ColumnTypes.JSONB:
                 // pg(pg-types) have done JSON.parse conversion
                 // https://github.com/brianc/node-pg-types/blob/ed2d0e36e33217b34530727a98d20b325389e73a/lib/textParsers.js#L170
                 return value;
@@ -236,6 +237,7 @@ export class PostgresDriver implements Driver {
             //     }
 
             case ColumnTypes.JSON:
+            case ColumnTypes.JSONB:
                 if (typeof value === "string") {
                     return JSON.parse(value);
                 } else {
--- a/src/driver/postgres/PostgresQueryRunner.ts
+++ b/src/driver/postgres/PostgresQueryRunner.ts
@@ -790,6 +790,8 @@ where constraint_type = 'PRIMARY KEY' and tc.table_catalog = '${this.dbName}'`;
                 }
             case "json":
                 return "json";
+            case "jsonb":
+                return "jsonb";
             case "simple_array":
                 return typeOptions.length ? "character varying(" + typeOptions.length + ")" : "text";
         }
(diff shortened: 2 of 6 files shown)


pleerock:
Hi. Thank you very much for your PR.

Test coverage is required. There is no guide for test requirements yet, I suggest to add your test into `test/functional/your-functionality` and implement tests. You need to cover at least functionality that saves and retrieves data in json format. Also you can use QueryRunner to get a table structure and check if column in the database has a correct type. Just take a look how other tests are implemented. Let me know if you have issues on how to run tests locally if you need that.

Also if you have time would be great if you can add JSON support for other drivers as requested in #41.

CreepGin:
Thanks! I have no problem running the tests locally (though I did spot a few oddities such as hard-coded MySql username+password).

I'll update this patch with the tests hopefully by the end of today. For this PR, I think I'll just focus on getting Postgres jsonb working correctly since us Postgres users really can't live without it. 

Also I do have a few additional questions concerning implementing real (non-text-based) JSON support for other DB's (such as "how are we doing version/compatibility checking here?").

CreepGin:
@pleerock Can you take a look? I'm not sure if there's something wrong with CI. Tests are running fine on my end (all passing).

pleerock:
> failing tests on CI

for some reason sometimes tests are failing on CI. I'll need to figure out what is wrong there. Just try to restart build if it failed for unknown reason

> how are we doing version/compatibility checking here?

right now we don't have versions limitations, so probably we can rely on latest versions of dbs, and maybe in the future change things if someone request it

> test

can you please also add simple check: persist entity with JSON object set in the json column has been succesfully converted and saved, then load it and make sure that your json object has been successfully transformed into javascript object?

CreepGin:
Sure, I added another test case.

Postgres has had real JSON support for ages, but MySql has just added the `json` data type this year. So maybe it's good to do some kind of compatibility check right now. 

I haven't used MySql for a long time. Maybe people using it can give some suggestions.

pleerock:
Great. Also need to add docs [here](https://typeorm.github.io/tables-and-columns.html#column-types) and add release notes [here](https://github.com/typeorm/typeorm/blob/master/CHANGELOG.md).

Also it will be great if you can investigate json/jsonb support for other databases too. Im sure its pretty quick. Other drivers need implementation before we can close the issue in tracker + dont really want to add in the docs notes that only postgres support it.

pleerock:
> So maybe it's good to do some kind of compatibility check right now.

I think we can add this feature, and if users will use json (if they do it and realize that its not working)  they will quickly understand that their version of mysql does not support it. We can add version check later, its not a big priority in this case (but if you want to add it go ahead to do it any time :D)

> mysql

what about other dbs, sql server?

CreepGin:
I'm fairly sure that `jsonb` is Postgres-specific. Everything in this patch are just for that. 

As for other DBs, I will have to resort to google as I never used Sql Server, Oracle, and MariaDB.

pleerock:
> Great. Also need to add docs here and add release notes here for next version - 0.0.6.

Hey @CreepGin , any updates on these so far?

pleerock:
Okay I'll add docs by myself, thank you for this PR!

pleerock:
@CreepGin can you take a look on [Subject#buildDiffColumns](https://github.com/typeorm/typeorm/blob/master/src/persistence/Subject.ts#L324) . are you sure you dont need similar code for JSONB type?

pleerock:
This code is used to determine if column value should  be updated or not. Without this code ORM can produce extra query each time when you persist already exist object without need for that

Issue #131, opened 2016-12-19
Limited queries without `where` throw error


subvertallchris:
If I am trying to return a limited number of records without also filtering on `where`, an illegal query is generated:

This query:

```
    const configsQuery = repo.createQueryBuilder('exercise_configs')
      .orderBy('exercise_configs.created_at', 'DESC')
      .setMaxResults(15);
    const configs = await configsQuery.getMany();
```

Gives this output:

```
SELECT DISTINCT("distinctAlias"."exercise_configs_id") as ids_id, "distinctAlias"."exercise_configs_created_at" FROM (SELECT "exercise_configs"."id" AS "exercise_configs_id", "exercise_configs"."created_at" AS "exercise_configs_created_at",  "exercise_configs"."type" AS "exercise_configs_type" FROM "exercise_configs" "exercise_configs" WHERE  AND type=$1) "distinctAlias" ORDER BY "distinctAlias"."exercise_configs_created_at" DESC LIMIT 15 -- PARAMETERS: ["ExerciseResultConfig"]
```

The issue is at the end, that `WHERE  AND type=$`. Including any `where` filter acts as a workaround:

```
    const configsQuery = repo.createQueryBuilder('exercise_configs')
      .where('1 = 1')
      .orderBy('exercise_configs.created_at', 'DESC')
      .setMaxResults(15);
    const configs = await configsQuery.getMany();
```

I see the same thing when using `find` with filters:

```
    // gives an error
    return await repo.find({ alias: 'exercise_configs', limit: 15 })
    // fixes the error
    return await repo.find({ alias: 'exercise_configs', where: '1 = 1', limit: 15 })
```

I am using STI, if that matters at all.

eduardoweiland:
I've found the same issue, but it's not related to limited queries. It happens with all queries with STI models that do not specify additional WHERE conditions.

It seems like an easy fix, and is probably caused by [this line](https://github.com/typeorm/typeorm/blob/08634b5c083bc76b38617109914b5db6f0d1bec3/src/query-builder/QueryBuilder.ts#L1479). I'm already working on a PR, I just need to write some tests for it.

pleerock:
thanks @eduardoweiland . I'll make a new release with this change at the end of this week.

pleerock:
released in `0.0.6`

PR #132, opened 2016-12-19
Fixed: error with single table inheritance query without additional conditions

eduardoweiland:
#131


pleerock:
Thank you @eduardoweiland !

PR #134, opened 2016-12-19
TIME is converted to 'HH-mm' instead of 'HH:mm'

cserron:
TIME was converted to 'HH-mm' instead of 'HH:mm', that was causing an error in the SQL query.

### DIFF
--- a/.idea/vcs.xml
+++ b/.idea/vcs.xml
@@ -1,6 +1,0 @@
-<?xml version="1.0" encoding="UTF-8"?>
-<project version="4">
-  <component name="VcsDirectoryMappings">
-    <mapping directory="$PROJECT_DIR$" vcs="Git" />
-  </component>
-</project>
--- a/.idea/vcs.xml
+++ b/.idea/vcs.xml
@@ -1,0 +1,6 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project version="4">
+  <component name="VcsDirectoryMappings">
+    <mapping directory="$PROJECT_DIR$" vcs="Git" />
+  </component>
+</project>
--- a/src/util/DataTransformationUtils.ts
+++ b/src/util/DataTransformationUtils.ts
@@ -18,7 +18,7 @@ export class DataTransformationUtils {
      */
     static mixedDateToTimeString(value: Date|any): string|any {
         if (value instanceof Date)
-            return value.getHours() + "-" + value.getMinutes() + "-" + value.getSeconds();
+            return value.getHours() + ":" + value.getMinutes() + ":" + value.getSeconds();
 
         return value;
     }
--- a/src/util/DataTransformationUtils.ts
+++ b/src/util/DataTransformationUtils.ts
@@ -14,7 +14,7 @@ export class DataTransformationUtils {
     }
 
     /**
-     * Converts given value into time string in a "HH-mm-ss" format.
+     * Converts given value into time string in a "HH:mm:ss" format.
      */
     static mixedDateToTimeString(value: Date|any): string|any {
         if (value instanceof Date)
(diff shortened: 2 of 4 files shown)


src/util/DataTransformationUtils.ts
```diff
@@ -18,7 +18,7 @@ export class DataTransformationUtils {
      */
```
cserron: Hi @eduardoweiland just changed the title of the PR to match the issue's tittle.
Thanks for your feedback. Cheers!

pleerock:
@cserron Thank you very much for this PR. I can merge it, however it will be better if you cover you case with test, so next time if someone make a change in this file, they will be aware of such functionality and how it should work. Just take a look how other tests are working, it should be pretty easy. Thank you again.

src/util/DataTransformationUtils.ts
eduardoweiland: Actually, I was referring to the comment in the code, line 17, that says `Converts given value into time string in a "HH-mm-ss" format`. GitHub did not allowed me to comment on the right line, sorry... :smile:

pleerock:
@cserron can you please add tests too?

cserron:
@pleerock I'd be glad if I knew how. Could you point me to some closed PR where that's been done in order to know how to reproduce a test? 
Thanks.

pleerock:
for example [this](https://github.com/typeorm/typeorm/pull/132/files) amazing PR by @eduardoweiland

cserron:
@pleerock ok, just pushed my test, hope is ok. Thanks!

test/github-issues/134/issue-134.ts
```diff
@@ -0,0 +1,31 @@
...
+
+
+    it("should successfully persist the post with creationDate in HH:mm and return persisted entity", () => Promise.all(connections.map(async connection => {
+        const post = new Post();
+        post.title = "Hello Post #1";
+        post.creationDate = new Date();
+        const returnedPost = await connection.entityManager.persist(post);
+
+        expect(returnedPost).not.to.be.empty;
+        returnedPost.should.be.equal(post);
```
pleerock: can you also add few more lines there: load post where post.id = returnedPost.id and assert if its creationDate is correct?

.idea/vcs.xml
```diff
@@ -0,0 +1,6 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
pleerock: please remove your working file, or add .idea into your global gitignore config

cserron:
Don't really understand how this unit tests work. Please check it out and send me your feedback @pleerock Thx.

test/github-issues/134/issue-134.ts
```diff
@@ -0,0 +1,44 @@
...
+        const savedPost = await postRepository.persist(post);
+        const loadedPost = await connection.entityManager
+            .createQueryBuilder(Post, "post")
+            .where("post.id=:id", { id: savedPost.id })
+            .getOne();
+
+
+        expect(post).not.to.be.empty;
+        expect(loadedPost).not.to.be.empty;
+        expect(loadedPost).to.be.eql(savedPost);
```
pleerock: The problem here that `creationDate` must be a string, not a `Date` object. There is a Date support in `time` column type, however it exist there just like an extra feature.

When you load a post, `creationDate` will be returned as a `string`, not `Date` object. Because column type = `time` is converted by orm to string, not a Date. Thats why 

In your assertion you are checking if loadedPost is equal to savedPost, however they aren't equal. Because typeof `savedPost.creationDate` === Date, and in `loadedPost.creationDate` === "string".

You should make a small correction in your test - you should assert only loadedPost.creationDate and check if its equal to `savedPost.creationDate.getDate() + "-" + savedPost.creationDate.getMonth()  + "-" + savedPost.creationDate.getYear();`. Something like that.

Also note that if you want support your `creationDate` Date object, you'll also need it to support a `string` type, because otherwise it wont be type-safe. Something like that:

`creationDate: Date|string`

pleerock:
take a look how I fixed issues in thouse commits

pleerock:
Thank you for your PR. This change will be released in next typeorm version very soon

PR #138, opened 2016-12-20
fix #137

### LINKED ISSUE #137: Cannot save relation entity.
In the following code: 
When I am persistent Job, category field will not be saved.
```
@Table()
export default class Category {

  @Column()
  public level3: string

  @PrimaryColumn('string', { length: '16' })
  public level3_number: string

  @OneToMany(type => Job, job => job.category)
  public jobs: Job[]

}


@Table()
export default class Job extends BaseTable {

  @ManyToOne(type => Category, category => category.jobs)
  public category: Category

}
```

I guess this may be related to the code
https://github.com/typeorm/typeorm/blob/master/src/persistence/Subject.ts#L371
```
            const updatedEntityRelationId: any =
                this.entity[relation.propertyName] instanceof Object ?
                    this.metadata.getEntityIdMixedMap(this.entity[relation.propertyName])
                    : this.entity[relation.propertyName];
```
I think here ```this.metadata.getEntityIdMixedMap``` should be changed to ```relation.inverseEntityMetadata.getEntityIdMixedMap```.

mingyang91:
If entity's primary column is different from the primary column of relation, ```buildDiffRelationalColumns``` can not give the correct result.

Because the ```getEntityIdMixedMap``` internal ```this.firstPrimaryColumn``` Entity's Column rather than the ```Relation``` of the column

### DIFF
--- a/src/persistence/Subject.ts
+++ b/src/persistence/Subject.ts
@@ -368,7 +368,7 @@ export class Subject {
             // todo: what to do if there is a column with relationId? (cover this too?)
             const updatedEntityRelationId: any =
                 this.entity[relation.propertyName] instanceof Object ?
-                    this.metadata.getEntityIdMixedMap(this.entity[relation.propertyName])
+                    relation.inverseEntityMetadata.getEntityIdMixedMap(this.entity[relation.propertyName])
                     : this.entity[relation.propertyName];
 
 


pleerock:
Thank you so much, I'll make release with this change at the end of this week.

Issue #114, opened 2016-12-09
Can not be parsed correctly the URL of pg.


mingyang91:
DATABASE_URL: postgres://username@ooo.xxx.com.cn:5432/database

```afterBase -> '/database1'``` **```preBase.substr(secondSlash) + 1 ```Here are some strange**
```afterBase.split("/")[0] -> ''``` **is correct?**

ERROR stack:
```
/Users/famer.me/Repository/vein/node_modules/typeorm/driver/postgres/PostgresDriver.js:70
            throw new DriverOptionNotSetError_1.DriverOptionNotSetError("database");
            ^
DriverOptionNotSetError: Driver option (database) is not set. Please set it to perform connection to the database.
    at new DriverOptionNotSetError (/Users/famer.me/Repository/vein/node_modules/typeorm/driver/error/DriverOptionNotSetError.js:13:28)
    at new PostgresDriver (/Users/famer.me/Repository/vein/node_modules/typeorm/driver/postgres/PostgresDriver.js:70:19)
    at ConnectionManager.createDriver (/Users/famer.me/Repository/vein/node_modules/typeorm/connection/ConnectionManager.js:361:24)
    at ConnectionManager.create (/Users/famer.me/Repository/vein/node_modules/typeorm/connection/ConnectionManager.js:94:27)
    at Object.<anonymous> (/Users/famer.me/Repository/vein/dist/entity/index.js:38:35)
    at Module._compile (module.js:571:32)
    at Object.Module._extensions..js (module.js:580:10)
    at Module.load (module.js:488:32)
    at tryModuleLoad (module.js:447:12)
    at Function.Module._load (module.js:439:3)
```
https://github.com/typeorm/typeorm/blob/master/src/driver/DriverUtils.ts#L67

pleerock:
released in `0.0.6`

PR #155, opened 2016-12-28
fix issue #114

mingyang91:


src/driver/DriverUtils.ts
```diff
@@ -50,21 +51,20 @@ export class DriverUtils {
...
 
         return {
-            host: host,
+            host: hostname,
             username: username,
             password: password,
             port: port ? parseInt(port) : undefined,
-            database: afterBase ? afterBase.split("/")[0] : undefined
+            database: path ? path.substr(1) : undefined
         };
```
pleerock: you should revert using url package. Previously I used url package, however I decided not to use it anymore and replaced with my own code. The reason is that typeorm can be used in the browser and we should avoid using third-party dependencies. Im sure what you did - same can be achieved with previous code + bugfix

mingyang91: OK

pleerock:
Thank you!

Issue #152, opened 2016-12-27
Can I change the (postgresql) default schema?


mingyang91:
I would like to design a SaaS multi-tenant software, the need to segregate tenant data. But I found `table_schema` can not be amended in `PostgresQueryRunner.ts`.
Can you add a way to modify `table_schema`?

pleerock:
no, there is no. However you can contribute here and make a `schemaName` extra option in `DriverOptions` and use this option in a `PostgresQueryRunner` class. It should be easy.

PR #160, opened 2016-12-30
Can change postgres default schema name

mingyang91:
add `schemaName` property, default value is `public`
Issue #152

### DIFF
--- a/ormconfig.json.dist
+++ b/ormconfig.json.dist
@@ -56,6 +56,19 @@
     }
   },
   {
+    "skip": false,
+    "name": "postgres-schema",
+    "driver": {
+      "type": "postgres",
+      "host": "localhost",
+      "port": 5432,
+      "username": "test",
+      "password": "test",
+      "database": "test",
+      "schemaName": "test-schema"
+    }
+  },
+  {
     "skip": true,
     "name": "mssql",
     "driver": {
--- a/test/functional/connection/connection.ts
+++ b/test/functional/connection/connection.ts
@@ -3,7 +3,7 @@ import {expect} from "chai";
 import {Post} from "./entity/Post";
 import {View} from "./entity/View";
 import {Category} from "./entity/Category";
-import {createTestingConnections, closeTestingConnections, setupSingleTestingConnection} from "../../utils/test-utils";
+import {createTestingConnections, closeTestingConnections, setupSingleTestingConnection, setupTestingConnections} from "../../utils/test-utils";
 import {Connection} from "../../../src/connection/Connection";
 import {Repository} from "../../../src/repository/Repository";
 import {TreeRepository} from "../../../src/repository/TreeRepository";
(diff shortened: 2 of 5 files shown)


test/functional/connection/connection.ts
```diff
@@ -359,5 +359,31 @@ describe("Connection", () => {
...
+        let connection: Connection;
+        afterEach(() => connection.isConnected ? connection.close() : {});        
+        it("schema name can be set", async () => {
+            const testSchema = "test-schema";
+            const testOption = setupSingleTestingConnection("postgres", {
+                name: "default",
+                entities: [Post]
+            });
+            // Only for testing
+            (testOption.driver as any).schemaName = testSchema;
```
pleerock: why did you set schema name this name, and didnt pass schema name in connection options?

mingyang91: Ok，modified.

ormconfig.json.dist
```diff
@@ -56,6 +56,19 @@
...
+    "driver": {
+      "type": "postgres",
+      "host": "localhost",
+      "port": 5432,
+      "username": "test",
+      "password": "test",
+      "database": "test",
+      "schemaName": "test-schema"
+    }
+  },
```
pleerock: oh noo, not this way.

mingyang91: So...what should i do?

test/functional/connection/connection.ts
```diff
@@ -359,5 +359,33 @@ describe("Connection", () => {
...
+                await connection.syncSchema(true);
+
+                const post = new Post();
+                post.title = "ChangeSchemaName";
+
+                const PostRepo = connection.getRepository(Post);
+                await PostRepo.persist(post);
+
+                const query = await connection.driver.createQueryRunner();
+                const schemaName = connection.driver.options.schemaName || "public";
```
pleerock: looks like your test is completely wrong, you enable "postgres" instead of "postgres-schema" and here you use `|| "public"` which fixes this mistake, thats why your test is passing. it actually checks if query is executed in public schema. Please remove `schemaName` variable and use `"test-schema"."post"` right in the query.

ormconfig.json.dist
pleerock: this is not necessary. Please remove it

test/functional/connection/connection.ts
```diff
@@ -359,5 +359,33 @@ describe("Connection", () => {
         })));
 
     });
+
+    describe("Can change postgres default schema name", () => {
+        let connections: Connection[];
+        beforeEach(async () => {
+            connections = await createTestingConnections({ 
+                enabledDrivers: ["postgres"],
+                entities: [Post]
```
pleerock: you just need to send `schemaName` just like `entities` are. You probably need to modify `test-utils`.

ormconfig.json.dist
pleerock: see my comment below about test-utils

test/functional/connection/connection.ts
mingyang91: In fact, here were tested to run the two connections. `EnabledDrivers` only select the type is postgres config.

mingyang91: If schemaName not set, it should be public. If set, it should be config.schemaName.

pleerock: but here in your test it will be always "public", because connection.driver.options.schemaName is not set because you are enabling wrong driver. just console.log(schemaName); and you'll see it

mingyang91: I think
```
> "test-schema" || "public"
"test-schema"
> undefined || "public"
"public"
```

and

console print `console.log(schemaName, connection.driver.options);` is
```
test-schema { type: 'postgres',
  host: 'localhost',
  port: 5432,
  username: 'test',
  password: 'test',
  database: 'test',
  schemaName: 'test-schema' }
public { type: 'postgres',
  host: 'localhost',
  port: 5432,
  username: 'test',
  password: 'test',
  database: 'test' }
```

mingyang91: code
console-print

test/functional/connection/connection.ts
mingyang91: I will be changed to this.

pleerock:
Lets merge it and I'll take a close look on it

pleerock:
If I simply add console.log:

```typescript
                const query = await connection.driver.createQueryRunner();
                const schemaName = connection.driver.options.schemaName || "public";
                console.log(schemaName);
                const rows = await query.query(`select * from "${schemaName}"."post" where id = $1`, [post.id]);
                expect(rows[0]["title"]).to.be.eq(post.title);
```

here is what I have:

```
Connection
    Can change postgres default schema name
public
      ✓ schema name can be set (118ms)
```

Which is wrong, because `test-schema` is actually tested.

pleerock:
Ive made a commit. Thats something I wanted, however there is still issue with readonly property. Not sure how to fix that efficiently

mingyang91:
It looks like your local ormconfig.json is older.

Issue #176, opened 2017-01-06
@CreateDateColumn and @UpdateDateColumn does not read back in UTC


ykim:
I'm currently using TypeORM 0.0.5, which is persisting into a Postgres 9.6 instance. When I create an entity with CreateDateColumn and UpdateDateColumn, it persists the date as an UTC timestamp correctly. However, when I used findOneById to get the entity, the corresponding fields are being parsed in local time. I have set both storeInLocalTimezone and loadInLocalTimezone to false. Is this supposed to be the expected behavior?

pleerock:
It looks like dates this is how underlaying drivers work. In [mysql](https://github.com/mysqljs/mysql#connection-options) there is  `timezone: "utc"` that can be used to make it work correctly. in postgres [here is issue](https://github.com/brianc/node-postgres/issues/783), somebody suggest to set machine time to timezone and problem will be resolved

pleerock:
I made a fix, it should work now as expected. Released in `0.0.7-alpha.6`. Please let me know if something is still wrong there

PR #178, opened 2017-01-08
Fixed missing zeros in date and date localization

### LINKED ISSUE #177: Date columns loses zeroes in sqlite
When saving column with a date or when using `@CreateDateColumn` or `@UpdateDateColumn`, it saves in this format: `2017-1-8 00:57:4` and sqlite doesn't like that (`DATETIME('2017-1-8 00:57:4')` returns null). Not sure how about other databases.

TypeORM: v0.0.6
NodeJS: v7.2.0

Luke265:
Fix #177


pleerock:
Hi! Thank you for this PR. You should cover your PR with tests otherwise code you have written may broke too soon by others. Also take a look why tests are failing right now. Thanks

pleerock:
Tests are failing because of tslint, `"` should be used instead of `'`

Luke265:
Tests passed. I've removed #134 test, because date, datetime and time columns are now deserialized into Date objects.

pleerock:
> date, datetime and time columns are now deserialized into Date objects.

its not correct. only datetime should be mapped into the Date object.
Problem with time that it cannot be mapped to date object because date object cannot exist with time and without date. But we cant set any random date to such object. Thats why we just operate with strings there. Same decision was made regarding to date type, because date object cant exist without time and setting time to some default value looks incorrectly. Thats why we choose to use only datetime for date object. Its by design and was discussed earlier in already closed github issues.

pleerock:
Ive made a fix that fixes #176 and conflicts with your changes in 4b09c4419af9acbc6a4031604bfd6045bce47567 sorry about that =(

package.json
```diff
@@ -74,6 +74,7 @@
     "@types/node": "^6.0.54",
     "app-root-path": "^2.0.1",
     "glob": "^7.1.1",
+    "moment": "^2.17.1",
```
pleerock: since typeorm is also can be used in browser it should be dependless from other packages to prevent complex configuration in the browser. In short words we cant use moment, we need to write small code that does what you need from moment

pleerock:
@Luke265 are there still issues that this PR resolves?

pleerock:
@Luke265 this PR is already outdated so Im going to close it. I think this issue is already resolved, please check it and let me know if its not if you are still interested in this issue.

PR #187, opened 2017-01-10
Allow empty host, if extra.socketPath is set

johncoffee:
I found this issue while deploying to google cloud, as they use unix sockets for mysql.

A workaround is just to define `host` as some string, because the mysql driver is preferring `socketPath` over host.

### DIFF
--- a/ormconfig.travis.json
+++ b/ormconfig.travis.json
@@ -1,4 +1,3 @@
-
 [
   {
     "skip": false,
--- a/ormconfig.travis.json
+++ b/ormconfig.travis.json
@@ -1,17 +1,5 @@
+
 [
-  {
-    "skip": false,
-    "name": "mysql",
-    "driver": {
-      "type": "mysql",
-      "username": "test",
-      "password": "test",
-      "database": "test",
-      "extra": {
-        "socketPath": "/var/run/mysqld/mysqld.sock"
-      }
-    }
-  },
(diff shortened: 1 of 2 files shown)


johncoffee:
Affirmative.

johncoffee:
@pleerock could you help me? Do you know if mysql on travis CI is binding to a unix socket? I have never used travis before :)

pleerock:
ahhh okay we probably can skip test for this case for now

pleerock:
wait you should have make a simple unit test. Just create a new mysql driver with and without socket url option and make sure that it works as expected without having to connect to mysql.

nishantkCB:
@johncoffee  , Can you please help me in connecting my node App enginee with sql cloud . I am using typeorm and with extra.socketPath in ormconfig file . 


Error : Error: connect ENOENT /cloudsql/[project:region:instance] at PipeConnectWrap.afterConnect [as oncomplete] (net.js:1113:14)

ormConfig.json  file contains below json . 
{
  "type": "mysql",
  "extra": {
    "socketPath": "/cloudsql/[project:region:instance]"
  },
  "username": "root",
  "password": "xxxx",
  "database": "yyy",
  "synchronize": true,
  "logging": false,
  "entities": [
    "build/entity/**/*.js"
  ]
}

thanks,
Nishant K

PR #189, opened 2017-01-10
Allow zero, false and empty string to be used as default value.

Luke265:


pleerock:
Thank you! Make sure to include tests!

PR #208, opened 2017-01-16
Single table inheritance 184

dcworldwide:
Test case for issue 184. Shows that we can't use the base class repository for CRUD in a STI schema.


pleerock:
failing btw

dcworldwide:
Yeah figured they'd pass once an issue is found... can defer merge if you want to keep it all green

PR #213, opened 2017-01-19
I made a mistake

mingyang91:
In the past pr #83 , I modified the wrong line.

### DIFF
--- a/test/functional/json/jsonb.ts
+++ b/test/functional/json/jsonb.ts
@@ -34,4 +34,27 @@ describe("jsonb type", () => {
         expect(foundRecord!.data.foo).to.eq("bar");
     })));
 
+    it("should persist jsonb string correctly", () => Promise.all(connections.map(async connection => {
+        await connection.syncSchema(true);
+        let recordRepo = connection.getRepository(Record);
+        let record = new Record();
+        record.data = JSON.stringify({ foo: "bar" });
+        let persistedRecord = await recordRepo.persist(record);
+        let foundRecord = await recordRepo.findOneById(persistedRecord.id);
+        expect(foundRecord).to.be.not.undefined;
+        expect(foundRecord!.data).to.be.a("string");
+        expect(JSON.parse(foundRecord!.data)!.foo).to.eq("bar");
+    })));
+
+    it("should persist jsonb array correctly", () => Promise.all(connections.map(async connection => {
+        await connection.syncSchema(true);
+        let recordRepo = connection.getRepository(Record);
+        let record = new Record();
+        record.data = [1, "2", { a: 3 }];
+        let persistedRecord = await recordRepo.persist(record);
+        let foundRecord = await recordRepo.findOneById(persistedRecord.id);
+        console.log("array", foundRecord!.data);        
+        expect(foundRecord).to.be.not.undefined;
+        expect(foundRecord!.data).to.deep.include.members([1, "2", { a: 3 }]);
+    })));
 });
--- a/test/functional/json/jsonb.ts
+++ b/test/functional/json/jsonb.ts
@@ -35,19 +35,17 @@ describe("jsonb type", () => {
     })));
 
     it("should persist jsonb string correctly", () => Promise.all(connections.map(async connection => {
-        await connection.syncSchema(true);
         let recordRepo = connection.getRepository(Record);
         let record = new Record();
-        record.data = JSON.stringify({ foo: "bar" });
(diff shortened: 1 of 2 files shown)


mingyang91:
This code：
```
                if (typeof value === "string") {
                    return JSON.parse(value);
                } else {
                    return value;
                }
```
If i have a JSON like `'"some string"'`, may cause some errors. Because the use of a re-call string JSON.parse is not correct.

And

In `preparePersistentValue` function
If JSON value is `[1, 2, 3]`, node-pg will throw the following exception.
```
query failed: INSERT INTO "competency"("tags") VALUES ($1)  RETURNING "id" -- PARAMETERS: [[1,2,3]]
error during executing query:error: invalid input syntax for type json
```

pleerock:
Hi! Thank you for the PR. Please add tests this time to make sure it works correctly. Also add test cases when you test where you are persisting arrays.

mingyang91:
OK, has been added

test/functional/json/jsonb.ts
```diff
@@ -34,4 +34,27 @@ describe("jsonb type", () => {
         expect(foundRecord!.data.foo).to.eq("bar");
     })));
 
+    it("should persist jsonb string correctly", () => Promise.all(connections.map(async connection => {
+        await connection.syncSchema(true);
```
pleerock: schema should be in sync already.

test/functional/json/jsonb.ts
```diff
@@ -34,4 +34,27 @@ describe("jsonb type", () => {
...
+        record.data = JSON.stringify({ foo: "bar" });
+        let persistedRecord = await recordRepo.persist(record);
+        let foundRecord = await recordRepo.findOneById(persistedRecord.id);
+        expect(foundRecord).to.be.not.undefined;
+        expect(foundRecord!.data).to.be.a("string");
+        expect(JSON.parse(foundRecord!.data)!.foo).to.eq("bar");
+    })));
+
+    it("should persist jsonb array correctly", () => Promise.all(connections.map(async connection => {
+        await connection.syncSchema(true);
```
pleerock: schema should be in sync already.

test/functional/json/jsonb.ts
```diff
@@ -34,4 +34,27 @@ describe("jsonb type", () => {
         expect(foundRecord!.data.foo).to.eq("bar");
     })));
 
+    it("should persist jsonb string correctly", () => Promise.all(connections.map(async connection => {
+        await connection.syncSchema(true);
+        let recordRepo = connection.getRepository(Record);
+        let record = new Record();
+        record.data = JSON.stringify({ foo: "bar" });
```
pleerock: Why do you stringify json here? I think `record.data` should access an object since its "json" column.

test/functional/json/jsonb.ts
```diff
@@ -34,4 +34,27 @@ describe("jsonb type", () => {
...
+        let persistedRecord = await recordRepo.persist(record);
+        let foundRecord = await recordRepo.findOneById(persistedRecord.id);
+        expect(foundRecord).to.be.not.undefined;
+        expect(foundRecord!.data).to.be.a("string");
+        expect(JSON.parse(foundRecord!.data)!.foo).to.eq("bar");
```
pleerock: Why do you manually parse json here? I think `record.data` should already return you a parsed object

test/functional/json/jsonb.ts
```diff
@@ -35,19 +35,17 @@ describe("jsonb type", () => {
     })));
 
     it("should persist jsonb string correctly", () => Promise.all(connections.map(async connection => {
-        await connection.syncSchema(true);
         let recordRepo = connection.getRepository(Record);
         let record = new Record();
-        record.data = JSON.stringify({ foo: "bar" });
+        record.data = "foo";
```
pleerock: wait, why are you storing string here? Isnt jsonb supposed to support json objects? Can't you do `record.data = { a: 1 };` then after you load it `foundRecord.data === { a: 1 }` ?

pleerock: adding @CreepGin

mingyang91: Just want to test a string type has not been repeatedly parse.

mingyang91: Because strings, arrays, numbers, Boolean belong to json.

pleerock: ahh now I understood the purpose. Thank you

Issue #215, opened 2017-01-20
Bug - invalid replacements of join conditions


netnexus:
Please see this example:

```typescript
const sql = connection.entityManager
    .createQueryBuilder(Post, "p")
    .leftJoinAndMapOne("p.author", Author, "n", "p.author_id = n.id")
    .leftJoinAndMapOne("p.abbreviation", Abbreviation, "ab", "p.abbreviation_id = ab.id")
    .getGeneratedQuery();
```
creates this query:

```sql
SELECT p.id AS p_id, p.author AS p_author, p.abbreviation AS p_abbreviation, n.id AS n_id, n.name AS n_name, ab.id AS ab_id, ab.name AS ab_name FROM post p LEFT JOIN author n ON p.author_id = n.id  LEFT JOIN abbreviation ab ON p.abbreviation.id = ab.id
```

Please notice the "p.abbreviation.id" which should be "p.abbreviation_id". Bug comes from a unescaped dot in a regex inside the QueryBuilder.

PR #216, opened 2017-01-20
encode dot in regex

netnexus:
Please see #215

### DIFF
--- a/test/github-issues/215/issue-215.ts
+++ b/test/github-issues/215/issue-215.ts
@@ -1,0 +1,32 @@
+import "reflect-metadata";
+import {createTestingConnections, closeTestingConnections, reloadTestingDatabases} from "../../utils/test-utils";
+import {Connection} from "../../../src/connection/Connection";
+import {Post} from "./entity/Post";
+import {Author} from "./entity/Author";
+import {Abbreviation} from "./entity/Abbreviation";
+
+describe("github issues > #215 invalid replacements of join conditions", () => {
+
+    let connections: Connection[];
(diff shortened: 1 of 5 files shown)


test/github-issues/215/issue-215.ts
```diff
@@ -0,0 +1,32 @@
...
+            .createQueryBuilder(Post, "p")
+            .leftJoinAndMapOne("p.author", Author, "n", "p.author_id = n.id")
+            .leftJoinAndMapOne("p.abbreviation", Abbreviation, "ab", "p.abbreviation_id = ab.id")
+            .getGeneratedQuery();
+
+        sql.should.be.equal("SELECT p.id AS p_id, p.author AS p_author, p.abbreviation AS " +
+            "p_abbreviation, n.id AS n_id, n.name AS n_name, ab.id AS ab_id, ab.name AS " +
+            "ab_name FROM post p LEFT JOIN author n ON p.author_id = n.id  LEFT JOIN " +
+            "abbreviation ab ON p.abbreviation_id = ab.id");
+    })));
```
pleerock: Maybe also good to to persist object with author and abbreviation and then select these data and check if returned result is correct?

pleerock: as you can see test is failing because on different platforms generated sql is different. Maybe its better to only check result?

netnexus: makes sense

pleerock:
Thank you for this fix! Released in `0.0.7-alpha.24`

PR #225, opened 2017-01-22
test for exception on missing host or socketPath 

johncoffee:


test/functional/connection/connection.ts
```diff
@@ -81,6 +82,39 @@ describe("Connection", () => {
...
+                    dropSchemaOnConnection: false,
+                    schemaCreate: false,
+                    enabledDrivers: ["mysql"],
+                });
+            }
+            catch (e) {
+                didThrow = true;
+            }
+            expect(didThrow).to.be.true;
+        });
```
pleerock: btw you can do 

```
expect(() => {
     connection = getConnectionManager().create(<ConnectionOptions>{
                     driver: {
                         "type": "mysql",
                         "username": "test",
                         "password": "test",
                         "database": "test",
                     },
                     entities: [],
                     entitySchemas: [],
                     dropSchemaOnConnection: false,
                     schemaCreate: false,
                    enabledDrivers: ["mysql"],
                 });

}).to.throw(Error);

File: test/functional/connection/connection.ts [hunk 2]
```diff
@@ -81,6 +82,39 @@ describe("Connection", () => {
...
+    describe("establishing connection", function() {
+        let connection: Connection;
+        after(() => {
+            if (connection && connection.isConnected)
+                return connection.close();
+
+            return Promise.resolve();
+        });
+
+        it("should throw DriverOptionNotSetError when extra.socketPath is missing", function() {
```
pleerock: probably correct statement is `when both "extra.socketPath" and "host" are missing`

test/functional/connection/connection.ts
```diff
@@ -81,6 +82,39 @@ describe("Connection", () => {
 
     });
 
+    describe("establishing connection", function() {
+        let connection: Connection;
+        after(() => {
+            if (connection && connection.isConnected)
+                return connection.close();
```
pleerock: will it even be connected in this test?

test/functional/connection/connection.ts
johncoffee: Changed it.

test/functional/connection/connection.ts
johncoffee: I left it to be overly secure. Now I removed it.

pleerock:
Thank you!

Issue #229, opened 2017-01-24
Broke mssql Azure connection


jmai00:
I was running fine with mssql driver from build 0.0.7-alpha.19 and earlier. With 0.0.7-alpha.20 and later, I don't get a connection to Azure DB.

In file src/driver/sqlserver/SqlServerDriver.ts, if remove the line 109:
```typescript
options.options = { useUTC: false };
```
 I get a connection.

pleerock:
can you provide more details, what kind of error do you have?

jmai00:
It doesn't do anything, I don't get any errors. But the **then** function never gets called. My code below: 
```typescript
return createConnection( <ConnectionOptions>{
            driver: {
                type: "mssql",
                host: config["host"],
                username: config["user"],
                password: config["password"],
                database: config["database"],
                extra: { options: { encrypt: true}},
            entities: [
                __dirname + "/db/typeorm/*.js",
            ],
            // autoSchemaSync: true,
            logging: {
                logQueries: true,
                // logOnlyFailedQueries: true,
            },
        }).then( (connection: Connection) => {
            this.connectionManager = getConnectionManager();
            this.alive = true;
            global["syslog"].info(`${this.name}: Connected`);
            return true;
        }).catch( (error) => {
            global["syslog"].warn(`${this.name}: ${error.message}`);
            return false;
        });
```

Oh. if i drop the **s** from attibute options, it works again. So there might be an attribute options you're conflicting for azure connection.
> options.option = { useUTC: false };

pleerock:
actually its being passed to underlying driver, [documentation about it here](https://github.com/patriksimek/node-mssql#drivers-1). So it should work. It also works with my local mysql installation. Not sure why its not working with azure.

jmai00:
Ok I figured out the problem. 
Your code blows away my setting in options.encrypt by setting a new object.
the code should be below:
```typescript
if (options.options) options.options.useUTC = false 
else  options.options = { useUTC: false }
```

pleerock:
ahh right. Do you want to create a PR for this change?

jmai00:
I'll submit the PR

pleerock:
will be released in `0.0.8-alpha.2`

PR #236, opened 2017-01-25
Add condition to set useUTC default option #229

jmai00:
Fix for #229 
Add condition to set useUTC default option if it doesn't exist


pleerock:
Thank you!

PR #239, opened 2017-01-26
Some fixes in MigrationCreateCommand

Tobias4872:
Hi,

despite adding the missing import for class "EntityManager", I propose some changes to meet some more strict tslint rules.

Please let me know if this will work for you.

Tobias.

### DIFF
--- a/src/commands/MigrationCreateCommand.ts
+++ b/src/commands/MigrationCreateCommand.ts
@@ -73,14 +73,16 @@ export class MigrationCreateCommand {
      * Gets contents of the migration file.
      */
     protected static getTemplate(name: string, timestamp: number): string {
-        return `import {MigrationInterface, QueryRunner, Connection} from "typeorm";
+        return `import {Connection, EntityManager, MigrationInterface, QueryRunner} from "typeorm";
 
 export class ${name}${timestamp} implements MigrationInterface {
 
-    async up(queryRunner: QueryRunner, connection: Connection, entityManager?: EntityManager): Promise<any> {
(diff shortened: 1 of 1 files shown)


src/commands/MigrationCreateCommand.ts
```diff
@@ -73,14 +73,16 @@ export class MigrationCreateCommand {
...
      */
     protected static getTemplate(name: string, timestamp: number): string {
-        return `import {MigrationInterface, QueryRunner, Connection} from "typeorm";
+        return `import {Connection, EntityManager, MigrationInterface, QueryRunner} from "typeorm";
 
 export class ${name}${timestamp} implements MigrationInterface {
 
-    async up(queryRunner: QueryRunner, connection: Connection, entityManager?: EntityManager): Promise<any> {
+    public async up(queryRunner: QueryRunner, connection: Connection, entityManager?: EntityManager): Promise<any> {
+        ;
```
pleerock: this is really disgusting tslint rule.

Tobias4872: We have not been setting this rule explicitly, it is part of the default rule set. In any case, would you still consider this pull request? Or do you need me to change anything?

pleerock: yes, of course we need to have `EntityManager` in imports. I'm okay with `public` keyword (however don't really like it). But having `;` is not acceptable. Please change only this part.

Tobias4872:
Done. Removed semicolons in empty functions.

pleerock:
Thank you!

PR #247, opened 2017-02-01
Some async problems (esp. when using sqlite)

netnexus:
After introducing a more complex entity in our application I found the application e2e tests failing sporadically with some SQLITE constraint problems during remove operations or less often with a "query failed: DROP TABLE "xyz" / "SQLITE_ERROR: no such table: main.my_referenced_table" when syncing/recreating the db schema.

This lead to some investigation of the async functions used in typeorm. Hope you can include the changes, but unfortunately it was not possible to write a test for the changes as the problem only arises sporadically. So please let me explain the changes here:

1. Changed log query time
For better debugging I have moved the query log statement into the promise. This way we will see the query log output when the query is executed, not when the promise with the query is being created. This is important, because without this change the log output is in a different order than the actually executed queries.

2. Made some remove operations sequentially
Some operations run concurrently but should not IMHO. Was the intention really to have them run in parallel? Please have a look at this simplified example how the `await Promise.all(array.map(async ...))` calls are executed. They are started sequentially, but directly after each other without waiting for the predecessor to complete. The `await Promise.all` just assures that all promises resolved before continuing. When removing entities with relations this can lead to an unexpected behavior. 

```typescript
const array = [1, 2, 3, 4, 5, 6];
async function testConcurrentAsyncMap() {
    await Promise.all(array.map(async (num) => {
        await new Promise<void>((resolve) => setTimeout(resolve, Math.random() * 100)); // random delay
        console.log(num);
    }));
}

testConcurrentAsyncMap();
```

> Returns
> 1
> 3
> 2
> 5
> 6
> 4

```typescript
const array = [1, 2, 3, 4, 5, 6];
async function testSeqAsync() {
    for (const num of array) {
        await new Promise<void>((resolve) => setTimeout(resolve, Math.random() * 100)); // random delay
        console.log(num);
    }
}

testSeqAsync();
```

> Returns
> 1
> 2
> 3
> 4
> 5
> 6

3. The previous behavior has some implications when creating the schema with sqlite, because we run recreateTable after each step in e.g. createForeignKeys. As said in 2. `Promise.all(this.entityToSyncMetadatas.map(async metadata => {` will run `recreateTable` multiple times at parallel, which can lead to a race condition when a referenced table was removed (in favor of being replaced with the temporary table in another thread soon).
To work around this issue the PR implements a simple lock/release mechanism for this method, so that it cannot be run concurrently.


pleerock:
Hi, thank you for your efforts and such detailed comment in this PR!

As I understood the problem is that some methods are called in parallel, which call `recreateTable` method, which executes in parallel. And because it executes in parallel - there are problems because it recreates the table and operations are overlapped. Right?

If so, then solution should be execution of those methods which call `recreateTable` in sequence, instead of parallel. I already faced with this problem which happened sometimes on me and made a fix in [this](https://github.com/typeorm/typeorm/commit/f689077c9869a9501e0de9e37ac8679403f95cb4) commit. So looks like I missed something.

Current implementation looks like a very rough monkey patch, which is not acceptable. If I understood the problem correctly you just need to make sure that all operations are executed in sequence, to do that I recommend to use `PromiseUtils.runInSequence` method instead of those patches.

netnexus:
Hi, thank you for the fast feedback. Please let us look at each issue separately:

1. log query time
Do you agree that it is better to log inside the promise? Are you ok with the provided changes?
see e.g. https://github.com/typeorm/typeorm/pull/247/files#diff-f449ebcfffaeef0bb22211a613c8e283R149

2. Remove operation failing
Your fix did not change anything in terms of the issue when removing entities. But I have adjusted to code to use the `PromiseUtils.runInSequence` method in favor of the `for (... of ...)` construct to have the code aligned with your fix. Only one small change is required, please see https://github.com/typeorm/typeorm/pull/247/files#diff-3bd548a8706c8ccbbe2d1d64edf56fc3R745

3. SchemaBuilder changes
My fault - my fork was not synced with your most recent changes with `PromiseUtils.runInSequence`. During my tests the problem only was present when using sqlite, because other drivers (except websql) do not have that method. That's why I decided to add a mutex to this method instead of sequencing the methods which call `recreateTable` (because it seemed to work fine for other drivers). However - your fix just works perfectly here, so I have reverted my changes on this topic, leaving only 1. and 2. in this PR.

Kind regards
Joe

pleerock:
Hi again, thank you for detailed description of each issue. Yes I agree with what you did and will merge this PR. Thank you!

Issue #271, opened 2017-02-08
Support for UUID columns in PostgreSQL (maybe other databases too)


MaikelH:
Currently the PostgreSQL system for TypeORM does not support UUID columns. (see[ PostgresQueryRunner.ts](https://github.com/typeorm/typeorm/blob/6897da035ea028b5214327d9540d27ee1a7c9dc0/src/driver/postgres/PostgresQueryRunner.ts#L746)

Since the UUID type is a supported type in PostgreSQL, TypeORM should also support it.

Some example code:

```typescript
@Entity()
export class Product {
    @PrimaryColumn('uuid')
    public id: string;
}
```

This code will error with the following output:
```
Specified type (uuid) is not supported by Postgres driver
```

pleerock:
Yes, we should add this column type. I'll implement it once I get more free time. But its easy feature, if you want to contribute - feel free to create PR for it.

PR #272, opened 2017-02-08
add support for UUID column type in PostgreSQL (#271)

MaikelH:
Support for UUID columns in PostgreSQL (see #271)


pleerock:
thank you! Will be released in next typeorm version. Let me know if you need this release urgently.

MaikelH:
For us currently the sooner would be better. We use use UUID's extensively in our application.

pleerock:
Released in `0.0.9-alpha.3`

pleerock:
yes should be so, also covered by #284

PR #277, opened 2017-02-11
Mongodb support

pleerock:


dcworldwide:
Huge pr great work. Hopefully mongo brings more developers to this lib

src/index.ts
```diff
@@ -10,8 +10,10 @@ import {ObjectType} from "./common/ObjectType";
 import {Repository} from "./repository/Repository";
 import {EntityManager} from "./entity-manager/EntityManager";
 import {PlatformTools} from "./platform/PlatformTools";
+import {TreeRepository} from "./repository/TreeRepository";
+import {MongoRepository} from "./repository/MongoRepository";
 
-// -------------------------------------------------------------------------
+// -----------------------------------------------------------------   --------
```
19majkel94: space-typo?

pleerock: looks like so, thanks

package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "typeorm",
   "private": true,
-  "version": "0.0.9",
+  "version": "0.1.0-alpha.1",
```
19majkel94: Why no SemVer? I understand the `-alpha.x` suffix but switching from 0.0.9 to 0.1.0 is MINOR release, so no breaking changes, just new optional features. So no removing `Table` decorators or changing from `getResults` to `getMany` between 0.0.2 and 0.0.3. It's really bad that users has to save excact version of TypeORM to avoid this kind of changes because they can't rely on semver.
Why are you afraid of releasing 1.0.0? There always will be stuffs to improve or change or add, so you can't stay in 0.x.x forever 😉

pleerock: no semver until 1.0.0 . Simply because I know that there will be lot of breaking changes. Lot of most popular API are already stable. But also lot of less popular but still needed features aren't. I'll go to 1.0.0 once I cover most of features with tests and implement most of features needed to implement plus make refactorings to make a code robust and maintainable I'll go with 1.0.0. You are right that there will be always be stuff to improve or change or add, however right now typeorm is in active development which means we still expect breaking changes. And I don't want 1.0.0 to be come 17.0.0 in one year of development.

pleerock: I know that it bad for users but I hope they understand that its hard to maintain such a complex big library to a single person **effectively**. So for now strategy is to try to make breaking changes only between 0.x versions and all other non breaking changes in 0.0.x versions.

19majkel94: Acutally, it's a semver 😄 There is a special treatment for 0.x.x releases, as you said - 0.MAJOR.MINOR. npm also supports it, however for releases 0.0.x it always stick to the version (like 0.0.12), but when `^0.1.0` it will accept all 0.1.x releases:
https://nodesource.com/blog/semver-tilde-and-caret/
So all is ok, for rapid development you can continue using 0.0.x (so no 0.1.0 just 0.0.10) if TypeORM is in early stage and you plan to do a lot of API changes in near feature 😉 npm will keep installing 0.0.9 for `^0.0.9` even if there is 0.0.23 or 0.5.12 available

pleerock: ahh semver is so complex 😫 Thanks for pointing out a special treatment for 0.x.x releases. However why you are saying that typeorm still should follow 0.0.10 and not be 0.1.0 ? As I understood 0.1.0 will be `0.MAJOR.MINOR` which means api changes on MAJOR level are acceptable and on minor levels are not. This is equal to what I was talking:

> So for now strategy is to try to make breaking changes only between 0.x versions and all other non breaking changes in 0.0.x versions.

valeriy-devpronet: in simple words.
let's say we have version `1.2.3` where:
`x.x.3` - is version for bug fixing or patching which does not affect public API;
`x.2.x` - is version where can be complex changes inside methods for public API, there may be marks that methods will be deprecated, but in that release they are still exists and there exists backward compatibility to deprecated methods;
`1.x.x` - is version which breaks backward compatibility, removes all methods which marked as deprecated add new big changes;

so even simple
```
 1.2.3
 ^ ^ ^
 ^ ^ ^----- PATCH
 ^ ^------- MINOR
 ^--------- MAJOR
```


So yes adding MongoDB this is big feature, public API was changed it may be major version in terms of semver

pleerock: thanks for the reply @valeriy-devpronet . right thats basics of semver. complexity adds additional conditions when you are under `1.0.0`. So in case of `0.x.y` - `x` is major and `y` is minor as @19majkel94 said.

19majkel94: > Using the caret for versions less than 0.1.0 offers no flexibility at all. Only the exact version specified will be valid. 
For example, ^0.0.3 will only permit only exactly version 0.0.3.

> For versions greater than or equal to 0.1.0, but less than 1.0.0, the caret adopts the same behaviour as a tilde and will allow flexibility in patch versions (only).
For example, ^0.1.3 will permit all versions from 0.1.3 to the next minor, 0.2.0.

So if you decide to go to 0.1.0, from now you have to keep semver, so 0.1.1 has to be minor release with no breaking changes. If you want to keep active development and introduce many changes, I would recommend to stick to 0.0.x because you don't have to worry about breaking changes - ^0.0.9 will install only 0.0.9, not the breaking things release 0.0.10 😉 Or you might have 0.23.0 before 1.0.0 or big, rarely updates with cumulative breaking changes.

pleerock: I think we can do so 🙏 :crossed_fingers:

pleerock: but someone needs to control me 🙈 🙉 🙊

19majkel94:
If the `Table` has been renamed to `Entity`, why `Column` hasn't be renamed to `Field`? Now it's look really weird, Mongo with column 😆

PR #284, opened 2017-02-13
Correct generated UUID behavior

seanski:
This pull request is a quick fix for generated UUID fields.  The `PrimaryGeneratedColumn` decorator works correctly, and newly inserted UUIDs are returned with the entity.

### DIFF
--- a/src/driver/postgres/PostgresQueryRunner.ts
+++ b/src/driver/postgres/PostgresQueryRunner.ts
@@ -290,7 +290,9 @@ where constraint_type = 'PRIMARY KEY' and tc.table_catalog = '${this.dbName}'`;
                 .filter(dbColumn => dbColumn["table_name"] === tableSchema.name)
                 .map(dbColumn => {
                     const columnType = dbColumn["data_type"].toLowerCase() + (dbColumn["character_maximum_length"] !== undefined && dbColumn["character_maximum_length"] !== null ? ("(" + dbColumn["character_maximum_length"] + ")") : "");
-                    const isGenerated = dbColumn["column_default"] === `nextval('${dbColumn["table_name"]}_id_seq'::regclass)` || dbColumn["column_default"] === `nextval('"${dbColumn["table_name"]}_id_seq"'::regclass)`;
+                    const isGenerated = dbColumn["column_default"] === `nextval('${dbColumn["table_name"]}_id_seq'::regclass)` 
+                        || dbColumn["column_default"] === `nextval('"${dbColumn["table_name"]}_id_seq"'::regclass)` 
+                        || /^uuid\_generate\_v\d\(\)/.test(dbColumn["column_default"]);
 
                     const columnSchema = new ColumnSchema();
                     columnSchema.name = dbColumn["column_name"];
@@ -532,7 +534,7 @@ where constraint_type = 'PRIMARY KEY' and tc.table_catalog = '${this.dbName}'`;
 
         // update sequence generation
         if (oldColumn.isGenerated !== newColumn.isGenerated) {
-            if (!oldColumn.isGenerated) {
+            if (!oldColumn.isGenerated && newColumn.type !== "uuid") {
                 await this.query(`CREATE SEQUENCE "${tableSchema.name}_id_seq" OWNED BY "${tableSchema.name}"."${oldColumn.name}"`);
                 await this.query(`ALTER TABLE "${tableSchema.name}" ALTER COLUMN "${oldColumn.name}" SET DEFAULT nextval('"${tableSchema.name}_id_seq"')`);
             } else {
@@ -849,9 +851,9 @@ where constraint_type = 'PRIMARY KEY' and tc.table_catalog = '${this.dbName}'`;
      */
     protected buildCreateColumnSql(column: ColumnSchema, skipPrimary: boolean) {
         let c = "\"" + column.name + "\"";
-        if (column.isGenerated === true) // don't use skipPrimary here since updates can update already exist primary without auto inc.
+        if (column.isGenerated === true && column.type !== "uuid") // don't use skipPrimary here since updates can update already exist primary without auto inc.
             c += " SERIAL";
-        if (!column.isGenerated)
+        if (!column.isGenerated || column.type === "uuid")
             c += " " + column.type;
         if (column.isNullable !== true)
             c += " NOT NULL";
@@ -870,6 +872,8 @@ where constraint_type = 'PRIMARY KEY' and tc.table_catalog = '${this.dbName}'`;
                 c += " DEFAULT " + column.default + "";
             }
         }
+        if (column.isGenerated && column.type === "uuid" && !column.default)
+            c += " DEFAULT uuid_generate_v4()";
         return c;
     }
 
--- a/test/functional/uuid/uuid.ts
+++ b/test/functional/uuid/uuid.ts
@@ -7,10 +7,16 @@ import {createTestingConnections, closeTestingConnections} from "../../utils/tes
 describe("uuid type", () => {
 
     let connections: Connection[];
-    before(async () => connections = await createTestingConnections({
-        entities: [Record],
-        enabledDrivers: ["postgres"] // because only postgres supports jsonb type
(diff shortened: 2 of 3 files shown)


seanski:
Not sure why the build keeps hanging.  The tests are passing... not sure what I'm doing wrong.

src/driver/postgres/PostgresQueryRunner.ts
```diff
@@ -870,6 +870,8 @@ where constraint_type = 'PRIMARY KEY' and tc.table_catalog = '${this.dbName}'`;
                 c += " DEFAULT " + column.default + "";
             }
         }
+        if (column.isGenerated && column.type === "uuid")
```
pleerock: if column.default will be set it will break

src/driver/postgres/PostgresQueryRunner.ts
```diff
@@ -870,6 +870,8 @@ where constraint_type = 'PRIMARY KEY' and tc.table_catalog = '${this.dbName}'`;
...
+            c += " DEFAULT uuid_generate_v4()";
```
pleerock: is it only can be `uuid_generate_v4`? if there other functions like uuid_generate_v6 or something similar? If yes, then whats possibility that users may want to use that instead?

seanski: There are several other uuid functions, but everyone other ORM I have used uses the v4 function.  https://www.postgresql.org/docs/9.5/static/uuid-ossp.html... Fixing the other issue you brought up, still allowing for a default value to be set when generated, should give users the ability to use another uuid function.

test/functional/uuid/jsonb.ts
```diff
@@ -7,10 +7,17 @@ import {createTestingConnections, closeTestingConnections} from "../../utils/tes
...
-    }));
+    before(async () => {
+        connections = await createTestingConnections({
+            entities: [Record],
+            enabledDrivers: ["postgres"] // because only postgres supports jsonb type
+        });
+
+        connections.forEach(async connection => {
+            let queryRunner = await connection.driver.createQueryRunner();
+            await queryRunner.query(`CREATE extension IF NOT EXISTS "uuid-ossp"`);
```
pleerock: when you are creating your own query runner you need to release it. Probably thats why your test is hanging. But, probably better here just to call `connection.entityManager.query()` instead

test/functional/uuid/jsonb.ts
```diff
@@ -7,10 +7,17 @@ import {createTestingConnections, closeTestingConnections} from "../../utils/tes
...
 
     let connections: Connection[];
-    before(async () => connections = await createTestingConnections({
-        entities: [Record],
-        enabledDrivers: ["postgres"] // because only postgres supports jsonb type
-    }));
+    before(async () => {
+        connections = await createTestingConnections({
+            entities: [Record],
+            enabledDrivers: ["postgres"] // because only postgres supports jsonb type
```
pleerock: change comment - not jsonb, but uuid

src/driver/postgres/PostgresQueryRunner.ts
```diff
@@ -849,9 +849,9 @@ where constraint_type = 'PRIMARY KEY' and tc.table_catalog = '${this.dbName}'`;
      */
     protected buildCreateColumnSql(column: ColumnSchema, skipPrimary: boolean) {
         let c = "\"" + column.name + "\"";
-        if (column.isGenerated === true) // don't use skipPrimary here since updates can update already exist primary without auto inc.
+        if (column.isGenerated === true && column.type !== "uuid") // don't use skipPrimary here since updates can update already exist primary without auto inc.
             c += " SERIAL";
-        if (!column.isGenerated)
+        if (!column.isGenerated || (column.isGenerated === true && column.type === "uuid"))
```
pleerock: actually this expression can be simplified to `if (!column.isGenerated || column.type === "uuid")` and will work same and be easier to read

test/functional/uuid/jsonb.ts
pleerock: another question is, why need to execute this statement, do user always need to do this before he is going to use uuid? Isnt there a way not to do this and use uuid?

test/functional/uuid/jsonb.ts
```diff
@@ -7,10 +7,16 @@ import {createTestingConnections, closeTestingConnections} from "../../utils/tes
...
-        entities: [Record],
-        enabledDrivers: ["postgres"] // because only postgres supports jsonb type
-    }));
+    before(async () => {
+        connections = await createTestingConnections({
+            entities: [Record],
+            enabledDrivers: ["postgres"] // because only postgres supports uuid type
+        });
+
+        connections.forEach(async connection => {
```
pleerock: also you probably know that you are not awaiting here execution of `query` ? Because if you want to await it here you need to do: `await Promise.all(connections.map(async connection => ... ));` instead

test/functional/uuid/uuid.ts
```diff
@@ -7,10 +7,16 @@ import {createTestingConnections, closeTestingConnections} from "../../utils/tes
...
-        enabledDrivers: ["postgres"] // because only postgres supports jsonb type
-    }));
+    before(async () => {
+        connections = await createTestingConnections({
+            entities: [Record],
+            enabledDrivers: ["postgres"] // because only postgres supports uuid type
+        });
+
+        await Promise.all(connections.map(async connection => {
+            await connection.entityManager.query(`CREATE extension IF NOT EXISTS "uuid-ossp"`);
```
pleerock: now you even dont need async/await you can just do `connection => connection.entityManager.query(`CREATE extension IF NOT EXISTS "uuid-ossp"`);`

pleerock: or

```
connection => {
     return connection.entityManager.query(CREATE extension IF NOT EXISTS "uuid-ossp");
}
```

test/functional/uuid/jsonb.ts
seanski: In order to use the generate_uuid functions, you have to enable the uuid-ossp extension.  You may be able to configure travis to do this, but, even still, anyone who runs the tests locally will need to enable them manually.  It may be worth having the schema synchronizer do this automatically, but I haven't even looked at how that works yet.

test/functional/uuid/uuid.ts
```diff
@@ -7,10 +7,16 @@ import {createTestingConnections, closeTestingConnections} from "../../utils/tes
...
-        enabledDrivers: ["postgres"] // because only postgres supports jsonb type
-    }));
+    before(async () => {
+        connections = await createTestingConnections({
+            entities: [Record],
+            enabledDrivers: ["postgres"] // because only postgres supports uuid type
+        });
+
+        await Promise.all(connections.map(connection => {
+            connection.entityManager.query(`CREATE extension IF NOT EXISTS "uuid-ossp"`);
```
pleerock: you missed return!

pleerock: another question is, why need to execute this statement, do user always need to do this before he is going to use uuid? Isnt there a way not to do this and use uuid?

seanski:
@pleerock -- I answered this above, but it got hidden in an outdated diff: 

In order to use the generate_uuid functions, you have to enable the uuid-ossp extension. You may be able to configure travis to do this, but, even still, anyone who runs the tests locally will need to enable them manually. It may be worth having the schema synchronizer do this automatically, but I haven't even looked at how that works yet.

pleerock:
yeah, if users will use uuid and won't work they may ask question why. The best way to check if there any entity using uuid column and enable this extension. This should be easy fix

seanski:
Okay, cool... I'll look into that now.  Thanks!

galkowskit:
I just created a UUID `@PrimaryGeneratedColumn({ type: 'uuid' })`. It doesn't run `CREATE EXTENSION "uuid-ossp";` out of the box. After running this manually - it does work correctly.

pleerock:
right, we need additional code to determine if there is a uuid column used and run this query on app start. You can implement this feature or create an issue and wait until somebody will do that

PR #291, opened 2017-02-16
Adding count method

### LINKED ISSUE #290: Count method
I would LOVE to see a `count` method on the repositories, just just returns how many would match. Wouldnt need to any of the extra stuff that find does

aequasi:
Resolves #290


pleerock:
Tests are failing. Also make sure to add `count` method to entity manager as well

pleerock:
Awesome! Thank you! Will be released in next version

Issue #329, opened 2017-03-03
Adding .select() to createQueryBuilder return undefined


valeriy-devpronet:
```typescript
@EntityRepository(SomeEntity)
export class SomeEntityRepository extends Repository<SomeEntity>
{
   async handleThis()
  {
      const q = this.createQueryBuilder('e')
            .select(['e.id', 'e.anotherProperty'])
            //or 
            //.select('e.id, e.anotherProperty')
            //or even
            //.select('e.id')
            .addOrderBy('e.createdAt', 'DESC')
            .getOne();
  }
}
```
`q` is always `undefined`

but if I remove `.select()` or if i'm using it without parameters it works. But I do not need a complete list of columns in the result.
I did not found how to use `.select()` in the right way nor in examples neither in source.
Thanks.

pleerock:
> But I do not need a complete list of columns in the result.

this feature of partial selection is not supported yet. However it will be supported in next version. Adding @AlexMesser

valeriy-devpronet:
I saw that mentions in another issues. Seems like this issue #245  relates also to that enhancement and if i'm not mistaken this also #296. Am I correct?

pleerock:
partial selection is already supported in latest typeorm versions.

valVk:
does it in stable release or in unstable?

pleerock:
there are some features broken (like cascades and class-table and single-table inheritances, but they did not work well before too), as well lot of bugfixes and new features.

JaosnHsieh:
having same issue here. current work around is to comment out partial select.

from

```typescript
      Device.createQueryBuilder('device')
      .innerJoinAndSelect('device.xxx', 'xxx', 'xxx.yyyy = :s', {
        s: '123',
      })
      .innerJoinAndSelect('device.a', 'a')
      .innerJoinAndSelect('device.b', 'b')
      .select(['a.id',', 'a.id', 'a.name'])
      .getOne();
```


to

```typescript
  Device.createQueryBuilder('device')
      .innerJoinAndSelect('device.xxx', 'xxx', 'xxx.yyyy = :s', {
        s: '123',
      })
      .innerJoinAndSelect('device.a', 'a')
      .innerJoinAndSelect('device.b', 'b')
//      .select(['a.id',', 'a.id', 'a.name'])
      .getOne();
```

versions
```
"pg": "^7.3.0",
 "typeorm": "0.2.25"
```

Issue #245, opened 2017-01-30
Select columns to list on mapped entities


Tobias4872:
Hi,

when I used "innerJoinAndMapOne" to join in let's say a parent element. Can I - e.g. - decide that I only want to see the "id" and the "name" column of that parent element in the map instead of the complete element?

Thanks,
   Tobias.

pleerock:
no, partial selection is not supported yet, but I'm planning to implement it the future versions of orm.

Tobias4872:
OK, thanks. Not sure if you want to close or keep this. Up to you.

ascoders:
Hello @pleerock ,I find that if use `.select` method, it will override all select sql sentence:

```typescript
const user = await this.userRepository.createQueryBuilder('user')
            .where('user.id=:id', {id: Number(req.session['userId'])})
            .leftJoinAndMapOne('user.gameSimulatedPlanetUser', GameUser, 'gameUser', 'gameUser.user=user.id')
            .leftJoinAndMapMany('gameUser.planets', GamePlanet, 'gamePlanet', 'gameUser.id=gamePlanet.gameUser')
            .leftJoinAndMapMany('gamePlanet.buildings', GameBuilding, 'gameBuilding', 'gamePlanet.id=gameBuilding.planet')
            .select(['user.id as user_id']) // amazing
            .getSql()
```

And then i got a sql like `SELECT user.id as user_id FROM user user LEFT JOIN...`, and do not support syntax like `gameUser.something as gameUser_something`. Hope that the future can be supported!

Another question is that i don't know why this is true `.select(['user.id as user_id'])`, and `select['user.id']` doesn't,  may be a magic.

pleerock:
you can use such code temporary until partial selection feature will come.

pleerock:
Partial selection is already implemented in latest versions, so I'll close this issue.

vsainiwelt:
what about if i want to select a column of gameUser.planets

from query as below
```
const user = await this.userRepository.createQueryBuilder('user')
            .where('user.id=:id', {id: Number(req.session['userId'])})
            .leftJoinAndMapOne('user.gameSimulatedPlanetUser', GameUser, 'gameUser', 'gameUser.user=user.id')
            .leftJoinAndMapMany('gameUser.planets', GamePlanet, 'gamePlanet', 'gameUser.id=gamePlanet.gameUser')
[...]

PR #330, opened 2017-03-04
implemented partial select, issues #329 #245

AlexMesser:


valVk:
Hi, which release will contain these changes?

pleerock:
once QueryBuilder refactoring will be finished #334

valVk:
Great, thanks for clarification.

PR #337, opened 2017-03-09
refactored query builder - refactored JOIN part

AlexMesser:


johncoffee:
Hi @AlexMesser. Looks like this pull requests could be a super improvement. I would love to review all the changes to know what has improved, but it's impossible with 500+ files changed! Is it possible to avoid unnecessary renaming, to produce less changed files? I can't see the value in renaming `connection.entityManager` to `connection.manager`, or `generatedColumn.fullName` => `generatedColumn.databaseName`, etc.?
Otherwise keep up the good work.

pleerock:
Hi @johncoffee all changes made on this branch we made together with @AlexMesser (for example renaming connection.entityManager to connection.manager is what I did to simplify naming for users). We are refactoring codebase by the way, and its impossible to do good refactoring to produce better code style and structured code without having to renaming and move things around. 

Yeah I know this PR is hard to review, and this is because of all refactoring and changes we made from March. @AlexMesser and me will try to make smaller commits so for everyone else it will be easier to review this commit-by-commit.

PR #339, opened 2017-03-12
Fractional seconds#267

jmai00:
Adding Fraction seconds length option to datetime to possible DB:
- mysql
- mssql
- postgres


pleerock:
Hi, thank you for this PR. Why there are two PRs? #339 and #340 ? As I understood it should be a single PR, because second fixes this one's tests? Can you merge them?

src/driver/mysql/MysqlQueryRunner.ts
```diff
@@ -709,23 +709,37 @@ export class MysqlQueryRunner implements QueryRunner {
...
-
                 } else {
                     return "decimal";
-
                 }
             case "date":
-                return "date";
+                if (!typeOptions.length) {
+                    return "date";
+                } else if (typeOptions.length <= 6) {
```
pleerock: It seems to me that we should not put this logic, because if user will set length to `10` but database will generate him `date(6)` he will be confused why its not generating him proper value, and he will not be aware of it. Probably better just to pass value back to db and left its validation on it. Alternatively we can make our own validation during schema create time.

pleerock:
can #340 be closed? Also looks like tests are still failing

pleerock:
FYI still is failing....

jmai00:
I need help trying to figure out why it's failing. If the test is running against other DB, then i can't confirm the results that it should. But my unit test is from mysql db.

Issue #267, opened 2017-02-07
Support Time, Datetime, Timestamp for fractional seconds


jmai00:
The current TypeORM version doesn't support length in the column options for type time, timestamp, and datetime. MySql supports this use for storing decimal fractional seconds.

Time(3)
Timestamp(3)
Datetime(6)

pleerock:
This should be easy feature for contribution :stuck_out_tongue_winking_eye:

jmai00:
yes i'm working on it.

rmoore-a3-digital:
I'd like to add a +1 for getting this into the shipping release.

jmai00:
Wow lots of things have changed, seems for the better. Great work on refactoring. I've been waiting for these changes. I saw the fractional fix for dates but there are situations not handled in the millisecond padding. I'm trying to submit a change.
1) since js only natively supports up to milliseconds, the padding doesn't handle when 0 or less than 10
2) there's a mix up on the driver parameter datetime. The creationdate and updatedate has precision parameter while in the mssql parameter says for time, datetime2, datetimeoffset uses scale. I'm trying to understand which one will be the convention? It seems you're going with precision with the latest updates.

Thanks.

pleerock:
I'll close this issue because @AlexMesser added fractional seconds support and precision as well. Feel free to open another issue if you have some problems with latest changes.

> precision parameter while in the mssql parameter says for time, datetime2, datetimeoffset uses scale

it uses `precision` everywhere including official documentation. It does not use `scale` word.

PR #340, opened 2017-03-12
Fix gh issue 176 unit test

jmai00:
Minor fix for the rounding error from cascading change of fractional seconds in gh issue #267


jmai00:
Merge with #339

PR #342, opened 2017-03-13
Move @types/mongodb to devDependencies

Spoki4:
I transpile typescript code to es6 and install only dependencies without devDependencies. But I whatever install types.

### DIFF
--- a/package.json
+++ b/package.json
@@ -36,7 +36,8 @@
     "websql-orm"
   ],
   "devDependencies": {
-    "@types/chai": "^3.4.35",
+    "@types/chai": "^3.4.35",    
+    "@types/mongodb": "^2.1.41",
     "@types/chai-as-promised": "0.0.29",
     "@types/mocha": "^2.2.39",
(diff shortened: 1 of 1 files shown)


pleerock:
We can't move this to dev dependency because users projects will throw compiler errors because of missing mongodb dependencies in mongodb driver.

Spoki4:
What I can do if I don't wont have a types in dependencies on my production server? But they are installing.

pleerock:
I understand the issue, however there is no good solution for this yet. One solution is to refactor typeorm completely and make it modular however that requires lot of time to think of its new design and implement it. This is planned for the future. For now you have to ignore the fact that you have extra package with typings on your environment. In fact they aren't add so much.

Spoki4:
I'm a perfectionist, but I'm ready to sacrifice this. Thx you for answer.

pleerock:
yeah me too, however we both know that nothing is perfect and there always not perfect choices and tradeoffs we forced to choose. Thanks for PR.

PR #358, opened 2017-03-20
Export ObjectType and ObjectLiteral

CWSpear:
I was creating an API that has a base service class to do common rest commands (i.e. `get`, `find`, `create`), etc, and I was trying to do something like this:

```ts
import { getConnection, EntityManager, ObjectType } from 'typeorm';

export abstract class AbstractService<T> {
  protected entityManager: EntityManager;

  constructor(private entityClass: ObjectType<T>) {
    this.entityManager = getConnection().entityManager;
  }

  async find(params): Promise<T[]> {
    return this.entityManager.find(this.entityClass);
  }
  // ... other methods...
}

// usage in another file:
export class UserService extends AbstractService<User> {
  constructor() {
    super(User);
  }
}
```

but `ObjectType` isn't exported, so TypeScript was complaining. It wouldn't let me do `import { ObjectType } from 'typeorm/common` either (`module not found`).

This PR adds them as exports so that the above is possible without TypeScript complaining.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -18,6 +18,8 @@ import {MongoRepository} from "./repository/MongoRepository";
 // -------------------------------------------------------------------------
 
 export * from "./container";
+export * from "./common/ObjectType";
+export * from "./common/ObjectLiteral";
 export * from "./decorator/columns/Column";
 export * from "./decorator/columns/CreateDateColumn";
 export * from "./decorator/columns/DiscriminatorColumn";


CWSpear:
Thank you for this! Would it be possible to get a version published with this change? :-)

pleerock:
its merged to master, however there are few changes on master that I wanted to finish before publish a new version... Now sure what is the best way to do now - publish new version with everything I had made before but without finishing some of them, or cherry pick your commit on separate branch with old revision and publish a new version with old changes + your commit....

pleerock:
If you need a new version urgently I'll do the last

CWSpear:
It's not terribly urgent. I think in general, it may be wise to have a strategy to handle publishing new versions that doesn't mess with other ongoing work. Perhaps a `develop` branch as the default branch on GitHub and the `master` branch is the production/publish branch, then you cherry pick/merge commits from `develop` as they're ready. That way doc changes and other minor (but important) stuff like these PRs aren't blocked behind other work.

pleerock:
You are completely right. I was using single branch always simply to move faster. I didn't expect that I'll delay publishing of a new version. But I did because of some internal problems. I'll change a strategy and will maintain multiple branches in the near future.

CWSpear:
👍 thanks!

PR #364, opened 2017-03-22
Persist/Remove multiple unrelated entities in one entityManager call

### LINKED ISSUE #363: EntityManager can't save/delete 2 unrelated entity types in a single persist/remove call
TypeORM: 0.9
Node: 6.10.0
Postgres: 9.6.2

Consider 2 entities that have no relation and 2 fields, `id` and `name`:

```ts
const banana = new Banana();
banana.name = 'The Awesome Banana';

const apple = new Apple();
apple.name = 'The Awesome Apple';

await connection.entityManager.persist([banana, apple]);
```

I would expect this to save a row in the `apple` table and a row in the `banana` table, but it saves 2 rows in the `banana` table.

This _may_ be intentional, but seems like a bug to me? Especially looking at the code, it seemed like it wasn't doing anything that should cause it to do this behavior.

CWSpear:
This changes the behavior of `entityManager` to correctly handle persisting and removing multiple unrelated entities in one call.

Resolves #363

### DIFF
--- a/src/entity-manager/EntityManager.ts
+++ b/src/entity-manager/EntityManager.ts
@@ -103,7 +103,9 @@ export class EntityManager extends BaseEntityManager {
                     if (target.length === 0)
                         return Promise.resolve(target);
 
-                    return this.getRepository<Entity[]>(target[0].constructor).persist(entity as Entity[], options);
+                    return Promise.all(target.map((t, i) => {
+                        return this.getRepository<Entity>(t.constructor).persist((entity as Entity[])[i], options);
+                    }));
                 } else {
                     return this.getRepository<Entity>(target.constructor).persist(entity as Entity, options);
                 }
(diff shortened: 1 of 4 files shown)


src/entity-manager/EntityManager.ts
```diff
@@ -178,7 +180,9 @@ export class EntityManager extends BaseEntityManager {
         } else {
             // todo: throw exception if constructor in target is not set
             if (target instanceof Array) {
-                return this.getRepository<Entity[]>(target[0].constructor).remove(entity as Entity[], options);
+                return Promise.all(target.map((t, i) => {
```
pleerock: i know its a short peace of code, but please prefer longer and more clear code in such places as well

pleerock:
Thank you so much!

src/entity-manager/EntityManager.ts
CWSpear: I do not understand. Are you referring to `t` and `i`? If we're discussing trying to make the code more clear, let's talk about how `target` and `entity` are singular words, but here in the code, they actually both arrays (so `targets` and `entities`) and referring the _same_ array... it's as confusing as heck 😄.

I thought about trying to clear that up a bit, but I try to make as few changes as I can in PRs for large projects if I can help it.

pleerock: I don't like short variables like `t` and `i`, and prefer to name them full name, e.g. `target` or `targetItem`, or simply `item` and `index`, instead of `t` and `i`. But its not that important, I can do it by myself next time I face this code. Thank you again

CWSpear: You already merged it, but normally I don't mind updating! Don't be afraid to ask for things on PRs to lessen your load, even if you think they're minor :-)

Anyway, thanks for merging.

pleerock: its really not something important. But if you want to lessen my load there are bunch of issues needs to be fixed from easy to hard. Just let me know or pick any 😃

Issue #343, opened 2017-03-13
Datatype: Fixed character char(100)


navossoc:
Hi, I like to know if it is possible to create a `character` column on postgres.

I was reading the docs, but I didn't find anything related to create a column with fixed length.
Am I missing something here or it's not possible at the moment?

Is there any chances this data type will be added?

```sql
CREATE TABLE users (
    id integer NOT NULL,
    login character varying(100) NOT NULL,
    hash character(64) NOT NULL
);
```

Thanks.

pleerock:
you can do that:

```typescript
@Entity("users")
export class User {
     // ....

     @Column({ length: 100 })
     login: string;

}
```

navossoc:
@pleerock Yes, but that result in `character varying(100)`.

I'm trying to achieve the `character` type, like on the hash column.

pleerock:
ah no, there is no separate type for character. I will do that in the future once I get more free from more priority tasks. If you have a time you can contribute, its pretty easy feature to implement.

navossoc:
Sure, but give me some guidelines here, maybe I can help.

How do you plan to implement it? I mean...

Are you planning to add a new option `type` at the @Column() decorator?

Something like this:
```ts
@Entity("users")
export class User {
     // ....

     @Column({  type: "stringfixed", length: 64})
     hash: string;
}
```
Maybe, you are planning to add a new option flag like "fixed"?

```ts
@Entity("users")
export class User {
     // ....

     @Column({  type: "string", length: 64, fixed: true})
     hash: string;
}
```

Personally, I think the second option would make more sense since it could be used for other types of data in the future. For example, in MySQL:
```
char / varchar
binary / varbinary
```

Or maybe in a totally different way...
Let me know what are you plans on this and I can push a PR to help you on that.

[]'s

pleerock:
Having `fixed` looks a better option then adding separate types for each of them. Its a bit more complicated to implement (but still easy) and it looks like a better approach. Should it be named `fixed` or `fixedLength`? Adding @19majkel94

19majkel94:
I think that `length` should set the fixed length n chars and `maxLength` should do the varchar like.
You can get typesafe in the same way like [here](https://github.com/typeorm/typeorm/issues/289#issuecomment-282101194)
If you don't want to introduce a breaking change, the best substitute would be `fixed` option I think.

navossoc:
I think `fixedLength` sounds better to me than just `fixed`.

[...]

Issue #344, opened 2017-03-14
@TypeORM splitting drivers into separate packages 


pleerock:
I was thinking a lot how to solve all problems with new differing drivers like mongodb one. 
Solution is to separate packages:

* `@typeorm/mysql`
* `@typeorm/mariadb`
* `@typeorm/sqlite`
* `@typeorm/sql-server`
* `@typeorm/oracle`
* `@typeorm/websql`
* `@typeorm/mongodb`

and possible future drivers like:

* `@typeorm/redis`
* `@typeorm/elasticsearch`
* `@typeorm/cassandra`
* `@typeorm/rethinkdb`
* `@typeorm/neo4j`

etc.

This requires lot of refactoring efforts and increases orm complexity, but that seems the only way to go further with different driver types. 

We may also have general packages like `@typeorm/core`, `@typeorm/sql`,  `@typeorm/nosql` however I'm not sure about them. Alternatively we can simply pack everything needed into their packages using different build scripts and tsconfigs.

Also another solution is to have `@typeorm/typeorm` and everything go from there, other packages just use declaration merging ability of TypeScript and add additional functionality, and users just use single  `@typeorm/typeorm` namespace everywhere.

Need help - suggestions, proposals or coding help if possible.

pleerock:
@19majkel94 maybe its time to appear in list of typeorm contributors 😉

madCoder24:
As a personal preference, I would prefer different packages for each driver. It would help me as a developer keep things clear as to what driver is being used in the project. A single namespace would be nice... but I can see declaration merging becoming increasing complicated as time went on and more drivers were added. 

Thoughts?

I have been following this project for about a month and find it very interesting. Great work so far!

Salimlou:
Hi @pleerock 

you can use something a monorepo like lerna, I'm using it for my projects.
But it comes with restrictions:

-Publishing  you cannot use an output folder for publishing, it publishes the project folder. you can overcome this by adding a .npmignore file to ignore unwanted files in the tarball

-Linking : on bootstrap lerna links dependents packages, and in case of peerDependenies, it May cause some problems.

perhaps we can start by refactoring so we have clear packages first, and then add a monorepo tool.

pleerock:
thanks guys for the feedback!

@madCoder24  as for me I see declaration merging is the best approach, it will help to easy migrations between different drivers, it will be more consistent, and documentation maybe easier to write, and overall users experience will be better I think. However its less flexible and requires good design of single namespace.

[...]

PR #369, opened 2017-03-25
Support fixed length column data type.

navossoc:
Well, I had some time and as I had mentioned in #343. This pull request tries to implement that feature.
Please take a look and if needed feel free to make any changes in it.

**Some notes:**
I didn't add support to SQLite and Web SQL because I want to discuss some another minor changes.

For example:
`CHARACTER`, `VARCHAR`, `VARYING CHARACTER`, `NCHAR`, `NVARCHAR` are all the same as `TEXT`.
Refer to [3.1.1. Affinity Name Examples](https://www.sqlite.org/datatype3.html#affinity_name_examples)

For this two specific drivers, there is no way to add a `fixedLength` attribute.

This brings some improvements, like this piece of code:
```ts
        switch (typeOptions.type) {
            case "string":
                return "character varying(" + (typeOptions.length ? typeOptions.length : 255) + ")";
            case "text":
                return "text";
        }
```

Can be simplified to this:
```ts
        switch (typeOptions.type) {
            case "string":
            case "text":
                return "text";
        }
```

It's just a line, but it ends up being one less type to worry about.

Also are some inconsistencies on the `normalizeType` signature.
https://github.com/typeorm/typeorm/blob/master/src/query-runner/QueryRunner.ts#L82

Most drivers don't have a return type, except for Postgres (string).
For example:
**MySQL:** https://github.com/typeorm/typeorm/blob/master/src/driver/mysql/MysqlQueryRunner.ts#L688
**Postgres**: https://github.com/typeorm/typeorm/blob/master/src/driver/postgres/PostgresQueryRunner.ts#L747

**--- EDIT ---**
Also, is there any special reason for SQL Server uses `NVARCHAR` and Oracle uses `VARCHAR2` type?
I mean, Oracle also have a type with unicode support (`NVARCHAR2`).
https://docs.oracle.com/cd/B28359_01/server.111/b28318/datatype.htm#CNCPT1821

I didn't make this changes on this PR, if you agree with that we can do another one later.
Let me know what you think about it.

[]'s

### DIFF
--- a/src/decorator/options/ColumnOptions.ts
+++ b/src/decorator/options/ColumnOptions.ts
@@ -80,4 +80,10 @@ export interface ColumnOptions {
      */
     readonly localTimezone?: boolean;
 
+    /**
+     * Indicates if column's type will be set as a fixed-length data type.
+     * Works only with "string" columns.
+     */
+    readonly fixedLength?: boolean;
+
 }
--- a/src/driver/mongodb/MongoQueryRunner.ts
+++ b/src/driver/mongodb/MongoQueryRunner.ts
@@ -643,7 +643,7 @@ export class MongoQueryRunner implements QueryRunner {
     /**
      * Creates a database type from a given column metadata.
      */
-    normalizeType(typeOptions: { type: ColumnType, length?: string|number, precision?: number, scale?: number, timezone?: boolean, fixedLength?: boolean }) {
+    normalizeType(typeOptions: { type: ColumnType, length?: string|number, precision?: number, scale?: number, timezone?: boolean, fixedLength?: boolean }): string {
         throw new Error(`Schema update queries are not supported by MongoDB driver.`);
     }
 
--- a/src/driver/mongodb/MongoQueryRunner.ts
+++ b/src/driver/mongodb/MongoQueryRunner.ts
@@ -643,7 +643,7 @@ export class MongoQueryRunner implements QueryRunner {
     /**
      * Creates a database type from a given column metadata.
      */
-    normalizeType(typeOptions: { type: ColumnType, length?: string|number, precision?: number, scale?: number, timezone?: boolean }) {
+    normalizeType(typeOptions: { type: ColumnType, length?: string|number, precision?: number, scale?: number, timezone?: boolean, fixedLength?: boolean }) {
         throw new Error(`Schema update queries are not supported by MongoDB driver.`);
     }
 
--- a/src/driver/mysql/MysqlQueryRunner.ts
+++ b/src/driver/mysql/MysqlQueryRunner.ts
@@ -685,11 +685,15 @@ export class MysqlQueryRunner implements QueryRunner {
     /**
      * Creates a database type from a given column metadata.
      */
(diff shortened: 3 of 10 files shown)


pleerock:
Hi. Thank you for this PR. I does look good.

> Also are some inconsistencies on the normalizeType signature.
> Most drivers don't have a return type, except for Postgres (string).

right, they do should have by a convention, even if they are inferred. 

> Also, is there any special reason for SQL Server uses NVARCHAR and Oracle uses VARCHAR2 type?
> I mean, Oracle also have a type with unicode support (NVARCHAR2).

I'm not sure. Shall we use nvarchar everywhere or simply varchar? Or shall it be choosable by a user?

> CHARACTER, VARCHAR, VARYING CHARACTER, NCHAR, NVARCHAR are all the same as TEXT.
Refer to 3.1.1. Affinity Name Examples

Thats new for me. Didn't know about it. However we can left it as it is because of below:

There is types redesign in a next major typeorm version. See #344 for more details. Also take a look on https://github.com/pleerock/typeorm-experiment how types are implemented new-way for postgres and other drivers (right now only postgres is ready, however it shows principles for other drivers)

navossoc:
@pleerock Sure...

> right, they do should have by a convention, even if they are inferred.

Ok, but what signature is the right one? 😆 
**any:** https://github.com/typeorm/typeorm/blob/master/src/query-runner/QueryRunner.ts#L82
**string**: https://github.com/typeorm/typeorm/blob/master/src/driver/postgres/PostgresQueryRunner.ts#L747

For me, makes more sense the `string` type, but I don't have a deep knowledge on the project, maybe I'm missing something and be wrong.

If you let me know, I can push one more commit into this PR fixing this also.

> I'm not sure. Shall we use nvarchar everywhere or simply varchar? Or shall it be choosable by a user?

AFAIK, JavaScript and Node are designed to be Unicode compliant.
So I think it make more sense defaults to UTF.

I don't know the inner details, so it is UTF-8, UTF-16, UTF-32? Little Endian, Big Endian?
How Oracle implements it too? It will be problems? No idea at all!

I just saw the difference and raised the flag, sorry, but I can't help much on that question.
I think it needs a proper research and people with experience with Oracle to say what works or not.

> Thats new for me. Didn't know about it. However we can left it as it is because of below:

Of course, then we'll leave it as it is for now.

pleerock:
> Ok, but what signature is the right one?

They always return string, isn't it? Then it should be string, not `any`. If you want/can you can fix this too.

>  we'll leave it as it is for now.

Lets do so for now.

navossoc:
Yes, at least for now, they all return a string.
Pushed another commit fixing that.

If I find more cases like that, do you want I create an issue to report?

[]'s

pleerock:
> If I find more cases like that, do you want I create an issue to report?

you can send a fix its something small, or create an issue if its something big and needs discussion.

Also take a look on  https://github.com/pleerock/typeorm-experiment and #344  if you want to contribute on a next minor version of typeorm.

Thank you, Ive merged your PR.

PR #380, opened 2017-03-31
bugfix query condition

mingyang91:
I missed some special cases in the code previously submitted(#160).

This causes the second schema to be initialized in the same database, and an exception is thrown when the `DROP CONSTRAINT` is run.
Because `indicesSql`, `foreignKeysSql`, `uniqueKeysSql`, `primaryKeysSql` do not add constraints (`table_schema =?`), The query retrieves all the constraints in the entire database and deletes it in a particular schema, but it does not exist.

I have fixed the problem and added the test case.


mingyang91:
The CI check has passed, except for Node.js: stable.
https://travis-ci.org/typeorm/typeorm/jobs/217457743

`Cannot create a new connection named "sqlite", because connection with such name already exist and it now has an active connection session.`
The above error seems to be caused by other reasons.

test/functional/connection/connection.ts
```diff
@@ -384,30 +387,39 @@ describe("Connection", () => {
...
                 await connection.syncSchema(true);
 
-                const post = new Post();
-                post.title = "ChangeSchemaName";
+                const comment = new Comment();
+                comment.context = "ChangeSchemaName";
 
-                const PostRepo = connection.getRepository(Post);
-                await PostRepo.persist(post);
+                const CommentRepo = connection.getRepository(Comment);
```
pleerock: please follow a common convention and naming it something like `repository` or `commentRepository`

pleerock:
> Cannot create a new connection named "sqlite", because connection with such name already exist and it now has an active connection session.

yes there is unknown issue with sqlite for some reason sometimes it fails. I restarted the build and its okay now

pleerock:
Before merging this PR I would like to ask you to properly test schema creation, because you touched it and its not test covered (I mean overall test coverage of schema creation in orm) we need to make sure that this change did not break anything else. Thank you.

test/functional/connection/connection.ts
mingyang91: I need to create some `fk constraints` through `ManyToOne`/`OneToMany` to test whether they are isolated.
If I created on the `Post` entity, I would have to modify all the use cases (`entities: [Post, /* ??? */]`) to include the `fk` relationship.
So I created two new entities.

mingyang91:
👌，overall

ormconfig.json.dist
```diff
@@ -56,6 +56,22 @@
...
+      "port": 5432,
+      "username": "test",
+      "password": "test",
+      "database": "test",
+      "schemaName": "another-schema"
+    },
+    "logging": {
+      "logQueries": false
+    }
+  },
```
pleerock: Do you think we need these changes?

pleerock: because I would like to avoid new connections

ormconfig.travis.json
```diff
@@ -44,6 +44,19 @@
...
+    "name": "postgres-schema",
+    "driver": {
+      "type": "postgres",
+      "host": "localhost",
+      "port": 5432,
+      "username": "test",
+      "password": "test",
+      "database": "test",
+      "schemaName": "another-schema"
+    }
```
pleerock: but same left here

mingyang91:
Do you mean that? 389c250bf1ac85638afb599e9e80fc88c4a38878

test/functional/connection/connection.ts
```diff
@@ -1,10 +1,16 @@
 import "reflect-metadata";
 import {expect} from "chai";
 import {Post} from "./entity/Post";
+import {Guest as GuestV1} from "./entity/v1/Guest";
+import {Comment as CommentV1} from "./entity/v1/Comment";
+import {Guest as GuestV2} from "./entity/v1/Guest";
+import {Comment as CommentV2} from "./entity/v1/Comment";
```
pleerock: I did not understand why are you doing this? GuestV1 and GuestV2 are same just like CommentV1  and CommentV2. What is the point to do it?

test/functional/connection/connection.ts
```diff
@@ -381,33 +387,89 @@ describe("Connection", () => {
...
+                    schemaName: "test-schema",
+                    dropSchemaOnConnection: true,
+                }),
+                createTestingConnections({ 
+                    enabledDrivers: ["postgres"],
+                    entities: [CommentV1, GuestV1],
+                    schemaName: "another-schema",
+                    dropSchemaOnConnection: true
+                })
+            ]);
```
pleerock: looks quite complex, maybe better to write them separately without Promise.all?:

```ts
const connection1 = await createTestingConnections({ 
                    enabledDrivers: ["postgres"],
                    entities: [CommentV1, GuestV1],
                    schemaName: "first-schema",
                    dropSchemaOnConnection: true
                });

const connection2 = await createTestingConnections({ 
                    enabledDrivers: ["postgres"],
                    entities: [CommentV1, GuestV1],
                    schemaName: "second-schema",
                    dropSchemaOnConnection: true
                });

```

test/functional/connection/connection.ts
```diff
@@ -381,33 +387,89 @@ describe("Connection", () => {
...
+                    schemaName: "another-schema",
+                    dropSchemaOnConnection: false,
+                    schemaCreate: true
+                })
+            ]);
+            connections = [connection1, connection2];
+        });
+    });
+
+    describe.only("Can change postgres default schema name", () => {
```
pleerock: don't forget to remove `. only`

test/functional/connection/connection.ts
```diff
@@ -381,33 +387,89 @@ describe("Connection", () => {
...
+                const comment = new CommentV1();
+                comment.title = "Change SchemaName";
+                comment.context = `To ${schemaName}`;
 
-                const post = new Post();
-                post.title = "ChangeSchemaName";
-
-                const PostRepo = connection.getRepository(Post);
-                await PostRepo.persist(post);
+                const CommentRepo = connection.getRepository(CommentV1);
```
pleerock: please name repositories as standard - camelCase and simple like: `commentRepository` or simply `repository`

test/functional/connection/connection.ts
mingyang91: In the previous test case, the result of loadTableSchemas is always empty. Because each time the establishment of connection will drop schema.
Only in this way can we cover the case of sync schema.

pleerock: are you sure? Can you check once again? Because what you did is same as:

```ts
const a = {};
const b = a;
```

You have same instances of GuestV1 and GuestV2, CommentV1 and CommentV2

pleerock: it should be: GuestV1 === GuestV2 and CommentV1 === CommentV2

mingyang91: typo 😂

mingyang91:
ok

pleerock:
thank you ;)

PR #386, opened 2017-04-06
Correct table name to 'photometadata' in one-to-one relation example

rodrigopandini:
Correct table name to 'photometadata' in one-to-one relation example in README.md file


pleerock:
thank you very much!

PR #431, opened 2017-04-29
Let know to relation wrapper that entity was loaded using query build…

serega3000:
Without this line related entity will be loaded again when request it from current entity


serega3000:
Wow test failed. But i think that change PR is correct whatever. Its or wrong tests or another bug. I will try to understand

Issue #234, opened 2017-01-25
lazy loading, many-to-one, querying on the wrong entity id.


ericwooley:
***I updated the details below after further investigation***

I am having an issue where I am lazy loading a relationship, but a where an [the wrong entity id is being queried](https://github.com/ericwooley/graphql-typeorm-passport-boilerplate/blob/master/src/graphql/queries/user.ts#L19)

`executing query: SELECT "lists"."id" AS "lists_id", "lists"."name" AS "lists_name", "lists"."owner" AS "lists_owner" FROM "shopping_list" "lists" INNER JOIN "user" "User" ON "User"."id"="lists"."owner" WHERE "lists"."id"=$1 -- PARAMETERS: [1]`

Should just be 
`executing query: SELECT "lists"."id" AS "lists_id", "lists"."name" AS "lists_name", "lists"."owner" AS "lists_owner" FROM "shopping_list" "lists" INNER JOIN "user" "User" ON "User"."id"="lists"."owner" WHERE "user"."id"=$1 -- PARAMETERS: [1]`

If I don't lazy load and use add the join to the repo.findById (everything works fine)[https://github.com/ericwooley/graphql-typeorm-passport-boilerplate/blob/master/src/graphql/queries/user.ts#L6]

```
const user = await userRepo.findOneById(request.user.id, {
      alias: 'user',
      innerJoinAndSelect: {
        "lists": "user.lists"
    }
})
```

I have a [user](https://github.com/ericwooley/graphql-typeorm-passport-boilerplate/blob/master/src/models/user.ts#L26) class, with a lazy loaded one to many relationship: 
```
@Table()
export default class User {
	// ...
	@OneToMany(type => ShoppingList, shoppingList => shoppingList.owner, {
		cascadeAll: true
	})
	lists: Promise<ShoppingList[]>
        // ... 
}
```

and a class for [ShoppingList](https://github.com/ericwooley/graphql-typeorm-passport-boilerplate/blob/master/src/models/shoppingList.ts)
```
@Table()
       // ...
	@ManyToOne(type => User, user => user.lists, {
		cascadeAll: true,
	})
	owner: Promise<User>
}
```

I cannot figure out for the life of me, why this where clause is being added. Maybe it's something dumb I am doing, but I have tried messing with it every way I can think of.

pleerock:
when you can `user.list` you want to load all lists of this SPECIFIC user. If your query will be:

[...]

Issue #223, opened 2017-01-21
@ManyToMany return all rows in table


Fokir:
I have model
```
@Table()
export class User extends Model{
    @PrimaryGeneratedColumn()
    id: number;
    ...
    @ManyToMany(type=>Role, (role:Role)=>role.users)
    roles: Promise<Array<Role>>;
    ...
}
```
and 
```
@Table()
export class Role extends Model{
    static __repository: Repository<Role>;

    @PrimaryGeneratedColumn()
    id: number;

    @Column()
    @Index({unique: true})
    role: string;

    @ManyToMany(type => User, (user: User) => user.roles)
    @JoinTable()
    users: Promise<Array<User>>;
}
```

![table](https://cloud.githubusercontent.com/assets/1648384/22175236/87a4b53c-e00a-11e6-84fb-45fd2c50995e.png)

```
let user: User = await token.user;
user.roles.then((d)=>{
    console.log(d);
});
```
Returns all roles from the role of the table, ignoring the staging table?
 I'm use 0.0.7-alpha.22
 
![unknown 1](https://cloud.githubusercontent.com/assets/1648384/22175285/6eef78dc-e00b-11e6-8859-9d81eff04999.png)

Fokir:
Can I wait for error correction in coming days?

ItAlexander:
+

SetgorS:
+

pleerock:
haha you have a strong team 😆 . okay I'll make a fix for this this week.

Fokir:
Thank you, it's just an important place in the application. A "typeorm" has great potential in applications Typescript

Fokir:
The problem is fixed?

pleerock:
Yes. released in `0.0.8-alpha.2`

PR #432, opened 2017-04-29
Let know to lazy relation wrapper that entity was loaded using query builder

serega3000:
Maked for lazy relations related entity to be not loaded again after selecting with join and fetching it from current entity. Added `__has__*=true` to RawSqlResultsToEntityTransformer.
But then some tests make failed, and i founded that if we have inverse side relation, ex. category.posts <-> post.category, when we fetching post from category (category.posts), categories (post.categories) will be also added in result set with empty array, because it persist in join method in LazyRelationWrapper, and it will be added as alias to query builder and then to RawSqlResultsToEntityTransformer. The only solution i found is to remove `__has__*` value from result object in LazyRelationWrapper.

### DIFF
--- a/src/lazy-loading/LazyRelationsWrapper.ts
+++ b/src/lazy-loading/LazyRelationsWrapper.ts
@@ -56,6 +56,12 @@ export class LazyRelationsWrapper {
                     }
 
                     this[promiseIndex] = qb.getMany().then(results => {
+                        // remove information about current entity because we joined it
+                        if (relation.hasInverseSide) {
+                            results.map(result => {
+                                delete (result as any)["__has__" + relation.inverseRelation.propertyName + "__"];
+                            });
+                        }
                         this[index] = results;
                         this[resolveIndex] = true;
                         delete this[promiseIndex];
@@ -73,6 +79,10 @@ export class LazyRelationsWrapper {
                         .where(relation.entityMetadata.targetName + "." + relation.inverseEntityMetadata.firstPrimaryColumn.propertyName + "=:id", { id: relation.entityMetadata.getEntityIdMixedMap(this) });
 
                     this[promiseIndex] = qb.getMany().then(results => {
+                        // remove information about current entity because we joined it
+                        results.map(result => {
+                            delete (result as any)["__has__" + relation.inverseRelation.propertyName + "__"];
+                        });
                         this[index] = results;
                         this[resolveIndex] = true;
                         delete this[promiseIndex];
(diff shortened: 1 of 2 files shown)


pleerock:
Hi and thank you for your PR!

Please make sure to add tests for the functionality you add.

Sorry but I didn't understood the reason why you added `remove __has__*` code into `LazyRelationWrapper`. Maybe you can comment that code and add some tests which shows the problem?

serega3000:
Ok, i will add tests and comments.
I try to explain why i added `remove __has__*` into `LazyRelationWrapper`, may be exists better solution, but i did not founded it.

When i just added `entity["__has__" + propertyName + "__"] = true;` tests failed on `github issues > #234 and #223 lazy loading does not work correctly from one-to-many and many-to-many sides`

I made some debugging and understand problem:
For example, we have lazy relations and inverse side relation, `category.posts` <-> `post.category`.
We have variable `post`
we try to get category from post `let category = await post.category`
and then we want to get all posts in this category `let posts = await  category.posts`.
We always will get empty array.

It is because when `LazyRelationWrapper` creating query for finding category, he also joining post. https://github.com/typeorm/typeorm/blob/master/src/lazy-loading/LazyRelationsWrapper.ts#L91
When post table joined, it added to `AliasMap` in `QueryBuilder`
Then `QueryBuilder` send `AliasMap` to `RawSqlResultsToEntityTransformer`
Next, when `RawSqlResultsToEntityTransformer` build result, line 163:
https://github.com/typeorm/typeorm/blob/master/src/query-builder/transformer/RawSqlResultsToEntityTransformer.ts#L163

` const relationAlias = this.aliasMap.findAliasByParent(alias.name, relation.propertyName);`
relation alias of 'post' founded.


`const relatedEntities = this.groupAndTransform(rawSqlResults, relationAlias);`
returns empty array, although no one column of post selected in query


```
const isResultArray = relation.isManyToMany || relation.isOneToMany;
const result = !isResultArray ? relatedEntities[0] : relatedEntities;
```
result still empty array


```
                if (result) {
                    ....
                    if (relation.isLazy) {
                        entity["__" + propertyName + "__"] = result;
                        entity["__has__" + propertyName + "__"] = true;
                    } else {
                   .....
```
Result of empty posts will be added to category, and when we try to get posts from this category, we always get empty array.

pleerock:
@serega3000 Im sorry for late response here, but are you still interested in this fix?

pleerock:
If you are please check latest version if your issue is already fixed. If its not then please create a test that reproduce your issue and I'll make a fix by myself.

Issue #334, opened 2017-03-07
QueryBuilder Refactoring Plan


pleerock:
Plan:

- [x] partial selection
- [ ] fix problem with long aliases e.g getMaxIdentifierLength
- [x] refactor `getManyAndCount()` method - uses two query runners
- [ ] fix all issues with relation ids functionality
- [ ] fix issues with escaping
- [ ] fix relation count functionality
- [ ] test partial selection with embeddables
- [ ] implement `addSelectAndMap` functionality #296
- [ ] refactor QueryBuilder code
- [x] implement first level cache
- [ ] `.addCount` and `.addCountSelect()` and similar functionality
- [ ] add exceptions everywhere possible
- [ ] test coverage everything is possible
- [x] implement streaming
- [x] fix issues #320 #293 #275 #261 #255 #241 #200 #199 #89 #347

pleerock:
Most of query builder refactoring has been already finished. There will be other rounds of refactorings in the future, for now I think we can close this issue.

PR #445, opened 2017-05-02
chore: move @types/mongodb to devDependencies

NoNameProvided:


pleerock:
I would like to. But if you package typeorm without these typings users gonna have bunch of compiler errors because of "mongodb" usages in typeorm codebase. Fix can be done after #344

pleerock:
I'll merge this one for now. #334 fixes this some specific way...

Issue #452, opened 2017-05-03
Complete Oracle tests


gintsgints:
I would like to create this thread for issues on Oracle test execution.

pleerock:
we are lucky if we can setup docker container with oracle inside and run it on CI. btw are you using oracle?

gintsgints:
Some questions I have:
- Why normal tests includes sample2-one-to-one.ts (line 24) file with only MySQL driver? This means you cannot skip MySQL tests?
- Probably there will be lot of issues about ORA-00972: identifier is too long. Need to find out to change name generator for it

gintsgints:
"setup docker container with oracle inside and run it on CI" is big aim :)

Oracle answered that it is ok to use they image at store. But this requests signing up to docker store. Not sure how to do this with travis.

https://github.com/oracle/docker-images/issues/367#issuecomment-298252450

pleerock:
1. `sample*` tests were first tests I developed and at that time only mysql were supported. We need to refactor those tests to make them to work with all drivers.

2. `ORA-00972: identifier is too long.` What is this issue about? Identifier of what? table? column? alias,?

I remember I setup oracle in docker on my mac machine. But it was a long time ago. I think we can setup docker for oracle too (but images are huuuuge).

gintsgints:
This time it was column alias.

pleerock:
this happens sometimes on a really very very long generated names, probably only on many-to-many tables. You can setup custom column names for many to many tables.

gintsgints:
I already using custom generator for project I use typeorm. Option could be setup custom generator for tests too.

gintsgints:
Dealt with column names and Join table names so far. Now comes index:

```
query failed: CREATE  INDEX "ind_e25153a614c8d27da45a08a506f" ON "ps_cats"("sample4PostCategoryId")
      1) "before all" hook: reloadDatabase
```

gives

```
Error: ORA-00972: identifier is too long
```

gintsgints:
@pleerock Please check my pull requests.

pleerock:
@gintsgints done. Can you please setup docker container for travis so we can run oracle tests?

AlexMesser:
fixed and released in `typeorm@0.2.0-alpha.23`. You can try it via `npm i typeorm@next`.

Kononnable:
I'm not sure what was fixed here. According to ormconfig.travis.json Travis is skipping testing on oracle.

[...]

PR #454, opened 2017-05-03
#452 First changes for Oracle tests to pass.

gintsgints:
For you to review.

### DIFF
--- a/sample/sample2-one-to-one/entity/PostDetails.ts
+++ b/sample/sample2-one-to-one/entity/PostDetails.ts
@@ -7,10 +7,10 @@ export class PostDetails {
     @PrimaryGeneratedColumn()
     id: number;
 
-    @Column()
+    @Column({name: "auth_nam"})
     authorName: string;
 
-    @Column()
+    @Column({name: "cmnt"})
     comment: string;
 
     @Column()
--- a/sample/sample2-one-to-one/entity/PostMetadata.ts
+++ b/sample/sample2-one-to-one/entity/PostMetadata.ts
@@ -7,7 +7,7 @@ export class PostMetadata {
     @PrimaryGeneratedColumn()
     id: number;
 
-    @Column()
+    @Column({name: "desc"})
     description: string;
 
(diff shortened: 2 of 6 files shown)


gintsgints:
Guilty was this one:

```
"sample3_post_metadata"."description" AS "sample3_post_metadata_description" 
```

Of coarse you can shorten also table name. Wish me go that way?

pleerock:
shortening column names isnt a fix there. Do you expect users would be pleasured to do so too? We need some good fix. I don't understand why oracle support on aliases length is so limited.

gintsgints:
May be we can generate aliases somehow shorter?

pleerock:
ahhh Im not sure... Do you have any ideas?

gintsgints:
Hmm just a question. Why you need to give them aliases anyway?

pleerock:
ahhh thats a long story.... we need aliases to map data correctly to entities. You need to learn code of `QueryBuilder` and `RawSqlResultsToEntityTransformer` to understand why they are needed.

pleerock:
Probably we have only one way of fixing this - shorten them. But im not sure what is effective way of doing this. Maybe cut half of table name and column name in query builder? Not really sure. Is there any way we can increase alias length in oracle via orm?

Also, if you are planning to work on QB I recommend to switch [to this](https://github.com/typeorm/typeorm/tree/query-builder-refactoring) branch, because there are latest changes of everything

gintsgints:
How about giving them numbers? Like Alias1, Alias2, .... AliasN

pleerock:
we can try numbers, however it will make things much much more complicated. And its already quite complicated

pleerock:
@gintsgints we are planning to finish complete oracle support in `0.2.0`. @AlexMesser already fixed most of issues oracle has on [his own branch](https://github.com/typeorm/typeorm/tree/schema-builder-refactoring). Soon all those changes will be landed in `next` and final 0.2.0 will be released next month. Thank you very much for your contribution.

PR #459, opened 2017-05-06
Support Oracle for indices.

gintsgints:


src/naming-strategy/DefaultNamingStrategy.ts
```diff
@@ -31,7 +31,7 @@ export class DefaultNamingStrategy implements NamingStrategyInterface {
             return customName;
 
         const key = "ind_" + tableName + "_" + columns.join("_");
-        return "ind_" + RandomGenerator.sha1(key).substr(0, 27);
+        return "in_" + RandomGenerator.sha1(key).substr(0, 27);
```
pleerock: why "in_"? If its about length then batter to change 27 to 26, because I would prefer "ind_" to stay same in all databases

gintsgints: Oki... I will change it to 26 then.

Issue #453, opened 2017-05-03
Connect using Active Directory account


StickNitro:
Hi, this is a question really, is it possible with typeorm to connect to an MSSQL database using a Windows Active Directory account? In other words is it possible to specify the DOMAIN with the connection either separately or with the username?

for example:
```
{
  username: "test.user",
  password: "...",
  domain: "mydomain"
}
```
or
```
{
  username: "mydomain/test.user",
  password: "..."
}
```

I am looking at options for using an ORM in an upcoming NodeJS service but one of the caveats is that we must use a Windows Domain account to connect to the database (we cannot use an sql account)

pleerock:
Im not familiar with sql-server specific options, so I can't answer your question. Orm uses [this underlying driver](https://github.com/patriksimek/node-mssql) which is responsible for the whole connection. Probably you can ask them if they are supporting what you are talking about?

StickNitro:
Hi,

It is supported on the referenced driver, you can include a ```"domain": "mydomain"``` property when connecting to mssql.

From looking at the driver code in the repository it looks like this is not included in the ```ConnectionOptions``` or ```DriverOptions``` types and is therefore not catered for in the ```SqlServerDriver```.

I can take a look at this and see if I can get a PR together to include this option.

pleerock:
you can do this:

```json
[
  {
    "name": "default",
    "driver": {
      "type": "mysql",
      "host": "localhost",
      "port": 3306,
      "username": "xxxx",
      "password": "xxxx",
      "database": "xxxx",
      "extra": {
           "domain": "mydomain"
      }
    },
    "autoSchemaSync": true
  }
]
```

PR #469, opened 2017-05-11
Connect using AD

StickNitro:
Pull request for #453 - primarily for the SqlServerDriver

added domain property to DriverOptions
updated DriverUtils to parse domain from url (i.e. domain/username:password for domain based logins
updated SqlServerDriver to include domain when connecting

### DIFF
--- a/src/driver/DriverOptions.ts
+++ b/src/driver/DriverOptions.ts
@@ -80,7 +80,10 @@ export interface DriverOptions {
     readonly tablesPrefix?: string;
 
     /**
-     * Specify the domain for active directory connections
+     * Specify the domain for active directory connections.
+     *
+     * NOTE: This options is MSSQL specific.
+     *
      */
     readonly domain?: string;
 }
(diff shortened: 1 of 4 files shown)


src/driver/DriverOptions.ts
```diff
@@ -79,4 +79,8 @@ export interface DriverOptions {
      */
     readonly tablesPrefix?: string;
 
+    /**
+     * Specify the domain for active directory connections
```
pleerock: Please add dot at the end of comment. Also please add note that this option is mssql-specific.

pleerock:
Thank you for the PR! Please check why tests are failing. Also please create one another PR against 0.0.x-version branch

StickNitro:
Closing PR as identified that you are able to pass ```"extra":{"domain": "mydomain"}``` which do actually get passed down to the driver

pleerock:
I think you can reopen this PR, because even `"extra":{"domain": "mydomain"}` is supported this will be handy if this option will be on the top level

pleerock:
I still insist you to recreate this PR. This would be very useful if we will have domain right in mssql driver options. Please do it if you have a free time.

PR #470, opened 2017-05-11
Added domain property for mssql driver to suport AD auth

StickNitro:
relates to #453 Connect using Active Directory account


StickNitro:
Closing PR as identified that you are able to pass ```"extra":{"domain": "mydomain"}``` which do actually get passed down to the driver

Issue #460, opened 2017-05-08
still not support int array in Postgres?


Rwing:
table

```
CREATE TABLE test
(
    a INTEGER[]
);
```

and entity.ts

```
export class test {
    @Column("simple_array")
    a: number[];
}
```

and insert a record
```
let t = new test();
t.a = [1,2,3];
repository.persist(t);
```
I got a error

> nodejs.error: malformed array literal: "1,2,10"
    at Connection.parseE (/Users/XXX/node_modules/pg/lib/connection.js:569:11)
    at Connection.parseMessage (/Users/XXX/node_modules/pg/lib/connection.js:396:17)
    at Socket.<anonymous> (/Users/XXX/node_modules/pg/lib/connection.js:132:22)
    at emitOne (events.js:96:13)
    at Socket.emit (events.js:189:7)
    at readableAddChunk (_stream_readable.js:176:18)
    at Socket.Readable.push (_stream_readable.js:134:10)
    at TCP.onread (net.js:551:20)
name: 'error'
length: 142
severity: 'ERROR'
code: '22P02'
detail: 'Array value must start with "{" or dimension information.'
hint: undefined
position: undefined
internalPosition: undefined
internalQuery: undefined
where: undefined
schema: undefined
table: undefined
column: undefined
dataType: undefined
constraint: undefined
file: 'arrayfuncs.c'
line: '319'
routine: 'array_in'

NoNameProvided:
Simply use `autoSchemaSync: true` and it will work. 

Simple array behind the scenes is a comma separated string, so `[1, 2, 3, 4]` will become `"1,2,3,4"`.

Rwing:
Sorry, I didnt catch your meaning, the field's type is INTEGER[] in database, so string "1,2,3,4" can't insert into database yet.

NoNameProvided:
When you write: 

```
@Column({ type: 'simple_array', default: '' })
categories: number[];
```

The created field by TypeORM is a `text` field.

The table you create by hand is incorrect, you should create:

```sql
CREATE TABLE test(a TEXT NOT NULL);
``

Rwing:
I got it, but for same reason, I can't change the database schema so the field must be INTEGER[] 😢

Rwing:
I solved it by a temp solution :
use return value instant of at this line
https://github.com/typeorm/typeorm/blob/master/src/driver/postgres/PostgresDriver.ts#L235

pleerock:
[...]

PR #489, opened 2017-05-23
fixed the issue #460

Rwing:


pleerock:
I have to close this PR because its not properly implemented. "simple array" type and postgres arrays are different. Support for postgres array is implemented in latest typeorm versions. See [here](https://github.com/typeorm/typeorm/blob/master/test/functional/database-schema/column-types/postgres/entity/Post.ts) example how to use them

PR #496, opened 2017-05-26
fixed getCount for oracle alias case sensitivy

hajekj14:
Fixed getCount with Oracle case sensitivity for aliases

### DIFF
--- a/src/query-builder/QueryBuilder.ts
+++ b/src/query-builder/QueryBuilder.ts
@@ -1145,7 +1145,7 @@ export class QueryBuilder<Entity> {
 
         try {
             const results = await queryRunner.query(countQuerySql, countQueryParameters);
-            if (!results || !results[0] || !results[0]["cnt"] || !results[0]["CNT"])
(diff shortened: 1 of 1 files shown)


hajekj14:
Is this solution ok for you? Or you prefered add quotas to getCount query(... as "cnt" ...) or complet differet way?

Thanks

pleerock:
probably better to do `add quotas to getCount query(... as "cnt" ...)`. Are you able to do it?

hajekj14:
Ok, no problem.

hajekj14:
problem with travisCI: Cannot create a new connection named "sqlite", because connection with such name already exist and it now has an active connection session

pleerock:
yeah simply restart the build...

pleerock:
thanks for this awesome pr ;)

hajekj14:
:-) no problem, it was be really complicated

Dne 2. 6. 2017 20:56 napsal uživatel "Umed Khudoiberdiev" <
notifications@github.com>:

PR #507, opened 2017-05-31
Fix oracle offset limit

hajekj14:
It's not best solution as you mentioned in QueryBuilder for SqlServerDriver, but from my point is it OK before you finished QueryBuilder refactoring.

btw.: If you want some help with Oracle stuff, we have strong background with this DB


pleerock:
Why did you close it?

hajekj14:
It was broken, I worked on better solution.

JH.

Dne 3. 6. 2017 10:42 napsal uživatel "Umed Khudoiberdiev" <
notifications@github.com>:

PR #516, opened 2017-06-04
fix oracle offset limit

hajekj14:
It's a quick fix, only for setLimit and setOffset, methods skip and take has a problem with order by "ids_s". Problem with Oracle 11 and less is they have not any built in query limitation. I thinking about move query building to Driver specific class, what you think about it? It will be better for handling db diffs.


hajekj14:
How I can restart travis job?

src/query-builder/QueryBuilder.ts
```diff
@@ -1639,13 +1646,26 @@ export class QueryBuilder<Entity> {
         return "";
     }
 
+    createSpecificExpression(sql: string): string {
```
pleerock: why did you make this method public? Please change it to protected.

pleerock: also please name it better, its not understandable what this specific is. Something like `createLimitOffsetSpecificExpression`

src/query-builder/QueryBuilder.ts
```diff
@@ -1639,13 +1646,26 @@ export class QueryBuilder<Entity> {
...
+         if ((this.expressionMap.offset || this.expressionMap.limit) && this.connection.driver instanceof OracleDriver) {
+             sql = "SELECT * FROM (" + sql + ") WHERE ";
+             if (this.expressionMap.offset) {
+                 sql += "\"RN\" > " + this.expressionMap.offset;
```
pleerock: use exist `this.escapeAlias` method instead of manually doing `\"RN\"`

src/query-builder/QueryBuilder.ts
```diff
@@ -1397,6 +1401,9 @@ export class QueryBuilder<Entity> {
         switch (this.expressionMap.queryType) {
             case "select":
                 const selection = allSelects.map(select => select.selection + (select.aliasName ? " AS " + ea(select.aliasName) : "")).join(", ");
+                if ((this.expressionMap.limit || this.expressionMap.offset) && this.connection.driver instanceof OracleDriver) {
+                    return "SELECT ROWNUM \"RN\"," + selection + " FROM " + this.escapeTable(tableName) + " " + ea(aliasName) + lock;
```
pleerock: use exist `this.escapeAlias` method instead of manually doing `\"RN\"`

pleerock:
You can't restart job - only I can. Don't worry if one of the platform fails - worry only if all of them are failing.

pleerock:
> I thinking about move query building to Driver specific class, what you think about it?

yeah I was also thinking about it but the answer for now is "no". Its quite hard to design good interface here without making it too complex.

pleerock:
Also, are you able to setup docker container that can run oracle tests? Currently its a pain to run oracle tests we need to find mechanism to run them automatically, because otherwise oracle features may break periodically a lot.

hajekj14:
I'm fixed stuff from your comments. Answers for your other questions send later.

hajekj14:
@pleerock problem with skip and take is simple, Oracle has problem with aliases without quotas for example take & skip block in getEntitiesAndRawResults method create alias for primary key without qt and use it in ORDER BY with qt is problem, I will fix it by adding qt for primary key alias:
``DISTINCT(${distinctAlias}.${propertyName}) as ids_${primaryColumn.databaseName}`;` => ``DISTINCT(${distinctAlias}.${propertyName}) as "ids_${primaryColumn.databaseName}"`;`

pleerock:
what about this my request:

> Also, are you able to setup docker container that can run oracle tests? Currently its a pain to run oracle tests we need to find mechanism to run them automatically, because otherwise oracle features may break periodically a lot.

I really afraid this will break very soon if we won't setup testing infrastructure to run oracle tests

hajekj14:
I talk about it with our SA, we can create container with oracle linux & oracle 11 xe, but there is a lot of problems with typeorm and Oracle 11 like identifiers length, limits etc. We would like to run on Oracle 12 because this version resolve most of this stuff. I try to solve container with 11 and we will see.

pleerock:
Oracle is the most problematic database I have worked with

PR #523, opened 2017-06-07
fix #521 related to typo in QB

### LINKED ISSUE #521: Update values not getting replaced in 0.1.0 alpha4
Its about postgresql, I don't know if same applies for other drivers.
When doing .update using query builder, the values from provided object literal don't get replaced with arguments. So for example if I do .update({ something: true }) then in the query it ends up as something=:updateset__something instead of something=$1, also the "true" isnt present in arguments array.

I will try to investigate it further.

bartlomiej-korpus:
fix #521

### DIFF
--- a/src/query-builder/QueryBuilder.ts
+++ b/src/query-builder/QueryBuilder.ts
@@ -1405,7 +1405,7 @@ export class QueryBuilder<Entity> {
                 const updateSet = Object.keys(this.expressionMap.updateSet).map(key => key + "=:updateSet__" + key);
                 const params = Object.keys(this.expressionMap.updateSet).reduce((object, key) => {
                     // todo: map propertyNames to names ?
-                    object["updateSet_" + key] = this.expressionMap.updateSet![key];
+                    object["updateSet__" + key] = this.expressionMap.updateSet![key];
                     return object;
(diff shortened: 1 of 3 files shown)


src/query-builder/QueryBuilder.ts
```diff
@@ -1405,7 +1405,7 @@ export class QueryBuilder<Entity> {
                 const updateSet = Object.keys(this.expressionMap.updateSet).map(key => key + "=:updateSet__" + key);
                 const params = Object.keys(this.expressionMap.updateSet).reduce((object, key) => {
                     // todo: map propertyNames to names ?
-                    object["updateSet_" + key] = this.expressionMap.updateSet![key];
+                    object["updateSet__" + key] = this.expressionMap.updateSet![key];
```
dcworldwide: What does ! mean in the statement `this.expressionMap.updateSet![key];`

pleerock: when strict typescript compiler option enable it does not allow to access properties of objects without `if` check if it can be undefined. Here we explicitly tell compiler that `updateSet` cannot be undefined by using `!`

pleerock:
thank you very much for this awesome PR!

Issue #542, opened 2017-06-14
Repository.findByIds() throws a TypeError


aldahick:
I'm using version 0.1.0-alpha.8.
When calling `Repository.findByIds()`, a TypeError is thrown like so:
```
TypeError: Cannot read property 'map' of undefined
      at EntityManager.findByIds (C:\Users\Alex\development\foo\src\entity-manager\EntityManager.ts:472:18)
      at Repository.findByIds (C:\Users\Alex\development\foo\src\repository\Repository.ts:281:29)
      at Function.<anonymous> (C:\Users\Alex\development\foo\foo.ts:41:39)
```

aldahick:
Resolved by #543.

PR #543, opened 2017-06-14
Resolve issue #542

aldahick:
It appears that `Repository.findByIds()` wasn't actually passing the `ids` parameter to `EntityManager.findByIds()`.


pleerock:
thank you!

PR #545, opened 2017-06-15
Issue 528

ar2r13:
## Problem:
Currently, TypeORM migrations fail if new migrations have older timestamps than already applied migrations. But this is an often case when 2 or more developers work on the project simultaneously.

## Solution:
1. Find first out-of-order migration
2. Rollback all migrations after its timestamp (attention on the order)
3. Apply migrations in the correct order.

### DIFF
--- a/test/github-issues/528/issue-528.ts
+++ b/test/github-issues/528/issue-528.ts
@@ -8,10 +8,16 @@ import {Author} from "./entity/Author";
 
 describe("github issues > #528 Migrations failing on timestamp validation", () => {
 
+    class MockMigrationExecutor extends MigrationExecutor {
+        async getExecutedMigrations() {
+            return await this.loadExecutedMigrations();
+        }
+    }
+
     let connections: Connection[];
     before(async () => connections = await createTestingConnections({
         entities: [__dirname + "/entity/*{.js,.ts}"],
-        migrations: [__dirname + "/migrations/FirstReleaseChanges{.js,.ts}"],
+        migrations: [__dirname + "/migrations/NewerMigration{.js,.ts}"],
         schemaCreate: true,
         dropSchemaOnConnection: true,
     }));
@@ -33,7 +39,7 @@ describe("github issues > #528 Migrations failing on timestamp validation", () =
         connections = await createTestingConnections({
             entities: [__dirname + "/entity/*{.js,.ts}"],
             // SecondReleaseChanges has older timestamp than FirstReleaseChanges
-            migrations: [__dirname + "/migrations/SecondReleaseChanges{.js,.ts}"],
+            migrations: [__dirname + "/migrations/OldestMigration{.js,.ts}"],
             schemaCreate: true
         });
 
@@ -42,14 +48,13 @@ describe("github issues > #528 Migrations failing on timestamp validation", () =
     it("should be right migrations order", () => Promise.all(connections.map(async connection => {
         await connection.runMigrations();
 
-        class MockMigrationExecutor extends MigrationExecutor {
-            async getExecutedMigrations() {
-                return await this.loadExecutedMigrations();
-            }
-        }
         const migrations = await new MockMigrationExecutor(connection).getExecutedMigrations();
+        const [OldestMigration, NewerMigration] = migrations;
 
-        migrations[1].timestamp.should.be.greaterThan(migrations[0].timestamp);
+        OldestMigration.name.should.be.eq("OldestMigration1481283582120");
+        NewerMigration.name.should.be.eq("NewerMigration1481283582123");
+
+        NewerMigration.timestamp.should.be.greaterThan(OldestMigration.timestamp);
     })));
 
     after(function () {
--- a/test/github-issues/528/migrations/NewerMigration.ts
+++ b/test/github-issues/528/migrations/NewerMigration.ts
@@ -2,7 +2,7 @@ import {MigrationInterface} from "../../../../src/migration/MigrationInterface";
 import {Connection} from "../../../../src/connection/Connection";
 import {QueryRunner} from "../../../../src/query-runner/QueryRunner";
 
-export class FirstReleaseChanges1481283582123 implements MigrationInterface {
+export class NewerMigration1481283582123 implements MigrationInterface {
 
     async up(queryRunner: QueryRunner, connection: Connection): Promise<any> {
         await queryRunner.renameColumn("post", "title", "name");
--- a/test/github-issues/528/migrations/OldestMigration.ts
+++ b/test/github-issues/528/migrations/OldestMigration.ts
@@ -2,7 +2,7 @@ import {MigrationInterface} from "../../../../src/migration/MigrationInterface";
 import {Connection} from "../../../../src/connection/Connection";
 import {QueryRunner} from "../../../../src/query-runner/QueryRunner";
 
-export class SecondReleaseChanges1481283582120 implements MigrationInterface {
+export class OldestMigration1481283582120 implements MigrationInterface {
 
     async up(queryRunner: QueryRunner, connection: Connection): Promise<any> {
         await queryRunner.renameColumn("post", "title", "name");


pleerock:
thank you for the PR! Can you please briefly explain what was the problem and how did you fix it?

src/migration/MigrationExecutor.ts
```diff
@@ -55,17 +55,27 @@ export class MigrationExecutor {
...
             return true;
         });
 
+        // find migration with newer timestamp
+        const oldestPendingMigrationTimestamp = Math.min(...pendingMigrations.map(migration => migration.timestamp));
+
+        // migration is new and not executed. now check if its timestamp is correct
+        if (lastTimeExecutedMigration && oldestPendingMigrationTimestamp < lastTimeExecutedMigration.timestamp) {
+            // find every executed migrations with a timestamp older than oldest pending migrations timestamp
+            await executedMigrations.forEach(async migration => {
```
pleerock: you can't `await` what `forEach` returns you, right? You should do `await Promise.all(executedMigrations.map(async ...`

test/github-issues/528/issue-528.ts
```diff
@@ -0,0 +1,57 @@
...
+        const postOld = new Post();
+        postOld.title = "Hello World!";
+        postOld.author = author;
+
+        const postNew = new Post();
+        postNew.title = "Hello!";
+        postNew.author = author;
+
+        const postRepository = connection.getRepository(Post);
+
```
pleerock: i didn't understand point of your tests... you create two posts, save first one, then second... but you dont create migrations that should run, right? Don't understand what exactly you are testing here...

src/migration/MigrationExecutor.ts
```diff
@@ -55,17 +55,27 @@ export class MigrationExecutor {
...
+        // find migration with newer timestamp
+        const oldestPendingMigrationTimestamp = Math.min(...pendingMigrations.map(migration => migration.timestamp));
+
+        // migration is new and not executed. now check if its timestamp is correct
+        if (lastTimeExecutedMigration && oldestPendingMigrationTimestamp < lastTimeExecutedMigration.timestamp) {
+            // find every executed migrations with a timestamp older than oldest pending migrations timestamp
+			await Promise.all(executedMigrations.map(async migration => {
+				if (migration.timestamp > oldestPendingMigrationTimestamp) {
+                    await this.undoMigration(migration);
+                    pendingMigrations.push(migration);
```
icewind: Logically, reverted migration shouldn't be appended to the `pendingMigrations`.  They should be prepended to the `pendingMigrations` in the same order.

src/migration/MigrationExecutor.ts
```diff
@@ -55,17 +55,27 @@ export class MigrationExecutor {
         const pendingMigrations = allMigrations.filter(migration => {
             // check if we already have executed migration
             const executedMigration = executedMigrations.find(executedMigration => executedMigration.name === migration.name);
-            if (executedMigration)
-                return false;
 
-            // migration is new and not executed. now check if its timestamp is correct
-            if (lastTimeExecutedMigration && migration.timestamp < lastTimeExecutedMigration.timestamp)
-                throw new Error(`New migration found: ${migration.name}, however this migration's timestamp is not valid. Migration's timestamp should not be older then migrations already executed in the database.`);
+            if (executedMigration) return false;
```
apetushkov: could be simplified to 
return !executedMigration;

src/migration/MigrationExecutor.ts
```diff
@@ -55,17 +55,27 @@ export class MigrationExecutor {
...
             return true;
         });
 
+        // find migration with newer timestamp
+        const oldestPendingMigrationTimestamp = Math.min(...pendingMigrations.map(migration => migration.timestamp));
+
+        // migration is new and not executed. now check if its timestamp is correct
+        if (lastTimeExecutedMigration && oldestPendingMigrationTimestamp < lastTimeExecutedMigration.timestamp) {
+            // find every executed migrations with a timestamp older than oldest pending migrations timestamp
+			await Promise.all(executedMigrations.map(async migration => {
```
icewind: `await Promise.all` is not suitable here since order of migrations is crucial so we need to make sure we respect it. It is better to use `for (... of ...)` here

src/migration/MigrationExecutor.ts
```diff
@@ -55,17 +55,27 @@ export class MigrationExecutor {
...
+				if (migration.timestamp > oldestPendingMigrationTimestamp) {
```
apetushkov: Migrations should be rolled back starting from newer to oldest

src/migration/MigrationExecutor.ts
apetushkov: It is not enough to just push pending migrations to the queue. You need to implement correct order including allMigrations.
Ex., you can apply "allMigrations" starting the rollback point.

apetushkov:
@pleerock We got an issue, when 2 or more developers work on the same project and create migrations
Ex:
Dev A started work first, made a migration locally, but he works for a long time
Dev B started worrk after A, made a migration and quickly merged to master 

So we have migrations:
Dev A : 1 2 3  (locally)
Dev B:  1 2 _ 4 (locally)
then Dev B completes 
Stage: 1 2 4 (succesfully executed)
then Dev A completes 
Stage: 1 2 3 4 (fails: 3 is out of order)

So we need to rollback 4, then apply 3 then 4.
Hope it clarifies a little bit. 

But this PR need some fixes and better tests coverage.

pleerock:
@apetushkov thanks I understood the issue from @ar2r13 PR description. This is very important PR but it needs to be finished and well test covered

src/migration/MigrationExecutor.ts
```diff
@@ -55,17 +55,25 @@ export class MigrationExecutor {
...
-            if (lastTimeExecutedMigration && migration.timestamp < lastTimeExecutedMigration.timestamp)
-                throw new Error(`New migration found: ${migration.name}, however this migration's timestamp is not valid. Migration's timestamp should not be older then migrations already executed in the database.`);
-
-            // every check is passed means that migration was not run yet and we need to run it
-            return true;
+            return !executedMigration;
         });
 
+        // find migration with oldest timestamp
+        const oldestPendingMigrationTimestamp = Math.min(...pendingMigrations.map(migration => migration.timestamp));
```
apetushkov: what if pendingMigrations === [] ?

src/migration/MigrationExecutor.ts
```diff
@@ -55,17 +55,25 @@ export class MigrationExecutor {
...
+
+        // migration is new and not executed. now check if its timestamp is correct
+        if (lastTimeExecutedMigration && oldestPendingMigrationTimestamp < lastTimeExecutedMigration.timestamp) {
+            // find every executed migrations with a timestamp older than oldest pending migrations timestamp
+            for (let migration of executedMigrations) {
```
apetushkov: looks like an incorrect order for undo

src/migration/MigrationExecutor.ts
ar2r13: if pendingMigrations.length is 0 then nothing to migrate

test/github-issues/528/issue-528.ts
```diff
@@ -0,0 +1,34 @@
...
+    let connections: Connection[];
+    before(async () => connections = await createTestingConnections({
+        entities: [__dirname + "/entity/*{.js,.ts}"],
+        migrations: [__dirname + "/migrations/*{.js,.ts}"],
+        schemaCreate: true,
+        dropSchemaOnConnection: true,
+    }));
+    after(() => closeTestingConnections(connections));
+
```
apetushkov: Missing a test for made changes. 
You need to check correct order of undo / apply (at least)

src/migration/MigrationExecutor.ts
ar2r13: why?
1. Find the oldest pending migration
2. Compare each executed migration with oldest pending migration and undo if it older then...
3. After concat with all migrations and sort

src/migration/MigrationExecutor.ts
apetushkov: oldestPendingMigrationTimestamp will be Infinity

src/migration/MigrationExecutor.ts
apetushkov: if you applied A then B, you need to rollback B then A

src/migration/MigrationExecutor.ts
apetushkov: Also, lastTimeExecutedMigration could be unset

ar2r13:
How i can get executed migrations list in the test?

Ostrzy:
@pleerock I would like to ask - why do we care about timestamps ordering that much? 

Most of the operations are really independent of each other and in general if some migrations are that much connected that you need to have them in order, you should work on them together anyway. I was working with Rails migrations previously and I have never run into situation where error about timestamps not being in order would be useful.

I am asking this question mainly because I'm a little worried about the approach taken in this PR. From my experience it's pretty typical that you may have a long running branch which eventually gets merged into master. With approach taken here, we may accidentally erase data from production environment if the long running branch had a migration. To get back to the moment where we want to apply our new migration, we would drop tables and remove columns, effectively erasing the data.

luchillo17:
@Ostrzy You sure you talking the same Rails i know? as far as i know Rails (actually Active Record, the Rails ORM) does takes in account the timestamp to determine the order of migrations to run, though not sure what does when it finds some out of order.

If you look at Rails docs about [Active Record#Creating a standalone migration](http://edgeguides.rubyonrails.org/active_record_migrations.html#creating-a-standalone-migration), it does say that it takes in account the timestamps to determine which should be run and in what order in paragraph 1 in last 3 lines.

Ostrzy:
@luchillo17 I double checked on some simple example and it confirmed that ActiveRecord does not care if migration is executed out of order. It just looks at `schema_migrations` table and if anything in `db/migrations` is missing there, it will run in without problems or notice.
ActiveRecord uses timestamps to determine the order of applying/reverting the migrations. Even if migration was just applied out of order, if you try to revert single migration, it will revert the migration with top timestamp, not the one migrated out of order.

luchillo17:
So it will try to run in order but if there's just pending migrations out of order it will run it regardless, is that right?
Also it will revert in order even if last migration run was out of order, is that right as well?

Ostrzy:
Yes and yes.

pleerock:
Sorry, but Im going to close this PR since its not relevant anymore. Lets discuss a new proposal implementation in #1315 and fix this issue for everybody.

PR #565, opened 2017-06-22
Add missing TYPEORM_MIGRATIONS

netes:
It seems that `TYPEORM_MIGRATIONS` variable is missing in `ConnectionOptionsEnvReader.ts`. I've added it. This probably caused migrations failure when using `ormconfig.env` (migrations couldn't be found).


pleerock:
thanks!

PR #574, opened 2017-06-23
Exit process after finishing migrations

netes:
Hi there,
For some reason, after running migrations or reverting them, node process doesn't quit, which makes it hard to run on staging servers or CI (results in timeouts because of hanged process).
I've made a quick fix but I'm not 100% convinced this is the right way to handle this.

What do you think?

### DIFF
--- a/src/commands/MigrationRevertCommand.ts
+++ b/src/commands/MigrationRevertCommand.ts
@@ -45,6 +45,8 @@ export class MigrationRevertCommand {
         } catch (err) {
             console.error(err);
             // throw err;
+        } finally {
+            process.exit(0);
         }
     }
 
--- a/src/commands/MigrationRunCommand.ts
+++ b/src/commands/MigrationRunCommand.ts
@@ -45,6 +45,8 @@ export class MigrationRunCommand {
         } catch (err) {
             console.error(err);
             // throw err;
+        } finally {
+            process.exit(0);
         }
     }
 


src/commands/MigrationRevertCommand.ts
```diff
@@ -45,7 +45,9 @@ export class MigrationRevertCommand {
         } catch (err) {
             console.error(err);
             // throw err;
+        } finally {
+            process.exit(0);
```
pleerock: its not a fix, we need to find a reason why its not finishing instead of forcely close it.

netes: I agree, but this helped me as a hotfix. If I'll have some more time I'll try to narrow the problem.

pleerock:
Are you using latest typeorm version? And what database are you using? Because I don't such problem, process is finished successfully:

```shell
Umeds-MacBook-Pro-2:typescript-example pleerock$ typeorm migrations:run
executing query: SELECT * FROM INFORMATION_SCHEMA.COLUMNS WHERE TABLE_SCHEMA = 'test' AND TABLE_NAME = 'migrations'
executing query: SELECT * FROM `migrations` `migrations`
0 migrations are already loaded in the database.
1 migrations were found in the source code.
1 migrations are new migrations that needs to be executed.
executing query: START TRANSACTION
executing query: CREATE TABLE `post_categories_category` (`postId` int(11) NOT NULL, `categoryId` int(11) NOT NULL, PRIMARY KEY(`postId`, `categoryId`)) ENGINE=InnoDB
executing query: ALTER TABLE `post` DROP `name`
executing query: ALTER TABLE `post` DROP `int`
executing query: ALTER TABLE `post` DROP `tinyint`
executing query: ALTER TABLE `post` DROP `smallint`
executing query: ALTER TABLE `post` DROP `mediumint`
executing query: ALTER TABLE `post` DROP `bigint`
executing query: ALTER TABLE `post` DROP `float`
executing query: ALTER TABLE `post` DROP `double`
executing query: ALTER TABLE `post` DROP `decimal`
executing query: ALTER TABLE `post` DROP `date`
executing query: ALTER TABLE `post` DROP `datetime`
executing query: ALTER TABLE `post` DROP `timestamp`
executing query: ALTER TABLE `post` DROP `time`
executing query: ALTER TABLE `post` DROP `year`
executing query: ALTER TABLE `post` DROP `char`
executing query: ALTER TABLE `post` DROP `varchar`
executing query: ALTER TABLE `post` DROP `blob`
executing query: ALTER TABLE `post` DROP `tinyblob`
executing query: ALTER TABLE `post` DROP `tinytext`
executing query: ALTER TABLE `post` DROP `mediumblob`
executing query: ALTER TABLE `post` DROP `mediumtext`
executing query: ALTER TABLE `post` DROP `longblob`
executing query: ALTER TABLE `post` DROP `longtext`
executing query: ALTER TABLE `post` DROP `enum`
executing query: ALTER TABLE `post` DROP `simpleArray`
executing query: ALTER TABLE `post` ADD `title` varchar(255) NOT NULL
executing query: CREATE INDEX `ind_bf0cba7428bf40a2c8d3848ade` ON `post_categories_category`(`postId`)
executing query: CREATE INDEX `ind_df5ae6425e71923e23734eb18c` ON `post_categories_category`(`categoryId`)
executing query: ALTER TABLE `post_categories_category` ADD CONSTRAINT `fk_00ee6a3b536439f11fb9e65d762` FOREIGN KEY (`postId`) REFERENCES `post`(`id`)
executing query: ALTER TABLE `post_categories_category` ADD CONSTRAINT `fk_6e347a1626d70d8dbc0e2ea8bd7` FOREIGN KEY (`categoryId`) REFERENCES `category`(`id`)
executing query: INSERT INTO `migrations`(`timestamp`, `name`) VALUES (?,?) -- PARAMETERS: [1498207416999,"first1498207416999"]
Migration first1498207416999 has been executed successfully.
executing query: COMMIT
Umeds-MacBook-Pro-2:typescript-example pleerock$ 
```

I checked it on mysql.

netes:
I'm running migrations via `ts-node`, maybe that's the cause. I'll check it in a few hours probably and let you know.

netes:
@pleerock sorry for the delay, I've checked without `ts-node` and still got the problem. DB is Postgres:
```executing query: INSERT INTO "role" (name) VALUES ('admin')
executing query: INSERT INTO "role" (name) VALUES ('landlord')
executing query: INSERT INTO "role" (name) VALUES ('tenant')
executing query: INSERT INTO "role" (name) VALUES ('applicant')
executing query: INSERT INTO "migrations"("timestamp", "name") VALUES ($1,$2)  -- PARAMETERS: [1498547082640,"addRoles1498547082640"]
Migration addRoles1498547082640 has been executed successfully.
executing query: COMMIT```

And it hangs forever.

pleerock:
okay I'll check it with postgres

netes:
It seems that connection with DB is not closed properly. I've checked `connection.isConnected` in the last `finally` clause and it's still connected.

bartlomiej-korpus:
@netes I've got the same problem with my app, with 0.1.0 the app doesn't close properly when restarted by nodemon because the DB connection doesn't close. Postgres

pleerock:
@netes @bartlomiej-korpus can you guys confirm that it still works wrong in `0.1.0-alpha.25` ?

netes:
Yup. Just updated and connection is still not closed.

pleerock:
It was postgres-only issue. Fixed in `0.1.0-alpha.28`

netes:
Great! Thanks!

Issue #587, opened 2017-06-27
Wrong ordering of fields in composite index.


bartlomiej-korpus:
When creating a composite index, the ordering of fields isn't maintained, instead, ordering of the columns returned by metadata is used.
The bug is in file
[https://github.com/typeorm/typeorm/blob/0.1.0/src/metadata/IndexMetadata.ts](IndexMetadata.ts)

in lines 78-100, the filtering used to detect inexistent fields changes the order.

I would like to make a PR against 0.0.x branch first, as thats what I am using at the moment.

pleerock:
if you want to create PR against 0.0.x feel free to do that, but be aware that in 0.0.x composite primary keys were not working correctly (especially in relations). Correct composite primary keys support is one of the major change in 0.1.0.

Also link you sent is wrong, there is no 0.1.0 branch in typeorm repo.

bartlomiej-korpus:
Wow, I have no idea where I got that link, though I copy-pasted it from the browser.
Yes I've read the composite primary keys don't work correctly, these indices I am using for indexing for faster searches.

This is the correct [https://github.com/typeorm/typeorm/blob/0.0.x-version/src/metadata/IndexMetadata.ts#L97](link)

Also I can see 0.1.0 branch in TypeORM repo.

pleerock:
Okay, send a PR. Why are you using `0.0.x`. I strongly recommend you to switch to 0.1.0 because it has lot of important fixes.

bartlomiej-korpus:
I thought that -alpha postfix meant it shouldn't be used in production yet. Should it?

pleerock:
theoretically everything under 1.0.0 should not be used in production 😆 I recommend you to use it, but I constantly make changes, so be ready to change your code and table schema if needed. Also probably incremental changes will be easier for you then migrate from 0.0.x to 0.1.0 then.

bartlomiej-korpus:
Okay, I will migrate! How many versions do you expect between now and 0.1.0?

pleerock:
many, hundreds probably 😆

bartlomiej-korpus:
Do you document changes between alphas? Or is there some place with docs for 0.1.0? typeorm.github.io is for 0.0.11 right?

pleerock:
[...]

PR #607, opened 2017-07-03
fix ordering of columns in IndexMetadata

bartlomiej-korpus:
This PR fixes issue #587 
This probably isn't the most elegant solution to this problem that one can come up with. This is the safest modification I could do with my lacking knowledge about the codebase.


pleerock:
Thank you! I'll publish a new version with this change tomorrow.

pleerock:
released in 0.1.0-alpha.27

PR #611, opened 2017-07-04
fix issue-587 for relations and refactor field matching in IndexMetadata

bartlomiej-korpus:
Another fix regarding issue #587 which now also fixes it for relations. I am sorry I did not see it coming with previous fix. 
Please review this code thoroughly, I've pretty much changed it to the other direction, from searching of entity's fields and entity's relations in index given column names, now index column names are mapped into their corresponding column/relation.joinColumns. This lets us preserve the ordering.


bartlomiej-korpus:
Okay so I fixed the failing tests, it was the build that failed because of the "noImplicitAny", still I have no idea why it wasn't and still isn't failing locally for me(not for tsc or npm test). I rebased it with upstream/master because I thought something might have changed, but it wasnt a case.

pleerock:
Can you please create a clean PR with only your changes, because its impossible to review your change with other merged changes?

PR #615, opened 2017-07-05
fixed: default orderby

taoqf:


pleerock:
Hi and thank you for your fix! Would be great to have a test that will check this functionality, to prevent this functionality break in the future.

taoqf:
@pleerock ,Actually I am not quite familiar with this testing system yet.
A test like [sample30-default-order-by](https://github.com/typeorm/typeorm/tree/master/sample/sample30-default-order-by) will be proper.

pleerock:
` sample30-default-order-by` is a sample, its not a `test`. For tests refer `test` directory. Its easy to write test for this issue, just take a look how other tests are implemented.

taoqf:
I will close this issue because it will resolve all the problems default order causes.

pleerock:
what do you mean? Why did you close it?

taoqf:
I will try to fix this and add test later.

Issue #610, opened 2017-07-04
ECONNREFUSED not thrown correctly


joffray589:
I created a sample repository (https://github.com/joffray589/typeorm-sample) which expose this problem, here is the main.js
````
let dbOptions: ConnectionOptions = { /* ... */ };

(async () => {
    try{
        let connection = await createConnection(dbOptions); // (1)
        console.log("connection.isConnected = " + connection.isConnected);
        await connection.getRepository(FooEntity).find();       // (2)
        console.log("Mysql connection ok");
    }
    catch(e){
        console.error("Mysql connection error : " + e );
    }
})();
````

If I stop mysql server and run this example with autoSchemaSync=true, ECONNREFUSED will be thrown at (1).
But if i set autoSchemaSync to false and run this example again, ECONNREFUSED will be thrown at (2) and connection.isConnected will return true.

Here is my environnement: 

Windows 7,
typeorm: 0.0.10,
mysql: 2.13.0

**Comment by joffray589** [comment]
After some investigations, I think that this issue comes from the MysqlDriver.connect() method. 
The pool attribute is instanciated but the connection should be tested after.
I think this issue should be resolved by adding this code at the end of the connect method :
 ```
        return new Promise<void>((resolve, reject) => {
            this.pool.getConnection(function(err, connection) {
                if(err){
                    reject("Mysql connection error : " + err);
                }
                else{
                    resolve();
                }
            });
        });
```

I'm going to make some tests and I'll make a pull request if it works.

joffray589:
And here is the pull request : #618

pleerock:
fixed by #619

PR #619, opened 2017-07-05
Update MysqlDriver.ts to fix issue #610

joffray589:


pleerock:
tests are failing

pleerock:
I don't understand the reason of this change. Why do you need to get the connection? How exactly it fixes your problem? I need to understand the logic behind this change.

joffray589:
We need to get the connection because it looks like that the mysql library allow pool creation even if the mysql server is stopped. 
This PR fixes the problem because it ensure that the connection is working at the end of the MysqlDriver.connect() method call.

joffray589:
The tests were failing because of tslint whitespace rule. I fixed this on my branch,
but it actually looks like that this issue is more related to the mysql library than typeorm.

I'm going to report this issue on mysql repository, I can also submit a clean PR if you want to.

joffray589:
The test were failing because of tslint whitespace rule, I fixed it but the travis build gave me a strange error related to event.js. It looks like that this is because another build was running at the same time ([source](https://stackoverflow.com/questions/34290796/events-js-141-throw-er-unhandled-error-event)).
Can you restart this build job? [build-link](https://travis-ci.org/typeorm/typeorm/jobs/250437142)

src/driver/mysql/MysqlDriver.ts
```diff
@@ -154,6 +154,19 @@ export class MysqlDriver implements Driver {
         }, this.options.extra || {});
 
         this.pool = this.mysql.createPool(options);
+        
+        return new Promise<void>((ok, fail) => {
+            this.pool.getConnection((err: any, connection: any) => {
```
pleerock: you need to call `connection.release()` because you need to release each connection you are getting from pool

src/driver/mysql/MysqlDriver.ts
```diff
@@ -154,6 +154,19 @@ export class MysqlDriver implements Driver {
         }, this.options.extra || {});
 
         this.pool = this.mysql.createPool(options);
+        
+        return new Promise<void>((ok, fail) => {
```
pleerock: also please write a commend why are you doing it because its not really oblivious why do you need this code block

pleerock:
Okay I'll do rest changes by my own and release a new version. Thank you for your contribution!

joffray589:
You're welcome, and thanks for this awesome library!

PR #626, opened 2017-07-06
Improve DirectoryExportedClassesLoader

asvetliakov:
1. Moved array checking on top of the object checking since Array is the instance/typeof object
2. Replaced ```instanceof``` calls with ```typeof``` calls.

For some reason, the ```exported instanceof Object``` check is giving ```false``` in my Jest test files when i'm creating DB connection so the typeorm is unable to process entities/migrations/subscribers. This may be caused by heavy VM contexts manipulating in Jest itself.

### DIFF
--- a/src/util/DirectoryExportedClassesLoader.ts
+++ b/src/util/DirectoryExportedClassesLoader.ts
@@ -6,16 +6,16 @@ import {PlatformTools} from "../platform/PlatformTools";
 export function importClassesFromDirectories(directories: string[], formats = [".js", ".ts"]): Function[] {
 
     function loadFileClasses(exported: any, allLoaded: Function[]) {
-        if (exported instanceof Function) {
+        if (typeof exported === "function") {
             allLoaded.push(exported);
 
-        } else if (exported instanceof Object) {
+        } else if (Array.isArray(exported)) {
+            exported.forEach((i: any) => loadFileClasses(i, allLoaded));
+
+        } else if (typeof exported === "object") {
             Object.keys(exported).forEach(key => loadFileClasses(exported[key], allLoaded));
 
-        } else if (exported instanceof Array) {
-            exported.forEach((i: any) => loadFileClasses(i, allLoaded));
         }
-
         return allLoaded;
     }
 


asvetliakov:
There aren't any tests since looks like you're not testing ```util/``` directory at all

src/util/DirectoryExportedClassesLoader.ts
```diff
@@ -6,16 +6,16 @@ import {PlatformTools} from "../platform/PlatformTools";
 export function importClassesFromDirectories(directories: string[], formats = [".js", ".ts"]): Function[] {
 
     function loadFileClasses(exported: any, allLoaded: Function[]) {
-        if (exported instanceof Function) {
+        if (typeof exported === "function") {
```
pleerock: why?

src/util/DirectoryExportedClassesLoader.ts
```diff
@@ -6,16 +6,16 @@ import {PlatformTools} from "../platform/PlatformTools";
...
             allLoaded.push(exported);
 
-        } else if (exported instanceof Object) {
+        } else if (Array.isArray(exported)) {
```
pleerock: why? `exported instanceof Array` already worked good

src/util/DirectoryExportedClassesLoader.ts
asvetliakov: For consistency with ```typeof === "object"```. Change it back?

src/util/DirectoryExportedClassesLoader.ts
asvetliakov: ```Array.isArray``` is working better within different context/iframes. Wanna to be have changed it back?

pleerock: No you can left it as it is, however next time don't change exist code if its the same. Because for example `instanceof Array` used in 100 other places and when one change differs it brings questions and inconsistency

src/util/DirectoryExportedClassesLoader.ts
pleerock: same answer as for array, you can left it however in the codebase everywhere instanceof Function is used and it brings inconsistency with other places

pleerock:
Thank you for the fix!

asvetliakov:
Thanks for merging! Can you push a new build? This issue is killing TDD workflow with jest for us 😄

PR #627, opened 2017-07-07
fix column ordering in composite index with relation fields

bartlomiej-korpus:
Clean PR regarding issue #587 (previous PR #611 was cluttered with merged changes) with fix for relation fields.

### DIFF
--- a/src/metadata/IndexMetadata.ts
+++ b/src/metadata/IndexMetadata.ts
@@ -112,26 +112,18 @@ export class IndexMetadata {
                 }
             }
 
-            // console.log("columnPropertyNames:", columnPropertyNames);
-            // console.log("this.entityMetadata.columns:", this.entityMetadata.columns.map(column => column.propertyPath));
-            const columns = this.entityMetadata.columns
-                .filter(column => columnPropertyNames.indexOf(column.propertyPath) !== -1)
-                .sort((a, b) => columnPropertyNames.indexOf(a.propertyPath) - columnPropertyNames.indexOf(b.propertyPath));
-            // console.log("columns:", columns.map(column => column.propertyPath));
-            this.entityMetadata.relations
-                .filter(relation => relation.isWithJoinColumn && columnPropertyNames.indexOf(relation.propertyName) !== -1)
-                .forEach(relation => columns.push(...relation.joinColumns));
-
-            // todo: better to extract all validation into single place if possible
-            const missingColumnNames = columnPropertyNames.filter(columnPropertyName => {
-                return !this.entityMetadata.columns.find(column => column.propertyPath === columnPropertyName) &&
-                    !this.entityMetadata.relations.find(relation => relation.isWithJoinColumn && relation.propertyPath === columnPropertyName);
-            });
-            if (missingColumnNames.length > 0) {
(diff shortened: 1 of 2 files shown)


bartlomiej-korpus:
@pleerock No idea why Maria on node 6 errored.

src/metadata/IndexMetadata.ts
```diff
@@ -112,26 +112,18 @@ export class IndexMetadata {
...
-            const missingColumnNames = columnPropertyNames.filter(columnPropertyName => {
-                return !this.entityMetadata.columns.find(column => column.propertyPath === columnPropertyName) &&
-                    !this.entityMetadata.relations.find(relation => relation.isWithJoinColumn && relation.propertyPath === columnPropertyName);
-            });
-            if (missingColumnNames.length > 0) {
-                throw new Error(`Index ${this.givenName ? "\"" + this.givenName + "\" " : ""}contains columns that are missing in the entity: ` + missingColumnNames.join(", "));
-            }
-
-            this.columns = columns;
+            this.columns = columnPropertyNames.map(propertyName => {
```
pleerock: Im not sure about your changes. Why resulted code is less then original? Where are some checks like checking `relation.propertyPath` ? I think this change broke untested functionality.

pleerock:
Sometimes sqlite fail on CI I don't know why (and cant find a reason). Just need to re-run ci tests. Please take a look on my commend. Also remember you need to provide a test for your change.

src/metadata/IndexMetadata.ts
bartlomiej-korpus: relation.propertyPath was used to check if some columns were missing in the model but it wasn't used when building "columns" array, which makes it of no use anyway(relation.propertyName was used here instead in the old code).

Following this, I assumed that if relation.propertyName is being used for building "columns", its the valid field to use, and I rebuilt checking of missing columns to checking if ANY column is missing instead of checking WHICH columns are missing, that makes the code shorter(I also changed the error message).

I would love to fix what you think is broken but to do this, I need to ask you whats the difference and why is .propertyName being used to build "columns" but .propertyPath is being used to find missing columns instead.

pleerock:
are you sure your tests are correct and actually check what you actually do in this PR? Because you didn't add new assertions based on what you want to do in this PR. I'm sure if you don't add good test coverage this functionality will be broken very soon because there are lot of other issues with indices that needs to be fixed and fixing code will change this code again.

bartlomiej-korpus:
@pleerock 
I can write tests for old code, and then make the new code pass them, I just need you to help me figuring out what would be good tests for the "build" method in IndexMetadata. 
I can think of following ones:
should build proper index fields from simple columns
should build proper index fields from relations (simple key? more?)
should build proper index fields with embeddeds (should?)
what more?

In this PR i changed the old test for this issue to include relations.

pleerock:
you probably need to check if indices are generated in the database in proper order. Would be great to have scenarios for embeds too.

bartlomiej-korpus:
The test i commited checks if the indices are generated by IndexMetadata in proper order, not in database. Its hard to deal with indices in DB because there is no unified way to get info about index.

pleerock:
Thank you for your hard work! This will be released in next version this week.

PR #634, opened 2017-07-09
feat(Logger): use debug package for logging

### LINKED ISSUE #633: feature request: use debug package for logging
With #505 closed, log messages are highlighted in the terminal which is awesome, however it cannot work when using a custom logger function. The reason we use custom logger is to control an universal way of what we log in different environments.

The package we use is `debug` used both by Express & Koa for logging. The idea is to specify different name spaces in the `DEBUG` environment variable on startup to select what to log. 

An example: `typeorm:*,-typeorm:query`. This would log every message from TypeORM except the ones when written on queries. (These are listed here: [src/logger/Logger.ts#L73-L92](https://github.com/typeorm/typeorm/blob/master/src/logger/Logger.ts#L73-L92) )

To achieve this the following have to be done:
- use `debug` package for logging
- do not return when no logging options are present
- remove old way of specifying logging options (via `LoggerOptions`)

Things worth to discuss: 
- console.log calls in cli commands, I left them untouched, I think there is no benefit filtering them. When someone runs a command she/he wants to see the output. 
- where to add this in the docs

NoNameProvided:
This PR will

- use `debug` package for logging
   - introducing the `typeorm:schema`, `typeorm:query`, `typeorm:log`, `typeorm:info`, `typeorm:warn`, `typeorm:error` namespaces
- remove old way of specifying what to log
   - removing the `LoggerOptions.logQueries`, `LoggerOptions.logFailedQueryError `, `LoggerOptions.logOnlyFailedQueries` preferences
   - still keeping the ability to add the custom `LoggerOptions.logger` function
   - making the logger function do not depend on the availability of the `LoggerOptions` object

closes #633

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,0 +1,6792 @@
+{
+  "name": "typeorm",
+  "version": "0.1.0-alpha.31",
+  "lockfileVersion": 1,
+  "requires": true,
+  "dependencies": {
+    "@gulp-sourcemaps/identity-map": {
+      "version": "1.0.1",
+      "resolved": "https://registry.npmjs.org/@gulp-sourcemaps/identity-map/-/identity-map-1.0.1.tgz",
+      "integrity": "sha1-z6I7xYQPkQTOMqZedNt+epdLvuE=",
+      "dev": true,
+      "requires": {
+        "acorn": "5.1.1",
+        "css": "2.2.1",
+        "normalize-path": "2.1.1",
+        "source-map": "0.5.6",
+        "through2": "2.0.3"
+      },
+      "dependencies": {
+        "acorn": {
+          "version": "5.1.1",
+          "resolved": "https://registry.npmjs.org/acorn/-/acorn-5.1.1.tgz",
+          "integrity": "sha512-vOk6uEMctu0vQrvuSqFdJyqj1Q0S5VTDL79qtjo+DhRr+1mmaD+tluFSCZqhvi/JUhXSzoZN2BhtstaPEeE8cw==",
+          "dev": true
+        },
+        "source-map": {
+          "version": "0.5.6",
+          "resolved": "https://registry.npmjs.org/source-map/-/source-map-0.5.6.tgz",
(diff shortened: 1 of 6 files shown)


NoNameProvided:
Ahh I see, needs to fix bunch of tests because the changes in `LoggerOptions`, before fixing them @pleerock can you give your blessing on these changes? Can be this merged when tests are fixed?

pleerock:
I don't see point of using `debug` package. In TypeORM you can easily provide a custom logger where anyone who want can use `debug` or any other package he wants.

NoNameProvided:
Hi @pleerock! 

The `debug` package is the 9yh most dependent package ((ref)[https://www.npmjs.com/browse/depended/]), it's used by default for express & koa and i am sure a few other web framework as well. 

While in one hands it allows greater flexibility to have a custom solution. In the other hand it's good to provide battle tested, widely used libs for standard stuff, so developers can re-use their knowledge.

pleerock:
I still don't understand why we should use debug package over winston or something else. Or use anything at all. 

 I have refactored how logger is working in latest commits. See how it now works [here](https://github.com/typeorm/typeorm/blob/master/docs/logging.md).

I suggest you to create separate package for this purpose, something like `typeorm-logger-debug` and use new functionality I provided in latest commits.

NoNameProvided:
Will do thanks!

pravdomil:
**before**
before

**after**
after


@pleerock do you consider about adding one more `logger` into typeorm called `"debug"`?

pravdomil:
here is my implementation 
```ts
export class DebugLogger implements Logger {
    private debugQuery = debug("typeorm:query");
		
    logQuery(query: string, parameters?: any[], queryRunner?: QueryRunner) {
        if (this.debugQuery.enabled) {
            const sql = query + (parameters && parameters.length ? " -- PARAMETERS: " + this.stringifyParams(parameters) : "");
            this.debugQuery(PlatformTools.highlightSql(sql));
        }
    }
}
```

pleerock:
Im worrying about growing typeorm's dependency list...

Okay let's add it experimental. Create a PR

PR #640, opened 2017-07-10
Use the naming strategy on TreeRepository

hisamu:
While trying to use the `findRoots`method, the `parentPropertyName` field is not using the `joinColumnName` method from the naming strategy. This PR fixes the issue.


pleerock:
Thanks!

hisamu:
@pleerock it seems the version `0.1.0-alpha.32` is not yet available on npm/yarn. Is it possible to install it?

pleerock:
I did not release it yet but will do today or tomorrow

PR #650, opened 2017-07-12
Insert a column in a specific position on MySQL

hisamu:
Right now is not possible to add a column specifying the position it should be placed on the schema. With this PR this becomes possible for MySQL.

I'm not sure if this column could be helpful on another DB, though. If there's a way of adding an option only for a certain driver I can refactor it.

### DIFF
--- a/src/driver/mysql/MysqlQueryRunner.ts
+++ b/src/driver/mysql/MysqlQueryRunner.ts
@@ -421,9 +421,9 @@ export class MysqlQueryRunner implements QueryRunner {
     /**
      * Creates a new column from the column schema in the table.
      */
-    async addColumn(tableSchemaOrName: TableSchema|string, column: ColumnSchema): Promise<void> {
+    async addColumn(tableSchemaOrName: TableSchema|string, column: ColumnSchema, afterColumn?: string): Promise<void> {
         const tableName = tableSchemaOrName instanceof TableSchema ? tableSchemaOrName.name : tableSchemaOrName;
-        const sql = `ALTER TABLE \`${tableName}\` ADD ${this.buildCreateColumnSql(column, false)}`;
+        const sql = `ALTER TABLE \`${tableName}\` ADD ${this.buildCreateColumnSql(column, false, afterColumn)}`;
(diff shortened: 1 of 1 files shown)


pleerock:
`after` should not be placed in `schema`. Table schema is representation of database schema, but you add "operation" into it. This method looks like it should be part of `QueryRunner` alongside with `addColumn` method. We can add method like `addColumnAfter` and/or `addColumnBefore`, but its better if we do something like `insertColumnAt(position, column)`, but we also need to store a position of each columns in `ColumnSchema`. We also need ability to simply move columns ordering. We can create a new method like `moveColumn` which can move column to any position. 

Also PR must implement this method for all drivers.

hisamu:
Agreed, it should be part of the `QueryRunner`. A `addColumnAfter` method would be a good option.

Regarding `before` and `position` options`, this isn't possible in MySQL. You can either add a column as first or after another column. 

As of other drivers, postgres doesn't support setting the column order and as of my research it seems to be the case for MSSQL, Oracle, SQLite and WebSQL as well. MongoDB doesn't have a schema, so this is only suitable for MySQL.

That being the case, this method could be added to the MySQL driver and we type hint the migration like this if needed:

```TypeScript
public async up (queryRunner: MysqlQueryRunner): Promise<any>
```

hisamu:
I updated the code to reflect that change on the `addColumn` method. This works, but maybe having a driver specific options interface would be better. 

What do you think?

pleerock:
I think better to create a separate method `addColumnAfter` instead of making ugly current addColumn method. And yeah it will be mysql-specific and won't be in interface until we implement same functionality for all other drivers.

Once you finish, please make sure to add tests in this PR

hisamu:
I started to do that but I had to repeat almost all the code from `addColumn`. I'll try to refactor it to minimize copy-paste. I'll add some tests as well.

pleerock:
@hisamu any further progress?

mzygmunt:
It will nice to see this feature in future releases.

pleerock:
Closing as no further actions from PR author.

Issue #145, opened 2016-12-23
Relation update failing during normal table update


MattJeanes:
Hi, I've been trying to do a simple update on a table with a one-to-many, but it's failing with the following SQLite error: `SQLITE_CONSTRAINT: NOT NULL constraint failed: comments.ticket_id`

Dumping relevant stripped down code...

```typescript
// Ticket

import { Table, Column, PrimaryGeneratedColumn, ManyToOne, OneToMany } from 'typeorm';
import { User } from './User';
import { Comment } from './Comment';

@Table("tickets")
export class Ticket {
    @PrimaryGeneratedColumn()
    id: number;

    @OneToMany(type => Comment, comment => comment.ticket)
    comments: Comment[] = [];

    @Column("int", { name: "c_vsts" })
    vsts: number;
}
```

```typescript
// Comment

import { Table, Column, PrimaryGeneratedColumn, ManyToOne, JoinColumn } from 'typeorm';
import { Ticket } from './Ticket';
import { User } from './User';

@Table("comments")
export class Comment {
    @PrimaryGeneratedColumn()
    id: number;

    @ManyToOne(type => Ticket, ticket => ticket.comments)
    @JoinColumn({ name: "ticket_id" })
    ticket: Ticket;

    @Column("text")
    body: string;
}
```

```typescript
// In this case, key is 'vsts' and value is '5335'
public async edit<K extends keyof Ticket>(id: number, key: K, value: Ticket[K]) {
    var conn = await this.db.getConnection();
    var ticketRepo = conn.getRepository(Ticket);
    var ticket = await ticketRepo.findOneById(id);
    if (!ticket) { throw "Invalid ticket" };

    ticket[key] = value;
    await ticketRepo.persist(ticket);
};
```

```typescript
// https://github.com/typeorm/typeorm/blob/fc37ca1262e6bc22c9e55083330af43360f9f9ed/src/persistence/SubjectBuilder.ts#L664-L668
} else if (relation.isOneToMany && relation.inverseRelation) {
	relatedEntitySubject.relationUpdates.push({
		relation: relation.inverseRelation,
		value: null
	}); // todo: implement same for one-to-one
}
```

So basically after crawling around the code it looks like the above code from the SubjectBuilder appears to be pushing a null relation update into it. This results in the following query being attempted during `SubjectOperationExecutor.executeUpdateRelations`, resulting in the SQLite error at the top:

`UPDATE "comments" SET "ticket_id"=null WHERE "id"=81220`

[...]

PR #652, opened 2017-07-12
Update documentation to remove references to self-initialised arrays

guscastro:
As per discussion in [#145](https://github.com/typeorm/typeorm/issues/145), self-initialised arrays on relations may lead to unintended updates as typeorm assumes an empty array means all relations have been removed. Hence, remove references to this practice from the documentation to avoid confusion.

### DIFF
--- a/README-zh_CN.md
+++ b/README-zh_CN.md
@@ -863,7 +863,7 @@ export class Album {
         cascadeRemove: true  // 在移除Album时，会自动移除相册里的Photo
     })
     @JoinTable()
-    photos: Photo[] = []; // 初始化个Photo数组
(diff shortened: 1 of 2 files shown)


guscastro:
I may have assumed what the Chinese meant. 🙂

pleerock:
can you please check everywhere places where these properties are used to make sure they are not using `push` without in-place array initialization.

guscastro:
Did you mean within the documentation examples? I couldn't find any examples that use those properties.
I'm also not sure how this PR would've broken the build. Any ideas?

pleerock:
Thanks

Issue #660, opened 2017-07-13
Specifying a RETURNING or OUTPUT clause with QueryBuilder


tjme:
Is there a way to explicitly specify a PostgreSQL RETURNING (or MSSQL OUTPUT) clause for INSERT and UPDATE queries, using QueryBuilder in TypeORM version 0.1.0?
I note it is mentioned in #361, but that is for EntityManager and appears to be implicitly generated.

pleerock:
explicitly there is no way at the moment. But theoretically we can do that via `InsertQueryBuilder` by adding `returning` method. But it will work only in postgres and mssql. Its easy feature to implement, if you want feel free to contribute.

tjme:
I've had a go at drafting some simple changes as below (please also see attached zip for more details).
They may be wrong, and there's no parameter checking or error reporting!

Note also that it seems that PostgreSQL and MSSQL support the clause for DELETE as well.

Add to QueryExpressionMap.ts (perhaps after valuesSet):
    returning: string = "";

In each of DeleteQueryBuilder.ts, InsertQueryBuilder.ts and UpdateQueryBuilder.ts, add:
import {SqlServerDriver} from "../driver/sqlserver/SqlServerDriver";
import {PostgresDriver} from "../driver/postgres/PostgresDriver";
...
    returning(returning: string): this {
        this.expressionMap.returning = returning;
        return this;
...
and replace existing return statement with (e.g. for insert):
        if (this.expressionMap.returning !== "" && this.connection.driver instanceof PostgresDriver) {
            return `INSERT INTO ${tableName}(${columnNames}) VALUES ${values} RETURNING ${this.expressionMap.returning}`;
        } else if (this.expressionMap.returning !== "" && this.connection.driver instanceof SqlServerDriver) {
            return `INSERT INTO ${tableName}(${columnNames}) OUTPUT ${this.expressionMap.returning} VALUES ${values}`;
        } else {
            return `INSERT INTO ${tableName}(${columnNames}) VALUES ${values}`;
        }

[TypeORM-Returning.zip](https://github.com/typeorm/typeorm/files/1149287/TypeORM-Returning.zip)

pleerock:
zip? OMG dude its github, you should publish your code in git repos! Also please format your code properly, its very hard to read what you wrote. Yeah you need to do something like what you wrote, but try to remove duplication of "INSERT INTO" statement. Simply create 4 blocks of code - insert, output, values, returning. You are on the right way so start creating a pull request.

[...]

PR #670, opened 2017-07-15
#660 Specifying a RETURNING or OUTPUT clause with QueryBuilder

tjme:
2nd attempt!

### DIFF
--- a/src/query-builder/DeleteQueryBuilder.ts
+++ b/src/query-builder/DeleteQueryBuilder.ts
@@ -29,7 +29,6 @@ export class DeleteQueryBuilder<Entity> extends QueryBuilder<Entity> {
      */
     getQuery(): string {
         let sql = this.createDeleteExpression();
-        sql += this.createWhereExpression();
         return sql.trim();
     }
 
@@ -172,13 +171,13 @@ export class DeleteQueryBuilder<Entity> extends QueryBuilder<Entity> {
      */
     protected createDeleteExpression() {
         const tableName = this.escape(this.getMainTableName());
+        const whereExpression = this.createWhereExpression();
         if (this.expressionMap.returning !== "" && this.connection.driver instanceof PostgresDriver) {
-            return `DELETE FROM ${tableName} RETURNING ${this.expressionMap.returning}`;
-
+            return `DELETE FROM ${tableName}${whereExpression} RETURNING ${this.expressionMap.returning}`;
         } else if (this.expressionMap.returning !== "" && this.connection.driver instanceof SqlServerDriver) {
-            return `DELETE FROM ${tableName} OUTPUT ${this.expressionMap.returning}`;
+            return `DELETE FROM ${tableName} OUTPUT ${this.expressionMap.returning}${whereExpression}`;
         } else {
-            return `DELETE FROM ${tableName}`; // todo: how do we replace aliases in where to nothing?
+            return `DELETE FROM ${tableName}${whereExpression}`; // todo: how do we replace aliases in where to nothing?
         }
     }
 
--- a/src/query-builder/DeleteQueryBuilder.ts
+++ b/src/query-builder/DeleteQueryBuilder.ts
@@ -150,10 +150,15 @@ export class DeleteQueryBuilder<Entity> extends QueryBuilder<Entity> {
      * Optional returning/output clause.
      */
     returning(returning: string): this {
-        this.expressionMap.returning = returning;
-        return this;
+        if (this.connection.driver instanceof SqlServerDriver || this.connection.driver instanceof PostgresDriver) {
+            this.expressionMap.returning = returning;
+            return this;
+        } else throw new Error(`OUTPUT or RETURNING clause only supported by MS SQLServer or PostgreSQL`);
     }
 
+    /**
+     * Optional returning/output clause.
+     */
     output(output: string): this {
          return this.returning(output);
     }
--- a/src/query-builder/DeleteQueryBuilder.ts
+++ b/src/query-builder/DeleteQueryBuilder.ts
@@ -3,6 +3,8 @@ import {ObjectLiteral} from "../common/ObjectLiteral";
 import {ObjectType} from "../common/ObjectType";
 import {Connection} from "../connection/Connection";
 import {QueryRunner} from "../query-runner/QueryRunner";
+import {SqlServerDriver} from "../driver/sqlserver/SqlServerDriver";
+import {PostgresDriver} from "../driver/postgres/PostgresDriver";
 
 /**
  * Allows to build complex sql queries in a fashion way and execute those queries.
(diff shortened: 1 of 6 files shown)


pleerock:
please provide tests as well

pleerock:
tests are failing

src/query-builder/DeleteQueryBuilder.ts
```diff
@@ -144,16 +146,31 @@ export class DeleteQueryBuilder<Entity> extends QueryBuilder<Entity> {
...
     }
 
+    /**
+     * Optional returning/output clause.
+     */
+    returning(returning: string): this {
+        this.expressionMap.returning = returning;
+        return this;
+    }
+
```
pleerock: probably we can add `output` method as well (to be friendly to sqlserver users) like this:

```ts
 output(output: string): this {
     return this.returning(output);
}
```

test/github-issues/660/issue-660.ts
```diff
@@ -0,0 +1,116 @@
...
+            .where("name = :name", { name: user.name })
+            .returning("DELETED.*")
+            .disableEscaping()
+            .getSql();
+
+        if (connection.driver instanceof SqlServerDriver) {
+            expect(sql).to.equal("DELETE FROM user OUTPUT DELETED.* WHERE name = @0"); }
+        else if (connection.driver instanceof PostgresDriver) {
+            expect(sql).to.equal("DELETE FROM user RETURNING DELETED.* WHERE name = $1"); }
+        // else { // this is arguably an error case, since .returning() is only supported by PostgreSQL and MSSQL
```
pleerock: probably we should check in `returning` method if current driver supports it + we need to test this case as well

test/github-issues/660/issue-660.ts
```diff
@@ -0,0 +1,116 @@
...
+        const sql = connection.createQueryBuilder()
+            .delete()
+            .from(User)
+            .where("name = :name", { name: user.name })
+            .returning("DELETED.*")
+            .disableEscaping()
+            .getSql();
+
+        if (connection.driver instanceof SqlServerDriver) {
+            expect(sql).to.equal("DELETE FROM user OUTPUT DELETED.* WHERE name = @0"); }
```
pleerock: currently CI is not configured to run sql server tests... You are able to run them manually?

test/github-issues/660/issue-660.ts
```diff
@@ -0,0 +1,116 @@
...
+            .set({ name: "Joe Bloggs" })
+            .where("name = :name", { name: user.name })
+            .returning("INSERTED.*")
+            .disableEscaping()
+            .getSql();
+
+        if (connection.driver instanceof SqlServerDriver) {
+            expect(sql).to.equal("UPDATE user SET name = @0 OUTPUT INSERTED.* WHERE name = @1"); }
+        else if (connection.driver instanceof PostgresDriver) {
+            expect(sql).to.equal("UPDATE user SET name = $1 RETURNING INSERTED.* WHERE name = $2"); } // Fudged, until PostgreSQL RETURNING clause can be correctly placed after WHERE clause
```
pleerock: I didn't understand what this comment means?

test/github-issues/660/issue-660.ts
tjme: Initially, I was ONLY able to run SQLServer tests, and I believe the RETURNING/OUTPUT clause is now working correctly for it.

test/github-issues/660/issue-660.ts
tjme: I am now able to run the PostgreSQL test locally, and the INSERT seems to work correctly, but UPDATE does not, because, unlike for SQLServer, PostgreSQL wants the RETURNING clause at the end (e.g. after any WHERE clause), and I'm not sure how to achieve that, and our advice would be greatly appreciated!

test/github-issues/660/issue-660.ts
tjme: More tests now added, but not for PostgreSQL UPDATE or DELETE

src/query-builder/DeleteQueryBuilder.ts
```diff
@@ -144,16 +146,35 @@ export class DeleteQueryBuilder<Entity> extends QueryBuilder<Entity> {
...
     protected createDeleteExpression() {
         const tableName = this.escape(this.getMainTableName());
-        return `DELETE FROM ${tableName}`; // todo: how do we replace aliases in where to nothing?
+        if (this.expressionMap.returning !== "" && this.connection.driver instanceof PostgresDriver) {
+            return `DELETE FROM ${tableName} RETURNING ${this.expressionMap.returning}`;
+
+        } else if (this.expressionMap.returning !== "" && this.connection.driver instanceof SqlServerDriver) {
+            return `DELETE FROM ${tableName} OUTPUT ${this.expressionMap.returning}`;
+        } else {
+            return `DELETE FROM ${tableName}`; // todo: how do we replace aliases in where to nothing?
```
pleerock: What does this todo mean? Aren't they disabled when "disableEscaping" is set to true?

src/query-builder/InsertQueryBuilder.ts
```diff
@@ -23,6 +24,10 @@ export class InsertQueryBuilder<Entity> extends QueryBuilder<Entity> {
         return sql.trim();
     }
 
+    output(output: string): this {
```
pleerock: add method comment as well please

src/query-builder/UpdateQueryBuilder.ts
```diff
@@ -143,6 +148,14 @@ export class UpdateQueryBuilder<Entity> extends QueryBuilder<Entity> {
         return this;
     }
 
+    /**
+     * Optional returning/output clause.
+     */
+    returning(returning: string): this {
+        this.expressionMap.returning = returning;
```
pleerock: it should throw exception here is driver does not support returning

src/query-builder/DeleteQueryBuilder.ts
tjme: Sorry, I didn't change that code, or add the todo.

src/query-builder/InsertQueryBuilder.ts
tjme: Will do

tjme:
Though some tests are now failing, I think the only remaining problem stems from the PostgreSQL RETURNING clause being added before, rather than after the WHERE clause. Could anyone advise how I might address this?

pleerock:
to fix this issue you can merge `getQuery` and `createInsertExpression`/... into `getQuery` method and control order of WHERE expression

tjme:
Thanks, I think it's now all fine (though there's a false negative test result elsewhere).

pleerock:
Thank you for all your hard work! This will be released in next version this week.

PR #697, opened 2017-07-23
Support type-safe returns when using transactions

### LINKED ISSUE #696: Support type-safe returns when using transactions
Currently there is no way to return "type-safe" data from inside of a transaction in a "non-hackish" way. That's because signature of the transaction method always returns ``Promise<any>``. The two possible workaround that I've seen are:
* Assign values to variables in outer scope (use by existing test cases in TypeORM repository, eg. this one: https://github.com/typeorm/typeorm/blob/master/test/functional/transaction/transaction-in-entity-manager/transaction-in-entity-manager.ts#L34)
    * Mentioned snippet:
    ```typescript
    let postId: number|undefined = undefined, categoryId: number|undefined = undefined;

    await connection.manager.transaction(async entityManager => {

        const post = new Post();
        post.title = "Post #1";
        await entityManager.save(post);

        const category = new Category();
        category.name = "Category #1";
        await entityManager.save(category);

        postId = post.id;
        categoryId = category.id;

    });
    ```
    * Obviously, it sucks that we have to allow ``undefined`` being assigned to ``postId`` because it is late-bound by the transaction. We should be able to catch it using ``const``.
* Unsafe typecasting from ``Promise<any>`` to our desired type.
    * Example:  
    ```typescript
    const postId = await (await connection.manager.transaction(async entityManager => {

        const post = new Post();
        post.title = "Post #1";
        await entityManager.save(post);

        return post.id;

    })) as number;
    ```
    * If someone changes what type ``post.id`` uses, this code will silently break, because we are forcing TS to use ``number`` type.

It's kind of strange that ``transaction`` forces to use ``Promise<any>`` when there are generics and type inference working pretty well in TypeScript. We should be able to do something like this:
```typescript
// Type inference should guess what is returned by the arrow function
// If needed, we could also define returning type ourselves
const { postId, categoryId } = await connection.manager.transaction(async entityManager => {

    const post = new Post();
    post.title = "Post #1";
    await entityManager.save(post);

    const category = new Category();
    category.name = "Category #1";
    await entityManager.save(category);

    return {
        postId: post.id,
        categoryId: category.id
    };

});

// Types of "postId" and "categoryId" are known at "compile-time"
```

mdziekon:
Resolves #696 

## Problem:
``manager.transaction`` does not allow to use type-safe returns.

## Solution:
``manager.transaction`` should leverage generics & type inference to allow type-safe transaction returns.

## Notes:
It seems that TS compiler sometimes might shout at us for not using explicit returns (look at the "fixed" testcase in file ``query-builder-locking.ts``). It's probably due to the fact that TypeORM is configured to use ``noImplicitReturns`` flag, and with my fix, compiler has to create a union type (earlier it was always ``any``, so it didn't care much). I hope this won't affect acceptance of my PR, I don't think it's that much of a deal (but well, I might be wrong).

### DIFF
--- a/src/entity-manager/EntityManager.ts
+++ b/src/entity-manager/EntityManager.ts
@@ -84,7 +84,7 @@ export class EntityManager {
      * Wraps given function execution (and all operations made there) in a transaction.
      * All database operations must be executed using provided entity manager.
      */
-    async transaction(runInTransaction: (entityManger: EntityManager) => Promise<any>): Promise<any> {
+    async transaction<T>(runInTransaction: (entityManger: EntityManager) => Promise<T>): Promise<T> {
 
         if (this.connection.driver instanceof MongoDriver)
             throw new Error(`Transactions aren't supported by MongoDB.`);
--- a/test/functional/query-builder/locking/query-builder-locking.ts
+++ b/test/functional/query-builder/locking/query-builder-locking.ts
@@ -275,6 +275,8 @@ describe("query builder > locking", () => {
                         .getOne().should.be.rejectedWith(LockNotSupportedOnGivenDriverError)
                 ]);
             });
+
+        return;
     })));
 
 });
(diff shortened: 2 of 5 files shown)


pleerock:
Thank you!

PR #723, opened 2017-08-01
Escape parent id join condition

eyalhakim:
Before this fix, this exception would be thrown from the db:
" missing FROM-clause entry for table "parentidcolumn_XX"
due to lowercase vs uppercase


eyalhakim:
Hi @pleerock, please have a look at this.

PR #726, opened 2017-08-02
Try finding relations in both child and parent entities

eyalhakim:
Prior to this fix, we assumed that if inheritance exists, relations would only be found on the parent entity.


eyalhakim:
@pleerock Please check this one out as well

PR #727, opened 2017-08-02
First load parent metadata, then create primary columns

eyalhakim:
and only then we can create join columns/foreign keys.

Prior to this fix, creating foreign keys / junction entity metadatas did not work since we didn't yet load join columns.


eyalhakim:
@pleerock another fix

eyalhakim:
Hi @pleerock, would appreciate your help on this one and the other PR's.
We are relying heavily on TypeORM and really need the inheritance feature.

eyalhakim:
@pleerock can you look why this build is failing?
Doesn't seem like something caused by my changes

pleerock:
I'll close this issue since I decided to remove CTI support from typeorm functionality for now.

PR #728, opened 2017-08-02
First try getting property name from referencedColumn

eyalhakim:
as it is not as it is not always equal to the property name itself (like in the case of inheritance)


eyalhakim:
@pleerock and another one. feel like i'm close to getting it to work :]

eyalhakim:
@pleerock I see that i broke something. Do you have an idea how to solve this in another way?

Issue #737, opened 2017-08-06
Both null and undefined find option value should generate IS NULL query


yujiahaol68:
I found that when use find function with option. If I want to find a column's value that is null. I will get the right result. But if `tslint` prefer changing `null` into `undefined`, it won't work great because it didn't map both null and undefined value to generate the `IS NULL` sql query.

It cause a big problem and it will take a long time to debug if my project use `tslint` very late.

NoNameProvided:
This has been discussed multiple times across multiple projects of @pleerock . In short this is the correct behavior.

In JS `null` and `undefined` means different things. A `null` value is an explicit way to express that nothing should be there. In the other hand `undefined` means a non-existing value. With `undefined` TypeORM wouldn't be able to make a difference why it isn't there. You just forgot to put the value there? Or was it intentional? With `undefined` there is no way to tell the difference.

For the above reason it's unlikely that this will be changed. I advice you to turn of that particular TSLint rule.

yujiahaol68:
However some teams have set this `no-null-keyword` rule. Maybe the lib should give the right to the team to choose instead of insisting all the team should obey the rules of lib. What do you think?

pleerock:
@NoNameProvided you are absolutely right. I recommend to turn off that tslint rule as well

binki:
I used to use `null` a lot before I started using TypeScript. With TypeScript, it’s a lot easier to only use `undefined` instead. Using `null` in TypeScript requires checking for more conditions and being more explicit because you can’t take advantage of the default value of variables being `undefined`.

Additionally, TypeScript itself’s own “optional class property” behavior favors `undefined` over `null`. This is because it is easier for an object to end up with a property access returning `undefined` than it is to get `null` as `null` will only be returned if the property is initialized.

Regarding `undefined` as a property value, you already can use `Object.hasOwnProperty.call(o, propertyName)` to determine if the property is defined on an object. If the value shouldn’t be on the object, it can be removed by using `delete` instead of overwriting it with `undefined`.

What I think I care about more than **how** TypeORM chooses to handle `undefined`/`null` is that there is consistent behavior and documentation. If I have time later I may open a new docs issue, but just dumping this here for now. Here is what I am seeing with postgresql backend currently (typeorm-0.2.7):

```ts
const o = await manager.findOne(MyEntity, {
  col1: 'asdf',
  col2: undefined,
  col3: null,
});
```

generates SQL like this:

```sql
[...]

PR #738, opened 2017-08-06
Both null and undefined find option value can generate IS NULL query

yujiahaol68:
#737


pleerock:
Sorry, but I don't think I can accept this PR. `null` and `undefined` are different things and this difference play big role in typeorm. I don't see any reason why `undefined` should lead to `null` in this case.

PR #751, opened 2017-08-09
Fix engine option being ignored

as-com:
`MySqlQueryRunner` references `TableSchema#engine`, but the engine parameter wasn't set when creating tables.


pleerock:
Thank you!

PR #765, opened 2017-08-15
Update ColumnOptions.ts

kherkeladze:
type error if column has 'generated' option


pleerock:
This option was removed in latest version (probably docs aren't updated yet), instead following syntax must be used if you want to use column with generated option:

```ts
@Column()
@Generated()
someId;
```

luchillo17:
If that's the case @kherkeladze should test if that works and then reject this PR?

pleerock:
yeah. He can also make fixes in documentation if he wants

PR #775, opened 2017-08-18
fixes #760, prevent setting serial directly in ColumnOptions

### LINKED ISSUE #760: QueryFailedError: type "serial" does not exist on schema sync
**Problem**
When autosync is enabled the schema is synced correctly on the first run, when the application is restarted it will fail with the following error.

```
QueryFailedError: type "serial" does not exist
```

The following entity was used.
```typeorm
@Entity()
class testEntity {
    @PrimaryGeneratedColumn()
    public sourceInventoryId: number;

    @Column({ type: "serial" })
    public testColumn: number;
}
```
**Possible cause**
It tries to execute the following query `ALTER TABLE "test_entity" ALTER COLUMN "sourceInventoryId" TYPE serial`. This fails because `serial` is not supported in `ALTER COLUMN` statements. 

**Possible solution**
Check the current type of the column, if it is already integer with a sequence don't try to change it. If it already exists, the only solution is to delete the column (don't think this is something we want) or to fail with an exception.

Another solution is not to support `serial` like types on `Column` decorators. This then can be solved by introducing a `GeneratedColumn` decorator similar to `PrimaryGeneratedColumn` (which already does not support a `serial` type).

---
**database:** postgres 9.6
**typeorm version:** 0.1.0-alpha.34
**typescript:** 2.4.2

MaikelH:
This is a fix to prevent #760. 

This will remove the option to set a `serial` directly in the `ColumnOptions`. Serial column can still be used by specifying the `PrimaryGeneratedColumn` or `Generated` decorator.


pleerock:
How do we make primary column with `bigserial` now ? Looks like we dont have such possibility

MaikelH:
You can't. The best way is probably to an extra option to the `Generated()` decorator that specifies the type of the column.

MaikelH:
The problem here is essentially that there is ambiguity about what `serial` actually is. In Postgres it used more like an alias, which causes troubles on `alter` statements like in #760.  

The best way to go forward with this is in my opinion to use the `PrimaryGeneratedColumnType` for that. Don't use `serial` then, but use underlying column types (`int`, `bigint`, etc.). We can use the same columntypes for the `Generated` decorator, so you will be able to use `bigserial` columns. Also this will prevent setting the column type directly.

PR #782, opened 2017-08-20
Fix contributing guidelines url in DEVELOPER.md

luchillo17:
The url pointed to the contributing guidelines in angular repo, changed towards the one in this repo.

### DIFF
--- a/DEVELOPER.md
+++ b/DEVELOPER.md
@@ -8,7 +8,7 @@ This document describes how to set up your development environment and run TypeO
 * [Building](#building)
 * [Running Tests Locally](#running-tests-locally)
 
-See the [contribution guidelines](https://github.com/angular/angular/blob/master/CONTRIBUTING.md)
+See the [contribution guidelines](https://github.com/typeorm/typeorm/blob/master/CONTRIBUTING.md)
 if you'd like to contribute to Angular.
 
 ## Prerequisite Software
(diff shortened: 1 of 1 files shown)


luchillo17:
If this get's reviewed, have you considered start to use the GitFlow workflow for versioning?

Here's a few things it would solve:
- It would keep master at a release version, meaning people could actually install from git with npm in case they wanted to test the next release without waiting for the npm upload.
- It would auto-tag your versions (depending on the tool you use for gitflow, GitKraken works great and when finishing a release it will tag your release/hotfix).
- Tags are important to differentiate where the previous version ends and the next version begins, for example, since there's no tags for alpha versions in this repo, i cannot make a diff between 2 alpha versions to check what changed, and CHANGELOG.md don't have alpha version changes.
- Branching and merging guidelines of GitFlow methodology helps keep versions under control when working with big teams, and this repo surely has many contributors.

Lastly is there any guide on understanding the repo structure, how it works? for me, since i'm starting, it's quite big and dunno where to start.

pleerock:
I would like to start using GitFlow once 0.1.0 released. Can you create a github issue for your requests or remind me once 0.1.0 released?

> Lastly is there any guide on understanding the repo structure, how it works? for me, since i'm starting, it's quite big and dunno where to start.

for now you can simply ask me what you are interested in, but in the future we definitely need explanation about process in typeorm source and internal stuff overview

luchillo17:
@pleerock Thanks, i think i start to get how it works, it sends all metadata to a global variable, so i can get the metadata and make introspection to auto-generate the relations in query builder, why? GraphQL can request a relation if he wants, and it can be recursive, though idk if i can make TypeORM return recursive data like that, for example in a User, Post, Comment, User: User -> Post -> Comments -> User (user that made the comment).

PR #787, opened 2017-08-21
Select only specific columns

### LINKED ISSUE #718: Question: would it be valid to implement an extra option on find/findOne methods?
I mean, currently I have to use query builder to select just a few columns from my table, using the find methods on a specific repository returns all columns.

I think it would be nice to have a key on "options" for specifying the columns that I wish to select. 

Any thoughts?

BrainMaestro:
Hey @pleerock, so I tried to implement #718. It works up to the point where the [raw result is transformed into an entity](https://github.com/typeorm/typeorm/blob/master/src/query-builder/SelectQueryBuilder.ts#L1557). Since the result does not have all the relevant columns, it is not converted into an entity. I'm not sure exactly how to handle that. Thoughts?

Fixes #718

### DIFF
--- a/src/find-options/FindOptionsUtils.ts
+++ b/src/find-options/FindOptionsUtils.ts
@@ -96,8 +96,10 @@ export class FindOptionsUtils {
             return qb;
 
         // apply all options from FindOptions
-        if (options.select)
+        if (options.select) {
             qb.select(options.select.map(selection => qb.alias + "." + selection));
+            qb.partialSelect = true;
+        }
 
         if (options.where)
             this.applyConditions(qb, options.where);
--- a/src/find-options/FindOptionsUtils.ts
+++ b/src/find-options/FindOptionsUtils.ts
@@ -15,6 +15,7 @@ export class FindOptionsUtils {
         const possibleOptions: FindOneOptions<any> = obj;
         return possibleOptions &&
                 (
+                    possibleOptions.select instanceof Array ||
                     possibleOptions.where instanceof Object ||
                     possibleOptions.relations instanceof Array ||
                     possibleOptions.join instanceof Object ||
@@ -29,6 +30,7 @@ export class FindOptionsUtils {
         const possibleOptions: FindManyOptions<any> = obj;
         return possibleOptions &&
                 (
+                    possibleOptions.select instanceof Array ||
                     possibleOptions.where instanceof Object ||
                     possibleOptions.relations instanceof Array ||
                     possibleOptions.join instanceof Object ||
@@ -94,6 +96,9 @@ export class FindOptionsUtils {
             return qb;
 
         // apply all options from FindOptions
+        if (options.select)
+            qb.select(options.select);
+
         if (options.where)
             this.applyConditions(qb, options.where);
 
--- a/src/find-options/FindOptionsUtils.ts
+++ b/src/find-options/FindOptionsUtils.ts
@@ -97,7 +97,7 @@ export class FindOptionsUtils {
 
         // apply all options from FindOptions
         if (options.select)
-            qb.select(options.select);
+            qb.select(options.select.map(selection => qb.alias + "." + selection));
 
         if (options.where)
             this.applyConditions(qb, options.where);
--- a/src/query-builder/SelectQueryBuilder.ts
+++ b/src/query-builder/SelectQueryBuilder.ts
@@ -33,6 +33,8 @@ import {QueryRunner} from "../query-runner/QueryRunner";
  */
 export class SelectQueryBuilder<Entity> extends QueryBuilder<Entity> {
 
+    partialSelect: boolean = false;
+
     // -------------------------------------------------------------------------
     // Public Implemented Methods
     // -------------------------------------------------------------------------
@@ -1403,11 +1405,8 @@ export class SelectQueryBuilder<Entity> extends QueryBuilder<Entity> {
 
     protected buildEscapedEntityColumnSelects(aliasName: string, metadata: EntityMetadata): SelectQuery[] {
         const hasMainAlias = this.expressionMap.selects.some(select => select.selection === aliasName);
-        const hasSelectedColumn = metadata.columns.some(column => {
-            return this.expressionMap.selects.some(select => select.selection === aliasName + "." + column.propertyName);
-        });
 
-        if (!hasMainAlias && hasSelectedColumn) {
+        if (!hasMainAlias && this.partialSelect) {
             metadata.primaryColumns.forEach(column => {
                 const selection = aliasName + "." + column.propertyName;
                 if (!this.expressionMap.selects.some(select => select.selection === selection)) {
--- a/src/query-builder/SelectQueryBuilder.ts
+++ b/src/query-builder/SelectQueryBuilder.ts
@@ -1403,6 +1403,21 @@ export class SelectQueryBuilder<Entity> extends QueryBuilder<Entity> {
 
     protected buildEscapedEntityColumnSelects(aliasName: string, metadata: EntityMetadata): SelectQuery[] {
         const hasMainAlias = this.expressionMap.selects.some(select => select.selection === aliasName);
+        const hasSelectedColumn = metadata.columns.some(column => {
+            return this.expressionMap.selects.some(select => select.selection === aliasName + "." + column.propertyName);
+        });
+
+        if (!hasMainAlias && hasSelectedColumn) {
+            metadata.primaryColumns.forEach(column => {
+                const selection = aliasName + "." + column.propertyName;
+                if (!this.expressionMap.selects.some(select => select.selection === selection)) {
+                    this.expressionMap.selects.push({
+                        selection: selection,
+                        virtual: true,
+                    });
+                }
+            });
+        }
 
         const columns: ColumnMetadata[] = hasMainAlias ? metadata.columns : metadata.columns.filter(column => {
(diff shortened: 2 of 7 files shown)


BrainMaestro:
I would like to implement the second one. However, I'm not sure why but the select I added is not aliasing the columns. It generates the sql **SELECT name, description FROM "photo" "Photo** instead of **SELECT `Photo`.`name` AS `Photo_name`, `Photo`.`description` AS `Photo_description` FROM `photo` `Photo`**. I've tried looking everywhere, and I can't seem to find out where that's happening. It's relevant because the aliased columns are [assumed](https://github.com/typeorm/typeorm/blob/master/src/query-builder/transformer/RawSqlResultsToEntityTransformer.ts#L52) when doing the transform.

pleerock:
ah I know why. You need to select using main alias. Append it. In query builder we do:

```ts
createQueryBuilder("post")
.select("post.name")
.addSelect("post.description")
...
```

Same must be done there - apply main alias before each select user does

pleerock:
I made commit that should fix it, please check that

BrainMaestro:
Thanks! That's exactly what I needed. 

So, I found a way to handle the partial results, but it seems a little hacky to me. It's the simplest one I could think of though. Keeping track of the id and removing it from the entity seemed like it would be more complex. Let me know what you think of everything

pleerock:
I don't think its possible to implement it and have a correct behaviour without selecting entity id. You need to write a more complex test with lot of data (lets say 10 posts and each post has 5 categories). Create `name` column in both post and category and partial select it. And you'll need to make partial loading of post with non partial loaded category (e.g. full category) and load partial post with partial category data.

I think you need to load entity id even user didnt specified them (check in createSelectExpression if entity primary columns are not specified and them if so, maybe with "virtual" flag or some other criteria we need to determine that we should not put it into destination entity). In goruping we still use entity id, but we simply don't write entity id into final entity

BrainMaestro:
Okay, the tricky thing though is how to pass the data around. There's no access to EntityMetadata in the SelectQueryBuilder which would have been useful. Passing the metadata into the FindOptionUtils various apply methods generates too many changes. I'm thinking of doing a separate refactor that might make this easier

BrainMaestro:
I ended up adding virtual selects that are removed after transforming the raw results. I think EntityManager needs a little refactoring though. I had to duplicate a lot there. Also, this way involved passing metadata into FindOptionUtils methods like I mentioned. If you have a better suggestion, I would like to know. 

Also, I rewrote the tests with more data. Let me know if something is not done right or missing

BrainMaestro:
Can you rerun the travis build? Only the `node: 6` failed, and it doesn't even seem related

src/find-options/FindOptionsUtils.ts
```diff
@@ -87,13 +90,22 @@ export class FindOptionsUtils {
...
 
         // if options are not set then simply return query builder. This is made for simplicity of usage.
         if (!options || !this.isFindOneOptions(options))
             return qb;
 
         // apply all options from FindOptions
+        if (options.select) {
+            qb.select(options.select.map(selection => qb.alias + "." + selection));
+
+            metadata.primaryColumns
```
pleerock: its a wrong place for this code. This code should be part of query builder. Most probably in `createSelectExpression` method where you can check if primary columns are not selected and select them just like what you do here. Also add test which do partial selection using query builder only (without using find methods - because find methods are just delegation to query builder)

pleerock: Once you do it you won't need to pass this metadata everywhere.

src/query-builder/SelectQueryBuilder.ts
```diff
@@ -117,35 +117,35 @@ export class SelectQueryBuilder<Entity> extends QueryBuilder<Entity> {
     /**
      * Adds new selection to the SELECT query.
      */
-    addSelect(selection: (qb: SelectQueryBuilder<any>) => SelectQueryBuilder<any>, selectionAliasName?: string): this;
+    addSelect(selection: (qb: SelectQueryBuilder<any>) => SelectQueryBuilder<any>, selectionAliasName?: string, virtual?: boolean): this;
```
pleerock: "virtual" is internal. We should not expose this functionality to the user. Once you move logic about into query builder you won't need to change this method at all

src/query-builder/SelectQueryBuilder.ts
```diff
@@ -1561,9 +1561,19 @@ export class SelectQueryBuilder<Entity> extends QueryBuilder<Entity> {
...
 
+        const virtualColumns = this.expressionMap.selects
+            .filter(select => select.virtual)
+            .map(select => select.selection.replace(this.alias + ".", ""));
+
+        // remove virtual columns from the entity
+        entities = entities.map(entity => {
+            virtualColumns.forEach(virtualColumn => delete entity[virtualColumn]);
+            return entity;
+        });
```
pleerock: I think this logic should be somewhere inside raw result to entity transformer, because here you remove virtuals from top-level entities. Add test like this:

```ts
createQueryBuilder("post")
.select(["post.title", "post.description")
.addSelect(["category.title"])
.leftJoin("post.categories", "category")
.getMany();
```

you should except following data:

```json
[{
   title: "..",
   description: "..",
   categories: [{
      title: "..",
   }]
}]
```

Something like that

src/find-options/FindOptionsUtils.ts
```diff
@@ -94,6 +96,11 @@ export class FindOptionsUtils {
             return qb;
 
         // apply all options from FindOptions
+        if (options.select) {
+            qb.select(options.select.map(selection => qb.alias + "." + selection));
+            qb.partialSelect = true;
```
BrainMaestro: I couldn't figure out a way to differentiate this select and the regular ones that are done with the query builder. Using virtual columns makes other query builder tests to break because a primary column is inserted into the sql. I don't like this, and I'm hoping you can think of something better.

test/functional/repository/find-options/repository-find-options.ts
```diff
@@ -51,4 +52,75 @@ describe("repository > find options", () => {
...
+        })));
+
+        expect(loadedPhotos2).to.have.deep.members(photos.map(photo => ({
+            id: photo.id,
+            name: photo.name,
+            description: photo.description,
+            categories,
+        })));
+
+        // expect(loadedPhotos3).to.have.deep.members(photos.map(photo => ({
```
BrainMaestro: I commented out this test because I'm not sure how to handle this case. If we allow virtual primary columns for this case, then the test passes. But it might be unexpected behaviour for users of querybuilder to have ids inserted into their queries when not specified. I think we should only use the virtual columns when coming from the repository

src/find-options/FindOptionsUtils.ts
pleerock: > because a primary column is inserted into the sql

you need to insert primary columns only if they aren't selected at all (if something is selected, but not primary columns). You don't need this ugly flag. Try to redesign this thing

BrainMaestro: If you check, that's exactly what I'm doing. I'm only selecting the primary columns if they have not already been selected. However it's not enough.

Also, you're coming off as a bit rude you know? I have been courteous and willing to listen to every suggestion.

Jorgevillada:
This is a great feature. Will it be on the next release?

src/find-options/FindOptionsUtils.ts
pleerock: I'm rude? Sorry if I am, I don't want to be so haha. I just wanted to say that this code can be better and must be redesigned (you know it too, right)

pleerock:
@Jorgevillada yes most probably so

src/find-options/FindOptionsUtils.ts
BrainMaestro: Okay that's good. I know the flag isn't right. I definitely do not want it and would prefer if it is removed. I just wanted to demonstrate the issue to you.

If I only check if there have been selects, it breaks the other query builder tests that check the sql output. I'm willing to wait until you're back from vacation since you have a better understanding of the code. There might be something obvious that would solve this that I've missed. 

Enjoy your holiday.

pleerock:
Thank you very much for this PR. I merged it and made small fixes we were talking about in above commit. This feature will be released in `0.1.0-alpha.36`. And sorry for being rude again 😆

pleerock:
Next step for this feature is to:

* add tests for partial selection using query builder instead of find options and make sure ids are automatically selected and they arent stored in entity
* add tests with partial selection of columns inside embed entities
* add tests with partial selection and not selected ids inside embed entity
* add tests for partial selection of joined entities

If you have time and want to contribute feel free please :)

chriszrc:
So if I have an entity like this:

```
@Entity()
@Index(["country","iso3","region","name","pollutant"], { unique: true })
export class ExposureRegion implements Region {

    @PrimaryColumn()
    @Column({ length: 56 })
    country?: string;

    @PrimaryColumn()
    @Column({ length: 6})
    iso3?: string;

    @PrimaryColumn()
    @Column({ length: 12})
    region?: string;

    @PrimaryColumn()
    @Column({ length: 56})
    name?: string;

    @PrimaryColumn()
    @Column({ length: 5})
    pollutant?: string;

}

```

Where the primary key is actually a composite key of all the columns in the table, then there would be no way right now to do a query where you select only one of those columns. Correct?

pleerock:
Correct. Primary keys are very important and lot of operations inside ORM are based on them. So if you don't want to get benefits of orm make partial selections and take a raw sql results instead

PR #795, opened 2017-08-24
Support for injecting repositories into transactions 

19majkel94:
As suggested in:
https://github.com/typeorm/typeorm-typedi-extensions/pull/16#issuecomment-323574728

This PR introduce `@TransactionRepository` parameter decorator which injects repository (or custom repository) into method decorated with `@Transaction` decorator.

Additionally, the dev dependencies has been updated to the newest ones to be able to develop on Node 8.x.


pleerock:
Add docs [here](https://github.com/typeorm/typeorm/blob/master/docs/transactions.md) and [here](https://github.com/typeorm/typeorm/blob/master/docs/decorator-reference.md)

src/metadata-args/TransactionRepositoryMetadataArgs.ts
```diff
@@ -0,0 +1,32 @@
+/**
+ * Used to inject transaction's entity managed into the method wrapped with @Transaction decorator.
```
pleerock: update comment

Issue #772, opened 2017-08-17
Unknown column error when using @RelationCount decorator


ruiming:
I have an `Article` table,  `User` table, and an `Article_Readers_User` table to record each user's behavior to the article.

```typescript
// Article Entity
@Entity()
export class Article {
  @PrimaryGeneratedColumn()
  id: number
  
  // I want to cacluate how many user like this article
  @RelationCount((article: Article) => article.readers, 'readers', qb => qb.andWhere('readers.like=:like', { like: true }))
  like: number
  
  @OneToMany(type => ArticleReadersUser, item => item.article)
  readers?: ArticleReadersUser[]
}

// User Entity
@Entity()
export class User {
  @PrimaryColumn()
  uid: number

  @OneToMany(type => ArticleReadersUser, item => item.user)
  articles?: ArticleReadersUser[]
}

// Article_Readers_User
@Entity()
export class ArticleReadersUser {
  @PrimaryColumn()
  articleId: number

  @PrimaryColumn()
  userUid: number

  @Column({ default: false })
  like: boolean

  @ManyToOne(type => Article)
  @JoinColumn({ name: 'articleId' })
  article: Article

  @ManyToOne(type => User)
  @JoinColumn({ name: 'userUid' })
  user: User
}
```
However, when I query the articles it throw an error:

```typescript
 { QueryFailedError: ER_BAD_FIELD_ERROR: Unknown column 'readers.id' in 'field list'
    at ...
  code: 'ER_BAD_FIELD_ERROR',
  errno: 1054,
  sqlMessage: 'Unknown column \'readers.id\' in \'field list\'',
  sqlState: '42S22',
  index: 0,
  sql: 'SELECT `readers`.`articleId` AS `parentId`, COUNT(`readers`.`id`) AS `cnt` FROM `article_readers_user` `readers` WHERE `readers`.`articleId` IN (99, 100, 101, 106, 107, 109, 117, 119, 120) AND `readers`.`like`=true GROUP BY `readers`.`articleId`',
  name: 'QueryFailedError',
  query: 'SELECT `readers`.`articleId` AS `parentId`, COUNT(`readers`.`id`) AS `cnt` FROM `article_readers_user` `readers` WHERE `readers`.`articleId` IN (?) AND `readers`.`like`=? GROUP BY `readers`.`articleId`',
  parameters: [ [ 99, 100, 101, 106, 107, 109, 117, 119, 120 ], true ] }
```

[...]

PR #808, opened 2017-08-30
fix unknown column error when using relationCount

ruiming:
This pr fix this issue: #772 

I am not familiar with SQL and I do not know whether this change will break something. But this change work for me which solve above issue.

### DIFF
--- a/src/query-builder/relation-count/RelationCountLoader.ts
+++ b/src/query-builder/relation-count/RelationCountLoader.ts
@@ -47,10 +47,10 @@ export class RelationCountLoader {
                     return { relationCountAttribute: relationCountAttr, results: [] };
 
                 // generate query:
-                // SELECT category.post as parentId, COUNT(category.id) AS cnt FROM category category WHERE category.post IN (1, 2) GROUP BY category.post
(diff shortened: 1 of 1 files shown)


pleerock:
Relation count currently is almost completely broken functionality. I don't recommend you to use it. TBH I think to remove this functionality at all because of value, complexity and issues it brings

ruiming:
That is a little pity for me because I think `RelationCount` is very powerful and easy to use.

For example, calculate how many people like&comment this article&comment. Without `RelationCount`, I need to always join it manually which brings more code.

Before this being removed, can this pr be merged first in order to fix that issue? I did not find any other questions when using `RelationCount` except the above issue.

pleerock:
Then can you tell me how exactly this change (counting exact column vs *) fixes your issue? I need to understand what exactly your issue is and if your change really fixed root cause of issue

ruiming:
The problem is described in this issue #772 . 
The point is that the exact column to be counted is not existed in that table in the example that issue had described. So I change it to `count(*)` which can work fine and get the result exactly I want.

pleerock:
Okay I'll merge it. However Im not sure if its correct fix. I think problem is in "referencedColumn". It uses original entity id instead of target entity ids. I think proper fix would be to count there multiple primary keys. And, again, I recommend not to use relation count functionality

Issue #811, opened 2017-08-30
[Bug] Compiling and Package breaks in new clone


luchillo17:
Hi, just cloned the project to test a few changes, but before that i tried compiling, and also packaging and none of those worked, here's the command i ran:
```sh
yarn && yarn setup:config && yarn compile
```

That installs all deps with yarn, create the base `ormconfig.json` and should compile with TSC, however the result is this:
![image](https://user-images.githubusercontent.com/9698639/29883062-adde897c-8d74-11e7-9181-e6396d4dc5d8.png)

I think it's complaining by typings, same happens with `npm run package` equivalent:
![image](https://user-images.githubusercontent.com/9698639/29884047-c4bc46a4-8d77-11e7-9862-21dafa075397.png)

However i've also cloned the package as a SubModule to my project and when i import directly from `src` folder i get many other issues which makes me think i have to wait until i can package.

luchillo17:
Ok so it does package, it just complains for typings, but the thing is that using it from SubModule gives this:
![image](https://user-images.githubusercontent.com/9698639/29885181-86983cc6-8d7b-11e7-9a4a-506d7a794779.png)
It's like if it couldn't detect the parent `pg` install, i'll try with npm link and installing `pg` in the SubModule itself.

luchillo17:
Ok with npm link inside the SubModule works fine, which means i have to keep using `npm package`, at least as SubModule i can commit the specific `commit`, work in changes and test in my project those changes.

pleerock:
@Ostrzy do you think your PR fixes this issue?

Ostrzy:
@pleerock I believe so - I had exactly the same errors regarding `decorator/Index.ts` and `MongoEntityManager.ts`. First commit in my PR fixed them for me.

pleerock:
@luchillo17 can you please check it and close it if you fixed it

luchillo17:
TSC errors are gone, but Package got's a new one:
![image](https://user-images.githubusercontent.com/9698639/30005725-5102d4be-90ae-11e7-8635-3249f2ad9199.png)

luchillo17:
Using `src` directly gives a lot less errors but still some:
![image](https://user-images.githubusercontent.com/9698639/30005765-5e443cc0-90af-11e7-89a3-fc052ca55049.png)

pleerock:
[...]

PR #816, opened 2017-09-03
Fix problem with ConnectionOptionsReader when loading config with entities specified directly

Ostrzy:
Due to incorrect check, `ConnectionOptionsReader` failed when anything but a array of strings was provided to `entities`, `migrations` or `subscribers` connection options. Fix was to simply change `||` to `&&`.

As for commit with types changes it was simply needed for me to start the work on fix. Typescript 2.4.2 doesn't have a problem with what was previously but using 2.5.2 caused compilation to fail.


Ostrzy:
I just realised that the types problem I encountered was the same as in https://github.com/typeorm/typeorm/issues/811.

pleerock:
Thank you!

Issue #548, opened 2017-06-16
make sqlite database work in browser for production


mobidev111:
### a decent cross-platform sql implementation is available on Android, iOS & Windows with a native sqlite plugin called 'cordova-sqlite-storage'
- Cordova is a cross-platform framework covering Android, iOS & Windows
- Sqlite is available via the cordova plugin 'Cordova-sqlite-storage' https://github.com/litehelpers/Cordova-sqlite-storage - 1500+ stars & actively maintained
- it does remove most of the limitations of the native browser databases when looking from a cross-platform perspective: 
  - on iOS: removes need to use deprecated webSQL
  - on iOS & Android: features better performance than native databases
  - on Windows: Provides a working implementation

### targeting the large browser-based communities and extends the reach of typeorm massively: Angular & possibly React
 - AngularJS (Angular 1) with Ionic 1
 - Angular (Angular 2+) with Ionic 2+
 - React?

### proposal: Interface the sqlite database using the 'cordova-sqlite-storage' Cordova plugin

Possible tasks are:

- [ ] make statement: actively maintain browser support within typeorm
  - some of the existing approaches in the wild date a few months back and are not actively maintained
  - to have lasting support the browser we should maintain this as part of the typeorm/typeorm repo

- [ ] consolidate the existing browser approaches in the typeorm/typeorm repo
  - @BradyLiles could you post your excellent work in BradyLiles/ionic-orm as a pull-request against the typeorm/typeorm repo?
  - in case that @BradyLiles doesn't have time - @pleerock could you help in bringing this back into the typeorm/typeorm repo? as mentioned in https://github.com/typeorm/typeorm/issues/62#issuecomment-262148315
  - @sidneycorreia could you post your work (https://github.com/sidneycorreia/ionic-orm) as pull-request against the typeorm/typeorm repo?

- [ ] support sqlite database via cordova-sqlite-storage
  - use the sql interface -> switch the websql-based sql interface of the existing proof-of-concept to sqlite
  - adapt/create new sqlite driver, as discussed in https://github.com/typeorm/typeorm/issues/169#issuecomment-294363884

- [ ] extend example project with browser sqlite example - [typeorm/browser-example](https://github.com/typeorm/browser-example/blob/master/src/index.ts)

- [ ] make support ready for production
[...]

PR #821, opened 2017-09-04
Moving Node.js dependencies into PlatformTools class + Cordova SQLite Driver

daniel-lang:
This pull request adds a driver for cordova's sqlite plugin, since that has to be called differently then the node.js sqlite module. It also moves calls to Node.js modules into the PlatformTools class as instructed by @pleerock in #548.
The BrowserPlatformTools implementation isn't perfect. Especially the `chalk` mock is a work in progress.

The following steps are in my opinion necessary after this PR:
* [ ] Update the [browser-example repo](https://github.com/typeorm/browser-example) and the browser docs (and keep it up to date)
* [ ] Think about a way to include the WebSQL and Cordova-SQLite drivers in the automated tests
* [ ] Update the docs to feature cordova support
* [ ] Enhance and move my [cordova test app](https://github.com/daniel-lang/typeorm-cordova-example) to the typeorm organisation (and keep it up to date)
* [ ] Maybe add docs for other frameworks like Ionic

### DIFF
--- a/src/driver/DriverFactory.ts
+++ b/src/driver/DriverFactory.ts
@@ -4,6 +4,7 @@ import {WebsqlDriver} from "./websql/WebsqlDriver";
 import {SqlServerDriver} from "./sqlserver/SqlServerDriver";
 import {OracleDriver} from "./oracle/OracleDriver";
 import {SqliteDriver} from "./sqlite/SqliteDriver";
+import {CordovaSqliteDriver} from "./cordova-sqlite/CordovaSqliteDriver";
 import {MysqlDriver} from "./mysql/MysqlDriver";
 import {PostgresDriver} from "./postgres/PostgresDriver";
 import {Driver} from "./Driver";
@@ -28,6 +29,8 @@ export class DriverFactory {
                 return new MysqlDriver(connection);
             case "sqlite":
                 return new SqliteDriver(connection);
+            case "cordova-sqlite":
+                return new CordovaSqliteDriver(connection);
             case "oracle":
                 return new OracleDriver(connection);
             case "mssql":
--- a/src/driver/DriverFactory.ts
+++ b/src/driver/DriverFactory.ts
@@ -4,7 +4,7 @@ import {WebsqlDriver} from "./websql/WebsqlDriver";
 import {SqlServerDriver} from "./sqlserver/SqlServerDriver";
 import {OracleDriver} from "./oracle/OracleDriver";
 import {SqliteDriver} from "./sqlite/SqliteDriver";
-import {CordovaSqliteDriver} from "./cordova-sqlite/CordovaSqliteDriver";
+import {CordovaDriver} from "./cordova/CordovaDriver";
 import {MysqlDriver} from "./mysql/MysqlDriver";
 import {PostgresDriver} from "./postgres/PostgresDriver";
 import {Driver} from "./Driver";
@@ -29,8 +29,8 @@ export class DriverFactory {
                 return new MysqlDriver(connection);
             case "sqlite":
                 return new SqliteDriver(connection);
-            case "cordova-sqlite":
-                return new CordovaSqliteDriver(connection);
+            case "cordova":
+                return new CordovaDriver(connection);
             case "oracle":
                 return new OracleDriver(connection);
             case "mssql":
--- a/src/driver/cordova-sqlite/CordovaSqliteConnectionOptions.ts
+++ b/src/driver/cordova-sqlite/CordovaSqliteConnectionOptions.ts
@@ -1,0 +1,22 @@
+import {BaseConnectionOptions} from "../../connection/BaseConnectionOptions";
+
+/**
+ * Sqlite-specific connection options.
+ */
+export interface CordovaSqliteConnectionOptions extends BaseConnectionOptions {
+
+    /**
+     * Database type.
+     */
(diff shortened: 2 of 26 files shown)


src/driver/mongodb/typings.ts
```diff
@@ -1,5 +1,4 @@
-import {EventEmitter} from "events";
-import {Readable, Writable} from "stream";
+import {EventEmitterWrapper, StreamReadableWrapper, StreamWritableWrapper} from "../../platform/PlatformTools";
```
pleerock: do we have alternative to create such "wrappers"? Maybe we simply can define for browser package additional typings, like:

```ts
declare module "stream";
declare module "events";
```

Would this approach work?

src/driver/DriverFactory.ts
```diff
@@ -28,6 +29,8 @@ export class DriverFactory {
                 return new MysqlDriver(connection);
             case "sqlite":
                 return new SqliteDriver(connection);
+            case "cordova-sqlite":
+                return new CordovaSqliteDriver(connection);
```
pleerock: can you briefly explain what are differences between sqlite and Cordova drivers?

pleerock:
Good start, I think you are on the right direction

src/driver/DriverFactory.ts
daniel-lang: There are three main differences in the cordova plugin:
* it has to be loaded differently (see `loadDependencies()` in the Driver class
* it does not support named parameters. Therefore I had to overwrite `insert()` and `parametrize()`
* sql-queries are executed with `databaseConnection.executeSql()` and not `databaseConnection.all()`

Other than that they work identical

src/driver/mongodb/typings.ts
daniel-lang: I will give it a try

src/driver/DriverFactory.ts
pleerock: if it does not support named parameters then how you escape parameters?

pleerock: Also since both websql and cordova plugins have almost same codebase, maybe its good time to make a good abstraction for them? Are you able to do that?

daniel-lang: * parameters are escaped with a ? and have to be in the correct order to mapped correctly
* It looks like WebSQL, Cordova SQLite and SQLite could get an abstraction, right?

src/driver/mongodb/typings.ts
daniel-lang: I don't know where I would define these module so I commited a version where the normal PlatformTools just exports the classes from `stream` and `event` and the browser version an empty classes for it. This would make it less of a change

src/driver/DriverFactory.ts
pleerock: understood. yes they all have only few differences since they all are based on sqlite

daniel-lang: Should I add that to this pull request or open a seperate one?

pleerock: as you wish. We can finish with this PR and then do it, it will be easier to review

src/platform/PlatformTools.ts
```diff
@@ -1,7 +1,11 @@
 import * as path from "path";
 import * as fs from "fs";
 import {highlight, Theme} from "cli-highlight";
-const chalk = require("chalk");
+export {ReadStream} from "fs";
+export {EventEmitter} from "events";
+export {Readable, Writable} from "stream";
```
pleerock: Looks like you made even better solution 👍 💯

pleerock:
Does it 100% work in cordova now? We probably need some tests to prevent broking of what you did...

pleerock:
Also, what about other tasks you listed in PR?

pleerock:
Also, do you have experience in electron? We can make it work in electron as well

src/driver/cordova-sqlite/CordovaSqliteConnectionOptions.ts
```diff
@@ -0,0 +1,22 @@
...
+
+/**
+ * Sqlite-specific connection options.
+ */
+export interface CordovaSqliteConnectionOptions extends BaseConnectionOptions {
+
+    /**
+     * Database type.
+     */
+    readonly type: "cordova-sqlite";
```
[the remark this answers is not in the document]
daniel-lang: will change it

daniel-lang:
This the cordova app I have tested it with: [https://github.com/daniel-lang/typeorm-cordova-example](https://github.com/daniel-lang/typeorm-cordova-example). It creates to tapes with a relationship between them and inserted and selected data. nothing to complicated. It was intended as a prove of concept and an entry point if anyone is interested in using it.

daniel-lang:
About the tasks:
* I can create a PR for the browser-example repo to update it
* I have seen a Dockerfile with android and cordova support. I will look into it and see if automated testing is possible
* I can also create a PR for typeorm.github.io to include it in the docs
* I would wait with adding the cordova example since it will change a bit, I will message you with it is ready
* I'm still trying to get it to work with Ionic and will include it in the typeorm.github.io PR if it works

So I think all of those tasks are follow-ups after this PR.
And I don't have any experience with electron

pleerock:
Thank you very much for your hard work. I'll wait other PRs as well once you'll be able to do them.

PR #830, opened 2017-09-06
SQLite Drivers abstraction + PlatformTools chalk integration

daniel-lang:
This is an attempt at having one base sqlite driver that cordova, sqlite (node) and websql extend. I hope this is what you had in mind.
Unfortunately I added something else that was supposed to go in a PR of its own. It is another attempt to fix the compiler error that are caused by my chalk mock in the BrowserPlatformTools file by moving the calls to chalk into the PlatformTools class. Let me know if you like that approach and whether or not you are happy to have it in this PR.


pleerock:
This is too bulky to review, so I'll merge it by trust, you checked everything, right?

pleerock:
Also is it optimized to make it to work with sqljs as well?

daniel-lang:
Don't merge it yet, i'm removing the chalk thing and I tested cordova and websql not node!

daniel-lang:
This turned into a mess. I will open a new PR

PR #831, opened 2017-09-06
SQLite Drivers abstraction

daniel-lang:
This is an attempt at having one base sqlite driver that cordova, sqlite (node) and websql extend. I hope this is what you had in mind. This is also a preparation for sql.js since it can use the same abstract driver and query runner. But I'm not quiet sure yet how to make sql.js work in the browser and in node (since it supports both), therefore it's not in this PR.

(same as #830 without the BrowserPlatform changes).


pleerock:
Nice, thank you

PR #837, opened 2017-09-07
Fix for multiple Ionic and browser problems

daniel-lang:
This PR contains
* changed to `gulpfile.ts` to create an es5 module and a systemjs version for the browser. it also prepares the content of `build/browser` to be used in the `typeorm-browser` package
* removes a duplicate `Entity` decorator export in `index.ts` that webpack doesn't like
* Replaces `require("mongodb")` with `PlatformTools.load("mongodb")` in places I missed before

The changes to `AdvancedConsoleLogger` and `PlatformTools` aren't a must. It's just a suggestion and would remove my not so pretty chalk mock. If you are not a fan of it I can revert it back to the way it was.


gulpfile.ts
```diff
@@ -229,10 +258,11 @@ export class Gulpfile {
     package() {
         return [
             "clean",
-            ["browserCopySources", "browserCopyMainBrowserFile", "browserCopyPlatformTools"],
-            ["packageCompile", "browserCompile"],
+            ["browserCopySources", "browserCopyPlatformTools", "browserCopyMainBrowserFile"],
```
pleerock: you probably aware that all 3 execute in parallel and order of these does not matter?

src/logger/AdvancedConsoleLogger.ts
```diff
@@ -46,16 +46,16 @@ export class AdvancedConsoleLogger implements Logger {
      */
     logQuerySlow(time: number, query: string, parameters?: any[], queryRunner?: QueryRunner) {
         const sql = query + (parameters && parameters.length ? " -- PARAMETERS: " + this.stringifyParams(parameters) : "");
-        console.log(chalk.underline.yellow(`query is slow:`) + " " + chalk.bold(PlatformTools.highlightSql(sql)));
-        console.log(chalk.underline.yellow(`execution time:`), time);
+        PlatformTools.logWarn(`query is slow:`, PlatformTools.highlightSql(sql, true));
```
pleerock: told on sql does not work (it highlights only first word of sql query), so you can remove it.

pleerock: just what I did yesterday with `query failed` error message - I removed bold from there

pleerock:
What do you think if we simply remove all drivers (except websql, sqljs) for browser package? And probably stub them somehow elegant? Same with advanced console logger, maybe remove that too? And do same for ionic/cordova?

gulpfile.ts
daniel-lang: Yes I am aware of this, this change comes because I had a version that wouldn't need the `browserCopyMainBrowserFile` so I removed it and later added it back in but at a different place in the list. I will revert that change

daniel-lang:
It would make life easier I guess but I'm not quiet sure how that would be managable with all the imports.

daniel-lang:
Should I create an issue for your idea to seperate the browser package so we can brainstorm there?

pleerock:
as you wish - you can create a new issue or pr or even we can continue discussion here

Issue #838, opened 2017-09-08
Time zones for timestamp columns are incorrectly fetched and persisted in PostgreSQL


slavafomin:
Hello!

I've encountered a strange behavior with `timestamp with time zone` columns in PostgreSQL. It looks like such columns are incorrectly fetched and persisted. I'm going to push the test to demonstrate the problem.

slavafomin:
When working through TypeORM entities this problem will not surface, because it looks like it has a symmetrical properties (i.e. two problems negate each other). But this becomes apparent when you insert data to the underlying database directly, e.g. through `psql`. The timestamps returned from TypeORM are actually off by some offset. You could also insert timestamp with `manager.save()` and then query the row directly with `pgsql` to see that persisted value is incorrect as well.

pleerock:
Can you provide a short examples of what you have, what you do, what is wrong and what is expected result? Looks like test you have added is passing

pleerock:
Ah, no. It says:

```ts
  1) github issues > #838 Time zones for timestamp columns are incorrectly fetched and persisted in PostgreSQL "before all" hook:
     DataTypeNotSupportedError: Data type "timestamp with time zone" is not supported in "mysql" database.
```

for some reason CI shows it as passing 😨

pleerock:
You need to enable only posgtres by using `enabledDrivers: ["postgres"]`

slavafomin:
The test should be able to explain the problem, but I will also add a step-by-step reproduction scenario:

1. Insert the timestamp `"1989-08-16T10:00:00+03:00"` to the PostgreSQL column with type `timestamp with time zone` directly through, e.g. `psql`.

2. Fetch the entity using ORM and examine the timestamp. It would be off by amount of hours in server's time zone offset.

---

1. Insert the timestamp `"1989-08-16T10:00:00+03:00"` to the PostgreSQL column with type `timestamp with time zone` using ORM.

2. Fetch the timestamp directly through, e.g. `psql` and examine it. Timestamp would be off by amount of hours in server's time zone offset.

--

I've also updated the test as you've suggested.

slavafomin:
Hello @pleerock! Has you been able to look into this? I think that severity of this problem is very high, because it could lead to a potential data corruption and business logic issues. I've implemented a workaround for our application, but I would be glad to assist in fixing it in the source.

robinsummerhill:
I've looked into this problem and can also reproduce the problem seen by @slavafomin.

The problem is this:

[...]

PR #839, opened 2017-09-08
Added test for issue #838

slavafomin:


pleerock:
Let me merge this test and take a look on it deeply.

pleerock:
What is proper syntax for inserting date time with timezone in postgres? When I do `INSERT INTO "flight" ("id", "date") VALUES (4, '1989-08-16T10:00:00+03:00');`, I have following record in database:

```
4	1989-08-16 14:00:00.000000 +07:00
```

which does not look correct

pleerock:
If you simply do:

```ts

        await postgresConnection.query(`INSERT INTO "flight" ("id", "date") VALUES (1, '1989-08-16 14:00:00.000000 +03:00');`);
        const results = await postgresConnection.query(`SELECT date FROM "flight" WHERE id = 1`);
```

It will give you following:

```
executing query:  INSERT INTO "flight" ("id", "date") VALUES (1, '1989-08-16 14:00:00.000000 +03:00');
executing query:  SELECT date FROM "flight" WHERE id = 1
[ anonymous { date: 1989-08-16T11:00:00.000Z } ]
```

TypeORM DOES NOT DO ANYTHING HERE, 0 ACTIONS, it just delegates query further to pg driver. It means problem is on another layer or there is an issue on how we insert the data into the database.

robinsummerhill:
@pleerock Your insert/select above is showing expected behaviour. Postgres will convert a datetime to UTC when persisting a value to a `timestamp with time zone` field. Any client that displays it (e.g. psql, pgadmin etc) will convert it back to localtime using the current timezone.

It is unfortunate that the name 'timestamp with timezone' is misleading. Postgres **does not** store the timezone at all - it just does a conversion to UTC before persisting it. That's **all** it does.

You input ('1989-08-16 14:00:00.000000 +03:00') and output ('1989-08-16T11:00:00.000Z') are completely equivalent and point to the exact same moment in time.

pleerock:
@robinsummerhill thank you for your explanation, so postgres converts to zero utc whatever date without whatever utc I pass to it

Issue #834, opened 2017-09-06
Schema update does not work for indexes when changing fields or uniqueness


iz-iznogood:
Hi,

schema update for indexes only works when the index name changes.
It does not work for field changes (adding, removing or changing the order of fields) or changing the uniqueness of the index.

It would be nice if this included those cases too

pleerock:
Released in `0.1.0-alpha.40`

Issue #833, opened 2017-09-06
Unique Index Schema update issue


iz-iznogood:
Adding a Unique index using the Index decorator works great
If you remove the decorator then the index is not dropped.

The problem happens when the decorator is on the entity or the column(s) and it only happens for unique indexes. Non unique are dropped fine
I use MSSQL and I traced the issue on this query 

`SELECT TABLE_NAME = t.name, INDEX_NAME = ind.name, IndexId = ind.index_id, ColumnId = ic.index_column_id, COLUMN_NAME = col.name, ind.*, ic.*, col.* FROM sys.indexes ind INNER JOIN sys.index_columns ic ON ind.object_id = ic.object_id and ind.index_id = ic.index_id INNER JOIN sys.columns col ON ic.object_id = col.object_id and ic.column_id = col.column_id INNER JOIN sys.tables t ON ind.object_id = t.object_id WHERE ind.is_primary_key = 0 AND ind.is_unique = 0 AND ind.is_unique_constraint = 0 AND t.is_ms_shipped = 0 ORDER BY t.name, ind.name, ind.index_id, ic.index_column_id`

that excludes unique indexes.

pleerock:
Released in `0.1.0-alpha.40`

Issue #770, opened 2017-08-16
Adding "unique: true" to existing @Column yields constraint with wrong name, throwing "relation ... already exists" on every restart


avbentem:
A mismatch in the naming of `UNIQUE` constraints between newly created tables, and tables that have `unique: true` added to new or existing `@Column` definitions at a later time, causes errors like:

    error: relation "uk_my_column" already exists

When using `autoSchemaSync: true` in typeorm 0.1.0-alpha.34 along with pg 7.1.2 towards PostgreSQL 9.6.3 on x86_64-pc-linux-gnu, compiled by gcc (Debian 4.9.2-10) 4.9.2, 64-bit:

1. All is fine when creating a new table with a column which has `unique: true` specified for the very first run. For that, the following is used in the `CREATE TABLE` statement:

    ```sql
    CREATE TABLE ... CONSTRAINT "uk_my_table_my_column" UNIQUE ("my_column")
    ```

2. However, when **adding `unique: true` to an existing column**, TypeORM adds a constraint with a name that lacks the table name. The `ALTER TABLE ... ADD` uses:

    ```sql
    ALTER TABLE ... ADD CONSTRAINT "uk_my_column" UNIQUE ("my_column")
    ```
    This works fine initially (the constraint is added), but fails for subsequent restarts for which TypeORM tries to add the very same constraint again and again. So:

    **The erroneous shorter name seems to be the culprit, as TypeORM probably expects the long name when trying to determine if the unique constraint already exists**. 

3. And when **adding a column to an existing table and immediately specifying `unique: true` for that new column**, then the `UNIQUE` constraint is not added immediately. Instead, TypeORM adds the column on the first restart, and only adds the constraint on the second restart, with the wrong shorter name just like above, so then also failing for subsequent restarts.

I guess I could create a PR for (2), which would also partly fix (3), but the latter should obviously add the new constraint in the first `ALTER TABLE ... ADD COLUMN`.

- This might also be the cause for problems with `unique: true` in `@Index`, like reported in #721, but I am not sure. 
- This might also be the cause for #719, though that (erroneously?) refers to foreign keys. (And uses `schema:sync`, but that might work the same as `autoSchemaSync: true`.)

More details below, but I guess the points above cover it all.

## Adding `unique: true` to existing `@Column`

With:

```typescript
import {Column, PrimaryGeneratedColumn} from 'typeorm';
import {Entity} from 'typeorm/decorator/entity/Entity';

@Entity('my_table')
[...]

PR #843, opened 2017-09-08
Fixes for Issues #833 and #834 (Schema updates for indexes)

iz-iznogood:
Implement Issue #833 by changing the SQL Server metadata query to return the whether an index is unique  or not. 
Implement issue #834 by extending the drop index logic to include the changes in columns - uniqueness so that the index is recreated properly


iz-iznogood:
Please review and provide feedback

Thanks

src/schema-builder/RdbmsSchemaBuilder.ts
```diff
@@ -357,7 +357,20 @@ export class RdbmsSchemaBuilder implements SchemaBuilder {
 
             // drop all indices that exist in the table, but does not exist in the given composite indices
             const dropQueries = tableSchema.indices
-                .filter(indexSchema => !metadata.indices.find(indexMetadata => indexMetadata.name === indexSchema.name))
+                .filter(indexSchema => {
+                    var metadataIndex = metadata.indices.find(indexMetadata => indexMetadata.name === indexSchema.name);
```
pleerock: don't use `var` - use `const` instead. Also take a look, you have tslint errors

src/schema-builder/RdbmsSchemaBuilder.ts
```diff
@@ -357,7 +357,20 @@ export class RdbmsSchemaBuilder implements SchemaBuilder {
...
+                    if (!metadataIndex)
+                        return true;
+                    if (metadataIndex.isUnique !== indexSchema.isUnique)
+                        return true;
+                    if (metadataIndex.columns.length !== indexSchema.columnNames.length)
+                        return true;
+                    for (let i = 0; i < metadataIndex.columns.length; ++i)
```
pleerock: can you please make this code more functional? (without `for` loop)

pleerock:
Please fix tslint issues, merge master into your branch (to get latest fixes to failing tests)

pleerock:
We also need to make this to work on all platforms, not only mssql. Also would be great if we write tests for this functionality

src/schema-builder/RdbmsSchemaBuilder.ts
ckalogeropoulos: Done

src/schema-builder/RdbmsSchemaBuilder.ts
ckalogeropoulos: Done

ckalogeropoulos:
Hi again, I fixed the review comments and I will send you a pull request soon.
However when testing it against MySQL, and Postgres I found several other issues in other areas.
In Postgres the indexes where updated fine but there is some issue with the sequences, the system is always trying to recreate them and fails. This is not related to this code
In MySQL the problems are worse, it tries to recreate the entire table in every run even there are no changes in the configuration. This is also unrelated to this code.
I could not test SQLite or Oracle since for SQLite I could not install the driver (some compile issues and I have no Oracle installed).
I will try to investigate the issues on Postgres and MySQL and fix them if possible.

iz-iznogood:
Hi, please pull

pleerock:
To make this PR complete we need to fix issues in mysql, postgres and sqlite. And add tests that confirms new fixes. Oracle can be skipped for now.

test/functional/database-schema/indices/indices-create-modify.ts
```diff
@@ -0,0 +1,97 @@
...
+                expect(tableSchema!.indices.length).to.be.equal(1);
+                expect(tableSchema!.indices[0].name).to.be.equal("IDX_TEST"); 
+                expect(tableSchema!.indices[0].isUnique).to.be.false; 
+                expect(tableSchema!.indices[0].columnNames.length).to.be.equal(2); 
+                expect(tableSchema!.indices[0].columnNames[0]).to.be.equal("firstname"); 
+                expect(tableSchema!.indices[0].columnNames[1]).to.be.equal("lastname");
+    
+            };
+
+            const checkUniquenessFn = async function () {
```
pleerock: why did you make such structure? If you want separate blocks then better to create multiple `it` code blocks

iz-iznogood: Done

iz-iznogood:
Hi, I have made some more fixes that I found while looking at #770.
This pull request is needed so that more fixes in other areas can come in (fix+test for #770).

Please review and merge if Ok. 
Thanks

iz-iznogood:
Hi, this is getting a bit messy, I will provide a clean pull request with all the necessary commits.
Sorry for this

PR #844, opened 2017-09-08
Clarifying Migration Executer Error Msg

puttputt:
"Migration class name should contain a class name at the end of the file. ${migrationClassName} migration name is wrong."

Was receiving this error and was not sure what it meant. I discovered that the migration class name requires a timestamp, but the error thrown did not describe this properly.

${migrationClassName} migration name is wrong. Migration class name should have a UNIX timestamp appended.


pleerock:
Looks like it was really wrong message. Thank you for your PR!

Issue #721, opened 2017-07-31
IndexOption unique:true causing unhandledPromiseRejection MSSQL


DavidPMacNeil:
Hello, If I add or modify an index using TypeORM alpha 31, and MSSQL to use `{ unique: true}`, I will get an unhandled promise rejection warning. 

So if I have:
```
@Entity()
@Index('IDX_UNIQUE_CATS', (cat: Cat) => [cat.CatID, cat.Sequence], { unique: true })
class Cat {
    ...
}
```

I will receive this error:
```
(node:3764) UnhandledPromiseRejectionWarning: UnhandledPromiseRejection (rejection id: 1): RequestError: The operation failed because an index or statistics with name 'IDX_UNIQUE_CATS' already exists on table 'Cat'.
```

Thanks for the time.

avbentem:
Same with PostgreSQL, which would throw something like `error: relation "uk_my_column_name" already exists`. 

typeorm 0.1.0-alpha.34 along with pg 7.1.2 towards PostgreSQL 9.6.3 on x86_64-pc-linux-gnu, compiled by gcc (Debian 4.9.2-10) 4.9.2, 64-bit

MaikelH:
Are both of you using the `autoSchemaSync` option?

avbentem:
Yes, I am using `autoSchemaSync: true`, which otherwise works great. (Like everything in TypeORM works great!) And I am sorry for not seeing this earlier:

1. All is fine when creating a new table with a column which has `unique: true` specified for the very first run. For that, the following is used in the `CREATE TABLE` statement:

    ```sql
    CONSTRAINT "uk_my_table_2_my_column_name" UNIQUE ("my_column_name")
    ```

2. However, when **adding `unique: true` to an existing column**, TypeORM adds a constraint with a name that lacks the table name. The `ALTER TABLE ... ADD` uses:

    ```sql
    CONSTRAINT "uk_my_column_name" UNIQUE ("my_column_name")
    ```
    This works fine initially, but fails for subsequent restarts for which TypeORM tries to add the very same constraint again. So, this shorter name seems to be the culprit.

3. And when **adding a column to an existing table and immediately specifying `unique: true` for that new column**, then the `UNIQUE` constraint is not added immediately. Instead, TypeORM only adds the constraint on the next restart, just like above, so then also fails for subsequent restarts.

[...]

Issue #840, opened 2017-09-08
EntitySchemas does not support indexes


iz-iznogood:
Hi,

I am trying to use typeorm with entity schemas for the entity definitions and I found that it is not supported. It would be great to add it there too

pleerock:
Released in `0.1.0-alpha.40`

PR #852, opened 2017-09-11
Fix for issue #770

iz-iznogood:
Fix unique constraint names in Postgres and add a test


pleerock:
Thank you for your hard work. I have released all your changes in `0.1.0-alpha.39`. What issues must be closed? #770 , #833, #834 and #840 ? Anything else?

iz-iznogood:
Thank you, it was a pleasure working with your platform, you have setup a great infrastructure and library.
I believe it also closes #721 as it is related to these fixes and the tests I added cover this case too.

pleerock:
Thank you very much

Issue #824, opened 2017-09-05
[Bug] Wrong lazy relation value as parameter on ManyToOne relations


luchillo17:
So i'm debugging why it uses the primary key value of the `Many` side instead of the one in the relation defined by `@JoinColumn`.

Here's how it looks:
```ts
@Entity('a_asset', {
  skipSync: true,
  orderBy: {
    id: 'DESC',
  },
} as any)
export class Asset {
  @PrimaryColumn('int', {
    name: 'a_asset_id',
  })
  id: number

  // Relations
  @ManyToOne(
    (type) => Organization,
    (organization) => organization.assets,
  )
  @JoinColumn({
    name: 'ad_org_id',
  })
  organization: Promise<Organization>
}
```
And the organization one:
```ts
@Entity('ad_org', {
  skipSync: true,
  orderBy: {
    id: 'DESC',
  },
} as any)
export class Organization {
  @PrimaryColumn('int', {
    name: 'ad_org_id',
  })
  id: number

  // Relations
  @OneToMany(
    (type) => Asset,
    (asset) => asset.organization,
  )
  assets: Promise<Asset[]>
}
```
It seems to me that the value of `@JoinColumn` property is not being saved to the `Asset` result object before using it for the `Lazy-Relation`, so the Lazy relation by default tries with the `primaryKey`.

makamekm:
What version do you use?

luchillo17:
Directly from GitHub.

luchillo17:
Ok so i'm logging the entity before it get's passed to the `LazyRelationsWrapper` and this shows:
```ts
let result = [ Asset {
    id: '1015712',
    name: '582ADJ',
    value: '582ADJ',
    versionno: '2018',
    organizationId: '0' } ]
```
As you can see there's `organizationId`, however when i go to the wrapper, when i go into calling the `getEntityValue` function on the join column, i see this:
![image](https://user-images.githubusercontent.com/9698639/30169108-0c43bc7a-93b1-11e7-9af0-58a4a238b9ee.png)
Take note that there's an `entityMetadata`, but no `embeddedMetadata` nor `relationMetadata` which are the ones the `getEntityValue` asks for, in such case it goes directly for the `id` field of entity instead of the relation one.

luchillo17:
Check https://github.com/typeorm/typeorm/issues/826 to see my situation regarding testing, skipping schema tests meanwhile.

[...]

PR #853, opened 2017-09-11
Extend Lazy Relation test entities

luchillo17:
Extending the Lazy-Relation testing entities with named tables, named PK columns, and a combination of both generates a few issues with MySql databases.


test/functional/lazy-relations/entity/Category.ts
```diff
@@ -24,4 +30,193 @@ export class Category {
...
+    onePostNamedAll: Promise<PostNamedAll>;
+
+}
+
+@Entity("s_category", {
+    orderBy: {
+        id: "ASC",
+    }
+})
+export class CategoryNamedTable {
```
pleerock: Dont put two entities in same class. Bad practice.

pleerock:
It is hell to work with such entity. Lets start from scratch and create a separate test case for each issue you have. Create a new test with minimal reproduction problem of first issue you have. You create a broken test with minimal reproduction code -> I fix it. Then you create a second broken test with minimal reproduction code -> I fix it. And each test in separate directory, lets put all tests related to lazy relations inside `test/functional/relation/lazy/XXX` directory, okay? For example:

* `test/functional/relation/lazy/lazy-with-custom-table-name/...`
* `test/functional/relation/lazy/lazy-with-custom-join-column-name/...`

okay? This way we will fix all issues efficiently and will keep better maintainability of the codebase.

test/functional/lazy-relations/entity/Category.ts
luchillo17: What do you mean 2 entities in same class? did you meant 2 entities in same file?

pleerock: yes sorry, I meant 2 entities in same file

luchillo17:
@pleerock i thought `test/functional/lazy-relations` was the right place for all `LazyRelations` tests, and as for using the same file, i wanted to make a test that covered all posible combinations:
+ Normal entity - normal entity.
+ Normal - Named Table.
+ Normal - Named Column.
+ Normal - Named Table & Column.
+ Named Table - Normal.
+ Named Table - Named Table.
+ ...

test/functional/lazy-relations/entity/Category.ts
luchillo17: Check 3rd comment on why i tried to put them there.

pleerock:
It could be right place if it was a small change, but you added too much and that test become a maintenance hell. To make it easier to work with this test and prevent maintenance hell we need to do what I described above.

luchillo17:
However as i told you that's not the test yet, those are only the entities for the tests to come.

pleerock:
Normal entity - Normal entity.
Normal - Named Table.
Normal - Named Column.
Normal - Named Table & Column.
Named Table - Normal.
Named Table - Named Table.


=>>

test/functional/lazy-relations/
* lazy-relations-named-table
* lazy-relations-named-columns

Do you really think you should create normal from one then from another side as separate tests? Maybe use named tables in ` lazy-relations-named-table` in all its entities? Same with `lazy-relations-named-columns` - use named columns in all its entities?

luchillo17:
Well originally i wanted each Post entities to have relations to each Category entities to ensure the test covers all possible combination, including owner & not owner sides.

I thought it was necessary because the names of the tables and PK columns are often taken in account for when making CRUD operations. 

So you suggest i should use NamedTable with NamedTable and that should just cover those cases?

pleerock:
If you think its enough then yes lets create less cases. If you think its not covering all then create multiple directories and you can name directories this way: lazy-relations-named-table-case1, lazy-relations-named-table-case2

luchillo17:
Ok, i'll try with less cases and then test inside my project, and if issues persists, i'll extend to multiple cases combining owning side of relations and all that, however could you be more precise where would i put the tests? like folder structure and such? i see in the original you have an `entity` folder.

pleerock:
inner structure is same in all tests. just put your test with its entities inside `test/functional/relation/lazy/lazy-relations-named-table` directory as we decided, same for others

luchillo17:
Btw `case1`, `case2`... aren't very descriptive, i've had a hard time with other tests with such naming, had to get inside each case and understand what such test case is about, don't you think?

pleerock:
right but what is an alternative?

luchillo17:
Well, since we already inside `lazy` folder, we already know it's about lazy relations, just describe the actual case like `lazy/named-table-to-named-column` or `named-column-to-named-both` or something in this specific case.

luchillo17:
@pleerock i've come to notice that i had to clean up the build folder for it to pick up the change in folder structure, after that seems like the `json` files for `schema` aren't being copied to the build folder, shouldn't they be imported at the top of the file just like the entities? or am i thinking too much in the Webpack way?

pleerock:
yeah its a problem, but I dont copy things instead just change path to schemas. Take a look how I implemented it in other places where entity schema is used. Its not a perfect solution but works for now

pleerock:
also I dont think you should use entity schema in your tests. No need to complicate things, use regular entities

luchillo17:
I just tried to keep the normal one just like you left it, but now i understand why that many `..` back's in schema paths.

luchillo17:
@pleerock I've done a last push and the only thing left is fixing the issue with named columns, seems named tables work ok but named columns break in both tests that involves such feature.

luchillo17:
My bad, it seems it lacks a few more things, i didn't noticed until now that in the `describe`'s `before` function of each `lazy` test there's an `enabledDrivers` array to tell the test in which DB's can it test, i've changed to this for now:
```ts
        entities: [
            Post,
            Category,
        ],
        entitySchemas: [ userSchema, profileSchema ],
        schemaCreate: true,
        dropSchema: true,
        enabledDrivers: ["mysql", "postgres"] // we can properly test lazy-relations only on one platform
    }));
```
With that and skipping MySql from the `ormconfig.json` file, i get the same errors i have in my app, let me setup that and make a new PR.

luchillo17:
@pleerock Dude, i think i understand the issue with `lazy-relations`, with my last commit the issue is reduced to getting the wrong relation id, take as example this image where i have above the green line the queries of the normal relation tests, while those under the green line are the queries and error of the new tests i made commit in this PR:

image

In the upper part those SQL queries relates to the successful ones from the basic lazy relations, you know, the ones you had before i added more tests, please note the `WHERE` clause on the last one, where it uses `"Post"."id" = $1"`, since in basic relations tests both have a basic id field, there's no harm there and works ok.

However in the lower half, it represents the named PK tests i made, here the `Post` entity uses a field named `s_post_id` and the `Category` entity uses a field named `s_category_id`, you can notice this time they are different, and in this case the code is picking the wrong PK name for the entity over which it is going to base the `WHERE` clause.

The code that relates to this issue is this: [src/lazy-loading/LazyRelationsWrapper.ts#L108-L111](https://github.com/luchillo17/typeorm/blob/feature/extend-lazy-relation-testing/src/lazy-loading/LazyRelationsWrapper.ts#L108-L111)

For now i've managed to fix this with the following code, however i think this doesn't play nice with the entities that have more than 1 PK, after all i'm accessing the first (index 0) of the entity's `primaryColumns` property:
```ts
joinColumns.forEach(joinColumn => {
  
  qb.andWhere(`${qb.escape(relation.entityMetadata.name)}.${qb.escape(relation.entityMetadata.primaryColumns[0].databaseName)} = :${joinColumn.referencedColumn!.databaseName}`)
        .setParameter(`${joinColumn.referencedColumn!.databaseName}`, joinColumn.referencedColumn!.getEntityValue(entity));
  });
```

luchillo17:
@pleerock Here you go, this should fix the `loadManyToOneOrOneToOneOwner` relations, since the testing for the other relations is ok i think it's finally fixed.

If tests still fails, they aren't from mine but from other tests that has nothing to do with `Lazy-Relations` so this PR should be able to be merged into master.

pleerock:
I've fixed failing tests on master, please merge master and re-run your tests

src/lazy-loading/LazyRelationsWrapper.ts
```diff
@@ -105,9 +106,9 @@ export class LazyRelationsWrapper {
             .from(relation.type, relation.propertyName) // Category, category
             .innerJoin(relation.entityMetadata.target as Function, relation.entityMetadata.name, conditions);
 
-        joinColumns.forEach(joinColumn => {
-            qb.andWhere(`${relation.entityMetadata.name}.${joinColumn.referencedColumn!.databaseName} = :${joinColumn.referencedColumn!.databaseName}`)
-                .setParameter(`${joinColumn.referencedColumn!.databaseName}`, joinColumn.referencedColumn!.getEntityValue(entity));
+        primaryColumns.forEach(primaryColumn => {
+            qb.andWhere(`${qb.escape(relation.entityMetadata.name)}.${qb.escape(primaryColumn.databaseName)} = :${primaryColumn.databaseName}`)
+                .setParameter(`${primaryColumn.databaseName}`, primaryColumn.getEntityValue(entity));
```
pleerock: Can you replace:

```ts
qb.andWhere(`${qb.escape(relation.entityMetadata.name)}.${qb.escape(primaryColumn.databaseName)} = :${primaryColumn.databaseName}`)
    .setParameter(`${primaryColumn.databaseName}`, primaryColumn.getEntityValue(entity));
```

to

```ts

qb.andWhere(`${relation.entityMetadata.name}.${primaryColumn.propertyPath} = :${primaryColumn.propertyName}`)          
.setParameter(`${primaryColumn.propertyName}`, primaryColumn.getEntityValue(entity));
```

It must give same result as right now and automatically escape column names

luchillo17:
Let me try 1 time after merge, then i'll try your suggestion.

luchillo17:
Done, ready to merge

pleerock:
Nice, merged

luchillo17:
How fast does it make it to npm repo? i'm tired of executing cli commands from `node_modules/typeorm/cli.js`, and with this do you consider the feat in beta state or released?

pleerock:
No, I can release it right now

luchillo17:
Nice, now how was it to close the issue that references this PR?

pleerock:
released in `typeorm@0.1.0-alpha.43`. Sorry I did not get what do you mean

luchillo17:
Ah in some issue trackers that have integrations with version control, you can tell it to generate a branch off  the issue, then when finished tell it to make the PR and when accepted it closses the issue with a message that references the PR, i've just done that manually for #824.

Issue #855, opened 2017-09-11
Postgres primary key issue


iz-iznogood:
Hello,

if my primary key column is not named 'id' then the created sequence in the postgres db has a hardcoded dependency on the column name (it is created as 'nextval("tableName"_id_seq:reqclass'
This makes the column marked as not generated and in subsequent restarts it tries to recreate it

PR #856, opened 2017-09-11
Fix for issue #855

iz-iznogood:
This adds a fix for #855 by changing the sequence naming code.
Also add test


iz-iznogood:
Hi, please merge this if you agree. It fixes #855

test/functional/database-schema/sequences/entity/Person.ts
```diff
@@ -0,0 +1,16 @@
...
+@Entity()
+export class Person {
+
+    @PrimaryGeneratedColumn()
+    Id: number;
+
+    @Column({
+        unique: true
+    })
+    Name: string;
```
pleerock: can you please use camelCase in property names in this, others and future models, its its more natural to use camelCase in js world.

pleerock:
Looks like a proper fix, thank you

ckalogeropoulos:
Since I am looking at schema update and metadata, can you please close the issues (or I can do this if you want) so that I can check what remains that I can fix ? Also for the previous commits - merges

Thank you

test/functional/database-schema/sequences/entity/Person.ts
iz-iznogood: Sure, I did it this way since this is how I discovered the issue. So instead of using different database names in the decorator, I chose to take the class as is in my code

PR #860, opened 2017-09-11
Remove deprecated imports for generated migrations

lukaszklis:
Spotted two unnecessary imports (according to the latest changelog) in a migration I’ve just generated.


pleerock:
Thank you!

PR #862, opened 2017-09-11
@AbstractEntity missing in documenation.

puttputt:
I'm not sure where you want the updated documentation but I noticed this missing in the docs.

### DIFF
--- a/docs/entities.md
+++ b/docs/entities.md
@@ -615,6 +615,7 @@ To reduce duplication and produce a better abstraction we can create a base clas
 
 
 ```typescript
+@AbstractEntity
 export abstract class Content {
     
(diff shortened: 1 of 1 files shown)


pleerock:
This is deprecated decorator, and its not required to use this decorator on inherited entities anymore - you can inherit any class you want and parent class will contain all columns and relations from childs

puttputt:
Really? I am using 0.0.11 and this wasn't the case. Or is it fixed in upcoming versions?

puttputt:
Oh. Nevermind I see the 0.1 changelog. Thanks.

pleerock:
I recommend you to upgrade to 0.1.0

PR #868, opened 2017-09-12
fix bug when apply `skip` and `take` option to findAndCount(fix #867)

### LINKED ISSUE #867: result of `findAndCount` is wrong when apply `skip` and `take` option
bellow test case failed

````js
        const userRepository = connection.getRepository(User);
        const users = new Array(5).fill(0).map((n, i) => {
            const user = new User();
            user.username = `User_${i}`;
            return user;
        });
        await userRepository.save(users);
        const [ foundUsers, totalCount ] = await userRepository.findAndCount({
            skip: 1,
            take: 2
        });
        expect(foundUsers).to.have.lengthOf(2); 
````

the output is:

`AssertionError: expected [ Array(5) ] to have a length of 2 but got 5`

YuhangGe:
see #867 

then function `applyOptionsToQueryBuilder` will be called by both `applyFindOneOptionsOrConditionsToQueryBuilder` and `applyFindManyOptionsOrConditionsToQueryBuilder`，but it only check `isFindOneOptions` in source file `src/find-options/FindOptionsUtils.ts` at line 95

### DIFF
--- a/src/find-options/FindOptionsUtils.ts
+++ b/src/find-options/FindOptionsUtils.ts
@@ -92,7 +92,7 @@ export class FindOptionsUtils {
     static applyOptionsToQueryBuilder<T>(qb: SelectQueryBuilder<T>, options: FindOneOptions<T>|FindManyOptions<T>|undefined): SelectQueryBuilder<T> {
 
         // if options are not set then simply return query builder. This is made for simplicity of usage.
-        if (!options || !this.isFindOneOptions(options))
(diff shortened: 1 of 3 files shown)


pleerock:
One your test has failed because of chaotic insertion order. Can you please insert user one-by-one or simply add order by to find method

pleerock:
I added fix I was talking about. Thank you for PR and sorry again for misunderstanding you.

YuhangGe:
@pleerock never mind. My english is poor, so it's just okey~

Issue #857, opened 2017-09-11
[Discussion] The future of the browser package


daniel-lang:
This discussion started in PR #837 

@pleerock: "What do you think if we simply remove all drivers (except websql, sqljs) for browser package? And probably stub them somehow elegant? Same with advanced console logger, maybe remove that too? And do same for ionic/cordova?"

It would be great to have a package that includes the SystemJS and TypeScript version so that it can be either used in a SystemJS project or retranspiled for a different module loader
I like the idea from @pleerock to only include the needed drivers (websql, cordova, sqljs - at some point) but I'm not sure who replacing or stubbing them would work.
The way it is done right now with the PlatformTools class is fine for that one file but is going to be a nightmare to maintain if it is done for multiple files/all drivers. Same would apply to creating a reperate repository.

Is there a better way to maybe generate mocks automatically?

daniel-lang:
@pleerock How hard would it be to upload the `build/browser` folder as it is for now? Would be great since the `typeorm-browser` package is on version `0.0.6`

pleerock:
I can do it easily, however did you check it? I'll publish contents of `build/browser` directory, however what I see it right now it looks like this:

```
- build/browser
cache
common
connection
decorator
driver
entity-manager
entity-schema
error
find-options
lazy-loading
logger
metadata
metadata-args
metadata-builder
migration
naming-strategy
persistence
platform
query-builder
query-runner
repository
schema-builder
subscriber
util
container.ts
index.ts
package.json
README.md
typeorm-browser.js
typeorm-class-transformer-shim.js
typeorm-model-shim.js
```

However what I made last time it was looking this way:

```
- typeorm-browser
package.json
README.md
typeorm-browser.d.ts
typeorm-browser.js
typeorm-browser.js.map
typeorm-browser.min.js
```

So my question is why did you change it this way, I don't think its correct

daniel-lang:
That's what I tried to explain. For projects that don't use SystemJS `typeorm-browser` is unusable right now, since it only includes the one compiled version. Ionic for example uses webpack and therefore needs every TypeScript file compiled seperate.
I added the other folders to make the library usable in Ionic and other projects that don't use SystemJS to load modules

pleerock:
Okay, does it work with both systemjs and webpack properly now? Also, why do you pack source ts files instead of .d.ts files ?

daniel-lang:
[...]

PR #871, opened 2017-09-12
Browser Package: Adds cli.js and fixes wrong naming for the SystemJS build

daniel-lang:
As mentioned in #857 the `cli.js` file was missing from the package, which npm complains about. For consistency, I also renamed the internal name for the SystemJS build since otherwise the package would be called `typeorm-browser` but in the code it would still be `import { Entity } from "typeorm"` instead of `import { Entity } from "typeorm-browser"`. Is this a welcome change?

### DIFF
--- a/gulpfile.ts
+++ b/gulpfile.ts
@@ -105,7 +105,7 @@ export class Gulpfile {
         return [
             tsResult.js
                 .pipe(sourcemaps.write(".", { sourceRoot: "", includeContent: true }))
-                .pipe(gulp.dest("./build/browser"))
+                .pipe(gulp.dest("./build/package"))
         ];
     }
 
@@ -121,10 +121,10 @@ export class Gulpfile {
             .pipe(tsProject());
 
         return [
-            tsResult.dts.pipe(gulp.dest("./build/browser")),
+            tsResult.dts.pipe(gulp.dest("./build/package/browser")),
             tsResult.js
                 .pipe(sourcemaps.write(".", { sourceRoot: "", includeContent: true }))
-                .pipe(gulp.dest("./build/browser"))
+                .pipe(gulp.dest("./build/package/browser"))
         ];
     }
 
@@ -136,21 +136,15 @@ export class Gulpfile {
         return gulp.src("./build/browser/typeorm-browser.js")
             .pipe(uglify())
             .pipe(rename("typeorm-browser.min.js"))
-            .pipe(gulp.dest("./build/browser"));
+            .pipe(gulp.dest("./build/package"));
     }
 
     @Task()
     browserClearPackageDirectory(cb: Function) {
         return del([
             "./build/systemjs/**",
-             "./build/browser/src/**"
+             "./build/browser/**"
         ])
-    }
-
-    @Task()
-    browserCopyCliFile() {
-        return gulp.src("./build/package/cli.js")
-            .pipe(gulp.dest("./build/browser"))
     }
 
     // -------------------------------------------------------------------------
@@ -234,9 +228,7 @@ export class Gulpfile {
     packagePreparePackageFile() {
         return gulp.src("./package.json")
             .pipe(replace("\"private\": true,", "\"private\": false,"))
-            .pipe(gulp.dest("./build/package"))
-            .pipe(replace("\"name\": \"typeorm\",", "\"name\": \"typeorm-browser\","))
-            .pipe(gulp.dest("./build/browser"));
+            .pipe(gulp.dest("./build/package"));
     }
 
     /**
@@ -246,8 +238,7 @@ export class Gulpfile {
     packageCopyReadme() {
         return gulp.src("./README.md")
             .pipe(replace(/```typescript([\s\S]*?)```/g, "```javascript$1```"))
-            .pipe(gulp.dest("./build/package"))
-            .pipe(gulp.dest("./build/browser"));
+            .pipe(gulp.dest("./build/package"));
     }
 
     /**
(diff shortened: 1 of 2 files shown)


pleerock:
I thought it is already `import { Entity } from "typeorm"`, isn't it? [See here](https://github.com/typeorm/browser-example/tree/master/src).

daniel-lang:
It is right now, but this PR would change that so that it would be `import { Entity } from "typeorm-browser"` which would line up with the package name.

pleerock:
No, I would totally like to avoid it.

daniel-lang:
Okay so it should stay `import { Entity } from "typeorm"` for systemjs?

pleerock:
Yes I would like it to stay consistent everywhere. Plus it allows to reuse models between backend and frontend (not something usual, but anyway nice to have feature)

daniel-lang:
Okay I can change it back for the systemjs version but webpack uses the folder name of the package so there it will always be `import { Entity } from "typeorm-browser"`

pleerock:
Thats a problem as well. I don't really like it. Can we change it somehow?

daniel-lang:
> Plus it allows to reuse models between backend and frontend (not something usual, but anyway nice to have feature)
Not sure if that is possible since the PlatformTools are the BrowserPlatformTools in the `typeorm-browser` package

daniel-lang:
> Can we change it somehow?

As far as I know, it looks at the folder name in `node_modules` which is `typeorm-browser`

pleerock:
Maybe some webpack option? Probably something like `map` ?

daniel-lang:
Another option would be to include a browser folder with the normal package and do `import { Entity } from "typeorm/browser"`. Not sure if that's any better and would even work.

pleerock:
> Not sure if that is possible since the PlatformTools are the BrowserPlatformTools in the typeorm-browser package

No, for example if you have a backend with `User` entity and lets say you have a frontend where you use `User` model as well. You can share such entities between frontend and backend, no matter what is inside typeorm package, because all what is used are decorators from "typeorm" namespace

pleerock:
I think its possible to setup webpack, can you please research this question?

daniel-lang:
So it is possible with two modifications
1. for the typescript compiler by adding (tsconfig.json)
```
"baseUrl": ".",
"paths": {
    "typeorm": ["node_modules/typeorm-browser"]
 }
```

2. for webpack (package.json)
```
"browser": {
    "typeorm": "typeorm-browser"
}
```

After that I can write `import { Entity } from "typeorm-browser"` or `import { Entity } from "typeorm"`, since both will work

daniel-lang:
The same changes to `tsconfig.json` have to be done for systemjs as well since the typescript compiler only looks at the folder name

pleerock:
> import { Entity } from "typeorm-browser" or import { Entity } from "typeorm"

I don't think its a good idea to use or even give opportunity to use both ways. We need to keep only one way (preferable second as I told). What step is required to do only `import { Entity } from "typeorm"`? Is it possible to configure only webpack and avoid configuring tsconfig?

daniel-lang:
One way would be to create a factory which loads the right PlatformTools class depending on config or even driver. Than everything could work with one package and the  `import { Entity } from "typeorm"` syntax

daniel-lang:
Not sure, if that would work for the compiler

daniel-lang:
> Is it possible to configure only webpack and avoid configuring tsconfig?

No, since no matter which loader is used (webpack, systemjs or something completely different) the typescript files have to be transpiled to javascript and therefore tsconfig has to be configured so it knows where to look for the package

pleerock:
So what is the conclusion on how it shall work? As I understood we cannot easily make user to use `import { Entity } from "typeorm"` imports. Thus users must use `import { Entity } from "typeorm-browser"` syntax. If for some reason they would want to use `import { Entity } from "typeorm"` syntax instead, they configure their webpack/systemjs and tsconfig. Right?

daniel-lang:
No, I found a way to use `import { Entity } from "typeorm"` everwhere and only have one package for browser and Node.js. I'm going to update the PR in a second

daniel-lang:
So the conclusion is
* use `import { Entity } from "typeorm"` everywhere
* use one package `typeorm` for node and browser which works by including both versions in the same package
  * drawback: the package gets bigger and has a lot of duplicate code
  * advantage: one universal import syntax, easier to maintain
  * webpack looks at the `browser` tag in `package.json` and uses that path if it exists so when using typeorm with webpack it automatically chooses the right version with any config by the user

pleerock:
Solution looks good, are you sure this change does not affect anything in both nodejs and browser usages? Is compiler good in both cases?

pleerock:
Would be great if you update browser example (also we need to create both cordova and ionic example repositories in typeorm organization)

daniel-lang:
I tested the ionic app (webpack) and the modified browser example (systemjs) with this solution and both versions compile and run with no problem. Other than the automated tests, I didn't test the nodejs module but that should not be affected by this.
As soon as a working version is out I will create a PR for the browser example. It is ready but for now I waited for a working package.
How would you handle the cordova and ionic example? Should I create them and you reupload them to the organisation so I can keep them up to date by creating PRs for them?

pleerock:
> How would you handle the cordova and ionic example? Should I create them and you reupload them to the organisation so I can keep them up to date by creating PRs for them?

I think if you create repos we can move them to typeorm organization, even if no we'll figure out something. We probably need two repositories: "ionic-example" and "cordova-example". Both must follow same code style as "browser-example" and contain a very simple and minimal code for users to understand and work with.

If you are planning to continue active contribution on typeorm I'll add you to organization and you'll be able to create repositories on your own, just let me know.

pleerock:
I published a new `0.1.0-alpha.42` version with your changes

daniel-lang:
Thanks!

I can definitely create example repos. The one for cordova already exists and just needs to be updated. And for ionic I can use the app I'm working on as a template.
As of right now I going to continue contributing to typeorm, at least to the browser specific part, maybe also to the rest if I have to time.

Issue #403, opened 2017-04-13
Will there be auto-migrations generator like in DoctrineMigrations?


idchlife:
Hi! Right now without auto-migrations I'm wondering, is there easy way to just collect SQL from syncing and just save it into file?

pleerock:
This is must-have feature and I plan to do it. However right now there so many other issues needs to be resolved, my hands just didnt touch this part yet.

idchlife:
@pleerock got it, thanks!

pleerock:
basic generation functionality is landed since `0.1.0-alpha.17`. It has some features missing plus it do not generate sql for `down` method yet. Start testing it 🏭

GuillaumeNury:
@pleerock Awesome work !
I tried `migrations:generate` in `0.1.0-alpha-19`. Migration script contains all what is needed to update/add/remove columns, but he does not have anything about creating missing table.

vance:
Any good way to configure it with naming strategy?

pleerock:
@GuillaumeNury hmm thats strange, are you sure about it? It should create schema for new entities as well. Maybe you forgot to specify your entities in orm configuration?

@vance what do you mean?

ddyrcz:
Is there any instruction how to use the `migrations:generate` command?

yoz:
@vance maybe see #1482?

AlexMesser:
fixed and released in `typeorm@0.2.0-alpha.23`. You can try it via `npm i typeorm@next`.

LogansUA:
@pleerock Sorry for popping this issue up but question is still remain for me.

Can we somehow collect all synching queries and write them to some file?

For now the only way I see to write custom logger and define method `logQuery` but filter those which are `SELECT/DELETE/UPDATE`. Is there better way to do it?

pleerock:
yes we can, there is a migration:generate command that you can use, see CLI documentation.

PR #875, opened 2017-09-12
add initial support for postgres migration generation

llwt:
I wasn't sure how to test this, happy to add some if you've got an example or idea of how to add them. I worked off of the implementation in the MySqlQueryRunner for most of this but didn't see any coverage added when that functionality was added.

Relates to #403.

### DIFF
--- a/src/driver/postgres/PostgresQueryRunner.ts
+++ b/src/driver/postgres/PostgresQueryRunner.ts
@@ -84,7 +84,7 @@ export class PostgresQueryRunner implements QueryRunner {
     /**
      * Sql-s stored if "sql in memory" mode is enabled.
      */
-    protected sqlsInMemory: string[] = [];
+    protected sqlsInMemory: (string|{ up: string, down: string })[] = [];
 
     /**
      * Mode in which query runner executes.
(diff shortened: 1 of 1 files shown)


pleerock:
Hi and thank you for your work. I think before I can merge this PR you'll need to complete all todos you left.

Regarding to tests, I didn't think how to test it yet... Probably we need to make a change, check if change in in the database, then revert it and check if change was reverted.

llwt:
The TODOs here were just explicitly listing the down steps that weren't implemented in the migrations so they are called out and could be manually implemented. The forward steps should all be there.

In the MysqlQueryRunner there are places where you are currently not passing through the `schemaQuery` function and hitting the DB which seems undesired. I opted for the DB comment down script with a `TODO` statement instead.

For example in [MysqlQueryRunner #L566](https://github.com/typeorm/typeorm/blob/master/src/driver/mysql/MysqlQueryRunner.ts#L566):
```typescript
        if (newColumn.isUnique === false && oldColumn.isUnique === true)
            await this.query(`ALTER TABLE \`${tableSchema.name}\` DROP INDEX \`${oldColumn.name}\``); // todo: add revert code
```

I'm happy to implement the down steps when I get time, I just figured I'd push up the "up" logic since it seemed like it might be useful to others even without fully implemented "down" generation.

pleerock:
I'll merge it and I hope you'll get some time to finish all todos.

llwt:
Thanks! Sorry I haven't been able to fully flush this out. It's on the top of my list of TODOs, migration support is super helpful for our workflow.

Issue #790, opened 2017-08-22
[Postgres] Fatal error during schema sync


bcherny:
I had an entity with the following column:

```ts
@Column()
@Index()
foo: string
```

I then deleted that column and replaced it with:

```ts
@ManyToOne(() => Foo, foo => foo.bars)
foo: Foo
```

Now I'm seeing this error on startup (using 0.1.0-alpha.35):

```
0|index    | dropping an index: ind_8531cdf4c5900186c2a8a5cea4
0|index    | executing query: DROP INDEX "ind_8531cdf4c5900186c2a8a5cea4"
0|index    | query failed: DROP INDEX "ind_8531cdf4c5900186c2a8a5cea4"
0|index    | error: { error: index "ind_8531cdf4c5900186c2a8a5cea4" does not exist
0|index    |     at Connection.parseE (/Users/bcherny/Sites/mousepig/server/node_modules/pg/lib/connection.js:546:11)
0|index    |     at Connection.parseMessage (/Users/bcherny/Sites/mousepig/server/node_modules/pg/lib/connection.js:371:19)
0|index    |     at Socket.<anonymous> (/Users/bcherny/Sites/mousepig/server/node_modules/pg/lib/connection.js:114:22)
0|index    |     at emitOne (events.js:115:13)
0|index    |     at Socket.emit (events.js:210:7)
0|index    |     at addChunk (_stream_readable.js:252:12)
0|index    |     at readableAddChunk (_stream_readable.js:239:11)
0|index    |     at Socket.Readable.push (_stream_readable.js:197:10)
0|index    |     at TCP.onread (net.js:589:20)
0|index    |   name: 'error',
0|index    |   length: 124,
0|index    |   severity: 'ERROR',
0|index    |   code: '42704',
0|index    |   detail: undefined,
0|index    |   hint: undefined,
0|index    |   position: undefined,
0|index    |   internalPosition: undefined,
0|index    |   internalQuery: undefined,
0|index    |   where: undefined,
0|index    |   schema: undefined,
0|index    |   table: undefined,
0|index    |   column: undefined,
0|index    |   dataType: undefined,
0|index    |   constraint: undefined,
0|index    |   file: 'tablecmds.c',
0|index    |   line: '762',
0|index    |   routine: 'DropErrorMsgNonExistent' }
0|index    | executing query: ROLLBACK
0|index    | You have triggered an unhandledRejection, you may have forgotten to catch a Promise rejection:
[...]

Issue #197, opened 2017-01-13
[Postgres] Fails to drop indexes when removing fields


lktvlm:
Version typeorm@0.0.7-alpha.7

I had a class with column
`@Column({type: "text", nullable: true})`
`@Index()`
`text2: string;`
Removing this column leads to error while deleting index:
- <...>
- executing query: SELECT t.relname AS table_name, i.relname AS index_name, a.attname AS column_name  FROM pg_class t, pg_class i, pg_index ix, pg_attribute a
- WHERE t.oid = ix.indrelid AND i.oid = ix.indexrelid AND a.attrelid = t.oid
- AND a.attnum = ANY(ix.indkey) AND t.relkind = 'r' AND t.relname IN ('type_test_post') ORDER BY t.relname, i.relname
- <...>
> Result - index exists:
> type_test_post	ind_5701659c6d9617e3cb8100b9eb4	likesCount
> type_test_post	ind_9c4cdb13eec7ef29ee990c9dafd	text3
> type_test_post	ind_d8ff91f7221b6df7c5b684ffa9c	title
> type_test_post	ind_ebfa1f9180c7d7b54861cc1afc2	text2
> type_test_post	my_index_with_id_and_title	title
> type_test_post	my_index_with_id_and_title	id
> type_test_post	type_test_post_pkey	id
- columns dropped in type_test_post: text3
- **executing query: ALTER TABLE "type_test_post" DROP "text3"**
- dropping an index: ind_9c4cdb13eec7ef29ee990c9dafd
- executing query: DROP INDEX "ind_9c4cdb13eec7ef29ee990c9dafd"
- **query failed: DROP INDEX "ind_9c4cdb13eec7ef29ee990c9dafd"**
- executing query: ROLLBACK
- Cannot connect:  { error: **index "ind_9c4cdb13eec7ef29ee990c9dafd" does not exist**

It probably should drop indexes **first**, and only then drop the column.

lktvlm:
I tried to make a clean test with a new class and a new column, but then the order was correct.
Probably there is a possibility of both a correct order and an incorrect order in the code depending on something.

pleerock:
yeah catching bug in schema sync is difficult because process is different each time. I need more clear steps how to reproduce this issue from the beginning, e.g. schema create first time, then "x" change and it works, then "y" change and it breaks.

lktvlm:
Here you go.
First create a column, sync:
>@Column({nullable: true})
>text6: string;
[...]

PR #878, opened 2017-09-13
Issue #197 and #790

iz-iznogood:
This is a fix for issues #197 and #790, now column removal and indices works well.

Please pull

### DIFF
--- a/src/schema-builder/RdbmsSchemaBuilder.ts
+++ b/src/schema-builder/RdbmsSchemaBuilder.ts
@@ -6,7 +6,6 @@ import {IndexSchema} from "./schema/IndexSchema";
 import {QueryRunner} from "../query-runner/QueryRunner";
 import {PrimaryKeySchema} from "./schema/PrimaryKeySchema";
 import {ColumnMetadata} from "../metadata/ColumnMetadata";
-import {IndexMetadata} from "../metadata/IndexMetadata";
 import {EntityMetadata} from "../metadata/EntityMetadata";
 import {PromiseUtils} from "../util/PromiseUtils";
 import {Connection} from "../connection/Connection";
@@ -407,24 +406,13 @@ export class RdbmsSchemaBuilder implements SchemaBuilder {
      */
     protected async dropColumnReferencedIndices(tableName: string, columnName: string): Promise<void> {
 
-        const allIndexMetadatas = this.connection.entityMetadatas.reduce(
-            (all, metadata) => all.concat(metadata.indices),
-            [] as IndexMetadata[]
-        );
-
         const tableSchema = this.tableSchemas.find(table => table.name === tableName);
         if (!tableSchema)
             return;
 
         // find depend indices to drop them
-        const dependIndices = allIndexMetadatas.filter(indexMetadata => {
-            return indexMetadata.tableName === tableName && !!indexMetadata.columns.find(column => column.databaseName === columnName);
-        });
-        if (!dependIndices.length)
-            return;
-
         const dependIndicesInTable = tableSchema.indices.filter(indexSchema => {
-            return !!dependIndices.find(indexMetadata => indexSchema.name === indexMetadata.name);
+            return indexSchema.tableName === tableName && !!indexSchema.columnNames.find(columnDatabaseName => columnDatabaseName === columnName);
         });
         if (dependIndicesInTable.length === 0)
             return;
--- a/src/driver/postgres/PostgresQueryRunner.ts
+++ b/src/driver/postgres/PostgresQueryRunner.ts
@@ -434,7 +434,7 @@ where constraint_type = 'PRIMARY KEY' AND c.table_schema = '${this.schemaName}' 
                     const columnNames = columnPositions
                         .map((pos: number) => dbIndicesInfos.find(idx => idx.attnum === pos)!["column_name"]);
 
-                    return new IndexSchema(dbTable["TABLE_NAME"], dbIndexName, columnNames, dbIndicesInfos[0]["is_unique"]);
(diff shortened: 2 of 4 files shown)


src/schema-builder/RdbmsSchemaBuilder.ts
```diff
@@ -407,24 +406,13 @@ export class RdbmsSchemaBuilder implements SchemaBuilder {
...
-            (all, metadata) => all.concat(metadata.indices),
-            [] as IndexMetadata[]
-        );
-
         const tableSchema = this.tableSchemas.find(table => table.name === tableName);
         if (!tableSchema)
             return;
 
         // find depend indices to drop them
-        const dependIndices = allIndexMetadatas.filter(indexMetadata => {
```
pleerock: can you please explain this removal?

iz-iznogood: Yes sorry, it was in the commit message. I am putting it here as well

We change the logic for finding the indices on dropped columns. Instead of searching in the metadata we just look in the retrieved schema.

We change this logic for 2 reasons:

1. If the user of TORM drops a column then it is his responsibility to maintain the rest of the schema, removing all references of this column from everywhere. TORM has a validator for that thus it catches such errors on startup. In any case the previous code was leaving the dropped index inside the metadata but was removing it from DB which is bad.

2. If the index is removed from the metadata, then all checks where failing since the allIndexMetadata filtering was not returning anything

pleerock: Im sorry but I didnt not understand both reasons.

> previous code was leaving the dropped index inside the metadata but was removing it from DB which is bad.

what does it mean? How it can leave dropped index inside metadata? If user removes index from entity it will be dropped from the database. If user added index to entity it will be created.


> If the user of TORM drops a column then it is his responsibility to maintain the rest of the schema, removing all references of this column from everywhere.

We should help him everywhere we really can to maintain his schema because maintaining schema is a pain and we are trying to abstract people from this process.

In our case I think we can remove indices if he removed them from entity and change or add them when he does it, right?

iz-iznogood: Hi again, please let me clarify. 

Let's say the user wants to remove a column. In this case it is normal to assume that he will also remove all references to this column (indices, foreign keys etc) from the rest of the metadata. If not there should be proper validations. This works perfectly, the validations are there.

But the code I removed implies that although the column is removed, the reference of the column remains inside the metadata and it searches for the such a reference in the indices (of the metadata) when it should be instead searching for such a reference in the indices of the db schema (which is not removed yet). 

Since the reference of the column is not inside the metadata - the user has also removed it (otherwise the schema validations that run early would have caught such bad configuration), this search fails so and the system can not find what indices needs to be removed.

Regarding the second point, you do this already. You have the validations and it is perfect.

iz-iznogood: There is only one hole in this system that I see and it looked like this.
The user has a table where a column is referenced by an index (as an example).

The user removes the column from the metadata at runtime (removing it from the metadata array) and forgets to remove the reference in the index (or just remove the entire index). This breaks at runtime, but I do not consider this a proper way to do it. We can catch this if we run the validations early every time the schema is updating, not just on create connection. This is bullet proof.

Or you can change the entire metadata structure and instead of providing direct reference to the Columns array, wrap it in an ColumnList class, and implement add and remove methods.
At remove you can also remove all references of this column from other schema objects.

That's what i meant by assisting the user on schema changes but I believe that it's the wrong thing to do. 

Hope this clears things a bit

Thanks

pleerock:
okay, lets merge it

pleerock:
can #197 and #790 be closed now?

pleerock:
for some reason tests started to fail most of times. I added mssql to docker and now CI runs mssql tests as well and after that it started to fail. I can't find why.

iz-iznogood:
Yes both issues can be closed.

Regarding CI is there a specific test that fails ? Or it is random ?

pleerock:
See fails [here](https://travis-ci.org/typeorm/typeorm/branches). Sometimes it fails because of timeout without running any tests, last time it also failed few times on some test related to unique indices

iz-iznogood:
Ok can you please open an issue regarding mssql tests so that we can continue there ?

Also I am considering about running the metadata validations every time synchronize is executed (early) so that we can avoid any unpleasantries. Do you agree, so that I can provide a pull request ?

pleerock:
I disabled mssql tests on travis because I couldn't find source of issues.

>  running the metadata validations every time synchronize

metadata validation is one time operation because metadata is not changed anywhere in the app, why do you need to revalidate again things that are not changed anywhere in the app?

iz-iznogood:
As I wrote a few comments before : 

> There is only one hole in this system that I see and it looked like this.
The user has a table where a column is referenced by an index (as an example).

>The user removes the column from the metadata at runtime (removing it from the metadata array) and forgets to remove the reference in the index (or just remove the entire index). Then he runs the schema synchronization. This breaks at runtime, but I do not consider this a proper way to do it. We can catch this if we run the validations early every time the schema is updating, not just on create connection. This is bullet proof.

Since the metadata data structures are not read only, someone can remove columns at runtime and run sync again. If you believe that I am paranoid :laughing: then ok. But code wise I believe we can add this validation.
It's your call anyways, AFAIC I am fully covered

pleerock:
> Since the metadata data structures are not read only, someone can remove columns at runtime and run sync again

at runtime isnt an option. Lot of things can be done in runtime.

iz-iznogood:
Ok then we close this. Thanks !!!!

Issue #861, opened 2017-09-11
normalizeType function issue


iz-iznogood:
Hi,

I checked on how typeorm decides on the proper datatype for common javascript types
For example if column.type == String then it translates to nvarchar (mssql) or character varying (postgres) etc
This is good when we work with entities for metadata. But creates an issue regarding entity schemas. If we use json for that then the comparison fails since column.type == "string" is false etc.
I see that for some drivers (and for Number datatype) you have expander the check to include this case
e.g. sqlite 
`        if (column.type === Number || column.type === "int") {
            return "integer";
`
mysql
`        if (column.type === Number || column.type === "integer") {
            return "int";
`

I propose that there should be some string aliases for the common types to allow the json case to work
For example "string", "integer", "boolean", "date" with case insensitive comparison.
I believe it is mandatory to have real cross database support, unless I miss something

If you agree I can send some patches and tests for this change.

Thanks

pleerock:
Actually `column.type === Number` is needed because `Number` here is reflection of property type, e.g.:

```ts
@Column()
sum: number;
```

This `number` is translated into JavaScript `Number` type which then we translate into concrete types. 

So, what Im thinking is that json does not need those abstract types, and users can set exact type they really need, like varchar, int, text, etc.

iz-iznogood:
Hi,

I understand your POV but I beg to differ. What I propose is a simple enhancement of what already exist. 
All normalizeType methods have an else clause so that anyone can pass whatever specific type he wants. So this case is covered in all scenarios (decorators, js objects, json files). But in the later case we miss the generic functionality that is present in all other cases.
For instance I wanted to declare a boolean datatype. In the ts classes case I had just to add the field and the ORM converted it to bit column for mssql, etc This did not work for the json objects, I had to set it to bit manually, thus killing the abstraction. 
Also most users will want to use the same schema for different connections (imagine also the replication case) so common type aliases for at least string, boolean, integer and date are (IMHO) mandatory.
I could also add float, double, datetime etc but lets stay for the common js datatypes.

So in code instead of 
`       
 } else if (column.type === String) {
return "nvarchar";
` 
I propose

`
        } else if (column.type === String || (typeof(column.type) === "string" && column.type.toLowerCase() === "string") {
            return "nvarchar";
`
same for other types. This does not break anything and is symmetrical to all cases

What do you say ?

pleerock:
[...]

PR #879, opened 2017-09-13
Add common type aliases to normalizeTypes (#861)

iz-iznogood:
I have done a (very) simple implementation, I have used "string", "integer", "date" and "boolean".
I believe it would be nice to have something common for float and double. 

Please take a look and if you agree I can add these + tests so that it is complete.

Thanks

### DIFF
--- a/src/driver/mysql/MysqlDriver.ts
+++ b/src/driver/mysql/MysqlDriver.ts
@@ -332,16 +332,16 @@ export class MysqlDriver implements Driver {
         if (column.type === Number || column.type === "integer") {
             return "int";
 
-        } else if (column.type === String) {
+        } else if (column.type === String || column.type === "string") {
             return "varchar";
 
-        } else if (column.type === Date) {
+        } else if (column.type === Date || column.type === "date") {
             return "datetime";
 
         } else if ((column.type as any) === Buffer) {
             return "blob";
 
-        } else if (column.type === Boolean) {
+        } else if (column.type === Boolean || column.type === "boolean") {
             return "tinyint";
 
         } else if (column.type === "uuid") {
--- a/src/driver/oracle/OracleDriver.ts
+++ b/src/driver/oracle/OracleDriver.ts
@@ -311,16 +311,16 @@ export class OracleDriver implements Driver {
      */
     normalizeType(column: { type?: ColumnType, length?: number, precision?: number, scale?: number, isArray?: boolean }): string {
         let type = "";
-        if (column.type === Number) {
+        if (column.type === Number || column.type === "integer") {
             type += "integer";
 
-        } else if (column.type === String) {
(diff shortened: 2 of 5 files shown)


pleerock:
You must remove checks for "date" because there are database types called "date". And this database is not equal to `Date`. Usually "date" is simply dd:mm:yy, but `Date` is more then that - it also contains hrs, mm, ms

pleerock:
any updates on this?

iz-iznogood:
I need aliases for at least the common types including date, otherwise it is useless IMO.
Since we can not use date, what other options are there ?

pleerock:
other option is to use other name. But I don't think we can find name better then "date" =) So we have really only two options: 1. close this PR; 2. Left everything else and exclude date at all.

iz-iznogood:
Right now we need these data types (mabe more in the future):
Boolean, String, Int32, Int64, Int16, Byte, Decimal, Float, Double, Date, Datetime, Time, Guid, ByteArray

As it stands now I maintain the mapping in my own code, which is not a problem. But having something common across the board would be great. Also the user can change the mapping at initialization, you have something like this inside the drivers I believe (I found it while looking on the Version decorator)

We could also use specific names like torm_boolean, torm_string, etc or somenamespacing like torm.commonTypes.string, torm.commonTypes.boolean
So we can have the string names for the json and some constants or enum for the code (decorators) 
I know it is not pretty but at least this reduces the overhead on having to handle DB specific stuff on code that is totally DB agnostic.

What's your opinion ?

pleerock:
> Right now we need these data types (mabe more in the future):
Boolean, String, Int32, Int64, Int16, Byte, Decimal, Float, Double, Date, Datetime, Time, Guid, ByteArray

I don't think so. Int32 ? Why not integer32 ? Double? Why not Numeric or Real? Guid? Why not UUID? Date, Datetime, Time are already busy. See, the problem is to call it "double" or "real" is that its the same meaning, just depend on database (and programmers) preference. Im sure we shall not confuse things even more.

> We could also use specific names like torm_boolean, torm_string, etc or somenamespacing like torm.commonTypes.string, torm.commonTypes.boolean

ahh looks very very bad

My opinion is that we can add "string" and maybe "boolean" and not really more then that. Or not do do anything and simply let user use database-specific types as it was designed.

iz-iznogood:
Ok lets close it then, there is no point in implementing it only for 2 data types.

However I still believe that you have thought of it backwards. 
Usually in ORM design you begin with the abstraction and at a later stage you add the possible overrides. 

In TORM you want to have DB specific types but you don't really use the extra functionality each DB provides, for example in the query builder you have a common interface (which is good) but I do not see how you can use specialized queries that only a few RDBMS's support. And how are you going to handle the cases that a feature is supported in a small subset of databases....

PR #885, opened 2017-09-14
Add support for value transformers

### LINKED ISSUE #289: first class support for custom (un)marshallers
So i want to use momentjs instead of Date for my records and this is how I was able to get it typeorm to work:

```typescript
    @BeforeInsert() _setDt() {this.date = this.date.toDate() as any}
    @BeforeUpdate() _setDt1() {this.date = this.date.toDate() as any}
    @AfterLoad() _setDt2() {this.date = moment(this.date)}
    @Column({ type: "date", nullable: false }) date: moment.Moment
```

it would be great if there was a first class column option for setting up our own marshalling and unmarshalling so we didn't have to break types with `as any`

```typescript
@Column({ 
    type: "date", 
    nullable: false, 
    marshaller: {
        toDb: d => d.toDate(), 
        fromDb: d => moment(d)
    }
}) date: moment.Moment

```

iRath96:
This closes #289. Let me know if there's anything that needs to be changed. 😄

### DIFF
--- a/src/decorator/options/ColumnCommonOptions.ts
+++ b/src/decorator/options/ColumnCommonOptions.ts
@@ -53,6 +53,6 @@ export interface ColumnCommonOptions {
      * Specifies a value transformer that is to be used to (un)marshal
      * this column when reading or writing to the database.
      */
-    transformer?: ValueTransformer<any, any>;
+    transformer?: ValueTransformer;
 
 }
--- a/src/decorator/options/ColumnCommonOptions.ts
+++ b/src/decorator/options/ColumnCommonOptions.ts
@@ -53,6 +53,6 @@ export interface ColumnCommonOptions {
      * Specifies a value transformer that is to be used to (un)marshal
      * this column when reading or writing to the database.
      */
-    valueTransformer?: ValueTransformer<any, any>;
+    transformer?: ValueTransformer<any, any>;
 
 }
--- a/src/decorator/options/ColumnCommonOptions.ts
+++ b/src/decorator/options/ColumnCommonOptions.ts
@@ -1,3 +1,5 @@
+import {ValueTransformer} from "./ValueTransformer";
+
 /**
  * Column options specific to all column types.
  */
@@ -46,5 +48,11 @@ export interface ColumnCommonOptions {
      * Supported only by postgres.
      */
     isArray?: boolean;
+
+    /**
+     * Specifies a value transformer that is to be used to (un)marshal
+     * this column when reading or writing to the database.
+     */
+    valueTransformer?: ValueTransformer<any, any>;
 
 }
--- a/src/decorator/options/ValueTransformer.ts
+++ b/src/decorator/options/ValueTransformer.ts
@@ -7,12 +7,12 @@ export interface ValueTransformer<P, Q> {
      * Used to marshal data when writing to
      * the database.
      */
-    transformTo (value: P): Q;
+    to (value: P): Q;
 
     /**
      * Used to unmarshal data when reading from
      * the database.
      */
-    transformFrom (value: Q): P;
+    from (value: Q): P;
 
 }
--- a/src/decorator/options/ValueTransformer.ts
+++ b/src/decorator/options/ValueTransformer.ts
@@ -1,0 +1,18 @@
+/**
+ * Interface for objects that deal with (un)marshalling data.
+ */
+export interface ValueTransformer<P, Q> {
+
+    /**
+     * Used to marshal data when writing to
+     * the database.
+     */
+    transformTo (value: P): Q;
+
+    /**
+     * Used to unmarshal data when reading from
+     * the database.
+     */
+    transformFrom (value: Q): P;
+
+}
--- a/src/decorator/options/ValueTransformer.ts
+++ b/src/decorator/options/ValueTransformer.ts
@@ -1,18 +1,18 @@
 /**
  * Interface for objects that deal with (un)marshalling data.
  */
-export interface ValueTransformer<P, Q> {
+export interface ValueTransformer {
 
     /**
      * Used to marshal data when writing to
      * the database.
      */
-    to (value: P): Q;
+    to (value: any): any;
 
     /**
      * Used to unmarshal data when reading from
      * the database.
      */
-    from (value: Q): P;
+    from (value: any): any;
 
 }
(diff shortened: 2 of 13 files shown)


pleerock:
Thank you for this PR. Lets see what we can improve here and what details left to finish

src/decorator/options/ColumnCommonOptions.ts
```diff
@@ -47,4 +49,10 @@ export interface ColumnCommonOptions {
      */
     isArray?: boolean;
 
-}
\ No newline at end of file
+    /**
+     * Specifies a value transformer that is to be used to (un)marshal
+     * this column when reading or writing to the database.
+     */
+    valueTransformer?: ValueTransformer<any, any>;
```
pleerock: What do you think if we just name it `transform`?

src/decorator/options/ValueTransformer.ts
```diff
@@ -0,0 +1,18 @@
+/**
+ * Interface for objects that deal with (un)marshalling data.
+ */
+export interface ValueTransformer<P, Q> {
+
+    /**
+     * Used to marshal data when writing to
+     * the database.
+     */
+    transformTo (value: P): Q;
```
pleerock: What do you think if we name them as `to` and `from` ? Also, we should be able to do: 

```ts

 +@Entity()
 +export class Post {
 +
 +    id: number;
 +
 +    @Column()
 +    title: string;
 +
 +    @Column({ type: String, transform: { from: tags => tags.join(","), to: tags => tags.split(",") } })
 +    tags: string[];
 +
 +}
```

src/decorator/options/ValueTransformer.ts
```diff
@@ -0,0 +1,18 @@
+/**
+ * Interface for objects that deal with (un)marshalling data.
+ */
+export interface ValueTransformer<P, Q> {
```
pleerock: Also I don't think we need `<P, Q>`. They arent really used anywhere and does not really provide type safety anywhere

src/decorator/options/ColumnCommonOptions.ts
iRath96: @pleerock Sounds good to me. `valueTransformer` is indeed a bit too long.

src/decorator/options/ValueTransformer.ts
iRath96: @pleerock Yeah, that's a lot better.

src/decorator/options/ValueTransformer.ts
iRath96: @pleerock I think they're useful. If you accidentally did something like

```javascript
transform: {
  from(v: string) {
    return v.split(",");
  },
  to(v: string[]) {
    return v.concat(","); // obvious mistake here, should use .join
  }
}
```

Then the template parameters would catch that this is not a valid `ValueTransformer`.

iRath96:
@pleerock Do you want to rename the `valueTransformer` field and `transform` methods or should I do that?

pleerock:
You do, since its your PR 😉

pleerock:
You'll also need to update docs [here.](https://github.com/typeorm/typeorm/blob/master/docs/entities.md#column-options)

src/driver/oracle/OracleDriver.ts
```diff
@@ -252,8 +252,8 @@ export class OracleDriver implements Driver {
      * Prepares given value to a value to be persisted, based on its column type and metadata.
      */
     preparePersistentValue(value: any, columnMetadata: ColumnMetadata): any {
-        if (columnMetadata.valueTransformer)
-            value = columnMetadata.valueTransformer.transformTo(value);
+        if (columnMetadata.transformer)
+            value = columnMetadata.transformer.to(value);
```
pleerock: You need to do better test coverage. What if `to` is not defined in `transformer`. This code will fail. Same with `from`.

src/decorator/options/ColumnCommonOptions.ts
```diff
@@ -53,6 +53,6 @@ export interface ColumnCommonOptions {
      * Specifies a value transformer that is to be used to (un)marshal
      * this column when reading or writing to the database.
      */
-    valueTransformer?: ValueTransformer<any, any>;
+    transformer?: ValueTransformer<any, any>;
```
pleerock: I think you can remove those generic parameters from `ValueTransformer`. I don't really see value in them

iRath96: @pleerock It ensures that the `from` and the `to` method are compatible, which can be handy in some cases. Take a look at the example I made in response to your comment on the `ValueTransformer`-interface itself.

src/driver/oracle/OracleDriver.ts
iRath96: @pleerock Do you think the `ColumnMetadata` constructor would be a good place to ensure that both `from` and `to` exist on the given transformer?

src/decorator/options/ColumnCommonOptions.ts
pleerock: I don't think it really brings any value in practice, it just makes things complicated in this particular cases in my opinion

iRath96:
@pleerock Do you think the `ColumnMetadata` constructor would be a good place to ensure that both from and to exist on the given transformer?

pleerock:
no, you don't need to ensure. Users may have any of them, not necessary both. Just change:

```ts
 if (columnMetadata.transformer)
             value = columnMetadata.transformer.from(value);
```

to:


```ts
 if (columnMetadata.transformer && columnMetadata.transformer.from)
             value = columnMetadata.transformer.from(value);
```

Same with `to`.

What is also important is to add tests for these cases - with only `from` and with only `to` to make sure we are covering all possible cases

iRath96:
@pleerock I am not sure if I understand. When exactly would one specify a `from` method but not a `to` method? As I understand it, those go hand in hand together.

pleerock:
I thought people may have such use cases

pleerock:
Okay, I dont want to by annoying if you think people won't have use cases then don't do it

iRath96:
@pleerock I don't think so. If you were to only specify either `from` or `to`, then the property type and the column type would have to be identical (if they weren't, either loading from the database or storing to the database would break).

The only functionality left then would be changing the value (incrementing a counter, for example) - but people should really use `@BeforeUpdate` / `@AfterLoad` for this, because it's a lot cleaner.

pleerock:
Yes those decorators should not change any state and do side effects. Okay I think I can merge it. Thank you very much for your contribution!

Christilut:
@iRath96 It seems the transformers don't have async support? See #1050

iRath96:
@Christilut This is indeed the case. Adding async support to those would require quite a lot of effort ( value transformers are invoked in the `preparePersistentValue` method of database drives, which itself is called in numerous places which require it to be synchronous). I unfortunately don't have time to implement this right now.

Christilut:
Alright, I understand :)

Issue #572, opened 2017-06-23
Column Index exception using dropDatabase


muyu66:
Hello, 

This is my code:

```
class User {
    @Column('int', { name: 'user_id' })
    @Index({ unique: true })
    userId: number;
}
```

When i execute dropDatabase(), it throw a exception

    Error: ER_CANT_DROP_FIELD_OR_KEY: Can't DROP 'user_id'; check that column/key exists

	query failed: ALTER TABLE `user` DROP INDEX `user_id`

But i try to input raw sql "ALTER TABLE `user` DROP  `user_id`" , it is just successful.

Is this my usage problem ?

addition:
The problem may be that the column create a random name INDEX (eg. ind_7ad3e9ad3b262a0236929bf7024) when using **@Index({ unique: true })**.
So **DROP INDEX 'user_id'** must be not exist.

Fokir:
[#221](https://github.com/typeorm/typeorm/issues/221) The problem is also related to this issue

pleerock:
@iz-iznogood can you please take a look, maybe you have already fixed this issue?

iz-iznogood:
This is related to #549,  #244 and #221. 
We need to discuss this to decide on a behavior.
The workaround in #221 is correct but the problem remains.

pleerock:
fixed by @iz-iznogood , fix will be released in next version

Issue #244, opened 2017-01-30
ER_CANT_DROP_FIELD_OR_KEY on start when there is a unique index


stonecold123:
I'm getting ER_CANT_DROP_FIELD_OR_KEY on start when there is a unique column in table `user`. The problem is gone only when the unique key is deleted. I also tried putting `index()` at the top of the class like this:

`@Index("unique_user_twitter_id", (post: UserData) => [post.twitter_id], { unique: true }) `

but the problem still occurs. I have seen several posts on this. My problem only occurs when there is an unique index in the following table. The column `twitter_id` isn't linked as foreign keys to other table

````
@Entity()
export class User {
    @PrimaryGeneratedColumn({type:'int'})
    user_id: number;

    @Column({type:'string',length:255})
    user_name: string;

    @Column({type:'string',length:25})
    @Index({unique:true})
    twitter_id: string;
}
````

Error:
````
{ Error: ER_CANT_DROP_FIELD_OR_KEY: Can't DROP 'twitter_id'; check that column/key exists
    at Query.Sequence._packetToError (/home/website/node/node_modules/mysql/lib/protocol/sequences/Sequence.js:52:14)
    at Query.ErrorPacket (/home/website/node/node_modules/mysql/lib/protocol/sequences/Query.js:77:18)
    at Protocol._parsePacket (/home/website/node/node_modules/mysql/lib/protocol/Protocol.js:280:23)
    at Parser.write (/home/website/node/node_modules/mysql/lib/protocol/Parser.js:75:12)
    at Protocol.write (/home/website/node/node_modules/mysql/lib/protocol/Protocol.js:39:16)
    at Socket.<anonymous> (/home/website/node/node_modules/mysql/lib/Connection.js:103:28)
    at emitOne (events.js:96:13)
    at Socket.emit (events.js:188:7)
    at readableAddChunk (_stream_readable.js:176:18)
    at Socket.Readable.push (_stream_readable.js:134:10)
    at TCP.onread (net.js:551:20)
    --------------------
    at Protocol._enqueue (/home/website/node/node_modules/mysql/lib/protocol/Protocol.js:141:48)
    at PoolConnection.query (/home/website/node/node_modules/mysql/lib/Connection.js:208:25)
    at /build/compiled/src/driver/mysql/MysqlQueryRunner.ts:150:48
    at MysqlQueryRunner.query (/build/compiled/src/driver/mysql/MysqlQueryRunner.ts:149:16)
[...]

Issue #221, opened 2017-01-21
Ubuntu Mysql error second start application


Fokir:
Discovered the problem! When you run the application again, an error occurs. Locally, all right, what could be the problem?

```
(node:11918) UnhandledPromiseRejectionWarning: Unhandled promise rejection (rejection id: 1): Error: ER_CANT_DROP_FIELD_OR_KEY: Can't DROP 'role'; check that column/key exists
(node:11918) DeprecationWarning: Unhandled promise rejections are deprecated. In the future, promise rejections that are not handled will terminate the Node.js process with a non-zero exit code.

```
 MySQL 5.6.28-0ubuntu0.15.04.1 (Ubuntu)

Fokir:
Now I have done so
`let user = await User.__repository.createQueryBuilder('user').where('user.uplay = :uplay', {uplay: uplay})
            .andWhere('user.email is null').andWhere('user.password is null').getOne();`
It has previously been
`let user = await User.__repository.find({uplay: uplay, email: null, password: null})`

pleerock:
Have no idea. Needs more info and reproduction steps. Do you run it and you already have a database?

Fokir:
```
root@131785:~# lsb_release -a
No LSB modules are available.
Distributor ID: Ubuntu
Description:    Ubuntu 15.04
Release:        15.04
Codename:       vivid
```

`this.connection = await createConnection(databaseConfig);`

```
export let databaseConfig: ConnectionOptions = {
    driver: {
        type: "mysql",
        host: env<string>('DB_HOST', 'localhost'),
        port: env<number>('DB_PORT', 3306),
        username: env<string>('DB_USER', null),
        password: env<string>('DB_PASSWORD', null),
        database: env<string>('DB_NAME', null)
    },
    entities: [
        User, Role, Token, ....
    ],
    subscribers: [UpdateListeners],
    autoSchemaSync: true,
    logging: {
        logger: Log.logTypeorm,
        logQueries: env<boolean>('DEBUG', false),
        logSchemaCreation: env<boolean>('DEBUG', false),
        logFailedQueryError: env<boolean>('DEBUG', false)
    }
};
```

MySQL 5.6.28
base charset utf8_general_ci

Tables are created typeorm. When you restart, the error occurs. It disappears when run again. And so on to infinity. 

All logs
```
[...]

PR #886, opened 2017-09-14
Fix for issues #549, #244, #572 and #221.

### LINKED ISSUE #549: Unique Index Name on recreation
Hi,

When using a unique index like this`@Index({ unique: true })` the column is successfully created but when restarting TORM it throws an error complaining that the index does not exist.

The fix is `@Index("myCol", { unique: true })`

What appears to be happening is that the "ALTER TABLE / DROP INDEX" statement is no recognising the index by it's uid. Instead it tries to drop the index using the column Name for the index name. 

`ALTER TABLE `user` DROP INDEX myCol`

instead of

`ALTER TABLE `user` DROP INDEX ind_705816b51e5f6f664aacaa94a75`

An alternative fix is to simply restart TORM again.

cheers

JT

iz-iznogood:
Add normalizeIsUnique method to driver interface and use it to check if a column is isUnique compared to what the DB schema reports. There are differences on how a DB handles the unique column case.

MySQL - MariaDB consider a column as unique if there is a unique index on the column, without the need for a unique constraint. 
So we need to adjust on how we resolve differences so that we take this into account (only for this DB type).

This fixes #549, #244, #572 and #221

### DIFF
--- a/src/driver/Driver.ts
+++ b/src/driver/Driver.ts
@@ -110,6 +110,11 @@ export interface Driver {
     normalizeDefault(column: ColumnMetadata): string;
 
     /**
+     * Normalizes "isUnique" value of the column.
+     */
+    normalizeIsUnique(column: ColumnMetadata): boolean;
+
+    /**
      * Normalizes "default" value of the column.
      */
     createFullType(column: ColumnSchema): string;
--- a/src/driver/mongodb/MongoDriver.ts
+++ b/src/driver/mongodb/MongoDriver.ts
@@ -242,6 +242,13 @@ export class MongoDriver implements Driver {
     }
 
     /**
+     * Normalizes "isUnique" value of the column.
+     */
+    normalizeIsUnique(column: ColumnMetadata): boolean {
+        throw new Error(`MongoDB is schema-less, not supported by this driver.`);
+    }
+    
+    /**
      * Normalizes "default" value of the column.
      */
     createFullType(column: ColumnSchema): string {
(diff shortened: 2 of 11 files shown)


pleerock:
Do you think we shall add tests for this case as well?

iz-iznogood:
Yes I will send the tests shortly, if you agree with the approach though... 

At least 3 of them are exactly the same scenario so for them one test will do. The other (#572) I will see if we need a separate test.

So are you good with how it is implemented so that I can code the tests ?

pleerock:
Implementation does not affect the tests, isnt it?

pleerock:
Im not sure about this implementation yet, I need to understand you why did so and isnt there a better solution....

pleerock:
are you sure this issue is mysql-specific? For example as I see right now posgres has same logic - it creates constraints if unique is used, also it can create an unique index. Then what is the difference? 

```ts
!!column.entityMetadata.indices.find(index => index.isUnique && index.columns.length === 1 && index.columns[0] === column)
```

For example, why this code cannot be applied for other drivers as well?

iz-iznogood:
Ok, I just pushed the test...

iz-iznogood:
The difference is that when you read the schema from the database there is no way to know in MySql if the column is unique or not since they report it unique not because of unique constraint but also because of unique index.

For example in sql server a column is not unique unless you create a unique constraint for that column.
Same for postgres

In mysql this does not work though constraints but directly though indices

pleerock:
Now I completely understood the issue, thx for explanation, however... I don't really like current fix. Maybe we can came up to something more elegant?

pleerock:
Okay we can merge it. I cannot came up to something better. In the future Im planning to refactor this `findChangedColumns` method and maybe move it into the `QueryRunner` and make it completely driver-specific. Then this will be refactored as well

Issue #798, opened 2017-08-26
sqlite: "database" path in ormconfig.json is not relative


stahlstift:
### Given

```
# Example ormconfig.json
[
  {
    "name": "default",
    "type": "sqlite",
    "database": "./data/example.sq3",
    "autoSchemaSync": true,
    "entities": [
      "src/Model/*.js"
    ]
  }
]
```

```
#Filestructure
src/**.*
package.json
ormconfig.json
```

### My expection

```
data/example.sq3
src/**.*
package.json
ormconfig.json
```

### Result
Error: SQLITE_CANTOPEN: unable to open database file

### Question
Bug? Should I fix this and make a pull request?

pleerock:
where it creates it right now?

stahlstift:
I think in process.cwd()

pleerock:
okay I think you can create a PR for this

pleerock:
hey @stahlstift didn't you had a chance to work on it yet?

stahlstift:
@pleerock No sorry. My laptop is broken... I think I can start next week on this topic.

pleerock:
This change will be released in 0.1.2

PR #896, opened 2017-09-17
#798 sqlite: A relative 'database' path is now relative to ormconfig

stahlstift:
Here we go @pleerock


src/connection/ConnectionOptionsReader.ts
```diff
@@ -138,6 +138,13 @@ export class ConnectionOptionsReader {
                 });
                 Object.assign(connectionOptions, { migrations: migrations });
             }
+            if (options.type === "sqlite") {
```
pleerock: It does not look like a good place to handle it. You need to handle this option in sqlte driver itself.

src/driver/sqlite/SqliteDriver.ts
```diff
@@ -77,11 +80,55 @@ export class SqliteDriver extends AbstractSqliteDriver {
     // Protected Methods
     // -------------------------------------------------------------------------
 
+    protected createDatabaseFile(): Promise<void> {
+        return new Promise<void>((resolve, reject) => {
+            mkdirp(dirname(this.options.database), (err: NodeJS.ErrnoException) => {
```
pleerock: we cannot use nodejs dependencies because typeorm is crossplatform (there is no node deps in browser for example). replace to "any" in this case

src/driver/sqlite/SqliteDriver.ts
```diff
@@ -7,6 +7,9 @@ import {SqliteConnectionOptions} from "./SqliteConnectionOptions";
 import {ColumnType} from "../types/ColumnTypes";
 import {QueryRunner} from "../../query-runner/QueryRunner";
 import {AbstractSqliteDriver} from "../sqlite-abstract/AbstractSqliteDriver";
+import {close, open, stat} from "fs";
+import {dirname} from "path";
+import * as mkdirp from "mkdirp";
```
pleerock: we cannot use nodejs dependencies because typeorm is crossplatform (there is no node deps in browser for example). extract "fs" and "path" logic into PlatformTools and add corresponding code into BrowserPlatformTools. Also probably its better to write recursive directory creation code manually then adding another dependency...

src/driver/sqlite/SqliteDriver.ts
```diff
@@ -77,11 +80,55 @@ export class SqliteDriver extends AbstractSqliteDriver {
...
+                    }
+
+                    return reject(err);
+                }
+
+                return resolve(stats.isFile());
+            });
+        });
+    }
+
```
pleerock: both those 2 functions looks like they can be extracted into separate class related to filesystem (not sqlite anyhow) and then used there.

pleerock: Maybe PlatformUtils is that place...

src/driver/sqlite/SqliteDriver.ts
pleerock: Looks like we already have `mkdirp` package, so you only need to abstract things into PlatformTools

stahlstift:
Thanks for your response, but I don't want to get so deeply into typeorm. I even don't use it ;)
I thought it would be a simple and quick solution. So feel free to decline my merge request and to close the tickets.

pleerock:
Okay, thanks for your contribution. I'll merge it apply other changes by myself.

PR #901, opened 2017-09-18
One more fix for #855

iz-iznogood:
We should use the same pattern for sequence names everywhere. 
There is one more wrong usage in dropIndex but it is never used with the isGenerated == true so we can leave it for later. Also in dropIndex we do not have the column name so it is not possible to create the correct sequence name.
This is a followup of commit c0a00ad68b52dffeb4da1bd57de8092d01e237a0


iz-iznogood:
This is added to another pull request with more changes to go in

Issue #905, opened 2017-09-18
Support length for schema changes


iz-iznogood:
Currently changing length for supported properties does not update the schema

Issue #897, opened 2017-09-17
Is it possible to set nvarchar(MAX) ?


iz-iznogood:
Hi, how can I configure nvarchar(max) in mssql ?

pleerock:
Good question. Actually it should be:

```ts
@Column({ type: "nvarchar" length: "MAX" })
```

But Im not sure if it will work since as I remember we convert length to number later on.
If it won't work then we'll need to change some logic and convert it to string instead of number everywhere. We need a single format (or number or string) because we make comparing down the road.

iz-iznogood:
Yes, I also checked that you never take into account the column size for finding changed columns.
These should be fixed together since they are related.

We should change the length property to string and fix all comparisons.
We should change the TableSchema.findChangedColumns to use length to compare only datatypes that size matters.
Now it is the time to move a part of TableSchema.findChangedColumns into the driver (the actual comparison part) since the datatype names are different in each database engine and we are going to use that so that we know if a datatype supports the length attribute.

@pleerock  Is it ok if I try this ? do you agree with the approach ?

pleerock:
Yes, I was thinking about it too. Im not satisfied with current implementation of `findChangedColumns` as well. Lets do this 👍

iz-iznogood:
Hi, I have changed the length property in ColumnMetadata and ColumnSchema to be number | string and it seems to work fine.

However I need to add some validations on whether the column type supports length - size.
I see that you have this 

`
export type WithLengthColumnType = "int" // mysql, postgres, mssql, oracle, sqlite
    |"tinyint" // mysql, mssql, sqlite
    |"smallint" // mysql, postgres, mssql, oracle, sqlite
    |"mediumint" // mysql, sqlite
    |"bigint" // mysql, postgres, mssql, sqlite
    |"character varying" // postgres
    |"varying character" // sqlite
    |"nvarchar" // mssql
    |"character" // mysql, postgres, sqlite
    |"native character" // sqlite
    |"varchar" // mysql, postgres, mssql, sqlite
    |"char" // mysql, postgres, mssql, oracle
    |"nchar" // mssql, oracle, sqlite
    |"varchar2" // oracle
    |"nvarchar2" // oracle, sqlite
    |"binary" // mssql
    |"varbinary"; // mssql

`

This however does not emit any runtime type information so that I can validate if a Datatype supports length at runtime. 
What I need is a method columnTypeSupportsLength(column: ColumnType): boolean.
[...]

PR #906, opened 2017-09-18
Issue #855, #897 and #905

iz-iznogood:
This implements issue #897 and #905.

We change the length property to be string and adjust all related comparisons (#897).
We add an array per driver with the properties that support length and use this to properly compare length during schema update (#905)
Add tests for mysql, mssql, postgres.

Please review and merge if possible


iz-iznogood:
Please note that I did not move the findChangedColumns in the driver, this I can do in another commit once these changes are settled

src/driver/Driver.ts
```diff
@@ -47,6 +47,11 @@ export interface Driver {
     dataTypeDefaults: DataTypeDefaults;
 
     /**
+     * Gets list of column data types that support length by a driver.
+     */
+    withLengthColumnType: ColumnType[];
```
pleerock: this should be called `withLengthColumnTypes` since its an array.

pleerock: Also I don't really like `withLengthColumnType` name but I cannot suggest something better at the moment

src/driver/Driver.ts
```diff
@@ -115,6 +120,11 @@ export interface Driver {
     normalizeIsUnique(column: ColumnMetadata): boolean;
 
     /**
+     * Calculates column length taking into account the default length values.
+     */
+    getColumnLength(column: ColumnMetadata): number | string | undefined;
```
pleerock: Do you really think we shall deal with `number|string` pain everywhere? Maybe we just cast into into string (for example when we read from decorator or somewhere in column metadata, I don't know) and use that string everywhere then?

src/driver/Driver.ts
iz-iznogood: Sorry for the typo, I fixed it

src/driver/Driver.ts
iz-iznogood: I am a bit worried about such a change. 
Now, the only reason length is string is to allow the "ΜΑΧ" value (or similar) and this could have been avoided using a convention like -1 for max, still the type would have been integer and we could handle this inside each driver (although it might look a bit strange or even create problems in some cases).

What worries me is that at some places we pass length to the actual db driver(s) and they might expect this to be number not string. Now we convert it to number where it is possible so the property type would be correct. In the other case we might have to check a ton of scenarios. 

I have not looked into each driver in depth, this would take huge amount of time to get into such details.

It is still your call, what would you like me to do with this ?

pleerock: > What worries me is that at some places we pass length to the actual db driver(s) and they might expect this to be number not string.

we do not pass length anywhere into underlying db drivers. We use length only internally to concatenate length to column type. Also we use length to compare things to understand if column has changed.

iz-iznogood: How about MsSqlParameter ? This also passes the length implicitly i believe ...

iz-iznogood:
Why some tests fail randomly ? Node 4 passes and Node stable fails for no reason .....
Any clues ?

src/driver/Driver.ts
pleerock: not sure can you check that, anyway Im sure its better to pass things in a single format then in two formats

pleerock:
This particular test failed because of ordering. Sometimes when we aren't awaiting things to be inserted in order it inserts things in order we aren't expect and ids aren't match our expectations. Solution is to insert things in order. Ive restarted failing build

src/driver/Driver.ts
iz-iznogood: Ok I will change it

pleerock: yeah lets see what will be

iz-iznogood:
Wow, to my surprise tests pass !!!!
Can you take a look please ? It's a lot of small changes all over, but most of them are trivial.
If you don't like something please say so that I can change it if I can.

Also would you prefer a new 'clean' pull request or this is ok ?

pleerock:
merged, thank you

iz-iznogood:
Thanks !!!
This closes #905 and #897

PR #909, opened 2017-09-19
Fixes #904

### LINKED ISSUE #904: Using closure tables without @TreeLevelColumn will always fail on insert
Hi,

the issue can be seen on the following line:
https://github.com/typeorm/typeorm/blob/51caf9fe30462ab44eacab69dbff1a4882bfd1f1/src/driver/postgres/PostgresQueryRunner.ts#L315

There should be a check using `hasLevel` otherwise you'll get the following error on insert:

> QueryFailedError: column "level" does not exist

Other functions which call `insertIntoClosureTable` probably have to be updated as well because they dependent on the returned number.

martinschleiss:


test/github-issues/904/issue-904.ts
```diff
@@ -0,0 +1,204 @@
...
+import {createTestingConnections, closeTestingConnections, reloadTestingDatabases} from "../../utils/test-utils";
+import {Connection} from "../../../src/connection/Connection";
+import {Category} from "./entity/Category";
+
+describe("github issues > #904 Using closure tables without @TreeLevelColumn will always fail on insert", () => {
+
+    let connections: Connection[];
+    before(async () => connections = await createTestingConnections({
+        entities: [__dirname + "/entity/*{.js,.ts}"],
+        enabledDrivers: ["mssql", "mysql", "oracle", "postgres", "sqlite"],
```
pleerock: no need to list all drivers if you want it to run for all of them

martinschleiss: ok

pleerock:
Thank you for your contribution! This change will be released in next alpha version.

PR #918, opened 2017-09-20
Entity schema split

iz-iznogood:
Small trivial pull request to split EntitySchema into more interfaces (EntitySchemaTable, EntitySchemaColumn, EntitySchemaIndex, EntitySchemaRelation) so that a large entity can be created easier using typed objects

### DIFF
--- a/src/entity-schema/EntitySchema.ts
+++ b/src/entity-schema/EntitySchema.ts
@@ -29,247 +29,256 @@ export interface EntitySchema { // todo: make it-to-date
     /**
      * Entity table's options.
      */
-    table?: {
-
-        /**
-         * Table name.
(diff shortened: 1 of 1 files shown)


src/entity-schema/EntitySchema.ts
```diff
@@ -29,247 +29,256 @@ export interface EntitySchema { // todo: make it-to-date
...
+     * Entity column's options.
+     */
+    columns: {
+        [columnName: string]: EntitySchemaColumn 
+    };
+
+    /**
+     * Entity relation's options.
+     */
+    relations: {
```
pleerock: probably relations can be "?" as well

PR #927, opened 2017-09-22
More Entity schema exports

iz-iznogood:
Adding new exports


src/index.ts
```diff
@@ -122,7 +122,7 @@ export {UpdateEvent} from "./subscriber/event/UpdateEvent";
 export {RemoveEvent} from "./subscriber/event/RemoveEvent";
 export {EntitySubscriberInterface} from "./subscriber/EntitySubscriberInterface";
 export {BaseEntity} from "./repository/BaseEntity";
-export {EntitySchema} from "./entity-schema/EntitySchema";
+export {EntitySchema, EntitySchemaTable, EntitySchemaColumn, EntitySchemaIndex, EntitySchemaRelation} from "./entity-schema/EntitySchema";
```
pleerock: Can you extract each interface into its own file? Looks like a mess there

iz-iznogood: Ok will do ASAP

PR #941, opened 2017-09-26
Optimized usage of ValueTransformer

dropdevcoding:
* Optimized usage of ValueTransformer
* Removed ValueTransformers from drivers
* Added whole handling to ColumnMetadata
* Enabled possibility of transforming primary key columns (even in embedded mode)


pleerock:
`getEntityValue` and `getEntityValueMap` are very sensitive places and its very dangerous to change something there. Can you please explain me what is the exact reason of moving transformation logic into there?

dropdevcoding:
If you look at the driver implementations there is nothing driver specific happening. The code is duplicated multiple times and thus it is not the responsibility of a driver imho. Besides that the value transformation didn't work for primary keys in several scenarios (writing and reading). 

As you already mentioned it is a very sensitive place and why is that? Because all the value handling of a column happens there and since it is a **Value**Transformer which transforms a **value** of a **column** it should exactly belong to this place. 

I didn't have to change that much in other places after placing the logic in there which is quite a good indicator that the architecture is well done if you put things into the right place.

If you don't agree you possibly may have a look at a more driver related solution like Doctrine does it with their Custom Mapping Types: http://docs.doctrine-project.org/projects/doctrine-orm/en/latest/cookbook/custom-mapping-types.html

pleerock:
> If you look at the driver implementations there is nothing driver specific happening. The code is duplicated multiple times and thus it is not the responsibility of a driver imho.

I think I can agree with you, however I think getEntityValue and getEntityValueMap not a good place where this logic must be extracted. The place where it goes should be is during persistence and hydration operations - update value right before saving it into the database, and update value right after loading them from the database.

> very sensitive place and why is that?

`getEntityValue(entity)` means you pass entity to it and it returns its column value.
For example `dateColumn.getEntityValue(Entity{ date: moment() }` must return `moment()`, not a `Date` which transformation does.
Also its used very deep inside, which are hard to explain without knowledge of codebase.

>  Because all the value handling of a column happens there and since it is a ValueTransformer which transforms a value of a column it should exactly belong to this place.

same naming does not mean same meaning and contexts 

> I didn't have to change that much in other places after placing the logic in there which is quite a good indicator that the architecture is well done if you put things into the right place.

you have zero test coverage of value transformation + million of other use cases, thats why you don't have issues right now. But changing logic of `getEntityValue` and `getEntityValueMap` non careful may bring unexpected results in places where you may not expect.

> If you don't agree you possibly may have a look at a more driver related solution like Doctrine does it with their Custom Mapping Types: http://docs.doctrine-project.org/projects/doctrine-orm/en/latest/cookbook/custom-mapping-types.html

Doctrine's custom mapping types is a bit different solution which aims to provide a **custom database type*,* not a **simple value transformation** just like what typeorm does. At the moment Im not sure if typeorm needs *additional* functionality like in doctrine, because in doctrine its more useful since it does not support native database column types like what typeorm has. But we may have such feature as well in the future if some users will have specific use cases which such solution covers.

You told you have fixed issues with primary columns and embeds, can you please add tests for them, so I can take a look on a problem more deeply? 

btw looks like there is a wrongly resolved conflict

dropdevcoding:
> you have zero test coverage of value transformation

I've already added some test cases for the specific issues I was facing. If everything else is covered well you should experience broken tests if the implementation would break the existing one, but the tests are still green.

The former implementation only worked correctly for properties which didn't have a special functionality such as being a primary key.

If I defined an id as a ValueObject I 
* expect the ValueObject to be transformed
* expect that I also may use 
```
repository.findOneById(instanceOfValueObjectId)
``` 
to be transformed before executing the query

* expect that I also may use 
```
repository.findOneById({
  id: instanceOfValueObject,
  embedded: {
    id: instanceOfEmbeddedValueObjectId
  }
})
```
 to be transformed before executing the query for fetching objects containing embedded objects which also contain a primary key and so on.

Besides that the ```ValueTransformer``` interface wasn't exported for reuse and wasn't respected in ```EntitySchemaColumn``` and ```EntitySchemaTransformer```

> Doctrine's custom mapping types is a bit different solution which aims to provide a custom database type, not a simple value transformation just like what typeorm does.

Yes, I only mentioned it because the handling of a custom database type and the containing mapping would make sense to reside inside the drivers (how it was before) since they depend on the driver - a simple transformation does not and should therefore not reside in the driver.

pleerock:
Sorry for late answer, last three weeks were really busy since I was working actively on a new site and preparing a 0.1.0 release

pleerock:
This PR needs a detailed review, since as I already told getEntityValue and getEntityValueMap are very sensitive places. So I would ask you to help me to understand all problems currently exist, then I can review all places where those methods are used and maybe apply your changes, or move this logic into another method and apply column.newMethod(column.getEntityValue(...)) if needed.
One place as you shown is `find` method:

```ts
repository.findOneById({
  id: instanceOfValueObject,
  embedded: {
    id: instanceOfEmbeddedValueObjectId
  }
})
```

what are other places? We definitely need failing tests for this and other cases.

pleerock:
Closing because of no answer and activity from the developer.

haseebnaseem:
@pleerock @dropdevcoding Can we please open this again ? Happy to contribute whatever is needed to get this done. We want to use value objects in entities, and transformers are not applied when a foreign key in an entity is a value object. Example:

```
@Entity()
Class Organization {
  @PrimaryColumn({
        type: "uuid",
        transformer: {
            from: (value: string) => new Uuid(value),
            to: (value: Uuid) => value.toString(),
        },
    })
    id: Uuid;
  
  @Column()
  name: string
}

class Member {
    @PrimaryColumn({
        type: "uuid",
        transformer: {
            from: (value: string) => new Uuid(value),
            to: (value: Uuid) => value.toString(),
        },
    })
    id: Uuid;
    
    @Column()
    name: string
    
    @ManyToOne(() => Organization)
    organization: Organization
}

const member = new Member()
member.organization = SampleOrganization
this.manager.save<Member>(member);

```

This throws an error with something like `invalid format for uuid "{value: \"xxx-xxx-xxxx--xxxx\"}"

PR #946, opened 2017-09-27
Issue 945: Sqlite with multiple primary keys

### LINKED ISSUE #945: Sqlite: composite primary key ignored on synchronization
When using two columns as primary key in sqlite like this:
```typescript
import { Entity, PrimaryColumn, Column } from 'typeorm';

@Entity()
export class ExampleEntity {
  @PrimaryColumn()
  id1: string;

  @PrimaryColumn()
  id2: string;

  @Column()
  otherColumn: boolean;
}
```
When making a change a creation sql like this is generated:
```sql
CREATE TABLE "temporary_example_entity"
("id1" varchar NOT NULL, "id2" varchar NOT NULL, "otherColumn" boolean, PRIMARY KEY(id1))
```
But is should be
```sql
CREATE TABLE "temporary_example_entity"
("id1" varchar NOT NULL, "id2" varchar NOT NULL, "otherColumn" boolean, PRIMARY KEY(id1,id2))
```

This issue only exists when syncing. The initial creation is correct.

daniel-lang:
fixes #945

### DIFF
--- a/test/github-issues/945/issue-945.ts
+++ b/test/github-issues/945/issue-945.ts
@@ -17,18 +17,16 @@ describe("github issues > #945 synchronization with multiple primary keys", () =
         const tableSchema = await connection.createQueryRunner().getTable("test_entity");
 
         if (tableSchema) {
-            const firstId = tableSchema.columns.find(column => {
-                return column.name === "id1";
+            const firstId = tableSchema.primaryKeys.find(column => {
+                return column.columnName === "id1";
             });
-            const secondId = tableSchema.columns.find(column => {
-                return column.name === "id2";
+            const secondId = tableSchema.primaryKeys.find(column => {
+                return column.columnName === "id2";
             });
 
-            if (firstId && secondId) {
-                expect(tableSchema.primaryKeys).length(2);
-                expect(firstId.isPrimary).to.be.true;
-                expect(secondId.isPrimary).to.be.true;
-            }
+            expect(tableSchema.primaryKeys).length(2);
+            expect(firstId).not.to.be.undefined;
+            expect(secondId).not.to.be.undefined;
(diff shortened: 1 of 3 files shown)


daniel-lang:
According to the [sqlite docs](https://sqlite.org/pragma.html#pragma_table_info) the column "pk" in `pragma table_info()` is 0 when the column isn't a primary key and otherwise the index

pleerock:
Looks like test is failing

daniel-lang:
It is failing for postgres. I currently checking why

daniel-lang:
For the postgres drivers the columns aren't marked with `isPrimary`. But the `primaryKeys` array includes the correct keys. If this is by design, you can merge this PR

test/github-issues/945/issue-945.ts
```diff
@@ -0,0 +1,33 @@
...
+    let connections: Connection[];
+    before(async () => connections = await createTestingConnections({
+        entities: [__dirname + "/entity/*{.js,.ts}"],
+        dropSchema: true
+    }));
+    beforeEach(() => reloadTestingDatabases(connections));
+    after(() => closeTestingConnections(connections));
+
+    it("schema should include two primary keys", () => Promise.all(connections.map(async connection => {
+        const tableSchema = await connection.createQueryRunner().getTable("test_entity");
```
pleerock: when you create a query runner, you must always release it

daniel-lang:
Thank you for pointing that out. Is the postgres behaviour indended?

pleerock:
I think no. That part is quite old and needs refactoring. Most probably its a bug

daniel-lang:
Okay, so this request is good to go or should I investigate the problem?

Issue #953, opened 2017-09-28
MySQL 5.7 JSON column parse


Diluka:
I stored a array `["ADMIN"]` in mysql json column, when I load it here comes an error `SyntaxError: Unexpected token A in JSON at position 0`

MySQL Driver: `"mysql2": "^1.4.2"`

I inspect the [code](https://github.com/typeorm/typeorm/blob/master/src/driver/mysql/MysqlDriver.ts#L336-L337), I found the`MysqlDriver` in `TypeORM` doesn't check the value is `string` or `object`.

Using `"mysql2": "^1.4.2"` connect to mysql5.7 will parse json string to object automatically.

Diluka:
@pleerock I create a PR to fix this

PR #954, opened 2017-09-28
fix: typeorm/typeorm#953

Diluka:
fix: typeorm/typeorm#953

### DIFF
--- a/test/github-issues/953/issue-953.ts
+++ b/test/github-issues/953/issue-953.ts
@@ -1,14 +1,15 @@
 import "reflect-metadata";
-import { closeTestingConnections, createTestingConnections, reloadTestingDatabases } from "../../utils/test-utils";
-import { Connection } from "../../../src/connection/Connection";
-import { expect } from "chai";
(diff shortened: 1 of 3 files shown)


daniel-lang:
You should add
```
enabledDrivers: ["mysql"]
```
to your test (issue-953.ts) on line 12, otherwise travis is running it on sqlite, postgres and mariadb as well. Sqlite and postgres don't support json as far as I know. I'm not sure about mariadb.

test/github-issues/953/issue-953.ts
```diff
@@ -0,0 +1,31 @@
...
+        let user = repo.create({
+            username: "admin",
+            password: "admin",
+            roles: ["ADMIN"],
+            lastLoginAt: new Date()
+        });
+        await repo.save(user);
+
+        let user1 = await repo.findOne({ username: "admin" });
+        expect(user1).to.be.an("object");
```
pleerock: also add assertion what your user1 object should be equal to, not simply check if it an object.

PR #955, opened 2017-09-28
SchemaBuilder: Table.addPrimaryKeys and Table.removePrimaryKeys (issue 952)

### LINKED ISSUE #952: SchemaBuilder: Inconsistency of primary keys
Follow up to #946 with another problem.

When a primary key is added to a table [`RdbmsSchemaBuilder.updatePrimaryKeys()`](https://github.com/typeorm/typeorm/blob/master/src/schema-builder/RdbmsSchemaBuilder.ts#L362) adds it to the primary key array but leaves out the `isPrimary` flag on the column itself.
This causes the primary key to not be created in Sqlite since [there](https://github.com/typeorm/typeorm/blob/master/src/driver/sqlite-abstract/AbstractSqliteQueryRunner.ts#L691) the flag is checked instead of the `primaryKeys` array.

@pleerock you mentioned, that you want to refactor the code. What is your idea for this?

daniel-lang:
@iz-iznogood is this what you had in mind? I'm not sure, that's the most beautiful way of doing it.

fixes #952

### DIFF
--- a/src/schema-builder/schema/Table.ts
+++ b/src/schema-builder/schema/Table.ts
@@ -162,7 +162,13 @@ export class Table {
      * Adds all given primary keys.
      */
     addPrimaryKeys(addedKeys: TablePrimaryKey[]) {
-        addedKeys.forEach(key => this.primaryKeys.push(key));
+        addedKeys.forEach(key => {
+            this.primaryKeys.push(key);
+            const index = this.columns.findIndex(column => column.name === key.columnName);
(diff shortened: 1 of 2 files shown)


iz-iznogood:
Hi @daniel-lang 

Well in IMHO this is a nice simple localized and contained change. 
Unless @pleerock  or @AlexMesser who are the orm gurus have something else in mind I ACK this

PR #969, opened 2017-10-01
Fix column naming for deeper embedded entities

Ostrzy:
Example from provided test would be "countersInformationDescription" which previously would get database name "countersCountersInformationDescription". Problem was that parentEmbeddedMetadata was already taken into consideration when building prefix so anything deeper than one level would add earlier prefixes multiple times.

Previously `buildParentPrefixes` for `countersInformationDescription` would be `["counters", "counters_information]` and now it is simply `["counters_information"]`.

The base test for insert, update, remove is basically the same as the one for embedded columns for MongoDB.


pleerock:
Thank you!

PR #981, opened 2017-10-03
In the case of .env file, it will load the default (global) .env file name.

erickponce:
It will introduce a breaking change because it will not locate ormconfig.env, instead it will search for a .env file.

Maybe its better to implement in a way that the 2 files work.


pleerock:
> Maybe its better to implement in a way that the 2 files work.

then changes should be in this PR

pleerock:
Thank you!

Issue #977, opened 2017-10-03
0.1.0 final is going to be released 


pleerock:
Hey guys, Im going to release `0.1.0` tomorrow. I have updated a website and documentation. I wanted to ask everybody who can [review everything there](http://typeorm.io). Also need to do a grammar check, Im sure you'll find lot of mistakes there.

Thank you!

luchillo17:
Wow new docs, nice.

shousper:
New website looks great!

Strongly suggest closing off [this issue about timestamps](https://github.com/typeorm/typeorm/issues/838) before release though. Kind of looks like it should already be closed??

shousper:
Some broken links on this page: http://typeorm.io/#/connection-options -- I think just the "Learn more about <blah>" links need updating as they point to `.md` files at the moment.

pleerock:
@shousper thank you. Are you able to create PRs for all issues are you able to resolve by yourself? Its really easy, you can edit page directly from the website (link on the bottom of each page)

shousper:
@pleerock Submitted a PR for the link fixes, but the others you might wanna look at ;)

rajshrimohanks:
@pleerock Man, I was off for 3 weeks and so much has happened. The new website looks fantastic. Kudos! :)

pleerock:
Ive made a final `0.1.0 ` release and published links in reddit. Please upvote and left some comments there if you are a reddit users!

NoNameProvided:
You should send it to Hacker News as well. With the “Show HN: ... “ format.

pleerock:
Upvote [this one.](https://www.reddit.com/r/programming/comments/759gal/typeorm_010_is_released_besides_mysql_postgresql/) Some a****** downvoted it.

rajshrimohanks:
I'm going to create a Reddit account just to upvote this. 👍

jcudzich:
Good job @pleerock, do not worry about haters! Keep it up!!!
I voted up on your reddit post.

pleerock:
@jcudzich thanks 😆

PR #982, opened 2017-10-03
[Docs] spelling and general improvements

daniel-lang:
Thank you for your hard work putting together the new docs. Here are some improvements for the first four doc pages. I don't know if you like "Learn more about x", but to me, it sounds better the original wording.
Also added cordova connection options to the docs.

Related to #977


pleerock:
Thank you very much!

PR #984, opened 2017-10-03
[docs] spelling and general improvements (part 2)

daniel-lang:
The next four pages. I have a couple of questions here. Please see below.

Related to #977

### DIFF
--- a/docs/decorator-reference.md
+++ b/docs/decorator-reference.md
@@ -37,17 +37,17 @@
 
 #### `@Entity`
 
-Marks your model as an entity. Entity is a class which is transformer into database table.
-You can specify table name into the entity:
+Marks your model as an entity. Entity is a class which is transformed into a database table.
+You can specify the table name in the entity:
 
 ```typescript
 @Entity("users")
 export class User {
 ```
 
-This code will create database table named "users".
+This code will create a database table named "users".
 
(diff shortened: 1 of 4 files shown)


docs/decorator-reference.md
```diff
@@ -97,19 +97,19 @@ export class User {
 }
 ```
 
-`@Column` decorator accept several options you can use:
+`@Column` accept several options you can use:
 
 * `type: ColumnType` - Column type. One of the type listed [above](#column-types).
```
daniel-lang: This document does not contain a header called "column types". So this link goes nowhere

docs/decorator-reference.md
```diff
@@ -690,66 +688,41 @@ export class User {
...
-    @Post("/users")
-    save(@TransactionManager() manager: EntityManager, @Body() user: User) {
-        return manager.save(user);
-    }
-    
-    
+
+@Transaction()
+save(@TransactionManager() manager: EntityManager, @Body() user: User) {
+    return manager.save(user);
```
daniel-lang: I simplified these examples since it felt like they contain irrelevant code

docs/decorator-reference.md
pleerock: I moved different sections in places so many times...

PR #985, opened 2017-10-03
[docs] spelling and general improvements (part 3)

daniel-lang:
Part 3 of I don't know how many...

The content of [entity-inheritance.md](https://github.com/typeorm/typeorm/tree/master/docs/entity-inheritance.md) is also in [entities.md#entity-inheritance](https://github.com/typeorm/typeorm/blob/master/docs/entities.md#entity-inheritance). Do you want to have two files or one? I would remove one of the two.

Related to #977

### DIFF
--- a/docs/decorator-reference.md
+++ b/docs/decorator-reference.md
@@ -99,7 +99,7 @@ export class User {
 
 `@Column` accept several options you can use:
 
-* `type: ColumnType` - Column type. One of the type listed [above](#column-types).
+* `type: ColumnType` - Column type. One of the [supported column types](./entities.md#column-types).
 * `name: string` - Column name in the database table. 
 By default the column name is generated from the name of the property.
 You can change it by specifying your own name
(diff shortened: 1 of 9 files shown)


docs/decorator-reference.md
```diff
@@ -99,7 +99,7 @@ export class User {
 
 `@Column` accept several options you can use:
 
-* `type: ColumnType` - Column type. One of the type listed [above](#column-types).
+* `type: ColumnType` - Column type. One of the [supported column types](./entities.md#column-types).
```
pleerock: i guess website will not create a proper non-github link...

docs/embedded-entities.md
```diff
@@ -17,10 +17,10 @@ export class User {
     id: string;
     
     @Column()
-    nameFirst: string;
```
pleerock: I called them inverse for some reason, I dont remember why... something with final database column names

docs/decorator-reference.md
daniel-lang: What do you mean with "non-github link"?

docs/embedded-entities.md
daniel-lang: Do you want me to revert these changes?

daniel-lang:
Could you tell me, what you think of this
> The content of entity-inheritance.md is also in entities.md#entity-inheritance. Do you want to have two files or one? I would remove one of the two.

Issue #983, opened 2017-10-03
Schema sync tries to "change" nullable columns and recreates foreign keys


d-fischer:
Nullable fields don't seem to be recognized properly in the schema sync, and it tries to alter the columns every time, dropping and recreating any related foreign keys in the process. It even wants to change the automatically generated ID column for a relation!

Version: 0.1.0-alpha.50 (definitely, double-checked it with `node_modules/.bin/typeorm -v`)

Code (this is the *whole* code I'm running, in a single file):

```typescript
import 'reflect-metadata';
import { Column, createConnection, Entity, ManyToOne, OneToMany, PrimaryGeneratedColumn } from 'typeorm';

@Entity()
class User {
	@PrimaryGeneratedColumn()
	id: number;

	@Column('int', {nullable: true})
	accessTokenExpires: number | null;

	@OneToMany(type => Session, session => session.user)
	sessions: Session[];

	@Column('text', {nullable: true})
	bio: string | null;
}

@Entity()
class Session {
	@PrimaryGeneratedColumn()
	id: number;

	@ManyToOne(type => User, user => user.sessions)
	user: User;
}

(async () => {
	try {
		await createConnection({
			type: 'mariadb',
			host: 'localhost',
			username: 'root',
			password: '',
			database: 'testcase',

			entities: [
				User, Session
			],
			synchronize: true,
			logging: 'all'
		});
		console.log('Connection to database successful');
	} catch (e) {
		console.error('Application startup failed', e);
	}
})();
```

Output on first run:
```
executing query:  START TRANSACTION
executing query:  SELECT * FROM INFORMATION_SCHEMA.TABLES WHERE TABLE_SCHEMA IN ('testcase') AND TABLE_NAME IN ('user', 'session')
executing query:  SELECT * FROM INFORMATION_SCHEMA.COLUMNS WHERE TABLE_SCHEMA IN ('testcase')
executing query:  SELECT * FROM INFORMATION_SCHEMA.STATISTICS WHERE TABLE_SCHEMA IN ('testcase') AND INDEX_NAME != 'PRIMARY' ORDER BY SEQ_IN_INDEX
executing query:  SELECT * FROM INFORMATION_SCHEMA.KEY_COLUMN_USAGE WHERE TABLE_SCHEMA IN ('testcase') AND REFERENCED_COLUMN_NAME IS NOT NULL
creating a new table: user
[...]

PR #986, opened 2017-10-03
Fix for #983

iz-iznogood:
MariaDB reports default value as "NULL" sometimes (string with "NULL" literal).
MySQL does the right thing and has null for default

This is a small workaround only for the MariaDB case


pleerock:
hmm strange that mariadb has such issues

iz-iznogood:
I will look deepee, I found it strange too. Also there was no reference on the web for this

ckalogeropoulos:
Hi @pleerock it is strange but this is what happens

```sql
ALTER TABLE `user` ADD COLUMN `col1` INT NULL DEFAULT NULL AFTER `bio`
or
ALTER TABLE `user` ADD COLUMN `col1` INT NULL AFTER `bio`
```

If you run do 
```sql
SELECT * FROM INFORMATION_SCHEMA.COLUMNS WHERE TABLE_SCHEMA IN ('test') 
```
this reports the column col1 with default value 'NULL' literal

If you do 
```sql
ALTER TABLE `user` ADD COLUMN `col2` INT(11) NOT NULL AFTER `col1`
```
and run the query again this comes back as null (not literal)

This is not related to the default value but to the nullable flag
If you ask the column to be required even if you don't specify the default value it still comes back as 'NULL'. 

I can make the workaround even stricter if you prefer by taking into account the required flag.

Thanks

Issue #1002, opened 2017-10-05
PrimaryGeneratedColumn('uuid') is not working


maddindeiss:
If I mark my id property in my entity as a `@PrimaryGeneratedColumn('uuid')`  it is set to `AUTOINCREMENT`. 
This will result in a SQLError:
```
query failed:  CREATE TABLE "project" ("id" varchar NOT NULL PRIMARY KEY AUTOINCREMENT, ...
```
```
SQLError {code: 5, message: "could not prepare statement (1 AUTOINCREMENT is only allowed on an INTEGER PRIMARY KEY)"}
```

But my id is type `string` because of the UUID.

```javascript
@Entity()
export class Project {
    @PrimaryGeneratedColumn('uuid')
    id: string;
}
```

So the example in the Documentation won't work: 
[http://typeorm.io/#/decorator-reference/primarygeneratedcolumn](http://typeorm.io/#/decorator-reference/primarygeneratedcolumn)

Am I doing something wrong or is this a bug?

And thanks for the great work!

daniel-lang:
Hey, what database system are you using?

pleerock:
I guess its something related to sqlite since "AUTOINCREMENT" is sqlite-specific keyword.

daniel-lang:
@maddindeiss I created a [test case](https://github.com/daniel-lang/typeorm/tree/issue-1002/test/github-issues/1002) which works on postgres, mariadb, sqlite and mysql. It simply inserts a an entity with a uuid primary column.

Is your use-case more complex? What are the steps to reproduce this error?

maddindeiss:
@daniel-lang You are right. I was using the `websql` driver and got that SQLError.
I switched over to the `cordova` driver and everything works fine now.

So maybe this is just a `websql` problem.

daniel-lang:
Maybe, I will have to check

daniel-lang:
Yes, it's a problem with the websql driver, #1003 contains a fix

pleerock:
thanks to @daniel-lang this can be closed now

maddindeiss:
@daniel-lang thanks for fixing the issue

PR #1003, opened 2017-10-05
Issue #1002: @PrimaryGeneratedColumn("uuid") in WebSQL

daniel-lang:
This is a test for #1002.
On my local machine this test runs without problems in mysql, mariadb, sqlite and postgres. So it's either something different or a different database.


pleerock:
yeah we already have tests that check `@PrimaryGeneratedColumn("uuid")` functionality in the codebase. Probably author have this issue after certain actions, maybe he is updating exist table

daniel-lang:
Okay, I will leave this test until we know more about the issue.

daniel-lang:
fixes #1002

WebSQL should use `buildCreateColumnSql` from AbstractSqlQueryRunner and not implement it's own, since the abstract one already supports uuid columns

PR #1035, opened 2017-10-17
expression was missing mode

elthrasher:
Hi, I was trying to get replication working in mysql and saw every request was going to slave. I believe this will fix that issue.


pleerock:
Looks like a proper fix, thank you for your contribution!

PR #1061, opened 2017-10-21
Cordova: PRAGMA  foreign_keys (#1053)

### LINKED ISSUE #1053: Ionic Foreign Keys
Hi,
I am creating a ManyToOne relation :
```
@Entity()
export class User{
    @PrimaryGeneratedColumn()
    Id: number;

    @ManyToOne(type => Workplace)
    Workplace: Workplace;
}

@Entity()
export class Workplace extends BaseEntity {
   @PrimaryGeneratedColumn()
    Id: number;
}
```

Then I insert workplaces in database. After that I try to insert a user with a workplace.Id that doesn't exist for workplace table. The insertion is done althought the foreign key einserted is not correct:
` let user: User = new User();
const workplace = new Worplace();
workplace.Id = 143214321432
        user.Workplace = workplace;
userRepository.save(user);`

I think that in sqlite databases, foreign keys are disabled by default in order to achieve retrocompatibility. So, in your sqlite drivers you run PRAGMA foregn_keys=ON, in order to enable foreign keys. This isn't done in your cordova driver. If I execute that sentence through ORM connection, foreign keys are enabled and user with wrong workplace isn't inserted in database:
` await this.connection.manager.query("PRAGMA foreign_keys=ON;");`

Could you add this to your cordova drivers? Would it work for WebSQL also?

daniel-lang:
Up to now, foreign keys where disabled in the cordova driver. This enables them by default.
I don't know if this has to be considered a breaking change, since but to know the cordova driver didn't enforce foreign key constraints.

Fixes #1053

### DIFF
--- a/src/driver/cordova/CordovaDriver.ts
+++ b/src/driver/cordova/CordovaDriver.ts
@@ -74,7 +74,14 @@ export class CordovaDriver extends AbstractSqliteDriver {
         return new Promise<void>((ok, fail) => {
             this.sqlite.openDatabase({name: this.options.database, location: this.options.location}, (db: any) => {
                 const databaseConnection = db;
-                ok(databaseConnection);
+
+                // we need to enable foreign keys in sqlite to make sure all foreign key related features
+                // working properly. this also makes onDelete to work with sqlite.
+                databaseConnection.executeSql(`PRAGMA foreign_keys = ON;`, [], (result: any) => {
+                    ok(databaseConnection);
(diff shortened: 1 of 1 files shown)


daniel-lang:
I forgot to add a line to the changelog. Is that necessary in your opinion?

pleerock:
I think its not mandatory because of popularity of cordova driver, but you can add it, as you wish

PR #1064, opened 2017-10-22
add missing link to slides

EreckGordon:
http://typeorm.io/#/tree-entities has the slides, but http://typeorm.io/#/entities does not.


pleerock:
Thank you!

PR #1143, opened 2017-11-05
explicit require() statements for the drivers that call them

EreckGordon:
This allows us to bundle into a server.js with webpack. I only needed it for Postgres for myself, but figured I should make it available for any driver that is calling PlatformTools.load().


pleerock:
Hi and thank you for your contribution. Did you check all usages of the load method? tslint is failing because of broken code style. Also, mkdirp and path are under "sqlite" comment section which is not correct

EreckGordon:
The Sqlite driver seems to call both mkdirp and path with PlatformTools.load(), so I added them to the switch case. https://github.com/typeorm/typeorm/blob/master/src/driver/sqlite/SqliteDriver.ts#L121 I would be happy to remove these from the driver switch statement.

I'll fix the style to conform to tslint rules, sorry about that.

The other potential problem with this implementation is it does give warnings about missing node_modules for every other type of driver. I could see that being irritating for users, (even though it is harmless) but did not know if you would prefer this type of switch to PlatformTools, or switching from the use of PlatformTools for require() statements in each individual driver.

```
WARNING in ./node_modules/typeorm/platform/PlatformTools.js
Module not found: Error: Can't resolve 'mongodb' in 'C:\Users\Ereck\Desktop\angular-universal-pwa-starter-project\angula
r-universal-pwa-starter\node_modules\typeorm\platform'
 @ ./node_modules/typeorm/platform/PlatformTools.js 38:27-45
 @ ./node_modules/typeorm/entity-manager/MongoEntityManager.js
 @ ./node_modules/typeorm/index.js
 @ ./server/database-connection.ts
 @ ./server/server.ts

WARNING in ./node_modules/typeorm/platform/PlatformTools.js
Module not found: Error: Can't resolve 'mssql' in 'C:\Users\Ereck\Desktop\angular-universal-pwa-starter-project\angular-
universal-pwa-starter\node_modules\typeorm\platform'
 @ ./node_modules/typeorm/platform/PlatformTools.js 78:27-43
 @ ./node_modules/typeorm/entity-manager/MongoEntityManager.js
 @ ./node_modules/typeorm/index.js
 @ ./server/database-connection.ts
 @ ./server/server.ts

WARNING in ./node_modules/typeorm/platform/PlatformTools.js
Module not found: Error: Can't resolve 'mysql' in 'C:\Users\Ereck\Desktop\angular-universal-pwa-starter-project\angular-
universal-pwa-starter\node_modules\typeorm\platform'
 @ ./node_modules/typeorm/platform/PlatformTools.js 43:27-43
 @ ./node_modules/typeorm/entity-manager/MongoEntityManager.js
 @ ./node_modules/typeorm/index.js
 @ ./server/database-connection.ts
 @ ./server/server.ts

WARNING in ./node_modules/typeorm/platform/PlatformTools.js
Module not found: Error: Can't resolve 'mysql2' in 'C:\Users\Ereck\Desktop\angular-universal-pwa-starter-project\angular
-universal-pwa-starter\node_modules\typeorm\platform'
 @ ./node_modules/typeorm/platform/PlatformTools.js 45:27-44
 @ ./node_modules/typeorm/entity-manager/MongoEntityManager.js
 @ ./node_modules/typeorm/index.js
 @ ./server/database-connection.ts
 @ ./server/server.ts

WARNING in ./node_modules/typeorm/platform/PlatformTools.js
Module not found: Error: Can't resolve 'oracledb' in 'C:\Users\Ereck\Desktop\angular-universal-pwa-starter-project\angul
ar-universal-pwa-starter\node_modules\typeorm\platform'
 @ ./node_modules/typeorm/platform/PlatformTools.js 50:27-46
 @ ./node_modules/typeorm/entity-manager/MongoEntityManager.js
 @ ./node_modules/typeorm/index.js
 @ ./server/database-connection.ts
 @ ./server/server.ts

WARNING in ./node_modules/typeorm/platform/PlatformTools.js
Module not found: Error: Can't resolve 'redis' in 'C:\Users\Ereck\Desktop\angular-universal-pwa-starter-project\angular-
universal-pwa-starter\node_modules\typeorm\platform'
 @ ./node_modules/typeorm/platform/PlatformTools.js 64:27-43
 @ ./node_modules/typeorm/entity-manager/MongoEntityManager.js
 @ ./node_modules/typeorm/index.js
 @ ./server/database-connection.ts
 @ ./server/server.ts

WARNING in ./node_modules/typeorm/platform/PlatformTools.js
Module not found: Error: Can't resolve 'sqlite3' in 'C:\Users\Ereck\Desktop\angular-universal-pwa-starter-project\angula
r-universal-pwa-starter\node_modules\typeorm\platform'
 @ ./node_modules/typeorm/platform/PlatformTools.js 69:27-45
 @ ./node_modules/typeorm/entity-manager/MongoEntityManager.js
 @ ./node_modules/typeorm/index.js
 @ ./server/database-connection.ts
 @ ./server/server.ts
Listening at http://localhost:8000
```

pleerock:
>  I would be happy to remove these from the driver switch statement.

No, I meant to simply extract them into separate code block, something like "other modules"

pleerock:
> The other potential problem with this implementation is it does give warnings about missing node_modules for every other type of driver. I could see that being irritating for users, (even though it is harmless) but did not know if you would prefer this type of switch to PlatformTools, or switching from the use of PlatformTools for require() statements in each individual driver.

If you have those errors... then Im not sure if this fix really fixes something...

pleerock:
adding @daniel-lang

EreckGordon:
You could switch https://github.com/typeorm/typeorm/blob/master/src/driver/postgres/PostgresDriver.ts#L587 for `this.postgres = require("pg");` and leave PlatformTools alone entirely, and this would be compatible with webpack, but for purely dynamic import statements, it blows up.

daniel-lang:
I didn't know, that webpack doesn't like these require statements. I agree, that the error about the missing node package isn't every nice but the only way I can think of, to fix this would be to exclude drivers and make them seperate node.js packages.
@pleerock some time ago we agreed to split the codebase into browser and node version and remove/change PlatformTools. Iguess this is another argument why we have to do this.

pleerock:
> @pleerock some time ago we agreed to split the codebase into browser and node version and remove/change PlatformTools. Iguess this is another argument why we have to do this.

I don't think it will work since people are using node version with webpack, not a browser version

pleerock:
> The split would happen for driver and platform specific code. So that Node.js packages can be used directly in the drivers.

this will make things much more complicated, I would like avoid doing that only because of this webpack issue. 

What if users just ignore those packages, maybe using [this](https://webpack.github.io/docs/list-of-plugins.html#ignoreplugin) ?

EreckGordon:
@daniel-lang yeah, you are right, I was too early in suggesting just moving the requires to the driver -- I still get the same webpack warnings about missing drivers.

@pleerock I tried ignore plugin and it caused errors, however, the FilterWarningsPlugin allowed me to just suppress the warnings without affecting the code. I think this is the winner.

```
const FilterWarningsPlugin = require('webpack-filter-warnings-plugin');

//in plugins array:
//ignore the drivers you don't want, in this case everything but the postgres drivers.
    new FilterWarningsPlugin({
    	exclude: [/mongodb/, /mssql/, /mysql/, /mysql2/, /oracledb/, /redis/, /sqlite3/]
    })
```

pleerock:
@daniel-lang what do you think about FilterWarningsPlugin?

daniel-lang:
Well, it's not the most user friendly solution but if you don't want to split we have to relay the configuration to the user and let them map to the correct version and suppress warnings.

pleerock:
Split will make things much more complicated something I would like to avoid as much as possible. Having to split because of this particular issue seems worthless to me.

So if you don't have alternatives I suggest to accept this PR once it gets finished and add to "FAQ" receipt how to configure typeorm with webpack.

daniel-lang:
I don't have any alternatives

docs/faq.md
```diff
@@ -179,4 +180,22 @@ Also if you want to use the ts-node CLI you can execute TypeORM the following wa
...
+
+```js
+const FilterWarningsPlugin = require('webpack-filter-warnings-plugin');
+
+module.exports = {
+    ...
+    plugins: [
+        //ignore the drivers you don't want. This is the complete list of all drivers -- remove the suppressions for drivers you want to use.
+        new FilterWarningsPlugin({
+            exclude: [/mongodb/, /mssql/, /mysql/, /mysql2/, /oracledb/, /pg/, /pg-native/, /pg-query-stream/, /redis/, /sqlite3/]
```
pleerock: what about mkdirp other non-driver modules? Don't we need them?

EreckGordon: Those loaded for me no problem, no errors. I don't know if they actually were needed for the switch statement, but it felt safer to put them in because PlatformTools.load() was getting called to load them. Loaded because they are standard nodejs modules rather than add ons like the drivers.

pleerock: Hmm strange you didnt had errors with `mkdirp`, its not a standard nodejs module....

pleerock: ah, okay so we have this package in dependencies, so you have it installed unlike driver packages

pleerock:
Looks good to me. @daniel-lang please review it and if you think its a good to go and you don't have anything to add then I'll merge it.

daniel-lang:
@pleerock good to go, I will create a seperate PR that removes the flag from package.json so that webpack can be easily used with the node.js version and not only with the browser version.

pleerock:
@daniel-lang thank you! @EreckGordon thank you for your contribution!

Vandivier:
@EreckGordon @daniel-lang 

"The other potential problem with this implementation is it does give warnings about missing node_modules for every other type of driver. I could see that being irritating for users, (even though it is harmless)"

Yes, it is annoying. It makes debugging a pain. Is there an issue tracking silencing for that?

daniel-lang:
@EreckGordon mentioned the `FilterWarningsPlugin` plugins. Did you give that a try?

Vandivier:
tldr; either it doesn't integrate with my project or, if it does, I need instruction on how to do so.

just above in the issue discussion history you can see 'Vandivier added a commit to Vandivier/ng-express-cms that referenced this pull request 12 hours ago.' Treat that repo as a repro.

It uses Angular Universal, which means it has tsconfig.json but webpack is not actually exposed.

Earlier than that you can see 'Vandivier referenced this pull request in angular/angular-cli.' There I was asking them about exposing webpack, something they currently don't want to do. There is talk about it being planned for Angular CLI 2, but I don't know the timeline or seriousness of that talk.

daniel-lang:
According to [this comment](https://github.com/typeorm/typeorm/issues/1290#issuecomment-354574475) it is currently not possible use TypeORM and Angular CLI together.
Angular Universal might work because it doesn't use the browser but rather the node version. But, as you pointed you, Angular CLI doesn't expose webpack configs, so you cannot use the mentioned plugin.

Vandivier:
Understood. I have segregated the API server from the rendering server. I'll proceed with TypeORM integration on the Express API server that doesn't depend on Angular CLI and provide an update here with how it goes.

PR #1146, opened 2017-11-06
fixed bad query when using take on entity with multiple pk

elthrasher:
Both conditions and parameters were invalid. Try running my tests against the old code. These changes fix the problem - though I'm not sure if they match the original intent.


pleerock:
Looks proper, thank you for your contribution

PR #1183, opened 2017-11-15
added IRepository interface

KyleGalvin:
Added IRepository allowing TypeMoq to automagically create mock objects for test harnesses

### DIFF
--- a/src/repository/Repository.ts
+++ b/src/repository/Repository.ts
@@ -347,3 +347,52 @@ export class Repository<Entity extends ObjectLiteral> {
     }
 
 }
+
+export interface IRepository<Entity extends ObjectLiteral> {
+    readonly manager: EntityManager;
+    readonly metadata: EntityMetadata;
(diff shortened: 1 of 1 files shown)


src/repository/Repository.ts
```diff
@@ -12,7 +12,7 @@ import {SelectQueryBuilder} from "../query-builder/SelectQueryBuilder";
 /**
  * Repository is supposed to work with your entity objects. Find entities, insert, update, delete, etc.
  */
-export class Repository<Entity extends ObjectLiteral> {
+export class Repository<Entity extends ObjectLiteral> implements IRepository<Entity> {
```
pleerock: why do you need it? Why you simply don't do `implements Repository` where `Repository` is a class? (in typescript you can implement classes as well)

KyleGalvin: @pleerock you just made my life a little bit easier.

Cheers!

PR #1202, opened 2017-11-20
Update working-with-entity-manager.md

babbarankit:
Typo Mistake, first parameter was missing. Also, have used findOne instead of findOneById.


pleerock:
Thank you very much! I returned `findOne` because `master` branch is `0.1.x` which has `findOneById`. This change must be in the `next` branch (and it has it already I think)

PR #1208, opened 2017-11-21
Additional findOptions - range and where OR

radarsu:
Sending findOptions like:
```
{ where: { property: ['X', 'Y'] } }
```
Will cause "or" select.

Select range
```
{ "where": { "amount": { "range": [10, 20] } } }
```
Select >=
```
{ "where": { "amount": { "range": [10, null] } } }
```
Select <=
```
{ "where": { "amount": { "range": [null, 10] } } }
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,5 @@
 {
-  "name": "typeorm",
-  "private": true,
+  "name": "radarsu-typeorm",
   "version": "0.2.0-alpha.6",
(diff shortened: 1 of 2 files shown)


pleerock:
Hi and thank you for your contribution. Unfortunately I cannot accept this PR since we are going to implement those features with a bit different design, other from what you did. See #1101 for more info.

Issue #1211, opened 2017-11-21
CLI Invalid Command Help


patrickhousley:
When using the CLI, if an invalid command is entered, nothing happens. Can you please update the CLI to print the help screen or an invalid command message when this happens?

mrkmg:
Just made a PR for both master and next

pleerock:
@mrkmg thank you very much, this fix will be published in `0.1.7` maybe next week

PR #1216, opened 2017-11-22
Fixed issue #1211

mrkmg:
Updated demand to demandCommand (demand is depreciated)
Added "strict"


pleerock:
Thank you very much!

PR #1218, opened 2017-11-22
Fixed broken test - test/github-issues/1034/issue-1034.ts

mrkmg:
test/github-issues/1034/issue-1034.ts was using a removed method


pleerock:
I merged and it still fails... Are you able to investigate why?

PR #1222, opened 2017-11-23
feature: findOneOrFail

### LINKED ISSUE #1149: Feature: findOneOrFail
Continuing on the discussion started in #122 .
Add a method `findOneOrFail` which returns a `Promise<T>` instead of a `Promise<T|undefined>` which is what `findOne` returns. The promise is rejected in case no result was found.

**motivation**: Helps to focus on the happy path when writing controller code (rejected promises naturally tend to trickle down into a single or a few error handlers which is less cumbersome than manually handling undefined values).

**additional consideration**: Expose a conventional error format to help users distinguish this kind of error from other errors. For instance, an HTTP framework could recognize the error and spit out a 404 instead of a 500 which it would normally produce.

I'll try to come up with a PR some time soon.

hmil:
Close #1149 

Seems that simple to implement.

I'm not quite satisfied with testing though:
- I could not find the test for `findOne` on `EntityManager` to add the tests for `findOneOrFail`
- Also I'm not sure copy/pasting the tests for `findOne` is the right thing to do. With pure unit tests I would simply assert that `manager.findOneOrFail` calls `manager.findOne` and does the right thing if it gets `undefined` in return. Then I would test that `repository.findOneOrFail` forwards the call to `manager.findOneOrFail`. These tests are obviously higher level and I don't have enough knowledge of this project to know what's right to do.

### DIFF
--- a/src/entity-manager/EntityManager.ts
+++ b/src/entity-manager/EntityManager.ts
@@ -1,6 +1,7 @@
 import {Connection} from "../connection/Connection";
 import {FindManyOptions} from "../find-options/FindManyOptions";
 import {ObjectType} from "../common/ObjectType";
+import { EntityNotFoundError } from "../error/EntityNotFoundError";
 import {QueryRunnerProviderAlreadyReleasedError} from "../error/QueryRunnerProviderAlreadyReleasedError";
 import {FindOneOptions} from "../find-options/FindOneOptions";
 import {DeepPartial} from "../common/DeepPartial";
@@ -536,6 +537,33 @@ export class EntityManager {
         }
 
         return qb.getOne();
+    }
+
+    /**
+     * Finds first entity that matches given find options or rejects the returned promise on error.
+     */
+    findOneOrFail<Entity>(entityClass: ObjectType<Entity>|string, id?: string|number|Date|ObjectID, options?: FindOneOptions<Entity>): Promise<Entity>;
+
+    /**
+     * Finds first entity that matches given find options or rejects the returned promise on error.
+     */
+    findOneOrFail<Entity>(entityClass: ObjectType<Entity>|string, options?: FindOneOptions<Entity>): Promise<Entity>;
+
+    /**
+     * Finds first entity that matches given conditions or rejects the returned promise on error.
+     */
+    findOneOrFail<Entity>(entityClass: ObjectType<Entity>|string, conditions?: DeepPartial<Entity>, options?: FindOneOptions<Entity>): Promise<Entity>;
+
+    /**
+     * Finds first entity that matches given conditions or rejects the returned promise on error.
+     */
+    findOneOrFail<Entity>(entityClass: ObjectType<Entity>|string, idOrOptionsOrConditions?: string|string[]|number|number[]|Date|Date[]|ObjectID|ObjectID[]|FindOneOptions<Entity>|DeepPartial<Entity>, maybeOptions?: FindOneOptions<Entity>): Promise<Entity> {
+        return this.findOne(entityClass, idOrOptionsOrConditions as any, maybeOptions).then((value) => {
+            if (value === undefined) {
+                return Promise.reject(new EntityNotFoundError());
+            }
+            return Promise.resolve(value);
+        });
     }
 
     /**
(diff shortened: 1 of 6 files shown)


src/error/EntityNotFoundError.ts
```diff
@@ -0,0 +1,13 @@
+/**
+ * Thrown when no result could be found in methods which are not allowed to return undefined or an empty set.
+ */
+export class EntityNotFoundError extends Error {
+    name = "EntityNotFound";
+
+    constructor() {
+        super();
+        Object.setPrototypeOf(this, EntityNotFoundError.prototype);
+        this.message = `No such entity.`;
```
pleerock: We need to provide a better message. Something like `No entity "${entityClass.constructor.name}" was found by following criteria: "${idOrOptionsOrConditions}` ?

src/entity-manager/EntityManager.ts
```diff
@@ -538,6 +539,33 @@ export class EntityManager {
...
+
+    /**
+     * Finds first entity that matches given conditions or rejects the returned promise on error.
+     */
+    findOneOrFail<Entity>(entityClass: ObjectType<Entity>|string, idOrOptionsOrConditions?: string|string[]|number|number[]|Date|Date[]|ObjectID|ObjectID[]|FindOneOptions<Entity>|DeepPartial<Entity>, maybeOptions?: FindOneOptions<Entity>): Promise<Entity> {
+        return this.findOne(entityClass, idOrOptionsOrConditions as any, maybeOptions).then((value) => {
+            if (value === undefined) {
+                return Promise.reject(new EntityNotFoundError());
+            }
+            return Promise.resolve(value);
```
pleerock: you can simply `return value;`

pleerock:
> I could not find the test for findOne on EntityManager to add the tests for findOneOrFail

not everything covered, maybe this method is out of coverage

> These tests are obviously higher level and I don't have enough knowledge of this project to know what's right to do.

all tests are kinda functional/integration tests. I don't like to test if one thing simply calls another thing until its something complicated and cannot be easily covered with functional tests. So, in your case you can call `manager.findOneOrFail` and `repository.findOneOrFail` and check results of both methods instead of checking if one of them (or both) call some another method, I don't find it better then actual results checking.

src/entity-manager/EntityManager.ts
hmil: In this instance I cannot because I'm returning a rejected Promise in the first branch and TypeScript expects the return type to be homogeneous (either all promise or all direct value). Also I don't want to go async/throw for this very simple use case.

pleerock: ah okay

src/entity-manager/EntityManager.ts
```diff
@@ -538,6 +539,35 @@ export class EntityManager {
...
+
+    /**
+     * Finds first entity that matches given conditions or rejects the returned promise on error.
+     */
+    findOneOrFail<Entity>(entityClass: ObjectType<Entity>|string, idOrOptionsOrConditions?: string|string[]|number|number[]|Date|Date[]|ObjectID|ObjectID[]|FindOneOptions<Entity>|DeepPartial<Entity>, maybeOptions?: FindOneOptions<Entity>): Promise<Entity> {
+        return this.findOne(entityClass, idOrOptionsOrConditions as any, maybeOptions).then((value) => {
+            if (value === undefined) {
+                return Promise.reject(new EntityNotFoundError(
+                        (typeof entityClass === "string") ? entityClass : entityClass.constructor.name,
+                        JSON.stringify(idOrOptionsOrConditions)));
```
hmil: I have to stringify this parameter otherwise the error message could show a frustrating `[Object object]`. Can you make sure this is safe? If there is a circular object here we end up throwing an error while attempting to create another error :/

hmil:
> all tests are kinda functional/integration tests. I don't like to test if one thing simply calls another thing until its something complicated and cannot be easily covered with functional tests. So, in your case you can call manager.findOneOrFail and repository.findOneOrFail and check results of both methods instead of checking if one of them (or both) call some another method, I don't find it better then actual results checking.

So you mean what I did here is OK ?

src/entity-manager/EntityManager.ts
pleerock: > Can you make sure this is safe?

Not really sure... We can either wrap into try/catch or show `idOrOptionsOrConditions` only if its a primitive.

btw move all boilerplate code into `EntityNotFoundError` to make this code cleaner.

pleerock:
> So you mean what I did here is OK ?

yeah

pleerock:
Looks like there are issues with test:

```
 repository > find methods findOneOrFail should return entity by a given id and find options:
EntityNotFound: Could not find any entity of type "User" matching: 1
      at new EntityNotFoundError (src/error/EntityNotFoundError.ts:10:9)
      at src/entity-manager/EntityManager.ts:573:39
      at <anonymous>
```

since this must fail:

```ts
            loadedUser = await userRepository.findOneOrFail(1, {
                where: {
                    secondName: "Dorian"
                }
            });
            expect(loadedUser).to.be.undefined;
```

Now it can be `undefined` if `findOneOrFail` is used

src/error/EntityNotFoundError.ts
danhvv-1254: @pleerock please help me. i print entityClass from exception (EntityNotFoundError) but it is undefined.
` return this.userRepository.findOneOrFail(id);`

Issue #281, opened 2017-02-12
Table Names using Underscores instead of Continuous Name


rightisleft:
The latest alpha 3 is incorrectly changing class names into multiple underscores:

Database: Postgres
OS: Mac OS X

```typescript
@Table()
export class DTODocument {
  @PrimaryGeneratedColumn() id: number;
}
```

Expected - dto_document OR DTODocument
Actual -  public | d_t_o_document       | table | rightisleft

pleerock:
Looks like a small bug. You are able to fix it [there](https://github.com/typeorm/typeorm/blob/master/src/util/StringUtils.ts) ?

valeriy-devpronet:
what if use [lodash.snakeCase](https://github.com/lodash/lodash/blob/4.17.4/lodash.js#L14542)  or [lodash.camelCase](https://github.com/lodash/lodash/blob/4.17.4/lodash.js#L14117) or this is unexpected dependecies?

pleerock:
no, typeorm should not have external dependecies like this because its also packed for browser where user will require to install lodash separately to make it to work. Thats why we need our own function.

pleerock:
thanks to @mrkmg this is fixed and will be released in next 0.2.0-alpha version (`npm i typeorm@next`)

hi2u:
@rightisleft when you mentioned...

> Expected - dto_document OR DTODocument

Do you know if there's some way to switch TypeORM between using camelCase and under_scores ?

I'm trying to figure out how to use underscores in foreign key column names: https://github.com/typeorm/typeorm/issues/2608 ... not sure if my issue is related.

PR #1229, opened 2017-11-24
Fix Issue 281

mrkmg:
Fix Issue #281 by updating regex

Test is here: https://regex101.com/r/QeSm2I/1

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -41,6 +41,7 @@ Use `findOne(id)` method instead now.
 * removed deprecated `removeById`, `removeByIds` methods
 * removed `deleteById` method - use `delete(id)` method instead now
 * removed `updateById` method - use `update(id)` method instead now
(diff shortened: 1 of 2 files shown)


mrkmg:
Obviously a change like this will cause new tables to be created if the string changed.

Not sure how we want to handle rolling this out, as anyone with a db will need to manually rename their tables to keep their data.

pleerock:
Yes its a breaking change, please update a changelog.

also, can you please take a look why tests are failing?

mrkmg:
The failing tests appears to be an issue in `next` on its own. Looking into the issue now.

mrkmg:
Tests Failing

Issue: https://github.com/typeorm/typeorm/issues/1242
PR: https://github.com/typeorm/typeorm/pull/1243

pleerock:
Thank you!

PR #1234, opened 2017-11-26
Support for persistence operations without wrapped transactions

goulash1971:
Hi Guys

Firstly, congratulations on typeOrm (coming from an enterprise java background, I appreciate the effort that has gone into putting this together).

The pull request covers some changes that enables the 'save' and 'remove' persistence operations to be performed without wrapping transactions so that the transaction scope can be managed outside of typeOrm (whilst still using the typeOrm interfaces into the underlying persistence store).

The need comes from a system that I am working on where the transaction model includes various application level guards, and transaction state is distributed across a bunch of underlying persistence stores.

I originally made the mods against 0.1.6.0 (happy to provide these to you if you would like) but thought that it would be better to redo them against the 0.2.0 branch given the amount of work you guys have done on the persistence layer.

For my sample tests, using 2000 double entity inserts I get a 50% performance increase in my use cases because where I am handling the transaction state out-of-line of the persistence operations.  Obviously this is very particular to my use case, however by allowing the 'save' and 'remove' operations to be (optionally) performed without a transaction wrapper (as per the low level 'insert', 'update', 'delete' etc. operations) I suspect others will get performance boosts.

One thing to note, at 0.2.0 the 'save' operation (on MySQL) there is an 'INSERT' operation and a 'SELECT' operation at 0.2.0, whilst at 0.1.6.0 there was only an 'INSERT' operation.  As a result the time for the 'save' operation nearly doubles (in my tests the average time for 2000 entity inserts was 4792ms at 0.1.6.0 which increases to 8301ms at 0.2.0).

Obviously you are doin this 'SELECT' operation for a good reason :)

Do you think it is possible to optionally overrode the 'SELECT' (which I assume is an entity reload) through a save option?  In may cases I imagine that the need for an entity reload is not required or, the reload has already been taken account of in client code.

I will have a look  to see if I can find the place where the reload occurs myself and propose a pull-request, but it could be something you might want to consider when you look at 0.2.0 or beyond.

Once again, thanks for all of the efforts on this.

Cheers
Stuart

### DIFF
--- a/src/entity-manager/EntityManager.ts
+++ b/src/entity-manager/EntityManager.ts
@@ -322,8 +322,7 @@ export class EntityManager {
     const options = target ? maybeOptions : maybeEntityOrOptions as SaveOptions;
 
     // execute save operation
-    await new EntityPersistExecutor(this.connection, this.queryRunner, "save", target, entity, options)
-      .executeNoTransaction();
+    await new EntityPersistExecutor(this.connection, this.queryRunner, "save", target, entity, options).executeNoTransaction();
     return entity;
   }
 
@@ -394,8 +393,7 @@ export class EntityManager {
     const options = target ? maybeOptions : maybeEntityOrOptions as SaveOptions;
 
     // execute save operation
-    await new EntityPersistExecutor(this.connection, this.queryRunner, "remove", target, entity, options)
-      .executeNoTransaction()
+    await new EntityPersistExecutor(this.connection, this.queryRunner, "remove", target, entity, options).executeNoTransaction();
     return entity;
   }
 
--- a/src/entity-manager/EntityManager.ts
+++ b/src/entity-manager/EntityManager.ts
@@ -1,36 +1,36 @@
-import {Connection} from "../connection/Connection";
-import {FindManyOptions} from "../find-options/FindManyOptions";
-import {ObjectType} from "../common/ObjectType";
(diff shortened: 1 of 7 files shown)


goulash1971:
I've added in something that allows the reload to be optional and the time to perform the 1000 inserts drops back to what it was at 0.1.6.0.

As I mentioned in my earlier comment, there obviously are good reasons for the reload of the entity therefore I have set it so that the default is to perform the reload i.e. the reloadEntity option must be set to false to explicitly stop the reload.

I've also fixed a problem I introduced that made the Travis build fall over.

pleerock:
Hi and thank you for your contribution.

Please take a look on your code in the "files changed" tab. Obviously its not possible for me to review this PR, but its very important for me to review each line you changed in details because you changed very very sensitive part of ORM. Remove custom code style formatting you applied and make incremental and clear changes which I'll be able to review. 

Also, please add tests which test functionality you have added.

Thank you for making this PR against 0.2.0 it was a good decision.

pleerock:
in the last commit to next branch I added `listeners: boolean` and `chunk?: number` flags to `SaveOptions`, so I decided to implement this as well. 

Because of change diff I don't know how did you implement transaction disabling, but I added a fag to save/remove options called `transaction: boolean`.

I also applied your changes on `reloadEntity` but called it simply `reload`

goulash1971:
Thanks @pleerock that looks great :)

pleerock:
If that satisfied you I think this PR can be closed now. Thanks for your efforts

Issue #1245, opened 2017-11-28
`findByIds` ignores `FindManyOptions`


mskv:
`manager.findByIds(Entity, ids, { where: { property: "value" } })`

results in `WHERE IN (...)` without `property = 'value'`.

I'll submit a PR with a fix soon.

mskv:
Here's the PR:
https://github.com/typeorm/typeorm/pull/1246

PR #1246, opened 2017-11-28
[issue-1245] Use `andWhere` instead of `where` in `findByIds`

mskv:
should fix https://github.com/typeorm/typeorm/issues/1245


pleerock:
Thank you for your contribution!

PR #1269, opened 2017-11-29
test case for issue #1268 (sqlite insert with default columns)

### LINKED ISSUE #1268: SQLite: INSERT with columns that have default values
When `repository.insert()` is used on an entity that includes columns with default values, the insert will fail.

**Example code**
Entity
```typescript
@Entity()
export class User {
    @PrimaryGeneratedColumn()
    id: number;

    @Column()
    name: string;

    @Column({default: true})
    isAdmin: boolean;
}
```
Insert
```typescript
const repository = connection.getRepository(User);

let user1 = new User();
user1.name = "Daniel Lang";

let user2 = new User();
 user2.name = "Daniel Kurz";
user2.isAdmin = false;

let user3 = new User();
user3.name = "Daniel Lang3";
user3.isAdmin = true;

await repository.insert([user1, user2, user3]);
```

This will generate the following statement
```sql
INSERT INTO "user"("name", "isAdmin") VALUES ($1,NULL), ($2,$3), ($4,$5)
-- PARAMETERS: ["Daniel Lang","Daniel Kurz",0,"Daniel Lang3",1]
-- SQLITE_CONSTRAINT: NOT NULL constraint failed: user.isAdmin
```

When using `.save()` each entity is saved seperately and there the default column is left with default is left out:
```sql
INSERT INTO "user"("name") VALUES ($1) -- PARAMETERS: ["Daniel Lang"]
INSERT INTO "user"("name", "isAdmin") VALUES ($1,$2) -- PARAMETERS: ["Daniel Kurz",0]
INSERT INTO "user"("name", "isAdmin") VALUES ($1,$2) -- PARAMETERS: ["Daniel Lang3",1]
```
Fixing this won't be easy because entities have to be seperated into those that have the default value set and those that don't.

daniel-lang:
still don't how to fix #1268, but this is a test case for it

### DIFF
--- a/test/github-issues/1268/default.ts
+++ b/test/github-issues/1268/default.ts
@@ -1,0 +1,38 @@
+import "reflect-metadata";
+import {expect} from "chai";
+import {closeTestingConnections, createTestingConnections, reloadTestingDatabases} from "../../utils/test-utils";
+import {Connection} from "../../../src/connection/Connection";
+import {User} from "./entity/User";
+
+describe("github issues > #1268 sqlite batch insert with default", () => {
+
+    let connections: Connection[];
+    before(async () => connections = await createTestingConnections({
+        entities: [__dirname + "/entity/*{.js,.ts}"]
+    }));
+    beforeEach(() => reloadTestingDatabases(connections));
+    after(() => closeTestingConnections(connections));
+
+    it("should save entities where default value is not specified", () => Promise.all(connections.map(async connection => {
+
+        const repository = connection.getRepository(User);
+
+        let user1 = new User();
+        user1.name = "Daniel Lang";
+
+        let user2 = new User();
+        user2.name = "Daniel Kurz";
+        user2.isAdmin = false;
+
+        let user3 = new User();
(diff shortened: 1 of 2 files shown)


pleerock:
Oh I see you already added it, but how it passes if it should fail since its should work only in `next` ?

daniel-lang:
It passed because Travis CI for some reason doesn't fail when a test for sql.js fails. If you check the log you will see, that it failed.
This PR is targeted for `master`. Therefore I will close it and reopen one for `next`.

pleerock:
Okay. We'll need to fix sqljs issue as well.

pleerock:
@daniel-lang any updates

daniel-lang:
I haven't yet managed to reproduce the error of the test not failing locally, so I couldn't investigate it.

daniel-lang:
These tests fail on master as this feature is only supported on `next`!

pleerock:
hmm so we need to remove those changes from master?

daniel-lang:
I guess so, yes

pleerock:
can you do that?

Issue #1282, opened 2017-11-30
FEATURE REQUEST - Naming strategy joinTableColumnName has to know if it is called from the owning or owned (inverse) context


risa:
Hello,

I have a legacy database I need to work with that uses slightly odd naming conventions for M:N relation. Example:

```shell
+-------------+--------------+----------------------------+
|                prefix__Question                         |
+-------------+--------------+----------------------------+
| id          | int(11)      | PRIMARY KEY AUTO_INCREMENT |
| title       | varchar(255) |                            |
+-------------+--------------+----------------------------+

+-------------+--------------+----------------------------+
|                prefix__Category                         |
+-------------+--------------+----------------------------+
| id          | int(11)      | PRIMARY KEY AUTO_INCREMENT |
| name        | varchar(255) |                            |
+-------------+--------------+----------------------------+

+-------------+--------------+----------------------------+
|           prefix__Question__categories                  |
+-------------+--------------+----------------------------+
| slot        | int(11)      | PRIMARY KEY FOREIGN KEY    |
| col         | int(11)      | PRIMARY KEY FOREIGN KEY    |
| item        | int(11)      | PRIMARY KEY FOREIGN KEY    |
+-------------+--------------+----------------------------+
```
The ```slot``` is autogenerated ID, ```col``` refers to ```Category.id``` and ```item``` refers to ```Question.id```

I do not need to insert data (yet). I am trying to build a custom naming strategy for these join tables.
For the ```joinTableColumnName``` I need to know if it is called in the junctionColumns context (i.e. the owning part
of the relationship) or the inverseJunctionColumns (i.e. the owned part of the relationship).

I have prepared a small fix which adds the fourth parameter to the joinTableColumnName called inverse. If it is called
from the owned part its value is false, if from the inverse part it is true. This allowed me to prepare a naming strategy
which works also with my legacy database structure.

Trivial tests are included too.

Please let me know if this PR is sane and even possibly acceptable.

Thank you

Richard

pleerock:
Closing since feature was already merged.

PR #1283, opened 2017-11-30
Naming strategy joinTableColumnName has to know if it is called from the owning or owned (inverse) context

risa:
PR for feature request https://github.com/typeorm/typeorm/issues/1282

Including simple tests.

### DIFF
--- a/src/naming-strategy/DefaultNamingStrategy.ts
+++ b/src/naming-strategy/DefaultNamingStrategy.ts
@@ -60,8 +60,12 @@ export class DefaultNamingStrategy implements NamingStrategyInterface {
         return columnName + "_" + index;
     }
 
-    joinTableColumnName(tableName: string, propertyName: string, columnName?: string, inverse?: boolean): string {
+    joinTableColumnName(tableName: string, propertyName: string, columnName?: string): string {
         return camelCase(tableName + "_" + (columnName ? columnName : propertyName));
+    }
+
+    joinTableInverseColumnName(tableName: string, propertyName: string, columnName?: string): string {
+        return this.joinTableColumnName(tableName, propertyName, columnName);
     }
 
     foreignKeyName(tableName: string, columnNames: string[], referencedTableName: string, referencedColumnNames: string[]): string {
--- a/src/naming-strategy/DefaultNamingStrategy.ts
+++ b/src/naming-strategy/DefaultNamingStrategy.ts
@@ -60,7 +60,7 @@ export class DefaultNamingStrategy implements NamingStrategyInterface {
         return columnName + "_" + index;
     }
 
-    joinTableColumnName(tableName: string, propertyName: string, columnName?: string): string {
+    joinTableColumnName(tableName: string, propertyName: string, columnName?: string, inverse?: boolean): string {
         return camelCase(tableName + "_" + (columnName ? columnName : propertyName));
     }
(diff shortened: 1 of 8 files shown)


src/naming-strategy/DefaultNamingStrategy.ts
```diff
@@ -60,7 +60,7 @@ export class DefaultNamingStrategy implements NamingStrategyInterface {
         return columnName + "_" + index;
     }
 
-    joinTableColumnName(tableName: string, propertyName: string, columnName?: string): string {
+    joinTableColumnName(tableName: string, propertyName: string, columnName?: string, inverse?: boolean): string {
```
pleerock: Maybe its better to create separate `joinTableInverseColumnName` method instead?

test/github-issues/1282/issue-1282.ts
```diff
@@ -0,0 +1,39 @@
...
+
+        // joinTbelColumnName was called at least twice
+        expect(namingStrategy.calls.length).greaterThan(1);
+
+        // the first call was with inverse=false - i.e. for the owning part of the relation
+        expect(namingStrategy.calls[0]).to.be.false;
+
+        // the second call was with invers=true - i.e. for the owned part of the relation
+        expect(namingStrategy.calls[1]).to.be.true;
+
```
pleerock: probably you can check how your database column are named in the database this way:

```ts
connection.getManyToManyMetadata(Animal, "categories")
     .columns.find(column => column.databaseName === "...")
     .should.not.be.undefined;
```

src/naming-strategy/DefaultNamingStrategy.ts
risa: That would definitely be more transparent, I just did not want to introduce new methods. But if that is OK with you then I am going to change it.

test/github-issues/1282/issue-1282.ts
risa: Looks better, I will give it a try. Thanks.

src/naming-strategy/DefaultNamingStrategy.ts
pleerock: yeah lets change it

risa:
@pleerock, sorry for the delay - I have updated the PR to include both functions as you suggested:
- ```joinTableColumnName``` and
- ```joinTableInverseColumnName```
Could you please have a look at the PR? 

Thanks


Richard

pleerock:
thank you, what about this my comment:

> probably you can check how your database column are named in the database this way:
```ts
connection.getManyToManyMetadata(Animal, "categories")
     .columns.find(column => column.databaseName === "...")
     .should.not.be.undefined;
```

risa:
Sorry, I forgot about this one, I will try it later today ...

risa:
Updated the testing logic. Unfortunately I could not make the call as elegant as you proposed because there is strictNullChecks=true parameter in tsconfig.json and the getEntityMetadata returns a potentially undefined value, so typescript compiler kept bothering me until I made the explicit assignment with if. If you know a more elegant way please leave me a comment. Thanks!

pleerock:
That's fine, thank you very much!

pleerock:
Will be released in 0.1.8 probably next week, but let me know if you need it super urgently

PR #1280, opened 2017-11-30
Webpack: Remove automatic selection of browser version

daniel-lang:
This will finally support node projects that use webpack. I also put together the needed changed to make it work in Ionic over at the [example](https://github.com/typeorm/ionic-example#typeorm--018).

fixes #1009

PR #1287, opened 2017-12-01
Improves docs for webpack changes and new find option

daniel-lang:
@pleerock, do you think this would be enough?

related to #1280


pleerock:
Thank you, its definitely better then nothing, but it can be improved more later.

PR #1302, opened 2017-12-05
New DebugLogger

pravdomil:
## Motivation
<a href="https://github.com/visionmedia/debug">The debug package</a> is the <a href="https://www.npmjs.com/browse/depended">9th most dependent package</a>, it's used by  express, koa, yeoman, babel, mocha, socket.io, eslint, morgan and <a href="https://www.npmjs.com/browse/depended/debug">more</a>. It works in both node.js and browser.
https://github.com/typeorm/typeorm/pull/634 https://github.com/typeorm/typeorm/issues/633

## What's new
- `ConnectOptions.logger` has new value `"debug"`
- New debug namespaces introduced:
```
typeorm:query
typeorm:queryError
typeorm:querySlow
typeorm:schemaBuild
typeorm:migration
typeorm:log
typeorm:info
typeorm:warn
```

## Drawbacks
- Debug dependency (debug itself has only one dependency, module ms)
- `ConnectOptions.logging?: LoggerOptions` is no longer used.
You need set env variable `DEBUG=typeorm:*` or call `debug.enable('typeorm:*');` in order to set log level.

## Benefits
- TypeORM log fits nicely next to others.
- You can start debugging every aspect of application by setting `DEBUG=*` and `logger: "debug"`.

## Example
Here's rest api example, I got a get request and then it reports which query TypeORM executed.

api

## Discussion
- Use it as default logger and deprecate `logging` option and control everything via debug lib. That's basically how express does logging.
- Documentation changes

### DIFF
--- a/src/logger/DebugLogger.ts
+++ b/src/logger/DebugLogger.ts
@@ -17,10 +17,6 @@ export class DebugLogger implements Logger {
     private debugInfo = debug("typeorm:info");
     private debugWarn = debug("typeorm:warn");
     
-    constructor() {
-        
-    }
-    
     /**
      * Logs query and parameters used in it.
      */
--- a/src/logger/DebugLogger.ts
+++ b/src/logger/DebugLogger.ts
@@ -1,6 +1,6 @@
-import { Logger } from "./Logger";
-import { QueryRunner } from "../";
-import { PlatformTools } from "../platform/PlatformTools";
+import {Logger} from "./Logger";
+import {QueryRunner} from "../";
+import {PlatformTools} from "../platform/PlatformTools";
 
 const debug = PlatformTools.load("debug");
 
--- a/src/logger/DebugLogger.ts
+++ b/src/logger/DebugLogger.ts
@@ -1,7 +1,8 @@
-import * as debug from "debug";
 import { Logger } from "./Logger";
 import { QueryRunner } from "../";
 import { PlatformTools } from "../platform/PlatformTools";
+
+const debug = PlatformTools.load("debug");
 
 /**
  * Performs logging of the events in TypeORM via debug library.
--- a/src/logger/DebugLogger.ts
+++ b/src/logger/DebugLogger.ts
@@ -26,7 +26,7 @@ export class DebugLogger implements Logger {
      */
     logQuery(query: string, parameters?: any[], queryRunner?: QueryRunner) {
         if (this.debugQuery.enabled) {
-            this.debugQuery(PlatformTools.highlightSql(query));
+            this.debugQuery(PlatformTools.highlightSql(query) + ";");
             if (parameters && parameters.length) {
                 this.debugQuery("parameters:", parameters);
             }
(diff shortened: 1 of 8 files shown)


src/logger/DebugLogger.ts
```diff
@@ -0,0 +1,103 @@
+import * as debug from "debug";
```
pleerock: we need to dynamically load any third-party package using `require` from `PlatformTools` class. Its requirement for typeorm to be cross-platform

src/logger/DebugLogger.ts
```diff
@@ -0,0 +1,103 @@
...
+    private debugQueryError = debug("typeorm:queryError");
+    private debugQuerySlow = debug("typeorm:querySlow");
+    private debugSchemaBuild = debug("typeorm:schemaBuild");
+    private debugMigration = debug("typeorm:migration");
+    
+    private debugLog = debug("typeorm:log");
+    private debugInfo = debug("typeorm:info");
+    private debugWarn = debug("typeorm:warn");
+    
+    constructor() {
```
pleerock: remove it

src/logger/DebugLogger.ts
```diff
@@ -0,0 +1,103 @@
...
+import { Logger } from "./Logger";
+import { QueryRunner } from "../";
+import { PlatformTools } from "../platform/PlatformTools";
+
+/**
+ * Performs logging of the events in TypeORM via debug library.
+ */
+export class DebugLogger implements Logger {
+    private debugQuery = debug("typeorm:query");
+    private debugQueryError = debug("typeorm:queryError");
```
pleerock: > typeorm:queryError

queryError looks rough, is there naming convention in `debug`? Maybe better name it `typeorm:query:error` or `typeorm:query-error` ?

pleerock:
Thank you for your contribution, can you please update documentation and changelog for the next version as well

src/logger/DebugLogger.ts
pravdomil: https://github.com/visionmedia/debug#conventions

src/logger/DebugLogger.ts
pravdomil:
```ts
import * as path from "path";

/**
 * Platform-specific tools.
 */
export class PlatformTools {
    ...
		
    /**
     * Debug module needed by DebugLogger
     */
    static debug = debug;
}
```
like that? what do you prefer?

pravdomil:
what is the next version number?

pleerock:
0.1.8

src/logger/DebugLogger.ts
pleerock: no, just see how many other libraries like mkdirp, redis, mysql, etc. are loaded using PlatformTools, its dynamic loading using `require` function

pravdomil: ahh you meant `PlatformTools.load`

pravdomil:
maybe we can change query debug namespace to
```
typeorm:query:log
typeorm:query:error
typeorm:query:slow
```

Issue #1147, opened 2017-11-06
How to use LIKE in findAndCount


Flexiink:
<pre>
 const sqlConditions = {
            take: limit,
            skip: offset,
            where: 'name LIKE "%1323%"'
 };
 const [ items, count ] = await this._getRepository().findAndCount(sqlConditions);
</pre>

The code works correctly, but Webpack generates an error

<pre>
ERROR in ../categories.ts (96,75): Type '{ take: any; skip: number; where: string; }' has no properties in common with type 'DeepPartial<Categories>'
</pre>

Began to receive after update **Typescript** the top up to **2.4.2**

pleerock:
right now `where` accepts object, not a string.

so what you are attempting to do is not by a design. But we probably can change this behaviour and support a string as well.

For now you must use query builder instead.

Flexiink:
How can I extend this class so that I do not get an error?

<pre>
export interface FindOneOptions<Entity> {
    where?: Partial<Entity> | ObjectLiteral;
}
</pre>
to <code> where?: Partial<Entity> | ObjectLiteral **| string**; </code>

pleerock:
Just change this interface and create a PR and I'll merge it.

pleerock:
this must be already released in 0.1.3

brendan-roche:
@pleerock this fix was never applied. Please see
https://github.com/typeorm/typeorm/pull/1307

mantiqh:
We can doing it something like this.

```
async findAll(
    query: FindAllCategoriesQueryDto,
  ): Promise<Paginated<EcostoreCategory>> {
    const [result, total] = await this.categoryRepo.findAndCount({
      take: query.limit,
      skip: query.limit * (query.page - 1),
      select: ['id', 'image', 'description', 'name', 'active'],
      where: {
        active: query.active,
        name: query.name && ILike(`%${query.name}%`),
      },
    });

    return new Paginated<EcostoreCategory>({
      page: query.page,
      result,
      total,
    });
  }
```

PR #1307, opened 2017-12-05
Issues #1147

brendan-roche:
This issue was closed, but this fix was never merged to master. 
Also would be nice to get merged back to `next` soon too please

### DIFF
--- a/src/entity-manager/MongoEntityManager.ts
+++ b/src/entity-manager/MongoEntityManager.ts
@@ -517,7 +517,14 @@ export class MongoEntityManager extends EntityManager {
         if (!optionsOrConditions)
             return undefined;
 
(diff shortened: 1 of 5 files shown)


pleerock:
typescript compilation failed

brendan-roche:
Yeah saw this after I pushed. I thought it was simple fix but fails because this `query` parameter is passed through to the mongo driver on collection.find - MongoQueryRunner.ts:102, which does not accept a string for an sql query.

http://mongodb.github.io/node-mongodb-native/2.1/api/Collection.html#find

So an option to get around this would be to ignore the string and pass an empty object instead, as mongo is not a sql database. 

Would you be happy with this solution?

pleerock:
yeah looks like we have no option if we want to add a string condition

brendan-roche:
@pleerock ready to review again

pleerock:
Thank you

PR #1320, opened 2017-12-08
Fix MissingDriverError

hmil:
`MissingDriverError`'s message incorrectly listed the available drivers.

Instead of writing a regression test I factored out the list of available drivers and added a compile-time check which ensures the list is correct. The list can even be reused for maximum DRYness should one need the list of drivers somewhere else in the code.


pleerock:
emm... I don't know how to express my feelings about this code... but what this code really adds? What is the purpose of `__ensureConnectionTypesIsCorrect` ? Also, driver type already have a compile time check, I can't understand what this code adds

hmil:
Yeah sorry for not leaving enough details. Let me rephrase what I did here:

**tl;dr** _Creates a list of available drivers `ConnectionTypes` which is **guaranteed** to be correct thanks to the utility method `__ensureConnectionTypesIsCorrect`. This does not affect the public API. You trade verbosity for long-term maintainability._

**long version**
The class `MissingDriverError` was giving a misleading error message because it showed a wrong list of available drivers. Now one way to fix this is to just add the missing drivers to the list and be done with it. However this has two drawbacks:
1. No one will ever think of changing that piece of code when they add or remove a driver which means the bug is bound to come back.
2. If another piece of code needs the list, then you will duplicate the hard-coded list and the problem in 1. increases exponentially.

The clean way to fix this bug is to define a list of available drivers somewhere and derive the error message from that list. This solves 2. because now we can re-use the list wherever we need it.

However, in order to address 1. we need a way to enforce that the list is consistent with the available drivers. One way to achieve this is by letting TypeScript inform us when the list is wrong. Here's where it gets more complex. The only type available which will tell us exactly which drivers are available is the field "type" in `ConnectionOptions`. This field does not have an explicit type defined, instead it uses the values aggregated by the union of types under `ConnectionOptions` (which is good because it's DRY 👍 ).
**In order to make sure the list has all and only values which are acceptable in `ConnectionOptions.type` is to force the TypeScript compiler to type check this list for us.**  
The function `__ensureConnectionTypesIsCorrect` is exactly this. It exploits the TypeScript compiler to make sure the list is accurate. Note that the function doesn't need to be called, we only need it during the compilation stage.

Give it a try: If you make a typo in the list or if you forget a value the code won't compile.

Now whenever we need a list of available drivers, we have one at disposal. Whenever we add, remove or rename a driver, we know this list will stay accurate.

---

### Other bits and pieces

To achieve the above, we need the `tuple` factory method to create accurately typed tuples. This is a very useful basic tool which can be used anywhere to increase type safety. It won't hurt to have it in the codebase. 😉 

Finally, if you look at the [DriverFactory](https://github.com/typeorm/typeorm/blob/4d7ea319ce04a22bc99e3231d87b70fe432d9283/src/driver/DriverFactory.ts#L22) there is a switch statement that goes over the value of interest. Wouldn't it be nice if typescript could tell us if we forgot something ? Say no more, I've got you! If you change the parameter of `MissingDriverError` to the type `never`, then typescript will tell you if you forgot a case in your switch statement. Effectively this means it is mathematically impossible to get an error message such as `Wrong driver: "mysql" given. Supported driver are: ..., "mysql", ...`
Give it a try: Remove a case statement and you'll see what I'm talking about.

---

I acknowledge it may seem over-engineered but consider the following: On such a large project there are so many things to do that you'll want to automate as much as possible. This is an instance of trading verbosity for maintainability. At the expense of a dozen lines of code you get in return the guarantee that some bug will never come back ever again and you get a reliable tool (`ConnectionTypes`) which you can safely reuse across the whole codebase with no additional maintenance cost.

pleerock:
Thank you very much for such detailed explanation.

I understood what problem this code tries to solve, however I don't think that having such approach is better then manually adding drivers each time. So I cannot accept this PR, Im very sorry. Programming is all about deciding among cons and pros as we know, and in this case I make my decision that its better to manually change few places rather then have this ugly code which brings complexity and requires good documentation to explain why we need it. We could make same comment and ask users to change x and y places and such comment could be much simpler then comment for this code

hmil:
No problem I get the tradeoff. Though you've got to admit the hack itself is pretty sick. 🤣

I'll push the clean fix then you can just squash and merge so there's no trace left of this "ugly" code. 😉

pleerock:
Thank you very much, really appreciate your efforts and contribution, merged

PR #1321, opened 2017-12-08
Enable foreign keys in sqljs

hmil:
This is mostly a copy-paste of the code in the `sqlite` driver into the `sqljs` driver.

Since SQL.js is a port of sqlite, I think they should behave the same as much as possible.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -84,9 +84,7 @@ By default its true.
 
 ## 0.1.3
 
-* added support for `
-
-`. To use it you just need to install `npm i sql.js` and use `sqljs` as driver type ([#894](https://github.com/typeorm/typeorm/pull/894)).
+* added support for `sql.js`. To use it you just need to install `npm i sql.js` and use `sqljs` as driver type ([#894](https://github.com/typeorm/typeorm/pull/894)).
 * added explicit require() statements for drivers ([#1143](https://github.com/typeorm/typeorm/pull/1143))
 * fixed bug where wrong query is generated with multiple primary keys ([#1146](https://github.com/typeorm/typeorm/pull/1146))
 * fixed bug for oracle driver where connect method was wrong ([#1177](https://github.com/typeorm/typeorm/pull/1177))
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -52,6 +52,7 @@ Use `findOne(id)` method instead now.
 * `@DiscriminatorColumn` decorator has been removed, use `@TableInheritance` options instead now
 * `skipSync` in entity options has been renamed to `synchronize`. Now if it set to false schema synchronization for the entity will be disabled.
 By default its true.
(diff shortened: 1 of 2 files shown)


src/driver/sqljs/SqljsDriver.ts
```diff
@@ -209,7 +209,14 @@ export class SqljsDriver extends AbstractSqliteDriver {
             this.databaseConnection = new this.sqlite.Database();
         }
 
-        return Promise.resolve(this.databaseConnection);
+        return new Promise<void>(async (ok, fail) => {
```
daniel-lang: `async` shouldn't be needed here, right?

hmil: Indeed

daniel-lang:
@pleerock PR is fine with me. Is the failing build related to this PR or a general problem? Also is a note in the changelog needed?

hmil:
I'm anticipating this calls for an entry in the changelog since it might break existing applications.

Also @daniel-lang if you've got time would you mind giving your opinion on #1320 ? I'm feeling guilty for the wall of text I left there especially since this is borderline bikeshedding.

daniel-lang:
Yes it might break them, but @pleerock  will have to decide if it's big/bad enough to include it

CHANGELOG.md
```diff
@@ -83,7 +84,9 @@ By default its true.
 
 ## 0.1.3
 
-* added support for `sql.js`. To use it you just need to install `npm i sql.js` and use `sqljs` as driver type ([#894](https://github.com/typeorm/typeorm/pull/894)).
+* added support for `
+
+`. To use it you just need to install `npm i sql.js` and use `sqljs` as driver type ([#894](https://github.com/typeorm/typeorm/pull/894)).
```
daniel-lang: What happened here?

hmil: What the hell? I used Github's web editor, no idea where this is coming from.

pleerock:
Looks good, thank you!

pleerock:
Is there `this.databaseConnection.exec` method in sqljs? For some reason those tests failed on next, not sure why they didn't on master. I replaced this code with the following:

```ts
        const queryRunner = this.createQueryRunner("master");
        await queryRunner.query(`PRAGMA foreign_keys = ON;`);
        return this.databaseConnection;
```

pleerock:
Also @daniel-lang I have one comment regarding to sqljs tests - you use `createConnection` and pass connection options which aren't taken from ormconfig.json which means if we do 'skip: false` it will not affect those tests

daniel-lang:
Yes, sqljs has `this.databaseConnection.exec(sqlquery)` which might be better in this case. I will create a PR with for it.

And for the sqljs tests: I used `createConnection` because otherwise I cannot pass options that aren't defined in `TestingOptions` to the driver.

pleerock:
> I used createConnection because otherwise I cannot pass options that aren't defined in TestingOptions to the driver.

what if we add those options to TestingOptions then?

Issue #1045, opened 2017-10-18
type "enum" does not exist


mkzaker:
Hi,
my database is postgres and i have this error 
QueryFailedError: type "enum" does not exist

pleerock:
enums are only supported in [mysql](https://github.com/typeorm/typeorm/blob/e7c516e7f4f162e1411d65a8aa30c43ef8b9220e/test/functional/database-schema/column-types/mysql/entity/Post.ts#L83). We are planning to add postgres enum support in 0.2.0 . Duplicate of #671

pleerock:
I expect it in December

RDeluxe:
Hello, any news about this ? i'm currently starting a new project with TypeORM and I would like to avoid hacking around if the new version is not far away

pleerock:
Work is in progress, we are expecting to land 0.2.0 in the beginning of January

RDeluxe:
Hi,

Any news about this ? :)

pleerock:
Hey Ive been very busy last two months, so 0.2.0 milestone is moved to March.

AlexMesser:
fixed and released in `typeorm@0.2.0-alpha.23`. You can try it via `npm i typeorm@next`.

PR #1325, opened 2017-12-10
enum type currently not supported in postgres

jakoblorz:
As issue #1045 points out, typeorm does not support `enum` on postgres type, yet it is referenced in the docs as supported.

### DIFF
--- a/docs/entities.md
+++ b/docs/entities.md
@@ -248,7 +248,7 @@ or
 `character varying`, `character`, `varchar`, `char`, `int2`, `integer`, `int4`, `int8`, 
 `float4`, `float8`, `smallserial`, `serial2`, `serial`, `serial4`, `bigserial`, `serial8`, 
 `money`, `boolean`, `bool`, `text`, `citext`, `bytea`, `date`, `interval`, `point`, `line`, `lseg`, `box`, 
-`circle`, `path`, `polygon`, `enum`, `cidr`, `inet`, `macaddr`, `bit`, `bit varying`,
+`circle`, `path`, `polygon`, `cidr`, `inet`, `macaddr`, `bit`, `bit varying`,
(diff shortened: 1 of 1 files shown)


pleerock:
did you check if this really brings enum support into postgres?

jakoblorz:
Hahaha :laughing: this pr is **not** supposed to add enum support, I just discovered that the docs document its support which I think is wrong:
> *from /docs/entities.md*
> ### Column types for `postgres`
> `int2`, `int2`, `int4`, `int8`, `integer`, `smallint`, `bigint`, `decimal`, `numeric`, `decimal`, 
> `numeric`, `real`, `double precision`, `time`, `time with time zone`, `time without time zone`,
> `timestamp`, `timestamp without time zone`, `timestamp with time zone`, `int`, `smallint`, `bigint`,
> `character varying`, `character`, `varchar`, `char`, `int2`, `integer`, `int4`, `int8`, 
> `float4`, `float8`, `smallserial`, `serial2`, `serial`, `serial4`, `bigserial`, `serial8`, 
> `money`, `boolean`, `bool`, `text`, `citext`, `bytea`, `date`, `interval`, `point`, `line`, `lseg`, `box`, 
> `circle`, `path`, `polygon`, `enum`, `cidr`, `inet`, `macaddr`, `bit`, `bit varying`,
> `varbit`, `tsvector`, `tsquery`, `uuid`, `xml`, `json`, `jsonb` 

I myself expected that enums are implemented using postgres due to the docs but then discovered that there is no support!

**this pr removes the wrong mention of enum support for postgres from the docs; it does not implement enums for postgres**

Sorry for explaining this pr insufficient!

pleerock:
haha sorry I misunderstood, I thought this PR brings postgres enum support to typeorm 😆  Thank you for your contribution, I'll accept this PR. We are planning to add enum postgres support in 0.2.0 which is going to be released in a month or two.

PR #1338, opened 2017-12-13
sql.js: enable foreign keys and change tests

daniel-lang:
Follow up to #1321 as that PR went into `next` instead of master.
Also @pleerock's comment about sql.js tests is adressed.
These changes will have to go into `next` as well

### DIFF
--- a/src/driver/sqljs/SqljsDriver.ts
+++ b/src/driver/sqljs/SqljsDriver.ts
@@ -182,7 +182,16 @@ export class SqljsDriver extends AbstractSqliteDriver {
             this.databaseConnection = new this.sqlite.Database();
         }
 
-        return Promise.resolve(this.databaseConnection);
+        // Enable foreign keys for database
+        return new Promise<any>((ok, fail) => {
+            try {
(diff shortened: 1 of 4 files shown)


src/driver/sqljs/SqljsDriver.ts
```diff
@@ -182,7 +182,16 @@ export class SqljsDriver extends AbstractSqliteDriver {
             this.databaseConnection = new this.sqlite.Database();
         }
 
-        return Promise.resolve(this.databaseConnection);
+        // Enable foreign keys for database
+        return new Promise<any>((ok, fail) => {
+            try {
+                this.databaseConnection.exec(`PRAGMA foreign_keys = ON;`);
```
pleerock: its not a promise, right?

daniel-lang: sqljs is synchronous, sadly

pleerock:
Can you please merge master into the `next` as well? I guess you'll have conflicts with my changes

daniel-lang:
Another test that is marked as "passed" even though in reality it failed

PR #1341, opened 2017-12-13
reverted sqlite batch insert tests (#1269)

daniel-lang:
They won't work on `master` only on `next`


pleerock:
Thank you very much

PR #1344, opened 2017-12-14
bug(require/browser): debug package was required outside of class

### LINKED ISSUE #1336: Browsersupport broken in 0.1.8
Typeorm broke support for usage in the browser with webpack as bundler, even with using the `NormalModuleReplacementPlugin`.

The following error is displayed:

```
Uncaught Error: This option/function is not supported in the browser environment. Failed operation: require("debug").
    at Function.PlatformTools.load (PlatformTools.js:23)
    at Object.<anonymous> (DebugLogger.js:2)
    at __webpack_require__ (bootstrap 3533b48…:19)
    at Object.<anonymous> (OffsetWithoutLimitNotSupportedError.js:26)
    at __webpack_require__ (bootstrap 3533b48…:19)
    at Object.<anonymous> (ConnectionManager.js:62)
    at __webpack_require__ (bootstrap 3533b48…:19)
    at Object.<anonymous> (bundle.js:51464)
    at __webpack_require__ (bootstrap 3533b48…:19)
    at Object.<anonymous> (inherits_browser.js:23)
```

My guess is that this has to do with introducing the new DebugLogger in 7b98e216.

I think the relevant line leading to this problem is 

```
const debug = PlatformTools.load("debug");
```

from commit 717e6540.

It was working fine in version 0.1.7.

daniel-lang:
The debug package was required outside of DebugLogger which broke the browser support. It has to be required inside the class to only be loaded when the logger is used (in node).

fixes #1336 
fixes typeorm/ionic-example#4

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ feel free to ask us and community.
 ## 0.1.10
 
 * `sqljs` driver now enforces FK integrity by default (same behavior as `sqlite`)
+* fixed issue that broke browser support in 0.1.8 because of the debug package ([#1344](https://github.com/typeorm/typeorm/pull/1344))
 
 ## 0.1.9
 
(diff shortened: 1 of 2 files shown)


pleerock:
Thank you very much! I'll publish new version tomorrow.

PR #1346, opened 2017-12-14
CI tests: fixed tests not failing when they should

daniel-lang:
This PR is just for me to try and reproduce and fix CI not marking builds with sqljs tests that failed as failed


daniel-lang:
@pleerock I'm not quiet sure why the extra call to the istanbul reporter is in the gulpfile but it seems like something is wrong with it as you can see in the first four tests where a failing test didn't mark the whole test as failed. The second to last test fails correctly while the last one only contains the correct task.
If you don't want to have all the commits in the project I can create a new PR with just the two changes that matter

pleerock:
I don't know. Isn't `writeReports` suppose to save reports into a `coverage` directory?

daniel-lang:
It has something to do with test coverage as the output we had, that reported the coverage at the end is gone.

pleerock:
hmmm okay...

daniel-lang:
Should I try to include it again and make it work?

pleerock:
yeah we need it

pleerock:
it also creates a "coverage" directory with reports saved in html files

PR #1347, opened 2017-12-14
Changed ColumnMode to ColumnCategory

mrkmg:
While looking through the src code, saw a todo to rename mode to something else. Seeing as its a type, but that would be confusing due to type being the DB data type, I figured "category" would be a good alternative.

Feel free to reject the new name. Just couldn't think of a better name myself


pleerock:
Sorry but I don't think `Category` is better naming then `mode`. The best word is `type` here but we know that we cannot use it because `type` for column means `column type`. Mode is better word here because it indicates that this column works in a specific `mode` while category more associates like this column belongs to some category. Thank you for your efforts!

PR #1374, opened 2017-12-21
fix #1373 - check require('mysql') call

### LINKED ISSUE #1373: mysql2 not `require`d properly in jest
Yet another issue with using typeorm x Jest!

It appears that Jest is creating a mock for `mysql` when I did not install `mysql` and want to use `mysql2` instead. Not only that but it waits until **after** the first attempt to require `mysql` to create the mock! This leads to a crazy heisenbug where the first time one calls `createConnection`, typeorm uses the `mysql2` driver, and the second time it uses the bogus `mysql` driver and crashes!

This issue can be worked around by adding an additional check to make sure we don't get garbage in return of the call to `require('mysql')`.

hmil:
#1373

### DIFF
--- a/src/driver/mysql/MysqlDriver.ts
+++ b/src/driver/mysql/MysqlDriver.ts
@@ -530,6 +530,13 @@ export class MysqlDriver implements Driver {
     protected loadDependencies(): void {
         try {
             this.mysql = PlatformTools.load("mysql");  // try to load first supported package
+            /*
+             * Some frameworks (such as Jest) may mess up Node's require cache and provide garbage for the 'mysql' module
+             * if it was not installed. We check that the object we got actually contains something otherwise we treat
+             * it as if the `require` call failed.
+             *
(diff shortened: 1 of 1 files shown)


src/driver/mysql/MysqlDriver.ts
```diff
@@ -530,7 +530,9 @@ export class MysqlDriver implements Driver {
     protected loadDependencies(): void {
         try {
             this.mysql = PlatformTools.load("mysql");  // try to load first supported package
-
+            if (Object.keys(this.mysql).length === 0) {
+                throw new Error("'mysql' was found but it is empty. Falling back to 'mysql2'.");
```
hmil: The message is there for documentation purposes only. It is never shown to the user since the error gets caught right below.

pleerock: please add detailed comments on this code and explain why you did this and left github issue number.

PR #1422, opened 2018-01-08
Join eager relations of a specified relation

### LINKED ISSUE #1416: Loading a relation should load that relations eager relations
I have a case where there is a relationship between two entities (`User` and `Photo`). `Photo` meanwhile has an eager relationship with `PhotoMetadata`. If I run this query,  
```js
const user = userRepository.find({ where: { id: 1}, relations: ['photo'] })

// I should be able to access this
user.photo.photoMetadata
```

shouldn't the `Photo` that gets joined with `User` have `PhotoMetadata` eagerly loaded? Currently, the eager relations for the relation are ignored

BrainMaestro:
Fixes #1416

### DIFF
--- a/src/entity-manager/EntityManager.ts
+++ b/src/entity-manager/EntityManager.ts
@@ -25,7 +25,6 @@ import {RepositoryNotFoundError} from "../error/RepositoryNotFoundError";
 import {RepositoryNotTreeError} from "../error/RepositoryNotTreeError";
 import {RepositoryFactory} from "../repository/RepositoryFactory";
 import {TreeRepositoryNotSupportedError} from "../error/TreeRepositoryNotSupportedError";
-import {EntityMetadata} from "../metadata/EntityMetadata";
 import {QueryPartialEntity} from "../query-builder/QueryPartialEntity";
 import {EntityPersistExecutor} from "../persistence/EntityPersistExecutor";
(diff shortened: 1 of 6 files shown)


src/entity-manager/EntityManager.ts
```diff
@@ -496,7 +495,14 @@ export class EntityManager {
         const metadata = this.connection.getMetadata(entityClass);
         const qb = this.createQueryBuilder(entityClass, FindOptionsUtils.extractFindManyOptionsAlias(optionsOrConditions) || metadata.name);
         FindOptionsUtils.applyFindManyOptionsOrConditionsToQueryBuilder(qb, optionsOrConditions);
-        this.joinEagerRelations(qb, qb.alias, metadata);
+
+        ids = ids.map(id => {
```
pleerock: why do you need those changes? Looks like an old code from the `master` branch. Are you sure you did a proper branch from next and not from master?

pleerock:
Tests are failing as well. I think because of code from the master.

src/entity-manager/EntityManager.ts
BrainMaestro: Ah, rebase gone wrong. I branched from master at first, then rebased with next. Fixed

BrainMaestro:
All good now

pleerock:
Thank you very much!

PR #1423, opened 2018-01-09
Handle create extension exceptions for Postgres driver

### LINKED ISSUE #1407: Create extension query fails in Postrges connection for any users w/o superuser rights
I've found these little helpers to work with UUID and citext column types: 
https://github.com/typeorm/typeorm/blob/b96fc776cc37896b85cd3c21c2bc68655cc29bd9/src/driver/postgres/PostgresDriver.ts#L234
https://github.com/typeorm/typeorm/blob/b96fc776cc37896b85cd3c21c2bc68655cc29bd9/src/driver/postgres/PostgresDriver.ts#L236
 But in production database user doesn't have superuser rights to create any extensions in the database, so `executeQuery` will eventually fail. I'm wondering can these `executeQuery` calls be wrapped into a try-catch block and print some kind of warning message about required extensions in such situation?

maxmalov:
Fixes #1407

### DIFF
--- a/src/driver/postgres/PostgresDriver.ts
+++ b/src/driver/postgres/PostgresDriver.ts
@@ -229,12 +229,20 @@ export class PostgresDriver implements Driver {
             await Promise.all([this.master, ...this.slaves].map(pool => {
                 return new Promise((ok, fail) => {
                     pool.connect(async (err: any, connection: any, release: Function) => {
+                        const { logger } = this.connection;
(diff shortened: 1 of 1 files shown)


pleerock:
Thank you very much!

maxmalov:
np @pleerock, do you have any expectations when this fix will be released to npm?

pleerock:
Done. `0.1.11`

PR #1426, opened 2018-01-10
Specs to show one-to-one relation integrity isn't respected

dfournier:
Hi!

Before investing more time on a bugfix, do you think those specs are correct?

For relational databases, I want to add a `UNIQUE` constraint on the foreign key column in order to enforce `OneToOne` relation in both ways.

### DIFF
--- a/src/metadata-builder/EntityMetadataBuilder.ts
+++ b/src/metadata-builder/EntityMetadataBuilder.ts
@@ -108,10 +108,13 @@ export class EntityMetadataBuilder {
                 // create entity's relations join columns (for many-to-one and one-to-one owner)
                 entityMetadata.relations.filter(relation => relation.isOneToOne || relation.isManyToOne).forEach(relation => {
                     const joinColumns = this.metadataArgsStorage.filterJoinColumns(relation.target, relation.propertyName);
-                    const foreignKey = this.relationJoinColumnBuilder.build(joinColumns, relation); // create a foreign key based on its metadata args
+                    const { foreignKey, uniqueConstraint } = this.relationJoinColumnBuilder.build(joinColumns, relation); // create a foreign key based on its metadata args
                     if (foreignKey) {
                         relation.registerForeignKeys(foreignKey); // push it to the relation and thus register there a join column
                         entityMetadata.foreignKeys.push(foreignKey);
+                    }
+                    if (uniqueConstraint) {
+                        entityMetadata.uniques.push(uniqueConstraint);
                     }
                 });
 
--- a/src/metadata-builder/RelationJoinColumnBuilder.ts
+++ b/src/metadata-builder/RelationJoinColumnBuilder.ts
@@ -1,4 +1,5 @@
 import {ColumnMetadata} from "../metadata/ColumnMetadata";
+import {UniqueMetadata} from "../metadata/UniqueMetadata";
 import {ForeignKeyMetadata} from "../metadata/ForeignKeyMetadata";
 import {RelationMetadata} from "../metadata/RelationMetadata";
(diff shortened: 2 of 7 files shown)


pleerock:
no, one-to-one relations aren't necessary unique. You can have a single object shared between multiple users.

dfournier:
In that case it's a one-to-many?

pleerock:
what do you mean? Case you provided is not valid for both one-to-many and one-to-one relations

dfournier:
For a one-to-one relation, you shouldn't be able to have multiple access tokens for one user. If that were the case, it would be a many-to-one relation on the access token and one-to-many on the user.

At the moment, it doesn't appear that there is any distinction between the columns created for these two types of relation. This is what the second spec is testing for.

pleerock:
yes, looks like I was wrong. What is your suggestion? Add unique constraint on one-to-one relations?

dfournier:
Yes, I think it's a legitimated constraint to add 😄 

I also think it could be done on top of this PR

pleerock:
We can try to add `unique: relation.isOneToOne ? true : false` [here](https://github.com/pleerock/typeorm/blob/0884335a2dbdfd0794806ae600876a48244d873a/src/metadata-builder/RelationJoinColumnBuilder.ts#L127) and check if it does the trick

dfournier:
I push the fix you proposed to see if the CI is happy with that.

dfournier:
@pleerock sorry for the delay, I've been busy.

Setting unicity on the column works only when the table is joined with one column. That's why I add a unique index when the table is joined with multiple columns. Here is the created indexes on table `post` defined on the spec [`functional/relations/multiple-primary-keys/multiple-primary-keys-one-to-one`](https://github.com/typeorm/typeorm/tree/master/test/functional/relations/multiple-primary-keys/multiple-primary-keys-one-to-one):
screen shot 2018-03-04 at 22 21 27

What do you think of this solution so far?

Some specs are missing, the same kind I added (checking trying to link the same object twice with a `OneToOne` relation) but on `functional/relations/multiple-primary-keys/multiple-primary-keys-one-to-one`.

pleerock:
Can you please create this PR against `next` branch? We are actively working on `next` version and planning to release this month plus this change is kinda serious so it will be better if it land in `next`. Also in `next` we are supporting UNQIUE CONSTRAINTS (not indices, except for mysql where INDEX and UNIQUE CONSTRAINT is the same) so you can use `UniqueMetadata`

dfournier:
The branch has been rebased and  I used a UNIQUE CONSTRAINT instead of an INDEX. The CI doesn't pass but it looks like it's already broken on `next` branch. Don't hesitate to ping me back if there is anything I could do in order to see this PR live 😄

pleerock:
Lets merge it and see how it will work. Thank you very much for contribution!

pleerock:
@dfournier tests are failing currently on next and looks like because of this PR. Can you please take a look once? 

Why do you have `referencedColumns.length > 1 ` [here](https://github.com/typeorm/typeorm/blob/d3f2bd9b17ec32626c1773e710a413d3812e9971/src/metadata-builder/RelationJoinColumnBuilder.ts#L72)? Shouldn't it be more the zero?

dfournier:
Because otherwise the constraint is set on the column itself [here](https://github.com/typeorm/typeorm/blob/d3f2bd9b17ec32626c1773e710a413d3812e9971/src/metadata-builder/RelationJoinColumnBuilder.ts#L145).

pleerock:
thats quite confusing logic and it should be set only in one place, otherwise it will be source of problems.

dfournier:
I agree, this comes from the very first implementation. I create a fix now :)

Issue #1369, opened 2017-12-20
EntitySubscriber not firing events on abstract class entity


hvostoblud:
Having trouble to run code like this in my app with `"typeorm": "0.1.9"`. Please, check this one, mb i am doing something wrong:
```
import 'reflect-metadata';
import { createConnection, Connection, PrimaryGeneratedColumn, Column, Entity, EventSubscriber, EntitySubscriberInterface, InsertEvent, UpdateEvent } from "typeorm";
import * as path from 'path';
import * as fs from 'fs';

abstract class AbstractEntity {
    @PrimaryGeneratedColumn() id: number;
    @Column() firstname: string;
    @Column() lastname: string;
    @Column() fullname: string;
}

@Entity()
class ConcreteEntity extends AbstractEntity {
    @Column() description: string;
}

@EventSubscriber()
export class AbstractEntitySubscriber implements EntitySubscriberInterface<AbstractEntity> {
	listenTo() {
        return AbstractEntity;
	}
	async beforeInsert(event: InsertEvent<AbstractEntity>) {
		this.updateFullName(event.entity);
	}
	async beforeUpdate(event: UpdateEvent<AbstractEntity>) {
		this.updateFullName(event.entity);
	}
	updateFullName(o: AbstractEntity) {
		o.fullname = o.firstname + " " + o.lastname;
	}
}

const DATABASE_PATH = path.join(__dirname, 'test.db');
var dbExists = fs.existsSync(DATABASE_PATH);
if(dbExists) fs.unlinkSync(DATABASE_PATH);

(async () => { 

    const connection: Connection = await createConnection({
        type: 'sqlite',
        database: DATABASE_PATH,
        entities: [ AbstractEntity, ConcreteEntity ],
        subscribers: [ AbstractEntitySubscriber ],
        synchronize: true
    });

    var e = connection.manager.create(ConcreteEntity);
    e.description = "description";
    e.firstname = "firstname";
    e.lastname = "lastname";
    connection.manager.save(e);

})();
```
 This error is fired on this example `Error: SQLITE_CONSTRAINT: NOT NULL constraint failed: concrete_entity.fullname`

pleerock:
Looks like this case was not covered. `listenTo` isn't supposed to support abstract entities, only concrete ones. But we can fix it. If you are willing to contribute you can do it on your own, its not a hard fix, or at least provide a failing test and we can fix it very quickly.

pleerock:
Fixed and will be released in `0.1.12`.

PR #1432, opened 2018-01-10
Added check for subclasses in Broadcaster isAllowedSubscriber method

meel-io:
Added test for #1369
Added check for subclasses in Broadcaster isAllowedSubscriber method


pleerock:
Thank you very much!

pleerock:
Also would be great if you can update a changelog for the next version!

meel-io:
Sure, will do
Thank you for the great work!

Issue #1428, opened 2018-01-10
Migration not working on SQLITE at Ionic Framework


aluco100:
I'm trying two migration like this:

```
import {MigrationInterface, QueryRunner} from "typeorm";

export class ProductRefactoring1515586923234 implements MigrationInterface {

    public async up(queryRunner: QueryRunner): Promise<any> {
      await queryRunner.renameColumn("product","thumbnail","photo");
      // await queryRunner.query("ALTER TABLE 'product' RENAME 'thumbnail' to 'photo'");
    }

    public async down(queryRunner: QueryRunner): Promise<any> {
      await queryRunner.renameColumn("product","photo","thumbnail");
      // await queryRunner.query("ALTER TABLE 'product' RENAME 'photo' to 'thumbnail'");
    }

}
```
```
import {MigrationInterface, QueryRunner} from "typeorm";

export class ProductRefactoring1515606002592 implements MigrationInterface {

    public async up(queryRunner: QueryRunner): Promise<any> {
      await queryRunner.query("ALTER TABLE 'product' ADD COLUMN 'author' VARCHAR(50)");
    }

    public async down(queryRunner: QueryRunner): Promise<any> {
      await queryRunner.query("ALTER TABLE 'product' DROP COLUMN 'author'");
    }

}

```

Now I have that code for trigger them:

 ```
 public connection: Connection;

  async createSchema(): Promise<any> {
    return new Promise(async (fulfill, reject) => {
      await createConnection({
        type: 'cordova',
        database: 'test',
        location: 'default',
        logging: ['error', 'query', 'schema'],
        synchronize: true,
        entities: [
          Product
        ]
      }).then((connection) =>{
        this.connection = connection;
        fulfill();
      })
        .catch(error => reject(error));
    })

  }

  async runMigrations(){
    await this.connection.close();
    let migration = await createConnection({
      type: 'cordova',
      database: 'test',
      location: 'default',
      logging: ['error', 'query', 'schema'],
      synchronize: true,
      entities: [
        Product
      ],
      migrations: [ProductRefactoring1515586923234, ProductRefactoring1515606002592]
    }).then(async migration => {
      // await connection.runMigrations();
      // run all migrations
      await migration.runMigrations();

      // and undo migrations two times (because we have two migrations)
      await migration.undoLastMigration();
      await migration.undoLastMigration();

      console.log("Done. We run two migrations then reverted them.");
      await migration.close();
    });

    // await connection.runMigrations()
[...]

PR #1443, opened 2018-01-12
QueryRunner.renameColumn in Sqlite and Postgres

daniel-lang:
Because auf #1428 I looked at the sqlite code for `renameColumn` which didn't change anything, it just recreated the table. The fix that I wrote isn't the prettiest but I don't know of a better way right now.
Thanks to the test I also notices a syntax error in postgres when renaming columns.

It seems like this is an area (query runner methods that change the table) without many tests. Might be a good idea to add some more?!

### DIFF
--- a/src/driver/postgres/PostgresQueryRunner.ts
+++ b/src/driver/postgres/PostgresQueryRunner.ts
@@ -625,17 +625,18 @@ where constraint_type = 'PRIMARY KEY' AND c.table_schema IN (${schemaNamesString
         if (!oldColumn)
             throw new Error(`Column "${oldTableColumnOrName}" was not found in the "${tableOrName}" table.`);
 
-        if (this.connection.driver.createFullType(oldColumn) !== this.connection.driver.createFullType(newColumn) ||
-            oldColumn.name !== newColumn.name) {
-
+        if (this.connection.driver.createFullType(oldColumn) !== this.connection.driver.createFullType(newColumn)) {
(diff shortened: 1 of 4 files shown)


pleerock:
@AlexMesser works on query runners and their refactoring and test coverage several month already. Its going to be the biggest change of 0.2.0. If you aren't completely sure about this change its better not to merge it (but we still need this test for @AlexMesser) and wait for @AlexMesser to finish his work (planning to finish completely end of next month)

daniel-lang:
Okay, I didn't know that he is already working on this. So should I create a merge request for the test for next or just drop it completely?

pleerock:
first question is about this issue urgency. If its something very important we need to fix it. Second question is about this fix if we are confident with it (hard to say since there are no tests in this area). If we decide not to fix and merge this issue right now we can simply left this PR open and @AlexMesser can handle everything else later

daniel-lang:
> first question is about this issue urgency

I'm not sure about that. I only found this because of #1428. The SQLite fix is more a hack to get it working with the current system. The Postgres fix actually addresses the todo that was there, so it should be fine. If @AlexMesser has his changes ready for 0.2 in two weeks, I guess isn't not big deal to wait.

pleerock:
@daniel-lang can this PR be closed now?

daniel-lang:
I can't test it on `next` right now since the tests aren't passing

daniel-lang:
The tests that I included in this pr pass on `next` so it seems to be fixed.

Issue #1115, opened 2017-10-30
Performance problems when saving multiple documents


MaikelH:
### Problem
In our system we noticed slowdowns when we needed to insert a lot entities (up to 10.000). To see what the exact problem was we created a test script to test the insert speed of TypeORM. 

For the test we used the `save()` method inside a transaction ([see here](https://github.com/MaikelH/db-access-perf-tests/blob/master/src/tests/TypeormTest.ts)). In the screenshot below we can see the results of the test we ran with different number of documents.

![chart 1](https://user-images.githubusercontent.com/2207042/32171367-2ada6c88-bd78-11e7-995b-c3134cb8af39.png)

We compare against raw queries using `pg-promise`. The expected behavior is that TypeORM is slower than `pg-promise`, but should still scale linear. What can be seen in the graph is that TypeORM exhibits `O(n^2)` runtime complexity instead. Inserting 16000 documents takes about 7.5 minutes using TypeORM. 

The table below gives a more detailed view (times are in milliseconds), all the numbers are the average of 10 runs.

Documents | PG-Batch | TypeORM | PG-Promise | PG-InsertHelper
-- | -- | -- | -- | --
500 | 179 | 824 | 250 | 54
1000 | 268 | 1931 | 469 | 93
2000 | 476 | 12602 | 993 | 130
4000 | 863.1 | 58696 | 2015.7 | 216.5
8000 | 1618.9 | 116990.8 | 3886.1 | 385.9
16000 | 3281.2 | 438171.8 | 7706.2 | 736.8

What can be seen in the table is that when inserting a smaller amount of documents the performance is fine. 

After every test run the NodeJS process was restarted. If we do all the test runs in the same process (loop) the test results are slightly better on average but the `O(n^2)` runtime complexity is still there. See the google doc sheet `Machine 2 (continuous)` for the results.

### Solution
Currently not known.

### Environment
TypeORM: 0.1.1
Database: Postgres 9.6
Node: 6.11
OS: Ubuntu

Entity used:

```typescript
@Entity()
export class Document {
	@PrimaryColumn('text')
	id: string;

	@Column('text')
	docId: string;

	@Column('text')
	label: string;

	@Column('text')
	context: string;

	@Column({type: 'jsonb'})
	distributions: Distribution[];

	@Column({type: 'timestamp with time zone'})
	date: Date;
}

export interface Distribution {
	weight: string,
	id: number,
	docId: number
}
```

### More information
[...]

PR #1444, opened 2018-01-12
Optimize 'group' function in 'RawSqlResultsToEntityTransformer'

sviterok:
Hi @pleerock,
I think there is an opportunity to optimize `group` function that sits inside `RawSqlResultsToEntityTransformer` by switching from quadratic to the linear algorithm.

What do you think about the following fix?
I've got a quite significant performance improvement out of it on my local test (e.g. select of 140k records works ~10 seconds instead of 4 minutes).

### DIFF
--- a/src/query-builder/transformer/RawSqlResultsToEntityTransformer.ts
+++ b/src/query-builder/transformer/RawSqlResultsToEntityTransformer.ts
@@ -52,9 +52,9 @@ export class RawSqlResultsToEntityTransformer {
             const id = alias.metadata.primaryColumns.map(column => rawResult[alias.name + "_" + column.databaseName]).join("_"); // todo: check partial
             if (!id) return;
 
-            let group: { id: any, items: any[] }
+            let group: { id: any, items: any[] };
             if (id in groupedResults) {
-                group = groupedResults[id]
+                group = groupedResults[id];
             } else {
                 group = { id: id, items: [] };
             }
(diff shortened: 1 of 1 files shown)


src/query-builder/transformer/RawSqlResultsToEntityTransformer.ts
```diff
@@ -47,20 +47,24 @@ export class RawSqlResultsToEntityTransformer {
...
+            } else {
                 group = { id: id, items: [] };
-                groupedResults.push(group);
             }
 
             group.items.push(rawResult);
         });
-        return groupedResults.map(group => group.items);
+
+        return Object.keys(groupedResults).map(function(key, index) {
```
pleerock: Also when doing persistence optimizations I found out `Object.keys` is expensive operation. If you are able to test it with significantly more data (hundred of millions) I think you'll found Object.keys price as well. Alternative to this method is store keys in the array in a method above.

pleerock: also you can use regular `=>` function and you don't need an `index` here.

pleerock:
Thank you for this PR. `RawSqlResultsToEntityTransformer` is the place that needs optimization and thank you for putting efforts into fixing those issues. 

Please take a look why CI failed, looks like tests are not passing.

Also this PR requires tests which loads lot of data to make sure we will not break anything next time we change things there. You can write a dummy test which selects 100k rows in lets say 10 seconds and mocha will be fine with them since it has 15 seconds limitation on test run. If we break something and it will become 4 minutes again test will fail as well.

Also you might be interested in recent performance optimizations in persistence of `next` version of typeorm in #1115

pleerock:
I have fixed this issue using es maps instead, thanks for the report and PR, this helped to find the most serious place that needed in optimization.

sviterok:
Your solution with ES Maps looks a lot cleaner, actually. Good job on that!
I’m glad to help with development of the project:)

PR #1446, opened 2018-01-12
buxfix to create query using websql and conditional with boolean value

kadupenido:


daniel-lang:
Just for your information, you currently don't run websql on CI and I'm not sure we ever will. So you won't get feedback from CI if you fix works.
I will let @pleerock decide if he wants the test case for it.

pleerock:
Looks like a proper fix lets merge it

Issue #1433, opened 2018-01-11
After updating to the newest version, ".limit" isn't allowed on type UpdateQueryBuilder


bporter-a3digital:
I'm upgrading from typeorm version 0.0.11 to 0.1.11, and when I tried to fix the "setLimit" method to "limit" as shown below, the compiler states _"Property 'limit' does not exist on type 'UpdateQueryBuilder<NotificationEntity>'."_

Here's the code: 
```typescript

let result = await this.notificationRepository.createQueryBuilder('notification')
            .update({
                processBatch: batch
            })
            .where('notification.processBatch is null')
            .andWhere(`notification.status = '${NotificationStatus.PENDING}'`)
            .andWhere(`notification.nextRunDate < now()`)
            .limit(3)
            .orderBy('priority')
            .execute();

```

I tried moving the update statement to the end of the chain before "execute", but this doesn't work for us as it ends up changing the code's logic.

My team and I are wondering what happened to the ability to set a limit on an update statement, as this is allowed in MYSQL.

pleerock:
Did it actually work in old versions? Even query builder had a `limit` method I don't think it was used in generated query. Previously all methods of different query builders were mixed and it was source of confusion. Limit in update query looks really weird and looks like its only supported in mysql

swilliams-a3digital:
It did work in the previous version (at least with mysql).   I think it was prevously ("setLimit" not "limit" in typeorm). It generated sql that followed this format:

UPDATE [LOW_PRIORITY] [IGNORE] table_reference
    SET assignment_list
    [WHERE where_condition]
    [ORDER BY ...]
    [LIMIT row_count]

https://dev.mysql.com/doc/refman/5.7/en/update.html

Limit and order by are valid options in a mySQL updates.  We can always drop down to writing raw sql, but we'd prefer to stick with typeorm.

Thanks
Scott

swilliams-a3digital:
Using limit and order by is a super cool feature of mysql that is incredibly useful for processing large updates in small batches.  Combined with order by you can also write code where you are updating / working with the highest priority records first.

pleerock:
Okay we can add it back. This is an easy feature to implement and I'll left it for your contributions.

bporter-a3digital:
I thought I made the relevant changes, but I'm having trouble getting a test to pass. The test output online is different than what I'm seeing locally, so I don't really know what else to do for now.

pleerock:
[...]

PR #1447, opened 2018-01-12
UpdateQueryBuilder will now support "limit" and "orderBy" functions for MySQL

bporter-a3digital:
#1433


test/functional/query-builder/update/query-builder-update.ts
```diff
@@ -150,4 +150,29 @@ describe("query builder > update", () => {
 
     })));
 
+    it("should perform update with limit correctly", () => Promise.all(connections.map(async connection => {
```
pleerock: please add checks for your test, something like this:

```ts
if (connection.driver instanceof MysqlDriver) { ... your code } else { code to handle exception since for other drivers you need to check if using those methods really throw you errors }
```

pleerock:
for some reason ci does not execute tests at all. I'll merge this branch and run tests locally.

bporter-a3digital:
Okay, thanks Umed!

PR #1462, opened 2018-01-17
test(sqljs): fixed auto save tests so they can be skipped

daniel-lang:
This will make skipping the sqljs auto save tests possible.

@pleerock, should these changes also go into master?


pleerock:
Thank you, yes since its an issue it should be fixed on master too, but its not mandatory. This functionality is needed for people who want to run test against lets say only mysql, I use this feature quite often during development.

Issue #1465, opened 2018-01-18
Saving relation makes foreign key field `null` using typeorm @next


snaquaye:
Hello,

I have the following entities, using typeorm@next:
```
Account.ts
@Entity()
export default class Account {

  @PrimaryGeneratedColumn('uuid')
  id: string;

  @Column({
    nullable: false,
    unique: true,
  })
  @IsNotEmpty({
    groups: ['registration']
  })
  @IsEmail({
    allow_display_name: true,
    allow_utf8_local_part: true,
    require_tld: true
  }, {
    always: true
  })
  email: string;

  @Column({
    nullable: false,
  })
  @IsNotEmpty({
    groups: ['registration', 'changePassword']
  })
  password: string;

  @OneToOne(type => AccountActivationToken, 'account', { cascade: ['insert', 'remove'] })
  accountActivationToken: AccountActivationToken;

  public encodePassword() {
    if (!this.password) return;

    const salt = genSaltSync();
    this.password = hashSync(this.password, salt);

    return this;
  }
}


Token.ts(STI)
@Entity()
@TableInheritance({ column: { name: 'type', type: 'varchar' } })
export default abstract class Token {

  @PrimaryGeneratedColumn('uuid')
  id: string;

  otpCode: string;

  @Column()
  tokenSecret: string;

  abstract account: Account

  accountId: string;
}

AccountActivationToken.ts
@ChildEntity()
export default class AccountActivationToken extends Token {

  @OneToOne(type => Account, 'accountActivationToken', {
    cascade: ['insert', 'update']
  })
  @JoinColumn({ referencedColumnName: 'id' })
  account: Account

  constructor(public tokenSecret: string, public expiresOn: Date) {
    super();
  }

  markAsUsed() {
    if (this.hasExpired()) throw new Error('Token has expired');

    if (this.isUsed) throw new Error('Token has already been used');

    this.isUsed = true;
  }
}

In my controller
const account = new Account();
account.email = 'test@user.com'
account.password = '123456'
const tokenSecret = 'XXXXXXXXXXXXX'
const expiresOn: new Date()
account.accountActivationToken = new AccountActivationToken(tokenSecret, expiresOn)
const manger = getManager();
manager.save(account)
```

both the account data and accountActivationToken data are saved, but the foreign key remains null. Any ideas to why it is happening this way.

Thanks

pleerock:
are you able to create a PR with a failing test with reproduction of this issue?

snaquaye:
[...]

PR #1471, opened 2018-01-18
test case for #1465

snaquaye:
#1465


test/github-issues/1465/save-relation.ts
```diff
@@ -0,0 +1,39 @@
...
+    async () =>
+      (connections = await createTestingConnections({
+        entities: [__dirname + "/entity/*{.js,.ts}"],
+      }))
+  );
+  beforeEach(() => reloadTestingDatabases(connections));
+  after(() => closeTestingConnections(connections));
+
+  it("account property in accountActivationToken should not be null", () =>
+    Promise.all(
```
pleerock: please put return before Promise.all, tests are not finished because of this mistake

pleerock:
Looks like your particular test is passing on a CI:

```
  save child and parent entity
    ✓ account property in accountActivationToken should not be null (51ms)
```

PR #1472, opened 2018-01-19
Issue 1469 manager find support conditions and options

brendan-roche:


brendan-roche:
replaced by https://github.com/typeorm/typeorm/pull/1474

PR #1477, opened 2018-01-19
Issue template for new issues

daniel-lang:
@pleerock that's the template that I would go with as it includes some guidelines and information that is always nessecary, but not to much.


pleerock:
can you please take a look on changes I have made? Does they make sense? I know I was asking for something short, but previous version wasn't meaningful IMO.

daniel-lang:
Looks great, if that isn't too much for you. I added the cordova driver, but other than that it looks great.

Issue #1478, opened 2018-01-19
Inconsistent documentation on `eager` option


HillTravis:
The docs regarding how the `eager` option works are contradictory regarding use with `QueryBuilder`.

On [the **What are Relations?** page, under the section **Relation Options**](http://typeorm.io/#/relations/relation-options), the first option is `eager`, and it says:

> If set to true, the relation will always be loaded with the main entity when using find* methods **or QueryBuilder** on this entity

Then, on [the **Eager and Lazy Relations** page, at the bottom of the **Eager Relations** section](http://typeorm.io/#/eager-and-lazy-relations/eager-relations), it says:

> Eager relations only work when you use find* methods. If you use QueryBuilder eager relations are disabled and have to use leftJoinAndSelect to load the relation.

pleerock:
Right, its a mistake. Eager relations are only working with find* methods currently. I'll be really grateful if you create a PR with those changes.

PR #1492, opened 2018-01-23
Corrected documentation on `eager` relation option

HillTravis:
https://github.com/typeorm/typeorm/issues/1478


pleerock:
Awesome! Thank you.

PR #1494, opened 2018-01-23
WIP: Joins with Subquery

### LINKED ISSUE #1476: leftJoinAndSelect with subquery returns Unknown column in field list error
Using the Latest 0.1.11 version. Why does `leftJoinAndSelect` create an invalid query using the subquery function?

For example, joining a count subquery

````
         this.plansRepository
            .createQueryBuilder("b")
            .leftJoinAndSelect(subQuery=>{
                return subQuery
                    .select(`COUNT("planId")`,`total`)
                    .addSelect(`planId`)
                    .from(Items,"items")
                    .groupBy(`items.planId`)
            },"i",`i.planId = b.planId`)
            .getRawMany()
````
The query returns an SQL error because there's an redundant subquery alias `i` in front of the `FROM` clause, "i".*, **i** <<<<< FROM

````
ER_BAD_FIELD_ERROR: Unknown column 'i' in 'field list'
````

````
SELECT `b`.`planId` AS `b_planId`, `b`.`planName` AS `b_planName`,`i`.*, i FROM `plans` `b`
LEFT JOIN (SELECT COUNT(planId) AS `total`, planId 
FROM `items` `items` GROUP BY `items`.`planId`) `i` ON i.planId = `b`.`planId`

````

Even if I use a custom select in the plansRepository: 
````
 this.plansRepository
            .createQueryBuilder("b")
            .select(1)
            .leftJoinAndSelect(subQuery......
````
I still get that redundant alias in front of the FROM clause.

### PR DESCRIPTION
**Pull request by daniel-lang**
I'm not completely happy with this fix it doesn't work for postgres and works around the problem rather than fixing it.

The reason postgres is excluded from the tests is that the query that gets generated with the fix looks like this:
```sql
SELECT "b"."planId" AS "b_planId", "b"."planName" AS "b_planName", "i".* FROM "plan" "b"
LEFT JOIN (SELECT COUNT("planId") AS "total",
planId FROM "item" "items" GROUP BY "items"."planId") "i" ON i.planId = "b"."planId"
```
Postgres doesn't like the unescaped column planId in the subquery and i.planId on the "ON"-statement.

From my findings, the both are not escaped because [`SelectQueryBuilder::replacePropertyNames()`](https://github.com/typeorm/typeorm/blob/master/src/query-builder/QueryBuilder.ts#L514) doesn't have any metadata to compare and replace it with. The metadata is missing because [`JoinAttribute::metadata`](https://github.com/typeorm/typeorm/blob/master/src/query-builder/JoinAttribute.ts#L136) can't find any metadata as `entityOrProperty` is neither an entity nor a property but rather a special kind of subquery. I couldn't come up with a proper way to differentiate between an entity and a subquery.
But if we somehow get metadata for subqueries, [`SelectQueryBuilder::join()`](https://github.com/typeorm/typeorm/blob/master/src/query-builder/SelectQueryBuilder.ts#L1129) will have to change as subqueries are currently only evaluated if no metadata exists.

Fixes #1476

### DIFF
--- a/gulpfile.ts
+++ b/gulpfile.ts
@@ -375,6 +375,7 @@ export class Gulpfile {
     @SequenceTask("ci-tests")
     ciTests() {
         return [
+            "clean",
             "compile",
             "tslint",
             "wait",
--- a/src/query-builder/SelectQueryBuilder.ts
+++ b/src/query-builder/SelectQueryBuilder.ts
@@ -1210,7 +1210,8 @@ export class SelectQueryBuilder<Entity> extends QueryBuilder<Entity> implements 
                     const hasMainAlias = this.expressionMap.selects.some(select => select.selection === join.alias.name);
                     if (hasMainAlias) {
                         allSelects.push({ selection: this.escape(join.alias.name!) + ".*" });
-                        excludedSelects.push({ selection: this.escape(join.alias.name!) });
+                        const excludedSelect = this.expressionMap.selects.find(select => select.selection === join.alias.name);
+                        excludedSelects.push(excludedSelect!);
                     }
                 }
             });
--- a/test/github-issues/1476/entity/Item.ts
+++ b/test/github-issues/1476/entity/Item.ts
@@ -1,0 +1,11 @@
+import { Entity, PrimaryColumn, Column } from "../../../../src/index";
+
+
+@Entity()
+export class Item {
+    @PrimaryColumn()
+    itemId: number;
+
+    @Column()
+    planId: number;
+}
--- a/test/github-issues/1476/entity/Plan.ts
+++ b/test/github-issues/1476/entity/Plan.ts
@@ -1,0 +1,11 @@
+import { Entity, PrimaryColumn, Column } from "../../../../src/index";
+
+
+@Entity()
+export class Plan {
+    @PrimaryColumn()
+    planId: number;
+
+    @Column()
+    planName: string;
+}
--- a/test/github-issues/1476/issue-1476.ts
+++ b/test/github-issues/1476/issue-1476.ts
@@ -1,0 +1,59 @@
+import "reflect-metadata";
+import { expect } from "chai";
+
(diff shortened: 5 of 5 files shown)


daniel-lang:
And btw I also added `clean` to the tests gulp task so that the build directory is cleared before the project is compiled.

daniel-lang:
One option to tell entities and subqueries apart could simply be a boolean value inside `JoinAttribute` like `isSubQuery` which was to be set when any joining method is called

daniel-lang:
@pleerock any thoughts on this?

pleerock:
`replacePropertyNames` is the worth place in the ORM. However I don't see any other way to implement "automatic" escaping. 

I don't know what to do in this case, subquery can return anything, we cannot anyhow reliable escape this, so its on user hands to resolve this properly.

To fix this test for all drivers you need to do use `escape` method from the driver or query builder.
Im going to release 0.1.12 with this change, can I merge it?

daniel-lang:
you can merge it, but as I mentioned without escaping from the user, postgres won't work.

daniel-lang:
So what you are saying is, that the subquery will have to look like this:
```typescript
const plans = await planRepo
   .createQueryBuilder("b")
   .leftJoinAndSelect(
      subQuery => {
         return subQuery
            .select(`COUNT("planId")`, `"total"`)
            .addSelect(`"planId"`)
            .from(Item, "items")
            .groupBy(`"items.planId"`);
      }, "i", `"i"."planId" = "b"."planId"`)
   .getRawMany();
```

pleerock:
postgres works if we manually escape, right?

pleerock:
I think this can be called "by design" currently, because I don't see any solution for this problem.

daniel-lang:
> postgres works if we manually escape, right?

Yes, but I can modify the test to prove that it works.

pleerock:
btw give me your npm account, I'll grant you publishing access

Issue #1490, opened 2018-01-23
Undefined does not affect model objects


crutchcorn:
When trying to update a user, I mistakenly ran the following commands:
```
      user.resetPasswordToken = undefined;
      user.resetPasswordExpires = undefined;
      userRepo.save(user)
```

But I noticed that it wasn't doing anything to the user object. However, when I changed this to:
```
      user.resetPasswordToken = null;
      user.resetPasswordExpires = null;
      userRepo.save(user)
```

It had the expected behavior. This could be different from most people's expectations (even if this is expected behavior from the project), and should be clearly documented

pleerock:
Its by design and allows partial updations. You can lets say do `userRepo.save({ id: 1, middleName: "Lalala" })` without being worried that other data will be removed. Also setting properties to `null` is actually making column values null in the database as well, so logically everything fit good.

and btw you always can contribute in documentation, its quite easy to do and will help others to solve issues you faced.

crutchcorn:
I understand and think it's a great design choice. 

I know I can contribute, and do plan to, but often find these bugs in the middle of a code session and just need to log them out. I'll try getting a PR in with doc additions here soon.

pleerock:
nice!

Issue #1483, opened 2018-01-20
BeforeUpdate is not running


crutchcorn:
Adding `@BeforeUpdate()` fails to run in my model when updating the model.

A full code sample can be found here:
https://github.com/crutchcorn/PET-Template/commit/721d4bac85e6cb0e2b3d6a4e28098538c17b9fc0

crutchcorn:
I am incredibly dumb. Ignore this issue, as it is not valid

crutchcorn:
@pleerock I thought this was a bug as, even though I was hitting `save` to update, I was not passing any new information to the model. This should probably be documented as this is drastically different behavior than what I think most would expect

pleerock:
before update means before actual updation in the database. If there are no changes in the model it does not to do database updation, thus listener is not being called.

crutchcorn:
Right, but this is behavior that many might not expect. You might want to document that so that it's more explicit

lordnynex:
I understand the underlying issue in this ticket that there is no new properties to the model so the hook is not firing. In my case, quite a few parameters are being updated but I'm not seeing the hook fire at all. The model is being updated by `repository.updateById(id, params)`. Is this expected? Our particular work flow is 'create empty entity' and then 'update the entity with actual data'.  

Unrelated, is there a way for the BeforeUpdate listener to have some context into what fields are changing? In my case, I'm attempting to perform a cryptographic operation and need to find away around performing this operation more than once on a field.

crutchcorn:
@lordnynex 
Sounds like you should create a new bug - as that's unrelated to the behavior found here

You can use `BeforeUpdate`, query the database for the old model, and compare against `this` (`this` contains the new object). However, if you're using `select: false` (which I would suggest for crypto anything), you're more or less out of luck. Related thread here: https://github.com/typeorm/typeorm/issues/1514

lordnynex:
@crutchcorn Thanks for getting back to me. I read it, the whole thing seems complicated and confusing. In the last hour or so I'm having much better luck with implementing a custom `ValueTransformer` on the column. Take a look at this test case example.

https://github.com/typeorm/typeorm/blob/d36ae18a887f3045e5c24f6b798e3afe13707043/test/functional/columns/value-transformer/entity/Post.ts#L6

crutchcorn:
[...]

Issue #828, opened 2017-09-06
Optionally disallow columns from showing in query 


crutchcorn:
I'd like to have a way to prevent a column from showing up in a query if the option is enabled. (for example, blacklisting a password and hash unless explicitly requesting for them)

As part of this feature request, I'd also like a way to explicitly request these blacklisted columns when requested. This is a very similar feature I'm requesting as [`{select: false}`](mongoosejs.com/docs/2.7.x/docs/schematypes.html) in mongoose.

pleerock:
There is a possibility to partial only select columns you need, for example:

```ts
const user = await userRepository
     .createQueryBuilder('row')
     .select('row.id')
     .addSelect('row.username')
     .where('row.username = :name', {name: 'username'})
     .getOne();
```

crutchcorn:
But is it possible to change this from the model definition? I would like password and hash to ALWAYS be hidden from TypeORM queries unless otherwise explicit calling. This will cut down on the possibility of developer errors and make things easier to secure

pleerock:
We can add a new option and/or decorator to support this, however how many use cases it will have? I don't know any other cases other then user passwords in general applications where it can be helpful. Isn't it enough to simply delete property once you load user?:

```ts
const user = await ...;
delete user.password;
```

crutchcorn:
`delete` is horribly slow IIRC, and goes back to my "developers managing things they need not always manage". I think that this could have implamentations for anything that might hold private information (SSN, passwords, PII, etc)

pleerock:
This feature is implemented and will be released in next (`0.1.0-alpha.38`) version. Usage:

```ts
import {Entity} from "../../../../../src/decorator/entity/Entity";
import {Column} from "../../../../../src/decorator/columns/Column";
import {PrimaryColumn} from "../../../../../src/decorator/columns/PrimaryColumn";

@Entity()
export class Post {

    @PrimaryGeneratedColumn()
    id: number;

    @Column()
    title: string;

    @Column()
    text: string;

    @Column({ select: false })
    authorName: string;

}
```

crutchcorn:
While I am able to do what you recommended with:
```
 const user = await userRepository
    .createQueryBuilder('row')
    .select('row.id')
    .addSelect('row.email')
    .addSelect('row.password')
    .where('row.email = :name', {name: 'email'})
    .getOne();
```

[...]

PR #1513, opened 2018-01-26
Update documentation for `select`, `NULL`, and subscribers

crutchcorn:
This should add a few lines of documentation that were mentioned in a few issues I was involved in:
#1490 #1483 #828


pleerock:
Thank you very much!

PR #1520, opened 2018-01-27
Webpack, browser, warnings fix: Module not found

darekf77:
Fix for webpack (browser build) warning: Module not found. This will fix all webpack errors if you are build for example angular 5 app.

WARNING in ../node_modules/typeorm/platform/PlatformTools.js
Module not found: Error: Can't resolve '**database module name**' in '/Users/dariusz/ProjectsMac/npm/tsc-npm-project/projects/workspace/node_modules/typeorm/platform'
resolve 'redis' in '/Users/dariusz/ProjectsMac/npm/tsc-npm-project/projects/workspace/node_modules/typeorm/platform'
  Parsed request is a module

### DIFF
--- a/package.json
+++ b/package.json
@@ -9,6 +9,18 @@
     "name": "Umed Khudoiberdiev",
     "email": "pleerock.me@gmail.com"
   },
+  "browser": {
+    "redis": false,
+    "pg-query-stream": false,
+    "pg-native": false,
+    "pg": false,
+    "oracledb": false,
+    "mysql2": false,
+    "mysql": false,
+    "mssql": false,
+    "mongodb": false,
+    "sqlite3": false
+  },
   "repository": {
     "type": "git",
     "url": "https://github.com/typeorm/typeorm.git"


pleerock:
Can someone explain me how this works?

adding webpack expert @daniel-lang

daniel-lang:
The Angular CLI doesn't open its webpack config to its users. Therefore the fix that I describe in [our docs](http://typeorm.io/#/supported-platforms/browser) cannot be applied.
This would silence webpack but I'm fairly certain, that @darekf77 will run into other problem as he is using the nodejs version and not the browser version.
A fix for Angular would be to import from `typeorm/browser` directly instead of `typeorm`. That should also get rid of the webpack warnings. Maybe @darekf77 can try that and see if it helps.

pleerock:
but how `browser` section make things to work? Who reads it and uses? Webpack? Angular CLI?

daniel-lang:
The browser section is, in this case, read by webpack. I found some information about `browser` [here](https://github.com/defunctzombie/package-browser-field-spec). These entries tell webpack to ignore the missing packages that the nodejs version requires in PlatformTools.

pleerock:
@daniel-lang it sounds good. We shall merge it, right?

daniel-lang:
Well yes, I guess it helps.
But I'm certain @darekf77 is going to run into other problem because he is not using the browser version.

pleerock:
Some of the modules like mkdirp, path and debug are not included. why @darekf77  ?

darekf77:
@daniel-lang I didn't know if there is a browser version of typeorm. Now in my isomorphic lib I will just replace (using webpack): require('typeorm') with require('typeorm/browser'). But before I do that this "browser" section in package.json is preventing in my build lots of warnings... So I think should be there... anyway @pleerock.

pleerock:
decision to choose browser or node is depend on what application you do. If you are storing data in the browser you choose browser version, if you are storing data on the backend you use backend version.

What is your use case @darekf77 ?

darekf77:
I am storing data in sqlite backend, but I am using classes from backend on brower side as angular services.

darekf77:
@pleerock **typeorm/browser** works for me perfect. Thanks for awesome library.

PR #1524, opened 2018-01-28
Fix issue#1140

martynsmith:
This changes the postgres driver behaviour such that if a transform exists on a column, the driver doesn't try to do further manipulation of the value after the transform has run.

### DIFF
--- a/src/driver/postgres/PostgresDriver.ts
+++ b/src/driver/postgres/PostgresDriver.ts
@@ -324,7 +324,7 @@ export class PostgresDriver implements Driver {
      */
     prepareHydratedValue(value: any, columnMetadata: ColumnMetadata): any {
         if (columnMetadata.transformer)
-            value = columnMetadata.transformer.from(value);
+            return columnMetadata.transformer.from(value);
 
         if (value === null || value === undefined)
             return value;


pleerock:
Thank you very much for your contribution. 

I don't think its a correct fix, because, if let's say user wants to transform boolean into "yes" or "no", but database will pass to transform method instead of `true` or `false` -> `1` or `0` values and transformation logic will rely on boolean it will fail. It will not be obvious why numbers are returned to user instead of boolean. So, I think transformation logic must be moved under everything else and `return ...` in each block of code must be replaced with `value = ...`. 

Also few things needs to be done before this PR can be merged.

* You should include tests which test functionality you made, its a requirement because we need to make sure functionality you have fixed will not be broken again (its easy just copy paste any exist code and replace with your data)
* This issue must be fixed in all drivers (they all have those lines of code duplicated)
* Add note about this change in the changelog for the next version

tmstieff:
Do you think a flag on the transformer interface, 'skipAdditionalTransformations' or something like that would suffice? That way it can be handled by the transformer itself, and there's no need to delegate the responsibility to each column definition.

martynsmith:
Yes, I think that would be a good option too :-)

martynsmith:
Closing this PR since it's now replaced by #1547

PR #1533, opened 2018-01-31
Fixed typo in mongodb.md

Yonben:
Line 193 was importing getMongoManager instead of getMongoRepository.


pleerock:
Thank you very much!

PR #1535, opened 2018-01-31
Updated the hidden column to reflect the query

umarov:
I noticed this while browsing the docs. There were two columns with the name `name`. The example query builder to fetch that column was selecting `password`. So updated the column to reflect that


pleerock:
Thank you very much!

Issue #1530, opened 2018-01-30
Error: Uncaught (in promise): Error: FOREIGN KEY constraint failed


davidmpaz:
**Issue type:**

[x] question
[ ] bug report
[ ] feature request
[ ] documentation issue

**Database system/driver:**

[ ] `cordova`
[ ] `mongodb`
[ ] `mssql`
[ ] `mysql` / `mariadb`
[ ] `oracle`
[ ] `postgres`
[ ] `sqlite`
[x] `sqljs`
[ ] `websql`

**TypeORM version:**

[ ] `latest`
[ ] `@next`
[x] `0.1.11` (or put your version here)

**Steps to reproduce or a small repository showing the problem:**

Inside an ionic application I create the connection and populate it with data. My models contains relations from `Product` going down to other models, this mean in `Product` model I set up relation with cascade insert/updates so I only need to hydrate products and save instead saving all models separately.

Connection setup:
```typescript
if (isBrowser(this.platform)) {
    dbOptions = {
        type: 'sqljs',
        autoSave: true,
        location: 'products',
    } as SqljsConnectionOptions;
}

return createConnection(Object.assign(dbOptions, {
    entities: this.getModelsToLoad(),
    synchronize: true,
    logging: true,
}));
```

After I populate database everything works fine, and then after I reload page, i can see queries ran in console, because `synchronize: true`, i assume. Last query is:
```
executing query:  CREATE TABLE "temporary_product" ("identifier" integer NOT NULL PRIMARY KEY AUTOINCREMENT, "id" varchar NOT NULL, ....)
executing query:  INSERT INTO "temporary_product"("identifier", "id", ....) SELECT "identifier", "id", ... FROM "product"
executing query:  DROP TABLE "product"
executing query:  ROLLBACK
```
than I receive the error.

Could you guys help me on how to overcome this. My expected outcome would be that I can use same data after a page reload, since that's happens a lot when developing in browser an ionic app. Foreign keys enabled are not allowing me to re-initialize database again from existent data.

Thanks in advance,
David

pleerock:
@daniel-lang do you have something in mind?

giorgiofellipe:
Any workaround?

AlexMesser:
please check this issue on latest `@next` version, it should be fixed

daniel-lang:
so #1536 is no longer needed?

davidmpaz:
@daniel-lang, @pleerock, @AlexMesser thanks for taking care of this!

[...]

PR #1536, opened 2018-01-31
SQLite: add integer primary keys to primary key list

daniel-lang:
I discovered, that SQLite doesn't create primary columns that are specified as `integer` but rather creates an alias to an internal column called `rowid` which is created for every table. When the primary key column is not an integer the column and an index is created. Before this fix, the primary key was only added to the list if the index existed. This is why SQLite based drivers tried to add primary keys that already existed when an `integer` was used.
This fixed the first of the two problems I mentioned in #1530. For the second one a special method would be needed that can be called `RdbmsSchemaBuilder::build()` to disable and enable foreign key constraints for database systems that don't support `ALTER TABLE` statements. Or is there a better way/place to enable and disable them?

### DIFF
--- a/src/driver/sqlite-abstract/AbstractSqliteQueryRunner.ts
+++ b/src/driver/sqlite-abstract/AbstractSqliteQueryRunner.ts
@@ -272,6 +272,12 @@ export class AbstractSqliteQueryRunner implements QueryRunner {
                 if (tableColumn.isGenerated) {
                     tableColumn.generationStrategy = "increment";
                 }
+                // According to the SQlite docs, integer primary keys don't have an index.
+                // Therefore the primary key has to be added seperately.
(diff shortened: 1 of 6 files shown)


daniel-lang:
This is the quote from the SQLite docs about indexes and constraints:
> In most cases, UNIQUE and PRIMARY KEY constraints are implemented by creating a unique index in the database. (The exceptions are INTEGER PRIMARY KEY and PRIMARY KEYs on WITHOUT ROWID tables.)

And the "INTEGER PRIMARY KEY"
> With one exception noted below, if a rowid table has a primary key that consists of a single column and the declared type of that column is "INTEGER" in any mixture of upper and lower case, then the column becomes an alias for the rowid. Such a column is usually referred to as an "integer primary key". A PRIMARY KEY column only becomes an integer primary key if the declared type name is exactly "INTEGER".

PR #1539, opened 2018-01-31
Fix typos

jamesrichford:
Hey guys,

Really enjoying typeorm so far - just noticed a few typos so thought I'd send them on over and be as helpful as I can be (maybe next PR will be something actually useful :) haha)


pleerock:
Each even smallest possible contribution is valuable. Thank you very much!

PR #1541, opened 2018-02-01
fixing repeating column types for mysql / mariadb

pablocarreraest:
removing repeated types of `int`, `tinyint`, `smallint`, `mediumint`, `bigint` at  column types for mysql / mariadb


pleerock:
Thank you!

pablocarreraest:
I will remove also the other types of other db's that are repeated in the documentation

Issue #1545, opened 2018-02-02
Typeorm runs insert query instead of update query on save of existing entity for ManyToOne relationships 


SamuelColacchia:
**Issue type:**

[x] bug report

**Database system/driver:**

[x] `postgres`

**TypeORM version:**

[x] `@next`

***Relevant models***

``` ts
@Entity("main") 
export class mainModel extends BaseEntity {
	@Column({ type: "bigint", unique: true, primary: true })
	@Generated("increment")
	id: string;
	
	//other columns ommited
	@OneToMany(
		type => dataModel,
		dataModel=> dataModel.id,
		{ cascade: true, eager: true }
	)
	data: dataModel[];	
}

@Entity("valid")
export class validModel extends BaseEntity {
	@Column({
		type: "text",
		primary: true
	})
	valid: string;

	@Column({
		type: "boolean",
		default: "true"
	})
	active: boolean;

	@OneToMany(
		type => SourceOptionsDataModel,
		sourceOptionsDataModel => sourceOptionsDataModel.sourceOption
	)
	sourceOptionsData: SourceOptionsDataModel[];
}

@Entity("data")
export class dataModel extends BaseEntity {
	@ManyToOne(type => validModel , {
		primary: true,
		eager: true
	})
	@JoinColumn({
		name: "valid",
		referencedColumnName: "valid"
	})
	valid: validModel;

	@ManyToOne(type => RmaModel, {
		primary: true
	})
	@JoinColumn({
		name: "mainId",
		referencedColumnName: "id"
	})
	id: mainModel ;

	@Column({
		type: "boolean",
		default: false
	})
	checked: boolean;
}
```

**Example Data**

***valid***
``` json
{
"valid": "somevalue"
}
````

***main*** 
``` json
{
"data": [{"valid": "somevalue", "checked": true }]
}
```

### Steps to reproduce

1.  Save valid to database
2. Save main to database
3. Attempt to update main yields issues 

***Example code to produce issue***
```ts
const theId = \\the id you get from saving main
const theManager = getManager();
const theModel: any = await theManager.findOne(model, theId);
const theSave = await theManager.save(model, theModel);
```

***Issue***
Typeorm attempts to to run a insert query instead of a update query for values in dataModel entity.

***Attempts to resolve***
Added id to data array in main.

pleerock:
You did not provide a "save" code. Are you able to create a PR with failing test case? (easy, just take a look how tests are organized)

[...]

PR #1564, opened 2018-02-07
Issue #1545 Added Tests, for updating a relational entity

SamuelColacchia:


pleerock:
Why original issue is already closed? Is this test wrong?

PR #1619, opened 2018-02-21
chore(common): export DeepPartial from common

cojack:
For a compatibility with a Repository API, when build an 3rd party tools


cojack:
This will solve #1619

pleerock:
thanks, will be released in next version.

Issue #1620, opened 2018-02-21
Transformer is not applied when reading back from DB during save


pdeszynski:
**Issue type:**

[ ] question
[ x] bug report
[ ] feature request
[ ] documentation issue

**Database system/driver:**

[ ] `cordova`
[ ] `mongodb`
[ ] `mssql`
[ ] `mysql` / `mariadb`
[ ] `oracle`
[ ] `postgres`
[ x] `sqlite`
[ ] `sqljs`
[ ] `websql`
(Maybe other too)

**TypeORM version:**

[ ] `latest`
[ x] `@next`
[ ] `0.x.x` (or put your version here)

**Steps to reproduce or a small repository showing the problem:**
Update with a transformer will fail as the field in not firstly transformed. Here's an example stack trace:

```
 at Object.to (test/functional/repository/basic-methods/entity/Post.ts:18:40)
      at SqliteDriver.AbstractSqliteDriver.preparePersistentValue (build/compiled/src/driver/sqlite-abstract/AbstractSqliteDriver.js:9:12465)
      at build/compiled/src/query-builder/UpdateQueryBuilder.js:9:29999
      at Array.forEach (<anonymous>)
      at build/compiled/src/query-builder/UpdateQueryBuilder.js:9:29332
      at Array.forEach (<anonymous>)
      at UpdateQueryBuilder.createUpdateExpression (build/compiled/src/query-builder/UpdateQueryBuilder.js:9:29111)
      at UpdateQueryBuilder.getQuery (build/compiled/src/query-builder/UpdateQueryBuilder.js:9:11113)
      at UpdateQueryBuilder.QueryBuilder.getQueryAndParameters (build/compiled/src/query-builder/QueryBuilder.js:9:19425)
      at UpdateQueryBuilder.<anonymous>  #(build/compiled/src/query-builder/UpdateQueryBuilder.js:9:14908)
```

Here's an update `Photo` entity, which will cause this exception: https://github.com/pdeszynski/typeorm/blob/next/test/functional/repository/basic-methods/entity/Post.ts

And added a test that will throw an exception: https://github.com/pdeszynski/typeorm/blob/next/test/functional/repository/basic-methods/repository-basic-methods.ts#L323

and run it with ` npm test -- --grep="should update existing entity"`

I didn't test it with other engines. 

This duplicates probably, an already closed issue (lack of additional info): https://github.com/typeorm/typeorm/issues/1291

andresgariglio:
[...]

PR #1627, opened 2018-02-21
Fix problem with tranformers having invalid value when saving entity read from DB

pdeszynski:
Solves: https://github.com/typeorm/typeorm/issues/1620


test/utils/test-setup.ts
```diff
@@ -1,3 +1,4 @@
+import "source-map-support/register";
```
pleerock: do we have source-map-support in package.json?

pleerock:
thanks for the contribution!

andresgariglio:
Hi! I'm still reproducing this issue, but it only happens for `@PrimaryColumn`, it's working ok for `@Column` fields.

PR #1647, opened 2018-02-25
Edit database replication document

dsaki:
Hi
In database replication document, 
https://github.com/typeorm/typeorm/blob/master/docs/multiple-connections.md#replication
`setQueryRunner` returns `void`.

So, this document is wrong.


pleerock:
thanks for the PR, but this code fixes consequence. Instead can you please add `return this` to `QueryBuilder`? This would be a proper fix.

pleerock:
Thank you very much!

dsaki:
Thank you 😇

Issue #1351, opened 2017-12-15
Remove missing child objects in a relation


timothyvanderaerden:
Hello

I have a question where I cannot find the solution for in the documentation.

My Understanding is that when you use the `.save()` function, the foreign key of all all missing child objects will be set to null. Is there a 'nice' way to remove those objects from the database instead of setting the foreign key to null, using the save methode?

So for example if I have the following books in my database:

| id | name |
| -- | ------ |
| 1  | Book x |
| 2  | Book y |

And save the following object:
```
Author: {
   name: "...",
   books: [
       {
           name: "Book z",
       },
       {
            id: 2
            name: "Book yy",
       }
    ]
}
```
```
const authorRepository = getConnection('database').getRepository(Author);
await authorRepository.save(author);
```
The result should be:

| id | name |
| -- | ------ |
| 2  | Book yy |
| 3  | Book z |

I'm using version: 0.2.0-alpha.12.

Thanks

pleerock:
No there is no way to "nicely" remove such related object, I was thinking about cascade remove for such use case, but not sure about its implementation yet. Lets left this issue open for now.

wux5:
I have exactly same problem. Wondering why we chose to implement it with the current approach (update the foreign key with null), isn't it make more sense to delete the removed records from db? Maybe make it an option to either delete or null the fk?

My workaround is to delete all records that's not in the books array (based on the example above, delete from books where author_id = 123 and id not in (2);), and then call author.save(), wrapping them in one transaction. (still concerned about the overhead of .save() which tries to NULL the fk though there is not any).

mazyvan:
Having the same issue. Any fix. I think it's necessary to fix the internal behavior. 
I have an entity that has a relation with cascades enabled. 
I want to remove the relation entities with just: `primaryEntity.relationsEntities = []` 
It's more clean than the workarounds out there

nolazybits:
Hello, it would be nice to add this in the doc that at the moment this is not supported.
Took me a good hour to come to this thread.
Thanks again for this nice library :+1:

Kononnable:
@nolazybits 
PRs from the community to the docs are always welcomed.

Pimnr47:
Note that this issue also arises when updating an existing entity with children using the repo's save():

Please do tell me if I get this wrong:

[...]

PR #1665, opened 2018-02-27
Feature: Delete orphaned rows

adenhertog:
Also, fix: https://github.com/typeorm/typeorm/issues/1351

This adds an additional option on one-to-many relationships where child entities are deleted, rather than having their foreign key set to null, when removed from their parent. 

Many times we have the situation where a parent entity "owns" a child. An example of this is when a `User` has a set of `RoleAssignment` in the context of permissions. When a `RoleAssignment` is removed from the user, we expect for the row to be deleted. Previously the row would remain in the database but with the `userId` column set to null, leaving an orphaned row that would clog up the table.

This fix allows for `RoleAssignment` to specify on its relationship with `User` that it should be deleted, rather than orphaned, when the relationship is removed.

By default, the original behaviour of setting the foreign key to `null` is retained.


pleerock:
Im sorry but I cannot accept this PR. 

First, persistence was completely refactored in `@next`, cascade removes are not working currently in next and Im not yet sure about their design and implementation.

If you want have this feature in `@next` then its better to start discussion before implementation since this area is very sensitive and must be implemented very carefully and right design should be chosen first.

guscastro:
I'd personally really like to have this feature in `@next`. Could we start the discussion with your concerns about the current implementation, @pleerock ?

pleerock:
@adenhertog I see you pushed some changes, but as I told if it will be implemented it must be re-implemented against `next` branch (and all changes it has)

adenhertog:
Hey @pleerock, I'm happy to discuss how this might fit into `@next` as for certain scenarios it fits better than setting FKs to null. 

I'm currently using `0.1.13` in production and needed this change immediately, so can close this PR without merging if you don't think it fits into `@latest`

kirliam:
Just for your information, this is a common use-case for me as well, and it would be nice if I could have the option to tell typeorm to delete the orphan entities, since today I am required to do it manually!

kristophjunge:
Hello Guys,

please note that i have looked on this issue from a bug perspective in #1761 and also filed a pull request against @next.

Having cascade remove working in these cases is essential. Setting the foreign key to NULL only covers a portion of the possible situations.

If cascade remove is not working it basically renders all cascade methods to be unusable. I cannot implement "insert" and "update" one way and have some extra code to fill the missing "remove" part. 

I can go only full with these features or completely without.

adenhertog:
Closing this since master has diverged by a major version. I'd still find it very useful having this addressed in the latest release

haroldo-ok:
I am quite surprised that one of the top JS/TS ORM's does not support orphan removal...

PR #1678, opened 2018-02-28
Proper default column type for postgres

### LINKED ISSUE #1677: Array type default for Postgres always shows update needed in schema:log
Related to #1532 
**Issue type:**

[ ] question
[x] bug report
[ ] feature request
[ ] documentation issue

**Database system/driver:**

[ ] `cordova`
[ ] `mongodb`
[ ] `mssql`
[ ] `mysql` / `mariadb`
[ ] `oracle`
[x] `postgres`
[ ] `sqlite`
[ ] `sqljs`
[ ] `websql`

**TypeORM version:**

[x] `latest`
[ ] `@next`

When using an array column in Postgres, using the array literal as a string works to set the default value in the database:

```js
@Column({
  nullable: false,
  type: 'smallint',
  array: true,
  default: '{5}',
})
```

However, when running `schema:log` or `schema:sync` with the cli, it always reports that it needs to update the default value of this column:

```sh
ALTER TABLE "public"."user" ALTER COLUMN "roles" SET DEFAULT '{5}';
```

This is because the actual default value in postgres is `'{5}'::smallint[]`. In other words, the default value needs to be the literal string, wrapped in single quotes, with `::column.type[]` appended to the end.

jdforsythe:
Fixes #1677

### DIFF
--- a/src/driver/postgres/PostgresDriver.ts
+++ b/src/driver/postgres/PostgresDriver.ts
@@ -311,7 +311,7 @@ export class PostgresDriver implements Driver {
 
         } else if (columnMetadata.type === "simple-array") {
             return DateUtils.simpleArrayToString(value);
-
+            
         } else if (columnMetadata.type === "simple-json") {
             return DateUtils.simpleJsonToString(value);
         }
@@ -472,7 +472,7 @@ export class PostgresDriver implements Driver {
      * Normalizes "default" value of the column.
      */
     normalizeDefault(column: ColumnMetadata): string {
-        const arrayCast = column.isArray ? `::${column.type}[]` : '';
+        const arrayCast = column.isArray ? `::${column.type}[]` : "";
 
         if (typeof column.default === "number") {
             return "" + column.default;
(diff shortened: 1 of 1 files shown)


pleerock:
Thank you for your contribution! Can you please create a PR against `next` ? There are lot of changes in next currently and I want these changes to be there.

adding @AlexMesser

jdforsythe:
@pleerock this is a bigger change in `next`. There has been a change to sending only `columnMetadata.default` into the `[driver].normalizeDefault()` method. The `isArray` property won't be available unless the whole `columnMetadata` object is passed in.

For consistency, it should probably be changed in `MysqlDriver.ts`, `OracleDriver.ts`, etc.

pleerock:
@jdforsythe its up to you, but you can already use `next` - its much much better in terms of performance and number of bugs. We are planning to release it this month.

pleerock:
If you still prefer master I can merge this one as well.

jdforsythe:
We are going to look into switching to `next` since it's that close to release and we are pre-release, but I think it would be helpful to have this in npm for now

Issue #1493, opened 2018-01-23
Error parsing pg connection string


andrewdhazlett:
**Issue type:**

[ ] question
[x] bug report
[ ] feature request
[ ] documentation issue

**Database system/driver:**

[ ] `cordova`
[ ] `mongodb`
[ ] `mssql`
[ ] `mysql` / `mariadb`
[ ] `oracle`
[x] `postgres`
[ ] `sqlite`
[ ] `sqljs`
[ ] `websql`

**TypeORM version:**

[ ] `latest`
[ ] `@next`
[x] `0.1.9` (or put your version here)

**Steps to reproduce or a small repository showing the problem:**
attempt to establish a connection with a username or password containing the `@` character
<!--
    To answer those questions you need to put "x" inside the square brackets, for example:
[x] `mysql`
[ ] `postgres`
!>

pleerock:
@andrewdhazlett Im very sorry but accidentally removed your message content, really sorry about that, but can you write it again

andrewdhazlett:
this can be fixed by using a mature package to parse the url https://github.com/typeorm/typeorm/blob/feada9a3dd06b12c38b9f021bb3bd3dc72fd0094/src/driver/DriverUtils.ts#L56

pleerock:
@andrewdhazlett if its an easy fix, do you want to send a PR with the fix?

pleerock:
We cannot use third-party package btw because typeorm must work in browsers without requiring third-party libs. I'll left this issue for the community to fix.

Rhymmor:
I guess it can be appropriate to just split be last `@` character. There is something similar in [node-url](https://github.com/defunctzombie/node-url):
https://github.com/defunctzombie/node-url/blob/13a35bd35a2cdf3fbfa7ee9c6ed5b927a48d6821/url.js#L200-L210
If it is, I can create PR.

pleerock:
No, I don't think so `username@password` where password can contain `@` can break. Im not sure how people handle such except for having to backslash this character. And Im not even sure if postgres really allows such symbols in username and password.

Rhymmor:
Couldn't find anything in postgres docs but I was able to set password contains `@` locally.

[...]

PR #1690, opened 2018-03-02
Issue #1493. Fix connection url parsing

Rhymmor:
In addition to the discussion in #1493, there is a fix for the case where password contains colons


pleerock:
Thank you very much! Would be great if you'll have a time and wish if you research this question for other databases as well.

Rhymmor:
Does it not work with other databases right now? I see that it is used for other databases with credentials and urls should be parsed correctly no matter which prefix they have.

pleerock:
ah yes its not driver-specifc and used by all of them

Issue #1680, opened 2018-02-28
Delete & Update applies to all entities in table if criteria is undefined or empty 


alex-red:
**Issue type:**

[x] question
[x] bug report
[ ] feature request
[ ] documentation issue

**Database system/driver:**

[ ] `cordova`
[ ] `mongodb`
[ ] `mssql`
[x] `mysql` / `mariadb`
[ ] `oracle`
[ ] `postgres`
[ ] `sqlite`
[ ] `sqljs`
[ ] `websql`

**TypeORM version:**

[ ] `latest`
[x] `@next`
[ ] `0.x.x` (or put your version here)

**Steps to reproduce or a small repository showing the problem:**

Using the repository API and calling any of the following:

`await repository.delete(null)`
`await repository.delete(undefined)`
`await repository.delete([])`

Will drop the entire table.

I don't think this is the intended functionality, or if it is, the documentation should make it very clear that it will delete everything. Also may be related to #1226.

pleerock:
Correct. Since its easy to implement, do you want to contribute?

alex-red:
Sure, sounds good to me, I'll make a PR soon then :)

alex-red:
Hey @pleerock some questions for you:

1) What should the return value look like in this case? Should I return an empty DeleteResult?

2) I see that the same issue occurs with repository.update, ex: `repository.update(null, { someColumn: "test" }` will update the entire table, should I apply the same checks here?

pleerock:
> I see that the same issue occurs with repository.update, ex: repository.update(null, { someColumn: "test" } will update the entire table, should I apply the same checks here?

yes, please

> What should the return value look like in this case? Should I return an empty DeleteResult?

Probably returning empty delete result is a bad idea. Maybe its better to throw an error?

MajidJafari:
How we can remove all entries of a table at once, now? Surely we can use `QueryBuilder`, but I'm sure `TypeORM` needs a simple way to do that, too.

andreasblueher:
Hey @pleerock did you see this question? I can't find a way to delete all data for a specific entity either. Could you please help us?

nebkat:
> How we can remove all entries of a table at once, now? Surely we can use QueryBuilder, but I'm sure TypeORM needs a simple way to do that, too.

`repository.delete(() => "")` should work, although it is a workaround.

NenadJovicic:
@MajidJafari @andreasblueher you can use `repository.clear()` it will delete everything from that table

[...]

PR #1691, opened 2018-03-02
fix for #1680 - Delete & Updates checks for empty criterias

alex-red:
Let me know if any changes are needed :)

### DIFF
--- a/src/entity-manager/EntityManager.ts
+++ b/src/entity-manager/EntityManager.ts
@@ -357,12 +357,19 @@ export class EntityManager {
     }
 
     /**
-     * Updates entity partially. Entity can be found by a given conditions.
+     * Updates entity partially. Entity can be found by a given condition(s).
      * Unlike save method executes a primitive operation without cascades, relations and other operations included.
      * Executes fast and efficient UPDATE query.
      * Does not check if entity exist in the database.
+     * Condition(s) cannot be empty.
      */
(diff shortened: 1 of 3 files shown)


src/entity-manager/EntityManager.ts
```diff
@@ -357,12 +357,19 @@ export class EntityManager {
...
      * Unlike save method executes a primitive operation without cascades, relations and other operations included.
      * Executes fast and efficient UPDATE query.
      * Does not check if entity exist in the database.
+     * Condition(s) cannot be empty.
      */
     update<Entity>(target: ObjectType<Entity>|EntitySchema<Entity>|string, criteria: string|string[]|number|number[]|Date|Date[]|ObjectID|ObjectID[]|DeepPartial<Entity>, partialEntity: DeepPartial<Entity>, options?: SaveOptions): Promise<UpdateResult> {
+        
+        // if user passed empty criteria or empty list of criterias, then throw an error
+        if (!criteria || (criteria instanceof Array && criteria.length === 0)) {
+            throw new Error(`Empty criteria(s) are not allowed for the update method.`);
```
pleerock: you need to use Promise.reject(new Error(...)) since this method returns promise, it should be rejected

pleerock: also instead of `!criteria` please add explicit checks: `criteria === undefined || criteria === null`

pleerock:
thank you very much for your contribution!

PR #1695, opened 2018-03-02
cli: removed "s" from some migration commands

### LINKED ISSUE #1281: CLI: migration:run and migration:generate name difference between code and docs
The command to create a new migration is `migrations:create` but docs states `migration:create`. Same goes for `migrations:generate` and `migration:generate` (docs).
It seems that the commands where renamed for 0.0.5
What should change? Do we update docs or change the name back?
I don't know the original reason for the rename but, just looking at the name, it would make more sense to call them `migration:create` and `migration:generate` as only one file/migration is created.

TypeORM version: 0.1.7
Doc page: [Using CLI](http://typeorm.io/#/using-cli/generate-a-migration-from-exist-table-schema)

daniel-lang:
* `migrations:create` becomes `migration:create`
* `migrations:generate` becomes `migration:generate`
* `migrations:revert` becomes `migration:revert`

**BUT** `migrations:run` stays `migrations:run`

fixed #1281

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -61,7 +61,7 @@ This fixed various issues on how real arrays must work
 composite unique constraint, on table level. E.g. `@Unique("uq_id_name", ["id", "name"])`
 * fixed `Oracle` issues, now it will be maintained as other drivers 
 * implemented migrations functionality in all drivers
-* CLI commands changed from `migrations:create`, `migrations:generate` and `migrations:revert` to `migration:create`, `migration:generate` and `migration:revert`
+* CLI commands changed from `migrations:create`, `migrations:generate`, `migrations:revert` and `migrations:run` to `migration:create`, `migration:generate`, `migration:revert` and `migration:run`
 
 ## 0.1.13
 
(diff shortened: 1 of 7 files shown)


pleerock:
But why `migrations:run` stays same? We run `migrations:run` but for reverting we run `migration:revert` ? That sounds bad including the fact that revert multiple migrations. If we won't rename it as well then having `migration:create` and others start with migrations:* that would not be good too. We need some solution.

daniel-lang:
Okay, I thought `migrations:run` would be the only command that works with multiple migrations, which is why I left the "s".
I don't know if renaming very command to `migration:*` would make sense, but it might be the most logical thing to do.
btw, the docs state, that `migration:revert`  only reverts the latest migration.

pleerock:
Okay lets do everything "migration:*". We'll also need to add note in changelog and update docs. Related PR: #1700

daniel-lang:
Okay, I renamed all the commands, updated docs and added a note to the changelog.

pleerock:
Thank you

PR #1701, opened 2018-03-05
Proper default column for postgres

### LINKED ISSUE #1677: Array type default for Postgres always shows update needed in schema:log

jdforsythe:
Fixes #1677 for next branch

See #1678

### DIFF
--- a/src/driver/Driver.ts
+++ b/src/driver/Driver.ts
@@ -121,7 +121,7 @@ export interface Driver {
     /**
      * Normalizes "default" value of the column.
      */
-    normalizeDefault(defaultValue: string|Function): string;
+    normalizeDefault(columnMetadata: ColumnMetadata): string;
 
     /**
      * Normalizes "isUnique" value of the column.
--- a/src/driver/mongodb/MongoDriver.ts
+++ b/src/driver/mongodb/MongoDriver.ts
@@ -255,7 +255,7 @@ export class MongoDriver implements Driver {
     /**
      * Normalizes "default" value of the column.
      */
-    normalizeDefault(defaultValue: string): string {
+    normalizeDefault(columnMetadata: ColumnMetadata): string {
         throw new Error(`MongoDB is schema-less, not supported by this driver.`);
     }
 
--- a/src/driver/mysql/MysqlDriver.ts
+++ b/src/driver/mysql/MysqlDriver.ts
@@ -401,7 +401,9 @@ export class MysqlDriver implements Driver {
     /**
      * Normalizes "default" value of the column.
      */
-    normalizeDefault(defaultValue: any): string {
+    normalizeDefault(columnMetadata: ColumnMetadata): string {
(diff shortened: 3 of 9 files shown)


pleerock:
You did lot of non necessary changes like replacing `defaultValue: string|Function` to `columnMetadata: ColumnMetadata`. Please be careful with changes you bring.

About the problem itself, Im not sure if appending type automatically will not lead to other problems. (for example does type we have always match type user specifies? What if default will be wrapped into a function, for example `xxx(column)`, etc.)

Isn't it better if user simply specify array type on his own?

jdforsythe:
About the parameter change, see my [comment on the issue](https://github.com/typeorm/typeorm/pull/1678#issuecomment-370530178). There was a change in `next` to only pass in the `columnMetadata.defaultValue` string instead of the whole object (as is the case in `master`). In order to know if the column is an array, you also need access to the `columnMetadata.isArray` boolean, so the whole object must be passed in. The parameters have to be the same across all the different drivers, so I changed all references to the function to pass or expect the full `ColumnMetadata`.

If you see the problem in #1677, postgres returns the array default as an array literal with a cast, so without this the `schema:log` and `schema:sync` will _always_ report there is a change to the default value for every array column.

I'm not sure what you mean about the user type - isn't `columnMetadata.type` the type the user specified in the model? The default column as a function is handled - I added the cast to the resulting value of the function.

If you have a different way to fix this bug, I'd be willing to help

pleerock:
Thank you very much.

I mean what if we enforce people to do following:

```ts
@Column({
  nullable: false,
  type: 'smallint',
  array: true,
  default: '{5}::smallint[]',
})
```

Instead of simply {5} ?

pleerock:
Okay, nervermind. 

Yes, we can merge this, however I removed your code from function-style value, I want to left full  control for users who don't want to have this type applied.

Issue #1694, opened 2018-03-02
Timezone handling error in DateUtils.mixedDateToDate()


dario1985:
**Issue type:**

[ ] question
[x ] bug report
[ ] feature request
[ ] documentation issue

**Database system/driver:**

[ ] `cordova`
[ ] `mongodb`
[ ] `mssql`
[x ] `mysql` / `mariadb`
[ ] `oracle`
[ ] `postgres`
[ ] `sqlite`
[ ] `sqljs`
[ ] `websql`

**TypeORM version:**

[x ] `latest`
[ ] `@next`
[x ] `^0.1.13` (or put your version here)

**Steps to reproduce or a small repository showing the problem:**

```
const datestr = '2018-03-01T10:00:00.000Z' // with Timezone explicit to UTC!
DateUtils.mixedDateToDate(datestr)
> 2018-03-01T10:00:00.000Z
DateUtils.mixedDateToDate(datestr, true) // e.g. used on MySQL
> 2018-03-01T09:00:00.000Z
DateUtils.mixedDateToDate(datestr) === DateUtils.mixedDateToDate(datestr, true)
> false // should be true
```

What is the use of the [`toUtc`](https://github.com/typeorm/typeorm/blob/82b1410f0e4d7e481f596874fa5c303e3c50019c/src/util/DateUtils.ts#L47) flag, when should apply? It can cause double timezone conversions issues

dario1985:
Nice! I think this it's pretty critical as it affects all no-utc systems. 
When is expected next branch to be released? May this fix can be released in the current 1.x too!

AlexMesser:
Dates behavior changed in `0.2.0-alpha.29`. Please check it via `npm i typeorm@next`

dario1985:
IHMO this is not a feature... it's a bug and it should be fixed also in the current stable branch too (as people may want to use it on production instead of a next/alpha lol)

pleerock:
Released in `0.1.18`. Let us know if you still have issues.

Issue #1328, opened 2017-12-11
timezone issue


taehyoung:
[edit]
Hello,
I set timezone +09:00 in node and mysql.

```js
user.date = new Date()
console.log(new Date()) // 2017-12-11T05:33.15.467Z
```

I expected user.date to be saved as “2017-12-11 14:33.15” but the actual value was “2017-12-11 05:33:15".

I want “2017-12-11 14:33.15” to be saved in the local time. 

I tracked the log and found the parameter was “2017-12-10T20:33:15.451Z” in the query statement. it was not “2017-12-11T05:33:15.451Z”.

I thought the problem is that converting is executed one more internally.

so, what should I do if it is saved as the local time?

Should I set timezone +18:00?

AlexMesser:
Dates behavior changed in `typeorm@0.2.0-alpha.29`. Please check it via `npm i typeorm@next`

pleerock:
Released in `0.1.18`.

Issue #1240, opened 2017-11-26
Wrong timezone using date in find method


TheOnlyReadOnly:
Let's take this entity:

```javascript
@Entity("users")
export class User {

    @PrimaryGeneratedColumn()
    id: number;

    @Column({ length: 100 })
    name: string;

    @Column({ name: 'birth_date' })
    birthDate: Date;
}
```
and the following code:

```javascript
    let birthDate = new Date(2017, 1, 15);

    let user = new User();
    user.name = "TEST";
    user.birthDate = birthDate;

    let usersRepo = connection.getRepository(User);
    usersRepo.save(user).then((saved: User) => {

        console.log(saved);

        usersRepo.find({ birthDate: birthDate }).then((found: User[]) => {

            console.log(found);

        }).catch(error => console.log(error));
    }).catch(error => console.log(error));
```
with logging enabled this what I get:

```
executing query:  START TRANSACTION
executing query:  INSERT INTO `users`(`name`, `birth_date`) VALUES (?,?) -- PARAMETERS: ["TEST","2017-02-14T22:00:00.000Z"]
executing query:  COMMIT
User { name: 'TEST', birthDate: 2017-02-14T23:00:00.000Z, id: 2 }
executing query:  SELECT `User`.`id` AS `User_id`, `User`.`name` AS `User_name`, `User`.`birth_date` AS `User_birth_date` FROM `users` `User` WHERE`User`.`birth_date`=? -- PARAMETERS: ["2017-02-14T23:00:00.000Z"]
[]
```
When inserting entity birthDate is translated to `2017-02-14T22:00:00Z` but on DB I read `2017-02-14 23:00:00`. My timezone is CET, should be UTC+1 in winter so console value is wrong while DB value is correct. Using find method translation to UTC appears correct in console (`14T23:00:00Z`)  but nothing is found. May be it's searching for `15T00:00:00Z`?

TypeORM: 0.1.6 (**Works fine with TypeORM 0.1.0-alpha.32**)
Node: 8.0.0
Database: MySql 5.5

makamekm:
I see the save problem:
```
    let connection = await ORM.Manager.Connect();
    let mtiRepository = connection.getRepository(MTI);

    const date = moment(new Date()).subtract(2, 'minutes').toDate();

    const course = await mtiRepository.createQueryBuilder('course')
    .where("(course.last_date IS NULL OR course.last_date < :date) AND course.paused = :paused AND course.done = :done")
[...]

Issue #1189, opened 2017-11-16
[mysql] select entity date type column, but shifted 1 day.


taehyoung:
typeorm 0.1.2

```js
@Entity()
export class User {
    @PrimaryGeneratedColumn()
    idx: number;

    @Column()
    name: string;

    @Column("date")
    date: string;
}
```

I make user and save user,

```js
 let tempUser = new User();
 tempUser.name = "mike";
 tempUser.date = "2017-11-11";
 await getManager().getRepository(User).save(tempUser);
```

and then 

```js
let user = await getManager().getRepository(User).findOneById(1);
```

I expect user.date = "2017-11-11" but i got "2017-11-10"

so, I run raw query but i got "2017-11-10T06:00:00.000Z"

is this bug?

daniel-lang:
Hey, are you running your mysql database and node on different systems?
It is possible, that your systems use different timezones?

I would suggest checking out the `timezone` option to compensate these differences.

taehyoung:
@daniel-lang 
thank you for answering me.
yes. my timezone option is +09:00 in connection option. (MySQL also)
I want to use { localTimezone: true } in @Column(). but it is deprecated.
what should I do?

daniel-lang:
So node and MySQL have the same timezone? If so, you shouldn't have to specify the `timezone` option.

taehyoung:
I remove `timezone` option in connectionOption.
But, `@CreateDateColumn()` or `@UpdateDateColumn` are saved as UTC time.
and `new Date() ` or `'2017-11-24 13:00:00'` are also save as UTC time.
(ex: I expect 'DATETIME' type column '2017-11-24 13:00:00' in mysql, but saved '2017-11-24 04:00:00')

daniel-lang:
Dates being saved in UTC is the expected behavior. But when node and mysql have the same timezone, dates should be correct when loaded.
@pleerock Do you have an idea?

pablocarreraest:
@daniel-lang  is this the problem of init process.env.TZ before you start node...

AlexMesser:
Dates behavior changed in `typeorm@0.2.0-alpha.29`. Please check it via `npm i typeorm@next`

pleerock:
Released in `0.1.18`. Let us know if you still have issues.

taehyoung:
thank you.
Now, this is work.

PR #1717, opened 2018-03-07
send timestamp to database without converting it into UTC

### LINKED ISSUE #1716: Proper handling of timestamps with timezone information
**Issue type:**

[ ] question
[x] bug report
[ ] feature request
[ ] documentation issue

**Database system/driver:**

[ ] `cordova`
[ ] `mongodb`
[ ] `mssql`
[ ] `mysql` / `mariadb`
[ ] `oracle`
[x] `postgres`
[ ] `sqlite`
[ ] `sqljs`
[ ] `websql`

**TypeORM version:**

[x] `latest`
[ ] `@next`

Timestamps should not be converted by the ORM to UTC as the database is perfectly capable of handling these inputs. It should be the developers responsibility/choice how timezones and offsets are handled correctly.
I do agree that at the end of the day timestamps should be stored in UTC so that any client can later derive the correct date/time from it based on the local timezone.
This means it should be possible to send in timestamps with timezone information as well.
Right now the ORM strips away that important information.

Imagine having the app server with typeorm and the database at UTC+1.
If I want to update a timestamp (with timezone) with:
```
2018-03-07 14:00:00+05
```
I expect to see the following value stored in the database:
```
2018-03-07 10:00:00+01
```
which is basically the same as:
```
2018-03-07 09:00:00+00
```
But unfortunately I end up having an offset of 1 hour in the database due to my server's timezone:
```
2018-03-07 09:00:00+01
```
In a different timezone the deviation would be even worse.

This issue obviously has a huge impact on time-driven applications in a global market. The solution is quite simple though: Let the developer decide how to apply timezones.

btw, great module 🙌 , I appreciate the work you've put into this project @pleerock (and all the others).

chriskalmar:
fixes #1716


pleerock:
What about other drivers?

chriskalmar:
If you agree with this approach I'll be more than happy to continue on this for the other drivers. 
I'll just need some time to check how the other engines are handling timezones. MySQL should be an easy one and Oracle should be pretty close to Postgres, don't know about SQL Server though.

pleerock:
Can you please review [all issues related to timezone](https://github.com/typeorm/typeorm/issues?q=is%3Aopen+is%3Aissue+milestone%3A0.2.0) I guess this will fix some of them.

chriskalmar:
sure, will have a look at them

pleerock:
Thank you. Few important notices however:

1. tests are failing for mssql right now
2. please make PR clean so it can be easily reviewed
3. try to keep test as simple as possible, currently it looks like a bit complex, in tests its completely okay to have duplication, the most important is simplicity

chriskalmar:
As per request I rewrote all tests and rebased my commits. 
The mssql tests were failing because there is no SQL Server running on travis (ormconfig), so I made each driver test optional based on the current ormconfig. 

I went through all relevant issues and I believe the following will be fixed with this PR:
#1716, #1694, #1240, #976, #838 and #1328 
(possibly even #1189)

pleerock:
This change looks like it breaking things (even if it was wrong) and I would like to avoid merging it into the master. What do you think?

chriskalmar:
I absolutely understand your concern, but in my experience it's always better to fix problems like that as soon as possible. Of course you have users of typeorm that are already used to (or simply not aware of) the current timestamp/timezone issue and a fix in master would introduce a breaking change for them. 

But on the other hand - due to the growing popularity of typeorm - you have to think of all the new users you get every day and it would be a shame to let them run into problems that could be fixed already. 
PR  #1765 fixes the issue for the next branch, but in my understanding it's not quite ready for prime time. So naturally most users will use the current master releases.

The way I believe it could work out for both, existing and new users, is a compromise:
What if we would introduce a flag or configuration option named e.g. `LEGACY_TIMEZONE_SUPPORT`.
By default typeorm would use the proper timezone handling, but for those that are already having a production system with wrong timezones this flag could bring back the behaviour they are already used to. 

With this approach you enable new users with a correct timestamp/timezone behaviour from day one, and at the same time you provide full backwards compatibility to existing users. And if they decide to upgrade their database to reflect correct timestamps some day it will be their own choice. 

I think everybody would be satisfied. What do you think?

br0wn:
I would just note that this now affects `@CreateDateColumn` and  `@UpdateDateColumn` dates as, if I understood correctly, values for those are set with only `new Date()`. Therefore, if the server is running in a timezone, date stored in the database will be stored in the timezone of the server for those columns.

chriskalmar:
@br0wn 
`new Date()` creates a timestamp based on UTC. Depending on your timezone you would see a given representation including timezone information. But still, it is UTC.
So the database will store it in UTC as well, which is the correct behaviour.

jwclark:
This appears to not be working as intended in 0.2.6 using postgres. The entity and test call below result in a timestamp in server time(rather thanUTC) and without timezone information. 

@Entity()
export class DateTest {

    public id?: string;

    @Column()
    public date?: Date;
}

testRepository.save({date: new Date()});

jwclark:
This has been going back and forth since at 466: https://github.com/typeorm/typeorm/issues/466. This was working before.

VinceOPS:
Same issue encountered with PostgreSQL

PR #1157, opened 2017-11-09
0.2.0 release changes

pleerock:
everything is going to land into 0.2.0 final. You can try next TypeORM version by installing it `npm i typeorm@next`

PR #1744, opened 2018-03-12
next - cordova and websql inserts

daniel-lang:
This is the fix to address my comment in #1157 and fix the inserts for websql and cordova.


pleerock:
Thank you very much

Issue #1749, opened 2018-03-13
Can't delete entities from tables in non-default schema


adenhertog:
**Issue type:**

[ ] question
[x] bug report
[ ] feature request
[ ] documentation issue

**Database system/driver:**

[ ] `cordova`
[ ] `mongodb`
[ ] `mssql`
[ ] `mysql` / `mariadb`
[ ] `oracle`
[x] `postgres`
[ ] `sqlite`
[ ] `sqljs`
[ ] `websql`

**TypeORM version:**

[x] `latest`
[ ] `@next`
[ ] `0.x.x` (or put your version here)

**Steps to reproduce or a small repository showing the problem:**

Given the following entity declaration:
```typescript
@Entity('bar', { schema: 'foo'})
class Bar {
  ...
}

await barRepository.deleteById(...)
```

I expect the executed SQL to be:
```sql
DELETE from foo.bar where id = ...
```

But instead I get
```sql
DELETE from bar where id = ...
```

This errors as `bar` is not in the default schema

PR #1750, opened 2018-03-13
#1749: Can't delete entities from tables in non-default schema

adenhertog:
This is a fix to issue #1749 that where an entity from a table created in a non-default schema can't be deleted as the schema isn't included by the executor.


pleerock:
@adenhertog thank you for contribution.

Can you please create PR against `@next` branch, but before please check if it already fixed there. `@next` is going to be released soon and it contain ton of fixes, and this change is conflicting with next since it does not have files you changed.

adenhertog:
@pleerock No problem, I'll have a look in `@next` to see if the same bug exists and raise a separate PR

This PR is a fix for an open bug in `master` - is this branch still supported, and will you still consider merging this? I'm currently using `v0.1.16` and can't delete anything from tables in the non-default schemas.

pleerock:
Thank you for contribution!

Issue #1736, opened 2018-03-10
Milliseconds are parsed incorrectly from datetime string


andreasrueedlinger:
Thanks to unit testing if found the following bug: datetime strings like `"2018-03-10T16:13:08.002Z"` (milliseconds below 10) are parsed as `"2018-03-10T16:13:08.020Z"` since the follwing code adds only one leading zero:
https://github.com/typeorm/typeorm/blob/82b1410f0e4d7e481f596874fa5c303e3c50019c/src/util/DateUtils.ts#L202

AlexMesser:
fixed and released in `typeorm@0.2.0-alpha.36`. You can try it via `npm i typeorm@next`.

PR #1757, opened 2018-03-14
Fix date conversion issue

zzd258147:
I'm working on a Ionic project, I have a problem that the date always display as "Invalid Date".

After some research, I found:

If the format of the date string is "2018-03-14 02:33:33.906", Safari (and iOS WKWebView) will convert it to an invalid date object.
We need to modify the date string to "2018-03-14T02:33:33.906Z" and Safari will convert it correctly.

ISO 8601
https://www.w3.org/TR/NOTE-datetime

untitled2

### DIFF
--- a/src/util/DateUtils.ts
+++ b/src/util/DateUtils.ts
@@ -14,7 +14,16 @@ export class DateUtils {
         if (!mixedDate)
             return mixedDate;
 
-        const date = typeof mixedDate === "string" ? new Date(mixedDate) : mixedDate as Date;
+        /**
+         * Fix date conversion issue
+         * 
+         * If the format of the date string is "2018-03-14 02:33:33.906", Safari (and iOS WKWebView) will convert it to an invalid date object.
+         * We need to modify the date string to "2018-03-14T02:33:33.906Z" and Safari will convert it correctly.
+         *
+         * ISO 8601
+         * https://www.w3.org/TR/NOTE-datetime
+         */
+        const date = typeof mixedDate === "string" ? (!isNaN(new Date(mixedDate).getTime()) ? new Date(mixedDate) : new Date(mixedDate.replace(" ", "T") + "Z")) : mixedDate as Date;
         // if (!storedInLocal) {
 
         // else if it was not stored in local timezone, means it was stored in UTC


pleerock:
Not sure why this change must be part of ORM. You need to provide a date in a valid format if you have it in a non-acceptable format.

zzd258147:
Our project is using ionic 3.9.2, the target platform is iOS(iPad).

We use typeorm and cordova-sqlcipher-adapter to access local SQLite DB.

We add createdDate and updatedDate column for all the tables.

```
    @CreateDateColumn({ nullable: true })
    createdDate: string;

    @UpdateDateColumn({ nullable: true })
    updatedDate: string;
```

But if I save an entity into DB, then execute select, the date column will be "Invalid Date".

```
    let a1: AgentProfile = new AgentProfile();
    a1.agentCode = "123";
    await a1.save();

    let result: AgentProfile = await AgentProfile.findOne();
    console.log("a2.createdDate:" + result.createdDate);
```

The console of Xcode:
untitled4

The web inspector of Safari:
untitled6

untitled7

untitled8

The @CreateDateColumn() is auto manage by typeorm.

When you call entity.save() for the first time, typeorm will get current time then  convert it into date string.

AbstractSqliteDriver:
untitled10

DateUtils:
untitled11

If the date save into SQLite DB in this format, then execute select. Typeorm will use DateUtils to convert it back to date, here the problem arises.

As the AbstractSqliteDriver and DateUtils both are a part of typeorm, I need to fix this issue in DateUtils (or in AbstractSqliteDriver).

Please consider merging this PR, or give me some advise.

Solution:
untitled9

More info:
https://stackoverflow.com/questions/4310953/invalid-date-in-safari

pleerock:
adding @daniel-lang

daniel-lang:
This is what I found on this issue:
The SQlite function `datetime()` which is used as the default value for `@CreateDateColumn()` and `@UpdateDateColumn()` returns a date with the format `YYYY-MM-DD HH:MM:SS` which seems to be fine with SQlite (at least if you use `string`/`varchar` as the column type).
Selecting the columns seems to not work in safari because it is not compatible with the formats it supports.

So I guess we will need a fix like @zzd258147 proposed, but in a more readable form (not in a single line). But I'm not sure where the best place for such a fix might be.

zzd258147:
I'd like to help fix this issue, please advise where to fix it.

I can make this fix more readable.

zzd258147:
Hi @pleerock, can you give me some advise?

pleerock:
If this change affect only sqlite databases then probably its better to apply change in `AbstractSqliteDriver#prepareHydratedValue` method since Im not sure how this change in DateUtils affect other drivers. Also please take a look why tests are failing

zzd258147:
OK, I will create a new PR to fix it.

zzd258147:
Last time the tests are failing because: `typeorm/test/github-issues/1736/issue-1736.ts`

```
    it("should correctly parse milliseconds from datetime string", () =>  {
        expect(DateUtils.mixedDateToDatetimeString("2018-03-10T16:13:08.002Z")).to.be.equal("2018-03-10 21:13:08.002");
        expect(DateUtils.mixedDateToDatetimeString("2018-03-10T16:13:08.020Z")).to.be.equal("2018-03-10 21:13:08.020");
        expect(DateUtils.mixedDateToDatetimeString("2018-03-10T16:13:08.200Z")).to.be.equal("2018-03-10 21:13:08.200");

        expect(DateUtils.mixedDateToUtcDatetimeString("2018-03-10T16:13:08.002Z")).to.be.equal("2018-03-10 16:13:08.002");
        expect(DateUtils.mixedDateToUtcDatetimeString("2018-03-10T16:13:08.020Z")).to.be.equal("2018-03-10 16:13:08.020");
        expect(DateUtils.mixedDateToUtcDatetimeString("2018-03-10T16:13:08.200Z")).to.be.equal("2018-03-10 16:13:08.200");
    });
```

#1736 Milliseconds are parsed incorrectly from datetime string

But I'm not sure why `2018-03-10T16:13:08.002Z` should be equal to `2018-03-10 21:13:08.002`?

Is this related to the time zone of the CI server?

@AlexMesser is work on this and skipped this test temporary, so it passed all the tests now.

Issue #1761, opened 2018-03-15
Cascade Remove Set Foreign Key to Null Fails When Inverse Side Is Primary


kristophjunge:
**Issue type:**

[ ] question
[x] bug report
[ ] feature request
[ ] documentation issue

**Database system/driver:**

[ ] `cordova`
[ ] `mongodb`
[ ] `mssql`
[x] `mysql` / `mariadb`
[ ] `oracle`
[ ] `postgres`
[ ] `sqlite`
[ ] `sqljs`
[ ] `websql`

**TypeORM version:**

[ ] `latest`
[x] `@next`
[ ] `0.x.x` (or put your version here)

**Steps to reproduce or a small repository showing the problem:**

The steps for reproduction are supplied in the unit test in the related pull request.

---

Hello,

thank you for the great work on typeorm so far!

The use case here is to have a many to many relation over a relation table (Comment) with additional properties (message) using two one-to-many relations.

```
+----------+       +-----------+
|Post      +       |Comment    |       +----------+
+----------+  1:x  +-----------+       |User      |
|id (PK)   | <---> |postId (PK)|  x:1  +----------+
|comments[]|       |userId (PK)| <---> |id (PK)   |
+----------+       |message    |       |comments[]|
                   +-----------+       +----------+
```

The primary key of `Comment` is a combination of the two foreign keys.

In this use case i want to use eager loading together with cascade "insert", "update" and "remove".

Here are the entity definitions:

```ts
@Entity()
export class Post {

    @PrimaryGeneratedColumn()
    id: number;

    @OneToMany(type => Comment, comment => comment.post, {cascade: ["insert", "update", "remove"], eager: true})
    comments: Comment[];

}
```

```ts
@Entity()
export class User {

    @PrimaryGeneratedColumn()
    id: number;

    @OneToMany(type => Comment, comment => comment.user)
    comments: Comment[];

}
```

```ts
@Entity()
export class Comment {

    @ManyToOne(type => User, user => user.comments, {primary: true})
    user: User;

    @ManyToOne(type => Post, post => post.comments, {primary: true})
    post: Post;

    @Column()
    message: string;

}
```

Cascade `insert` and `update` are working like a charm but when it comes to cascade `remove` it triggers the following error:

```
QueryFailedError: ER_BAD_NULL_ERROR: Column 'postId' cannot be null
```

The reason is that only the logic is implemented to detach a relation by setting its foreign key field to NULL. However there are situations where that is not working.

[...]

PR #1762, opened 2018-03-15
Cascade Remove Set Foreign Key to Null Fails When Inverse Side Is Primary

kristophjunge:
See #1761

### DIFF
--- a/src/persistence/subject-builder/OneToManySubjectBuilder.ts
+++ b/src/persistence/subject-builder/OneToManySubjectBuilder.ts
@@ -165,13 +165,22 @@ export class OneToManySubjectBuilder {
                 const removedRelatedEntitySubject = new Subject({
                     metadata: relation.inverseEntityMetadata,
                     parentSubject: subject,
-                    canBeUpdated: true,
                     identifier: removedRelatedEntityRelationId,
-                    changeMaps: [{
+                });
+
+                // in case the inverse relation is primary we cannot remove the relation by setting the
+                // foreign key field to null. Instead remove the entity.
+                if (relation.inverseRelation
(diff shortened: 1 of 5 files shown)


src/persistence/subject-builder/OneToManySubjectBuilder.ts
```diff
@@ -165,13 +165,22 @@ export class OneToManySubjectBuilder {
...
                     metadata: relation.inverseEntityMetadata,
                     parentSubject: subject,
-                    canBeUpdated: true,
                     identifier: removedRelatedEntityRelationId,
-                    changeMaps: [{
+                });
+
+                // in case the inverse relation is primary we cannot remove the relation by setting the
+                // foreign key field to null. Instead remove the entity.
+                if (relation.inverseRelation
```
pleerock: WE CANNOT SIMPLY REMOVE INVERSE ENTITY IF CASCADE WERE NOT DEFINED. Where cascade remove check is?

pleerock:
Cascade removes is a complex topic. If you want this to be merged then we'll need to implement all cascade remove features. This PR currently covers only one to many case. We'll need to implement this for other relations as well.

src/persistence/subject-builder/OneToManySubjectBuilder.ts
pleerock: also, this case only covers when foreign key of entity is a primary key, and does not cover simple cases. Cascade remove logic should not be tight to the fact of foreign key column being primary or not.

pleerock:
Any updates?

leandromsa:
hello, exist some solution while there is no fix realeased?

Kononnable:
Closing as no response from the author.

Issue #976, opened 2017-10-03
Is it internally converting the date object to UTC before saving.


tontonskie:
I have an entity like this:

```
@Entity('jobs')
export class Job {

  @Column(type => DateTimeWithOffset)
  start: DateTimeWithOffset;
}

export class DateTimeWithOffset {

  @Column({ name: '_offset' })
  offset: number;

  @Column('datetime', { name: '_date' })
  date: Date;
}
```

I'm saving it like this:
```
let job = new Job();
job.start = new DateTimeWithOffset()
job.start.offset = offset;
job.start.date = moment('2017-11-22 16:00').toDate();
job.save()
```

But what stored in database is `2017-11-22 08:00:00`. I check the moment object and it is parsed correctly and `toDate()` return is correct.

I'm using mysql 5.7 and my machine is in `Asia/Manila` timezone `+08:00` 

it seems like internally the date object is converted to UTC before saving.

pleerock:
Yes, it is., dates should be stored in utc in the database to avoid all kind of issues.

tontonskie:
@pleerock thanks for quick response.

Is there a typeorm way to converting this to local time when fetching ?

pleerock:
it does it, to be precision javascript itself does it.

bartlomiej-korpus:
@tontonskie check out node's TZ env variable, it affects the way dates are parsed.

pleerock:
any updates on this issue?

Diluka:
mysql `DATETIME` should store as mysql server's localtime using `timezone` config, `SHOW VARIABLES WHERE Variable_name = 'time_zone';` or `SELECT NOW();` to calibrate.

If you want to store UTC time, you should store it into `TIMESTAMP` column.

Do not store UTC time in localtime column(`DATETIME`), you are messing up my database.
If the database is migrated to a different time zone server, it will be disastrous.

tontonskie:
@pleerock i’m already okay on this issue

dhcmrlchtdj:
@pleerock, I have the same problem.

https://github.com/typeorm/typeorm/blob/master/src/driver/mysql/MysqlDriver.ts#L305
https://github.com/typeorm/typeorm/blob/master/src/util/DateUtils.ts#L48

I think it is a bug with `new Date()` at `DateUtils`.
Maybe we should use `Date.UTC()` instead of `new Date()`.

[...]

Issue #1261, opened 2017-11-29
onDelete property on foreign key is not modified on sync


Ionaru:
**Reproduction code**
```typescript
import { BaseEntity, createConnection, Entity, JoinColumn, OneToOne, PrimaryGeneratedColumn } from 'typeorm';
import 'reflect-metadata';

@Entity()
class Foo extends BaseEntity {

    @PrimaryGeneratedColumn()
    public id: number;
}

@Entity()
class Bar extends BaseEntity {

    @PrimaryGeneratedColumn()
    public id: number;

    @OneToOne(type => Foo, {
        onDelete: 'RESTRICT'
    })
    @JoinColumn()
    foo: Foo;
}

async function run() {
    await createConnection({
        type: 'sqlite',
        database: 'test.db',
        entities: [
            Foo, Bar
        ],
        logging: 'all',
        synchronize: true,
    });

    const foo = new Foo();
    const bar = new Bar();
    bar.foo = foo;
    await foo.save();
    await bar.save();
    await foo.remove();
}

run().then();
```
**Run steps**
1. Run the program once, it will error because of a foreign key constraint when deleting `foo` (correct).
2. For the property `onDelete` change `RESTRICT` to `SET NULL` or `CASCADE`
3. Run the program again, it will error because of a foreign key constraint when deleting `foo`. (wrong)

**Info**
We changed the foreign key `onDelete` property, but this change was not saved to the database. This resulted in a foreign key constraint error.

I don't know if this is the only property that is not saved, there might be others.

pleerock:
Its a bug. @AlexMesser working currently on schema synchronization refactoring and we'll ship all new changes and this fix as well in 0.2.0

Also please take a look how I added test for your issue, you can same next time ;)

stonecold123:
When the table already exists, I'm not able to add/update the onDelete property as well for the relationship.
I have to add it manually or delete the entire table to add the onDelete property.

AlexMesser:
fixed and released in `typeorm@0.2.0-alpha.23`. You can try it via `npm i typeorm@next`.

PR #1765, opened 2018-03-16
send timestamp to database without converting it into UTC (next branch)

### LINKED ISSUE #1716: Proper handling of timestamps with timezone information

chriskalmar:
fixes #1716, #1694, #1240, #976, #838 and #1328 
(next branch PR of #1717)

### DIFF
--- a/src/driver/mysql/MysqlDriver.ts
+++ b/src/driver/mysql/MysqlDriver.ts
@@ -316,8 +316,8 @@ export class MysqlDriver implements Driver {
         } else if (columnMetadata.type === "json") {
             return JSON.stringify(value);
 
-        } else if (columnMetadata.type === "datetime" || columnMetadata.type === Date) {
-            return DateUtils.mixedDateToDate(value, true);
+        } else if (columnMetadata.type === "timestamp" || columnMetadata.type === "datetime" || columnMetadata.type === Date) {
+            return DateUtils.mixedDateToDate(value);
 
         } else if (columnMetadata.type === "simple-array") {
             return DateUtils.simpleArrayToString(value);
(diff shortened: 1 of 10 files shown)


chriskalmar:
looks like the build failed due to an unrelated problem (#1261)
```
  github issues > #1261 onDelete property on foreign key is not modified on sync
    1) should order by added selects when pagination is used
```

AlexMesser:
thank you for your contribution!

chriskalmar:
thanks, any chance your could review my PR #1717 for the same issue on the master branch as well?

PR #1773, opened 2018-03-17
Skip constructor when entity is loaded from database

### LINKED ISSUE #1772: Skip constructor when entity is loaded from database
**Issue type:**

[ ] question
[x] bug report
[x] feature request
[ ] documentation issue

**Database system/driver:**

[x] `cordova`
[x] `mongodb`
[x] `mssql`
[x] `mysql` / `mariadb`
[x] `oracle`
[x] `postgres`
[x] `sqlite`
[x] `sqljs`
[x] `websql`

**TypeORM version:**

[x] `latest`
[ ] `@next`

Create some entity with non-empty constructor. For example:

```ts
@Entity()
export class Post {

    @PrimaryGeneratedColumn()
    id: number;

    @Column()
    private title: string;

    constructor(title: string) {
        if (title === undefined) {
            throw new Error("Invalid title value");
        }

        this.title = title;
    }

}
```

When entity is loaded from db, `Invalid title value` is thrown, because constructor is invoked without args. I think calling constructor must be skipped somehow in this case.

pepakriz:
fixes #1772

### DIFF
--- a/src/metadata/EntityMetadata.ts
+++ b/src/metadata/EntityMetadata.ts
@@ -19,6 +19,7 @@ import {PostgresDriver} from "../driver/postgres/PostgresDriver";
 import {SqlServerDriver} from "../driver/sqlserver/SqlServerDriver";
 import {PostgresConnectionOptions} from "../driver/postgres/PostgresConnectionOptions";
 import {SqlServerConnectionOptions} from "../driver/sqlserver/SqlServerConnectionOptions";
+import {ObjectInstantiator} from "../util/ObjectInstantiator";
 
 /**
  * Contains all entity metadata.
@@ -442,7 +443,7 @@ export class EntityMetadata {
 
         // if target is set to a function (e.g. class) that can be created then create it
         if (this.target instanceof Function)
-            return new (<any> this.target)();
+            return ObjectInstantiator.createInstanceWithoutConstructor(<any> this.target);
 
         // otherwise simply return a new empty object
         const newObject = {};
--- a/src/util/ObjectInstantiator.ts
+++ b/src/util/ObjectInstantiator.ts
@@ -1,0 +1,10 @@
+export class ObjectInstantiator {
+
+    static createInstanceWithoutConstructor<T>(objectClass: { new (): T }): T {
+        class NewObject {}
+        NewObject.prototype = objectClass.prototype;
+
+        return <T> new NewObject();
+    }
+
+}
--- a/test/github-issues/1772/entity/Post.ts
+++ b/test/github-issues/1772/entity/Post.ts
@@ -1,0 +1,25 @@
+import {Entity} from "../../../../src/decorator/entity/Entity";
+import {Column} from "../../../../src/decorator/columns/Column";
+import {PrimaryGeneratedColumn} from "../../../../src/decorator/columns/PrimaryGeneratedColumn";
+
+@Entity()
+export class Post {
+
+    @PrimaryGeneratedColumn()
+    id: number;
+
+    @Column()
+    private title: string;
+
+    initialized?: true;
+
+    constructor(title: string) {
+        this.title = title;
+        this.initialized = true;
+    }
+
+    public getTitle(): string {
+        return this.title;
+    }
+
+}
--- a/test/github-issues/1772/issue-1772.ts
+++ b/test/github-issues/1772/issue-1772.ts
@@ -1,0 +1,35 @@
+import "reflect-metadata";
+import {expect} from "chai";
+import {closeTestingConnections, createTestingConnections, reloadTestingDatabases} from "../../utils/test-utils";
+import {Connection} from "../../../src/connection/Connection";
+import {Post} from "./entity/Post";
+
+describe("github issues > #1772 Skip constructor when entity is loaded from database", () => {
+
+    let connections: Connection[];
+    before(async () => connections = await createTestingConnections({
+        entities: [Post],
+        dropSchema: true
+    }));
+    beforeEach(() => reloadTestingDatabases(connections));
+    after(() => closeTestingConnections(connections));
+
+    it("should not call constructor", () => Promise.all(connections.map(async connection => {
+
+        const postRepository = connection.getRepository(Post);
+
+        // create and save a post first and check entity is initialized
+        const post = new Post("About columns");
+        await postRepository.save(post);
+
+        expect(post.initialized).to.be.true;
+
+        // check if entity loaded from database is not initialized by constructor
+        const loadedPost = await postRepository.findOneById(1);
+        expect(loadedPost).to.be.instanceof(Post);
+        expect(loadedPost!.initialized).to.be.undefined;
+
+    })));
+
+
+});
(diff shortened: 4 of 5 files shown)


pleerock:
Looks like a big hack. What if user **wants** to have something initialized in the constructor?

pleerock:
Im going to close it, thanks for your efforts.

pepakriz:
@pleerock I agree. It is big hack, but correct behavior in my opinion. I was inspired by Doctrine ORM. When entity is deserialized from db, constructor is omitted by reflection. So you have free constructor for initial checks and you don't need write special hack for undefined values.

To answer your question: If someone want to initialize something during deserialization from db, there should be some event (http://typeorm.io/#/listeners-and-subscribers/what-is-an-entity-listener)

vimmerru:
@pleerock 

I agree that calling constructor during deserialization is a wrong behavior and suggest to merge this.

1. You noticed that constructor can have additional logic, but in the most cases this logic is invalid during deserialization process. Very simple example is assigning properties in constructor. As user doesn't expect undefined here it will cause crashing in runtime. Here are multiple bugs reported related to this.
2. It is near impossible to use typeorm in 'strict' compiler mode. Especially, if you want to get some additional safety.
3. Direct assignments of prototypes is more-less well known approach in js world.

pleerock:
Reopening this issue and according PR.

pleerock:
I think I have change my mind since I see people have inconvenience with strict mode. @Kononnable @vlapo @havenchyk what you think about it?

pepakriz:
@pleerock Wow, glad to hear it! Tell me if you need some help

Kononnable:
There are reasons why some libraries requires such classes(e.g. entity classes) to have a constructor without parameters. 

I'm good with both solutions:
- not calling a constructor(since it's deserialization) as long as there is a way to easy propagate fields not saved in db(without a Column decorator)
- use default values and constructor without parameters(current implementation)
We can achieve same results with both of them so it's just a matter of preference.

src/metadata/EntityMetadata.ts
```diff
@@ -442,7 +443,7 @@ export class EntityMetadata {
 
         // if target is set to a function (e.g. class) that can be created then create it
         if (this.target instanceof Function)
-            return new (<any> this.target)();
+            return ObjectInstantiator.createInstanceWithoutConstructor(<any> this.target);
```
XanderEmu: You know you could do something like:
```js
const object = {};
Reflect.setPrototypeOf(object, this.target.prototype);
return object
```

jtremback:
Just want to chime in to say that this makes strict mode difficult. To make strict mode work, you need to use the ! operator on every key in the entity, turning off null checking.

XanderEmu:
Probably.

pleerock:
I can merge this branch if it fix all issues and will be created against `next`

XanderEmu:
So I tried to do just that. I basically copied the code (submitted by @pepakriz into a new branch created against `next`, only I used JS `Reflect` instead of `ObjectInstantiator`).

https://github.com/XanderEmu/typeorm/tree/skip-constructor-when-entity-is-loaded-from-the-database

The tests are failing for `entity-metadata-validator > initialized relations` which `should throw error if relation with initialized array was found on many-to-many relation`.

I'm not sure how I should proceed. Basically, I don't understand why this is the case. Perhaps, with the change proposed by @pepakriz, it is not a problem anymore?

pleerock:
@XanderEmu I'll need to check tests on `next`, maybe they are failing not because of your changes. Just create a PR with all changes you propose, make sure your test is passing and I'll take of the rest once I get a chance to review and merge it.

XanderEmu:
I'm afraid they weren't failing on next. I submitted a PR anyway:
https://github.com/typeorm/typeorm/pull/3845

src/metadata/EntityMetadata.ts
vlapo: What about `Object.create(this.target.prototype);` ? Looks more clear for me.

vlapo:
I am good with both solution too. But I have to say that most of my entity classes have a lot of properties so mostly I use only one parameter with data to initialize and always optional. 

Right now we have some issues and questions about parameters in constructor, after merge we will have question/issues about "non called constructors". So if this change will merged we have to update documentation and inform developers about deserialization process.

jtremback:
@vlapo If you are using strict mode, Typescript complains if you have a non optional property on a class that is not explicitly instantiated in the constructor. This seems like a pretty integral part of type correctness to me.

If you are using an optional argument to instantiate your entities, are you doing a null check every time you access a property on an entity? If not, it seems like you are missing out on some of the correctness checking that Typescript could provide.

src/metadata/EntityMetadata.ts
nicomouss: @vlapo, yes `Object.create` will do the job, also `Reflect.construct`.
@XanderEmu not sure about `Reflect.setPrototypeOf`, see examples below

```js
class Post {
    constructor(title) {
        console.log("Inside constructor for " + title);
        this._title = title;
    }
}

let post1 = new Post("post 1"); // > 'Inside constructor for post 1'
console.log(post1 instanceof Post); // > true (OK !)
console.log(post1._title); // > 'post 1'  (OK !)

let post2 = Reflect.construct(function() {}, [], Post);
console.log(post2 instanceof Post); // > true  (OK !)
console.log(post2._title); // > undefined  (OK !)

const object = {};
let post3 = Reflect.setPrototypeOf(object, Post.prototype);
console.log(post3 instanceof Post); // > false  (NOT OK !)

let post4 = Object.create(Post.prototype);
console.log(post4 instanceof Post); // > true  (OK !)
console.log(post4._title); // > undefined (OK !)
```

nicomouss:
@pepakriz nice work. I was about to submit a PR tackling the same issue, and I stumbled upon your work.

The example you have implemented in your test cases underline the right behavior the ORM should have in my opinion. With this solution, the business process of creating domain object and the technical process of re-creation/hydratation of these pre-existing domain objects are clearly separated.

NHibernate for example requires to have a default constructor on the domain classes (that you would usually hide by marking it as `protected`) to tackle the same design principle of separating the initial object creation (by the user) from their re-creation by the ORM later on.

Here the solution provided is even more elegant, as it does not require `polluting` domain classes with useless empty constructor (useless as they add no business meaning to your domain).

PS: just for information, I was using `return Reflect.construct(function() {}, [], this.target);` in my intended PR, end-result is same apparently I guess.

nicomouss:
> Right now we have some issues and questions about parameters in constructor, after merge we will have question/issues about "non called constructors". So if this change will merged we have to update documentation and inform developers about deserialization process.

Yes documentation should clearly state that TypeOrm will not call any constructor while re-constructing entities. 

But again this design sounds more logical, so I am not sure it will raise more questions than before. In my case the question was more to understand "why is the ORM calling the constructor when re-constructing my entities". I add to put if/else within my constructors to avoid execution of some business code by the ORM, raising more questions from other members of the team asking why these if/else were here for ...

src/metadata/EntityMetadata.ts
nicomouss: @XanderEmu Sorry, my bad, I messed it up, my example with `Reflect.setPrototypeOf` should have been coded properly:

```js
const object = {};
Reflect.setPrototypeOf(object, Post.prototype);
post3 = object;
console.log(post3 instanceof Post); // > true
```

All good then !

pleerock:
Closing in favor of #3845 . @pepakriz thanks for the contribution!

Issue #1779, opened 2018-03-19
SQL math operations apply to string ids in tree entity with closure table


PisaJan:
**Issue type:**

[ ] question
[x] bug report
[ ] feature request
[ ] documentation issue

**Database system/driver:**

[ ] `cordova`
[ ] `mongodb`
[ ] `mssql`
[x] `mysql` / `mariadb`
[ ] `oracle`
[ ] `postgres`
[ ] `sqlite`
[ ] `sqljs`
[ ] `websql`

**TypeORM version:**

[x] `latest`
[ ] `@next`
[ ] `0.x.x` (or put your version here)

**Steps to reproduce or a small repository showing the problem:**

There is a problem with string IDs in case of tree entity with closure table. 
If id is formed like "1/2" mysql transforms the value to 0.5 and therefore for example insert for new entity fails.

Entity example:
```ts
import { PrimaryColumn, TreeChildren, TreeParent, ClosureEntity} from 'typeorm';

@ClosureEntity()
export class Group {

    @PrimaryColumn({ type: 'varchar' })
    public id: string;

    @TreeChildren()
    public children: Group[];

    @TreeParent()
    public parent: Group;

}
```

PisaJan:
Posted PR fixes just MySQL driver, maybe some other drivers would also similar fix.

pleerock:
This was already fixed in `@next` together with other related changes. Can you use `@next` version instead (it will be released in 0.2.0 soon)?

PisaJan:
Okey I will switch to @next for a while. Is there any date set for 0.2.0? Just for the idea. 

Thanks :)

pleerock:
first two weeks of april

PisaJan:
Great, thanks!

PR #1781, opened 2018-03-19
Fix: SQL math operations apply to string ids in tree entity

PisaJan:
Partial fix for issue: https://github.com/typeorm/typeorm/issues/1779


pleerock:
I'll close it since `@next` already has those fixes and will be released quite soon. Thank you very much for your valuable contribution

Issue #1674, opened 2018-02-28
BaseEntity should have reload method


svicalifornia:
**Issue type:** Feature request
**TypeORM version:** `latest`

ActiveRecord (Rails) supports a `reload` method to fetch the latest state of the record from the database and set the ActiveRecord object's state to match:

http://api.rubyonrails.org/classes/ActiveRecord/Persistence.html#method-i-reload

This reload functionality is missing from TypeORM, and this omission forces us to jump through hoops like:

    book = await Book.findOneById(book.id)

This is especially sad when using typeorm-routing-controllers-extensions, which selects objects for us using params — we shouldn't have to add the above object selection logic back in.

Please add the `reload` method to the BaseEntity class as it behaves in Rails ActiveRecord. (Of course this will return a Promise and generally be used with `await`.)

pleerock:
can you please provide a method signature with usages example - how you think it should look like?

svicalifornia:
public reload(): Promise<BaseEntity> {…}

The returned Promise should resolve to a reference to the original entity, so that it can be chained to further methods:

    return (await book.reload()).currentBorrower;

svicalifornia:
Example use cases:

* When saving a new BaseEntity object, the default values of columns not specified are not currently loaded from the database and added to the entity object, so the entity object is missing any default values. Calling `myEntity.reload()` would load any values that are missing from the entity object.
* After updating one or more records via `manager` or `queryBuilder`, if you have a reference to a BaseEntity, you can easily get its new values by calling `myEntity.reload()`.
* Currently a `manager` update is required to set a timestamp (other than create/update timestamps) to the current time. Calling `myEntity.reload()` afterward will get the actual timestamp and put it in the entity object:

    await manager.query('UPDATE project SET "submittedAt" = CURRENT_TIMESTAMP WHERE id = $1', [project.id]);
    await project.reload();

pleerock:
okay, we can do that. Are you able to create a PR for this feature?

svicalifornia:
Sorry, I'd like to help with that, but at the moment I don't have time to dive into the code.

pleerock:
Then I'll left this open. If you really need it then you can do it very quickly, its very trivial feature to implement.

svicalifornia:
Just a note that this method should mutate the original entity object in place. It should **not** simply load and return a second instance of the record.

If I knew more about how BaseEntity objects managed their internal state, then I might contribute a PR, but I haven’t had time to look into that yet.

[...]

PR #1787, opened 2018-03-19
#1674 add reload method to repository

meel-io:


src/repository/Repository.ts
```diff
@@ -345,5 +345,20 @@ export class Repository<Entity extends ObjectLiteral> {
...
+    reload(entity: Entity): Promise<void> {
+        return new Promise(async (resolve, reject) => {
+            const newEntity = await this.findOneById(entity.id);
+            
+            if (!newEntity) {
+                reject("An error ocurred while reloading entity");
+            }
+            
+            Object.assign(entity, newEntity);
+            resolve();
```
pepakriz: Why not resolve newEntity? Why mutate argument? In my opinion `Object.assign` should be in BaseEntity.

src/repository/Repository.ts
```diff
@@ -345,5 +345,20 @@ export class Repository<Entity extends ObjectLiteral> {
     async clear(): Promise<void> {
         return this.manager.clear(this.metadata.target);
     }
-
+    
+    /**
+     * Reloads the supplied entity into the latest state of the record from the database
+     */
+    reload(entity: Entity): Promise<void> {
+        return new Promise(async (resolve, reject) => {
```
pepakriz: Promise can be rewritten to async reload function.

test/functional/repository/reload/repository-reload.ts
```diff
@@ -0,0 +1,54 @@
...
+        post!.text.should.be.eql("This is default text.");
+        assertCategory.should.be.eql({
+            id: 1,
+            name: "Persistence"
+        });
+
+        category.name = "Persistence and Entity";
+        await category.save();
+        
+        await post.reload();
```
pepakriz: I don't like activeRecord-like test when test is about `repository > reload method`, but I am not the boss here :)

meel-io: Yes, I wasn't sure about this either. I think I'll make a repository test instead

src/repository/Repository.ts
meel-io: Fair point, I'll change it

src/repository/Repository.ts
meel-io: I hadn't thought about it. Makes more sense to assign it inside BaseEntity

src/repository/Repository.ts
```diff
@@ -345,5 +345,20 @@ export class Repository<Entity extends ObjectLiteral> {
...
+            const newEntity = await this.findOneById(entity.id);
```
pleerock: Let's make this feature part of `@next` branch which will be released in a few weeks. In `@next` there is a method called `findOneOrFail` that you can use instead of this check

src/repository/Repository.ts
pleerock: TBH I don't find `reload` method useful in repository, so I think the whole method body must be moved into BaseEntity

src/repository/Repository.ts
meel-io: Got it, will do

src/repository/Repository.ts
meel-io: Ok, sounds good to me. I'll update 👍

pleerock:
Thank you very much for your contribution!

Issue #1780, opened 2018-03-19
Support for insertion ignore on duplicate error


dimiork:
**Issue type:**

[ ] question
[ ] bug report
[x] feature request
[ ] documentation issue

**Database system/driver:**

[ ] `cordova`
[ ] `mongodb`
[ ] `mssql`
[x] `mysql` / `mariadb`
[x] `oracle`
[x] `postgres`
[x] `sqlite`
[x] `sqljs`
[ ] `websql`

**TypeORM version:**

[x] `latest`
[ ] `@next`
[ ] `0.x.x` (or put your version here)

**Steps to reproduce or a small repository showing the problem:**

There is special syntax for ignoring insertion on duplicate error in most RDBMS engines.
Here is some research about it:

> MySQL: INSERT **IGNORE** INTO test (x, y) VALUES (1,1);
> SQLite: INSERT **OR IGNORE** INTO test (x, y) VALUES (1,1);
> OracleDB: INSERT /*+ ignore_row_on_dupkey_index(test,unique-idx) */ INTO test (x, y) VALUES (1,1);
> PostgreSQL: INSERT INTO test (x, y) VALUES (1,1) **ON CONFLICT DO NOTHING**;

SQLServer is not support this feature in query request at all (You can only set parameter _ignore_dup_key_ during creation index).

So we have two patterns: exception at end of query for Postgres and between INSERT and INTO keywords for others. 
Additionally there is already implemented QueryBuilder's method "onConflict" that's why ignore syntax can be easy implemented by having changed way to generate insert sql query for all of above cases.

I'll try to provide solution in PR soon.

<!--
    To answer those questions you need to put "x" inside the square brackets, for example:
    [x] `mysql`
    [ ] `postgres`

    Also, please format your code properly (by taking code blocks into ```ts .... ```)
!>

benfavre:
Would also need this for Many-to-many relations and relation().of().add()

EvanDarwin:
What's the status of this issue?

josephbuchma:
Looks like Mysql and Postgres are supported ( #2983 )
Meanwhile, here is generic workaround I use for **sqlite** (ActiveRecord pattern):
```ts
function insertOrIgnore<
  Model extends typeof BaseEntity,
  Instance = InstanceType<Model>
>(
  model: Model,
  values: Array<{ [key in keyof Instance]: Instance[key] }>,
) {
  const q = model
    .createQueryBuilder()
    .insert()
    .into(model)
[...]

PR #1793, opened 2018-03-21
Added support for insertion ignore clause on duplicate error for MySQL/MariaDB, PostgreSQL, SQlite/Sqljs and Oracle databases

dimiork:
See [#1780](https://github.com/typeorm/typeorm/issues/1780)

### DIFF
--- a/src/query-builder/InsertQueryBuilder.ts
+++ b/src/query-builder/InsertQueryBuilder.ts
@@ -4,7 +4,11 @@ import {ObjectType} from "../common/ObjectType";
 import {QueryPartialEntity} from "./QueryPartialEntity";
 import {SqlServerDriver} from "../driver/sqlserver/SqlServerDriver";
 import {PostgresDriver} from "../driver/postgres/PostgresDriver";
+import {MysqlDriver} from "../driver/mysql/MysqlDriver";
+import {OracleDriver} from "../driver/oracle/OracleDriver";
 import {AbstractSqliteDriver} from "../driver/sqlite-abstract/AbstractSqliteDriver";
+import {SqljsDriver} from "../driver/sqljs/SqljsDriver";
+import {SqliteDriver} from "../driver/sqlite/SqliteDriver";
 
 /**
  * Allows to build complex sql queries in a fashion way and execute those queries.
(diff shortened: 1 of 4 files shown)


dimiork:
Something strange is going.
My local copy passed all tests.
This test failed on [basic-eager-relations.ts](https://github.com/typeorm/typeorm/blob/master/test/functional/relations/eager-relations/basic-eager-relations/basic-eager-relations.ts) but it has no any insertion cases.
what should I do?

test/github-issues/1780/issue-1780.ts
```diff
@@ -0,0 +1,176 @@
...
+
+    it("should save two rows without duplicate error in SQLite/Sqljs", () => Promise.all(connections.map(async connection => {
+
+        try {
+            if ([SqljsDriver, SqliteDriver].find(currentDriver => connection.driver instanceof currentDriver)) {
+
+                await connection.manager
+                    .createQueryBuilder()
+                    .insert()
+                    .onConflict("OR IGNORE")
```
pleerock: Maybe its better to add `orIgnore` method into query builder? Because this string looks weird. If we do so, then we can use this method in mysql as well.

pleerock:
For some reasons on CI tests sometimes fails =( I'll restart CI.

src/query-builder/InsertQueryBuilder.ts
```diff
@@ -20,7 +24,8 @@ export class InsertQueryBuilder<Entity> extends QueryBuilder<Entity> {
      */
     getQuery(): string {
         let sql = this.createInsertExpression();
-        return sql.trim();
+        // regexp converting multiline query string into single line
```
pleerock: can you please provide more details why we need this?

test/github-issues/1780/issue-1780.ts
dimiork: Ok, I'll update PR with orIgnore and orUpdate methods.

src/query-builder/InsertQueryBuilder.ts
dimiork: To get more readability I changed sql query from long single line to line with breaks.
It's for what regexp is used.

pleerock: Can you please make another fix of this issue without regexp applied? Im really scary about using regexp in user defined expressions since it can contain string with values which should not be parsed with regexp. Also regexp brings a performance issue with is critical in insertion process when multiple inserts are called.

lambrojos:
@dimiork Sorry to revive this thread, but is this PR dead? This is a feature I would very much like to be merged.. If there is anything I can do to help, let me know :)

pleerock:
Lets close it in favour of #2983

PR #1800, opened 2018-03-22
`this.migrationsTableName` was used before assigned

Mordred:
Hi, I'm testing actual `next` version, but I cannot run migrations because of wrong ordering in the constructor.


pleerock:
Just fixed this issue and found out your PR. Sorry about that. Thank you very much for contribution, it is very valuable. Issue is fixed in latest `@next`

PR #1815, opened 2018-03-23
Fixed the issue that given object may not has hasOwnProperty function

hkjeffchan:
If the given object does not come from standard object, it will not have the "hasOwnProperty" function and raise error.

Lots of library in node (e.g. graphql) is creating object by Object.create(null) and does not "hasOwnProperty".

https://stackoverflow.com/questions/16585209/node-js-object-object-has-no-method-hasownproperty

### DIFF
--- a/src/query-builder/transformer/PlainObjectToNewEntityTransformer.ts
+++ b/src/query-builder/transformer/PlainObjectToNewEntityTransformer.ts
@@ -28,12 +28,12 @@ export class PlainObjectToNewEntityTransformer {
 
         // copy regular column properties from the given object
         metadata.columns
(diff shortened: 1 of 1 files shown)


pleerock:
Can you please check what is current behaviour in `@next` ?

hkjeffchan:
No error in @next !!! Thanks!

hkjeffchan:
By the way, seems lots of change in @next for this file. Will this be merged in master in this 0.1 version soon or need to wait for version 0.2? Thanks!

pleerock:
Im planning to release 0.2.0 in a few weeks.

PR #1821, opened 2018-03-25
Add isEager option to the entity schema

afmelsaidy:


pleerock:
Thank you very much! I was planning to remove all "is" in entity schema, e.g. isLazy => lazy, isEager => eager. If you can do it now it would be great, if no I'll do it later.

Issue #1825, opened 2018-03-26
Collisions in new shorter field aliases


bartlomiej-korpus:
**Issue type:**

[ ] question
[x] bug report
[ ] feature request
[ ] documentation issue

**Database system/driver:**

[ ] `cordova`
[ ] `mongodb`
[ ] `mssql`
[x ] `mysql` / `mariadb`
[ ] `oracle`
[x ] `postgres`
[ ] `sqlite`
[ ] `sqljs`
[ ] `websql`

**TypeORM version:**

[ ] `latest`
[x ] `@next`
[ ] `0.x.x` (or put your version here)

**Steps to reproduce or a small repository showing the problem:**

This is a bug in new shorter aliases that are generated using first 2 letters of camelCased field name.
Two fields with similar names (2 first letters of each word being same is enough, and it is even worse in embeddeds since embedded field names aren't properly camelcased so the bug triggers even if only 2 letters of the field are equal in embedded).

I am PR'ing a test case for this.

Here is example SQL generated in my test case:

`
SELECT
  "Thing"."id"                                            AS "Thing_id",
  "Thing"."embeddedThingSomeseriouslylongfieldnamefirst"  AS "Thing_emThSo",
  "Thing"."embeddedThingSomeseriouslylongfieldnamesecond" AS "Thing_emThSo"
FROM "thing" "Thing"
WHERE ("Thing"."id" = $1)
`

We can see here that two fields have the same alias.

bartlomiej-korpus:
@AlexMesser I tried to figure this out but unfortunately I don't know enough about the codebase. Even if I disable abbreviations in SelectQueryBuilder's createAliasName, the bug just moves from one place to another and different fields are not being populated.

AlexMesser:
@bartlomiej-korpus I have no idea how we can elegantly solve the problem with maximum character limit on labels. Perhaps I just turn off this functionality for every database except Oracle, because other databases have a very large limit (63 bytes for Postgres).

AlexMesser:
#1785

bartlomiej-korpus:
@AlexMesser maybe we could do some field ordering? So my field "Thing_emThSo"  would end up being "0_Thing_emThSo"  and the other one "1_Thing_emThSo".
But that would make labels dependant on their order in select clause, and I don't know if its even possible to do now(what about joins? inner query would need to know the ordering of fields in the outer one? that doesn't make sense at all).

AlexMesser:
[...]

PR #1826, opened 2018-03-26
failing test case for issue #1825

bartlomiej-korpus:
failing test case for #1825


bartlomiej-korpus:
This is failing becuase of some other tests that fails to compile.

PR #1835, opened 2018-03-27
Fix #1833: Add only one error listener to MySQL connections

### LINKED ISSUE #1833: MySQL Driver adds multiple event listeners to a single connection
**Issue type:**

[X] bug report

**Database system/driver:**

[X] `mysql` / `mariadb`

**TypeORM version:**

[X] `latest`

[index.ts](https://gist.github.com/m1cr0man/e70c066c6b8c3009d2bed61d37e10632) (Instructions at the top)

Every time a connection is pulled from the pool, the MySQL driver [adds a listener](https://github.com/typeorm/typeorm/blame/6df6c0f2321bcbd37450658ecde4f6186f88252c/src/driver/mysql/MysqlDriver.ts#L573) to catch and log connection errors.

In the case of a single instance of MySQL (and possibly for clusters too) the connection returned by [pool.getConnection](https://github.com/typeorm/typeorm/blame/6df6c0f2321bcbd37450658ecde4f6186f88252c/src/driver/mysql/MysqlDriver.ts#L466) is always the same object, and so the error listener is added multiple times to it.

After around 9 transactions, TypeORM produces this error:
```
(node:9054) MaxListenersExceededWarning: Possible EventEmitter memory leak detected. 11 error listeners added. Use emitter.setMaxListeners() to increase limit
MaxListenersExceededWarning: Possible EventEmitter memory leak detected. 11 error listeners added. Use emitter.setMaxListeners() to increase limit
    at _addListener (events.js:224:19)
    at PoolConnection.addListener (events.js:241:10)
    at MysqlDriver.prepareDbConnection (.../node_modules/typeorm/driver/mysql/MysqlDriver.js:535:20)
    at .../node_modules/typeorm/driver/mysql/MysqlDriver.js:439:48
    at Ping.onOperationComplete [as _callback] (.../node_modules/mysql/lib/Pool.js:110:5)
    at Ping.Sequence.end (.../node_modules/mysql/lib/protocol/sequences/Sequence.js:88:24)
    at Ping.Sequence.OkPacket (.../node_modules/mysql/lib/protocol/sequences/Sequence.js:97:8)
    at Protocol._parsePacket (.../node_modules/mysql/lib/protocol/Protocol.js:279:23)
    at Parser.write (.../node_modules/mysql/lib/protocol/Parser.js:76:12)
    at Protocol.write (.../node_modules/mysql/lib/protocol/Protocol.js:39:16)
```
Increasing max listeners would not be the best solution here - it's important to ensure that only one error listener is added to each database connection.

m1cr0man:
Check if there are any listeners attached to the database connection already. If not, add one.

This solves an event listener overflow issue as described in #1833 

I haven't figured out a way to write tests for this yet because the raw connection seems to be a protected property. Working on it, but opening a PR for feedback now.

### DIFF
--- a/src/driver/mysql/MysqlDriver.ts
+++ b/src/driver/mysql/MysqlDriver.ts
@@ -570,7 +570,9 @@ export class MysqlDriver implements Driver {
           Attaching an error handler to connection errors is essential, as, otherwise, errors raised will go unhandled and
           cause the hosting app to crash.
          */
-        connection.on("error", (error: any) => logger.log("warn", `MySQL connection raised an error. ${error}`));
+        if (connection.listeners("error").length === 0) {
+            connection.on("error", (error: any) => logger.log("warn", `MySQL connection raised an error. ${error}`));
+        }
         return connection;
     }
 


pleerock:
Thank you very much. Looks like bug was introduced recently. Adding @guscastro to review. @guscastro please think if this problem can be solved more elegant way (other then checking number of errors in listener), also please check if same issue does not occur for other drivers. Thank you.

guscastro:
@pleerock I went through the mysqljs docs and their code and I'm not seeing another way of handling these. The problem here is that the pool reuses connections, to which we might already have attached error listeners. `mysqljs`'s `PoolCluster` doesn't offer an interface to the connection error handler (see https://github.com/mysqljs/mysql/blob/master/lib/Pool.js#L43 for when they create the connection).

As far as node `EventEmitter`s go, I think @m1cr0man 's solution is probably good enough. The downside of this solution is that, if there is already another handler attached that is not our handler, our handler is never attached. To work around this, you could use `removeListener` and reattach the listener.

Other drivers don't suffer from this issue as we attach the error handler on newly created `Pool`s rather than pooled connections.

pleerock:
okay, thank you @guscastro for investigation

markFromMST:
This same issue exists for postgres:  https://github.com/typeorm/typeorm/blob/master/src/driver/postgres/PostgresDriver.ts#L812

PR #1850, opened 2018-03-29
add authentication mechanism option

Rhymmor:
Just added the option to MongoConnectionOptions and the corresponding parameter to MongoClient's connect (from docs: http://mongodb.github.io/node-mongodb-native/3.0/api/MongoClient.html#.connect)

### DIFF
--- a/src/driver/mongodb/MongoConnectionOptions.ts
+++ b/src/driver/mongodb/MongoConnectionOptions.ts
@@ -245,4 +245,8 @@ export interface MongoConnectionOptions extends BaseConnectionOptions {
      */
     readonly loggerLevel?: "error"|"warn"|"info"|"debug";
 
+    /**
+     * Specifies the list of authentication mechanisms the server accepts.
+     */
+    readonly authMechanism?: string;
 }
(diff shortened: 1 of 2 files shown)


pleerock:
Thank you very much!

src/driver/mongodb/MongoConnectionOptions.ts
```diff
@@ -245,4 +245,8 @@ export interface MongoConnectionOptions extends BaseConnectionOptions {
      */
     readonly loggerLevel?: "error"|"warn"|"info"|"debug";
 
+    /**
+     * Specifies the list of authentication mechanisms the server accepts.
```
vlapo: Description is little bit strange. What about "**Sets the authentication mechanism that MongoDB will use to authenticate the connection.**". Also there is `docs/connection-options.md` file with all options.

Rhymmor: Oh, I forgot to change it after copy/paste. Sure, I fix it!

vlapo: Nice :)

PR #1859, opened 2018-03-30
Doc: Migration commands should be singular

clouless:
The docs say

```
typeorm migrations:*
```

It should be

```
typeorm migration:*
```

### DIFF
--- a/docs/migrations.md
+++ b/docs/migrations.md
@@ -78,7 +78,7 @@ Here we setup three options:
 Once you setup connection options you can create a new migration using CLI:
 
 ```
-typeorm migrations:create -n PostRefactoring
+typeorm migration:create -n PostRefactoring
 ```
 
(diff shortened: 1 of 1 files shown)


pleerock:
What version are you using? We already have this change in `next` docs

clouless:
I am using `0.2.0-alpha.40` and just did a `yarn add typeorm`

and then `./node_modules/.bin/typeorm -v`

So if you already have another PR for that you can just close this one :)

pleerock:
Looks like we accidentally pushed 0.2.0 to the `latest`, I've fixed that, thanks for the report. 
Feel free to use `0.2.0-alpha.40` since final 0.2.0 will be released in a few days.
This PR can be closed since once we release 0.2.0 we are going to merge `next` branch into `master` which already contains those changes.

clouless:
ok cool thx :) Really love the whole "design" of the framework.
I checked a lot of other ORMs but this is my favorite :) great work!

PR #1866, opened 2018-04-01
translate cn doc

huanz:
I am translating the Chinese document, but it is not finished yet.you can see it [here](http://typeorm.noonme.com/)


pleerock:
Thank you!

PR #1878, opened 2018-04-03
bugfix: allow lazy load within transaction

DR-BoneZ:
Lazy loaded relations to an entity should have access to the context of the transaction in which the entity was selected.

I.E.
```
getManager.transaction(async (manager) => {
  // create User with relation Email
  let user = manager.findOneById(User, user.id);
  let emails = await user.emails; // should fetch the email we just created, but won't because lazy load executes outside of transaction
});

### DIFF
--- a/gulpfile.ts
+++ b/gulpfile.ts
@@ -406,4 +406,3 @@ export class Gulpfile {
     }
 
 }
-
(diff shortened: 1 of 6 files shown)


DR-BoneZ:
If anyone knows this codebase a little better and could shed some light that would be great. The failing test seems completely unrelated to my change.

DR-BoneZ:
tests are passing locally... wtf??

DR-BoneZ:
Got it. I'll work on it.

DR-BoneZ:
It looks like the RawSqlResultsToEntityTransformer doesn't send a QueryRunner..

DR-BoneZ:
changing pr to next

PR #1899, opened 2018-04-09
React-Native driver

### LINKED ISSUE #1114: Support react-native-sqlite-storage?
I can't go back writing raw query anymore after using typeorm in node.js. I need to use sqlite extensively in a react native app. Has anyone used typeorm with any react native local storages? Do you have plans to support react-native-sqlite-storage? I know that it will support sql.js but the app crashes with sql.js as soon as it starts.

daniel-lang:
This PR finally adds a driver for React-Native. I "fixed" the Buffer problem by simply importing the polyfill in `PlatformTools.ts`.
I'm going to add the example and link it to the main repo as soon as an alpha with the driver is available.

Closes #1114


pleerock:
🎆 🙌  yahooo, awesome job! We must have a "example" repository just like what we have for ionic

daniel-lang:
I have the example ready, but I would like to have an official alpha containing this driver to test the example before releasing it

pleerock:
Released in alpha.44

PR #1908, opened 2018-04-10
format readme and add table of content

wy193777:
Have a table so users can follow the very long readme file.

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,23 +20,64 @@
   <br>
 </div>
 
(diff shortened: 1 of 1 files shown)


pleerock:
Thank you very very much for your hard work, however I don't want to have a table of contents, because the same document (README.md) is rendered on the homepage http://typeorm.io where I would like to avoid this list of links and start with the real action. Also it will look ugly on the site since there is a list of links on the left panel

wy193777:
Thanks for your reply, just closed the PR.

pleerock:
Sorry about that =(

PR #1915, opened 2018-04-11
Optional `prefix` cache option for the redis cache key 

agubler:
**Enhancement**

Add a new optional cache configuration option `prefix` that can be used to differentiate between identical statements in the query cache across connections that could point at different databases.

If the cache option `identifier` is also passed the `prefix` will be prepended to the identifier, otherwise it will be prepended to the serialised query.

### DIFF
--- a/src/cache/QueryResultCacheOptions.ts
+++ b/src/cache/QueryResultCacheOptions.ts
@@ -10,6 +10,12 @@ export interface QueryResultCacheOptions {
     identifier: string;
 
     /**
+     * Prefix to be applied to the cache identifer
+     *
+     */
+    prefix?: string;
+
+    /**
      * Time, when cache was created.
      */
     time?: number;
@@ -28,5 +34,4 @@ export interface QueryResultCacheOptions {
      * Query result that will be cached.
      */
     result?: any;
-
 }
--- a/src/cache/RedisQueryResultCache.ts
+++ b/src/cache/RedisQueryResultCache.ts
@@ -73,22 +73,18 @@ export class RedisQueryResultCache implements QueryResultCache {
      */
     getFromCache(options: QueryResultCacheOptions, queryRunner?: QueryRunner): Promise<QueryResultCacheOptions|undefined> {
         return new Promise((ok, fail) => {
-
-            if (options.identifier) {
-                this.client.get(options.identifier, (err: any, result: any) => {
-                    if (err) return fail(err);
-                    ok(JSON.parse(result));
-                });
-
-            } else if (options.query) {
-                this.client.get(options.query, (err: any, result: any) => {
-                    if (err) return fail(err);
-                    ok(JSON.parse(result));
-                });
-
-            } else {
+            let key = options.identifier || options.query;
+            if (!key) {
                 ok(undefined);
             }
+            if (options.prefix) {
+                key = `${options.prefix}~${key}`;
+            }
+
+            this.client.get(key, (err: any, result: any) => {
+                if (err) return fail(err);
+                ok(JSON.parse(result));
+            });
         });
     }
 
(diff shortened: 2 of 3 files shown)


src/cache/RedisQueryResultCache.ts
```diff
@@ -73,22 +73,18 @@ export class RedisQueryResultCache implements QueryResultCache {
...
-                    ok(JSON.parse(result));
-                });
-
-            } else {
+            let key = options.identifier || options.query;
+            if (!key) {
                 ok(undefined);
             }
+            if (options.prefix) {
+                key = `${options.prefix}~${key}`;
```
pleerock: what is `~` here?

src/cache/QueryResultCacheOptions.ts
```diff
@@ -9,6 +9,11 @@ export interface QueryResultCacheOptions {
      */
     identifier: string;
 
+    /**
+     * Prefix to be applied to the cache identifer.
+     */
+    prefix?: string;
```
pleerock: this option should work if database cache (instead of redis) is used as well.

src/cache/RedisQueryResultCache.ts
agubler: Is simply a delimiter that would help it be more human readable when looking directly in the cache store. I can remove it if you would like?

src/cache/QueryResultCacheOptions.ts
agubler: Ha of course, no problems.

src/cache/RedisQueryResultCache.ts
pleerock: probably yes. let them decide and put this symbol in prefix if needed

agubler: Sure no problem

src/cache/DbQueryResultCache.ts
```diff
@@ -17,7 +17,7 @@ export class DbQueryResultCache implements QueryResultCache {
     // Constructor
     // -------------------------------------------------------------------------
 
-    constructor(protected connection: Connection) {
+        constructor(protected connection: Connection) {
```
pleerock: =(

src/cache/DbQueryResultCache.ts
```diff
@@ -60,6 +59,11 @@ export class DbQueryResultCache implements QueryResultCache {
                 type: driver.normalizeType({ type: driver.mappedDataTypes.cacheIdentifier }),
                 isNullable: true
             }),
+            new TableColumn({
```
pleerock: it shouldnt be a separate column i think. Just prepend to cacheIdentifier

pleerock: how? each database will have its own cache table

agubler: If that is the case then there isn't an actual need to have a prefix for the database cache, it is a redis only requirement.

pleerock: correct, but this option is exist and it global for all caches, and if user uses a table cache and set it and it won't affect anything it will be strange, right?

agubler: @pleerock So perhaps we should rename the option to make it explicit that it is a redis only option, `redisKeyPrefix` or similar? Doesn't seem like useful to implement it for the database cache if it has zero benefit.

pleerock: no, its a bad idea. Cache can be anything, not only table or redis and its abstract from a real implementation. So probably its better to append prefix to queries and identifier I think. Those who wont use it simply will ignore it.

agubler: Okay if that's what you want for the database cache implementation I will do that. 

Even as we're leaving it as an abstract option, it does seems arbitrary to implement it in the database caching by prepending to `identifier` (which might be empty) and have zero affect on the end functionality - having it stored in a separate column for me is more explicit and less arbitrary (and more expected) but I figure thats a breaking change for existing users of the database cache (so not great).

Thanks for your time, it's great to have these discussions! I don't have time tonight but I should be able to get that done for tomorrow :)

pavelkalin:
Not very clear whether prefix was added or not, so for everyone looking like I was, this works:

`cache: {
		type: 'redis',
		options: {
			prefix: 'sql_cache::',
			host: REDIS_HOST,
			port: REDIS_PORT,
		},
		ignoreErrors: true,
	},
`

SamuelGaona:
> Not very clear whether prefix was added or not, so for everyone looking like I was, this works: `cache: { type: 'redis', options: { prefix: 'sql_cache::', host: REDIS_HOST, port: REDIS_PORT, }, ignoreErrors: true, },`

Not working with ioredis.

SamuelGaona:
> Not very clear whether prefix was added or not, so for everyone looking like I was, this works: `cache: { type: 'redis', options: { prefix: 'sql_cache::', host: REDIS_HOST, port: REDIS_PORT, }, ignoreErrors: true, },`

Nevermind, the option is called `keyPrefix`. This should be documented aswell.

Rahul1d:
Does the prefix works ? I have tried but its not working

Issue #1918, opened 2018-04-12
Remove websql and SystemJS support


pleerock:
@daniel-lang I would like to open a discussion if we really need to support a websql. websql was a feature to support typeorm in browser however it does have multiple limitations:

* only chrome-like browsers support it which means it can't be used in a real production scenarios
* lot of serious issues with it (like problems in schema sync, inconsistent behaviour, etc.)

Then we got a sqljs implementation which does not have any of above issues. So, shall we deprecate a websql and make sqljs as our first-citizen browser driver. Same I guess will work for cordova, ionic, etc., right?

What is your thoughts here?

Also btw I wanted to discuss a deprecation of systemjs support (since almost nobody use it) and make webpack-first support.

pleerock:
btw I just checked sqljs implementation and have a question. Why in

https://github.com/typeorm/typeorm/blob/6d0507233b78342db0acecbd39a5f48955af7cbe/src/driver/sqljs/SqljsDriver.ts#L85

we have a fileExist check? Shouldn't it create a database if it does not exist? I mean what if I just created app without having a db yet?

Because I wasn't be able to setup a basic app:

```js
const typeorm = require("typeorm");
const Post = require("./post.js").Post;

typeorm.createConnection({
    type: "sqljs",
    autoSave: true,
    synchronize: true,
    location: __dirname  + "/db.sqlite",
    entities: [
        Post
    ]
})
    .then(() => console.log("done"))
    .catch(err => console.error(err));
```

daniel-lang:
The only reason to keep websql has been that is nice to have when developing an ionic app in the browser. But sqljs would work just as fine (or even better considering the limitations) in this situation.
As mention in [this comment (#1081)](https://github.com/typeorm/typeorm/issues/1081#issuecomment-346166472), the current state of the driver is unusable for anything beyond a simple test.

SystemJS is a similar story. I recently tested `0.2` and noticed, that, like websql, it's only usable for simple tests because we have `require`-statements in multiple places to prevent circular dependencies. But those `require`-statements cannot be interpreted by SystemJS. So this is another indicator, that it is not used. Another thing that I'm quiet unhappy with in regards to SystemJS is that the file that tsc creates only works up to a certain version of SystemJS (newer version are not supported).

TL:DR; I'm all for dropping Websql and SystemJS support

pleerock:
> TL:DR; I'm all for dropping Websql and SystemJS support

[...]

PR #1922, opened 2018-04-12
Sql.js: Fixed startup error when file does not yet exist

daniel-lang:
problem mentioned by @pleerock in #1918


pleerock:
I also added [electron app example](https://github.com/typeorm/electron-typescript-example). You'll like it! Looks like the only platform left to support is native script...

PR #1923, opened 2018-04-12
Drop WebSQL and System.js support

### LINKED ISSUE #933: websql: after error creating table, `temporay_foo` table is not deleted
Tried to use unsupported data type (`char` instead of `character`).

Table creation first time aborted with error "data type not supported" when creating temporary_foo table.

After I fixed the problem in my code, further processing of schema sync stopped due to "table temporary_foo already exists".

I suppose schema-sync should better drop temporary table after error.

### LINKED ISSUE #1081: Transactions on WebSql
Should transactions on WebSql work? We are making a test with an Ionic app. Transactions work on Android device and SQlite database but don't work for browser and WebSQL. On WebSQL when there is an error, no rollback is done.
Example of code: 
```typescript
await maintenanceRepository.entityManager.transaction(async entityManagerTx => {
    entityManagerTx.save(currentMaintenance); // Insert OK
    entityManagerTx.save(maintenanceFormValues); // This should fail and rollback
});
```

The maintenanceFormValue.save is modified in order to generate an error. Despite of that, current maintenance is persisted and maintenances are inserted in database for WebSQL.

### LINKED ISSUE #1918: Remove websql and SystemJS support

daniel-lang:
Support for WebSql and System.js have has been dropped after a
discussion (#1918) as they are cannot be fully supported and aren't used
that much.

fixed #1918
fixed #933 (i guess)
fixed #1081 (i guess)

The last two issues are both WebSql related.


pleerock:
Thank you, also please don't forget to update our examples repositories to latest alphas. Once we release 0.2.0 we just remove alpha postfix.

PR #1927, opened 2018-04-12
Adding failing test case for GitHub issue #1926

csuich2:
Failing test case for #1926


csuich2:
@pleerock The test case file was named issue.ts not issue-1926.ts, so I fixed that, but it still looks like Travis CI isn't executing my test. Am I missing something here?

PR #1942, opened 2018-04-14
removing cursor and table types from sql server driver

Kononnable:
Removed cursor and table types from sql server driver. These types are special, there is no way to store them permanently(in a table).


pleerock:
Thank you!

sadokmtir:
Can anyone explain me what is the reason behind deleting the cursor and forcing use of `getMany()` and loading everything in memory ? or I am missing something ? Thanks

Kononnable:
We've removed types from entity definitions(aka column type in a table). You can't store cursor in a table column, it only exist in run time. This doesn't affect ability to use cursors.

sadokmtir:
Thanks you for your answer, can you point me please for the documentation about the cursor, I could not find it I can only find pagination build with `.skip(n)` and ` .take(m);`

Kononnable:
I don't think cursor based pagination is implemented in typeorm. You can take a look at https://github.com/benjamin658/typeorm-cursor-pagination

Kononnable:
I would say that if you're not taking all of records at once you're using some kind of pagination. As for cursors in general - depends on a database structure and how it is used they can be both good and bad in terms of performance.
I think it would be a good feature for typeorm to have, but I'm sure I wouldn't like it to be the default one - you might hit performance issues if you've got your business logic in your db.

It's also worth pointing out that we're living in JavaScript world, so fetching next row have to be an async function(even if row is already pre-fetched).

PR #1949, opened 2018-04-15
feat(test): Use ts-node with typechecking and timeout

pdeszynski:
@pleerock One more try with timeout flags and color output. It's important that first run might be long (more than 30s) as the cache is generated then. All subsequent calls will be fast thought. In comparison running single test with `ts-node` takes around 1-2s vs 30s with tsc compiling everything.


pleerock:
Lets try it.

PR #1994, opened 2018-04-22
Improved container types

saboya:
I'm writing a plugin to better integrate TypeORM and Hapi. While doing so, I felt the need to use the container feature. However, TypeORM currently doesn't expose a generic container implementation, making it harder than it should be.

This PR introduces `DefaultContainer`, which itself implements the new `ContainerInterface`. Also cleaned up types a bit unifying the argument to the container to the `ContainedType<T>` type.

This would allow anyone that just wants to have separate containers to simply import `Defaultcontainer` and use it with `useContainer`.

Thank you for this great library.


pleerock:
can I ask you for what purpose do you use container? (with TypeORM)

saboya:
I was using containers to manage separate connections, but I realized I was actually misusing the feature and actually could just handle `ConnectionManager` directly :) I'm just picking up TypeORM, still figuring out the best way to do things.

I think the PR is still valid anyway, but I'll let you be the judge of that :)

pleerock:
Yes this functionality can be merged, but I don't see lot of value in it TBH. I think I'll close it until someone else provide more feedback about his use cases.

saboya:
While exposing the container itself might be a needed feature, I still think the type chances make for a better / cleaner design. Would you like me to change it to just be type cleanup PR and merge that?

pleerock:
Okay

saboya:
Last rebase only has the type fixes. Let me know if there's anything you want adjusted!

PR #2017, opened 2018-04-24
added test for nested child entities

Gecko222:


Gecko222:
This wasnt a problem before 0.2.1

PR #2024, opened 2018-04-25
 Add a Brackets example on Where section in select-query-build.md

astrohsy:
I found there was no example of `Brackets` on docs. So, I added an example!


pleerock:
Thanks!

PR #2029, opened 2018-04-25
hotfix: get new query runner if old one is released for lazy relations

DR-BoneZ:


DR-BoneZ:
failing on `gulp createTravisOrmConfig`?

src/query-builder/RelationLoader.ts
```diff
@@ -22,6 +22,7 @@ export class RelationLoader {
      * Loads relation data for the given entity and its relation.
      */
     load(relation: RelationMetadata, entityOrEntities: ObjectLiteral|ObjectLiteral[], queryRunner?: QueryRunner): Promise<any[]> { // todo: check all places where it uses non array
+        if (queryRunner && queryRunner.isReleased) queryRunner = undefined; // get new one if already closed
```
pleerock: I think question is - how released query runner come here?

DR-BoneZ: Here's what happens:
- Entity is loaded within transaction, lazy load methods are generated with the transaction's query runner passed in. 
- Transaction is committed
- Entity still has lazy load methods with query runner from transaction.
- Upon lazy load, tries to use closed query runner

PR #2045, opened 2018-04-27
Update migrations.md

befitsandpiper:
your migration commands are out of date, and are all now pluralized


pleerock:
Thanks for the contribution, however migration api was changed in 0.2.0 - now it starts with `migration:`

Issue #2044, opened 2018-04-27
The @ManyToOne() relationship will be broken when you use custom ids.


sevenryze:
**Issue type:**

[ ] question
[X] bug report
[ ] feature request
[ ] documentation issue

**Database system/driver:**

[ ] `cordova`
[ ] `mongodb`
[ ] `mssql`
[X] `mysql` / `mariadb`
[ ] `oracle`
[ ] `postgres`
[ ] `sqlite`
[ ] `sqljs`
[ ] `react-native`

**TypeORM version:**

[X] `latest`
[ ] `@next`
[ ] `0.x.x` (or put your version here)

**Steps to reproduce or a small repository showing the problem:**
Ok, another issue related to #1542 , i think.

When you use the `@BeforeInsert()` to achieve the goal to set the custom Ids, even the normal save() method will be fine, the save method of relationship decorators will not set the relationed-entity's ID.

1. use `typeorm init` to init a project.
1. mysql 5.7
1. run the code snips below. 

```js
@Entity()
export class User {
  @PrimaryColumn() id: number; // Just do not use auto generated ID.

  @Column() name: string;

  @OneToMany(type => Photo, photo => photo.user)
  photos: Photo[];

  @BeforeInsert()
  insertId = () => {
    this.id = 123;
  };
}
```

```js
@Entity()
export class Photo {
  @PrimaryColumn() id: number;

  @Column() url: string;

  @ManyToOne(type => User, user => user.photos)
  user: User;

  @BeforeInsert()
  insertId = () => {
    this.id = 123 + 1000 * Math.random();
  };
}

```

```js
createConnection()
  .then(async connection => {
    const photo1 = new Photo();
    photo1.url = "me.jpg";
    await connection.manager.save(photo1);

    const photo2 = new Photo();
    photo2.url = "me-and-bears.jpg";
    await connection.manager.save(photo2);

    const user = new User();
    user.photos = [photo1, photo2];
    await connection.manager.save(user);
  })
  .catch(error => console.log(error));
```
Even though those entity are saved, the userId field of photo table is null.

If you use auto generated IDs, the result will be fine. 

Can you give me some clues about how to fix this problem? thank you. 

<!--
    To answer those questions you need to put "x" inside the square brackets, for example:
    [x] `mysql`
    [ ] `postgres`

    Also, please format your code properly (by taking code blocks into ```ts .... ```)
!>

sevenryze:
i find a way to fix this issue, see #2053 for reference.

[...]

Issue #1542, opened 2018-02-01
Creating Custom Generated Column Value


PawanKr2feb:
I am looking for a way to create some custom value for my Column in entity.

Ex:

```ts
@Entity("DemandRecord")
   export class DemandRecord {

    @CustomColumn()
     demandRecordNumber: number;
```

Ex:- demandRecordNumber values should be like YYYYMMDD001,YYYYMMDD002..like this where YYYYMMDD is year,month and date.

Am i not able to find any approach for this. In Hibernate we have ways to this but in TypeORM..i am not able to find any documents regarding this

pleerock:
Its not clear what you are requesting - please provide a detailed description of your feature request.

pleerock:
Closing as no answer from reporter.

PawanKr2feb:
Added the extra information

sevenryze:
I find that i need the same feature.

For now, when use `@PrimaryGeneratedColumn()` decorator, we get only two types of generation strategy - increment and uuid.  What i need is the ability to set a custom value to the generated column, maybe the option object could add a property - value - which could get a function to call to return the value we need. 

if you think this feature is useful, i think i could give a PR.

pleerock:
there is no need to create a new decorator, simply use `@PrimaryColumn` decorator and generate its value anywhere - in the class itself, in the listeners, simply by assigning your value to class property.

PawanKr2feb:
Can you please give a small code snippet for the above requirement..not able to get what you are trying to say

pleerock:
```ts
@Entity()
export class Post {

   @PrimaryColumn()
   id: string;

}
```

You want to have a manually generated value for id, right? Then Just Do It.

```ts
const post = new Post();
post.id = "your-value";
manager.save(post);
```

PawanKr2feb:
What we are trying to say that in hibernate, there is a special way to achieve the scenario mentioned above..since typeorm is almost similar to hibernate so why don't we can achieve that in typeorm..what u mention above will help but it will time consuming for a developer if they have to autogenreate a value which is dependent on a DB column

pleerock:
There is no need for additional functionality in there.

Alternative to that approach you can do:

```ts
@Entity()
export class Post {

   @PrimaryColumn()
   id: string;

  @BeforeInsert()
  private beforeInsert() {
         this.id = generateAnything();
  }

}
```

sevenryze:
[...]

PR #2053, opened 2018-04-30
Fix the issue that should not get embedded column value twice

sevenryze:
Related to the issue #2044 and the issue #1542.

All i need is a way to use my custom generated Id but it doesn't work since relationship id will not be saved with custom id.

I dig some and find we get the embedded value twice and the todo in code already said we should consider remove this function. 😄


pleerock:
Hope it won't break anything 😨 But looks like todo in there says us that we can remove this line.

sevenryze:
@pleerock yeh, 🔮 that's why we need the tests suit... since no one could predict what's happening to our code, which is just 🌫 Mystic Forest 🌫.

even so, i'm nerve-wracking...

again, thanks for your quick reply!

pleerock:
correct we have tests however I can count at least a million different cases where we can provide tests, because of complexity and feature-richness of orm, but obviously we can't provide test for everything.

PR #2071, opened 2018-05-03
Minor changes in Many-to-one / one-to-many relations examples

jithinkmatthew:
While trying out the examples given in 'Many-to-one / one-to-many relations' , getting an error. `error: null value in column "name" violates not-null constraint`. So added a minor change in that example to insert name in User Table.


pleerock:
Thank you!

PR #2079, opened 2018-05-03
Issue #2067 - Fixed UnhandledPromiseRejectionWarning in PostgresQueryRunner

### LINKED ISSUE #2067: Unhandled promise rejection warning on postgres connection issues
**Issue type:**

[ ] question
[X] bug report
[ ] feature request
[ ] documentation issue

**Database system/driver:**

[ ] `cordova`
[ ] `mongodb`
[ ] `mssql`
[ ] `mysql` / `mariadb`
[ ] `oracle`
[X] `postgres`
[ ] `sqlite`
[ ] `sqljs`
[ ] `react-native`

**TypeORM version:**

[X] `latest`
[X] `@next`

**Miscellaneous:**

npm: 6.0.0
node: 9.10.1
postgres: 10.3

I'm writing a small application using typeorm together with `typeorm-typedi-extensions` and a postgres database inside a docker container. Currently I'm running into an uncatchable error when the database connection breaks during runtime:

1. Start the database server.
2. Start the application and connect to the database.
3. Execute queries - everything works fine.
4. Stop the database server.
5. Try to execute a query (e.g. `SomeRepository.find()`).
6. You will now see an `UnhandledPromiseRejectionWarning` inside the terminal and your application has no idea what's going on.

```
(node:2899) UnhandledPromiseRejectionWarning: Error: connect ECONNREFUSED 127.0.0.1:5432
    at TCPConnectWrap.afterConnect [as oncomplete] (net.js:1173:14)
(node:2899) UnhandledPromiseRejectionWarning: Unhandled promise rejection. This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was nothandled with .catch(). (rejection id: 1)
(node:2899) [DEP0018] DeprecationWarning: Unhandled promise rejections are deprecated. In the future, promise rejections that are not handled will terminate the Node.js process with a non-zero exit code.
```

A small debugging session showed that the error is thrown at the `query(...)`-Method of the `PostgresQueryRunner`. The problem seems to be the async function inside the Promise call:

```ts
return new Promise<any[]>(async (ok, fail) => {
    const databaseConnection = await this.connect();
    ...
}
```

In my case the QueryRunner tries to create a new connection and execute the query, but since the database is not available an error is thrown. This error is caught inside the aync function, but since the surrounding Promise is only triggered when either the `ok` or the `fail` callback is called, the result (or in this case the error) returned by the async function is unreachable. The Promise call should be three lines below and wrapping the `databaseConnection.query`-method with the callback and should not contain an async method - instead the query method should be async. Another way is to wrap all the code inside the Promise into a try-catch block, like it's done in most other QueryRunners.

nilshohmann:
Fixes #2067 

- Wrapped the code inside the Promise into a try-catch block
- Wrote a test that failed with a timeout in the previous code and now succeeds

### DIFF
--- a/src/driver/postgres/PostgresQueryRunner.ts
+++ b/src/driver/postgres/PostgresQueryRunner.ts
@@ -152,26 +152,31 @@ export class PostgresQueryRunner extends BaseQueryRunner implements QueryRunner 
             throw new QueryRunnerAlreadyReleasedError();
 
         return new Promise<any[]>(async (ok, fail) => {
-            const databaseConnection = await this.connect();
-            this.driver.connection.logger.logQuery(query, parameters, this);
-            const queryStartTime = +new Date();
+            try {
+                const databaseConnection = await this.connect();
+                this.driver.connection.logger.logQuery(query, parameters, this);
+                const queryStartTime = +new Date();
 
-            databaseConnection.query(query, parameters, (err: any, result: any) => {
+                databaseConnection.query(query, parameters, (err: any, result: any) => {
 
-                // log slow queries if maxQueryExecution time is set
-                const maxQueryExecutionTime = this.driver.connection.options.maxQueryExecutionTime;
-                const queryEndTime = +new Date();
-                const queryExecutionTime = queryEndTime - queryStartTime;
-                if (maxQueryExecutionTime && queryExecutionTime > maxQueryExecutionTime)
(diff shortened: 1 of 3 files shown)


pleerock:
Thank you very much. Will be released in 0.2.5

PR #2082, opened 2018-05-04
Support varbinary data type for mysql driver

asvetliakov:
https://dev.mysql.com/doc/refman/5.7/en/binary-varbinary.html


pleerock:
you shall add test [here](https://github.com/typeorm/typeorm/blob/master/test/functional/database-schema/column-types/mysql/column-types-mysql.ts).

asvetliakov:
@pleerock , @AlexMesser  Done

AlexMesser:
thank you!

PR #2098, opened 2018-05-08
Update SUPPORTED-PLATFORMS.md

lucashsilva:
Added sql.js to plugins array, needed in webpack configuration for browser.


pleerock:
Thank you!

Issue #1981, opened 2018-04-20
[Bug] Boolean values not casted properly when used in .find() condition


robmarti:
**Issue type:**

[ ] question
[x]  bug report
[ ] feature request
[ ] documentation issue

**Database system/driver:**

[x] `cordova`
[ ] `mongodb`
[ ] `mssql`
[ ] `mysql` / `mariadb`
[ ] `oracle`
[ ] `postgres`
[ ] `sqlite`
[ ] `sqljs`
[ ] `react-native`

**TypeORM version:**

[x] `latest`
[ ] `@next`
[x] `0.2.0` (or put your version here)

Dear @pleerock and community,

before describing my actual problem, I want to thank you for this **awesome** project :+1:

When developing a mobile application with Ionic (Cordova, respectively), I stumbled upon the following issue. Consider the following example:

I have a `Product` model, that looks as follows:
```
@Entity()
export class Product {

    // various columns here

    @Column()
    liked: boolean;
}
```

and a respective `ProductRepository` in order to handle access to this model. This works so far, and i can create / update / delete models fine.
However, if i want to show only the `Product`s that are `liked`, I request data like this:

```
// repository is the ProductRepository!
products = await repository.find({
	liked : true
});
```

However, when executing this query, i get an empty result. Problem is, that the database used is a `SQLite` database (remember that i use Ionic!). Connection is set up properly (remember that I can save / update / delete models respectively). The `Boolean` attribute is internally represented as an `integer` in the `SQLite` database. When I use `1` instead of `true`, however, it works.
```
// repository is the ProductRepository!
products = await repository.find({
	liked : 1
});
```

How can i properly "configure" the ORM to use `boolean` values instead of `tinyint` **for the `find()` calls**?

Thank you very much for your help, really appreciate it!
Cheers

pleerock:
strange this must work. Can you contribute and create a PR with a failing test that reproduces this issue?

robmarti:
I tried but I was not able to set up the `TestingConnections`. 
I added the following to the `ormconfig.json` file
```
{
    "type": "cordova",
    "database": "test",
    "location": "default",
    "logging": false
  }
```
and tried to create a connection inside the `before` hook 

```
before(async () => connections = await createTestingConnections({
        enabledDrivers: ["cordova"],
        entities: [__dirname + "/entity/*{.js,.ts}"],
        schemaCreate: true
    }));
```
[...]

PR #2110, opened 2018-05-09
fix issue 1981

robmarti:
This PR addresses issue #1981 where boolean values could not be used to `find` entities in a repository using `cordova`. This is because boolean values are represented as numeric values.  

This PR maps the boolean values to their corresponding numerical representation before submitting a query.

### DIFF
--- a/src/driver/sqlite-abstract/AbstractSqliteDriver.ts
+++ b/src/driver/sqlite-abstract/AbstractSqliteDriver.ts
@@ -301,7 +301,15 @@ export abstract class AbstractSqliteDriver implements Driver {
      * and an array of parameter names to be passed to a query.
      */
     escapeQueryWithParameters(sql: string, parameters: ObjectLiteral, nativeParameters: ObjectLiteral): [string, any[]] {
-        const builtParameters: any[] = Object.keys(nativeParameters).map(key => nativeParameters[key]);
+        const builtParameters: any[] = Object.keys(nativeParameters).map(key => {
+            // Mapping boolean values to their numeric representation
+            if (typeof nativeParameters[key] === "boolean") {
+                return nativeParameters[key] === true ? 1 : 0;
+            }
+
+            return nativeParameters[key];
+        });
+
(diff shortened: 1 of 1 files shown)


pleerock:
thanks for the PR, you need to add test otherwise this change can be simply deleted

robmarti:
Dear @pleerock ,

thank you for fast response on my PR. Let me briefly summarize issue #1981 again, which is tackled by this PR:
- Calling `repository.find({"key" : true});` results in creating a "wrong" query parameter when using the `Cordova` driver (which internally uses the `AbstractSqlite` driver on mobile devices.
- `Cordova` uses `SQLite` as a database, but represents `boolean` values as numerical values
- Querying for `liked : true` results in an empty set, because the database contains `liked : 1` values

As @daniel-lang states, this may be because `cordova` does not internally convert `boolean` to `integer` values.

Based on your comment, you state that you need to use the `sqlite` instead of the `cordova` driver in order to test it, beceause you cannot run the test on the mobile device. Furthermore, you also commited a test addressing this issue ( https://github.com/typeorm/typeorm/commit/bc6bf763c2e2b7a9c27ed6b234fa1d75c662af29 ).

However, as you correctly stated, this test passes successfully, because it is **not** run on a mobile device (that uses the `cordova` driver)! This makes it so hard to reproduce the error and test for the correct behavior of the submitted PR.

As you can see, the PR still allows to pass your initially written test (see commit linked above). However, I also created a new version of `typeorm` that contained the submitted PR. This version, in turn, was included in my customized `ionic-example` application (I have linked in the issue #1981) so I can confirm the proper behavior of the submitted solution.

Because you already had a test to address this issue (which does not exist on regular `sqlite` driver, but only when using the `cordova` one!) I did not explicitly create a new test. Furthermore, it may not be possible to create a test that automatically checks for this incorrect behavior, because it need to run on a mobile device.

It would be very kind if you still could merge this PR or give feedback how to properly test for the new (correct) behavior.

Kind Regards

pleerock:
Okay, thank you very much

PR #2150, opened 2018-05-15
exclude read-only columns from the insert columns list

mike-bhtp:
Included in a PR for reference; I'm not sure if this is the best place for this.  Fixed issue 2041 for me (readonly columns were being included on the insert statement).

### DIFF
--- a/src/query-builder/InsertQueryBuilder.ts
+++ b/src/query-builder/InsertQueryBuilder.ts
@@ -294,7 +294,7 @@ export class InsertQueryBuilder<Entity> extends QueryBuilder<Entity> {
 
             // if user did not specified such list then return all columns except auto-increment one
             // for Oracle we return auto-increment column as well because Oracle does not support DEFAULT VALUES expression
-            if (column.isGenerated && column.generationStrategy === "increment" && !(this.connection.driver instanceof OracleDriver))
+            if (column.isReadonly || (column.isGenerated && column.generationStrategy === "increment" && !(this.connection.driver instanceof OracleDriver)))
                 return false;
 
             return true;


mike-bhtp:
Odd how that test failed...

mike-bhtp:
I probably closed this prematurely, but I hadn't run unit tests against everything, and probably wouldn't be able to for some time.  It looks like the `should not update columns marked with readonly property` test for `test/functional/columns/readonly/columns-readonly.ts` isn't correct?  I would expect that a `readonly` column would neither insert nor update, thus if you pull the Post from the repo, authorName would be null.  

There's a NOT NULL constraint on the `authorName` column (at least in Postgres), so the insert should never work, and thus the test I broke should not have been working.

pleerock:
It is by-design. `readonly` shall only block updates, not inserts.

ps2goat:
I'd say this is a bad design.  Other ORMs use the 'readonly' flag to indicate a column that _cannot_ be inserted into, such as computed columns, or columns that they don't want to be updated. How is this handled in typeorm 0.2.x?

It seems like a better approach would be to restore the functionality from `0.1.16` for `readonly`, and add new properties for what you're describing.  Something like `insertOnly` to mark it as insertable but not updatable or deletable.

I can create an issue for this, but it definitely seems like a regression.

Also, at least one test that failed is set up incorrectly, as the `post.autherName` is marked as readonly, yet the test expects the authorName to be the new value: `should not update columns marked with readonly property`, from `test/functional/columns/readonly/columns-readonly.ts`.

pleerock:
yeah you are right, not sure how come test is wrong. We can implement your proposal, add `update: boolean` and `insert: boolean` flags. Are you able to contribute?

mike-bhtp:
Yes, I can.  I don’t have a lot of free time to learn the entire code base, but if you could point me in a general direction, I could spend some time on it.

Thanks,
Mike

________________________________

vlapo:
@mike-bhtp Do you want implement this feature or could I close this PR?

ps2goat:
You can close it-- I will not be able to implement it at this time.

vlapo:
@ps2goat this PR was created from account @mike-bhtp

Kononnable:
Closing as no response from the author.

PR #2170, opened 2018-05-18
use ReadonlyArray for parameters

jbpionnier:


pleerock:
Thanks for the PR. Motivation for this change?

jbpionnier:
Typeorm does not need to modify array...
with this change I will be able to give `ReadonlyArray` to Typeorm

pleerock:
so looks like your are trying to solve your specific problem in a typeorm codebase? Sorry, but I can't accept this PR, you have to fix issue in your codebase.

jbpionnier:
Hi, 
Sorry if i wasn't very clear, it's not so much an issue in my code base, but it was rather to make the typeorm code base more secure by using immutable arrays. From what i understand you shoudn't have to push in arrays in typeORM, is that correct?

pleerock:
yes its not pushing parameters, however there are hundred other places where arrays are not being changed but we still use regular T[] syntax. So, if we do what you suggest then we have to do it everywhere. But Im not yet sure if we need to do it, since this change has such a small value.

Kononnable:
@jbpionnier Are you still interested in this PR? Nothing has changed for a few months.

jbpionnier:
No. I would not want to get in a fight

PR #2177, opened 2018-05-20
instanceof doesn't work as intended in typeorm for javascript

Alphapage:
When using javascript with typeorm, `optionsOrName instanceof Object` returns `false`. So, typeorm tries to find a configuration file which is unnecessary when options are already submitted.


pleerock:
We can merge this PR. But its not the only place where we use instanceof Object instead of typeof object. Its not about javascript actually, its about how you create a connection. I know only one way of object creation which leads to object not being instance of Object. Can you please provide your example, how do you initialize options?

Alphapage:
Not enough testing before submitting the PR. I updated my project and now it works as expected. I will close it for now.
Sorry for the inconvenience.

PR #2198, opened 2018-05-23
Working In Progress: add support for rowversion type for mssql/SQL Server

wy193777:
Documents related to rowversion type: https://docs.microsoft.com/en-us/sql/t-sql/data-types/rowversion-transact-sql?view=sql-server-2017


AlexMesser:
thank you!

wy193777:
You are welcome. I guess we also need to update some documents. I will look into that if I got time tonight.

Issue #2199, opened 2018-05-23
Inserting value for @PrimaryGeneratedColumn() ?


MarkMatute:
**Issue type:**

[x] question
[ ] bug report
[ ] feature request
[ ] documentation issue

**Database system/driver:**

[ ] `cordova`
[ ] `mongodb`
[ ] `mssql`
[x] `mysql` / `mariadb`
[ ] `oracle`
[ ] `postgres`
[ ] `sqlite`
[ ] `sqljs`
[ ] `react-native`

**TypeORM version:**

[x] `latest`
[ ] `@next`
[ ] `0.x.x` (or put your version here)

Is there a way to key in custom id for entity?
The code below does work on 0.1.17, does not work on 0.1.18 onwards.

```
@Entity()
class SampleEntity {

@PrimaryGeneratedColumn()
id: number

@Column()
name: string
}

 return entityManager.create(SampleEntity, {
      id: 100,
      name: 'ABC'
    });
```

MarkMatute:
For the record it works on 0.1.18, 0.1.19, but not on 0.2.0 onwards. 
any response for this?

Thanks

PR #2218, opened 2018-05-26
fix: enable explicitly insert primary key value into mysql database . #Issue-2199

teyou:
fix : https://github.com/typeorm/typeorm/issues/2199


pleerock:
Thank you!

teyou:
Thanks

blackherohd:
@teyou I see that it solved for mysql, sqlite. So, how about postgres. I am trying but look like it dont work with postgres

PR #2222, opened 2018-05-28
two issues:

nitzantomer:
1. when naming the primary columns in a junction table it creates two columns per one: 'userId' and 'user_id'
 2. when updating the main entity ('Record'), the ORM tries to update the junction table as well, twice, and the 2nd time it fails.
    there's no reason to do so even once.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -4,11 +4,6 @@
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
-    "8": {
-      "version": "0.0.1",
-      "resolved": "https://registry.npmjs.org/8/-/8-0.0.1.tgz",
-      "integrity": "sha1-J9BJqqwyJnucc3e5Ryvy3xW1Pig="
-    },
     "@babel/code-frame": {
       "version": "7.0.0-beta.42",
       "resolved": "https://registry.npmjs.org/@babel/code-frame/-/code-frame-7.0.0-beta.42.tgz",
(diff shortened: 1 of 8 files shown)


nitzantomer:
The ORM log for the 2nd test:
```
query: UPDATE "records" SET "status" = ? WHERE "id" = ? -- PARAMETERS: ["failed","record1"]
query: UPDATE "record_contexts" SET "recordId" = ? WHERE "recordId" = ? AND "userId" = ? -- PARAMETERS: ["record1","record1","user1"]
query: UPDATE "record_contexts" SET "recordId" = ? WHERE "recordId" = ? AND "userId" = ? -- PARAMETERS: [null,"record1","user1"]
```
The two last updates should not happen at all, and the 2nd one fails:
```
query failed: UPDATE "record_contexts" SET "recordId" = ? WHERE "recordId" = ? AND "userId" = ? -- PARAMETERS: [null,"record1","user1"]
error: { Error: SQLITE_CONSTRAINT: NOT NULL constraint failed: record_contexts.recordId errno: 19, code: 'SQLITE_CONSTRAINT' }
query: ROLLBACK
```

pleerock:
First test works properly. Join column do not care about your primary keys thats why it creates its own join columns. To make join columns use exist columns you need to specify needed column name in join column options:

```ts
    @ManyToOne(type => Record, record => record.contexts)
    @JoinColumn({ name: "record_id" })
    public readonly record: Record;

    @ManyToOne(type => User, user => user.contexts)
    @JoinColumn({ name: "user_id" })
    public readonly user: User;

```

Second test is a complicated topic. It was already fixed in `@next` (npm i typeorm@next)

nitzantomer:
@pleerock 
what's the status of the next version?  is it stable enough for production?

pleerock:
its recommended if you fix current next version if you go for production with it. Right now its quite stable but I'll start doing big changes in a near future in next.

Issue #2185, opened 2018-05-21
Transaction Isolation Level Proposal


travtarr:
**Issue type:**

[ ] question
[ ] bug report
[x] feature request
[ ] documentation issue

It was mentioned in #233 to create an issue defining how one would see isolation levels being used.

I believe it would be as simple as an `isolationLevel` parameter on the transaction methods.  Alternatively an `options` parameter could be specified with the `isolationLevel` as a property of those options.

For the *EntityManager* an overloaded function would be needed in order to specify the options before the callback:
```ts
async transaction<T>(isolationLevel: IsolationLevel, runInTransaction: (entityManger: EntityManager) => Promise<T>): Promise<T>;
async transaction<T>(runInTransaction: (entityManger: EntityManager) => Promise<T>): Promise<T> {

  ...

  try {
    if (isolationLevel) {
      await queryRunner.startTransaction(isolationLevel);
    } else {
      await queryRunner.startTransaction();
    }
    const result = await runInTransaction(queryRunner.manager);
    await queryRunner.commitTransaction();
    return result;
    ...
}
```

For *PostgresQueryRunner*:
```ts
async startTransaction(isolationLevel?: IsolationLevel): Promise<void> {
  if (this.isTransactionActive)
      throw new TransactionAlreadyStartedError();

  this.isTransactionActive = true;
  await this.query("START TRANSACTION");
  if (isolationLevel) {
    await this.query("SET TRANSACTION ISOLATION LEVEL" + isolationLevel)
  }
}
```

Where the following are defined:
```ts
enum IsolationLevel {
  READ_UNCOMMITTED = 'READ UNCOMMITTED',
  READ_COMMITED = 'READ COMMITTED',
  REPEATABLE_READ = 'REPEATABLE READ',
  SERIALIZABLE = 'SERIALIZABLE'
}
```

Then to use it, it would be as simple as:
```ts
getManager().transaction(IsolationLevel.SERIALIZABLE, (manager) => {
  ...
});
```

References:
[Postgres](https://www.postgresql.org/docs/9.6/static/sql-set-transaction.html)
[MySQL](https://dev.mysql.com/doc/refman/8.0/en/innodb-transaction-isolation-levels.html)

pleerock:
Thank you very much for your efforts! Did you check other databases support as well? We will need to implement this in every database. Also looks like an easy feature to implement, are you able to contribute?

pleerock:
[...]

PR #2227, opened 2018-05-28
feature: add ability to specify isolation level of transactions

travtarr:
In reference to #2185.

A few notes:

- There is an issue with isolation levels in *SQL Server* versions before 2014 as described [here](http://www.levibotelho.com/development/plugging-isolation-leaks-in-sql-server/) and I did not account for this in the implementation. The **TLDR** is that prior versions do not reset the isolation level when a connection returns to the pool (ie. all future transactions on a connection that explicitly set the isolation level in a previous transaction will use that isolation level and not the default)

- I didn't implement SQL Server's specific `SNAPSHOT` isolation, if that should be required (since the connection options state it is a valid isolation level) then I can see about implementing it
- For Sqlite databases, I couldn't figure out a way to detect if `shared_cache_mode` is enabled at run-time to throw an error if the method is called with the isolation level that requires `shared_cache_mode`

- The exported type `IsolationLevel` in *EntityManager.ts* is a string literal type, but if the comment in the linked issue meant that you wanted the actual values of the type in each method declaration, I can change that
    ie. 
    ```typescript
    async transaction<T>(isolationLevel: "READ UNCOMMITTED" | "READ COMMITTED" | "REPEATABLE READ" | "SERIALIZABLE", runInTransaction: (entityManger: EntityManager) => Promise<T>): 
    Promise<T>;
    ```
   instead of:
    ```typescript
    export type IsolationLevel = "READ UNCOMMITTED" | "READ COMMITTED" | "REPEATABLE READ" | 
    "SERIALIZABLE";

    async transaction<T>(isolationLevel: IsolationLevel, runInTransaction: (entityManger: EntityManager) 
    => Promise<T>): Promise<T>;
    ```

- The tests I created just test that the transactions still work with each isolation level that is possible for each database; the tests do not test the actual limitations imposed by each isolation level. If those tests are required please let me know and I'll start working on them

### DIFF
--- a/docs/transactions.md
+++ b/docs/transactions.md
@@ -1,8 +1,10 @@
 # Transactions
 
-* [Creating and using transactions](#creating-and-using-transactions)
-* [Transaction decorators](#transaction-decorators)
-* [Using `QueryRunner` to create and control state of single database connection](#using-queryrunner-to-create-and-control-state-of-single-database-connection)
+- [Transactions](#transactions)
+    - [Creating and using transactions](#creating-and-using-transactions)
+        - [Specifying Isolation Levels](#specifying-isolation-levels)
+    - [Transaction decorators](#transaction-decorators)
+    - [Using `QueryRunner` to create and control state of single database connection](#using-queryrunner-to-create-and-control-state-of-single-database-connection)
 
 ## Creating and using transactions
 
@@ -45,6 +47,30 @@ If you'll use global manager (from `getManager` or manager from connection) you'
 You also cannot use classes which use global manager or connection to execute their queries.
 All operations **MUST** be executed using the provided transactional entity manager.
 
+### Specifying Isolation Levels
+
+Specifying the isolation level for the transaction can be done by supplying it as the first paramter:
+
+```typescript
+import {getManager} from "typeorm";
+
+await getManager().transaction("SERIALIZABLE", transactionalEntityManager => {
+    
+});
+```
+
+Isolation level implementations are *not* agnostic across all databases.
+
+The following database drivers support the standard isolation levels (`READ UNCOMMITTED`, `READ COMMITTED`, `REPEATABLE READ`, `SERIALIZABLE`):
+* MySQL
+* Postgres
+* SQL Server
+
+**SQlite** defaults transactions to `SERIALIZABLE`, but if *shared cache mode* is enabled, a transaction can use the `READ UNCOMMITTED` isolation level.
+
+**Oracle** only supports the `READ COMMITTED` and `SERIALIZABLE` isolation levels.
+
+
 ## Transaction decorators
 
 There are a few decorators which can help you organize your transactions - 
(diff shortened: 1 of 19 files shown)


docs/transactions.md
```diff
@@ -1,8 +1,10 @@
 # Transactions
 
-* [Creating and using transactions](#creating-and-using-transactions)
-* [Transaction decorators](#transaction-decorators)
-* [Using `QueryRunner` to create and control state of single database connection](#using-queryrunner-to-create-and-control-state-of-single-database-connection)
+- [Transactions](#transactions)
```
pleerock: I think we don't need this top-level category since there is a single item only?

src/decorator/transaction/Transaction.ts
```diff
@@ -1,4 +1,9 @@
-import {getConnection, getMetadataArgsStorage, MongoRepository, Repository, TreeRepository} from "../../";
+import {getConnection, getMetadataArgsStorage, MongoRepository, Repository, TreeRepository, IsolationLevel, EntityManager} from "../../";
+
+export interface TransactionOptions {
```
pleerock: can you please extract it into separate file `TransactionOptions.ts`?

pleerock: into decorator/options directory

pleerock:
Looks very good. I think everything you pointed in the comments is not required for now. We can do it once we get people issues or you can do it anytime you want.

travtarr:
Made the changes you suggested except for one difference: I placed the `IsolationLevel.ts` into the `driver/types/` folder instead of the `driver/` folder. 
If you prefer it the way you suggested, let me know and I can move it up a directory.

pleerock:
Thank you very much! Would be great if you update a changelog for the next version as well.

src/driver/mysql/MysqlQueryRunner.ts
```diff
@@ -99,12 +100,15 @@ export class MysqlQueryRunner extends BaseQueryRunner implements QueryRunner {
...
      */
-    async startTransaction(): Promise<void> {
+    async startTransaction(isolationLevel?: IsolationLevel): Promise<void> {
         if (this.isTransactionActive)
             throw new TransactionAlreadyStartedError();
 
         this.isTransactionActive = true;
         await this.query("START TRANSACTION");
+        if (isolationLevel) {
+            await this.query("SET SESSION TRANSACTION ISOLATION LEVEL " + isolationLevel);
```
victordidenko: Wouldn't it be better to use `"SET TRANSACTION ISOLATION LEVEL " + isolationLevel` (without SESSION) _before_ `"START TRANSACTION"`, so isolation level will only apply to one transaction, not all subsequent transactions within the current session?

travtarr: You are correct, this was an oversight on my part.  I'll make the change and push up another PR.

travtarr: Actually, I'll have to change the logic so that the `SET` happens before the transaction starts.

PR #2240, opened 2018-05-30
Expo driver

### LINKED ISSUE #2183: Is it possible to create a driver for Expo's SQLite API?
**Issue type:**

[x] question
[ ] bug report
[x] feature request
[ ] documentation issue

**Database system/driver:**

[ ] `cordova`
[ ] `mongodb`
[ ] `mssql`
[ ] `mysql` / `mariadb`
[ ] `oracle`
[ ] `postgres`
[ ] `sqlite`
[ ] `sqljs`
[ ] `react-native`

**TypeORM version:**

[x] `latest`
[ ] `@next`

[Expo](https://expo.io) has a [SQLite API](https://docs.expo.io/versions/latest/sdk/sqlite.html#resultset), but it is extremely rudimentary. I wanted to get some opinions on whether it's even possible to create a TypeORM driver for it.

  * The `openDatabase` method accepts a `name` parameter and Expo's internal API will figure out the database location depending on the platform. This returns a `Database` object.
  * The `Database` object has a single `transaction` method with 3 arguments: a `callback` function that accepts a `Transaction` object as its only argument, and `error` function, and a `success` function.
  * The `Transaction` object has a single method with the following signature: `executeSql(sqlStatement, arguments, success_callback, error_callback)`

Comparing what's available in Expo's SQLite API to how TypeORM's `Driver` works, there doesn't seem to be a good way to do connection management, i.e. I can't 'open' or 'close' a connection to the database. Would it still be possible to create a driver for this or is it a lost cause?

cuibonobo:
Add a driver to support the [Expo](https://expo.io/) platform. In particular it is using the [Expo SQLite API](https://docs.expo.io/versions/latest/sdk/sqlite.html) that ships with Expo.

A few things to note about this pull request:
- Expo apps still need to install `react-native-sqlite-storage` as a dependency. The issue is that the Expo packager still attempts to include this dependency (even though it doesn't need it) because of the `this.sqlite = require("react-native-sqlite-storage");` line in `src/driver/react-native/ReactNativeDriver.ts`. If someone knows how to make the Expo packager ignore this line, I'm all ears! Still, needing to install  `react-native-sqlite-storage` isn't such a big deal. No linking or ejecting of the Expo app is required.
- Expo has an odd way of handling transactions: basically every query to the database is in a transaction context. In fact, running a `BEGIN TRANSACTION` query throws a `Transaction already started` error. As a result, I needed to override the `startTransaction`, `commitTransaction`, and `rollbackTransaction` methods on the `AbstractSqliteQueryRunner` class to account for the weird behavior. Besides this, the Expo driver is almost identical to how React Native is being handled.

I have a working Expo example at [https://github.com/cuibonobo/expo-example](https://github.com/cuibonobo/expo-example). To try it, download the Expo app from the App Store or Google Play Store, then `cd` into the `expo-example` directory and run:

```bash
npm install
npm start
```

If your phone is on the same network as your development machine, scanning the generated QR code will automatically open the Expo app and run the example.

---
Fixes #2183.

### DIFF
--- a/src/driver/expo/ExpoQueryRunner.ts
+++ b/src/driver/expo/ExpoQueryRunner.ts
@@ -52,7 +52,15 @@ export class ExpoQueryRunner extends AbstractSqliteQueryRunner {
     }
 
     /**
-     * Starts transaction.
+     * Starts transaction. Within Expo, all database operations happen in a
+     * transaction context, so issuing a `BEGIN TRANSACTION` command is
+     * redundant and will result in the following error:
+     * 
+     * `Error: Error code 1: cannot start a transaction within a transaction`
+     * 
+     * Instead, we keep track of a `Transaction` object in `this.transaction`
+     * and continue using the same object until we wish to commit the
+     * transaction.
      */
     async startTransaction(): Promise<void> {
         if (this.isTransactionActive && typeof this.transaction !== "undefined")
@@ -64,6 +72,10 @@ export class ExpoQueryRunner extends AbstractSqliteQueryRunner {
     /**
      * Commits transaction.
      * Error will be thrown if transaction was not started.
+     * Since Expo will automatically commit the transaction once all the
+     * callbacks of the transaction object have been completed, "committing" a
+     * transaction in this driver's context means that we delete the transaction
+     * object and set the stage for the next transaction.
      */
     async commitTransaction(): Promise<void> {
         if (!this.isTransactionActive && typeof this.transaction === "undefined")
@@ -76,6 +88,9 @@ export class ExpoQueryRunner extends AbstractSqliteQueryRunner {
     /**
      * Rollbacks transaction.
      * Error will be thrown if transaction was not started.
+     * This method's functionality is identical to `commitTransaction()` because
+     * the transaction lifecycle is handled within the Expo transaction object.
+     * Issuing separate statements for `COMMIT` or `ROLLBACK` aren't necessary.
      */
     async rollbackTransaction(): Promise<void> {
         if (!this.isTransactionActive && typeof this.transaction === "undefined")
@@ -96,6 +111,7 @@ export class ExpoQueryRunner extends AbstractSqliteQueryRunner {
             const databaseConnection = await this.connect();
             this.driver.connection.logger.logQuery(query, parameters, this);
             const queryStartTime = +new Date();
+            // All Expo SQL queries are executed in a transaction context
             databaseConnection.transaction((transaction: ITransaction) => {
                 if (typeof this.transaction === "undefined") {
                     this.startTransaction();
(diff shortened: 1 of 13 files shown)


pleerock:
Thank you very much!

Looks good I think, but I need @daniel-lang confirmation on this PR since he is expert in sqlite databases and mobiles in TypeORM.

Also, once PR will be merged we'll need to move your expo-example repository into typeorm organization on github. 

Also, you'll have to maintain this driver all your life (😆 ) because we bring a new functionality and we can't one day remove it if maintainer has gone, I hope you understand the responsibility.

cuibonobo:
Thanks for looking! I will be happy to tranfer the example repo and maintain the Expo driver. I have also been keeping an eye on changes to the SQLite API within Expo so I can make changes to the driver if necessary.

src/driver/expo/ExpoQueryRunner.ts
```diff
@@ -0,0 +1,136 @@
...
+        super();
+        this.driver = driver;
+        this.connection = driver.connection;
+        this.broadcaster = new Broadcaster(this);
+    }
+
+    /**
+     * Starts transaction.
+     */
+    async startTransaction(): Promise<void> {
```
daniel-lang: Could you write a comment documenting why you had to override the transaction methods?

cuibonobo: Yes, that's a good idea. I'll update my PR today.

pleerock:
yeah TBH I definitely would like to avoid adding specific sqlite3 drivers into typeorm. @cuibonobo what are exact expo specifics that required you to write a separate driver for it?

cuibonobo:
@stevefan1999 The API that is available in `Expo.SQLite` *is* a wrapper around `react-native-sqlite-storage`, but Expo projects are designed in such a way that you need to use Expo APIs if you want to access any native platform functionality (touching the file system, accessing sensors, etc). Expo abstracts these functions so that there are no `android` and `ios` projects in your repository. As a result, there is no way to link the `react-native-sqlite-storage` libraries to these platforms.

In fact, when you attempt to use the `react-native` driver with an Expo project, the error you get is: `undefined is not an object (evaluating 'NativeModules["SQLite"][method]')`. This is the same error you would get if you attempt to create a connection with `react-native-sqlite-storage` but haven't properly linked your libraries.

cuibonobo:
@daniel-lang I've updated the comments for those transaction lifecycle methods to give a little more insight into what is happening. Thanks for your suggestion!

I rebased my branch against current master to make sure everything still works ok.

daniel-lang:
The problem with all the different sqlite drivers that we have in TypeORM is that the way they are loaded and called is always a little different. Therefore if we want to support the different mobile platforms we will have to have a lot of sqlite drivers that are quite similar.

pleerock:
@cuibonobo I'll merge it, however in the future we might change the way we do drivers right now. Please track for #2479. Thank you very much for your contribution!

Can you please "send" your example repository to "typeorm" organization?

cuibonobo:
@pleerock Err, I spoke too soon. I attempted to transfer the repository to the "typeorm" organization but received the following error:
```
You don’t have the permission to create repositories on typeorm
```

cuibonobo:
I've added you as a collaborator on the example repo. Hopefully after accepting you can transfer it to the typeorm org.

pleerock:
its not owner access, I can't do it. Let me just add you to typeorm organization and you can move it on your own.

PR #2252, opened 2018-06-01
Fix saving of duplicate entities

### LINKED ISSUE #2251: Unexpected behavior when passing duplicate entities to repository.save()
**Issue type:**

[ ] question
[x] bug report
[ ] feature request
[ ] documentation issue

**Database system/driver:**

[ ] `cordova`
[ ] `mongodb`
[ ] `mssql`
[x] `mysql` / `mariadb`
[ ] `oracle`
[ ] `postgres`
[ ] `sqlite`
[ ] `sqljs`
[ ] `react-native`

**TypeORM version:**

[x] `latest`
[ ] `@next`

I noticed this problem while upgrading from typeorm 0.1.11 to 0.2.6. For better or for worse (okay, for worse), my team has a UI that sometimes passes duplicate entities to the server to be saved. For example, a payload might look like:
```
[ { id: 1, description: "test1a" },
  { id: 2, description: "test2a" },
  { id: 1, description: "test1a" },
  { id: 2, description: "test2a" }]
```
Under typeorm 0.1.11 the behavior here was be that four update statements are generated. After upgrading to 0.2.6, we observed that we get two update and two insert statements (which will likely fail).

It might be good to throw some kind of error or warning because it's clearly not my intent to update the same row twice, however the behavior under 0.1.11 was working for my app while it's now broken under 0.2.6.

In the interest of preserving backward compatibility I will quickly submit a PR that allows the above save to work, generating four update statements (two of them useless). If the maintainers of this library wish to deprecate this behavior and instead throw warnings or errors, I'm happy to contribute there as well.

elthrasher:
Fixes #2251

### DIFF
--- a/src/persistence/SubjectDatabaseEntityLoader.ts
+++ b/src/persistence/SubjectDatabaseEntityLoader.ts
@@ -91,16 +91,15 @@ export class SubjectDatabaseEntityLoader {
                 .findByIds(allIds, findOptions);
 
             // now when we have entities we need to find subject of each entity
-            // and insert that entity into database entity of the found subject
+            // and insert that entity into database entity of the found subjects
             entities.forEach(entity => {
-                const subject = this.findByPersistEntityLike(subjectGroup.target, entity);
-                if (subject) {
-                    subject.databaseEntity = entity;
-                    if (!subject.identifier)
-                        subject.identifier = subject.metadata.hasAllPrimaryKeys(entity) ? subject.metadata.getEntityIdMap(entity) : undefined;
-                }
+                const subjects = this.findByPersistEntityLike(subjectGroup.target, entity);
+                subjects.forEach(subject => {
+                  subject.databaseEntity = entity;
+                  if (!subject.identifier)
+                      subject.identifier = subject.metadata.hasAllPrimaryKeys(entity) ? subject.metadata.getEntityIdMap(entity) : undefined;
+                });
             });
-
         });
 
         await Promise.all(promises);
@@ -111,11 +110,13 @@ export class SubjectDatabaseEntityLoader {
     // ---------------------------------------------------------------------
 
     /**
-     * Finds subject where entity like given subject's entity.
+     * Finds subjects where entity like given subject's entity.
      * Comparision made by entity id.
+     * Multiple subjects may be returned if duplicates are present in the subject array.
+     * This will likely result in the same row being updated multiple times during a transaction.
      */
-    protected findByPersistEntityLike(entityTarget: Function|string, entity: ObjectLiteral): Subject|undefined {
-        return this.subjects.find(subject => {
+    protected findByPersistEntityLike(entityTarget: Function|string, entity: ObjectLiteral): Subject[] {
+        return this.subjects.filter(subject => {
             if (!subject.entity)
                 return false;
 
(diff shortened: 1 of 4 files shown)


elthrasher:
@pleerock Sure. Will take a look and write a test + add to this PR if necessary.

pleerock:
Thank you!

elthrasher:
@pleerock Added a test for cascades and looks like no additional changes are needed as the function you referenced is specifically de-duping so there will only be one update per entity. This is different behavior than the collection save, but it's non-breaking, so maybe good enough for now and circle back?

pleerock:
Im not sure but I think this should be changed:

```ts

                const alreadyExistRelationEntitySubject = this.findByPersistEntityLike(relationEntityMetadata.target, relationEntity);
                if (alreadyExistRelationEntitySubject) {
                    if (alreadyExistRelationEntitySubject.canBeInserted === false) // if its not marked for insertion yet
                        alreadyExistRelationEntitySubject.canBeInserted = relation.isCascadeInsert === true;
                    if (alreadyExistRelationEntitySubject.canBeUpdated === false) // if its not marked for update yet
                        alreadyExistRelationEntitySubject.canBeUpdated = relation.isCascadeUpdate === true;
                    return;
                }
```

when there will be multiple subjects it will update canBeInserted and canBeUpdated only on first one. In some situations it may lead to bugs?

elthrasher:
It looks to me like that's a guard against deeply nested entities which may have differing cascading strategies. Let me just say I'm very glad I don't have that problem. ;)

Anyway, since the cascade will be the same for each entity in a collection, I don't see this coming up. The same logic is just a bit lower in the function:
```
                const relationEntitySubject = new Subject({
                    metadata: relationEntityMetadata,
                    parentSubject: subject,
                    entity: relationEntity,
                    canBeInserted: relation.isCascadeInsert === true,
                    canBeUpdated: relation.isCascadeUpdate === true
                });
```
Or did I miss something?

pleerock:
it pushes `relationEntitySubject` into `allSubjects`:

                this.allSubjects.push(relationEntitySubject);

which comes from the outside, so allSubjects still can have multiple entities of the same time just before any push of relationEntitySubject is happening.

elthrasher:
I'll give it another look.

elthrasher:
@pleerock I'm still not seeing anything there that concerns me. Even though both of these classes have methods called findByPersistEntityLike, they are pretty different and I think just one needs remediation in this PR.

SubjectDatabaseEntityLoader starts by executing a SELECT statement with all the ids passed in and then loops over it. This is where we have a problem. If my id list is `[1, 2, 2, 3]` I wind up with a SELECT statement where that becomes the WHERE IN clause and I get back three rows. We loop over those three rows and flag three of the four entities for updates, then the last one winds up being an insert.

CascasdesSubjectBuilder doesn't execute any sql to get a list. Instead it pulls all the child entities from the entity being saved and loops on them. Here findByPersistEntityLike is de-duping the list of entities. If a dupe is found, we may update the insert/update metadata, but we don't add to allSubjects due to the return statement on line 48. So the list is deduped and we shouldn't wind up with unwanted inserts.

pleerock:
okay, let's merge it.

Thank you.

PR #2261, opened 2018-06-03
Oracle tests working on travis #452

Kononnable:
Sorry for the delay, I was busy lately.

As I mentioned earlier - oracle test will be omitted on non-colabolator builds(travis encrypted variables). They're also very slow on travis(8 vs 24 min).

To make it work(if slower ci isn't a problem) you would need to add two travis secure environment variables(DOCKER_USERNAME,DOCKER_PASSWORD) and accept oracle licenses(https://store.docker.com/images/oracle-instant-client https://store.docker.com/images/oracle-database-enterprise-edition).

Example travis logs on
Colabolator build: https://travis-ci.org/Kononnable/typeorm/jobs/387403215
Non-Colabolator build: https://travis-ci.org/Kononnable/typeorm/jobs/387403216

Builds on node 4, 6 fails(at least in my repo) but it doesn't seem to be connected to this PR - it failed on forked repo without additional commits too(on clean master branch).


pleerock:
Sorry but can you please explain what does it mean "collaborator" and non-collaborator builds? What are differences? 

> To make it work(if slower ci isn't a problem) you would need to add two travis secure environment variables(DOCKER_USERNAME,DOCKER_PASSWORD) and accept oracle licenses(https://store.docker.com/images/oracle-instant-client https://store.docker.com/images/oracle-database-enterprise-edition).

what shall I do to make it to work locally?

Kononnable:
Travis doesn't inject secure environment variables on builds made from forks, PRs made from forks(non-collaborator build). So there is no way to login to docker(except for storing docker password as plain text).
If commit which started build was made directly to typeorm repo(collaborator build), Travis consider it as trustworthy and will inject secure environment variables - CI can download and test oracle db.

To work locally with oracle db on docker:
- Accept oracle licenses(links in first comment)
- Login to docker in console  `docker login`
- Use new docker-compose, ormconfig files
- install oracledb without saving it as dependency (https://github.com/Kononnable/typeorm-model-generator/issues/70) `npm i oracledb --no-save`
- Download and configure(add to environment variables) Oracle Instant Client(depends on operating system)
- Create oracle user:
```sql
CREATE USER typeorm IDENTIFIED BY Passw0rd;
GRANT CONNECT TO typeorm;
GRANT UNLIMITED TABLESPACE TO typeorm;
```

To make it work on Travis:
- Accept oracle licenses(links in first comment)
- Add secure environment variables `DOCKER_USERNAME`,`DOCKER_PASSWORD`

pleerock:
Local workflow looks a bit complicated. We setup this [docker oracle](https://hub.docker.com/r/sath89/oracle-12c/) on @AlexMesser windows machine and it worked fine out of the box. What do you think about it?

Kononnable:
Local setup would look like:

- ~Accept oracle licenses(links in first comment)~
- ~Login to docker in console docker login~
- Use new docker-compose, ormconfig files
- install oracledb without saving it as dependency (Kononnable/typeorm-model-generator#70) `npm i oracledb --no-save`
- Download and configure(add to environment variables) Oracle Instant Client(depends on operating system)
- Create oracle user

I'm not sure about last step, maybe there is already created non-system account which isn't locked by default, but I didn't find one. Oracle instant client is still a dependency of oracledb package so it have to be installed manually. And you still have to install oracledb manually because oracle provide binaries only for most popular architectures.

On travis docker login still would be necessary - If I remember correctly oracle instant client license forbids redistributing, so you need to download it after accepting oracle license(from oracle or docker).

pleerock:
yes you are right. We only remove first two steps but its probably better then nothing? What about creating user, not sure as well maybe we'll need to execute a command inside container to create it? or create a custom Dockerfile with command inside? Do you think we shall go this way or better to use what you already setup?

Kononnable:
Yes, fewer steps is better approach. 
I've changed docker image to `sath89/oracle-12c` it's a bit slower to download and start but on developer machine it shouldn't be a problem. On travis it takes ~4 more minutes to run tests.

Setting a custom docker image would be a nice solution - it would omit step for creating oracle user, but it's another piece of software which should be maintained. I'm not sure if it is worth the trouble. Creating oracle user, installing oracledb could be added as some sort of script(bash or gulp task) so end user would have less steps to setup environment.

In the mean time oracle published new version of driver and changed node version compatibility. Nodejs 10, 8, 6 can run on node-oracledb v2.3.0 and Nodejs 9, 4 on node-oracledb v2.2.0. I need to add such distinction for this PR too. Also I'm not satisfied  with current method of checking if oracle is fully functional. sath89 image acts a bit different then original image and do some more work before service is available from outside of the container - I've added a sleep command after database is available inside a container as a quick-fix but there should be some more elegant solution.

pleerock:
What are a new workflow for both development and travis? Can you please document how to enable oracle tests on local machines somewhere?

Also I still did not completely understand why do we need `docker-compose-without-login.yml` and `ormconfig.travis.no-login.json`. Is there any way to avoid those files?

Kononnable:
Closing. I would implement it a bit differently now.
I will get back to it when(if) we will have more time. Right now other issues got more priority.

PR #2273, opened 2018-06-04
fix rowversion type for sql server

wy193777:


src/driver/mysql/MysqlDriver.ts
```diff
@@ -107,7 +107,6 @@ export class MysqlDriver implements Driver {
         "longtext",
         "enum",
         "json",
-        "rowversion",
```
pleerock: howcome =(

PR #2281, opened 2018-06-05
Add missing methods increment and decrement

droplet92:
add the methods increment and decrement to entity-manager-api.md and repository-api.md


pleerock:
Thank you very much!

PR #2295, opened 2018-06-06
Fix wrong ES version in docs for async/await

abenhamdine:
Async/await have been standardized in ES2017 (aka ES8) not ES7


pleerock:
Thank you!

PR #2323, opened 2018-06-11
refactor: update changelog with isolation level feature

travtarr:
I realized my change to the `transactions.md` file were not actually committed in the original PR, this was due to a plugin I was using that auto-creates and formats the table of contents in markdown files, so I went ahead and added that change to this PR in addition to the changelog update.


pleerock:
Thank you very much!

PR #2327, opened 2018-06-12
Update migrations.md

dimiork:
Update of method name: typeorm migration:[cmd] --> typeorm migration**s**:[cmd]


pleerock:
hey thanks for the contribution. What version are you using? In latest 0.2.x version we renamed all commands to `migration` instead of `migrations`

havenchyk:
@pleerock can be closed

Issue #2331, opened 2018-06-13
Inconsistent handling of "undefined" between save() and update()


hallsbyra:
**Issue type:**

[ ] question
[x] bug report
[ ] feature request
[ ] documentation issue

**Database system/driver:**

[ ] `cordova`
[ ] `mongodb`
[ ] `mssql`
[ ] `mysql` / `mariadb`
[ ] `oracle`
[x] `postgres`
[ ] `sqlite`
[ ] `sqljs`
[ ] `react-native`

**TypeORM version:**

[x] `latest`
[ ] `@next`
[ ] `0.x.x` (or put your version here)

If I `update` a nullable column with `undefined`, the value is changed to `null`.

This passes:
```
        // Given
        await testFixture.companyRepo.save(dummy.company({id: "C0", logo: Buffer.from("test", "base64")}))

        // When
        await testFixture.companyRepo.update("C0", {logo: undefined})

        // Then
        const c1 = await testFixture.companyRepo.findOneOrFail("C0")
        expect(c1.logo).to.be.null
```

If I instead `save` a nullable column with `undefined`, the value is not changed.

This also passes:

```
        // Given
        await testFixture.companyRepo.save(dummy.company({id: "C0", logo: Buffer.from("test", "base64")}))

        // When
        const c1 = await testFixture.companyRepo.findOneOrFail("C0")
        c1.logo = undefined
        await testFixture.companyRepo.save(c1)

        // Then
        const c2 = await testFixture.companyRepo.findOneOrFail("C0")
        expect(c2.logo).to.not.be.null
```

I think the behavior of `save` is more semantically correct.

pleerock:
You are right, `update` shall ignore the undefined values as well. Feel free to contribute BTW.

gao-sun:
Is there any update on this?

filipsuk:
seems to be still an issue

jamie3:
2 years old? Seems like a good bug to fix.

albertcito:
3 years old? Seems like a good bug to fix.

imnotjames:
Feel free to contribute.

PR #2340, opened 2018-06-14
Updating values to undefined leaves them untouched.

hallsbyra:
Fix for #2331

### DIFF
--- a/src/query-builder/UpdateQueryBuilder.ts
+++ b/src/query-builder/UpdateQueryBuilder.ts
@@ -55,6 +55,11 @@ export class UpdateQueryBuilder<Entity> extends QueryBuilder<Entity> implements 
 
         try {
 
+            // do nothing if there are no values to update
+            if (Object.keys(this.getValueSet()).length === 0) {
+                return new UpdateResult();
+            }
+
             // start transaction if it was enabled
             if (this.expressionMap.useTransaction === true && queryRunner.isTransactionActive === false) {
                 await queryRunner.startTransaction();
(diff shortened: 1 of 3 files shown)


src/query-builder/UpdateQueryBuilder.ts
```diff
@@ -129,7 +134,10 @@ export class UpdateQueryBuilder<Entity> extends QueryBuilder<Entity> implements
      * Values needs to be updated.
      */
     set(values: ObjectLiteral): this {
-        this.expressionMap.valuesSet = values;
+        const copiedValues = {...values};
```
pleerock: I won't do it this way. `values` can be an entity and it used as an entity down the road so we'll have the issues. Also if its an entity it may have properties other then column values so we need to check only column values.

koffmoff:
Oh seems this broke some existing test cases. I'll fix them if you think this is a viable solution.

koffmoff:
Took a look at the first failing test case: `embedded-many-to-one-case1.ts\inverse side`. I thought it would be just matter of updating the test case now that `undefined !== null`. However, it seems a deeper problem has been uncovered. When a `likedPost `is removed from a `User`, `Counter.likedUserId` is supposed to be set to `null`. However, it is being set to `undefined`. Looks like a bug that has not been visible before, since `undefined `had the same behavior as `null`.

I'm afraid I don't have the time to dig any deeper into this.

imnotjames:
Bummer that you don't have time to dig further.  I'm gonna close this pull request for now but if you do ever have time again please feel free to open another with the changes.

PR #2341, opened 2018-06-14
feature: add ability to get all registered connections from manager

posgarou:
This PR adds the ability to get all the connections currently registered with a manager.

Use case: In our application, we dynamically establish 0 or more connections.  This change would allow us to simplify shutdown code to something like:

```typescript
const connections = getAllConnections();

for (const connection of connections) {
  if (connection.isConnected) {
    await connection.close();
  }
}
```

### DIFF
--- a/src/connection/ConnectionManager.ts
+++ b/src/connection/ConnectionManager.ts
@@ -16,7 +16,7 @@ export class ConnectionManager {
     /**
      * List of connections registered in this connection manager.
      */
-    protected readonly connections: Connection[] = [];
+    public readonly connections: Connection[] = [];
 
     // -------------------------------------------------------------------------
     // Public Methods


pleerock:
You can simply expose `connections` then. No need to create a separate method that returns them.

posgarou:
The main reason I implemented it this way was to prevent external modifications to the `connections` array.  If that's not a concern, I can change this PR to make `connections` a public property.  Would you prefer that route?

pleerock:
> The main reason I implemented it this way was to prevent external modifications to the connections array. 

yeah I understood that. I don't think its necessary currently. I don't like connection manager design tbh and maybe in the future I'll change its design. For now simply expose connections property.

posgarou:
Sounds good.  I made the change and rebased.  👍

pleerock:
Thank you. Btw you might be interested by #2358

Issue #2259, opened 2018-06-03
Incorrect SQL query


WeekendMan:
**Issue type:**

[ x ] question
[ x ] bug report
[ ] feature request
[ ] documentation issue

**Database system/driver:**

[ ] `cordova`
[ ] `mongodb`
[ ] `mssql`
[ ] `mysql` / `mariadb`
[ ] `oracle`
[ x ] `postgres`
[ ] `sqlite`
[ ] `sqljs`
[ ] `react-native`

**TypeORM version:**

[ x ] `latest`
[ ] `@next`
[ ] `0.x.x` (or put your version here)

I have app based on node.js/express/typescript. When I add the following entity in the array of the typeorm entities app crashes.
I tried to use the @PrimaryGeneratedColumn decorator without the params, like in the docs, but it doesn't work neither.
The reason of crash - incorrect syntax of the SQL query (PostgreSQL) that generated by typeorm. The query is
`CREATE TABLE "social_account" ("id" NOT NULL, "first_name" character varying(255) NOT NULL, "last_name" character varying(255) NOT NULL, "image" character varying(255) NOT NULL, CONSTRAINT "PK_a69e41dd6fa725d17b040eda5e6" PRIMARY KEY ("id"))`
As you can see, there is no type after the `id` field.
Where is the problem?

```
@Entity()

export class SocialAccount {
  @PrimaryGeneratedColumn('increment', { type: 'int' })
  public id: number;

  @Column({ length: 255, name: 'first_name' })
  public firstName: string;

  @Column({ length: 255, name: 'last_name' })
  public lastName: string;

  @Column({ length: 255 })
  public image: string;

  @JoinColumn({ name: 'id' })
  @ManyToOne(() => SocialNetwork, (socialNetwork): number => socialNetwork.id)
  public socialNetworkID: number;

  @JoinColumn({ name: 'id' })
  @ManyToOne(() => User, (user): string => user.id)
  public userID: string;
}
```

agubler:
Hi @pleerock I created a failing test for this issue (that I've also just run into) here: https://github.com/typeorm/typeorm/compare/master...agubler:postgres-incorrect-create-table-syntax

Is there anything else you need?

pleerock:
Please create it in a PR against master branch.

agubler:
Done @pleerock

Kononnable:
Looks like issue was resolved(by fixing test).

PR #2362, opened 2018-06-19
Failing test for incorrect create table syntax

agubler:
Demonstrates creation of incorrect create table syntax in postgres when using generated columns.

Refs: #2259


agubler:
Okay having a look at the source code I think this is because when using a `uuid` generation strategy the `type` needs to be `uuid` also...

pleerock:
yes, correct, because column type is actually uuid, not a string.

PR #2365, opened 2018-06-19
Updated to accept Oracle connectString for TNS based Host resolution

tejzpr:
Updated driver to accept Oracle connect string as specified in https://github.com/oracle/node-oracledb/blob/master/doc/api.md#-813-embedded-connection-strings since Oracle connections can also be made with TNS strings when connecting to a load-balanced service.


pleerock:
Thank you!

PR #2387, opened 2018-06-22
fix for issue when using order by with the fields database name when …

deanmowat:
When using take or skip and attempting to run an order by you get a databaseName not found error if the orderby string contains the databaseName of the field and the databaseName and the propertry name in the model do not match.

This seems to be inconsistent with limit and offset which will build the query with the database name or the property name in the model.

I've added a test at test/other-issues/take-order-by-property-name-different-to-database-name

Which is passing with the change I've added. if you remove the two :

`if (!column) {
    column = alias.metadata.findColumnWithDatabaseName(propertyPath);
}`

 then you will see this test fails with the error mentioned above.

The fix could be cleaned up however I wanted to get an opinion on whether or not this is the right way to go to fix this problem.


pleerock:
not sure about this PR... automatic escaping is such a sensitive topic.... Right now by design you need to use property names in your query to automatic escape properties. Why don't you use property names?

PR #2392, opened 2018-06-23
[Feature Request] Add new connection option to sqlite for SQLCipher #…

JeongJun-Lee:
Hello Umed,

I tried to make a test case but basically it's not impossible to test a connection if a target sqlite database is not encrypted by SQLCipher extension. The extension is not default for sqlite. We need to apply it manually by building from source code like this link(https://github.com/mapbox/node-sqlite3#building-for-sqlcipher).

If you have a good idea to make a test case for this feature, please let me know.

Thanks,
JJ

### DIFF
--- a/src/driver/sqlite/SqliteConnectionOptions.ts
+++ b/src/driver/sqlite/SqliteConnectionOptions.ts
@@ -15,4 +15,9 @@ export interface SqliteConnectionOptions extends BaseConnectionOptions {
      */
     readonly database: string;
 
+    /**
+     * Encryption key for for SQLCipher.
+     */
+    readonly key?: string;
+
 }
--- a/src/driver/sqlite/SqliteDriver.ts
+++ b/src/driver/sqlite/SqliteDriver.ts
@@ -97,6 +97,14 @@ export class SqliteDriver extends AbstractSqliteDriver {
                     if (err) return fail(err);
                     ok(databaseConnection);
                 });
+
+                // in the options, if encryption key for for SQLCipher is setted.
+                if (this.options.key) {
+                  databaseConnection.run(`PRAGMA key = ${this.options.key};`, (err: any, result: any) => {
+                    if (err) return fail(err);
+                    ok(databaseConnection);
+                  });
+                }
             });
         });
     }


pleerock:
Hi,

Since its an extension, Im not sure if this change should be part of the package. Why don't you execute sql query once your connection is made in your codebase?

JeongJun-Lee:
I have already verified it by below test code. But as I have mentioned before, only in case the database is encrypted by the encryption key, this test case will be passed.
```
import "reflect-metadata";
import {createTestingConnections, closeTestingConnections, reloadTestingDatabases} from "../../utils/test-utils";
import {Connection} from "../../../src/connection/Connection";
import {expect} from "chai";

describe("github issues > #1437 [Feature Request] SQLCipher connection", () => {

        enabledDrivers: ["sqlite"],
        driverSpecific: {
          key: "test"
        },
        schemaCreate: true,
        dropSchema: true,
    }));
    beforeEach(() => reloadTestingDatabases(connections));
    after(() => closeTestingConnections(connections));

    it("should make a connection to encrypted database", () => Promise.all(connections.map(async connection => {

      const post = new Post();
      post.title = " About My Post   ";
      await connection.manager.save(post);
      post.title.should.be.equal(" About My Post   ");

      const loadedPost = await connection.manager.findOne(Post, { id: 1 });
      expect(loadedPost).not.to.be.undefined;
      loadedPost!.title.should.be.equal(" About My Post   ");

    })));

});
```

JeongJun-Lee:
Especially at mobile App, this usage of this extension is very broad and almost default for protecting data and design of sqlite database. Actually I need this extension for Desktop App made by Electron because of same reason as in case of mobile App. Thanks.

pleerock:
Okay lets merge it, thanks

JeongJun-Lee:
Hello @pleerock,

We're waiting for new version release including this merge commit since v0.2.7. I have tried to apply v0.3.0-alpha series to our codebase but it's impossible to use it in our codebase by this issue https://github.com/typeorm/typeorm/issues/2702. Do you have a plan to release v0.2.8 with bug fixes before v0.3-public? If then, pls certainly include this commit by cherry-pick. Very thanks.

PR #2416, opened 2018-06-27
Add test case for issue 2373

dominic-simplan:
Test for for deleting an entity with a cascaded nested entity.


imnotjames:
Rebased against master.

imnotjames:
Looking at this test case, there's some issues.

You don't wait for the for count to complete.
You don't wait for the remove to complete.

I'm going to fix the test case.

imnotjames:
Tests are passing, I don't think this is an actual issue.

PR #2423, opened 2018-06-28
Spatial support for PostgreSQL using PostGIS

### LINKED ISSUE #370: Support for PostGIS data types
Do you plan to support [PostGIS](http://postgis.net/docs/reference.html) geography data types?

mojodna:
This includes support for both geometry and geography columns as well as GiST indices (used when passing the `spatial: true` option). Client representations use GeoJSON (existing MySQL and MS SQL drivers use WKT (well-known text)) for compatibility with geospatial libraries such as Turf, JSTS, etc.

Fixes #370

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -39,7 +39,7 @@ services:
 
   # postgres
   postgres:
-    image: "postgres:9.6.1"
+    image: "mdillon/postgis:9.6"
     container_name: "typeorm-postgres"
     ports:
       - "5432:5432"
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -39,6 +39,9 @@ services:
 
   # postgres
   postgres:
+    # mdillon/postgis is postgres + PostGIS (only). if you need additional
+    # extensions, it's probably time to create a purpose-built image with all
+    # necessary extensions. sorry, and thanks for adding support for them!
     image: "mdillon/postgis:9.6"
     container_name: "typeorm-postgres"
     ports:
--- a/src/driver/postgres/PostgresDriver.ts
+++ b/src/driver/postgres/PostgresDriver.ts
@@ -145,13 +145,18 @@ export class PostgresDriver implements Driver {
         "numrange",
         "tsrange",
         "tstzrange",
-        "daterange"
+        "daterange",
+        "geometry",
+        "geography"
     ];
 
     /**
      * Gets list of spatial column data types.
      */
-    spatialTypes: ColumnType[] = [];
+    spatialTypes: ColumnType[] = [
+        "geometry",
+        "geography"
+    ];
 
     /**
      * Gets list of column data types that support length by a driver.
@@ -283,7 +288,10 @@ export class PostgresDriver implements Driver {
         const hasHstoreColumns = this.connection.entityMetadatas.some(metadata => {
             return metadata.columns.filter(column => column.type === "hstore").length > 0;
         });
-        if (hasUuidColumns || hasCitextColumns || hasHstoreColumns) {
+        const hasGeometryColumns = this.connection.entityMetadatas.some(metadata => {
+            return metadata.columns.filter(column => this.spatialTypes.indexOf(column.type) >= 0).length > 0;
+        });
+        if (hasUuidColumns || hasCitextColumns || hasHstoreColumns || hasGeometryColumns) {
             await Promise.all([this.master, ...this.slaves].map(pool => {
                 return new Promise((ok, fail) => {
                     pool.connect(async (err: any, connection: any, release: Function) => {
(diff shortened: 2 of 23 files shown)


mojodna:
@AlexMesser

OKNoah:
@mojodna Thanks, I was just about to do this! I'm going to try add a test or function for `orderBy` based on distance from a certain input.

mojodna:
See https://github.com/mojodna/typeorm/pull/1 for more discussion on this.

test/functional/query-builder/order-by/entity/Post.ts
```diff
@@ -21,4 +22,8 @@ export class Post {
     @Column()
     num2: number = 1;
 
-}
\ No newline at end of file
+    @Column({ type: "geometry" })
+    @Index({ spatial: true })
```
mojodna: This looks like it broke the build; some of the before/after hooks are now hanging. Is there a good way to diagnose this?

mojodna: Ah, it looks like introducing support for spatial indexes in Postgres is conflicts with tests that also execute in MySQL. I think I've just pushed a fix for this.

src/driver/postgres/PostgresDriver.ts
```diff
@@ -370,7 +384,7 @@ export class PostgresDriver implements Driver {
             || columnMetadata.type === "timestamp without time zone") {
             return DateUtils.mixedDateToDate(value);
 
-        } else if (columnMetadata.type === "json" || columnMetadata.type === "jsonb") {
+        } else if (this.spatialTypes.concat(["json", "jsonb"]).indexOf(columnMetadata.type) >= 0) {
```
pleerock: previous one was better =)

docker-compose.yml
```diff
@@ -39,7 +39,7 @@ services:
 
   # postgres
   postgres:
-    image: "postgres:9.6.1"
+    image: "mdillon/postgis:9.6"
```
pleerock: I guess somebody else can easily create a PR for the feature with specific extension and put another docker image with that extension, but without gis let's say. So better solution would be to have a base postgres image with all modifications applied

pleerock:
wow, this is huge. This wasn't an easy thing to implement, how much time it took you to bring all this into TypeORM?)

Lets ask @AlexMesser if he can review it.

docker-compose.yml
mojodna: Agreed. `mdillon/postgis` is `postgres` + PostGIS. Is this worth punting on until someone else needs to add another extension? (We can add a note/apology here ;-)

src/driver/postgres/PostgresDriver.ts
mojodna: _waits for `Array.includes` in TS..._

Local style should be:

```typescript
} else if (columnMetadata.type === "json" ||
           columnMetadata.type === "jsonb" ||
           this.spatialTypes.indexOf(columnMetadata.type) >= 0) {
```

❓

mojodna:
Thanks! I'm guessing ~8 hours (including getting familiar with TypeORM generally while on a plane) over the span of a week and a half..

src/driver/postgres/PostgresDriver.ts
pleerock: no, thats fine as well, you can keep it. I simply don't like `concat`, I would either go with "local style" you wrote or `["json", "jsob", ... this.spatialTypes].indexOf()` which looks much better then concat

docker-compose.yml
pleerock: haha I yeah... I hope it won't be me 😆

src/driver/postgres/PostgresDriver.ts
mojodna: Ah, right, I always forget about spreads in this context.

pleerock:
You are the beast!

mojodna:
@AlexMesser docs pushed.

pleerock:
Thank you very much! This will be published in 0.2.8 soon!

RDeluxe:
Awesome work @mojodna , thanks again !

htonkovac:
When can we expect 0.2.8 to be released?

sublimator:
Yes please to 0.2.8!

lorenzopicoli:
Any update on the 0.2.8 release @pleerock?

lucashsilva:
Looking forward for this

Vincz:
Hi @pleerock , Any update about the next release ?

hussainwali74:
I just want a simple example please:
how to insert lat-long point dataa into postgres, find distance between 2 points??

I have this column:
 ```
 @Index({ spatial: true })
  @Column({
    type: 'geometry',
    srid: 4326,
    nullable: true,
    spatialFeatureType: 'Point',
    transformer: {
      to: (v: Point) => {
        console.log(v);
        console.log(JSON.stringify(v));
        return eval(
          `ST_GeomFromGeoJSON('{"type":"Point","coordinates":[39.807222,-76.984722]}')`,
        );
      },
      from: (v: any) => {
        return v;
      },
    },
  })
  current_location: string;


```

when I try to insert data:
  "current_location": "{\"type\":\"Point\",\"coordinates\":[39.807222,-76.984722]}",

it says ST_GeomFromGeoJSON is not defined. Please help

mojodna:
It looks like you don’t have the PostGIS extension installed. To enable it in a specific database, run `create extension postgis` directly from `psql`.

PR #2435, opened 2018-06-29
fix splice error 2nd param

monkeymon:
**In transaction decorator, instead of replacing the "argsWithInjectedTransactionManagerAndRepositories" variable array index, it's inserting new value into the error.

In array.splice, we should use 1 in second parameter instead of 0 to keep inserting. 

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/splice

Signed-off-by: Chris Sim <monkeymon@gmail.com>

### DIFF
--- a/src/decorator/transaction/Transaction.ts
+++ b/src/decorator/transaction/Transaction.ts
@@ -87,7 +87,7 @@ export function Transaction(connectionOrOptions?: string | TransactionOptions): 
                     }
 
                     // replace method param with injection of repository instance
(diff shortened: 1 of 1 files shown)


pleerock:
Not sure if completely got the issue. 

But anyway Im not sure if I shall merge this PR since I want to deprecated transaction decorator and want to completely remove it in the next version. Please consider having your own decorator in your codebase.

monkeymon:
ohh no problem

pleerock:
its too way hacky and bring lot of confusion for some users, please consider moving it into your codebase or creating a separate package :)

Issue #2329, opened 2018-06-13
updating geometry Point type doesn't support "POINT(x y)" format unlike inserting


ingun37:
**Issue type:**

[ ] question
[x] bug report
[ ] feature request
[ ] documentation issue

**Database system/driver:**

[ ] `cordova`
[ ] `mongodb`
[ ] `mssql`
[x] `mysql` / `mariadb`
[ ] `oracle`
[ ] `postgres`
[ ] `sqlite`
[ ] `sqljs`
[ ] `react-native`

**TypeORM version:**

[ ] `latest`
[x] `@next`
[ ] `0.x.x` (or put your version here)

**Steps to reproduce or a small repository showing the problem:**

I have a property defined as 
```ts
@Column("point")
    geo: string
```
Updating value using "Point(x y)" format to this raises ER_CANT_CREATE_GEOMETRY_OBJECT (Inserting is fine)
e.g)
```ts
repository.create({"geo":"POINT(1 0)"}) //OK
repository.create({"geo":"POINT(0 0)"}) //OK
let a = repository.findOne(1)
a.geo = "POINT(1 0)"
repository.save(a) //ERROR
updateQueryBuilder.set({"geo":"POINT(0 0)"}).where("id = 1").execute() //ERROR
updateQueryBuilder.set({"geo":"POINT(1 0)"}).where("id = 1").execute() //ERROR
```

thank you

pleerock:
I need help from you and debug this thing. What query builder produces you when you do `updateQueryBuilder.set({"geo":"POINT(1 0)"}).where("id = 1").execute()` checkout SQL logs. Second, please provide SQL query that actually works if you run it directly in the database.

ingun37:
Here is a raw SQL query that works, production tested
```
UPDATE store SET geo=ST_GeomFromText('${geoRaw}') WHERE id=${storeid}
```
Heres output queryBuilder produced
```
(node:32729) UnhandledPromiseRejectionWarning: QueryFailedError: ER_CANT_CREATE_GEOMETRY_OBJECT: Cannot get geometry object from data you send to the GEOMETRY field
    at new QueryFailedError (/Users/ingun/Desktop/youngmoney/server/src/error/QueryFailedError.ts:7:9)
    at Query._callback (/Users/ingun/Desktop/youngmoney/server/src/driver/mysql/MysqlQueryRunner.ts:157:37)
[...]

PR #2449, opened 2018-07-02
Fix update of geospatial type column with mysql driver in UpdateQueryBuilder.ts

dreamersp:
In InsertQueryBuilder the geospatial types are addressed with GeomFromText() function passed to MySQL/MariaDB so we replicated the same behaviour on UpdateQuerBuilder.
We faced this reported issue: https://github.com/typeorm/typeorm/issues/2329 and needed a quick fix to continue working.


pleerock:
Thank you!

dreamersp:
A pleasure, thanks to you for doing such a remarkable work!!

PR #2459, opened 2018-07-04
fix bug of OrmUtils.compare2Objects

yoshiaki-natsume:
when id is binary and relation is nullable, compare2Objects raise error.

>TypeError: Argument must be a Buffer
>    at Buffer.equals (buffer.js:502:11)
>    at Function.OrmUtils.compare2Objects (../src/util/OrmUtils.ts:176:47)
>    at Function.OrmUtils.compare2Objects (../src/util/OrmUtils.ts:232:31)
>    at Function.OrmUtils.deepCompare (../src/util/OrmUtils.ts:114:23)
>    at Function.EntityMetadata.compareIds (../src/metadata/EntityMetadata.ts:723:25)
>    at ../src/persistence/SubjectChangedColumnsComputer.ts:150:13
>    at Array.forEach (native)
>    at SubjectChangedColumnsComputer.computeDiffRelationalColumns (../src/persistence/SubjectChangedColumnsComputer.ts:119:51)
>    at ../persistence/SubjectChangedColumnsComputer.ts:21:18
>    at Array.forEach (native)
>    at SubjectChangedColumnsComputer.compute (../src/persistence/SubjectChangedColumnsComputer.ts:19:18)
>    at SubjectExecutor.recompute (../src/persistence/SubjectExecutor.ts:167:45)
>    at new SubjectExecutor (../src/persistence/SubjectExecutor.ts:78:14)
>    at EntityPersistExecutor.<anonymous> (../src/persistence/EntityPersistExecutor.ts:119:28)
>    at step (../node_modules/typeorm/persistence/EntityPersistExecutor.js:32:23)
>    at Object.next (../node_modules/typeorm/persistence/EntityPersistExecutor.js:13:53)

this is an error case.

```TypeScript
class A {
  @ManyToOne(type => B, B => B.a, {})
  @JoinColumn({ name: "bId" })
  b: B | null;
}

class B {
  @Column("binary", {name: "id"})
  id: Buffer;

  @OneToMany(type => A, A => A.b)
  a: A[];
}
```

```TypeScript
const a = new A();
b.id = new Buffer([0]);
await a.save();

const b = new B();
await b.save();

a.b = b;
await a.save(); // this raise error
```

### DIFF
--- a/src/util/OrmUtils.ts
+++ b/src/util/OrmUtils.ts
@@ -177,7 +177,9 @@ export class OrmUtils {
         if (x === y)
             return true;
 
-        if (x.equals instanceof Function && x.equals(y))
+        // Check if both argument is Buffer
+        // Buffer.equals Argument must be a Buffer
+        if (Buffer.isBuffer(x) && Buffer.isBuffer(y) && x.equals(y))
             return true;
 
         // Works in case when functions are created in constructor.
--- a/src/util/OrmUtils.ts
+++ b/src/util/OrmUtils.ts
@@ -179,7 +179,12 @@ export class OrmUtils {
 
         // Check if both argument is Buffer
         // Buffer.equals Argument must be a Buffer
-        if (Buffer.isBuffer(x) && Buffer.isBuffer(y) && x.equals(y))
+        if (Buffer.isBuffer(x) && Buffer.isBuffer(y))
+            return x.equals(y);
+        if (Buffer.isBuffer(x) || Buffer.isBuffer(y))
+            return false;
+
+        if (x.equals instanceof Function && x.equals(y))
             return true;
 
         // Works in case when functions are created in constructor.
--- a/test/functional/util/compare-objects.ts
+++ b/test/functional/util/compare-objects.ts
@@ -1,0 +1,51 @@
+import "reflect-metadata";
+import {closeTestingConnections, createTestingConnections, reloadTestingDatabases} from "../../utils/test-utils";
+import {Connection} from "../../../src/connection/Connection";
+import {Post} from "./entity/Post";
+import {Counters} from "./entity/Counters";
+
+describe("util > compare2Objects functionality", () => {
+
+    let connections: Connection[];
+    before(async () => connections = await createTestingConnections({
+        entities: [Post, Counters],
+        enabledDrivers: ["mysql"]
+    }));
+    beforeEach(() => reloadTestingDatabases(connections));
+    after(() => closeTestingConnections(connections));
+
+    it("should compare two binary objects correctly", () => Promise.all(connections.map(async connection => {
+
+        // create and save a post and a counters first
+        const post = new Post();
+        post.id = new Buffer([0]);
+        await connection.manager.save(post);
+        const counters = new Counters();
+        counters.id = new Buffer([0]);
+        await connection.manager.save(counters);
+
+        // then update its properties and save again
+        const loadedPost = await connection.getRepository(Post).findOne();
+        loadedPost!.counters = counters!;
+        await connection.manager.save(loadedPost!);
+
+    })));
+
+    it("should compare binary and null objects correctly", () => Promise.all(connections.map(async connection => {
+
+        // create and save a post and a counters first
+        const counters = new Counters();
+        counters.id = new Buffer([0]);
+        await connection.manager.save(counters);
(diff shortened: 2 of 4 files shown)


src/util/OrmUtils.ts
```diff
@@ -177,7 +177,9 @@ export class OrmUtils {
         if (x === y)
             return true;
 
-        if (x.equals instanceof Function && x.equals(y))
+        // Check if both argument is Buffer
+        // Buffer.equals Argument must be a Buffer
+        if (Buffer.isBuffer(x) && Buffer.isBuffer(y) && x.equals(y))
```
yoshiaki-natsume: Does anyone know x and y have `.equals` method expect Buffer?

pleerock: isn't `.equals` come from buffer? If yes then maybe we can check if x and y are both instances of objects?

yoshiaki-natsume: equals come from Buffer.
Object.equals is not defined
```TypeScript
const a = new Buffer([]);
const b = {}
a.equals(b); // TypeError: Argument must be a Buffer
b.equals // undefined
```
So you must check instances are both Buffer or not.

pleerock: ah sorry looks like equals in in a Function. Why did you remove previous line?

```
if (x.equals instanceof Function && x.equals(y))
```

?

yoshiaki-natsume: Because it seems that line is just for Buffer class.
So I fixed that line(not remove).

pleerock: hmmm I wouldn't assume that. For example mongodb's ObjectId has equals method as well. I don't remember why exactly we need this (my bad on missing comment) but we need to put it back until we figure out that we really don't need it.

pleerock: Also please add a test that **fails** if someday somebody remove lines of code you have added.

yoshiaki-natsume: yes, I think we need more unit tests to describe logic.
ok, I'll prepare unit test for this case.

yoshiaki-natsume:
added unit test and fix condition

yoshiaki-natsume:
I made related Pull Request #2476

yoshiaki-natsume:
fixed unit tests

src/util/OrmUtils.ts
```diff
@@ -177,6 +177,13 @@ export class OrmUtils {
         if (x === y)
             return true;
 
+        // Check if both argument is Buffer
+        // Buffer.equals Argument must be a Buffer
+        if (Buffer.isBuffer(x) && Buffer.isBuffer(y))
+            return x.equals(y);
+        if (Buffer.isBuffer(x) || Buffer.isBuffer(y))
```
imnotjames, line 184: So there is never a case where a buffer could match something like a uintarray?

test/functional/util/compare-objects.ts
```diff
@@ -0,0 +1,51 @@
...
+        post.id = new Buffer([0]);
+        await connection.manager.save(post);
+        const counters = new Counters();
+        counters.id = new Buffer([0]);
+        await connection.manager.save(counters);
+
+        // then update its properties and save again
+        const loadedPost = await connection.getRepository(Post).findOne();
+        loadedPost!.counters = counters!;
+        await connection.manager.save(loadedPost!);
```
imnotjames, line 30: So there shouldn't be an exception here?  That's the expected results?

test/functional/util/compare-objects.ts
```diff
@@ -0,0 +1,51 @@
...
+import {Connection} from "../../../src/connection/Connection";
+import {Post} from "./entity/Post";
+import {Counters} from "./entity/Counters";
+
+describe("util > compare2Objects functionality", () => {
+
+    let connections: Connection[];
+    before(async () => connections = await createTestingConnections({
+        entities: [Post, Counters],
+        enabledDrivers: ["mysql"]
```
imnotjames, line 12: Should this be opened up for all drivers?  Or at least, more drivers?

imnotjames (commented):
Can you rebase and fix conflicts

AlexMesser:
This is quite old PR, and since there is no community feedback on this issue, I assume there is no interested in this change from the community which makes these changes disputable. Closing for now.

Issue #2464, opened 2018-07-04
ManyToMany onDelete option


lucas-coelho:
**Issue type:**

[ ] question
[x] bug report
[ ] feature request
[ ] documentation issue

**Database system/driver:**

[ ] `cordova`
[ ] `mongodb`
[ ] `mssql`
[ ] `mysql` / `mariadb`
[ ] `oracle`
[x] `postgres`
[ ] `sqlite`
[ ] `sqljs`
[ ] `react-native`

**TypeORM version:**

[x] `latest`
[ ] `@next`
[ ] `0.x.x` (or put your version here)

**Steps to reproduce or a small repository showing the problem:**

ManyToMany decorator `onDelete` options is not working. 
The  default value is 'CASCADE' and its not possible to change. 

```
export class Foo {

  @PrimaryGeneratedColumn()
  id: number;

  @JoinTable()
  @ManyToMany(_ => Bar, bar => bar.foos,  {
    cascade: false,
  })
  bars: Bar[]

}
```

pleerock:
Im going to release this change end of this week. Thank you!

P0oOOOo0YA:
Still the same using typeorm with SQLite onDelete: CASCADE doesn't work. Entities have ManyToMany relation.

PR #2465, opened 2018-07-05
Fix many to many on delete option

lucas-coelho:
## Changes

  - Update ORM config in developer documentation
  - Remove `.only` in `value-transformer.ts` test 
  - Fix ManyToMany onDelete option

## Issue
 - #2464


pleerock:
Thank you!

PR #2466, opened 2018-07-05
Let's remain positive.

rhalff:


rhalff:
Sorry, I assumed it was a typo,  I should have read the example better.

PR #2471, opened 2018-07-06
Added a test to see if the returned object from the DB is a Date object

qbodart:
It seems that the function save of the repository does not return the object from the DB but rather a mix from the previous reference of the variable updated with some data from the DB.


qbodart:
Apparently my test does not fail so... I still need to work on it :-)

Issue #2470, opened 2018-07-06
Error persisting quoted field names inside entity


marceloverdijk:
**Issue type:**

[x] question
[x] bug report
[ ] feature request
[ ] documentation issue

**Database system/driver:**

[ ] `cordova`
[ ] `mongodb`
[ ] `mssql`
[ ] `mysql` / `mariadb`
[ ] `oracle`
[ ] `postgres`
[x] `sqlite`
[ ] `sqljs`
[ ] `react-native`

**TypeORM version:**

[x] `latest`
[ ] `@next`
[ ] `0.x.x` (or put your version here)

**Steps to reproduce or a small repository showing the problem:**

I have an `Building` entity containing a quoted (`"`) field like:

```
@JoinColumn({ name: "constructor_id" })
@ManyToOne(type => Constructor, { eager: true })
public "constructor"?: Constructor;
```

I'm quoting the constructor field as `constructor` is a reserved word and and I have the need to call it `constructor` anyway.

When trying:

```
// constructing building and setting other properties omitted...
building["constructor"] = await constructorRepository.findOne(constructorId);
buildingRepository.save(building);
```

It gives the following error:

```
query failed: INSERT INTO "building"("name", "city_id", "constructor_id") VALUES (?, ?, [object Object]) -- PARAMETERS: ["building 1","amsterdam",1]
error: { Error: SQLITE_ERROR: no such column: object Object errno: 1, code: 'SQLITE_ERROR' }
```

I think the problem is that typeorm tries to save the reference to he constructor as a `[object Object]` as can be seen in the logging.

I will make a test case later.

imnotjames:
This should be fixed as of #7821 but I'll double check

imnotjames:
Actually, I can't replicate this because it's no longer valid Typescript.

I think the issue primarily had to do with you overloading `constructor` as a field rather than the quotes.  I tried a few other ways to try to replicate this and couldn't.

PR #2472, opened 2018-07-06
PR for proving issue 2470

marceloverdijk:
This PR proves #2470.

The test currently succeeds as I test explicitly on the error that I'm getting. 
The assertion should be negated when a solution is found.


imnotjames:
>Actually, I can't replicate this because it's no longer valid Typescript.
>
> I think the issue primarily had to do with you overloading `constructor` as a field rather than the quotes.  I tried a few other ways to try to replicate this and couldn't.

PR #2476, opened 2018-07-06
fixed fail update relation entity bug when relation entity have binary primary key.

yoshiaki-natsume:


src/query-builder/UpdateQueryBuilder.ts
```diff
@@ -370,7 +370,7 @@ export class UpdateQueryBuilder<Entity> extends QueryBuilder<Entity> implements
                     //
                     let value = column.getEntityValue(valuesSet);
                     if (column.referencedColumn && value instanceof Object) {
-                        value = column.referencedColumn.getEntityValue(value);
+                        value = column.referencedColumn.getEntityValue(valuesSet);
```
yoshiaki-natsume: I don't know this is destructive change or not.
But I passed unit tests (on my local env).
Does these unit tests cover all cases?

pleerock: believe me or not but there are at least million of use cases in orm. So no, it does not cover all million cases :) but change is most probably destructive.

src/metadata/ColumnMetadata.ts
```diff
@@ -592,7 +592,11 @@ export class ColumnMetadata {
                 value = this.referencedColumn.getEntityValue(PromiseUtils.extractValue(entity[this.propertyName]));
 
             } else {
-                value = entity[this.propertyName];
+                if (Buffer.isBuffer(entity)) {
+                    value = entity;
+                } else {
+                    value = entity[this.propertyName];
+                }
```
yoshiaki-natsume: I tried to change logic to not destructive.
What do you think?

yoshiaki-natsume:
in order to pass this unit tests, you need to merge #2459.

src/metadata/ColumnMetadata.ts
pleerock: looks like a hack. I'll left this PR opened and take a look on a problem by myself once I get time for it.

yoshiaki-natsume: I got it 😃

imnotjames (commented):
Please include a description and if there is one an issue number to link this to.

You have the PR you referenced as part of this but the tests still don't pass. Can you rebase and fix the tests?

AlexMesser:
I'm closing this as it's failing, has conflicts and hasn't been updated in a few years. Anyway, thank you for contribution.

Issue #1767, opened 2018-03-16
Change removeById method to act like remove method


jdforsythe:
**Issue type:**

[ ] question
[ ] bug report
[x] feature request
[x] documentation issue

**Database system/driver:**

[ ] `cordova`
[ ] `mongodb`
[ ] `mssql`
[ ] `mysql` / `mariadb`
[ ] `oracle`
[x] `postgres`
[ ] `sqlite`
[ ] `sqljs`
[ ] `websql`

**TypeORM version:**

[x] `latest`
[x] `@next`
[ ] `0.x.x` (or put your version here)

**Steps to reproduce or a small repository showing the problem:**

In the `EntityManager` the methods that modify data come in two groups - you pass an entity in and cascades/relations/other operations happen (`save()` and `remove()`), or you pass in an id/conditions and a primitive operation happens without the cascades/relations (`insert()`, `update()`, `updateById()`, `delete()`, `deleteById()`)

When updating by id, you just add the id property to the entity and save. However, when deleting, if you want the operations to happen instead of primitive, and you just have the id, you must perform two database operations:
  1. get the entity with `findOneById()`
  2. pass the entity to `remove()` to perform the DELETE

There is a `removeById()` method but it is marked as deprecated in the code and just calls the primitive `deleteById()`. It would be more convenient to have a `removeById()` method that performs the same operations as `remove()` but doesn't require the extra step of doing the `find` to get entities to pass in.

Also note that in the [Repository API](https://github.com/typeorm/typeorm/blob/master/docs/repository-api.md) docs the `removeById()` and `removeByIds()` methods are still listed, not as deprecated, and the `delete` methods aren't listed. In `next` [Repository API](https://github.com/typeorm/typeorm/blob/next/docs/repository-api.md) docs, none of the 'byId' methods are listed. The docs would lead you to believe that `delete()` is the way to delete by id, but doesn't mention that none of the cascade/relations/etc will be run.

So there are two propositions here:

1. implement a `removeById()` method to remove entities by id and still perform the cascades
2. update the documentation to show all available methods and explain the loss of cascading in certain methods

pleerock:
In latest you can do `repository.remove(repository.create({ id: 1 }))`. In `next` we already support following syntax:

```ts
repository.save({  id: 1, ... })
```

because of using DeepPartial instead of Entity as a first argument. Im planning to do same with `remove` method and you'll be able to use following signature:

```ts
repository.remove({  id: 1  })
```

[...]

PR #2477, opened 2018-07-06
Fixed the Init template code so that the deprecated RemoveById call is no longer there

mandreko:
Per: https://github.com/typeorm/typeorm/issues/1767


pleerock:
Thank you! Would be great if you can check other commands if something obsolete is left there as well! Thanks!

mandreko:
I'll be looking. I just started using typeorm for a project, so we will see if anything else is outdated. Thanks!

PR #2478, opened 2018-07-07
Fix manager.decrement in docs

guuurgle:


pleerock:
Thank you very much GMan!

PR #2484, opened 2018-07-08
Issue 1738 advanced console formatted

aaronstaves:
I've put this off for a while just because I've had stuff working locally and haven't had a need for it.

This is a pretty simple addition, but could easily be tweaked going forward if need be.  For now I've just extended AdvancedConsoleLogger and just added in the bits we'd need to format SQL for logQuery calls.  I'm completely open to reworking this if another method is preferred (i.e. build this into it's own standalone logger).

One thing I did notice on one comment a while back, someone was complaining about extra packages included for the debug/loggers.  I'm wondering if it would be worthwhile to just add this as a separate package, something like `@typeorm/advanced-console-formatted` and then people can import and use it as they see fit.

Anyways, apologies again for not following through on this sooner. Let me know if/how this needs to be reworked.  Right now I just want to get this out there to discuss and see what, if anything, fails in the pipeline.  Thanks!

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "typeorm",
-  "version": "0.2.6",
+  "version": "0.2.7",
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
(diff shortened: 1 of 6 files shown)


pleerock:
> Anyways, apologies again for not following through on this sooner.

Its open source and any efforts even latest ones are highly appreciated! Thank you!

> . I'm wondering if it would be worthwhile to just add this as a separate package, something like @typeorm/advanced-console-formatted and then people can import and use it as they see fit.

I hope we'll do things this way in the future.

pleerock:
Can you please add documentation to the new logger (you can attach screenshot btw)

imnotjames:
Some of this has been brought in via other PRs.  I'll be closing this out but if you think there's still more please rebase & open a new PR.

PR #2495, opened 2018-07-10
Exposed the ValueTransformer

mohyt:
Updated the index.ts, to expose the ValueTransformer interface


mohyt:
@pleerock : We are using the typeORM, but it seems ValueTranformer is not exposed. So in order to create a custom transformer we need to extend the ValueTransformer.

lorenzleutgeb:
Maybe squash these two commits to hide the typo.

PR #2507, opened 2018-07-12
Redis should set expiry time

kikar:


kikar:
@pleerock Any ETA on next release in which this will come out?

Issue #2531, opened 2018-07-16
nativescript support breaks browser in react-native


arolson101:
**Issue type:**

[ ] question
[X] bug report
[ ] feature request
[ ] documentation issue

**Database system/driver:**

[ ] `cordova`
[ ] `mongodb`
[ ] `mssql`
[ ] `mysql` / `mariadb`
[ ] `oracle`
[ ] `postgres`
[ ] `sqlite`
[ ] `sqljs`
[X] `react-native`
[ ] `expo`

**TypeORM version:**

[X] `latest`
[ ] `@next`
[ ] `0.x.x` (or put your version here)

**Steps to reproduce or a small repository showing the problem:**
I've got a react-native application using typeorm.  I cloned the repository (trying to fix https://github.com/typeorm/typeorm/issues/2502) and the package I build uses nativescript-sqlite though I don't want to use it.

```
Loading dependency graph, done.
error: bundling failed: Error: Unable to resolve module `nativescript-sqlite` from `C:\crap\node_modules\typeorm\build\package\browser\driver\nativescript\NativescriptDriver.js`: Module `nativescript-sqlite` does not exist in the Haste module map
```

pleerock:
thanks for the report we are going to fix it

championswimmer:
requires are not supported in browser runtime https://github.com/typeorm/typeorm/blob/master/src/platform/BrowserPlatformTools.template#L33 
so how exactly do we work around this ?

championswimmer:
ALso @arolson101 did this error occur on `typeorm@latest` ? If I am not wrong `typeorm@0.2.7` is latest on npmjs.com and that version doesn't even have nativescript support on it.

pleerock:
@championswimmer he build repository manually. 

Yeap so I have another report that tells that typeorm in browser has same issue with react native when he simply runs in browser. This issue should be fixed for both react-native and nativescript.

@daniel-lang but I guess we cannot replace to BrowserPlatformTools.load in react native (and nativescript) right? Is react-native uses `typeorm/browser` ? Looks like a serious issue.

arolson101:
@championswimmer I’m using a forked version of the latest to fix a (different) bug.

arolson101:
@pleerock yes, react-native uses typeorm/browser

championswimmer:
We need to whitelist the drivers in Webpack so that webpack doesn't try to parse the require statements for optional modules.

championswimmer:
[...]

PR #2553, opened 2018-07-20
require react native and nativescript drivers dynamically

championswimmer:
these require lines should not be resolved by Webpack, so using
runtime resolved variable here instead of constant strings.

Signed-off-by: Arnav Gupta <arnav@codingblocks.com>

### DIFF
--- a/src/driver/nativescript/NativescriptConnectionOptions.ts
+++ b/src/driver/nativescript/NativescriptConnectionOptions.ts
@@ -15,4 +15,9 @@ export interface NativescriptConnectionOptions extends BaseConnectionOptions {
      */
     readonly database: string;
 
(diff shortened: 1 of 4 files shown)


championswimmer:
@pleerock @arolson101 @daniel-lang 

This solves https://github.com/typeorm/typeorm/issues/2531 possibly
Also lets people use their own custom Nativescript/ReactNative driver by passing it into constructor (if they want to use a different driver, or else falls back to using default)

Kononnable:
@championswimmer 
Since #2854 is merged and it resolves problem with browser based drivers are we still needing this PR?

championswimmer:
yeah closing this now, we'll see for v3

Issue #2499, opened 2018-07-11
Postgres DELETE query result is useless


newsiberian:
**Issue type:**
Hi, @pleerock, I faced with strange results from .remove() and .delete()  queries. delete() query result are pretty much useless currently. It is always `{ raw: [] }`.

Here is debugging result from [here](https://github.com/typeorm/typeorm/blob/master/src/driver/postgres/PostgresQueryRunner.ts#L177)
![typeorm_1](https://user-images.githubusercontent.com/11130487/42551885-c20dc070-8503-11e8-9143-6a052d23dadc.png)

In addition, there is a lack in docs query results. Is it possible to consider the possibility of changing api and return something valuable. In my opinion, the number of removed rows will be the best for .remove() and .delete() queries.

[ ] question
[ ] bug report
[x] feature request
[x] documentation issue

**Database system/driver:**

[ ] `cordova`
[ ] `mongodb`
[ ] `mssql`
[ ] `mysql` / `mariadb`
[ ] `oracle`
[x] `postgres`
[ ] `sqlite`
[ ] `sqljs`
[ ] `react-native`
[ ] `expo`

**TypeORM version:**

[x] `latest`
[ ] `@next`
[ ] `0.x.x` (or put your version here)

**Steps to reproduce or a small repository showing the problem:**

<!--
    To answer those questions you need to put "x" inside the square brackets, for example:
    [x] `mysql`
    [ ] `postgres`

    Also, please format your code properly (by taking code blocks into ```ts .... ```)
!>

pleerock:
yes, we already had such feature request. We can return more information in Insert/Update/DeleteResult objects. Feel free to contribute!

vlapo:
@newsiberian Is this issue resolved?

newsiberian:
@vlapo, nope, even after pr was merged, I still receive the same results

pleerock:
how come? it was merged a new version was released, why do you still have same results? do you have your changes in the latest version you have installed?

Kononnable:
Closing as no response from the author.

PR #2626, opened 2018-08-06
Return number of affected rows on delete

newsiberian:
Hi, this is related with #2499 
Here we additionally receive new variable `affected` which is [`rowCount`](https://github.com/brianc/node-postgres/blob/master/lib/result.js#L46-L49) from `pg`.

UPD:
Number of affected rows returns for `mysql`, `mariadb`, `mssql`, `postgres`

I was unable to run Oracle locally, so it is not covered. `sqlite` & `sqljs` seems requires additional queries to getting such information. Mongodb seems to have an internal bug. It is always returns `{ n: 0; ok: 1 }` for me and this is goes not from the nodejs `mongo-core` package, but from db itself, so I think db is bugged. at least for node.js. Mongo console operations works as expected.

Also I limited the launch of tests only for covered databases.


newsiberian:
@pleerock, please review

pleerock:
Sorry for such delayed merge, was really busy recently. PR looks good, thank you very much!

PR #2664, opened 2018-08-15
Wrong decrement documentation in manager api

gabrielberlanda:
Fix decrement documentation in entity manager api


pleerock:
Thank you very much!

PR #2665, opened 2018-08-15
Simple ENV addition

jstewart37:
Addition of debug to env variables is usefull for myself and maybe others.

### DIFF
--- a/docs/using-ormconfig.md
+++ b/docs/using-ormconfig.md
@@ -125,6 +125,7 @@ List of available env variables you can set:
 * TYPEORM_MIGRATIONS_DIR
 * TYPEORM_SUBSCRIBERS_DIR
 * TYPEORM_DRIVER_EXTRA
+* TYPEORM_DEBUG
 
 `ormconfig.env` should be used only during development.
 On production you can set all these values in real ENVIRONMENT VARIABLES.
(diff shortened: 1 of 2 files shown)


pleerock:
Looks like this is mysql specific. Not sure if I should merge it, this is not scalable way of doing things. What lets say if we some option `x` that makes one thing and accepts one value in one driver and makes different thing and accepts different value in a different driver.

adding @Kononnable

Kononnable:
I wouldn't worry here about two drivers doing different things with such values. This should be validated when option is added to `ConnectionOptions` type.
But I think we already have got to many options which can be introduced as environment variables. We should either choose to define every option introduced in ConnectionOptions in environment variables or choose only environment specific(database connection info etc).

It's not like we're polluting environment variables, but we're encouraging people to do so. 
Typeorm is just a lib, not an end-application, so a better design would be to configure only end-application in env variables, not every lib used by it. End-application would then pass expected values to libs.

pleerock:
Yeah thats why I want to extract this feature out of typeorm in the future. e.g. remove ormconfig. configuration loading and all that stuff, maybe into separate package. All this was done for a convenience but people have different use cases and we can't cover all of them with this additional convenience.

What shall we do with this PR then

pleerock:
@Kononnable you are kind man haha. Yeah you are right, lets merge it

PR #2674, opened 2018-08-16
Update express example

rhlsthrm:
Current example does nothing when the endpoints are called.


afurculita:
There is no need for `res.json`. A middleware can be used to convert a response to JSON. Something like: https://github.com/jakjothi/express-response/blob/master/lib/express-response.js

cvh23:
Update operation (HTTP PUT) is missing. How can I update an entity with a partial update request object?

Issue #2441, opened 2018-07-02
MySQL Connection Pool size and close/timeout issue


sagarch88:
**Issue type:**

[x ] question
[ ] bug report
[ ] feature request
[ ] documentation issue

**Database system/driver:**

[ ] `cordova`
[ ] `mongodb`
[ ] `mssql`
[ x] `mysql` / `mariadb`
[ ] `oracle`
[ ] `postgres`
[ ] `sqlite`
[ ] `sqljs`
[ ] `react-native`

**TypeORM version:**

[x ] `latest`
[ ] `@next`
[ ] `0.x.x` (or put your version here)

1. Does TypeORM for MySQL by default uses connection pool? (I believe it does)
2. How do I specify the connectionPool limit in TypeORM for MySQL?
3. Should I manually release the connection from connectionPool? If yes, how do I release the connection?

I have followed the documentation for the implementation of TypeORM. Following is the ormconfig.json
`[

    {
        "name": "default",
        "type": "mysql",
        "host": process.env.DATABASE_HOST,
        "port": process.env.DATABASE_PORT,
        "username": process.env.DATABASE_USER,
        "password": process.env.DATABASE_PASSWORD,
        "connectTimeout": 1500,
        "database": process.env.DATABASE_NAME,
        "entities": [
            "src/models/entities/**/*.ts"
        ],
        "logging": true
    }
]`

I am having issues with TypeORM with MySQL.

1. Immediately after NodeJS application is started , the MySQL query takes 1 second to respond. If no more requests are received to the server, the connection to MySQL is still active, but under sleep state(By running `show processlist;` on MySQL - application user is still active.). After the idle time of around 10 minutes, If there is other request, I get the response in around 20 seconds for the 1st time and then immediately, for subsequent requests.

I have traced the issue to the following line of code in **my Application**:
        `await getConnectionManager().get().transaction        // Takes around 20sec.`
I have tried with `getManager().transaction` too but the issue persists. By splitting the above statement into 2 lines, 
    `const connectionManager = getManager();`
    `connectionManager.transaction();`, getting the transaction is taking the time.

2. The connections are never closed on the MySQL server side. In the following image, where application user is `unified`. 20+ users are under sleep connection even when there were no requests to the server.
![screen shot 2018-07-02 at 9 56 00 am](https://user-images.githubusercontent.com/23547857/42145906-c2c052cc-7de1-11e8-9e1f-b084790364d2.png)

[...]

PR #2681, opened 2018-08-19
Update transactions.md (add `queryRunner.release()` statement)

seungha-kim:
Hi! I'm a typeorm beginner :)  It would be good if there is some examples in transaction doc, about that a manually created query runner needs to be released. It will be helpful to beginners like me.

Followings are long explanation about my situation.

I'm using transaction in my unit tests. My transaction helper function is here:

```ts
export const withTx = (f: (manager: EntityManager) => Promise<any>) => {
  return async () => {
    const connection = getConnection()
    const qr = connection.createQueryRunner()
    await qr.connect()
    await qr.startTransaction()
    try {
      await f(qr.manager)
    } finally {
      await qr.rollbackTransaction()
    }
  }
}
```

And I almost **copied and pasted** this from [here](http://typeorm.io/#/transactions/using-queryrunner-to-create-and-control-state-of-single-database-connection).

But I've run into a problem that query runners cannot start transaction after 10 test started. It was problem about connection pool, and it was caused because there is no release statement in my transaction helper function.

So I added release statement like this, and the problem is solved:

```ts
export const withTx = (f: (manager: EntityManager) => Promise<any>) => {
  return async () => {
    const connection = getConnection()
    const qr = connection.createQueryRunner()
    await qr.connect()
    await qr.startTransaction()
    try {
      await f(qr.manager)
    } finally {
      await qr.rollbackTransaction()
      await qr.release() // ***added***
    }
  }
}
```

Before test error occurred, I had no idea what problem is in my code. The added example can warn beginners to add release statement.

### DIFF
--- a/docs/transactions.md
+++ b/docs/transactions.md
@@ -153,6 +153,10 @@ try {
     // since we have errors lets rollback changes we made
     await queryRunner.rollbackTransaction();
     
+} finally {
+    
+    // you need to release query runner which is manually created:
+    await queryRunner.release();
 }
 ```
 


Slessi:
Related to https://github.com/typeorm/typeorm/issues/2441#issuecomment-415055432

PR #2692, opened 2018-08-21
fix(transactions): remove session keyword from mysql transaction init…

travtarr:
In response to #2227  [comment](https://github.com/typeorm/typeorm/pull/2227#pullrequestreview-148154828)

For some reason I specified `SESSION` in the initiation of a transaction when an isolation level was provided for the MySQL runner, this removes that keyword.


travtarr:
Also had to change ordering around since MySQL doesn't allow the isolation level to be changed on an active transaction.

pleerock:
Thank you! Sorry for late merge (

PR #2717, opened 2018-08-27
Fix misspelled column type

afurculita:
Fix misspelled column type.

This causes the following problem: if you have a `simple-array` column, when doing a save on the entity, because the diff algorithm has this column type name misspelled, the column will be always marked for update, even though the value didn't change.


afurculita:
@Kononnable can you please take a look over this? This should be an easy one. Thanks

pleerock:
Thank you very much!

Issue #2215, opened 2018-05-25
Inserting id (primary column) from code.


PavanBahuguni:
**Issue type:**

[ ] question
[ ] bug report
[ x ] feature request
[ ] documentation issue

**Database system/driver:**

[ ] `cordova`
[ ] `mongodb`
[ ] `mssql`
[ ] `mysql` / `mariadb`
[ ] `oracle`
[ x ] `postgres`
[ ] `sqlite`
[ ] `sqljs`
[ ] `react-native`

**TypeORM version:**

[ x ] `latest`
[ ] `@next`
[ ] `0.x.x` (or put your version here)

**Steps to reproduce or a small repository showing the problem:**

I have a table with id as PrimaryGeneratedColumn. In some scenario i need to give the id in code and dont want to use the autogenerated id. Even if i give id, it is being ignored.

krislefeber:
What is your use case for this? If you don't want it to auto-generate, then don't use the PrimaryGeneratedColumn. If you manually set this id, the auto-increment could break on future entities that try to auto-generate an id with that value. I think it would be easier to have a separate field for the cases where you want to override the value.

PavanBahuguni:
I have a set of masterdata which have fixed ids, i have set the autogenerate sequence to start from after the ids which are allocated to masterdata. I want the masterdata to be inserted with the specific ids. Other data can get the db generated ids. Is it possible to do this?

pleerock:
I think this is duplicate of #2199 ?

teyou:
FYI Postgres has different way of `remembering` AUTO_INCREMENT value. 
I don't think my PR cater for Postgres database.

Tiketti:
I would also like to see this behavior reverted. Our use case is related to tests and inserting entities into the database with preset ID's and asserting API returns the right values. An example in pseudo code:

```
const person = new Person();
person.id = 1;
personRepo.save(person);
...

describe('GET /api/person/{id}', async () => {
  test(('should return person with correct id') => {
    const person = await fetch(url);
    expect(person).not.toBeUndefined();
  });
}
```

Above test fails, because `person` is inserted with an autoincremented id instead of `1` and can't be found.

hauau:
As a temporary workaround for MySQL/MariaDB/InnoDB setting autoincrement value for table and inserting in the desirable order is possible. This query rebuilds table and requires that no existing id is bigger than a new autoincrement value, you should probably truncate it beforehand.
[...]

PR #2722, opened 2018-08-28
Inserting id (primary column) from code.

andreasrueedlinger:
Add flag allowGeneratedValuesFromCode to connection options to configure the behavior of values set from code for generated columns (#2215 )

It has to be discussed if the current exceptions should be kept in code or (at least MySql) could be removed and just this option is provided for transparent design.

### DIFF
--- a/src/connection/BaseConnectionOptions.ts
+++ b/src/connection/BaseConnectionOptions.ts
@@ -104,6 +104,13 @@ export interface BaseConnectionOptions {
     readonly extra?: any;
 
     /**
+     * Allow to set values from code for generated columns for this connection.
+     * Use this only if you know what your are doing (e.g. inserting legacy data with pre-defined ids),
+     * and make sure the generated values will never collide with such a value from code.
+     */
+    readonly allowGeneratedValuesFromCode?: boolean;
+
+    /**
      * Allows to setup cache options.
      */
     readonly cache?: boolean|{
--- a/src/connection/Connection.ts
+++ b/src/connection/Connection.ts
@@ -111,6 +111,11 @@ export class Connection {
      */
     readonly relationIdLoader: RelationIdLoader;
 
+    /**
+     * Allow to set values from code for generated columns for this connection.
+     */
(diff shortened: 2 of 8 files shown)


andreasrueedlinger:
@pleerock of course we can add another exception for all drivers that support it. But I think it would be nicer design if you let the user decide if he wants to enable it in his specific situation. It would be more transparent and delegates the responsibility to the user (so you don't have to bother anymore which drivers would support it). In any case I would like to see some documentation about it somewhere.

pleerock:
> @pleerock of course we can add another exception for all drivers that support it. 

no, we don't need to add any exception. If user passed value and db supports we just use it, if user passed value and db doesnt support we just ignore it just like what we do currently.

This behaviour is too small to make it configurable. Configuration for it is redundant and adds complexity.

andreasrueedlinger:
> no, we don't need to add any exception. If user passed value and db supports we just use it, if user passed value and db doesnt support we just ignore it just like what we do currently.

But you can't say that db does not support it without knowing the exact situation. If you write unit tests with hard-coded primary keys and not using any sequence values I guess every db supports it? And it is exactly in such a scenario I would like to enable it through some kind of configuration. For production I agree with you about when to support it.

pleerock:
> This behaviour is too small to make it configurable. Configuration for it is redundant and adds complexity.

Please remove configuration and make it as #2218 If users will have demand for such configuration we'll rethink it again.

andreasrueedlinger:
> > This behaviour is too small to make it configurable. Configuration for it is redundant and adds complexity.
> 
> Please remove configuration and make it as #2218 If users will have demand for such configuration we'll rethink it again.

For which drivers then? Sqlite, postgres, ...?

pleerock:
what driver support having value as a primary generated column?

imnotjames:
This is pretty out of date and we haven't seen much activity.  If you get time to handle the conflicts & address the concerns, please feel free to open another PR!

Issue #2006, opened 2018-04-23
Columns are being set to null after saving the entity


asvetliakov:
**Issue type:**

[ ] question
[x] bug report
[ ] feature request
[ ] documentation issue

**Database system/driver:**

[ ] `cordova`
[ ] `mongodb`
[ ] `mssql`
[x] `mysql` / `mariadb`
[ ] `oracle`
[ ] `postgres`
[ ] `sqlite`
[ ] `sqljs`
[ ] `react-native`

**TypeORM version:**

[x] `latest`
[ ] `@next`
[ ] `0.x.x` (or put your version here)

**Steps to reproduce or a small repository showing the problem:**

Repro:

```ts

@Entity()
class User {
    @PrimaryGeneratedColumn()
    id: number;
    
    @Column("varchar", { nullable: true })
    token: string | null = null;
}

const u = new User();
u.token = "sometoken";

await db.manager.save(u);
u.token; // -> !!! null

const u2 = await db.manager.findOne(User, u.id);
u2.token; // "sometoken"

pleerock:
Added test for this issue and everything works fine. Please provide more details.

@asvetliakov can you please next time create a failing test once you submit issues you can reproduce with a test.

asvetliakov:
I added failing test. Sorry the repro example was incompleted.

Issue #2005, opened 2018-04-23
Boolean columns are being set to false after saving the entity


asvetliakov:
**Issue type:**

[ ] question
[ ] bug report
[ ] feature request
[ ] documentation issue

**Database system/driver:**

[ ] `cordova`
[ ] `mongodb`
[ ] `mssql`
[x] `mysql` / `mariadb`
[ ] `oracle`
[ ] `postgres`
[ ] `sqlite`
[ ] `sqljs`
[ ] `react-native`

**TypeORM version:**

[x] `latest`
[ ] `@next`
[ ] `0.x.x` (or put your version here)

**Steps to reproduce or a small repository showing the problem:**

Repro:

```ts

@Entity()
class User {
    @PrimaryGeneratedColumn()
    id: number;
    
    @Column({
        type: "tinyint",
        transformer: {
            from: val => !!val,
            to: val => val,
        },
    })
    activated: boolean;
}

const u = new User();
u.activated = true;

await db.manager.save(u);
u.activated; // false

const u2 = db.manager.findOne(User, u.id);
u2.activated; // true
```

asvetliakov:
Please check this repo:
https://github.com/asvetliakov/typeorm-bugs/tree/master/2005

For some reason the test in typeorm repo itself passes.

aasmpro:
same issue, Entity:
```
@Entity({ name: 'devices' })
export class Device {
	// other columns ...
	@Column({ name: 'haveWhy', default: 0, transformer: new NumberToBooleanTransformer() })
	haveWhy!: boolean;
}
```
transformer: 
```
export class NumberToBooleanTransformer implements ValueTransformer {
    from (value: number): boolean {
        console.log('sssssssssssssssssssssss',value, typeof value)
        if(value === 1){
            return true
        } else {
            return false
        }
    }

    to (value: boolean): number {
        console.log('xxxxxxxxxxxxxxxxxxxxxxx',value,  typeof value)
        if(value === true){
            return 1
        } else {
            return 0
        }
    }
}
```
my test:
```
it('test null value', async () => {
    await deviceRepository.clear();
    await deviceRepository.save({
        id: 1,
        haveWhy: true,
    });

    const device: Device | undefined = await deviceRepository.findOne(1);

    await expect(device!.serialNumber).toBe(null)
    await expect(device!.touchpointTypeId).toBe(1)
    await expect(device!.haveWhy).toBe(true) // this line fails
    });
```
test output:
```
[...]

PR #2723, opened 2018-08-28
Prevent setting to default value the columns that are being updated

afurculita:
This is a problem related to https://github.com/typeorm/typeorm/issues/2005 and https://github.com/typeorm/typeorm/issues/2006 and fixed for inserts in https://github.com/typeorm/typeorm/commit/6acbf70a98b7ba1035cd40b5cbb9a802c6b04024.

This PR fixes it also for updates.


afurculita:
@Kononnable can you please take a look over this? Please let me know if more has to be done. Thanks.

Kononnable:
I'm not familiar with this part of typeorm functionality, so I won't be much of help here. It looks like you took @pleerock fix and applied it in one more place. 
It would be great if that aditional fix will also be added to tests - probably `test/github-issues/2005/issue-2005.ts` or `test/github-issues/2006/issue-2006.ts` are good places to start.

pleerock:
Looks correct, thanks @afurculita !

Issue #2221, opened 2018-05-28
column option readonly is not working


whcarrot:
**Issue type:**

[ ] question
[x] bug report
[ ] feature request
[ ] documentation issue

**Database system/driver:**

[ ] `cordova`
[ ] `mongodb`
[ ] `mssql`
[x] `mysql` / `mariadb`
[ ] `oracle`
[ ] `postgres`
[ ] `sqlite`
[ ] `sqljs`
[ ] `react-native`

**TypeORM version:**

[x] `latest`
[ ] `@next`
[ ] `0.x.x` (or put your version here)

**Steps to reproduce or a small repository showing the problem:**

I set column's option below, but when updating, below column included.

```ts
	@Column({type: 'varchar', select: false, readonly: true})
	public provider_name: string;
```

So I inserted exception code query-builder/UpdateQueryBuilder.js:359

```ts
                    if(column.isReadonly) {
                        return ;
                    }
```

please fix it.

hielkehoeve:
I have created a PR for this issue: https://github.com/typeorm/typeorm/pull/2749

vlapo:
Close as duplicate of https://github.com/typeorm/typeorm/issues/2041

Issue #2041, opened 2018-04-27
Read-only setting is not working


DrunkenPoney:
It would be nice if there was a read-only option. It would prevent any non-voluntary edits to the database. I wanted to create a personal to for my job but I'm gonna look for an alternative since I'm not sure of what I'm doing and I don't want to risk to edit anything in the database.

pleerock:
There is a `readonly` option in column decorator: `@Column({ readonly: true })`

pleerock:
@ps2goat you able to provide a failing test for your issue?

marceloverdijk:
I have a similar use case. I would like to open a sqlite db in read only mode.

`new sqlite3.Database(filename, [mode], [callback])`
where mode = `sqlite3.OPEN_READONLY`

Is it possible to pass db specific options in ormconfig.json?

pleerock:
@marceloverdijk no, its not. But its quite easy feature to implement, so feel free to contribute. You only need to add option in `SqliteConnectionOptions` and pass it into the new sqlite3.Database in SqliteDriver

djgould:
The current tests for readonly seem broken. https://github.com/typeorm/typeorm/blob/6fcaa0f762807743fbf8c346193348b1c66147a9/test/functional/columns/readonly/columns-readonly.ts#L17-L38

in this case `authorName` is readonly but the test is expecting the value to change (despite what the comments say). I believe the last line should be `expect(loadedPost!.authorName).to.be.equal("Umed");`

I didn't actually run the tests myself but if they are passing it seems there is a problem with the readonly property

hielkehoeve:
I have created a PR for this issue: https://github.com/typeorm/typeorm/pull/2749

vlapo:
Lets track this as main issue for this problem. Just summary:
 - readonly is not working as expected - please see comments here https://github.com/typeorm/typeorm/pull/2150#issuecomment-389915199
 - as @ps2goat pointed there is small inconsistency in current read only definition and we should add new properties as `updatable` and `insertable`
 - our current read only test is wrong and have to be fixed

imnotjames:
Tests have been fixed in master and this feature seems to have been fixed now.

PR #2749, opened 2018-08-31
Fix column option readonly

hielkehoeve:
As described in issue #2041 and issue #2221 the readonly option is not working correctly.

I have fixed and updated de testcase to use `.save` and `.update` and fixed the code accordingly.

### DIFF
--- a/src/persistence/Subject.ts
+++ b/src/persistence/Subject.ts
@@ -193,7 +193,7 @@ export class Subject {
             if (this.metadata.isJunction && changeMap.column) {
                 valueMap = changeMap.column.createValueMap(changeMap.column.referencedColumn!.getEntityValue(value));
 
-            } else if (changeMap.column) {
+            } else if (changeMap.column && (!changeMap.column.isReadonly || !this.databaseEntity)) {
                 valueMap = changeMap.column.createValueMap(value);
 
             } else if (changeMap.relation) {
--- a/test/functional/columns/readonly/columns-readonly.ts
+++ b/test/functional/columns/readonly/columns-readonly.ts
@@ -34,7 +34,7 @@ describe("columns > readonly functionality", () => {
         const loadedPost = await postRepository.findOne(1);
         expect(loadedPost!.title).to.be.equal("About columns1");
         expect(loadedPost!.text).to.be.equal("Some text about columns1");
-        expect(loadedPost!.authorName).to.be.equal("Umed1");
+        expect(loadedPost!.authorName).to.be.equal("Umed");
 
     })));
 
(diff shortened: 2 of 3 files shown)


hielkehoeve:
Is this framework still alive?

hielkehoeve:
...

nightwolfz:
Looks like no one is around :(

nightwolfz:
Looks like no one is around :(

pleerock:
Sorry for a long response here.

We need to think how it should work now.

1. when we call `save` if it does insert it should only insert the entity, if it does update it should skip this entity column value

2. we have UpdateQueryBuilder that can execute any update query. Should we get in count readonly in there? From one side we can update an entity using update query builder, from another side update query builder should work like regular SQL and let you update anything.

Maybe for second we can disallow readonly columns updation, but add some method into UpdateQueryBuilder like `readOnlyColumns(boolean)` ?

adding @Kononnable

src/persistence/Subject.ts
```diff
@@ -193,7 +193,7 @@ export class Subject {
             if (this.metadata.isJunction && changeMap.column) {
                 valueMap = changeMap.column.createValueMap(changeMap.column.referencedColumn!.getEntityValue(value));
 
-            } else if (changeMap.column) {
+            } else if (changeMap.column && (!changeMap.column.isReadonly || !this.databaseEntity)) {
```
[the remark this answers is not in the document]
hielkehoeve: if: the column is not readonly OR the action is an insert action.

test/functional/columns/readonly/columns-readonly.ts
```diff
@@ -34,9 +34,20 @@ describe("columns > readonly functionality", () => {
...
         expect(loadedPost!.text).to.be.equal("Some text about columns1");
-        expect(loadedPost!.authorName).to.be.equal("Umed1");
+        expect(loadedPost!.authorName).to.be.equal("Umed");
 
-    })));
 
+        // then update all its properties and save again
+        post.title = "About columns1";
+        post.text = "Some text about columns1";
+        post.authorName = "Umed1";
```
Kononnable: I think this should throw an error. If property is readonly we should get an error on assignment. Just ignoring such situations may lead to bugs in app logic.

ejaustin: Any updates here? Reading `tests` code and saw this bug.
`expect(loadedPost!.authorName).to.be.equal("Umed1");`

I'd expect `post.authorName` NOT be updated and get an error or at least a consistent expected behavior.

imnotjames:
There's conflicts and tests now show this feature as working in master, going to close.

PR #2750, opened 2018-08-31
Update DriverFactory.ts

odykyi:
fix code style


pleerock:
we can't call any of those styles bad or good. So I dont see a point to change it. But anyway I'll merge it to add another contributor into typeorm's piggy bank haha

Issue #2756, opened 2018-09-01
The default column is not present after calling save


rianby64:
**Issue type:**

[x] question
[x] bug report
[ ] feature request
[ ] documentation issue

**Database system/driver:**

[ ] `cordova`
[ ] `mongodb`
[ ] `mssql`
[ ] `mysql` / `mariadb`
[ ] `oracle`
[x] `postgres`
[ ] `sqlite`
[ ] `sqljs`
[ ] `react-native`
[ ] `expo`

**TypeORM version:**

[x] `latest`
[ ] `@next`
[ ] `0.x.x` (or put your version here)

**Steps to reproduce or a small repository showing the problem:**
Given the following migration
```ts
export class Post1535782493710 implements MigrationInterface {

    public async up(queryRunner: QueryRunner): Promise<any> {
      await queryRunner.query(`CREATE SEQUENCE IF NOT EXISTS incr_by_db_seq;`);
      await queryRunner.query(`
        CREATE TABLE IF NOT EXISTS "post"
          (
              id serial NOT NULL,
              incr integer NOT NULL DEFAULT nextval('incr_by_db_seq'),
              title character varying(255),
              CONSTRAINT post_pkey PRIMARY KEY (id)
          );
      `);
    }

    public async down(queryRunner: QueryRunner): Promise<any> {
      await queryRunner.query(`DROP SEQUENCE IF EXISTS incr_by_db_seq;`);
      await queryRunner.query(`DROP TABLE IF EXISTS "post";`);
    }

}
```
The entity looks like
```ts

@Entity()
export class Post {

    @PrimaryGeneratedColumn()
    id: number;

    @Column()
    incr: number;

    @Column()
    title: string;

}
```

If perform a save call, then the result doesn't contain the column ```incr```
```ts
const post = new Post();
post.title = "Super title";
const result = await connection.manager.save(post);

expect(result).not.to.be.empty;
expect(result!.title).not.to.be.empty;
result!.title.should.be.equal("Super title");

expect(result!.incr_by_db, "incr field is present").not.to.be.empty;
expect(result!.incr_by_db).to.be.greaterThan(0);
```

rianby64:
If you configure the entity as follows
```ts
@Entity()
export class Post {

    @PrimaryGeneratedColumn()
    id: number;

    @Column({
        default: 'whatever'
    })
    incr: number;

    @Column()
    title: string;

}
```

then, the field ```incr``` will be present after calling ```save```.

[...]

PR #2757, opened 2018-09-01
test: the column that has a default definition

rianby64:
In order to expose this case, I modified the interface ```TestConnection``` in order to allow to run migrations.

After running the migrations, the following is to catch the case when a field that is defined to have a default value, must be present in the object returned by calling ```save```. That field should behave as behaves the field ```id```.

Related to #2756

### DIFF
--- a/test/github-issues/2756/entity/post.entity.ts
+++ b/test/github-issues/2756/entity/post.entity.ts
@@ -8,7 +8,11 @@ export class Post {
     @PrimaryGeneratedColumn()
     id: number;
 
-    @Column()
+    @Column({
+        name: 'incr',
+        nullable: true,
+        default: 10,
+    })
     incr: number;
 
     @Column()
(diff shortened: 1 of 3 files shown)


imnotjames:
Can this instead be done by calling the migration executor programmatically?  Instead of calling it on the configured connection?

rianby64:
Yeap... Let met think about it.

imnotjames:
You actually don't need to modify any of this - it's possible to use migrations in tests without changing anything

https://github.com/typeorm/typeorm/blob/2f275816375a58c9ad136d5f02dc3508797152c2/test/github-issues/4697/issue-4697.ts#L25

rianby64:
Hey, @imnotjames I moved the code to its correspondent directory.
Hope this change will reflect the error I mentioned in the issue #2757 .

rianby64:
@imnotjames finally I got the error running. Please, take a look.

imnotjames:
I switched up the `empty` check with a `defined` check and it passes.

Something's up with negating chai's `empty` check, I think.

imnotjames:
Tests passed with that.  Cannot replicate.

PR #2761, opened 2018-09-03
updated the repository.

pvaidhya:
updated the repository with different code examples like create , soft delete, get all data and get specific data by filtering e.c.t.


pleerock:
I dont think what you wrote here should part of documentation since it uses specific code style, specific libs, etc.

Thanks for your efforts.

Issue #2758, opened 2018-09-02
Insert fails when related OneToOne entity's primary key is also a foreign key 


yazshel:
**Issue type:**

[ ] question
[x] bug report
[ ] feature request
[ ] documentation issue

**Database system/driver:**

[ ] `cordova`
[ ] `mongodb`
[ ] `mssql`
[ ] `mysql` / `mariadb`
[ ] `oracle`
[x] `postgres`
[ ] `sqlite`
[ ] `sqljs`
[ ] `react-native`
[ ] `expo`

**TypeORM version:**

[x] `latest`
[ ] `@next`
[ ] `0.x.x` (or put your version here)

TypeORM fails to set the foreign key value on insert when the related entity's primary key is also part of a foreign key.

I have a three level model hierarchy which models user, person and party. Nested inserts  The relationships look like User -> Person -> Party, however the relationship between person and party is somewhat unusual - it's a `@OneToOne` relationship where the primary key of `Person` is also a foreign key to `Party`.

User` also has a `OneToOne` relationship to `Person` but with a normal foreign key.

With this arrangement, Creating a new `Person` with a nested `Party` works as expected, but attempting to create a `User` with nested `Person` and `Party` records fails to set the `User.personId` foreign key field to the inserted person's ID.

Further, when attempting to create a new `User` and providing an existing `Person` object as `User.person` value, the same issue occurs.

Please see the minimal example below:

`entity/entities.ts`:
```ts
import { Column, Entity, JoinColumn, OneToOne, PrimaryColumn, PrimaryGeneratedColumn } from 'typeorm';

@Entity()
export class Party {
  @PrimaryGeneratedColumn('uuid')
  id: string;
}

@Entity()
export class Person {

  // Party ID also acts as PK for Person (ie. inheritance)
  @PrimaryColumn('uuid')
  id: string;

  @OneToOne(() => Party, { cascade: true, onDelete: "CASCADE" })
  @JoinColumn({ name: 'id' })
  party: Party;
}


@Entity()
export class User {

  @PrimaryGeneratedColumn('uuid')
  id: string;

  @Column('uuid')
  personId: string;

  @OneToOne(() => Person, { cascade: true, onDelete: "CASCADE" })
  person: Person;
}
```

`index.ts`:
```ts
import 'reflect-metadata';

import { createConnection } from 'typeorm';

import { Person, User } from './entity/entities';

createConnection().then(async connection => {

    const personRepository = connection.getRepository(Person);
[...]

PR #2589, opened 2018-07-26
Github2588

csuich2:
Failing test case for #2588

Issue #2373, opened 2018-06-20
Primary Key column not included in returned object if primary key is is a foreign key of one-to-one relationship


dominic-simplan:
**Issue type:**

[x] question
[ ] bug report
[ ] feature request
[ ] documentation issue

**Database system/driver:**

[x] `cordova`
[ ] `mongodb`
[ ] `mssql`
[ ] `mysql` / `mariadb`
[ ] `oracle`
[ ] `postgres`
[ ] `sqlite`
[ ] `sqljs`
[ ] `react-native`

**TypeORM version:**

[x] `latest`
[ ] `@next`
[ ] `0.x.x` (or put your version here)

**Steps to reproduce or a small repository showing the problem:**

I have a table (ProjectSettings) whose primary key is also the foreign key of a one-to-one relation. My entities look like this:
```
@Entity()
export class Project {
    @PrimaryGeneratedColumn()
    id: number;
    @OneToOne(number => ProjectSettings, projectSettings => projectSettings.project, { onDelete: 'CASCADE', cascade: true })
    settings: ProjectSettings;
}
```
and
```
@Entity()
export class ProjectSettings {
    @JoinColumn()
    @OneToOne(type => Project, { primary: true })
    project: Project;
}
```

However, if I use the repository to find the ProjectSettings by the Project, the returned class does not contain the projectId:
```
let repository = getConnection().getRepository(ProjectSettings);
repository.findOne({ project: project }).then(projectSettings => { ...})
```
Thus, when trying to persist the ProjectSettings again, a new entry is created, as the projectId is unknown.
Actually I've tried many combinations of various annotation properties, however I am not able to get it working.

Could someone advise what I am doing wrong here?

pleerock:
> the returned class does not contain the projectId:

because your class does not contain projectId property.

To avoid this kind of issues we are deprecating this format of definition of primary foreign keys in favour of following format:

```ts
@Entity()
export class ProjectSettings {

    @PrimaryColumn()
    projectId: number;

    @OneToOne(type => Project)
    @JoinColumn()
    project: Project;
}
```

please use it and probably you'll get rid of "strange" issues.

dominic-simplan:
@pleerock thanks for the explanation! Saving and loading is working fine now, however I am getting an error when trying to delete the Project entity. (I want that the ProjectSettings are cascaded on insert, update and delete of the Project entity.)

```
@Entity('Project')
export class Project {
[...]

PR #2047, opened 2018-04-28
Tests for issues #1989 and #1993 

filipjnc:

PR #2765, opened 2018-09-04
Add tests for #2758

yazshel:
Tests for `@OneToOne` and `@ManyToOne` relations to an entity which has a foreign key as its primary key.

### DIFF
--- a/test/github-issues/2758/entity/Bar.ts
+++ b/test/github-issues/2758/entity/Bar.ts
@@ -1,0 +1,17 @@
+import { Column, Entity, JoinColumn, OneToOne, PrimaryColumn } from "../../../../src";
+import { Foo } from "./Foo";
+
+@Entity()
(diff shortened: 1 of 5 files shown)


hielkehoeve:
Your tests are not working. If this is caused by a bug in TypeORM please fix it in this PR.

yazshel:
Sorry I've just realised I mis-read the documentation about logging issues - I mistakenly understood that PRs with unit tests demonstrating the issue were desired.

I've done a fair bit of code tracing but haven't been able to track down why the relation's foreign key is not being set in these circumstances, so I'll close the PR for now.

Kononnable:
@yazshel There are many PRs with such tests(#2589, #2047, #2416, #2472). @pleerock asked many times for such PRs(#2373, #1989). If problem is in your code(not in typeorm) PR just will be rejected.

PR #2777, opened 2018-09-07
fix transform.

bendela6:
transformer.from does not works when column is null. I think it should.

### DIFF
--- a/src/driver/mysql/MysqlDriver.ts
+++ b/src/driver/mysql/MysqlDriver.ts
@@ -408,10 +408,9 @@ export class MysqlDriver implements Driver {
      * Prepares given value to a value to be persisted, based on its column type or metadata.
      */
     prepareHydratedValue(value: any, columnMetadata: ColumnMetadata): any {
-        if (value === null || value === undefined)
-            return value;
-
-        if (columnMetadata.type === Boolean) {
+        if (value === null || value === undefined) {
+            //DO NOTHING
+        } else if (columnMetadata.type === Boolean) {
             value = value ? true : false;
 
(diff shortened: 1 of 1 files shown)


afurculita:
@bendela6 could you please give us more info about your use case?

src/driver/mysql/MysqlDriver.ts
```diff
@@ -408,10 +408,9 @@ export class MysqlDriver implements Driver {
      * Prepares given value to a value to be persisted, based on its column type or metadata.
      */
     prepareHydratedValue(value: any, columnMetadata: ColumnMetadata): any {
-        if (value === null || value === undefined)
-            return value;
-
-        if (columnMetadata.type === Boolean) {
+        if (value === null || value === undefined) {
+            //DO NOTHING
```
odykyi: pls do not use comments

Kononnable:
>Create your patch, **including appropriate test cases**. Without tests your PR will not be accepted.

@bendela6 You need to add tests for this PR. Thanks to it we will know exactly what issue has been fixed by this PR and make sure it won't be broken again.

kikar:
We have the same issue. Our case is that we don't want to store a default value in the db, to keep the table lighter.
So with the `transformer.from` we could replace the `NULL` with our default value.

mcwebb:
This also affects us. Our use case is that a null value in a nullable Postgres column is assigned as `null` for the property on the entity, when we actually want the property to simply be undefined if the database value is null.

pleerock:
can somebody add a test, because believe me, this is going to break soon without a test. Test should be simple for this change.

PR #2796, opened 2018-09-13
Improve Repository.save() return type

cuchi:
When saving a new entity with `repository.save(fields)`, the current type definition fails to return to me the right fields types for the generated columns, for example:

```typescript
@Entity()
export default class User {
    @PrimaryGeneratedColumn('uuid') id: string
    @Column() name: string
    @Column({ nullable: true }) address?: string
}

// ...

const user = { name: 'John Doe', address: 'Somewhere' }
const createdUser = await this.repository.save(user)

console.log(createdUser.name) // okay 
console.log(createdUser.address) // okay
console.log(createdUser.id) // compilation error: Property 'id' does not exist on type ...
```
This seems to only make sense when `reload: false` is sent, which is not the case. 
This little change fixes that, while overriding the optional type when the values are present.

I'm not sure if this breaks some special case, maybe some maintainer could look into this :sweat_smile:

### DIFF
--- a/src/repository/Repository.ts
+++ b/src/repository/Repository.ts
@@ -123,13 +123,15 @@ export class Repository<Entity extends ObjectLiteral> {
      * Saves all given entities in the database.
      * If entities do not exist in the database then inserts, otherwise updates.
      */
-    save<T extends DeepPartial<Entity>>(entities: T[], options?: SaveOptions): Promise<T[]>;
+    save<T extends DeepPartial<Entity>>(entities: T[], options: SaveOptions & { reload: false }): Promise<T[]>;
+    save<T extends DeepPartial<Entity>>(entities: T[], options?: SaveOptions): Promise<(T & Entity)[]>;
 
     /**
      * Saves a given entity in the database.
      * If entity does not exist in the database then inserts, otherwise updates.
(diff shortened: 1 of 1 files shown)


pleerock:
Thanks for contribution!

by making it `& Entity` it means we add all entity properties which is not truth either. Not sure if we should merge it.

cuchi:
> by making it `& Entity` it means we add all entity properties which is not truth either. Not sure if we should merge it.

That's why I define those nullable properties as TypeScript optional types. See the prop `address` in my example.

With my change, the following is true, because `{ foo: type } & { foo?: type }` still resolves to `{ foo: type }`
```typescript
@Entity()
export default class User {
    @PrimaryGeneratedColumn('uuid') id: string
    @Column() name: string
    @Column({ nullable: true }) address?: string
    @Column({ nullable: true }) number?: number
}

// ...

const user = { name: 'John Doe', address: 'Somewhere' }
const createdUser = await this.repository.save(user)

createdUser.address // okay
createdUser.age // error, might be undefined

```

pleerock:
okay, lets merge it, and lets see how people react to this change. Thank you for the contribution!

PR #2801, opened 2018-09-14
Fix overriden embedded in STI

fwoelffel:
- Fix the embedded columns overriding in STI implementations
- Fix the embedded columns not being marked as nullable


elthrasher:
👍 Looks solid to me.

fwoelffel:
Let me know if there is something keeping us from merging this.

PR #2823, opened 2018-09-20
Enum value lack of ‘NO ACTION’

walter211:


pleerock:
Its NO ACTION by default I believe.

Issue #2584, opened 2018-07-25
Cascaded table data is dropped on column add during migration


dominic-simplan:
**Issue type:**

[x] question
[x] bug report
[ ] feature request
[ ] documentation issue

**Database system/driver:**

[x] `cordova`
[ ] `mongodb`
[ ] `mssql`
[ ] `mysql` / `mariadb`
[ ] `oracle`
[ ] `postgres`
[ ] `sqlite`
[ ] `sqljs`
[ ] `react-native`
[ ] `expo`

**TypeORM version:**

[x] `latest`
[ ] `@next`
[ ] `0.x.x` (or put your version here)

I have a tableA which has ON CASCADE DELETE foreign key contstraint on tableB.
Now I add a column to tableB during a migration, and all data of tableA is gone, as the migration drops and recreates tableB:

```
await queryRunner.addColumn("Project", new TableColumn({
            name: "CustomVersion",
            type: "varchar",
            isNullable: true
        }));

query:  CREATE TABLE "temporary_Project" (...)
query:  INSERT INTO "temporary_Project"...
query:  DROP TABLE "Project"
```

Shouldn't the foreign key constraints be handled/disabled before dropping the table? Or am I doing it wrong?

dominic-simplan:
Also seems like `await queryRunner.query("PRAGMA foreign_keys=OFF;");` as a workaround has no effect as  the [foreign_key is always set to on](https://github.com/typeorm/typeorm/blob/47e8f875c3500614d7a93fdb05164b2939034d9f/src/driver/cordova/CordovaDriver.ts#L85) when creating a connection (and a new connection is created on every `queryRunner.query`?)

dominic-simplan:
The workaround is to change the foreign_key before starting the migration:
```
await connection.query("PRAGMA foreign_keys=OFF;");
await connection.runMigrations();
await connection.query("PRAGMA foreign_keys=ON;");
```

The foreign_key cannot be changed in the migration itself as migrations are executed in a transaction and foreign_key cannot be changed within a transaction:

> It is not possible to enable or disable foreign key constraints in the middle of a multi-statement transaction (when SQLite is not in autocommit mode). Attempting to do so does not return an error; it simply has no effect. 
Source: https://www.sqlite.org/foreignkeys.html

pleerock:
Interesting issue...

[...]

PR #2826, opened 2018-09-20
check for referencing foreign keys when altering a table using sqlite

hielkehoeve:
As described in issue #2576 TypeORM did not take into account whether tables have data and foreign constraints to the given table.

I have created a testcase to illustrate the situation.


dominicdesu:
Looks like this would also fix #2584 !?

hielkehoeve:
...

pleerock:
looks like CI is failing?

hielkehoeve:
It is. However this is not in the testcase I introduced but in `schema builder > change check constraint`...

pleerock:
no, its not that hard. You can ask people for help in slack if you stuck somewhere. You have to do that because tests are failing because of you changes and you need to check what and why you have broke and fix that.

imnotjames:
I'm gonna close this as it's failing, has conflicts and has not been updared in a few years

Issue #2557, opened 2018-07-20
Simple Postgres timestamp transformer is munging the value before transformer.to()


yang:
**Issue type:**

[ ] question
[x] bug report
[ ] feature request
[ ] documentation issue

**Database system/driver:**

[ ] `cordova`
[ ] `mongodb`
[ ] `mssql`
[ ] `mysql` / `mariadb`
[ ] `oracle`
[x] `postgres`
[ ] `sqlite`
[ ] `sqljs`
[ ] `react-native`
[ ] `expo`

**TypeORM version:**

[x] `latest`
[ ] `@next`
[ ] `0.x.x` (or put your version here)

**Steps to reproduce or a small repository showing the problem:**

With the following simple test case trying to use luxon.DateTime as a timestamp column:

```ts
import {DateTime} from "luxon";

...

@Entity()
export class Dummy {
  @PrimaryGeneratedColumn()
  id: number;

  @Column('timestamp', {transformer: {from: (d: Date) => DateTime.fromJSDate(d), to: (d: DateTime) => d.toJSDate()}})
  date: DateTime;
}

...

createConnection().then(async connection => {
  const dummy = new Dummy();
  dummy.date = DateTime.local();
  connection.manager.save(dummy);
})
```

I get:

```
(node:73270) UnhandledPromiseRejectionWarning: QueryFailedError: invalid input syntax for type timestamp: "0NaN-NaN-NaNTNaN:NaN:NaN.NaN+NaN:NaN"
    at new QueryFailedError (/Users/yang/proj/lastmileai/web/src/error/QueryFailedError.ts:7:9)
    at Query.callback (/Users/yang/proj/lastmileai/web/src/driver/postgres/PostgresQueryRunner.ts:171:30)
    at Query.handleError (/Users/yang/proj/lastmileai/web/node_modules/pg/lib/query.js:143:17)
    at Connection.connectedErrorHandler (/Users/yang/proj/lastmileai/web/node_modules/pg/lib/client.js:132:26)
    at emitOne (events.js:116:13)
    at Connection.emit (events.js:211:7)
    at Socket.<anonymous> (/Users/yang/proj/lastmileai/web/node_modules/pg/lib/connection.js:123:12)
    at emitOne (events.js:116:13)
    at Socket.emit (events.js:211:7)
    at addChunk (_stream_readable.js:263:12)
```

pleerock:
what do you mean `munging the value`? Can you provide an example? Also please provide a code without any third-party library

[...]

Issue #2280, opened 2018-06-05
DateTime(luxon) infomation goes missing after insertion into sqlite


YLRong:
**Issue type:**

[ ] question
[x] bug report
[ ] feature request
[ ] documentation issue

**Database system/driver:**

[ ] `cordova`
[ ] `mongodb`
[ ] `mssql`
[ ] `mysql` / `mariadb`
[ ] `oracle`
[ ] `postgres`
[x] `sqlite`
[ ] `sqljs`
[ ] `react-native`

**TypeORM version:**

[ ] `latest`
[ ] `@next`
[x] `0.2.6`

**Steps to reproduce or a small repository showing the problem:**

```
import { createConnection, Entity, PrimaryGeneratedColumn, Column, getRepository } from 'typeorm'
import { ValueTransformer } from 'typeorm/decorator/options/ValueTransformer'
import { DateTime } from 'luxon'

export class DateTimeTransformer implements ValueTransformer {
  public from (value: string): DateTime {
    console.log(`FROM ${value}, ${DateTime.fromSQL(value).toISO()}`)
    return DateTime.fromSQL(value)
  }

  public to (value: DateTime): string {
    console.log(`TO ${value.toISO()}, ${value.toSQL()}`)
    return value.toSQL()
  }
}

@Entity()
export class User {
  @PrimaryGeneratedColumn()
  public id: number

  @Column()
  public firstName: string

  @Column()
  public lastName: string

  @Column()
  public age: number

  @Column({ type: String, nullable: true, transformer: new DateTimeTransformer() })
  public startDate: DateTime = DateTime.fromJSDate(new Date())
}

test('test basic typeorm with sqlite', async () => {
  await createConnection({
    type: 'sqlite',
    database: ':memory:',
    entities: [User],
    synchronize: true
  })
  console.log('DB connected')

  const repository = getRepository(User)
  const user = new User()
  user.firstName = 'Timber'
  user.lastName = 'Saw'
  user.age = 25
  user.startDate = DateTime.local(2018, 6, 5)
  await repository.save(user)

  const allUsers = await repository.find()
  console.log(allUsers)
  expect(allUsers.length).toBe(1)

})
```

Running the test code, I got the following output:
```
$ jest --config=jest.config.json TypeOrm
 PASS  dist/entities/tests/TypeOrm.test.js
  √ test basic typeorm with sqlite (324ms)

  console.log dist/entities/tests/TypeOrm.test.js:61
    DB connected

[...]

PR #2831, opened 2018-09-21
Fix deepMerge to keep object protoype

adriensamson:
In #2557 and maybe also in #2280, luxon DateTime prototype is lost during `save()` before applying the transformer. luxon then sees the DateTime as invalid.

I've added a test with a custom wrapping object to avoid a dependency on luxon only for tests.


pleerock:
Thank you!

PR #2832, opened 2018-09-22
Fix typo in closure table section of entities.md

asztal:
This now matches the example in http://typeorm.io/#/tree-entities (that one doesn't show `@TreeLevelColumn`, despite importing it, but I'm not sure if that's intentional or not).


pleerock:
Thank you!

PR #2840, opened 2018-09-24
fix #2502- TypeError: One of the sources for assign has an enumerable key on the prototype chain 

### LINKED ISSUE #2502: TypeError: One of the sources for assign has an enumerable key on the prototype chain
**Issue type:**

[ ] question
[X] bug report
[ ] feature request
[ ] documentation issue

**Database system/driver:**

[ ] `cordova`
[ ] `mongodb`
[ ] `mssql`
[ ] `mysql` / `mariadb`
[ ] `oracle`
[ ] `postgres`
[X] `sqlite`
[ ] `sqljs`
[ ] `react-native`
[ ] `expo`

**TypeORM version:**

[ ] `latest`
[ ] `@next`
[X] `0.2.7` (or put your version here)

I'm running under react-native which is pedantic about Object.assign.  It's tripping in [```Object.assign(this, alias)```](https://github.com/typeorm/typeorm/blob/2758aec2a15745ec653bb107677bf33e7007afdb/src/query-builder/Alias.ts#L23) on ```target``` and ```metadata```.

I get this error:
> TypeError: One of the sources for assign has an enumerable key on the prototype chain. Are you trying to assign a prototype property? We don't allow it, as this is an edge case that we do not support. This error is a performance optimization and not spec compliant.

arolson101:
fix for https://github.com/typeorm/typeorm/issues/2502

I created a separate function ```ObjectUtils.assign``` to use instead of ```Object.assign``` that will only copy the properties returned by ```Object.getOwnPropertyNames```.

### DIFF
--- a/src/connection/Connection.ts
+++ b/src/connection/Connection.ts
@@ -33,6 +33,7 @@ import {RelationIdLoader} from "../query-builder/RelationIdLoader";
 import {EntitySchema} from "../";
 import {SqlServerDriver} from "../driver/sqlserver/SqlServerDriver";
 import {MysqlDriver} from "../driver/mysql/MysqlDriver";
+import {ObjectUtils} from "../util/ObjectUtils";
 import {PromiseUtils} from "../";
 
 /**
@@ -178,7 +179,7 @@ export class Connection {
             await this.queryResultCache.connect();
 
         // set connected status for the current connection
-        Object.assign(this, { isConnected: true });
+        ObjectUtils.assign(this, { isConnected: true });
 
         try {
 
@@ -224,7 +225,7 @@ export class Connection {
         if (this.queryResultCache)
             await this.queryResultCache.disconnect();
 
-        Object.assign(this, { isConnected: false });
+        ObjectUtils.assign(this, { isConnected: false });
     }
 
     /**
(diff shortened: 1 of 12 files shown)


arolson101:
note I didn't see #2502; maybe that fix is enough.

hristostf:
> note I didn't see #2502; maybe that fix is enough.

How can i use that fix? I use version 0.2.7 and that issue still exists.

adhambadr:
@hristostf i also still struggle with a clean way to do it, for now I went to arolson101 branch "[with_build](https://github.com/arolson101/typeorm/tree/with_build)" and manually patched the build/package folder into my node_modules as a temp fix till the PR is merged. it def. breaks down on other dev's computers. But at least its a solid fix, thanks @arolson101

jerolimov:
FYI, added this info also to #2502 

I had the same issue in my react-native project and could fix it by adding the [babel polyfill](https://babeljs.io/docs/en/babel-polyfill).

1. Add it as dependency (it is a runtime dependency, so you should not add it as devDependency)
    * with npm (`npm install --save @babel/polyfill`)
    * or yarn (`yarn add @babel/polyfill`)
2. Load the polyfills by loading it before your app-code / typeorm was loaded. For example, add this import to the top of your `index.js`:
    * `import '@babel/polyfill'`

pleerock:
Interesting how it would work if we use object spread instead.. Thanks merged, will be released in 0.2.9

stonecold123:
@pleerock  @jerolimov  
I am not able to fix the issue using `@babel/polyfill`. Please release the 0.2.9 version soon.

pleerock:
You can already use it via `npm i typeorm@rc`

PR #2843, opened 2018-09-25
Fix TYPEORM_URL to actually work

jussikinnula:
Fix TYPEORM_URL to actually work (e.g. switch Object.assign( ... ) to…… actually update urlDriverOptions to options)


pleerock:
Thank you!

PR #2854, opened 2018-09-26
unbreak release: externalise nativescript

championswimmer:
Going into the future drivers should come from the userland
or other packages of @typeorm/driver-xxxx

But for now we need to unbreak the 0.2.x release cycle
and only NativeScript broke it

Signed-off-by: Arnav Gupta <arnav@codingblocks.com>


championswimmer:
@Kononnable true that. fixed

championswimmer:
@pleerock can we make a 0.2.8 release now ?

PR #2879, opened 2018-10-02
Github2588 fix

### LINKED ISSUE #2588: createQueryBuilder always eager loads all relations, even if a condition is specified
**Issue type:**

[ ] question
[x] bug report
[ ] feature request
[ ] documentation issue

**Database system/driver:**

[ ] `cordova`
[ ] `mongodb`
[ ] `mssql`
[ ] `mysql` / `mariadb`
[ ] `oracle`
[x] `postgres`
[ ] `sqlite`
[ ] `sqljs`
[ ] `react-native`
[ ] `expo`

**TypeORM version:**

[ ] `latest`
[x] `@next`

A recent change in `next` that added `disableEagerRelations()` to `SelectQueryBuilder` appears to have broken `leftJoinAndSelect` conditions. From what I can tell, if eager relations are enabled, `SelectQueryBuilder` adds left joins for __all__ relations of an entity, even if one is already specified with a condition. Because of this, all relations are fully joined and added to the entity, rather than using the condition.

Here is the select query output from the new test case PR I've opened:

SELECT *
FROM "post" "post"
LEFT JOIN "post_review" "post_review" ON post_review."postId" = "post"."id" AND "post_review"."rating" >= 3
LEFT JOIN "post_review" "post_reviews" ON "post_reviews"."postId"="post"."id" WHERE "post"."id" = $1

The first left join is the one I've added with my custom condition. The second left join was __not__ added by me.

Link to PR that reproduces issue:

csuich2:
I've attempted to fix #2588.

My approach is to check for existing joins in the query and to not add a leftJoinAndSelect for any relations that are already specified.

### DIFF
--- a/src/query-builder/SelectQueryBuilder.ts
+++ b/src/query-builder/SelectQueryBuilder.ts
@@ -1853,7 +1853,9 @@ export class SelectQueryBuilder<Entity> extends QueryBuilder<Entity> implements 
             if (this.expressionMap.eagerRelations === true) {
                 if (!this.findOptions.options || this.findOptions.options.eagerRelations !== false) {
                     // Create a list of all of the alias+propertyPaths that were manually joined, so we don't join them again
-                    const manuallyJoinedRelations = this.expressionMap.joinAttributes.filter(join => join.relationPropertyPath).map(join => join.parentAlias + "." + join.relationPropertyPath);
+                    const manuallyJoinedRelations = this.expressionMap.joinAttributes
+                                                        .filter(join => join.relationPropertyPath)
+                                                        .map(join => join.parentAlias + "." + join.relationPropertyPath);
                     const joinEagerRelations = (alias: string, metadata: EntityMetadata) => {
                         metadata.eagerRelations.forEach(relation => {
                             const relationAlias = alias + "_" + relation.propertyPath.replace(".", "_");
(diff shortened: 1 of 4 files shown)


pleerock:
Looks fine, @Kononnable can you please review?

src/query-builder/SelectQueryBuilder.ts
```diff
@@ -1852,10 +1852,16 @@ export class SelectQueryBuilder<Entity> extends QueryBuilder<Entity> implements
 
             if (this.expressionMap.eagerRelations === true) {
                 if (!this.findOptions.options || this.findOptions.options.eagerRelations !== false) {
+                    // Create a list of all of the alias+propertyPaths that were manually joined, so we don't join them again
+                    const manuallyJoinedRelations = this.expressionMap.joinAttributes.filter(join => join.relationPropertyPath).map(join => join.parentAlias + "." + join.relationPropertyPath);
```
Kononnable: Line is pretty long. Maybe it would look better written like that:
```typescript
    const manuallyJoinedRelations = this.expressionMap.joinAttributes
                                        .filter(join => join.relationPropertyPath)
                                        .map(join => join.parentAlias + "." + join.relationPropertyPath);
```

test/github-issues/2588/issue-2588.ts
```diff
@@ -0,0 +1,47 @@
...
+import {expect} from "chai";
+
+import {Post} from "./entity/Post";
+import {PostReview} from "./entity/PostReview";
+
+describe("github issues > #2588 - createQueryBuilder always does left joins on relations", () => {
+    let connections: Connection[];
+    before(async () => connections = await createTestingConnections({
+        entities: [__dirname + "/entity/*{.js,.ts}"],
+        enabledDrivers: ["postgres"],
```
Kononnable: Bug was reported for postgres, but doesn't it occur on every database using SelectQueryBuilder? It doesn't look like database engine related. If that's the case we should probably test it on every driver.

csuich2:
@pleerock @Kononnable Any other thoughts on this?

PR #2889, opened 2018-10-03
docs: Fix separated bulleted list

nicogreenarry:
A linebreak was making a single bulleted list look like two separate lists.


pleerock:
Thank you very much!

PR #2891, opened 2018-10-04
Fix check version command

mabc224:


pleerock:
There is a `typeorm version` command that does a bit different than node's `--version`.

```
$ typeorm version
No local installed TypeORM was found.
Global installed TypeORM version: 0.2.0-alpha.46
```

Thanks for your efforts!

Issue #2381, opened 2018-06-21
Hydration performance issue on complex dataset


leoperria:
**Issue type:**

[X] question
[ ] bug report
[ ] feature request
[ ] documentation issue

**Database system/driver:**

[ ] `cordova`
[ ] `mongodb`
[ ] `mssql`
[ ] `mysql` / `mariadb`
[ ] `oracle`
[X] `postgres`
[ ] `sqlite`
[ ] `sqljs`
[ ] `react-native`

**TypeORM version:**

[ ] `latest`
[ ] `@next`
[X] `0.2.6` 

**Steps to reproduce or a small repository showing the problem:*

Hi, we have a performance issue in hydration for a (quite) complex query like this:

```ts
const ids = [655, 8888, 9999, 12324, 1111, 78, 7, 6, 5, 4, 3, 2, 1, 15, 14, 13, 12, 11, 10, 9];
const res = await this._entityRepository
    .createQueryBuilder("a");
    .leftJoinAndSelect(`a.table2`, "am")
    .leftJoinAndSelect(`a.table1`, "o")
    .leftJoinAndSelect(`am.table3`, "m")
    .leftJoinAndSelect(`a.table4List`, "e")
    .leftJoinAndSelect(`e.table5`, "em")
    .leftJoin(`e.table6`, "el")
    .addSelect("el.id")
    .leftJoinAndSelect(`em.table7`, "emm")
    .leftJoin(`a.table8`, "ap")
    .addSelect("ap.id")
    .leftJoin(`a.table9`, "as")
    .addSelect("as.id")
    .leftJoin(`a.table10`, "l")
    .addSelect("l.id")
    .leftJoin(`a.table11`, "s")
    .addSelect("s.id")
    .leftJoin(`a.table12`, "pc")
    .addSelect("pc.id")
    .where(`a.id IN (${ids.join(",")})`)
    .getMany();
```

This generates a SQL query like the following:

```sql

select
    "a"."id" as "a_id",
	"a"."f1" as "a_f1",
	"a"."f2" as "a_f2",
	"a"."f3" as "a_f3",
	"a"."f4" as "a_f4",
	"a"."f5" as "a_f5",
	"a"."f6" as "a_f6",
	"a"."f7" as "a_f7",
	"a"."f8" as "a_f8",
	"a"."f9" as "a_f9",
	"a"."f10" as "a_f10",
	"a"."f11" as "a_f11",
	"a"."f12" as "a_f12",
[...]

PR #2894, opened 2018-10-04
Performance of hydration

Kononnable:
#2381 
I don't like the way I implemented it but I don't have a concept how to do it better. Decorators won't be of help here because they work when object is being created. Using generic class just for caching would require changing how it is used  - we can't have a getter to property of type Cache<T> which returns type T, so instead of `obj.relation` we would have to use `obj.relation.value` or something similar.

I hope community will come up with some nicer solution. On problem reproduction repo https://github.com/stelltec/typeorm_testcase1 cached version works ~15x faster with 10240 rows returned.

### DIFF
--- a/src/query-builder/JoinAttribute.ts
+++ b/src/query-builder/JoinAttribute.ts
@@ -68,19 +68,30 @@ export class JoinAttribute {
         return false;
     }
 
+    
+    isSelectedCache: boolean;
+    isSelectedEvalueated: boolean = false;
     /**
      * Indicates if this join is selected.
      */
     get isSelected(): boolean {
-        for (const select of this.queryExpressionMap.selects) {
-            if (select.selection === this.alias.name)
-                return true;
+        if (!this.isSelectedEvalueated) {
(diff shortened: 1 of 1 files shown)


dmitrirussu:
@Kononnable 
https://nodejs.org/docs/latest-v8.x/api/addons.html
https://github.com/nlohmann/json

Kononnable:
@dmitrirussu please add few words and don't force everyone to ask themselves what you could have in mind :smiley: 
If I guessed correctly you suggest to serialize json with native api using n-api, but the problem with performance wasn't with serializing the object but with creating one.

dmitrirussu:
@Kononnable, I suggest to be done hydration on cpp module I think node api provide that posibility...   the  json library could help to parse any raw colllection from js... etc... there could be diff solution for that... the main ideia is to use c++ and node api for that...

pleerock:
Im okay with this fix. Better solution is to refactor QueryBuilder, there are lot of room for optimization. Hope we do it some day.

PR #2895, opened 2018-10-04
Fix andWhere and orWhere signatures

davewasmer:
`where()`, `andWhere()` and `orWhere` all delegate to `computeWhereParameter()`, but only `where()` allows object literals as the condition to evaluate. This fixes the signatures of `andWhere()` and `orWhere()` to also allow object literal conditions.


pleerock:
Thanks for the contribution!

I think this won't work properly. Is it working if we do `.where({ findOptions1 }).andWhere({ findOptions2 }).orWhere({ findOptions3 })` ? We need tests that confirm that its all working.

pleerock:
Closing as there is no response from author for now. Lets open it once it gets some response.

PR #2898, opened 2018-10-05
Support specifying full config file path in cli using -f option

### LINKED ISSUE #2339: confusing command line
**Issue type:**

[ ] question
[ ] bug report
[x ] feature request
[ x] documentation issue

**Database system/driver:**

[ ] `cordova`
[ ] `mongodb`
[ ] `mssql`
[x ] `mysql` / `mariadb`
[ ] `oracle`
[ ] `postgres`
[ ] `sqlite`
[ ] `sqljs`
[ ] `react-native`

**TypeORM version:**

[ x] `latest`
[ ] `@next`

I would actually vote to change this behavior but at least should be spelled out more clearly. When using the CLI, you can use the -f command to specify an alternate configuration file.

The normal expectation would be a complete filename, such as

ts-node ./node_modules/typeorm/cli.js schema:sync -c test-connection -f ormconfig-local.yml

However, the system is set up to expect no extension on the file (it rotates and checks for all possibilities), and including the extension gives the error "Error: No connection options were found in any of configurations file."

I would propose that manually specifying a file would normally include the extension and that should be allowed to work with TypeORM.

<!--
    [x] `mysql`
    [ ] `postgres`

!>

steevsachs:
Support specification of a full file path with extension using the `-f` option in the cli.

Omitting a file extension will retain the current behavior of looping over possible file extensions for support.

I also updated `normalizeConnectionOptions` to use `baseDirectory` since this seemed more correct, but please let me know if that's not desirable.

Resolves #2339

### DIFF
--- a/src/connection/ConnectionOptionsReader.ts
+++ b/src/connection/ConnectionOptionsReader.ts
@@ -83,7 +83,7 @@ export class ConnectionOptionsReader {
         const fileFormats = ["env", "js", "ts", "json", "yml", "yaml", "xml"];
 
         // Detect if baseFilePath contains file extension
-        const possibleExtension = this.baseFilePath.substr(this.baseFilePath.indexOf("."));
+        const possibleExtension = this.baseFilePath.substr(this.baseFilePath.lastIndexOf("."));
         const fileExtension = fileFormats.find(extension => `.${extension}` === possibleExtension);
 
         // try to find any of following configuration formats
--- a/src/connection/ConnectionOptionsReader.ts
+++ b/src/connection/ConnectionOptionsReader.ts
@@ -78,9 +78,16 @@ export class ConnectionOptionsReader {
      * todo: get in count NODE_ENV somehow
      */
     protected async load(): Promise<ConnectionOptions[]|undefined> {
+        let connectionOptions: ConnectionOptions|ConnectionOptions[]|undefined = undefined;
+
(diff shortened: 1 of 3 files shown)


src/connection/ConnectionOptionsReader.ts
```diff
@@ -78,9 +78,16 @@ export class ConnectionOptionsReader {
      * todo: get in count NODE_ENV somehow
      */
     protected async load(): Promise<ConnectionOptions[]|undefined> {
+        let connectionOptions: ConnectionOptions|ConnectionOptions[]|undefined = undefined;
+
+        const fileFormats = ["env", "js", "ts", "json", "yml", "yaml", "xml"];
+
+        // Detect if baseFilePath contains file extension
+        const possibleExtension = this.baseFilePath.substr(this.baseFilePath.indexOf("."));
```
pleerock: its not reliable way of getting a file extension

steevsachs: Good catch, thank you.  Would you consider `lastIndexOf` reliable?

pleerock: I was thinking you'll use some method that gets you file extension... Or yeah `lastIndexOf` at least.

pleerock:
okay I added lastIndexOf for now. lets merge it.

indera-shsp:
Is there any documentation describing this change?

PR #2899, opened 2018-10-05
Removed .only from tests causing only 20 tests to run

csuich2:
@pleerock @Kononnable It appears that only 20 tests are being run on the next branch build.

Take a look at the output of the vNext build vs master build.

vNext build: https://travis-ci.org/typeorm/typeorm/jobs/437460875
master build: https://travis-ci.org/typeorm/typeorm/jobs/437717744

It appears that some .only statements are causing a subset of the tests to run. In this PR I've removed those and you can see that more than 20 tests are now running:
https://travis-ci.org/csuich2/typeorm/jobs/437822569

### DIFF
--- a/test/functional/observers/observe-count/observers-count-insert.ts
+++ b/test/functional/observers/observe-count/observers-count-insert.ts
@@ -4,7 +4,7 @@ import {Connection} from "../../../../src";
 import {closeTestingConnections, createTestingConnections, reloadTestingDatabases} from "../../../utils/test-utils";
 import {Post} from "./entity/Post";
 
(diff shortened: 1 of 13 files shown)


Kononnable:
We should also fix broken tests - running tests locally it looks like 20 were broken in the meantime(mostly by breaking changes I think).
But broken tests shoulldn't block this PR because they're not introduced by it.

csuich2:
@pleerock I hate to be the bearer of bad news, but should this hold up the release if there are tests broken due to regressions?

csuich2:
Nevermind! I now realize the release people are asking for is from master, not next.

PR #2909, opened 2018-10-09
Added info & package.json script useful for devs

csuich2:
I figured I'd throw this out there since it has been very useful for me as I've started to contribute to this project.

```tsc -w``` has helped avoid long compilations
```npm run test-no-compile``` skips the ```tsc``` step when I'm using ```tsc -w```

I'm happy to chat about these if you'd like and my feelings won't be hurt if you don't want to accept them, but I figured I'd share since they have been very helpful to me.


pleerock:
No that seems fine, thank you!

Issue #2311, opened 2018-06-08
Disable default OUTPUT selects for mssql 


arnas:
**Issue type:**

[ ] question
[ ] bug report
[x] feature request
[ ] documentation issue

**Database system/driver:**

[ ] `cordova`
[ ] `mongodb`
[x] `mssql`
[ ] `mysql` / `mariadb`
[ ] `oracle`
[ ] `postgres`
[ ] `sqlite`
[ ] `sqljs`
[ ] `react-native`

**TypeORM version:**

[x] `latest`
[ ] `@next`
[ ] `0.x.x` (or put your version here)

Hello,
in the application I am developing I am required to use triggers ( more precise azure data sync that adds triggers on selected database tables). 
But as table has triggers queries fails if it has OUTPUT clause: "DML statement cannot have any enabled triggers if the statement contains an OUTPUT clause without INTO clause.". And it seems that this clause are added on default on every repository save, update, insert if entity has properties with CreateDateColumn or UpdateDateColumn decorators ( even select: false doesn't affect this).
I found I way to disable it in the driver configuration by changing function `        this.connection.driver.isReturningSqlSupported = () => false;`. But then typeorm does many one by one select after initial operation even if don't need any results. 
Maybe it would be good to have additional setting to disable output altogether or add to specif columns that I don't want this default behavior?

pleerock:
output isn't reasonless there. If we add option to disable them then some functionality like CreateDateColumn and UpdateDateColumn won't work completely properly. Maybe in your very specific case there is another solution? Of if there is no other solutions maybe its better not to use CreateDateColumn and UpdateDateColumn decorators in your case?

arnas:
Could you explain why output is required for CreateDateColumn ? As far I know it just adds Current_timestamp to a createdOn column ?

pleerock:
because `save` method updates original entity you pass - updates all properties with DEFAULT set, so you have data in your model reflected from the database.

arnas:
I see. So yeah global change could result in some weird behavior. But maybe there could be additional option for these column that are enabled be default?

pleerock:
All additional options gonna look weird as well. Maybe its better to get know why mssql does not allow using output in your cases and how this situation can be handled?

arnas:
Yeah I agree about that adding additional options can be an unnecessary overhead.
[...]

PR #2911, opened 2018-10-09
Add flag to disable output for sql server

arnas:
I have add this flag, though I am having problems running tests in my current dev environment and haven't added them, though as I seen other options also doesn't have test.

https://github.com/typeorm/typeorm/issues/2311


pleerock:
I added semicolons

Issue #1686, opened 2018-03-01
Migration generate drops/creates all constraints


ProTip:
**Issue type:**

[ ] question
[x ] bug report
[ ] feature request
[ ] documentation issue

**Database system/driver:**

[ ] `cordova`
[ ] `mongodb`
[ ] `mssql`
[ ] `mysql` / `mariadb`
[ ] `oracle`
[x] `postgres`
[ ] `sqlite`
[ ] `sqljs`
[ ] `websql`

**TypeORM version:**

[ ] `latest`
[x] `@next`
[ ] `0.x.x` (or put your version here)

On an existing database in PostgreSQL 9.6.2 migration generation wants to drop all unique constraints and create them again. I ran this migration however it wants to do it on the next generate as well. The only thing I can tell(without digging into the code) is that the order in ```down``` is backwards on the columns in the SQL statement..

pleerock:
> On an existing database in PostgreSQL 9.6.2

what is existing database? Database generate from older typeorm version? If yes, then its okay that it drops them since there is a new foreign keys naming strategy in latest next.

ProTip:
Database tables were created outside of TypeORM initially. I have let it drop and re-add the constraints multiple times but it always generates the commands again in the migrations as of ```0.2.0-alpha.27```.

I will dive into the code soon and see if I can track down the issue; will open a PR if it's a straightforward fix.

pleerock:
no, its okay that typeorm drops and re-create exist keys. To make a proper comparison of exist keys TypeORM uses its own naming strategy for them. So it simply checks that it does not have fks and there are outdated fks - it removes old one a create new ones. Its expected behaviour.

ProTip:
Currently it does it for every migration. I've allowed it to drop and recreate the unique constraints, however it wants to do it again for *every* migration.

So, currently, every migration I generate is cluttered with drop/add constraint commands even though I keep letting it run them.

pleerock:
hmm I need a reproduction example. Can you reproduce this issue without pre-created database?

pleerock:
Closing as no answer from reporter.

asvetliakov:
Got the same problem,
Repro:
Entities:

User.ts
```ts
import { PrimaryGeneratedColumn, Column, Entity, PrimaryColumn, OneToOne } from "typeorm";
import { Profile } from "./Profile";

@Entity()
export class User {
    @PrimaryColumn("varchar", { length: 255 })
    public id: string;


[...]

PR #2918, opened 2018-10-11
Fix duplicate unique index removal in mysql

adriensamson:
With https://github.com/typeorm/typeorm/issues/1686#issuecomment-382204047 at some point, you get 3 indices on uuid : `FK_`, `REL_` and `IDX_`. `IDX_` is a duplicate of `REL_` and should be removed.

When the schema builder tries to remove `IDX_`, it resets `tableColumn.isUnique` to false even if the `REL_` index keeps it unique. The column is then seen as changed and thus foreign key is dropped and recreated and `IDX_` is also recreated.

I've changed `Table.removeIndex()` to check for other unique indices.

I'm not sure how to add tests.
And `Table.removeUniqueConstraint()` should maybe be changed the same way for other dbms.


pleerock:
Thank you!

PR #2922, opened 2018-10-12
QueryBuilder getCount() fails when counting projected dataset

csuich2:
I run into this issue where I was attempting to do some aggregation in a subquery, then count the results based on a where condition. I found that QueryBuilder getCount() was requiring that the alias have metadata, which mine did not.

My solution is to check if the entity *has* metadata, and if not, use a simple `COUNT(*) as "cnt"` select.


pleerock:
The `getCount` method suppose to work with entities, not raw queries. Not sure if we should merge it. 

Also it seems tests are failing, we should fix them before merging such big changes.

Kononnable:
Closing - PR looks abandoned, tests are failing and we were not sure if it should be merged in the first place.

Issue #2578, opened 2018-07-24
ActiveRecord model is returning empty results everytime


rangelfinal:
**Issue type:**
[X] bug report

**Database system/driver:**
[X] `postgres`

**TypeORM version:**
[X] `latest` (0.2.7)

**Problem**
ActiveRecord model is returning empty update results even when it's updating stuff in the database.

**Steps to reproduce or a small repository showing the problem:**
```
export default class OAuthAccessToken extends BaseEntity {
@Column() id: string;

@Column({ nullable: false })
  expires: Date;

public static async revoke(conditions: Partial<OAuthAccessToken>) {
    const now = new Date();
    return this.update({ ...conditions, expires: MoreThan(now) as DeepPartial<Date> }, { expires: now });
  }
}
```
```
const updateResult = await token.revoke({id: 'abc'}); // Token with id abc was updated in the database
console.log(updateResult); // UpdateResult { generatedMaps: [], raw: [] }
```

<!--
    To answer those questions you need to put "x" inside the square brackets, for example:
    [x] `mysql`
    [ ] `postgres`

    Also, please format your code properly (by taking code blocks into ```ts .... ```)
!>

pleerock:
UpdateResult is somewhat system. What exactly do you expect `update` method to return?

rangelfinal:
Any kind of indication that the update was successful.  
The full entities, their IDs or just a integer with the number of rows affected.  
Even a boolean would do, but a indication of the number of rows affected would be better.

rangelfinal:
I will close this issue and follow the feature request, then.  
For future reference, the feature request is the issue #1308.

Issue #2415, opened 2018-06-27
DeleteResult / UpdateResult not implemented for all Drivers


marko-asplund:
**Issue type:**

[ ] question
[x] bug report
[ ] feature request
[ ] documentation issue

**Database system/driver:**

[ ] `cordova`
[ ] `mongodb`
[ ] `mssql`
[ ] `mysql` / `mariadb`
[ ] `oracle`
[x] `postgres`
[ ] `sqlite`
[ ] `sqljs`
[ ] `react-native`

**TypeORM version:**

[ ] `latest`
[ ] `@next`
[x] `0.2.7` (or put your version here)

**Steps to reproduce or a small repository showing the problem:**
1. invoke `Repository.delete`
2. the returned `DeleteResult` does not include any info at all. Expecting to include the number of affected rows.

I'm having trouble with `Repository.delete`: the operation successfully removes the entity from the database, but `DeleteResult` does not contain count of entities deleted, it's just `DeleteResult { raw: [] }`. This occurs with v0.2.7.

Also, there doesn't appear to be a tag for version 0.2.7 on GitHub though there's a release on npmjs. Why is that?

Why doesn't `DeleteResult` specify a property for the number of rows affected?

pleerock:
> Also, there doesn't appear to be a tag for version 0.2.7 on GitHub though there's a release on npmjs. Why is that?

I'll fix that.

> Why doesn't DeleteResult specify a property for the number of rows affected?

because it wasn't implemented. We already have this feature reported, but noone still took on its implementation. Please feel free, its an easy feature to implement.

marko-asplund:
@pleerock Can you give a reference to an earlier report? I just tried, but was unable to find one.

pleerock:
I could find only #1308 which is related, need to search deeper to find that one. But anyway, you don't need that issue if you want to implement it. You know what your goal, you can easily do that. But please be sure to implement it for all drivers if you are going to do it. That's easy feature to implement

thibaud-sanchez:
I have the same problem with SQLite and 0.2.8
It returns : `{"raw": []}`

pleerock:
`raw` is just raw result returned by underlying database driver. As I pointed above we need to implement this feature request.

HardySoftware:
MongoDb does not return anything, neither raw nor affected regardless if the record is found/deleted or not.

akosasante:
In case it helps anyone else in the future, for postgres I ended up just using 
```typescript
[...]

Issue #1308, opened 2017-12-06
Raw Postgresql Update query result is always an empty array


barczaG:
The query result of a raw postgresql update should be `UPDATE n` where `n` is the number of rows but it always returns an empty array.

barczaG:
@pleerock If you give me some hint I can start developing/fixing it myself.

pleerock:
it will be implemented in `0.2.0` (`npm i typeorm@next`) , `next` branch. We already have an `UpdateResult` object returned by query builder which we can use and add `affectedRows` or something like that property into it. You can investigate and implement this feature on your own, this will be a very nice contribution

kodayashi:
@pleerock - Is this something community needs to do? I see it was pulled from `0.2.0`.

pleerock:
@kodayashi yes I marked this since it can be implemented by a community. 0.2.0 has a `UpdateResult` object we just need to populate needed info for each database after update operation.

barczaG:
So the main issue here is that we return `result.rows` instead of `result` here: https://github.com/typeorm/typeorm/blob/master/src/driver/postgres/PostgresQueryRunner.ts#L173
By returning the `result` object we could access the `rowCount` property - https://node-postgres.com/api/result#-code-result-rowcount-int-code- then which could be added to `UpdateResult` object

I ended up using a fork because wasn't sure If I would break something by changing the return value of the `query` function. 

I can fix it if you think it's safe to alter the return value

pleerock:
Probably we'll need to refactor things a bit... We can create a `QueryResult` and let `query` method to return that, and make sure to update all query method usages. Then we can map QueryResult into UpdateResult and InsertResult where needed. This change should be in the `next` branch since its a breaking change.

kodayashi:
@barczaG - Happy to test this out in `next` if you need.

shaunc:
@pleerock - ... also DeleteResult (?)

PachowStudios:
Any update on this?

mkonicek:
Hi, just to double check - the fix for this hasn't landed, correct?

I'm using typeorm `0.2.16` and just did `console.log(updateResult)`.

I get an empty result:

```
UpdateResult { generatedMaps: [], raw: [] }
```

saarw:
Using a fork with this fix/workaround, but the rows end up at position raw[1], I'm guessing a real fix would have an object with an affectedRows field?
[...]

PR #2944, opened 2018-10-17
postgres: return whole result on updates and deletes

edevil:
Currently on updates and deletes nothing useful is returned to the caller. By returning the whole result from the postgres driver we can know how many rows were altered, etc.

Addresses: #2415, #1308, #2578


edevil:
This doesn't seem to be possible without breaking changes.

adrianhara:
May I ask why this issue was closed? Currently the updates return a result object. Is it not possible to augment that one instead of changing it completely?

PR #2957, opened 2018-10-21
Added Adminer to compose

gintsgints:
- Added adminer image for fast/simple database tool.
- Added bit instructions on README.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -68,6 +68,11 @@ services:
     ports:
       - "27017:27017"
 
+  adminer:
+    image: adminer
+    restart: always
+    ports:
+      - 8080:8080
   # redis
   # redis:
   #   image: "redis:3.0.3"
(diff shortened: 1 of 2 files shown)


docker-compose.yml
```diff
@@ -68,6 +68,11 @@ services:
     ports:
       - "27017:27017"
 
+  adminer:
+    image: adminer
+    restart: always
+    ports:
+      - 8080:8080
```
pleerock: we need to use different port something that isn't busy for users for sure

pleerock:
okay I checked adminer and tbh I don't like it. Isn't there modern alternatives? If there isn't I think lets just let users to use what they wanna use and do not include some specific tools in docker compose.

gintsgints:
Ok. I remove adminer then.

pleerock:
sorry about that 😞 Thanks for the contribution!

Issue #2965, opened 2018-10-22
Reuse preloaded lazy relations


StrikeForceZero:
**Issue type:**

[ ] question
[ ] bug report
[x] feature request
[ ] documentation issue

**Database system/driver:**

[x] `all?`
[ ] `cordova`
[ ] `mongodb`
[ ] `mssql`
[ ] `mysql` / `mariadb`
[ ] `oracle`
[ ] `postgres`
[ ] `sqlite`
[ ] `sqljs`
[ ] `react-native`
[ ] `expo`

**TypeORM version:**

[x ] `latest`
[ ] `@next`
[ ] `0.x.x` (or put your version here)

**Steps to reproduce or a small repository showing the problem:**

```typescript
import { Note } from './note';

@TypeOrm.Entity()
export class Person {
    @TypeOrm.PrimaryGeneratedColumn()
    public id: number;

    @TypeOrm.Column()
    public name: string;

    @TypeOrm.OneToMany(type => Note, note => note.owner, { lazy: true })
    public notes: Promise<Note[]> | Note[];
}
```

```typescript
import { Person } from './person';

@TypeOrm.Entity()
export class Note {
    @TypeOrm.PrimaryGeneratedColumn()
    public id: number;

    @TypeOrm.Column()
    public label: string;

    @TypeOrm.ManyToOne(type => Person, { lazy: true })
    public owner: Promise<Person> | Person;
}
```

```typescript
const repoPerson = connection.getRepository(Person);
const repoNote = connection.getRepository(Note);

const personA  = await repoPerson.create({ name: 'personA' });
const personB    = await repoPerson.create({ name: 'personB' });

await repoPerson.save([
        personA,
        personB,
])

await repoNote.insert({ label: 'note1', owner: personA });
await repoNote.insert({ label: 'note2', owner: personB });

const res = await repoPerson.find({ relations: [ 'notes' ] });
console.log(res);
console.log(await res[ 0 ].notes);
```
results:
```
query: SELECT "Person"."id" AS "Person_id", "Person"."name" AS "Person_name", "Person_notes"."id" AS "Person_notes_id", "Person_notes"."label" AS "Person_notes_label", "Person_notes"."ownerId" AS "Person_notes_ownerId" FROM "person" "Person" LEFT JOIN "note" "Person_notes" ON "Person_notes"."ownerId"="Person"."id"
[ Person { id: 1, name: 'personA', __notes__: [ [Object] ] },
[...]

PR #2902, opened 2018-10-08
Lazy relation promise persistence

yazshel:
await and persist resolved value of Promise instances on lazy relation properties (closes #2729)

PR #2966, opened 2018-10-22
Reuse preloaded lazy relations

StrikeForceZero:
Resolves https://github.com/typeorm/typeorm/issues/2965

### DIFF
--- a/test/github-issues/2965/entity/person.ts
+++ b/test/github-issues/2965/entity/person.ts
@@ -1,0 +1,14 @@
+import { Entity, Column, OneToMany, PrimaryGeneratedColumn } from '../../../../src/';
+import { Note } from './note';
+
+@Entity()
+export class Person {
+    @PrimaryGeneratedColumn()
+    public id: number;
+
+    @Column()
+    public name: string;
+
+    @OneToMany(type => Note, note => note.owner, { lazy: true })
+    public notes: Promise<Note[]> | Note[];
+}
--- a/test/github-issues/2965/index.ts
+++ b/test/github-issues/2965/index.ts
@@ -1,0 +1,58 @@
+import "reflect-metadata";
+
+import {closeTestingConnections, createTestingConnections, reloadTestingDatabases} from "../../utils/test-utils";
+import {Connection} from "../../../src";
+
+import {Person} from "./entity/person";
+import {Note} from "./entity/note";
+
+describe("github issues > #2965 Reuse preloaded lazy relations", () => {
+    let connections: Connection[];
+    before(async () => connections = await createTestingConnections({
+        entities: [ __dirname + "/entity/*{.js,.ts}" ],
+        // use for manual validation
+        // logging: true,
+    }));
+    beforeEach(() => reloadTestingDatabases(connections));
+    after(() => closeTestingConnections(connections));
+
+    it("should resuse preloaded lazy relations", () => Promise.all(connections.map(async connection => {
+
+        const repoPerson = connection.getRepository(Person);
+        const repoNote = connection.getRepository(Note);
+
+        const personA  = await repoPerson.create({ name: 'personA' });
+        const personB    = await repoPerson.create({ name: 'personB' });
+
+        await repoPerson.save([
+            personA,
+            personB,
+        ])
+
+        await repoNote.insert({ label: 'note1', owner: personA });
+        await repoNote.insert({ label: 'note2', owner: personB });
+
+
+        const originalLoad: (...args: any[]) => Promise<any[]> = connection.relationLoader.load;
+        let loadCalledCounter = 0;
+        connection.relationLoader.load = (...args: any[]): Promise<any[]> => {
+            loadCalledCounter++;
+            return originalLoad.call(connection.relationLoader, ...args);
+        }
+
+        {
+            const res = await repoPerson.find({ relations: ['notes'] });
+            const personANotes = await res[0].notes;
+            loadCalledCounter.should.be.equal(0);
+            personANotes[0].label.should.be.equal('note1');
+        }
+
+        {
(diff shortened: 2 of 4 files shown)


test/github-issues/2965/entity/person.ts
```diff
@@ -0,0 +1,14 @@
+import * as TypeOrm from '../../../../src/';
+import { Note } from './note';
+
+@TypeOrm.Entity()
```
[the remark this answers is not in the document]
StrikeForceZero: done in https://github.com/typeorm/typeorm/pull/2966/commits/0bd376be2de46e19906ff4f641567a79c59626ae

test/github-issues/2965/index.ts
```diff
@@ -0,0 +1,64 @@
...
+
+        const originalLoad: (...args: any[]) => Promise<any[]> = connection.relationLoader.load;
+        let loadCalledCounter = 0;
+        connection.relationLoader.load = (...args: any[]): Promise<any[]> => {
+            loadCalledCounter++;
+            return originalLoad.call(connection.relationLoader, ...args);
+        }
+
+        {
+            console.log('preload notes');
```
Kononnable: Let's clean these console logs - If test passes there is no need for it.

test/github-issues/2965/index.ts
```diff
@@ -0,0 +1,64 @@
...
+            const res = await repoPerson.find({ relations: [ 'notes' ] });
+            console.log(res);
+            const personANotes = await res[ 0 ].notes;
```
Kononnable: You use `[ 0 ]` here and `[0]` in line 50, it looks weird. Why different formatting? Same goes for lines 57 and 60.

Kononnable:
Looks good for me except for what I pointed out earlier.

pleerock:
Im not sure if we should continue to merge PRs related to "lazy relations feature" because I want to deprecate this experimental feature (since this feature brings a bad design and bring too much problems). What do you think @Kononnable

Kononnable:
If work is already done we can merge it and eventually remove it in 0.3.

I think some people use this feature and will be disappointed if we remove it. There are very few situations where this feature really shines, but most of the times it is just to make developer happy in cost of performance(lazy loading in general).
Lazy loading could get rid of some of it problems if getters/setters could accept different types and convert them to property type. Last I checked such feature request is marked as open in typescript repo and there are no plans to implement it in the near future. Until then lazy loading isn't very useful for me, but some folks use it for some reasons.

pleerock:
> I think some people use this feature and will be disappointed if we remove it.

of course they will be disappointed. But they will be disappointed even more if they choose the wrong way from the beginning and some day will simply stop using ORM because it will become too problematic for them

I can call current lazy loading implementation as an anti-pattern because it brings more problems then it solve.

Kononnable:
I don't like how it is implemented too. The question is: will we be able to implement it differently(in eg. 1 year), or are we going to say it is to complex to be implemented correctly(language, orm limitations etc.) and this feature probably won't come back in future versions of typeorm.

pleerock:
Its not about complexity, its about design and how good or bad this particular feature fits the design and philosophy of the framework. This particular feature can be deprecated because it imposes a bad design, it become to be a bad practice I would say.

The point of this feature is to load relation entities in some way, e.g. `const categories = await post.categories`. Same can be already done, maybe more verbose but basically same using relational query builder. We can simply deprecate this feature in favour of that syntax, and/or introduce a new additional syntax as I mentioned in #2902, something like `model.owner = repository.loadRelation("owner", model)`

test/github-issues/2965/index.ts
StrikeForceZero: copy/paste and intellij trying to be smart and format the pasted text. fixed.

test/github-issues/2965/index.ts
StrikeForceZero: all removed

StrikeForceZero:
@pleerock I totally understand why lazy loading can create some massive headaches.

That being said, is there an active issue being used to discuss this?

I guess while I'm here, only thing I would really like to see, before it being removed, is a way to load a relationship for an entity similar to what you described:  `model.owner = repository.loadRelation("owner", model)`, maybe even something like:
```typescript
model.owner = repository.loadRelation("owner");
// these options would be great for type saftey and refactoring support
model.owner = modelRepository.loadRelation(model => model.owner);
model.owner = repository.loadRelation(model, model => model.owner);
// bonus points for Entity Manager
model.owner = entityManager.loadRelation(model, model => model.owner);
modelRelationLoader = entityManager.relationLoader(model);
model.owner = modelRelationLoader(model => model.owner);
// or.. maybe getting a little crazy
modelOwnerRelationLoader = entityManager.relationLoader(Model)(model => model.owner);
model.owner = modelOwnerRelationLoader(model);
```

Kononnable:
@pleerock I didn't quite well put my thoughts into words. What I meant was that right now we have to remember about many things when using lazy loading. Some of these problems can be dealt with by using better design others will stay. We will never be able to achieve similar transparency as eg. Entity Framework because of language limitations. In EF there is no difference from user code perspective between using lazy loading and having each entity already loaded.

As for `model.owner = repository.loadRelation("owner", model)` - I don't like passing a string parameter, better to make use of type system. Also what if someone would try to use `model.owner` before it is loaded - should we return undefined or try to throw an error to make developer aware.

I don't think there is a issue with this proposal. Maybe we should create one and put ours ideas there - it will be easier to find and more people will be able to share their opinions.

pleerock:
Okay, lets merge this one then. But @StrikeForceZero be aware, I think we'll have to deprecate lazy loading functionality in the future and replace with suggested or similar api.

PR #2967, opened 2018-10-23
Use most specific matching relation type

amaranth:
This allows using a single table for multiple entities without using a
type column. Some databases infer the type from the context of how/where
the row is loaded.


amaranth:
I've rebased against the latest master to resolve the conflicts, should be good now.

AlexMesser:
thank you for contribution!

Issue #2968, opened 2018-10-23
Incompatability with tc39 decorators proposal


NilSet:
**Issue type:**

[ ] question
[x] bug report
[ ] feature request
[ ] documentation issue

**Database system/driver:**

[ ] `cordova`
[ ] `mongodb`
[ ] `mssql`
[ ] `mysql` / `mariadb`
[ ] `oracle`
[ ] `postgres`
[x] `sqlite`
[ ] `sqljs`
[ ] `react-native`
[ ] `expo`

**TypeORM version:**

[x] `latest`
[ ] `@next`
[ ] `0.x.x` (or put your version here)

**Steps to reproduce or a small repository showing the problem:**
Repro repo: https://github.com/NilSet/typeorm-decorator-incompat

tc39 class decorators are supplied not the class itself, but the descriptor which will be used to create the class.

The export of the module contains the constructed class, so when comparing the export to the annotation capture, they are not `===`. This makes `EntityMetadataBuilder` not find any matching annotated classes.

<img width="1128" alt="screen shot 2018-10-22 at 10 04 44 pm" src="https://user-images.githubusercontent.com/4490501/47336970-a5a66380-d646-11e8-9bb0-23513f12c22f.png">

https://github.com/tc39/proposal-decorators

NilSet:
After hacking together an Entity decorator which works with tc39 decorators, I ran into the same issue with the Column decorators, where it expects a class and instead gets a descriptor.

imnotjames:
We are currently using typescript decorators which are not compatible with the proposal. 

Once the proposal gets through and typescript implements we'll start using that.

purerosefallen:
TypeScript 5.0 comes with the new proposal. Reconsider this?

pleerock:
Yes, we are willing to support tc39 proposal. Feel free to contribute.

NSExceptional:
@pleerock can we reopen this?

andrew-demb:
Hi everyone. 
Does anyone have advice on how to implement support for TC39 decorators in TypeORM?

PR #2970, opened 2018-10-23
Proof of concept for handling tc39 decorator proposal

NilSet:
#2968  This is a proof of concept only updating the decorators used in https://github.com/NilSet/typeorm-decorator-incompat , but it does get that repo to run.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "typeorm",
-  "version": "0.2.6",
+  "version": "0.2.8",
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
--- a/src/decorator/columns/Column.ts
+++ b/src/decorator/columns/Column.ts
@@ -84,7 +84,7 @@ export function Column(type: (type?: any) => Function, options?: ColumnEmbeddedO
  * Only properties decorated with this decorator will be persisted to the database when entity be saved.
  */
 export function Column(typeOrOptions?: ((type?: any) => Function)|ColumnType|(ColumnOptions&ColumnEmbeddedOptions), options?: (ColumnOptions&ColumnEmbeddedOptions)): Function {
-    return function (object: Object, propertyName: string) {
+    return function (object: Object, propertyName?: string) {
 
         // normalize parameters
         let type: ColumnType|undefined;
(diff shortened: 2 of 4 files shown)


pleerock:
okay this is reproducible only with babel that supports latest decorator proposal, right?

We need to:

* make sure we support both new and old proposals
* make this support in all decorators

@Kononnable can you take care about this PR?

NilSet:
I believe the method parameter annotations are not supported by babel yet (the tc36 proposal for that is still in stage 0), so those ones will probably have to be skipped.

Kononnable:
https://tc39.github.io/proposal-decorators/ says it's in stage 2 out of 4(5). [The TC39 Process](https://tc39.github.io/process-document/) 
I'm not very familiar with how TC39 process works but I think until it's a candidate a lot of things can be changed there. Wouldn't it be wiser to wait a bit with that? Also most of the times TC39 proposals are later looked up by typescript maintainers and they try to implement it into language if possible.

asvetliakov:
@Kononnable the blocker for stage 3 is the champions can't make final conclusion should decorators go before or after export (or both, or block with export). Just the styling thing.

pleerock:
If decorators will become part of ES final they most likely won't be in the same form as they are right now. I also heard that they can't decide about decorator being before or after export keyword.

@NilSet can you take care about two points I've written in this PR?

We need to:

make sure we support both new and old proposals
make this support in all decorators

NilSet:
Sure, with the caveat that some of the decorators will not be possible to use via babel quite yet, but I can try to match what the stage 0 proposal for method parameter decorators currently looks like.

pleerock:
sounds like a plan :)

imnotjames:
This is really interesting!  Thank you for the contribution, but I'll be closing this.  It's a really cool proof of concept, though.

There's a new proposal for decorators which was to be proposed at the Sept TC39 meeting.  Not sure what came of that.  I wonder if there'd be a way to extract the decorators as they stand & then have two decorator packages?  Not sure what would be the best path there.

Anywho - yeah - thanks for the proof of concept!  Can also discuss this kinda stuff on the [community slack](https://join.slack.com/t/typeorm/shared_invite/zt-gej3gc00-hR~L~DqGUJ7qOpGy4SSq3g).

purerosefallen:
TypeScript 5.0 is a thing now.
Reconsider this?

PR #2983, opened 2018-10-24
Add support to Mysql / Postgres for duplicate entry update or ignore

apellizzn:
Hello, starting from issue #1780 and its relative PR #1793 I've updated the code of #1793 which now is based on the current master.
This solves #1780, support is provided only for MySql and Postgres at the moment.

### DIFF
--- a/src/query-builder/InsertQueryBuilder.ts
+++ b/src/query-builder/InsertQueryBuilder.ts
@@ -226,6 +226,29 @@ export class InsertQueryBuilder<Entity> extends QueryBuilder<Entity> {
         return this;
     }
 
+    /**
+     * Adds additional ignore statement supported in databases.
+     */
+    orIgnore(statement: string | boolean = true): this {
+        this.expressionMap.onIgnore = statement;
+        return this;
+    }
+
+    /**
+     * Adds additional update statement supported in databases.
+     */
+    orUpdate(statement?: { columns?: string[], conflict_target?: string | string[] }): this {
+      this.expressionMap.onUpdate = {};
+      if (statement && statement.conflict_target instanceof Array)
+          this.expressionMap.onUpdate.conflict = ` ( ${statement.conflict_target.join(", ")} ) `;
+      if (statement && typeof statement.conflict_target === "string")
+          this.expressionMap.onUpdate.conflict = ` ON CONSTRAINT ${statement.conflict_target} `;
+      if (statement && statement.columns instanceof Array)
+          this.expressionMap.onUpdate.columns = statement.columns.map(column => `${column} = :${column}`).join(", ");
+      return this;
+  }
+
+
     // -------------------------------------------------------------------------
     // Protected Methods
     // -------------------------------------------------------------------------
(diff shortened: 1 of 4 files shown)


pleerock:
Looks okay I think, what do you think @Kononnable

src/query-builder/InsertQueryBuilder.ts
```diff
@@ -226,6 +226,29 @@ export class InsertQueryBuilder<Entity> extends QueryBuilder<Entity> {
         return this;
     }
 
+    /**
+     * Adds additional ignore statement supported in databases.
```
Kononnable: Should we add information which database engines suport such statements? Like in line 222.

test/github-issues/1780/issue-1780.ts
```diff
@@ -0,0 +1,160 @@
...
+     let user2 = new User();
+    user2.first_name = "John";
+    user2.last_name = "Lenon";
+    user2.is_updated = "yes";
+     // let data = [user1, user2];
+    // Bulk insertion with duplicated data through same query with duplicate error exception is not supported in PostgreSQL
+    // https://doxygen.postgresql.org/nodeModifyTable_8c_source.html : Line 1356
+     it("should save one row without duplicate error in MySQL/MariaDB", () => Promise.all(connections.map(async connection => {
+      try {
+         if (connection.driver instanceof MysqlDriver) {
```
Kononnable: We're creating connection to every db engine supported by typeorm but using only three. Reloading is also not necessary since we use different driver in different test.
It isn't much difference but it will take another second or two if every supported db engine will be tested.

Kononnable:
Changes looks good. Should we mention them somewhere(docs,changelog)?

Formatting is kind of weird(for me), but it's more like a global problem. Maybe we should think about forcing similar formatting in typeorm codebase - like tslint/prettier running on every `git commit`.

pleerock:
Yeah Im keeping merging PRs someone will need to update a changelog once we release next version

pleerock:
lets merge this one

apellizzn:
Yes I also feel formatting is not the best but I've tried to follow what is already there :)

src/query-builder/InsertQueryBuilder.ts
```diff
@@ -234,14 +257,17 @@ export class InsertQueryBuilder<Entity> extends QueryBuilder<Entity> {
      * Creates INSERT express used to perform insert query.
      */
     protected createInsertExpression() {
-
         const tableName = this.getTableName(this.getMainTableName());
         const valuesExpression = this.createValuesExpression(); // its important to get values before returning expression because oracle rely on native parameters and ordering of them is important
         const returningExpression = this.createReturningExpression();
         const columnsExpression = this.createColumnNamesExpression();
+        let query = "INSERT "
```
jsnick: @apellizzn tslint: missing semicolon

src/query-builder/InsertQueryBuilder.ts
```diff
@@ -234,14 +257,17 @@ export class InsertQueryBuilder<Entity> extends QueryBuilder<Entity> {
...
 
-        // generate INSERT query
-        let query = `INSERT INTO ${tableName}`;
+        if(this.connection.driver instanceof MysqlDriver) {
```
jsnick: @apellizzn tslint: missing whitespace

src/query-builder/InsertQueryBuilder.ts
```diff
@@ -234,14 +257,17 @@ export class InsertQueryBuilder<Entity> extends QueryBuilder<Entity> {
...
+          query+= `${this.expressionMap.onIgnore ? " IGNORE " : ""}`
```
jsnick: tslint: missing semicolon

PR #2985, opened 2018-10-24
Only check for discriminator conflicts on STI entities

### LINKED ISSUE #2984: Discriminator conflict reported even for non-inherited tables
**Issue type:**

[ ] question
[X] bug report
[ ] feature request
[ ] documentation issue

**Database system/driver:**

[ ] `cordova`
[ ] `mongodb`
[ ] `mssql`
[X] `mysql` / `mariadb`
[ ] `oracle`
[ ] `postgres`
[ ] `sqlite`
[ ] `sqljs`
[ ] `react-native`
[ ] `expo`

**TypeORM version:**

[X] `latest`
[ ] `@next`

If you have two entity classes with the same name and one of them uses `@TableInheritance` TypeORM will check to see if other entities have the same discriminator and fail because the other class does. However, as this other class is just a plain entity the conflict shouldn't matter, they don't have any effect on each other.

`entities/issue/note.ts`:
```typescript
@Entity({name: "issueNote"})
@TableInheritance({column: {type: "varchar", name: "type"}})
export class Note {

    @PrimaryGeneratedColumn()
    public id: number;

}
```

`entities/commit/note.ts`:
```typescript
@Entity({name: "commitNote"})
export class Note {

    @PrimaryGeneratedColumn()
    public id: number;

}
```

```
Error: Entities Note and Note as equal discriminator values. Make sure their discriminator values are not equal using @DiscriminatorValue decorator.
    at EntityMetadataValidator.validate (/Users/travis/Projects/velocity/typeorm/src/metadata-builder/EntityMetadataValidator.ts:71:23)
```

amaranth:
If an entity has the same discriminator value as one using
@TableInheritance this should be allowed as there is no way for them to
conflict. Fixes #2984.

### DIFF
--- a/test/github-issues/2984/issue-2984.ts
+++ b/test/github-issues/2984/issue-2984.ts
@@ -1,25 +1,21 @@
 import "reflect-metadata";
 
-import {closeTestingConnections, reloadTestingDatabases, setupSingleTestingConnection} from "../../utils/test-utils";
-import {createConnection, Connection} from "../../../src";
+import {closeTestingConnections, createTestingConnections, reloadTestingDatabases} from "../../utils/test-utils";
+import {Connection} from "../../../src";
 
 describe("github issues > #2984 Discriminator conflict reported even for non-inherited tables", () => {
-    let connection: Connection;
 
-    before(async () => connection = await createConnection(setupSingleTestingConnection("postgres", {
+    let connections: Connection[];
+    before(async () => connections = await createTestingConnections({
         entities: [__dirname + "/entity/**/*{.js,.ts}"],
-        schemaCreate: true,
-        dropSchema: true,
-    })));
-    beforeEach(async () => {
-        await reloadTestingDatabases([connection]);
-    });
-    after(() => closeTestingConnections([connection]));
+    }));
+    beforeEach(() => reloadTestingDatabases(connections));
+    after(() => closeTestingConnections(connections));
 
-    it("should load entities even with the same discriminator", async () => {
+    it("should load entities even with the same discriminator", () => Promise.all(connections.map(async connection => {
         connection.entityMetadatas.should.have.length(2);
         connection.entityMetadatas.forEach(metadata =>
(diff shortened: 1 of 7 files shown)


test/github-issues/2984/issue-2984.ts
```diff
@@ -0,0 +1,25 @@
+import "reflect-metadata";
+
+import {closeTestingConnections, reloadTestingDatabases, setupSingleTestingConnection} from "../../utils/test-utils";
+import {createConnection, Connection} from "../../../src";
+
+describe("github issues > #2984 Discriminator conflict reported even for non-inherited tables", () => {
+    let connection: Connection;
+
+    before(async () => connection = await createConnection(setupSingleTestingConnection("postgres", {
```
Kononnable: Why test it on just one database if fix is not database type specific?

amaranth: Whoops, that was a copy/paste from another test. I should be able to push a change for that tomorrow if you want.

pleerock: looks like in this case he doesn't need to do it since he don't really use connection. He just needs to check entity validation process.

pleerock:
I think there is a different problem. Why would you get this error if you don't use entity children? Discriminator values are created for entity children right? In this case your second Node isn't child of that parent. It looks we need to check if your second node is a child of your first node. You probably need to use `entityMetadata.inheritanceTree` property for this purpose.

amaranth:
So while adding the change to make it only fail conflicts in the same `inheritanceTree` I discovered the original check was actually broken. Only the class with `@TableInheritance` gets marked as "STI", the rest are just "entity-child" type. That means the original check would block situations that should be allowed (like what this PR was originally about) but would allow situations that should be blocked (like having two children with the same `discriminatorValue`).

imnotjames (approved):
This looks good.  Just want to rebase so all tests run and then we are good to go.

amaranth:
Rebased and all the tests passed, should be good to go.

PR #2986, opened 2018-10-24
Make children inherit database option from parent

amaranth:
Without this children use the default database for the connection which may be undefined or not the same as the one on the parent entity.


pleerock:
Thanks!

PR #3030, opened 2018-11-01
Fix tslint errors and turn on tslint commit hook and on CI

amaranth:
The first commit in this PR fixes all reported tslint errors. The second one adds a commit hook to run tslint and try to autofix errors, like was done previously. It also changes the travis-ci configuration to run tslint and generate code coverage reports again.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,10 +16,10 @@ before_script:
   - sudo service mysql stop
   - sudo service postgresql stop
   - docker-compose up -d
-  - cp ormconfig.travis.json ormconfig.json
+  - npm run setup:config
 
 script:
-  - npm test
+  - npm run test-ci
 
 after_success:
   - bash <(curl -s https://codecov.io/bash)
--- a/package.json
+++ b/package.json
@@ -55,7 +55,7 @@
     "gulp-tslint": "^8.1.3",
     "gulp-typescript": "^4.0.2",
     "gulpclass": "^0.1.2",
-    "husky": "^0.14.3",
+    "husky": "^1.1.2",
     "lint-staged": "^7.0.0",
     "mocha": "^5.0.5",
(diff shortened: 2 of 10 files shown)


.travis.yml
```diff
@@ -16,10 +16,10 @@ before_script:
   - sudo service mysql stop
   - sudo service postgresql stop
   - docker-compose up -d
-  - cp ormconfig.travis.json ormconfig.json
+  - npm run setup:config
```
pleerock: this is not needed really, previous command is simple and I would prefer not to use gulp too much.

.travis.yml
```diff
@@ -16,10 +16,10 @@ before_script:
...
 
 script:
-  - npm test
+  - npm run test-ci
```
pleerock: can you please explain this change?

package.json
```diff
@@ -91,17 +91,24 @@
...
     "*.ts": [
-      "tslint --fix",
+      "tslint --project . --fix",
       "git add"
     ]
   },
   "scripts": {
     "test-ci": "gulp ci-tests",
-    "test": "rimraf ./build && tsc && mocha --file ./build/compiled/test/utils/test-setup.js --bail --recursive --timeout 30000  ./build/compiled/test",
+    "test": "gulp tests",
```
pleerock: can you please explain this change?

pleerock:
> Right now things are in a weird hybrid state of using gulp or just using npm. If you want to avoid gulp should I make a PR to remove gulp usage entirely? Otherwise, we should port everything back to going through gulp for consistency which this PR attempts to do.

I would like to avoid using gulp but Im not sure if its efficiently possible.

imnotjames:
We're no longer using tslint nor travis so I'll be closing this out.  Cheers!

PR #3039, opened 2018-11-03
Fix #3009 use newer extension for postgress UUID gen.

### LINKED ISSUE #3009: PostgreSQL UUID
**Issue type:**

[ ] question
[ ] bug report
[x] feature request
[ ] documentation issue

**Database system/driver:**

[ ] `cordova`
[ ] `mongodb`
[ ] `mssql`
[ ] `mysql` / `mariadb`
[ ] `oracle`
[x] `postgres`
[ ] `sqlite`
[ ] `sqljs`
[ ] `react-native`
[ ] `expo`

**TypeORM version:**

[ ] `latest`
[ ] `@next`
[x] `0.2.8` (or put your version here)

> Historically this module depended on the OSSP UUID library, which accounts for the module's name. While the OSSP UUID library can still be found at http://www.ossp.org/pkg/lib/uuid/, it is not well maintained, and is becoming increasingly difficult to port to newer platforms.

`If you only need randomly-generated (version 4) UUIDs, consider using the gen_random_uuid() function from the pgcrypto module instead.`

[Source](https://www.postgresql.org/docs/10/static/uuid-ossp.html).

So, when @PrimaryGeneratedColumn('uuid') used, seems it's recommended to use another extension.

**By the way, how to change the name of constraint that PrimaryGeneratedColumn create?**

gintsgints:
Attention! New extension does not support Old postgres versions: 9.2 / 9.1 / 9.0 / 8.4 / 8.3

### DIFF
--- a/src/driver/postgres/PostgresDriver.ts
+++ b/src/driver/postgres/PostgresDriver.ts
@@ -299,9 +299,9 @@ export class PostgresDriver implements Driver {
                         if (err) return fail(err);
                         if (hasUuidColumns)
                             try {
-                                await this.executeQuery(connection, `CREATE EXTENSION IF NOT EXISTS "uuid-ossp"`);
+                                await this.executeQuery(connection, `CREATE EXTENSION IF NOT EXISTS "pgcrypto"`);
                             } catch (_) {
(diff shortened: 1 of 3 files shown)


pleerock:
Hey, thank you for contribution!

I think its better to create this PR against `next` branch since its a breaking change :)
Also we'll need to update a changelog

gintsgints:
Made another on next branch.

PR #3055, opened 2018-11-06
Support development in node10 (LTS)

jsnick:


test/functional/database-schema/column-types/mssql/column-types-mssql.ts
```diff
@@ -137,7 +137,7 @@ describe("database schema > column types > mssql", () => { // https://github.com
         table!.findColumnByName("binary")!.type.should.be.equal("binary");
         table!.findColumnByName("varbinary")!.type.should.be.equal("varbinary");
         table!.findColumnByName("image")!.type.should.be.equal("image");
-        table!.findColumnByName("rowversion")!.type.should.be.equal("rowversion");
```
pleerock: can you please explain this change?

since mssql tests aren't running on CI I can't test if it worked wrong before or your changes are wrong.

package.json
```diff
@@ -99,7 +99,7 @@
   },
   "scripts": {
     "test-ci": "gulp ci-tests",
-    "test": "rimraf ./build && tsc && mocha --file ./build/compiled/test/utils/test-setup.js --bail --recursive --timeout 30000  ./build/compiled/test",
+    "test": "rimraf ./build && tsc && mocha --file ./build/compiled/test/utils/test-setup.js  --recursive --timeout 60000 ./build/compiled/test",
```
pleerock: whats wrong with `--bail`?

amaranth: `--bail` makes the tests stop after the first failure. This can be a good thing as tests that run later in the same file seem to all fail anyway due to transaction issues but, on the other hand, it can make finding and fixing test failure across the entire test suite a massive slog as you have to keep running all of the tests over and over each time you fix the latest failure.

test/functional/database-schema/column-types/mssql/column-types-mssql.ts
amaranth: https://docs.microsoft.com/en-us/sql/t-sql/data-types/rowversion-transact-sql?view=sql-server-2017 seems to imply `timestamp` and `rowversion` are identical, perhaps `rowversion` only exists in the query engine and the backend is just always `timestamp`. I tried to enable mssql for tests but ran in to a number of issues so couldn't reproduce this either way.

jsnick: I don't know why. But, If expected value is `rowversion`, only this test fails. Others succeed.
```
1) database schema > column types > mssql
       all types should work correctly - persist and hydrate:

      AssertionError: expected 'timestamp' to equal 'rowversion'
      + expected - actual

      -timestamp
      +rowversion
```

My test environment.
```
Node 10.13.0
Mac OS X 10.13.6 (High Sierra)
msssql instanced by docker-compose.yml (in this pull-request)
("mcr.microsoft.com/mssql/server:2017-GA-ubuntu")
```

package.json
jsnick: Ooops, this change happened while I was setting typeorm development environment for checking total failed tests.
it's my commit mistake. Thanks for reviewing!

test/functional/database-schema/column-types/mssql/column-types-mssql.ts
pleerock: @amaranth so tests are currently passing on your side right? what if you change in your own test from 'rowversion' to 'timestamp' ?

amaranth: Tests aren't passing normally but after #3162 to make this test work and disabling others that fail for different reasons I can confirm that even on node 8.12 and more or less current TypeORM master this also returns `timestamp` instead of `rowversion`.

jsnick: I reverted this. It is maybe better to handle this in new issue.

havenchyk:
LGTM :)

havenchyk:
@pleerock @Kononnable can we merge this?

pleerock:
Thank you!

PR #3056, opened 2018-11-06
Add support to use current values in insert Mysql/Postgres

### LINKED ISSUE #3047: Mysql / Postgres on duplicate key update / on conflict use current values
**Issue type:**

[ ] question
[ ] bug report
[x ] feature request
[ ] documentation issue

**Database system/driver:**

[ ] `cordova`
[ ] `mongodb`
[ ] `mssql`
[x ] `mysql` / `mariadb`
[ ] `oracle`
[ x] `postgres`
[ ] `sqlite`
[ ] `sqljs`
[ ] `react-native`
[ ] `expo`

**TypeORM version:**

[x ] `latest`
[ ] `@next`

In MySQL when we want to perform an INSERT and handle what happens ON DUPLICATE KEY we can use the current values to perform an update, this may happens usually when performing multiple rows insert at the same time.

Also we want to do the same in Postgres ON CONFLICT

Example
- Table
```
users (unique index on (first_name, last_name)
| id | first_name | last_name | is_updated |
|----|------------|-----------|------------|
| 1  | John       | Lennon    | no         |
```
- Insert sql
```sql
INSERT INTO users `user`(`id`, `first_name`, `last_name`, `is_updated`) 
VALUES 
(DEFAULT, "Ringo", "Star", "no")
(DEFAULT, "John", "Lennon", "yes")
```

What we want as result is the following
```
users (unique index on (first_name, last_name)
| id | first_name | last_name | is_updated |
|----|------------|-----------|------------|
| 1  | John       | Lennon    | yes        |
| 2  | Ringo      | Star      | no         |
```

To achieve this we can use the following syntax (MYSQL)
```sql
INSERT INTO users `user`(`id`, `first_name`, `last_name`, `is_updated`) 
VALUES 
(DEFAULT, "Ringo", "Star", "no")
(DEFAULT, "John", "Lennon", "yes")
ON DUPLICATE KEY UPDATE is_updated=VALUES(is_updated)
```

(POSTGRES)
```sql
INSERT INTO users `user`(`id`, `first_name`, `last_name`, `is_updated`) 
VALUES 
(DEFAULT, "Ringo", "Star", "no")
(DEFAULT, "John", "Lennon", "yes")
ON CONFLICT ( first_name, last_name )
DO UPDATE SET is_updated = EXCLUDED.is_updated
```

I would like to proceed with a PR about this, what do you think about ?

apellizzn:
In Mysql / Postgres add support to update record during INSERT on conflict / duplicate key using current values, implements, fixes #3047

### DIFF
--- a/src/metadata/ColumnMetadata.ts
+++ b/src/metadata/ColumnMetadata.ts
@@ -532,7 +532,7 @@ export class ColumnMetadata {
      * If column is in embedded (or recursive embedded) it extracts its value from there.
      */
      getEntityValue(entity: ObjectLiteral, transform: boolean = false): any|undefined {
-        // if (entity === undefined || entity === null) return undefined; // uncomment if needed
+        if (entity === undefined || entity === null) return undefined;
 
         // extract column value from embeddeds of entity if column is in embedded
         let value: any = undefined;
--- a/src/metadata/RelationMetadata.ts
+++ b/src/metadata/RelationMetadata.ts
@@ -325,7 +325,7 @@ export class RelationMetadata {
      * If column is in embedded (or recursive embedded) it extracts its value from there.
      */
     getEntityValue(entity: ObjectLiteral, getLazyRelationsPromiseValue: boolean = false): any|undefined {
-
+        if (entity === null || entity === undefined) return undefined;
         // extract column value from embeddeds of entity if column is in embedded
         if (this.embeddedMetadata) {
 
--- a/src/query-builder/InsertQueryBuilder.ts
+++ b/src/query-builder/InsertQueryBuilder.ts
@@ -237,7 +237,7 @@ export class InsertQueryBuilder<Entity> extends QueryBuilder<Entity> {
     /**
      * Adds additional update statement supported in databases.
      */
-    orUpdate(statement?: { columns?: string[], conflict_target?: string | string[] }): this {
+    orUpdate(statement?: { columns?: string[], overwrite?: string[], conflict_target?: string | string[] }): this {
       this.expressionMap.onUpdate = {};
       if (statement && statement.conflict_target instanceof Array)
           this.expressionMap.onUpdate.conflict = ` ( ${statement.conflict_target.join(", ")} ) `;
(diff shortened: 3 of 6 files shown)


jsnick:
I think there is atomic update case like `column = column + 1`

apellizzn:
@jsnick I think we can use `column = column + 1` when we want to update the value based on the current value in the database.
What i am doing with this PR is to update the value based on the current record that i am trying to insert.
Using `column = column + 1` will produce
```
| id | counter |
|----|---------|
| 1  | 5       |
```
```sql
INSERT INTO foo (id, counter) VALUES (1, 10) ON DUPLICATE KEY UPDATE counter  = counter +1 
```
```
| id | counter |
|----|---------|
| 1  | 6       |
```
Using `column = VALUES(column)` will produce
```
| id | counter |
|----|---------|
| 1  | 5       |
```
```sql
INSERT INTO foo (id, counter) VALUES (1, 10) ON DUPLICATE KEY UPDATE counter  = VALUES(counter)
```
```
| id | counter |
|----|---------|
| 1  | 10       |
```

marshall007:
It would be nice if a high-level `conflict: 'ignore' | 'replace' | 'error'` option were also exposed on `EntityManager`'s `save/update/insert` APIs.

apellizzn:
@marshall007 Sure, to me this sounds like another PR i would like to merge this one first.

pleerock:
QueryBuilder is a complex task since we replace property names in database names and use property names in queries in most of places. In this case looks like we have to use database names, and its confusing. 

Lets merge it for now, but someday we'll need to figure out what to do with this confusion. 

pleerock:
ooups looks like a conflict, can you please resolve it?

apellizzn:
@pleerock Sorry for the delay, resolved :)

src/metadata/ColumnMetadata.ts
```diff
@@ -532,7 +532,7 @@ export class ColumnMetadata {
      * If column is in embedded (or recursive embedded) it extracts its value from there.
      */
      getEntityValue(entity: ObjectLiteral, transform: boolean = false): any|undefined {
-        // if (entity === undefined || entity === null) return undefined; // uncomment if needed
```
pleerock: so, why its needed?)

apellizzn: Because in some case when you actually insert multiple rows which cause a conflict `entity` is not defined, so therefore line 595 will break `value = entity[this.propertyName];`

Database with columns c1, c2, c3 with a unique index on c1 and the following data
```
| A (c1) | Foo (c2) | Bar (c3) |
```
If i try to insert the following
```
| A | Foo2 | Bar2 |
| A | Foo3 | Bar3 |
```
I just want the database to be  `| A | Foo3 | Bar3 |` i guess this has something to do with the fact that you are actually trying to insert N rows but returning N-1

pleerock: mm not sure I got it, but okay, lets merge it

pleerock:
Thank you!

Issue #2641, opened 2018-08-09
DISTINCT in query builder


dariusmb:
**Issue type:**

[x] question
[ ] bug report
[ ] feature request
[ ] documentation issue

**Database system/driver:**

[ ] `cordova`
[ ] `mongodb`
[ ] `mssql`
[x] `mysql` / `mariadb`
[ ] `oracle`
[ ] `postgres`
[ ] `sqlite`
[ ] `sqljs`
[ ] `react-native`
[ ] `expo`

**TypeORM version:**

[x] `latest`
[ ] `@next`
[ ] `0.x.x` (or put your version here)

**Steps to reproduce or a small repository showing the problem:**

<!--
    To answer those questions you need to put "x" inside the square brackets, for example:
    [x] `mysql`
    [ ] `postgres`

    Also, please format your code properly (by taking code blocks into ```ts .... ```)
!>

Is there any way to SELECT DISTINCT using query builder?

afurculita:
Why not simply:

```
const applications = await Application.createQueryBuilder("applications")
    .select('DISTINCT applications.id', 'id')
    .addSelect('applications.loan_number', 'loanNumber')
    .addSelect('applicant.id', 'applicantId')
    .addSelect('ass.status', 'signature_status')
    .leftJoin('applications.applicationApplicants', 'applicant')
    .leftJoin('applications.applicationSignatureStatus', 'ass')
    .where("applications.deleted_at = null")
    .orderBy("applications.id", "DESC")
    .limit(4)
    .getRawMany();
```
?

gabrielshawandpartners:
@afurculita The Postgres `DISTINCT ON` behaves a bit different from `DISTINCT` clause, I simplified the query a bit just to post here, but anyway, the same problem happens with `DISTINCT`

gabrielshawandpartners:
I think we could have a `distinct`, like KNEXT does, like:
```
knex('tablename').distinct(knex.raw('ON (column) column'))
.orderBy('column')
.orderBy('another_column');
```

exoszajzbuk:
Same issue here, the `DISTINCT ON` select is not getting placed as the first select => syntax error

ollija:
I made a PR fixing this issue https://github.com/typeorm/typeorm/pull/3065

kajalj3098:
select distinct p.id_company as idCompany
[...]

PR #3065, opened 2018-11-07
Fix invalid SQL syntax when using distinct and addSelect on Postgresql

ollija:
Fixes this: https://github.com/typeorm/typeorm/issues/2641 for **Postgresql**

When using `distinct` with `addSelect`, the `distinct`-selection will be behind all the other `selects` in the resulting SQL.

So both this:
```javascript
createQueryBuilder(Bar, "bar")
    .select('DISTINCT(bar.name)', 'name')
    .addSelect('bar.description', 'description')
    .getRawMany();
```

and this:

```javascript
createQueryBuilder(Bar, "bar")
    .select('bar.description', 'description')
    .addSelect('DISTINCT(bar.name)', 'name')
    .getRawMany();
```

Will produce the same sql resulting to this error:
```
ERROR:  syntax error at or near "DISTINCT" at character 46
STATEMENT:  SELECT "bar"."description" AS "description", DISTINCT("bar"."name") AS "name" FROM "bar" "bar"
```
The same error is produced when using` DISTINCT ON (...)` with `addSelect`.
Valid syntax would be with the distinct selection before other selections:

`SELECT DISTINCT("bar"."name") AS "name", "bar"."description" AS "description" FROM "bar" "bar"`

This change ensures that `DISTINCT` / `DISTINCT ON (...)` selections are added to the front of the select clauses.

### DIFF
--- a/src/query-builder/SelectQueryBuilder.ts
+++ b/src/query-builder/SelectQueryBuilder.ts
@@ -107,15 +107,15 @@ export class SelectQueryBuilder<Entity> extends QueryBuilder<Entity> implements 
     select(selection?: string|string[]|((qb: SelectQueryBuilder<any>) => SelectQueryBuilder<any>), selectionAliasName?: string): SelectQueryBuilder<Entity> {
         this.expressionMap.queryType = "select";
         if (selection instanceof Array) {
-            this.expressionMap.selects = selection.map(selection => ({ selection: selection }));
+            this.expressionMap.selects = selection.map(this.getSelectionWithColumnPath);
 
         } else if (selection instanceof Function) {
             const subQueryBuilder = selection(this.subQuery());
             this.setParameters(subQueryBuilder.getParameters());
-            this.expressionMap.selects.push({ selection: subQueryBuilder.getQuery(), aliasName: selectionAliasName });
+            this.expressionMap.selects.push({ ...this.getSelectionWithColumnPath(subQueryBuilder.getQuery()), aliasName: selectionAliasName });
 
         } else if (selection) {
-            this.expressionMap.selects = [{ selection: selection, aliasName: selectionAliasName }];
+            this.expressionMap.selects = [{ ...this.getSelectionWithColumnPath(selection), aliasName: selectionAliasName }];
         }
 
         return this;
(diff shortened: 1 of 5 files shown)


pleerock:
The issue itself is very weird - why query builder adds selects with "distinct" after everything else. We need to find a root issue and fix that proper way, suggested fix is just a hack.

pleerock:
@ollija any news here?

rgpass:
Hoping to have this fixed. Having to do a manual SQL call for this currently.

kernel-io:
@pleerock what fix would be deemed as 'proper' for this?

pleerock:
@kernel-io proper fix is just finding a root cause of this issue and fixing it.

ollija:
@pleerock Please review

src/query-builder/SelectQueryBuilder.ts
```diff
@@ -1962,4 +1973,20 @@ export class SelectQueryBuilder<Entity> extends QueryBuilder<Entity> implements
...
     }
 
+    /**
+     * Gets the selected column from the selection. Is not always the same as the selection itself:
+     * 'DISTINCT ON (foo.bar) foo.bar' has the columnPath of 'foo.bar'
+     */
+    protected getSelectionWithColumnPath(selection: string) {
+        let columnPath: string = selection;
+
+        if (/(distinct)/i.test(selection) && this.connection.driver instanceof PostgresDriver) {
```
ollija: I would've put `this.connection.driver instanceof PostgresDriver`-condition before the regexp, but that way some tests broke with `this.connection` being undefined

ricardovf:
When will this be merged? Need to use `distinct on` feature of postgres. Thanks!

src/query-builder/SelectQueryBuilder.ts
Kononnable:
```suggestion
        if (this.connection.driver instanceof PostgresDriver && /(distinct)/i.test(selection)) {
```
`this.connection.driver instanceof PostgresDriver` is easier to check(performance) then regexp, so it should be probably as first condition

TheNoim:
When does this finally gets merged?

src/query-builder/SelectQueryBuilder.ts
nikolovjd: > ollija on Feb 28  Author
> I would've put this.connection.driver instanceof PostgresDriver-condition before the regexp, but that way some tests broke with this.connection being undefined

This was commented 2 months before your review. This is a very serious bug, makes MANY complex queries completely impossible to build via querybuilder.

Kononnable: Looks like I've missed it. Thanks

JosephTico:
Why hasn't this been merged?

smonv:
@pleerock Please review this PR. It's quite annoying to working with `DISTINCT ON` right now.

MarZab:
Do you have any plans to merge this in?

yagoferrer:
I was running into the same issue and I found out that there are two methods you can use to fix issues with **distinct**

```
.select('user.id', 'id')
.distinct(true)
```

```
.select('user.id', 'id')
.distinctOn(['user.id'])
```

test/github-issues/2641/issue-2641.ts
```diff
@@ -0,0 +1,55 @@
...
+import {Connection} from "../../../src/connection/Connection";
+import {Foo} from "./entity/Foo";
+import {Bar} from "./entity/Bar";
+
+describe("github issues > #2641 - DISTINCT in query builder", () => {
+
+    let connections: Connection[];
+    before(async () => connections = await createTestingConnections({
+        entities: [__dirname + "/entity/*{.js,.ts}"],
+        enabledDrivers: ["postgres"]
```
imnotjames, line 12: This is the query builder.  Run tests against all drivers please!

src/query-builder/SelectQueryBuilder.ts
imnotjames, line 1983: What if my column name is `distinct`

imnotjames (changes requested):
Please rebase and fix conflicts

AlexMesser:
Superseded by #4109 and #4954. Anyway, thank you for contribution!

PR #3081, opened 2018-11-09
More accurate type for Postgres ssl parameters

rixtox:
Reference to https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/pg/index.d.ts#L35


pleerock:
Hey, thank you for the contribution!

PR #3082, opened 2018-11-09
Fix app-root-path import not working with webpack

alfaproject:
This happens in a production bundle `Cannot find module '/var/task/node_modules/app-root-path'` because Webpack doesn't include `app-root-path` in the bundle.

I'm not quite sure why this module is being lazy loaded, but if it's meant to be lazy loaded then it should use dynamic imports, which I'm happy to change to if that's preferred.

### DIFF
--- a/src/platform/PlatformTools.ts
+++ b/src/platform/PlatformTools.ts
@@ -90,7 +90,7 @@ export class PlatformTools {
                 */
                 case "sql.js":
                     return require("sql.js");
-                    
+
                 /**
                 * sqlserver
                 */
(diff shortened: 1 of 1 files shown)


pleerock:
Hi and thank you for contribution!

The reason why its being lazily loaded is because typeorm is cross-platform and we don't wanna go all node dependencies into browser package for example (we remove all those dependencies on the fly for browser version).

To make this to work in webpack properly its better to change PlatformTools' load method, there is a code where we explicitly compare each module and require it in a format like `(if module === "app-root-path") { return require("app-root-path") }` which makes webpack happy.

alfaproject:
@pleerock alright, fixed it that way now. (:

pleerock:
Thank you!

PR #3083, opened 2018-11-09
Fix incorrect property name in documentation

misterGF:
The correct property should be pool.evictionRunIntervalMillis instead of pool.victionRunIntervalMillis


havenchyk:
Seems to be legit. @pleerock @Kononnable please approve :)

PR #3091, opened 2018-11-10
[Doc] Typo?

RevealedFrom:
Shouldn't it be the table `user`?


havenchyk:
You're right, it seems to be a typo.

@pleerock @Kononnable could you please take a look?

PR #3139, opened 2018-11-21
Fix issue template MD checkboxes

PapsOu:
The aim of this PR is only to facilitate user's issue reporting


havenchyk:
@pleerock @Kononnable

pleerock:
thanks for the contribution, but Im going to revert this change since we can't use md checkboxes. The problem is that if you use md checkboxes github will treat it as a task list and for example for "mysql, [x] postgres, oracle" it will treat it as 1/3 tasks are resolved.

PR #3140, opened 2018-11-21
feat: add a ttl to the redis key

killix:
On the next branch, the redis key doesn't have an expiry time.
I found this [PR](https://github.com/typeorm/typeorm/pull/2507) for `master`. But no information about why this was not done on `next`.


pleerock:

its because latest master wasn't merged into next yet.

Issue #3125, opened 2018-11-19
CreateDateColumn, UpdateDateColumn, VersionColumn not present in some case


GitStorageOne:
**Issue type:**

[ ] question
[x] bug report
[ ] feature request
[ ] documentation issue

**Database system/driver:**

[ ] `cordova`
[ ] `mongodb`
[ ] `mssql`
[ ] `mysql` / `mariadb`
[ ] `oracle`
[x] `postgres`
[ ] `sqlite`
[ ] `sqljs`
[ ] `react-native`
[ ] `expo`

**TypeORM version:**

[ ] `latest`
[ ] `@next`
[x] `0.2.9` (or put your version here)

**Steps to reproduce or a small repository showing the problem:**

**model:**

```ts
export abstract class BaseCollection extends BaseEntity {

  @PrimaryColumn('uuid', { name: 'id' })
  id: string;

  @CreateDateColumn({ type: 'timestamp', name: 'create_date', default: () => 'LOCALTIMESTAMP' })
  createDate: string;
  
  @UpdateDateColumn({ type: 'timestamp', name: 'update_date', default: () => 'LOCALTIMESTAMP' })
  updateDate: string;

  @VersionColumn({ name: 'data_version' })
  dataVersion: number;

  constructor() {
    super();
    if (!this.id) {
      this.id = uuidv4();
    }
  }

}
```

```ts
@Entity('users')
export class User extends BaseCollection {

  @Column('varchar', { name: 'surname', length: 100, nullable: true })
  surname: string | null;

  @Column('varchar', { name: 'name', length: 100, nullable: true })
  name: string;

  @Column('varchar', { name: 'email', length: 200, unique: true })
  email: string;

}
```

**Test case:**

```ts
  User.create(user).save()
    .then((result) => {
      res.set('Item-Id', result.id);
      res.set('Data-Version', result.dataVersion.toString());
      res.sendStatus(200);
    })
```
where `result` is object with:
```json
{
   "id":"d8a66a45-339b-4cbe-97e3-2c7e81b73b24",
   "surname":"test surname",
   "name":"test name",
   "email":"Test@test.com",
   "createDate":"2018-11-19T11:17:39.816Z",
   "updateDate":"2018-11-19T11:17:39.816Z",
   "dataVersion":1
}
```

Now, if we try to save entity with same data again (all fields are the same), `result` object looks like this:

```json
{
   "id":"d8a66a45-339b-4cbe-97e3-2c7e81b73b24",
[...]

PR #3143, opened 2018-11-21
Add a unit test for issue #3125

alfaproject:
Here is a unit test for issue #3125

1. The primary key is not autogenerated but an identifier exists because it's set manually in the entity.
2. `SubjectDatabaseEntityLoader` catches this manually set identifier and loads the entity.
3. Now that the identity is loaded the `save` method thinks nothing changed and doesn't really do anything.

I've tried to see how I can get around this, but I'm afraid that there's a lot going on. Maybe we just need an auto-generated strategy called 'manual' or something unless there's a more elegant way to sort this issue out.

Help appreciated.

### DIFF
--- a/src/persistence/EntityPersistExecutor.ts
+++ b/src/persistence/EntityPersistExecutor.ts
@@ -94,9 +94,9 @@ export class EntityPersistExecutor {
 
                     // load database entities for all subjects we have
                     // next step is to load database entities for all operate subjects
-                    // console.time("loading...");
+                    console.time("loading...");
                     await new SubjectDatabaseEntityLoader(queryRunner, subjects).load(this.mode);
-                    // console.timeEnd("loading...");
+                    console.timeEnd("loading...");
 
                     // console.time("other subjects...");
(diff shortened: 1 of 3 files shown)


src/persistence/EntityPersistExecutor.ts
```diff
@@ -94,9 +94,9 @@ export class EntityPersistExecutor {
 
                     // load database entities for all subjects we have
                     // next step is to load database entities for all operate subjects
-                    // console.time("loading...");
+                    console.time("loading...");
                     await new SubjectDatabaseEntityLoader(queryRunner, subjects).load(this.mode);
```
alfaproject: This is where the identity gets 'reloaded' because the identifier is already set. To be fair, even if the identifier wasn't set, the identity shouldn't be reloaded, because it's a new one. It should just go straight to the INSERT.

imnotjames:
Looks like the issue this was for has been closed - as such, I'll be closing out this PR.  Thanks for the contribution!

PR #3160, opened 2018-11-26
Fix STI types on children in joins

amaranth:
Note: This requires (and includes the commit from) PR #2967 otherwise you don't get the right type in the child's relation anyway.

If an STI child overrides a parent property and uses a different type doing a query on the parent type will skip the relation. This happens because the query's join is built using the parent relation but when mapping to an entity the child's relation is fetched so they don't match. Instead, we now use the relation's `propertyPath` to allow using relations that reference the same property. We also copy the child's relation type into the join to ensure we get the right type in the child as well.


AlexMesser:
thank you for contribution!

PR #3014, opened 2018-10-30
Fix findOne not setting underlying query to limit 1

StrikeForceZero:
Resolves #3013 

not sure if there is a applicable unit test for this or not

PR #3162, opened 2018-11-27
Add dummy order by to make offset/limit work for mssql

amaranth:
SQL Server only supports offset and limit functionality as a part of their ORDER BY syntax. This means if a query doesn't have an order clause we will generate invalid syntax when setting a limit. To avoid this when a limit or offset is requested and no order is given we add an ORDER BY for the first primary key of the table. This shouldn't have any effect on the query performance or the order of the results but the alternative would be a non-functional query so this will always be an improvement anyway.

**Note:** Tests still fail when mssql is enabled for other reasons. This only fixes the issue created in #3014.

### DIFF
--- a/src/query-builder/SelectQueryBuilder.ts
+++ b/src/query-builder/SelectQueryBuilder.ts
@@ -1531,13 +1531,22 @@ export class SelectQueryBuilder<Entity> extends QueryBuilder<Entity> implements 
         }
 
         if (this.connection.driver instanceof SqlServerDriver) {
+            // Due to a limitation in SQL Server's parser implementation it does not support using
+            // OFFSET or FETCH NEXT without an ORDER BY clause being provided. In cases where the
+            // user does not request one we insert a dummy ORDER BY that does nothing and should
+            // have no effect on the query planner or on the order of the results returned.
+            // https://dba.stackexchange.com/a/193799
+            let prefix = "";
+            if ((limit || offset) && Object.keys(this.expressionMap.allOrderBys).length <= 0) {
+                prefix = " ORDER BY (SELECT NULL)";
+            }
 
             if (limit && offset)
-                return " OFFSET " + offset + " ROWS FETCH NEXT " + limit + " ROWS ONLY";
+                return prefix + " OFFSET " + offset + " ROWS FETCH NEXT " + limit + " ROWS ONLY";
             if (limit)
-                return " OFFSET 0 ROWS FETCH NEXT " + limit + " ROWS ONLY";
+                return prefix + " OFFSET 0 ROWS FETCH NEXT " + limit + " ROWS ONLY";
             if (offset)
-                return " OFFSET " + offset + " ROWS";
+                return prefix + " OFFSET " + offset + " ROWS";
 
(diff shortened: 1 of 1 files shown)


amaranth:
Changed in the second commit to always insert an ORDER BY when needed instead of just in the specific instance where EntityManager was adding a limit implicitly.

amaranth:
We're kind of not, they asked for an offset or limit and this is the only way to provide that with SQL Server.

src/query-builder/SelectQueryBuilder.ts
```diff
@@ -1531,13 +1531,22 @@ export class SelectQueryBuilder<Entity> extends QueryBuilder<Entity> implements
         }
 
         if (this.connection.driver instanceof SqlServerDriver) {
+            let segment = "";
+
+            if ((limit || offset) && Object.keys(this.expressionMap.allOrderBys).length === 0) {
+                // MSSQL requires an ORDER BY if you use limits
+                this.addOrderBy(`${this.alias}.${this.expressionMap.mainAlias!.metadata.primaryColumns[0].propertyPath}`);
+                segment = this.createOrderByExpression();
```
pleerock: this confuses code order a bit, maybe better to move this logic on top before query building, something like if its sql server and no order buys but limit or offset presents we addOrderBy.

vlapo:
Why so complicated solution? Simple add "do nothing select" `SELECT NULL` as mentioned on [stackexchange](https://dba.stackexchange.com/questions/193783/sql-server-error-invalid-usage-of-the-option-first-in-the-fetch-statement?answertab=votes#tab-top) 
```typescript
let offsetPrefix = "";
f ((limit || offset) && Object.keys(this.expressionMap.allOrderBys).length <= 0)
     offsetPrefix = " ORDER BY (SELECT NULL)";
}
if (limit && offset)
     return offsetPrefix + " OFFSET " + offset + " ROWS FETCH NEXT " + limit + " ROWS ONLY";
if (limit)
     return offsetPrefix + " OFFSET 0 ROWS FETCH NEXT " + limit + " ROWS ONLY";
if (offset)
     return offsetPrefix + " OFFSET " + offset + " ROWS";
```
Plus add comments why this "hack" is there with some links with explanation.

Kononnable:
From the query performance perspective `ORDER BY (SELECT NULL)` looks better. Setting `ORDER BY` on primary key column would almost always force MSSQL Server to use index on primary key(mostly clustered index) while `SELECT NULL` would allow other indices to be used instead(if they're more performant for that query).

rjmacarthy:
Hey guys. 

We are experiencing this issue, what needs to be done in order to get the PR merged?  

Thanks!

pleerock:
it should be re-implemented with all comments from us applied.

amaranth:
Changed to the method suggested by @vlapo, added a comment explaining what we're doing and why, and rebased and squashed the commits. I linked to the StackExchange post instead of somewhere more official because Microsoft shut down their Connect platform so links are dead.

pleerock:
Looks good now :) Thank you!

rjmacarthy:
Awesome thanks guys.

PR #3168, opened 2018-11-27

### LINKED ISSUE #1139: @PrimaryGeneratedColumn("uuid") doesn't work with mysql driver
I use MySQL driver.  I get some exception after running my app.

> Connection.js:103:28)
    at emitOne (events.js:116:13)
    at Socket.emit (events.js:211:7)
  code: 'ER_TOO_LONG_KEY',
  errno: 1071,
  sqlMessage: 'Specified key was too long; max key length is 767 bytes',

The logger says the serial is VARCHAR(255). Is this a bug or something? Also this is not optimal for uuid.

huanz:


src/driver/mysql/MysqlDriver.ts
```diff
@@ -509,13 +509,17 @@ export class MysqlDriver implements Driver {
         if (column.length)
             return column.length.toString();
 
+        /**
+         * fix https://github.com/typeorm/typeorm/issues/1139
+         */
+        if (column.generationStrategy === 'uuid')
```
havenchyk: could you please provide a test for this change? Also, it would be nice if you could add a record in changelog and place this PR against `next` branch

Issue #3159, opened 2018-11-26
0.2.9 CONTAINS BREAKING CHANGES!!!


jd-carroll:
The recent 0.2.9 release has a single changelog entry of: 
```
* UpdateEvent now returns with contains updatedColumns and updatedRelations
```

However, there are a great many set of changes between the two versions.  So I am **very confident** that 0.2.9 contains changes that were not supposed to be released.

**Breaking Change:**
In 0.2.8, the `DriverUtils` will parse any URL provided and merge the contents from the URL with the config strategy using the code:
```javascript
  return Object.assign(options, urlDriverOptions);
```

In 0.2.9, the `DriverUtils` changes this approach to merge the parsed options to:
```javascript
  return Object.assign(urlDriverOptions, options);
```

The difference is that if properties like `username` and `password` are specified in the configuration object directly **and not in the url**, they will now be `undefined`.

Whether or not this was intentional, it at least needs to be called out more explicitly in the changelog

havenchyk:
@jd-carroll thank you! Would you mind to provide a PR for that issue you found?

jd-carroll:
@havenchyk Before submitting a PR, which I am happy to do, I would like to first confirm what is and is not supposed to be in the 0.2.9 release.  Since there isn't a `0.2.8` or `0.2.9` tag, I would recommend npm install'ing to two empty directories and performing a diff on the file system.

havenchyk:
@jussikinnula could you please clarify this change?

jussikinnula:
> @jussikinnula could you please clarify this change?

Sorry of the confusion. I did had to double check, and the fix works on my test application with both ConnectionOptions `url` set and with regular parameters (e.g. `username`, `password`, etc). If providing both, then the parameters are overwritten those in the URL, e.g.:

```
const options: ConnectionOptions = {
  url: 'postgres://user1:pass2@localhost:5432/db1',
  type: 'mysql',
  host: 'localhost',
  port: 3306,
  username: 'user1',
  password: 'user2',
  database: 'db2',    
}
```

This will get end result of `type` being `mysql`, but rest of the parameters are extracted from url. I guess this is undesirable?

Anyway, without the change `return Object.assign(urlDriverOptions, options);` none of the URL parameters will get through.

A working test application is here: https://github.com/jussikinnula/react-typeorm-todo

[...]

PR #3173, opened 2018-11-28
Change so that parameters from "url" connection option can be overridden by specific connection options

jussikinnula:
Based on discussion in https://github.com/typeorm/typeorm/issues/3159, it would be desirable to allow users to override parameters set in *url* connection option (only in mysql/mariadb, postgres, mssql and mongodb).

The current implementation prioritises `url` over the other parameters.

For example...

```
const options = {
  url: 'postgres://localhost:5432',
  username: 'myuser',
  password: 'mypass',
  ...
};```

...would get initial options from url, and then set username and password as defined in options.

### DIFF
--- a/docs/connection-options.md
+++ b/docs/connection-options.md
@@ -103,7 +103,7 @@ Read more about caching [here](./caching.md).
 
 ## `mysql` / `mariadb` connection options
 
-* `url` - Connection url where perform connection to.
+* `url` - Connection url where perform connection to. Please note that other connection options will override parameters set from url.
 
 * `host` - Database host.
 
@@ -162,7 +162,7 @@ See [SSL options](https://github.com/mysqljs/mysql#ssl-options).
 
 ## `postgres` connection options
 
-* `url` - Connection url where perform connection to.
+* `url` - Connection url where perform connection to. Please note that other connection options will override parameters set from url.
 
 * `host` - Database host.
 
@@ -198,7 +198,7 @@ See [SSL options](https://github.com/mysqljs/mysql#ssl-options).
 
 ## `mssql` connection options
 
-* `url` - Connection url where perform connection to.
+* `url` - Connection url where perform connection to. Please note that other connection options will override parameters set from url.
 
 * `host` - Database host.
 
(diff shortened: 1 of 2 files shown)


havenchyk:
Ah, since it's a breaking change again, should it be opened against next @pleerock @Kononnable ?

pleerock:
@havenchyk its unclear when next is going to be released.

Since we already did a breaking change for this url stuff probably its okay to merge this one. We'll need to add notes in the changelog.

marines:
This PR breaks the previous fix #2843 for TYPEORM_URL. `Object.assign(urlDriverOptions, options)` overwrites all the credentials taken from the URL with undefineds if only the URL is being passed.
@pleerock @jussikinnula

jussikinnula:
@marines, it should work so that single parameters come first, then larameters from TYPEORM_URL.

We also had a new PR to allow using only TYPEORM_URL lately (handy with typeorm cli).

We did had regression before this PR, which was TYPEORM_URL overriding single parameters - which after discussed we found undesirable.

Do you think something still should be changed or have you actually had problems with new versions?

marines:
I'm fine with single parameters taking precedence over the URL but currently TYPEORM_URL can't be used to define connection string (as of 0.2.12-rc2). I will create a new issue if you think we should move the conversation from here.

marines:
Unsurprisingly it doesn't work with TYPEORM_CONNECTION defined. If you don't provide single parameters then those empty values overwrite the actual credentials provided by TYPEORM_URL while they should be ignored if empty (as I mentioned in my first comment).

jussikinnula:
@marines, you were right. There was still an issue with the Object.assign(). It's now fixed in https://github.com/typeorm/typeorm/pull/3442. Sorry about that. I simplified the structure now even more, so it should clearly work as needed.

@pleerock, we should pick https://github.com/typeorm/typeorm/pull/3442 also for the upcoming next RC.

marines:
No problem at all! I didn't expect it to be fixed so quickly. Thanks a lot!

PR #3175, opened 2018-11-28
Add support for or query to find conditions

### LINKED ISSUE #2929: How to filter objects with OR query
**Issue type:**
[ ✓ ] question
[ ✓  ] feature request
[ ✓  ] documentation issue

**Database system/driver:**
[ ✓ ] `postgres`

**TypeORM version:**
[ ✓ ] `latest`

I want find results in my repo where firstname is like 'john' **OR** lastname is like 'doe'
but findOptions where clause treat it as **AND**.

What I tried :

```
let results = await userRepo.find({
  where : {
    firstname : Like('%John%'),
    lastname : Like('%Doe%'),
  }
});
```

What I expect of this :

```
let results = await userRepo.find({
  where : {
    firstname : Like('%John%'),
    lastname : Or(Like('%Doe%')),
  }
});
```
Any help on how can I use 'OR' within **where** object?

BrainMaestro:
Fix #2929 

Hi I took a stab at this, but the naming and implementation might not be the best. Any pointers would be nice. Also, I'm not 100% sure about the feature because order matters which can be tricky. There's a test showing this


pleerock:
Hey, thank you for PR. I provided answer in #2929 please check it.

BrainMaestro:
Oh nice. If that works, then this can be closed

pleerock:
Not sure about master but in next it works for sure. Thank you for your efforts!

PR #3176, opened 2018-11-28
Fix #3009 user newer extension for postgress UUID gen.

### LINKED ISSUE #3009: PostgreSQL UUID

gintsgints:
New extension does not support Old postgres versions: 9.2 / 9.1 / 9.0 / 8.4 / 8.3

As it is braking change, it goes to next branch.

### DIFF
--- a/src/driver/postgres/PostgresDriver.ts
+++ b/src/driver/postgres/PostgresDriver.ts
@@ -299,9 +299,9 @@ export class PostgresDriver implements Driver {
                         if (err) return fail(err);
                         if (hasUuidColumns)
                             try {
-                                await this.executeQuery(connection, `CREATE EXTENSION IF NOT EXISTS "uuid-ossp"`);
+                                await this.executeQuery(connection, `CREATE EXTENSION IF NOT EXISTS "pgcrypto"`);
                             } catch (_) {
-                                logger.log("warn", "At least one of the entities has uuid column, but the 'uuid-ossp' extension cannot be installed automatically. Please install it manually using superuser rights");
+                                logger.log("warn", "At least one of the entities has uuid column, but the 'pgcrypto' extension cannot be installed automatically. Please install it manually using superuser rights");
                             }
                         if (hasCitextColumns)
                             try {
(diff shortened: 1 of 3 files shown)


pleerock:
We need to add a note in the changelog. Also, maybe its worth to support both syntaxes (im not thinking about users who are using too old postgres version, Im thinking about users who use any postgres version but already have used this uuid type, what is going to happen with their data). If they loose their data its better to left old extension support since its easy to do and just a few lines of code, and we can add some option, like `legacyUiidExtension`

donifer:
Would love to see this functionality. Currently in the Kubernetes world the Bitnami postgres image (which is the helm chart de-facto image) doesn't support `uuid-ossp` making it impossible to use with typeorm.

pleerock:
I think best if we support both extension and add postgres connection option `uuidExtension: "pgcrypto"|"uuid-ossp"` defaulted to "uuid-ossp". This way everyone can used what they need without any breaking changes. In the future versions we'll switch default to "pgcrypto"

pleerock:
closed in favor of #3537

PR #3190, opened 2018-12-03
Add flag to enable ON CONFLICT support for Sqlite.

saboya:
Sqlite `ON CONFLICT` clause is supported since `sqlite` `3.24.0`. This PR adds a flag to enable explicit support for it, to avoid breaking changes, and preserves the current behavior otherwise.

It might be necessary to build `sqlite3` from source since precompiled binaries from npm apparently are not build with a recent enough `sqlite` lib.

Tests were added for the driver option and `ON CONFLICT` functionality.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,6 +17,7 @@ before_script:
   - sudo service postgresql stop
   - docker-compose up -d
   - cp ormconfig.travis.json ormconfig.json
+  - npm install sqlite3 --build-from-source
 
 script:
   - npm test
--- a/src/query-builder/InsertQueryBuilder.ts
+++ b/src/query-builder/InsertQueryBuilder.ts
@@ -16,6 +16,7 @@ import {SqljsDriver} from "../driver/sqljs/SqljsDriver";
 import {BroadcasterResult} from "../subscriber/BroadcasterResult";
 import {EntitySchema} from "../";
 import {OracleDriver} from "../driver/oracle/OracleDriver";
+import {SqliteDriver} from "../driver/sqlite/SqliteDriver";
 
 /**
  * Allows to build complex sql queries in a fashion way and execute those queries.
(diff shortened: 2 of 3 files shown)


src/driver/sqlite/SqliteConnectionOptions.ts
```diff
@@ -20,4 +20,8 @@ export interface SqliteConnectionOptions extends BaseConnectionOptions {
      */
     readonly key?: string;
 
-}
\ No newline at end of file
+    /**
+     * Enable ON CONFLICT clauses. (SQLITE >= v3.24.0)
+     */
+    readonly conflictSupport?: boolean;
```
pleerock: Think we don't need this flag. If users don't specify `this.expressionMap.onIgnore` or `this.expressionMap.onConflict` we don't break anything, if they specify, well they know what they are using

saboya: My rationale behind the flag was maintaining the current behavior and avoid breaking changes. But I do agree that users should know what they are using.

I'll remove the flag.

saboya: @pleerock Force-pushed the changes without the flag.

.travis.yml
```diff
@@ -17,6 +17,7 @@ before_script:
   - sudo service postgresql stop
   - docker-compose up -d
   - cp ormconfig.travis.json ormconfig.json
+  - npm install sqlite3 --build-from-source
```
havenchyk: why sqlite? shouldn't docker-compose.yml handle this?

saboya: `sqlite` tests are not run inside a `docker` container, so `docker-compose` isn't used for it.

This command has to be run because the pre-built versions of `sqlite3` were not compiled with a recent enough `sqlite` lib, as in, `>= 3.24.0`. When you install `sqlite3` from source, it builds using the bundled version that goes with it, `3.24.0`. or with a custom version that can be specified with the `--sqlite` flag.

havenchyk: I missed it installs npm package. thanks for the explanation!

havenchyk:
LGTM

pleerock:
Thank you very much!

PR #3204, opened 2018-12-06
Fix #2288 - Increment/decrement accept string as input value

### LINKED ISSUE #2288: Increment and decrement should take string also as a value
**Issue type:**

[ ] question
[x] bug report
[ ] feature request
[ ] documentation issue

**Database system/driver:**

[ ] `cordova`
[ ] `mongodb`
[ ] `mssql`
[ ] `mysql` / `mariadb`
[ ] `oracle`
[x] `postgres`
[ ] `sqlite`
[ ] `sqljs`
[ ] `react-native`

**TypeORM version:**

[ ] `latest`
[ ] `@next`
[x] `0.2.6` (or put your version here)

Because of the limit of javascript, typeorm is using a string for big integer. So, to use `increment` and `decrement` with big integer, we need to provide a string as a value.

And, I think we need a validator to check if the given string is a number because it cannot be handled by Typescript.

https://github.com/typeorm/typeorm/blob/master/src/entity-manager/EntityManager.ts#L652

If you think that we have to fix this, I'll create a PR for it.

vlapo:
- Fix #2288
 - add tests with bigint

### DIFF
--- a/src/entity-manager/EntityManager.ts
+++ b/src/entity-manager/EntityManager.ts
@@ -606,7 +606,7 @@ export class EntityManager {
         findOptions = {
             ...(findOptions || {}),
             take: 1,
-        }
+        };
 
         FindOptionsUtils.applyOptionsToQueryBuilder(qb, findOptions);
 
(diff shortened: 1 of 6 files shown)


pleerock:
Awesome! Thank you!

Issue #2693, opened 2018-08-21
Option to run migrations in 1-transaction-per-migration mode


alexturek:
**Issue type:**

[ ] question
[ ] bug report
[x] feature request
[ ] documentation issue

**Database system/driver:**

[ ] `cordova`
[ ] `mongodb`
[ ] `mssql`
[ ] `mysql` / `mariadb`
[ ] `oracle`
[x] `postgres`
[ ] `sqlite`
[ ] `sqljs`
[ ] `react-native`
[ ] `expo`

**TypeORM version:**

[x] `latest` (0.2.7)

**Steps to reproduce or a small repository showing the problem:**

I just had a specific issue when running migrations (on my local DB, not against a production system), where running all migrations in a single transaction caused them all to fail. What I ran into was that I was running

1. <early timestamp> create table `foo`
2. <later timestamp> `insert into foo values ...`

The CLI output showed the create-table migration succeeding, and then insert-data failed with "no such relationship `tablename`"

I was able to fix this with `-t=false` (from #1408). But my ideal solution would run these migrations transactionally, just one transaction-per-migration. Some of this is my expectation from other ORM libraries I've used, where migrations always happen outside a transaction (bad), but at least if the last migration fails, the earlier ones have succeeded.

So my suggestion (not sure of the interface) is to add this third option:

1. (default, existing) Run all migrations inside a transaction. I assume this really only works for schema-change-in-transaction databases.
2. (existing) Run all migrations outside of a transaction. Just no `BEGIN` statements that the user doesn't specify. Solved with `typeorm migration:run -t=false`
3. (new) Run each migration inside a transaction. This will avoid partially applied migrations, but also let us lock in progress, so that it doesn't matter if users run all migrations at once, or one every day.

I can see the arguments both ways here. I'm just wondering if you'd be amenable to adding the third option. Also I'm not sure of the best interface, since it seems like `-t` is already a boolean flag, and I'm not sure I'd want a trilean option 😄

Diluka:
> (new) Run each migration inside a transaction. This will avoid partially applied migrations, but also let us lock in progress, so that it doesn't matter if users run all migrations at once, or one every day.

I have encountered this situation
If defined a key too long the migration will break database structure

kodayashi:
I just hit this too. In short I was migrating schema in migration 1 and seeding in migration 2. Running them one at a time worked, but `-t=false` allowed me to run them all in one `migration:run`. Nice find @alexturek.

[...]

PR #3208, opened 2018-12-06
[WIP] Add new mode to wrap each migration in transaction

garbles:
Addresses https://github.com/typeorm/typeorm/issues/2693

Creates a new, backwards compatible mode for running migrations each in their own transaction. This behaviour is what I think most expect as default from an ORM, so running all migrations in a single transaction seems odd. Using a single transaction also does not allow you to do certain things (in Postgres at least) if you expect to be able to run all migrations at once.

For example, if you have the following two migrations, you can **never** run all of from scratch.

```ts
import { MigrationInterface, QueryRunner } from "typeorm";

export class CreateUuidExtension1544044606093 implements MigrationInterface {
  public up(queryRunner: QueryRunner): Promise<any> {
    return queryRunner.query(`CREATE EXTENSION IF NOT EXISTS "uuid-ossp";`);
  }

  public down(queryRunner: QueryRunner): Promise<any> {
    return queryRunner.query('DROP EXTENSION "uuid-ossp"');
  }
}
```

```ts
import { MigrationInterface, QueryRunner, Table } from "typeorm";

export class CreateUsers1543965157399 implements MigrationInterface {
  public up(queryRunner: QueryRunner): Promise<any> {
    return queryRunner.createTable(
      new Table({
        name: "users",
        columns: [
          { name: "id", type: "uuid", isPrimary: true, default: "uuid_generate_v4()" },
        ]
      })
    );
  }

  public down(queryRunner: QueryRunner): Promise<any> {
    return queryRunner.dropTable("users");
  }
}
```

The error is `function uuid_generate_v4() does not exist` because extensions cannot be referenced from the same transaction which they are created.

Changes to the `-t` flag to accept: `all` (run all migrations in a single transaction), `each` (run each migration in a separate transaction), `none` (run all migrations without transactions) or `false` (same as none).

TODO:
- [x] Add new tests for 'each' mode
- [ ] Clean up code

### DIFF
--- a/src/connection/Connection.ts
+++ b/src/connection/Connection.ts
@@ -272,12 +272,12 @@ export class Connection {
      * Runs all pending migrations.
      * Can be used only after connection to the database is established.
      */
-    async runMigrations(options?: { transactionMode?: 'all' | 'none' | 'each' }): Promise<Migration[]> {
+    async runMigrations(options?: { transactionMode?: "all" | "none" | "each" }): Promise<Migration[]> {
         if (!this.isConnected)
             throw new CannotExecuteNotConnectedError(this.name);
 
         const migrationExecutor = new MigrationExecutor(this);
-        migrationExecutor.transactionMode = (options && options.transactionMode) || 'all';
+        migrationExecutor.transactionMode = (options && options.transactionMode) || "all";
 
         const successMigrations = await migrationExecutor.executePendingMigrations();
         return successMigrations;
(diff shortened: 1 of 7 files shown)


src/connection/Connection.ts
```diff
@@ -272,14 +272,13 @@ export class Connection {
      * Runs all pending migrations.
      * Can be used only after connection to the database is established.
      */
-    async runMigrations(options?: { transaction?: boolean }): Promise<Migration[]> {
+    async runMigrations(options?: { transactionMode?: "all" | "none" | "each" }): Promise<Migration[]> {
```
pleerock: I think we can simplify it and just call it `transaction`

MatthiasKunnen: I would prefer `transactionMode` or a variant thereof in this case. When a parameter is called `transaction` I would expect to pass an object of type `Transaction`.

garbles:
@mordof hey! Sorry I haven't had time to finish it. Please feel free to take on this PR!

PR #3209, opened 2018-12-07
Add .editorconfig to reduce formatting issues

havenchyk:


.editorconfig
```diff
@@ -0,0 +1,12 @@
...
+# top-most EditorConfig file
+root = true
+
+# Unix-style newlines with a newline ending every file
+[*]
+end_of_line = lf
+insert_final_newline = true
+charset = utf-8
+indent_style = space
+indent_size = 2
```
pleerock: are you sure it completely follows our code style? I think we have indent_size set to 4 currently, right?

havenchyk: I'll re-check :)

Kononnable:
We could also think about adding `trim_trailing_whitespace` rule.

havenchyk:
@Kononnable added `trim_trailing_whitespace`

pleerock:
Thank you!

Issue #3200, opened 2018-12-05
No pending migrations - needs some guidance.


pallymore:
**Issue type:**

[x] question
[x] bug report
[ ] feature request
[ ] documentation issue

**Database system/driver:**

[ ] `cordova`
[ ] `mongodb`
[ ] `mssql`
[ ] `mysql` / `mariadb`
[ ] `oracle`
[x] `postgres`
[ ] `sqlite`
[ ] `sqljs`
[ ] `react-native`
[ ] `expo`

**TypeORM version:**

[x] `latest`
[ ] `@next`
[ ] `0.x.x` (or put your version here)

**Steps to reproduce or a small repository showing the problem:**

So I have a small app that was working fine until today `typeporm` doesn't seem to find the proper migrations. I need some guidance on how to debug my configurations.

I'm using `awesome-typescript-loader` (with default settings) to parse all migration files to `db/migration/[name].js`- the [name] part is the same as the ts files.

The app is packed into a Docker image, with the working directory set to the app's root folder.

predefined task `db:migrate` 
![image](https://user-images.githubusercontent.com/2566530/49489100-8e4ac080-f817-11e8-8fe7-f9a5d1eb49b8.png)

When I run it:
![image](https://user-images.githubusercontent.com/2566530/49489121-9b67af80-f817-11e8-956a-9766ddf78992.png)

Now when I check what's in migrations table 
![image](https://user-images.githubusercontent.com/2566530/49489143-ad495280-f817-11e8-9ad8-0bea0966a036.png)

those 3 are from my pervious operations.

When I check what files are in the `db/migration`:
![image](https://user-images.githubusercontent.com/2566530/49489169-c7833080-f817-11e8-9078-522edeeae3e5.png)

And this is what my `ormconfig.js` looks like:
![image](https://user-images.githubusercontent.com/2566530/49489207-ef729400-f817-11e8-8edf-011dab081021.png)

Migration works when I'm in development mode (with the `ts` files) - but doesn't work with `js` files? What I might be doing wrong here?

pallymore:
tried to create a migration file with `typeorm migration:create` in the container, it seems to be able to find the `ts` file but not the `js` ones - so, webpack issue?

havenchyk:
@pallymore please, create a small repo to reproduce the problem

pallymore:
Hi, it turns out to be an issue with webpack 4 & UglifyJS. 

Adding those settings works for me now:
[...]

PR #3214, opened 2018-12-08
Adds some information about transpiling migration files with Webpack

pallymore:
Related: https://github.com/typeorm/typeorm/issues/3200#issuecomment-444744508


pleerock:
Thank you!

PR #3215, opened 2018-12-08
Issue 2856: feat: added support for IORedis in TypeORM caching

HussainAliAkbar:


Kononnable:
Most of the code is same for redis and IORedis. Would be nicer if we could reuse that code instead of copy pasting it.
As for the tests: I think same code that test current redis cache mechanism should test ioredis because it's a substitute ... but to be honest I'm not sure we currently have any tests for redis caching.

HussainAliAkbar:
You are right. I'll see what can be done about reusing the code and update the PR accordingly. 

For the tests, i'll double check to see if tests exist or not.

ignacioblit:
Excellent PR, I was much in need for the same and about to work on something of this sort. 

Does it support clustering? 

Do we have any news of it being accepted?

Thanks!

acarvallo-lc:
Agree, cluster support would be great to have. I'm developing a project that needs it and the default Redis support is not enough :(

IORedis is much more feature-rich than the default package and it has better maintenance from what I tried in previous projects and would be great to have in the ORM queries.

pleerock:
I think this can be created against `master`?

ignacioblit:
That sounds awesome @pleerock, does that mean you will move ahead with the merge?

HussainAliAkbar:
exactly. IORedis is much more feature rich than node_redis. Lets get this one merged first and then we can brainstorm on what other features of IORedis can be leveraged in TypeORM.

sure @pleerock, i'll make this PR against master. i also need to work on refactoring the code before i can do that. I have figured out a way to merge both the classes into a single one. I'll try to get it done in the next 2-3 days.

HussainAliAkbar:
I have created a new PR against the master branch: https://github.com/typeorm/typeorm/pull/3289. Should we close this one then?

havenchyk:
@HussainAliAkbar yes

PR #3220, opened 2018-12-09
Add more specific types for cli commands

havenchyk:
- Allow dropped migrations CLI prefix
- Add a note that old style is deprecated


src/cli.ts
```diff
@@ -1,5 +1,6 @@
 #!/usr/bin/env node
 import "reflect-metadata";
+import * as yargs from "yargs";
```
pleerock: do yargs include typings now?

havenchyk: @pleerock https://github.com/typeorm/typeorm/pull/3220/files#diff-b9cfc7f2cdf78a7f4b91a753d10865a2R45 yes

havenchyk: https://github.com/DefinitelyTyped/DefinitelyTyped/blob/aad7b771590a78176ee5beed711da512f54d7e15/types/yargs/index.d.ts

pleerock: ah sorry didn't saw that

pleerock:
Thank you!

pleerock:
wait, so the reason why I don't like to include something that doesn't have its own package is that we have to either include those additional typings in dependencies, either force people to install them. Both approaches are bad.

So, right now I see `import * as yargs from "yargs"` in commands (definition files), does it mean when user installs typeorm compiler will give him errors about missing "yargs" dependency? Or does it ignore them in node_modules? (I don't remember exactly but something tells me it first)

havenchyk:
@pleerock so, I finally checked - made a build from the latest master branch and copied it to another working example, then made `npx tsc` - no errors. Could you please release another `rc` or `next` tag from latest master to install package from npm?

pleerock:
released in `0.2.10-rc.2`

havenchyk:
@pleerock just installed this version to the simple local project with TS, works without any problem

PR #3223, opened 2018-12-10
Add linting ci step. Fix current linting issues

havenchyk:
As far as I see, `lint-staged` specified in package.json doesn't run (there is no pre-commit hook that triggers it), so it makes sense to add an additional CI for linting


pleerock:
Thank you!

PR #3230, opened 2018-12-11
Add metadata to subscriber events

milutinovici:
When creating a general subscriber, if prototype of an inserted object is not set, it is impossible to get metadata. Typeorm has this information from used repository, but doesn't pass it here. This pull request just passed down this information.


pleerock:
Thank you for the contribution.

Its highly recommended to add tests because next time somebody change this code he can accidentally remove this metadata and break this thing.

PR #3231, opened 2018-12-11
Uses defineProperty so that __has_ and __promise_ variables are not enumerable

joelgallant:
Seeing these properties when JSON.stringifying and like functions is jarring, and as far as I can tell there's no need for these properties to be enumerable

### DIFF
--- a/src/query-builder/RelationLoader.ts
+++ b/src/query-builder/RelationLoader.ts
@@ -189,6 +189,7 @@ export class RelationLoader {
         const promiseIndex = "__promise_" + relation.propertyName + "__"; // in what property of the entity loading promise will be stored
         const resolveIndex = "__has_" + relation.propertyName + "__"; // indicates if relation data already was loaded or not, we need this flag if loaded data is empty
 
+        // using defineProperty in these so that private properties are enumerable: false
         Object.defineProperty(entity, relation.propertyName, {
             get: function() {
                 if (this[resolveIndex] === true || this[dataIndex]) // if related data already was loaded then simply return it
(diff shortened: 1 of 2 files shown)


pleerock:
Hey, thanks for the contribution.

Can you please take a look why tests are failing?
Also can you please add test for the change you made?

joelgallant:
I can't really tell what test is failing from the logs. I'll try and add a test for this change, though I'm unfamiliar with the setup.

joelgallant:
As far as I can tell, `test/functional/relations/lazy-relations/basic-lazy-relation` should act as a regression test on this change, since it doesn't introduce any changed or new behavior.

Running the tests locally works (though I like the docs for dev setup could be improved - I'll try and follow up with a different PR for that).

amaranth:
You've made `resolveIndex` read-only but the setter tries to write to it. Making that writable again makes the tests pass.

```diff
diff --git a/src/query-builder/RelationLoader.ts b/src/query-builder/RelationLoader.ts
index da4705bf..9abadbe7 100644
--- a/src/query-builder/RelationLoader.ts
+++ b/src/query-builder/RelationLoader.ts
@@ -206,7 +206,10 @@ export class RelationLoader {
                         this[dataIndex] = result;

                         // ensures enumerable: false
-                        Object.defineProperty(this, resolveIndex, { value: true });
+                        Object.defineProperty(this, resolveIndex, {
+                            value: true,
+                            writable: true,
+                        });

                         delete this[promiseIndex];
                         return this[dataIndex];

```

joelgallant:
Thanks!

pleerock:
Can you please add test as well? Because without test next time somebody makes a change this functionality can break without tests.

joelgallant:
You got it. Added a test that ensures the properties are not enumerable, which would make reverting this break the test.

pleerock:
Thank you very much!

pleerock:
@joelgallant can you please take a look at #3576 ? This PR definitely broke current behaviour. If we won't hotfix it soon I'll have to revert this PR.

joelgallant:
I'll take a look at this sure

pinguinjkeke:
Any news here?

PR #3234, opened 2018-12-12

huanz:


pleerock:
Tests are failing (formatting issues)

pleerock:
I had to revert your change because they broke CI tests. The reason why it broke that you don't await for result of `synchronize` call, and since you don't await it I guess your test is wrong as well. Please create a new PR and await result of `synchronize` call, not sure maybe you need to add `return` to that block, need to check tooling docs

Issue #3249, opened 2018-12-13
Cannot set migrationsTableName through environment variables


lukejagodzinski:
**Issue type:**

[ ] question
[ ] bug report
[x] feature request
[ ] documentation issue

**TypeORM version:**

[x] `latest`
[ ] `@next`
[ ] `0.x.x` (or put your version here)

I've checked in the source code and it looks like it's not possible to set the `migrationsTableName` through the environment variables. I'm creating a project that connects to the old database where there is already `migrations` table and for the new app I want to create separate migrations table.

pleerock:
feel free to PR it, its easy to add a new one

lukejagodzinski:
Sure I will add that. Is the `MIGRATIONS_TABLE_NAME` good var name?

PR #3254, opened 2018-12-14
Add env variable for the migrationsTableName conn option

lukejagodzinski:
Add a new variable called `TYPEORM_MIGRATIONS_TABLE_NAME` as discussed here https://github.com/typeorm/typeorm/issues/3249

PR does not contain tests for this change as there are no tests for the other env variables


havenchyk:
:+1:

Issue #2333, opened 2018-06-13
datetime column showing changed on every schema:sync run


ktwbc:
**Issue type:**

[ ] question
[ X] bug report
[ ] feature request
[ ] documentation issue

**Database system/driver:**

[ ] `cordova`
[ ] `mongodb`
[ ] `mssql`
[X ] `mysql` / `mariadb`
[ ] `oracle`
[ ] `postgres`
[ ] `sqlite`
[ ] `sqljs`
[ ] `react-native`

**TypeORM version:**

[ ] `latest`
[ ] `@next`
[ X] `0.2.6` (or put your version here)

**Steps to reproduce or a small repository showing the problem:**
1. Create new entity file:

```
import { Entity, PrimaryGeneratedColumn, Column } from 'typeorm';


@Entity('test')
export class Test {

  @PrimaryGeneratedColumn()
  id: number;

  @Column({type: 'datetime', nullable: true, default: null})
  publish_date: Date;

}

```
2. Run command schema:sync
Initial run of command shows:
```
CREATE TABLE `test` (`id` int NOT NULL AUTO_INCREMENT, `publish_date` datetime NULL, PRIMARY KEY (`id`)) ENGINE=InnoDB
```
3. Confirm table has been created
4. Run schema:sync again

Run will show an alter table erroneously (and will repeat on every run)
```
ALTER TABLE `test` CHANGE `publish_date` `publish_date` datetime NULL
```
<!--
    To answer those questions you need to put "x" inside the square brackets, for example:
    [x] `mysql` 5.6.36 on Mac (installed with homebrew)
    [ ] `postgres`

    Also, please format your code properly (by taking code blocks into ```ts .... ```)
!>

iamwindy:
try using
`
@Column({ precision: 6,type: "timestamp", nullable: true })
  public deletedAt?: Date;
`
for the meantime, it will not fetch changes for now

ktwbc:
still doing it, just different statement when it runs each time but is updating it over and over, even though the db actually updated on the first sync

columns changed in "applicants". updating: publish_date
query: ALTER TABLE `applicants` CHANGE `publish_date` `publish_date` timestamp(6) NULL

ktwbc:
```
{precision: 6, type: 'timestamp', nullable: true}
```
worked without default:null on this variation.

iamwindy:
i know, thats why i told you to use that :D

ManuelMolina97:
Got the same problem but a bit different:

[...]

Issue #609, opened 2017-07-03
Custom precision on CreateDateColumn and UpdateDateColumn


sanlyx:
MySQL 5.6 and later releases provide support for fractional timestamps via the `DATETIME(1)` to `DATETIME(6)` column types. It would be nice for TypeORM to support them on CreateDateColumn and UpdateDateColumn decorators:

```ts
@CreateDateColumn('datetime(6)')  // does not work because default has to be CURRENT_TIMESTAMP(6) instead of CURRENT_TIMESTAMP
createdAt: Date;
```

pleerock:
as of `0.1.0-alpha.29` created date and updated date use `datetime(6)` as data types.

ruimgoncalves:
This breaks support for MySql 5.5 when creating a table since this version does not support specifying precision for DateTime or Timestamp types.

daggett206:
the same problem for mysql 5.5
`You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near \'(6) NOT NULL DEFAULT CURRENT_TIMESTAMP(6)\' at line 1'
`

svicalifornia:
@pleerock Reopen this issue and fix for MySQL 5.5.

pleerock:
in latest `@next` you are able to do following:

```ts
@CreateDateColumn({ precision: 4 /* or any other */, default: () => "CURRENT_TIMESTAMP" })
```

To make this functionality to work on old mysql versions.

svicalifornia:
@pleerock This issue was reopened because the addition of precision caused MySQL 5.5 to fail when creating a table. Since this has now been closed twice, perhaps someone should open a new issue for the MySQL 5.5 compatibility problem.

@ruimgoncalves or @daggett206 Would either of you like to open a new issue for the MySQL 5.5 precision problem?

svicalifornia:
@pleerock Sorry, I'm trying to understand how you have resolved the MySQL 5.5 issue raised by @ruimgoncalves or @daggett206 above.

Last July, you said that the new default for CreateDateColumn and UpdateDateColumn to `DATETIME(6)`. Then they posted that MySQL 5.5 doesn't support any precision for `DATETIME`:

```
You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near \'(6) NOT NULL DEFAULT CURRENT_TIMESTAMP(6)\' at line 1'
```

[...]

PR #3262, opened 2018-12-16
schema sync on columns without precision, width or default (close #2737,#2333)

### LINKED ISSUE #2737: MySQLDriver findChangedColumns (fields: width, precision)
**Issue type:**

[ ] question
[x] bug report
[ ] feature request
[ ] documentation issue

**Database system/driver:**

[ ] `cordova`
[ ] `mongodb`
[ ] `mssql`
[x] `mysql` / `mariadb`
[ ] `oracle`
[ ] `postgres`
[ ] `sqlite`
[ ] `sqljs`
[ ] `react-native`
[ ] `expo`

**TypeORM version:**

[ ] `latest`
[ ] `@next`
[x] `0.2.7` (or put your version here)

MySQL version: 5.7
Typescript version: 3.0.1

Everytime schema:sync is called, typeorm drops all fk and re-create them. It also performs alter table change column to timestamp columns. The problem is when findChangedColumns is called.
[Line](https://github.com/typeorm/typeorm/blob/a4dec02cc59d3219a29c7be0322af2253e1452dc/src/driver/mysql/MysqlDriver.ts#L623)

```ts
console.log("width:", tableColumn.width, columnMetadata.width);
console.log("precision:", tableColumn.precision, columnMetadata.precision);
```

output:

for fk:
```ts
width: 10 undefined
```

for timestamp:
```ts
precision: 0 null
```

Example entity:

```ts
@Entity('assignments')
export class Assignment {

  @PrimaryGeneratedColumn({unsigned: true})
  id: number;

  @CreateDateColumn({precision: null, type: 'timestamp', default: () => 'CURRENT_TIMESTAMP'})
  from: Date;

  @Column({precision: null, type: 'timestamp', default: () => null, nullable: true})
  to: Date;
```

Kononnable:
Fixes syncing problems described in #2737


src/driver/mysql/MysqlDriver.ts
```diff
@@ -640,12 +640,14 @@ export class MysqlDriver implements Driver {
             // console.log("isUnique:", tableColumn.isUnique, this.normalizeIsUnique(columnMetadata));
             // console.log("isGenerated:", tableColumn.isGenerated, columnMetadata.isGenerated);
             // console.log("==========================================");
-
+            const mysqlDefaultLength = 10;
```
Kononnable: Not sure if it is always 10(and why), but I experienced this property filled only on int type with even if it wasn't specified in column options.

pleerock:
Thank you!

is #2333 related somehow?

src/driver/mysql/MysqlDriver.ts
pleerock: fixes in "sync" area are very dangerous without understanding what and why. First, you compare "width" with "length". Second it might be default to `10`, but usually it depends on column type. So its better to figure those things out and find a proper documentation-based solution.

Kononnable:
Yes, it looks like it fixes that issue too.

src/driver/mysql/MysqlDriver.ts
Kononnable: Sorry, I guess I was too tired at the time. Forgot to put [WIP] and named width as length.

Found out why there is 10 - default width declared in `dataTypeDefaults` is decreased by 1 when column is unsigned (it behaves differently with `bigint`). I should be able to fix that tomorrow.

src/driver/mysql/MysqlDriver.ts
```diff
@@ -640,12 +640,13 @@ export class MysqlDriver implements Driver {
...
             // console.log("isGenerated:", tableColumn.isGenerated, columnMetadata.isGenerated);
             // console.log("==========================================");
-
             return tableColumn.name !== columnMetadata.databaseName
                 || tableColumn.type !== this.normalizeType(columnMetadata)
                 || tableColumn.length !== columnMetadata.length
-                || tableColumn.width !== columnMetadata.width
-                || tableColumn.precision !== columnMetadata.precision
+                || (tableColumn.width !== columnMetadata.width
+                    && (tableColumn.width || columnMetadata.width))
```
pleerock: a bit complex expression. Can you please explain it?

src/driver/mysql/MysqlQueryRunner.ts
```diff
@@ -1245,8 +1240,14 @@ export class MysqlQueryRunner extends BaseQueryRunner implements QueryRunner {
                     tableColumn.isPrimary = dbPrimaryKeys.some(dbPrimaryKey => {
                         return this.driver.buildTableName(dbPrimaryKey["TABLE_NAME"], undefined, dbPrimaryKey["TABLE_SCHEMA"]) === tableFullName && dbPrimaryKey["COLUMN_NAME"] === tableColumn.name;
                     });
+
                     tableColumn.zerofill = dbColumn["COLUMN_TYPE"].indexOf("zerofill") !== -1;
                     tableColumn.unsigned = tableColumn.zerofill ? true : dbColumn["COLUMN_TYPE"].indexOf("unsigned") !== -1;
+                    if (this.driver.withWidthColumnTypes.indexOf(tableColumn.type as ColumnType) !== -1) {
```
pleerock: also can you please explain why this was moved?

Kononnable: Default width changes when column is unsigned - so I needed to move it after that property is changed.

src/driver/mysql/MysqlDriver.ts
Kononnable: We strict check if values aren't equals, but column is changed when at least one of the values is defined. For this case undefined and null should be treated as equals(didn't want to use ```==```).

I'm not sure atm which one returned ```undefined``` and which ```null```. I can check that out and find out why there are different if that's necessary.

pleerock: Solution looks a bit complex and it looks like the root of the problem is somewhere else. Maybe we should figure out why its somewhere is undefined and somewhere is null?

Kononnable: I guess first few bugfixes gonna be hard :)
It looks like this is not necessary for width. Precision is more problematic - it accepts nulls(#609 cc789b410eb9a0e962101b22175284da37b5f387). 

In case of MySql driver when that check is executed `tableColumn.precision=0` and `columnMetadata.precision=null`. The question is what `tableColumn.precision` value should be - should we change it always to null when it is 0 on time columns? It is set on line https://github.com/typeorm/typeorm/blob/master/src/driver/mysql/MysqlQueryRunner.ts#L1283 

On other drivers there is comparison between `tableColumn.precision=undefined` and `columnMetadata.precision=null`(because we set it explicitly to null in entity options).

pleerock: Is it possible for precision to be equal to zero? If no then maybe we can replace null with zero. If yes, then maybe we'll use undefined values instead of nulls everywhere ? (except here in this edge-case precision case)

Kononnable: mysql acts a bit strange for my taste:
- timestamp with precision = 0 - creates timestamp without precision
- decimal with precision = 0 - creates decimal with precision 10(default)

I see two options how to solve it:
- when reading database state set column precision to null if db returns 0 - it will break synchronization for time columns with precision set to 0(in entity definition), ppl would have to change precision to null(setting precision to 0 doesn't work on mysql<5.6)
- deprecate null as precision value in column options(metadata) and  when syncing db add columns without passing precision value if precision is set to 0(current implementation for null precision)

pleerock: sounds okay to me. Just make sure you know all the cases when precision can be zero or not for all databases. Schema synchronization is a fragile place, we need to be careful with changes.

Kononnable: Ok, I think I will go with first option - I guess more people use precision = null than precision = 0.
Should it(logic about null precision) be implemented in all of the drivers or just mysql(where that edge case really exists)? In other words do we want `precision:null` to be accepted anywhere or just on old mysql servers where it was intended to do?

Kononnable: @vlapo 
Default values are declared in `dataTypeDefaults` and mostly handled in a different place(internal structures don't contain properties with default values most of the time) so I wouldn't put them to this PR.
Of course writing more tests would be helpful, but it takes a lot of time before making actual changes to the code - I'm not sure we've got enough time since none of us works full-time on the project atm. I think fixing one problem at the time might be a better solution for someone who doesn't really know how much time he/she would be able to contribute next week/month.

Kononnable: I can implement second one if anyone likes it more. I don't really like any of them :)
First one keeps nulls as precision so nobody would have to change theirs entity schemas after upgrading. But leaves a question how null should be handled on other dbs.
Second one introduce 0 as a magic number - depending on column type it will act differently. But in a long run it might be better, because we wouldn't allow null as precision for any(others) db driver.


After a little thought I guess my choice may be due to how such cases are handled at daily job - no depreciation/breaking change policy(which isn't a good policy).

pleerock: I prefer second approach. We just need to make sure if can always omit precision for drivers if its equal to zero (e.g. make sure there is no "0" as a precision value for any driver we have).

Regarding to depreciation/breaking change policy - currently we try to be flexible and at the same time avoid critical breaking changes in master. I don't think we should follow depreciation strategy so much at the moment. It all takes time, and time is limited, so I prefer to stay flexible instead.

Kononnable: Checked every type with precision on mssql, mysql, postgres - either acts like no precision was specified or throws an error when trying to create a column. Sqlite however allows columns with precision = 0.

Kononnable: Since Extended Support for Mysql 5.5 Ends this December do we actually need `precision: null` anymore? I'm not sure if we should support if it is eol. However it would be breaking for folks who still use 5.5 so, probably should be against next if we decide to remove that option.

http://www.oracle.com/us/support/library/lifetime-support-technology-069183.pdf Oracle’s MySQL Releases

pleerock: Let me ask people is slack. But yeah probably we need to drop it then.

pleerock: Okay lets drop 5.5 support.

Kononnable:
To summarize what this PR do:
- removed support for `precision: null` as discussed earlier(breaking, so not sure if should be against master or next)
- fixed default values handling in schema sync(`defualt: null` is same as `default: undefined`)
- fixed width property handling on columns set as unsigned

Kononnable:
Note for the future: changing base branch on PR should be done only after base branch is merged, otherwise we got a lot of mess :)

dantman:
What's the status of this? What needs to be done?

It's starting to get painful having to ignore junk migrations being generated and not 100% knowing that my database schema is correct.

Kononnable:
Implementation is finished. Tests on next were(and still are) failing but it doesn't seem related to this PR.

dantman:
Other PRs seem to have successful builds, perhaps try merging master into the branch and push. That may trigger a successful build.

Kononnable:
To make it work we would need 'green' next branch. Since merging master with next branch have test which are failing, so rebase on any of those commits will also result in failing build. We can't use older next version, because changes are required in CockroachDriver which is new.
However there are 2 PRs open which will fix current tests, so it shouldn't be long now.

kirushi:
Looking forward to this release, removing redundant alters are getting annoying.

You guys are all doing a fantastic job - trying to convince my company to donate to typeorm 👍

Miniplop:
@Kononnable 

Any news on this branch ?

dantman:
@Kononnable Which PRs are the 2 PRs that are needed to fix the tests?

Kononnable:
Sorry guys, it should have been merged long time ago.
Only one unrelated sqlite test and most of cockroachdb tests are failing, so looks ok to merge.

I guess I should check participating notification more often when I'm not checking all notifications for some time.

PR #3268, opened 2018-12-17
Corrected Docs @CreateDate and @UpdateDateColumn

wederer:
Docs were not clear that you can not pass values to properties decorated with @CreateDateColumn or @UpdateDateColumn. These values will be ignored and automatically set.


wederer:
nvm, I think this is somewhat of a "bug". If a value is provided in the entity object the value should be used and not `NOW()`. I will open an issue.

Issue #1675, opened 2018-02-28
Migration should either read .ts or deliver a .js


GitBalou:
This is either a bug or a new feature

[ ] question
[ x] bug report
[x ] feature request
[ ] documentation issue

**Database system/driver:**

[ ] `cordova`
[ ] `mongodb`
[ ] `mssql`
[ ] `mysql` / `mariadb`
[ ] `oracle`
[ x] `postgres`
[ ] `sqlite`
[ ] `sqljs`
[ ] `websql`

**TypeORM version:**

[x ] `latest`
[ ] `@next`
[ ] `0.x.x` (or put your version here)

**Steps to reproduce or a small repository showing the problem:**
`typeorm migrations:generate -n test` this create typecript file test.ts
`typeorm migrations:run` this fails with `Unexpected token import`

**What solves the issue**
Either call typeorm using ts-node, as explained here : https://github.com/typeorm/typeorm/issues/1561
Either transpile .ts file .js using typescript

**Expected behaviour**
Typeorm should be able to read the file it generated, which means : reading ts file or providing .js file

pleerock:
any application you write in typescript you need to compile and run. So its quite normal that you need to compile before you run a typescript code.

eyelidlessness:
The issue as reported is that **TypeORM** is generating files that it then cannot read. I have seen this behavior as well.

pleerock:
what does it mean **it cannot read** ?! TypeORM does not read files, node does. TypeORM generates you a file, you compile it, then your node runs js files.

eyelidlessness:
What I mean is:

1. `migrations:generate` generates migrations, in the form of files
2. It is expected that `migrations:run` runs those migrations, by reading those files and running them - but it does not do that

eyelidlessness:
If the expectation is that `migrations:run` accepts only JS files, then `migrations:generate` should generate only JS files. It is very confusing for the automatic migration generator to generate migrations that cannot be run.

pleerock:
wrong logic. If node can run only js files it does not mean that we need to write js files. are you new in typescript?

eyelidlessness:
It's worth noting that nothing in the documentation about migrations indicates that TS migration files (whether written manually or generated by `migrations:generate`) need to be compiled to be run. The assumption based on the documentation is that `migrations:run` handles TS migrations.

eyelidlessness:
[...]

Issue #371, opened 2017-03-26
Migration cli unexpected token import


adithep:
I wrote everything in typescript.
The application and everything work fine, but as soon as I call 

`typeorm migrations:run`

I get this error

src/entity/User.ts:1
(function (exports, require, module, __filename, __dirname) { import {Entity, Column, PrimaryGeneratedColumn} from "typeorm";
                                                              ^^^^^^
SyntaxError: Unexpected token import
    at Object.exports.runInThisContext (vm.js:78:16)
    at Module._compile (module.js:543:28)
    at Object.Module._extensions..js (module.js:580:10)
    at Module.load (module.js:488:32)
    at tryModuleLoad (module.js:447:12)
    at Function.Module._load (module.js:439:3)
    at Module.require (module.js:498:17)
    at require (internal/module.js:20:19)
    at Function.PlatformTools.load (/usr/local/lib/node_modules/typeorm/platform/PlatformTools.js:26:20)
    at /usr/local/lib/node_modules/typeorm/util/DirectoryExportedClassesLoader.js:29:69
(node:38469) UnhandledPromiseRejectionWarning: Unhandled promise rejection (rejection id: 5): SyntaxError: Unexpected token import
(node:38469) DeprecationWarning: Unhandled promise rejections are deprecated. In the future, promise rejections that are not handled will terminate the Node.js process with a non-zero exit code.

Not sure if it matters but this is my tsconfig.json:

{
  "compilerOptions": {
    "declaration": true,
    "emitDecoratorMetadata": true,
    "experimentalDecorators": true,
    "isolatedModules": false,
    "module": "commonjs",
    "moduleResolution": "node",
    "outDir": "dist",
    "sourceMap": true,
    "sourceRoot": "src",
    "noEmitHelpers": false,
    "target": "es5",
    "typeRoots": [
      "node_modules/@types"
    ],
    "lib": ["es6", "dom"]
  },
  "exclude": [
    "node_modules"
  ]
}

What ever I change in tsconfig.json seems to not effect the error at all.

pleerock:
you node environment tries to load `src/entity/User.ts`, but it should load `.js`. So you either need to run app with ts-node, or set proper configuration for JS FILES in your ormconfig cli section, e.g. `dist/migrations/.*js`

pleerock:
did you resolve your issue?

adithep:
Yes. Thankyou.

[...]

PR #3276, opened 2018-12-18
Note that migrations work with `js` files only

wederer:
Multiple Issues such as #1675 #371 are opened because of this missing documentation.

### DIFF
--- a/docs/migrations.md
+++ b/docs/migrations.md
@@ -141,6 +141,8 @@ export class PostRefactoringTIMESTAMP implements MigrationInterface {
 
 ## Running and reverting migrations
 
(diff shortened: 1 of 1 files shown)


docs/migrations.md
```diff
@@ -141,6 +141,8 @@ export class PostRefactoringTIMESTAMP implements MigrationInterface {
 
 ## Running and reverting migrations
 
+**`typeorm migration:create` and `typeorm migration:generate` will create `.ts` files. The `migration:run` and `migration:revert` commands only work on `.js` files. Thus the typescript files need to be compiled before running the commands.** Alternatively you can use `ts-node` in conjunction with `typeorm` to run `.ts` migration files.
```
havenchyk: @wederer could you please also add an example of using `ts-node`? It should be somewhere in the issues you mentioned

wederer: Will do.

pleerock:
Thank you very much! We definitely were missing this in the docs!

PR #3282, opened 2018-12-19
Fix generated migration file types in README

fan-jin:
Through the cli, migrations generated are in `.ts`, not `.js`.


pleerock:
Migrations generated are in `.ts` but they are supposed to be compiled into `.js` thus paths are correct since without ts-node you can only use `.js` files. 

Thanks for the contribution anyway!

Issue #3258, opened 2018-12-15
No connection options were found in any of configurations file


krivochenko:
**Issue type:**

[ ] question
[*] bug report
[ ] feature request
[ ] documentation issue

**Database system/driver:**

[ ] `cordova`
[ ] `mongodb`
[ ] `mssql`
[*] `mysql` / `mariadb`
[ ] `oracle`
[ ] `postgres`
[ ] `sqlite`
[ ] `sqljs`
[ ] `react-native`
[ ] `expo`

**TypeORM version:**

[ ] `latest`
[ ] `@next`
[*] `0.2.9` (or put your version here)

**Steps to reproduce or a small repository showing the problem:**

1. Add `"typeorm": "ts-node ./node_modules/typeorm/cli.js --config .env"` to `scripts` in package.json. 
2. Run `npm run typeorm -- migration:generate -n MigrationName`. On 0.2.8 it works, but on 0.2.9 it causes error: `Error: No connection options were found in any of configurations file.`

vlapo:
I think your problem is caused by this change https://github.com/typeorm/typeorm/commit/3271ce57971384cd9f076129604ecaa713db92f4#diff-dc3f0fa4dfd1d42b636f6d95a19dc19a

As workaround you can run your command without config argument. (Please let me know if it is working)

But I think this may be changed as bugfix or some change request.

krivochenko:
@vlapo what do you mean? Is it not allowed to store settings of connection in file `.env`?

pleerock:
@steevsachs can you please elaborate?

steevsachs:
Definitely. I’m traveling today but will work on this tonight or tomorrow.

vlapo:
It is allowed @krivochenko, I just saw that connection options reader is using `.env` file if exists and config path is not provided . Sry for misunderstanding.

krivochenko:
@vlapo without `config` argument it works. 

Thank you!

pleerock:
Think we can close this one.

PR #3283, opened 2018-12-19
Support specifying .env as config

steevsachs:
If `baseFilePath` was specified as `.env`, it would not properly load because `.env` was appended to the file path.

Reference typeorm/typeorm#3258


pleerock:
tests are failing...

steevsachs:
@pleerock 

I see, the issue with this and the previously skipped test was that those config files weren't being copied over during build.  I can fix that for the `.js` file byh just making it `.ts`, but the `.env` is different. 

Do you have a suggestion for the best/cleanest way to make a mock `.env` file accessible for testing?

pleerock:
lets just skip those tests with todo comment added then

LeartS:
This broke `ormconfig.env` files, now they are not loaded. `this.baseFilePath` is `ormconfig`, `.env` is not appended anymore and it doesn't work.

PR #3289, opened 2018-12-19
feat: added support for IORedis in TypeORM caching

HussainAliAkbar:


HussainAliAkbar:
@pleerock @Kononnable , this PR is against the master branch. I have also refactored the code a bit so that there is no duplication. almost all the methods were reused except for the connect method which i have handled. 

Do let me know if this works for you guys or not.

src/cache/RedisQueryResultCache.ts
```diff
@@ -38,17 +44,28 @@ export class RedisQueryResultCache implements QueryResultCache {
...
-        } else {
-            this.client = this.redis.createClient();
+        if (this.clientType === "redis") {
+            if (cacheOptions && cacheOptions.options) {
+                this.client = this.redis.createClient(cacheOptions.options);
+            } else {
+                this.client = this.redis.createClient();
+            }
+        } else if (this.clientType === "ioredis") {
+            if (cacheOptions && cacheOptions.options) {this.client = this.redis.createClient();
```
pleerock: is it typo? this.client is being assigned two times?

HussainAliAkbar: indeed it was. i have fixed it.

HussainAliAkbar:
@pleerock , if you are going ahead with the merge, should i update the caching documentation as well?

pleerock:
yes, please :)

src/cache/RedisQueryResultCache.ts
```diff
@@ -23,11 +23,17 @@ export class RedisQueryResultCache implements QueryResultCache {
      */
     protected client: any;
 
+    /**
+     * Type of the Redis Client (redis or ioredis).
+     */
+    protected clientType: string;
```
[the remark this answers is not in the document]
HussainAliAkbar: I have updated the type to ```protected clientType: "redis" | "ioredis";```. However, i have still kept the comment to make it consistent with other protected properties. I thought that if in the future, the type was changed, the comment could be changed with it as well. 

Let me know if the comment still needs to be removed.

HussainAliAkbar:
@pleerock documentation updated.

ignacioblit:
Hey guys, how are you?

Now that i've read the documentation update, and the options. I'm thinking that this new implementation does not support the Cluster options for ioredis is that correct? 

https://github.com/luin/ioredis/blob/master/API.md#Cluster

if thats the case, do you have any plans on implementing this? If not, may I take a look into supporting such feature?

Thanks!

pleerock:
Think we can merge it, thanks for the contribution!

Issue #3169, opened 2018-11-27
Developer environment not working


victorschinaider:
**Issue type:**

[x] question
[ ] bug report
[ ] feature request
[ ] documentation issue

**TypeORM version:**

[x] `latest`
[ ] `@next`
[ ] `0.x.x` (or put your version here)

I'm trying to setup environment of typeorm.
I did what `DEVELOPER.md` file says, but tests are failing in first test (Bulk-save), and are not giving details of what is wrong.

havenchyk:
@victorschinaider what steps did you try and what's the output?

victorschinaider:
Now, i retry steps, and stops at  third test.

First, my environment: 
I'm using `nvm`, with:
- node v8.11.4 (npm v5.6.0)

So, i folllowed exactly what is said in `DEVELOPER.md`.

First i forked TypeORM and clone in my machine. (https://github.com/victorschinaider/typeorm)
 Then i setup upstream. Then i installed the dependencies. After that, i execute `docker-compose up -d` in terminal to setup databases. I copy and renamed `ormconfig.json.dist` to `ormconfig.json`.
So, i finally execute : `npm test`, and the output is:

```
> typeorm@0.2.9 test /home/schinaider/Documentos/typeorm
> rimraf ./build && tsc && mocha --file ./build/compiled/test/utils/test-setup.js --bail --recursive --timeout 30000  ./build/compiled/test



  benchmark > bulk-save > case1
    ✓ testing bulk save of 10.000 objects (742ms)

  benchmark > bulk-save > case2
    ✓ testing bulk save of 10000 objects (1077ms)

  cascades > should insert by cascades from both sides (#57)
writeStringToMemory is deprecated and should not be called! Use stringToUTF8() instead!
tedious deprecated The `IntN` data type is internal and will be removed. node_modules/mssql/lib/tedious.js:63:20
tedious deprecated The `BitN` data type is internal and will be removed. node_modules/mssql/lib/tedious.js:71:40
    ✓ should insert by cascades from owner side (153ms)

  columns > embedded columns
    1) should insert / update / remove entity with embedded correctly


  3 passing (9s)
  1 failing

  1) columns > embedded columns
       should insert / update / remove entity with embedded correctly:
     QueryFailedError: Error: Invalid usage of the option NEXT in the FETCH statement.
      at new QueryFailedError (src/error/QueryFailedError.ts:9:9)
[...]

PR #3292, opened 2018-12-19
docs(develop): Use ormconfig.travis.json in DEVELOPER.md

vlapo:
There are two ormconfig files now `travis` and `dist`. `dist` file is used in `DEVELOPER.md` and mostly have to be same as `travis` file (because of docker). 

I think we should have only one ormconfig file. Mostly because most of new contributors will be use docker and it is necessary run localy same tests as in travis builds.

Issue with non-updated `dist` file is e.g. here https://github.com/typeorm/typeorm/issues/3169


pleerock:
No, I don't think we should do this.

I don't want local configuration to be same as travis configuration. Ideally user needs to run tests on all databases. There are some problems that should be fixed to make it possible to run all databases on travis.

But I understand why you want for users to run the same tests as we are running on travis. Okay, we can do it until we add oracle and mssql into travis. Lets just update `dist` file accordingly, but  not touch `dist` file. I don't like the idea of copying travis file into local docker configuration file.

pleerock:
@vlapo thank you very much

PR #3295, opened 2018-12-20
update mongo 3.1.10 and add mongo transations !

wzquyin:
update mongo 3.1.10 and add mongo transations !
add forceTransation to mongoqueryrunner...
only when u use entitymanager.transation and mongoconnection is support transation     forceTransation wil be true in the transation context.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -47,6 +47,14 @@
         "through2": "^2.0.3"
       }
     },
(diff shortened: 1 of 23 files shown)


pleerock:
Migrations were already implemented in #3304 

@wzquyin if you are interested please provide another PR with clear description of what you propose and clear changes

PR #3298, opened 2018-12-20
Pass referencedColumn's databaseName instead of propertyName

hanpama:
Pass referencedColumn's databaseName instead of propertyName while creating default joinColumnName

When I tried to define chained multiple keys relations like:

```typescript
@Entity()
export class Semester {

    @PrimaryColumn()
    year: number;

    @PrimaryColumn()
    season: string;

}

@Entity()
export class Lecture {

    @ManyToOne(type => Semester, { primary: true })
    semester: Semester;

    @PrimaryColumn()
    course: string;

}

@Entity()
export class Project {

    @ManyToOne(type => Lecture, { primary: true })
    lecture: Lecture;

    @PrimaryColumn()
    studentId: string;

}
```

It throws an error trying to create two `lectureSemester` columns on `project` table.

I expected the join columns for Project.lecture should be `lectureSemesterYear`, `lectureSemesterSeason`

This is caused by the logic creating the default joinColumnName with the referencedColumn's propertyName which can collides.
(https://github.com/typeorm/typeorm/blob/master/src/metadata-builder/RelationJoinColumnBuilder.ts#L130)

### DIFF
--- a/src/metadata-builder/RelationJoinColumnBuilder.ts
+++ b/src/metadata-builder/RelationJoinColumnBuilder.ts
@@ -127,7 +127,7 @@ export class RelationJoinColumnBuilder {
                 return (!joinColumn.referencedColumnName || joinColumn.referencedColumnName === referencedColumn.propertyName) &&
                     !!joinColumn.name;
             });
-            const joinColumnName = joinColumnMetadataArg ? joinColumnMetadataArg.name : this.connection.namingStrategy.joinColumnName(relation.propertyName, referencedColumn.propertyName);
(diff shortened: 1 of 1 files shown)


pleerock:
In `@next` branch those things has changed. Can you test that as well? Also please add a test to your changes.

hanpama:
Cool👍 I will take a look on `@next` branch. I will push some tests about this.

imnotjames:
Is this still a work in progress?  We need tests to verify that this fixes the issue you're expecting it to & ideally a rebase against master so the CI runs.

imnotjames:

PR #3301, opened 2018-12-20
Made CREATE INDEX SQL consistent across dialects

MatthiasKunnen:
Postgres had two spaces before columns.
MySQL, Oracle, and SQL Server had no space between the table and columns.


pleerock:
Thanks :)

PR #3305, opened 2018-12-21
 Update doc to explain how to use the CLI with typescript setup

hkjeffchan:


docs/using-cli.md
```diff
@@ -14,6 +15,34 @@
...
+Install ts-node globally:
+```
+npm install -g ts-node
+```
+
+Add typeorm command under scripts section in package.json
+```
+"script" {
+    ...
+    "typeorm": "ts-node -r tsconfig-paths/register ./node_modules/.bin/typeorm"
```
pleerock: what is `tsconfig-paths/register` here? shouldn't it be installed separately? why do we need users to enforce to register it?

hkjeffchan: Sorry, I am actually using typeorm with nestjs setup and I cannot run properly without registering tsconfig-paths/register (complaining MODULE NOT FOUND)

havenchyk: afaik `tsconfig-paths/register` is a way to allow `ts-node` to work with `paths`, I don't think it's really needed to be here

docs/using-cli.md
```diff
@@ -14,6 +15,34 @@
 * [Clear cache](#clear-cache)
 * [Check version](#check-version)
 
+## Notes on entity files written in typescript
+This CLI tool is written in javascript and to be run on node. If your entity files are in typescript, you will need to transpile them to javascript before using CLI. You may skip this section if you only use javascript.
+
+You may setup ts-node in your project to ease the operation as follows:
+
+Install ts-node globally:
```
havenchyk: I would vote for using `npx ts-node` instead

docs/using-cli.md
havenchyk: btw, shouldn't just `typeorm` work? without `./node_modules/.bin?

hkjeffchan: I just create a fresh project with typeorm only according to the doc. Can confirm that it can run without tsconfig-paths/register.

No idea why it doesn't work on nestjs. Checking...

docs/using-cli.md
pleerock: never used `npx` on my own

docs/using-cli.md
hkjeffchan: Oh... I found out why now.

In a typical setup, ppl will put entity files under 'src/entities' and register the tsconfig-path somewhere else (like app.ts as suggested in doc). So if the entity references another entity with like OneToOne relation, without tsconfig-paths/register, it will fail with "Cannot find module" error.

It will fail in my fresh typeorm only setup if there is one entity referencing another file.

I think you better keep that in your doc to save your time from answering dumb ppl like me from complaining.

Thanks for this great lib again!

hkjeffchan: > btw, shouldn't just `typeorm` work? without `./node_modules/.bin`?

@havenchyk ppl may not install typeorm globally

docs/using-cli.md
havenchyk: It worth trying

docs/using-cli.md
havenchyk: Okay, I think I found the reason. During the start of npm script, npm adds `./node_modules/.bin` to the path, so if you create a script like `"typeorm": "typeorm"`, it will use `typeorm` from the `.bin` folder, but if `"typeorm": "ts-node typeorm"`, npm will not resolve `./node_moduels/.bin/typeorm`. Even better to use `$(npm bin)` - it's an absolute path to `node_modules/.bin`, so next code will work from any place `"typeorm": "npx ts-node $(npm bin)/typeorm"`

wederer:
This should not work on windows.
I am unsure why, but using `./node_modules/.bin/typeorm` on windows fails.
Using `./node_modules/typeorm/cli.js` works on both windows and linux.

pleerock:
yeah windows is weird about paths and executables :)

havenchyk:
I'm fine with current state of PR, as for Windows support, I'd handle it in a separate PR

docs/using-cli.md
FrankFang: What if I use babel instead of ts-node?

PR #3308, opened 2018-12-21
ts-node migration example does not work on windows

wederer:
`ts-node ./node_modules/.bin/typeorm migration:run` does not work on windows
`ts-node ./node_modules/typeorm/cli.js migration:run` works on windows and on linux

### DIFF
--- a/docs/migrations.md
+++ b/docs/migrations.md
@@ -151,7 +151,7 @@ typeorm migration:run
 
 Example with `ts-node`:
 ```
-ts-node ./node_modules/.bin/typeorm migration:run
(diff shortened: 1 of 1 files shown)


pleerock:
can somebody else who use windows and mac and linux confirm this?

havenchyk:
Shouldn’t path use backslashes in windows?

pleerock:
no, I think windows understands both backslashes and standard slashes

Kononnable:
Confirming - on windows only second one works, on linux both. Didn't test on mac.

Issue #3071, opened 2018-11-08
MySQL support for 'bit' data type 


yonatanmn:
**Issue type:**
[X] feature request

**Database system/driver:**
[X] `mysql` / `mariadb`

**TypeORM version:**
[X] `latest`

I'm working with existing DB, that was created with hibernate ORM. All booleans were created with `bit` as it is recommended with new versions of MySQL (not sure about it though)

Now I'm adding nestjs server with typeorm, and when I try to connect to the DB, it throws that `bit` data type is not supported by mysql driver. 
If I change it to tinyint, (or not specifying anything) it alters the table, and then the hibernate server is not working. 

What can I do? why not supporting 'bit'? 

Thanks

frozenex:
+1

vlapo:
I think `bit` should be represented as string (same as in pgsql).

Also we should also add `bool` and also `boolean` to mysql driver as they are synonyms for `TINYINT(1)`

https://dev.mysql.com/doc/refman/5.5/en/numeric-type-overview.html

pleerock:
Yeah I think we should add those new data types. Its easy task to do actually.

PR #3310, opened 2018-12-21
chore(mysql): sync mysql data types with mysql docs

vlapo:
Sync mysql supported data types against mysql docs
  - added types - `bit`, `national char`, `national varchar`
  - added synonyms - `integer`, `double precision`, `real`, `fixed`, `bool`, `boolean`, `nchar`, `nvarchar`
  - update docs
  - update tests
  - close https://github.com/typeorm/typeorm/issues/3071

Maybe I missed something so maybe @AlexMesser should check this PR.


pleerock:
The most important is to check synchronization process. For example you set `bit`, then you call sync, then you call sync again and make sure no additional queries were executed. We don't have such tests and we did those checks manually.

Some databases, when you request column+type return different type other then you originaly specified. For example if you set "integer" it can return you "int" when we request schema information. Then checker makes `integer != int` and it runs synchoronization. Thats why we normalize it before comparison.

vlapo:
Thnx for comment. I did not realize problem with sync before. I just updated `normalizeType` method in `MysqlDriver` and add all synonyms like `integer`, `bool`, `boolean`,... Now there are no additional queries during sync process related to data type synonyms.

But I still have a lot of change queries related to scale and precision. E.g.
```
query: ALTER TABLE `post` CHANGE `double` `double` double NOT NULL
```
But I see PR (https://github.com/typeorm/typeorm/pull/3262) and Issues about this so it should be not related to this PR.

vlapo:
Yeah, I think now it is ready to merge

Issue #2905, opened 2018-10-09
cannot connect  to mssql server show invalid object name 'SYS.DATABASES'


arkasaha2:
**Issue type:**

[x] question
[ ] bug report
[ ] feature request
[ ] documentation issue

**Database system/driver:**

[ ] `cordova`
[ ] `mongodb`
[x] `mssql`
[ ] `mysql` / `mariadb`
[ ] `oracle`
[ ] `postgres`
[ ] `sqlite`
[ ] `sqljs`
[ ] `react-native`
[ ] `expo`

**TypeORM version:**

[x] `latest`
[ ] `@next`
[ ] `0.x.x` (or put your version here)

Error:

query failed: SELECT "NAME", "COLLATION_NAME" FROM "SYS"."DATABASES"
error: { RequestError: Invalid object name 'SYS.DATABASES'.
    at handleError (D:\nodeORM\ormNodeTest\node_modules\mssql\lib\tedious.js:546:15)
    at emitOne (events.js:116:13)
    at Connection.emit (events.js:211:7)
    at Parser.<anonymous> (D:\nodeORM\ormNodeTest\node_modules\tedious\lib\connection.js:611:16)
    at emitOne (events.js:116:13)
    at Parser.emit (events.js:211:7)
    at Parser.<anonymous> (D:\nodeORM\ormNodeTest\node_modules\tedious\lib\token\token-stream-parser.js:54:15)
    at emitOne (events.js:116:13)
    at Parser.emit (events.js:211:7)
    at addChunk (D:\nodeORM\ormNodeTest\node_modules\readable-stream\lib\_stream_readable.js:291:12)
  code: 'EREQUEST',
  number: 208,
  lineNumber: 1,
  state: 1,
  class: 16,
........

the table which it is trying to search doesn't exist since the name is in smaller case as shown in the image below: 
![capture](https://user-images.githubusercontent.com/23381789/46661751-65bb9700-cbd7-11e8-8e5e-4611eb1162e1.PNG)

Kononnable:
By default mssql server is case insensitive. Can you check if it is changed in yours case or is problem elsewhere?
`SELECT SERVERPROPERTY('COLLATION')` or `SELECT collation_name FROM sys.databases WHERE name = 'your_database_name'` should tell if server/database is CS or CI.

vedun:
[...]

PR #3312, opened 2018-12-21
invalid object name 'SYS.DATABASES'. #2905 fix

vedun:


pleerock:
so simple, thank you!

PR #3315, opened 2018-12-22
fix(test): rowversion type is presented as timestamp type in SQL server

vlapo:
In MSSQL rowversion is represented as timestamp column data type. This PR just fix test with type checking rowversion.

I think this is last puzzle to solve issue  https://github.com/typeorm/typeorm/issues/3169. I think we can also enable mssql tests on travis and will see if it is more stable now.


vlapo:
Checked 2 times. I believe synchronisation is already correctly handled in `SqlServerDriver.normalizeType` by https://github.com/typeorm/typeorm/blob/4b8b371a937f9d087c8dcb37cc6d7e7c48efe5c1/src/driver/sqlserver/SqlServerDriver.ts#L445-L448
and this PR is just test correction.

Issue #2066, opened 2018-05-02
Doc issue: migration from sequelize


stevenmusumeche:
**Issue type:**

[ ] question
[ ] bug report
[ ] feature request
[ x] documentation issue

**Database system/driver:**

[ ] `cordova`
[ ] `mongodb`
[ ] `mssql`
[x ] `mysql` / `mariadb`
[ ] `oracle`
[ ] `postgres`
[ ] `sqlite`
[ ] `sqljs`
[ ] `react-native`

**TypeORM version:**

[ x] `latest`
[ ] `@next`
[ ] `0.x.x` (or put your version here)

Thank you for this project.  It is awesome.  We are converting an API from sequelize and ran into a small bug in the documentation.

In the sequelize migration documentation, the following line is used to demonstrate how to create a new entity:

```
const employee = Employee.create({ name: "John Doe", title: "senior engineer" });
```

In order for this to work, the create method needs to be called on the repository, not the entity.

```
const employee = await EmployeeRepository.create({ name: "John Doe", title: "senior engineer" });
```

pleerock:
if you are using active record (e.g extend BaseEntity) then `create` method is available to you

stevenmusumeche:
Would you like a PR to the documentation to make that more clear?

pleerock:
Okay, sure!

PR #3316, opened 2018-12-22
doc(sequelize): Add few links to other docs and small updates

vlapo:
- Added few links to other part of codumentation in sequelize migration part. 
 - Also mentioned active record pattern and repository api so I hope we can close https://github.com/typeorm/typeorm/issues/2066


pleerock:
Thank you very much!

Issue #2849, opened 2018-09-25
Query builder update generates broken SQL


marano:
**Issue type:**

[x] bug report

**Database system/driver:**

[x] `postgres`

**TypeORM version:**

[x] `latest`

```js
const ids = [1]
const query = manager
  .createQueryBuilder()
  .update(Entity)
  .set({ brandId: 2 })
  .where(`"brandId" IN (:...ids)`, { ids })
console.log(query.getSql())
```

This logs:
```
query: UPDATE "image_classification" SET WHERE "brandId" IN ($1) -- PARAMETERS: [1]
```

Expected result:
Generated SQL to include object passed to `set` method.

marano:
I've figured this works:
```
const brand = await Brand.findOne(2)
...
.set({ brand })
```

But it seems really weird that typeorm won't let me set the id field directly. Also it shouldn't fail silently. Because of that I will let this issue open for further discussion.

vlapo:
I think it should be good manner to throw error in case final map of values to update is empty. In `UpdateQueryBuilder.getValueSet()` is small empty check and throw. But later in `createUpdateExpression()` is this map filtered and possible ends with empty array of values to change.

Also It should be nice to allow update "columns generated by relations" as in this case where `brandId` is generated by ORM.

pleerock:
@vlapo I think we shall check both emptiness and wrong properties in.

vlapo:
Sure I agree. But what about join column not defined in entity as property like `brandId` in this issue? 

Maybe we should allow update `brandId` even if it is not defined in entity.

Example:
```typescript
@Entity()
export class Post {

    @PrimaryGeneratedColumn()
    id: number;

    @ManyToOne(type => Author, author => author.posts)
    @JoinColumn()
    author: Author;
}
```
```typescript
const query = connection.manager
            .createQueryBuilder()
            .update(Post)
            .set({ authorId: 2 })
            .where({ id: 1 });
console.log(query.getSql());
```
This should be valid way to update author in post.

pleerock:
no, I don't think so. This design isn't clear. What is `authorId` in there? It means we would allow to specify any database column as well? Why only database column and property name? What about other expressions `UPDATE SET` supports? 

In terms of this issue we only need to add those exceptions and close it. If we want to extend update query builder functionality, we need to create a separate issue and describe a new proposal of update query builder that covers all cases.

vlapo:
[...]

PR #3324, opened 2018-12-25
fix(query builder): Throw error on missing or unknown properties

vlapo:
- handle emptiness and wrong properties in update query builder
 - fix https://github.com/typeorm/typeorm/issues/2849

### DIFF
--- a/src/error/UpdateValuesMissingError.ts
+++ b/src/error/UpdateValuesMissingError.ts
@@ -1,0 +1,13 @@
+/**
+ * Thrown when user tries to update using QueryBuilder but do not specify what to update.
+ */
+export class UpdateValuesMissingError extends Error {
+    name = "UpdateValuesMissingError";
+
+    constructor() {
+        super();
+        Object.setPrototypeOf(this, UpdateValuesMissingError.prototype);
+        this.message = `Cannot perform update query because update values are not defined. Call "qb.set(...)" method to specify updated values.`;
+    }
+
+}
--- a/src/error/EntityColumnNotFound.ts
+++ b/src/error/EntityColumnNotFound.ts
@@ -1,0 +1,13 @@
+/**
+ *
+ */
+export class EntityColumnNotFound extends Error {
+    name = "EntityColumnNotFound";
+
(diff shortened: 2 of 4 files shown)


src/error/UpdateValuesMissingError.ts
```diff
@@ -0,0 +1,13 @@
+/**
+ * Thrown when user tries to update using QueryBuilder but do not specify what to update.
+ */
+export class UpdateValuesMissingError extends Error {
+    name = "InsertValuesMissingError";
```
pleerock: name is wrong

src/error/UpdateValuesMissingError.ts
```diff
@@ -0,0 +1,13 @@
...
+
+    constructor() {
+        super();
+        Object.setPrototypeOf(this, UpdateValuesMissingError.prototype);
+        this.message = `Cannot perform update query because update values are not defined. Call "qb.set(...)" method to specify inserted values.`;
```
pleerock: "updated values"

src/error/ColumnMetadataNotFound.ts
```diff
@@ -0,0 +1,13 @@
+/**
+ *
+ */
+export class ColumnMetadataNotFoundError extends Error {
+    name = "ColumnMetadataNotFoundError";
+
+    constructor(propertyPath: string) {
+        super();
+        Object.setPrototypeOf(this, ColumnMetadataNotFoundError.prototype);
+        this.message = `No column metadata for property "${propertyPath}" was found.`;
```
pleerock: probably we need to give more user-friendly error. Users don't know what "metadata" means.

vlapo: You are right. We need more user friendly errors. What do you suggest?
EntityColumnNotFound -> `No entity column "${propertyPath}" was found.`?

vlapo:
Totally forgot. Updated by comments.

pleerock:
Thank you. This is important to mention in changelog. In can break some exist code in a way.

Issue #3318, opened 2018-12-24
Document "Example using TypeORM with Express" Missing Update


rationalthinker1:
**Issue type:**

[ ] question
[ ] bug report
[ ] feature request
[X] documentation issue

**Database system/driver:**

[ ] `cordova`
[ ] `mongodb`
[ ] `mssql`
[ ] `mysql` / `mariadb`
[ ] `oracle`
[ ] `postgres`
[ ] `sqlite`
[ ] `sqljs`
[ ] `react-native`
[ ] `expo`

**TypeORM version:**

[X] `latest`
[ ] `@next`
[ ] `0.x.x` (or put your version here)

If you look at the document, it doesn't include the update function. When writing the routes, it defines the update route and doesn't show it's done with TypeORM. Also, I want to use Active Record, it doesn't show it's done. I see that there's a lot of documentation and I'm surprised that there isn't a document about basic CRUD operations using Active Records with Express.
https://github.com/typeorm/typeorm/blob/master/docs/example-with-express.md

pleerock:
Following code can be used to both insert and update:

```ts
    app.post("/users", async function(req: Request, res: Response) {
        const user = userRepository.create(req.body);
        return userRepository.save(user);
    });
```

you just need to pass an id in `req.body` and it will update user with a given id.

This express example is just `an example`. You can do it the way we described, or any other way you prefer.

rationalthinker1:
In the documents, it shows the route:
```
app.put("/users/:id", function(req: Request, res: Response) {
    // here we will have logic to update a user by a given user id
});
```

But in the actual implementation, it skips over it.
```
import * as express from "express";
import {Request, Response} from "express";
import * as bodyParser from  "body-parser";
import {createConnection} from "typeorm";
import {User} from "./User";

// create typeorm connection
createConnection().then(connection => {
    const userRepository = connection.getRepository(User);
    
    // create and setup express app
    const app = express();
    app.use(bodyParser.json());
    
    // register routes
    
    app.get("/users", async function(req: Request, res: Response) {
        return userRepository.find();
    });
    
    app.get("/users/:id", async function(req: Request, res: Response) {
        return userRepository.findOne(req.params.id);
    });
    
    app.post("/users", async function(req: Request, res: Response) {
        const user = userRepository.create(req.body);
        return userRepository.save(user);
    });
    
    app.delete("/users/:id", async function(req: Request, res: Response) {
        return userRepository.remove(req.params.id);
    });
    
    // start express server
[...]

PR #3325, opened 2018-12-25
docs(express): small correction of express example

vlapo:
- add update endpoint implementation
 - closes https://github.com/typeorm/typeorm/issues/3318


docs/example-with-express.md
```diff
@@ -201,30 +201,34 @@ import {User} from "./User";
...
-    
+
     app.post("/users", async function(req: Request, res: Response) {
         const user = userRepository.create(req.body);
         return userRepository.save(user);
     });
-    
+
+    app.put("/users/:id", function(req: Request, res: Response) {
+        return userRepository.update(req.params.id, req.body);
```
pleerock: I would like to avoid using `update` method. `update` method does a very simple thing - it executes update query and basically thats it. `save` method does everything orm is able to do. And many users, especially newcomers may not understand why update isnt working as they expected.

So in this case, if you want to have a separate `put` method I would recommend to `find` object by `req.params.id`, assign req.body and save that object.

vlapo: Sure,I understand your concern

Issue #2807, opened 2018-09-16
Running npm test only executes 11 tests?


dominicdesu:
**Issue type:**

[x] question
[ ] bug report
[ ] feature request
[ ] documentation issue

**Database system/driver:**

[ ] `cordova`
[ ] `mongodb`
[ ] `mssql`
[ ] `mysql` / `mariadb`
[ ] `oracle`
[ ] `postgres`
[x] `sqlite`
[ ] `sqljs`
[ ] `react-native`
[ ] `expo`

**TypeORM version:**

[x] `latest`
[ ] `@next`
[ ] `0.x.x` (or put your version here)

Hi I am trying to run all tests with the sqlite database enabled.
However, it seems like only 11 tests are being executed, and no error message is displayed either. This is the output:

```
PS C:\Projects\typeorm> npm test

> typeorm@0.2.7 test C:\Projects\typeorm
> tsc && mocha --file ./build/compiled/test/utils/test-setup.js --bail --recursive --timeout 30000  ./build/compiled/test



  benchmark > bulk-save > case1
    √ testing bulk save of 10.000 objects

  benchmark > bulk-save > case2
    √ testing bulk save of 10000 objects

  cascades > should insert by cascades from both sides (#57)
    √ should insert by cascades from owner side (47ms)

  columns > embedded columns
    √ should insert / update / remove entity with embedded correctly (49ms)
    √ should properly generate column names

  columns > getters and setters
    √ should not update columns marked with readonly property

  columns > no-selection functionality
    √ should not select columns marked with select: false option
    √ should not select columns with QueryBuilder marked with select: false option
    √ should select columns with select: false even columns were implicitly selected

  columns > readonly functionality
    √ should not update columns marked with readonly property

  columns > value-transformer functionality
    √ should marshal data using the provided value-transformer

  Connection
    before connection is established

  11 passing (678ms)

      1) "before all" hook
      2) "after all" hook
```

It seems to stop when it would execute an mysql test, but mysql is disabled in my ormconfig.json:

```
[
  {
    "skip": true,
    "name": "mysql",
    "type": "mysql",
    "host": "localhost",
    "port": 3306,
    "username": "test",
    "password": "test",
    "database": "test",
    "logging": false
  },
  {
    "skip": true,
    "name": "mariadb",
    "type": "mariadb",
    "host": "localhost",
    "port": 3307,
    "username": "test",
    "password": "test",
    "database": "test",
    "logging": false
  },
  {
    "skip": false,
[...]

PR #3332, opened 2018-12-26
Test cleanup

elthrasher:
See if we can speed up tests by running in parallel on travis by db. I needed to rewrite a few tests as they would fail if mysql was not enabled. Some of this needs to be tweaked to be a bit more elegant. Can look at adding back mssql and oracle if this works out.


elthrasher:
@Kononnable @pleerock @vlapo I couldn't get Travis to really work the way I wanted it. The main issue is an inability to share cache between the build stages. If I had access to a Docker registry, I'd build an image which would then cache the install, lint and build stages so we could basically focus on spending all our time in tests. As things are, the overhead of running those processes over and over kills any advantage we get from running everything in parallel. And Travis does not provide any sort of Docker cache so you have to pull each image on each run. We could spin our own and store everything in an S3 bucket, but that seems like too much of a hack.

That's not to say we can't run things in parallel. If we don't care so much about build times, we can do this quite easily:
screen shot 2018-12-27 at 10 18 05 am

Running like that builds take around 16-17 minutes (and can take more if there's a wait for jobs to start or especially if there are several commits at once).

I also cleaned up some of the tests and externalized config so you can set `CONNECTIONS` in your environment, which would be a comma-separated list that overrides the skip values in `ormconfig.json`. It's better for CI to be able to control such things via environment vars and the cleanup I did now allows someone to enable any (save mssql or oracle) connection and run tests locally with no bleed over.

I know I started out talking about jest, but I wanted to first see how things run in CI. To take full advantage of jest, I think we'd need to introduce some magic layer where each test creates and runs in its own db or schema. I also removed `npm install sqlite3 --build-from-source` because it doesn't seem necessary and takes some time.

As I see it, the options at this point are:
1. Parallel gives useful insights into how the tests are running and build times aren't that important so let's clean this up and publish it.
2. Keep the test cleanup, but don't bother with parallel execution. Instead figure out how to speed things up with jest.
3. Find a different CI tool with better support for build stages.
4. Something else?

Kononnable:
>We could spin our own and store everything in an S3 bucket, but that seems like too much of a hack.
It would be a problem on non-collaborator builds(majority of PRs) - Travis doesn't allow to use secrets in such builds(because someone might steal them).

We might get it a little bit faster by separating compilation from running tests and compile in the background while downloading docker images(https://stackoverflow.com/questions/19543139/bash-script-processing-limited-number-of-commands-in-parallel). However db often accepts connections few seconds(even minutes in case of oracle) after container is up - so we would have to retry until it is ready.

As for jest - I didn't have an opportunity to use it yet but I'm not sure if we will gain much from it if new layer needs to be added. Database creation for each test probably will be slow(often it creates new files in db container). New schema might be an option(few tests use multiple schemas, so schema with prefix would be necessary) but I think mysql, mariadb and oracle support only one schema per database.

Another option would be to make tests separated(each test run single db driver) and run them in parallel but make sure that two tests with same driver won't be running at the same time. Not sure if this is easily achievable.

test/utils/test-utils.ts
```diff
@@ -158,13 +158,22 @@ export function setupSingleTestingConnection(driverType: DatabaseType, options:
...
         } catch (err) {
-            return require(__dirname + "/../../ormconfig.json");
+          config = require(__dirname + "/../../ormconfig.json");
+        }
+        // environment var override
+        // if CONNECTIONS is present, its values will override ALL skip values
+        if (process.env.CONNECTIONS) {
+          const connections = process.env.CONNECTIONS.split(",");
+          config = config.map((entry: TestingConnectionOptions) => {
+            entry.skip = connections.indexOf(entry.type) === -1;
```
havenchyk: includes?

test/utils/test-utils.ts
```diff
@@ -158,13 +158,22 @@ export function setupSingleTestingConnection(driverType: DatabaseType, options:
...
     try {
-
+        let config;
         try {
-            return require(__dirname + "/../../../../ormconfig.json");
-
+          config = require(__dirname + "/../../../../ormconfig.json");
         } catch (err) {
-            return require(__dirname + "/../../ormconfig.json");
+          config = require(__dirname + "/../../ormconfig.json");
```
havenchyk: `path.join`?

elthrasher:
@havenchyk The trouble there is Travis. There's no local Docker cache so I would have to either build the image in each process or build it, push it to some registry (now credentials are a problem), then pull it in each process.

havenchyk:
@elthrasher I was talking about the second approach (build, upload, use), but also have you tried using `matrix`? I haven't used this option, but it can solve at least the problem with sqlite. I'm checking docs at the moment, so can't suggest anything better before I finish and try this option

elthrasher:
Yeah, a few of my attempts did something like that. The problem still is caching Docker. However, looks like there may be a way to hack around it: https://github.com/travis-ci/travis-ci/issues/5358#issuecomment-246302666
Probably worth a try. The alternative would be to move to some other CI provider that supports Docker caching better.

elthrasher:
Okay, lots of failures and a few successes here.

There's been a long-standing issue that you can't just run tests against one kind of db. For example, disable all but postgres and run the tests and you'll get failures because it can't find mysql. I've fixed that so if mysql is disabled in ormconfig.json, no tests that require it will be run.

I added an env variable `CONNECTIONS` that will override the `skip` property of ormconfig.json. That makes it easier to manage what runs and when.

I tried to do a sort of pipeline where the build happens once and then we run a bunch of parallel processes, one for each db connection. TravisCI just doesn't support this. There's no way to pass files from one stage to the next. Caches are not sharable. The only way we could do this on TravisCI is to publish a Docker image and then pull it for each parallel stage or to stage files on S3, both of which have their own complications. The real gain here would probably be fairly minimal as each stage would have to pull the Docker image, cutting into the time we saved.

I experimented with running parallel mocha processes in the same job (sort of how it would work with jest, but explicitly by db connection). This works and gives us a slight improvement in time taken, however we get six times the logs so I had to switch to a more terse logging format.

I tried running the setup stuff in parallel and saw some performance gains, but it gets complicated because travis doesn't do a great job of waiting for the setup to finish before the tests begin. If we were running 100 builds per day, I'd still optimize this, but I don't think the added complexity is worth it for typeorm.

I removed node 9, which saw end-of-life last summer in favor of node 11. Fortunately all the tests still pass. I assume this is just a chore that nobody has gotten to yet rather than a commitment to support node 9.

I tried altering the tests so that each test would create its own schema or database, which would let us convert to jest and full parallel execution (though it still remains to be seen how much improvement we'd get out of that). However, it's extremely complicated to do that for all the databases typeorm supports. I think it's doable, but should probably be its own PR if we decide we want to try it.

Thoughts? If we want to go ahead with this, I'll add some more comments.

vlapo:
Just reference to issue related to this changes https://github.com/typeorm/typeorm/issues/2807

pleerock:
> There's been a long-standing issue that you can't just run tests against one kind of db. For example, disable all but postgres and run the tests and you'll get failures because it can't find mysql. I've fixed that so if mysql is disabled in ormconfig.json, no tests that require it will be run.

mysql is required because very-very-old tests were using it (in the times when there was only mysql support in typeorm) and we still didn't refactor those old tests

> TRAVIS

if travis is a problem lets try some other CI. I have experience working only with travis and circleci. Are there better and popular/perspective alternatives?

elthrasher:
@pleerock This PR refactors those tests so that mysql won't always be required.

It's not that Travis is ruining the project or anything. It's just that some of the suggestions made above won't work on Travis. It kills me a little that CI runs for 4-5 minutes before the tests even start, but is this really a big problem for typeorm? Probably not.

I've added my comments so this is good to go from my perspective if we like these changes. If not, happy to discuss or revise.

tsconfig.json
```diff
@@ -23,6 +23,7 @@
         "tmp",
         "temp",
         "build",
-        "node_modules"
+        "node_modules",
+        "test/utils/test-runner.ts"
```
Kononnable: Why this file is excluded and runned with ts-node?

elthrasher: If that file is included in tranpilation, no tests are executed. I think it's some issue with the mocha global. I had started to run tests using ts-lint entirely but stopped after having some problems with Travis. Probably a good idea to take that up again and see if it will work. It should make local runs easier and allow us to exclude all tests from the build, which will make things faster.

elthrasher:
Tests run via ts-node only now and are much faster! I'm surprised, but we've cut over two minutes off. This should also make development much nicer as compilation isn't necessary to run tests. I think we're now getting out of mocha what we could've hoped to get out of jest, so that conversion won't be necessary with the current feature set.

Build times are still running 6-7 minutes but only 150 seconds are spent actually executing the tests. If we want to run faster, we could look at another CI, but I would merge this PR and start another for that.

elthrasher:
This commit: https://github.com/typeorm/typeorm/commit/82ad52b5437840ca9bfaab40e860fbf9e0318a4a#diff-165e041541ffbf626d08c6ac034a9c58R7 has removed almost all of the tests from execution via `only`. I've fixed that here.

elthrasher:
Ah well, running only one test explained the incredible test performance. Looks like we're back to normal levels, but I still think this is worth it.

Kononnable:
Ok, to sum up what are Pros and Cons of this PR?

Pro:
CONNECTIONS environment variable
Some tests fixed
Db drivers don't wait for each other(so tests should be faster)
ts-node - faster development ( I was using `tsc -w` for compiling so no big difference for me)

Con:
We don't see result for each test(benchmark tests doesn't make much sense now)


I probably missed something here, so feel free to update the list. Is there any failing CI report so we can see how it looks like when CI fails?

elthrasher:
Here's how it looks when it fails: https://travis-ci.org/typeorm/typeorm/jobs/474388884

The performance gains for running parallel seem minor, so that could be removed for more verbose output, thus keeping most of the pros without the con.

.travis.yml
```diff
@@ -1,28 +1,25 @@
...
-#  - stable
+  - 11
   - 10
-  - 9
   - 8
-#  - 6
-#  - 4
-#  - 0.12
-#  - 0.11
+cache: npm
```
Kononnable: Are we utilizing cache if we use `npm ci`? I think it removes `.\node_modules` at the beginning.

.gitignore
```diff
@@ -7,4 +7,5 @@ node_modules/
 ormconfig.json
 ormlogs.log
 npm-debug.log
+test/functional/sqljs/*.sqlite
```
Kononnable: Why not use temp folder? It has it own gitignore file.

test/github-issues/182/entity/Post.ts
```diff
@@ -12,7 +12,7 @@ export class Post {
     @Column()
     title: string;
 
-    @Column({ nullable: false })
+    @Column("int", { nullable: false })
```
Kononnable: Why was that needed? Shouldn't change much but don't see any obvious reason.

.travis.yml
elthrasher: See the docs: https://docs.travis-ci.com/user/languages/javascript-with-nodejs/#npm-ci-support
`$HOME/.npm` is cached, so yes, this does help.

.gitignore
elthrasher: That output had previously gone to `build/compiled/test...` which is gitignored already. I could redirect this output, but that would involve changing more tests. Think that would be better?

test/github-issues/182/entity/Post.ts
elthrasher: This test was a little tricky. It implies it's testing an enum type, but it never creates an enum column in the db, just uses an enum in the orm code. When the code is transpiled, for whatever reason the column is created as an int, but running with ts-node it was sending `object` to mysql as the column type, which of course doesn't work. Adding `int` there made the test execute correctly - the same SQL statements are sent to the db for both the transpiled and ts-node versions.

.travis.yml
Kononnable: Oh, forgot about that, somebody mentioned it before somewhere

elthrasher:
Doesn't seem like the parallel execution is getting us anything other than confusing output so I removed it. I still think there are enough good things here to consider merging and maybe think about another CI which would support parallel better - or just solve some completely different problem. 10 minutes for CI isn't that bad.

package.json
```diff
@@ -100,7 +100,7 @@
   },
   "scripts": {
     "test-ci": "gulp ci-tests",
-    "test": "rimraf ./build && tsc && mocha --file ./build/compiled/test/utils/test-setup.js --bail --recursive --timeout 60000 ./build/compiled/test",
+    "test": "ts-node test/utils/test-runner.ts",
```
pleerock: I would like to keep previous setup. `ts-node` works very slow in windows (in my own tests) and I don't want to use it.

pleerock: it was few years ago last time I reproduced issue on windows however. Maybe we can keep ts-node

elthrasher: Well, let me know. We could support both if desired.

test/github-issues/182/entity/Post.ts
pleerock: really useful to add according comment :)

pleerock:
Im okay to merge it

.gitignore
Kononnable: If that how it always worked let's leave it for now.

Kononnable:
I've measured times on my machine:
Enabled drivers: mysql, mariadb, postgres, sqljs
```
configuration    this PR      master
linux1           16m2s        16m9s  
linux2           1m50s        2m32s
windows          5m49s        5m37s
```
Configurations where different(hardware, db engine config) so there is no point in comparing linux to windows etc.  It looks like ts-node doesn't introduce performance problems on windows anymore(at least in our case). I think it's good or, we could support both ways if it is easy to implement.

elthrasher:
I'll give circleci a look when I can. As for running both ts-node and compiled, the main downside of that is that we don't need to ever compile the tests when we use ts-node - in fact they are excluded in the tsconfig.json in this PR. It does save a few seconds.

havenchyk:
@elthrasher I wasn't sure about this when I proposed this try, but I use ava as a test runner and there is an option there to run tests without pre-compilation, with requiring ts-node and it seemed to be faster :)

vlapo:
So whats a status of this ? :-)

imnotjames:
I like the idea of using `ts-node` for our tests - can we get a PR that's just that?

It seems there's a bunch of changes here that kinda all got conflated & made it harder to agree on merging them all in together

We aren't using `travis` any longer so that part's no longer relevant.

`setupConnection` was dropped, too.

I'll be closing this one, though - cause there's lots of conflicts and not too much  movement for over a year.

PR #3338, opened 2018-12-27
gitignore

huanz:


Kononnable:
Not sure if we should add all files with might be added not just `yarn-error.log` which probably was a reason why this PR was made. E.g.: `Files that might appear in the root of a volume`  seems unnecessary for me.
But it's probably a topic for discussion like `tabs or spaces`.

havenchyk:
You can specify ignores files globally on your machine, so I’d not extend this gitognore files, even if you did investigation and wanted to contribute, I don’t think this change is necessary, so it’s up to maintainers

PR #3358, opened 2019-01-03
Store UUID as char

elderapo:
UUID has fixed length (36) hence it's better to store it as char instead of varchar.


pleerock:
Please try to not reformat code when changing files, commit only changes related to your feature request / bugfix.

pleerock:
First, I don't understand why you changed `uniqueidentifier` to `char`. Second, I don't think this worse changing because it will break all current user data.

elderapo:
I disabled prettier in vscode for typeorm workspace to keep old formatting style but something formatted code anyways.

You are right, I should've kept ```uniqueidentifier``` instead of ```char``` for SqlServerDriver, my bad.

I guess there is no point of fixing this pull request if it can break all current user data.

pleerock:
Yes I think it simply doesn't worth it. Thank you for your efforts anyway!

PR #3362, opened 2019-01-04
double default value fixed

kandros:
Default value connectionName to “default” is already set in ConnectionManager’s get function

https://github.com/typeorm/typeorm/blob/master/src/connection/ConnectionManager.ts#L37


pleerock:
Thanks for the contribution!

You are right, its already defined in the connection, but can you please tell me what is the point of this change?

pleerock:
From another perspective its handy to see that its by default equals to `default` when you hover over `getConnection` method, right :)

I'm okay to merge this branch, I have in plans to change those methods signature in 0.4.0 anyway.

PR #3364, opened 2019-01-04
Enchancement - Add support for ioredis Cluster object initialization

ignacioblit:
As a continuation to:

https://github.com/typeorm/typeorm/pull/3289

and my comment about Ioredis cluster object I submit the following tweaks which allow to initialize a Ioredis cluster object as a caching option.

https://github.com/luin/ioredis/blob/master/API.md#Cluster


pleerock:
Thank you!

Issue #3347, opened 2019-01-01
Can't Connect To MongoDB Atlas Cluster


pointtoken:
**Issue type:**

[ ] question
[X] bug report
[ ] feature request
[ ] documentation issue

**Database system/driver:**

[ ] `cordova`
[X] `mongodb`
[ ] `mssql`
[ ] `mysql` / `mariadb`
[ ] `oracle`
[ ] `postgres`
[ ] `sqlite`
[ ] `sqljs`
[ ] `react-native`
[ ] `expo`

**TypeORM version:**

[X] `latest`
[ ] `@next`
[ ] `0.x.x` (or put your version here)

**Steps to reproduce or a small repository showing the problem:**

I have a valid connection string to a Mongo Atlas DB that I am trying to use with TypeORM. I know the connection string works because I have used it from Python and from the Mongo shell. But it doesn't work from typeorm. I have tried both the **mongodb://** and **mongo+srv** synax. Not sure how to debug or figure this one out. Here is the error:

```
(node:19257) DeprecationWarning: current URL string parser is deprecated, and will be removed in a future version. To use the new parser, pass option { useNewUrlParser: true } to MongoClient.connect.
{ MongoNetworkError: connection 5 to cluster0-XXXX.mongodb.net:27017 closed
    at Socket.<anonymous> (/home/administrator/Source/Repos/emoon-relayer/node_modules/mongodb-core/lib/connection/connection.js:276:9)
    at Object.onceWrapper (events.js:273:13)
    at Socket.emit (events.js:182:13)
    at TCP._handle.close (net.js:611:12)
  name: 'MongoNetworkError',
  errorLabels: [ 'TransientTransactionError' ],
  [Symbol(mongoErrorContextSymbol)]: {} }
```

vlapo:
Did you try add `useNewUrlParser: true` to your connection config?

rustamwin:
Try add `dbName: <your database name>` to your connection options

pointtoken:
Thanks guys -- but how do I actually pass parameters when MongoDB  connects. The JSON object passed on createConnection in typeORM wraps when MongoDB calls connect.

vlapo:
After few tries I have two valid configurations for typeorm and mongo atlas:
```typescript
const connection = await createConnection({
            type: "mongodb",
            useNewUrlParser: true,
[...]

Issue #1741, opened 2018-03-11
Cannot define SSL connection options for MongoDB in environment variables


PRo25:
**Issue type:**

[ ] question
[X] bug report
[ ] feature request
[ ] documentation issue

**Database system/driver:**

[ ] `cordova`
[X] `mongodb`
[ ] `mssql`
[ ] `mysql` / `mariadb`
[ ] `oracle`
[ ] `postgres`
[ ] `sqlite`
[ ] `sqljs`
[ ] `websql`

**TypeORM version:**

[ ] `latest`
[ ] `@next`
[X] `0.1.11`

**Steps to reproduce or a small repository showing the problem:**

It isn't possible to configure a connection to a MongoDB server that uses SSL by using the environment variables. I tried, without success, to use the variable "TYPEORM_DRIVER_EXTRA" by defining it in the ".env" file like this: "TYPEORM_DRIVER_EXTRA={ "ssl": true, "sslValidate": false }". But after checking the MongoDB driver code (https://github.com/typeorm/typeorm/blob/master/src/driver/mongodb/MongoDriver.ts) I noticed it doesn't use the extra property of the options object for anything so it's normal that it doesn't work.
If I configure the connection options with SSL directly in the code then it works fine.

pleerock:
Thank you for the report. PR will be accepted for this feature request.

osdiab:
I'm using postgres, but I believe my pr should fix this? #3409 at least for setting it to `true`.

PR #3366, opened 2019-01-04
fix(mongodb): do not pass false undefined options to mongo client

vlapo:
In mongodb driver `passed` options take precedence over connection string option.
Thats reason why our users have problems with connect to mongo using `url` property like in https://github.com/typeorm/typeorm/issues/3347

 - This change will check all valid mongo client options and pass only exists one from `MongoConnectionOptions`.
 - Options updated according to http://mongodb.github.io/node-mongodb-native/3.1/api/MongoClient.htm

Do not see any way how to test so if you have any idea give me a kick. Tested locally with atlas cluster.

### DIFF
--- a/src/driver/mongodb/MongoDriver.ts
+++ b/src/driver/mongodb/MongoDriver.ts
@@ -116,6 +116,72 @@ export class MongoDriver implements Driver {
      */
     protected mongodb: any;
 
+    /**
+     * Valid mongo connection options
+     * NOTE: Keep sync with MongoConnectionOptions
+     * Sync with http://mongodb.github.io/node-mongodb-native/3.1/api/MongoClient.html
+     */
+    protected validOptionNames: string[] = [
+        "poolSize",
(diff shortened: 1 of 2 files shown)


pleerock:
Thank you!

src/driver/mongodb/MongoDriver.ts
```diff
@@ -394,4 +420,21 @@ export class MongoDriver implements Driver {
...
+     */
+    protected buildConnectionOptions(): any {
+        const mongoOptions: any = {};
+
+        for (let index = 0; index < this.validOptionNames.length; index++) {
+            const optionName = this.validOptionNames[index];
+
+            if (optionName in this.options) {
+                mongoOptions[optionName] = (this.options as any)[optionName];
+            }
```
rustamwin: need check `this.options.extra` ? https://github.com/typeorm/typeorm/issues/1741

PR #3368, opened 2019-01-05
Update docs to recommend using npx instead of global docs

ustun:
npx tool, installed by default in node will download the library and execute.
This will also prevent a global installation and ensure that the binary run is
always the latest stable version.

### DIFF
--- a/README-zh_CN.md
+++ b/README-zh_CN.md
@@ -213,16 +213,10 @@ await timber.remove();
 快速开始只有在NodeJS应用程序中使用TypeORM才可以使用。
 如果你正在使用其他平台，请看[分步指南](#分步指南)。
 
-首先全局安装TypeORM：
-
-```
-npm install typeorm -g
-```
-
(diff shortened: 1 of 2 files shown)


vlapo:
npx is by default installed with npm v5.2.0 and later. Right now we support node v8 and older version of 8 was delivered with npm v5.0.x. Plus node v6 have end of life scheduled on April 2019. 

So I think we should add only note with `npx` command and still have global installation presented in our documentation.

pleerock:
Even through I still use `-g` to install things, I'm fine to use npx instead

ustun:
@vlapo How about if we do as create-react-app does and put a note for people in older versions? 

See: https://github.com/facebook/create-react-app#quick-overview

vlapo:
Sure @ustun. I think it is better solution than I suggested.

vlapo:
Ping @ustun 🏓  How this looking from your side?

pleerock:
fixed by #4155. Thanks guys

PR #3380, opened 2019-01-07
fix #3379

### LINKED ISSUE #3379: Migration will keep create and drop indexes if index name is the same across tables
**Issue type:**

[x] bug report

**Database system/driver:**

[x] `mysql` / `mariadb`

**TypeORM version:**

[x] `latest`

album.ts
```
@Entity('album')
@Index('album_id', ['id'])
export class Album {
    @Column('int', {
        nullable: false,
        primary: true,
        name: 'album_id',
    })
    id: number;
}
```

albumData.ts
```
@Entity('albumData')
@Index('album_id', ['id'])
export class AlbumData {
    @Column('int', {
        nullable: false,
        primary: true,
        name: 'album_id',
    })
    id: number;
}
```
This migration or schema sync will keep dropping and creating the index every time because the index columnNames are ['album_id', 'album_id'] and it should be ['album_id'] only.

hkjeffchan:
Sorry, don't really know how to create a test yet. I test it on mysql with my test case and it should have fixed that.

This bug is probably in PostgresQueryRunner.ts too but I don't really have a postgresDb to test on hand.

### DIFF
--- a/src/driver/mysql/MysqlQueryRunner.ts
+++ b/src/driver/mysql/MysqlQueryRunner.ts
@@ -1314,7 +1314,7 @@ export class MysqlQueryRunner extends BaseQueryRunner implements QueryRunner {
             }), dbIndex => dbIndex["INDEX_NAME"]);
 
             table.indices = tableIndexConstraints.map(constraint => {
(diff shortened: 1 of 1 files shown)


Kononnable:
Original issue was fixed for each driver in #3892.

Issue #3375, opened 2019-01-07
Add metadata to migrations table


jeremija:
**Issue type:**

[x] question
[ ] bug report
[x] feature request
[ ] documentation issue

**Database system/driver:** any

**TypeORM version:**

[x] `latest`
[ ] `@next`
[ ] `0.x.x` (or put your version here)

**Steps to reproduce or a small repository showing the problem:**

I would like to add some metadata to the `migrations` table, like `date_executed` and a `md5sum` or `sha1sum` of the filename to ensure nobody modified a past migration that was already executed. Ideally, it should be possible to add any kind of metadata.

Looks like some methods in `MigrationExecutor` would need to be modified to allow for these changes. If there is interest I could work on a pull request for this.

pleerock:
migrations table is used internally by orm, and if orm doesn't need those additional columns it shouldn't have them. I suggest you to create separate table where you can store what you want.

jeremija:
True, I could create a wrapper around the `migration:run` and `migration:revert` and store this  metadata in a separate table.

However, I just wanted to point out the benefits of the aforementioned columns. A `date_executed` timestamp has proven to be useful when debugging a production instance, and the migration checksum is used to prevent a scenario where new migrations are tested on an empty database (in a CI job), and then a migration suddenly fails on a test or production environment because somebody has modified a past migration. Nothing is worse than having a migration that fails mid-way. Unfortunately MySQL does not really support [transactions for DDL statements](https://dev.mysql.com/doc/refman/8.0/en/cannot-roll-back.html).

Hashing a migration file is new concept:

- [Flyway](https://flywaydb.org/getstarted/how) supports it out of the box (column is called `checksum`)
- It was described in [Stack Overflow: How we do deployment](https://nickcraver.com/blog/2016/05/03/stack-overflow-how-we-do-deployment-2016-edition/#database-migrations) blog post:

   > In memory, we compare the list of migrations that already ran to those needing to run then execute what needs running, in file order. If we find the hash of a filename doesn’t match the migration with the same file name in the table, we abort as a safety measure. We can --force to resolve this in the rare cases a migration should have changed (almost always due to developer error). After all migrations have run, we’re done.

- [postgres-migrations](https://www.npmjs.com/package/postgres-migrations) provides hash checks migration files

   > **Hash checks for previous migrations**
   >
   > Previously run migration scripts shouldn't be modified, since we want the process to be repeated in the same way for every new environment.
   >
[...]

PR #3386, opened 2019-01-08
Add hash and create_date to migrations table

jeremija:
Related to #3375.

Added tests and did my best to try to make it backwards compatible, but unfortunately I do not have any projects to try it on.

Since the migrations can either be specified as objects or wildcard paths, hashing the file contents was not an option, so I ended up using `sha1` from `object-hash`.

Not sure if this was intentional, but [this](https://github.com/typeorm/typeorm/blob/3bd211c/src/migration/MigrationExecutor.ts#L325) looked like a runaway promise so I fixed it:

```typescript
        if (this.connection.driver instanceof MongoDriver) {  
               const mongoRunner = queryRunner as MongoQueryRunner;
/* no await */ mongoRunner.databaseConnection.db(this.connection.driver.database!).collection(this.migrationsTableName).insert(values);               
        } else {
```


docs/migrations.md
```diff
@@ -167,6 +168,10 @@ typeorm migration:revert
 This command will execute `down` in the latest executed migration. 
 If you need to revert multiple migrations you must call this command multiple times. 
 
+## Migration hash
+
+A hash checksum will be calculated for every executed migration and stored in the `migrations` table. If there is a hash mismatch, the migration will not be executed and the migration process aborted. Alternatively, a `--force` flag can be used to ignore the hash check. The `object-hash.sha1()` function is used to generate a checksum. Additionally, a migration process will be aborted if a source migration cannot be found for an already executed migration. This is to help ensure a reproducible database environent by preventing developers from making changes in the past migrations, which might have already been executed in a production environment. More details about this functionality is available in issue #3375.
```
pleerock: instead of `issue #3375` it should be `[here](https://github.com/typeorm/typeorm/pull/3375)`

src/commands/MigrationRunCommand.ts
```diff
@@ -46,6 +50,7 @@ export class MigrationRunCommand implements yargs.CommandModule {
                 subscribers: [],
                 synchronize: false,
                 migrationsRun: false,
+                migrationsForce: args["force"] ? true : connectionOptions.migrationsForce,
```
pleerock: probably its better to check if its `args["force"]  === "1" || args["force"] === "true"`

src/commands/MigrationRevertCommand.ts
```diff
@@ -45,6 +49,7 @@ export class MigrationRevertCommand implements yargs.CommandModule {
                 subscribers: [],
                 synchronize: false,
                 migrationsRun: false,
+                migrationsForce: args["force"] ? true : connectionOptions.migrationsForce,
```
pleerock: probably its better to check if its args["force"] === "1" || args["force"] === "true"

pleerock:
>  did my best to try to make it backwards compatible, but unfortunately I do not have any projects to try it on.

this is the most important part of this PR. We cannot merge it until we 100% sure it won't break anything, so we need to carefully and deeply test everything before merging it. Please make sure to setup some dummy projects and make any possible test.

src/driver/sqlserver/SqlServerDriver.ts
```diff
@@ -166,6 +166,7 @@ export class SqlServerDriver implements Driver {
         version: "int",
         treeLevel: "int",
         migrationId: "int",
+        migrationHash: "char",
```
pleerock: is this column being created as nullable? (because I guess it will break exist tables in the case if its not)

src/commands/MigrationRunCommand.ts
jeremija: Would setting the type to `boolean` suffice instead?

```
> require('yargs').option('force', { default: false, type: 'boolean' }).parse(['--force']).force
true
> require('yargs').option('force', { default: false, type: 'boolean' }).parse(['--force', 'false']).force
false
```

The only difference is that any other argument other than `"false"` would be recognized as `true`, so `--force 0` would evaluate to true. Are there any other options where `--option 0` is already being treated as false?

docs/migrations.md
jeremija: Done!

src/driver/sqlserver/SqlServerDriver.ts
jeremija: It is being create as a nullable first, then hashes of all available existing migrations are calculated, and then it is being set to `not null`.

I just added a change which will only set it to `not null` when all hashes are set.

jeremija:
> this is the most important part of this PR. We cannot merge it until we 100% sure it won't break anything, so we need to carefully and deeply test everything before merging it. Please make sure to setup some dummy projects and make any possible test.

I understand. I've added the test cases I could think of under the `test/github-issues/3375`. Will try to  come up some more cases.

src/commands/MigrationRunCommand.ts
pleerock: thats okay

src/migration/MigrationExecutor.ts
```diff
@@ -1,15 +1,25 @@
...
 import {MssqlParameter} from "../driver/sqlserver/MssqlParameter";
 import {SqlServerConnectionOptions} from "../driver/sqlserver/SqlServerConnectionOptions";
 import {PostgresConnectionOptions} from "../driver/postgres/PostgresConnectionOptions";
-import { MongoDriver } from "../driver/mongodb/MongoDriver";
-import { MongoQueryRunner } from "../driver/mongodb/MongoQueryRunner";
+import {MongoDriver} from "../driver/mongodb/MongoDriver";
+import {MongoQueryRunner} from "../driver/mongodb/MongoQueryRunner";
+import {sha1} from "object-hash";
+
+export const COLUMN_CREATE_DATE = "create_date";
```
pleerock: can we remove those constants and move strings where they are used? I don't want to have in the project unknown `COLUMN_CREATE_DATE` when I search "column".

src/migration/MigrationExecutor.ts
```diff
@@ -1,15 +1,25 @@
...
 import {QueryRunner} from "../query-runner/QueryRunner";
 import {SqlServerDriver} from "../driver/sqlserver/SqlServerDriver";
 import {MssqlParameter} from "../driver/sqlserver/MssqlParameter";
 import {SqlServerConnectionOptions} from "../driver/sqlserver/SqlServerConnectionOptions";
 import {PostgresConnectionOptions} from "../driver/postgres/PostgresConnectionOptions";
-import { MongoDriver } from "../driver/mongodb/MongoDriver";
-import { MongoQueryRunner } from "../driver/mongodb/MongoQueryRunner";
+import {MongoDriver} from "../driver/mongodb/MongoDriver";
+import {MongoQueryRunner} from "../driver/mongodb/MongoQueryRunner";
+import {sha1} from "object-hash";
```
pleerock: Currently by design we should import all external dependencies via `PlatformTools.load` to make it possible to use library in browser without webpack (something that is going to change in the future, but we need it for now).

pleerock:
This is important change so adding everyone @Kononnable @vlapo @havenchyk @rustamwin

src/migration/MigrationExecutor.ts
```diff
@@ -284,14 +384,18 @@ export class MigrationExecutor {
...
+            const hash = this.calculateHash(migration);
+            return new Migration(undefined, migrationTimestamp, migrationClassName, hash, migration);
         });
 
         // sort them by timestamp
         return migrations.sort((a, b) => a.timestamp - b.timestamp);
     }
 
+    protected calculateHash(migration: MigrationInterface): string {
+        return sha1(migration);
```
rustamwin: What about use `RandomGenerator.sha1`

src/migration/MigrationExecutor.ts
vlapo: IMHO there is so much columns in typeorm codebase. I think 5 more is not so bad (of course if it helps to readable of code). Maybe move it inside of class or call it `COLUMN_NAME_*`? Just thinking out loud.

2r50gm

src/migration/MigrationExecutor.ts
vlapo: What library is `RandomGenerator.sha1`? `object-hash` looks like reasonable choice. Popular on http://npmjs.com and support node & browser.

docs/migrations.md
```diff
@@ -167,6 +168,10 @@ typeorm migration:revert
 This command will execute `down` in the latest executed migration. 
 If you need to revert multiple migrations you must call this command multiple times. 
 
+## Migration hash
+
+A hash checksum will be calculated for every executed migration and stored in the `migrations` table. If there is a hash mismatch, the migration will not be executed and the migration process aborted. Alternatively, a `--force` flag can be used to ignore the hash check. The `object-hash.sha1()` function is used to generate a checksum. Additionally, a migration process will be aborted if a source migration cannot be found for an already executed migration. This is to help ensure a reproducible database environent by preventing developers from making changes in the past migrations, which might have already been executed in a production environment. More details about this functionality is available [here](https://github.com/typeorm/typeorm/issues/3375).
```
vlapo: Just typo environent => environment

test/github-issues/3375/issue-3375.ts
```diff
@@ -0,0 +1,141 @@
...
+}
+
+describe("github issues > #3375 add metadata to migrations table", () => {
+    let connections: Connection[];
+
+    async function createConnections(migrationsPath = __dirname + "/migration/*.js", dropSchema = true) {
+        return createTestingConnections({
+            entities: [__dirname + "/entity/*{.js,.ts}"],
+            migrations: [migrationsPath],
+            enabledDrivers: ["postgres", "mssql", "mysql", "mariadb", "sqlite", "mongodb"],
```
vlapo: If we support migration on all drivers we should remove this option and run tests for all of them.

test/github-issues/3375/issue-3375.ts
```diff
@@ -0,0 +1,141 @@
...
+            try {
+                await connection.runMigrations();
+            } catch (err) {
+                error = err;
+            }
+            expect(error).to.be.ok;
+            expect(error!.message).to.match(/migration hash/i);
+        })));
+
+        it("should succeed when migrationIgnoreHash is truthy", () => Promise.all(connections.map(async connection => {
```
vlapo: migrationsForce instead of migrationIgnoreHash?

src/migration/MigrationExecutor.ts
rustamwin: it is typeorm util
https://github.com/typeorm/typeorm/blob/0860194ead50c266adec696d0220f41eed5e550e/src/util/RandomGenerator.ts#L15-L16

src/migration/MigrationExecutor.ts
```diff
@@ -232,28 +261,99 @@ export class MigrationExecutor {
...
+            .getRawOne();
+
+            if (count === 0) {
+                await queryRunner.changeColumn(this.migrationsTable, COLUMN_HASH, new TableColumn({
+                    name: COLUMN_HASH,
+                    type: this.connection.driver.normalizeType({type: this.connection.driver.mappedDataTypes.migrationHash}),
+                    isNullable: false,
+                    length: "40",
+                }));
+            }
```
vlapo: Maybe comment about this use case and nullable should be fine.

src/migration/MigrationExecutor.ts
```diff
@@ -72,9 +85,14 @@ export class MigrationExecutor {
         // find all migrations that needs to be executed
         const pendingMigrations = allMigrations.filter(migration => {
             // check if we already have executed migration
-            const executedMigration = executedMigrations.find(executedMigration => executedMigration.name === migration.name);
-            if (executedMigration)
+            const executedMigration = executedMigrationsByName[migration.name];
+            if (executedMigration) {
+                delete executedMigrationsByName[migration.name];
+                if (!this.migrationsForce && executedMigration.hash !== migration.hash) {
+                    throw new Error(`Migration hash for ${executedMigration.name} does not match: ${executedMigration.hash} !== ${migration.hash}`);
```
vlapo: Add some info about force option? Something like "Use force if you want to skip this check". Maybe better message

vlapo:
@jeremija really great job. Love to see so high quality PR on typeorm. Hope you will continue and improve typeorm with us. 

Just one question. What about rollback and `undoLastMigration` method? Should we not add hash consistency check too?

src/driver/sqlserver/SqlServerDriver.ts
Kononnable: Why column type changes depending on the driver? It's char in some, text in sqlite, int in mongo.

src/migration/MigrationExecutor.ts
```diff
@@ -41,6 +52,7 @@ export class MigrationExecutor {
 
         const options = <SqlServerConnectionOptions|PostgresConnectionOptions>this.connection.driver.options;
         this.migrationsTableName = connection.options.migrationsTableName || "migrations";
+        this.migrationsForce = !!connection.options.migrationsForce;
```
Kononnable: Connection options should have boolean value, so no `!!` necesarry. Even if someone passed other type it should be parsed as soon as possible to boolean type.

src/migration/MigrationExecutor.ts
```diff
@@ -232,28 +261,99 @@ export class MigrationExecutor {
...
+                await queryRunner.changeColumn(this.migrationsTable, COLUMN_HASH, new TableColumn({
+                    name: COLUMN_HASH,
+                    type: this.connection.driver.normalizeType({type: this.connection.driver.mappedDataTypes.migrationHash}),
+                    isNullable: false,
+                    length: "40",
+                }));
+            }
+        }
+
+        const createDateColumnExist = table!.columns.some(c => c.name.toLowerCase() === COLUMN_CREATE_DATE);
```
Kononnable: Is `toLowerCase()` necesary here? It isn't obvious that COLUMN_CREATE_DATE have to be lowercased(and it's extracted to const so it would be easier to change in the future).

src/migration/MigrationExecutor.ts
```diff
@@ -232,28 +261,99 @@ export class MigrationExecutor {
...
+                default: this.connection.driver.mappedDataTypes.createDateDefault,
+                precision: this.connection.driver.mappedDataTypes.createDatePrecision,
+                isNullable: true,
+            }));
+        }
+    }
+
+    protected async updateHashes(queryRunner: QueryRunner): Promise<void> {
+        const executedMigrations = await this.loadExecutedMigrations(queryRunner);
+        const migrations = this.getMigrations().filter(migration => executedMigrations.find(executedMigration => executedMigration.name === migration.name));
```
Kononnable: Is it me or is that line hard to read? You're looking for migrations which wasn't executed? It should be possible to rewrite it to more readeable format or just use more precise name - you're not selecting all migrations, just portion of them.

src/migration/MigrationExecutor.ts
havenchyk: typeorm/src/util/RandomGenerator.ts seems okay to me, no need to install another package, but tbh I would choose a battery tested solution instead

test/github-issues/3375/issue-3375.ts
```diff
@@ -0,0 +1,141 @@
...
+        return createTestingConnections({
+            entities: [__dirname + "/entity/*{.js,.ts}"],
+            migrations: [migrationsPath],
+            enabledDrivers: ["postgres", "mssql", "mysql", "mariadb", "sqlite", "mongodb"],
+            schemaCreate: true,
+            dropSchema
+        });
+    }
+
+    function setup(migrationsPath = __dirname + "/migration/*.js", dropSchema = true) {
```
Kononnable: This looks weird for me. `beforeEach`,`afterEach` specified in a separate function with name that doesn't exactly gives any new information. Mayebe we should add another `describe` block for it.

src/util/CollectionUtils.ts
```diff
@@ -0,0 +1,18 @@
...
+     */
+    static indexBy<T>(list: T[], key: keyof T): {[key: string]: T} {
+        return list.reduce((obj, item) => {
+            const id = item[key] as any;
+            if (obj.hasOwnProperty(id)) {
+                throw new Error(`Key "${id}" is not unique!`);
+            }
+            obj[id] = item;
+            return obj;
+        }, {} as {[key: string]: T});
```
havenchyk: I'm not a typescript expert, but a few points/questions:

1) isn't `[key: string]: T` a `Record<string, T>`? 
2) shouldn't `const id = item[key] as any` be number or string instead of `any`?

Kononnable:
Is there any way to enable migrations in browser? If there is we should test there before merging it.

Tbh. I'm not sure this feature is really orm responsibility. If someone changed already applied migration it should be checked 'near' commit phase(not when running migration).
If this gets merged we would lock ourselves to current implementation - if we decide to change/extend MigrationInterface it will break that functionality. Right now it would be hard but it is possible.
There also might be cases where changes in already existing migrations are necessary. Let's say we have product which is installed in few environments. We've got migration that insert some rows to a table and it depends on selected environment(external code; out of migration file). No one will be able to even change variable name in that external code if it is used in migration somewhere.

vlapo:
Ping @jeremija 🏓 How is this looking? Could you finish this PR?

jeremija:
Hi, sorry for the delay. I've been travelling without internet and will
have a look when I get back!

ruscon:
@jeremija 
I have one issue with typeorm@latest and with your branch too (for mongodb)

```bash
# install deps
npm i npx typeorm ts-node typescript

npx typeorm migration:create -n first
# Migration src/migration/1549667304345-first.ts has been generated successfully.

npx typeorm migration:create -n second
# Migration src/migration/1549667388712-second.ts has been generated successfully.

# with the @latest version, it will perform all migrations, but only one migration will be performed with your branch!
npx ts-node ./node_modules/typeorm/cli.js migration:run
# 0 migrations are already loaded in the database.
# 2 migrations were found in the source code.
# 2 migrations are new migrations that needs to be executed.
# Migration first1549667304345 has been executed successfully.
# Migration second1549667388712 has been executed successfully.
# (node:88910) DeprecationWarning: collection.insert is deprecated. Use insertOne, insertMany or bulkWrite instead.

# the second call is needed only in your branch
npx ts-node ./node_modules/typeorm/cli.js migration:run
# 1 migrations are already loaded in the database.
# 2 migrations were found in the source code.
# first1549667304345 is the last executed migration. It was executed on Sat Feb 09 2019 01:08:24 GMT+0200 (EET).
# 1 migrations are new migrations that needs to be executed.
# Migration second1549667388712 has been executed successfully.
# (node:88933) DeprecationWarning: collection.insert is deprecated. Use insertOne, insertMany or bulkWrite instead.

# will revert the first migration, but should revert the second
npx ts-node ./node_modules/typeorm/cli.js migration:revert
# 2 migrations are already loaded in the database.
# first1549667304345 is the last executed migration. It was executed on Sat Feb 09 2019 01:08:24 GMT+0200 (EET).
# Now reverting it...
# Migration first1549667304345 has been reverted successfully.

# will revert the second migration, but should revert the first
npx ts-node ./node_modules/typeorm/cli.js migration:revert
# 1 migrations are already loaded in the database.
# second1549667388712 is the last executed migration. It was executed on Sat Feb 09 2019 01:09:48 GMT+0200 (EET).
# Now reverting it...
# Migration second1549667388712 has been reverted successfully.
```

can you check ?

ruscon:
@jeremija ping

src/commands/MigrationRunCommand.ts
jeremija: Looks like an hour later you wrote the opposite in your review for `MigrationRevertCommand`:

> probably its better to check if its args["force"] === "1" || args["force"] === "true"

But in this case simply specifying `--force` in the CLI would not set the force arg to true. One would have to write either `--force true` or `--force 1`. I can make it that way, just think it might be confusing, since usually specifying a boolean `--flag` enables the flag in most CLI utilities I've used so far.

This is the way parsing currently works:

```
> var yargs = require('yargs').option('force', { default: false, type: 'boolean' })
undefined
> yargs.parse(['--force', 'true']).force
true
> yargs.parse(['--force', 'false']).force
false
> yargs.parse(['--force', '1']).force
true
> yargs.parse(['--force', '0']).force
true
```

src/driver/sqlserver/SqlServerDriver.ts
jeremija: IIRC, I do not think that the types in `MongoDriver` are actually being used. There is a comment:

> Mongodb does not need to have a strong defined mapped column types because they are not used in schema sync.

I was just using `int` because all other fields were set to `int`, and they are not all `int` columns.

As for SQLite, according to my understanding of the documentation [here](https://www.sqlite.org/datatype3.html), specifying `char` will result with a column of type `text`, so I just used text.

> The following table shows how many common datatype names from more traditional SQL implementations are converted into affinities by the five rules of the previous section. This table shows only a small subset of the datatype names that SQLite will accept. Note that numeric arguments in parentheses that following the type name (ex: "VARCHAR(255)") are ignored by SQLite - SQLite does not impose any length restrictions (other than the large global SQLITE_MAX_LENGTH limit) on the length of strings, BLOBs or numeric values.
>
> Example Typenames From TheCREATE TABLE Statement or CAST Expression:  CHARACTER(20), VARCHAR(255), VARYING CHARACTER(255), NCHAR(55), NATIVE, CHARACTER(70), NVARCHAR(100), TEXT, CLOB
> 
> Resulting Affinity: TEXT

src/migration/MigrationExecutor.ts
jeremija: I'm used to extracting repeated strings into constants to prevent typos. I can revert it back to strings, or change to `COLUMN_NAME_*`, whichever you prefer.

src/migration/MigrationExecutor.ts
jeremija: Thanks, didn't know about that. Should be fixed!

docs/migrations.md
jeremija: Thanks, should be fixed!

test/github-issues/3375/issue-3375.ts
jeremija: I just removed it. Are CI tests run on Oracle?

test/github-issues/3375/issue-3375.ts
jeremija: True, this was an old name of the variable. Changed it!

src/migration/MigrationExecutor.ts
jeremija: Thanks, I just pushed a change.

src/migration/MigrationExecutor.ts
jeremija: All of the `BaseConnectionOptions` are optional (they can also be `undefined`). Removing the `!!` results in a compile error and I'd also have to set the private `migrationsForce` variable to `boolean | undefined`.

src/migration/MigrationExecutor.ts
jeremija: Looks like my local tests will pass if I remove this check. I did this because Oracle keeps all the column names in uppercase, but I have no way to test that locally.

src/util/CollectionUtils.ts
jeremija: I fixed 1). For 2) I'd rather have TypeScript check at compile time if the value of a specified key is a string or number and fail if it is not, but I'm not sure what's the best way to specify this. Do you have a suggestion?

jeremija:
Sorry for the delay.

@ruscon I believe you are right, create_date should be used over timestamp while sorting because the migrations might not be necessarily executed in the same order. If a patch is merged upstream, it could have a migration that was created earlier than the latest migration in upsteam, but will be executed later.

@havenchyk I'll write the comment here about `object-hash` because I think it's important. I'd be the happiest if there was an easy way to create a `sha1` of the source file, rather than an object, but by the time the `MigrationExecutor` is created, the `this.connection.migrations` is already populated with migration objects. It gets even more complicated because the `BaseConnectionOptions` has migrations defined as:

```typescript
    readonly migrations?: (Function|string)[];
```

So if we can already have a function, what would be the best way to calculate a hash? `Function.toString()` does not seem like the best idea so I'm all ears for suggestions 🙂 

---

Also, I just merged upstream/master to this branch because of a conflict in `package-lock.json` and I had to cast `{hash} as any` to prevent compilation errors (commit 9ae17aa). Not sure how else to get rid of this error:

```
Argument of type '{ hash: string; }' is not assignable to parameter of type 'QueryDeepPartialEntity<any>'.
  Property 'hash' is incompatible with index signature.
    Type 'string' is not assignable to type 'QueryDeepPartialEntity<any> | (() => string) | QueryDeepPartialEntity<{}>[] | ReadonlyArray<Query...'.
```

test/github-issues/3375/issue-3375.ts
jeremija: I just noticed the `mongodb` driver has this set up in both `ormconfig.json` and `ormconfig.travis.json`:

```
    "disabledIfNotEnabledImplicitly": true,
```

Does this mean that tests would not run on mongodb if we remove `enabledDrivers`?

ruscon:
@jeremija any news ?

jeremija:
Hi @ruscon, I believe sorting by `id` in `getLatestExecutedMigration` is causing the issue you described:

```typescript
    /**
     * Finds the latest migration (sorts by id) in the given array of migrations.
     */
    protected getLatestExecutedMigration(migrations: Migration[]): Migration|undefined {
        const sortedMigrations = migrations.map(migration => migration).sort((a, b) => ((a.id || 0) - (b.id || 0)) * -1);
        return sortedMigrations.length > 0 ? sortedMigrations[0] : undefined;
    }
```

Since loading migrations from MongoDB is handled differently, the objects look different. Notice how it has no `id` fields, but has `_id`:

```javascript
sqlite [ Migration {
    id: 1,
    timestamp: 1546882838853,
    name: 'Test1546882838853',
    instance: undefined,
    hash: '2e11f68db800cb65dccec7c1040fd8747b06cfdf' } ]
mongodb [ { _id: 5c95fc8fd6229df06c483c34,
    timestamp: 1546882838853,
    name: 'Test1546882838853',
    create_date: 2019-03-23T09:29:51.359Z,
    hash: '2e11f68db800cb65dccec7c1040fd8747b06cfdf' } ]
postgres [ Migration {
    id: 1,
    timestamp: 1546882838853,
    name: 'Test1546882838853',
    instance: undefined,
    hash: '2e11f68db800cb65dccec7c1040fd8747b06cfdf' } ]
mariadb [ Migration {
    id: 1,
    timestamp: 1546882838853,
    name: 'Test1546882838853',
    instance: undefined,
    hash: '2e11f68db800cb65dccec7c1040fd8747b06cfdf' } ]
mysql [ Migration {
    id: 1,
    timestamp: 1546882838853,
    name: 'Test1546882838853',
    instance: undefined,
    hash: '2e11f68db800cb65dccec7c1040fd8747b06cfdf' } ]
cockroachdb [ Migration {
    id: 1,
    timestamp: 1546882838853,
    name: 'Test1546882838853',
    instance: undefined,
    hash: '2e11f68db800cb65dccec7c1040fd8747b06cfdf' } ]
```

>  I have one issue **typeorm@latest** and with your branch too (for mongodb)

Since this is happening on typeorm@latest as well, I would suggest you to open a separate issue about this.

imnotjames:
I believe this is now considered stale.  I will be closing it out as it's been abandoned.  If anyone would like to open a new PR with this functionality it can be considered.

Issue #2708, opened 2018-08-24
Bug in queryBuildier when summing


elderapo:
**Issue type:**

[ ] question
[x] bug report
[ ] feature request
[ ] documentation issue

**Database system/driver:**

[ ] `cordova`
[ ] `mongodb`
[ ] `mssql`
[x] `mysql` / `mariadb`
[ ] `oracle`
[ ] `postgres`
[ ] `sqlite`
[ ] `sqljs`
[ ] `react-native`
[ ] `expo`

**TypeORM version:**

[ ] `latest`
[ ] `@next`
[x] `0.3.0-alpha.7`, `0.3.0-alpha.10`, `0.2.7`

**Steps to reproduce or a small repository showing the problem:**

```ts
import {
  Column,
  ConnectionOptions,
  createConnection,
  Entity,
  ManyToOne,
  OneToMany,
  PrimaryGeneratedColumn
} from "typeorm";

@Entity()
export class User {
  @PrimaryGeneratedColumn()
  id: number;
}

const options: ConnectionOptions = {
  type: "mysql",
  host: "localhost",
  port: 6667,
  username: "test",
  password: "test",
  database: "test",

  synchronize: true,
  entities: [User]
};

createConnection(options).then(
  async connection => {
    await connection.synchronize(true);

    const u1 = new User();

    await connection.getRepository(User).save(u1);

    {
      const raw = await connection
        .getRepository(User)
        .createQueryBuilder("user")
        .select("SUM(user.id)", "sum")
        .getRawOne();

      const { sum } = raw;

      console.log("typeof sum", typeof sum); /// "string" ???
      console.log("sum === 1", sum === 1); // false becuase "1" !== 1
      console.log("res", raw);
    }

    {
      const raw = await connection
        .getRepository(User)
        .createQueryBuilder("user")
        .where("user.id = :id", { id: 1 })
        .getRawOne();

      console.log("typeof raw.user_id", typeof raw.user_id); // "number" - ok
      console.log("raw.user_id === 1", raw.user_id === 1); // true - ok
    }
  },
  error => console.log("Cannot connect: ", error)
);
```

elderapo:
SQL passed to [this queryRunner.query](https://github.com/typeorm/typeorm/blob/master/src/query-builder/SelectQueryBuilder.ts#L1903) is ```"SELECT SUM(`user`.`id`) AS `sum` FROM `user` `user`"```.

Same query works just fine with normal `mysql` library:

```js
[...]

PR #3391, opened 2019-01-09
Issue #2708 testcase

elderapo:
Test fails for all drivers except for `SqliteDriver`.


imnotjames:
Closed #2708 as not a bug & working as intended.

PR #3394, opened 2019-01-09
Fix #2770

### LINKED ISSUE #2770: NTEXT, TEXT, IMAGE columns deprecated in mssql as of 2008
**Issue type:**

[ ] question
[ x ] bug report
[ x? ] feature request
[ ] documentation issue

**Database system/driver:**

[ ] `cordova`
[ ] `mongodb`
[ x ] `mssql`
[ ] `mysql` / `mariadb`
[ ] `oracle`
[ ] `postgres`
[ ] `sqlite`
[ ] `sqljs`
[ ] `react-native`
[ ] `expo`

**TypeORM version:**

[ ] `latest`
[ ] `@next`
[ x ] `0.2.6` (or put your version here)

1. Create an entity with a column @Column({ type: "simple-json" })
2. Using EntityManager.query, query the database using the column in JSON_QUERY / JSON_VALUE function.

Causes issues since these functions seem to be expecting NVARCHAR.

According to [here](https://docs.microsoft.com/en-us/sql/t-sql/data-types/ntext-text-and-image-transact-sql?view=sql-server-2017) NTEXT, TEXT, and IMAGE types are deprecated as of 2008.

Workaround for this right now is just to use CAST( \<columnName\> as NVARCHAR(MAX)).

Will take a look if I can find some time.

leoiii12:
Fix #2770
Modified simple-array and simple-json to use nvarchar(MAX) in mssql

https://docs.microsoft.com/en-us/sql/t-sql/data-types/ntext-text-and-image-transact-sql?view=sql-server-2017

> IMPORTANT! ntext, text, and image data types will be removed in a future version of SQL Server. Avoid using these data types in new development work, and plan to modify applications that currently use them. Use nvarchar(max), varchar(max), and varbinary(max) instead.

One more reason for using nvarchar(max) is STRING_SPLIT is applicable to nvarchar only. In the following example, Tags is a simple-array column.
```sql
SELECT ProductId, Name, Tags  
FROM Product  
WHERE EXISTS (SELECT *  
    FROM STRING_SPLIT(Tags, ',')  
    WHERE value IN ('clothing', 'road')); 
```

https://docs.microsoft.com/en-us/sql/t-sql/functions/string-split-transact-sql?view=sql-server-2017

### DIFF
--- a/src/driver/sqlserver/SqlServerDriver.ts
+++ b/src/driver/sqlserver/SqlServerDriver.ts
@@ -434,7 +434,8 @@ export class SqlServerDriver implements Driver {
             return "uniqueidentifier";
 
         } else if (column.type === "simple-array" || column.type === "simple-json") {
-            return "ntext";
+            column.length = "MAX";
+            return "nvarchar";
 
         } else if (column.type === "dec") {
             return "decimal";


src/driver/sqlserver/SqlServerDriver.ts
```diff
@@ -434,7 +434,8 @@ export class SqlServerDriver implements Driver {
             return "uniqueidentifier";
 
         } else if (column.type === "simple-array" || column.type === "simple-json") {
-            return "ntext";
+            column.length = "MAX";
```
vlapo: I dont think so this is right place to modify `column.length`. I am not 100% sure but this should be moved to `getColumnLength` method of driver.

pleerock: this is definitely bad place for column length assignment to be. Also its not clear what `column` is in this method. Those places were developed by @AlexMesser and unfortunately I don't know most of the details.

pleerock: @leoiii12 we already have similar logic for uuid in mysql. We have code [here](https://github.com/pleerock/typeorm/blob/7922f12da3cba3d6605929a7fea93cdf88c2d98e/src/driver/mysql/MysqlDriver.ts#L716) and [here](https://github.com/typeorm/typeorm/blob/7922f12da3cba3d6605929a7fea93cdf88c2d98e/src/driver/mysql/MysqlDriver.ts#L585). Can't we do same in this case?

vlapo: Problem is with `getColumnLength` where first argument is possible `TableColumn` and this object contains only normalized data type (database data type).

imnotjames:
@leoiii12 are  you still working on this PR?  It seems like you deleted the upstream repo, so I have a feeling you've stopped.

hktv-mkchoi:
Please close.

leoiii12:
please close. thank you.

PR #3396, opened 2019-01-09
Delete .DS_Store

RobinCK:


pleerock:
hmmm who pushed that. Thanks!

PR #3401, opened 2019-01-10
Fixes decorator reference to use the renamed synchronize instead of skipSync

joelgallant:


vlapo:
Description should be updated too - right now it is really confusing.
There is example under paragraph you are updating. Please update it properly.

joelgallant:
You're totally right, my bad. Should be better now.

PR #3409, opened 2019-01-12
Config SSL parameter in .env

osdiab:
I noticed that i cannot add the SSL parameter when configuring through .env, so this allows that, where anything except the input "true" defaults to false.

This is useful for me since I am on Heroku postgres and it by default only accepts ssl connections.

~~TODO: add test case~~ after looking at existing tests for this, they seem entirely threadbare; so i guess not high pri https://github.com/osdiab/typeorm/tree/osdiab/config-ssl-in-env/test/functional/connection-options-reader


jussikinnula:
I’ll check the SSL parameters. A month ago we revisited env options, might be that SSL needs some love also :-)

rustamwin:
You can pass this option `TYPEORM_DRIVER_EXTRA`

osdiab:
Thank you @rustamwin ! adding this to my `.env` worked:

```bash
TYPEORM_DRIVER_EXTRA='{ "ssl": true }'
```

This really ought to be documented. will add something there in a PR.

PR #3411, opened 2019-01-12
Fix mongodb config (added `useNewUrlParser`)

rustamwin:


RobinCK:
Why reformat configuration files?

pleerock:
@RobinCK integrating prettier is a hard topic because I strongly disagree with some aspects of its formatting and its not customizable enough. 

Regarding to this PR, yes, formatting needs to be reverting. PR should do what it wants to do, no more than that. Also its not possible to review changes when formatting is changed.

rustamwin:
@RobinCK you are right, i'll fix it

Issue #3393, opened 2019-01-09
Enum: migration generation problem


Evilart86:
**Issue type:**

[X] bug report

**Database system/driver:**

[X] `postgres`

**TypeORM version:**

[X] `0.2.7` (or put your version here)

**Steps to reproduce or a small repository showing the problem:**

1. Define enum:
``` ts
enum Roles {
    guest,
    user,
    admin
}
```
2. Define entity:
```ts
@Entity()
export class Role {
   @PrimaryGeneratedColumn()
    id: number;

    @Column({
        type: "enum",
        enum: Roles,
        default: [Roles.guest]
    })
    type: Roles;
}
```
3. Run command:
ts-node typeorm migration:run

> Error: invalid input value for enum role_type_enum: "[0]"

In generated migration I can see next:
```ts
await queryRunner.query(`CREATE TYPE "role_type_enum" AS ENUM('guest', 'user', 'admin', '0', '1', '2')`);
await queryRunner.query(`CREATE TABLE "role" ("id" SERIAL NOT NULL, "type" "role_type_enum" NOT NULL DEFAULT '[0]', CONSTRAINT "PK_b36bcfe02fc8de3c57a8b2391c2" PRIMARY KEY ("id"))`);
```
I think that problem is there `DEFAULT '[0]'`
Ok if I fix it to `DEFAULT 'guest'` and it works - migration:run was success. I check database, there everything is in order. 
But in next step I tried to link my user entity and role, and generate migration:
```ts
@Entity()
export class User {
    @PrimaryGeneratedColumn()
    id: number;

    @Column({unique: true})
    email: string;

    @Column()
    password: string;

    @ManyToOne(type => Role)
    role: Role;
}
```
And also I got the same error:

> QueryFailedError: invalid input value for enum role_type_enum: "[0]"

Generated Migration:
```ts
export class ImplementRoleToUser1547033224503 implements MigrationInterface {

    public async up(queryRunner: QueryRunner): Promise<any> {
        await queryRunner.query(`ALTER TABLE "user" ADD "roleId" integer`);
        await queryRunner.query(`ALTER TYPE "role_type_enum" RENAME TO "role_type_enum_old"`);
        await queryRunner.query(`CREATE TYPE "role_type_enum" AS ENUM('guest', 'user', 'admin', '0', '1', '2')`);
        await queryRunner.query(`ALTER TABLE "role" ALTER COLUMN "type" DROP DEFAULT`);
[...]

Issue #3392, opened 2019-01-09
Enum sample


Evilart86:
**Issue type:**

[X] documentation issue

**TypeORM version:**

[X] `latest`

**Steps to reproduce or a small repository showing the problem:**

Open [this file](https://github.com/typeorm/typeorm/blob/master/sample/sample11-all-types-entity/entity/EverythingEntity.ts) 
There are no samples for Enum

Issue #3323, opened 2018-12-25
Column for Array[Enum] gives Postgres & TypeOrm an error


aronraja:
**Issue type:**

[ x ] question
[ ] bug report
[ ] feature request
[ ] documentation issue

**Database system/driver:**

[ ] `cordova`
[ ] `mongodb`
[ ] `mssql`
[ ] `mysql` / `mariadb`
[ ] `oracle`
[ x ] `postgres`
[ ] `sqlite`
[ ] `sqljs`
[ ] `react-native`
[ ] `expo`

**TypeORM version:**

[ ] `latest`
[ ] `@next`
[ x ] `0.2.9` (or put your version here)

Hello there!

I have a problem with creating a `Column` for my UserEntity. I want to insert some enums inside an Array. I've looked at some issues and seemed fixed but in my case I couldn't find an answer.

**User.entity**
```typescript
@Column({
    type:"enum", 
    enum: AppRoles, 
    array: true, 
    default: AppRoles[AppRoles.USER]
})
roles: AppRoles;
```

*Roles.ts*
```typescript
export enum AppRoles {
    USER = "USER",
    ADVANCED = "ADVANCED",
    USERADMIN = "USERADMIN",
    ADMIN = "ADMIN"
}
```

*Error*
```
ExceptionHandler] type "enum[]" does not exist +77ms

QueryFailedError: type "enum[]" does not exist
    at new QueryFailedError (/DIRECTORY/src/error/QueryFailedError.ts:9:9)
    at Query.callback (/DIRECTORY/src/driver/postgres/PostgresQueryRunner.ts:175:30)
    at Query.handleError (/DIRECTORY/node_modules/pg/lib/query.js:142:17)
    at Connection.connectedErrorMessageHandler (/DIRECTORY/node_modules/pg/lib/client.js:160:17)
    at Connection.emit (events.js:182:13)
    at Connection.EventEmitter.emit (domain.js:442:20)
    at Socket.<anonymous> (/DIRECTORY/node_modules/pg/lib/connection.js:124:12)
    at Socket.emit (events.js:182:13)
    at Socket.EventEmitter.emit (domain.js:442:20)
    at addChunk (_stream_readable.js:283:12)
```

vlapo:
First `default: AppRoles[AppRoles.USER]` is really weird construction and is incorrect.

But still I think we have bug related to array like default values:

For postgres enum types this will generate error:
```typescript
@Column({
    type:"enum", 
[...]

Issue #3072, opened 2018-11-08
Entity properties which are Enums represented by `number` are returned as `string`


vtsatskin:
**Issue type:**

[ ] question
[x] bug report
[ ] feature request
[ ] documentation issue

**Database system/driver:**

[ ] `cordova`
[ ] `mongodb`
[ ] `mssql`
[ ] `mysql` / `mariadb`
[ ] `oracle`
[x] `postgres`
[ ] `sqlite`
[ ] `sqljs`
[ ] `react-native`
[ ] `expo`

**TypeORM version:**

[x] `latest`
[ ] `@next` (could not test due to breaking changes)
[x] `0.2.7`
[x] `0.2.8`

**Steps to reproduce or a small repository showing the problem:**

1. Define an Entity with an Enum property `prop`, represented as a number
2. Sync the database
3. Create an instance of the Entity and save it to the database
4. Fetch the Entity
5. Look at the `prop` type and observe it is a `string` rather than a `number`.

```ts
enum UserRole {
    USER = 1,
    ADMIN = 2
}

@Entity()
export class User {
  // ... other properties

  @Column("enum", { enum: UserRole })
  role: UserRole
}

const user = await dbConnection.manager.getRepository(User).findOne(userId);
console.log("user.role type:", typeof user.role, "; expected: number")
// output: user.role type: string ; expected: number
```

MontoyaAndres:
Hi, you could resolve it?

pleerock:
enum also can be following:

```ts
enum UserRole {
    USER = 1,
    ADMIN = 2,
    SUPERADMIN = 'superadmin'
}
```

And this case brings problems

MontoyaAndres:
@pleerock yeah, and if you change the enum to for example:

enum User {
  NAME: "USERNAME"
}

In the database should save as "USERNAME" but is saved as "NAME"

pleerock:
@MontoyaAndres if you think its a bug provide a PR with a failing test and we'll fix it

MontoyaAndres:
I did it in the slack group, but they think is not.

pleerock:
@MontoyaAndres you need to provide a PR with a failing test to confirm that its a real bug and we'll try to fix it.

PR #3414, opened 2019-01-12
fix(postgres,mysql): better support for enum and enum array

vlapo:
- better support all types of typescript enums (numeric, string, heterogeneous) + tests + docs - Fix https://github.com/typeorm/typeorm/issues/3393, Fix https://github.com/typeorm/typeorm/issues/3072
```typescript

export enum NumericEnum {
    ADMIN,
    EDITOR,
    MODERATOR,
    GHOST
}

export enum StringEnum {
    ADMIN = "a",
    EDITOR = "e",
    MODERATOR = "m",
    GHOST = "g"
}

export enum HeterogeneousEnum {
    NO = 0,
    YES = "YES",
}
```
 - default values for enum array coluns (postgres) could be now defined as array - Fix https://github.com/typeorm/typeorm/issues/3323
```typescript
@Column({
    type: "enum",
    enum: StringEnum,
    array: true,
    default: [StringEnum.ADMIN]
})
stringEnums: StringEnum[];
```
 - add enum type to sample - Fix https://github.com/typeorm/typeorm/issues/3392


src/driver/mysql/MysqlDriver.ts
```diff
@@ -471,6 +473,9 @@ export class MysqlDriver implements Driver {
 
         } else if (columnMetadata.type === "simple-json") {
             value = DateUtils.stringToSimpleJson(value);
+
+        } else if (columnMetadata.type === "enum" && !isNaN(value)) {
```
pleerock: what if value was defined as `"1"` and user expects numbers?

pleerock: I don't think we'll be able to make "heterogeneous" way possible to work. The only option I see is to support string-only or number-only. If enum was defined in a   "heterogeneous" way - we either throw error either enforce strings.

(to understand if its string or numeric based we do `Object.keys(enum).every(key => typeof enum[key] === "number")`)

pleerock:
This PR is amazing~

src/driver/mysql/MysqlDriver.ts
vlapo: Yeah I decided to force all string numbers to classic javascript numbers, so values of this enum will be transformed to number after select from databases:
```typescript
enum TestEnum {
    NO = "0",
    YES = "1"
}
```
But in line you are highlighted we can check if exists numeric enum value (`Object.keys(enum).some(key => enum[key] === parseInt(value))`) and convert it only if exists. After that change the only one edge case will be enum like this:
```typescript
enum TestEnum {
    YES1 = 1
    YES = "1"
}
```
And this looks like really edge case.

pleerock: Right, last case is really edge case and I think its okay if we don't cover it. But first case should work properly (convert "0" to string, not number).

vlapo: Updated src and added test enum with numeric values as string.

Noob7:
Doesnt work on mysql and its 2023. What am i doing wrong? 
  @Exclude()
  @Column({
    type: 'enum',
    enum: UserRole,
    array: true,
  })
  roles: UserRole;

PR #3415, opened 2019-01-13
Explain how to use the DRIVER_EXTRA field

osdiab:


osdiab:
After #3409 was resolved with @rustamwin 's advice

docs/using-ormconfig.md
```diff
@@ -124,7 +124,9 @@ List of available env variables you can set:
 * TYPEORM_ENTITIES_DIR
 * TYPEORM_MIGRATIONS_DIR
 * TYPEORM_SUBSCRIBERS_DIR
-* TYPEORM_DRIVER_EXTRA
+* TYPEORM_DRIVER_EXTRA: expects a JSON-parseable string with the variables you want to include.
+  * For example, to enable the `ssl` property, define this variable in your `.env` file like so:
+  * `TYPEORM_DRIVER_EXTRA='{ "ssl": true }'`
```
pleerock: its not clear where `ssl` options comes from. Its not an option to define all options from all drivers either.

Also, only one option is documented this way, why others aren't? It doesn't look good now.

osdiab: I personally think having useful information is better than having it look a little ugly, and while I do think it's a good idea to actually document all the different properties it is redundant with existing documentation for it.

I think this ought to be specifically called out because the dotenv configurations are the only ones in which this wildcard "extra" JSON property is used—when providing it programmatically (not through a declarative config), for instance, `ssl` is defined alongside all the rest of the properties, not in a separate "extra" sub-object. so this breaks the convention from elsewhere.

pleerock: > separate "extra" sub-object

no, its actually same. In connection options it is "extra", in env its `TYPEORM_DRIVER_EXTRA`, just like any other option - in connection options its "entities" and in env its `TYPEORM_ENTITIES_DIR`. So its consistent and if you think docs for other options are redundant, then I can tell same about this option since its covered in connection options docs

vlapo: I think we have sufficient description in connection options doc http://typeorm.io/#/connection-options
> extra - Extra connection options to be passed to the underlying driver. Use it if you want to pass extra settings to underlying database driver.

So why to describe this in two places? Maybe just provide example how to define this object as string. ``TYPEORM_DRIVER_EXTRA='{ "ssl": true }'``

pleerock:
Three people told this change isn't necessary. So Im going to close this PR. Thank you very much for your efforts!

PR #3410, opened 2019-01-12
Fix mongodb entity listeners

rustamwin:
Issues #1527

PR #3424, opened 2019-01-15
MongoDB tests

rustamwin:
#3410 should fix `load` test


pleerock:
If this depends on changes of #3410 then why its not part of it ?

rustamwin:
Okay I'm rebase it into #3410

PR #3431, opened 2019-01-16
simplifying findByIds for single primary key entities

silentroach:
Trying to use `in` operator in `findByIds` method for single primary key entities.

```js
await userRepository.findByIds([1, 2, 4]);
```

Before:
```sql
SELECT "User"."id" AS "User_id", "User"."firstName" AS "User_firstName", "User"."secondName" AS "User_secondName"
FROM "user" "User"
WHERE (("User"."id" = $1) OR ("User"."id" = $2) OR ("User"."id" = $3))
-- PARAMETERS: [1,2,4]
```

After:

```sql
SELECT "User"."id" AS "User_id", "User"."firstName" AS "User_firstName", "User"."secondName" AS "User_secondName"
FROM "user" "User"
WHERE ("User"."id" IN ($1, $2, $3))
-- PARAMETERS: [1,2,4]
```

### DIFF
--- a/src/query-builder/QueryBuilder.ts
+++ b/src/query-builder/QueryBuilder.ts
@@ -18,6 +18,7 @@ import {SqlServerDriver} from "../driver/sqlserver/SqlServerDriver";
 import {OracleDriver} from "../driver/oracle/OracleDriver";
 import {EntitySchema} from "../";
 import {FindOperator} from "../find-options/FindOperator";
+import {In} from "../find-options/operator/In";
 
 // todo: completely cover query builder with tests
 // todo: entityOrProperty can be target name. implement proper behaviour if it is.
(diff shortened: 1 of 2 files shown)


silentroach:
Something is broken, sorry, will check

vlapo:
Next time please read https://github.com/typeorm/typeorm/blob/master/DEVELOPER.md and run tests locally. I think using docker is really simple way.

silentroach:
@vlapo yep, my fault, just tested small bunch of tests in local Postgres only.

silentroach:
fixed

src/query-builder/QueryBuilder.ts
```diff
@@ -697,14 +698,28 @@ export abstract class QueryBuilder<Entity> {
...
         const metadata = this.expressionMap.mainAlias!.metadata;
+        const normalized = (Array.isArray(ids) ? ids : [ids]).map(id => metadata.ensureEntityIdMap(id));
+
+        // using in(...ids) for single primary key entities
+        if (!metadata.hasMultiplePrimaryKeys
+            && metadata.embeddeds.length === 0
+        ) {
+            const primaryColumn = metadata.primaryColumns[0];
+
+            if (!primaryColumn.transformer) {
```
pleerock: can you please explain this check? (please don't tell me "just because tests were failing" 😆 )

silentroach: seems like there is a separate bug with find options like this:

```js
where: {
    [field-with-transform]: In(...values)
}
```

it will try to transform `in` operator

pleerock: mm, interesting. Would be great to have a separate test for find options that confirm this bug... Okay, can you please put a comment in there (why you added this check)?

silentroach: yep, will add a comment and pending test to confirm a bug, ok?

silentroach: done

pr is here https://github.com/typeorm/typeorm/pull/3435

src/query-builder/QueryBuilder.ts
```diff
@@ -697,14 +698,30 @@ export abstract class QueryBuilder<Entity> {
...
+        const normalized = (Array.isArray(ids) ? ids : [ids]).map(id => metadata.ensureEntityIdMap(id));
+
+        // using in(...ids) for single primary key entities
+        if (!metadata.hasMultiplePrimaryKeys
+            && metadata.embeddeds.length === 0
+        ) {
+            const primaryColumn = metadata.primaryColumns[0];
+
+            // getEntityValue will try to transform `In`, it is a bug
+            // todo: remove this transformer check after #2390 is fixed
```
pleerock: 👍

pleerock:
We actually miss a test case that actually checks that it executed `IN` query instead of `ORs`, but that's fine, if you want you can add this test in a separate PR. I'll merge this one, thank you very much for your contribution!

PR #3442, opened 2019-01-17
Fix DriverUtils.buildDriverOptions

jussikinnula:
Fix DriverUtils.buildDriverOptions static method to actually assign TYPEORM_URL originated parameters. There has been couple of past PR's for this, so now it's better to make it simpler with only one way to get the parameters.

The new logic assigns parameters by default from the options values, and secondly from parsed TYPEORM_URL originated parameters. Finally the parsed parameters are assigned on top of options.

### DIFF
--- a/src/driver/DriverUtils.ts
+++ b/src/driver/DriverUtils.ts
@@ -1,4 +1,4 @@
-/**
+    /**
  * Common driver utility functions.
  */
 export class DriverUtils {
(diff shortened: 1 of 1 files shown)


src/driver/DriverUtils.ts
```diff
@@ -14,27 +14,18 @@ export class DriverUtils {
...
-                    username: options.username || parsedUrl.username,
-                    password: options.password || parsedUrl.password,
-                    port: options.port || parsedUrl.port,
-                    database: options.database || parsedUrl.database
-                };
-                return Object.assign(urlDriverOptions, options);
-            }
+                urlDriverOptions.sid = options.sid || options.database || parsedUrl.database
+            };
+            return Object.assign(urlDriverOptions, options);
```
marines: Maybe
```suggestion
            return Object.assign({}, options, urlDriverOptions);
```
?
Otherwise empty values from options will still override parsed url params.

jussikinnula: That works even better, now I was able to even simplify the urlDriverOptions further. Please check the revised version.

marines: lgtm

src/driver/DriverUtils.ts
```diff
@@ -14,27 +14,18 @@ export class DriverUtils {
...
-                    type: options.type || parsedUrl.type,
-                    host: options.host || parsedUrl.host,
-                    username: options.username || parsedUrl.username,
-                    password: options.password || parsedUrl.password,
-                    port: options.port || parsedUrl.port,
-                    database: options.database || parsedUrl.database
-                };
-                return Object.assign(urlDriverOptions, options);
-            }
+                urlDriverOptions.sid = parsedUrl.database
```
marines:
```suggestion
                urlDriverOptions.sid = parsedUrl.database;
```

src/driver/DriverUtils.ts
```diff
@@ -14,27 +14,18 @@ export class DriverUtils {
...
+            };
```
marines:
```suggestion
            }
```

jussikinnula: Fixed.

pleerock:
Thank you. Would be great to have a test coverage for this functionality to prevent issues in the future.

PR #3446, opened 2019-01-18
translate docs into chinese

abingoal:


pleerock:
Hey! Thank you for the contribution!

Docs should be in a separate directory, e.g. `docs/chinese`

Issue #2103, opened 2018-05-08
Query builder regression in 0.2


vesse:
**Issue type:**

[ ] question
[X] bug report
[ ] feature request
[ ] documentation issue

**Database system/driver:**

[ ] `cordova`
[ ] `mongodb`
[ ] `mssql`
[ ] `mysql` / `mariadb`
[ ] `oracle`
[X] `postgres`
[ ] `sqlite`
[ ] `sqljs`
[ ] `react-native`

**TypeORM version:**

[X] `latest`
[ ] `@next`
[X] `0.1.20` (or put your version here)

**Steps to reproduce or a small repository showing the problem:**

Query builder generated SQL changed after upgrade to 0.2 in a way that affects results. This

```typescript
connection
  .getRepository(Model)
  .createQueryBuilder('model')
  .select('id')
  .whereInIds([ 1, 2 ])
  .andWhere('column = 1')
  .getMany());
```

produces 

```
WHERE ("model"."id" = $1) OR ("model"."id" = $2) AND column = 1 -- PARAMETERS: [1,2]
``` 

in 0.2.5, and 

```
WHERE ("model"."id"=$1 OR "model"."id"=$2) AND column = 1 -- PARAMETERS: [1,2]
```

in 0.1.20. Because of the parentheses over `id` conditions have changed, a different set of results is returned. If the order of the conditions is changed to

```typescript
connection
  .getRepository(Model)
  .createQueryBuilder('model')
  .select('id')
  .where('column = 1')
  .andWhereInIds([ 1, 2 ])
  .getMany());
```

the generated SQL is

```
WHERE column = 1 AND (("model"."id" = $1) OR ("model"."id" = $2)) -- PARAMETERS: [1,2]
```

for 0.2.5 and

```
WHERE column = 1 AND ("model"."id"=$1 OR "model"."id"=$2) -- PARAMETERS: [1,2]
```

for 0.1.20, and both thus return the same results.

vesse:
Maybe it's then a difference of opinion, but I don't see any practical use for the query that

```typescript
.whereInIds([1, 2, 3)
.andWhere('x = 1')
```

produces. I would expect it to return items that have id (1, 2, 3) AND that also have x = 1, but the current query returns items that have id (1, 2) + an item with id = 3 and x = 1.

Especially if we consider SQL IN query. Consider the following table:

```
db=# select * from items;
 id | x
----+---
  1 | 1
  2 | 1
  3 | NULL
  4 | NULL
[...]

PR #3447, opened 2019-01-18
fixing query builder regression #2103

silentroach:
respecting logical operators precedence in `whereInIds` (using brackets for multiple `OR` operators)


pleerock:
Thank you very much!

silentroach:
thank you for typeorm :)

Issue #3457, opened 2019-01-19
Use tslib to reduce package size


webleaf:
We should use [tslib](https://github.com/Microsoft/tslib), because it avoid duplicate declarations of things like `__generator`, `__awaiter`, `__extends`, `__assign`, etc.
It's easy. Just read [this](https://github.com/Microsoft/tslib). It reduce package size.

PR #3458, opened 2019-01-19
add tslib

webleaf:
#3457

### DIFF
--- a/package.json
+++ b/package.json
@@ -90,7 +90,7 @@
     "reflect-metadata": "^0.1.12",
     "xml2js": "^0.4.17",
     "yargonaut": "^1.1.2",
-    "tslib": "^1.7.0",
+    "tslib": "^1.9.0",
     "yargs": "^12.0.5"
   },
   "lint-staged": {
--- a/package.json
+++ b/package.json
@@ -90,6 +90,7 @@
     "reflect-metadata": "^0.1.12",
     "xml2js": "^0.4.17",
     "yargonaut": "^1.1.2",
+    "tslib": "^1.9.3",
     "yargs": "^12.0.5"
   },
   "lint-staged": {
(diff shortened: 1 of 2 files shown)


pleerock:
but users will have to install it as well, right? e.g. they must do `importHelpers: true` and install tslib in their projects manually, right?

webleaf:
@pleerock Nope.
`importHelpers: true` is ts compiler option. So it works only when you make `typeorm` build. `typeorm` npm package contains only `*.js`, `.d.ts` files and `package.json` which should contain `tslib` dependency.
I add it as dependency to package.json (if `package.json` which I change will be lie in builded `typeorm` npm package).
So it works as all npm-packages works. When user install typeorm via `npm install typeorm` it wiil be automatically dowloaded to one of two places:
- `userProject/node_modules/typeorm/node_modules/tslib`
- `userProject/node_modules/tslib`

To which place it will be automatically placed depends on `tslib` version of user project.
Or if user project have not `tslib` it will be just placed to `userProject/node_modules/typeorm/node_modules/tslib`
For example this screenshot of my `node_modules` dir tree (as you can see there are `typeorm/node_modules` dir with `debug` and `ms` dirs which I don't install manually):

pleerock:
ah okay it only adds `var tslib_1 = require("tslib")`, for some reason I was always thinking it requires a globally included shim-like library.

pleerock:
The only concern I have if mobile-based builds will continue to work.

webleaf:
@pleerock first of all, this feature is important for mobile-based and browser-based builds (reduce package size). I use nativescript, that's why I do this. In the coming hours I will check it one more time and let you know.

webleaf:
@pleerock works like a charm. Tested with [nativescript-vue-typeorm-sample](https://github.com/typeorm/nativescript-vue-typeorm-sample)

pleerock:
Okay lets try to merge it and see how it works for users.

vlapo:
Fixed https://github.com/typeorm/typeorm/issues/3457

PR #3465, opened 2019-01-20
chore(changelog): 0.2.12 changelog + add Unreleased section

vlapo:
- 0.2.12 changelog 
 - I think all PR and changes (or most of them) should include changelog record. So I added **Unreleased** section and added note to the contributing guide

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,13 +5,50 @@ however since API is already quite stable we don't expect too much breaking chan
 If we missed a note on some change or you have a questions on migrating from old version,
 feel free to ask us and community.
 
-## 0.2.12
+## Unreleased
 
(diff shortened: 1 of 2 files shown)


pleerock:
Thank you!!!!

pleerock:
> I think all PR and changes (or most of them) should include changelog record.

yeah we need to do it. We should also consider automation as you suggested before.

rustamwin:
I've variant too

0.2.13 under development <- when released should change to date

* [change type e.g Bug] [link to issue or issue num] - [description] (username)

vlapo:
Yeah I would love to use automation in future too. I will check some open source github repositories and find some simple way. I love format like angular:
```
```

Issue #3428, opened 2019-01-16
QueryBuilder/SelectQueryBuilder - last where parameter overrides others with the same name


foxdb:
**Issue type:**

[ ] question
[x] bug report
[ ] feature request
[ ] documentation issue

**Database system/driver:**

[ ] `cordova`
[ ] `mongodb`
[ ] `mssql`
[ ] `mysql` / `mariadb`
[ ] `oracle`
[x] `postgres`
[ ] `sqlite`
[ ] `sqljs`
[ ] `react-native`
[ ] `expo`

**TypeORM version:**

[ ] `latest`
[ ] `@next`
[x] `0.2.9` (or put your version here)

## Reproduction

The following code:

```ts
const [result, count] = await queryBuilder
      .where('person.name = :name', { name: 'Luke Skywalker' })
      .orWhere('person.name = :name', { name: 'Leia Organa' })
      .getManyAndCount()
```

resulted in this query:
```
(...) WHERE "person"."name" = $1 OR "person"."name" = $2
-- PARAMETERS: ["Leia Organa","Leia Organa"]
```

the query I expected was:
```
(...) WHERE "person"."name" = $1 OR "person"."name" = $2
-- PARAMETERS: ["Luke Skywalker","Leia Organa"]
```

I expected the parameters of the first `where` to be separated from the parameters of the subsequent `orWhere`. It looks like the last one overrides the rest. This happens only when using the same parameter name. Using different names fixes the issue:

```ts
const [result, count] = await queryBuilder
      .where('person.name = :name', { name: 'Luke Skywalker' })
      .orWhere('person.name = :otherName', { otherName: 'Leia Organa' })
      .getManyAndCount()
```

## Potential cause

> Please correct my understanding of the codebase, I haven't spent much time in there!

- In `SelectQueryBuilder`'s implementation, `where`, `andWhere` and `orWhere` are pushing in the query builder's `expressionMap.wheres` array. there is no issue on that side.
- then, if there are any parameters attached to the condition, they call `setParameters`. setParameters populates `expressionMap.parameters`, which is a map of `[parameterName]: parameterValue`. This map is shared amongst all the `wheres`.

As a result, when I set a new where using an existing parameter name, I interfere with an existing where's parameters.

## Potential solution

I am unsure if this should:
- be considered as a bug and fixed
- or documented ("Don't use the same parameter name")

If it gets considered as a bug, one potential solution would be to hold the `parameters` inside the `wheres` array. 

[...]

PR #3467, opened 2019-01-20
Note about reusing parameters in SelectQueryBuilder's docs

foxdb:
Documentation: actions discussion on: https://github.com/typeorm/typeorm/issues/3428

### DIFF
--- a/docs/select-query-builder.md
+++ b/docs/select-query-builder.md
@@ -274,6 +274,19 @@ is a shortcut for:
 .setParameter("name", "Timber")
 ```
 
+Note: a set of parameters is unique per-query. A parameter's value will be overriden if you set it multiple times:
+
+```typescript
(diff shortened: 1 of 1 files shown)


docs/select-query-builder.md
```diff
@@ -274,6 +274,19 @@ is a shortcut for:
 .setParameter("name", "Timber")
 ```
 
+Note: a set of parameters is unique per-query. A parameter's value will be overriden if you set it multiple times:
```
pleerock: Thank you for your contribution! But I think we just need a simple one statement, like "note: do not use same parameter names for different values across your query builder" without code examples to simplify docs.

foxdb: Updated!

pleerock:
Thank you!

vtgn:
I totally disagree on this easy solution. It is really horrible and illogical! It would have been fixed in the code.

PR #3475, opened 2019-01-21
Refactoring NativeScript driver

webleaf:
Tested with [nativescript-vue-typeorm-sample](https://github.com/typeorm/nativescript-vue-typeorm-sample). Backward compatible.

Also `NativescriptDriver.driver` property is redundant (we can access driver via `NativescriptDriver.options.driver`... no one other `Driver` have such property) and should be removed in future release.
https://github.com/typeorm/typeorm/blob/0abf794c22d8aee03c0cf2aa449a409f2679c04d/src/driver/nativescript/NativescriptDriver.ts#L28

### DIFF
--- a/src/driver/nativescript/NativescriptDriver.ts
+++ b/src/driver/nativescript/NativescriptDriver.ts
@@ -56,11 +56,9 @@ export class NativescriptDriver extends AbstractSqliteDriver {
     /**
      * Closes connection with database.
      */
-    async disconnect(): Promise<void> {
-        return new Promise<void>((ok, fail) => {
-            this.queryRunner = undefined;
-            this.databaseConnection.close().then(ok).catch(fail);
-        });
+    async disconnect() {
+        this.queryRunner = undefined;
+        return this.databaseConnection.close();
     }
 
     /**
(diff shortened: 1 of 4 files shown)


pleerock:
Think we should ask @championswimmer for review. There were reasons why it was made this way.

championswimmer:
Seems good to me. Works fine !

pleerock:
@championswimmer do you remember why we decided to specify a driver as a parameter?

webleaf:
@pleerock I am talking only about `NativescriptDriver` property, not about avoid ability to pass your driver as parameter (may be it can be useful for someone)

webleaf:
don't merge. one more commit coming soon :)

webleaf:
Make some refactoring. `nativescript-sqlite` driver (constructor and methods) returns Promises.
So promosifying them one more time is redundant. Using async/await code become clearly.

src/driver/nativescript/NativescriptDriver.ts
```diff
@@ -87,37 +85,43 @@ export class NativescriptDriver extends AbstractSqliteDriver {
...
     }
 
     /**
      * If driver dependency is not given explicitly, then try to load it via "require".
      */
     protected loadDependencies(): void {
-        this.sqlite = this.driver;
-        if (!this.driver) {
+        try {
+            this.sqlite = this.options.driver || require("nativescript-sqlite");
```
championswimmer: this require line here will break webpack builds for people who want to use typeorm/browser, and do not have `nativescript-sqlite` in node modules (which, a react-native or web-sqlite user need not have).

championswimmer:
@pleerock yes there was  reason driver was passed as parameter, and we have to continue doing that. inlining the require line cannot be allowed, it breaks for non nativescript users.

src/driver/nativescript/NativescriptDriver.ts
webleaf: @championswimmer `react-native` driver have same line and it is not problem (i'm just get warning about that and nothing breaks).
https://github.com/typeorm/typeorm/blob/0a46810a026e147e6cac80d9a5ee16ea3b4bfe67/src/driver/react-native/ReactNativeDriver.ts#L92

championswimmer: Yes but it was already added in v2.x
We cannot break builds without major version change. 

Keeping require('') lines does break react native builds, as people have reported back here.

webleaf:
@pleerock you can merge it with v3, if what championswimmer is talking about is a real problem.

pleerock:
@webleaf the problem is that I do not use typeorm with react native or nativescript and @championswimmer knows better since he implemented driver for nativescript.

For me the most dangerous is to merge something that can break functionality for exist users, and since I do not work with those platforms I afraid to merge things. I can only ask for people who worked on this implementation to review.

Ideally we need somebody who can take care maintenance of both nativescript and react native platforms.

imnotjames:
Given the danger in merging this in I'll be closing this PR until we're able to verify that changes like this won't break anything.

I'd be happy to merge in the changes flattening the promise bits - but not the `driver` changes.

PR #3495, opened 2019-01-23
Added columnAliasName to Naming strategy to customize the alias generation

t4nz:
Added a method to the naming strategy to make the generation of column aliases customizable.
`columnAliasName(aliasName: string, columnName: string): string;`

### DIFF
--- a/src/naming-strategy/DefaultNamingStrategy.ts
+++ b/src/naming-strategy/DefaultNamingStrategy.ts
@@ -34,6 +34,10 @@ export class DefaultNamingStrategy implements NamingStrategyInterface {
         return customName ? customName : propertyName;
     }
 
+    columnAliasName(aliasName: string, columnName: string): string {
+      return aliasName + "_" + columnName;
+    }
+
     relationName(propertyName: string): string {
         return propertyName;
     }
--- a/src/query-builder/SelectQueryBuilder.ts
+++ b/src/query-builder/SelectQueryBuilder.ts
@@ -1836,7 +1836,7 @@ export class SelectQueryBuilder<Entity> extends QueryBuilder<Entity> implements 
             // transform raw results into entities
             const rawRelationIdResults = await relationIdLoader.load(rawResults);
             const rawRelationCountResults = await relationCountLoader.load(rawResults);
-            const transformer = new RawSqlResultsToEntityTransformer(this.expressionMap, this.connection.driver, rawRelationIdResults, rawRelationCountResults, this.queryRunner);
+            const transformer = new RawSqlResultsToEntityTransformer(this.expressionMap, this.connection, rawRelationIdResults, rawRelationCountResults, this.queryRunner);
             entities = transformer.transform(rawResults, this.expressionMap.mainAlias!);
 
             // broadcast all "after load" events
@@ -1930,7 +1930,7 @@ export class SelectQueryBuilder<Entity> extends QueryBuilder<Entity> implements 
      * If alias length is more than 29, abbreviates column name.
      */
     protected buildColumnAlias(aliasName: string, columnName: string): string {
-        const columnAliasName = aliasName + "_" + columnName;
+        const columnAliasName = this.connection.namingStrategy.columnAliasName(aliasName, columnName);
         if (columnAliasName.length > 29 && this.connection.driver instanceof OracleDriver)
             return aliasName  + "_" + abbreviate(columnName, 2);
 
--- a/src/query-builder/relation-id/RelationIdLoader.ts
+++ b/src/query-builder/relation-id/RelationIdLoader.ts
@@ -185,7 +185,7 @@ export class RelationIdLoader {
      * If alias length is more than 29, abbreviates column name.
      */
     protected buildColumnAlias(aliasName: string, columnName: string): string {
-        const columnAliasName = aliasName + "_" + columnName;
+        const columnAliasName = this.connection.namingStrategy.columnAliasName(aliasName, columnName);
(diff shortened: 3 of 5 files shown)


src/query-builder/transformer/RawSqlResultsToEntityTransformer.ts
```diff
@@ -1,4 +1,4 @@
-import {Driver} from "../../driver/Driver";
+import { Connection } from "../../connection/Connection";
```
RobinCK: Import style is different

src/query-builder/SelectQueryBuilder.ts
```diff
@@ -1836,7 +1836,7 @@ export class SelectQueryBuilder<Entity> extends QueryBuilder<Entity> implements
             // transform raw results into entities
             const rawRelationIdResults = await relationIdLoader.load(rawResults);
             const rawRelationCountResults = await relationCountLoader.load(rawResults);
-            const transformer = new RawSqlResultsToEntityTransformer(this.expressionMap, this.connection.driver, rawRelationIdResults, rawRelationCountResults, this.queryRunner);
+            const transformer = new RawSqlResultsToEntityTransformer(this.expressionMap, this.connection, rawRelationIdResults, rawRelationCountResults, this.queryRunner);
```
RobinCK: I don't like that the whole connection should be transferred to transformer.
I like `Single responsibility principles`

src/query-builder/transformer/RawSqlResultsToEntityTransformer.ts
t4nz: Ok, I can fix it

src/query-builder/SelectQueryBuilder.ts
t4nz: Do you have any suggestions?

src/naming-strategy/DefaultNamingStrategy.ts
```diff
@@ -34,6 +34,10 @@ export class DefaultNamingStrategy implements NamingStrategyInterface {
         return customName ? customName : propertyName;
     }
 
+    columnAliasName(aliasName: string, columnName: string): string {
```
vlapo: Function name is little bit misleading and my first thought was it is about alias naming. Maybe something like `columnNameWithAlias`. Just opening discussion about it.

src/query-builder/SelectQueryBuilder.ts
```diff
@@ -1930,7 +1930,7 @@ export class SelectQueryBuilder<Entity> extends QueryBuilder<Entity> implements
      * If alias length is more than 29, abbreviates column name.
      */
     protected buildColumnAlias(aliasName: string, columnName: string): string {
-        const columnAliasName = aliasName + "_" + columnName;
+        const columnAliasName = this.connection.namingStrategy.columnAliasName(aliasName, columnName);
         if (columnAliasName.length > 29 && this.connection.driver instanceof OracleDriver)
             return aliasName  + "_" + abbreviate(columnName, 2);
```
vlapo: You are not using naming strategy for abbreviate column name.

src/query-builder/relation-id/RelationIdLoader.ts
```diff
@@ -185,7 +185,7 @@ export class RelationIdLoader {
      * If alias length is more than 29, abbreviates column name.
      */
     protected buildColumnAlias(aliasName: string, columnName: string): string {
-        const columnAliasName = aliasName + "_" + columnName;
+        const columnAliasName = this.connection.namingStrategy.columnAliasName(aliasName, columnName);
         if (columnAliasName.length > 29 && this.connection.driver instanceof OracleDriver)
             return aliasName  + "_" + abbreviate(columnName, 2);
```
vlapo: Again abbreviate name without naming strategy applied.

src/query-builder/transformer/RawSqlResultsToEntityTransformer.ts
```diff
@@ -309,8 +309,8 @@ export class RawSqlResultsToEntityTransformer {
      * If alias length is more than 29, abbreviates column name.
      */
     protected buildColumnAlias(aliasName: string, columnName: string): string {
-        const columnAliasName = aliasName + "_" + columnName;
-        if (columnAliasName.length > 29 && this.driver instanceof OracleDriver)
+        const columnAliasName = this.connection.namingStrategy.columnAliasName(aliasName, columnName);
+        if (columnAliasName.length > 29 && this.connection.driver instanceof OracleDriver)
             return aliasName  + "_" + abbreviate(columnName, 2);
```
vlapo: And again.

src/query-builder/SelectQueryBuilder.ts
t4nz: > 1. First at all is this change required? I do not see any issue where this change will help to resolve some real life issue. Lets see what @pleerock.

We try to understand if this change is needed. The problem has occurred using an upper-snake-case nomenclature with Oracle driver. 
The current buildColumnAlias function, abbreviate for aliases with longer than 29 characters, the name of the column. In doing so though in case of similar fields ex. PRODUCT_CATEGORY_ID, PRODUCT_CATALOG_ID, you will get the same result: **PR**, and then an incorrect result of the query. The need was therefore to be able to customize this function.

> You are not using naming strategy for abbreviate column name.

Not being able to check the driver inside the new naming strategy function, the only chance that I saw was using a custom implementation into the new naming strategy function that returns a string shorter than 29 characters, to skip the condition.

t4nz:
Closing in favor of #3514. Thank you guys.

Issue #3460, opened 2019-01-20
Argument of type '{ name: FindOperator<any>; }' is not assignable to parameter of type 'DeepPartial<Entity>'


barmaglot92:
**Issue type:**

[x ] question
[ ] bug report
[ ] feature request
[ ] documentation issue

**Database system/driver:**

[ ] `cordova`
[ ] `mongodb`
[ ] `mssql`
[ ] `mysql` / `mariadb`
[ ] `oracle`
[x ] `postgres`
[ ] `sqlite`
[ ] `sqljs`
[ ] `react-native`
[ ] `expo`

**TypeORM version:**

[ ] `latest`
[ ] `@next`
[x] `0.2.11`

Hello. My question is about https://github.com/typeorm/typeorm/blob/master/docs/find-options.md.
I can't use condition "In" at Entity.find method.

```
    Group.find{ name: In(["test"]) }

Argument of type '{ name: FindOperator<any>; }' is not assignable to parameter of type 'DeepPartial<Group>'.
  Types of property 'name' are incompatible.
    Type 'FindOperator<any>' is not assignable to type 'string'
```
but

```
getRepository(Group).find{ name: In(["test"]) }
still works
```

I compared Repository.find and BaseEntity.find methods interfaces and seen
1.    find(conditions?: FindConditions<Entity>): Promise<Entity[]>;
2.   static find<T extends BaseEntity>(this: ObjectType<T>, conditions?: DeepPartial<T>): Promise<T[]>;

What the reason this differences?

pleerock:
It's a bug. `BaseEntity` declaration is old and invalid. I'll push a fix.

barmaglot92:
Thank you. Will you close issue when fix will be deployed

barmaglot92:
Unfortunately typings have't updated
```
    static find<T extends BaseEntity>(this: ObjectType<T>, conditions?: DeepPartial<T>): Promise<T[]>;
```

i think some troubles with build..

barmaglot92:
It actuals only for npm package, in sources all good

ricardovf:
@pleerock can you update the NPM version? I need this fix urgent too. Thanks!

barmaglot92:
This issue still without upgrade version in npm from two weeks ago, as I see it's already released, but not deployed. I'm so interested in so interested why it's not in npm avail for now

barmaglot92:
Sorry for bad English😒

ShanniL:
[...]

PR #3503, opened 2019-01-24
Fix signatures of BaseEntity.

leak4mk0:
Fix signatures of `BaseEntity`.
It seems that an argument `criteria` and `conditions` allow `FindConditions` instead of `DeepPartial`.
Them arguments of `Repository` was changed at #1846, but one of `BaseEntity` was not changed..


leak4mk0:
Sorry.
This is duplicated with #3460.

Issue #3350, opened 2019-01-02
ER_DUP_FIELDNAME with simple find on 0.2.10


Slessi:
**Issue type:**

[x] bug report

**Database system/driver:**

[x] `mysql 5.7.24`

**TypeORM version:**

[x] `0.2.10`

**Steps to reproduce or a small repository showing the problem:**

Given the following two entities:
```ts
@Entity({ name: "payrolls" })
export class Payroll {
    @PrimaryGeneratedColumn()
    public id: number;
}

@Entity({ name: "payroll_payslips" })
export class PayrollPayslip {
    @PrimaryGeneratedColumn()
    public id: number;

    @Column({ name: "payroll_id" })
    public payrollId: number;

    @ManyToOne((type) => Payroll)
    @JoinColumn({ name: "payroll_id" })
    public payroll: Payroll;
}
```

And the following find:

```ts
dbConn.getRepository(PayrollPayslip).findOne(id, { relations: ["payroll"] });
```

I now receive `ER_DUP_FIELDNAME` after upgrading to 0.2.10

```
error: QueryFailedError: ER_DUP_FIELDNAME: Duplicate column name 'PayrollPayslip_payroll_id'
    at new QueryFailedError (<project_dir>/node_modules/typeorm/error/QueryFailedError.js:28:28)
    at Query.<anonymous> (<project_dir>/node_modules/typeorm/driver/mysql/MysqlQueryRunner.js:227:45)
    at Query.<anonymous> (<project_dir>/node_modules/mysql/lib/Connection.js:502:10)
    at Query._callback (<project_dir>/node_modules/mysql/lib/Connection.js:468:16)
    at Query.Sequence.end (<project_dir>/node_modules/mysql/lib/protocol/sequences/Sequence.js:83:24)
    at Query.ErrorPacket (<project_dir>/node_modules/mysql/lib/protocol/sequences/Query.js:90:8)
    at Protocol._parsePacket (<project_dir>/node_modules/mysql/lib/protocol/Protocol.js:278:23)
    at Parser.write (<project_dir>/node_modules/mysql/lib/protocol/Parser.js:76:12)
    at Protocol.write (<project_dir>/node_modules/mysql/lib/protocol/Protocol.js:38:16)
    at Socket.<anonymous> (<project_dir>/node_modules/mysql/lib/Connection.js:91:28)
    at Socket.<anonymous> (<project_dir>/node_modules/mysql/lib/Connection.js:502:10)
[...]

Issue #3118, opened 2018-11-16
TypeORM Postgres identifier limit is 63 bytes, column aliases get truncated


jakari:
**Issue type:**

[ ] question
[x] bug report
[ ] feature request
[ ] documentation issue

**Database system/driver:**

[ ] `cordova`
[ ] `mongodb`
[ ] `mssql`
[ ] `mysql` / `mariadb`
[ ] `oracle`
[x] `postgres`
[ ] `sqlite`
[ ] `sqljs`
[ ] `react-native`
[ ] `expo`

**TypeORM version:**

[ ] `latest`
[ ] `@next`
[x] `0.2.8`

I have found that there are some issues with column aliasing when doing queries from the database. The column names might get very long inside the actual query when they concatenate the entity name into the column name. This becomes an issue in Postgres since the maximum length of a Postgres column is 63 bytes. This is especially a problem with embedded entities.

This same 63 byte name limit restriction applies to column aliases as well.

Currently, when TypeORM creates a query it creates an alias for each column it selects. The alias is of format "tablename_columnname". Now if we have a long table name and a long column name, that are both under 63 by themselves, but concatenated over 63 bytes, TypeORM won't find the column data because of the fact that Postgres also truncates the alias to 63 bytes.

A description of the limit is found in the  [postgres docs](https://www.postgresql.org/docs/current/sql-syntax-lexical.html#SQL-SYNTAX-IDENTIFIERS)

This issue might not be limited to Postgres only, however I haven't tested other databases.

**Steps to reproduce or a small repository showing the problem:**

Repository at [https://github.com/jakari/typeorm-column-length-issue](https://github.com/jakari/typeorm-column-length-issue)

An example of the issue by doing a query in Postgres:

```
[typeorm_test=# SELECT "andHereWehaveAnTooLongEntityNameThatIsVeryDescriptive" AS "EntityWithVeryLongAndDescriptiveName_andHereWehaveAnTooLongEntityNameThatIsVeryDescriptive" FROM entity_with_very_long_and_descriptive_name;
NOTICE:  identifier "EntityWithVeryLongAndDescriptiveName_andHereWehaveAnTooLongEntityNameThatIsVeryDescriptive" will be truncated to "EntityWithVeryLongAndDescriptiveName_andHereWehaveAnTooLongEnti"
 EntityWithVeryLongAndDescriptiveName_andHereWehaveAnTooLongEnti
-----------------------------------------------------------------
                                                               1
(1 row)
```

[...]

PR #3514, opened 2019-01-25
fix(postgres): length limit for alias names #3118

VinceOPS:
Hi there!

### Context

I just made a small Pull request to address https://github.com/typeorm/typeorm/issues/3118. Postgres has a ([built-in](https://www.postgresql.org/docs/current/sql-syntax-lexical.html#SQL-SYNTAX-IDENTIFIERS)) limit of 63 characters for the name of columns and aliases used in SQL queries.

This prevent TypeORM from being able to properly transform **raw results** to Entities when querying "deeply nested" relations, as it can generate aliases as long as `"CategoryWithVeryLongName__postsWithVeryLongName__authorWithVeryLongName_firstName"` (dummy example, but still 81 characters long).

Thus, the "populated" (joined) `authorWithVeryLongName` is never converted from a raw result to an entity (and stays `undefined`).

### Solution

A similar issue has already been addressed for Oracle [here](https://github.com/typeorm/typeorm/blob/c051810bfaaf8781e28f251754ba8eb4dc922cce/src/query-builder/SelectQueryBuilder.ts#L1932) by abbreviating the name of the selected column.
My concern with this code is that when `alias` is `"CategoryWithVeryLongName__postsWithVeryLongName__authorWithVeryLongName"` and `columnName` is `"firstName"`, abbreviating `"firstName"` to 2 characters might create collisions with other columns starting with `"fi"`.
That's why I decided to write `shorten`, which is focused on shortening **aliases**.

I also added a test which guarantees that `authorWithVeryLongName` is correctly loaded/instantiated as an entity.

### Why

Some E2E tests started failing in our application when [this fix](https://github.com/typeorm/typeorm/issues/3350) was introduced. Using `__` instead of `_` (as a separator for the path of tables) lengthened the selection aliases.

Any feedback is most welcome, this is only my second PR in the project. Thanks.


pleerock:
Maybe I'm asking for too much, but can't we make an ideal and proper solution from the beginning and cover all exist drivers and database and resolve this issue forever?

src/query-builder/transformer/RawSqlResultsToEntityTransformer.ts
```diff
@@ -313,6 +314,9 @@ export class RawSqlResultsToEntityTransformer {
         if (columnAliasName.length > 29 && this.driver instanceof OracleDriver)
             return aliasName  + "_" + abbreviate(columnName, 2);
 
+        if (columnAliasName.length > 63 && this.driver instanceof PostgresDriver)
+            return `${shorten(aliasName)}_${columnName}`;
+
         return columnAliasName;
```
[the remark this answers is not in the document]
t4nz: Could not you use a unique hash on the alias, to make sure you do not have duplicates?
Ex.
```typescript
function hash(): string {
    const n = Date.now();
    const last = hash.last || n;
    hash.last = n > last ? n : last + 1;
    return hash.last.toString(36);
}

return abbreviate(aliasName, 2)  + "_" + abbreviate(columnName, 2) + "_" + hash();
```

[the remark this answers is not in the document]
VinceOPS: I guess this is a solution, yup; @pleerock? 
This deals with an other issue, which concerns potential (and already existing) collisions with long identifiers, but I don't mind holding it in this PR - two birds, one stone!

VinceOPS:
@pleerock Alright, I made a new commit (for the sake of an easier review) with a more generic approach. Let me know. Please also have a look at [t4nz's comment](https://github.com/typeorm/typeorm/pull/3514#discussion_r251335556) about "alias collisions".

Thanks

t4nz:
Any news? I don't want to be rude but the ORM with Oracle is quite unusable right now

src/util/StringUtils.ts
```diff
@@ -38,4 +38,45 @@ export function abbreviate(str: string, abbrLettersCount: number = 1): string {
...
+      // "OrderItemList" becomes "OrItLi", while "company" becomes "comp"
+      const length = segmentTerms.length > 1 ? termLength : segmentLength;
+      const shortSegment = segmentTerms.map(term => term.substr(0, length)).join("");
+
+      acc.push(shortSegment);
+      return acc;
+    }, []);
+
+    return shortSegments.join(separator);
+  }
```
vlapo: Formating issue. I see some extra two spaces here. And also we are using 4 spaces as indent style. Why is build green? 🤔

src/driver/mongodb/MongoDriver.ts
```diff
@@ -107,6 +107,12 @@ export class MongoDriver implements Driver {
      */
     dataTypeDefaults: DataTypeDefaults;
 
+    /**
+     * No documentation specifying a maximum length for identifiers could be found
+     * for MongoDB.
+     */
+    maxAliasLength: number;
```
vlapo: Maybe initialize this prop to -1 as infinity? I do not have any special reason. Just looking better :)

test/github-issues/3118/max-alias-length-in-postgres.ts
```diff
@@ -0,0 +1,52 @@
...
+ * "The system uses no more than NAMEDATALEN-1 bytes of an identifier; longer names can be
+ * written in commands, but they will be truncated. By default, NAMEDATALEN is 64 so the
+ * maximum identifier length is 63 bytes. If this limit is problematic, it can be raised
+ * by changing the NAMEDATALEN constant in src/include/pg_config_manual.h."
+ */
+describe("other issues > shorten alias names for Postgres when they are longer than 63 characters", () => {
+    let connections: Connection[];
+    before(async () => connections = await createTestingConnections({
+        entities: [__dirname + "/entity/*{.js,.ts}"],
+        enabledDrivers: ["postgres"],
```
pleerock: can we remove enabledDrivers to see if all drivers pass this test?

CHANGELOG.md
```diff
@@ -16,6 +16,7 @@ feel free to ask us and community.
 * fixed increment/decrement value of embedded entity ([#3182](https://github.com/typeorm/typeorm/issues/3182))
 * fixed missing call `transformer.from()` in case column is NULL ([#3395](https://github.com/typeorm/typeorm/issues/3395))
 * fixed signatures of `update`/`insert` methods, some `find*` methods in repositories, entity managers, BaseEntity and QueryBuilders
+* fixed "deep relations" not loaded/mapped due to the built-in max length of Postgres ([#3118](https://github.com/typeorm/typeorm/issues/3118))
```
pleerock: @vlapo another issue with changelog... I think we need some other strategy...

vlapo: yeah it is pretty annoying. I think auto generate changelog from commit messages should works better. I will check more OSS projects on github and create some simple guide about it.

pleerock: Thank you

ajxs:
> Any news? I don't want to be rude but the ORM with Oracle is quite unusable right now
I'm in the same boat, our application is coming unstuck because of this issue.

ajxs:
What is actually even more troubling, is that the relations which silently failed to be populated will overwrite values with `null` during a cascading save, which has the capability to destroy all kinds of valuable data. We're just lucky we caught this in testing.
I've already heard all about the creators views on cascading, but this is pretty standard ORM functionality, and if used can have pretty dreadful consequences because of this bug.

pleerock:
We almost ready to merge this PR, just few small things left to finish, hope @VinceOPS will find a time to take a look on this PR.

@ajxs sorry, but how is your last comment related to this PR?

ajxs:
> We almost ready to merge this PR, just few small things left to finish, hope @VinceOPS will find a time to take a look on this PR.
> 
> @ajxs sorry, but how is your last comment related to this PR?

This bug is causing data clobbering in our product. This is something you should be aware of, that under certain conditions this bug can lead to serious data corruption.

pleerock:
Please DON'T CALL IT BUG. Because its not a BUG. Its how it is designed to work. Your statements can mislead other people. And I didn't get the actual answer to my question - how is your problem related to this PR?

ajxs:
> 
> 
> Please DON'T CALL IT BUG. Because its not a BUG. Its how it is designed to work. Your statements can mislead other people. And I didn't get the actual answer to my question - how is your problem related to this PR?

Pardon me, I thought I was commenting on the 'issue'. I think I ended up here following what I thought was a 'related issue' link. I understand what you mean now.
And yes, this is a bug. If the ORM generates an invalid query, which does not populate the entity in the prescribed manner, which in turn leads to unintended, undefined behavior then it is a bug.

pleerock:
@ajxs please create a new issue if you haven't yet and link all our conversations on this topic and lets continue discussion in there

adrianhara:
Great work guys! I stumbled onto this issue by hitting the postgres limitation. One question: if I understand the code correctly the `shorten` function takes some options but it's never called with any, therefore always splitting on double-underscore `__`. In our case the segments are separated by a simple underscore `_` and won't be split/shortened correctly. One option I see is to define this separator at the driver level, same as max length and then send it down to the function when calling. What do you think, would that work or do you see a better option?

VinceOPS:
Hi guys & sorry for the delay. I'm actually open for change requests and improvements, but I cannot test this feature for Oracle. At least I don't really have an interest/time in trying to make my workstation able to run Oracle and installing the drivers required to let the Node app connect to it... :roll_eyes: 

However I'd be glad to do whatever is required for MySQL & other DB systems "directly" available through the docker-compose.

pleerock:
> Hi guys & sorry for the delay. I'm actually open for change requests and improvements, but I cannot test this feature for Oracle. At least I don't really have an interest/time in trying to make my workstation able to run Oracle and installing the drivers required to let the Node app connect to it... 🙄
> **EDIT** - Don't want to look indelicate/rude so I'd rather be accurate: any help/commit/code change is most welcome regarding Oracle, if someone can/wants to do this part of the job.

Thats absolutely fine! Thank you for all the great job you already did. We all hate Oracle too.

Please review all comments people did, see if you won't break anything, uncomment `enabledDrivers: ["postgres"]` (since all drivers should pass, right? ignore oracle in your local ormconfig by setting skip: true in ormconfig)

src/util/StringUtils.ts
VinceOPS: Yup, done!

src/driver/mongodb/MongoDriver.ts
VinceOPS: I think this can lead to counterintuitive logics like `if (maxAliasLength !== -1)`, which looks weird in TypeScript, where we'd usually set the property as optional and consider it only if a value is associated. Don't you think so?

VinceOPS: I think this can lead to very counterintuitive logic like `if (maxAliasLength !== -1)` which is kinda "C-style checking"; I'd rather set the property `optional` and only consider its numeric value when it has one. What do you think?

VinceOPS:
Hi @pleerock 

I solved the branch conflicts and handled most of the comments. Two discussions are pending:
https://github.com/typeorm/typeorm/pull/3514#discussion_r251335556
https://github.com/typeorm/typeorm/pull/3514#discussion_r254594398

The integration test has been left to "default drivers" and it helped me to notice an other issue: MySQL is throwing an error when trying to create a table whose name is longer than 64 characters (this was the case of the many-yo-many join-column, in my example). Thus, I had to apply the shortening logic to the table names too. I also added a test to check this behavior. All other tests passed too.

Any comments are most welcome.

tsconfig.json
```diff
@@ -1,29 +1,30 @@
...
-}
+{
+    "version": "2.1.1",
+    "compilerOptions": {
+        "lib": ["es5", "es6"],
+        "outDir": "build/compiled",
+        "target": "es5",
+        "module": "commonjs",
+        "moduleResolution": "node",
+        "skipLibCheck": true,
```
pleerock: can you please explain this option? Does it really bring good performance in? I remember I had some issues with this option

CHANGELOG.md
```diff
@@ -35,6 +35,7 @@ feel free to ask us and community.
 * fixed signatures of `update`/`insert` methods, some `find*` methods in repositories, entity managers, BaseEntity and QueryBuilders
 * handle embedded documents through multiple levels in mongodb ([#3551](https://github.com/typeorm/typeorm/issues/3551))
 * fixed hanging connections in `mssql` driver ([#3327](https://github.com/typeorm/typeorm/pull/3327))
+* fixed "deep relations" not loaded/mapped due to the built-in max length of Postgres ([#3118](https://github.com/typeorm/typeorm/issues/3118))
```
pleerock: we need to merge master into this branch and move this block to the newer section under 0.2.15

src/driver/sqlserver/SqlServerDriver.ts
```diff
@@ -194,6 +194,12 @@ export class SqlServerDriver implements Driver {
         "datetimeoffset": { precision: 7 }
     };
 
+    /**
+     * Max length allowed by MSSQL Server for aliases (identifiers).
+     * @see https://docs.microsoft.com/en-us/sql/sql-server/maximum-capacity-specifications-for-sql-server
+     */
+    maxAliasLength = 128;
```
pleerock: shouldn't it be in a `AbstractSqliteDriver` instead?

tsconfig.json
VinceOPS: @pleerock Hi! This is only to avoid trying to compile the code with the definitions in `node_modules` as the project is using TS 2.9 and some type definitions (I think these are those from `yargs`) are using things from TS 3+ (`unknown` etc). I can remove it if you're not bothered by the same issue on your machine / in the CI (I assume so, as it is passing)

VinceOPS:
@pleerock Should be fine now. Please let me know!

pleerock:
there are some compiler errors ci fails because of them

VinceOPS:
@pleerock 

> there are some compiler errors ci fails because of them

:ok_hand:

pleerock:
Finally merged! Thank you very much for your hard work!

adrianhara:
Could this be merged also into next?

hohler:
I noticed that it's also a problem when getting relations. Specifically in the `query-builder/RelationIdLoader.ts` we don't use this "buildColumnAlias" at all. But it can also happen, that the alias column names are longer than e.g. 63 characters there.

PR #3523, opened 2019-01-27
fixing count performance issue

jhonatanTeixeira:
Counting joined results is never a easy task, however we can use an old technique, which is to recreate the query with an count and distinct, and typeorm uses this technique as other super popular orms like doctrine, sqlalchemy and hibernate does. However, it isn't doing it the right way, when it uses SELECT COUNT(DICTINCT(`table`.`id`)) FROM table -- the database will ignore indexes and will perform a full table scan and degrade performance (it is happening on my project, when counting results on a 50000+ registries, its taking from 6 to 15 seconds on my staging machine) .
To solve that problem, i changed it to the counting style that is used on the older and traditional orms like hibernate, doctrine and entities framework, they use count on a parent query encapsulating the current query as a subquery with distinct, that way the database engine doesn't ignore the indexes of the primary key, heres an example: "SELECT COUNT(1) cnt FROM (SELECT DISTINCT table.id FROM table) as result".
Thanks

### DIFF
--- a/src/query-builder/SelectQueryBuilder.ts
+++ b/src/query-builder/SelectQueryBuilder.ts
@@ -1702,39 +1702,23 @@ export class SelectQueryBuilder<Entity> extends QueryBuilder<Entity> implements 
         const mainAlias = this.expressionMap.mainAlias!.name; // todo: will this work with "fromTableName"?
         const metadata = this.expressionMap.mainAlias!.metadata;
 
-        const distinctAlias = this.escape(mainAlias);
-        let countSql: string = "";
-        if (metadata.hasMultiplePrimaryKeys) {
-            if (this.connection.driver instanceof AbstractSqliteDriver) {
-                countSql = `COUNT(DISTINCT(` + metadata.primaryColumns.map((primaryColumn, index) => {
-                    const propertyName = this.escape(primaryColumn.databaseName);
-                    return `${distinctAlias}.${propertyName}`;
-                }).join(" || ") + ")) as \"cnt\"";
+        const querySelects = metadata.primaryColumns.map(primaryColumn => {
+            const distinctAlias = this.escape("distinctAlias");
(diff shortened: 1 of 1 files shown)


Kononnable:
>database will ignore indexes and will perform a full table scan

Why do you think so? Also as it is about performance - did you test it on each db driver or just on one you're using in your project?
.. and tests where broken by those changes so I guess it's not working as you expected.

jhonatanTeixeira:
I don't think its slower, i know it is. I tested on mysql5.7 and podtgres. On 2 large databases i have right now and i got improved performance. Also i'm already using my changes on production and a got overall speedup on api listings.
I don't know why the tests broke, as the broken tests are not related to my changes. However i will dig deeper to find out

vlapo:
🤦‍♂️

Kononnable:
Look at https://github.com/typeorm/typeorm/blob/master/CONTRIBUTING.md. If you want us to look at this PR later you have to rewrite branch history(or make a new PR with changes) - PR in its current condition is unreadable.

This is a performance issue, so we would need test it on almost every driver. Otherwise we would merge changes which can e.g. speed up things a bit on postgres and mysql, but make mssql and oracle much slower.

jhonatanTeixeira:
Yeah, i will try to improve it.
About the change. If you take a look on other orms solutions, like doctrine, jpa and sqlalchemy, you will notice they all do the same, wraps up the main query into a count query with distinct

pleerock:
@jhonatanTeixeira first of all, please fix tests. Second, please provide a detailed explanation of what we had and what we have right now. Third, please provide links that prof the concept, abstract statements like see how other orms do aren't telling us anything.

imnotjames:
Hi there!  Thanks for the contribution!

It seems like the tests have been failing and this hasn't been touched in over a year.  I'll be closing it but I'm sure it's still a great performance improvement.  If you have the time to clean it up & continue the discussion please open a new PR.

Cheers!

PR #3526, opened 2019-01-27
fix(3517): fixed undefined object id field in case property name is _id

### LINKED ISSUE #3517: mongodb: .save() method does not set _id column value
**Issue type:**

[ ] question
[x] bug report
[ ] feature request
[ ] documentation issue

**Database system/driver:**

[ ] `cordova`
[x] `mongodb`
[ ] `mssql`
[ ] `mysql` / `mariadb`
[ ] `oracle`
[ ] `postgres`
[ ] `sqlite`
[ ] `sqljs`
[ ] `react-native`
[ ] `expo`

**TypeORM version:**

[x] `latest`
[ ] `@next`

```ts
@Entity()
class EntityClass {
    @ObjectIdColumn()
    _id: ObjectId;

    @Column()
    name: string;
}

let ent = new EntityClass();
ent.name = "abc";

await connection.mongoManager.save(ent);

console.log(ent._id); // inserts new document but does not set _id. _id is undefined
```

vlapo:
We are deleting `_id` property from entity to avoid duplication of object id columns. I did not realize case when `_id` is possible property name. Now we check if property name is not same as "column" name in db.
 
Fixed #3517

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ feel free to ask us and community.
 ### Bug Fixes
 
 * fixed signatures of `update`/`insert` methods, some `find*` methods in repositories, entity managers, BaseEntity and QueryBuilders
+* fixed undefined object id field in case property name is `_id` ([3517](https://github.com/typeorm/typeorm/issues/3517))
 
 ### Features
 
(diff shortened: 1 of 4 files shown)


rustamwin:
@vlapo conflicts :sweat_smile:

PR #3529, opened 2019-01-28
Update .editorconfig

rustamwin:


vlapo:
Why only package.json? We have more *.json files in our repository.

rustamwin:
@vlapo Good point!

Issue #3538, opened 2019-01-29
ioredis/cluster 2nd(options) parameter support


scvgoe:
**Issue type:**

[ ] question
[ ] bug report
[x] feature request
[ ] documentation issue

**Database system/driver:**

[ ] `cordova`
[ ] `mongodb`
[ ] `mssql`
[ ] `mysql` / `mariadb`
[ ] `oracle`
[ ] `postgres`
[ ] `sqlite`
[ ] `sqljs`
[ ] `react-native`
[ ] `expo`

**TypeORM version:**

[x] `latest`
[ ] `@next`
[ ] `0.x.x` (or put your version here)

**Question**
Hi, I appreciate your support for ioredis/cluster.

Actually, ioredis.Cluster constructor allow two parameters (startupNodes and options).
You can check them [here](https://github.com/luin/ioredis/blob/master/API.md#new-clusterstartupnodes-options)

But I cannot inject 2nd parameter (options) to [here](https://github.com/typeorm/typeorm/pull/3364/files#diff-7ef1785809892325abcbdb5d564dce09R64).

Do you have any plan to fix this problem?

vlapo:
`ioredis` and `ioredis-cluster` caches were thankfully provided by @HussainAliAkbar and @ignacioblit. Could you please check this issue?

Also feel free to contribute.

Kononnable:
I think this issue can be closed now since, PR was merged.

PR #3541, opened 2019-01-29
Enchancement - Add support for ioredis Cluster object 2nd parameter

scvgoe:
It comes from #3538

### DIFF
--- a/docs/caching.md
+++ b/docs/caching.md
@@ -145,23 +145,32 @@ In case you want to connect to a redis-cluster using IORedis's cluster functiona
     username: "test",
     cache: {
         type: "ioredis/cluster",
-        options: [
-            {
-                host: 'localhost',
-                port: 7000,
-            },
-            {
-                host: 'localhost',
-                port: 7001,
-            },
-            {
-                host: 'localhost',
-                port: 7002,
+        options: {
+            startupNodes: [
+                {
+                    host: 'localhost',
+                    port: 7000,
+                },
+                {
+                    host: 'localhost',
+                    port: 7001,
+                },
+                {
+                    host: 'localhost',
+                    port: 7002,
+                }
+            ],
+            options: {
+                scaleReads: 'all',
+                clusterRetryStrategy: function (times) { return null },
+                redisOptions: {
+                    maxRetriesPerRequest: 1
+                }
             }
-        ],
+        }
     }
 }
 ```
-Just specify all the nodes in the cluster inside an array with their hosts  and ports.
+Just specify all the nodes in the cluster inside an array with their hosts and ports.
 
 You can use `typeorm cache:clear` to clear everything stored in the cache.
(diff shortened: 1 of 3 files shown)


src/cache/RedisQueryResultCache.ts
```diff
@@ -60,14 +60,12 @@ export class RedisQueryResultCache implements QueryResultCache {
             } else {
                 this.client = new this.redis();
             }
-
         } else if (this.clientType === "ioredis/cluster") {
-            if (cacheOptions && cacheOptions.options) {
-                this.client = new this.redis.Cluster(cacheOptions.options);
+            if (cacheOptions && cacheOptions.options && cacheOptions.options.startupNodes) {
```
vlapo: `options` argument is not required according to source code from `ioredis` so do not have to check `cacheOptions.options` existence.
https://github.com/luin/ioredis/blob/fbcf692f5de8c653139e4b053c6356ab6d67e463/lib/cluster/index.ts#L65

vlapo:
First at all thank you for contributing. We are appreciate it. I have one inline comment and please also add record to `CHANGELOG.md`.

@HussainAliAkbar, @ignacioblit if you have any comments.

src/cache/RedisQueryResultCache.ts
ignacioblit: Nice enhancement @scvgoe!

This looks good to me. The optional value is actually cacheOptions.options.options.  

cacheOptions.options is the typeORM options object with the general Cluster Ioredis properties including the mandatory one: startupNodes. So in the case of Cluster Redis its always required. 

I have to say that: cacheOptions.options.options looks pretty weird and maybe confusing, as it has plenty of options. Though I have no better suggestion than changing the name of the property to clusterOptions. Yet this is none important, so ignore this if you want :)

vlapo: Sry man :-) did not realize it because of so much "options".

ignacioblit: Hahaha, exactly!

vlapo: So just add recordt to `CHANGELOG.md` and it is ready to merge.

HussainAliAkbar:
This looks good to me :+1: Nice enhancement indeed!

src/cache/RedisQueryResultCache.ts
scvgoe: Thank you for your reviewing!!

scvgoe:
When will be released this pr?

scvgoe:
Graet idea!

docs/caching.md
```diff
@@ -158,10 +158,46 @@ In case you want to connect to a redis-cluster using IORedis's cluster functiona
...
+                clusterRetryStrategy: function (times) { return null },
+                redisOptions: {
+                    maxRetriesPerRequest: 1
+                }
+            }
+        }
+    }
+}
+```
+Just specify all the nodes in the cluster inside an array with their hosts and ports.
```
vlapo: This sentence after object config is really confusing. Also object based configuration should be first (as it provide more options) and array based config should be only as simple note with description.

pleerock:
Thank you!

Issue #3496, opened 2019-01-23
json and jsonb comparison does not work


grin0c:
**Issue type:**

[ ] question
[x] bug report
[ ] feature request
[ ] documentation issue

**Database system/driver:**

[ ] `cordova`
[ ] `mongodb`
[ ] `mssql`
[ ] `mysql` / `mariadb`
[ ] `oracle`
[x] `postgres`
[ ] `sqlite`
[ ] `sqljs`
[ ] `react-native`
[ ] `expo`

**TypeORM version:**

[ ] `latest`
[ ] `@next`
[ 0.2.8] `0.x.x` (or put your version here)

**Steps to reproduce or a small repository showing the problem:**
If we save such an object in the database:
```
[{"message": "", "attributeKey": "files", "level": "wrong"}]
```
then postgresql saves this object as:
```
[{"level": "wrong", "message": "", "attributeKey": "files"}]
```
The problem is here: https://github.com/typeorm/typeorm/blob/master/src/persistence/SubjectChangedColumnsComputer.ts#L80-L83

<!--
    To answer those questions you need to put "x" inside the square brackets, for example:
    [x] `mysql`
    [ ] `postgres`

    Also, please format your code properly (by taking code blocks into ```ts .... ```)
!>

pleerock:
please provide more details including minimal entity code for reproduction.

grin0c:
```
@Entity()
export class Post {
  @PrimaryGeneratedColumn()
  id: number

  @UpdateDateColumn({ type: 'timestamptz' })
  updatedAt: Date

  @CreateDateColumn({ type: 'timestamptz' })
  createdAt: Date

  @Column({ type: 'jsonb', nullable: true })
  problems: object
}

const connection = await createConnection(connectionOptions)
const repository = connection.getRepository(Post)

const p = await repository.save(repository.create({ problems: [{ attributeKey: 'files', level: 'wrong', message: '' }] }))

await repository.save(p)
```
The second save method is to run update.

pleerock:
> The problem is here: 

can you please explain why problem is in there? Json.stringify doesn't change properties order, right?

grin0c:
The order of keys changes postgres when saves json.
<img width="925" alt="2019-01-24 10 48 06" src="https://user-images.githubusercontent.com/7963898/51663647-5ca8d700-1fc8-11e9-8c98-405b30f57cf4.png">

When save requests a model from the database, the keys in json come in a different order and are obtained:
```
[...]

PR #3543, opened 2019-01-30
Fix jsonb comparison.

grin0c:
PR for #3496

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ feel free to ask us and community.
 
 ### Bug Fixes
 
+* fixed json/jsonb column data types comparison ([#3496](https://github.com/typeorm/typeorm/issues/3496))
 * fixed signatures of `update`/`insert` methods, some `find*` methods in repositories, entity managers, BaseEntity and QueryBuilders
 
 ### Features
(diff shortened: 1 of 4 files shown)


vlapo:
What about `json` data type? Is there same problem? Also `mysql` support `json` data type.

Also please add suitable record to `CHANGELOG.md` file.

grin0c:
@vlapo Only JSONB, I wrote a comment in the code. 
What should I write in the CHANGELOG.md?

vlapo:
Sure but from mysql docs:

> MySQL also discards extra whitespace between keys, values, or elements in the original JSON document. To make lookups more efficient, it also sorts the keys of a JSON object. You should be aware that the result of this ordering is subject to change and not guaranteed to be consistent across releases.

https://dev.mysql.com/doc/refman/8.0/en/json.html

I do not say your solution is bad. Just want have all covered by tests. So maybe add `json` column and enable `mysql` driver in your test would be nice.

> What should I write in the CHANGELOG.md?

Just simple description about fixing json/jsonb column data types comparison and link to issue :-)

grin0c:
I checked on mysql 5.7, doesn't sort. But in any case, the method of comparison is universal for json and jsonb

pleerock:
Thank you very much!

PR #3547, opened 2019-01-30
Insert rowsAffected in oracle database in delete function

RobinsonLuiz:
rowsAffected using the delete method with oracle database was returning undefined, these modifications return the number of rows affected in the database.

### DIFF
--- a/src/driver/oracle/OracleQueryRunner.ts
+++ b/src/driver/oracle/OracleQueryRunner.ts
@@ -172,7 +172,7 @@ export class OracleQueryRunner extends BaseQueryRunner implements QueryRunner {
                         this.driver.connection.logger.logQueryError(err, query, parameters, this);
                         return fail(new QueryFailedError(query, parameters, err));
                     }
-                    ok(result.rows || result.outBinds);
+                    ok(result.rows || result.outBinds || result.rowsAffected);
                 };
                 const executionOptions = {
                     autoCommit: this.isTransactionActive ? false : true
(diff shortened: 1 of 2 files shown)


RobinsonLuiz:
rowsAffected using the delete function with oracle database was returning undefined, these modifications return the number of affected rows in the database.

src/driver/oracle/OracleQueryRunner.ts
```diff
@@ -172,7 +172,7 @@ export class OracleQueryRunner extends BaseQueryRunner implements QueryRunner {
                         this.driver.connection.logger.logQueryError(err, query, parameters, this);
                         return fail(new QueryFailedError(query, parameters, err));
                     }
-                    ok(result.rows || result.outBinds);
+                    ok(result.rows || result.outBinds || result.rowsAffected);
```
pleerock: This approach doesn't look good. I suggest to add a third flag paramater to `query` method, something like `fullResult: boolean` (don't like this name however) and make a check here and return some `QueryResult` interface (need to create a new interface) with `rowsAffected` inside as well with all other additional data. Then you just call a method in `delete query runner` with fullResult = true.

Also, can you please implement same for other methods (update and insert)? Also would be really helpful if you can do it for all other drivers as well and we can officially support this feature and add documentation to it.

src/query-builder/DeleteQueryBuilder.ts
```diff
@@ -80,6 +80,9 @@ export class DeleteQueryBuilder<Entity> extends QueryBuilder<Entity> implements
                     deleteResult.affected = typeof result[1] === "number" ? result[1] : null;
                     break;
                 }
+                case "oracle":
+                    deleteResult.raw = result.rowsAffected;
```
pleerock: and also looks like you need to change `deleteResult.affected` not `deleteResult.raw`. Careful!

imnotjames:
@RobinsonLuiz are you looking to return to this to continue the work on it?  There's a few open comments that need resolved & conflicts between this change and the master branch.  This also needs tests to be good to bring in.

imnotjames (changes requested):
This needs tests to verify the behavior works as expected.

imnotjames:

PR #3557, opened 2019-02-02
Remove an unused import in the Data Mapper example

dguo:
My guess is that it was accidentally copied over from the Active Record example.


pleerock:
@vlapo please feel free to merge changes you think doesn't need everybody's attention instantly ;)

PR #3566, opened 2019-02-03
Add documentation for array parameter syntax

### LINKED ISSUE #2845: Passing an array as a bound parameter for a `WHERE IN` clause (sqlite)
**Issue type:**

[ ] question
[x] bug report
[ ] feature request
[ ] documentation issue

**Database system/driver:**

[ ] `cordova`
[ ] `mongodb`
[ ] `mssql`
[ ] `mysql` / `mariadb`
[ ] `oracle`
[ ] `postgres`
[x] `sqlite`
[ ] `sqljs`
[ ] `react-native`
[ ] `expo`

**TypeORM version:**

[x] `latest`
[ ] `@next`

It is currently not possible to pass an array as a bound variable to a QueryBuilder:

```ts
userRepository.createQueryBuilder('user')
  .where('user.id IN ?', [1,2,3])
  .getMany()
```

This will fail with an `SQLITE_ERROR: syntax error`, mainly because the sqlite3 doesn't allow passing dynamically sized arrays. Instead with plain sqlite it is possible to dynamically generate a string of the form `(?,?,?,...`) of appropriate length and then simply pass all bound `?`'s as an array.

However, typeorm forces you to use named aliases, so you would have to work around that by generating an array of aliases which is just a pain to do...

davewasmer:
Fixes #2845

### DIFF
--- a/docs/select-query-builder.md
+++ b/docs/select-query-builder.md
@@ -274,6 +274,19 @@ is a shortcut for:
 .setParameter("name", "Timber")
 ```
 
+You can also supply an array of values, and have them transformed into a list of values in the SQL
+statement, by using the special expansion syntax:
+
+```typescript
+.where("category IN (:...categories)", { categories: [ "one", "two" ] })
+```
+
+Which becomes:
+
+```sql
+WHERE category IN ('one', 'two')
+```
+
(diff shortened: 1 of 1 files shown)


pleerock:
Thank you! I changed names a bit to match context.

Issue #3554, opened 2019-02-01
Extend sqljs driver with localforage lib for asynchronous save and load operations


tstoppe89:
**Issue type:**

[ ] question
[ ] bug report
[x] feature request
[ ] documentation issue

**Database system/driver:**

[ ] `cordova`
[ ] `mongodb`
[ ] `mssql`
[ ] `mysql` / `mariadb`
[ ] `oracle`
[ ] `postgres`
[ ] `sqlite`
[x] `sqljs`
[ ] `react-native`
[ ] `expo`

**TypeORM version:**

[x] `latest`
[ ] `@next`
[ ] `0.x.x` (or put your version here)

**Description**

Hey guys,
we are currently running into performance problems with the sql.js library in the browser, because of it's synchron (and blocking) save and load operations of the whole database. In several ionic & angular apps the initialization takes 3 - 9 seconds and freezes the application in this time, because of the local storage usage. We also tested your sql.js driver solution by replacing the localstorage operations with localforage (https://github.com/localForage/localForage), which is using the non-blocking asynchronous indexedDB - but also as key-value store, like the local storage is doing. By replacing it, the app can initialize the database asynchronously in the background while starting (or displaying a loading screen) without freezing.

Our fork & the specific branch: https://github.com/ApinautenGmbH/typeorm/tree/driver-sqljs-localforage

I've already started working on this extension by adding another connection option to the `SqljsConnectionOptions` like:

```
    /**
     * Enables the usage of the localforage library to save & load the database asynchronously from the
     * indexedDB instead of using the synchron local storage methods in a browser environment.
     */
    readonly useLocalforage?: boolean;
```

in the SqljsDriver.ts i'm doing soemthing like this:
```
    if (this.options.useLocalforage) {
        if (window.localforage) {
            await window.localforage.setItem(path, JSON.stringify(databaseArray));
        } else {
            throw new Error(`localforage is not defined - please import localforage.js into your site`);
        }
    } else {
        PlatformTools.getGlobalVariable().localStorage.setItem(path, JSON.stringify(databaseArray));
    }
```

I really would like to write tests and searched browser tests for the sqljs - are there some browser tests in typeorm? Or how do you test the sqljs browser-version? 
I could also set up an ionic test project if this would help.

P.S.: I really like your library and your work here!
P.S.S: We also would like to implement an own driver for indexedDB (in the future), because we are thinking, that this would give those browser application another performance boost. We also saw the #2048 issue

[...]

PR #3572, opened 2019-02-04
Add localforage as alternative to local storage for sqljs driver

tstoppe89:
For issue #3554.

I've added an option `useLocalforage` to the SqljsConnectionOptions to switch between the usage of the synchronous local storage (which could freeze the browser, when having lot's of data) and the async. indexedDB for the sql.js driver.

With using this flag, the localforage node module is needed and the localforage.js needs to be added to the page.

Having it enabled, the save & load operations are now asynchron, which means, those operations are not blocking the main thread of the browser anymore, which gives you the opportunity the display smooth loading screens. 

I've also created a small pre-realese, which allows us to use this version of typeorm in current angular projects: https://github.com/ApinautenGmbH/typeorm/releases

### DIFF
--- a/src/driver/sqljs/SqljsConnectionOptions.ts
+++ b/src/driver/sqljs/SqljsConnectionOptions.ts
@@ -39,5 +39,5 @@ export interface SqljsConnectionOptions extends BaseConnectionOptions {
      * Enables the usage of the localforage library to save & load the database asynchronously from the
      * indexedDB instead of using the synchron local storage methods in a browser environment.
      */
-    readonly useLocalforage?: boolean;
+    readonly useLocalForage?: boolean;
 }
--- a/src/driver/sqljs/SqljsConnectionOptions.ts
+++ b/src/driver/sqljs/SqljsConnectionOptions.ts
@@ -34,4 +34,10 @@ export interface SqljsConnectionOptions extends BaseConnectionOptions {
      * location is used to automatically save the database.
      */
     readonly location?: string;
+
+    /**
+     * Enables the usage of the localforage library to save & load the database asynchronously from the
+     * indexedDB instead of using the synchron local storage methods in a browser environment.
+     */
+    readonly useLocalforage?: boolean;
 }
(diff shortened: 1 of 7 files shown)


tstoppe89:
closes  #3554

tstoppe89:
it's weird, locally the tests are running fine and even with improved promise handling, errors still occur in the travis build [https://i.ibb.co/K9QQJvJ/typeormtests.png](https://i.ibb.co/K9QQJvJ/typeormtests.png)

src/driver/sqljs/SqljsConnectionOptions.ts
```diff
@@ -34,4 +34,10 @@ export interface SqljsConnectionOptions extends BaseConnectionOptions {
      * location is used to automatically save the database.
      */
     readonly location?: string;
+
+    /**
+     * Enables the usage of the localforage library to save & load the database asynchronously from the
+     * indexedDB instead of using the synchron local storage methods in a browser environment.
+     */
+    readonly useLocalforage?: boolean;
```
pleerock: I think it should be called `useLocalForage` instead

src/driver/sqljs/SqljsDriver.ts
```diff
@@ -101,8 +102,17 @@ export class SqljsDriver extends AbstractSqliteDriver {
...
                 // browser
-                // fileNameOrLocalStorageOrData should be a local storage key
-                const localStorageContent = PlatformTools.getGlobalVariable().localStorage.getItem(fileNameOrLocalStorageOrData);
+                // fileNameOrLocalStorageOrData should be a local storage / indexedDB key
+                let localStorageContent = null;
+                if (this.options.useLocalforage) {
+                    if (window.localforage) {
+                        localStorageContent = await window.localforage.getItem(fileNameOrLocalStorageOrData);
+                    } else {
+                        throw new Error(`localforage is not defined - please import localforage.js into your site`);
```
pleerock: I guess it can be not only site

pleerock:
try to rewrite problematic assertion using different way

src/driver/sqljs/SqljsDriver.ts
tstoppe89: localforage is developed for the browser, it uses indexedDB (and WebSQL as fallback), which are only available in the browser

i've added this variant, because the usage of localforage is a lot easier for (web-)app developer than importing it as node module, because there is no special webpack or browserify configuration needed

pleerock:
Looks like we can merge it now. Thank you!

Issue #3182, opened 2018-11-30
Cannot execute increment on column defined by embedded entity


smonv:
**Issue type:**

[ ] question
[x] bug report
[ ] feature request
[ ] documentation issue

**Database system/driver:**

[ ] `cordova`
[ ] `mongodb`
[ ] `mssql`
[ ] `mysql` / `mariadb`
[ ] `oracle`
[x] `postgres`
[ ] `sqlite`
[ ] `sqljs`
[ ] `react-native`
[ ] `expo`

**TypeORM version:**

[ ] `latest`
[ ] `@next`
[x] `0.2.8` (or put your version here)

I have a embedded entity contain some friend statistics of a User. When user send a friend request to other user, I want increase `friend.sent` counter but the query always try to update `null` to `frientSent`.

My entity:
 
```ts
class FriendStats {
    @Column('bigint', { default: 0 })
    count: number;

    @Column('bigint', { default: 0 })
    sent: number;

    @Column('bigint', { default: 0 })
    received: number;
}

@Entity('users')
export class User {
    @Column(type => FriendStats)
    friend: FriendStats;
}
```

Query: 

```ts
await this.manager.increment(User, { id: user.id }, 'friend.sent', 1);
```

TypeORM generated SQL:

```
UPDATE "users" SET "friendSent" = $2, "updatedAt" = CURRENT_TIMESTAMP WHERE "id" = $1 -- PARAMETERS: ["123456",null]
```

Error

```
query failed: UPDATE "users" SET "friendSent" = $2, "updatedAt" = CURRENT_TIMESTAMP WHERE "id" = $1 -- PARAMETERS: ["123456",null]

error: { error: null value in column "friendSent" violates not-null constraint ......
```

apellizzn:
This seems a bit strange, considering that the following is the output of the increment function in the `EntityManager.ts`
```
await this
            .createQueryBuilder(entityClass, "entity")
            .update(entityClass)
            .set({
                [propertyPath]: () => this.connection.driver.escape(column.databaseName) + " + " + Number(value)
            })
            .where(conditions)
            .execute();
```
There should be at least a `+` in your query output!

smonv:
I'm tried with `createQueryBuilder` and `update` but got same error and same generated query.

```ts
await manager.createQueryBuilder(User, 'u')
                .update()
                .set({
                    'friend.sent': () => `"friendSent" + ${Number(1)}`,
                })
                .where({ id: user.id }).execute();
```

Query:

[...]

PR #3578, opened 2019-02-04
fix(3182): fixed increment/decrement values of embeded entity

vlapo:
**Root of problem:**
Current implementation of increment/decrement method for embedded property define values to update like this:
```typescript
// ...
.set({
    "embedded.property": () => ...
})
// ...
```
But update query builder do not support define key as string with "dot notation path".

**Solution:**
Convert:
```typescript
.set({
    "embedded.property": () => ...
})
```
to: 
```typescript
.set({
    embedded: {
         property: () => ...
    }
})
```

*Note: Sure we can add support of "dot notation path" into `set()` method but right now `set()` method accept strongly typed argument and I think it is better keep this functionality and just update increment/decrement function as it is presented in this PR.*

Fixes https://github.com/typeorm/typeorm/issues/3182

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ feel free to ask us and community.
 
 ### Bug Fixes
 
+* fixed increment/decrement value of embedded entity ([#3182](https://github.com/typeorm/typeorm/issues/3182))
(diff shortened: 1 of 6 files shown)


pleerock:
Thank you!

Issue #3395, opened 2019-01-09
Transform.from does nothing when column is NULL


rrgoncalves:
**Issue type:**

[ ] question
[x] bug report
[ ] feature request
[ ] documentation issue

**Database system/driver:**

[ ] `cordova`
[ ] `mongodb`
[x] `mssql`
[x] `mysql` / `mariadb`
[x] `oracle`
[x] `postgres`
[ ] `sqlite`
[ ] `sqljs`
[ ] `react-native`
[ ] `expo`

**TypeORM version:**

[ ] `latest`
[ ] `@next`
[x] `0.2.11` (or put your version here)

**Steps to reproduce or a small repository showing the problem:**

Have a NULL value in a column like user.name and try to transform it with the following code:

```js
// user.ts

@Entity("usuario", { schema: "myschema" })
export class user { 
      @Column({
        name: "name",
        type: 'text',
        transformer: {
          to(n) { 
            return n === undefined ? null : n 
          },
          from(n) { 
            return n === null ? undefined : n
          },
        }
    })
}
```

So, I was checking the code from PostgresDriver.js (`/driver/postgres/PostgresDriver.js`) and found out that there's no check if there's a transformer function set in the `@Column` decorator, and the code always checks first if the value if either null or undefined and returns it. As shown below:

```js
PostgresDriver.prototype.prepareHydratedValue = function (value, columnMetadata) {
        if (columnMetadata.type === Boolean) {
            value = value ? true : false;

        // [...]
}
```

The fix should be something as simple as checking if the transformer exists and if it does, execute it:

```js
PostgresDriver.prototype.prepareHydratedValue = function (value, columnMetadata) {
        if (columnMetadata.transformer) 
           return columnMetadata.transformer.from(value);

        if (columnMetadata.type === Boolean) {
            value = value ? true : false;

        // [...]
}
```

Some other people (#https://github.com/typeorm/typeorm/pull/2777) and I believe the transformer should be executed nonetheless.

I'm willing to send a PR and add tests to fix such issue.

PS: MongoDriver got it right: 

```js
MongoDriver.prototype.prepareHydratedValue = function (value, columnMetadata) {
  if (columnMetadata.transformer)
    value = columnMetadata.transformer.from(value);
  return value;
};
```

Thanks!

PR #3581, opened 2019-02-04
fix(3395): handle transformer.from method in case column is NULL

vlapo:
Fixes https://github.com/typeorm/typeorm/issues/3395

Also we can close https://github.com/typeorm/typeorm/pull/2777 in favor of this PR.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ feel free to ask us and community.
 
 ### Bug Fixes
 
+* fixed missing call `transformer.from()` in case column is NULL ([#3395](https://github.com/typeorm/typeorm/issues/3395))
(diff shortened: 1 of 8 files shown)


CHANGELOG.md
```diff
@@ -9,6 +9,7 @@ feel free to ask us and community.
 
 ### Bug Fixes
 
+* fixed missing call `transformer.from()` in case column is NULL ([#3395](https://github.com/typeorm/typeorm/issues/3395))
```
pleerock: this is a bit breaking since users will have `null` now instead of value and their transform methods can give them `cannot read property x of null` now. I'm okay to merge it to master, I just aware you about it.

pleerock:
Thank you!

Issue #3587, opened 2019-02-06
Schema log regenerates enum every time (even if no changes)


onionhammer:
**Issue type:**

[ ] question
[x] bug report
[ ] feature request
[ ] documentation issue

**Database system/driver:**

[ ] `cordova`
[ ] `mongodb`
[ ] `mssql`
[ ] `mysql` / `mariadb`
[ ] `oracle`
[x] `postgres`
[ ] `sqlite`
[ ] `sqljs`
[ ] `react-native`
[ ] `expo`

**TypeORM version:**

[x] `latest`
[ ] `@next`
[ ] `0.x.x` (or put your version here)

**Steps to reproduce or a small repository showing the problem:**

I've created several enum columns, every time I create a migration or run schema:log typeorm calls `CREATE TYPE` for them, even if there have been no changes.

```ts
@Entity("equipmentmodels")
@Unique(["name"])
export class EquipmentModel {
  /* snipped other properties for simplicity */
  
  @PrimaryGeneratedColumn("uuid")
  id: string

  /** Type of equipment */
  @Column("enum", { enum: EquipmentModelType })
  type: EquipmentModelType

}
export enum EquipmentModelType {
  Thing1,
  Thing2,
  Thing3,
  Thing4
}
```

Consistently yields this

```
DROP TYPE "pcc"."devicemodels_type_enum_old";
ALTER TYPE "pcc"."equipmentmodels_type_enum" RENAME TO "equipmentmodels_type_enum_old";
CREATE TYPE "pcc"."equipmentmodels_type_enum" AS ENUM('0', '1', '2', '3');
```

PR #3594, opened 2019-02-07
fix(postgres): Fixed migration issue with postgres numeric enum type

vlapo:
Postgres store enum type as string and we support now numeric TS enum so we have to compare array of possible values and values in DB schema as strings.

Fixes https://github.com/typeorm/typeorm/issues/3587

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ feel free to ask us and community.
 
 ### Bug fixes
 
+* fixed migration issue with postgres numeric enum type - change queries are not generated if enum is not modified ([#3587](https://github.com/typeorm/typeorm/issues/3587))
 * fixed mongodb entity listeners in optional embeddeds ([#3450](https://github.com/typeorm/typeorm/issues/3450))
 * fixes returning invalid delete result
 
--- a/src/driver/postgres/PostgresDriver.ts
+++ b/src/driver/postgres/PostgresDriver.ts
@@ -741,7 +741,7 @@ export class PostgresDriver implements Driver {
             if (!tableColumn)
                 return false; // we don't need new columns, we only need exist and changed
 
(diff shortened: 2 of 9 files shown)


src/driver/postgres/PostgresDriver.ts
```diff
@@ -751,7 +751,15 @@ export class PostgresDriver implements Driver {
                 || tableColumn.isPrimary !== columnMetadata.isPrimary
                 || tableColumn.isNullable !== columnMetadata.isNullable
                 || tableColumn.isUnique !== this.normalizeIsUnique(columnMetadata)
-                || (tableColumn.enum && columnMetadata.enum && !OrmUtils.isArraysEqual(tableColumn.enum, columnMetadata.enum))
+                || (
+                    tableColumn.enum
+                    && columnMetadata.enum
+                    // enums in postgres are always strings
```
pleerock: couldn't we just store enum values as strings in both sides? (e.g. tableColumn.enum and columnMetadata.enum both to be a `string[]` ?)

vlapo: You are half right. `tableColumn.enum` is always `string[]` but for `columnMetadata.enum` we need type `(string|number)[]` because we support numeric and heterogeneous based typescript enums. I updated this PR with typings. Please check it now :)

CHANGELOG.md
```diff
@@ -9,6 +9,7 @@ feel free to ask us and community.
 
 ### Bug Fixes
 
+* fixed migration issue with postgres numeric enum type - change queries are not generated if enum is not modified ([#3587](https://github.com/typeorm/typeorm/issues/3587))
```
pleerock: now outdated changelog

vlapo:
Sry for delay. Rebased and updated.

pleerock:
Thank you!!

Issue #3583, opened 2019-02-05
Add `browser: browser/index.js`, `main: index.js` to package.json


danielmhanover:
This would allow re-use of the same source files across both server and client environments in Webpack, as Webpack will intelligently choose the right TypeORM build based on the dev environment.

vlapo:
Does it mean that this webpack config will be no more necessary ? 

http://typeorm.io/#/supported-platforms/browser

danielmhanover:
that is correct

chriswep:
@danielmhanover I was looking for the same thing. Since this is a small change, i created a pull request: https://github.com/typeorm/typeorm/pull/3623

I additionally added dummy classes to replace backend-only drivers in browser builds via the same configuration option in `package.json`. Using my fork of typeorm i could successfully import typeorm in webpack builds with zero additional configuration and bundle size reduced by one third (@vlapo)

@pleerock what do you think about the approach with the dummy classes? I think it's safe to assume that people don't use mysql/oracle in a browser/mobile environment? It would bring a significant advantage to browser builds by default. However i can understand if you don't like this approach. From the few places i looked into the code, i feel there would be some refactoring needed to have a more "solid" option to disable drivers. Do you see any other areas where we can save on bundle size in browsers? I would be interested to look into this.

pleerock:
I would really like to see some better approach to prevent having multiple instances of the same classes, this is going to be annoying in IDE because it will suggest you multiple instances of classes

chriswep:
ya, the IDE issue also came to my mind. i don't have enough insight in the code to determine if there is another simple way to achieve a similar result (hints?). from my perspective (and use case) the benefit would outweigh the IDE issue - but for sure there are other valid perspectives on this. Would it help if i add jsdoc comments to the classes like "DO NOT USE THIS CLASS"? :-) 

i think the same effect can be achieved via webpack configuration, however i think a smaller default bundle size would make using typeorm in the browser less painful. don't know how many people use it in this environment though.

alfaproject:
People that develop in serverless environments also want small bundles to save costs on storage and execution time as well. That's a back-end environment but would also benefit from this kind of optimizations.

chriswep:
since this is fixed in https://github.com/typeorm/typeorm/pull/3623, i think this issue can be closed @pleerock  @danielmhanover

vlapo:
Close as merged in https://github.com/typeorm/typeorm/pull/3623

PR #3623, opened 2019-02-12
added browser entry point to package.json and excluded backend-only drivers

chriswep:
There is a configuration option for `package.json` to instruct bundlers like `webpack` to use a different entry point in browser environments without any further configuration needed:  https://github.com/defunctzombie/package-browser-field-spec

Using this, we can additionally replace backend-only drivers by dummy classes to reduce the bundle size. Disabling Mongo, Postgres, SqlServer, Mysql and Oracle reduces the bundle size by one third.

With this change, `typeorm` can be imported in browser environments with zero additional configuration.


vlapo:
Fixes https://github.com/typeorm/typeorm/issues/3583

pleerock:
what about query runners? Are they excluded this way?

chriswep:
> what about query runners? Are they excluded this way?

i checked with the weback BundleAnalyzer: somehow the mongodb query runner survived, however the other excluded drivers are completely gone.

@pleerock as mentioned in #3583 - concerning the duplicate classes and IDE suggestions: would it be a compromise to add jsdoc warnings to the classes? that should prevent most cases of importing the wrong class. maybe there is also a possibility to ignore the file via `editorconfig`?

Anyway, if want to have driver-replacements out of this pull request i can remove them and just leave the new browser entry point in `package.json`.

chriswep:
> > what about query runners? Are they excluded this way?
> 
> i checked with the weback BundleAnalyzer: somehow the mongodb query runner survived, however the other excluded drivers are completely gone.

seems like MongoDb has a special spot in the codebase (for historic reasons?) - apart from the mongodb query runner also the mongodb entity manger and repository are imported from outside the driver (this doesn't happen for the other drivers). Anyway, those files can be replaced by the same method via `package.json` if we wanted to (bringing down the bundle size a bit more).

pleerock:
> would it be a compromise to add jsdoc warnings to the classes? that should prevent most cases of importing the wrong class. maybe there is also a possibility to ignore the file via `editorconfig`?

I don't think editorconfig has such option. jsdoc also wouldn't help. The only option I know is to have this file in format other than `.ts`, and create a separate build configuration that will inject this file only in the build process (we already do it this way, see BrowserPlatformTools file).

Yes, we need to completely strip out mongo as well.

chriswep:
> The only option I know is to have this file in format other than `.ts`, and create a separate build configuration that will inject this file only in the build process (we already do it this way, see BrowserPlatformTools file).

done
 
> Yes, we need to completely strip out mongo as well.

done

@pleerock any other concerns?

pleerock:
Hope you won't have me for the solution I suggested. Hope we can came up to something better in the future.

Thank you very much for your contribution!

timoschwarzer:
There's one problem with this. When using Electron in combination with Vue CLI (which uses Webpack internally), TypeORM stops working with this patch. Version 0.2.13 worked just fine but versions since 0.2.14 don't work anymore. What would be the best way to fix this? :smiley:

Kononnable:
A lot of libs(electron, vue, webpack), so it might be hard to find a problem. It probably doesn't work out with some of webpack configuration, but with which option? What error does it throw? We should probably create new issue for this - it would be hard to follow something in already merged PR.

chriswep:
@timoschwarzer you can tell webpack to use another entry point (i.e. main, not browser, which is the default)

timoschwarzer:
@Kononnable you're right. For more details see my newly created issue. (tl;dr: target=node does not work with Electron)

timoschwarzer:
@chriswep Do you mind telling me how to do that? :)

PR #3625, opened 2019-02-12
Fix incorrect pagination argument in documentation

nubs:
The documentation for `findAndCount` referenced the pagination settings as `from` and `take`, but they should be `skip` and `take` instead.


pleerock:
Thank you!

Issue #3644, opened 2019-02-14
FindOneOptions["where"] type


tlaziuk:
**Issue type:**

- [x] question
- [x] bug report
- [ ] feature request
- [x] documentation issue

**Database system/driver:**

- [ ] `cordova`
- [ ] `mongodb`
- [ ] `mssql`
- [ ] `mysql` / `mariadb`
- [ ] `oracle`
- [ ] `postgres`
- [ ] `sqlite`
- [ ] `sqljs`
- [ ] `react-native`
- [ ] `expo`

**TypeORM version:**

- [ ] `latest`
- [ ] `@next`
- [x] `0.2.13` (and previous versions)

**Steps to reproduce or a small repository showing the problem:**

according to doc (https://github.com/typeorm/typeorm/blob/0.2.13/docs/find-options.md) results can be filtered by the `where` find option and alternative conditions (`OR`) can be separeted by passing an array of conditions to that option,

yet the type doesn't reflect that, is there any particular reason why the type `Array<FindConditions<Entity>>` isn't there? https://github.com/typeorm/typeorm/blob/0.2.13/src/find-options/FindOneOptions.ts#L18

Diluka:
> yet the type doesn't reflect that, is there any particular reason why the type `Array<FindConditions<Entity>>` isn't there? https://github.com/typeorm/typeorm/blob/0.2.13/src/find-options/FindOneOptions.ts#L18

https://github.com/typeorm/typeorm/blob/4448f4a5553454c391aaf448fbed1f68f2b479e8/docs/find-options.md#L49-L58

PR #3646, opened 2019-02-15
update FindOneOptions["where"] type

tlaziuk:
according to docs (https://github.com/typeorm/typeorm/blob/0.2.13/docs/find-options.md) this type should also accept an array of find options

array of options was previously covered by `ObjectLiteral` type, should it be removed then?

#3644

### DIFF
--- a/src/find-options/FindOneOptions.ts
+++ b/src/find-options/FindOneOptions.ts
@@ -15,7 +15,7 @@ export interface FindOneOptions<Entity = any> {
     /**
      * Simple condition that should be applied to match entities.
      */
-    where?: Array<FindConditions<Entity>>|FindConditions<Entity>|ObjectLiteral|string;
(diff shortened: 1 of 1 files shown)


pleerock:
hmm I don't remember if this functionality works in 0.2.x? Does it? I know for sure it works in `next`. If we have this functionality in 0.2.x then why don't we have tests for it?

tlaziuk:
@pleerock seems like it works, I'm not very familiar to this project internals, but I can try to write the tests

pleerock:
I added test by my own, thank you. Yes, looks like we can to merge this PR.

> array of options was previously covered by `ObjectLiteral` type, should it be removed then?

no, ObjectLiteral is also used for different purposes... `find*` methods signature has changed in `next` and ObjectLiteral was already removed in there.

Issue #2358, opened 2018-06-19
Change ormconfig behaviour and connection creation design in next TypeORM version


pleerock:
In this issue I want to describe few features I would like to introduce in 0.3.0 

I consider following changes:

1. **remove connection manager and don't manage connections at all. Users will be able to define and manage multiple connections following way:**

```ts
export const defaultConnection = createConnection({ ... });
export const anotherConnection = createConnection({ ... });
export const thirdConnection = createConnection({ ... });
```

2. **remove `getManager`, `getConnection`, `getRepository` and other exported methods, we don't need them, we will be able to simply do following:**

```ts
export const connection = createConnection({ ... }); // same as getConnection()
export const manager = connection.manager; // same as getManager()
export const getRepository<T> = (entity: T) => manager.getRepository(entity); // same as getRepository()
```

This approach fits more JavaScript world then current "connection manager" based approach. 

3. **remove automatic ormconfig loading** 

Not yet sure about this approach yet since CLI needs it, just like it needs `entities` sections of it, but still we need some solution. The best looking solution is to have ormconfig.ts (which compiles to ormconfig.js)

4. **remove support for directory paths in entities, migrations and subscribers options in ormconfig**

There are various issues user face with loading entities by directory path, e.g.

```ts
entites: ["entity/*.js"]
```

Most of issues are related to different compiler options and execution environments. Few examples of problems:

1. If you are using ts-node or similar approach of running typescript code instead of javascript you'll need to use following configuration: `entites: ["entity/*.ts"]`
2. If you are using ts-node for development, but node for production you'll need dynamic configuration or following configuration `entites: ["entity/*{.ts,.js}"]` but this may not work in rare specific cases
3. If you are using outDir of typescript you need to use following configuration: `entites: ["./out-dir/entity/*.js"]`. If you mix outDir, ts-node for development and node for production the situation gets ugly
4. If you remove or rename entity with outDir option enabled users get serious not obvious issues on why they have a wrongly synchronized database schema (they need to manually cleanup outDir directory to resolve this issue)

5. **rename Connection to DataSource and QueryRunner to Connection**

This is a huge change. Currently Connection is a wrong word since it does not establish a real database connection, its just a database source. And QueryRunner is actually is a connection, so I think they should be renamed.

posgarou:
These all sounds like reasonable changes.

If you decide to remove them, it would be helpful to deprecate them for the 0.3.x releases and only take them out in the 0.4.0 release.

[...]

PR #3651, opened 2019-02-16
Add support to load other .env variations

RodolfoSilva:
Add support to load other .env variations.

https://github.com/bkeepers/dotenv#what-other-env-files-can-i-use


RodolfoSilva:
Closed in favor of the #2358. The task expose a better api to use the configurations.

PR #3653, opened 2019-02-16
Fix invalid delete result

Hossein-s:
Currently delete queries return wrongly initialized DeleteResult object. The logic uses connection name to check type of driver that is not correct for production environment.


pleerock:
Thank you very much!

PR #3655, opened 2019-02-17
Fix the quirk bug that the buffer compare function is not working.

sevenryze:
The buffer compare function shows a quirk bug, see the picture i captured below:

annotation 2019-02-17 123514

The function instanceof Function check failed and we need to add more checks.

BWT. How many frequency we publish a new version? which including bug fix version. 
I think the predictable frequency is very help for more contributing and more robust for formal application.

### DIFF
--- a/src/util/OrmUtils.ts
+++ b/src/util/OrmUtils.ts
@@ -177,7 +177,9 @@ export class OrmUtils {
         if (x === y)
             return true;
 
-        if (x.equals instanceof Function && x.equals(y))
+        // Fix the buffer compare bug.
+        // See: https://github.com/typeorm/typeorm/issues/3654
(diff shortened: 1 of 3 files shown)


sevenryze:
Any update on this PR? @pleerock Please review 😄

pleerock:
Thank you!

Strange that instance of Function doesn't work...

pleerock:
> BWT. How many frequency we publish a new version? which including bug fix version.
I think the predictable frequency is very help for more contributing and more robust for formal application.

I would like to avoid fixed schedule on version publishing to be flexible. I usually release a new version if:

* I feel there is enough changes to make a new release
* there is some critical/important change
* somebody ask to release

sevenryze:
Fine, got your points and thanks for your review.

Will this fix be included in next release?

pleerock:
Just released it.

PR #3656, opened 2019-02-17
Hotfix: entity cloning with new non-emurable properties

joelgallant:
I unfortunately don't have the testing setup on my machine at the moment, but making this fix in node_modules of the failing test (https://github.com/P4sca1/typeorm-problem) works as intended.


joelgallant:
This bug looks like it's caused by https://github.com/typeorm/typeorm/blob/master/src/persistence/Subject.ts#L147 which was introduced here 018ee51399ecdd2e2810a39372a179819510b230

joelgallant:
@pleerock I'll need your eyes on this, since I don't really know what 018ee513 is doing and why it was necessary. It seems that this is outside of the current test coverage since tests pass fine.

src/persistence/SubjectDatabaseEntityLoader.ts
```diff
@@ -64,7 +64,7 @@ export class SubjectDatabaseEntityLoader {
 
                     // gets all relation property paths that exist in the persisted entity.
                     subject.metadata.relations.forEach(relation => {
-                        const value = relation.getEntityValue(subject.entityWithFulfilledIds!);
+                        const value = relation.getEntityValue(subject.entity!);
```
pleerock: uhh wait wait those places are complex and thus sensitive to change, can you explain your change? Its scary to me to merge changes in this area.

pleerock:
you are talking about one year old commit... Tests are passing in the master, so it looks like your change broke current test you are referring

joelgallant:
Yeah. I don't have the bandwidth to understand how all these things interact, it's really complex. Tests pass with 9fad2ab, but I tried to distill the fix down more.

pleerock:
I had to revert your commit to release a new version with a fix.

Please create a new PR with those changes, add a tests that checks reported issue and try to fix issue with the following suggestion:

I checked source code, we needed Object.assign here because otherwise it would change original entity. Try to replace Object.assign({}, ...) with Object.assign(this.metadata.create(), ...)

joelgallant:
Appreciate the tip, thanks. Would be good to get test coverage if it's not there already for that failing case, and get a robust patch in.

imnotjames:
Is this still relevant?

joelgallant:
I don't think so

imnotjames:
Alrighty!  Thanks for the contribution anyway. :)

PR #3674, opened 2019-02-19
Update README: missing property in example model

neminovno:
Property `isPublished` is missing in starting example model.


pleerock:
Thank you!

PR #3675, opened 2019-02-20
pass column name for `SQL expressions` value in update query

s97712:


vlapo:
Please describe more about this PR. What is problem you want to resolve with this PR.

s97712:
this PR can pass column name to function of generating `SQL expressions`  in update query.
```
        const qb = connection.createQueryBuilder();
        await qb
            .update(User)
            .set({ likesCount: () => qb.escape(`likesCount`) + " + 1" }) 
            /* Replace with 
            .set({ likesCount: name => `${name} + 1` })
            */
            .where("likesCount = 1")
            .execute();
```

vlapo:
okey, but first at all fix all issues with failed build.

pleerock:
I don't think we need it. In practice people use a single database and they don't need cross-database escaping mechanizm. So,  `.set({ likesCount: () => `"likesCount" + 1` })` is already quite fine.

Kononnable:
I believe if you specify db name you also have to specify schema name(at least in some database engines).

Tests failing, author didn't reply. I also think this is very marginal use case(and it's not even complete).
Closing.

PR #3680, opened 2019-02-20
add codecov badge

championswimmer:


pleerock:
I guess some people wouldn't like that its not 100% 😆

PR #3683, opened 2019-02-21
fix a bug that loadEagerRelations not working

superhuman54:
```
someRepository.findOne("anyId", {
 loadEagerRelations: false
});
```

in case of the above code, FindOptionsUtils.isFindOneOptions returns false and we need to check one more option(typeof loadEagerRelations).

please check this out. :D


pleerock:
Thank you! Would be nice to have tests thought

PR #3686, opened 2019-02-21
Make subject entity’s properties enumerable

### LINKED ISSUE #3576: Error when saving an entity after lazy loading its relations: TypeError: Cannot read property 'find' of undefined
**Issue type:**

[ ] question
[x] bug report
[ ] feature request
[ ] documentation issue

**Database system/driver:**

[ ] `cordova`
[ ] `mongodb`
[ ] `mssql`
[x] `mysql` / `mariadb`
[ ] `oracle`
[ ] `postgres`
[ ] `sqlite`
[ ] `sqljs`
[ ] `react-native`
[ ] `expo`

**TypeORM version:**

[x] `latest`
[ ] `@next`

**Repository demonstrating the problem:**

https://github.com/P4sca1/typeorm-problem

vinsidious:
Overview
-----
This PR fixes #3576 by making all configurable properties of `Subject.entity` enumerable again. This maintains the original intent of #3231 while fixing the underlying issue that it introduced.

**Note:** Unfortunately, I didn't noticed #3656 until just before opening this PR. That being said, I think that these changes are still worth proposing since they take a different approach and include tests.

-----

@pleerock Please let me know if the impact of these changes are larger than I'm anticipating. As far as I can tell, this makes the properties on `Subject.entity` the same as they were prior to #3231.

### DIFF
--- a/src/util/OrmUtils.ts
+++ b/src/util/OrmUtils.ts
@@ -63,6 +63,24 @@ export class OrmUtils {
     }
 
     /**
+     * Iterates over an object's own property names and ensures that all
+     * configurable properties are also enumerable.
+     */
+    static makeEnumerable(object: any) {
+        if (this.isObject(object)) {
+            Object.getOwnPropertyNames(object).forEach(property => {
+                const descriptor = Object.getOwnPropertyDescriptor(object, property);
+                if (descriptor && descriptor.configurable && !descriptor.enumerable) {
+                    descriptor.enumerable = true;
+                    Object.defineProperty(object, property, descriptor);
+                }
+            });
+        }
+
+        return object;
+    }
+
+    /**
      * Deep Object.assign.
      *
      * @see http://stackoverflow.com/a/34749873
--- a/test/github-issues/3576/entity/Book.ts
+++ b/test/github-issues/3576/entity/Book.ts
@@ -1,0 +1,14 @@
+import { Column, Entity, PrimaryGeneratedColumn, ManyToOne } from "../../../../src";
+import { User } from "./User";
+
+@Entity()
+export class Book {
+    @PrimaryGeneratedColumn("uuid")
+    id: string;
+
+    @Column()
+    title: string;
+
+    @ManyToOne(type => User, user => user.books, { lazy: true })
+    author: User;
+}
--- a/test/github-issues/3576/entity/User.ts
+++ b/test/github-issues/3576/entity/User.ts
@@ -1,0 +1,14 @@
+import { Column, Entity, OneToMany, PrimaryGeneratedColumn } from "../../../../src";
+import { Book } from "./Book";
+
+@Entity()
+export class User {
+    @PrimaryGeneratedColumn("uuid")
+    id: string;
+
+    @Column()
+    name: string;
+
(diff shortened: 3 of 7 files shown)


vinsidious:
Hmm actually, it looks like there are still cases where the error occurs. Not only am I having a tough time reproducing via a test case, I'm concerned that the changes added by #3231 are affecting more areas than originally anticipated and that the benefits in this case don't really justify the risk.

@pleerock any chance we can revert #3231?

test/github-issues/3576/entity/User.ts
```diff
@@ -0,0 +1,14 @@
...
+@Entity()
+export class User {
+    @PrimaryGeneratedColumn("uuid")
+    id: string;
+
+    @Column()
+    name: string;
+
+    @OneToMany(type => Book, book => book.author, { lazy: true })
+    books: Book[];
```
pleerock: for lazy relations its type should be Promise<Book[]>

test/github-issues/3576/entity/Book.ts
```diff
@@ -0,0 +1,14 @@
...
+@Entity()
+export class Book {
+    @PrimaryGeneratedColumn("uuid")
+    id: string;
+
+    @Column()
+    title: string;
+
+    @ManyToOne(type => User, user => user.books, { lazy: true })
+    author: User;
```
pleerock: for lazy relations its type should be Promise<User>

test/github-issues/3576/issue-3576.ts
```diff
@@ -0,0 +1,48 @@
...
+        await booksRepo.save(book1);
+
+        const testAuthor = await usersRepo.findOneOrFail();
+
+        // Historically, this was the step that would cause the subsequent save
+        // operation to throw.
+        await testAuthor.books;
+
+        await usersRepo.save(testAuthor);
+
```
pleerock: our goal is to check if its not enumerable. Can you add test at the end and confirm using JSON.stringify that after `save` call properties aren't enumerable again?

src/util/OrmUtils.ts
```diff
@@ -62,6 +62,24 @@ export class OrmUtils {
...
+     * Iterates over an object's own property names and ensures that all
+     * configurable properties are also enumerable.
+     */
+    static makeEnumerable(object: any) {
+        if (this.isObject(object)) {
+            Object.getOwnPropertyNames(object).forEach(property => {
+                const descriptor = Object.getOwnPropertyDescriptor(object, property);
+                if (descriptor && descriptor.configurable && !descriptor.enumerable) {
+                    descriptor.enumerable = true;
+                    Object.defineProperty(object, property, descriptor);
```
pleerock: are you sure its not changing original object? (for me it looks like it does)

pleerock:
@vinsidious btw do you need entity’s properties to be not enumerable just like just like original author of #3231 needs? I thinking about pros and cons of them being not enumerable

vinsidious:
@pleerock I never minded having the properties enumerable which is why I'd _personally_ prefer if we reverted #3231 so that I can update from `0.2.11` to the latest version.

pleerock:
@vinsidious I reverted it in `0.2.14`

PR #3697, opened 2019-02-23
Moves directory search to it's own function

gintsgints:
This small change allows to use TypeORM internal procedure of searching entity directories, outside ORM, by moving part of code to it's own exported function.


vlapo:
Please, run tests locally before submit PR. I think with docker it is so simple. https://github.com/typeorm/typeorm/blob/master/DEVELOPER.md#running-tests-locally

pleerock:
Also can you please explain motivation of those changes once you fix tests?

gintsgints:
I close this while I get time to do it properly.

Issue #1414, opened 2018-01-06
Support enum for DB who don't support it


antoine-pous:
Currently SQLite doesn't have `ENUM` type, but TypeORM can use it's own validator before inserting a value into the DB and emit an error when the value is not in the enum.

```typescript
enum MyEnum { 
    zero = 0; 
    one = 1;
}

@Entity()
export class Test {
    @Column({enum: MyEnum})
    myVar: number; // Must be 0 or 1
}
```

mhombach:
I'm a big fan of making ORM-Systems really universal or at least as much universal as it can get.
If i create my business-logic within it and i later need to change the database-engine, then there should not be a bunch of stuff i need to change. Sure, i can then optimize everything for the new engine, but it should "work" flawless without changing much stuff...
So i would vote for implementing this as universal type into the ORM and when a database doenst support that column-type, then internally it should just get converted to a string or an integer, which represents that ENUM values. And sure, then it should be validated exactly to match the declared ENUM-Object.

pleerock:
@mhombach sorry to disappoint you, but Im biggest anti-fun of universal ORM-systems, simply because:

* any abstraction come with cost, and cost of making everything universal is too high - you loose to much functionality that each database provides which values higher then having things "universal", because of...
* being universal is mostly useless, because in practice it does not give anything except of "i later need to change the database-engine". But reality that you are not going to change your database engine almost never. And if you'll have such need ever it will be easier to you to change all places in your code to match new database rather then loosing benefits provided by point 1

But this particular feature request we can implement, because it does not restrict / limit us in anything, its just an addition.

atheros:
I once wrote a full app using an orm(not typeorm) using linux + mysql for in-house usage, then my boss came and told me I have to deploy same app to one of our clients. The catch was it was supposed to run on windows with mssql. Obviously it took me some time to get the job done and I would have greatly appreciated if the orm I was using would allow me to abstract things more.

In any case since there is already simple-array and simple-json, why not add simple-enum? Also an option with "useNativeTypeIfAvailable" would be even better.

pleerock:
This issue is opened for the community. Feel free to implement `simple-enum` type, its easy feature to implement.

davidmpaz:
My 5 cents...

@antoine-pous  I came into this issue a couple of day before looking for the same kind of functionality. The outcome of it did not look well so i tried to solve it working around. I am working with `cordova` and `sqljs` drivers. So, what I did was this:
```typescript
[...]

PR #3700, opened 2019-02-23
Added simple-enum column type

borremosch:
I have added the `simple-enum` column type discussed in #1414. It works like this:

- For MySQL, MariaDB and Postgres, the native `ENUM` type is used
- For SQLite and SQL Server, a `CHECK()` constraint is used, like so:
  ```
    colname CHECK (colname in ('A','B','C')) NOT NULL DEFAULT ('A')
  ```
  The CHECK is also reverse engineered back into an enum when code is being generated from a DB schema
- For Oracle, I found that enum is currently not supported in TypeORM. Since Oracle does have a native `ENUM` type, I believe `simple-enum` should be based on it. I have not added `simple-enum` for Oracle.

### DIFF
--- a/src/driver/sqlite-abstract/AbstractSqliteDriver.ts
+++ b/src/driver/sqlite-abstract/AbstractSqliteDriver.ts
@@ -246,6 +246,8 @@ export abstract class AbstractSqliteDriver implements Driver {
 
         } else if (columnMetadata.type === "simple-json") {
             return DateUtils.simpleJsonToString(value);
+        } else if (columnMetadata.type === "simple-enum") {
+            return DateUtils.simpleEnumToString(value);
         }
 
         return value;
@@ -288,6 +290,10 @@ export abstract class AbstractSqliteDriver implements Driver {
 
         } else if (columnMetadata.type === "simple-json") {
             value = DateUtils.stringToSimpleJson(value);
+
+        } else if( columnMetadata.type === "simple-enum" ) {
+            value = DateUtils.stringToSimpleEnum(value, columnMetadata);
+
         }
 
         if (columnMetadata.transformer)
@@ -384,6 +390,9 @@ export abstract class AbstractSqliteDriver implements Driver {
 
         } else if (column.type === "simple-json") {
             return "text";
+
+        } else if (column.type === "simple-enum") {
+            return "varchar";
 
         } else {
             return column.type as string || "";
--- a/src/driver/sqlite-abstract/AbstractSqliteDriver.ts
+++ b/src/driver/sqlite-abstract/AbstractSqliteDriver.ts
@@ -291,7 +291,7 @@ export abstract class AbstractSqliteDriver implements Driver {
         } else if (columnMetadata.type === "simple-json") {
             value = DateUtils.stringToSimpleJson(value);
 
-        } else if( columnMetadata.type === "simple-enum" ) {
+        } else if ( columnMetadata.type === "simple-enum" ) {
             value = DateUtils.stringToSimpleEnum(value, columnMetadata);
 
         }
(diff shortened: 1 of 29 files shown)


src/driver/sqlite-abstract/AbstractSqliteDriver.ts
```diff
@@ -288,6 +290,10 @@ export abstract class AbstractSqliteDriver implements Driver {
 
         } else if (columnMetadata.type === "simple-json") {
             value = DateUtils.stringToSimpleJson(value);
+
+        } else if ( columnMetadata.type === "simple-enum" ) {
```
pleerock: since sqlite supports `enum` now, it should have `enum` type as well

pleerock: also, don't forget to list new types in the docs (website)

pleerock: or using CHECK constraint in sqlite can't be called `enum` ?

pleerock:
Great work! Asking @AlexMesser for review

src/driver/sqlite-abstract/AbstractSqliteDriver.ts
borremosch: I cannot find native enum support on the SQLite website: https://www.sqlite.org/datatype3.html.

I would say it makes sense to have `enum` for native enum types, and `simple-enum` for the emulated enum type. That way it will be more clear when you are using a derived/emulated type.

borremosch: There is currently no section on either enums or `simple-` types on the website. Do you want me to create one in a pull request?

Also, are you aware that the website certificate is invalid when accessing it from https://typeorm.io/?

pleerock: > Also, are you aware that the website certificate is invalid when accessing it from https://typeorm.io/?

yes =(

> There is currently no section on either enums or `simple-` types on the website. Do you want me to create one in a pull request?

yes feel free please

pleerock:
Thank you so much!

ToonvanStrijp:
@borremosch thanks! :)

JVMartin:
Found a bug in this PR:
```typescript
export enum FromType {
  NO_REPLY = 'no-reply',
  RANDOM = 'random',
}

@Entity('broadcast')
class BroadcastEntity {
  @PrimaryGeneratedColumn()
  id: number;

  @Column({
    type: 'simple-enum',
    enum: FromType,
    nullable: false,
  })
  from: FromType;
}
```

Here's the error:
```
  console.log ../node_modules/typeorm/platform/PlatformTools.js:200
    query failed: CREATE TABLE "broadcast" ("id" integer PRIMARY KEY AUTOINCREMENT NOT NULL, "from" varchar CHECK( from IN ('no-reply','random') ) NOT NULL)

  console.log ../node_modules/typeorm/platform/PlatformTools.js:200
    error: [Error: SQLITE_ERROR: near "from": syntax error] {
      errno: 1,
      code: 'SQLITE_ERROR'
    }
```

As you can see, the generated column:
`"from" varchar CHECK( from IN ('no-reply','random') ) NOT NULL`
is missing the double quotes that should be surrounding the column name, and so a protected keyword like `from` cannot be used.

It should look like this:
`"from" varchar CHECK( "from" IN ('no-reply','random') ) NOT NULL`

opgbaudouin:
When using simple-enum and using 'synchronise' in the connection options i get a continuous "column has changed errors" (i.e. it keeps ALTER table).

`
   //simple-enum causes TypeORM to think columns have always changed...
   @Column({name: 'type'/*, type: 'simple-enum', enum: ProjectType*/}) 
   @IsEnum(ProjectType) //class-validator
    type: ProjectType;
`
Removing it naturally removes the CHECK IN feature but there is some error in detecting if a column schema is the same and should be altered.

This is in TypeORM 0.2.25 and SQLite 3 adapter 4.2.0.

src/driver/sqlite-abstract/AbstractSqliteDriver.ts
janpio: 2 other `simple-` types are now mentioned, but `simple-enum` is not:
https://typeorm.io/entities#simple-array-column-type
https://github.com/typeorm/typeorm/issues/10077

Issue #3705, opened 2019-02-25
no such file or directory highlight.js/tools/build.js when installing typeorm


yogevlahyani:
**Issue type:**

[ ] question
[x] bug report
[ ] feature request
[ ] documentation issue

**Database system/driver:**

[ ] `cordova`
[ ] `mongodb`
[ ] `mssql`
[ ] `mysql` / `mariadb`
[ ] `oracle`
[ ] `postgres`
[ ] `sqlite`
[ ] `sqljs`
[ ] `react-native`
[ ] `expo`

**TypeORM version:**

[x] `latest`
[ ] `@next`
[ ] `0.x.x` (or put your version here)

**Steps to reproduce or a small repository showing the problem:**
I tried to install typeorm in a clean new directory (no other dependencies except typeorm) and I got this error:
`ENOENT: no such file or directory, chmod '/Users/user/dev/nestjs-custom-base-entity/node_modules/highlight.js/tools/build.js'`

Also happens when I tried to install typeorm globally with sudo. tried to clean npm cache couple of time.

yogevlahyani:
Works for me when I run `npm i typeorm --no-bin-links`

Vinlow:
Got the same issue. Thanks for the quickfix @yogevlahyani

NachtmahrEDM:
Got the same problem, @yogevlahyani workaround helped.

hk-Edifixio:
How would this workaround work on a deployment server's npm install?

BelgianCrowbar:
Looks like this is a bug from highlight.js latest version 
https://github.com/highlightjs/highlight.js/issues/1984

abingoal:
> Looks like this is a bug from highlight.js latest version
> [highlightjs/highlight.js#1984](https://github.com/highlightjs/highlight.js/issues/1984)

My solution: `npm i highlight.js@9.14.2 -D`.
They have solved this problem,waiting for the new version.

WiRight:
Try use `yarn`
It's solved for me

LucasBrd:
Seems that typeorm depends on cli-highlight which itself depends on highlight.js 9.6.0

> "dependencies": {
>     "chalk": "^2.3.0",
>     "highlight.js": "^9.6.0",
>     "mz": "^2.4.0",
>     "parse5": "^4.0.0",
>     "yargs": "^11.0.0"
>   },

Stradivario:
[...]

PR #3714, opened 2019-02-25
fix-hilight-cli

NaturesProphet:
remove this bugged lib and get it working again please!


abingoal:
I think that highlighting the sql string in the terminal is very helpful for debugging. Why not try this awsome feature?
About the problem of highlight-cli, you can use a [temporary solution](https://github.com/typeorm/typeorm/issues/3705) and wait for the official fix.

pleerock:
Right

PR #3727, opened 2019-02-27
Connection.dropDatabase() leaks an unreleased db client

mwilniewiec:
QueryRunner in Connection.dropDatabase() must be released similarly to Connection.query() method.
```
try {
...
} finally {
    queryRunner.release()
}
```
I had a situation where no more queries could be done because I had 10 clients leaked in the query pool.
In my case dropDatabase() method failed because of a deadlock in postgres, this caused no release() to be called. Postgres Pool has a default of max 10 connections. When all 10 connections were failed dropDatabase calls (not released), no more queries could be performed and hanged forever.


pleerock:
Thank you!

PR #3729, opened 2019-02-27
@next test fixes

csuich2:
I've attempted to fix the tests failing on @next.

### DIFF
--- a/package.json
+++ b/package.json
@@ -101,8 +101,8 @@
   },
   "scripts": {
     "test-ci": "gulp ci-tests",
-    "test": "tsc && mocha --file ./build/compiled/test/utils/test-setup.js --bail --recursive --timeout 30000  ./build/compiled/test",
-    "test-no-compile": "mocha --file ./build/compiled/test/utils/test-setup.js --bail --recursive --timeout 30000  ./build/compiled/test",
+    "test": "tsc && mocha --file ./build/compiled/test/utils/test-setup.js --recursive --timeout 30000  ./build/compiled/test",
+    "test-no-compile": "mocha --file ./build/compiled/test/utils/test-setup.js --recursive --timeout 30000  ./build/compiled/test",
     "compile": "tsc",
     "package": "gulp package"
   },
--- a/package.json
+++ b/package.json
@@ -101,8 +101,8 @@
   },
   "scripts": {
     "test-ci": "gulp ci-tests",
-    "test": "tsc && mocha --file ./build/compiled/test/utils/test-setup.js --recursive --timeout 30000  ./build/compiled/test",
-    "test-no-compile": "mocha --file ./build/compiled/test/utils/test-setup.js --recursive --timeout 30000  ./build/compiled/test",
+    "test": "tsc && mocha --file ./build/compiled/test/utils/test-setup.js --bail --recursive --timeout 30000  ./build/compiled/test",
+    "test-no-compile": "mocha --file ./build/compiled/test/utils/test-setup.js --bail --recursive --timeout 30000  ./build/compiled/test",
     "compile": "tsc",
     "package": "gulp package"
   },
--- a/src/entity-manager/EntityManager.ts
+++ b/src/entity-manager/EntityManager.ts
@@ -554,6 +554,7 @@ export class EntityManager {
         }
 
         if (findOptions.where || metadata.primaryColumns.length > 1) {
+            qb.setFindOptions(findOptions);
             return qb.andWhereInIds(ids).getMany();
         }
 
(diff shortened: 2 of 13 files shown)


src/entity-manager/EntityManager.ts
```diff
@@ -554,6 +554,7 @@ export class EntityManager {
         }
 
         if (findOptions.where || metadata.primaryColumns.length > 1) {
+            qb.setFindOptions(findOptions);
```
csuich2: qb.setFindOptions() was called later on in this method, but not in this case. Without it, tests were failing because there were no findOptions.

src/find-options/FindOptionsUtils.ts
```diff
@@ -90,6 +90,10 @@ export function normalizeFindOptions<T>(options: FindOptions<T>): FindOptions<T>
         if (where instanceof Array)
             return where.map((where: any) => recursivelyWhere(where));
 
+        if (Object.keys(where).every(key => !(where[key] instanceof Object && !(where[key] instanceof FindOperator)))) {
+            return where;
+        }
```
csuich2: In the test case from the same commit, a transformer was throwing an error because ultimately the transformed value was replaced by a recreation of the value, but it didn't have the getValue() method that the transformer expected. So, if we all values of this object are not objects do not meet the criteria we're about to use to recursively search, we can bail out early and not modify it.

src/query-builder/SelectQueryBuilder.ts
```diff
@@ -2318,6 +2318,10 @@ export class SelectQueryBuilder<Entity> extends QueryBuilder<Entity> implements
         }
     }
 
+    protected transformParameter(value: any, column: ColumnMetadata) {
+        return column.transformer && column.transformer.to ? column.transformer.to(value) : value;
+    }
```
csuich2: This addresses another part of the same test issue in the same commit. In the event the value is a FindOperator, we need to transform the parameters inside it, not the FindOperator itself.

test/functional/persistence/entity-updation/persistence-entity-updation.ts
```diff
@@ -28,7 +28,7 @@ describe("persistence > entity updation", () => {
         const post = new PostUuid();
         post.text = "Hello Post";
         await connection.manager.save(post);
-        const loadedPost = await connection.manager.findOne(PostUuid);
+        const loadedPost = await connection.manager.findOne(PostUuid, post.id);
```
csuich2: There are several tests that I fixed like this. I'm not sure if this is your preferred fix, though. findOne() now throws an error if no findOptions are specified.

test/github-issues/184/issue-184.ts
```diff
@@ -26,7 +26,7 @@ describe("github issues > #184 [Postgres] Single-Inheritance not working with in
         employee.shared = "e";
 
         await employeeRepository.save(employee);
-        await employeeRepository.findOne(1);
+        await employeeRepository.findOne(employee.id);
```
csuich2: ```employee.id``` is a string, not an integer, so Postgres throws an error. Is this supposed to work?

csuich2:
@pleerock  @Kononnable @AlexMesser I took a stab at attempting to fix the tests that are broken in @next. I believe I found the root cause and fix for some of them. Others, I may have just put a bandaid on.

There are still 2 tests that are failing on the node 9 CI tests, but I can't reproduce them locally.

package.json
```diff
@@ -101,8 +101,8 @@
   },
   "scripts": {
     "test-ci": "gulp ci-tests",
-    "test": "tsc && mocha --file ./build/compiled/test/utils/test-setup.js --bail --recursive --timeout 30000  ./build/compiled/test",
```
Kononnable: Was it intentional?

csuich2: Good question. Not really. It was very useful for my testing, but I did meant to ask as a part of the PR what people's thoughts on removing it are. It would be nice during test runs & CI builds to know if more than 1 test is failing, rather than having the test run stop after 1 failure.

I'll undo this if people prefer the fail early approach.

Kononnable: Sometimes when test fails there are still open db connections. It makes following tests fail because they cannot open new connection. I'm not sure, but I think this is why bail option was added - to fail quickly and don't print next 200 errors which are passing when only one test is ran.

In local development I use both commands(with and without bail) so for me it doesn't really matter what is 'the standard one'.

csuich2: Interesting. I never experienced that, but I can definitely see how that would happen. I've reverted this change and will keep working without bail locally.

pleerock:
I merged next into master just now, we can close it now.

pleerock:
thank you anyway 😆

csuich2:
Won't these tests still be failing in master now due to the changes in next that were merged in?

pleerock:
what do you mean? I merged master into the next

csuich2:
I misunderstood - I thought you said you merged next into master and were moving on vNext dev.

Either way, I'm surprised that the tests are fixed just by merging master. Regardless, glad they got fixed!

pleerock:
no, of course I fixed them after I merged into master. Not all of them fixed however, I realized that I only fixed mysql. I'll take care about rest of them maybe this weekend

PR #3737, opened 2019-03-01
Decode the user info of a database URI

### LINKED ISSUE #3716: The user info part of an URI should be decoded
**Issue type:**
Bug report

**Database system/driver:**
- `mysql` / `mariadb`
- `postgres`

**TypeORM version:**
`0.2.14`

If I use this `ormconfig.yaml` file, I can connect to my database:

```yaml
default:
  type: postgres
  host: localhost
  port: 5432
  database: foo_bar
  username: postgres
  password: Foo@Bar@      # Note the "at" signs!
```

But if I use an `url` instead, my password is refused:

```yaml
default:
  type: postgres
  url: 'postgres://postgres:Foo%40Bar%40@localhost:5432/foo_bar'
```

The password sent to the database is `Foo%40Bar%40` instead of `Foo@Bar@`. So if I want to use an `url`, I must use a password with only "standard" characters.

That's a wrong behavior : the user info in an URI can contain special characters that are URL-encoded, and so they must be decoded before using them to connect to the database.

cedx:
Fixes #3716.
Does it require a test?

### DIFF
--- a/src/driver/DriverUtils.ts
+++ b/src/driver/DriverUtils.ts
@@ -61,8 +61,8 @@ export class DriverUtils {
         return {
             type: type,
             host: host,
-            username: username,
-            password: password,
+            username: decodeURIComponent(username),
+            password: decodeURIComponent(password),
             port: port ? parseInt(port) : undefined,
             database: afterBase || undefined
         };
--- a/test/github-issues/3737/issue-3737.ts
+++ b/test/github-issues/3737/issue-3737.ts
@@ -1,0 +1,34 @@
+import { DriverUtils } from "../../../src/driver/DriverUtils";
+import { expect } from "chai";
+
+describe("github issues > #3737 Should URL-decode the user info of a data source URI", () => {
+
+    it("should parse URL with percent-encoded username", () => {
+        const obj: any = {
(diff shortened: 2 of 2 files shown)


src/driver/DriverUtils.ts
```diff
@@ -61,8 +61,8 @@ export class DriverUtils {
         return {
             type: type,
             host: host,
-            username: username,
-            password: password,
+            username: decodeURIComponent(username),
+            password: decodeURIComponent(password),
```
pleerock: are you sue other parameters like host shouldn't be decoded as well?

pleerock:
ideally we need a test but not sure if we can setup integration test. So we probably have to make a unit test.

src/driver/DriverUtils.ts
cedx: In theory, we should decode every part that contains "reserved" characters (cf. https://en.wikipedia.org/wiki/Percent-encoding). In practice, in the 15 years of my career, I haven't see any hostname or database name using reserved characters.

cedx:
I added some unit tests.

pleerock:
Thank you for your contribution!

Issue #2498, opened 2018-07-11
Virtual Field in Entity


totalard:
**Issue type:**

[X] question
[ ] bug report
[ ] feature request
[ ] documentation issue

**Database system/driver:**

[ ] `cordova`
[ ] `mongodb`
[ ] `mssql`
[ ] `mysql` / `mariadb`
[ ] `oracle`
[X] `postgres`
[ ] `sqlite`
[ ] `sqljs`
[ ] `react-native`
[ ] `expo`

**TypeORM version:**

[X] `latest`
[] `@next`
[ ] `0.x.x` (or put your version here)

How can i create virtual fields in Entity
that is model has field, But it will not stored in the DB. it is defined as a function. the function defined at the entity, Every time the entity loaded this field act as normal fields, But only Readonly,

Hope You understood

pleerock:
Provide a detailed proposal on a feature request with code examples.

totalard:
It is Very Easy to Understand @pleerock 

Just a Field in Entity, Which is not stored in DB (Has No relation to DB)
But the value is calculated using other fields value (Eg : Total Amount)

For Every Query to Entity or in Json Response,, The Field Must be there as other DB Field. But cannot Write value in it,, Because it is a Calculated Field.

Hope U Understood.

A Definition:
Fields can have values calculated by a function, instead of simply reading a database stored value. A computed field is declared just like a regular field,

s97712:
Like `asExpression`, but support all database in `typeorm`?

s97712:
I also need this, are you accept PR for this? @pleerock

nilfalse:
+1

thatisuday:
@totalard Follow my answer here: https://github.com/typeorm/typeorm/issues/1288#issuecomment-478337948

danjudex:
+1

totalard:
Any Feedback?

victorigualada:
I believe that what you are looking for if the method `getRawMany()`.

If you have an entity:
```
@Entity()
export class Employee {

  @Column('name')
  name: string;

  @Column('start_date')
  startDate: Date;

  // this is your "virtual field", as you called it, that remains a TS property
  // without the @Column() decoractor
  daysInEnterprise: number;
}
```

Then, in your query you could do:

```
  const employee: Employee = await getConnection()
    .createQueryBuilder(Employee, 'employee')
[...]

PR #3742, opened 2019-03-02
feat: Virtual column

s97712:
Support virtual column, that add new option `asVirtual` for `Column`, like `asExpression`, but support more database. 

Usage in [test cases](https://github.com/s97712/typeorm/tree/virtual-column/test/functional/repository/virtual-column)

Related issue: #2498

### DIFF
--- a/src/decorator/options/ColumnOptions.ts
+++ b/src/decorator/options/ColumnOptions.ts
@@ -1,6 +1,7 @@
 import {ColumnType} from "../../driver/types/ColumnTypes";
 import {ValueTransformer} from "./ValueTransformer";
 import { ColumnCommonOptions } from "./ColumnCommonOptions";
+import { DatabaseType } from "../..";
 
 /**
  * Describes all column's options.
@@ -118,6 +119,11 @@ export interface ColumnOptions extends ColumnCommonOptions {
     asExpression?: string;
 
     /**
+     * Like asExpression, but support all database in typeorm.
+     */
+    asVirtual?: (names: {[col: string]: string }, type: DatabaseType) => string;
+
+    /**
      * Generated column type. Supports only in MySQL.
      */
     generatedType?: "VIRTUAL"|"STORED";
@@ -150,4 +156,7 @@ export interface ColumnOptions extends ColumnCommonOptions {
      * SRID (Spatial Reference ID (EPSG code))
      */
     srid?: number;
+
 }
+
+
--- a/test/functional/repository/virtual-column/entity/User.ts
+++ b/test/functional/repository/virtual-column/entity/User.ts
@@ -1,0 +1,28 @@
+import { Entity } from "../../../../../src/decorator/entity/Entity";
+import { Column, ColumnName, PrimaryGeneratedColumn } from "../../../../../src";
+import { concat } from "./utils";
+
+
+@Entity()
+export class User {
+    @PrimaryGeneratedColumn()
+    id: number;
+
+    @Column({
+        nullable: true,
+        asVirtual: (names: ColumnName<User>, type) => {
+            try {
+                return concat(type, names.firstName, "'-'", names.lastName);
+            } catch {
+                return "'not implemented!'";
+            }
+        }
+    })
+    fullName: string;
+
+    @Column()
+    firstName: string;
+
+    @Column()
+    lastName: string;
+}
--- a/test/functional/repository/virtual-column/entity/User.ts
+++ b/test/functional/repository/virtual-column/entity/User.ts
@@ -9,7 +9,6 @@ export class User {
     id: number;
 
     @Column({
-        nullable: true,
         asVirtual: (names: ColumnName<User>, type) => {
             try {
                 return concat(type, names.firstName, "'-'", names.lastName);
--- a/test/functional/repository/virtual-column/entity/utils.ts
+++ b/test/functional/repository/virtual-column/entity/utils.ts
@@ -16,13 +16,13 @@ export function concat(type: DatabaseType, ...strings: string[]) {
 
 export function summary(type: DatabaseType, body: string, max_length: number) {
     switch (type) {
-        //case "oracle":
+        // case "oracle":
         case "sqlite":
-        //case "mysql":
-        //case "mariadb":
+        // case "mysql":
+        // case "mariadb":
             return concat(type, `substr(${body}, 0, ${max_length})`, "'...'");
         case "postgres":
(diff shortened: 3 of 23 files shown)


pleerock:
@s97712 tests are failing.

@Kononnable @vlapo what do you think about this feature?

s97712:
It can work on my test [https://travis-ci.org/s97712/typeorm/builds/501284747](https://travis-ci.org/s97712/typeorm/builds/501284747), they are same commit,  I am also confused.

src/decorator/options/ColumnOptions.ts
```diff
@@ -1,6 +1,7 @@
 import {ColumnType} from "../../driver/types/ColumnTypes";
 import {ValueTransformer} from "./ValueTransformer";
 import { ColumnCommonOptions } from "./ColumnCommonOptions";
+import { DatabaseType } from "../..";
```
Kononnable: I know what that mean, but such import looks weird.

src/decorator/options/ColumnOptions.ts
```diff
@@ -117,6 +118,11 @@ export interface ColumnOptions extends ColumnCommonOptions {
      */
     asExpression?: string;
 
+    /**
+     * Like asExpression, but support all database in typeorm.
```
Kononnable: If it's doing exactly as `asExpression` then we should enable it in other drivers(why duplicate functionality for MySQL). Probably this comment should be changed to describe what feature does not what it is similar to.

test/functional/repository/virtual-column/run_test.sh
```diff
@@ -0,0 +1,2 @@
+npx mocha --require ts-node/register repository-virtual-column.ts
```
Kononnable: unnecessary file

test/functional/repository/virtual-column/entity/utils.ts
```diff
@@ -0,0 +1,30 @@
...
+    switch (type) {
+        case "postgres":
+        case "oracle":
+        case "sqlite":
+            return strings.join("||");
+        case "mysql":
+        case "mariadb":
+            return "concat(" + strings.join(",") + ")";
+        default:
+            throw new Error(`dialect '${type}' still not implement 'concat()'`);
```
Kononnable: Adding new driver would require changes in this test file?

src/decorator/options/ColumnOptions.ts
```diff
@@ -117,6 +118,11 @@ export interface ColumnOptions extends ColumnCommonOptions {
...
+     */
+    asVirtual?: (names: {[col: string]: string }, type: DatabaseType) => string;
```
Kononnable: Why is `type` parameter needed here?

src/decorator/options/ColumnOptions.ts
s97712: I didn't notice that it was generated by ide...

src/decorator/options/ColumnOptions.ts
s97712: The type parameter can avoid to use global variables to determine database type and help to encapsulate code. If you don't consider the compatibility when defining the model, just ignore this parameter @Kononnable

test/functional/repository/virtual-column/entity/utils.ts
s97712: Don't require, it still pass test.

src/decorator/options/ColumnOptions.ts
s97712: This is just a reference implementation and may need more improvements.

Kononnable: I don't really see use of it in real application. It looks to me like it would be used only to show that tests are working in multiple database types(but test column could also do something simple like `1+1` and it would be easier to read).

s97712: It can let you encapsulate a group of sql generating function and don't consider what the database type when using they, such as  the `concat` and `summary` function in my test cases. if haven't the parameter, you need write function like `mysql_concat`, `pg_concat`, `mssql_concat`, etc, that is very inconvenience and error-prone, and the `concat` is common scenarios.

src/decorator/options/ColumnOptions.ts
pleerock: yeah some ides make auto-imports this way, don't like that either

test/functional/repository/virtual-column/entity/User.ts
```diff
@@ -0,0 +1,27 @@
...
+import { concat } from "./utils";
+
+
+@Entity()
+export class User {
+    @PrimaryGeneratedColumn()
+    id: number;
+
+    @Column({
+        asVirtual: (names: ColumnName<User>, type) => {
```
pleerock: whats typeof `type`?

pleerock:
I think we should introduce a new `@VirtualColumn` decorator instead. Also what are alternative names to `virtual`?

test/functional/repository/virtual-column/entity/User.ts
s97712: DatabaseType  

```
/**
 * Database type.
 */
export type DatabaseType =
    "mysql"|
    "postgres"|
    "cockroachdb"|
    "mariadb"|
    "sqlite"|
    "cordova"|
    "react-native"|
    "nativescript" |
    "sqljs"|
    "oracle"|
    "mssql"|
    "mongodb"|
    "expo";
```

pleerock: don't think we need it. I also don't like first parameter. It might be useful but I don't like current design. I would prefer to make it parameterless for now and think what parameters we should add based on the future feedback.

s97712: yes, may need more feedback

s97712: I think it need  alias name of table at least, the name is no fixed.

s97712:
Is there a plan to implement?

pleerock:
@s97712 yes, but you didn't provide answers to my comments

imnotjames:
Given that there's a variety of conflicts & the tests aren't passing & that it's been a while since this has been worked on I'm gonna close it out.  I believe we did add some support for these things in later versions of TypeORM.

If you do fix the conflicts and want to continue work please open a new PR that references this one!

royeradames:
https://pietrzakadrian.com/blog/virtual-column-solutions-for-typeorm#5-decorator-method

Issue #3730, opened 2019-02-28
[EntitySchema] Extend BaseSchema


johannesschobel:
**Issue type:**

[x] question
[ ] bug report
[ ] feature request
[x] documentation issue

**Database system/driver:**

[ ] `cordova`
[ ] `mongodb`
[ ] `mssql`
[ ] `mysql` / `mariadb`
[ ] `oracle`
[x] `postgres`
[ ] `sqlite`
[ ] `sqljs`
[ ] `react-native`
[ ] `expo`

**TypeORM version:**

[x] `latest`
[ ] `@next`
[ ] `0.x.x` (or put your version here)

**Steps to reproduce or a small repository showing the problem:**

Dear TypeORM maintainers and community,

first of all, i would like to thank you for this awesome project - i really like using TypeORM.. Recently, I stumbled upon `EntitySchema` for decoupling the "database structure" from the actual models - and i like this approach **very** much!

Before, when using the normal `Entity Decorators`, i did something like this in order to DRY up my boilerplate code for my entities:
```ts
export abstract class AbstractEntity {
  @PrimaryGeneratedColumn('uuid') id: string;
  @CreateDateColumn() createdAt: Date;
  @UpdateDateColumn() updatedAt: Date;
}

export class Category extends AbstractEntity {
   @Column() slug: string;
   // ...
}
```

When switching to `EntitySchema`, i found, that there is some kind of `extends` param, which can be used to `extend` an existing Schema (see code here)
https://github.com/typeorm/typeorm/blob/b9d2756c1726728da85c36aeedd924230dc601e6/src/entity-schema/EntitySchemaOptions.ts#L13-L18

However, I did not find any real documentation about this feature so far - not even in the sample.

Basically, what i would like to do, is something like this:
```ts
export const BaseSchema = new EntitySchema({
  name: 'BaseSchema',
  columns: {
    id: {
      type: 'uuid',
      generated: 'uuid',
      primary: 'true',
    },
    // ... more default column descriptions here
  },
  // maybe some other "default" descriptions here
});
```

and then "use" this schema as blueprint for other schemas, like so:
```ts 
export const CategorySchema = new EntitySchema<Category>({
  name: 'Category',
  tablename: 'categories',
  extends: 'BaseSchema',
  columns: {
    slug: {
      type: 'varchar',
      length: 50,
      unique: true,
    }
  }
});
```
[...]

PR #3747, opened 2019-03-02
[Docs] Write about EXTENDING and USING EntitySchemas

johannesschobel:
This PR adds some basic information about extending and using `EntitySchema`. I think, the docs for `EntitySchema` is not that prominent as it should be.. `EntitySchema` seems so cool!

This PR introduces changes discussed in #3730 

All the best


docs/separating-entity-definition.md
```diff
@@ -131,3 +137,73 @@ export const CategoryEntity = new EntitySchema<Category>({
...
+
+## Using Schemas to Query / Insert Data
+
+Of course, you can use the defined schemas in your repositories or entity manager as you would use the decorators.
+Consider the previously defined `Category` example (with its `Interface` and `CategoryEntity` schema in order to get 
+some data or manipulate the database.
+
+```ts
+// request data
+const categoryRepository = getRepository<Category>('CategoryEntity');
```
pleerock: actually what you need to supply isn't `'CategoryEntity'`, but `CategoryEntity`, e.g. 

```ts
const categoryRepository = getRepository(CategoryEntity); // type is inferred
```

string support is legacy and it might be removed in the future.

johannesschobel: I will check this tomorrow and give feedback on this.. Thanks a lot for your helpful comments!

johannesschobel:
Dear @pleerock ,

i tried testing your suggested changes in this minimal example here:
```ts
export interface User {
  id: string;
}

export const UserSchema = new EntitySchema<User>({
  name: 'UserSchema',
  tableName: 'users',

  columns: {
    id: {
      type: 'uuid',
      primary: true,
      generated: 'uuid',
    },
  }
});
```

However, the IDE makes "red squiggly lines" when creating the repository like so:
```ts
const userRepo = getRepository<User>(UserSchema); // note the missing ''!
```
whereas `User` is the `Interface` and `UserSchema` is the `EntitySchema` described above.

The error raised by TS is:
```
TSError: ⨯ Unable to compile TypeScript:
src/modules/application/foobar.service.ts(28,39): error TS2345: Argument of type 'EntitySchema<User>' is not assignable to parameter of type 'string | Function | (new () => User)'.
  Type 'EntitySchema<User>' is not assignable to type 'new () => User'.
    Type 'EntitySchema<User>' provides no match for the signature 'new (): User'.
```

am i missing something!?

johannesschobel:
wait.. sorry.. i did not have the latest version of `typeorm` installed (although i have it defined as `^0.2.11`, which should install the `^0.2.14` as well!)
However, with the latest version it properly compiles!

please ignore the message before, haha

docs/separating-entity-definition.md
johannesschobel: i pushed a new commit with the changes

johannesschobel:
the error should not be related to my PR

pleerock:
Thank you!

PR #3745, opened 2019-03-02
Fix Docs (Typos, Markdown, ...)

johannesschobel:
This PR fixes some minor issues in the `EN` docs.
* it corrects some Markdown Anchors to refer to the correct headlines
* it fixes some typos

Feedback is welcome!

PR #3752, opened 2019-03-04
sync doc update

abingoal:
Due to language differences, some typo errors in #3745 not be changed in Chinese. So this PR only fix anchors.


pleerock:
Thank you!

Issue #3127, opened 2018-11-19
typeorm v0.2.9 MySQL - Ignoring invalid configuration option passed to Connection


rmuchall:
**Issue type:**

[ ] question
[x] bug report
[ ] feature request
[ ] documentation issue

**Database system/driver:**

[ ] `cordova`
[ ] `mongodb`
[ ] `mssql`
[x] `mysql` / `mariadb`
[ ] `oracle`
[ ] `postgres`
[ ] `sqlite`
[ ] `sqljs`
[ ] `react-native`
[ ] `expo`

**TypeORM version:**

[ ] `latest`
[ ] `@next`
[x] `0.2.9` (or put your version here)

When using typeorm v0.2.9 on MySQL the following error message occurs in the console when establishing a connection:
`
Ignoring invalid configuration option passed to Connection: acquireTimeout. This is currently a warning, but in future versions of MySQL2, an error will be thrown if you pass an invalid configuration options to a Connection
`

rmuchall:
Reverting to typeorm v0.2.8 "fixes" the issue.

bmccann36:
I am using v0.2.8 and just started experiencing this issue

bmccann36:
I was inadvertently using 2.9
I got confused because I was actually getting a different error that had to do with ssl config and I was conflating the two but they are caused by different things.
downgrading the version fixes the error mentioned in this thread
thanks

zaksesti-ns8:
I think there is a related issue maybe with options in general.  I also experienced the above.   
In addition I'm not able to connect with `mysql@2.6.4` and `typeorm@0.2.9`.   
`typeorm@0.2.8` is OK.   
` Access denied for user ''@'172.17.0.1' (using password: NO)`  

Options:
```
  const options: ConnectionOptions = {
    type: 'mysql',
    url: 'mysql://localhost:3306/',
    username: 'root',
    database: 'protect',
    entities: [Post],
    migrations: [`${__dirname}/migrations/**/*.{t,j}s`],
    migrationsRun: true,
    timezone: 'Z',
    logging: true,
    dropSchema: true,
  };
```

josecolella:
I'm also experiencing this issue after upgrade to latest typeorm

NoelBaron:
I received this error as well using 0.2.9. Although it doesn't prohibit a connection today, it looks like the MySQL team will reject connections soon.

[...]

PR #3753, opened 2019-03-04
avoid option.acquireTimeout when using mysql2

rchaser53:
related: https://github.com/typeorm/typeorm/issues/3127

### DIFF
--- a/src/driver/mysql/MysqlDriver.ts
+++ b/src/driver/mysql/MysqlDriver.ts
@@ -820,7 +820,6 @@ export class MysqlDriver implements Driver {
         return Object.assign({}, {
             charset: options.charset,
             timezone: options.timezone,
-            acquireTimeout: options.acquireTimeout,
             connectTimeout: options.connectTimeout,
             insecureAuth: options.insecureAuth,
             supportBigNumbers: options.supportBigNumbers !== undefined ? options.supportBigNumbers : true,
@@ -837,7 +836,11 @@ export class MysqlDriver implements Driver {
             database: credentials.database,
             port: credentials.port,
             ssl: options.ssl
-        }, options.extra || {});
+        },
+        options.acquireTimeout === undefined
+          ? {}
+          : { acquireTimeout: options.acquireTimeout },
+        options.extra || {});
     }
 
     /**


rchaser53:
I have started using this library today. But this library emits the warning message when using with mysql2. It's a bit noisy. So I sent the this PR.

src/driver/mysql/MysqlDriver.ts
```diff
@@ -802,7 +808,7 @@ export class MysqlDriver implements Driver {
         } catch (e) {
             try {
                 this.mysql = PlatformTools.load("mysql2"); // try to load second supported package
-
+                this.version = 2;
```
johannesschobel: can we have the same order as above (first set version, then load packages) or vice-versa.. but not mixed?

src/driver/mysql/MysqlDriver.ts
```diff
@@ -829,7 +834,12 @@ export class MysqlDriver implements Driver {
             debug: options.debug,
             trace: options.trace,
             multipleStatements: options.multipleStatements,
-            flags: options.flags
+            flags: options.flags,
+            ...(
+                this.version === 1
+                    ? { acquireTimeout: options.acquireTimeout } 
+                    : {}
+                )
```
johannesschobel: i think this statement is particularly hard to read.. but that is just my opinion

src/driver/mysql/MysqlDriver.ts
```diff
@@ -788,6 +793,7 @@ export class MysqlDriver implements Driver {
      */
     protected loadDependencies(): void {
         try {
+            this.version = 1;
```
johannesschobel: i would rather not use `1`, `2` as version tags but rather `mysql`, `mysql2` and so on..
further, can we use this maybe as an enum for checks later on?

rchaser53:
Thank you the review. I fixed it.

src/driver/mysql/MysqlDriver.ts
```diff
@@ -37,6 +45,11 @@ export class MysqlDriver implements Driver {
      */
     mysql: any;
 
+    /**
+     * Mysql version.
+     */
+    version: MySQLVersion;
```
pleerock: I'm sorry to annoy you but we don't really need enum here. Regular string literal type is enough. (e.g. replace `version: MySQLVersion` to `version: "mysql"|"mysql2")

pleerock:
Thank you for your contribution! I have left one comment, but its not critical to fix. If you have time to fix that please send another PR while I'm going to merge this one.

rchaser53:
Good idea! I changed my PR.

pleerock:
Actually I meant just traverse over all object properties and filter out undefined onces. But thats fine too.

pleerock:
Thank you!

PR #3756, opened 2019-03-04
Allow mongodb projection

### LINKED ISSUE #3266: Nested Select in mongodb
**Issue type:**

[x] question
[ ] bug report
[ ] feature request
[ ] documentation issue

**Database system/driver:**

[ ] `cordova`
[x] `mongodb`
[ ] `mssql`
[ ] `mysql` / `mariadb`
[ ] `oracle`
[ ] `postgres`
[ ] `sqlite`
[ ] `sqljs`
[ ] `react-native`
[ ] `expo`

**TypeORM version:**

[x] `latest`
[ ] `@next`

I am trying to do a nested select for fetching out particular field of a nested array of objects in mongodb. Currently the select that worked for me looks like this.

```ts 
    const data = await StudentRepository.find({
      select: ["id", "name"],
    });
```

The structure of the object in mongo looks like this

```json
{  
   "_id":"5c07b4bac10fcb38083d39b5",
   "name":"Tim",
   "address":[  
      {  
         "pincode":123,
         "other":"lorem ipsum"
      },
      {  
         "pincode":456,
         "other":"lorem ipsum"
      }
   ]
}
```

Here `address` is an array contained inside the `student object`

I would like to do something like this 

```ts 
    const data = await StudentRepository.find({
      select: ["id", "name","address.pincode"],
    });
```

The final structure that I am trying to achieve is 

```json
{  
   "_id":"5c07b4bac10fcb38083d39b5",
   "name":"Tim",
   "address":[  
      {  
         "pincode":123
      },
      {  
         "pincode":456
      }
   ]
}
```

rustamwin:
Fixes #3266


pleerock:
We can't go with such changes. If we keep going with mongodb-specific changes entity-manager/repository is going to be a mess for rdbms and we'll always be blocked on some absolute incompatible features. 

I suggest to remove dependency of MongoEntityManager from EntityManager (e.g. remove extends) and completely re-implement MongoEntityManager with mongodb-specific features.

ElianCordoba:
Seams reasonable, could you provide some initial code/steps to get to that? I don't feel confident enough to take that task ¯\_(ツ)_/¯

pleerock:
1. remove `extends EntityManager`
2. implement same functions we have before in EntityManager
3. start improving mongodb support

jyang0110:
why am I getting redirected to here when no solution is given...
Thanks for the work tho

phil294:
@rustamwin I am trying to understand your pull request: Why is it necessary to add a new `projection` option? There is already a
https://github.com/typeorm/typeorm/blob/master/src/entity-manager/MongoEntityManager.ts#L96
method which changes `select` into `projection`.
Wouldnt the linked issue (#3266) be fixed by simply adjusting `convertFindOptionsSelectToProjectCriteria` so that nested keys in `select` are put into the projection too?

I am having a hard time understanding why the syntax `  select: ["id", "name","address.pincode"],` doesnt work already in general.

Thanks!

rustamwin:
> Why is it necessary to add a new projection option? There is already a
https://github.com/typeorm/typeorm/blob/master/src/entity-manager/MongoEntityManager.ts#L96
method which changes select into projection.

because `select` doesn't allow to use mongo `projection`, but just allow fields
mongodb has a magical `projection` :smile: 
https://docs.mongodb.com/manual/reference/operator/projection/

phil294:
@rustamwin I'm afraid I dont understand. I know that for "select" syntax in MongoDB, you have to use `projection`. But thats what https://github.com/typeorm/typeorm/blob/master/src/entity-manager/MongoEntityManager.ts#L96 already does, doesnt it? It transforms `select` into `projection` when using the mongo driver.  `convertFindOptionsSelectToProjectCriteria` already transforms a `["name", "address.pincode"]` into a `{ name: 1, 'address.pincode': 1 }` projection. So it seems to me the nested mongo "select" (internally handled as a projection) should already work. What am I missing?

gruckion:
Is this going to be merged? Would be nice to have this support.

kateile:
waiting for this merge. projection is awesome feature in mongoDb

robbyemmert:
Is this going to be merged? I think this will help my situation (and probably a lot of others, by the look of it).

rustamwin:
Won't be merged before this:
> I suggest to remove dependency of MongoEntityManager from EntityManager (e.g. remove extends) and completely re-implement MongoEntityManager with mongodb-specific features.

MowiliMi:
any progress with merge?

imnotjames:
@phil294 I agree with you - one, on your most recent comment about courtesy and etiquette.

Thank you. 

But secondly, on reusing select. Reusing select would bypass a lot of the work pleerock had suggested and would maintain a smaller, more cohesive API.  We wouldn't have lots of custom mongodb options.

I'd personally prefer that we had a not as great mongodb experience for a more consistent experience across all of these data stores.

If you're after something with super deep mongodb feature support a generic ORM probably ain't gonna cut it.  It'll never have feature parity with something that's built out with mongodb as it's primary goal.  It can't.

---

I also see that more or less development on this PR has stalled for a few years.  I'm gonna close it out because otherwise we end up with folks spamming it with.. well, various comments that aren't constructive towards the PR itself. 


Please feel free to open a new PR and definitely continue discussion of this feature request in #3266

Issue #1934, opened 2018-04-13
Mongo: db actions aren't logged in console even if passing in logging: true


mcMickJuice:
**Issue type:**

[ ] question
[x] bug report
[ ] feature request
[ ] documentation issue

**Database system/driver:**

[ ] `cordova`
[x] `mongodb`
[ ] `mssql`
[ ] `mysql` / `mariadb`
[ ] `oracle`
[ ] `postgres`
[ ] `sqlite`
[ ] `sqljs`
[ ] `websql`

**TypeORM version:**

[] `latest`
[ ] `@next`
[x] `0.1.20` (or put your version here)

I'm not seeing any database related logging when setting `logging: true` in the connectionOptions.  Nor am I seeing logging when setting logging to true in the `ormconfig.json` file

Here is a repo where this is occurring:

https://github.com/mcMickJuice/typeorm-logger-bug

cojack:
Still nothing? After year?!

vlapo:
Just adding reference to PR https://github.com/typeorm/typeorm/pull/3768

jackykwandesign:
> Still nothing? After year?!

yes, after 3 year still nothing

imnotjames:
> > Still nothing? After year?!
> 
> yes, after 3 year still nothing, just tested and find here..
> seems typeorm had a bad support in mongo

We're always accepting pull requests improving support.

mihailShumilov:
hm... issue still actual...

danielehrhardt:
Still Issue?

KCFeng:
I still have this issue.

jackykwandesign:
same here

VictorManduca:
Same here

r0und:
Same here.

xenoterracide:
yet another bump, tf... it's not even calling the logger, I created a stub with throw errors.... and no errors are even being thrown.

xenoterracide:
@imnotjames @pleerock if I were to revive, or rewrite the PR how soon could this happen?

mostafa8026:
any update :(

zgldh:
"typeorm": "^0.2.41",
    "@nestjs/typeorm": "8.0.4",
    "nestjs-pino": "^2.5.0",


Not working.

denysshnurenko:
still actual

riarheos:
I believe it is not a typeorm bug but rather a mongodb driver one.
Their official documentation states that they removed the logger from it and they specify the workaround:
[...]

PR #3768, opened 2019-03-06
Mongo logging

Matthijn:
As in this issue, Mongo DB queries were not logged. Now most of them are. (Delete's,  fetches, updates, etc).

https://github.com/typeorm/typeorm/issues/1934


src/logger/SimpleConsoleLogger.ts
```diff
@@ -44,7 +44,7 @@ export class SimpleConsoleLogger implements Logger {
      * Logs query that is slow.
      */
     logQuerySlow(time: number, query: string, parameters?: any[], queryRunner?: QueryRunner) {
-        const sql = query + (parameters && parameters.length ? " -- PARAMETERS: " + this.stringifyParams(parameters) : "");
+        const sql = query + (parameters && (parameters.length > 0 || parameters.length === undefined) ? " -- PARAMETERS: " + this.stringifyParams(parameters) : "");
```
pleerock: can you please explain this `parameters.length === undefined` part?

pleerock: I guess we have it because parameters can be object now? If yes then few concerns: 

* better and more intuitive logic should be applied instead of this code
* do we really need convert mongodb queries into strings in logging?

pleerock:
Thank you for your contribution! Adding @rustamwin

src/logger/SimpleConsoleLogger.ts
Matthijn: Yes, the mongo parameters are a dictionary indeed, they don't have a length. Did it this way so the minimum amount of code would be changed. You can have own loggers, and I don't know if all loggers would just be fine with having an object in there instead of a string.

rustamwin: I agree with @pleerock's think. We should keep it

rustamwin: I guess query already string, small problem with parameters.

rustamwin:
@Matthijn Thank you for your contribution. It was part of mongo's plans

> Adding @rustamwin 

@pleerock Would be great adding me to typeorm :smile:

src/logger/SimpleConsoleLogger.ts
pleerock: @Matthijn thank you for making it as minimal as possible but we need to improve it a bit to make it clean and maintainable, and improve functionality itself

Matthijn:
No problem! Do whatever you want with it. 😄

src/logger/SimpleConsoleLogger.ts
renatogbp: I am new to typescript, but how an object can be assigned to 'any[]'? Shouldn't be only 'any'? Unless it is an array of object so it would have the length defined.

ruscon:
ping

sunada-ke:
When will this be merged?

labs-plus:
Any updates? This would be super useful.

src/logger/AdvancedConsoleLogger.ts
```diff
@@ -25,7 +25,7 @@ export class AdvancedConsoleLogger implements Logger {
      */
     logQuery(query: string, parameters?: any[], queryRunner?: QueryRunner) {
         if (this.options === "all" || this.options === true || (this.options instanceof Array && this.options.indexOf("query") !== -1)) {
-            const sql = query + (parameters && parameters.length ? " -- PARAMETERS: " + this.stringifyParams(parameters) : "");
+            const sql = query + (parameters && (parameters.length > 0 || parameters.length === undefined) ? " -- PARAMETERS: " + this.stringifyParams(parameters) : "");
```
imnotjames, line 28: Should just use `Array.isArray` instead of checking if length is undefined

imnotjames (changes requested):
Needs tests & has conflicts.

AlexMesser:

PR #3776, opened 2019-03-07
Fix compare2Objects handling of null.

brandonbloom:
If you try to save a JSONB column with a null value in an object,
you get this error:

TypeError: Cannot read property 'equals' of null

### DIFF
--- a/src/util/OrmUtils.ts
+++ b/src/util/OrmUtils.ts
@@ -177,8 +177,9 @@ export class OrmUtils {
         if (x === y)
             return true;
 
-        // Unequal, but either is null or undefined.
-        if (x == null || y == null)
+        // Unequal, but either is null or undefined (use case: jsonb comparasion)
+        // PR #3776, todo: add tests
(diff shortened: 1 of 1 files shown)


pleerock:
Thank you for contribution!

We can't go with unstrict checks, if you want to check for both `null` and `undefined` better to specify both comparisons instead.

pleerock:
Also we really need a test that confirms this functionality, because next time somebody change this peace of code he can easily remove/change it and break this functionality.

brandonbloom:
Hi @pleerock, We're on an older version of TypeORM and aren't ready to upgrade anytime soon, so there is no pressing need for us to land this in to master. I probably won't get around to improving this PR for quite a while. I'm OK for you to consider this PR a bug report rather than a merge request. If you'd like to take these steps yourself, test it by creating a jsonb column in Postgresql and try changing a property in the json object to/from `null`.

pleerock:
Got you. Thank you!

PR #3782, opened 2019-03-08
Added exclude option to find methods in mongodb

### LINKED ISSUE #3781: There is no way to exclude fields easily in mongo.
**Issue type:**

[ ] question
[ ] bug report
[x] feature request
[ ] documentation issue

**Database system/driver:**

[ ] `cordova`
[x] `mongodb`
[ ] `mssql`
[ ] `mysql` / `mariadb`
[ ] `oracle`
[ ] `postgres`
[ ] `sqlite`
[ ] `sqljs`
[ ] `react-native`
[ ] `expo`

**TypeORM version:**

[ ] `latest`
[x] `@next`

The `@Column({ select: false })` does not work in mongo, so I thought about adding (PR coming soon™) an `exclude` option in all `find` methods. Basically the current `select` uses the value 1 to the projection of the query, so the `exclude` will use the value 0.

Use case: when you want to return a user from the DB and want to filter sensible information (password, salt, credit card info, etc).

[Relevant doc](https://docs.mongodb.com/manual/tutorial/project-fields-from-query-results/#return-all-but-the-excluded-fields)

ElianCordoba:
Fixes #3781 and #3266

### DIFF
--- a/docs/find-options.md
+++ b/docs/find-options.md
@@ -13,6 +13,12 @@ All repository and manager `find` methods accept special options you can use to 
 userRepository.find({ select: ["firstName", "lastName"] });
 ```
 
+* `exclude` - indicates which properties of the main object must be excluded
+
+```typescript
+userRepository.find({ exclude: ["lastName"] });
+```
+
 * `relations` - relations needs to be loaded with the main entity. Sub-relations can also be loaded (shorthand for join and leftJoinAndSelect)
 
 ```typescript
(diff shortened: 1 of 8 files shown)


ElianCordoba:
Well I feel like an idiot, just realize of #3756

pleerock:
Thank you very much! Please participate in a discussion in #3756

michaeljota:
After reading #3756, I think this would be a better approach than the one the PR introduces. Though, I understand this still adds some Mongo only options, this could be also used to introduce some kind of exclude to other managers. 

I have some ideas to improve this PR, reducing the complexity this have to introduce, but I would prefer doing something that last. 

What I like of the current implementation in TypeORM is that, even when some options aren't compatible with other RBD, it would be relative straightforward to update the DB provider in the needed case.

PR #3790, opened 2019-03-09
[WIP] installed jest and added test example

pleerock:


imnotjames:
Last updated march 15th, lots of conflicts.  I'm gonna close this out but I'm very interested in this!!  Let's chat on slack sometime about switching to jest? :)

I think we can even switch to jest while still using chai..

pleerock:
this is really good to have, but once we merge master into the next and release `next`

PR #3791, opened 2019-03-09
Change master query docs to not leak connections

ericdcobb:
Example code that demonstrates querying master instead of a replica for select queries will leak connections.

This was biting us in production and took some time to troubleshoot because we weren't actually getting good errors (on my todolist as to why good errors weren't being thrown in the case, maybe I'll submit another PR when I figure that out), but I think the example code should not be leaking connections and we should mention that the query runner needs to be released manually.

Cheers!


pleerock:
I can imagine how hard it was to reproduce issue. Thank you!

Issue #3536, opened 2019-01-28
Sync only works once for enums on entities with capital letters in entity name


onionhammer:
**Issue type:**

[ ] question
[x] bug report
[ ] feature request
[ ] documentation issue

**Database system/driver:**

[ ] `cordova`
[ ] `mongodb`
[ ] `mssql`
[ ] `mysql` / `mariadb`
[ ] `oracle`
[x] `postgres`
[ ] `sqlite`
[ ] `sqljs`
[ ] `react-native`
[ ] `expo`

**TypeORM version:**

[x] `latest`
[ ] `@next`
[ ] `0.x.x` (or put your version here)

**Steps to reproduce or a small repository showing the problem:**

Create a column of type enum array, sync the database, then try to sync it again

Example

```ts
@Column("enum", { enum: AccountPermission, array: true, default: '{}' })
accountPermission: AccountPermission[]
```

When I run typeorm schema:sync on an empty database it works, when I run it again it crashes with error

`type "pcc._roles_accountpermission_enum" does not exist`

It seems to be adding a leading `_` before the type name in one spot but not another; in the database the type is `Roles_accountpermission_enum`

onionhammer:
Additional notes

1. This does not only affect arrays

2. This only seems to fail if your `@Entity()` name has capital letters in it; such as "Roles" in the above example.

nikolovjd:
I have a similar issue, however for me the issue appears if the column name in the db has an underscore.

onionhammer:
Hey @vlapo what was the name of the \@Entity you used? I can try to create a failing test tomorrow

vlapo:
This is entity and enum definition:
```typescript
export enum AccountPermission {
    Thing1 = 1,
    Thing2 = 4,
    Thing3 = 3,
    Thing4 = 2
}

@Entity("Roles")
export class Roles {

    @PrimaryGeneratedColumn()
    id: string;

    @Column("enum", { enum: AccountPermission, array: true, default: '{}' })
    accountPermission: AccountPermission[];
}
```
Are you sure you are using latest version from npm?

vlapo:
Ou sry boys! I see error now. Had a drop schema setting enabled from previous test. Will check this later today. 🙌

OLDIN:
I have a similar problem:
Error: `type "public.customfield_type_enum" does not exist`
My entity name: `customField`
My synchronization and migrations do not work.

[...]

Issue #2724, opened 2018-08-28
Error reverting migration with enum postgres


zveljkovic:
**Issue type:**

[ ] question
[x ] bug report
[ ] feature request
[ ] documentation issue

**Database system/driver:**

[ ] `cordova`
[ ] `mongodb`
[ ] `mssql`
[ ] `mysql` / `mariadb`
[ ] `oracle`
[x] `postgres`
[ ] `sqlite`
[ ] `sqljs`
[ ] `react-native`
[ ] `expo`

**TypeORM version:**

[ ] `latest`
[ ] `@next`
[x] `0.27` (or put your version here)

**Steps to reproduce or a small repository showing the problem:**
Create migration below, run migration:run and then migration:revert.

***Expected***
To undo migration

***Actual***
Throws error `Error during migration revert:
{ QueryFailedError: type "test_status_enum" does not exist`

```import {MigrationInterface, QueryRunner, Table} from 'typeorm';

export class CreateTest1535031863784 implements MigrationInterface {

		public async up(queryRunner: QueryRunner): Promise<any> {
			await queryRunner.createTable(new Table({
				columns: [
					{
						isGenerated: true,
						isPrimary: true,
						name: 'id',
						type: 'int',
					},
					{
						enum: ['active', 'pending', 'inactive'],
						name: 'status',
						type: 'enum',
					},
				],
				name: 'Test',
			}));
		}

		public async down(queryRunner: QueryRunner): Promise<any> {
			await queryRunner.dropTable('Test');
		}
}

```

Kononnable:
May be related to #1997.

PR #3814, opened 2019-03-12
fix(3536): fix sync schema issue with postgres enum in case capital letters in entity name

vlapo:
Using uppercase letters in enum names seems to be problem in postgres. This PR convert whole enum name into lowercase (not only column name part).

Fixes https://github.com/typeorm/typeorm/issues/3536 https://github.com/typeorm/typeorm/issues/2724
Close https://github.com/typeorm/typeorm/pull/1978 in favour of this PR with tests for inactivity of original PR author.


OLDIN:
How do I get these changes now?
After all, the patch version has not changed
Or when should I wait for the patch version?

Issue #3823, opened 2019-03-14
Compilation error since v0.2.15 with TypeScript 2 (working on v0.2.14)


LoicPoullain:
**Issue type:**

[ ] question
[x] bug report
[ ] feature request
[ ] documentation issue

**Database system/driver:**

[ ] `cordova`
[ ] `mongodb`
[ ] `mssql`
[ ] `mysql` / `mariadb`
[ ] `oracle`
[ ] `postgres`
[ ] `cockroachdb`
[x] `sqlite`
[ ] `sqljs`
[ ] `react-native`
[ ] `expo`

**TypeORM version:**

[x] `latest`
[ ] `@next`
[ ] `0.x.x` (or put your version here)

**Steps to reproduce or a small repository showing the problem:**

*Project directory*
```
node_modules/
index.ts
package.json
```

*index.ts*
```typescript
import { createConnection } from 'typeorm';

async function main() {
  await createConnection();
}
```

*package.json (version 1)*
```json
{
  "scripts": {
    "build": "tsc index.ts"
  },
  "dependencies": {
    "@types/node": "^11.11.3",
    "typeorm": "^0.2.15",
    "typescript": "^3.3.3333"
  }
}
```

*package.json (version 2)*
```json
{
  "scripts": {
    "build": "tsc index.ts"
  },
  "dependencies": {
    "@types/node": "^11.11.3",
    "typeorm": "^0.2.15",
    "typescript": "^2.9.2"
  }
}
```

Running `npm run build` works with the first version of `package.json`(TypeScript version is 3).

Running `npm run build` throws an error with the second version of `package.json` (TypeScript version is 2).

```
> @ build /Users/loicpoullain/Desktop/toto
> tsc index.ts

node_modules/typeorm/connection/Connection.d.ts:40:28 - error TS1039: Initializers are not allowed in ambient contexts.

40     readonly isConnected = false;
                              ~~~~~
```

In the [documentation](https://github.com/typeorm/typeorm#typescript-configuration), TypeORM is supposed to support version 2.3 and greater.

Installing `typeorm@0.2.14` fixes the issue.

Issue firstly reported [here](https://github.com/FoalTS/foal/issues/377) by @rick-software.

LoicPoullain:
And thank you so much for developing and maintaining this ORM. It is excellent!

eugenchio:
@LoicPoullain , check your TS version. The latest version requires ^3.3.3333.

rustamwin:
[...]

PR #3827, opened 2019-03-14
[README.md] Update the TS compiler version to 3.3 or greater

LoicPoullain:
v0.2.15 requires to use the version 3.3 or greater of the TypeScript compiler. This PR updates the documentation with this new version.

Related issue: #3823

### DIFF
--- a/README.md
+++ b/README.md
@@ -220,7 +220,7 @@ await timber.remove();
 
 ##### TypeScript configuration
 
-Also, make sure you are using TypeScript compiler version **2.3** or greater,
+Also, make sure you are using TypeScript compiler version **3.3** or greater,
 and you have enabled the following settings in `tsconfig.json`:
 
 ```json


rustamwin:
Thank you!

pleerock:
Thanks for the contribution!

Are you sure `3.3` is required? Are you sure it's not working in 3.0 or 3.1 ?

LoicPoullain:
It's working with 3.0 as well. I put `3.3` because the TypeORM repo uses `3.3.3333` in its `package.json` and `package-lock.json`. It was a way to prevent unexpected bugs in feature releases.

For example, TypeScript [introduces a breaking change](https://github.com/Microsoft/TypeScript/wiki/Breaking-Changes) between v3.0 and v3.1. A future version of TypeORM may use this one causing the library to fail on applications using v3.0.

Maybe a workaround could be to use v3.0 in the `package[-lock].json` of the repo. Maybe it's a wrong idea.

PR #3832, opened 2019-03-15
fix(relationLoader): added quotes to junctionAlias

Liapin:
I have
```@Entity('actions')
export class Action {
...
  @ManyToMany(() => User)
  @JoinTable({
    name: 'actions_users',
    joinColumn: { name: 'action_id', referencedColumnName: 'id' },
    inverseJoinColumn: { name: 'user_id', referencedColumnName: 'id' }
  })
  public users: User[]
...
```
and when I try doing smth like this:

```
action.users = { id: '1' }
await transactionManager.getRepository(Action).save(action)
```

TypeORM generated query smth like this:

```
SELECT Action_users_rid.user_id AS "user_id", Action_users_rid.action_id AS "action_id"
FROM "company_cefb9781_e8ce_43fa_8cb1_52e111cc3869"."users" "users"
       INNER JOIN "company_cefb9781_e8ce_43fa_8cb1_52e111cc3869"."actions_users" "Action_users_rid"
                  ON (Action_users_rid.action_id = $1 AND Action_users_rid.user_id = "users"."id")
ORDER BY Action_users_rid.user_id ASC, Action_users_rid.action_id ASC
```
(alias name no double-quoted). But join alias with quotes

### DIFF
--- a/src/query-builder/relation-id/RelationIdLoader.ts
+++ b/src/query-builder/relation-id/RelationIdLoader.ts
@@ -143,12 +143,12 @@ export class RelationIdLoader {
                     return Object.keys(mappedColumn).map(key => {
                         const parameterName = key + index;
                         parameters[parameterName] = mappedColumn[key];
-                        return junctionAlias + "." + key + " = :" + parameterName;
+                        return "\"" + junctionAlias + "\"." + key + " = :" + parameterName;
                     }).join(" AND ");
                 });
 
                 const inverseJoinColumnCondition = inverseJoinColumns.map(joinColumn => {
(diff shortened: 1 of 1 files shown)


src/query-builder/relation-id/RelationIdLoader.ts
```diff
@@ -143,12 +143,12 @@ export class RelationIdLoader {
                     return Object.keys(mappedColumn).map(key => {
                         const parameterName = key + index;
                         parameters[parameterName] = mappedColumn[key];
-                        return junctionAlias + "." + key + " = :" + parameterName;
+                        return "\"" + junctionAlias + "\"." + key + " = :" + parameterName;
```
pleerock: not all drivers have quotes as escaping character! You need to use `connection.driver.escape` method. Also please attach test so this functionality won't break ever again. Thank you!

PR #3845, opened 2019-03-19
Skip constructor when entity is loaded from database

XanderEmu:
I basically copied the code submitted by @pepakriz  into a new branch created against `next`, only I used JS's `Reflect` instead of `ObjectInstantiator`).

Original PR: https://github.com/typeorm/typeorm/pull/1773

### DIFF
--- a/src/metadata/EntityMetadata.ts
+++ b/src/metadata/EntityMetadata.ts
@@ -502,7 +502,8 @@ export class EntityMetadata {
         // if target is set to a function (e.g. class) that can be created then create it
         let ret: any;
         if (this.target instanceof Function) {
-            ret = new (<any> this.target)();
+            ret = {};
+            Reflect.setPrototypeOf(ret, this.target.prototype);
             this.lazyRelations.forEach(relation => this.connection.relationLoader.enableLazyLoad(relation, ret, queryRunner));
             return ret;
         }
--- a/test/github-issues/1772/entity/Post.ts
+++ b/test/github-issues/1772/entity/Post.ts
@@ -1,0 +1,24 @@
+import {Entity} from "../../../../src/decorator/entity/Entity";
+import {Column} from "../../../../src/decorator/columns/Column";
+import {PrimaryGeneratedColumn} from "../../../../src/decorator/columns/PrimaryGeneratedColumn";
+
+@Entity()
+export class Post {
+    @PrimaryGeneratedColumn()
+    id: number;
+
+    @Column()
+    private title: string;
+
+    initialized?: true;
+
+    constructor(title: string) {
+        this.title = title;
+        this.initialized = true;
+    }
+
+    public getTitle(): string {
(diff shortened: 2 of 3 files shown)


pleerock:
Thank you!

pepakriz:
Wow! Thank you too

vlapo:
What about documentation? We have a note in docs about optionality of constructor parameter (https://typeorm.io/#/entities). It should be updated in next branch too.

pleerock:
I just found one issue with this change. Right now property initializations won't work with this change:

```ts
export class Post {
  @Column()
  isLoaded = false
}
```

we don't get isLoaded initialized. I'm not yet sure about this behaviour, but it looks right too. But I would like to get your thoughts about it. Is it something expected now? Adding @nicomouss too.

pepakriz:
@pleerock It's expected in my opinion

nicomouss:
Same here, it's expected as constructor is skipped when entities are built from the ORM. Thus properties initialisations are also skipped.

Kononnable:
Right now 3 tests in `@next` are broken because of this change:
`/functional/entity-metadata-validator/initialized-relations`
`/functional/repository/basic-methods`
`/github-issues/3256`
We also have to change `Avoid relation property initializers` in relations-faq since it's outdated after merging this change. If we don't initialize properties on creation we don't need `InitializedRelationError` anymore.
But support for old method was partially introduced in #3910 so maybe we should just update those tests/docs(otherwise delete all/parts of it).

pkuczynski:
Any prospects when this would be released?

pleerock:
@pkuczynski its already released

pkuczynski:
Hmm, this is strange and does not seem to work for ActiveRecord. My example:

```js
export class Car extends BaseEntity {
  @PrimaryGeneratedColumn()
  id: number

  registration: string

  constructor() {
    console.log('Constructor has been called')
  }
}
```

And when I call `Car.findOne(1)` I can see in the console:

```
Constructor has been called
```

Kononnable:
This change is only available for 0.3.0 version. It's released under @next tag.

pkuczynski:
`0.3` is still in alpha, so a bit risky to use in production, right?

Kononnable:
I believe some folks use @next in production environments. I think is preety stable right now, but it would be best to ask people who actually use it daily to tell if they think it's usable in production.
We(as typeorm) will move to 0.3 soon, but there is still some work which need to be done and we don't have much free time to do it, so no even approximate release date yet.

pkuczynski:
Got it! Thanks for the detailed explanation. Do you have any high-level description of changes introduced by 3.0?

Kononnable:
0.3.0 wasn't released, so there is no migration guide or similiar documents. You can check [changelog](https://github.com/typeorm/typeorm/blob/next/CHANGELOG.md) - some of the changes are pointed out there.

PR #3848, opened 2019-03-19
Documentation update: add extra `includes` to migration example

AbigailMcP:
Adds necessary includes (`TableIndex`, `TableColumn` and `TableForeignKey` ) to example migration.


pleerock:
Thank you!

PR #3870, opened 2019-03-21
Fix #3549

### LINKED ISSUE #3549: Bug when using findOne and In
**Issue type:**

[x] bug report

**Database system/driver:**

[x] `mysql` / `mariadb`

**TypeORM version:**

[x] `latest`
[x] `0.2.12` (or put your version here)

Repo to reproduce: https://github.com/Slessi/typeorm-findone-in-bug

When using the following query, the parameters don't seem to be loaded correctly.

```ts
connection.getRepository(MyEntity).findOne(1, {
  where: {
    myField: In([2, 3]),
  },
}));
```

As per the following SQL log from typeorm: (Note how parameters are `[1,3]` and not `[1,2,3]`

```sql
SELECT `MyEntity`.`id` AS `MyEntity_id`, `MyEntity`.`myField` AS `MyEntity_myField` FROM `my_entity` `MyEntity` WHERE `MyEntity`.`myField` IN (?, ?) AND `MyEntity`.`id` IN (?) LIMIT 1
-- PARAMETERS: [1,3]
```

Which results in the following error

> ER_PARSE_ERROR: You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near '?) LIMIT 1'

However it does work if you include everything inside the `where: {}` like so

```ts
connection.getRepository(MyEntity).findOne({
  where: {
    id: 1,
    myField: In([2, 3]),
  },
}));
```

Kononnable:
Fixes #3549

Tests were failing only on master. On next they were passing(before a fix) because next no longer use parameters in those generated queries. We've got a regression there unless it was done on purpose.

### DIFF
--- a/src/query-builder/QueryBuilder.ts
+++ b/src/query-builder/QueryBuilder.ts
@@ -753,6 +753,7 @@ export abstract class QueryBuilder<Entity> {
                             const aliasPath = this.expressionMap.aliasNamePrefixingEnabled ? `${this.alias}.${propertyPath}` : column.propertyPath;
                             let parameterValue = column.getEntityValue(where, true);
                             const parameterName = "where_" + whereIndex + "_" + propertyIndex + "_" + columnIndex;
+                            const parameterBaseCount = Object.keys(this.expressionMap.nativeParameters).filter(x => x.startsWith(parameterName)).length;
 
                             if (parameterValue === null) {
                                 return `${aliasPath} IS NULL`;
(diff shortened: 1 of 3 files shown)


pleerock:
Not sure about next changes.. Fortunately we have a test so we can fix this on next separately later. Best if you can merge this into next too and check if test is passing and if not fix it taking in count next changes. If no thats fine we can do it later.

PR #3877, opened 2019-03-22
Added handling of synchronous errors from .connect() methods

Ginden:
This is workaround for issue that we have found:
* Sometimes `this.master` is `undefined` in PostgresDriver.ts - we are not sure why this happened, because our code never calls `.disconnect`.
* Errors thrown from `.connect` method synchronously (present in eg. `pg-pool`) can propagate as synchronous errors - leaving users surprised (end if this occurs in callback, it can kill your process).

After this PR, these errors will still occur, but users can handle them as rejected promises.

### DIFF
--- a/src/driver/cockroachdb/CockroachDriver.ts
+++ b/src/driver/cockroachdb/CockroachDriver.ts
@@ -534,7 +534,7 @@ export class CockroachDriver implements Driver {
      * Used for replication.
      * If replication is not setup then returns default connection's database connection.
      */
-    obtainMasterConnection(): Promise<any> {
+    async obtainMasterConnection(): Promise<any> {
         return new Promise((ok, fail) => {
             this.master.connect((err: any, connection: any, release: any) => {
                 err ? fail(err) : ok([connection, release]);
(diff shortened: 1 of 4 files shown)


pleerock:
hm, wait. Every method returns Promise, what `async` changes here?

Ginden:
@pleerock Consider this:
```javascript
class Foo {
   constructor() {
       this.qaz = null;
   }

   async foo() {
        return new Promise((res, rej) => this.qaz((err) => err ? rej(err) : resolve()))
   }
   bar() {
        return new Promise((res, rej) => this.qaz((err) => err ? rej(err) : resolve()))
   }
}
const foo = new Foo();
foo.foo(); // Returns rejected promise
foo.bar(); // Throws synchronously
```

Calling `.bar()` will throw synchronously, because `this.qaz` is null. Calling `.foo()` will return rejected promise.

pleerock:
Looks like motivation was wrong? Please specify why you closed this issue without any description.

Ginden:
> Looks like motivation was wrong?

Yes, I was wrong about behaviour of Promise constructor.

PR #3882, opened 2019-03-23
New npm script to run specific test

nicomouss:
Integrated a new `test-fast-single` npm script in package.json enabling developpers to execute specific tests by targeting the directory where the tests are located.

Thus the command below will only execute the tests located within the specified directory (and by extension in all its sub-directories):

`npm run test-fast-single build/compiled/test/functional/schema-builder`

It is just another way of doing than using the  `--grep` option that requires a regular expression to be written.

This method here might be more intuitive than having to write a regular expression to target the subset of tests we want to execute.

### DIFF
--- a/DEVELOPER.md
+++ b/DEVELOPER.md
@@ -140,6 +140,18 @@ All the tests are executed on our Continuous Integration infrastructure and a PR
 describe.only('your describe test', ....)
 ```
 
+>**Hint:** you can use the `test-fast-single` npm script to only execute tests contained within a specific file or directory tree.
+>
+> Example: to only execute tests related to github issue #363:
+>```shell
+>npm run test-fast-single build/compiled/test/github-issues/363
+>```
+>
+> Example: To only execute tests related to schema-builder:
+>```shell
+>npm run test-fast-single build/compiled/test/functional/schema-builder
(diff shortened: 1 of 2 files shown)


pleerock:
Thanks for the contribution!

We are migrating to jest soon, so I don't think this change is necessary.

Kononnable:
You don't have to use complex regex expression - you can just pass a string as grep argument -   `npm run test-fast -- --grep "schema builder"`. Test localized in same directory usually share first part of the test name.

nicomouss:
> You don't have to use complex regex expression - you can just pass a string as grep argument - `npm run test-fast -- --grep "schema builder"`. Test localized in same directory usually share first part of the test name.

Yes, only if "schema builder" is part of two different tests names, then you have unexpected tests that are being executed, although you don't want them to be executed initially. Then you think 'why the heck this other test is popping out', then you have to check the names in order to change your regex to make it target only your test ... That's a lots of thinking/tasks just a execute a test I have put in a folder.

With a folder target, we are sure to target exclusively the specific tests we want. More intuitive. Less cognitive overload.

Anyway both methods work, so everybody can pickup her/his preferred method.

PR #3883, opened 2019-03-23
[Docs] Clarification for save/remove in Repository API

weijunyu:
- Updated Repository API docs
Updated description and examples to reflect that save() and remove() return the entities passed in as arguments.


pleerock:
Its not necessary to add those returns to the example. Save returns same instance of object you send to it.

pleerock:
Thank you for your efforts!

dcworldwide:
I needed an answer to this today. It would be nice if the docs clearly stated that save(x) mutates x. Hence, no need to return anything.

PR #3885, opened 2019-03-23
Add pause/resume stream for mysql driver

Srar:


src/driver/mysql/MysqlQueryRunner.ts
```diff
@@ -187,6 +187,10 @@ export class MysqlQueryRunner extends BaseQueryRunner implements QueryRunner {
                 const stream = databaseConnection.query(query, parameters);
                 if (onEnd) stream.on("end", onEnd);
                 if (onError) stream.on("error", onError);
+                
+                stream.pause = databaseConnection.pause.bind(databaseConnection);
```
pleerock: can you please refer to the docs?

pleerock:
Thank you for contribution. Without tests this functionality can be easily broken, can you please add some tests too?

src/driver/mysql/MysqlQueryRunner.ts
pleerock: no, sorry, I wasn't talking about our docs. I was asking about some link in the code to the documentation of those methods and motivation why we placed this code here.

docs/select-query-builder.md
```diff
@@ -762,6 +762,8 @@ const stream = await getRepository(User)
     .stream();
 ```
 
+You can use `stream.pause()` or `stream.resume()` method to stop/resume stream when retrieving a large result set.
```
pleerock: since this is supported only in mysql we don't need those docs probably

src/driver/mysql/MysqlQueryRunner.ts
Srar: stream is not implemented in mysql driver. only way is pause socket.

[https://github.com/mysqljs/mysql#streaming-query-rows](https://github.com/mysqljs/mysql#streaming-query-rows)

pleerock: Current solution doesn't look proper. Looks like you are resolving issue you have. You should be able to access databaseConnection where you need it different way.

Srar: the problem is the `stream()` method returns stream type.
https://github.com/typeorm/typeorm/blob/master/src/driver/mysql/MysqlQueryRunner.ts#L179
https://github.com/typeorm/typeorm/blob/master/src/platform/PlatformTools.ts#L4

imnotjames:

Issue #3828, opened 2019-03-14
Conflicting pull-requests accepted to fix postgres schema:log with uppercase table names and enums


romain-gilliotte:
**Issue type:**

[ ] question
[x] bug report
[ ] feature request
[ ] documentation issue

**Database system/driver:**

[ ] `cordova`
[ ] `mongodb`
[ ] `mssql`
[ ] `mysql` / `mariadb`
[ ] `oracle`
[x] `postgres`
[ ] `cockroachdb`
[ ] `sqlite`
[ ] `sqljs`
[ ] `react-native`
[ ] `expo`

**TypeORM version:**

[ ] `latest`
[ ] `@next`
[ ] `0.2.15` (or put your version here)

**Steps to reproduce or a small repository showing the problem:**

See https://github.com/typeorm/typeorm/issues/3536 and https://github.com/typeorm/typeorm/issues/3792

**What happened:**

Accepting both pull-request is causing typeorm migration tool to crash again when trying to generate migrations on databases created with any version (<= 0.2.14) that contained:
- uppercase letters in the table names.
- at least one enum in one of those tables.

```
{ QueryFailedError: type "transaction.detail_type_enum" does not exist
    at new QueryFailedError (.../src/error/QueryFailedError.ts:9:9)
    at Query.callback (.../src/driver/postgres/PostgresQueryRunner.ts:176:30)
    at Query.handleError (.../node_modules/pg/lib/query.js:142:17)
    at Connection.connectedErrorMessageHandler (.../node_modules/pg/lib/client.js:160:17)

# The enum is named "*D*etail_type_enum" in my database, because it was created by an older
# version of typeorm
```
When TypeOrm named the enums before this version, it kept uppercase letters from the table name when generating the enum type name, it no longer does.

My pull-request corrected the missing quotes in the driver that were causing the crash when uppercase letters existed in enums names.

@onionhammer pull-request corrected the same problem by keeping bad quoting, but lowercasing all enum names.

Accepting both patches into the 0.2.15 version cause typeorm to **both** lowercase and quote the enum type, which result in the enum type not being found, and crashing again.

Both solutions were OK but don't play well together

romain-gilliotte:
Actually i wrote this issue a bit too fast.
I think that the PR that came with https://github.com/typeorm/typeorm/issues/3536 was only fixing the problem for database created with typeorm >= 0.2.15.

Even without quoting, postgres would not be able to find an enum type created with a former version of typeorm.

[...]

PR #3889, opened 2019-03-24
Add failing test for issue #3828

romain-gilliotte:
This test shows the crash that happens when trying to migrate tables with uppercase name containing at least one enum that were created with typeorm <= 0.2.14

See issue #3828


vlapo:
@Kononnable I think this test will fail every time. Typeorm will not find enum type with different name.

Kononnable:
Yeah, I've messed up a bit. Original issue was fixed so I thought it can be safely merged.
I've added skip for this test to let master pass. I will look into that later today - it looks like just changing the type name in the test doesn't solve the problem.

vlapo:
Yeah sorry. I forgot close this PR after my PR with fix conflict was merge :)

Issue #2191, opened 2018-05-22
TableForeignKeyOptions missing params for deferrable


jkl1337:
**Issue type:**

[ ] question
[ ] bug report
[X] feature request
[ ] documentation issue

**TypeORM version:**

[ ] `latest`
[X] `@next`
[ ] `0.x.x` (or put your version here)

**Steps to reproduce or a small repository showing the problem:**

`TableForeignKeyOptions` is missing params to allow setting deferrable foreign key constraints. Not sure what the best interface for this is. sqlalchemy uses a bool and a free form string/expression for "initially".

pleerock:
What database are you using? TableForeignKeyOptions tries to cover general foreign key functionality that any database has. Looks like deferrable foreign key constraints you are talking about is some database-specific? (I haven't used them). 

If some very-specific functionality is missing you can always call a `query` method and execute any SQL query you would like to.

jkl1337:
Deferrable constraints are feature ID F721 of SQL-99. They are supported by a number of common engines, and also not supported by some notable engines.

https://www.postgresql.org/docs/9.0/static/sql-createtable.html
https://oracle-base.com/articles/8i/constraint-checking-updates
https://www.sqlite.org/foreignkeys.html#fk_deferred
I don't have the time to look at SQL Server, but as of 2012 it did not support them. Mysql unsurprisingly does not support them. Mariadb also does not.

Here is how sqlalchemy addresses it:
http://docs.sqlalchemy.org/en/latest/core/constraints.html#sqlalchemy.schema.ForeignKeyConstraint.__init__

Deferred FK constraints are one of those features that when you need them it is hard not to use them because the alternative is manageable but quite annoying.

I am cognizant of the issues of defining general functionality. But, the lowest common denominator is quite low. MySQL does not support CHECK constraints, Mariadb only does very recently. Both of these are provided explicitly in TypeORM. Similarly for partial indices. So there are arguments either way for this.

https://stackoverflow.com/questions/955671/dealing-with-circular-reference-when-entering-data-in-sql
https://github.com/cockroachdb/cockroach/issues/9897

pleerock:
Okay, got you. We'll investigate how to add this feature to typeorm in the future. Meanwhile, feel free to contribute. For now you'll have to use raw sql queries.

JanPeter:
Is there any update on this?

Kononnable:
PR with feature implementation was merged, will be released by end of the week in `typeorm@0.2.17`

[...]

PR #3904, opened 2019-03-27
Deferrable options for foreign keys (postgres)

jahuuar:
Related feature request: #2191

I've added deferrable options for foreign keys on Postgres, see `DEFERRABLE` on [CREATE TABLE](https://www.postgresql.org/docs/11/sql-createtable.html), or on [SET CONSTRAINTS](https://www.postgresql.org/docs/11/sql-set-constraints.html). Based on that documentation the following declarations are equivalent (at final of column or constraint definition):

- "" == "`NOT DEFERRABLE`"  (default)
- "`DEFERRABLE`" == "`DEFERRABLE INITIALLY IMMEDIATE`"
- "`DEFERRABLE INITIALLY DEFERRED`"

Thus, with a single property on foreign column options all three cases are covered (undefined means "`NOT DEFERRABLE`", default in postgres):
- `deferrable`: "`INITIALLY IMMEDIATE`" | "`INITIALLY DEFERRED`"

This configuration could be extensible to [sqlite](https://www.sqlite.org/foreignkeys.html#fk_deferred) later.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,8 @@ feel free to ask us and community.
 
 ### Features
 
+* added deferrable options for foreign keys (postgres) ([#2191](https://github.com/typeorm/typeorm/issues/2191))
+
 ## 0.2.16 (2019-03-26)
 
 ### Bug fixes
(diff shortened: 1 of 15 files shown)


pleerock:
Thank you! Would be great to have this feature for other databases too.

PR #3910, opened 2019-03-27
Implemented entity factory to create entities

nicomouss:
Following PR #3845 

1) Refactored the code to use an EntityFactory to create the entities
2) Added optional "entityFactory" configuration option: this gives full flexibility to end-user to fine-tune how the entities are created by the ORM (for example he might want to stay with previous behaviour when entities are creating by calling the constructor). Default behaviour stays the one implemented in PR #3845 (skipping constructor).

Also using a custom entity factory, end-user can now use ES6 Proxy API to profile the entities created by the ORM.

See examples implemented as tests in the PR.

### DIFF
--- a/src/connection/BaseConnectionOptions.ts
+++ b/src/connection/BaseConnectionOptions.ts
@@ -3,6 +3,7 @@ import {LoggerOptions} from "../logger/LoggerOptions";
 import {NamingStrategyInterface} from "../naming-strategy/NamingStrategyInterface";
 import {DatabaseType} from "../driver/types/DatabaseType";
 import {Logger} from "../logger/Logger";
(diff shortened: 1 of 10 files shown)


pleerock:
Looks good. Can we have some docs for that too?

nicomouss:
Yes will try to document this soon, should I create another PR for that?

Kononnable:
This PR is merged, so yes - another PR.
I wonder if we should add 'old EntityFactory` and export it from typeorm to make it easier to migrate if someone was using property initialization in models.

nicomouss:
I agree with Old Entity Factory. Should we find another more meaningful name for it?
For now, everything implemented in PR #3962

Issue #1505, opened 2018-01-25
Nesting transactions


eloyleonardo:
**Issue type:**

[x] question
[ ] bug report
[ ] feature request
[ ] documentation issue

**Database system/driver:**

[ ] `cordova`
[ ] `mongodb`
[ ] `mssql`
[ ] `mysql` / `mariadb`
[ ] `oracle`
[x] `postgres`
[ ] `sqlite`
[ ] `sqljs`
[ ] `websql`

**TypeORM version:**

[x] `latest`
[ ] `@next`
[ ] `0.x.x` (or put your version here)

I have a transactional method and from it i call other transactional methods. When i do this the other method create a new transaction, there are any option to create a nested transaction ? Something like the propagation options of [jpa/spring transaction](https://docs.spring.io/spring-framework/docs/current/javadoc-api/org/springframework/transaction/TransactionDefinition.html#PROPAGATION_MANDATORY).

The code above illustrate the question.

**Steps to reproduce or a small repository showing the problem:**

```ts 

@Transaction()
public async save(@RequestBody() band: Band, @TransactionRepository(Band) bandRepository: Repository<Band>) {
        /* new transaction 1 started */ 
        await bandRepository.save(band);
        await this.snooze(30000);
        await  this.saveMoreAwesomeBand(band.name, [band, band, band, band, band, band]);
        /* in this case the transaction 2 was commited but the 1 will be rollbacked */ 
        throw new Error('teste');
       //return await bandRepository.findAll();
    }

@Transaction()
private async saveMoreAwesomeBand(name: string, bands: Array<Band>, @TransactionRepository(Band) bandRepository?: Repository<Band>) {
        /* new transaction 2 started */
        for (let i = 0; i < bands.length; i++) {
            const band = bands[i];
            delete band.id;
            band.name = name + ' ' + i;
            await bandRepository.save(band);
        }
        /* transaction 2 commited */
    }


 ```

ProTip:
I'm *super* interested in this as well. The unit of work experience in entity framework is fantastic, however the technical limitations of the javascript/node seem to make that impossible without a sort of ```AsyncLocal<T>``` storage :|

RDeluxe:
What's the current workaround for this, if any @pleerock  ?

chadjaros:
+1. I would really like to set up nested transactions so that I can do db integration tests without having to re-seed the database for every test.

odavid:
Hi All,
[...]

PR #3921, opened 2019-03-30
Add failing test for issue #1505

clovis1122:
See #1505.

The 1st test case is concerned with calling transactional methods inside another transaction. The 2nd test case (closer to what I need) is related to detect if there's a transaction already running in the current connection. In both test cases, the solution would be that Typeorm somehow detects that the transaction is open and switch over to savepoint/commit savepoint approach.


AlexMesser:
fixed in #8541

PR #3922, opened 2019-03-31
Resolve warning declaration Connection class: Initializers are not allowed in ambient contexts.

Quynh-Nguyen:
I got this warning when using Webpack compile typeorm.

```
node_modules/typeorm/connection/Connection.d.ts (40,28): Initializers are not allowed in ambient contexts.
```
Maybe relate to this issue:
https://github.com/typeorm/typeorm/issues/3823

### DIFF
--- a/src/connection/Connection.ts
+++ b/src/connection/Connection.ts
@@ -62,7 +62,7 @@ export class Connection {
     /**
      * Indicates if connection is initialized or not.
      */
-    readonly isConnected = false;
+    readonly isConnected: boolean;
 
     /**
      * Database driver used by this connection.
@@ -128,6 +128,7 @@ export class Connection {
         this.queryResultCache = options.cache ? new QueryResultCacheFactory(this).create() : undefined;
         this.relationLoader = new RelationLoader(this);
         this.relationIdLoader = new RelationIdLoader(this);
+        this.isConnected = false;
     }
 
     // -------------------------------------------------------------------------


yangukmo:
I hope to merge quickly.

vlapo:
Hey @yangukmo. I think problem is with your TS version. Please check this issue https://github.com/angular/angular-cli/issues/5693 and try use latest version of typescript (or maybe 3.3).

yangukmo:
@vlapo Thank you for reply :)
I use Angular6 and TS 2.9.2.
So, I can't use more than TS3.
Is not this code wrong?

Quynh-Nguyen:
@vlapo I got this problem when implement serverless framework with typescript.
And it's should have definition at constructor, I guess that.
Thank you so much.

vlapo:
Sure I understand a problem. But this is caused by compiling typeorm with `typescript@3.3`. So what is a right solution of problem? 
 - Check whole codebase and remove explicit property initialization everywhere? (Also do not use it in new changes).
 - Is there any flag for `tsc` to remove explicit property initialization from declaration files?
 - Do not care and drop support older version of typescript?

yangukmo:
@vlapo This problem seems to have been caused by the update to the patch version.
Wouldn't it be better to update in minor version than patch versions?

* 0.2.17 - TypeScript reverts for version 2.X.X
* 0.3.X - Change to support only typescript 3.X.X

pleerock:
Its just a single change and Im okay to merge this change since its not principal for a codebase right now. @vlapo do we loose in something if we merge this one? Or do you think authors aren't fixing root of the problem? What do you think about merging it?

vlapo:
I am ok to merge this too. I just think this is not the only problem caused by compiling typeorm with TS3.3. So maybe we will have more issues of this type :)

pleerock:
Okay lets merge this time

Issue #3640, opened 2019-02-14
UPDATE Error when attribut set to NULL


alroux:
**Issue type:**

[ ] question
[x ] bug report
[ ] feature request
[ ] documentation issue

**Database system/driver:**

[ ] `cordova`
[ ] `mongodb`
[ ] `mssql`
[ ] `mysql` / `mariadb`
[ ] `oracle`
[ ] `postgres`
[ ] `sqlite`
[ ] `sqljs`
[ ] `react-native`
[ ] `expo`

**TypeORM version:**

[ ] `latest`
[ ] `@next`
[x ] `0.2.12`

**Steps to reproduce or a small repository showing the problem:**
Hi,

I uses Typeorm in a Nativescript Mobile project, and I have a problem, on Android, setting some data to null if this data was setting to a value before. I use save method which make an INSERT if object does not exist and an UPDATE else. Save method calls  not working UPDATE query  ( see below)

```
  @Column({
    nullable: true
  })
  public maidenName: string | null;
```
```
/ Working
await Patient.getRepository().query(
    "UPDATE patient SET maidenName = NULL WHERE id = ?",
    [1]
  );

// Not Working
await Patient.getRepository().query(
    "UPDATE patient SET maidenName = ? WHERE id = ?",
    [null, 1]
  );
```

I assigned a value (string)  to maidenName , but if then I set maidenName to null and save, I get an error there.
```
QueryFailedError: java.lang.IllegalArgumentException: the bind value at index 1 is null
JS:     android.database.sqlite.SQLiteProgram.bindString(SQLiteProgram.java:164)
JS:     android.database.sqlite.SQLiteProgram.bindAllArgsAsStrings(SQLiteProgram.java:200)
JS:     android.database.sqlite.SQLiteDirectCursorDriver.query(SQLiteDirectCursorDriver.java:49)
JS:     android.database.sqlite.SQLiteDatabase.rawQueryWithFactory(SQLiteDatabase.java:1408)
JS:     android.database.sqlite.SQLiteDatabase.rawQuery(SQLiteDatabase.java:1347)
JS:     com.tns.Runtime.callJSMethodNative(Native Method)
JS:     com.tns.Runtime.dispatchCallJSMethodNative(Runtime.java:1116)
JS:     com.tns.Runtime.callJSMethodImpl(Runtime.java:996)
JS:     com.tns.Runtime.callJSMethod(Runtime.java:983)
[...]

PR #3923, opened 2019-04-01
Fix for #3640: Setting Null values in query rather than passing them as query parameters for AbstractSqliteDriver

velval:
This fixes issues with typeorm when using with NativeScript in Android platform. Underlying sqlite driver for Android requires all query parameter not to be null, so when updating an existing column value with null typeorm should set it directly into the query rather than passing null as one of the query parameters.

https://github.com/typeorm/typeorm/issues/3640

### DIFF
--- a/src/query-builder/UpdateQueryBuilder.ts
+++ b/src/query-builder/UpdateQueryBuilder.ts
@@ -388,6 +388,10 @@ export class UpdateQueryBuilder<Entity> extends QueryBuilder<Entity> implements 
                     // todo: duplication zone
                     if (value instanceof Function) { // support for SQL expressions in update query
                         updateColumnAndValues.push(this.escape(column.databaseName) + " = " + value());
+                    } else if ( (value === undefined || value === null) && this.connection.driver instanceof AbstractSqliteDriver ) {
+                        // If value is null and we are working with SQLite simply set the null value directly into the query instead of passing it on the paramters.
+                        // This is to avoid underlying drivers to complain avout null params. See  https://github.com/typeorm/typeorm/issues/3640
+                        updateColumnAndValues.push(this.escape(column.databaseName) + " = NULL");
                     } else {
                         if (this.connection.driver instanceof SqlServerDriver) {
                             value = this.connection.driver.parametrizeValue(column, value);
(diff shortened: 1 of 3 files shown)


webleaf:
@velval there are [tslint errors](https://travis-ci.org/typeorm/typeorm/jobs/513914057)

velval:
Hey @webleaf , I've fixed those tslint errors and all checks seemed have passed now.

src/query-builder/UpdateQueryBuilder.ts
```diff
@@ -388,6 +388,10 @@ export class UpdateQueryBuilder<Entity> extends QueryBuilder<Entity> implements
                     // todo: duplication zone
                     if (value instanceof Function) { // support for SQL expressions in update query
                         updateColumnAndValues.push(this.escape(column.databaseName) + " = " + value());
+                    } else if ( (value === undefined || value === null) && this.connection.driver instanceof AbstractSqliteDriver ) {
```
pleerock: `undefined !== null`. Why do you check for undefined value here?

src/query-builder/UpdateQueryBuilder.ts
```diff
@@ -434,6 +438,10 @@ export class UpdateQueryBuilder<Entity> extends QueryBuilder<Entity> implements
                 // todo: duplication zone
                 if (value instanceof Function) { // support for SQL expressions in update query
                     updateColumnAndValues.push(this.escape(key) + " = " + value());
+                } else if ( (value === undefined || value === null) && this.connection.driver instanceof AbstractSqliteDriver ) {
```
pleerock: same here

webleaf:
I think we could do `value == null`. It checks for `undefined` too.
But (even in this case) set entity field to `undefined` does not produce `UPDATE` query either (looks like it by design in typeorm)
About the fix as a whole, it seems to work for me.

pleerock:
Im not asking to do `value == null`, Im asking why we need to do more than `value === null`.

> But (even in this case) set entity field to undefined does not produce UPDATE query either (looks like it by design in typeorm)

right, this is why Im asking why this undefined check is needed here

webleaf:
@pleerock please, make as it would be better

jscti:
Can this be merged ?

imnotjames (changes requested):
If this is only affecting native script, why update all SQLite drivers?

imnotjames:
Given there were unanswered questions, last activity over a year ago & this would negatively impact other drivers I'll be closing it.

I'll be closing this PR but if you ever get time to come back to it please feel welcome to open a new PR addressing the concerns.

PR #3943, opened 2019-04-05
View entity

AlexMesser:
implemented View functionality


src/decorator/columns/ViewColumn.ts
```diff
@@ -0,0 +1,21 @@
+import {getMetadataArgsStorage} from "../../";
+import {ColumnMetadataArgs} from "../../metadata-args/ColumnMetadataArgs";
+
+/**
+ * ViewColumn decorator is used to mark a specific class property as a view column.
+ */
+export function ViewColumn(): Function;
```
Kononnable: Why do we need separate definition here?

AlexMesser: not needed, just a mistake

PR #3944, opened 2019-04-05
Fixed issue with adding columns to existing tables in mssql.

mrkmg:
If you try to add a non null column to an existing table, the query will fail on MS-SQL due to a restriction of non-null non-default columns being forbidden when the table already contains rows of data.

In order to work around this, you need to define a default value, but the driver does not add the default value until after the column is created. On MSSQL this causes you to not be able to ever add the column.

This simple patch fixes that.

### DIFF
--- a/src/driver/sqlserver/SqlServerQueryRunner.ts
+++ b/src/driver/sqlserver/SqlServerQueryRunner.ts
@@ -654,7 +654,7 @@ export class SqlServerQueryRunner extends BaseQueryRunner implements QueryRunner
         const upQueries: Query[] = [];
         const downQueries: Query[] = [];
 
-        upQueries.push(new Query(`ALTER TABLE ${this.escapePath(table)} ADD ${this.buildCreateColumnSql(table, column, false, false)}`));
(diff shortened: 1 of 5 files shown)


mrkmg:
Just rebased for newest updates.

mrkmg:
Just added a test to demo why this is needed

mrkmg:
The last commit should pass tests. It failed because MSSQL was not ready when the test started.

mrkmg:
Rebased again on top of master. I will keep this up to date to ensure easiest merge. This patch is crucial for me, and I would have to assume others as well.

pleerock:
I need @AlexMesser review. As usual its scary to me to merge schema-related changes.

mrkmg:
Any update on this, @AlexMesser ?

Fairly breaking change if you use SyncSchema with MSSQL as without this patch it's impossible to add non-null columns to existing tables.

mrkmg:
Just rebased on master.

AlexMesser:
looks good. I think other drivers may have the same issue )

PR #3962, opened 2019-04-08
Entity Factory

nicomouss:
Following PR #3910 

Implemented Old Entity Factory.
Added optional EntityMetadata parameter to createEntity method.
Created documentation.


nicomouss:
@pleerock When you review and merge this PR, would you mind also exporting `EntityFactoryInterface` from typeorm.

I forgot to add this line in `src/index.ts`:

```ts
export {EntityFactoryInterface} from "./entity-factory/EntityFactoryInterface";
```

Kononnable:
The only names I came up with are `EntityFactoryWithInitialization` and `EntityFactoryWithoutInitialization`. They're long etc. but I doubt in year or so we will remember what is the difference between`OldEntityFactory` and `DefaultEntityFactory`. There is also a question what if we decide to change defaults to something new in the future.

test/functional/repository/basic-methods/repository-basic-methods.ts
```diff
@@ -27,6 +27,8 @@ describe("repository > basic methods", () => {
     let connections: Connection[];
     before(async () => connections = await createTestingConnections({
         entities: [Post, Blog, Category, UserEntity, QuestionEntity],
+        entityFactory:  new OldEntityFactory()
```
pleerock: why do we need it now? Are tests failing without old entity factory now?

test/github-issues/3256/issue-3256.ts
```diff
@@ -2,13 +2,15 @@ import "reflect-metadata";
...
 import {Post} from "./entity/Post";
+import { OldEntityFactory } from "../../../src";
 
 describe("github issues > #3256 wrong subscriber methods being called", () => {
 
     let connections: Connection[];
     before(async () => connections = await createTestingConnections({
         entities: [__dirname + "/entity/*{.js,.ts}"],
         subscribers: [__dirname + "/subscriber/*{.js,.ts}"],
+        entityFactory:  new OldEntityFactory()
```
pleerock: why do we need it now? Are those tests failing without old entity factory now?

src/entity-factory/OldEntityFactory.ts
```diff
@@ -0,0 +1,14 @@
+import { EntityFactoryInterface } from './EntityFactoryInterface';
+
+export class OldEntityFactory implements EntityFactoryInterface {
```
pleerock: Alternatively we could just point code example in the changelog how user can do same without this being in the codebase. Im okay with OldEntityFactory name btw since it would be very temporary.

pleerock:
> @pleerock When you review and merge this PR, would you mind also exporting `EntityFactoryInterface` from typeorm.
> 
> I forgot to add this line in `src/index.ts`:
> 
> ```ts
> export {EntityFactoryInterface} from "./entity-factory/EntityFactoryInterface";
> ```

export it if you need it

test/functional/repository/basic-methods/repository-basic-methods.ts
Kononnable: https://github.com/typeorm/typeorm/blob/d3ee978501a4f1b74c99e2fc24ed5d8105919322/test/functional/repository/basic-methods/repository-basic-methods.ts#L165
Yes, we need it, otherwise this line is failing. With new entity  factory fields aren't initialized. We could remove that part of test if we don't want to use OldEntityFactory(if it's just temporary).

test/github-issues/3256/issue-3256.ts
Kononnable: As above - fields are initialized. To make it work without OldEntityFactory we would have to change them to check for undefined values.
https://github.com/typeorm/typeorm/blob/d3ee978501a4f1b74c99e2fc24ed5d8105919322/test/github-issues/3256/issue-3256.ts#L32

nicomouss:
> > @pleerock When you review and merge this PR, would you mind also exporting `EntityFactoryInterface` from typeorm.
> > I forgot to add this line in `src/index.ts`:
> > ```ts
> > export {EntityFactoryInterface} from "./entity-factory/EntityFactoryInterface";
> > ```
> 
> export it if you need it

I don't need it, but some people might it if they want to implement a factory class that implements `EntityFactoryInterface`.

Thanks @Kononnable for exporting that interface.

pleerock:
@Kononnable can we merge it?

Kononnable:
Yes, I think we can merge it.

PR #3975, opened 2019-04-10
Map TYPEORM_CONNECT_STRING to the connectString ConnectionOption

bgfriend0:
In order for Oracle DB to connect over secure networking protocols, I need access to the connectString. I noticed that the connectString is configurable when specified as a JS object and an ormconfig.json file, but not from an environment variable. I mapped the appropriate environment variable to cover that gap.


vlapo:
I dont know if this option is not too oracle-specific. Did you try use `TYPEORM_DRIVER_EXTRA` ? 
```
TYPEORM_DRIVER_EXTRA='{ "connectString": "<connectString>" }'
```

bgfriend0:
I presumed connectString was general across all DBs. Nevertheless, I'll give that a try, thanks.

PR #3977, opened 2019-04-11
fixing tests

Kononnable:
Fixing failing tests on next branch.
Remaining 4 tests require old entity factory, so I will fix them after #3962 is merged(or just in that PR).

### DIFF
--- a/test/github-issues/1576/entity/Category.ts
+++ b/test/github-issues/1576/entity/Category.ts
@@ -4,12 +4,12 @@ import {Column, PrimaryGeneratedColumn, Entity} from "../../../../src";
 export class Category {
 
     @PrimaryGeneratedColumn()
-    id: number|null;
+    id: number;
 
     @Column()
     name: string;
 
     @Column()
-    name2: string|null;
+    name2: string;
 
 }
--- a/test/github-issues/184/issue-184.ts
+++ b/test/github-issues/184/issue-184.ts
@@ -26,7 +26,7 @@ describe("github issues > #184 [Postgres] Single-Inheritance not working with in
         employee.shared = "e";
 
         await employeeRepository.save(employee);
-        await employeeRepository.findOne(1);
+        await employeeRepository.findOne("1");
 
         // let homesitterRepository = connection.getRepository(Homesitter);
         // const homesitter = new Homesitter();
--- a/test/other-issues/sqlite-relations-too-many-parameters/entity/Category.ts
+++ b/test/other-issues/sqlite-relations-too-many-parameters/entity/Category.ts
@@ -5,12 +5,12 @@ import {PrimaryColumn} from "../../../../src/decorator/columns/PrimaryColumn";
 export class Category {
 
     @PrimaryColumn()
-    id: string;
+    id: number;
 
     @Column()
     name: string;
 
-    constructor(id?: string, name?: string) {
+    constructor(id?: number, name?: string) {
         if (id) this.id = id;
         if (name) this.name = name;
     }
--- a/test/other-issues/sqlite-relations-too-many-parameters/entity/Category.ts
+++ b/test/other-issues/sqlite-relations-too-many-parameters/entity/Category.ts
@@ -5,12 +5,12 @@ import {PrimaryColumn} from "../../../../src/decorator/columns/PrimaryColumn";
 export class Category {
 
     @PrimaryColumn()
-    id: number;
+    id: string;
 
     @Column()
     name: string;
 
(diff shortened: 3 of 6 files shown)


test/github-issues/1576/entity/Category.ts
```diff
@@ -4,12 +4,12 @@ import {Column, PrimaryGeneratedColumn, Entity} from "../../../../src";
...
 
     @PrimaryGeneratedColumn()
-    id: number|null;
+    id: number;
 
     @Column()
     name: string;
 
     @Column()
-    name2: string|null;
```
Kononnable: Not sure why `|null` was added here(only in next branch). Reflect metadata thinks it's an object then and we guess wrong column type.

test/github-issues/184/issue-184.ts
```diff
@@ -26,7 +26,7 @@ describe("github issues > #184 [Postgres] Single-Inheritance not working with in
         employee.shared = "e";
 
         await employeeRepository.save(employee);
-        await employeeRepository.findOne(1);
+        await employeeRepository.findOne("1");
```
Kononnable: Id is a string. In the past we passed ids always as parameters(conversion to string was done inside pg library), now we pass them in query content if they're numbers.

test/other-issues/sqlite-relations-too-many-parameters/sqlite-relations-too-many-parameters.ts
```diff
@@ -51,9 +51,8 @@ describe("other issues > sqlite relations too many parameters", () => {
         const postWithCategory = await connection.manager.findOne(Post, 1, {
             relations: ["categories"]
         });
-        console.log(postWithCategory);
 
-        postWithCategory!.categories.length.should.be.equal(2000);
+        postWithCategory!.categories.length.should.be.equal(1200);
```
Kononnable: I have no idea why it was 2000 before, we've got just 1200 items.

test/other-issues/sqlite-relations-too-many-parameters/entity/Category.ts
```diff
@@ -5,14 +5,14 @@ import {PrimaryColumn} from "../../../../src/decorator/columns/PrimaryColumn";
 export class Category {
 
     @PrimaryColumn()
-    id: string;
+    id: number;
```
Kononnable: This one is a bit tricky. If ids are strings they're passed in query parameters, if numbers in query text. 1200 parameters are more than standard sqlite can handle. We should either accept that in some rare cases when PK is string there will be errors on sqlite or remove optimizations on how entities are loaded.
https://github.com/typeorm/typeorm/blob/20ff8c60a50df13517ed13451c0bba1c10d4f1c8/src/query-builder/SelectQueryBuilder.ts#L2071-L2094

We could also escape string ids and pass them as query text, but it would be harder to implement and more error prone.

pleerock: yeah we need to resolve this one. Don't think we should change this test just to make things work... there is a real issue in there.

test/other-issues/sqlite-relations-too-many-parameters/sqlite-relations-too-many-parameters.ts
pleerock: @Kononnable I have some not pushed changes on my machine, let me push them before merging these changes. Please take a look on them

pleerock: weird but test is passing on master? How it is?

Kononnable: Test was introduced in next, it doesn't exist in master. It was introduced during time, when only 20 tests were run(`.only` problem), after removing `.only` tests on CI always failed(and bailed) before that test.

test/other-issues/sqlite-relations-too-many-parameters/entity/Category.ts
Kononnable: Ok, so we probably will have to remove that many-to-many optimization on query builder. I will look into that after you push code you mentioned earlier(if it won't fix the problem ofc.).

Kononnable:
@pleerock I'm gonna merge this one without changes to sqlite-relations-too-many-parameters which kind of disabled this test case. There still will be one test failing on @next, but it's running close to the end so if anything else break we should know it(after examining test logs).

Issue #3953, opened 2019-04-07
cant compile EntitySchema with spatialFeatureType as column option


xxRockOnxx:
**Issue type:**

[ ] question
[ ] bug report
[x] feature request
[x] documentation issue

**Database system/driver:**

[ ] `cordova`
[ ] `mongodb`
[ ] `mssql`
[ ] `mysql` / `mariadb`
[ ] `oracle`
[x] `postgres`
[ ] `cockroachdb`
[ ] `sqlite`
[ ] `sqljs`
[ ] `react-native`
[ ] `expo`

**TypeORM version:**

[x] `latest`
[ ] `@next`
[ ] `0.x.x` (or put your version here)

**Steps to reproduce or a small repository showing the problem:**

Here's how the definition is created

```
new EntitySchema<MyType>({
  name: "MyType",
  columns: {
    location: {
      type: "geometry",
      spatialFeatureType: "Point"
    }
  }
})
```

Error message:

> Object literal may only specify known properties, and 'spatialFeatureType' does not exist in type 'EntitySchemaColumnOptions'.

## Workaround

```
/* tslint:disable:no-submodule-imports */

import { EntitySchema, EntitySchemaColumnOptions } from "typeorm";
import { SpatialColumnOptions } from "typeorm/decorator/options/SpatialColumnOptions";

interface LocationColumn extends EntitySchemaColumnOptions, SpatialColumnOptions {}

const location: LocationColumn = {
  type: "geometry",
  spatialFeatureType: "Point"
};

new EntitySchema<MyType>({
  name: "MyType",
  columns: {
    location
  }
})
```

I'm not sure if this is the proper way to do this or it's just a hack, but can we have this in documentation and/or the proper type?

PR #3990, opened 2019-04-13
fix(3953): add missing spatial options in entity schema column definition

vlapo:
Adding `spatialFeatureType` and `srid` into `EntitySchemaColumnOptions`.

Fix https://github.com/typeorm/typeorm/issues/3953


pleerock:
Thank you! Hope we will have some day all entity schema stuff test covered.

Issue #3973, opened 2019-04-10
`leftJoinAndMapOne` suggested entity definition incorrect (documentation)


jalada:
**Issue type:**

[ ] question
[ ] bug report
[ ] feature request
[x] documentation issue

On https://github.com/typeorm/typeorm/blob/master/docs/select-query-builder.md#joining-and-mapping-functionality the documentation suggests:

```typescript
export class User {    
    /// ...
    profilePhoto: Photo;
}
```

But without any kind of initializer (which would defeat the purpose) this isn't allowed as it can't be undefined. You probably need:

```typescript
export class User {
    /// ...
    profilePhoto: Photo | undefined;
}
```

Unless there's a different recommendation?

vlapo:
IMHO this is matter of strict mode enabled in typescript or not. But if you think it is doc issue feel free to contribute :)

pleerock:
Yes its matter of strict mode and it really depends how you are using typescript. It complicates docs and isn't consistent with other docs we have. Thank you for PR anyway.

PR #3994, opened 2019-04-15
Make the `leftJoinAndMapOne` entity definition satisfy a strict TypeScript config

jalada:
#3973

### DIFF
--- a/docs/select-query-builder.md
+++ b/docs/select-query-builder.md
@@ -681,7 +681,7 @@ Add `profilePhoto` to `User` entity and you can map any data into that property 
 ```typescript
 export class User {    
     /// ...
-    profilePhoto: Photo;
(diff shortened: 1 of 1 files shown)


pleerock:
Yes its matter of strict mode and it really depends how you are using typescript. It complicates docs and isn't consistent with other docs we have. Thank you for PR anyway.

jalada:
Is there information somewhere in the documentation on what TS config has been used to define the docs, and what should be used so that if you _use_ the documentation your code doesn't syntax error?

pleerock:
it will emit error only if strict mode is enabled. If you create a new project with empty tsconfig and start setting up things it won't emit you errors until you add strict flag into tsconfig. As far as I know most of people don't use strict mode

jalada:
OK gotcha, thanks @pleerock.

PR #3998, opened 2019-04-15
use distinct discriminator_value for each child entity saved

### LINKED ISSUE #2253: Passing an array of objects removing property of child-table from object, causing query failure
**Issue type:**

[ ] question
[X] bug report
[ ] feature request
[ ] documentation issue

**Database system/driver:**

[ ] `cordova`
[ ] `mongodb`
[ ] `mssql`
[ ] `mysql` / `mariadb`
[ ] `oracle`
[X] `postgres`
[ ] `sqlite`
[ ] `sqljs`
[ ] `react-native`

**TypeORM version:**

[X] `latest`
[ ] `@next`

Utilizing an extended child-table of an ORM table, required fields for the child end up removed when being passed to the database, whether the child table or parent table name is used in the query.

When passing an array of objects to either `save` or `insert`, the ORM is removing the `anotherColumn` value from the objects within the array, which causes the query to fail due to a missing, but expected parameter.

```ts
import * as orm from "typeorm";

@orm.TableInheritance({ column: "type" })
@orm.Entity()
class User {
	constructor(init?: Partial<User>) {
		Object.assign(this, init);
	}

	@orm.PrimaryGeneratedColumn()
	id: number;

	@orm.Column()
	type: string;
}

@orm.ChildEntity("sub")
class SubUser extends User {
	constructor(init?: Partial<SubUser>) {
		super(init);
		Object.assign(this, init);
	}

	@orm.Column()
	anotherColumn: number;
}
// ...

const SubUsers = orm.getRepository(SubUser);
let subUsers: Partial<SubUser>[] = [/*...*/];

// all of these do the same thing
subUsers = subUsers;
subUsers = subUsers.map(u => SubUsers.create(u));
subUsers = SubUsers.create(subUsers);
subUsers = subUsers.map(u => new SubUser(u));

/*
all throw "QueryFailedError: bind message supplies 9 parameters, but prepared statement "" requires 10"
when subUsers.length === 2,
"supplies 13 but requires 15"
when subUsers.length === 3, etc
No error when subUsers.length === 1
*/
await SubUsers.save(subUsers);
await SubUsers.createQueryBuilder("sub_user")
	.insert().into(SubUser).values(subUsers)
	.execute();
await SubUsers.createQueryBuilder("sub_user")
}```

paulfitz:
Currently, saving multiple child entities together results in failure, due to a name collision between the parameter supplied to specify the entity type (`discriminator_value`).  This commit makes the parameter name unique (`discriminator_value_NNN`) in order to avoid that collision and allow insertion of multiple child entities to succeed.

See #2253 for examples of the problem.  Fixes #2253.

### DIFF
--- a/src/query-builder/InsertQueryBuilder.ts
+++ b/src/query-builder/InsertQueryBuilder.ts
@@ -414,8 +414,8 @@ export class InsertQueryBuilder<Entity> extends QueryBuilder<Entity> {
                     //     expression += subQuery;
 
                     } else if (column.isDiscriminator) {
-                        this.expressionMap.nativeParameters["discriminator_value"] = this.expressionMap.mainAlias!.metadata.discriminatorValue;
-                        expression += this.connection.driver.createParameter("discriminator_value", parametersCount);
+                        this.expressionMap.nativeParameters["discriminator_value_" + parametersCount] = this.expressionMap.mainAlias!.metadata.discriminatorValue;
+                        expression += this.connection.driver.createParameter("discriminator_value_" + parametersCount, parametersCount);
                         parametersCount++;
                         // return "1";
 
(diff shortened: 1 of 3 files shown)


pleerock:
Thank you!

PR #4005, opened 2019-04-16
Prevent errors when synchronizing Sqlite or MSSql DBs with `simple-enum` types

louisremi:
**Issue type:**

[ ] question
[x] bug report
[ ] feature request
[ ] documentation issue

**Database system/driver:**

[ ] `cordova`
[ ] `mongodb`
[ ] `mssql`
[ ] `mysql` / `mariadb`
[ ] `oracle`
[ ] `postgres`
[ ] `cockroachdb`
[x] `sqlite`
[ ] `sqljs`
[ ] `react-native`
[ ] `expo`

**TypeORM version:**

[x] `latest`
[ ] `@next`

**Steps to reproduce:**

I tried to use the new (and **undocumented**) `simple-enum` type with sqlite. I'm surprised to see that Typeorm is able to synchronize only once when the table hasn't been created yet. Once the table has been created, synchronizing throws with the following error message:
```
{ QueryFailedError: SQLITE_ERROR: near "-": syntax error
    at new QueryFailedError (/home/louisremi/Workspace/monbelappart/src/error/QueryFailedError.ts:9:9)
    at handler (/home/louisremi/Workspace/monbelappart/src/driver/sqlite/SqliteQueryRunner.ts:53:26)
    at replacement (/home/louisremi/Workspace/monbelappart/node_modules/sqlite3/lib/trace.js:19:31)
    at Statement.errBack (/home/louisremi/Workspace/monbelappart/node_modules/sqlite3/lib/sqlite3.js:16:21)
  message: 'SQLITE_ERROR: near "-": syntax error',
  errno: 1,
  code: 'SQLITE_ERROR',
  name: 'QueryFailedError',
  query: 'CREATE TABLE "temporary_room" ("id" varchar PRIMARY KEY NOT NULL, "reference" varchar NOT NULL, "name" varchar NOT NULL, "floorArea" integer NOT NULL, "status" simple-enum CHECK( status IN (\'draft\',\'active\',\'maintenance\',\'archived\',\'deleted\') ) NOT NULL DEFAULT (\'draft\'), "createdAt" datetime NOT NULL DEFAULT (datetime(\'now\')), "updatedAt" datetime NOT NULL DEFAULT (datetime(\'now\')), CONSTRAINT "UQ_8c1de74045cb8a30adbb6614f7c" UNIQUE ("reference"))',
  parameters: [] }
```
I am a complete noob with Typeorm, but from poking around the source code, here's what I'm guessing is going wrong:
- at some point [`loadTables`](https://github.com/typeorm/typeorm/blob/63d993b1ef9f68a6e12bd535ac8ca8be39702502/src/driver/sqlite-abstract/AbstractSqliteQueryRunner.ts#L754) is executed to detect a difference between the entity and the actual Table layout.
- for some reason, the person who added simple-enum support wanted `loadTables()` to return `simple-enum` instead of the actual `varchar` type that was used when creating the table in sqlite, see https://github.com/typeorm/typeorm/blob/63d993b1ef9f68a6e12bd535ac8ca8be39702502/src/driver/sqlite-abstract/AbstractSqliteQueryRunner.ts#L818 This code block is troubling because there is no such "inverse conversion" for other `simple-*` types such as `simple-json` and `simple-array`
- because of this, two problems happen: Typeorm incorrectly detects a difference between the entity and the table layout, so it tries to recreate the table… and this time uses `simple-enum` as the column type in the query, which causes the above error.

I removed the "inverse conversion" code blocks in both Sqlite and MSSql query runners, and I removed the tests that were expecting `findColumnByName` to return the entity definition instead of the actual table layout. Again, the reason I did this is because the implementation and tests of `simple-array` and `simple-json` don't have this behavior and assertions… and also because using `simple-enum` only appears to work in sqlite when I remove this code.

I'm willing to add documentation for simple-enum to this PR if you are OK with the proposed changes.

### DIFF
--- a/test/functional/database-schema/column-types/sqlite/column-types-sqlite.ts
+++ b/test/functional/database-schema/column-types/sqlite/column-types-sqlite.ts
@@ -128,11 +128,11 @@ describe("database schema > column types > sqlite", () => {
         table!.findColumnByName("datetime")!.type.should.be.equal("datetime");
         table!.findColumnByName("simpleArray")!.type.should.be.equal("text");
         table!.findColumnByName("simpleJson")!.type.should.be.equal("text");
-        table!.findColumnByName("simpleEnum")!.type.should.be.equal("simple-enum");
+        table!.findColumnByName("simpleEnum")!.type.should.be.equal("varchar");
         table!.findColumnByName("simpleEnum")!.enum![0].should.be.equal("A");
         table!.findColumnByName("simpleEnum")!.enum![1].should.be.equal("B");
         table!.findColumnByName("simpleEnum")!.enum![2].should.be.equal("C");
-        table!.findColumnByName("simpleClassEnum1")!.type.should.be.equal("simple-enum");
+        table!.findColumnByName("simpleClassEnum1")!.type.should.be.equal("varchar");
         table!.findColumnByName("simpleClassEnum1")!.enum![0].should.be.equal("apple");
         table!.findColumnByName("simpleClassEnum1")!.enum![1].should.be.equal("pineapple");
         table!.findColumnByName("simpleClassEnum1")!.enum![2].should.be.equal("banana");
--- a/test/functional/database-schema/column-types/sqlite/column-types-sqlite.ts
+++ b/test/functional/database-schema/column-types/sqlite/column-types-sqlite.ts
@@ -129,13 +129,7 @@ describe("database schema > column types > sqlite", () => {
         table!.findColumnByName("simpleArray")!.type.should.be.equal("text");
         table!.findColumnByName("simpleJson")!.type.should.be.equal("text");
         table!.findColumnByName("simpleEnum")!.type.should.be.equal("varchar");
-        table!.findColumnByName("simpleEnum")!.enum![0].should.be.equal("A");
-        table!.findColumnByName("simpleEnum")!.enum![1].should.be.equal("B");
-        table!.findColumnByName("simpleEnum")!.enum![2].should.be.equal("C");
         table!.findColumnByName("simpleClassEnum1")!.type.should.be.equal("varchar");
-        table!.findColumnByName("simpleClassEnum1")!.enum![0].should.be.equal("apple");
-        table!.findColumnByName("simpleClassEnum1")!.enum![1].should.be.equal("pineapple");
-        table!.findColumnByName("simpleClassEnum1")!.enum![2].should.be.equal("banana");
 
     })));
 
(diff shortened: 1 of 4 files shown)


pleerock:
We can't simply remove exist line of code.. Please ping author of those changes and ask him why he added them here.

louisremi:
Hello @borremosch, you are the author of #3700 and your input here would be very much appreciated :-)

Do you remember the rationale for adding the "inverse-transformation" of "varchar + CHECK()" back to `simple-enum`? This appears to cause errors when synchronizing with SQLite. Such behavior wasn't implemented for `simple-json` and `simple-array`, so I'm assuming you have been overzealous here, and I suggest this behavior should be removed for `simple-enum`. What do you think?

louisremi:
Hello again @borremosch, your input on this issue ↑ is more than welcome :-)

borremosch:
Hey @louisremi, apologies for the delay, I have been quite busy lately. I have taken a quick look and it looks like you might be right. I thought it would be nice if Typeorm would infer certain check statements back to a simple enum so that's probably why I added the code for the inverse transformation, but if this is causing issues then it is probably a good idea to remove that part of the code.

test/functional/database-schema/column-types/sqlite/column-types-sqlite.ts
```diff
@@ -128,14 +128,8 @@ describe("database schema > column types > sqlite", () => {
         table!.findColumnByName("datetime")!.type.should.be.equal("datetime");
         table!.findColumnByName("simpleArray")!.type.should.be.equal("text");
         table!.findColumnByName("simpleJson")!.type.should.be.equal("text");
-        table!.findColumnByName("simpleEnum")!.type.should.be.equal("simple-enum");
-        table!.findColumnByName("simpleEnum")!.enum![0].should.be.equal("A");
```
pleerock: can you tell why did you remove those checks for data equality?

louisremi: It isn't data equality that is checked here. Data equality is still tested [on line 118](https://github.com/typeorm/typeorm/pull/3700/files#diff-ce7c0509b75f61f6f160d566ef049087R118). These tests were written to ensure that `findByColumnName("simpleEnum")` returned the original definition of the `simple-enum`, (instead of the actual `nvarchar` used by Sqlite). This is the behavior I am proposing to fix. So these tests are no longer relevant.

imnotjames (changes requested):
Please add a test that confirms that this change allows the intended behavior to work as expected.  I think it's that when you do a second sync it will work correctly?

And that changes to the enum are still reflected - unless there's already a test for that.

PR #4002, opened 2019-04-16
[docs] Link relations.md to other relations-related doc pages

Muph0:
Each of the listed relation types has its own doc page, so I made the list items link to those pages.

PR #4006, opened 2019-04-17
[doc][zh_CN] add view entity docs and link relations.md to other relations-related doc pages

abingoal:
view entity docs #3943 
link relations.md to other relations-related doc pages #4002


rustamwin:
Thank you!

PR #4009, opened 2019-04-17
feat: Support array of value-transformers

### LINKED ISSUE #4007: Support array of ValueTransformer in column options
**Issue type:**

[ ] question
[ ] bug report
[x] feature request
[ ] documentation issue

**Database system/driver:**

[ ] `cordova`
[ ] `mongodb`
[ ] `mssql`
[ ] `mysql` / `mariadb`
[ ] `oracle`
[x] `postgres`
[ ] `cockroachdb`
[ ] `sqlite`
[ ] `sqljs`
[ ] `react-native`
[ ] `expo`

**TypeORM version:**

[ ] `latest`
[ ] `@next`
[x] `^0.2.14` (or put your version here)

Hi, I'd like to be able to apply multiple ValueTransformers to a field:
```ts
import { ValueTransformer } from "typeorm";

export const lowercase: ValueTransformer = {
    to: (entityValue: string) => { return entityValue.toLocaleLowerCase() },
    from: (databaseValue: string) => { return databaseValue }
}

export const encrypt: ValueTransformer = {
    to: (entityValue: string) => { return crypto.encrypt(entityValue) },
    from: (databaseValue: string) => { return crypto.decrypt(databaseValue) }
}
```
And use them like this:
```ts
  @Column({ unique: true, transformer: [lowercase, encrypt] })
  email: string;
```

Is it on the road map ? 
I'd be more than happy to do a PR if this sounds a good idea to you !😇

Thanks

Samox:
Closes #4007


pleerock:
Thank you for contribution.

We can accept this functionality, but it needs a test coverage and cover all databases.

Samox:
Hello @pleerock. Glad to hear that :) The PR is ready 🤗

src/util/ApplyValueTransformers.ts
```diff
@@ -0,0 +1,21 @@
+import { ValueTransformer } from "../decorator/options/ValueTransformer";
+
+export class ApplyValueTransformers {
+    static innerTransform(transformer: ValueTransformer | ValueTransformer[], databaseValue: any) {
```
vlapo: I am not sure about this methods names `inner`, `outer`. Maybe `toTransform`, `fromTransform` ? It's just my opinion. I am good with this if other says it is ok :)

vlapo:
What about move tests into `typeorm/test/functional/columns/value-transformer/` instead of `issue` folder? 

Also please update changelog :)

vlapo:
Some refactor idea @pleerock @Kononnable @AlexMesser :
I have to say this `transformer` feature is more core feature for me but we use this functions in all `Driver`s. I think we should move this out of driver and call before `preparePersistentValue` and `prepareHydratedValue` methods. Also undefined check move into `ApplyValueTransformers` functions. 

This change will:
 -  encapsulate `transform` feature into core classes like `UpdateQueryBuilder`/`InsertQueryBuilder`/... (or maybe some new abstract `Driver` class?)
 - reduce issues and bugs e.g. on creating new drivers
 - more separate core from driver

Let me know what do you think :)

Kononnable:
As a general rule I agree - everything which is the same in all/almost all drivers should be put elsewhere. 
Maybe after some changes we might extract abstractDriver instead of Driver interface which would keep 'core logic' of all drivers. Specific drivers would be then just 'configuration data'(which features are supported by db engine) and implementation of db-specific features.

However, I think we should suspend bigger refactoring until 0.4.0 - current next branch is alive for a year now and we should focus our effort on releasing it.  After 'migration period' it should allow us to spend a bit less time on maintenance.

src/util/ApplyValueTransformers.ts
Samox: I wasn't very inspired, `inner` and `outer` are not clear ^^. I'll go with your opinion

Samox: `transformFrom` and `transformTo` are even clearer

Samox:
Hi guys,
Thanks for the feedback @vlapo, I took your comments in account and made changes :). 

What about the refactor idea, can we merge as it is ? I could open another PR after this one, but I might need help ^^.

test/functional/columns/value-transformer/value-transformer.ts
```diff
@@ -54,5 +55,17 @@ describe("columns > value-transformer functionality", () => {
...
+    it("should apply three transformers", () => Promise.all(connections.map(async connection => {
+        const userRepository = await connection.getRepository(User);
+        const email = `${connection.name}@JOHN.doe`;
+        const user = new User();
+        user.email = email;
+
+        await userRepository.save(user);
+
+        const dbUser = await userRepository.findOne();
+        dbUser && dbUser.email.should.be.eql(email.toLocaleLowerCase());
```
vlapo: What if there is a bug and only first value transformer (lowercase in this case) is applied ? :) Also maybe add test with empty array - for sure it will not throw errors or something unexpected in future.

Samox: Good thinking, I'll improve the tests

Samox:
@vlapo I added a test with `[lowercase, trim]` to make sure both are applied. And one with empty array.

About the docs, I had it challenged by 2 guys that said they understood 😇.

Samox:
Hi all, can I do anything else for the feature to be merged ? 😬

vlapo:
Just wait to last review by pleerock.

PR #4027, opened 2019-04-22
fixing schema sync of enum columns

Kononnable:
Enabling tests added in #3889. Original issue was resolved some time ago, however tests failed because we had some issues with synchronizing enum types.
Typeorm didn't know how to act when enum type had different name then generated one(name was changed manually/column wasn't created by typeorm in the first place) and schema sync was failing.

### DIFF
--- a/src/driver/postgres/PostgresQueryRunner.ts
+++ b/src/driver/postgres/PostgresQueryRunner.ts
@@ -318,6 +318,7 @@ export class PostgresQueryRunner extends BaseQueryRunner implements QueryRunner 
             .filter(column => column.type === "enum" || column.type === "simple-enum")
             .map(async column => {
                 const hasEnum = await this.hasEnumType(table, column);
+                // TODO: Should also check if values of existing type matches expected ones
                 if (!hasEnum) {
                     upQueries.push(this.createEnumTypeSql(table, column));
                     downQueries.push(this.dropEnumTypeSql(table, column));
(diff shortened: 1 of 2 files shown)


src/driver/postgres/PostgresQueryRunner.ts
```diff
@@ -1993,6 +1996,22 @@ export class PostgresQueryRunner extends BaseQueryRunner implements QueryRunner
         }).join(".");
     }
 
+    protected async getEnumTypeName(table: Table, column: TableColumn) {
```
Kononnable: Postgres type name is used only during schema sync/migration, so I've added a new method instead of adding another property to TableColumn.

I can add JSDoc comment before method definition, but tbh. I'm not much of a fan of that. Comments have a tendency of being forgotten when functionality changes and quickly becomes out of date(and in many cases don't bring much more info than function name - just look at the function above this one).

romain-gilliotte:
Thanks a lot!
This should solve the problem we've been having on our project (as typeorm users), and that motivated my initial PR :)

It's pretty cool to come back from holidays and see that this new PR aiming to fix the issue was submitted

IMHO those issues with naming in pgsql are there because postgresql seems to be the only database where `enum` and `sequences` need to be named.
There are no tests and I did not check, but it is likely that typeorm has the same issues with `sequence` naming than with `enums`

Would it make sense to move the code that is responsible for naming those to `src/naming-strategy/DefaultNamingStrategy.ts` instead of having everything in the driver?

This would allow the code of PostgreSQL driver to be shorter / easier to read as names will be handled the same way as with FKs, indexes, etc... and it would allow users to override the naming strategy if they wish.

pleerock:
We need a review from @AlexMesser

src/driver/postgres/PostgresQueryRunner.ts
```diff
@@ -901,10 +904,8 @@ export class PostgresQueryRunner extends BaseQueryRunner implements QueryRunner
 
             // if primary key have multiple columns, we must recreate it without dropped column
             if (clonedTable.primaryColumns.length > 0) {
-                const pkName = this.connection.namingStrategy.primaryKeyName(clonedTable.name, clonedTable.primaryColumns.map(column => column.name));
```
AlexMesser: can you please explain this change?

Kononnable: Unfortunately I can't.
I have no idea why I've changed it like that and when I look at it right now it just feels wrong. This code changes primary key and has nothing to do with enum type.

I remember working on it(took me a while to realize why `primaryColumns.length` is compared to 0), but I no longer think it's a valid change. I've checked changes after making PR, so at the time I thought change was necessary. I'll remove them since I have no clue why it is there and it looks like it would break functionality instead of fixing it.

Thanks for pointing that out.

src/driver/postgres/PostgresQueryRunner.ts
```diff
@@ -479,13 +480,14 @@ export class PostgresQueryRunner extends BaseQueryRunner implements QueryRunner
         });
 
         // rename ENUM types
-        newTable.columns
+        await Promise.all(newTable.columns
```
pleerock: I think its better to run them in sequence to prevent different side effects

TheNoim:
I am pretty sure this pr introduced this issue: #4350 I prepared a fix. However, I don't have any idea how to write a test for it.

PR #4035, opened 2019-04-23
adds an option to control whether columns are inserted by default

### LINKED ISSUE #85: Column option insert: false, update: false
JPA provides a column option {insert: false, update: false} that allows to exclude the column from insert and update statements, respectively.
It would be nice to add this feature to typeorm : can be really useful with columns updated internally in the database.
Also this feature may be a first step towards columns defined as formulas (ex: subselect)
I can contribute.

By the way, thank for your work on typeorm. Javascript was missing an ORM of this quality !

paulfitz:
This adds an `insert` option, analogous to the existing `select` and `readonly` options, for use in the `Column` decorator.  Just like a column specified with `@Column({select: false})` will not be selected by default in queries, and a column with`@Column({readonly: true;})` will not be `UPDATE`d, a column specified with `@Column({insert: false})` will not be inserted by default in `INSERT`s.

Adds a test to verify that an entity can be created even if an alleged column is not present, if that column is marked `insert: false`.

I believe this fixes #85, and is relevant to (though not the same as) the work in #2150.


pleerock:
Shall we deprecate `readonly` and introduce a `update` with the same functionality to prevent naming confusion?

test/github-issues/85/issue-85.ts
```diff
@@ -0,0 +1,35 @@
...
+    before(async () => connections = await createTestingConnections({
+        entities: [__dirname + "/entity/*{.js,.ts}"],
+        schemaCreate: true,
+        dropSchema: true
+    }));
+    beforeEach(() => reloadTestingDatabases(connections));
+    after(() => closeTestingConnections(connections));
+
+  it("should be able to have a non-inserted Column", () => Promise.all(connections.map(async connection => {
+    const queryRunner = connection.createQueryRunner();
```
pleerock: you **must** release queryRunner or CI will fail sometimes

pleerock: also, not sure Ive got meaning of this test. If you don't want to allow values to be inserted I guess you need to set a `DEFAULT` or don't set anything at all and write some value into entity before calling `save` then assert if value isn't equal to the value you wrote and equal to DEFAULT or null if you didn't set a default.

paulfitz: Thanks for the tip, I've released the `queryRunner`.  I also added a simpler, more intuitive test, and then documented my somewhat weirder one a bit better (I hope).

paulfitz:
> Shall we deprecate `readonly` and introduce a `update` with the same functionality to prevent naming confusion?

I'd be happy to include that in this PR, given that you're in favor of it.  To deprecate something in typeorm, is it sufficient to add a @deprecated tag in the right place, and a note to the ChangeLog?

pleerock:
> I'd be happy to include that in this PR, given that you're in favor of it. To deprecate something in typeorm, is it sufficient to add a @deprecated tag in the right place, and a note to the ChangeLog?

correct. Feel free to do that if you are able to!

paulfitz:
@pleerock I was tracing the current `readonly` flag so I could add `update` instead (slightly non-trivial since the logic is inverted).  I trace the effect of the flag to `ColumnMetaData.isReadonly`, but then I can't find a place where that flag is read.  I tried working back from the unit test in `test/functional/columns/readonly/columns-readonly.ts` but that test looks strange, it seems to expect all columns to change including the one marked `readonly`.  I've found an existing discussion of this at https://github.com/typeorm/typeorm/pull/2749#discussion_r240050076.  In summary, the existing implementation of `readonly` looks broken, so I'm scared about touching it in this PR since it'll be hard to validate the changes.

pleerock:
@paulfitz maybe you just deprecate it and introduce new update: boolean parameter that would do what readonly supposed to do? I won't break any exist functionality, right?

paulfitz:
@pleerock I took a shot at implementing the `update` option, deprecating `readonly`, leaving the existing `readonly` test untouched and adding a new test for `update`+`insert` options.  Let me know how it feels.

CHANGELOG.md
```diff
@@ -17,6 +17,7 @@ feel free to ask us and community.
 * sqlite date hydration is susceptible to corruption ([#3949](https://github.com/typeorm/typeorm/issues/3949))
 * fixed mongodb uniques, support 3 ways to define uniques ([#3986](https://github.com/typeorm/typeorm/pull/3986))
 * fixed mongodb TTL index ([#4044](https://github.com/typeorm/typeorm/pull/4044))
+* deprecate column `readonly` option in favor of `update` and `insert` options ([#4035](https://github.com/typeorm/typeorm/pull/4035))
```
pleerock: since 0.2.17 has been released this should be moved into new section. damn hate it.

pleerock:
Thank you very much for this amazing PR!

CHANGELOG.md
Kononnable: The only thing I can think of which could minimize this is to use some kind of changelog generator, but to make it work we would have to write commit messages in specific format.
We wouldn't have to fore contributors to do so - changing messages of merge commits should be enough. Which means a bit more work when merging PRs(decide if change should be seen in changelog, and what entry should be added there).

pleerock: I guess it will be our job - change title before merging. Thats okay I think... we need to find some good changelog generator solution then.

Issue #3993, opened 2019-04-15
Rename default table used in query caching


DarkInsight85:
**Issue type:**

[X] question
[ ] bug report
[X] feature request
[ ] documentation issue

**Database system/driver:**

[ ] `cordova`
[ ] `mongodb`
[ ] `mssql`
[X] `mysql` / `mariadb`
[ ] `oracle`
[ ] `postgres`
[ ] `cockroachdb`
[ ] `sqlite`
[ ] `sqljs`
[ ] `react-native`
[ ] `expo`

**TypeORM version:**

[ ] `latest`
[ ] `@next`
[X] `0.2.7` (or put your version here)

Hi, for the default table "query-result-cache" used in caching, is there a way to rename this table?

Kononnable:
No, it's not possible right now but it should be easy to make a PR with such enhancement if you need it. 
`src/cache/DbQueryResultCache.ts` is the file where this name is defined. If you want to change it to be customizable there should be another property in connection options through which new name could be passed.

YehezkielSaputra:
@Kononnable Just for your information, I have created some changing based on this issue but I have no access to push my commit and create Pull Request.
Would you please help me solve this issue?
Any of your help would be appreciated.

PR #4040, opened 2019-04-24
Fixing Issue # 3993 - Rename default table used in query caching

YehezkielSaputra:
Please refer to :
https://github.com/typeorm/typeorm/issues/3993


src/connection/BaseConnectionOptions.ts
```diff
@@ -133,6 +133,11 @@ export interface BaseConnectionOptions {
          */
         readonly duration?: number;
 
+        /**
+         * Configurable Table Name or query-result-cache as Default.
+         */
+        readonly cacheTableName?: string;
```
vlapo: We should move this option inside `options` property as this is setting only for "database" type of cache. 

Also `tableName` should be sufficient. It is already inside of `cache` object so `cache` prefix is not necessary.

src/cache/DbQueryResultCache.ts
```diff
@@ -28,7 +28,8 @@ export class DbQueryResultCache implements QueryResultCache {
     constructor(protected connection: Connection) {
 
         const options = <SqlServerConnectionOptions|PostgresConnectionOptions>this.connection.driver.options;
-        this.queryResultCacheTable = this.connection.driver.buildTableName("query-result-cache", options.schema, options.database);
+        const cacheOptions = typeof this.connection.options.cache === "object" ? this.connection.options.cache : {};
+        this.queryResultCacheTable = this.connection.driver.buildTableName(cacheOptions.cacheTableName || "query_result_cache" , options.schema, options.database);
```
vlapo: Previous default table name was "query-result-cache" not "query_result_cache". Please be careful. 

*My personal preference (optional):
Maybe expose this part `cacheOptions.cacheTableName || "query_result_cache"` as separate variable. For me it is more readable in cases long lines.*

src/connection/BaseConnectionOptions.ts
```diff
@@ -133,6 +133,11 @@ export interface BaseConnectionOptions {
          */
         readonly duration?: number;
 
+        /**
+         * Configurable Table Name or query-result-cache as Default.
```
vlapo: Yes I know I am perfectionist but more clear comment is:
```ts
/**
 * Configurable table name for "database" type cache.
 * Default value is "query-result-cache"
 */
```

src/connection/BaseConnectionOptions.ts
```diff
@@ -119,7 +119,15 @@ export interface BaseConnectionOptions {
         /**
          * Used to provide redis connection options.
          */
-        readonly options?: any;
+        readonly options?: any | {
```
pleerock: `options` are just some options to be sent to underlying caching driver, you shouldn't use it for your new option. Move tableName one level up.

YehezkielSaputra: ok noted, thanks

src/cache/DbQueryResultCache.ts
YehezkielSaputra: Hi @vlapo 
Noted with thanks.
I have changed as you requested.
Unfortunately, I couldn't add a new feature in file CHAGELOG.md since I don't have access to add it.
Please lemme know if there is anything else I could.

src/connection/BaseConnectionOptions.ts
```diff
@@ -1,8 +1,8 @@
-import {EntitySchema} from "../entity-schema/EntitySchema";
-import {LoggerOptions} from "../logger/LoggerOptions";
-import {NamingStrategyInterface} from "../naming-strategy/NamingStrategyInterface";
-import {DatabaseType} from "../driver/types/DatabaseType";
-import {Logger} from "../logger/Logger";
+import { EntitySchema } from "../entity-schema/EntitySchema";
+import { LoggerOptions } from "../logger/LoggerOptions";
+import { NamingStrategyInterface } from "../naming-strategy/NamingStrategyInterface";
+import { DatabaseType } from "../driver/types/DatabaseType";
+import { Logger } from "../logger/Logger";
```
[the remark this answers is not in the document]
Kononnable: Some IDEs autoformat code on save. There was a change latter in this file and I guess that's why whole file is formatted.
If we disable whitespace changes in git ui it looks much cleaner.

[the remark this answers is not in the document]
YehezkielSaputra: @Kononnable Yes, you're right. It was autoformatted, but I have reverted it back.
@vlapo Thanks for noticing it.

pleerock:
Thanks guys!

PR #4043, opened 2019-04-25
Add UUID type for primary generated column

seesemichaelj:
#284 ensured TypeORM will support UUID primary generation; this PR adds `"uuid"` to the appropriate type


pleerock:
Thanks

PR #4046, opened 2019-04-25
Add driver mariadb

nurdiansyah:
add driver https://github.com/MariaDB/mariadb-connector-nodejs

### DIFF
--- a/src/driver/mysql/MysqlDriver.ts
+++ b/src/driver/mysql/MysqlDriver.ts
@@ -352,17 +352,32 @@ export class MysqlDriver implements Driver {
 
         if (this.poolCluster) {
             return new Promise<void>((ok, fail) => {
-                this.poolCluster.end((err: any) => err ? fail(err) : ok());
+                const end = this.poolCluster.end((err: any) => err ? fail(err) : ok());
+                if (end instanceof Promise) {
+                    end.then(() => {
+                        this.poolCluster.end();
+                        ok();
+                    }).catch(err => {
+                        fail(err);
+                    });
+                }
                 this.poolCluster = undefined;
             });
         }
         if (this.pool) {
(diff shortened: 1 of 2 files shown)


Kononnable:
Do we really need files like  `src/driver/mariadb/MariaDriver.ts` , `src/driver/mariadb/MariaQueryRunner.ts`? We should avoid this as long as possible.

vlapo:
I agree. We do not need another driver for now. Lets just import `mariadb` package in case mysql/mysql2 is not provided. 

https://github.com/typeorm/typeorm/blob/87121eee5e00f55c69c4bfd1d46dc8c227bf9fc3/src/driver/mysql/MysqlDriver.ts#L789-L810

Will see what mariadb specific features we will implement in future. We can switch it into separate driver anytime. Maybe in some refactor.

**Btw. Please do not format code you are not changing.**

nurdiansyah:
yes agree.. before i try it use package in case mysql package.

different with mysql use callback and mariadb use Promise

nurdiansyah:
done switch to mysql case

pleerock:
Not sure I like what we have now. It feels like a mess now. We use `mysql2` instead of `mysql` because its api 100% (i guess) same as mysql package. Can we tell same about `mariadb` package? If yes then why there is additional code? If no, then those incompatibilities in api make code messy and less maintainable

nurdiansyah:
i think same,  additional code for promise api, used default by `mariadb`.  If not approved not problem

pleerock:
From docs:

> The MariaDB Node.js Connector can use different APIs on the back-end: Promise and Callback.
The default API is Promise API.
> Callback API is provided for compatibility with the mysql and mysql2 APIs.

so, lets use only callback API to prevent confusion in the code since we are using same class for all 3  drivers.

Issue #3949, opened 2019-04-07
sqlite date hydration is susceptible to corruption


jacobg:
**Issue type:**

[ ] question
[X] bug report
[ ] feature request
[ ] documentation issue

**Database system/driver:**

[X] `cordova`
[ ] `mongodb`
[ ] `mssql`
[ ] `mysql` / `mariadb`
[ ] `oracle`
[ ] `postgres`
[ ] `cockroachdb`
[X] `sqlite`
[ ] `sqljs`
[ ] `react-native`
[ ] `expo`

**TypeORM version:**

[X] `latest`
[ ] `@next`
[ ] `0.x.x` (or put your version here)

**Steps to reproduce or a small repository showing the problem:**

The AbstractSqliteDriver.prepareHydratedValue function contains this code:
```
            /**
             * Fix date conversion issue
             *
             * If the format of the date string is "2018-03-14 02:33:33.906", Safari (and iOS WKWebView) will convert it to an invalid date object.
             * We need to modify the date string to "2018-03-14T02:33:33.906Z" and Safari will convert it correctly.
             *
             * ISO 8601
             * https://www.w3.org/TR/NOTE-datetime
             */
            if (value && typeof value === "string") {
                value = value.replace(" ", "T") + "Z";
            }

            value = DateUtils.normalizeHydratedDate(value);
```

This code works fine if the datetime text stored in sqlite is of the format `"2018-03-14 02:33:33.906"`. But sqlite also supports storing datetime strings in ISO 8601 format `"2018-03-14T02:33:33.906Z"`. So if it's already stored in sqlite in ISO 8601 format, then typeorm will change the string to `"2018-03-14T02:33:33.906ZZ"`, i.e., two `Z`s the end. Or alternatively, it would change `"2018-03-14T02:33:33.906+00:00"` to `"2018-03-14T02:33:33.906+00:00Z"`, which is also invalid. The result is that typeorm will call the Date constructor on the invalid date string, and the Date object will also be an invalid date when the entity class is instantiated.

The reason this code normally works fine is that if the database row was saved using typeorm, then typeorm will save the date in the expected format. But if the database row was already created outside typeorm, then it is valid to save a sqlite database row in that ISO 8601 format. So because it's valid sqlite datetime format, typeorm should also support it.

[...]

PR #4050, opened 2019-04-25
sqlite date hydration is susceptible to corruption #3949

jacobg:
`AbstractSqliteDriver:prepareHydratedValue` should check for expected time string format before applying safari fixes.


Kononnable:
I looked briefly through the issue(don't have much time right now) and it looks ok. However it would be great to test performance of it(even outside of normal tests) - regex might be a bit costly. And we should know if it isn't a major performance issue before merging.

jacobg:
Thanks @Kononnable for taking a look. I'm not having a performance issue with it. I think regex can be costly when it's more flexible with greedy captures. But these two regexes are pretty strict, so I think they're relatively quick. Also I think generally sqlite is used client side rather than server side, so the transactional throughput may not need to be as high. If you have any more specific suggestions for evaluation, please let me know.

jacobg:
Thanks @Kononnable

pleerock:
Thanks guys!

PR #4063, opened 2019-04-29
[Postgres] Query runner generates wrong PK name (with custom schema)

XBeg9:
*BUG report:* if you use a custom schema name, you will be not able to drop / rename / change your PK constraint. `typeorm` (latest stable version and `next`) doesn’t include schema name inside `createTableSql` method when it creates PKs, but it includes it inside other methods.
*Question here:*
1) If we will fix it inside `createTableSql`, this will fix an issue for new tables, but not for old ones. (`table.name = this.driver.buildTableName(table.name, _this.driver.options.schema);)`)
2) or we will fix it inside other methods (change / delete / update), so we will not use `schema` name for PK namingStrategy (`this.connection.namingStrategy.primaryKeyName(table.name, columnNames);`)

### DIFF
--- a/test/github-issues/4063/issue-4063.ts
+++ b/test/github-issues/4063/issue-4063.ts
@@ -1,0 +1,42 @@
+import "reflect-metadata";
+import {createTestingConnections, closeTestingConnections, reloadTestingDatabases} from "../../utils/test-utils";
+import {expect} from "chai";
+import {Connection} from "../../../src/connection/Connection";
+
+describe("github issues > #4063 [Postgres] Query runner generates wrong PK name (with custom schema)", () => {
+    let connections: Connection[];
+    before(async () => connections = await createTestingConnections({
+        enabledDrivers: ["postgres"],
+        entities: [__dirname + "/entity/*{.js,.ts}"],
+        schema: "custom",
+        schemaCreate: true
+    }));
+    beforeEach(() => reloadTestingDatabases(connections));
+    after(() => closeTestingConnections(connections));
+
+    it("should generate correct PK name", () => Promise.all(connections.map(async connection => {
+        const queryRunner = connection.createQueryRunner();
+
+        let table = await queryRunner.getTable("post");
+        const textColumn = table!.findColumnByName("id")!;
+        table!.findColumnByName("id")!.isPrimary.should.be.true;
+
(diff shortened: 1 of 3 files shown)


pleerock:
I've asked @AlexMesser and he told its a wrong solution since constraint name is required to be in the key name. @AlexMesser can you propose different solution?

AlexMesser:
table schema is required in constraint names to support multi schema functionality. Constraint names must be unique across all schemas. Based on this, the first decision is correct. I assume that problems with synchronization will not arise. Typeorm will just recreate the changed constraints during synchronization. But it needs more investigations.

XBeg9:
@AlexMesser I agree, that scheme should be included in both cases. I will try to change my code to reflect your suggestions. And fix several tests.

test/github-issues/4063/issue-4063.ts
```diff
@@ -0,0 +1,42 @@
...
+    before(async () => connections = await createTestingConnections({
+        enabledDrivers: ["postgres"],
+        entities: [__dirname + "/entity/*{.js,.ts}"],
+        schema: "custom",
+        schemaCreate: true
+    }));
+    beforeEach(() => reloadTestingDatabases(connections));
+    after(() => closeTestingConnections(connections));
+
+    it("should generate correct PK name", () => Promise.all(connections.map(async connection => {
```
imnotjames, line 17: Can this be made more descriptive of the scenario?

test/github-issues/4063/issue-4063.ts
```diff
@@ -0,0 +1,42 @@
+import "reflect-metadata";
+import {createTestingConnections, closeTestingConnections, reloadTestingDatabases} from "../../utils/test-utils";
+import {expect} from "chai";
+import {Connection} from "../../../src/connection/Connection";
+
+describe("github issues > #4063 [Postgres] Query runner generates wrong PK name (with custom schema)", () => {
```
imnotjames, line 6: Is it possible for this scenario to be placed under `test/functional/query-runner/drop-primary-key.ts`? & the create primary key?

imnotjames (changes requested):
Tests are failing - can you rebase and fix them?

AlexMesser:
I'm closing this as it's failing and hasn't been updated in a few years. Anyway, thank you for contribution.

PR #4079, opened 2019-05-03
Add TYPEORM_MSSQL_ENCRYPT config environment variable

schafferer:
Add option to enable encryption using environment variables when connecting to MSSQL in Azure.


vlapo:
This is mssql specific only. Please use  `TYPEORM_DRIVER_EXTRA` to pass extra options into native driver ?
```
TYPEORM_DRIVER_EXTRA='{ "<key>": "<value>" }'
```

PR #4086, opened 2019-05-05
A document description

Sinkyears:
The document description is problematic.  Line 574 in the document。 entities: [
        __dirname + "/entity/*.ts"
    ],


pleerock:
it depends on how you setup/using typescript. What you suggested is going to work only if you are using something like ts-node. If you just compile typescript into javascript this won't work since runtime is using javascript by default.

PR #4087, opened 2019-05-05
Use proper checkboxes in issue template

HoldYourWaffle:
A small change to the issue template changing the formatting to actual checkboxes

**Before**
[ ] A
[ ] B
[x] C

**After**
- [ ] A
- [ ] B
- [X] C

### DIFF
--- a/ISSUE_TEMPLATE.md
+++ b/ISSUE_TEMPLATE.md
@@ -1,36 +1,36 @@
 **Issue type:**
 
-[ ] question
-[ ] bug report
-[ ] feature request
-[ ] documentation issue
+- [ ] question
+- [ ] bug report
+- [ ] feature request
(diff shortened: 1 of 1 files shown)


Kononnable:
This change was made some time ago and we reverted it because of how github treats checkboxes in first post of an issue.

>thanks for the contribution, but Im going to revert this change since we can't use md checkboxes. The problem is that if you use md checkboxes github will treat it as a task list and for example for "mysql, [x] postgres, oracle" it will treat it as 1/3 tasks are resolved.

_Originally posted by @pleerock in https://github.com/typeorm/typeorm/pull/3139#issuecomment-441350468_

HoldYourWaffle:
Hadn't thought about that, I can understand why that'd be annoying.

PR #4088, opened 2019-05-05
optimizing CI build times

Kononnable:
Removed steps no longer used from CI configs.

Right now circle CI test takes about [3 mins](https://circleci.com/gh/Kononnable/typeorm/18?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link). We could enable mssql there, but it would double the time(so I'm not sure if we need it there). TravisCI still takes it time, but it tests each driver we currently test on CI.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -23,12 +23,12 @@ jobs:
           POSTGRES_PASSWORD: "test"
           POSTGRES_DB: "test"
       - image: circleci/mongo:3.4.18
-      - image: mcr.microsoft.com/mssql/server:2017-GA-ubuntu
-        environment:
-          SA_PASSWORD: "Admin12345"
-          ACCEPT_EULA: "Y"
-      - image: cockroachdb/cockroach-unstable:latest
-        command: start --insecure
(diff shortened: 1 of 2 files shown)


pleerock:
There was a reason why `npm install sqlite3 --build-from-source` was added into travis config. Not sure if we need it right now, but you might want to check git history and find a PR with a conversation on this change.

Regarding to mssql - yes we need it, we need to cover as much databases as we can, because we want to know that all databases functioning without issues.

Kononnable:
> There was a reason why `npm install sqlite3 --build-from-source` was added into travis config. Not sure if we need it right now, but you might want to check git history and find a PR with a conversation on this change.

Yes, we was using bleeding edge sqlite features which wasn't yet available in compiled version. But it was some time ago and since then new version of compiled sqlite was made available.

Adding mssql won't be an issue. Cockroachdb may be. When I enabled it for tests on my fork it didn't work well with CircleCI - it was very slow and errored after some time. However I didn't take a longer look at it so it might(or not) be easy to fix.

PR #4091, opened 2019-05-05
Code coverage

Kononnable:
Right now we've got coverage badge in readme, but it shows coverage status from quite some time ago. I've enabled code coverage only for TravisCI - no need to slow down CircleCI builds since it's not testing all of the drivers.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -23,5 +23,4 @@ script:
   - npx nyc npm test
 
 after_success:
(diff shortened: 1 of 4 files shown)


.travis.yml
```diff
@@ -1,4 +1,3 @@
-sudo: required
```
Kononnable: Not related to main change. TravisCI no longer have two environments, so this line is no longer required.

Issue #4096, opened 2019-05-07
SQLite support for orUpdate


emibloque:
**Issue type:**

[ ] question
[ ] bug report
[x] feature request
[ ] documentation issue

**Database system/driver:**

[ ] `cordova`
[ ] `mongodb`
[ ] `mssql`
[ ] `mysql` / `mariadb`
[ ] `oracle`
[ ] `postgres`
[ ] `cockroachdb`
[x] `sqlite`
[ ] `sqljs`
[ ] `react-native`
[ ] `expo`

**TypeORM version:**

[x] `latest`
[ ] `@next`
[ ] `0.x.x` (or put your version here)

In this PR https://github.com/typeorm/typeorm/pull/3056, support for `ON CONFLICT` was added for MySQL and Postgres when using `orUpdate`, and in this PR https://github.com/typeorm/typeorm/pull/3190 support for Postgres `onConflict` was added for SQLite as they have the same syntax.

We want to have support of `orUpdate` override for SQLite, in order to have compatibility when using MySQL in production and SQLite in tests.

I'm going to create a PR for this as well.

PR #4097, opened 2019-05-07
Support orUpdate for SQLite

emibloque:
#### Sumary

This PR adds support of `QueryBuilder`'s `orUpdate` for SQLite.

#### Issue

This closes issue #4096 

#### Background

In PR #3056, support for `ON CONFLICT` was added for MySQL and Postgres when using `orUpdate`, and in this PR #3190 support for Postgres `onConflict` was added for SQLite as they have the same syntax.

### DIFF
--- a/src/query-builder/InsertQueryBuilder.ts
+++ b/src/query-builder/InsertQueryBuilder.ts
@@ -249,7 +249,7 @@ export class InsertQueryBuilder<Entity> extends QueryBuilder<Entity> {
       if (statement && statement.overwrite instanceof Array) {
         if (this.connection.driver instanceof MysqlDriver) {
           this.expressionMap.onUpdate.overwrite = statement.overwrite.map(column => `${column} = VALUES(${column})`).join(", ");
-        } else if (this.connection.driver instanceof PostgresDriver) {
+        } else if (this.connection.driver instanceof PostgresDriver || this.connection.driver instanceof AbstractSqliteDriver) {
           this.expressionMap.onUpdate.overwrite = statement.overwrite.map(column => `${column} = EXCLUDED.${column}`).join(", ");
         }
       }
--- a/test/github-issues/4096/issue-4096.ts
+++ b/test/github-issues/4096/issue-4096.ts
@@ -1,0 +1,61 @@
+import "reflect-metadata";
+import {createTestingConnections, closeTestingConnections, reloadTestingDatabases} from "../../utils/test-utils";
+import {Connection} from "../../../src/connection/Connection";
+import {expect} from "chai";
+import {User} from "./entity/User";
+
+describe("github issues > #4096 SQLite support for orUpdate", () => {
+    let connections: Connection[];
+
+    before(async () => connections = await createTestingConnections({
+        entities: [User],
+        enabledDrivers: ["sqlite"],
+        schemaCreate: true,
+        dropSchema: true,
+    }));
+
+    beforeEach(() => reloadTestingDatabases(connections));
+
+    after(() => closeTestingConnections(connections));
+
+    const user1 = new User();
+    user1.email = "example@example.org";
+    user1.username = "example";
+    user1.bio = "My bio";
+
+    const user2 = new User();
+    user2.email = "example@example.org";
+    user2.username = "example";
+    user2.bio = "Updated bio";
+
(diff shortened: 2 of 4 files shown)


test/github-issues/4096/issue-4096.ts
```diff
@@ -0,0 +1,61 @@
...
+            overwrite: ["bio"],
+          })
+          .execute();
+
+        const users = await UserRepository.find();
+        expect(users).not.to.be.undefined;
+        expect(users).to.have.lengthOf(1);
+        expect(users[0]).to.includes({ bio: "Updated bio" });
+      } catch (err) {
+        throw new Error(err);
```
Kononnable: Catching errors just to throw them anyway isn't the best practice. Why not just remove whole try ... catch?

test/github-issues/4096/issue-4096.ts
```diff
@@ -0,0 +1,61 @@
...
+        enabledDrivers: ["sqlite"],
+        schemaCreate: true,
+        dropSchema: true,
+    }));
+
+    beforeEach(() => reloadTestingDatabases(connections));
+
+    after(() => closeTestingConnections(connections));
+
+    const user1 = new User();
```
Kononnable: Object isn't shared between tests in any way, so we could put it in `it()` block.

emibloque: Moved 👌

test/github-issues/4096/issue-4096.ts
emibloque: Totally agree.
I just copied the structure in this tests https://github.com/typeorm/typeorm/pull/3056/files#diff-44f5f56ba7aa275b0e0ed9f63a771c3e just in case I was missing something.

pleerock:
> We want to have support of orUpdate override for SQLite, in order to have compatibility when using MySQL in production and SQLite in tests.

do you want to say that SQLite doesn't actually support onUpdate?

test/github-issues/4096/issue-4096.ts
```diff
@@ -0,0 +1,57 @@
+import "reflect-metadata";
+import {createTestingConnections, closeTestingConnections, reloadTestingDatabases} from "../../utils/test-utils";
+import {Connection} from "../../../src/connection/Connection";
+import {expect} from "chai";
+import {User} from "./entity/User";
+
+describe("github issues > #4096 SQLite support for orUpdate", () => {
```
pleerock: onUpdate

src/query-builder/InsertQueryBuilder.ts
```diff
@@ -249,7 +249,7 @@ export class InsertQueryBuilder<Entity> extends QueryBuilder<Entity> {
       if (statement && statement.overwrite instanceof Array) {
         if (this.connection.driver instanceof MysqlDriver) {
           this.expressionMap.onUpdate.overwrite = statement.overwrite.map(column => `${column} = VALUES(${column})`).join(", ");
-        } else if (this.connection.driver instanceof PostgresDriver) {
+        } else if (this.connection.driver instanceof PostgresDriver || this.connection.driver instanceof AbstractSqliteDriver) {
           this.expressionMap.onUpdate.overwrite = statement.overwrite.map(column => `${column} = EXCLUDED.${column}`).join(", ");
```
pleerock: hm so confusing, is it onUpdate or orUpdate? Looks like second, not sure who named it onUpdate before then.

emibloque:
@pleerock I want to mean SQLite doesn't seem to support Typeorm's `orUpdate`.

src/query-builder/InsertQueryBuilder.ts
emibloque: I could take advantage of this PR and change the key within `expressionMap` to `orUpdate` if you see it ok.

pleerock: we can left it as it is right now, and change it later (do it if you want), I just commented on what confused me...

emibloque:
I'm not sure if the error in CI is related with this PR (the last commit was only to solve the conflict in the Changelog file).
Is there anything I can do?

pleerock:
@emibloque its not your fault sometimes it hangs.

pleerock:
Looks like okay to merge it @Kononnable

Issue #4104, opened 2019-05-08
Support sql.js v1.0


michaelbromley:
**Issue type:**

[ ] question
[ ] bug report
[x] feature request
[ ] documentation issue

**Database system/driver:**

[ ] `cordova`
[ ] `mongodb`
[ ] `mssql`
[ ] `mysql` / `mariadb`
[ ] `oracle`
[ ] `postgres`
[ ] `cockroachdb`
[ ] `sqlite`
[x] `sqljs`
[ ] `react-native`
[ ] `expo`

**TypeORM version:**

[x] `latest`
[ ] `@next`
[ ] `0.x.x` (or put your version here)

Sql.js recently released v1.0, now compiled to WebAssembly. It has a new API for creating a connection, see [docs on upgrading](https://github.com/kripken/sql.js/tree/fbe8427c32d5077be06bb808283c2efb22a84c97#upgrading-from-0x-to-1x)

Supporting the new API should only involve a rather minor change to [`SqljsDriver.createDatabaseConnectionWithImport()`](https://github.com/typeorm/typeorm/blob/1917bff9c82568fc91ae0b9be8f83213f7ff4c9f/src/driver/sqljs/SqljsDriver.ts#L248) which I will attempt in a PR to follow.

PR #4105, opened 2019-05-08
Support sql.js v1.0

michaelbromley:
Fixes https://github.com/typeorm/typeorm/issues/4104

Updated the devDependency to sql.js ^1.0.0, all sql.js tests passing.

Also tested against sql.js ^0.5.0 and all tests still pass.


pleerock:
Thank you!

PR #4109, opened 2019-05-08
feat: add support for DISTINCT queries

### LINKED ISSUE #2641: DISTINCT in query builder

Benjamin-Dobell:
# Description

A bare-bones `DISTINCT` implementation that just supports standard SQL row-level `SELECT DISTINCT`, nothing fancy like PSQL's `DISTINCT ON` specific columns.

I've added support to the `QueryBuilder`, and also to `Repository#find` (via `JoinOptions`).

# Relevant Issues

Closes #2641 - _However_, there are users who have posted in that issue wanting two separate things `DISTINCT` (i.e. this PR) and `DISTINCT ON`, which is a (very useful) non-standard SQL extension that is unfortunately not covered by this PR. If this merges, I'd suggest opening another issue specifically to track `DISTINCT ON`. 

# Use Cases

The use-cases for this functionality in `QueryBuilder` is immediately obvious. However, the benefit to `Repository#find` probably warrants an example. 

e.g. A forum (User + Post models) and you want to get all users who have at least 1 post, but not load in their posts. This can be achieved by  `leftJoin` _without_ selecting, we specify `distinct: true` to ensure we don't load into memory duplicate rows (one for each user's post):

```typescript
const users = getRepository<User>(User).find({
	join: {
		alias: 'user',
		distinct: true,
		leftJoin: {
			post: 'user.posts',
		}
	},
	where: "user.id = post.user_id",
})
```

### DIFF
--- a/test/functional/query-builder/select/query-builder-select.ts
+++ b/test/functional/query-builder/select/query-builder-select.ts
@@ -27,6 +27,21 @@ describe("query builder > select", () => {
             "FROM post post");
     })));
 
+    it("should append all entity mapped columns from main selection to SELECT DISTINCT statement", () => Promise.all(connections.map(async connection => {
+        const sql = connection.manager.createQueryBuilder(Post, "post")
+            .distinct()
+            .disableEscaping()
+            .getSql();
+
+        expect(sql).to.equal("SELECT DISTINCT post.id AS post_id, " +
+            "post.title AS post_title, " +
+            "post.description AS post_description, " +
+            "post.rating AS post_rating, " +
+            "post.version AS post_version, " +
+            "post.categoryId AS post_categoryId " +
+            "FROM post post");
+    })));
+
     it("should append all entity mapped columns from both main selection and join selections to select statement", () => Promise.all(connections.map(async connection => {
         const sql = connection.createQueryBuilder(Post, "post")
             .leftJoinAndSelect("category", "category")
--- a/src/query-builder/QueryExpressionMap.ts
+++ b/src/query-builder/QueryExpressionMap.ts
@@ -47,6 +47,11 @@ export class QueryExpressionMap {
     selects: SelectQuery[] = [];
 
     /**
+     * Whether SELECT is DISTINCT.
+     */
+    selectDistinct: boolean = false;
+
+    /**
      * FROM-s to be selected.
      */
     // froms: { target: string, alias: string }[] = [];
--- a/src/query-builder/SelectQueryBuilder.ts
+++ b/src/query-builder/SelectQueryBuilder.ts
@@ -159,6 +159,14 @@ export class SelectQueryBuilder<Entity> extends QueryBuilder<Entity> implements 
     }
 
     /**
+     * Sets whether the selection is DISTINCT.
+     */
+    distinct(distinct: boolean = true): this {
+        this.expressionMap.selectDistinct = distinct;
+        return this;
+    }
+
+    /**
      * Specifies FROM which entity's table select/update/delete will be executed.
      * Also sets a main string alias of the selection data.
      * Removes all previously set from-s.
@@ -1400,8 +1408,9 @@ export class SelectQueryBuilder<Entity> extends QueryBuilder<Entity> implements 
 
                 return this.getTableName(alias.tablePath!) + " " + this.escape(alias.name);
             });
+        const select = "SELECT " + (this.expressionMap.selectDistinct ? "DISTINCT " : "");
         const selection = allSelects.map(select => select.selection + (select.aliasName ? " AS " + this.escape(select.aliasName) : "")).join(", ");
-        return "SELECT " + selection + " FROM " + froms.join(", ") + lock;
+        return select + selection + " FROM " + froms.join(", ") + lock;
     }
 
     /**


pleerock:
There is no `join` option in `next` so I would like to ask you not to change that part. You'll be able to do distinct only using select query builder.

Benjamin-Dobell:
> There is no `join` option in `next`

Ah, a bit silly of me, but I didn't even think to look at the `next` branch.

It might be worth updating the [contributing doc](https://github.com/typeorm/typeorm/blob/master/CONTRIBUTING.md) to mention PRs should be against `next` instead of `master`.

> You'll be able to do distinct only using select query builder.

Is there a issue/discussion around this I can refer to in order to get a grasp on the motivation for the change?

Based on the way TypeORM is _presently_ designed, being able to specify joins (an in particular relationship aliases) with the "entity manger" is critical functionality for me at the moment. If there's an alternative then that's fine, but needing to fallback to the `QueryBuilder` would _significantly_ detract from the usefulness of TypeORM as an ORM.

Specifically because relations will be split across tables, we regularly need a way to perform queries based on these relations. The example in my post above is pretty straight-forward i.e. "We want all users with at least one post".

MongoDB has no problem performing this queries on relations because they can be embedded e.g. profile pics can be embedded in users, so it's trivial to query users who have a profile pic. 

Without being able to reference reference relations (alias) in a `find()` operation then relational databases will be _significantly_ handicapped compared to what TypeORM offers document-oriented DBs (MongoDB).

If there's some alternative to the `join` syntax, then that's fine, but not being able to perform queries based on relations would be a deal breaker for us.

Benjamin-Dobell:
Haven't had a chance to try out `next`, but I see the change log mentions `QueryBuilder` will now load eager relationships - if that might be enough to address my concerns.

I'll rebase this on `next` (minus the `JoinOptions` stuff) when I get a chance.

pleerock:
> Ah, a bit silly of me, but I didn't even think to look at the next branch.
> It might be worth updating the contributing doc to mention PRs should be against next instead of master.

its not required for you to know about `next` and create PRs against it. Until there is a change I notice that is important on next, and I let you know about it.

>  Needing to fallback to the QueryBuilder would significantly detract from the usefulness of TypeORM as an ORM.

its a wrong assumption. `find` is designed to make a simple queries and `QueryBuilder` is designed to make more complex queries. If you don't like QueryBuilder I would like to suggest you to like it because to feel all benefits of it.

>  "We want all users with at least one post".

how do you do it with `find` right now? In `next` where expressions automatically create inner joins, e.g.

```ts
find(User, {
  where: {
      post: {
           id: 1
      }
  }
})
```

will create `SELECT * FROM user INNER JOIN post WHERE post.id=1`

Benjamin-Dobell:
>> "We want all users with at least one post".

> how do you do it with find right now? In next where expressions automatically create inner joins, e.g.
> [...]
> `SELECT * FROM user INNER JOIN post WHERE post.id=1`

That's not the correct query, we're after:

> `SELECT DISTINCT * FROM user INNER JOIN post ON post.user_id=user.id`

The example code in my initial post is actually performing this same query, albeit in an overly verbose fashion i.e.

```typescript
const users = getRepository<User>(User).find({
	join: {
		alias: 'user',
		distinct: true,
		leftJoin: {
			post: 'user.posts',
		}
	},
	where: "user.id = post.user_id",
})
```

This outputs the query:

> SELECT DISTINCT \<USER FIELD SELECTIONS\> FROM "users" "user" LEFT JOIN "posts" "post" ON "post"."user_id"="user"."id" WHERE "user"."id" = post.user_id

`LEFT JOIN` is just used for illustration purposes; an `INNER JOIN` (`innerJoin` instead of `leftJoin`) would of course eliminate the need for the `WHERE` clause entirely.  


**Anyway...**

The changes you've already made to `QueryBuilder` on `next` (e.g. enabled eager loading), actually address my concerns with using it. Basically, I just didn't like the idea that using `QueryBuilder` would cause me to miss out on some ORM benefits. For example, in the future if I wanted to change a model to enable eager loading, I wouldn't then want to have to go update every single `QueryBuilder` query simply because it didn't support eager loading.

I'm actually very comfortable using `QueryBuilder`, it's a far superior API to `find()`. It _was_ just missing functionality (on `master`) that `find()` supported that concerned.

Now that's no longer the case I actually see very little reason for `find()` to exist at all; if it were deprecated I wouldn't miss it.

hoangnguyen1247:
Merge this, please

fbrbovic:
Any chance of merging this any time soon?

Benjamin-Dobell:
> There is no `join` option in `next` so I would like to ask you not to change that part. You'll be able to do distinct only using select query builder.

@pleerock As requested, I've dropped the changes to `find`.

I've also rebased on the `HEAD` of `master`. If you'd prefer me to target `next`, please let me know.

test/functional/query-builder/select/query-builder-select.ts
```diff
@@ -27,6 +27,21 @@ describe("query builder > select", () => {
             "FROM post post");
     })));
 
+    it("should append all entity mapped columns from main selection to SELECT DISTINCT statement", () => Promise.all(connections.map(async connection => {
+        const sql = connection.manager.createQueryBuilder(Post, "post")
+            .distinct()
+            .disableEscaping()
+            .getSql();
+
+        expect(sql).to.equal("SELECT DISTINCT post.id AS post_id, " +
```
Kononnable: In general we should try to not compare query text in our tests, but the behavior.  If for some reason we would change default naming strategy in the future we would have to modify many tests. However this(wrong) approach is used in this file, so I'm ok with it here.

CHANGELOG.md
```diff
@@ -24,6 +24,7 @@ feel free to ask us and community.
 * extend afterLoad() subscriber interface to take LoadEvent ([issue #4185](https://github.com/typeorm/typeorm/issues/4185))
 * relation decorators (e.g. `@OneToMany`) now also accept `string` instead of `typeFunction`, which prevents circular dependency issues in the frontend/browser ([issue #4190](https://github.com/typeorm/typeorm/issues/4190))
 * added support for metadata reflection in typeorm-class-transformer-shim.js ([issue #4219](https://github.com/typeorm/typeorm/issues/4219))
+* added `SelectQueryBuilder#distinct` for building distinct (`SELECT DISTINCT`) queries.
```
Kononnable: We could remove this, since we're now using auto generating changelog entries from commit messages.

Benjamin-Dobell: Done. Also added `feat: ` to the commit message so it ought to auto-detect.

pleerock:
thank you!

PachowStudios:
Any ETA for when this will be released? This is critical for us :)

PR #4110, opened 2019-05-08
Allow find() to accept more complex where clauses

Benjamin-Dobell:
# Description

I've introduced a new class called `WhereFactory` whose implementation is identical to that of the existing `Brackets`, it only varies in semantics i.e. `QueryBuilder` won't insert brackets around a `WHERE` clause generated by a `WhereFactory`.

`Brackets` now inherits from `WhereFactory` and defines no additional functionality of its own, it's purely semantic i.e. `QueryBuilder` _will_ put brackets around a `WHERE` clause generated by `Brackets`.

Any APIs that took `Brackets` now take `WhereFactory`, meaning they'll continue to accept `Brackets` in addition to `WhereFactory`, so the API changes are non-breaking / backward compatible.

# Motivation

Simply put, more elaborate `WHERE` clauses returning entities - not just for type safety, but also to leverage functionality like eager loading.

With this PR we now support (to my knowledge) _all_ possible `WHERE` conditions when using `find`; including brackets, sub-queries, combinations of string and object literal (property) conditions etc.

e.g. Here we load user entities, eager load in their posts, but restrict to:

* User's that _have posts_ (no users with zero posts).
* If the _current user_ isn't an admin, we chain (AND) in the condition:
   - Only consider public posts, OR,
   - Posts made by friends.

```typescript
function findAccessibleUserPosts(user : User): Promise<User[]> {
	return getRepository<User>(User).find({
		join: {
			alias: 'user',
			leftJoinAndSelect: {
				post: 'user.posts',
			}
		},
		where: new WhereFactory((qb: WhereExpression) => {
			qb.where("user.id = post.user_id")

			if (user.role !== Role.admin) {
				qb.andWhere(new Brackets(qb => {
						qb.where("post.public = true"})
							.orWhereInIds(user.friendIds)
					}
				))

			}
		}),
	})
}
```

which would spit out something _like_ the following:

`WHERE "user"."id" = post.user_id AND (post.public = true OR "user"."id" IN (?, ?, ?))`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ feel free to ask us and community.
 ### Features
 
 * deprecate column `readonly` option in favor of `update` and `insert` options ([#4035](https://github.com/typeorm/typeorm/pull/4035))
+* added `WhereFactory` which can be passed to `Repository#find` to facilitate constructing more complex queries.
 
 ## 0.2.17 (2019-05-01)
 
--- a/src/find-options/FindOneOptions.ts
+++ b/src/find-options/FindOneOptions.ts
@@ -1,6 +1,7 @@
 import {JoinOptions} from "./JoinOptions";
 import {ObjectLiteral} from "../common/ObjectLiteral";
 import {FindConditions} from "./FindConditions";
+import {WhereFactory} from "../query-builder/WhereFactory";
 
 /**
  * Defines a special criteria to find specific entity.
@@ -13,9 +14,9 @@ export interface FindOneOptions<Entity = any> {
     select?: (keyof Entity)[];
 
     /**
-     * Simple condition that should be applied to match entities.
+     * Where condition that should be applied to match entities.
      */
-    where?: FindConditions<Entity>[]|FindConditions<Entity>|ObjectLiteral|string;
+    where?: FindConditions<Entity>[]|FindConditions<Entity>|ObjectLiteral|string|WhereFactory;
 
     /**
      * Indicates what relations of entity should be loaded (simplified left join form).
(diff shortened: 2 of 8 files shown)


Benjamin-Dobell:
Just force pushed away a white-space only change, no functional changes made.

pleerock:
Thanks for contribution!

For some reason I don't like naming of `WhereFactory`. Can we came up with better name. Maybe `RawWhere` ? `WhereExpression` looks good alternative but looks like its already busy. Now Im thinking that we should have called what is called `WhereExpression` now with a better name.

Also, I think I would have issues when I merge this into `next` branch. Can you make same changes on next branch too?

Benjamin-Dobell:
I think https://github.com/typeorm/typeorm/commit/abd4212f0c5207b3829e0a01b820058f1c9bad0f might actually make this PR mostly irrelevant.

This PR essentially allows you to construct any SQL query you want using `find()`. However, [`find()` has had functionality removed on `next`](https://github.com/typeorm/typeorm/pull/4109#issuecomment-490874943) and [`QueryBuilder` has had functionality enhanced](https://github.com/typeorm/typeorm/blame/next/CHANGELOG.md#L11).

With the decreased functionality offered by `find()` I don't believe offering additional `WHERE` flexibility makes all that much sense. Given `QueryBuilder` will be able to do everything `find()` can do (i.e. eager load relationships) and more, then it seems like the way forward.

Does my understanding seem correct? If so, I'll happily close this PR.

pleerock:
Yes, maybe you are right. Probably its a good idea left what you wanted to do on query builder itself.

PR #4114, opened 2019-05-09
fix: handle hydration of uncommon datetime formats in sqlite

louisremi:
Sequelize uses an uncommon format for datetime in Sqlite: `'YYYY-MM-DD HH:mm:ss.SSS Z'` (see [dataTypes.js](https://github.com/sequelize/sequelize/blob/7445423b9effc212e125d76d58e02713e808ffc6/lib/data-types.js#L467)). It's uncommon because it has a space instead of a T between the date and the time, and a space is inserted before the timezone. The change I am proposing allows properly hydrating those datetimes, instead of returning `Date { Invalid }` objects.

I also suggest removing the [comment block on line 280](https://github.com/monbelappart/typeorm/blob/fcfa5b91378005bada8b81d881931d7d6407c41f/src/driver/sqlite-abstract/AbstractSqliteDriver.ts#L280) as its content is mostly redundant with the comment section on line 290. Let me know if you agree and I will push that change as well.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ feel free to ask us and community.
 
 ## 0.2.18 (UNRELEASED)
 
+### Bug fixes
+
+* fixed Sqlite hydration of uncommon datetime formats ([#4114](https://github.com/typeorm/typeorm/pull/4114))
+
 ### Features
 
 * deprecate column `readonly` option in favor of `update` and `insert` options ([#4035](https://github.com/typeorm/typeorm/pull/4035))
--- a/src/driver/sqlite-abstract/AbstractSqliteDriver.ts
+++ b/src/driver/sqlite-abstract/AbstractSqliteDriver.ts
@@ -289,10 +289,15 @@ export abstract class AbstractSqliteDriver implements Driver {
             if (value && typeof value === "string") {
                 // There are various valid time string formats a sqlite time string might have:
                 // https://www.sqlite.org/lang_datefunc.html
-                // There are two separate fixes we may need to do:
-                //   1) Add 'T' separator if space is used instead
-                //   2) Add 'Z' UTC suffix if no timezone or offset specified
-
+                // There are three separate fixes we may need to do:
+                //   1) Remove spaces between the time and the timezone if any
+                //      (Sequelize serializes dates this way)
+                //   2) Add 'T' separator if space is used instead so that Safari
+                //      (and iOS WKWebView) is able to parse it
+                //   3) Add 'Z' UTC suffix if no timezone or offset specified
+                if (/ (Z|[+-]\d\d:\d\d)$/.test(value)) {
+                    value = value.replace(/ +(Z|[+-]\d\d:\d\d)$/, "$1");
+                }
                 if (/^\d\d\d\d-\d\d-\d\d \d\d:\d\d/.test(value)) {
                     value = value.replace(" ", "T");
                 }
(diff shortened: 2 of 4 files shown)


pleerock:
can you tell me why do we need to handle this uncommon case in our code?

louisremi:
@pleerock The documentation has a section titled "Migration from Sequelize to TypeORM", so I assume I am not the first one to make the transition. I might be one of the first to do it for a project based one Sqlite, though ;-)

I'm just realizing that the three distinct fixes could probably be converted into a single regexp based replace:
```js
// normalize the different datetime formats which might have been used in an existing database
// see https://www.sqlite.org/lang_datefunc.html
value = value.replace(
    /^(\d\d\d\d-\d\d-\d\d)[ T](\d\d:\d\d(?::\d\d(?:\.\d\d\d)?)?) ?(Z|[+-]\d\d:\d\d)?$/,
    (match, date, time, timezone) => `${date}T${time}${timezone || "Z"}`
);
```
This would greatly simplify the code and comments.

pleerock:
Documentation says "Migration from Sequelize to TypeORM" to help people to understand some basic differences we have with sequelize. It doesn't mean we want to do everything same as in Sequelize to make this migration.

So, I would like to ask you why exactly do we need this change? Why Sequelize adopted this date format if its not a standard one? For example it might be something that left for legacy reasons or something that was needed to specific set of people, we don't really need side-effect solutions in our codebase. Please let us know motivation behind this change, why majority of typeorm users will need it?

Kononnable:
We've changed this part of code recently. https://github.com/typeorm/typeorm/commit/7a1fc80b3966348b5c8b90738edb6005733b36d4#diff-028edddbf412743bcc7f82126ec0cc02L289
Didn't test if old code was working well with your example before, but if it was I think we should still support it.

We're saying it's uncommon format, but is it compatible with sqlite specification?

louisremi:
The fix I'm proposing in my previous comment is simpler than the current two-steps fix, and it should handle Typeorm's own format more efficiently, while supporting a larger range of uncommon formats.
I understand that you are being cautious about adding more complexity to the code, but I believe I managed to avoid that.
My particular use-case is to use Typeorm on a project where Sequelize was originally used, but this fix improves compatibility with any existing Sqlite database.

CHANGELOG.md
```diff
@@ -7,6 +7,10 @@ feel free to ask us and community.
 
 ## 0.2.18 (UNRELEASED)
 
+### Bug fixes
```
imnotjames, line 10: Please remove this from the changelog - we update it upon release via tooling

src/driver/sqlite-abstract/AbstractSqliteDriver.ts
```diff
@@ -289,10 +289,15 @@ export abstract class AbstractSqliteDriver implements Driver {
...
-                // There are two separate fixes we may need to do:
-                //   1) Add 'T' separator if space is used instead
-                //   2) Add 'Z' UTC suffix if no timezone or offset specified
-
+                // There are three separate fixes we may need to do:
+                //   1) Remove spaces between the time and the timezone if any
+                //      (Sequelize serializes dates this way)
+                //   2) Add 'T' separator if space is used instead so that Safari
+                //      (and iOS WKWebView) is able to parse it
+                //   3) Add 'Z' UTC suffix if no timezone or offset specified
```
imnotjames, line 297: To make this clearer can you move the relevant comment above the block of code it applies to

imnotjames (changes requested):

AlexMesser:
Since there were no demand on this feature request for over two years, I think this feature is too specific and we shouldn't merge it. Thank you for your contribution anyway!

Issue #4042, opened 2019-04-24
MSSQL Server NOLOCK


alfonsoal1983:
**Issue type:**

[x] question
[ ] bug report
[ ] feature request
[ ] documentation issue

**Database system/driver:**

[ ] `cordova`
[ ] `mongodb`
[x] `mssql`
[ ] `mysql` / `mariadb`
[ ] `oracle`
[ ] `postgres`
[ ] `cockroachdb`
[ ] `sqlite`
[ ] `sqljs`
[ ] `react-native`
[ ] `expo`

**TypeORM version:**

[x] `latest`
[ ] `@next`
[ ] `0.x.x` (or put your version here)

**Steps to reproduce or a small repository showing the problem:**

Hello, I would like to know if the library supports the NOLOCK command. Query example:

SELECT * FROM User with(nolock)

alfonsoal1983:
Does anyone know anything about this?

PR #4133, opened 2019-05-13
added dirty read(NOLOCK) in SQLServer

alfonsoal1983:
Added support for NOLOCK in mssql
Linked to this issue

https://github.com/typeorm/typeorm/issues/4042

### DIFF
--- a/src/query-builder/SelectQueryBuilder.ts
+++ b/src/query-builder/SelectQueryBuilder.ts
@@ -1626,9 +1626,12 @@ export class SelectQueryBuilder<Entity> extends QueryBuilder<Entity> implements 
                 if (driver instanceof SqlServerDriver) {
                     return " WITH (NOLOCK)";
                     
-                } else if (driver instanceof MysqlDriver || driver instanceof PostgresDriver || driver instanceof OracleDriver) {
-                    return "";
+                } else if (driver instanceof PostgresDriver) {
+                    return " NOWAIT";
 
+                } else if (driver instanceof MysqlDriver || driver instanceof OracleDriver) {
+                    return "";
+                    
                 } else {
                     throw new LockNotSupportedOnGivenDriverError();
                 }
--- a/src/query-builder/SelectQueryBuilder.ts
+++ b/src/query-builder/SelectQueryBuilder.ts
@@ -949,12 +949,12 @@ export class SelectQueryBuilder<Entity> extends QueryBuilder<Entity> implements 
     /**
      * Sets locking mode.
      */
-    setLock(lockMode: "pessimistic_read"|"pessimistic_write"): this;
+    setLock(lockMode: "pessimistic_read"|"pessimistic_write"|"dirty_read"): this;
 
     /**
      * Sets locking mode.
      */
-    setLock(lockMode: "optimistic"|"pessimistic_read"|"pessimistic_write", lockVersion?: number|Date): this {
+    setLock(lockMode: "optimistic"|"pessimistic_read"|"pessimistic_write"|"dirty_read", lockVersion?: number|Date): this {
         this.expressionMap.lockMode = lockMode;
         this.expressionMap.lockVersion = lockVersion;
         return this;
@@ -1385,6 +1385,9 @@ export class SelectQueryBuilder<Entity> extends QueryBuilder<Entity> implements 
                 case "pessimistic_write":
                     lock = " WITH (UPDLOCK, ROWLOCK)";
                     break;
+                case "dirty_read":
+                    lock = " WITH (NOLOCK)";
+                    break;
             }
         }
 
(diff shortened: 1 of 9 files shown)


src/query-builder/SelectQueryBuilder.ts
```diff
@@ -1616,6 +1619,16 @@ export class SelectQueryBuilder<Entity> extends QueryBuilder<Entity> implements
...
 
+                } else {
+                    throw new LockNotSupportedOnGivenDriverError();
+                }
+            case "dirty_read":
+                if (driver instanceof SqlServerDriver) {
+                    return " WITH (NOLOCK)";
+                    
+                } else if (driver instanceof MysqlDriver || driver instanceof PostgresDriver || driver instanceof OracleDriver) {
+                    return "";
```
Kononnable: Why just ignore here? I doubt it's the default setting for those drivers.

test/functional/query-builder/locking/query-builder-locking.ts
```diff
@@ -99,6 +99,21 @@ describe("query builder > locking", () => {
...
+    it("should attach dirty read lock statement on query if locking enabled", () => Promise.all(connections.map(async connection => {
+        if (connection.driver instanceof AbstractSqliteDriver || connection.driver instanceof CockroachDriver)
+            return;
+
+        const sql = connection.createQueryBuilder(PostWithVersion, "post")
+            .setLock("dirty_read")
+            .where("post.id = :id", { id: 1 })
+            .getSql();
+
+        if (connection.driver instanceof SqlServerDriver) {
```
Kononnable: We limit drivers on wchich we want the test to run in two separate statements. It should be done only in one place.

src/query-builder/SelectQueryBuilder.ts
alfonsoal1983: Hello, it's my first climb so I apologize in advance for any possible errors. Searching the internet I found this about equivalents to this instruction in other databases:

MySQL:
When WITH (nolock) is used in SQL Server to perform operation, it does not place shared lock (S) and also does not honor exclusive locks on table.  The WITH (nolock) table hint is equivalent to READ UNCOMMITTED also known as "dirty read", it is the lowest level of isolation. If you specify table hint then it will override the current isolation level.
MySQL default isolation level is REPEATABLE READ which means locks will be placed for each operation but multiple connections can read data concurrently with out making table unavailable during read. MySQL database server does not support changing the default isolation mode as a hint like SQL Server.

Oracle:
Oracle by default provides a read-consistent view of data.  The upshot of this design is that readers never block writers.

PostgreSQL:
PostgreSQL uses MVCC to minimize lock contention in order to allow for reasonable performance in multiuser environments. Readers do not conflict with writers nor other readers. The closest thing would be the NOWAIT

It's what I found about it.

test/functional/query-builder/locking/query-builder-locking.ts
alfonsoal1983: I do not know what the correction would be, any suggestions?

Kononnable: What I had in mind is line 103 - we're limiting there some drivers, and here in if condition we choose only one driver after all. We could put this if condition in the beginning and there will be only one place which controls test execution based on driver.

src/query-builder/SelectQueryBuilder.ts
Kononnable: I still don't understand why we just pass empty sting in some drivers and not throwing `LockNotSupportedOnGivenDriverError`.
If driver doesn't support this it's fine but I don't think we should throw an error, not just ignore this - otherwise people using it will think that everything is working as expected while underneath typeorm would just ignore some unsupported features.

alfonsoal1983: I think it would be better to throw in error, although really this kind of cases do not think it is an error, it would be more informative that the driver does not have that functionality. What do you think?

test/functional/query-builder/locking/query-builder-locking.ts
alfonsoal1983: Do you mean to put row 103 in a generic way and import it from those test cases?

src/query-builder/SelectQueryBuilder.ts
alfonsoal1983: I just realized that this part of the code should not be set, it is enough with the change of line 1388. I delete this part of the code.

test/functional/query-builder/locking/query-builder-locking.ts
alfonsoal1983: File updated

alfonsoal1983:
Is everything correct or is something missing so that the merge can be done?

pleerock:
Looks good to me.

alfonsoal1983:
When will the changes be merged with the master branch?

Kononnable:
Looks fine, so let's merge it.
Thank you for the contribution.

PR #4144, opened 2019-05-14
 Fixed the incorrect sql query for changing the column name in sql

shijiezhou1:
Fixed the incorrect sql query for changing the column name in sql, 
it has to be like this:

ALTER TABLE table_name RENAME COLUMN old_name to new_name;

But the docs give the wrong query. I keep failing of making the migration.

reference: 
[https://www.1keydata.com/sql/alter-table-rename-column.html](https://www.1keydata.com/sql/alter-table-rename-column.html)

[https://www.w3schools.com/sql/sql_alter.asp](https://www.w3schools.com/sql/sql_alter.asp)

### DIFF
--- a/docs/migrations.md
+++ b/docs/migrations.md
@@ -128,14 +128,12 @@ import {MigrationInterface, QueryRunner} from "typeorm";
 export class PostRefactoringTIMESTAMP implements MigrationInterface {
     
     async up(queryRunner: QueryRunner): Promise<any> {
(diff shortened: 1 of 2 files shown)


shijiezhou1:
@pleerock would you look at this change that I made ? Thank you

moltar:
Does this work across all engines?

shijiezhou1:
@moltar yeah, I had tested it already.

PR #4157, opened 2019-05-18
fixes #4156 - QueryExpressionMap not cloning all values correctly

### LINKED ISSUE #4156: QueryExpressionMap doesn't clone all values correctly
**Issue type:**

[x] bug report

**Database system/driver:**

[x] `postgres`

**TypeORM version:**

[x] `0.2.17` (or put your version here)

`QueryExpressionMap`'s `clone` method doesn't clone all values correctly, specifically `insertColumns`, `whereEntities`, and `nativeParameters` are currently cloned by reference. The persistent reference to `nativeParameters` between clones causes issues with postgres as it generates the wrong parameter names (`$1`, `$2`, `$3`...) in `QueryBuilder.computeWhereParameter()`

```ts
...
const qb = connection.manager
  .createQueryBuilder("Post", "post");

await Promise.all([
  qb.clone().where({ id: 1 }).getOne(),
  qb.clone().where({ id: 1 }).getOne(),
]);

// postgres error - "could not determine data type of parameter $1"
```

I'm opening this issue so I can reference it in a PR that fixes this bug

pippinthegreat:
Fixes #4156

### DIFF
--- a/src/query-builder/QueryExpressionMap.ts
+++ b/src/query-builder/QueryExpressionMap.ts
@@ -409,7 +409,7 @@ export class QueryExpressionMap {
         map.updateEntity = this.updateEntity;
         map.callListeners = this.callListeners;
         map.useTransaction = this.useTransaction;
-        map.nativeParameters = this.nativeParameters;
+        map.nativeParameters = Object.assign({}, this.nativeParameters);
         return map;
     }
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ feel free to ask us and community.
 ### Bug fixes
 
 * fixed loadRelationCountAndMap when entities' primary keys are strings ([#3946](https://github.com/typeorm/typeorm/issues/3946))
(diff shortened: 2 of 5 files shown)


birkir:
Thanks @pippinthegreat, paging mr @pleerock

src/query-builder/QueryExpressionMap.ts
```diff
@@ -404,12 +404,12 @@ export class QueryExpressionMap {
         map.cacheDuration = this.cacheDuration;
         map.relationPropertyPath = this.relationPropertyPath;
         map.of = this.of;
-        map.insertColumns = this.insertColumns;
-        map.whereEntities = this.whereEntities;
+        map.insertColumns = this.insertColumns.map((column) => column);
+        map.whereEntities = this.whereEntities.map((entity) => ({ ...entity }));
```
pleerock: I don't think we should change those. I think your particular problem is in `nativeParameters`, isn't it?

pippinthegreat: okay, I've removed those changes

PR #4161, opened 2019-05-18
make nulls optional

thisrashid:


pleerock:
Why you didn't make it `{ 'message.createdAt': 'DESC' }` ?

thisrashid:
This is also one possible solution but IMHO nulls should be optional as it is not supported by all databases.

PR #4166, opened 2019-05-20
Fix promise waiting for sqlite with `key` option

hansl:
If a `key` option is passed to the Sqlite driver, the connection promise does not wait on it and resolves on the first pragma. This PR makes sure that the Promise returned will wait for the key PRAGMA to be run.

Also took some time to simplify the reading of the createDatabaseConnection() function.

### DIFF
--- a/src/driver/sqlite/SqliteDriver.ts
+++ b/src/driver/sqlite/SqliteDriver.ts
@@ -85,28 +85,36 @@ export class SqliteDriver extends AbstractSqliteDriver {
     /**
      * Creates connection with the database.
      */
-    protected createDatabaseConnection() {
-        return new Promise<void>(async (ok, fail) => {
-            await this.createDatabaseDirectory(this.options.database);
-            const databaseConnection = new this.sqlite.Database(this.options.database, (err: any) => {
+    protected async createDatabaseConnection() {
+        await this.createDatabaseDirectory(this.options.database);
+
+        const databaseConnection: any = await new Promise((ok, fail) => {
+            const connection = new this.sqlite.Database(this.options.database, (err: any) => {
                 if (err) return fail(err);
-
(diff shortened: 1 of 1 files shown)


src/driver/sqlite/SqliteDriver.ts
```diff
@@ -85,28 +85,36 @@ export class SqliteDriver extends AbstractSqliteDriver {
...
-                }
             });
-        });
+        }
+
+        // we need to enable foreign keys in sqlite to make sure all foreign key related features
+        // working properly. this also makes onDelete to work with sqlite.
+        await run(`PRAGMA foreign_keys = ON;`);
+
+        // in the options, if encryption key for for SQLCipher is setted.
```
pleerock: `for for`

src/driver/sqlite/SqliteDriver.ts
```diff
@@ -85,28 +85,36 @@ export class SqliteDriver extends AbstractSqliteDriver {
...
+        if (this.options.key) {
+            await run(`PRAGMA key = ${JSON.stringify(this.options.key)};`);
```
pleerock: `this.options.key` looks weird. is 'pragma key' SQLCipher specific stuff? If yes, maybe option should be called `sqlCipherKey` ?

src/driver/sqlite/SqliteDriver.ts
hansl: Fixed the comment.

src/driver/sqlite/SqliteDriver.ts
hansl: This is code that's already released. I don't think changing the API is in spec for this PR.

pleerock: oh I didn't realized that we already have this feature. Okay, thanks for the fix.

Issue #3954, opened 2019-04-08
[Proposal] cli show migration enhancement


aa-neg:
**Issue type:**

[ ] question
[ ] bug report
[X] feature request
[ ] documentation issue

**Database system/driver:**

[X] `cordova`
[X] `mongodb`
[X] `mssql`
[X] `mysql` / `mariadb`
[X] `oracle`
[X] `postgres`
[X] `cockroachdb`
[X] `sqlite`
[X] `sqljs`
[X] `react-native`
[X] `expo`

**TypeORM version:**

[X ] `latest`
[ ] `@next`
[ ] `0.x.x` (or put your version here)

**Goals**

To have greater confidence before applying / reverting migrations on a database.

**Current state**

Currently when running migrations from the cli useful information such as which migrations have been executed and the specific sql which would be run are displayed *after* the fact.

```bash
ts-node ./node_modules/typeorm/cli.js migration:run
```

**Proposal**

We propose to have functionality similar to https://docs.djangoproject.com/en/2.2/ref/django-admin/#django-admin-showmigrations which will allow the developer to receive the additional information before applying the changes. 

e.g some options

```bash
# New command (which would list which has been applied or not and developer can determine which is going to be run / reverted)
ts-node ./node_modules/typeorm/cli.js migration:show

# additional flag
ts-node ./node_modules/typeorm/cli.js migration:run --plan
ts-node ./node_modules/typeorm/cli.js migration:revert --plan
```

Hoping to generate some discussion on the feature and we are happy to do the work and make the PR. 

cc @Juddling

vidhu:
I agree and if TypeORM provided feature comparable to django's ORM. That'd be a game changer. Especially in the migrations area

Juddling:
Example output from Laravel's migration CLI tool:

```
> php artisan migrate:status
+------+------------------------------------------------------------------+-------+
| Ran? | Migration                                                        | Batch |
+------+------------------------------------------------------------------+-------+
| Y    | 2018_06_26_212752_create_test_table                              | 0     |
| Y    | 2018_06_26_212752_create_test2_table                             | 0     |
| Y    | 2018_06_26_212752_create_test3_table                             | 0     |
| Y    | 2018_06_26_212752_create_test4_table                             | 0     |
```

and another from Django:
```
> python manage.py showmigrations
admin
 [X] 0001_initial
 [X] 0002_logentry_remove_auto_add
auth
 [X] 0001_initial
[...]

PR #4173, opened 2019-05-21
typeorm migration:show

Juddling:
As proposed here: https://github.com/typeorm/typeorm/issues/3954

This PR adds a new command `typeorm migration:show`

This lists all migrations and shows whether they are applied or pending.

The command also returns an error code if there are unapplied migrations, so that it can easily be used  in CI pipelines.

### DIFF
--- a/src/commands/MigrationShowCommand.ts
+++ b/src/commands/MigrationShowCommand.ts
@@ -20,11 +20,6 @@ export class MigrationShowCommand implements yargs.CommandModule {
         default: "default",
         describe: "Name of the connection on which run a query."
       })
-      .option("transaction", {
-        alias: "t",
-        default: "default",
-        describe: "Indicates if transaction should be used or not for migration run. Enabled by default."
-      })
       .option("config", {
         alias: "f",
         default: "ormconfig",
@@ -48,11 +43,7 @@ export class MigrationShowCommand implements yargs.CommandModule {
         logging: ["query", "error", "schema"]
       });
       connection = await createConnection(connectionOptions);
-
-      const options = {
-        transaction: args["t"] === "false" ? false : true
-      };
-      const unappliedMigrations = await connection.showMigrations(options);
+      const unappliedMigrations = await connection.showMigrations();
       await connection.close();
 
       // return error code if there are unapplied migrations for CI
--- a/src/commands/MigrationShowCommand.ts
+++ b/src/commands/MigrationShowCommand.ts
@@ -1,0 +1,69 @@
+import {createConnection} from "../index";
+import {ConnectionOptionsReader} from "../connection/ConnectionOptionsReader";
+import {Connection} from "../connection/Connection";
+import * as process from "process";
+import * as yargs from "yargs";
+const chalk = require("chalk");
+
+/**
+ * Runs migration command.
+ */
+export class MigrationShowCommand implements yargs.CommandModule {
+
+  command = "migration:show";
+  describe = "Show all migrations and whether they have been run or not";
+
+  builder(args: yargs.Argv) {
+    return args
+      .option("connection", {
+        alias: "c",
+        default: "default",
+        describe: "Name of the connection on which run a query."
+      })
+      .option("transaction", {
+        alias: "t",
+        default: "default",
+        describe: "Indicates if transaction should be used or not for migration run. Enabled by default."
+      })
+      .option("config", {
+        alias: "f",
+        default: "ormconfig",
+        describe: "Name of the file with connection configuration."
+      });
+  }
+
+  async handler(args: yargs.Arguments) {
+    let connection: Connection|undefined = undefined;
+    try {
+      const connectionOptionsReader = new ConnectionOptionsReader({
+        root: process.cwd(),
+        configName: args.config as any
+      });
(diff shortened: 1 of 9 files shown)


ashbadger:
@Juddling could you also add this to the [using-cli.md](https://github.com/typeorm/typeorm/blob/master/docs/using-cli.md)?

i took a stab:
## Show migrations and whether they've been run or not

To show all migrations and whether they've been run or not use following command:
```
typeorm migration:show
```

This command also returns an error code if there are unapplied migrations.

Juddling:
Added, thank you @ashbadger

src/commands/MigrationShowCommand.ts
```diff
@@ -0,0 +1,70 @@
...
+  describe = "Show all migrations and whether they have been run or not";
+
+  builder(args: yargs.Argv) {
+    return args
+      .option("connection", {
+        alias: "c",
+        default: "default",
+        describe: "Name of the connection on which run a query."
+      })
+      .option("transaction", {
```
Kononnable: It's probably not necessary if we're just going to show migrations without running them.

src/migration/MigrationExecutor.ts
```diff
@@ -48,6 +48,40 @@ export class MigrationExecutor {
...
+        const executedMigrations = await this.loadExecutedMigrations(queryRunner);
+
+        // get all user's migrations in the source code
+        const allMigrations = this.getMigrations();
+
+        for (const migration of allMigrations) {
+            const executedMigration = executedMigrations.find(executedMigration => executedMigration.name === migration.name);
+
+            if (executedMigration) {
+                this.connection.logger.logSchemaBuild(` ✅ ${migration.name}`);
```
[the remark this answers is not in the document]
Juddling: I just had a dig through yarn's codebase and it looks like they only enable it for specific terminals, so I'll remove these emojis 😢 😞 

https://github.com/yarnpkg/yarn/blob/d4805c302d9a02fd2a52cb4e02f9092b4689fbb1/src/cli/index.js#L116

```js
  commander.option(
    '--emoji [bool]',
    'enable emoji in output',
    boolify,
    process.platform === 'darwin' ||
      process.env.TERM_PROGRAM === 'Hyper' ||
      process.env.TERM_PROGRAM === 'HyperTerm' ||
      process.env.TERM_PROGRAM === 'Terminus',
  );
```

Juddling:
@Kononnable thanks for the review, I've pushed changes with your recommendations

src/connection/Connection.ts
```diff
@@ -305,6 +305,18 @@ export class Connection {
         await migrationExecutor.undoLastMigration();
     }
 
+    /**
+     * Runs all pending migrations.
```
vlapo: Could you please update or remove this comment ? :)

Juddling: Done, good spot

test/functional/migrations/show-command/command.ts
```diff
@@ -0,0 +1,26 @@
+import "reflect-metadata";
+import {createTestingConnections, closeTestingConnections, reloadTestingDatabases} from "../../../utils/test-utils";
+import {Connection} from "../../../../src/connection/Connection";
+
+describe("migrations > show command", () => {
+    let connections: Connection[];
+    before(async () => connections = await createTestingConnections({
+        migrations: [__dirname + "/migration/*.js"],
+        enabledDrivers: ["postgres"],
```
rustamwin: Could we enable more drivers?

vlapo: There are no driver specific changes. I think run this test only for one driver is ok.

Juddling: I've enabled them all for now, if it adds too much of a time hit to the tests, we can drop to fewer drivers

Juddling: I'll drop back down to just postgres then 👌

pleerock:
Thank you very much guys!

ryanoboril:
Great work, thank you!

glen-84:
@Juddling,

This makes absolutely no sense as the default behaviour. It is in no way an error for there to be unapplied migrations. Now, every time that this command is run manually by a user, they get 10 `npm ERR!` lines? That's just wrong.

If you need different behaviour in a CI environment, then add a CLI argument. A command run in a CI environment is usually only defined once, as opposed to users running this command many times manually. `|| 1` is _not_ a solution.

Juddling:
> @Juddling,
> 
> This makes absolutely no sense as the default behaviour. It is in no way an error for there to be unapplied migrations. Now, every time that this command is run manually by a user, they get 10 `npm ERR!` lines? That's just wrong.
> 
> If you need different behaviour in a CI environment, then add a CLI argument. A command run in a CI environment is usually only defined once, as opposed to users running this command many times manually. `|| 1` is _not_ a solution.

Initial use case for this was to run in CI pre deploy, i.e. all code migrations should be run before new code version is deployed.

If you're running this locally I can see why the error code isn't useful. I'd suggest opening a github issue tagging me, outlining the issue (are you having any problems beyond the fact that you don't like seeing those err lines) and a proposed solution - baring in mind that people may depend on this command returning the error code so this is probably a breaking change

alper:
Can we get this to also output the SQL?

PR #4182, opened 2019-05-24
fix: length limit for join table alias in FindOptions (table name specified more than once)

aethys256:
Hello!

### Context

This is a follow-up of the PR https://github.com/typeorm/typeorm/pull/3514

The issue is kinda the same except it's happening in join alias instead of select alias resulting in a `ERROR:  table name "CategoryWithVeryLongName__postsWithVeryLongName__authorWithVery" specified more than once` due to the fact names are truncated.

### Solution

The way I've implemented a fix is similar to what @VinceOPS did, by re-using his `shorten()`.

### Tests

I'm going to write a test that will cover what this PR is addressing, I only tested it on Postgres right now but considering it's re-using existing Driver's option `maxAliasLength` it should be safe.

### Comments

Let me know if there is anything else I should do in order to get the PR merged.  
We are currently maintaining a fork of TypeORM with this fix on our apps until it's being reviewed and merged.


cdstamper:
@pleerock what woul dit take to get this one merged as well?

pleerock:
Looks good, but would be great to have a review from  @VinceOPS

aethys256:
I've updated the PR to include an update to the existing test related to alias max length, this time both for select & join alias.
Let me know if there is anything missing to get it merged.
cc @VinceOPS

pleerock:
Let's merge it, hope @VinceOPS will find a time to review it later. Thanks!

aethys256:
Looks like changelog has to be manually updated, should I open a PR to add a line about this fix ?

pleerock:
@aethys256 yes, please

Issue #4185, opened 2019-05-24
afterLoad() subscriber interface missing additional info available on other events


lovrop:
**Issue type:**

[ ] question
[ ] bug report
[x] feature request
[ ] documentation issue

**TypeORM version:**

[x] `latest`
[ ] `@next`
[ ] `0.x.x` (or put your version here)

**Steps to reproduce or a small repository showing the problem:**

The [subscriber interface for `afterLoad()`](https://github.com/typeorm/typeorm/blob/master/src/subscriber/EntitySubscriberInterface.ts#L19) is inconsistent with the other events. All the other hooks take objects which have the entity and more context about the event (e.g. QueryRunner and EntityManager used).

The reason for this seems to be largely historical, due to how the interfaces evolved over time.  The info is available in `Broadcaster.ts` but not passed because it doesn't fit the interface.

This was previously reported in #2510 but closed after 2 months without comment.

lovrop:
I think this should be easily fixable in a backward-compatible way by adjusting the interface like so:

```ts
afterLoad?(entity: Entity, event?: LoadEvent<Entity>): Promise<any>|void;
```

The optionality of the `event` parameter will keep existing subscribers working. New subscribers can use the new parameter to get additional info about the event.

I'm looking into making the change, hope to have a pull request ready soon.

PR #4186, opened 2019-05-24
Extend afterLoad() subscriber interface to take LoadEvent

lovrop:
This brings it in line with the other events which already have the Connection, EntityManager, QueryRunner etc available.

Test Plan: npm test, and more specifically npm test -- --grep='github issues > #4185'


lovrop:
I couldn't get a clean full test run on my laptop, but most tests passed and I didn't see any regressions with the proposed code change.

pleerock:
Looks good to go.

Issue #4188, opened 2019-05-25
Documentation for One to Many has an erroneous reference


s-chand:
**Issue type:**

[ ] question
[ ] bug report
[ ] feature request
[x ] documentation issue

**Database system/driver:**

[ ] `cordova`
[ ] `mongodb`
[ ] `mssql`
[ ] `mysql` / `mariadb`
[ ] `oracle`
[ ] `postgres`
[ ] `cockroachdb`
[ ] `sqlite`
[ ] `sqljs`
[ ] `react-native`
[ ] `expo`

**TypeORM version:**

[x] `latest`
[ ] `@next`
[ ] `0.x.x` (or put your version here)

**Steps to reproduce or a small repository showing the problem:**
The documentation has a reference to the following text:
```
Here we added `@ManyToOne` to the `photos` property and specified the target relation type to be `Photo`.

```

This is incorrect and should be corrected to

```
Here we added `@OneToMany` to the `photos` property and specified the target relation type to be `Photo`.

```
 as this is actually what is done in the example referenced in the documentation.

I'm submitting a PR to fix this shortly.
<!--
    To answer those questions you need to put "x" inside the square brackets, for example:
    [x] `mysql`
    [ ] `postgres`

    Also, please format your code properly (by taking code blocks into ```ts .... ```)
!>
<img width="947" alt="Screenshot 2019-05-25 at 15 36 09" src="https://user-images.githubusercontent.com/1958765/58370100-de6f2980-7f02-11e9-8ca0-5437ba611871.png">

PR #4189, opened 2019-05-25
Correct error in decorator reference in documentation example for One to Many/ Many to One.

s-chand:
The documentation for OneToMany and ManyToOne has an error in reference to the decorator applied to photos property in the User entity. This commit rectifies this by switching the ManyToOne reference to the correct OneToMany decorator used in the example.

It fixes issue https://github.com/typeorm/typeorm/issues/4188


pleerock:
Thank you very much!

PR #4195, opened 2019-05-26
feat(RelationDecorators): allow to pass string instead of typeFunction

### LINKED ISSUE #4190: Relation decorators: allow to pass string instead of typeFunction
**Issue type:**

[ ] question
[ ] bug report
[x] feature request
[ ] documentation issue

---

To define table relations using decorators, this is the current way:

Photo.ts
```ts
    @ManyToOne(type => User, user => user.photos)
    user: User;
```

User.ts
```ts
    @OneToMany(type => Photo, photo => photo.user)
    photos: Photo[];
```

This works fine on Node.js, but can cause circular dependency issues on the frontend side (e.g. when using Angular CLI). See: #2059 or #1290

We could overcome those issues, if the following would be possible:

Photo.ts
```ts
    @ManyToOne('User', 'photos')
    user: User;
```

User.ts
```ts
    @OneToMany('Photo', 'user')
    photos: Photo[];
```

This should be fairly easy to implement, as it turns out that TypeORM already supports `string` instead of types (e.g. when using `EntitySchema`) - the only thing to do is to adjust the `typeFunction` type of the decorator functions.

@pleerock If that feature would be ok I could post a PR :)

haschu:
Closes #4190

In addition to
```ts
@OneToOne(type => Profile)
@ManyToOne(type => User, user => user.photos)
@ManyToMany(type => Category)
```
now it's also possible to write:
```ts
@OneToOne('Profile')
@ManyToOne('User','photos')
@ManyToMany('Category')
```

Using this approach can prevent circular dependency issues when TypeORM is used within a frontend/browser application (e.g. when running through Angular CLI).

### DIFF
--- a/src/decorator/relations/ManyToMany.ts
+++ b/src/decorator/relations/ManyToMany.ts
@@ -6,14 +6,15 @@ import {RelationMetadataArgs} from "../../metadata-args/RelationMetadataArgs";
  * multiple instances of Entity1. To achieve it, this type of relation creates a junction table, where it storage
  * entity1 and entity2 ids. This is owner side of the relationship.
  */
-export function ManyToMany<T>(typeFunction: (type?: any) => ObjectType<T>, options?: RelationOptions): Function;
+export function ManyToMany<T>(typeFunctionOrTarget: string|((type?: any) => ObjectType<T>), 
+                              options?: RelationOptions): Function;
 
 /**
  * Many-to-many is a type of relationship when Entity1 can have multiple instances of Entity2, and Entity2 can have
  * multiple instances of Entity1. To achieve it, this type of relation creates a junction table, where it storage
  * entity1 and entity2 ids. This is owner side of the relationship.
  */
-export function ManyToMany<T>(typeFunction: (type?: any) => ObjectType<T>,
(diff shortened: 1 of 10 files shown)


haschu:
Cool, tests failed for Node.js 9! I'm not sure if that's related to my changes... Error in `before each`, can anybody restart it?

pleerock:
Thanks, it's a good change. We should have it much earlier.

pleerock:
Oh, we should make a change in a changelog .

haschu:
Argh... sorry, I overlooked the changelog part in CONTRIBUTING.md - going to submit a PR to fix that soon.

I also realized, that [typeorm-class-transformer-shim.js](https://github.com/typeorm/typeorm/blob/a4dec02cc59d3219a29c7be0322af2253e1452dc/extra/typeorm-class-transformer-shim.js#L142) wouldn't work with the "string instead of typeFunction" approach. I'm also trying to fix that soon.

@pleerock Thank you so much for the fast merge - I have noticed that you're struggling at the moment to find enough time for TypeORM. I really appreciate it, as my team is migrating our in-house framework to TypeORM and we were stuck at this point. And also thanks for your effort to bring up such an amazing project! :)

PR #4202, opened 2019-05-27
Update FindOneOptions.ts

dilame:
Improve relations attribute typings.


pleerock:
We can't accept this change since it break exist functionality with relations being inside embeds, e.g. `user.counters.info`. This, however is addressed and fixed in `@next`.

PR #4211, opened 2019-05-30
fix(changelog): Added missing line for PR #4195

haschu:
I forgot to update the changelog in #4195 - here is the fix for that.


pleerock:
Thank you!

PR #4222, opened 2019-06-01
feat(class-transformer-shim): Added support for metadata reflection

### LINKED ISSUE #4219: class-transformer-shim: support metadata reflection
**Issue type:**

[ ] question
[ ] bug report
[x] feature request
[ ] documentation issue

**TypeORM version:**

[x] `latest`
[ ] `@next`

---

TypeORM provides a [class-transformer-shim](https://github.com/typeorm/typeorm/blob/master/extra/typeorm-class-transformer-shim.js), which is awesome for browser/fronend application usage.

However, the transformation of `Date` types for example - or any other type, that isn't passed via a decorator and a typeFunction - requires additional `class-transformer` code:

```ts
@Entity()
export class Photo {

    @PrimaryGeneratedColumn()
    id: number;

    @Column()
    url: string;
    
    // Works fine => class-transformer creates User objects :)
    @ManyToOne(type => User, user => user.photos)
    user: User;
    
    // No Date object, class-transformer leaves that as string :(
    @CreateDateColumn()
    createdAt: Date;

}
```

To make it work, we need to write:

```ts
import { Type } from 'class-transformer';

//...

@Type(() => Date)
@CreateDateColumn()
createdAt: Date;
```

which I find annoying, especially when sharing entities between backend and frontend. One would also have to import the `class-transformer` library into his backend, but actually it's only used on the frontend side. Plus of course, it saves us lots of code lines.

We could solve this, if [typeorm-class-transformer-shim.js](https://github.com/typeorm/typeorm/blob/master/extra/typeorm-class-transformer-shim.js) would support metadata reflection. 
If the client imports "reflect-metadata", we could use `design:type` and automatically pass it to the `@Type` decorator of class-transformer, as we already do with typeFunctions.

PR is on the way.

haschu:
closes #4219

Solution:

- First check if a `typeFunction` is passed to a decorator and then pass `typeFunction` to `@Type` (this is the current behavior)
- If no `typeFunction` is passed and "reflect-metadata" is imported, try to use `design:type`

So for example:

```ts
@Entity()
export class Photo {
     
    // ...

    @CreateDateColumn()
    createdAt: Date;

}

// ...

const photo = plainToClass(Photo, {
  //...
  createdAt: 'Sat Jun 01 2019'
});

photo.createdAt instanceof Date // true
```

Just a small note: 

In order to make it testable I've added "class-transformer" to the devDependencies.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,7 @@ feel free to ask us and community.
 * added support for `dirty_read` (NOLOCK) in SQLServer ([#4133](https://github.com/typeorm/typeorm/pull/4133))
 * extend afterLoad() subscriber interface to take LoadEvent ([issue #4185](https://github.com/typeorm/typeorm/issues/4185))
 * relation decorators (e.g. `@OneToMany`) now also accept `string` instead of `typeFunction`, which prevents circular dependency issues in the frontend/browser ([issue #4190](https://github.com/typeorm/typeorm/issues/4190))
+* added support for metadata reflection in typeorm-class-transformer-shim.js ([issue #4219](https://github.com/typeorm/typeorm/issues/4219))
 
 ## 0.2.17 (2019-05-01)
 
--- a/extra/typeorm-class-transformer-shim.js
+++ b/extra/typeorm-class-transformer-shim.js
@@ -1,5 +1,6 @@
 // this "shim" can be used on the frontend to make class-transformer to work with typeorm models automatically
 // without having to put @Type decorator on properties that already have type information inside relational decorators.
+// if "reflect-metadata" is imported, you can also leave out the @Type decorator for Date or other types.
 // using this shim you can share same models across backend and frontend more easily.
 // to use this shim simply configure your systemjs/webpack configuration to use this file instead of typeorm module.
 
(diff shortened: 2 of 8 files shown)


pleerock:
I can merge this one, but I think those "extra" stuff should be extracted out of the core library in the future.

haschu:
Thank you! 

I'm working on some features to improve the frontend experience of TypeORM and going to submit some PRs in the future.

I wouldn't disagree - maybe it makes sense to outsource this kind of stuff to a "typeorm/typeorm-frontend" repository, including examples on how to share entities between backend/frontend? 
But I think that atm, the feature amount would be a bit to small for a separate project.

pleerock:
Personally I don't recommend to share entities between frontend/backend (current way) since its hacky and tricky. So I don't want to recommend this way as "official way".

haschu:
I think this can improve the developer experience, if done right.

E.g. our in-house framework is now able to generate a websocket api from TypeORM entities. On the frontend side we kind of re-implemented the repository pattern, and share entities like:

```ts
const hero = heroRepository.create({
  name: 'Batman',
  mail: 'bruce@wayne.de',
  powers: [batPower]
});
await heroRepository.save(hero).toPromise();
```

The time that we save with this approach is enormous.

But I do agree, that its kind of tricky and hacky - but I think that can be improved :)

PR #4226, opened 2019-06-03
feat: add support for SQLite ON CONFLICT clause

rsbh:
Sqlite provides `ON CONFLLICT` at the time of table creation. 
https://sqlite.org/conflict.html


pleerock:
Thanks for contribution.

What if we change `onConflict`? Would ORM properly react on changes? Also, this feature needs a test coverage.

rishabh09:
@pleerock i will add some tests for this change

pleerock:
tests are failing

imnotjames (changes requested):
This still needs tests that confirm this is working as expected

PR #4230, opened 2019-06-03
Reuse preloaded null values for lazy loads

davewasmer:
The lazy load getter checks to see if the relation was already loaded by checking to see if the data for the relation is stored on the `dataIndex` key of the entity (i.e. `post.author` will check for `post.__author__` and reuse it if present).

Previously, it checked for a truthy value (just a simple `if (post.__author__)`). However, if the data for that relation was previously loaded, and that relation is empty, then `post.__author__` will be `null`. The simple truthy check will fail, and it will query the DB to attempt to load the relation.

This PR changes the check to explicitly check for `undefined`, rather than any falsey value. That way, preloaded `null` relations can reuse the preloaded value.

### DIFF
--- a/src/query-builder/RelationLoader.ts
+++ b/src/query-builder/RelationLoader.ts
@@ -191,7 +191,7 @@ export class RelationLoader {
 
         Object.defineProperty(entity, relation.propertyName, {
             get: function() {
-                if (this[resolveIndex] === true || this[dataIndex]) // if related data already was loaded then simply return it
+                if (this[resolveIndex] === true || this[dataIndex] !== undefined) // if related data already was loaded then simply return it
                     return Promise.resolve(this[dataIndex]);
 
                 if (this[promiseIndex]) // if related data is loading then return a promise relationLoader loads it


pleerock:
There is a `this[resolveIndex] === true` check that already resolves this issue. No matter if you loaded `null` or some real value from the database, once it's loaded `this[resolveIndex]` is set to true and this if block will be executed and value will be returned without reloading it again.

davewasmer:
@pleerock in my testing, it seems like `resolveIndex` is not set for data preloaded through a query builder that uses left joins and the relation shorthand:

```
let user = await User.createQueryBuilder('User')
  .leftJoinAndSelect('User.profile', 'Profile')
  .leftJoinAndSelect('Profile.headshot', 'Headshot')
  .where({ id })
```

`user` ends up looking like this:

image

The `dataIndex` (in this case, `__profile__`) is set, but I don't see the `resolveIndex` (in this case, should be `__has_profile__`).

Perhaps there's a bug with the relation loading? But in either case, seems like there's no downside to making the `dataIndex` check here more accurate?

pleerock:
I think we can accept this change too, but we need to find a source of the problem.

pleerock:
Okay, let's merge it

davewasmer:
Thanks @pleerock! Agreed, it would probably be a good idea to find the underlying reason that `resolveIndex` is not being set here.

Issue #3732, opened 2019-02-28
@Unique decorator don't recognize column names


webleaf:
**Issue type:**

[ ] question
[x] bug report
[ ] feature request
[x] documentation issue

**Database system/driver:**

[ ] `cordova`
[ ] `mongodb`
[ ] `mssql`
[ ] `mysql` / `mariadb`
[ ] `oracle`
[ ] `postgres`
[ ] `sqlite`
[ ] `sqljs`
[ ] `react-native`
[ ] `expo`

**TypeORM version:**

[x] `latest`
[ ] `@next`
[ ] `0.x.x` (or put your version here)

**Steps to reproduce or a small repository showing the problem:**

If column name is specified via `@Column({name: string})` or `@JoinColumn({name: string})`, such column don't recognized by `@Unique` and we still should use Entity original property name.

```typescript
@Entity()
@Unique(["field_name_1", "field_name_2"])
export class EntityName {

    @PrimaryGeneratedColumn()
    public id: number;

    @ManyToOne(type => OtherEntityName)
    @JoinColumn({name: "field_name_1"})
    public field1: OtherEntityName;

    @Column({name: "field_name_2"})
    public field2: string;
}
```
```
Error: Unique constraint contains column that is missing in the entity: field_name_1
```

vlapo:
Hey @webleaf. You have to use entity field name not a DB column name:
```typescript
@Unique(["field1", "field2"])
```

webleaf:
@vlapo yes, I know. Then it's docs bug:

> This decorator allows you to create a database unique constraint **for a specific column or columns**.

vlapo:
Feel free to contribute :) Note in https://typeorm.io/#/indices should be useful.

sheychen290:
Maybe related...
```
@Entity()
@Unique(["field1", "field2"])
export class EntityName {
    @ManyToOne(type => OtherEntityName, {primary: true})
    @JoinColumn({name: "field_name_1"})
    public field1: OtherEntityName;

    @Column({name: "field_name_2", primary: true})
    public field2: string;
}
```
Produce ```Error: Index contains column that is missing in the entity: field1```
Note: Same problem with ```field_name_1```

pleerock:
fixed by #4233

qathom:
Hello,

I'd like to share my approach.
Hope it can help someone.

```js
function notUnique<Entity>(entityA: Entity, entityB: Entity, tableUnique: (keyof Entity)[]): boolean {
[...]

PR #4233, opened 2019-06-04
give more clarity on arguments passed to unique

kedar-moghe:
I found the documentation for https://typeorm.io/#/decorator-reference/unique kind of confusing at first because I thought it was expecting database column names.

I also found an issue where a user asked for more clarity on this.  https://github.com/typeorm/typeorm/issues/3732


pleerock:
Thank you!

Issue #4234, opened 2019-06-04
default value of @UpdateDateColumn is `CURRENT_TIMESTAMP ON UPDATE CURRENT_TIMESTAMP`


koheiiwamura:
**Issue type:**

[ ] question
[x] bug report
[ ] feature request
[ ] documentation issue

**Database system/driver:**

[ ] `cordova`
[ ] `mongodb`
[ ] `mssql`
[x] `mysql` / `mariadb`
[ ] `oracle`
[ ] `postgres`
[ ] `cockroachdb`
[ ] `sqlite`
[ ] `sqljs`
[ ] `react-native`
[ ] `expo`

**TypeORM version:**

[ ] `latest`
[ ] `@next`
[x] `0.2.16` (or put your version here)

**Steps to reproduce or a small repository showing the problem:**

I use UpdateDateColumn to entity. 

but when I generated migrationfile, it wrote that 

```
'await queryRunner.query("ALTER TABLE `users` ADD `updated_at` datetime(6) NOT NULL DEFAULT CURRENT_TIMESTAMP(6)");'
```

default of updated_at should be `DEFAULT CURRENT_TIMESTAMP ON UPDATE CURRENT_TIMESTAMP`

<!--
    To answer those questions you need to put "x" inside the square brackets, for example:
    [x] `mysql`
    [ ] `postgres`

    Also, please format your code properly (by taking code blocks into ```ts .... ```)
!>

I use UpdateDateColumn to entity. 

but when I generated migrationfile, it is written that 

```
'await queryRunner.query("ALTER TABLE `users` ADD `updated_at` datetime(6) NOT NULL DEFAULT CURRENT_TIMESTAMP(6)");'
```

default of updated_at should be `DEFAULT CURRENT_TIMESTAMP ON UPDATE CURRENT_TIMESTAMP`

vlapo:
As I mentioned (https://github.com/typeorm/typeorm/pull/4236#issuecomment-498850543) this is by design and we update `UpdateDateColumn` columns manually. But I think it is nice enhancement of current behaviour. We should use native DB functionality every time it is possible.

koheiiwamura:
I see. thanks!

PR #4236, opened 2019-06-04
change dafault value of @updateDateColumn

koheiiwamura:
https://github.com/typeorm/typeorm/issues/4234


pleerock:
I don't understand one thing. Since its a basic functionality it must currently work. If so, how does it work right now? If no, is there any proof on that?

vlapo:
Currently we support more than mysql driver. Other DBs like sqlite do not support `ON UPDATE` syntax. So we update `UpdateDateColumn` "manualy" every time `save` method is used. We should close this as it is not a solution of root cause.

koheiiwamura:
I see, thanks!

PR #4239, opened 2019-06-05
Dealt with numbers in change detection

MatthiasKunnen:
Previously, number columns would always be marked as different and an update was executed. I changed the `SubjectChangedColumnsComputer.computeDiffColumns` to normalize numbers.

### DIFF
--- a/src/persistence/SubjectChangedColumnsComputer.ts
+++ b/src/persistence/SubjectChangedColumnsComputer.ts
@@ -1,3 +1,4 @@
+import {NumericTypes} from "../driver/types/ColumnTypes";
 import {Subject} from "./Subject";
 import {DateUtils} from "../util/DateUtils";
 import {ObjectLiteral} from "../common/ObjectLiteral";
@@ -78,6 +79,11 @@ export class SubjectChangedColumnsComputer {
 
                     } else if (column.type === "time") {
                         normalizedValue = DateUtils.mixedDateToTimeString(entityValue);
+
+                    } else if (NumericTypes.some(t => t === column.type)) {
+                        normalizedValue = column.precision != null
+                            ? normalizedValue.toPrecision(column.precision)
+                            : normalizedValue.toString();
 
                     } else if (column.type === "datetime" || column.type === Date) {
                         normalizedValue = DateUtils.mixedDateToUtcDatetimeString(entityValue);
--- a/src/driver/types/ColumnTypes.ts
+++ b/src/driver/types/ColumnTypes.ts
@@ -194,3 +194,33 @@ export type ColumnType = WithPrecisionColumnType
     |DateConstructor
     |NumberConstructor
     |StringConstructor;
+
+export const NumericTypes = [
+    "bigint", // mysql, postgres, mssql, sqlite, cockroachdb
+    "dec", // oracle, mssql
+    "dec", // oracle, mssql, mysql
+    "decimal", // mysql, postgres, mssql, sqlite
+    "decimal", // mysql, postgres, mssql, sqlite
+    "double precision", // postgres, oracle, sqlite, mysql, cockroachdb
+    "double", // mysql, sqlite
+    "fixed", // mysql
+    "fixed", // mysql
+    "float", // mysql, mssql, oracle, sqlite
+    "float4", // postgres, cockroachdb
+    "float8", // postgres, cockroachdb
+    "int", // mysql, mssql, oracle, sqlite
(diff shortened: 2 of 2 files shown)


src/persistence/SubjectChangedColumnsComputer.ts
```diff
@@ -79,6 +80,11 @@ export class SubjectChangedColumnsComputer {
                     } else if (column.type === "time") {
                         normalizedValue = DateUtils.mixedDateToTimeString(entityValue);
 
+                    } else if (NumericTypes.some(t => t === column.type)) {
+                        normalizedValue = column.precision != null
+                            ? normalizedValue.toPrecision(column.precision)
+                            : normalizedValue.toString();
```
pleerock: Why do you put `.toString()` here? I guess your problem in in precision? If so, how exactly your change fixed your problem and what exact problem you had that forced you to add `toString()`? Please provide a detailed explanations on your changes. Why it didn't work before and works now? Also please add tests to your change.

MatthiasKunnen: When arriving at line 84, normalizedValue is of type `number` and database value is of type `string`. To make a correct comparison, I convert the normalized value to a string by using a) `toPrecision` if the precision is known b) `toString` when the precision is not known. This is the safest comparison I could think of that wouldn't say both are equal when they are not while still recognizing equal values.

About tests, I've looked for existing tests to append but have not found any relating to the SubjectChangedColumnsComputer. Could you point me to the right files? Alternatively, I could extract the comparison to a utility function and add tests for the function specifically.

pleerock: Yeah, there are no similar tests, but we should have one...

pleerock: Also I feel like you are breaking some stuff we already have and since there are no tests in this area I feel it dangerous to merge it.

pleerock: Why I'm thinking you are breaking - because I don't understand why you need toString. You are saying that `databaseValue` is a string and `normalizedValue` is a number - are you sure its true for all number types??? Because if so, you are saying that currently check doesn't work properly for all numbers and it make redundant updates all the times, but it can't be true.

MatthiasKunnen: I definitely understand your concern. 

What I aim to achieve is preventing unnecessary update queries when number columns are involved. Unless people rely on updates of unchanged numeric values it should not break anything. I could however be wrong as I don't know the intricacies of the source code. My manual tests do succeed.

pleerock: Try to remove your code and check on a simple number. What you are saying it will make unnecessary update. What I'm saying it should not do it. Can you try and check it again for a simple number case?

MatthiasKunnen: I just tried with an int, it compares a number to a number there. I did my tests with decimals (Postgres).

I reckon this is because `bigNumberStrings` is turned off. So this is an edge case :thinking:

MatthiasKunnen: Although, `bigNumberStrings` is only for mysql/mariadb so that's not it.

MatthiasKunnen: Current tests for Postgres:

| Type             | normalizationValue | databaseValue |
|------------------|--------------------|---------------|
| decimal          | number             | string        |
| float            | number             | number        |
| numeric          | number             | string        |
| int              | number             | number        |
| double precision | number             | number        |
| real             | number             | number        |

Want me to try other types?

MatthiasKunnen: I just added bigint to the table which apparently also is a string as database value

pleerock: So we need `toString()` only for `decimal`, `numeric` and `bigint`, am I right? And using it for other types would bring issues since type of normalizationValue wouldn't be equal to value emitted by `toString()`?

MatthiasKunnen:
We need toPrecision and toString only for those types indeed.

pleerock:
I guess if we would merge this PR we would have bad bugs. We'll need a good test coverage for this functionality

imnotjames (changes requested):
Needs tests to show this fixes the expected behaviors at the very least

AlexMesser:

MatthiasKunnen:
I might come back to it at some point

Issue #3501, opened 2019-01-24
Querybuilder generating syntax errors


robbyemmert:
**Issue type:**

[ x] question
[x ] bug report
[ ] feature request
[x ] documentation issue

**Database system/driver:**

[ ] `cordova`
[ ] `mongodb`
[ ] `mssql`
[ ] `mysql` / `mariadb`
[ ] `oracle`
[ x] `postgres`
[ ] `sqlite`
[ ] `sqljs`
[ ] `react-native`
[ ] `expo`

**TypeORM version:**

[ ] `latest`
[ ] `@next`
[ ] `^0.2.7`

QueryBuilder.getMany() and .getManyAndCount() are both wrapping my queries in a SELECT DISTINCT clause, which breaks my query when I use .orderBy. It's deceptive, because it's formatting the SQL different from .getSql(). The SQL from .getSql() works just fine when I run it manually.

I'm not sure why queryBuilder is wrapping my query in a SELECT DISTINCT clause... Is there a reason for this, or a way to disable it? 

I'm not sure if this is a bug, question, or just something that needs better documentation.

Can .orderBy be used with queryBuilder?

My Code:
```ts
const query = this.serviceRepository.createQueryBuilder('service')
            .leftJoinAndSelect('service.type', 'service_type')
            // .where("LOWER(\"service\".\"countryName\") LIKE LOWER(:text)", {
            //     text: '%' + search.country + '%'
            // })
            // .orWhere("LOWER(\"service\".\"countryCode\") LIKE LOWER(:text)", {
            //     text: '%' + search.country + '%'
            // })
            .orderBy('\"service_countryName\"', 'ASC')
            .take(pageSettings.pageSize)
            .skip(pageSettings.pageSize * indexPage)

        const sql = await query.getSql()

        console.log('SQL OUT: ', sql)

        const services = await query.getMany()
```

The SQL as defined by .getSql() (Which works):
```sql
SELECT 
    "service"."id" AS "service_id", 
    "service"."created" AS "service_created", 
    "service"."updated" AS "service_updated", 
    "service"."notes" AS "service_notes", 
    "service"."active" AS "service_active", 
    "service"."associateFee" AS "service_associateFee", 
    "service"."governmentFee" AS "service_governmentFee", 
    "service"."miscMargin" AS "service_miscMargin", 
    "service"."chainGangMargin" AS "service_chainGangMargin", 
    "service"."typeId" AS "service_typeId", 
[...]

Issue #2736, opened 2018-08-29
SelectQueryBuilder addSelect should check for duplicate field 


alex-osman:
**Issue type:**

[ ] question
[x] bug report
[x] feature request
[ ] documentation issue

**Database system/driver:**

[ ] `cordova`
[ ] `mongodb`
[x] `mssql`
[ ] `mysql` / `mariadb`
[ ] `oracle`
[ ] `postgres`
[ ] `sqlite`
[ ] `sqljs`
[ ] `react-native`
[ ] `expo`

**TypeORM version:**

[ ] `latest`
[ ] `@next`
[x] `0.2.7` (or put your version here)

**Steps to reproduce or a small repository showing the problem:**

If you run `addSelect(fieldname)` to a query that is already selecting that field, it will throw 
```
QueryFailedError: ER_DUP_FIELDNAME: Duplicate column name 'order_created_at'
```
For example.
```ts 
const qs = this.orderRepository.createQueryBuilder('order')
    .leftJoinAndSelect('order.shippingLabel', 'shippingLabel')
...
...
qs.addSelect('order.created_at') // add a field that is already included in the query
qs.getMany() // throws ER_DUP_FIELDNAME error
```

The reason I need this is because I am sometimes doing
```ts
const qs = this.orderRepository.createQueryBuilder('order').select(['order.id'])
...
...
const toOrderBy = 'order.created_at'
// want to order by this field
qs.addSelect(toOrderBy) 
qs.orderBy(toOrderBy)
```

If I take away the `addSelect` and only select the id I get an error:
```
QueryFailedError: ER_BAD_FIELD_ERROR: Unknown column 'distinctAlias.order_created_at' in 'field list'
```

If I could see if the QueryBuilder is already selecting the field I would ignore it.  Or this could be done internally so this error is avoided and you could add multiple of the same fields.
```ts
if (!qs.selectFields.includes(toOrderby)) {
   qs.addSelect(toOrderBy)
}
```

deeptechuz:
My issue was not selecting ```orderBy``` column..
Let's say I am ordering data by ```created_at``` so I should add column in select too.
I don't know if it's helpful or not but try it if got something similar :)

DJEDAINI:
Just add brackets between column name will bypass the issue :)

.addSelect(
            `(${alias}.${orderBy})`,
            `${orderBy}`,
        )

PR #4257, opened 2019-06-07
fix: wrong query on QueryBuilder containing skip, take and orderBy

Hossein-s:
QueryBuilder has strange behaviour when it comes to ordering rows.

I couldn't understand why TypeORM breaks fetching rows to two parts😬. first fetching ids and then fetching actual data.

By the way seems does this job wrongly. columns that are needed for ordering (outside of inner query) are not selected and then not available outside it.

Addresses #2736 and #3501.

### DIFF
--- a/src/query-builder/SelectQueryBuilder.ts
+++ b/src/query-builder/SelectQueryBuilder.ts
@@ -1783,18 +1783,36 @@ export class SelectQueryBuilder<Entity> extends QueryBuilder<Entity> implements 
             const metadata = this.expressionMap.mainAlias.metadata;
             const mainAliasName = this.expressionMap.mainAlias.name;
 
+            const distinctAlias = this.escape("distinctAlias");
             const querySelects = metadata.primaryColumns.map(primaryColumn => {
-                const distinctAlias = this.escape("distinctAlias");
                 const columnAlias = this.escape(DriverUtils.buildColumnAlias(this.connection.driver, mainAliasName, primaryColumn.databaseName));
                 if (!orderBys[columnAlias]) // make sure we aren't overriding user-defined order in inverse direction
                     orderBys[columnAlias] = "ASC";
                 return `${distinctAlias}.${columnAlias} as "ids_${DriverUtils.buildColumnAlias(this.connection.driver, mainAliasName, primaryColumn.databaseName)}"`;
             });
 
+            const innerQuery = this.clone().orderBy();
+
+            // select only columns that are necessary for orderBys
+            innerQuery.expressionMap.selects = [
+                ...selects,
+                ...metadata.primaryColumns.map(pc => {
+                    const alias = DriverUtils.buildColumnAlias(this.connection.driver, mainAliasName, pc.databaseName);
+                    if (!selects.some(s => s.aliasName === alias)) {
+                        return {
+                            selection: `${this.escape(mainAliasName)}.${this.escape(pc.databaseName)}`,
+                            aliasName: alias,
+                        } as SelectQuery;
+                    }
+                    return null;
+                })
+                .filter((s): s is SelectQuery => !!s),
+            ];
+
             rawResults = await new SelectQueryBuilder(this.connection, queryRunner)
                 .select(`DISTINCT ${querySelects.join(", ")}`)
-                .addSelect(selects)
-                .from(`(${this.clone().orderBy().getQuery()})`, "distinctAlias")
+                .addSelect(selects.map(s => `${distinctAlias}.${this.escape(s.aliasName!)}`).join(", "))
+                .from(`(${innerQuery.getQuery()})`, "distinctAlias")
                 .offset(this.expressionMap.skip)
                 .limit(this.expressionMap.take)
                 .orderBy(orderBys)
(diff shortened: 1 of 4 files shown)


pleerock:
try to add some joins and you'll understand why it does so. I don't recommend to change those areas, they are sensitive and hard to review

Hossein-s:
@pleerock 

Yes it is so harder that it seems😃 By the way it should be fixed somehow, shouldn't? 🤔

pleerock:
ORM adds order by because it is required, and you must select columns used in order by statement, so consider it as a requirement. If you don't like this requirement consider re-building your query different way.

Hossein-s:
> ORM adds order by because it is required, and you must select columns used in order by statement, so consider it as a requirement.

I didn't get you. It is a problem in the codes? It does select columns in orderBy and it's obviously a requirement.

Hossein-s:
It seems I have chosen inappropriate PR message.

Second paragraph describes a question came to my mind.
But I have not changed the behavior of querybuilder in the PR and tried to fix issues occurred with current implementation.

Hossein-s:
Something is still causing problem for this part. I have the following code but queryBuilder is not escaping ServiceIdSquare.
It's a bug or intended?

```connection.createQueryBuilder()
            .select("Service")
            .from(Service, "Service")
            .addSelect("Service.id * Service.id", "ServiceIdSquare")
            .orderBy("ServiceIdSquare");```

**Comment by anodynos** [comment]
I can confirm that this nasty bug (we recently discovered it in production) occurs only when `join`, `skip`/`take` & `orderBy` are used together.  The remedy seems to be the painful refactoring of all our queries to use "where exists" structure etc instead of joins, which is pretty awful :-( 

When all these 3 culprits are used, the ORM synthesises a query with this structure:  

```SELECT DISTINCT "distinctAlias".""Entity_alias_id" as "ids_Entity_alias_id"` from (...) ORDER BY "SomeOrderByField" ASC LIMIT 10``` but without adding the `"SomeOrderByField"` in the select clause (and exposing it from the inner query - although I add it manually with the same name). 

Postgres complains that `ERROR: for SELECT DISTINCT, ORDER BY expressions must appear in select list`. If I manually add  `"SomeOrderByField`" in the failing SQL query it all works fine.

I can't assess if its a "simple fix" or not, but it needs to be fixed... :+1:

**Comment by anodynos** [comment]
Any updates on this - its breaking a lot of functionality and there's no way around it if we want to use joins :-(

**Comment by MarKayS** [comment]
> I can confirm that this nasty bug (we recently discovered it in production) occurs only when `join`, `skip`/`take` & `orderBy` are used together. The remedy seems to be the painful refactoring of all our queries to use "where exists" structure etc instead of joins, which is pretty awful :-(
> 
> When all these 3 culprits are used, the ORM synthesises a query with this structure:
> 
> `` SELECT DISTINCT "distinctAlias".""Entity_alias_id" as "ids_Entity_alias_id"` from (...) ORDER BY "SomeOrderByField" ASC LIMIT 10 `` but without adding the `"SomeOrderByField"` in the select clause (and exposing it from the inner query - although I add it manually with the same name).
> 
> Postgres complains that `ERROR: for SELECT DISTINCT, ORDER BY expressions must appear in select list`. If I manually add `"SomeOrderByField`" in the failing SQL query it all works fine.
> 
> I can't assess if its a "simple fix" or not, but it needs to be fixed... 👍
> 
> FYI: I don't have any complex order by (only a single one), only one single join, and it fails with just `skip = 0` and `take = 10`.

We ran into this specific issue and for now resolved it by using limit and offset instead of skip and take. So far it seems to work with non-complex joins and order bys (two left joins and order bys on those nested columns).

**Comment by anodynos** [comment]
Any updates on this issue / PR? It's been 6 months and we're facing major problems using TypeORM because of this...

**Comment by rostamiani** [comment]
Still waiting for a merge...

**Review by imnotjames — CHANGES_REQUESTED** [review]
The tests aren't comprehensive enough.  With various joins this code will break - so can we add those to the tests per the previous comment by pleerock?  It sounds like they caused problems and broke some functionality so we need to verify that's working first.

**Comment by imnotjames** [comment]
> Please merge this request if it works. Anyone tested?

It seems to fix some scenarios and breaks others.

**Comment by Hossein-s** [comment]
> The tests aren't comprehensive enough. With various joins this code will break - so can we add those to the tests per the previous comment by pleerock? It sounds like they caused problems and broke some functionality so we need to verify that's working first.

Thanks for reviewing this PR.

Actually I couldn't find any problem as all the project tests were passing. Would you give some queries or (guide me to build them) that this PR will break so that I can investigate more and update PR to consider them too?

**Comment by manavjaiswal** [comment]
Having the same issue. SQL fails with Unknown column 'distinctAlias.xxx_id' in 'field list'. 
I am using querybuilder with order by and pagination. Is there any updates on this please. I request to get this resolved soon. Thanks.

**Comment by imnotjames** [comment]
Here's an extra test which shows that this PR breaks other cases.

```
                // create some users
      
      
                const user1 = new User("user 1");
      
      
                const user2 = new User("user 2");
      
      
                const user3 = new User("user 3");
      
      
        

      
      
                // create some services
      
      
                const service1 = new Service("service 1 for user1", user1);
      
      
                const service2 = new Service("service 2 for user1", user1);
      
      
                const service3 = new Service("service 1 for user2", user2);
      
      
                const service4 = new Service("service 1 for user3", user3);
      
      
                const service5 = new Service("service 2 for user3", user3);
      
      
        

      
      
                // store entities
      
      
                await connection.manager.save([user1, user2, user3, service1, service2, service3, service4, service5]);
      
      
        

      
      
                // create complex QueryBuilder
      
      
                const queryBuilder = connection.getRepository(User).createQueryBuilder("user");
      
      
                queryBuilder
      
      
                    .leftJoinAndSelect("user.services", "service")
      
      
                    .skip(0)
      
      
                    .take(3)
      
      
                    .addOrderBy("service.title", "DESC");
      
      
        

      
      
                console.log(queryBuilder.getQuery());
      
      
        

      
      
                const users = await queryBuilder.getMany();
      
      
        

      
      
                // should return services for user1 and user3
      
      
                expect(users).to.have.length(3);
      
      
        

      
      
                // should be in right order
      
      
                expect(users).to.eql([
      
      
                    user1,
      
      
                    user2,
      
      
                    user3,
      
      
                ]);
 ```

PR #4259, opened 2019-06-08
build: update sqlite3 package

Hossein-s:
Building current version on windows OS is failing 🤒

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "typeorm",
-  "version": "0.2.17",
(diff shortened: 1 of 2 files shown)


pleerock:
@vlapo lets say we don't need to have this change in the changelog. Is there a way to merge it and avoid being added to the changelog?

vlapo:
Yes. Only feat, fix and perf commit types are converted into changelog record.

pleerock:
Oh, okay, cool, so I'm safe to "merge pull request"?

PR #4260, opened 2019-06-08
fix: call listeners for array embeddeds

### LINKED ISSUE #4071: [MongoDB] Event listener in list of nested documents gives error
**Issue type:**

[x] question
[x] bug report
[ ] feature request
[ ] documentation issue

**Database system/driver:**

[ ] `cordova`
[x] `mongodb`
[ ] `mssql`
[ ] `mysql` / `mariadb`
[ ] `oracle`
[ ] `postgres`
[ ] `cockroachdb`
[ ] `sqlite`
[ ] `sqljs`
[ ] `react-native`
[ ] `expo`

**TypeORM version:**

[ ] `latest`
[ ] `@next`
[x] `0.2.16`

```ts
export class Category {
    @Column()
    public slug: string;

    @Column()
    public name: string;
    
    @AfterLoad()
    someFunk() {
      console.log('Loaded')
    }
}

@Entity()
export class Article {
    @ObjectIdColumn()
    public id: ObjectID;

    @Column(type => Category)
    public categories: Category[];
}
```

Gives error 
```
entity[propertyPath][this.propertyName] is not a function
```

When the entities column is not an array it works, i.e
```ts
...
@Column(type => Category)
public category: Category;
...
```

I would expect the event listeners to be called once for every instance in the array.

rustamwin:
fix #4071


test/functional/mongodb/basic/embedded-columns-listeners/entity/Post.ts
```diff
@@ -19,4 +19,7 @@ export class Post {
     @Column(type => Counters)
     counters?: Counters;
 
-}
\ No newline at end of file
+    @Column(type => Counters)
```
pleerock: It's really better to add a new separate test that covers exactly this use case

rustamwin: Yeah!

vlapo:
Is this fix only for mongodb? What about another databases? Is it works as expected for them?

rustamwin:
Not sure. I think array embeddeds doesn't supported in RDBMS? @pleerock

vlapo:
lol you are right sry :)

Issue #3847, opened 2019-03-19
FK/PK naming strategy or name override question


zelid:
**Issue type:**

[x] question
[ ] bug report
[ ] feature request
[ ] documentation issue

**Database system/driver:**

[ ] `cordova`
[ ] `mongodb`
[ ] `mssql`
[ ] `mysql` / `mariadb`
[ ] `oracle`
[x] `postgres`
[ ] `cockroachdb`
[ ] `sqlite`
[ ] `sqljs`
[ ] `react-native`
[ ] `expo`

**TypeORM version:**

[x] `latest`
[ ] `@next`
[ ] `0.x.x` (or put your version here)

**Steps to reproduce or a small repository showing the problem:**

I have some custom UI logic which takes into consideration names of foreign keys.
`TypeORM` automatically gives random FK names like `FkC6Cc2C34Bd5420Cf0F3Fe38695D` which seems to be different after each DB re-creation and similar way for PKs like `Pk51F365875Ce04758B07Bc160204`.

Having for example 
```ts
@Entity()
export class WebsiteEmail {
  @PrimaryGeneratedColumn()
  id: number

  @ManyToOne(type => Website)
  @JoinColumn()
  website: Website
  @RelationId((self: WebsiteEmail) => self.website)
  websiteId: number

  @Column()
  address: string
}
``` 

I'm interested in both FK/PK naming strategy and manual FK/PK name override. Reading documentation showed only JoinColumn name overrides or I've missed the part about FK/PK names...

How to specify FK name for `website: Website` to be `fk_website_email_website_id` and PK name for `id: number` to be `pk_website_email_id` ?

update 1:
There is some [givenName](https://github.com/typeorm/typeorm/blob/master/src/metadata/CheckMetadata.ts#L32) property in the source code. Not sure how to pass there a value from attributes yet.

Kononnable:
I don't think we've got manual name override for PKs/FKs available as some column option etc. but You can  achieve it by using custom [NamingStrategy](https://github.com/typeorm/typeorm/blob/master/docs/naming-strategy.md).

Just extend DefaultNamingStrategy and override `primaryKeyName()` and `foreignKeyName()` functions.

tkvw:
> Just extend DefaultNamingStrategy and override `primaryKeyName()` and `foreignKeyName()` functions.

[...]

Issue #1355, opened 2017-12-18
Allow explicitly named primary keys, foreign keys, and indices


timc13:
a `constraintName` option on `@Index`, `@PrimaryGeneratedColumn`, and `@JoinColumn` would be nice.

example:
```
@Entity()
class Post {
  @PrimaryGeneratedColumn({ 
    constraintName: 'pk_post_id' 
  })
  id: number

  @ManyToOne(author => Author, author => author.posts)
  @JoinColumn({ 
    constraintName: 'fk_post_author_id'
  })
  author: Author
}
```

jsuryahyd:
So, is there no way to specify a foreign key constraint_name directly? I could not find anything in the documentation. My use case is to find which foreign key constraint fails by checking `mysqlErr.sqlMessage` and send specific error messages.

vegerot:
2020, is this still not an option?

vegerot:
If having unique names is a must, then what about something similar to what you have for migrations: where there is a unique id generated in the front, then a custom name can be given after?  Best of both worlds

hugo-dutra:
I Saw it on documentation.

@ManyToOne(type => Category)
@JoinColumn({ name: "cat_id" })
category: Category;

wodka:
@hugo-dutra this is for the field name, not the name of the foreign key itself

hugo-dutra:
> @hugo-dutra this is for the field name, not the name of the foreign key itself


Strange, I had the impression of seeing a foreign key with the same name created in the database. And a field with that name created in the table with the name of the foreign key. But, thanks for the comment. Thank you.

lmjavascriptprogrammer:
Well, it seems I need this feature as well...

hvpaiva:
I need as well. 
I already can see my DBA complaining...

wenerme:
I need this as well when I see the error message

```
QueryFailedError: constraint "PK_0669fe20e252eb692bf4d344975" of relation "settings" does not exist
```

Who knows what's the meaning of `PK_0669fe20e252eb692bf4d344975` ?

amirh715:
I can't believe I don't have this option. Almost 3 years have passed! Any response?! :-/

juanjalvarez:
Can't use auto-migrations for my use-case, which means the FK constraints expected by TypeORM won't match up with what I called them, which means no lazy loading :(

kaykhancheckpoint:
[...]

PR #4274, opened 2019-06-11
feat: Add referenced table metadata to namingstrategy to resolve foreignKeyName

tkvw:
@see https://github.com/typeorm/typeorm/issues/3847 and https://github.com/typeorm/typeorm/issues/1355

### DIFF
--- a/src/naming-strategy/NamingStrategyInterface.ts
+++ b/src/naming-strategy/NamingStrategyInterface.ts
@@ -60,7 +60,7 @@ export interface NamingStrategyInterface {
     /**
      * Gets the name of the foreign key.
      */
-    foreignKeyName(tableOrName: Table|string, columnNames: string[]): string;
+    foreignKeyName(tableOrName: Table|string, columnNames: string[], referencedTablePath?: string, referencedColumnNames?: string[]): string;
 
     /**
      * Gets the name of the index - simple and compose index.
--- a/src/driver/cockroachdb/CockroachQueryRunner.ts
+++ b/src/driver/cockroachdb/CockroachQueryRunner.ts
@@ -505,7 +505,7 @@ export class CockroachQueryRunner extends BaseQueryRunner implements QueryRunner
         // rename foreign key constraints
         newTable.foreignKeys.forEach(foreignKey => {
             // build new constraint name
-            const newForeignKeyName = this.connection.namingStrategy.foreignKeyName(newTable, foreignKey.columnNames);
+            const newForeignKeyName = this.connection.namingStrategy.foreignKeyName(newTable, foreignKey.columnNames, foreignKey.referencedTableName, foreignKey.referencedColumnNames);
 
             // build queries
             upQueries.push(new Query(`ALTER TABLE ${this.escapePath(newTable)} RENAME CONSTRAINT "${foreignKey.name}" TO "${newForeignKeyName}"`));
@@ -702,7 +702,7 @@ export class CockroachQueryRunner extends BaseQueryRunner implements QueryRunner
                     // build new constraint name
                     foreignKey.columnNames.splice(foreignKey.columnNames.indexOf(oldColumn.name), 1);
                     foreignKey.columnNames.push(newColumn.name);
-                    const newForeignKeyName = this.connection.namingStrategy.foreignKeyName(clonedTable, foreignKey.columnNames);
+                    const newForeignKeyName = this.connection.namingStrategy.foreignKeyName(clonedTable, foreignKey.columnNames, foreignKey.referencedTableName, foreignKey.referencedColumnNames);
 
                     // build queries
                     upQueries.push(new Query(`ALTER TABLE ${this.escapePath(table)} RENAME CONSTRAINT "${foreignKey.name}" TO "${newForeignKeyName}"`));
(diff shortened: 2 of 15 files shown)


src/naming-strategy/NamingStrategyInterface.ts
```diff
@@ -60,7 +60,7 @@ export interface NamingStrategyInterface {
     /**
      * Gets the name of the foreign key.
      */
-    foreignKeyName(tableOrName: Table|string, columnNames: string[]): string;
+    foreignKeyName(tableOrName: Table|string, columnNames: string[], referencedTablePath: string, referencedColumnNames: string[]): string;
```
pleerock: we need to mark them with `?` to prevent breaking release

pleerock:
Also would be great to have a test for this use case :) Also please add a PR title according to our standards. Thanks :)

tkvw:
Fixed conversation, rebased commit and added test

kedar-moghe:
Will this give the ability to give our own custom constraint names, or will it be generated by the scheme in the code?

tkvw:
> Will this give the ability in the future to give our own custom constraint names, or will it just be generated by the scheme in the code?

No, that is already possible using the namingstrategy. This PR just add some additional arguments to generate a foreignkey name based on referenced columns. This is not possible at te moment.

pleerock:
merged, sorry for late response here

tdnghia98:
> namingstrategy

Is there any support for this? It could be really useful when debugging, having generic constraints name is confusing

FernandoRicardez:
So it is posible to change the FK_0669fe20e252eb692bf4d344975, To something custom or at least more readable if it is some documentation would be neat!

kaykhancheckpoint:
For those struggling with this, it is possible to explicitly set the name of FK constraints by building a naming strategy.

**CustomNamingStrategy**

The following naming convention is used below: `<table1>_<table2>_<table1_column>` then i hash this because it could get long and mysql will throw an identifier too long error.

This method makes the foreign keys deterministic so that i can use the same migrations for both test and production db.

```
import { DefaultNamingStrategy, Table, NamingStrategyInterface } from "typeorm";
import crypto from "crypto";

export class CustomNamingStrategy extends DefaultNamingStrategy implements NamingStrategyInterface {
    foreignKeyName(tableOrName: Table | string, columnNames: string[], referencedTablePath?: string, referencedColumnNames?: string[]): string {
        
        tableOrName =
            typeof tableOrName === "string" ? tableOrName : tableOrName.name;

        const name = columnNames.reduce(
            (name, column) => `${name}_${column}`,
            `${tableOrName}_${referencedTablePath}`,
        );

        return`fk_${crypto.createHash('md5').update(name).digest("hex")}`
    }
}
```


**ormconfig.ts**

Then import the custom naming strategy in your config.

```
module.exports = {
    "type": "mysql",
    "host": process.env.DB_HOST,
    ...
    namingStrategy: new CustomNamingStrategy()
}
```

references:

https://github.com/typeorm/typeorm/tree/master/sample/sample12-custom-naming-strategy
https://github.com/typeorm/typeorm/issues/5286
https://github.com/typeorm/typeorm/pull/4274/files

emtiajium:
> For those struggling with this, it is possible to explicitly set the name of FK constraints by building a naming strategy.
> 
> **CustomNamingStrategy**
> 
> The following naming convention is used below: `<table1>_<table2>_<table1_column>` then i hash this because it could get long and mysql will throw an identifier too long error.
> 
> This method makes the foreign keys deterministic so that i can use the same migrations for both test and production db.
> 
> ```
> import { DefaultNamingStrategy, Table, NamingStrategyInterface } from "typeorm";
> import crypto from "crypto";
> 
> export class CustomNamingStrategy extends DefaultNamingStrategy implements NamingStrategyInterface {
>     foreignKeyName(tableOrName: Table | string, columnNames: string[], referencedTablePath?: string, referencedColumnNames?: string[]): string {
>         
>         tableOrName =
>             typeof tableOrName === "string" ? tableOrName : tableOrName.name;
> 
>         const name = columnNames.reduce(
>             (name, column) => `${name}_${column}`,
>             `${tableOrName}_${referencedTablePath}`,
>         );
> 
>         return`fk_${crypto.createHash('md5').update(name).digest("hex")}`
>     }
> }
> ```
> 
> **ormconfig.ts**
> 
> Then import the custom naming strategy in your config.
> 
> ```
> module.exports = {
>     "type": "mysql",
>     "host": process.env.DB_HOST,
>     ...
>     namingStrategy: new CustomNamingStrategy()
> }
> ```
> 
> references:
> 
> https://github.com/typeorm/typeorm/tree/master/sample/sample12-custom-naming-strategy #5286 https://github.com/typeorm/typeorm/pull/4274/files

Which TypeORM version are you using? I use v0.2.32, and it simply ignores my custom naming strategy. 

**My codes**
https://github.com/emtiajium/ielts-gibberish/blob/b5717f4a52b9f24ead1e11fc9220f1c3e127b4c0/src/common/persistence/DatabaseModule.ts#L46

https://github.com/emtiajium/ielts-gibberish/blob/0e8703b52dc8dd8fd7a4c17bd0eb7f63b49c838e/src/common/persistence/DatabaseNamingStrategy.ts#L16

Issue #4252, opened 2019-06-06
Single Table Inheritance with composite primary key not issuing correct SQL SELECT statement


aolvictim:
**Issue type:**

[ ] question
[X] bug report
[ ] feature request
[ ] documentation issue

**Database system/driver:**

[ ] `cordova`
[ ] `mongodb`
[ ] `mssql`
[ ] `mysql` / `mariadb`
[ ] `oracle`
[X] `postgres`
[ ] `cockroachdb`
[X] `sqlite`
[ ] `sqljs`
[ ] `react-native`
[ ] `expo`

**TypeORM version:**

[ ] `latest`
[ ] `@next`
[X] `0.2.17` (or put your version here)

**Steps to reproduce or a small repository showing the problem:**

Please take a look at the following code:

```typescript
// Person
@TableInheritance({ column: { type: "varchar", name: "type" } })
export class Person {

    @PrimaryGeneratedColumn('uuid')
    id: string;

    @PrimaryColumn()
    type: string; // type property is part of the composite primary key along with id

    @Column()
    name: string;
}

// User
@ChildEntity()
export class User extends Person { }

// create an user
let u1 = new User();
u1.name = "John Doe";
await connection.getRepository('User').save(u1);
```

I am using SQLite to test the code. Because SQLite does not return the inserted entry so `typeorm` issues an additional `SELECT` statement to retrieve the data just inserted. But look at the SQL log below you can see that the `SELECT` statement is incorrect.

```SQL
INSERT INTO "person"("id", "name", "type") VALUES (?, ?, ?) -- PARAMETERS: ["931f28ad-752d-48e2-b077-4c51755a535e","John Doe","User"] // this is correct

SELECT "User"."id" AS "User_id", "User"."type" AS "User_type" FROM "person" "User" WHERE  "User"."type" IN (?) -- PARAMETERS: ["User"] // this is NOT correct. "User"."id" = "931f28ad-752d-48e2-b077-4c51755a535e" should be part of the WHERE clause here
```

I dug a bit into the source code and found the problem is in the `insert` function in [ReturningResultsEntityUpdator.ts](https://github.com/typeorm/typeorm/blob/0.2.17/src/query-builder/ReturningResultsEntityUpdator.ts#L121):
```typescript
const entityId = metadata.getEntityIdMap(entity)!;
``` 
The `entityId` ends up being `undefined` while it suppose to be 
```javascript
{
    "id": "931f28ad-752d-48e2-b077-4c51755a535e",
[...]

PR #4275, opened 2019-06-11
fixed incorrect WHERE caused by type as primary key in STI

aolvictim:
This is a potential fix for the issue https://github.com/typeorm/typeorm/issues/4252

Basically it checks if the discriminator value is present in the entity, if not, add it.

### DIFF
--- a/src/metadata/EntityMetadata.ts
+++ b/src/metadata/EntityMetadata.ts
@@ -570,7 +570,11 @@ export class EntityMetadata {
         if (!entity)
             return undefined;
 
-        return EntityMetadata.getValueMap(entity, this.primaryColumns, { skipNulls: true });
+        let fullEntity = entity;
+        if (this.discriminatorColumn && !entity.hasOwnProperty(this.discriminatorColumn.databasePath)){
+            fullEntity[this.discriminatorColumn.databasePath] = this.discriminatorValue;
+        }
+        return EntityMetadata.getValueMap(fullEntity, this.primaryColumns, { skipNulls: true });
     }
 
     /**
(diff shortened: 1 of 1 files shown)


YegorZaremba:
It'll be better if you write a tests for this case

pleerock:
First of all we need to narrow down issue scope. Is this issue only when you use STI? Can we confirm it by having a test for this case (your issue without STI, the most simple case). If issue exist then this fix isn't correct. The reason I ask is because we need to fix a source of problem, which is hard to find in this case.

imnotjames (changes requested):
This needs tests to confirm that the behavior you expect from this code is exhibited.

src/metadata/EntityMetadata.ts
```diff
@@ -570,7 +570,11 @@ export class EntityMetadata {
         if (!entity)
             return undefined;
 
-        return EntityMetadata.getValueMap(entity, this.primaryColumns, { skipNulls: true });
+        let fullEntity = entity;
+        if (this.discriminatorColumn && !entity.hasOwnProperty(this.discriminatorColumn.databasePath))
+            fullEntity[this.discriminatorColumn.databasePath] = this.discriminatorValue;
```
imnotjames, line 575: Not sure I understand why you've assigned this variable and then are changing the prop in this variable.  Could you explain it to me?

If it's an object it's the same object in both variables - you haven't cloned it - so you might as well just `entity[this.discriminatorColumn.databasePath] = this.discriminatorValue;`

AlexMesser:

PR #4290, opened 2019-06-14
Fix: change CURRENT_TIMESTAMP to DEFAULT so that time precision is pr…

VolatileBit:
…eserved when updating entities

Currently date-time precision is lost on updates for the UpdateDateColumn.

image

When setting up the entity, a precision of 6 was configured, but as it can be seen in the above query, entities having **version > 1** will lose 3 decimal places of precision.

Found this problem because when running local tests that have fast inserts, some updated entities will fall into the same seconds for UpdateDateColumn, which causes inconsistencies when sorting them by time, the order of updates cannot be distinguished within the same second.

Looking at the original comment, for performance issues CURRENT_TIMESTAMP(6) was not used anymore. But if tables have default `updatedDate datetime(6) DEFAULT CURRENT_TIMESTAMP(6)` then we should be okay to use DEFAULT, which will preserve precision in the cases of `updatedDate datetime(3) DEFAULT CURRENT_TIMESTAMP(3)`.

### DIFF
--- a/src/query-builder/UpdateQueryBuilder.ts
+++ b/src/query-builder/UpdateQueryBuilder.ts
@@ -427,7 +427,7 @@ export class UpdateQueryBuilder<Entity> extends QueryBuilder<Entity> implements 
             if (metadata.versionColumn)
                 updateColumnAndValues.push(this.escape(metadata.versionColumn.databaseName) + " = " + this.escape(metadata.versionColumn.databaseName) + " + 1");
             if (metadata.updateDateColumn)
-                updateColumnAndValues.push(this.escape(metadata.updateDateColumn.databaseName) + " = CURRENT_TIMESTAMP"); // todo: fix issue with CURRENT_TIMESTAMP(6) being used, can "DEFAULT" be used?!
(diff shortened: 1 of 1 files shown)


VolatileBit:
CI/CD fails with a syntax error. I am a bit lost here, I may need help fixing this.

pleerock:
Looks like at least sqlite3 doesn't support this syntax. Maybe you can try to google solution for sqlite?

pleerock:
Can you please add test as well?

mpvosseller:
Ran into this today too. If I quickly create and update an entity the `updatedDate` can end up being before the `createdDate` (which was also the initial value of `updatedDate`)

I'm using MySQL and all the defaults like:
```javascript
  @CreateDateColumn()
  createdDate!: Date

  @UpdateDateColumn()
  updatedDate!: Date
```

The db types are `datetime(6)` and the defaults are `CURRENT_TIMESTAMP(6)`. 
When the entity is created `createdDate` and `updatedDate` are given a `CURRENT_TIMESTAMP(6)` value like `2020-08-24 15:45:45.140696`.
Then when I update the entity the `updatedDate` is set to a `CURRENT_TIMESTAMP` value like `2020-08-24 15:45:45.000000` which is now before the `createdDate` and the original `updatedDate`

imnotjames:

Issue #4285, opened 2019-06-13
Unstable mongodb test


vlapo:
**Issue type:**

[x] bug report

**Database system/driver:**

[x] `mongodb`

**TypeORM version:**

[x] `latest`

**Steps to reproduce or a small repository showing the problem:**

We have one unstable test in mongodb. I saw it more then one time. Log from test:
```
1) mongodb > timestampable columns
       should persist timestampable columns:
      AssertionError: expected 1560453732935 to equal 1560453732936
      + expected - actual
      -1560453732935
      +1560453732936
      
      at Object.<anonymous> (test/functional/mongodb/basic/timestampable-columns/timestampable-columns.ts:35:45)
      at step (node_modules/tslib/tslib.js:133:27)
      at Object.next (node_modules/tslib/tslib.js:114:57)
      at fulfilled (node_modules/tslib/tslib.js:104:62)
      at process._tickCallback (internal/process/next_tick.js:68:7)
```
Link: https://travis-ci.org/typeorm/typeorm/jobs/545410478

@rustamwin as most active mongodb contributor could you check it ? :)

vlapo:
Merged.

PR #4292, opened 2019-06-14
test(mongodb): time comparing is stable

YegorZaremba:
Test has `+/-` delta range of 5 milliseconds, because earlier this test fell due to the difference of 1 millisecond.

It fixes issue #4285


pleerock:
Thank you!

PR #4296, opened 2019-06-16
fix: length limit for eager join table alias in FindOptions

aethys256:
Hello!

### Context

This is a follow-up of the PR https://github.com/typeorm/typeorm/pull/4182 which itself was a follow-up of the PR https://github.com/typeorm/typeorm/pull/3514.

The issue is still the same except it's happening in join alias from relations defined as `eager` instead of join alias (defined in FindOption `relations` array) resulting in a `ERROR:  table name "CategoryWithVeryLongName__postsWithVeryLongName__authorWithVery" specified more than once` due to the fact names are truncated.

### Solution

The way I've implemented a fix is similar to what I did in the previous PR, i.e. by re-using @VinceOPS `shorten()`. This time I noticed the eager alias were generated from the --not-released-- NamingStrategy class, I added a method so both type of relations would be handled there.

### Tests

I did not write a test about it, I guess updating the previous one to add an eager relation would do the trick. I only tested it on Postgres but considering it's re-using existing Driver's option `maxAliasLength` it should be safe.

### Comments

Let me know if there is anything else I should do in order to get the PR merged.  
We are currently maintaining a fork of TypeORM with this fix on our apps until it's being reviewed and merged.

PS: In the previous PR I commented about updating manually the changelog, since then I noticed that conventional commit was adopted. Since my commits are respecting it, should I still update it manually or all commits will be picked-up automatically ?

### DIFF
--- a/src/naming-strategy/NamingStrategyInterface.ts
+++ b/src/naming-strategy/NamingStrategyInterface.ts
@@ -120,5 +120,10 @@ export interface NamingStrategyInterface {
     /**
      * Gets the name of the alias used for relation joins.
      */
-    eagerJoinRelationAlias(alias: string, propertyPath: string): string;
+    joinRelationAlias(alias: string, propertyPath: string, maxAliasLength?: number): string;
+
+    /**
+     * Gets the name of the alias used for eager relation joins.
+     */
+    eagerJoinRelationAlias(alias: string, propertyPath: string, maxAliasLength?: number): string;
 }
--- a/src/find-options/FindOptionsUtils.ts
+++ b/src/find-options/FindOptionsUtils.ts
@@ -3,7 +3,6 @@ import {FindOneOptions} from "./FindOneOptions";
 import {SelectQueryBuilder} from "../query-builder/SelectQueryBuilder";
 import {FindRelationsNotFoundError} from "../error/FindRelationsNotFoundError";
 import {EntityMetadata} from "../metadata/EntityMetadata";
-import {shorten} from "../util/StringUtils";
 
 /**
  * Utilities to work with FindOptions.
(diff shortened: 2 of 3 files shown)


joemaidman:
Thanks for this fix @aethys256, we just hit this one. Can we get this one reviewed?

src/naming-strategy/NamingStrategyInterface.ts
```diff
@@ -120,5 +120,10 @@ export interface NamingStrategyInterface {
     /**
      * Gets the name of the alias used for relation joins.
      */
-    eagerJoinRelationAlias(alias: string, propertyPath: string): string;
+    joinRelationAlias(alias: string, propertyPath: string, maxAliasLength?: number): string;
+
+    /**
+     * Gets the name of the alias used for eager relation joins.
+     */
+    eagerJoinRelationAlias(alias: string, propertyPath: string, maxAliasLength?: number): string;
```
Kononnable: Tbh. I'm not a fan of adding this to NamingStrategy - I think there should be names which are stored in the db, not aliases used in queries(which are runtime only). However `eagerJoinRelationAlias` is there already.

aethys256: I was not a fan either, although since there were one about it already I figured it would be less misleading to have both.
We could split it up in two parts, i.e. the one about the representation in the DB and the one about querying.

kshitij-srv:
Can this change be merged? I have a similar issue arising due to eagerly loaded relations.

traviscollins:
I'm running into this issue when using TypeOrm with a GraphQL service. The aliases used by the eager joins are far longer than postgres allows for uniqueness (63 characters). Its super easy to create a GraphQL query that causes joins names to be longer than 63 characters.

imnotjames (changes requested):
We need tests to verify this behavior.

imnotjames:

PR #4298, opened 2019-06-17
[dnm] cockroachdb: use numbers for int2/int4

jordanlewis:
Previously, CockroachDB represented all integer widths as 64-bit
integers to the client. CockroachDB's 19.2 release will remove this
limitation.

This commit upgrades the test suite to expect numbers, not strings, for
int2 and int4.


jordanlewis:
Hey @pleerock, this shouldn't be merged as such because it won't work with CockroachDB 19.1 and below. How would I go about making this change for just newer versions of CRDB?

Thanks!

vlapo:
We do not have a proper way how to handle different DB versions right now. I hope we will introduce new possibilities in driver in next versions of typeorm and enable this kind of changes.

AlexMesser:
Hi @jordanlewis. For now we dont have any mechanism to auto-detect users driver version. The only way is to introduce a new flag/parameter in connection settings, where user can define its crdb version. I plan to implement this on the weekend.

vlapo:
@AlexMesser I dont think this is the best way how to do this. Each DB has some function to detect DB version. We should automatically load this version in `Driver` and handle this without another connection option. @Kononnable introduce some idea about refactoring on slack. I think we should include this change into refactoring.

Kononnable:
This might not be the nicest way, but this probably is the easiest one. We don't have such mechanism in place, so developing it would probably take some time.
I believe returning numbers as strings can be a big problem to end-users - unless we're converting it to numbers in hydration process. If we're currently returning strings we could add a flag in connection options and return to this matter when 'versioning mechanism' is implemented. If we're converting it to numbers based on our metadata(I don't recall the implementation right now) then probably some dirty patch in code instead of connection option should be implemented.
Both solutions requires to take care of this later(fix won't be permanent), but leaving it as it is for some time might make more problems to end-users(and us) - it might be best to deal with it as soon as possible(even if implementation isn't final).

AlexMesser:
@jordanlewis, I tried a crdb v19.2.0-alpha.20190606, but I still getting strings instead of numbers for `int2` and `int4`. Is it right version?

jordanlewis:
@AlexMesser it didn't make it into that one. Was merged on June 14th.

imnotjames:
Thanks for the contribution.  Unfortunately, the two tests changed here are causing failures & have conflicts.  As it's been a year since there was any activity I'm gonna close this out.

PR #4303, opened 2019-06-17
refactor(cockroachdb): redundant executeQuery()

YegorZaremba:
`executeQuery` method is not used in `cockroachdb` driver

redundant-executeQuery


vlapo:
Thank you for code cleanup.

PR #4308, opened 2019-06-18
docs: add explanation ManyToMany with custom properties

maxsommer:
Since I myself ran into the issue and had to rewrite my code based on this solution I hope it may be helpful for others as well 😊

### DIFF
--- a/docs/many-to-many-relations.md
+++ b/docs/many-to-many-relations.md
@@ -169,3 +169,35 @@ const categoriesWithQuestions = await connection
     .leftJoinAndSelect("category.questions", "question")
     .getMany();
 ```
+
+In case you need to have additional properties to your ManyToMany relationship you should create a new Entity yourself. For example if you would like entities A and B to have a ManyToMany relationship with an `order` property associated to it you want to create entity C like the following:
+```typescript
+import { Entity, Column, ManyToOne, PrimaryGeneratedColumn } from "typeorm";
+import { A } from "./a";
+import { B } from "./b";
+
+@Entity()
+export class C {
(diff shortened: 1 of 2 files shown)


vlapo:
Hi @maxsommer. Thank you for your contribution. We have small note in FAQ section about this topic. https://typeorm.io/#/faq/how-do-i-add-extra-columns-into-many-to-many-junction-table So feel free extend this section of docs and add link from https://typeorm.io/#/relations-faq and https://typeorm.io/#/many-to-many-relations

About your example. Please use more descriptive names of entities. Maybe some real-life scenario like relation between Post and Tag and extra column with `createdAt` or something like that. Such a case is easier to understand.

image

maxsommer:
Alright, that's good to know – I did not notice that. I'll follow your advice and extend a little further and make the example more graspable. Thanks for the very quick and constructive feedback 😊

docs/many-to-many-relations.md
```diff
@@ -169,3 +169,40 @@ const categoriesWithQuestions = await connection
...
+    @ManyToOne(type => Category, category => category.postToCategories)
+    public category!: Category;
+}
+```
+
+Additionally you will have to add a relationship like the following to `Post` and `Category`:
+
+```typescript
+@OneToMany((type) => PostToCategory, (postToCategory) => postToCategory.post)
+public postToCategories!: number;
```
vlapo: `PostToCategory[]` instead of `number` ?
```ts
public postToCategories!: PostToCategory[];
```

maxsommer:
Thank you! Was my first open source PR so I'm very happy🙃

Issue #3309, opened 2018-12-21
FindOptions uses same parameter for different where options


jannikkeye:
**Issue type:**

[x] bug report

**Database system/driver:**
[x] `postgres`

**TypeORM version:**

[x] `@next`

**Steps to reproduce or a small repository showing the problem:**

I have a simple entity called `Reports`which is related to another entity by a ManyToOne-Relation. I am trying to load some reports based on the id of the related entity and another `archer`field.
Note that I am using find with findOptions, which is probably the issue here.

```typescript
const reports = await Report.find({
      where: {
        process: {
          id: processId
        },
        type
      },
      order: {
        createdAt: "DESC"
      }
})
```

This query is currently being generated and executed:

```sql
SELECT
    "Report"."id" AS "Report_id",
    "Report"."createdAt" AS "Report_createdAt",
    "Report"."updatedAt" AS "Report_updatedAt",
    "Report"."processId" AS "Report_processId",
    "Report"."type" AS "Report_type",
    "Report"."totalEvents" AS "Report_totalEvents",
    "Report"."newEvents" AS "Report_newEvents",
    "Report"."newApplicants" AS "Report_newApplicants",
    "Report"."totalApplicants" AS "Report_totalApplicants",
    "Report"."newFirstPriorityApplications" AS "Report_newFirstPriorityApplications",
    "Report"."totalFirstPriorityApplications" AS "Report_totalFirstPriorityApplications",
    "Report"."newSecondPriorityApplications" AS "Report_newSecondPriorityApplications",
    "Report"."totalSecondPriorityApplications" AS "Report_totalSecondPriorityApplications",
    "Report"."newThirdPriorityApplications" AS "Report_newThirdPriorityApplications",
    "Report"."totalThirdPriorityApplications" AS "Report_totalThirdPriorityApplications",
    "Report"."newOtherPriorityApplications" AS "Report_newOtherPriorityApplications",
    "Report"."totalOtherPriorityApplications" AS "Report_totalOtherPriorityApplications"
FROM
    "report" "Report"
    INNER JOIN "process" "Report_process" ON "Report_process"."id" = "Report"."processId"
    WHERE ("Report_process"."id" = $1
        AND "Report"."type" = $1)
ORDER BY
    "Report"."createdAt" DESC -- PARAMETERS: ["475a9cbe-17d5-41fd-9067-776da2443f81","application"]
```

[...]

PR #4310, opened 2019-06-18
WIP fix(query-builder): where conditions parameter

YegorZaremba:
TypeORM version **@next**

Fixes: #3309

Earlier `parameterIndex ` var was used at the very beginning of the `buildWhere` method, but should be in the loop
Because when this loop call `this.buildWhere()` there are two separate `parameterIndex` variables

But `parameterIndex` also used in this part of code:

```js
this.expressionMap.nativeParameters[parameterName + realParameterValueIndex] = realParameterValue;
parameterIndex++;
parameters.push(this.connection.driver.createParameter(parameterName + realParameterValueIndex, parameterIndex - 1));
```

Just want to run test by ci (my local machine is so weak) and recheck functionality one more time, I definitely write one more test for code above if it not exists

Also I have some questions:
1) Should I keep comment in the code or it looks messy?
2) Should I update `changelog`?

### DIFF
--- a/src/query-builder/SelectQueryBuilder.ts
+++ b/src/query-builder/SelectQueryBuilder.ts
@@ -2318,7 +2318,6 @@ export class SelectQueryBuilder<Entity> extends QueryBuilder<Entity> implements 
 
     protected buildWhere(where: any, metadata: EntityMetadata, alias: string, embedPrefix?: string): string {
         let condition: string = "";
-        let parameterIndex = Object.keys(this.expressionMap.nativeParameters).length;
         if (where instanceof Array) {
             condition = ("(" + where.map(whereItem => {
                 return this.buildWhere(whereItem, metadata, alias, embedPrefix);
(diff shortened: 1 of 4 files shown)


YegorZaremba:
hold branch  because of #4311

Kononnable:
Ad. 1 - Yes, you should remove that comment - it says why have moved the line, but after a month from merging we won't need to know where it was before and why it was moved - but we still could get this information from `git blame`.
You can comment code on PR why some specific change was made(like [here](https://github.com/typeorm/typeorm/pull/3977/files#r274307949)).

Ad.2. - In master there was a change to generate a chanelog based on commit history. It wasn't merged to next yet, so I guess we still should update it manually.

YegorZaremba:
Thanks for answers, but I closed PR
Yes, my PR fixes this issue, but after read commit history/debugging/passing tests for a 3 hours I still not sure that my changes don't break exist functionality

PR #4314, opened 2019-06-18
fix: make oracle tests on next branch work again

Kononnable:
Some of the fixes are based on current implementation on master branch.

I think some of the find options(regarding order of results) tests are non deterministic right now, but I'm not yet sure which ones(and if this is a test fault of bug in functionality).


src/driver/oracle/OracleQueryRunner.ts
```diff
@@ -1429,7 +1432,7 @@ export class OracleQueryRunner extends BaseQueryRunner implements QueryRunner {
         if (column.isNullable !== true && !column.isGenerated) // NOT NULL is not supported with GENERATED
             c += " NOT NULL";
         if (column.isGenerated === true && column.generationStrategy === "increment")
-            c += " GENERATED ALWAYS AS IDENTITY";
+            c += " GENERATED BY DEFAULT AS IDENTITY";
```
Kononnable: https://www.oracletutorial.com/oracle-basics/oracle-identity-column/
If we want to be able to override default generated value(identity insert) we have to use `BY DEFAULT`

src/query-builder/QueryBuilder.ts
```diff
@@ -730,10 +730,10 @@ export abstract class QueryBuilder<Entity> {
 
         } else {
             const [primaryColumn] = metadata.primaryColumns;
-            const areAllNumbers = normalized.every((id: any) => typeof id === "number");
+            const normalizedValues = normalized.map(ent => ent[Object.keys(ent)[0]]);
+            const areAllNumbers = normalizedValues.every((id: any) => typeof id === "number");
             if (areAllNumbers) {
-                return `${alias + this.escape(primaryColumn.databaseName)} IN (${normalized.join(", ")})`;
-
+                return `${alias + this.escape(primaryColumn.databaseName)} IN (${normalizedValues.join(", ")})`;
```
Kononnable: This was a dead code. Still not sure if we should put number parameters inside query string. Parametrized queries can use already compiled query plans(performance).

src/query-builder/SelectQueryBuilder.ts
```diff
@@ -2341,7 +2341,7 @@ export class SelectQueryBuilder<Entity> extends QueryBuilder<Entity> implements
                 if (column) {
 
                     const aliasPath = `${alias}.${propertyPath}`;
-                    const parameterName = alias + "_" + propertyPath.replace(".", "_") + "_" + parameterIndex;
+                    const parameterName = alias + "_" + propertyPath.split(".").join("_") + "_" + parameterIndex;
```
Kononnable: Replace all dots with underscores - there was a problem with STI fields.

test/functional/table-inheritance/single-table/database-option-inherited/database-option-inherited.ts
```diff
@@ -6,7 +6,8 @@ describe("table-inheritance > single-table > database-option-inherited", () => {
 
     let connections: Connection[];
     before(async () => connections = await createTestingConnections({
-        entities: [__dirname + "/entity/*{.js,.ts}"]
+        entities: [__dirname + "/entity/*{.js,.ts}"],
+        enabledDrivers: ["postgres", "cockroachdb", "mariadb", "mssql", "mysql", "sqlite", "sqljs"]
```
Kononnable: I think on Oracle we have to unmount database first before creating another one - however creating more databases isn't always possible(e.g. it's blocked on official docker images) so we probably should just disable that test for oracle.

test/github-issues/2518/issue-2518.ts
```diff
@@ -10,7 +10,8 @@ describe("github issues > #2518 TreeRepository.findDescendantsTree does not load
     before(
         async () =>
         (connections = await createTestingConnections({
-            entities: [__dirname + "/entity/*{.js,.ts}"]
+            entities: [__dirname + "/entity/*{.js,.ts}"],
+            enabledDrivers: ["postgres", "cockroachdb", "mariadb", "mssql", "mysql", "sqlite", "sqljs"]
```
Kononnable: Not compatible data type(text) on one of the entity columns.

test/github-issues/3363/issue-3363.ts
```diff
@@ -17,6 +18,10 @@ describe("github issues > #3363 Isolation Level in transaction() from Connection
 
     it("should execute operations in READ UNCOMMITED isolation level", () => Promise.all(connections.map(async function(connection) {
 
+        if (connection.driver instanceof OracleDriver) {
```
Kononnable: Oracle doesn't support that transaction isolation level.

test/github-issues/2518/issue-2518.ts
pleerock: we should left such comments inside code

pleerock:
please copy your comments into code and merge PR on your own

PR #4321, opened 2019-06-19
docs: fix typographical error in faq

SrChip15:
change "typedi" to "typed" in how to use TypeORM with a dependency injection tool question


pleerock:
Actually I really meant `typedi` since there is an example of using typeorm with **typedi** service container library.

vlapo:
Sry. My bad 🤦‍♂

PR #4325, opened 2019-06-19
Export additional schema builder classes

michaeldzjap:
This PR exports several table related schema builder classes that were previously not accessible for use with:

```typescript
queryRunner.createCheckContraint(s);
queryRunner.createExclusionContraint(s);
queryRunner.createUniqueConstraint(s);
```


vlapo:
@pleerock are these classes mean like public? If yes we should export also `*Options` interfaces (`TableOptions`, `TableCheckOptions`, ...)

pleerock:
yes, I think we can make them public

tzellman:
Thanks for the PR @michaeldzjap!

Do we have an idea of which release this changeset will be included in? Thanks!!

Kononnable:
You can still use them without exporting directly from index file `import TableCheck from "typeorm/schema-builder/table/TableCheck"`. However it doesn't change the fact, that they might be worth exporting - no point in exporting only part of similar API functions.

Issue #4332, opened 2019-06-20
PostgreSQL 12 compatibility (lack of consrc)


luzat:
**Issue type:**
[x] bug report
[x] feature request

**Database system/driver:**
[x] `postgres`

**TypeORM version:**
[x] `latest`
[x] `@next`

**Steps to reproduce or a small repository showing the problem:**
PostgreSQL 12 removes the `pg_constraint.consrc` column. This makes `constraintsSql` from `loadTables` in [PostgresQueryRunner.ts](https://github.com/typeorm/typeorm/blob/master/src/driver/postgres/PostgresQueryRunner.ts#L1372) fail with an exception and aborting the transaction. Documentation:

* [PostgreSQL 12](https://www.postgresql.org/docs/12/catalog-pg-constraint.html)
* [PostgreSQL 11](https://www.postgresql.org/docs/11/catalog-pg-constraint.html)

The docs for 11 still include `consrc`, but recommend to use `pg_get_constraintdef`,  which only seems to be used for exclusions constraints (`x`) here. This has been introduced in #2710 for exclusions constraints. It seems that `expression` is used in some places, but giving it a very rough look I think whatever `pg_get_constraintdef` returns might work as well. Always using `pg_get_constraintdef` at least gets my 1 entity test running, but this obviously does not use many constraints.

viktor-ku:
@Kononnable I see that it's fixed in the fork of typeorm. Will it be fixed also in the main typeorm (this) repo some day?

danvy:
Is there any work around until a new release is published?

MaSydJun:
I think the only workaround right now is to build https://github.com/eliziario/typeorm yourself and use its command line tool to achieve certain goals. 

Make sure you build eliziario-typeorm first like this:
```
npm install
npm run compile
npm run package
```

After that you could implement a custom npm-script like this:
```
...
"scripts": {
  ...
  "typeorm": "node /path/to/eliziario-typeorm/build/package/cli.js",
  ...
}
...
```

Then you can use this for instance:
```
npm run typeorm -- migration:run
```

Hope this helps & works, didn't really test it...

Domiserver:
I already have a NestJS project built with Angular ready to be deployed.
How should I add this temporary solution to the project?
Thanks

stefanpartheym:
> I already have a NestJS project built with Angular ready to be deployed.
> How should I add this temporary solution to the project?
> Thanks

[...]

PR #4333, opened 2019-06-21
fix(postgres,cockroachdb): remove consrc usage

luzat:
Replace usage of pg_constraint.consrc with pg_get_constraintdef() in all
cases. PostgreSQL 12 no longer supports consrc. Use the same code in
CockroachDB for consistency.

Closes: #4332

### DIFF
--- a/src/driver/cockroachdb/CockroachQueryRunner.ts
+++ b/src/driver/cockroachdb/CockroachQueryRunner.ts
@@ -1336,7 +1336,7 @@ export class CockroachQueryRunner extends BaseQueryRunner implements QueryRunner
         }).join(" OR ");
 
         const constraintsSql = `SELECT "ns"."nspname" AS "table_schema", "t"."relname" AS "table_name", "cnst"."conname" AS "constraint_name", ` +
-            `CASE "cnst"."contype" WHEN 'x' THEN pg_get_constraintdef("cnst"."oid", true) ELSE "cnst"."consrc" END AS "expression", ` +
+            `pg_get_constraintdef("cnst"."oid") AS "expression", ` +
             `CASE "cnst"."contype" WHEN 'p' THEN 'PRIMARY' WHEN 'u' THEN 'UNIQUE' WHEN 'c' THEN 'CHECK' WHEN 'x' THEN 'EXCLUDE' END AS "constraint_type", "a"."attname" AS "column_name" ` +
             `FROM "pg_constraint" "cnst" ` +
             `INNER JOIN "pg_class" "t" ON "t"."oid" = "cnst"."conrelid" ` +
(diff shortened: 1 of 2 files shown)


vlapo:
What about postgres 9 and 10? We cannot merge this as it is. We need introduce some mechanism how to handle more DB versions.

luzat:
pg_get_constraintdef with boolean second parameter was already used previously for exclusion constraints and has been introduced in PostgreSQL 7.4 (without second parameter in 7.3). It has already been in use. The tests run fine with the PostgreSQL 9.6 Docker version provided with TypeORM, but most likely all the way down to 7.4 (at least concerning this part of the code; don't know what guarantees the rest of the code gives).

luzat:
Also, PostgreSQL since 7,.4 [recommends](https://www.postgresql.org/docs/7.4/catalog-pg-constraint.html):

> Note: consrc is not updated when referenced objects change; for example, it won't track renaming of columns. Rather than relying on this field, it's best to use pg_get_constraintdef() to extract the definition of a check constraint.

vlapo:
Sorry. I made a fast comment and did not check docs before :) I  think the most important are versions 9.4+ (lower versions are not supported https://www.postgresql.org/support/versioning/)

src/driver/cockroachdb/CockroachQueryRunner.ts
```diff
@@ -1336,7 +1336,7 @@ export class CockroachQueryRunner extends BaseQueryRunner implements QueryRunner
         }).join(" OR ");
 
         const constraintsSql = `SELECT "ns"."nspname" AS "table_schema", "t"."relname" AS "table_name", "cnst"."conname" AS "constraint_name", ` +
-            `CASE "cnst"."contype" WHEN 'x' THEN pg_get_constraintdef("cnst"."oid", true) ELSE "cnst"."consrc" END AS "expression", ` +
+            `CASE "cnst"."contype" WHEN 'c' THEN regexp_replace(pg_get_constraintdef("cnst"."oid", true), '^CHECK\\s*\\((.*)\\)\\s*$', '\\1', 'i') ELSE pg_get_constraintdef("cnst"."oid", true) END AS "expression", ` +
```
Kononnable: Why do we have to use regexp here? Can't we just update the place where this value is read?

luzat: I was trying to write a minimally invasive patch, but it's probably nicer to do the replacement later when building the `TableCheck`. Will provide a patch for that.

luzat:
Some notes:
  * the regex is somewhat more generic (not case-sensitive and allowing spaces) than it needs to be to account for possible changes in PostgreSQL's or CockroachDB's format over time
  * the previously existing comment that the expression could be problematic because of lack of escaped columns is probably obsolete now (if it ever was relevant): `pg_get_constraint_def` should return a valid definition, including escaped columns where necessary.

PodaruDragos:
is there any progress regarding this ?

xWiiLLz:
Any updates on this?

src/driver/cockroachdb/CockroachQueryRunner.ts
luzat: Can this be resolved? Replacing the regular expression by some other parsing is probably not worth it and storing the constriaint definition including `CHECK(...)` would require changes in many other places which handle constraints.

Szayet:
@Kononnable could this be merged?

vincentdchan:
Any updates on this?

pleerock:
I usually really scare to merge database-schema related changes, but this looks like got a lot of attention. Let's merge it.

Issue #3685, opened 2019-02-21
Brackets syntax failed when use where with object literal


vietthang:
**Issue type:**

[ ] question
[x ] bug report
[ ] feature request
[ ] documentation issue

**Database system/driver:**

[ ] `cordova`
[ ] `mongodb`
[ ] `mssql`
[ ] `mysql` / `mariadb`
[ ] `oracle`
[x ] `postgres`
[ ] `sqlite`
[ ] `sqljs`
[ ] `react-native`
[ ] `expo`

**TypeORM version:**

[ ] `latest`
[ ] `@next`
[ x] `0.2.12` (or put your version here)

**Steps to reproduce or a small repository showing the problem:**

When I use Brackets syntax with object literal as below:

```ts
const generatedSql = entityManager
    .createQueryBuilder()
    .from(Product, 'Product')
    .where(new Brackets(qb => qb.where({ id: 1 })))
```

I got exception:

```
(node:97360) UnhandledPromiseRejectionWarning: TypeError: Cannot read property 'hasMetadata' of undefined
    at SelectQueryBuilder.QueryBuilder.computeWhereParameter (/Users/thangnv/Develop/typeorm-report/src/query-builder/QueryBuilder.ts:763:47)
    at SelectQueryBuilder.where (/Users/thangnv/Develop/typeorm-report/src/query-builder/SelectQueryBuilder.ts:690:32)
    at Brackets.entityManager.createQueryBuilder.from.where.typeorm_1.Brackets.qb [as whereFactory] (/Users/thangnv/Develop/typeorm-report/src/test.ts:27:34)
    at SelectQueryBuilder.QueryBuilder.computeWhereParameter (/Users/thangnv/Develop/typeorm-report/src/query-builder/QueryBuilder.ts:750:19)
    at SelectQueryBuilder.where (/Users/thangnv/Develop/typeorm-report/src/query-builder/SelectQueryBuilder.ts:690:32)
    at test (/Users/thangnv/Develop/typeorm-report/src/test.ts:27:6)
    at process._tickCallback (internal/process/next_tick.js:68:7)
```

From my quick research, the issue come from this line:

```ts
if (this.expressionMap.mainAlias!.hasMetadata) {
```

Seem like omit null check in `mainAlias` causes this issue.

Thanks,

pie6k:
I have this issue too

AndKiel:
Encountered the same thing today.
Any news on this?

lucashsilva:
[...]

PR #4337, opened 2019-06-21
Fix query builder crashing when using object in .where inside Brackets

Ephys:
This is an attempt at fixing #3685 where the following piece of code

```typescript
const generatedSql = entityManager
    .createQueryBuilder()
    .from(Product, 'Product')
    .where(new Brackets(qb => qb.where({ id: 1 })))
```

would crash because the QueryBuilder could not determine the alias to use for `id`


pleerock:
This small functionality got too complex in my opinion.

tonivj5:
Any progress?

viczam:
can we get this merged?

imnotjames (changes requested):

AlexMesser:
Superseded by #5659. Anyway, thank you for contribution.

Issue #4094, opened 2019-05-06
Delete on JSON column: QueryFailedError: operator does not exist: json = unknown


christopherdiehl:
**Issue type:**

[ ] question
[X] bug report
[ ] feature request
[ ] documentation issue

**Database system/driver:**

[ ] `cordova`
[ ] `mongodb`
[ ] `mssql`
[ ] `mysql` / `mariadb`
[ ] `oracle`
[X] `postgres`
[ ] `cockroachdb`
[ ] `sqlite`
[ ] `sqljs`
[ ] `react-native`
[ ] `expo`

**TypeORM version:**

[ ] `latest`
[ ] `@next`
[ ] `0.2.16` (or put your version here)

**Steps to reproduce or a small repository showing the problem:**
```
@Entity()
export class Organization extends BaseEntity {
  // allows us to pass in an object in the constructor
  @PrimaryGeneratedColumn()
  id: number;

  @Column({ type: "text" })
  name: string;

  @Column({ type: "boolean", default: true })
  active: boolean;
  
  @Column({ type: "json", default: {} })
  metadata?: any;

  @UpdateDateColumn()
  updatedDate: Date;

  @CreateDateColumn()
  createdDate: Date;

}

const organization : Organization= {
   name : "TEST",
}
await organization.save();
await Organization.delete(organization);

```

Result: 
The generated sql uses a where clause on a JSON column.
```
    QueryFailedError: operator does not exist: json = unknown

      at new QueryFailedError (src/error/QueryFailedError.ts:9:9)
      at Query.callback (src/driver/postgres/PostgresQueryRunner.ts:176:30)
      at Query.Object.<anonymous>.Query.handleError (node_modules/pg/lib/query.js:142:17)
      at Connection.connectedErrorMessageHandler (node_modules/pg/lib/client.js:183:17)
      at Socket.<anonymous> (node_modules/pg/lib/connection.js:125:12)
```

Generated SQL: 
```
    query: DELETE FROM "organization" WHERE "name" = $1 AND "id" = $2 AND "active" = $3 AND "metadata" = $4 AND "updatedDate" = $5 AND "createdDate" = $6 -- PARAMETERS: ["tempget",14,true,{},"2019-05-07T00:11:11.053Z","2019-05-07T00:11:11.053Z"]
```

Fix: 
`await Organization.delete({id: organization.id})`

Kononnable:
You're looking for `remove()`.
[...]

Issue #4049, opened 2019-04-25
How to fetch data from two MySQL tables and There is no common field to join. So my question is how we fetch the data from multiple MySQL tables without using join methods?  


RaghunathVadlakonda:
**Issue type:**

[ ] question
[ ] bug report
[ ] feature request
[ ] documentation issue

**Database system/driver:**

[ ] `cordova`
[ ] `mongodb`
[ ] `mssql`
[ ] `mysql` / `mariadb`
[ ] `oracle`
[ ] `postgres`
[ ] `cockroachdb`
[ ] `sqlite`
[ ] `sqljs`
[ ] `react-native`
[ ] `expo`

**TypeORM version:**

[ ] `latest`
[ ] `@next`
[ ] `0.x.x` (or put your version here)

**Steps to reproduce or a small repository showing the problem:**

<!--
    To answer those questions you need to put "x" inside the square brackets, for example:
    [x] `mysql`
    [ ] `postgres`

    Also, please format your code properly (by taking code blocks into ```ts .... ```)
!>

PR #4351, opened 2019-06-24
test(4094): shouldn't throw an error during deletion

YegorZaremba:
Fixes: [#4049](https://github.com/typeorm/typeorm/issues/4094)

I cannot reproduce this issue, added test just in case


vlapo:
You have to change last line:
```ts
await expect(connection.manager.delete(Organization, result)).to.not.be.rejected;
```
Problem is with `delete` and `result` as input.

YegorZaremba:
It was hard for me to fix it, so I close this PR

Issue #4354, opened 2019-06-25
Switch from dotenv to dotenv-flow


TheAifam5:
**Issue type:**

[ ] question
[ ] bug report
[x] feature request
[ ] documentation issue

**Database system/driver:**

[ ] `cordova`
[ ] `mongodb`
[ ] `mssql`
[ ] `mysql` / `mariadb`
[ ] `oracle`
[ ] `postgres`
[ ] `cockroachdb`
[ ] `sqlite`
[ ] `sqljs`
[ ] `react-native`
[ ] `expo`

**TypeORM version:**

[x] `latest`
[ ] `@next`
[ ] `0.x.x` (or put your version here)

**Steps to reproduce or a small repository showing the problem:**

_dotenv-flow extends dotenv adding the ability to have multiple .env* files like .env.development, .env.test and .env.production, also allowing defined variables to be overwritten individually in the appropriate .env*.local file._

imnotjames:
I don't think this makes sense for a project like TypeORM, does it?

It also wouldn't allow for mutiple `.env` files - the connection option reader would need to be updated to effectively support that itself.

It only uses `dotenv` to load a VERY SPECIFIC file.

TheAifam5:
I forgot to close this issue, more info you can find the closed PR.

scbird:
For those wishing to use dotenv-flow, you can place this line at the top of your `ormconfig.js` file:
```javascript
require("dotenv-flow").config({ silent: true, purge_dotenv: true });
```

PR #4355, opened 2019-06-25
Change dotenv to dotenv-flow

TheAifam5:
Related issue: https://github.com/typeorm/typeorm/issues/4354


TheAifam5:
Should be ready to merge. Please review.

pleerock:
`donenv` is super popular while `dotenv-flow` is a new and not so popular yet. Why they didn't make those changes part of `dontenv` itself? Is it reliable to use them? 

Also, I guess if we would merge it people will start asking about same functionality for other configuration types, e.g. `ormconfig.dev.json` and `ormconfig.prod.json`.

I'm not sure if we should merge it or not, so let's ask @vlapo and @Kononnable opinions.

pleerock:
_(another reason why I don't want a configuration to be core of the framework - people always use different implementations/solutions). But since we still have it, we need to do something with it._

TheAifam5:
The idea with `ormconfig.dev.json` would be really nice too :D 
If you think that PR is not needed, you can close it but that you need to decide.
I use `dotenv-flow` in production and works really well.

pleerock:
Let's see what other people say

Kononnable:
Sorry for being little late for the party.
I'm not comfortable with adding dependency which isn't widely used in the community(yet). It's a matter of cybersecurity.
Typeorm is used by many people(and corporations) and if we introduce a vulnerability we might harm many projects along the way. We don't have the time to check every line of code of dependencies(recursively), but having dependencies which are used by millions gives us a safety net. Popular packages got their code read by more people and even corporations which specialize in node package system security looks at it regularly.

TheAifam5:
@Kononnable 	no problem. we can just leave that PR, and reopen when will be widely used. :)

pleerock:
@TheAifam5 thanks for understanding.

PR #4405, opened 2019-07-08
feat: Add pool error handler

Ginden:
Add option to customize pool error handling. Users can decide to log these errors with higher level (eg. error), crash application or reconnect.


Ginden:
Changing source branch: https://github.com/typeorm/typeorm/pull/4474

Issue #3721, opened 2019-02-27
duplicate column name


pomeisl:
**Issue type:**
When we have columns similar to each other and their names are too long the lib will ganereate wrong aliases in the statements like below.

SELECT "selection"."CEID" AS "selection_CEID", 
"selection"."ACCOUNT_ID" AS "selection_ACCOUNT_ID", 
"selection"."FOADOS" AS "selection_FOADOS", 
"selection"."CLIENT_NAME" AS "selection_CLIENT_NAME", 
"selection"."TOTAL_BALANCE" AS "selection_TOTAL_BALANCE", 
"selection"."EGYENLEGKATEGORIA" AS "selection_EGYENLEGKATEGORIA", 
"selection"."EXECUTOR_CASE_ID" AS "selection_EXECUTOR_CASE_ID", 
"selection"."EXECUTOR_NAME" AS "selection_EXECUTOR_NAME", 
"selection"."EX_ADDR_PERMANENT_POSTCODE" AS "selection_EX", 26
"selection"."EX_ADDR_PERMANENT_CITY" AS "selection_EX", 22
"selection"."EX_ADDR_PERMANENT_STREET" AS "selection_EX", 24
"selection"."VH_EXECUTOR_COST_PAID_DATE" AS "selection_VH", 
"selection"."BEHAJTHATATLAN" AS "selection_BEHAJTHATATLAN", 
"selection"."UTOLSOBEFIZETESDATUMA" AS "selection_UT", 
"selection"."JOG_19_26OS" AS "selection_JOG_19_26OS", 
"selection"."JOG27ES" AS "selection_JOG27ES", 
"selection"."VH_LEVELES" AS "selection_VH_LEVELES", 
"selection"."STATUSZMODOSITASDATUMA" AS "selection_ST", 
"selection"."UTOLSODOKSIDATUMA" AS "selection_UTOLSODOKSIDATUMA", 17 + 10
"selection"."MEGYE" AS "selection_MEGYE" 
FROM "FAKTOR_KOLLECTO"."V_VH_VISITING_SELECTION_TAB" "selection"

The result entity:

{
  ceid: 11410283,
  accountId: 11448441,
  mainDebtorName: '*',
  clientName: '*',
  totalBalance: 1297044.32,
  balanceType: '*',
  executorCaseId: '*',
  executorName: '*',
  executorPermanentPostcode: 'Liszt F. u. 6.',
  executorPermanentCity: 'Liszt F. u. 6.',
[...]

Issue #2200, opened 2018-05-23
Issue with snake_case naming strategy


bssergy:
**Issue type:**

[ ] question
[x] bug report
[ ] feature request
[ ] documentation issue

**Database system/driver:**

[ ] `cordova`
[ ] `mongodb`
[x] `mssql`
[x] `mysql` / `mariadb`
[ ] `oracle`
[ ] `postgres`
[ ] `sqlite`
[ ] `sqljs`
[ ] `react-native`

**TypeORM version:**

[x] `latest`
[ ] `@next`
[ ] `0.x.x` (or put your version here)

**Steps to reproduce or a small repository showing the problem:**

I have two tables:

product
-----------
id

product_attachment
-----------
id
product_id

`var productAttachment = repository.findAndCount({ relations: 'product'});`

QueryFailedError: ER_DUP_FIELDNAME: Duplicate column name 'ProductAttachment_product_id'

This issue happens because query builder generate alias ProductAttachment_product_id twice:
product_attachment.product_id -> ProductAttachment_product_id
product_attachment.product.id -> ProductAttachment_product_id

pleerock:
Can you provide more info? `relations: 'product'` is not correct, it shall be `{ relations: ['product'] }`. Can you show your entities and what are actually doing?

bssergy:
@pleerock 
```
@Entity('product')
export class Product {

  @PrimaryGeneratedColumn()
  id: number;

  @OneToMany(() => ProductAttachment, productAttachment => productAttachment.product)
  @JoinColumn({ referencedColumnName: 'product_id' })
  product_attachments: ProductAttachment[];
}

@Entity('product_attachment')
export class ProductAttachment {

  @PrimaryGeneratedColumn()
  id: number;

  @Column()
  product_id: number;

  @ManyToOne(() => Product, product => product.product_attachments)
  @JoinColumn({ name: 'product_id', referencedColumnName: 'id' })
  product: Product;
}

const attachments = await connection
  .getRepository(ProductAttachment)
  .findAndCount({ relations: ["product"] });
```

pleerock:
Sounds like a bug. In 0.3.0 "relations" are completely refactored, so this issue shall be fixed in `@next`.

ValiDraganescu:
I am using a custom naming strategy where all the interface functions are implemented and this is how I add it to the connection options:
```
let baseConfig: ConnectionOptions = {
  "type": "mysql",
  "host": localHost,
  "port": 3306,
[...]

Issue #1969, opened 2018-04-19
Duplicate column alias


awn70:
**Issue type:**

[ ] question
[x] bug report
[ ] feature request
[ ] documentation issue

**Database system/driver:**

[ ] `cordova`
[ ] `mongodb`
[ ] `mssql`
[ ] `mysql` / `mariadb`
[x] `oracle`
[ ] `postgres`
[ ] `sqlite`
[ ] `sqljs`
[ ] `react-native`

**TypeORM version:**

[ ] `latest`
[x] `@next`
[ ] `0.x.x` (or put your version here)

**Steps to reproduce or a small repository showing the problem:**

The buildColumnAlias function in SelectQueryBuilder tends to return duplicate column alias names. If the columnAliasName is more than 29 chars long you get the table alias and 2 chars from the column name: 
e.g. MyLongTableOrClassName_MyLongColumnName -> MyLongTableOrClassName_MY

josephktcheung:
Related to https://github.com/typeorm/typeorm/issues/1785?

pleerock:
We are aware about this issue, however its all about Oracle's limitation on alias names. Solution is to use shorter alias when you create a query builder, e.g. createQueryBuilder("shorter") and obviously don't use too long column names.

awn70:
@josephktcheung Yes, same problem as #1785 

@pleerock Why don't you use something like string-hash to generate unique alias names. Or the first n chars from the original column name and append a generated hash to make them unique?

awn70:
Ok.
I thought about something like this:
```javascript
// original code for hash function from https://github.com/darkskyapp/string-hash
// creates hashes from 0 to 4294967295
function hash( str ) {
    let hash = 5381,
        i = str.length;

    while( i ) {
        hash = ( hash * 33 ) ^ str.charCodeAt( --i );
    }
    return hash >>> 0;
}

function calcAlias( tableName, columnName ) {
    let h = hash( tableName + "." + columnName );
    return columnName.substr( 0, 19 ) + "_" + h;
}

calcAlias( 'MY_LONG_TABLE_NAME', 'MY_VERY_LONG_COLUMN_NAME' );
// -> MY_VERY_LONG_COLUMN_526535556
```
Maybe it helps...

wdmcgrory:
I implemented  the hash correction suggested by @awn70, and this appears to take care of my issues anyway. 

[...]

PR #4435, opened 2019-07-14
Fixing #1969, #3721, #2200 duplicate column alias

neverson-silva:
Duplicate column error appears since version 2.10. 
Please can you fix it as it is impacting the applications


neverson-silva:
@vlapo this is an old bug, can anyone take a look?

PR #4444, opened 2019-07-16
feat: allow negating brackets

### LINKED ISSUE #4442: feat: allow not predicate on querybuilder
**Issue type:**

[x ] feature request

**TypeORM version:**

[x] `latest`

**Description** 
Querybuilder does not allow to add a nested `not` predicate. I was expecting something like: 
```
createQueryBuilder(User)
.whereNot('...')
```
or something like: 
```
createQueryBuilder(User)
.where(new Brackets(wb => ,{negate: true}))
```
I think the latter is most easy to implement and I'll create an pr for this.

tkvw:
Closes #4442

### DIFF
--- a/docs/select-query-builder.md
+++ b/docs/select-query-builder.md
@@ -341,12 +341,15 @@ createQueryBuilder("user")
         qb.where("user.firstName = :firstName", { firstName: "Timber" })
           .orWhere("user.lastName = :lastName", { lastName: "Saw" })
     }))
+    .andWhere(new Brackets(qb => {
+        qb.where("user.firstName = :firstName2", { firstName2: "Bob" })
+    }, { negate : true }))
 ```
 
 Which will produce the following SQL query:
 
 ```sql
-SELECT ... FROM users user WHERE user.registered = true AND (user.firstName = 'Timber' OR user.lastName = 'Saw')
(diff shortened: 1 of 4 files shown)


pleerock:
I really don't like us having this `Brackets` and now this ugly option... Not fun of this, so I need some opinions @Kononnable @vlapo

Kononnable:
Yeah, adding `not` as option might not be the most readable solution. People reading code would have to first understand specified condition, than realize that what they really want is to revert it.

Another option would be to add `not` exactly like brackets, but generating negated condition.
Instead of
```typescript
.andWhere(new Brackets(qb => {
        qb.where("user.firstName = :firstName2", { firstName2: "Bob" })
    }, { negate : true }))
```
we would use something like
```typescript
.andWhere(new Not(qb => {
        qb.where("user.firstName = :firstName2", { firstName2: "Bob" })
    }))
```

tkvw:
@Kononnable and @pleerock : I like the suggestion of @Kononnable more than mine solution. Not entirely sure howto create a PR like this. I'll give it a shot.

pleerock:
@Kononnable's suggestion is definitely better. However we already have Not operator and its used in a bit different context. Wouldn't adding new signatures to that operator confuse people?

imnotjames:
Given we're not going to move forward with this PR I'm going to close it out.

PR #4458, opened 2019-07-18
docs: fix missing async

davidpodhola:
In the second example the lambda is correctly started with an `async`. It is missing in the first example.


Kononnable:
We have to return a Promise here, so theoretically this could be done without async, but having async function would be the most common scenario.

Issue #4359, opened 2019-06-25
Column comments are persisted in MySQL but not PostgreSQL


rmainseas:
**Issue type:**

- [ ] question
- [X] bug report
- [ ] feature request
- [ ] documentation issue

**Database system/driver:**

- [ ] `cordova`
- [ ] `mongodb`
- [ ] `mssql`
- [ ] `mysql` / `mariadb`
- [ ] `oracle`
- [X] `postgres`
- [ ] `cockroachdb`
- [ ] `sqlite`
- [ ] `sqljs`
- [ ] `react-native`
- [ ] `expo`

**TypeORM version:**

- [ ] `latest`
- [ ] `@next`
- [x] `0.2.18` (or put your version here)

**Steps to reproduce or a small repository showing the problem:**

```ts
import {
  Entity, Column,
} from 'typeorm';

@Entity()
export class User {
  @Column({
    type: 'varchar',
    nullable: false,
    comment: 'The name of the user (i.e Samuel Jenkins)',
  })
  public name: string;
}
```

In MySQL this comment would be persisted to the database when a migration is generated (the migration would contain a SQL statement including the comment). In PostgreSQL.....that's not the case.

<!--
    To answer those questions you need to put "x" inside the square brackets, for example:
    - [x] `mysql`
    - [ ] `postgres`

    Also, please format your code properly (by taking code blocks into ```ts .... ```)
!>

Possibly related to #3360, though that seems more related to missing functionality (support for table comments) whereas, this functionality works in MySQL....just not PostgreSQL (hence why I categorized it as a bug)

nscarlson:
@nionoku what's the status on this?

imnotjames:
There's open issues with the PR that haven't been resolved.  If anyone would like to pick that up we could get that change in.

imnotjames:
Duplicate of #3360

Issue #3360, opened 2019-01-04
setting COMMENT on columns and tables (PostgreSQL)


ulo:
**Issue type:**

[ ] question
[ ] bug report
[x] feature request
[ ] documentation issue

**Database system/driver:**

[ ] `cordova`
[ ] `mongodb`
[ ] `mssql`
[ ] `mysql` / `mariadb`
[ ] `oracle`
[x] `postgres`
[ ] `sqlite`
[ ] `sqljs`
[ ] `react-native`
[ ] `expo`

**TypeORM version:**

[x] `latest`
[ ] `@next`
[ ] `0.x.x` (or put your version here)

**Feature request:**
When I define my entities I would like to be able to set a COMMENT to the database object. From looking at the source code this seems to be currently possible only for **MySQL** and not for **PostgreSQL**:
```ts
@Entity()
export class User {
  ...
  @Column({comment: 'The age in years.'})  // <--
  age: number
}
```

Using PostgreSQL one can also set a COMMENT on the **table** itself, which is not yet supported by TypeORM at all, but could look like this:
```ts
@Entity({comment: 'A user profile in the wep app.'})  // <--
export class User {
  ...
  @Column({comment: 'The age in years.'})
  age: number
}
```

I am asking for this feature as I am using the awesome PostGraphile tool, which heavily relies on COMMENTs for its configuration, and it would be really useful to do this by using TypeORM entity decorators!!

Many thanks,
Ulrich

vlapo:
This should be fixed also for mysql. And investigate comments in all supported databases and implemented it too.

Diluka:
Please save comment in meta first. It helps generating API docs.

DavidTheProgrammer:
Is anybody currently working on this?

imagine10255:
+1

auerbachb:
+1

imnotjames:
Code in #6902 is ready to go, has tests, should be good to merge.

AlainBartmanDilaw:
> This should be fixed also for mysql. And investigate comments in all supported databases and implemented it too.

Has this feature been implemented for mysql databases ?

derushio:
```ts
@Entity({comment: 'A user profile in the wep app.'})
```

Is this option not implemented?

nawaz-pasha:
> ```ts
> @Entity({comment: 'A user profile in the wep app.'})
> export class User {
> ...
> }
> ```
> 
> Is this option not implemented?

Don't think so.

[...]

Issue #3357, opened 2019-01-03
Migration generation drops and creates columns instead of altering resulting in data loss


ryuuji3:
**Issue type:**

[ ] question
[x] bug report
[ ] feature request
[ ] documentation issue

**Database system/driver:**

[ ] `cordova`
[ ] `mongodb`
[ ] `mssql`
[ ] `mysql` / `mariadb`
[ ] `oracle`
[x ] `postgres`
[ ] `sqlite`
[ ] `sqljs`
[ ] `react-native`
[ ] `expo`

**TypeORM version:**

[ ] `latest`
[ ] `@next`
[x] `0.2.11` (or put your version here)

**Steps to reproduce or a small repository showing the problem:**

I apologize- I can't help but think I've seen a similar issue but I couldn't find it in opened or closed issues.

Let's say I have a varchar column with length of 50;

```ts
@Column({
   length: 50
})
public example: string;
```

Which results in this migration:
```sql
CREATE TABLE "bug" ("text" character varying(50))
```

And then I go ahead and change the length to 51;

```ts
@Column({
   length: 51
})
public example: string;
```

The resulting generated migration code is:
```sql
ALTER TABLE "bug" DROP COLUMN "example"
ALTER TABLE "bug" ADD "example" character varying(51)
```

Which obviously would result in data loss.
I would have expected;

```sql
ALTER TABLE "bug" ALTER COLUMN "example" character varying(51)
```

pleerock:
no, its not related to #3352. We already had this question before, I'll provide answer once again.

Its by design. Let's say you have length equal to `50`. What will happen if you change it to `10`? Right, data loss again. Some databases even don't allow to do it without dropping the column. Some allow however without any warning. We could handle all possible cases, but this is a bit more complex in implementation, that's why decided to make it simple this way.

ryuuji3:
Thanks for getting back to me!

It makes sense that data truncation would result in *some* data loss; but not the entire column. However, something like increasing the size of a column would surely not result in data loss? I think this is worth pursuing.

Understandably something like SQLite which does not support `ALTER COLUMN` would not benefit from this feature... but that's also only one of many RMBDS' that do.

I think migration generation is a killer feature of many ORMs and TypeORM should not be left out. We personally chose TypeORM for our production app based on the fact that it at has migration generation where something like Sequelize has very limited support.
If this isn't something high on the road map, I understand. But I do not agree that it shouldn't be up for consideration whatsoever.

[...]

PR #4461, opened 2019-07-18
feat: adding column comments for postgres

nionoku:
Adding a comment to a column when creating a column for postgres.

Related #4359

### DIFF
--- a/src/driver/postgres/PostgresQueryRunner.ts
+++ b/src/driver/postgres/PostgresQueryRunner.ts
@@ -545,6 +545,12 @@ export class PostgresQueryRunner extends BaseQueryRunner implements QueryRunner 
             downQueries.push(new Query(`ALTER TABLE ${this.escapePath(table)} DROP CONSTRAINT "${uniqueConstraint.name}"`));
         }
 
+        // create column's comment
+        if (column.comment) {
+            upQueries.push(new Query(`COMMENT ON COLUMN ${this.escapePath(table)}.${column.name} IS '${column.comment}'`));
+            downQueries.push(new Query(`COMMENT ON COLUMN ${this.escapePath(table)}.${column.name} IS '${column.comment}'`));
+        }
+
         await this.executeQueries(upQueries, downQueries);
 
         clonedTable.addColumn(column);
@@ -1725,7 +1731,11 @@ export class PostgresQueryRunner extends BaseQueryRunner implements QueryRunner 
             sql += `, CONSTRAINT "${primaryKeyName}" PRIMARY KEY (${columnNames})`;
         }
 
-        sql += `)`;
+        sql += `);`;
+
+        table.columns
+            .filter(it => it.comment)
+            .forEach(it => sql += ` COMMENT ON COLUMN ${this.escapePath(table)}.${it.name} IS '${it.comment}';`);
 
         return new Query(sql);
     }
--- a/test/other-issues/postgre-add-comment/entity/Car.ts
+++ b/test/other-issues/postgre-add-comment/entity/Car.ts
@@ -1,0 +1,19 @@
+import {Entity, PrimaryGeneratedColumn, Column} from "../../../../src/index";
+
+@Entity("cars")
+export class Car {
+    public static build(item: { wheelsCount: number }) {
+        const entity = new Car();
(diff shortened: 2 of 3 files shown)


nionoku:
Fix test, comments are now added when creating a table in PostgreSQL.

rmainseas:
Just to clarify the title of the PR...


Postgres column comments are already supported. They just don't work. But it's cool that you have a PR to fix this, thank you!

nionoku:
> Just to clarify the title of the PR...
> 
> Postgres column comments are already supported. They just don't work. But it's cool that you have a PR to fix this, thank you!

Indeed, but the moment I was thinking about the name, there was a slightly different thought in my head. Thanks for comment.

rmainseas:
Yeah, not trying to be pedantic - I just thought it was worth pointing out :) <3

rmainseas:
Also, just to add - I've noticed that column comments actually already work (sort of) with postgres.....if you create a column it doesn't add the column comment in the migration. But if you somehow change something about that column (the length, the type etc.) the changes written into a migration also include the column comment.

nionoku:
> Also, just to add - I've noticed that column comments actually already work (sort of) with postgres.....if you create a column it doesn't add the column comment in the migration. But if you somehow change something about that column (the length, the type etc.) the changes written into a migration also include the column comment.

Yes, a comment for the column was added when changing the model fields.
My PR adds a comment to the column when creating the table :)

nionoku:
@rmainseas hello! Thank you for your attention to the question, regarding the attached issue #3357, I created a small repository, as typeorm I used my fork accordingly, and tested: created an entity, synchronized it with the database, conducted a test: the first module was executed, the second failed (as expected), then I changed the length of the field in the entity and repeated synchronization with the database, and repeated testing - all tests have been passed. That is, the comment is saved (by the way, this functionality is not included in my PR).
https://github.com/nionoku/typeorm-field-options-test

jjosef:
This is really basic and simple, any reason it is not being merged?

src/driver/postgres/PostgresQueryRunner.ts
```diff
@@ -545,6 +545,12 @@ export class PostgresQueryRunner extends BaseQueryRunner implements QueryRunner
             downQueries.push(new Query(`ALTER TABLE ${this.escapePath(table)} DROP CONSTRAINT "${uniqueConstraint.name}"`));
         }
 
+        // create column's comment
+        if (column.comment) {
+            upQueries.push(new Query(`COMMENT ON COLUMN ${this.escapePath(table)}.${column.name} IS '${column.comment}'`));
+            downQueries.push(new Query(`COMMENT ON COLUMN ${this.escapePath(table)}.${column.name} IS '${column.comment}'`));
```
AlexMesser: `column.name` should be escaped here

src/driver/postgres/PostgresQueryRunner.ts
```diff
@@ -1725,7 +1731,11 @@ export class PostgresQueryRunner extends BaseQueryRunner implements QueryRunner
             sql += `, CONSTRAINT "${primaryKeyName}" PRIMARY KEY (${columnNames})`;
         }
 
-        sql += `)`;
+        sql += `);`;
+
+        table.columns
+            .filter(it => it.comment)
+            .forEach(it => sql += ` COMMENT ON COLUMN ${this.escapePath(table)}.${it.name} IS '${it.comment}';`);
```
AlexMesser: `it.name` should be escaped here

AlexMesser:
@nionoku we also need to obtain information about column comments from database and put it into ColumnMetadata for subsequent schema synchronization, like we did it with column charset and collation for example. e.g. in your current PR comment won't be applied on existing column. It also won't be removed once you remove comment from the decorator (or change existing comment)

imnotjames:
Any luck on addressing these issues?  Without pulling it into column metadata, synchronize will stop working.

Can you rebase this & fix conflicts, as well?

imnotjames:
Related issue #3360

imnotjames (changes requested):
This needs to update the change detection code to handle checking for comment changes.

Also comments need to be escaped somehow

PR #4474, opened 2019-07-23
feat: Add pool error handler

Ginden:
Add option to customize pool error handling. Users can decide to log these errors with higher level (eg. error), crash application or reconnect.


vlapo:
Why not use `Logger` (https://typeorm.io/#/logging/using-custom-logger) and add only one new method? e.g. `logPoolError`

Ginden:
Why should I want to put eg. application termination logic in *logger*?

vlapo:
Sorry I read "pool error logger". Could you please update docs?

Ginden:
@vlapo Done!

pleerock:
Thanks guys!

Issue #4450, opened 2019-07-17
Missing socketPath when setup read/write replication


JarvisYuan:
**Issue type:**

[ ] question
[x] bug report
[ ] feature request
[ ] documentation issue

**Database system/driver:**

[ ] `cordova`
[ ] `mongodb`
[ ] `mssql`
[x] `mysql` / `mariadb`
[ ] `oracle`
[ ] `postgres`
[ ] `cockroachdb`
[ ] `sqlite`
[ ] `sqljs`
[ ] `react-native`
[ ] `expo`

**TypeORM version:**

[x] `latest`
[ ] `@next`
[ ] `0.x.x` (or put your version here)

**Steps to reproduce or a small repository showing the problem:**

When I used extra socketPath with read/write replication，config like this:
```ts
const config = {
        type: 'mysql',
        synchronize: false,
        entities: [path.resolve(__dirname, '../**/*.entity.ts')]
        replication: {
            master: {
                database: 'test_data',
                username: 'test_data_write',
                extra: {
                    socketPath: '/app/service/xxxxx/mysql.sock'
                }
            },
            slaves: [{
                database: 'test_data',
                username: 'test_data_read',
                extra: {
                    socketPath: '/app/service/xxxxx/mysql.sock'
                }
            }]
        }
}
```

There is no socketPath passed to Mysql. Then I found the problem in `typeorm/driver/mysql/MysqlDriver.js`.
 
```js
    MysqlDriver.prototype.createConnectionOptions = function (options, credentials) {
        credentials = Object.assign(credentials, DriverUtils_1.DriverUtils.buildDriverOptions(credentials)); // todo: do it better way
        // build connection options for the driver
        return Object.assign({}, {
            charset: options.charset,
            timezone: options.timezone,
            connectTimeout: options.connectTimeout,
            insecureAuth: options.insecureAuth,
            supportBigNumbers: options.supportBigNumbers !== undefined ? options.supportBigNumbers : true,
            bigNumberStrings: options.bigNumberStrings !== undefined ? options.bigNumberStrings : true,
            dateStrings: options.dateStrings,
            debug: options.debug,
            trace: options.trace,
            multipleStatements: options.multipleStatements,
            flags: options.flags
        }, {
            host: credentials.host,
            user: credentials.username,
            password: credentials.password,
            database: credentials.database,
            port: credentials.port,
            ssl: options.ssl
        }, options.acquireTimeout === undefined
            ? {}
[...]

PR #4475, opened 2019-07-23
feat: add socketPath connection option for MYSQL

joseph1125:
Attempt to perform a temp fix for issue #4450


pleerock:
just imagine if somebody read this code - what he would think. Of course he would think - let's simplify it and he will do a refactor. And since there is no test for this case your functionality will break. I can merge it, but I guarantee you this won't stay too long without tests.

imnotjames:
I'm pretty sure this can be accomplished by using the `extra` options instead.

imnotjames:

PR #4480, opened 2019-07-24
add skipLibCheck: true

davewasmer:
This mostly just reduces noise during `gulp package`. There's 1000+ library errors that are emitted during the build, but are ignored (they don't fail the build). Might as well silence them?

### DIFF
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -18,7 +18,8 @@
         "pretty": true,
         "strictNullChecks": true,
         "noUnusedLocals": true,
-        "downlevelIteration": true
+        "downlevelIteration": true,
+        "skipLibCheck": true
     },
     "exclude": [
         "tmp",


csuich2:
I'm don't think I agree with this. The main point of typescript is to perform type checking against your code and your dependencies. `skipLibCheck` removes all type checking from your dependencies.

If you're having issues with gulp package, maybe we can add a `--skipLibCheck` to just the `gulp package` command, but I really don't think it should be enabled in the main `tsconfig.json`.

davewasmer:
`skipLibCheck` (at least, as I understand it) disables type checking *within* your dependencies. It doesn't stop type checking against the interfaces exposed and used within your project.

I guess I don't see the point of type checking inside all those dependencies. There's currently over a thousand type errors produced by TypeORM's dependencies. If those 1000+ errors are never addressed, and never lead to any action on the part of TypeORM devs (which is fine - we don't need to tackle those errors as long as the deps stick to their public APIs), then what is the use of keeping them around?

csuich2:
I could be mistaken, but the TS docs say this about `--skipLibCheck`:
```
Skip type checking of all declaration files (*.d.ts).
```

Which to me means it ignores *ALL* type definitions, not just inter-dependency type checking.

I'm not familiar with the gulp package script, but I think it is doing something funky. Notice how tsc doesn't produce errors when run in an IDE or from the CLI against the project? That should mean that there aren't any typing errors. But it looks like `gulp package` is doing something extra. If there were truly typing errors, then the `npm compile` would be failing.

pleerock:
I don't have any errors when I use gulp command, so I'm not sure if this PR fixes something.

imnotjames:
This has been open for a year without much activity.  I don't have these errors, either.   Can we close it out?

PR #4495, opened 2019-07-26
See if we can run ci tests in parallel, one DB at a time

csuich2:
I don't mean for this PR to be accepted as is, I just wanted to open a discussion about an approach like this.

While working on stuff, I was getting frustrated at how the CI tests fail due to an issue with one DB that isn't present on the others. I took a stab at seeing if we could change the mocha `describe` and `it` usage to indicate which DB failed, but I didn't come up with anything good.

So, I figured maybe we could try running all the tests for each DB separately, that way we get a better sense of what is failing. This change would cause Travis to run `npm test` with a config unique to each DB so that only a single DB is tested. It then runs all of the DBs in parallel (it appears 5 at a time).

Right now I think it is a bit slower than the current approach, because it has to spin up, `npm install`, etc. for each test. We use CircleCI, so I know how I'd make it work more efficiently there, but I'm not too familiar with Travis.

Anyone have thoughts on how to speed this up? I think it would be really nice to know if my changes cause a break for multiple DBs or just a single DB.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,11 +17,21 @@ before_script:
   - sudo service mysql stop
   - sudo service postgresql stop
   - docker-compose up -d
-  - cp ormconfig.travis.json ormconfig.json
   - npm install sqlite3 --build-from-source
+
+env:
+  - TEST_DB=mariadb
+  - TEST_DB=mongodb
+  - TEST_DB=mssql
+  - TEST_DB=mysql
+  - TEST_DB=oracle
+  - TEST_DB=postgres
+  - TEST_DB=sqlite
+  - TEST_DB=sqljs
 
 script:
   - npm run lint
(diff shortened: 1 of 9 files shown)


ruscon:
> Right now I think it is a bit slower than the current approach, because it has to spin up, npm install, etc. for each test. We use CircleCI, so I know how I'd make it work more efficiently there, but I'm not too familiar with Travis.

What about this ?
https://stackoverflow.com/a/52387639/3408246

pleerock:
I think we just need to migrate to circleci and that's it. We started the process but something left unfinished, I think we had some tests sometimes failing on CircleCI and we didn't finish investigation on it. I think there is an open PR about it.

imnotjames:
We migrated to CircleCI.  The testing code makes it easy to split out other databases.  If you'd like to make that happen with CIrcleCI please open a new PR with that change.  Cheers!

Issue #4502, opened 2019-07-29
Lazy relations cannot be read immediately after being setup - promise is resolving to undefined


OJezu:
After setting up properties being relation mapped to promises resolving to mapped entities, properties hold promises resolving to `undefined` instead of the mapped entity. Saving the entity probably fixes that.

**Issue type:**

[ ] question
[X] bug report
[ ] feature request
[ ] documentation issue

**Database system/driver:**

[ ] `cordova`
[ ] `mongodb`
[ ] `mssql`
[X] `mysql` / `mariadb`
[ ] `oracle`
[ ] `postgres`
[ ] `cockroachdb`
[ ] `sqlite`
[ ] `sqljs`
[ ] `react-native`
[ ] `expo`

**TypeORM version:**

[X] `latest`
[ ] `@next`
[ ] `0.x.x` (or put your version here)

**Steps to reproduce or a small repository showing the problem:**
```typescript
@Entity()
export class Post {
    @PrimaryGeneratedColumn()
    id: number;

    @Column()
    title: string;

    @ManyToOne(() => Category, category => category.posts)
    category: Promise<Category>;
}
```

```typescript
@Entity()
export class Category {
    @PrimaryGeneratedColumn()
    id: number;

    @Column()
    name: string;

    @OneToMany(() => Post, post => post.category)
    posts: Promise<Post[]>;
}
```

```
const category1 = new Category();
category1.name = "category #1";

const post1 = new Post();
post1.title = "Hello Post #1";
post1.category = Promise.resolve(category1);

expect(await post1.category).to.be.equal(category1); // fails
```

```
const post1 = new Post();
post1.title = "Hello Post #1";
const post2 = new Post();
post2.title = "Hello Post #2";

const category1 = new Category();
category1.name = "category #1";
category1.posts = Promise.resolve([post1, post2]);

expect(await category1.posts).to.be.equal([post1, post2]); // fails
```

imnotjames:
`to.be.equal()` will compare by reference.

`to.be.eql` will compare them deeply & should be used in this case.

I've updated your tests to use `eql`.

```
const post1 = new Post();
post1.title = "Hello Post #1";
const post2 = new Post();
post2.title = "Hello Post #2";

const category1 = new Category();
category1.name = "category #1";
category1.posts = Promise.resolve([post1, post2]);

expect(await category1.posts).to.be.eql([post1, post2]); // fails
```

These pass.

imnotjames:
[...]

PR #4503, opened 2019-07-29
Tests for #4502 - lazy relations cannot be read after being set

OJezu:
Tests for #4502


imnotjames:
Tests pass with those changes, this isn't an issue.  Included more info in the issue this was for - TLDR `eql` works when `equal` didn't

PR #4505, opened 2019-07-29
fix: Transform properties when creating or updating entities

### LINKED ISSUE #4293: Transformed value not returned on save
**Issue type:**

[ ] question
[X] bug report
[ ] feature request
[ ] documentation issue

**Database system/driver:**

[ ] `cordova`
[ ] `mongodb`
[ ] `mssql`
[ ] `mysql` / `mariadb`
[ ] `oracle`
[X] `postgres`
[ ] `cockroachdb`
[ ] `sqlite`
[ ] `sqljs`
[ ] `react-native`
[ ] `expo`

**TypeORM version:**

[X] `latest`
[ ] `@next`
[X] `0.2.18` (or put your version here)

```ts
@Entity()
class Example extends BaseEntity {
  @PrimaryGeneratedColumn()
  id: number;

  @Column({
    transformer: {
      to: value => String(value).toUpperCase(),
      from: value => value,
    },
  })
  name: string;
}

describe('Example', () => {
  it('Should uppercase the name', async () => {
    const saved = await Example.create({name: 'test'}).save();
    expect(saved).to.have.property('name', 'TEST');
    // expected: TEST
    // actual: test
  });
});
```

When saving a column that has a transformer, the un-transformed value is returned from  the save function. This happens even if a `from` transformer is provided -- though none should be necessary. I've spent a while looking into this and can't figure out if it's design intent or not at this point. I'm going to read the source of `@CreateDateColumn` and see how it works, since it behaves the way I would expect.

loganyott:
Fixes #4293. Entities no longer need to be retrieved after being saved to have transformed properties.

Updates `SubjectExecutor` to check `diffColumns` and transform the values if columns were changed as the result of a `save`.

During testing, I found columns in the `generatedMap` were specifically not included in `diffColumns` and vice-versa, so there shouldn't be columns being transformed twice.

Added in a `@CreateDateColumn` in the test to demonstrate that generated columns are not affected.

### DIFF
--- a/src/persistence/SubjectExecutor.ts
+++ b/src/persistence/SubjectExecutor.ts
@@ -328,6 +328,13 @@ export class SubjectExecutor {
             }
 
             subjects.forEach(subject => {
+                if (subject.diffColumns) {
+                    subject.diffColumns.forEach(column => {
+                        const value = column.getEntityValue(subject.entity!);
+                        const preparedValue = this.queryRunner.connection.driver.prepareHydratedValue(value, column);
+                        column.setEntityValue(subject.entity!, preparedValue);
+                    });
+                }
                 if (subject.generatedMap) {
                     subject.metadata.columns.forEach(column => {
                         const value = column.getEntityValue(subject.generatedMap!);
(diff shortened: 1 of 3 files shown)


vlapo:
Check failing tests please. Also try run tests on your localhost before submitting PR.

imnotjames:
more than a year since last update - gonna close this in favor of #6150

PR #3900, opened 2019-03-26
Fix transform embeddeds (mongodb)

rustamwin:
Fixes #3899 #2642

PR #4511, opened 2019-07-30
fix: transform nested embedded array column overwritten undefined (mongodb)

tilfin:
**Issue type:**

[ ] question
[x] bug report
[ ] feature request
[ ] documentation issue

**Database system/driver:**

[ ] `cordova`
[x] `mongodb`
[ ] `mssql`
[ ] `mysql` / `mariadb`
[ ] `oracle`
[ ] `postgres`
[ ] `cockroachdb`
[ ] `sqlite`
[ ] `sqljs`
[ ] `react-native`
[ ] `expo`

**TypeORM version:**

[x] `latest`
[ ] `@next`

```typescript
export class Post {
  @Column()
  title: string

  @Column()
  comments: Comment[]
}

export class Comment {
  @Column()
  text: string

  @Column(type => Attachment)
  attachments: Attachment[]  // new column added

  constructor() {
    this.attachments = [] // added
  }
}

export class Attachment {
  @Column()
  subject: string
}
...
const oldPost = new Post()
oldPost.title = 'title'
const comment  = new Comment()
comment.text = 'hello'
oldPost.comments = [comments]
await repo.save(oldPost); // before attachments column is added

// Add attachments to Comment

const [post] = await repo.find();
post.comments[0].attachments  // is undefined, empty array expected
```

For example, adding new column to a collection, existtant documents have no columns in DB.
I want to set the column default value in the constructor and then I can handle it without whether its value is null or not.

I think that skipping the assignment of columns that has no value should be implemented in the commit (https://github.com/typeorm/typeorm/commit/0ae68e49aab371cd56dd253e8ddbf13f3e1a57bb#diff-5ca83022308e329a128f5a2bf58ba6c8).

### DIFF
--- a/src/query-builder/transformer/DocumentToEntityTransformer.ts
+++ b/src/query-builder/transformer/DocumentToEntityTransformer.ts
@@ -89,7 +89,10 @@ export class DocumentToEntityTransformer {
                     entity[embedded.propertyName] = (document[embedded.prefix] as any[]).map((subValue: any, index: number) => {
                         const newItem = embedded.create();
                         embedded.columns.forEach(column => {
-                            newItem[column.propertyName] = subValue[column.databaseNameWithoutPrefixes];
+                            const value = subValue[column.databaseNameWithoutPrefixes];
+                            if (value === undefined) return;
+
+                            newItem[column.propertyName] = value;
                         });
                         addEmbeddedValuesRecursively(newItem, document[embedded.prefix][index], embedded.embeddeds);
                         return newItem;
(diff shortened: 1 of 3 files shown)


rustamwin:
Hmm. This will be a continuation of the #3900

pleerock:
@rustamwin how it is?

pleerock:
I'm not sure if we should merge this change or not. In relational databases we completely ignore undefined properties and it works great.

However for mongodb it feels debatable. Reason is different way of storing things by mongo and mongo being schemaless on its own. 

What if user wants to completely remove some property from the document? Such changes would make it impossible, but such task feels absolutely natural for mongodb usage.

AlexMesser:
Looks like current implementation is by design. Closing.

Issue #4513, opened 2019-07-30
CockroachDB support for onConflict


HNicolas:
**Issue type:**

[ ] question
[ ] bug report
[x] feature request
[ ] documentation issue

**Database system/driver:**

[ ] `cordova`
[ ] `mongodb`
[ ] `mssql`
[ ] `mysql` / `mariadb`
[ ] `oracle`
[ ] `postgres`
[x] `cockroachdb`
[ ] `sqlite`
[ ] `sqljs`
[ ] `react-native`
[ ] `expo`

**TypeORM version:**

[x] `latest`
[ ] `@next`
[ ] `0.x.x` (or put your version here)

**Steps to reproduce or a small repository showing the problem:**

CockroachDB support `ON CONFLICT` statement, it would be great to add support for it in typeorm query builder.

PR #4518, opened 2019-07-31
feat: add support for ON CONFLICT for cockroach

HNicolas:
Closes: #4513


HNicolas:
I don't think that the problem with the CI build is correlated with my changes.
Do I have to fix something ?

pleerock:
Sorry for being late here, I think we can merge this one. Thanks for contribution!

PR #4531, opened 2019-08-02
MsSql: stringy port exception

arkady-zelensky:
There was a problem with MsSql driver - [tedious](https://github.com/tediousjs/tedious), when getting DB connection parameters from environment. There was [a type checking of number](https://github.com/tediousjs/tedious/blob/master/src/connection.js#L534), but port, passed from TypeORM was a stringy number, so it throws an error.

Solution: convert port to number.

### DIFF
--- a/src/connection/options-reader/ConnectionOptionsEnvReader.ts
+++ b/src/connection/options-reader/ConnectionOptionsEnvReader.ts
@@ -21,7 +21,7 @@ export class ConnectionOptionsEnvReader {
             type: PlatformTools.getEnvVariable("TYPEORM_CONNECTION") || (PlatformTools.getEnvVariable("TYPEORM_URL") ? PlatformTools.getEnvVariable("TYPEORM_URL").split("://")[0] : undefined),
             url: PlatformTools.getEnvVariable("TYPEORM_URL"),
             host: PlatformTools.getEnvVariable("TYPEORM_HOST"),
-            port: Number(PlatformTools.getEnvVariable("TYPEORM_PORT")),
+            port: PlatformTools.getEnvVariable("TYPEORM_PORT") && !isNaN(PlatformTools.getEnvVariable("TYPEORM_PORT")) ? Number(PlatformTools.getEnvVariable("TYPEORM_PORT")) : undefined,
             username: PlatformTools.getEnvVariable("TYPEORM_USERNAME"),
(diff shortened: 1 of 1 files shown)


Kononnable:
Which version of tedious driver? Did they publish fixed version?
There are plenty of options to pass config to typeorm, so if it has to be addressed in our code there might be other lines to be changed. Also `+` does the job done, but it is cryptic, easy to miss - `Number(x)` should be easier to read.

arkady-zelensky:
Used tedious driver of version 4.2.0.
No, they didn't. In the current version - 6.3.0 problem still wasn't solved.
Changed parsing function from `+` to `Number()`.

pleerock:
Maybe we should check if PlatformTools.getEnvVariable("TYPEORM_PORT") is really defined before doing it, since I'm not sure what Number(PlatformTools.getEnvVariable("TYPEORM_PORT")) will return. (I guess later there is a check if value is defined and use default port in case if its not, and some NaN can mess things up)

niccolofanton:
> Maybe we should check if PlatformTools.getEnvVariable("TYPEORM_PORT") is really defined before doing it, since I'm not sure what Number(PlatformTools.getEnvVariable("TYPEORM_PORT")) will return. (I guess later there is a check if value is defined and use default port in case if its not, and some NaN can mess things up)
@arkady-zelensky can you add this check before converting to int? Thanks 
`port: PlatformTools.getEnvVariable("TYPEORM_PORT") && !isNaN(PlatformTools.getEnvVariable("TYPEORM_PORT")) ? Number(PlatformTools.getEnvVariable("TYPEORM_PORT")) : undefined`

arkady-zelensky:
@niccolofanton 
Done! Sorry for so late response :)

pilleye:
Hi, is there any update on this PR? The issue still seems to be persisting for me.

arkady-zelensky:
Fortunately, no :(

pilleye:
Looks like I've got to use Postgres for the time being :(

Chup4Chups:
Hi, any ETA on this ? 
we are using typeorm with mssql and we face the error below when trying to run migration from .env:
TypeError: The "config.options.port" property must be of type number.

imnotjames:
This was fixed in a different PR.

PR #4549, opened 2019-08-06
docs: update outdated Expo SQLite link

cruzach:
Self-explanatory title. The old link would result in a 404, and that's no fun


rustamwin:
Thank you!

Issue #1705, opened 2018-03-06
AfterUpdate subscriber bug


bo20202:
**Issue type:**

[ ] question
[x] bug report
[ ] feature request
[ ] documentation issue

**Database system/driver:**

[ ] `cordova`
[ ] `mongodb`
[ ] `mssql`
[ ] `mysql` / `mariadb`
[ ] `oracle`
[x] `postgres`
[ ] `sqlite`
[ ] `sqljs`
[ ] `websql`

**TypeORM version:**

[ ] `latest`
[ ] `@next`
[x] `0.1.9` (or put your version here)

**Steps to reproduce or a small repository showing the problem:**

1. Create entity.
2. Add method with `@AfterUpdate` decorator to it.
3. Update entity (with `getRepository(Entity).save()`).

**Expected behaviour:** 
When method with `@AfterUpdate` runs, I can get updated entity from database.

**Actual behaviour:**
Here's what actually happens:

1. Transaction is created (`executing query:  START TRANSACTION`)
2. `UPDATE` query is executed (`executing query:  UPDATE "entity" SET "something"=$1 WHERE "id"=$2 -- PARAMETERS: [1,2]`)
3. Method with `@AfterUpdate` is called.
4. Transaction is commited.

So, when `@AfterUpdate` method is called, entity in the database has not been updated yet. Therefore, I can not get updated entities from the database.

pleerock:
It is expected. If something is failing in AfterUpdate transaction should be reverted, it is by design.

bo20202:
@pleerock so, is there any way to notify listeners that something was updated or inserted?

pleerock:
Listeners and subscribers already do it.

bo20202:
@pleerock they are notified before transaction is commited. I need to notify listeners when the data is already in the database. How to do it proprerly?

pleerock:
Depend of your needs.

You can call your method without listeners simply after you call `save` method.

But I guess you need a data that inserted but not yet committed, in this case you need to use same `entityManager` instance (transactional entity manager), for this purpose you need to use subscribers instead of listeners since subscribers pass you `event` object with `entityManager` inside you'll need to use to get the data using transactional entity manager and you'll have what you need.

bo20202:
> But I guess you need a data that inserted but not yet committed
No, I need a confirmation that data is updated AND already in the database.  `AfterUpdate` confirms that data is updated, but not in the database.

pleerock:
[...]

PR #4554, opened 2019-08-08
pass [update|insert|remove|load]-events to listeners (as it's done for subscribers)

daweedm:
see #1705 

Make uniform the parameters received by subscribers and listeners

### DIFF
--- a/src/metadata/EntityListenerMetadata.ts
+++ b/src/metadata/EntityListenerMetadata.ts
@@ -66,11 +66,11 @@ export class EntityListenerMetadata {
     /**
      * Executes listener method of the given entity.
      */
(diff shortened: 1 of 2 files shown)


pleerock:
Can you please add tests to this functionality? (quite easy). Thanks!

Kononnable:
ping @daweedm

neuro-sys:
Looks like this would be very useful. As I just realized that it's not possible to use the surrounding transaction from inside Entity Listeners. So I am planning to move all such logic to Entity Subscribers which get an event object containing the surrounding transaction instance.

imnotjames:

Issue #4559, opened 2019-08-08
Add options to input config when initializing sql.js


DavidChen-minted:
**Issue type:**

[ ] question
[ ] bug report
[x] feature request
[ ] documentation issue

**Database system/driver:**

[ ] `cordova`
[ ] `mongodb`
[ ] `mssql`
[ ] `mysql` / `mariadb`
[ ] `oracle`
[ ] `postgres`
[ ] `cockroachdb`
[x] `sqlite`
[ ] `sqljs`
[ ] `react-native`
[ ] `expo`

**TypeORM version:**

[x] `latest`
[ ] `@next`
[ ] `0.x.x` (or put your version here)

Typeorm added supports for Sql.js v1.0 https://github.com/typeorm/typeorm/issues/4104 but it needs a new connection option to pass in initConfigs, such as [`{ locateFile: filename => ``/dist/${filename}`` }`](https://github.com/kripken/sql.js/tree/fbe8427c32d5077be06bb808283c2efb22a84c97#upgrading-from-0x-to-1x).

exequielc:
Hi,  I'm using this old sql.js v0.5.0  and don't understand how to change the conection in order to upgrade to 1.0

I'm using the Ionic implementation.

```
let dbOptions: ConnectionOptions;
 dbOptions = {
          type: 'sqljs',
          location: 'browser',
          autoSave: true
 };
 Object.assign(dbOptions, {
      synchronize: true,
      autoSchemaSync: true,
      migrationsRun: true,
      entities: [    tables   ]
    });
     return createConnection(dbOptions).then(connection => {
      console.log("Data Access connected!");
      return connection;
    }).catch(error => {
      //console.log("Data Access Error : ", );
      return null;
    });

```

PR #4560, opened 2019-08-08
add options to input init config for sql.js

DavidChen-minted:
Fixes https://github.com/typeorm/typeorm/issues/4559

added `sqlJsConfig` connection option for sql.js (v1.x) to inititialize.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,7 @@ feel free to ask us and community.
 * extend afterLoad() subscriber interface to take LoadEvent ([issue #4185](https://github.com/typeorm/typeorm/issues/4185))
 * relation decorators (e.g. `@OneToMany`) now also accept `string` instead of `typeFunction`, which prevents circular dependency issues in the frontend/browser ([issue #4190](https://github.com/typeorm/typeorm/issues/4190))
 * added support for metadata reflection in typeorm-class-transformer-shim.js ([issue #4219](https://github.com/typeorm/typeorm/issues/4219))
+* added `sqlJsConfig` to input config when initializing sql.js ([issue #4559](https://github.com/typeorm/typeorm/issues/4559))
 
 ## 0.2.17 (2019-05-01)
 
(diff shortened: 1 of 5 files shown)


exequielc:
Hi,  I'm using this old sql.js v0.5.0  and don't understand how to change the conection in order to upgrade to 1.0

I'm using the Ionic implementation.

```
let dbOptions: ConnectionOptions;
 dbOptions = {
          type: 'sqljs',
          location: 'browser',
          autoSave: true
 };
 Object.assign(dbOptions, {
      synchronize: true,
      autoSchemaSync: true,
      migrationsRun: true,
      entities: [    tables   ]
    });
     return createConnection(dbOptions).then(connection => {
      console.log("Data Access connected!");
      return connection;
    }).catch(error => {
      //console.log("Data Access Error : ", );
      return null;
    });

```

DavidChen-minted:
> Hi, I'm using this old sql.js v0.5.0 and don't understand how to change the conection in order to upgrade to 1.0
> 
> I'm using the Ionic implementation.

I never learnt Ionic but I think the current version of typeorm can adapt either sql.js v1.0 or v0.5 under the layer. The [browser example](https://github.com/typeorm/browser-example/blob/master/src/index.ts) would work for sql.js v1.0 as well. This pr is only to allow adding optional connection option w.r.t. v1.0.

exequielc:
I'v got this error with the same work config connection for the prev version  "sql.js": "^0.5.0".

`SqljsDriver.js:319 Uncaught Error: Cannot find module "sql.js/js/sql.js"`

clfhhc:
> I'v got this error with the same work config connection for the prev version "sql.js": "^0.5.0".
> 
> `SqljsDriver.js:319 Uncaught Error: Cannot find module "sql.js/js/sql.js"`
> 
> There's no example to use the new version of sql 1.0.0

I'm posting my working example here for those who are interested:
https://github.com/Project-Setup/github_sql_pwa#typeorm-and-sqljs

Issue #4570, opened 2019-08-11
PrimaryColumn decorator modifies passed options


sthuck:
**Issue type:**

[ ] question
[x ] bug report
[ ] feature request
[ ] documentation issue

**Database system/driver:**

[ ] `cordova`
[ ] `mongodb`
[ ] `mssql`
[ ] `mysql` / `mariadb`
[ ] `oracle`
[ ] `postgres`
[ ] `cockroachdb`
[ ] `sqlite`
[ ] `sqljs`
[ ] `react-native`
[ ] `expo`
[x] all/not db specific
**TypeORM version:**

[x] `latest`
[ ] `@next`
[ ] `0.x.x` (or put your version here)

some times it's useful to have a common column settings object.  It's good if you need some compatibility between different db's, or even just for plain readability. 
```ts
const dbConf = {
  idColumn: {nullable: false, type: 'varchar', length: 36},
  json: {type: 'json', nullable: true},
  shortTextField: {type: 'varchar', length: 64},
  numeric: {type: 'int'}
}

//Entity1.ts
@Entity
export class Entity1 {
  @PrimaryColumn(dbConf.idColumn)
   id!: string;

  @Column(dbConf.shortTextField)
   someColumn: string;
}

//Entity2.ts
@Entity
export class Entity2 {
  @PrimaryColumn(dbConf.numeric)
   id: number;

  @Column(dbConf.idColumn)
   someIdColumn: string;
}
```

unfortunately, `PrimaryColumn` modifies the passed options, and now `Entity2` will have a compound key, as `dbConf.idColumn` has changed to include `primary: true`.

PR #4571, opened 2019-08-11
fix: change PrimaryColumn decorator to clone passed options

sthuck:
Closes: #4570


pleerock:
think we can merge it

Issue #4598, opened 2019-08-18
Sqlite connections dont ignore the schema property on entities


roman-acumen:
**Issue type:**

[ ] question
[x] bug report
[ ] feature request
[ ] documentation issue

**Database system/driver:**

[ ] `cordova`
[ ] `mongodb`
[ ] `mssql`
[ ] `mysql` / `mariadb`
[ ] `oracle`
[x] `postgres`
[ ] `cockroachdb`
[x] `sqlite`
[ ] `sqljs`
[ ] `react-native`
[ ] `expo`

**TypeORM version:**

[x] `latest`
[ ] `@next`
[ ] `0.x.x` (or put your version here)

The Entity schema property is not ignored by databases that don't need it

**Steps to reproduce or a small repository showing the problem:**

```ts 
@Entity("test", { schema: "test" })
export class Test {
	@PrimaryColumn()
	public id: number;

	@Column()
	public test: string;
}
```
For postgres all will work ok, both syncronize and find will work ok:
``` ts
const connection = await typeorm.createConnection({
			type: "postgres",
			url: databaseConnectionUrl,
			entities: [Test]
		});
await connection.synchronize(true);
const res = await connection.getRepository(Test).save(new Test(42, "Meaning of life"));
const res2 = await connection.getRepository(Test).findOrFail({id: 42});
```

The same code on sqlite will pass the creation, but will fail on working

``` ts
const connection = await typeorm.createConnection({
			type: "sqlite",
			database: ":memory:",
			entities: [Test]
		});
await connection.synchronize(true);
// This will pass and all tables will be created
const res = await connection.getRepository(Test).save(new Test(42, "Meaning of life"));
// This will fail saying table test.test doesn't exist
const res2 = await connection.getRepository(Test).findOrFail({id: 42});
```

PR #4599, opened 2019-08-18
Github #4598 issue fix

roman-acumen:
Fix for issues #4598


pleerock:
AFAIK sqlite doesn't support schemas, whats purpose of this change?

roman-acumen:
In all places across the code (for example table creation) sqlite driver ignores this property. This is the only place where it doesn't. 

Also as I would expect that drivers will ignore properties not related to them - hence the fix.

Issue #4608, opened 2019-08-19
entityManager.delete() proving to be a big footgun


osdiab:
**Issue type:**

[x] question
[ ] bug report
[ ] feature request
[ ] documentation issue

**Database system/driver:**

[ ] `cordova`
[ ] `mongodb`
[ ] `mssql`
[ ] `mysql` / `mariadb`
[ ] `oracle`
[x] `postgres`
[ ] `cockroachdb`
[ ] `sqlite`
[ ] `sqljs`
[ ] `react-native`
[ ] `expo`

**TypeORM version:**

[x] `latest`
[ ] `@next`
[ ] `0.x.x` (or put your version here)

**Steps to reproduce or a small repository showing the problem:**

If you do this, what happens?

```ts
entityManager.delete(User, { where: { name: "Omar" } });
```

Answer: It deletes every `User` in your database, because `delete` doesn't take `FindOptions`, and therefore the `where` key appears to be completely ignored.

This is easily avoidable by using it as the documentation says, and instead doing `delete(User, { name: "Omar" })`; but since TypeScript won't tell you that you're doing anything wrong due to the overly general types, my team is finding this to be a particularly volatile API choice for TypeORM to not warn against.

Better would be stricter types to forbid usage of keys that aren't a part of `User`, but as a general complaint, wish the TypeORM APIs just generally had stricter enforcement of a consistent API invocation syntax.

imnotjames:
Duplicate of #3416

PR #4609, opened 2019-08-20
FindConditions instead of any for EntityManager.delete()

osdiab:
Context is in #4608 . I'm not 100% sure if this is correct for using `delete` properly (it passes the code directly to a `QueryBuilder.where()` clause, which I think can take `FindConditions`? But I'm not totally sure if that's how `QueryBuilder` works.

Having something like this would save my team and probably other's from blowing up their own dbs.

### DIFF
--- a/src/entity-manager/EntityManager.ts
+++ b/src/entity-manager/EntityManager.ts
@@ -533,7 +533,7 @@ export class EntityManager {
      * Does not check if entity exist in the database.
      * Condition(s) cannot be empty.
      */
-    delete<Entity>(targetOrEntity: ObjectType<Entity>|EntitySchema<Entity>|string, criteria: string|string[]|number|number[]|Date|Date[]|ObjectID|ObjectID[]|any): Promise<DeleteResult> {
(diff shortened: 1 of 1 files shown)


pleerock:
it won't support all FindConditions features, only some trivial ones

Kononnable:
Ok, so what should we do about this PR? Close it or change `FindConditions` to some subtype of it which would contain only features that work with `delete`? There are also tests missing which I consider necessary with such API.

@osdiab Are you still interested in this PR?

osdiab:
It would have been nice for past me but I'm used to it now that I'm not messing as much with the particular code - I do think it should probably go through in some form but I'm not sure what exactly to put there, as you've highlighted. If there's documentation somewhere about what `delete` can do that would be useful, otherwise I can delve into the `delete` code and grok it directly

osdiab:
This is the query builder code that criteria eventually gets added to: https://github.com/typeorm/typeorm/blob/master/src/query-builder/DeleteQueryBuilder.ts#L144

osdiab:
That would imply that a general, potentially correct option might be `string|((qb: QueryBuilder<Entity>) => string)|Brackets|ObjectLiteral|ObjectLiteral[]`, though that is quite general/footgun prone

pleerock:
On the next branch we already have this change, and I suppose because its working better in there.

AlexMesser:
Closing for now.

Issue #4630, opened 2019-08-24
Enum string not escaping resulting in broken migrations.


duckies:
**Issue type:**

[ ] question
[*] bug report
[ ] feature request
[ ] documentation issue

**Database system/driver:**

[ ] `cordova`
[ ] `mongodb`
[ ] `mssql`
[ ] `mysql` / `mariadb`
[ ] `oracle`
[*] `postgres`
[ ] `cockroachdb`
[ ] `sqlite`
[ ] `sqljs`
[ ] `react-native`
[ ] `expo`

**TypeORM version:**

[*] `latest` (_0.2.18 for reference_)
[ ] `@next`
[ ] `0.x.x` (or put your version here)

**Steps to reproduce or a small repository showing the problem:**

I am encountering an odd issue where providing an enum of strings with an apostrophe in it is not being escaped, so the SQL fails. A small repository that demonstrates the issue along with the full error is [here](https://github.com/duckies/nest-enum-escaping).

```Typescript
export enum Realm {
  Nerzhul = "Ner'zhul",
}

@Entity("character")
export class Character {
  @PrimaryGeneratedColumn()
  id: number;

  @Column({ type: "enum", enum: Realm })
  realm: Realm;
}

```
Produces the following incorrect query string for the realm column type:
``CREATE TYPE "character_realm_enum" AS ENUM('Ner'zhul')``

Trying to format the enum differently or escape it in the code did not help, Typeorm always generated incorrect migrations. I also don't know how to stop Typeorm from trying to create migrations for this field. Any help is appreciated.

pleerock:
fixed by #4631

PR #4631, opened 2019-08-24
Fix: Apostrophe in Postgres enum strings breaks query

duckies:
Patch for #4630

I'm not a SQL hero, but a simple way of escaping an apostrophe in this context is to simply use two apostrophes and this appears to work well for me. This is likely an issue in other drivers, but I am not familiar.


pleerock:
Thanks!

PR #4680, opened 2019-09-02
docs: Update many-to-many-relations.md, make it easier to understand

QoVoQ:
Make the example of the last section `many-to-many relations with custom properties` more clear and general.


pleerock:
Thanks!

Issue #3330, opened 2018-12-26
Creating a WITHOUT ROWID table


P0oOOOo0YA:
**Issue type:**

[*] question

**Database system/driver:**
[*] `sqlite`

**TypeORM version:**
[*] `latest`

In SQLite, one can create a [`WITHOUT ROWID`](https://www.sqlite.org/withoutrowid.html) table with the following query.

```sql
CREATE TABLE IF NOT EXISTS wordcount(
  word TEXT PRIMARY KEY,
  cnt INTEGER
) WITHOUT ROWID;
```
What is its equivalent in TypeORM?

pleerock:
Its not supported. We can make `@Entity` to accept this option and use in during query building. Feel free to contribute.

PR #4688, opened 2019-09-04
feat: implements Sqlite 'WITHOUT ROWID' table modifier

bnegrao:
This new feature adds a 'withoutRowid' option to EntityOptions in order to
enable Sqlite 'WITHOUT ROWID' modifier to the 'CREATE TABLE' statement.
See https://www.sqlite.org/withoutrowid.html

Closes: #3330


pleerock:
Thanks for contribution!

Looks like you don't track changes of `withoutRowid` (let's say it was set to true, then suddenly you decided to set to false) and schema sync won't catch up this change and your database won't be synchronized

bnegrao:
Hello Umed, thank you for reviewing. 

Can you point me to some code that does that tracking so I can have that as a guide in my implementation?

pleerock:
its in the same sqlite query runner file, you need to figure solution for your feature.

bnegrao:
Hmmm, SQlite [does not support](https://www.sqlite.org/omitted.html) an "alter table" command that would change this parameter after the table is created. 
That means if someone changes that option after a table was created that table must be recreated. Also if the table had relations with other tables those relations would have to be rebuilt too because the related primary key has changed.

What should I do in that case?

I am super new to this framework, if you can give any help that would be greatly appreciated.

bnegrao:
Changing the _withoutRowid_ option would be a "secret" change in the primary key of the table. Why secret? Because when you disable that option SQlite [will secretly create an INTEGER primary key for the table, and what you defined as the primary key will be secretly indexed](https://www.sqlite.org/withoutrowid.html).

The current implementation[ _AbstractSqliteQueryRunner.ts_](https://github.com/typeorm/typeorm/blob/63d993b1ef9f68a6e12bd535ac8ca8be39702502/src/driver/sqlite-abstract/AbstractSqliteQueryRunner.ts) does nothing when there is an update in the primary keys, see:

    /**
     * Updates composite primary keys.
     */
    async updatePrimaryKeys(tableOrName: Table|string, columns: TableColumn[]): Promise<void> {
        await Promise.resolve();
    }

Given that, i believe it does not make sense to track for changes in that property. Agree?

bnegrao:
> its in the same sqlite query runner file, you need to figure solution for your feature.

Can you please read my comment above?

pleerock:
Right, sqlite doesn't support alter table command, that's why we always re-create table (and backup and restore all the data) after anything changed in your schema.

Example: 

* you added a new column in entity
* we read columns we have in the database
* we read your entity columns
* we find a difference - a new column added
* since we can't alter we re-create table with a new column added
  (for other dbs we just call alter)
* that's it

same process with just everything to make sure your code in the entity always match your database representation. Same you need to do - read information about `WITHOUT ROWID` check if it exist or not in the table and in the entity, compare, do the job.

bnegrao:
But the problem with the _withoutRowid_ option is it is an invisible change in the schema, that option defines the creation of the _rowid_ column that is not declared in your Entity class.

This method from _RdbmsSchemaBuilder.ts_ will never be able to detect that there should be modification in the schema:

    /**
     * Updates composite primary keys.
     */
    protected async updatePrimaryKeys(): Promise<void> {
        await PromiseUtils.runInSequence(this.entityToSyncMetadatas, async metadata => {
            const table = this.queryRunner.loadedTables.find(table => table.name === metadata.tablePath);
            if (!table)
                return;

            const primaryMetadataColumns = metadata.columns.filter(column => column.isPrimary);
            const primaryTableColumns = table.columns.filter(column => column.isPrimary);
            if (primaryTableColumns.length !== primaryMetadataColumns.length && primaryMetadataColumns.length > 1) {
                const changedPrimaryColumns = primaryMetadataColumns.map(primaryMetadataColumn => {
                    return new TableColumn(TableUtils.createTableColumnOptions(primaryMetadataColumn, this.connection.driver));
                });
                await this.queryRunner.updatePrimaryKeys(table, changedPrimaryColumns);
            }
        });
    }


That last line `await this.queryRunner.updatePrimaryKeys(table, changedPrimaryColumns);` would never be invoked with the change of the _withoutRowid_ option.

Am I mistaken? do you have any ideas?

bnegrao:
And I don't understand why are you asking to implement that right now if the current implementation of _updatePrimaryKeys_ in [ _AbstractSqliteQueryRunner.ts_](https://github.com/typeorm/typeorm/blob/63d993b1ef9f68a6e12bd535ac8ca8be39702502/src/driver/sqlite-abstract/AbstractSqliteQueryRunner.ts) does nothing:

    /**
     * Updates composite primary keys.
     */
    async updatePrimaryKeys(tableOrName: Table|string, columns: TableColumn[]): Promise<void> {
        await Promise.resolve();
    }

If the users of typeorm + SQLite were happy with that implementation why do we need to worry now?

pleerock:
Okay let's merge it, since this feature most likely will be used during first-time table creation its okay for now not to have auto-syncing feature.

olosegres:
@pleerock Please tell me, is code from this PR lets use `withoutRowid` option with `EntitySchema`?

pthrasher:
@olosegres I've opened a PR to add the support you're requesting here: #8432

PR #4698, opened 2019-09-05
Update view-entities.md

Mikkal24:
docs: update view-entities.md

Add @PrimaryColumn() decorator to at least @ViewColumn(). Without the decorator we can't use repository methods such as count(). Was difficult to find the solution so I wanted to include this in the docs to save others time.

### DIFF
--- a/docs/view-entities.md
+++ b/docs/view-entities.md
@@ -109,6 +109,7 @@ import {ViewEntity, ViewColumn} from "typeorm";
 export class PostCategory {
 
     @ViewColumn()
+    @PrimaryColumn()
     id: number;
 
     @ViewColumn()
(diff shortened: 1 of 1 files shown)


pleerock:
I don't think we should merge this one, since right now its not by design. Instead we should create an issue with the description of the problem for the future to design it and build a proper solution.

Mikkal24:
Okay! I'll open an issue today. Can I use this PR for reference? @pleerock

pleerock:
yes, sure. Thanks @Mikkal24

Mikkal24:
Added issue -> https://github.com/typeorm/typeorm/issues/4699

pleerock:
Thanks!

Issue #296, opened 2017-02-18
select additional computed columns


D4edalus:
assuming i have some table with a column name, that may be null
in a query i would want to have a computed column "hasName", that is true if the name field is not null

i could do that as follows with the query builder
```
... createQueryBuilder("person")
.addSelect("name IS NOT NULL as hasName")
```
this is of course not mapped, and therefore only visible in "getRawMany()" .. but it shows what i want to achieve.

assume now i have a query for orders

```
... createQueryBuilder("order")
.leftJoinAndSelect("order.person", "person")
.addSelect("person.name IS NOT NULL as person.hasName") // does NOT work
```

question1: can addSelect be mapped to a custom variable, like it can be done with leftJoinAndMapOne?
question2: is there something like addSelectAndMap?

pleerock:
no there is no `addSelectAndMap` feature right now but I had plans to implement it. So its a feature request for the future.

vincentbouillart:
When will this new feature be present ?

voodooattack:
@pleerock Any news on this? I am also interested in this feature. I can contribute if you point me in the right direction.

pleerock:
not yet sure about design of this feature yet, but I think it will land somewhere in `0.4.0`

ToonvanStrijp:
For everyone that wants a work around: #1822

vlapo:
@pleerock Any progress in this feature? Some design, feature spec?

ekifox:
Update...

Offlein:
There might be a flaw in my reasoning, but I believe this can be accomplished by specifying extra properties on your entity, and then implementing the `@AfterLoad()` decoration?

Such as:

```
@Entity()
export class User {
  @PrimaryGeneratedColumn()
  id: number;

  @Column()
  firstName: string;

  @Column()
  lastName: string;

  fullName: string;

  @AfterLoad()
  setComputed() {
    this.fullName = this.firstName + ' ' + this.lastName;
  }
}
```

Does that accomplish what people are looking for? Apologies if not.

dan-h-ch:
So this does work, but I'm not sure it is exactly the same as being able to modify query builder with a `addSelectAndMap`. It feels a bit much to have to update the entity. For example, I'm trying to create a query to fetch days from expiration. I don't really care for that on the entity outside of this one situation.

But in a bind that works for now.

nikolovjd:
@pleerock Any updates on addSelectAndMap()?

[...]

PR #4703, opened 2019-09-06
fix: added support for selecting computed columns

Asarew:
This fixes some basic functionality to select computed columns discussed in issue #296 
```ts
@Entity()
export class Post {

    @PrimaryGeneratedColumn()
    id: number;

    @Column()
    title: string;

    @Column({ type: String, nullable: true })
    text: string|null;

    @Column({ nullable: true, insert: false, update: false, select: false })
    textSize: number;
}
```

```ts
const query = connection.manager.createQueryBuilder(Post, "post");
const loadedPost = await query
    .select("post.id")
    .addSelect("LENGTH(post.text)", `${query.alias}_textSize`)
    .addSelect("SUBSTRING(post.text, 1, 20)", "post_text")
    .getOne();
```

### DIFF
--- a/src/query-builder/transformer/RawSqlResultsToEntityTransformer.ts
+++ b/src/query-builder/transformer/RawSqlResultsToEntityTransformer.ts
@@ -137,8 +137,15 @@ export class RawSqlResultsToEntityTransformer {
 
             // if user does not selected the whole entity or he used partial selection and does not select this particular column
             // then we don't add this column and its value into the entity
-            if (!this.expressionMap.selects.find(select => select.selection === alias.name || select.selection === alias.name + "." + column.propertyPath))
+            if (!this.expressionMap.selects.find(select => {
+                return !!(
+                    select.selection === alias.name
+                    || select.selection === alias.name + "." + column.propertyPath
+                    || (select.aliasName && select.aliasName === alias.name + "_" + column.propertyPath)
+                );
+            })) {
                 return;
+            }
 
             column.setEntityValue(entity, this.driver.prepareHydratedValue(value, column));
             if (value !== null) // we don't mark it as has data because if we will have all nulls in our object - we don't need such object
--- a/test/github-issues/296/entity/Post.ts
+++ b/test/github-issues/296/entity/Post.ts
@@ -1,0 +1,19 @@
+import {Entity} from "../../../../src/decorator/entity/Entity";
+import {PrimaryGeneratedColumn} from "../../../../src/decorator/columns/PrimaryGeneratedColumn";
+import {Column} from "../../../../src/decorator/columns/Column";
+
+@Entity()
+export class Post {
+
+    @PrimaryGeneratedColumn()
+    id: number;
+
+    @Column()
+    title: string;
+
+    @Column({ type: String, nullable: true })
+    text: string|null;
+
+    @Column({ nullable: true, insert: false, update: false, select: false })
+    textSize: number;
+}
--- a/test/github-issues/296/issue-296.ts
+++ b/test/github-issues/296/issue-296.ts
@@ -15,7 +15,7 @@ describe("github issues > #296 select additional computed columns", () => {
     after(() => closeTestingConnections(connections));
 
 
-    it("should correctly substring the text column and populate the entity property", () => Promise.all(connections.map(async connection => {
+    it("should correctly select computed size of text column and populate given entity property", () => Promise.all(connections.map(async connection => {
         const post = new Post();
         post.title = "hello post";
         post.text = "Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiu";
@@ -33,6 +33,6 @@ describe("github issues > #296 select additional computed columns", () => {
         expect(loadedPost!.text).not.to.be.undefined;
         // Because some drivers return int's as strings's, so just to be sure we
         // cast both to string
-        expect("" + (loadedPost!.textSize)).to.be.equal("" + post.text.length);
+        expect("" + loadedPost!.textSize).to.be.equal("" + post.text.length);
     })));
 });
--- a/test/github-issues/296/issue-296.ts
+++ b/test/github-issues/296/issue-296.ts
@@ -1,0 +1,38 @@
+import "reflect-metadata";
+import {closeTestingConnections, createTestingConnections, reloadTestingDatabases} from "../../utils/test-utils";
+import {Connection} from "../../../src/connection/Connection";
+import {Post} from "./entity/Post";
+import {expect} from "chai";
+
+
+describe("github issues > #296 select additional computed columns", () => {
+
+    let connections: Connection[];
+    before(async () => connections = await createTestingConnections({
+        entities: [__dirname + "/entity/*{.js,.ts}"],
+    }));
+    beforeEach(() => reloadTestingDatabases(connections));
+    after(() => closeTestingConnections(connections));
+
+
+    it("should correctly substring the text column and populate the entity property", () => Promise.all(connections.map(async connection => {
+        const post = new Post();
+        post.title = "hello post";
+        post.text = "Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiu";
+        await connection.manager.save(post);
+
+        const query = connection.manager
+            .createQueryBuilder(Post, "post");
+        const loadedPost = await query
+            .select("post.id")
+            .addSelect("post.text")
+            .addSelect("LENGTH(post.text)", `${query.alias}_textSize`)
+            .getOne();
+
+        expect(loadedPost).not.to.be.undefined;
+        expect(loadedPost!.text).not.to.be.undefined;
+        // Because some drivers return int's as strings's, so just to be sure we
+        // cast both to string
+        expect("" + (loadedPost!.textSize)).to.be.equal("" + post.text.length);
+    })));
+});
--- a/test/github-issues/296/issue-296.ts
+++ b/test/github-issues/296/issue-296.ts
@@ -14,25 +14,28 @@ describe("github issues > #296 select additional computed columns", () => {
     beforeEach(() => reloadTestingDatabases(connections));
     after(() => closeTestingConnections(connections));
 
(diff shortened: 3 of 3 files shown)


TheNoim:
I hope this will get merged soon. Until then I will try to integrate this into an private mirror package. I really need this feature.

Asarew:
I'm thinking about turning this in to a WIP, so to discuss the options of adding the `addSelectAndMap` method's. As this won't require a lot of additional work. 🤔

TheNoim:
You should try fixing the CI.

Asarew:
Didn't notice the CI failed, pushed a fix for using proper function name based on driver.

tholander:
When does this release is planned for ?

Asarew:
@pleerock Thnx for assigning the reviewers! What are your thoughts on my [comment](https://github.com/typeorm/typeorm/pull/4703#issuecomment-528919535) i posted earlier?

Kononnable:
I don't like changing the entity model just for the sake of one query. Right now column is also stored in db(but value isn't persisted).
```typescript
@Column({ nullable: true, insert: false, update: false, select: false })
    textSize: number;
```
I think this would be ok(except for storing column in db), if column definition is set somewhere inside model(like another column decorator option). 

To be type-safety I think we should extend Entity type somehow and pass such type directly to query builder(or extend this type directly in query builder which would be better, but it might take more time).

Asarew:
In the test i provided 2 very broad way's on how this PR can be used: https://github.com/typeorm/typeorm/blob/0e402387fdde4ddf7921dfad61bfa64aeeb4778f/test/github-issues/296/issue-296.ts#L30...L31
One where i use a non existing column and mark the property as such(`textSize`), and another where i use an existing column (`text`).
This PR also works with regular properties and doesn't enforce a specific entity definition for it to work:
```typescript
    @Column({ type: String, nullable: true })
    text: string|null;
```

If we want to add another decorator and/or `@Column` option is in my opinion outside of the scope of this issue, as this PR is just a fix for missing(?) functionality.

when just selecting properties from the entity a query gets build with an alias that is used to transform the raw results in to entity results:
```typescript
query.select("post.id").addSelect("post.text");
```
will produce sql query:
```sql
SELECT
    `post`.`id` AS `post_id`,
    `post`.`text` AS `post_text` 
FROM
    `post` `post`
```
So i expected when i do something like:
```typescript
query
    .select("post.id")
    .addSelect("SUBSTR(post.text, 1, 20)", "post_text")
    .addSelect("LEN(post.text)", "post_textSize");
```
which produces SQL
```sql
SELECT
    "post"."id" AS "post_id",
    SUBSTR("post"."text", 1, 20) AS "post_text",
    LEN("post"."text") AS "post_textSize" 
FROM
    "post" "post" 
```
that they would map to the corresponding entity properties.

The idea behind this implementation is to let definition of the entity up to the end user, and just expose a basic expected `selectAndMap` functionality.

Asarew:
That being sad, it's possible to add on top of this PR something like:
```typescript
class Post {
    @Computed("LEN(post.text)")
    // or just
    @Virtual() // which would just allow the entity transformer to populate this entiy when the correspond alias is used.
    textSize: number;
}
```
and combined that with the `*SelectAndMap` methods that have been discussed in: #296

pleerock:
yeah so I agree with @Kononnable  comment. if you want to add `select`, `selectAndMap` and other functionality - all these values should be computed, and we definitely shouldn't have a column in the database. Having a column would bring issues in the future and its completely unexpected for users of this functionality.

Asarew:
@pleerock This doesn't add any `selectAndMap` functionality. My goal with this PR is to **fix** building of computed select queries and map the values to their corresponding properties. 

Because the way that the query builder API works today, i expected that this would already work. 
How the user decides to use this missing functionality is (imho) up to them. The test/examples that i provided; are nothing more that a show case of the broad use cases that come with this fix. (see [my comment](https://github.com/typeorm/typeorm/pull/4703#issuecomment-534943637) for a detailed explanation)

So based on your comment i want to make my self clear:
- It's not required/enforced **AT ALL** to have additional (virtual) columns, this fix is for any/every type of entity configuration.
- It allows a lot of requested features to be build upon this fix; but that is from my perspective outside the scope of this PR. (like: `selectAndMap`, computed & virtual properties)

Asarew:
@Kononnable can you please merge this PR?

twittwer:
@pleerock @Kononnable @vlapo It would be really helpful to see this merged 🙏

duckies:
I encountered this issue today as it would have seemed natural that TypeORM allow you to create computed selects. Reviewing this change, since it can be used without any side effects on the entity schema I don't see why this could not be merged. `selectAndMap` would be nice to have, but it was talked about in 2017 and there's no current progress on the feature. The recommendation of using multiple queries is not dissimilar than saying never use computed selectors, which I don't find acceptable

yasirhantoush:
@Asarew , will this work without using the column decorater in the entity as in below example:

```
@Entity()
export class Post {

    @PrimaryGeneratedColumn()
    id: number;

    @Column()
    title: string;

    @Column({ type: String, nullable: true })
    text: string|null;

    textSize: number;
}
```

If so ,, then there is no concern / worries about users having to introduce a fake column in the db ,, I am facing this issue right now and ready to use this fix given no need to add extra columns in my tables.

janbuerling:
Is there any reason that this PR is not merged yet? Is it possible to merge it soon, please?

Thanks :)

Asarew:
@pleerock i'm currently working on adding the feature to 'hide' a column from the database, and i hope you can give me some input on the direction i'm planning to take:

- Add a column option `hidden: boolean`
- Filter out hidden columns in the SchemaBuilder's (method: `executeSchemaSyncOperationsInProperOrder`)
- handle the hidden columns by default as they would have the options: `{ nullable: true, insert: false, update: false, select: false }`

Please let me know if i'm missing more items on this list

**@typeorm team** (@vlapo, @Kononnable) please let me know if you guys can meet me half-way and give this pull request a little bit more priority, love & support. if so, I'll be willing to invest my time to add this feature.
i'm doing this not because i need the have the hidden column support (because i don't sync entities). but just to speed up the merging of this pull request.

wojciechbator:
+1

This is awesome @Asarew that You've implemented this stuff. Right now, the common case for this is grabbing the count of some relation (like `employeesCount` in `Companies` table) and sorting over this column - it's very hard to do without either writing the whole sql in raw query (which I'd like to avoid, that why we're using ORM, right?) or second queries and some glueing.

nilswx:
Please merge this. TypeORM has been without `selectAndMap` since it was first mentioned in 2017. At this rate I doubt that it's coming anytime soon.

The only solution right now is to use `getRaw...` and stitch the data together yourself.

wojciechbator:
@nilswx I used the mentioned method, but the produced column is not sortable in paginated table. Instead, the thing that works is to add subquery with count as a result, this way it works. Using subquery to have sortable aggregated result is not quite efficient. In generał, it would be perfect if TypeORM supports selecting aggregated results and I think that's one of the reasons they try with `addSelectAndMap` method

STIENLILLY:
MERGE PLEASE

giorgiofellipe:
@pleerock @Kononnable @vlapo guys, please...

JoaquinAraujo:
We need this :(

Overfacto:
any news?

danielkv:
+1

pavel-luhin:
+1
I cannot use `loadRelationCountAndMap` because I am joining on a table without relations and foreign keys. Really need this feature to use aggregated results in such cases.

duprez:
Hi guys, please check out my [pr(6855)](https://github.com/typeorm/typeorm/pull/6855) about the commented feature (`addSelectAndMap`) and feel free to give any feedback

danielkv (approved):

ZhaoKunLong (approved):

ZhaoKunLong:
I really need this part. now I use the `getRawAndEntities()`  then merge the Raw result and Entitie result.

ZhaoKunLong:
I really need this part. now I use the getRawAndEntities() then merge the Raw result and Entitie result.

NIV54:
Where is this standing? This is still a wanted feature.
If the only problem is the extra DB column there are acceptable offers made by @Asarew with `@Column({ hidden: true })` or by @yasirhantoush with his idea of not using a decorator at all.
A new decorator such as `@CumputedColumn` which will not create an actual DB column can also be proposed (this way future editing of this feature will be easier since the decorator is already there and no breaking changes will be necessary).

pleerock:
Okay, I'm back to give some love to this PR.
Obviously it can't be merged like this, even if it's simple, the usage isn't obvious.
We need to came up with a proper design, most probably it should be `selectAndMap`.

But this issue is 3(!) years old and I'm afraid @Asarew already loosed his interest in this PR.
@Asarew let me know if you are still in, otherwise we should close this PR and wait for a new implementation.

Asarew:
@pleerock I dont work for the company/project anymore. But I'm in! 
What are you requirements? The discussed issues in this thread are:
* Mark a column with a decorator to make sure a column isn't created.
* Add additional method on query builder (`addSelectAndMap`)

Am I missing something?

TheNoim:
Maybe first there should be a proposal how the ´selectAndMap´ should work. Then somebody can start working on a PR.

Asarew:
I would agree but I expect that the resources necessary to make a good proposal are not available for the typeorm team.

Asarew:
closed due to lack of attention

Jedliu:
That's unbelievable! People ask for this merge for more than 3 years but I didn't see it in the latest release. And I did some research on the issues and many more related found as well. Can't understand!

Asarew:
just use MikroOrm, it's written from the ground up with a higher quality of code than Typeorm. From my perspective this project is not only dead, but has also become irrelevant as MikroOrm does (almost) the same thing only better.

SnowingFox:
i can't understand, i feel dispointed to this project

giorgiofellipe:
The TypeORM is dead, long live MikroORM!

Issue #4701, opened 2019-09-06
Duplicate migrations are executed.


jurasan:
**Issue type:**

[ ] question
[x] bug report
[ ] feature request
[ ] documentation issue

**Database system/driver:**

[ ] `cordova`
[ ] `mongodb`
[ ] `mssql`
[ ] `mysql` / `mariadb`
[ ] `oracle`
[ ] `postgres`
[ ] `cockroachdb`
[ ] `sqlite`
[ ] `sqljs`
[ ] `react-native`
[ ] `expo`

**TypeORM version:**

[x] `latest`
[ ] `@next`
[ ] `0.x.x` (or put your version here)

**Steps to reproduce or a small repository showing the problem:**

Create two migrations in 2 files, but with the same class name.

There are two scenarios:

1. You run one migration and then another in 2 steps. Then the second migration is skipped, because its class name is already present in the migrations table.
2. You run both migrations in one step. This time both migrations are executed

Thus we have different behavior. This creates a problem when you make an error copy-pasting migrations and not changing class name, and the first migration is already applied. When you run tests that run all migrations at once all the migrations will be executed despite duplicate class name. When you apply it on live database, the second migration will be skipped.

PR #4704, opened 2019-09-06
Fix 4701

jurasan:
fixing issue with duplicate migrations https://github.com/typeorm/typeorm/issues/4701

### DIFF
--- a/src/migration/MigrationExecutor.ts
+++ b/src/migration/MigrationExecutor.ts
@@ -357,9 +357,9 @@ export class MigrationExecutor {
             values["timestamp"] = migration.timestamp;
             values["name"] = migration.name;
         }
-        if (this.connection.driver instanceof MongoDriver) {
+        if (this.connection.driver instanceof MongoDriver) {  
             const mongoRunner = queryRunner as MongoQueryRunner;
-            mongoRunner.databaseConnection.db(this.connection.driver.database!).collection(this.migrationsTableName).insert(values);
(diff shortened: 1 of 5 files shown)


ssonne:
Shouldn't the fix be to throw an error if there are duplicate migrations? In what circumstance would you want to allow duplicate migration names and skip one? IME duplicate names are are a result of creating a new migration by copy/pasting code, and the new migration ends up silently not being run.

jurasan:
@ssonne I agree that throwing error is better. But now when I have duplicate migration class names and I'm applying it separately, that's what's happening (skipping second migration). 
I'd like the behavior at least be consistent.
If we're ok with throwing error I can do it this way, but it could potentially break older duplicate migrations, if they happened to be applied together in one step and thus both were applied correctly. Are we OK with that?

ssonne:
> it could potentially break older duplicate migrations
That's fair, it would be safer to make throwing an error on duplicates an opt-in flag.

> Are we OK with that?
I'm just an interested user; I don't have write access. But I think this PR is a strict improvement since it makes the behavior consistent.

jurasan:
@pleerock 
Would it make sense to have this change to have consistency now and then another PR which would throw exception but with the next major version, as this is kind of breaking change, sorf of.

pleerock:
we do not consider fixing a broken scenario as a breaking change, its a fix.

pleerock:
and yes, I think we should throw exception in this case

jurasan:
@pleerock Changed to throwing error

PR #4712, opened 2019-09-09
chore: add reference documentation generation

Evertras:
Adds a gulp command that will generate full reference documentation
using TypeDoc.  Current documentation is guide-based and does not
contain a full reference to all classes/interfaces/decorators/etc.

This is half a PR, half a proposal.  However, I've found the result to be immediately useful when working with a project that's using typeorm rather than trying to learn it from scratch, so I'd love to see either this or something like it be accepted.

The documentation location, distribution, etc. still need to be finalized.  I don't know how this would work with current hosting, I've only been using it locally.

There is a markdown plugin for TypeDoc, but using it with typeorm resulted in an essentially unreadable file; the HTML version is far nicer to actually use.  So I don't think this can be simply dropped into the current doc site.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,3 +9,4 @@ ormconfig.json
 ormlogs.log
 npm-debug.log
 /test/github-issues/799/tmp/*
+docs/ref
--- a/gulpfile.ts
+++ b/gulpfile.ts
@@ -16,6 +16,7 @@ const sourcemaps = require("gulp-sourcemaps");
 const istanbul = require("gulp-istanbul");
 const remapIstanbul = require("remap-istanbul/lib/gulpRemapIstanbul");
 const ts = require("gulp-typescript");
+const typedoc = require("gulp-typedoc");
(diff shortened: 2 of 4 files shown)


pleerock:
Last time I used typedoc with typeorm result was not satisfiable, not sure if something has changed. Can you finish the implementation and publish it somewhere? Or give me instructions on a publishing process? (we are currently hosting only on a github so we'll need to push results into that repo I guess)

Evertras:
I could get it published somewhere over the next few days, but if you want to see what it looks like for yourself it should be pretty simple.  If you checkout this branch, then run the docs gulp task, you'll get a full static site in docs/ref.  Opening the index.html file should be sufficient.

AlexMesser:
I think we can add TypeDoc support, but probably we need to finish current implementation and add links into https://github.com/typeorm/typeorm.github.io repository. Closing for now, but feel free to re-open.

Issue #4719, opened 2019-09-10
TypeORM + Postgres: HStore implementation broken & subject to "hstore injection"


tobyhinloopen:
**Issue type:**

[ ] question
[x] bug report
[ ] feature request
[ ] documentation issue

**Database system/driver:**

[ ] `cordova`
[ ] `mongodb`
[ ] `mssql`
[ ] `mysql` / `mariadb`
[ ] `oracle`
[x] `postgres`
[ ] `cockroachdb`
[ ] `sqlite`
[ ] `sqljs`
[ ] `react-native`
[ ] `expo`

**TypeORM version:**

[x] `latest` (0.2.18)
[ ] `@next`
[x] `0.2.18` (or put your version here)

**Steps to reproduce or a small repository showing the problem:**

For an Entity column like this:

```typescript
class Thing {
  // ...
  @Column({ type: "hstore", hstoreType: "object" })
  public values: Record<string, string>;
}
```

The `values` property is not restored properly when saving & restoring the
entity: Empty string values in the map (EG `{ foo: "1", bar: ""}` cause the object to break.

For clarity & reproduction, I've created an isolated bare-minimum repository that acts as an example of the issue.

See https://github.com/tobyhinloopen/typeorm-failure for a runnable example.

tobyhinloopen:
```
    const thing = new Thing();
    thing.values = { username: `", admin=>"1`, admin: "0" };
    await thingRepo.save(thing);
    const foundThing = await thingRepo.findOneOrFail();
    console.log(foundThing.values); // { admin: '1' }
```

🤔

PR #4720, opened 2019-09-10
Fix "Hstore injection" & properly handle NULL, empty string, backslashes & quotes in hstore key/value pairs

tobyhinloopen:
Closes https://github.com/typeorm/typeorm/issues/4719

Add support for empty string keys/values, NULL values, and properly escape key/values.


pleerock:
Thanks!

PR #4722, opened 2019-09-10
Update many-to-one-one-to-many-relations.md

DavidTheProgrammer:
Explicitly defining the optional nature of the ManyToOne and OneToMany relationship when viewed from the child entity.

### DIFF
--- a/docs/many-to-one-one-to-many-relations.md
+++ b/docs/many-to-one-one-to-many-relations.md
@@ -45,7 +45,7 @@ export class User {
 Here we added `@OneToMany` to the `photos` property and specified the target relation type to be `Photo`.
 You can omit `@JoinColumn` in a `@ManyToOne` / `@OneToMany` relation.
 `@OneToMany` cannot exist without `@ManyToOne`.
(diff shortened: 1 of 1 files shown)


docs/many-to-one-one-to-many-relations.md
```diff
@@ -45,7 +45,7 @@ export class User {
 Here we added `@OneToMany` to the `photos` property and specified the target relation type to be `Photo`.
 You can omit `@JoinColumn` in a `@ManyToOne` / `@OneToMany` relation.
 `@OneToMany` cannot exist without `@ManyToOne`.
-If you want to use `@OneToMany`, `@ManyToOne` is required.
+If you want to use `@OneToMany`, `@ManyToOne` is required. However, the inverse is not required, if you only care about the `@ManyToOne` relationship, you can define it without having `@OneToMany` on the parent entity.
```
tobyhinloopen:
```suggestion
If you want to use `@OneToMany`, `@ManyToOne` is required. However, the inverse is not required: If you only care about the `@ManyToOne` relationship, you can define it without having `@OneToMany` on the related entity.
```

Issue #4123, opened 2019-05-11
Views generating broken Migrations


jeanfortheweb:
**Issue type:**

[ ] question
[x] bug report
[ ] feature request
[ ] documentation issue

**Database system/driver:**

[ ] `cordova`
[ ] `mongodb`
[ ] `mssql`
[x] `mysql` / `mariadb`
[ ] `oracle`
[ ] `postgres`
[ ] `cockroachdb`
[ ] `sqlite`
[ ] `sqljs`
[ ] `react-native`
[ ] `expo`

**TypeORM version:**

[x] `latest`
[ ] `@next`
[ ] `0.x.x` (or put your version here)

**Steps to reproduce or a small repository showing the problem:**

When working with views, in migrations a line like the following is generated:

```ts
  await queryRunner.query("INSERT INTO `development`.`typeorm_metadata`(`type`, `schema`, `name`, `value`) VALUES (?, ?, ?, ?)");
```

As you can see, the placeholders are never filled, neither does a table named "typeorm_metadata" exist.

When I delete those lines, everything is working.

<!--
    To answer those questions you need to put "x" inside the square brackets, for example:
    [x] `mysql`
    [ ] `postgres`

    Also, please format your code properly (by taking code blocks into ```ts .... ```)
!>

jeanfortheweb:
I just noticed that `schema:sync` would at least create the `typeorm_metadata` table, which migrations do not

jakubste:
Just noticed the same thing in postgres. After bumping version of typeorm to 0.2.17 and adding a view, migrations fail on 
`INSERT INTO "typeorm_metadata"("type", "schema", "name", "value") VALUES ($1, $2, $3, $4)`
because of `error: relation "typeorm_metadata" does not exist`

From migrations logs it also looks like parameters are not set properly:
```
  query: 'INSERT INTO "typeorm_metadata"("type", "schema", "name", "value") VALUES ($1, $2, $3, $4)',
  parameters: []
```

RobRoseKnows:
Does anyone know if the typeorm_metadata migrations are safe to delete or if that will cause problems elsewhere?

jakubste:
It'll generate problems if you'll try to create new migrations.

Basically, it looks like typeorm checks if views exist in the database looking into this table. If it will not exist it will generate migrations for creating views every time you'll create new migrations.

I think there are 2 solutions:
1) Create and fill `typeorm_metadata` by hand.
or
[...]

PR #4726, opened 2019-09-11
fix: view table migrations

evgeniytar:
MigrationGenerateCommand didn't apply parameters to query and information about ViewTables couldn't be inserted into typeorm_metadata table. Also added code that creates typeorm_metadata table if ViewTables exists

Fixed issue #4123


pleerock:
Thank you!

Issue #4753, opened 2019-09-16
@CavidM @pleerock This seems to cause an error for connection configs with replication. Please note the lack of the `database` property on the root object.


CavidM:
@CavidM @pleerock This seems to cause an error for connection configs with replication. Please note the lack of the `database` property on the root object.

Example config:
```js
{
  type: "mysql",
  replication: {
    master: {
      host: "DB_HOST",
      port: "DB_PORT",
      username: "DB_USER",
      password: "DB_PASSWORD",
      database: "DB_NAME"
    },
    slaves: [
      {
        host: "RO_DB_HOST",
        port: "DB_PORT",
        username: "DB_USER",
        password: "DB_PASSWORD",
        database: "DB_NAME"
      }
    ]
  },
  entities: [...]
};
```

_Originally posted by @cajames in https://github.com/typeorm/typeorm/pull/4694#issuecomment-531664364_

j-tegen:
Having the same issues at the moment. Would you recommend fixing the typeorm version or adding the database property to the root object until this is fixed?

CavidM:
Yes, for now you can use v0.2.18.

elthrasher:
My team was impacted by this as well. Please give https://github.com/typeorm/typeorm/pull/4826 a look.

PR #4754, opened 2019-09-16
Issue #4753

CavidM:
Add condition for check exist database name in driver


pleerock:
looks like duplicate of #4756

PR #4756, opened 2019-09-16
Fixed issue #4753

CavidM:
Add condition for check exist database name in driver.


pleerock:
I think this was already fixed by #4826

PR #4760, opened 2019-09-17
Improve replacePropertyNames performance

mohd-akram:
I did some profiling and found that this function was a bottleneck when using a cache. With these changes I was able to get about 5x better performance.

### DIFF
--- a/src/query-builder/QueryBuilder.ts
+++ b/src/query-builder/QueryBuilder.ts
@@ -565,27 +565,45 @@ export abstract class QueryBuilder<Entity> {
      * Replaces all entity's propertyName to name in the given statement.
      */
     protected replacePropertyNames(statement: string) {
-        this.expressionMap.aliases.forEach(alias => {
-            if (!alias.hasMetadata) return;
-            const replaceAliasNamePrefix = this.expressionMap.aliasNamePrefixingEnabled ? alias.name + "\\." : "";
-            const replacementAliasNamePrefix = this.expressionMap.aliasNamePrefixingEnabled ? this.escape(alias.name) + "." : "";
-            alias.metadata.columns.forEach(column => {
-                const expression = "([ =\(]|^.{0})" + replaceAliasNamePrefix + column.propertyPath + "([ =\)\,]|.{0}$)";
-                statement = statement.replace(new RegExp(expression, "gm"), "$1" + replacementAliasNamePrefix + this.escape(column.databaseName) + "$2");
-                const expression2 = "([ =\(]|^.{0})" + replaceAliasNamePrefix + column.propertyName + "([ =\)\,]|.{0}$)";
-                statement = statement.replace(new RegExp(expression2, "gm"), "$1" + replacementAliasNamePrefix + this.escape(column.databaseName) + "$2");
-            });
-            alias.metadata.relations.forEach(relation => {
-                [...relation.joinColumns, ...relation.inverseJoinColumns].forEach(joinColumn => {
-                    const expression = "([ =\(]|^.{0})" + replaceAliasNamePrefix + relation.propertyPath + "\\." + joinColumn.referencedColumn!.propertyPath + "([ =\)\,]|.{0}$)";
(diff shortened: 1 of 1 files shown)


mohd-akram:
@pleerock Could this be merged?

pleerock:
@Kononnable @vlapo can you guys help to review this one?

mohd-akram:
This seems more useful now given #5309.

thynson:
I also run into these code when profiling. @mohd-akram's work looks good.

henrinormak:
Any update on this? Anything blocking this improvement?

joras:
This function is most prominent one in the profiling results of our services, and I guess it applies for other DB heavy code. Any speedup to this would be with significant effect.

imnotjames:
Are there tests that verify this?  There's none here, but I'm guessing there's tests that verify the functionality otherwise.

src/query-builder/QueryBuilder.ts
```diff
@@ -546,27 +546,45 @@ export abstract class QueryBuilder<Entity> {
...
+            const replacementKeys = Object.keys(replacements)
+                .map(k => k.replace(/\./g, "\\."));
+            if (replacementKeys.length) {
+                statement = statement.replace(new RegExp(
+                    `([ =\(]|^.{0})${replaceAliasNamePrefix}(${
+                    replacementKeys.join("|")})([ =\)\,]|.{0}$)`,
+                    "gm"
+                ), (_, pre, p, post) =>
+                    `${pre}${replacementAliasNamePrefix}${this.escape(replacements[p].databaseName)}${post}`
+                );
```
imnotjames, line 584: This can use positive lookbehinds and lookaheads prevent having to pass pre & post.

```suggestion
                statement = statement.replace(new RegExp(
                    `(?<=[ =\(]|^)${replaceAliasNamePrefix}(${replacementKeys.join("|")})(?=[ =\)\,]|.{0}$)`,
                    "gm"
                ), (_, p) =>
                    `${replacementAliasNamePrefix}${this.escape(replacements[p].databaseName)}`
                );
```

imnotjames:
Rebased against master.  Gonna go looking to make sure we've got tests.

imnotjames:
Oracle fails with this - doesn't in `master` and it's not the changes I added.  Looking..

imnotjames:
Oh.. oh wow.

This passes before & fails now because NOW we're correctly escaping regular expressions.... but before we weren't.  So a `.` was getting included in the pattern inadvertently & allowing issue 215's test to pass.

imnotjames:
Bonus win - this fixes some undesired and undefined behavior!  Wonderful.

imnotjames:
Thanks for the contribution!

ldiqual:
@imnotjames Any chance you could push this in a new release? This should be a huge improvement and I couldn't find a way to install the master branch in my project directly (unless there's a way to do that?)

pleerock:
a new release will be available this week

Vluf:
This breaks QueryBuilder from `0.2.29` on React Native due to lack of positive lookbehinds support.

mohd-akram:
The lookbehind could be removed by adjusting the replacement function (see the first commit for inspiration).

chriswep:
@imnotjames @pleerock this broke typeorm on iOS webkit (which doesn't support lookbehind, not even in iOS 14).

unfortunately all of the last typeorm releases where instantly failing on my app project (via Ionic). Even if a node-environment is the most common use of typeorm, I'd think that hybrid apps are still a significant use case. i think there needs to be some way to catch at least those basic issues before merging PRs.

sandro-salles:
@mohd-akram @pleerock @imnotjames is it possible to revert or add a new version of this change without the lookbehind so it can work on React native again? I'd love to help... but the regex is a little bit too complex for my knowledge... :/

pleerock:
Reverted by #7269 . Fix will be available in `0.2.31`.

djflex68:
But what about the initial fix about replacePropertyNames performance ? Isn't it merged in anymore ?

PR #4769, opened 2019-09-18
better entity values comparison on save

silentroach:
```typescript
@Entity()
export class Post {
    @PrimaryGeneratedColumn()
    id: number;
    @Column({ name: "dt", type: "timestamp with time zone" })  // <-- type is important here
    date: Date;
}

const post = new Post();
post.id = 1;
post.date = new Date();
await postRepository.save(post);       // <-- first save, it is ok

const loadedPost = await postRepository.findOneOrFail(1);
await postRepository.save(loadedPost); // <-- was not changed, 
                                       // but typeorm will update the record with the same date
```

Entity property with column type "timestamp with time zone" value is always different for TypeORM on save.

Seems like you forget to normalize these values on value comparison function.

### DIFF
--- a/src/persistence/SubjectChangedColumnsComputer.ts
+++ b/src/persistence/SubjectChangedColumnsComputer.ts
@@ -73,28 +73,50 @@ export class SubjectChangedColumnsComputer {
                 let normalizedValue = entityValue;
                 // normalize special values to make proper comparision
                 if (entityValue !== null) {
-                    if (column.type === "date") {
-                        normalizedValue = DateUtils.mixedDateToDateString(entityValue);
+                    switch (column.type) {
+                        case "date":
+                            normalizedValue = DateUtils.mixedDateToDateString(entityValue);
+                            break;
 
-                    } else if (column.type === "time") {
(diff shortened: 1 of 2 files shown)


src/persistence/SubjectChangedColumnsComputer.ts
```diff
@@ -73,28 +73,50 @@ export class SubjectChangedColumnsComputer {
...
+
+                        case "simple-array":
+                            normalizedValue = DateUtils.simpleArrayToString(entityValue);
+                            databaseValue = DateUtils.simpleArrayToString(databaseValue);
+                            break;
+                        case "simple-enum":
+                            normalizedValue = DateUtils.simpleEnumToString(entityValue);
+                            databaseValue = DateUtils.simpleEnumToString(databaseValue);
+                            break;
+                        case "simple-json":
```
silentroach: this normalization was missed too

silentroach:
week without any reaction, very sad :{

silentroach:
two weeks without any reaction, what the fuck?

pleerock:
Thanks for waiting @silentroach. I'll merge this change since Im releasing a new version today, however this functionality needs a tests, you know it =)

PR #4770, opened 2019-09-18
Fix many-to-one-one-to-many-relations.md example

ohjeyong:
In table description of photo and user, FOREIGN KEY note is omitted.


pleerock:
thanks

Issue #4241, opened 2019-06-05
Argument of type 'ObjectId' is not assignable to parameter of type 'string | number | Date | ObjectID | FindOneOptions


davidquintard:
**Issue type:**

[ ] question
[ x] bug report
[ ] feature request
[ ] documentation issue

**Database system/driver:**

[ ] `cordova`
[ x] `mongodb`
[ ] `mssql`
[ ] `mysql` / `mariadb`
[ ] `oracle`
[ ] `postgres`
[ ] `cockroachdb`
[ ] `sqlite`
[ ] `sqljs`
[ ] `react-native`
[ ] `expo`

**TypeORM version:**

[x ] `latest`
[ ] `@next`
[ ] `0.x.x` (or put your version here)

Impossible to execute findOne(<ObjectID>)

````
import {ObjectID} from "mongodb";
....
async changeGalerie(id_photo:ObjectID, acces: number) {
    let photo = await this.candidatPhotoRepository.findOne(id_photo);
}
````
error TS2345: Argument of type 'ObjectId' is not assignable to parameter of type 'string | number | Date | ObjectID | FindOneOptions<CandidatPhoto> | Partial<CandidatPhoto>'.
  Property 'generate' is missing in type 'ObjectId' but required in type 'ObjectID'.

**Comment by pairmix** [comment]
CandidatPhotoRepository:
````
import {Service} from "typedi";
import { MongoRepository, EntityRepository } from "typeorm";
import {ObjectID} from "mongodb";
import {CandidatPhoto} from "../entity/CandidatPhoto";

// create custom Repository class
@Service()
@EntityRepository(CandidatPhoto)
export class CandidatPhotoRepository extends MongoRepository<CandidatPhoto>
{
    ....
}
````

**Comment by pairmix** [comment]
It seems i have to replace:
````
let photo = await this.candidatPhotoRepository.findOne(id_photo);
````
By 
````
let photo = await this.candidatPhotoRepository.findOne({id: id_photo});
````

to let it work.

**Comment by vlapo** [comment]
You have to import `ObjectID` from `typeorm`.
```ts
import { ObjectID } from "typeorm";
```

pairmix:
I dit it.
TypeORM is waiting for a typeorm "ObjectID" type. (import { ObjectID } from "typeorm";)
But my entities are defined with a mongodb "ObjectID" type (import {ObjectID} from "mongodb";)
which is deprecated and is an alias of "ObjectId".

@types/bson/index.t.ts
[...]

PR #4777, opened 2019-09-19
change mongodb ObjectID to ObjectId

ruscon:
issue #4241

bson lib types
https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/bson/index.d.ts#L352

A little *BC* 
from `ObjectID`
```typescript
generate(time?: number): string;
```
to `ObjectId`
```typescript
static generate(time?: number): Buffer;
```

Old `ObjectID` will still be supported, but deprecated
https://github.com/typeorm/typeorm/commit/9c8fa67c02b163fd737eec8299a0cf3887553680#diff-859ccae412ed029cdf257edd7258b283R1934-R1939

### DIFF
--- a/docs/decorator-reference.md
+++ b/docs/decorator-reference.md
@@ -260,9 +260,9 @@ Learn more about [entity columns](entities.md#entity-columns).
 
 #### `@ObjectIdColumn`
 
-Marks a property in your entity as ObjectID.
+Marks a property in your entity as ObjectId.
 This decorator is only used in MongoDB.
-Every entity in MongoDB must have a ObjectID column.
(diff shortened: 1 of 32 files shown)


ruscon:
@pleerock, @pairmix, @vlapo ^

ruscon:
@vlapo ping

pleerock:
It's a breaking change. We can't accept this change at the moment. Maybe in the future.

imnotjames:
I don't see tests verifying that the old object ID is supported.

Is there any way to accept the new one as input but still only return the old one?  To preserve BC

imnotjames:

PR #4779, opened 2019-09-19
fix(aurora): apply mysql query fixes to aurora

ArsenyYankovsky:
There's a lot of `instanceof MysqlDriver` checks in the code and we need these for aurora driver as well for queries like updates to work correctly.


TreeMan360:
@pleerock I know you only recently pushed out a release but I would love to see this get merged in along with a new release soon as I am using this in production and would rather not have to monkey-patch my node modules.

PR #4783, opened 2019-09-19
Fix "current_timestamp(6)" !== "CURRENT_TIMESTAMP(6)" for MariaDB >= 10.2

### LINKED ISSUE #4782: CreateDate and UpdateDate columns always get recreated when generating migrations for latest version of MariaDB
**Issue type:**

[ ] question
[x] bug report
[ ] feature request
[ ] documentation issue

**Database system/driver:**

[ ] `cordova`
[ ] `mongodb`
[ ] `mssql`
[x] `mysql` / `mariadb`
[ ] `oracle`
[ ] `postgres`
[ ] `cockroachdb`
[ ] `sqlite`
[ ] `sqljs`
[ ] `react-native`
[ ] `expo`

**TypeORM version:**

[x] `latest`
[ ] `@next`

# The Issue

I've described this issue [here](https://github.com/typeorm/typeorm/issues/2737#issuecomment-532705221). After taking a deeper look into this problem, I've realized that the cause of issue #2737 and this issue is not the same.

MariaDB 10.2.7 introduced the following [change](https://mariadb.com/kb/en/library/changes-improvements-in-mariadb-102/#information-schema) in `information_schema.COLUMNS` table:

> Literals are now quoted in the `COLUMN_DEFAULT` column to distinguish them from expressions

Since `10.2` (the earliest version I could find on Docker Hub was `10.2.5`) the expressions are returned in lowercase. This means we now get `current_timestamp(6)` instead of `CURRENT_TIMESTAMP(6)` for create/update date columns like it is in MariaDB `10.1` and MySQL.

Since the `MysqlDriver#mappedDataTypes.createDateDefault` and `updateDateDefault` are defined as `CURRENT_TIMESTAMP(6)`, the `MysqlDriver#findChangeColumns` function marks these date columns as changed, and new migrations for these columns will be generated indefinitely.

# Default Values Quoted Twice

While trying to fix this issue, I noticed that the literal default values are double quoted in MariaDB 10.2, since they will already have quotes in the `information_schema.COLUMNS` table, and quotes will be added again on this line:

https://github.com/typeorm/typeorm/blob/13ac22242ad1e2a46cc3aa6af870f4b1cfae1af3/src/driver/mysql/MysqlQueryRunner.ts#L1294

---

# More Info

> `COLUMN_DEFAULT`: Default value for the column. From MariaDB 10.2.7, literals are quoted to distinguish them from expressions. NULL means that the column has no default. In MariaDB 10.2.6 and earlier, no quotes were used for any type of default and NULL can either mean that there is no default, or that the default column value is NULL.

- https://mariadb.com/kb/en/library/information-schema-columns-table/

---

# Examples:

### MariaDB 10.1.41

```sql
Welcome to the MariaDB monitor.  Commands end with ; or \g.
Server version: 10.1.41-MariaDB-1~bionic mariadb.org binary distribution

MariaDB [test]> CREATE TABLE test(stamp datetime(6) NOT NULL DEFAULT CURRENT_TIMESTAMP(6));
Query OK, 0 rows affected (0.02 sec)

MariaDB [test]> SELECT table_name, column_name, column_default FROM information_schema.COLUMNS WHERE table_schema = DATABASE();
+------------+-------------+----------------------+
| table_name | column_name | column_default       |
+------------+-------------+----------------------+
| test       | stamp       | CURRENT_TIMESTAMP(6) |
+------------+-------------+----------------------+
1 row in set (0.00 sec)
```

### MariaDB 10.2

```sql
Welcome to the MariaDB monitor.  Commands end with ; or \g.
Server version: 10.2.5-MariaDB-10.2.5+maria~jessie mariadb.org binary distribution

MariaDB [test]> CREATE TABLE test(stamp datetime(6) NOT NULL DEFAULT CURRENT_TIMESTAMP(6));
Query OK, 0 rows affected (0.01 sec)

MariaDB [test]> SELECT table_name, column_name, column_default FROM information_schema.COLUMNS WHERE table_schema = DATABASE();
+------------+-------------+----------------------+
| table_name | column_name | column_default       |
+------------+-------------+----------------------+
| test       | stamp       | current_timestamp(6) |
+------------+-------------+----------------------+
1 row in set (0.00 sec)
```

### MySQL 8.0.17

```sql
Welcome to the MySQL monitor.  Commands end with ; or \g.
Server version: 8.0.17 MySQL Community Server - GPL

mysql> CREATE TABLE test(stamp datetime(6) NOT NULL DEFAULT CURRENT_TIMESTAMP(6));
Query OK, 0 rows affected (0.03 sec)

mysql> SELECT table_name, column_name, column_default FROM information_schema.COLUMNS WHERE table_schema = DATABASE();
+------------+-------------+----------------------+
| TABLE_NAME | COLUMN_NAME | COLUMN_DEFAULT       |
+------------+-------------+----------------------+
| test       | stamp       | CURRENT_TIMESTAMP(6) |
+------------+-------------+----------------------+
1 row in set (0.01 sec)
```

jeremija:
Fixes #4782. I've explained the issue in great detail there.

- Fixed a check for `CURRENT_TIMESTAMP`, and made it more generic to cover cases like `CURRENT_TIMESTAMP`, `CURRENT_TIMESTAMP()` and `CURRENT_TIMESTAMP(6)`, ignoring the case.
- Fixed a test failure in `test/functional/query-builder/add-column.ts` due to double-quoting:
https://github.com/typeorm/typeorm/blob/13ac22242ad1e2a46cc3aa6af870f4b1cfae1af3/test/functional/query-runner/add-column.ts#L73
- Added `semver` dependency to make it easy to check the version of MariaDB.

### DIFF
--- a/src/util/VersionUtils.ts
+++ b/src/util/VersionUtils.ts
@@ -1,0 +1,20 @@
+export type Version = [number, number, number];
+
+export class VersionUtils {
+    static isGreaterOrEqual(version: string, targetVersion: string): boolean {
+        const v1 = parseVersion(version);
+        const v2 = parseVersion(targetVersion);
+
+        return v1[0] > v2[0] ||
+            v1[0] === v2[0] && v1[1] > v2[1] ||
+            v1[0] === v2[0] && v1[1] === v2[1] && v1[2] >= v2[2];
+    }
+}
+
+function parseVersion(version: string = ""): Version {
+    const v: Version = [0, 0, 0];
+
+    version.split(".").forEach((value, i) => v[i] = parseInt(value, 10));
+
+    return v;
+}
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -27,7 +27,7 @@ services:
 
   # mariadb
   mariadb:
-    image: "mariadb:10.1.37"
+    image: "mariadb:10.4.8"
     container_name: "typeorm-mariadb"
     ports:
       - "3307:3306"
--- a/src/driver/mysql/MysqlQueryRunner.ts
+++ b/src/driver/mysql/MysqlQueryRunner.ts
@@ -21,6 +21,7 @@ import {ColumnType, PromiseUtils} from "../../index";
 import {TableCheck} from "../../schema-builder/table/TableCheck";
 import {IsolationLevel} from "../types/IsolationLevel";
 import {TableExclusion} from "../../schema-builder/table/TableExclusion";
+import { VersionUtils } from "../../util/VersionUtils";
 
 /**
  * Runs queries on a single mysql database connection.
@@ -1243,6 +1244,7 @@ export class MysqlQueryRunner extends BaseQueryRunner implements QueryRunner {
             return [];
 
         const isMariaDb = this.driver.options.type === "mariadb";
+        const dbVersion = await this.getVersion();
 
         // create tables for loaded tables
         return Promise.all(dbTables.map(async dbTable => {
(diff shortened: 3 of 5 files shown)


src/util/VersionUtils.ts
```diff
@@ -0,0 +1,20 @@
+export type Version = [number, number, number];
+
+export class VersionUtils {
+    static isGreaterOrEqual(version: string, targetVersion: string): boolean {
+        const v1 = parseVersion(version);
+        const v2 = parseVersion(targetVersion);
+
+        return v1[0] > v2[0] ||
```
pleerock: adding "(" ")" would really simplify understanding of this code

pleerock:
Thank you!

PR #4794, opened 2019-09-22
feat: support spatial functions of MySQL 8+

### LINKED ISSUE #3702: MySQL Spatial Type Support : GeomFromText function is not supported
**Issue type:**

[x] question
[x] bug report
[ ] feature request
[ ] documentation issue

**Database system/driver:**

[ ] `cordova`
[ ] `mongodb`
[] `mssql`
[x] `mysql` / `mariadb`
[ ] `oracle`
[ ] `postgres`
[ ] `sqlite`
[ ] `sqljs`
[ ] `react-native`
[ ] `expo`

**TypeORM version:**

[x] `latest`
[ ] `@next`

As per the documentation, I have the following column in my table

```
@Column("geometry", {
        spatialFeatureType: "Point",
        srid: 4326
})
position: string;
```

It successfully creates the column of type "geometry" in MySQL Table but when you are inserting some data like

```
const location = new Location();
location.name = 'my-location';
location.tags = [ 'tag-1', 'tag-2' ];
location.position = 'POINT(1 1)';

const _location = await connection.manager.save( location );
```

It gives the following error

```
(node:40807) UnhandledPromiseRejectionWarning: QueryFailedError: ER_SP_DOES_NOT_EXIST: FUNCTION mydb.GeomFromText does not exist
```

This is because MySQL 8.0 and higher doesn't recognize `GeomFromText` function (deprecated), it supports these spatial functions with `ST_` prefix like `ST_GeomFromText`. Please read my article on Medium on the spatial data type in MySQL (https://itnext.io/playing-with-geometry-spatial-data-type-in-mysql-645b83880331).

> After fixing this issue my modifying typeorm code, I also found that SRID specification while mentioning column type is not working. typeorm is inserting data with SRID 0 (default MySQL).

Is there anything I am doing wrong or **typeorm** doesn't support spatial data types in MySQL.
 

**Proposed solution**
- Add new connection option `legacySpatialSupport` which defaults to `true`. If `false`, newer spatial functions will be used like `ST_GeomFromText` and `ST_AsText` etc.
- Like PostgreSQL, make use of `srid` while creating spatial column in simple format `@Column( { type: 'point', srid: 4326 } ) position: string;`

jeiea:
* Add `legacySpatialSupport` option for replacing
  dropped functions to new ones

* Use srid column option like PostgreSQL

Implemented the OP's suggestion naively.
Close #3702

### DIFF
--- a/docs/connection-options.md
+++ b/docs/connection-options.md
@@ -154,6 +154,8 @@ Slight performance penalty for most calls. (Default: `true`)
 * `multipleStatements` - Allow multiple mysql statements per query. Be careful with this, it could increase the scope 
 of SQL injection attacks. (Default: `false`)
 
+* `legacySpatialSupport` - Use spatial functions like GeomFromText and AsText which are removed in MySQL 8. (Default: true)
+
 * `flags` - List of connection flags to use other than the default ones. It is also possible to blacklist default ones.
  For more information, check [Connection Flags](https://github.com/mysqljs/mysql#connection-flags).
  
--- a/docs/zh_CN/connection-options.md
+++ b/docs/zh_CN/connection-options.md
@@ -98,6 +98,8 @@
 
 - `multipleStatements` - 每个查询允许多个 mysql 语句。请注意，它可能会增加 SQL 注入攻击的范围。 （默认值：`false`）
 
+- `legacySpatialSupport` - Use spatial functions like GeomFromText and AsText which are removed in MySQL 8. (Default: true)
+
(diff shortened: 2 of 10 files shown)


pleerock:
Can you please fix conflicts? thanks :)

jeiea:
Fixed. Thanks for the reply.

jeiea:
Please let me know if any consideration or elaboration is needed.

pleerock:
Sorry for waiting and thanks for your great efforts!

PR #4800, opened 2019-09-23
test: add test which demonstrates incorrect JOIN clause for issue 1668

sgronblo:
Note: no fix here yet, just a new test case that demonstrates breakage when getting a relation using find when using postgres. Not tested on other DBs.

It gets a bit hairy to see what's going on as I tried to boil away anything that was related to the original problem domain. However you can make the query work by changing the following in the broken query:

`ON C__a.aId="C"."aId"` -> `ON "C__a"."id"="C"."aId"`

So, two problems. Double quotes are missing, and the id column is incorrect.

-- Original commit message follows

The problem demonstrated here is not exactly the same as in the original
issue for 1668. However the symptoms are similar. The original creator
of 1668 was experiencing a lack of double quotes around the table/column
name for a join condition. That matches this case. However there is an
additional problem where the column used to join to the related table is
also incorrect.

When running the tests, and viewing the database logs you can see the
following error:

2019-09-23 08:45:08.233 UTC [6013] ERROR:  missing FROM-clause entry for table "c__a" at character 178
2019-09-23 08:45:08.233 UTC [6013] STATEMENT:  SELECT "C"."aId" AS "C_aId", "C"."fooCode" AS "C_fooCode", "C"."barId" AS "C_barId", "C__a"."id" AS "C__a_id", "C__a"."name" AS "C__a_name" FROM "c" "C" LEFT JOIN "a" "C__a" ON C__a.aId="C"."aId" WHERE "C"."barId" = $1


imnotjames:
Rebased & cleaned up the test a bit.

hinogi:
Is this still an issue?

aphofstede:
This is still an issue in 0.3.17. Any ideas for a workaround?

yevhen-komarov:
If you need to fix this issue or want to test the solution, install [#11400](https://github.com/typeorm/typeorm/pull/11400#issuecomment-2849034549)

Issue #3427, opened 2019-01-16
[serverless] Connection Reuse is broken in a Lambda environment


Wintereise:
**Issue type:**

[ ] question
[x] bug report
[ ] feature request
[ ] documentation issue

**Database system/driver:**

[ ] `cordova`
[ ] `mongodb`
[ ] `mssql`
[ ] `mysql` / `mariadb`
[ ] `oracle`
[x] `postgres`
[ ] `sqlite`
[ ] `sqljs`
[ ] `react-native`
[ ] `expo`

**TypeORM version:**

[x] `latest`
[ ] `@next`
[ ] `0.x.x` (or put your version here)

**Steps to reproduce or a small repository showing the problem:**

I'm primarily expanding further on https://github.com/typeorm/typeorm/issues/2598#issue-345445322, the fixes described there are catastrophic in nature, because:

1. They either involve closing connections without understanding that the lambda may not just be serving that one request.
2. Full bootstrap of the connection a second time anyway (defeating the purpose of caching it in the first place)) and I believe this is something that should be addressed in core.
3. Even the somewhat saner metadata rebuild causes problems (since it cannot be awaited and runs in parallel). This results in metadata lookups while queries are running (for another controller, perhaps) randomly start failing.

The issue is exactly as described, if we attempt to reuse connections in a lambda environment, the entity manager no longer seems to know anything about our entities.

The first request made completes successfully (and primes our connection manager to reuse the same connection). Subsequent requests quit with `RepositoryNotFoundError: No repository for "TData" was found. Looks like this entity is not registered in current "default" connection?`

Here's a complete test case (with no external dependencies other than typeorm and uuid): https://gist.github.com/Wintereise/3d59a0414419b4ecca5137c20fc29622

When the else block (line #22 onwards) is hit, all is well, things work fine. When the cached connection from `manager.get("default")` is hit however, we can no longer run queries.

If it's needed, I can setup a project with `serverless-offline` for click-and-go testing.

Wintereise:
The problem effectively originates here: https://github.com/typeorm/typeorm/blob/ad04ec9207483cc12dc5fb9ed1c3fa3ecae9d8f1/src/entity-manager/EntityManager.ts#L730

For all subsequent attempts using a cached connection, `Connection#hasMetaData` inexplicably seems to return false.

[...]

PR #4804, opened 2019-09-25
Connection Reuse is broken in a Lambda environment:

phong-innomizetech:
- Fix issue find metadata does not exist with webpack

### DIFF
--- a/src/connection/Connection.ts
+++ b/src/connection/Connection.ts
@@ -480,7 +480,8 @@ export class Connection {
      */
     protected findMetadata(target: Function|EntitySchema<any>|string): EntityMetadata|undefined {
         return this.entityMetadatas.find(metadata => {
-            if (metadata.target === target)
+            // @ts-ignore
+            if (metadata.target.name === target.name) {
                 return true;
             if (target instanceof EntitySchema) {
                 return metadata.name === target.options.name;


src/connection/Connection.ts
```diff
@@ -480,7 +480,8 @@ export class Connection {
      */
     protected findMetadata(target: Function|EntitySchema<any>|string): EntityMetadata|undefined {
         return this.entityMetadatas.find(metadata => {
-            if (metadata.target === target)
+            // @ts-ignore
```
silentroach: no

src/connection/Connection.ts
```diff
@@ -480,7 +480,8 @@ export class Connection {
...
+            if (metadata.target.name === target.name) {
```
silentroach: `target` can be a `string`

silentroach: missing closing brackets

what is wrong with you?

craigcartmell:
Pushed a couple of fixes to @phong-innomizetech 's original commit.

craigcartmell:
Original monkeypatch is here which I used for reference: https://github.com/typeorm/typeorm/issues/3427#issuecomment-504134702

Benjamin-Dobell:
@pleerock This PR seems to be responsible for a _lot_ of open issues at the moment.

It doesn't fix Webpack support, but instead breaks it. This is trivially verifiable by looking at the line:

https://github.com/typeorm/typeorm/pull/4804/files#diff-9ada25e7bb500fdf98eeb214f7c94e9aR483

What this does is compare _runtime_ names of functions, that's almost certainly going to fail in production, as production code is typically minified. Minified names are by no means guaranteed to be unique, so they shouldn't ever be used in a lookup.

In my case, I've got two classes that were both minified to `e`, so when I tried `getRepository(SomeModel)` it returns the wrong repository.

The [previous functionality](https://github.com/typeorm/typeorm/pull/4804/files#diff-9ada25e7bb500fdf98eeb214f7c94e9aL483) on the other hand was correct, in that it compared functions themselves, not their runtime names. This still occurs, but only if a match isn't found on the runtime name (which won't occur).

Benjamin-Dobell:
Actually, even in non-minified code, there's no guarantee class (function) names are unique. They probably shouldn't be involved in the lookup at all.

Benjamin-Dobell:
@pleerock My apologies, I see this has been discussed in #4967.

Issue #4820, opened 2019-09-26
EntityMetadata ignores Schemas for OracleDriver


jf214:
**Issue type:**

[ ] question
[X] bug report
[ ] feature request
[ ] documentation issue

**Database system/driver:**

[ ] `cordova`
[ ] `mongodb`
[ ] `mssql`
[ ] `mysql` / `mariadb`
[X] `oracle`
[ ] `postgres`
[ ] `cockroachdb`
[ ] `sqlite`
[ ] `sqljs`
[ ] `react-native`
[ ] `expo`

**TypeORM version:**

[X] `latest`
[ ] `@next`
[ ] `0.x.x` (or put your version here)

**Steps to reproduce or a small repository showing the problem:**

I'm using the OracleDriver and after upgrading to the 0.2.19, the generated queries ignore the schemas defined on my Entity Definition.

Example Entity
```ts
import {
  Column,
  Entity,
  PrimaryColumn,
} from "typeorm";

@Entity({
  name: "PERSON",
  schema: "DW_ADMIN"
})
export class Person {
    @PrimaryColumn({ name: "ID" })
    public id: number;

    @Column({ name: "NAME" })
    public name: string;
}
```

The resulting query:
```sql
SELECT "p"."ID" AS "p_ID", "p"."NAME" AS "p_NAME" FROM "PERSON"
```

The result is an oracle error because there is no schema attached to the table in the query
```ORA-00942: table or view does not exist``` 

Prior to upgrading the query was:
```sql
SELECT "p"."ID" AS "p_ID", "p"."NAME" AS "p_NAME" FROM "DW_ADMIN"."PERSON"
```

<!--
    To answer those questions you need to put "x" inside the square brackets, for example:
    [x] `mysql`
    [ ] `postgres`

    Also, please format your code properly (by taking code blocks into ```ts .... ```)
!>

marrasen:
We have a similar problem using MySQL. I think this is a problem for all databases except Postgres and SqlServer since this commit:

https://github.com/typeorm/typeorm/commit/d8f1c81dcfed33bf73493fccd162dff70f2679f9

kdi-1:
The same problem. Any workaround?

marrasen:
> The same problem. Any workaround?

We reverted to version 0.2.18 until this is resolved

jf214:
We've done the same.

marrasen:
[...]

PR #4821, opened 2019-09-26
Proposed Fix for #4820 

jf214:
Including schema in table path for OracleDriver in EntityMetadata


pleerock:
as I know in oracle there is no real schema. There is a schema but is equal to user. If so then our schema related functionality might not work properly.

jf214:
So far, in my experience using Typeorm with Oracle, I have not had any issues with how schemas are treated by the library. It wasn't until I updated to latest that my queries ignored the schema I had defined in the entities, which caused my queries to fail in Oracle. I think it started with d8f1c81dcfed33bf73493fccd162dff70f2679f9.

I tested my fix and it seems to work for Oracle. Though in hindsight, it isn't ideal because it doesn't address the issue for other driver types.

kdi-1:
@jf214  I think in number 8 should be 
import {OracleDriver} from "../driver/oracle/OracleDriver";

imnotjames (changes requested):
This needs tests to confirm the behaviors

kdi-1:
@imnotjames @jf214  I think this can be closed as fixed in #6673

Issue #3841, opened 2019-03-18
Unique index gets droppped and generated each time I use "typeorm migration:generate"


svegalopez:
**Issue type:**

[x ] question
[ ] bug report
[ ] feature request
[ ] documentation issue

**Database system/driver:**

[ ] `cordova`
[ ] `mongodb`
[x ] `mssql`
[ ] `mysql` / `mariadb`
[ ] `oracle`
[ ] `postgres`
[ ] `cockroachdb`
[ ] `sqlite`
[ ] `sqljs`
[ ] `react-native`
[ ] `expo`

**TypeORM version:**

[x ] `latest`
[ ] `@next`
[ ] `0.x.x` (or put your version here)

**Steps to reproduce or a small repository showing the problem:**

1. Add an index to an entity using decorator :
```javascript
@Index(["countryCode", "locode"], { unique:true, where: 'locode IS NOT NULL' })
```
2. use the cli to generate the migration: 
``` 
typeorm migration:generate
```

It comes up with this:
```javascript
await queryRunner.query(`CREATE UNIQUE INDEX "IDX_a4465d440921efae77ef84b626" ON "core"."cities" ("country_code", "locode") WHERE locode IS NOT NULL`);
```

3. Run the migration
```
typeorm migration:run
```

4.  Keep working on your project, add other entities etc.. and then generate a new migration with your latest work:
```
typeorm migration:generate
```

5. You will see this everytime you generate a new migration using the cli:
```javascript
public async up(queryRunner: QueryRunner): Promise<any> {
        await queryRunner.query(`DROP INDEX "IDX_a4465d440921efae77ef84b626" ON "core"."cities"`);
        await queryRunner.query(`ALTER TABLE "core"."products" ADD "name_test" varchar(100) NOT NULL`);
        await queryRunner.query(`CREATE UNIQUE INDEX "IDX_a4465d440921efae77ef84b626" ON "core"."cities" ("country_code", "locode") WHERE locode IS NOT NULL`);
    }
```

As you can see, the UNIQUE INDEX that was added in steps 2 and 3, is getting dropped and generated again and again...

smonv:
Any update on this issue?

Hovik123:
You need to add index on entity and add  { synchronize: false } option for make sure that you disabled the sync for the column.

@Index('customer-id-idx', ['customerId'], { fulltext: true })
@Index('customer-name-idx', ['name'], { fulltext: true })
[...]

Issue #2943, opened 2018-10-17
Inappropriate migration generated


btxtiger:
**Issue type:**

[ ] question
[ x ] bug report
[ ] feature request
[ ] documentation issue

**Database system/driver:**

[ ] `cordova`
[ ] `mongodb`
[ ] `mssql`
[ x ] `mysql` / `mariadb`
[ ] `oracle`
[ ] `postgres`
[ ] `sqlite`
[ ] `sqljs`
[ ] `react-native`
[ ] `expo`

**TypeORM version:**

[ x ] `latest` 0.2.7
[ ] `@next`
[ ] `0.x.x` (or put your version here)

**Steps to reproduce or a small repository showing the problem:**

When using an `unsigned` **PrimaryGenerateColumn**, and generating a migration using `typeorm migration:generate`, the generated migrations get into a loop. The generated migration is always the same, because the query leads to an undocumented change in the database, which causes migration:generate to detect the change again and again.

Reproduction:
```ts
import { Entity, PrimaryGeneratedColumn, Column, PrimaryColumn, Generated } from 'typeorm';

/**
 * Log action
 */
@Entity('log_action')
export class LogAction {
   /**
    * logActionId
    */
   @PrimaryGeneratedColumn({
      name: 'log_action_id',
      unsigned: true
   })
   public logActionId: number;

   /**
    * title
    */
   @Column('varchar', {
      nullable: false,
      length: 150,
      name: 'title'
   })
   public title: string;
   /**
    * created
    */
   @Column('datetime', {
      nullable: false,
      default: () => 'CURRENT_TIMESTAMP',
      name: 'created'
   })
   public created: Date;

   /**
    * last_modified
    */
   @Column('datetime', {
      nullable: false,
      default: () => 'CURRENT_TIMESTAMP',
      onUpdate: 'CURRENT_TIMESTAMP',
      name: 'last_modified'
   })
   public lastModified: Date;
}
```

- **Create table**
`ts-node ./node_modules/.bin/typeorm schema:sync`
- **Create migration**
`ts-node ./node_modules/.bin/typeorm migration:generate -n test`

You can run `migration:generate` and `migration:run` multiple times, there will always be this migration generated, again and again.
```ts
import {MigrationInterface, QueryRunner} from "typeorm";

export class test1539781487192 implements MigrationInterface {

    public async up(queryRunner: QueryRunner): Promise<any> {
[...]

PR #4819, opened 2019-09-26
Fix #3841 - unique index in `migration:generate`

fw623:
fixed #3841 (at least for mysql)
before: drop and readd unique index in every `migration:generate`

PR #4822, opened 2019-09-27
fix: resolve MySQL unique index check when bigNumberStrings is false

SteveByerly:
Part of fix for #2737 and #2943

### DIFF
--- a/src/driver/aurora-data-api/AuroraDataApiQueryRunner.ts
+++ b/src/driver/aurora-data-api/AuroraDataApiQueryRunner.ts
@@ -1217,8 +1217,18 @@ export class AuroraDataApiQueryRunner extends BaseQueryRunner implements QueryRu
                 .map(dbColumn => {
 
                     const columnUniqueIndex = dbIndices.find(dbIndex => {
-                        return this.driver.buildTableName(dbIndex["TABLE_NAME"], undefined, dbIndex["TABLE_SCHEMA"]) === tableFullName
-                            && dbIndex["COLUMN_NAME"] === dbColumn["COLUMN_NAME"] && dbIndex["NON_UNIQUE"] === "0";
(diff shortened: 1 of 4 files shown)


fw623:
I found the same problem with unique Index getting dropped and readded every migration. 
My attempt at fixing: [PR4819](https://github.com/typeorm/typeorm/pull/4819)
This should also fix Issue #3841

src/driver/aurora-data-api/AuroraDataApiQueryRunner.ts
```diff
@@ -1340,7 +1344,7 @@ export class AuroraDataApiQueryRunner extends BaseQueryRunner implements QueryRu
                     table: table,
                     name: constraint["INDEX_NAME"],
                     columnNames: indices.map(i => i["COLUMN_NAME"]),
-                    isUnique: constraint["NON_UNIQUE"] === "0",
+                    isUnique: constraint["NON_UNIQUE"] == "0", // tslint:disable-line:triple-equals - trying to find truthy value
```
pleerock: imagine somebody reading this code. Can you please explicitly specify what logic should be in there (as I understood we should check for === "0" && === 0? If so, please change it

pleerock:
@fw623 correct its a duplicate, but here we also have a test and both drivers covered. We'll need to close your PR once this one is completed

src/driver/aurora-data-api/AuroraDataApiQueryRunner.ts
SteveByerly: @pleerock updated the code to be more explicit.

Pulled in latest from master and rebased.

I debated pulling that logic out into a method on the QueryRunner, but didn't want to overstep and make assumptions.
something like:
`private isUniqueIndex(constraint): boolean`

If you'd rather go that route, I can update. Otherwise let me know if there's anything else you'd like changed.

pleerock:
I think it looks okay now, thanks

PR #4826, opened 2019-09-28
fixes 4753

elthrasher:
This should fix https://github.com/typeorm/typeorm/issues/4753 and also provide a better way to actually get the database name.

### DIFF
--- a/src/connection/Connection.ts
+++ b/src/connection/Connection.ts
@@ -517,10 +517,26 @@ export class Connection {
         const migrations = connectionMetadataBuilder.buildMigrations(this.options.migrations || []);
         ObjectUtils.assign(this, { migrations: migrations });
 
-        this.driver.database = <string> this.options.database;
+        this.driver.database = this.getDatabaseName();
 
         // validate all created entity metadatas to make sure user created entities are valid and correct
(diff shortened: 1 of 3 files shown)


src/connection/Connection.ts
```diff
@@ -517,10 +517,26 @@ export class Connection {
...
-        this.driver.database = <string> this.options.database;
+        this.driver.database = this.getDatabaseName();
 
         // validate all created entity metadatas to make sure user created entities are valid and correct
         entityMetadataValidator.validateMany(this.entityMetadatas.filter(metadata => metadata.tableType !== "view"), this.driver);
     }
 
+    // This database name property is nested for replication configs.
+    protected getDatabaseName(): string {
+    const options = this.options;
```
ruscon:
```typescript
const {replication, database} = this.options;
// ...
```

elthrasher: This causes an issue with the typeguard. I can't be sure `this.options` contains `replication` until after the guard. I was trying to avoid the use of `any`. I like destructuring, but don't think it's worth the cost here.

pleerock:
Thanks @elthrasher so much!

PR #4834, opened 2019-09-30
docs: fix entities broken link

blackpr:
The `decorator-reference` link is broken. I believe this will fix it
see: https://typeorm.io/#/entities


pleerock:
Thanks!

PR #4844, opened 2019-10-02
test(postgres): add test case for zero-arity cubes

liaujianjie:
Add test case for a bug where Postgres `cube` of zero arity incorrectly hydrates as `[0]` instead of `[]`.

### DIFF
--- a/test/functional/cube/postgres/cube-postgres.ts
+++ b/test/functional/cube/postgres/cube-postgres.ts
@@ -100,6 +100,17 @@ describe("cube-postgres", () => {
     it("should persist cube of arity 0 correctly", () =>
         Promise.all(
             connections.map(async connection => {
+                // Get Postgres version because zero-length cubes are not legal
+                // on all Postgres versions. Zero-length cubes are only tested
+                // to be working on Postgres version >=10.6.
+                const [{ server_version }] = await connection.query(
+                    "SHOW server_version"
+                );
+                const semverArray = server_version.split(".").map(Number);
+                if (!(semverArray[0] >= 10 && semverArray[1] >= 6)) {
+                    return;
+                }
+
                 const color: number[] = [];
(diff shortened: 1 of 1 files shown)


liaujianjie:
The broken test has to do with the Postgres version. Test fails on `9.6.11` but works on `10.6`. Not sure how to proceed from here.

pleerock:
@liaujianjie looks like we already have this change in #4848 ? If so, can we just left a test from this PR since it looks like is missing in #4848

liaujianjie:
@pleerock The test are still failing because CircleCI uses Postgres v9.6.11 though.

```sql
CREATE EXTENSION cube;
CREATE TABLE test (color CUBE);
INSERT INTO test(color) VALUES ('()'); -- works for Postgres v10.6 but not v9.6
```

Error throw in Postgres v9.6:

```
ERROR:  bad cube representation
LINE 1: INSERT INTO test(color) VALUES ('()');
                                        ^
DETAIL:  syntax error at or near ")"
```

pleerock:
maybe we should update a postgres version in config to make these tests pass?

liaujianjie:
I'm all for it, but that means we're not going to guarantee support for Postgres versions <10. Are we ok with that?

pleerock:
then maybe we can add some check to this test - test if its a postgres >= 10 then execute it, otherwise ignore?

liaujianjie:
Sounds good. How can I check the Postgres version?

swim:
Adding a getVersion to the QueryRunner (SHOW server_version) would be nice however, it would need to be implemented for each driver.

liaujianjie:
> Adding a getVersion to the QueryRunner (SHOW server_version) would be nice however, it would need to be implemented for each driver.

I think he was referring to just the codebase.

PR #4854, opened 2019-10-03
entity property name helper for propertyPath on increment/decrement

silentroach:
`propertyPath` argument in `increment`/`decrement` methods of `manager` as a simple string is not really safe way to pass the field name and it is so abstract just to support embedded fields.

Here is a helper for current entity property names, just to help developer pass valid name in simple cases.

Also it will be great to have a default value in this methods.

Screen Shot 2019-10-03 at 14 04 10

### DIFF
--- a/src/repository/Repository.ts
+++ b/src/repository/Repository.ts
@@ -339,14 +339,14 @@ export class Repository<Entity extends ObjectLiteral> {
     /**
      * Increments some column by provided value of the entities matched given conditions.
      */
-    increment(conditions: FindConditions<Entity>, propertyPath: Extract<keyof Entity, string> | string, value: number | string): Promise<UpdateResult> {
+    increment(conditions: FindConditions<Entity>, propertyPath: Extract<keyof Entity, string> | string, value: number | string = 1): Promise<UpdateResult> {
         return this.manager.increment(this.metadata.target, conditions, propertyPath, value);
     }
 
     /**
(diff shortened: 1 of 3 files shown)


src/repository/Repository.ts
```diff
@@ -339,14 +339,14 @@ export class Repository<Entity extends ObjectLiteral> {
     /**
      * Increments some column by provided value of the entities matched given conditions.
      */
-    increment(conditions: FindConditions<Entity>, propertyPath: string, value: number | string): Promise<UpdateResult> {
+    increment(conditions: FindConditions<Entity>, propertyPath: Extract<keyof Entity, string> | string, value: number | string = 1): Promise<UpdateResult> {
```
pleerock: its opinionated if we should make `1` default or not. Need some additional opinions on this small change.

pleerock:
Usually I reject such PRs because people don't know about embedded functionality and make it simply keyof Entity and remove string. Such PRs make great sense without embedded functionality in. But in your case you also specified a string and such changes make it only half sense. You won't have type safety, the only thing you have is type hints in your IDE.

silentroach:
🤷‍♂️

pleerock:
Not sure if you understood my point, but it wasn't in rejecting this PR. I just wanted to say that we can merge this PR but we also can improve it with my future planned strategy.

PR #4860, opened 2019-10-04
docs: don't await .merge in example with express

mguida22:
In the [examples with express](https://typeorm.io/#/example-with-express/adding-typeorm-to-the-application) the PUT request awaits when it doesn't need to, since .merge doesn't return a promise.


pleerock:
Thanks!

PR #4861, opened 2019-10-04
Updated CircleCI config to run tests in parallel

csuich2:
Per discussion in #4495, I figured I'd give CircleCI a try. The config is pretty verbose right now and I'm continuing to look in to how to simply it, but this config fans out to the 3 node versions we test in TravisCI, then fans out to all of the DBs in parallel.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -10,6 +10,8 @@ executors:
         name: mariadb
         environment:
           MYSQL_ROOT_PASSWORD: "admin"
+          MYSQL_USER: "test"
+          MYSQL_PASSWORD: "test"
           MYSQL_DATABASE: "test"
   mongodb:
     docker:
@@ -28,6 +30,8 @@ executors:
       - image: mysql:5.7.24
         environment:
           MYSQL_ROOT_PASSWORD: "admin"
+          MYSQL_USER: "test"
+          MYSQL_PASSWORD: "test"
           MYSQL_DATABASE: "test"
   oracle:
     docker:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,13 +1,64 @@
 # Javascript Node CircleCI 2.0 configuration file
 # Check https://circleci.com/docs/2.0/language-javascript/ for more details
-version: 2
+version: 2.1
+
+commands:
+  run-tests:
+    steps:
+      - attach_workspace:
+          at: ~/typeorm/
+      - run: sudo npm install -g npm@latest
+      - run: cp ormconfig.circleci.json ormconfig.json
+
+      # Download and cache dependencies
+      - restore_cache:
+          keys:
+            - ${NODE_VERSION}-v1-dependencies-{{ checksum "package.json" }}
+            # fallback to using the latest cache if no exact match is found
+            - ${NODE_VERSION}-v1-dependencies-
+
+      - run: npm install
+
+      - save_cache:
+          paths:
+            - node_modules
+          key: ${NODE_VERSION}-v1-dependencies-{{ checksum "package.json" }}
+
+      # - run: docker-compose up -d
+
+      # run tests
+      - run: npm run lint
+      - run: npm test
+
 jobs:
-  build:
-    working_directory: ~/typeorm
+  checkout_code:
     docker:
-      # specify pre-built docker images
-      # documented at https://circleci.com/docs/2.0/circleci-images/
       - image: circleci/node:10.15.0
+    working_directory: ~/typeorm/checkout/
+    steps:
+      - checkout
+      - run:
+          name: Copy source for node versions
+          command: |
+            mkdir ~/typeorm/node8/
+            mkdir ~/typeorm/node10/
+            mkdir ~/typeorm/node12/
+
+            cp -R ~/typeorm/checkout/. ~/typeorm/node8/
+            cp -R ~/typeorm/checkout/. ~/typeorm/node10/
+            cp -R ~/typeorm/checkout/. ~/typeorm/node12/
+      - persist_to_workspace:
+          root: ~/typeorm/
+          paths:
+            - ./checkout/
+            - ./node8/
+            - ./node10/
+            - ./node12
+
+  test-node8: &test-template
+    working_directory: ~/typeorm/node8/
+    docker:
+      - image: circleci/node:8.16.1
       - image: mysql:5.7.24
         environment:
           MYSQL_ROOT_PASSWORD: "admin"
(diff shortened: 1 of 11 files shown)


csuich2:
[CircleCI](https://circleci.com/gh/csuich2/typeorm)

csuich2:
This is what the workflow UI looks like:
image

csuich2:
I'm still struggling with one last MySQL failure and I'm not sure why. The config is based off the docker-compose.yml, so the same permissions should be the same as when tests are run locally.

image

pleerock:
Can you please explain all these changes? I see changes, but I also need to understand motivation behind each change.

.circleci/config.yml
```diff
@@ -1,56 +1,378 @@
 # Javascript Node CircleCI 2.0 configuration file
 # Check https://circleci.com/docs/2.0/language-javascript/ for more details
-version: 2
-jobs:
-  build:
-    working_directory: ~/typeorm
+version: 2.1
+
+executors:
```
csuich2: Executors are ways of sharing the configuration that a job will need to run. In this case, I've defined an executor for each DB that we test, that way when we run the tests on each version of Node, each of those jobs can simply reference the executor for that DB rather than duplicating it.

csuich2: Note: All of these configs were copied from `docker-compose.yml`

.circleci/config.yml
```diff
@@ -1,56 +1,378 @@
...
-      # - image: cockroachdb/cockroach-unstable:latest
-      #   command: start --insecure
+  sqlite:
+    docker:
+      - image: circleci/node:10.15.0 # sqlite doesn't require another image
+  sqljs:
+    docker:
+      - image: circleci/node:10.15.0 # sqljs doesn't require another image
 
+commands:
```
csuich2: Commands are job `steps` that can have parameters so they can be shared between jobs. In this case, I've created one command for running `npm install` and copying the ormconfig (which I now realize is unnecessary since we copy the DB specific ormconfigs for each test). This step also leverage's CircleCI's `restore_cache` and `save_cache` in order to cache the `npm install` to speed up builds.

.circleci/config.yml
```diff
@@ -1,56 +1,378 @@
...
             - node_modules
-          key: v1-dependencies-{{ checksum "package.json" }}
-
-      # - run: docker-compose up -d
+          key: << parameters.node_folder >>-v1-dependencies-{{ checksum "package.json" }}
 
       # run tests
       - run: npm run lint
+
+  run-db-tests:
```
csuich2: This command is used by all of the jobs that are the permutations of Node version & DB. Much like the TravisCI PR I PoC'd, this simply copies the appropriate ormconfig, then runs `npm test` which tests just a single DB.

.circleci/config.yml
```diff
@@ -1,56 +1,378 @@
...
+          - sqlite
+          - sqljs
+    steps:
+      - attach_workspace:
+          at: ~/typeorm/
+      - run: cp ormconfigs/ormconfig.<< parameters.database >>.json ormconfig.json
       - run: npm test
+
+jobs:
+  checkout_code:
```
csuich2: This is the first step that runs, which checks out the code, then copies it to separate folders for each node version. that way we can run `npm install` and `npm test` in the same node version-specific folder.

.circleci/config.yml
```diff
@@ -1,56 +1,378 @@
...
+            cp -R ~/typeorm/checkout/. ~/typeorm/node12/
+      - persist_to_workspace:
+          root: ~/typeorm/
+          paths:
+            - ./checkout/
+            - ./node8/
+            - ./node10/
+            - ./node12
+
+  test-node8: &test-template
```
csuich2: This step's name is a misnomer that I need to fix. It actually leverages the `prepare-for-tests` command which runs `npm install`. In this case, it is for Node 8. The `&test-template` is a YAML feature which allows this YAML object to be referenced later on and copied, so that the Node 10 & 12 respective steps only need to specify the values that are different. Right now, that actually means everything, since each property has node version specific values, but it could be used to only specify some values once.

.circleci/config.yml
```diff
@@ -1,56 +1,378 @@
...
+      - image: circleci/node:8.16.1
+    steps:
+      - prepare-for-tests:
+          node_folder: node8
+      - persist_to_workspace:
+          root: ~/typeorm/
+          paths:
+            - ./node8/
+
+  test-node8-mariadb: &mariadb-template
```
csuich2: This step actually runs tests. In this case, it runs the MariaDB tests on Node 8. Once again, we see `&mariadb-template`, which allows the Node 10 & 12 MariaDB tests to only specify values which are different. Unlike the last step, this one does actually reduce duplicate configuration since the only property with node specific values is `working_directory`.

Each of the test running steps will specify:
* which executor to use (so that the appropriate DB container us used)
* the working directory to use (which depends on the node version)
* the steps to run (which is almost always just `run-db-tests) with which DB is being tested, so we can copy the ormconfig for that DB.

.circleci/config.yml
```diff
@@ -1,56 +1,378 @@
...
+
+  test-node12-sqlite:
+    <<: *sqlite-template
+    working_directory: ~/typeorm/node12/
+
+  test-node12-sqljs:
+    <<: *sqljs-template
+    working_directory: ~/typeorm/node12/
+
+workflows:
```
csuich2: I've started leveraging CircleCI workflows in this config which allow the tests to run in parallel. This is easiest to visualize in the CircleCI dashboard, but essentially:

* `checkout_code` runs to get the code from GH, the copy that code for each version of node to test
* `test_node8`, `test_node10` and `test_node12` run which runs `npm install` for their respective versions of node
* `test_node*_*` runs for each version of node and each DB to test

I'm definitely not happy with how verbose this config is, but as far as I can tell, CircleCI workflows don't allow for parameters or any other way of specifying a matrix of jobs to run for each value. While the config is definitely overly verbose, it doesn't seem too bad since we don't add new DBs that often.

csuich2:
Definitely! I've added some comments which should help and I'm happy to keep explaining where I can. The basic idea is to leverage CircleCI workflows which allow for jobs to run in parallel. We start out by checking out the code, the fanning the build out to test every permutation of node version and DB we support/test.

The config is way more verbose than I'd like right now, but CircleCI doesn't allow for parameters to be used in every place where it would be helpful. I've considered sending this config to someone at CircleCI and seeing if they have any suggestions of how to better implement what we want to do, but this should get us started.

Kononnable:
>The basic idea is to leverage CircleCI workflows which allow for jobs to run in parallel.

This describes what you want to use, but doesn't say anything about what you want to achieve. Is it to make CI tests run faster, ability to see if error is db specific or something completely different?

Kononnable:
General idea is great, not sure if everything I mention here is possible to be achieved. Some of the points aren't really a problem of this PR.
Problems I see with this PR:
- very verbose - it's not so simple anymore to change tested node versions(and every half a year/a year we have to change it)
- some tests fails without enabled specific db drivers - it was fixed in master, but in next some new tests requires specific driver to be enabled to pass
- long job names - to see which db failed you have to hover over the job
- code compiles multiple times(every `npm test`)
- sometimes tests fails for various reason not related to code changes and we have to restart it manually - with more jobs this situation might be more common
- does it take much longer than current approach? we test more targets etc. but there also might be some parallelism restrictions

csuich2:
Good points:
>     * very verbose - it's not so simple anymore to change tested node versions(and every half a year/a year we have to change it)
This is something I do want to continue looking in to, but I don't have a solution yet. It might be helpful to reach out to CIrcleCI support and see if they have suggestions.
>     * some tests fails without enabled specific db drivers - it was fixed in master, but in next some new tests requires specific driver to be enabled to pass
There were two tests that were failing if sqlite wasn't set up in the ormconfig, but I've fixed that as part of this (on master).
>     * long job names - to see which db failed you have to hover over the job
This was cleanup I wanted to do as well. I've pushed a couple commits that should improve the job names.
>     * code compiles multiple times(every `npm test`)
Good find. I thought we were using ts-node for tests. I've pulled the compile step out of the test jobs and into the prepare/setup job.
>     * sometimes tests fails for various reason not related to code changes and we have to restart it manually - with more jobs this situation might be more common
Anecdotally, I haven't experienced that very much. Once I got things working in CircleCI, I didn't see any transient failures.
>     * does it take much longer than current approach? we test more targets etc. but there also might be some parallelism restrictions
I'm currently seeing builds take ~20 minutes compared to TravisCI's ~30 minutes. By default, CircleCI allows open source projects to have 4 containers at a time for running jobs. They of course offer additional containers at an additional fee, but I'm already seeing lower build times with just 4 containers. We could also potentially reach out and ask about additional free containers.

csuich2:
image

Kononnable:
On travis there is also CockroachDb being tested - if I recall correctly it was the slowest one.
As for build time - we could alternatively test just some of those settings - e.g. one container for node 8, node 10 and separate container for each db on node 12. We would know if bug is db driver specific(it will fail on db specific container with node 12), but if it only happens on specific combination of node and db we would get general container failing(without knowledge of db).

We also have to look on build times using two different point of view:
- pushing changes to one PR
- merging multiple PRs
In second option we would queue multiple builds, so this time might sum up. However we can use different settings for CI builds on PRs and on master/next branch(if we would have to lower the build times).

imnotjames:
I'm going to close this -

* The CircleCI config has been changed pretty drastically from this PR
* tripling the builds mean much slower as we only have a handful of available workers with CircleCI

PR #4865, opened 2019-10-06
Prevents the In operator from crashing when called with empty parameters

0xClpz:
The In() operator crashes if the array passed in parameters is empty, this makes it painful to work with dynamic data for the in operator.

This PR fixes that.

### DIFF
--- a/src/find-options/FindOperator.ts
+++ b/src/find-options/FindOperator.ts
@@ -107,7 +107,7 @@ export class FindOperator<T> {
             case "between":
                 return `${aliasPath} BETWEEN ${parameters[0]} AND ${parameters[1]}`;
             case "in":
-                return parameters.length > 0 ? `${aliasPath} IN (${parameters.join(", ")})` : "";
+                return parameters.length > 0 ? `${aliasPath} IN (${parameters.join(", ")})` : "1 != 1";
             case "any":
                 return `${aliasPath} = ANY(${parameters[0]})`;
             case "isNull":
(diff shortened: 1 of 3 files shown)


pleerock:
this way database will return **all** rows and that might be unexpected. not sure what is the best design choose here:

* return nothing if empty array was passed
* throw an error if empty array was passed
* ignore condition and return all without matching - the way you did it

I'm not sure last approach is right - maybe we should return nothing in the case if empty array was passed.

0xClpz:
Ohh nice catch! I think it should indeed return nothing if an empty array is passed! Can you point me in the right direction to implement that? Thanks!

pleerock:
If it should return nothing then we can do something like if its empty then apply IS NULL to this property. But I'm not sure we should do it.

We should think what we would like this to lead to:

```ts
           {
                where: {
                    uuid: In([])
                }
            }
```

* give us all users?
* give us users where uuid is null?

What is the point to do second? Doing `uuid IS NULL` seems odd in this case, isn't it?

I need some opinions here @Kononnable @vlapo

bogdan:
Other ORMs return nothing in this case by adding `where 1 != 1` into the where condition to make sure it is really nothing.

0xClpz:
> Other ORMs return nothing in this case by adding `where 1 != 1` into the where condition to make sure it is really nothing.

Just implemented that in my PR, i guess that's what makes the most sense

PR #4873, opened 2019-10-08
Add optional name to MigrationInterface (fixes #3933)

### LINKED ISSUE #3933: React Native: create tables query fails in release mode with query syntax error
**Issue type:**

[ ] question
[x] bug report
[ ] feature request
[ ] documentation issue

**Database system/driver:**

[ ] `cordova`
[ ] `mongodb`
[ ] `mssql`
[ ] `mysql` / `mariadb`
[ ] `oracle`
[ ] `postgres`
[ ] `cockroachdb`
[ ] `sqlite`
[ ] `sqljs`
[x] `react-native`
[ ] `expo`

**TypeORM version:**

[ ] `latest`
[ ] `@next`

I only added TypeORM to my project last week. Everything was working great while developing but as soon as I began making release builds of my app the entire database stopped working. Turning on logging shows me this:
```
2019-04-03 19:08:05.864 [error][tid:com.facebook.react.JavaScript] 'query failed: ', 'CREATE TABLE "temporary_t" ()'
2019-04-03 19:08:05.866 [error][tid:com.facebook.react.JavaScript] 'error: ', { message: 'near ")": syntax error', code: 5 }
```

I've been unable to identify what could be causing this and from a cursory search I haven't been able to find any other reports of this issue.

I believe the error is being thrown immediately on creation of the database in the root of the app, and no further database related operation work in the application. My initialisation code:
```js
const sqliteConnectionManager = getConnectionManager();
const connection = sqliteConnectionManager.create({
  type: 'react-native',
  database: DATABASE_NAME,
  synchronize: true,
  logging: true,
  // migrations,
  entities,
  location: 'default',
});

await connection.connect();
```

I'm using plain JavaScript in this project, not TypeScript. Any help would be much appreciated.

May also be worth mentioning I'm using ES6 with decorators for defining my entities, like so:
```js
@Entity()
class ReferenceDataGroup {
  @PrimaryColumn('text')
  groupId;

  @Column({ type: 'text', nullable: true })
  lastModified;

  @OneToMany(() => ReferenceDataValue, rdValue => rdValue.groupId, { eager: false })
  values;
}
```

leonardofalk:
This fixes a problem when after minification process, migrations cannot be loaded because their name is changed to some random value, this happens mostly when using react-native/expo.

There are several topics about this issue in facebook/metro and typeorm repositories, but I couldn't find any pull request for it. It could be fixed by changing the default minifier, but that can break other stuff.

There are also a few steps before merging this PR:

- [x] If this field sould be called 'migrationName'? Other options could be 'displayName' or simply 'name'.

- [x] Should migrations generated with the CLI contain this attribute by default?

**EDIT**

Created a babel-plugin to automatically add name property:

```js
module.exports = function({ types: t }, options) {
  return {
    visitor: {
      ClassDeclaration(path) {
        const migrationName = t.stringLiteral(path.node.id.name)
        const migrationTimestamp = parseInt(migrationName.value.substr(-13), 10)

        if (!isNaN(migrationTimestamp)) {
          const prop = t.classProperty(t.identifier(options.property), migrationName, null, null)

          path.node.body.body.push(prop)
        }
      },
    },
  }
}
```

```js
{
  presets: ['module:metro-react-native-babel-preset'],
  plugins: [
    ['babel-plugin-typeorm-migration-name', { property: 'migrationName' }]
  ],
}
```

### DIFF
--- a/src/commands/MigrationGenerateCommand.ts
+++ b/src/commands/MigrationGenerateCommand.ts
@@ -128,9 +128,12 @@ export class MigrationGenerateCommand implements yargs.CommandModule {
      * Gets contents of the migration file.
      */
     protected static getTemplate(name: string, timestamp: number, upSqls: string[], downSqls: string[]): string {
+        const migrationName = `${camelCase(name, true)}${timestamp}`;
+
         return `import {MigrationInterface, QueryRunner} from "typeorm";
 
-export class ${camelCase(name, true)}${timestamp} implements MigrationInterface {
+export class ${migrationName} implements MigrationInterface {
+    name = '${migrationName}'
 
     public async up(queryRunner: QueryRunner): Promise<any> {
 ${upSqls.join(`
--- a/src/migration/MigrationExecutor.ts
+++ b/src/migration/MigrationExecutor.ts
@@ -314,7 +314,7 @@ export class MigrationExecutor {
      */
     protected getMigrations(): Migration[] {
         const migrations = this.connection.migrations.map(migration => {
-            const migrationClassName = (migration.constructor as any).name;
+            const migrationClassName = migration.name || (migration.constructor as any).name;
             const migrationTimestamp = parseInt(migrationClassName.substr(-13));
             if (!migrationTimestamp)
                 throw new Error(`${migrationClassName} migration name is wrong. Migration class name should have a JavaScript timestamp appended.`);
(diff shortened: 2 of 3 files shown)


pleerock:
> If this field sould be called 'migrationName'? Other options could be 'displayName' or simply 'name'. 

simply name looks fine

> Should migrations generated with the CLI contain this attribute by default?

not sure, but we can add it

leonardofalk:
> > If this field sould be called 'migrationName'? Other options could be 'displayName' or simply 'name'.
> 
> simply name looks fine
> 
> > Should migrations generated with the CLI contain this attribute by default?
> 
> not sure, but we can add it

Done, changed to name and added it to the migration template.

PR #4874, opened 2019-10-08
fix: correcting the data returned by queries with join, orderBy and take

separovichabel:
Queries with the 'orderBy' in a joined column will not return less entities than 'take' specifies.

(sorry about my English)
The case where I found this problem is the following:
```typescript
const user1 = new User();
const user2 = new User();
await connection.manager.save([user1, user2]); // creates users id 1 and 2

    
const phone1 = new Phone();
phone1.user = user1;
phone1.number = 1;

const phone2 = new Phone();
phone2.user = user1;
phone2.number = 2;
await connection.manager.save([phone1, phone2]); // register 2 phones to user1

// it does not returns the 2 expected users
// [ User { id: 1, phones: [ [Phone], [Phone] ] } ] instead
const usersWithSelect = await connection.manager.createQueryBuilder(User, 'user')  
        .leftJoinAndSelect('user.phones', 'phone')
        .orderBy('phone.number', 'DESC')
        .take(2)
        .getMany();
```
The reason is the first of the two querys made. The one that take the Ids:

```sql
SELECT DISTINCT `distinctAlias`.`user_id` as "ids_user_id", `distinctAlias`.`phone_number` 
     FROM (
          SELECT `user`.`id` AS `user_id`, `phone`.`id` AS `phone_id`, `phone`.`number` AS `phone_number`, `phone`.`userId` AS `phone_userId` 
          FROM `user` `user` 
          LEFT JOIN `phone` `phone` ON `phone`.`userId`=`user`.`id`
     ) `distinctAlias` 
     ORDER BY `distinctAlias`.`phone_number` DESC, `user_id` ASC
     LIMIT 2
```
Since de `ORDER BY` is in the outer query it is needed to have the `distinctAlias.phone_number`.
But It crashes the `take` propose when there is more than one `distinctAlias.phone_number` with diferent values.
The result of this (in MySql) is User1 twice as response.

My solution was: 

```sql
SELECT
	DISTINCT `distinctAlias`.`user_id` as "ids_user_id"
FROM
	(
	SELECT
		`user`.`id` AS `user_id`,
		`phone`.`id` AS `phone_id`,
		`phone`.`number` AS `phone_number`,
		`phone`.`userId` AS `phone_userId`,
		`phone_number`
	FROM
		`user` `user`
	LEFT JOIN `phone` `phone` ON
		`phone`.`userId` = `user`.`id`
	ORDER BY
		`phone`.`number` DESC) `distinctAlias`
LIMIT 2
```

Passing the `ORDER BY` to the SubQuery, the `distinctAlias.phone_number` is no more necessary.
Now the `Distinct` will effectively returns Id just once. No more data will be included in the first query response.

Thanks!


pleerock:
You need ordering in distinct query otherwise data won't be sorted properly. Tests are failing as you can see.

separovichabel:
> You need ordering in distinct query otherwise data won't be sorted properly. Tests are failing as you can see.

But we need to choose: 
- A - ordering in distinct query breaking the take functionality 
- B - Ordering in the subquery and loose the ordenation

None of these options are correct. How do we correct both problems?

imnotjames:
Thanks for the contribution.  However, there's a few problems with this PR:

* You've included `dist` erroneously
* You've updated the package lock
* There are conflcits elsewhere
* Tests are failing

I'm going to close this out as your upstream repository has been deleted and I believe you've stopped working on it.

Please feel free to open a new PR once you've corrected the issues.

PR #4478, opened 2019-07-24
feat: add materialized View support for Postgres

Ginden:
feat: add option to synchronize or not to synchronize ViewEntity

Fixes #4317
Fixes  #3996

Tests must be added probably, but I have some issues with running them in Docker.

PR #4877, opened 2019-10-09
fix(postgres): postgres query runner to create materialized view

JackCuthbert:
We've just started using a materialized view in our project (postgres DB) and noticed that the schema sync task during our tests was not creating a materialized version of the view when passing `{ materialized: true }` in the entity definition.

Looks like a few things were missed in https://github.com/typeorm/typeorm/pull/4478:

1. materialized SQL logic left out of `PostgresQueryRunner.ts`.
2. materialized condition logic was backwards and _always_ adding `MATERIALIZED` to the create view SQL
3. The `materialized` option in the entity definition was always being set to false due to not being passed through
4. [Materialized views are not supported in cockroachdb](https://marsettler.com/docs/v2.1/postgres-ddl-support.html#create-materialized-view) so I've disabled the cockroach driver for the postgres test (this was causing tests to fail without a meaningful reason). I'm happy to add some boilerplate tests for cockroach but I'm not experienced with it so I'm not certain how much value I could add to that suite.

We're using my fork internally in production and the new materialized views are working great. Would be awesome to have someone take a look at this :smile:

### DIFF
--- a/src/driver/postgres/PostgresQueryRunner.ts
+++ b/src/driver/postgres/PostgresQueryRunner.ts
@@ -1739,10 +1739,13 @@ export class PostgresQueryRunner extends BaseQueryRunner implements QueryRunner 
     }
 
     protected createViewSql(view: View): Query {
+        const materializedClause = view.materialized ? "MATERIALIZED " : "";
+        const viewName = this.escapePath(view);
+
         if (typeof view.expression === "string") {
-            return new Query(`CREATE VIEW ${this.escapePath(view)} AS ${view.expression}`);
+            return new Query(`CREATE ${materializedClause}VIEW ${viewName} AS ${view.expression}`);
         } else {
-            return new Query(`CREATE VIEW ${this.escapePath(view)} AS ${view.expression(this.connection).getQuery()}`);
+            return new Query(`CREATE ${materializedClause}VIEW ${viewName} AS ${view.expression(this.connection).getQuery()}`);
         }
     }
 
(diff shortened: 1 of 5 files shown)


src/driver/postgres/PostgresQueryRunner.ts
```diff
@@ -1739,10 +1739,12 @@ export class PostgresQueryRunner extends BaseQueryRunner implements QueryRunner
     }
 
     protected createViewSql(view: View): Query {
+        const materializedClause = view.materialized ? "MATERIALIZED " : "";
+
         if (typeof view.expression === "string") {
-            return new Query(`CREATE VIEW ${this.escapePath(view)} AS ${view.expression}`);
+            return new Query(`CREATE ${materializedClause}VIEW "${view.name}" AS ${view.expression}`);
```
pleerock: what is the point of removing `this.escapePath` here?

JackCuthbert: Oops, didn't notice that! Will amend, good catch :smile: 

(I'll be honest, it was copy pasta from the previous PR :wink:)

JackCuthbert: Looks like that change causes a syntax error. I'll see if I can get this sorted over the next day or so.

pleerock:
Not sure what you did in the last commit, but looks like its still not using this.escapePath

pleerock:
oh sorry, my inattentiveness. Let's merge it.

PR #4882, opened 2019-10-09
Allow Unique decorator to accept generic types

alk831:
It stricts `Unique` decorator array of column names to only be properties of the provided type.
Below example will print type error (User class doesn't have `idd` property).
```ts
@Entity
@Unique<User>('unique_key', ['idd'])
class User {
    id: number;
}
```

Works as usual when no type is provided (no type error):
```ts
@Entity
@Unique('unique_key', ['idd'])
class User {
    id: number;
}
```

### DIFF
--- a/src/decorator/Unique.ts
+++ b/src/decorator/Unique.ts
@@ -4,28 +4,28 @@ import {UniqueMetadataArgs} from "../metadata-args/UniqueMetadataArgs";
 /**
  * Composite unique constraint must be set on entity classes and must specify entity's fields to be unique.
  */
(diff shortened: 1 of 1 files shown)


pleerock:
I would like to merge it, however it breaks some exist functionality.

We have functionality called "embedded", it's when you have a column inside some class, for example: user.names.first, and you have to specify such columns in this unique decorator. With this change we won't be able to specify it.

alk831:
You are right, and I couldn't find any workaround for this.

PR #4884, opened 2019-10-09
fix: allow EntitySchema to be passed to EntityRepository

da-mkay:
Currently we cannot pass an `EntitySchema<T>` to the `@EntityRepository` decorator unless we cast it to `any`. But actually it works with an `EntitySchema`, just the types are missing. This PR adds the types.


pleerock:
Looks like we can merge it

PR #4892, opened 2019-10-10
feat: export Migration Execution API from main package (fixes #4880)

### LINKED ISSUE #4880: Make MigrationExecutor API more controllable and public
**Issue type:**

[ ] question
[ ] bug report
[x] feature request
[ ] documentation issue

**Database system/driver:**

Doesn't apply to a specific driver.

**TypeORM version:**

[x] `latest`
[ ] `@next`

I'd like to use MigrationExecutor to deal with migrations myself, today all that can be done is calling `runMigrations`, and a few other methods.

In my scenario, if by some reason a duplicated table/column error is thrown I'd like to ignore this migration, add it to done migrations and call the next one, this is impossible without calling MigrationExecutor private API or reimplementing it.

This is a piece of code that show how I'm doing it today:

```ts
import { Connection, QueryRunner } from 'typeorm/browser'
import { MigrationExecutor } from 'typeorm/browser/migration/MigrationExecutor'
import _ from 'lodash'

const IGNORED_ERRORS = [
  /table .+ already exists/i,
  /index .+ already exists/i,
  /duplicate column name/i,
  /no such table: .+/i,
  /no such index: .+/i,
].map(r => new RegExp(r))

export class DatabaseMigrationService {
  executor: any
  queryRunner: QueryRunner

  constructor(connection: Connection) {
    this.executor = new MigrationExecutor(connection)
    this.queryRunner = this.executor.queryRunner || connection.createQueryRunner('master')
  }

  call() {
    const queryRunner = this.queryRunner

    return new Promise(async (resolve, reject) => {
      try {
        const migrations: Migration[] = this.executor.getMigrations()
        const doneMigrations = await this.executor.loadExecutedMigrations(queryRunner)
        const newMigrations: any = _.reject(migrations, migration =>
          _.find(doneMigrations, ['timestamp', migration.timestamp])
        )

        _.each(newMigrations, async migration => await this.executeMigration(queryRunner, migration))

        resolve()
      } catch (e) {
        reject(e)
      } finally {
        if (this.queryRunner) {
          this.queryRunner.release()
        }
      }
    })
  }

  executeMigration(queryRunner: QueryRunner, migration: Migration): Promise<boolean> {
    this.executor.transaction = false

    return new Promise(async (resolve, reject) => {
      try {
        this.registerLog(`Migration ${migration.name} started`)

        await (migration.instance as any).up(queryRunner)
        await this.insertExecutedMigration(migration)

        resolve(true)
      } catch (e) {
        this.registerLog(e.message, { level: 'ERROR' })

        if (this.shouldIgnoreError(e)) {
          this.insertExecutedMigration(migration)
          resolve(true)
        } else {
          reject(e)
        }
      } finally {
        this.registerLog(`Migration ${migration.name} finished`)
      }
    })
  }

  shouldIgnoreError(error: Error): boolean {
    return _.some(IGNORED_ERRORS, r => r.test(error.message))
  }

  registerLog(content: string, props = {}) {
    // register this with external apis ...
  }

  insertExecutedMigration(migration: Migration) {
    return this.executor.insertExecutedMigration(this.queryRunner, migration)
  }
}
```

I'd like to propose this changes:

- Make getMigrations, loadExecutedMigrations, insertExecutedMigration, deleteExecutedMigration, methods public
- Make `import { MigrationExecutor } from 'typeorm'` available
- Maybe add/rename some methods (ie.: showMigrations could be hasMigrationsPending, while showMigrations could be preserved as a way to log migration status), add getPendingMigrations, rename/add loadExecutedMigrations to getDoneMigrations, and maybe other methods

If this is approved I could put a work on it.

leonardofalk:
This may need a lot of rework and review, but it describes what I tried to explain in #4880

### DIFF
--- a/src/migration/MigrationExecutor.ts
+++ b/src/migration/MigrationExecutor.ts
@@ -52,6 +52,78 @@ export class MigrationExecutor {
     // -------------------------------------------------------------------------
 
     /**
+     * Tries to execute a single migration given.
+     */
+    public async executeMigration(migration: Migration): Promise<Migration> {
+        return this.withQueryRunner(async (queryRunner) => {
+            await this.createMigrationsTableIfNotExist(queryRunner);
+            await (migration.instance as any).up(queryRunner);
+            await this.insertExecutedMigration(queryRunner, migration);
+
+            return migration;
+        });
+    }
+
+    /**
+     * Returns an array of all migrations.
+     */
+    public async getAllMigrations(): Promise<Migration[]> {
+        return Promise.resolve(this.getMigrations());
+    }
+
+    /**
+     * Returns an array of all executed migrations.
+     */
+    public async getExecutedMigrations(): Promise<Migration[]> {
+        return this.withQueryRunner(async queryRunner => {
+            await this.createMigrationsTableIfNotExist(queryRunner);
+
+            return await this.loadExecutedMigrations(queryRunner);
+        });
+    }
+
+    /**
+     * Returns an array of all pending migrations.
+     */
+    public async getPendingMigrations(): Promise<Migration[]> {
+        const allMigrations = await this.getAllMigrations();
+        const executedMigrations = await this.getExecutedMigrations();
+
+        return allMigrations.filter(migration =>
+            executedMigrations.find(
+                executedMigration =>
+                    executedMigration.name === migration.name
+            )
+        );
+    }
+
+    /**
+     * Inserts an executed migration.
+     */
+    public insertMigration(migration: Migration): Promise<void> {
+        return new Promise((resolve, reject) => {
+            this.withQueryRunner(queryRunner => {
+                this.insertExecutedMigration(queryRunner, migration)
+                    .then(resolve)
+                    .catch(reject);
+            });
+        });
+    }
+
+    /**
+     * Deletes an executed migration.
+     */
+    public deleteMigration(migration: Migration): Promise<void> {
+        return new Promise((resolve, reject) => {
+            this.withQueryRunner(queryRunner => {
+                this.deleteExecutedMigration(queryRunner, migration)
+                    .then(resolve)
+                    .catch(reject);
+            });
+        });
+    }
+
+    /**
      * Lists all migrations and whether they have been executed or not
      * returns true if there are unapplied migrations
      */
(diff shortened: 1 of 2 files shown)


src/migration/MigrationExecutor.ts
```diff
@@ -48,6 +48,40 @@ export class MigrationExecutor {
...
+        return this.withQueryRunner(async (queryRunner) => {
+            await this.createMigrationsTableIfNotExist(queryRunner);
+            await (migration.instance as any).up(queryRunner);
+            await this.insertExecutedMigration(queryRunner, migration);
+
+            return migration;
+        })
+    }
+
+    public async getAllMigrations(): Promise<Migration[]> {
```
pleerock: why do we need it if we already have getMigrations ?

pleerock:
Think we can merge it.. but tests are failing because of tslint errors, can you please fix them

src/migration/MigrationExecutor.ts
leonardofalk: getMigrations is a protected method where this one is part of the public API proposed returning a Promise<Mutation[]>, like its sibling methods getExecutedMigrations and getPendingMigrations.

leonardofalk:
> Think we can merge it.. but tests are failing because of tslint errors, can you please fix them

Yeah I can fix it, I was hoping to get some feedback first, because there are some other changes in the original issue which needs to be done, if they make any sense.

pleerock:
For me those methods don't make sense since I never had such use case. But we still can make some methods public if users need it. I would like to hear other people opinions around this question but looks like nobody is interested

src/migration/MigrationExecutor.ts
pleerock: not sure about it. Not sure we really need this "all" name. But we can left it as you did.

leonardofalk: I just thought that for consistency they should all return promises instead and I didn't want to introduce any changes to current behaviour. I'll change getMigrations to return a Promise and make it public, if that's ok.

pleerock: choose any approach you prefer

leonardofalk:
> Think we can merge it.. but tests are failing because of tslint errors, can you please fix them

Done.

PR #4893, opened 2019-10-10
Fix CreateDateColumn and UpdateDateColumn defaults

Elassyo:
This fixes an issue where the default for columns using CreateDateColumn and UpdateDateColumn are considered as changed because the database server reports the default for these columns with the CURRENT_TIMESTAMP function in lowercase.

### DIFF
--- a/src/driver/mysql/MysqlDriver.ts
+++ b/src/driver/mysql/MysqlDriver.ts
@@ -232,10 +232,10 @@ export class MysqlDriver implements Driver {
     mappedDataTypes: MappedColumnTypes = {
         createDate: "datetime",
         createDatePrecision: 6,
-        createDateDefault: "CURRENT_TIMESTAMP(6)",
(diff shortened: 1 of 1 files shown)


SteveByerly:
wouldn't it be better to fix the part that does the comparison in case this behavior differs between environments?

pleerock:
@SteveByerly is right, but we need to make sure to lowercase not all defaults, but only for those type of columns

Elassyo:
What other drivers have a lowerDefaultValueIfNecessary method to convert the schema column default values to lowercase before comparing them with the database column metadata.
The MySQL driver doesn't have this so it seems that all default values are compared in a case sensitive way so the problem isn't limited to CreateDateColumn and UpdateDateColumn.

pleerock:
> it seems

Schema sync is dangerous place to change. We need to make sure before changes.

PR #4896, opened 2019-10-11
Add fix for 'pg' turning cockroachdb ints to strings

csuich2:
In an attempt to get `next` building, I found the reason for one of the CockroachDB failures. According to https://github.com/cockroachdb/cockroach/issues/14701, `pg` treats bigint values as strings, because they can be 64-bit, but JS only supports 53-bit numbers.

There is a proposed workaround which I have put into TypeORM and verified it fixes the test. However, additional tests are now failing because the `column-types-cockroach.ts` tests were coded to assume that cockroach numbers are actually strings.

I *think* the fix from that PR is a good start, but I'm not an expert at what is going on here.

### DIFF
--- a/src/driver/cockroachdb/CockroachDriver.ts
+++ b/src/driver/cockroachdb/CockroachDriver.ts
@@ -672,6 +672,23 @@ export class CockroachDriver implements Driver {
     protected loadDependencies(): void {
         try {
             this.postgres = PlatformTools.load("pg");
+            // https://github.com/cockroachdb/cockroach/issues/14701
+            this.postgres.types.setTypeParser(20, (value: string) => {
+                const parsed = parseInt(value);
+
+                // Is this number safe to convert?
+                if (parsed.toString() !== value) {
+                  return value;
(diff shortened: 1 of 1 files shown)


pleerock:
This isn't a good approach and can break both types and give unpredicted errors. For a regular "number" we use ints, if user specify some specific type, for example "integer" we can use strings.

This code helps to understand when we do parseInt:

https://github.com/pleerock/typeorm/blob/a925be97a6e21c58b2941ba5e11be9eeb91ac30c/src/driver/cockroachdb/CockroachDriver.ts#L339

This code helps to understand how we handle int and non int column types

https://github.com/typeorm/typeorm/blob/15a3e876f415856c594eb08458e907395b16ddc8/test/functional/database-schema/column-types/cockroachdb/column-types-cockroach.ts#L30

imnotjames:
@csuich2 Tests fail for this & you didn't add any tests that confirm the behavior that's intended.  Can you do that?

PR #4900, opened 2019-10-14
fix typo

fatihky:
hey it seems `save` must be `remove` but I'm not sure about the rest of the sentence.


pleerock:
you are right, thanks!

PR #4912, opened 2019-10-15
doc: clarify comments on many-to-many sample

mguida22:
While trying to better understand how TypeORM handles many-to-many relations, I came across this sample code with comments. Some of the comments seemed mistakenly copy/pasted, so I tried to clarify them as best as I could.

Thanks for all of the work on TypeORM!


mguida22:
Looks like the build errored? Is there a way I can trigger a rebuild?

pleerock:
Thanks!

Issue #2794, opened 2018-09-13
Cache breaks date field


ph55:
**Issue type:**
[V] bug report

**Database system/driver:**
[V] `mysql` / `mariadb`

**TypeORM version:**
[V] `5.2.0`

**Steps to reproduce:**
When using cache on table that have date field, it returns date with time.

Entity example:
```ts
@Entity()
export class RateMinute {
  @Index('IDX_symbol')
  @PrimaryColumn()
  symbol: string;

  @Column(decimal)
  open: number;

  @Column(decimal)
  close: number;

  @Column(decimal)
  high: number;

  @Column(decimal)
  low: number;

  @Column(decimal)
  volume: number;

  @Index('IDX_date')
  @PrimaryColumn('date')
  date: string;

  @Index('IDX_time')
  @PrimaryColumn('time')
  time: string;
}
```

```sql
CREATE TABLE `rate_minute` (
  `symbol` varchar(255) NOT NULL,
  `open` decimal(20,8) NOT NULL,
  `close` decimal(20,8) NOT NULL,
  `high` decimal(20,8) NOT NULL,
  `low` decimal(20,8) NOT NULL,
  `volume` decimal(20,8) NOT NULL,
  `date` date NOT NULL,
  `time` time NOT NULL,
  PRIMARY KEY (`symbol`,`date`,`time`),
  KEY `IDX_symbol` (`symbol`),
  KEY `IDX_date` (`date`),
  KEY `IDX_time` (`time`)
) ENGINE=InnoDB DEFAULT CHARSET=utf8;
```

No cache result:
```
{
symbol: "USDJPY",
open: "111.57900000",
close: "111.58400000",
high: "111.58400000",
low: "111.57900000",
volume: "4.93080000",
date: "2018-09-13",
time: "11:15:00"
}
```

Cache result (see date):
```
{
symbol: "USDJPY",
open: "111.57900000",
close: "111.58400000",
high: "111.58400000",
low: "111.57900000",
volume: "4.93080000",
date: "2018-09-12T21:00:00.000Z",
time: "11:15:00"
}
```

JClackett:
are you using redis for you cache?

ph55:
No, default one.
https://typeorm.io/#/caching

JClackett:
hmm, i was having the same issue with redis where when storing the payload it would Json.stringify it and then would convert all dates to full datetime format, wonder if the default one has the same issue...

leonidasv:
[...]

PR #4918, opened 2019-10-17
Parse date strings loaded from cache

Slessi:
Closes: #2794


pleerock:
From issue description it looks like we are storing the date wrongly. It says original date is in `2018-09-13` format, but we save in the cache in the `2018-09-12T21:00:00.000Z` format. 

So instead of changing value after loading we should store date in a proper format. To fix it we need to change the way we serialize model before putting into the cache, iterate over every column (entity column metadatas), get column value using `columnMetadata.getEntityValue` and use driver's `preparePersistentValue` method to convert into the format we need to save.

imnotjames (changes requested):
Tests are failing.  Can this be rebased & fix the tests?

Slessi:
@imnotjames I don't really use this anymore, maybe somebody else can pick this up

Issue #4687, opened 2019-09-04
Why `repository.findOne(null)` returns 1st record of the table?


anthonylau:
**Issue type:**

[x] question
[ ] bug report
[ ] feature request
[ ] documentation issue

**Database system/driver:**

[ ] `cordova`
[ ] `mongodb`
[ ] `mssql`
[x] `mysql` / `mariadb`
[ ] `oracle`
[ ] `postgres`
[ ] `cockroachdb`
[ ] `sqlite`
[ ] `sqljs`
[ ] `react-native`
[ ] `expo`

**TypeORM version:**

[ ] `latest`
[ ] `@next`
[x] `0.2.15` (or put your version here)

**Steps to reproduce or a small repository showing the problem:**

```ts
const user = await userRepository.findOne(null); 
// user.id is 1
```

I expect it would returns `undefined`. Is it intended?

DimaIT:
As we have in documentation: `findOne - Finds first entity that matches some id or find options.`. If no find options given then every record matches.

It may be confusing, especially when you are expecting `findOne` to search by id only. In such case it may be better to search by id explicitly: `userRepository.findOne({ id: value })`. Then you 'll get `undefined` for `null` value.

Issue #4448, opened 2019-07-17
findOneOrFail on undefined


chungchi300:
**Issue type:**

[ ] question
[ x] bug report
[ ] feature request
[ ] documentation issue

**Database system/driver:**

[ ] `cordova`
[ ] `mongodb`
[ ] `mssql`
[x ] `mysql` / `mariadb`
[ ] `oracle`
[ ] `postgres`
[ ] `cockroachdb`
[ ] `sqlite`
[ ] `sqljs`
[ ] `react-native`
[ ] `expo`

**TypeORM version:**

[x ] `latest`
[ ] `@next`
[ ] `0.x.x` (or put your version here)

**Steps to reproduce or a small repository showing the problem:**
```javascript
await photoRepository.findOneOrFail(undefined);

//it returns photo element !
```

LuisAngelMoralesRomo:
There is already a discussion about this on issue #2500

vlapo:
Close as duplicate of #2500

Issue #2500, opened 2018-07-11
.findOne(undefined) returns first item in the database instead of undefined


ct-reposit:
**Issue type:**

[ ] question
[x] bug report
[ ] feature request
[ ] documentation issue

**Database system/driver:**

[ ] `cordova`
[ ] `mongodb`
[ ] `mssql`
[ ] `mysql` / `mariadb`
[ ] `oracle`
[x ] `postgres`
[ ] `sqlite`
[ ] `sqljs`
[ ] `react-native`
[ ] `expo`

**TypeORM version:**

[x] `latest`
[ ] `@next`
[ ] `0.x.x` (or put your version here)

**Steps to reproduce or a small repository showing the problem:**

```
user.id = undefined;
User.findOne(user.id); // expect to see undefined / null but returns the first item in the table
```

pleerock:
Looks like an issue. Contributions are welcomed!

kvnvo:
Same here, why there is no fix?

vlapo:
Sure from this point it seem like a bug but we have to consider more than one using of `findOne` e.g.:
 - Active record
```
const user = await User.findOne()
```
 - repository
```
const repository = connection.getRepository(User);
const users = await repository.findOne();
```
 - entity manager
```
const user = getManager().findOne(User);
```
All these examples demonstrate using `findOne` with `undefined` value as id input and it is expected return one entity.
Also documentation from sources say:
```
/**
 * Finds first entity that matches given conditions.
 */
findOne(id?: string|number|Date|ObjectID, options?: FindOneOptions<Entity>): Promise<Entity|undefined>;
```
Given condition is nothing (`undefined`) so from my view it is "find one record and I do not have any condition". I think it is all right to return first record in table.

And of course about:
```
User.findOne({ where: { id: null } })
```
It will return undefined because there is no record with `id` = `null`

vlapo:
Is this real issue? I mean `User.findOne(undefined);` is same as `User.findOne();` but what is correct behaviour of `User.findOne();`? 

From jsdoc:
```
/**
 * Finds first entity that matches given conditions.
 */
```
But there are no conditions (`undefined`) and I think if there is no condition we have to return first entity like from `User.findOne({});`

Maybe change method signature and mark id or condition as required but it is really big change and it will be necessary sync at least all `find*` methods behaviour.

vlapo:
But there is function overloading:
```typescript
[...]

PR #4921, opened 2019-10-17
docs: adds note regarding dangerous findOne behavior

tjhillard:
Calling `findOne(undefined)` or `findOne(null)` results in potentially unexpected and dangerous behavior.

Ref:
* https://github.com/typeorm/typeorm/issues/4687
* https://github.com/typeorm/typeorm/issues/2500
* https://github.com/typeorm/typeorm/issues/4448

Given the amount of issues opened regarding this behavior, I think it warrants a note in the docs.

### DIFF
--- a/docs/repository-api.md
+++ b/docs/repository-api.md
@@ -212,6 +212,8 @@ const user = await repository.findOneOrFail(1);
 const timber = await repository.findOneOrFail({ firstName: "Timber" });
 ```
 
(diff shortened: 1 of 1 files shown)


pleerock:
Instead of adding this to the docs I would improve current behavior and throw an error if there are arguments and first one is null or undefined

pleerock:
Check my last comment in #2500

imnotjames:
Given that this will be available in the next minor release, we should probably document it in this one with a message that the behavior is deprecated and will be going away.

PR #4801, opened 2019-09-23
docs: link Active Record and Data Mapper to appropriate writeup about…

jayarjo:
… both

helps to quickly orientate newcomers and discover a precious explanation about ORM phylosphies

PR #4928, opened 2019-10-19
Fix README.md active-record-data-map link missing issue

calebx:


pleerock:
thank you!

jayarjo:
@calebx @pleerock this PR didn't fix the problem, did it? I think it's a problem in parser (that thing that converts md to the documentation on the site) - it fails to parse # properly.

I already mentioned it in a related PR as well: https://github.com/typeorm/typeorm/pull/4801#issuecomment-544487350

calebx:
@jayarjo it works good on current master branch, for the first two link in README.
is it what we expected?
while I still saw two bad link, maybe more. I will pick it up.

jayarjo:
@calebx I don't think it does. Check the raw MD and then what it is being converted to.

calebx:
@jayarjo 
oh. I see. the ```./docs``` will convert to ```#```
sorry did not see the covert logic in behind for website.

now it would work.
https://github.com/typeorm/typeorm/pull/5065

@pleerock

PR #4943, opened 2019-10-21
docs: Decorator Reference Mistake | line 254

carlocorradini:
Found a mistake at line 254.
The type of a _@PrimaryGeneratedColumn("uuid")_ must be a **string** and **not a number**.

Annotation 2019-10-21 203403


pleerock:
thanks!

PR #4945, opened 2019-10-21
docs: explain cascade options

mguida22:
Currently the docs only describe passing a `boolean` for the cascade option. It's also possible to provide an array of cascade options, so I've updated the docs to reflect that, and provided an example.


pleerock:
thanks

PR #4954, opened 2019-10-23
feat: add distinct on() support for postgres

swim:
Following on from #4109. This PR adds support for Postgres distinct on clause.

todo;
- [x] Update docs to cover distinct on clause use (orderBy etc)


vinimdocarmo:
I'm looking forward to it!

Issue #4923, opened 2019-10-18
typeorm_metadata table not in migrations when generating view sql


timojokinen:
**Issue type:**

[ ] question
[x] bug report
[ ] feature request
[ ] documentation issue

**Database system/driver:**

[ ] `cordova`
[ ] `mongodb`
[ ] `mssql`
[x] `mysql` / `mariadb`
[ ] `oracle`
[ ] `postgres`
[ ] `cockroachdb`
[ ] `sqlite`
[ ] `sqljs`
[ ] `react-native`
[ ] `expo`

**TypeORM version:**

[x] `latest`
[ ] `@next`
[ ] `0.x.x` (or put your version here)

**Steps to reproduce or a small repository showing the problem:**
When generating the SQL for a view with the cli command "migration:generate" the command creates a table called typeorm_metadata in the database without adding it to the migrations file. If you drop the schema and run the migrations again or run the migration on another machine it fails because the table doesn't exist.

A solution would be if the migration:run command creates the typeorm_metadata table if it doesn't exist like migration:generate does or if the migration:generate command adds the create SQL script to the migration.

A workaround is to add the sql create script for typeorm_metadata to your migrations manually.

Up:
``CREATE TABLE `typeorm_metadata` (`type` varchar(255) NOT NULL,`database` varchar(255) DEFAULT NULL,`schema` varchar(255) DEFAULT NULL,`table` varchar(255) DEFAULT NULL,`name` varchar(255) DEFAULT NULL,`value` text) ENGINE=InnoDB``

Down:
DROP TABLE `typeorm_metadata`

pablosproject:
Thanks, I just banged my head against this.

zaro:
Same here. What would make most sense I think is to move the creation of the table on run, instead of generate.

timojokinen:
@zaro There's an open PR that addresses this

jtushman:
Just ran into this as well.  Thanks for the work around.

moltar:
Same query for PG:

```sql
            CREATE TABLE typeorm_metadata (
                "type" varchar(255) NOT NULL,
                "database" varchar(255) DEFAULT NULL,
                "schema" varchar(255) DEFAULT NULL,
                "table" varchar(255) DEFAULT NULL,
                "name" varchar(255) DEFAULT NULL,
                "value" text
            )
```

niccolofanton:
[...]

PR #4956, opened 2019-10-23
fix: create typeorm_metatable when running migrations

BitPatty:
Proposed fix for #4923 
When generating migrations for views a `typeorm_metadata` table is created in which the views select statements are stored during the migration. However, the table won't be created during the actual execution of the migration. Thus, running a migration on a database which doesn't have the `typeorm_metadata` table will eventually fail. To avoid that, the table is created during a migration
if necessary.


pablosproject:
Any chances we got this merged any soon? Just ran into this issue and I think is important to have this in the library.

src/schema-builder/RdbmsSchemaBuilder.ts
```diff
@@ -98,17 +95,24 @@ export class RdbmsSchemaBuilder implements SchemaBuilder {
         }
     }
 
+    /**
+     * If the schema contains views, create the typeorm_metadata table if it doesn't exist yet
+     */
+    async createMetadataTableIfNecessary(): Promise<void> {
+        if (!this.queryRunner) this.queryRunner = this.connection.createQueryRunner("master");
```
pleerock: why did you add this check? it looks like this.queryRunner is created in all places where we use this method. Having it just for type is super confusing. But if its really needed (but I don't think so) it should also be properly released.

src/commands/MigrationRunCommand.ts
```diff
@@ -73,6 +74,12 @@ export class MigrationRunCommand implements yargs.CommandModule {
                     // noop
             }
 
+            const schemaBuilder = connection.driver.createSchemaBuilder();
+
+            if (schemaBuilder instanceof RdbmsSchemaBuilder) {
+                await schemaBuilder.createMetadataTableIfNecessary();
+            }
```
pleerock: can we make same inside MigrationExecutor#executePendingMigrations ? Near "await this.createMigrationsTableIfNotExist(queryRunner);" line to make sure its consistent and works not only when called from the CLI

pleerock: I guess need to remove this code now?

GSKryachko:
I also would like to see it merged as soon as possible.

pleerock:
This PR isn't finished. For some reason @BitPatty added metadata table creation code into `undoLastMigration` method, I'm completely confused why

BitPatty:
I'm just as confused on how this happened :/

michaeldever:
This would be awesome, hitting my head against this in test for the last little while.

coyoteecd:
@pleerock can you merge and release this? I've run into this too

imnotjames:
Can you fix the conflict

pleerock:
we don't have tests for cli commands yet.. if someone can contribute and start writing tests for CLI for the first time, this one is a good opportunity. If nobody wants, probably we can merge this one.

BitPatty:
@imnotjames fixed the conflict

BitPatty:
@imnotjames Can you check the tests?

imnotjames:
> @imnotjames Can you check the tests?

They're currently failing, it looks like?

BitPatty:
The CI build had a timeout during service setup, it didn't run the tests

pleerock:
@AlexMesser can you please review this PR?

pleerock:
Okay, I have to review it by myself.

src/schema-builder/RdbmsSchemaBuilder.ts
```diff
@@ -99,17 +96,24 @@ export class RdbmsSchemaBuilder implements SchemaBuilder {
         }
     }
 
+    /**
+     * If the schema contains views, create the typeorm_metadata table if it doesn't exist yet
+     */
+    async createMetadataTableIfNecessary(): Promise<void> {
+        if (this.viewEntityToSyncMetadatas.length > 0) {
+            this.queryRunner = this.queryRunner || this.connection.createQueryRunner();
```
pleerock, line 104: `|| this.connection.createQueryRunner()` this code can be very dangerous since its not clear here where it gets **released**. Use `this.queryRunner` if it's available, or pass it to `createMetadataTableIfNecessary` method from the places where you call this method.

BitPatty: On a side note, with both the current implementation [1]/[2] as well as this PR the migration table is created when **generating** migrations. Is this intended behavior?

[1] https://github.com/typeorm/typeorm/blob/a92120ac8821461055993a36ddba17d3466de925/src/commands/MigrationGenerateCommand.ts#L105
[2] https://github.com/typeorm/typeorm/blob/a92120ac8821461055993a36ddba17d3466de925/src/schema-builder/RdbmsSchemaBuilder.ts#L116-L125

pleerock: > the migration table is created when generating migrations. Is this intended behavior?

Probably it shouldn't be inside `log` method, since you are creating metadata table in `executePendingMigrations`, right?

TheNoim:
I just noticed this could also be relevant for my pr #6469, because it uses the ´typeorm_metadata´ table.

AlexMesser:
thank you for contribution!

issackr:
@pleerock @BitPatty 
Hi
I am getting this error while having my server that run migrations going up
typeorm_metadata table already exists
It looks like this behaviour started around the merging of this fix, do you think it might be connected?

Thanks

BitPatty:
You might have patched a migration manually to explicitly create the typeorm_metadata table. Check your migration files for any `CREATE TABLE typeorm_metadata` statements.

PR #4992, opened 2019-10-30
Fix typo: "Sasha" to "Sascha"

Sascha8a:


LorenzHenk:
According to [this](https://en.wikipedia.org/wiki/Sasha_(name)) Wikipedia entry, Sascha is the German form of Sasha.

Please close this PR.

pleerock:
Yeah there can be many form of same name in different countries / regions. Closing as not relevant PR. Thanks.

PR #5006, opened 2019-10-31
Add note to execute query without 'null' values

dreamchasersuon:
It's useful for those who trying to do so first time. I will point it also in 'find-options' section below 'IsNull' explanation.


pleerock:
please read my comment here https://github.com/typeorm/typeorm/pull/5009#issuecomment-549341320

PR #5009, opened 2019-10-31
Add reference to 'is not null' expression

dreamchasersuon:
It's confusing when you write about `IsNull()` function to compare that value is equal to `null`, but never wrote about how to get value which `is not null`.

### DIFF
--- a/docs/find-options.md
+++ b/docs/find-options.md
@@ -325,6 +325,7 @@ const loadedPosts = await connection.getRepository(Post).find({
     title: IsNull()
 });
 ```
+Note: if you want execute expression where value not equal to `null` - follow the note in this section https://typeorm.io/#/select-query-builder/adding-where-expression
 
 will execute following query: 
 


pleerock:
Not sure if I get your request... How is this particular operator related to IS NOT NULL? Checkout the last operator called "Not". You can combine it with any other operator and do "title: Not(isNull())"

dreamchasersuon:
> Not sure if I get your request... How is this particular operator related to IS NOT NULL? Checkout the last operator called "Not". You can combine it with any other operator and do "title: Not(isNull())"

Sure. This will work fine:
 ```
userRepository.find({ 
    join: {
        alias: "user",
        leftJoinAndSelect: {
            profile: "user.profile",
            photo: "user.photos",
            video: Not(IsNull())
        }
    }
});
```

But what if your approach is different:
```
await this.createQueryBuilder()
      .leftJoin(BatchEntity, 'batch', 'company.id = batch.companyId')
      .andWhere('batch.id IN (:batchIds)', { batchIds })
      .andWhere('batch.task is not null)
      .getMany();
```

So, my thoughts to include it for another side.

pleerock:
Since this documentation is about find options and not query builder, I don't see a point including a link to a query builder implementation, this can confuse people. Thanks

PR #5016, opened 2019-11-01
RedisCache: getFromCache

wisekaa03:
When redis is not initialized


pleerock:
can you please explain it in details? when happens what and why? Thanks.

wisekaa03:
when I used hot.module, I closed the connection and reopened. but the connection to the cache was not restored.

wisekaa03:
I understand that this could be written differently, but at that time I did not have time.

pleerock:
I'm going to close this issue. The problem most probably in your code. You should access things without waiting for a connection to be established. Also, your code is going to cause a never-resolved-promise.

PR #5020, opened 2019-11-01
Fix docs description 

rhogeranacleto:
## This pull request fixes the description of `synchronize` method

I guess the `synchronize` method has the wrong description. I've changed the description to indicate that `synchronize` is usually used on application starting instead of application shut down.


pleerock:
Thanks!

PR #5023, opened 2019-11-02
Upgrade app-root-path

BlackDark:
Fixes incompatbility with pnpm which couldn't find the `ormconfig.json` in the root path.


pleerock:
Thanks!

PR #5026, opened 2019-11-04
Update DeepPartial Type

Benno007:
Update the types to match this code in the typescript bug (https://github.com/Microsoft/TypeScript/issues/21592) that is causing `Excessive stack depth comparing types`

https://github.com/Microsoft/TypeScript/issues/21592#issuecomment-496723647


Benno007:
Somehow accidentally opened this instead of our fork :|

PR #5036, opened 2019-11-05
fix: mysql driver query streaming

### LINKED ISSUE #5035: MySQL driver: incorrect return value in QueryRunner.stream()
**Issue type:**

[ ] question
[x] bug report
[ ] feature request
[ ] documentation issue

**Database system/driver:**

[ ] `cordova`
[ ] `mongodb`
[ ] `mssql`
[x] `mysql` / `mariadb`
[ ] `oracle`
[ ] `postgres`
[ ] `cockroachdb`
[ ] `sqlite`
[ ] `sqljs`
[ ] `react-native`
[ ] `expo`

**TypeORM version:**

[x] `latest`
[ ] `@next`

Type definition for QueryRunner.stream:
```ts
stream(query: string, parameters?: any[], onEnd?: Function, onError?: Function): Promise<ReadStream>;
```

Let's see if this is true for MySQL:

```ts
const q = this.connection.createQueryRunner();
const s = await q.stream('select * from users order by id limit 10');
console.log(s)
```

`s` appears to be node-mysql's `Query` which is an EventEmitter but not a ReadableStream. In order to get a stream one needs to call `.stream()` on a Query:

```ts
const q = this.connection.createQueryRunner();
const s = (await q.stream('select * from users order by id limit 10') as any).stream();
console.log(s)
```

The bug appears to be here: https://github.com/typeorm/typeorm/blob/master/src/driver/mysql/MysqlQueryRunner.ts#L190
The value returned is a query, not a stream. After setting event handlers, the promise should be resolved with `query.stream()`.

constb:
Mysql driver's `connection.query()` returns `Query` object which is an EventEmitter but not a ReadableStream. In order to get a stream one needs to call `.stream()` on it.

Fixes #5035

### DIFF
--- a/src/driver/mysql/MysqlQueryRunner.ts
+++ b/src/driver/mysql/MysqlQueryRunner.ts
@@ -187,10 +187,10 @@ export class MysqlQueryRunner extends BaseQueryRunner implements QueryRunner {
             try {
                 const databaseConnection = await this.connect();
                 this.driver.connection.logger.logQuery(query, parameters, this);
-                const stream = databaseConnection.query(query, parameters);
-                if (onEnd) stream.on("end", onEnd);
-                if (onError) stream.on("error", onError);
-                ok(stream);
+                const databaseQuery = databaseConnection.query(query, parameters);
+                if (onEnd) databaseQuery.on("end", onEnd);
+                if (onError) databaseQuery.on("error", onError);
+                ok(databaseQuery.stream());
 
             } catch (err) {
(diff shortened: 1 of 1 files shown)


pleerock:
Thanks! Can we have a test that will make sure we won't have this issue again in the future?

constb:
I'm afraid I don't have enough experience writing tests. I don't even know how to run ones that already exist in the repo. :(

Issue #5011, opened 2019-10-31
Invalid sequence creation when adding to well known


Ginden:
**Issue type:**

[ ] question
[x] bug report
[ ] feature request
[ ] documentation issue

**Database system/driver:**

[ ] `cordova`
[ ] `mongodb`
[ ] `mssql`
[ ] `mysql` / `mariadb`
[ ] `oracle`
[x] `postgres`
[ ] `cockroachdb`
[ ] `sqlite`
[ ] `sqljs`
[ ] `react-native`
[ ] `expo`

**TypeORM version:**

[x] `latest`
[ ] `@next`
[ ] `0.x.x` (or put your version here)

tl;dr:

Given class:
```typescript
class Banner {
   @PrimaryGeneratedColumn()
   public id: number;

    @Column({type: 'int', generated: 'increment'})
    public camelCaseColumnId: number;
}
```
TypeORM generates wrong migration.
```
await queryRunner.query(`CREATE SEQUENCE "banner_camelCaseColumnId_seq" OWNED BY "nps_banner"."pollId"`);
await queryRunner.query(`ALTER TABLE "banner" ALTER COLUMN "camelCaseColumnId" SET DEFAULT nextval('banner_camelCaseColumnId_seq`);
 ```

This fails with message `ERROR: relation "banner_camelcasecolumnid_seq" does not exist`.

Second line should be:
```typescript
await queryRunner.query(`ALTER TABLE "banner" ALTER COLUMN "camelCaseColumnId" SET DEFAULT nextval('"banner_camelCaseColumnId_seq"`); // double quote marks in string
````

imnotjames:
This seems to have been fixed already.

PR #5039, opened 2019-11-05
Escape sequence name in nextval call

Ginden:
fixes: #5011


imnotjames:
Can you include tests on this to confirm it fixes that behavior?

imnotjames:

PR #5052, opened 2019-11-06
Fix repo for app-root-path in lock file

BlackDark:
Missed some configurations on my side in the last PR.
See https://github.com/typeorm/typeorm/commit/7f87f0c9da6f74e14d167c3dfffdd351ca2d2f9f#diff-32607347f8126e6534ebc7ebaec4853dR460

Now reverted back to official npm repo for the `app-root-path` package

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -457,8 +457,8 @@
     },
     "app-root-path": {
       "version": "3.0.0",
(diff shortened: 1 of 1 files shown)


pleerock:
Thanks!

PR #5053, opened 2019-11-06
docs: fix broken DataMapper/ActiveRecord links

mcky:


pleerock:
Thanks a lot, we already fixed it in #5065.

PR #5063, opened 2019-11-10
feat: support options that can be used when clearing/truncating tables

### LINKED ISSUE #2978: Support for CASCADEd TRUNCATE in PostgreSQL
**Issue type:**

[ ] question
[ ] bug report
[x] feature request
[ ] documentation issue

**Database system/driver:**

[ ] `cordova`
[ ] `mongodb`
[ ] `mssql`
[ ] `mysql` / `mariadb`
[ ] `oracle`
[x] `postgres`
[ ] `sqlite`
[ ] `sqljs`
[ ] `react-native`
[ ] `expo`

**TypeORM version:**

[ ] `latest`
[ ] `@next`
[x] `0.2.7` (or put your version here)

Have 2 two tables defined like:

```
@Entity()
export class User 
{
    @PrimaryGeneratedColumn()
    id: number;

    @Column()
    email: string;
    
    @Column()
    username: string;

    @Column({type: "time with time zone"})
    signedUp: Date;

    @Column()
    firstName: string;

    @Column()
    lastName: string;

    @Column({type: "enum", enum: RoleType})
    role: RoleType;
    
    @Column()
    hashedPassword: string;

    @Column()
    hashedPasswordSalt: string;
}

```

```
@Entity()
export class Session 
{
    @PrimaryGeneratedColumn()
    id: number;

    @Column({name: "userId", type: "bigint"})
    @ManyToOne(type => User)
    @JoinColumn()
    user: User;
    
    @Column()
    sessionHash: string;

    @Column()
    sessionDetails: string;

    @Column({type: "time with time zone"})
    validTo: Date;

    @Column({type: "time with time zone"})
    lastActive: Date;
 }

```

Then attempt to truncate the data using:

`        await getConnection().getRepository(User).clear();`

This will produce error like:

> ERROR:  cannot truncate a table referenced in a foreign key constraint
> DETAIL:  Table "session" references "user".
> HINT:  Truncate table "session" at the same time, or use TRUNCATE ... CASCADE.

Interim hack is to:
`await getConnection().getRepository(User).query("TRUNCATE \"user\" CASCADE");`

I would expect solution like:
```
    /**
     * Clears all the data from the given table/collection (truncates/drops it).
     *
     * Note: this method uses TRUNCATE and may not work as you expect in transactions on some platforms.
     * @see https://stackoverflow.com/a/5972738/925151
     */
    clear(cascaded? boolean): Promise<void>;
```

tzellman:
current change only affects the downstream Postgres QueryRunner, which supports cascading when truncating

Closes #2978

### DIFF
--- a/src/driver/BaseClearOptions.ts
+++ b/src/driver/BaseClearOptions.ts
@@ -1,0 +1,12 @@
+import {DatabaseType} from "./types/DatabaseType";
+
+/**
+ * BaseClearOptions is set of options shared by all database types.
+ */
+export interface BaseClearOptions {
+
+    /**
+     * Database type. This value is required.
+     */
+    readonly type: DatabaseType;
+}
--- a/src/driver/postgres/PostgresClearOptions.ts
+++ b/src/driver/postgres/PostgresClearOptions.ts
@@ -1,0 +1,17 @@
+import {BaseClearOptions} from "../BaseClearOptions";
+
+/**
+ * Postgres-specific connection options.
+ */
+export interface PostgresClearOptions extends BaseClearOptions {
+
+    /**
+     * Database type.
+     */
+    readonly type: "postgres";
+
+    /**
+     * Cascade when truncating.
+     */
+    readonly cascade?: boolean;
+}
--- a/src/driver/postgres/PostgresQueryRunner.ts
+++ b/src/driver/postgres/PostgresQueryRunner.ts
@@ -22,6 +22,7 @@ import {OrmUtils} from "../../util/OrmUtils";
 import {Query} from "../Query";
 import {IsolationLevel} from "../types/IsolationLevel";
 import {PostgresDriver} from "./PostgresDriver";
+import {PostgresClearOptions} from "./PostgresClearOptions";
 
 /**
  * Runs queries on a single postgres database connection.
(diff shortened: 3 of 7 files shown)


tzellman:
Definitely open to feedback on this, especially if there is a solution/strategy that fits better within the typeorm style and goals.

The reasoning behind the approach I took was to avoid leaking driver-specific capabilities into the abstract API. For example, only the Postgres driver supports a `CASCADE` option. Additionally (and not included in this PR), Oracle supports some additional options when truncating (https://docs.oracle.com/cd/B19306_01/server.102/b14200/statements_10006.htm).

pleerock:
I know you wanted to make types database-specific, but passing "type" all the time to this method (and same methods if we would have them in the future) looks stupid for API users.

Having it typed is great, but not this way. And there is no elegant way of doing it typed, without a full rewrite of connection, repository and entity manager objects. But its definitely something that we are going to do in the future, to allow such features. So for now, I recommend to make it same options for all drivers, add a runtime check (not supported options) and test-coverage everything. 

Thanks :)

imnotjames:
This needs to be rebased, conflicts fixed, and needs tests before it can be merged.

imnotjames:

PR #5065, opened 2019-11-11
fix README link to docs

calebx:
https://github.com/typeorm/typeorm/pull/4928
this should work


jayarjo:
Chinese docs have to be fixed as well.

calebx:
@jayarjo yeah.
it could make the html page happy.
while, in https://github.com/typeorm/typeorm/blob/master/README-zh_CN.md github readme file.
those links would redirect to EN version. and that is an issue for all link there. 
I'd keep it still since that README is for page only.

pleerock:
It's still seems to be broken. Can you please check how do we do it in other places? Also, we need to make sure that links are working from both site and github when you click the link from the github site.

pleerock:
okay so as was mentioned it was a site parser issue. Fixed in https://github.com/typeorm/typeorm.github.io/commit/2b6a05432eafa7f750023087a4216a37413585b2

calebx:
👍

PR #5072, opened 2019-11-12
fix: mssql change column default value

szamanis:
Hello, 
I m using typeorm with mssql and I stumble upon a problem when I tried to change the default value of a column and then synchronize. 
The problem was that the default column was being added again and not dropped first, resulting to an sql error. 
I looked into the code and made a fix and added the corresponding unit test into functional directory. This should also fix migration issues.

I believe this test is mandatory regardless of the database.


pleerock:
These are too much dangerous changes to merge. Hope @AlexMesser can review them.

AlexMesser:
I've tested it and it looks like the issue was already fixed. Thank you for contribution!

Issue #4452, opened 2019-07-17
Support for SYSDATE in oracle Date column


rjmohammadww:
**Issue type:**

[ ] question
[ ] bug report
[x] feature request
[ ] documentation issue

**Database system/driver:**

[ ] `cordova`
[ ] `mongodb`
[ ] `mssql`
[ ] `mysql` / `mariadb`
[ x] `oracle`
[ ] `postgres`
[ ] `cockroachdb`
[ ] `sqlite`
[ ] `sqljs`
[ ] `react-native`
[ ] `expo`

**TypeORM version:**

[x ] `latest`
[ ] `@next`
[ ] `0.x.x` (or put your version here)

**Steps to reproduce or a small repository showing the problem:**

Can't use SYSDATE in date column fields for Oracle because of this line 
https://github.com/typeorm/typeorm/blob/master/src/driver/oracle/OracleDriver.ts#L364
the logic tries to parse the string or function being passed down.

Sample Oracle Query
```
INSERT INTO SOME_TABLE (Last_Upd_By, Last_Upd_Date)

VALUES ( P_Last_Upd_By, SYSDATE);
```

Sample Code:
Entity:
```
@Column('date', {
    nullable: true,
    name: 'CANCEL_REQ_DATE'
  })
  CANCEL_REQ_DATE: Date | null;

  @Column('date', {
    nullable: true,
    name: 'PLAN_END_DATE'
  })
  PLAN_END_DATE: Date | null;

@Column('varchar2', {
    nullable: true,
    length: 20,
    name: 'LAST_UPD_BY'
  })
  LAST_UPD_BY: string | null;
```
insertion:

```await cancelRepo.insert({
    LAST_UPD_BY: lastUpdatedBy,
    CANCEL_REQ_DATE: 'SYSDATE', // for issue sample
    LAST_UPD_DATE: () => 'SYSDATE', // for issue sample
  });
``` 
Both **SYSDATE** format gives error.

Possible solutions may include adding a conditional on line #360
```
if (typeof value === 'function') {
  return value
}
```
Above proposed solution will give alot of flexibility to the users to run their desired queries/SQL functions if they don't want the default behavior.

The current workaround on my end includes creating a patch file that patches that piece of code at the time of my docker container creation.

 I can make a PR if we are fine with my proposed solution. 

Thanks!
RJ

PR #5076, opened 2019-11-12
fix: resolves oracle sql expression bug for date columns

rjmohammadww:
Closes: #4452

### DIFF
--- a/docs/many-to-many-relations.md
+++ b/docs/many-to-many-relations.md
@@ -237,11 +237,11 @@ Additionally you will have to add a relationship like the following to `Post` an
 ```typescript
 // category.ts
 ...
-@OneToMany((type) => PostToCategory, (postToCategory) => postToCategory.category)
+@OneToMany(type => PostToCategory, postToCategory => postToCategory.category)
 public postToCategories!: PostToCategory[];
 
 // post.ts
 ...
-@OneToMany((type) => PostToCategory, (postToCategory) => postToCategory.post)
+@OneToMany(type => PostToCategory, postToCategory => postToCategory.post)
 public postToCategories!: PostToCategory[];
 ```


pleerock:
We can't fix it only for Oracle driver and only for a date column.

Instead, I recommend you to put a function check [here](https://github.com/typeorm/typeorm/blob/ee61c518d746139babbbeaa19b8f3d6e87e34f08/src/query-builder/InsertQueryBuilder.ts#L406) and see how its going to work (it should cover all drivers and column types). You can see we already do it in update method (https://github.com/pleerock/typeorm/blob/ee61c518d746139babbbeaa19b8f3d6e87e34f08/src/query-builder/UpdateQueryBuilder.ts#L397).

Also we'll need a test coverage for this issue, otherwise it can easily break very soon.
 Thank you.

rjmohammadww:
@pleerock  please correct me if I am mistaken but based on my understanding from the driver code, this problem only exists for oracle driver because the way this statement is written

```

//OracleDriver.ts Line #367
else if (columnMetadata.type === "date") {
            if (typeof value === "string")
                value = value.replace(/[^0-9-]/g, "");
  ---> This one          return () => `TO_DATE('${DateUtils.mixedDateToDateString(value)}', 'YYYY-MM-DD')`;

        }
```

https://github.com/typeorm/typeorm/blob/1047989796e441a084001570f1656d49e44a9a07/src/driver/oracle/OracleDriver.ts#L367


This statement forces to return a function without any checks to see if the value is a function.
```() => `TO_DATE('${DateUtils.mixedDateToDateString(value)}', 'YYYY-MM-DD')`;```


This problem does not exist for other drivers because other drivers simply do
```
else if (columnMetadata.type === "date") {
            value = DateUtils.mixedDateToDateString(value);
        } 
```
e.g https://github.com/typeorm/typeorm/blob/1047989796e441a084001570f1656d49e44a9a07/src/driver/postgres/PostgresDriver.ts#L408
https://github.com/typeorm/typeorm/blob/1047989796e441a084001570f1656d49e44a9a07/src/driver/sqlserver/SqlServerDriver.ts#L424

and . `mixedDateToDateString` just returns the value, if the value is not an instance of date
https://github.com/typeorm/typeorm/blob/0a13e6ac3c5be1fc0a29aba4f5e5b69a474a5927/src/util/DateUtils.ts#L29

**Comment by pleerock** [comment]
no, here your original request is to use a function-syntax to insert SQL expressions as a column value. This is proper for any value in any database, no matter if its a date value or not.

**Comment by rjmohammadww** [comment]
@pleerock  Sorry for the confusion, and let me try to elaborate on what I am saying. If my understanding is wrong, I promise, I shall change my PR with the updates from the review.

The problem I am trying to fix only exists for the Oracle driver and date column.
I have tried other SQL expressions on other column types and they work as expected for Oracle. Furthermore, this test would fail if that wasn't the case for any other driver

https://github.com/typeorm/typeorm/blob/17f3224c58b7126a9c1360ce21f43cda83a35e04/test/functional/query-builder/insert/query-builder-insert.ts#L79


The reason this problem exists only for oracle date column is that this statement only exists in Oracle diver
https://github.com/typeorm/typeorm/blob/1047989796e441a084001570f1656d49e44a9a07/src/driver/oracle/OracleDriver.ts#L367


What ends up happening is that when the date column value is a SQL expression function, Line 367 returns it as 

```
 () => `TO_DATE('${DateUtils.mixedDateToDateString(() => SQL EXPRESSION)}', 'YYYY-MM-DD')`;
```

All other drivers just simply return the value without any conversion if the value is not an instance of  date
https://github.com/typeorm/typeorm/blob/0a13e6ac3c5be1fc0a29aba4f5e5b69a474a5927/src/util/DateUtils.ts#L26

pleerock:
> The problem I am trying to fix only exists for the Oracle driver and date column.

its not origin problem. Source of the problem is that we should allow users to use functions when they insert a column value, no matter if they are using oracle or date column. () => string SHOULD BE SUPPORTED BY ANY DRIVER ON ANY COLUMN TYPE DURING INSERTION, which is not properly working right now from some cases like date column and others I guess. That's original problem I suggest you to fix.

PR #5078, opened 2019-11-13
remove connection options mutation

silentroach:
Just accidently discovered that connection options are modified if url is passed.

Steps to reproduce:

```ts
const options = {
    type: 'postgres',
    url: 'postgresql://postgres:somepassword@localhost:5432/postgres',
    ...
}

await createConnection(options).connect(); // ok

console.log(options.type); // 'postgresql' (changed from valid postgres)

await createConnection(options).connect(); // will not connect cause of 'postgresql' type is not supported
```


pleerock:
I'll merge it, however its not test covered. And since change isn't obvious it can easily break again next time somebody change this code. If you want a long-term support for this change, please add tests in a different PR. Thanks.

PR #5097, opened 2019-11-17
fix: better mergeDeep implementation and tests

### LINKED ISSUE #5096: Inserting entities with complex objects with circular references results in stack overflow
**Issue type:**
[x] bug report

**Database system/driver:**
[x] `sqlite`

**TypeORM version:**

[x] `latest`
[x] `0.2.20`

```js
import {
  Instant
} from '@js-joda/core';

class InstantTransformer implements ValueTransformer {
  to(entityValue: Instant | null): number | null {
    if (entityValue === null) { return null; }
    return entityValue.toEpochMilli()
  }

  from(databaseValue: number | null): Instant | null {
    if (databaseValue === null) { return null; }
    return Instant.ofEpochMilli(databaseValue)
  }
}

@Entity()
export class Project {
  uuid: string;

  @Column({
    transformer: new InstantTransformer()
  })
  creationInstant: Instant;
}
```

When attempting to create a `Project`, something like this:
```js
  const project = new Project();
  project.creationInstant = Instant.parse(creationInstant);
  await connection.manager.save(project);
```

We end up with:
```
RangeError: Maximum call stack size exceeded
    at Array.shift (<anonymous>)
    at Function.OrmUtils.mergeDeep (node_modules/typeorm/util/OrmUtils.js:70:30)
    at Function.OrmUtils.mergeDeep (node_modules/typeorm/util/OrmUtils.js:86:26)
...
    at Function.OrmUtils.mergeDeep (node_modules/typeorm/util/OrmUtils.js:63:19)
    at node_modules/typeorm/persistence/Subject.js:169:33
    at Array.reduce (<anonymous>)
    at Subject.createValueSetAndPopChangeMap (node_modules/typeorm/persistence/Subject.js:125:41)
    at node_modules/typeorm/persistence/SubjectExecutor.js:237:85
    at Array.forEach (<anonymous>)
    at SubjectExecutor.<anonymous> (node_modules/typeorm/persistence/SubjectExecutor.js:225:58)
    at step (node_modules/tslib/tslib.js:136:27)
    at Object.next (node_modules/tslib/tslib.js:117:57)
    at node_modules/tslib/tslib.js:110:75
```

---

The best solution should probably change this mess:
https://github.com/typeorm/typeorm/blob/ed5ebb082560f8fd56ffd5b7eaf60fcd5a3d26c5/src/util/OrmUtils.ts#L84-L88
Into a solution which uses `getPrototypeOf`, `constructor`, and etc to see if an object is a plain object like [lodash does](https://github.com/lodash/lodash/blob/4.17.15/lodash.js#L12036).

fer22f:
The current mergeDeep implementation tries to recursively copy complex objects like class instances. So, a new implementation of `mergeDeep` and `isPlainObject` (previously `isObject`) is given. This fixes #5096, which demonstrates the problem.

As a bonus, this creates tests for these methods which weren't tested before.

### DIFF
--- a/src/persistence/SubjectExecutor.ts
+++ b/src/persistence/SubjectExecutor.ts
@@ -352,7 +352,7 @@ export class SubjectExecutor {
 
             // for mongodb we have a bit different updation logic
             if (this.queryRunner instanceof MongoQueryRunner) {
-                const partialEntity = OrmUtils.mergeDeep({}, subject.entity!);
+                const partialEntity = OrmUtils.mergeDeep({}, { ...subject.entity! });
                 if (subject.metadata.objectIdColumn && subject.metadata.objectIdColumn.propertyName) {
                     delete partialEntity[subject.metadata.objectIdColumn.propertyName];
                 }
--- a/src/util/OrmUtils.ts
+++ b/src/util/OrmUtils.ts
@@ -70,11 +70,10 @@ export class OrmUtils {
         // Filters out Buffer and custom instances of classes
         const prototype = Object.getPrototypeOf(item);
         return prototype === null || prototype === Object.getPrototypeOf({});
-
     }
 
     /**
-     * Deep Object.assign.
+     * Deep Object.assign for simple objects.
      *
      * @see http://stackoverflow.com/a/34749873
      */
--- a/src/util/OrmUtils.ts
+++ b/src/util/OrmUtils.ts
@@ -58,8 +58,19 @@ export class OrmUtils {
         }, [] as T[]);
     }
 
-    static isObject(item: any) {
-        return (item && typeof item === "object" && !Array.isArray(item));
+    // Checks if it's an object made by Object.create(null), {} or new Object()
+    static isPlainObject(item: any) {
+        // Filters out undefined, null, function, and primitive array, number, string
+        // Less expensive call than toString
+        const isObject = item != null && typeof item === "object" && !Array.isArray(item);
+        if (!isObject) { return false; }
+        // Filters out Date, Set, Map, String, Math, Number and a few others
+        const isObjectObject = Object.prototype.toString.call(item) === "[object Object]";
+        if (!isObjectObject) { return false; }
+        // Filters out Buffer and custom instances of classes
+        const prototype = Object.getPrototypeOf(item);
+        return prototype === null || prototype === Object.getPrototypeOf({});
+
     }
 
     /**
@@ -71,7 +82,7 @@ export class OrmUtils {
         if (!sources.length) return target;
         const source = sources.shift();
 
-        if (this.isObject(target) && this.isObject(source)) {
+        if (this.isPlainObject(target) && this.isPlainObject(source)) {
             for (const key in source) {
                 let propertyKey = key;
                 if (source[key] instanceof Promise)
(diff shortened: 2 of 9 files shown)


fer22f:
I guess the tests pass, but I'm a bit worried about edge cases, maybe something like passing a class instance to an `insert`, or some kind of recursive scenario?

pleerock:
Tests can easily pass because we don't have a large test coverage for monogdb driver.

I think your changes won't work properly because in the cases when embedded objects (nested objects) are used, we actually need class instances, we can't skip them.

Solution to this issue is to rewrite current way of "copying object" - traverse over defined columns in the class and all its embeddeds and copy only object properties that are defined in these columns. Columns definition tells us what schema should be used for saving and this schema can't be recursive.

fer22f:
@pleerock Ok, thanks! I will work some more on this then. I will add more tests regarding having nested embedded objects and objects with more information than the schema allows (should we throw an error or something if we detect that, or just ignore?).

I will work on making a new function to transform nested entities into plain objects using schema (probably using `getMetadata` and `.prototype`, I still don't know exactly how and how well it will scale), and then apply that instead of `{ ... }`, and I think then we will be ready!

fer22f:
> Tests can easily pass because we don't have a large test coverage for monogdb driver.
> 
> I think your changes won't work properly because in the cases when embedded objects (nested objects) are used, we actually need class instances, we can't skip them.

@pleerock I was looking into this again, and it looks like there is a test for this, and [it passed](https://travis-ci.org/typeorm/typeorm/jobs/613215120#L695):

https://github.com/typeorm/typeorm/blob/4307a596117306b916f131810cbc45bde6404a0c/test/functional/mongodb/basic/embedded-columns/mongodb-embedded-columns.ts#L134-L158

So I don't know exactly how I could break this, since I saw that it passed that, it looks to me to be pretty solid (even though it really doesn't look like it).

papaiatis:
@pleerock is this ever going to get merged any time soon?

src/persistence/SubjectExecutor.ts
```diff
@@ -352,7 +352,7 @@ export class SubjectExecutor {
 
             // for mongodb we have a bit different updation logic
             if (this.queryRunner instanceof MongoQueryRunner) {
-                const partialEntity = OrmUtils.mergeDeep({}, subject.entity!);
+                const partialEntity = OrmUtils.mergeDeep({}, { ...subject.entity! });
```
imnotjames, line 355: Why shallow cloning the object here?  Doesn't this code that already?

src/util/OrmUtils.ts
```diff
@@ -58,20 +58,30 @@ export class OrmUtils {
         }, [] as T[]);
     }
 
-    static isObject(item: any) {
-        return (item && typeof item === "object" && !Array.isArray(item));
+    // Checks if it's an object made by Object.create(null), {} or new Object()
+    static isPlainObject(item: any) {
```
imnotjames, line 62: If you're dropping `isObject` can you make this private?

test/functional/columns/value-transformer/entity/Post.ts
```diff
@@ -27,4 +59,6 @@ export class Post {
     @Column({ type: String, transformer: new TagTransformer() })
     tags: string[];
 
+    @Column({ type: String, transformer: new ComplexTransformer(), nullable: true })
+    complex: Complex | null;
```
imnotjames, line 63: Is it possible to add this to a different entity in a different entity file?

That way we're not muddying up the original tests

test/functional/util/OrmUtils.ts
```diff
@@ -0,0 +1,106 @@
...
+        expect(isPlainObject([])).to.be.false;
+        expect(isPlainObject(new Foo())).to.be.false;
+        expect(isPlainObject(null)).to.be.false;
+        expect(isPlainObject(Math)).to.be.false;
+        expect(isPlainObject(Error)).to.be.false;
+        expect(isPlainObject(0)).to.be.false;
+        expect(isPlainObject(false)).to.be.false;
+        expect(isPlainObject(NaN)).to.be.false;
+    });
+ });
```
imnotjames, line 37: since we're doing all the tests below - do we need to test `isPlainObject` if it's just an internal helper?  Do the below tests not go through all the possible permutations?

imnotjames (changes requested):
Can you rebase & fix the conflict?

PR #5099, opened 2019-11-18
Change getManager to getRepository

PedramMarandi:
docs: update the mistaken import in doc.

replace `getRepository` with `getManager`.


pleerock:
Thank you!

PR #5122, opened 2019-11-21
Update links to Active Record and Data Mapper doc

denilsonsa:
Follow-up of e42091e1506148ba761737d7977331b5c1d60307 and #4928

This is probably duplicate of #5065. Still, this commit shows a way to define links without having to put the URLs inline.


pleerock:
Thanks for the PR, but we need inline approach. 
Also, #5065 doesn't fix the issue.
Also, we need to make sure that links are working from both site and github when you click the link from the github site.

PR #5133, opened 2019-11-22
Fix circular dependency error message

glentakahashi:
I believe this should say `nullable: true`. You only have an impossible circular dependency if both relations are `NOT NULL`.


pleerock:
Thanks

Issue #5114, opened 2019-11-20
double comma raw query if use order by FIELD id


winzaa123:
**Issue type:**

[ ] question
[x] bug report
[ ] feature request
[ ] documentation issue

**Database system/driver:**

[ ] `cordova`
[ ] `mongodb`
[ ] `mssql`
[x ] `mysql` / `mariadb`
[ ] `oracle`
[ ] `postgres`
[ ] `cockroachdb`
[ ] `sqlite`
[ ] `sqljs`
[ ] `react-native`
[ ] `expo`

**TypeORM version:**

[ ] `latest`
[ ] `@next`
[x] `0.2.19` (or put your version here)

**Steps to reproduce or a small repository showing the problem:**

I need order 2 condition if i order below

```ts
query.orderBy(`FIELD(job_statusId,${[3,2].join(',')})`,'DESC')
        .addOrderBy('job.createdAt','DESC') 
```
i will get Query
![image](https://user-images.githubusercontent.com/15700798/69222124-66c94100-0bab-11ea-881e-a0ed4689e9c2.png)

if i selects.replace(',','') will can run my code but not sure position code in Typescript (query-builder\SelectQueryBuilder.js)
![image](https://user-images.githubusercontent.com/15700798/69222062-44cfbe80-0bab-11ea-9e38-f576a1177953.png)

<!--
    To answer those questions you need to put "x" inside the square brackets, for example:
    [x] `mysql`
    [ ] `postgres`

    Also, please format your code properly (by taking code blocks into ```ts .... ```)
!>

winzaa123:
0.2.20 same too.

ong-ar:
0.2.24 same too

smdehaas:
Did anyone find a workaround this issue?

RajThuppanna-wt:
Is there a plan to fix this issue? I am forced to put upgrade plans on hold due to this issue. I badly need to upgrade other modules that depend on this version of TypeORM.

coerdes-chris:
This is still happening in 0.3.7
A potential, temporary, workaround:
```js
const back = SelectQueryBuilder.prototype.addSelect;
SelectQueryBuilder.prototype.addSelect = function (...args) {
  // If the first character of the first argument is a ',',
  if (args[0][0] === ',') {
    // Remove the ','
    args[0] = args[0].substring(1);
  }
  return back.apply(this, args);
};
```

Nightbr:
[...]

PR #5138, opened 2019-11-25
fix: filter create order case empty string

winzaa123:
fix issue : #5114


pleerock:
Not sure I can merge it right now since I don't remember why we needed that `.replace(",", "")` feature for.

winzaa123:
image

i just don't need "" empty string in createOrderByCombinedWithSelectExpression function

imnotjames:
Can you create tests that verify this fixes the issue?

imnotjames (changes requested):
Can you create tests that verify this fixes the issue?

imnotjames:

PR #5147, opened 2019-11-26
feat: add migrations transaction option to connection options

mightyYaroslav:
It is the extension of pull request https://github.com/typeorm/typeorm/pull/4629 which allows
 applying same transaction options when running transactions in automatic mode


src/connection/BaseConnectionOptions.ts
```diff
@@ -47,6 +47,11 @@ export interface BaseConnectionOptions {
      */
     readonly migrationsTableName?: string;
 
+    /**
+     * Transaction mode for migrations to run in
+     */
+    readonly migrationsTransaction?: "all" | "none" | "each";
```
pleerock: maybe rename it so it can be clear that its a "mode" and not a "transaction"?

mightyYaroslav: Yep, it is reasonable. Updated the naming

pleerock:
Great, thanks!

PR #5164, opened 2019-12-02
fix: use global promise name

### LINKED ISSUE #5156: async-listener breaks lazy property discovery
**Issue type:**
async listener (in this instance required in by aws library) overrides the global Promise with WrappedPromise which extends Promise.

in the ManyToOne relationship decorator this results in a failure to recognise the promise type of properties:
```ts
        if (!isLazy && Reflect && (Reflect as any).getMetadata) { // automatic determination
            const reflectedType = (Reflect as any).getMetadata("design:type", object, propertyName);
            if (reflectedType && typeof reflectedType.name === "string" && reflectedType.name.toLowerCase() === "promise")
                isLazy = true;
        }
```

[ ] question
[x] bug report
[ ] feature request
[ ] documentation issue

**Database system/driver:**

[x] `cordova`
[x] `mongodb`
[x] `mssql`
[x] `mysql` / `mariadb`
[x] `oracle`
[x] `postgres`
[x] `cockroachdb`
[x] `sqlite`
[x] `sqljs`
[x] `react-native`
[x] `expo`

**TypeORM version:**

[x] `latest`
[ ] `@next`

aws-xray-sdk-core:
  └─┬ aws-xray-sdk-core@2.2.0
    └─┬ continuation-local-storage@3.2.1
      └── async-listener@0.6.10  

create an entity with a lazy property. this will not be loaded

jclyons52:
when checking if a property is lazy, use the name of the current global
Promise, this way if it has been changed lazy properties will still be
identified

Closes #5156

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -181,6 +181,15 @@
         "@types/node": "*"
       }
     },
+    "@types/continuation-local-storage": {
+      "version": "3.2.2",
+      "resolved": "https://registry.npmjs.org/@types/continuation-local-storage/-/continuation-local-storage-3.2.2.tgz",
+      "integrity": "sha512-aItm+aYPJ4rT1cHmAxO+OdWjSviQ9iB5UKb5f0Uvgln0N4hS2mcDodHtPiqicYBXViUYhqyBjhA5uyOcT+S34Q==",
+      "dev": true,
+      "requires": {
+        "@types/node": "*"
+      }
+    },
     "@types/debug": {
       "version": "4.1.2",
       "resolved": "https://registry.npmjs.org/@types/debug/-/debug-4.1.2.tgz",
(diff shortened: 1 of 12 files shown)


pleerock:
don't really think we need all this trickiness only to cover such a rare use case considering lazy relations being an experimental feature. I suggest you simply specify your relation as lazy using `{ lazy: true }` in your relation options.

jclyons52:
one Library that causes the issue gets [~730,000 monthly downloads](https://www.npmjs.com/package/async-listener), and given that it's used in continuation local storage, (used in express apps) and AWS SDKs, it seems like there is a pretty high probability of it showing up in server side apps that use typeorm. Excluding the test and package.json code this PR changes 17 lines in the codebase, so the complexity cost doesn't seem that high. Maybe using global without a null check is a bit risky (I can add one if it helps) but aside from that it seems clear enough. I've already solved this issue in my code bases, so I'm not worried about a work around. This is just a particularly difficult issue to debug (library load order may be different under test, hiding the issue) and pretty high severity as relationships stop working. I'm trying to prevent others from having the same problem.

PR #5178, opened 2019-12-04
fix: use a prefix on SelectQueryBuilder internal parameters

### LINKED ISSUE #5174: `selectQueryBuilder.take` messes up the query when using the `ids` parameter
**Issue type:**

[ ] question
[x] bug report
[ ] feature request
[ ] documentation issue

**Database system/driver:**

[ ] `cordova`
[ ] `mongodb`
[ ] `mssql`
[ ] `mysql` / `mariadb`
[ ] `oracle`
[x] `postgres`
[ ] `cockroachdb`
[ ] `sqlite`
[ ] `sqljs`
[ ] `react-native`
[ ] `expo`

**TypeORM version:**

[ ] `latest`
[ ] `@next`
[x] `0.2.20`

This issue happens when I have something like:
```typescript
const query = this.createQueryBuilder('employee')
// ... complex joins and stuff...
query.where('task.id IN (:...ids)', { ids: [foo, bar, etc] })

const result = await query.take(5).getMany()
```
This causes TypeORM to bahave unexpectedly, loading an empty result.

I can see in the generated query that it substitutes by `:...ids` with the parameters `.take()` uses internally.

I consider it a bug because it can frustrate a lot of unsuspecting people. Maybe I would rename it to something more obscure to the end user.

cuchi:
Fixes #5174


pleerock:
If I remember it right oracle driver allows parameters only to start from a-z. So, I recommend to bring instead some non trivial name instead of using `_` sign.

pleerock:
Thanks!

PR #5195, opened 2019-12-06
added in memory caching option

CameronSima:
For testing or in cases when Redis is unsuitable/ not wanted, I think an in-memory cache option is a nice feature. Also, I don't quite see the utility of a db based cache which is the current default setting. This will allow caching without the need to hit db or implement Redis.

### DIFF
--- a/src/cache/InMemoryQueryResultCache.ts
+++ b/src/cache/InMemoryQueryResultCache.ts
@@ -1,6 +1,6 @@
 import {QueryResultCache} from "./QueryResultCache";
-import { QueryResultCacheOptions } from './QueryResultCacheOptions';
-import { QueryRunner } from '..';
+import { QueryResultCacheOptions } from "./QueryResultCacheOptions";
+import { QueryRunner } from "..";
 
 /**
  * Caches query result into simple in-memory map.
@@ -14,7 +14,7 @@ export class InMemoryQueryResultCache implements QueryResultCache {
     /**
      * Map to store cached values.
      */
-    protected cache: Record<string, QueryResultCacheOptions>
+    protected cache: Record<string, QueryResultCacheOptions>;
 
     // -------------------------------------------------------------------------
     // Constructor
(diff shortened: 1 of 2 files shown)


pleerock:
I'm not sure about this PR. Looks like we really don't need it. Because:

* there is a possibility of creating a custom cache types, without any changes in the core. That's what you already did. Anyone can just copy/paste this and use in their codebase. Or you can create an extension if you want =)
* it doesn't make sense in production. You most probably will have many node processes, load balancer and so on. In memory cache won't be effective or even can produce unexpected results 
* for simple and test cases we already have a database cache

Thanks :)

justblender:
> * there is a possibility of creating a custom cache types, without any changes in the core.

Sorry for bumping this issue, but there's no documention describing how to implement a custom caching solution and I have absolutely no clue how to integrate it with TypeORM without forking it, maybe you should allow users to pass an instance of a class that implements QueryResultCache to the 'cache' property of BaseConnectionOptions?

pleerock:
@justblender checkout the source code, take a look how current redis is implemented and you'll understand how to implement a custom caching.

ovr:
@pleerock 

> it doesn't make sense in production. You most probably will have many node processes, load balancer and so on. In memory cache won't be effective or even can produce unexpected results

IMHO: But memory adapter can be a great feature for:

- When you needed to cache data from db that are dictionary (languages/countries/cities/whatever else). This data can be changes once per year.

Will be cool if memory adapter will be built-in (better with LRU/lifetime support)

But: ORM doesnot provide region cache support, when you can cache some data in memory, another in redis by decorator as example

pleerock:
> IMHO: But memory adapter can be a great feature for

I can't imagine any serious production server running on a single process, and even if such production app exists, I don't believe it would need a caching

I'll merge this PR if it will get more votes from people (and more real-world motivation why they need it)

leandrogehlen:
@pleerock, Cordova apps nedds this feature

justblender:
@leandrogehlen You can set a custom cache provider starting from 0.2.22. Check out documentation on [caching](https://typeorm.io/#/caching) to get started. Sadly, there is no IMDB `QueryResultCache` implementation built-in, so you'll have to write your own one.

leandrogehlen:
@jussikinnula, I know and I am using my own implementation. But I think this feature should be in core framework

sumeet-bansal:
Would like for this to be merged as well.

imnotjames:

Issue #5192, opened 2019-12-06
QueryBuilder to compare ROW() sets


boenrobot:
**Issue type:**

[x] question
[ ] bug report
[x] feature request
[ ] documentation issue

**Database system/driver:**

[ ] `cordova`
[ ] `mongodb`
[ ] `mssql`
[x] `mysql` / `mariadb`
[ ] `oracle`
[ ] `postgres`
[ ] `cockroachdb`
[ ] `sqlite`
[ ] `sqljs`
[ ] `react-native`
[ ] `expo`

**TypeORM version:**

[x] `latest`
[ ] `@next`
[ ] `0.x.x` (or put your version here)

**Steps to reproduce or a small repository showing the problem:**

<!--
    To answer those questions you need to put "x" inside the square brackets, for example:
    [x] `mysql`
    [ ] `postgres`

    Also, please format your code properly (by taking code blocks into ```ts .... ```)
!>

In MySQL, I can compare a row against a set of values using MySQL's ROW() function, coupled with the IN operator.

f.e.

```
SELECT e.system_event_id FROM `system_events` `e` INNER JOIN `system_event_handlers` `h` ON e.system_event_id = h.system_event_id AND h.finished_at IS NOT NULL WHERE ROW(e.system_event_id, COUNT(*)) = (
('ab', 0),
('cd', 1),
('ef', 0)
) GROUP BY system_event_id)
```

will get me just the specified event IDs if those event IDs have the expected count of finished system_event_handler rows.

I have the list as a nested array in TS, i.e.

```
const events = [
  ['ab', 0],
  ['cd', 1],
  ['ef', 0]
];
```

But the question is how I can feed the list into TypeORM's query builder?

I tried

```
eventsRepo
          .createQueryBuilder('e')
          .select('e.system_event_id')
          .innerJoin(
            'system_event_handlers',
            'h',
            'e.system_event_id = h.system_event_id AND h.finished_at IS NOT NULL',
          )
          .groupBy('system_event_id')
          .where('ROW(e.system_event_id, COUNT(*)) = (:events)', {
            events,
          })
          .getQuery()
```

But that leaves the :events unparsed, resulting in a syntax error.

If this is a missing feature, consider this a feature request... But either way, any way of generating the top query?

boenrobot:
I ended up with what I think is an ugly workaround, though it does work...

[...]

PR #5196, opened 2019-12-06
feat: Added List find operator.

boenrobot:
See #5192.

This PR adds an operator to output an array as a comma separated value. Primary useful in query builder.

I haven't actually tested this in my own code, as I can't really run the tests locally, and in my app I have other dependencies that require an official TypeORM version.


pleerock:
find operators were not designed to work with query builder... For now my recommendation is to use separate parameters per each array item

PR #5202, opened 2019-12-08
Speed ​​up id search in buildChildrenEntityTree

SoraKumo001:
The processing speed improves when the number of data is large.


pleerock:
Thanks!

PR #5212, opened 2019-12-10
docs: use @RelationId instead of @Column to load relation id

CarstenWalther:


pleerock:
No, I can't recommend it. @Column is a natural and cost-free way while relation id is complex and has some additional computations applied.

PR #5217, opened 2019-12-11
Implement driver options for NativeScript

jerbob92:
This pull request implements the extra driver options for NativeScript and adds them to the driver specific `ConnectionOptions`, it also fixes passing options using the extra option from `BaseConnectionOptions`. This is required to use the Commercial version of the driver.


pleerock:
Thanks!

PR #5227, opened 2019-12-12
feat: hash aliases to avoid conflicts

papermana:
When selecting relations (e.g. in `find` methods), if a generated alias is longer than the maximum identifier length allowed by the current driver, it would get shortened: for example "Something_somethingElse" would become "Some_soEl" etc.

However, if multiple relations have similar enough aliases, their shortened versions might end up identical. This would lead to conflicts that cause the query to fail.

This commit hashes aliases that are too long, instead. This way there is still a theoretical possibility of a collision, but it's astronomically low. In practice, alias conflicts should disappear.

One drawback of this solution is that it leads to more vague aliases. This may make queries less inspectable. For what it's worth, though, `Some_soEl` doesn't grant a lot of clarity about what it represents, either.

BREAKING CHANGE: aliases for very long relation names may be replaced with hashed strings

---

This change only affects one place where identifiers are `shorten`ed. This solves the problem that I encountered personally. Perhaps it doesn't solve all problems from this category, but I don't understand the TypeORM codebase enough to want to change anything more than what I know is necessary.

I didn't open this PR in response to a specific ticket, but instead based on my own problem. But it appears like it might solve https://github.com/typeorm/typeorm/issues/3721 and https://github.com/typeorm/typeorm/issues/1969, maybe others as well.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "typeorm",
-  "version": "0.2.20",
+  "version": "0.2.21",
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
--- a/src/driver/DriverUtils.ts
+++ b/src/driver/DriverUtils.ts
@@ -1,5 +1,5 @@
 import { Driver } from "./Driver";
-import { shorten } from "../util/StringUtils";
+import { hash } from "../util/StringUtils";
 
     /**
  * Common driver utility functions.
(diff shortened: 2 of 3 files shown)


yzimmermann:
My Team would appreciate a merge. Would be nice to get a eta

pleerock:
Okay, let's merge it. 

The downside of this issue is critical for Oracle users, since in Oracle max alias name is super short and I guess this change would lead to hashes in most Oracle queries even if there are no relations at all. 

Maybe best fix is to have a map of aliases used per query builder to prevent duplication. Also maybe we can give ability to define this logic in a naming strategy can work better for some people.
Hope we can apply any of these solutions in the future.

jcarrillo7:
@pleerock  I think this also happens on table names. I have hit issues like 

table name "..." specified more than once

And have to fallback to query builder

PR #5231, opened 2019-12-15
fix(postgres): spatialFeatureType mapping

hoonoh:
Fixes incorrect spatial feature type mapping when there are multiple tables with different spatial types.

For example, when there are two tables `A` and `B` with columns `A.point` as `Point` and `B.polygon` ans `Polygon` type, both columns where recognized as `Point`.


imnotjames:
@hoonoh Can you add tests to this that confirm that these changes fix the offending behavior?

Is there a related github issue?

imnotjames:

PR #5236, opened 2019-12-17
Clarify example

maclanphier:
`name` and `id` are often reserved keys, so the example should be made less ambiguous.


pleerock:
no, there is no problem with id and name, and this example works without any issues, I don't see a reason for this change :)

pleerock:
id and name are used in many many places in the docs, if we change here... with the same logic we can change everywhere too. Ive got your concern but it can be applied to many places and it won't look good then. Maybe we can improve this, but not this way I think.

PR #5240, opened 2019-12-19
Add array type to CockroachDB default parser

sam3d:
This PR implements the following fix to the CockroachDB driver:

- The default array parser from the PostgresDriver has been copied over.

Previously, when the following column was provided:

```ts
@Column("varchar", { array: true, default: ["hello"] })
```

An error was thrown that the array type did not match, as arrays were not getting parsed. The alternative solution was to set the default to use PostgreSQL syntax:

```ts
@Column("varchar", { array: true, default: "{'hello'}" })
```

### DIFF
--- a/src/driver/cockroachdb/CockroachDriver.ts
+++ b/src/driver/cockroachdb/CockroachDriver.ts
@@ -479,7 +479,10 @@ export class CockroachDriver implements Driver {
         const defaultValue = columnMetadata.default;
         const arrayCast = columnMetadata.isArray ? `::${columnMetadata.type}[]` : "";
 
-        if (typeof defaultValue === "number") {
+        if (columnMetadata.isArray && Array.isArray(defaultValue)) {
+            return `'{${defaultValue.map((val: string) => `${val}`).join(",")}}'`;
+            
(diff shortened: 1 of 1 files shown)


src/driver/cockroachdb/CockroachDriver.ts
```diff
@@ -479,7 +479,10 @@ export class CockroachDriver implements Driver {
         const defaultValue = columnMetadata.default;
         const arrayCast = columnMetadata.isArray ? `::${columnMetadata.type}[]` : "";
 
-        if (typeof defaultValue === "number") {
+        if (columnMetadata.isArray && Array.isArray(defaultValue)) {
+            return `'{${defaultValue.map((val: string) => `${val}`).join(",")}}'`;
```
pleerock: looks like it force casts array values to string even if column isn't a string-typed. What it can lead to? Also, tests are missing for this feature :) Thanks

sam3d:
Ah good point, will get those sorted ASAP. Thanks!

sam3d:
Didn't have time to work on this, might come back to it later.

Issue #5249, opened 2019-12-24
Unable to resolve module `crypto` from `node_modules/typeorm/browser/util/StringUtils.js`


nochkin:
**Issue type:**

[ ] question
[x] bug report
[ ] feature request
[ ] documentation issue

**Database system/driver:**

[ ] `cordova`
[ ] `mongodb`
[ ] `mssql`
[ ] `mysql` / `mariadb`
[ ] `oracle`
[ ] `postgres`
[ ] `cockroachdb`
[ ] `sqlite`
[ ] `sqljs`
[x] `react-native`
[ ] `expo`

**TypeORM version:**

[ ] `latest`
[ ] `@next`
[x] `0.2.22` (this is the `latest` at this moment)

When using with React Native I'm getting "Unable to resolve module `crypto`" error when trying to use TypeORM.
The issue appears in the 0.2.22 only. This is due to the following recent change:
https://github.com/typeorm/typeorm/commit/edc8e6dfd2c250656ab5b143297c60f57787ce77#diff-3fbb9e2b917220327e7c2fe6522c9b6d

**Steps to reproduce or a small repository showing the problem:**

```react-native init TypeormHash```
```cd TypeormHash```
```Add "import {Entity} from 'typeorm/browser';" at the top of App.js```
```react-native run-android```

nochkin:
I guess this is open for discussions as I'm not sure this is the right approach.
My own fix is to use sha.js instead of "crypto" which is not available in RN:
https://github.com/nochkin/typeorm/commit/ba160304f89e8a533b075d4af9961066a09bd05a

Kononnable:
It looks like a regression, so yes it is acceptable :)
Although this library was last updated 2 years ago, so there might be better alternatives.

@papermana There was any particular reason to use crypto, or just because it was available without need to download more dependencies.

papermana:
@Kononnable No particular reason. As you say, it was simply there (or so I assumed). Any library that produces a SHA256 hash (or maybe a different one as well) should be ok.

nochkin:
Thanks for confirmation.
The [sha.js](https://www.npmjs.com/package/sha.js) has over 3M weekly downloads according to npmjs. This was my main concerns when I picked up. Others were not so popular.
I'll create PR now.

Kononnable:
My bad, I was having internet troubles and download graph wasn't showing properly.  Looks like it isn't updated frequently, because it is pretty stable.

sallar:
[...]

PR #5270, opened 2019-12-31
Use sha.js instead of crypto for hash calculation

nochkin:
Initial discussion @ https://github.com/typeorm/typeorm/issues/5249

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "typeorm",
-  "version": "0.2.21",
+  "version": "0.2.22",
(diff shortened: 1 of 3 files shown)


pleerock:
Thanks guys!

thynson:
Why not use isomorphic-crypto instead, which still use native SHA implementation on node environment.

nochkin:
@thynson,
That will probably break using TypeORM in React-Native (this is how I stumbled onto this issue initially).
At least, the "sha.js" is more universal solution as it's available in other worlds too.

PR #5273, opened 2020-01-01
chore: update node versions tested on travis

Kononnable:
Update node.js versions tested on travis:
- removed nodejs 8.x - EOL
- added 13.x - current
- removed @sap/hdbext from dev dependencies

Looks like @sap/hdbext is not compatible with node 13. I think we should deal with this same way we're currently handling oracledb. It doesn't work in every environment(32 bit), so developer need to install it manually if he wants to use it.

### DIFF
--- a/.npmrc
+++ b/.npmrc
@@ -1,1 +1,0 @@
-@sap:registry=https://npm.sap.com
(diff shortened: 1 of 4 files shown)


pleerock:
Got you, thanks!

PR #5276, opened 2020-01-01
Fix: Enums with spaces are not converted properly

duckies:
Postgres returns values for enum arrays in formats like `{"Guild Master", Recruit}` and are incorrectly converted to `['"Guild Master"', 'Recruit' ]`.

Potential issues arise depending on the fix implementation if the user wanted quotes in their enum. I utilized a dirty patch that removes the first and last characters from the string and determines if it's a valid enum.

- [x] Functionality testing using arrays.
- [x] Functionality testing without arrays.


RobertWHurst:
My team just ran into this issue. It's a pretty crucial bug, would be really nice to see this fixed. Our product does hotel bookings, and this caused some serious issues for us.

src/driver/postgres/PostgresDriver.ts
```diff
@@ -537,12 +537,20 @@ export class PostgresDriver implements Driver {
             if (columnMetadata.isArray) {
                 // manually convert enum array to array of values (pg does not support, see https://github.com/brianc/node-pg-types/issues/56)
                 value = value !== "{}" ? (value as string).substr(1, (value as string).length - 2).split(",") : [];
-                // convert to number if that exists in poosible enum options
+                // Remove extraneous delimiting quotes around enum strings with spaces.
+                value = value.map((val: string) => {
+                    // Returns an empty string if the length is negative.
+                    const stripped = val.substr(1, val.length - 2);
```
pleerock: This code isn't obvious to me. As I understood you are trying to remove quotes around enum strings by this code. But what if there are no quotes around? Wouldn't it remove regular characters from it?

javierlinked:
We are facing the same issue. And waiting this fix to be included into main branch. Thanks.

src/driver/postgres/PostgresDriver.ts
imnotjames: Seconding this.

imnotjames (commented):
Souls this be changed to make it more general?  Does this issue apply to other drivers as well or only Postgres?  What about cockroach?

duckies:
I personally no longer use TypeORM, so if a more generalized solution is desired feel free to close this.

PR #5293, opened 2020-01-06
chore: MSSQL, Cockroachdb, Oracle tests on CI

Kononnable:
Enables mssql, cockroachdb and oracle tests on CI. CircleCI will finally test all of db tested on travis(plus oracle).
Travis still is used for:
- checking different nodejs versions(can be moved but probably will make CI tests slower - we will hit free parallel job limit)
- code coverage(easy to change)

Updated mssql image - it contains latest security patches and is much more resource friendly(no longer needs 3.25GB RAM to start).
Cockroachdb tests are running on separate jobs - cockroachdb eats too much resources which caused tests to crash due to insufficient RAM.
Oracle builds will be skipped on PRs from outside of typeorm repository(to prevent stealing docker credentials).

There are few more ideas worth checking up in this area, but I think there are other priorities right now:
- Try to configure cockroachdb to use less resources
	- check if we're not dropping schema and syncing it again two times for some tests(it forces cockroach to eat more resources)
- Check if hana can be configured to use less resources(this is currently a blocker for running it on CI)
- Test some non-node based environments(to avoid regression when e.g. using node core dependencies


pleerock:
yaaaaai, thanks @Kononnable !

PR #5309, opened 2020-01-09
feat: Add basic support for custom cache providers

justblender:
As of right now, TypeORM only allows to use built-in cache providers, this pull request adds an option to specify a custom cache provider (that should implement QueryResultCache spec) in the ConnectionOptions object:

```
await createConnection({
  // ...all of the required options for establishing connection

  cache: {
    cacheProvider(connection) {
      return new MockQueryResultCache(connection);
    }
  }
});
```

### DIFF
--- a/src/connection/BaseConnectionOptions.ts
+++ b/src/connection/BaseConnectionOptions.ts
@@ -126,7 +126,7 @@ export interface BaseConnectionOptions {
         /**
          * Factory function for custom cache providers that implement QueryResultCache.
          */
-        readonly cacheProvider?: (connection: Connection) => QueryResultCache;
+        readonly provider?: (connection: Connection) => QueryResultCache;
 
         /**
          * Configurable table name for "database" type cache.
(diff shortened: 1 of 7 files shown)


src/connection/BaseConnectionOptions.ts
```diff
@@ -121,6 +123,11 @@ export interface BaseConnectionOptions {
          */
         readonly type?: "database" | "redis" | "ioredis" | "ioredis/cluster"; // todo: add mongodb and other cache providers as well in the future
 
+        /**
+         * Factory function for custom cache providers that implement QueryResultCache.
+         */
+        readonly cacheProvider?: (connection: Connection) => QueryResultCache;
```
pleerock: rename it to just "provider". We already have a "cache" word in a parent.

pleerock:
Thanks for contribution! Can you please add some test that will check this functionality and docs to describe it for others

justblender:
Smh, made a typo in the commit name. 

@pleerock, I've added information in the docs and renamed the property, but I'm not used to write the tests. 😞

pleerock:
Its never too late to learn :) Just copy paste exist test, replace few lines of code and it work.

pleerock:
Thanks!

PR #5318, opened 2020-01-11
Added note about using CreateDateColumn and UpdateDateColumn decorators

NiKlimenko:
Motivation: it may not be obvious to beginners that `INSERT` script will use the DEFAULT value, and the default behavior must be set for the column either in the decorator options or directly in migration script.

### DIFF
--- a/docs/decorator-reference.md
+++ b/docs/decorator-reference.md
@@ -280,7 +280,8 @@ Learn more about [MongoDB](mongodb.md).
 #### `@CreateDateColumn`
 
 Special column that is automatically set to the entity's insertion time.
-You don't need to write a value into this column - it will be automatically set.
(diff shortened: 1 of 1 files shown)


pleerock:
Based on your comment it feels like user needs to perform additional steps to make it to work (like "do I need to manually set the DEFAULT to this column or what?"). But reality is that ORM manages that for you if you have schema sync enabled.

NiKlimenko:
Yes, if you are using a migration scripts (production ready solution), you also need to set the `default ' column property. I think that the documentation should take into account migration-first examples.
`{
    name: 'updatedDate',
    type: 'timestamptz',
    default: 'now()',
  }`

pleerock:
I don't feel like this is an appropriate place for your note... Maybe we need to create a separate docs, but this needs more work. Thanks for your efforts!

PR #5336, opened 2020-01-13
Fixed tests and related functionality on next

Benjamin-Dobell:
Fixed tests and any corresponding broken functionality on the `next` branch.

Most things seem to have been broken as a result of merging master back into `next`, some recent, some issues were introduced a while back.

### DIFF
--- a/src/find-options/FindOptions.ts
+++ b/src/find-options/FindOptions.ts
@@ -206,7 +206,7 @@ export type FindOptions<E> = {
     /**
      * Enables or disables query result caching.
      */
-    lock?: { mode: "optimistic", version: number|Date } | { mode: "pessimistic_read"|"pessimistic_write" };
+    lock?: { mode: "optimistic", version: number|Date } | { mode: "pessimistic_read"|"pessimistic_write"|"dirty_read" };
 
     /**
      * Offset (paginated) where from entities should be taken.
--- a/src/metadata/ColumnMetadata.ts
+++ b/src/metadata/ColumnMetadata.ts
@@ -10,6 +10,7 @@ import {MongoDriver} from "../driver/mongodb/MongoDriver";
 import {PromiseUtils} from "../util/PromiseUtils";
 import {FindOperator} from "../find-options/FindOperator";
 import {ApplyValueTransformers} from "../util/ApplyValueTransformers";
+import {CockroachDriver} from "../driver/cockroachdb/CockroachDriver";
 
 /**
  * This metadata contains all information about entity's column.
@@ -656,10 +657,21 @@ export class ColumnMetadata {
      * Compares given entity's column value with a given value.
      */
     compareEntityValue(entity: any, valueToCompareWith: any) {
-        const columnValue = this.getEntityValue(entity);
+        let columnValue = this.getEntityValue(entity);
+
+        if (this.entityMetadata.connection.driver instanceof CockroachDriver && this.entityMetadata.connection.driver.normalizeType({type: this.type}) === "int8") {
+            if (typeof columnValue === "number") {
+                columnValue = columnValue.toString();
+            }
+            if (typeof valueToCompareWith === "number") {
+                valueToCompareWith = valueToCompareWith.toString();
+            }
+        }
+
         if (columnValue instanceof Object) {
             return columnValue.equals(valueToCompareWith);
         }
+
         return columnValue === valueToCompareWith;
     }
 
--- a/src/query-builder/QueryExpressionMap.ts
+++ b/src/query-builder/QueryExpressionMap.ts
@@ -434,6 +434,7 @@ export class QueryExpressionMap {
         map.callObservers = this.callObservers;
         map.useTransaction = this.useTransaction;
         map.nativeParameters = Object.assign({}, this.nativeParameters);
+        map.eagerRelations = this.eagerRelations;
         return map;
     }
 
--- a/src/query-builder/SelectQueryBuilder.ts
+++ b/src/query-builder/SelectQueryBuilder.ts
@@ -1105,7 +1105,6 @@ export class SelectQueryBuilder<Entity> extends QueryBuilder<Entity> implements 
         const queryRunner = this.obtainQueryRunner();
         let transactionStartedByUs: boolean = false;
         try {
-
             // start transaction if it was enabled
             if (this.expressionMap.useTransaction === true && queryRunner.isTransactionActive === false) {
                 await queryRunner.startTransaction();
@@ -1113,7 +1112,7 @@ export class SelectQueryBuilder<Entity> extends QueryBuilder<Entity> implements 
             }
 
             this.expressionMap.queryEntity = true;
-            this.applyFindOptions();
+
             const results = await this.executeEntitiesAndRawResults(queryRunner);
 
             // close transaction if we started it
@@ -1145,6 +1144,8 @@ export class SelectQueryBuilder<Entity> extends QueryBuilder<Entity> implements 
      * Gets single entity returned by execution of generated query builder sql.
      */
     async getOne(): Promise<Entity|undefined> {
+        this.applyFindOptions();
+
         const results = await this.getRawAndEntities();
         const result = results.entities[0] as any;
 
@@ -1170,6 +1171,8 @@ export class SelectQueryBuilder<Entity> extends QueryBuilder<Entity> implements 
      * Gets entities returned by execution of generated query builder sql.
      */
     async getMany(): Promise<Entity[]> {
+        this.applyFindOptions();
+
         if (this.expressionMap.lockMode === "optimistic")
             throw new OptimisticLockCanNotBeUsedError();
 
@@ -1182,6 +1185,9 @@ export class SelectQueryBuilder<Entity> extends QueryBuilder<Entity> implements 
      * Count excludes all limitations set by setFirstResult and setMaxResults methods call.
      */
     async getCount(): Promise<number> {
+
+        this.applyFindOptions();
+
         if (this.expressionMap.lockMode === "optimistic")
             throw new OptimisticLockCanNotBeUsedError();
 
@@ -1196,7 +1202,7 @@ export class SelectQueryBuilder<Entity> extends QueryBuilder<Entity> implements 
             }
 
             this.expressionMap.queryEntity = false;
-            this.applyFindOptions();
+
             const results = await this.executeCountQuery(queryRunner);
 
             // close transaction if we started it
@@ -1229,6 +1235,9 @@ export class SelectQueryBuilder<Entity> extends QueryBuilder<Entity> implements 
      * This method is useful to build pagination.
      */
     async getManyAndCount(): Promise<[Entity[], number]> {
+
+        this.applyFindOptions();
+
         if (this.expressionMap.lockMode === "optimistic")
             throw new OptimisticLockCanNotBeUsedError();
 
@@ -1242,7 +1251,6 @@ export class SelectQueryBuilder<Entity> extends QueryBuilder<Entity> implements 
                 transactionStartedByUs = true;
             }
 
-            this.applyFindOptions();
             this.expressionMap.queryEntity = true;
             const entitiesAndRaw = await this.executeEntitiesAndRawResults(queryRunner);
             this.expressionMap.queryEntity = false;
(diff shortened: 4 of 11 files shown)


Benjamin-Dobell:
_Dang!_ Tests related to ordering are failing on the CI, but not locally.

Admittedly, I haven't looked in details. However, just glancing at the order functionality, it doesn't really seem sound to me. ASC/DESC being provided as values in nested objects won't work for many DBs, as the order of sort expressions determines the final order:

e.g.

> When more than one expression is specified, the later values are used to sort rows that are equal according to the earlier values

https://www.postgresql.org/docs/9.5/queries-order.html

 Thee iteration order of keys in JavaScript objects isn't guaranteed, so the resultant general SQL has no guaranteed order.

Unless I'm missing something.


_Anyway_, even with that particular test failing, this PR is still a big improvement on `next` as without this PR the tests won't even _start_ running.

pleerock:
@Kononnable I think you need to take a look on this

Benjamin-Dobell:
@pleerock Any reason #5347 was merged over this whilst neither have been through review?

pleerock:
> @pleerock Any reason #5347 was merged over this whilst neither have been through review?

I'm sorry, but @Kononnable is now actively working on next branch because we want to merge it to master soon, and we had discussion about his changes (and maybe he already had them) before I (maybe we) see your changes.

Benjamin-Dobell:
@pleerock You commented on this pull request 3 hours before you merged #5347, so you definitely saw this pull request.

Let's do a quick comparison:

|                                                                                                                                                                                                             | This PR                          | #5347                       |   |   |
|-------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|----------------------------------|-----------------------------|---|---|
| PR Size                                                                                                                                                                                                     | **1,968**                        | 2,126                       |   |   |
| Excluding yarn.lock (%)                                                                                                                                                                                     | **100%**                         | 194%                        |   |   |
| Pull requested submitted                                                                                                                                                                                    | **First**                        | Last                        |   |   |
| Commits                                                                                                                                                                                                     | **9**                            | 4                           |   |   |
| Commit messages follow [project guidelines](https://github.com/typeorm/typeorm/blob/master/CONTRIBUTING.md#-commit-message-guidelines)                                                                      | **Yes**                          | No                          |   |   |
| Commits do one thing?                                                                                                                                                                                       | **Yes**                          | No                          |   |   |
| Unnecessary Whitespace Changes                                                                                                                                                                              | **No**                           | Yes                         |   |   |
| Unnecessary Refactoring                                                                                                                                                                                     | **No**                           | Yes                         |   |   |
| [Improper use of Utility Types](https://github.com/typeorm/typeorm/pull/5347/files#diff-7da0fc32ecaf4de82fdafaaac5d5c9aeR55-R63) [*](https://www.typescriptlang.org/docs/handbook/utility-types.html#picktk) | **No**                           | Yes                         |   |   |
| TODOs                                                                                                                                                                                                       | **Eliminates outstanding TODOs** | Introduced additional TODOs |   |   |
| **Tests Passed**                                                                                                                                                                                            | **2,663**                        | 15... Yes, just 15!            |   |   |

😖 

Is this pull request perfect? Absolutely _not!_ That's what reviews are for.

However, irrespective of whether this PR should have passed review, #5437 clearly shouldn't have.

Benjamin-Dobell:
Because this PR is correctly split up into atomic tasks, even if you didn't want all of it. You could have picked what you wanted and rebased #5437 on this. However, because #5437 conflates fixes with refactoring, there's no easy way to rebase this on #5437.

Now here we are two weeks after I submitted this PR and `next` is still passing a grand total of 15 tests. 🤦‍♂

pleerock:
@Benjamin-Dobell I'm sorry for confusion. I definitely see this PR, but I didn't review it and left it to @Kononnable . The reason I merged his PR because of changes we discussed with him weeks earlier.

> Is this pull request perfect? Absolutely not! That's what reviews are for.

@Kononnable is a second core contributor right now (after me) who works on `next` branch and he wants to push it to the master, it's not necessary for him to make a perfect PRs. I never do perfect PRs because I know what I'm pushing (of course it doesn't mean everything I do is right) 

Just relax, we can revert merged PR (if its still possible), and merge yours instead, or I can update your changes to make sure they fit well into his PR. I just need @Kononnable 's review of this PR.

src/metadata/ColumnMetadata.ts
```diff
@@ -656,10 +657,21 @@ export class ColumnMetadata {
      * Compares given entity's column value with a given value.
      */
     compareEntityValue(entity: any, valueToCompareWith: any) {
-        const columnValue = this.getEntityValue(entity);
+        let columnValue = this.getEntityValue(entity);
```
Kononnable: Not sure this is the proper way to handle this, however as long as we use it only in RelationIdLoader for comparing primary keys and return to the user fully transformed entities we should be ok.
In the future if we decide to use this function for something else we might have some problems. This lines will also be changed once we separate library core from driver implementations, but it's problem for another day.

Kononnable:
As for reverting #5347 and merging this one instead I'm ok with that, but there are two points worth taking care of it that's the case(it might be better to merge both PR after conflict resolution):
- cloning query builder with find options(not very nice code, but it works in mentioned PR)
- find options precedence over some QueryBuilder methods e.g. lock option set by find options wasn't overwritten by QueryBuilder setLock method(before #5347).

As for the 'comparison table' - I won't even comment on that. Unless you really don't know what's wrong with it and you choose you really need to know that.

pleerock:
Its not necessary to revert these changes, I can cherry pick changes from this branch. Let me just understand what we need and what we don't.

src/metadata/ColumnMetadata.ts
pleerock: yeah Im not sure about these changes too. Probably I'll need to check myself this issue.

src/find-options/FindOptions.ts
```diff
@@ -206,7 +206,7 @@ export type FindOptions<E> = {
     /**
      * Enables or disables query result caching.
      */
-    lock?: { mode: "optimistic", version: number|Date } | { mode: "pessimistic_read"|"pessimistic_write" };
+    lock?: { mode: "optimistic", version: number|Date } | { mode: "pessimistic_read"|"pessimistic_write"|"dirty_read" };
```
pleerock: this change -> looks like we can copy it over.

src/query-builder/SelectQueryBuilder.ts
```diff
@@ -1105,15 +1105,14 @@ export class SelectQueryBuilder<Entity> extends QueryBuilder<Entity> implements
...
-
             // start transaction if it was enabled
             if (this.expressionMap.useTransaction === true && queryRunner.isTransactionActive === false) {
                 await queryRunner.startTransaction();
                 transactionStartedByUs = true;
             }
 
             this.expressionMap.queryEntity = true;
-            this.applyFindOptions();
+
```
pleerock: how do I understand motivation behind this change?

src/query-builder/SelectQueryBuilder.ts
```diff
@@ -1929,6 +1937,16 @@ export class SelectQueryBuilder<Entity> extends QueryBuilder<Entity> implements
                 }
             }
 
+            if (this.findOptions.lock) {
```
pleerock: @Kononnable did you have this change already? I need to know if I should move this change or not.

test/functional/find-options/basic-usage/find-options-relations.ts
```diff
@@ -45,9 +44,6 @@ describe("find options > relations", () => {
 
     it("complex relation #1", () => Promise.all(connections.map(async connection => {
```
pleerock: can you please tell why previous version didn't work and new version works?

test/functional/repository/find-options/repository-find-options.ts
```diff
@@ -199,16 +199,16 @@ describe("repository > find options > cache", () => {
         user4.name = "Ginny";
         await connection.manager.save(user4);
 
-        // without cache it must return really how many there entities are
-        const users2 = await connection.getRepository(User).find();
-
-        expect(users2.length).to.be.equal(4);
-
-        // but with cache enabled it must not return newly inserted entity since cache is not expired yet
+        // with the cache enabled (and not expired) we expect our previously cached results i.e. not the newly inserted entity
```
pleerock: I guess I need explanation on changes in every test. Why it didn't work before, and how new changes fixed it

src/query-builder/QueryExpressionMap.ts
```diff
@@ -434,6 +434,7 @@ export class QueryExpressionMap {
         map.callObservers = this.callObservers;
         map.useTransaction = this.useTransaction;
         map.nativeParameters = Object.assign({}, this.nativeParameters);
+        map.eagerRelations = this.eagerRelations;
```
Kononnable: Not sure why this line is marked as new - it is already present in next branch(introduced by #5347)

src/query-builder/SelectQueryBuilder.ts
Kononnable: I think we should skip changes related to find options. Both approach looks like they're doing exactly same thing, but in a different way. 
As for this change it's probably because of this line https://github.com/Benjamin-Dobell/typeorm/blob/86134f7ab9e9cd07d33ac360d928236df77dd95c/src/query-builder/SelectQueryBuilder.ts#L1845 - we don't have metadata available in some contexts(migrations) and we're using getter witch throws an error instead of `hasMetadata` method(in current version of next `hasMetadata` is used).

src/query-builder/SelectQueryBuilder.ts
Kononnable: Yes, it's there.
https://github.com/typeorm/typeorm/blob/ba4bebe5d9045ec8655ca50e0e022fb606149f13/src/query-builder/SelectQueryBuilder.ts#L1985-L1991

src/query-builder/SelectQueryBuilder.ts
Kononnable: I think we should skip changes related to findOptions in QueryBuilder. Both approaches looks like they're doing same thing with a different way(apart from find options precedence overs query builder methods).
As for the change - it's probably because we don't have metadata available in all contexts(in migrations), and there is a check on `metadata` property(which throws an error if there is none) instead of `hasMetadata` method. Current version of next uses `hasMetadata` in code where similar problem happened.

test/functional/find-options/basic-usage/find-options-relations.ts
Benjamin-Dobell: The current ordering functionality is fundamentally broken. The reasoning is covered in a comment further up in this PR. This is an _attempt_ to work around that, however realistically the underlying functionality needs to be fixed.

However, a public facing API change is required, so I'm reluctant to make such a change myself.

test/functional/repository/find-options/repository-find-options.ts
Benjamin-Dobell: The reasoning is covered in the commit message for this change. Basically there's a potential timing issue any time the cache is involved as the cache expires after 1 second.

Previously the test went:

1. Insert
2. Verify we can retrieve from DB
3. Verify we can retrieve from cache

If step 2 takes too long the cache expires by the time we reach 3 (in practice this was occurring for me sporadically when running the tests in Docker on macOS i.e. a VM). So I've reordered the test to:

1. Insert
2. Verify we can retrieve from cache
3. Verify we can retrieve from DB

Whilst in _theory_ the cache can still expire, it's now exceedingly unlikely as we're using the cache immediately.

src/query-builder/SelectQueryBuilder.ts
Benjamin-Dobell: @pleerock To understand this change you should be looking at it in the context of the commit, which is why this PR is split up into several commits:

https://github.com/typeorm/typeorm/pull/5336/commits/aa023e47cd1437ea3be793481eb14f93b3437468

The commit message states that I've fixed lock mode. You'll then note that I've not simply deleted this line, but rather moved all calls to `applyFindOptions` to be _before_ any usages of `this.expressionMap.lockMode`.

Prior to this commit we were incorrectly checking the lock mode before the find options had been applied, thus it was effectively being ignored when the query builder was being used as part of a `find`.

src/metadata/ColumnMetadata.ts
Benjamin-Dobell: I wasn't thrilled about this change either. However, this was the most appropriate location I could find for the functionality, however I'm far from expert on typeorm's inner workings. Very happy to defer the fix for this one to either of you two.

src/query-builder/QueryExpressionMap.ts
neuro-sys: Because this branch needs to be rebased from the latest master, most likely.

imnotjames:
We've since merged master into `next` again.  There's some conflicts here.  Any chance you're wanting to get it resolved & can get it all reviewed up?

Issue #5245, opened 2019-12-21
Support for specific postgres range operators


H6LS1S:
**Issue type:**

[x] question
[ ] bug report
[ ] feature request
[ ] documentation issue

**Database system/driver:**

[ ] `cordova`
[ ] `mongodb`
[ ] `mssql`
[ ] `mysql` / `mariadb`
[ ] `oracle`
[x] `postgres`
[ ] `cockroachdb`
[ ] `sqlite`
[ ] `sqljs`
[ ] `react-native`
[ ] `expo`

**TypeORM version:**

[x] `latest`
[ ] `@next`
[ ] `0.x.x` (or put your version here)

When there will be support for operators to work with range / array / geometry and .etc? In my opinion, such operators should be included in the same level with operators less*, more*, equal, and so on, since range filtering ( range is contained by - `<@`, contains range - `@>`, overlap - `&&`, [more](https://www.postgresql.org/docs/current/functions-range.html) )  is ubiquitous. If it is already released, can I get documentation or examples ?

**Steps to reproduce or a small repository showing the problem:**
At the moment, I need to use such a crutch in a situation where the fields in the request may be optional and the external interface is different from the search interface. 
**!!!The approach described below is not clean and not suitable as a permanent solution.!!!**
```ts
import { Transform, Type } from 'class-transformer';
import { ApiProperty } from '@nestjs/swagger';
import { In, Raw} from 'typeorm';
import {
  IsUUID,
  IsOptional,
} from 'class-validator';

import { Range } from '../../../common/decorators/range.decorator';

export class GameFilter {
  @IsUUID()
  @IsOptional()
  @ApiProperty({
    required: false,
    example: 'b4e19ca1-48ea-482c-81ea-1f646d7f75d9',
  })
  public readonly id: string;

  @IsOptional()
  @Range(1, 40)
  @Transform(value =>
    Raw((alias: string) => `${alias} <@ '${JSON.stringify(value)}'`),
  )
  @ApiProperty({
    minimum: 1,
    maximum: 40,
    isArray: true,
    required: false,
    example: [1, 40],
    default: [1, 40],
    type: () => [Number, Number],
  })
  public readonly amount: [number, number];
```
Thanks in advance for reply

swim:
I'm keen on the overlaps op as well!

HELSIS666:
It is strange that in almost a month two people became interested in this problem.

imnotjames:
[...]

PR #5339, opened 2020-01-14
Feature/postgres overlaps operator

swim:
This PR addresses part of #5245 adding support for the OVERLAPS operator to QueryBuilder.

### DIFF
--- a/docs/select-query-builder.md
+++ b/docs/select-query-builder.md
@@ -356,6 +356,19 @@ Note: be careful with `orWhere` - if you use complex expressions with both `AND`
 keep in mind that they are stacked without any pretences.
 Sometimes you'll need to create a where string instead, and avoid using `orWhere`.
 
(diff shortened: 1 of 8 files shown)


pleerock:
wait.. no, no. we don't need it as part of QueryBuilder in a following manner. If we can add new find operators (like Not(), LessThan(), etc) then its okay we if came up with a good design.

swim:
Ah okay, I thought QueryBuilder would require an implementation first. The higher level abstraction via EntityManager would be next with a new operator.

imnotjames:
Until we can have driver-specific extensions of the querybuilder it's probably safer to use the escape hatches to accomplish driver specific SQL.

1. FOr `where` you can just as easily use `.where('field OVERLAPS :value', { value: "string" })`
2. For `FindOperators` you can use `Raw` - `const Overlaps = (pattern) => Raw(alias => `${alias} OVERLAPS :pattern`, { pattern })`

As such, I'm going to close out this PR.

PR #5343, opened 2020-01-14
fix: SAP HANA driver fixes

Szbuli:
SAP Hana driver creates incorrect foreign keys.


Szbuli:
We have found more bugs in the sap hana driver when trying to create migrations. I will add those fixes to this pull request soon.

pleerock:
Thanks for fixes! Are you using TypeORM with HANA? If yes, can you please provide feedback on it? 

Also, are you interested in improving HANA support in TypeORM? If yes, we really need some help to enable its test running on CI, because its super hard to review such PRs without knowing for sure it doesn't break exist functionality

Szbuli:
Prior to the official driver we were using our custom hana driver for typeorm. We are moving towards the official driver hence the fixes. 
We are using docker based hana express as well as hana services hosted on sap cloud platform.

I will try to look into the problem about the CI. I saw that you have tried, can you help me and explain what was the problem?

pleerock:
> Prior to the official driver we were using our custom hana driver for typeorm.

this is cool

> I will try to look into the problem about the CI. I saw that you have tried, can you help me and explain what was the problem?

yes, I had some strange issues on startup, I don't remember what exactly TBH. But I didn't spend too much time on trying to fix these issues. So, if you can help it would be great.

btw, are you satisfied with new driver? Is it same/better than what you have on your own?

Szbuli:
The problem with the CI was with a wrong port configuration for Hana. I dont know how can I start circleci tests.

In a previous pull request #5273 the hana driver was removed, how should the test run then?

pleerock:
probably we can manually add npm i hana-blabla into CI script, its not necessary for it to be in package.json

pleerock:
> I dont know how can I start circleci tests.

CI triggers on your branch when you push the changes

Szbuli:
Only travis ci run, circleci not

Szbuli:
I will not try to fix hana  with travis ci as #5293 states that it is not desired. But I still need some way to run circle ci on this PR.

Kononnable:
It is desired, but not sure if it is possible. There are limited resources on free CI and I couldn't make hana docker image use less resources, so it would be possible to run on our CI.

Szbuli:
Ok i got it now, the docker version you used is the smallest available, HANA uses a lot of memory.... On circle ci there is enough memory, but it is not enabled for "public" pull requests? 
The problem was that you used port 39015, but for the hana service only version port 39017 is open for connections.

Kononnable:
When I started hana docker image locally I ended up with it taking 4GB of ram, if I assigned less container wouldn't start. CircleCI standard(free tier) machines got 4GB of ram. Maybe there is a way to customize hana to use less memory on startup, but I couldn't make it happen.

Szbuli:
I don't think 4GB will work... I would say 6-8 minimum. The official minimum for express is 12GB and 16GB recommended.

pleerock:
@Szbuli I enabled CircleCI on PR branches... Can you please take a look now?

> I will not try to fix hana with travis ci as #5293 states that it is not desired. But I still need some way to run circle ci on this PR.

as @Kononnable said we desire to have them on both CIs if possible...

Szbuli:
Thanks for enabling circleCI. Is the $DOCKER_USER and  $DOCKER_PASSWORD env variables are removed from circleCI context? 

As I see now that there is only 4GB available, I am not sure if we can manage to make HANA work.

Kononnable:
Environment variables won't be set for builds outside of parent repository - security concerns. To test it with docker login you would have to enable circleCI on your forked repo and add variables there.

Szbuli:
As @Kononnable  said, the main problem is the memory. I may reduce it by a little with a custom docker image, if I disable some services, but 4GB will not be enough.

I can only run the tests locally if the max memory is 4GB.

pleerock:
If you think we cannot do anything, then revert your changes on whats not working right now and I'll merge rest of PR. Thanks for taking your time on research

Szbuli:
The currently used @sap/hdbext is not compatible with newer versions of node and is not giving any configuration options for the pool used. There is an open source pool library based on @sap/hana-client (which is also used by @sap/hdbext), with the open source [hdb-pool](https://github.com/ckyycc/hdb-pool/) there are config options for the pool min-max count, etc.
My proposal is to use the @sap/hana-client and hdb-pool combination vs the @sap/hdbext

We need the pool size options as there is currently a severe bug in the hana-client, which makes a stackoverflow error if there are more than 2 concurrent connections. We issued a ticket to SAP with the problem.
Also there are some project where we need 20-30 concurrent connections, with hdbext and hana-client we cannot set this.

pleerock:
@Szbuli I'm okay with these changes.

Kononnable:
TravisCI machines got 8GB of RAM. So hana tests might be able to work there. Not sure if we'll use it for long - there were plans to drop it in favor of circleci, but we may also keep it for testing multiple node versions etc. I guess if e.g. hana tests works only on travis, we could use both of them.

pleerock:
Since we don't pay for travis, maybe its okay to use it as a backup solution or solution that can run HANA for us.

pleerock:
Is it finished? We probably need to update docs and show what packages user need to install to make hana to work.

pleerock:
Thanks for your amazing work!

Szbuli:
Thank you for merging in the pull request! In the meantime we are trying to enable hana to run in travis ci. Looks like 8GB is still not enough to run the tests, they fail after a while. We will try to disable some services in Hana.

pleerock:
@Szbuli right now I have issue with connection to hana... If I remove this line from SapDriver: if (this.options.database) dbParams.databaseName = this.options.database; - it works. I'm trying to understand why I have this problem and hdb-pool doesn't seem to have this option (databaseName) in their docs

Szbuli:
Sorry I was very busy at the time, and forget to reply. Do you still need help?

PR #5345, opened 2020-01-14
fixing 'in' clause case for ORACLE

gerwinbrunner:
This is a fix for the Oracle database.

If you have a query like this:
```
const listOfIds = [];
DefinitionsEntityRepo.find({
  where: { id: In(listOfIds) },
})
```

it will result in the SQL
`SELECT "DefinitionsEntity"."id" AS "DefinitionsEntity_id" "DefinitionsEntity"."status" AS "DefinitionsEntity_status" FROM "KRI_DEFINITIONS" "DefinitionsEntity" WHERE "DefinitionsEntity"."id" IN ()
`

Then on a SQLlite DB the query will work just fine.

On Oracle it will raise an error:
error: { Error: ORA-00936: missing expression errorNum: 936, offset: 1432 }

The cause of this behavior is that oracle needs to handle an empty array differently.
This is how a working query for Oracle would look like:

`SELECT "DefinitionsEntity"."id" AS "DefinitionsEntity_id" "DefinitionsEntity"."status" AS "DefinitionsEntity_status" FROM "KRI_DEFINITIONS" "DefinitionsEntity" WHERE "DefinitionsEntity"."id" IN (null)
`

This PR fixes the emtpy array case so oracle queries will work and result in a correct return


mastilver:
I don't think it's an issue only for Oracle
We have the same issue with Postgres, we fixed it by changing it to:

```diff
-                return `${aliasPath} IN (${parameters.join(", ")})`;
+                return parameters.length > 0 ? `${aliasPath} IN (${parameters.join(", ")})` : 'FALSE';
```

Issue #5163, opened 2019-12-01
Add distinct for entity manager find options


ziofat:
**Issue type:**

[ ] question
[ ] bug report
[x] feature request
[ ] documentation issue

**Database system/driver:**

[ ] `cordova`
[ ] `mongodb`
[ ] `mssql`
[x] `mysql` / `mariadb`
[ ] `oracle`
[ ] `postgres`
[ ] `cockroachdb`
[ ] `sqlite`
[ ] `sqljs`
[ ] `react-native`
[ ] `expo`

**TypeORM version:**

[x] `latest`
[ ] `@next`
[ ] `0.x.x` (or put your version here)

**Steps to reproduce or a small repository showing the problem:**

It will be nice to have distinct option for select in find options when using entity manager.

<!--
    To answer those questions you need to put "x" inside the square brackets, for example:
    [x] `mysql`
    [ ] `postgres`

    Also, please format your code properly (by taking code blocks into ```ts .... ```)
!>

swim:
Agreed, exposing the distinct option to the find method would be super handy. However, there's a minor complication to this request. The find method maps to getMany which in turn sets the expression map query entity property to true. This changes the SQL statement by including the id in the selection which breaks the distinct result set.

PR #5349, opened 2020-01-14
Feature/find distinct

swim:
This PR addresses #5163

### DIFF
--- a/src/query-builder/SelectQueryBuilder.ts
+++ b/src/query-builder/SelectQueryBuilder.ts
@@ -1044,8 +1044,14 @@ export class SelectQueryBuilder<Entity> extends QueryBuilder<Entity> implements 
                 transactionStartedByUs = true;
             }
 
-            this.expressionMap.queryEntity = true;
+            if (!this.expressionMap.selectDistinct)
+                this.expressionMap.queryEntity = true;
+
             const results = await this.executeEntitiesAndRawResults(queryRunner);
+
+            // if entities is empty seed with raw
+            if (results.entities.length === 0)
+                results.entities = results.raw;
 
             // close transaction if we started it
             if (transactionStartedByUs) {
--- a/src/find-options/FindOneOptions.ts
+++ b/src/find-options/FindOneOptions.ts
@@ -13,6 +13,11 @@ export interface FindOneOptions<Entity = any> {
     select?: (keyof Entity)[];
 
     /**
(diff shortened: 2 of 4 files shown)


src/query-builder/SelectQueryBuilder.ts
```diff
@@ -1044,9 +1044,15 @@ export class SelectQueryBuilder<Entity> extends QueryBuilder<Entity> implements
                 transactionStartedByUs = true;
             }
 
-            this.expressionMap.queryEntity = true;
+            if (!this.expressionMap.selectDistinct)
+                this.expressionMap.queryEntity = true;
```
pleerock: can you please explain these lines? Its not obvious why we need to enable this flag

src/query-builder/SelectQueryBuilder.ts
```diff
@@ -1044,9 +1044,15 @@ export class SelectQueryBuilder<Entity> extends QueryBuilder<Entity> implements
...
+
             const results = await this.executeEntitiesAndRawResults(queryRunner);
 
+            // if entities is empty seed with raw
+            if (results.entities.length === 0)
+                results.entities = results.raw;
```
pleerock: also, why do we need this?

src/query-builder/SelectQueryBuilder.ts
swim: Previously this flag was always set as enabled in getRawAndEntities. When this flag is enabled the id of the selected table is added to the query. This means the selection is no longer distinct. Does this flag need to be enabled by default?

src/query-builder/SelectQueryBuilder.ts
swim: Ugh yes I should have raised this as a talking point / temporary amendment. This needs a better solution. The results object entities is empty without the queryEntity flag.

imnotjames: Can this only be done only when you've set distinct?

src/query-builder/SelectQueryBuilder.ts
imnotjames: If the Id could be omitted with select would it remove the need for these lines?

imnotjames (commented):
Can you rebase so the tests run?

imnotjames:
This has been stale for over 6 months, I'm going to be closing it out but the issue will remain.

PR #5361, opened 2020-01-15
fix: use OUTPUT INTO on SqlServer for returning columns

### LINKED ISSUE #5160: DML statement cannot have any enabled triggers if the statement contains an OUTPUT clause without INTO clause
**Issue type:**

[ ] question
[x] bug report
[ ] feature request
[ ] documentation issue

**Database system/driver:**

[ ] `cordova`
[ ] `mongodb`
[x] `mssql`
[ ] `mysql` / `mariadb`
[ ] `oracle`
[ ] `postgres`
[ ] `cockroachdb`
[ ] `sqlite`
[ ] `sqljs`
[ ] `react-native`
[ ] `expo`

**TypeORM version:**

[x] `latest`
[ ] `@next`

I have tried to insert Person into MSSQL table which contains one trigger, and I have run into this exception:

```
The target table \'Person.person\' of the DML statement cannot have any enabled triggers if the statement contains an OUTPUT clause without INTO clause.
```

Typeorm tried to execute this query:

```
exec sp_executesql @statement=N'INSERT INTO "Person"."person"("FirstName", "LastName", "Title") OUTPUT INSERTED."BusinessEntityID" VALUES (@0, @1, @2)',@params=N'@0 nvarchar(50), @1 nvarchar(50), @2 nvarchar(8)',@0=N'Alija',@1=N'Bobija',@2=N'Mr.'
```

I have used database `AdventureWorks2017` that is open source and available on Microsoft online websites.

This is my `Person` entity:

```ts
@Entity({
    schema: 'Person'
})
export default class Person extends BaseEntity {
    @PrimaryGeneratedColumn({
        name: 'BusinessEntityID'
    })
    Id!: number;

    @Column({
        length: 50
    })
    FirstName! : string;

    @Column({
        length: 50
    })
    LastName!: string;

    @Column({
        length: 8
    })
    Title!: string;
}
```

And this is my code for adding new person:

```ts
    await Person.insert({
        FirstName: 'Alija',
        LastName: 'Bobija',
        Title: 'Mr.'
    });
```

Best regards

sgarner:
Fixes #5160

When inserting or updating an entity, `QueryBuilder` was using SQL Server's `OUTPUT` clause to return updated values for columns with generated or default values and then update the entity representation.

This breaks when the target table has a trigger defined, with the following error:

```
The target table '...' of the DML statement cannot have any enabled triggers
if the statement contains an OUTPUT clause without INTO clause.
```

This PR changes the behavior of `QueryBuilder` (when connected to SQL Server) to use `OUTPUT INTO` instead of just `OUTPUT`, so the returned values are inserted to a table variable and then selected back, rather than returned directly from the `INSERT` statement.

SQL Server reference: https://docs.microsoft.com/en-us/sql/t-sql/queries/output-clause-transact-sql?view=sql-server-ver15#inserting-data-returned-from-an-output-clause-into-a-table

### DIFF
--- a/src/driver/sqlserver/SqlServerDriver.ts
+++ b/src/driver/sqlserver/SqlServerDriver.ts
@@ -700,6 +700,20 @@ export class SqlServerDriver implements Driver {
         }, {} as ObjectLiteral);
     }
 
+    buildTableVariableDeclaration(identifier: string, columns: ColumnMetadata[]): string {
+        const outputColumns = columns.map(column => {
+            return `${this.escape(column.databaseName)} ${this.createFullType(new TableColumn({
+                name: column.databaseName,
+                type: this.normalizeType(column),
+                length: column.length,
+                isNullable: column.isNullable,
+                isArray: column.isArray,
+            }))}`;
+        });
+
+        return `DECLARE ${identifier} TABLE (${outputColumns.join(", ")})`;
+    }
+
     // -------------------------------------------------------------------------
     // Protected Methods
     // -------------------------------------------------------------------------
--- a/src/driver/sqlserver/SqlServerQueryRunner.ts
+++ b/src/driver/sqlserver/SqlServerQueryRunner.ts
@@ -1370,6 +1370,11 @@ export class SqlServerQueryRunner extends BaseQueryRunner implements QueryRunner
                 return Promise.all(dropFkQueries.map(result => result["query"]).map(dropQuery => this.query(dropQuery)));
             }));
             await Promise.all(allTablesResults.map(tablesResult => {
+                if (tablesResult["TABLE_NAME"].startsWith("#")) {
+                    // don't try to drop temporary tables
+                    return;
+                }
+
                 const dropTableSql = `DROP TABLE "${tablesResult["TABLE_CATALOG"]}"."${tablesResult["TABLE_SCHEMA"]}"."${tablesResult["TABLE_NAME"]}"`;
                 return this.query(dropTableSql);
             }));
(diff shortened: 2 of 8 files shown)


pleerock:
sooo, looks like now it will always trigger 3 queries instead of one?

pleerock:
I don't really like it taking in mind how specific request is. Can we have a better solution?

sgarner:
> sooo, looks like now it will always trigger 3 queries instead of one?

Correct, when an insert or update query needs to return generated columns then a batch will be executed that looks like:

```sql
DECLARE @OutputTable (columns...);
INSERT INTO table (columns...) OUTPUT INSERTED.generatedCol1, INSERTED.generatedCol2... INTO @OutputTable VALUES (values...);
SELECT * FROM @OutputTable;
```

As this is still a single request and the extra 2 queries only operate on variables I don't believe this will have any performance impact. This solution is based on examples recommended in the SQL Server documentation.

> I don't really like it taking in mind how specific request is. Can we have a better solution?

Can you explain more what you don't like?

pleerock:
Okay, let's merge it. PR looks good.

sgarner:
Thank you @pleerock 😀

nebkat:
See linked issue/PR above (#7100, #7101), this completely broke `QueryBuilder.returning()` for custom columns. It can be fixed with some further integration of the specific functionality with `QueryBuilder` but I'm not sure that's the path that should be taken as this is already very different from any other driver.

---

I think this should either be reverted and stick to the previous behavior, and user can manually deal with the situations where there are triggers, or we should create handlers inside `QueryRunner` for `INSERT`/`UPDATE`/`DELETE` where highly DB specific stuff like this can be appended.

The latter option could work quite nicely if we progressed from raw string queries to objects, e.g:

```typescript
    getQuery() {
        return {
            comment: this.createComment(),
            update: this.createUpdateExpression(),
            returning: this.createReturningExpression()
            orderBy: this.createOrderByExpression(),
            limit: this.createLimitExpression()
        }
    }
```

Then `QueryRunner.insertQuery()` would form the database specific query as needed.

PR #5380, opened 2020-01-19
feat: add fulltext parser option

### LINKED ISSUE #3777: MySQL Fulltext index with ngram parser
**Issue type:**

[ ] question
[ ] bug report
[x] feature request
[ ] documentation issue

**Database system/driver:**

[ ] `cordova`
[ ] `mongodb`
[ ] `mssql`
[x] `mysql` / `mariadb`
[ ] `oracle`
[ ] `postgres`
[ ] `sqlite`
[ ] `sqljs`
[ ] `react-native`
[ ] `expo`

**TypeORM version:**

[x] `latest`
[ ] `@next`

We're able to create full text indexes in MySQL like so:
```
@Index(["name"], { fulltext: true })
```
This generates SQL like so:
```
CREATE FULLTEXT INDEX `name` ON `table`(`name`)
```

However, MySQL allows you to also specify a parser for fulltext indexes, specifically `ngram`. This lets you perform fuzzy search.
```
CREATE FULLTEXT INDEX `name` ON `table`(`name`) WITH PARSER ngram
```

It would be great if we could have an additional index option for this. I'm thinking this could be something like
```
@Index(["name"], { fulltext: true, ngram: true })
```
or
```
@Index(["name"], { fulltext: true, parser: 'ngram'|'mecab'|null })
```

That way, we can still use generated migrations and schema sync without losing specific parsers.

bizen241:
This adds support for parser option for fulltext index.

[MySQL > ngram Full-Text Parser](https://dev.mysql.com/doc/refman/8.0/en/fulltext-search-ngram.html)
[MySQL > Full-Text Parser Plugins](https://dev.mysql.com/doc/refman/8.0/en/plugin-types.html#full-text-plugin-type)

Closes #3777

### DIFF
--- a/src/decorator/Index.ts
+++ b/src/decorator/Index.ts
@@ -77,6 +77,7 @@ export function Index(nameOrFieldsOrOptions?: string|string[]|((object: any) => 
             unique: options && options.unique ? true : false,
             spatial: options && options.spatial ? true : false,
             fulltext: options && options.fulltext ? true : false,
+            parser: options ? options.parser : undefined,
             sparse: options && options.sparse ? true : false,
             background: options && options.background ? true : false,
             expireAfterSeconds: options ? options.expireAfterSeconds : undefined
--- a/src/decorator/options/IndexOptions.ts
+++ b/src/decorator/options/IndexOptions.ts
@@ -21,6 +21,12 @@ export interface IndexOptions {
     fulltext?: boolean;
 
     /**
+     * Fulltext parser.
+     * Works only in MySQL.
+     */
+    parser?: string;
+
+    /**
      * Index filter condition.
      */
     where?: string;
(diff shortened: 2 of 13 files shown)


pleerock:
Thanks for contribution!

We need to read information about parser being used on a table, store this information about compare with what we have in entity to make a proper updates on a schema changes. Looks like there is no such code. Try to apply parser on exist index, or remove it from exist index, ORM won't update a schema and this is bug.

You need to modify code on line 1410...

bizen241:
To synchronize parser, I tried using `COMMENT` clause.

pleerock:
sorry, we can't use "comment" clause. What if user already using COMMENT for different purposes - problem. What if user don't want to have dirty comments - problem? What if we'll have some other property that will be relying on COMMENT - problem. We need to read information on parser being applied from INFORMATION_SCHEMA, other approaches are not acceptable.

bizen241:
Thank you for your reviewing.

I can't read information about parser from `INFORMATION_SCHEMA`.
[MySQL Bugs > no metadata for parser name of fulltext index](https://bugs.mysql.com/bug.php?id=87974)

Sorry for lack of consideration.

pleerock:
If you are sure there is no way - please revert back "comment" changes, we'll just add support without schema sync working, its okay if its not supported by the database.

pleerock:
Thanks

PR #5391, opened 2020-01-22
.synchronize() drops json column on mariadb

### LINKED ISSUE #3636: synchronize drops (and then re-adds) json column in mariadb
**Issue type:**

[ ] question
[x] bug report
[ ] feature request
[ ] documentation issue

**Database system/driver:**

[ ] `cordova`
[ ] `mongodb`
[ ] `mssql`
[x] `mysql` / `mariadb`
[ ] `oracle`
[ ] `postgres`
[ ] `sqlite`
[ ] `sqljs`
[ ] `react-native`
[ ] `expo`

**TypeORM version:**

[ ] `latest`
[ ] `@next`
[0.2.13] `0.x.x` (or put your version here)

I created and entity with a json column. Here is part of the entity:

```
@Entity()
export class Sport {

  id: string;

  @Column({nullable: false})
  name: string;

  // columns skipped

  @Column('json', {
    nullable: true,
    default: '{}',
  })
  lexicon: object;

}
```
I'll note that after creation, the "lexicon" column in the database is typed longtext.
Whenever synchronize is run, I see:
```
query: ALTER TABLE `sport` DROP COLUMN `lexicon`
query: ALTER TABLE `sport` ADD `lexicon` json NULL DEFAULT '{}'
```
Which, of course deletes any data I had in that column.

I'm using MariaDB 10.2.10. In fact I upgraded to this version today specifically because it has support for json columns.

Before today I was just using a varchar column rather than a json column and storing strings - but of course that means converting objects to strings and back.

creduo:
Fixes #3636 

https://mariadb.com/kb/en/json-data-type/

The JSON column type is implemented using LONGTEXT in MariaDB, so every time a table with JOSN column gets synchronized will drop the column and add it again. 

This PR could fix the issue

### DIFF
--- a/src/driver/mysql/MysqlDriver.ts
+++ b/src/driver/mysql/MysqlDriver.ts
@@ -537,6 +537,14 @@ export class MysqlDriver implements Driver {
         } else if (column.type === "uuid") {
             return "varchar";
 
+        } else if (column.type === "json" && this.options.type === "mariadb") {
+            /*
+             * MariaDB implements this as a LONGTEXT rather, as the JSON data type contradicts the SQL standard,
+             * and MariaDB's benchmarks indicate that performance is at least equivalent.
+             * @see https://mariadb.com/kb/en/json-data-type/
+             */
+            return "longtext";
+
         } else if (column.type === "simple-array" || column.type === "simple-json") {
             return "text";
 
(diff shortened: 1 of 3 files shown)


pleerock:
Thanks!

imnotjames:
Was it intentional for this to stop applying the `CHECK` constraint `JSON_VALID`?  The `JSON` alias applies `JSON_VALID` as a check constraint in mariadb.

> In order to ensure that a a valid json document is inserted, the JSON_VALID function can be used as a CHECK constraint. This constraint is automatically included for types using the JSON alias from MariaDB 10.4.3.

creduo:
@imnotjames Apparantly I missed that part. And yes, changing a LONGTEXT field to JSON type will cause the issue you mentioned. And I can't come up with a quick patch to prevent that, because any JSON type is actually LONGTEXT type after creation.  

However, the patch in  this PR only works after the field has been created, so, adding a JSON field will just work as **JSON** type with `JSON_VALID()` not a LONGTEXT type.

Issue #1583, opened 2018-02-12
Repository create returns an array at all times


crutchcorn:
**Issue type:**

[ ] question
[x] bug report
[ ] feature request
[ ] documentation issue

**Database system/driver:**

[ ] `cordova`
[ ] `mongodb`
[ ] `mssql`
[ ] `mysql` / `mariadb`
[ ] `oracle`
[x] `postgres`
[ ] `sqlite`
[ ] `sqljs`
[ ] `websql`

**TypeORM version:**

[ ] `latest`
[ ] `@next`
[x] `0.1.11` (or put your version here)

**Steps to reproduce or a small repository showing the problem:**

I have the following code:
```
  const updatedPost = postRepository.create({
    ...req.body,
    user: undefined,
    comments: undefined,
    upvotes: undefined
  });
  postRepository.updateById(req.post.id, updatedPost)
    .then((user) => {
      res.send(user);
    })
```

However, when I try to compile this code, I get:
```
C:\Users\crutchcorn\git\Rucksack\rucksack-backend [mvp-routes ≡ +0 ~7 -0 !]
λ  npm start

> typeorm-typescript-express-example@0.0.1 start C:\Users\crutchcorn\git\Rucksack\rucksack-backend
> tsc && node src/index.js

src/modules/Posts/controllers/posts.controller.ts(37,42): error TS2559: Type 'Post[]' has no properties in common with type 'DeepPartial<Post>'.
npm ERR! code ELIFECYCLE
npm ERR! errno 2
npm ERR! typeorm-typescript-express-example@0.0.1 start: `tsc && node src/index.js`
npm ERR! Exit status 2
npm ERR!
npm ERR! Failed at the typeorm-typescript-express-example@0.0.1 start script.
npm ERR! This is probably not a problem with npm. There is likely additional logging output above.

npm ERR! A complete log of this run can be found in:
npm ERR!     C:\Users\crutchcorn\AppData\Roaming\npm-cache\_logs\2018-02-12T04_27_06_357Z-debug.log
```

I believe this code is because of the type definition for create:
```
    create(entityLikeArray: DeepPartial<Entity>[]): Entity[];
    create(entityLike: DeepPartial<Entity>): Entity;
```
But no matter how I change it I can't seem to get things working

pleerock:
[...]

PR #5400, opened 2020-01-22
fix(base-entity): set create return type to T[]

merceyz:
**What's the problem this PR addresses?**

The version of `BaseEntity.create` that takes an array of `entityLike` objects returns an array of entities but the return type incorrectly specifies that it's a single entity.

**How did you fix it?**

Changed the return type from `T` to `T[]`

### DIFF
--- a/src/repository/BaseEntity.ts
+++ b/src/repository/BaseEntity.ts
@@ -126,7 +126,7 @@ export class BaseEntity {
      * Creates a new entities and copies all entity properties from given objects into their new entities.
      * Note that it copies only properties that present in entity schema.
      */
-    static create<T extends BaseEntity>(this: ObjectType<T>, entityLikeArray: DeepPartial<T>[]): T;
+    static create<T extends BaseEntity>(this: ObjectType<T>, entityLikeArray: DeepPartial<T>[]): T[];
 
     /**
      * Creates a new entity instance and copies all entity properties from this object into a new entity.


pleerock:
Thanks!

jalooc:
FYI this broke existing code in a way described in this issue: https://github.com/typeorm/typeorm/issues/1583#issuecomment-596021889

tl;dr; this code:
```ts
const foo = Foo.create({ ...req.body }) // req.body is `any`
console.log(foo.bar)
```
suddenly breaks, because before the update `foo` used to be of `Foo` type and after the update it becomes `Foo[]` (note the array).

This might be TS fault instead TypeORM's fault, I'm not sure. But since it's breaking working code, it definitely requires a second look.

sherbakov1:
Same for me. This broke the existing code.

Diluka:
I thought it was TypeScript(https://github.com/microsoft/TypeScript/issues/37471), but turned out to be this.

itsjbecks:
Has there been any update on this / more background on why this was changed. It also broke existing code on our project.

Diluka:
@pleerock look here. this pr introduce a breaking change

pleerock:
Type signature seems to be right. It expects an array and we get array back. I don't see how it can be wrong right now, can somebody point me what can be wrong ?

jalooc:
@pleerock The code looks good, but its usage is crooked. The problem is that when using object spread operator in the `create({ ...x  })` argument, TS somehow thinks (probably because of some TS error) that [this](https://github.com/typeorm/typeorm/blob/07a8aead6e0a23f9d82f9a1c37bc10417732ca6b/src/repository/BaseEntity.ts#L129) definition should be used, not [that](https://github.com/typeorm/typeorm/blob/07a8aead6e0a23f9d82f9a1c37bc10417732ca6b/src/repository/BaseEntity.ts#L135). Therefore doing something like this: `const y = Model.create({ ...y })` makes `y` of type `Model[]` instead of `Model`, which was not the case before this PR. So while what you wrote [here](https://github.com/typeorm/typeorm/issues/1583#issuecomment-365244154) seems totally logical, it does not work this way.

pleerock:
okay, but this PR definitely fixes issue. Not sure how can I help with particular this use case when you use `any` object.

PR #5406, opened 2020-01-23
rename a sequence related to generated primary key when a table is renamed

### LINKED ISSUE #4937: Typeorm does not rename sequence of PK on renaming of table
**Issue type:**

[ ] question
[x] bug report
[ ] feature request
[ ] documentation issue

**Database system/driver:**

[ ] `cordova`
[ ] `mongodb`
[ ] `mssql`
[ ] `mysql` / `mariadb`
[ ] `oracle`
[x] `postgres`
[ ] `cockroachdb`
[ ] `sqlite`
[ ] `sqljs`
[ ] `react-native`
[ ] `expo`

**TypeORM version:**

[x] `latest`
[ ] `@next`

1. Create `Table1` with primary key id as serial type
2. Run cli command
```
typeorm migration:create -n RenameTable1
```
3. Write on migration
```
queryRunner.renameTable('Table1', 'Table2')
```
4. Run cli command
```
typeorm migration:generate -n CheckActualShema
```
Expected result - new migration is not generated.
Actual result - migration is generated.
```
await queryRunner.query(
      `CREATE SEQUENCE "Table2_id_seq" OWNED BY "Table2"."id"`
    )
    await queryRunner.query(
      `ALTER TABLE "Table2" ALTER COLUMN "id" SET DEFAULT nextval('Table2_id_seq')`
    )
    await queryRunner.query(
      `ALTER TABLE "Table2" ALTER COLUMN "id" DROP DEFAULT`
    )
```

fullc0de:
fixes #4937

### DIFF
--- a/src/driver/postgres/PostgresQueryRunner.ts
+++ b/src/driver/postgres/PostgresQueryRunner.ts
@@ -438,6 +438,20 @@ export class PostgresQueryRunner extends BaseQueryRunner implements QueryRunner 
             downQueries.push(new Query(`ALTER TABLE ${this.escapePath(newTable)} RENAME CONSTRAINT "${newPkName}" TO "${oldPkName}"`));
         }
 
+        // rename sequences
+        newTable.columns.map(col => {
+            if (col.isGenerated) {
+                const seqName = this.buildSequenceName(oldTable, col.name, undefined, true, true);
+                const newSeqName = this.buildSequenceName(newTable, col.name, undefined, true, true);
+
+                const up = schemaName ? `ALTER SEQUENCE "${schemaName}"."${seqName}" RENAME TO "${newSeqName}"` : `ALTER SEQUENCE "${seqName}" RENAME TO "${newSeqName}"`;
+                const down = schemaName ? `ALTER SEQUENCE "${schemaName}"."${newSeqName}" RENAME TO "${seqName}"` : `ALTER SEQUENCE "${newSeqName}" RENAME TO "${seqName}"`;
+
+                upQueries.push(new Query(up));
+                downQueries.push(new Query(down));
+            }
+        });
+
         // rename unique constraints
         newTable.uniques.forEach(unique => {
             // build new constraint name
--- a/src/driver/postgres/PostgresQueryRunner.ts
+++ b/src/driver/postgres/PostgresQueryRunner.ts
@@ -440,7 +440,7 @@ export class PostgresQueryRunner extends BaseQueryRunner implements QueryRunner 
 
         // rename sequences
         newTable.columns.map(col => {
-            if (col.isGenerated) {
+            if (col.isGenerated && col.generationStrategy === "increment") {
                 const seqName = this.buildSequenceName(oldTable, col.name, undefined, true, true);
                 const newSeqName = this.buildSequenceName(newTable, col.name, undefined, true, true);
 
--- a/test/functional/query-runner/rename-table.ts
+++ b/test/functional/query-runner/rename-table.ts
@@ -36,18 +36,18 @@ describe("query runner > rename table", () => {
         table = await queryRunner.getTable("question");
         table!.should.be.exist;
 
-        await queryRunner.renameTable("question", "user");
-        table = await queryRunner.getTable("user");
+        await queryRunner.renameTable("question", "answer");
+        table = await queryRunner.getTable("answer");
         table!.should.be.exist;
 
         if (connection.driver instanceof PostgresDriver) {
-            let result = await queryRunner.query(`SELECT COUNT(*) FROM "pg_class" "c" WHERE "c"."relkind" = 'S' and "c"."relname" = 'user_id_seq'`);
+            let result = await queryRunner.query(`SELECT COUNT(*) FROM "pg_class" "c" WHERE "c"."relkind" = 'S' and "c"."relname" = 'answer_id_seq'`);
             result[0].count.should.be.equal("1");
         }
 
         await queryRunner.executeMemoryDownSql();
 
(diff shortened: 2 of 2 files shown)


neuro-sys:
Not completely related, but do you know why the following line is commented in master?

https://github.com/typeorm/typeorm/blob/6c6bde7ebdc0698a938bb0576ea588236f5cf84a/src/schema-builder/RdbmsSchemaBuilder.ts#L156-L173

fullc0de:
@neuro-sys Thank you for giving me a feedback. Unfortunately, I don't know why it doesn't relate #4937 and I have seen a post about why migration doesn't support renaming table before but I don't know why that issue relates to the issue where I am on. Could you give me a hint or any explanation about that?

neuro-sys:
@fullc0de 

> I have seen a post about why migration doesn't support renaming table before but I can't understand why that issue relates to the issue where I am on.

Okay, it is not related to your issue. I was only curious (I'm not a contributor or a maintainer of this project).

Looking into the code though, I see that the problem is with `OWNED BY` statement in the sequence creation:

https://github.com/typeorm/typeorm/blob/ec9034125db31c1800a5fe732f3588cb48c384a3/src/driver/postgres/PostgresQueryRunner.ts#L831-L832

So one option would be to drop `OWNED BY` in the creation of sequence there, so that `renameTables` would not require changing the DDL for sequence any more. But this may be dangerous if `OWNED BY` is critical, which I am not 100% sure. My guess is that it's not needed.

The other option is like you do here to recreate the sequences with the new table name, whih seems correct. However there's already inconsistency in the master about sequence creation. `renameColumn` does not use `OWNED BY` when re-creating the sequence:

https://github.com/typeorm/typeorm/blob/ec9034125db31c1800a5fe732f3588cb48c384a3/src/driver/postgres/PostgresQueryRunner.ts#L646-L647

Neither your version.

I would personally combine this into a function, and do not replicate the same code three times, and drop `OWNED BY` altogether (or use it everywhere).

fullc0de:
sorry, I don't understand why the code in `changeColumn` method relates to the issue. I can see an intent from the code that a column which has a sequential property needs to have its related sequence renamed. In this idea, I presumed that same scheme should be applied to `renameTable`.

As you know, PostgreSQL doesn't change sequences automatically when renaming a table. Despite it works like that, there is no issue from DBMS perspective. because `nextval('hello_id_seq'::regclass)` on the sequence column is used as a link between a table and a sequence even though a table name and a sequence name are mismatched. Technically, Unchanging sequence name is no matter when it comes to DB operations.

The problem I like to point out is the way of TypeORM migration to work. as you can see in #4937, we have to deal with additional migration right after performing `renamingTable` migration. That's about changing a name of a sequence. It looks like annoyed task.

I think there are two options TypeORM can choose as follows:

1) Don't deal with the mismatching.
2) Do renaming a name of a sequence along with its table's name.

I'm afraid that there might be some missed points that I had to know while I was doing the work. I'm looking forward to that this thread is good point to get to know that. :-)

Thanks @neuro-sys

neuro-sys:
> sorry, I don't understand why the code in changeColumn method relates to the issue.

You are right, it's not related.

There are no calls to `.renameTable()` in the codebase (other than in tests), so it has nothing to do with the issue here, my bad. 

Whether `renameTable()` should handle renaming of sequences or not, I don't know (what about all other objects linked to the table, should rename everything?). Maybe a separate call before/after `renameTable()` to `renameColumn()` or `changeColumn()` would handle the problem with that issue?

My knowledge is very limited, so I'll bail out from this issue.

fullc0de:
@neuro-sys 
I think some users may have their own workarounds including me. And, your suggestion can also be one of them as well. But, these are just workarounds not a solution. Almost workarounds should be altered to the right solution someday.

src/driver/postgres/PostgresQueryRunner.ts
```diff
@@ -438,6 +438,20 @@ export class PostgresQueryRunner extends BaseQueryRunner implements QueryRunner
             downQueries.push(new Query(`ALTER TABLE ${this.escapePath(newTable)} RENAME CONSTRAINT "${newPkName}" TO "${oldPkName}"`));
         }
 
+        // rename sequences
+        newTable.columns.map(col => {
+            if (col.isGenerated) {
```
AlexMesser: you need `&& col.generationStrategy !== "uuid"` here, because uuid columns does not have sequences

fullc0de: Thank you, @AlexMesser 
If that code is supposed to be only effective to "increment" strategy, wouldn't it be better `col.generationStrategy === "increment"`? We can avoid an extra condition of `undefined`.

AlexMesser: yes, for Postgres we can only check for `increment`

fullc0de:
I applied a feedback from @AlexMesser and also fixed test failure.
But, i encountered the following failure message during test.

```
  other issues > entity listeners must work in embeddeds as well
    ✓ getters and setters should work correctly

  embedded > embedded-many-to-many-case1
    owner side
Segmentation fault (core dumped)
npm ERR! Test failed.  See above for more details.
```

I guess it doesn't seem there is nothing to do with my job. If there is, please feel free to give an advice.

fullc0de:
@AlexMesser 
If there is no more to check, could you please let the PR go ahead at your earliest convenience?

AlexMesser:
@fullc0de sorry for delay. I added additional checks in the test and for some reason it fails on CI now, but it passes on my machine. I will merge this PR and fix it on master if error appears there. 

Thank you for contribution!

PR #5413, opened 2020-01-24
Add NotImplementedError for driver functions that do nothing

neuro-sys:
This is very minor, but recently bitten by this. I don't see any reason why unimplemented functions should return silently doing nothing.


pleerock:
This change can be breaking for users who rely on different drivers and were using methods even if they don't do anything (`getDatabases` for example). So, I would recommend to create this PR against `next` branch.

neuro-sys:
I see, makes sense. I rebased it on top of `next` branch.

Overall, this is intended to break people's builds to make them not use functions which don't do anything for their Driver. Since this is a breaking change, it makes sense that it should go into a major version release.

pleerock:
Great, thanks!

PR #5419, opened 2020-01-26
perf: Optimized version of EntityMetadata#compareIds() for the common case

njam:
We've found a TypeORM-based application spending most CPU time in `OrmUtils.deepCompare()`, called from `compareIds()`.

This PR implements a faster code path for the common case when the ID object has the type `{id: string}`. Our application's overall response time has improved from 1.2s to 0.4s!

Before:  
image

After:  
image

wdyt?

cc @fresswolf

### DIFF
--- a/src/metadata/EntityMetadata.ts
+++ b/src/metadata/EntityMetadata.ts
@@ -746,6 +746,16 @@ export class EntityMetadata {
         if (firstId === undefined || firstId === null || secondId === undefined || secondId === null)
             return false;
 
+        // Optimized version for the common case
+        if (
+          typeof firstId.id === "string" &&
+          typeof secondId.id === "string" &&
+          Object.keys(firstId).length === 1 &&
+          Object.keys(secondId).length === 1
+        ) {
+            return firstId.id === secondId.id;
+        }
+
         return OrmUtils.deepCompare(firstId, secondId);
     }
 
--- a/src/metadata/EntityMetadata.ts
+++ b/src/metadata/EntityMetadata.ts
@@ -613,7 +613,7 @@ export class EntityMetadata {
         const secondEntityIdMap = this.getEntityIdMap(secondEntity);
         if (!secondEntityIdMap) return false;
 
-        return EntityMetadata.compareIds(firstEntityIdMap, secondEntityIdMap);
+        return OrmUtils.compareIds(firstEntityIdMap, secondEntityIdMap);
     }
 
     /**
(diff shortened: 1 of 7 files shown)


src/metadata/EntityMetadata.ts
```diff
@@ -746,6 +746,14 @@ export class EntityMetadata {
         if (firstId === undefined || firstId === null || secondId === undefined || secondId === null)
             return false;
 
+        // Optimized version for the common case
+        if (
+          typeof firstId.id === "string" && Object.keys(firstId).length === 1 &&
```
merceyz: In the event `typeof secondId.id !== "string"` wouldn't it be faster to do the two `typeof` checks first, then `Object.keys`?

```js
if (
  typeof firstId.id === 'string' &&
  typeof secondId.id === 'string' &&
  Object.keys(firstId).length === 1 &&
  Object.keys(secondId).length === 1
) {
  return firstId.id === secondId.id;
}
```

njam: You mean because `typeof` should be faster than `Object.keys()` in general?

merceyz: Yes

njam: Agree, I've udpated the code, thx!

src/metadata/EntityMetadata.ts
```diff
@@ -746,6 +746,16 @@ export class EntityMetadata {
...
             return false;
 
+        // Optimized version for the common case
+        if (
+          typeof firstId.id === "string" &&
+          typeof secondId.id === "string" &&
+          Object.keys(firstId).length === 1 &&
+          Object.keys(secondId).length === 1
+        ) {
+            return firstId.id === secondId.id;
```
neuro-sys: Wouldn't it make sense to have this as the first check in `OrmUtils.deepCompare` to benefit from it in all places it's used, such as:

- https://github.com/typeorm/typeorm/blob/a4dec02cc59d3219a29c7be0322af2253e1452dc/src/persistence/subject-builder/OneToOneInverseSideSubjectBuilder.ts#L142
- https://github.com/typeorm/typeorm/blob/a4dec02cc59d3219a29c7be0322af2253e1452dc/src/persistence/subject-builder/OneToManySubjectBuilder.ts#L120

njam: Makes sense!

Should I create a helper function `OrmUtils.compareIds()` for this use case?

njam: I pushed a commit with `OrmUtils.compareIds()`

njam:
@pleerock this should be ready

It's a backwards compatible change, which should improve performance in many use cases.

src/util/OrmUtils.ts
```diff
@@ -121,6 +121,26 @@ export class OrmUtils {
...
+    /**
+     * Check if two entity-id-maps are the same
+     */
+    static compareIds(firstId: ObjectLiteral|undefined, secondId: ObjectLiteral|undefined) {
+        if (firstId === undefined || firstId === null || secondId === undefined || secondId === null)
+            return false;
+
+        // Optimized version for the common case
+        if (
+            typeof firstId.id === "string" &&
```
pleerock: but what if ids are numbers? wouldn't it be an optimized version for numeric ids too?

njam: @pleerock agreed, I've pushed a new version which works for "string" and "number". wdyt?

njam:
@pleerock the commit you added makes sense, thanks. Is there anything else to do?

njam:
@pleerock ping

pleerock:
Sorry for late merge and thanks for contribution!

njam:
> Sorry for late merge and thanks for contribution!

Thanks for the assistance on the PR!

Issue #4947, opened 2019-10-22
beforeUpdate subscriber entity argument is undefined


simpert:
**Issue type:**

[ ] question
[X] bug report
[ ] feature request
[ ] documentation issue

**Database system/driver:**

[ ] `cordova`
[ ] `mongodb`
[ ] `mssql`
[ ] `mysql` / `mariadb`
[ ] `oracle`
[ ] `postgres`
[ ] `cockroachdb`
[X] `sqlite`
[ ] `sqljs`
[ ] `react-native`
[ ] `expo`

**TypeORM version:**

[ ] `latest`
[ ] `@next`
[X] `typeorm@0.2.19` (or put your version here)

**Steps to reproduce or a small repository showing the problem:**

```
@EventSubscriber()
export class UpdateModifiedSubscriber implements EntitySubscriberInterface {
    beforeUpdate ( event: UpdateEvent<any> ) {
        if ( ( event.entity as IEntityBase ) != null )
            ( event.entity as IEntityBase ).dateModified = new Date();
    }

    beforeInsert ( event: InsertEvent<any> ) {
        if ( ( event.entity as IEntityBase ) != null )
            ( event.entity as IEntityBase ).dateModified = new Date();
    }
}

export interface IEntityBase {
    id?: string;
    dateModified?: Date;
}

export abstract class EntityBase<T> implements IEntityBase {
    constructor ();
    constructor ( entity: Partial<T> );
    constructor ( entity?: any ) {
        if ( !entity ) return;

        for ( const key in entity )
            this[ key ] = entity[ key ];
    }

    @PrimaryGeneratedColumn( 'uuid' )
    id: string;

    @Column( { type: dateTimeType(), nullable: true } )
    dateModified: Date;
}

@Entity( 'Player' )
export class Player extends EntityBase<IPlayer> implements IPlayer {
    @Column( { type: 'nvarchar', length: 20, nullable: false } )
    firstName: string;

    @Column( { type: 'nvarchar', length: 20, nullable: false } )
    lastName: string;

    @Column( { type: 'integer', nullable: false } )
    jerseyNumber: number;

    @Column( { type: 'integer', nullable: false } )
    weight: number;

    @Column( { type: 'text', nullable: true } )
    stats?: string;

    @Column( { nullable: false } )
    teamId: string;

    /**
     * The players complete name
     *
     * @readonly
     * @memberof Player
     */
    get fullName () { return `${ this.firstName } ${ this.lastName }`; }
}

[...]

PR #5443, opened 2020-01-30
fix: entity object undefined in update subscriber

simpert:
When defining a `beforeUpdate` implementation in a subscriber the entity object is undefined on the passed in `event.entity` property when calling `update` on a repository and passing a updated entity object.

Closes: #4947


simpert:
I was unable to get all database's set up locally and therefore could not execute all tests for all databases. For SqlServer, MongoDb, and Sqlite I was able to run and ensure all tests pass. I did however make some edits after reviewing failed CI test run logs and was able to get it passing in more environments.

simpert:
For the common check it fails in `npm test` with a segmentation fault error after running a fraction of the tests and the oracle check fails to spin up environment.

I do not think I can do much more on this PR. Sorry.

pleerock:
Thanks!

PR #5479, opened 2020-02-05
migration generation: update enumName for Postgres

### LINKED ISSUE #5478: Setting enumName doesn't change how migrations get generated
I want to share enum names with the same values across several tables, but after making the changes `generateMigrations` doesn't actually change anything. `enumNames` ought to also be reflected in migration generation

### Original comment in enumName PR

@ejose19 - am i'm right in thinking this won't affect the migrations generated? It appears they're still being prefixed with table names and ignoring the new `enumName`..

I can take a crack at a PR if i can get a few pointers..

_Originally posted by @elmpp in https://github.com/typeorm/typeorm/pull/4824#issuecomment-545484163_

osdiab:
Resolves #5478 for Postgres databases (which is what I use, so I didn't do this for the other drivers).

If you add an `enumName` field for an entity's column, generated migrations for Postgres will now properly edit the name of the enum for that column.

## Old comment

this is my ~~first~~ second! code contribution so I want to make sure I'm going the right direction before I put a bunch of effort applying this kind of change all over the place.

Still need to figure out how this repo does tests for things like this, how i can try it out to see if it works

My understanding is that the query building in the individual drivers determines whether or not a change in a generated migration will actually get outputted - and i noticed that the optional enumName parameter wasn't being passed in in a lot of cases. this makes it non-optional to force the client code to provide something, as a way to smoke out all the places where it may be missing.


osdiab:
hmm i dont think this will do anything different, given that i just filled in what the code defaults to. hmmm not totally sure where the names get generated for migrations then...

osdiab:
reverted the old changes (but left my previous ones in place), and tried instead to modify `findChangedColumns()`  (in postgres only for now) to check for changed enum name. maybe this is the right way to go?

osdiab:
(after looking through the code I believe the names should be right but the migration generation code just isn't detecting the change, which gets called by each of the appropriate `SchemaBuilder.log()` -> `RdbmsSchemaBuilder.updateExistColumns()` -> `driver.findChangedColumns()` calls i believe)

in `PostgresDriver`, `createEnumTypeSql()` is called in `addColumn()`, `createTable()`, `changeColumns()`, and `dropColumn()` - i think `createTable()`, `addColumn()` and `dropColumn()` should all be fine because the `createEnumTypeSql()` method already produces the correct names, so I guess `changeColumns()` needs to be called, which is in the `RdbmsSchemaBuilder.updateExistColumns()` method - so I think what I did might be on the right track then. Need to figure out how to try it out now

src/driver/postgres/PostgresQueryRunner.ts
```diff
@@ -711,21 +711,24 @@ export class PostgresQueryRunner extends BaseQueryRunner implements QueryRunner
...
                 // rename old ENUM
-                upQueries.push(new Query(`ALTER TYPE "${enumTypeBeforeColumnChange.enumTypeSchema}"."${enumTypeBeforeColumnChange.enumTypeName}" RENAME TO ${oldEnumNameWithoutSchema}`));
-                downQueries.push(new Query(`ALTER TYPE ${oldEnumName} RENAME TO  "${enumTypeBeforeColumnChange.enumTypeName}"`));
+                upQueries.push(new Query(`ALTER TYPE "${prevEnumSchema}"."${prevEnumName}" RENAME TO ${prevEnumNameWithOldSuffix}`));
+                downQueries.push(new Query(`ALTER TYPE ${newEnumName} RENAME TO ${newEnumNameWithOldSuffix}`));
 
                 // create new ENUM
                 upQueries.push(this.createEnumTypeSql(table, newColumn));
-                downQueries.push(this.dropEnumTypeSql(table, oldColumn));
+                downQueries.push(this.createEnumTypeSql(table, oldColumn));
```
pleerock: looks like a typo?

osdiab: whoops thanks for pointing it out!

osdiab: so actually I think that this isn't wrong - when you're doing a `down` query we want to create the old enum type, not delete it - because the old enum type doesn't exist in the database yet. What do you think?

osdiab:
cc @pleerock @AlexMesser @Kononnable

src/driver/postgres/PostgresQueryRunner.ts
AlexMesser: down queries runs in reverse order. In this case `dropEnumTypeSql` will be executed at the end of Down process and drops a "new" enum type, which was created by Up query. So if you replace it with `createEnumTypeSql` it will try to create already existiing "new" enum type

osdiab: Oh I see didn’t know that, will fix then!

osdiab:
OK ready for a review again

osdiab:
hmmm nm looks like i broke other tests, checking...

JipSterk:
@osdiab any eta on this pull request?

osdiab:
@JipSterk sorry ended up focusing on other stuff on my company but i'll fix it up first thing tomorrow! (i'm on japan time)

osdiab:
@JipSterk hopefully tests pass now, heading to bed but if they do should be ready for re-review from maintainers.

osdiab:
@Kononnable @pleerock @AlexMesser  ready for a review!

JipSterk:
@osdiab Thanks a bunch 🌈

osdiab:
Pinging on this @Kononnable @pleerock @AlexMesser

davidlj95:
Thanks for the fix @osdiab , once it gets merged, we'll use it! 👀

osdiab:
@Kononnable @pleerock @AlexMesser pinging again - any chance you can review? its been a month and a half and people are pinging me about when this will get merged lol

osdiab:
pinging again, hope someone can review this... @pleerock @AlexMesser @Kononnable

JipSterk:
Im sorry for pinging again but would like to know some kind of ETA since it currently broken. @pleerock @AlexMesser @Kononnable

Kononnable:
I do not consider myself as part of typeorm team(or however we want to call it) for several months now, nor do I plan to come back to this role with the way things work here(not directly related to number of issues, prs). Because of that I won't be the one to merge this PR, so please don't ping me next time.

All I can say is that I'm sorry for the state of the typeorm, but I don't think I will be the one who change it. At least not with its current organization structure which I doubt will be ever changed.

osdiab:
Bummer :P yeah I'm pretty frustrated with the unresponsiveness of the maintainers and the huge number of issues. Will probably be looking for alternatives for my company, but in the meantime I hope @pleerock and whoever else might be maintaining the project can help.

osdiab:
Calling out to @imnotjames since I see you've been active lately on TypeORM and this PR's been waiting for a while!

src/driver/postgres/PostgresQueryRunner.ts
```diff
@@ -711,31 +711,32 @@ export class PostgresQueryRunner extends BaseQueryRunner implements QueryRunner
...
                 upQueries.push(this.createEnumTypeSql(table, newColumn));
-                downQueries.push(this.dropEnumTypeSql(table, oldColumn));
+                downQueries.push(this.dropEnumTypeSql(table, newColumn, newEnumNameWithOldSuffixAndSchema));
 
-                // if column have default value, we must drop it to avoid issues with type casting
+                // if column has default value, we must drop it to avoid issues with type casting
                 if (newColumn.default !== null && newColumn.default !== undefined) {
                     upQueries.push(new Query(`ALTER TABLE ${this.escapePath(table)} ALTER COLUMN "${newColumn.name}" DROP DEFAULT`));
-                    downQueries.push(new Query(`ALTER TABLE ${this.escapePath(table)} ALTER COLUMN "${newColumn.name}" SET DEFAULT ${newColumn.default}`));
+                    downQueries.push(new Query(`ALTER TABLE ${this.escapePath(table)} ALTER COLUMN "${newColumn.name}" SET DEFAULT ${oldColumn.default}`));
```
imnotjames, line 734: Is this correct?  This is setting to the old column def default

src/driver/postgres/PostgresQueryRunner.ts
```diff
@@ -747,17 +748,19 @@ export class PostgresQueryRunner extends BaseQueryRunner implements QueryRunner
...
+                // down: rename new enum type and create old enum type
+                upQueries.push(this.dropEnumTypeSql(table, newColumn, prevEnumNameWithOldSuffixAndSchema));
+                downQueries.push(this.createEnumTypeSql(table, oldColumn));
+                downQueries.push(new Query(`ALTER TYPE ${newEnumNameWithSchema} RENAME TO ${newEnumNameWithOldSuffix}`));
             }
 
             if (oldColumn.isNullable !== newColumn.isNullable) {
                 if (newColumn.isNullable) {
-                    upQueries.push(new Query(`ALTER TABLE ${this.escapePath(table)} ALTER COLUMN "${oldColumn.name}" DROP NOT NULL`));
+                    upQueries.push(new Query(`ALTER TABLE ${this.escapePath(table)} ALTER COLUMN "${newColumn.name}" DROP NOT NULL`));
```
imnotjames, line 760: Switched from old column to new column.  Intended?

src/driver/postgres/PostgresQueryRunner.ts
```diff
@@ -747,17 +748,19 @@ export class PostgresQueryRunner extends BaseQueryRunner implements QueryRunner
...
                     downQueries.push(new Query(`ALTER TABLE ${this.escapePath(table)} ALTER COLUMN "${oldColumn.name}" SET NOT NULL`));
                 } else {
-                    upQueries.push(new Query(`ALTER TABLE ${this.escapePath(table)} ALTER COLUMN "${oldColumn.name}" SET NOT NULL`));
+                    upQueries.push(new Query(`ALTER TABLE ${this.escapePath(table)} ALTER COLUMN "${newColumn.name}" SET NOT NULL`));
```
imnotjames, line 763: Old column to new column via diff.  Intentional?

imnotjames (commented):
Looks pretty good.  Probably didn't need to change variable names which would have made this change smaller.

Does this leave the _old around?  Does it ever clean it up?

src/driver/postgres/PostgresQueryRunner.ts
osdiab: It's counterintuitive but I believe it's correct - lines 733 - 749 are a block together, if you read the up queries in order from top down and the down queries from bottom up, they match one another i believe - its been a while since i've looked at this but i believe what's going on is applying the `oldColumn.default` value restores whatever was previously the default. Though perhaps we should be only doing this logic if the old column had a default at all, rather than always. a test for this would probably also be helpful!

src/driver/postgres/PostgresQueryRunner.ts
osdiab: At this point i believe technically `oldColumn` shouldn't exist, it would have already been renamed. but that said, i dont believe that the postgres generated migrations actually ever rename a column (always seems to drop and recreate) so it may be a moot point.

src/driver/postgres/PostgresQueryRunner.ts
osdiab: yeah, see above - i think lol. need to review my old code or maybe write some extra tests to make sure haha.

imnotjames: Ok - once you're ready, re-request my review.

JipSterk:
hey @osdiab i was just checking in on the status of the pr. i hope you have some spare time to close this issue. keep safe & keep sane 🌈❤️

osdiab:
sorry for the delay, will take a look soon! thanks for the reminder.

osdiab:
@JipSterk sorry still on my radar but just haven't had the time. Hopefully will get to it this week.

Issue #4958, opened 2019-10-23
TypeORM 0.2.20 getRepository returns results from another Repo.


mabintact:
**Issue type:**

[ ] question
[x] bug report
[ ] feature request
[ ] documentation issue

**Database system/driver:**

[ ] `cordova`
[ ] `mongodb`
[ ] `mssql`
[ ] `mysql` / `mariadb`
[ ] `oracle`
[x] `postgres`
[ ] `cockroachdb`
[ ] `sqlite`
[ ] `sqljs`
[ ] `react-native`
[ ] `expo`

**TypeORM version:**

[ ] `latest`
[ ] `@next`
[x] `0.2.20`

**Steps to reproduce or a small repository showing the problem:**

As soon as I update to TypeORM 0.2.20 (NPM Version) when we ran a graphql query the results obtain had nulls everywhere except for the keys. I added a console log to the query and discovered the results were for another repository NOT included in the query's file.

async getQuerySomething(_root: any, _args: any) {
  const queryData = await getConnection("mainConn")
  .getRepository(Repo)
  .find();
  
  console.log("/** getQuerySomething * Start ***");
  console.log(queryData);
  console.log("*** getQuerySomething * End   **/");
  
  return queryData;
},

The output of queryData showed results that were inconsistent with the repository. When outputing the result of getRepository, we could clearly see the table name was wrong. It wasn't even an Entity included in the Resolver file. The Entity came from another Resolver.

Reproducing the issue was easy as all we had to do was go back and forth from 0.2.19 to 0.2.20.
0.2.20 is the only version that seemed to cause the issue.

pleerock:
can you check if its the same issue as #4940 ?

mabintact:
The class names are different in our case.

One is:
@Entity("request_type", {schema: "public" } )
export class RequestType {...}

and the other:
@Entity("request_status", {schema: "public" } )
export class RequestStatus {...}

smallg:
@mabintact also catch this issue, how about solution?

Soinou:
We are also encountering this issue. Here's what happens:

## first.ts

```typescript
import { Column, Entity } from "typeorm";

@Entity({ name: "first" })
export class First {
    @Column("int", {
        nullable: false,
        primary: true,
        unique: true,
    })
    public id!: number;

    @Column("varchar", {
        nullable: true,
        primary: false,
        unique: false,
    })
    public name?: string;
}
```

## second.ts

```typescript
[...]

Issue #4967, opened 2019-10-24
0.2.20 release breaks entity repo resolving mechanism


sevenryze:
**Issue type:**

[ ] question
[x] bug report
[ ] feature request
[*] documentation issue

**Database system/driver:**

[ ] `cordova`
[ ] `mongodb`
[ ] `mssql`
[x] `mysql` / `mariadb`
[ ] `oracle`
[ ] `postgres`
[ ] `cockroachdb`
[ ] `sqlite`
[ ] `sqljs`
[ ] `react-native`
[ ] `expo`

**TypeORM version:**

[X] `latest`
[ ] `@next`
[ ] `0.x.x` (or put your version here)

**Steps to reproduce or a small repository showing the problem:**

Before the 0.2.20 release, e.g. 0.2.19, we could use this kind of code to specify a count column in db:
```ts
  @Index()
  @Column({
    comment: "The count, without setting default option!"
  })
  private count: number = 0;
```

Everything works as expected, until the latest version released.

For now, if you use 0.2.20 version, during the db initialization stage, ORM will give the error like this:
```
The field `count` default value has not been set. 
```

I highly recommend listing this as a breaking warning as it's a big breaks for code.
Or, is it a regression bug introducing by accident?

antoine-pous:
I don't find the way you're using to assign default value into the documentation :thinking: 

Then you've probaby misused the defaut value assignation, IMO this is not really a bug/breaking change if it's not used the way it was designed for

sevenryze:
So what did you mean that I should rely on the `default` option in the
@column decorator rather than the default JS class field feature to specify
the default value for a field?
Best regards
Seven Ryze

pleerock:
We are trying not to introduce a breaking changes. Are you sure this error comes from TypeORM? Can you provide a stack trace of this error? Where exactly this errors comes?

sevenryze:
@pleerock yes, I'm sure to say that's typeorm's breaking change.

I dig some, I believe this issue is caused by #4804 PR, which breaks the repository metadata resolving logic. 

```ts
protected findMetadata(target: Function|EntitySchema<any>|string): EntityMetadata|undefined {
        return this.entityMetadatas.find(metadata => {
            if (typeof metadata.target === "function" && typeof target === "function" && metadata.target.name === target.name) {  // <--- See this? Mark 1
                return true;
            }
            if (metadata.target === target) {
                return true;
            }
            if (target instanceof EntitySchema) {
[...]

PR #5486, opened 2020-02-06
revert: "fix: connection Reuse is broken in a Lambda environment: (#4804)"

millsmcilroy:
This reverts commit 7962036be3c009f8de3030e83002be0346e35d48 

The original PR came about from this issue: https://github.com/typeorm/typeorm/issues/3427.
The author of the original issue has since realized the root problem they were facing was unrelated: https://github.com/typeorm/typeorm/issues/3427#issuecomment-549770047

It seems this "monkeypatch" is causing more problems than any workaround it might have provided is worth. The conditional `metadata.target.name === target.name` in the `findMetadata` function is not a thorough-enough comparison. Function names are often the same in large codebases and especially in minified code (as @Benjamin-Dobell points out https://github.com/typeorm/typeorm/pull/4804#issuecomment-580075220). This results in false positives returned from `this.entityMetadatas.find()` depending on the order of the `EntityMetadata` in the array.

Closes:
https://github.com/typeorm/typeorm/issues/4958 TypeORM 0.2.20 getRepository returns results from another Repo.
https://github.com/typeorm/typeorm/issues/4967 0.2.20 release breaks entity repo resolving mechanism

### DIFF
--- a/src/connection/Connection.ts
+++ b/src/connection/Connection.ts
@@ -479,12 +479,8 @@ export class Connection {
      */
     protected findMetadata(target: Function|EntitySchema<any>|string): EntityMetadata|undefined {
         return this.entityMetadatas.find(metadata => {
-            if (typeof metadata.target === "function" && typeof target === "function" && metadata.target.name === target.name) {
+            if (metadata.target === target)
                 return true;
-            }
-            if (metadata.target === target) {
-                return true;
-            }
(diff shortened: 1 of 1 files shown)


pleerock:
ahhh... okay...

pleerock:
thanks : )

PR #5491, opened 2020-02-06
feat: Base overload function signatures on Repository methods

DrewVartanian:
This change adds overload function signatures that match the actual function argument and return types to Repository methods. This allows these methods to be called with arguments that match the function argument types, but don't match one of the other specific overload signatures.

Without these signatures, generically extending one of these methods becomes difficult.  For example, the following typescript will not compile, saying "No overload matches this call", ultimately complaining "Type 'FindManyOptions<BaseEntity>' has no properties in common with type 'FindConditions<BaseEntity>'":

```
import {Repository, BaseEntity, FindManyOptions, FindConditions} from './typeorm/src';

function find(repo: Repository<BaseEntity>, optionsOrConditions?: FindManyOptions<BaseEntity>|FindConditions<BaseEntity>) {
    return repo.find(optionsOrConditions); // error because no signature matches (optionsOrConditions? FindManyOptions<Entity>|FindConditions<Entity>) => Promise<Entity[]
}
```

This is because neither of the declared signatures:
`find(options?: FindManyOptions<Entity>): Promise<Entity[]>;`
and
`find(conditions?: FindConditions<Entity>): Promise<Entity[]>;`
can accept an argument that is:
`FindManyOptions<Entity>|FindConditions<Entity>`

Adding the signature
`find(optionsOrConditions?: FindManyOptions<Entity>|FindConditions<Entity>): Promise<Entity[]>`
means that we can still utilize the more specific signatures above (which is particularly useful when the return type is different), but still allows us to use the less specified argument type for when we want to do something like extend the method.

If this PR is accepted, I can add these base overload signatures to the methods in other files, but I wanted to submit one file first, in case there was any feedback that majorly changed this approach.

### DIFF
--- a/src/repository/Repository.ts
+++ b/src/repository/Repository.ts
@@ -144,6 +144,12 @@ export class Repository<Entity extends ObjectLiteral> {
     save<T extends DeepPartial<Entity>>(entity: T, options?: SaveOptions): Promise<T & Entity>;
 
     /**
+     * Saves one or many given entities in the database.
+     * If entity or entities do not exist in the database then inserts, otherwise updates.
+     */
+    save<T extends DeepPartial<Entity>>(entityOrEntities: T|T[], options?: SaveOptions): Promise<T|T[]>;
+
+    /**
      * Saves one or many given entities.
      */
     save<T extends DeepPartial<Entity>>(entityOrEntities: T|T[], options?: SaveOptions): Promise<T|T[]> {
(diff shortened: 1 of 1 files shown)


pleerock:
I aware about this issue however I feel like TypeScript itself should handle this. The reason why they are separated definitions, is because its much easier for users to understand 

```ts
find(options?: FindManyOptions<Entity>): Promise<Entity[]>;
find(conditions?: FindConditions<Entity>): Promise<Entity[]>;
```

than 

```ts
find(conditions?: FindManyOptions<Entity>|FindConditions<Entity>)
```

its actually a very trivial example, now imagine more union types in, and more parameters any of each can be union too.

To fix your issue you either need to use `any` either need to create a type guards.

I don't want you to use `any` as in your case, and at the same time I don't want you to use runtime-enabled `if` statements  for type guards, but at the same time I don't want to complicate types with additional type signature. I feel like if we go with your approach we'll need to change some other more complicated places too, and overall I don't like this change.

imnotjames:
Agreed, well written, it doesn't add much, breaks uniformity across the project, & is difficult to maintain..  let's stick with leveraging TypeScript here!

Thank you for the contribution either way & don't let this being closed discourage you from contributing again in the future!

PR #5492, opened 2020-02-07
Fix typo

intelliot:
Per https://github.com/typeorm/typeorm/blob/master/docs/decorator-reference.md#column

You can explicitly specify a column type. Fix in README.


pleerock:
Thanks, always mess these two words.

PR #5500, opened 2020-02-10
docs: delete single quote from "'age' +1"

bcking92:
single quotes make error called ER_TRUNCATED_WRONG_VALUE: Truncated incorrect DOUBLE value.
It is working without single quotes.


pleerock:
its really depend on a database you use, and single quoting was to make sure column name is escaped. But okay we can merge it since it doesn't matter a lot

PR #5505, opened 2020-02-11
feat: add updateAllColumns to SaveOptions

### LINKED ISSUE #3490: Method save does not return the entire model
**Issue type:**

[ ] question
[x] bug report
[ ] feature request
[ ] documentation issue

**Database system/driver:**

[ ] `cordova`
[ ] `mongodb`
[ ] `mssql`
[ ] `mysql` / `mariadb`
[ ] `oracle`
[x] `postgres`
[ ] `sqlite`
[ ] `sqljs`
[ ] `react-native`
[ ] `expo`

**TypeORM version:**

[ ] `latest`
[ ] `@next`
[0.2.8] `0.x.x` (or put your version here)

```
class Post {
  @PrimaryGeneratedColumn()
  id: number

  @UpdateDateColumn({ type: 'timestamptz' })
  updatedAt: Date

  @CreateDateColumn({ type: 'timestamptz' })
  createdAt: Date

  @Column({ type: 'varchar' })
  title: string

  @Column({ type: 'text' })
  description: string
}

const connection = await createConnection(connectionOptions)
const repository = connection.getRepository(Post)

const p = await repository.save(repository.create({ title: 'D' }))
console.log(p)
#----------------------------------------
Post {
  title: 'D',
  id: 1,
  updatedAt: 2019-01-23T06:07:18.769Z,
  createdAt: 2019-01-23T06:07:18.769Z }
#----------------------------------------

const p1 = await repository.save(repository.create({ id: p.id, title: 'F' }))
console.log(p1)
#----------------------------------------
Post { id: 1, title: 'F', updatedAt: 2019-01-23T06:07:18.800Z }
#----------------------------------------
```

Using the PUT method, a part of the model comes to us. The save method accepts `DeepPartial<Entity>`, but in this case it does not return the entire model.

<!--
    [x] `mysql`
    [ ] `postgres`

!>

ejose19:
Allows retrieving all columns with .save() on both insert and update within a single operation.

Fixes #3490

### DIFF
--- a/src/persistence/SubjectExecutor.ts
+++ b/src/persistence/SubjectExecutor.ts
@@ -285,6 +285,7 @@ export class SubjectExecutor {
                         .into(subjects[0].metadata.target)
                         .values(bulkInsertMaps)
                         .updateEntity(this.options && this.options.reload === false ? false : true)
+                        .updateAllColumns(this.options && this.options.updateAllColumns === true ? true : false)
                         .callListeners(false)
                         .execute();
 
@@ -311,6 +312,7 @@ export class SubjectExecutor {
                             .into(subject.metadata.target)
                             .values(subject.insertedValueSet)
                             .updateEntity(this.options && this.options.reload === false ? false : true)
+                            .updateAllColumns(this.options && this.options.updateAllColumns === true ? true : false)
                             .callListeners(false)
                             .execute()
                             .then(insertResult => {
@@ -385,6 +387,7 @@ export class SubjectExecutor {
                     .update(subject.metadata.target)
                     .set(updateMap)
                     .updateEntity(this.options && this.options.reload === false ? false : true)
+                    .updateAllColumns(this.options && this.options.updateAllColumns === true ? true : false)
                     .callListeners(false);
 
                 if (subject.entity) {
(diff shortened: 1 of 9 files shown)


pleerock:
I left my feedback on #3490, maybe we can came up with a different solution

ejose19:
Ok regarding your proposition on #3490, you want .save to return the complete entity correct? In that case what we will do regarding {"reload": false}? It will be supported? In that case .save will return a partial entity.

pleerock:
> you want .save to return the complete entity correct?

yes

> In that case what we will do regarding {"reload": false}? It will be supported? In that case .save will return a partial entity.

reload: false simply should work as it currently works when its reload: false

pleerock:
Also, changes should be against `next` branch

ejose19:
Ok, sounds good. I will tackle it somewhere this weekend. One thought, does next branch has a release date yet? Because if it's nowhere near yet I guess users would like to have the feature on master, so a non breaking change (like this PR) would help until next is released.

ejose19:
I wrote an WIP on #5680, please take a look as your feedback can help me to finish it.

PR #5517, opened 2020-02-13
Return NULL when normalize default null value

### LINKED ISSUE #5509: Inappropiate migration generated for 'default: null'
**Issue type:**

[ ] question
[X] bug report
[ ] feature request
[ ] documentation issue

**Database system/driver:**

[ ] `cordova`
[ ] `mongodb`
[ ] `mssql`
[X] `mysql` / `mariadb`
[ ] `oracle`
[ ] `postgres`
[ ] `cockroachdb`
[ ] `sqlite`
[ ] `sqljs`
[ ] `react-native`
[ ] `expo`

**TypeORM version:**

[X] `latest`
[ ] `@next`

When using the option default: null for an entity's column, and generating a migration using typeorm migration:generate, the generated migrations get into a loop. The generated migration is always the same, because 'null' default value is being normalized as null instead of NULL. When typeorm checks default value for the column it finds 'NULL' !==  'null'.

This bug is related to #3345

In order to reproduce with a simple User entity:

```ts
import {
  Entity,
  PrimaryGeneratedColumn,
  Column,
  Unique,
  CreateDateColumn,
  UpdateDateColumn
} from 'typeorm';
import { Length } from 'class-validator';

@Entity()
@Unique(['username'])
export default class User {
  @PrimaryGeneratedColumn()
  id: number;

  @Column()
  @Length(4, 20)
  username: string;

  @Column()
  @Length(4, 100)
  password: string;

  @Column({ nullable: true, default: null })
  country: string;

  @Column('simple-array')
  @IsNotEmpty()
  roles: string[];

  @Column()
  @CreateDateColumn()
  createdAt!: Date;

  @Column()
  @UpdateDateColumn()
  updatedAt: Date;
}
```

- Sync Schema

```sh
ts-node ./node_modules/typeorm/cli.js schema:sync
```

- Generate Migration

```sh
ts-node ./node_modules/typeorm/cli.js migration:generate -n DefaultNullTest
```

Obtained:

- Up

```sql
ALTER TABLE `user` CHANGE `country` `country` varchar(255) NULL DEFAULT null
```

- Down

```sql
ALTER TABLE `user` CHANGE `country` `country` varchar(255) NULL DEFAULT NULL
```

I think modifying [MysqlDriver.ts line 593](https://github.com/typeorm/typeorm/blob/8c67682fa41b8336cd96150baf654adcf5f954b5/src/driver/mysql/MysqlDriver.ts#L593) to:
```ts
return `NULL`;
```
should fix the problem

<!--
    [x] `mysql`
    [ ] `postgres`

!>

thefat32:
This will fix #5509

### DIFF
--- a/src/driver/mysql/MysqlDriver.ts
+++ b/src/driver/mysql/MysqlDriver.ts
@@ -590,7 +590,7 @@ export class MysqlDriver implements Driver {
             return `'${defaultValue}'`;
 
         } else if (defaultValue === null) {
-            return `null`;
+            return `NULL`;
 
         } else {
             return defaultValue;


pleerock:
Thanks! It would be great to have a test so this won't break again.

PR #5524, opened 2020-02-14
fix: un-quoted column name is not safe

JVMartin:
This is a fix for an issue was introduced by #3700.

Column names must be quoted, as otherwise there are obviously issues.

Here's an example of the error that this fix addresses:
```typescript
enum FromType {
  NO_REPLY = 'no-reply',
  RANDOM = 'random',
}

@Entity('broadcast')
class BroadcastEntity {
  @PrimaryGeneratedColumn()
  id: number;

  @Column({
    type: 'simple-enum',
    enum: FromType,
    nullable: false,
  })
  from: FromType;
}
```

Trying to synchronize produces:
```
  console.log ../node_modules/typeorm/platform/PlatformTools.js:200
    query failed: CREATE TABLE "broadcast" ("id" integer PRIMARY KEY AUTOINCREMENT NOT NULL, "from" varchar CHECK( from IN ('no-reply','random') ) NOT NULL)

  console.log ../node_modules/typeorm/platform/PlatformTools.js:200
    error: [Error: SQLITE_ERROR: near "from": syntax error] {
      errno: 1,
      code: 'SQLITE_ERROR'
    }
```

### DIFF
--- a/src/driver/sqlite-abstract/AbstractSqliteQueryRunner.ts
+++ b/src/driver/sqlite-abstract/AbstractSqliteQueryRunner.ts
@@ -1084,7 +1084,7 @@ export abstract class AbstractSqliteQueryRunner extends BaseQueryRunner implemen
         }
 
         if (column.enum)
-            c += " CHECK( " + column.name + " IN (" + column.enum.map(val => "'" + val + "'").join(",") + ") )";
+            c += " CHECK( \"" + column.name + "\" IN (" + column.enum.map(val => "'" + val + "'").join(",") + ") )";
         if (column.isPrimary && !skipPrimary)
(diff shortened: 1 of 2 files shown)


pleerock:
Thanks for PR!

There are some issues in this PR:

* its not test-covered. It means next time somebody change something in this code can easily break your efforts
* tests are failing because PR is not finished. Looks like you need to fix other places that rely on this code (looks like there is regular expression that extract name out of CHECK expression)
* would be great if you can fix this issue for every database driver, if any of them have this problem too

JVMartin:
Good call-outs, thank you. When I find some time, I will design some tests and ensure that existing tests pass.

imnotjames:

PR #5529, opened 2020-02-15
docs: update many to many relations with custom properties

econdie:
for this example, it is referencing associating a custom **createdAt** property but then the actual following example uses an **order** property. I thought it would make sense to match these up.


pleerock:
Thanks for pointing to the problem! I applied few changes, so instead of changing the actual code, I changed the description! Thanks

PR #5539, opened 2020-02-17
refactoring instance of with Array.isArray()

naidishuli:
This pull request is based on the content on this page: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/instanceof. This will solve and prevent troubles with typeorm functionalities when calling functions from different contexts. When using typeorm from REPL you will not be able to obtain relations of a specific entity. This PR will solve the issue.


pleerock:
Thanks!

Issue #4750, opened 2019-09-16
Save + null prototype object = Cannot convert object to primitive value


mcrawshaw:
**Issue type:**

[x] bug report

**Database system/driver:**

[x] `postgres` (but likely not db specific)

**TypeORM version:**

[x] `latest`

**Steps to reproduce or a small repository showing the problem:**

Error + stack:
```ts
error: GraphQL failure: TypeError: Cannot convert object to primitive value
    at Array.join (<anonymous>)
    at Array.toString (<anonymous>)
    at isNaN (<anonymous>)
    at Function.Object.<anonymous>.OrmUtils.compare2Objects (/Users/markcrawshaw/Code/metamap-server/src/util/OrmUtils.ts:171:13)
    at Function.Object.<anonymous>.OrmUtils.deepCompare (/Users/markcrawshaw/Code/metamap-server/src/util/OrmUtils.ts:118:23)
    at /Users/markcrawshaw/Code/metamap-server/src/persistence/SubjectChangedColumnsComputer.ts:90:38
```
Error reported when null prototype object is passed.  It is due to global `isNaN` not being safe with null prototype objects.  It can be fixed by either correcting the order of the if conditions or by using `Number.isNaN`.

`util/OrmUtils.ts@compare2Objects` from
```ts
if (isNaN(x) && isNaN(y) && typeof x === "number" && typeof y === "number")
```
to
```ts
if (typeof x === "number" && typeof y === "number" && isNaN(x) && isNaN(y))
```

And if you're wondering if global `isNaN` could really do this, try this:
```js
let result;

const nullProtoObject = Object.create(null);
nullProtoObject.prop = "value";

try {
  isNaN(nullProtoObject);
}
catch (e) {
  result = `isNaN NOT SAFE (${e})`;
}

console.info(result || "isNaN SAFE")
```

Regards,
Mark.

mcrawshaw:
@pleerock really quick fix here...  just correcting the order of conditions within an `if`... any chance you could have a look for me?

mcrawshaw:
@pleerock Mark (from Aplas) here, any chance you could have a look at this one for us?  Should be a quick fix...

pleerock:
hey @mcrawshaw, sorry I'm not watching for all issues and didn't see your request. It would be better if you ping me directly in TypeORM's slack next time! Is #5566 fixing your issue?

mcrawshaw:
No worries @pleerock, thanks @zbessette!

[...]

PR #5566, opened 2020-02-24
Small fix for iss #4750

zbessette:
Change check in util/OrmUtils to prevent 'Cannot convert object to primitive value' error iss #4750


pleerock:
Thanks! Would be great if you can cover it with tests in a meantime

PR #5571, opened 2020-02-26
fix: SAP HANA driver fixes

Szbuli:
Unfortunately left some obvious copy paste errors. I checked the whole file, I hope there are no more errors like this... #5343


pleerock:
Thanks!

Issue #4418, opened 2019-07-11
ILIKE find operator


ivanveselin:
**Issue type:**

[ ] question
[ ] bug report
[x] feature request
[ ] documentation issue

**Database system/driver:**

[ ] `cordova`
[ ] `mongodb`
[ ] `mssql`
[ ] `mysql` / `mariadb`
[ ] `oracle`
[x] `postgres`
[ ] `cockroachdb`
[ ] `sqlite`
[ ] `sqljs`
[ ] `react-native`
[ ] `expo`

**TypeORM version:**

[x] `latest`
[ ] `@next`
[ ] `0.x.x` (or put your version here)

Would it be a big hassle to include ILIKE find operator for case insensitive string query? I see that LIKE is implemented, ILIKE would be a nice touch

manuhortet:
Perform case insensitive operations right now is pretty counterintuitive, this is a needed feature IMO

wafs:
If you're using postgres, I've got a solution to have a custom operator:

```typescript
import { Connection, FindOperator, FindOperatorType } from 'typeorm';

class FindOperatorWithExtras<T> extends FindOperator<T> {
  constructor(
    type: FindOperatorType | 'ilike',
    value: FindOperator<T> | T,
    useParameter?: boolean,
    multipleParameters?: boolean,
  ) {
    // @ts-ignore
    super(type, value, useParameter, multipleParameters);
  }

  public toSql(
    connection: Connection,
    aliasPath: string,
    parameters: string[],
  ): string {
    // @ts-ignore
    if (this._type === 'ilike') {
      return `${aliasPath} ILIKE ${parameters[0]}`;
    }

    return super.toSql(connection, aliasPath, parameters);
  }
}

/**
 * Find Options Operator.
 * Example: { someField: Like("%some sting%") }
 */
export function ILike<T>(
  value: T | FindOperator<T>,
): FindOperatorWithExtras<T> {
  return new FindOperatorWithExtras('ilike', value);
}
```

I would suggest opening the `FindOperator` to be extensible by consumers via something like the strategy pattern or a class, so that we don't have to do hacky things like telling TS to ignore private variable usage.

jacobator:
I used `Raw` operator for Postgres to achieve `ILIKE` like this:

```
import { Raw } from 'typeorm';

const name = 'hello';

const books = await this.booksRepository.find({
  where: {
    name: Raw(alias => `${alias} ILIKE '%${name}%'`),
  },
});
```

[...]

PR #5574, opened 2020-02-27
Feature: Refactor FindOperator to be extendable

wafs:
Inspired by this issue: https://github.com/typeorm/typeorm/issues/4418

Currently the only way (I can see) to add custom where operators is to extend the base FindOperator class, but the class isn't made very extendable, all the potential operators are handled by a switch statement inside it.

With this refactor, each operator is in charge of creating and providing its own SQL statement, so now writing a new operator is very simple, and the constructor can be locked down to restrict types at compile time.

Alternatively if the class based approach is not a route you want to take with the code base, we can just add another parameter to the FindOperator constructor which is `statement` and pass it in.

### DIFF
--- a/docs/find-options.md
+++ b/docs/find-options.md
@@ -23,7 +23,7 @@ userRepository.find({ relations: ["profile", "photos", "videos", "videos.video_a
 * `join` - joins needs to be performed for the entity. Extended version of "relations".
 
 ```typescript
-userRepository.find({ 
+userRepository.find({
     join: {
         alias: "user",
         leftJoinAndSelect: {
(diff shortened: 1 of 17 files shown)


pleerock:
Thanks for your efforts, but I consider FindOperator to mostly internal part of ORM for now. I don't like current implementation either, and for sure it must be extensible. However I wanted to refactor it and make it to look nice and extensible, but without using classes at all.

wafs:
> Thanks for your efforts, but I consider FindOperator to mostly internal part of ORM for now. I don't like current implementation either, and for sure it must be extensible. However I wanted to refactor it and make it to look nice and extensible, but without using classes at all.

Understood. 

What should people do in cases like here (  https://github.com/typeorm/typeorm/issues/4418) where they want to use db-domain specific operators (eg: `ILIKE` / `regex` / `@>`) using a repository instead of query builder?

imnotjames:
At this point in time we don't have a great way of handling database specific dialect in general. ):

As a step towards fixing that - in #6797 the FindOperator toSQL was moved to the query builder.  Perhaps in a beautiful new future to support a multitude of dialect - the driver package has its own dialect (and maybe query builder) that knows what to do with driver exported `FindOperator`s!

But.. just a small step towards one of many possibilities.

Unfortunately, that does put a hamper on merging this PR as it no longer applies.

As such I'll be closing this PR but please don't let that discourage you - I think custom find operators for a dialect would rock and any PR towards that goal is awesome.

PR #5599, opened 2020-03-02
add default value note for unique @Column option

jlvallelonga:
adds note about the "default" value of the unique option of the `@Column` decorator

I'm pretty sure this is true. Feel free to just close this PR if not. I just noticed some explicit `unique: false`'s in our project and wanted to make this clear in the docs


pleerock:
> I'm pretty sure this is true

that's why it's not mentioned in the docs.

but we can merge this PR

Issue #3416, opened 2019-01-13
Unknown fields are stripped from WHERE clause


realyze:
**Issue type:**

[ ] question
[x] bug report
[ ] feature request
[x] documentation issue

**Database system/driver:**

[ ] `cordova`
[ ] `mongodb`
[ ] `mssql`
[ ] `mysql` / `mariadb`
[ ] `oracle`
[x] `postgres`
[ ] `sqlite`
[ ] `sqljs`
[ ] `react-native`
[ ] `expo`

**TypeORM version:**

[x] `latest`
[ ] `@next`
[ ] `0.x.x` (or put your version here)

**Steps to reproduce or a small repository showing the problem:**
Reproducible with https://github.com/typeorm/typescript-example

1. Create some categories in the DB

Add this snippet
```
const categoryRepository = getRepository(Category);
const categories = await categoryRepository.find({doesNotExist: 'any value'} as any);
console.log(categories);
```

I would expect `categories` to be empty since we're querying a non-existent field. However the query actually returns everything (i.e., every record in the `Category` table).

Is there any reason the `doesNotExist` key is not present in the resulting SQL query please?

pleerock:
@vlapo this is fixed by #3324 right?

vlapo:
> this is fixed by #3324 right?

3324 was about update. This is select. In this case are unknown properties removed without error.

pleerock:
ah, okay... Shall we apply same strategy for selects?

vlapo:
Original question:
> Is there any reason the `doesNotExist` is not present in the resulting SQL query please?

@realyze if typeorm do not remove non-exists columns from your query, you will get error from database. So thats reason why we do it.

> ah, okay... Shall we apply same strategy for selects?

This is breaking change. Lot of users will get errors after it. But IMHO it is correct behaviour to throw error if someone try to use unknown column in where statement of SQL (database will return error in this case, too). @Kononnable, @havenchyk what do you think about this?

pleerock:
Im for throwing an error. If its so-breaking we need to do it against `next`.

silentroach:
Just got the same thing - thought that the field was used in the query, but it was just skipped :{

Will be great to throw an error.

realyze:
How about adding a flag that toggles between the new/old behaviour? IMO the default should be to throw (because of principle of least surprise) but it would be good to allow users to opt out.

silentroach:
Here is an issue for this #2528

[...]

Issue #3654, opened 2019-02-16
Repository.save() doesn't not check id correctly when id is buffer type.


sevenryze:
**Issue type:**

[ ] question
[*] bug report
[ ] feature request
[ ] documentation issue

**Database system/driver:**

[ ] `cordova`
[ ] `mongodb`
[ ] `mssql`
[x] `mysql` / `mariadb`
[ ] `oracle`
[ ] `postgres`
[ ] `sqlite`
[ ] `sqljs`
[ ] `react-native`
[ ] `expo`

**TypeORM version:**

[x] `latest`
[x] `@next`
[ ] `0.x.x` (or put your version here)

**Steps to reproduce or a small repository showing the problem:**
1. Use buffer as the id type. Define column type to `binary`.
1. Add new entity into db and read it, then update it's field.
1. Invoke the save() method of the entity repository.
1. The save() is always INSERT a new entity, which cause duplication on primary constrains check.
1. Then, change id type to `string` type. The save() works as normal, UPDATE.

I guess we maybe not implement the proper buffer compare logic, as only compare the object/string etc. currently.
<!--
    To answer those questions you need to put "x" inside the square brackets, for example:
    [x] `mysql`
    [ ] `postgres`

    Also, please format your code properly (by taking code blocks into ```ts .... ```)
!>

sevenryze:
I can confirm the conjecture i have made and will send a PR to fix this bug soon.

Feel free to test and merge this PR.

sevenryze:
This bug will be fixed in #3655.
Do anyone have questions?

sevenryze:
PR is merged and this issue should be closed.

PR #5603, opened 2020-03-03
fix: Unknown fields are stripped from WHERE clause (issue #3416)

johannessjoberg:
This fixes issue #3416  with `where` criteria silently passing with unknown properties instead of throwing an error.

I saw that fixing this issue is considered a breaking change and is aimed to be a part of `0.3.0`, but I know many users would really appreciate if this fix is included as soon as possible 🙏 One way for it to be not breaking and be shipped with `0.2.X` would be to use some kind of configuration/option to be able to opt-in. Do you think this would be a good idea and if-so where should it be applied?

The test `test/github-issues/3654/issue-3654.ts` was removed due to it's not correct. It simply returns all users as the `id` property in the `where` criteria are stripped making the test redundant. It would probably be a good idea to look into if issue #3654 is actually fixed.


johannessjoberg:
@pleerock Should this PR be aimed towards the `next` branch instead?

pleerock:
Thank you for contribution and sorry for a late review. Since this seems to be critical for many people, let's just merge it.

johannessjoberg:
Thank you @pleerock! Is there an ETA on the next release?

tjhiggins:
@johannessjoberg Thank you! This has been plaguing our team for awhile.

Issue #5613, opened 2020-03-04
Cannot Run migration:generate with geo columns, Regression from PR #5525


krazibit:
**Issue type:**

[ ] question
[x] bug report
[ ] feature request
[ ] documentation issue

**Database system/driver:**

[ ] `cordova`
[ ] `mongodb`
[ ] `mssql`
[ ] `mysql` / `mariadb`
[ ] `oracle`
[x] `postgres`
[ ] `cockroachdb`
[ ] `sqlite`
[ ] `sqljs`
[ ] `react-native`
[ ] `expo`

**TypeORM version:**

[ ] `latest`
[ ] `@next`
[x] `0.2.24` (or put your version here)

**Steps to reproduce or a small repository showing the problem:**

run `migration:generate` with entities that have geo column.

Fix in #5525 created a new issue.
The new clause added AND "table_name" = '${table.name}'
But table.name includes the schema name as prefix making the query return no result

Causing error `Cannot read property ‘type’ of undefined` on second line below

```ts
 const results: ObjectLiteral[] = await this.query(geographyColumnSql);
  tableColumn.spatialFeatureType = results[0].type;
  tableColumn.srid = results[0].srid;
```

krazibit:
PR https://github.com/typeorm/typeorm/pull/5614
@pleerock @ohjeyong

PR #5614, opened 2020-03-04
fix: issue #5613

krazibit:


ohjeyong:
This PR cause same error as you said on #5613 on my project which means my PR(#5525) breaks yours and this PR breaks mine. I found the problem.

In [PostgresDriver#L606](https://github.com/typeorm/typeorm/blob/b99b4ad869cbf482ae9a5558acf64084157cbedd/src/driver/postgres/PostgresDriver.ts#L606),
```
buildTableName(tableName: string, schema?: string): string {
    return schema ? `${schema}.${tableName}` : tableName;
}
```

This is the problem.
Even though the default value of connection option `schema` is `public`, if `schema` option is set on connection option, then `buildTableName` returns `public.some_table_name` but if not, then `buildTableName` returns just `some_table_name`.

My project doesn't set `schema` option and your project does.

I think there are two solutions.

1. Merge this PR and change `buildTableName` function like below.
```
buildTableName(tableName: string, schema?: string): string {
    return schema ? `${schema}.${tableName}` : `public.${tableName}`;
}
```

2. Change this where statement like this.
```
WHERE (${tablesCondition})
    AND "column_name" = '${tableColumn.name}'
    AND (
        ("table_schema" || '.' || "table_name") = '${table.name}' OR
        "table_name" = '${table.name}'
    )`;
```

Solution 1 is much cleaner and better but `buildTableName` function is referenced a lot, so I don't know is it ok to change like this. @pleerock 

Solution 2 is kind of dirty but it will not make any side effects.

krazibit:
Yeah, you are right @ohjeyong .
I'll add the second solution to the PR

krazibit:
Or even shorter as a `schema` variable is available in the method  can do ` "table_name" = '${table.name.replace(`${schema}.`)}'`

cknuteson-firstlook:
Looking forward to this getting merged. Had to rollback to 0.2.22 for my project

ohjeyong:
@krazibit Can you add the commit?

krazibit:
Sure, would add the commit in a bit

krazibit:
@ohjeyong, can you verify please? thanks

ohjeyong:
Thanks. LTGM. @pleerock Can you verify please?

winterec:
This patch fixes it for me

pleerock:
oh I already merged #5801 before I saw this PR. It looks like author tried to fix same issue, but a bit different way. Is merged approach is correct and I can close this PR? adding @igoraguiar

krazibit:
closing this as already fixed in PR-5801 as per @pleerock last comment

Issue #5592, opened 2020-02-29
0.2.24 JEST tests suddenly throwing RepositoryNotFound, no changes to code


seandearnaley:
**Issue type:**

[ ] question
[x] bug report
[ ] feature request
[ ] documentation issue

**Database system/driver:**

[ ] `cordova`
[ ] `mongodb`
[ ] `mssql`
[ ] `mysql` / `mariadb`
[ ] `oracle`
[x] `postgres`
[ ] `cockroachdb`
[ ] `sqlite`
[ ] `sqljs`
[ ] `react-native`
[ ] `expo`

**TypeORM version:**

[x] `latest`
[ ] `@next`
[ ] `0.x.x` (or put your version here)

**Steps to reproduce or a small repository showing the problem:**

Not sure how exactly this started, I ran an dependency update to upgrade to TypeORM 0.2.24, and then suddenly my tests started failing, frustratingly just rolling back to .22 the error still happens, I believe some other dependency was updated with my yarn.lock, in general this issue has been problematic during development, but these tests had been working for 3 months without issue.

tsconfig.json

```
{
  "compilerOptions": {
    "lib": ["es5", "es6", "es2016", "es2017", "dom"],
    "target": "ES2016",
    "module": "commonjs",
    "moduleResolution": "node",
    "resolveJsonModule": true,
    "outDir": "./build",
    "noImplicitAny": true,
    "sourceMap": true,
    "emitDecoratorMetadata": true,
    "experimentalDecorators": true,
    "allowSyntheticDefaultImports": true,
    "baseUrl": ".",
    "paths": {
      "*": ["node_modules/"],
      "typings/*": ["src/typings/*"]
    }
  },
  "include": ["src/**/*.ts"],
  "exclude": ["src/test/**/*.spec.ts", "node_modules"]
}

```

jest.config.js

```
module.exports = {
  preset: "ts-jest",
  testEnvironment: "node",
  collectCoverage: true,
  collectCoverageFrom: [
    "./src/**/*.{ts,tsx,js,jsx}",
    "!**/node_modules/**",
    "!**/build/**",
    "!./src/generated/**",
    "!./src/migration/**"
  ],
  testPathIgnorePatterns: [
    "/node_modules/",
    "/build/",
    "/__utils",
    "/__testData",
    "/__queries"
  ],
  globalSetup: "./src/jest.globalSetup.ts"
};
```

schema config:
```
export const schemaConfig = {
[...]

Issue #5587, opened 2020-02-28
Repository cannot be found when using existing connection


bwermes:
**Issue type:**

[ ] question
[X] bug report
[ ] feature request
[ ] documentation issue

**Database system/driver:**

[ ] `cordova`
[ ] `mongodb`
[X] `mssql`
[ ] `mysql` / `mariadb`
[ ] `oracle`
[ ] `postgres`
[ ] `cockroachdb`
[ ] `sqlite`
[ ] `sqljs`
[ ] `react-native`
[ ] `expo`

**TypeORM version:**

[X] `latest`
[ ] `@next`
[ ] `0.x.x` (or put your version here)

**Steps to reproduce or a small repository showing the problem:**

I'm encountering an issue where a getRepository(<repo name>) call is providing the repository after the SQL connection is initially made. However, after a subsequent call to the database, where I'm reusing the existing connection, the repository is not found.

Response: 
```
{
    "error": "No repository for \"Store\" was found. Looks like this entity is not registered in current \"default\" connection?"
}
```

Here is a screen shot of the initial call followed by a second call a moment later. The function being triggered in both calls is the same.
![typeorm-repo-not-found](https://user-images.githubusercontent.com/22311634/75593294-c3a18680-5a39-11ea-9a9f-fbb6025aec35.png)

[StorePack.zip](https://github.com/typeorm/typeorm/files/4269451/StorePack.zip)

millsmcilroy:
Related to this issue https://github.com/typeorm/typeorm/issues/3427? If you're using sls-offline, invoke it with `--skipCacheInvalidation`.

rmolinamir:
@millsmcilroy thanks for the response. I'm not sure if it might be related to that. I was using a singleton for my connections in order to never lose it. I had no problems w/ `0.2.22` but after updating to `0.2.24` I started having issues w/ this. There's another open issue regarding this same problem, #5592.

pleerock:
can you please provide a reproduction repo?

jukra:
Same thing happening here, and I also verified that it is due to this change https://github.com/typeorm/typeorm/commit/e3be9e34e00768af3304b238389fb80ce13bb567.

With serverless offline using  `--skipCacheInvalidation` kind of solves the issue, but with that you are not able to hot reload any changes. What I have done for now as temp fix is to disable the connection reusing with serverless offline as I don't mind it creating new connections when developing locally.

phong-innomizetech:
Look at my commented at #5592

[...]

PR #5627, opened 2020-03-06
no-repository-for-XEntity-was-found

phong-innomizetech:
Adding condition once entity and target have typed is a function

### DIFF
--- a/src/connection/Connection.ts
+++ b/src/connection/Connection.ts
@@ -479,8 +479,11 @@ export class Connection {
      */
     protected findMetadata(target: Function|EntitySchema<any>|string): EntityMetadata|undefined {
         return this.entityMetadatas.find(metadata => {
-            if (metadata.target === target)
+            if (metadata.target === target ||
+                typeof metadata.target === "function" && typeof target === "function" && metadata.target.name === target.name) {
                 return true;
+            }
+
             if (target instanceof EntitySchema) {
                 return metadata.name === target.options.name;
             }


millsmcilroy:
@phong-innomizetech This is ultimately a revert of this revert: https://github.com/typeorm/typeorm/pull/5486. This change would end up breaking things again that were just fixed for people like this guy: https://github.com/typeorm/typeorm/issues/4958#issuecomment-594092242.

seandearnaley:
@millsmcilroy are there any guidelines for resolving this RepositoryNotFound message, I understand that the update fixed something for some people but it has broken things for people like me and I can't see how to address it: https://github.com/typeorm/typeorm/issues/5592

millsmcilroy:
@seandearnaley Are you running in a serverless environment? If so, this might be an environment issue and not a code issue, one of which has a solution in this thread https://github.com/typeorm/typeorm/issues/3427#issuecomment-549770047. If not, it would be great to have a reproducible example. Introducing changes like this without tests to back them up just adds more even more fragility to a project as complex as this one.

seandearnaley:
@millsmcilroy , the tests fail with the new version on my ubuntu desktop, I don't really see anything special about my test suite, this must be affecting others-- you can see it for yourself here: https://github.com/seandearnaley/brainstrike-typescript-starter/tree/master/server , just pull and upgrade to .24, it'll start failing with yarn test

seandearnaley:
@millsmcilroy , @YoonSeongKyeong has done some work to resolve this, could you take a look at his code on: https://github.com/typeorm/typeorm/issues/5592#issuecomment-600077109 , I'm not sure if it's a similar issue or not.

OliverLeighC:
My coworker just updated typeorm today and we are getting this same error. Code has not changed at all, in fact my git branch is still on the older typeorm version and I'm not getting this error. We are not running it in a serverless environment.

seandearnaley:
@millsmcilroy , note this issue has been reported by many users now:  https://github.com/typeorm/typeorm/issues/5587 , https://github.com/typeorm/typeorm/issues/5676 , https://github.com/typeorm/typeorm/issues/5592

rpeg:
I just wanted to point out that as of right now, no version of typeorm is compatible with Expo SDK 36. .23 and .24 have the repo bug, .22 fails due to a node std lib dependency, and .21 has an issue with SDK 36's new sqlite API.

YoonSeongKyeong:
```javascript
Object.getPrototypeOf(metadata.target) === Object.getPrototypeOf(target)&&metadata.target.name===target.name
```
This condition is more acceptable. because original PR have crash when it allow the non-related function with same name in accident. The condition only check if target is function and name is the same

davecarlson:
@pleerock - is there an for fixing this? it seems to be stopping a lot of people ( myself included ) from moving past 0.2.22

lmiller1990:
I had this problem when upgrading to Node 14. I downgraded to Node 12 (which is LTS) and it worked.

imnotjames:
Thanks for the contribution!  This is indeed basically a revert of #5486 - and would cause new problems instead even if it would fix this one case.

Given that this change only handles a subset of cases and breaks existing working cases in going to be closing this PR as we work towards a more suitable solution.

imnotjames:
I'm going to rebase this

Issue #5629, opened 2020-03-06
migration script linting warning


apiel:
**Issue type:**

[ ] question
[* ] bug report
[ ] feature request
[ ] documentation issue

**TypeORM version:**

[* ] `latest`
[ ] `@next`
[ ] `0.x.x` (or put your version here)

**Steps to reproduce or a small repository showing the problem:**
Run:
```
npx typeorm migration:create -n Example
```
Will generate:
```
import { MigrationInterface, QueryRunner } from 'typeorm';

export class Example1580910649971 implements MigrationInterface {
    public async up(queryRunner: QueryRunner): Promise<any> {
    }

    public async down(queryRunner: QueryRunner): Promise<any> {
    }
}
```
Why does the method `up` and `down` return `any`? Shouldn't this be a `void`?

When I look at https://github.com/typeorm/typeorm/blob/6cc1a781a7d782701e6fcd3b30fb68296dcefe40/src/migration/MigrationExecutor.ts it doesn't seem that you are using any return value from those method.

The problem is that we always get linting issue in our setup `Unexpected any. Specify a different type.eslint(@typescript-eslint/no-explicit-any)`. Would it be possible to change it to `void`?

apiel:
Issue solved in PR and got merged

PR #5630, opened 2020-03-06
update cli migration up and down from any to void

apiel:
Following the open issue https://github.com/typeorm/typeorm/issues/5629 I created a pull request.

I kept the `migration/MigrationInterface.d.ts` to `Promise<any>` so use-case like `test/github-issues/2693/migration/0000000000003-InsertUser.ts` still work.


pleerock:
thanks!

Issue #2096, opened 2018-05-08
[mysql] Database name isn't read from url


jcudzich:
**Issue type:**

[x] question

**Database system/driver:**

[x] `mysql` / `mariadb`

**TypeORM version:**

[x] `latest`

**Question:**

Hello,

I use mysql url connection string, but I get error:

```Option "database" is not set in your connection options, please define "database" option in your connection options or ormconfig.json```

When i add ```database``` property to connection options everything is fine, but why database name can't be read from url? My url:

```mysql://username:password@host:port/database```

I think that database should be reads from url when it's defined.
What do you think?

**Update:**
I'm not sure, but I think is bug, becouse when I debug connection I see that database name i corectly read from url, but error is throw by ```EntityMetadataValidator``` which not found database options in ```EntityMetadata```.

**Comment by pleerock** [comment]
are you able to dig into source code a bit and find a place with a problem?

**Comment by jcudzich** [comment]
I found that in `src/driver/mysql/MysqlDriver.ts` options property is set in line 242:

```
this.options = connection.options as MysqlConnectionOptions;
```

and in line 248 same file database property is set like this:

```
this.database = this.options.replication ? this.options.replication.master.database : this.options.database;
```

So problem is really simple. To create database property TypeORM use raw options, insted of options built by `DriverUtils.buildDriverOptions()` which has database parsed from url.

Replace above mentioned method to set options property in MysqlDriver solve problem.
After fix, line 242 in `src/driver/mysql/MysqlDriver.ts` is:
```
this.options = DriverUtils.buildDriverOptions(connection.options) as MysqlConnectionOptions;
```
I hope it was helpful.

pleerock:
thanks for the investigations, are you able to provide a PR?

jcudzich:
Sure, just give me some time :)

vlapo:
I think this is no more issue. @jcudzich could you please test this on latest version of typeorm?

nicolasigot:
Hello @vlapo ,
I just found this issue while getting it with version 0.2.11 of typeORM.
I provide URL with this format as explained above and no database as it is within the URL.
Same issue...

vlapo:
Sorry for misunderstanding. I think it is fixed. But now I see we have this kind bug in all drivers. But in MySQL it is checked and throw error. I will check it later

AndKiel:
I think it's not fixed, happening to me on `v0.2.13`.

AlashaHonore:
[...]

PR #5640, opened 2020-03-07
fix: respect database from connection urls

ant45de:
Database names can be defined in the options object. Now also connection urls that contain a database can be used to have the database set in the drivers object.

Closes: #2096

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "typeorm",
-  "version": "0.2.24",
+  "version": "0.2.22",
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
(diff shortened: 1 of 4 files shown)


pleerock:
Cockroachdb is failing because mysql is disabled in there. We need to check in this test if mysql tests were enabled in ormconfig.

ant45de:
Thanks for the hint!
I surrended the connection test with the ormconfig check.

glennreyes:
Can you catch up on this @pleerock? Thanks!

ant45de:
@pleerock can this be merged?

pleerock:
Sorry guys for a late review. Thanks!

PR #5651, opened 2020-03-09
Aurora Data API - Postgres Support

ArsenyYankovsky:
I've been working on Postgress support for the [aurora-data-api driver](https://github.com/ArsenyYankovsky/typeorm-aurora-data-api-driver)
It's currently in the WIP state with most of the basic things working. I believe this PR is all that needed in terms of changes in the TypeORM repository.


pleerock:
Can you please explain me what is the main difference between `postgres and aurora postgres` and `mysql and aurora mysql`? We have so much duplication (and things to maintain) because of what? Thanks!

ArsenyYankovsky:
@pleerock The main difference is the underlying driver these use. The aurora ones use [the driver ](https://github.com/ArsenyYankovsky/typeorm-aurora-data-api-driver)that goes to the [AWS data api](https://docs.aws.amazon.com/AmazonRDS/latest/AuroraUserGuide/data-api.html) which is very useful in serverless environments. 

Ideally I think it should be refactored to just extend the mysql and postgres classes to reduce the complexity. The issue I faced while trying to do that is that connection options have different properties in them so they cannot be easily replaced and their type cannot be overridden in the child class.

Let me know how I should proceed.

filipsuk:
I am planning to use typeorm in lambdas to connect with serverless aurora postgress. What are the limitations of current (mysql?) data api driver when using serverless postgres? Or it is not compatible at all? @ArsenyYankovsky 

Thanks for your amazing work!

ArsenyYankovsky:
@filipsuk Unfortunately, it will generate mysql-specific queries which won't work in postgres version of aurora.

ArsenyYankovsky:
@pleerock I refactored the code a bit to reduce code duplication and make support easier. Unfortunately, I had to put the Aurora driver in the same file as Postgres driver to avoid circular dependency issues. Let me know if that's ok.

pleerock:
Great! Thank you a lot!

ctrlplusb:
Gosh, this is an epic addition! 💜

pleerock:
thanks to @ArsenyYankovsky !

PR #5656, opened 2020-03-10
Added multi-column informations for JoinColumn

PierreTurnbull:
A bit of information I thought was missing.

https://stackoverflow.com/questions/60619027/referenced-column-undefined-was-not-found-in-entity-postsentity


pleerock:
Thanks!

PR #5661, opened 2020-03-11
Next branch changes

Kononnable:
- fix some incorrect merge decisions(master->next)
- remove some old, unused code
- remove unimplemented select query builder methods(orWhere, andWhere with FindOptions parameter)
- precision can no longer be null - cleanup
- fix locking on oracledb
- fix nondeterministic tests
- removal of transaction decorators
- update dependencies
- update typescript version and target ES version
- enable previously skipped tests which are passing now


src/metadata/ColumnMetadata.ts
```diff
@@ -355,15 +355,15 @@ export class ColumnMetadata {
             this.default = options.args.options.default;
         if (options.args.options.onUpdate)
             this.onUpdate = options.args.options.onUpdate;
-        if (options.args.options.scale !== null && options.args.options.scale !== undefined)
```
pleerock: are you sure about these checks? If I remember correctly @AlexMesser were added these two checks explicitly

Kononnable: They were here because we allowed precision to be null previously(to work well with old mysql versions).
As for scale: it is defined as `scale?: number;` so only types allowed by typescript are number and undefined. We don't overwrite it anywhere(casting to any before writing), so we should be good.

Issue #4646, opened 2019-08-27
Add support for temporal (system-versioned) table 


dennis-d:
**Issue type:**

[ ] question
[ ] bug report
[ x ] feature request
[ ] documentation issue

**Database system/driver:**

[ ] `cordova`
[ ] `mongodb`
[ x ] `mssql`
[ x] `mysql` / `mariadb`
[ ] `oracle`
[ ] `postgres`
[ ] `cockroachdb`
[ ] `sqlite`
[ ] `sqljs`
[ ] `react-native`
[ ] `expo`

**TypeORM version:**

[x  ] `latest`
[ ] `@next`
[ ] `0.x.x` (or put your version here)

It would be nice to add support for temporal (SCD) table to typeorm. Ability to created temporal table with migration and make entity away for historical table and query types like `between<time1> and <time2> `, `as of <time> `, `from <time>` , `contained in <time>, <time1>`

gianlazz:
I too would very much like to see this feature!

Travvolta:
Hi there, is there any news on the addition of this functionality?

dennis-d:
We have some working functionality , but it's deferentially doesn't cover everything, but I can create a pull request

Travvolta:
That would be really helpful thank you!

overthemike:
Any chance postgres can be added as well?

gyanprakash48:
@imnotjames By anychance a tenttaive timeline can be shared for this feature to be avaiable ?

DanielSoCra:
This would be a feature that will be needed more and more since for example audits become more and more important. 👍

panuchka:
Also in need of this, so we can replace some of our hacky solutions 👍 Hopefully will be in the next release?

rohaldb:
Any update on this The PR seems to be passing tests but still open

LucasHaddad:
is this being implemented?

PR #5589, opened 2020-02-29
feat: Implement temporal table support for MSSQL

dennis-d:
Add support for migration of system version table for MSSQL. 
Initial take on the implementation 

WIP: #4646

PR #5663, opened 2020-03-11
Feature/implementation of temporal queries to query builder and find methods

Travvolta:
This relates to the work done here https://github.com/typeorm/typeorm/issues/4646 which is the implementation on the actual model / table side of temporal tables.

This pull request is to implement functionality for creating the necassary FOR SYSTEM TIME clauses when building queries through typeorm


Travvolta:
Implementes code from https://github.com/typeorm/typeorm/pull/5589

Travvolta:
Note: the project I was working on no longer requires the system versioning functionality meaning I will not be able to work on this request anywhere near as frequently.

PR #5677, opened 2020-03-13
[sqljs] Provide a default empty array for parameters.

graham-:
The statement bind method in sqljs assumes that either an object or an array has been provided.


atifsyedali:
Hi @pleerock ... can you please merge this? Can't query using latest sqljs@1.2.1 without this fix as it throws an error. See issue #5720.

Thank you!

PR #5680, opened 2020-03-13
refactor: make save method fetch all columns by default

ejose19:
Proposed solution for #3490


ejose19:
The removal of the iterators on  src/persistence/SubjectExecutor.ts were needed as databases without returning were applying transformers twice making test fails when the first transform converts the row into an array

specifically 
```ts
    from (value: string): string[] {
        return value.split(", ");
    }
```
on test/functional/columns/value-transformer/entity/Post.ts

The refactoring on src/query-builder/ReturningResultsEntityUpdator.ts regarding the .select and .addSelect is due to in some cases metadata.targetName was an empty string, making the query fail (again this happened only for databases without returning)

Failing test as stated in comments, are the postgres weird conversion from Date to Datetime on .save(), and for test/functional/persistence/multi-primary-key/multi-primary-key.ts

the category: null only gets bringed up by returning-enabled databases, I was trying many options on transformRawResultsGroup() function without luck, as any change would break another test.

Currently there's 14 test failing, will be updating this as I work however a review is appreciated @pleerock

ejose19:
Uh, postgres date issue is an upstream bug (
see https://github.com/brianc/node-postgres/issues/1844 and https://github.com/brianc/node-pg-types/issues/50) 

For that one either we adjust the test or a patch needs to be done in typeorm (if that's what we want)

ejose19:
@pleerock

src/entity-manager/LiteralEntityManager.ts
```diff
@@ -261,10 +263,11 @@ export function createLiteralEntityManager<Entity>({ connection, queryRunner }:
                 .insert()
                 .into(target)
                 .values(entity)
+                .updateEntity(options?.reload === false ? false : true)
                 .execute();
         },
 
-        update<Entity>(target: EntityTarget<Entity>, criteria: string | string[] | number | number[] | Date | Date[] | ObjectID | ObjectID[] | FindOptionsWhere<Entity>, partialEntity: QueryDeepPartialEntity<Entity>): Promise<UpdateResult> {
+        update<Entity>(target: EntityTarget<Entity>, criteria: string | string[] | number | number[] | Date | Date[] | ObjectID | ObjectID[] | FindOptionsWhere<Entity>, partialEntity: QueryDeepPartialEntity<Entity>, options?: UpdateOptions): Promise<UpdateResult> {
```
pleerock, line 270: `update` method is used to execute `UPDATE` query, it not supposed to perform complex operations, we should't add reload to this method.

src/entity-manager/LiteralEntityManager.ts
```diff
@@ -250,7 +252,7 @@ export function createLiteralEntityManager<Entity>({ connection, queryRunner }:
                 .then(() => entity);
         },
 
-        async insert<Entity>(target: EntityTarget<Entity>, entity: QueryDeepPartialEntity<Entity> | (QueryDeepPartialEntity<Entity>[])): Promise<InsertResult> {
+        async insert<Entity>(target: EntityTarget<Entity>, entity: QueryDeepPartialEntity<Entity> | (QueryDeepPartialEntity<Entity>[]), options?: InsertOptions): Promise<InsertResult> {
```
pleerock, line 255: `insert` method is used to execute `INSERT` query, it not supposed to perform complex operations, we should't add reload to this method.

src/driver/cockroachdb/CockroachDriver.ts
```diff
@@ -583,7 +583,8 @@ export class CockroachDriver implements Driver {
         return Object.keys(insertResult).reduce((map, key) => {
             const column = metadata.findColumnWithDatabaseName(key);
             if (column) {
-                OrmUtils.mergeDeep(map, column.createValueMap(this.prepareHydratedValue(insertResult[key], column)));
+                OrmUtils.mergeDeep(map, column.createValueMap(insertResult[key]));
```
pleerock, line 586: can you please explain this change?

src/driver/sqlserver/SqlServerDriver.ts
```diff
@@ -585,7 +585,8 @@ export class SqlServerDriver implements Driver {
         return Object.keys(insertResult).reduce((map, key) => {
             const column = metadata.findColumnWithDatabaseName(key);
             if (column) {
-                OrmUtils.mergeDeep(map, column.createValueMap(this.prepareHydratedValue(insertResult[key], column)));
+                OrmUtils.mergeDeep(map, column.createValueMap(insertResult[key]));
```
pleerock, line 588: can you please explain this change?

src/entity-manager/LiteralEntityManager.ts
ejose19: Then we have an inconsistency with current behavior (both in master and next) as they are indeed updating if an entity was given:

```ts
  const user = new User();

  user.firstName = "John";
  user.lastName = "Doe";
  user.age = 30;

  await User.insert(user);

  console.log(user.id);

  const user2 = new User();

  user2.firstName = "John";
  user2.lastName = "Doe";
  user2.age = 30;

  await getConnection()
    .createQueryBuilder()
    .insert()
    .into(User)
    .values(user2)
    .execute();

  console.log(user2.id);
```

In that code for example, ids are set, meaning typeorm asked the db for it (either with returning or a new query). If you want insert and update to be simple, that shouldn't be allowed and they should query for additional columns. Or we allow insert/update to keep that behavior with the switch I proposed.

src/driver/sqlserver/SqlServerDriver.ts
ejose19: Changes on drivers were needed to avoid a "double hydration" which was giving issues to columns that couldn't handle it (like custom transforms). All the other drivers didn't had that .prepareHydratedValue and that's how I found why it wasn't working for those 2 drivers.

witalobenicio:
Any updates on this?

pleerock:
@ejose19 still have interest in this PR? If yes, we need to re-think it and most probably open against `master`. We are going to port changes from `next` to `master` step-by-step, it doesn't make sense to merge things into `next` anymore.

pleerock:
Closing for now.

romain130492:
still happening., any updates on this?

DrazekKD:
I bump this topic still exist bug in production build

PR #5690, opened 2020-03-15
fix: redundant undefined parameters are not generated in migration files anymore

olegmdev:
When migration was generated without parameters, it had been adding `undefined` 2nd parameter.

image


pleerock:
Thank you!

Issue #5691, opened 2020-03-15
RelationId is too slow


satanTime:
**Issue type:**

[ ] question
[x] bug report
[ ] feature request
[ ] documentation issue

**Database system/driver:**

[ ] `cordova`
[ ] `mongodb`
[ ] `mssql`
[x] `mysql` / `mariadb`
[ ] `oracle`
[ ] `postgres`
[ ] `cockroachdb`
[ ] `sqlite`
[ ] `sqljs`
[ ] `react-native`
[ ] `expo`

**TypeORM version:**

[x] `latest`
[ ] `@next`
[ ] `0.x.x` (or put your version here)

**Steps to reproduce or a small repository showing the problem:**
`@RelationId()` is too slow on big amount of related records (1k+ rows), see test in the PR.

satanTime:
Hi @pleerock, @Kononnable, could you review and provide your feedback?

satanTime:
Hi, any update on this?

satanTime:
Hi there, any update?

satanTime:
Hi @dhritzkiv, could you review?

satanTime:
Hi @pleerock, any news? I need this fix for out prod.

dhritzkiv:
Btw, if you need this sooner than later, you can point at your own version of Typeorm in a Git repo:

npm install git://github.com/YOURGITUSERNAME/typeorm.git#BRANCH_OR_COMMIT_HASH --save

satanTime:
thanks, I'll try, not sure if it helps with packages where typeorm is a dependency.

satanTime:
Unfortunately it doesn't work because it requires a build:
> ENOENT: no such file or directory, chmod '/Users/michael/projects/amc/backend/node_modules/typeorm/cli.js'

I could do a fork of the repo and so on, but I think the better way to be to add the solution to the main package.

satanTime:
closing based on the comment https://github.com/typeorm/typeorm/pull/5692

satanTime:
I would like to bring this topic back, because there is no interface for 3rd-party plugins in typeorm.

satanTime:
ping @pleerock, might you check this comment: https://github.com/typeorm/typeorm/pull/5692#issuecomment-767077537.

thank you in advance.!

satanTime:
Hi @imnotjames and @pleerock, just a friendly reminder, please take a look when you have time.

Thanks.

nebkat:
[...]

PR #5692, opened 2020-03-15
perf: RelationId is too slow

satanTime:
When we join a lot of relations we can get 1k+ records from db even it is only 10 entities. Then when relationId are loaded the query contains too many duplciates for the same ids. The solution is to check that the relationId query fetches only unique values.

Closes: #5691

### DIFF
--- a/src/query-builder/relation-id/RelationIdLoader.ts
+++ b/src/query-builder/relation-id/RelationIdLoader.ts
@@ -32,17 +32,34 @@ export class RelationIdLoader {
                 if (relationIdAttr.queryBuilderFactory)
                     throw new Error("Additional condition can not be used with ManyToOne or OneToOne owner relations.");
 
+                const duplicates: Array<string> = [];
                 const results = rawEntities.map(rawEntity => {
                     const result: ObjectLiteral = {};
+                    const duplicateParts: Array<string> = [];
                     relationIdAttr.relation.joinColumns.forEach(joinColumn => {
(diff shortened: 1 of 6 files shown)


satanTime:
Hi @pleerock,

the PR has been updated.

satanTime:
Hi @pleerock, could you review the PR?

pleerock:
Thanks for contribution!

`@RelationId` is experimental decorator - it it will be removed in the future versions. To prevent spending time on reviewing and fixing functionality that is going to be removed, I recommend you to create a separate package for this decorator and keep all code in there.

satanTime:
Hi @pleerock, thanks for the info. Then I'll create a separate package as you suggest.

satanTime:
Hi @pleerock,

I've started to work on the package and can't find a way how to integrate a separate package into typeorm flow.

All metadata is stored in `MetadataArgsStorage` and it doesn't allow a way how to add custom info here.
Also I didn't find hooks how to influence on query builder from outside.

In `BaseConnectionOptions` I can't see a way of adding plugins too.

Might you point me to a doc or to an example how to write a plugin / package for typeorm?

Thank you in advance.

satanTime:
Hi @pleerock, I would like to bring this PR back and to find a way to merge it.

I need a solution for GraphQL, but with code first approach.
Unfortunately, prisma and sequelize do not solve it as typeorm does,
but the problem with typeorm is this bug.

What should be done to merge it?
If it is an experimental feature - why not to merge this PR until the feature is gone?

PR #5700, opened 2020-03-17
fix: columns with transformer should be normalized for update

### LINKED ISSUE #2703: Column with transformer is not normalized for update
**Issue type:**

[ ] question
[x] bug report
[ ] feature request
[ ] documentation issue

**Database system/driver:**

[x] all

**TypeORM version:**

[x] `latest`
[ ] `@next`

If you add a transformer to a column which serializes the property value to the desired format, this transformation is not considered in `SubjectChangedColumnsComputer` and the column is always marked as changed if you are working with a complex object type for the column which is serialized to string for example.

```ts
export interface MyComplexType {
  foo: string;
  bar: number;
  xy: Date;
}

export const myTransformer: ValueTransformer = {
  from: dbValue => {
    const parsed = (dbValue || '').split(',');
    return { foo: parsed[0], bar: +parsed[1] || undefined };
  },
  to: entityValue => `${entityValue.foo},${entityValue.bar}`
};

@Entity('foo')
export class Foo {
  @PrimaryGeneratedColumn({ type: 'int' })
  readonly id: number;

  @Column('text', { transformer: myTransformer })
  transformed: MyComplexType;
}
```

Why is `column.getEntityValue()` not called with the transform flag to true in `computeDiffColumns` of `SubjectChangedColumnsComputer`?

PachowStudios:
Normalizes columns with transformers when computing changes.

Fixes #2703

Note: I had to update `TestingOptions` to allow a custom logger in order to write the tests for this. It needs to check the queries performed by the repository.


PachowStudios:
@pleerock Any chance we can get this merged for the next release?

gaghan430:
pls merge this asap

pleerock:
Thank you!

PR #4470, opened 2019-07-22
fix the Excessive stack depth comparing types `FindConditions<?>` and…

mophy:
Fix the this error:

`error TS2321: Excessive stack depth comparing types 'FindConditions<?>' and 'FindConditions<?>'.`

Affected TypeScript versions: >=3.2.0

The fixing is according to this link: https://github.com/microsoft/TypeScript/issues/21592#issuecomment-496723647

PR #5710, opened 2020-03-18
Fix BaseEntity finder methods to allow  lazy relation conditions

bogdan:
The following code:

```
const category = await Category.findOne();
const post = Post.findOne({category})
```

Results in typecheck error when `Post#category` is a lazy relation while the finder does work.


bogdan:
@pleerock any feedback here?

pleerock:
TBH I would like to avoid merging this change, because I want to change lazy relation concept in the future versions.

bogdan:
@pleerock is there any described plan on that? Will there be breaking changes? What is expected date when this issue can be fixed with "lazy relation concent in the future versions"?  

It is causing problems on our end because developers are very confused with this bug.

pleerock:
There is no roadmap, but I was previously mentioned about it in other issues. Okay, let's merge it if its problematic with current version.

src/find-options/FindConditions.ts
```diff
@@ -3,6 +3,9 @@ import {FindOperator} from "./FindOperator";
 /**
  * Used for find operations.
  */
-export type FindConditions<T> = {
-    [P in keyof T]?: T[P] extends never ? FindConditions<T[P]>|FindOperator<FindConditions<T[P]>> : FindConditions<T[P]>|FindOperator<FindConditions<T[P]>>;
+type FindCondition<T> = FindConditions<T> | FindOperator<FindConditions<T>>;
+export type FindConditions<T> = T | {
```
pleerock, line 7: I don't think I remember why we had `T[P] extends never`, but since we have it most probably we had some issue. But why did you add `T | ` ? Isn't second part already cover T?

bogdan: `T[P] extends never` is to workaround the typescript bug: https://github.com/typeorm/typeorm/pull/4470.

`T | ` is required because of weird behavior of `{[P keyof T]}` for generics. Example:

```
type A<T> = {[P in keyof T]: T[P]}

type B = A<string>;
```

The computed type of `B` is `string` which is counter intuitive.
So such magic doesn't work for non-generic type which will always result in recursion types call in previous implementation. So I can not write the condition like: `{category: category}` becuase `category` is not generic. The thing I can only do without `T |` is the following `{category: {id: category.id}}` which is working btw. Let me know if you want such query to also be added to the test if it is not tested yet in other place.

pleerock: hmm I didn't understand. How "string" is related to `{category: category}`? And I don't understand how `{category: category}` wouldn't work. Can you send a small typescript playground example?

bogdan: Sure: https://www.typescriptlang.org/play/?ssl=28&ssc=1&pln=29&pc=1#code/PTAEGUEsFsAcBtIDNIFMAmoBGBPUAnVaAewDdIA7Ac1ADFL0B5WVfAQwBdj8BYAKH4ccLOgwDCxCukgdIkgM4AeACoA+UAF5QAb36h9oANoAFUJVABrVDmJJQygLoB+AFyipEqTLkUlykw6qANz8AL4hApGoAB6w3BygQiL0HpLSsgoAcqgA7irqWsqgAD46egYmZhSW1rb2zm7+xg6gMRyoUvKgxvjE0JDyqIqUSKygAKqq5fpO7uie6T7y2XmT06BuKfNp3lm5KgHBYfwnfADG8GzyXWKcqFTceLp8BmbobhQArtBYY1oAjBFXrA2IQKBw3D0+gMhrd2g98HhSp8pKgUBQMAVur1+oMAHSEeTEeCkVAAChR6DRlAwAEp1vIOJxIGdQOj0GTSGx4J9UJtxDsMr5FHD7o9VLSyi8DKEGUzZKz2StOdzefzUl4hct9qKETgJTpQLK+Mb+GcFAkQWCEloMTlQLrHmTaRF+I7EXj2WTtKArR0EqF6XwQPYABaoAioK6SRLh0CIDGgNhYMgR9DEVDyCgAcgS5rgkHgEYG2FQZzYn0G-BD7rwxCwACsywkSxxCHdMIgrL7LuZ602zi3fJAqXMFrtfKBq2A2FJQFTzZ8EBgqlwqtJyOhPtzfb0WPhZJnQDkZKHQAADUzmKw2OzKc-xyDdokuN13PWehjev3gtzaEduD+HB4iOAA0vqgv6gGQeCeJAaEganLWn5SMq2jwbSQA

test/other-issues/lazy-relations-loading-via-base-entitiy-finders/index.ts
```diff
@@ -0,0 +1,33 @@
+import {Post} from "./entity/Post";
+import {Category} from "./entity/Category";
+import "reflect-metadata";
+import {closeTestingConnections, createTestingConnections, reloadTestingDatabases} from "../../utils/test-utils";
+import {Connection} from "../../../src/connection/Connection";
+import {expect} from "chai";
+import {PromiseUtils} from "../../../src";
+
+describe("lazy-relations-loading-via-base-entity-finders", () => {
```
imnotjames, line 9: Can this love in the standard functional tests instead of in this part of the tests

src/find-options/FindConditions.ts
imnotjames: Can you add that to the rest just in case - or ensure its in the right place elsewhere in the codebase.

test/other-issues/lazy-relations-loading-via-base-entitiy-finders/index.ts
bogdan: I am not sure what do you mean here. Can you explain?

bogdan:
@imnotjames fixed the issue. Can you please review again?

bogdan:
@pleerock @imnotjames can you please merge this one?

alumni:
@bogdan This PR broke `find`/`findOne` when used from `EntityManager`.

```ts
const p = entityManager.findOne(Post, { where: { id: 1 }});
```

Under TS 4.4.4, the inferred type of `p` is `{ id: number }` instead of `Post`

bogdan:
@alumni I can not reproduce that on current master. For me the type of `p` is `Post | undefined`.

alumni:
@bogdan I check a bit more and it seems it's not particularly because of this PR, it only happens more often now, probably because of the added `T | ...` in line 11. We were already explicitly specifying the generic type in many cases (`post = entityManager.find<Post>(Post, ...)`) to avoid this problem. Probably if I find a simpler example than what I have now, I would also have a solution to it.

brandon-leapyear:
Note: the `T | ...` addition causes a regression: https://github.com/typeorm/typeorm/issues/8600

Issue #4980, opened 2019-10-28
onUpdate: 'CASCADE' doesn't work on many-to-many relation


stepanh:
**Issue type:**

[ ] question
[x] bug report
[ ] feature request
[ ] documentation issue

**Database system/driver:**

[ ] `cordova`
[ ] `mongodb`
[ ] `mssql`
[ ] `mysql` / `mariadb`
[ ] `oracle`
[x] `postgres`
[ ] `cockroachdb`
[ ] `sqlite`
[ ] `sqljs`
[ ] `react-native`
[ ] `expo`

**TypeORM version:**

[x] `latest`
[ ] `@next`
[x] `0.2.20`

**Steps to reproduce or a small repository showing the problem:**

`onUpdate: 'CASCADE',` works well on `many-to-one` relations, however, on `many-to-many` it seems to have no effect.
```ts
export const SampleEntity = new EntitySchema<Sample>({
  name: 'sample',
  columns: {
    ...BaseColumns,
    id: {
      type: String,
      primary: true,
    },
  },
  relations: {
    company: {
      type: 'many-to-one',
      target: 'company',
      onUpdate: 'CASCADE', // works, foreign key shows "ON UPDATE CASCADE"
    },
    owners: {
      type: 'many-to-many',
      target: 'user',
      joinTable: true,
      cascade: true,
      onUpdate: 'CASCADE', // doesn't work, junction table foreign key shows "ON UPDATE NO ACTION"
    },
  }
});

```

michaelbazos:
Encountered the same issue.

```ts
  @JoinTable()
  @ManyToMany(type => Employee, {
    onUpdate: 'CASCADE',        // query log shows "ON UPDATE NO ACTION" no matter the value
    onDelete: 'CASCADE'          // property is correctly picked up
  })
  employees: Employee[];
```

DiCrEn:
Same here with mysql too

AndresSp:
Same here... Someone could attend this?

fan-tom:
Is this because of here https://github.com/typeorm/typeorm/blob/master/src/metadata-builder/JunctionEntityMetadataBuilder.ts#L146
`onUpdate: relation.unUpdate || 'CASCADE'` is missed?

fan-tom:
[...]

Issue #3505, opened 2019-01-24
Allow different onDelete behavior for inverse side of ManyToMany


csuich2:
**Issue type:**

[ ] question
[ ] bug report
[x] feature request
[ ] documentation issue

**Database system/driver:**

[ ] `cordova`
[ ] `mongodb`
[ ] `mssql`
[ ] `mysql` / `mariadb`
[ ] `oracle`
[ ] `postgres`
[ ] `sqlite`
[ ] `sqljs`
[ ] `react-native`
[ ] `expo`

**TypeORM version:**

[ ] `latest`
[x] `@next`
[ ] `0.x.x` (or put your version here)

**Steps to reproduce or a small repository showing the problem:**

ManyToMany relationships currently use the same onDelete option for both directions of the join table. It would be nice to be able to specify a different onDelete option for the inverse side of a ManyToMany relation.

For example: If I have ```Post``` which have a ManyToMany relationship with ```Category```, I may want the deletion of a ```Post``` to have an no action on delete, while deletion of a ```Category``` would cascade delete all ```Post```s which have that category.

I believe this could be fixed an ```onDeleteInverse``` option to ```JoinTableOptions``` and ```JoinTableMultipleColumnsOptions``` so that ```JunctionEntityMetadataBuilder.build()``` can create the foreign keys with two different ```ON DELETE``` behaviors.

pleerock:
We need a PR for this feature.

csuich2:
I'm happy to open the PR, I just wanted to talk through the solution with you before opening it. Do you think this is the right approach?

pleerock:
currently we have no action onDelete right?

csuich2:
Yeah, JoinColumn has no onDelete property. I believe ```JunctionEntityMetadataBuilder.build)``` uses the onDelete option from one side of the relation metadata. I'm not sure which, but I assume the side with the JoinColumn on it.

pleerock:
no, it shouldn't use any other's onDelete.

csuich2:
I'm on @next, so my line numbers might be a bit different, but this is what I've found so far:

```JunctionEntityMetadataBuilder``` line 130:
``` typescript
        // create junction table foreign keys
        entityMetadata.foreignKeys = [
            new ForeignKeyMetadata({
                entityMetadata: entityMetadata,
                referencedEntityMetadata: relation.entityMetadata,
                columns: junctionColumns,
                referencedColumns: referencedColumns,
[...]

PR #5714, opened 2020-03-18
fix: pass `ManyToMany` `onUpdate` option to foreign key metadata

fan-tom:
Closes: #4980

I'm not sure if onDelete/onUpdate options are handled properly, it seems to me that they must be applied to foreign keys pointing to OPPOSITE entity, not to entity they are used in, but such changes break existing tests, so I considered to fix this issue not breaking them.

### DIFF
--- a/src/metadata-builder/JunctionEntityMetadataBuilder.ts
+++ b/src/metadata-builder/JunctionEntityMetadataBuilder.ts
@@ -148,14 +148,16 @@ export class JunctionEntityMetadataBuilder {
                 referencedEntityMetadata: relation.entityMetadata,
                 columns: junctionColumns,
                 referencedColumns: referencedColumns,
-                onDelete: relation.onDelete || "CASCADE"
+                onDelete: relation.onDelete || "CASCADE",
+                onUpdate: relation.onUpdate || "CASCADE"
             }),
             new ForeignKeyMetadata({
                 entityMetadata: entityMetadata,
                 referencedEntityMetadata: relation.inverseEntityMetadata,
                 columns: inverseJunctionColumns,
                 referencedColumns: inverseReferencedColumns,
-                onDelete: relation.onDelete || "CASCADE"
+                onDelete: relation.inverseRelation ? relation.inverseRelation.onDelete : "CASCADE",
+                onUpdate: relation.inverseRelation ? relation.inverseRelation.onUpdate : "CASCADE",
             }),
         ] : [];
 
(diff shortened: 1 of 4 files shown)


pleerock:
Based on your comments what I understood is that we need to set `onUpdate` and `onDelete` for every column in the junction table separately, am I right? If so, this temporary change doesn't make sense and we probably need to remove this option and introduce it for `@JoinTable`'s `joinColumn` option?:

```ts
@ManyToMany(
        () => Book,
        book => book.authors
    )
    @JoinTable({
        name: "author_to_books",
        joinColumn: {
            name: "author_id",
            referencedColumnName: "id",
            onDelete: "CASCADE",
            onUpdate: "CASCADE"
        },
        inverseJoinColumn: {
            name: "book_id",
            referencedColumnName: "id",
            onDelete: "CASCADE",
            onUpdate: "CASCADE"
        },
    })
``` 

Also, before any change will be merged we need to be sure nothing will be broken for exist users and ORM wont generate any new queries for those who update the version. I'm a bit scary to merge such kind of things.

fan-tom:
I just want to say that when I write
```typescript
class Author {
  @ManyToMany(type => Book, {
    onUpdate: 'CASCADE',
    onDelete: 'CASCADE',
  })
  books: Book[];
}

class Book {
  @JoinTable()
  @ManyToMany(type => Author, {
    onUpdate: 'NO ACTION',
    onDelete: 'NO ACTION',
  })
  authors: Author[];
}
```
I expect that I can update id's of books and get fk updated accordingly, but cannot update author's id's, like when you use ManyToOne and OneToOne, you apply those decorators inside referencing table and onDelete/onUpdate options are applied to links to REFERENCED table. But in case of ManyToMany when I set `onDelete: 'CASCADE'` inside `Author` entity, I get it applied to link to CURRENT table(Author), instead of link to REFERENCED table(Book).

What about affecting current users, I just added missing property handling code for `onUpdate` option, that works the same way `onDelete` works currently, just like people requested in issue, and fixed the bug that single `onDelete` option was used for both sides, so I don't understand what things you are scary about.

fan-tom:
Is it related issue? https://github.com/typeorm/typeorm/issues/3505

imnotjames (commented):
This makes sense.  Can you rebase and get the new test suite kicking off?

fabianscheidt:
I am facing both of the issues that are mentioned here. Wouldn't it make sense to merge @fan-tom's suggestion now and later come up with a change that involves the `@JoinTable` decorator as suggested by @pleerock?

Ami777:
Why isn't this merged... 1 year later :|?

jcrben:
@fan-tom up for rebasing for a merge as suggested by @imnotjames?

fan-tom:
rebased

AlexMesser:
thank you for contribution!

PR #5728, opened 2020-03-20
Fix sha.js import

guybedford:
When using this pattern in the browser it will throw because namespace objects can never be functions, which also throws in Rollup even if it might work in other build tools.

Using the default import form ensures the instance is the exact `module.exports` instance provided by sha.js.


pleerock:
I even won't think to review these changes by a modules expert. Thanks!

PR #5737, opened 2020-03-21
Names of extra columns for specific tree types moved to NamingStrategy

steerio:
This change makes column names used by the nested set and materialized path tree types configurable through the naming strategy, keeping the ones already in use as defaults.

The rationale is that some of us have to work with legacy databases accessed by other tools as well, in which case altering column names is off the table.


pleerock:
This is a bit breaking for people who create their own strategies and didn't extend DefaultStrategy.

Issue #5734, opened 2020-03-21
EntityManager.insert(AnyEntity, []) crashes, should be no-op


lovrop:
**Issue type:**

[x] bug report

**Database system/driver:**

[x] `postgres`
[x] `sqlite`

**TypeORM version:**

[x] `latest`

**Steps to reproduce or a small repository showing the problem:**

Either of these calls fails for any/most entity types:

  - `entityManager.insert(AnyEntity, [])`
  - `anyRepository.insert([])`

I first encountered this when I found my subscriber was getting invoked with an empty array where it of course expects an entity. Even in the absence of a subscriber these calls likely fail because the empty array is let through as an entity by TypeORM, resulting in carnage. :)

For example, we can add this trivial test to the test suite:

```diff
$ git diff -u
diff --git i/test/functional/query-builder/insert/query-builder-insert.ts w/test/functional/query-builder/insert/query-builder-insert.ts
index 118cb37f..f6f3ebe6 100644
--- i/test/functional/query-builder/insert/query-builder-insert.ts
+++ w/test/functional/query-builder/insert/query-builder-insert.ts
@@ -141,4 +141,8 @@ describe("query builder > insert", () => {

     })));

+    it("Repository.insert should not crash on an empty array", () => Promise.all(connections.map(async connection => {
+        await connection.createEntityManager().insert(User, []);
+    })));
+
 });

$ npm test -- --grep="empty array"
> typeorm@0.2.24 test /Users/lovro/repos/third-party/typeorm
> rimraf ./build && tsc && mocha --file ./build/compiled/test/utils/test-setup.js --bail --recursive --timeout 60000 ./build/compiled/test "--grep=empty array"

  query builder > insert
    1) Repository.insert should not crash on an empty array

  0 passing (189ms)
  1 failing

  1) query builder > insert
       Repository.insert should not crash on an empty array:
     QueryFailedError: SQLITE_CONSTRAINT: NOT NULL constraint failed: user.name
      at new QueryFailedError (src/error/QueryFailedError.ts:9:9)
      at Statement.handler (src/driver/sqlite/SqliteQueryRunner.ts:53:26)

npm ERR! Test failed.  See above for more details.
```

I expect inserting an empty array of entities to be a no-op. Indeed it is so for `save()` so `insert()` is inconsistent.

lovrop:
I found special handling for this in EntityManager.save() so I believe the correct fix is to do the same thing in EntityManager.insert(). I will prepare a PR.

PR #5745, opened 2020-03-23
fix: calling EntityManager.insert() with an empty array of entities

lovrop:
As described in issue #5734, the current behaviour seems like an oversight and inconsistent with save() and remove() which already have this handled as a special case.

Test Plan: npm test, and more specifically npm test -- --grep='#5734'

Closes: #5734


lovrop:
@pleerock mind looking at this simple bugfix? Thanks!

pleerock:
Thanks for contribution! Sorry for such a late review!

PR #5754, opened 2020-03-24
Aurora Data API - Support for AWS configuration options through aurora driver

cklam2:
The underlying AWS RDS database, accessible through the aurora-data-api driver supports configurations settings, also known as [AWS.ConfigurationOptions](https://docs.aws.amazon.com/AWSJavaScriptSDK/latest/AWS/RDSDataService.html#constructor-property). It allows you to pass options like max number of retries, Arn regions and impersonation secrets.

The [typeorm-aurora-data-api-driver](https://github.com/ArsenyYankovsky/typeorm-aurora-data-api-driver), which is referenced in this project has now added support for this, see pull request state [here](https://github.com/ArsenyYankovsky/typeorm-aurora-data-api-driver/pull/31). Minimum required version is 1.2.0. 

Changes included in this pull request  allows users of TypeORM to pass the values through AuroraDataApiConnectionOptions. Example:

```
createConnection({
    name,
    type: 'aurora-data-api',
    ...
    serviceConfigOptions: {
        credentials,
        maxRetries: 3,
        retryDelayOptions: {
           base: 5000,
        }
    }
  });
```

### DIFF
--- a/ormconfig.json.dist
+++ b/ormconfig.json.dist
@@ -92,7 +92,7 @@
     "useUnifiedTopology": true
   },
   {
-    "skip": false,
+    "skip": true,
     "name": "aurora-data-api",
     "type": "aurora-data-api",
     "region": "us-east-1",
--- a/ormconfig.json.dist
+++ b/ormconfig.json.dist
@@ -90,5 +90,18 @@
     "logging": false,
     "useNewUrlParser": true,
     "useUnifiedTopology": true
+  },
+  {
+    "skip": false,
+    "name": "aurora-data-api",
+    "type": "aurora-data-api",
+    "region": "us-east-1",
+    "secretArn": "mysecret",
+    "resourceArn": "app-dbcluster",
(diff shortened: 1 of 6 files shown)


filipsuk:
this would be really useful for me 🎉 as I need to set data api endpoint to be able to test locally with https://github.com/koxudaxi/local-data-api/

filipsuk:
Probably found a dirty workaround before this is merged.

```ts
import AWS from 'aws-sdk';

// Run this before creating the connection
AWS.config.update({ retryDelayOptions: { base: 1000 }, maxRetries: 60 });
```

cklam2:
It's a workaround but I believe those settings are global so will affect all service requests. Be careful with this.

pleerock:
Thanks!

webMan1:
@cklam2 This did _not_ add support for the 'aurora-data-api-pg' driver, correct? Is there something that remains to be done to make that happen?

Issue #5762, opened 2020-03-25
Using URL as a rich column type breaks


dhritzkiv:
**Issue type:**

[x] bug report

**Database system/driver:**

[x] `postgres`

**TypeORM version:**

[x] `latest`

**Steps to reproduce or a small repository showing the problem:**

```ts
// ./Doc.entity.ts
@Entity()
export class Doc {
	@PrimaryGeneratedColumn("uuid")
	id: string;

	@Column("varchar", {
		length: 2000,
		nullable: true,
		// marshall 
		transformer: {
			from(value: string | null): URL | null {
				return value !== null ? new URL(value) : value;
			},
			to(value: URL | null): string | null {
				return value?.toString() || null;
			},
		},
	})
	url: URL | null;
}

// elsewhere…

const doc = new Doc();
doc.url = new URL("https://typeorm.io/")

await repository.save(doc); // <- throws here with `TypeError: Cannot set property 'flags' of undefined`
```

The stack trace suggests that Node internals for URL are throwing here 
https://github.com/nodejs/node/blob/2fa74e3e38bb028339e48763138456b3ed10ed97/lib/internal/url.js#L240 (`ctx` is undefined?)
 after `href#set` is called: https://github.com/nodejs/node/blob/2fa74e3e38bb028339e48763138456b3ed10ed97/lib/internal/url.js#L458

which is triggered from the TypeORM library here:
https://github.com/typeorm/typeorm/blob/685d76b20464388c6f86c7d268770577452124b6/src/util/OrmUtils.ts#L90

I've previously employed richer types + transformers for other data types flawlessly. However, this data type (URL) fails before it even gets to the transformers to Marshall to the database.

I'm not familiar enough with the inner workings of how TypeORM serializes its data to break this problem down. Is there something about the URL data type specifically that prevents it from working as expected?

In the meantime, I can work with the data type as a string, and transform it to a URL manually.

dhritzkiv:
Due to React Native's lack of `URL` (see #6142), a simple import of `URL` and `instanceof` check against it won't work as in #5771.

There exist URL polyfills such as [whatwg-url](https://github.com/jsdom/whatwg-url) and [react-native-url-polyfill](https://github.com/charpeni/react-native-url-polyfill), but these are heavy-weight and over-kill.

Ultimately, the fix doesn't need to do anything crazy with `URL`. It only needs to know not to treat it as a simple enumerable object, and cast the value to a string for serialization and equality comparison.

[...]

PR #5771, opened 2020-03-26
Handle URL objects as column field values

dhritzkiv:
Addresses #5762


pleerock:
I have to revert this one because of #6142 . We'll need to figure out some different solutions to such kind of problems.

dhritzkiv:
@pleerock should I re-open #5762, or create a new issue?

pleerock:
I have reopened it.

PR #5776, opened 2020-03-28
Update Entity decorator return type to ClassDecorator

GrayStrider:
I've been making a class decorator composer function, and Function type is not specific enough. I'll augment it locally for the time being.

I suppose some other types would have to be updated as well, hopefully there's a CI in place


GrayStrider:
So far can't augment the types, probably due to typeorm using `declare` instead of proper interfaces

PR #5779, opened 2020-03-29
Add MySQL REGEXP

SalemCode8:
Adding MySQL REGEXP Operator

### DIFF
--- a/docs/find-options.md
+++ b/docs/find-options.md
@@ -23,7 +23,7 @@ userRepository.find({ relations: ["profile", "photos", "videos", "videos.video_a
 * `join` - joins needs to be performed for the entity. Extended version of "relations".
 
 ```typescript
-userRepository.find({ 
+userRepository.find({
     join: {
         alias: "user",
         leftJoinAndSelect: {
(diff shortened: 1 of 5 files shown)


docs/find-options.md
```diff
@@ -23,7 +23,7 @@ userRepository.find({ relations: ["profile", "photos", "videos", "videos.video_a
 * `join` - joins needs to be performed for the entity. Extended version of "relations".
 
 ```typescript
-userRepository.find({ 
+userRepository.find({
```
klevamane: This change shouldn' t be added. The same goes for others in this PR

SalemCode8: You can remove what you  see isn't necessary

src/find-options/FindOperator.ts
```diff
@@ -112,6 +112,8 @@ export class FindOperator<T> {
                 return `${aliasPath} = ANY(${parameters[0]})`;
             case "isNull":
                 return `${aliasPath} IS NULL`;
+            case "Regexp":
```
pleerock, line 115: to make it consistent, here "Regexp" should be "regexp"

pleerock:
Thank for contribution! This PR needs a test before it can be merged.

SalemCode8:
> Thank for contribution! This PR needs a test before it can be merged.

Yeah, Sure

imnotjames:
We haven't seen any activity on this for a while, there aren't tests, and now have a few conflicts.  I'm going to close this but as soon as you have more time to work on fixing the issues, please open a PR and we'll be gracious to review it again!

For the time being, you can use the new features with `Raw` to pass parameters to that instead!

```
const Regexp = (regex: string) => Raw(alias => `${alias} REGEXP :regex`, { regex });
```

PR #5786, opened 2020-03-31
feat: use postgres 10 identity constraint

jaschaephraim:
Old versions of Postgres used SERIAL pseudo-type for auto-incrementing values.
Versions >=10 use SQL standard-conforming GENERATED AS IDENTITY.
This allows direct insertion of values into PrimaryGeneratedColumn using Postgres.

BREAKING CHANGE: breaks compatibility with postgres versions <10
Closes: #2215


jaschaephraim:
Note: This PR includes updated `docker-compose` using Postgres 10. The current `docker-compose` using Postgres 9 will prevent CI from passing.

imnotjames:
Can this be rebased?

src/driver/postgres/PostgresQueryRunner.ts
```diff
@@ -634,21 +634,6 @@ export class PostgresQueryRunner extends BaseQueryRunner implements QueryRunner
                     downQueries.push(new Query(`ALTER TABLE ${this.escapePath(table)} RENAME CONSTRAINT "${newPkName}" TO "${oldPkName}"`));
                 }
 
-                // rename column sequence
```
imnotjames, line 637: Wouldn't dropping a lot of this code be bad because we'd never undo it and remove the sequence?

Like if someone upgrades to this version and have an existing database wouldn't they be stuck & the migration code would fail?

How do we migrate from the v9 compatible code to newer code?

src/driver/postgres/PostgresQueryRunner.ts
```diff
@@ -826,20 +811,13 @@ export class PostgresQueryRunner extends BaseQueryRunner implements QueryRunner
...
 
-            if (oldColumn.isGenerated !== newColumn.isGenerated && newColumn.generationStrategy !== "uuid") {
-                if (newColumn.isGenerated === true) {
-                    upQueries.push(new Query(`CREATE SEQUENCE ${this.buildSequenceName(table, newColumn)} OWNED BY ${this.escapePath(table)}."${newColumn.name}"`));
-                    downQueries.push(new Query(`DROP SEQUENCE ${this.buildSequenceName(table, newColumn)}`));
-
-                    upQueries.push(new Query(`ALTER TABLE ${this.escapePath(table)} ALTER COLUMN "${newColumn.name}" SET DEFAULT nextval('${this.buildSequenceName(table, newColumn, undefined, true)}')`));
-                    downQueries.push(new Query(`ALTER TABLE ${this.escapePath(table)} ALTER COLUMN "${newColumn.name}" DROP DEFAULT`));
-
+            if (newColumn.isGenerated !== oldColumn.isGenerated) {
```
imnotjames, line 814: Is it possible to keep track of the version of postgres we're operating against & choose to use this new technique on newer versions of postgres?

imnotjames (changes requested):
Dropping support for postgres 9 is really difficult to do here - unless we can continue to support postgres 9 I don't think we can move forward with this change.

PR #5789, opened 2020-03-31
Prevent TypeError when calling bind function with sql.js 1.2.X

fb64:
for information check discussion on https://github.com/sql-js/sql.js/pull/375


pleerock:
Thanks! This isn't test-covered however, so I strongly recommend to create a separate PR with a test-case to prevent this issue in the future.

PR #5805, opened 2020-04-04
Add Schema for OracleDriver

pirouet:
A bug exists that causes he schema name to be skipped when using Oracle. This fix puts it back in


pirouet:
Hi - I've been meaning to catch up on this. Is anyone able to shed light on this why would have been closed with no reason as to why, please?

pleerock:
I'm not sure, it looks like you have some bot installed that automatically closes your pull requests or branches that stay for too long?

PR #5828, opened 2020-04-08
Add ILike FindOperator for case insensitive LIKE

brussee:
(depends on support by database driver)

### DIFF
--- a/docs/find-options.md
+++ b/docs/find-options.md
@@ -268,6 +268,22 @@ will execute following query:
 SELECT * FROM "post" WHERE "title" LIKE '%out #%'
 ```
 
+* `ILike`
+
+```ts
+import {ILike} from "typeorm";
+
+const loadedPosts = await connection.getRepository(Post).find({
+    title: ILike("%out #%")
+});
+```
+
+will execute following query:
+
+```sql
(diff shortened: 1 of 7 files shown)


brussee:
Small contribution, which takes care of the case insensitive LIKE use case.
Solution, until the intended refactor is done at some point: https://github.com/typeorm/typeorm/pull/5574#issuecomment-592302691

brussee:
Fixes #4418

brussee:
Here's a gist for anyone using https://www.npmjs.com/package/patch-package
https://gist.github.com/brussee/d66ebabe8556168b69e61c5420491318

pleerock:
We already have this functionality on `next` branch. There is a failing test - you probably missed the point to exclude unsupported databases.

pleerock:
Also, can you please make sure there is minimal amount of differences with `next` branch? To simplify merge to the `next` branch. Thanks!

brussee:
> ... - you probably missed the point to exclude unsupported databases.

Yes, I don't know how to do that.

> Also, can you please make sure there is minimal amount of differences with `next` branch?

I've just checked and there are no differences on the `next` branch regarding this patch, so I'll close this.

@pleerock Sorry for the delay and thanks for your replies!

PR #5847, opened 2020-04-10
Oracle 10g-11g to support limit offset Some fixes after test

vmsh:
I adapted typeorm for Oracle 10g-11g to support limit offset for my old projects

### DIFF
--- a/src/driver/Driver.ts
+++ b/src/driver/Driver.ts
@@ -27,6 +27,11 @@ export interface Driver {
     database?: string;
 
     /**
+     * Server version (for Oracle)
+     */
+    serverversion: string;
+
+    /**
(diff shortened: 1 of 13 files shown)


pleerock:
I'm sorry but old Oracle versions are too problematic, and the problem you are trying to fix is just one of them. We don't have plan to support it, probably its easier for you to upgrade your Oracle version or keep maintaining your functionality on your own fork.

vmsh:
> I'm sorry but old Oracle versions are too problematic, and the problem you are trying to fix is just one of them. We don't have plan to support it, probably its easier for you to upgrade your Oracle version or keep maintaining your functionality on your own fork.

PR #5854, opened 2020-04-11
Feature: Add soft remove and recover methods to entity

jannostern:
Add soft remove and recover methods to entity.
This PR improves https://github.com/typeorm/typeorm/pull/5034

### DIFF
--- a/src/repository/BaseEntity.ts
+++ b/src/repository/BaseEntity.ts
@@ -58,6 +58,20 @@ export class BaseEntity {
     }
 
     /**
+     * Records the delete date of current entity.
+     */
+    softRemove(options?: SaveOptions): Promise<this> {
+        return (this.constructor as any).getRepository().softRemove(this, options);
+    }
+
+    /**
+     * Recovers a given entity in the database.
+     */
+    recover(options?: SaveOptions): Promise<this> {
+        return (this.constructor as any).getRepository().recover(this, options);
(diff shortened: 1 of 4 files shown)


pleerock:
Looks like tests are failing

jannostern:
@pleerock do you have any recommendations why? 
When I run them locally, they seem to pass (I have tested only with postgres though).

The `entity-soft-remove.ts` test is basically same as `repository-soft-remove.ts` test (which is already merged to master and is located in the same directory), with only couple of minor modifications.

I wonder if this could be a database engine specific issue? I can't see that from the CircleCI output.
But I could debug this further, if I knew the the database engine to use, so I could reproduce the failing test.

pleerock:
this shouldn't be database specific... I mostly think you missed something in your test, but I you say that locally you have everything working with postgres, maybe you'll need to run all these tests for all dbs. You just need to run `docker-compose up` and you'll have all dbs on your local.

jannostern:
@pleerock thank you for the input. 
I ran the tests using the docker database instances and found the issue.

I had not noticed that I needed to use `PromiseUtils.runInSequence` to go through the database connections in sequence.

Also I wanted to say, that you have done great work with TypeORM. 
I just recently started to use it, and I really love it.

pleerock:
Thank you!

Issue #3365, opened 2019-01-04
Repository update does not respect readonly flag on CreateDateColumn or UpdateDateColumn


jdforsythe:
**Issue type:**

[ ] question
[x] bug report
[ ] feature request
[ ] documentation issue

**Database system/driver:**

[ ] `cordova`
[ ] `mongodb`
[ ] `mssql`
[ ] `mysql` / `mariadb`
[ ] `oracle`
[x] `postgres`
[ ] `sqlite`
[ ] `sqljs`
[ ] `react-native`
[ ] `expo`

**TypeORM version:**

[ ] `latest`
[ ] `@next`
[x] `0.2.11` (or put your version here)

**Related to:**

#3271

**Steps to reproduce or a small repository showing the problem:**

I recently switched our update method to use `repository.update()` instead of `repository.save()` to prevent the needless extra `SELECT` before `UPDATE`.

**pizza-route.ts**
```ts
@JsonController('/pizzas')
export class PizzaController {
  private readonly pizzaRepository: Repository<Pizza>;

  constructor() {
    const connection: Connection = getConnectionManager().get();
    this.pizzaRepository = connection.getRepository<Pizza>(Pizza);
  }

  @Put('/:id')
  async put(@Param('id') id: number, @EntityFromBody({ required: true }) pizza: Pizza) {
    /* before */
    // return this.pizzaRepository.save(pizza);

    /* after */
    return this.pizzaRepository.update(id, pizza).then((res: UpdateResult) => {
      console.log(res); // always empty { generatedMaps: [], raw: [] }
      return pizza; // can't return updated DB object
   });
  }
}
```

I was surprised by two things:

1. It started throwing this same error from #3271:

  `multiple assignments to same column \"updatedAt\"`. 

  I thought adding the `readonly` flag:

  ```ts
  @UpdateDateColumn({ readonly: true }) updatedAt: Date;
  ```

  would fix the problem, but the `repository.update()` method appears to ignore the `readonly` flag for the `UpdateDateColumn()` and `CreateDateColumn()`.

2. The `repository.update()` method does not give access to the updated DB value like `repository.save()` does. I expected to see the query like:

  ```sql
  UPDATE "pizza" SET "name" = $2, "updatedAt" = CURRENT_TIMESTAMP WHERE "id" = $1 RETURNING *
  ```

  but the query was

  ```sql
[...]

Issue #3271, opened 2018-12-17
@CreateDateColumn and @UpdateDateColumn should use value if provided.


wederer:
**Issue type:**

[ ] question
[x] bug report
[ ] feature request
[ ] documentation issue

**Database system/driver:**

[ ] `cordova`
[ ] `mongodb`
[ ] `mssql`
[ ] `mysql` / `mariadb`
[ ] `oracle`
[x] `postgres`
[ ] `sqlite`
[ ] `sqljs`
[ ] `react-native`
[ ] `expo`

**TypeORM version:**

[x] `latest`
[ ] `@next`
[ ] `0.x.x` (or put your version here)

**Steps to reproduce or a small repository showing the problem:**
Entity User:
```
    @PrimaryColumn({
        nullable: false,
        primary: true,
    })
    id: number;

     @CreateDateColumn({
        nullable: false,
        name: "dt_create"
    })
    createdOn: Date;

    @UpdateDateColumn({
        nullable: false,
        name: "dt_modified"
    })
    modifiedOn: Date;
```
And then run:
```typescript
    const date = new Date(2017, 11, 24, 12);
    const user = new User();
    user.id = 1;
    user.modifiedOn = date;
    user.createdOn= date;

    getRepository(User).insert(user);
    getRepository(User).update(user);
```

The behaviour of `save`, `update` and `insert` are pretty strange when it comes the the `@CreateDateColumn` and `@UpdateDateColumn` decorators.

When running the above code and a user with the same id is already present following error is thrown:
```
multiple assignments to same column "dt_modified"
```

TypeORM is probably adding any `@UpdateDateColumn` property to `update` queries which causes this error. I would expect to be able to override the value if needed.

`save` just ignores the value instead. Values passed in the `insert` however are used and the outcome is as expected.

Is this behaviour expected/intended or is this a bug?

feather-jmalone:
i agree with @wederer, there are times when it would be nice to manually override the value of an `@UpdateDateColumn`.

pleerock:
its not supposed for it, just write create a regular date column column and set date you need when you need

wederer:
@pleerock this should be better documented then. I will open up a PR for documentation changes. Please leave the issue open until then.

isotopeee:
Use `repository.update()` to override column values with `@CreateDateColumn` and `@UpdateDateColumn` programatically in automated tests.

tinaroh:
It would be a nice option to be able to override the values for generating fake test data.

[...]

PR #5861, opened 2020-04-11
docs: specify that columns using CreateDateColumn/UpdateDateColumn cannot be set manually

liaujianjie:
Follows the discussion on #3365.

Columns using the decorators `@CreateDateColumn()` and `@UpdateDateColumn()` cannot have their values manually set. This behaviour is not documented in the docs.

This PR documents that behaviour.


pleerock:
There are many requests on this feature. I think instead we need to implement this feature to prevent a future confusion. #3271

pleerock:
I created #5867 . I can close this PR, feel free to update documentation in a different way now.

liaujianjie:
That's great news, thank you!

PR #5867, opened 2020-04-13
feat: create-column, update-column, version-column column kinds now support user specified values

### LINKED ISSUE #3271: @CreateDateColumn and @UpdateDateColumn should use value if provided.

pleerock:
closes #3271 - create-column, update-column, version-column column kinds now support user specified values


test/core/column-kinds/create-date-column/create-date-column.ts
```diff
@@ -0,0 +1,96 @@
...
+        await sleep(1000);
+
+        // create post once again
+        post.title = "Updated Title";
+        await postRepository.save(post);
+
+        // check if date was created
+        const loadedPostAfterUpdate = await postRepository.findOne();
+        expect(loadedPostAfterUpdate!.createdAt.toString()).to.be.eql(loadedPostBeforeUpdate!.createdAt.toString());
+    })));
```
petrosschilling, line 76: This is bad. The whole purpose of having the `@UpdateDateColumn` is to update it on every save. And it has been disabled. WTF?

Issue #3280, opened 2018-12-18
Migration generate continuously repeats entries


ktwbc:
**Issue type:**

[ ] question
[ X] bug report
[ ] feature request
[ ] documentation issue

**Database system/driver:**

[ ] `cordova`
[ ] `mongodb`
[ ] `mssql`
[X ] `mysql` / `mariadb`
[ ] `oracle`
[ ] `postgres`
[ ] `sqlite`
[ ] `sqljs`
[ ] `react-native`
[ ] `expo`

**TypeORM version:**

[X ] `latest`
[ ] `@next`
[ ] `0.x.x` (or put your version here)

**Steps to reproduce or a small repository showing the problem:**

When building model files and relying on building and running migrations, TypeORM will constantly drop and add the same indexes over and over.

Table A:
```import { Entity, Column, ManyToOne, JoinColumn, PrimaryColumn } from 'typeorm';
import { TableB } from './table-b';

@Entity('table_a')
export class TableA {

  @PrimaryColumn('uuid')
  id: string;

  @Column({ nullable: true })
  etag: string;

  @ManyToOne(type => TableB, item => item.id)
  @JoinColumn({name: 'category_id'})
  category: TableB;

  @Column({ nullable: true })
  file_name: string;


}
```
Table B:
```import { Entity, PrimaryGeneratedColumn, Column, Index } from 'typeorm';

@Entity('table_b')
export class TableB {

  @PrimaryGeneratedColumn()
  id: number;

  @Column({ nullable: true})
  @Index("IDX_2a8279b83232e2243609048a69")
  uuid: string;

  @Column()
  title: string;


}
```

Steps:
a. build model files per above
b. Run migrate build:
```ts-node ./node_modules/typeorm/cli.js migration:generate -f ormconfig-local -c migrate -n 'mySql'```
c. view build migration:

```import {MigrationInterface, QueryRunner} from "typeorm";

export class mySql1545169032433 implements MigrationInterface {

    public async up(queryRunner: QueryRunner): Promise<any> {
        await queryRunner.query("CREATE TABLE `table_b` (`id` int NOT NULL AUTO_INCREMENT, `uuid` varchar(255) NULL, `title` varchar(255) NOT NULL, INDEX `IDX_2a8279b83232e2243609048a69` (`uuid`), PRIMARY KEY (`id`)) ENGINE=InnoDB");
[...]

PR #5869, opened 2020-04-13
Fix migrations being generated for FK even if there are no changes

zaro:
This is related to #3280 

We are using 0.2.22 with postrgres, but upgrading to 0.2.24 is impossible for us at that point, since migration:generate always generates bunch of FK recreations.
The deleted and the newly created FK are exactly  the same.

In the end I managed to trace this down to the fact that we are explicitly specifying the schema for each entity (we have entities in multiple schema).

So here is a fix. Metadata was building correctly the full table name for table changes, but this wasn't the case for  the FK definitions.


moltar:
legend! this was super annoying

pleerock:
Thanks!

PR #5874, opened 2020-04-14
fix: correct processing object keys in group method

Enemur:
Group method doesn't work with key with object type like jsonb


imnotjames:
Needs tests confirming behavior is corrected.

imnotjames:

PR #5886, opened 2020-04-16
doc: set User entity in the Repository decorator

eechava6:
added to the EntityRepository decorator the parameter User, since it won't work without it and following the docs in: 
https://typeorm.io/#/custom-repository/


pleerock:
No, it should work, why you think it doesn't work?

BTW this custom repository syntax will be removed in 0.3.0

null-char:
@pleerock What do you mean by this custom repository syntax will be removed? Do you mean custom repositories itself will be removed entirely?

pleerock:
no, it will be changed. You can take a look how it's going to look like on `next` branch (source code).

PR #5896, opened 2020-04-17
Fix validateCertificate is not valid option for the hdb-pool

wwasowicz:
Main changes:

According to **@sap/hana-client** doc _validateCerificate_ is not valid property:

https://help.sap.com/viewer/0eec0d68141541d1b07893a39944924e/2.0.03/en-US/4fe9978ebac44f35b9369ef5a4a26f4c.html

_sslValidateCertificate_ should be passed instead to the **hdb-pool**

Additional changes:

Fix for validation if a property is defined, especially important for numbers and booleans.

### DIFF
--- a/src/driver/sap/SapConnectionCredentialsOptions.ts
+++ b/src/driver/sap/SapConnectionCredentialsOptions.ts
@@ -36,7 +36,7 @@ export interface SapConnectionCredentialsOptions {
     /**
      * Validate database certificate
      */
-    readonly validateCertificate?: boolean;
+    readonly sslValidateCertificate?: boolean;
 
     /**
      * Key for encrypted connection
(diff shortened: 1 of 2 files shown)


elsbrock:
stumbled over the same - can confirm this fixes it

pleerock:
Thanks a lot @wwasowicz for contribution! I have already merged @elsbrock PR [that contained `](https://github.com/typeorm/typeorm/pull/5938/files) validateCerificate` change. So I want to ask if we really need "undefined" checks you added? The only case when we need it is when we need a zero as a valid value, do we really have such case?

imnotjames:
Going to close this as it doesn't bring any advantages over what we currently have in place.

PR #5902, opened 2020-04-18
DOCS: update connection-options.md

tareqdayya:
The "entitySchemas" connection option doesn't work. Instead, schemas are recognized when added to the "entities" connection option. The docs were updated to reflect that.


pleerock:
Thanks!

PR #5859, opened 2020-04-11
WIP: FIX for #2500. findOne(undefined|null|false) returns undefined.

V1c70r:
FIX for https://github.com/typeorm/typeorm/issues/2500.

PR #5905, opened 2020-04-19
fix: findOne(undefined|null|false) returns undefined

V1c70r:
FIX for #2500 in the next branch. Remade https://github.com/typeorm/typeorm/pull/5859.


pleerock:
Thanks a lot!

codingbrowny:
FInally

PR #5924, opened 2020-04-21
refactor deprecated new Buffer to Buffer.from

kaznovac:
`new Buffer()` api is deprecated as of node 10, refactor to `Buffer.from` available from node 5.10


pleerock:
Thanks!

PR #5925, opened 2020-04-21
test: multiple primary keys many-to-many can fail with 'Specified key was too long'

kaznovac:
set explicitly lengths of primary key fields

### DIFF
--- a/test/functional/relations/multiple-primary-keys/multiple-primary-keys-many-to-many/entity/Category.ts
+++ b/test/functional/relations/multiple-primary-keys/multiple-primary-keys-many-to-many/entity/Category.ts
@@ -10,10 +10,14 @@ import {Unique} from "../../../../../../src";
 @Unique(["code", "version", "description"])
 export class Category {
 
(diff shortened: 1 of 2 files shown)


pleerock:
Motivation? Tests are passing on master without these changes.

kaznovac:
i've run test on mysql 8 and this one was one of the failed.

explicit column lengths also guard against different column lengths between distributions, making this test more deterministic

imnotjames:
Can you create some tests that verify this behavior?

kaznovac:
@imnotjames these are changes in tests to make them more robust and not rely on the default values of column lenght

no need to test the tests

imnotjames:
Oh, duh :) I misread the PR somehow.  I've been looking at a bunch, forgive me. :sweat_smile:  Easy to mix things up!


So then this should be `test:` not `fix:` - I'll update the title.

imnotjames (approved):
Looks good to me.  Hopefully we'll soon be able to split drivers out and test them against multiple database versions!

PR #5926, opened 2020-04-21
feat: mysql8 geomcollection is synonymous to geometrycollection

kaznovac:
mysql reports `geomcollection` as column data_type. according to https://dev.mysql.com/doc/refman/8.0/en/gis-class-geometrycollection.html these two types are synonymous.

reporting column type as "geometrycollection" to keep backward compatibility.

### DIFF
--- a/src/driver/mysql/MysqlQueryRunner.ts
+++ b/src/driver/mysql/MysqlQueryRunner.ts
@@ -1372,6 +1372,12 @@ export class MysqlQueryRunner extends BaseQueryRunner implements QueryRunner {
                         tableColumn.precision = parseInt(dbColumn["DATETIME_PRECISION"]);
                     }
 
+                    if (tableColumn.type === "geomcollection" && !isMariaDb && VersionUtils.isGreaterOrEqual(dbVersion, "8.0.0")) {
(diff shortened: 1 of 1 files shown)


kaznovac:
should i add geomcollection as a valid type for column also?

pleerock:
Synonym (alias) names are always pain in ass, because orm compares db type with local type and perform schema sync if they mismatch.That's why orm "normalizes" alias names to a main one. Every database does the same internally. We need to map alias in the `normalizeType` method and make sure everything works right and sync works without making additional queries. Pain. Do we really need this alias?

kaznovac:
for mysql8 support i think we do need this. i can look what else needs to be done in `normalizeType` when i find some time

imnotjames:
There are no tests that verify this functionality.  Can you do that?

kaznovac:
@imnotjames sorry i no longer use the typeorm lib, if i get back to it i'll try to find the time to update pr with tests and update this `normalizeType` pleerock mentioned. i also wouldn't mind if someone takes over this pr.

imnotjames:
No worries on not using it anymore - thanks for the contribution anyway!  If you do end up on getting back to this, great!  If someone picks it up, also great. :)

Cheers!

PR #5927, opened 2020-04-21
Added support for NOWAIT & SKIP LOCKED in Postgres

Nopik:
This PR does add support for `SELECT .. FOR UPDATE SKIP LOCKED` & `SELECT .. FOR UPDATE NOWAIT` in Postgres.

It also does open an easy way to add this functionality for other DB servers.


pleerock:
Thanks! I'll merge it, but it would be great to have a test in a separate PR. I just [merged another PR ](https://github.com/typeorm/typeorm/pull/5971)that contained similar logic and tests.

PR #5929, opened 2020-04-21
chore: refactor mkdirp into native fs.mkdir with recursive

kaznovac:
use native mkdir with recursive option instead of mkdirp
(one less package to depend and audit)

fs.mkdir with recursive option is introduced in node 10.12 (which is oldest maintained version)

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -253,9 +253,9 @@
       "dev": true
     },
     "@types/node": {
-      "version": "9.6.42",
-      "resolved": "https://registry.npmjs.org/@types/node/-/node-9.6.42.tgz",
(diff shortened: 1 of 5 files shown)


pleerock:
We do support node 8 (even probably orm works on older versions too), I don't see a point to remove this support only because of one dependency for now.

kaznovac:
i thought when node 8 is no longer supported since (2019-12-31 https://github.com/nodejs/Release#end-of-life-releases) and node 10 is the oldest supported and having the needed api one less package to think about security advisories.

even mkdirp states that node10 is oldest supported and that lib uses native node api
https://github.com/isaacs/node-mkdirp/blob/master/CHANGELOG.md

imnotjames:
This will have to come in with a backwards compatibility break but I'm sure it can at that time. (I'm hoping for the `next` to have TS4.0 & Node >= 10 at the least.)

Issue #3695, opened 2019-02-22
Does "cli-highlight" really have to be a dependency or can it be a devDependency?


onionhammer:
**Issue type:**

[x] question
[ ] bug report
[ ] feature request
[ ] documentation issue

I am using webpack to perform treeshaking before pushing a library into a lambda; one of the largest dependencies in the output (according to the bundle analyzer) is cli-highlight. Can this be a devDependency only or is this massive package crucial to the runtime functioning of typeorm?

zackzachariah:
Adding to the concerns, `cli-highlight` is also responsible for all of the project's "High severity" warnings when you run `yarn audit` - it would be great to not have to worry about those in a production deploy situation.

pleerock:
yes, it have to be a regular dependency since CLI is using it. If we install it as dev-dependency then cli won't be able to use it.

Solution here is to extract cli functionality into separate typeorm-cli package, but this will not be implemented in a near future.

thatcort:
`cli-highlight` just broke our builds due to a problem with its dependence on `highlight.js`: https://github.com/highlightjs/highlight.js/issues/1984

I'd also love to see this dependency removed/moved/reduced.

zackzachariah:
Agree with your ideal solution @pleerock, though I wish there was an optional dependencies model we could use here in the meantime.

I'll keep following this issue for when the team gets around to splitting out the CLI code, though I'm sadly gonna be uncomfortable using `typeorm` in production code until then. I worry that the sheer number of vulnerabilities introduced via `cli-highlight`, even if they're not exposed to the open internet in production, makes it much more likely that a downstream user will miss new, important audit flags 😞

pleerock:
but why `cli-highlight` is bringing you such issues? Maybe worth reporting this issue on their repo?

zackzachariah:
(@pleerock - assuming your comment is in response to the user who just deleted their comment about a build break and not to me?)

pleerock:
no, it was to you. I'm talking about this statement:

> the sheer number of vulnerabilities introduced via `cli-highlight`

zackzachariah:
Ah ok - apologies for the misunderstanding.

I went to report an Issue with them as you suggested and found the following trail:
1. CLI-Highlight is aware of the issue and it's tracked by [felixfbecker/cli-highlight#58](felixfbecker/cli-highlight#58). It's an upstream dependency of _theirs_, `highlight.js`.
[...]

PR #5937, opened 2020-04-22
Optional cli-highlight

shz:
There's an existing issue around around the `cli-highlight` library (#3695).  Separately, the import time for `cli-highlight` is significant -- about 120ms on my local machine, for example -- which causes an unfortunate drop in cold-start times for any project using TypeORM.

Given the two points above, a fair solution seems to be to make the highlighting optional, so that any issues it causes (whether startup time, packaging issues, or others) can be skipped by projects that don't need the highlighting functionality.

Here's an example on cold start times from my development machine (measured on the execution time from `node -e "require('./build/compiled/src/index.js')"`) before/after making `cli-highlight` optional:

**Before**: `304.4154ms`
**After**: `186.2428ms`

And as a final note, I realize this may be a slightly controversial approach to a fix.  Happy to roll with whatever feedback you may have; I figured I'd start by suggesting a fix to the problem rather than just opening an issue.

### DIFF
--- a/docs/logging.md
+++ b/docs/logging.md
@@ -85,8 +85,9 @@ This code will log all queries which run more then `1 second`.
 
 TypeORM ships with 4 different types of logger:
 
-* `advanced-console` - this is the default logger which logs all messages into the console using color 
-and sql syntax highlighting (using [chalk](https://github.com/chalk/chalk)).
+* `advanced-console` - this is the default logger which logs all messages into the console using color
+and sql syntax highlighting (using [chalk](https://github.com/chalk/chalk), and only if
+[cli-highlight](https://github.com/felixfbecker/cli-highlight) is installed alongside TypeORM).
 * `simple-console` - this is a simple console logger which is exactly the same as the advanced logger, but it does not use any color highlighting.
 This logger can be used if you have problems / or don't like colorized logs.
 * `file` - this logger writes all logs into `ormlogs.log` in the root folder of your project (near `package.json` and `ormconfig.json`).
--- a/package-lock.json
+++ b/package-lock.json
@@ -449,11 +449,6 @@
       "integrity": "sha512-/FQM1EDkTsf63Ub2C6O7GuYFDsSXUwsaZDurV0np41ocwq0jthUAYCmhBX9f+KwlaCgIuWyr/4WlUQUBfKfZog==",
       "dev": true
     },
-    "any-promise": {
-      "version": "1.3.0",
-      "resolved": "https://registry.npmjs.org/any-promise/-/any-promise-1.3.0.tgz",
-      "integrity": "sha1-q8av7tzqUugJzcA3au0845Y10X8="
-    },
     "anymatch": {
       "version": "2.0.0",
       "resolved": "https://registry.npmjs.org/anymatch/-/anymatch-2.0.0.tgz",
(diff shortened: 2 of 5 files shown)


pleerock:
I understand, this is desirable by some people, but for most people it's not desirable. For most people highlighting is a quite good feature to have. And asking most people to separately install some additional package doesn't make lot of sense. Simply said we need less complexity and more convenience out of the box. So no, this approach won't work. Maybe we can came up a different approach, like sending a flag to disable this particular functionality? Can you tell because of what exactly problem comes from? If we simply don't highlight, does it work fast?

shz:
I guess there are really two underlying problems to unpack here:

 1. Including `cli-highlight` as a dependency increases bundle size (per https://github.com/typeorm/typeorm/issues/3695)
 2. *Importing* `cli-highlight` takes a significant amount of time (anywhere from 100ms - 200ms in my tests)

These are not huge issues in most places, but can have some fringe impacts.  For example, if you use TypeORM on a web API on GCP Cloud Run or AWS Lambda, that's an additional 120ms for cold-start requests, in an environment where the colorized logging isn't particularly helpful (it's often not possible to get colorized output from production logging tools).

And because those two problems are sort of at a module/import level, it's difficult to solve them with a light touch.  In theory we could use a conditional import to solve `2.` (though it's a little ugly), but `1.` is still going to be a problem.

Maybe something that would work is moving `cli-highlight` to an `optionalDependency`?  I'm not totally clear how it works, but the hope is that it would still be installed by default with `npm install typeorm`, but that `npm install typeorm --no-optional` would *not* install the highlighting.  Presumably with a `package-lock.json`, this means projects would have the ability to *opt-out* of having the highlighting for the ones that really need it.

Does that sound reasonable?

pleerock:
> And because those two problems are sort of at a module/import level

just to confirm - are you sure if only adding a single `import` of "cli-highlight" will bring problems performance issues?

filipsuk:
Just to add one argument - with recently added typeorm's support for AWS Aurora Data API, I think more and more people will use typeorm in serverless environments (like AWS lambda) where 100-200ms in cold start makes a big difference. So avoiding the import (2) in production env. would be a great addition.

shz:
> just to confirm - are you sure if only adding a single import of "cli-highlight" will bring problems performance issues?

Yep!  Here's an example:

```bash
mkdir test; cd test
npm install cli-highlight
echo 'require("cli-highlight")' >test.js
node test.js
```

Running `node test.js` on my development machine takes ~120ms, and I see similar numbers when running in production when benchmarked.

pleerock:
what if we dynamically require cli-highlight, but if we have some flag like `disableColors` (or a better name) then we don't require it

shz:
Sure that does sound like a good approach, and I'll update the PR to match that behavior.

Just as a note, this won't solve the bundle size issue in #3695, which I was originally trying to tackle as well in this PR.  Not so much a concern for me as the particular issue I care about here is import time, but I wanted to flag it down.

imnotjames:
It's definitely going to be important that we create a separate `cli` package where this would then be possible, but at this time we can't make `cli-highlight` a dev dependency, or optional dependency, or remove it for a number of folks.  It'd break a number of existing use cases.

If you'd like to make the change as suggested in this thread to have it dynamically imported & handled under a flag please feel free to open a new PR.

PR #5938, opened 2020-04-23
fix(hana-driver): SSL options, column delta detection mechanism

elsbrock:
fixes the following issues with the HANA driver:

* wrong SSL option being used: `sslValidateCertificate` (right) vs `validateCertificate`
* creation of composite FKs with wrong order
* column delta detection detecting changes erroneously for columns with default lengths and `uuid` type

@Szbuli @pleerock @AlexMesser


pleerock:
Thanks!

PR #5947, opened 2020-04-24
fix: fix insert and update query builder to handle mssql geometry column correctly

waikuen2010:
Query builder and methods of repository(e.g. insert, save) are handling geometry fields of mssql incorrectly.  

Current behavior is to set the input WKT (e.g. POINT(1 1)) to VALUES. However, according to [SQL server documentation](https://docs.microsoft.com/en-us/sql/t-sql/spatial-geometry/spatial-types-geometry-transact-sql?view=sql-server-ver15), the value should be set like this:
```
INSERT INTO SpatialTable (GeomCol1)  
VALUES (geometry::STGeomFromText('LINESTRING (100 100, 20 180, 180 180)', 0));  
```  
This pull request have make changes to query builders handling mssql geometry column.


pleerock:
Thank you!

Steffion:
My code stopped working because of this update. I figured out that the default SRID is now 0, which won't work for most people? A default value of 4326 (according to: https://stackoverflow.com/a/378941/4726278) would fix this issue (#6189)?

PR #5971, opened 2020-04-29
[Add] FOR NO KEY UPDATE lock mode for postgresql

ohjeyong:
Support `FOR NO KEY UPDATE` lock mode on postgresql.
If you will not update primary key, then `FOR NO KEY UPDATE` is better then `FOR UPDATE`, becaues `FOR UPDATE` can make deadlock with `SELECT FOR KEY SHARE` query.
The main difference between `FOR NO KEY UPDATE` and `FOR UPDATE` is `FOR NO KEY UPDATE` will not block `SELECT FOR KEY SHARE`.
The detail is described [this article](https://dba.stackexchange.com/questions/188169/select-for-update-locking-other-tables-in-postgresql).

### DIFF
--- a/src/query-builder/QueryExpressionMap.ts
+++ b/src/query-builder/QueryExpressionMap.ts
@@ -150,7 +150,7 @@ export class QueryExpressionMap {
     /**
      * Locking mode.
      */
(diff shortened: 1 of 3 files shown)


pleerock:
Thanks!

PR #5980, opened 2020-04-30
docs: Fix select query builder link

### LINKED ISSUE #5979: Query builder doc incorrect link
**Issue type:**

[ ] question
[ ] bug report
[ ] feature request
[X] documentation issue

https://github.com/typeorm/typeorm/blob/master/docs/select-query-builder.md

In the link above, the `What are aliases for?` doesn't lead to the right section or any section at all.

The problem is that the question mark is omitted from the link so it doesn't match with the title.

I've tried to escape it but with no luck. 

Hopefully, someone else can give it a shot.

vuki656:
Fixes #5979 if removing the question mark from the link is acceptable.


pleerock:
Thank you!

PR #6006, opened 2020-05-04
WIP Drop env autoload

Andreyco:
- [x] Provide failing test
- [ ] Make test passing
- [ ] Update docs and whatever


imnotjames:
I agree with this but it'll need to come in during a minor release where a breaking change is expected.

Cheers

PR #6007, opened 2020-05-04
docs: Update env variables listing

Andreyco:
Some of listed variables are not used anymore, some were are missing.
List is extracted from src/connection/options-reader/ConnectionOptionsEnvReader.ts


pleerock:
Thanks!

PR #6016, opened 2020-05-05
Add @DeleteDateColumn to special columns docs

rednebmas:
Mostly copied text from the decorator reference: https://github.com/typeorm/typeorm/blob/master/docs/decorator-reference.md#deletedatecolumn


pleerock:
Thanks!

PR #6027, opened 2020-05-07
Fix expo sqlite driver disconnect()

allandiego:
Currently the disconnect() function does not perform a real call to databaseConnection close, it only sets the queryRunner and databaseConnection to undefined, this behavior may cause problems when you do something like this:

1 - await getConnection().close();
2 - Replace the database file
3 - Try to open connection createConnection(); and do some insert

more info:
https://github.com/expo/expo/issues/8109

### DIFF
--- a/src/driver/expo/ExpoDriver.ts
+++ b/src/driver/expo/ExpoDriver.ts
@@ -40,6 +40,7 @@ export class ExpoDriver extends AbstractSqliteDriver {
         return new Promise<void>((ok, fail) => {
             try {
                 this.queryRunner = undefined;
+                this.databaseConnection._db.close();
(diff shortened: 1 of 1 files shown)


macfanatic:
Lots of discussion on this topic over the years, linking to make discovery easier for others trying to solve this issue.

- https://github.com/expo/expo/issues/2278
- https://github.com/expo/expo/issues/8109
- https://github.com/expo/expo/issues/639 (see comment specifically [here](https://github.com/expo/expo/issues/639#issuecomment-419723890))

pleerock:
Thanks! Would be great to have a test in a meanwhile.

Issue #2614, opened 2018-08-01
How can i use PostgreSQL through Unix socket?


MikalaiR:
**Issue type:**

[x] question
[ ] bug report
[ ] feature request
[ ] documentation issue

**Database system/driver:**

[ ] `cordova`
[ ] `mongodb`
[ ] `mssql`
[ ] `mysql` / `mariadb`
[ ] `oracle`
[x] `postgres`
[ ] `sqlite`
[ ] `sqljs`
[ ] `react-native`
[ ] `expo`

**TypeORM version:**

[x] `latest`
[ ] `@next`
[ ] `0.x.x` (or put your version here)

**Steps to reproduce or a small repository showing the problem:**

I tried something like:

```
{
  "type": "postgres",
  "url": "postgresql://user:pass@%2Fcloudsql%2Fpath:5432/db",

  "entities": [
    "dist/**/**.entity.js"
  ],
  "synchronize": true
}
```
but it doesn't working.

Kononnable:
@MikalaiR 
Did you resolve your issue? It doesn't look ORM related - can you connect with similar options using postgres driver directly?

MikalaiR:
@Kononnable I have resolved this issue using temporary 'crutch'.
```
{
  "type": "postgres",
  "url": "postgresql://user:pass@localhost:5432/db",

  "extra": {
    "host": "/cloudsql/...."
  },
```

It don't work in right way because TypeORM parses `url` itself instead of pass it to driver. And `extra` works because `extra` apllied to parsed object. But it's temporary crutch i think, TypeORM should support it through `url`.

fullofcaffeine:
I'd also like the know if this is currently supported / the status of this issue. Currently I'm using password auth restricted by host and pgsql is protected by a frewall, but it'd much more convenient if I had a unix user per db/app and could use pg_ident ithrough socket instead.

paeolo:
The issue is still present in the latest release. Why TypeORM doesn't pass the url to the driver directly ?

guillenotfound:
For anyone having the same issue, here are the variables I've set on my Cloud Run service:
![image](https://user-images.githubusercontent.com/5492960/89021248-7eeb8580-d320-11ea-9193-8590d84d33b4.png)

It does connect via socket, here is how the variable looks like:
`TYPEORM_HOST=/cloudsql/INSTANCE_CONNECTION_NAME`

[...]

PR #6042, opened 2020-05-10
fix: postgresql connection URL can use an UNIX Socket

paeolo:
Solves https://github.com/typeorm/typeorm/issues/2614.
* Current behavior:
Users may provide an URL in the connection options which is parsed by a generic function to fill the following fields: host, port, username, password, database.
* Issue:
You can't give an UNIX socket as host with PostgreSQL.
* Corrected behavior:
Provide a `ConnectionUrlParser` class with different parser functions. The default parser function is left unchanged to prevent any regression with other database types. Concerning PostgreSQL a new function is provided which solves the issue. To prevent any regression, I tested myself the function with multiples strings against the default parser.
* Exemple of tested strings
`postgres://user:password@host:5432/database` (standard)
`postgres://user:password@/var/run/postgresql?dbname=database` (UNIX socket)
`postgres:///var/run/postgresql?dbname=database` (without username)

### DIFF
--- a/src/driver/postgres/PostgresDriver.ts
+++ b/src/driver/postgres/PostgresDriver.ts
@@ -2,7 +2,6 @@ import {Driver} from "../Driver";
 import {ConnectionIsNotSetError} from "../../error/ConnectionIsNotSetError";
 import {ObjectLiteral} from "../../common/ObjectLiteral";
 import {DriverPackageNotInstalledError} from "../../error/DriverPackageNotInstalledError";
-import {DriverUtils} from "../DriverUtils";
 import {ColumnMetadata} from "../../metadata/ColumnMetadata";
 import {PostgresQueryRunner} from "./PostgresQueryRunner";
 import {DateUtils} from "../../util/DateUtils";
@@ -922,11 +921,12 @@ export class PostgresDriver implements Driver {
      */
     protected async createPool(options: PostgresConnectionOptions, credentials: PostgresConnectionCredentialsOptions): Promise<any> {
 
-        credentials = Object.assign({}, credentials, DriverUtils.buildDriverOptions(credentials)); // todo: do it better way
+        credentials = Object.assign({}, credentials);
 
         // build connection options for the driver
         // See: https://github.com/brianc/node-postgres/tree/master/packages/pg-pool#create
         const connectionOptions = Object.assign({}, {
+            connectionString: credentials.url,
             host: credentials.host,
             user: credentials.username,
             password: credentials.password,


pleerock:
Thanks for contribution.

You probably don't have to use `DriverUtils` at all. Just create a local function in postgres driver - that's it. `DriverUtils` was created for common usage, if some driver have its own logic, better if it won't use DriverUtils at all and use its own implementation.

Also, few other notes: 

* is there some popular package that is able to perform URL parsing? (to prevent extra maintainable code in the codebase). If no, that's fine.
* can we have a tests for this feature?

paeolo:
You're welcome!
* Yes there is: https://www.npmjs.com/package/pg-connection-string. I wanted to use this package first but wasn't sure about the dependencies policies, since there are only a few dependencies in the package.json.
* Yes I can make a test.
Paul.

pleerock:
I would like to prevent extra dependencies, because people who use `mysql` don't need `pg-connection-string`. I checked, it looks like `pg-connection-string` is installed when you install `pg` itself, and since this code is running only when `pg` is used, maybe we can just add a check if `pg-connection-string` and use it? (dynamic `require`). If nothing was found - throw an exception (not possible case I guess)

paeolo:
That's fine for me. I don't have time to do it this week-end but I can do it next week;)
Paul.

paeolo:
@pleerock Seems better now.
Paul.

guillenotfound:
Any update on this? 😄

paeolo:
> Any update on this?

It's ready for merge, waiting for @pleerock to review it.

src/driver/postgres/PostgresDriver.ts
```diff
@@ -922,7 +933,14 @@ export class PostgresDriver implements Driver {
      */
     protected async createPool(options: PostgresConnectionOptions, credentials: PostgresConnectionCredentialsOptions): Promise<any> {
 
-        credentials = Object.assign({}, credentials, DriverUtils.buildDriverOptions(credentials)); // todo: do it better way
+        credentials = Object.assign({}, credentials);
+
+        if (credentials.url) {
+            credentials = Object.assign(
+                credentials,
+                PostgresDriver.parseConnectionUrl(credentials.url)
```
imnotjames, line 941: Is it possible to pass the URL as `connectionString` 

See "connection uri" in https://node-postgres.com/features/connecting

src/driver/postgres/PostgresDriver.ts
```diff
@@ -880,6 +879,18 @@ export class PostgresDriver implements Driver {
         return "$" + (index + 1);
     }
 
+    // -------------------------------------------------------------------------
+    // Public Static Methods
+    // -------------------------------------------------------------------------
+
+    /**
+     * Return database credentials from a postgreSQL URL.
+     */
```
imnotjames, line 888: Can this be private?

imnotjames (changes requested):

src/driver/postgres/PostgresDriver.ts
paeolo: Well it's possible. But then the method "parseConnectionUrl" I added has no more reason to be, ie. the fix is one line (pass the URL as connectionString). I think it's better to do as you suggest but I am not sure then how to add a test for this.

imnotjames: Hmm.. well, less is more!  :)

I'm not sure, though - what happens if you try to open a connection to a socket that doesn't exist in postgres?  Can we confirm the driver handles that as expected?

paeolo: Yes of course, it would be the same as opening a TCP/IP connection that can't be open. I am ready to push this one line fix but the only thing you should be aware is I am not sure how to test this: CI platforms don't provide a way to connect through an UNIX socket (I mean maybe but it's not standard). I think it's not necessary to add a test for this. What do you think? Anyway i will push the one line in a moment.

guillenotfound:
I used what's described here and was able to connect from cloud run: https://github.com/typeorm/typeorm/issues/2614#issuecomment-667028124

src/driver/postgres/PostgresDriver.ts
imnotjames: If it's not possible to test it then so be it.  If there's any way we can or if you think of something, let's get more testing coverage.

src/driver/postgres/PostgresDriver.ts
```diff
@@ -922,11 +921,12 @@ export class PostgresDriver implements Driver {
      */
     protected async createPool(options: PostgresConnectionOptions, credentials: PostgresConnectionCredentialsOptions): Promise<any> {
 
-        credentials = Object.assign({}, credentials, DriverUtils.buildDriverOptions(credentials)); // todo: do it better way
```
imnotjames, line 925: Dropping this line drops something specifically for Oracle & the URL parsing which we don't want anyway.

Sounds good to me!

imnotjames (approved):

imnotjames:
Thanks for the contribution!

PR #6056, opened 2020-05-13
fix: escape column comment in mysql driver

BTOdell:
Closes: #6066


BTOdell:
Closes: #6066

pleerock:
Thanks!

Issue #3874, opened 2019-03-22
Using an (empty string) enum as the type of a primary key column


BTOdell:
**Issue type:**

[ ] question
[x] bug report
[ ] feature request
[ ] documentation issue

**Database system/driver:**

[ ] `cordova`
[ ] `mongodb`
[ ] `mssql`
[x] `mysql` / `mariadb`
[ ] `oracle`
[ ] `postgres`
[ ] `cockroachdb`
[ ] `sqlite`
[ ] `sqljs`
[ ] `react-native`
[ ] `expo`

**TypeORM version:**

[x] `latest`
[ ] `@next`
[x] `0.2.15`

**Steps to reproduce or a small repository showing the problem:**

```ts
import {BaseEntity, Entity, PrimaryColumn, Column} from "typeorm";

enum Singleton {
    EMPTY = ""
}

/**
 * Global application settings.
 */
@Entity("settings")
export class Settings extends BaseEntity {

    @PrimaryColumn()
    readonly singleton: Singleton = Singleton.EMPTY;

    @Column()
    value!: string;

}
```

I have a "settings" table in my database that holds global application settings (one setting per column). In non-TypeORM projects, I've always used an enum primary key to enforce the existence of only one row. But TypeORM is not loading anything when I run:

```ts
Settings.find();
```

The array is empty (despite having a fully populated row in the database).

If I change the entity definition to:

```ts
import {BaseEntity, Entity, PrimaryColumn, Column} from "typeorm";

/**
 * Global application settings.
 */
@Entity("settings")
export class Settings extends BaseEntity {

    @PrimaryColumn()
    readonly singleton: number = 0;

    @Column()
    value!: string;

}
```

then it starts working (after updating the database schema).

Kononnable:
It works if you don't set `EMPTY = ""` - set EMPTY as some string(with length>0) and you will be good.
Not sure why it's not working with empty string - probably because it's cast to false somewhere but it doesn't seem like a major bug.

BTOdell:
I ran into this bug again now (a year later) and this time I did some initial investigation.
I have deduced that the bug exists in the code path of the `RawSqlResultsToEntityTransformer#transform` function: https://github.com/typeorm/typeorm/blob/master/src/query-builder/transformer/RawSqlResultsToEntityTransformer.ts#L39

The raw SQL results contain a single row (as expected) but this transformer fails to produce a valid entity from it.

I am going to continue digging through the code and eventually create a pull request with the fix.

PR #6063, opened 2020-05-14
fix: use an empty string enum as the type of a primary key column

BTOdell:
Closes: #3874

### DIFF
--- a/src/query-builder/transformer/RawSqlResultsToEntityTransformer.ts
+++ b/src/query-builder/transformer/RawSqlResultsToEntityTransformer.ts
@@ -72,7 +72,6 @@ export class RawSqlResultsToEntityTransformer {
 
                 return keyValue;
             }).join("_"); // todo: check partial
-            if (!id) return;
 
             const items = map.get(id);
             if (!items) {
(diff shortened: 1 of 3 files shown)


pleerock:
I'm a bit scary to merge this one... Hope it won't break anything.

BTOdell:
I understand your concerns. I was very careful in reading through the code thoroughly before making this change.

The `id` variable will always be a `string`. It will never be `null` or `undefined` because it's the return value from the `Array.join` function: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/join

That means the ```if (!id)``` condition only succeeds if `id` is an empty string. `id` is used as a key into a `Map` but there is nothing invalid about using an empty string as a Map key.

This pull request will also fix issues with empty strings in `varchar` primary columns.

pleerock:
That's right, it's what we have at the end. However what if there is no primary key at all (someone selected data without `id` column lets say)? With previous changes it wont return anything. With new changes it will treat every selected row as a row of the same entity with id = `""`. Entity selection without id isn't valid case in the framework. But before we returned `[]` for such cases, now we'll return rows - this is more confusing and people may use this behaviour without understanding of what they should do.

pleerock:
I can agree your case is valid, however its super rare I think. Okay, let's merge it and see how it goes.

Issue #3147, opened 2018-11-22
queryRunner.query drops column when I try to modify a column


its-vincesanity:
**Issue type:**

[ ] question
[x] bug report
[ ] feature request
[ ] documentation issue

**Database system/driver:**

[ ] `cordova`
[ ] `mongodb`
[ ] `mssql`
[x] `mysql` / `mariadb`
[ ] `oracle`
[ ] `postgres`
[ ] `sqlite`
[ ] `sqljs`
[ ] `react-native`
[ ] `expo`

**TypeORM version:**

[x] `latest`
[ ] `@next`
[ ] `0.x.x` (or put your version here)

I wanted to migrate a column type from VARCHAR(255) to TEXT with:
```
export class NoTypeToTextMigration1542892455634 implements MigrationInterface {
  async up(queryRunner: QueryRunner): Promise<any> {
    // audit_log
    await queryRunner.query('ALTER TABLE `audit_log` MODIFY COLUMN `body` TEXT');
  }
```
Sometimes it works as expected and sometimes it drops the data from my column.
So, I setup an audit log in my MySQL:
```
SET GLOBAL log_output = 'TABLE';
SET GLOBAL general_log = 'ON';
```
My logs say:
```
| 2018-11-22 13:16:08 | root[root] @  [172.17.0.1] |        22 |         0 | Query        | ALTER TABLE `audit_log` DROP COLUMN `body`                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                       |
| 2018-11-22 13:16:08 | root[root] @  [172.17.0.1] |        22 |         0 | Query        | ALTER TABLE `audit_log` ADD `body` text NOT NULL
```

Why did it drop my table, I don't want lose my data If I change a column type :\

vincevaghalsen:
Same results with:
```
await queryRunner.renameColumn('audit_log', 'body', new TableColumn( {
  name: 'body',
  type: 'text',
  isNullable: true
}));
```
and
```
await queryRunner.changeColumn('audit_log', 'body', new TableColumn( {
  name: 'body',
  type: 'text',
  isNullable: true
}));
```
Both call 
```
ALTER TABLE `audit_log` DROP COLUMN `body`
```

jyang0110:
Is there any explanation about why data is lost?

Kononnable:
@vincevaghalsen 
[...]

PR #6081, opened 2020-05-17
refactor: update conditions for generate queries and avoid drop columns unnecessarily

ronielramos:
remove comparison of field length and type for generate queries with "drops"
issue: https://github.com/typeorm/typeorm/issues/3147


imnotjames:
Can you add tests that verify this behavior?

imnotjames:

PR #6084, opened 2020-05-18
SAP Hana bug fixes

pleerock:


test/integration/sample1-simple-entity.ts
```diff
@@ -4,7 +4,7 @@ import {Connection} from "../../src/connection/Connection";
 import {Post} from "../../sample/sample1-simple-entity/entity/Post";
 import {closeTestingConnections, createTestingConnections, reloadTestingDatabases} from "../utils/test-utils";
 
-describe("insertion", function() {
+describe.only("insertion", function() {
```
Andreyco, line 7: Fyi, `only` leftover

pleerock: yeah, thanks. I just experimenting over CI to run a single test

pleerock:
Okay, it looks like we can run SAP Hana tests on CI, but it took 10 minutes to launch a container.

PR #6085, opened 2020-05-18
Update query deep partial TypeScript definition

sdomagala:
User should be able to pass function if field is an array, right now it is only enabled for single values.

Example:
```
await getConnection()
      .createQueryBuilder()
      .update(Entity)
      .set({
        arrayField: (() => 'some update query') as unknown as string[],
      })
      .where('1 = 1')
      .execute();
```

This is the workaround I've used so far to hack around TypeScript, with this fix merged, you can drop `as unknown as string[]` and it works as expected

### DIFF
--- a/src/query-builder/QueryPartialEntity.ts
+++ b/src/query-builder/QueryPartialEntity.ts
@@ -11,7 +11,9 @@ export type QueryPartialEntity<T> = {
  */
 export type QueryDeepPartialEntity<T> = {
     [P in keyof T]?:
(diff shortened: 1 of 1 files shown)


pleerock:
what is the type of `arrayField` in the entity definition?

sdomagala:
@pleerock Sorry for the long response, I've missed the e-mail.
`arrayField` is supposed to be `string` but since the TS definition won't let me define like it like that, I had to cast type to `string[]`. IMHO the TypeScript definition should always allow type `() => string` to be passed not after some conditional checks

pleerock:
Think this is okay to merge.

Issue #4350, opened 2019-06-24
Array of enums column doesn't work at all


iFonny:
**Issue type:**

[ ] question
[ x ] bug report
[ ] feature request
[ ] documentation issue

**Database system/driver:**

[ ] `cordova`
[ ] `mongodb`
[ ] `mssql`
[ ] `mysql` / `mariadb`
[ ] `oracle`
[ x ] `postgres`
[ ] `cockroachdb`
[ ] `sqlite`
[ ] `sqljs`
[ ] `react-native`
[ ] `expo`

**TypeORM version:**

[ x ] `latest`
[ ] `@next`
[ ] `0.x.x` (or put your version here)

**Steps to reproduce or a small repository showing the problem:**

Hello!

I have a problem with creating a Column for my ReviewEntity. I want to insert some enums inside an Array. It's working with an enum with only 1 char (like in this [example](https://github.com/typeorm/typeorm/blob/master/test/functional/database-schema/simple-enums-array/entity/EnumArrayEntity.ts), I have no idea why) but not if I add a value in enum after creation or if I use words.

It's working on the first launch (creation), but If I run again (with sync enabled) I will get an error.

**review.entity**
```ts
  @Column({
    name: 'type',
    type: 'enum',
    enum: ReviewTypeEnum,
    array: true
  })
  type: ReviewTypeEnum[];
```

**review.enums**
```ts
export enum ReviewTypeEnum {
  Text = 'text',
  Image = 'image',
  Audio = 'audio',
  Video = 'video',
}
```

Full error:
```js
query failed: ALTER TYPE "public"."_review_types_enum" RENAME TO "review_types_enum_old"
error: { error: cannot alter array type review_types_enum[]
    at Connection.parseE (/Users/ifonny/Travail/Lucette/Lucette-API-nestjs/node_modules/pg/lib/connection.js:602:11)
    at Connection.parseMessage (/Users/ifonny/Travail/Lucette/Lucette-API-nestjs/node_modules/pg/lib/connection.js:399:19)
    at Socket.<anonymous> (/Users/ifonny/Travail/Lucette/Lucette-API-nestjs/node_modules/pg/lib/connection.js:121:22)
    at Socket.emit (events.js:189:13)
    at Socket.EventEmitter.emit (domain.js:441:20)
    at addChunk (_stream_readable.js:284:12)
    at readableAddChunk (_stream_readable.js:265:11)
    at Socket.Readable.push (_stream_readable.js:220:10)
[...]

PR #6126, opened 2020-05-25
fix: resolve postgres array enum migration issue

TheNoim:
Closes: #4350

### DIFF
--- a/src/driver/postgres/PostgresQueryRunner.ts
+++ b/src/driver/postgres/PostgresQueryRunner.ts
@@ -2065,11 +2065,11 @@ export class PostgresQueryRunner extends BaseQueryRunner implements QueryRunner 
         // You can not modify the array type. You need to edit the enum itself.
         if (column.isArray) {
             // Find the information about the array type
-            const typeResult = await this.query(`select "typelem" from pg_type where "typname" = '${udtName}';`);
+            const typeResult = await this.query(`SELECT "typelem" FROM pg_type WHERE "typname" = $1`, [udtName]);
 
             const typelem = typeResult[0]["typelem"];
             // Get with the oid of the enum array type the enum type
-            const baseTypeNameResult = await this.query(`select "typname" from pg_type where oid = ${typelem}`);
+            const baseTypeNameResult = await this.query(`SELECT "typname" FROM pg_type WHERE oid = $1`, [typelem]);
 
             const baseTypeName = baseTypeNameResult[0]["typname"];
 
--- a/src/driver/postgres/PostgresQueryRunner.ts
+++ b/src/driver/postgres/PostgresQueryRunner.ts
@@ -2059,9 +2059,27 @@ export class PostgresQueryRunner extends BaseQueryRunner implements QueryRunner 
         }
         const result = await this.query(`SELECT "udt_schema", "udt_name" ` +
             `FROM "information_schema"."columns" WHERE "table_schema" = '${schema}' AND "table_name" = '${name}' AND "column_name"='${column.name}'`);
+
+        let udtName = result[0]["udt_name"];
+
+        // You can not modify the array type. You need to edit the enum itself.
+        if (column.isArray) {
+            // Find the information about the array type
+            const typeResult = await this.query(`select "typelem" from pg_type where "typname" = '${udtName}';`);
+
+            const typelem = typeResult[0]["typelem"];
+            // Get with the oid of the enum array type the enum type
+            const baseTypeNameResult = await this.query(`select "typname" from pg_type where oid = ${typelem}`);
+
+            const baseTypeName = baseTypeNameResult[0]["typname"];
+
+            // Replace the udtName with the enum type
+            udtName = baseTypeName;
+        }
+
         return {
             enumTypeSchema: result[0]["udt_schema"],
-            enumTypeName: result[0]["udt_name"]
+            enumTypeName: udtName
         };
     }
 
--- a/src/driver/postgres/PostgresQueryRunner.ts
+++ b/src/driver/postgres/PostgresQueryRunner.ts
@@ -727,50 +727,6 @@ export class PostgresQueryRunner extends BaseQueryRunner implements QueryRunner 
                 downQueries.push(new Query(`ALTER TABLE ${this.escapePath(table)} ALTER COLUMN "${newColumn.name}" TYPE ${this.driver.createFullType(oldColumn)}`));
             }
 
-            if (
-                (newColumn.type === "enum" || newColumn.type === "simple-enum")
-                && (oldColumn.type === "enum" || oldColumn.type === "simple-enum")
-                && !OrmUtils.isArraysEqual(newColumn.enum!, oldColumn.enum!)
-            ) {
-                const enumName = this.buildEnumName(table, newColumn);
-                const arraySuffix = newColumn.isArray ? "[]" : "";
-                const oldEnumName = this.buildEnumName(table, newColumn, true, false, true);
-                const oldEnumNameWithoutSchema = this.buildEnumName(table, newColumn, false, false, true);
-                const enumTypeBeforeColumnChange = await this.getEnumTypeName(table, oldColumn);
-
-                // rename old ENUM
-                upQueries.push(new Query(`ALTER TYPE "${enumTypeBeforeColumnChange.enumTypeSchema}"."${enumTypeBeforeColumnChange.enumTypeName}" RENAME TO ${oldEnumNameWithoutSchema}`));
-                downQueries.push(new Query(`ALTER TYPE ${oldEnumName} RENAME TO  "${enumTypeBeforeColumnChange.enumTypeName}"`));
-
-                // create new ENUM
-                upQueries.push(this.createEnumTypeSql(table, newColumn));
-                downQueries.push(this.dropEnumTypeSql(table, oldColumn));
-
-                // if column have default value, we must drop it to avoid issues with type casting
-                if (newColumn.default !== null && newColumn.default !== undefined) {
-                    upQueries.push(new Query(`ALTER TABLE ${this.escapePath(table)} ALTER COLUMN "${newColumn.name}" DROP DEFAULT`));
-                    downQueries.push(new Query(`ALTER TABLE ${this.escapePath(table)} ALTER COLUMN "${newColumn.name}" SET DEFAULT ${newColumn.default}`));
-                }
-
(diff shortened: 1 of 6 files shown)


pleerock:
Thanks for contribution! Can we have a test that covers these changes?

TheNoim:
The thing is, I already tried to make some test cases. But I am not sure how to do a test case for a data model which changes. The only thing I could imagen was to spawn to different node process after each other with different typescript files which represent the model changes. However, I though, man this is dirty.

stewartmcgown:
@TheNoim is this fix compatible with the latest RC builds? I'm running into this exact issue.

TheNoim:
I only tested it on 0.2.25, but I think it should. Currently using the patch in production.

TheNoim:
@pleerock Any idea how to implement a test for this case? The only idea I came up is: forking the current process to use different typescript classes and enums. However, I think this isn't ideal.

TheNoim:
@pleerock Finally, this should be ready for merge.

TheNoim:
Any news?

src/driver/postgres/PostgresQueryRunner.ts
```diff
@@ -2019,9 +2019,27 @@ export class PostgresQueryRunner extends BaseQueryRunner implements QueryRunner
...
+        // You can not modify the array type. You need to edit the enum itself.
+        if (column.isArray) {
+            // Find the information about the array type
+            const typeResult = await this.query(`SELECT "typelem" FROM pg_type WHERE "typname" = $1`, [udtName]);
+
+            const typelem = typeResult[0]["typelem"];
+            // Get with the oid of the enum array type the enum type
+            const baseTypeNameResult = await this.query(`SELECT "typname" FROM pg_type WHERE oid = $1`, [typelem]);
+
+            const baseTypeName = baseTypeNameResult[0]["typname"];
```
imnotjames, line 2034: Is there a case where there won't be at least 1 value in this result set?

test/github-issues/4350/issue-4350.ts
```diff
@@ -0,0 +1,55 @@
...
+describe("github issues > #4350 Array of enums column doesn't work at all", () => {
+    let connections: Connection[];
+    after(() => closeTestingConnections(connections));
+
+    it("should migrate postgres enums correctly", async () => {
+        connections = await createTestingConnections({
+            entities: [__dirname + "/entity/Post1{.js,.ts}"],
+            schemaCreate: true,
+            dropSchema: true,
+            enabledDrivers: ["postgres"]
```
imnotjames, line 17: Can this also be enabled for cockroach DB

src/driver/postgres/PostgresQueryRunner.ts
TheNoim: Should always be 1 or zero.

src/driver/postgres/PostgresQueryRunner.ts
```diff
@@ -2019,9 +2019,27 @@ export class PostgresQueryRunner extends BaseQueryRunner implements QueryRunner
...
+        let udtName = result[0]["udt_name"];
+
+        // You can not modify the array type. You need to edit the enum itself.
+        if (column.isArray) {
+            // Find the information about the array type
+            const typeResult = await this.query(`SELECT "typelem" FROM pg_type WHERE "typname" = $1`, [udtName]);
+
+            const typelem = typeResult[0]["typelem"];
+            // Get with the oid of the enum array type the enum type
+            const baseTypeNameResult = await this.query(`SELECT "typname" FROM pg_type WHERE oid = $1`, [typelem]);
```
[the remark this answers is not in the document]
TheNoim: I could. However, I would just leave it like that. Better documentation and a not that long query.

src/driver/postgres/PostgresQueryRunner.ts
imnotjames: If there are 0 went this fail with an index error?  Is that what we want?

TheNoim: I mean, it can only fail if there is no array enum. It should always be 1 if there is one. And if not, then there is already something wrong somewhere else. Should I throw an actual error or what is your suggestion?

test/github-issues/4350/issue-4350.ts
TheNoim: Well, I would just revert https://github.com/typeorm/typeorm/pull/6126/commits/9e2276c5b673f7c3d8ba7fab4bbc1d019a01f917 again. I just notice cockroachdb as a separate driver which doesn't include my fix. So the test will fail.

TheNoim: The cockroachdb driver doesn't even handle enum cases. I think this is an issue by itself. I could open an issue for someone else.

TheNoim: Ok, will remove it. Is not even listed here: https://typeorm.io/#/entities/column-types-for-cockroachdb

test/github-issues/4350/issue-4350.ts
```diff
@@ -0,0 +1,55 @@
...
+import {Post1, TestEnum1} from "./entity/Post1";
+import {Post2, TestEnum2} from "./entity/Post2";
+import {Connection} from "../../../src";
+import {expect} from "chai";
+
+describe("github issues > #4350 Array of enums column doesn't work at all", () => {
+    let connections: Connection[];
+    after(() => closeTestingConnections(connections));
+
+    it("should migrate postgres enums correctly", async () => {
```
imnotjames, line 12: Can you add a test of going from an array to a non-array and back again?

Not sure if that's supported or not but if it worked before I'm curious if this keeps it operating.

If it didn't work before we don't need it.

src/driver/postgres/PostgresQueryRunner.ts
imnotjames: Could there be no array enum if you are making a column an array?

imnotjames (commented):
Just question about a test then we should be good

test/github-issues/4350/issue-4350.ts
TheNoim: I think this is a bit out of scope for this issue. The issue itself is only about changing an existing array column. I guess there should be a test for this behaviour, but not with this pull request.

src/driver/postgres/PostgresQueryRunner.ts
TheNoim: I don't think so.

test/github-issues/4350/issue-4350.ts
imnotjames: Ok - this test needs to exist somewhere before we can merge this in - it is technically out of scope.

Unfortunately, without a test we cannot tell that this change is causing that to break or not.

TheNoim:
Any news @imnotjames ? Don't want to annoy you, but it feels like forever since I opened this pr.

wallopthecat:
Can we get some eyes on this? Super annoying bug

TheNoim:
Any news?

JipSterk:
@imnotjames sorry for the ping but could you please re-review?

TheNoim:
Maybe @pleerock could look into this?

renehauck:
really need this :-/

TheNoim:
Soon I am at the point where I just cry. Both of my prs are stuck for nearly a year.

pleerock:
@TheNoim I'm sorry so much. PR is complex and its hard for me to review it. I had to ask @AlexMesser to help me.

Our conclusion on this PR is: even if it properly works, we aren't sure we want to merge it because of the value/complexity criteria. Let me explain. Right now it's super hard to review and merge changes in schema syncs. It's also hard to make changes to it even if you know (or think you know) how it works. Cost of mistake is high, because breaking change can lead to data loss on user side who didn't backup data before TypeORM upgrade. That's why in this area (rdbms sync) we prefer simple and obvious solutions as much as possible with feature sacrification if needed.

I don't want to say we won't fix this issue, we definitely need to fix it. But we are looking for the most simple approach. To understand if we can make it simpler let me ask first question:

* in TypeORM if we change the column type, e.g. from `varchar` to `int` we drop previous column and create a new column. This is implemented this way mostly due to simplicity
* I think if we change from `enum` to `enum array` we should consider it same way - its a column type change, and we should drop previous column and create a new column

Can we do it? I think it will lead to more consistent and simpler code with a small feature sacrification. What do you think about it?

stewartmcgown:
You're right to be cautious of the new functionality @pleerock. If it's any consolation, I've used this patch in production and it had the intended effect of migrating the existing enum values to a single member array.

pleerock:
@stewartmcgown thank you, any feedback on PR changes before they being merged are super valuable!

TheNoim:
I will comment on this later. I am a bit busy today.

TheNoim:
> @TheNoim I'm sorry so much. PR is complex and its hard for me to review it. I had to ask @AlexMesser to help me.
> 
> Our conclusion on this PR is: even if it properly works, we aren't sure we want to merge it because of the value/complexity criteria. Let me explain. Right now it's super hard to review and merge changes in schema syncs. It's also hard to make changes to it even if you know (or think you know) how it works. Cost of mistake is high, because breaking change can lead to data loss on user side who didn't backup data before TypeORM upgrade. That's why in this area (rdbms sync) we prefer simple and obvious solutions as much as possible with feature sacrification if needed.
> 
> I don't want to say we won't fix this issue, we definitely need to fix it. But we are looking for the most simple approach. To understand if we can make it simpler let me ask first question:
> 
> * in TypeORM if we change the column type, e.g. from `varchar` to `int` we drop previous column and create a new column. This is implemented this way mostly due to simplicity
> * I think if we change from `enum` to `enum array` we should consider it same way - its a column type change, and we should drop previous column and create a new column
> 
> Can we do it? I think it will lead to more consistent and simpler code with a small feature sacrification. What do you think about it?

Isn't this only about the changes I did later on this pr? I mean my original fix only included the pure bug fix for the change of an enum. It didn't include any other fixes like migration from enum to array enum and back. I only started to include this kind of change, because I thought @imnotjames required these. However, after looking through his replies I may misunderstood him. I could revert this changes and make this an pr with only the core bug fix. I mean this is the primary part of the pr: https://github.com/typeorm/typeorm/pull/6126/files#diff-b924b9a6d4ab4328f4ad59d4d96b794f6a1975fae8d1d3708909e31c151730b9R2101

Any thoughts?

PR #6133, opened 2020-05-26
OracleDriver :: Add Schema Option

pirouet:
Currently OracleDriver does not respect the `schema` option provided in the Entity decorator. This change is a simple one to add it to the list of Drivers that will have the schema set


liudonghua123:
The format of the code changed a lot, maybe preserve the original format, or make another pr about the format.

pirouet:
Hi, change of format is unintentional but the way my styling is in the IDE. 

I'll probably need to can this merge and create a new one. When are you looking to do a merge and push new version? This change would be ideal to have available for my Production apps

Loving the work 😁

pirouet:
Superseded by #6673. Closing this down - other change is the same but without changing the formatting

PR #6145, opened 2020-05-27
fix: revert "Handle URL objects as column field values"

pleerock:
Reverts typeorm/typeorm#5771


visar-polymath:
Is there a plan to publish this change as an NPM version?

sebsastianek:
URL && value instanceof URL could also be a solution

PR #6156, opened 2020-05-29
Fix issue with arguments being inserted

Destreyf:
According to https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/splice the splice method with a 0 delete will insert, where as a 1 will replace, this has been causing me massive headaches when i'm using 4-5 transaction repo's in a method, and have an argument that comes -after- the methods.  this leads to having a bunch of undefined arguments in the function calls.

### DIFF
--- a/src/decorator/transaction/Transaction.ts
+++ b/src/decorator/transaction/Transaction.ts
@@ -87,7 +87,7 @@ export function Transaction(connectionOrOptions?: string | TransactionOptions): 
                     }
 
                     // replace method param with injection of repository instance
-                    argsWithInjectedTransactionManagerAndRepositories.splice(metadata.index, 0, repositoryInstance);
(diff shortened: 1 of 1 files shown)


pleerock:
We are going to remove `Transaction` decorators in the future version. Please consider to have your own implementation (with changes you need) in your own codebase.

Destreyf:
@pleerock got it.

I already created my own decorator for this fix and just wanted to contribute back.

If you don't mind, can you explain the rationale behind removing the transaction decorators?  I'm just curious as to what the underlying reasoning is.

pleerock:
it will be removed because it is too specific. This specific arise many different problems to users. So, it's better for users to implement this decorator the way they need according to their specific use case.

PR #6160, opened 2020-05-29
Add postgres connection timeout option

mothershipper:
Commit message below to explain more of what's going on -- but before that, I just want to say I'd love some help writing tests. The contribution guidelines suggested this wouldn't get pulled in without them, however, I didn't see any existing test cases that would give me a starting point.

--- 

There was no documented way of setting a connection timeout for the
postgres driver. We recently ran into an issue with our network that
caused a container to hang indefinitely attempting to connect to
postgres.

We managed to resolve the issue by setting 'connectionTimeoutMillis' in
the 'extra' field of our connection options. This approach does not
appear to be documented anywhere.

Seeing that MongoDB and MySQL drivers both support a connection timeout
as part of the documented API, we felt it made sense to add a similar
option to the Postgres driver and hopefully avoid some headaches down
the road.

This commit adds a 'connectTimeoutMS' option to the postgres driver
that gets translated to the appropriate field for the pg library. It
also updates the documentation to reflect this new option.

Because the default behavior of the underlying pg library is to attempt
to connect indefinitely, we didn't feel like it was a safe change to
introduce a default timeout, even if that's more sane behavior.

As mentioned earlier, both the MySQL and MongoDB drivers support a
connection timeout option. MySQL uses 'connectTimeout' while MongoDB
uses 'connectTimeoutMS'. We went with 'connectTimeoutMS' as to not
introduce yet another name for a connection timeout, and because the
'MS' suffix makes it clearer what is expected.

We hope a future PR may adjust the MySQL connection options to adopt the
same name, but will leave that up to someone with a stronger opinion.

### DIFF
--- a/docs/connection-options.md
+++ b/docs/connection-options.md
@@ -173,6 +173,8 @@ See [SSL options](https://github.com/mysqljs/mysql#ssl-options).
 
 * `schema` - Schema name. Default is "public".
 
+* `connectTimeoutMS` - The milliseconds before a timeout occurs during the initial connection to the postgres server. If `undefined`, or set to `0`, there is no timeout. Defaults to `undefined`. 
+
 * `ssl` - Object with ssl parameters. See [TLS/SSL](https://node-postgres.com/features/ssl).
 
 * `uuidExtension` - The Postgres extension to use when generating UUIDs. Defaults to `uuid-ossp`. Can be changed to `pgcrypto` if the `uuid-ossp` extension is unavailable.
--- a/src/driver/postgres/PostgresConnectionOptions.ts
+++ b/src/driver/postgres/PostgresConnectionOptions.ts
@@ -34,6 +34,12 @@ export interface PostgresConnectionOptions extends BaseConnectionOptions, Postgr
     };
 
     /**
+     * The milliseconds before a timeout occurs during the initial connection to the postgres
+     * server. If undefined, or set to 0, there is no timeout. Defaults to undefined.
+     */
+    readonly connectTimeoutMS?: number;
+
+    /**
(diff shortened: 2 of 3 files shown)


stewartmcgown:
I actually cannot tell you how glad I am that you have experienced the same thing. I feel like I'm going insane with these infinitely hanging connections in Typeorm. I'll apply this patch and hopefully get some sleep tonight 🎉

mothershipper:
@stewartmcgown Not sure if you want to use this as a patch as there is a valid work-around. You can add the following to your connection options in the short term:

```typescript
createTypeORMConnection({
  ...,
  extra: {
    connectionTimeoutMillis: 10_000,
  },
})
```

I haven't heard from anyone on this PR yet, and honestly not sure when you could expect it to land.

stewartmcgown:
After adding that key to my extra config I get non-stop timeout configs:

```
nest_1       | Error: timeout exceeded when trying to connect
nest_1       |     at Timeout._onTimeout (/app/node_modules/pg-pool/index.js:180:27)
nest_1       |     at listOnTimeout (internal/timers.js:549:17)
nest_1       |     at processTimers (internal/timers.js:492:7)
```

mothershipper:
You could install `psql` in your container, exec into it and ensure you have connectivity between the container and the database. If you don't have connectivity with `psql`, then it's definitely the network or bad connection settings.

stewartmcgown:
What it seems to be is that sometimes the pg-pool is being overloaded, even with 64+ connections max. I'm wondering if there is some way to configure connection backpressure etc. I'm currently using entityManager.getRepository(XXX) for everything, which may be the cause? Not sure really.

pleerock:
If you have a high load and this error - make sure you have enough pool size capacity to handle your load.

PR #6165, opened 2020-05-31
feat: add scopes

oney:
Add scopes feature like ORMs in Ruby on Rails and Laravel.
See document https://github.com/oney/typeorm/blob/add-scopes/docs/scope.md

Please give any feedback or idea.
Might need more tests.

### DIFF
--- a/src/decorator/scopes/Scope.ts
+++ b/src/decorator/scopes/Scope.ts
@@ -1,0 +1,12 @@
+import {getMetadataArgsStorage} from "../../";
+import {ScopeMetadataArgs} from "../../metadata-args/ScopeMetadataArgs";
+
+export function Scope(global: boolean = false): Function {
+    return function (object: Object, propertyName: string) {
+        getMetadataArgsStorage().scopes.push({
+            target: object,
+            propertyName: propertyName,
+            global: global,
+        } as ScopeMetadataArgs);
+    };
+}
--- a/src/find-options/FindOneOptions.ts
+++ b/src/find-options/FindOneOptions.ts
@@ -1,6 +1,7 @@
 import {JoinOptions} from "./JoinOptions";
 import {ObjectLiteral} from "../common/ObjectLiteral";
 import {FindConditions} from "./FindConditions";
+import {SelectQueryBuilder, ScopeFn} from "../query-builder/SelectQueryBuilder";
 
 /**
  * Defines a special criteria to find specific entity.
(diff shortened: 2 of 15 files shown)


src/decorator/scopes/Scope.ts
```diff
@@ -0,0 +1,12 @@
+import {getMetadataArgsStorage} from "../../";
+import {ScopeMetadataArgs} from "../../metadata-args/ScopeMetadataArgs";
+
+export function Scope(global: boolean = false): Function {
```
Ginden, line 4: `MethodDecorator`

src/find-options/FindOneOptions.ts
```diff
@@ -32,6 +33,10 @@ export interface FindOneOptions<Entity = any> {
      */
     order?: { [P in keyof Entity]?: "ASC"|"DESC"|1|-1 };
 
+    unscoped?: boolean;
```
Ginden, line 36: I would prefer `scoped` with default `true`.

src/decorator/scopes/Scope.ts
Ginden, line 4: Boolean parameters are usually considered bad. `{global: boolean}` will be future-proof.

src/metadata/ScopeMetadata.ts
```diff
@@ -0,0 +1,59 @@
...
+
+    // ---------------------------------------------------------------------
+    // Public Build Methods
+    // ---------------------------------------------------------------------
+
+    /**
+     * Builds some depend scope properties.
+     */
+    build(): this {
+        return this;
```
Ginden, line 56: Why is it needed?

src/query-builder/SelectQueryBuilder.ts
```diff
@@ -2027,4 +2031,32 @@ export class SelectQueryBuilder<Entity> extends QueryBuilder<Entity> implements
...
+        this.expressionMap.globalScoped = globalScoped;
+        return this;
+    }
+
+    scope(scopeFn: ScopeFn<this>): this;
+    scope(scopeFns: Array<ScopeFn<this>>): this;
+    scope(scopeFn:  ScopeFn<this> | Array<ScopeFn<this>>): this {
+        const scopeFns = Array.isArray(scopeFn) ? scopeFn : [scopeFn];
+        scopeFns.forEach(scopeFn => {
+            scopeFn(this);
```
Ginden, line 2058: Should provide alias as second parameter - two entities can have the same field name.

pleerock:
According to author last comments in a different threads it doesn't seem he is interesting in doing anything in this repository.

PR #6169, opened 2020-05-31
Fix MySQL Foreign Key schema query

### LINKED ISSUE #6168: Wrong foreign key information loaded from schema
**Issue type:**

[ ] question
[x] bug report
[ ] feature request
[ ] documentation issue

**Database system/driver:**

[ ] `cordova`
[ ] `mongodb`
[ ] `mssql`
[x] `mysql` / `mariadb`
[ ] `oracle`
[ ] `postgres`
[ ] `cockroachdb`
[ ] `sqlite`
[ ] `sqljs`
[ ] `react-native`
[ ] `expo`

**TypeORM version:**

[x] `latest`
[ ] `@next`
[x] `0.2.5`

When using MySQL/MariaDB if you have an entity with a foreign key and it is created in two different schemas (i.e. a development schema and a staging schema) on the same server, the information loaded from the schema is non-deterministic, as it can come from either schema, due to a missing clause in the query used to load the data from MySQL's information schema.

A simple way to test this is:

1. Create an entity with a `ManyToOne` relation to another entity
1. Generate the migration and synchronize to the database
1. Change the schema for the entity
1. Change the `ManyToOne` relation `onUpdate` and/or `onDelete` values
1. Generate the migration for the new schema
1. Apply the migration
1. Generate a new migration. Due to the wrong information loaded the `onUpdate` or `onDelete` values might come from the wrong schema and generate an unnecessary foreign key recreation

weckx:
Add a new clause for the join with `INFORMATION_SCHEMA.REFERENTIAL_CONSTRAINTS` to prevent errors when there is more than one database in the server with foreign keys with same name.

Closes #6168


imnotjames:
Unfortunately, we did pull in a different PR that accomplished this :thinking: but looking at this, I like that you used the `inner join` to make the query.

I'll be closing this but I will probably make the change you made here the actual change we keep in the repo.

PR #6173, opened 2020-06-01
Update docs on replication and raw queries

Neo-Zhixing:
The docs were not very clear about the behavior of read/write replication when performing raw queries.

### DIFF
--- a/docs/multiple-connections.md
+++ b/docs/multiple-connections.md
@@ -275,7 +275,8 @@ Example of replication connection settings:
 ```
 
 All schema update and write operations are performed using `master` server.
-All simple queries performed by find methods or select query builder are using a random `slave` instance. 
+All simple queries performed by find methods or select query builder are using a random `slave` instance.
+All raw queries performed by query method are always performed using `master` server.
 
 If you want to explicitly use master in SELECT created by query builder, you can use the following code:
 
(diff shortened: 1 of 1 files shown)


pleerock:
I don't think your statement is correct. What do you call "raw queries" ?

Neo-Zhixing:
@pleerock When you directly write a SQL query in plain text (like `connection.query('SELECT * FROM users WHERE users.id=$1', [id])`) without a repository or a query builder.

In this case TypeORM will always use the master server to perform the query.

The behavior or raw queries was never defined in the docs. This PR makes it clear to the user that if they want to perform the raw query on a slave server they'd have to specify this manually. TypeORM will not perform `SELECT` on the slave server automatically in this case.

docs/multiple-connections.md
```diff
@@ -275,7 +275,8 @@ Example of replication connection settings:
 ```
 
 All schema update and write operations are performed using `master` server.
-All simple queries performed by find methods or select query builder are using a random `slave` instance. 
+All simple queries performed by find methods or select query builder are using a random `slave` instance.
+All raw queries performed by query method are always performed using `master` server.
```
imnotjames, line 279:
```suggestion
All queries performed by query method are performed using the `master` instance.
```

docs/multiple-connections.md
```diff
@@ -291,6 +292,20 @@ try {
...
+```typescript
+
+const slaveQueryRunner = connection.createQueryRunner("slave");
+await connection.query('SELECT * FROM users WHERE id = $1', [userId], slaveQueryRunner)
+  .then((result) => {
+    // Do your stuff
+  })
+  .finally(() => {
+    return slaveQueryRunner.release()
+  })
```
imnotjames, line 306: If this is going to be a separate query please match the `try`/`finally` syntax as above.

docs/multiple-connections.md
```diff
@@ -291,6 +292,20 @@ try {
         
 ```
 
+If you want to use `slave` in raw queries, you also need to explicitly specify the query runner.
```
imnotjames, line 295: Why not change the above message and example  so it shows both?  Could lead to more concise documentation.

Neo-Zhixing: I think it's better to keep them separate. Although the way we create `QueryRunner` was similar in both examples, what I'm really trying to show here is "how to specify QueryRunner for raw queries". Imo this saves readers' time and makes things more clear.

imnotjames:
Thanks for the contribution

Issue #6174, opened 2020-06-01
QueryBuilder generates invalid SQL for update when using both `returning` and `order by`


jahudka:
This is a **bug report**. The driver used is **postgres**, TypeORM version **0.2.25** (latest at the time of writing this report). To reproduce:

```typescript
import { getConnection } from 'typeorm';
import { ExampleEntity } from '../entities';

(async () => {
  const { manager: em } = await getConnection();

  const builder = em.createQueryBuilder()
    .update(ExampleEntity)
    .set({
      exampleProp: () => 'exampleProp + 1',
    })
    .where('exampleProp < 10')
    .orderBy('exampleProp', 'ASC')
    .limit(1)
    .returning('*');

  const { raw: entities } = await builder.execute();
})();
```

This generates the following SQL:

```sql
UPDATE "example_entity"
SET "example_prop" = "example_prop" + 1
WHERE "example_prop" < 10
RETURNING *
ORDER BY "example_prop" ASC
LIMIT 1
```

As you can see, the `RETURNING *` clause is inserted early, which results in a parse error in PostgreSQL. From what I can gather from TypeORM source this is caused by the `UpdateQueryBuilder`: the [`getQuery()`](https://github.com/typeorm/typeorm/blob/master/src/query-builder/UpdateQueryBuilder.ts#L49) method concatenates the results of `createUpdateExpression()`, `createOrderByExpression()` and `createLimitExpression()`, but the `RETURNING` clause is inserted [within the `createUpdateExpression()` method](https://github.com/typeorm/typeorm/blob/master/src/query-builder/UpdateQueryBuilder.ts#L526) and so it can never work in PostgreSQL (and potentially other platforms [other than SQLServer](https://github.com/typeorm/typeorm/blob/master/src/query-builder/UpdateQueryBuilder.ts#L529)) when an `ORDER BY` or `LIMIT` clause is also applied.

I'll be posting a PR to fix this issue shortly.

jahudka:
:-( so it turns out that Postgres doesn't support UPDATE ... LIMIT, which I was sure it did for some reason.. Nevermind, sorry about that, closing..

PR #6175, opened 2020-06-01
fix: support UpdateQueryBuilder orderBy() / limit() and returning() at the same time

jahudka:
Closes: #6174


jahudka:
:-( so it turns out that Postgres doesn't support UPDATE ... LIMIT, which I was sure it did for some reason.. Nevermind, sorry about that, closing this PR & the associated issue..

PR #6180, opened 2020-06-02
Point users to typeorm.io for docs

mikemaccana:
Hello! I hope you don't mind, but it seems that:

 - typeorm.io is the official website for this project
 - typeorm.io has more extensive documentation than this README does 
 - for whatever reason, search engines don't know about typeorm.io, so many users don't realise that site exists

Please let me know if any of those are incorrect!

This PR sends users to the more complete documentation at typeorm.io


pleerock:
You are right! However I love when project github home pages have a compressive documentation without having to go to some other site! I'm sure somebody who needs more than basic can easily find a documentation site. Most probably we'll change this in the future, but for now I would like to keep it as it is. Thanks!

PR #6199, opened 2020-06-04
fix: enable explicitly inserting IDENTITY values into mssql

dklischies:
Allow explicit insertion of preset values for mssql IDENTITY columns. These are the equivalent to
auto-incrementing primary keys in MySQL, but require special conditional treatment.

Closes: #2199 for mssql

### DIFF
--- a/src/query-builder/InsertQueryBuilder.ts
+++ b/src/query-builder/InsertQueryBuilder.ts
@@ -347,6 +347,18 @@ export class InsertQueryBuilder<Entity> extends QueryBuilder<Entity> {
             query += ` RETURNING ${returningExpression}`;
         }
 
+
+        // Inserting a specific value for an auto-increment primary key in mssql requires enabling IDENTITY_INSERT
+        // IDENTITY_INSERT can only be enabled for tables where there is an IDENTITY column and only if there is a value to be inserted (i.e. supplying DEFAULT is prohibited if IDENTITY_INSERT is enabled)
+        if (this.connection.driver instanceof SqlServerDriver
+            && this.expressionMap.mainAlias!.hasMetadata
+            && this.expressionMap.mainAlias!.metadata.columns
+                .filter((column) => this.expressionMap.insertColumns.length > 0 ? this.expressionMap.insertColumns.indexOf(column.propertyPath) !== -1 : column.isInsert)
+                .some((column) => this.isOverridingAutoIncrementBehavior(column))
+        ) {
+            query = `SET IDENTITY_INSERT ${tableName} ON; ${query}; SET IDENTITY_INSERT ${tableName} OFF`;
+        }
+
         return query;
     }
 
@@ -372,7 +384,8 @@ export class InsertQueryBuilder<Entity> extends QueryBuilder<Entity> {
                 && !(this.connection.driver instanceof OracleDriver)
                 && !(this.connection.driver instanceof AbstractSqliteDriver)
                 && !(this.connection.driver instanceof MysqlDriver)
-                && !(this.connection.driver instanceof AuroraDataApiDriver))
+                && !(this.connection.driver instanceof AuroraDataApiDriver)
+                && !(this.connection.driver instanceof SqlServerDriver && this.isOverridingAutoIncrementBehavior(column)))
                 return false;
 
             return true;
@@ -598,6 +611,21 @@ export class InsertQueryBuilder<Entity> extends QueryBuilder<Entity> {
             return [this.expressionMap.valuesSet];
 
         throw new InsertValuesMissingError();
+    }
+
+    /**
+     * Checks if column is an auto-generated primary key, but the current insertion specifies a value for it.
+     * 
+     * @param column
+     */
+    protected isOverridingAutoIncrementBehavior(column: ColumnMetadata): boolean {
+        return column.isPrimary 
+                && column.isGenerated 
+                && column.generationStrategy === "increment"
+                && this.getValueSets().some((valueSet) => 
+                    column.getEntityValue(valueSet) !== undefined 
+                    && column.getEntityValue(valueSet) !== null
+                );
     }
 
 }
--- a/test/github-issues/2199/issue-2199.ts
+++ b/test/github-issues/2199/issue-2199.ts
@@ -1,15 +1,16 @@
 import { expect } from "chai";
 import "reflect-metadata";
 import { Connection } from "../../../src/connection/Connection";
+import { SqlServerDriver } from '../../../src/driver/sqlserver/SqlServerDriver';
 import { closeTestingConnections, createTestingConnections, reloadTestingDatabases } from "../../utils/test-utils";
 import { Bar } from "./entity/Bar";
 
-describe("github issues > #2199 - Inserting value for @PrimaryGeneratedColumn() for mysql and sqlite", () => {
+describe("github issues > #2199 - Inserting value for @PrimaryGeneratedColumn() for mysql, sqlite and mssql", () => {
 
     let connections: Connection[];
     before(async () => connections = await createTestingConnections({
         entities: [__dirname + "/entity/*{.js,.ts}"],
-        enabledDrivers: ["mysql", "mariadb", "sqlite", "better-sqlite3"],
+        enabledDrivers: ["mysql", "mariadb", "sqlite", "better-sqlite3", "mssql"],
         schemaCreate: true,
         dropSchema: true
     }));
@@ -43,5 +44,26 @@ describe("github issues > #2199 - Inserting value for @PrimaryGeneratedColumn() 
         });
         const thirdBar = await connection.manager.save(thirdBarQuery);
         expect(thirdBar.id).to.eql(5);
+    })));
+
+    it("should reset mssql's INSERT_IDENTITY flag correctly after failed queries", () => Promise.all(connections
+        .filter(connection => connection.driver instanceof SqlServerDriver)
+        .map(async connection => {
+            // Run a query that failes at the database level
+            await expect(connection.createQueryBuilder()
+                .insert()
+                .into(Bar)
+                .values({
+                    id: 20,
+                    description: () => "NONEXISTINGFUNCTION()",
+                })
+                .execute()
+            ).to.be.rejectedWith("Error: 'NONEXISTINGFUNCTION' is not a recognized built-in function name.");
+            // And now check that IDENTITY_INSERT is disabled by inserting something without an ID value and see if that works
+            const successfulBarQuery = connection.manager.create(Bar, {
+                description: "default id value"
+            });
+            const bar = await connection.manager.save(successfulBarQuery);
+            expect(bar.id).to.be.a('number');
     })));
 });


dklischies:
@pleerock Any opinion on this? Currently the behavior differs between database engines (i.e. Oracle DB, MySQL and Sqllite will update auto increment values, MSSQL will not). This is a step towards homogenizing this behavior.

pleerock:
shouldn't this be set on a connection-level or configuration level? I found it weird having it here.

dklischies:
I agree that this is rather weird, but that weirdness is a result of the way Microsoft implemented enabling manual insertion of values into identity columns. You can only enable `INSERT_IDENTITY` for one table per connection at the same time (see [docs](https://docs.microsoft.com/en-us/sql/t-sql/statements/set-identity-insert-transact-sql?view=sql-server-ver15#remarks)). That means we cannot enable this once when connecting to the database, because that would mean only enabling it for one table. Additionally, enabling this option disables the default value for the identity column. That means that we only want to enable `INSERT_IDENTITY` if the user supplies a value for the identity column (for specific queries) and not enable it if the user does not supply a value. These two issues lead to the requirement of enabling and disabling the option on a per query basis.

Lastly you are not allowed to set `INSERT_IDENTITY` if there is no Identity column in a table, hence all these checks in line 376ff of src/query-builder/InsertQueryBuilder.ts

src/query-builder/InsertQueryBuilder.ts
```diff
@@ -335,6 +335,19 @@ export class InsertQueryBuilder<Entity> extends QueryBuilder<Entity> {
...
+        // Inserting a specific value for an auto-increment primary key in mssql requires enabling IDENTITY_INSERT
+        // IDENTITY_INSERT can only be enabled for tables where there is an IDENTITY column and only if there is a value to be inserted (i.e. supplying DEFAULT is prohibited if IDENTITY_INSERT is enabled)
+        if (this.connection.driver instanceof SqlServerDriver
+            && this.expressionMap.mainAlias!.hasMetadata
+            && this.expressionMap.mainAlias!.metadata.columns
+                .filter((column) => this.expressionMap.insertColumns.length > 0 ? this.expressionMap.insertColumns.indexOf(column.propertyPath) !== -1 : column.isInsert)
+                .some((column) => column.isPrimary && column.isGenerated && column.generationStrategy === "increment"
+                    && this.getValueSets().some((valueSet) => column.getEntityValue(valueSet) !== undefined && column.getEntityValue(valueSet) !== null))
+        ) {
+            query = `SET IDENTITY_INSERT ${tableName} ON; ${query}; SET IDENTITY_INSERT ${tableName} OFF`;
```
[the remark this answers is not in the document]
dklischies: That works correctly, as long as there is no syntax error in the query (but some other error like a missing function), at which point there is likely a bug in the query generator.
I've added a test case for that.

src/query-builder/InsertQueryBuilder.ts
```diff
@@ -335,6 +335,19 @@ export class InsertQueryBuilder<Entity> extends QueryBuilder<Entity> {
...
+
+        // Inserting a specific value for an auto-increment primary key in mssql requires enabling IDENTITY_INSERT
+        // IDENTITY_INSERT can only be enabled for tables where there is an IDENTITY column and only if there is a value to be inserted (i.e. supplying DEFAULT is prohibited if IDENTITY_INSERT is enabled)
+        if (this.connection.driver instanceof SqlServerDriver
+            && this.expressionMap.mainAlias!.hasMetadata
+            && this.expressionMap.mainAlias!.metadata.columns
+                .filter((column) => this.expressionMap.insertColumns.length > 0 ? this.expressionMap.insertColumns.indexOf(column.propertyPath) !== -1 : column.isInsert)
+                .some((column) => column.isPrimary && column.isGenerated && column.generationStrategy === "increment"
+                    && this.getValueSets().some((valueSet) => column.getEntityValue(valueSet) !== undefined && column.getEntityValue(valueSet) !== null))
+        ) {
```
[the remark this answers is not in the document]
dklischies: The `hasMetadata` check was required since there was a code path that led to this line without the metadata being set (rerunning the tests without that check should show that).
I've refactored that last line into a separate function, which we can use in `getInsertedColumns` as well. For an explanation, please see the comment above the if. Please feel free to let me know if there is something else you need to know.

src/query-builder/InsertQueryBuilder.ts
```diff
@@ -360,7 +373,8 @@ export class InsertQueryBuilder<Entity> extends QueryBuilder<Entity> {
                 && !(this.connection.driver instanceof OracleDriver)
                 && !(this.connection.driver instanceof AbstractSqliteDriver)
                 && !(this.connection.driver instanceof MysqlDriver)
-                && !(this.connection.driver instanceof AuroraDataApiDriver))
+                && !(this.connection.driver instanceof AuroraDataApiDriver)
+                && !(this.connection.driver instanceof SqlServerDriver && this.getValueSets().some((valueSet) => column.getEntityValue(valueSet) !== undefined && column.getEntityValue(valueSet) !== null)))
```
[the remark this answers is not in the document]
dklischies: Yes it is. There are three scenarios:

- You have a column which is not an entity value, then including the column is fine.

- You have a column which is an entity value, but you are not specifying a valid value for that (i.e. `undefined` or `null`). Then the primary key must be generated automatically. To get this working in mssql, you are not allowed to include the column name in the query, otherwise it will complain because you are not allowed to insert something unless `IDENTITY_INSERT` is enabled. However if you enable that, it will not generate a value anymore.

- You have a column which is an entity value, and you are specifying a value for that. Then `IDENTITY_INSERT` will be enabled and you must specify the column name.

I hope that helps. If not, please let me know. I spent some time on figuring out the behavior of mssql server with regard to that `IDENTITY_INSERT` option.

src/query-builder/InsertQueryBuilder.ts
```diff
@@ -347,6 +347,18 @@ export class InsertQueryBuilder<Entity> extends QueryBuilder<Entity> {
...
+
+        // Inserting a specific value for an auto-increment primary key in mssql requires enabling IDENTITY_INSERT
+        // IDENTITY_INSERT can only be enabled for tables where there is an IDENTITY column and only if there is a value to be inserted (i.e. supplying DEFAULT is prohibited if IDENTITY_INSERT is enabled)
+        if (this.connection.driver instanceof SqlServerDriver
+            && this.expressionMap.mainAlias!.hasMetadata
+            && this.expressionMap.mainAlias!.metadata.columns
+                .filter((column) => this.expressionMap.insertColumns.length > 0 ? this.expressionMap.insertColumns.indexOf(column.propertyPath) !== -1 : column.isInsert)
+                .some((column) => this.isOverridingAutoIncrementBehavior(column))
+        ) {
+            query = `SET IDENTITY_INSERT ${tableName} ON; ${query}; SET IDENTITY_INSERT ${tableName} OFF`;
```
imnotjames, line 359: Are you able to escape the path for the table name here?

src/query-builder/InsertQueryBuilder.ts
```diff
@@ -600,4 +613,19 @@ export class InsertQueryBuilder<Entity> extends QueryBuilder<Entity> {
...
+     * 
+     * @param column
+     */
+    protected isOverridingAutoIncrementBehavior(column: ColumnMetadata): boolean {
+        return column.isPrimary 
+                && column.isGenerated 
+                && column.generationStrategy === "increment"
+                && this.getValueSets().some((valueSet) => 
+                    column.getEntityValue(valueSet) !== undefined 
+                    && column.getEntityValue(valueSet) !== null
```
imnotjames, line 627: If it is null isn't that technically a valid value that we'd want to count as a "supplied" value

imnotjames:
I'm making a change now to loosen up that insert test.

imnotjames:
ACTUALLY, I'm confused - why is it running it for that code?

The DDL for the table that's in the test that failed is:

```
create table test
(
	id int not null
		constraint PK_5417af0062cf987495b611b59c7
			primary key,
	description nvarchar(255) not null
)
```

It's not an `IDENTITY` type.

src/query-builder/InsertQueryBuilder.ts
imnotjames, line 359: Also, Instead of doing this as part of building the query itself, how would it look as part of the `execute` function?

That way it's not muddying up the SQL we've generated - and we already do something similar to that for `@OutputTable`.  That way it's more for the session than the query itself we're generating.

dklischies:
> 
> 
> ACTUALLY, I'm confused - why is it running it for that code?
> 
> The DDL for the table that's in the test that failed is:
> 
> ```
> create table test
> (
> 	id int not null
> 		constraint PK_5417af0062cf987495b611b59c7
> 			primary key,
> 	description nvarchar(255) not null
> )
> ```
> 
> It's not an `IDENTITY` type.
> 
> But it should be - 🤔 This isn't a problem with your changes.

It is an entity type. Look at the related entity/test.ts in that folder

```
@Entity()
export class Test {

    @PrimaryGeneratedColumn()
    id: number;

}
```

src/query-builder/InsertQueryBuilder.ts
dklischies: I thought about that. However, i think its more appropriate to have it here. Otherwise, the query generation method will generate a query which cannot be run, unless you execute it with the execute function. In other words, this method would build an invalid query, without additional context provided by execute. There is also a lot of driver dependent code in this method anyways.

The OutputTable handling in the execute method on the other hand has nothing to do with the insert query, but the way we retrieve results. In other words, without the OutputTable handling, the query would still work, it just could not retrieve the results correctly (which is a concern with regard to the return value of the execute function).

imnotjames:
The DDL it generates is not an identity, though.  It should be.  Not sure why - I think there's a bug elsewhere in typeorm.  Unrelated to this PR.

src/query-builder/InsertQueryBuilder.ts
dklischies: In general, `null` is not a valid value for a primary key (because that is a clustered index that determines on-disk ordering of entries). Most databases (including mssql) prevent you from doing this.

Now, one could argue that this should fail instead of falling back to auto increment. However, that would be inconsistent with common SQL behavior, as databases cannot distinguish between undefined and null, and as such would insert a auto generated primary key if you supply null in a query.

src/query-builder/InsertQueryBuilder.ts
dklischies: Have a look at line 304. Either the table name is already escaped, or there is an issue with injection already.

dklischies:
> 
> 
> The DDL it generates is not an identity, though. It should be. Not sure why - I think there's a bug elsewhere in typeorm. Unrelated to this PR.

I'm pretty sure you are looking at the wrong DDL. Where does that description field come from?

imnotjames:
> > The DDL it generates is not an identity, though. It should be. Not sure why - I think there's a bug elsewhere in typeorm. Unrelated to this PR.
> 
> I'm pretty sure you are looking at the wrong DDL. Where does that description field come from?

From sql server studio - I also confirmed by inspecting the table manually.

dklischies:
Is there anything required from my side to get this merged? I'll happily discuss and make changes.
If you feel that you don't want to merge this at all please let me know, so I can stop rebasing/resolving conflicts.

fabianscheidt:
I am having the exact issue that seems to be solved with this PR. Is there a plan to merge this in the near future?

pleerock:
I usually prefer not to have these kind of changes (because of additional complexities it brings, and many "why"-s on many "if"-s. But you did a great job on investigation on how it works in sql server and I can't simply ignore these amount of efforts. Thank you for contribution!

geersch:
Debatable if this should be considered as a breaking change since the behaviour can change suddenly causing issues.
At the very least it is a new feature and I would like to see this reflected in the version number. Only the patch portion was updated giving me no indication of new features, let alone breaking changes. Should only be incremented for backwards compatible fixes.

Updated to this version recently and a lot of tests started failing on the CI server. Turned out we had some code which was generating a value of `0` for the primary, generated columns. Granted this is a bug on our part. TypeOrm of course now thought this value should be inserted into the IDENTITY column.

Perhaps it is time to drop the "anything can change at any time" leading 0 major version? `0.3.0` would now be considered a major update, so the only option you have now is to update the patch version.

src/query-builder/InsertQueryBuilder.ts
EatingJoey: What if it is '0'. I want to use the 'save()' method which can update and insert and when I want to insert then I set the data of identity column is 0. And it really insert 0 into the identity column. So I can not insert more data because it only update this  row.

EatingJoey:
So how you reslove this issue？ I face the same issue with you and I did not find the early version which don`t have this function.
Thanks

geersch:
@EatingJoey Browsing through the GIT history I see version `0.2.30` is the last version to not contain this change. 

IIRC we fixed it by making sure we do not specify an explicit value for the identity based column. That was a bug on our part.

EatingJoey:
> @EatingJoey Browsing through the GIT history I see version `0.2.30` is the last version to not contain this change.
> 
> IIRC we fixed it by making sure we do not specify an explicit value for the identity based column. That was a bug on our part.

@geersch Thank you for your information.

wy193777:
Is there a way to disable `SET IDENTITY_INSERT ON` when using query builder or `getRepository().save()`? This feature makes trigger after insert and update unusable.

dklischies:
`SET IDENTITY_INSERT ON` is only set if you supply a value for a primary key during insertion. Do not supply a value, and it will not be set.

Do you have a test case for the broken behavior you are seeing?

wy193777:
Seems only first call of `getRepository().save()` doesn't have `SET IDENTITY_INSERT ON`, all following calls will have it and id will be automatically provided by typeorm. Let me try create a test for it.

wy193777:
I found the problem. `getRepository().save(data)` will give `data`'s object id if they are just inserted.

Issue #2216, opened 2018-05-25
Ability to capture Postgres Extension already installed output in logger?


awwong1:
**Issue type:**

[x] question

**Database system/driver:**

[x] `postgres`

**TypeORM version:**

[X] `latest` `0.2.0`

**Steps to reproduce or a small repository showing the problem:**

I'm currently using the `uuid` and `citext` column types in my entity.

```js
import {
  Column,
  Entity,
  PrimaryGeneratedColumn,
} from "typeorm";

@Entity()
export class Sample {
  @PrimaryGeneratedColumn("uuid") public uuid: string;

  @Column({ type: "citext", nullable: false })
  public lowercaseval: string;
}
```

This has a side effect of outputting the following to console whenever I start my application.

```text
NOTICE:  extension "uuid-ossp" already exists, skipping
NOTICE:  extension "citext" already exists, skipping
```

Is there any way to redirect or capture this output? I have already [implemented my own custom logger](https://github.com/typeorm/typeorm/blob/master/docs/logging.md#using-custom-logger) but I still see direct output to stdout.

Thanks again for such a great library!

PR #6215, opened 2020-06-06
feat: add option to pass postgres notice messages to the client logger

artysidorenko:
Additional option `logNotifications` for Postgres driver. When true, passes any notice messages or listen/notify events to the client logger with `info` level.

This should close off [issue 2216](https://github.com/typeorm/typeorm/issues/2216).

Please reach out in case of questions or feedback.


artysidorenko:
(update: tests were a little verbose and poorly organised, so re-worked them to be more robust + updated commit message to adhere to the conventions)

pleerock:
Thanks!

Issue #2544, opened 2018-07-18
Concrete type isn't written down in table inheritance


lev-kuznetsov:
**Issue type:**

[X] question
[X] bug report

**Database system/driver:**

[X] `postgres`
[X] `sqlite`

**TypeORM version:**

[X] `0.2.7` (or put your version here)

**Steps to reproduce or a small repository showing the problem:**

Maybe I'm doing something wrong; when I create a polymorphic container and save, writing down the entities the type written down for each entity is the supertype not the concrete type. Here's a snippet to reproduce, I also tried postgres with the same result

```ts
import {
  ChildEntity, Column, createConnection, Entity, ManyToOne, OneToMany, PrimaryGeneratedColumn, TableInheritance
} from 'typeorm'

@Entity()
export class ContainerOfThings {
  constructor(things?: Thing[]) {
    this.things = things
  }

  @PrimaryGeneratedColumn()
  private id: number

  @OneToMany(() => Thing, thing => thing.container, {cascade: ['insert', 'remove']})
  things: Thing[]
}

@Entity()
@TableInheritance({column: {type: 'varchar', name: 'type'}})
export abstract class Thing {
  @PrimaryGeneratedColumn()
  private id: number

  @ManyToOne(() => ContainerOfThings, c => c.things)
  container: ContainerOfThings
}

@ChildEntity()
export class ParticularThing extends Thing {
  constructor(foo: string = 'foo') {
    super()
    this.foo = foo
  }

  @Column({type: 'varchar'})
  foo: string
}

describe('class inheritance', () => {
  it('should write down type', async () => {
    const connection = await createConnection({
      synchronize: true,
      type: 'sqlite',
      database: 'inheritance-test',
      logging: true,
      entities: [ContainerOfThings, Thing, ParticularThing]
    })

    await connection.transaction(async em => {
      await em.getRepository(ContainerOfThings).save(new ContainerOfThings([
        new ParticularThing('bar'), new ParticularThing('baz'), new ParticularThing()
      ]))
    })
  })
})
```

If you look at the logs and also inspect the database you'll see it's written down `Thing` as type not `ParticularThing`

lev-kuznetsov:
I saw https://github.com/typeorm/typeorm/issues/184

I tried saving using the concrete repository and the correct type is written down. After that I can use the base repository to findX and that gets resolved to the correct type which I would think would be the most difficult thing. So is writing down the base type when using the base repository a bug?

Krivich:
[...]

PR #6219, opened 2020-06-06
fix: Child entities not being saved correctly with cascade actions

ZBAGI:
`extractRelationValuesFromEntity` used `inverseEntityMetadata` which may contains abstract or base entity metadata which coursed child entity that we actually want to save to be treated as it's parent. This commit adds additional check for it.

Fixes: #2544


Garethp:
@pleerock Is there anything I can do, or anything I can contribute, to help get this PR merged?

pleerock:
test coverage in a separate PR is a good to have.

PR #6224, opened 2020-06-07
feat: add better-sqlite3 driver

K024:
* feat: better-sqlite3 driver which is significantly faster than node-sqlite3
* test: added all tests for sqlite to better-sqlite3
* test: "query runner > drop column" modified compatible
* docs: added better-sqlite3 related docs

### DIFF
--- a/docs/connection-options.md
+++ b/docs/connection-options.md
@@ -5,6 +5,7 @@
 * [`mysql` / `mariadb` connection options](#mysql--mariadb-connection-options)
 * [`postgres` / `cockroachdb` connection options](#postgres--cockroachdb-connection-options)
 * [`sqlite` connection options](#sqlite-connection-options)
+* [`better-sqlite3` connection options](#better-sqlite3-connection-options)
 * [`cordova` connection options](#cordova-connection-options)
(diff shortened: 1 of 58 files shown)


jackple:
I think `better-sqlite3` is a better idea!

pleerock:
quick question - is `better-sqlite3` and `sqlite3` api is the same? Maybe its better to change exist SqliteDriver file instead of so much duplication.

K024:
@pleerock They are looking similar but:
- `better-sqlite3` is fully sync api running on event loop (main) thread,
- `sqlite3` uses `napi_queue_async_work` to dispatch queries to libuv worker threads.

But it's strange to tell that `better-sqlite3` is much more performant on light weight queries (less blocking locks and context switches I guess) and shows no significant difference on heavy queries.

Actually sqlite3 is not suitable for concurrent queries, that's the duty of a full RDBMS. So I'm also considering the necessity of improving it.  

Btw, sqlite3 doesn't provide statement cache, nor the two driver libs. This could be influential if you never reuse statement objects (in many benchmarks on the web).

pleerock:
Thanks for explanation and your hard work! I think we can merge it.

Issue #6235, opened 2020-06-10
Mysql FindOperator In([]) will insert wrong sql


zstiu:
**Issue type:**

[ ] question
[x] bug report
[ ] feature request
[ ] documentation issue

**Database system/driver:**

[ ] `cordova`
[ ] `mongodb`
[ ] `mssql`
[x] `mysql` / `mariadb`
[ ] `oracle`
[ ] `postgres`
[ ] `cockroachdb`
[ ] `sqlite`
[ ] `sqljs`
[ ] `react-native`
[ ] `expo`

**TypeORM version:**

[x] `latest`
[ ] `@next`
[ ] `0.x.x` (or put your version here)

**Steps to reproduce or a small repository showing the problem:**   

   

I found this in toSql() function:
```ts
case "in":
                if ((connection.driver instanceof OracleDriver) && parameters.length === 0) {
                    return `${aliasPath} IN (null)`;
                }
                return `${aliasPath} IN (${parameters.join(", ")})`;
```

But when In() function (with mysql) get empty array, also insert wrong sql, so it should be like this:
```ts
case "in":
                if ((connection.driver instanceof OracleDriver || connection.driver instanceof MysqlDriver) && parameters.length === 0) {
                    return `${aliasPath} IN (null)`;
                }
                return `${aliasPath} IN (${parameters.join(", ")})`;
```

PR #6237, opened 2020-06-10
fix: insert IN(null) instead of IN() when In([]) empty array for mysq…

zstiu:
Close: #6235


pleerock:
Thanks! Would be great to have a test for this feature in a meantime!

Issue #4878, opened 2019-10-09
URL Connection string not overridden by supplied options


ryanmr:
**Issue type:**

[ ] question
[x] bug report
[ ] feature request
[ ] documentation issue

**Database system/driver:**

[ ] `cordova`
[ ] `mongodb`
[ ] `mssql`
[x] `mysql` / `mariadb`
[ ] `oracle`
[ ] `postgres`
[ ] `cockroachdb`
[ ] `sqlite`
[ ] `sqljs`
[ ] `react-native`
[ ] `expo`

**TypeORM version:**

[x] `latest`
[ ] `@next`
[ ] `0.x.x` (or put your version here)

**Steps to reproduce or a small repository showing the problem:**

From the docs:

> url - Connection url where perform connection to. Please note that other connection options will override parameters set from url.

In my particular situation, I pass `url` for host, port, database, etc in the TypeORM options but supply username and password directly.

```ts
    TypeOrmModule.forRoot({
      type: 'mysql',
      url: DB_CONNECTION, // CONNECTION STRING
      // for example: mysql://localhost:3306/dev
      username: DB_USERNAME, // abc
      password: DB_PASSWORD, // abc
      database: 'dev',
      entities: ['src/**/**.entity{.ts,.js}'],
    }),
```

The _username_ and _password_ are not set in the given connection string, and based on the documentation, the two values should have priority from the options vs the unspecified values in the connection string url.

It seems like [this could be reversed](https://github.com/typeorm/typeorm/blob/358e8a1a4872a950b471835cb48ff2bbc74cf6ce/src/driver/DriverUtils.ts#L31) and this would work as documented.

```ts
return Object.assign({}, options, urlDriverOptions);
```

```
// before
{
  type: 'mysql',
  url: 'mysql://localhost:3306/dev',
  username: '',
  password: '',
  database: 'dev',
  entities: [ 'src/**/**.entity{.ts,.js}' ],
  host: 'localhost',
  port: 3306
}

// after
{
  type: 'mysql',
  url: 'mysql://localhost:3306/dev',
  username: 'abc',
  password: 'abc',
  database: 'dev',
  entities: [ 'src/**/**.entity{.ts,.js}' ],
  host: 'localhost',
  port: 3306
}
```

[...]

PR #6247, opened 2020-06-11
 fix: properly override database url properties

Redmega:
As per the documentation, the url properties should be getting overridden by the specifically-passed configuration options. They currently are not because of the order of the `Object.assign` call.

This reverses the order of the assign to ensure that any properties which were already defined outside of the url are properly persisted in the final configuration.

There are various reasons for this PR. What drove me to make the change was CI/CD usage. In Gitlab, you are provided a DATABASE_URL, which can have a dynamic url and port due to the kubernetes setup. My typeorm connection kept connecting to the default user's database instead of the `test` database I was providing it, because the url property was overriding the provided `database`.

This PR fixes an issue opened October of last year! #4878

### DIFF
--- a/src/driver/DriverUtils.ts
+++ b/src/driver/DriverUtils.ts
@@ -28,7 +28,7 @@ export class DriverUtils {
             if (buildOptions && buildOptions.useSid) {
                 urlDriverOptions.sid = parsedUrl.database;
             }
-            return Object.assign({}, options, urlDriverOptions);
+            return Object.assign({}, urlDriverOptions, options);
         }
         return Object.assign({}, options);
     }
--- a/test/github-issues/4878/issue-4878.ts
+++ b/test/github-issues/4878/issue-4878.ts
@@ -1,0 +1,19 @@
+import { DriverUtils } from "../../../src/driver/DriverUtils";
+import { expect } from "chai";
+
+describe("github issues > #4878 URL Connection string not overridden by supplied options", () => {
+    it("should override url-built options with user-supplied options", () => {
+        const obj: any = {
+            username: "user",
+            password: "password",
+            host: "host",
+            database: "database",
+            port: 8888
+        };
+
+        const url = `postgres://url_user:${obj.password}@${obj.host}:${obj.port}/${obj.database}`;
+        obj.url = url;
+        const options = DriverUtils.buildDriverOptions(obj);
+        expect(options.username).to.eql(obj.username);
+    });
+});


Redmega:
@pleerock I'm sorry to tag you and I know you're quite busy, but this is getting in the way of our production release as we need to rely on the overriding of the url property due to CI/CD variables, and it seems like a minor PR.

This PR brings the code behaviour in line with the documented behaviour. Would welcome comments or feedback.

pleerock:
Okay, I'll merge it since this is how it supposed to work. However it can lead to a small breaking change, in the case if somebody relied on reversed order before.

I don't really like the way it works right now, since behaviour isn't obvious and requires documentation notes and catchups. Maybe in the future we'll change it, so only one way is acceptable and other way is leading to fatal error.

Redmega:
I agree, if someone tried to use it as documented, figured out it was reversed, and then is relying on that reversed behavior, that would be a breaking change for them.

However, I personally don't think that's a _bad_ breaking change, as if you are deliberately doing something contrary to how the docs say, without raising an alarm about it, it seems like that was a brittle application anyway.

ngraef:
> @pleerock: However it can lead to a small breaking change, in the case if somebody relied on reversed order before.

Just chiming in here to note that this change in v0.2.26 broke us.

We're getting static base options using `ConnectionOptionsReader` then applying a dynamic connection URL on top of it:

```ts
const connectionOptionsReader = new ConnectionOptionsReader({
  root: this.settings.ormConfigDirectory,
});

const connectionOptions = await connectionOptionsReader.get('default');

Object.assign(connectionOptions, { url: "<some_url>" });

const connection = await createConnection(connectionOptions);
```

The problem is the `connectionOptions` object [created by `ConnectionOptionsEnvReader`](https://github.com/typeorm/typeorm/blob/4baac7108652d11af2e06a7702c18179e556adbd/src/connection/options-reader/ConnectionOptionsEnvReader.ts#L20-L50) has a bunch of properties with `undefined` value because they're not set in the environment:

```ts
host: undefined,
port: undefined,
username: undefined,
password: undefined,
database: undefined,
```

So with the change in this PR, these `undefined` values overwrite the `urlDriverOptions` parsed out of the `url` and we end up with a `connectionOptions` object that doesn't have any connection information.

We'll come up with a workaround because it sounds like this is the intended behavior, but I wanted to add this to the conversation.

/cc @alvarocantador @keawade

Redmega:
That's a definite use-case that I did not consider! A check to ensure the overriding values are not undefined or null might be a good thing.

pleerock:
guys in case of breaking changes please consider directly ping me in slack. I'm going to revert this commit since there are other complains about this breaking change. I'll release a new 0.2.27 version.

jonathan-mothership:
@pleerock Did this ever make it into a 0.2.27 version? I don't see this in any tag: https://github.com/typeorm/typeorm/blob/0.2.34/src/driver/DriverUtils.ts#L31

This is blocking us form doing some nice overrides on the host.

Redmega:
It was reverted due to undefined values overwriting defined ones, but a PR to reimplement while addressing that issue would probably be accepted

Cprakhar:
@pleerock, what is the current status of addressing this issue for the upcoming release?

Issue #5999, opened 2020-05-04
PostgresQueryRunner not handling idle connection errors


joras:
**Issue type:**

[ ] question
[x] bug report
[ ] feature request
[ ] documentation issue

**Database system/driver:**

[ ] `cordova`
[ ] `mongodb`
[ ] `mssql`
[ ] `mysql` / `mariadb`
[ ] `oracle`
[x] `postgres`
[ ] `cockroachdb`
[ ] `sqlite`
[ ] `sqljs`
[ ] `react-native`
[ ] `expo`

**TypeORM version:**

[ x] `latest`
[ ] `@next`
[ ] `0.x.x` (or put your version here)

**Steps to reproduce or a small repository showing the problem:**

PostgresQueryRunner is not handling idle connection errors, thus all the connection related errors are unhandled and process usually dies with unexpected error.

```Error: Connection terminated unexpectedly
    at Connection.<anonymous> (/Users/jaan/Projects/typeormtest/node_modules/pg/lib/client.js:275:71)
    at Object.onceWrapper (events.js:417:28)
    at Connection.emit (events.js:311:20)
    at Connection.EventEmitter.emit (domain.js:482:12)
    at Socket.<anonymous> (/Users/jaan/Projects/typeormtest/node_modules/pg/lib/connection.js:126:10)
    at Socket.emit (events.js:323:22)
    at Socket.EventEmitter.emit (domain.js:482:12)
    at endReadableNT (_stream_readable.js:1204:12)
    at processTicksAndRejections (internal/process/task_queues.js:84:21)
```

It is not quite obvious when reading node-postgres documentation, but it is there under https://node-postgres.com/api/client - 'event' section. The *client returned from pool.connect() must handle on('error') events* because this is where idle connection error are returned. It seems pool.on('error') is not enough.

Also if you look node-postgres pool.query() function you can see they are handling on('error') **and** client.query() errors.

Now that said,  the issue is made worse by a bug in in node-postgres - when idle connection error handler is not attached, client.query() does not actually throw at all https://github.com/brianc/node-postgres/issues/2191 . 

Of course if this issue is fixed without fixing https://github.com/typeorm/typeorm/issues/5112 the connection would be returned to the pool until the pool fills with broken connections.

Small example to play with, if you stop postgres, the exceptions are not caught and process dies :
https://gist.github.com/joras/03bf07a7eb0ab0f7bc755f15b510b432

AlexDerbenew:
[...]

PR #6262, opened 2020-06-16
fix: handle 'error' events from pool connection

joras:
as touched in https://github.com/typeorm/typeorm/issues/5999

fixes following bug: connections returned from pg-pool do not have 'error' event handler. 'error' events are triggered on connection errors(https://node-postgres.com/api/client events section), thus when a connection error happens it will trigger uncaught exception and process is stopped.

fix:
Add an 'error' event handler to connections returned from pool. The error handler does not do anything else except releasing the pool connection with error - this will remove this connection from the pool, and this connection will not be reused. Note that there is no logging, or anything else in the handler, this is due to fact that if the event is handled and the process does not die, the query() promise will be correctly rejected, thus the error will be logged and handled there.

Some test code for reviewer to play with, run it, try to kill pg process, see how it handles before and after. I'm not sure if it is possible to write an unit test for this https://gist.github.com/joras/8ac861709a8f1b879944cfa3575adc43

### DIFF
--- a/src/driver/postgres/PostgresQueryRunner.ts
+++ b/src/driver/postgres/PostgresQueryRunner.ts
@@ -102,10 +102,14 @@ export class PostgresQueryRunner extends BaseQueryRunner implements QueryRunner 
      * Releases used database connection.
      * You cannot use query runner methods once its released.
      */
-    release(): Promise<void> {
+    release(err?: any): Promise<void> {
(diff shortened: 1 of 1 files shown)


RDeluxe:
This looks good to me and should be considered ASAP.

pleerock:
Thank you

PR #6269, opened 2020-06-16
feat: migrate to monorepo nrwl

luisvt:
Migrate to Monorepo using nrwl, this will allow to have every driver as a separate library which at the same time will allow an esier way to use this project in the browser


pleerock:
Sorry this is too huge change, don't think we can do it right now.

luisvt:
When it will be good time?

imnotjames:
Thanks for the contribution.  However, given the amount of churn this creates we'll address splitting the package into sub-packages closers to the release of the `next` branch.

PR #6274, opened 2020-06-17
docs: list all supported types for andWhere and orWhere

cainlevy:
SelectQueryBuilder's `where(), andWhere(), and orWhere() methods all support the same types, which are passed through to computeWhereParameter().

### DIFF
--- a/src/query-builder/SelectQueryBuilder.ts
+++ b/src/query-builder/SelectQueryBuilder.ts
@@ -717,7 +717,7 @@ export class SelectQueryBuilder<Entity> extends QueryBuilder<Entity> implements 
      * Adds new AND WHERE condition in the query builder.
      * Additionally you can add parameters used in where expression.
      */
-    andWhere(where: string|Brackets|((qb: this) => string), parameters?: ObjectLiteral): this {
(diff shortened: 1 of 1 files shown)


pleerock:
no, no, no. You shouldn't use ObjectLiteral for `.where` method too since it was created for internal purposes and not documented anywhere. Purpose of QueryBuilder is to create a string-based query. I mentioned it multiple times, in the future we'll plan to change current `.where` signature.

cainlevy:
Understood.

When you say "not documented anywhere" I think you mean on typeorm.io? But the types are documentation too. Changing the current `.where` signature is the only way to actually make it undocumented. Until then it's part of TypeORM's API, even if you wish it were not.

If you're still gathering developer feedback: I wish it could be documented. You undoubtedly have reasons (complexity? redundancy?) but as a developer I enjoy the ObjectLiteral syntax as the simplest way to express my most common need.

Thanks for all you've done. I know you're busy, and I do not expect a response.

pleerock:
It's just not designed to work with object literals on a QueryBuilder level. If you go deeper with this idea in QueryBuilder, you'll find many features might not work (because they were not designed, thus were not implemented) and you can arise many new issues.

cainlevy:
I appreciate the update! We'll be careful.

Issue #6284, opened 2020-06-20
support .cjs extension for orm config


dpickett:
**Issue type:**

[ ] question
[ ] bug report
[x] feature request
[ ] documentation issue

**Database system/driver:**

[ ] `cordova`
[ ] `mongodb`
[ ] `mssql`
[ ] `mysql` / `mariadb`
[ ] `oracle`
[ ] `postgres`
[ ] `cockroachdb`
[ ] `sqlite`
[ ] `sqljs`
[ ] `react-native`
[ ] `expo`

**TypeORM version:**

[x ] `latest`
[ ] `@next`
[ ] `0.x.x` (or put your version here)

I will have a pull request forthcoming for this. Need the issue # to write a test

imnotjames:
Seems like this was merged in and we're good to close it, yeah?

dpickett:
yes confirmed. Thanks! I'm closing this now.

PR #6285, opened 2020-06-20
support cjs extension for ormconfig

dpickett:
addresses #6284 - if you're running a NodeJS library that has ESM modules enabled, using a `cjs` extension for the `ormconfig` is required to express the configuration in commonjs.

This script allows me to run the CLI in a local environment and a test case is supplied.


pleerock:
Looks good, thank you

Issue #655, opened 2017-07-13
Referenced object doesn't get initialized (MongoDB)


Masterrg:
Hello everyone,

I am currently implementing a basic user/group system with TypeORM on MongoDB. I got the following relations:

1. Many users got one group
2. One group got many users

Everything gets saved correctly, the problem is that the referenced group object in the user object doesn't get loaded when I am loading my user object from the database.

**Small example:**
user.entity.ts
```ts
@Entity()
export class User {
    @ObjectIdColumn()
    id: string;

    @Column()
    name: string;

    @Column("string")
    @Index({unique: true})
    email: string;

    @Column()
    password: string;

    @ManyToOne(type => Group, group => group.users)
    group: Group;

    @Column()
    code_activation: string;

    @Column()
    code_forgotmail: string;

    @CreateDateColumn()
    registered: string;
}
```

group.entity.ts
```ts
@Entity()
export class Group {
    @ObjectIdColumn()
    id: string;

    @Column("int")
    power: number;

    @Column("string")
    @Index({unique: true})
    name: string;

    @OneToMany(type => User, user => user.group)
    users: User[];

    @UpdateDateColumn()
    updated: string;
}
```

testme.ts
```ts
let repository = await this.db.getRepository(User);
let user = await repository.findOne({email: email});

return user.group // is null
```

Best Regards

pleerock:
`ManyToOne`/`OneToMany`/`ManyToMany`/`OneToOne` - those decorators are only supported in relational databases, they are not supported by MongoDB. Mongodb is relational-less, so you can't do what you want easily as in relational databases. You should load your needed data (group) manually and set it to your entity (user) manually.

Masterrg:
Hello @pleerock and thank you very much for your reply,

are there any plans to add support for those decorators in future? We are trying to combine relational and non-relational strategies on MongoDB to create a flexible system which fits our needs - on just one database.

Joins are already possible since MongoDB 3.2 (https://docs.mongodb.com/manual/reference/operator/aggregation/lookup/). I think that this approach (at least in the end) results in less work in terms of coding and less round trips to the database, as everything is handled internally by MongoDB.

Nevertheless, I am dealing with this problem the way you already mentioned in your comment. In case other people are facing the same issue, let me just show you a snippet how to deal with it:


user.entity.ts
```ts
@Entity()
export class User {
    @ObjectIdColumn()
    id: string;

    @Column()
    name: string;

    @Column("string")
    @Index({unique: true})
[...]

PR #6300, opened 2020-06-24
fix: transform relations for aggregate $lookup

rhutchison:
$lookup aggregation support is lacking and I haven't seen much movement.  I am not very familiar with the library, but didn't want to just log another issue on this topic.  Please provide feedback and let me know how I can improve to get this supported in typeorm.

I've tested this with aggregateEntity queries, but have not fully tested for saves -- this probably needs work.  I will continue to run my own tests and add more commits, but I wanted to get eyes on this.

Resolves https://github.com/typeorm/typeorm/issues/655

User.ts
```ts
@Entity({ name: 'users' })
export class User implements IUser {
  @PrimaryGeneratedColumn()
  id: number;

  @Column()
  name: string;
}
```

Photo.ts
```ts
@Entity({ name: 'photos' })
export class Photo implements IPhoto {
  @PrimaryGeneratedColumn()
  id: number;

  @Column()
  url: string;

  @RelationId((photo: Photo) => photo.user)
  userId: number;

  @ManyToOne((type) => User)
  user: IUser;
}
```

photo.service.ts
```ts
  async findById(id: number) {
    const cursor = await this.repository.aggregateEntity([
      {
        $match: { id },
      },
      {
        $lookup: {
          from: 'users',
          localField: 'userId',
          foreignField: 'id',
          as: 'user',
        },
      }
    ]);

    const result = await cursor.limit(1).toArray();

    return result.length > 0 ? result[0] : undefined;
  }
```

### DIFF
--- a/src/query-builder/transformer/DocumentToEntityTransformer.ts
+++ b/src/query-builder/transformer/DocumentToEntityTransformer.ts
@@ -116,6 +116,20 @@ export class DocumentToEntityTransformer {
         addEmbeddedValuesRecursively(entity, document, metadata.embeddeds);
 
         // if relation is loaded then go into it recursively and transform its values too
+        metadata.relations.forEach(relation => {
+            const isResultArray = relation.isManyToMany || relation.isOneToMany;
+            const relatedEntities = document[relation.propertyName];
+
+            if (relatedEntities) {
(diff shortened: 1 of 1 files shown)


rhutchison:
The implementation is deeper than this PR.  I might revisit, but seems like focus on mongodb support in typeorm is not there.

PR #6314, opened 2020-06-26
refactor: update browser shims

its-lucas:
Update the browser shims to include new column types and repositories
resolving issues whereby the sharing of types resulting in compilation
issues.


pleerock:
thanks for contribution, it looks like I already merged similar PR. Can you please check it once? Does it have everything you have? Its a bit harder to review your PR since it contain a redundant formatting.

its-lucas:
Hi, it has the vast majority of things contained in this commit. Only thing I can see in this that isn't in there are the stubs for repositories which is useful for those exporting the entirety of their entity definitions where repositories may leak through. 

Sorry about the formatting trouble, it didn't appear that way during a local diff and must be a local prettier quirk.

pleerock:
I don't think you should share any code with repositories inside because its all server-specific, so I guess this PR can be closed.

PR #6316, opened 2020-06-26
escape columns in InsertQueryBuilder.orUpdate

Vrtak-CZ:
Fixies this use-case

```typescript
await repository
        .createQueryBuilder()
        .insert()
        .values({
            name,
            code,
            createdAt: new Date(),
            updatedAt: new Date(),
        })
        .orUpdate({
            conflict_target: ['name'],
            overwrite: ['code', 'updatedAt'],
        })
        .execute();
```

because it generates:

```sql
INSERT INTO "my_table"("id", "name", "code", "createdAt", "updatedAt") VALUES (DEFAULT, $1, $2, $3, $4) ON CONFLICT  ( name )  DO UPDATE SET code = EXCLUDED.code, updatedAt = EXCLUDED.updatedAt
```

and it causes:

```
ERROR:  column excluded.updatedat does not exist
LINE 1: ...  DO UPDATE SET code = EXCLUDED.code, updatedAt = EXCLUDED.u...
                                                             ^
HINT:  Perhaps you meant to reference the column "excluded.updatedAt".
```

instead of:

```sql
INSERT INTO "my_table"("id", "name", "code", "createdAt", "updatedAt") VALUES (DEFAULT, $1, $2, $3, $4) ON CONFLICT  ( "name" )  DO UPDATE SET "code" = EXCLUDED."code", "updatedAt" = EXCLUDED."updatedAt"
```

### DIFF
--- a/src/query-builder/InsertQueryBuilder.ts
+++ b/src/query-builder/InsertQueryBuilder.ts
@@ -267,16 +267,16 @@ export class InsertQueryBuilder<Entity> extends QueryBuilder<Entity> {
     orUpdate(statement?: { columns?: string[], overwrite?: string[], conflict_target?: string | string[] }): this {
       this.expressionMap.onUpdate = {};
       if (statement && Array.isArray(statement.conflict_target))
-          this.expressionMap.onUpdate.conflict = ` ( ${statement.conflict_target.join(", ")} ) `;
+          this.expressionMap.onUpdate.conflict = ` ( ${statement.conflict_target.map((columnName) => this.escape(columnName)).join(", ")} ) `;
       if (statement && typeof statement.conflict_target === "string")
-          this.expressionMap.onUpdate.conflict = ` ON CONSTRAINT ${statement.conflict_target} `;
+          this.expressionMap.onUpdate.conflict = ` ON CONSTRAINT ${this.escape(statement.conflict_target)} `;
       if (statement && Array.isArray(statement.columns))
-          this.expressionMap.onUpdate.columns = statement.columns.map(column => `${column} = :${column}`).join(", ");
+          this.expressionMap.onUpdate.columns = statement.columns.map(column => `${this.escape(column)} = :${column}`).join(", ");
       if (statement && Array.isArray(statement.overwrite)) {
         if (this.connection.driver instanceof MysqlDriver || this.connection.driver instanceof AuroraDataApiDriver) {
           this.expressionMap.onUpdate.overwrite = statement.overwrite.map(column => `${column} = VALUES(${column})`).join(", ");
         } else if (this.connection.driver instanceof PostgresDriver || this.connection.driver instanceof AbstractSqliteDriver || this.connection.driver instanceof CockroachDriver) {
-          this.expressionMap.onUpdate.overwrite = statement.overwrite.map(column => `${column} = EXCLUDED.${column}`).join(", ");
+          this.expressionMap.onUpdate.overwrite = statement.overwrite.map(column => `${this.escape(column)} = EXCLUDED.${this.escape(column)}`).join(", ");
         }
       }
       return this;
--- a/test/functional/query-builder/insert-on-conflict/entity/Post.ts
+++ b/test/functional/query-builder/insert-on-conflict/entity/Post.ts
@@ -1,8 +1,10 @@
 import {Entity} from "../../../../../src/decorator/entity/Entity";
 import {Column} from "../../../../../src/decorator/columns/Column";
+import {Unique} from "../../../../../src/decorator/Unique";
 import {PrimaryColumn} from "../../../../../src/decorator/columns/PrimaryColumn";
 
 @Entity()
+@Unique(["date"])
 export class Post {
 
     @PrimaryColumn()
@@ -10,5 +12,8 @@ export class Post {
 
     @Column()
     title: string;
+
+    @Column()
+    date: Date;
 
 }
--- a/test/functional/query-builder/insert-on-conflict/query-builder-insert-on-conflict.ts
+++ b/test/functional/query-builder/insert-on-conflict/query-builder-insert-on-conflict.ts
@@ -13,11 +13,12 @@ describe("query builder > insertion > on conflict", () => {
     beforeEach(() => reloadTestingDatabases(connections));
     after(() => closeTestingConnections(connections));
 
-    it("should perform insertion correctly", () => Promise.all(connections.map(async connection => {
+    it("should perform insertion correctly using onConflict", () => Promise.all(connections.map(async connection => {
 
         const post1 = new Post();
         post1.id = "post#1";
         post1.title = "About post";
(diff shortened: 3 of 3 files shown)


src/query-builder/InsertQueryBuilder.ts
```diff
@@ -255,16 +255,16 @@ export class InsertQueryBuilder<Entity> extends QueryBuilder<Entity> {
...
+          this.expressionMap.onUpdate.conflict = ` ( ${statement.conflict_target.map((columnName) => this.escape(columnName)).join(", ")} ) `;
       if (statement && typeof statement.conflict_target === "string")
-          this.expressionMap.onUpdate.conflict = ` ON CONSTRAINT ${statement.conflict_target} `;
+          this.expressionMap.onUpdate.conflict = ` ON CONSTRAINT ${this.escape(statement.conflict_target)} `;
       if (statement && Array.isArray(statement.columns))
-          this.expressionMap.onUpdate.columns = statement.columns.map(column => `${column} = :${column}`).join(", ");
+          this.expressionMap.onUpdate.columns = statement.columns.map(column => `${this.escape(column)} = :${column}`).join(", ");
       if (statement && Array.isArray(statement.overwrite)) {
         if (this.connection.driver instanceof MysqlDriver || this.connection.driver instanceof AuroraDataApiDriver) {
           this.expressionMap.onUpdate.overwrite = statement.overwrite.map(column => `${column} = VALUES(${column})`).join(", ");
```
pleerock, line 265: why did you ignore a mysql?

pleerock:
You are sure `conflict_target` and `overwrite ` only accept a column names as their values? Can you please add tests, so we can be sure functionality is properly working for all platforms and won't break in future

src/query-builder/InsertQueryBuilder.ts
Vrtak-CZ: i do not work with MySQL so i don't know if it's needed

Vrtak-CZ:
> You are sure `conflict_target` and `overwrite ` only accept a column names as their values?

I'm not.

> Can you please add tests, so we can be sure functionality is properly working for all platforms and won't break in future.

Few tests added

imnotjames:
Can you rebase against master so the tests kick off fully?

Vrtak-CZ:
@imnotjames done

imnotjames:
Unfortunately, some test failures.

```
query builder > insertion > on conflict 1) "before all" hook for "should perform insertion correctly" 2) "after all" hook for "should perform insertion correctly" 371 passing (2m) 9 pending 2 failing 1) query builder > insertion > on conflict "before all" hook for "should perform insertion correctly": DataTypeNotSupportedError: Data type "timestamp with time zone" in "Post.date" is not supported by "better-sqlite3" database. at new DataTypeNotSupportedError (build/compiled/src/error/DataTypeNotSupportedError.js:2:568) at /typeorm/build/compiled/src/metadata-builder/EntityMetadataValidator.js:29:2316 at Array.forEach (<anonymous>) at EntityMetadataValidator.validate (build/compiled/src/metadata-builder/EntityMetadataValidator.js:29:2045) at /typeorm/src/metadata-builder/EntityMetadataValidator.ts:27:100 at Array.forEach (<anonymous>) at EntityMetadataValidator.validateMany (src/metadata-builder/EntityMetadataValidator.ts:27:100) at Connection.buildMetadatas (src/connection/Connection.ts:177:29) at Connection.<anonymous> (build/compiled/src/connection/Connection.js:39:31) at step (node_modules/tslib/tslib.js:141:27) at Object.next (node_modules/tslib/tslib.js:122:57) at fulfilled (node_modules/tslib/tslib.js:112:62) 2) query builder > insertion > on conflict "after all" hook for "should perform insertion correctly": TypeError: Cannot read property 'map' of undefined at Object.closeTestingConnections (test/utils/test-utils.ts:289:36) at Context.<anonymous> (test/functional/query-builder/insert-on-conflict/query-builder-insert-on-conflict.ts:14:17) at processImmediate
```

src/query-builder/InsertQueryBuilder.ts
gintsgints: MySQL might not need this as it preserves case sensitivity for column names if no quotes. https://seeq.atlassian.net/wiki/spaces/KB/pages/443088907/SQL+Column+Names+and+Case+Sensitivity

test/functional/query-builder/insert-on-conflict/query-builder-insert-on-conflict.ts
```diff
@@ -51,8 +67,25 @@ describe("query builder > insertion > on conflict", () => {
...
+            .insert()
+            .into(Post)
+            .values(post2)
+            .orUpdate({
+                conflict_target: ['date'],
+                overwrite: ['title']
+            })
+            .execute();
+
+        await connection.manager.findOne(Post, "post#1").should.eventually.be.eql({
```
imnotjames, line 84: Can you create a separate `it` block just for this scenario?  Instead of including it in this one.

It means we can test edge cases separately and determine with them exactly what breaks

src/query-builder/InsertQueryBuilder.ts
imnotjames: Can we escape MySQL anyways as part of this to improve that side of things as well?

test/functional/query-builder/insert-on-conflict/entity/Post.ts
```diff
@@ -11,4 +11,7 @@ export class Post {
     @Column()
     title: string;
 
-}
\ No newline at end of file
+    @Column()
+    date: Date;
```
imnotjames, line 15: Why would adding this exhibit the failure?

imnotjames (commented):
Looks like a great change to make.  Test cases can be improved, maybe.

src/query-builder/InsertQueryBuilder.ts
gintsgints: In code for columns pleerock removed extra escaping from columns some two years ago :)

Vrtak-CZ: so what is the decision? should i add escaping for MySQL or not? I didn't work with MySQL for few years now so i do not know if it's needed or if it breaks something when it's escaped when you don't need to escape.

Vrtak-CZ:
is there anything more I can do?

src/query-builder/InsertQueryBuilder.ts
pleerock: @gintsgints do you have any insights on why we removed extra escaping?

pleerock: @Vrtak-CZ it will be consistent if we add for MySQL as well. Case sensitivity preservation is a configuration option that is different by default in different databases.

pleerock:
Sorry for delayed merge. Thank you for contribution!

Issue #3632, opened 2019-02-13
Embedded entities with entity schema


javix:
**Issue type:**

[x] question
[ ] bug report
[ ] feature request
[ ] documentation issue

**Database system/driver:**

[ ] `cordova`
[ ] `mongodb`
[ ] `mssql`
[ ] `mysql` / `mariadb`
[ ] `oracle`
[x] `postgres`
[ ] `sqlite`
[ ] `sqljs`
[ ] `react-native`
[ ] `expo`

**TypeORM version:**

[ ] `latest`
[ ] `@next`
[x] `0.2.12` (or put your version here)

**Steps to reproduce or a small repository showing the problem:**

In the actual project I can not use decorators, therefore I describe the entities with entity schema.
Now, I wont use embedded entities., but it does not work.

```
export const userSchema: EntitySchema<User> = new EntitySchema<User>( {
    name: 'user',
    target: User,
    columns: {
        username: {
            type: 'varchar',
            length: 20,
            unique: true
        },
        password: {
            type: 'varchar',
            length: 20,
            transformer: new PasswordValueTransformer()
        },
        lastLogin: {
            type: 'timestamp with time zone',
            nullable: true,
            name: 'lastlogin'
        }
    }
} );
```

```
export const metadataSchema: EntitySchema<Metadata> = new EntitySchema<Metadata>( {
    name: 'metadata',
    target: Metadata,
    columns: {
        createdBy: {
            type: 'varchar'
        },
        createdAt: {
            type: 'timestamp with time zone',
            createDate: true
        },
        modifiedBy: {
            type: 'varchar'
        },
        modifiedAt: {
            type: 'timestamp with time zone',
            updateDate: true
        }
    }
} );
```

```
export const dbUserSchema: EntitySchema<DbUser> = new EntitySchema<DbUser>( {
    name: 'dbUser',
    target: DbUser,
    columns: {
        id: {
            type: Number,
            primary: true,
            generated: true
        },
        user: {
            type: User   <--- compile failed
        },
    },
    relations: {
        user: {
            type: User   <--- compile failed
        },
        metadata: {
            type: Metadata   <--- compile failed
        }
    }
} );
```

Rovaro:
Check how EntitySchemas are specified: http://typeorm.io/#/separating-entity-definition.

This should work:
```javascript
export const dbUserSchema: EntitySchema<DbUser> = new EntitySchema<DbUser>( {
    name: 'dbUser',
    target: DbUser,
[...]

PR #6318, opened 2020-06-27
feat: embedded entities for entity schema

igoraguiar:
Allows to define embedded entities when using EntitySchema

Closes: #3632

### DIFF
--- a/docs/separating-entity-definition.md
+++ b/docs/separating-entity-definition.md
@@ -112,6 +112,44 @@ export const PersonSchema = new EntitySchema({
 });
 ```
 
+Example with embedded entity:
+
+```ts
+import {EntitySchema} from "typeorm";
+
+const AddressEntity = new EntitySchema({
+    name: 'address',
+    columns: {
+        city: {
+            type: String
+        },
+        zipcode: {
+            type: String
+        }
+    }
+});
+
+export const PersonEntity = new EntitySchema({
+    name: "person",
+    columns: {
+        id: {
+            type: Number,
+            primary: true,
+            generated: true
+        },
+        name: {
+            type: String
+        }
+    },
+    embeddeds: {
+        address: {
+            isArray: false,
+            type: () => AddressEntity
+        }
+    }
+});
+```
+
 If you want to make your entity typesafe, you can define a model and specify it in schema definition:
 
 ```ts
--- a/src/entity-schema/EntitySchemaEmbeddedOptions.ts
+++ b/src/entity-schema/EntitySchemaEmbeddedOptions.ts
@@ -1,0 +1,19 @@
+import { EntitySchema } from "./EntitySchema";
+
+export interface EntitySchemaEmbeddedOptions {
+    /**
+     * Type of the class to be embedded.
+     */
+    type: ((type?: any) => Function | EntitySchema);
+
+    /**
+     * Indicates if this embedded is array or not.
+     */
+    isArray: boolean;
+
+    /**
+     * Prefix of the embedded, used instead of propertyName. If set to empty string, then prefix is not set at all.
+     */
+    prefix?: string | boolean;
+
+}
--- a/src/metadata-args/EmbeddedMetadataArgs.ts
+++ b/src/metadata-args/EmbeddedMetadataArgs.ts
@@ -1,3 +1,5 @@
+import {EntitySchema} from "../entity-schema/EntitySchema";
+
 /**
  * Arguments for EmbeddedMetadata class.
  */
@@ -6,7 +8,7 @@ export interface EmbeddedMetadataArgs {
     /**
      * Class to which this column is applied.
      */
-    target: Function;
+    target: Function | string;
 
     /**
      * Class's property name to which this column is applied.
@@ -22,11 +24,11 @@ export interface EmbeddedMetadataArgs {
      * Prefix of the embedded, used instead of propertyName.
      * If set to empty string, then prefix is not set at all.
      */
-    prefix?: string|boolean;
+    prefix?: string | boolean;
 
     /**
      * Type of the class to be embedded.
      */
-    type: ((type?: any) => Function);
+    type: ((type?: any) => Function | EntitySchema);
 
 }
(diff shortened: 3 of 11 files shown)


src/entity-schema/EntitySchemaEmbeddedOptions.ts
```diff
@@ -0,0 +1,19 @@
+import { EntitySchema } from "./EntitySchema";
+
+export interface EntitySchemaEmbeddedOptions {
+    /**
+     * Type of the class to be embedded.
+     */
+    type: ((type?: any) => Function | EntitySchema);
```
pleerock, line 7: I think here we should point to either Function either entity schema name, e.g. `Function | string`, don't really want to have a circular deps within entity schemas too.

src/metadata-args/EmbeddedMetadataArgs.ts
```diff
@@ -22,11 +24,11 @@ export interface EmbeddedMetadataArgs {
...
      * If set to empty string, then prefix is not set at all.
      */
-    prefix?: string|boolean;
+    prefix?: string | boolean;
 
     /**
      * Type of the class to be embedded.
      */
-    type: ((type?: any) => Function);
+    type: ((type?: any) => Function | EntitySchema);
```
pleerock, line 32: ` | string` instead of ` | EntitySchema`

src/metadata-builder/EntityMetadataBuilder.ts
```diff
@@ -53,7 +54,7 @@ export class EntityMetadataBuilder {
     // -------------------------------------------------------------------------
 
     constructor(private connection: Connection,
-                private metadataArgsStorage: MetadataArgsStorage) {
+        private metadataArgsStorage: MetadataArgsStorage) {
```
pleerock, line 57: this file has much of redundant formatting...

docs/separating-entity-definition.md
```diff
@@ -112,6 +112,44 @@ export const PersonSchema = new EntitySchema({
 });
 ```
 
+Example with embedded entity:
+
+```ts
+import {EntitySchema} from "typeorm";
+
+const AddressEntity = new EntitySchema({
```
pleerock, line 120: since technically `entity` is an equivalent of a database `table`, it's better to call it `AddressEmbed` to avoid confusion.

docs/separating-entity-definition.md
```diff
@@ -112,6 +112,44 @@ export const PersonSchema = new EntitySchema({
...
+            generated: true
+        },
+        name: {
+            type: String
+        }
+    },
+    embeddeds: {
+        address: {
+            isArray: false,
+            type: () => AddressEntity
```
pleerock, line 147: would be great to have an ability to specify a name too (by embed name instead of variable reference)

src/entity-schema/EntitySchemaEmbeddedOptions.ts
pleerock, line 7: What `Function` can be here?

src/metadata-builder/EntityMetadataBuilder.ts
```diff
@@ -556,30 +557,57 @@ export class EntityMetadataBuilder {
     protected createEmbeddedsRecursively(entityMetadata: EntityMetadata, embeddedArgs: EmbeddedMetadataArgs[]): EmbeddedMetadata[] {
         return embeddedArgs.map(embeddedArgs => {
             const embeddedMetadata = new EmbeddedMetadata({ entityMetadata: entityMetadata, args: embeddedArgs });
-            const targets = MetadataUtils.getInheritanceTree(embeddedMetadata.type);
+            if (typeof embeddedMetadata.type === "function") {
```
pleerock, line 560: why do we need these changes? Basically, we shouldn't change this file and only `EntitySchemaTransformer` should be changed.

src/entity-schema/EntitySchemaEmbeddedOptions.ts
pleerock: as I understood it can be a non-entity-schema entity

src/metadata/EmbeddedMetadata.ts
```diff
@@ -32,7 +33,7 @@ export class EmbeddedMetadata {
     /**
      * Embedded target type.
      */
-    type: Function;
+    type: Function | EntitySchema;
```
pleerock, line 36: I think what we should do instead is to mark `type` as optional, and don't store anything here if type is entity schema is used.

src/metadata/EmbeddedMetadata.ts
```diff
@@ -190,7 +191,7 @@ export class EmbeddedMetadata {
      * Creates a new embedded object.
      */
     create(): any {
-        return new (this.type as any);
+        return typeof this.type === "function" ? new (this.type as any) : {};
```
pleerock, line 194: and here we'll have a check for type not being undefined.

src/metadata-builder/EntityMetadataBuilder.ts
pleerock: as I wrote below, we need to make embeddedMetadata.type equal to Function | undefined then we can rewrite this code just a bit and avoid duplication.

src/metadata-args/EmbeddedMetadataArgs.ts
pleerock, line 32: Lets make it undefined if entity schema is used.

pleerock:
Great additional to entity schemas, thank you

igoraguiar:
Hi @pleerock .
Just letting you know that a saw and agree with the reviews.
I will be glad to work on them as soon as I have the time, which is probably next week.
Thank you for this great library.

Yorex93:
Currently needing this myself as I'm trying to implement clear separations in a DDD styled project. Thanks @igoraguiar and @pleerock

jeesus:
What's the status with this?

db306:
Can we help anyhow ?

imnotjames:
@db306 

> Can we help anyhow ?

If you make the changes requested & open a new PR I will close this one and review yours.

toneysix:
any progress in it? from my perspective, it's the most important feature which ought to have kept up with embedded entities feature (e.g. by means of decorators)

pleerock:
I have to close this PR because work is not finished and its abandoned. Please feel free to create a new PR with required changes when you'll have a time.

PR #6324, opened 2020-06-28
style: migrate from slave/master terms to primary/replica

imnotjames:
this change updates terminology to "primary"/"replica"

more inclusive terminology can lead to a more inclusive project

the terminology master/slave is considered non-inclusive by many
and most databases that this project uses are also migrating away
from the terminology.

this leaves the original "master" and "slave" terminology in place
but marks them as deprecated & points them at the primary/replica
terminology.  at the next breaking change branch these can be removed
with a warning about the breaking change in the changelog

### DIFF
--- a/docs/multiple-connections.md
+++ b/docs/multiple-connections.md
@@ -36,7 +36,7 @@ const connections = await createConnections([{
 }]);
 ```
 
-This approach allows you to connect to any number of databases you have 
+This approach allows you to connect to any number of databases you have
 and each database will have its own configuration, own entities and overall ORM scope and settings.
 
 For each connection a new `Connection` instance will be created.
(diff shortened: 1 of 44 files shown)


imnotjames:
I did my best to ensure this wouldn't break existing code that may rely on this.

However, if we wanted to make this into a breaking change that could be arranged. :)

pleerock:
Unfortunately this is a breaking change, and we can't do it right now because it will break user code. But in the future versions, I'm sure we'll apply these changes.

imnotjames:
This is a breaking change?

imnotjames:
Either way, though, thanks for looking at it!  Looking forward to when this change can be made.  Happy to help out with it against whatever branch is the next breaking release. :)

imnotjames:
I'll close this for now so we don't clutter up the pull requests with things that aren't gonna be merged in right away. :)

PR #6335, opened 2020-07-01
fix: allow simple-json columns to hold non-object types (fixes #5501)

### LINKED ISSUE #5501: Incorrect data loading from JSON string for column type 'simple-json'. Why?
**Issue type:**

[x] question
[x] bug report
[ ] feature request
[ ] documentation issue

**Database system/driver:**

[ ] `cordova`
[ ] `mongodb`
[ ] `mssql`
[x] `mysql` / `mariadb`
[ ] `oracle`
[ ] `postgres`
[ ] `cockroachdb`
[ ] `sqlite`
[ ] `sqljs`
[ ] `react-native`
[ ] `expo`

**TypeORM version:**

[x] `latest`
[ ] `@next`

Why are some correct JSON values returned as Object when typeorm loading JSON string (simple-json)?
```ts 
JSON.parse('{"foo": "bar"}') // { foo: 'bar' }
JSON.parse('[ 1, 2, 3]') // [1, 2, 3]
JSON.parse('true') // true => But typeorm return: { }
JSON.parse('123') // 123 => But typeorm return: { }
JSON.parse('"message"') // 'message' => But typeorm return: { }
```
In my project, I noticed this strange behavior.
- Data inside entity is correct before saving.
- Inside database, data is also correct.
- But loaded entity has incorrect data.

_Example - Part of my problem:_
```ts

@Entity()
export class Foo {
  @Column('simple-json')
  returnValue: Json = false
} 

// ...

let foo = await connection.manager.save(new Foo())
// foo => Foo { returnValue: false }
// Inside DB => select 'foo'.'returnValue' from 'foo' => returnValue: "false"
let findFoo = await connection.getRepository(Foo).findOneOrFail()
// findFoo => Foo { returnValue: {} }
```

I looked into typeorm code and I found [explicit test on type Object](https://github.com/typeorm/typeorm/blob/8c67682fa41b8336cd96150baf654adcf5f954b5/src/util/DateUtils.ts#L180) before returning parsed values.

**Why is this behavior chosen?**

dapriett:


pleerock:
This basically reverts #4476 . First, you need ask initial authors why then made this change, then you can explain why change was wrong and your proposal is right. At the end we need to came to conclusion what we should do.

imnotjames:
I'm going to close this in favor of #6574

PR #6336, opened 2020-07-01
fix: synchronize column comment in mysql driver

lusess123:
column comment  change synchronize to database


pleerock:
I don't know why old code was comment, I guess for some particular reason. Since changes in schema sync can lead to problems I would avoid merging this PR for now, until @AlexMesser or somebody else can properly check it.

imnotjames:
Looks like it was commented out for all drivers at some point because it wasn't available when the code was being originally written.

@lusess123 if you could write a test to validate that this is working as expected that'd be stellar to get this merged in!

PR #6344, opened 2020-07-03
docs: Add advance subdocument query documentation

Ibcs16:
Easy peasy examples for querying subdocuments and array of subdocuments


pleerock:
Thanks!

PR #6372, opened 2020-07-09
Fix getPendingMigrations

axos88:
getPendingMigrations currently returns the executed migrations instead of the non-executed ones.


pleerock:
Thank you!

PR #6380, opened 2020-07-10
Update documentation for RelationQueryBuilder

fstn:
RelationQueryBuilder is not compatible with transaction. I updated the doc to specify it.


pleerock:
Well, its not a desirable behaviour. If it works so, we should change current implementation and make it to work with transaction.

fstn:
Are you going to change it?

pleerock:
Feel free to contribute, or alternatively you can create an issue and let's wait if someone want to contribute.

Issue #4588, opened 2019-08-15
migration handling for multiple instances


rightisleft:
**Issue type:**

[ ] question
[X] bug report
[] feature request
[ ] documentation issue

**Database system/driver:**

[ ] `cordova`
[ ] `mongodb`
[ ] `mssql`
[ ] `mysql` / `mariadb`
[ ] `oracle`
[X] `postgres`
[ ] `cockroachdb`
[ ] `sqlite`
[ ] `sqljs`
[ ] `react-native`
[ ] `expo`

**TypeORM version:**

[ ] `latest`
[ ] `@next`
[X] `0.2.11` (or put your version here)

**Steps to reproduce or a small repository showing the problem:**

Given that there is a collection of separate node services running typeorm (service 0-9),
AND each service has a connection to a single postgres database,
WHEN service process 0 acquires a database lock on the migrations table
Then service instances 1-9 should NOT run the same migration 

**What actually happens**

Service 0 sets a lock on the migrations table
Service 1-9 pause and wait
Service 0 finishes the migration and releases the lock
Service 1-9 attempt to re-run the migration

**What Should Happen**

Services 1-9 should wait until the lock is released, then check to see if the migration they are executing _now_ exists in the migrations table AND skip if found

<!--
    To answer those questions you need to put "x" inside the square brackets, for example:
    [x] `mysql`
    [ ] `postgres`

    Also, please format your code properly (by taking code blocks into ```ts .... ```)
!>

jfairley:
It's really unfortunately that TypeORM _seems_ to not have some sort of locking mechanism to handle parallel executions. Here's what I coded into my app to support my clustered servers.

The following is literally the first thing my app does, and is coupled with the `migrationsRun: false,` configuration.

```typescript
/**
 * 1. create the database connection.
 * 2. run migrations
 * 2a. needs home-rolled migration locking mechanism, because TypeORM lacks one
 */
async function createDatabaseConnection(): Promise<Connection> {
  // connect to the database
  const connection = await createConnection();

  // create database lock for migrations
  do {
    try {
      await connection.query("CREATE TABLE `_lock` (`id` int NOT NULL AUTO_INCREMENT, PRIMARY KEY (`id`))");
      break;
    } catch (err) {
      // lock table already exists. wait for release and try again.
      logger.info("Database lock table (_lock) already exists. Waiting for release...");
      await new Promise(resolve => setTimeout(resolve, Math.random() * 1000));
    }
  } while (true);

[...]

PR #6388, opened 2020-07-12
feat(migrations): add lock mechanism

ruscon:
ref https://github.com/typeorm/typeorm/issues/4588

**Issue type:**

[ ] question
[ ] bug report
[ x ] feature request
[ ] documentation issue

**Database system/driver:**
All

**BC: ?**
No

**Description of implementation:**

```typescript
private readonly migrationsLock: {
    enabled: boolean; // by default false
    tableName: string; // by default "migrations_lock"
    strategy: "wait-for-unlock"  | "throw-error" | "skip-migrations"; // by default ""wait-for-unlock""
    waitForUnlockInterval: number; // by default 1000 ms
};
```

**strategies:**
1. **`"wait-for-unlock"`**  - `migrations:run` command, which was launched in parallel with another launch, will wait until `migrationsLock.tableName` is removed from the first process that created `migrationsLock.tableName` table
2. **`"throw-error"`** - `migrations:run` command, which was launched in parallel, seeing an existing `migrationsLock.tableName` table, will simply throw the `MigrationsLockTableAlreadyExistsError` error and stopping its work
3. **`"skip-migrations"`** - `migrations:run` command, which was launched in parallel, seeing an existing `migrationsLock.tableName` table, will completely ignore the start of migrations and finish its work without error

**on process "SIGINT" or "SIGTERM":**
if the `migrationsLock.tableName` table was created by a process that intercepted the signal, the `dropMigrationsLockTable` function will be triggered


ruscon:
@pleerock ping

src/driver/mongodb/MongoQueryRunner.ts
```diff
@@ -13,7 +13,7 @@ import {
     Code,
     Collection,
     CollectionAggregationOptions,
-    CollectionBulkWriteOptions,
+    CollectionBulkWriteOptions, CollectionCreateOptions,
```
ddimitrioglo, line 16: new line required

pleerock:
I'm trying to understand problem described in #4588 . Can you please help me to explain how is it possible to run migrations:run command in parallel? Because from my understanding even if you have a several nodes, you shouldn't run this command in parallel, you should run it on a single node.

ruscon:
> I'm trying to understand problem described in #4588 . Can you please help me to understand how is it possible to run migrations:run command in parallel? Because from my understanding even if you have a several nodes, you shouldn't run this command in parallel, you should run it on a single node.

When pushing the code to the master branch, the code automatically goes through the CD / CI and is uploaded to several instances of the same service (or different ones) asynchronously, for example, to heroku or kubernetes. At the same time, until the migrations are completed, we should not restart the services with the new code (there can be a lot of different situations).

pleerock:
So, from my understanding in a such workflow all your nodes gets the latest master and you have **automatic sync enabled** (or call run migrations, no matter), that's why you get this issue.

But, in my opinion you should avoid such workflow. Why you don't have a separate step in your pipeline that is going to run migrations before launching / restarting your nodes?

ruscon:
At the moment we are doing this, but this slows down the deploy of the code in our case by 15 minutes (this is a lot)

ruscon:
I am sure that for many developers it will be much easier to use this lock mechanism than to reinvent their own wheel

pleerock:
> but this slows down the deploy of the code in our case by 15 minutes (this is a lot)

This is a lot, but considering being a production environment and not so often deploys in theory, maybe you can live with it? Or, of course I don't know specifics of your project, but I believe you can improve this time by creating a separate microservice that's purpose would be to simply run connection and migrations (minimal code + minimal node dependencies = most probably faster build and execution time).

Please try to understand me, as many features I get merged, as harder it becomes to maintain this thing. So I would like to avoid features that I find redundant or too much opinionated, or features that imply a risky strategies. I see many likes on this PR, but I guess they come from your friends 🤣 I would like to get this feature a time and see new opinions from other people.

pleerock:
Not sure, but just as a suggestion: if you are using Kubernetes for clusterization, "StatefulSet" might be useful to you - you can check from your node, if its a master node ("0"-node), only then you need to run migrations. For all other nodes you just need to disable migrations and schema sync.

imnotjames:
While I think that a multi-instance setup is common enough that you'll run into these kinds of problems - I think that writing a wrapper around TypeORM's migration to add this functionality would make more sense than including it within.

I went to look at a few different similar ORMs - the following don't support this:

* Mikro-ORM
* Sequelize
* Django
* SQLAlchemy
* Laravel

On face value this doesn't add a ton of code but it does introduce a number of new failure states that we probably want to avoid.  It'd be one thing if we were using some sort of built in table locking mechanism to prevent multiple workers from touching the migration table at once, but implementing and maintaining our own pessimistic locking mechanism is a bit much.

pleerock:
I think we should have this feature. There are many 👍 on this PR and it does matter. Also, I have faced this issue in a real world, and my solution might be not perfect for everyone and having this improvement might be handy for many people.

> I went to look at a few different similar ORMs - the following don't support this:

some of ORMs are really less immature... so I wouldn't use comparison in this case.

>  It'd be one thing if we were using some sort of built in table locking mechanism to prevent multiple workers from touching the migration table at once, but implementing and maintaining our own pessimistic locking mechanism is a bit much.

do we have any database-level options to implement this feature? If yes, and it would be simpler - let's do it. If no, we need to improve this PR and merge it.

Regarding to this PR - to merge it, I have few notes:

* I suggest following option signature:

```ts
private readonly migrationsLock?: {
    strategy: "none" | "wait-for-unlock"  | "throw-error" | "skip-migrations"; // "none" is default
    waitForUnlockInterval: number; // by default 1000 ms
};
```

* use `typeorm_metadata` table, because we already have some other functionality storing metadata information, I would like to avoid creating many typeorm-specific tables, and store everything inside one table (table name isn't customizable yet, but we can add this as another feature). Also, we don't really need to drop this table in case if something fails.

Also, can you please provide a real-world scenarios where "throw-error" and "skip-migrations" are really needed?

But again, before applying these changes I would like to be sure if there are no ways to implement same logic using database-level locks. 

Sorry @ruscon for being slow on reviewing this one.

imnotjames:
> I think we should have this feature. There are many 👍 on this PR and it does matter. 

It definitely matters - but creating an explicit locking mechanism to prevent a script from running multiple times is much easier for a specific situation when it's your codebase.  A generalized solution that uses a pessimistic locking mechanism is very difficult to maintain long term.

I'm open for other solutions, of course!

> some of ORMs are really less immature... so I wouldn't use comparison in this case.

Only Mikro ORM :)

You can also add Ruby on Rails to the list.  It doesn't implement an explicit lock.

[Like the other more mature ORMs it also uses a database level locking mechanism.](https://github.com/rails/rails/blob/master/activerecord/lib/active_record/migration.rb#L1238-L1244)

Perhaps we could get something like that going?

hwaterke:
@imnotjames Rails uses an advisory lock and has been for 5 years.
See https://github.com/rails/rails/commit/2c2a8755460ec3d32ece91c9766dbd0304ece028

Ecto (Elixir) also performs a lock guaranteeing two different servers cannot run the same migration as the same time.

I think this feature is clearly relevant.

imnotjames:
> @imnotjames Rails uses an advisory lock and has been for 5 years.
> See [rails/rails@2c2a875](https://github.com/rails/rails/commit/2c2a8755460ec3d32ece91c9766dbd0304ece028)

Yup, that's what I had more or less said in my comment.

There's no explicit table that defines locks.

Instead, it uses database-level locking features to implement the locking mechanism - which I'd noted in my comment.

This is also what I suggested we should do instead of creating a table that persists.

imnotjames:
I'm all for this if we can use database features to make it happen. 

I think it's a mistake if we decide to write our own locking mechanism using tables.

pleerock:
Okay, so what shall we do with this PR? Close it?

Based on discussion it looks like we should go with database-level locking features (e.g. advisory locks) since its more reliable solution. But more complex to implement.

Issue #6389, opened 2020-07-12
bug: mongodb url with query params is incorrectly parsed


ruscon:
**Issue type:**

[x] bug report

**Database system/driver:**

[x] `mongodb`

**TypeORM version:**

[x] the bug is reproduced starting from `0.2.25` (0.2.24 is OK)

**Steps to reproduce or a small repository showing the problem:**

Mongo DB URL without query params is OK
`TYPEORM_URL = mongodb://127.0.0.1:27017/testing`
the database called `testing` will be created

Mongo DB URL with query params (bug)
`TYPEORM_URL = mongodb://127.0.0.1:27017/testing?readPreference=primary&appName=xxx`
the database called `testing?readPreference=primary&appName=xxx` will be created

The bug appeared in this commit (src/connection/Connection.ts:538):
https://github.com/typeorm/typeorm/commit/ed75d59f27e673c58c4c7bfef85835c132ac5836#diff-9ada25e7bb500fdf98eeb214f7c94e9aR538

PR #6390, opened 2020-07-12
fix(driver-utils): correctly parse connection URI with query params

ruscon:
ref #6389

**BC?:**
No

In the future, it may be better to use this library
https://github.com/vitaly-t/connection-string


ruscon:
@pleerock ping

pleerock:
Thank you!

PR #6392, opened 2020-07-13
chore(package): update typescript to version 3.9.6

ruscon:
\+ an example of using the new typescript functionality "?" in code

### DIFF
--- a/src/entity-manager/EntityManager.ts
+++ b/src/entity-manager/EntityManager.ts
@@ -161,17 +161,17 @@ export class EntityManager {
     /**
      * Creates a new query builder that can be used to build a sql query.
      */
-    createQueryBuilder<Entity>(entityClass: ObjectType<Entity>, alias: string, queryRunner?: QueryRunner): SelectQueryBuilder<Entity>;
+    createQueryBuilder<Entity = any>(entityClass: ObjectType<Entity>, alias: string, queryRunner?: QueryRunner): SelectQueryBuilder<Entity>;
 
     /**
      * Creates a new query builder that can be used to build a sql query.
      */
-    createQueryBuilder<Entity>(entityClass: EntitySchema<Entity>, alias: string, queryRunner?: QueryRunner): SelectQueryBuilder<Entity>;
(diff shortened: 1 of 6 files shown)


src/entity-manager/EntityManager.ts
```diff
@@ -161,17 +161,17 @@ export class EntityManager {
     /**
      * Creates a new query builder that can be used to build a sql query.
      */
-    createQueryBuilder<Entity>(entityClass: ObjectType<Entity>, alias: string, queryRunner?: QueryRunner): SelectQueryBuilder<Entity>;
```
pleerock, line 164: I don't think we should have `= any` here. What is the motivation behind this change?

src/query-builder/SelectQueryBuilder.ts
```diff
@@ -181,21 +181,21 @@ export class SelectQueryBuilder<Entity> extends QueryBuilder<Entity> implements
      * Also sets a main string alias of the selection data.
      * Removes all previously set from-s.
      */
-    from<T>(entityTarget: (qb: SelectQueryBuilder<any>) => SelectQueryBuilder<any>, aliasName: string): SelectQueryBuilder<T>;
+    from<T = any>(entityTarget: (qb: SelectQueryBuilder<any>) => SelectQueryBuilder<any>, aliasName: string): SelectQueryBuilder<T>;
```
pleerock, line 184: same comment here

src/entity-manager/EntityManager.ts
ruscon: > I don't think we should have `= any` here. What is the motivation behind this change?

Yes, I think it should be unknown, but many tests fall

pleerock: I don't think we should put a default type at all, what can be a motivation? It looks like a monkey patching instead of root problem resolving solution.

ruscon: In the latest typescript version, for example in this block `createQueryBuilder<Entity>`, `Entity` by default will be `unknown` (in the current version used it is interpreted as `any`)
You can just upgrade the typescript version and see the difference

Issue #5444, opened 2020-01-30
EntitySchema missing support for multiple joinColumns in relations


pwbrown:
**Issue type:**

[x] bug report

**TypeORM version:**

[x] `0.2.22`

**Inconsistent Feature:**

When defining a `JoinColumn` using decorators, it accepts an array of JoinColumnOptions:
```
@Entity()
class Example {
    .
    .
    .
    @ManyToOne(type => Entity, entity => entity.col)
    @JoinColumn([
        { name: 'col1', referencedColumnName: 'col1' },
        { name: 'col2', referencedColumnName: 'col2' },
    ])
    relationship!: Entity;
}
```

However when defining a relation using an EntitySchema, the joinColumn option only accepts a single column object
```
interface Example {
    col1: string;
    relationship?: Entity;
}

const ExampleEntity = new EntitySchema<Example>({
    name: 'Example',
    columns: { ... },
    relation: {
        relationship: {
            type: 'many-to-one',
            target: 'Entity',
            inverseSide: 'col',
            /** ALLOWED */
            joinColumn: {
                name: 'col1',
                referencedColumnName: 'col2',
            },
            /** NOT ALLOWED */
            joinColumn: [
                { name: 'col1', referencedColumnName: 'col1'  },
                { name: 'col2', referencedColumnName: 'col2' },
            ],
        },
    },
})
```

**Code References:**

1. `src/decorator/relations/JoinColumn.ts` - Correct functionality reference (iterates over join columns and adds metadata for each)

lines 30-42
```
export function JoinColumn(optionsOrOptionsArray?: JoinColumnOptions|JoinColumnOptions[]): Function {
    return function (object: Object, propertyName: string) {
        const options = optionsOrOptionsArray instanceof Array ? optionsOrOptionsArray : [optionsOrOptionsArray || {}];
        options.forEach(options => {
            getMetadataArgsStorage().joinColumns.push({
                target: object.constructor,
                propertyName: propertyName,
                name: options.name,
                referencedColumnName: options.referencedColumnName
            } as JoinColumnMetadataArgs);
        });
    };
}
```

2. `src/entity-schema/EntitySchemaRelationOptions.ts` - definition should allow for an array of JoinColumnOptions

line 58 - ***incorrect***
```
joinColumn?: boolean|JoinColumnOptions;
```

line 58 - ***corrected***
```
joinColumn?: boolean|JoinColumnOptions|JoinColumnOptions[];
```

[...]

PR #6397, opened 2020-07-14
fix: support multiple `JoinColumn`s in EntitySchema

imnotjames:
update type definition and schema transformer
so that - like the decorator - the EntitySchema can define
composite `JoinColumn` definitions

Closes: #5444


imnotjames:
PS - I couldn't find anywhere that this was exercised in tests, so I created some under `test/github-issues/5444`.

Let me know if there's a different way to add them.

pleerock:
Thanks a lot!

Issue #2089, opened 2018-05-05
How can I use TreeEntity and Listener with EntitySchema?


codingnan:
**Issue type:**

[x] question
[ ] bug report
[ ] feature request
[ ] documentation issue

**Database system/driver:**

[ ] `cordova`
[ ] `mongodb`
[ ] `mssql`
[ ] `mysql` / `mariadb`
[ ] `oracle`
[x] `postgres`
[ ] `sqlite`
[ ] `sqljs`
[ ] `react-native`

**TypeORM version:**

[x] `latest`
[ ] `@next`
[ ] `0.x.x` (or put your version here)

**Steps to reproduce or a small repository showing the problem:**

How can I use TreeEntity and Listener with EntitySchema?

codingnan:
ok, thanks to answer my question.

imnotjames:
Hi there!

Until #6401 or similar gets merged in, I've created a monkeypatch which extends the transformer to support listeners.

```
export async function patchSupportListenersTransformer () {
  const { EntitySchemaTransformer } = await import('typeorm/entity-schema/EntitySchemaTransformer.js');

  const transform = EntitySchemaTransformer.prototype.transform;

  if ('isPatchedForIssue2089' in transform) {
    return;
  }

  EntitySchemaTransformer.prototype.transform = function (schemas) {
    const entityListeners = [];

    for (const schema of schemas) {
      const options = schema.options;

      if (!('listeners' in options)) {
        continue;
      }

      for (const { type, propertyName } of options.listeners) {
        entityListeners.push({
          target: options.target || options.name,
          propertyName,
          type
        });
      }

      delete options.listeners;
    }

    const metadataArgsStorage = transform.call(this, schemas);

    metadataArgsStorage.entityListeners.push(...entityListeners);

    return metadataArgsStorage;
  };

  EntitySchemaTransformer.prototype.transform.isPatchedForIssue2089 = true;
}
```

PR #6401, opened 2020-07-14
feat: support Entity Listeners in EntitySchema

imnotjames:
Update the `EntitySchema` to add support for `listeners` - `EntityListeners` similar to the [decorator listeners](https://typeorm.io/#/listeners-and-subscribers/what-is-an-entity-listener).

Also update `EntityListenerMetadataArgs.target` to support `string` as well as `Function`.  This should be supported per [MetadataArgsStorage.filterListeners](https://github.com/typeorm/typeorm/blob/1e1595e4c561995c60eef60426ef4f198d8b48cb/src/metadata-args/MetadataArgsStorage.ts).

PARTIAL RESOLUTION for #2089 (Listener)

### DIFF
--- a/src/entity-schema/EntitySchemaListenerOptions.ts
+++ b/src/entity-schema/EntitySchemaListenerOptions.ts
@@ -1,0 +1,13 @@
+import {EventListenerType} from "../metadata/types/EventListenerTypes";
+
+export interface EntitySchemaListenerOptions {
+    /**
+     * Type of Event to Listen to
+     */
+    type: EventListenerType;
+
+    /**
+     * Entity's property name to which listener is applied.
+     */
+    propertyName: string;
+}
--- a/src/entity-schema/EntitySchemaOptions.ts
+++ b/src/entity-schema/EntitySchemaOptions.ts
@@ -7,6 +7,7 @@ import {TableType} from "../metadata/types/TableTypes";
 import {EntitySchemaUniqueOptions} from "./EntitySchemaUniqueOptions";
 import {EntitySchemaCheckOptions} from "./EntitySchemaCheckOptions";
 import {EntitySchemaExclusionOptions} from "./EntitySchemaExclusionOptions";
+import {EntitySchemaListenerOptions} from "./EntitySchemaListenerOptions";
 
 /**
  * Interface for entity metadata mappings stored inside "schemas" instead of models decorated by decorators.
@@ -86,6 +87,11 @@ export class EntitySchemaOptions<T> {
     * Entity exclusion options.
     */
     exclusions?: EntitySchemaExclusionOptions[];
+
+    /**
+     * Entity listener options.
+     */
+    listeners?: EntitySchemaListenerOptions[];
 
     /**
      * Indicates if schema synchronization is enabled or disabled for this entity.
--- a/src/entity-schema/EntitySchemaTransformer.ts
+++ b/src/entity-schema/EntitySchemaTransformer.ts
@@ -13,6 +13,7 @@ import {GeneratedMetadataArgs} from "../metadata-args/GeneratedMetadataArgs";
 import {UniqueMetadataArgs} from "../metadata-args/UniqueMetadataArgs";
 import {CheckMetadataArgs} from "../metadata-args/CheckMetadataArgs";
 import {ExclusionMetadataArgs} from "../metadata-args/ExclusionMetadataArgs";
+import {EntityListenerMetadataArgs} from "../metadata-args/EntityListenerMetadataArgs";
 
 /**
  * Transforms entity schema into metadata args storage.
@@ -237,6 +238,18 @@ export class EntitySchemaTransformer {
                         expression: exclusion.expression
                     };
                     metadataArgsStorage.exclusions.push(exclusionArgs);
+                });
+            }
+
+            // add listener metadata args from the schema
+            if (options.listeners) {
+                options.listeners.forEach(listener => {
+                    const listenerArgs: EntityListenerMetadataArgs = {
+                        target: options.target || options.name,
+                        propertyName: listener.propertyName,
+                        type: listener.type
+                    };
+                    metadataArgsStorage.entityListeners.push(listenerArgs);
                 });
             }
 
--- a/src/metadata-args/EntityListenerMetadataArgs.ts
+++ b/src/metadata-args/EntityListenerMetadataArgs.ts
@@ -8,7 +8,7 @@ export interface EntityListenerMetadataArgs {
     /**
      * Class to which listener is applied.
      */
-    readonly target: Function;
+    readonly target: Function|string;
 
     /**
      * Class's property name to which listener is applied.


src/entity-schema/EntitySchemaListenerOptions.ts
```diff
@@ -0,0 +1,13 @@
+import {EventListenerType} from "../metadata/types/EventListenerTypes.js";
```
pleerock, line 1: can you please remove `.js` extension from all imports in all files?

pleerock:
I don't really want to merge this feature, because of design considerations. 

Entity schemas are recommended for usage in the case if you don't prefer to use classes and decorators (personally I always use entity schemas in my projects). If you don't use classes and decorators you are dealing with literal objects ("{ ... }"). I find it a bad practice to have a methods / functions in your dummy model literal objects. Instead, I believe they should be in a separate classes / utility functions. It's not only solves you many problems you might have with having functions/methods inside your models, but is also a separation of concerns. 

So, IMO we should have this feature for entity schemas, but it should look different (not to call entity object property, but instead list all functions that should be executed on a particular event). Should this list of functions be a part of entity schema definition? It's a separate question, but if you want to proceed with the changes I proposed - we can discuss it.

imnotjames:
> Entity schemas are recommended for usage in the case if you don't prefer to use classes and decorators (personally I always use entity schemas in my projects).

In our case, we aren't using TypeScript so decorators are a bit difficult anyway!

> If you don't use classes and decorators you are dealing with literal objects ("{ ... }"). I find it a bad practice to have a methods / functions in your dummy model literal objects. Instead, I believe they should be in a separate classes / utility functions. It's not only solves you many problems you might have with having functions/methods inside your models, but is also a separation of concerns.

Having these live outside of the entity models would require a bit of a refactor - perhaps a separate code path?  There's a few approaches that could work - some of which are breaking changes (eg, updating decorator listeners to pass the function they're decorating directly instead of property name), others less so.

I think that would be a more interesting way to approach the problem.  Calling `propertyName` is definitely kind of kludgy here -

> So, IMO we should have this feature for entity schemas, but it should look different (not to call entity object property, but instead list all functions that should be executed on a particular event). Should this list of functions be a part of entity schema definition? It's a separate question, but if you want to proceed with the changes I proposed - we can discuss it.

This was mostly to bring the `EntitySchema` in-line feature-wise with the decorators.  If this were to be merged in I'd still be interested in implementing a more elegant approach to the feature - and supporting both `EntitySchema` & Decorators.

src/entity-schema/EntitySchemaListenerOptions.ts
imnotjames: Ah, sure!  A bit of muscle memory from other projects. :)

Might be good to bring in `eslint-plugin-import` & enforce via linting as well.

temp/.gitignore
```diff
@@ -1,2 +0,0 @@
-**
```
imnotjames, line 1: Oops!  I didn't mean to remove this.  I had thought it was a file I had accidentally added.

src/entity-schema/EntitySchemaListenerOptions.ts
imnotjames: The extensions should be removed.

temp/.gitignore
imnotjames: I put the file back :)

pleerock:
Here is what I'm thinking about.

We have subscribers that looks like this:

```ts
@EventSubscriber()
export class UserEntitySubscriber implements EntitySubscriberInterface<UserEntity> {
  listenTo() {
    return UserEntity;
  }
  async beforeInsert(event: InsertEvent<UserEntity>) {
    event.entity.fullname = event.entity.firstname + " " + event.entity.lastname;
  }
  async beforeUpdate(event: UpdateEvent<UserEntity>) {
    event.entity.fullname = event.entity.firstname + " " + event.entity.lastname;
  }
}
```

I think what we need to do is to introduce a new syntax that stays consistent with entity schemas, e.g.:

```ts

export const Listeners = {
  User: {
    beforeUpdate(event) {
      event.entity.fullName = event.entity.firstName + " " + event.entity.lastName
    },
  },
  SomeOtherEntity: {
    beforeInsert(event) {
      event.entity.fullName = event.entity.firstName + " " + event.entity.lastName
    },
  }
}
```

Use can specify multiple such listeners in the connection options, in the same "subscriptions" option. Its possible to make it to work with old syntax and new one.

We need to make it simple as possible, use **exist** code functionality as much as possible, e.g. reuse class-based subscription functionality as much as possible, or change it a bit if needed to support both cases.

pleerock:
so my idea is to forget about "entity listeners" because of the issues I listed in previous message and use subscriptions as a method what changing entity before / after some event. This new syntax is just alternative to class-based subscriptions.

imnotjames:
Ok - I was digging into another issue for forever (#6416) but finally getting back to this.

Maybe I'm a little confused - why is there a difference between `Subscribers` and `Listeners` anyway?

Is your suggestion to effectively bring them all together under one umbrella? Possibly deprecating the `Subscribers` with a translation layer to the new `Listeners` API & the decorators effectively defining new `Listeners` on the fly through this API?

pleerock:
> Maybe I'm a little confused - why is there a difference between Subscribers and Listeners anyway? The documentation isn't really clear and the features look rather similar in nature.

there is no difference, it's just a two different "ways to use" the same functionality, depend on user preference. 

> Is your suggestion to effectively bring them all together under one umbrella? 

Well, I just suggest to add another way of use. Right now we have two ways, just add another way:

* inside entity ("entity listeners")
* inside a separate class ("subscribers")
* inside a literal object ("new way")

Motivation? To keep it consistent with entity schemas, since many entity schemas prefer not to use classes since they are redundant in most cases, including this one.

> Possibly deprecating the Subscribers with a translation layer

no, we won't deprecate anything for now, just adding new functionality.

imnotjames:
What if we did roughly this PR but instead had

```

import {EventListenerType} from "../metadata/types/EventListenerTypes";

export interface EntitySchemaListenerOptions {
    /**
     * Type of Event to Listen to
     */
    type: EventListenerType;

    /**
     * Entity's property name to which listener is applied.
     */
    callback: Function;
}
```

Perhaps even matching up `type` as a discriminator value so the callback types matched the subscriber interfaces?

pleerock:
My goal is to avoid having a function inside entity, that's why I want to avoid approach that we use in classes and something proposed in this PR.

imnotjames:
If we use the subscriber definition you'd suggested we wouldn't be able to target entities that haven't been given an explicit name - eg, ones by which we'd normally use `listenTo` for referencing by the class name of the entity.  (`{ User: { afterLoad() {} } }` refers to the string `"User"` and not a class named `User`.  If it can't support both that seems like a step backward.  You don't have to name your entity schemas, they can still be based on classes.

If we want to have a subscriber that isn't class-based it can be done via a few minor changes so we can pass the following as subscribers:

```
subscribers: [ { listenTo: () => Foo, afterLoad: foo => foo.bar() } ]
```

I could open a separate PR with that, but it doesn't feel like a huge win over the current mechanisms.

imnotjames:
Separately, it sounds like the `listener` feature won't ever be supported as part of Entity, and only for those that are using Typescript and the decorator format.  This means ECMAScript cannot use listeners without hacks & workarounds (nobody will do that).

I somewhat disagree on the function bit in schemas, but that's your call.  In a project I'd been wokring on were using functions in the entity schema to support transformers, targets by class name, as well as defining the `inverseSide` of relationships in our entity schema. :shrug:

By having to rely on subscribers it does mean that we have to split the logic that defines the entity as the generation of the ID was being done via the listener.  It can probably still be done via a subscriber that lives in the same file but that's a bit verbose.

PR #6404, opened 2020-07-14
fix(aurora): pass formatOptions to Data API Client, fix extensions

ArsenyYankovsky:
This PR fixes an issue where the `AuroraDataApiPostgresDriver` would incorrectly query the database to setup extensions (like UUID or hstore).

It also passes down the `formatOptions` to the driver so that different handling of dates could be configured.


src/driver/postgres/PostgresDriver.ts
```diff
@@ -1090,4 +1091,80 @@ export class AuroraDataApiPostgresDriver extends PostgresWrapper {
         return this.client.query(query);
     }
 
+    /**
+     * Makes any action after connection (e.g. create extensions in Postgres driver).
+     */
+    async afterConnect(): Promise<void> {
```
pleerock, line 1097: this code confused me a lot. And I just found that `AuroraDataApiPostgresDriver` in the same file as `PostgresDriver`. Can we have it in a separate file?

Also found there is a weird `PostgresWrapper`... Can we avoid it? 

Also why do we need this duplication with `afterConnect`? Even hard-coded to a specific implementation `if` check inside PostgresDriver would be better than this kind-of duplication.

ArsenyYankovsky: `PostgresConnectionOptions | AuroraDataApiPostgresConnectionOptions` option I don't think will work because it means we need to cast options everywhere we use them in the `PostgresDriver` class. 
I also tried moving `AuroraDataApiPostgresDriver ` out in a separate file but I found no way around the circular dependency issue.

PR #6417, opened 2020-07-17
fix: support combination of many-to-one/cacade/explicit composite PK 

### LINKED ISSUE #4122: .save() is throwing duplicate key value violates unique constraint
**Issue type:**

[x] question
[ ] bug report
[ ] feature request
[ ] documentation issue

**Database system/driver:**

[ ] `cordova`
[ ] `mongodb`
[ ] `mssql`
[ ] `mysql` / `mariadb`
[ ] `oracle`
[x] `postgres`
[ ] `cockroachdb`
[ ] `sqlite`
[ ] `sqljs`
[ ] `react-native`
[ ] `expo`

**TypeORM version:**

[x] `latest`
[ ] `@next`

This was working for me earlier but now it is not. I am using a composite key of three different properties. Every time I try to update and existing row, I get the error `QueryFailedError: duplicate key value violates unique constraint "PK_8b1375852f33d30b328388e5a5c"`.

This is my entity:
```
'use strict';

import * as _ from 'lodash';
import { Entity, Column, PrimaryColumn, Generated } from "typeorm";


@Entity()
export class Assortment {

	@Column("uuid")
	@Generated("uuid")
	id!: string;

	@PrimaryColumn()
	accountId!: string;

	@PrimaryColumn()
	entityType!: string;

	@PrimaryColumn()
	parentId!: string;

	@Column({
		nullable: true
	})
	parentType!: string;

	@Column('varchar', {
		nullable: true
	})
	sortedIds!: string[];

	@Column('jsonb', {
		nullable: true
	})
	children!: Assortment[];
};
```

### LINKED ISSUE #4969: Problem saving entity with cascade true
**Issue type:**

[ ] question
[x] bug report
[ ] feature request
[ ] documentation issue

**Database system/driver:**

[ ] `cordova`
[ ] `mongodb`
[ ] `mssql`
[ ] `mysql` / `mariadb`
[ ] `oracle`
[ ] `postgres`
[ ] `cockroachdb`
[x] `sqlite`
[ ] `sqljs`
[ ] `react-native`
[ ] `expo`

**TypeORM version:**

[x] `latest`
[ ] `@next`

Here is a repository to reproduce the bug:
[https://github.com/youtix/nest-typerorm-test](https://github.com/youtix/nest-typerorm-test)

I m trying to save a user conversation with messages. I set the messages field on the conversation entity to **_cascade: true_**. But When I try this code:
```
const user3: User = { login: 'admin', createdBy: 'system', lastModifiedBy: 'system' };
const user4: User = { login: 'user', createdBy: 'system', lastModifiedBy: 'system' };
const message1: Message = { content: 'Hello How are you? ', createdBy: user3.login, lastModifiedBy: user3.login };
const conversation1: Conversation = { sender: user3, reciever: user4, messages: [message1] };
getConnection().getRepository(Conversation).save(conversation1);
```
It creates this query:
```
INSERT INTO "message"("id", "createdBy", "createdDate", "lastModifiedBy", "lastModifiedDate", "content", "conversationSenderId", "conversationRecieverId", "conversationDate") VALUES (?, ?, datetime('now'), ?, datetime('now'), ?, ?, ?, ?)
-- PARAMETERS: ["67348880-6897-47cb-92ef-5f66ffb2e88c","admin","admin","Hello How are you? ","b2420445-c8ee-4080-86b3-98ab12ef576b","09f3672d-9b2f-4f68-b47e-a7ca5d806ec6","2019-10-24 14:41:40.000"]
```

with this error:

`{ [Error: SQLITE_CONSTRAINT: FOREIGN KEY constraint failed] errno: 19, code: 'SQLITE_CONSTRAINT' }`

I can see it's a problem with the foreign key of messages table but I can't fix it because it's TypeORM library which handle the creation, not me. I searched on internet (documentation, github, stackoverflow) without success. I think it's bug in TypeORM library but I m a novice in it so I prefer asking if I did something wrong

Here are my entities :

```
export abstract class BaseEntity {
  @ObjectIdColumn()
  id?: string;

  @Column()
  createdBy?: string;

  @CreateDateColumn()
  createdDate?: Date;

  @Column()
  lastModifiedBy?: string;

  @UpdateDateColumn()
  lastModifiedDate?: Date;
}
```

```
@Entity('user')
export class User extends BaseEntity {
  @Column()
  login?: string;

  @OneToMany(type => Conversation, conversation => conversation.sender)
  sentConversations?: Conversation[];

  @OneToMany(type => Conversation, conversation => conversation.reciever)
  recievedConversations?: Conversation[];
}
```

```
@Entity()
export class Conversation {
  @PrimaryColumn()
  senderId?: string;

  @PrimaryColumn()
  recieverId?: string;

  @CreateDateColumn({ primary: true })
  date?: Date;

  @OneToMany(type => Message, message => message.conversation, { cascade: true })
  messages: Message[];

  @ManyToOne(type => User, user => user.sentConversations)
  @JoinColumn({ name: 'senderId' })
  sender?: User;

  @ManyToOne(type => User, user => user.recievedConversations)
  @JoinColumn({ name: 'recieverId' })
  reciever?: User;
}
```

```
@Entity('message')
export class Message extends BaseEntity {
  @Column({ length: 10000 })
  content: string;

  @ManyToOne(type => Conversation, conversation => conversation.messages)
  conversation?: Conversation;
}
```

### LINKED ISSUE #6416: Combining ManyToOne, Cascade, & Composite Primary Key causes Unique Constraint issues
**Issue type:**

[ ] question
[x] bug report
[ ] feature request
[ ] documentation issue

**Database system/driver:**

[ ] `cordova`
[ ] `mongodb`
[ ] `mssql`
[ ] `mysql` / `mariadb`
[ ] `oracle`
[ ] `postgres`
[ ] `cockroachdb`
[x] `sqlite`
[ ] `sqljs`
[ ] `react-native`
[ ] `expo`

**TypeORM version:**

[x] `latest`
[ ] `@next`

---

I've found that a relationship between two entities where one is `ManyToOne`, cascading, and the relationship is part of a composite primary key causes various problems when persisting the entities.

Generally, what happens is that during persistence when the [subject recomputes](https://github.com/typeorm/typeorm/blob/3227c0b7b9f79868f755eb40e72287bdb60d71b6/src/persistence/Subject.ts#L290-L294), the `ManyToOne` primary key field gets set with the entity on the other side of the relationship.

However, when `SubjectDatabaseEntityLoader.findByPersistEntityLike` gets called it can query the `Entity` from the database (because `{ where: { otherId: OtherEntity() } }` works but `subject.metadata.compareEntities` can't understand the relationship.

This means that the first time we save a record and the related entities it works & inserts everything as expected.  However, the second time it does not and tries to insert the related entities again - even with their primary keys - causing a unique constraint issue.

In my case, because we're using transformers on the primary keys, I am struggling to get the relationship to work at all without the field.

I'll be creating a test and digging into a few possible solutions for this.

---

Minimal Example Entities which exhibit the behavior below.  To note, `PostTag` has a custom generation because `sqlite` does not support composite primary key `AUTOINCREMENT`.

```
import { EntitySchema } from '../../../../src/entity-schema/EntitySchema';

import PostTag from './PostTag';

export default class Post {
    postId: number;

    title: string;

    tags: PostTag[];
}

export const PostSchema = new EntitySchema<Post>({
    name: 'Post',
    target: Post,
    columns: {
        postId: {
            generated: true,
            type: Number,
            primary: true
        }
    },
    relations: {
        tags: {
            target: () => PostTag,
            type: "one-to-many",
            inverseSide: (tag: PostTag) => tag.post,
            cascade: true
        }
    }
});
```

```
import { EntitySchema } from '../../../../src/entity-schema/EntitySchema';

import Post from './Post';

let id = 0;

export default class PostTag {
    postTagId: number;

    postId: string;

    name: string;

    post: Post;

    constructor () {
        this.postTagId = id++;
    }
}

export const PostTagSchema = new EntitySchema<PostTag>({
    name: 'PostTag',
    target: PostTag,
    columns: {
        postId: {
            type: Number,
            primary: true
        },
        postTagId: {
            type: Number,
            primary: true
        }
    },
    relations: {
        post: {
            nullable: false,
            target: () => Post,
            inverseSide: (post: Post) => post.tags,
            type: 'many-to-one',
            joinColumn: { name: 'postId' }
        }
    }
});
```

The error received on the second save of the Entity when using SQLite is:

```
QueryFailedError: SQLITE_CONSTRAINT: UNIQUE constraint failed: post_tag.otherId, post_tag.postId, post_tag.postTagId
    at new QueryFailedError (typeorm/src/error/QueryFailedError.ts:9:9)
    at Statement.handler (typeorm/src/driver/sqlite/SqliteQueryRunner.ts:53:26) {
  errno: 19,
  code: 'SQLITE_CONSTRAINT',
  query: 'INSERT INTO "post_tag"("otherId", "postId", "postTagId") VALUES (?, ?, ?)',
  parameters: [ 3, 2, 1 ]
}
```

imnotjames:
handle setting values deeply in entities with relations in ColumnMetadata,
also check for a virtual relationship column in rawsqlresultstoentitytransformer

this allows us to handle a case where `many-to-one` with explicit composite PKs
columns were failing to persist a second time- instead of correctly updating the
field they would cause an insert to occur leading to a unique PK constraint error

Fixes #6416 
Fixes #4969
Fixes #4122

### DIFF
--- a/src/metadata/ColumnMetadata.ts
+++ b/src/metadata/ColumnMetadata.ts
@@ -662,7 +662,18 @@ export class ColumnMetadata {
             return extractEmbeddedColumnValue([...this.embeddedMetadata.embeddedMetadataTree], entity);
 
         } else {
-            entity[this.propertyName] = value;
+            // we write a deep object in this entity only if the column is virtual
+            // because if its not virtual it means the user defined a real column for this relation
+            // also we don't do it if column is inside a junction table
+            if (!this.entityMetadata.isJunction && this.isVirtual && this.referencedColumn && this.referencedColumn.propertyName !== this.propertyName) {
+                if (!(this.propertyName in entity)) {
+                    entity[this.propertyName] = {};
+                }
+
+                entity[this.propertyName][this.referencedColumn.propertyName] = value;
+            } else {
+                entity[this.propertyName] = value;
+            }
         }
     }
 
--- a/src/persistence/Subject.ts
+++ b/src/persistence/Subject.ts
@@ -289,7 +289,8 @@ export class Subject {
             if (this.parentSubject) {
                 this.metadata.primaryColumns.forEach(primaryColumn => {
                     if (primaryColumn.relationMetadata && primaryColumn.relationMetadata.inverseEntityMetadata === this.parentSubject!.metadata) {
-                        primaryColumn.setEntityValue(this.entityWithFulfilledIds!, this.parentSubject!.entity);
+                        const value = primaryColumn.referencedColumn!.getEntityValue(this.parentSubject!.entity!);
+                        primaryColumn.setEntityValue(this.entityWithFulfilledIds!, value);
                     }
                 });
             }
--- a/src/query-builder/transformer/RawSqlResultsToEntityTransformer.ts
+++ b/src/query-builder/transformer/RawSqlResultsToEntityTransformer.ts
@@ -237,7 +237,7 @@ export class RawSqlResultsToEntityTransformer {
                 const idMap = columns.reduce((idMap, column) => {
                     let value = result[column.databaseName];
                     if (relation.isOneToMany || relation.isOneToOneNotOwner) {
-                        if (column.referencedColumn) // if column is a relation
+                        if (column.isVirtual && column.referencedColumn && column.referencedColumn.propertyName !== column.propertyName) // if column is a relation
                             value = column.referencedColumn.createValueMap(value);
 
                         return OrmUtils.mergeDeep(idMap, column.createValueMap(value));
--- a/test/github-issues/6416/entity/Post.ts
+++ b/test/github-issues/6416/entity/Post.ts
@@ -1,0 +1,52 @@
+import { EntitySchema } from "../../../../src";
+
+import PostTag from "./PostTag";
+import PostAttachment from "./PostAttachment";
+
+let id = 0;
+
+export default class Post {
+    postId: number;
+
+    otherId: number;
+
+    tags: PostTag[];
+
+    attachments: PostAttachment[];
+
+    constructor() {
+        this.postId = id++;
+        this.otherId = id++;
+    }
+}
+
+export const PostSchema = new EntitySchema<Post>({
+    name: "Post",
+    target: Post,
+    columns: {
+        otherId: {
+            type: Number,
+            primary: true,
+            nullable: false
+        },
+        postId: {
+            type: Number,
+            primary: true,
+            nullable: false
+        }
+    },
+    relations: {
+        tags: {
+            target: () => PostTag,
+            type: "one-to-many",
+            inverseSide: "post",
+            cascade: true
+        },
+        attachments: {
+            target: () => PostAttachment,
+            type: "one-to-many",
+            inverseSide: "post",
+            cascade: true
+        }
+    }
+});
--- a/test/github-issues/6416/entity/PostAttachment.ts
+++ b/test/github-issues/6416/entity/PostAttachment.ts
@@ -1,0 +1,35 @@
+import { EntitySchema } from "../../../../src";
+
+import Post from "./Post";
+
+let id = 0;
+
+export default class PostAttachment {
+    attachmentId: number;
+
+    post: Post;
+
+    constructor () {
+        this.attachmentId = id++;
+    }
+}
+
+export const PostAttachmentSchema = new EntitySchema<PostAttachment>({
+    name: "PostAttachment",
+    target: PostAttachment,
+    columns: {
+        attachmentId: {
+            type: Number,
+            primary: true,
+            nullable: false
+        }
+    },
+    relations: {
+        post: {
+            primary: true,
+            nullable: false,
+            target: () => Post,
+            type: "many-to-one"
+        }
+    }
+});
--- a/test/github-issues/6416/entity/PostTag.ts
+++ b/test/github-issues/6416/entity/PostTag.ts
@@ -1,0 +1,51 @@
+import { EntitySchema } from "../../../../src";
+
+import Post from "./Post";
+
+let id = 0;
+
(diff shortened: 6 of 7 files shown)


imnotjames:
PS, through doing this I've found that there's a number of ways the surrounding code could be refactored to make something like this easier to update -

* The status of "virtual relation-created column" - eg, a column that has been created as part of a relation existing without
  an explicit column already being there is mostly implied [through a few particular field checks](https://github.com/typeorm/typeorm/blob/880f58e6eb9a668d339f8dd5f540a65bd9e348b2/src/metadata/ColumnMetadata.ts#L720).  This should be codified
  either as a `getter` or as a property that gets set when the `ColumnMetadata` is created
* Embedded column code is just peppered throughout everything.  It might be worthwhile to have a separate layer - eg, an `EmbeddedColumnMetadata extends ColumnMetdata` - to support this logic instead of the `if` blocks.. As it stands, it adds a lot of complexity to the classes.
* One thing I've found is that there's a number of footguns in the configuration of a schema.  It might be worthwile to clamp down on support of features with more explicit error messages being thrown instead of letting things get in a state that's broken.  Buuut... that's a bit more vague to implement.
* Not everything is the same between decorators and the entity schema.  You can create some things in decorators that you can't in EntitySchema and some of the namings aren't quite the same.  Not sure the best way to fix that, though, without breaking backwards compatibility.. There's also a number of defaults present in the decorators that aren't in `EntitySchema` - though in some cases it's because the decorators have more knowledge of where they're being associated..

jmc420:
I just raised an issue which I think is linked to this: https://github.com/typeorm/typeorm/issues/10450

In my case the composite primary key being updated includes the foreign key to the parent and you get a constraint error. When updating a child with a single primary key or a composite key that does not include the foreign key, updating works ok.

PR #6440, opened 2020-07-21
fix: prevent wrong returned entity in ReturningResultsEntityUpdator

imnotjames:
in the `ReturningResultsEntityUpdator` we have to check that the `entityId`
we pull from the entity is actually there.  if we don't we will create a
where clause that's undefined - effectively querying for every record
in the database and returning the wrong value


imnotjames:
I should probably add a test that validates this - I found it via cascades causing some other bug..

HopperGithub:
the version 0.2.26 also can hit the bug, when use Repository insert API to insert entities

example  `await this.repository.insert(insertList)`

when I  debug, mysql had work successfully, then program walk in this snippets and throw error

`TypeError: Cannot convert undefined or null to object`

image

Who can give some advice kindly

pleerock:
The question is: do we need to return error or just skip setting returning result? We need to know a real-case scenario of how this could happen and why

imnotjames:
At this moment it's a temporary defensive and preventative measure to throw an exception - preventing the incorrect return & invalid retrieval from the database.

There's underlying issues that cause this behavior that haven't been determined yet.

You can see this same technique applied in the update method - https://github.com/typeorm/typeorm/pull/6440/files#diff-39f14c63414acd9b002e8d238893fa25L59

marshalys:
when I insert item bulk, I get this error:

```
 Error: Cannot update entity because entity id is not set in the entity.
```

my code like this:

```

const items = ['xx', 'xxx'].map((value) => {
  const item = new Model();
  item.name = value;
  return item;
});

await getRepository(Model).insert(items);
```

how can I fix this?

Issue #6442, opened 2020-07-22
@JoinTable does not respect inverseJoinColumns referenced column width


michael-golden:
**Issue type:**

[ ] question
[x ] bug report
[ ] feature request
[ ] documentation issue

**Database system/driver:**

[ ] `cordova`
[ ] `mongodb`
[x ] `mssql`
[ ] `mysql` / `mariadb`
[ ] `oracle`
[ ] `postgres`
[ ] `cockroachdb`
[ ] `sqlite`
[ ] `sqljs`
[ ] `react-native`
[ ] `expo`

**TypeORM version:**

[ ] `latest`
[ ] `@next`
[ x] `0.2.25`

**Steps to reproduce or a small repository showing the problem:**

Specifying joinColumn in the `@JoinTable()` decorator does not pass the inverse join column referenced column width.

```ts
@Entity()
export class FooEntity {
    @PrimaryColumn({ type: "int", width: 10, unsigned: true, nullable: false })
    @Generated()
    public id: number;

    @ManyToMany(() => BarEntity)
    @JoinTable({
        name: "foo_bars",
        joinColumns: [
            {
                name: "foo_id",
            }
        ],
        inverseJoinColumns: [
            {
                name: "bar_id",
            }
        ]
    })
    public fooBars: BarEntity[];
}

@Entity()
export class BarEntity {
    @PrimaryColumn({ type: "int", width: 10, unsigned: true })
    @Generated()
    public id: number;

    @Column("varchar", { nullable: false, length: 50 })
    public code: string;
}
```

The [JunctionEntityMetadataBuilder](https://github.com/typeorm/typeorm/blob/master/src/metadata-builder/JunctionEntityMetadataBuilder.ts) [build()](https://github.com/typeorm/typeorm/blob/9abab82d51e3b76d717f2f473cbf3e4d96f51488/src/metadata-builder/JunctionEntityMetadataBuilder.ts#L31) implementation does not pass the 'width' as an arg.

See: https://github.com/typeorm/typeorm/blob/master/src/metadata-builder/JunctionEntityMetadataBuilder.ts#L110-L125

suggested change:
```
options: {
    length: !inverseReferencedColumn.length
                        && (this.connection.driver instanceof MysqlDriver || this.connection.driver instanceof AuroraDataApiDriver)
                        && (inverseReferencedColumn.generationStrategy === "uuid" || inverseReferencedColumn.type === "uuid")
                            ? "36"
[...]

PR #6444, opened 2020-07-22
#6442 fix: @JoinTable does not respect inverseJoinColumns referenced column width

michael-golden:
fix: @JoinTable does not respect inverseJoinColumns referenced column width

Closes: #6442


pleerock:
Thanks for contribution! Looks like you have a CI failed, can you please check that? There are no errors, so I guess you have missed `await` somewhere that's why CI can't finish tests.

michael-golden:
@pleerock - I've added awaits, but I'm not seeing any further reason for the timeout - any ideas?

michael-golden:
@pleerock - Tests are fixed!

pleerock:
Thank you!

PR #6452, opened 2020-07-23
chore: use eslint instead of tslint

imnotjames:
This updates the repository to use `eslint` instead of `tslint`, using a close approximation of the `tslint` config as an `eslint` config.

This is because `eslint` has been deprecated and the suggested upgrade route is to switch to `eslint` using the `tslint-to-eslint` conversion tool.  There were a few notices, but overall everything seems to work as expected. No changes to the linting errors (or lack thereof) in the repository. Notices are as follows:

```
5 ESLint rules behave differently from their TSLint counterparts:
  * @typescript-eslint/no-unused-vars:
    - Please read the following article as the rule behaviour may change on the short term: https://github.com/typescript-eslint/typescript-eslint/issues/1856
  * brace-style:
    - ESLint's brace-style will check all tokens.
  * eqeqeq:
    - Option "smart" allows for comparing two literal values, evaluating the value of typeof and null comparisons.
  * camelcase:
    - Leading and trailing underscores (_) in variable names will now be ignored.
  * no-underscore-dangle:
    - Leading or trailing underscores (_) on identifiers will now be forbidden.

1 rule is not known by tslint-to-eslint-config to have an ESLint equivalent:
  * tslint-to-eslint-config does not know the ESLint equivalent for TSLint's "whitespace".
```

Also, a few configuration items were dropped because the repository does not pass:

* `@typescript-eslint/member-delimiter-style`
* `no-underscore-dangle`
* `camel-case`

I suggest enabling these after this is merged in with the related linting fixes if there's interest to do so.

Also, spacing related items were turned off because A: Too many actual issues & B: Apparently lots of typescript incompatibility / wonk.

* `@typescript-eslint/indent`
* `spaced-comment`

I suggest fixing some of the actual issues (some files are indented with 3 spaces, what?) & probably using something more opinionated like `prettier` via the `eslint-plugin-prettier`.


pleerock:
I hate prettier, but I do use it for all my projects last few years. So, I'm good with switching to it. **However** first we need to merge this branch into **next** branch, then apply prettier on both. We need to avoid conflicts as much as possible.

imnotjames:
I'm not a huge fan of prettier either, but I am a fan of uniformity.  Per the Typescript ESLint folks - they suggested prettier. :\ ( https://github.com/typescript-eslint/typescript-eslint/issues/1824 )

Agreed on making sure we avoid merge conflicts.

For reference, branch with prettier support - https://github.com/imnotjames/typeorm/tree/chore/prettier

PR #6460, opened 2020-07-24
chore: Use CircleCI for all CI builds

imnotjames:
Drop travis-ci from the builds by using the matrix functionality in circleci to run multiple node versions.

Refactor the way builds work to allow splitting and joining per-database builds together more easily.

Add support back in for MSSQL

Drops the (unused) oracle job definition.

You can see this CI configuration having run [at the CircleCI for imnotjames/typeorm](https://app.circleci.com/pipelines/github/imnotjames/typeorm/97/workflows/208af43a-3e57-4a71-a467-f783b03eaee9) & the [Code Coverage reported as expected.](https://codecov.io/gh/imnotjames/typeorm/commit/1f55a7789dd1c962957483950d20b2486d65f6cf)

~~**However, as part of this - the code coverage will need to be entered into CircleCI just as it's in Travis-CI - either using the same CodeCov key or a different one - with the project environment variable `CODECOV_TOKEN`.**~~ Token is in place and code coverage is publishing.


imnotjames:
Also - CockroachDB remains very slow.

I've actually reached out to the folks at CockroachDB on this - if there's any options & I've been experimenting with some tweaks to no avail.  I've gotten it down from 25m to 18m but..

pleerock:
Thank you, looks great! Few questions:

* Do we loose something by dropping travis ci? We had many discussions and decisions in past, as well as problems with both circleci and travis. 
* Can we add other drivers we have (for which we don't run tests)? Like mssql and SAP Hana
* Can we name Items from "test-common-node12" to something like "test-oracle" and "test-mysql", etc. ? Its not really clear now which one is oracle for example

imnotjames:
> Do we loose something by dropping travis ci? We had many discussions and decisions in past, as well as problems with both circleci and travis.

TravisCI was doing code coverage and builds for NodeJS 10,12, & 13.  This change adds that to the CircleCI builds.

~However, keep in mind - until the `CODECOV_TOKEN` is added to CircleCI it cannot publish to `codecov.io`~ It seems like the CodeCov token is in place in the CircleCI environment.

> Can we add other drivers we have (for which we don't run tests)? Like mssql and SAP Hana

Not sure I follow - could you elaborate?  Add how?

If you mean include them in the builds so we do run tests against them, yes, that could be done for MSSQL.  It was removed because it was failing.. somehow.  SAP Hana requires a private docker image which I don't know who has access to and was never enabled. Oracle as well, it seems. ):

> Can we name Items from "test-common-node12" to something like "test-oracle" and "test-mysql", etc. ? Its not really clear now which one is oracle for example

`test-common-node{version}` maps up to the current `common`.  We _could_ split these out further if there's interest - but too many jobs on either Travis or CircleCI will cause some blocking. (there's only so many concurrent jobs that can run)

I tried to keep the jobs roughly the same in this PR so there's only `common` and `cockroachdb`.

Also: This drops the oracle job as it didn't really ever run, it seems.

If we want to split `common` out it might be worthwhile to investigate how we can only run the servers we want to run.
This probably would be jobs specific to each database instead of the way I've implemented it here, abstracting the runnable bits to commands.

imnotjames:
I've updated the job names to make them clearer as to what they are.

pleerock:
Thank you a lot

PR #6463, opened 2020-07-24
test: remove hardcoded test from github issue tst 2096

imnotjames:
the test had assumed the username, password, host, and port of
the mysql server

this update changes that so the test instead uses the config properly
while still checking that the test works as expected


pleerock:
Thank you!

PR #6469, opened 2020-07-26
feat: implement generated columns for postgres 12 driver

TheNoim:
The implementation has the potential to make full-text search much faster when using Postgres. You can simply pre-generate all tsvector's.

Example:

```typescript
export class Post {

    @PrimaryColumn()
    id: number;

    @Column({ unique: true })
    version: number;

    @Column({ default: "My post" })
    name: string;

    @Column()
    text: string;

    @Column()
    tag: string;

    @Column({ generatedType: "STORED", asExpression: "to_tsvector('english', text)" })
    tsv: string[];
}
```

Note:

My implementation is a bit different from the MySQL implementation. For me, it was important that a generated column only gets dropped and added again if the as expression changes. Maybe I should implement the same mechanic for the MySQL driver.


TheNoim:
Ok, this is a problem. The tests are running on Postgres 9. However, this is a Postgres 12 feature.

TheNoim:
Resolved the merge conflicts and adjusted the ci to also test Postgres 12.

TheNoim:
Rebased and resolved conflicts.

src/driver/postgres/PostgresQueryRunner.ts
```diff
@@ -344,6 +344,22 @@ export class PostgresQueryRunner extends BaseQueryRunner implements QueryRunner
...
+            // if table have column with generated type, we must add the expression to the meta table
+            await Promise.all(table.columns
+                .filter(column => column.generatedType === "STORED" && column.asExpression)
+                .map(async column => {
+                    const tableName = await this.getTableNameWithSchema(table.name);
+                    const deleteQuery = new Query(`DELETE FROM typeorm_generation_meta WHERE table_name = $1 AND column_name = $2`, [tableName, column.name]);
+                    upQueries.push(deleteQuery);
+                    upQueries.push(new Query(`INSERT INTO typeorm_generation_meta(table_name, column_name, generation_expression) VALUES ($1, $2, $3)`, [tableName, column.name, column.asExpression]));
+                    downQueries.push(deleteQuery);
+                    return Promise.resolve();
```
imnotjames, line 357: Do you actually need to return `Promise.resolve()` here?

src/driver/postgres/PostgresQueryRunner.ts
```diff
@@ -344,6 +344,22 @@ export class PostgresQueryRunner extends BaseQueryRunner implements QueryRunner
                 return Promise.resolve();
             }));
 
+        if (await this.driver.isGeneratedColumnsSupported(this.mode)) {
+            // if table have column with generated type, we must add the expression to the meta table
+            await Promise.all(table.columns
+                .filter(column => column.generatedType === "STORED" && column.asExpression)
+                .map(async column => {
+                    const tableName = await this.getTableNameWithSchema(table.name);
```
imnotjames, line 352: Can't this be done outside the `map` call - then all of this can be sync - no need for all of the `Promise.all` and async.

src/driver/postgres/PostgresQueryRunner.ts
```diff
@@ -1592,6 +1645,19 @@ export class PostgresQueryRunner extends BaseQueryRunner implements QueryRunner
                         }
                     }
 
+                    if (dbColumn["is_generated"] === "ALWAYS" && dbColumn["generation_expression"] && supportsGeneratedColumns) {
+                        // In postgres there is no VIRTUAL generated column type
+                        tableColumn.generatedType = "STORED";
```
imnotjames, line 1650: Is there a case where a `VIRTUAL` might be attempted here?

src/schema-builder/RdbmsSchemaBuilder.ts
```diff
@@ -786,6 +791,36 @@ export class RdbmsSchemaBuilder implements SchemaBuilder {
         ), true);
     }
 
+    /**
+     * Creates typeorm service table for storing user defined generated columns as expressions.
+     */
+    protected async createTypeormGeneratedMetadataTable() {
```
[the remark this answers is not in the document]
TheNoim: Sadly no. Postgres, of course, saves the generated expression in a metatable, but it gets formatted differently and there is no way of checking if the expressions are the same. You would need to build your own sql parser to check if there are the same and I think this is a bit out of scope for this.

src/schema-builder/RdbmsSchemaBuilder.ts
```diff
@@ -786,6 +791,36 @@ export class RdbmsSchemaBuilder implements SchemaBuilder {
...
+        const options = <SqlServerConnectionOptions|PostgresConnectionOptions>this.connection.driver.options;
```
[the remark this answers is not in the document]
TheNoim: Well, I don't know anymore. But it works without it. I will remove it.

.circleci/config.yml
```diff
@@ -213,3 +209,11 @@ workflows:
             - build
           databases: "oracle"
           node-version: "12"
+
+      - test:
```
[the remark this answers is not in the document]
TheNoim: I think this is not my decision if we should switch to Postgres 12 only as test target. This is why I just introduced Postgres 12 and not just changed it to 12. If you think we should change the target I will do it.

.circleci/config.yml
```diff
@@ -125,10 +124,9 @@ jobs:
               npx js-yaml ./docker-compose.yml \
                 | jq -r '.services | keys | map(select(. | IN($ARGS.positional[]))) | join(" ")' --args << parameters.databases >>
             )
-
             docker-compose --project-name typeorm --no-ansi up --detach $SERVICES
       - install-packages:
-         cache-key: node<< parameters.node-version >>
+          cache-key: node<< parameters.node-version >>
```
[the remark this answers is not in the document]
TheNoim: I will. When I updated the pr I copied the config.yml from the repo.

src/driver/postgres/PostgresDriver.ts
```diff
@@ -910,6 +912,18 @@ export class PostgresDriver implements Driver {
         return true;
     }
 
+    /**
+     * Returns true if postgres supports generated columns
+     */
+    async isGeneratedColumnsSupported(mode: ReplicationMode): Promise<boolean> {
+        const runner = this.createQueryRunner(mode);
+        const results = await runner.query("SHOW server_version;", []);
```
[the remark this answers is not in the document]
TheNoim: You mean caching it? Haven't thought about this. This is a good idea. Any suggestions where to put it? Directly in the connect function? Or in afterConnect?

src/driver/postgres/PostgresDriver.ts
```diff
@@ -910,6 +912,18 @@ export class PostgresDriver implements Driver {
...
+        await runner.release();
+        const versionString = results[0]["server_version"] as string;
+        const version = parseFloat(versionString);
```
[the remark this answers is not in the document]
TheNoim: Didn't know we have something like this. Will update this.

src/driver/postgres/PostgresQueryRunner.ts
TheNoim: Not really. I will remove it.

src/driver/postgres/PostgresQueryRunner.ts
TheNoim: I mean I could transform it to a for loop. But does this make it better? I mean we need to do at least one async call. Is it really important where?

src/driver/postgres/PostgresQueryRunner.ts
TheNoim: This just makes sure it is marked as a generated column. We don't have a boolean switch `is_generated_column` or something similar. This is similar to it.

.circleci/config.yml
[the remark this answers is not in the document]
TheNoim: Well, I think this whitespace addition is correct. But I removed the other ones.

src/driver/postgres/PostgresQueryRunner.ts
```diff
@@ -361,6 +361,17 @@ export class PostgresQueryRunner extends BaseQueryRunner implements QueryRunner
                 return Promise.resolve();
             }));
 
+        // if table have column with generated type, we must add the expression to the meta table
+        await Promise.all(table.columns
+            .filter(column => column.generatedType === "STORED" && column.asExpression)
+            .map(async column => {
+                const tableName = await this.getTableNameWithSchema(table.name);
+                const deleteQuery = new Query(`DELETE FROM typeorm_generation_meta WHERE table_name = $1 AND column_name = $2`, [tableName, column.name]);
```
pleerock, line 369: I'm not sure if you aware or not, but we already have a `typeorm_metadata` table where we store metadata... we should probably use it instead of generating of a new table....

src/driver/postgres/PostgresDriver.ts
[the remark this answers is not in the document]
pleerock: yes, I think we can do it in `afterConnect`

src/driver/postgres/PostgresQueryRunner.ts
pleerock: Otherwise PR looks good and can be merged

TheNoim: > I'm not sure if you aware or not, but we already have a `typeorm_metadata` table where we store metadata... we should probably use it instead of generating of a new table....

I was aware. However, I was not able to find a usage of this table somewhere and therefore didn't know how to use it correctly. I didn't want to cause any conflicts. And it is hard to identify whether you could cause any conflicts if you don't know how it is already used.

TheNoim: @pleerock I would appreciate an response :) I need to know If I need to refactor the code to use the other meta table. However, I still don't know if this meta table fits the usage.

TheNoim: @pleerock Just another small reminder ;) Really want this support in typeorm.

TheNoim:
*sigh*

TheNoim:
It would be cool if this pr gets some love too, after so many prs got merged this week.

TheNoim:
It seems like, tests are failing because of #8017. I hope this doesn't prevent this pr to finally happen.

src/driver/postgres/PostgresQueryRunner.ts
```diff
@@ -435,6 +435,17 @@ export class PostgresQueryRunner extends BaseQueryRunner implements QueryRunner
             }
         }
 
+        // if table have column with generated type, we must add the expression to the meta table
+        await Promise.all(table.columns
```
imnotjames, line 439: Isn't it possible for the queries here to be inconsistently generated because of the `Promise.all`?  The order that they're generated in can change drastically because of it.

TheNoim: Could you please elaborate more? As far as I know the order with Promise.all never changes. And even if it does, I think it isn't really an issue, because this line only affects the meta table. There it doesn't matter in which order they get inserted/removed.

TheNoim: Ok, I think I known what you mean. This is an alternative approach:

```typescript
await Promise.all(table.columns
    .filter(column => column.generatedType === "STORED" && column.asExpression)
    .map(async column => {
        const tableName = await this.getTableNameWithSchema(table.name);
        const deleteQuery = new Query(`DELETE FROM typeorm_generation_meta WHERE table_name = $1 AND column_name = $2`, [tableName, column.name]);
        const insertQuery = new Query(`INSERT INTO typeorm_generation_meta(table_name, column_name, generation_expression) VALUES ($1, $2, $3)`, [tableName, column.name, column.asExpression]);
        return [ [deleteQuery, insertQuery], [deleteQuery] ];
    })
)
    .forEach(([ providedUpQueries, providedDownQueries ]) => {
        upQueries.push(...providedUpQueries);
        downQueries.push(...providedDownQueries);
    });
```

pleerock:
Sorry your PR is on the last page of the list, that's why I never reach it, since I always start from the first page 😆 Let's figure out what to do - merge or close it, I'm going to review it.

TheNoim:
> Sorry your PR is on the last page of the list, that's why I never reach it, since I always start from the first page 😆 Let's figure out what to do - merge or close it, I'm going to review it.

This is still open: https://github.com/typeorm/typeorm/pull/6469/files#r584112670

src/driver/postgres/PostgresDriver.ts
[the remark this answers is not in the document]
pleerock: @TheNoim any updates here?

src/driver/postgres/PostgresQueryRunner.ts
pleerock: I think what @imnotjames meant is to replace `Promise.all(table.columns` to `for (let column of table.columns)` in order to have a more predictable order of generated queries in migration.

src/driver/postgres/PostgresQueryRunner.ts
pleerock: My response is the same - we don't want to have separate table for each feature we have inside schema sync. We should use `typeorm_metadata` table to store schema metadata information for any use case we have. You can see how its used in the same `RdbmsSchemaBuilder` file.

test/functional/query-runner/add-column.ts
```diff
@@ -72,6 +90,35 @@ describe("query runner > add column", () => {
         column2.length.should.be.equal("100");
         column2!.default!.should.be.equal("'this is description'");
 
+        if (connection.driver instanceof MysqlDriver || connection.driver instanceof PostgresDriver) {
+            const isMySQL = connection.driver instanceof MysqlDriver && connection.options.type === "mysql";
+            let postgresSupported = false;
+
+            if (connection.driver instanceof PostgresDriver) {
```
pleerock, line 97: why `connection.driver.isGeneratedColumnsSupported` is not used here?

pleerock (changes requested):
All comments from previous review should be resolved before this can be merged.

src/driver/postgres/PostgresQueryRunner.ts
TheNoim: Ok, I can do this

src/driver/postgres/PostgresQueryRunner.ts
TheNoim: But it only gets generated in RdbmsSchemaBuilder. And there is no clear documentation ? Of course `database`, `schema` and `table` are self explaining. But what is `type` and `name` used for? I don't want to override anything which may be in use. Otherwise I would do it like this:

- `type`: `generated`
- `name`: Column name
- `value`: Expression

Is this ok?

pleerock: > And there is no clear documentation ?

This stuff is internal, that's why there is no clear documentation. Also, this code wasn't written by me, so I also don't know all the details. Source code is your best friend 😆

pleerock:
Tests are failing btw

TheNoim:
> Tests are failing btw

Yes, I see it. Will try to solve it as soon as possible.

dunklesToast:
Since tests are passing now, can this land soon?

AlexMesser:
* added a new `MetadataTableType` enum to make metadata table types more stricter and predictable
```ts
export enum MetadataTableType {
    VIEW = "VIEW",
    MATERIALIZED_VIEW = "MATERIALIZED_VIEW",
    GENERATED_COLUMN = "GENERATED_COLUMN"
}
```
* implemented `insertTypeormMetadataSql()` and `deleteTypeormMetadataSql()` functions in `BaseQueryRunner` to unify metadata insertion and deletetion across all drivers and metadata types
* some code refactoring

AlexMesser:
thank you for contribution!

TheNoim:
I am so happy this finally landed on master <3

Migushthe2nd:
@TheNoim Schema synchronization seems to fail.

```log
query failed: SELECT * FROM "typeorm_metadata"  WHERE "table" = 'theme_previews' AND "name" = 'image720Hash' AND "
schema" = 'typeorm' AND "database" = 'themezer_dev' AND "type" = 'GENERATED_COLUMN'
error: error: relation "typeorm_metadata" does not exist
    at Parser.parseErrorMessage (node_modules\pg-protocol\src\parser.ts:369:69)
    at Parser.handlePacket (node_modules\pg-protocol\src\parser.ts:188:21)
    at Parser.parse (node_modules\pg-protocol\src\parser.ts:103:30)
    at Socket.<anonymous> (node_modules\pg-protocol\src\index.ts:7:48)
    at Socket.emit (node:events:394:28)
    at addChunk (node:internal/streams/readable:312:12)
    at readableAddChunk (node:internal/streams/readable:287:9)
    at Socket.Readable.push (node:internal/streams/readable:226:10)
    at TCP.onStreamRead (node:internal/stream_base_commons:190:23) {
  length: 115,
  severity: 'ERROR',
  code: '42P01',
  detail: undefined,
  hint: undefined,
  position: '15',
  internalPosition: undefined,
  internalQuery: undefined,
  where: undefined,
  schema: undefined,
  table: undefined,
  column: undefined,
  dataType: undefined,
  constraint: undefined,
  file: 'parse_relation.c',
  line: '1381',
  routine: 'parserOpenTable'
}
query: ROLLBACK
query failed: SELECT * FROM "typeorm_metadata"  WHERE "table" = 'theme_previews' AND "name" = 'image360Hash' AND "
schema" = 'typeorm' AND "database" = 'themezer_dev' AND "type" = 'GENERATED_COLUMN'
error: error: current transaction is aborted, commands ignored until end of transaction block
...
```

It is fine the first sync, and the columns seem to be generated correctly in the database, however on second sync all generated columns fail. My decorator:
`@PrimaryColumn({type: "varchar", generatedType: "STORED", asExpression: "to_hex(counter)"})`

TheNoim:
> @TheNoim Schema synchronization seems to fail.
> 
> ```
> query failed: SELECT * FROM "typeorm_metadata"  WHERE "table" = 'theme_previews' AND "name" = 'image720Hash' AND "
> schema" = 'typeorm' AND "database" = 'themezer_dev' AND "type" = 'GENERATED_COLUMN'
> error: error: relation "typeorm_metadata" does not exist
>     at Parser.parseErrorMessage (node_modules\pg-protocol\src\parser.ts:369:69)
>     at Parser.handlePacket (node_modules\pg-protocol\src\parser.ts:188:21)
>     at Parser.parse (node_modules\pg-protocol\src\parser.ts:103:30)
>     at Socket.<anonymous> (node_modules\pg-protocol\src\index.ts:7:48)
>     at Socket.emit (node:events:394:28)
>     at addChunk (node:internal/streams/readable:312:12)
>     at readableAddChunk (node:internal/streams/readable:287:9)
>     at Socket.Readable.push (node:internal/streams/readable:226:10)
>     at TCP.onStreamRead (node:internal/stream_base_commons:190:23) {
>   length: 115,
>   severity: 'ERROR',
>   code: '42P01',
>   detail: undefined,
>   hint: undefined,
>   position: '15',
>   internalPosition: undefined,
>   internalQuery: undefined,
>   where: undefined,
>   schema: undefined,
>   table: undefined,
>   column: undefined,
>   dataType: undefined,
>   constraint: undefined,
>   file: 'parse_relation.c',
>   line: '1381',
>   routine: 'parserOpenTable'
> }
> query: ROLLBACK
> query failed: SELECT * FROM "typeorm_metadata"  WHERE "table" = 'theme_previews' AND "name" = 'image360Hash' AND "
> schema" = 'typeorm' AND "database" = 'themezer_dev' AND "type" = 'GENERATED_COLUMN'
> error: error: current transaction is aborted, commands ignored until end of transaction block
> ...
> ```
> 
> On initial sync the columns seem to be generated correctly in the database and there is no issue. On second sync all generated columns fail. My decorator: `@PrimaryColumn({type: "varchar", generatedType: "STORED", asExpression: "sha256(\"image720File\")"})`
> 
> Can you replicate this?

I was hoping #4956 solves the issue. But it seems like it doesn't. Maybe I need to look into it a bit further.

PR #6553, opened 2020-08-12
fix: Fix Mongodb delete by ObjectId. Closes #6552

### LINKED ISSUE #6552: MongoRepository delete by ObjectId deletes the wrong entity
**Issue type:**

[ ] question
[x] bug report
[ ] feature request
[ ] documentation issue

**Database system/driver:**

[ ] `cordova`
[x] `mongodb`
[ ] `mssql`
[ ] `mysql` / `mariadb`
[ ] `oracle`
[ ] `postgres`
[ ] `cockroachdb`
[ ] `sqlite`
[ ] `sqljs`
[ ] `react-native`
[ ] `expo`

**TypeORM version:**

[x] `latest`
[ ] `@next`

```ts
@Entity()
export class Post {

    @ObjectIdColumn()
    _id: ObjectID;

    @Column()
    title: string;

}

// try to delete a post by passing directly it's ObjectId (as the signature suggests should be possible)
// delete(criteria: string|string[]|number|number[]|Date|Date[]|ObjectID|ObjectID[]|FindConditions<Entity>): Promise<DeleteResult>
const post: Post = getSomePost();
repository.delete(post._id);
```
If you look at the actual query executed (in MongoQueryRunner) you will see `deleteOne({})`.
Notice the empty query! This will have a side effect of removing the first Post from the collection, whatever that is ...

If the Post entity has the id field renamed to `postId` the query will be `deleteOne({ _id: Buffer})`, which is still wrong, because the type should be ObjectId not Buffer. In this case the sideffect is less severe, it just doesn't delete anything

The workaround is to use `repository.delete({_id: post._id})` which works correctly.

The problem is in [MongoEntityManager.convertMixedCriteria](https://github.com/typeorm/typeorm/blob/0.2.25/src/entity-manager/MongoEntityManager.ts#L633) which tests for instanceof ObjectId in the wrong order:

```ts
protected convertMixedCriteria(metadata: EntityMetadata, idMap: any): ObjectLiteral {
        if (idMap instanceof Object) {
            return metadata.columns.reduce((query, column) => {
                const columnValue = column.getEntityValue(idMap);
                if (columnValue !== undefined)
                    query[column.databasePath] = columnValue;
                return query;
            }, {} as any);
        }

        // means idMap is just object id
        const objectIdInstance = PlatformTools.load("mongodb").ObjectID;
        return {
            "_id": (idMap instanceof objectIdInstance) ? idMap : new objectIdInstance(idMap)
        };
    }
```
If `idMap` is an ObjectId it will obviously also be an instance of Object and it will enter wrongly in the first branch, never reaching the code at the bottom.

danmana:


imnotjames:
As I understand it - without this change only the second of the two test cases pass?

imnotjames:
Can you add the case of the previous required behavior?  You'd mentioned it in the #6552 

> The workaround is to use `repository.delete({_id: post._id})` which works correctly.

If you add that I think this is more or less good to go.

danmana:
@imnotjames I added the regression test as suggested.

You are correct, before this fix:
* test 1: failed (deleted wrong entity)
* test 2: failed (didn't delete anything)
* test 3: passed

imnotjames:
Thanks for the contribution!

PR #6554, opened 2020-08-12
test: Fix typo in github issue 6442 test name

danmana:


pleerock:
@imnotjames do you have any ideas why this small change could fail?

imnotjames:
https://github.com/npm/cli/issues/558

In particular

https://github.com/npm/cli/issues/558#issuecomment-580018468

> npm v7 won't have this problem. (I'm working in the code that'll do this now.) I haven't checked into how involved it'd be to fix this issue for npm v6, but there's a good chance it'll just end up being "upgrade to v7 for the fix". In the meantime, I'd recommend using npm install instead of npm ci if this is affecting you.

We also probably want to use `lock-verify` to ensure the lockfile is in sync with the package list.

I'll take a look at a PR later today to try to resolve it.

imnotjames:
Awesome.  Looks like we're all good to go!

imnotjames:
Oh - one little thing for the future - this would be a `test` change rather than a `chore` change 

https://github.com/typeorm/typeorm/blob/master/CONTRIBUTING.md#commit

It's a bit confusing to find the right one sometimes.  I'll update the title for ya!

imnotjames:
Thanks for the contribution :tada:

PR #6567, opened 2020-08-14
Added links to database GUIs

rathboma:
Hey folks.

We use TypeORM to build Beekeeper Studio (which I list), and using Beekeeper Studio was helpful for viewing TypeORM data, so figured I'd add a section.

We're 100% foss, so no commercial incentive, so I've also added DBeaver and TablePlus which I think are the two other biggies.


pleerock:
First, it's not directly related to TypeORM, so I don't see a point of adding such link.
Second, it's an ad, no matter if related product is non-commercial or not.
Sorry, I can't accept it.

rathboma:
Ok, no worries, thanks for reviewing!

PR #6588, opened 2020-08-19
feat: backport SQLite Busy handler & WAL mode enable

imnotjames:
This is a backport of the SQLite logic to enable WAL & Busy retries


imnotjames:
@pleerock I've moved where the `PRAGMA` gets set for WAL to effectively once per connection - where other `PRAGMA` get called.

Was there an important reason for the `PRAGMA` call to have been in the query runner?

pleerock:
No, I don't think there was. Also in sqlite there is only one instance of query runner always.

pleerock:
Thank you! I'm sure this can close many opened sqlite issues we have.

Issue #5067, opened 2019-11-11
ORA-00972: identifier is too long on joins with FindOptions


derSoerrn95:
**Issue type:**

[ ] question
[x] bug report
[ ] feature request
[ ] documentation issue

**Database system/driver:**

[ ] `cordova`
[ ] `mongodb`
[ ] `mssql`
[ ] `mysql` / `mariadb`
[x] `oracle`
[ ] `postgres`
[ ] `cockroachdb`
[ ] `sqlite`
[ ] `sqljs`
[ ] `react-native`
[ ] `expo`

**TypeORM version:**

[x] `latest`
[ ] `@next`
[ ] `0.x.x` (or put your version here)

**Steps to reproduce or a small repository showing the problem:**

It's a copy of #693, cause the issue was closed, but the error still exists - at least with oracle.

When I'm using one of the `getRepository(Entity).find({})` functions with realations array set in the `FindOptions`, I'm getting the `too long identifier` error. I think the main reason it doesn't work is because of Oracle, but I cannot be sure about it and I'm not able to change my db.. 

So it would be really cool, if anyone could make a PR, that allows the user to add an alias for `RelationOptions` or (the easier way) let the aliases only contain out of 1-2 letters.

cesc1802:
i have same issue when table name and column name too long, alias name sometime wrong. for example i have 2 columns name CREATE_DATE and CREATE_BY into table BLABLABLABLABLABLA. alias name is BLABLABLABLABLABLA.CREATE_ and BLABLABLABLABLABLA.CREATE_ for 2 columns

PR #6591, opened 2020-08-19
fix: resolve issues ora-00972:identifier is too long

MuratGundes:
Use index instead of parameter name for make it shorter alias name.

Closes: #5067

### DIFF
--- a/src/driver/oracle/OracleDriver.ts
+++ b/src/driver/oracle/OracleDriver.ts
@@ -627,7 +627,7 @@ export class OracleDriver implements Driver {
      * Creates an escaped parameter.
      */
     createParameter(parameterName: string, index: number): string {
-        return ":" + parameterName;
+        return ":" + (index + 1);
     }
 
     /**
--- a/test/github-issues/5067/issue-5067.ts
+++ b/test/github-issues/5067/issue-5067.ts
@@ -2,6 +2,7 @@ import "reflect-metadata";
 import {expect} from "chai";
 import {Connection} from "../../../src";
 import {closeTestingConnections, createTestingConnections} from "../../utils/test-utils";
+import {OracleDriver} from "../../../src/driver/oracle/OracleDriver";
 
 describe.only("github issues > #5067 ORA-00972: identifier is too long", () => {
     let connections: Connection[];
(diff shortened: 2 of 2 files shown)


pleerock:
@imnotjames right now we have no idea if we can merge this small, but suspicious change because we have no oracle tests on CI running 😭

imnotjames:
We can get them running - they weren't before because they were disabled & I think the Oracle image couldn't be pulled 'cause it's not free.

Happy to take a look at adding an oracle build to the mix - I was poking at it the other day but gave up because it just hasn't been running before.

pleerock:
@imnotjames thanks a lot

MuratGundes:
Thanks a lot guys! We will be appreciated.

Rainson12:
merging this will be much appreciated it seems to be working for us and it seems like there is not other solution other than this to fix the issue with column& table names be longer than 30 characters. Its also not possible to shorten the column names as its a legacy system.

imnotjames:
Just to quickly update this thread on the status of OracleDB CI ---

I've refactored the CI to use docker-compise to start up services.  This
simplifies bringing in other services because they just need to be defined
in docker compose like the others & can even be "private" corporate images
like the OracleDB ones.

https://github.com/imnotjames/typeorm/blob/build/remote-docker/

I'll be trying to get this in as a PR today.

As an extended part of this I've added the client libraries (why would it
need separate libraries like this? Licensing I assume..? Ugh..) for
oracleDB as well as a docker compose service to start up oracle for us to
run tests against it.

However, currently running into an issue with oracle during the first test
that gets run against it - I've got a feeling I've set it up to connect to
the wrong SID? Unclear.

QueryFailedError: ORA-24005: Inappropriate utilities used to perform DDL on
AQ table SYSTEM.DEF$_AQCALL

imnotjames:
Ok - issue with tests is that I had tried to spin up Oracle 11 - which typeorm doesn't support.

imnotjames:
Oracle tests seem to be working and they catch another spot where this happens -

https://github.com/typeorm/typeorm/blob/master/src/query-builder/SelectQueryBuilder.ts#L1864

The column alias can end up being `30` characters long because of the `ids_` prefix.  I'd fixed it by replacing it with:

https://github.com/imnotjames/typeorm/blob/619b2bd42e6a4abfe116e6cad3a35a0b4a72eb85/src/query-builder/SelectQueryBuilder.ts#L1865-L1871

imnotjames:
Should this instead be done similar to aliases?  if > some predetermined driver specific length, `hash` it?

Like https://github.com/typeorm/typeorm/blob/5516aca07db4137df05ad2c789e10b943a939ae6/src/driver/DriverUtils.ts#L52

imnotjames:
Also @MuratGundes please rebase against `master` - the tests for Oracle should run now :)

imnotjames:
Hey @MuratGundes - I'm gonna open another PR with these changes - the commits will have you as the author, but will be rebased against `master` so the oracle tests will run!

pleerock:
thank you @imnotjames it would be great if we can merge this one as fast as possible

PR #6602, opened 2020-08-22
build: use docker-compose for database services

imnotjames:
This updates the CircleCI configuration to utilize `docker-compose` for running the dependent database services we run tests against.

This allows us to more easily deploy more varied and complex systems - such as databases that could normally run us out of memory (eg, CockroachDB) or databases that require authenticated login to fetch the images (eg, OracleDB & SAP Hana)

It also grants us access to a larger machine on the other end of the `docker-remote` - a system with 8GB of RAM instead of the 4GB the normal runner gets.

While the tests take a bit longer to run, there may be mechanisms we could employ to improve that down the line.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -93,42 +93,72 @@ jobs:
     working_directory: ~/typeorm
     docker:
       - image: circleci/node:<< parameters.node-version >>
-      - image: mysql:5.7.24
-        environment:
-          MYSQL_ROOT_PASSWORD: "admin"
-          MYSQL_DATABASE: "test"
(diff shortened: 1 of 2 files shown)


imnotjames:
This increases build times by about 1.5x on average.  This is not because of the images being cached or not.

It does seem to improve the stability of some of the tests & also makes it so the CI versions match up with dev versions with docker compose.

I'm not sure the cause, yet.  It could be that the remote docker disks used for volumes are slower so we would want to use in-memory storage for them.  That's a more in depth change, though.

Issue #3152, opened 2018-11-23
Webpack support is limited.


itonics-tbeauvais:
**Issue type:**

[ ] question
[x] bug report
[ ] feature request
[ ] documentation issue

**Database system/driver:**

[ ] `cordova`
[ ] `mongodb`
[ ] `mssql`
[ ] `mysql` / `mariadb`
[ ] `oracle`
[x] `postgres`
[ ] `sqlite`
[ ] `sqljs`
[ ] `react-native`
[ ] `expo`

**TypeORM version:**

[ ] `latest`
[ ] `@next`
[x] `0.2.9` (or put your version here)

**Steps to reproduce or a small repository showing the problem:**

All I did here was us the `typeorm init` to create create a new project that works well. Adding a simple webpack is where problems arise.

The problem seems to be with webpack. I have seen a few of these as I have been sifting through your closed and open issues.

Using a pretty basic `webpack.config.js`
```
const path = require('path');

module.exports = {
    entry: './src/index.ts',
    mode: 'development',
    module: {
        rules: [
            {
                test: /\.tsx?$/,
                use: 'ts-loader',
                exclude: /node_modules/
            }
        ]
    },
    resolve: {
        extensions: ['.tsx', '.ts', '.js','.json']
    },
    // added as a result of the available node libraries
    externals: {
        fs: '{}',
        tls: '{}',
        net: '{}',
        dns: '{}',
        readline: '{}'
    },
    output: {
        filename: 'bundle.js',
        path: path.resolve(__dirname, 'dist')
    }
};
```

Running this..
```
node dist/bundle.js
```

Results in the following runing error.
```
{ Error: Cannot find module '/node_modules/app-root-path'
    at webpackEmptyContext (webpack:///./node_modules/typeorm/platform_sync?:2:10)
    at Function.PlatformTools.load (webpack:///./node_modules/typeorm/platform/PlatformTools.js?:136:93)
    at ConnectionOptionsReader.get [as baseDirectory] (webpack:///./node_modules/typeorm/connection/ConnectionOptionsReader.js?:247:50)
    at ConnectionOptionsReader.get [as baseFilePath] (webpack:///./node_modules/typeorm/connection/ConnectionOptionsReader.js?:235:25)
    at ConnectionOptionsReader.eval (webpack:///./node_modules/typeorm/connection/ConnectionOptionsReader.js?:130:50)
[...]

PR #6608, opened 2020-08-23
Fix some implicit dependencies not bundled by webpack

mmvsk:
This PR adds some missing dependencies (dotenv, xml2js, js-yaml) into `PlatformTools.ts`, allowing webpack to bundle them.

Closes: #3152

### DIFF
--- a/src/platform/PlatformTools.ts
+++ b/src/platform/PlatformTools.ts
@@ -128,6 +128,15 @@ export class PlatformTools {
                 case "glob":
                     return require("glob");
 
+                case "dotenv":
+                    return require("dotenv");
+
+                case "xml2js":
+                    return require("xml2js");
+
+                case "js-yaml":
+                    return require("js-yaml");
+
                 case "typeorm-aurora-data-api-driver":
                     return require("typeorm-aurora-data-api-driver");
                 /**


pleerock:
But how do changing only `PlatformTools.ts` works? I thought their usages should be changed as well?

imnotjames:
Why is `PlatformTools.load` even used for things like `dotenv`, `glob`, `xml2js`, or `jsyaml`?  Those are actual dependencies - so the global/local nature that `PlatformTools.load` is trying to work around shouldn't be a case with those.

Would a different solution to this kind of problem be to drop the calls to `PlatformTools.load` for these dependencies that are full dependencies?

mmvsk:
@imnotjames Yeah, of course those libraries shouldn't be dynamically loaded with that custom loader.

Probably, it's just a rookie over-engineering attempt... But who knows, there may be a reason to do that, a kind of undocumented hack. So it would require some extended testing, or at least asking the authors of those hacks.

But right now some basic TypeORM features simply doesn't work with webpack (such as calling `getConnectionOptions` that relies on `dotenv`). The errors aren't catched during compile time but during runtime, and the errors are misleading (some module wasn't found in node_modules, that doesn't exist anymore after the bundling).

So I advice to merge this PR to at least make it work, without risking any breaking changes. For this fix I just made a quick grep to see which generic libraries are dynamically loaded and not listed in `PlatformTools.ts`. Because when listed in the switch-case, Webpack is somehow able to bundle them. It's a small change that just make it work.

And I agree, all libraries that are not specific to drivers should be removed from that custom dynamic loader.

imnotjames:
To poke at that idea a bit further I've opened a draft PR #6630

mmvsk:
> To poke at that idea a bit further I've opened a draft PR #6630

Looks good!

pleerock:
@imnotjames looks like we can close it in favor of #6630 ?

pleerock:
btw, the motivation behind PlatformTools working this way is because back 3 or 4 years ago we tried to support not only webpack, but also other bundlers like systemjs. I remember the times when systemjs was popular. But today it probably doesn't make sense because most uses webpack, and even any other bundler probably work the same way as webpack does. (systemjs was completely different)

pleerock:
in the past without "require" we could even do `<script src="typeorm.js">` (compiled into a single file using only typescript)  and run TypeORM in browser without a webpack. Cost of that was a requirement not to have 3rd party dependencies in runtime.

PR #6621, opened 2020-08-26
feat: dynamic embedded entities

nebkat:
Pass the target object and property name to embedded entity type function. Allows an entity to decide at runtime which entity it would like to embed depending on the target object.

This makes it possible to create generic entity types where an extending class can change it's embedded types by using a static property, or other class specific logic, without having to duplicate the column definition.

For example:

```typescript
export abstract class PlayerEntity<A> {
	@Column((type, target, propertyName) => target.attributesType()) attributes: A;
}

@Entity('players_2019')
export class Game2019PlayerEntity extends PlayerEntity<Game2019Attributes> {
	static attributesType = () => Game2019Attributes;
}

@Entity('players_2020')
export class Game2020PlayerEntity extends PlayerEntity<Game2020Attributes> {
	static attributesType = () => Game2020Attributes;
}
```

### DIFF
--- a/src/decorator/columns/Column.ts
+++ b/src/decorator/columns/Column.ts
@@ -90,13 +90,13 @@ export function Column(type: "hstore", options?: ColumnCommonOptions & ColumnHst
  * single table of the entity where Embedded is used. And on hydration all columns which supposed to be in the
  * embedded will be mapped to it from the single table.
  */
-export function Column(type: (type?: any) => Function, options?: ColumnEmbeddedOptions): PropertyDecorator;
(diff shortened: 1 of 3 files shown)


pleerock:
> In writing this example I've realized that this is arguably not much better than redefining the column in each class and setting the embedded type as before

I have the same feeling.

Thanks for contribution anyway!

imnotjames:
Given that we're in agreement that this feature probably shouldn't get merged in I'll be closing it out.  If anyone feels that it could be improved and be a worthy addition to TypeORM please re-open it as a new PR!

PR #6623, opened 2020-08-27
build: add oracle to CI

imnotjames:
Adds an Oracle server we can run tests against to the CI build.

This is to support PRs such as https://github.com/typeorm/typeorm/pull/6591

Unfortunately, some tests won't pass with Oracle because it cannot handle identifiers that are longer than 30 characters & table aliases are generated to be longer than that.  I've disabled those tests for Oracle.


package.json
```diff
@@ -84,6 +84,7 @@
     "mssql": "^4.3.2",
     "mysql": "^2.15.0",
     "mysql2": "^1.6.5",
+    "oracledb": "^5.0.0",
```
pleerock, line 87: The reason why we didn't have this before in `package.json` is because it will cause many problems to users who want to pull typeorm source, simply run `npm i` without any headache and continue tweaking things.

Is it still problematic to install this package? If yes, I recommend to remove it, and manually add `npm i oracledb` command to CI config

imnotjames: I did not have any issues installing this - at least, no more than sqlite3 with node-gyp.  I didn't have to install anything extra.

But - that's just me.  I tried on Linux and a Mac running high sierra (with homebrew).  Is there any way we can test that further?

pleerock: I'm talking about these steps:

https://oracle.github.io/node-oracledb/INSTALL.html#-2-quick-start-node-oracledb-installation

From what I remember it was required to do all these steps otherwise `npm i` on typeorm root would fail. That's why I'm talking, if it still necessary to do these steps for users, it's better to remove oracledb from dependencies.

imnotjames: The steps required for the install to work only apply if there's no prebuilt binary.

This does not include fetching the client libraries and setting up the ld config path which is only required if you are actually exercising the OracleDB module.

Instead, the steps include installing dependencies for building a nodejs binary extension using gyp - but ONLY if your architecture and node version does not have a prebuilt binary.  The same requirement applies for sqlite3 as it's also a binary module.

If this is a concern we can move this to a peer or optional dependency?  But then the same should be done for sqlite3 & better-sqlite.

pleerock: I didn't have any problems installing sqlite3 & better-sqlite.... Don't remember any at least. But oracle definitely is a problem to install. 

Anyway, my point is - any user should be able to just clone the repo, run `npm i` and just use it. With oracle its complicated (and we had users complaining about it few years ago). And since it looks like nothing has changed, and it's still a problem. I didn't heard anything about sqlite from users regarding to its issues.

So I suggest to remove it, and make oracle tests to run only on CI, so it won't bring problems to users on their local machines.

imnotjames: Sure, I can do that.  I'll get that later today.  And as it stands, oracle tests wouldn't run locally unless you enabled them.

Also, the issues you're probably referring to are that the client libraries were needed to import the module before v3.1.0.  However, that's been changed to not be the case

The v4 release brought a refactor to use a different node API so the builds are portable allowing for a binary distributable to be used instead of having to always compile from source - using node pre gyp just like sqlite3.

pleerock: okay, if they updated the installations steps, then we can left it in deps.

imnotjames: I'm happy to move forward in whatever direction you think is best, here.  It's fine either way by me. The one thing I'd want to make sure is that there's at least a hint of somewhere with the expected version of `oracledb` so we can easily pull it in with CI without hard-coding the version in the CI config.

pleerock:
thank you!

imnotjames:
Looks like a commit from between when I created this and when it got merged is causing the build to fail. ):

https://app.circleci.com/pipelines/github/typeorm/typeorm/1416/workflows/036dfdfc-ef4c-451a-b739-235c83f2c5f4/jobs/2474/parallel-runs/0/steps/0-114

```
github issues > #4410 allow custom filepath for FileLogger when no option is passed writes to the base path: QueryFailedError: ORA-00933: SQL command not properly ended at new QueryFailedError (src/error/QueryFailedError.ts:1:51) at handler (src/driver/oracle/OracleQueryRunner.ts:51:72)
```

imnotjames:
Looks like it's because Oracle can't handle the unquoted table name in the raw query that's executed.

PR #6624, opened 2020-08-27
bugfix: pass `ids_` to alias builder to prevent length overflow

imnotjames:
When creating the query for the `distinctAlias` - Instead of prepending `ids_` to the built alias name we include `ids_` as a prefix.

This means that the generated column alias could potentially be 4 characters longer than the maximum allowed column length defined by the driver.  In the case of Oracle, with its hilariously short identifier length, this comes up very frequently and causes the `ORA-00972` error code to emit.


pleerock:
Thank you!

PR #6630, opened 2020-08-27
techdebt: only use PlatformTools.load for optional dependencies

imnotjames:
This updates the usages of `PlatformTools.load` to only be for items that are optional dependencies - such as driver-related modules.  With this, if you attempt to use `PlatformTools.load` on something that's not specified explicitly it will fail with a `TypeError`.  Not only does this improve webpack functionality - it also enforces correct types for the modules that we import.


src/platform/PlatformTools.ts
```diff
@@ -111,40 +119,21 @@ export class PlatformTools {
...
-                    return require(name);
-
+                 * react-native-sqlite
+                 */
+                case "react-native-sqlite-storage":
+                    return require("react-native-sqlite-storage");
             }
 
         } catch (err) {
-            if (!path.isAbsolute(name) && name.substr(0, 2) !== "./" && name.substr(0, 3) !== "../") {
```
pleerock, line 142: not sure why we had this check before (I think it was a PR), are you sure we don't need it anymore?

imnotjames: This check was for the following case:

1. We are running the global typeorm
2. We want a module installed in a local project
3. We want to check to make sure we aren't trying to load a relative file

Because `PlatformTools.load` is now limited in what it can require through a specific array of modules we cannot have relative files being loaded.

As such, leaving this check in would have been fine but it would never be true - the module name cannot have these values.

imnotjames:
Looks like something's up with the merge.  I'll take a look in about an hour.  It's trying to pull in something through the load function it shouldn't.

pleerock:
Thank you!

PR #6632, opened 2020-08-28
fix: Unnecessary migrations for unsigned numeric types

### LINKED ISSUE #2943: Inappropriate migration generated

michaelbromley:
In MariaDB, unsigned numeric types (`<tiny,small,medium>int`) without an explicit `width` property set would generate migrations on every run. This is due to an error in setting the default width for unsigned types.

Fixes #2943

### DIFF
--- a/src/driver/mysql/MysqlQueryRunner.ts
+++ b/src/driver/mysql/MysqlQueryRunner.ts
@@ -1711,11 +1711,10 @@ export class MysqlQueryRunner extends BaseQueryRunner implements QueryRunner {
             && this.connection.driver.dataTypeDefaults[column.type].width;
 
         if (defaultWidthForType) {
-            // In MariaDB, the default widths of certain numeric types are 1 less than
+            // In MariaDB & MySQL 5.7, the default widths of certain numeric types are 1 less than
             // the usual defaults when the column is unsigned.
-            const isMariaDb = this.driver.options.type === "mariadb";
             const typesWithReducedUnsignedDefault = ["int", "tinyint", "smallint", "mediumint"];
-            if (isMariaDb && column.unsigned && -1 < typesWithReducedUnsignedDefault.indexOf(column.type)) {
+            if (column.unsigned && -1 < typesWithReducedUnsignedDefault.indexOf(column.type)) {
                 return (defaultWidthForType - 1) === width;
             } else {
                 return defaultWidthForType === width;
--- a/src/driver/mysql/MysqlQueryRunner.ts
+++ b/src/driver/mysql/MysqlQueryRunner.ts
@@ -1694,4 +1694,35 @@ export class MysqlQueryRunner extends BaseQueryRunner implements QueryRunner {
         return result[0]["version"];
     }
 
+    /**
+     * Checks if column display width is by default.
+     */
+    protected isDefaultColumnWidth(table: Table, column: TableColumn, width: number): boolean {
+        // if table have metadata, we check if length is specified in column metadata
+        if (this.connection.hasMetadata(table.name)) {
+            const metadata = this.connection.getMetadata(table.name);
+            const columnMetadata = metadata.findColumnWithDatabaseName(column.name);
+            if (columnMetadata && columnMetadata.width)
+                return false;
+        }
+
+        const defaultWidthForType = this.connection.driver.dataTypeDefaults
+            && this.connection.driver.dataTypeDefaults[column.type]
+            && this.connection.driver.dataTypeDefaults[column.type].width;
+
+        if (defaultWidthForType) {
+            // In MariaDB, the default widths of certain numeric types are 1 less than
+            // the usual defaults when the column is unsigned.
+            const isMariaDb = this.driver.options.type === "mariadb";
+            const typesWithReducedUnsignedDefault = ["int", "tinyint", "smallint", "mediumint"];
+            if (isMariaDb && column.unsigned && -1 < typesWithReducedUnsignedDefault.indexOf(column.type)) {
+                return (defaultWidthForType - 1) === width;
+            } else {
+                return defaultWidthForType === width;
+            }
+        }
+
+        return false;
+    }
+
 }
--- a/src/driver/mysql/MysqlQueryRunner.ts
+++ b/src/driver/mysql/MysqlQueryRunner.ts
@@ -1296,6 +1296,7 @@ export class MysqlQueryRunner extends BaseQueryRunner implements QueryRunner {
                     tableColumn.name = dbColumn["COLUMN_NAME"];
                     tableColumn.type = dbColumn["DATA_TYPE"].toLowerCase();
 
+                    tableColumn.unsigned = tableColumn.zerofill ? true : dbColumn["COLUMN_TYPE"].indexOf("unsigned") !== -1;
                     if (this.driver.withWidthColumnTypes.indexOf(tableColumn.type as ColumnType) !== -1) {
                         const width = dbColumn["COLUMN_TYPE"].substring(dbColumn["COLUMN_TYPE"].indexOf("(") + 1, dbColumn["COLUMN_TYPE"].indexOf(")"));
                         tableColumn.width = width && !this.isDefaultColumnWidth(table, tableColumn, parseInt(width)) ? parseInt(width) : undefined;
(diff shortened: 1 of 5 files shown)


src/query-runner/BaseQueryRunner.ts
```diff
@@ -324,10 +324,18 @@ export abstract class BaseQueryRunner {
...
 
-        if (this.connection.driver.dataTypeDefaults
+        const defaultWidthForType = this.connection.driver.dataTypeDefaults
             && this.connection.driver.dataTypeDefaults[column.type]
-            && this.connection.driver.dataTypeDefaults[column.type].width) {
-            return this.connection.driver.dataTypeDefaults[column.type].width === width;
+            && this.connection.driver.dataTypeDefaults[column.type].width;
+
+        if (defaultWidthForType) {
+            // the default widths of these numeric types are 1 less when the column is unsigned.
```
pleerock, line 332: is this statement true for all databases?

michaelbromley: No, actually only MariaDB. The method this code lives in (`isDefaultColumnWidth`) is only called for the MySQL driver, as stated in the comment above it, and it turns out that the method is actually only hit when using MariaDB specifically, because [this line](https://github.com/typeorm/typeorm/pull/6632/files/414a3faada18a7574e7dfba98006304cb2df0d36#diff-f449ebcfffaeef0bb22211a613c8e283R1301) always results in `width = undefined` for MySQL due to it's representation of `dbColumn["COLUMN_TYPE"]` not including the width part.

pleerock: Having this check for all drivers is confusing. We need to move this logic to a driver-specific method, e.g. `driver.compareColumnWidths` or something like that.

michaelbromley: Makes sense, I agree. I can do this refractor as an addition to this PR.

Ps if you are planning to make a release soon let me know so I can do the changes in time.

pleerock: yeah its already quite a while I didn't make a release, so yeah, expect a release soon

michaelbromley: OK I just looked into this more. The `isDefaultColumnWidth` method is actually used in 2 QueryRunners: MysqlQueryRunner & AuroraQueryRunner. I guess that's why it was put on the base class.

I don't see any tests for Aurora in CI, so I've no way to be sure that my changes are valid for Aurora. 

I question the value of putting it in the Driver interface because as far as I can tell it is an (almost useless) feature of MySQL _only_. [Here's a good explanation](https://alexander.kirk.at/2007/08/24/what-does-size-in-intsize-of-mysql-mean/) of what it actually does.

Here's another suggestion:

1. Move the `isDefaultColumnWidth` method from the Base class to both MysqlQueryRunner and AuroraQueryRunner.
2. Leave the AuroraQueryRunner version as it was before the fix, since I do not know whether it breaks anything there.
3. In the MysqlQueryRunner include the changes from this PR, with an additional check of the existing `isMariaDb` boolean to further make sure (and make explicit) that the fix only applies to MariaDB.

What do you think?

pleerock: yeah, go ahead

michaelbromley:
Strangely that last commit now causes a test failure. Is there any way to tell from CircleCI which DB driver is failing?

0xCAP:
This would be an awesome fix. I'm getting redundant migrations being generated all the time!

pleerock:
looks like you need to figure it out from the failed SQL

michaelbromley:
OK I figured out the failure is happening with MySQL 5.7, I tested locally with MySQL 8. I'll get this fixed.

michaelbromley:
@pleerock tests all passing, looks good to merge 👍

src/driver/mysql/MysqlQueryRunner.ts
```diff
@@ -1296,6 +1296,7 @@ export class MysqlQueryRunner extends BaseQueryRunner implements QueryRunner {
                     tableColumn.name = dbColumn["COLUMN_NAME"];
                     tableColumn.type = dbColumn["DATA_TYPE"].toLowerCase();
 
+                    tableColumn.unsigned = tableColumn.zerofill ? true : dbColumn["COLUMN_TYPE"].indexOf("unsigned") !== -1;
```
pleerock, line 1299: `tableColumn.zerofill` is set below, why did you move it?

src/driver/mysql/MysqlQueryRunner.ts
```diff
@@ -1694,4 +1694,34 @@ export class MysqlQueryRunner extends BaseQueryRunner implements QueryRunner {
...
+
+        const defaultWidthForType = this.connection.driver.dataTypeDefaults
+            && this.connection.driver.dataTypeDefaults[column.type]
+            && this.connection.driver.dataTypeDefaults[column.type].width;
+
+        if (defaultWidthForType) {
+            // In MariaDB & MySQL 5.7, the default widths of certain numeric types are 1 less than
+            // the usual defaults when the column is unsigned.
+            const typesWithReducedUnsignedDefault = ["int", "tinyint", "smallint", "mediumint"];
+            if (column.unsigned && -1 < typesWithReducedUnsignedDefault.indexOf(column.type)) {
```
pleerock, line 1717: `column.unsigned && -1` can we replace this check to something more obvious?

src/driver/mysql/MysqlQueryRunner.ts
michaelbromley: Good catch. The value of `tableColumn.unsigned` is used inside the `isDefaultColumnWidth()` method, so it needs to be set prior to checking the width, which is why I moved it up. I missed the dependency on the `zerofill` value. I'll move the `zerofill` assignment above the `unsigned` assignment.

src/driver/mysql/MysqlQueryRunner.ts
michaelbromley: Will do

michaelbromley:
Those changes are done now 👍

pleerock:
Thank you

PR #6634, opened 2020-08-28
fix: Unnecessary migrations for fulltext indices

### LINKED ISSUE #6633: Postgres, migrations: Fulltext indices continually dropped & re-created
**Issue type:**

[ x ] bug report
[ ] feature request
[ ] documentation issue

**Database system/driver:**

[ ] `cordova`
[ ] `mongodb`
[ ] `mssql`
[ ] `mysql` / `mariadb`
[ ] `oracle`
[ x ] `postgres`
[ ] `cockroachdb`
[ ] `sqlite`
[ ] `sqljs`
[ ] `react-native`
[ ] `expo`

**TypeORM version:**

[ x ] `latest`
[ ] `@next`

Given the following entity:

```ts
@Entity()
export class Test {
    @PrimaryColumn()
    id: number;

    @Index("description_index", { fulltext: true })
    @Column()
    description: string;
}
```

Running a migration with postgres will _always_ generate the following queries, even though no changes are made to the schema:

```ts
export class someMigration1598601015234 implements MigrationInterface {
        await queryRunner.query(`DROP INDEX "description_index"`, undefined);
        await queryRunner.query(`CREATE INDEX "description_index" ON "test" ("description") `, undefined);
    }

        await queryRunner.query(`DROP INDEX "description_index"`, undefined);
        await queryRunner.query(`CREATE INDEX "description_index" ON "test" ("description") `, undefined);
    }
}
```

The reason seems to be that postgres does not support fulltext indices, but in `RdbmsSchemaBuilder.ts` the `tableIndex.isFulltext` property is checked even for postgres, which makes any fulltext indices give false positives:

https://github.com/typeorm/typeorm/blob/d1ed5723bb55d4124be13b21e57d9f729d0bae1a/src/schema-builder/RdbmsSchemaBuilder.ts#L280-L281

This may be related to #5655 and #5159. It also may occur with other database types.

I intend to submit a PR with a fix shortly.

michaelbromley:
Fixes #6633 (see issue for root cause explanation)

### DIFF
--- a/src/driver/Driver.ts
+++ b/src/driver/Driver.ts
@@ -194,6 +194,11 @@ export interface Driver {
     isUUIDGenerationSupported(): boolean;
 
     /**
+     * Returns true if driver supports fulltext indices.
+     */
+    isFullTextColumnTypeSupported(): boolean;
+
+    /**
      * Creates an escaped parameter.
      */
     createParameter(parameterName: string, index: number): string;
--- a/src/driver/aurora-data-api/AuroraDataApiDriver.ts
+++ b/src/driver/aurora-data-api/AuroraDataApiDriver.ts
@@ -751,6 +751,13 @@ export class AuroraDataApiDriver implements Driver {
     }
 
     /**
+     * Returns true if driver supports fulltext indices.
+     */
+    isFullTextColumnTypeSupported(): boolean {
+        return true;
+    }
+
+    /**
      * Creates an escaped parameter.
      */
     createParameter(parameterName: string, index: number): string {
--- a/src/driver/cockroachdb/CockroachDriver.ts
+++ b/src/driver/cockroachdb/CockroachDriver.ts
@@ -650,6 +650,13 @@ export class CockroachDriver implements Driver {
     }
 
     /**
+     * Returns true if driver supports fulltext indices.
+     */
+    isFullTextColumnTypeSupported(): boolean {
+        return false;
+    }
+
+    /**
      * Creates an escaped parameter.
      */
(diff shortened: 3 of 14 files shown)


pleerock:
Thanks for PR!

What is the point of using `fulltext` flag for a database that doesn't support it? TypeORM's entities don't try to be database-agnostic.

I can merge it, however it would be better if we change it a bit - we probably need to have a boolean flag in driver like `isFullTextColumnTypeSupported(): boolean`. If you don't want to return back to it, no problem, I'll merge it.

michaelbromley:
> What is the point of using fulltext flag for a database that doesn't support it? TypeORM's entities don't try to be database-agnostic.

Because in my framework (vendure.io) I _do_ want the entities to be DB-agnostic. The end-user passes in the ConnectionOptions to specify which DB they want to use. I have a `SearchIndexItem` entity using fulltext fields, and I aim to support MySQL/MariaDB/Postgres/SQLite. Having DB-specific models is perhaps possible (dynamically defining the metadata at runtime) but it is certainly not convenient or intuitive to figure out what's going on.

I think it makes sense that for non-supported features like this, the DB just totally ignores them and they shouldn't have unexpected side-effects like with the current issue.

> I can merge it, however it would be better if we change it a bit - we probably need to have a boolean flag in driver like isFullTextColumnTypeSupported(): boolean. If you don't want to return back to it, no problem, I'll merge it.

This makes sense, and is consistent with the handling of the other features where support differs between drivers. I can push another commit later this week.

michaelbromley:
Done!

pleerock:
Thank you!

PR #6637, opened 2020-08-28
chore: instead of require, import chalk from chalk

imnotjames:
There's not much point to require `chalk` in the way it's been done - so use `import` instead.

This allows us to use the type definitions for chalk which we weren't before - `require` makes it `any` instead of the `chalk` types


pleerock:
Thanks!

PR #6638, opened 2020-08-28
fix: Migration issues with scale & precision in sqlite/sql.js

### LINKED ISSUE #6636: SQLite, sqljs migration issues with scale & precision
**Issue type:**

[ x ] bug report
[ ] feature request
[ ] documentation issue

**Database system/driver:**

[ ] `cordova`
[ ] `mongodb`
[ ] `mssql`
[ ] `mysql` / `mariadb`
[ ] `oracle`
[ ] `postgres`
[ ] `cockroachdb`
[ x ] `sqlite`
[ x ] `sqljs`
[ ] `react-native`
[ ] `expo`

**TypeORM version:**

[ x ] `latest`
[ ] `@next`

Given the following entity:
```ts
@Entity()
export class Test {

    @PrimaryColumn()
    id: number;

    @Column({ nullable: true, precision: 6 })
    startedAt?: Date;

    @Column({ type: 'decimal', precision: 5, scale: 2 })
    value: number;

}
```

Running a migration on a newly-synced SQLite/sql.js DB results in the following uneccessary migration commands:

```ts
[
     {
        query: 'CREATE TABLE "temporary_test" ("id" integer PRIMARY KEY NOT NULL, "startedAt" datetime(6), "value" decimal(5,2) NOT NULL)',
        parameters: undefined
    },
    {
        query: 'INSERT INTO "temporary_test"("id", "startedAt", "value") SELECT "id", "startedAt", "value" FROM "test"',
        parameters: undefined
    },
    {query: 'DROP TABLE "test"', parameters: undefined},
    {
        query: 'ALTER TABLE "temporary_test" RENAME TO "test"',
        parameters: undefined
    }
]
```

This is caused by a false positive in the `findChangedColumns()` method:

https://github.com/typeorm/typeorm/blob/d1ed5723bb55d4124be13b21e57d9f729d0bae1a/src/driver/sqlite-abstract/AbstractSqliteDriver.ts#L553-L554

because the **precision** and **scale** arguments are not correctly inferred when reading the table. Then length property, however, _is_ correctly inferred:

https://github.com/typeorm/typeorm/blob/d1ed5723bb55d4124be13b21e57d9f729d0bae1a/src/driver/sqlite-abstract/AbstractSqliteQueryRunner.ts#L829-L839

All that is missing is to do the same for precision & scale.

I intend to submit a PR shortly.

michaelbromley:
Specifying precision or scale properties on columns with SQLite/sql.js would result in migrations being generated even on an unchanged schema.

This was due to the precision and scale arguments not correctly being inferred when reading the table. This change handles scale and precision in the same way that "length" was already being correctly handled.

Fixes #6636

### DIFF
--- a/test/github-issues/6636/issue-6636.ts
+++ b/test/github-issues/6636/issue-6636.ts
@@ -1,0 +1,25 @@
+import { Connection } from "../../../src";
+import { closeTestingConnections, createTestingConnections, reloadTestingDatabases } from "../../utils/test-utils";
+import { Test } from "./entity/Test";
+import { expect } from "chai";
+
+describe("github issues > #6636 migration issues with scale & precision", () => {
+
+    let connections: Connection[];
+    before(async () => connections = await createTestingConnections({
+        entities: [Test],
+        enabledDrivers: ["sqljs", "sqlite", "better-sqlite3"],
+    }));
+    beforeEach(() => reloadTestingDatabases(connections));
+    after(() => closeTestingConnections(connections));
+
+    it("should not create migrations columns with precision", () => {
+        Promise.all(connections.map(async (connection) => {
+                const sqlInMemory = await connection.driver.createSchemaBuilder().log();
+                expect(sqlInMemory.upQueries).to.eql([]);
+                expect(sqlInMemory.downQueries).to.eql([]);
+            }
+        ))
+    });
+
+});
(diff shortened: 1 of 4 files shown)


test/github-issues/6636/issue-6636.ts
```diff
@@ -0,0 +1,25 @@
...
+    let connections: Connection[];
+    before(async () => connections = await createTestingConnections({
+        entities: [Test],
+        enabledDrivers: ["sqljs", "sqlite", "better-sqlite3"],
+    }));
+    beforeEach(() => reloadTestingDatabases(connections));
+    after(() => closeTestingConnections(connections));
+
+    it("should not create migrations columns with precision", () => {
+        Promise.all(connections.map(async (connection) => {
```
pleerock, line 17: there is no `await`, this will cause issues

Issue #6640, opened 2020-08-30
Cannot ORDER BY json field in query with JOIN and pagination


szlori:
**Issue type:**

[x] bug report
[ ] feature request
[ ] documentation issue

**Database system/driver:**

[ ] `cordova`
[ ] `mongodb`
[ ] `mssql`
[ ] `mysql` / `mariadb`
[ ] `oracle`
[x] `postgres`
[ ] `cockroachdb`
[ ] `sqlite`
[ ] `sqljs`
[ ] `react-native`
[ ] `expo`

**TypeORM version:**

[x] `latest`
[ ] `@next`
[ ] `0.x.x` (or put your version here)

**Steps to reproduce or a small repository showing the problem:**

This seems related to #3356 and #5670, but more specific about JSON (or JSONB) fields and I have a fix proposal (will submit PR) together with test that can illustrate the issue.

Short background, in POSTGRES you can have JSON/JSONB columns and then access fields in the JSON object with `->` or `->>` operators. So if I want to use such a field of JSON column in a query, I can use `"entity.property ->'field'"`.

Notice the **space** between `entity.property` and `->`, this is allowed by POSTGRES syntax, and using this ensures that TypeORM will correctly process `entity.property` part and will leave the `->'field'` unchanged. (NOTE: the same way -using the space- some other POSTGRES syntax can be sneaked in after property, such as `::text` or `::int` for casting.) 

So this works almost everywhere, which is great, given the fact TypeORM doesn't have much support for json fields, but it doesn't work in the case mentioned by the title, when all the below 3 conditions are satisfied:

- the json field is used in ORDER BY
- query has a JOIN
- pagination (.take()) is used

Ex query (note that `form.extraSettings` is a JSONB column and it has a `someNb` field):

```
    const qb = connection.createQueryBuilder(Form, "form")
      .innerJoinAndSelect("form.formTmpl", "formTmpls", "formTmpls.origin = :orig", { orig: 1 })
      .orderBy({
        "form.extraSettings ->'someNb'": {
            order: "DESC",
            nulls: "NULLS LAST",
          }
      })
      .take(2);
```

The reason is that this case is handled separately in `createOrderByCombinedWithSelectExpression()` and here the specified property is not transformed in the same way as in most other parts, but here it is assumed that only an `entity.property` (i.e. `[alias, propertyPath]`) is present in order by key.

However, this does not need to be the case, ORDER BY works just fine when ->field is added in all other cases.

[...]

PR #6644, opened 2020-08-31
fix: support ORDER BY json field with JOIN and pagination

szlori:
Closes: #6640 

Allow specifying jsonb fields (`property ->'field'` in Postgres) in a GROUP BY, even for the case the query has JOIN and uses pagination.

### DIFF
--- a/src/query-builder/SelectQueryBuilder.ts
+++ b/src/query-builder/SelectQueryBuilder.ts
@@ -2051,10 +2051,17 @@ export class SelectQueryBuilder<Entity> extends QueryBuilder<Entity> implements 
                 if (orderCriteria.indexOf(".") !== -1) {
                     const criteriaParts = orderCriteria.split(".");
                     const aliasName = criteriaParts[0];
-                    const propertyPath = criteriaParts.slice(1).join(".");
+                    const fieldSpec = criteriaParts.slice(1).join(".");
+                    let propertyPath = fieldSpec;
+                    let rawFieldSpecs = "";
                     const alias = this.expressionMap.findAliasByName(aliasName);
-                    const column = alias.metadata.findColumnWithPropertyPath(propertyPath);
-                    return this.escape(parentAlias) + "." + this.escape(DriverUtils.buildAlias(this.connection.driver, aliasName, column!.databaseName));
+                    const spIdx = propertyPath.indexOf(" ");
+                    if (spIdx !== -1) {
+                        propertyPath = propertyPath.slice(0, spIdx);
+                        rawFieldSpecs = propertyPath.slice(spIdx);
+                    }
+                    const column = alias.metadata.findColumnWithPropertyName(propertyPath);
+                    return this.escape(parentAlias) + "." + this.escape(DriverUtils.buildAlias(this.connection.driver, aliasName, column!.databaseName)) + rawFieldSpecs;
                 } else {
                     if (this.expressionMap.selects.find(select => select.selection === orderCriteria || select.aliasName === orderCriteria))
                         return this.escape(parentAlias) + "." + orderCriteria;
@@ -2069,10 +2076,17 @@ export class SelectQueryBuilder<Entity> extends QueryBuilder<Entity> implements 
             if (orderCriteria.indexOf(".") !== -1) {
                 const criteriaParts = orderCriteria.split(".");
                 const aliasName = criteriaParts[0];
-                const propertyPath = criteriaParts.slice(1).join(".");
+                const fieldSpec = criteriaParts.slice(1).join(".");
+                let propertyPath = fieldSpec;
+                let rawFieldSpecs = "";
                 const alias = this.expressionMap.findAliasByName(aliasName);
+                const spIdx = propertyPath.indexOf(" ");
+                if (spIdx !== -1) {
+                    propertyPath = propertyPath.slice(0, spIdx);
+                    rawFieldSpecs = propertyPath.slice(spIdx);
+                }
                 const column = alias.metadata.findColumnWithPropertyPath(propertyPath);
-                orderByObject[this.escape(parentAlias) + "." + this.escape(DriverUtils.buildAlias(this.connection.driver, aliasName, column!.databaseName))] = orderBys[orderCriteria];
+                orderByObject[this.escape(parentAlias) + "." + this.escape(DriverUtils.buildAlias(this.connection.driver, aliasName, column!.databaseName)) + rawFieldSpecs] = orderBys[orderCriteria];
             } else {
                 if (this.expressionMap.selects.find(select => select.selection === orderCriteria || select.aliasName === orderCriteria)) {
                     orderByObject[this.escape(parentAlias) + "." + orderCriteria] = orderBys[orderCriteria];
--- a/test/github-issues/6640/issue-6640.ts
+++ b/test/github-issues/6640/issue-6640.ts
@@ -1,0 +1,93 @@
+import "reflect-metadata";
+import {closeTestingConnections, createTestingConnections, reloadTestingDatabases} from "../../utils/test-utils";
+import {Connection} from "../../../src/connection/Connection";
+import {Form} from "./entity/Form";
+import {FormTmpl} from "./entity/FormTmpl";
+import { expect } from "chai";
+
+
+
+describe("github issues > #6640 Cannot ORDER BY json field in query with JOIN and pagination", () => {
+
+    let connections: Connection[];
+    before(async () => connections = await createTestingConnections({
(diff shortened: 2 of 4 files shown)


imnotjames:
This breaks CockroachDB tests.  Can you look into that?

szlori:
> This breaks CockroachDB tests. Can you look into that?

OK, let me install it and give it a try.

imnotjames:
Can use `docker-compose` and the configs we have in the root of our repo if that helps :)

szlori:
Hmmm, I checked the circleci error and it is some timeout, running the same test locally succeeds for me.
How to proceed?

imnotjames:
I'll try kicking it off from failed and we'll see what happens -- https://app.circleci.com/pipelines/github/typeorm/typeorm/1385/workflows/aa4e567b-d052-4874-80bd-3afa2b8644db/jobs/3986

szlori:
Still failing, but now seems even earlier stage...

imnotjames:
I restarted the entire build from the start this time. :)

szlori:
Aha, no issue after all :)

test/github-issues/6640/issue-6640.ts
```diff
@@ -0,0 +1,93 @@
...
+        someNb: 40,
+        someStr: "extra4",
+      };
+      await repo.save(form4);
+
+      const orig = 1;
+      const qb = connection.createQueryBuilder(Form, "form")
+        .innerJoinAndSelect("form.formTmpl", "formTmpls", "formTmpls.origin = :orig", { orig })
+        .orderBy({
+          "form.extraSettings ->'someNb'": {
```
imnotjames, line 79: What about using the actual JSON notation, `form.extraSettings->'someNb'`?

src/query-builder/SelectQueryBuilder.ts
```diff
@@ -1937,10 +1937,17 @@ export class SelectQueryBuilder<Entity> extends QueryBuilder<Entity> implements
         const selectString = Object.keys(orderBys)
             .map(orderCriteria => {
                 if (orderCriteria.indexOf(".") !== -1) {
-                    const [aliasName, propertyPath] = orderCriteria.split(".");
```
imnotjames, line 1940: Is it possible to treat the `orderBy` similarily to how we treat `WHERE`?  That way we can do something more generalized?

As it stands, I'd want to gate this behind a postgres check

test/github-issues/6640/issue-6640.ts
```diff
@@ -0,0 +1,93 @@
...
+      form4.settings = "setting4";
+      form4.extraSettings = {
+        someNb: 40,
+        someStr: "extra4",
+      };
+      await repo.save(form4);
+
+      const orig = 1;
+      const qb = connection.createQueryBuilder(Form, "form")
+        .innerJoinAndSelect("form.formTmpl", "formTmpls", "formTmpls.origin = :orig", { orig })
```
imnotjames, line 77: Do we actually need this? Couldn't we simplify the model significantly and remove the extra pieces we aren't using?

For tests, I think we only really need...

```
@Entity()
export class Test {
  @PrimaryColumn()
  id: number;

  // JSONB column
  @Column("jsonb", {nullable: true})
  json?: any;
}
```

We don't even need to read the data in the json blob, right?  We can compare the `id` values & make sure the sort order is correct in that way.

test/github-issues/6640/issue-6640.ts
```diff
@@ -0,0 +1,93 @@
...
+        someNb: 5,
+        someStr: "extra3",
+      };
+      await repo.save(form3);
+
+      const form4 = new Form();
+      form4.formTmplId = 1;
+      form4.settings = "setting4";
+      form4.extraSettings = {
+        someNb: 40,
```
imnotjames, line 70: Just to throw things off, can you include other JSON data in the fields as well - stuff that alphabetically comes before, etc

test/github-issues/6640/issue-6640.ts
```diff
@@ -0,0 +1,93 @@
...
+      const orig = 1;
+      const qb = connection.createQueryBuilder(Form, "form")
+        .innerJoinAndSelect("form.formTmpl", "formTmpls", "formTmpls.origin = :orig", { orig })
+        .orderBy({
+          "form.extraSettings ->'someNb'": {
+              order: "DESC",
+              nulls: "NULLS LAST",
+            }
+        })
+        .take(2);
```
imnotjames, line 84: Is it possible to do the test with `take` separate from this case?

imnotjames (changes requested):

test/github-issues/6640/issue-6640.ts
szlori: Using the JSON notation (without the space before '->') does not work in other places either. Typeorm will not treat them as separate fields, only the '.' is recognized for it. And this will mess up the entity to table column translation.

But using the space before '->' does the trick in most cases (i.e. other parts of the code).

src/query-builder/SelectQueryBuilder.ts
szlori: Hmmm, that was my original idea actually, but turns out it's not as straightforward here, because of the generated alias.

But IIRC first I tried to use/change a bit `replacePropertyNames()` but it didn't really work or the change was too invasive. If you think it's worth taking another look, I think implementing a separate similar function for this case would be the way to go.

The way it is now, though I indeed only considered real-life case for Postgres, this is relatively generic as it just makes sure that the entity to table column translation is only done on the string up till the space and what's after the space is left unchanged...

test/github-issues/6640/issue-6640.ts
szlori: Significantly no... but can simplify (I simplified our real-life example, so at one point I got tired removing stuff ;) )

But we need to have JOIN, ORDER BY and PAGINATION, only in that case we end up in that function I patched. In other case it works already.

test/github-issues/6640/issue-6640.ts
szlori: In `extraSettings->someStr` that would alphabetically come before the numbers?

test/github-issues/6640/issue-6640.ts
szlori: Not quite understand, could you explain?

imnotjames: You're testing a few different functionalities here.  One is order by with JSON notation, the other is grouping by & ordering by a JSON notation.

Ideally we would have one test that's just a very basic case - eg, just ordering by the JSON field.  Nothing else, that way it's in isolation.

Then a separate test which joins & uses take.

szlori: Yes, indeed this issue occurs only in this case when several functionalities are used together. This is not an issue with json fields, it's only this specific case using json fields. That is why I test this particular case.

If I test simpler cases (which were working fine before), it will be a non-regression test. I don't think these tests are supposed to cover non-regression... right?

imnotjames: :thinking: the notation `property ->'field'` works to order normally?  Wouldn't it get escaped erroneously?

szlori: Yes, it works, `replacePropertyNames()`, which is used in other parts (ex. in where) will translate `property`, but will leave `->'field'` untouched, due to the separating space and the fact that there is no property that can match "->'field'".

And therefore in the resulting SQL we properly get `alias.column_name ->'field'`.

pbn4:
What's the status of this PR? I'd really appreciate it.

pbn4:
@imnotjames @szlori (I forgot to add a mention)

laudibert:
Hello, is this PR not ok to merge ? 

If it is, can it be merged ?

If it not, I am willing to help finishing it by completing any missing tests or anything else on a new PR. But I am not sure what's missing here.

@imnotjames @szlori

laudibert:
@pleerock Sorry to tag you, I don't know who else to ask about this as both the reviewer and the author seem to have abandoned this PR.

Cyberuben:
@szlori @imnotjames could you please review or fix this PR? I really need this change

pleerock:
I'll close this PR since its abandoned and not sure if it has a right solution to the problem.
Main `SelectQueryBuilder`'s responsibility is to join SQL string parts into a single SQL query.
And for some reason `.orderBy` designed not the best way to do it. I think the proper solution (and simplest) to the problem is to allow any string into `.orderBy` that will be concatenated to a `ORDER BY ` section of the SQL query.

e.g.:

```ts
.orderBy(`form.extraSettings ->'someNb'`)
```

most work as expected. Please feel free to contribute in a new PR.

PR #6660, opened 2020-09-02
Support absolute paths in migrationsDir for the CLI

tomasreimers:
Right now the CLI will ALWAYS prepend the CWD to `options.cli.migrationsDir`. 

This prevents us from using absolute paths there. 

How would we feel about changing that to support absolute paths (by only prepending CLI if the path DOES NOT start with "/")? If we're open to it, I'm happy to make the change for the other CLI commands as well.

### DIFF
--- a/src/commands/MigrationCreateCommand.ts
+++ b/src/commands/MigrationCreateCommand.ts
@@ -59,7 +59,7 @@ export class MigrationCreateCommand implements yargs.CommandModule {
                 } catch (err) { }
             }
 
-            const path = process.cwd() + "/" + (directory ? (directory + "/") : "") + filename;
+            const path = (directory.startsWith("/") ? "" : process.cwd() + "/") + (directory ? (directory + "/") : "") + filename;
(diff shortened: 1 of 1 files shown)


pleerock:
Looks good, thanks

imnotjames:
Looks like this may have broke the build.

```
src/commands/MigrationCreateCommand.ts:62:27 - error TS2571: Object is of type 'unknown'.

62             const path = (directory.startsWith("/") ? "" : process.cwd() + "/") + (directory ? (directory + "/") : "") + filename;
                             ~~~~~~~~~


Found 1 error.
```

Thoughts?

tomasreimers:
Immediate thought is: go ahead and revert to unblock the build.

I was expecting to have a little more back and forth on the commit (and to
add the behavior to the other methods) so I don't mind rewriting and
figuring out the TS error then?

Does that work?

pleerock:
I pushed a type fix, not sure how I merged it without seeing that build is failing 🤔

PR #6671, opened 2020-09-04
test: Testing that the discriminatorValue gets saved for ChildEntity when saved by cascade

Garethp:
Adding a test for [the fix](https://github.com/typeorm/typeorm/pull/6219#issuecomment-687258512) for #2544


pleerock:
thank you!

Issue #4410, opened 2019-07-10
Passing File Path for FileLogger as an option


halufa:
**Issue type:**

[ ] question
[ ] bug report
[x] feature request
[ ] documentation issue

**Database system/driver:**

[ ] `cordova`
[ ] `mongodb`
[ ] `mssql`
[ ] `mysql` / `mariadb`
[ ] `oracle`
[ ] `postgres`
[ ] `cockroachdb`
[ ] `sqlite`
[ ] `sqljs`
[ ] `react-native`
[ ] `expo`

**TypeORM version:**

[x] `latest`
[ ] `@next`
[ ] `0.x.x` (or put your version here)

**Steps to reproduce or a small repository showing the problem:**

The FileLogger is working really well. The only problem is the file path it's using. I would like to have the file path destination of the file log as an option instead of being hardcoded.

PlatformTools_1.PlatformTools.appendFileSync(basePath + "/ormlogs.log", strings.join("\r\n") + "\r\n");

Thank you

halufa:
It's up to you to implement my request or not. Meanwhile I came up with a simple solution using a custom logger extending FileLogger.

export class CustFileLogger extends FileLogger implements Logger {

    protected write(strings: string | string[]) {
        strings = strings instanceof Array ? strings : [strings];
        const basePath = AppSettings.env.logsPath
        const now = moment()
        strings = strings.map((str) => "[" + now.toLocaleString() + "]" + str)
        appendToTextFile(strings.join("\r\n") + "\r\n", basePath + "/ormlogs_" + moment(now).format('YYYYMMDD') + ".log")
    }
}

PR #6672, opened 2020-09-05
escape table name for Github #4410

imnotjames:
Merging in #6623 caused a test that was merged in slightly before #6642 to start failing because the SQL written is not compatible with Oracle.

As such, this updates that test to properly escape the table name queried.


pleerock:
Thank you!

PR #6676, opened 2020-09-06
fix: oracle table escaping

pirouet:
The way the table name is made up in typeorm isn't always right. It can cause the table name to be escaped and truncated (based on underscores) and the inclusion of quotes around the table name means pipelined queries cannot be easily run. With this change, ther

A new option on @Entity decorator exists called `exact` and the purpose of this is to ensure that the name of the table does not get modified in anyway and is used as it. This aids where you need to call a function (within a package) pipelined but the length is greater than 30 characters.

Tests to follow

### DIFF
--- a/docs/decorator-reference.md
+++ b/docs/decorator-reference.md
@@ -60,6 +60,7 @@ You can also specify some additional entity options:
 * `engine` - database engine to be set during table creation (works only in some databases).
 * `synchronize` - entities marked with `false` are skipped from schema updates.
 * `orderBy` - specifies default ordering for entities when using `find` operations and `QueryBuilder`.
+* `exact` - determines if the `name` should not be modified and used exactly as it is specified
 
 Example:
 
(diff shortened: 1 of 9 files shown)


pirouet:
Unsure why the test failed - excluding SAP from local testing shows this test passing:

```
  ts-typeorm git:(oracle-table-escaping) ✗ npm test -- --grep="column kinds > create date column"

> typeorm@0.2.25 test /Volumes/Developer/ts-typeorm
> rimraf ./build && tsc && mocha --file ./build/compiled/test/utils/test-setup.js --bail --recursive --timeout 60000 ./build/compiled/test "--grep=column kinds > create date column"



  column kinds > create date column
    ✓ create date column should automatically be set by a database (124ms)
    ✓ create date column can also be manually set by user (109ms)
    ✓ create date column should not be updated automatically on every change (1157ms)
    ✓ create date column should set a custom date when specified (165ms)


  4 passing (6s)
```

pirouet:
@pleerock Hey! Sorry to tag you but would you happen to have any ideas if this is just the CI setup, or something I have done wrong that would explain the Oracle test failing? As mentioned in the comments, running tests locally passes that test - and all others for that matter too

Thanks in advance!
Michael

pleerock:
I guess its your changes, since there are other PRs (newer) which successfully pass.

malvisa01:
Hi @pirouet pirouet 

I am facing same issue. But I can see this commit is not merged to master branch. Is there any other way to solve this issue.
Thank you.

pirouet:
@malvisa01 hi there! - Yeah, this is one that's still on my list, not quite managed to get around to the finishing pieces needed to get Oracle to work properly but it's one I'm desperate to get back at again

malvisa01:
@pirouet Thankyou for your replay.
Please update is there anything. :)

PR #6681, opened 2020-09-08
Update select-query-builder.md

TulioMolina:
Found a typo on the "**Adding WHERE expression**" section: ```user.firstName``` is used in one of the SQL snippets, but you were previously referring to ```user.id``` for that example.


pleerock:
Thank you

TulioMolina:
Never mind!

Issue #4415, opened 2019-07-10
Beautify generated SQL for migrations


moltar:
**Issue type:**

[ ] question
[ ] bug report
[x] feature request
[ ] documentation issue

**Database system/driver:**

[ ] `cordova`
[ ] `mongodb`
[ ] `mssql`
[ ] `mysql` / `mariadb`
[ ] `oracle`
[ ] `postgres`
[ ] `cockroachdb`
[ ] `sqlite`
[ ] `sqljs`
[ ] `react-native`
[ ] `expo`

**TypeORM version:**

[x] `latest`
[ ] `@next`
[ ] `0.x.x` (or put your version here)

**Steps to reproduce or a small repository showing the problem:**

It would improve readability and edit-ability of generated SQL migration strings if they were "beautified", in other words, formatted.

Right now it is just a single long line for the entire table.

Thanks.

PR #6685, opened 2020-09-09
feat: Beautify generated SQL for migrations

artysidorenko:
Allows user to pass an optional flag to beautify generated SQL for migrations

Closes: #4415

---

Had a stab at [issue #4415](https://github.com/typeorm/typeorm/issues/4415) - adds an external dependency `@sqltools/formatter` to actually do the formatting, though it's a pretty small package.

Tests may be a little overkill 🙈  - checks for the main sql drivers that the formatting matches the provided templates for an example set of tables. Below is a visual example with postgres:

<details><summary>before/without the flag</summary>

</details>
<details><summary>after/with the flag</summary>

</details>

Any questions let me know.


pleerock:
Thanks for the PR and attached screenshoots. I'm thinking if we really need a flag for this feature, maybe we can make this behaviour by default? Do you see any cons of this approach?

artysidorenko:
> Do you see any cons of this approach?

My only concern is if there's ever any formatter issue in the future, then it might be annoying to the user if they can't disable it? I think it looks pretty reliable but every now and again an issue gets raised, for example [this one](https://github.com/mtxr/vscode-sqltools/issues/605), or [this one](https://github.com/mtxr/vscode-sqltools/issues/460). Neither looks too serious but I'm always pretty risk averse if the feature is a nice-to-have rather than a must-do 😅

pleerock:
okay, let's make it flag-based for now, maybe in the future, once we decide its reliable, we'll make this by default. Thank you!

src/commands/MigrationGenerateCommand.ts
```diff
@@ -7,6 +7,7 @@ import {camelCase} from "../util/StringUtils";
 import * as yargs from "yargs";
 import {AuroraDataApiDriver} from "../driver/aurora-data-api/AuroraDataApiDriver";
 import chalk from "chalk";
+import { format } from "@sqltools/formatter/lib/sqlFormatter";
```
pleerock, line 10: is it a safe import? Don't they have it exported on `@sqltools/formatter` ?

artysidorenko: Unfortunately the TS compiler's unhappy with that (`Could not find a declaration file for module '@sqltools/formatter'`).. The declaration file is in there but I think it's unable to pick it up from the top level folder.

Require statement works ok, if you prefer that? `const { format } = require("@sqltools/formatter");`

I'm a bit newb-ish on how the packaging works though - here's the contents of `node_modules/@sqltools/formatter` if you can spot anything in that structure:

image

pleerock: okay, lets merge it as it is... but I bet if there is something wrong with typing on `@sqltools/formatter` they will release a new version with change and it can break our stuff... so I would probably fix this package version for now.

pleerock:
Thank you a lot!

test/github-issues/4415/entity/Username.ts
```diff
@@ -0,0 +1,15 @@
+import { Column } from "../../../../src/decorator/columns/Column";
+import { PrimaryColumn } from "../../../../src/decorator/columns/PrimaryColumn";
+import { Entity } from "../../../../src/decorator/entity/Entity";
```
agaidulean-theunitdev, line 3: Bruh

PR #6698, opened 2020-09-11
bugfix: sql.js v1.2+ don't support undefined parameters

### LINKED ISSUE #5720: Latest Sqljs@1.2.1 throws an error when querying
**Issue type:**

[ ] question
[x] bug report
[ ] feature request
[ ] documentation issue

**Database system/driver:**

[ ] `cordova`
[ ] `mongodb`
[ ] `mssql`
[ ] `mysql` / `mariadb`
[ ] `oracle`
[ ] `postgres`
[ ] `cockroachdb`
[ ] `sqlite`
[x] `sqljs`
[ ] `react-native`
[ ] `expo`

**TypeORM version:**

[x] `latest`
[ ] `@next`

Trying to query using the latest Sqljs and TypeORM throws an error: `QueryFailedError: TypeError: Cannot convert undefined or null to object`.

It turns out that when we have parameters in `SqljsQueryRunner#query` is undefined/null, the latest sqljs throws an error.

Harmless fix is to either skip the bind or give default value for parameters as an empty array within the SqljsQueryRunner only.

Tested in browser main thread, as well as worker thread. Both works.

imnotjames:
before sql.js 1.2 it seems undefined were treated as if they were null,
but as of 1.2 they cause a query error & fail to execute

this change swaps out any undefined parameters with `null`s

closes #5720

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -9149,9 +9149,9 @@
       "integrity": "sha1-BOaSb2YolTVPPdAVIDYzuFcpfiw="
     },
     "sql.js": {
-      "version": "1.0.0",
-      "resolved": "https://registry.npmjs.org/sql.js/-/sql.js-1.0.0.tgz",
-      "integrity": "sha512-fl/cQu8hDbu9bXNXTrmwwtS2wvHttWrXS9fzINVtedOBOJJWMjwSADR1aQ+DKmCdGNF4jrg/15AedJ5kcrWRJA==",
+      "version": "1.3.0",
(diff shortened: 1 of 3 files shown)


pleerock:
Thank you! btw titles must be named according to [conventions](https://github.com/typeorm/typeorm/blob/master/CONTRIBUTING.md#type) in order for changelog to be properly generated.

imnotjames:
🙇 oops.  Thought it was bugfix, remembered wrong.  I'll make sure in the future to use the correct prefix!

PR #6701, opened 2020-09-11
techdebt: create EntityTarget and use instead of EntitySchema / ObjectType / etc

imnotjames:
This backports the `EntityTarget` type from the `@next` branch and uses it instead of the bunches of `ObjectType`/`EntitySchema`/`string` union types.


pleerock:
Thank you!

Issue #6382, opened 2020-07-10
Remove orphaned entities from a relation?


RobbieGM:
**Issue type:**

[x] question
[ ] bug report
[ ] feature request
[ ] documentation issue

**Database system/driver:**

[ ] `cordova`
[ ] `mongodb`
[ ] `mssql`
[ ] `mysql` / `mariadb`
[ ] `oracle`
[x] `postgres`
[ ] `cockroachdb`
[ ] `sqlite`
[ ] `sqljs`
[ ] `react-native`
[ ] `expo`

**TypeORM version:**

[ ] `latest`
[ ] `@next`
[ ] `0.x.x` (or put your version here)

**Steps to reproduce or a small repository showing the problem:**

I have a one-to-many relation and I would like to delete rows from the many side of the relationship rather than unlinking them by setting their foreign key to null (the default behavior), leaving them orphaned. I saw [a PR](https://github.com/typeorm/typeorm/pull/1665) for this feature but it was rejected. Is there a way to do this, or any plans to support it in the future?

I would like to be able to do something like this:
```ts
@Entity()
@TableInheritance({ column: { name: 'type', type: 'text' } })
export default class Geolocation {
  @PrimaryGeneratedColumn()
  id!: number;

  @IsGeoJSONPoint
  @Column('geography')
  point!: Point;
}

@ChildEntity('business')
export default class BusinessGeolocation extends Geolocation {
  @ManyToOne(type => Business, business => business.geolocations, { onDelete: 'CASCADE' })
  business!: Business;
}


@Entity()
export default class Business {
  @PrimaryGeneratedColumn()
  id!: number;

  // I would like to remove orphaned business geolocations
  @OneToMany(type => BusinessGeolocation, businessGeolocation => businessGeolocation.business, { cascade: true })
  geolocations!: BusinessGeolocation[];
}

// somewhere in the API...

const business = new Business();
business.geolocations = [{ type: 'Point', coordinates: [3, 5] }, { type: 'Point', coordinates: [8, 2] }];
await manager.save(business);
business.geolocations = [{ type: 'Point', coordinates: [6, -3] }]; // I want all the old geolocation rows to be deleted here
await manager.save(business);
```

vieweg:
I have a same question!

majorhayes:
I'm trying to do the same thing.

savolkov:
#1286

ml1nk:
Related: #2467

tobiaskraus:
[...]

Issue #1286, opened 2017-12-01
Delete Referenced Record instead of setting fk reference to null on OneToMany save


crowebird:
Setup:
```

@Entity()
export Class EntityA extends BaseEntity {
  @PrimaryGeneratedColumn()
  EntityAId: number;

  @Column()
  SomeCol: string;

  @OneToMany(type => EntityB, entity => entity.EntityA, {
    cascadeInsert: true
  })
  EntityBs: EntityB[];
}

@Entity()
export Class EntityB extends BaseEntity {
  @PrimaryGeneratedColumn()
  EntityAId: number;

  @ManyToOne(type => EntityA, (entity: EntityA) => entity.EntityBs)
  @JoinColumn({ name: 'fkEntityA' })
  EntityA: EntityA;
}

...
let entity = await EntityA.findOne({
  SomeCol: 'SomeVal'
});

entity.EntityBs = [
  EntityB.create()
];

entity.save();
```

Say I already have an EntityA in the database with some EntityBs that reference it.

If I query entityA and then change entity.EntityBs to some new array of EntityBs (and maybe update some existing), the ones that are no longer in the array I want to DELETE from the database.

Right now they only get their fkEntityA reference set to NULL.  Is this a feature I can set?

Thanks

(I have tried the cascaseremove on onDelete CASCADE option but the reference record was still just having the value set to NULL)

AlexMesser:
If you set `onDelete="CASCADE"`, the related entity is removed **only** when the owning entity is. Supplying the new array of EntityBs without records that you want to delete causes _detach_, not _delete_ of related records. It means that this records now do not belong to EntityA.

If you want to delete related records, you must manually call entityMenager.remove() on these records.

pleerock:
Looks like question is answered.

izundo-viennv:
@AlexMesser How to delete related records instead detaching. Because some cases I just update parent entity without deleting.

PR #6704, opened 2020-09-11
fix: Cannot delete entities on 1:N relationship with nullable: false

### LINKED ISSUE #1761: Cascade Remove Set Foreign Key to Null Fails When Inverse Side Is Primary

### LINKED ISSUE #2467: Cannot delete entities on 1:N relationship with nullable: false
**Issue type:**

[X] question
[?] bug report

**Database system/driver:**

[X] `mysql` / `mariadb`

**TypeORM version:**

[X] `0.2.6` (or put your version here)

My entities are like this
```
@Entity('User')
class User {
  @PrimaryColumn('varchar', { length: 120 })
  public id: string;

  @Column('varchar', { length: 120 })
  public name: string;

  @OneToMany(() => Comment, (comment: Comment) => comment.user, {
    cascade: true,
    eager: true
  })
  public comments: Comment[];
}

@Entity('Comment')
class Comment {
  public id: string;

  @Column('varchar', { length: 500 })
  public content: string;

  @ManyToOne(() => User, (user: User) => user.comments, {
    onDelete: 'CASCADE'
  })
  @JoinColumn({ name: 'userId' })
  public comment: Comment;

  @Column('datetime', { nullable: true })
  public deletedAt: Date | null;
}
** I guess it could have some typos or mistakes because it is just code sample for describing my settings
```

My problem is
- I am working with DDD, so `User` is an aggregate which contains `Comments`
- In `User` aggregate,  `User` is aggregate root and entity, comment is just value object.
- In DDD, VO must be replaced new value object when it should be modified.
- So, my job is here. I want to modify user's comment, in this time, i should create new VO as following it's change and save it. Including deletion of original comment. But I can't do that, because original comments are not deleted, just have `null` in it's foreign key. I know that I didn't set `nullable=false`, but If I set this option to false, an error occurs: `QueryFailedError: ER_BAD_NULL_ERROR: Column 'UserId' cannot be null`.

** Of course TypeOrm entities will be mapped to Domain Objects.
May they be like this:
```
class User {
	private userId: number;
	private comments: Comment[]

	...
	public modifyComment(origin: Comment) {
		comments.push(new Comment('blah blah'));
		this.removeComment(origin);
	}
}
```

- I can handle it modifying orignal object instead of creating new one and deleting it, however, it is not what I want to do. I am trying to strictly follow rules of DDD.

What I want is
- How can I remove an `comment` clearly instead of inserting `null` value to it's foreign key `userId`?
- and Why `nullable=false` is not working as I expected?

Thank you for your supporting.

### LINKED ISSUE #5645: Delete entity on OneToMany update

ml1nk:
Delete the entity if the relation id is not nullable instead of erroring. 
Solution adapted from https://github.com/typeorm/typeorm/issues/2467#issuecomment-500465793.

closes #1761
closes #2467
closes #5645

related #1286
related #1351
related #6382

### DIFF
--- a/src/persistence/subject-builder/OneToManySubjectBuilder.ts
+++ b/src/persistence/subject-builder/OneToManySubjectBuilder.ts
@@ -165,12 +165,13 @@ export class OneToManySubjectBuilder {
                 const removedRelatedEntitySubject = new Subject({
                     metadata: relation.inverseEntityMetadata,
                     parentSubject: subject,
-                    canBeUpdated: true,
+                    canBeUpdated: relation.inverseRelation!.isNullable,
+                    mustBeRemoved: !relation.inverseRelation!.isNullable,
                     identifier: removedRelatedEntityRelationId,
-                    changeMaps: [{
+                    changeMaps: relation.inverseRelation!.isNullable ? [{
                         relation: relation.inverseRelation!,
                         value: null
-                    }]
+                    }] : undefined
                 });
                 this.subjects.push(removedRelatedEntitySubject);
             });
--- a/test/github-issues/2467/entity/Bar.ts
+++ b/test/github-issues/2467/entity/Bar.ts
@@ -1,0 +1,14 @@
+import { Foo } from "./Foo";
+import { Entity, ManyToOne, Column, PrimaryGeneratedColumn } from "../../../../src";
+
+@Entity()
+export class Bar {
+    @PrimaryGeneratedColumn()
+    barId: number;
+
+    @ManyToOne(() => Foo, foo => foo.bars, { nullable: false })
+    foo: Foo;
+
+    @Column()
+    name: string;
+}
--- a/test/github-issues/2467/entity/Baz.ts
+++ b/test/github-issues/2467/entity/Baz.ts
@@ -1,0 +1,14 @@
+import { Foo } from "./Foo";
+import { Entity, ManyToOne, Column, PrimaryGeneratedColumn } from "../../../../src";
+
+@Entity()
+export class Baz {
+    @PrimaryGeneratedColumn()
+    bazId: number;
+
+    @ManyToOne(() => Foo, foo => foo.bazs, { nullable: true })
+    foo: Foo;
+
+    @Column()
+    name: string;
+}
--- a/test/github-issues/2467/entity/Foo.ts
+++ b/test/github-issues/2467/entity/Foo.ts
@@ -1,0 +1,21 @@
+import { Entity, OneToMany, PrimaryColumn } from "../../../../src";
+import { Bar } from "./Bar";
+import { Baz } from "./Baz";
+
+@Entity()
+export class Foo {
+    @PrimaryColumn()
+    fooId: number;
+
+    @OneToMany(type => Bar, bar => bar.foo, {
+        cascade: ["insert", "remove"],
+        eager: true
+    })
+    bars: Bar[];
+
+    @OneToMany(type => Baz, baz => baz.foo, {
+        cascade: ["insert", "remove"],
+        eager: true
+    })
+    bazs: Baz[];
+}
--- a/test/github-issues/2467/issue-2467.ts
+++ b/test/github-issues/2467/issue-2467.ts
@@ -1,0 +1,55 @@
+import "reflect-metadata";
+import {createTestingConnections, closeTestingConnections, reloadTestingDatabases} from "../../utils/test-utils";
+import {Connection} from "../../../src/connection/Connection";
+import {expect} from "chai";
+import { Foo } from "./entity/Foo";
+import { Bar } from "./entity/Bar";
+import { Baz } from "./entity/Baz";
+
+describe("github issues > #2467 Cannot delete entities on 1:N relationship with nullable: false", () => {
+
+    let connections: Connection[];
+    before(async () => connections = await createTestingConnections({
+        entities: [__dirname + "/entity/*{.js,.ts}"],
+        schemaCreate: true,
+        dropSchema: true,
+    }));
+    beforeEach(() => reloadTestingDatabases(connections));
+    after(() => closeTestingConnections(connections));
+
+    it("should delete related entity relation if not nullable", () => Promise.all(connections.map(async connection => {
+
+        const repoFoo = connection.getRepository(Foo);
+        const repoBar = connection.getRepository(Bar);
+        const repoBaz = connection.getRepository(Baz);
+
+        await repoFoo.save({ fooId: 1, bars: [{ name: "A" }, { name: "B" }],  bazs: [{ name: "C" }, { name: "D" }] });
+
+        let foo = await repoFoo.findOneOrFail();
+        let bars = await repoBar.find();
+        let bazs = await repoBaz.find();
+
+        expect(foo).to.not.be.undefined;
+        expect(foo.bars.length).to.equal(2);
+        expect(foo.bazs.length).to.equal(2);
+        expect(bars.length).to.equal(2);
+        expect(bazs.length).to.equal(2);
+
+        foo.bars = [];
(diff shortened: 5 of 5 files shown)


filipsuk:
This would be a great addition! 👏

egervari:
Will this PR be reviewed? I'm actually really interested in this feature personally - it would solve a lot of headaches.

egervari:
Can we ship this if it's good please?

TheBlue-1:
Yeah, it's a really necessary feature! Thank you for implementing it, but could you also add an option to delete the entity even if the foreign key is nullable? I think just because its nullable it shouldn't remain in a cascading relation.

ml1nk:
There ist already a merge request for this purpose #1665.

hakimio:
@ml1nk that one was closed in 2018.

ml1nk:
@hakimio That's the problem, it was rejected and next isn't merged yet. It shouldn't be hard to reopen, but until there is new input from a maintainer it doesn't seem to make any sense to me. There is more activity recently, so there is still hope.

javiermrz:
Please merge this, it's really necessary. @imnotjames I suppose you have a lot of work to do, but you self-requested a review and then ignored it. Could you please check this and merge it? All checks passed, no conflicts. There will be conflicts if this is not merged soon, and that will delay this feature even more. Thanks in advance.

tariqwebstar:
Much needed feature. Please merge it ASAP.

hakimio:
More like extremely important fix for 4 year old bug.

vicky-gonsalves:
@imnotjames Please merge this PR. It is necessary feature.

gfborges:
Also need this fix

nebkat:
There is the reopened version of the orphaned row action PR ^ (#7105).

I think the action needs to be configurable with something like `orphanedRowAction` rather than being based on whether the column is nullable. If the column is non-nullable, perhaps there was a mistake in deleting, so the approach of this PR doesn't seem right.

---

BTW, I know the lack of interaction is frustrating but spamming comments will not change that. If you care about an issue or pull request just thumbs up it. They can be sorted by reactions, so the important issues will get to the top, don't worry. It is bad enough that the backlog is so large, if these PRs are eventually reviewed the discussion should be centered around the contents of the change, not flooded with requests for merge.

pleerock:
closing in favor of #7105

PR #6707, opened 2020-09-12
feat: additional @entity options

pirouet:
New option in @Entity decorator as well as other changes related to the OracleDriver


src/cache/DbQueryResultCache.ts
```diff
@@ -29,7 +29,7 @@ export class DbQueryResultCache implements QueryResultCache {
 
         const options = <SqlServerConnectionOptions|PostgresConnectionOptions>this.connection.driver.options;
         const cacheOptions = typeof this.connection.options.cache === "object" ? this.connection.options.cache : {};
-        const cacheTableName = cacheOptions.tableName || "query-result-cache";
+        const cacheTableName = cacheOptions.tableName || "queryresultcache";
```
imnotjames, line 32: Why this change?

src/driver/DriverUtils.ts
```diff
@@ -45,8 +45,11 @@ export class DriverUtils {
      *
      * @return An alias allowing to select/transform the target `column`.
      */
-    static buildColumnAlias({ maxAliasLength }: Driver, alias: string, column: string): string {
-        const columnAliasName = alias + "_" + column;
+    static buildColumnAlias({ maxAliasLength, options }: Driver, alias: string, column: string): string {
+        let columnAliasName = alias + "_" + column;
+        if(options.type === "oracle") {
+            columnAliasName = columnAliasName.toUpperCase();
+        }
```
imnotjames, line 52: If this needs to be the case shouldn't this be done within the oracle query runner?

src/driver/oracle/OracleDriver.ts
```diff
@@ -339,15 +339,15 @@ export class OracleDriver implements Driver {
      * Escapes a column name.
      */
     escape(columnName: string): string {
-        return `"${columnName}"`;
+        return columnName;
     }
```
imnotjames, line 343: Even if this isn't technically needed, does it hurt things to have it?

test/functional/cache/custom-cache-provider.ts
```diff
@@ -29,12 +29,12 @@ describe("custom cache provider", () => {
...
       const queryResultCacheTable = queryResultCache.queryResultCacheTable;
-      expect(queryResultCacheTable).to.contain("mock");
+      expect(queryResultCacheTable.toLowerCase()).to.contain("mock");
     })));
 
     it("should cache results properly", () => Promise.all(connections.map(async connection => {
 
-        // first prepare data - insert users
+        // first prepare data - insert user_cache
         const user1 = new User();
```
imnotjames, line 38: Why are we changing the cache tests?

They don't seem related

imnotjames:
For now I'm gonna close this draft out - haven't seen much activity - if you're still working on it we can reopen it and continue the review.  There's a lot of unrelated changes, though.

src/driver/oracle/OracleDriver.ts
pirouet: Hey, 

As part of the changes needed, if I was to use a package in the FROM statement, I can't have quotes around it or it throws an ORA exception. This was a first attempt at removing the quotes (since Oracle doesn't need them) but - as discovered - it does it for more than just table names, also column, aliases, etc from what I seen. 

I am working on a better solution for this

src/driver/DriverUtils.ts
pirouet: Yeah, ideally within the runner so it's self contained, this is an early draft as I came across issues but definitely not the final solution

src/cache/DbQueryResultCache.ts
pirouet: Oracle doesn't like table names with dashes inside it as it will then throw an ORA exception. As per other comments, this is an early draft as I came across issues but definitely not the final solution

test/functional/cache/custom-cache-provider.ts
pirouet: The trouble with Oracle is that it returns results in upper case. I know the escaping would mean it comes back in lowercase/camelcase but with the other issues experienced I need to find a way to take the raw results from Oracle and match what typeorm does. 

I am looking at how I can do this specifically for OracleDriver/QueryRunner only to not affect other things. You can safely ignore these ones as I have closed this down as it was getting unwieldy and not nice (sure you would agree) and working on a better solution

PR #6714, opened 2020-09-14
Migration:generate issue with onUpdate using mariadb 10.4

jesussegado:
New versions of MariaDB return expressions in lowercase we need to set it  in uppercase so the comparison in MysqlDriver#compareExtraValues does not fail.

 example:
```sql
CREATE TABLE `test` (
  `test_id` int(11) NOT NULL AUTO_INCREMENT,
  `test_update` timestamp() NOT NULL DEFAULT current_timestamp() ON UPDATE current_timestamp(),
  PRIMARY KEY (`test_id`)
) ENGINE=InnoDB DEFAULT CHARSET=latin1
```

When you try to generate a new migration always create the next migration file:

```ts
import { MigrationInterface, QueryRunner } from 'typeorm';

export class test261600082802966 implements MigrationInterface {
  name = 'test261600082802966';

    await queryRunner.query(
      'ALTER TABLE `test` CHANGE `test_update` `test_update` timestamp() NOT NULL DEFAULT CURRENT_TIMESTAMP() ON UPDATE CURRENT_TIMESTAMP()'
    );
  }

    await queryRunner.query(
      'ALTER TABLE `test` CHANGE `test_update` `test_update` timestamp() NOT NULL DEFAULT CURRENT_TIMESTAMP() ON UPDATE current_timestamp()'
    );
  }
}
```
Entity file test.ts
```ts
import { BaseEntity, Column, Entity, PrimaryGeneratedColumn } from 'typeorm';

@Entity()
export class test extends BaseEntity {
  @PrimaryGeneratedColumn({})
  test_id: number;

  @Column({
    type: 'timestamp',
    default: () => 'CURRENT_TIMESTAMP()',
    onUpdate: 'CURRENT_TIMESTAMP()',
    nullable: false,
  })
  test_update: Date;
}
```

### DIFF
--- a/src/driver/mysql/MysqlQueryRunner.ts
+++ b/src/driver/mysql/MysqlQueryRunner.ts
@@ -1321,7 +1321,9 @@ export class MysqlQueryRunner extends BaseQueryRunner implements QueryRunner {
                     }
 
                     if (dbColumn["EXTRA"].toLowerCase().indexOf("on update") !== -1) {
-                        tableColumn.onUpdate = dbColumn["EXTRA"].substring(dbColumn["EXTRA"].toLowerCase().indexOf("on update") + 10);
+                        // New versions of MariaDB return expressions in lowercase.  We need to set it in
+                        // uppercase so the comparison in MysqlDriver#compareExtraValues does not fail.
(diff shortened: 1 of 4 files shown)


pleerock:
Thank you! It will be useful if you'll add some test, so we can make sure in future this functionality won't break.

jesussegado:
I just added a test to issue 6714

pleerock:
Thank you!

PR #6724, opened 2020-09-16
chore: explicitly pull in typescript 3.9

imnotjames:
this bumps the version of typescript accepted to >=3.9.7,<4.0
we were implicitly pulling in that version (per the package-lock.json)
so this just codifies it in the package.json as well


pleerock:
Thank you!

PR #6742, opened 2020-09-18
fix: Improve simple-enum handling in migrations

jlhogan:
Use an explicit name when creating simple-enum CHECK constraints so that
they can be ignored when finding old (to be deleted) CHECK constraints.
Additionally, update query runners for simple-enum-compatible types to
maintain true DB type ('varchar'/'nvarchar') as tableColumn.type, as opposed
to setting it to 'simple-enum' in `loadTables`. This fixes broken
`findChangedColumn` in simple-enum-compatible drivers, which check
tableColumn.type against normalized columnMetadata.type. According to
in-line documentation, `normalizeType` creates a DB type from a metadata
type, implying that tableColumn.type is expected to be a true DB type,
not a metadata type like 'simple-enum'.


imnotjames:
Tests have failed - can you look into that?

Also, is it possible to add tests to confirm that the behavior was fixed?

imnotjames:

PR #6745, opened 2020-09-19
fix: count() method for multiple primary keys for cockroachdb

temnov98:
CockroachDB does not support concat() for different types at the moment.
To fix this problem, each primary key is cast to the text type.

An example of a generated SQl query before the error is fixed:
```
SELECT COUNT(DISTINCT(CONCAT("Test"."varcharField", "Test"."uuidField", "Test"."intField"))) as "cnt" 
FROM "tests" "Test
```

An error that occurs while executing this SQL query:
```QueryFailedError: count(): unknown signature: concat(varchar, uuid, int)```

Sample SQL query after fixing the error:
```
SELECT COUNT(DISTINCT(CONCAT("Test"."varcharField"::text, "Test"."uuidField"::text, "Test"."intField"::text))) as "cnt" 
FROM "tests" "Test"
```

### DIFF
--- a/src/query-builder/SelectQueryBuilder.ts
+++ b/src/query-builder/SelectQueryBuilder.ts
@@ -1793,6 +1793,11 @@ export class SelectQueryBuilder<Entity> extends QueryBuilder<Entity> implements 
                     const propertyName = this.escape(primaryColumn.databaseName);
                     return `${distinctAlias}.${propertyName}::text`;
                 }).join(", ") + "))) as \"cnt\"";
+            } else if (this.connection.driver instanceof OracleDriver) {
+                countSql = `COUNT(DISTINCT(` + metadata.primaryColumns.map((primaryColumn, index) => {
+                    const propertyName = this.escape(primaryColumn.databaseName);
+                    return `${distinctAlias}.${propertyName}`;
+                }).join(" || ") + ")) as \"cnt\"";
             } else {
                 countSql = `COUNT(DISTINCT(CONCAT(` + metadata.primaryColumns.map((primaryColumn, index) => {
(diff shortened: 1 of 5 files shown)


temnov98:
I fixed my lint errors in second commit. But failed checks (ci/circleci: lint) not related to my changes now.

imnotjames:
Looks like it's here - https://github.com/typeorm/typeorm/pull/6745/files#diff-7da0fc32ecaf4de82fdafaaac5d5c9aeR39

imnotjames:
Is it possible for the test to live somewhere under `test/functional/query-builder` & apply to more than just cockroach DB?  Eg, run for all drivers?

That way we can ensure this functionality works across everything & this change doesn't negatively impact other drivers.  I mean, it shouldn't 'cause it applies explicitly for that driver but yknow.. good to check. :)

imnotjames:
Looks like it sailed for Oracle.  Can you update the test to check if it's oracle & return if so?  that way we can come back in afterwards and fix the Oracle implementation & enable it again later.

Unless you want to fix this case with be oracle as well :)

temnov98:
> Oof - this case fails with oracle. Pretty sure that's unrelated to your changes.
> 
> Can you update the test to check if it's oracle & return if so? that way we can come back in afterwards and fix the Oracle implementation & enable it again later.
> 
> Unless you want to fix this case with be oracle as well :)

Thank you for your help! 

Yes, I noticed that there was an error in oracle. The problem arose due to the fact that oracle does not support ```CONCAT()``` for more than 2 arguments. 

To fix this problem, I replaced ```CONCAT()``` with ```||``` operator.

pleerock:
Thank you!

PR #6746, opened 2020-09-20
refactor: drop promise utils

imnotjames:
this drops the usage of `PromiseUtils` and replaces it with either built in `Promise` functions
or analogous (& clearer?) standard Typescript - such as `for (let f of funcs) { await f(); }` instead
of `Promise.runInSequence(funcs, f => f());`

In general, switching from `Promise.runInSequence` to `Promise.all` in will bring those test cases
in line with every other test case that runs in parallel rather than sequentially.  a handful of
test cases need to remain sequential & have had comments added to them to notate why that is.

this also drops the export of `PromiseUtils`.  as far as I can tell this should not have been
exported in the first place and is an undocumented private API export from TypeORM.


imnotjames:
Story of why I did this - I was testing some other things and ran into weirdness with test for github issue 1055 - which led me to look at PromiseUtils & the usage of it - then wonder what it was doing & why we even needed it.

pleerock:
> for (let f of funcs) { await f(); } 

this syntax is definitely more clear than PromiseUtils. runInSequence

> In general, switching from Promise.runInSequence to Promise.all in tests will accelerate
those test cases by running the varying test cases against different connections in parallel
rather than sequentially. this does not seem to affect the tests outcomes.

Promise.runInSequence is great in tests because it simplifies debugging, especially when logging is enabled (otherwise different sqls from different dbs are mixed and its impossible to understand the SQL flow and you have to enable each driver and check them in isolation). When tests are being running in . So, my question is - how faster tests become? (in percentage)

pleerock:
Also regarding to your changes on other methods:

* I'm not sure if removing `.extractValue` won't bring any problems. There might be problems when using lazy relations, cases when we need to compare values, but one of the value is a promise (with value inside).

* regarding to issue-1055, if I'm not wrong its not possible to change this find: `await manager.findOne(Child, { parent: PromiseUtils.create(loadedParent) })` to `await manager.findOne(Child, { parent: loadedParent })` because of type mismatch, since "parent" is a lazy relation.

* I can't comment on `.settle` because I don't really remember why exactly we did it so

pleerock:
And one another suggestion.. is to try to avoid big changes because we would like to merge and release a `next` branch first. (to prevent big conflicts)

imnotjames:
Tests aren't that much faster - it's negligible & I shouldn't have added it as a reason for the change.  If you'd like, I can update the tests to run sequentially anywhere we were using `PromieUtils.runInSequence` - but instead I had updated them to use `Promise.all` like most of the other tests we have.

> I'm not sure if removing .extractValue won't bring any problems. There might be problems when using lazy relations, cases when we need to compare values, but one of the value is a promise (with value inside).

`extractValue` does nothing for normal promises - only promises created using `PromiseUtils.create` which was only used in `issue-1055` - nowhere else in the code or documentation.  Normal promises are just returned back as promises - which includes our lazy relations.

> regarding to issue-1055, if I'm not wrong its not possible to change this find: await manager.findOne(Child, { parent: PromiseUtils.create(loadedParent) }) to await manager.findOne(Child, { parent: loadedParent }) because of type mismatch, since "parent" is a lazy relation.

Unfortunately, no - the test then fails because `findOne` does not understand how to unpack a real promise.  This was an issue even before any of these other changes.

> And one another suggestion.. is to try to avoid big changes because we would like to merge and release a next branch first. (to prevent big conflicts)

Agreed - this got a little bigger than I had wanted it to be.  Perhaps I can be culling this to just updating tests & a separate PR update  test 1055 and remove the use of `extractValue`?

Hopefully we can get `master` merged back into `next` & could then check what conflicts this would create?

pleerock:
> If you'd like, I can update the tests to run sequentially anywhere we were using PromieUtils.runInSequence - but instead I had updated them to use Promise.all like most of the other tests we have.

okay, if we have them mixed already, lets keep go with `Promise.all`

> extractValue does nothing for normal promises - only promises created using PromiseUtils.create which was only used in issue-1055 - nowhere else in the code or documentation. Normal promises are just returned back as promises - which includes our lazy relations.

> Unfortunately, no - the test then fails because findOne does not understand how to unpack a real promise. This was an issue even before any of these other changes.

this is weird. Maybe we had something that uses "__value__" before... okay let's merge it, but we need to keep track on new bugs on lazy loading

> Hopefully we can get master merged back into next & could then check what conflicts this would create?

yes, I mean we can continue to merge things into `master`, however more we merge, harder it would be for us to merge master into the next. At some point we need to focus on merging master into the next and versa.

PR #6747, opened 2020-09-20
techdebt: create a ReplicationMode type and update function defs

imnotjames:
create a type to track ReplicationMode instead of writing out
`"master"|"slave"` everywhere.

update to drop the default from the QueryRunner constructor
as they will always receive the mode from the driver when
it's part of the QueryRunner

also drop the default from Driver.createQueryRunner in the
implementations - the interface mandates the mode to be defined
so it will never be omitted anyway

also drop the explict "master" from any connection.createQueryRunner
calls so we leave it either the default or `slave` when needed

all of this makes it easier to eventually migrate to
other naming convetions for the replication mode
should it be deemed the right path to go down

### DIFF
--- a/src/cache/DbQueryResultCache.ts
+++ b/src/cache/DbQueryResultCache.ts
@@ -229,7 +229,7 @@ export class DbQueryResultCache implements QueryResultCache {
         if (queryRunner)
             return queryRunner;
 
-        return this.connection.createQueryRunner("master");
+        return this.connection.createQueryRunner();
     }
 
 }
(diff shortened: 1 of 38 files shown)


imnotjames:
This is a chunk of #6324 which doesn't change any of the actual behavior or values - instead just abstracts them a bit :)

This means a change to implement something like #6324 would be much smaller in scope and it'd be easier to tell if it's safe to merge in or not :tada:

pleerock:
We also need to re-export newly created type in main `index.ts` for other people to use.

imnotjames:
Sure, I can do that.  Only reason I didn't was because other driver types such as `IsolationLevel` hadn't been exported.

pleerock:
> Sure, I can do that. Only reason I didn't was because other driver types such as IsolationLevel hadn't been exported.

sometimes we forget to add things to `index`. `IsolationLevel` was another contributed type that we forgot to add. Thank you!

PR #6759, opened 2020-09-22
Fixed bug behavior

Tequelequeteque:
Behavior equal create migration

### DIFF
--- a/src/commands/EntityCreateCommand.ts
+++ b/src/commands/EntityCreateCommand.ts
@@ -51,13 +51,9 @@ export class EntityCreateCommand implements yargs.CommandModule {
                 } catch (err) { }
             }
 
-            const path = process.cwd() + "/" + (directory ? (directory + "/") : "") + filename;
-            const fileExists = await CommandUtils.fileExists(path);
-            if (fileExists) {
-                throw `File ${chalk.blue(path)} already exists`;
-            }
+              const path = (directory.startsWith("/") ? "" : process.cwd() + "/") + (directory ? (directory + "/") : "") + filename;
             await CommandUtils.createFile(path, fileContent);
-            console.log(chalk.green(`Entity ${chalk.blue(path)} has been created successfully.`));
+            console.log(`Migration ${chalk.blue(path)} has been generated successfully.`);
 
         } catch (err) {
             console.log(chalk.black.bgRed("Error during entity creation:"));


imnotjames:
Can you clarify what the bug is?

src/commands/EntityCreateCommand.ts
```diff
@@ -51,13 +51,9 @@ export class EntityCreateCommand implements yargs.CommandModule {
                 } catch (err) { }
             }
 
-            const path = process.cwd() + "/" + (directory ? (directory + "/") : "") + filename;
-            const fileExists = await CommandUtils.fileExists(path);
-            if (fileExists) {
-                throw `File ${chalk.blue(path)} already exists`;
-            }
+              const path = (directory.startsWith("/") ? "" : process.cwd() + "/") + (directory ? (directory + "/") : "") + filename;
```
imnotjames, line 54: It kinda looks to me that you're trying to handle absolute paths.

In that case, perhaps look at `path.resolve`?  Something like `path.resolve(directory, filename)` would get you probably much of the behavior to handle both relative and absolute paths.

src/commands/EntityCreateCommand.ts
```diff
@@ -51,13 +51,9 @@ export class EntityCreateCommand implements yargs.CommandModule {
...
             await CommandUtils.createFile(path, fileContent);
-            console.log(chalk.green(`Entity ${chalk.blue(path)} has been created successfully.`));
+            console.log(`Migration ${chalk.blue(path)} has been generated successfully.`);
```
imnotjames, line 56: Was this change accidental?

Tequelequeteque:
I pass absolute path on cli migrateDir and directory create ok but in entityDir he create all subdir in src ...
example :
I pass /user/myuser/dir/src/entity
so he create /user/myuser/dir/src//user/myuser/dir/src/entity

Tequelequeteque:
```ts
import { resolve, join } from 'path';
import 'dotenv/config';
import { PostgresConnectionOptions } from 'typeorm/driver/postgres/PostgresConnectionOptions';

interface IENV{
  [key:string]: any,
}

const {
  DB_TYPE,
  DB_PORT,
  DB_USER,
  DB_PASS,
  DB_NAME,
} = process.env as unknown as IENV;

const entitiesPath = resolve(__dirname, '..', 'entities');
const migrationsPath = resolve(__dirname, 'migrations');

// behavior diferent
const entitiesCliPath = entitiesPath.split('/').slice(-2).join('/');

const dbConfig = {
  synchronize: true,
  name: 'default',
  type: DB_TYPE,
  host: 'localhost',
  port: DB_PORT,
  username: DB_USER,
  password: DB_PASS,
  database: DB_NAME,
  migrationsRun: true,
  entities: [entitiesPath.concat('/**/*ts')],
  migrations: [migrationsPath.concat('/**/*ts')],
  cli: {
    entitiesDir: entitiesCliPath,
    migrationsDir: migrationsPath,
  },
  dropSchema: true,
  logging: true,
} as PostgresConnectionOptions;

export { dbConfig };

```

src/commands/EntityCreateCommand.ts
Tequelequeteque: I just copy code from MigrationCreateCommand...
Maybe delegate task create file =T

src/commands/EntityCreateCommand.ts
Tequelequeteque: I just copy code from MigrationCreateCommand...
Maybe delegate task create file =T

imnotjames:
Given that this breaks the build, I'll be closing this in favor of #6807

PR #6763, opened 2020-09-23
fix: add dummy for FileLogger, ConnectionOptionsReaders, and update gulpfile

### LINKED ISSUE #6739: 0.2.26 brakes browser builds (by node-only PlatformTools refactoring commit)
**Issue type:**
[x] bug report

**Database system/driver:**
all / browser-builds

**TypeORM version:**
[x] `0.2.26`

Building a project with TypeORM 0.2.26 with webpack for a browser target leads to multiple errors, the following is just just an excerpt (with duplicates removed):
```
WARNING in ../node_modules/typeorm/browser/connection/ConnectionOptionsReader.js 114:45-64
Critical dependency: the request of a dependency is an expression

WARNING in ../node_modules/typeorm/browser/util/DirectoryExportedClassesLoader.js 38:38-78
Critical dependency: the request of a dependency is an expression

Module not found: Error: Can't resolve 'path'

Module not found: Error: Can't resolve 'stream'

Module not found: Error: Can't resolve 'fs'

Module not found: Error: Can't resolve 'timers'
```

I could track this down to this commit:
https://github.com/typeorm/typeorm/commit/490ad0d424e674de5be38b0c3bc83b9a14a383a5

I reverted this commit locally and could then successfully use 0.2.26 with this modification in a browser Webpack build. 

I guess that https://github.com/typeorm/typeorm/pull/6630 didn't take browser builds into account for its refactoring. On a quick look those errors (and the code changes in the pull request) do look like node-only stuff.

I suggest reverting the commit/PR @pleerock.

imnotjames:
Add browser dummies for `FileLogger`, `ConnectionOptionsReader`, and change how we process the `*.template` files in `gulp`.

Ideally, we should move away from this whole process for creating browser builds by monkey-patching random parts of TypeORM in this way - such as restructuring the package or finding some other mechanism to make this happen.. but until then, this solves some of the remaining issues with browser builds according to @chriswep 

Fixes #6739

### DIFF
--- a/gulpfile.ts
+++ b/gulpfile.ts
@@ -64,38 +64,18 @@ export class Gulpfile {
             "!./src/commands/*.ts",
             "!./src/cli.ts",
             "!./src/typeorm.ts",
-            "!./src/typeorm-model-shim.ts",
-            "!./src/platform/PlatformTools.ts"
+            "!./src/typeorm-model-shim.ts"
         ])
         .pipe(gulp.dest("./build/browser/src"));
     }
 
     /**
-     * Replaces PlatformTools with browser-specific implementation called BrowserPlatformTools.
+     * Copies templates for compilation
      */
     @Task()
-    browserCopyDirectoryExportedClassesLoader() {
-        return gulp.src("./src/platform/BrowserDirectoryExportedClassesLoader.template")
-            .pipe(rename("BrowserDirectoryExportedClassesLoader.ts"))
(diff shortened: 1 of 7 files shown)


imnotjames:
We still don't have a way to automatically test this kind of thing - technically it wasn't preventing folks from building TypeORM for browsers but it was making the build they'd have to make larger than desirable.

PR #6780, opened 2020-09-24
Fix schema synchronization for partitioned tables with PostgreSQL 12+

nlenepveu:
### Consider partitioned tables as regular tables

Typeorm currently excludes partitioned tables from schema synchronization. We could also take them into account as their behavior regarding schema modifications is the same as regular tables.

### Ignore foreign keys on partitions of the partitioned tables when syncing schema on PostgreSQL 12+

When a table has foreign keys referencing a partitioned table, the `PostgresQueryRunner` will load the constraints from every partition.
We need to ignore those foreign keys as they are not managed by the user.

Example:
`table_A` has a foreign key constraint referencing `table_B` which is a partitioned table on `key1` and `key2`.

The command `typeorm schema:log` will return this kind of list of queries to remove the FK constraints from `table_A`.
```
------------------------------------------------------------------
-- Schema syncronization will execute following sql queries (220):
------------------------------------------------------------------
ALTER TABLE "table_A" DROP CONSTRAINT "table_A_table_B_key1_key2__fkey";
ALTER TABLE "table_A" DROP CONSTRAINT "table_A_table_B_key1_key2_fkey1";
ALTER TABLE "table_A" DROP CONSTRAINT "table_A_table_B_key1_key2_fkey2";
ALTER TABLE "table_A" DROP CONSTRAINT "table_A_table_B_key1_key2_fkey3";
ALTER TABLE "table_A" DROP CONSTRAINT "table_A_table_B_key1_key2_fkey4";
ALTER TABLE "table_A" DROP CONSTRAINT "table_A_table_B_key1_key2_fkey5";
ALTER TABLE "table_A" DROP CONSTRAINT "table_A_table_B_key1_key2_fkey6";
ALTER TABLE "table_A" DROP CONSTRAINT "table_A_table_B_key1_key2_fkey7";
ALTER TABLE "table_A" DROP CONSTRAINT "table_A_table_B_key1_key2_fkey8";
ALTER TABLE "table_A" DROP CONSTRAINT "table_A_table_B_key1_key2_fkey9";
ALTER TABLE "table_A" DROP CONSTRAINT "table_A_table_B_key1_key2_fkey10";
ALTER TABLE "table_A" DROP CONSTRAINT "table_A_table_B_key1_key2_fkey11";
ALTER TABLE "table_A" DROP CONSTRAINT "table_A_table_B_key1_key2_fkey12";
ALTER TABLE "table_A" DROP CONSTRAINT "table_A_table_B_key1_key2_fkey13";
ALTER TABLE "table_A" DROP CONSTRAINT "table_A_table_B_key1_key2_fkey14";
ALTER TABLE "table_A" DROP CONSTRAINT "table_A_table_B_key1_key2_fkey15";
ALTER TABLE "table_A" DROP CONSTRAINT "table_A_table_B_key1_key2_fkey16";
ALTER TABLE "table_A" DROP CONSTRAINT "table_A_table_B_key1_key2_fkey17";
ALTER TABLE "table_A" DROP CONSTRAINT "table_A_table_B_key1_key2_fkey18";
ALTER TABLE "table_A" DROP CONSTRAINT "table_A_table_B_key1_key2_fkey19";
ALTER TABLE "table_A" DROP CONSTRAINT "table_A_table_B_key1_key2_fkey20";
ALTER TABLE "table_A" DROP CONSTRAINT "table_A_table_B_key1_key2_fkey21";
ALTER TABLE "table_A" DROP CONSTRAINT "table_A_table_B_key1_key2_fkey22";
ALTER TABLE "table_A" DROP CONSTRAINT "table_A_table_B_key1_key2_fkey23";
ALTER TABLE "table_A" DROP CONSTRAINT "table_A_table_B_key1_key2_fkey24";
ALTER TABLE "table_A" DROP CONSTRAINT "table_A_table_B_key1_key2_fkey25";
...
```

Typeorm should ignore those constraints.

### DIFF
--- a/src/driver/postgres/PostgresQueryRunner.ts
+++ b/src/driver/postgres/PostgresQueryRunner.ts
@@ -1434,6 +1434,10 @@ export class PostgresQueryRunner extends BaseQueryRunner implements QueryRunner 
             }
             return `("ns"."nspname" = '${schema}' AND "cl"."relname" = '${name}')`;
         }).join(" OR ");
+
+        const hasRelispartitionColumn = await this.hasSupportForPartitionedTables();
+        const isPartitionCondition = hasRelispartitionColumn ? ` AND "cl"."relispartition" = 'f'` : "";
+
         const foreignKeysSql = `SELECT "con"."conname" AS "constraint_name", "con"."nspname" AS "table_schema", "con"."relname" AS "table_name", "att2"."attname" AS "column_name", ` +
             `"ns"."nspname" AS "referenced_table_schema", "cl"."relname" AS "referenced_table_name", "att"."attname" AS "referenced_column_name", "con"."confdeltype" AS "on_delete", ` +
             `"con"."confupdtype" AS "on_update", "con"."condeferrable" AS "deferrable", "con"."condeferred" AS "deferred" ` +
@@ -1449,7 +1453,7 @@ export class PostgresQueryRunner extends BaseQueryRunner implements QueryRunner 
             `WHERE "con1"."contype" = 'f' AND (${foreignKeysCondition}) ` +
             `) "con" ` +
             `INNER JOIN "pg_attribute" "att" ON "att"."attrelid" = "con"."confrelid" AND "att"."attnum" = "con"."child" ` +
-            `INNER JOIN "pg_class" "cl" ON "cl"."oid" = "con"."confrelid" ` +
+            `INNER JOIN "pg_class" "cl" ON "cl"."oid" = "con"."confrelid" ${isPartitionCondition}` +
             `INNER JOIN "pg_namespace" "ns" ON "cl"."relnamespace" = "ns"."oid" ` +
             `INNER JOIN "pg_attribute" "att2" ON "att2"."attrelid" = "con"."conrelid" AND "att2"."attnum" = "con"."parent"`;
         const [dbTables, dbColumns, dbConstraints, dbIndices, dbForeignKeys]: ObjectLiteral[][] = await Promise.all([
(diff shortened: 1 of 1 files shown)


src/driver/postgres/PostgresQueryRunner.ts
```diff
@@ -1434,6 +1434,10 @@ export class PostgresQueryRunner extends BaseQueryRunner implements QueryRunner
             }
             return `("ns"."nspname" = '${schema}' AND "cl"."relname" = '${name}')`;
         }).join(" OR ");
+
+        const hasIsPartition = await this.hasPartitionedTables();
+        const isPartitionCondition = hasIsPartition ? ` AND "cl"."relispartition" = 'f'` : "";
```
pleerock, line 1439: what if we wouldn't have `hasIsPartition` check here? according to this check it looks like we do need items with "AND "cl"."relispartition" = 't' ?

nlenepveu: @pleerock the column `relispartition` does not exist in PostgreSQL versions that have no support for partitioned tables (prior to version 10). This is why we need this check here.
Then, when using a version of PostgreSQL with partitioned tables, we need to ignore the partitions and only check the constraints on the "main" table ie: `"relispartition" = "f"` (which are not partitions).

nlenepveu: Maybe we should rename `hasPartitionedTables` to `hasSupportForPartitionedTables` to make it clearer?

nlenepveu:
@pleerock let me know if anything is unclear

pleerock:
Looks good, thanks

prevostc:
Hi there sorry to interrupt, this seem to be the only issue related to pg partitioning.
I couldn't find in the documentation how to add `PARTITION BY RANGE (x,y)` to the table definition.

doominkim:
Don't you have any plans to officially support partitioning in typeorm?

PR #6785, opened 2020-09-25
techdebt: move aurora postgres driver to the right dir

imnotjames:
for some reason the aurora postgres driver was living with the postgres
driver - instead of in the directory explicitly for the aurora postgres
driver.

this moves the aurora-pg driver into that directory so there's a clearer
delineation between the two drivers


germangamboa95 (approved):

pleerock:
Thank you!

PR #6796, opened 2020-09-28
fix: set the search path for postgres when schema is set

### LINKED ISSUE #6734: `schema` option in postgres driver is not doing anything
[x] bug report
[ ] feature request
[x] documentation issue

**Database system/driver:**

[ ] `cordova`
[ ] `mongodb`
[ ] `mssql`
[ ] `mysql` / `mariadb`
[ ] `oracle`
[x] `postgres`
[ ] `cockroachdb`
[ ] `sqlite`
[ ] `sqljs`
[ ] `react-native`
[ ] `expo`

**TypeORM version:**

[x] `latest`
[ ] `@next`
[x] `0.2.26`

The [documented `schema` parameter for postgres](https://typeorm.io/#/connection-options/postgres--cockroachdb-connection-options) does not seem to be doing anything. There was a [PR for adding a default schema in the past](https://github.com/typeorm/typeorm/pull/160), though under the `schemaName` property, but it seems this feature was removed completely.

All that remains is a `schema` option defined in the postgres driver, which does not have any effect it seems.

imnotjames:
when the schema option is set for postgres we used to set
the search path - to much rejoice - but we no longer do
because it was accidentally removed

however, as of this change, it'll be back

fixes #6734

### DIFF
--- a/src/driver/postgres/PostgresDriver.ts
+++ b/src/driver/postgres/PostgresDriver.ts
@@ -818,11 +818,7 @@ export class PostgresDriver implements Driver {
      * If replication is not setup then returns default connection's database connection.
      */
     obtainMasterConnection(): Promise<any> {
-        return new Promise((ok, fail) => {
-            this.master.connect((err: any, connection: any, release: any) => {
-                err ? fail(err) : ok([connection, release]);
-            });
-        });
+        return this.obtainPoolConnection(this.master);
     }
 
     /**
@@ -834,12 +830,8 @@ export class PostgresDriver implements Driver {
         if (!this.slaves.length)
             return this.obtainMasterConnection();
 
-        return new Promise((ok, fail) => {
-            const random = Math.floor(Math.random() * this.slaves.length);
-            this.slaves[random].connect((err: any, connection: any, release: any) => {
-                err ? fail(err) : ok([connection, release]);
-            });
-        });
+        const random = Math.floor(Math.random() * this.slaves.length);
+        return this.obtainPoolConnection(this.slaves[random]);
     }
 
     /**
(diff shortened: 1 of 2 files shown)


pleerock:
Sorry but I think I it's better to left as it is.

It was 3 years ago when change was made by @AlexMesser and I do remember we had an offline discussion about these changes and he removed default schema set because it bring problems with multi-schema setups. I would like to avoid breaking changes where I'm not completely sure, but here I'm sure this change was intended.

In #6734 author claims that schema option is not doing anything, but it's not completely true. It is used where it should be used. We don't take a responsibility on raw queries like `.query("SELECT FROM database.table")` - this is true not only for schema, but also for database. We also try to avoid changing database settings, including session settings. In his case, if he wants a schema to be default for everything - he can just apply change to a database configuration, it is more rational decision, rather than executing and applying setting on each database connection obtained from the pool.

imnotjames:
I can think of a few cases where this would make for a clearer API for users - but I'm going to spend a bit of time thinking this through, I'll do some research, and come back to this one shortly. :)

imnotjames:
> Sorry but I think I it's better to left as it is.
> 
> It was 3 years ago when change was made by @AlexMesser and I do remember we had an offline discussion about these changes and he removed default schema set because it bring problems with multi-schema setups. I would like to avoid breaking changes where I'm not completely sure, but here I'm sure this change was intended.
> 
> In #6734 author claims that schema option is not doing anything, but it's not completely true. It is used where it should be used. We don't take a responsibility on raw queries like `.query("SELECT FROM database.table")` - this is true not only for schema, but also for database. We also try to avoid changing database settings, including session settings. In his case, if he wants a schema to be default for everything - he can just apply change to a database configuration, it is more rational decision, rather than executing and applying setting on each database connection obtained from the pool.



> Sorry but I think I it's better to left as it is.
> 
> It was 3 years ago when change was made by @AlexMesser and I do remember we had an offline discussion about these changes and he removed default schema set because it bring problems with multi-schema setups. I would like to avoid breaking changes where I'm not completely sure, but here I'm sure this change was intended.
> 
> In #6734 author claims that schema option is not doing anything, but it's not completely true. It is used where it should be used. We don't take a responsibility on raw queries like `.query("SELECT FROM database.table")` - this is true not only for schema, but also for database. We also try to avoid changing database settings, including session settings. In his case, if he wants a schema to be default for everything - he can just apply change to a database configuration, it is more rational decision, rather than executing and applying setting on each database connection obtained from the pool.

I disagree with this - mostly because of the location of the option.  If this only applies to query builder code / schema builder code, it should be denoted as such in the configuration.  As it stands, it reads as a default that gets set similar to `database` for all connections.

However, this behavior is across postgres, cockroach, SAP, and SQL Server. So be it.

Issue #6378, opened 2020-07-09
ormconfig.js doesn't work with two set of options in it


denissabramovs:
Config for .js is not working with two configs in it as with json example above it.
https://github.com/typeorm/typeorm/blob/master/docs/using-ormconfig.md#using-ormconfigjs
So basically this is not working:
```
module.exports = [
  {
    "name": "admin-api",
    "type": process.env.TYPEORM_CONNECTION,
    "host": process.env.TYPEORM_HOST,
    "port": process.env.TYPEORM_PORT,
    "username": process.env.TYPEORM_USERNAME,
    "password": process.env.TYPEORM_PASSWORD,
    "database": process.env.TYPEORM_DATABASE,
    "migrationsTableName": process.env.TYPEORM_ADMIN_API_MIGRATIONS_TABLE_NAME,
    "migrations": [
      process.env.TYPEORM_ADMIN_API_MIGRATIONS_DIR
    ]
  },
  {
    "name": "tenants-api",
    "type": process.env.TYPEORM_CONNECTION,
    "host": process.env.TYPEORM_HOST,
    "port": process.env.TYPEORM_PORT,
    "username": process.env.TYPEORM_USERNAME,
    "password": process.env.TYPEORM_PASSWORD,
    "database": process.env.TYPEORM_DATABASE,
    "migrationsTableName": process.env.TYPEORM_TENANTS_API_MIGRATIONS_TABLE_NAME,
    "migrations": [
      process.env.TYPEORM_TENANTS_API_MIGRATIONS_DIR
    ]
  }
];
```

**Issue type:**

[ ] question
[ x] bug report
[ ] feature request
[ ] documentation issue

**Database system/driver:**

[ ] `cordova`
[ ] `mongodb`
[ ] `mssql`
[ ] `mysql` / `mariadb`
[ ] `oracle`
[ x] `postgres`
[ ] `cockroachdb`
[ ] `sqlite`
[ ] `sqljs`
[ ] `react-native`
[ ] `expo`

**TypeORM version:**

[ ] `latest`
[ ] `@next`
[^0.2.25] `0.x.x` (or put your version here)

**Steps to reproduce or a small repository showing the problem:**

<!--
    To answer those questions you need to put "x" inside the square brackets, for example:
    [] `mysql`
    [x] `postgres`

    Also, please format your code properly (by taking code blocks into ```ts .... ```)
--!>

denissabramovs:
[...]

PR #6812, opened 2020-09-30
fix: Cannot find connection ${name} because its not defined in any orm configuration files.

II-alex-II:
issue:
image

emulate issue:
image

see https://github.com/typeorm/typeorm/issues/6378#issue-654213226

### DIFF
--- a/src/connection/ConnectionOptionsReader.ts
+++ b/src/connection/ConnectionOptionsReader.ts
@@ -11,6 +11,11 @@ import {ConnectionOptionsXmlReader} from "./options-reader/ConnectionOptionsXmlR
  */
 export class ConnectionOptionsReader {
 
+    static findOptionByName(options: ConnectionOptions[], name = "default") {
+        const [targetOptions] = options.filter(options => options && options.name === name || (name === "default" && options && !options.name));
+        return targetOptions;
+    }
+
     // -------------------------------------------------------------------------
     // Constructor
     // -------------------------------------------------------------------------
@@ -50,7 +55,7 @@ export class ConnectionOptionsReader {
      */
     async get(name: string): Promise<ConnectionOptions> {
         const allOptions = await this.all();
-        const targetOptions = allOptions.find(options => options.name === name || (name === "default" && !options.name));
+        const targetOptions = ConnectionOptionsReader.findOptionByName(allOptions, name);
         if (!targetOptions)
             throw new Error(`Cannot find connection ${name} because its not defined in any orm configuration files.`);
 
@@ -65,7 +70,7 @@ export class ConnectionOptionsReader {
         if (!allOptions)
             return false;
 
-        const targetOptions = allOptions.find(options => options.name === name || (name === "default" && !options.name));
+        const targetOptions = ConnectionOptionsReader.findOptionByName(allOptions, name);
         return !!targetOptions;
     }
 


imnotjames:
Wouldn't this just hide the issue in question?

src/connection/ConnectionOptionsReader.ts
```diff
@@ -50,7 +55,7 @@ export class ConnectionOptionsReader {
      */
     async get(name: string): Promise<ConnectionOptions> {
         const allOptions = await this.all();
-        const targetOptions = allOptions.find(options => options.name === name || (name === "default" && !options.name));
```
imnotjames, line 53: Can we instead update & add the `options` check here to verify that `options` is truthy instead of create a new static function?

imnotjames:
Ok - this is completely unrelated to #6378 

It looks like you've passed invalid data into the the connection options & this is trying to make the code for connection options reader a bit more defensive as we can't really type it.  I can dig that.

We probably would want to do `typeof options === 'object'` though to be more explicit in what we're looking for, wouldn't we?

src/connection/ConnectionOptionsReader.ts
```diff
@@ -11,6 +11,11 @@ import {ConnectionOptionsXmlReader} from "./options-reader/ConnectionOptionsXmlR
  */
 export class ConnectionOptionsReader {
 
+    static findOptionByName(options: ConnectionOptions[], name = "default") {
```
imnotjames, line 14: Can we keep the API for `ConnectionOptionsReader` leaner?  I don't think we need to add public helper methods onto the API.

src/connection/ConnectionOptionsReader.ts
```diff
@@ -11,6 +11,11 @@ import {ConnectionOptionsXmlReader} from "./options-reader/ConnectionOptionsXmlR
...
+        const [targetOptions] = options.filter(options => options && options.name === name || (name === "default" && options && !options.name));
```
imnotjames, line 15: Why use destructuring + filter instead of `find`?

II-alex-II: @imnotjames filter does not throw error
image

II-alex-II:
We had correctly settings 
image

I updated after the error the settings
image

src/connection/ConnectionOptionsReader.ts
II-alex-II: yep, need will create a pure function

src/connection/ConnectionOptionsReader.ts
imnotjames: Right - filter skips over the empty item you've entered into the array but I don't really see that as being a huge benefit when you're updating to defend against an undefined value anyway.  The code in your screenshot is different from the code you've entered in this PR.

II-alex-II:
I think  function normalize does not correctly work
image

II-alex-II:
but I didn't understand why I had targetOption got undefined

image

imnotjames:
Function normalize doesn't actually return an array of options because during runtime the type system for Typescript is "best effort" - most of the typing happens during the compilation phase using static analysis.

I think that you're passing in something that's undefined still - too many commas or something like that.  Your `ormconfig.js`  is truncated - can you paste it instead of taking screenshots?  Please remove the passwords, of course, but I am pretty sure there's something that's not as expected in the file.

II-alex-II:
ormconfig.js
`require('dotenv/config');
const { DB_PORT, DB_USERNAME, DB_PASSWORD, DB_DATABASE, DB_HOST } = process.env;

module.exports = [
  {
    name: 'production',
    type: 'postgres',
    schema: 'public',
    synchronize: false,
    dropSchema: false,
    logging: true,
    host: DB_HOST,
    port: DB_PORT,
    username: DB_USERNAME,
    password: DB_PASSWORD,
    database: DB_DATABASE,
  },
  {
    name: 'development',
    type: 'postgres',
    schema: 'public',
    host: 'localhost',
    port: 5432,
    username: 'test',
    password: 'test',
    database: 'test',
    synchronize: false,
    logging: true,
    entities: [ 'src/**/*.entity.ts' ],
    migrations: [ 'src/migration/**/*.ts' ],
    subscribers: [ 'src/subscriber/**/*.ts' ],
    cli: {
      entitiesDir: 'src/**/*.entity.ts',
      migrationsDir: 'src/migration',
      subscribersDir: 'src/subscriber',
    },
  },
  {
    type: 'postgres',
    schema: 'public',
    host: 'localhost',
    port: 5432,
    username: 'test',
    password: 'test',
    database: 'test',
    synchronize: false,
    logging: true,
    entities: [ 'src/**/*.entity.ts' ],
    migrations: [ 'src/migration/**/*.ts' ],
    subscribers: [ 'src/subscriber/**/*.ts' ],
    cli: {
      entitiesDir: 'src',
      migrationsDir: 'src/migration',
      subscribersDir: 'src/subscriber',
    },
  },
];
`

II-alex-II:
Why is **normalizeContentOptions** adding array?
image

II-alex-II:
@imnotjames 
Isn't it weird
image

II-alex-II:
> Функция `normalizeContentOptions`не обязана возвращать массив `ConnectionOption`s, потому что во время выполнения система типов для Typescript является «лучшей из возможных» - большая часть типизации происходит на этапе компиляции с использованием статического анализа.
> 
> Я думаю, что вы передаете что-то еще не определенное - слишком много запятых или что-то в этом роде. Ваш `ormconfig.js`скриншот усечен - можете ли вы вставить его вместо того, чтобы делать скриншоты? Пожалуйста, удалите пароли, конечно, но я почти уверен, что в файле что-то не так, как ожидалось.

image

Issue #6800, opened 2020-09-29
MySQL loadTables performance


ronits-cx:
**Issue type:**

[x] bug report
[ ] feature request
[ ] documentation issue

**Database system/driver:**

[ ] `cordova`
[ ] `mongodb`
[ ] `mssql`
[x] `mysql` / `mariadb`
[ ] `oracle`
[ ] `postgres`
[ ] `cockroachdb`
[ ] `sqlite`
[ ] `sqljs`
[ ] `react-native`
[ ] `expo`

**TypeORM version:**

[ ] `latest`
[ ] `@next`
[ ] `0.x.x` (or put your version here)

**Steps to reproduce or a small repository showing the problem:**

<!--
    To answer those questions you need to put "x" inside the square brackets, for example:
    [x] `mysql`
    [ ] `postgres`

    Also, please format your code properly (by taking code blocks into ```ts .... ```)
--!>

In `MysqlQueryRunner`, `loadTables` performs queries to `INFORMATION_SCHEMA`, which is not optimized.<br/>
MySQL recommends adding a where clause with the schema and table for better performance - https://dev.mysql.com/doc/refman/5.7/en/information-schema-optimization.html.<br/><br/>

There are 2 queries with a JOIN that have a where clause on the schema and table, but the where clause is only for one of the tables in the join.<br/>
In the foreign key query, this causes an actual bug (which is what https://github.com/typeorm/typeorm/pull/6169/files solves).<br/><br/>

For large databases with many schemas, this can be a crucial improvement.<br/>
In our database, for example, each of these queries takes about **7.5 seconds**.<br/>
If a where clause is added to the second table in the join, the query takes around **200 milliseconds**.<br/>

PR #6820, opened 2020-10-01
Fix MySQL loadTables performance

ronits-cx:
This change will fix a [performance issue](https://github.com/typeorm/typeorm/issues/6800) in MySQL loadTables, which is especially bad in multi-tenanted environments.

loadTables performs queries to INFORMATION_SCHEMA, which is not optimized.
MySQL recommends adding a where clause with the schema and table for better performance - https://dev.mysql.com/doc/refman/5.7/en/information-schema-optimization.html.

There are 2 queries with a JOIN that have a where clause on the schema and table, but the where clause is only for one of the tables in the join.

For large databases with many schemas, this can be a crucial improvement.
In our database, for example, it improved each query from about **7.5 seconds** to around **200 milliseconds**.


imnotjames:
Is it possible to split these two changes apart?  The performance vs the FK query check?

It's nicer to have that from a changelog perspective & so if we have to roll things back we can isolate what we're rolling back.

ronits-cx:
@imnotjames It's possible to separate it to:
1. The FK change - changes the query result and fixes performance for that query
2. The indices change - fixes performance for that query without changing the query result
So it's not possible to completely separate performance from changing the query result
I'll prepare another PR with only the first change

imnotjames:
Is the subquery needed if we're already doing this in the where clause in the main query?

Does the `EXPLAIN` show this problem?  If so, can you post them as a reply

imnotjames (commented):
Is the test for this a duplicate of the one for 6168?

ronits-cx:
The subquery is needed for the indices because it's a left join. If it were an inner join, like the foreign keys, then we could use the main where clause, but since it's a left join we also need in the results the rows where REFERENTIAL_CONSTRAINTS columns are null. I also tried adding it in the main where clause with and OR IS NULL, but that doesn't get the performance improvement.

The explain shows the problem. For example for the foreign key query it has 2 rows, one for kcu and one for rc.
For kcu we get: `Using where; Open_full_table; Scanned 0 databases`
For rc we get: `Using where; Open_full_table; **Scanned all databases**; Using join buffer (Block Nested Loop)`
After adding the where clause for the table, we get: `Using where; Open_full_table; **Scanned 1 database**; Using join buffer (Block Nested Loop)`

ronits-cx:
The test isn't exactly a duplicate, because the 6168 PR only touched the foreign keys, so I only tested the foreign keys there. Here I also added index checks to see they're unaffected by the change.

ronits-cx:
Before - scanned all databases:
Before - scanned all databases

After - scanned 1 database:
After - scanned 1 database

ronits-cx:
@imnotjames @pleerock Hi, I noticed you merged [this refactor](https://github.com/typeorm/typeorm/pull/6851) yesterday.
I think adding the condition in the `ON` will make it impossible to have a performance improvement, even with the extra conditions in the `WHERE` clause (not to mention they're duplicate of one another).

See explain only with the `WHERE` clause conditions (rc scanned 0 databases):
Screen Shot 2020-10-06 at 9 36 42

Explain with both `WHERE` clause and `ON` conditions (rc scanned all databases):
Screen Shot 2020-10-06 at 9 36 19

However I did just notice that I used KEY_COLUMN_USAGE's columns TABLE_SCHEMA and CONSTRAINT_SCHEMA interchangeably, which is a possible bug. If TABLE_SCHEMA !== CONSTRAINT_SCHEMA then it's wrong to add the `rc` constraints inside the `foreignKeysCondition` like I did here. So I'm not sure how it's possible to add this performance improvement...

imnotjames:
I'm finding that having an `OR` in the WHERE clause at all forces a data directory scan.  This is even with a simpler query like:

```
        SELECT
            *
        FROM `INFORMATION_SCHEMA`.`KEY_COLUMN_USAGE` `kcu`

        WHERE
            (`kcu`.`TABLE_SCHEMA` = 'test' AND `kcu`.`TABLE_NAME` = 'question')
           OR
            (`kcu`.`TABLE_SCHEMA` = 'test' AND `kcu`.`TABLE_NAME` = 'category')

```

PR #6825, opened 2020-10-01
sample 35: one-to-one cascade deletion

gtpan77:
**Reason**
Users has doubt that removing any side of row in a relation would result in deletion of both side of row in the relation. [link](https://stackoverflow.com/questions/64002621/how-to-use-ondelete-cascade-on-one-to-one-relationship/64005071#comment113189983_64005071)

**Solution**
I found that only deleting referencing row would delete referenced row. Otherwise, it will only delete only one side.

If it looks legit, You can also close the issue.

### DIFF
--- a/sample/sample35-one-to-one-cascade-deletion/entity/User.ts
+++ b/sample/sample35-one-to-one-cascade-deletion/entity/User.ts
@@ -1,0 +1,16 @@
+import { Entity, PrimaryGeneratedColumn, Column, OneToOne, JoinColumn } from "../../../src/index";
+import { Profile } from "./Profile";
+
+@Entity()
+export class User {
+
+  @PrimaryGeneratedColumn()
+  id: number;
+
+  @Column()
+  name: string;
+
+  @OneToOne(() => Profile, (profile: { user: any; }) => profile.user, { eager: true, onDelete: "CASCADE" })
+  @JoinColumn()
+  profile: Profile;
+}
--- a/sample/sample35-one-to-one-cascade-deletion/entity/User.ts
+++ b/sample/sample35-one-to-one-cascade-deletion/entity/User.ts
@@ -4,13 +4,13 @@ import { Profile } from "./Profile";
 @Entity()
 export class User {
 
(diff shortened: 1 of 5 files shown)


gtpan77:
@imnotjames, Please review this PR

sample/sample35-one-to-one-cascade-deletion/entity/User.ts
```diff
@@ -0,0 +1,16 @@
...
+@Entity()
+export class User {
+
+  @PrimaryGeneratedColumn()
+  id: number;
+
+  @Column()
+  name: string;
+
+  @OneToOne(() => Profile, (profile: { user: any; }) => profile.user, { eager: true, onDelete: "CASCADE" })
```
pleerock, line 13: why do you need `(profile: { user: any; })` ? why not just `profile => profile.user` ?

pleerock:
can you please add a description to this PR - motivation behind this PR? Thanks.

sample/sample35-one-to-one-cascade-deletion/entity/User.ts
gtpan77: Thanks, I'll fix it.

gtpan77:
@pleerock, I have provided the description and motivation for this PR.

pleerock:
I changed your code and tests are still passing - it means you don't need onDelete from both sides. Only from the side where JoinColumn persists. It's a basic behaviour of any RDBMS and relations, it's not really related to ORM.

gtpan77:
Thanks for your change 👍

pleerock:
I don't see enough reasons to merge this new sample. I you had a question, and I think you got your answer now. It will be helpful to others if you answer question on SO. 

p.s. looks like `subscribersDir` change is needed since its a bugfix.

gtpan77:
Sure, You can close this PR. `subscribersDir` change was needed. Now, It is fixed in master branch.

pleerock:
Great, thanks!

PR #6828, opened 2020-10-02
Mysql foreign key multi tenant fix

ronits-cx:
This change will fix [issue 6168](https://github.com/typeorm/typeorm/issues/6168), where the foreign key queried might be wrong.
There is already another open PR for this issue: [PR 6169](https://github.com/typeorm/typeorm/pull/6169/files).
The difference is that this PR is a split from another pending [performance fix](https://github.com/typeorm/typeorm/pull/6820), which needs the condition to be where it is here in order to later add the performance fix.

The foreign key query has a JOIN and a where clause on the schema and table names, but the where clause is only for one of the tables in the join. This way in a multi-tenanted environment, it's possible for several schemas to have the same foreign key name, resulting in duplicate results in the query.


test/github-issues/6168/issue-6168.ts
```diff
@@ -0,0 +1,115 @@
...
+    before(async () => {
+        connections = await createTestingConnections({
+            entities: [__dirname + "/entity/*{.js,.ts}"],
+            enabledDrivers: ["mysql"],
+            schemaCreate: false,
+            dropSchema: false,
+        });
+
+        if (connections.length > 0) {
+            const queryRunner = connections[0].createQueryRunner();
```
imnotjames, line 74: Since you already have to have an extra line here, could this instead be -

```suggestion
        for (const connection of connections) {
          const queryRunner = connection.createQueryRunner();
```

Same for the other similar ones like this.

Right now it's just the 1 connection but if you do this it makes splitting them up easier (eg, adding maria to this)

test/github-issues/6168/issue-6168.ts
```diff
@@ -0,0 +1,115 @@
...
+    after(async () => {
+        if (connections.length > 0) {
+            const queryRunner = connections[0].createQueryRunner();
+            await queryRunner.dropDatabase("test1");
+            await queryRunner.dropDatabase("test2");
+            await queryRunner.release();
+        };
+
+        await closeTestingConnections(connections);
+        await Promise.all(testQueryRunners.map(queryRunner => queryRunner.release()));
```
imnotjames, line 92: You can create and release this in the test itself - as long as it's before the `expect`

test/github-issues/6168/issue-6168.ts
```diff
@@ -0,0 +1,115 @@
...
+        await closeTestingConnections(testConnections);
+    });
+
+    it("should only have one foreign key column", () => Promise.all(connections.map(async connection => {
+        const options = connection.options as MysqlConnectionOptions;
+
+        const connectionTest1 = await createConnection({ ...options, name: "test1", database: "test1" });
```
imnotjames, line 99: I think this should be `createTestingConnection` somehow.

Instead of creating 2 databases, couldn't you just create 1 database & then perform this test by operating against `test` (the standard test database) and `test2`?  That way you wouldn't even need to open this connection here - you could use `connection` directly.

imnotjames (changes requested):
Left a few comments :)

imnotjames (approved):
🍏

Just waiting on tests to pass!

ronits-cx:
Awesome thanks! 😄

imnotjames:
Wonderful.  Thank you for the contribution!

Issue #6509, opened 2020-08-04
If you are using SnakeNamingStrategy of "typeorm-naming-strategies" library, then you can't use entity classes for getting repositories if their name is not in snake case.


gtpan77:
If you are using SnakeNamingStrategy of "typeorm-naming-strategies" library, then you can't use entity classes for getting repositories if their name is not in snake case.
I created a entity class UserLedger.
`import { UserLedger } from "../entity/UserLedger";`
`this.userLedger = getRepository(UserLedger);`
It was failing and throwing error. RepositoryNotFoundError: No Repository for "UserLedger" was found.

changed the entity name to "user_ledger", it fixed the issue.
`this.userLedger = getRepository("user_ledger");`

_Originally posted by @gtpan77 in https://github.com/typeorm/typeorm/issues/5064#issuecomment-668500256_

korolariya:
I have same problem

imnotjames:
I believe this is a duplicate of #5676 - let me know if you think it's different, though.

Issue #5676, opened 2020-03-13
v0.2.24 not registering entities as v0.2.22


nealdyrkacz:
**Issue type:**

[ ] question
[x ] bug report
[ ] feature request
[x ] documentation issue

**Database system/driver:**

[ ] `cordova`
[ ] `mongodb`
[ ] `mssql`
[ ] `mysql` / `mariadb`
[ ] `oracle`
[x ] `postgres`
[ ] `cockroachdb`
[ ] `sqlite`
[ ] `sqljs`
[ ] `react-native`
[ ] `expo`

**TypeORM version:**

[ ] `latest`
[ ] `@next`
[x ] `0.2.24`

Been using v0.2.22 and my ormconfig.js is as follows
```
module.exports = {
  type: 'postgres',
  host: process.env.TYPEORM_HOST,
  port: process.env.TYPEORM_PORT,
  username: process.env.TYPEORM_USERNAME,
  password: process.env.TYPEORM_PASSWORD,
  database: process.env.TYPEORM_DATABASE,
  synchronize: false,
  logging: false,
  entities: ['build/entity/*.js'],
  migrations: ['build/migration/*.js'],
  subscribers: ['build/subscriber/*.js'], //'build/subscriber/**/*.js'],
  cli: {
    entitiesDir: 'src/entity',
    migrationsDir: 'src/migration',
    subscribersDir: 'src/subscriber',
  },
};

```

which works fine,  but I updated to v0.2.24 and I'm getting 
"No repository for \"<entity>\" was found. Looks like this entity is not registered in current \"default\" connection?"

Is there a documentation I am missing?

YoonSeongKyeong:
I also found this bug with mysql & typeorm v0.2.24 arising with `getRepository(<Entity>)`
everything was fine with typeorm v0.2.22

following is Error Message
```powershell
RepositoryNotFoundError: No repository for "<Entity>" was found. Looks like this entity is not registered in current "default" connection?
```

YoonSeongKyeong:
# **Debugger Behavior in RepositoryNotFoundError**

### this is v0.2.24 behavior with `getRepository(<Entity>)`
![image](https://user-images.githubusercontent.com/46678939/76674640-0c8a3b00-65f5-11ea-8966-d774b104e4c1.png)
None of above condition fall in to true

### this is v0.2.22 behavior with `getRepository(<Entity>)`
[...]

PR #6829, opened 2020-10-02
fix: snake naming strategy

gtpan77:
fix for issue #6509 

**Summary:**
Solution for this problem -
```
getRepository(UserProfile); // didn't work - throws error No Repository "UserProfile" found
getRepository("user_profile"); // works fine
```

**Reason:**
`triple equal comparison between [Function: User] and [Function: User] was returning false`

**Hack used for workaround:**
By using "user" string instead of passing entity class.

This change is working with both methods - string `"user"` and entity class `User`.

### DIFF
--- a/src/connection/Connection.ts
+++ b/src/connection/Connection.ts
@@ -484,7 +484,7 @@ export class Connection {
      */
     protected findMetadata(target: EntityTarget<any>): EntityMetadata|undefined {
         return this.entityMetadatas.find(metadata => {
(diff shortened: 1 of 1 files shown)


imnotjames:
Thanks for the contribution.

Unfortunately, this change could cause problems when there's entities that share the same class name but in different areas of the code.

As such, I'll close this, but please see #5676 as I believe it's related.

PR #6830, opened 2020-10-02
fix: Cannot find connection ${name} because its not defined in any orm configuration files.

II-alex-II:
discussion https://github.com/typeorm/typeorm/pull/6812

- [x] the setting is an Object
- [x] fix: normalizeContentOptions not updated array options (
https://visdup.blogspot.com/2019/01/how-use-external-entities-in-nestjs.html
we can add productions setting with format ts and js entity  file
)

emulate issue:
image


imnotjames:
Can you create tests that validate this fixes the issue you found?

imnotjames:
Without tests we can't confirm that this fixes any issues.

Issue #6833, opened 2020-10-02
JSON Keys break RawResult to EntityTransformer


Maddimax:
( I've provided a PR for this issue here: https://github.com/typeorm/typeorm/pull/6834 )
For the following entity:

```
class MyId {
   first: number;
   second: number;
}

@Entity({ name: 'tests' })
export class Test {
  @PrimaryColumn('jsonb')
  id: MyId;
}
```

With the following data:

| id  |
|----------------|
| {"first": 1, "second": 2} |
| {"first": 1, "second": 3} |

The following code will only return one entity:

```
getRepository('tests').createQueryBuilder()
        .select()
        .where('id IN (:...ids)', {
          ids: [
            { first: 1, second: 2 },
            { first: 1, second: 3 },
          ],
        })
        .getMany();
```

This is due to:

https://github.com/typeorm/typeorm/blob/5ef94509b89f11f8337e18046c3f9d9632d234df/src/query-builder/transformer/RawSqlResultsToEntityTransformer.ts#L66

Adding 

```
if (typeof keyValue === 'object') {
  return JSON.stringify(keyValue)
}
```

solves the issue

imnotjames:
I think we can improve this solution... Somehow.  Not sure how exactly...  JSONB and JSON aren't the same - least of which is the on-disk serialization format.

Duplicate keys, unsorted keys, etc, will cause differences between the two as well, which this will not handle.

Maddimax:
Could you elaborate what you mean by Duplicate / unsorted keys and how this would not be handled? I'm totally open to improve the solution.

Maddimax:
I've added tests for this issue in https://github.com/typeorm/typeorm/issues/6833 as well

imnotjames:
Sure - if you flip the keys around it should fail to find them but technically it's the same thing.  Both to JSONB and as parsed JSON.

```
const payload = { second: 2, first: 1 }
manager.save({ id: payload })
manager.find ({ where: { id: payload } })
```

Because JSONB won't store it as you wrote it - it should sort the keys - this should fail if it were a test.

Maddimax:
Understood. I've added a test for this and it seems as if the order does not matter: https://github.com/typeorm/typeorm/pull/6834/files#diff-09760a400d18a7e291718cb6c8f624b9R69

PR #6834, opened 2020-10-02
fix: allow for complex jsonb primary key columns 

Maddimax:
Entities with 'jsonb' primary key are not correctly returned. ( see #6833 )

This PR adds a fix and tests for Issue #6833


pleerock:
Thank you!

PR #6843, opened 2020-10-04
fix: prevent dangling transactions in sqlite

### LINKED ISSUE #3919: queryRunner.release() leaves a dangling transaction with sqlite
**Issue type:**

[ ] question
[ ] bug report
[ ] feature request
[x] documentation issue

**Database system/driver:**

[ ] `cordova`
[ ] `mongodb`
[ ] `mssql`
[ ] `mysql` / `mariadb`
[ ] `oracle`
[ ] `postgres`
[ ] `cockroachdb`
[x] `sqlite`
[ ] `sqljs`
[ ] `react-native`
[ ] `expo`

**TypeORM version:**

[x] `latest`
[ ] `@next`

```ts
  await queryRunner.startTransaction();
  try {
    if (err) return; // THIS LINE CAUSES A TRANSACTION LEAK
    await queryRunner.commitTransaction();
    return '';
  } catch(err) {
    console.log("Error. Rolling back request", info, err);
    await queryRunner.rollbackTransaction();
  } finally {
    await queryRunner.release();
  }
```

imnotjames:
when you release the sqlite runner and had an open transaction
we wouldn't actually do anything with it.  in other drivers,
such as postgres, releasing the runner closes the connection
and that causes the transaction to roll back.

to keep the behavior uniform across drivers, the sqlite
release now rolls back transactions if they're open

fixes #3919

### DIFF
--- a/src/driver/cockroachdb/CockroachQueryRunner.ts
+++ b/src/driver/cockroachdb/CockroachQueryRunner.ts
@@ -112,15 +112,17 @@ export class CockroachQueryRunner extends BaseQueryRunner implements QueryRunner
      * Releases used database connection.
      * You cannot use query runner methods once its released.
      */
-    release(): Promise<void> {
+    async release(): Promise<void> {
+        if (this.isTransactionActive) {
+            await this.rollbackTransaction();
+        }
+
         this.isReleased = true;
         if (this.releaseCallback)
             this.releaseCallback();
 
         const index = this.driver.connectedQueryRunners.indexOf(this);
         if (index !== -1) this.driver.connectedQueryRunners.splice(index);
-
-        return Promise.resolve();
     }
 
     /**
--- a/src/driver/sqlite-abstract/AbstractSqliteQueryRunner.ts
+++ b/src/driver/sqlite-abstract/AbstractSqliteQueryRunner.ts
@@ -55,12 +55,16 @@ export abstract class AbstractSqliteQueryRunner extends BaseQueryRunner implemen
 
     /**
      * Releases used database connection.
-     * We just clear loaded tables and sql in memory, because sqlite do not support multiple connections thus query runners.
+     * We just roll back any open transactions and clear loaded tables / sql in memory,
+     * because sqlite does not support multiple connections thus query runners.
      */
-    release(): Promise<void> {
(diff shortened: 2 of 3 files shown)


imnotjames:
Failing tests, oh no.

Seems cockroach (and probably Postgres... And who else...) Is affected by this.

It seems like this behavior is gonna be in a lot of places where pooling comes into play.  I feel like this is not the desired behavior to leave in place, though, so I'll continue forward on this.

imnotjames:
Ouch - turns out that we hit the same connection at the same time meaning without a lock in the SQLite query runner we end up using the same query runner across multiple async queries. ):

pleerock:
why anyone needs to release a query runner in sqlite? We have only one query runner in sqlite, right? - it shouldn't be closed.

imnotjames:
Per documentation, the `QueryRunner` is noted as running queries on a single database connection.  It keeps state & when released is supposed to "reset" the underlying database connection.  That's what the `release` function does in `AbstractSqliteQueryRunner`, as well as other `QueryRunner`s

However, what it doesn't clean up is transactions that have been started in that `QueryRunner`.  

In many of the tests we run into this problem - for example, we call `save` twice in a `Promise.all` and that ends up causing the transactions to overlap and mess with each other.  In the case of one of them completing and the other failing and requiring a rollback - that ends up not happening as expected.

Even worse, some of the tests abuse this behavior - where they will maintain a transaction from one test to another - causing general weirdness & causing some unclean test states.

If you're calling release on any other `QueryRunner` - you expect it to return the underlying connection to a pristine state, right?  Why would that be different here - even though we know `createQueryRunner` always returns the same `QueryRunner` even the TypeORM internals calls `release` and expects that to bring the connection back to a pristine state.

pleerock:
> Per documentation, the QueryRunner is noted as running queries on a single database connection.

it's important to understand that SQLite* is the only driver where it's not possible to have multiple query runners. Because there is only one connection in SQLite, we have only one QueryRunner and it doesn't make sense to create other QueryRunners, there can be only one connection. Same, it doesn't make sense to release a query runner. Releasing it means disconnecting from the database, e.g. connection.disconnect()

PR #6846, opened 2020-10-04
Use defaults provided in entity definition decorators

jougene:
Imagine then we already have some database with no default provided in schema. But in entity(model) definition we want to determine default value, without touching database schema (default modification on big table can take a very long time).
Now when you have default value provided in Column decorator (`@Column({ default: 'default_value'})`) but not defined in database schema cause error `null value in column <column_name> violates not-null constraint. But honestly I wait that this value can be inserted with "default_value" value. So I prepare repository with this issue reproduce 
- https://github.com/jougene/typeorm-default-issue

And make a pr that fix this issue.

### DIFF
--- a/src/query-builder/InsertQueryBuilder.ts
+++ b/src/query-builder/InsertQueryBuilder.ts
@@ -472,17 +472,18 @@ export class InsertQueryBuilder<Entity> extends QueryBuilder<Entity> {
 
                     // if value for this column was not provided then insert default value
                     } else if (value === undefined) {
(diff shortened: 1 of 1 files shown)


pleerock:
`default` in `@Column` is all about change in the database schema. I don't see a point to confuse things and introduce this change. If you don't need to change a database schema and instead just want to write some value, just use hooks or preset value before save.

imnotjames:
There are a number of ways to do this that don't break assumptions made about how typeorm interacts with the database.

I'm also gonna be a :-1: here - if you need this functionality please use other mechanisms to make it happen.  You can even pull the default value from the columns as part of the model initialization.

Issue #6815, opened 2020-09-30
RelationId() on nullable relation returns 'null' string


angel200sdnot:
**Issue type:**

[x] bug report
[ ] feature request
[ ] documentation issue

**Database system/driver:**

[ ] `cordova`
[ ] `mongodb`
[ ] `mssql`
[ ] `mysql` / `mariadb`
[ ] `oracle`
[x] `postgres`
[ ] `cockroachdb`
[ ] `sqlite`
[ ] `sqljs`
[ ] `react-native`
[ ] `expo`

**TypeORM version:**

[ ] `latest`
[ ] `@next`
[x] `0.2.25` (or put your version here)

**Steps to reproduce or a small repository showing the problem:**
```typescript
@Entity()
export class ChildEntity {
  @PrimaryGeneratedColumn({ type: 'bigint' })
  id: string;
}

@Entity()
export class ParentEntity {
  @PrimaryGeneratedColumn({ type: 'bigint' })
  id: string;

  @OneToOne(() => ChildEntity, { nullable: true })
  @JoinColumn()
  child: ChildEntity | null;

  @RelationId((parent: ParentEntity) => parent.child)
  childId: string | null;
}

const parent = await entityManager.getRepository(ParentEntity).findOne(parentId);
console.log(parent) // parent.childId will be 'null' if column is null in database
```
**Expected behaviour**

childId should be `null` instead of `'null'`

**Additional comments**

I am getting around this issue by using `Column()` instead of `RelationId()`

PR #6848, opened 2020-10-04
fix: return null for nullable RelationId() column

anttimaki:
Fix issue where attempting to cast bigints to strings also did it for
null values.

Fix issue where transformRelationIds() filtered out null values when
it should only be done for undefined.

Closes: #6815


pleerock:
Thank you!

Issue #2475, opened 2018-07-06
Safe Raw find operator


asafigan:
**Issue type:**

[ ] question
[ ] bug report
[x] feature request
[ ] documentation issue

Can there be a find operator that is like `Raw` but will safely insert values like the query builder's `where`?

API

`SafeRaw("title = :title", {title: "value"})`

I would like it so that like can do something like this:

```
{
    where: {
        id: In(SafeRaw(
            "select sluggableId from slug where slug in :slugs and sluggableType = :sluggableType", 
            {slugs: ['userInput'], sluggableType: Type.name}
        ))
    }
}
```

darky:
:+1: to second parameter to exist `Raw`

u8sand:
I'm also keen on getting this implemented; it will make full-text-search very simple to create an operator for. i.e. `SafeRaw(`ts_vector(${alias}) @@ plainto_tsquery('english', :1)`, query)`

I've done some work on this; perhaps it will provide a simple starting point for it to be incorporated. The following patch is a hack--I've coerced things to `any` and it should not be thought of as a ready-to-go implementation. Nonetheless, the existing `raw` operator is extended to support parameters and a working test is demonstrated. It does so by simply passing the value itself and any parameters to the `value` of `FindOperator` and sets the `useParameter` & `multipleParameters` to reuse the existing parameter passing code. The public `value` getter returns the `parameters` externally so that it gets used as expected. Internally, the raw function is given another argument for use in the query itself.

E.g. for my use case:
```
find({
  field: Raw(
    (alias, params) => `ts_vector(${alias}) @@ plainto_tsquery('english', ${params[0]})`,
    [query]
  )
})  
```

```
diff --git src/find-options/FindOperator.ts src/find-options/FindOperator.ts
index f2b8a6c5..83b35573 100644
--- src/find-options/FindOperator.ts
+++ src/find-options/FindOperator.ts
@@ -71,10 +71,15 @@ export class FindOperator<T> {
      * Gets the final value needs to be used as parameter value.
      */
     get value(): T {
-        if (this._value instanceof FindOperator)
+        if (this._value instanceof FindOperator) {
             return this._value.value;
+        }
 
-        return this._value;
+        if (this._type === 'raw') {
+            return (this._value as any).parameters;
+        } else {
+            return this._value;
+        }
     }
 
     // -------------------------------------------------------------------------
@@ -113,10 +118,10 @@ export class FindOperator<T> {
             case "isNull":
[...]

PR #6850, opened 2020-10-05
feat: add ability for escaping for Raw() find operator

temnov98:
Currently, find operator ```Raw()``` not supported escaping.
I have updated the code so that it is now possible to use an escaping.
The updated syntax is as follows:

```
await connection.getRepository(Post).find({
    likes: Raw((columnAlias, parameters) => {
        return `(${columnAlias} = ${parameters[0]}) OR (${columnAlias} = ${parameters[1]})`
    }, [2, 3]),
});

await connection.getRepository(Post).find({
    likes: Raw((columnAlias, parameters) => {
        return `(${columnAlias} = ANY(${parameters[0]})) AND (${columnAlias} < ${parameters[1]})`
    }, [[1, 4, 5, 6], 6]),
});

await connection.getRepository(Post).find({
    title: Raw((columnAlias, parameters) => {
        return `${columnAlias} = ANY(${parameters[0]})`;
    }, [["About #1", "About #3", "About #5"]]),
    likes: Raw((columnAlias, parameters) => `${columnAlias} IN (${parameters[0]}, ${parameters[1]})`, [5, 1]),
});
```

The tests for the new functionality have been written. The old functionality works unchanged.

### DIFF
--- a/src/find-options/operator/Raw.ts
+++ b/src/find-options/operator/Raw.ts
@@ -1,20 +1,47 @@
 import {FindOperator} from "../FindOperator";
+import {ObjectLiteral} from "../../common/ObjectLiteral";
 
 /**
  * Find Options Operator.
  * Example: { someField: Raw([...]) }
+ */
+export function Raw<T>(value: string): FindOperator<any>;
+
+/**
+ * Find Options Operator.
  * Example: { someField: Raw((columnAlias) => `${columnAlias} = 5`) }
- * 
+ */
+export function Raw<T>(sqlGenerator: ((columnAlias: string) => string)): FindOperator<any>;
+
+/**
+ * Find Options Operator.
  * For escaping parameters use next syntax:
  * Example: { someField: Raw((columnAlias, parameters) => `${columnAlias} = ${parameters[0]}`, [5]) }
  */
+export function Raw<T>(sqlGenerator: ((columnAlias: string, parameters: any[]) => string), parameters: any[]): FindOperator<any>;
+
+/**
+ * Find Options Operator.
+ * For escaping parameters use next syntax:
+ * Example: { someField: Raw((columnAlias) => `${columnAlias} = :value`, { value: 5 }) }
+ */
+export function Raw<T>(sqlGenerator: ((columnAlias: string) => string), parameters: ObjectLiteral): FindOperator<any>;
+
 export function Raw<T>(
     valueOrSqlGenerator: string | ((columnAlias: string, parameters: any[]) => string),
-    sqlGeneratorParameters?: any[],
+    sqlGeneratorParameters?: any[] | ObjectLiteral,
 ): FindOperator<any> {
-    if (typeof valueOrSqlGenerator === 'function') {
-        return new FindOperator("raw", sqlGeneratorParameters || [], true, true, valueOrSqlGenerator);   
-    } else {
+    if (typeof valueOrSqlGenerator !== 'function') {
         return new FindOperator("raw", valueOrSqlGenerator, false);
-    }    
+    }
+
+    if (!sqlGeneratorParameters) {
+        return new FindOperator("raw", [], true, true, valueOrSqlGenerator);
+    }
+
+    if (Array.isArray(sqlGeneratorParameters)) {
+        return new FindOperator("raw", sqlGeneratorParameters, true, true, valueOrSqlGenerator);
+    }
+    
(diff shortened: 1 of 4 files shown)


temnov98:
This functionality is useful in the case when it is necessary to pass user input in a raw query. Also you can create custom operators via ```Raw()``` operator. 

For example, operator ```ILIKE()``` which is in the next branch, but not in master can be implemented as follows:

```
function ILike(value: string) {
    return Raw((columnAlias, parameters) => `${columnAlias} ILIKE ${parameters[0]}`, [value]);
}

```

imnotjames:
Aha!  This is very nice! :) I've actually seen a few people asking about this BUT for the life of me I can't find the issues where that was.

I'll try to get to reviewing this over the next few days as I'm sure it would make a lotta people happy.

As I understand it - this is backwards compatible, as well?  With the previous functionality?  Looks to be the case there from the source.

imnotjames:
I think this Closes #2475

temnov98:
I added the ability to specify parameters using an object literal parameters for ```Raw()``` find operator. 

Example:

```
await connection.getRepository(Post).find({
    likes: Raw((columnAlias) => `${columnAlias} < :maxValue`, { maxValue: 6 }),
});
```

temnov98:
> As I understand it - this is backwards compatible, as well? With the previous functionality? Looks to be the case there from the source.

yes, all changes are backward compatible.

src/find-options/operator/Raw.ts
```diff
@@ -1,9 +1,47 @@
 import {FindOperator} from "../FindOperator";
+import {ObjectLiteral} from "../../common/ObjectLiteral";
 
 /**
  * Find Options Operator.
  * Example: { someField: Raw([...]) }
```
pleerock, line 6: I think example is incorrect here

temnov98: > I think example is incorrect here

fixed

src/find-options/operator/Raw.ts
```diff
@@ -1,9 +1,47 @@
...
+/**
+ * Find Options Operator.
+ * Example: { someField: Raw((columnAlias) => `${columnAlias} = 5`) }
+ */
+export function Raw<T>(sqlGenerator: ((columnAlias: string) => string)): FindOperator<any>;
+
+/**
+ * Find Options Operator.
+ * For escaping parameters use next syntax:
+ * Example: { someField: Raw((columnAlias, parameters) => `${columnAlias} = ${parameters[0]}`, [5]) }
```
pleerock, line 19: I can't understand why do we need this syntax. Isn't it redundant?

temnov98: > I can't understand why do we need this syntax. Isn't it redundant?

Do you suggest leaving only parameters via ObjectLiteral and remove via array? I can do it.

pleerock: Yes, it's just seems redundant. It's better to left only one syntax if they both do same and there are no technical issues.

temnov98: > Yes, it's just seems redundant. It's better to left only one syntax if they both do same and there are no technical issues.

Done. Waiting for the tests to complete.

pleerock:
Thank you a lot!

sanderevers:
Having parameters with the Raw operator is a very welcome feature for us, thanks.

I did however encounter a little catch: what happens when you use the same `:name` in two Raw instances? To some people it will probably be obvious that this should map to the same `$parameter` in the generated SQL (as it currently does). But I think you could also argue that this parameter should be local to this Raw instance.
For example, in our project, these Raw instances are automatically generated, instead of specified by hand. When they use the same `:name`, the SQL produces unexpected results (or, in the best case, breaks because of type errors). In order to guarantee a unique parameter for each instance, I now use UUIDs as names.

Now, I'm not saying that it shouldn't be allowed to share parameters between Raw instances, but
(a) this represents a design choice which I don't know you were aware of, and
(b) it would be good to document this, so users know what to expect

imnotjames:
Unfortunately, there's only support in query builders (& thus find operators) for query-builder-global variable scopes.

We don't allow for local parameter scopes at this time - and making that change would be a breaking one but we're definitely accepting pull requests that want to explore that.

This is documented [here](https://github.com/typeorm/typeorm/blob/master/docs/select-query-builder.md#using-parameters-to-escape-data)

bhaveshdaswani93:
I cannot use this change when installing using yarn add typeorm.
What would step to get this change

Issue #6268, opened 2020-06-16
MySQL timezone is not applied for Date decorators


yorickdevries:
**Issue type:**

[ ] question
[x] bug report
[ ] feature request
[ ] documentation issue

**Database system/driver:**

[ ] `cordova`
[ ] `mongodb`
[x] `mssql`
[ ] `mysql` / `mariadb`
[ ] `oracle`
[ ] `postgres`
[ ] `cockroachdb`
[ ] `sqlite`
[ ] `sqljs`
[ ] `react-native`
[ ] `expo`

**TypeORM version:**

[x] `latest`
[ ] `@next`
[ ] `0.x.x` (or put your version here)

**Steps to reproduce or a small repository showing the problem:**

<!--
When using the date decorators like ```@CreateDateColumn``` or ```@UpdateDateColumn```, the timezone shift (set in connection options) isn't applied. When retrieving these columns, the dates are however shifted, resulting in wrong timestamps.
--!>

ivanguimam:
Same problem

ivanguimam:
My solution:

```ts
const transformer: ValueTransformer = {
  from: (value) => moment.utc(value, 'YYYY-MM-DD HH:mm:ss').toDate(),
  to: (value) => value,
}

@Entity({ name: 'articles' })
class Article {
  @UpdateDateColumn({ transformer, type: 'timestamp' })
  updatedAt: Date
}
```

Obs: I saved the dates in my bank in UTC 0, only when I searched the data in the bank the dates were converted to the time of the machine that my backend is running. This way the dates will continue in UTC 0 and on my frontend I convert to the local time of the client.

jbjhjm:
I'm experiencing the same behavior with mariadb!

PR #6851, opened 2020-10-05
refactor: check mysql constraint schema on join

imnotjames:
in #6268 we pulled in a change that fixed the wrong FK
getting loaded in a multi-database / multi-schema environment
but in #6169 the approach was more desirable as it creates
a simpler query that should be easier to change in the future

as such, this pulls over that change from #6169


imnotjames:
Tests are already in place - so as long as the tests pass for the previous PR I think this is good to merge in.

pleerock:
I'm always super scary to merge changes on QueryRunners and SQLs used for schema sync. Let's merge it, but keep track on breaking change issues.

imnotjames:
:+1:

PR #6862, opened 2020-10-06
feat: backport ilike from next

imnotjames:
the next branch has support for ilike and this backports
the FindOperator into the main branch

### DIFF
--- a/src/find-options/FindOperatorType.ts
+++ b/src/find-options/FindOperatorType.ts
@@ -11,5 +11,6 @@ export type FindOperatorType = "not"
     | "in"
     | "any"
     | "isNull"
+    | "ilike"
     | "like"
     | "raw";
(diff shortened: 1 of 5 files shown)


imnotjames:
ILike is not supported in oracle.  Need to disable that test for oracle

imnotjames:
HMM, one of the mainline drivers don't support it either but I'm not sure which one.  I'll have to check locally and then publish a change to the tests to disable.

pleerock:
> HMM, one of the mainline drivers don't support it either but I'm not sure which one.

It would be great I think to quickly understand from CI what exact driver failed, to improve our developer experience.

imnotjames:
Definitely - a big reason why it's difficult is because of how we've structured our tests, with the `connections` promises.

pleerock:
Thank you!

PR #6869, opened 2020-10-08
fix(data-api): Fixed how data api driver uses and reuses a client

ArsenyYankovsky:
Now each queryRunner will have it's own client which will allow it to handle transactions correctly.


src/driver/aurora-data-api-pg/AuroraDataApiPostgresQueryRunner.ts
```diff
@@ -29,6 +29,8 @@ export class AuroraDataApiPostgresQueryRunner extends PostgresQueryRunnerWrapper
      */
     driver: AuroraDataApiPostgresDriver;
 
+    client: any;
```
imnotjames, line 32: Can this be private?

src/driver/aurora-data-api-pg/AuroraDataApiPostgresDriver.ts
```diff
@@ -90,7 +90,19 @@ export class AuroraDataApiPostgresDriver extends PostgresWrapper implements Driv
      * Creates a query runner used to execute database queries.
      */
     createQueryRunner(mode: ReplicationMode) {
-        return new AuroraDataApiPostgresQueryRunner(this, mode);
+        return new AuroraDataApiPostgresQueryRunner(
+            this,
+            new this.DataApiDriver(
```
imnotjames, line 95: Do we need the client on the driver still if we have it here?

src/driver/aurora-data-api/AuroraDataApiQueryRunner.ts
```diff
@@ -36,6 +36,8 @@ export class AuroraDataApiQueryRunner extends BaseQueryRunner implements QueryRu
 
     driver: AuroraDataApiDriver;
 
+    client: any
```
imnotjames, line 39: Can this be private?

imnotjames (changes requested):
Looks like you removed the client from the `AuroraDataApiDriver` but not the postgres one.

Do we need the client on the postgres driver still?

imnotjames:
Oof, because this is the aurora stuff we don't have tests for it.

Also, is there a related github issue?

src/driver/aurora-data-api-pg/AuroraDataApiPostgresDriver.ts
ArsenyYankovsky: Yeah it runs a couple of statements like enabling postgres extensions.

src/driver/aurora-data-api-pg/AuroraDataApiPostgresQueryRunner.ts
ArsenyYankovsky: I just made it the same level of visibility as all the other properties. If I had to I'd do protected to allow access to everyone who wants to extend the class.

ArsenyYankovsky:
> Looks like you removed the client from the `AuroraDataApiDriver` but not the postgres one.
> 
> Do we need the client on the postgres driver still?

Yeah it runs a couple of statements like enabling postgres extensions.

ArsenyYankovsky:
> Oof, because this is the aurora stuff we don't have tests for it.
> 
> Also, is there a related github issue?

I have tests in the driver repository. This is a fix for a bug that was reported there as well (https://github.com/ArsenyYankovsky/typeorm-aurora-data-api-driver/issues/5)

src/driver/aurora-data-api-pg/AuroraDataApiPostgresDriver.ts
imnotjames: The base `PostgresDriver` effectively uses a query runner to accomplish this - by calling `this.connection.query` which creates a new query runner & executes them.

Can we rely on the base `PostgresDriver` functionality instead of keeping an extra connection open at all times for this?

ArsenyYankovsky: Good catch, will fix.

src/driver/aurora-data-api-pg/AuroraDataApiPostgresQueryRunner.ts
imnotjames: The other property, driver, is on every query runner.

Well, they can't code against it if we make it private, right?  That's some of the use of visibility because it means you know what is and isn't a public API that can or can't change.

Is it possible that we can make it private and then if it's needed we can loosen up the protections on it then?

Or are you saying that there are already implementations that expect this to be available?

src/driver/aurora-data-api-pg/AuroraDataApiPostgresDriver.ts
```diff
@@ -110,7 +110,7 @@ export class AuroraDataApiPostgresDriver extends PostgresWrapper implements Driv
      * Executes given query.
      */
     protected executeQuery(connection: any, query: string) {
```
imnotjames, line 112: Can we just drop this entirely?

Does the `postgres` version of this work?

```
    /**
     * Executes given query.
     */
    protected executeQuery(connection: any, query: string) {
            connection.query(query, (err: any, result: any) => {
                if (err) return fail(err);
                ok(result);
            });
        });
    }
```

ArsenyYankovsky: Postgres version misuses the interface defined in the Connection class which defined second parameter of the query method as:
`parameters?: any[]`
And the code you pasted sends in a function.

It feels dirty and it also breaks the code.

src/driver/aurora-data-api-pg/AuroraDataApiPostgresQueryRunner.ts
ArsenyYankovsky: I made them protected. I think it's the best way forward because all the other properties there seems to have it, it makes it non-public and also allows extensibility. Also I don't think it's good making it private because the change if the need arises will take some time because people will need to wait for the release of a new version of the ORM. I also don't like making it possible for people to do their thing and if they decide to extend the class they probably know what they're doing.

imnotjames (approved):

imnotjames:
Given we can't really test this, I think it's good to merge in. :shrug:

SethVerishop:
@ArsenyYankovsky  would have been nice to keep the transactionId inside queryRunner instead of client, it's a bit confusing how the queryrunner takes a driver and client when those things reference each other

Issue #4998, opened 2019-10-30
Typeorm add select distinct in my query. I dont want select distinct. Help!


HenriqueRamos13:
**Issue type:**

[x] question

[x] `postgres`

Hello I want to make a query where I get the data above, but the typeorm is putting select distinct in the query and because of this is giving error.

`const [response, count] = await this.database
        .getRepository(Rating)
        .createQueryBuilder("rating")
        .select(
          'rating.id as id, station.name as stationname, rating.stationtext as text, rating.operatorrating as orating, rating.stationrating as srating, fillin.created_at, customer.name as cname, rating.operatorname as oname'
        )
        .innerJoin("rating.fillin", "fillin")
        .innerJoin("fillin.customer", "customer")
        .innerJoin("rating.station", "station")
        .where("fillin.is_deleted = false AND station.chain_id = :chainId", {
          chainId
        })
        //.setParameter("chainId", chainId)
        .skip(skip)
        .take(take)
        .getManyAndCount();`

The console return this: 

`query: SELECT DISTINCT "distinctAlias"."rating_id" as "ids_rating_id" FROM (SELECT "rating"."id" as id, "station"."name" as stationname, rating.stationtext as text,
[1]            rating.operatorrating as orating, rating.stationrating as srating, fillin.created_at,
[1]             "customer"."name" as cname, rating.operatorname as oname FROM "rating" "rating" INNER JOIN "fillin" "fillin" ON "fillin"."id"="rating"."fillinid"  INNER JOIN "customer" "customer" ON "customer"."id"="fillin"."customer_id"  INNER JOIN "station" "station" ON "station"."id"="rating"."stationid" WHERE fillin.is_deleted = false AND station.chain_id = $1) "distinctAlias" ORDER BY "rating_id" ASC LIMIT 10 -- PARAMETERS: ["aedbcc7f-d336-45d3-b04b-b598bd2445ee"]
`
`[1] query failed: SELECT DISTINCT "distinctAlias"."rating_id" as "ids_rating_id" FROM (SELECT "rating"."id" as id, "station"."name" as stationname, rating.stationtext as text,
[1]            rating.operatorrating as orating, rating.stationrating as srating, fillin.created_at,
[...]

PR #6870, opened 2020-10-08
fix: handle count multiple PK & edge cases more gracefully

### LINKED ISSUE #4550: getCount() not DISTINCT()?
**Issue type:**

[ ] question
[ ] bug report
[x ] feature request
[ ] documentation issue

**Database system/driver:**

[ ] `cordova`
[ ] `mongodb`
[x ] `mssql`
[ ] `mysql` / `mariadb`
[ ] `oracle`
[ ] `postgres`
[ ] `cockroachdb`
[ ] `sqlite`
[ ] `sqljs`
[ ] `react-native`
[ ] `expo`

**TypeORM version:**

[ x] `latest`
[ ] `@next`

now getCount() => select count(DISTINCT(id)), i want select count(id)` or `select count(*)

```ts
.getCount({ DISTINCT: false })
```

<!--
    [x] `msql`
    [ ] `postgres`

!>

### LINKED ISSUE #5314: getCount(). Perfs issues for multi columns primary key
**Issue type:**

[X ] question

**Database system/driver:**
[X ] `postgres`

Version 0.2.18

I'm facing a perf issue when running getCount()
Why when counting all records of a query, queryBuilder creates a SQL query that concatenates the columns that are part of the primary key ?

**Typescript  :** 
```
const query: SelectQueryBuilder<TEntity> = dbConnection
					.getRepository<TEntity>(entity)
					.createQueryBuilder(entity.name)
					.where(`"${entity.name}"."${id}"= :val`, {
						val: value.toLowerCase(),
					});

				const totalRecords: number = await query.getCount();
```

**generated sql**
```
SELECT COUNT(DISTINCT(
	CONCAT(
		"u"."a", 
		"u"."b",
		"u"."c")
	)
) as "cnt" FROM "sc"."tbl" "u"
```

That is destroying the performances for a table of 140K records. 
Creating an index on CONCAT using is not efficient, hard to maintain with 100 of tables, knowing that for some tables, some primay comumns are timestamp...

Is there a way or a parameter to set to avoid the concat and just running a kind of (`select count(1) FROM "sc"."tbl" "u"`) ?

Rgds

### LINKED ISSUE #5989: Count returns incorrect results for multiple primary keys
When using an entity with multiple primary keys, `count()` can return an incorrect result. 

The current behavior of count with primary keys is to `concat` the two keys together and then count the distinct result of that. If the primary keys are of the same data type and can contain the same values, this will result in an incorrect count. See below for an example.

There are other reasons why the current method of `concat` is not desirable, particular regarding performance as it causes indexes to be ignored.

Some related issues

- #5314 
- #4550
- #2535

**Issue type:**

[ ] question
[X] bug report
[ ] feature request
[ ] documentation issue

**Database system/driver:**

[ ] `cordova`
[ ] `mongodb`
[X] `mssql`
[ ] `mysql` / `mariadb`
[ ] `oracle`
[ ] `postgres`
[ ] `cockroachdb`
[ ] `sqlite`
[ ] `sqljs`
[ ] `react-native`
[ ] `expo`

**TypeORM version:**

[X] `latest`
[ ] `@next`

The following example incorrectly returns a count of `1` when there are `2` records

```ts
@Entity()
export class MyEntity {
	@PrimaryColumn()
	field1: number;

	@PrimaryColumn()
	field2: number;
}
```

```js
{
	field1: 11,
	field2: 101
	// concat(field1, field2) = '11101'
},
{
	field1: 1,
	field2: 1101
	// concat(field1, field2) = '11101'
},
```

<!--
    [x] `mysql`
    [ ] `postgres`

--!>

imnotjames:
currently we use concatenation of multiple primary keys and a
COUNT DISTINCT of that to figure out how many records we have
matched in a query.

however, that fails if the records have keys when the keys
are ambigious when concatenated (`"A", "AA"` & `"AA", "A"`)

the fact that we do a distinct can also be a performance impact
that isn't needed when we aren't doing joins

as such, in MySQL & Postgres we can use the built in counting
of multiple distinct values to resolve some of the issues,
and in other environments we can make it SLIGHTLY better by
adding delimiters between the concatenated values.  It is not
perfect because it technically could run into the same issue
if the delimiters are in the primary keys but it's BETTER
in most cases.

also, in cases where we do not perform any joins we can
short circuit all of this and do a much more performant
`COUNT(1)` operation

fixes #5989
fixes #5314
fixes #4550
fixes: #4998

### DIFF
--- a/src/query-builder/SelectQueryBuilder.ts
+++ b/src/query-builder/SelectQueryBuilder.ts
@@ -1774,43 +1774,74 @@ export class SelectQueryBuilder<Entity> extends QueryBuilder<Entity> implements 
         });
     }
 
-    protected async executeCountQuery(queryRunner: QueryRunner): Promise<number> {
-
+    private computeCountExpression() {
         const mainAlias = this.expressionMap.mainAlias!.name; // todo: will this work with "fromTableName"?
         const metadata = this.expressionMap.mainAlias!.metadata;
 
+        const primaryColumns = metadata.primaryColumns;
         const distinctAlias = this.escape(mainAlias);
-        let countSql: string = "";
-        if (metadata.hasMultiplePrimaryKeys) {
-            if (this.connection.driver instanceof AbstractSqliteDriver) {
-                countSql = `COUNT(DISTINCT(` + metadata.primaryColumns.map((primaryColumn, index) => {
-                    const propertyName = this.escape(primaryColumn.databaseName);
-                    return `${distinctAlias}.${propertyName}`;
-                }).join(" || ") + ")) as \"cnt\"";
 
-            } else if (this.connection.driver instanceof CockroachDriver) {
-                countSql = `COUNT(DISTINCT(CONCAT(` + metadata.primaryColumns.map((primaryColumn, index) => {
-                    const propertyName = this.escape(primaryColumn.databaseName);
-                    return `${distinctAlias}.${propertyName}::text`;
-                }).join(", ") + "))) as \"cnt\"";
-            } else if (this.connection.driver instanceof OracleDriver) {
-                countSql = `COUNT(DISTINCT(` + metadata.primaryColumns.map((primaryColumn, index) => {
-                    const propertyName = this.escape(primaryColumn.databaseName);
-                    return `${distinctAlias}.${propertyName}`;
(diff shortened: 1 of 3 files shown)


cdupetit:
Hello,
Any news on this PR ?
Thanks in advance.

PR #6879, opened 2020-10-09
fix: redundant migration with decimal default

### LINKED ISSUE #5407: Wrong migration created because of default column value format
**Issue type:**

[ ] question
[x] bug report
[ ] feature request
[ ] documentation issue

**Database system/driver:**

[ ] `cordova`
[ ] `mongodb`
[ ] `mssql`
[x] `mysql` / `mariadb`
[ ] `oracle`
[ ] `postgres`
[ ] `cockroachdb`
[ ] `sqlite`
[ ] `sqljs`
[ ] `react-native`
[ ] `expo`

**TypeORM version:**

[ ] `latest`
[ ] `@next`
[x] `0.2.22` (or put your version here)

I have the following entity
```
@Entity({
  name: "poi"
})
export class Poi {

  @PrimaryGeneratedColumn()
  id: number;

  @Column()
  normalizedNameEn: string;

  @Column({
    type: "decimal",
    precision: 11,
    scale: 8,
    default: 0
  })
  lat: number;

  @Column({
    type: "decimal",
    precision: 11,
    scale: 8,
    default: 0
  })
  lon: number;

  @Column()
  isActivated: Boolean;
}
```
`lat` and `lon` ware wrongly defined as `int` previously so I updated to decimal

when I run migrate:generate I it creates a migration file that (amongst others) contains the following lines
```
await queryRunner.query("ALTER TABLE `poi` CHANGE `lat` `lat` decimal(11,8) NOT NULL DEFAULT 0", undefined);
await queryRunner.query("ALTER TABLE `poi` CHANGE `lon` `lon` decimal(11,8) NOT NULL DEFAULT 0", undefined);
```
after I run the migration and later generate a new one I always get the following
```
        await queryRunner.query("ALTER TABLE `poi` CHANGE `lat` `lat` decimal(11,8) NOT NULL DEFAULT 0", undefined);
        await queryRunner.query("ALTER TABLE `poi` CHANGE `lon` `lon` decimal(11,8) NOT NULL DEFAULT 0", undefined);
    }

        await queryRunner.query("ALTER TABLE `poi` CHANGE `lon` `lon` decimal(11,8) NOT NULL DEFAULT '0.00000000'", undefined);
        await queryRunner.query("ALTER TABLE `poi` CHANGE `lat` `lat` decimal(11,8) NOT NULL DEFAULT '0.00000000'", undefined);
    }
```

However, I did some tests and I found that if I define `lat` and `lon` with default `'0.00000000'` typeorm will report no changes (as it should)

### LINKED ISSUE #6140: typeorm decimal value of column repeatedly generating migration ?
**Issue type:**

[ ] question
[x] bug report
[ ] feature request
[ ] documentation issue

**Database system/driver:**

[ ] `cordova`
[ ] `mongodb`
[ ] `mssql`
[x] `mysql` / `mariadb`
[ ] `oracle`
[ ] `postgres`
[ ] `cockroachdb`
[ ] `sqlite`
[ ] `sqljs`
[ ] `react-native`
[ ] `expo`

**TypeORM version:**

[x] `latest`
[ ] `@next`

```
import {Entity, PrimaryGeneratedColumn, Column} from "typeorm";

@Entity()
export class User {

    @PrimaryGeneratedColumn()
    id: number;

    @Column()
    firstName: string;

    @Column()
    lastName: string;

    @Column()
    age: number;

    @Column()
    city: number;

    @Column({ type: 'decimal', precision: 22, scale: 2,default: 0.00 })
    bday: number;

}

```

**Migration** 

```

export class Adddecl1590562535409 implements MigrationInterface {
    name = 'Adddecl1590562535409'

        await queryRunner.query("ALTER TABLE `user` CHANGE `bday` `bday` decimal(22,2) NOT NULL DEFAULT 0");
    }

        await queryRunner.query("ALTER TABLE `user` CHANGE `bday` `bday` decimal(22,2) NOT NULL DEFAULT '0.00'");
    }

}
```

I continually getting migrations generated even after running them. I am not making any changes in Schema.
Step:1 Generate a migration 
Step :2 Run a migration 
Step :3 Generate a migration 

As per **step 3** migration should not be generated because i have already ran the migration and made no changes in DB scheme since i ran migration.

**Note:**
If i don't set the default value in schema then i don't get any migration generated on step 3.

jeiea:
MySQL/MariaDB decimal column is represented like '0.0'.
MysqlDriver.normalizeDefault didn't consider trailing zeros.
User can bypass this by passing default as string, but we
can't guide the zeros by type, so we should convert user input.

Close #6140
Close #5407

### DIFF
--- a/src/driver/cockroachdb/CockroachDriver.ts
+++ b/src/driver/cockroachdb/CockroachDriver.ts
@@ -484,7 +484,7 @@ export class CockroachDriver implements Driver {
         const arrayCast = columnMetadata.isArray ? `::${columnMetadata.type}[]` : "";
 
         if (typeof defaultValue === "number") {
-            return "" + defaultValue;
+            return `(${defaultValue})`;
 
         } else if (typeof defaultValue === "boolean") {
             return defaultValue === true ? "true" : "false";
@@ -618,7 +618,7 @@ export class CockroachDriver implements Driver {
                 || tableColumn.type !== this.normalizeType(columnMetadata)
                 || tableColumn.length !== columnMetadata.length
                 || tableColumn.precision !== columnMetadata.precision
-                || tableColumn.scale !== columnMetadata.scale
+                || (columnMetadata.scale !== undefined && tableColumn.scale !== columnMetadata.scale)
                 // || tableColumn.comment !== columnMetadata.comment // todo
                 || (!tableColumn.isGenerated && this.lowerDefaultValueIfNecessary(this.normalizeDefault(columnMetadata)) !== tableColumn.default) // we included check for generated here, because generated columns already can have default values
                 || tableColumn.isPrimary !== columnMetadata.isPrimary
--- a/src/driver/postgres/PostgresDriver.ts
+++ b/src/driver/postgres/PostgresDriver.ts
@@ -393,7 +393,7 @@ export class PostgresDriver implements Driver {
             return metadata.columns.filter(column => this.spatialTypes.indexOf(column.type) >= 0).length > 0;
         });
         const hasLtreeColumns = this.connection.entityMetadatas.some(metadata => {
-            return metadata.columns.filter(column => column.type === 'ltree').length > 0;
+            return metadata.columns.filter(column => column.type === "ltree").length > 0;
         });
         const hasExclusionConstraints = this.connection.entityMetadatas.some(metadata => {
             return metadata.exclusions.length > 0;
@@ -498,7 +498,7 @@ export class PostgresDriver implements Driver {
             return `(${value.join(",")})`;
 
         } else if (columnMetadata.type === "ltree") {
-            return value.split(".").filter(Boolean).join('.').replace(/[\s]+/g, "_");
+            return value.split(".").filter(Boolean).join(".").replace(/[\s]+/g, "_");
         } else if (
             (
                 columnMetadata.type === "enum"
--- a/src/driver/postgres/PostgresDriver.ts
+++ b/src/driver/postgres/PostgresDriver.ts
@@ -734,7 +734,7 @@ export class PostgresDriver implements Driver {
         }
 
         if (typeof defaultValue === "number") {
-            return "" + defaultValue;
+            return `'${defaultValue}'`;
 
         } else if (typeof defaultValue === "boolean") {
             return defaultValue === true ? "true" : "false";
@@ -874,7 +874,7 @@ export class PostgresDriver implements Driver {
                 || tableColumn.type !== this.normalizeType(columnMetadata)
                 || tableColumn.length !== columnMetadata.length
                 || tableColumn.precision !== columnMetadata.precision
-                || tableColumn.scale !== columnMetadata.scale
+                || (columnMetadata.scale !== undefined && tableColumn.scale !== columnMetadata.scale)
                 // || tableColumn.comment !== columnMetadata.comment // todo
                 || (!tableColumn.isGenerated && this.lowerDefaultValueIfNecessary(this.normalizeDefault(columnMetadata)) !== tableColumn.default) // we included check for generated here, because generated columns already can have default values
                 || tableColumn.isPrimary !== columnMetadata.isPrimary
--- a/src/driver/cockroachdb/CockroachQueryRunner.ts
+++ b/src/driver/cockroachdb/CockroachQueryRunner.ts
@@ -1492,6 +1492,7 @@ export class CockroachQueryRunner extends BaseQueryRunner implements QueryRunner
 
                         } else {
                             tableColumn.default = dbColumn["column_default"].replace(/:::.*/, "");
+                            tableColumn.default = tableColumn.default.replace(/^(-?[\d\.]+)$/, "($1)");
                         }
                     }
 
--- a/src/driver/mysql/MysqlDriver.ts
+++ b/src/driver/mysql/MysqlDriver.ts
@@ -591,7 +591,7 @@ export class MysqlDriver implements Driver {
         }
 
         if (typeof defaultValue === "number") {
-            return "" + defaultValue;
+            return `'${defaultValue.toFixed(columnMetadata.scale)}'`;
 
         } else if (typeof defaultValue === "boolean") {
             return defaultValue === true ? "1" : "0";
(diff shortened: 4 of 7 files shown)


test/functional/migrations/generate-command/command.ts
```diff
@@ -27,10 +27,7 @@ describe("migrations > generate command", () => {
         await connection.driver.createSchemaBuilder().build();
 
         const sqlInMemory = await connection.driver.createSchemaBuilder().log();
-
-        console.log(sqlInMemory.upQueries);
```
imnotjames, line 31: :thinking: I've been looking for this!  Thank you for removing it.

test/functional/migrations/generate-command/entity/post.entity.ts
```diff
@@ -23,4 +23,6 @@ export class Post extends BaseEntity {
     @JoinTable()
     categories: Category[];
 
+    @Column("decimal", { default: -0, precision: 3, scale: 1 })
```
imnotjames, line 26: Is it possible to create another entry where there are multiple test cases?

A few that I can think of:

* Omitting the `scale`
* A bigger `default`
* A negative `scale`?  Is that possible?
* A negative default
* A string default
* A column of type int with a number default

test/functional/migrations/generate-command/command.ts
```diff
@@ -7,7 +7,7 @@ describe("migrations > generate command", () => {
     let connections: Connection[];
     before(async () => connections = await createTestingConnections({
         migrations: [],
-        enabledDrivers: ["postgres"],
+        enabledDrivers: ["mysql", "postgres"],
```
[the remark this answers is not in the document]
jeiea: Yes, I saw mysql test failure.

test/functional/migrations/generate-command/entity/post.entity.ts
jeiea: Did you mean it is necessary to add the tests for `default` feature thoroughly?
My objective is resolving those issues. I can't sure your suggestion corresponds to the pull request title.

imnotjames: Oh - `string` default won't be necessary actually, you're right - since it won't hit this case.  But the others are still relevant - the issue I'm thinking of is that your change may affect other cases of when the default is a number and we did not expect.

The test case you have here only tests for one very specific scenario - however, other permutations exist that we should check for.

EG:

```
@Column("decimal", { default: -0 })
```

What happens when the scale is omitted and undefined?  We don't really know & the change you made may or may not handle it & we should add test cases to verify that functionality.

jeiea: Yeah, actually you suggestion is great. I'll add it.
I'm thinking about adding column to the new entity of the same test. Because I don't want to increase test time. Is it permitted?

test/github-issues/5407/entity/EnumUser.ts
```diff
@@ -0,0 +1,17 @@
...
+
+enum Gender {
+  UNSPECIFIED = "",
+  FEMALE = "F",
+  MALE = "M",
+  OTHER = "O",
+}
+
+@Entity()
+export class EnumUser extends User {
```
imnotjames, line 13: Is this needed for the test?  If not, can we remove it?

test/github-issues/5407/entity/CommonUser.ts
```diff
@@ -0,0 +1,5 @@
+import {Entity} from "../../../../src/decorator/entity/Entity";
+import { User } from "./User";
+
+@Entity()
+export class CommonUser extends User { }
```
imnotjames, line 5: Same here - is this needed for the test?

imnotjames (commented):
:tada: Lookin good!  Two small questions then we're good to go I think

test/github-issues/5407/entity/EnumUser.ts
jeiea: Did you mean testing enum? I just checked it is covered in `test\functional\database-schema\enums\enums.ts`. It can be removed.

imnotjames:
Failure was an issue with circleci - kicking the tests :)

src/driver/cockroachdb/CockroachDriver.ts
```diff
@@ -618,7 +618,7 @@ export class CockroachDriver implements Driver {
                 || tableColumn.type !== this.normalizeType(columnMetadata)
                 || tableColumn.length !== columnMetadata.length
                 || tableColumn.precision !== columnMetadata.precision
-                || tableColumn.scale !== columnMetadata.scale
+                || (columnMetadata.scale && tableColumn.scale !== columnMetadata.scale)
```
imnotjames, line 621: Do we actually want to do the first check?  What if there's a `scale` on the table column but not on the column metadata?  We don't want to remove it?

Or is this for an undefined value?  In that case, should this be `(typeof columnMetadata.scale === 'undefined' && tableColumn.scale !== columnMetadata.scale)`?  As `columnMetadata.scale = 0`  would short circuit it as written.

src/driver/postgres/PostgresDriver.ts
```diff
@@ -874,7 +874,7 @@ export class PostgresDriver implements Driver {
                 || tableColumn.type !== this.normalizeType(columnMetadata)
                 || tableColumn.length !== columnMetadata.length
                 || tableColumn.precision !== columnMetadata.precision
-                || tableColumn.scale !== columnMetadata.scale
+                || (columnMetadata.scale && tableColumn.scale !== columnMetadata.scale)
```
imnotjames, line 877: Same as above - what if `columnMetadata.scale` is 0?

src/driver/cockroachdb/CockroachDriver.ts
jeiea: I wrote it for undefined value. I missed it except mysql driver. Nice comment.

test/functional/migrations/generate-command/command.ts
```diff
@@ -11,10 +11,8 @@ describe("migrations > generate command", () => {
         schemaCreate: false,
         dropSchema: true,
         entities: [Post, Category],
-        logging: true,
```
imnotjames, line 14: Can you drop the changes to this file?

imnotjames (approved):

test/functional/migrations/generate-command/command.ts
jeiea: I will do that. Can I know the reason?

imnotjames: Because the changes were mostly stylistic - so removing them reduces the footprint of this code change.

I'll separately from this drop the logging flag from testing connections.

imnotjames:
Wonderful. Thanks for the contribution!

PR #6880, opened 2020-10-09
test: remove console logging from tests

imnotjames:
we have a bunch of console loggers in our tests that cause
extra output to be emitted when we don't need it.  because
these were for debugging & not part of the tests success
this removes them


imnotjames:
I went through all of these because of #6879

Issue #2065, opened 2018-05-02
TypeError: Cannot convert object to primitive value


shashikiran797:
**Issue type:**

[ ] question
[x] bug report
[ ] feature request
[ ] documentation issue

**Database system/driver:**

[ ] `cordova`
[ ] `mongodb`
[ ] `mssql`
[ ] `mysql` / `mariadb`
[ ] `oracle`
[x] `postgres`
[ ] `sqlite`
[ ] `sqljs`
[ ] `react-native`

**TypeORM version:**

[ ] `latest`
[ ] `@next`
[x] `0.2.0` (or put your version here)

**Steps to reproduce or a small repository showing the problem:**

I am using NestJs with Typeorm
I get the request object form graphql, graphql parameters do not have ```__proto__``` set.
I am directly passing the graphql paramter object to the save method. I am getting the below error when I do that:
```
TypeError: Cannot convert object to primitive value
    at new MustBeEntityError (/Users/shashikiranms/projects/vcbackend/src/error/MustBeEntityError.ts:10:43)
    at EntityPersistExecutor.execute (/Users/shashikiranms/projects/vcbackend/node_modules/typeorm/persistence/EntityPersistExecutor.js:77:35)
    at EntityManager.save (/Users/shashikiranms/projects/vcbackend/src/entity-manager/EntityManager.ts:298:14)
    at UserRepository.Repository.save (/Users/shashikiranms/projects/vcbackend/src/repository/Repository.ts:138:29)
```

When I debugged the issue, it is beacause
EntityPersistExecutor.js checks for the entity to be an *instanceof Object*, which is false, because ```__proto__``` is not set for the entity

```js
if (!this.entity || !(this.entity instanceof Object)) {
     return Promise.reject(new MustBeEntityError_1.MustBeEntityError(this.mode, this.entity));
}
```

I think changing the above condition ``` this.entity instanceof Object``` to ``` typeof this.entity === 'object'``` solves the problem

PS: It was working in version 0.1.20. 
It is breaking once I upgraded to 0.2.0

pleerock:
`typeof null` also is an object, if value you have isn't an object then what it is?

shashikiran797:
this.entity is not null but an object, but it does not inherit from Object (i.e ```this.entity.__proto__``` is undefined) that's why ```this.entity instanceof Object``` is false.

pleerock:
okay.. are you able to provide a PR?

abrahamx97:
[...]

PR #6884, opened 2020-10-10
fix: resolves issue proto-less object validation

maayanbar13:
Due to `instanceof` comparison, object without prototype is mistaken for a privative value, throwing `TypeError: Cannot convert object to primitive value.` when trying to `.toString` said object (inside template string).

This mostly effects usage with `graphql` as `graphql-js`, [by design](https://github.com/graphql/graphql-js/pull/504), creates objects by using Object.create(null).
This fix allows saving these objects.    

The fix uses `typeof` instead of `toString` comparison since I see no reason why `new Number/Boolean/Date` should be a supported use case.

Closes: #2065

### DIFF
--- a/src/persistence/EntityPersistExecutor.ts
+++ b/src/persistence/EntityPersistExecutor.ts
@@ -41,7 +41,7 @@ export class EntityPersistExecutor {
     execute(): Promise<void> {
 
         // check if entity we are going to save is valid and is an object
-        if (!this.entity || !(this.entity instanceof Object))
+        if (!this.entity || typeof this.entity !== "object")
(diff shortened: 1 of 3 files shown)


imnotjames:
Can you create tests that verify this fixes the erroneous behavior?

maayanbar13:
@imnotjames Added tests :)

imnotjames (approved):
This makes sense to me.  Doesn't seem to introduce any backwards compatibility issues.  Has tests.  Wonderful!

imnotjames:
Thanks for the contribution and fixing this long outstanding issue :)

PR #6887, opened 2020-10-11
fix: support empty `IN` clause across all dialects

### LINKED ISSUE #2195: SQL Syntax error when empty array is given into "In" method
**Issue type:**
bug report

**Database system/driver:**
`mysql` / `mariadb`

**TypeORM version:**
`0.2.5` 

example code:
 ```ts
import { In } from 'typeorm'
// ...
const res = await Entity.find({ where: { id: In([]) } })
// ER_PARSE_ERROR: You have an error in your SQL syntax; ...
 ```
Generated SQL is similar to
 ```sql
SELECT * FROM users WHERE id IN();
```
In MySql `IN ()` (empty IN) is a syntax error. To fix it i propose that either:
* improve type annotations not to allow empty arrays to be inserted inside `In` method with guard for checking array.length to be > 0 (this would be closest to mysql spec)
* `In` method converts empty array to `FALSE` so the result would be similar to
```sql 
SELECT * FROM users WHERE id IN(FALSE); -- FALSE seems to work if used also in NOT IN(FALSE)
```
* remove `IN()` clause (may or may not create other sql syntax errors)
```sql
SELECT * FROM users WHERE id;
```

imnotjames:
we were supporting an empty `IN` clause for MySQL and Oracle
and this updates the handling to be for all other dialects as well
by making the "empty" clause be `0=1`

This follows the lead of a number of other projects in a similar space:

* [Laravel: `1=0`](https://github.com/laravel/framework/blob/8.x/src/Illuminate/Database/Query/Grammars/Grammar.php#L265-L272)
* [SQLAlchemy: `1!=1`](https://docs.sqlalchemy.org/en/13/core/sqlelement.html#sqlalchemy.sql.operators.ColumnOperators.in_)
* [Knex: `1=0`](http://knexjs.org/#Builder-whereIn)
* [Sequel: `1!=1`](https://github.com/jeremyevans/sequel/blob/67cc4245227a6c01ecb20d0138aaf37b05a5aebc/doc/release_notes/3.32.0.txt#L171)
    
I've also chosen to use `1=0` in the empty case because it's the most likely technique which allows the various query optimizers to do their magic. :)

Closes #4865
fixes #2195

### DIFF
--- a/src/query-builder/QueryBuilder.ts
+++ b/src/query-builder/QueryBuilder.ts
@@ -17,7 +17,6 @@ import {ColumnMetadata} from "../metadata/ColumnMetadata";
 import {SqljsDriver} from "../driver/sqljs/SqljsDriver";
 import {SqlServerDriver} from "../driver/sqlserver/SqlServerDriver";
 import {OracleDriver} from "../driver/oracle/OracleDriver";
-import {MysqlDriver} from "../driver/mysql/MysqlDriver";
 import {EntitySchema} from "../";
 import {FindOperator} from "../find-options/FindOperator";
 import {In} from "../find-options/operator/In";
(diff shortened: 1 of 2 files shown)


imnotjames:
The test commit had CI ran against it to show the failures - the results can be found [here](https://app.circleci.com/pipelines/github/imnotjames/typeorm/372/workflows/ffa0d826-0591-4c0e-84c0-80d12659eb14)

imnotjames:
From a performance standpoint, all query optimizers that I've checked prevent the query from even doing any work (EG, this will turn the entire query into a no-op, so no table scan happens).

I checked this against the following databases via an EXPLAIN or comparable tool:

* MSSQL
* PostgreSQL
* MySQL
* MariaDB
* SQLite (I had trouble reading it but I think the `HALT` is above the actual table scan?)
* CockroachDB

MultiWar:
Cool, thanks for working on this

Issue #3582, opened 2019-02-04
npm install when use typeorm --init


jaigm:
**Issue type:**

[ ] question
[ ] bug report
[x] feature request
[ ] documentation issue

**Database system/driver:**

[ ] `cordova`
[ ] `mongodb`
[ ] `mssql`
[ ] `mysql` / `mariadb`
[ ] `oracle`
[ ] `postgres`
[ ] `sqlite`
[ ] `sqljs`
[ ] `react-native`
[ ] `expo`

**TypeORM version:**

[x] `latest`
[ ] `@next`
[ ] `0.x.x` (or put your version here)

Any reason why are you not running `npm install` after create the project structure when using: `typeorm init --name MyProject --database mysql --express`?

vlapo:
I think the reason is that some users are using `yarn` instead of `npm`. But it is no to killing feature and maybe someone could contribute :-) 

E.g. angular-cli is good point to check:
https://github.com/angular/angular-cli/blob/master/packages/angular/cli/tasks/npm-install.ts

PR #6889, opened 2020-10-11
init cli: add option for installing package

gtpan77:
Closes: #3582

Added options for installing package either using npm or yarn by providing `--pm` or `--manager` option

### DIFF
--- a/src/commands/InitCommand.ts
+++ b/src/commands/InitCommand.ts
@@ -38,6 +38,7 @@ export class InitCommand implements yargs.CommandModule {
             .option("pm", {
                 alias: "manager",
                 choices: ["npm", "yarn"],
+                default: "npm",
                 describe: "Install packages, expected values are npm or yarn."
             });
     }
@@ -98,7 +99,7 @@ export class InitCommand implements yargs.CommandModule {
         return new Promise<string>((ok, fail) => {
             exec(command, (error: any, stdout: any, stderr: any) => {
                 if (stdout) return ok(stdout);
-                if (stderr) return ok(stderr);
+                if (stderr) return fail(stderr);
                 if (error) return fail(error);
                 ok("");
             });
--- a/src/commands/InitCommand.ts
+++ b/src/commands/InitCommand.ts
@@ -1,8 +1,9 @@
-import {CommandUtils} from "./CommandUtils";
-import {ObjectLiteral} from "../common/ObjectLiteral";
+import { CommandUtils } from "./CommandUtils";
+import { ObjectLiteral } from "../common/ObjectLiteral";
 import * as path from "path";
 import * as yargs from "yargs";
 import chalk from "chalk";
+import { exec } from "child_process";
 
 /**
  * Generates a new project with TypeORM.
@@ -33,6 +34,10 @@ export class InitCommand implements yargs.CommandModule {
             })
             .option("docker", {
                 describe: "Set to true if docker-compose must be generated as well. False by default."
+            })
+            .option("pm", {
+                alias: "manager",
+                describe: "Install packages, expected values are npm or yarn."
             });
     }
 
@@ -43,6 +48,7 @@ export class InitCommand implements yargs.CommandModule {
             const isDocker = args.docker !== undefined ? true : false;
             const basePath = process.cwd() + (args.name ? ("/" + args.name) : "");
             const projectName = args.name ? path.basename(args.name as any) : undefined;
+            const installNpm = args.pm === "yarn" ? false : true;
             await CommandUtils.createFile(basePath + "/package.json", InitCommand.getPackageJsonTemplate(projectName), false);
             if (isDocker)
                 await CommandUtils.createFile(basePath + "/docker-compose.yml", InitCommand.getDockerComposeTemplate(database), false);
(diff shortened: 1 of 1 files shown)


src/commands/InitCommand.ts
```diff
@@ -81,11 +93,22 @@ export class InitCommand implements yargs.CommandModule {
     // Protected Static Methods
     // -------------------------------------------------------------------------
 
+    protected static executeCommand(command: string) {
+        return new Promise<string>((ok, fail) => {
+            exec(command, (error: any, stdout: any, stderr: any) => {
+                if (stdout) return ok(stdout);
+                if (stderr) return ok(stderr);
```
imnotjames, line 100: Wouldn't `stderr` then a failure exit code cause this to not reject?  Because it would resolve first and then never reject.

src/commands/InitCommand.ts
```diff
@@ -202,7 +225,7 @@ export class InitCommand implements yargs.CommandModule {
                 sourceMap: true
             }
         }
-        , undefined, 3);
+            , undefined, 3);
```
imnotjames, line 228: Can you undo this change?

src/commands/InitCommand.ts
```diff
@@ -221,13 +244,13 @@ temp/`;
...
-        return `import {Entity, ${ database === "mongodb" ? "ObjectIdColumn, ObjectID" : "PrimaryGeneratedColumn" }, Column} from "typeorm";
+        return `import {Entity, ${database === "mongodb" ? "ObjectIdColumn, ObjectID" : "PrimaryGeneratedColumn"}, Column} from "typeorm";
 
 @Entity()
 export class User {
 
-    ${ database === "mongodb" ? "@ObjectIdColumn()" : "@PrimaryGeneratedColumn()" }
-    id: ${ database === "mongodb" ? "ObjectID" : "number" };
+    ${database === "mongodb" ? "@ObjectIdColumn()" : "@PrimaryGeneratedColumn()"}
+    id: ${database === "mongodb" ? "ObjectID" : "number"};
```
imnotjames, line 253: Can you undo thee changes?

src/commands/InitCommand.ts
```diff
@@ -33,6 +34,10 @@ export class InitCommand implements yargs.CommandModule {
             })
             .option("docker", {
                 describe: "Set to true if docker-compose must be generated as well. False by default."
+            })
+            .option("pm", {
```
imnotjames, line 38: Can we instead just make a flag for `yarn` and install with `npm` by default?  What does webpack do?

src/commands/InitCommand.ts
```diff
@@ -33,6 +34,10 @@ export class InitCommand implements yargs.CommandModule {
...
+                alias: "manager",
+                describe: "Install packages, expected values are npm or yarn."
```
imnotjames, line 40: If these are the valid values, should we set `choices` to these two?

src/commands/InitCommand.ts
```diff
@@ -70,6 +76,12 @@ export class InitCommand implements yargs.CommandModule {
                 console.log(chalk.green(`Project created inside current directory.`));
             }
 
+            if (args.pm && installNpm) {
+                InitCommand.executeCommand("npm install");
+            } else {
+                InitCommand.executeCommand("yarn install");
```
imnotjames, line 82: Shouldn't you `await` this?  Or we don't wait for them to complete ever?

imnotjames (changes requested):

src/commands/InitCommand.ts
gtpan77: Sure, but it is not done by me. VScode auto formatter make this change on saving the file. This change will not affect the code.

src/commands/InitCommand.ts
gtpan77: same here, these changes are only formatting. It will not affect the functioning of the code.

src/commands/InitCommand.ts
gtpan77: I thought of making npm default option if flag is not provided. We can make one flag for yarn installation only. Need your confirmation here.

src/commands/InitCommand.ts
gtpan77: my bad. `await` should be there.

src/commands/InitCommand.ts
gtpan77: Hmm.. I took this implementation from `VersionCommand.ts`. Should I call fail for `stderr`?

src/commands/InitCommand.ts
gtpan77: done

src/commands/InitCommand.ts
gtpan77: @imnotjames, need your final confirmation here.

pleerock:
I'm sorry for so late review, just found a time to review it.
Are you still interested in this change? Looks like imnotjames don't have a time currently to answer your questions. Let me review it.

src/commands/InitCommand.ts
pleerock: default should be npm, if users use `init` command without `pm` option specified - it should work and default should be npm.

src/commands/InitCommand.ts
pleerock: Yes, I think `fail` should be called here. Also, probably `VersionCommand` is incorrect.

gtpan77:
Yes, I'm interested in these changes.

src/commands/InitCommand.ts
gtpan77: done

src/commands/InitCommand.ts
gtpan77: done

AlexMesser:
thank you for contribution!

Issue #6288, opened 2020-06-21
Complement shim.js : Is there any problem adding @Unique to shim ?


Overfacto:
I've error on Angular 9 with TypeORM shim, and it is resolved with manually @Unique empty patch. 

If when I add @Unique and something decorator empty function, will this be problem?

**Issue type:**
[x] feature request

**TypeORM version:**
[x] `latest`

imnotjames:
You'll need to elaborate on your problem and what feature you're looking for - I'm not sure I understand what you'd like as a new behavior.

imnotjames:
Duplicate of #5920

Issue #5920, opened 2020-04-20
there is no Unique decoration export in typeorm-model-shim


Overfacto:
**Issue type:**
[v] feature request

**TypeORM version:**
[v] `latest`

Hello, I'm developer who loves typeorm that holy powerful and super comfortable ORM framework since using Hibernate. 

 I found the error on my front-end when use typeorm-shims with `@Unique` decorators in entity. 

typeorm shim.js is set on tsconfig. but after when I add @Unique decoration, In my case, Angular framework throw ```webpack_imported_module _9_ default(...) is not a function``` error.

I was manually patching Unique fake function to  `typeorm-model-shim.js`, then fixing this issue.

Can you add Unique for typeorm shims? Or How can I support for you? 
I hope to help typeorm. Typeorm has insufficient features but the best on all javascript ORM. I'm totally fan of you.

Overfacto:
Ah, It was error when I avoid using `@Unique` decorator in my entity. 

When I configure `@Unique` to each columns in one Entity, typeorm was threw error.

So I have to set `@Unique` on model class ( next line of `@Entity` )

PR #6891, opened 2020-10-11
fix: sync the typeorm-model-shim

imnotjames:
this updates the shim to include all current decorators
exported from index.ts - adding @Unique and a few others

fixes: #6288
fixes: #5920


pleerock:
Thanks!

PR #6892, opened 2020-10-12
feat: support query comments in the query builder

### LINKED ISSUE #3643: Query Comments
**Issue type:**

[ ] question
[ ] bug report
[x] feature request
[ ] documentation issue

Would you be open to a PR that added support for "naming" queries? I.e. something like:

```ts
Post.createQueryBuilder()
  .comment(`Lookup all the posts in ${ category } category`)
  .where({ category })
  .getSql()

/*
-- Lookup all the posts in foo category
SELECT ...
```

It would only be stored / used in TypeORM, obviously not submitted to the driver layer.

I've found a few times where I'm trying to diagnose a API call that makes multiple complex queries and is failing due to an SQL error. I end up trying to read the generated SQL and trying to infer where in my code that query came from. It would be nice to be able to label the queries such that I could easily trace them back when the error causes TypeORM to dump `getSql()` output.

imnotjames:
add a `comment` method to the QueryBuilder so we can include an
arbitrary comment in our queries for a variety of purposes -  from
query plan stability to SQL reverse proxy routing, to debugging.
the comment builder is supported in selects, inserts, deletes, and
updates

this is directly inspired by the functionality supported by hibernate
to handle SQL Query comments

it uses C-style queries which are ANSI SQL 2003 & supported in all
of the dialects of SQL that we support as drivers

fixes #3643

### DIFF
--- a/src/query-builder/DeleteQueryBuilder.ts
+++ b/src/query-builder/DeleteQueryBuilder.ts
@@ -40,7 +40,8 @@ export class DeleteQueryBuilder<Entity> extends QueryBuilder<Entity> implements 
      * Gets generated sql query without parameters being replaced.
      */
     getQuery(): string {
-        let sql = this.createDeleteExpression();
+        let sql = this.createComment();
+        sql += this.createDeleteExpression();
         return sql.trim();
     }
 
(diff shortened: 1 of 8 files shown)


pleerock:
Thank you!

PR #6903, opened 2020-10-13
fix: support changing comments in MySQL columns

imnotjames:
in the mysql driver, column comments were only half supported -
adding them to new columns & new tables was possible, but updating
existing columns to include comments was not possible

adds support to add comments to existing columns & creates tests
to verify the behavior

Closes #6336

### DIFF
--- a/src/driver/mysql/MysqlQueryRunner.ts
+++ b/src/driver/mysql/MysqlQueryRunner.ts
@@ -637,7 +637,7 @@ export class MysqlQueryRunner extends BaseQueryRunner implements QueryRunner {
                 oldColumn.name = newColumn.name;
             }
 
-            if (this.isColumnChanged(oldColumn, newColumn, true)) {
+            if (this.isColumnChanged(oldColumn, newColumn, true, true)) {
                 upQueries.push(new Query(`ALTER TABLE ${this.escapePath(table)} CHANGE \`${oldColumn.name}\` ${this.buildCreateColumnSql(newColumn, true)}`));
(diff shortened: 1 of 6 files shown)


src/driver/mysql/MysqlQueryRunner.ts
```diff
@@ -1688,8 +1704,8 @@ export class MysqlQueryRunner extends BaseQueryRunner implements QueryRunner {
             c += " PRIMARY KEY";
         if (column.isGenerated && column.generationStrategy === "increment") // don't use skipPrimary here since updates can update already exist primary without auto inc.
             c += " AUTO_INCREMENT";
-        if (column.comment)
-            c += ` COMMENT '${column.comment.replace("'", "''")}'`;
+        if (column.comment != undefined && column.comment.length > 0)
```
pleerock, line 1707: `!=` ?

pleerock:
It's super important for us to avoid breaking changes, and changes in schema sync are the most dangerous ones. That's why I'm always super dangerous on merging such PRs. Please make sure to check schema sync under different circumstances before merging it.

src/driver/mysql/MysqlQueryRunner.ts
imnotjames: I was doing this to prevent `NULL` as well as `undefined` - holdover ideas from JavaScript.  Not needed because the type shouldn't allow null.

imnotjames:
Entirely agree.  There's a number of edge cases tested here and the only code path that's not being a comments changing test should be safe.

I've tested various possible failure states and also changing the comments.  There's existing tests that test changing columns without changing comments.

Nothing is 100% safe, but these new tests help bring that % up a bit :)

PR #6909, opened 2020-10-15
refactor: create base error and simplify error creation

### LINKED ISSUE #3479: All Error thrown by TypeOrm shall have a base type
**Issue type:**

[ ] question
[ ] bug report
[x] feature request
[ ] documentation issue

**Database system/driver:**

[ ] `cordova`
[ ] `mongodb`
[ ] `mssql`
[ ] `mysql` / `mariadb`
[ ] `oracle`
[ ] `postgres`
[ ] `sqlite`
[ ] `sqljs`
[ ] `react-native`
[ ] `expo`

**TypeORM version:**

[x] `latest`
[ ] `@next`

Since the error thrown by `TypeOrm` has no unique base class, it is very difficult to accurately catch errors from `ORM` with the type-based `ExceptionFilter` of `nest.js`.

imnotjames:
This creates a base TypeORM Error that follows the conventions needed to fix the prototype tree & allow hierarchical error trees 

fixes #3479

### DIFF
--- a/src/driver/DriverFactory.ts
+++ b/src/driver/DriverFactory.ts
@@ -67,7 +67,13 @@ export class DriverFactory {
             case "capacitor":
                 return new CapacitorDriver(connection);
             default:
-                throw new MissingDriverError(type);
+                throw new MissingDriverError(
+                    type,
+                    [
+                        "cordova", "expo", "mariadb", "mongodb", "mssql", "mysql", "oracle", "postgres",
+                        "sqlite", "better-sqlite3", "sqljs", "react-native", "aurora-data-api", "aurora-data-api-pg"
+                    ]
+                );
         }
     }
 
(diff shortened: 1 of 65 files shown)


imnotjames:
Converting to draft - tests need more work.

pleerock:
I'm not that much fan of custom error classes. Do we _really_ need them? This convention came to me (like to many people probably) from java background and I think this isn't necessary have to be here, in JavaScript. Same question applies to having a single error per file. The way I design errors last few years [looks so](https://github.com/pleerock/frameworkx/blob/b397a078eb840f69645559279f46337152745760/packages/parser/src/errors.ts#L5).

imnotjames:
With non-custom errors you are unable to include extra metadata that could be useful in either debugging or gracefully handling the errors.  Right now creating an error means a bunch of boilerplate & if you miss some of that it breaks.

Per file, it's not a huge deal either way but it does make it slightly easier from the git side of things as a single file full of errors listed out like that ends up with conflicts when you have multiple people working on a single project.

PR #6924, opened 2020-10-16
fix(Postgres): generate count expression for relations without primary

MessiasLima:
the generation of count expression is based on an array of primary columns.

There are relations that don't have a primary key (e.g views). In these cases, we have to distinct count by all row.

In this PR, I put a verification for Cockroach/Postgres connection.


imnotjames:
Can you add tests for this?

imnotjames (changes requested):
This needs tests to confirm the new behavior works as expected.

dimitryvolkov:
@MessiasLima I also needed that behavior for views, temporarily I managed to work around this by definition a custom decorator for view columns delegating to the original view column code and passing 'primary: true'.
Since I need this behavior as well if you could grant me access to your fork I can write the necessary tests missing to get this moving

imnotjames:
You can always grab their branch into your own fork & open a new PR as well.

https://github.com/MessiasLima/typeorm

imnotjames:
Given this has been idle for about 6 months I'm going to close this PR.  Please feel free to re-open it fixing the issue.

PR #6926, opened 2020-10-17
fix: wrap where clauses in parentheses

### LINKED ISSUE #6170: Surround where clauses with parentheses
**Issue type:**

[ ] question
[ ] bug report
[x] feature request
[ ] documentation issue

**Database system/driver:**

[ ] `cordova`
[ ] `mongodb`
[x] `mssql`
[x] `mysql` / `mariadb`
[x] `oracle`
[x] `postgres`
[ ] `cockroachdb`
[x] `sqlite`
[ ] `sqljs`
[ ] `react-native`
[ ] `expo`

**TypeORM version:**

[ ] `latest`
[ ] `@next`
[x] `0.2.25`

It would be nice if the expression passed to `where()` and `andWhere()` methods of `QueryBuilder` was surrounded by parentheses.

For example, the following code

```ts
qb("User", "user")
  .where("user.deleteDate IS NULL")  // expression (A)
  .andWhere("user.name = 'foo' OR user.birthDate IS NULL"); // expression (B or C) 
```
currently generates the following SQL sentence:

```sql
SELECT `user`.`id`, `user`.`name`, `user`.`birthDate`, `user`.`deleteDate`
FROM `User` `user`
WHERE `user`.`deleteDate` IS NULL
AND `user`.`name` = 'foo'
OR `user`.`birthDate` IS NULL
```

As can be seen, the resulting `WHERE` clause is `A and B or C`. But it would make more sense if it was `(A) and (B or C)` - because `A` was passed to `where()` and `B or C` was passed to `andWhere()`.

I know that I can achieve this result by manually including the parentheses to every where condition. But, in my opinion, adding the parentheses automagically in SQL sentence generation would be better.

imnotjames:
prevents issues where order of operations takes precedence
over the intended query builder query

fixes #6170

### DIFF
--- a/src/query-builder/QueryBuilder.ts
+++ b/src/query-builder/QueryBuilder.ts
@@ -757,11 +757,11 @@ export abstract class QueryBuilder<Entity> {
         return this.expressionMap.wheres.map((where, index) => {
             switch (where.type) {
                 case "and":
-                    return (index > 0 ? "AND " : "") + this.replacePropertyNames(where.condition);
+                    return (index > 0 ? "AND " : "") + "(" + this.replacePropertyNames(where.condition) + ")";
                 case "or":
-                    return (index > 0 ? "OR " : "") + this.replacePropertyNames(where.condition);
+                    return (index > 0 ? "OR " : "") + "(" + this.replacePropertyNames(where.condition) + ")";
                 default:
-                    return this.replacePropertyNames(where.condition);
+                    return "(" + this.replacePropertyNames(where.condition) + ")";
             }
         }).join(" ");
     }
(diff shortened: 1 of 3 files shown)


pleerock:
Would like to understand what problem this will solve. Also, as a note, it would make sense to omit parentheses in the case if the aren't really needed.

nebkat:
If this is merged the same thing should also be added to
https://github.com/typeorm/typeorm/blob/c4a36da62593469436b074873eba186f0f8b990d/src/query-builder/SelectQueryBuilder.ts#L1719-L1734

pleerock:
I think I'll close this issue for now since its in draft and since we can break other people code currently they have relying on current behaviour. But I definitely think we should have proposed change.

Issue #2434, opened 2018-06-29
QueryBuilder insert for Oracle failed


yuluyi:
**Issue type:**

[ ] question
[X] bug report
[ ] feature request
[ ] documentation issue

**Database system/driver:**

[ ] `cordova`
[ ] `mongodb`
[ ] `mssql`
[ ] `mysql` / `mariadb`
[X] `oracle`
[ ] `postgres`
[ ] `sqlite`
[ ] `sqljs`
[ ] `react-native`

**TypeORM version:**

[X] `latest`
[ ] `@next`
[ ] `0.x.x` (or put your version here)

**Steps to reproduce or a small repository showing the problem:**

```ts
await queryRunner.manager.createQueryBuilder()
          .insert()
          .into(Person)
          .values(people)
          .execute()

// { Error: ORA-00933: SQL command not properly ended errorNum: 933, offset: 140 }
```

Oracle seems doesn't support syntax like INSERT INTO xxx VALUES (value1, ...), (value2, ...)

pleerock:
ah batch insertion is not supported by oracle. Please use a regular for loop over your values and insert data within multiple queries

wfpena:
I created a PR to solve this issue #6927

PR #6927, opened 2020-10-17
fix: support multiple row insert on oracle

wfpena:
change generated insert query to allow inserting an array of objects to oracle

fixes: #2434

### DIFF
--- a/src/query-builder/InsertQueryBuilder.ts
+++ b/src/query-builder/InsertQueryBuilder.ts
@@ -87,8 +87,9 @@ export class InsertQueryBuilder<Entity> extends QueryBuilder<Entity> {
             // console.time(".prepare returning statement");
             const returningResultsEntityUpdator = new ReturningResultsEntityUpdator(queryRunner, this.expressionMap);
             if (this.expressionMap.updateEntity === true && this.expressionMap.mainAlias!.hasMetadata) {
-                this.expressionMap.extraReturningColumns = returningResultsEntityUpdator.getInsertionReturningColumns();
-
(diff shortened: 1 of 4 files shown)


imnotjames:
Can you explain briefly how this works?

wfpena:
@imnotjames 
I've just followed the idea based on this :https://stackoverflow.com/questions/39576/best-way-to-do-multi-row-insert-in-oracle 

You can insert on oracle based on a query. So you can do something like:

```sql
ìnsert into table1
select ...
```

so what this change does is, create a query with the union of all the values on the array to be inserted, then place them on the insert statement.

imnotjames:
Can you add a test that uses a generated column

wfpena:
I've already added that

imnotjames:
Aha, I had only looked at the second model, didn't see the first one 🙈

wfpena:
I've had to add tests with primary generated columns and date columns, etc.

wfpena:
> Aha, I had only looked at the second model, didn't see the first one 🙈 I'll try to fully review this as soon as I can. Trying to get through the backlog while keeping up with new PRs.

Thank you

wfpena:
@imnotjames just realized I dind't leave the id null on the tests with primary generated, I might have to fix that.

wfpena:
> @imnotjames just realized I dind't leave the id null on the tests with primary generated, I might have to fix that.

@imnotjames the tests are fix now.

pleerock:
Good we have oracle tests running on CI now. Probably this can be merged?

PR #6928, opened 2020-10-17
test: fix test 6900 in CI

imnotjames:
test 6900 only ran correctly in local testing & didn't correctly
handle the different hostnames in CI / CD


imnotjames:
(This fixes the broken build on master - no idea why tests didn't run for the previous PR - we might want to enforce passing tests for merges)

Issue #6527, opened 2020-08-06
[DOCS] Inverted decorators in "Adjacency list" example


manuelnucci:
**Issue type:**

[ ] question
[ ] bug report
[ ] feature request
[x] documentation issue

**Database system/driver:**

[ ] `cordova`
[ ] `mongodb`
[ ] `mssql`
[ ] `mysql` / `mariadb`
[ ] `oracle`
[ ] `postgres`
[ ] `cockroachdb`
[ ] `sqlite`
[ ] `sqljs`
[ ] `react-native`
[ ] `expo`

**TypeORM version:**

[ ] `latest`
[ ] `@next`
[ ] `0.x.x` (or put your version here)

**Steps to reproduce or a small repository showing the problem:**

The decorators used in the "[Adjacency list](https://typeorm.io/#/entities/adjacency-list)" could possibly be inverted.

Here's the example as I think is correct:

```
import {Entity, Column, PrimaryGeneratedColumn, ManyToOne, OneToMany} from "typeorm";

@Entity()
export class Category {

    @PrimaryGeneratedColumn()
    id: number;

    @Column()
    name: string;

    @Column()
    description: string;

    @ManyToOne(type => Category, category => category.children) // <-- each child has only one parent
    parent: Category;

    @OneToMany(type => Category, category => category.parent) // <-- each parent has one or more children
    children: Category[];
}
```

manuelnucci:
**Update** 

It seems that only [this](https://typeorm.io/#/entities/adjacency-list) example was outdate. The one that I found [here](https://typeorm.io/#/tree-entities/adjacency-list) is correct, as I proposed initially.

Therefore, the first link should be updated to mimic the second one.

imnotjames:
Pull requests always welcome!

PR #6932, opened 2020-10-19
docs: fix decorators in example class using adjacency list

manuelnucci:
Closes: #6527

### DIFF
--- a/docs/entities.md
+++ b/docs/entities.md
@@ -705,11 +705,11 @@ export class Category {
     @Column()
     description: string;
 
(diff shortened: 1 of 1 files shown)


manuelnucci:
Inverted decorators in [this](https://typeorm.io/#/entities/adjacency-list) example to agree with the one found [here](https://typeorm.io/#/tree-entities/adjacency-list). The latter is the correct one.

imnotjames:
Makes sense to me!  Thanks for the contribution. :)

PR #6934, opened 2020-10-19
Allow Error to be a type of logQueryError param

### LINKED ISSUE #6933: Logger typings appear to be wrong
**Issue type:**

[ ] bug report
[ ] feature request
[x] documentation issue

**Database system/driver:**

[ ] `cordova`
[ ] `mongodb`
[ ] `mssql`
[ ] `mysql` / `mariadb`
[ ] `oracle`
[x] `postgres`
[ ] `cockroachdb`
[ ] `sqlite`
[ ] `sqljs`
[ ] `react-native`
[ ] `expo`

**TypeORM version:**

[ ] `latest`
[ ] `@next`
[x] `0.2.25` (or put your version here)
 but it still seems to be the case in the source code

I made a custom logger and followed the types provided by the library for `logQueryError` to take a first parameter `error` of type `string`, but when i actually inspect the `error` parameter it appears to be an instance of `Error`, not a string. This is misleading; are there certain cases where it will really be a string, or can the typing be changed entirely to `Error` or some subtype of `Error`?

osdiab:
I haven't inspected the code to know if string or any other type does get returned, but this can at least inform people without needing to inspect the code that it can be an easily parsable Error instance.

Closes #6933

### DIFF
--- a/src/logger/Logger.ts
+++ b/src/logger/Logger.ts
@@ -13,7 +13,7 @@ export interface Logger {
     /**
      * Logs query that is failed.
      */
-    logQueryError(error: string, query: string, parameters?: any[], queryRunner?: QueryRunner): any;
+    logQueryError(error: string | Error, query: string, parameters?: any[], queryRunner?: QueryRunner): any;
 
     /**
      * Logs query that is slow.


imnotjames (approved):
This makes sense - looking at where we call `logQueryError` it may only be an error but it's unclear.  Keeping the string & adding error seems like a good path forward.

PR #6945, opened 2020-10-20
fix: ILike operator generally available for any driver

leonardofalk:
https://github.com/typeorm/typeorm/issues/4418

Although this issue is already closed, I thought it could do a bit more. It makes ILIKE compatible with non Postgres drivers.

### DIFF
--- a/src/query-builder/QueryBuilder.ts
+++ b/src/query-builder/QueryBuilder.ts
@@ -916,7 +916,7 @@ export abstract class QueryBuilder<Entity> {
      * Gets SQL needs to be inserted into final query.
      */
     protected computeFindOperatorExpression(operator: FindOperator<any>, aliasPath: string, parameters: any[]): string {
-        const { driver } = this.connection
+        const { driver } = this.connection;
 
         switch (operator.type) {
             case "not":
--- a/src/query-builder/QueryBuilder.ts
+++ b/src/query-builder/QueryBuilder.ts
@@ -15,6 +15,8 @@ import {QueryDeepPartialEntity} from "./QueryPartialEntity";
 import {EntityMetadata} from "../metadata/EntityMetadata";
 import {ColumnMetadata} from "../metadata/ColumnMetadata";
 import {SqljsDriver} from "../driver/sqljs/SqljsDriver";
+import {PostgresDriver} from "../driver/postgres/PostgresDriver";
+import {CockroachDriver} from "../driver/cockroachdb/CockroachDriver";
 import {SqlServerDriver} from "../driver/sqlserver/SqlServerDriver";
 import {OracleDriver} from "../driver/oracle/OracleDriver";
 import {EntitySchema} from "../";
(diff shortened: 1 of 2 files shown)


src/query-builder/QueryBuilder.ts
```diff
@@ -932,7 +933,11 @@ export abstract class QueryBuilder<Entity> {
             case "equal":
                 return `${aliasPath} = ${parameters[0]}`;
             case "ilike":
-                return `${aliasPath} ILIKE ${parameters[0]}`;
+                if (this.connection.driver instanceof PostgresDriver) {
```
imnotjames, line 936: Also cockroachDB

src/query-builder/QueryBuilder.ts
```diff
@@ -932,7 +933,11 @@ export abstract class QueryBuilder<Entity> {
...
+                    return `${aliasPath} ILIKE ${parameters[0]}`;
+                }
+
+                return `UPPER(${aliasPath}) LIKE UPPER(${parameters[0]})`;
```
imnotjames, line 940: I believe ANSI SQL supports the `COLLATE` operator - which may be more likely to hit an index than this?

For example, SQLite supports something along the lines of..

```
foo COLLATE NOCASE LIKE "test%"
```

Would something like this be more performant?  Is it portable?  (I don't know the answer to these :smile:)

leonardofalk: Doesn't look good, I don't know if it could be more performant, but it sound like we're lying to the dev that just wanted to do `x ILIKE y`.

leonardofalk: Also, this is exactly how hibernate works:

https://github.com/hibernate/hibernate-orm/blob/master/hibernate-core/src/main/java/org/hibernate/criterion/LikeExpression.java

https://github.com/hibernate/hibernate-orm/blob/master/hibernate-core/src/main/java/org/hibernate/criterion/IlikeExpression.java

imnotjames: Ah, ok!  Great!  I'm all on board then with `lower` - I'd just remembered seeing someone do that for MSSQL once upon a time and wanted to bring it up :)

src/query-builder/QueryBuilder.ts
```diff
@@ -914,6 +916,8 @@ export abstract class QueryBuilder<Entity> {
      * Gets SQL needs to be inserted into final query.
      */
     protected computeFindOperatorExpression(operator: FindOperator<any>, aliasPath: string, parameters: any[]): string {
+        const { driver } = this.connection
```
imnotjames, line 919: You'll need a semicolon here for linting :)

pleerock (approved):

imnotjames (approved):
🍏 Once tests are good I'm happy!

leonardofalk:
Thanks guys!

PR #6978, opened 2020-10-27
Exit with code 1 if no changes were found

### LINKED ISSUE #3037: Schema migration dry run or generate .sql file
**Issue type:**

[x ] question
[ ] bug report
[ x] feature request
[ ] documentation issue

**Database system/driver:**

[ ] `cordova`
[ ] `mongodb`
[ ] `mssql`
[ ] `mysql` / `mariadb`
[ ] `oracle`
[ x] `postgres`
[ ] `sqlite`
[ ] `sqljs`
[ ] `react-native`
[ ] `expo`

**TypeORM version:**

[ x] `latest`
[ ] `@next`

Typeorm migrations seems like missing the ability to dry run migrations to look on the sql generated for migrations. This feature is impremented on [db-migrate](https://www.npmjs.com/package/db-migrate). 

It seems similar to the `schema:log`: Shows sql to be executed by schema:sync command. It shows sql log only for your default connection. To run update queries on a concrete connection use -c  option.

holm:
### Description of change

This makes `migration:generate` exit with error code 1 if no changes were found. In general I think this qualifies as a non-successful run, but the main benefit is that it allows the command to be used to verify that there are no changes between the code and the current database schema.

I am not sure how best to make a test case for this. Any pointers on this would be appreciated.

Fixes #3037

### Pull-Request Checklist

### DIFF
--- a/docs/using-cli.md
+++ b/docs/using-cli.md
@@ -203,6 +203,8 @@ Learn more about [Migrations](./migrations.md).
 Automatic migration generation creates a new migration file
 and writes all sql queries that must be executed to update the database.
 
+If no there were no changes generated, the command will exit with code 1.
+
 ```
 typeorm migration:generate -n UserMigration
 ```
--- a/src/commands/MigrationGenerateCommand.ts
+++ b/src/commands/MigrationGenerateCommand.ts
@@ -1,6 +1,5 @@
 import {ConnectionOptionsReader} from "../connection/ConnectionOptionsReader";
 import {CommandUtils} from "./CommandUtils";
-import {Connection} from "../connection/Connection";
 import {createConnection} from "../index";
 import {MysqlDriver} from "../driver/mysql/MysqlDriver";
 import {camelCase} from "../util/StringUtils";
@@ -69,7 +68,6 @@ export class MigrationGenerateCommand implements yargs.CommandModule {
             } catch (err) { }
         }
 
-        let connection: Connection|undefined = undefined;
         try {
             const connectionOptionsReader = new ConnectionOptionsReader({
                 root: process.cwd(),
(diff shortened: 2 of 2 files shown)


pleerock:
Thank you!

SrdjanCosicPrica:
This does not make much sense IMO. What happens if the command itself exists with 1 because something else is wrong with the entities/schema?

The use case for me is having a pipeline that checks that a developer has not forgotten to commit migration files. Before this fix I was able to do
```
yarn run typeorm:cli migration:generate -n __SHOULD_NOT_EXIST__
# Look for SHOULD_NOT_EXIST in the migration folder to check if developer
# has not commited changes to database entities.
occurances=$(find src/database/migrations -type f -name "*__SHOULD_NOT_EXIST__*" | wc -l)
# If there are more than 0 occurances, we throw an error
if [ "$occurances" -ne 0 ]; then
    rm src/database/migrations/*SHOULD_NOT_EXIST*
    echo "Managed to generate migration files. Be sure to commit database changes."
    exit 1
fi
```

but this fix breaks it and running the command and checking for exit 1 does not make sense as explained above. It would be much better to add a `--check` flag that does exactly this, checks if it's possible to generate migrations and exists with 1 if it can. Django ORM does exactly this and works very well for a pipeline
https://docs.djangoproject.com/en/3.1/ref/django-admin/#cmdoption-makemigrations-check

AntonPuko:
agreed, non sense change

AntonPuko:
@SrdjanCosicPrica  that is how we do it after the upgrade

migrationCheck.sh
```
echo $1
if [[ $1 == *"Your schema is up to date"* ]] ; then
    exit 0
fi
exit 1
```

package.json
```
 "migration:check": "./checkMigrations.sh \"$(node ../../node_modules/typeorm/cli.js schema:log)\"",
```

SrdjanCosicPrica:
@AntonPuko Yeah I ended up doing something similar

```
EXIT_CODE=0
output=$(yarn run typeorm:cli migration:generate -n __SHOULD_NOT_EXIST__) || EXIT_CODE=$?

echo "$output"
if [[ "$output" != *"No changes in database schema were found"* ]]; then
  rm src/database/migrations/*SHOULD_NOT_EXIST*
  echo "Managed to generate migration files. Be sure to commit database changes."
  exit 1
elif [[ "$output" == *"No changes in database schema were found"* ]]; then
  exit 0
fi

exit $EXIT_CODE
```

holm:
I added a follow up here: https://github.com/typeorm/typeorm/pull/7275

Happy to hear feedback so I don't ruin anyones day again.

Issue #6977, opened 2020-10-26
Relation columns in embedded entities are not prefixed


nebkat:
## Issue Description

When using relations inside embedded entities (`@ManyToOne`), their columns are not prefixed with the embedded entity prefix, unlike with normal `@Column`s. Whether a custom prefix is provided or the property name is used makes no difference.

See also: #3226, #3132 and #2254. Partially fixed by PR #3025.

Example scenario:

```typescript
import { Column, createConnection, Entity, ManyToOne, PrimaryGeneratedColumn } from "typeorm";

@Entity()
export class UserEntity {
    @PrimaryGeneratedColumn()
    id: number;

    @Column(() => BanInfo)
    ban: BanInfo;
}

export class BanInfo {
    @Column({nullable: true}) reason: string;

    @ManyToOne(() => UserEntity, {nullable: true}) user: UserEntity;
}

createConnection().then(async () => {}).catch(error => console.log(error));
```

### Expected Behavior
```sql
CREATE TABLE `user_entity` (
  `id` int NOT NULL AUTO_INCREMENT,
  `banUserId` int NULL,
  `banReason` varchar(255) NULL,
  PRIMARY KEY (`id`)
) ENGINE = InnoDB
```

### Actual Behavior
```sql
CREATE TABLE `user_entity` (
  `id` int NOT NULL AUTO_INCREMENT,
  `userId` int NULL, # Missing ban prefix
  `banReason` varchar(255) NULL,
  PRIMARY KEY (`id`)
) ENGINE = InnoDB
```

### Steps to Reproduce

1.
2.

### My Environment

| Dependency          | Version  |
| ---                 | ---      |
| Operating System    | Arch Linux         |
| Node.js version     | v14.14.0 | 
| Typescript version  | v4.0.3 | 
| TypeORM version     | v0.2.28 | 

### Additional Context

A single line change from #3025 (https://github.com/typeorm/typeorm/pull/3025/files#diff-c49455edefb07335bf9bd8b26a6fc6b15e6232177b5f56c2119d59043dd3a163R136) fixes the issue described above, however it is not compatible with relation ID columns. With this change applied, the code below produces a duplicate column error (both resolve correctly to banUserid but are not combined as would normally be expected).

```typescript
export class BanInfo {
    @Column({nullable: true}) reason: string;

    @ManyToOne(() => UserEntity, {nullable: true}) user: UserEntity;
    @Column() userId: number;
}
```
```sql
[...]

Issue #3226, opened 2018-12-10
Typeorm wrong generate field name for relations field in embedded entity


dimensi:
**Issue type:**

[ ] question
[x] bug report
[ ] feature request
[ ] documentation issue

**Database system/driver:**

[ ] `cordova`
[ ] `mongodb`
[ ] `mssql`
[ ] `mysql` / `mariadb`
[ ] `oracle`
[x] `postgres`
[ ] `sqlite`
[ ] `sqljs`
[ ] `react-native`
[ ] `expo`

**TypeORM version:**

[x] `latest`
[ ] `@next`
[ ] `0.x.x` (or put your version here)

**Steps to reproduce or a small repository showing the problem:**
Create relations in embedded entity and look into database.
```ts
// File.ts
import { Column, Entity, PrimaryGeneratedColumn } from 'typeorm'

@Entity()
export class File {
  @PrimaryGeneratedColumn()
  id: number

  @Column()
  path: string

  @Column()
  filename: string

  @Column()
  mimetype: string
}

// Test.ts
import { Column, Entity, JoinColumn, OneToOne, PrimaryGeneratedColumn } from 'typeorm'
import { File } from './File'

class AdditionalInfoTest {
  @OneToOne(type => File, { onDelete: 'CASCADE' })
  @JoinColumn()
  logo?: File
}

@Entity()
export class Test {
  @PrimaryGeneratedColumn()
  id: number

  @Column(type => AdditionalInfoTest)
  info: AdditionalInfoTest = new AdditionalInfoTest()
}
```

![screenshot 2018-12-10 at 19 48 04](https://user-images.githubusercontent.com/11390039/49747464-8aee8380-fcb4-11e8-9f53-a8befc6bcb3a.png)

<!--
    To answer those questions you need to put "x" inside the square brackets, for example:
    [x] `mysql`
    [ ] `postgres`

    Also, please format your code properly (by taking code blocks into ```ts .... ```)
!>

pleerock:
it doesn't look like a duplicate of #3132 but both saying that generated name is wrong. In this particular case it should be `infoLogoId`. @dimensi can you please create a failing test for this issue, and maybe even research a fix?

dimensi:
@pleerock i create [pr](https://github.com/typeorm/typeorm/pull/3248/files) with tests for this bug. I can't research a fix, because this project too hard for my skill. 
In logs i can see wrong behaviour typeorm.
```
query: BEGIN TRANSACTION
[...]

Issue #3132, opened 2018-11-20
Foreign key columns within embedded entities are not prefixed


marshall007:
**Issue type:**

[ ] question
[x] bug report
[ ] feature request
[ ] documentation issue

**Database system/driver:**

[ ] `cordova`
[ ] `mongodb`
[ ] `mssql`
[ ] `mysql` / `mariadb`
[ ] `oracle`
[x] `postgres`
[ ] `sqlite`
[ ] `sqljs`
[ ] `react-native`
[ ] `expo`

**TypeORM version:**

[x] `latest`
[ ] `@next`
[ ] `0.x.x` (or put your version here)

**Steps to reproduce or a small repository showing the problem:**

```ts
@Entity()
export class ChangeSet {
  @PrimaryColumn('uuid')
  public id: string;

  @Column(() => ChangeSetAuthority)
  public initializer: ChangeSetAuthority;
  @Column(() => ChangeSetAuthority)
  public finalizer?: ChangeSetAuthority;
}

export class ChangeSetAuthority {
  @ManyToOne(() => Artifact)
  @JoinColumn({
    name: 'forArtifactId',
  })
  public artifact?: Artifact;
  @Column()
  public forArtifactId?: string;

  @ManyToOne(() => User)
  @JoinColumn({
    name: 'forUserId',
  })
  public user: User;
  @Column()
  public forUserId: string;
}
```

In the above scenario, duplicate definitions created for the relation ID columns with _different names_ and duplicate FK constraints are generated with the _same names_. See the following migration output as an example:

```
ALTER TABLE "change_sets" ADD "forArtifactId" uuid
ALTER TABLE "change_sets" ADD "forUserId" uuid
ALTER TABLE "change_sets" ADD "initializerForartifactid" character varying NOT NULL
ALTER TABLE "change_sets" ADD "initializerForuserid" character varying NOT NULL
ALTER TABLE "change_sets" ADD "finalizerForartifactid" character varying NOT NULL
ALTER TABLE "change_sets" ADD "finalizerForuserid" character varying NOT NULL
ALTER TABLE "change_sets" ADD CONSTRAINT "FK_f0e1be463c54638152058446def" FOREIGN KEY ("forArtifactId") REFERENCES "artifacts"("id")
ALTER TABLE "change_sets" ADD CONSTRAINT "FK_7c584ae18dac9a1d4fa42c2ab03" FOREIGN KEY ("forUserId") REFERENCES "users"("id")
[...]

Issue #2254, opened 2018-06-01
 Not working properly: One-to-many relation in a Embedded entity inside Single Table Inheritance


cesarve77:
**Issue type:**

[x] bug report

**Database system/driver:**

[x] `mssql`

**TypeORM version:**

[x] `latest`

**Steps to reproduce or a small repository showing the problem:**

```
git clone https://github.com/cesarve77/typeorm-bug
cd typeorm-bug
git checkout Embedded
npm i
npm run start
```

## Entities:

```javascript

@ChildEntity()
export class Guest extends Member {

    @Column()
    cars: number
}

@Entity()
@TableInheritance({column: {name: "type", type: "varchar"}})
export  class Member {

    @PrimaryGeneratedColumn()
    id: number;

    @Column(type => BlueCard)
    blueCard: BlueCard

    @Column("varchar",{length: 50})
    name: string

    @Column("varchar",{length: 50})
    surname: string

}

export class BlueCard {

    @ManyToOne(type=>BlueCardStatus, blueCardStatus=>blueCardStatus.members, {eager: true})
    status: BlueCardStatus;

    @Column("text", {nullable: true,})
    notes: Date;

    constructor(init?: Partial<BlueCard>) {
        Object.assign(this, init);
    }
}

@Entity()
export class BlueCardStatus {

    @PrimaryGeneratedColumn()
    id: number;

    @Column("varchar", {length: 50})
    status: string

    @OneToMany(type => Member, member =>  member.blueCard  && member.blueCard.status)
    members: Member[]
}


```

## Expecting behaviour:

when:
```javascript
        const guest: Guest = new Guest()
        const parent: Parent = new Parent()

        parent.lots=12
        parent.name='name1'
        parent.surname='surname1'
        parent.blueCard=new BlueCard()
        parent.blueCard.status=new BlueCardStatus()
        parent.blueCard.status.status='status 1'

        guest.cars=1
        guest.name='name2'
        guest.surname='surname2'
        guest.blueCard=new BlueCard()
        guest.blueCard.status=new BlueCardStatus()
        guest.blueCard.status.status='status 2'

        await connection.manager.save(parent.blueCard.status)
        await connection.manager.save(guest.blueCard.status)
        await connection.manager.save(parent)
```
Everything saved (and it is saved but not related)

and the table looks like: 
![image](https://user-images.githubusercontent.com/3970983/40863060-e516805c-6631-11e8-9bec-4a3068f28e1a.png)

# Result:
[...]

PR #6981, opened 2020-10-27
fix: prefix relation id columns contained in embedded entities (#6977)

nebkat:
Searches embedded entity columns for relation ID column if relation column is in embedded entity. If not found, creates new relation ID with embedded metadata set to match the relation column.

Fixes: #2254
Fixes: #3132
Fixes: #3226
Fixes: #6977

### Description of change

See #6977. 

Currently the relation column's `entityMetadata` is searched for columns matching the join column, rather than `embeddedMetadata` (when present).

https://github.com/typeorm/typeorm/blob/a5eb946117a18d94c0157188b6a39542c8d50756/src/metadata-builder/RelationJoinColumnBuilder.ts#L136.

This changes that behavior to match 

https://github.com/typeorm/typeorm/blob/a5eb946117a18d94c0157188b6a39542c8d50756/src/metadata-builder/EntityMetadataBuilder.ts#L152-L156

### Pull-Request Checklist

### DIFF
--- a/src/metadata-builder/RelationJoinColumnBuilder.ts
+++ b/src/metadata-builder/RelationJoinColumnBuilder.ts
@@ -133,11 +133,19 @@ export class RelationJoinColumnBuilder {
             });
             const joinColumnName = joinColumnMetadataArg ? joinColumnMetadataArg.name : this.connection.namingStrategy.joinColumnName(relation.propertyName, referencedColumn.propertyName);
 
-            let relationalColumn = relation.entityMetadata.ownColumns.find(column => column.databaseName === joinColumnName);
+            let relationalColumn: ColumnMetadata | undefined;
+            if (this.connection.options.enableNextFeatures) {
+                // TODO: NEXT 0.3.0 default to fixed embedded columns
+                const relationalColumns = relation.embeddedMetadata ? relation.embeddedMetadata.columns : relation.entityMetadata.ownColumns;
+                relationalColumn = relationalColumns.find(column => column.databaseNameWithoutPrefixes === joinColumnName);
+            } else {
+                relationalColumn = relation.entityMetadata.ownColumns.find(column => column.databaseName === joinColumnName);
+            }
             if (!relationalColumn) {
                 relationalColumn = new ColumnMetadata({
                     connection: this.connection,
                     entityMetadata: relation.entityMetadata,
+                    embeddedMetadata: this.connection.options.enableNextFeatures ? relation.embeddedMetadata : undefined, // TODO: NEXT 0.3.0 default to fixed embedded columns
                     args: {
                         target: "",
                         mode: "virtual",
--- a/test/functional/entity-metadata/entity/Counters.ts
+++ b/test/functional/entity-metadata/entity/Counters.ts
@@ -18,7 +18,7 @@ export class Counters {
     @Column()
     favorites: number;
 
-    @Column(() => Subcounters)
+    @Column(() => Subcounters, {prefix: "sub"})
     subcounters: Subcounters;
 
     @ManyToMany(type => User, user => user.likedPosts)
--- a/src/connection/BaseConnectionOptions.ts
+++ b/src/connection/BaseConnectionOptions.ts
@@ -175,4 +175,8 @@ export interface BaseConnectionOptions {
 
     };
 
+    /**
+     * Flag to enable breaking changes that are planned for release in next version.
+     */
+    readonly enableNextFeatures?: boolean;
 }
--- a/src/connection/options-reader/ConnectionOptionsEnvReader.ts
+++ b/src/connection/options-reader/ConnectionOptionsEnvReader.ts
@@ -46,7 +46,8 @@ export class ConnectionOptionsEnvReader {
                 subscribersDir: PlatformTools.getEnvVariable("TYPEORM_SUBSCRIBERS_DIR"),
             },
             cache: this.transformCaching(),
-            uuidExtension: PlatformTools.getEnvVariable("TYPEORM_UUID_EXTENSION")
+            uuidExtension: PlatformTools.getEnvVariable("TYPEORM_UUID_EXTENSION"),
+            enableNextFeatures: !!PlatformTools.getEnvVariable("TYPEORM_ENABLE_NEXT_FEATURES")
         }];
     }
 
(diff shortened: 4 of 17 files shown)


imnotjames:
Looks like this change causes the oracle tests to fail?

nebkat:
Oracle issue with max identifier length being 30 characters, looks like it never occurred in the past because the column names always excluded the embedded prefixes. Fixed by shortening a prefix in the edge cases with long identifiers.

src/metadata-builder/RelationJoinColumnBuilder.ts
```diff
@@ -133,11 +133,13 @@ export class RelationJoinColumnBuilder {
             });
             const joinColumnName = joinColumnMetadataArg ? joinColumnMetadataArg.name : this.connection.namingStrategy.joinColumnName(relation.propertyName, referencedColumn.propertyName);
 
-            let relationalColumn = relation.entityMetadata.ownColumns.find(column => column.databaseName === joinColumnName);
+            const relationalColumns = relation.embeddedMetadata ? relation.embeddedMetadata.columns : relation.entityMetadata.ownColumns;
```
pleerock, line 136: what if there is an embedded, but we point to a relation that not in the embedded?

test/functional/entity-metadata/entity/Counters.ts
```diff
@@ -18,7 +18,7 @@ export class Counters {
     @Column()
     favorites: number;
 
-    @Column(() => Subcounters)
+    @Column(() => Subcounters, {prefix: "sub"})
```
pleerock, line 21: Why did you change exist test? You need to create a separate test for a new use case.

nebkat: The columns inside embedded entities were being incorrectly generated previously. The correct column name is `countersSubcountersWatchedUserId`, which is 32 characters long, and too long for Oracle DB which can only take 30 characters. Before this fix, the column name would have been `watchedUserId` or something, so there was no problem, but the wrong column name was going unnoticed.

This test only checks the ORM side of things anyway, so the exact name in database was irrelevant, but I can change the variable name from `subcounters` to `sub` directly if you'd prefer.  I have already created a new test case for the issue #6977.

src/metadata-builder/RelationJoinColumnBuilder.ts
nebkat: I'm not sure what you mean, I think `relation.embeddedMetadata` is only set if `relation` is a column within an embedded entity. It was somewhat confusing initially but `relation` refers only to the foreign key column, not the target relation entity. 

This particular section is what searches for an existing relation id column, and if the relation is located inside an embedded then it only makes sense to search for the relation id column also in the embedded rather than in the base of the entity.

test/functional/entity-metadata/entity/Counters.ts
pleerock: the problem is, if we merge these changes, we introduce a breaking change. Right now Oracle users have same case working (even if column name is invalid), but if we merge these changes their codebases won't be working at all. We can't tell them "hey guys just monkey patch it using "prefix" like we did it on our tests". Proper solution is to fix related Oracle issue.

pleerock:
@imnotjames it looks like this will introduce a serious breaking change for the users who have same use case. Right now what we have is a bug, this is true. If we merge these changes bug will be fixed, but it will also lead to schema re-creation and data loss for users. We need to decide how much acceptable it is to merge these changes not into the `next`.

src/metadata-builder/RelationJoinColumnBuilder.ts
pleerock: okay, thank you for explanation

test/functional/entity-metadata/entity/Counters.ts
nebkat: For this test anyway there is no solution for Oracle unless we plan to truncate column names automatically? Could potentially introduce a more informative error during column generation if target DB is Oracle since the DB error is very vague.

pleerock: we already do truncate in [some other places](https://github.com/typeorm/typeorm/blob/494866a999606e4acf8275aaae14099b0c1a0be1/src/find-options/FindOptionsUtils.ts#L229) for Oracle

nebkat:
I see the difficulty of maintaining the project now 🙂 Would it be worth introducing a temporary option property to relations that applies this fix until next major release? Definitely shouldn't break existing setups, but unless the next release is coming soon it might be frustrating for users and will also lead to more broken columns being initiated in new projects. In my case I have multiple userId references and it fails because of duplicate column so I can't even keep the ORM side correct.

`@ManyToOne(() => UserEntity, {fixEmbeddedRelationColumn: true})`?

pleerock:
if we go this way, but better to make same a connection-level option

nebkat:
Sounds good, if you think this is best I can implement it. I propose maybe a field in `BaseConnectionOptions` e.g. `targetVersion: "0.3.0"` which defaults to the current release, string comparison as a condition for the breaking changes, along with a strong warning for users that things are likely to break if they target future versions. Once that version is released the old code gets removed, but hopefully additional feedback can be gained from users willing to test new features.

src/metadata-builder/RelationJoinColumnBuilder.ts
nebkat: I've had a second look at this, and in theory we could allow relation IDs to not be in the embedded entity together with the relation, but I'm not sure whether that would cause additional confusion.

```typescript
@Entity()
class Other {
    @PrimaryColumn() id: number;
}
@Entity()
class User {
    @Column(() => Embed) embed: Embed;

    @Column() embedOther1Id: number;
    @ManyToOne(() => Other) embedOther2: Other;
}
class Embed {
    @ManyToOne(() => Other) other1: Other;
    @Column() other2Id: number;
}
```

Both relations here could match up correctly if we want, and if they are not matched up they will lead to a duplicate column error anyway. MongoDB is an exception since it maintains the embedded objects in DB, so both relation/ID have to be in the same entity.

pleerock:
I think it's better just to call them "enableNextFeatures", don't think we'll really try to support more than one next breaking version.

test/functional/entity-metadata/entity/Counters.ts
nebkat: Might be something worth looking at with naming strategies for the future, gives the user the option for more meaningful change than simple truncation.

nebkat:
Since latest TypeScript is merged the next release should be considered breaking, can the `enableNextFeatures` flag be removed and have the change applied by default?

PR #6996, opened 2020-10-31
transactional events in subscriber interface

pleerock:


pleerock:
@imnotjames I'm going to merge this one and release a new version. Do you want to review it before I do that?

src/find-options/FindOneOptions.ts
```diff
@@ -59,4 +59,9 @@ export interface FindOneOptions<Entity = any> {
      */
     loadEagerRelations?: boolean;
 
+    /**
+     * If this is set to true, SELECT query in a `find` method will be executed in a transaction.
+     */
+    transaction?: boolean
```
imnotjames, line 65: I'm assuming this was intentionally added but seems to be a separate feature from the rest of this PR

imnotjames (approved):
Overall makes sense.

This does make the disparity between listeners, schema listeners, and subscribers wider, but there's already disparity there as it stands.

src/find-options/FindOneOptions.ts
pleerock: oh yeah I forgot I need to create a separate branch for this separate feature

pleerock:
> This does make the disparity between listeners, schema listeners, and subscribers wider, but there's already disparity there as it stands.

yes it does, and here is why:

when you run:

```sql
START TRANSACTION
SELECT * FROM user
COMMIT
```

you execute:

```ts
entity.beforeTransactionStart()
entity.afterTransactionStart()
```

there is no "entity" before SELECT on which we can call `beforeTransactionStart` method.

PR #7001, opened 2020-11-02
fix: reject nullable primary columns

lilred:
Every supported back-end requires that primary key columns be non-nullable.
As of this PR, this mistake is caught at compile-time rather than as part of execution.

BREAKING CHANGE: passing `ColumnOptions` to `@PrimaryKeyColumn` does not function anymore. One must use `PrimaryKeyColumnOptions` instead.

### Description of change

Forbid primary columns from being marked as nullable. This condition is expected by all currently-supported back-ends and is generally understood to be standard.

As of the time this PR was opened, some of the tests aren't passing on my machine, and I've not touched documentation. I'd like to validate that this change is desired and likely to be merged before spending more time on it.

### Pull-Request Checklist

- [x] This pull request links relevant issues as `Fixes #0000` (N/A)
- [x] There are new or updated unit tests validating the change (N/A)

### DIFF
--- a/src/decorator/columns/PrimaryColumn.ts
+++ b/src/decorator/columns/PrimaryColumn.ts
@@ -5,35 +5,41 @@ import {ColumnMetadataArgs} from "../../metadata-args/ColumnMetadataArgs";
 import {GeneratedMetadataArgs} from "../../metadata-args/GeneratedMetadataArgs";
 
 /**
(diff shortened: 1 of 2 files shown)


pleerock:
Looks correct. I don't think that much of a breaking change since there should be only few people (if not less) who could do it.

pleerock:
Thank you for contribution! 🎉

PR #7006, opened 2020-11-03
feat: export all errors

### LINKED ISSUE #7000: Export error types for checks on the outside
### The Problem

We use NestJS with the TypeOrm integration and try to catch TypeOrm errors with a NestJs-Filter. Unfortunatley only the `QueryFailedError` is exported from the TypeOrm-Library. This stops us from being able to check something like `error instanceOf EntityNotFoundError`. Worse, we cannot specify this error as part of the `Catch` clause from NestJS.

### The Solution

Export all errrors. 

Agreon:
### Description of change

This PR adds the export of all errors.

Fixes #7000

### Pull-Request Checklist


rubiin:
@imnotjames can have this in next realease, most awaited feature for me

imnotjames (commented):
Instead of exporting each of these individually in the `src/index.ts`, would it be clearer to create an `index.ts` in `error` which iterates through them?

That way we could do `export * from "./error"` and isolate the error export iteration to the error dir.

What do you think?

Agreon:
That's a good idea an i had it myself, i just oriented myself on the existing code to stay consistent. But of course, will do it in a second

Agreon:
@imnotjames ping!

rubiin:
I think this pr can be merged if there are no any alterations to be made

Agreon:
Why is this not merged? It's not even hard to review :/

rubiin:
@imnotjames little help here

nebkat:
There is still the unresolved conversation in #6909. Personally I think the classes should all be exported but there may be reasons not to, for example avoiding creating a public API that will later result in breaking changes if the errors are renamed.

Either way it's not being ignored, pleerock has just not gotten around to all the pending pull requests and there are many others waiting too. If you need it I suggest making a local repo with the changes applied and using `npm link`.

pleerock:
I don't think errors should be part of public API and I don't really like how errors are currently designed, but we can merge this PR if somebody needs it.

pleerock:
Thank you for contribution! Please keep in mind in the future we most probably will change errors public api.

Issue #7008, opened 2020-11-03
Unable to addSelect with computed result


Asarew:
when using a computed selection the property doesn't get hydrated

```ts
@Entity()
export class Post {

    @PrimaryGeneratedColumn()
    id: number;

    @Column()
    title: string;

    @Column({ type: String, nullable: true })
    text: string|null;
}
```
```ts
const query = connection.manager.createQueryBuilder(Post, "post");
const loadedPost = await query
    .select("post.id")
    .addSelect("SUBSTRING(post.text, 1, 20)", "post_text")
    .getOne();

if (!loadedPost.text) {
   throw new Error('text not hydrated');
} else if (loadedPost.length > 20) {
  throw new Error('text length to large');
} else {
  console.log('everything is all good')
}


```
### Expected Behavior
This would output `everything is all good`.

### Actual Behavior
Error is thrown: `text not hydrated`

### Are you willing to resolve this issue by submitting a Pull Request?
I've already fixed this issue in this PR: #4703 but it seems to getting ignored due to the fact that it doesn't support all use cases of the parties involved that are mainly interested in `addSelectAndMap`. 
But because a feature PR for that is currently in the works #6855, i think that we can just merge #4703 as it doesn't introduce any new features but fixes a Very long standing bug.

nebkat:
@Asarew I share your frustration at the lack of this feature for so long but this is really isn't constructive. The developers are clearly aware of this and you are simply wasting their time by creating new issues. I am working on a solution that will hopefully satisfy all parties because personally I don't think any of the provided solutions are actually going to be good long term.

Asarew:
The bug that I'm describing here has nothing to do with `addSelectAndMap`, So I'm not "waisting time" of any body except yours as you clearly haven't looked into the issue. `addSelectAndMap` works with plain properties and NOT with columns.

nebkat:
@Asarew I think the fact that you have been able to link to comments, issues and pull requests that all describe your situation shows that this is a known and duplicate issue, which therefore makes this a waste of their time. 

I can assure you I have looked into your issue and all of the related issues from the past.

[...]

Issue #1822, opened 2018-03-25
Add custom select and map to property of entity


ToonvanStrijp:
**Issue type:**

[ ] question
[ ] bug report
[X ] feature request
[ ] documentation issue

**Database system/driver:**

[X] `cordova`
[X] `mongodb`
[X] `mssql`
[X] `mysql` / `mariadb`
[X] `oracle`
[X] `postgres`
[X] `sqlite`
[X] `sqljs`
[X] `websql`

**TypeORM version:**

[ ] `latest`
[ ] `@next`
[ ] `0.x.x` (or put your version here)

**Steps to reproduce or a small repository showing the problem:**
What I to accomplish is to add a custom select that maps to a property of my Entity. I've found an work around but would like to know if there is an easier/cleaner way to do this.

So I've my model:

```
@Entity('groups')
export class Group extends BaseEntity{

    @PrimaryGeneratedColumn()
    id: number;

    @ManyToMany(type => User, user => user.groups)
    @JoinTable({
        name: 'groups_users',
        joinColumn: {
            name: 'group_id',
            referencedColumnName: 'id'
        },
        inverseJoinColumn:
            {name: 'user_id',
                referencedColumnName: 'id'
            }
    })
    users: User[];

    active: boolean;
}
```

This entity has a ManyToMany relationship to User. The `JoinTable` `groups_users` has one extra column `active` this `active` indicates if the group that the user is linked to is visible.

When I query the groups to get all the groups for one user (also the inactive groups) I do this:

```
let query = this.groupRepository
            .createQueryBuilder('group')
            .innerJoin('groups_users', 'gu')
            .innerJoin('group.users', 'user', 'user.deleted_at is null AND group_user.user_id = :id',
                {id: userId})
            .orderBy('group.id', 'ASC')
            .andWhere('group.deleted_at IS NULL');
```

As you can see the `Group` entity has a property called active. As for now the `active` property is not seen by `typeorm` so what I did was the following:

I added this to the active property:

```
 @Column('tinyint', {
        nullable: true,
        select: false
    })
    active: boolean;
```

By setting `select` to `false` the other queries won't break (since active isn't a real column inside the `group` table).

Now I added this to the query:

```
.addSelect(s => s.select('active').from('groups_users', 'gu').where('gu.group_id = group.id AND gu.user_id = :id', {id: userId}), 'group_active')
```

By doing this the active property is filled for each user and the `active` is true or false according the 0 or 1 in `JoinTable`.

[...]

PR #7010, opened 2020-11-03
feat: selectAndMap, computed/virtual/mapped columns (WIP)

### LINKED ISSUE #296: select additional computed columns

nebkat:
*Still WIP, gathering feedback*

Fixes #296, #1822, #7008
Alternative to #4703, #6855

### Description of change
Universal `selectAndMap` for computed/virtual/mapped columns with minimal disruption to existing systems and no changes to database drivers.

The column system in TypeORM can generally be broken down into two sides: entity -> database and database -> entity. With computed columns everything in the database -> entity side remains useful, namely the existing `RawSqlResultsToEntityTransformer`, `Driver.prepareHydratedValue` and `SelectQueryBuilder` column aliasing, while the entity -> database side should not be used at all.

With this solution for `selectAndMap` I have attempted to make the most of the column system and keep a similar pattern for defining computed properties.

#### 1. `@Virtual(type: ColumnType, options?: VirtualOptions)` decorator
A new decorator is introduced for marking entity properties as being virtual and storing their database type for later reference. It removes all the options from `@Column()` that are only relevant to database storage. The `ColumnMetadataArgs.mode` is set to `'computed'`.

Normally the type will be inferred from the property type, but for special values such as `'json'`, `'date'`, `'enum'` and `'hstore'` I have left the parameter in. Ideally the available `ColumnType`s would be more restrictive given that a `'tinyint'` and `'int'` are both just numbers once they have been parsed into the result, but there is no harm.

The name "Virtual" is slightly confusing currently since there are already 'virtual' columns from junction tables and table inheritance. I think the current 'virtual' columns should be renamed to 'internal' columns as they are internal to the ORM, while the new mapped properties should be known as virtual. Alternatively 'computed' or 'mapped' could be used, but I would like to see `@Computed` reserved for putting the select expression directly in the entity in future.

#### 2. `addSelectAndMap(selection: string|qb..., mapToProperty: string)`
Only the selection expression and mapped property are provided, rather than using an alias, since the main intention is to access the computed value at the property and not have to worry about how exactly that happens.

There might be situations where an alias is also useful if the computed value is being used further in other expressions, but I'm not sure what the best approach is for this. Since the virtual property is still defined as a metadata column it should be possible to use it in where, order, etc using its property name, although I have not yet tested this.

The type of the computed value is determined by the annotated property in the entity, unlike with #6855. I think this is a better approach for maintaining type safety, and avoids all the changes to `prepareHydratedValue`. 

#### 3. `EntityMetadata.databaseColumns`
A filtered version of `columns` is added to hold all the non computed columns and only these are used in schema building. This works well, although the naming of all the various column lists is now becoming quite messy, and it feels wrong introducing a new array just because of them.

A possible alternative could be to split up the concept of 'properties' and 'columns' to represent the entity -> database and database -> entity sides. The difficulty is that the new virtual columns mainly care about the property, but they do also need to maintain their `databaseName` for example, so way to split it isn't exactly obvious. Ideally `ColumnMetadata` would extend `PropertyMetadata`, and the appropriate one would be used throughout depending on what is being done.

A list just for virtual would also work, but I think this requires even more changes to the existing code (`[...columns, ...virtual]`, etc).

### Sample
```typescript
@Entity()
export class UserEntity {
    @PrimaryGeneratedColumn()
    id: number;

    @Virtual()
    test: number;
}

const entityManager = connection.createEntityManager();
await entityManager.save(UserEntity, { id: 1 });
const result = await entityManager.createQueryBuilder(UserEntity, 'user')
        .select()
        .addSelectAndMap('5 + 5', 'test')
        .getOne();
```

```sql
SELECT `user`.`id` AS `user_id`, 5 + 5 AS `user_test` FROM `user_entity` `user`
```

```typescript

UserEntity { id: 1, test: 10 }
```

### Feedback
Currently only tested with above sample code so might have issues for insert/update/etc, but I think this is a step in the right direction for `selectAndMap`. Please let me know what you think and send ideas.

### Pull-Request Checklist

### DIFF
--- a/src/decorator/columns/Virtual.ts
+++ b/src/decorator/columns/Virtual.ts
@@ -1,0 +1,100 @@
+import { ColumnOptions, getMetadataArgsStorage } from "../../";
+import {
+    ColumnType, SpatialColumnType
+} from "../../driver/types/ColumnTypes";
+import {ColumnMetadataArgs} from "../../metadata-args/ColumnMetadataArgs";
+import {ColumnTypeUndefinedError} from "../../error/ColumnTypeUndefinedError";
+import { ColumnEnumOptions } from '../options/ColumnEnumOptions';
+import { ColumnHstoreOptions } from '../options/ColumnHstoreOptions';
+import { SpatialColumnOptions } from '../options/SpatialColumnOptions';
+import { VirtualOptions } from '../options/VirtualOptions';
+
+/**
+ * Column decorator is used to mark a specific class property as a table column.
+ * Only properties decorated with this decorator will be persisted to the database when entity be saved.
+ */
+export function Virtual(): PropertyDecorator;
+
+/**
+ * Column decorator is used to mark a specific class property as a table column.
+ * Only properties decorated with this decorator will be persisted to the database when entity be saved.
+ */
+export function Virtual(type: ColumnType): PropertyDecorator;
+
+/**
+ * Column decorator is used to mark a specific class property as a table column.
+ * Only properties decorated with this decorator will be persisted to the database when entity be saved.
+ */
+export function Virtual(type: SpatialColumnType, options?: SpatialColumnOptions): PropertyDecorator;
+
+/**
+ * Column decorator is used to mark a specific class property as a table column.
+ * Only properties decorated with this decorator will be persisted to the database when entity be saved.
+ */
+export function Virtual(type: "enum", options?: ColumnEnumOptions): PropertyDecorator;
+
+/**
+ * Column decorator is used to mark a specific class property as a table column.
+ * Only properties decorated with this decorator will be persisted to the database when entity be saved.
+ */
+export function Virtual(type: "simple-enum", options?: ColumnEnumOptions): PropertyDecorator;
+
+/**
+ * Column decorator is used to mark a specific class property as a table column.
+ * Only properties decorated with this decorator will be persisted to the database when entity be saved.
+ */
+export function Virtual(type: "set", options?: ColumnEnumOptions): PropertyDecorator;
+
+/**
+ * Column decorator is used to mark a specific class property as a table column.
+ * Only properties decorated with this decorator will be persisted to the database when entity be saved.
+ */
+export function Virtual(type: "hstore", options?: ColumnHstoreOptions): PropertyDecorator;
+
+/**
+ * Column decorator is used to mark a specific class property as a table column.
+ * Only properties decorated with this decorator will be persisted to the database when entity be saved.
+ */
+export function Virtual(typeOrOptions?: ColumnType|VirtualOptions, options?: VirtualOptions): PropertyDecorator {
+    return function (object: Object, propertyName: string) {
+
+        // normalize parameters
+        let type: ColumnType|undefined;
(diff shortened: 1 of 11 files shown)


kkoomen:
When can this be implemented? We're waiting for this very urgently.

nathan815:
Glad this is being worked on 🙂 

Any updates on this? I'd be glad to help with this if you need any help.

martinsotirov:
I don’t have any preference for this or #6855 as long as there is a solution for that use case. Hopefully soon.

Btw, wouldn’t it make sense to call the decorator for computed properties `@Computed(…)`?

nebkat:
Wanted to reserve that for providing a query but perhaps one of these could work?

```ts
@Computed("varchar", { expression: `CONCAT(text, "-test")` })

// or

@Expression(`CONCAT(text, "-test")`)
@Computed("varchar")
```

Asarew:
@nebkat @why did you decide to close this PR?

nebkat:
@Asarew Maintainers have stated that they do not intend on merging the refactoring (#7058) necessary for this change and will instead focus on a new version of TypeORM (https://github.com/typeorm/typeorm/tree/typed-repository).

nathan815:
@nebkat Maybe it would be better to merge your improvements into a fork of typeorm. This is such a needed feature to plenty of people. (On a side note: Decorators are a nice way of declaring entities so I really will hate if they remove them. Having the info right next to the field is by far the nicest way of expressing the entity information.)

Yevhen-Molchaniuk:
I need solution for addSelectAndMap without declaring new attributes in the class, maybe anyone can help me?

mickaelmesdocteurs:
@Yevhen-Molchaniuk There is no good solution at this point.
The only way is to use `getRawOne()` or `getRawMany()` and map things as you want.

Please correct me if i'm wrong

vensauro:
> Asarew Maintainers have stated that they do not intend on merging the refactoring (#7058) necessary for this change and will instead focus on a new version of TypeORM ([https://github.com/typeorm/typeorm/tree/typed-repository](https://github.com/typeorm/typeorm/tree/typed-repository?rgh-link-date=2021-04-29T13%3A42%3A28Z)).

hi @nebkat, you can make an fork of typeorm with your changes?
i am thinking in migrate to mikro-orm, that has more evoluted than typeorm in the actual momento, but maybe your changes can do the work 🤩

frct1:
Hey
This feature actually should be prepared to be in master branch
Such a basic thing but not available to use , that's a bummer :(

newerton:
Please, more attention to this type of PR.

Issue #1090, opened 2017-10-24
upsert with return functionality


EvaLok:
proposing the following functionality (essentially `insert... on conflict.. returning..`):

```typescript
const result = model.upsert({
    id: 123,
    someValue: "test"
}, {
    returnAll: true
});

console.log(result); // outputs entire updated model, including preexisting fields, auto-updated fields

const result = model.upsert({
    id: 123,
    someValue: "test"
});

console.log(result); // outputs some indicator as to whether or not an insert or an update occurred

const result = model.upsert({
    id: 123,
    someValue: "test"
}, {
    returnPartial: [
        'id',
        'someValue',
        'someOtherValue'
    ]
);

console.log(result); // outputs a partial object with specified fields loaded
```

at least in postgres, this is possible in a single operation, but it's not currently possible in a single operation with TypeORM (example not tested, just for showing a current rough solution..):

```typescript
import { Column, PrimaryColumn, Repository, UpdateDateColumn } from "typeorm";
import { Entity } from "typeorm/decorator/entity/Entity";

export class SomeServiceClass {
    constructor(
        private repo: Repository<SomeAggregate>
    ) {}

    public async handleSomeEventWithTypeORMMethods( event: ISomeEvent ): Promise<SomeAggregate>  {
        let model = await this.repo.findOne({
            id: event.id,
        });

        if ( ! model ) {
            model = new SomeAggregate();
        }

        model.valueToUpdate = event.valueToUpdate;

        await this.repo.persist(model);

        const updatedModel = await this.repo.findOne({
            id: event.id,
        });

        return updatedModel;
    }

    public async handleSomeEventWithRawQuery( event: ISomeEvent ): Promise<SomeAggregate>  {
        const results = await this.repo.query(`
            INSERT INTO
                some_table
            (
                "id",
                "valueToUpdate"
            )
                VALUES
            (
                $1,
                $2
            )

            ON CONFLICT( "id" )
                DO UPDATE
                    SET
                        "valueToUpdate" = $2
                    WHERE
                        "some_table"."id" = $1

            RETURNING *
        `, [
            event.id,
            event.valueToUpdate
        ]);

        return results[0];
    }
}


@Entity()
export class SomeAggregate {
    @PrimaryColumn("uuid")
    public id: string;

    @Column("text")
    public valueToUpdate ?: string;

    @Column("text")
    public anotherValue ?: string;

    @UpdateDateColumn({type: "timestamp with time zone"})
[...]

PR #7029, opened 2020-11-07
feat: upsert/query-builder insert or update all columns

nebkat:
Improves `InsertQueryBuilder.orUpdate` to allow all columns being inserted to be automatically listed in the `ON CONFLICT DO UPDATE` / `ON DUPLICATE KEY UPDATE` clauses (postgre/mysql).

`EntityManager.upsert` with this enabled.

Fixes: #1090

### Pull-Request Checklist

### DIFF
--- a/docs/entity-manager-api.md
+++ b/docs/entity-manager-api.md
@@ -123,18 +123,40 @@ await manager.remove([
 * `insert` - Inserts a new entity, or array of entities.
 
 ```typescript
-await manager.insert(User, { 
-    firstName: "Timber", 
-    lastName: "Timber" 
+await manager.insert(User, {
+    firstName: "Timber",
+    lastName: "Timber"
 });
 
-await manager.insert(User, [{ 
-    firstName: "Foo", 
-    lastName: "Bar" 
-}, { 
-    firstName: "Rizz", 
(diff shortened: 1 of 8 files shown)


nebkat:
For now the requirement to provide conflict columns to `EntityManager.upsert` for Postgres is unideal, it would be better if they could be chosen automatically:
- Should it be based on the inputted columns?
- Is there a reason why one wouldn't want all of the entity's primary/unique columns listed?
- What is the expected behavior with composite keys/indices (what are the 'expressions' mentioned in the syntax)?
- Can a non-unique column be used as the conflict target?

I've only used MySQL where no conflict target is needed so I'm not sure how it should work, would be good to get some input from Postgres/SQLite users. For MySQL at least this works perfectly.

Additionally I'd like to add an extra argument for setting columns to specific values if a conflict occurs (e.g. count = count + 1) which should also be a `QueryDeepPartialEntity<Entity>`. Currently `InsertQueryBuilder` and `UpdateQueryBuilder` have a lot of duplicate code and are somewhat difficult to follow so they could do with some refactoring before this gets added.

nebkat:
Replaced by #7076

Issue #6997, opened 2020-11-01
Typeorm entities not working with javascript


theabner:
I'm trying to create a project with typeorm with javascript, a connection works correctly, but typeorm can't find my entities.
I am following the example provided by the type documentation: https://typeorm.io/#/usage-with-javascript

app.js:
```
typeorm.createConnection({
  type: 'mysql',
  host: process.env.MYSQL_IP, // Docker Env
  port: process.env.MYSQL_INTERNAL_PORT,
  username: process.env.MYSQL_ROOT_USERNAME,
  password: process.env.MYSQL_ROOT_PASSWORD,
  database: process.env.MYSQL_DATABASE,
  synchronize: true,
  entities: [
    require('./Model/TestSchema.js'),
  ],
}).then((connection) => {
  const app = express();

  app.use(cors(CorsConfigs));
  app.use(bodyParser.json());
  app.use(bodyParser.urlencoded({ extended: false }));
  app.use(authentication);
  app.use(router);

  const testRepository = connection.getRepository('Category');
  const teste = { name: 'TypeScript' };

  app.listen(process.env.BACKEND_PORT, () => {
    console.log(`Rodando na porta: ${process.env.BACKEND_PORT.toString()}`);
  });
}).catch((error) => {
  console.log('Error: ', error);
});
```

Model/TestSchema.js:
```
module.exports = {
  name: 'Category',
  columns: {
    id: {
      primary: true,
      type: 'int',
      generated: true,
    },
    name: {
      type: 'string',
    },
  },
};
```
the connection is working correctly, the problem and when I run it shows me:

Error: RepositoryNotFoundError: No repository for "Category" was found. Looks like this entity is not registered in current "default" connection?

at new RepositoryNotFoundError (/app/src/error/RepositoryNotFoundError.ts:11:9)

at EntityManager.getRepository (/app/src/entity-manager/EntityManager.ts:919:19)

at Connection.getRepository (/app/src/connection/Connection.ts:346:29)

at /app/src/app.js:32:37

at processTicksAndRejections (internal/process/task_queues.js:93:5)

edcolvin:
I think the documentation is incorrect. You need to pass EntitySchema objects in the entities array. 

```javascript
const typeorm = require("typeorm")
const EntitySchema = typeorm.EntitySchema
```

and then in the options passed to createConnection:
```javascript
entities: [
        new EntitySchema(require("./entity/Post")),
        new EntitySchema(require("./entity/Category"))
    ]
[...]

PR #7031, opened 2020-11-07
docs: fix javascript usage examples

edcolvin:
### Description of change
Updated Javascript usage example in docs/usage-with-javascript.md and docs/zh_CN/usage-with-javascript.md to pass EntitySchema objects in the ConnectionOptions instead of plain objects and changed type:"string" in the schemas to type:"varchar". Also updated docs/zh_CN/usage-with-javascript.md to use "entities" instead of "entitySchemas" in the ConnectionOptions.

### DIFF
--- a/docs/usage-with-javascript.md
+++ b/docs/usage-with-javascript.md
@@ -58,7 +58,9 @@ typeorm.createConnection({
 ##### entity/Category.js
 
 ```typescript
-module.exports = {
+var EntitySchema = require("typeorm").EntitySchema;
+
+module.exports = new EntitySchema({
     name: "Category", // Will use table name `category` as default behaviour.
     tableName: "categories", // Optional: Provide `tableName` property to override the default behaviour for table name. 
     columns: {
@@ -68,16 +70,18 @@ module.exports = {
             generated: true
         },
         name: {
-            type: "string"
+            type: "varchar"
         }
     }
-};
+});
 ```
 
 ##### entity/Post.js
 
 ```typescript
-module.exports = {
+var EntitySchema = require("typeorm").EntitySchema;
+
+module.exports = new EntitySchema({
     name: "Post", // Will use table name `post` as default behaviour.
     tableName: "posts", // Optional: Provide `tableName` property to override the default behaviour for table name. 
     columns: {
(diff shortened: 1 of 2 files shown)


edcolvin:
I saw issue #6997 where that reported the example code wasn't working.

docs/usage-with-javascript.md
```diff
@@ -17,8 +18,8 @@ typeorm.createConnection({
     database: "test",
     synchronize: true,
     entities: [
-        require("./entity/Post"),
-        require("./entity/Category")
+        new EntitySchema(require("./entity/Post")),
+        new EntitySchema(require("./entity/Category"))
```
imnotjames, line 22: Are you sure this is necessary?

edcolvin: Yes. 

The items in the entities array should be either a constructor function with ```@Entity()``` metadata, or an instanceof ```EntitySchema```, or string that is a file glob path to modules that export one of those things. Since this is an example using plain javascript, it needs to be an EntitySchema. This is the same technique for defining entities as described in [Separating Entity Definition](https://github.com/typeorm/typeorm/blob/master/docs/separating-entity-definition.md).

If if you try to run the example code as shown, it will throw a RepositoryNotFoundError. The code in the example repo linked at the bottom of this document works correctly. This updates the example code on the document itself to match the code in the repo.

imnotjames: Right, didn't read the whole example.  Sorry :)

Can we instead create the entity schemas in the entity objects instead?

edcolvin: OK, I like that better. I think the example repo does it in the entities array for importing from JSON, but that's not the case here. I will revise the pull request. Thanks!

edcolvin: So I used ```var EntitySchema=require("typeorm").EntitySchema``` in the schema files. Is that ok?

(I assume it's using var for compatibility with ES5, but I don't know if we really need to do that anymore.)

edcolvin:
Anything else I can do for this?

pleerock:
Thank you for contribution! 🎉

Issue #6993, opened 2020-10-30
Implement optional Domain Entity Mapper for better Object Oriented design patterns


Tomaszal:
### The Problem
Right now all Entity Properties must be public for TypeORM to work, which prevents the use of TypeORM Entities as Domain Entities for DDD architecture in a concise way. Of course it is still possible to use it by separating the Infrastructure Entity definitions from Domain Entity ones, however I think TypeORM has the potential of simplifying that a lot by combining them. As stated by some others in previous raised issues,
> In the end, what I'm looking for is not just a way to do encapsulation - there are plenty of those - but a way to have both encapsulation AND the ease of use/productivity of TypeORM.

### The Solution
I suggest to add a property to Entity Options called Domain Entity Mapper (or something like that, name could change). From DDD perspective, that would just be a Factory function which would create a Domain Entity out of the Raw Entity Object returned by TypeORM. From TypeORM perspective, it would allow to map the Persistence Entities to Domain Entities without needing to use any setters. Here is a very simple example to show what I mean:

```ts
interface IPersonProps {
    id: number;
    firstName: string;
    lastName: string;
}

interface IPersonMap {
    id: number;
    fullName: string;
}

@Entity({ domainEntityMapper: Person.createFromMap })
export class Post {
  private readonly props: IPersonProps;
    
  @PrimaryColumn()
  get id(): number {
    return this.props.id;
  }

  @Column()
  get fullName(): string {
    return this.props.firstName + " " + this.props.lastName;
  }

  get firstName(): string {
    return this.props.firstName;
  }

  get lastName(): string {
    return this.props.lastName;
  }

  private constructor(props: IPersonProps) {
    this.props = props;
  }

  static createFromMap(map: IPersonMap): Person {
    const names = map.fullName.split(" ");
    const firstName = names[0];
    const lastName = names[names.length - 1];

    return new Person({
      id: map.id,
      firstName,
      lastName,
    });
  }
}
```

This is obviously not a real world example (at least not a good one), but the point is it keeps the props immutable to outside of the Domain. If you try doing this right now, it will obviously fail to set props of the Entity, as they do not have any setters. However, if a user specifies a Domain Entity Mapper, TypeORM will use that to create an Entity, instead of trying to use setters directly.

This should be fairly simple to implement by modifying the Query Builder Transformers (`src/query-builder/transformer`). In fact I already have the SQL one working as a proof of concept on a local fork. I would be up to implement it properly and across other database providers if I get a green light.

[...]

PR #7036, opened 2020-11-08
feat: Implement Domain Entity Mapper

Tomaszal:
### Description of change

Domain Entity Mapper on Entities and Embedded Columns allows for more advanced object-oriented patterns (such as Domain Driven Design) by allowing to have Columns without setters. A full explanation can be found the issue that this closes.

In addition to what was described in the issue, I had to add a Domain Entity Mapper property to Embedded Column Options. This works, however perhaps it would be a good idea to add a new decorator for Embedded Entities. It could be used to specify Embedded Column Options globally, instead of in each entity where it is used. However I think that is something for a separate PR if we agree on this.

This requires documentation update, however I will do that after changes are implemented (if any are needed).

Closes: #6993

### Pull-Request Checklist

### DIFF
--- a/src/decorator/columns/Column.ts
+++ b/src/decorator/columns/Column.ts
@@ -133,7 +133,8 @@ export function Column(typeOrOptions?: ((type?: any) => Function)|ColumnType|(Co
                 propertyName: propertyName,
                 isArray: reflectMetadataType === Array || options.array === true,
                 prefix: options.prefix !== undefined ? options.prefix : undefined,
-                type: typeOrOptions as (type?: any) => Function
+                type: typeOrOptions as (type?: any) => Function,
+                domainEntityMapper: options.domainEntityMapper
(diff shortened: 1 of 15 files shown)


nebkat:
I worry that if this is added it will make it quite difficult to make future changes to methods such as `ColumnMetadata.getEntityValue()`/`setEntityValue()`, and everywhere else that has been modified really.

Is the main purpose of this to allow TypeORM internally to set values while external users cannot? Or is the fact that it runs through a constructor (as in some of the examples) the goal? I believe the former could be achieved with TypeScript alone, just by exporting a public facing version of the entities that has all of its properties set to `readonly`.

pleerock:
Thank you for your efforts!

But I don't think we need this functionality. I completely understand what do you want, but same goals can be achieved using different approaches. For example you can use entity schemas and completely separate database schema out of types and have your implementation based on a types.

PR #5253, opened 2019-12-26
Fix query-builder error for pessimistic_write locks with left joins

garlou:
Fix for problem described here: https://github.com/typeorm/typeorm/issues/4084
Usage: 
```
.createQueryBuilder(User, 'user')
  .setLock('pessimistic_write', 'user')
  .leftJoinAndSelect('user.someAnotherTable', 'someAnotherTable')
  .where('email LIKE :email', { email: 'email@mail.com' })
  .getOne();```

PR #7040, opened 2020-11-08
feat: implement "FOR UPDATE OF" for postgres driver

### LINKED ISSUE #4084: QueryBuilder with leftJoinAndSelect and pessimistic_write lock (postgres)
**Issue type:**

[ ] question
[ ] bug report
[x] feature request
[ ] documentation issue

**Database system/driver:**

[ ] `cordova`
[ ] `mongodb`
[ ] `mssql`
[ ] `mysql` / `mariadb`
[ ] `oracle`
[x] `postgres`
[ ] `cockroachdb`
[ ] `sqlite`
[ ] `sqljs`
[ ] `react-native`
[ ] `expo`

**TypeORM version:**

[ ] `latest`
[ ] `@next`

```typescript
.createQueryBuilder(User, 'user')
  .setLock('pessimistic_write')
  .leftJoinAndSelect('user.someAnotherTable', 'someAnotherTable')
  .where('email LIKE :email', { email: 'email@mail.com' })
  .getOne();
```

Code above generates query: 
```sql
SELECT "user"."id" AS "user_id",
       "someAnotherTable"."id" AS "someAnotherTable_id"
FROM "public"."users" "user"
LEFT JOIN "public"."some_another_table" "someAnotherTable" ON "someAnotherTable"."id"="user"."some_another_table_id"
WHERE email LIKE 'email@mail.com'
  FOR UPDATE
```

Which produce postgresql error:  

>ERROR:  FOR UPDATE cannot be applied to the nullable side of an outer join

Looks like DBMS trying to lock someAnotherTable as well, but typescript code above does not imply that every joined table should be locked (isn't it?).  
If so, generated sql code should include part `FOR UPDATE OF <>` instead of  `FOR UPDATE`:  

```sql
SELECT "user"."id" AS "user_id",
       "someAnotherTable"."id" AS "someAnotherTable_id"
FROM "public"."users" "user"
LEFT JOIN "public"."some_another_table" "someAnotherTable" ON "someAnotherTable"."id"="user"."some_another_table_id"
WHERE email LIKE 'email@mail.com'
  FOR UPDATE OF "user"
```
similar issue:
https://code.djangoproject.com/ticket/28010

ejose19:
### Description of change

Fixes #4084

This PR implements some parts of the unmerged #5253 PR, improving it to allow specifying multiple tables and checking whether the referenced tables exists in the query.

Also, there are some improvements to related methods JSDocs/Type definitions, removing invalid ```where``` space, and adding missing lock mode to ```FindOptions```.

**Note for whoever want to implement this for other drivers**
- Oracle uses a different syntax, specifying ```table.column``` or ```column``` instead of ```table``` after the ```OF``` clause, so an special treatment needs to be done when oracle is the driver. [source1](https://docs.oracle.com/en/database/oracle/oracle-database/19/sqlrf/SELECT.html#GUID-CFA006CA-6FF1-4972-821E-6996142A51C6) [source2](https://docs.oracle.com/en/database/oracle/oracle-database/19/sqlrf/img_text/for_update_clause.html)

- MySQL / Aurora do indeed support this syntax in their latest versions, but tests are using an outdated version of them, so a new updated image is needed for properly testing the feature. [source](https://dev.mysql.com/doc/refman/8.0/en/select.html)

- MySQL / Aurora in their latest version does not need to use a different ```if``` for ```return " LOCK IN SHARE MODE";```, as they support the ```FOR SHARE``` syntax. *same source as above

- MySQL / Aurora / Oracle supports ```SKIP LOCKED``` and ```NOWAIT```, so when new tests images are added they can be included in the corresponding ```if```. *same sources as above

### Pull-Request Checklist


AlexMesser:
thank you for contribution!

src/query-builder/SelectQueryBuilder.ts
```diff
@@ -960,26 +960,21 @@ export class SelectQueryBuilder<Entity> extends QueryBuilder<Entity> implements
...
      * Sets locking mode.
      */
-    setLock(lockMode: "optimistic", lockVersion: number): this;
+    setLock(lockMode: "optimistic", lockVersion: number | Date): this;
 
     /**
      * Sets locking mode.
      */
-    setLock(lockMode: "optimistic", lockVersion: Date): this;
+    setLock(lockMode: "pessimistic_read"|"pessimistic_write"|"dirty_read"|"pessimistic_partial_write"|"pessimistic_write_or_fail"|"for_no_key_update", lockVersion?: undefined, lockTables?: string[]): this;
```
ekifox, line 968: can you add a version of setLock without `lockVersion` param?

ekifox:
This code:
```ts
        const inventory = await this.inventoryRepository
            .createQueryBuilder('inv')
            .select('inv.id')
            .innerJoinAndSelect('inv.item', 'item')
            .andWhere('inv.id IN (:...ids)', {
                ids: inventoryIds
            })
            .andWhere('inv.userId = :id', {
                id: user.id
            })
            .andWhere('inv.lockedBy IS NULL')
            .setLock('pessimistic_write', null, ['inv'])
            .getMany()
```
produces:

QueryFailedError: invalid input syntax for type integer: "NaN"
    at new QueryFailedError (/root/crash/src/error/QueryFailedError.ts:9:9)
    at PostgresQueryRunner.<anonymous> (/root/crash/src/driver/postgres/PostgresQueryRunner.ts:228:19)
    at step (/root/crash/node_modules/tslib/tslib.js:143:27)
    at Object.throw (/root/crash/node_modules/tslib/tslib.js:124:57)
    at rejected (/root/crash/node_modules/tslib/tslib.js:115:69)
    at processTicksAndRejections (internal/process/task_queues.js:97:5)

ejose19:
You should use `typescript`, `null` is not a valid value to be passed as 2nd argument. It gives an error: `Argument of type 'null' is not assignable to parameter of type 'undefined'.ts(2345)`

The way the original `setLock` is designed, you need to pass `undefined` as 2nd argument if you want to use `lockTables`, changing it would be a breaking change and one that wouldn't have been merged into this major release of typeorm.

ekifox:
> You should use `typescript`, `null` is not a valid value to be passed as 2nd argument. It gives an error: `Argument of type 'null' is not assignable to parameter of type 'undefined'.ts(2345)`
> 
> The way the original `setLock` is designed, you need to pass `undefined` as 2nd argument if you want to use `lockTables`, changing it would be a breaking change and one that wouldn't have been merged into this major release of typeorm.

I have tried undefined before null. Anyway, I found a bug, it's on my side, thanks for this feature!

PR #7042, opened 2020-11-09
fix: handle overlapping property / database names in querybuilder

### LINKED ISSUE #7030: Column name aliasing broke in v 0.2.29
~~It looks like bug fix #6870 caused this regression. ~~ Edit from @imnotjames: It seems to be #4760

If you have a table with Primary key and another column with the same name, then the Primary ID name gets precedence.  0.2.28 was working correctly.

## Issue Description

We have a table for example called Book which had a DB primary key in the db. But we don't want to expose this to the users so we have created another column Book_id which has a UUID which we use in our API

```
@Entity({name: 'website_book'})
export class Book {
    @PrimaryGeneratedColumn({
        type: 'integer',
        name: 'id',     <========= 0.2.29 behavior 
    })
    oldBookId: number;

    @Column('int', {
        nullable: false,
        name: 'user_id',
    })
    user_id: number;

    @Column('character varying', {
        nullable: false,
        unique: true,
        length: 38,
        name: 'book_id',
    })
    id: string;    <========= 0.2.28 behavior 
}
```

### Expected Behavior

" Book.id " should be mapped to (book_id column in the database)

### Actual Behavior

" Book.id " is mapped to (id column in the database)

### My Environment

Not os or env dependent issue

| TypeORM version  : 0.2.29

- [ ] `aurora-data-api`
- [ ] `aurora-data-api-pg`
- [ ] `better-sqlite3`
- [ ] `cockroachdb`
- [ ] `cordova`
- [ ] `expo`
- [ ] `mongodb`
- [ ] `mysql`
- [ ] `nativescript`
- [ ] `oracle`
- [X] `postgres`
- [ ] `react-native`
- [ ] `sap`
- [ ] `sqlite`
- [ ] `sqlite-abstract`
- [ ] `sqljs`
- [ ] `sqlserver`

I can help test the resolution

imnotjames:
### Description of change

we should always prioritize replacement of any property path first,
then property names, then database names, then relation values
whenever there's ambiguous values

fixes #7030

### Pull-Request Checklist

### DIFF
--- a/src/query-builder/QueryBuilder.ts
+++ b/src/query-builder/QueryBuilder.ts
@@ -587,24 +587,37 @@ export abstract class QueryBuilder<Entity> {
 
             const replacements: { [key: string]: string } = {};
 
-            for (const column of alias.metadata.columns) {
-                if (!(column.propertyPath in replacements))
-                    replacements[column.propertyPath] = column.databaseName;
-                if (!(column.propertyName in replacements))
-                    replacements[column.propertyName] = column.databaseName;
-                if (!(column.databaseName in replacements))
-                    replacements[column.databaseName] = column.databaseName;
+            // Insert & overwrite the replacements from least to most relevant in our replacements object.
+            // To do this we iterate and overwrite in the order of relevance.
+            // Least to Most Relevant:
+            // * Relation Property Path to first join column key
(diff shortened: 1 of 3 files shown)


ddehghan:
Yeah. this fixed my issue.

Issue #7106, opened 2020-11-25
PostgreSQL sequence identifier length error


matthewberryman:
### Expected Behaviour

Generated sequence name should be < `NAMEDATALEN` bytes, where `NAMEDATALEN`=64 by default.

### Actual Behaviour

There is no check in https://github.com/typeorm/typeorm/blob/c4a36da62593469436b074873eba186f0f8b990d/src/driver/postgres/PostgresQueryRunner.ts#L2055 or https://github.com/typeorm/typeorm/blob/c4a36da62593469436b074873eba186f0f8b990d/src/driver/postgres/PostgresQueryRunner.ts#L2058 that the generated sequence length fits in the default limit for identifiers. A second schema sync will throw an unhandled error, e.g.:
```
Error during schema synchronization:
QueryFailedError: relation "soil_classification_order_sub_orders_soilclassificationordersub" does not exist
    at new QueryFailedError (/Users/matthew/code/CBB/Nepal/API/src/error/QueryFailedError.ts:9:9)
    at Query.callback (/Users/matthew/code/CBB/Nepal/API/src/driver/postgres/PostgresQueryRunner.ts:220:30)
    at Query.handleError (/Users/matthew/code/CBB/Nepal/API/node_modules/pg/lib/query.js:139:19)
    at Client._handleErrorMessage (/Users/matthew/code/CBB/Nepal/API/node_modules/pg/lib/client.js:326:17)
    at Connection.emit (node:events:329:20)
    at Connection.EventEmitter.emit (node:domain:467:12)
    at /Users/matthew/code/CBB/Nepal/API/node_modules/pg/lib/connection.js:109:12
    at Parser.parse (/Users/matthew/code/CBB/Nepal/API/node_modules/pg-protocol/src/parser.ts:102:9)
    at TLSSocket.<anonymous> (/Users/matthew/code/CBB/Nepal/API/node_modules/pg-protocol/src/index.ts:7:48)
    at TLSSocket.emit (node:events:329:20) {
  length: 161,
  severity: 'ERROR',
  code: '42P01',
  detail: undefined,
  hint: undefined,
  position: undefined,
  internalPosition: undefined,
  internalQuery: undefined,
  where: undefined,
  schema: undefined,
  table: undefined,
  column: undefined,
  dataType: undefined,
  constraint: undefined,
  file: 'namespace.c',
  line: '423',
  routine: 'RangeVarGetRelidExtended',
[...]

PR #7115, opened 2020-11-27
fix: resolves Postgres sequence identifier length error

matthewberryman:
Changes the sequence name generator to handle cases where PostgreSQL auto-generated sequence name differed from TypeORM generated sequence name.

Closes: #7106

### Description of change

Current behaviour: sequence name generator for PostgreSQL does not match PostgreSQL automatic sequence name  implementation for some cases.
New behaviour: now it does. 

### Pull-Request Checklist

### DIFF
--- a/src/driver/postgres/PostgresQueryRunner.ts
+++ b/src/driver/postgres/PostgresQueryRunner.ts
@@ -2051,11 +2051,14 @@ export class PostgresQueryRunner extends BaseQueryRunner implements QueryRunner 
             tableName = table.name.split(".")[1];
         }
 
+        let seqName = `${tableName}_${columnName}_seq`;
+        if (seqName.length > this.connection.driver.maxAliasLength!) // note doesn't yet handle corner cases where .length differs from number of UTF-8 bytes
+            seqName=`${tableName.substring(0,29)}_${columnName.substring(0,Math.max(29,63-tableName.length-5))}_seq`;
+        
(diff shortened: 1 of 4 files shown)


src/driver/postgres/PostgresQueryRunner.ts
```diff
@@ -2051,11 +2051,14 @@ export class PostgresQueryRunner extends BaseQueryRunner implements QueryRunner
             tableName = table.name.split(".")[1];
         }
 
+        let seqName = `${tableName}_${columnName}_seq`;
+        if (seqName.length > 63) // note doesn't yet handle corner cases where .length differs from number of UTF-8 bytes
```
pleerock, line 2055: instead of hard-coded magic value `63` I would recommend to use `Driver.maxAliasLength` property.

pleerock: I think in all other places we already use a `shorten` function to shorten things... Can we use it here as well to achieve consistency?

matthewberryman: I have pushed an amended commit that uses the `maxAliasLength` property. I have not used `shorten` because it doesn't match the truncation algorithm that Postgres uses, thus won't fix the issue.

matthewberryman:
In the failing circleci test, it does pass the additional test, however there was a segfault that occurred later for some reason unrelated to the latest change.

pleerock:
Thank you for contribution! 🎉

PR #7058, opened 2020-11-11
refactor: Query builders, drivers, entity/column/relation metadata, decorators

nebkat:
# Description of change

Major refactor of query builders and other parts of code, big improvements in readability and lots of duplicate code removed.

### Pull-Request Checklist

- [x] Code is up-to-date with the `master` branch
- [x] `npm run lint` passes with this change
- [x] `npm run test` passes with this change
- [x] This pull request links relevant issues as `Fixes #0000`
- [x] There are new or updated unit tests validating the change
- [x] Documentation has been updated to reflect this change
- [x] The new commits follow conventions explained in [CONTRIBUTING.md](https://github.com/typeorm/typeorm/blob/master/CONTRIBUTING.md)

PR #7119, opened 2020-11-27
feat: add index options to many-to-one relation decorator

### LINKED ISSUE #7118: Add index option to ManyToOne decorator
## Feature Description

Add a quality-of-life indexing option to ManyToOne decorator.

### The Problem

Not all RDBMSs auto-index the referencing-side of FKs (postgresql doesn't do it), and very frequently there are benefits of doing so. Since this ought to be a common idiom it makes sense for the ManyToOne decorator to accept an option which tell it to also create an index for the FK.

### The Solution

Add a `index?: boolean | IndexOptions & { name?: string }` property to ManyToOneRelations(a new interface which extends `RelationOptions`) which the decorator reads and creates the index metadata if it is set.

- [ ] `aurora-data-api`
- [ ] `aurora-data-api-pg`
- [ ] `better-sqlite3`
- [ ] `cockroachdb`
- [ ] `cordova`
- [ ] `expo`
- [ ] `mongodb`
- [ ] `mysql`
- [ ] `nativescript`
- [ ] `oracle`
- [x] `postgres`
- [ ] `react-native`
- [ ] `sap`
- [ ] `sqlite`
- [ ] `sqlite-abstract`
- [ ] `sqljs`
- [ ] `sqlserver`

pedrolcn:
### Description of change

Adds option to create index on the referencing side of Many-To-One relations.  
Closes #7118 

### Pull-Request Checklist

### DIFF
--- a/src/decorator/options/ManyToOneOptions.ts
+++ b/src/decorator/options/ManyToOneOptions.ts
@@ -1,0 +1,12 @@
+import {IndexOptions} from './IndexOptions';
+import {RelationOptions} from './RelationOptions';
+
+/**
+ * Describes a many-to-one relation's options
+ */
+export interface ManyToOneOptions extends RelationOptions {
+  /**
+   * Create an index on the referencing column of the many-to-one relation.
+   */
+  index?: boolean | IndexOptions & { name?: string };
+}
--- a/src/decorator/relations/ManyToOne.ts
+++ b/src/decorator/relations/ManyToOne.ts
@@ -1,12 +1,12 @@
-import {getMetadataArgsStorage, ObjectType, RelationOptions} from "../../";
+import {getMetadataArgsStorage, IndexOptions, ObjectType, RelationOptions, ManyToOneOptions} from "../../";
 import {RelationMetadataArgs} from "../../metadata-args/RelationMetadataArgs";
-
+import {Index} from '../Index';
 /**
  * Many-to-one relation allows to create type of relation when Entity1 can have single instance of Entity2, but
  * Entity2 can have a multiple instances of Entity1. Entity1 is an owner of the relationship, and storages Entity2 id
  * on its own side.
  */
-export function ManyToOne<T>(typeFunctionOrTarget: string|((type?: any) => ObjectType<T>), options?: RelationOptions): PropertyDecorator;
+export function ManyToOne<T>(typeFunctionOrTarget: string|((type?: any) => ObjectType<T>), options?: ManyToOneOptions): PropertyDecorator;
 
 /**
  * Many-to-one relation allows to create type of relation when Entity1 can have single instance of Entity2, but
@@ -15,7 +15,7 @@ export function ManyToOne<T>(typeFunctionOrTarget: string|((type?: any) => Objec
  */
 export function ManyToOne<T>(typeFunctionOrTarget: string|((type?: any) => ObjectType<T>),
                              inverseSide?: string|((object: T) => any),
-                             options?: RelationOptions): PropertyDecorator;
+                             options?: ManyToOneOptions): PropertyDecorator;
 
 /**
  * Many-to-one relation allows to create type of relation when Entity1 can have single instance of Entity2, but
@@ -24,7 +24,7 @@ export function ManyToOne<T>(typeFunctionOrTarget: string|((type?: any) => Objec
  */
 export function ManyToOne<T>(typeFunctionOrTarget: string|((type?: any) => ObjectType<T>),
                              inverseSideOrOptions?: string|((object: T) => any)|RelationOptions,
-                             options?: RelationOptions): PropertyDecorator {
+                             options?: ManyToOneOptions): PropertyDecorator {
 
     // normalize parameters
     let inverseSideProperty: string|((object: T) => any);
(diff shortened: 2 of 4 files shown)


src/decorator/options/ManyToOneOptions.ts
```diff
@@ -0,0 +1,13 @@
+import { IndexOptions } from './IndexOptions';
+import { RelationOptions } from './RelationOptions';
+
+
```
gabrielsanttana, line 4: Here you could remove this white space:

```suggestion
```

src/decorator/relations/ManyToOne.ts
```diff
@@ -55,5 +55,24 @@ export function ManyToOne<T>(typeFunctionOrTarget: string|((type?: any) => Objec
...
+                synchronize: indexOptions && (indexOptions as { synchronize: false }).synchronize === false ? false : true,
+                where: indexOptions ? indexOptions.where : undefined,
+                unique: indexOptions && indexOptions.unique ? true : false,
+                spatial: indexOptions && indexOptions.spatial ? true : false,
+                fulltext: indexOptions && indexOptions.fulltext ? true : false,
+                parser: indexOptions ? indexOptions.parser : undefined,
+                sparse: indexOptions && indexOptions.sparse ? true : false,
+                background: indexOptions && indexOptions.background ? true : false,
+                expireAfterSeconds: indexOptions ? indexOptions.expireAfterSeconds : undefined
+            } as IndexMetadataArgs);
```
nebkat, line 75: I suggest you reuse the existing `Index()` decorator here so we don't duplicate.
```suggestion
            Index(indexOptions)(object, propertyName);
```

nebkat (changes requested):

src/decorator/relations/ManyToOne.ts
pedrolcn: Done. Also added logic for handling custom index names without code duplication.

src/decorator/options/ManyToOneOptions.ts
pedrolcn: Done

nebkat:
Looks good, last thing just squash these 4 commits into one for less git spam :tada:

pedrolcn:
> Looks good, last thing just squash these 4 commits into one for less git spam 🎉

Done!

pedrolcn:
@nebkat 

This has been open for a few days, is there any action from me that is still needed before this can be merged ?

nebkat (approved):
@pedrolcn I am not a maintainer here, just helping out. There haven't been any merges in over a month so just a matter of waiting now. Approved by me anyway!

pedrolcn:
> @pedrolcn I am not a maintainer here, just helping out. There haven't been any merges in over a month so just a matter of waiting now. Approved by me anyway!

Oh, I didn't realize that, I see
Thanks for the response!

pleerock:
Thank you for your efforts!

However, why do we need this feature when we already have a `Index` decorator?
What you are suggesting is to introduce this:

```ts
@ManyToOne(() => Type, type => type.property, {
   option1: ... 
})
```

when we already have:

```ts
@Index({ option1: ... })
@ManyToOne(() => Type, type => type.property)
```

Don't see much of point. Next step would be to merge `@JoinColumn` into this decorator or even worse `@JoinTable`.

From my experience this kind of features only confuses people and complicates the code because of new flows in the code, e.g.:

* what if I put both `@Index({ someOptions })` and `@Index({ someOptions }) @ManyToOne({ index: otherOptions })`?
* user can ask what is the difference between putting index into `many-to-one` and separate index, needs additional documentation

nebkat:
But couldn't we then say the same for `@Column({ generated: true, unique: true, primary: true})` and every combination of `@Generated()`  `@Unique()` `@PrimaryColumn()` `@PrimaryGeneratedColumn()`? In both cases it's a choice between being explicit or concise, or even ultimately a matter of style. Personally I prefer to use `@Unique()`/`@Index()` only for multi-column indices.

If you really want to solve the user confusion with these sort of things I think we need to move towards separate documentation and auto generated API reference. The user should be able to look up `ColumnOptions` and see that `generated: true` is the same as `Generated()` and vice-versa, rather than someone having to manually document the little details (of which there are many) in `docs/`. It also wouldn't be all that difficult to have all the `MetadataArgsStorage` additions check if there are duplicate definitions and throw a warning or error.

----

What I've tried to do here https://github.com/typeorm/typeorm/pull/7058/commits/b032f6b0ef9be87fbd4ad7813a382519f9ada11a https://github.com/typeorm/typeorm/pull/7058/commits/63c9bd95eb82e5c55e73c06b392e0c520a519c1d https://github.com/typeorm/typeorm/pull/7058/commits/b321c020be76f1a90647728e5c939830a66474ef is at least have all the duplicate functionality use the same code by calling decorators within decorators. As long as this is the case I don't see any harm in having both `@Column({option: ...})` and `@Option`.

----

Also slight correction just in case you misread rather than made a typo, it would be:

```ts
@ManyToOne(() => Type, type => type.property, {
   index: {option1: ... }
})
```

Rather than:

```ts
@ManyToOne(() => Type, type => type.property, {
   option1: ... 
})
```

Which is a bit more obvious as to what is happening.

pedrolcn:
@pleerock 
The rationale is that indexing a FK is a common enough idiom (and most likely a best practice) that it should be encouraged by the API design as well something that should be very ergonomic to do. But as the linked issue mentions, this is mostly a quality-of-life improvement, it is not by any chance something that was not possible before.

Although, I do agree that "many ways of doing the same thing" is a problem and one that TypeORM definitely suffers from (sometimes with subtle differences between them) and it can confuse users. I myself haven't stopped to consider what happens if the user does something like
```typescript
@Index({/* indexOptions */})
@ManyToOne(() => Type, type => type.property, {
  index: { /* indexOptions */  }
})
```

leonidasv (approved):
Nice work.

pleerock:
> If you really want to solve the user confusion with these sort of things I think we need to move towards separate documentation and auto generated API reference. 

> I do agree that "many ways of doing the same thing" is a problem and one that TypeORM definitely suffers from (sometimes with subtle differences between them) and it can confuse users

I think you both admit that current design is confusing and I don't really want to continue with new features that bring a new confusion. Let me close it for now, until we have time to redesign current confusing things and make a better design.

PR #7137, opened 2020-12-02
Sqlite Attach Database

elmpp:
### Description of change

Adds support for SQLite [attach databases](https://sqlite.org/lang_attach.html). This allows database files additional to the default "main" file that TypeORM currently supports. 

This implementation is quite similar to current support for multi-database (postgres, mysql) but without the concept of schema paths. The databases are configured in the same way also - as entity decorators; albeit using filepath strings

The bulk of the complexity comes from platform filepath differences - i felt a big requirement would be that migration files, synchronised file locations etc would be portable between windows/posix. As a result there are a few path utils (with tests) to achieve that. 

Notes:
 - ~Foreign keys aren't supported cross-database so are turned off if any dbs are attached. It could be possible to still support them within the same db. What do you think?~ Foreign keys between tables of same db are included
 - ~i wasn't able to build 'better-sqlite' on my machine so hoping CI will test that~
 - adds `ATTACH` statements to migrations

### Pull-Request Checklist


src/metadata/EntityMetadata.ts
```diff
@@ -861,6 +862,11 @@ export class EntityMetadata {
             }
         }
 
+        if (this.connection.driver instanceof SqliteDriver) {
+            // @todo - perhaps use this driver callback in all cases?
```
nebkat, line 866: At this point lets just abstract `buildTablePath()` for all drivers and pass `tableName`, `schema` and `database`. The function could be made optional also and default to using `tableName` by default.

src/query-builder/QueryBuilder.ts
```diff
@@ -85,6 +85,7 @@ export abstract class QueryBuilder<Entity> {
      * QueryBuilder can be initialized from given Connection and QueryRunner objects or from given other QueryBuilder.
      */
     constructor(connectionOrQueryBuilder: Connection|QueryBuilder<any>, queryRunner?: QueryRunner) {
+
```
nebkat, line 88:
```suggestion
```

src/driver/sqlite/SqliteDriver.ts
```diff
@@ -81,6 +87,29 @@ export class SqliteDriver extends AbstractSqliteDriver {
         return super.normalizeType(column);
     }
 
+    /**
+     * For SQLite, the database may be added in the decorator metadata. It will be a filepath to a database file.
+     */
+    buildTableName(tableName: string, _schema?: string, database?: string): string {
+
+        if (!database) return tableName
```
nebkat, line 95: Missing semicolons throughout this function (and a few others).

src/driver/sqlite-abstract/AbstractSqliteQueryRunner.ts
```diff
@@ -1149,8 +1167,12 @@ export abstract class AbstractSqliteQueryRunner extends BaseQueryRunner implemen
             downQueries.push(this.createIndexSql(oldTable, index));
         });
 
+        const [databaseNew, tableNameNew] = this.splitTablePath(newTable.name)
+        // @ts-ignore // unused var
```
nebkat, line 1171: Can `_` be used for unused vars?

src/driver/sqlite-abstract/AbstractSqliteQueryRunner.ts
```diff
@@ -1197,4 +1218,18 @@ export abstract class AbstractSqliteQueryRunner extends BaseQueryRunner implemen
         this.replaceCachedTable(oldTable, newTable);
     }
 
+    /**
+     * tablePath e.g. "myDB.myTable", "myTable"
+     */
+    protected splitTablePath(tablePath: string): [string | undefined, string] {
+        return ((tablePath.indexOf('.') !== -1) ? tablePath.split('.') : [undefined, tablePath]) as [string | undefined, string]
```
nebkat, line 1225: Should be careful here, `split('.')` could return more than 2. Better to do something like
```suggestion
        let dotIndex = tablePath.indexOf(".");
        return (dotIndex >= 0) ? [tablePath.substr(0, dotIndex), tablePath.substr(dotIndex + 1)] : [undefined, tablePath];
```

nebkat: Also make sure all strings are in double quotes rather than single, should be picked up by ESLint in IDE?

nebkat (changes requested):

src/metadata/EntityMetadata.ts
elmpp: I didn't want to do this outright in my first PR because i'm still unclear the differences between `tableName` and `tablePath`.

I think it needs a bit more thought (and probably needs more consistent escaping strategies in certain places)

src/driver/sqlite-abstract/AbstractSqliteQueryRunner.ts
```diff
@@ -1150,7 +1172,12 @@ export abstract class AbstractSqliteQueryRunner extends BaseQueryRunner implemen
         });
 
         // change table name into 'temporary_table'
-        newTable.name = "temporary_" + newTable.name;
+        newTable.name = this.buildTemporaryTablePath(newTable.name)
+
+        // @ts-ignore // ignore var
```
elmpp, line 1177: would be nice if this was allowable?

src/driver/sqlite-abstract/AbstractSqliteQueryRunner.ts
nebkat: I checked now the syntax is actually to not put in any variable:

`[ , tableNameNew]`

src/metadata/EntityMetadata.ts
nebkat: The only difference should be that `tablePath` contains the full database/schema name, while `tableName` does not. I think drivers have also so far been responsible for escaping, but is there something else I'm missing here?

elmpp: the concept of `tablePath` vs `tableName` seems a little blurry throughout - it differs from here in the `EntityMetadata`, through the `Table` and `TableIndex` instances. Also, the same driver instance can expect differing values based on the platform (the `AbstractSqliteDriver/QueryRunner`). e.g. `sql.js` does not support tablePaths on indexes etc

elmpp: tl;dr; i'd rather not include cross-driver changes in this PR

src/connection/BaseConnectionOptions.ts
```diff
@@ -110,6 +110,11 @@ export interface BaseConnectionOptions {
      */
     readonly extra?: any;
 
+    /**
+     * The directory where ormconfig has been read from
+     */
+    baseDirectory?: string;
```
imnotjames, line 116: What happens when there's no ormconfig?

Why does this feature operate differently from how the standard database path resolution works for sqlite?

src/connection/ConnectionOptionsReader.ts
```diff
@@ -172,11 +176,10 @@ export class ConnectionOptionsReader {
                 Object.assign(connectionOptions, { migrations: migrations });
             }
 
-            // make database path file in sqlite relative to package.json
+            // make database path file absolute in sqlite
             if (options.type === "sqlite" || options.type === "better-sqlite3") {
                 if (typeof options.database === "string" &&
-                    options.database.substr(0, 1) !== "/" &&  // unix absolute
-                    options.database.substr(1, 2) !== ":\\" && // windows absolute
+                    !isAbsolute(options.database) &&
```
imnotjames, line 182: The attaching should probable be defined within the connection options for sqlite instead of in random entities so they are in the same place as the main file's database.  The entities could then reference the alias instead of the filename.

src/driver/sqlite-abstract/AbstractSqliteDriver.ts
```diff
@@ -39,6 +44,11 @@ export abstract class AbstractSqliteDriver implements Driver {
      */
     databaseConnection: any;
 
+    /**
+     * Any attached databases (excepting default 'main')
+     */
+    attachedDatabases: DatabasesMap = {};
```
imnotjames, line 50: can this be protected?

src/driver/sqlite/SqliteDriver.ts
```diff
@@ -139,8 +170,27 @@ export class SqliteDriver extends AbstractSqliteDriver {
     /**
      * Auto creates database directory if it does not exist.
      */
-    protected async createDatabaseDirectory(fullPath: string): Promise<void> {
-        await mkdirp(path.dirname(fullPath));
+    protected async createDatabaseDirectory(dbPath: string): Promise<void> {
+        await mkdirp(dbPath);
```
imnotjames, line 174: Probably can drop this function and instead use `mkdirp` directly

imnotjames (changes requested):
We aren't using yarn for this project, please remove references to it.

pleerock:
I'll close it since there is no answer from author. Feel free to re-open.

elmpp:
@pleerock  Hi there - I must have overlooked the second review of @imnotjames . If I find some time this week to respond is there a decent chance it could be considered? (there been a bit of a wait)

Not sure how to reopen this PR tbh either ?

imnotjames:
You should be able to reopen it yourself.

elmpp:
I see no option to reopen - only the ability to "reference in a new issue". I'd like to keep the comments etc if possible. Weirdly, i'm also unable to upload images in this reply

elmpp:
I have rebased the branch to latest master and added more tests. 

I thought adding commits would give me the option to reopen but looks like not 

I'm going to open another PR with a reference to here. I assume you are filtering closed PR comments 👍

PR #7163, opened 2020-12-06
allow explicit default values on saving an object without relying on db defaults

### LINKED ISSUE #7098: column default is ignored
### Expected Behavior
Defining a column with a default value should take that value on insert new and put it in the column

### Actual Behavior
puts ```DEFAULT``` value - since there's no default value in the db, it throws an error

```
query: START TRANSACTION
query: INSERT INTO "public"."FamilyAccounts"("createdAt", "updatedAt", "name") VALUES ($1, $2, DEFAULT) RETURNING "id", "createdAt", "updatedAt", "name" -- PARAMETERS: ["2020-11-22T20:50:20.187Z","2020-11-22T20:50:20.187Z"]
query failed: INSERT INTO "public"."FamilyAccounts"("createdAt", "updatedAt", "name") VALUES ($1, $2, DEFAULT) RETURNING "id", "createdAt", "updatedAt", "name" -- PARAMETERS: ["2020-11-22T20:50:20.187Z","2020-11-22T20:50:20.187Z"]
```

### Steps to Reproduce
1. Define entity FamilyAccount
2. Define a default value for name
3. Create an instance of FamiltyAccount
4. set all properties but NOT for name
5. Save

```typescript
@Index("FamilyAccounts_pkey", ["id"], { unique: true })
@Entity("FamilyAccounts", { schema: "public" })
export class FamilyAccount extends extends BaseEntity {
    @PrimaryGeneratedColumn({ type: "integer", name: "id" })
    id: number;

    @CreateDateColumn({ name: "createdAt" })
    createdAt: Date;
  
    @UpdateDateColumn({ name: "updatedAt" })
    updatedAt: Date;

    @Column("character varying", { name: "name", nullable: true, length: 255, default: "'test 123'" })
    name: string | null;

    @OneToMany(() => UserInFamilyAccount, userInFamily => userInFamily.familyAccount)
    userInFamilyAccounts: [];
}

// then try to create the family account:
  let fa = FamilyAccount.create();
  fa.createdAt = fa.updatedAt = new Date();
  fa = await fa.save();
```

### My Environment

| Node.js version     | v12.0.0| 
| Typescript version  | v3.9.7 | 
| TypeORM version     | v0.2.26 | 

- [ ] `aurora-data-api`
- [ ] `aurora-data-api-pg`
- [ ] `better-sqlite3`
- [ ] `cockroachdb`
- [ ] `cordova`
- [ ] `expo`
- [ ] `mongodb`
- [ ] `mysql`
- [ ] `nativescript`
- [ ] `oracle`
- [x] `postgres`
- [ ] `react-native`
- [ ] `sap`
- [ ] `sqlite`
- [ ] `sqlite-abstract`
- [ ] `sqljs`
- [ ] `sqlserver`

yossibeck:
…ults

Fixes #7098

per https://github.com/typeorm/typeorm/issues/7098 
allow setting default values from code and not just from the database 

### Pull-Request Checklist

- [v ] Code is up-to-date with the `master` branch
- [ x] `npm run lint` passes with this change
- [ v] This pull request links relevant issues as `Fixes #0000`
- [x ] There are new or updated unit tests validating the change
- [ v] Documentation has been updated to reflect this change

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "typeorm",
-  "version": "0.2.28",
+  "version": "0.2.29",
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
@@ -4227,7 +4227,6 @@
       "resolved": "https://registry.npmjs.org/fstream/-/fstream-1.0.12.tgz",
       "integrity": "sha512-WvJ193OHa0GHPEL+AycEJgxvBEwyfRkN1vhjca23OaPVMCaLCXTd5qAu82AjTcgP1UJmytkOKb63Ypde7raDIg==",
       "dev": true,
-      "optional": true,
       "requires": {
         "graceful-fs": "^4.1.2",
         "inherits": "~2.0.0",
@@ -4239,15 +4238,13 @@
           "version": "1.2.5",
           "resolved": "https://registry.npmjs.org/minimist/-/minimist-1.2.5.tgz",
           "integrity": "sha512-FM9nNUYrRBAELZQT3xeZQ7fmMOBg6nWNmJKTcgsJeaLstP/UODVpGsr5OhXhhXg6f+qtJ8uiZ+PUxkDWcgIXLw==",
-          "dev": true,
-          "optional": true
+          "dev": true
         },
         "mkdirp": {
           "version": "0.5.5",
           "resolved": "https://registry.npmjs.org/mkdirp/-/mkdirp-0.5.5.tgz",
           "integrity": "sha512-NKmAlESf6jMGym1++R0Ra7wvhV+wFW63FaSOFPwRahvea0gMUcGUhVeAg/0BC0wiv9ih5NYPB1Wn1UEI1/L+xQ==",
           "dev": true,
-          "optional": true,
           "requires": {
             "minimist": "^1.2.5"
           }
@@ -4257,7 +4254,6 @@
           "resolved": "https://registry.npmjs.org/rimraf/-/rimraf-2.7.1.tgz",
           "integrity": "sha512-uWjbaKIK3T1OSVptzX7Nl6PvQ3qAGtKEtVRjRuazjfL3Bx5eI409VZSqgND+4UNnmzLVdPj9FqFJNPqBZFve4w==",
           "dev": true,
-          "optional": true,
           "requires": {
             "glob": "^7.1.3"
           }
--- a/src/decorator/options/ColumnOptions.ts
+++ b/src/decorator/options/ColumnOptions.ts
@@ -70,6 +70,12 @@ export interface ColumnOptions extends ColumnCommonOptions {
     default?: any;
 
     /**
+     * Saves the default value (if one was given in this ColumnOptions) to the db in case none was given
+     * regardless of the database's default value
+     */
+    explicitDefault? : boolean;
+
+    /**
      * ON UPDATE trigger. Works only for MySQL.
      */
     onUpdate?: string;
--- a/src/metadata/ColumnMetadata.ts
+++ b/src/metadata/ColumnMetadata.ts
@@ -122,6 +122,12 @@ export class ColumnMetadata {
     default?: any;
 
     /**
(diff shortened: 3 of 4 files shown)


pleerock:
Thank you for contribution!

I don't think `explicitDefault` is a good name for this feature. I suggest at least to name it `computedDefault` instead. What do you think?

Also, can you please add tests for this new feature?

nebkat:
I think `explicitDefault` implies it will explicitly set the default values for columns when inserting as opposed to implicitly using DB default values, while `computedDefault` might make it seem like it is about using a computed value (e.g. "CURRENT_TIMESTAMP") in the database?

There is the unintended side effect that if a raw value with `() => calculateNewId()` is used the default effectively becomes "computed" at runtime, which is probably what you were thinking of here?

---

The `() => value` wrapper syntax has been useful for passing raw values but I think it would be better if there was an "expression builder" system which extended the existing find operators. Then anywhere that an SQL expression is used you could pass an `Expression` value such as `Raw('CURRENT_TIMESTAMP()')`, or better yet `Fn('CURRENT_TIMESTAMP')` . It's also worth noting that this is actually shorter to type than the function wrapper, and should be much more readable to less experienced users, so overall would be a very positive change.

I've done some preliminary work on this here https://github.com/nebkat/typeorm/tree/expression-builder/src/expression-builder/expression. Goal is to make it fully compatible with existing find operators to avoid any breaking changes.

If that was to be added, the `() => 'raw'` syntax could be deprecated and there would no longer be the possibility of using a computed default value like this, or better yet it could become the correct way of using a "computed" value like `() => calculateNewId()`.

pleerock:
> I think explicitDefault implies it will explicitly set the default values for columns when inserting as opposed to implicitly using DB default values, while computedDefault might make it seem like it is about using a computed value (e.g. "CURRENT_TIMESTAMP") in the database?

Most of options in `@Column` decorator are database-specific, including `default`. `explicitDefault` should have a name that would tell about this feature something like "hey dude this default isn't a database-level default, this value will be set in the **code** by ORM". `computedDefault` says like "this value will be used in ORM's computations, it won't be delegated to database". `explicitDefault` doesn't really tell me something like that. `computedDefault` is a first thing I came up with, it's not perfect and I'm open to a better alternatives.

> The () => value wrapper syntax has been useful for passing raw values

Yes, we can deprecate this feature in a future and use `Raw` or similar operator - it would be a better approach.

> I've done some preliminary work on this here https://github.com/nebkat/typeorm/tree/expression-builder/src/expression-builder/expression. Goal is to make it fully compatible with existing find operators to avoid any breaking changes.

Looks good. Do you have docs or examples of usage of these operators?

nebkat:
> Most of options in @Column decorator are database-specific

This is certainly a valid point, and there may be a broader discussion to be had regarding the division between database/ORM properties especially for relations and if select-and-map is introduced. That said, I'd be much more likely to think `computedDefault` is some kind of database feature since all SQL engines can do some kind of computations, including for default values. What about `insertExplicitDefault`/`insertDefaultExplicitly` as it ties it directly to the `insert()` operation?

---

I see you've mentioned in a few PRs the concern about users getting confused ormaking assumptions about functionality - I'd say again here as in https://github.com/typeorm/typeorm/pull/7119#issuecomment-758044921, I think a published API reference would really help in this regard. We should always try make function/option naming be as descriptive as possible but there's only so much that can be done, and for what is quite a complex library as a bare minimum the user should be expected to check if there is a comment about a field they are unsure of. Already this is possible through IDEs.

---

> Looks good. Do you have docs or examples of usage of these operators?

I'll make a draft PR with full examples soon 🎉

pleerock:
In order for this PR to be merged we need to do following changes and additions:

* we need to figure out a better name for this option
* new option should be able to accept functions, to allow dynamic values
* new option should be test-covered

ablankenship10:
Is this item progressing at all? I had just implemented the same before finding this, as `defaultValue`, field but open to whatever naming convention.

AlexMesser:
Closing in favor of #7680

PR #7166, opened 2020-12-07
Support maxdecimaldigits option by geometry type

### LINKED ISSUE #7165: Support maxdecimaldigits option of geometry type
### The Problem

PostGIS may return coordinates with a required precision. The option [maxdecimaldigits](https://postgis.net/docs/ST_AsGeoJSON.html) allows to reduce/increase the max number of decimal places, moreover, it rounds the value, so `30.0123456789` becomes `30.012345679`. In the existing solution, ST_AsGeoJSON relies on the default value (9) causing issues in high-precision applications, e.g. previously received geometries can't be used for searching the same data using ST_Intersects, ST_Overlaps, etc.

### The Solution

Typeorm already has `precision` field, which doesn't affect geometry type, but represents about the same idea as `maxdecimaldigits` in PostGIS. By adding `precision` as a second argument to ST_AsGeoJSON developers can override the default value, therefore either simplifying data or getting more precise results.

For instance,
```
selectionPath = `ST_AsGeoJSON(${selectionPath}, ${column.precision || 9})::json`;
```
will work pretty much everywhere without breaking any existing application.

- [ ] `aurora-data-api`
- [ ] `aurora-data-api-pg`
- [ ] `better-sqlite3`
- [ ] `cockroachdb`
- [ ] `cordova`
- [ ] `expo`
- [ ] `mongodb`
- [ ] `mysql`
- [ ] `nativescript`
- [ ] `oracle`
- [x] `postgres`
- [ ] `react-native`
- [ ] `sap`
- [ ] `sqlite`
- [ ] `sqlite-abstract`
- [ ] `sqljs`
- [ ] `sqlserver`

alexile:
### Description of change

Adds support of PostGIS `maxdecimaldigits` option. The value comes from the existing `precision` option. Currently, it isn't possible, so the precision is always 9, therefore small, detailed elements either overlap or cannot be found using ST_Intersects, ST_Overlaps, etc. By default, PostGIS rounds all decimal places beyond `maxdecimaldigits`, which causes precision issues/ruins searching. This PR makes `precision` not only affect `double` but `geometry` types as well.

Fixes #7165 

### Pull-Request Checklist

### DIFF
--- a/src/query-builder/SelectQueryBuilder.ts
+++ b/src/query-builder/SelectQueryBuilder.ts
@@ -1765,7 +1765,7 @@ export class SelectQueryBuilder<Entity> extends QueryBuilder<Entity> implements 
 
                 if (this.connection.driver instanceof PostgresDriver)
                     // cast to JSON to trigger parsing in the driver
-                    selectionPath = `ST_AsGeoJSON(${selectionPath})::json`;
+                    selectionPath = `ST_AsGeoJSON(${selectionPath}, ${column.precision || 9})::json`;
 
                 if (this.connection.driver instanceof SqlServerDriver)
                     selectionPath = `${selectionPath}.ToString()`;
--- a/src/query-builder/SelectQueryBuilder.ts
+++ b/src/query-builder/SelectQueryBuilder.ts
@@ -1765,8 +1765,11 @@ export class SelectQueryBuilder<Entity> extends QueryBuilder<Entity> implements 
 
                 if (this.connection.driver instanceof PostgresDriver)
                     // cast to JSON to trigger parsing in the driver
-                    selectionPath = `ST_AsGeoJSON(${selectionPath}, ${column.precision || 9})::json`;
-
+                    if (column.precision) {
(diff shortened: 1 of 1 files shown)


pleerock:
Thank you for contribution! 

Can you please add a test, so we can be sure added functionality won't break next time?

Where the number `9` comes from? Is it a postgres default? If yes, can it be added to the `PostgresDriver.dataTypeDefaults` property?

Or, since it's an optional argument, maybe for simplicity purpose we can simply append `column.precision` only if it was defined in there?

alexile:
@pleerock correct, the number comes from [ST_AsGeoJSON](https://postgis.net/docs/ST_AsGeoJSON.html) docs.

> Or, since it's an optional argument, maybe for simplicity purpose we can simply append column.precision only if it was defined in there?
Makes sense, I will take a look

pleerock:
I added a simplification I was asking about. If you'll find time to write tests for this feature, it would be great to have them to avoid future breaking. Thank you!

Issue #5371, opened 2020-01-17
Default enum null value causes a bug


OrkhanAlikhanov:
**Issue type:**

[x] bug report

**Database system/driver:**

[x] `mysql` / `mariadb`

**TypeORM version:**

[x] `latest`
[x] `@next`

**Steps to reproduce or a small repository showing the problem:**
Create an enum column with default value null
```ts
enum MyEnum {
  valueA = 'valueA',
  valueB = 'valueB',
}

@Entity()
export class Entity extends BaseEntity {
  @Column({
    type: 'enum',
    enum: MyEnum,
    nullable: true,
    default: null,
  })
  field?: MyEnum;
}
```

result 
```js
{
  message: "Invalid default value for 'field'",
  code: 'ER_INVALID_DEFAULT',
  errno: 1067,
  sqlState: '42000',
  sqlMessage: "Invalid default value for 'field'",
  name: 'QueryFailedError',
  query: "ALTER TABLE `entity` CHANGE `field` `field` enum ('valueA', 'valueB') NULL DEFAULT 'null'",
  parameters: []
}
```

It should've generated  `query: "ALTER TABLE ``entity`` CHANGE ``field`` ``field`` enum ('valueA', 'valueB') NULL DEFAULT null"`,

Adding `&& defaultValue !== null` seems to work
https://github.com/typeorm/typeorm/blob/e8eb98b0bd3c551d82c3f6fd0899259ea868c685/src/driver/mysql/MysqlDriver.ts#L572

robertrypula:
Any updates on this topic? I'm stuck in the same issue :(

agoulziprod:
this solution found here: [https://github.com/typeorm/typeorm/issues/4833](https://github.com/typeorm/typeorm/issues/4833) worked for me finally,
```
{
    name: 'status',
    type: 'enum',
    enum: ['active', 'inactive', 'paused'],
    enumName: 'statusEnum',
    default: '"inactive"'
}
```

falahati:
This happens with PostgreSQL as well.
Following is the result of TypeOrm field definition:
```sql
"status" "status_enum" DEFAULT \'null\'
```
From:
```typescript
@Column({ name: "status", type: "enum", enum: Status, nullable: true, default: undefined  })
public winner: Status| null = null;
```
Resulting in:

> 'invalid input value for enum status_enum: "null"'

Probably due to putting null inside `''` and escaping it.

[...]

PR #7221, opened 2020-12-23
Nullable default value for `Enum` types

falahati:
### Description of change
This PR is the result of a conversation at #5371 regarding a bug with TypeOrm and MySQL resulting in the generation of an invalid table create query for enum fields with a default value of `null`
This happens simply due to treating `null` as another value for the enum and therefore, escaping and passing it as a string to the database.
The fix was easy, we are already checking for `undefined` before searching in enum and escaping the value in case there was no default value specified, now we are also checking for `null` in case the passed default value is _special_ and should not be treated like others.
This is done by simply replacing `value !== undefined` with `value != null` therefore checking for both conditions.

### Pull-Request Checklist

This is a small change and so I did the changes on my browser without setting up the project for test, build, and lint. But honestly, I don't think it can break anything! Feel free to reject it if it doesn't fits the standards!

Thank you for the great library.

### DIFF
--- a/src/driver/mysql/MysqlDriver.ts
+++ b/src/driver/mysql/MysqlDriver.ts
@@ -582,7 +582,7 @@ export class MysqlDriver implements Driver {
     normalizeDefault(columnMetadata: ColumnMetadata): string {
         const defaultValue = columnMetadata.default;
 
-        if ((columnMetadata.type === "enum" || columnMetadata.type === "simple-enum") && defaultValue !== undefined) {
+        if ((columnMetadata.type === "enum" || columnMetadata.type === "simple-enum") && defaultValue != null) {
             return `'${defaultValue}'`;
         }
 
--- a/src/driver/postgres/PostgresDriver.ts
+++ b/src/driver/postgres/PostgresDriver.ts
@@ -725,7 +725,7 @@ export class PostgresDriver implements Driver {
             (
                 columnMetadata.type === "enum"
                 || columnMetadata.type === "simple-enum"
-            ) && defaultValue !== undefined
+            ) && defaultValue != null
         ) {
             if (columnMetadata.isArray && Array.isArray(defaultValue)) {
                 return `'{${defaultValue.map((val: string) => `${val}`).join(",")}}'`;


src/driver/mysql/MysqlDriver.ts
```diff
@@ -582,7 +582,7 @@ export class MysqlDriver implements Driver {
     normalizeDefault(columnMetadata: ColumnMetadata): string {
         const defaultValue = columnMetadata.default;
 
-        if ((columnMetadata.type === "enum" || columnMetadata.type === "simple-enum") && defaultValue !== undefined) {
+        if ((columnMetadata.type === "enum" || columnMetadata.type === "simple-enum") && defaultValue != null) {
```
pleerock, line 585: `!=` doesn't feel reliable

pleerock:
what is the point to set `default: null`, if when you set `nullable: true` database already sets you a null by default?

src/driver/mysql/MysqlDriver.ts
falahati: you think `defaultValue !== undefined && defaultValue !== null` is more descriptive? cuz `!=` is reliable, its just not as descriptive since we might have both these values here and each might mean a different think in context.
I will change it if you think that it will help.

falahati:
> what is the point to set `default: null`, if when you set `nullable: true` database already sets you a null by default?

true as I never saw one that doesn't automatically set the default to `null` when the column is `nullable`; however in many projects, I prefer to explicitly mention it in design and write it in code as it makes the design more obvious to the person implementing it.
Having a default value indicates to the programmer that you don't need to pass a value to this column to create a row.

However, the problem here with typeorm is not the usefulness of the `default: null` value, however, the discussion is about the inconsistency of typeorm behavior when a `default: null` value is used on an `enum` column.

pleerock:
> I prefer to explicitly mention it in design and write it in code as it makes the design more obvious to the person implementing it.

from other side, it might be confusing if you put both nullable: true and default: null. Other programmer might ask you "why did you use both if nullable **already** does null? What is the point? (Like I did) Can I remove one?"
Another problem is what will happen if you use nullable: false and default: null? Personally I don't know if database even allow such behaviour. If we make `default: null` to lead to *no behaviour*, it can also confuse because `undefined` served as a thing leading to no behaviour while null in TypeORM always does something else regarding to nulls.

Not sure what is the best approach here, maybe remove `null` support in defaults at all, but this will be a breaking change.

falahati:
> > I prefer to explicitly mention it in design and write it in code as it makes the design more obvious to the person implementing it.
> 
> from other side, it might be confusing if you put both nullable: true and default: null. Other programmer might ask you "why did you use both if nullable **already** does null? What is the point? (Like I did) Can I remove one?"
> Another problem is what will happen if you use nullable: false and default: null? Personally I don't know if database even allow such behaviour. If we make `default: null` to lead to _no behaviour_, it can also confuse because `undefined` served as a thing leading to no behaviour while null in TypeORM always does something else regarding to nulls.
> 
> Not sure what is the best approach here, maybe remove `null` support in defaults at all, but this will be a breaking change.

Well, I don't understand the other side's argument.
You see, I am not aware of such a standard that says every database needs to set the default value of a `nullable` field to `null` and if there is one, I don't know if all databases comply with that. Hence why I said, "Every database that I **saw**...". I assume that the majority of developers are also not sure about the standards of database design and therefore, nobody should find a `default: nullable` anything out of the ordinary. Not to mention that this type of descriptive design is not something that someone decides to remove since that decision should come with all sort of knowledge that I pointed out before.

So unless you do have this information, this confusion would never happen. But then if you have this knowledge, you can simply deduce that `default: null` is there just to be explicit.

My understanding is that the TypeOrm core should not limit the `null` value since it makes a lot of confusion without such knowledge that I mentioned.
However, drivers for each database engine are free to ignore it in the final create table query generation process. Or just does what it does now, pass it along; since again, I don't remember any database engine that I saw has any problem with an additional `DEFAULT NULL` statement in the field creation script. If they are ok with that, we should be too.

src/driver/mysql/MysqlDriver.ts
nebkat: Perhaps we could just place the null check before all the others? Then `null` is always `NULL` and the rest stays the same. Would apply the same fix for sets too.

```suggestion
        if (defaultValue === null) {
            return `NULL`;
        } else if ((columnMetadata.type === "enum" || columnMetadata.type === "simple-enum") && defaultValue !== undefined) {
```

nebkat:
All other column types (except 'set' which has the same problem) accept `default: null`, so unless you want to explicitly prevent it from ever being set as a default value (`default?: Exclude<any, null>`?, which itself would be a much larger breaking change), I think this should be allowed.

> from other side, it might be confusing if you put both nullable: true and default: null. Other programmer might ask you "why did you use both if nullable already does null? What is the point? (Like I did) Can I remove one?"

You could extend this logic to something like `ColumnOptions` `primary` and `nullable`. A programmer could ask why is `{primary: true, nullable: false}` defined, but the answer is surely quite obvious here. The extra information doesn't do any harm, and I don't think the library should ever implicitly prevent redundant information for the sake of enforcing conciseness.

> Another problem is what will happen if you use nullable: false and default: null? Personally I don't know if database even allow such behaviour. If we make default: null to lead to no behaviour, it can also confuse because undefined served as a thing leading to no behaviour while null in TypeORM always does something else regarding to nulls.

It should do what `{ primary: true, nullable: true }` does, which is throw a `PrimaryColumnCannotBeNullableError`, in this case a `NonNullableColumnCannotDefaultNullError`. Even without an error, the user has clearly contradicted themselves so it is their own fault if the database doesn't accept it.

`default: null`, wouldn't be leading to no behaviour, since it would explicitly set the column default to `NULL`. `undefined` would continue to do nothing (i.e. return `undefined` from `normalizeDefault`).

---

I would almost have said that `default: null` should automatically set `nullable: true` if it is undefined, but it is probably better not to do this in case of situations like:

```ts
function calculateDefault() {
    try {
        return doSomething();
    } catch (e) {
        return null;
    }

// ...

@Column({
    default: calculateDefault() // never actually meant for this to be null
})
```

PR #7229, opened 2020-12-30
fix: Fix unnecessary migrations for timestamps with default values of CURRENT_TIMESTAMP and precision

### LINKED ISSUE #3991: Migration keeps changing @CreateDateColumn/@UpdateDateColumn timestamp column to same definition
**Issue type:**

[ ] question
[x] bug report
[ ] feature request
[ ] documentation issue

**Database system/driver:**

[ ] `cordova`
[ ] `mongodb`
[ ] `mssql`
[x] `mysql` / `mariadb`
[ ] `oracle`
[ ] `postgres`
[ ] `cockroachdb`
[ ] `sqlite`
[ ] `sqljs`
[ ] `react-native`
[ ] `expo`

**TypeORM version:**

[ ] `latest`
[ ] `@next`

I have a bunch of tables with my create- and updateTimestamp columns defined as follows:
```ts
  @CreateDateColumn({ type: 'timestamp', precision: null, default: () => 'CURRENT_TIMESTAMP' })
  createTimestamp: Date;

  @UpdateDateColumn({ type: 'timestamp', precision: null, default: () => 'CURRENT_TIMESTAMP' })
  updateTimestamp: Date;
```

When I run `migration:generate`, typeorm is repeatedly (even after running it) generating a migration file that changes the column definition to the same definition it already has. You can tell because the `up` and `down` methods do the exact same thing. I would expect the migration to be empty.

Example migration file
```ts

export class TestTimestamps1555253092006 implements MigrationInterface {

        await queryRunner.query("ALTER TABLE `album` CHANGE `createTimestamp` `createTimestamp` timestamp NOT NULL DEFAULT CURRENT_TIMESTAMP");
        await queryRunner.query("ALTER TABLE `album` CHANGE `updateTimestamp` `updateTimestamp` timestamp NOT NULL DEFAULT CURRENT_TIMESTAMP");
    }

        await queryRunner.query("ALTER TABLE `album` CHANGE `updateTimestamp` `updateTimestamp` timestamp NOT NULL DEFAULT CURRENT_TIMESTAMP");
        await queryRunner.query("ALTER TABLE `album` CHANGE `createTimestamp` `createTimestamp` timestamp NOT NULL DEFAULT CURRENT_TIMESTAMP");
    }

}
```

I can work around it by defining my create- and updateTimestamp columns as follows
```ts
  @Column('timestamp', {
    default: () => 'CURRENT_TIMESTAMP',
  })
  createTimestamp: Date;

  @Column('timestamp', {
    default: () => 'CURRENT_TIMESTAMP',
    onUpdate: 'CURRENT_TIMESTAMP',
  })
  updateTimestamp: Date;
```

But it's a shame that it's not working with the `@CreateDateColumn` and `@UpdateDateColumn` decorators.

<!--
    [x] `mysql`
    [ ] `postgres`

!>

RohanTalip:
### Description of change

After upgrading `typeorm` from `0.2.12` to `0.2.29`, I started seeing `COMMENT ON COLUMN "tableName"."columnName" IS NULL` statements in both the `up` and `down` migrations for columns that had default values but no changes.

I saw that issue #3991 looked similar, so I fixed it.  I have a separate fix for the comment issue that I will send in another pull request.

Details:
- Returned the database specific equivalent of CURRENT_TIMESTAMP in normalizeDefault().
- Pulled out individual conditions for easier debugging.
- Intentionally used '!=' instead of '!==' for the precision comparison, so that undefined and null compare as equal.

Fixes #3991

### Pull-Request Checklist

### DIFF
--- a/src/driver/cockroachdb/CockroachDriver.ts
+++ b/src/driver/cockroachdb/CockroachDriver.ts
@@ -483,6 +483,9 @@ export class CockroachDriver implements Driver {
         const defaultValue = columnMetadata.default;
         const arrayCast = columnMetadata.isArray ? `::${columnMetadata.type}[]` : "";
 
+        // console.log(`normalizeDefault() : columnMetadata.propertyName = ${columnMetadata.propertyName}`);
+        // console.log(`normalizeDefault() : defaultValue = ${defaultValue}`);
+
         if (typeof defaultValue === "number") {
             return `(${defaultValue})`;
 
--- a/src/driver/cockroachdb/CockroachDriver.ts
+++ b/src/driver/cockroachdb/CockroachDriver.ts
@@ -490,7 +490,11 @@ export class CockroachDriver implements Driver {
             return defaultValue === true ? "true" : "false";
 
         } else if (typeof defaultValue === "function") {
-            return defaultValue();
+            const value = defaultValue();
+            if (value === "CURRENT_TIMESTAMP") {
+                return "current_timestamp()";
+            }
+            return value;
 
         } else if (typeof defaultValue === "string") {
             return `'${defaultValue}'${arrayCast}`;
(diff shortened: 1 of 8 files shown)


RohanTalip:
Actually, I think I got one of the tests wrong ... and likely some production code as a result ... I'll take a look at it soon.

src/driver/cockroachdb/CockroachDriver.ts
```diff
@@ -614,17 +621,43 @@ export class CockroachDriver implements Driver {
...
-                || tableColumn.isUnique !== this.normalizeIsUnique(columnMetadata)
-                || tableColumn.isGenerated !== columnMetadata.isGenerated;
+            // Pulled out individual conditions for easier debugging.
+
+            const nameCond = tableColumn.name !== columnMetadata.databaseName;
+            const typeCond = tableColumn.type !== this.normalizeType(columnMetadata);
+            const lengthCond = tableColumn.length !== columnMetadata.length;
+
+            // Intentionally using '!=' instead of '!==' so that undefined and null compare as equal.
+            const precisionCond = tableColumn.precision != columnMetadata.precision;
```
pleerock, line 631: not sure I really like `!=` approach instead of explicit check for null or undefined, because `!=` forces me to ask questions:

* what if tableColumn.precision is `0` and columnMetadata.precision is `undefined` or `null`? What behaviour should I expect?
* why the hell precision can be both `null` or `undefined`? In what situations it can be `null` and in what situations it can be `undefined`? Feels like it should be only one of them when the value isn't defined? Maybe normalization can be applied somewhere else before this check?

pleerock: same questions can arise in different heads, and instead of writing the comments maybe it's better to replace this check with explicit one?

nebkat: If we could bump the TypeScript version a nice check here would be
```suggestion
            const precisionCond = (tableColumn.precision ?? null) !== (columnMetadata.precision ?? null);
```

src/driver/cockroachdb/CockroachDriver.ts
```diff
@@ -483,14 +483,21 @@ export class CockroachDriver implements Driver {
         const defaultValue = columnMetadata.default;
         const arrayCast = columnMetadata.isArray ? `::${columnMetadata.type}[]` : "";
 
+        // console.log(`normalizeDefault() : columnMetadata.propertyName = ${columnMetadata.propertyName}`);
+        // console.log(`normalizeDefault() : defaultValue = ${defaultValue}`);
```
nebkat, line 487: Should probably remove all of the commented out logging lines.

src/driver/cockroachdb/CockroachDriver.ts
nebkat: This is now possible on latest TypeScript 😀

PR #7236, opened 2020-12-31
[FEAT] Add NOWAIT and SKIP LOCKED lock support for MySQL

lacunadream:
Adds `SKIP LOCKED` (pessimistic_partial_write) and `NOWAIT` (pessimistic_write_or_fail) support for Mysql drivers. 

Database support as follows
Database | SKIP LOCKED | NOWAIT
------------ | ------------- | ------------
MySQL | 8.0 | 8.0
MariaDB | Unsupported | 10.3.0

Added test cases, but they do not pass with the current versions listed in docker-compose. (Have tested with the versions listed above and they pass). Generally, how does typeorm handle cases like this?

References: 
https://dev.mysql.com/doc/refman/8.0/en/innodb-locking-reads.html
https://jira.mariadb.org/browse/MDEV-13115
https://mariadb.com/kb/en/wait-and-nowait/

### DIFF
--- a/src/query-builder/SelectQueryBuilder.ts
+++ b/src/query-builder/SelectQueryBuilder.ts
@@ -1686,14 +1686,14 @@ export class SelectQueryBuilder<Entity> extends QueryBuilder<Entity> implements 
                     throw new LockNotSupportedOnGivenDriverError();
                 }
             case "pessimistic_partial_write":
-                if (driver instanceof PostgresDriver) {
+                if (driver instanceof PostgresDriver || driver instanceof MysqlDriver) {
                     return " FOR UPDATE SKIP LOCKED";
 
                 } else {
                     throw new LockNotSupportedOnGivenDriverError();
                 }
             case "pessimistic_write_or_fail":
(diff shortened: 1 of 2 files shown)


jorenvandeweyer:
Nice, thanks for adding this!
I wanted to open a pull request myself for this too today. 

Any idea how fast this can be pushed into the next release?

nebkat:
Probably have to add a check for the driver version so it doesn't get used on earlier MySQL versions.

pleerock:
Thank you for contribution!

As @nebkat mentioned, we need to add a version check (can be added into test itself) in order to prevent CI fail.

lacunadream:
@pleerock added the tests with version checks. CI passes now.

nebkat:
Probably outside the scope of this PR but seeing as all the checks in `createLockExpression()` are solely based on `lockMode` and driver they should really be moved to the drivers themselves. Then the version logic can be implemented specifically for the MySQL driver rather than requiring the user to do so.

pleerock:
Thank you for contribution! 🎉

PR #7239, opened 2021-01-02
fix: get length attribute of postgres array columns

### LINKED ISSUE #6990: Postgres array column dropped and re-added during synchronization when no change is made.
## Issue Description
When using Postgres, if a length is specified for an array of type varchar and synchronization is enabled, Typeorm will drop and re-add the column on startup even if there are no schema changes.

If the length property is removed, the behavior will stop occurring.

### Expected Behavior
When there are no changes to the column, Typeorm should recognize the current database schema matches the column configuration, and not drop and re-add the column during synchronization.

### Actual Behavior
When synchronizing on startup, Typeorm issues SQL queries to drop and re-add the column, even though nothing has changed.

From the console output at startup:
```
query: ALTER TABLE "test"."user" DROP COLUMN "roles"
query: ALTER TABLE "test"."user" ADD "roles" character varying(64) array NOT NULL
```

### Steps to Reproduce

1. Configure Typeorm with a Postgres connection.
2. Enable synchronization.
3. Configure an entity with a column of type string[] and options set to { array: true, type: "character varying", length: 64 }.
4. Launch the program to create the new column, and then restart it to trigger synchronization again.

```typescript
 @Column({
    array: true,
    type: "character varying",
    length: 64,
  })
  roles: string[];
```

For a small example, see: https://github.com/edcolvin/array-sync-issue

### My Environment

| Operating System    |  CentOS 7     |
| Node.js version     | v12.16.3 | 
| Typescript version  | 4.0.5 | 
| TypeORM version     | 0.2.28 | 

### Additional Context
Tested using Postgres:11 Docker image 
PostgreSQL 11.9 (Debian 11.9-1.pgdg90+1) on x86_64-pc-linux-gnu, compiled by gcc (Debian 6.3.0-18+deb9u1) 6.3.0 20170516, 64-bit

- [ ] `aurora-data-api`
- [ ] `aurora-data-api-pg`
- [ ] `better-sqlite3`
- [ ] `cockroachdb`
- [ ] `cordova`
- [ ] `expo`
- [ ] `mongodb`
- [ ] `mysql`
- [ ] `nativescript`
- [ ] `oracle`
- [x] `postgres`
- [ ] `react-native`
- [ ] `sap`
- [ ] `sqlite`
- [ ] `sqlite-abstract`
- [ ] `sqljs`
- [ ] `sqlserver`

edcolvin:
### Description of change

Fixes #6990. 

PostgresQueryRunner method loadTables is used to determine if the database is in sync with the entity definitions. This change updates how loadTables determines the length value of array columns. It uses the postgres-specific pg_catalog.pg_attribute table and the pg_catalog.format_type function to find the data type name, which includes the length for types that have a length attribute. Previously, loadTables relied solely on character_maximum_length in information_schema.columns, which is null for array columns. Because of this, array columns with a length attribute were always dropped. 

This also includes a test that confirms the issues is resolved by this fix. 

### Pull-Request Checklist

### DIFF
--- a/src/driver/postgres/PostgresQueryRunner.ts
+++ b/src/driver/postgres/PostgresQueryRunner.ts
@@ -1417,12 +1417,28 @@ export class PostgresQueryRunner extends BaseQueryRunner implements QueryRunner 
             return `("table_schema" = '${schema}' AND "table_name" = '${name}')`;
         }).join(" OR ");
         const tablesSql = `SELECT * FROM "information_schema"."tables" WHERE ` + tablesCondition;
+        
+        /**
+         * Uses standard SQL information_schema.columns table and postgres-specific 
+         * pg_catalog.pg_attribute table to get column information.
+         * @see https://stackoverflow.com/a/19541865
+         */
         const columnsSql = `
-            SELECT
-                *,
-                pg_catalog.col_description(('"' || table_catalog || '"."' || table_schema || '"."' || table_name || '"')::regclass::oid, ordinal_position) as description,
-                ('"' || "udt_schema" || '"."' || "udt_name" || '"')::"regtype" AS "regtype"
-            FROM "information_schema"."columns"
+            SELECT columns.*,
+              pg_catalog.col_description(('"' || table_catalog || '"."' || table_schema || '"."' || table_name || '"')::regclass::oid, ordinal_position) AS description,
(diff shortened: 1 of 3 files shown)


pleerock:

It's super hard for me to review this kind of PRs because it's not possible to predict how these changes will actually work. Let's merge it and keep watching if it won't arise new issues.

Issue #7256, opened 2021-01-08
find where clauses params order is wrong


Diluka:
## Issue Description
here is a query like below
```ts
await OrderBill.find({
        where: {
          paidAt: Raw((alias) => `DATE(${alias}) = DATE(:date)`, { date: index.toDate() }),
          type: OrderBillType.HOSPITAL_FEE,
          isPaid: true,
        },
})
```

### Expected Behavior

generate right params array (omit select)
```sql
SELECT *
FROM `mrcs_order_bill` `OrderBill`
WHERE DATE(`OrderBill`.`paidAt`) = DATE(?)
  AND `OrderBill`.`type` = ?
  AND `OrderBill`.`isPaid` = ? -- PARAMETERS: ["2020-12-31T16:00:00.000Z","HOSPITAL_FEE",true]
```

### Actual Behavior

generate wrong params array (omit select)
```sql
SELECT *
FROM `mrcs_order_bill` `OrderBill`
WHERE DATE(`OrderBill`.`paidAt`) = DATE(?)
  AND `OrderBill`.`type` = ?
  AND `OrderBill`.`isPaid` = ? -- PARAMETERS: ["HOSPITAL_FEE",true,"2020-12-31T16:00:00.000Z"]
```

### Steps to Reproduce

1.
2.

```typescript
// insert code here
```

### My Environment

| Dependency          | Version  |
| ---                 | ---      |
| Operating System    |          |
| Node.js version     | v12.20.0 | 
| Typescript version  | v4.1.3 | 
| TypeORM version     | v0.2.29 | 

### Relevant Database Driver(s)

- [ ] `aurora-data-api`
- [ ] `aurora-data-api-pg`
- [ ] `better-sqlite3`
- [ ] `cockroachdb`
- [ ] `cordova`
- [ ] `expo`
- [ ] `mongodb`
- [x] `mysql`
- [ ] `nativescript`
- [ ] `oracle`
- [ ] `postgres`
- [ ] `react-native`
- [ ] `sap`
- [ ] `sqlite`
- [ ] `sqlite-abstract`
- [ ] `sqljs`
- [ ] `sqlserver`

### Are you willing to resolve this issue by submitting a Pull Request?

- [ ] Yes, I have the time, and I know how to start.
- [ ] Yes, I have the time, but I don't know how to start. I would need guidance.
- [ ] No, I don't have the time, although I believe I could do it if I had the time...
- [x] No, I don't have the time and I wouldn't even know how to start.

Issue #7104, opened 2020-11-24
Parameter in Raw when used with FindConditions OR


nyotiemon:
## Issue Description

According to pleerock answer on #2929 [link](https://github.com/typeorm/typeorm/issues/2929#issuecomment-442788892), when combining with Raw() function, the raw-parameter will be stacked at the last.

``` typescript
@Entity()
class Item {
    @PrimaryGeneratedColumn()
    id: number;

    @Column()
    name: string;

    @Column()
    deviceName: string;

    @Column({ type: 'json' })
    config: JSON;
}

const input: any = {bothName: "ABC", protocol:["T1","T2]};

const filter: FindManyOptions<Item> = {};
const wheres: FindConditions<Item> = {};

if (input.protocol) {
    wheres.config = Raw(col => `JSON_EXTRACT(${col}, '$.type') IN (:protocols)`, {protocols: input.protocol});
}
if (input.bothName) {
    output.where = [
        {name: Like(`%${input.bothName}%`), ...wheres},
        {deviceName: Like(`%${input.bothName}%`), ...wheres }
    ]
} else {
    output.where = wheres;
}

getRepository(Item).find(filter);
```

### Expected Behavior

The order of SQL parameter should be:

``` sql
SELECT ...
WHERE 
(`Item`.`name` LIKE ? AND JSON_EXTRACT(`Item`.`config`, '$.type') IN (?)) OR
(`Item`.`deviceName` LIKE ? AND JSON_EXTRACT(`Item`.`config`, '$.type') IN (?)) 

---PARAMETERS: ["%ABC%", ["T1","T2],  "%ABC%",  ["T1","T2]]
```

### Actual Behavior

It instead return this query:
``` sql
-- PARAMETERS: ["%ABC%", "%ABC%", ["T1","T2], ["T1","T2]]
```

### My Environment

Operating System: ubuntu 18
Node.js version: v12.13.1
Typescript version: Version 3.9.5
TypeORM version: 0.2.29

### Relevant Database Driver(s)

- [ ] `aurora-data-api`
- [ ] `aurora-data-api-pg`
- [ ] `better-sqlite3`
- [ ] `cockroachdb`
- [ ] `cordova`
- [ ] `expo`
- [ ] `mongodb`
- [x] `mysql`
- [ ] `nativescript`
- [ ] `oracle`
- [ ] `postgres`
- [ ] `react-native`
- [ ] `sap`
- [ ] `sqlite`
- [ ] `sqlite-abstract`
[...]

Issue #7026, opened 2020-11-06
QueryBuilder broken on web - Invalid regular expression: invalid group specifier name


Vluf:
## Issue Description
The QueryBuilder is broken in the new 0.2.29 release due to this change: https://github.com/typeorm/typeorm/pull/4760
Regex lookbehinds are not supported in on a lot of web environments, a fallback should be implemented.
This affects any driver that uses JavascriptCore (Safari, React Native, ...)

### Expected Behavior

I'd expect QueryBuilder to work

### Actual Behavior

QueryBuilder crashes with `Invalid regular expression: invalid group specifier name`

### Steps to Reproduce

I would add an reproducible example, however, I feel the issue is clear enough without?

### My Environment

| Dependency          | Version  |
| ---                 | ---      |
| Operating System    |     React Native, Android     |
| Typescript version  | v4.0.5 | 
| TypeORM version     | v0.2.29 | 

### Additional Context

Relevant:
- https://bugs.webkit.org/show_bug.cgi?id=174931
- https://stackoverflow.com/questions/641407/javascript-negative-lookbehind-equivalent

### Relevant Database Driver(s)

- [ ] `aurora-data-api`
- [ ] `aurora-data-api-pg`
- [ ] `better-sqlite3`
- [ ] `cockroachdb`
- [ ] `cordova`
- [ ] `expo`
- [ ] `mongodb`
- [ ] `mysql`
- [ ] `nativescript`
- [ ] `oracle`
- [ ] `postgres`
- [x] `react-native`
- [ ] `sap`
- [ ] `sqlite`
- [ ] `sqlite-abstract`
- [ ] `sqljs`
- [ ] `sqlserver`

### Are you willing to resolve this issue by submitting a Pull Request?

- [ ] Yes, I have the time, and I know how to start.
- [ ] Yes, I have the time, but I don't know how to start. I would need guidance.
- [x] No, I don't have the time, although I believe I could do it if I had the time...
- [ ] No, I don't have the time and I wouldn't even know how to start.

imnotjames:
What version of node does react native provide?

imnotjames:
Looking at the react-native issues.. looks like the problem is brought up time and again - but never fixed.

We can work around it, but that sure does make things a bit messier..

Vluf:
React Native runs on [JavascriptCore](http://trac.webkit.org/wiki/JavaScriptCore), which still does not support lookbehinds

hhunaid:
[...]

Issue #6607, opened 2020-08-23
Parameter order is sometimes wrong in SQLite


hediet:
Thanks for creating and maintaining typeorm!

**Issue type:**

[x] bug report
[ ] feature request
[ ] documentation issue

**Database system/driver:**

[ ] `cordova`
[ ] `mongodb`
[ ] `mssql`
[ ] `mysql` / `mariadb`
[ ] `oracle`
[ ] `postgres`
[ ] `cockroachdb`
[x] `sqlite`
[ ] `sqljs`
[ ] `react-native`
[ ] `expo`

**TypeORM version:**

[x] `latest`
[ ] `@next`
[ ] `0.x.x` (or put your version here)

**Steps to reproduce or a small repository showing the problem:**

I have this typescript query:
```ts
const result = await this.connection.versionLanguages
			.createQueryBuilder("lang")
			/* (1) */    .where({ owningVersionId: v.id })
			/* (2) */ // .where("lang.owningVersionId = :whatTheFuck", { whatTheFuck: v.id })
			.leftJoin("lang.translatedFormats", "translation")
			.leftJoin(
				"translation.owningFormat",
				"translatable",
				"NOT(translatable.isStale)"
			)
			.leftJoin(
				"translatable.translatedFormats",
				"defaultTranslation",
				"defaultTranslation.owningLanguageId = :defaultLang AND translation.acceptedTranslation NOT NULL",
				{ defaultLang: v.defaultLanguage.id }
			)
			.groupBy("languageCode")
			.select("languageCode")
			.addSelect("SUM(translation.words)", "translatedWords")
			.addSelect(
				"SUM(defaultTranslation.words)",
				"defaultTranslatedWords"
			)
			.getRawMany<{
				languageCode: string;
				translatedWords: number;
				defaultTranslatedWords: number;
			}>();
```

This query mixes up the parameters `v.id` and `v.defaultLanguage.id`. When sent to the db, the parameters end up in the wrong order i.e. the query is complete garbage.
When commenting out `(1)` and uncommenting `(2)`, it works as expected.

Luckily, this query happened to be a `select` query and not a `delete` query where messing up condition parameters could delete really important stuff.

I guess the problem is [here](https://github.com/typeorm/typeorm/blob/8d90d403627f949dc8bf1bf32dc33c918b55f6eb/src/driver/sqlite-abstract/AbstractSqliteDriver.ts#L581-L585):
```ts
createParameter(parameterName: string, index: number): string {
[...]

PR #7282, opened 2021-01-16
feat: Expression builder (WIP)

nebkat:
### Depends on #7058 (refactor: Query builders and entity/column/relation metadata)
Necessary due to large amount of duplicate code before refactor.

***Only last commit is relevant to this PR! See #7058 for rest.***

### Description of change

This introduces the concept of "expressions" in TypeORM that represent SQL expressions, as well as an accompanying "expression builder" system. SQL expressions are found in a variety of places, but most importantly in the `WHERE` clause, `INSERT VALUES` and `UPDATE SET`. Currently only `QueryBuilder.where()` has basic support for constructing expressions using `FindOperators`.
 
Whenever "raw" expressions are used a number of major problems arise, all of which can be solved by the expression builder:
 
- The syntax of the expression cannot be checked for errors without sending it to the database engine
- Type checking entity column names isn't possible, somewhat defeating the purpose of using an ORM
- There is a major security risk if user provided values are not escaped by an inexperienced developer
- Column/property/alias name replacement is expensive (#4760, #7026), and can break when there is a conflict in naming
- Parameter value replacement is expensive, especially in long queries
- Maintaining native parameter order is difficult (#7256, #7104, #6607)
- Ugly code when building queries
 
# Implementation
_*Slightly simplified from actual implementation for better readability*_

## Base Types
```ts
// Raw values, such as 1, "Paris", TRUE or NULL, which will always get converted to query parameters(!)
type ExpressionLiteral = boolean | string | number | null;

// Class which produces an expression string when its build() method is called
abstract class ExpressionBuilder {
    abstract build(ctx: ExpressionBuildContext): string;
}

type Expression = ExpressionLiteral | ExpressionBuilder;
```

## Build Context
Certain tasks of the expression builder are left to be implemented by the query builder, via an abstract `ExpressionBuildContext`.

```ts
export abstract class ExpressionBuildContext {
    // Query builder will replace property name or path and escape
    abstract buildColumn(column: string, alias?: string): string;
    // Query builder will add values to the native parameters list and replace with `?` / `$1`
    abstract buildLiteral(column: string): string;
    // Query builder will replace all property names and parameters
    abstract buildRaw(expression: string): string;
}
```

# Builders
Every part of an expression that isn't a literal will be constructed using builders. These are functional building blocks that combine other expressions to form the final expression _in raw string form_.

Builders extend the `ExpressionBuilder` class and have a function to initialize them (i.e. `XYZ()` instead of `new XYZBuilder()`). Some builder functions may even have [tagged template literal](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals#tagged_templates) equivalents for conciseness.

Expressions will be built as the final step in constructing queries in the query builder, in the order that they appear in the query. Builders themselves will also build their sub-expressions in the order that they appear. This way, the order of native parameters will always be preserved. (#7256, #7104, #6607).

#### Raw
Raw unedited value that is passed as provided.

```ts
function Raw(raw: string): RawBuilder;
Raw("1 = 1 AND ?'!@@").build(ctx); /// Becomes: 
r`1 = 1 AND ?'!@@` // Tagged template literal variant
```
```sql
1 = 1 AND ?'!@@ # Note everything remains unescaped
```
_This is precisely what the expression builder is trying to avoid but it remains for users that would like to still have raw SQL access. It will also replace the `() => "RAW"` syntax, which should become deprecated._

#### Column
Explicit reference to a column name or property name/path.
```ts
function Col(): ColumnBuilder;
function Col(column: string): ColumnBuilder;
function Col(alias: string, column: string): ColumnBuilder;
Col("userId").build(ctx);
c`userId` // Tagged template literal variant
```
```sql
`userId` # Note the column name is escaped
```

If `Col()` is used, it becomes a reference to the current column within a where conditions object or inset/update values object.

`replacePropertyNames()` can be used accurately and efficiently because we know the value we are dealing with is some kind of column, as opposed to having to search the entire `WHERE` clause for matching property names using regular expressions.

#### Function
Function within a query such as `IF`, `LENGTH`, or `CURRENT_TIMESTAMP`. Known functions have initializers with fixed parameters, but user can also manually call an arbitrary function with arbitrary parameters.

```ts
function Fn(name: string, parameters: Expression[]): UserFunctionBuilder;
Fn("CONCAT", [c`firstName`, " ", c`lastName`]).build(ctx);
```
```sql
CONCAT(`firstName`, " ", `lastName`)
```

#### Operator
All operators in the form of `operand *operator* operand`.
```ts
function *Operator*(a: Expression, b: Expression): *Operator*Builder;
Equal(1, 2).build(ctx);
LessThan(1, 2).build(ctx);
Like(c`name`), "John").build(ctx);
// MoreThan, LessThanOrEqual, Is, etc are all the same
```
```sql
WHERE 1 = 2
WHERE 1 < 2
WHERE `name` LIKE "John"
```

#### And, Or, Xor
```ts
function And(...expressions: []): AndBuilder;
function Or(...expressions: []): OrBuilder;
function Xor(...expressions: []): XorBuilder;
And(c`isUnread`), LessThan(c`date`, "2020-10-01"), Or(false, 1)).build(ctx);
```
```sql
WHERE `isUnread` AND `date` < "2020-10-01" AND (FALSE OR 1)
```

# Find Conditions
Normally when we are dealing with entities and the `WHERE` clause, we define the find conditions using an object.

When we write:
 
```ts
cities.find({
    city: "Paris",
    country: "France"
})
```
 
It is the equivalent of `AND`ing every `key = value` of the object i.e.:

```ts
And(...Object.entries(conditions).map(([key, value]) => Equal(Col(key), value)));
And(Equal(c`city`, "Paris"), Equal(c`country`, "France"));
```
```sql
WHERE `city` = "Paris" AND `country` = "France"
```

#### Comparators
However there is an exception to this rule if we are using comparison builders:
 
```ts
cities.find({
    population: MoreThan(10000),
    country: Equal("France")
})
```
 
We don't want to check if `population = > 10000 AND country = = "France"`, but `population > 10000 AND country = "France"`. _Note also that where previously there were two arguments to `MoreThan` and `Equal` there is now only one._

This is a special variant of the comparator where only one argument is given instead of two. In this variant, the first argument becomes `Col()`, so that it references the current column based on the key in the object, and a flag is set to achieve the desired behavior of not checking for equality. We update the object mapping code:

```ts
And(...Object.entries(conditions).map(([key, value]) => {
        // If the find condition value was already comparing to the current column, use it directly
        if (value instanceof ExpressionBuiler && value.columnComparator) return value;
        // Otherwise check for equality
        return Equal(value);
}));
```

 Which gives us the distinction between:

```ts
cities.find({ population: MoreThan(10000); });
cities.find({ population: MoreThan(c`population`, 10000); });
```

```sql
WHERE `population` > 10000
WHERE `population` = (`population` > 10000) # Broken query here, but could be useful elsewhere
```

#### And, Or, Xor
One remaining special case is the ability to use `And`, `Or`, and `Xor` in combination with a find object (long been requested #2100).

```ts
cities.find({
    name: Or("Paris", "London", Like("A%"), And(Like("B%"), Like("%B")),
    country: In(["England", "France"])
}); 
```

```sql
WHERE (`name` = "Paris" OR `name` = "London" OR `name` LIKE "A%" OR (`name` LIKE "B%" AND `name` LIKE "%B")
AND `country` IN ("England", "France")
```

When used as part of a find conditions object, `And`, `Or`, and `Xor` are always _distributed_ over their column.

To avoid this behavior, there is a wrapper function:

```ts
// Does nothing to the expression being wrapped, but prevents the distributive property from applying
function Ex(expression: Expression): WrappedExpressionBuilder;

cities.find({
    name: Ex(Or("Paris", "London")),
})
```

```sql
WHERE `name` = ("Paris" OR "London") # Broken query, but might be intended in other queries
```

# Examples

_When used with query builder all values get replaced with parameters (`?`, `$1`)_

#### Find All
```ts
cities.find(true);
```

```sql
WHERE 1
```

#### Basic Conditions

```ts
cities.find(Equal(c`name`, "London"));
cities.find({name: "London"});
```

```sql
WHERE `city` = "London"
```

#### OR Conditions
```ts
cities.find(Or(Equal(c`name`, "London"), Equal(c`name`, "Paris"));
cities.find(Or({name: "London"}, {name: "Paris"}));
cities.find(Or({name: "London"}, Equal(c`name`, "Paris"));
cities.find({name: Or("London", "Paris")});
```

```sql
WHERE `city` = "London" OR `city` = "Paris"
```

#### Not
```ts
cities.find({
    name: Not("Paris"),
    population: Not(MoreThan(10000),
    country: Not(In(["Germany", "Spain", "Italy"]),
    capital: Not(Not(true))
    rural: Not(Coalesce(null, null, true))
})
```

```sql
WHERE
`name` != "Paris AND
`population` <= 10000 AND 
`country` NOT IN ("Germany", "Spain", "Italy") AND
capital = 1 AND
rural = Not(Coalesce(NULL, NULL, 1))
```

#### Nested With Column
```ts
cities.find({
    name: MoreThan(Length(), 10)
})
```

```sql
WHERE LENGTH(`name`) > 10
```

#### Case Expression
```ts
cities.find({
    population: Case(c`country`,
            When("France", 10000),
            When("England", 50000),
            Else(100000)
    )
})
```
```sql
WHERE population = (CASE `country` WHEN "France" THEN 10000 WHEN "England" THEN 50000 ELSE 100000 END)
```

#### Join Conditions

```ts
qb.leftJoin(AddressEntity, "address", Equal(c`address.id`, c`contact.addressId`));
qb.leftJoin(AddressEntity, "address", { id: c`contact.addressId` });
```

```sql
LEFT JOIN `addresses` ON `address`.`id` = `contact`.`addressId`;
```

#### Update Increment

```ts
users.update({
    id: 1,
}, {
    score: Plus(Col(), 1)
});
```

```sql
SET `score` = `score` + 1 WHERE `id` = 1
```

#### Update Reference

```ts
cities.update({
    country: "France"
}, {
    large: MoreThan(c`population`, 10000
});
```

```sql
SET `large` = `population` > 10000 WHERE `country` = "France
```

#### Column Default

```ts
@Column({default: CurrentTimestamp() }) createDate: Date;
@Column({default: r`CURRENT_TIMESTAMP`}) createDate: Date;
```

```sql
createDate DATETIME DEFAULT CURRENT_TIMESTAMP
```

### Pull-Request Checklist


nebkat:
@pleerock

This is about 90% complete now and I have identified some _very very_ significant benefits from introducing this system and its related changes:

----
- Adds `And()`/`Or()` builders to `find()`/`where()`
- Adds ability to filter by relation and join columns
- Adds ability to create very complex expressions previously only possible with raw expressions
- Adds ability to cleanly use expressions in `select()`, `orderBy()`, `groupBy()`, `where()` column defaults, insert/update values
- Adds ability to generate subqueries without breaking parameter ordering
----
- Improves performance by using targeted property name replacement only on column expressions and raw expressions
- Improves performance by using targeted parameter replacement only on parameter expressions and raw expressions
- Improves safety from SQL injections by parameterizing all values except those used in raw expressions
- Improves user experience by moving query construction to code, where missing brackets, typos, etc are noticed
----
- Eliminates all usage of TypeORM parameters in built in functions and replaces with native parameters
- Eliminates all manual query construction in relations/joins
- Eliminates all problems with native parameter ordering
- Eliminates a large amount of duplicate code
----

I have also identified countless issues that can directly be solved with this: #7280, #7256, #7248, #7247, #7226, #7210, from the first 2 pages of issues alone 🎉 

Amazingly I believe it may even be possible to achieve all of this without introducing any breaking changes relative to the current version!

That said, a bunch of features remain that would effectively become obsolete, and would only make it more difficult to implement:
- `SelectQueryBuilder.select(select: string)`
    - Previously does not differentiate between column/property name/path and a raw expression
    - Now it should assume that a column is being selected, and if user wants expressions they can pass `qb.select(Raw("..."))`
    - Better yet, `qb.select(SomeExpressionBuild())`
    - If no breaking changes allowed, it must assume that a string could be either a column or an expression
- `SelectQueryBuilder.where(where: Brackets)`
    - Previously used for `qb.andWhere(new Brackets(qb => qb.where("1 = 2").orWhere("2 < 3")))`
    - Replaced by expressions `qb.andWhere(Or(Equal(1, 2), LessThan(2, 3))`
    - If no breaking changes allowed, all `Brackets` inputs have their wheres converted to expressions as above
- `SelectQueryBuilder.where(where: (qb: QueryBuilder) => string)`
    - Previously used for `qb.andWhere(qb => { const sq = qb.subQuery() ... return "post.title IN " + sq.getQuery() })`
    - Replaced by expressions `qb.andWhere(In(Col("post", "title"), SubQuery(qb => ...)))`
    - Better yet, `qb.andWhere({title: In(SubQuery(qb => ...))})`
    - If no breaking changes allowed, mapped to `Raw(where(this))`
- `SelectQueryBuilder.where(where: string)`
    - Previously used for passing string expression `qb.andWhere("post.title = :title", { title: "Title" })`
    - Replaced by expression `qb.andWhere(Equal(Col("post", "title", "Title")`, or `qb.andWhere(Raw("post.title = :title"), { title: "Title" })`
    - Better yet, `qb.andWhere({title: "Title"})`
    - If no breaking changes allowed, all string inputs are mapped to `Raw(where)`
- `SelectQueryBuilder.join(..., on: string)`
    - Previously used for `qb.leftJoin(..., "user", "user.id = post.userId")`
    - Replaced by expressions `qb.leftJoin(..., "user", Equal(Col("user", "id"), Col("post", "userId")))`
    - If no breaking changes allowed, all string inputs are mapped to `Raw(on)`
- `SelectQueryBuilder.orderBy(orderBy: string)`
    - Previously does not differentiate between column/property name/path and a raw expression
    - Now it should assume ordering by a column, and if user wants expression they can pass `qb.orderBy(Raw("..."))`
    - If no breaking changes allowed, all string inputs are mapped to `Raw(orderBy)`
- `Insert/UpdateQueryBuilder.set(values: { [key: string]: () => string })`
    - Previously used for passing raw expressions as values `qb.set({count: () => "count + 1"})`
    - Replaced by expressions `qb.set({count: Plus(Col(), 1)})`
    - If no breaking changes allowed, mapped to `Raw(value())`
 - `@Column({default: (() => string)})`
     - Previously used for passing raw expressions as values `@Column({default: () => "CURRENT_TIMESTAMP"})`
     - Replaced by expressions `@Column({default: CurrentTimestamp()})`
     - If no breaking changes allowed, mapped to `Raw(default())`

Considering this, it might be better to target the `next` branch, but I would first like to get your opinion.

Is there a possibility that we could work towards the release of `next` in the near future if a few features were to get merged, and if so what level of breaking changes would we be willing to accept? Aside from this if #7076 (upsert support) and #7010 (select and map) were to be merged there would already be a significant difference between the two versions that would be worthy of a major release.

draaglom:
you love to see it!

>Improves performance by using targeted property name replacement only on column expressions and raw expressions

Ever since looking into #4760 I was hoping someone would tackle this -- those regexes in replacePropertyNames dominate performance above a certain query rate, but the generic/blunt regex approach was necessary given the current QueryBuilder.

pleerock:
All problems you have listed are super important and it will be great if we solve them.

I'm not sure I can review this PR, however I read your description.
Everything looks good, but one important note:

```ts
cities.find({
    name: Not("Paris")
})
```

Here `Not` should only accept `typeof City.name`. And in this example: `cities.find(Or({name: "London"}))` - `{ name: ... }` must be `typeof City`. In order to implement it expression must have a generic type, and `find` should only accept expressions of `Expression<City>`. I'm not sure how it will play well with nested expressions, but we should try to make type-safe as much as possible.

michaelbromley:
Just chiming in here to say that this looks really great and would solve some particularly tricky issues I am dealing with. Has any decision been made about whether this will go in `next` or `master`? 

If there is any outside help you need which might expedite the merging of this work, please let me know!

pleerock:
@michaelbromley I'm currently experimenting on how I can make it completely type-safe on my own branch.

PR #7290, opened 2021-01-17
implemented formatting an entire query

### LINKED ISSUE #7271: Full formatting for whole query
## Feature Description
1) to implement format the query with templates 
```query('SELECT * FROM user WHERE user.id = :id').setParameters({id:1})```

2)  to implement method setEntities. which puts the entity target to sql, like "from" and allows us to make aliases for names (it is very convenient to use camelCase in names)
```query('SELECT * FROM user WHERE user.id = :id').setEntities({user: UserEntity})```

### The Problem

When I need to make long queries, the builder becomes very hard to read. I would like to format the whole query like this

```
getConnection()
      .createQueryBuilder()
      .setQuery(
        `
      SELECT user.id, user.name, userRole.name, 
      FROM user
      LEFT JOIN userRole.id = user.roleId
      WHERE userRole.name  = :roleName AND user.deviceId = :deviceId
      LIMIT 10
    `,
      )
      .setParameters({ deviceId: 1, roleName: 'user' })
      .setEntities({ user: UserEntity, userRole: UserRoleEntity })
```

- [ ] `aurora-data-api`
- [ ] `aurora-data-api-pg`
- [ ] `better-sqlite3`
- [ ] `cockroachdb`
- [ ] `cordova`
- [ ] `expo`
- [ ] `mongodb`
- [ ] `mysql`
- [ ] `nativescript`
- [ ] `oracle`
- [ x ] `postgres`
- [ ] `react-native`
- [ ] `sap`
- [ ] `sqlite`
- [ ] `sqlite-abstract`
- [ ] `sqljs`
- [ ] `sqlserver`

- [ x ] No, I don't have the time, although I believe I could do it if I had the time...

### PR DESCRIPTION
**Pull request by stepanzabelin**
The ability to set an entire sql query has been added to the builder. Along the way, the issue with the aliases of entities has been resolved.  Tables and columns are associated with entities using ampersand

### Description of change
UPD
1. implemented class QueryFormatBuilder class that extends SelectQueryBuilder
2. implemented QueryFormatter that has limited properties of QueryFormatBuilder
3. implemented createQueryFormatter method  that returns new QueryFormatter
4. QueryFormatter has some extended methods and new methods:
 -  "setQuery" that sets the entire query 
 - "setAliases", "addAlias"  that bind entities to aliases
5. Added tests

### The Problem

When I started using this library I really liked it, but over time my queries became more complex, longer and contain many subqueries. the resulting code is unreadable. it cannot be supported normally.
I would like to be able to use  "select" database queries without unnecessary sugar ( to exec pure queries of any kind with parameters )

###  The Solution

"setQuery" takes an entire query with templates

```
.createCustomQuery()
.setQuery(`SELECT "users"."nick_name" FROM "users" WHERE "users"."id" = :id`)
.setParameters({id:1})

// SELECT "users"."nick_name"  FROM "users" WHERE "users"."id" = '1'
```

I added a new construction with the ampersand.  This is needed in order to associate the name with the entity

For example, We have the following entity

```
@Entity("users")
export class UserEntity {
    @PrimaryGeneratedColumn()
    id?: number;

    @Index({ unique: true })
    @Column({
        type: "varchar",
        length: 64,
        name: "nick_name",
    })
    nickName: string;
}
```

to associate with a sql, we need to use the methods "addAlias" , "setAliases" and use the ampersand
``` SELECT &user.nickName  FROM &user WHERE user.id = :id ```

Table name alias `&user` will be replaced  by `"users" "user"`
Table columns alias `&user.nickName` will be replaced  by `"user"."nick_name" as 'nickName'`

```
.createCustomQuery()
.setQuery(`SELECT &user.nickName  FROM &user WHERE user.id = :id`)
.setParameters({id:1})
.addAlias(UserEntity, 'user')

// SELECT  "user"."nick_name" as  "nickName" FROM "users" "user" WHERE "user"."id" = '1'
```
also i added the ability to map all the column names and make them aliases matching with entity property names using  construction: ``` &alias.*```

```
.createCustomQuery()
.setQuery(`SELECT &user.*  FROM &user WHERE user.id = :id`)
.setParameters({id:1})
.addAlias(UserEntity, 'user')

// SELECT  "user"."id" as  "id" ,  "user"."nick_name" as  "nickName" FROM "users" "user" WHERE "user"."id" = '1'
```
#### More examples
```
await connection.createCustomQuery(`
    SELECT UPPER(CONCAT('Hello', ' ', 'World')) as "text";
  `).execute();
```

```
const insertResults = await connection.createCustomQuery(`
    INSERT INTO "users" ( email, role_id, password_hash, nick_name )
    VALUES (:email, :roleId, :passwordHash, :nickName)
    RETURNING *;
  `, {
     email: '666666@test.test',
     roleId:1,
     passwordHash: '',
     nickName: 'Jane'
   }).execute();
```

### Pull-Request Checklist

- [x] This pull request links relevant issues as `Fixes #7271`

### CODE REVIEW

**Comment by nebkat** [comment]
I like the ability to use `&user` as an alias, but I'm not sure about the overall idea.

I think if your code is becoming unreadable we need to improve the existing query builders so that that is no longer the case. If we accept something like this I worry that there will be much less incentive for others to improve things and people will end up using custom queries at the slightest inconvenience. I have a PR submitted for expression building https://github.com/typeorm/typeorm/pull/7282 which might be useful for you in making advanced queries a bit more bearable.

However if this does get added it should definitely have its own `QueryBuilder`, perhaps `CustomQueryBuilder`, rather than using `SelectQueryBuilder`.

**Comment by stepanzabelin** [comment]
Thanks @nebkat
i restored the original code and implemented createQueryFormatter (  class QueryFormatter ). It got much better

File: src/query-builder/QueryFormatBuilder.ts [hunk 1]
```diff
@@ -0,0 +1,120 @@
...
+import {SelectQueryBuilder} from "./SelectQueryBuilder";
+
+/**
+ * Allows to build full sql queries 
+ */
+export interface AliasesLiteral { 
+    [aliasName: string]: EntityTarget<any> 
+};
+
+export class QueryFormatBuilder extends SelectQueryBuilder<any> {
```
nebkat, line 11: It should extend just `QueryBuilder` since it doesn't use any features from `SelectQueryBuilder`.

src/query-builder/QueryFormatter.ts
```diff
@@ -0,0 +1,94 @@
+import {ObjectLiteral} from "../common/ObjectLiteral";
+import {Connection} from "../connection/Connection";
+import {EntityTarget} from "../common/EntityTarget";
+import {AliasesLiteral, QueryFormatBuilder} from "./QueryFormatBuilder";
+
+/**
+ * Allows to build full sql queries 
+ */
+export class QueryFormatter {
```
nebkat, line 9: What is the purpose of this class if it seems to just wrap everything that the `QueryFormatBuilder` does?

src/query-builder/QueryFormatBuilder.ts
nebkat: Also still think `CustomQueryBuilder` would be a more accurate description.

nebkat (changes requested):
Still unsure about the idea but here are some more comments anyway.

src/query-builder/QueryFormatBuilder.ts
stepanzabelin: 1) I've changed the description into CustomQueryBuilder. Looks better.
2) I've changed the extending into QueryBuilder instead SelectQueryBuilder. Functions "getRawOne" , "getRawMany" became unavailable, I had to duplicate them in CustomQueryBuilder

src/query-builder/QueryFormatter.ts
stepanzabelin: yes, yes it looks a little weird. I'll explain now. I tried not to change the code of QueryBuilder . so I created "CustomQueryBuilder" (that extends "QueryBuilder")  and rewrote some of the methods. but CustomQueryBuilder has a lot of unnecessary methods for this issue, so I made  class "CustomQuery" that takes some of the methods needed.  Maybe you can suggest a better solution

stepanzabelin:
Thans @nebkat   for code review. I took into account your notes and refactored the code

pleerock:
Thank you for your efforts! However I don't think we need this feature. I want to keep query builder pure SQL, without our own "features". And for sure I want to avoid `replacing` something in SQL. If we want to simplify using sub-queries, probably we need to came up with a new api-based query builder implementation. But this isn't easy to design and implement.

pleerock:
Closing it for now until it get more user feedback. Thank you!

PR #7295, opened 2021-01-18
useUTC for oracle and postgres

Alankarsharma:
If useUTC is set to true, then for Oracle set environment variable 'ORA_SDTZ' and for postgres set 'PGTZ'.

Based on documentation I thought 'useUTC' was used for all databases, but it was actually set only for MSSQL.
If this option is set to true then I am setting environment variables for oracle and Postgres.

### Pull-Request Checklist

### DIFF
--- a/src/driver/oracle/OracleConnectionOptions.ts
+++ b/src/driver/oracle/OracleConnectionOptions.ts
@@ -17,6 +17,11 @@ export interface OracleConnectionOptions extends BaseConnectionOptions, OracleCo
     readonly schema?: string;
 
     /**
+    * A boolean determining whether to pass time values in UTC or local time. (default: true).
+    */
+    readonly useUTC?: boolean;
+
+    /**
      * Replication setup.
      */
     readonly replication?: {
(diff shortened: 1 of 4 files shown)


pleerock:
Thank you for contribution!
It would be good if we'll have a documentation for this new option.

src/driver/oracle/OracleConnectionOptions.ts
```diff
@@ -16,6 +16,11 @@ export interface OracleConnectionOptions extends BaseConnectionOptions, OracleCo
      */
     readonly schema?: string;
 
+    /**
+    * A boolean determining whether to pass time values in UTC or local time. (default: true).
```
silentroach, line 20: isn't it a breaking change?

pleerock: there is no code in this PR where this option is set to `true` explicitly. Maybe comment is just wrong @Alankarsharma ?

Alankarsharma: this is part of connection options. If user set this value to true in that case I am setting those environment variables.

Alankarsharma: Default behaviour is false, so anyone who is using this for oracle or Postgres will not face any issue.

silentroach: so it will be great to change comments and documentation

PR #7296, opened 2021-01-19
fix: resolve issue with find with relations returns soft-deleted entities

### LINKED ISSUE #6265: find with relations returns soft-deleted entities
**Issue type:**

[x] question
[ ] bug report
[ ] feature request
[ ] documentation issue

**Database system/driver:**

[ ] `cordova`
[ ] `mongodb`
[ ] `mssql`
[ ] `mysql` / `mariadb`
[ ] `oracle`
[x] `postgres`
[ ] `cockroachdb`
[ ] `sqlite`
[ ] `sqljs`
[ ] `react-native`
[ ] `expo`

**TypeORM version:**

[x] `latest`
[ ] `@next`

Apologies for opening an issue for this, it seems like I am probably missing something very obvious but essentially the issue is that soft-deleted entities are being returned as part of relations.

Example:

Say we have "User" and "Post" - user has many posts

1. `const userWithPosts = await userRepo.findOne(1, { relations: [ 'posts' ] })`
2. `userWithPosts.posts` will contain all posts, even ones that have been deleted

Am I missing something obvious here? This functionality works as expected when not using relations (e.g. in the above example, `postRepo.find()` would only return non-deleted items) but when using relations I am running into this issue.

luiseariass:
### Description of change
This pull request includes fix for soft delete behavior when including soft deletes relationships. 
fix: resolve issue with find with relations returns soft-deleted entities #6265
### Pull-Request Checklist

- [x] This pull request links relevant issues as `Fixes #6265`

### DIFF
--- a/src/query-builder/SelectQueryBuilder.ts
+++ b/src/query-builder/SelectQueryBuilder.ts
@@ -1336,7 +1336,10 @@ export class SelectQueryBuilder<Entity> extends QueryBuilder<Entity> implements 
         this.expressionMap.joinAttributes.push(joinAttribute);
 
         if (joinAttribute.metadata) {
-
+           if (joinAttribute.metadata.deleteDateColumn && !this.expressionMap.withDeleted) {
+                const conditionDeleteColumn = `${aliasName}.${joinAttribute.metadata.deleteDateColumn.propertyName} IS NULL`;
+                joinAttribute.condition += joinAttribute.condition ? ` AND ${conditionDeleteColumn}`: `${conditionDeleteColumn}`;
+            }
             // todo: find and set metadata right there?
             joinAttribute.alias = this.expressionMap.createAlias({
                 type: "join",
--- a/test/github-issues/6265/entity/Role.ts
+++ b/test/github-issues/6265/entity/Role.ts
@@ -1,0 +1,23 @@
+import { Entity } from "../../../../src/decorator/entity/Entity";
+import {
+    OneToMany,
+    Column,
+    DeleteDateColumn,
+    PrimaryGeneratedColumn,
+} from "../../../../src";
+import { User } from "./User";
+
+@Entity()
+export class Role {
+    @PrimaryGeneratedColumn()
+    id: string;
+
+    @Column()
+    title: string;
+
+    @OneToMany((_) => User, (user) => user.role, { cascade: true })
+    users: User[];
+
+    @DeleteDateColumn()
+    deleteDate?: Date;
+}
--- a/test/github-issues/6265/entity/User.ts
+++ b/test/github-issues/6265/entity/User.ts
@@ -1,0 +1,23 @@
+import { Entity } from "../../../../src/decorator/entity/Entity";
+import {
+    ManyToOne,
+    Column,
+    DeleteDateColumn,
+    PrimaryGeneratedColumn,
+} from "../../../../src";
+import { Role } from "./Role";
+
+@Entity()
+export class User {
+    @PrimaryGeneratedColumn()
+    id: number;
+
+    @Column()
+    name: string;
+
+    @ManyToOne((_) => Role, (role) => role.users)
+    role: Role;
+
+    @DeleteDateColumn()
+    deleteAt?: Date;
+}
--- a/test/github-issues/6265/issue-6265.ts
+++ b/test/github-issues/6265/issue-6265.ts
@@ -1,0 +1,97 @@
+import "reflect-metadata";
+import { expect } from "chai";
+import { Connection } from "../../../src";
+import { User } from "./entity/User";
+import { Role } from "./entity/Role";
+import {
+    createTestingConnections,
+    reloadTestingDatabases,
+    closeTestingConnections,
+} from "../../utils/test-utils";
+
+describe("github issues > #6265 `fix: resolve issue with find with relations returns soft-deleted entities", () => {
+    let connections: Connection[];
+
+    before(async () => {
+        connections = await createTestingConnections({
+            entities: [User, Role],
+            schemaCreate: true,
+            dropSchema: true,
+        });
+    });
+    beforeEach(() => reloadTestingDatabases(connections));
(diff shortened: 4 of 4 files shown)


src/query-builder/SelectQueryBuilder.ts
```diff
@@ -1336,7 +1336,10 @@ export class SelectQueryBuilder<Entity> extends QueryBuilder<Entity> implements
         this.expressionMap.joinAttributes.push(joinAttribute);
 
         if (joinAttribute.metadata) {
-
+           if (joinAttribute.metadata.deleteDateColumn && !this.expressionMap.withDeleted) {
+                const conditionDeleteColumn = `${aliasName}.${joinAttribute.metadata.deleteDateColumn.propertyName} is null`;
```
nebkat, line 1340: Use all caps for SQL, i.e. `IS NULL`, `AND`.

luiseariass: done

nebkat (approved):
Ideally squash the two commits into one now but otherwise looks good to me 🎉

nebkat:
Actually you should also add tests to verify the functionality, there should already be some examples testing non relation soft deleted items that you can mostly copy paste.

caiomestres:
> Tests look reasonable. Strictly this should also be a "feat:" rather than "fix:" but not so important. One thing that will be an issue is that this is essentially a breaking change for current users so not sure how to work around that.

A bool property for nested soft deletes wouldn't work @nebkat ?

something like 
userRepository.find({ relations: ["profile", "photos", "videos"], nestedSoftDelete: true });

default would be false, that way maintaining backward compatibility

nebkat:
That works but it must be covered for both find and query builder, and it must also match the current naming (`withDeleted` I think).

luiseariass:
I propose we add a new flag `withDeletedRelations: boolean` to the @DeleteDateColumn decorator, with a default value of false. This way we don't affect current integrations.
The option `withDeleted: true` should overwrite `withDeletedRelations: true` behavior.

nebkat:
I don't think it should be on `@DeleteDateColumn` since that applies to the entity being deleted whether it is a relation or not. Whether deleted relations are included should be down to each query, and in future the default should switch to not including soft deleted relations.

luiseariass:
I was just looking for a way not to affect the current behavior, but I fervently believe that the behavior of the relationships should be the same as the main entity, we can work on updating the documentation of when the join is used and when it is used the soft delete

nebkat:
Personally I think there should just be a new major release with the breaking change but that doesn't seem to be planned so we have to work around it unfortunately.

jdniels:
it's fixed?

RAllner:
Hey guys. Is there any news on this PR? Looking forward to use it, since I could get rid of a bigger list of workarounds.

caiomestres:
> Hey guys. Are there any news on this PR? Looking forward to use it, since I could get rid of a bigger list of workarounds.

Waiting for @nebkat to deliberate about the best approach...

brennedith:
@nebkat, following on @caiomestres' train of thought:

Find
```javascript
const firstUser = await User.findOne({
  where: {
    id: 1,
  },
  withDeletedRelations: false,
});
```

QueryBuilder
```javascript
const firstUser = await connection
  .getRepository(User)
  .createQueryBuilder('user')
  .withDeletedRelations(false)
  .where('user.id = :id', { id: 1 })
  .getOne();
```

Is this something that can be achieved? If so, can you suggest to us where to start?

Thanks in advance

nebkat:
Yes that looks like the ideal solution.

pleerock:
I don't think we need extra flag at least because it can raise questions like "how to do it with deleted relations but without deleted relations of relations".

Also I don't think we need a backward, because without this change current behaviour looks like a bug.

Thank you for contribution, I'm going to merge this one.

luiseariass:
when is the next release @pleerock?

pleerock:
Available in the latest `0.2.31` version.

stovberpv:
Seems to work but `withDeleted: true` doesn't.

When i want to query deleted rows and use:

```javascript
builder.withDeleted();
```

the query builds as follows:

```sql
SELECT *
  FROM plan
 INNER JOIN blank ON blank.id = plan.blank_id
                 AND (blank.deleted_at IS NULL)            <-- still here
```

luiseariass:
> Seems to work but `withDeleted: true` doesn't.
> 
> When i want to query deleted rows and use:
> 
> ```js
> builder.withDeleted();
> ```
> 
> the query builds as follows:
> 
> ```sql
> SELECT *
>   FROM plan
>  INNER JOIN blank ON blank.id = plan.blank_id
>                  AND (blank.deleted_at IS NULL)            <-- still here
> ```

What query in typeom you use?

stovberpv:
@luiseariass i have 2 entities with one-to-one relation

```javascript
@Entity()
class Plan {
  @OneToOne(() => Blank, blank => blank.id)
  @JoinColumn({ name: 'blank_id' })
  blank: Blank;
}

@Entity()
class Blank {
  @OneToOne(() => Plan, plan => plan.blank)
  plan: Plan;

  @DeleteDateColumn()
  deleted_at: Date;
}
```

And query builder looks like this

```javascript
const res = await this.manager
  .getRepository(Plan)
  .createQueryBuilder('plan')
  .innerJoinAndSelect(`plan.blank`, 'blank')
  .withDeleted()
  .getMany();
```

stovberpv:
i figure it out, the `withDeleted()` needs to be added before the `innerJoinAndSelect`;

jdniels:
thank you for fix the bug ////

br-dr3:
Any suggestion on how to bring relation deleted without flag? Many soft deletes just exist to continue showing some text of a deleted entity..

siamahnaf:
I am using latest version. But not seeing any `nesteDeleted` flag. I just need to enable withDelete for only relations table not the current table. It merged PR, one said it is availe from `0.2.31` current version is '0.3.20`

PR #7298, opened 2021-01-19
better postgres uuid column handling

### LINKED ISSUE #6039: Postgresql UUID Default value missing on initial migration generation.
**Issue type:**

[ ] question
[ X] bug report
[ ] feature request
[ ] documentation issue

**Database system/driver:**

[ ] `cordova`
[ ] `mongodb`
[ ] `mssql`
[ ] `mysql` / `mariadb`
[ ] `oracle`
[ X] `postgres`
[ ] `cockroachdb`
[ ] `sqlite`
[ ] `sqljs`
[ ] `react-native`
[ ] `expo`

**TypeORM version:**

[ ] `latest`
[ ] `@next`
[ X] `0.2.24`

Create a basic entity with a primary generated column of the UUID type.

```ts
@Entity('tenant')
export class Tenant {
  @PrimaryGeneratedColumn({ name: 'tenant_id', type: 'uuid' })
  id: string;

  @CreateDateColumn({ name: 'created_on', type: 'timestamp' })
  createdOn: Date;
}
```

Generate a migration.

```
typeorm migration:generate -n "test"
```

View the contents of the migration .ts file. Note that `tenant_id` has the `NOT NULL` attribute, with no sequence or default value.

```ts
export class test1589086562460 implements MigrationInterface {
    name = 'test1589086562460'

        await queryRunner.query(`CREATE TABLE "tenant" ("tenant_id" uuid NOT NULL, "name" text NOT NULL, "url_name" text NOT NULL, "status" character NOT NULL, "created_on" TIMESTAMP NOT NULL DEFAULT now(), CONSTRAINT "PK_4e7cb4e84f82aa7842a0bafc673" PRIMARY KEY ("tenant_id"))`, undefined);
    }

        await queryRunner.query(`DROP TABLE "tenant"`, undefined);
    }

}
```

Run the migration, generate a second migration.

```
typeorm migration:run test
typeorm migration:generate -n second-test
```

Review the contents of the `second-test` migration .ts file. Note the `CREATE SEQUENCE` query followed by the `ALTER TABLE` query. At this point we have made no change to our entity, however we have pending database changes that were not included on the initial migration.

```ts

export class secondTest1589086875150 implements MigrationInterface {
    name = 'secondTest1589086875150'

        await queryRunner.query(`CREATE SEQUENCE "tenant_tenant_id_seq" OWNED BY "tenant"."tenant_id"`, undefined);
        await queryRunner.query(`ALTER TABLE "tenant" ALTER COLUMN "tenant_id" SET DEFAULT nextval('tenant_tenant_id_seq')`, undefined);
    }

        await queryRunner.query(`ALTER TABLE "tenant" ALTER COLUMN "tenant_id" DROP DEFAULT`, undefined);
        await queryRunner.query(`DROP SEQUENCE "tenant_tenant_id_seq"`, undefined);
    }

}

```

It is my belief that the second migration should be empty as no changes have been made to any entities after the first migration.

salzhrani:
### Description of change

passing `{type: "uuid"}` as a parameter to `PrimaryGeneratedColumn` is valid given the TS types. but strategy is set to `increment` instead of `uuid`
this change fixes it. I think it would be worthwhile to tighten the parameters type definition instead of having a large union
Fixes #6039
### Pull-Request Checklist

### DIFF
--- a/src/decorator/columns/PrimaryGeneratedColumn.ts
+++ b/src/decorator/columns/PrimaryGeneratedColumn.ts
@@ -44,7 +44,7 @@ export function PrimaryGeneratedColumn(strategyOrOptions?: "increment"|"uuid"|"r
             strategy = strategyOrOptions as "increment"|"uuid"|"rowid";
 
         if (strategyOrOptions instanceof Object) {
-            strategy = strategyOrOptions.type === "uuid" ? "uuid" : "increment";
+            strategy = "type" in strategyOrOptions && strategyOrOptions.type === "uuid" ? "uuid" : "increment";
             Object.assign(options, strategyOrOptions);
         }
     } else {
--- a/src/decorator/columns/PrimaryGeneratedColumn.ts
+++ b/src/decorator/columns/PrimaryGeneratedColumn.ts
@@ -44,7 +44,7 @@ export function PrimaryGeneratedColumn(strategyOrOptions?: "increment"|"uuid"|"r
             strategy = strategyOrOptions as "increment"|"uuid"|"rowid";
 
         if (strategyOrOptions instanceof Object) {
-            strategy = "increment";
+            strategy = strategyOrOptions.type === "uuid" ? "uuid" : "increment";
             Object.assign(options, strategyOrOptions);
         }
     } else {
--- a/src/decorator/columns/PrimaryGeneratedColumn.ts
+++ b/src/decorator/columns/PrimaryGeneratedColumn.ts
@@ -44,7 +44,7 @@ export function PrimaryGeneratedColumn(strategyOrOptions?: "increment"|"uuid"|"r
             strategy = strategyOrOptions as "increment"|"uuid"|"rowid";
 
         if (strategyOrOptions instanceof Object) {
-            strategy = "type" in strategyOrOptions && strategyOrOptions.type === "uuid" ? "uuid" : "increment";
+            strategy = "increment";
             Object.assign(options, strategyOrOptions);
         }
     } else {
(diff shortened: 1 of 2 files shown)


nebkat:
The `type` there actually refers only to the type of the column, not the generation strategy. If anything `"uuid"` should be removed as a valid option for `type` as it doesn't make sense to use it with the increment strategy. If we were to apply this change the strategy could be set to any of the values here:

https://github.com/typeorm/typeorm/blob/af2c5c4dbe360d2cdcdb3398f5000a2ef0999d3d/src/driver/types/ColumnTypes.ts#L1-L19

salzhrani:
do you prefer to fix the parameter type definition? would that be a breaking change ?

nebkat:
It is technically a breaking change but unless someone has an example of how `@PrimaryGeneratedColumn("increment", {type: "uuid"})` works it looks like a bug and should be removed.

salzhrani:
`uuid` is a valid column type in postgres.
how about we check for the cases where `uuid` is the first parameter and when it is passed as type with a different strategy. and maybe force having an object as a single parameter for the decorator with the next breaking change?

nebkat:
`uuid` is a valid column type but how would an auto incrementing column work with it? If the values are sequential then it defeats the purpose of using a UUID.

The correct way to use the `uuid` is just `@PrimaryGeneratedColumn("uuid")`. The decorator assumes if the first parameter is an object that the user wanted an "increment" generated column, so it treats it the same as passing `@PrimaryGeneratedColumn("increment", options)`.

salzhrani:
Whats your recommendation on solving this issue?. just remove `uuid` from `PrimaryGeneratedColumnType`?

salzhrani:
Any other changes would you like me to add?

nebkat:
That's all, just squash your commits into one and make sure you follow the commit message format in [CONTRIBUTING.md](https://github.com/typeorm/typeorm/blob/master/CONTRIBUTING.md) 🎉

pleerock:
It took me quite a while to figure this single-line change.

I can't understand why we did this #4043 change (liked by 3 people) two years ago.

nebkat:
This looks like an easy mistake to overlook, the type is `PrimaryGeneratedColumnType` rather than `PrimaryGeneratedIncrementColumnType` so it is confusing.

PR #7318, opened 2021-01-26
Fixing performance issues of `RelationId`.

### LINKED ISSUE #5691: RelationId is too slow

satanTime:
### Description of change

Fixes #5691

Fixing performance issues of `RelationId`.
For example, we have users and a company.
We want to select the company with relation ids.
There are 100k users, and all of them belong to the same company.

So we select `users.company`,
it selects 100k users and 100k companies, but these 100k companies are the same entity,
because all users belong to the same company.

Let's imagine, that the company has adminId, and we want it to be selected via `RelationId`.

The issue is that currently typeorm simply joins ids of entities,
so the query has 100k times `company.id = NNN or company.id = NNN or ...`, whereas NNN is the same id,
and because the query is very long, MySQL dies when parses it.

The changes reduce ids to be an unique array, so 100k ids will become a single id.

### Pull-Request Checklist

### DIFF
--- a/src/query-builder/relation-id/RelationIdLoader.ts
+++ b/src/query-builder/relation-id/RelationIdLoader.ts
@@ -32,17 +32,34 @@ export class RelationIdLoader {
                 if (relationIdAttr.queryBuilderFactory)
                     throw new Error("Additional condition can not be used with ManyToOne or OneToOne owner relations.");
 
+                const duplicates: Array<string> = [];
                 const results = rawEntities.map(rawEntity => {
                     const result: ObjectLiteral = {};
+                    const duplicateParts: Array<string> = [];
                     relationIdAttr.relation.joinColumns.forEach(joinColumn => {
                         result[joinColumn.databaseName] = this.connection.driver.prepareHydratedValue(rawEntity[DriverUtils.buildColumnAlias(this.connection.driver, relationIdAttr.parentAlias, joinColumn.databaseName)], joinColumn.referencedColumn!);
+                        const duplicatePart = `${joinColumn.databaseName}:${result[joinColumn.databaseName]}`;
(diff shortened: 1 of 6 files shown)


pleerock:
Thank you for contribution.

I'm not sure I should merge this PR because last time we wanted to deprecate this file in favor of [another RelationIdLoader](https://github.com/typeorm/typeorm/blob/master/src/query-builder/RelationIdLoader.ts) we already have. I believe it already have such optimization. We just need to replace old usages to new usages and make necessary modifications.

satanTime:
Hi @pleerock,

thanks, I'll take a look how it works and update the case.

My idea was - if you plan to deprecate / remove `RelationId` in near future, then it should be a big issue to merge the related PR, because it shouldn't cause any additional issues.

TheNoim:
> My idea was - if you plan to deprecate / remove RelationId in near future, then it should not be a big issue to merge the PR, because it should not cause any additional issues.

Even if it gets deprecated, in its last version with this feature, it should be in it's best shape and best performance.

pleerock:
Thank you for contribution! 🎉

satanTime:
Thanks @pleerock, happy coding!

PR #7373, opened 2021-02-14
bugfix: skip register many to many inverse relationship

mluxe:
### Description of change

this is a fix to registerJunctionEntityMetadata on many to many relationship

registerJunctionEntityMetadata applies relationship metadata information to relevant entities. 

in case of many to many relationship entities have inverse relationship to each other so that Entity B overwrites the junctionEntityMetadata on Entity A. 

with this behaviour on many-to-many relationship, the code is able to save data into database with code
`entityARepository.save(entityA)` but it will not be able to save data with `entityBRepository.save(entityB)`

to fix this, we need to skip inverse relationship assignment on many-to-many relationship

```ts
@Entity({ name: "Datasource" })
export class Datasource {
	@PrimaryGeneratedColumn()
	id?: number;
	@Column("varchar", { length: 255 })
	name?: string

    @ManyToMany(type => Edr, edr => edr.datasources)
    @JoinTable({
        name: 'EdrDatasourceMaps',
        joinColumn: { name: 'DatasourceId' },
        inverseJoinColumn: { name: 'EdrId' }
     })
    edrs?: Edr[];
}


@Entity({ name: "EDR" })
export class Edr {
    @PrimaryGeneratedColumn()
    id?: number;
    @Column("varchar")
    name?: string;

    @ManyToMany(type => Datasource, datasource => datasource.edrs)
    @JoinTable({
        name: 'EdrDatasourceMaps',
        joinColumn: { name: 'EdrId' },
        inverseJoinColumn: { name: 'DatasourceId' }
     })
    datasources?: Datasource[];
}
```

test code
```ts
const data: Edr = {
  id: 3,
   name: "test",
    datasources: [
        {id: 4}, {id: 5}
    ]
};

await edr_repository.save(data); // will fail
// generated sql 
// INSERT INTO "EdrDatasourceMaps"("EdrId", "DatasourceId") VALUES (?, ?), (?, ?) -- PARAMETERS: [4,3,5,3]
// it should be 
// INSERT INTO "EdrDatasourceMaps"("DatasourceId", "EdrId") VALUES (?, ?), (?, ?) -- PARAMETERS: [4,3,5,3]

const data2: Datasource = {
  name: "datasource",
  edrs: [
      { id: 1 }, {id: 2}
  ]
}
await ds_repository.save(data); // will success
```

### Pull-Request Checklist

### DIFF
--- a/src/metadata/RelationMetadata.ts
+++ b/src/metadata/RelationMetadata.ts
@@ -531,7 +531,7 @@ export class RelationMetadata {
         // we do not want to register inverse relationship
         // to the inverse entity because it is going to override
         // it's relationship information
-        if (this.inverseRelation && this.isManyToManyOwner == false){
+        if (this.inverseRelation && this.isManyToManyOwner === false){
             this.inverseRelation.junctionEntityMetadata = junctionEntityMetadata;
             this.joinTableName = junctionEntityMetadata.tableName;
         }
--- a/src/metadata/RelationMetadata.ts
+++ b/src/metadata/RelationMetadata.ts
@@ -527,7 +527,11 @@ export class RelationMetadata {
     registerJunctionEntityMetadata(junctionEntityMetadata: EntityMetadata) {
         this.junctionEntityMetadata = junctionEntityMetadata;
         this.joinTableName = junctionEntityMetadata.tableName;
-        if (this.inverseRelation) {
+        // incase of many to many relationship
+        // we do not want to register inverse relationship
+        // to the inverse entity because it is going to override
+        // it's relationship information
+        if (this.inverseRelation && this.isManyToManyOwner == false){
             this.inverseRelation.junctionEntityMetadata = junctionEntityMetadata;
             this.joinTableName = junctionEntityMetadata.tableName;
         }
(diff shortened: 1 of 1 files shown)


pleerock:
I don't understand motivation of this PR.

Why did you put `@JoinTable` from both sides of relationship? You should have it only from one side, and which side has a join table considered as a "owning side". Owning side is only reflected on a generated table names.

Having `@JoinTable` on both sides is prohibited, if there is no validation in TypeORM, we can add it.

mluxe:
> I don't understand motivation of this PR.
> 
> Why did you put `@JoinTable` from both sides of relationship? You should have it only from one side, and which side has a join table considered as a "owning side". Owning side is only reflected on a generated table names.
> 
> Having `@JoinTable` on both sides is prohibited, if there is no validation in TypeORM, we can add it.

that sounds fare. I re-read the many-to-many document again and I notice that I simply did not notice that message. I think it needs to be hi-lighted in the document so it can be more visible.

the code in the PR will be allow the relationship being registered properly if someone like me added `@JoinTable` on both sides without notice the __must not__ do. If it does not make any sense to you then drop the PR.

alternatively, like you said a validation to through an error can also help us to know there must only have one `@JointTable` on many-to-many relationships.

pleerock:
Let me close this PR, if you'll have time to add validation logic, please feel free to create a new PR. Thanks

PR #7375, opened 2021-02-14
fix: printSql and getSql no longer break InsertQueryBuilder with Postgres

### LINKED ISSUE #3016: "could not determine data type of parameter $1" in createQueryBuilder insert
**Issue type:**

[ ] question
[x ] bug report
[ ] feature request
[ ] documentation issue

**Database system/driver:**

[ ] `cordova`
[ ] `mongodb`
[ ] `mssql`
[ ] `mysql` / `mariadb`
[ ] `oracle`
[x ] `postgres`
[ ] `sqlite`
[ ] `sqljs`
[ ] `react-native`
[ ] `expo`

**TypeORM version:**

[ ] `latest`
[ ] `@next`
[x ] `0.2.8` (or put your version here)

I am trying to create an insertQuery with sql function in Nestjs framework like this :
```
const result = await this.userRepository
      .createQueryBuilder()
      .insert()
      .into(User)
      .values({
        column_1: () => `nextval('table_id_seq')`,
        column_2: user.column_2,
        column_3: user.column_3,
        column_4: user.column_4,
        created_at: user.created_at,
      })
      .printSql()
      .execute();
```
The error : 

```
error: { error: could not determine data type of parameter $1
    at Connection.parseE (C:\Users\trung\code\ims\exchange-nest\node_modules\pg\lib\connection.js:554:11)
    at Connection.parseMessage (C:\Users\trung\code\ims\exchange-nest\node_modules\pg\lib\connection.js:379:19)
    at Socket.<anonymous> (C:\Users\trung\code\ims\exchange-nest\node_modules\pg\lib\connection.js:119:22)
    at emitOne (events.js:116:13)
    at Socket.emit (events.js:211:7)
    at addChunk (_stream_readable.js:263:12)
    at readableAddChunk (_stream_readable.js:250:11)
    at Socket.Readable.push (_stream_readable.js:208:10)
    at TCP.onread (net.js:601:20)
  name: 'error',
  length: 111,
  severity: 'ERROR',
  code: '42P18',
  detail: undefined,
  hint: undefined,
  position: undefined,
  internalPosition: undefined,
  internalQuery: undefined,
  where: undefined,
  schema: undefined,
  table: undefined,
  column: undefined,
  dataType: undefined,
  constraint: undefined,
  file: 'postgres.c',
  line: '1350',
  routine: 'exec_parse_message' }
```
Is insertQuery not support param like **column_1: () => `nextval('table_id_seq')`,**  like updateQuery in [#465](https://github.com/typeorm/typeorm/issues/465)

simon-a-j:
This corrects parameter numbering on `InsertQueryBuilder` to start from zero each time a query and parameters are requested. This resolves an issue whereby calling `printSql` or `getSql` followed by `execute` would generate incorrect parameter numbering, causing an error `could not determine data type of parameter $1` with Postgres.

### Description of change

Currently, if using a query builder to insert with values, and trying to print the query prior to execution, such as:

```
const result = await repo
                    .createQueryBuilder()
                    .insert()
                    .into(User)
                    .values({
                        first_name: `Test`,
                        last_name: `User`,
                    })
                    .printSql()
                    .execute();
```
                   
The insert operation will fail with an error on Postgres similar to:

```
could not determine data type of parameter $1
```

The reason is that `printSql` and `getSql` populate `QueryBuilder.expressionMap.nativeParameters`, which subsequently affects the behaviour of `execute` due to this line in `InsertQueryBuilder.createValuesExpression`:

```
let parametersCount = Object.keys(this.expressionMap.nativeParameters).length;
```

Here, parameter numbering is started from the number of native parameters already present in the instance. Since `printSql` or `getSql` populate `nativeParameters`, the behaviour differs depending on whether they have been called prior to execution.

With Postgres, since we use numbered placeholders (`$1`), the misnumbering causes an error.

This PR resolves the issue by always starting parameter numbering from zero when generating a query and parameters. I also choose to rename `parametersCount` to `paramNum` to indicate that it increments per parameter rather than representing the total count.

I'm having a hard time seeing why we'd ever want parameter numbering in a builder to not start from zero, but the existing behaviour seems like it could have been intentional so I'm prepared to learn that I missed something :laughing:

Fixes #3016

### Pull-Request Checklist

### DIFF
--- a/src/query-builder/InsertQueryBuilder.ts
+++ b/src/query-builder/InsertQueryBuilder.ts
@@ -427,7 +427,7 @@ export class InsertQueryBuilder<Entity> extends QueryBuilder<Entity> {
         // if column metadatas are given then apply all necessary operations with values
         if (columns.length > 0) {
             let expression = "";
-            let parametersCount = Object.keys(this.expressionMap.nativeParameters).length;
+            let paramNum = 0;
             valueSets.forEach((valueSet, valueSetIndex) => {
                 columns.forEach((column, columnIndex) => {
                     if (columnIndex === 0) {
@@ -473,9 +473,9 @@ export class InsertQueryBuilder<Entity> extends QueryBuilder<Entity> {
                     //     expression += subQuery;
 
                     } else if (column.isDiscriminator) {
-                        this.expressionMap.nativeParameters["discriminator_value_" + parametersCount] = this.expressionMap.mainAlias!.metadata.discriminatorValue;
-                        expression += this.connection.driver.createParameter("discriminator_value_" + parametersCount, parametersCount);
-                        parametersCount++;
+                        this.expressionMap.nativeParameters["discriminator_value_" + paramNum] = this.expressionMap.mainAlias!.metadata.discriminatorValue;
+                        expression += this.connection.driver.createParameter("discriminator_value_" + paramNum, paramNum);
+                        paramNum++;
                         // return "1";
 
                     // for create and update dates we insert current date
@@ -490,8 +490,8 @@ export class InsertQueryBuilder<Entity> extends QueryBuilder<Entity> {
                         const paramName = "uuid_" + column.databaseName + valueSetIndex;
                         value = RandomGenerator.uuid4();
                         this.expressionMap.nativeParameters[paramName] = value;
-                        expression += this.connection.driver.createParameter(paramName, parametersCount);
-                        parametersCount++;
+                        expression += this.connection.driver.createParameter(paramName, paramNum);
+                        paramNum++;
 
                     // if value for this column was not provided then insert default value
                     } else if (value === undefined) {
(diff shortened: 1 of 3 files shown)


src/query-builder/InsertQueryBuilder.ts
```diff
@@ -427,7 +427,7 @@ export class InsertQueryBuilder<Entity> extends QueryBuilder<Entity> {
         // if column metadatas are given then apply all necessary operations with values
         if (columns.length > 0) {
             let expression = "";
-            let parametersCount = Object.keys(this.expressionMap.nativeParameters).length;
```
pleerock, line 430: I don't remember why this line of code was needed, but I don't think it was incidentally placed here. There is definitely a mess around parameters in QueryBuilder, and this area of code needs to be refactored. I'm not ready to break any exist functionality in order to fix `printSql` method. We'll need to figure out why this line of code was needed before we can make a decision about this change.

simon-a-j: Thanks for the feedback! Let me take another look at this and see if I can either give us more confidence or find an alternative implementation.

nebkat: This is here for cases when other parts of the query builder have already added parameters to `expressionMap.nativeParameters`. In certain drivers it may be possible that some are already added, even with an `INSERT` query so we need to maintain the correct index/order. `this.connection.driver.createParameter()` creates the `$X` string based on whatever number is given to it.

I've mostly resolved the ordering problem in #7282, but it still remains that `nativeParameters` is shared between multiple `printSql()`/`getSql()` calls, leading to the issue described. We need to move it to a temporary storage that isn't `expressionMap`, as the parameters are calculated every time the query is compiled.

nebkat (changes requested):
Would avoid merging for now, this may break certain drivers and I think the same problem exists in `UpdateQueryBuilder` as well.

AlexMesser:
Looks like conversation got staled, closing for now.

PR #7391, opened 2021-02-18
fix: improve EntityManager.save() return type

arosequist:
This brings it in line with the [equivalent method](https://github.com/typeorm/typeorm/blob/fde9f0772eef69836ff4d85816cfe4fd6f7028b4/src/repository/Repository.ts#L122-L144) in Repository, which was originally changed with #2796.

### Description of change

* As long as `{ reload: false }` isn't specific in the `options`, the response should be `T & Entity`, not just `T`.
* The docstrings for the non-array versions should be singular, not plural.

### Pull-Request Checklist

  - I can't find a relevant issue for this.
  - Since this is just a typings change, I don't think there are any additional unit tests to make for it?

### DIFF
--- a/src/entity-manager/EntityManager.ts
+++ b/src/entity-manager/EntityManager.ts
@@ -291,13 +291,25 @@ export class EntityManager {
      * Saves all given entities in the database.
      * If entities do not exist in the database then inserts, otherwise updates.
      */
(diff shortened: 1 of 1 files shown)


pleerock:
Thank you for contribution! 🎉

Issue #7386, opened 2021-02-17
UpdateResult and DeleteResult return affected as undefined for Aurora Data API driver


coyoteecd:
## Issue Description

Aurora Data API driver started as a clone of MysqlDriver and for DeleteQueryBuilder, it copies the number of deleted rows from result.affectedRows. But Aurora Data API driver returns this in a numberOfRecordsUpdated property, therefore affected is left undefined.

For UpdateQueryBuilder, affected is not handled at all (the driver implementation was copied before #5628 landed in master). The same fix needs to be applied there as well.

### Expected Behavior

When using `repository.insert(entity)`, the `InsertResult.affected` value should reflect what's being returned by the low-level API via the numberOfRecordsUpdated property.
Likewise, when using `repository.update(entity)`, the `UpdateResult.affected` should do the same as insert.

### Actual Behavior

In both cases, affected is `undefined`.

### My Environment

| Dependency          | Version  |
| ---                 | ---      |
| Operating System    |  Windows 10        |
| Node.js version     | v12.14.0 | 
| Typescript version  | v4.1.3 | 
| TypeORM version     | v0.2.31 | 

### Relevant Database Driver(s)

- [x] `aurora-data-api`
- [ ] `aurora-data-api-pg`
- [ ] `better-sqlite3`
- [ ] `cockroachdb`
- [ ] `cordova`
- [ ] `expo`
- [ ] `mongodb`
- [ ] `mysql`
- [ ] `nativescript`
- [ ] `oracle`
- [ ] `postgres`
- [ ] `react-native`
- [ ] `sap`
- [ ] `sqlite`
- [ ] `sqlite-abstract`
- [ ] `sqljs`
- [ ] `sqlserver`

### Are you willing to resolve this issue by submitting a Pull Request?

- [x] Yes, I have the time, and I know how to start.
- [ ] Yes, I have the time, but I don't know how to start. I would need guidance.
- [ ] No, I don't have the time, although I believe I could do it if I had the time...
- [ ] No, I don't have the time and I wouldn't even know how to start.

PR #7433, opened 2021-02-26
fix: [aurora-data-api] Return number of affected rows in UpdatedResult and DeleteResult

coyoteecd:
Closes: #7386

Add correct parsing of raw result for Aurora Data API (mysql-based) driver so that number of records updated are copied to the "generic" affected property.

### Pull-Request Checklist

- [ ] `npm run lint` passes with this change (there are already failures on my machines)

### DIFF
--- a/src/query-builder/DeleteQueryBuilder.ts
+++ b/src/query-builder/DeleteQueryBuilder.ts
@@ -73,9 +73,13 @@ export class DeleteQueryBuilder<Entity> extends QueryBuilder<Entity> implements 
             const result = await queryRunner.query(sql, parameters);
 
             const driver = queryRunner.connection.driver;
-            if (driver instanceof MysqlDriver || driver instanceof AuroraDataApiDriver) {
+            if (driver instanceof MysqlDriver) {
                 deleteResult.raw = result;
                 deleteResult.affected = result.affectedRows;
+
+            } else if (driver instanceof AuroraDataApiDriver) {
+                deleteResult.raw = result;
+                deleteResult.affected = result.numberOfRecordsUpdated;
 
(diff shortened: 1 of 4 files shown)


src/query-builder/DeleteQueryBuilder.ts
```diff
@@ -73,10 +73,14 @@ export class DeleteQueryBuilder<Entity> extends QueryBuilder<Entity> implements
...
-            if (driver instanceof MysqlDriver || driver instanceof AuroraDataApiDriver) {
+            if (driver instanceof MysqlDriver) {
                 deleteResult.raw = result;
                 deleteResult.affected = result.affectedRows;
 
+            } else if (driver instanceof AuroraDataApiDriver) {
+                deleteResult.raw = result;
+                deleteResult.affected = result.numberOfRecordsUpdated;
+
             } else if (driver instanceof SqlServerDriver || driver instanceof PostgresDriver || driver instanceof CockroachDriver) {
```
Olyno, line 84: You seems to fix the ``AuroraDataApiDriver`` thing, but you didn't change anything about this list of drivers. Is it normal? I don't know about the ``AuroraDataApiDriver``, but I know that this bug is available for ``PostgresDriver`` too and probably more drivers. Do the ``AuroraDataApiDriver`` fixes all?

coyoteecd: It obviously fixes only AuroraDataApiDriver (see the tag in the commit message). Would be nice to fix them all, unfortunately I do not have enough knowledge of the other drivers to submit a fix for everything.

This is a follow-up PR for the one that fixed MySQL driver, as explained in the issue.

AlexMesser:
thank you for contribution!

PR #7452, opened 2021-03-08
fix: Switch on logging for migration generate command 

zaro:
### Description of change
Revert, show , run have schema logging on by default, and for some reason generate has all logging disabled. 

This PR, simply enables it.

### Pull-Request Checklist

### DIFF
--- a/src/commands/MigrationGenerateCommand.ts
+++ b/src/commands/MigrationGenerateCommand.ts
@@ -97,7 +97,7 @@ export class MigrationGenerateCommand implements yargs.CommandModule {
                 synchronize: false,
                 migrationsRun: false,
                 dropSchema: false,
-                logging: false
+                logging: ["error", "schema"]
             });
 
             const upSqls: string[] = [], downSqls: string[] = [];


pleerock:
TBH I don't see much sense to output `schema` logs. It maybe even confusing, because we are generating a new migration instead of outputing like like `schema:log` command does.

Regarding to `error` logs - also don't see a point, because we already have `try/catch` block which will show the error if something fails.

zaro:
@pleerock The value I see form this logs is that it basically outputs simple human readable list of changes that are going to be in the migration, e.g. :
```
dropping an old view: public.service
dropping an old view: public.vehicle_view
....
renaming column "name" in to "fullName"
...
creating a new view: public.service
creating a new view: public.vehicle_view
```

Which gives a good overview of what is in the migration w/o combing trough tons of SQL statements.

Our models have grown quite large, and often migrations are quite lengthy, which leads to the more junior developers simply not bother to read the SQL at all. Having this kind of short list of changes is quite helpful IMO, but I agree that it probably it's not needed for everybody.

What do you think if it's made somehow optional via command line switch or env var ?

pleerock:
we have a `typeorm schema:log` command to see what SQL code is going to be generated. This command was purposely created for the use case you just mentioned - before migration generation just run `schema:log` to make sure everything is alright. IMO adding flag to migration generate command is a duplication of `schema:log`. Plus using `schema:log` prevents you from wrong migration generation in the case if output SQLs aren't correct.

zaro:
That's my point thought, it's not duplication of schema log, it textual representation of the changes. `schema:log` also shows SQL, which is anyway in the migration.

Also using it for preview, is not really useful because it outputs SQL again, which is fine but again if you just want a glance of what is changed SQL is too much if you have 500+ lines of SQL in the migration ( which is not uncommon in our project due the heavy use of views) :)

Anyway, it is not super important, I just saw it as an easy improvement but it's also not a big deal if not merged.

PR #7459, opened 2021-03-10
Added socketPath support for replicas

wieringen:
Currently, Typeorm doesn't support socket connections for replicas. We use Google CloudSQL (mysql) and the preferred way to connect to CloudSQL from Cloud Run is sockets. The implementation is pretty straight forward and works the same as the none replica configuration; when a socketPath is set the host and port are ignored.


AlexMesser:
thank you for contribution!

PR #7497, opened 2021-03-24
Fixed: edited compilerOptions.lib, added "DOM" propertie

Rupert-com:
**Fixed:** Added `DOM` to the `compilerOptions.lib`, which is needed in order to make console.log which is generated in index.ts

### Pull-Request Checklist

- [✅ ] Code is up-to-date with the `master` branch
- [N/A ] `npm run lint` passes with this change
- [N/A ] `npm run test` passes with this change
- [N/A ] This pull request links relevant issues as `Fixes #0000`
- [N/A ] There are new or updated unit tests validating the change
- [✅ ] Documentation has been updated to reflect this change
- [✅] The new commits follow conventions explained in [CONTRIBUTING.md](https://github.com/typeorm/typeorm/blob/master/CONTRIBUTING.md)

### DIFF
--- a/src/commands/InitCommand.ts
+++ b/src/commands/InitCommand.ts
@@ -217,7 +217,7 @@ export class InitCommand implements yargs.CommandModule {
     protected static getTsConfigTemplate(): string {
         return JSON.stringify({
             compilerOptions: {
-                lib: ["es5", "es6"],
+                lib: ["es5", "es6", "DOM"],
                 target: "es5",
                 module: "commonjs",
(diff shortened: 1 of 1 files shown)


AlexMesser:
I don't think this is a correct change. Mostly `dom` should be used if you are writing a browser-based application. If you are writing a node-base application you should install `@types/node` instead of using `dom` (until you don't need a dom-specific or browser-specific features). It means we can't accept this change - user needs to install what he needs depend on the platform he is using.

PR #7529, opened 2021-04-02
fix: define class properties for QueryFailedError

ejose19:
Defining those properties allow users to access a typed error like:

```ts
if (error instanceof QueryFailedError) {
  // This will be typed
  console.log(error.query, error.parameters);
}
```


AlexMesser:
thank you for contribution!

PR #7538, opened 2021-04-05
fix: find by Date object in sqlite driver

### LINKED ISSUE #2286: find operators like MoreThan and LessThan doesn't work properly for date fields
**Issue type:**

[ ] question
[x] bug report
[ ] feature request
[ ] documentation issue

**Database system/driver:**

[ ] `cordova`
[ ] `mongodb`
[ ] `mssql`
[ ] `mysql` / `mariadb`
[ ] `oracle`
[ ] `postgres`
[x] `sqlite`
[ ] `sqljs`
[ ] `react-native`

**TypeORM version:**

[x] `latest`
[ ] `@next`

find operators like MoreThan and LessThan doesn't work properly for date fields. some investigation showed that sqlite stores dates like ```2018-06-05 15:57:27.249``` but the query is like ```... WHERE "User"."createdAt" > ? -- PARAMETERS: ["2018-06-05T01:01:32.650Z"]``` meaning it queries in ISO date format. when you just need comparison till day factor, it doesnt look like a problem but when you also need the time/hours/mins/etc then it doesnt return proper results. i only tested for sqlite so i dont know if same problem goes for other dbs.

### LINKED ISSUE #3426: @PrimaryColumn on Date field SQLite save issue
**Issue type:**

[ ] question
[x ] bug report
[ ] feature request
[ ] documentation issue

**Database system/driver:**

[ ] `cordova`
[ ] `mongodb`
[ ] `mssql`
[ ] `mysql` / `mariadb`
[ ] `oracle`
[ ] `postgres`
[ x] `sqlite`
[ ] `sqljs`
[ ] `react-native`
[ ] `expo`

**TypeORM version:**

[x ] `latest`
[ ] `@next`

I create an Entity with a primary column using Date as type, i can save using the save method but the next time the save method cant find the already saved entities with the same date because two problems:

* The search query of the save method uses ISO dates instead of the SQLite format` "YYYY-MM-DD"`, i resolve this creating a custom ValueTransformer and momentjs (but this must be a default behavior)

* The search query don't use the `date` function when searching for primary date column

The output of the save method in a second attempt looks like:

```bash
CONSOLE INFO file:///app/vendor.js:234819:21: query:  SELECT "UserPodcastHistory"."time" AS "UserPodcastHistory_time", "UserPodcastHistory"."userId" AS "UserPodcastHistory_userId" FROM "user_podcast_history" "UserPodcastHistory" WHERE (("UserPodcastHistory"."time" = ?) OR ("UserPodcastHistory"."time" = ?)) -- PARAMETERS: ["2018-10-09","2018-10-08"]
CONSOLE INFO file:///app/vendor.js:234819:21: query:  BEGIN TRANSACTION
CONSOLE INFO file:///app/vendor.js:234819:21: query:  INSERT INTO "user_podcast_history"("time", "userId") VALUES (?, NULL) -- PARAMETERS: ["2018-10-09 00:00:00.000"]
CONSOLE ERROR file:///app/vendor.js:234822:22: query failed:  INSERT INTO "user_podcast_history"("time", "userId") VALUES (?, NULL) -- PARAMETERS: ["2018-10-09 00:00:00.000"]
CONSOLE ERROR file:///app/vendor.js:234822:22: error:  SQLITE.ExecSQL Failed 0
CONSOLE INFO file:///app/vendor.js:234819:21: query:  ROLLBACK
CONSOLE ERROR file:///app/vendor.js:47035:24: ERROR QueryFailedError: SQLITE.ExecSQL Failed 0
```

phil294:
### Description of change (copied from commit message)

In sqlite, Date objects are persisted as UtcDatetimeString.
But a Date object parameter was escaped with .toISOString(), making such queries impossible.
This commit aligns both transforms.
This bug does *not* apply to better-sql where you can only bind numbers, strings, bigints, buffers,
and null.
This is breaking for when the user inserted their dates manually as ISO and relied on this old
maltransformation, after this their find()s by Date won't work anymore.

BREAKING CHANGE: Change Date serialization in selects

fixes #2286
fixes #3426 

### Pull-Request Checklist

### DIFF
--- a/test/github-issues/2286/issue-2286.ts
+++ b/test/github-issues/2286/issue-2286.ts
@@ -3,20 +3,21 @@ import { createTestingConnections, closeTestingConnections, reloadTestingDatabas
 import { Connection } from "../../../src/connection/Connection";
 import { expect } from "chai";
 import { Post } from "./entity/Post";
+import { Example } from "./entity/Example";
 import { Between } from "../../../src";
 
 describe("github issues > #2286 find operators like MoreThan and LessThan doesn't work properly for date fields", () => {
 
     let connections: Connection[];
     before(async () => connections = await createTestingConnections({
-        entities: [__dirname + "/entity/*{.js,.ts}"],
+        entities: [ Post, Example ],
         schemaCreate: true,
         dropSchema: true,
         /* Test not eligible for better-sql where binding Dates is impossible */
         enabledDrivers: ["sqlite"]
     }));
     beforeEach(() => reloadTestingDatabases(connections));
-    
+
     after(() => closeTestingConnections(connections));
 
     it("should find a record by its datetime value with find options", () => Promise.all(connections.map(async connection => {
@@ -51,5 +52,29 @@ describe("github issues > #2286 find operators like MoreThan and LessThan doesn'
             .where("post.dateTimeColumn = :now", { now })
             .getOne();
         expect(postByDateEquals).to.not.be.undefined;
+    })));
+
+    it("should save, update, and load with a date PK", () => Promise.all(connections.map(async connection => {
+        const start = new Date("2000-01-01");
+        const middle = new Date("2000-06-30");
+        const end = new Date("2001-01-01");
+
+        await connection.manager.save(Example, { id: start, text: "start" });
+        await connection.manager.save(Example, { id: middle, text: "middle" });
+        await connection.manager.save(Example, { id: end, text: "end" });
+
+        const repo = connection.manager.getRepository(Example);
+
+        let example = await repo.findOneOrFail({ id: middle });
+
+        expect(example.text).to.be.equal("middle");
+
+        example.text = "in between";
+
+        await repo.save(example);
+
+        example = await repo.findOneOrFail({ id: middle });
+
+        expect(example.text).to.be.equal("in between");
     })));
 });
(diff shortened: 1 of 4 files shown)


osmar85:
This will not working when you need find exat Date. SqLite Date format is YYYY-MM-DD HH-mm-ss and this solution brings YYYY-MM-DD HH-mm-ss.sss

test/github-issues/2286/issue-2286.ts
```diff
@@ -0,0 +1,55 @@
...
+        const middle = new Date("2000-06-30");
+        const post = new Post();
+        post.dateTimeColumn = middle;
+
+        await connection.manager.save(post);
+
+        const postByDateEquals = await connection.manager.findOne(Post, {
+            dateTimeColumn: middle
+        });
+        expect(postByDateEquals).to.not.be.undefined;
```
phil294, line 34: @osmar85 Here is a working test for finding by exact date. Further below with query builder

phil294:
And in my DB, values are indeed stored like `2021-05-09 22:00:00.000`, so *with* ms. The column has no specific type annotated, but is a Javascript `Date`. Are actually seeing date values being saved without MS @osmar85 ? If so, what does the property definition and annotations look like in your case?

imnotjames:
Given the breaking change is fixing broken selects this is probably acceptable..

imnotjames:
rebased & add test that shows this fixes #3426 as well

pleerock (approved):

pleerock:
Since it's a breaking change, I'll merge it into 0.3.0 branch. Expect to see more breaking changes in 0.4.0 regarding to how Dates work - because there are many different issues currently with the current dates handling design.

PR #7540, opened 2021-04-06
Add example for yarn v2 with pnp

kevin-lindsay-1:
### Pull-Request Checklist

### DIFF
--- a/docs/migrations.md
+++ b/docs/migrations.md
@@ -151,6 +151,11 @@ Example with `ts-node`:
 ts-node ./node_modules/typeorm/cli.js migration:run
 ```
 
+Example with `ts-node` using `yarn v2` with `pnp`:
+```
+ts-node $(yarn bin typeorm) migration:run
+```
+
 This command will execute all pending migrations and run them in a sequence ordered by their timestamps.
 This means all sql queries written in the `up` methods of your created migrations will be executed.
 That's all! Now you have your database schema up-to-date.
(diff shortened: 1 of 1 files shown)


docs/migrations.md
```diff
@@ -151,6 +151,11 @@ Example with `ts-node`:
 ts-node ./node_modules/typeorm/cli.js migration:run
 ```
 
+Example with `ts-node` using `yarn v2` with `pnp`:
```
merceyz, line 154: This isn't Yarn v2 nor PnP specific, it should be done like this no matter what version of Yarn and/or if you're using `node_modules`. If you're using workspaces `node_modules/typeorm/cli.js` might not exist as it could have been hoisted or a future version of `typeorm` change the bin path.

kevin-lindsay-1: Updated; please resolve if you're good.

merceyz: Looks good 👍; I don't have access to resolve

AlexMesser:
@kevin-lindsay-1 your solution looks weird... I'm not sure how what magic in your suggestion resolves that issue. 

> But the better solution would be to list pg (and anything else typeorm needs to require) as a peerDependency.

can you please check if @brandon-leapyear suggested solution works, I think it will be a better approach to fix this issue.

kevin-lindsay-1:
@AlexMesser yarn v2's pnp uses a virtual directory system. This method is required because when using pnp the `node_modules` directory (and by extension the binary) is not easily available locally. `yarn bin` is the command that exposes the pnp location of the binary.

AlexMesser:
thank you for contribution!

PR #7559, opened 2021-04-09
fix:use host if hostReplicaSet is not provided

NthMetal:
Change was for MongoDriver's buildConnectionUrl method for building a connection url for a replica set.
Currently the host that is used is the config option called "hostReplicaSet"
The change updates this so that if this option is not provided it uses a provided "host" with the given port.
It also defaults to using "127.0.0.1" as a host if none is provided, and "27017" if no port was provided.

This change was needed because I did not provide a connection url to my typeorm config, but I provided all other neccessary info to build a connection url. However, I provided only a "host" and not "hostReplicaSet" which caused it to create a url with the host "undefined". I tested this on an environment with a replica set.

### Pull-Request Checklist

### DIFF
--- a/src/driver/mongodb/MongoDriver.ts
+++ b/src/driver/mongodb/MongoDriver.ts
@@ -444,13 +444,13 @@ export class MongoDriver implements Driver {
             : "";
 
         let connectionString = undefined;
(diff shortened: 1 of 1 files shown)


AlexMesser:
thank you for contribution!

PR #7564, opened 2021-04-12
fix: Make QueryRunner do not generate duplicate query when enum is same name

### LINKED ISSUE #7523: Do not create duplicate CREATE TYPE migration query when same 'enumName's are exists
### The Problem
When there's a parent entity class with enum field that has multiple children,  typeorm migration creates duplicate `CREATE TYPE` query in migration file. This causes `Type already defined` error when running migration.

For example,
```typescript
// foobar-entity.ts
enum FooEnum { FOO, BAR }

class ParentEntity {
  @Column({
    type: 'enum',
    enum: FooEnum,
    enumName: 'foo_enum',
  })
  foo: FooEnum;
}

@Entity()
class ChildEntity1 extends ParentEntity {}

@Entity()
class ChildEntity2 extends ParentEntity {}
```

Then generate migration,
```typescript
// TIMESTAMP-FoobarMigration.ts
export class FoobarMigrationTIMESTAMP implements MigrationInterface {
  name = 'FoobarMigrationTIMESTAMP';

  public async up(queryRunner: QueryRunner) {
    await queryRunner.query('CREATE TYPE "foo_enum" AS ENUM ~~');
    // (create table child entity 1)
    await queryRunner.query('CREATE TYPE "foo_enum" AS ENUM ~~');
    // (create table child entity 2)
  }
}
```

### The Solution
Storing `enumName` in set or map to prevent from generating duplicate query might solve the issue.

### Considered Alternatives

For now, as of my knowledge, removing duplicate queries manually is the only way to solve the issue.

- [ ] `aurora-data-api`
- [ ] `aurora-data-api-pg`
- [ ] `better-sqlite3`
- [ ] `cockroachdb`
- [ ] `cordova`
- [ ] `expo`
- [ ] `mongodb`
- [ ] `mysql`
- [ ] `nativescript`
- [ ] `oracle`
- [x] `postgres`
- [ ] `react-native`
- [ ] `sap`
- [ ] `sqlite`
- [ ] `sqlite-abstract`
- [ ] `sqljs`
- [ ] `sqlserver`

5d-jh:
### Description of change

This issue fixes #7523 by storing built enum names in set class property.

### Pull-Request Checklist

### DIFF
--- a/src/driver/postgres/PostgresQueryRunner.ts
+++ b/src/driver/postgres/PostgresQueryRunner.ts
@@ -53,6 +53,12 @@ export class PostgresQueryRunner extends BaseQueryRunner implements QueryRunner 
     protected releaseCallback: Function;
 
     // -------------------------------------------------------------------------
+    // Private Properties
+    // -------------------------------------------------------------------------
+
+    private enumNames: Set<string>;
+
+    // -------------------------------------------------------------------------
     // Constructor
     // -------------------------------------------------------------------------
 
@@ -62,6 +68,8 @@ export class PostgresQueryRunner extends BaseQueryRunner implements QueryRunner 
         this.connection = driver.connection;
         this.mode = mode;
         this.broadcaster = new Broadcaster(this);
+
+        this.enumNames = new Set();
     }
 
     // -------------------------------------------------------------------------
@@ -350,17 +358,21 @@ export class PostgresQueryRunner extends BaseQueryRunner implements QueryRunner 
 
         // if table have column with ENUM type, we must create this type in postgres.
         const enumColumns = table.columns.filter(column => column.type === "enum" || column.type === "simple-enum")
-        const createdEnumTypes: string[] = []
+        // const createdEnumTypes = new Set<string>();
+
         for (const column of enumColumns) {
             // TODO: Should also check if values of existing type matches expected ones
             const hasEnum = await this.hasEnumType(table, column);
             const enumName = this.buildEnumName(table, column)
 
             // if enum with the same "enumName" is defined more then once, me must prevent double creation
-            if (!hasEnum && createdEnumTypes.indexOf(enumName) === -1) {
-                createdEnumTypes.push(enumName)
-                upQueries.push(this.createEnumTypeSql(table, column, enumName));
-                downQueries.push(this.dropEnumTypeSql(table, column, enumName));
+            if (!hasEnum) {
+                if (!this.enumNames.has(enumName)) {
+                    upQueries.push(this.createEnumTypeSql(table, column, enumName));
+                    downQueries.push(this.dropEnumTypeSql(table, column, enumName));
+                }
+
+                this.enumNames.add(enumName);
             }
         }
 
--- a/test/github-issues/7523/issue-7523.ts
+++ b/test/github-issues/7523/issue-7523.ts
@@ -1,4 +1,5 @@
 import "reflect-metadata";
+import {expect} from "chai";
 import {Connection} from "../../../src";
 import {createTestingConnections, closeTestingConnections} from "../../utils/test-utils";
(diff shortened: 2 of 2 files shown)


MrBacony:
"Documentation has been updated to reflect this change N/A"

is there a point in the documentation which should be updated?

5d-jh:
@MrBacony As far as I know of there isn't any. That's why I marked as N/A.

MrBacony:
@5d-jh ok good to know.

Maybe we have to check in your code, if all duplicate Enums have the same Entries.

I hope this PR will be merged soon.

imnotjames (changes requested):
I don't think this is a complete solution.

Wouldn't this not handle the case where you aren't doing this all at once?

Like if you create a new table that shares the enum - but not at the same time as the first one.

Then it would fail again.

Is it possible to add a test for that case?

5d-jh:
@imnotjames

Come to think about it, this PR isn't a good solution to resolve the issue.

I think registering all of enums in connection options, then referencing them by name would be a better solution in terms of maintainability. Example below.

```typescript
import SomeEnum from './some-enum';

createConnection({
  enums: [{
    name: 'some_enum',
    enum: SomeEnum,
  }]
})
```

```typescript
@Entity()
class SomeEntity {
  @Column({
    type: 'enum',
    enumName: 'some_enum',
  })
  someEnumField: SomeEnum;
}
```

Please let me know if this might be a better solution.

5d-jh:
Closing this PR since there is better solution #8161

5d-jh:
* `same name and different values within one migration` -> abort migration and throw error
* `same name and same values within one migration && different values with already defined in database` -> perform alter type

Will this flow solve the issue? (If this is accepted, would specify this behavior in documentation)

AlexMesser:
Finally I was able to review this PR.

Right now we have `hasEnumType` call inside `createTable` which should not be there, because `createTable` should not have any database calls to work properly with both runtime sync and migration generation. According to proper schema sync design we should first load all necessary data from the db, then execute sync based on comparison of db schema and schema collected from entity metadatas.

So, we actually need to re-implement current enum sync approach before we can properly fix this issue.

I have a suggestion on how to change the current implementation:

1. Introduce a new decorator called `@CustomType` which can be used on entities:

```ts
@CustomType({ name: `foo_enum`, type: "enum", value: FooEnum })
class ParentEntity {
}
```

2. remove `enumName` from ColumnOptions and introduce custom types support:

```ts
@Column(`foo_enum`)
foo: FooEnum;
```

3. if user don't want to create a custom type we can still specify "inline" enum but without name (and if we have two usages of the same enum, ORM automatically create us two enum types with different generated names even through with the same values - it's expected behaviour until we introduce a `@CustomType`

4. inside `RdbmsSchemaBuilder` we sync custom types separately (just like how we do it for indices, fks, etc.)

```ts
        await this.dropOldViews();
        await this.dropOldForeignKeys();
        await this.dropOldIndices();
        await this.dropOldChecks();
        await this.dropOldExclusions();
        await this.dropCompositeUniqueConstraints();
        await this.renameColumns();
        await this.createNewTables();
        await this.dropRemovedColumns();
        await this.dropCustomTypes() // NEW
        await this.createCustomTypes() // NEW
        await this.addNewColumns();
        await this.updatePrimaryKeys();
        await this.updateExistColumns();
        await this.createNewIndices();
        await this.createNewChecks();
        await this.createNewExclusions();
        await this.createCompositeUniqueConstraints();
        await this.createForeignKeys();
        await this.createViews();
```

Unfortunately re-implementation will take lot of time, and we have to postpone it to 0.4.0

AlexMesser:
Unfortunately I have to close this PR because we should implement it slightly different... Closing in favor of #8672

PR #7593, opened 2021-04-25
feat: allow passing custom user-data through local entity manager

artysidorenko:
Only QueryRunner currently accepts custom user-data.

This commit allows data to be added to local instances of EntityManager (and QueryBuilder).

Makes it possible to pass request tracing data to the query logger.

### Description of change

**Current behaviour**: `data` property is only exposed via the `QueryRunner`, or passed automatically when using persist/remove methods. There is, to my knowledge, no way to attach the data to "transient" runners created and released by typeorm when executing regular queries.

**New behaviour**: you can pass user data to local instances of `EntityManager`, which in turn passes it to any "transient" query runners - examples in the tests.

**Rationale**: The `data` property is very useful for attaching tracing information at the request level. For example if you get some sort of `connection terminated` error, you can trace the log back to the original http request. However, you either need to use 1 query runner per request (makes no difference for transactional requests, but adds a bottleneck for the rest: once you've run out of connections in the pool, new requests have to wait for the last one to complete), or rewrite all the queries manually to get the runner and attach the data each time. We have some very long running `GET` requests that make both these alternatives difficult.

Let me know if there's any feedback or concerns.

### Pull-Request Checklist

- ~[ ] This pull request links relevant issues as `Fixes #0000`~ N/A

### DIFF
--- a/test/other-issues/custom-user-data/custom-user-data.ts
+++ b/test/other-issues/custom-user-data/custom-user-data.ts
@@ -1,0 +1,176 @@
+import sinon from "sinon";
+import {Connection} from "../../../src/connection/Connection";
+import {createTestingConnections, closeTestingConnections, reloadTestingDatabases } from "../../utils/test-utils";
+import { SimpleConsoleLogger, EntityManager } from "../../../src";
+import { Foo } from "./entity/Foo";
+
+describe("other issues > capture custom user data in logger (e.g. trace id)", () => {
+    let connections: Connection[];
+    let manager: EntityManager;
+    let logQueryStub: sinon.SinonStub;
+    const testFoo: Foo = {
+        id: 1,
+        name: "foo",
+    };
+
+    before(async() => {
+        logQueryStub = sinon.stub(SimpleConsoleLogger.prototype, "logQuery");
+        connections = await createTestingConnections({
+            enabledDrivers: ["postgres"],
+            entities: [Foo],
+            logging: true,
+            createLogger: () => new SimpleConsoleLogger(),
+        });
+    });
+    beforeEach(async () => {
+        await reloadTestingDatabases(connections);
+        await Promise.all(connections.map(connection => connection.getRepository(Foo).insert(testFoo)));
+    });
+    afterEach(async () => {
+        await Promise.all(connections.map(connection => connection.getRepository(Foo).delete({})));
+        logQueryStub.resetHistory();
+    });
+    after(() => {
+        closeTestingConnections(connections);
+        logQueryStub.restore();
+    });
+
+    describe("when attaching user data to request-scoped entity manager", () => {
+
+        const tests = [{
+            testTarget: "entity manager query",
+            testFunc: async() => await manager.query("SELECT * FROM foo"),
+            loggerArgs: [
+                ["SELECT * FROM foo", undefined]
+            ],
+        },{
+            testTarget: "entity manager transaction",
+            testFunc: async() => await manager.transaction(trx => trx.query("SELECT * FROM foo")),
+            loggerArgs: [
+                ["START TRANSACTION", undefined],
+                ["SELECT * FROM foo", undefined],
+                ["COMMIT", undefined]
+            ],
+        }, {
+            testTarget: "repository save query",
+            testFunc: async() => await manager.getRepository(Foo).save({id: 2, name: "secondFoo"}),
+            loggerArgs: [
+                ["START TRANSACTION", undefined],
+                [`INSERT INTO "foo"("name", "delete_date") VALUES ($1, DEFAULT) RETURNING "id", "delete_date"`, ["secondFoo"]],
+                ["COMMIT", undefined]
+            ],
+        }, {
+            testTarget: "repository remove query",
+            testFunc: async() => await manager.getRepository(Foo).remove({id: 2, name: "secondFoo"}),
+            loggerArgs: [
+                ["START TRANSACTION", undefined],
+                [`DELETE FROM "foo" WHERE "id" = $1`, [2]],
+                ["COMMIT", undefined]
+            ],
+        }, {
+            testTarget: "repository soft remove query",
+            testFunc: async() => await manager.getRepository(Foo).softRemove({id: 1, name: "foo"}),
+            loggerArgs: [
+                ["START TRANSACTION", undefined],
+                [`UPDATE "foo" SET "delete_date" = CURRENT_TIMESTAMP WHERE "id" IN ($1)`, [1]],
+                ["COMMIT", undefined]
+            ],
+        }, {
+            testTarget: "repository recover query",
+            testFunc: async() => await manager.getRepository(Foo).recover({id: 1, name: "foo"}),
+            loggerArgs: [
+                ["START TRANSACTION", undefined],
+                [`UPDATE "foo" SET "delete_date" = NULL WHERE "id" IN ($1)`, [1]],
+                ["COMMIT", undefined]
+            ],
+        }, {
+            testTarget: "repository insert query",
+            testFunc: async() => await manager.getRepository(Foo).insert({id: 2, name: "secondFoo"}),
+            loggerArgs: [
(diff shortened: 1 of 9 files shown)


test/other-issues/custom-user-data/custom-user-data.ts
```diff
@@ -0,0 +1,176 @@
...
+    afterEach(async () => {
+        await Promise.all(connections.map(connection => connection.getRepository(Foo).delete({})));
+        logQueryStub.resetHistory();
+    });
+    after(() => {
+        closeTestingConnections(connections);
+        logQueryStub.restore();
+    });
+
+    describe("when attaching user data to request-scoped entity manager", () => {
```
artysidorenko, line 38: sorry - quite a lot of test cases here. Different methods on the repository were creating queryRunners in different ways, so I erred on the side of caution and wrote tests for most of them to avoid any surprises.

AlexMesser:
I don't think this is a good and obvious property for the `EntityManager`. Since every entity manager has a queryRunner, why not to use it as `manager.queryRunner.data = x` ?

artysidorenko:
From what I understand, the EntityManager would need to be created with a single instance of QueryRunner for the lifetime of the request, to keep the same data (or we need to create a new EM on each query and keep re-attaching the data)? When the EM is created without a query runner, it will create a new one and release it each time on the fly ([here](https://github.com/artysidorenko/typeorm/blob/master/src/connection/Connection.ts#L390) was the example where I saw this first, in `Connection.ts`)

Going down the "single query runner" route works well for transactional requests as these are already doing it anyway. However it would introduce a lot of additional overhead on "non-transactional" queries, as these would now also need to hold onto a single db pool client throughout the duration of the request before releasing it. (we have a lot of long running read-only queries that might create bottlenecks in the connection pool)

An alternative approach I thought about initially was for an optional setting to allow the query runner to release its connection client after each query + take out a new one on the next query... that way we could use the single QR instance without the performance bottlenecks; but that felt like a much bigger change to how the query runner works, so I didn't want to mess with that.

Let me know if there's any alternatives you might be open to.. for now I had kept the property as already existed on the queryRunner, but don't have strong feelings about the structure.

pleerock:
I'm sorry @artysidorenko but I didn't get how your answer answers @AlexMesser question...

artysidorenko:
sorry @pleerock I dove a bit too much into some specific details.

Summary is: by default the entity manager does not have a single defined query runner, right? `manager.queryRunner` is normally `undefined`. A new runner - with blank `data` - is created and released each time a query is executed.

What I wrote in my last reply was trying to explain why using a transactional entity manager (which does have a single defined query runner) is not always feasible.

pleerock:
> by default the entity manager does not have a single defined query runner, right?

Right, because having query runner means having a connection used/opened. Having it in the global entity manager means we have it always opened and we assume we'll use it always in every query. I think it will slow down things (at the end if we need performance and we have replication setup, only master will be used all the time and no read replicas ever be used?). 

Can you please explain why exist solution `manager.queryRunner.data = x` won't work for you?

artysidorenko:
Here's a shortened example of what I'm trying to do; when passing a unique request id to the query runner. I don't think it's possible unless I always make the request handler transactional?

```typescript
class Controller {
  @Get('/foo')
  async requestHandler(
    req: Request,
  ) {
    const { requestId } = req.headers;
    const manager = getConnection().createEntityManager();    // (in practice this is done via middleware)

    // manager.queryRunner.data = { requestId };   <-- would fail as query runner does not exist yet

    const result = manager.query('SELECT * FROM foo;');

    // ...not shown: more queries and calcs before returning...

    return result;
  }
}
```

pleerock:
Okay now it's more clear what you are trying to do.

As I remember it doesn't make sense to call `const manager = getConnection().createEntityManager();` without `queryRunner` passed in. Same as using default `manager` from connection.

What makes sense however is to create and pass a queryRunner:

```ts
const queryRunner = getConnection().createQueryRunner();
const manager = getConnection().createEntityManager(queryRunner);
```

Don't forget to release `queryRunner`. Also here you can set data to `queryRunner`, it will be equivalent of what you are trying to achieve I guess.

artysidorenko:
ok thanks. I was hoping to avoid using 1 single queryRunner throughout the request and releasing it at the end (it impacts db performance because the request holds onto the pool connection and does not let any other request use it until the current request is finished), but it sounds like my idea in this PR might be muddying the entity manager class too much. I'll have a think how best to manage the db performance separately.

pleerock:
> I was hoping to avoid using 1 single queryRunner throughout the request and releasing it at the end (it impacts db performance because the request holds onto the pool connection and does not let any other request use it until the current request is finished)

that's right, I would also avoid doing that.

> but it sounds like my idea in this PR might be muddying the entity manager class too much.

to be honest it does. I don't say we can't have it, but having `data` in both entity manager and query runner doesn't seem to be good. But I completely understand what you want to achieve and had similar issues.

> const manager = getConnection().createEntityManager();    // (in practice this is done via middleware)

Also one question: if you have this manager defined in the `middleware` - how do you "access" this manager? through the `request`? Just curious and trying to understand other people practices.

artysidorenko:
> Also one question: if you have this manager defined in the middleware - how do you "access" this manager? through the request? Just curious and trying to understand other people practices.

In a nutshell it's through the request yes. We use a bunch of nestjs tools; an interceptor to create a custom `ExecutionContext` at the start of the request, with some request/additional logging metadata + the entity manager; and a custom decorator to pass it to the controller. But it works through the request indeed.

```typescript
class ContextInterceptor implements NestInterceptor {
  intercept(context: ExecutionContext, next: CallHandler): Observable<any> {
    const request = context.switchToHttp().getRequest();
    const manager = getConnection().createEntityManager();
    ctx = {
      logContext,
      ...
      getManager: () => Promise.resolve(manager);    // not shown: if endpoint is transactional - also starts a trx
    };
    request.ctx = ctx;
    .... return next() ...
  }
}

const Context = createParamDecorator((_, request?: any) => {
  return request.ctx;
});

class Controller {
  @Get('/read-only-endpoint')
  async getFoo(
    @Context()
    ctx: ExecutionContext,
  ): Promise<Foo[]> {
    const manager = await ctx.getManager();
    const foo = await manager.find(Foo);
    return foo;
  }
}
```
There's still a bit of prop drilling for all the downstream services (so we try to keep `ExecutionContext` as the first parameter consistently), but it really helps with observability to propagate the logging context throughout the application so we can pinpoint the cause when something goes wrong.

artysidorenko:
Hi again @pleerock 

Was thinking about ways to make this proposal more robust.. I thought about adding a private/protected option property on the manager instead, so it's not as vague/loosely defined. This could then trickle down into the query runner.

Just checking if this seems like a more robust model, in which case I could update this PR to see how it looks in practice? 

New property on `EntityManager.ts`:

```typescript
/**
  * Creation options for query runners initialised by this entity manager
  */
protected queryRunnerOptions: QueryRunnerOptions;
```

The new interface `QueryRunnerOptions.ts`:

```typescript
/**
 * Additional options when creating a new query runner
 */
export interface QueryRunnerOptions {
  /**
   * Attach data to the query runner
   */
  readonly data?: ObjectLiteral
}
```

Update the method on `Connection.ts` (and other places where required):

```typescript
createQueryRunner(mode: ReplicationMode = "master", options?: QueryRunnerOptions): QueryRunner {
    const queryRunner = this.driver.createQueryRunner(mode);
    const manager = this.createEntityManager(queryRunner);
    Object.assign(queryRunner, { manager: manager });

    /****  new section         ****/
    if (options?.data) {
        queryRunner.data = options.data;
    }
    /****  end of new section  ****/

    return queryRunner;
}
```

(I also thought about making it a more open-ended extension point e.g. allow the option to be a function that takes the QR and returns a modified instance, so people can configure/extend the query runner as they please, unless that's too open..)

pleerock:
@artysidorenko did you figure out solutions for your problems? I would like to close this PR since I don't think we need to increase complexity by adding extra rare case features.

artysidorenko:
Hi @pleerock, no problem, will close this PR.

In the end we largely went with node async-local-storage to attach data to the request context.

PR #7596, opened 2021-04-27
Fix #6958

### LINKED ISSUE #6958: Promises never get resolved in specific cases
**Issue type:**

[x] bug report
[ ] feature request
[ ] documentation issue

**Database system/driver:**

[ ] `cordova`
[x] `mongodb`
[ ] `mssql`
[ ] `mysql` / `mariadb`
[ ] `oracle`
[ ] `postgres`
[ ] `cockroachdb`
[ ] `sqlite`
[ ] `sqljs`
[ ] `react-native`
[ ] `expo`

**TypeORM version:**

[ ] `latest`
[x] `@next`

Sometimes promises just never resolve. At first I thought it might be related with running scripts with `ts-node`, however even when using pure JS the issue is the same.

The freeze always happens when I try to perform a query after another query was performed without closing the connection. Here closing and re-connecting seems to avoid the bug, however I had another case where I tried saving an entity to a repository after performing a raw query and it also froze on the `await repository.save(...)` method, re-connecting did not fix it.

Here's a code snippet in pure JS:

```.js
const createConnection = require("typeorm").createConnection;


(async () => {
    const con = await createConnection();
    await con.createQueryRunner().query(`SELECT 1 as foo`);
    // re-connecting would work here...
    // await con.close();
    // await con.connect();
    await con.createQueryRunner().query(`SELECT 1 as foo`);
    console.log(1);
    await con.close(); // freezes here
    console.log(2); // this is never reached
})();
```

The actual contents of the raw query do not seem to matter, it happens every time. The problem is probably related to raw queries since the problem didn't happen with repositories (until doing a raw query once, after that it does).

PauloRSF:
### Description of change

When releasing a used query runner on postgres' driver, `splice()` was used without limiting the number of elements to be deleted, causing all the query runners to be removed, thus, preventing the driver from releasing them properly. So, the callback on https://github.com/typeorm/typeorm/blob/f85f436f51fb000cd9959b44e8d7a79bf0cd10ab/src/driver/postgres/PostgresDriver.ts#L1056 was never called on node-postgres and the promise never resolved. 

I created a test based on the issue's demo, which works now.

Fixes #6958

### Pull-Request Checklist

### DIFF
--- a/src/driver/postgres/PostgresDriver.ts
+++ b/src/driver/postgres/PostgresDriver.ts
@@ -1051,7 +1051,10 @@ export class PostgresDriver implements Driver {
      * Closes connection pool.
      */
     protected async closePool(pool: any): Promise<void> {
-        await Promise.all(this.connectedQueryRunners.map(queryRunner => queryRunner.release()));
+        while (this.connectedQueryRunners.length) {
+            await this.connectedQueryRunners[0].release();
+        }
+
         return new Promise<void>((ok, fail) => {
             pool.end((err: any) => err ? fail(err) : ok());
         });
--- a/src/driver/postgres/PostgresQueryRunner.ts
+++ b/src/driver/postgres/PostgresQueryRunner.ts
@@ -127,7 +127,7 @@ export class PostgresQueryRunner extends BaseQueryRunner implements QueryRunner 
             this.releaseCallback();
 
         const index = this.driver.connectedQueryRunners.indexOf(this);
-        if (index !== -1) this.driver.connectedQueryRunners.splice(index);
+        if (index !== -1) this.driver.connectedQueryRunners.splice(index, 1);
 
         return Promise.resolve();
     }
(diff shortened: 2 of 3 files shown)


src/driver/postgres/PostgresDriver.ts
```diff
@@ -1051,7 +1051,10 @@ export class PostgresDriver implements Driver {
      * Closes connection pool.
      */
     protected async closePool(pool: any): Promise<void> {
-        await Promise.all(this.connectedQueryRunners.map(queryRunner => queryRunner.release()));
+        while (this.connectedQueryRunners.length) {
```
AlexMesser, line 1054: can you please explain why this change to `while` is needed? As I understand main motivation is to release them in sequence. If so, why its needed? Thanks!

PauloRSF: Because the `.release()` method of the query runners mutates the driver's `connectedQueryRunners` list, causing `.map()` to not iterate through all of them. I changed it to a while loop because it was the most obvious way i could think to ensure that all query runners got released and `connectedQueryRunners` was really empty before closing the pool 😅. If you have a better way to do it, let me know!

AlexMesser:
thank you for contribution!

Issue #5983, opened 2020-05-01
cache stored without identifier/key when using eager:true


kateile:
**Issue type:**
[x] bug report

**Database system/driver:**
[x] `mssql`

**TypeORM version:**
[x] `0.x.x` (0.2.24)

**Steps to reproduce or a small repository showing the problem:**
I am trying to cache users by their username, and everything works fine when `eager:false`. When `true` cache is stored without **key**. This bug could provide incorrect result when cache is enabled and `eager:true` is used. I have managed to reproduce it has below.

User.ts
```ts
@Entity()
export class User {
  @PrimaryGeneratedColumn()
  id: number;

  @Column({nullable: true, unique: true})
  username?: string;

  @OneToOne(
    () => Profile,
    (profile) => profile.user,
    {eager: true}
  )
  @JoinColumn()
  profile: Profile;
}
```
Profile.ts
```ts
@Entity()
export class Profile {
  @Field(() => ID)
  @PrimaryGeneratedColumn()
  id: number;

  @OneToOne(
    () => User,
    (user) => user.profile
  )
  user: User;
}
```
UserRepository .ts
```ts
@EntityRepository(User)
export class UserRepository extends Repository<User> {
  async findByUsername(username: string): Promise<User> {
    const user = await this.findOne({
      where: {username},
      cache: {
        id: `username:${username}`,
        milliseconds: 24 * 60 * 60 * 1000
      }
    });

    if (!user) {
      throw new Error(`user with ${username} was not found!!!`);
    }

    return user;
  }

}
```
UserRepository.test.ts
```ts
describe('User Tests', () => {
  it('should save cache with identifier when eager:true', async () => {
    const userRepository = getCustomRepository(UserRepository);

    const username = 'mwanza';

    try {
      await userRepository.findByUsername(username);
      expect(1).toEqual(5); //In case above passed.
    } catch (e) {
      expect(e.message).toMatch(new RegExp(username));
    }

    const redisKey = await redis.get(`username:${username}`); //Passing cache key as used above.
    expect(redisKey).not.toBeNull(); //Received: null
  });
});
```

kateile:
Also the same thing happen when using `relation`. Is this the intended behavior?

kateile:
@imnotjames The driver is **mysql** not mssql. I made a little error when reporting.

aamirpinger:
any progress on this?

4l3ss:
[...]

Issue #4277, opened 2019-06-12
Using cache in findAndCount and getManyAndCount returns 0 as count


Tallyrald:
**Issue type:**

[ ] question
[x] bug report
[ ] feature request
[ ] documentation issue

**Database system/driver:**

[ ] `cordova`
[ ] `mongodb`
[ ] `mssql`
[x] `mysql` / `mariadb`
[ ] `oracle`
[ ] `postgres`
[ ] `cockroachdb`
[ ] `sqlite`
[ ] `sqljs`
[ ] `react-native`
[ ] `expo`

**TypeORM version:**

[ ] `latest`
[x] `@next`
[x] `0.2.18`

**Steps to reproduce or a small repository showing the problem:**

Consider the following code (the function is inside an abstract class):
```ts
static async GetBasicList(userRepo: Repository<User>, limit: number, page: number): Promise<[User[], number]> {
	return userRepo.findAndCount({
		cache: {
			id: 'userBasicList',
			milliseconds: 100000,
		},
		order: {
			id: 'DESC',
		},
		relations: ['profile'],
		skip: page * limit,
		take: limit,
	});
}
```
Without the cache, the result is fully correct (returns the correct amount of users in the right order, loads the profile relation, correctly paginates and gives back the number of total hits). However when the above is run with caching (as it is specified above) the full result count returns 0.
For example if I have 10 users in my database and I run the query with limit set to 5 and page set to 0, I get 5 users and a count of 0 instead of 10.

The User entity is fairly complex with 11 relations and some other fields, however nothing is eagerly loaded. Some relations have cascade set to true (like the profile the query uses), but removing cascades does not help.

Caching is enabled in ormconfig.js and this issue is present with database cache and ioredis as well.

Also interesting to observe that in the cache, there are 2 queries inserted. One with the 'userBasicList' identifier and one with a NULL identifier. The latter contains a query that starts like this:
```sql
SELECT DISTINCT `distinctAlias`.`User_id` as "ids_User_id", `distinctAlias`.`User_id` FROM (SELECT...
```

Tried using @next, but the same thing happened.

EDIT:
Forgot to mention but the exact same pattern applies when a queryBuilder.getManyAndCount() is used.

Tallyrald:
It seems the problem originates from somewhere in SelectQueryBuilder's getManyAndCount() method.
Looking at the logs whenever the cache option is omitted (that is cache is disabled on the query) 3 queries are run.
[...]

PR #7610, opened 2021-05-02
Fix: remove unnecessary caching of SELECT DISTINCT query while using repository functions

### LINKED ISSUE #2721: Caching queries
**Issue type:**

[ ] question
[* ] bug report
[ ] feature request
[* ] documentation issue

**Database system/driver:**

[ ] `cordova`
[ ] `mongodb`
[ ] `mssql`
[* ] `mysql` / `mariadb`
[ ] `oracle`
[ ] `postgres`
[ ] `sqlite`
[ ] `sqljs`
[ ] `react-native`
[ ] `expo`

**TypeORM version:**

[ ] `latest`
[ ] `@next`
[* ] `0.2.7` (or put your version here)

In my case: **Redis cache**
https://github.com/typeorm/typeorm/blob/master/docs/caching.md : 
"... clearing cached results when you insert a new user" - similar (by remove) it doesn't

**repository<entity>**
1. find -> [1,2,3]
2. remove 2
3. find -> [1,2,3]; should be: [1,3]

Conclusion:
remove method doesn't clear cache key where was some entity

**Edited**
I forgot about config:
`
    cache: {
        type: 'redis',
        alwaysEnabled: true,
        duration: parseInt(AppConfig.DB.CACHE_REDIS_DURATION, 10),
        options: {}
    },`
and this line for "alwaysEnabled": https://github.com/typeorm/typeorm/blob/master/src/connection/BaseConnectionOptions.ts#L125

vasu2652:
Fix: remove unnecessary caching of preselect query while using the repository functions.

### Description of change

When using repository functions like findOne or find or findAndCount with cache enabled the default behavior of TypeOrm is trying to cache the intended query with the provided identifier and also a preselect query with a SELECT DISTINCT to resolve the PrimaryColumn.

Due to this data invalidation is going a bit out of control and we are getting to see stale data for a period of time after mutation on the entity.

Current Behaviour: Return stale data even after invalidation using the provided identifier.

Expected Behaviour: Return updated data after invalidation using provided the identifier.

Closes: #2721, #5983, #4277

### Pull-Request Checklist

- [x] This pull request links relevant issues as `Fixes #2721 #5983 #4277`


test/github-issues/2721/repository-caching.ts
```diff
@@ -0,0 +1,368 @@
...
+
+        // insert new entity
+        const user4 = new User();
+        user4.firstName = "Bakhrom";
+        user4.lastName = "Brochik";
+        user4.isAdmin = true;
+        await connection.getRepository(User).save(user4);
+
+        // without cache it must return really how many there entities are
+        const users2 = await await connection.getRepository(User).find({
```
shrujalshah28, line 69: Remove one `await`

vasu2652: oops. thank you!! 😊

imnotjames:
Won't this effectively make the caching useless?  The most expensive query is often the one to find the right PKs and order.  Looking those rows up again by PK is by comparison much faster.

What is the cache caching?

vasu2652:
> Won't this effectively make the caching useless? The most expensive query is often the one to find the right PKs and order. Looking those rows up again by PK is by comparison much faster.
> 
> What is cache caching?

@imnotjames  Yes, I agree with you. But...
Why would we need this explicit query caching when the user already has granular control over when to cache and when to invalidate. 
Due to this on the data change or mutation, there is no way we can invalidate this cached query, which ends up showing the state data.

What additional advantage this query is giving when we are already caching at a query level? and also this happens without the consent of the user until someone actually looks into the cache server.

imnotjames:
>  Why would we need this explicit query caching when the user already has granular control over when to cache and when to invalidate.

Except they don't have any sort of control over the cache because you're removing it in this PR.

> What additional advantage this query is giving when we are already caching at a query level? and also this happens without the consent of the user until someone actually looks into the cache server.

If you are caching the other part of the query you must cache this part.  That's how TypeORM works.

TypeORM does a first pass here to pull the primary keys then does a second pass to pull the data by primary key - which is usually incredibly fast.

However: If you don't want caching do not enable caching for the query.  If you do any sort of caching you will have stale data because the caching implementation built into TypeORM is simple at best.

PR #7648, opened 2021-05-12
Add CountOptions for Repository.count() not to be able to set distinct for counting

### LINKED ISSUE #7161: Add support for non-distinct count in getCount()
### The Problem

I noticed that when calling `Repository.count(findOptions)` the default SQL query generated by Typeorm is a `SELECT(COUNT(DISTINCT ...))` query [when using Postgres](https://github.com/typeorm/typeorm/blob/920e7812cd9d405df921f9ae9ce52ba0a9743bea/src/query-builder/SelectQueryBuilder.ts#L1811-L1817).

Postgres has some issues with counts that are distinct. There are [many](https://www.citusdata.com/blog/2016/10/12/count-performance/) [articles](https://www.peterbe.com/plog/best-way-to-count-distinct-indexed-things-in-postgresql) [and](https://www.postgresql.org/message-id/CAONnt+72Mtg6kyAFDTHXFWyPPY-QRbAtuREak+64Lm1KN1c-wg@mail.gmail.com) [discussions](https://stackoverflow.com/questions/11250253/postgresql-countdistinct-very-slow) on that subject. Basically, the performance can be very negatively impacted when that happens. The default would still be a distinct count, evidently.

Recently, I made a change in the codebase of the company to whom I work to use the QueryBuilder and build a non-distinct count, because there was no need for it to be `DISTINCT`.  The performance got significantly better.

If the developer knows that it is safe to make a non-distinct count, it would be nice if TypeORM provided a way to do that more seamlessly than by having the developer build a query with QueryBuilder.

### The Solution

There is some coupling between the options for count and for find, which makes a simple solution hard for me to define. One solution I can think of is creating an interface to be used as the type of the argument of a count instead of `FindOptions<T>`, something like:

```declare interface CountOptions<T> extends FindOptions<T> {
  distinct?: boolean
}
```

And then use the value of `distinct` to build the count expression in `computeCountExpression()`. I am pretty new to Typescript, however, so I am sure there are better solutions.

### Considered Alternatives

Keep having the `COUNT(DISTINCT(...))` as the only option for the counts.

- [ ] `aurora-data-api`
- [ ] `aurora-data-api-pg`
- [ ] `better-sqlite3`
- [ ] `cockroachdb`
- [ ] `cordova`
- [ ] `expo`
- [ ] `mongodb`
- [ ] `mysql`
- [ ] `nativescript`
- [ ] `oracle`
- [x] `postgres`
- [ ] `react-native`
- [ ] `sap`
- [ ] `sqlite`
- [ ] `sqlite-abstract`
- [ ] `sqljs`
- [ ] `sqlserver`

t2y:
### Description of change

Fixes #7161

The problem and solution are described in #7161, and I encountered the same issue in my application. I'm not the author of #7161, but I try to fix it.

There are 2 cases that we get counts with typeorm.

* Repository.count()
* SelectQueryBuilder.getCount()

When we call them with a table with relations, it queries `SELECT(COUNT(DISTINCT ...))`. This PR adds CountOptions for Repository.count() to control whether SELECT takes DISTINCT or not.

```typescript
export interface CountOptions<Entity = any> extends FindManyOptions<Entity> {
    /**
     * Specifies whether the selection is DISTINCT.
     */
    distinct?: boolean;
}
```

For example, we can pass CountOptions to Repository.count().

```typescript
orderRepository.count({
    distinct: false,
    relations: ["user", "products"],
    where: { user: userA },
});
```

Then, I added a new `disable-count-distinct` option to SelectQueryBuilderOption and control whether SELECT takes DISTINCT or not internally.

```typescript
export type SelectQueryBuilderOption = "disable-global-order"|"create-pojo"|"disable-count-distinct";
```

By doing so, when we use QueryBuilder, we can set the option directly. I think that's quite explicit and easy to handle.

```typescript
orderRepository
    .createQueryBuilder("order")
    .leftJoin("order.user", "user")
    .leftJoin("order.products", "product")
    .where("user.id = :id", { id: userA.id })
    .setOption("disable-count-distinct")
    .getCount();
```

For easy to use, `getCount()` also can take an option.

```typescript
orderRepository
    .createQueryBuilder("order")
    .leftJoin("order.user", "user")
    .leftJoin("order.products", "product")
    .where("user.id = :id", { id: userA.id })
    .getCount({distinct: false});
```

#### about QueryExpressionMap

QueryExpressionMap already has `selectDistinct` for distinct, but this attribute is for find() or raw query, not count(). The value is `false` by default, and it doesn't affect `SELECT(COUNT(DISTINCT ...))`. To keep the compatibility, we shouldn't reuse this attribute, I think.

```typescript
export class QueryExpressionMap {
    ...
    /**
     * Whether SELECT is DISTINCT.
     */
    selectDistinct: boolean = false;
    ...
}
```

I'm new to typeorm, so there might be a better design keeping the compatibility to do this. Tell me if you know.

### Count SQL changes

Count SQL will be changed below. The `issue-7161.ts` test code generates these SQLs.

#### Current

```sql
SELECT 
  COUNT(
    DISTINCT("c_order"."id", "c_order"."key")
  ) AS "cnt" 
FROM 
  "c_order" "c_order" 
  LEFT JOIN "user" "user" ON "user"."id" = "c_order"."userId" 
  LEFT JOIN "c_order_products_product" "c_order_product" ON "c_order_product"."cOrderId" = "c_order"."id" 
  AND "c_order_product"."cOrderKey" = "c_order"."key" 
  LEFT JOIN "product" "product" ON "product"."id" = "c_order_product"."productId" 
WHERE 
  "user"."id" = $1
```

#### Specifying `disable-count-distinct` option

```sql
SELECT 
  COUNT(1) AS "cnt" 
FROM 
  "c_order" "c_order" 
  LEFT JOIN "user" "user" ON "user"."id" = "c_order"."userId" 
  LEFT JOIN "c_order_products_product" "c_order_product" ON "c_order_product"."cOrderId" = "c_order"."id" 
  AND "c_order_product"."cOrderKey" = "c_order"."key" 
  LEFT JOIN "product" "product" ON "product"."id" = "c_order_product"."productId" 
WHERE 
  "user"."id" = $1
```

### Pull-Request Checklist


imnotjames:
If you aren't doing a distinct count, the concat does nothing except reduce performance, right?

t2y:
@imnotjames Thank you for reviewing.

> If you aren't doing a distinct count, the concat does nothing except reduce performance, right?

I think so, too. Some DBMS need `COUNT(CONCAT(...))` since they don't take multiple columns for `COUNT()`.

> Also, why not an optional distinct option for getCount directly?

I will implement the option if it's needed. Firstly, I thought that I should implement minimum changes with a consistent way for `Repository.count()` and `getCount()`. For example, I'm not sure that API can take wrong usage like `.setOption("disable-count-distinct").getCount({distinct: true})` is good or not.

imnotjames:
> @imnotjames Thank you for reviewing.
> 
> > If you aren't doing a distinct count, the concat does nothing except reduce performance, right?
> 
> I think so, too. Some DBMS need `COUNT(CONCAT(...))` since they don't take multiple columns for `COUNT()`.

To clarify: Without distinct, what's the difference between this query with the `CONCAT` and `COUNT(1)`?  Is there any?

> 
> > Also, why not an optional distinct option for getCount directly?
> 
> I will implement the option if it's needed. Firstly, I thought that I should implement minimum changes with a consistent way for `Repository.count()` and `getCount()`. For example, I'm not sure that API can take wrong usage like `.setOption("disable-count-distinct").getCount({distinct: true})` is good or not.

Try it, I think it will simplify the implementation.  An extra optional parameter is easier to grok than extra options you have to memorize, I'd think..

t2y:
> To clarify: Without distinct, what's the difference between this query with the CONCAT and COUNT(1)? Is there any? Can we drop the CONCAT in that case?

OK. I will investigate the difference.

> Try it, I think it will simplify the implementation.

Got it. I will add it.

t2y:
@imnotjames 

> To clarify: Without distinct, what's the difference between this query with the CONCAT and COUNT(1)? Is there any? Can we drop the CONCAT in that case?

You are right. We don't need CONCAT when a primary key would be passed to COUNT parameter. We can replace simple COUNT(1) in this case. The only difference is passing a column name including a null value.

I refactored my changes and updated the description of this issue.

pleerock:
Sorry for late review.

My question is - in what cases `COUNT(1)` doesn't work and we need to apply DISTINCT?

pleerock:
Closing as there is no answer from author.

Issue #3563, opened 2019-02-02
Retrieving ID in @AfterInsert operation


alex5207:
**Issue type:**

[X ] question
[ ] bug report
[ ] feature request
[ ] documentation issue

**Database system/driver:**

[ ] `cordova`
[ ] `mongodb`
[ ] `mssql`
[X ] `mysql` / `mariadb`
[ ] `oracle`
[ ] `postgres`
[ ] `sqlite`
[ ] `sqljs`
[ ] `react-native`
[ ] `expo`

**TypeORM version:**

[X] `latest`
[ ] `@next`
[ ] `0.x.x` (or put your version here)

**Steps to reproduce or a small repository showing the problem:**

I have two entities, **Person** and **Photo** which has a one-to-one relationship. After creating a person I'm trying to create a Photo in the `@AfterInsert` part. However, when trying to fetch the id of the newly created person, it returns undefined (

```@Entity()
export class Person {

  @AfterInsert()

  public async handleAfterInsert() {
    const person = await getConnection()
      .getRepository(Person)                     //<--- This is where the problem is
      .findOne(this.id);

    //Insert into Photo
      const photo = New Photo();
      photo.person = person;

    await getConnection()
      .getRepository(Photo)
      .save(photo);
  }

  @PrimaryGeneratedColumn()
  public id: number;

  @OneToOne(type => Photo, photo => photo.person)
  public photo: Photo
}
`

**Comment by vlapo** [comment]
Did you use `save()` or `insert()` method? Because `insert()` method is just simple INSERT in DB and save method do lot of magic around - like fill primary generated column.

**Comment by alex5207** [comment]
When creating a `Person` I am using the `save()` method.
Note: I can verify that the ID is present if I do a `console.log(this.id)`.

So, its like the repository is not updated quickly enough

**Comment by vlapo** [comment]
Understand now. I thought `this.id` is undefined before. This is problem of transactions. Insert statement is done but transaction is not committed yet so person is not available in new select.

But I do not understand why you need select person again? You have your entity in `this` so use it.
```typescript
@AfterInsert()
public async handleAfterInsert() {
    //Insert into Photo
    const photo = New Photo();
    photo.person = this;

    await getConnection()
      .getRepository(Photo)
      .save(photo);
}
```

vlapo:
Yeah you need use same `entityManager` instance because of transactions. In short you have to use subscriber instead of listener.

```typescript
@EventSubscriber()
[...]

PR #7664, opened 2021-05-18
Commit insert transaction before broadcasting after insert event

### LINKED ISSUE #6450: TypeORM Subscriber AfterInsert Can't Find Record
**Issue type:**

[x] question
[ ] bug report
[ ] feature request
[ ] documentation issue

**Database system/driver:**

[ ] `cordova`
[ ] `mongodb`
[ ] `mssql`
[ ] `mysql` / `mariadb`
[ ] `oracle`
[x] `postgres`
[ ] `cockroachdb`
[ ] `sqlite`
[ ] `sqljs`
[ ] `react-native`
[ ] `expo`

**TypeORM version:**

[ ] `latest`
[ ] `@next`
[x] `0.2.25` (or put your version here)

I have a few Entities `Action`, and `GPKAsset` 

### Action.ts
```typescript
export class Action extends BaseEntity {

	@PrimaryGeneratedColumn()
	id: number;

REMOVED COLUMNS AND OTHER RELATIONSHIPS TO REDUCE CLUTTER

	@ManyToMany(() => GPKAsset, gpkAsset => gpkAsset.actions)
	gpkAssets: GPKAsset[];
```

### GPKAsset.ts
```typescript
@Entity("gpk_assets")
export class GPKAsset extends BaseEntity {

	@PrimaryGeneratedColumn()
	id: number;

	@Index()
	@Column()
	contract: string;

REMOVED COLUMNS AND OTHER RELATIONSHIPS TO REDUCE CLUTTER

	@ManyToMany(() => Action, action => action.gpkAssets)
	@JoinTable({name:'gpkassets_actions'})
	actions: Action[];
```

### ActionSubscriber.ts
```typescript

@EventSubscriber()
export class ActionSubscriber implements EntitySubscriberInterface<Action> {
	listenTo() {
		return Action;
	}

	async afterInsert(event: InsertEvent<Action>): Promise<Promise<any> | void> {

		let action: Action = event.entity;
		switch (event.entity.contract) {
			case 'simpleassets:createlog':
				let gpkAsset: GPKAsset | null = new GPKAsset();
				gpkAsset = await gpkassetManager.createGPKAssetFromAction(action);
				await actionManager.addRelationAccount(action);
				if (gpkAsset !== null) {
				     await actionManager.addRelationGPKAsset(action, gpkAsset))
				}
				break;
			case 'simpleassets:transfer':
				await gpkassetManager.transferGPKAssetOwnership(action);
				break;
			default:
		}
	}
}
```

In the `ActionSubscriber.ts` I would like to perform certain things based on the contract column as you can see above. 

The problem I am encountering is on `await actionManager.addRelationAccount(action);` it is throwing exceptions saying that the record doesn't exist in actions. This is quite confusing because I have the id of the action and this is the `afterInsert` event. 

I am guessing either I am doing something very wrong or it's some sort of race condition, or it's a bug? 

Any guidance would be greatly appreciated. 

Thank you

Roemerb:
### Description of change

Fixes #6450 

Commits an insert transaction before broadcasting the after insert event. Fixes issue #6450. We have been having a similar issue as described there. When the `afterInsert` event is triggered, we would like to emit a job to our queue worker that will index the entity data in elastic search. However, we have noticed that we cannot fetch the entity from the database at the time that the subscriber event is triggered. As the issue describes, this is because the `afterInsert` event is broadcasted before the entity is committed to the database. This PR flips that code around.

### Pull-Request Checklist

### DIFF
--- a/src/query-builder/InsertQueryBuilder.ts
+++ b/src/query-builder/InsertQueryBuilder.ts
@@ -117,6 +117,13 @@ export class InsertQueryBuilder<Entity> extends QueryBuilder<Entity> {
                 // console.timeEnd(".updating entity");
             }
 
+            // close transaction if we started it
+            // console.time(".commit");
+            if (transactionStartedByUs) {
+                await queryRunner.commitTransaction();
+            }
+            // console.timeEnd(".commit");
+
             // call after insertion methods in listeners and subscribers
             if (this.expressionMap.callListeners === true && this.expressionMap.mainAlias!.hasMetadata) {
                 const broadcastResult = new BroadcasterResult();
@@ -125,13 +132,6 @@ export class InsertQueryBuilder<Entity> extends QueryBuilder<Entity> {
                 });
                 if (broadcastResult.promises.length > 0) await Promise.all(broadcastResult.promises);
             }
-
-            // close transaction if we started it
-            // console.time(".commit");
-            if (transactionStartedByUs) {
-                await queryRunner.commitTransaction();
-            }
-            // console.timeEnd(".commit");
 
             return insertResult;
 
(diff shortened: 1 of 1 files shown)


pleerock:
This is by design.

Listeners must run inside transaction, because some people want to do something else in the same transaction before committing it. Or revert transaction if something in their listeners fail.

AlexMesser:
here is the solution for the original issue https://github.com/typeorm/typeorm/issues/3563#issuecomment-460054971

Issue #5981, opened 2020-05-01
Getting redundant migrations generated


moltar:
**Issue type:**

[ ] question
[x] bug report
[ ] feature request
[ ] documentation issue

**Database system/driver:**

[ ] `cordova`
[ ] `mongodb`
[ ] `mssql`
[ ] `mysql` / `mariadb`
[ ] `oracle`
[x] `postgres`
[ ] `cockroachdb`
[ ] `sqlite`
[ ] `sqljs`
[ ] `react-native`
[ ] `expo`

**TypeORM version:**

[x] `latest`
[ ] `@next`
[ ] `0.x.x` (or put your version here)

**Steps to reproduce or a small repository showing the problem:**

Running `migration:generate` against a database that is already in sync, produces the following redundant change:

```ts
import {MigrationInterface, QueryRunner} from "typeorm";

export class sync1588321965836 implements MigrationInterface {
    name = 'sync1588321965836'

    public async up(queryRunner: QueryRunner): Promise<void> {
        await queryRunner.query(`CREATE SEQUENCE "foo_bar_level_manufacturer_id_seq" OWNED BY "foo_bar_level_manufacturer"."id"`, undefined);
        await queryRunner.query(`ALTER TABLE "foo_bar_level_manufacturer" ALTER COLUMN "id" SET DEFAULT nextval('foo_bar_level_manufacturer_id_seq')`, undefined);
        await queryRunner.query(`ALTER TABLE "foo_bar_level_manufacturer" ALTER COLUMN "id" DROP DEFAULT`, undefined);
        await queryRunner.query(`CREATE SEQUENCE "foo_baz_manufacturer_id_seq" OWNED BY "foo_baz_manufacturer"."id"`, undefined);
        await queryRunner.query(`ALTER TABLE "foo_baz_manufacturer" ALTER COLUMN "id" SET DEFAULT nextval('foo_baz_manufacturer_id_seq')`, undefined);
        await queryRunner.query(`ALTER TABLE "foo_baz_manufacturer" ALTER COLUMN "id" DROP DEFAULT`, undefined);
    }

    public async down(queryRunner: QueryRunner): Promise<void> {
        await queryRunner.query(`ALTER TABLE "foo_baz_manufacturer" ALTER COLUMN "id" SET DEFAULT nextval('foo_baz_manufactur_id_seq'`, undefined);
[...]

Issue #3076, opened 2018-11-08
migration:generate continuously updates default values


iCrawl:
**Issue type:**

[x?] question
[x] bug report
[ ] feature request
[ ] documentation issue

**Database system/driver:**

[ ] `cordova`
[ ] `mongodb`
[ ] `mssql`
[ ] `mysql` / `mariadb`
[ ] `oracle`
[x] `postgres`
[ ] `sqlite`
[ ] `sqljs`
[ ] `react-native`
[ ] `expo`

**TypeORM version:**

[x] `latest` 0.2.8
[ ] `@next`
[ ] `0.x.x` (or put your version here)

**Steps to reproduce or a small repository showing the problem:**

Somewhat relevant other issue #2943 (?)
Maybe this is just me misunderstanding how postgresql works, but since (seen in the screenshots below) the default values are correctly set. If the default values are really just supposed to be `'{}'` and `ARRAY[]` please feel free to close this issue 🙇 

Playlists.ts
```ts
import { Entity, Column, PrimaryGeneratedColumn } from 'typeorm';

@Entity('playlists')
export class Playlist {
	@PrimaryGeneratedColumn()
	id!: number;

	@Column({ type: 'bigint' })
	user!: string;

	@Column({ type: 'bigint' })
	guild!: string;

	@Column({ type: 'text' })
	name!: string;

	@Column({ type: 'text' })
	description!: string | null;

	@Column({ type: 'text', array: true, default: () => "ARRAY[]::text[]" })
	songs!: string[];

	@Column({ default: 0 })
	plays!: number;
}
```

Settings.js
```ts
import { Entity, Column, PrimaryColumn } from 'typeorm';

@Entity('settings')
export class Setting {
	@PrimaryColumn({ type: 'bigint' })
	guild!: string;

	@Column({ type: 'jsonb', default: () => "'{}'::jsonb" })
	settings: any;
}
```

1541719937415-Init.ts
```ts
import {MigrationInterface, QueryRunner} from "typeorm";

export class Init1541719937415 implements MigrationInterface {

    public async up(queryRunner: QueryRunner): Promise<any> {
[...]

Issue #1729, opened 2018-03-09
Queries generated with function as DEFAULT still generated later on migrations


ffd114:
**Issue type:**

[ ] question
[X] bug report
[ ] feature request
 [ ] documentation issue

**Database system/driver:**

[ ] `cordova`
[ ] `mongodb`
[ ] `mssql`
[ ] `mysql` / `mariadb`
[ ] `oracle`
[X] `postgres`
[ ] `sqlite`
[ ] `sqljs`
[ ] `websql`

**TypeORM version:**

[X] `latest`
[ ] `@next`
[ ] `0.x.x` (or put your version here)

**Steps to reproduce or a small repository showing the problem:**

1. Create entity, e.g: `User` with column `email`
2. Generate and run migration `typeorm migrations:generate -n Init && typeorm migrations:run`
3. Add new column `createdAt` with `default` set to `() => 'NOW()'`
4. Generate and run migration `typeorm migrations:generate -n AlterCreated && typeorm migrations:run`
5. Add new column `name`
6. If we generate migration again `typeorm migrations:generate -n Test`, the queries that generated because the changes on step 3 will show up again

**Expected**

Changes on step 3 already applied on step 4, the queries should not generated again later

I've setup repository [here](https://github.com/ffd114/migration-with-function) to reproduce this issue

pleerock:
Can you please check this issue in latest `@next` ? `npm i typeorm@next`

ffd114:
I've tried using `typeorm@next` and it's working

Many thanks!

ffd114:
Sorry, I'm reopening this, it seems it still not working, I will investigate this further.

federicofrade:
I have the same issue as @adrianhara by now my solution it's not use default value but charge my values always.
It is possible something like this
```ts
dafault: timestamp => timestamp ? timestamp : Date.now().valueOf()
```
?

4doge:
It's 2020 and issue is still here.
@pleerock I want to setup column defaults, which run PostgreSQL functions:

```js
@Column({
  default: () => 'md5((now())::text)',
})
token: string;
```

Which is generationg migrations like (which is correct):
```ts
import {MigrationInterface, QueryRunner} from "typeorm";

export class AddDefaultForUserToken1591177402278 implements MigrationInterface {

    public async up(queryRunner: QueryRunner): Promise<any> {
[...]

PR #7681, opened 2021-05-25
fix: correctly strip type conversion in postgres for default values

### LINKED ISSUE #5132: Default of -1 (minus 1) generates useless migrations
**Issue type:**

[ ] question
[X] bug report
[ ] feature request
[ ] documentation issue

**Database system/driver:**

[ ] `cordova`
[ ] `mongodb`
[ ] `mssql`
[ ] `mysql` / `mariadb`
[ ] `oracle`
[X] `postgres`
[ ] `cockroachdb`
[ ] `sqlite`
[ ] `sqljs`
[ ] `react-native`
[ ] `expo`

**TypeORM version:**

[X] `latest`
[ ] `@next`
[X] `0.2.20` (or put your version here)

See repo at https://github.com/anodynos/typeorm-bug-migrations-default-minus-1 
Execute `╰─$ scripts/migration-generate.sh` to reproduce the issue. 

Assuming an Entity : 

```
@Entity('foo')
export class Foo {
  ...
  @Column({ default: 1, type: 'bigint' })
  bar_default_plus_1_is_fine: number;
  
  @Column({ default: -1, type: 'bigint' })
  bar_default_minus_1_is_creating_repeated_migrations: number;
}
```

after the initial migration is added, *every time we generate a migration* (without any schema changes), we get a new migration like:

```
...
        await queryRunner.query(`ALTER TABLE "foo" ALTER COLUMN "bar_default_minus_1_is_creating_repeated_migrations" SET DEFAULT -1`, undefined);
    }

        await queryRunner.query(`ALTER TABLE "foo" ALTER COLUMN "bar_default_minus_1_is_creating_repeated_migrations" SET DEFAULT '-1'`, undefined);
    }
...
```

### LINKED ISSUE #7110: Typeorm Migrations ignore existing default value on column
## Issue Description
In Postgres, it's possible to apply current_setting values as a column default value. 
```      
CREATE TABLE "coffee" ("id" int ,"tenant_id" text DEFAULT current_setting('coffee.current_tenant'::text));

  Column   |  Type   | Collation | Nullable |                    Default                     
-----------+---------+-----------+----------+------------------------------------------------
 id        | integer |           |          | 
 tenant_id | text    |           |          | current_setting('coffee.current_tenant'::text)
```
Thus, when executing 'SET valuename=value', this value is applied as a default for that column.

**What happens actually?**
Typeorm ignores such a present default value. From my investigation so far, I suspect that it is related to the function parameter and the quotes ('coffee.current_tenant'). Using a function without parameters (e.g hotfix()) works perfectly.

Here's an example entity:

```
import { Column, Entity, PrimaryGeneratedColumn } from "typeorm";

@Entity()
export class Coffee {
    @PrimaryGeneratedColumn()
    id: number;

    @Column()
    name: string;
    
    @Column({
        type: "text",
        nullable: false,
        default: () => "current_setting('coffee.current_tenant'::text)",
    })
    tenant_id: string;
}
``` 
When executing `$ npm run start:dev` the database is generated (synchronize: true) and contains the default setting as expected.

```
postgres=# \d coffee
                                         Table "public.coffee"
  Column   |       Type        | Collation | Nullable |                    Default                     
-----------+-------------------+-----------+----------+------------------------------------------------
 id        | integer           |           | not null | nextval('coffee_id_seq'::regclass)
 name      | character varying |           | not null | 
 tenant_id | text              |           | not null | current_setting('coffee.current_tenant'::text)
Indexes:
    "PK_4d27239ee0b99a491ad806aec46" PRIMARY KEY, btree (id)
```
However, it seems that Typeorm Migrations does not recognize this type of default value when generating or checking for migrations using either `typeorm schema:log` or `typeorm migrations:generate -n row-level`.

After generating an additional migration, that migration contains the suggested changes. After applying that new migration, Typeorm still adds the same line for further migrations.

### Expected Behavior
```
$ npx typeorm schema:log
Your schema is up to date - there are no queries to be executed by schema synchronization.
``` 

### Actual Behavior
``` 
$ npx typeorm schema:log
----------------------------------------------------------------
-- Schema syncronization will execute following sql queries (1):
----------------------------------------------------------------
ALTER TABLE "coffee" ALTER COLUMN "tenant_id" SET DEFAULT current_setting('coffee.current_tenant'::text);
```

### Steps to Reproduce

I created a minimal example that shows the behaviour and a workaround. It's available here: https://github.com/fabianskii/typeorm-migration-ignores-default-bug

1. `$ git clone https://github.com/fabianskii/typeorm-migration-ignores-default-bug`
2. `$ cd typeorm-migration-ignores-default-bug/ && npm install`
3. start provided postgres docker: `$ docker-compose up -d`
4. `$ npm run build` (after each migration change)
5. `$ typeorm migration:run
6. `$ typeorm schema:log`
7. `$ typeorm migrations:generate -n row-level
8. `$ npm run build`
9. `$ typeorm schema:log`

```
----------------------------------------------------------------
-- Schema syncronization will execute following sql queries (1):
----------------------------------------------------------------
ALTER TABLE "coffee" ALTER COLUMN "tenant_id" SET DEFAULT current_setting('coffee.current_tenant'::text);
```

### My Environment

| Operating System    |  popOS 20.10        |
| Node.js version     | tried with: v10.19.0; 14.15.1; 15.3.0 | 
| Typescript version  | v4.1.2 | 
| TypeORM version     | v0.2.29 | 

- [ ] `aurora-data-api`
- [ ] `aurora-data-api-pg`
- [ ] `better-sqlite3`
- [ ] `cockroachdb`
- [ ] `cordova`
- [ ] `expo`
- [ ] `mongodb`
- [ ] `mysql`
- [ ] `nativescript`
- [ ] `oracle`
- [x] `postgres`
- [ ] `react-native`
- [ ] `sap`
- [ ] `sqlite`
- [ ] `sqlite-abstract`
- [ ] `sqljs`
- [ ] `sqlserver`

kauz:
Use better regex to remove only type conversion, not the whole string to the end

Closes: 7110

### Description of change

When loading schema with postgres driver typeorm currently strips explicit type conversions from column default value. But current RegExp in some cases does this incorrectly, that leads to part of string after conversion being stripped to the end. 
For example value `to_char(nextval('orders_display_id_seq'::regclass), 'FMU999'::text)` is stripped to just `to_char(nextval('orders_display_id_seq'`. This behaviour causes duplications when generating migrations.  

Fixes #7110
Fixes #5132
### Pull-Request Checklist

- [x] Documentation has been updated to reflect this change - no need

### DIFF
--- a/src/driver/postgres/PostgresQueryRunner.ts
+++ b/src/driver/postgres/PostgresQueryRunner.ts
@@ -1730,7 +1730,7 @@ export class PostgresQueryRunner extends BaseQueryRunner implements QueryRunner 
                         } else if (dbColumn["column_default"] === "now()" || dbColumn["column_default"].indexOf("'now'::text") !== -1) {
                             tableColumn.default = dbColumn["column_default"];
                         } else {
-                            tableColumn.default = dbColumn["column_default"].replace(/::[\w\s]+/g, "");
+                            tableColumn.default = dbColumn["column_default"].replace(/::[\w\s\[\]]+/g, "");
                             tableColumn.default = tableColumn.default.replace(/^(-?\d+)$/, "'$1'");
                         }
                     }
--- a/src/driver/postgres/PostgresQueryRunner.ts
+++ b/src/driver/postgres/PostgresQueryRunner.ts
@@ -1728,9 +1728,9 @@ export class PostgresQueryRunner extends BaseQueryRunner implements QueryRunner 
                             tableColumn.isGenerated = true;
                             tableColumn.generationStrategy = "uuid";
                         } else if (dbColumn["column_default"] === "now()" || dbColumn["column_default"].indexOf("'now'::text") !== -1) {
-                            tableColumn.default = dbColumn["column_default"]
+                            tableColumn.default = dbColumn["column_default"];
                         } else {
-                            tableColumn.default = dbColumn["column_default"].replace(/::.*/, "");
+                            tableColumn.default = dbColumn["column_default"].replace(/::\w+/g, "");
                             tableColumn.default = tableColumn.default.replace(/^(-?\d+)$/, "'$1'");
                         }
                     }
--- a/src/driver/postgres/PostgresQueryRunner.ts
+++ b/src/driver/postgres/PostgresQueryRunner.ts
@@ -1730,7 +1730,7 @@ export class PostgresQueryRunner extends BaseQueryRunner implements QueryRunner 
                         } else if (dbColumn["column_default"] === "now()" || dbColumn["column_default"].indexOf("'now'::text") !== -1) {
                             tableColumn.default = dbColumn["column_default"];
                         } else {
-                            tableColumn.default = dbColumn["column_default"].replace(/::[\w\s\[\]]+/g, "");
+                            tableColumn.default = dbColumn["column_default"].replace(/::[\w\s\[\]\"]+/g, "");
                             tableColumn.default = tableColumn.default.replace(/^(-?\d+)$/, "'$1'");
                         }
                     }
--- a/src/driver/postgres/PostgresQueryRunner.ts
+++ b/src/driver/postgres/PostgresQueryRunner.ts
@@ -1730,7 +1730,7 @@ export class PostgresQueryRunner extends BaseQueryRunner implements QueryRunner 
                         } else if (dbColumn["column_default"] === "now()" || dbColumn["column_default"].indexOf("'now'::text") !== -1) {
                             tableColumn.default = dbColumn["column_default"];
                         } else {
-                            tableColumn.default = dbColumn["column_default"].replace(/::\w+/g, "");
+                            tableColumn.default = dbColumn["column_default"].replace(/::[\w\s]+/g, "");
                             tableColumn.default = tableColumn.default.replace(/^(-?\d+)$/, "'$1'");
                         }
                     }
--- a/test/github-issues/5132/entity/foo.entity.ts
+++ b/test/github-issues/5132/entity/foo.entity.ts
@@ -1,0 +1,13 @@
+import { BaseEntity, Column, Entity, PrimaryGeneratedColumn } from "../../../../src";
+
+@Entity("foo")
+export class Foo extends BaseEntity {
+    @PrimaryGeneratedColumn()
+    id: number;
+
+    @Column({ default: 1, type: "int" })
+    bar_default_1: number;
+
+    @Column({ default: -1, type: "int" })
+    bar_default_minus_1: number;
+}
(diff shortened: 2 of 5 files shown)


kauz:
This should cover all existing types in typeorm. Regex I used for reference: [https://regex101.com/r/qHG4lQ/1](https://regex101.com/r/qHG4lQ/1)

kauz:
Also not sure if these two should be also marked as fixed: 
https://github.com/typeorm/typeorm/issues/5981
https://github.com/typeorm/typeorm/issues/5132

they are pretty old ones

AlexMesser:
#5132 has a reproduction example. Can you please add a test for it?

AlexMesser:
#1729, #3076 also may be fixed

kauz:
> #5132 has a reproduction example. Can you please add a test for it?

Added a test. I've change bigint to int, as it failed in oracle with usupported type errror. To tell the truth, it seems that this issue was already resolved before my changes.

kauz:
> #1729, #3076 also may be fixed

As for #1729, #3076, not sure if we'll still cover all possible cases after that:
From my tests:

`default: () => "NOW()"` - works as expected.

[default: () => "(now() at time zone 'utc')"](https://github.com/typeorm/typeorm/issues/1729#issuecomment-382011405)  - doesn't work, as postgres converts it to `timezone('utc', now())`, but this seems unrelated and I don't know an easy fix for that. You can always use `timezone('utc', now())`

`default: () => "'{}'::jsonb"` and
`default: () => "ARRAY[]::text[]"` - do not work. But in both cases you can use  `default: () => "'{}'"` and postgresql knows how to convert it implicitly.

Complex cases such as `default: () => MD5(NOW()::text))` do not have clear solution for now. Postgres can't convert timestamp to text implicitly in this case, and you're required to use explicit coversion. And even if you use `default: () => "MD5(CAST(NOW() as text))"` it's still converted to `md5((now())::text)`. A possible solution may be to store `tableColumn.defaultRaw` [here](https://github.com/typeorm/typeorm/blob/d27dd2af2ca320e74a17b3ab273cd3bf55d01923/src/driver/postgres/PostgresQueryRunner.ts#L1645-L1646) and compair either raw value or value with stripped types. Let me know if it works for you.

AlexMesser:
> A possible solution may be to store tableColumn.defaultRaw here and compair either raw value or value with stripped types.

I think this is only the way to handle such cases. We have a plan to add a sublayer between ORM and database where we can store mappings between entity and table metadata. It will allow us to handle any comparisons for defaults, checks, expressions, views, etc.

AlexMesser:
thank you for contribution!

Issue #3103, opened 2018-11-13
Add "conditional" method to query builder


lazarljubenovic:
**Issue type:**

[ ] question
[ ] bug report
[x] feature request
[ ] documentation issue

**Database system/driver:** All

**Steps to reproduce or a small repository showing the problem:**

I often conditionally add certain statements from the query based on some parameters. Those parameters in my case usually come from the query strings in a RESTful API, but the same could apply to many other use-cases.

For example:

```ts
const query = somehow
  .createQueryBuilder('users')
  .where(`user.deletedAt IS NOT NULL`)

if (params.hasImage) {
  query.andWhere(`user.image IS NOT NULL`)
}

if (params.q != '') {
  query.andWhere(`user.name LIKE :q`, { q: params.q })
}

const users = await sql.getMany()
```

This quite breaks the flow of the nice fluid chainable API that the query builder has. I propose a method `.if` or `.conditional` which accepts two arguments and then runs one of the given functions based on the given condition. 

```ts
const users = await somehow
  .createQueryBuilder('users')
  .where(`user.deletedAt IS NOT NULL`)
  .if(params.hasImage, query => query.andWhere(`user.image IS NOT NULL`))
  .if(params.q != '', query => query.andWhere(`user.username LIKE :q`, { q: params.q }))
  .getMany()
```

`.if` could accept a third argument which would be an `else` branch.

Thoughts?

Kononnable:
I'm not sure if it wouldn't make code less readable. We would introduce two/three parameters and lines will become very long.
Why not use plain ifs?
```typescript
let query = somehow
  .createQueryBuilder('users')
  .where(`user.deletedAt IS NOT NULL`);
if(params.hasImage)
  query=query.andWhere(`user.image IS NOT NULL`);
if(params.q != '')
  query=query.andWhere(`user.username LIKE :q`, { q: params.q });
const users = query.getMany();
```

pleerock:
@lazarljubenovic I completely understand your request and wanted this functionality for years, but when it comes to implementation another understanding comes: 

* it cover only some cases, when you'll have more non-trivial cases you'll have to fallback to if/else statements, and mixing both approaches isn't that fun anymore
* as @Kononnable mentioned it can be less readable, it works good mostly for a very short conditions and queries
* it breaks query builder semantics a bit since `if` isnt about building query
* it will bring lot of new questions regarding to this functionality and its inflexibility in the future

Kononnable:
[...]

PR #7455, opened 2021-03-09
feat: condition to enable where

fanuelson:
Fixes: #7405 

### Description of change

Now is possible to enable where based on condition
```ts
const searchTitle = undefined;
const posts = await postRepository.createQueryBuilder("p")
    .andWhere("p.title = :title", { title: searchTitle}, searchTitle !== undefined) <---
    .getMany();
```
This will result: ```select p from Post p```

### Pull-Request Checklist

- [x] Code is up-to-date with the `master` branch
- [x] `npm run lint` passes with this change
- [x] `npm run test` passes with this change
- [x] This pull request links relevant issues as `Fixes #0000`
- [x] There are new or updated unit tests validating the change
- [x] Documentation has been updated to reflect this change
- [x] The new commits follow conventions explained in [CONTRIBUTING.md](https://github.com/typeorm/typeorm/blob/master/CONTRIBUTING.md)

PR #7684, opened 2021-05-26
Improve conditional where (andWhere / orWhere)

jojoldu:
### Description of change

Suggest a new method for conditional where queries that were previously discussed(https://github.com/typeorm/typeorm/issues/3103).

If null is received, the corresponding `andWhere` / `orWhere` condition is not added to the conditional statement.

Querydsl library of JAVA language supports similar type.
So there is no big burden to actually make a dynamic query.

```java
// Querydsl)

query.from(customer)
     .where(
         isFirstNameLike(customerQueryInfo.getFirstName()), // if null return then erased condition
         isLastNameLike(customerQueryInfo.getLastName()),
         isStatusEq(customerQueryInfo.getStatus()))
     .list(customer);

private BooleanExpression isFirstNameLike(String firstName){
    return firstName != null ? customer.firstName.like(firstName) : null;        
}

private BooleanExpression isStatusEq(StatusEnum status){
    return status != null ? customer.status.eq(status) : null;
}

```
* https://stackoverflow.com/a/12287074

In typeORM, can use it like this:

Expect the following code.

```javascript
    findAllByDto (dto: ArticleSearchDto) {
        return createQueryBuilder()
            .select("article")
            .from(Article, "article")
            .andWhere(this.eqTitle(dto.title)) // or title? `article.title = :${title}` : null;
            .orderBy("article.id desc")
            .getMany();
    }
  
   // erase or andWhere
    private eqTitle(title: string): string {
        if(!title) {
            return null;
        }

        return `article.title = :${title}`;
    }
```

### Pull-Request Checklist

### DIFF
--- a/src/query-builder/SelectQueryBuilder.ts
+++ b/src/query-builder/SelectQueryBuilder.ts
@@ -729,7 +729,11 @@ export class SelectQueryBuilder<Entity> extends QueryBuilder<Entity> implements 
      * Adds new AND WHERE condition in the query builder.
      * Additionally you can add parameters used in where expression.
      */
-    andWhere(where: string|Brackets|((qb: this) => string), parameters?: ObjectLiteral): this {
+    andWhere(where: string|Brackets|null|((qb: this) => string), parameters?: ObjectLiteral): this {
+        if(!where) {
+            return this;
+        }
+
         this.expressionMap.wheres.push({ type: "and", condition: this.computeWhereParameter(where) });
         if (parameters) this.setParameters(parameters);
         return this;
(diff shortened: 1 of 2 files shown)


michaelzangl:
I don't like this idea - if you don't want to add a condition, just don't add it. Just ignoring the addition is an error that is really hard to spot in production.

imnotjames:
The example in your PR is a SQL injection threat.

Can you include the examples you pulled this from?  (Ideally 3 from varying DSLs)

My biggest concern with builders like these is the confusing order of operations.

imnotjames:
Also, to make this more explicit, I think we should - if ever - implement it similar to laravel's `when`.

This implementation makes it very easy to unintentionally omit a `where` clause.

AlexMesser:
Duplicate of #7455 and closing for the same reason:

> This is redundant. We can't continue extending query builder functionality to fit all user demands. You can simply use "if" statement to do same without any changes in the codebase.

PR #7694, opened 2021-05-28
Fix Caching won't work with replication enabled

### LINKED ISSUE #5919: ⚠Caching won't work with replication enabled
**Issue type:**

[ ] question
[ x] bug report
[ ] feature request
[ ] documentation issue

**Database system/driver:**

[ ] `cordova`
[ ] `mongodb`
[ ] `mssql`
[ ] `mysql` / `mariadb`
[ ] `oracle`
[x ] `postgres`
[ ] `cockroachdb`
[ ] `sqlite`
[ ] `sqljs`
[ ] `react-native`
[ ] `expo`

**TypeORM version:**

[x ] `latest`
[ ] `@next`

Hi everyone, I've configured typeorm with replication using two instances, a master(read-write) and a slave(read), so if my understading is correct all the select/get queries are executed in the slave instance and all the updates/set/insert/etc are executed in the master db but today I've enabled caching for my db and it doesn't work, it shows the following error when I try to get something from the db:

`{ error: cannot execute INSERT in a read-only transaction
    at Connection.parseE (/app/node_modules/pg/lib/connection.js:604:13)
    at Connection.parseMessage (/app/node_modules/pg/lib/connection.js:403:19)
    at Socket.<anonymous> (/app/node_modules/pg/lib/connection.js:123:22)
    at Socket.emit (events.js:198:13)
    at Socket.EventEmitter.emit (domain.js:448:20)
    at addChunk (_stream_readable.js:287:12)
    at readableAddChunk (_stream_readable.js:268:11)
    at Socket.Readable.push (_stream_readable.js:223:10)
    at TCP.onStreamRead [as onread] (internal/stream_base_commons.js:94:17)
  name: 'error',
  length: 118,
  severity: 'ERROR',
  code: '25006',
  detail: undefined,
  hint: undefined,
  position: undefined,
  internalPosition: undefined,
  internalQuery: undefined,
  where: undefined,
  schema: undefined,
  table: undefined,
  column: undefined,
  dataType: undefined,
  constraint: undefined,
  file: 'utility.c',
  line: '246',
  routine: 'PreventCommandIfReadOnly' }
`

I suppose that this is being caused when a get request is received on my API and then typeorm selects the slave database in order to execute the query and then it tries to do a `INSERT`  to chache the results using the slave instance and because it is `READ ONLY` it throws that error, so I guess that this is a bug, correct me if I'm wrong, thanks in advance!

madeindjs:
Fix #5919 

### Description of change

I just added a new method `QueryRunner.getReplicationMode` to check if we can insert cache into `DbQueryResultCache.storeInCache`. If query runner is in slave mode, we create a new one in master mode

### Pull-Request Checklist

- [x]  `npm run lint` passes with this change
- [x] Documentation has been updated to reflect this change N/A

### DIFF
--- a/src/cache/DbQueryResultCache.ts
+++ b/src/cache/DbQueryResultCache.ts
@@ -158,7 +158,9 @@ export class DbQueryResultCache implements QueryResultCache {
      * Stores given query result in the cache.
      */
     async storeInCache(options: QueryResultCacheOptions, savedCache: QueryResultCacheOptions|undefined, queryRunner?: QueryRunner): Promise<void> {
-        queryRunner = this.getQueryRunner(queryRunner);
+        if (queryRunner === undefined || queryRunner?.getReplicationMode() === "slave") {
+            queryRunner = this.connection.createQueryRunner();
+        }
 
         let insertedValues: ObjectLiteral = options;
         if (this.connection.driver instanceof SqlServerDriver) { // todo: bad abstraction, re-implement this part, probably better if we create an entity metadata for cache table
--- a/src/cache/DbQueryResultCache.ts
+++ b/src/cache/DbQueryResultCache.ts
@@ -158,8 +158,10 @@ export class DbQueryResultCache implements QueryResultCache {
      * Stores given query result in the cache.
      */
     async storeInCache(options: QueryResultCacheOptions, savedCache: QueryResultCacheOptions|undefined, queryRunner?: QueryRunner): Promise<void> {
-        if (queryRunner === undefined || queryRunner?.getReplicationMode() === "slave") {
-            queryRunner = this.connection.createQueryRunner();
+        const shouldCreateQueryRunner = queryRunner === undefined || queryRunner?.getReplicationMode() === "slave";
+
+        if (queryRunner === undefined || shouldCreateQueryRunner) {
+            queryRunner = this.connection.createQueryRunner("master");
         }
 
         let insertedValues: ObjectLiteral = options;
@@ -204,6 +206,10 @@ export class DbQueryResultCache implements QueryResultCache {
                 .into(this.queryResultCacheTable)
                 .values(insertedValues)
                 .execute();
+        }
+
+        if (shouldCreateQueryRunner) {
+            await queryRunner.release();
         }
     }
 
--- a/src/driver/mongodb/MongoQueryRunner.ts
+++ b/src/driver/mongodb/MongoQueryRunner.ts
@@ -1,20 +1,10 @@
-import {ObjectLiteral} from "../../common/ObjectLiteral";
-import {Connection} from "../../connection/Connection";
-import {MongoEntityManager} from "../../entity-manager/MongoEntityManager";
-import {TypeORMError} from "../../error";
-import {ReadStream} from "../../platform/PlatformTools";
 import {QueryRunner} from "../../query-runner/QueryRunner";
-import {Table} from "../../schema-builder/table/Table";
-import {TableCheck} from "../../schema-builder/table/TableCheck";
+import {ObjectLiteral} from "../../common/ObjectLiteral";
 import {TableColumn} from "../../schema-builder/table/TableColumn";
-import {TableExclusion} from "../../schema-builder/table/TableExclusion";
+import {Table} from "../../schema-builder/table/Table";
 import {TableForeignKey} from "../../schema-builder/table/TableForeignKey";
 import {TableIndex} from "../../schema-builder/table/TableIndex";
-import {TableUnique} from "../../schema-builder/table/TableUnique";
 import {View} from "../../schema-builder/view/View";
-import {Broadcaster} from "../../subscriber/Broadcaster";
-import {SqlInMemory} from "../SqlInMemory";
-import {ReplicationMode} from "../types/ReplicationMode";
 import {
     AggregationCursor,
     BulkWriteOpResultObject,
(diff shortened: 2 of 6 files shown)


imnotjames:
Adding extra API surface area & exposing more internal details kinda seems.. well, a bit over-kill for a feature that shouldn't be so tightly wound up in TypeORM anyway.

I think it'd be reasonable for `DbQueryResultCache` to just create a new query runner instead.  I'd just need to confirm it resolves the issue.

madeindjs:
@imnotjames , to be sure to have understood correctly, you propose:

1. remove `BaseQueryRunner.getReplicationMode` that I added
2. in `DbQueryResultCache.storeInCache`, remove `queryRunner` options and always create a new `queryRunner` inside

imnotjames:
> @imnotjames , to be sure to have understood correctly, you propose:
> 
> 1. remove `BaseQueryRunner.getReplicationMode` that I added
> 2. in `DbQueryResultCache.storeInCache`, remove `queryRunner` options and always create a new `queryRunner` inside
> 
> 

Yes.  Would that cause much overhead to create a new query runner, you think?

I feel like it's the most consistent.  You'd just need to release the query runner afterwards in a finally block.

madeindjs:
I do not think this is a good idea because:

1. `queryRunner`is passed everywhere. If I remove it from `storeInCache`, I should do it also on theses methods to be consitent
~~~diff
export interface QueryResultCache {
    connect(): Promise<void>;
    disconnect(): Promise<void>;
-   synchronize(queryRunner?: QueryRunner): Promise<void>;
+   synchronize(): Promise<void>;
    getFromCache(options: QueryResultCacheOptions, queryRunner?: QueryRunner): Promise<QueryResultCacheOptions|undefined>;
-   storeInCache(options: QueryResultCacheOptions, savedCache: QueryResultCacheOptions|undefined, queryRunner?: QueryRunner): Promise<void>;
+   storeInCache(options: QueryResultCacheOptions, savedCache: QueryResultCacheOptions|undefined): Promise<void>;
    isExpired(savedCache: QueryResultCacheOptions): boolean;
-   clear(queryRunner?: QueryRunner): Promise<void>;
+   clear(): Promise<void>;
-   remove(identifiers: string[], queryRunner?: QueryRunner): Promise<void>;
+   remove(identifiers: string[]): Promise<void>;
}
~~~
2. if we pass a `queryRunner` from a transaction, this will imply that cache will be stored before transaction commit. Even if we use a master `queryRunner`

What do you think about it?

imnotjames:
> if we pass a queryRunner from a transaction, this will imply that cache will be stored before transaction commit. Even if we use a master queryRunner

This is happening anyway with the code you're suggesting in the case of using a replica.

I almost think we shouldn't fix this at all is that's a concern and instead cause DBResultCache to bail out if you're trying to use it with any sort of replication enabled.

madeindjs:
> This is happening anyway with the code you're suggesting in the case of using a replica.

Not necessary if the user use a `querRunner` from a transaction on `master` server.

> instead cause DBResultCache to bail out if you're trying to use it with any sort of replication enabled.

I don't think so. Currently, I have an application on a production environment I want to scale with Postgres replication but I want to keep the cache system.

But if you ok with [modifications I proposed on `QueryResultCache`](https://github.com/typeorm/typeorm/pull/7694#issuecomment-862225392), I can give it a try and update this PR. What do you think about it?

imnotjames:
> > This is happening anyway with the code you're suggesting in the case of using a replica.
> 
> Not necessary if the user use a `querRunner` from a transaction on `master` server.

Which is why I said in the case of using a replica which is 99% of the time when you're caching.

> 
> > instead cause DBResultCache to bail out if you're trying to use it with any sort of replication enabled.
> 
> I don't think so. Currently, I have an application on a production environment I want to scale with Postgres replication but I want to keep the cache system.

You would be really better off building your own caching mechanism if you need specific code paths.

Also, if you're selecting in the middle of a transaction you maybe shouldn't be caching that value - because then you'd be getting the data from before your operations.  The whole point would be that you know changes have been made.

> 
> But if you ok with [modifications I proposed on `QueryResultCache`](https://github.com/typeorm/typeorm/pull/7694#issuecomment-862225392), I can give it a try and update this PR. What do you think about it?

I think the optimal solution would be to omit query runner's use in the built-in implementation and mark those field as deprecated in the interface but still pass them to the interface.

That would give folks time to migrate away from it for their own implementation.

Then at a later date we can remove these.  Otherwise everyone gets very grumpy. 😄 I mean, understandably.

pleerock:
Okay so here I think you both have valid points @madeindjs  and @imnotjames just different opinions.

If you ask me,

* I'm okay with new `getReplicationMode()` function, it doesn't bring something complex and logically makes sense to be part of QueryRunner

* I'm okay with proposed solution on using given query runner or create new one if needed.

here @imnotjames is right because in scaled app most of times queries will be served by read replicas and having this check doesn't make much sense. However, in specific user scenarios, when for example he opens complex transaction with saves and queries, it will be better to keep using given query runner in the case if transaction fail (e.g. data was inserted, "finds" executed, "cached" in a separate connection, but then transaction fails and we have no data inserted, but we do have data in the cache).  

* if user wants to use db cache for production, that's completely fine. Yes maybe it's not the most performant solution, but its the most performant from simplest solutions. And since we have such solution, and this bug makes this solution unusable - I think we need to fix it.

I'm good with this PR, we just need to fix conflicts.
Sorry for late review.

madeindjs:
Great @pleerock . I currently [use this branch on production](https://www.npmjs.com/package/@optimiz.me/typeorm) since june. So I'll try to correct conflicts as soon as possible.

madeindjs:
@pleerock , I just resolved conflicts.

pleerock:
Thank you for contribution! 🎉

PR #7712, opened 2021-06-06
add example to getting relation id without joining relation

juanp123:
### Description of change
The example in the documentation shows how to get a relation id without joining the relation but this does not work if the variable name of the column does not match the actual database column name.   Added an example

### DIFF
--- a/docs/relations-faq.md
+++ b/docs/relations-faq.md
@@ -120,7 +120,18 @@ export class User {
     
 }
 ```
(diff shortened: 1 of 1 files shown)


pleerock:
Thank you for contribution, however I think this is a bit redundant in terms of official documentation. Instead, I really suggest to create a StackOverflow question and answer it by yourself.

PR #7725, opened 2021-06-09
feat: make postgres extensions install optional

### LINKED ISSUE #7662: Creating extensions in Postgres should be optional
## Feature Description

PostgresDriver.enableExtensions should be an optional function.

### The Problem

For our specific use-case, we use TypeORM in a service that is using read-only client credentials. I imagine this to be a common occurrence. Since we have a `uuid` column in at least 1 of our entities, this line runs https://github.com/typeorm/typeorm/blob/master/src/driver/postgres/PostgresDriver.ts#L336-L341:

```
if (hasUuidColumns)
            try {
                await this.executeQuery(connection, `CREATE EXTENSION IF NOT EXISTS "${this.options.uuidExtension || "uuid-ossp"}"`);
            } catch (_) {
                logger.log("warn", `At least one of the entities has uuid column, but the '${this.options.uuidExtension || "uuid-ossp"}' extension cannot be installed automatically. Please install it manually using superuser rights, or select another uuid extension.`);
            }
```

Although there is no _real_ issue, we are being bombarded with logs from both Postgres and the App especially on those that periodically go up and down (think crons) 

### The Solution

One of the ways this can be solved by adding a flag at the start of the function to determine whether it should run or not. There are many (possibly better) ways to achieve this same effect.

### Considered Alternatives

Add a flag somewhere else to disable automatically running a `CREATE EXTENSION` script

- [ ] `aurora-data-api`
- [ ] `aurora-data-api-pg`
- [ ] `better-sqlite3`
- [ ] `cockroachdb`
- [ ] `cordova`
- [ ] `expo`
- [ ] `mongodb`
- [ ] `mysql`
- [ ] `nativescript`
- [ ] `oracle`
- [x] `postgres`
- [ ] `react-native`
- [ ] `sap`
- [ ] `sqlite`
- [ ] `sqlite-abstract`
- [ ] `sqljs`
- [ ] `sqlserver`

- [x] No, I don't have the time and I wouldn't even know how to start.

jansenignacio:
There are cases wherein the credentials given to the service are READONLY. In these cases, installing extensions automatically fails. There is no reason to run these everytime as they trigger unnecessary error logs.

Fixes #7662

### Pull-Request Checklist

### DIFF
--- a/docs/connection-options.md
+++ b/docs/connection-options.md
@@ -1,507 +1,510 @@
 # Connection Options
 
-* [What is `ConnectionOptions`](#what-is-connectionoptions)
-* [Common connection options](#common-connection-options)
-* [`mysql` / `mariadb` connection options](#mysql--mariadb-connection-options)
-* [`postgres` / `cockroachdb` connection options](#postgres--cockroachdb-connection-options)
-* [`sqlite` connection options](#sqlite-connection-options)
-* [`better-sqlite3` connection options](#better-sqlite3-connection-options)
-* [`capacitor` connection options](#capacitor-connection-options)
-* [`cordova` connection options](#cordova-connection-options)
-* [`react-native` connection options](#react-native-connection-options)
-* [`nativescript` connection options](#nativescript-connection-options)
-* [`mssql` connection options](#mssql-connection-options)
-* [`mongodb` connection options](#mongodb-connection-options)
-* [`sql.js` connection options](#sqljs-connection-options)
-* [`expo` connection options](#expo-connection-options)
-* [Connection options example](#connection-options-example)
-    
+-   [What is `ConnectionOptions`](#what-is-connectionoptions)
+-   [Common connection options](#common-connection-options)
+-   [`mysql` / `mariadb` connection options](#mysql--mariadb-connection-options)
+-   [`postgres` / `cockroachdb` connection options](#postgres--cockroachdb-connection-options)
+-   [`sqlite` connection options](#sqlite-connection-options)
+-   [`better-sqlite3` connection options](#better-sqlite3-connection-options)
+-   [`capacitor` connection options](#capacitor-connection-options)
+-   [`cordova` connection options](#cordova-connection-options)
+-   [`react-native` connection options](#react-native-connection-options)
+-   [`nativescript` connection options](#nativescript-connection-options)
+-   [`mssql` connection options](#mssql-connection-options)
+-   [`mongodb` connection options](#mongodb-connection-options)
+-   [`sql.js` connection options](#sqljs-connection-options)
+-   [`expo` connection options](#expo-connection-options)
+-   [Connection options example](#connection-options-example)
+
 ## What is `ConnectionOptions`
 
 Connection options is a connection configuration you pass to `createConnection`
- or define in `ormconfig` file. Different databases have their own specific connection options.
+or define in `ormconfig` file. Different databases have their own specific connection options.
 
 ## Common connection options
 
-* `type` - Database type. You must specify what database engine you use.
- Possible values are "mysql", "postgres", "cockroachdb", "mariadb", "sqlite", "better-sqlite3", "capacitor", "cordova", 
- "nativescript", "oracle", "mssql", "mongodb", "sqljs", "react-native".
- This option is **required**.
+-   `type` - Database type. You must specify what database engine you use.
+    Possible values are "mysql", "postgres", "cockroachdb", "mariadb", "sqlite", "better-sqlite3", "capacitor", "cordova",
+    "nativescript", "oracle", "mssql", "mongodb", "sqljs", "react-native".
+    This option is **required**.
 
-* `name` - Connection name. You'll use it to get connection you need using `getConnection(name: string)` 
-or `ConnectionManager.get(name: string)`. 
-Connection names for different connections cannot be same - they all must be unique.
-If connection name is not given then it will be called "default".
+-   `name` - Connection name. You'll use it to get connection you need using `getConnection(name: string)`
+    or `ConnectionManager.get(name: string)`.
+    Connection names for different connections cannot be same - they all must be unique.
+    If connection name is not given then it will be called "default".
 
-* `extra` - Extra connection options to be passed to the underlying driver. 
-Use it if you want to pass extra settings to underlying database driver.
+-   `extra` - Extra connection options to be passed to the underlying driver.
+    Use it if you want to pass extra settings to underlying database driver.
 
-* `entities` - Entities, or Entity Schemas, to be loaded and used for this connection.
-Accepts both entity classes, entity schema classes, and directories paths to load from.
-Directories support glob patterns.
-Example: `entities: [Post, Category, "entity/*.js", "modules/**/entity/*.js"]`.
-Learn more about [Entities](./entities.md). 
-Learn more about [Entity Schemas](separating-entity-definition.md).
+-   `entities` - Entities, or Entity Schemas, to be loaded and used for this connection.
+    Accepts both entity classes, entity schema classes, and directories paths to load from.
+    Directories support glob patterns.
+    Example: `entities: [Post, Category, "entity/*.js", "modules/**/entity/*.js"]`.
+    Learn more about [Entities](./entities.md).
+    Learn more about [Entity Schemas](separating-entity-definition.md).
 
-* `subscribers` - Subscribers to be loaded and used for this connection.
-Accepts both entity classes and directories to load from.
-Directories support glob patterns.
-Example: `subscribers: [PostSubscriber, AppSubscriber, "subscriber/*.js", "modules/**/subscriber/*.js"]`.
-Learn more about [Subscribers](listeners-and-subscribers.md).
+-   `subscribers` - Subscribers to be loaded and used for this connection.
+    Accepts both entity classes and directories to load from.
+    Directories support glob patterns.
+    Example: `subscribers: [PostSubscriber, AppSubscriber, "subscriber/*.js", "modules/**/subscriber/*.js"]`.
+    Learn more about [Subscribers](listeners-and-subscribers.md).
 
-* `migrations` - Migrations to be loaded and used for this connection.
-Accepts both migration classes and directories to load from.
-Directories support glob patterns.
-Example: `migrations: [FirstMigration, SecondMigration, "migration/*.js", "modules/**/migration/*.js"]`.
-Learn more about [Migrations](./migrations.md).
+-   `migrations` - Migrations to be loaded and used for this connection.
+    Accepts both migration classes and directories to load from.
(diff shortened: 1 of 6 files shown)


docs/connection-options.md
```diff
@@ -185,6 +185,8 @@ See [SSL options](https://github.com/mysqljs/mysql#ssl-options).
 
 * `logNotifications` - A boolean to determine whether postgres server [notice messages](https://www.postgresql.org/docs/current/plpgsql-errors-and-messages.html) and [notification events](https://www.postgresql.org/docs/current/sql-notify.html) should be included in client's logs with `info` level (default: `false`).
 
+* `skipInstallExtensions` - A boolean to control whether to install necessary postgres extensions automatically or not (default: `true`)
```
imnotjames, line 188: The documented default is incorrect.

Also, why not just make this `installExtensions` instead of negating it?

jansenignacio: > The documented default is incorrect.

True I overlooked this.


> Also, why not just make this installExtensions instead of negating it?

I initially did it that way (`installExtensions`) but I couldn't figure out how to make it `true` by default so I opted for `skipInstallExtensions` which will be `false` by default. We don't want to change the existing behavior

jansenignacio: I would prefer to rename it `installExtensions` if you can lead me to how I can make it default to `true`

jansenignacio: @imnotjames In any case, I've updated this documentation to make it correct for this case

imnotjames:
This could be optional but still `installExtensions` by checking if the value is undefined or true.

Also, can you add tests that confirm the behavior?

jansenignacio:
> This could be optional but still `installExtensions` by checking if the value is undefined or true.
> 
> Also, can you add tests that confirm the behavior?

Can you point me in the right direction of how to write tests for this? I couldn't find any example of tests for `afterConnect` or `enableExtensions` specifically. 

I can tell you that I knew to make it `skipInstallExtensions` because the tests initially broke when it was just `installExtensions`

docs/connection-options.md
```diff
@@ -1,507 +1,510 @@
...
-* [`better-sqlite3` connection options](#better-sqlite3-connection-options)
-* [`capacitor` connection options](#capacitor-connection-options)
-* [`cordova` connection options](#cordova-connection-options)
-* [`react-native` connection options](#react-native-connection-options)
-* [`nativescript` connection options](#nativescript-connection-options)
-* [`mssql` connection options](#mssql-connection-options)
-* [`mongodb` connection options](#mongodb-connection-options)
-* [`sql.js` connection options](#sqljs-connection-options)
-* [`expo` connection options](#expo-connection-options)
-* [Connection options example](#connection-options-example)
```
imnotjames, line 17: Can you undo all these formatting changes

imnotjames:
> Can you point me in the right direction of how to write tests for this? I couldn't find any example of tests for afterConnect or enableExtensions specifically.

Maybe something like this?

https://github.com/typeorm/typeorm/blob/69d3e0644a38d4deba8aca3461e2afae252666cc/test/github-issues/2703/issue-2703.ts#L8

jansenignacio:
> > Can you point me in the right direction of how to write tests for this? I couldn't find any example of tests for afterConnect or enableExtensions specifically.
> 
> Maybe something like this?
> 
> https://github.com/typeorm/typeorm/blob/69d3e0644a38d4deba8aca3461e2afae252666cc/test/github-issues/2703/issue-2703.ts#L8

I don't see how this helps me create the test, actually. Spent most of today trying to figure out how to create the test for this but no go. I don't see any tests that specifically test anything related to postgres extensions. Almost everything I see are tests wherein **the connection has already been established**. I can't a test or a test pattern to test the `afterConnect` behavior of the Driver.

imnotjames:
The test adds an in-memory query logger.  The `CREATE EXTENSION` query should be logged into that logger, shouldn't it?

jansenignacio:
I actually tested that but the `CREATE EXTENSION` was never logged even when I played around with different settings. Thanks to your explanation there, I managed to figure out that the reason it wasn't logged was because...it wasn't being logged!

Please check if the way I added the log for the `CREATE EXTENSION` queries there are correct.

---
I also apologize for the way I constructed my previous response. It was quite arrogant. You maintainers of good open source software don't deserve that kind of language. I thank you for your patience and hard work!

src/driver/postgres/PostgresDriver.ts
```diff
@@ -1084,6 +1085,7 @@ export class PostgresDriver implements Driver {
      */
     protected executeQuery(connection: any, query: string) {
         return new Promise((ok, fail) => {
+            this.connection.logger.logQuery(query);
```
jansenignacio, line 1088: @imnotjames I'm unsure if this is the right way to do this or not

imnotjames: Yeah, it probably should have been doing that anyway

jansenignacio:
@imnotjames can you help me out with the failing tests? I believe they're failing when the DB it's connecting to is not postgres but I'm not really sure. I can't replicate it on my local. I'm wondering if I need to put a short circuit in the test suite to skip it if the DB it's connecting to is not postgres...but I also don't know how to best do that.

test/github-issues/7662/issue-7662.ts
```diff
@@ -0,0 +1,105 @@
...
+            expect(connections).to.have.length(1);
+
+            connection = connections[0];
+
+            const logger = connection.logger as MemoryLogger;
+            const createExtensionQueries = logger.queries.filter((q) =>
+                q.startsWith("CREATE EXTENSION IF NOT EXISTS")
+            );
+
+            expect(createExtensionQueries).to.be.empty;
```
jansenignacio, line 33: This doesn't make sense 🤷 I'm so confused why it's erroring out like that. I can't replicate it locally

Screen Shot 2021-06-24 at 9 33 00 PM

jansenignacio:
Are the circleci tests really flaky? The Node12 Oracle pipeline failed but it doesn't show any errors and my change passed with no issues


Screen Shot 2021-06-24 at 9 33 07 PM

test/github-issues/7662/issue-7662.ts
imnotjames: If you think it's something odd that only shows up in CI You can always push it with a temporary change to print out the values before the expect?

imnotjames:
> Are the circleci tests really flaky? The Node12 Oracle pipeline failed but it doesn't show any errors and my change passed with no issues
> 
> Screen Shot 2021-06-24 at 9 33 07 PM

Looking at the test it is running and failing there.

test/github-issues/7662/issue-7662.ts
```diff
@@ -0,0 +1,105 @@
...
+                enabledDrivers: ["postgres"],
+                schemaCreate: false,
+                dropSchema: true,
+                createLogger: () => new MemoryLogger(true),
+                driverSpecific: {
+                    installExtensions: false,
+                },
+            });
+
+            expect(connections).to.have.length(1);
```
imnotjames, line 24: This won't be true in all test environments.

This is why your tests are failing.

In the oracle tests it fails because we aren't running postgres there so it gets omitted.

Instead, return early if the connections length are 0 or something like that

test/github-issues/7662/issue-7662.ts
imnotjames: This failed expectation is a few lines above, not this line.

jansenignacio: AHHH 💡 thanks bro. That was exactly it

docs/connection-options.md
```diff
@@ -185,6 +185,7 @@ See [SSL options](https://github.com/mysqljs/mysql#ssl-options).
 
 * `logNotifications` - A boolean to determine whether postgres server [notice messages](https://www.postgresql.org/docs/current/plpgsql-errors-and-messages.html) and [notification events](https://www.postgresql.org/docs/current/sql-notify.html) should be included in client's logs with `info` level (default: `false`).
 
+* `installExtensions` - A boolean to control whether to install necessary postgres extensions automatically or not (default: `true`)
```
imnotjames, line 188: Kind of a nit-pick but can you add another line under this.

src/driver/postgres/PostgresDriver.ts
```diff
@@ -304,8 +304,9 @@ export class PostgresDriver implements Driver {
     async afterConnect(): Promise<void> {
         const extensionsMetadata = await this.checkMetadataForExtensions();
 
-        if (extensionsMetadata.hasExtensions) {
-            return new Promise<void>((ok, fail) => {
+        const installExtensions = this.options.installExtensions === undefined || this.options.installExtensions === null || this.options.installExtensions === true;
```
imnotjames, line 307: `this.options.installExtensions === null` is impossible with the typing.

test/github-issues/7662/issue-7662.ts
```diff
@@ -0,0 +1,113 @@
...
+                driverSpecific: {
+                    installExtensions: false,
+                },
+            });
+
+            if (connections.length < 1) {
+                this.skip();
+            }
+
+            expect(connections).to.have.length(1);
```
imnotjames, line 28: You can omit this expectation.

test/github-issues/7662/issue-7662.ts
```diff
@@ -0,0 +1,113 @@
...
+                schemaCreate: false,
+                dropSchema: true,
+                createLogger: () => new MemoryLogger(true),
+                driverSpecific: {
+                    installExtensions: false,
+                },
+            });
+
+            if (connections.length < 1) {
+                this.skip();
```
imnotjames, line 25: You should probably `return` so you don't run the rest of the code.

I think that `this.skip()` emits an error somehow but it'd be clearer if you `return this.skip()` or `this.skip();` followed by `return;`

test/github-issues/7662/issue-7662.ts
```diff
@@ -0,0 +1,113 @@
...
+            const logger = connection.logger as MemoryLogger;
+            const createExtensionQueries = logger.queries.filter((q) =>
+                q.startsWith("CREATE EXTENSION IF NOT EXISTS")
+            );
+
+            expect(createExtensionQueries).to.be.empty;
+        } finally {
+            if (connection) {
+                const logger = connection.logger as MemoryLogger;
+                logger.clear();
```
imnotjames, line 41: Since you're creating a new one every time you don't need to do this.

Same for below.

test/github-issues/7662/issue-7662.ts
```diff
@@ -0,0 +1,113 @@
...
+                await closeTestingConnections([connection]);
```
imnotjames, line 42: This can be 

```
closeTestingConnections(connections);
```

right?

Same for below.

imnotjames (approved):

PR #7728, opened 2021-06-09
fix: Capacitor driver PRAGMA requests failing on Android

chriswep:
- fix: PRAGMA needs to be run through the query method, since it returns data (Android sqlite rejects the request otherwise)
- added WAL journal mode


imnotjames:
Just to confirm - there's no way of testing this, is there?  Aside from running it on Android.

src/driver/capacitor/CapacitorDriver.ts
```diff
@@ -81,9 +81,14 @@ export class CapacitorDriver extends AbstractSqliteDriver {
...
         );
         await connection.open();
+
         // we need to enable foreign keys in sqlite to make sure all foreign key related features
         // working properly. this also makes onDelete to work with sqlite.
-        await connection.execute(`PRAGMA foreign_keys = ON;`);
+        await connection.query(`PRAGMA foreign_keys = ON`, []);
+
+        // turn on WAL mode to enhance performance
+        await connection.query(`PRAGMA journal_mode = WAL`, []);
```
imnotjames, line 90: Shouldn't someone do this themselves if they want to enable it?

Why always enable it?  Per the comment, it's for performance?  For writes that might be true, for reads it's known to be slightly slower.

chriswep: i added `journalMode` as an optional setting

chriswep:
> Just to confirm - there's no way of testing this, is there? Aside from running it on Android.

i don't think there is a reasonable way within the scope of this project. if i didn't overlook something, i also don't see other drivers testing something like that (or test anything at all..). probably another reason to have plugin drivers in the future?

pleerock:
> if i didn't overlook something, i also don't see other drivers testing something like that

correct and that's why when somebody changes something and break something on these drivers - we don't even realize it. Then people have different issues they report. If we can automatize test running on this (or any other which isn't automatized) - it will save us a lot from many issues. And this question is not related to the future - automatization happens once no matter how project is structured.

chriswep:
@pleerock i agree. fwiw, i was bitten by this myself more than once after upgrading to a new TypeORM release. however in those cases the issues were changes outside the direct scope of the driver. So i guess what i was trying to say: moving drivers out of core and importing them as plugins prevents them from breaking other stuff on a conceptual level. furthermore it makes them easier to test. but i guess that is something we would all agree on anyway. 

as far as this PR is concerned, i think it wouldn't make sense to attempt testing this. this code doesn't affect other parts of TypeORM and this is about conforming to the SQLite spec in combination with the Android Interface which i failed concerning the `PRAGMA` command in the initial implementation. if drivers would test at this level, they'd need to test against the whole spec of their respective engine in combination with every supported client (in my case, multiple iOS and Android versions). maybe a more achievable goal would be a general purpose test that tests each driver through the TypeORM spec?

imnotjames:
Has this been tested on iOS?  Will it break iOS' implementation?

pleerock:
@chriswep yes I don't mind a new testing setup to be part of this PR. However we need contributions from anybody in order to fix the issues we have.

chriswep:
> Has this been tested on iOS? Will it break iOS' implementation?

yes it has been tested, no it doesn't break iOS @imnotjames

imnotjames (approved):

PR #7731, opened 2021-06-10
fix(#7730): Default values filter numeric types, and empty strings

### LINKED ISSUE #7730: TypeError: value.toUpperCase
When Default value is not a string, a method that does not exist is called

image

image
image

https://github.com/typeorm/typeorm/blob/ff28aad27958ef7763a305982f70caeb611f2448/src/driver/mysql/MysqlDriver.ts#L946

jinpy666:
### Pull-Request Checklist

- [x] This pull request links relevant issues as `Fixes #7730`

### DIFF
--- a/src/driver/mysql/MysqlDriver.ts
+++ b/src/driver/mysql/MysqlDriver.ts
@@ -940,7 +940,7 @@ export class MysqlDriver implements Driver {
      * Otherwise returns original input.
      */
     protected normalizeDatetimeFunction(value?: string) {
-        if (!value) return value
+        if (!value || typeof value === "number" || value === "''") return value
 
         // check if input is datetime function
         const isDatetimeFunction = value.toUpperCase().indexOf("CURRENT_TIMESTAMP") !== -1


imnotjames (changes requested):
This should instead just check that it's typeof string shouldn't it?  Probably could live with that instead of the number check.  Why do we add the empty string check? 

If this function can possibly accept non-strongs maybe it should be an `any` in the parameter type?

jinpy666:
> This should instead just check that it's typeof string shouldn't it? Probably could live with that instead of the number check. Why do we add the empty string check?

My case is： **default:()=>''**
That's why you need to determine the empty string.
My intention was to block it at the beginning and not have to go through the following process.
Please look at the second anser.

> If this function can possibly accept non-strongs maybe it should be an `any` in the parameter type?
Yes.
https://github.com/typeorm/typeorm/blob/53b33e105d1b0e25b1594c88c4405cacc34ff5bb/src/driver/mysql/MysqlDriver.ts#L604-L607

When the default is function, can always access to normalizeDatetimeFunction, inside is not very good type judgment, lead to syntax error, and through the Google search the default values demo, mostly use the function to input.

imnotjames:
> My case is： default:()=>''
> That's why you need to determine the empty string.

Your code change will not affect this because the value you're looking for is `"''"`, not an empty string (`""`).

What is the issue you're trying to solve for here?

imnotjames (changes requested):
Please also add tests for this

PR #7739, opened 2021-06-14
refactor: allow driver to define which query-builders to use

imnotjames:
### Description of change

this updates typeorm to request a new query builder
from the driver rather than using a hard-coded module.
in many cases we want to return query builders with
a different embedded dialect within them.

this is a tiny step to encapsulating some of the dialect
code out of the QueryBuilders and instead bundling
them together under the drivers

### Pull-Request Checklist

### DIFF
--- a/src/connection/Connection.ts
+++ b/src/connection/Connection.ts
@@ -392,9 +392,6 @@ export class Connection {
      * Executes raw SQL query and returns raw database results.
      */
     async query(query: string, parameters?: any[], queryRunner?: QueryRunner): Promise<any> {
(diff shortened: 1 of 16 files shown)


imnotjames:
This is currently just an idea I had to help with splitting up the drivers a bit more.

There's currently 133 cases where we use specific drivers to determine the behavior of TypeORM.  Half of those are in the `query-builder` because it has lots of `if` checks to do specific SQL depending on the dialect.

This is not the best solution but it's a first step.

I'm mostly curious to see if this builds.

imnotjames:
Closing this for now.

Issue #5365, opened 2020-01-16
Generated Identity


Css-IanM:
**Issue type:**

[ ] question
[ ] bug report
[ x] feature request
[ ] documentation issue

**Database system/driver:**

[ ] `cordova`
[ ] `mongodb`
[ ] `mssql`
[ ] `mysql` / `mariadb`
[ ] `oracle`
[ x] `postgres`
[ ] `cockroachdb`
[ ] `sqlite`
[ ] `sqljs`
[ ] `react-native`
[ ] `expo`

**TypeORM version:**

[ x] `latest`
[ ] `@next`
[ ] `0.x.x` (or put your version here)

**Steps to reproduce or a small repository showing the problem:**

Include Decorator or decorator parameters for primary key identity IE: Postgres 10+ 
```sql
GENERATED { ALWAYS | BY DEFAULT } AS IDENTITY { ( sequence_option ) }
```

mitsos1os:
IMO this should also be the default behavior for `@PrimaryGeneratedColumn()` under PostgresVesion > 10 using an options flag during the connection or something

mitsos1os:
@pleerock @AlexMesser I am willing to issue a non-breaking PR for implementing this feature in a flag-option enabled way.
Would this be something that you would consider accepting as a feature, or should I not deal with it at all?
Thanks!

PR #7716, opened 2021-06-07
fix: prevent dropping materialized views on Postgres versions < 9.3

AlexMesser:
### Description of change

Materialized views introduced in Postgres 9.3. This PR adds version check before dropping materialized views to prevent errors on Postgres versions lower than 9.3

Fixes #7714

### Pull-Request Checklist

- [ ] Code is up-to-date with the `master` branch
- [ ] `npm run lint` passes with this change
- [ ] `npm run test` passes with this change
- [ ] This pull request links relevant issues as `Fixes #0000`
- [ ] There are new or updated unit tests validating the change
- [ ] Documentation has been updated to reflect this change
- [ ] The new commits follow conventions explained in [CONTRIBUTING.md](https://github.com/typeorm/typeorm/blob/master/CONTRIBUTING.md)

PR #7741, opened 2021-06-14
Postgres IDENTITY Column support

mitsos1os:
### Description of change

Fixes: #5365
This PR adds the ability to create **IDENTITY** columns _(for **Postgres 10+!**)_ using the `PrimaryGeneratedColumn` decorator, instead of the default `SERIAL` that was being created so far.

ex:
```typescript
@Entity()
export class User {
  @PrimaryGeneratedColumn("identity")
  id!: number;
}
```

The user must be cautious not to use the `identity` option in the decorator when connecting to a Postgres <10 database because it will throw an error since it is not supported in the DB Engine.

I chose to add `identity` as an extra `generationStrategy` instead of adding an extra option flag, because it directly avoids rest of boilerplate code provided from the `increment` default strategy _(Adding Sequences)_. If extra flag was chosen and `increment` was kept as a `generationStrategy`, extra logic would have to be applied to check the flag's value and not create the **SERIAL** column with its associated **SEQUENCES** and **DEFAULT VALUES** accordingly.

Unit tests have been added that test the produced SQL for creating the column in a table.

However no functional tests added that can load the table from the database and apply proper `TableColumn` options _(although the realted code has been added)_, due to the fact that older Postgres version is used in docker-compose.yml. I could directly upgrade it in the compose file and apply my test, but I am not sure this is a wanted behavior.

Open to suggestions and comments!

Thanks!

### Pull-Request Checklist


imnotjames:
Why not use update things so that `increment` uses the `IDENTITY` syntax when connected to postgres 10?

mitsos1os:
> Why not use update things so that `increment` uses the `IDENTITY` syntax when connected to postgres 10?

Hi! Can you please explain what you mean by saying **update things **? I didn't understand your suggestion

imnotjames:
> > Why not use update things so that `increment` uses the `IDENTITY` syntax when connected to postgres 10?
> 
> Hi! Can you please explain what you mean by saying **update things**? I didn't understand your suggestion

Change the default behavior of `increment` for primary generated column so that it uses the new postgres feature when connected to postgres 10.

Is that possible?

mitsos1os:
@imnotjames From what I have seen there is no information for the version of the Database Server to which the client is connected to. The only thing I found with a quick look is something like that https://github.com/brianc/node-postgres/issues/2002 which would require extra tampering with the embedded driver functionality.

Also I stated in my PR description that I chose a new generation strategy to also circumvent the accompanying SEQUENCE creation that comes with SERIAL column definition.

imnotjames:
> @imnotjames From what I have seen there is no information for the version of the Database Server to which the client is connected to. The only thing I found with a quick look is something like that https://github.com/brianc/node-postgres/issues/2002 which would require extra tampering with the embedded driver functionality.

See https://github.com/typeorm/typeorm/pull/7716

> 
> Also I stated in my PR description that I chose a new generation strategy to also circumvent the accompanying SEQUENCE creation that comes with SERIAL column definition.

It seems that this is the new suggested mechanism for increment instead of serial when supported, right?

I don't think we should add yet another generation strategy that's custom for a specific database.  I also think we should drop `rowid` because cockroachdb seems to use that as the default anyway.

mitsos1os:
Ok thanks! I will take a look.
Just to make sure, will you review this for integration?
What I mean is it would be good to avoid a different opinion from a reviewer that would want to revert this change again

If it is OK, I have no problem changing it to what you suggested

mitsos1os:
OK, I took a look at the version checking, it is straightforward.
How should I proceed with my PR without re-writing the same code for version checking as [here](https://github.com/typeorm/typeorm/pull/7716/files#diff-b924b9a6d4ab4328f4ad59d4d96b794f6a1975fae8d1d3708909e31c151730b9R1928)?

imnotjames:
> Ok thanks! I will take a look.
> Just to make sure, will you review this for integration?
> What I mean is it would be good to avoid a different opinion from a reviewer that would want to revert this change again
> 
> If it is OK, I have no problem changing it to what you suggested

I can review it and I have the ability to merge code.

I can't 100% say that it will be merged in.

We can always leave this PR open, create a new branch, get that code working, and then make sure the tests show things are good.

The only thing I'm unsure about is what will happen with existing tables when this change gets in.  I think the answer is "nothing" - they'll stay the same, but if they are recreated they would get the new code / sequence / etc.  That's probably fine?

mitsos1os:
@imnotjames OK cool!
Regarding the #7716 PR where the `getVersion` is implemented, how do you want to handle this since I will have to use the same/similar util?
Should I also create it in my PR?

Also left a comment regarding my view on the implementation [here](https://github.com/typeorm/typeorm/pull/7716#discussion_r654630750)

imnotjames:
If you need it, write it.  If it gets merged in first we'll update that PR to match.

docs/entities.md
```diff
@@ -524,7 +524,7 @@ export class User {
 
 `uuid` value will be automatically generated and stored into the database.
 
-Besides "uuid" there is also "increment" and "rowid" (CockroachDB only) generated types, however there are some limitations
+Besides "uuid" there is also "increment" "identity" (Postgres 10+ only) and "rowid" (CockroachDB only) generated types, however there are some limitations
```
pleerock, line 527: missing "," mark between `"increment" "identity"`

src/driver/postgres/PostgresQueryRunner.ts
```diff
@@ -930,7 +930,7 @@ export class PostgresQueryRunner extends BaseQueryRunner implements QueryRunner
                 }
             }
 
-            if (oldColumn.isGenerated !== newColumn.isGenerated && newColumn.generationStrategy !== "uuid") {
+            if (oldColumn.isGenerated !== newColumn.isGenerated && !["uuid", "identity"].includes(newColumn.generationStrategy as string)) {
```
pleerock, line 933: Simply this expression. My stupid brain can't accept too tricky expressions 😆

pleerock (changes requested):

mitsos1os:
@pleerock thanks for getting back! I have rebased and addressed the comments on your conversations. Take a look and let me know what you think!
Thanks!

mitsos1os:
Great! Looking forward to it getting merged!

PR #7746, opened 2021-06-16
test: update test to not use `Promise.all` in some cases

imnotjames:
### Description of change

`Promise.all` will execute all promises at once.  Because of the
way that the current code is written these mostly execute
serially, but that's just coincidental.  It's possible for them
to execute in different orders if there are any extra async layers
between the test and the query executor, such as through a
refactoring.

single-connection databases can get out of order which breaks
the insert return code.  because these tests are not meant to
detect race conditions behavior, I've refactored them to avoid
this problem by removing `Promise.all` in some cases.

### Pull-Request Checklist


imnotjames:
It's possible that it could be caused by the `tslib` we're using which shims the async code.

PR #7748, opened 2021-06-16
fix: drop SAP statement after `prepare` per Hana client docs

### LINKED ISSUE #7344: SAP Hana Driver does not drop prepared statement
## Issue Description

After certain amount of db queries, an error occurs due to exceeded maximum number of prepared statements

### Expected Behavior

Single connection should be able to execute infinite number of queries and prepared statements should be closed/dropped when db lib does do it automatically.

### Actual Behavior

Error message obtained
```
Error: exceed maximum number of prepared statements: the number of prepared statements per connection cannot exceed the max statements
```

### Steps to Reproduce

Setup a nestjs with typeorm simple app on SAP hana db, define one entity, and in a loop run inserts for that entity.  This brakes after ~34k of inserts

```typescript
import { InjectRepository } from '@nestjs/typeorm';

  constructor(
    @InjectRepository(Schedule)
    private readonly repo: Repository<Schedule>,
  ) {
  }

for (let i=0; i<=200000; i++) {
      const ent = new Entity({...});
      await this.repo.insert(ent);
}
```

### My Environment

| Operating System    | MacOS BigSur         |
| Node.js version     | v12.20.1 | 
| Typescript version  | v3.8.3 | 
| TypeORM version     | v0.2.24 | 
| @sap/hana-client version | 2.6.58 |
| hdb-pool version | 0.1.4 |
| @nestjs/typeorm version | 7.0.0 |

### Additional Context

`SapQueryRunner.query()` is doing `databaseConnection.prepare()` but not doing `stmt.drop()` in the callback. In my test, after modifying local copy of the file with
```
// first thing in the prepare callback
statement.drop((err) => {
  if (err) {
    _this.driver.connection.logger.logQueryError(err, query, parameters, _this);
    resolveChain();
    return fail(new index_1.QueryFailedError(query, parameters, err));
  }
})
```
Note that patch is done on the compiled js in my local node_modules... Typescript would look different ofc.

I am able to go pass the 34k of inserts w/out the error state above. Let it running all the way up to 200k

Please verify error handling stated above is ok - if I'm gonna do the PR

- [ ] `aurora-data-api`
- [ ] `aurora-data-api-pg`
- [ ] `better-sqlite3`
- [ ] `cockroachdb`
- [ ] `cordova`
- [ ] `expo`
- [ ] `mongodb`
- [ ] `mysql`
- [ ] `nativescript`
- [ ] `oracle`
- [ ] `postgres`
- [ ] `react-native`
- [x] `sap`
- [ ] `sqlite`
- [ ] `sqlite-abstract`
- [ ] `sqljs`
- [ ] `sqlserver`

imnotjames:
### Description of change

each statement takes up a pool so if you query frequently you can
exhaust that pool of statement handles.  this drops the statement
after we're done with it so we don't exhaust the pool of statements

fixes #7344

### Pull-Request Checklist

### DIFF
--- a/src/driver/sap/SapQueryRunner.ts
+++ b/src/driver/sap/SapQueryRunner.ts
@@ -184,82 +184,101 @@ export class SapQueryRunner extends BaseQueryRunner implements QueryRunner {
         }
 
         const promise = new Promise(async (ok, fail) => {
-           try {
-               const databaseConnection = await this.connect();
-               // we disable autocommit because ROLLBACK does not work in autocommit mode
-               databaseConnection.setAutoCommit(!this.isTransactionActive);
-               this.driver.connection.logger.logQuery(query, parameters, this);
-               const queryStartTime = +new Date();
-               const isInsertQuery = query.substr(0, 11) === "INSERT INTO";
+            const resolveChain = () => {
+                let promiseIndex = this.queryResponsibilityChain.indexOf(promise);
+                let waitingPromiseIndex = this.queryResponsibilityChain.indexOf(waitingPromise);
 
-               const statement = databaseConnection.prepare(query);
-               statement.exec(parameters, (err: any, raw: any) => {
+                if (promiseIndex !== -1)
+                    this.queryResponsibilityChain.splice(promiseIndex, 1);
+                if (waitingPromiseIndex !== -1)
+                    this.queryResponsibilityChain.splice(waitingPromiseIndex, 1);
 
-                   // log slow queries if maxQueryExecution time is set
-                   const maxQueryExecutionTime = this.driver.options.maxQueryExecutionTime;
-                   const queryEndTime = +new Date();
-                   const queryExecutionTime = queryEndTime - queryStartTime;
-                   if (maxQueryExecutionTime && queryExecutionTime > maxQueryExecutionTime)
-                       this.driver.connection.logger.logQuerySlow(queryExecutionTime, query, parameters, this);
+                waitingOkay();
+            };
 
-                   const resolveChain = () => {
-                       if (promiseIndex !== -1)
-                           this.queryResponsibilityChain.splice(promiseIndex, 1);
-                       if (waitingPromiseIndex !== -1)
-                           this.queryResponsibilityChain.splice(waitingPromiseIndex, 1);
-                       waitingOkay();
-                   };
+            let statement: any;
 
-                   let promiseIndex = this.queryResponsibilityChain.indexOf(promise);
-                   let waitingPromiseIndex = this.queryResponsibilityChain.indexOf(waitingPromise);
-                   if (err) {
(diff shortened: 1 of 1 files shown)


imnotjames:
unfortunately, we still don't test SAP Hana.  I can barely get a server running to execute tests.  Based off the documentation this seems reasonable, though?

pleerock:
Any additional test infrustructure is highly appreciated 🤞

imnotjames:
SAP Hana is very, very hard to get testing because of the licensing around Hana.  It used to be easier but SAP restricted it further.

It now requires an account that has "purchased" the development kit for free in docker hub.  This can be done but it requires attention any time a new version is released.

It also caused the CI runner to OOM in my tests.

src/driver/sap/SapQueryRunner.ts
```diff
@@ -237,7 +239,9 @@ export class SapQueryRunner extends BaseQueryRunner implements QueryRunner {
                });
            } catch (err) {
                 fail(err);
-            }
+           } finally {
+               statement?.drop();
```
tmilos, line 243: This might execute before callback where data is read is being called, since exec is not awaited since its callback based. I would put drop in the callback

src/driver/sap/SapQueryRunner.ts
```diff
@@ -190,7 +192,7 @@ export class SapQueryRunner extends BaseQueryRunner implements QueryRunner {
                const queryStartTime = +new Date();
                const isInsertQuery = query.substr(0, 11) === "INSERT INTO";
 
-               const statement = databaseConnection.prepare(query);
+               statement = databaseConnection.prepare(query);
                statement.exec(parameters, (err: any, result: any) => {
 
                    // log slow queries if maxQueryExecution time is set
```
tmilos, line 198: I tested it with drop here and it worked

src/driver/sap/SapQueryRunner.ts
imnotjames: Good thinking. :)

imnotjames: I've reworked it a little bit.  Let me know what you think!  I tried to reduce code duplication.

To be fair... this entire query method needs to be rewritten, it's a bit difficult to work with...

src/driver/sap/SapQueryRunner.ts
imnotjames: `drop` is async as well?  Ouch..

src/driver/sap/SapQueryRunner.ts
```diff
@@ -237,8 +249,8 @@ export class SapQueryRunner extends BaseQueryRunner implements QueryRunner {
                });
            } catch (err) {
                 fail(err);
-            }
-        });
+           }
+        }).then(dropStatement, dropStatement)
```
imnotjames, line 253: Wait, this will swallow up any exceptions, won't it..?

Issue #6142, opened 2020-05-27
OrmUtils in 0.2.25 breaks browser builds due to node import


chriswep:
**Issue type:**

[ ] question
[x] bug report
[ ] feature request
[ ] documentation issue
**TypeORM version:**

[x] `latest`
[ ] `@next`
[ ] `0.x.x` (or put your version here)

**Steps to reproduce or a small repository showing the problem:**

using the `.save()` method (sometimes?) produces the error ` TypeError: Right hand side of instanceof is not an object` which is triggered by `OrmUtils.compare2Objects`. Since https://github.com/typeorm/typeorm/commit/50a0641 this uses the node-import `url` which is not available on browser targets. I guess this needs to be removed or polyfilled in `browser/utils/Ormutils`.

allandiego:
Having the same error:

The package at "node_modules\typeorm\browser\util\OrmUtils.js" attempted to import the Node standard library module "url". It failed because the native React runtime does not include the Node standard library.

pleerock:
fixed by #6145

imnotjames:
This is indeed fixed & has been released.  Bummer that we had to revert it. ):

PR #7761, opened 2021-06-17
Remove "properly loads config from yaml" as an exclusive test run

dhritzkiv:
Remove "properly loads config from yaml" as an exclusive test run

### Pull-Request Checklist

- [x] Code is up-to-date with the `master` branch
- [ ] `npm run lint` passes with this change
- [ ] `npm run test` passes with this change
- [ ] This pull request links relevant issues as `Fixes #0000`
- [ ] There are new or updated unit tests validating the change
- [ ] Documentation has been updated to reflect this change
- [ ] The new commits follow conventions explained in [CONTRIBUTING.md](https://github.com/typeorm/typeorm/blob/master/CONTRIBUTING.md)

PR #7762, opened 2021-06-17
fix: handle URL objects as column field values (redux)

### LINKED ISSUE #5762: Using URL as a rich column type breaks

dhritzkiv:
### Description of change

Fixes #5762
See previous PR (which was reverted) #5771 due to #6142

This time, `URL` isn't imported from the `url` module (which isn't present in React Native or browser environments). Instead, it's optionally accessed on the globalThis (if present) object, which is supported in node v.10+ and modern browser environments.

### Pull-Request Checklist

### DIFF
--- a/src/util/OrmUtils.ts
+++ b/src/util/OrmUtils.ts
@@ -1,4 +1,5 @@
 import {ObjectLiteral} from "../common/ObjectLiteral";
+import {PlatformTools} from "../platform/PlatformTools";
 
 /**
  * Access `URL` in environments that support it.
@@ -7,7 +8,8 @@ import {ObjectLiteral} from "../common/ObjectLiteral";
  *
  * Note: `globalThis` is being extended with a `URL` property as @types/node does not declare it (see https://github.com/DefinitelyTyped/DefinitelyTyped/issues/34960)
  */
-const GlobalURL = (typeof globalThis !== "undefined" ? globalThis as (typeof globalThis & {URL?: (typeof import("url").URL)}) : {URL: undefined}).URL;
+const _global = PlatformTools.getGlobalVariable() as (typeof globalThis & {URL?: (typeof import("url").URL)});
+const GlobalURL = (typeof _global !== "undefined" ? _global : {URL: undefined}).URL;
 
 export class OrmUtils {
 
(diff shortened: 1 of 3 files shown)


dhritzkiv:
Note: the relevant tests won't be run on CI until #7761 is merged

imnotjames:
Would something like https://github.com/typeorm/typeorm/pull/5097 fix this & other cases where object are passed?

dhritzkiv:
> Would something like #5097 fix this & other cases where object are passed?

Hmm. Could be. Looking through the code, `isPlainObject` should return `false` for an instance of `URL`, since `Object.prototype.toString.call(url)` would return `'[object URL]'`.

I'll pull that branch down and add the tests from this PR

dhritzkiv:
I think that branch is so outdated that tests won't run properly. :(

imnotjames:
I believe this should be fixed on the `master` branch now.  I've included your tests.

I'll be closing this PR.

PR #7767, opened 2021-06-19
feat: Error if postgresql enum is not specified

michaelzangl:
### Description of change

Print a nice error message if the enum value is missing for postgresql, including the column name that has the problem.

Message:
```
No enum values provided for column my_column of table my_table. Please provide the enum values using @Column({enum: ...})
```

Took me some time to figure out which column was causing this, since there is no error message currently and typeorm just crashes with an undefined object reference.

### Pull-Request Checklist

- -- This pull request links relevant issues as `Fixes #0000`
- -- Documentation has been updated to reflect this change

### DIFF
--- a/src/driver/postgres/PostgresQueryRunner.ts
+++ b/src/driver/postgres/PostgresQueryRunner.ts
@@ -2027,7 +2027,10 @@ export class PostgresQueryRunner extends BaseQueryRunner implements QueryRunner 
      */
     protected createEnumTypeSql(table: Table, column: TableColumn, enumName?: string): Query {
         if (!enumName) enumName = this.buildEnumName(table, column);
-        const enumValues = column.enum!.map(value => `'${value.replace("'", "''")}'`).join(", ");
+        if (!column.enum) {
+            throw new Error(`No enum values provided for column ${column.name} of table ${table.name}. Please provide the enum values using @Column({enum: ...})`);
+        }
+        const enumValues = column.enum.map(value => `'${value.replace("'", "''")}'`).join(", ");
         return new Query(`CREATE TYPE ${enumName} AS ENUM(${enumValues})`);
     }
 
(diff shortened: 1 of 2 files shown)


imnotjames (commented):
Is it possible to omit the enums if they haven't been defined in postgres?

michaelzangl:
@imnotjames What do you mean by omitting them? If you want to define a column as postgres enum type, you need to provide the enum values somewhere.

imnotjames:
Except you didn't - what happens if you create the type but omit all possible values?  Is the type still valid?

michaelzangl:
> Except you didn't - what happens if you create the type but omit all possible values? Is the type still valid?

Yes, it is. But you cannot use the column for anything useful → this is usually a programming mistake. When using an implicit empty enum values array, this issue is hard to track.

If some really wants to use an empty enum, one can use:
```
@Column({enum: []})
```

imnotjames:
Looking at how other drivers implement enums, they omit the values of the enums isn't defined.

https://github.com/typeorm/typeorm/blob/master/src/driver/mysql/MysqlQueryRunner.ts#L1862

https://github.com/typeorm/typeorm/blob/master/src/driver/sqlserver/SqlServerQueryRunner.ts#L2139

Why should postgres operate differently?

michaelzangl:
MySQL und SqlServer don't set the enum values to an empty array.

They use a text (varchar, ....) column with a constraint to model enums. If an enum is not specified, they don't add the enum constraint. That means, they allow all textual values for that column.

Postgres has a special enum column type. In that case, only whitelisted values are allowed by postgres. This is why postgres requires you to provide an array of allowed values. And setting this array to be an empty default is not the same as just allowing all values (as mysql/sqlserver do).

imnotjames:
Then we should emit an error when building metadata when the type is set to
an enum, not here, during the database schema builder much later (if at
all) in the process.

pleerock:
I think its better to add validation checks inside `EntityMetadataValidator` and make sure to cover all database drivers.

michaelzangl:
Feel free to do it - I just wanted to fix that error message of the NPE to something menaingful, which is a two line change (+ tests). Changing validation is much bigger and since I am not that deep into TypeORM, I won't be able to do it.

Issue #7135, opened 2020-12-02
Declare Driver Dependencies in package.json


brandon-leapyear:
## Issue Description

Yarn v2 uses Plug-n-Play, which means that it's starting to enforce packages needing to declare their dependencies. `typeorm` breaks because `pg` is declared as a dev dependency, but it's required at runtime.

https://github.com/typeorm/typeorm/blob/master/src/driver/postgres/PostgresDriver.ts#L956
https://github.com/typeorm/typeorm/blob/master/src/platform/PlatformTools.ts#L78

The workaround is to add the following to `.yarnrc.yml`:
```yaml
packageExtensions:
  'typeorm@*':
    peerDependencies:
      'pg': '*'
```

But the better solution would be to list `pg` (and anything else typeorm needs to require) as a `peerDependency`.

### Expected Behavior

Running `yarn node dist/main.js` works, after compiling my typescript code using typeorm.

### Actual Behavior

Running our NestJS server gives us

```
[Nest] 2260   - 2020-12-02 1:34:48 AM   [TypeOrmModule] Unable to connect to the database. Retrying (1)... +4ms
DriverPackageNotInstalledError: Postgres package has not been found installed. Try to install it: npm install pg --save
    at new DriverPackageNotInstalledError (/root/src/~/.yarn_cache/typeorm-npm-0.2.29-c30de22860-c70f986138.zip/node_modules/typeorm/error/DriverPackageNotInstalledError.js:10:28)
    at PostgresDriver.loadDependencies (/root/src/~/.yarn_cache/typeorm-npm-0.2.29-c30de22860-c70f986138.zip/node_modules/typeorm/driver/postgres/PostgresDriver.js:966:19)
    at new PostgresDriver (/root/src/~/.yarn_cache/typeorm-npm-0.2.29-c30de22860-c70f986138.zip/node_modules/typeorm/driver/postgres/PostgresDriver.js:202:14)
    at DriverFactory.create (/root/src/~/.yarn_cache/typeorm-npm-0.2.29-c30de22860-c70f986138.zip/node_modules/typeorm/driver/DriverFactory.js:35:24)
    at new Connection (/root/src/~/.yarn_cache/typeorm-npm-0.2.29-c30de22860-c70f986138.zip/node_modules/typeorm/connection/Connection.js:53:59)
[...]

PR #7768, opened 2021-06-19
Add peer dependencies

brandonchinn178:
### Description of change

Resolves https://github.com/typeorm/typeorm/issues/7135

Added a peer dependency for every call to `PlatformTools.load()` mentioning a dependency not declared in `dependencies` (which should be all of them).

### Pull-Request Checklist

### DIFF
--- a/package.json
+++ b/package.json
@@ -114,6 +114,78 @@
     "typeorm-aurora-data-api-driver": "^2.0.0",
     "typescript": "^4.2.2"
   },
+  "peerDependencies": {
+    "@sap/hana-client": "*",
+    "better-sqlite3": "*",
+    "hdb-pool": "*",
+    "ioredis": "*",
+    "ioredis": "*",
+    "ioredis/cluster": "*",
+    "mongodb": "*",
+    "mssql": "*",
+    "mysql2": "*",
+    "oracledb": "*",
+    "pg": "*",
+    "pg-native": "*",
+    "pg-query-stream": "*",
+    "redis": "*",
+    "sql.js": "*",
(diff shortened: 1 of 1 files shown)


imnotjames:
What npm version did this feature start

brandonchinn178:
Which feature are you referring to?

`peerDependencies` always existed. `npm` v3-6 didn't install these by default, but v7 does install them by default, unless it's specified as `optional` in `peerDependenciesMeta`, which was added in v7

imnotjames:
Got it.

Can you verify with an older version or find the documentation of npm to confirm that?

The issue before was that it ensured they were installed I'd thought.

Just making sure we understand the implications of setting this.

brandonchinn178:
The v7 docs talk about v3-6: https://docs.npmjs.com/cli/v7/configuring-npm/package-json#peerdependencies

You can also change the version at the top of the page to v6, which says it doesnt install them automatically

imnotjames:
Got it.  I'll verify that behavior with some older versions, then we'll merge it.

imnotjames:
In v6, there's a big warning that says to install the packages.  That's not great but not a show-stopper.  It will very likely lead to some issues being opened erroneously, though.

Given that v6 is older, now, I think that's acceptable - though maybe a little annoying for folks stuck with that version.

This also is a huge benefit because we can start hinting the right versions to use.  Yay!

imnotjames (approved):

imnotjames:
Commits should match the `CONTRIBUTING.md` but I'll handle this as part of the squash.

PR #7791, opened 2021-06-23
Postgres IDENTITY Column support #2

mitsos1os:
### Description of change

Fixes: #5365

### Also includes similar functionality to #7716 regarding **server version**

This PR adds the ability to create **IDENTITY** columns _(for **Postgres 10+!**)_ instead of the default `SERIAL` that was being created so far.

It is still based on `PrimaryGeneratedColumn("increment")`. A check is performed against the database version and proper functionality is selected according to IDENTITY Column support or not from the database server.

### Pull-Request Checklist

- [] `npm run lint` passes with this change
- [] `npm run test` passes with this change
- [] There are new or updated unit tests validating the change
- [] Documentation has been updated to reflect this change


imnotjames (changes requested):
Can you create tests that verify this behavior?

mitsos1os:
> Can you create tests that verify this behavior?

you beat me on sending the comment :smile: 

I just created this PR according to our previous conversation in #7741

However, I wanted you to see the problem that arises regarding changing existing schema operations and why I chose to introduce a new generation strategy `identity` in my original PR.

If you can take a look at the **PostgresQueryRunner** wherever `this.driver.supportsIdentityColumn()` is used in `renameTable` and `changeColumn`, there will be a bit of a mess if a change is made on a schema running in a Postgre 10+ version. The old creations would be based on sequences while the new ones will be based on IDENTITY and vice versa for the reverse queries.

Also I guess this will also lead to missed sequence numbers for the new columns that won't be able to resume from last sequence.

The only way to bypass this would be to introduce if column has a sequence attached and then apply corrective queries etc... which I believe will increase the complexity a lot!

Please take a look and let me know what you think.

Thanks

mitsos1os:
> Can you create tests that verify this behavior?

Hi @imnotjames ! Did you by any chance have a look at my response?

Sorry to be pressuring, but it would be great if we could reach a conclusion and merge an implementation of this feature, because we are currently using it by monkey patching the code in the running application!

Thanks for your time!

imnotjames:
> The old creations would be based on sequences while the new ones will be based on IDENTITY and vice versa for the reverse queries.

Isn't this what we want?  To use the sequence based generation when feasible.

> Also I guess this will also lead to missed sequence numbers for the new columns that won't be able to resume from last sequence number used.

Not sure I understand.  How will it miss sequence numbers?

mitsos1os:
Let me explain what I meant with an example.

Lets's say we are running **typeorm** against a PostgreSQL Server **version 10+**, using the functionality so far that handles `SERIAL` and `SEQUENCES` for **increment** columns

Without changing anything else on the server, we integrate the new functionality for `IDENTITY` columns, which based on server version are supported. Let's see some example operations:

**1. Regular usage**
During regular usage just upgrading the library will leave the old `SERIAL` columns unchanged without migrating to `IDENTITY`. This is due to the fact that all checks are valid in `findChangedColumns` [here](https://github.com/typeorm/typeorm/blob/f547dd7f8004d4a003c351f43c7bb29381dc6567/src/driver/postgres/PostgresDriver.ts#L919), since we are not actually chaning any column configuration

**2. Rename table**
If we rename the table containing the `SERIAL` column then the sequence name will be left matching the old table name due to the check (here)[https://github.com/typeorm/typeorm/blob/f547dd7f8004d4a003c351f43c7bb29381dc6567/src/driver/postgres/PostgresQueryRunner.ts#L508] where since the `IDENTITY` column is supported, no sequence change willl be run. So sequence name will not follow new table name.

**3. Rename `SERIAL` column**
This is the same case as the **rename table** operation, but now this is due to the code [here](https://github.com/typeorm/typeorm/blob/f547dd7f8004d4a003c351f43c7bb29381dc6567/src/driver/postgres/PostgresQueryRunner.ts#L729). Again sequence name will be left not matching new column name but new column will continue to use old sequence for default value

**4. Change `SERIAL` column to other non-generated type**
As seen [here](https://github.com/typeorm/typeorm/blob/f547dd7f8004d4a003c351f43c7bb29381dc6567/src/driver/postgres/PostgresQueryRunner.ts#L933), if an old `SERIAL` column is changed to a **non-generated** type, then the old accompanying sequence will be left intact and also the default value of the column won't be unassigned from the previous attached SEQUENCE

These are some of the problematic cases I can think of.

Also by this:

> Also I guess this will also lead to missed sequence numbers for the new columns that won't be able to resume from last sequence number used.

I meant that if somehow (I can't think of a way right now), an old `SERIAL` column is marked as being updated to `IDENTITY` and the last number used from the accompanying SEQUENCE was 5, we cannot easilty port this to the new IDENTITY configuration so that next inserted value will get the number 6

I hope my concerns were made clearer now.
What are your thoughs on these?

pleerock:
@mitsos1os I'm sorry for delayed review from my side. I checked both PRs and even through we don't need extra `identity` type - it's safer approach we should go with.

And yeah, sorry about it:

> What I mean is it would be good to avoid a different opinion from a reviewer that would want to revert this change again

pleerock:
So I'm going to close this PR in favour of #7741 - I reviewed it - almost good to be merged.

PR #7600, opened 2021-04-28
Upgrade version of mssql to fix a bug of date with mssql driver (tedious)

v1d3rm3:
### Description of change

Hello! I'm sending a PR to solve a problem in Date, as referenced here https://github.com/tediousjs/tedious/issues/1249.  Unfortanalety, it'll not be a version to solve the problem in `mssql v6`, so i'm sending this upgrade for `next´ branch, because i'm using this version in my job a long time motivated by this issue (https://github.com/typeorm/typeorm/issues/4270).

In PR, i did the following things (sorry for not separate them). First i upgraded mssql version to v7. Then i correct the way to send name parameters to mssql (tedious aparently not accept name as numbers, only strings).

### Pull-Request Checklist

- [x] (IT FOR NEXT BRANCH) Code is up-to-date with the `master` branch
- [x] `npm run lint` passes with this change
- [x] `npm run test` passes with this change (partially)
- [x] This pull request links relevant issues as `Fixes #0000`
- [x] There are new or updated unit tests validating the change
- [x] Documentation has been updated to reflect this change
- [x] The new commits follow conventions explained in [CONTRIBUTING.md](https://github.com/typeorm/typeorm/blob/master/CONTRIBUTING.md)

PR #7796, opened 2021-06-25
Fixing problem in date parser

v1d3rm3:
### Description of change

new Date(ISOString) is not a reliable parser to date strings.
 It's better to use 'date-fns' parser to parser string in ISO Format.

The problem here is with wrong timezone.

For example:

``new Date('2021-04-28')`` will generate `2021-04-28T00:00:00.000Z`
 in my timezone, which is not true for my timezone (GMT-0300). It should
 be `2021-04-28T03:00:00.000Z` as `new Date(2021, 3, 28)` generates.

https://stackoverflow.com/a/2587398

origin from #7600 

### Pull-Request Checklist


v1d3rm3:
The test on NodeV10 is falling cause of "Segmentation fault (core dumped)"

pleerock:
I don't understand the problem. From my understanding, `new Date('2021-04-28')` creates a Date object and it contains your local timezone. It's just when you use `.toISOString()` it removes that:

```ts
console.log(new Date('2021-04-28'))
console.log(new Date('2021-04-28').toString())
console.log(new Date('2021-04-28').toISOString())
```

Output (for me) is:

```shell
2021-04-28T00:00:00.000Z
Wed Apr 28 2021 05:00:00 GMT+0500 (Tajikistan Time)
2021-04-28T00:00:00.000Z
```

Now I don't understand what's your problem? There is a timezone inside your data object. Maybe if you provide usage examples I can understand?

v1d3rm3:
> I don't understand the problem. From my understanding, `new Date('2021-04-28')` creates a Date object and it contains your local timezone. It's just when you use `.toISOString()` it removes that:


I think it's not true. In your example, you've `GMT+0500 (Tajikistan Time)`. So, when you use ISO String to create Date object (`new Date('2021-04-28')`), it isn't giving back the timezone in your example: `2021-04-28T00:00:00.000Z`. So, it was expected that `new Date('2021-04-28')` give us back `2021-04-28T05:00:00.000Z`, not `2021-04-28T00:00:00.000Z`.

pleerock:
Okay I think I understood the difference. This would be a huge breaking change. We'll need to fix all other related date issues in TypeORM and release it in `0.3.0`.

I'll keep this PR opened for `0.3.0`.

pleerock:
Due to conflicts manually pushed your changes in 7e625b86b09c9ed675e6bfc7a16fecf589b0e2d2 . Will be released in `0.3.0`.

PR #7799, opened 2021-06-25
refactor: use TypeORMError instead of base Error

### LINKED ISSUE #7348: Add parent class for all errors
Is there possibility/sense to add one parent class for all errors?

Now all errors extend standard JS `Error`
https://github.com/typeorm/typeorm/tree/master/src/error
`export class EntityColumnNotFound extends Error {`

I see it `export class EntityColumnNotFound extends MyParentError {` and then `export class MyParentError extends Error {`

It will handy to catch/filter/detect errors from DB. And do something, for example show one common message for users about DB error, without any details.

For example, Sequelize has `DatabaseError` class https://sequelize.org/master/class/lib/errors/database-error.js~DatabaseError.html

Thanks.

imnotjames:
### Description of change

this means you can catch any TypeORM issues by catching the
TypeORM error type

fixes #7348

### Pull-Request Checklist


Shiroh1ge:
Is there a way to track when we'll get a release with this change?

PR #7821, opened 2021-06-30
refactor: use traditional parameters where possible

imnotjames:
### Description of change

this refactors a wide variety of use-cases for parameters.

* drops the native parameters out of use but leaves them in place for compatibility purposes
* refactors the parameter replacement to improve performance drastically
* allows for nested querybuilders without parameter overwriting (such as with `Brackets`
* aligns the escape parameters functions across drivers

### Pull-Request Checklist


src/query-builder/QueryBuilder.ts
```diff
@@ -72,6 +72,8 @@ export abstract class QueryBuilder<Entity> {
      */
     protected parentQueryBuilder: QueryBuilder<any>;
 
+    private parameterIndex = 0;
```
imnotjames, line 75: add comment explaining what this is

imnotjames:
I want to do some back of the napkin benchmarking before merging this in..

pleerock:
> but leaves them in place for compatibility purposes

can you please describe what compatibility purposes left?

imnotjames:
> > but leaves them in place for compatibility purposes
> 
> can you please describe what compatibility purposes left?

It's a public API and could be used by third parties.

imnotjames:
> I want to do some back of the napkin benchmarking before merging this in..

Back of the napkin, no negligible difference.  Timing was incredibly hard to see a difference over 10m inserts in batches of 100 & 10,000

imnotjames:
PS without the changes to the way we're doing replacement escapes, the difference was that for batches of 100 ~200x slower, 1,000 it never finished, 10,000 it crashed because of too many matchers.

PR #7831, opened 2021-07-03
fix: for mysql use boolean type alias for boolean type

### LINKED ISSUE #3622: TypeORM booleans use tinyint (4) instead of the standard tinyint (1)
**Issue type:** [x] bug report
**Database system/driver:** [x] `mysql` / `mariadb`
**TypeORM version:** [x] `latest`

Create a entity modal with a boolean type and use it in MySQL.
```ts
@Entity()
export class Foo {
  @PrimaryGeneratedColumn()
  id: number;

  @Column()
  isBar: boolean = false;
}
```

When you sync with the database you'll find that the isBar column is equivalent to `tinyint (4)` because TypeORM uses `tinyint` as the type without a size.

However, the standard for booleans in MySQL is `tinyint (1)`. MySQL has an official `BOOL`/`BOOLEAN` alias type, which is an alias to `tinyint (1)` not `tinyint (4)`. And other ORMs properly use `tinyint (1)` for their booleans, e.g. I'm migrating a Doctrine based project to TypeORM and need to manually specify the widths of my booleans to avoid counterproductive migrations being created.

imnotjames:
### Description of change

the boolean type in typeorm currently uses a tinyint(4)
when instead we could be using the mysql alias boolean
which is a tinyint(1)

fixes #3622

### Pull-Request Checklist

### DIFF
--- a/src/driver/mysql/MysqlDriver.ts
+++ b/src/driver/mysql/MysqlDriver.ts
@@ -167,6 +167,7 @@ export class MysqlDriver implements Driver {
      * Gets list of column data types that support length by a driver.
      */
     withWidthColumnTypes: ColumnType[] = [
+        "boolean",
         "bit",
         "tinyint",
         "smallint",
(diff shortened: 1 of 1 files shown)


imnotjames:
Right - this doesn't work because tinyint is what it gets created as.. because alias.  Oof...

PR #7839, opened 2021-07-05
refactor: use a `Map` for ConnectionManager to improve lookups

### LINKED ISSUE #7492: Change connections param from array to object
## Feature Description

Good morning,
I think to improve the performance with multiple connections you could change the _connections_ param in _ConnectionManager_ from array to object with key is the _name_ connection

### The Problem

I use typeorm with a lot of multiple connections and the _getConnection_ isn't very performant.

### The Solution (js)

`this.connections = {}`

when create a connection:

`this.connections['nameConnection'] = Connection`

the result will be (for example with _default_ connection and _log_ connection):

`this.connections = {
  default: Connection,
  log: Connection,
}`

when use for example get connection the performance it will be better (there is anymore find):

`return this.connections[nameConnection]`

### Additional Context

no additional context

all driver

no, I never used typescript

- [x] No, I don't have the time and I wouldn't even know how to start.

Thank you so much
Chiara

imnotjames:
### Description of change

closes #7492

### Pull-Request Checklist

### DIFF
--- a/src/connection/ConnectionManager.ts
+++ b/src/connection/ConnectionManager.ts
@@ -8,15 +8,19 @@ import {AlreadyHasActiveConnectionError} from "../error/AlreadyHasActiveConnecti
  * It also provides useful factory methods to simplify connection creation.
  */
 export class ConnectionManager {
-
-    // -------------------------------------------------------------------------
-    // Protected Properties
-    // -------------------------------------------------------------------------
-
     /**
      * List of connections registered in this connection manager.
+     *
+     * @deprecated
      */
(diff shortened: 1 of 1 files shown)


pc3b3r:
@imnotjames I guess it's better to keep the possibility to retrieve all connections available (as it was before this PR using the public readonly `connections`).

imnotjames:
> @imnotjames I guess it's better to keep the possibility to retrieve all connections available (as it was before this PR using the public readonly `connections`).

Can you elaborate the issue here

PR #7868, opened 2021-07-08
Capacitor driver additional connection options

chriswep:
- add additional connection options that the are available for capacitor-sqlite: encryption, mode and version
- adds check for correct values in the journalMode connection option


src/driver/capacitor/CapacitorDriver.ts
```diff
@@ -74,22 +74,33 @@ export class CapacitorDriver extends AbstractSqliteDriver {
...
-
-        if (this.options.journalMode) {
-            await connection.query(`PRAGMA journal_mode = ?`, [
-                this.options.journalMode,
-            ]);
+        await connection.query(`PRAGMA foreign_keys = ON`);
+
+        if (
+            this.options.journalMode &&
+            ["DELETE", "TRUNCATE", "PERSIST", "MEMORY", "WAL", "OFF"].indexOf(
```
imnotjames, line 97: The types already enforce this.

src/driver/capacitor/CapacitorDriver.ts
```diff
@@ -74,22 +74,33 @@ export class CapacitorDriver extends AbstractSqliteDriver {
...
+                this.options.journalMode
+            ) !== -1
+        ) {
+            await connection.query(
+                `PRAGMA journal_mode = ${this.options.journalMode}`
```
imnotjames, line 102: Why aren't we using parameters here?

src/driver/capacitor/CapacitorConnectionOptions.ts
```diff
@@ -9,15 +9,30 @@ export interface CapacitorConnectionOptions extends BaseConnectionOptions {
...
      * Database name (capacitor-sqlite will add the suffix `SQLite.db`)
      */
     readonly database: string;
 
     /**
-     * The capacitor-sqlite instance. For example, `new SQLiteConnection(CapacitorSQLite)`.
+     * Set to true for database encryption
      */
-    readonly driver: any;
+    readonly encrypted?: boolean;
```
imnotjames, line 25: Can't this just be `databaseMode !== undefined && databaseMode !== "no-encryption"`

src/driver/capacitor/CapacitorDriver.ts
chriswep: sqlite unfortunately does not allow parameters in PRAGMA queries/requests (that's what i tried before)

src/driver/capacitor/CapacitorDriver.ts
chriswep: yes. however since we can't use parameters, this makes the query safe while having the plus of preventing wrong values (in case somebody does not use the type system)

src/driver/capacitor/CapacitorConnectionOptions.ts
chriswep: i get what you mean. this mirrors the options that are available for capacitor-sqlite. i thought as well that they are redundant. the thinking is that in the end its better to stay close to the underlying driver.

imnotjames: I feel like that's just going to add extra confusion, isn't it?  

Setting `encrypted` to false just sets it to `no-encryption`.

https://github.com/capacitor-community/sqlite/blob/master/android/src/main/java/com/getcapacitor/community/database/sqlite/CapacitorSQLitePlugin.java#L168

I get that you're trying to mirror it but we're also trying to give a better API on top of capacitor, aren't we?

src/driver/capacitor/CapacitorDriver.ts
imnotjames: Got it - I see that in the sqlite stuff as well.

Can you move this change to a separate PR so we can have a better changelog & commit history?

src/driver/capacitor/CapacitorConnectionOptions.ts
chriswep: good catch from the capacitor-sqlite source. it's ok for me to remove the `encrypted` property. i pushed an update.

src/driver/capacitor/CapacitorDriver.ts
chriswep: done: https://github.com/typeorm/typeorm/pull/7873

imnotjames (approved):

PR #7908, opened 2021-07-15
fix: mongo replica set

klassm:
From typeorm > 0.2.28 parsing the connection string broke our database setup. In our setup we use a replicaSetup - however the required parameters tls and replicaSet were never added to the connection string.

This PR will add the two parameters, in addition to a new test that parses the connection string (as I was looking for how the parsing works).

There is no new test for the actual adding of the two query parameters. If you can point me towards an example test case, I can also provide one for this case.

### Pull-Request Checklist

### DIFF
--- a/src/driver/mongodb/MongoDriver.ts
+++ b/src/driver/mongodb/MongoDriver.ts
@@ -448,15 +448,16 @@ export class MongoDriver implements Driver {
             ? `${options.username}:${options.password}@`
             : "";
 
-        let connectionString = undefined;
+
         const portUrlPart = (schemaUrlPart === "mongodb+srv")
             ? ""
             : `:${options.port || "27017"}`;
 
+        let connectionString: string;
         if(options.replicaSet) {
-            connectionString = `${schemaUrlPart}://${credentialsUrlPart}${options.hostReplicaSet || options.host + portUrlPart || "127.0.0.1" + portUrlPart}/${options.database || ""}`;
+            connectionString = `${schemaUrlPart}://${credentialsUrlPart}${options.hostReplicaSet || options.host + portUrlPart || "127.0.0.1" + portUrlPart}/${options.database || ""}?replicaSet=${options.replicaSet}${options.tls ? "&tls=true" : ""}`;
         } else {
-            connectionString = `${schemaUrlPart}://${credentialsUrlPart}${options.host || "127.0.0.1"}${portUrlPart}/${options.database || ""}`;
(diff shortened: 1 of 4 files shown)


test/utils/test-utils.ts
```diff
@@ -42,6 +42,11 @@ export interface TestingOptions {
      */
     name?: string;
 
+    /**
+     * Url for the database connection.
+     */
+    url?: string;
```
imnotjames, line 48: Why this change?

klassm: This change was more about being able to test if. If you want we can also revert it - but generally I tried it out by modifying the `ormconfig.json`, which required me to allow setting a url here.

imnotjames: I get that it's for testing, but no test you added uses this.

We technically can't, either, because the mongo server we set up won't have many features related to multiple instances enabled.

klassm: Yes exactly. If you want I can also remove it again :-)

klassm:
I removed the corresponding commit, so the offending line is gone again :-)

test/utils/test-utils.ts
klassm: I removed the corresponding commit, so the offending line is gone again :-)

src/driver/mongodb/MongoDriver.ts
```diff
@@ -442,21 +442,22 @@ export class MongoDriver implements Driver {
     /**
      * Builds connection url that is passed to underlying driver to perform connection to the mongodb database.
      */
-    protected buildConnectionUrl(options: { [key: string]: any }): string {
+    static buildConnectionUrl(options: { [key: string]: any }): string {
```
imnotjames, line 445: Can this still be protected even if it's static?  I guess not because that's how you're confirming this with tests?

Is there any way to test that this change works WITHOUT exposing more functions as part of a public API?

klassm: Yeah we thought about this two. The unfortunate part is that the one public method using this is the connect method - so then we'd have to mock the world to get a test. But I wanted to have a test 🙈

klassm: Ok now I rewrote the test a bit - don't know whether it's really pretty now. But what it does now is mock the connect function and emulate what the mongodriver does, so we can actually catch the connection url and verify that it is correct. A bit hard to read, but that way there is no additional public method.

imnotjames (approved):

PR #7909, opened 2021-07-15
feat: update mongodb to major version 4

klassm:
Mongo released a new major version of mongodb, which should (tm) be backwards compatible with older versions, but is essentially a whole rewrite of the library in typescript. You can find the release notes [here](https://github.com/mongodb/node-mongodb-native/releases/).

The thing is: With npm dependencies and your own project upgrading to the new major version, typeorm will crash, as the API slightly changed. I guess typeorm should use the latest version, to here's the update to make it work again.

### Pull-Request Checklist


imnotjames:
With this now any old version will crash, won't it?  Can you update the peer dependencies required versions to match?

We also now have another mismatch between `ObjectID` and `ObjectId`.

imnotjames:
Also the write concern options were dropped, weren't they?

https://github.com/mongodb/node-mongodb-native/pull/2642

The mongo typings we store are also now incorrect.

imnotjames:
Given how big of a change this is, would it be safer to pin the mongo node driver version to 3 for now?

klassm:
Yes, but that means that everybody will need to pin it then :-( - that's also weird. I would still try to migrate the typings to mongo 4 (actually use the public typings from Mongo). But this involves public api changes then :-(

imnotjames:
> Yes, but that means that everybody will need to pin it then :-( 

Yes, it means they wouldn't be able to use `node-mongo@4` with TypeORM until we figure out how we want to roll out these changes.  It's been 2 days since they released v4, we can wait a little bit to upgrade.  If we pin the peer dependency version then it will be clear to anyone that tries to use TypeORM with `node-mongo` that we do not yet support v4.

> But this involves public api changes then :-(

Yeah, that's sort of the issue.

There's also cases where queries & usages of the Mongo TypeORM driver would break when upgrading from `node-mongo` 3 to 4 but keeping the same code.

I also feel like there's a few other deprecated features we were using but I cannot for the life of me remember at this point in time.

klassm:
I already have 90% done, so I would just finish it today after lunch. There seems to be three methods that don't exist any more. 
You can have a look and decide what to do :-)

klassm:
There you go - it is now quite a big PR, but effectively it uses the native types from mongo now - so next time you will at least see directly that some types don't match anymore, e.g. because mongo updated.
I am curious what you will do now - just do the peer dependency for version 3, or rather the mongo version 4 upgrade. Problem is that the upgrade will not get easier over the time :-(

src/driver/mongodb/MongoQueryRunner.ts
```diff
@@ -6,38 +7,41 @@ import {TableForeignKey} from "../../schema-builder/table/TableForeignKey";
...
-    ReadPreference,
-    ReplaceOneOptions,
+    OptionalId,
+    OrderedBulkOperation, ReduceFunction,
+    RenameOptions, ReplaceOptions,
     UnorderedBulkOperation,
-    UpdateWriteOpResult
-} from "./typings";
+    UpdateFilter, UpdateOptions, UpdateResult,
+} from "mongodb";
```
imnotjames, line 44: If we do this wouldn't we have to define `mongodb` as a full dependency rather than a peer and dev dependency?

klassm: No not really. At runtime the proper dependency is available, as it comes from the calling application. For compile time we have the one as dev dependency.
The only point is that we have to ensure that the peer dependency matches. But this also happened up to now (which is actually why I started investigating in typeorm).

imnotjames: I have a good feeling this will fail during runtime even if we aren't using mongo for cases like the browser at the very least.

>  The only point is that we have to ensure that the peer dependency matches. But this also happened up to now (which is actually why I started investigating in typeorm).

There's a very specific way we do imports of these peer dependencies - using `PlatformTools` - so that we don't bring in the dependencies in cases where it would cause failures.

klassm: Ah now I now what you mean - it's because the import is evaluated when reading the files. But that would mean that we'd have to copy all the dependency types to typeorm again - that's also weird.

imnotjames: > that's also weird.

I 100% agree.

imnotjames:
I really, really, really want to get us to a place where we can split the drivers OUT from the rest of the package.  Part of this is difficult because of how the MongoDB integration had been implemented. :cry:

The big issue I have with this vs just pinning to node-mongo v3 is that it's a breaking change & I am not comfortable pushing for a major version bump of TypeORM for this.  It's been 2 days since the node-mongo v4 release & it was an entire rewrite.

As much as I trust my friends over at mongo I'm very hesitant to make this jump.  (That reminds me.. I should ask them for coffee & chat about this a bit.)

klassm:
Yeah that's a good idea :-) But that means there is no real point in continuing here. On the upside of this, we found out what breaking changes really were there.

rubybb:
Are there any updates/progress on this?

klassm:
@rubybb We now decided to get rid of typeorm completely and use the mongo driver directly instead. Typeorm does not offer too much abstraction - so it simplifies the overall architecture.

pleerock:
Feel free to re-open in case of new interest.

rubybb:
> Feel free to re-open in case of new interest.

This issue should say opened as there's no reason to not eventually update the dependencies (unless typeorm intends to remove mongodb support) and staying on old dependencies will only lead to problems.

pleerock:
> This issue should say opened as there's no reason to not eventually update the dependencies (unless typeorm intends to remove mongodb support) and staying on old dependencies will only lead to problems.

Issue is closed because author showed that he doesn't have interest in this PR. If someone wants to make it on top of his changes, he should create a new PR.

PR #7910, opened 2021-07-15
fix: use ObjectLiteral in UpdateEvent rather than `Entity`

imnotjames:
### Description of change

The typing going into `UpdateEvent.entity` is always an `ObjectLiteral`, so update the typing on the `UpdateEvent` to match.

### Pull-Request Checklist


mathieu-bour:
Could you explain possible explain:

> The typing going into UpdateEvent.entity is always an ObjectLiteral

This modification wrecks a lot of my subscribers typings...

imnotjames:
We never sent an entity into `UpdateEvent`.  If it was close to your entity object then that was a coincidence.

mathieu-bour:
Thank you very much for the clarification!

PR #7922, opened 2021-07-16
fix: disable SQLite FK checks in synchronize / migrations

### LINKED ISSUE #2576: SQLite OneToMany ManyToOne synchronize: true Error: SQLITE_CONSTRAINT
**Issue type:**

[x] question
[x] bug report
[ ] feature request
[ ] documentation issue

**Database system/driver:**

[ ] `cordova`
[ ] `mongodb`
[ ] `mssql`
[ ] `mysql` / `mariadb`
[ ] `oracle`
[ ] `postgres`
[x] `sqlite`
[ ] `sqljs`
[ ] `react-native`
[ ] `expo`

**TypeORM version:**

[x] `latest`
[x] `@next`

First, this is my entity.

**User.ts**
```ts
import { Entity, Generated, Column, OneToMany, CreateDateColumn, UpdateDateColumn } from 'typeorm'
import Mutation from './Mutation'

@Entity('user')
export class User {
  @Column({
    unique: true,
    primary: true
  })
  @Generated('uuid')
  id: string

  @Column()
  name: string

  @Column({
    nullable: true
  })
  class: string

  @Column({
    nullable: true
  })
  nik: string

  @Column({
    nullable: true
  })
  address: string

  @Column({
    default: false
  })
  isDeleted: boolean

  @Column({
    default: 0
  })
  balance: number

  @OneToMany(type => Mutation, mutation => mutation.user)
  mutations: Mutation[]

  @CreateDateColumn()
  createDate: string

  @UpdateDateColumn()
  updateDate: string
}

export default User
```

**Mutation.ts**
```ts
import { Entity, Generated, Column, ManyToOne, CreateDateColumn, UpdateDateColumn } from 'typeorm'
import User from './User'

@Entity('mutation')
export class Mutation {
  @Column({
    unique: true,
    primary: true
  })
  @Generated('uuid')
  id: string

  @Column({
    enum: ['ADD', 'SUBSTRACT']
  })
  type: string

  @Column()
  balance: number

  @ManyToOne(type => User, user => user.mutations)
  user: User

  @CreateDateColumn()
  createDate: string

  @UpdateDateColumn()
  updateDate: string
}

export default Mutation
```

**ormconfig.js**
```js
module.exports = {
  type: 'sqlite',
  database: dbPath,
  autoSchemaSync: true,
  synchronize: true, /* I think this is cause the problem */
  autoSave: true,
  logging: true,
  entities: ['./app/database/entity/*.js']
}
```

So everything works fine. Insert new user, CRUD user, CRUD mutation.
But after close my app, and reopen my app, everything looks error.

On the console I see 
```
query: BEGIN TRANSACTION
query: SELECT * FROM "sqlite_master" WHERE "type" = 'table' AND "name" IN ('mutation', 'user')
query: SELECT * FROM "sqlite_master" WHERE "type" = 'index' AND "tbl_name" IN ('mutation', 'user')
query: PRAGMA table_info("user")
query: PRAGMA index_list("user")
query: PRAGMA foreign_key_list("user")
query: PRAGMA table_info("mutation")
query: PRAGMA index_list("mutation")
query: PRAGMA foreign_key_list("mutation")
query: CREATE TABLE "temporary_mutation" ("id" varchar PRIMARY KEY NOT NULL, "type" varchar NOT NULL, "balance" integer NOT NULL, "createDate" datetime NOT NULL DEFAULT (datetime('now')), "updateDate" datetime NOT NULL DEFAULT (datetime('now')), "userId" varchar, CONSTRAINT "FK_81b6138f2087ac02096ff1ed441" FOREIGN KEY ("userId") REFERENCES "user" ("id") ON DELETE NO ACTION ON UPDATE NO ACTION)
query: INSERT INTO "temporary_mutation"("id", "type", "balance", "createDate", "updateDate", "userId") SELECT "id", "type", "balance", "createDate", "updateDate", "userId" FROM "mutation"
query: DROP TABLE "mutation"
query: ALTER TABLE "temporary_mutation" RENAME TO "mutation"
query: CREATE TABLE "temporary_user" ("id" varchar PRIMARY KEY NOT NULL, "name" varchar NOT NULL, "class" varchar, "nik" varchar, "address" varchar, "isDeleted" boolean NOT NULL DEFAULT (0), "balance" integer NOT NULL DEFAULT (0), "createDate" datetime NOT NULL DEFAULT (datetime('now')), "updateDate" datetime NOT NULL DEFAULT (datetime('now')))
query: INSERT INTO "temporary_user"("id", "name", "class", "nik", "address", "isDeleted", "balance", "createDate", "updateDate") SELECT "id", "name", "class", "nik", "address", "isDeleted", "balance", "createDate", "updateDate" FROM "user"
query: DROP TABLE "user"
query failed: DROP TABLE "user"
error: { Error: SQLITE_CONSTRAINT: FOREIGN KEY constraint failed errno: 19, code: 'SQLITE_CONSTRAINT' }
query: ROLLBACK
Cannot make connection
```

Actually I can solve the error by drop all the data in mutation table, but I don't want to always delete all my data when I need to run the app.

After googling I found this https://github.com/typeorm/typeorm/issues/1530#issuecomment-361850741
I think looks similar, but I'm not sure how to fix correctly.

I've tried typeorm@next and @latest 
Any answer will be help

imnotjames:
### Description of change

this moves the `PRAGMA` calls to where we need them in the sqlite migration generation / synchronize

fixes #2576

### Pull-Request Checklist

### DIFF
--- a/src/driver/better-sqlite3/BetterSqlite3QueryRunner.ts
+++ b/src/driver/better-sqlite3/BetterSqlite3QueryRunner.ts
@@ -59,6 +59,20 @@ export class BetterSqlite3QueryRunner extends AbstractSqliteQueryRunner {
     }
 
     /**
+     * Called before migrations are run.
+     */
+    async beforeMigration(): Promise<void> {
+        await this.query(`PRAGMA foreign_keys = OFF`);
+    }
+
+    /**
+     * Called after migrations are run.
+     */
+    async afterMigration(): Promise<void> {
+        await this.query(`PRAGMA foreign_keys = ON`);
+    }
+
+    /**
      * Executes a given SQL query.
      */
     async query(query: string, parameters?: any[], useStructuredResult = false): Promise<any> {
--- a/test/functional/sqljs/auto-save.ts
+++ b/test/functional/sqljs/auto-save.ts
@@ -34,7 +34,7 @@ describe("sqljs driver > autosave", () => {
         await connection.createQueryBuilder().insert().into(Post).values(posts).execute();
         await connection.createQueryBuilder().update(Post).set({title: "Many posts"}).execute();
         await connection.createQueryBuilder().delete().from(Post).where("title = ?", {title: "third post"}).execute();
-        
+
         const repository = connection.getRepository(Post);
         let post = new Post();
         post.title = "A post";
@@ -52,7 +52,7 @@ describe("sqljs driver > autosave", () => {
 
         await connection.close();
 
-        expect(saves).to.be.equal(7);
+        expect(saves).to.be.equal(8);
     })));
 });
 
--- a/src/driver/capacitor/CapacitorQueryRunner.ts
+++ b/src/driver/capacitor/CapacitorQueryRunner.ts
@@ -26,6 +26,20 @@ export class CapacitorQueryRunner extends AbstractSqliteQueryRunner {
         this.broadcaster = new Broadcaster(this);
     }
 
+    /**
+     * Called before migrations are run.
+     */
+    async beforeMigration(): Promise<void> {
+        await this.query(`PRAGMA foreign_keys = OFF`);
+    }
+
+    /**
+     * Called after migrations are run.
+     */
+    async afterMigration(): Promise<void> {
+        await this.query(`PRAGMA foreign_keys = ON`);
+    }
+
     async executeSet(set: { statement: string; values?: any[] }[]) {
         if (this.isReleased) throw new QueryRunnerAlreadyReleasedError();
 
--- a/src/driver/cordova/CordovaQueryRunner.ts
+++ b/src/driver/cordova/CordovaQueryRunner.ts
@@ -29,6 +29,20 @@ export class CordovaQueryRunner extends AbstractSqliteQueryRunner {
     }
 
     /**
+     * Called before migrations are run.
+     */
+    async beforeMigration(): Promise<void> {
+        await this.query(`PRAGMA foreign_keys = OFF`);
+    }
+
+    /**
+     * Called after migrations are run.
+     */
+    async afterMigration(): Promise<void> {
+        await this.query(`PRAGMA foreign_keys = ON`);
+    }
+
+    /**
      * Executes a given SQL query.
      */
     async query(query: string, parameters?: any[], useStructuredResult = false): Promise<any> {
--- a/src/driver/expo/ExpoQueryRunner.ts
+++ b/src/driver/expo/ExpoQueryRunner.ts
@@ -127,6 +127,20 @@ export class ExpoQueryRunner extends AbstractSqliteQueryRunner {
     }
 
     /**
+     * Called before migrations are run.
+     */
+    async beforeMigration(): Promise<void> {
+        await this.query(`PRAGMA foreign_keys = OFF`);
+    }
+
+    /**
+     * Called after migrations are run.
+     */
+    async afterMigration(): Promise<void> {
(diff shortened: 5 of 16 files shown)


imnotjames:
>  It is not possible to enable or disable foreign key constraints in the middle of a multi-statement transaction (when SQLite is not in autocommit mode). Attempting to do so does not return an error; it simply has no effect.

imnotjames:
So I think this approach won't work but I have a test that should confirm that..

src/driver/better-sqlite3/BetterSqlite3QueryRunner.ts
```diff
@@ -58,6 +58,20 @@ export class BetterSqlite3QueryRunner extends AbstractSqliteQueryRunner {
         }
     }
 
+    /**
+     * Called before migrations are run.
+     */
+    async beforeMigration(): Promise<void> {
```
pleerock, line 64: can't these methods be defined in `AbstractSqliteQueryRunner`?

test/functional/sqljs/auto-save.ts
```diff
@@ -52,7 +52,7 @@ describe("sqljs driver > autosave", () => {
 
         await connection.close();
 
-        expect(saves).to.be.equal(7);
+        expect(saves).to.be.equal(8);
```
pleerock, line 55: 🤔

pleerock (approved):

phil294:
If I understand this correctly, migrations can now go haywire with FKs, but at the same time, they are guaranteed to not have you lose data acidentally because of `ON DELETE CASCADE` constraints. This was pretty daunting until now. However, is simply disabling FKs a good idea? It will now be possible to write migrations that break the internal consistency of the database. Until now, setting `PRAGMA foreign_keys=OFF` was tedious, but at least you had to know what you were doing.

ed-curran:
FYI for anybody that comes accross this trying to figure out how to use `PRAGMA foreign_keys=OFF` in migrations, this change doesn't seem to apply to migrations:run (or when using migrationsRun: true). I don't know if that was intentional or not.

Explanation in case it wasn't intentional:
`MigrationExecuter.executeMigration()` was updated for this change (to use `PRAGMA foreign_keys=OFF` if using an sqlite driver). However `Connection.runMigrations()` uses `MigrationExecuter.executePendingMigrations()` which doesn't use `MigrationExecuter.executeMigration()` so hence this change didn't do anything for migrations:run.


As a result the workaround to do this for 0.2.x is still to run the migrations yourself like this (and not use migrationsRun: true), however this means you can't use the cli for migrations:run.
```
connection.query(`PRAGMA foreign_keys=OFF`);
connection.runMigrations();
connection.query(`PRAGMA foreign_keys=ON`);
```

A longer term solution could be to have a `beforeTransaction` and  `afterTransaction` hook inside `MigrationInterface` which would allow the migration author to set `PRAGMA foreign_keys=ON` if they want it.

sinopsysHK:
Hello, this might be working well when we execute one specific migration.
But it does not apply for pending migrations being processed automatically when DataSource is initialize (if any):

https://github.com/typeorm/typeorm/blob/c669f50bddadb00d0f896e16ccfe5e1ecad16d3c/src/migration/MigrationExecutor.ts#L334-L361

Adding here as well beforeMigration and afterMigration would be consistent

netroy:
@imnotjames Hi 👋🏽 ,
was it intentional that the `beforeMigration` / `afterMigration` calls were not added in `executePendingMigrations` ?

Issue #3909, opened 2019-03-27
Define Tree Depth


raydirty:
**Issue type:**

[x] question
[ ] bug report
[ ] feature request
[ ] documentation issue

**Database system/driver:**

[ ] `cordova`
[ ] `mongodb`
[ ] `mssql`
[ ] `mysql` / `mariadb`
[ ] `oracle`
[x] `postgres`
[ ] `cockroachdb`
[ ] `sqlite`
[ ] `sqljs`
[ ] `react-native`
[ ] `expo`

**TypeORM version:**

[x] `latest`
[ ] `@next`
[ ] `0.x.x` (or put your version here)

**Steps to reproduce or a small repository showing the problem:**

Linking to the official [documentation](https://typeorm.io/#/tree-entities/working-with-tree-entities) about Trees.

Is it possible to limit the depth in a recursive tree hierarchy?

For example something like that:

```ts
const manager = getManager();
const trees = await manager.getTreeRepository(Category).findTrees({limit: 1});
```

and then the response would look like this:

```json
[{
    "id": 1,
    "name": "a1",
    "children": [{
        "id": 2,
        "name": "a11"
    }, {
        "id": 3,
        "name": "a12"
    }]
}]
```

This could be helpful to "paginate" trees and to improve the performance of APIs.

<!--
    To answer those questions you need to put "x" inside the square brackets, for example:
    [x] `mysql`
    [ ] `postgres`

    Also, please format your code properly (by taking code blocks into ```ts .... ```)
!>

johanneslumpe:
I am also interested in this. Right now the closure table strategy does not  store the tree depth in the table, so without code changes to how the strategy is implemented this probably isn't possible.

guiaramos:
Any workaround for this?

PR #7926, opened 2021-07-17
feat: add depth limiter optional parameter when loading nested trees

tiagojsag:
### Description of change

feat: add depth limiter optional parameter when loading nested trees when using TreeRepository's findTrees() and findDescendantsTree()

Modified TreeRepository's findTrees() and findDescendantsTree() public methods to now accept an optional object with config options.
For now, said object contains a depth parameter that, if set, will limit how far down the tree those methods will crawl and retrieve nested entities.

BREAKING CHANGE: TreeRepository's protected method buildChildrenEntityTree() now requires a 4th argument. Anyone affected by this break should also review and update their implementation, otherwise this feature will not work.

Closes: #3909

### Pull-Request Checklist

Unrelated tests are failing for me both on the branch and on master. I ran the individual tests I added to cover my changes, and hopefully CI will take care of the rest.

### DIFF
--- a/src/repository/FindTreesOptions.ts
+++ b/src/repository/FindTreesOptions.ts
@@ -1,0 +1,11 @@
+/**
+ * Special options passed to TreeRepository#findTrees
+ */
+export interface FindTreesOptions {
+
+    /**
+     * When loading a tree from a TreeRepository, limits the depth of the descendents loaded
+     */
+    depth?: number;
+
+}
--- a/src/repository/TreeRepository.ts
+++ b/src/repository/TreeRepository.ts
@@ -5,6 +5,7 @@ import {AbstractSqliteDriver} from "../driver/sqlite-abstract/AbstractSqliteDriv
 import { TypeORMError } from "../error/TypeORMError";
 import { FindTreeOptions } from "../find-options/FindTreeOptions";
 import { FindOptionsUtils } from "../find-options/FindOptionsUtils";
+import { FindTreesOptions } from "./FindTreesOptions";
 
 /**
  * Repository with additional functions to work with trees.
@@ -64,7 +65,11 @@ export class TreeRepository<Entity> extends Repository<Entity> {
 
         const entities = await qb.getRawAndEntities();
         const relationMaps = this.createRelationMaps("treeEntity", entities.raw);
-        this.buildChildrenEntityTree(entity, entities.entities, relationMaps);
+        this.buildChildrenEntityTree(entity, entities.entities, relationMaps, {
+            depth: -1,
+            ...options
+
+        });
 
         return entity;
     }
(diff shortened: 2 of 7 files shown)


src/repository/FindTreesOptions.ts
```diff
@@ -0,0 +1,11 @@
+/**
+ * Special options passed to TreeRepository#findTrees
+ */
+export interface FindTreesOptions {
+
+    /**
+     * When loading a tree from a TreeRepository, limits the depth of the descendents loaded
+     */
+    depth: number;
```
imnotjames, line 9: make `depth` optional.

when `depth` is undefined then treat that as the unlimited case.

I think that would be a clearer API & would match other cases where we have optional options

imnotjames (changes requested):

src/repository/FindTreesOptions.ts
tiagojsag: Thanks for the suggestion. I agree that it does indeed make more sense.

I have implemented the suggested change, and extended the test to include examples of using empty objects for the new arguments. Let me know if there's anything else you think can be improved here.

Thanks!

tiagojsag: @imnotjames I just rebased my branch to include the latest changes done to master. Assuming the tests (currently running) pass, is there anything else I can to do help move this PR forward?

tiagojsag: @imnotjames I've once again rebased my branch, and the tests are passing. Let me know if this PR is relevant to TypeORM, or if not, I can close it. thanks

pleerock:
@tiagojsag please rebase it once again and I'll merge it. Thank you!

pleerock (approved):

tiagojsag:
:+1: done

pleerock:
Thank you for contribution! 🎉

tiagojsag:
Thanks for maintaining!

src/repository/TreeRepository.ts
```diff
@@ -261,14 +266,18 @@ export class TreeRepository<Entity> extends Repository<Entity> {
         });
     }
 
-    protected buildChildrenEntityTree(entity: any, entities: any[], relationMaps: { id: any, parentId: any }[]): void {
+    protected buildChildrenEntityTree(entity: any, entities: any[], relationMaps: { id: any, parentId: any }[], options: (FindTreesOptions & { depth: number })): void {
```
mikeulvila, line 269: @tiagojsag @pleerock this definition requires me to add `depth` property to the `options` argument. 
 Since `FindTreesOptions` only has `depth` as optional property, is there a reason why `options` can't be optional and just use `FindTreesOptions` as it's type and not cause a breaking change?   As it is right now, it looks like I'll have to update `buildChildrenEntityTree` with `{ depth: -1 }` everywhere I don't want changing functionality.

tiagojsag: I think you may be correct, and this can probably be modified to be optional.

PR #7933, opened 2021-07-19
fix: allow char and varchar cases to fall through when creating parameters

### LINKED ISSUE #7932: non-ascii characters assigned to var/char columns in SQL are truncated to one byte
## Issue Description
We are using MS-SQL in typeorm, and our data model includes some `varchar` columns. When we are inserting or updating records using strings that contain non-ASCII characters, those characters are stored without any errors, but using only their second byte. For example, if we store `"\u2021"` (‡), the value actually stored to the database will be equal to `"\u0021"` (!). Note also that the object returned from e.g. `save` shows the original value, `"\u2021"` (‡). The collation setting for the database, tables, and columns are all `SQL_Latin1_General_CP1_CI_AS`, and the characters we are attempting to store can therefore be represented by the corresponding codepage.
### Expected Behavior
If a value can be stored correctly in a given var/char column in MS-SQL based on the relevant collation settings, we would expect that it would be stored unchanged. If it cannot, we would expect an error to be thrown when trying to store it. It seems that determining whether a value can be stored or not should be the responsibility of the database server, and we could just forward any error from it to the end consumer.

### Actual Behavior
Attempting to store any value that contains characters that do not fit into one byte in MS-SQL results in each character being stored as if only the second byte were set.

### Steps to Reproduce

1. Set up a project with the `mssql` driver, and an ormconfig with a `type: 'mssql'` connection which has an entity type that has at least one `char` or `varchar` column
2. Store any string into that or those columns that contain a character that must be encoded in ucs2 or above
3. Inspect the actual value stored to the database

#### ./index.ts
```typescript
import {
    createConnection,
    getConnection,
    Repository
} from 'typeorm';
import { OneModel } from './entity/OneModel';

async function main() {
    await createConnection();
    const connection = getConnection();
    const repository: Repository<OneModel> =
        connection.getRepository(OneModel);

    const one = new OneModel();
    one.content = '\u2021';

    const claimedSavedOne = await repository.save(one);
    const actualSavedOne = 
        await repository.findOne({ order: { created: 'DESC' } });

    // expect(claimedSavedOne.content).toBe(one.content) // assertion succeeds
    // expect(actualSavedOne.content).toBe(one.content); // assertion fails
    // or
    console.log(claimedSavedOne.content === one.content); // Output: true
    console.log(claimedSavedOne.content, one.content); // Output: ‡ ‡

    console.log(actualSavedOne?.content === one.content); // Output: false
    console.log(actualSavedOne?.content, one.content); // Output: ! ‡
}

main().then(() => process.exit(0));
```

#### ./entity/OneModel.ts
```typescript
import {
    Entity,
    PrimaryGeneratedColumn,
    Column,
    CreateDateColumn
} from 'typeorm';

@Entity()
export class OneModel {

    id?: string;

    @CreateDateColumn({ type: 'datetime' })
    created?: Date;

    @Column('varchar', { length: 10 })
    content: string = '';
}
```

### My Environment

| Operating System    | Windows 10.2004 (build 19041.1110) and macOS v11.2.3 |
| Node.js version     | v14.15.5 and v8.9.3 | 
| Typescript version  | v4.3.5 and v2.9.2 | 
| TypeORM version     | v0.2.34 and v0.2.22 | 
| `mssql` driver version | v7.1.3 and v4.1.0 |
| Database version | Microsoft SQL Server 2019 Express and Microsoft SQL Server 2017 |

### Additional Context

I have tracked down the point where truncation occurs to be within tedious' `data-types/varchar` module, in which varchar values are converted to bytes to be written into the SQL RPC request stream using `Buffer.from(value, 'ascii')`. Based on [this issue](https://github.com/tediousjs/tedious/issues/723) it seems that the onus is on consumers to use `nvarchar` parameters when appropriate as a solution to this issue, even when setting a `varchar` column.

- [ ] `aurora-data-api`
- [ ] `aurora-data-api-pg`
- [ ] `better-sqlite3`
- [ ] `cockroachdb`
- [ ] `cordova`
- [ ] `expo`
- [ ] `mongodb`
- [ ] `mysql`
- [ ] `nativescript`
- [ ] `oracle`
- [ ] `postgres`
- [ ] `react-native`
- [ ] `sap`
- [ ] `sqlite`
- [ ] `sqlite-abstract`
- [ ] `sqljs`
- [x] `sqlserver`

Ceshion:
### Description of change
When we are converting MS-SQL parameters from typeorm's to native, we allow the `varchar` case to fall through to `nvarchar`, and `char` to `nchar`, in order to correctly enable saving strings to columns of those types that contain characters longer than one byte.

Fixes #7932

### Pull-Request Checklist

### DIFF
--- a/src/driver/sqlserver/SqlServerQueryRunner.ts
+++ b/src/driver/sqlserver/SqlServerQueryRunner.ts
@@ -2244,15 +2244,12 @@ export class SqlServerQueryRunner extends BaseQueryRunner implements QueryRunner
             case "tinyint":
                 return this.driver.mssql.TinyInt;
             case "char":
-                return this.driver.mssql.Char(...parameter.params);
             case "nchar":
                 return this.driver.mssql.NChar(...parameter.params);
             case "text":
-                return this.driver.mssql.Text;
             case "ntext":
                 return this.driver.mssql.Ntext;
             case "varchar":
-                return this.driver.mssql.VarChar(...parameter.params);
             case "nvarchar":
                 return this.driver.mssql.NVarChar(...parameter.params);
             case "xml":
--- a/test/github-issues/7932/entity/Example.ts
+++ b/test/github-issues/7932/entity/Example.ts
@@ -1,0 +1,22 @@
+import {
+    Entity,
+    PrimaryGeneratedColumn,
+    Column,
+    CreateDateColumn
+} from '../../../../src';
+
+@Entity()
+export class Example {
+
+    @PrimaryGeneratedColumn('uuid')
+    id?: string;
+
+    @CreateDateColumn({ type: 'datetime' })
+    created?: Date;
+
+    @Column('varchar', { length: 10 })
+    content: string = '';
+
+    @Column('char', { length: 10 })
+    fixedLengthContent: string = '';
+}
--- a/test/github-issues/7932/issue-7932.ts
+++ b/test/github-issues/7932/issue-7932.ts
@@ -1,0 +1,56 @@
+import "reflect-metadata";
+import { expect } from "chai";
+import { Connection } from "../../../src";
+import { closeTestingConnections, createTestingConnections, reloadTestingDatabases } from "../../utils/test-utils";
+import { Example } from "./entity/Example";
+
+describe("github issues > #7932  non-ascii characters assigned to var/char columns in SQL are truncated to one byte", () => {
+
+    let connections: Connection[];
+    before(async () => {
+        connections = await createTestingConnections({
+            enabledDrivers: ["mssql"],
+            entities: [Example],
+            schemaCreate: false,
+            dropSchema: true
+        });
+    });
+    beforeEach(() => reloadTestingDatabases(connections));
+    after(() => closeTestingConnections(connections));
+
+    it("should store non-ascii characters in var/char without data loss", () => Promise.all(connections.map(async connection => {
+        const repo = connection.getRepository(Example);
+
+        const entity = new Example();
+        entity.content = '\u2021';
+        entity.fixedLengthContent = '\u2022';
+
+        await repo.save(entity);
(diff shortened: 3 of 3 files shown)


imnotjames:
This seems like a somewhat "nuclear" option, don't you think?

This ends up doubling the storage space of every text field & halving the maximum size.  Why not use the `nvarchar` type instead in your entity?

Also, this saves us on the two-byte unicode case.  What about 3 or 4 byte emoji? (I don't know the width of the example emoji you used)

Ceshion:
> This seems like a somewhat "nuclear" option, don't you think?
> 
> This ends up doubling the storage space of every text field & halving the maximum size. Why not use the `nvarchar` type instead in your entity?
> 
> Also, this saves us on the two-byte unicode case. What about 3 or 4 byte emoji? (I don't know the width of the example emoji you used)

I see the concern, but given this only changes the data type for parameters it would only even potentially have an impact on data length in transit, would it not? I believe it would not even change that, while there may be a performance overhead to using them when not needed, and a solution with more nuance (i.e. pre-checking the string for non-ascii characters and discriminating on that fact) *may* be more performant, but I have not checked. That kind of check would itself cause overhead of course, and I am trusting that the developers of ADO and JDBC have already taken that into account when choosing this approach for their own work (referencing the tedious issue linked in this PR's linked issue). Do you think we should re-examine it?

The emoji that I included in the test should both be 3-byte characters, and appear to throw an error on storage like they should- although SQL Server can (and will soon) begin storing those in char and varchar columns in some specific situations anyway, so shouldn't we leave determining what could or couldn't be stored in general to the database itself?

imnotjames (approved):
Given this isn't changing the way we way we're persisting on disk & only in-transit this seems reasonable.

I also happy to defer to the JDBC folks, too.

PR #7956, opened 2021-07-25
feat: add parseTableName to Driver interface

imnotjames:
### Description of change

this adds an analogue to the buildTableName method on every
driver which is the inverse.  parseTableName takes in a variety
of types (EntityMetadata, Table, View, etc) and creates an object
which has a normalized `database`, `schema`, and `tableName`

this may be used throughout the drivers & is also incredibly
useful for the Schema Builder to utilize

### Pull-Request Checklist


imnotjames:
I'm not the happiest with this because it breaks abstraction - not all drivers will have a schema.

However, because `buildTableName` requires a schema & database for all drivers I think that's acceptable until we rework that?

PR #7989, opened 2021-07-30
feat: add possibility to pass application_name via postgres specific config

das-mensch:
### Description of change

As the "extra" parameter should be deprecated in the future, the need for
passing driver specific options via the respective config objects is given.
The addition of the "applicationName" parameter is a step forward in this 
direction. This parameter is visible in logs and statistics and helps db
administrators to see which connection belongs to which application.
This change also enables the user to use code completion in case they
didn't even knew this options exists

### Pull-Request Checklist

### DIFF
--- a/src/driver/postgres/PostgresConnectionOptions.ts
+++ b/src/driver/postgres/PostgresConnectionOptions.ts
@@ -67,4 +67,10 @@ export interface PostgresConnectionOptions extends BaseConnectionOptions, Postgr
      * Automatically install postgres extensions
      */
      readonly installExtensions?: boolean;
+
+    /**
+     * sets the application_name var to help db administrators identify
+     * the service using this connection. Defaults to 'undefined'
+     */
+    readonly applicationName?: string;
 }
--- a/src/driver/postgres/PostgresDriver.ts
+++ b/src/driver/postgres/PostgresDriver.ts
@@ -1101,7 +1101,8 @@ export class PostgresDriver implements Driver {
             database: credentials.database,
             port: credentials.port,
             ssl: credentials.ssl,
-            connectionTimeoutMillis: options.connectTimeoutMS
+            connectionTimeoutMillis: options.connectTimeoutMS,
+            application_name: options.applicationName
         }, options.extra || {});
 
         // create a connection pool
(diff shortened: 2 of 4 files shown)


src/driver/postgres/PostgresDriver.ts
```diff
@@ -1101,7 +1101,8 @@ export class PostgresDriver implements Driver {
             database: credentials.database,
             port: credentials.port,
             ssl: credentials.ssl,
-            connectionTimeoutMillis: options.connectTimeoutMS
+            connectionTimeoutMillis: options.connectTimeoutMS,
+            application_name: options.applicationName || "typeorm app"
```
imnotjames, line 1105: Can this default to undefined instead

imnotjames (changes requested):

src/driver/postgres/PostgresDriver.ts
das-mensch: sure, my thought was: if any typeorm app is visible with "typeorm app" it'll help to distinguish between typeorm apps and all other apps in the database overview. A bit more "clean by default". But i see your point...

src/driver/postgres/PostgresConnectionOptions.ts
```diff
@@ -67,4 +67,10 @@ export interface PostgresConnectionOptions extends BaseConnectionOptions, Postgr
      * Automatically install postgres extensions
      */
      readonly installExtensions?: boolean;
+
+    /**
+     * sets the application_name var to help db administrators identify
+     * the service using this connection. Defaults to 'typeorm app'
```
shrujalshah28, line 73:
```suggestion
     * the service using this connection. Defaults to 'undefined'
```

shrujalshah28 (changes requested):

src/driver/postgres/PostgresConnectionOptions.ts
das-mensch: yeah saw that a bit too late. i've added it while you where reviewing i think, sorry

imnotjames (approved):

shrujalshah28 (approved):

PR #7993, opened 2021-07-30
Quick - backwards compatible fix for supporting mongo v4

PrescottJRynewicz:
### Description of change
While integrating `next-auth` into a full stack next project, I noticed that the type orm implementation of mongo v4 was breaking.  

There is another [much larger PR, and more complete PR](https://github.com/typeorm/typeorm/pull/7909) open to upgrade to mongo v4.

But, with the fix I added here, I was able to leverage the core functionality of the MongoEntity Manager without causing any regressions. 

### Pull-Request Checklist


imnotjames:
At this point in time we've pinned the peer dependencies for mongo to 3.x

Does this still apply in that case?  You really shouldn't be able to use mongo 4 with TypeORM..

PrescottJRynewicz:
Got it - @imnotjames thanks for the context - will close out if we're pinning to mongo 3.x.

PR #7805, opened 2021-06-25
 fix: better support of one-to-one and many-to-one relation-based properties in where clauses

imnotjames:
### Description of change

updates the `QueryBuilder` to support building `WHERE` clauses via `ObjectLiteral`s that go across relations.

Most of this was done to migrate the `whereInIDs` over to using native QueryBuilder features, and abstract the query compilation away from the `QueryBuilder` interface.

for example

```typescript
const postsWithFoo = await connection.createQueryBuilder(Post, "post")
            .select("post.id")
            .leftJoin("post.category", "category_join")
            .disableEscaping()
            .where({
                "category": {
                    "name": "Foo"
                }
            })
            .getMany()
```

This also applies to `FindOptions`

```typescript
const postsWithFoo = await connection
  .getRepository(Post).findOne({ where: { "category": { "name": "Foo" } } })
```

Fixes #1847
Fixes #2707
Fixes #7251
Fixes #4906
Fixes #6264
Fixes #4837
Fixes #4787

### Pull-Request Checklist

- [x] Code is up-to-date with the `master` branch
- [x] `npm run lint` passes with this change
- [x] `npm run test` passes with this change
- [x] This pull request links relevant issues as `Fixes #0000`
- [x] There are new or updated unit tests validating the change
- [x] Documentation has been updated to reflect this change
- [x] The new commits follow conventions explained in [CONTRIBUTING.md](https://github.com/typeorm/typeorm/blob/master/CONTRIBUTING.md)

PR #8005, opened 2021-08-02
Fix bug with new nested where filters

### LINKED ISSUE #8082: Cannot Query PK with FindOperator and other values correctly
## Issue Description

If you do a query that has all primary keys - even if they're `FindOperator`s TypeORM ignores the rest of the values because it assumes the PKs are uniquely identifying

### Expected Behavior

```

            it("should craft query with FindOperator with nested primary/non-primary key", () => Promise.all(connections.map(async connection => {
                const [sql, params] = connection.createQueryBuilder(HeroImage, "hero")
                    .leftJoin("hero.post", "posts")
                    .leftJoin("posts.category", "category")
                    .where({
                        post: {
                            category: {
                                id: In([1, 2, 3]),
                                name: In(["Foo", "Bar", "Baz"]),
                            }
                        }
                    })
                    .getQueryAndParameters();

                expect(sql).to.equal(
                    'SELECT "hero"."id" AS "hero_id", "hero"."url" AS "hero_url" ' +
                    'FROM "hero_image" "hero" ' +
                    'LEFT JOIN "post" "posts" ON "posts"."heroImageId"="hero"."id"  ' +
                    'LEFT JOIN "category" "category" ON "category"."id"="posts"."categoryId" ' +
                    'WHERE ("category"."id" IN (?, ?, ?) AND "category"."name" IN (?, ?, ?))'
                );

                expect(params).to.eql([1, 2, 3, "Foo", "Bar", "Baz"]);
            })));
```

### Actual Behavior

Query generated from test is:

```
SELECT "hero"."id" AS "hero_id", "hero"."url" AS "hero_url" ' +
                    'FROM "hero_image" "hero" ' +
                    'LEFT JOIN "post" "posts" ON "posts"."heroImageId"="hero"."id"  ' +
                    'LEFT JOIN "category" "category" ON "category"."id"="posts"."categoryId" ' +
                    'WHERE ("category"."id" IN (?, ?, ?))
```

### My Environment

| Node.js version     | x.y.zzz  | 
| Typescript version  | x.y.zzz  | 
| TypeORM version     | `0.2.36`  | 

| `postgres`           | no            |

tjhiggins:
### Description of change
Found a bug in the most recent nested where filter feature:
https://github.com/typeorm/typeorm/pull/7805

When combining nested primary keys and non-primary keys the query builder would build an incorrect query.
Included a test case that illustrates the issue. 

 fixes #8082

### Pull-Request Checklist

### DIFF
--- a/src/query-builder/QueryBuilder.ts
+++ b/src/query-builder/QueryBuilder.ts
@@ -993,6 +993,7 @@ export abstract class QueryBuilder<Entity> {
 
             if (metadata.hasRelationWithPropertyPath(path)) {
                 const relation = metadata.findRelationWithPropertyPath(path)!;
+                const isEntityTarget = relation.inverseEntityMetadata.target === entity[key].constructor;
                 const entityKeysLength = Object.keys(entity[key]).length;
 
                 // There's also cases where we don't want to return back all of the properties.
@@ -1007,21 +1008,34 @@ export abstract class QueryBuilder<Entity> {
                         .map(j => j.referencedColumn)
                         .filter((j): j is ColumnMetadata => !!j);
 
-                    const seenJoinColumns = new Set<string>();
-                    for (const joinColumn of joinColumns) {
-                        const entityValueMap = joinColumn.getEntityValueMap(entity[key]);
-                        for (const [columnName, value] of Object.entries(entityValueMap || {})) {
-                            if (value) {
-                                seenJoinColumns.add(columnName);
+                    if (isEntityTarget) {
+                        const hasAllJoinColumns = joinColumns.length > 0 && joinColumns.every(
+                            column => column.getEntityValue(entity[key], false)
+                        );
+
+                        if (hasAllJoinColumns) {
+                            paths.push(path);
+                            continue;
+                        }
+                    } else {
+                        const seenJoinColumns = new Set<string>();
+                        for (const joinColumn of joinColumns) {
+                            const entityValueMap = joinColumn.getEntityValueMap(entity[key]);
+                            for (const [columnName, value] of Object.entries(entityValueMap || {})) {
+                                if (value) {
+                                    seenJoinColumns.add(columnName);
+                                } else {
+                                    break;
+                                }
                             }
                         }
-                    }
 
-                    const hasOnlyJoinColumns = joinColumns.length > 0 && seenJoinColumns.size === entityKeysLength;
+                        const hasOnlyJoinColumns = joinColumns.length > 0 && seenJoinColumns.size === entityKeysLength;
 
-                    if (hasOnlyJoinColumns) {
-                        paths.push(path);
-                        continue;
+                        if (hasOnlyJoinColumns) {
+                            paths.push(path);
+                            continue;
+                        }
                     }
                 }
 
@@ -1036,23 +1050,39 @@ export abstract class QueryBuilder<Entity> {
                 // a HUGE where.
                 const primaryColumns = relation.inverseEntityMetadata.primaryColumns;
 
-                const seenPrimaryColumns = new Set<string>();
-                for (const primaryColumn of primaryColumns) {
-                    const entityValueMap = primaryColumn.getEntityValueMap(entity[key]);
-                    for (const [columnName, value] of Object.entries(entityValueMap || {})) {
-                        if (value) {
-                            seenPrimaryColumns.add(columnName);
+                if (isEntityTarget) {
+                    const hasAllPrimaryKeys = primaryColumns.length > 0 && primaryColumns.every(
+                        column => column.getEntityValue(entity[key], false)
+                    );
+
+                    if (hasAllPrimaryKeys) {
+                        const subPaths = primaryColumns.map(
+                            column => `${path}.${column.propertyPath}`
+                        );
+                        paths.push(...subPaths);
+                        continue;
+                    }
+                } else {
+                    const seenPrimaryColumns = new Set<string>();
+                    for (const primaryColumn of primaryColumns) {
+                        const entityValueMap = primaryColumn.getEntityValueMap(entity[key]);
+                        for (const [columnName, value] of Object.entries(entityValueMap || {})) {
+                            if (value) {
+                                seenPrimaryColumns.add(columnName);
+                            } else {
+                                break;
+                            }
                         }
                     }
-                }
-                const hasOnlyPrimaryKeys = primaryColumns.length > 0 && seenPrimaryColumns.size === entityKeysLength;
+                    const hasOnlyPrimaryKeys = primaryColumns.length > 0 && seenPrimaryColumns.size === entityKeysLength;
 
-                if (hasOnlyPrimaryKeys) {
-                    const subPaths = primaryColumns.map(
-                        column => `${path}.${column.propertyPath}`
-                    );
-                    paths.push(...subPaths);
-                    continue;
+                    if (hasOnlyPrimaryKeys) {
+                        const subPaths = primaryColumns.map(
+                            column => `${path}.${column.propertyPath}`
+                        );
+                        paths.push(...subPaths);
+                        continue;
+                    }
                 }
 
                 // If nothing else, just return every property that's being passed to us.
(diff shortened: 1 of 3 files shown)


tjhiggins:
@imnotjames how do I check if an entity is a transformed class vs a plain filter object? struggling with the differences between:

```ts
const foundChild = await manager.findOne(Child, { parent: loadedParent });
```

and 

```ts
.where({
    post: {
        category: {
            id: In([1, 2, 3]),
            name: In(["Foo", "Bar", "Baz"]),
        }
    }
})
```

tjhiggins:
@imnotjames The node 14 test failed due to a circleci issue - could you please re-run?

imnotjames:
Doesn't the test pass if you write the code as

```
const isEntityTarget = relation.inverseEntityMetadata.target === entity[key].constructor;

if (isEntityTarget) {
    // .... shortcuts for entity target
}

// If nothing else, just return every property that's being passed to us.
const subPaths = this.createPropertyPath(relation.inverseEntityMetadata, entity[key])
    .map(p => `${path}.${p}`);
paths.push(...subPaths);
```

tjhiggins:
> Doesn't the test pass if you write the code as
> 
> ```
> const isEntityTarget = relation.inverseEntityMetadata.target === entity[key].constructor;
> 
> if (isEntityTarget) {
>     // .... shortcuts for entity target
> }
> 
> // If nothing else, just return every property that's being passed to us.
> const subPaths = this.createPropertyPath(relation.inverseEntityMetadata, entity[key])
>     .map(p => `${path}.${p}`);
> paths.push(...subPaths);
> ```

I tried that after I added the isEntityTarget check, but a couple tests failed. Realizing now that those tests might be invalid now and it now generates the correct sql. Let me know if I'm wrong:

https://github.com/typeorm/typeorm/pull/8005/commits/6de932714bf4fa1f6fa112b5e143a892d16f069c#diff-fc53d8864169b2e2d0b175dc446c164524f59407c457d6b81a0fa1c004d8ca5cR206

Updated with the simple solution and changed those tests.

tjhiggins:
Nvm the simple solution doesn't play nice with updates:
```bash
1) embedded > embedded-many-to-one-case1
       owner side
         should insert, load, update and remove entities with embeddeds when embedded entity having ManyToOne relation2:
     EntityColumnNotFound: No entity column "counters.likedUser.id" was found.
```

tjhiggins:
@imnotjames Ready for another look. Would love your feedback on a better check for hasOnlyPrimaryKeys. Also please note that a couple tests that used to fail now pass - I'm not 100% if they are correct.

test/functional/connection-options-reader/connection-options-reader.ts
```diff
@@ -11,7 +11,7 @@ async function createDotenvFiles() {
 }
 
 async function createYamlFiles() {
-  await fs.mkdir(path.join(__dirname, "configs/yaml"));
+  await fs.mkdir(path.join(__dirname, "configs/yaml"), { recursive: true });
```
imnotjames, line 14: Why is this change needed?

src/query-builder/QueryBuilder.ts
```diff
@@ -993,51 +993,68 @@ export abstract class QueryBuilder<Entity> {
...
+
+                        if (hasAllJoinColumns) {
+                            paths.push(path);
+                            continue;
+                        }
+                    }
+
+                    if (relation.relationType === "one-to-many" || relation.relationType === "many-to-many") {
+                        throw new Error(`Cannot query across ${relation.relationType} for property ${path}`);
+                    }
```
imnotjames, line 1037: Pretty sure this needs to remain outside the `if (isEntityTarget || hasOnlyPrimaryKeys) { ... }` block as it's unrelated to the shortcuts taken.  We just do not currently support this kind of query at all.

src/query-builder/QueryBuilder.ts
```diff
@@ -993,51 +993,68 @@ export abstract class QueryBuilder<Entity> {
...
+                        const hasAllJoinColumns = joinColumns.length > 0 && joinColumns.every(
+                            column => column.getEntityValue(entity[key], false)
+                        );
+
+                        if (hasAllJoinColumns) {
+                            paths.push(path);
+                            continue;
+                        }
+                    }
+
```
imnotjames, line 1034: Does the `hasOnlyPrimaryKeys` actually apply here?  Isn't it just in the case of an `isEntityTarget`?
  Can this ever apply with `hasOnlyPrimaryKeys`?

src/query-builder/QueryBuilder.ts
```diff
@@ -993,51 +993,68 @@ export abstract class QueryBuilder<Entity> {
...
-                // a HUGE `where` to uniquely look up the entity.
+                const primaryColumns = relation.inverseEntityMetadata.primaryColumns;
+                const seenPrimaryColumns = new Set<string>();
+                for (const primaryColumn of primaryColumns) {
+                    const entityValueMap = primaryColumn.getEntityValueMap(entity[key]);
+                    for (const [columnName, value] of Object.entries(entityValueMap || {})) {
+                        if (value) {
+                            seenPrimaryColumns.add(columnName);
+                        } else {
+                            break;
```
imnotjames, line 1006: If we know this inner one is invalid, we know they all are.  Do we want to `break` further up?

src/query-builder/QueryBuilder.ts
```diff
@@ -993,51 +993,68 @@ export abstract class QueryBuilder<Entity> {
...
+                const isEntityTarget = relation.inverseEntityMetadata.target === entity[key].constructor;
 
-                // There's also cases where we don't want to return back all of the properties.
-                // These handles the situation where someone passes the model & we don't need to make
-                // a HUGE `where` to uniquely look up the entity.
+                const primaryColumns = relation.inverseEntityMetadata.primaryColumns;
+                const seenPrimaryColumns = new Set<string>();
+                for (const primaryColumn of primaryColumns) {
+                    const entityValueMap = primaryColumn.getEntityValueMap(entity[key]);
+                    for (const [columnName, value] of Object.entries(entityValueMap || {})) {
```
imnotjames, line 1002: What about embedded entity primary keys?  Is that a concern here?

src/query-builder/QueryBuilder.ts
```diff
@@ -993,51 +993,68 @@ export abstract class QueryBuilder<Entity> {
...
+                        if (value) {
```
imnotjames, line 1003: Is `null` a valid PK value?  I know `0` is.  So is an empty string.  Is that something we should be handling here?  Only checking for `undefined` instead?

src/query-builder/QueryBuilder.ts
```diff
@@ -993,51 +993,68 @@ export abstract class QueryBuilder<Entity> {
...
+                            seenPrimaryColumns.add(columnName);
+                        } else {
+                            break;
+                        }
+                    }
+                }
+                const hasOnlyPrimaryKeys = primaryColumns.length > 0 && seenPrimaryColumns.size === Object.keys(entity[key]).length;
```
imnotjames, line 1010: If it's only primary columns do we actually care if it hits the optimizations?

In theory, it would be the same either way, wouldn't it?

test/functional/connection-options-reader/connection-options-reader.ts
tjhiggins: I wasn't able to run test-fast multiple times in a row. It would complain that the directory already existed.

src/query-builder/QueryBuilder.ts
tjhiggins: The update flow uses this method and fails if you pass `contact.id` vs `contactId`. 
ex. https://github.com/typeorm/typeorm/pull/8005#issuecomment-891214828
That's the reason I added the hasOnlyPrimaryKeys check back.

src/query-builder/QueryBuilder.ts
tjhiggins: I believe you're right and I can have this block only apply to isEntityTarget.

src/query-builder/QueryBuilder.ts
tjhiggins: I'm not sure - happy to hand off this pr to you if its easier. 

Otherwise please point me to docs and I can try to add a test case.

test/functional/connection-options-reader/connection-options-reader.ts
imnotjames: That's a bigger problem that our tests have than just this.  Can you omit this change for now just to simplify this PR & what it touches?  If we have to revert or find the cause of something it makes it a whole lot easier if it's one PR for one change.

pleerock:
Closing in favour of #8082

Issue #8026, opened 2021-08-04
Inserting a value for a column that has a relation, and is also a date, results in the value being inserted as DEFAULT


jordansoltman:
## Issue Description

When inserting data into a table that has a date column that is used in a relation, Typeorm will not use the date and instead will set that value to DEFAULT in the insert query.

### Expected Behavior

When inserting dates on a column that is also used in a relation, that the date would be inserted and not set to DEFAULT.

### Actual Behavior

The query is built with DEFAULT and the date isn't used.

### Steps to Reproduce

First, we create two entities that are related on three columns (but we really only are concerned that `scheduled_departure_time` is related).

```typescript
@Entity()
export class ScheduledSailing extends BaseEntity {
    @PrimaryColumn()
    departure_terminal_id: number;

    @PrimaryColumn()
    scheduled_departure_time: Date;

    @Column()
    scheduled_arrival_time: Date;

    @Column()
    vessel_position_number: number | null;

    @ManyToOne(() => Sailing, (sailing) => sailing.scheduled_sailings)
    @JoinColumn([
        { referencedColumnName: "departure_terminal_id", name: "departure_terminal_id" },
        { referencedColumnName: "vessel_position_number", name: "vessel_position_number" },
        { referencedColumnName: "vessel_position_number", name: "vessel_position_number" }
    ])
    sailing: Sailing;
}

@Entity()
export class Sailing extends BaseEntity {
    @PrimaryColumn()
    departure_terminal_id: number;

    @PrimaryColumn()
    scheduled_departure_time: Date;

    @Column()
    vessel_position_number: number | null;

    @OneToMany(() => ScheduledSailing, (scheduledSailing) => scheduledSailing.sailing)
    @JoinColumn([
        { referencedColumnName: "departure_terminal_id", name: "departure_terminal_id" },
        { referencedColumnName: "vessel_position_number", name: "vessel_position_number" },
        { referencedColumnName: "scheduled_departure_time", name: "scheduled_departure_time" }
    ])
    scheduled_sailings: ScheduledSailing[];
}
```

Then we attempt to insert into the scheduled sailing table:

```typescript

const insertValue = {
    departure_terminal_id: 1,
    scheduled_departure_time: Date,
    scheduled_arrival_time: Date,
    vessel_position_number: 3,
}

const query = await queryBuilder
        .insert()
        .into(ScheduledSailing)
        .values([insertValue])
        .getQueryAndParameters()
```

[...]

PR #7774, opened 2021-06-21
refactor: only merge simple objects in `ORMUtils.mergeDeep`

imnotjames:
### Description of change

based on the work of #5097

fixes #5096
fixes #5762
fixes #6181
fixes #5487
fixes #5321

### Pull-Request Checklist

- [x] Code is up-to-date with the `master` branch
- [x] `npm run lint` passes with this change
- [x] `npm run test` passes with this change
- [x] This pull request links relevant issues as `Fixes #0000`
- [x] There are new or updated unit tests validating the change
- [ ] Documentation has been updated to reflect this change
- [x] The new commits follow conventions explained in [CONTRIBUTING.md](https://github.com/typeorm/typeorm/blob/master/CONTRIBUTING.md)

PR #8027, opened 2021-08-04
fix: resolve get entity value for date columns that are related

jordansoltman:
Closes: #8026

### Description of change

Allows for date columns that are part of a relation to have data inserted into them.

Fixes: #8026 

### Pull-Request Checklist

- [x] Documentation has been updated to reflect this change N/A

### DIFF
--- a/src/metadata/ColumnMetadata.ts
+++ b/src/metadata/ColumnMetadata.ts
@@ -615,7 +615,7 @@ export class ColumnMetadata {
                     if (relatedEntity && relatedEntity instanceof Object && !(relatedEntity instanceof FindOperator) && !(relatedEntity instanceof Buffer)) {
                         value = this.referencedColumn.getEntityValue(relatedEntity);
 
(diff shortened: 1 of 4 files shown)


imnotjames:
This needs a test to confirm the behavior

jordansoltman:
@imnotjames I added a test. Will it be an issue that my last commit message is a merge?

imnotjames:
No, it's all gravy.  We'll probably squash it.

imnotjames:
We dealt with something similar in https://github.com/typeorm/typeorm/pull/7774 - where only plain objects were treated with this.

Possibly tangentially related kinds of issues.

I'll try to get to actually reviewing this sooner rather than later.

imnotjames (approved):
This is fine for now.  We probably want to make this more general in the future but I think it'll help your exact issue.

In particular, we should really be checking the metadata to confirm that this property is either a plain object or a registered embedded type & failing otherwise.

jordansoltman:
It does feel like a bandage on a larger issue but I'm very new to TypeORM so I haven't had the time to really understand what the proper fix looks like. 

If I stick with TypeORM for future work I'll be happy to circle back and submit a better patch in the future.

PR #8085, opened 2021-08-15
refactor: Oracle Changes

pirouet:
Change for Oracle to be more in line with standards and syntax

BREAKING CHANGE: Use OUT_FORMAT_OBJECT and new error when Reserved words used

### Description of change

Change is being made to ensure the syntax and formatting is in line with Oracle standards. IT also introduces a breaking change to enforce the non-use of reserved words for table, column and alias names.

### Pull-Request Checklist

- [] Documentation has been updated to reflect this change N/A


pirouet:
Unexpected issues on tests for other drivers, cancelling Pull Request for now until sorted out then will reopen

src/driver/oracle/OracleDriver.ts
```diff
@@ -342,7 +373,7 @@ export class OracleDriver implements Driver {
      * Escapes a column name.
      */
     escape(columnName: string): string {
-        return `"${columnName}"`;
+        return columnName.toUpperCase();
     }
```
imnotjames, line 377: Why not just have error messages instead?  This implicitly does something else that might not be expected.

src/driver/oracle/OracleQueryRunner.ts
```diff
@@ -234,7 +236,7 @@ export class OracleQueryRunner extends BaseQueryRunner implements QueryRunner {
      * Returns raw data stream.
      */
     stream(query: string, parameters?: any[], onEnd?: Function, onError?: Function): Promise<ReadStream> {
-        throw new TypeORMError(`Stream is not supported by Oracle driver.`);
+        return this.databaseConnection.queryStream(query, parameters);
```
imnotjames, line 239: Can you do this as a separate PR next time

src/driver/oracle/OracleQueryRunner.ts
```diff
@@ -597,7 +598,7 @@ export class OracleQueryRunner extends BaseQueryRunner implements QueryRunner {
         const table = tableOrName instanceof Table ? tableOrName : await this.getCachedTable(tableOrName);
         const oldColumn = oldTableColumnOrName instanceof TableColumn ? oldTableColumnOrName : table.columns.find(c => c.name === oldTableColumnOrName);
         if (!oldColumn)
-            throw new TypeORMError(`Column "${oldTableColumnOrName}" was not found in the ${this.escapePath(table)} table.`);
+            throw new TypeORMError(`Column ${oldTableColumnOrName} was not found in the ${this.escapePath(table)} table.`);
```
imnotjames, line 601: This is an error message, not a query.  Why change it?

src/driver/oracle/OracleQueryRunner.ts
```diff
@@ -667,8 +668,8 @@ export class OracleQueryRunner extends BaseQueryRunner implements QueryRunner {
                     const newUniqueName = this.connection.namingStrategy.uniqueConstraintName(clonedTable, unique.columnNames);
 
                     // build queries
-                    upQueries.push(new Query(`ALTER TABLE ${this.escapePath(table)} RENAME CONSTRAINT "${unique.name}" TO "${newUniqueName}"`));
-                    downQueries.push(new Query(`ALTER TABLE ${this.escapePath(table)} RENAME CONSTRAINT "${newUniqueName}" TO "${unique.name}"`));
+                    upQueries.push(new Query(`ALTER TABLE ${this.escapePath(table)} RENAME CONSTRAINT ${unique.name} TO ${newUniqueName}`));
+                    downQueries.push(new Query(`ALTER TABLE ${this.escapePath(table)} RENAME CONSTRAINT ${newUniqueName} TO ${unique.name}`));
```
imnotjames, line 672: Why are we removing quotes from everything?

src/driver/oracle/OracleQueryRunner.ts
```diff
@@ -1491,8 +1492,10 @@ export class OracleQueryRunner extends BaseQueryRunner implements QueryRunner {
      */
     protected createTableSql(table: Table, createForeignKeys?: boolean): Query {
         const columnDefinitions = table.columns.map(column => this.buildCreateColumnSql(column)).join(", ");
+        if(this.driver.reservedWords.includes(table.name.toUpperCase())) {
```
imnotjames, line 1495: Isn't this not as much of a problem if they're escaped via quotes?

src/query-builder/SelectQueryBuilder.ts
```diff
@@ -2025,7 +2026,13 @@ export class SelectQueryBuilder<Entity> extends QueryBuilder<Entity> implements
             // transform raw results into entities
             const rawRelationIdResults = await relationIdLoader.load(rawResults);
             const rawRelationCountResults = await relationCountLoader.load(rawResults);
-            const transformer = new RawSqlResultsToEntityTransformer(this.expressionMap, this.connection.driver, rawRelationIdResults, rawRelationCountResults, this.queryRunner);
+
+            let transformer;
+            if(this.connection.driver instanceof OracleDriver) {
+                transformer = new OracleRawSqlResultsToEntityTransformer(this.expressionMap, this.connection.driver, rawRelationIdResults, rawRelationCountResults, this.queryRunner);
+            } else {
```
imnotjames, line 2033: Isn't this exactly the same except everything is uppercase?  Couldn't this be achieved by calling a function on the driver to normalize the column name

src/driver/oracle/OracleQueryRunner.ts
pirouet: It's not recommended therefore the intention is to align to any standard/recommendations set out by the provider (Oracle)

```
Oracle does not recommend using quoted identifiers for database object names. These quoted identifiers are accepted by SQL*Plus, but they may not be valid when using other tools that manage database objects.
```

src/driver/oracle/OracleQueryRunner.ts
pirouet: Fair point, simply caught up in the refactoring work. Will revert this part

src/driver/oracle/OracleQueryRunner.ts
pirouet: It's not but same with comment on the change, quotes are not recommended and Oracle stores details in uppercase. I don't believe we should be allowing reserved keywords to be used. From Oracle's site:
```
The following words are reserved by Oracle. That is, they have a special meaning to Oracle and so cannot be redefined. **For this reason, you cannot use them to name database objects such as columns, tables, or indexes.**
```

src/query-builder/SelectQueryBuilder.ts
pirouet: It is, yep - Couldn't see another way right now and intention - initially - to avoid messing with the RawSqlResultsToEntityTransformer and bring a wider range of drivers in to scope but happy to relook and see if I can find a better way with more time to check

src/driver/oracle/OracleDriver.ts
pirouet: Sorry, not sure I follow with error messages instead?

src/driver/oracle/OracleQueryRunner.ts
pirouet: Agreed - was meant to be part of separate branch

src/driver/oracle/OracleDriver.ts
imnotjames: If someone passes a column name that's invalid throw an error message instead of implicitly changing the column names to something the user did not intend.

imnotjames:
I'll wait until the next iteration of the PR but with this same thinking applied to other drivers we should always lowercase postgres because it's case-insensitive by default & makes everything lowercase unless you quote it.

I don't know if this makes sense because it's taking a very opinionated approach to naming for one specific driver when it isn't explicitly required by the underlying database.

src/driver/oracle/OracleDriver.ts
pirouet: Gotcha - that makes sense thanks, by adding the quote marks (pre my change) then we're already changing the column name to something that isn't what the user had initially.

I do have another change as part of this now-closed PR that if a reserved word is used then it will throw a new ReservedWordError.

pirouet:
Yeah, spot on with Oracle being upper case. I would say "yes" to making the other drivers become opinionated but that just my opinion anyway as I believe we should be looking to implement to vendor's specifications - Although we can get way with putting it around quotes, it doesn't necessarily mean we should but I'm more than happy to listen to other feedback

One bit I haven't mentioned for this work in the first place is to allow packages to be called. Trying to use a package function with typeorm as it currently is causes it to fail. For example, assume I want to use a pipelined result and set the following in my entity:

```
@Entity({
  name: 'TABLE(SCHEMA.PACKAGE.FUNCTION)'
})
```

Effectively I want to call a query that ends up like this: `SELECT * FROM TABLE(SCHEMA.PACKAGE.FUNCTION) WHERE 1 = 1;`

I can't recall exactly but this was split up using quote marks along the dots and becomes shorted/slashed up because of the 30 character limit in place (rightfully so in normal circumstances). The work in the PR is a pre-cursor to looking at the shorten/hash functions so that we can have "longer" names where applicable for Oracle only (selfishly that's all I'm interested in just now).

Issue #2363, opened 2018-06-19
[Feature Request] Upsert generation


tomer-amir-vonage:
**Issue type:**

[ ] question
[ ] bug report
[x] feature request
[ ] documentation issue

**Database system/driver:**

[ ] `cordova`
[ ] `mongodb`
[ ] `mssql`
[ ] `mysql` / `mariadb`
[ ] `oracle`
[x] `postgres`
[ ] `sqlite`
[ ] `sqljs`
[ ] `react-native`

**TypeORM version:**

[x] `latest`
[ ] `@next`
[ ] `0.x.x` (or put your version here)

**Steps to reproduce or a small repository showing the problem:**

Hi,

I am trying to do an `upsert` to a large object and I am having trouble.
I saw #1090 but this solution is not very good for me.

to upsert I need to build my query with `createQueryBuilder` but `onConflict` only excepts string queries, which makes it very difficult with large objects since I have to write my query manually.

Here is what I tied to do:
```typescript
await connenction.createQueryBuilder()
	.insert()
	.into(Model)
	.values(model)
	.onConflict(connection.createQueryBuilder().update(Model, model).where({
		uniqueField: someValue
	}).getQuery())
	.execute();
```

I think that having a function that if the insert fails just updates the whole object would be better.

is there any way to generate the `onConflict` query that I am missing?

Thanks!

pleerock:
> Here is what I tied to do:

does it work for you?

TomerAmirV:
Hi @pleerock 
it does not... 😞 
I'm having issues with parameters

pleerock:
Try this:

```ts
const updateQb = connection.createQueryBuilder().update(Model, model).where({
		uniqueField: someValue
	}); 

await connenction.createQueryBuilder()
	.insert()
	.into(Model)
	.values(model)
	.onConflict(updateQb.getQuery())
        .setParameters(updateQb.getParameters())
	.execute();
```

not sure, maybe you'll need to set native parameters as well.

TomerAmirV:
I did it, but the parameters are not ordered the same in both queries..
also, you need to start the `onConflict` with `'("violation_column")' DO` and remove the table name from the update query... 🤢

TomerAmirV:
any news?

pleerock:
What is your proposal?

TomerAmirV:
something like `onConflict(...conflictFields: string[]).doUpdate()` would be nice 😄

Devang142:
[...]

PR #8104, opened 2021-08-20
feat: add upsert methods for the drivers that support onUpdate

joeflateau:
### Description of change

This adds EntityManager#upsert, Repository#upsert and BaseEntity.upsert

Closes: #2363

Only implements upsert for drivers that InsertQueryBuilder supports onUpdate for. 

### Pull-Request Checklist


docs/entity-manager-api.md
```diff
@@ -147,6 +147,29 @@ await manager.update(User, 1, { firstName: "Rizzrak" });
 // executes UPDATE user SET firstName = Rizzrak WHERE id = 1
 ```
 
+* `upsert` - Inserts a new entity or array of entities unless they already exist in which case they are updated instead.
```
pleerock, line 150: I think we need to left a note what databases support this method

joeflateau: @pleerock I've added that

docs/entity-manager-api.md
```diff
@@ -147,6 +147,29 @@ await manager.update(User, 1, { firstName: "Rizzrak" });
 // executes UPDATE user SET firstName = Rizzrak WHERE id = 1
 ```
 
+* `upsert` - Inserts a new entity or array of entities unless they already exist in which case they are updated instead. Supported by AuroraDataApi, Cockroach, Mysql, Postgres, and Sqlite database drivers.
+
+```typescript
+await manager.upsert(User, { externalId: "abc123" }, { firstName: "Rizzrak" });
```
pleerock, line 153: TBH I find this signature confusing. I prefer to have only second signature:

```ts
await manager.update(User, ["externalId"], [
    { externalId:"abc123", firstName: "Rizzrak" },
    { externalId:"bca321", firstName: "Karzzir" },
    ]);
```

pleerock: Also I would swap parameters order, it makes more logically sense to have:

```ts
await manager.update(
  User, 
  [
      { externalId:"abc123", firstName: "Rizzrak" },
      { externalId:"bca321", firstName: "Karzzir" },
  ], 
  ["externalId"]
);
```

docs/entity-manager-api.md
```diff
@@ -147,6 +147,29 @@ await manager.update(User, 1, { firstName: "Rizzrak" });
...
+/** executes 
+ *  INSERT INTO user 
+ *  VALUES (externalId = abc123, firstName = Rizzrak) 
+ *  ON CONFLICT (externalId) DO UPDATE firstName = EXCLUDED.firstName
+ **/
+
+await manager.update(User, ["externalId"], [
```
pleerock, line 160: typo, `update` must be `upsert`

docs/entity-manager-api.md
joeflateau: I put a lot of thought into the signature and think it is best to match the style used by https://docs.mongodb.com/manual/reference/method/db.collection.findOneAndUpdate/. In addition, since the definition is essentially `upsert(Entity, ${things to match/conflict on}, ${things to update})` I felt that parameter order made more sense. I feel strongly about keeping the first signature, but a little less strongly about the order of the second signature.

docs/entity-manager-api.md
joeflateau: all set here

docs/entity-manager-api.md
pleerock: `${things to match/conflict on}` - but when there is no conflict, this value will be used for insertion, right? So it's actually `${things to match/conflict on/insert if no conflict merged with next parameter values}` - and this is what it makes this signature complicated in understanding.

joeflateau: this is where i'm hung up:

`await postRepository.upsert({ externalId }, { title: "Post title initial" });`

vs

`await postRepository.upsert({ externalId, title: "Post title initial" }, ["externalId"]);`

I feel that the first signature flows significantly better than the second. The second feels unnecessarily redundant. `MongoRepository#updateOne` has the same basic signature and I don't believe it's complicated to understand: `await mongoRepository.updateOne({ externalId }, { title: "Post title initial" }, { upsert: true });`

Perhaps there's a middle ground here? Maybe rather than 1 upsert method with 2 signatures I should introduce 2 upsert methods? `Repository#upsertOne(conditions, partial)`, a signature similar to `MongoRepository#updateOne(conditions, partial, { upsert: true })` and also `Repository#upsertMany(conflictPaths: string[], partialEntities: DeepPartial<Entity>[])`

I really don't want to lose the MongoDb style signature.

pleerock: I'm looking for both simplicity and obviousness. Adding new method won't help. Following MongoDB convention doesn't have to be there, maybe some people can appreciate it, but most prefer simplicity and obviousness. And I don't find `${things to match/conflict on/insert if no conflict merged with next parameter values}` parameter obvious. Maybe because I'm not MongoDB user. I would prefer to type extra code.

joeflateau: Well, I don't like it, but you're in charge here. I'll update the PR with your preferred style.

pleerock: sorry about that =( If you want we can wait for other people who are interested in this feature to provide their feedback.

joeflateau: no worries, I'm going to make the change. we can always reintroduce the second signature if feedback warrants it. thanks for everything you do for typeorm.

pleerock: thank you 🤞

docs/repository-api.md
```diff
@@ -154,6 +154,22 @@ await repository.update(1, { firstName: "Rizzrak" });
 // executes UPDATE user SET firstName = Rizzrak WHERE id = 1
 ```
 
+* `upsert` - Inserts a new entity or array of entities unless they already exist in which case they are updated instead.
```
pleerock, line 157: We also need to specify list of supported drivers here.

joeflateau: incoming

src/util/PropertyPath.ts
```diff
@@ -0,0 +1,35 @@
+export type MaxPropertyPathRecursionDepth = 5;
```
pleerock, line 1: This is going to be a breaking change for users since we support TypeScript versions far behind `4` where this syntax is supported. We have to use `string` for now, but in the future we have a strategy to use this syntax everywhere.

joeflateau: oh, didn't realize pre-4 was still supported :(

test/functional/repository/basic-methods/repository-basic-methods.ts
```diff
@@ -351,11 +369,187 @@ describe("repository > basic methods", () => {
...
+        function upsertableConnections(able: boolean | UpsertType) {
+            return function () {
+                return connections.filter((c) =>
+                    typeof able === "string"
+                        ? c.driver.supportedUpsertType === able
+                        : (c.driver.supportedUpsertType != null) === able
+                );
+            };
+        }
+        it("should first create then update an entity", asyncMapper(upsertableConnections(true), async (connection) => {
```
pleerock, line 387: we need to simplify code. Instead of quite complex for understanding `asyncMapper` and `upsertableConnections` we can have a simple checks like:

```ts
if (!connection.driver.supportedUpsertType) return
```

pleerock (changes requested):

src/util/PropertyPath.ts
pleerock: yeah 0.2.x is quite old and we are trying to avoid breaking changes. I hope when we finally release `0.3.0` it will use the latest TypeScript version.

joeflateau: let me know if I can be of assistance with `0.3.0` 👍

test/functional/repository/basic-methods/repository-basic-methods.ts
joeflateau: incoming

pleerock:
Thank you for your hard work!

joeflateau:
@AndrewKiri it was added in 0.2.40

AndrewKiri:
> @AndrewKiri it was added in 0.2.40

Ah sorry, thanks 🙏

joeflateau:
no prob and you're welcome! I'm happy to be able to give back to the project

Issue #8115, opened 2021-08-24
Logging option of `true` is ignored by certain logging operations


aardvarkk:
## Issue Description

I noticed this issue when trying to run `showMigrations` in the `MigrationExecutor`. Despite me having logging set to `true` in my options (which is supposed to log everything, just like `"all"`), nothing appeared in the logs. I was still seeing the query logs, but not the output from `showMigrations`.

### Expected Behavior

With logging set to `true`, everything should be logged, just as with `"all"`.

### Actual Behavior

Certain operations are not logged with logging set to `true`.

### Steps to Reproduce

1. Set logging to `true`
2. Try calling `logSchemaBuild` or `log` on the connection logger.

### Are you willing to resolve this issue by submitting a Pull Request?

 - ✅ Yes, I have the time, and I know how to start.

PR will be incoming shortly.

PR #8116, opened 2021-08-24
Fixes https://github.com/typeorm/typeorm/issues/8115

aardvarkk:
### Description of change

I noticed that certain logging operations (namely `logSchemaBuild` and `log` would not work if the TypeORM logging option was set to `true`. *Some* of the functions (such as `logQuery` and `logQueryError`) explicitly checked for `this.options === true`, but others did not.

The documentation says:

> You can enable logging of all queries and errors by simply setting logging: true in your connection options

But certain logs are left out. I created a helper function called `shouldLog` to hold the logic, and shared the code between `AdvancedConsoleLogger` and `FileLogger` since they contained effectively identical checks. I believe the helper reduces some of the visual complexity and makes the code more readable. Sharing the checks in a common function hopefully helps prevent mismatches between these different logging behaviours in the future.

### Pull-Request Checklist

### DIFF
--- a/src/logger/AdvancedConsoleLogger.ts
+++ b/src/logger/AdvancedConsoleLogger.ts
@@ -1,4 +1,4 @@
-import {LoggerOptions} from "./LoggerOptions";
+import {LoggerOptions, shouldLog} from "./LoggerOptions";
 import {PlatformTools} from "../platform/PlatformTools";
 import {QueryRunner} from "../query-runner/QueryRunner";
 import {Logger} from "./Logger";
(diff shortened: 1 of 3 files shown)


pleerock:
Okay so there are two "high levels" of logging:

* what user needs when he develops his app
* what TypeORM developer / digger needs when he debugs what TypeORM does

First group needs only logs that are created by business logic of his app - executed queries, errors.

Second group needs logs to understand what TypeORM does inside and debug something if needed.

I know it's confusing, and maybe we need a better API, but I don't want to enable `logSchemaBuild` and `log` for `logging: true` because they are rare cases for TypeORM developers. And `logging: true` is simple to use for users who don't care about TypeORM internal happening stuff.

I have to reject these changes, but thank you a lot for contribution!

PR #8127, opened 2021-08-27
feat: Allow custom update expression in InsertQueryBuilder's `orUpdate` method

### LINKED ISSUE #8125: InsertQueryBuilder `orUpdate` enhancement to support calculated values in `ON UPDATE`
## Feature Description

I am proposing an enhancement to the existing `orUpdate` function in the insert query builder before I go ahead and work on this and build an MR for the feature so we can discuss how this feature may look before I do the work.

### The Problem

The problem is that the existing `orUpdate` feature doesn't allow calculated values to be used in the update clause.

In my scenario I have an entity recording daily stats like so:
```
@Entity()
export class DailyCommunityFeedItemStats {
    @PrimaryColumn()
    public feedItemId: number;

    @PrimaryColumn('date')
    public date: string;

    @Column()
    public views: number;
}
```

I want to be able to do a single bulk insert / update query to update this table. In MySQL the query would look like so:
```
INSERT INTO daily_community_feed_item_stats ( ... )
VALUES ( ... )
ON DUPLICATE KEY UPDATE views = views + VALUES(views)
```

However, from reading InsertQueryBuilder code, I can only update views to same value as is used on insert.
```
INSERT INTO daily_community_feed_item_stats ( ... )
VALUES ( ... )
ON DUPLICATE KEY UPDATE views = VALUES(views)
```

### The Solution

Currently the signature of `orUpdate` is as follows:

```
    orUpdate(statement?: {
        columns?: string[];
        overwrite?: string[];
        conflict_target?: string | string[];
    }): this;
```

I am proposing we tweak the `overwrite` value to accept an array of objects containing `column` and `expression` fields like so
```
    orUpdate(statement?: {
        columns?: string[];
        overwrite?: string[] | {column: string, expression?: string}[];
        conflict_target?: string | string[];
    }): this;
```

Internally then I would tweak the InsertQueryBuilder to use the `expression` value if exists, otherwise fallback to a default expression of `${this.escape(column)} = VALUES(${this.escape(column)})` as it is currently

### Additional Context

My plan is to implement this for MySQL and SQLite as my use case needs both of these drivers and as they both support this custom expression functionality, see linked docs below.

I have no Postgres experience so would be reluctant to commit to support this driver also but will give it a shot also. 

[MySQL Docs](https://dev.mysql.com/doc/refman/8.0/en/insert-on-duplicate.html)
[SQLite Docs](https://www.sqlite.org/lang_UPSERT.html)

| `mysql`              | yes            |
| `postgres`           | no            |
| `sqlite`             | yes            |

wilson208:
Closes #8125 

### Description of change

This PR improves the existing `orUpdate` function within InsertQueryBuilder for MySQL, SQLite & Postgres. It allows a custom expression to be used rather that simply overriding the value.

Existing functionality allows the following:
```
await DailyStatsRepository
    .createQueryBuilder()
    .insert()
    .into(DailyStats)
    .values({feedItemId: 1, date: '2021-01-01', views: 2})
    .orUpdate(['views'], ['feedItemId', 'date'])
    .execute();
```

Which will produce a query as follows:

```
INSERT INTO daily_stats ( ... ) VALUES ( ... )
ON DUPLICATE KEY UPDATE views = VALUES(views)
```

My use case (and I have hit this in past projects also, hence the issue and PR this time), involves needing to append to an existing value, rather than simple replace it, so my solution allows the following:

```
await DailyStatsRepository
    .createQueryBuilder()
    .insert()
    .into(DailyStats)
    .values({feedItemId: 1, date: '2021-01-01', views: 2})
    .orUpdate([{column: 'views', expression: 'views + VALUES(views)'}], ['feedItemId', 'date'])
    .execute();
```

Which will produce a query as follows:

```
INSERT INTO daily_stats ( ... ) VALUES ( ... )
ON DUPLICATE KEY UPDATE views = views + VALUES(views)
```

This allows me to do a single query to update the rows, without the need to first read the data, manually update the `views` column and then insert/update. This would be fine in most instances, but this feature I am working on is going to receive high traffic and has the potential to lead to race conditions, hence being able to do this in a single insert/update is what I need.

Happy to re-work as desired to get this in, the code changes themselves aren't significant so wont take long to rework.

### Pull-Request Checklist

### DIFF
--- a/src/query-builder/InsertQueryBuilder.ts
+++ b/src/query-builder/InsertQueryBuilder.ts
@@ -371,8 +371,8 @@ export class InsertQueryBuilder<Entity> extends QueryBuilder<Entity> {
                 if (Array.isArray(overwrite)) {
                     query += ` ${conflictTarget} DO UPDATE SET `;
                     query += overwrite?.map((val: string | {column: string, expression?: string}) => {
-                        const column = this.escape(typeof val === 'string' ? val : val.column);
-                        const expression = typeof val !== 'string' && val.expression ? val.expression :  `EXCLUDED.${column}`
+                        const column = this.escape(typeof val === "string" ? val : val.column);
+                        const expression = typeof val !== "string" && val.expression ? val.expression :  `EXCLUDED.${column}`;
 
                         return `${column} = ${expression}`;
                     }).join(", ");
@@ -392,8 +392,8 @@ export class InsertQueryBuilder<Entity> extends QueryBuilder<Entity> {
                 if (Array.isArray(overwrite)) {
                     query += " ON DUPLICATE KEY UPDATE ";
                     query += overwrite?.map((val: string | {column: string, expression?: string}) => {
-                        const column = this.escape(typeof val === 'string' ? val : val.column);
-                        const expression = typeof val !== 'string' && val.expression ? val.expression :  `VALUES(${column})`
+                        const column = this.escape(typeof val === "string" ? val : val.column);
+                        const expression = typeof val !== "string" && val.expression ? val.expression :  `VALUES(${column})`;
 
                         return `${column} = ${expression}`;
                     }).join(", ");
--- a/src/query-builder/InsertQueryBuilder.ts
+++ b/src/query-builder/InsertQueryBuilder.ts
@@ -281,14 +281,14 @@ export class InsertQueryBuilder<Entity> extends QueryBuilder<Entity> {
     /**
      * @deprecated
      */
-    orUpdate(statement?: { columns?: string[], overwrite?: string[], conflict_target?: string | string[] }): this;
+    orUpdate(statement?: { columns?: string[], overwrite?: (string | {column: string; expression?: string})[], conflict_target?: string | string[] }): this;
 
-    orUpdate(overwrite: string[], conflictTarget?: string | string[]): this;
+    orUpdate(overwrite: string[] | {column: string; expression?: string}[], conflictTarget?: string | string[]): this;
 
     /**
      * Adds additional update statement supported in databases.
      */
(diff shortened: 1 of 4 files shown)


wilson208:
@imnotjames would it be possible to get a review and CI ran on this, I also don't see any documentation describing `orUpdate`, where would you like this documented to get this PR over the line?

src/query-builder/InsertQueryBuilder.ts
```diff
@@ -370,7 +370,12 @@ export class InsertQueryBuilder<Entity> extends QueryBuilder<Entity> {
 
                 if (Array.isArray(overwrite)) {
                     query += ` ${conflictTarget} DO UPDATE SET `;
-                    query += overwrite?.map(column => `${this.escape(column)} = EXCLUDED.${this.escape(column)}`).join(", ");
+                    query += overwrite?.map((val: string | {column: string, expression?: string}) => {
```
pleerock, line 373: wouldn't it apply `undefined` to the `query` if overwrite is undefined?

pleerock: okay, so `overwrite` seems to be defined here since its definitely an array. Then why do we need a question mark?

pleerock: I see it was before, but we need to correct this statement if its logically confuses.

pleerock:
If other dbs support this feature - we need to implement there as well. And yes, we need to add documentation.

pleerock:
I'm okay with this feature, do you have any concerns @imnotjames ?

pleerock:
Closing as there is no answer from author. Please feel free to re-open it with requested changes.

seancheung:
Any update on this?

vaportail:
Wow, just... wow. Yeah, who needs this :D

PR #8141, opened 2021-08-30
optional JSON/JSONB passthrough for postgres

### LINKED ISSUE #8128: PostgresDriver.preparePersistentValue for jsonb column should not call JSON.stringify if already a string
## Feature Description

I'd like to be able use a transformer to perform my own serialization/deserialization of json - for instance, using a parser which supports big integers. 

### The Problem

PostgresDriver.preparePersistentValue calls JSON.stringify even when the value given is already a string. What gets stored is therefore a quoted string containing the JSON one actually wants to store.

### The Solution

modify PostgresDriver.preparePersistentValue to pass through string values unchanged.

### Considered Alternatives

parametrize the JSON parser used in PostgresDriver so that (for example) json-bigint could be used.

### Additional Context

Large integers are pretty useful in the blockchain world - for instance IPLD's DAG-JSON codec uses many big integer fields.

| `postgres`           | yes |

softwareplumber:
### Description of change

Fixes #8128 - though better than originally proposed. Implements mechanism for postgres JSON/JSONB types similar to HSTORE. Permits developers to use their own JSON parser which is essential for supporting high precision/scale in JSON, while maintaining backward compatibility.

### Pull-Request Checklist

- [X] `npm run test` passes with this change (though I've only run for postgres)
- [X] Documentation has been updated to reflect this change (only english)

### DIFF
--- a/docs/entities.md
+++ b/docs/entities.md
@@ -573,7 +573,7 @@ You can change it by specifying your own name.
 * `asExpression: string` - Generated column expression. Used only in [MySQL](https://dev.mysql.com/doc/refman/5.7/en/create-table-generated-columns.html).
 * `generatedType: "VIRTUAL"|"STORED"` - Generated column type. Used only in [MySQL](https://dev.mysql.com/doc/refman/5.7/en/create-table-generated-columns.html).
 * `hstoreType: "object"|"string"` - Return type of `HSTORE` column. Returns value as string or as object. Used only in [Postgres](https://www.postgresql.org/docs/9.6/static/hstore.html).
-* `array: boolean` - Used for postgres and cockroachdb column types which can be array (for example int[])
+* `jsonType: "object"|"string"` - Return type of `JSON` or `JSONB` column. Returns value as string or as object. Used only in Postgres* `array: boolean` - Used for postgres and cockroachdb column types which can be array (for example int[])
 * `transformer: { from(value: DatabaseType): EntityType, to(value: EntityType): DatabaseType }` - Used to marshal properties of arbitrary type `EntityType` into a type `DatabaseType` supported by the database. Array of transformers are also supported and will be applied in natural order when writing, and in reverse order when reading. e.g. `[lowercase, encrypt]` will first lowercase the string then encrypt it when writing, and will decrypt then do nothing when reading.
 
 Note: most of those column options are RDBMS-specific and aren't available in `MongoDB`.
--- a/docs/entities.md
+++ b/docs/entities.md
@@ -573,7 +573,8 @@ You can change it by specifying your own name.
 * `asExpression: string` - Generated column expression. Used only in [MySQL](https://dev.mysql.com/doc/refman/5.7/en/create-table-generated-columns.html).
 * `generatedType: "VIRTUAL"|"STORED"` - Generated column type. Used only in [MySQL](https://dev.mysql.com/doc/refman/5.7/en/create-table-generated-columns.html).
 * `hstoreType: "object"|"string"` - Return type of `HSTORE` column. Returns value as string or as object. Used only in [Postgres](https://www.postgresql.org/docs/9.6/static/hstore.html).
-* `jsonType: "object"|"string"` - Return type of `JSON` or `JSONB` column. Returns value as string or as object. Used only in Postgres* `array: boolean` - Used for postgres and cockroachdb column types which can be array (for example int[])
+* `jsonType: "object"|"string"` - Return type of `JSON` or `JSONB` column. Returns value as string or as object. Used only in Postgres
+* `array: boolean` - Used for postgres and cockroachdb column types which can be array (for example int[])
 * `transformer: { from(value: DatabaseType): EntityType, to(value: EntityType): DatabaseType }` - Used to marshal properties of arbitrary type `EntityType` into a type `DatabaseType` supported by the database. Array of transformers are also supported and will be applied in natural order when writing, and in reverse order when reading. e.g. `[lowercase, encrypt]` will first lowercase the string then encrypt it when writing, and will decrypt then do nothing when reading.
 
 Note: most of those column options are RDBMS-specific and aren't available in `MongoDB`.
--- a/src/decorator/columns/Column.ts
+++ b/src/decorator/columns/Column.ts
@@ -17,6 +17,7 @@ import {ColumnEmbeddedOptions} from "../options/ColumnEmbeddedOptions";
 import {EmbeddedMetadataArgs} from "../../metadata-args/EmbeddedMetadataArgs";
 import {ColumnTypeUndefinedError} from "../../error/ColumnTypeUndefinedError";
 import {ColumnHstoreOptions} from "../options/ColumnHstoreOptions";
+import {ColumnJsonOptions} from "../options/ColumnJsonOptions";
 import {ColumnWithWidthOptions} from "../options/ColumnWithWidthOptions";
 import {GeneratedMetadataArgs} from "../../metadata-args/GeneratedMetadataArgs";
 import {ColumnOptions} from "../options/ColumnOptions";
@@ -90,6 +91,18 @@ export function Column(type: "hstore", options?: ColumnCommonOptions & ColumnHst
 /**
  * Column decorator is used to mark a specific class property as a table column.
  * Only properties decorated with this decorator will be persisted to the database when entity be saved.
+ */
+export function Column(type: "json", options?: ColumnCommonOptions & ColumnJsonOptions): PropertyDecorator;
+
+/**
+ * Column decorator is used to mark a specific class property as a table column.
+ * Only properties decorated with this decorator will be persisted to the database when entity be saved.
+ */
+export function Column(type: "jsonb", options?: ColumnCommonOptions & ColumnJsonOptions): PropertyDecorator;
+
+/**
+ * Column decorator is used to mark a specific class property as a table column.
+ * Only properties decorated with this decorator will be persisted to the database when entity be saved.
  *
  * Property in entity can be marked as Embedded, and on persist all columns from the embedded are mapped to the
  * single table of the entity where Embedded is used. And on hydration all columns which supposed to be in the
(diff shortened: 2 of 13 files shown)


softwareplumber:
The cockroachdb test failed with a segmentation fault... apparently after all the tests hard run. Did a local test with cockroachdb turned on and it seems to run OK.

thansen8641:
This would be amazing!!!! 🤙🏻

kordspace:
I really like how a large INT's can be represented in JSON but suddenly a JS parser won't be able to handle big INT's -- breaks JSON stringify... We need to use a different JSON parser and JSON stringify to deal with big INT's -- TypeORM can't use these big INTS, so if you want to be able to create an index on a 64 bit timestamp, in a JSON value. This fixes that!

src/decorator/options/ColumnJsonOptions.ts
```diff
@@ -0,0 +1,12 @@
+/**
+ * Column options for enum-typed columns.
```
pleerock, line 2: Wrong comment.

src/decorator/options/ColumnJsonOptions.ts
```diff
@@ -0,0 +1,12 @@
...
+ */
+export interface ColumnJsonOptions {
+
+    /**
+     * Return type of HSTORE column.
```
pleerock, line 7: Wrong comment.

docs/entities.md
```diff
@@ -573,7 +573,7 @@ You can change it by specifying your own name.
 * `asExpression: string` - Generated column expression. Used only in [MySQL](https://dev.mysql.com/doc/refman/5.7/en/create-table-generated-columns.html).
 * `generatedType: "VIRTUAL"|"STORED"` - Generated column type. Used only in [MySQL](https://dev.mysql.com/doc/refman/5.7/en/create-table-generated-columns.html).
 * `hstoreType: "object"|"string"` - Return type of `HSTORE` column. Returns value as string or as object. Used only in [Postgres](https://www.postgresql.org/docs/9.6/static/hstore.html).
-* `array: boolean` - Used for postgres and cockroachdb column types which can be array (for example int[])
+* `jsonType: "object"|"string"` - Return type of `JSON` or `JSONB` column. Returns value as string or as object. Used only in Postgres* `array: boolean` - Used for postgres and cockroachdb column types which can be array (for example int[])
```
pleerock, line 576: missing return line before `array: boolean`

src/decorator/columns/Column.ts
```diff
@@ -128,6 +141,18 @@ export function Column(typeOrOptions?: ((type?: any) => Function)|ColumnType|(Co
         if (options.type === "hstore" && !options.hstoreType)
             options.hstoreType = reflectMetadataType === Object ? "object" : "string";
 
+        if (options.type === "json" && !options.jsonType)
```
pleerock, line 144: I think better to remove this block for now to ensure its 100% non breaking change.

src/decorator/options/ColumnOptions.ts
```diff
@@ -149,6 +149,12 @@ export interface ColumnOptions extends ColumnCommonOptions {
      */
     hstoreType?: "object"|"string";
 
+    /**
+     * Return type of JSON/JSONB column.
+     * Returns value as string or as object.
```
pleerock, line 154: "Returns"? Based on issue description I thought we are talking about value we send to the database...

src/decorator/options/ColumnJsonOptions.ts
softwareplumber: Hmm, this was a copy-paste from ColumnHstoreOptions. I guess it doesn't make much sense there, either. Will ammend to 'options for JSON-typed'

docs/entities.md
softwareplumber: Sorry about that.

src/decorator/columns/Column.ts
softwareplumber: Spent quite a while thinking about this at the time. If a user manages map a string property of some entity to a json column in the database, in the expectation that json column will only ever contain strings, then of course their code will break if anyone ever inserts anything into that json column which is not a simple json string. The actual 'break' this change would imply is that in this corner case we would see quotes around their string when we didn't before (because the json parser is not being invoked). I felt that the balance was in favor of protecting the user from the consequence of their bad code design.

src/decorator/options/ColumnOptions.ts
softwareplumber: This is analogous to the HSTORE type. Yes, this applies to both the value sent and value returned from the database.

softwareplumber:
There are other builds unrelated to this PR failing with QueryFailedError: SQLITE_ERROR: cannot start a transaction within a transaction, so I'm going to assume that is not related to my update.

softwareplumber:
Virtually certain the failing test due to e4d46363917db57a9107048b973b6a12be8d61fd @TheProgrammer21

src/decorator/columns/Column.ts
softwareplumber: Happy to pull this out (and any associated unit test) if it is really the only way to get this merged, but since (for me) the only code that would break is arguably already broken, and this change should actually improve the behavior of that broken code, I think there's an argument for keeping this in. I'm going to fix everything else, rebase, and request a re-review.

TheProgrammer21:
> Virtually certain the failing test due to [e4d4636](https://github.com/typeorm/typeorm/commit/e4d46363917db57a9107048b973b6a12be8d61fd) @TheProgrammer21

@pleerock fixed the failing test on the master branch. I don't really get why it was failing initially since it ran without problems in my PR.

softwareplumber:
@TheProgrammer21 thanks that seems to have done the trick for me.

pleerock:
I don't think I want to merge this because it introduces double-way behavior. 

I would like to understand why do we have a `JSON.stringify` there at all? @imnotjames maybe you know why?

softwareplumber:
I suspect it is because, by default, node-postgres implements exactly 
the behavior that you wanted to avoid - if passed a string, it will 
parse it into JSON, as opposed to treating it as JSON string.

This is based on a vague memory rather than current research, but it 
would explain the JSON.stringify - this would put quotes around plain 
string?

------ Original Message ------
From: "Umed Khudoiberdiev" ***@***.***>
To: "typeorm/typeorm" ***@***.***>
Cc: "softwareplumber" ***@***.***>; "Author" 
***@***.***>
Sent: 11/18/2021 8:48:40 AM
Subject: Re: [typeorm/typeorm] optional JSON/JSONB passthrough for 
postgres (#8141)

pleerock:
@softwareplumber and what node-pg gives you back? Does it gives an object?

I think appropriate solution to this problem is to find out why `JSON.stringify` as used and remove it if there are no strong reasons behind.

softwareplumber:
That's not going to solve my problem though.

At the end of the day, JSON is transmitted to postgres over the wire as 
a string. At some point in the stack, the conversion will occur. 
Removing JSON.stringify from the typeorm layer will simply mean that the 
same thing happens in the postgres API layer, which will use the 
standard JSON.stringify, which will fail if the object passed to it 
contains a large integer.

The same problem applies in reverse when reading from the database.

The way this update works is that the node-postgres connection is 
created with a custom type parser which prevents node-postgres applying 
JSON.parse to JSON values incoming from the database.

if jsonType is 'object', JSON.stringify is invoked by typeorm on 
outgoing JSON parameters, and JSON.parse is applied to incoming JSON 
values. I have inspected the pg-types source to verify that JSON.parse 
is all that is applied by node-postgres, so this should maintain the 
exact same functionality as was the case before I applied the custom 
type parser.

if jsonType is 'string', typeorm doesn't apply any transformation. We 
send JSON strings to postgres and get JSON strings back from postgres. 
The only actual type conversion occurs on the postgres server, which is 
quite capable of handling large integer values (and other high-precision 
numerics). This is the new functionality I actually need.

So post this PR, we have a case where either typeORM does stringify AND 
parse, or it does neither, as controlled by jsonType. This is arguably 
cleaner than the current situation, where json.stringify is applied to 
outgoing parameters but we rely on node-postgres to parse incoming data.

Regards
Jon.

------ Original Message ------
From: "Umed Khudoiberdiev" ***@***.***>
To: "typeorm/typeorm" ***@***.***>
Cc: "softwareplumber" ***@***.***>; "Mention" 
***@***.***>
Sent: 11/18/2021 12:32:29 PM
Subject: Re: [typeorm/typeorm] optional JSON/JSONB passthrough for 
postgres (#8141)

>
***@***.*** <https://github.com/softwareplumber> and what node-pg

pleerock:
@softwareplumber sorry for late answer here... But can't we came up with a more elegant solution? Something like:

* we make node-pg to return `string` all the times
* we manually parse `JSON string` to [object with BigInts in mind](https://stackoverflow.com/questions/18755125/node-js-is-there-any-proper-way-to-parse-json-with-large-numbers-long-bigin)
* we manually stringify object by specifying to JSON.strigify [how to handle BigInts](https://stackoverflow.com/a/65156250/925151)

what do you think about it?

pleerock:
I think this issue can be closed via #8766 . Let me know if not.

softwareplumber:
Yes, sure, this PR can be summed up as *optionally* pass through JSON. Whereas #8766 appears to be #always# pass through JSON, which was *grits teeth* my original proposal, excepting that this was shot down as a breaking change.

pleerock:
since its `0.3.x` I permitted myself to introduce a breaking change to correctly support json/jsonb.

pleerock:
I had to revert my change introduced in #8766. More details in #8777. Also based on the problem I faced changes proposed in this PR won't play well. So, I'll ask to answer [my question](https://github.com/typeorm/typeorm/pull/8141#issuecomment-1041544197) regarding to further implementation of this issue.

hinogi:
@softwareplumber it's been a long while, but can you maybe review the changes once more regarding big int replacement during stringify?

gioboa:
@softwareplumber thanks for opening this PR.
I'm closing this for now because it's stale, feel free to create a new PR with your improvements. :muscle:

PR #8157, opened 2021-09-06
Typeorm creates migration that creates already existing unique constraint

### LINKED ISSUE #8158: Typeorm creates migration that creates already existing unique constraint
### Expected Behavior

Typeorm should detect the unique constraint for column was already created and skip its creation.

### Actual Behavior

Typeorm does not detect the constraint is present and tries to create it again:

```
error: error: relation "UQ_7acfebc1c4dab180b06c83c66b2" already exists
```

### Steps to Reproduce

1. Clone the repo https://github.com/styczynski/express-typeorm-starter/tree/reproduce-contraints-problem
2. Run `npm install`
3. Run `docker-compose -f ./docker-compose.yml up --abort-on-container-exit`
4. Run `npm run migrate:up`
5. Run `npm run migrate:generate NewMigration`
6. Now `npm run migrate:up` will fail (on the new migration).

The failure message will be:
```
error: error: relation "UQ_7acfebc1c4dab180b06c83c66b2" already exists
```

The problem is that if you change the order of constraints when setting up the table Typeorm will fail to correctly detect uniqueness of columns.
The problematic part is this (part of the `User.ts` model):
```
@Unique(["id", "userMetaId"])
@Unique(["userMetaId"])
/* ... */
@OneToOne(() => UserMeta)
```

The OneToOne relation requires unique constraint on the `userMetaId` column.
However if we create `["id", "userMetaId"]` constraint and then `["userMetaId"]`, Typeorm will catch only the first one
and decide that the constraint is composite so the column is not unique.

This means that if we edit the initial migration and change order of constraints so that ("userMetaId") precede the other one, everything will works fine.

### My Environment

| Operating System    |     MacOS Catalina 10.15.7   |
| Node.js version     | 12.7.0  | 
| Typescript version  | 3.9.7  | 
| TypeORM version     | 0.2.37  | 

PR for this issue: #8157

styczynski:
### Description of change

Fixes #8158

Typeorm was detecting whether unique constraint exists on column by using `find()` on all constraints for column (finding one with type `UNIQUE`) and then checking whether the constraint is a composite one. This is not a valid approach, because in case there's more than one constraint for a column and the first one is composite the code will not detect the column is in fact unique.

The code was adapted to use filter:
1. Select all constraints for column with type `UNIQUE`
2. Filter all composite constraints
3. Select first result

This approach will correctly handle situation in which there are many unique keys for a column and some of them are composite and are ordered randomly.

### Pull-Request Checklist

### DIFF
--- a/src/driver/postgres/PostgresQueryRunner.ts
+++ b/src/driver/postgres/PostgresQueryRunner.ts
@@ -1791,9 +1791,9 @@ export class PostgresQueryRunner extends BaseQueryRunner implements QueryRunner 
                     tableColumn.isPrimary = !!columnConstraints.find(constraint => constraint["constraint_type"] === "PRIMARY");
 
                     /* Find all contraints that are not composite one for specific column and take first one */
-                    const uniqueConstraint = columnConstraints.filter(constraint => constraint["constraint_type"] === "UNIQUE").map(constraint => {
+                    const uniqueConstraint: ObjectLiteral | undefined = columnConstraints.filter(constraint => constraint["constraint_type"] === "UNIQUE").filter(constraint => {
                         return !dbConstraints.find(dbConstraint => dbConstraint["constraint_type"] === "UNIQUE"
-                            && dbConstraint["constraint_name"] === uniqueConstraint["constraint_name"]
+                            && dbConstraint["constraint_name"] === constraint["constraint_name"]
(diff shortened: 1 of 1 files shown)


AlexMesser:
Your PR was superseded by #8364. Thank you for debugging and for contribution!

PR #8179, opened 2021-09-13
Option to set Postgres connection's search_path

IamBlueSlime:
### Description of change

Adds a Postgres driver-specific option to update the `search_path` of opened connections. This will enable query runners to work in the connection's schema by default, fall backing to the `default` schema (to use some extensions for instance, as uuid_generate_v4 is enabled on the default schema).

### Pull-Request Checklist

- [x] This pull request links relevant issues as `Fixes #0000` N/A

### DIFF
--- a/docs/connection-options.md
+++ b/docs/connection-options.md
@@ -195,6 +195,8 @@ See [SSL options](https://github.com/mysqljs/mysql#ssl-options).
 
 * `applicationName` - A string visible in statistics and logs to help referencing an application to a connection (default: `undefined`)
 
+* `updateSearchPath` - A boolean to control whether to update the connection's search path to prioritize the connection's schema (default: `false`)
+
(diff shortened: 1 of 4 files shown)


pleerock:
Similar PR was declined in #6796 . Why do you need it? I guess it needed only in queries executed by `.query` method with custom SQL query. Because in all other cases TypeORM places schema where its needed.

AlexMesser:
We removed `search_path` a long time ago because it caused problems with existing multi-schema functionality.

If current TypeORM schema design does not cover your needs, you can always use a "qualified object name", like `schema.table`, for example in raw queries.

outp1:
We certainly need to reopen this! @jeremylvln @AlexMesser
How can a standalone option break something? This is a feature, not a fix

PR #8204, opened 2021-09-21
✏️ Fix typos in word 'syncronization'

gradam:
Fixes typo in logs 
syncronization -> synchronization


imnotjames (approved):
Less a typo and more a misspelling but all the same this is something good to fix.

Issue #8212, opened 2021-09-23
[Bug] OneToManySubjectBuilder does not match RelationIds correctly


jbjhjm:
## Issue Description

While investigating Subject source code hunting down an issue, I found another one...
OneToManySubjectBuilder seems to not work correctly, in most cases it will fail to match persisted and incoming relationIds and will delete + insert items instead of updating.

The entities I used when I found the bug:

```ts
@Entity()
export class UserEntity {
	@PrimaryGeneratedColumn()
	id:number;
	@OneToMany("UserSettingEntity","user",{ cascade:true })
	@Expose()
	settings:UserSettingEntity[];
}

@Entity()
export class UserSettingEntity {
	@PrimaryColumn("int")
	userId:number;

	@ManyToOne("UserEntity","settings",{ cascade:false })
	user?:UserEntity;

	@PrimaryColumn("varchar")
	name:string;

	@Column({nullable:true})
	value:string;
}
```

### Expected Behavior

Running the following query...
```ts
repo.save({
  id:3,
  name:'testUser',
  settings:[
    {name:'foo',value:'bar'},
    {name:'test',value:'val'},
  ]
})
```
... should insert 1 user + 2 setting entities if database is empty.
... when user:3 and setting with {userId:3,name:'foo'} is already persisted, it should update 'foo' setting and insert 'test' setting.

### Actual Behavior

OneToManySubjectBuilder does not match the relationIds correctly.
For that reason, it will always generate "removedRelatedEntitySubject"s.
It doesn't look like wanted behavior to me, at least not following the variable names and comments.
But I'm not that proficient with TypeOrm core logic.
So I think it's best to first ask if this is how it is supposed to work or not.

The relevant code inside OneToManySubjectBuilder.ts:

L158:
```ts
EntityMetadata.difference(relatedEntityDatabaseRelationIds, relatedPersistedEntityRelationIds)
```

can never match, because relatedEntityDatabaseRelationIds contains no RelationIds but full row data:

L62:
```ts
let relatedEntityDatabaseRelationIds: ObjectLiteral[] = [];
if (subject.databaseEntity) {
  relatedEntityDatabaseRelationIds = relation.getEntityValue(subject.databaseEntity);
}
```

It seems to me that the naming of this variable is wrong and over time someone has made the assumption it does actually contain the RelationIds. 

This leads to data that EntityMetadata.difference can not match correctly:

```ts
relatedEntityDatabaseRelationIds=[
[...]

Issue #8175, opened 2021-09-11
Child entities with orphanedRowAction with foreign keys as composite primary key are deleted when updated


Metheny80:
### Expected Behavior
Child entities should exist after the `save` operation.

### Actual Behavior
Child entities are updated then deleted when calling `save`.

### Steps to Reproduce

1. Define an entity `Parent` with `OneToMany` relationship to another entity `Child`
3. Entity `Child` has a composite primary key: parent (`Parent`) id + another entity (`OtherEntity`) id
4. Save an object of type `Parent` with at least 1 child of type `Child`.
5. Make changes in one of the `Child` objects and save the parent again.
--> Children will be updated and then deleted (in the end there are no children in the DB), instead of being saved to the DB.

```typescript
@Entity({ name: 'parents' })
export class Parent {
    @PrimaryGeneratedColumn({ unsigned: true })
    id!: number;

    @OneToMany(() => Child, child => child .parent, { cascade: true })
    children!: Child[];
}


@Entity({ name: 'children' })
export class Child {
    @OneToOne(() => OtherEntity)
    @JoinColumn({ name: 'otherEntityId', referencedColumnName: "id" })
    otherEntity!: OtherEntity;

    @PrimaryColumn({ name: "otherEntityId" })
    otherEntityId!: number;  

    @ManyToOne(() => Parent, parent => parent.children, { orphanedRowAction: 'delete' })
    @JoinColumn({ name: "parentId", referencedColumnName: "id" })
    parent!: Parent;

    @PrimaryColumn({ name: "parentId" })
    parentId!: number;    

    @Column({ name: 'order' })
    order!: number; 
}

@Entity({ name: 'otherEntities' })
export class OtherEntity {
    @PrimaryGeneratedColumn({ unsigned: true })
    id!: number;
}
```

### My Environment

| Dependency          | Version  |

| Operating System   AWS Lambda
| Node.js version     14
| Typescript version  4.3.4
| TypeORM version     0.2.34

### Relevant Database Driver(s)

| DB Type              | Reproducible  |
| ---                  | ---           |
| `aurora-data-api`    | no            |
| `aurora-data-api-pg` | no            |
| `better-sqlite3`     | no            |
| `cockroachdb`        | no            |
| `cordova`            | no            |
| `expo`               | no            |
| `mongodb`            | no            |
| `mysql`              | yes            |
| `nativescript`       | no            |
| `oracle`             | no            |
| `postgres`           | no            |
[...]

PR #8221, opened 2021-09-28
Fix: OneToManySubjectBuilder generates incorrect subjects

jbjhjm:
Fixes https://github.com/typeorm/typeorm/issues/8212
Fixes https://github.com/typeorm/typeorm/issues/8175

Edit: On initial PR, I was not 100% right about the source of the issue.
After further investigation, I am now updating the description to better describe what is happening.

In certain cases, OneToManySubjectBuilder will use wrong relation key data which stops it from correctly matching keys of incoming data against database keys.
This will lead to entities being deleted and (re-)inserted instead of being updated and/or staying untouched.

Reason for this is the use of `relation.getEntityValue(subject.databaseEntity)`.
The code assumes that the returned value will only contain the related items key data.
However this can be incorrect. If one uses the afterLoad hook to generate entity properties, the returned object may include additional properties.

### Description of change

The required change is simple:
Pass the data within entityDatabaseValues to getEntityIdMap to get a correct relationId map in all cases.

- [X] `npm run lint` passes with this change -- Passed on CircleCI
- [X] `npm run test` passes with this change -- Passed on CircleCI
- [X] There are new or updated unit tests validating the change -- see info at top
- [X] Documentation has been updated to reflect this change -- not needed


imnotjames:
Lots of test failures.  Can you look into those?

pleerock:
Have you finished with changes? I pulled again, removed changes and still don't see failing tests. Please let me know by pinging me once you finish.

jbjhjm:
@pleerock sorry, forgot to submit the comment it seems.

you were 100% right when asking me to dig deeper.
I discovered that the fix works well, but I was not right about the source of the issue.
I added new commits to reflect new insights.

- refactored the tests for correctly working child composite keys into a new functional test suite, think it's good to have them to ensure it works fine in general
- written a new test that will fail without the fix being applied
- improved notes in source code

The issue does only appear in certain cases, when using a afterLoad subscriber to set some entity data.
This would add additional properties to the child key data item loaded when fetching the parent databaseEntity.
Until now, OneToManySubjectBuilder did not handle this situation correctly,
it tried to match incoming keys against the database keys possibly polluted with additional props.

In this situation, the key mismatch would lead to children being deleted from database and new ones being inserted, also beforeRemove / beforeInsert would be called instead of the expected beforeUpdate.

---

About your most recent comment, it's irritating to hear that.
I just checked if the test would fail if I roll back OneToManySubjectBuilder, and it does fail as expected:

```
 1) github issues > #8221
       afterLoad entity modifier must not make relation key matching fail:

      AssertionError: expected 2 to equal 0
      + expected - actual

      -2
      +0

      at J:\Projects\typeorm\test\github-issues\8221\issue-8221.ts:62:42
```
L62: expect(subscriber.counter.deletes).to.equal(0);
which happens because wrong subjects are being generated.

pleerock:
Yes, they are failing now.

I guess now you have introduced **another bug** because your test is invalid and actually it **must** generated `DELETE` queries, because these lines of code in the test:

```ts
await userRepo.save([{
    id:1,
    settings:[
	    {id:1,name:"A",value:"foobar"},
	    {id:1,name:"B",value:"testvalue"},
    ]
}]);
```

must be:

```ts
await userRepo.save([{
    id:1,
    settings:[
	    {assetId:1,name:"A",value:"foobar"},
	    {assetId:1,name:"B",value:"testvalue"},
    ]
}]);
```

pleerock:
Okay after hours of debugging I can say your solution looks correct. I found magic code that automatically finds first primary key (`assertId`) just based on parent entity and second primary key (`name`). Damn cascades are super complex.

jbjhjm:
> Okay after hours of debugging I can say your solution looks correct. I found magic code that automatically finds first primary key (`assertId`) just based on parent entity and second primary key (`name`). Damn cascades are super complex.

Great! Can you give me a hint on where to find the mentioned magic code? I'd want to take a look and maybe add some notes to the WIP documentation update on cascades https://github.com/typeorm/typeorm/issues/8277

pleerock:
It's subject.identifier gets filled automatically, actually not sure I can explain everything shortly, u need to spend few hours debugging to figure it all out 😆 Also I'm not sure I was completely right in all my assumptions, because some code goes too deeply. Hate cascades. It take too much time and efforts and sometimes I think better not to have this feature TBH.

Issue #8240, opened 2021-10-03
Support ViewEntity  dependency configuration


zaro:
## Feature Description
Currently TypeORM doesn't support any way of specifying dependencies between views, which often leads to  migrations  being able to execute only after manual editing of the generated  code.

### The Problem

Suppose we have ViewA  and ViewB. ViewB is using ViewA inside it's definition, so the normal order is to create first ViewA and then ViewB. Now if there is a change in ViewA, we can't execute it w/o first dropping ViewB, adjusting VIewA and then creating ViewB back.  In typeorm if we edit only ViewA, the resulting migration will generate only changes for ViewA , and the migration will fail to execute because ViewB still holds a reference to ViewA and the DROP VIEW statement will fail. A workaround this will be to make also some trivial change to ViewB so it also ends up being updated in the migration. Unfortunately even this is not sufficient as the order in which these are generated is based on import order or something like this, so the DROP/CREATE statements often end up in the wrong order , e.g dropping ViewA before ViewB, and a manual edit of the migration file is needed.
With more views and dependencies the process can get quite cumbersome and error prone.

### The Solution

Introduce new config for the VIewEntity decorator where we can explicitly list that a view is depending on another view and then generate the migration in the correct order based on this configuration. 

Example

```ts
////////////////////////////////////////////////////////////
// viewA.ts
@ViewEntity({
    expression: `
      select * from tableA    
    `
})

export class ViewA{
//...
}
////////////////////////////////////////////////////////////
// viewB.ts

import {ViewA} from './viewA.ts';

@ViewEntity({
    expression: `
      select * from ViewA  ​ where column > 5
   `,
   dependsOn: [ViewA]
})
export class ViewB{
//...
}

```

### Considered Alternatives
None that I could think of

### Relevant Database Driver(s)
Not connected to specific drivers, affects all SQL databases that support views.

### Are you willing to resolve this issue by submitting a Pull Request?

 - ✅ Yes, I have the time, and I know how to start.
 - ✖️ Yes, I have the time, but I don't know how to start. I would need guidance.
 - ✖️ No, I don’t have the time, but I can support (using donations) development.
 - ✖️ No, I don’t have the time and I’m okay to wait for the community / maintainers to resolve this issue.

mwanago:
@zaro Can this issue be closed, since you've merged the PR?
Cool work! :+1:

zaro:
@mwanago Yes, thanks for reminding me.

yanqic:
[...]

PR #8261, opened 2021-10-10
feat: Add dependency configuraiton for views #8240

zaro:
### Description of change
See #8240 for detailed description of the problem.

The PR  adds dependsOn option to @View decorator, where dependencies can be listed. Also use these dependencies to order drop/create view correctly when generating migrations

### Pull-Request Checklist

### DIFF
--- a/src/schema-builder/RdbmsSchemaBuilder.ts
+++ b/src/schema-builder/RdbmsSchemaBuilder.ts
@@ -456,42 +456,54 @@ export class RdbmsSchemaBuilder implements SchemaBuilder {
 
             this.connection.logger.logSchemaBuild(`dropping an old view: ${view.name}`);
 
-            // drop an old view
-            //await this.queryRunner.dropView(view);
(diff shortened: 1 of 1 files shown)


src/schema-builder/util/ViewUtils.ts
```diff
@@ -0,0 +1,27 @@
...
+      )) {
+          return 1;
+      }
+      if(metadataB.dependsOn && (
+          metadataB.dependsOn.has(metadataA.target) ||
+          metadataB.dependsOn.has(metadataA.name)
+      )){
+          return -1;
+      }
+      return 0;
```
alumni, line 24: Requires a topological sort, not an array sort.

Here, `return 0` will mean two items without an edge are equal, e.g. if we have the dependencies `C > B > A`, this function will return `0` for the pair `(C, A)`, meaning `C` and `A` are equal. Which actually makes the test for this issue fail :)

PR #8283, opened 2021-10-17
fix: create a different cacheId if present for count query in getManyAndCount

### LINKED ISSUE #4277: Using cache in findAndCount and getManyAndCount returns 0 as count

eugenio165:
### Description of change

Creates a different cacheId (if provided) for the count query in `getManyAndCount` so that it does not retreive the cache for the previous query (the select immediately above it). Previously, when providing a `cacheId` and using `getManyAndCount` or `findAndCount`, the count query would use the exact same `cacheId` as the select query above it, causing the `count` value to return as 0 (because it doesnt even exist in the select query result). Providing a different `cacheId` for the count query solves the problem. Only happens when explicitly defining a `cacheId`.

Fixes #4277 

### Pull-Request Checklist

- [ ] Documentation has been updated to reflect this change **N/A**

### DIFF
--- a/src/query-builder/SelectQueryBuilder.ts
+++ b/src/query-builder/SelectQueryBuilder.ts
@@ -1199,6 +1199,10 @@ export class SelectQueryBuilder<Entity> extends QueryBuilder<Entity> implements 
             this.expressionMap.queryEntity = true;
             const entitiesAndRaw = await this.executeEntitiesAndRawResults(queryRunner);
             this.expressionMap.queryEntity = false;
+            const cacheId = this.expressionMap.cacheId;
+            // Creates a new cacheId for the count query, or it will retreive the above query results
+            // and count will return 0.
+            this.expressionMap.cacheId = (cacheId) ? `${cacheId}-count` : cacheId;
             const count = await this.executeCountQuery(queryRunner);
             const results: [Entity[], number] = [entitiesAndRaw.entities, count];
 
--- a/test/github-issues/4277/entity/User.ts
+++ b/test/github-issues/4277/entity/User.ts
@@ -1,0 +1,10 @@
+import {Entity, PrimaryGeneratedColumn, Column} from "../../../../src";
+
+@Entity()
+export class User {
+    @PrimaryGeneratedColumn({ type: "integer" })
+    id: number;
+    
+    @Column()
+    name: string;
+}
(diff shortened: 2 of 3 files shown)


shrujalshah28:
This does solve #4277 but now we have another issue, as the `count` cache key is added separately and internally, it does not remove the `count` cache even if we remove cache by `cache-id` by that we added and so `count` in return value does not change until cache expired.

Issue #8275, opened 2021-10-14
Option to disable deletion/orphanage of unloaded relations


jbjhjm:
### The Problem

Due to the current implementation of orphanedRowAction, saving a parent of a OneToMany relation will always delete entries that were not loaded before saving. In some situations that's the wanted behavior, but in others this may delete entries "accidentally".

Example:
Let's assume we have a User with a number of related Settings.
Some settings are controlled by the user, some by admins and may even be invisible to the users themselves.
The user edits his/her profile and the data is sent to backend - a UserEntity with only those settings that the user can see/edit.
When saving the user, this would now delete any other persisted setting.

Has this been the case before introducing orphanedRowAction? I'm not sure.

However, for such situations, there should be an option to disable orphanedRowAction.

### The Solution

Pretty simple as far as I can see.
At https://github.com/typeorm/typeorm/blob/master/src/persistence/subject-builder/OneToManySubjectBuilder.ts#L160 ,
RemoveSubjects are being generated for all subjects that are persisted but not in loaded data.

So the modification would only be to use a new orphanedRowAction setting or a new config variable which does skip the whole section.

first I thought of allowing orphanedRowAction=none.
But for descriptive naming, I think it would be better to create a new config prop like orphaneUnloadedRows = true|false.

What do you think? It would be great to hear your opinion @nebkat as you do know best about the orphanedRowAction mechanism.

### Relevant Database Driver(s)

Relevant to all

### Are you willing to resolve this issue by submitting a Pull Request?

 - ✅ Yes, I have the time, and I know how to start.
 - ✖️ Yes, I have the time, but I don't know how to start. I would need guidance.
 - ✖️ No, I don’t have the time, but I can support (using donations) development.
 - ✖️ No, I don’t have the time and I’m okay to wait for the community / maintainers to resolve this issue.

pleerock:
> The user edits his/her profile and the data is sent to backend - a UserEntity with only those settings that the user can see/edit.

I think this can be easily fixed by controlling what you are sending to `save` after you receive data from client.
I strongly recommend to do it this way.

jbjhjm:
Will be fixed by https://github.com/typeorm/typeorm/pull/8678

PR #8285, opened 2021-10-18
Feat: orphanedRowAction=skip to disable auto-orphanage of children that are not loaded during updating parent

jbjhjm:
### Description of change

Fixes https://github.com/typeorm/typeorm/issues/8275

When saving a parent of a OneToMany relation, OneToManySubjectBuilder will currently orphane/delete all items that are not within the list of to-be-saved children.

This is not always wanted behavior. Example:

Let's assume we have a User [parent] with a number of [child] Settings.
Some of these settings may only be modified and visible to admins, others can be seen and edited by users themselves.
The user edits his/her profile and the data is sent to backend - a UserEntity with only those settings that the user can see/edit.

Currently, we'd now have to fetch all settings from database for all endpoints updating users, merge them with incoming changes and then save them - just to let OneToManySubjectBuilder that the persisted rows should not be removed.

Summing up, there are situations where it is just way simpler if one can disable the auto-orphaning mechanism.
Furthermore I believe that the behavior of auto-orphaning on update did not always exist and has been added some time ago.

Fortunately, this can be done very easily:
We add "skip" to allowed values for orphaneRowAction and do a simple if-condition skipping the auto-orphanage section.

### Pull-Request Checklist

- [x] `npm run lint` fails 737 times on code not modified by myself


pleerock:
The user edits his/her profile and the data is sent to backend - a UserEntity with only those settings that the user can see/edit.

I think this can be easily fixed by controlling what you are sending to save after you receive data from client.
I strongly recommend to do it this way.

e.g. don't send settings that you receive from the client to save method, since it's not a desirable action for you.

jbjhjm:
@pleerock 

I think you misunderstood the intention of this feature. Of course there are other ways to handle data.
But this is about adding one piece of configurability that has not been considered yet.

Currently we can specify HOW orphanage should work. But not IF it should be applied at all.
It is risky, data might be deleted by accident. The behavior of the mechanism is not obvious:

Saving children when saving parent is logical.
Deleting children when deleting parent is local.
Deleting children when updating parent is not obvious at all.

In many cases it is fine to have this feature enabled but one needs to be cautious with it.
For these reasons I strongly believe that we should be able to be in full control of this functionality.
It should be possible to disable it if one does not want to apply this behavior.

pleerock:
Okay since we are continuing merging cascade-related features (which I really hate), let's merge this one as well. I don't really like `skip` name because it sounds like something related to take/skip to me. What about to use `ignore` instead? Or do you have a better options?

jbjhjm:
Great!
Well we should be sure to find a term that makes sense in context to "orphanedRowAction".
"Ignore orphaned row action" might be too ambiguous to immediately understand what it does.
"Disable orphaned row action" sounds more explicit to me. What do you think about keyword `disable`?

Let me know what fits to typeORM, will do a rebase + wording update then.

jbjhjm:
@pleerock need a short feedback here to prepare PR update.
What fits better to typeorm naming conventions:

orphanedRowAction: disable
or
orphanedRowAction: ignore

thanks!

pleerock:
I'm okay with `disable`. Thank you for patience :)

jbjhjm:
closing in favor of rebased/updated PR https://github.com/typeorm/typeorm/pull/8678

Issue #7898, opened 2021-07-13
Extending an entity does not override property decorators


harveyappleton:
## Issue Description

If you have an abstract class with decorators, and extend it and override those decorators, the original decorators still affect the extended entity.

### Expected Behavior

Take a base entity, eg:

```typescript
export abstract class CoreEntity {

@Column({ unique: true, nullable: false })
test: string;

}
```

and then extend and override:

```typescript
export class MyEntity extends CoreEntity {

@Column({ unique: false, nullable: true })
test: string;

}
```

then the MyEntity table structure should have a column called test **that is nullable and no unique constraint**.

_(The reason this is useful is for audit logs, where my core entity is unique constrained/not nullable, but the audit log should not be unique constrained, otherwise I won't be able to add more than 1 log!!)_

### Actual Behavior

With the above entity example, the MyEntity table structure has a column called test **that is not nullable and has a unique constraint**.

### Steps to Reproduce

1. Create the abstract entity and the my entity (as per above)
2. Check the structure of the table to see the properties aren't being overridden

### My Environment

| Dependency          | Version  |
| ---                 | ---      |
| Operating System    |   macOS Big Sur 11.4       |
| Node.js version     | v16.4.2 | 
| Typescript version  | v4.3.5 | 
| TypeORM version     | v0.2.34 | 

### Relevant Database Driver(s)

- [ ] `aurora-data-api`
- [ ] `aurora-data-api-pg`
- [ ] `better-sqlite3`
- [ ] `cockroachdb`
- [ ] `cordova`
- [ ] `expo`
- [ ] `mongodb`
- [x] `mariadb`
- [x] `mysql`
- [ ] `nativescript`
- [ ] `oracle`
- [ ] `postgres`
- [ ] `react-native`
- [ ] `sap`
- [ ] `sqlite`
- [ ] `sqlite-abstract`
- [ ] `sqljs`
- [ ] `sqlserver`

### Are you willing to resolve this issue by submitting a Pull Request?

- [ ] Yes, I have the time, and I know how to start.
- [x] Yes, I have the time, but I don't know how to start. I would need guidance.
- [ ] No, I don't have the time, although I believe I could do it if I had the time...
- [ ] No, I don't have the time and I wouldn't even know how to start.

pleerock:
If you are ready to contribute, the best is to start from a PR with a failing test. If we see a failing test we can debug where is the problem. Most probably it's inside `EntityMetadataBuilder` class.

[...]

PR #8320, opened 2021-10-30
fix: entity override property

viveknathani:
for an extended entity to be able to override a property correctly, the
filtering functions need to take care of choosing correctly by using
the inheritance tree's topmost element

Closes: #7898

### Description of change

Debugging the build function in `EntityMetaDataBuilder` class show thats `entityMetaData` is being built incorrectly in `computeEntityMetadataStep1` function. During the step where `ownColumns` property needs to be created, the `filterColumns` function will choose the column for the base entity and not for the derived entity even when the column is to be built for the derived entity. This is fixed by ensuring that the column chosen has the target equal to the topmost element of the inheritance tree because the topmost element (or the one at index 0) is the derived entity. The same problem happens further during the setting of indices through unique constraint. The `filterUniques`  function should match the target with the first element of the inheritance tree. 

### Pull-Request Checklist


JakaBernardNudge:
@viveknathani Any news regarding this?

Issue #5829, opened 2020-04-08
Entity .find() running out of memory with nested eager relations


adamlong5:
**Issue type:**

[ ] question
[x] bug report
[ ] feature request
[ ] documentation issue

**Database system/driver:**

[ ] `cordova`
[ ] `mongodb`
[ ] `mssql`
[ ] `mysql` / `mariadb`
[ ] `oracle`
[x] `postgres`
[ ] `cockroachdb`
[ ] `sqlite`
[ ] `sqljs`
[ ] `react-native`
[ ] `expo`

**TypeORM version:**

[x] `latest`
[ ] `@next`
[ ] `0.x.x` (or put your version here)

**Steps to reproduce or a small repository showing the problem:**

1. I have an entity ("Entity A") that has 4 eager-loaded relations.
2. One of those four eager-loaded relations ("Entity B") in turn has 3 eager-loaded relations.
3. I'm querying multiple of Entity A via `[Entity A].find()` using the ActiveRecord pattern
4. When I do, I get a heap out of memory error when TypeORM is constructing the query string (full stack trace/error below)
5. If I toggle Entity A to no longer eager-load Entity B, I no longer get the error and the query works. If I then add `relations` to the `.find()` to bring back in Entity B and its sub-relations, everything works as well.

**Heap Error**
```console
<--- Last few GCs --->

[15498:0x110000000]    80987 ms: Mark-sweep 1355.1 (1439.8) -> 1355.0 (1439.8) MB, 302.9 / 0.0 ms  (average mu = 0.122, current mu = 0.000) last resort GC in old space requested
[15498:0x110000000]    81318 ms: Mark-sweep 1355.0 (1439.8) -> 1355.0 (1439.8) MB, 331.1 / 0.0 ms  (average mu = 0.062, current mu = 0.000) last resort GC in old space requested


<--- JS stacktrace --->

==== JS stack trace =========================================

    0: ExitFrame [pc: 0x29fb9595be3d]
    1: StubFrame [pc: 0x29fb9591059d]
Security context: 0x269d6c79e6e9 <JSObject>
    2: /* anonymous */ [0x269d5fa6e5a1] [[redacted]/node_modules/typeorm/query-builder/QueryBuilderUtils.js:~14] [pc=0x29fb9646a072](this=0x269d0932c559 <JSFunction QueryBuilderUtils (sfi = 0x269d5fa6e329)>,str=0x269d22c6ac89 <Very long string[77269]>)
[...]

PR #8323, opened 2021-11-01
feat: implement separate fetch relations in findOptions

GaryXiongxiong:
This new feature allow fetch relations in separated sql query, that  will fix memory running out when .find() called with multiple join relations. This issue is described in #5829 & #3857

Fixes: #5829

### Description of change

This new feature allow fetch relations in separated sql query, that  will fix memory running out when .find() called with multiple join relations. This issue is described in #5829 & #3857

You can fetch relations by using:
```typescript
    const posts = await connection.manager.find(Post, {
        relations: ["category"],
        where: {category:{name: "Category #2"}},
        separateFetchRelations: ["author","author.post"]
    });
```
Then the relation `author` and `post` of `author` will be queried in separated sql:
```sql
SELECT * FROM post 
LEFT JOIN category on category.post_id = post.id
WHERE category.name = 'Category #2';

SELECT * FROM author
WHERE author.post_id = post_id; --post's id just fetched

SELECT * FROM post
WHERE post.author_id = author_id; --author's id just fetched
```

### Pull-Request Checklist

### DIFF
--- a/src/find-options/FindOneOptions.ts
+++ b/src/find-options/FindOneOptions.ts
@@ -65,6 +65,10 @@ export interface FindOneOptions<Entity = any> {
     /**
      * If this is set to true, SELECT query in a `find` method will be executed in a transaction.
      */
-    transaction?: boolean
+    transaction?: boolean;
 
+    /**
+     * Indicates if this query going fetch relations in separated queries.
+     */
+    separateFetchRelations?: string[];
 }
--- a/src/find-options/FindOptionsUtils.ts
+++ b/src/find-options/FindOptionsUtils.ts
@@ -5,6 +5,7 @@ import {FindRelationsNotFoundError} from "../error/FindRelationsNotFoundError";
 import {EntityMetadata} from "../metadata/EntityMetadata";
 import {DriverUtils} from "../driver/DriverUtils";
 import { TypeORMError } from "../error";
+import { RelationSeparateAttributes } from "../query-builder/relation-separate/RelationSeparateAttributes";
 
 /**
  * Utilities to work with FindOptions.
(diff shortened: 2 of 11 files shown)


pleerock:
Functionality of separate queries for relations were implemented in `@next` branch **years** ago. Unfortunately we were not able to release them in time, and now `@next` is far away behind master in many aspects.

I think current implementation of this feature in `@next` is more advanced and we should port changes from there to `master`.

But the way it was implemented in `@next` is simply replacing of current `join` approach **completely** to separate queries. Some people were disagree, because it some cases `joins` are more effective. But mostly I think it's a better approach.

So, what if we port what we have in `@next`, keep what we have in `master` and introduce a new option called `relationLoadStrategy: "join" | "query"` ?

@GaryXiongxiong do you want to spend more of your time / will in order to check how it's currently implemented in `@next` and think how to properly port it to `master`?

YixiongJiang:
> Functionality of separate queries for relations were implemented in `@next` branch **years** ago. Unfortunately we were not able to release them in time, and now `@next` is far away behind master in many aspects.
> 
> I think current implementation of this feature in `@next` is more advanced and we should port changes from there to `master`.
> 
> But the way it was implemented in `@next` is simply replacing of current `join` approach **completely** to separate queries. Some people were disagree, because it some cases `joins` are more effective. But mostly I think it's a better approach.
> 
> So, what if we port what we have in `@next`, keep what we have in `master` and introduce a new option called `relationLoadStrategy: "join" | "query"` ?
> 
> @GaryXiongxiong do you want to spend more of your time / will in order to check how it's currently implemented in `@next` and think how to properly port it to `master`?

Yeah, will take some time for it

GeiOoo:
I dont wanna rush... but is there any progress on that topic?
We've got some huge performance problems with fairly complex entity relations.
A fully resolved query for a single record can easily result in a 75000 row result set taking about 2 seconds to execute.
We would love to spread those ``LEFT JOIN`` in separate queries to minimize data flow.

Penners:
Feature looks great! Looking forward to when it gets merged.

pleerock:
implemented in #8616

PR #8360, opened 2021-11-13
test: make tests consistent on machines which have different timezone…

### LINKED ISSUE #6931: test: "database schema > column types > postgres" is not working with different timezone
**Issue type:**

[x] bug report
[ ] feature request
[x] documentation issue

**Database system/driver:**

[ ] `cordova`
[ ] `mongodb`
[ ] `mssql`
[ ] `mysql` / `mariadb`
[ ] `oracle`
[x] `postgres`
[ ] `cockroachdb`
[ ] `sqlite`
[ ] `sqljs`
[ ] `react-native`
[ ] `expo`

**TypeORM version:**

[x] `latest`
[ ] `@next`

1. Open your postgresql.conf
2. Set:

```conf
timezone = 'Europe/Berlin'
```

Screen Shot 2020-10-18 at 14 19 25

This tests fails if you have a different timezone for your local postgres instance.

TheNoim:
…s configured for postgres

With this change, you can run the test local without changing your timezone in the postgres config. Before this change, two tests would fail, because of a different timezone in postgres

Closes #6931

### Description of change

When I was working on #6469 I noticed I can't run all test local on my machine, because I had a different timezone configured for my postgres test instance. This is why I always only run the test relevant for me on my local machine and test the rest of the test I only run as part of the CI process. With this change, I am able to run all test on my local machine. 

### Pull-Request Checklist


test/functional/database-schema/column-types/postgres/column-types-postgres.ts
```diff
@@ -4,18 +4,38 @@ import {Connection} from "../../../../../src/connection/Connection";
 import {closeTestingConnections, createTestingConnections, reloadTestingDatabases} from "../../../../utils/test-utils";
 import {PostWithoutTypes} from "./entity/PostWithoutTypes";
 import {Post} from "./entity/Post";
+import {PostgresDriver} from "../../../../../src/driver/postgres/PostgresDriver";
 
 describe("database schema > column types > postgres", () => {
+    // Save the original set timezone for later restore
+    let originalTimezone: string;
```
pleerock, line 11: do we really need to "restore it"? Isn't `SET TIME ZONE` per-session configuration?

TheNoim: I thought it is set until restart of postgres

pleerock:
What if we make this connection-configuration-level option? (e.g. `timezone` option ormconfig / ConnectionOptions)

I think it's better to keep UTC in all places, and where specific timezone is needed, we can manually trigger SET TIME ZONE within `query` method.

TheNoim:
> What if we make this connection-configuration-level option? (e.g. `timezone` option ormconfig / ConnectionOptions)
> 
> I think it's better to keep UTC in all places, and where specific timezone is needed, we can manually trigger SET TIME ZONE within `query` method.

I think this is a breaking change, or isn't it? I mean, this change here only effects tests and make them work on a different postgres configuration. It is probably a hint for a deeper issue.

pleerock:
> I think this is a breaking change, or isn't it?

no, it's optional property and can be set by user. I think it can be helpful for some users.

> I mean, this change here only effects tests and make them work on a different postgres configuration.

right but I think users can have same issues.

Maybe if you can checkout exist issues you can find something related.

TheNoim:
> > I think this is a breaking change, or isn't it?
> 
> no, it's optional property and can be set by user. I think it can be helpful for some users.
> 
> > I mean, this change here only effects tests and make them work on a different postgres configuration.
> 
> right but I think users can have same issues.
> 
> Maybe if you can checkout exist issues you can find something related.

Hey, let's do a deal. This fixes the issue at least for testing right now. How about you merge it (if I remove this timezone restore if necessary) and I try to find the underlining issue later. If I fix it, I can revert the changes in the tests again. But at least it wouldn't be an issue for local testing anymore.

pleerock:
> Hey, let's do a deal. This fixes the issue at least for testing right now. How about you merge it (if I remove this timezone restore if necessary) and I try to find the underlining issue later. If I fix it, I can revert the changes in the tests again. But at least it wouldn't be an issue for local testing anymore.

right, it's better to have it rather not to have it. But you probably know nothing is more permanent than temporary. I strongly suggest to make it the way I proposed. If you disagree - I'll merge it.

test/functional/database-schema/column-types/postgres/column-types-postgres.ts
TheNoim: It is unnecessary. I removed it.

PR #8372, opened 2021-11-16
Fix documentations for migrations.

seivan:
Since the model claims `@PrimaryGeneratedColumn`, the migration should as well.

### Description of change
Documentation change for adding `isGenerated: true` to a migration that uses a schema with `@PrimaryGeneratedColumn`

### Pull-Request Checklist

- [ ] Code is up-to-date with the `master` branch **N/A**
- [ ] `npm run lint` passes with this change **N/A**
- [ ] `npm run test` passes with this change **N/A**
- [ ] This pull request links relevant issues as `Fixes #0000` **N/A**
- [ ] There are new or updated unit tests validating the change **N/A**

### DIFF
--- a/docs/migrations.md
+++ b/docs/migrations.md
@@ -243,7 +243,8 @@ export class QuestionRefactoringTIMESTAMP implements MigrationInterface {
                 {
                     name: "id",
                     type: "int",
-                    isPrimary: true
+                    isPrimary: true,
+                    isGenerated: true
                 },
                 {
                     name: "name",


pleerock:
what model? there is no model defined for `Question` in documentation examples.

seivan:
> what model? there is no model defined for `Question` in documentation examples.

Because you're using `Post` as your first example and introducing `Question` - which is a problem of its own in terms of documentation. 
At the very least, make it clear that if you have `PrimaryGeneratedColumn` you need both `isGenerated: true` and `isPrimary: true` -  this caught me off guard.

PR #8377, opened 2021-11-17
fix(column): make CreateDateColumn unable to be changed

rluvaton:
Before, the column with `CreateDateColumn` could have been overridden

### Description of change

Before, you could change the insert date later on, which is odd as the created time is the time the object was inserted, so disable the insertion by default 

### Pull-Request Checklist

### DIFF
--- a/src/decorator/columns/CreateDateColumn.ts
+++ b/src/decorator/columns/CreateDateColumn.ts
@@ -13,7 +13,7 @@ export function CreateDateColumn(options?: ColumnOptions): PropertyDecorator {
             target: object.constructor,
             propertyName: propertyName,
             mode: "createDate",
-            options: options || {}
+            options: { update: false, ...options }
         } as ColumnMetadataArgs);
     };
 }


pleerock:
Validity of this change is subject of discussions that we need to do. But change is a breaking anyway, so I have to reject it for now.

rluvaton:
Looking at the tests, this seems like the intended result, which is odd, as the insertion should not be able to be changed by default (of course you will be able to set  `update: true`, to enable this)

pleerock:
If you take a look what test is failing:

> create date column should set a custom date when specified

it says that current behaviour is by design. As I told I'm not sure if we should support it, since there are other problems arise. But we currently do it, and we have to keep it for now.

pleerock:
I added `design issue: create/update date columns` label. Checkout other issues with the same label to understand other problems related to this decorator.

rluvaton:
Thank you!

As I said in the previous comment
> Looking at the tests, this seems like the intended result

rluvaton:
@dpickett Can I add a note in the docs and in the decorator comment to emphasize that behavior?

Issue #8319, opened 2021-10-30
JSON string value will be parsed twice with mysql2


fenying:
## Issue Description

While using `@Column('json')` with `mysql2` package, it will parse the JSON twice, if the value of column is a string value in JSON format like `"text"`.

> See the hardcoded parser in mysql2: https://github.com/sidorares/node-mysql2/blob/247ce10a364f0dc42c0720ca42925890d5db07f0/lib/parsers/text_parser.js#L59-L63
>
> And where the problem is https://github.com/typeorm/typeorm/blob/master/src/driver/mysql/MysqlDriver.ts#L565-L566

**However, it's okay with `mysql` pakcage.**

### Expected Behavior

Get the parsed string value

### Actual Behavior

```
$ node dist/app.js 
undefined:1
hello
^

SyntaxError: Unexpected token h in JSON at position 0
    at JSON.parse (<anonymous>)
    at MysqlDriver.prepareHydratedValue (/home/angus/projects/fenying/typeorm-mysql2-json-bug/node_modules/typeorm/driver/mysql/MysqlDriver.js:488:54)
    at /home/angus/projects/fenying/typeorm-mysql2-json-bug/node_modules/typeorm/query-builder/transformer/RawSqlResultsToEntityTransformer.js:125:56
    at Array.forEach (<anonymous>)
    at RawSqlResultsToEntityTransformer.transformColumns (/home/angus/projects/fenying/typeorm-mysql2-json-bug/node_modules/typeorm/query-builder/transformer/RawSqlResultsToEntityTransformer.js:114:26)
    at RawSqlResultsToEntityTransformer.transformRawResultsGroup (/home/angus/projects/fenying/typeorm-mysql2-json-bug/node_modules/typeorm/query-builder/transformer/RawSqlResultsToEntityTransformer.js:94:31)
    at /home/angus/projects/fenying/typeorm-mysql2-json-bug/node_modules/typeorm/query-builder/transformer/RawSqlResultsToEntityTransformer.js:34:32
    at Map.forEach (<anonymous>)
    at RawSqlResultsToEntityTransformer.transform (/home/angus/projects/fenying/typeorm-mysql2-json-bug/node_modules/typeorm/query-builder/transformer/RawSqlResultsToEntityTransformer.js:33:15)
[...]

PR #8389, opened 2021-11-19
fix: mysql2 processes incorrectly with JSON column

fenying:
This PR does:
- fix: the JSON column with JSON encoded string value like `row.column1 = "text"` should not be parsed twice using driver `mysql2`.
- fix: the JSON column should be allowed to save `null` in JSON encoding, even if it's not a nullable column.
- feat: allow using `{ "driver": "mysql2" }` in connection options to select driver explicitly.

Closes: #8319

### Pull-Request Checklist

### DIFF
--- a/src/driver/mysql/MysqlDriver.ts
+++ b/src/driver/mysql/MysqlDriver.ts
@@ -43,6 +43,8 @@ export class MysqlDriver implements Driver {
      */
     mysql: any;
 
+    usingMySQL2: boolean = false;
+
     /**
      * Connection pool.
      * Used in non-replication mode.
@@ -517,8 +519,12 @@ export class MysqlDriver implements Driver {
         if (columnMetadata.transformer)
             value = ApplyValueTransformers.transformTo(columnMetadata.transformer, value);
 
-        if (value === null || value === undefined)
+        if (value === null || value === undefined) {
+            if (columnMetadata.type === "json") {
+                return "null";
+            }
             return value;
+        }
 
         if (columnMetadata.type === Boolean) {
             return value === true ? 1 : 0;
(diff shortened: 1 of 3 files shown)


src/driver/mysql/MysqlDriver.ts
```diff
@@ -907,6 +916,11 @@ export class MysqlDriver implements Driver {
      * Loads all driver dependencies.
      */
     protected loadDependencies(): void {
+        if (this.options.driver === "mysql2") {
```
pleerock, line 919: why this code block is needed? You already set `usingMySQL2` down below.

fenying: because I didn't find a way to select driver explicitly, for unit test, Unless I uninstall the `mysql` package.

pleerock: the same applies versa. If you have mysql2 installed (or both) and you want to test `mysql` explicitly - you can't.

pleerock: (because of the lines you added)

pleerock:
> feat: allow using { "driver": "mysql2" } in connection options to select driver explicitly.

why do we need it if we already pickup available driver?

src/driver/mysql/MysqlDriver.ts
fenying: sorry, I can't get your point. My idea is, If a connection is specified with `mysql2`, it will use it. Otherwise, it keeps use `mysql` first. And obviously you can setup two connection to test both `mysql` and `mysql2`.

pleerock:
> fix: the JSON column with JSON encoded string value like row.column1 = "text" should not be parsed twice using driver mysql2.

mysql2 claims to have same API as mysql. Is this behaviour documented anywhere?

src/driver/mysql/MysqlDriver.ts
pleerock: ah okay, I misunderstood your code

pleerock: Anyway, I think this is redundant for now. I would like to keep current loading behaviour as it is now.

pleerock:
Closing as there is no answer from author. Feel free to reopen it.

voinmerk:
Apparently, the merger request was rejected. Unfortunately. One of the fixes in this request could solve my problem.

##### fix: the JSON column with JSON encoded string value like row.column1 = "text" should not be parsed twice using driver mysql2.

PR #8412, opened 2021-11-29
fix: find() throws error while using mongodb

Corky3892:
Add support for mongodb 4.2

Closes: 8146

### Description of change

mongodb@4.2 does not export a Cursor, instead it exports a FindCursor. This change provides compatibility support to TypeORM so it works with both mongodb@3.6.4 and mongodb@4.2

Some notes on the tests:
- test/functional/commands/migration-create.ts
- test/functional/commands/migration-generate.ts
- test/functional/connection-options-reader/connection-options-reader.ts
- test/functional/persistence/delete-orphans/delete-orphans.ts

These tests were not passing prior to any changes and were disabled to complete the other tests. The remaining tests were ran successfully for both MongoDB version 4.4.10 using mongdb@4.2 driver and using MongoDB version 3.7 (using the docker image) and mongodb@3.6.4 driver.

### Pull-Request Checklist

- [X] There are new or updated unit tests validating the change - N/A
- [X] Documentation has been updated to reflect this change - N/A

### DIFF
--- a/src/entity-manager/MongoEntityManager.ts
+++ b/src/entity-manager/MongoEntityManager.ts
@@ -683,7 +683,8 @@ export class MongoEntityManager extends EntityManager {
      * Overrides cursor's toArray and next methods to convert results to entity automatically.
      */
     protected applyEntityTransformationToCursor<Entity>(metadata: EntityMetadata, cursor: Cursor<Entity> | AggregationCursor<Entity>) {
-        const ParentCursor = PlatformTools.load("mongodb").Cursor;
+        // mongdb-3.7 exports Cursor, mongodb-4.2 exports FindCursor, provide support for both.
+        const ParentCursor = PlatformTools.load("mongodb").Cursor || PlatformTools.load("mongodb").FindCursor;
         const queryRunner = this.mongoQueryRunner;
         cursor.toArray = function (callback?: MongoCallback<Entity[]>) {
             if (callback) {


pleerock:
Is it the only change that will make TypeORM to work with mongodb@4 ? I thought we should have plenty of other changes as well. For example this pr #7909

Corky3892:
@pleerock - I took a look through #7909. The first relevant piece of information to note that in version 4.2.1 of the mongodb driver the following has been included back as an export.

```typescript
import { ObjectId } from 'bson';
/**
 * @public
 * @deprecated Please use `ObjectId`
 */
export const ObjectID = ObjectId;
```

This would explain why much of the changes in #7909 have been omitted. Additionally this PR does not attempt to include those typings provided by the mongodb driver itself (*which from that discussion there maybe is not the worst thing*). Finally I also notice a change was made in #7909 to  `src/migration/MigrationExecutor.ts`, as mentioned the migration tests were note passing prior to any changes and so I did not consider them in the scope of the change.

-------------------

So to answer your question, yes this appears to be the only change necessary to make TypeORM work with MongoDB 4. This is not to say it offer full support, there may be (*in fact probably are*) features in MongoDB that are still not supported, it is probably more accurate to describe this as a compatibility change. Also note that I did not update the peer dependency to require 4.2.1, to be very clear this change enables the use cases:

__Test Results__

- mongodb@4.2.1 with MongoDB 4.4.10 > **Pass**
- mongodb@3.6.4 with MongoDB 3.7 > **Pass**
- mongodb@4.2.1 with MongoDB 3.7 > **Fail**
- mongodb@3.6.4 with MongoDB 4.4.10 > **Fail**

----------------

Please let me know if you feel any of these omissions should be included and I will take a look at them.

pleerock:
Thank you for your efforts. Let's merge it and keep tracking if users have other issues that we can resolve that won't break @3 version support.

Corky3892:
You bet! We are overdue a TS native approach for working with mongo, love what you are doing here please feel free to tag me on anything which looks related.

alegbebomby:
When can this be released on npm?

oktapodia:
@pleerock Hello, do you have an idea when that will be released?

Jmallone:
Hi, i can hardly wait for this release :)

pleerock:
Expect to have release this week.

Issue #8408, opened 2021-11-25
Allow soft-deletion of orphaned relation rows using orphanedRowAction


DivCorleone:
## Feature Description

Allow soft-deletion of orphaned relation rows using `orphanedRowAction`.

### The Problem

`orphanedRowAction` supports nullifying the relation rows or deleting them but not soft-deleting them.

### The Solution

Add a `soft-delete` option for `orphanedRowAction`.

### Relevant Database Driver(s)

| DB Type              | Relevant      |
| ---                  | ---           |
| `aurora-data-api`    | no            |
| `aurora-data-api-pg` | no            |
| `better-sqlite3`     | no            |
| `cockroachdb`        | no            |
| `cordova`            | no            |
| `expo`               | no            |
| `mongodb`            | no            |
| `mysql`              | yes            |
| `nativescript`       | no            |
| `oracle`             | no            |
| `postgres`           | no            |
| `react-native`       | no            |
| `sap`                | no            |
| `sqlite`             | no            |
| `sqlite-abstract`    | no            |
| `sqljs`              | no            |
| `sqlserver`          | no            |

### Are you willing to resolve this issue by submitting a Pull Request?

 - ✖️ Yes, I have the time, and I know how to start.
 - ✅ Yes, I have the time, but I don't know how to start. I would need guidance.
 - ✖️ No, I don’t have the time, but I can support (using donations) development.
 - ✖️ No, I don’t have the time and I’m okay to wait for the community / maintainers to resolve this issue.

oxeye-yuvalk:
Cool!
We also need this feature
I am working on a small feature to add a soft-delete option for orphanedRowAction

PR #8414, opened 2021-12-01
feat: Allow soft-deletion of orphaned relation rows using orphanedRow…

oxeye-yuvalk:
Allow soft-deletion of orphaned relation rows using orphanedRowAction
related to - https://github.com/typeorm/typeorm/issues/8408

### Description of change

The Problem
orphanedRowAction supports nullifying the relation rows or deleting them but not soft-deleting them.

The Solution
Add a soft-delete option for orphanedRowAction.

### Pull-Request Checklist

### DIFF
--- a/docs/relations.md
+++ b/docs/relations.md
@@ -25,7 +25,7 @@ There are several options you can specify for relations:
 * `onDelete: "RESTRICT"|"CASCADE"|"SET NULL"` - specifies how foreign key should behave when referenced object is deleted
 * `primary: boolean` - Indicates whether this relation's column will be a primary column or not.
 * `nullable: boolean` - Indicates whether this relation's column is nullable or not. By default it is nullable.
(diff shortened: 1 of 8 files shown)


pleerock:
@jbjhjm what do you think about this feature request?

jbjhjm:
@pleerock Makes sense to me. Adds another facet to making the orphanedRowAction feature complete.
Implementation is really easy so not much worrying it might break anything.

Tests look fine, but I'm missing a check validating that there actually is a soft-deleted Post in database.
As far as I can see current tests would pass if the row was hard-deleted from database too.
So I think the "should delete the orphaned Post from the database" test should become "should mark orphaned Post as soft-deleted" @oxeye-yuvalk -- do you agree?

oxeye-yuvalk:
@pleerock @jbjhjm I just changed the test file and all check have passed!

PR #8424, opened 2021-12-03
catch JSON parse error

mshen7310:
### Catch JSON.parse exception in MysqlDriver.prepareHydratedValue
  I saved a string in mysql json field, then got an exception when I tried to retrieve that string. The string fed to JSON.parse doesn't have double quote around it. It looks like the string had been parsed before reaching prepareHydratedValue. By catching the exception, the value will be returned when it is a string and can not be parsed by JSON.parse.

### Pull-Request Checklist

- [ X ] Code is up-to-date with the `master` branch
- [ X ] `npm run lint` passes with this change
- [ N/A ] Documentation has been updated to reflect this change
- [ X ] The new commits follow conventions explained in [CONTRIBUTING.md](https://github.com/typeorm/typeorm/blob/master/CONTRIBUTING.md)

### DIFF
--- a/src/driver/mysql/MysqlDriver.ts
+++ b/src/driver/mysql/MysqlDriver.ts
@@ -568,8 +568,13 @@ export class MysqlDriver implements Driver {
             value = DateUtils.mixedDateToDateString(value);
 
         } else if (columnMetadata.type === "json") {
-            value = typeof value === "string" ? JSON.parse(value) : value;
-
+            if(typeof value === "string"){
+                try{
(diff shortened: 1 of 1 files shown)


mshen7310:
Re-submit the PR to trigger the auto-test.

pleerock:
> It looks like the string had been parsed before reaching prepareHydratedValue. 

who did that parsing?

Need more info on actual source of the problem, current fix looks like a monkey-patching.

Also PR needs to have tests before it can be merged.

Issue #8420, opened 2021-12-03
Caching with Redis broken with Redis v4


aardvarkk:
## Issue Description

I receive the error "The client is closed" when running a cached query via Redis.

The trouble is that the new version (v4) of the Redis npm package does **not automatically connect anymore**. See [here for the migration guide](https://github.com/redis/node-redis/blob/master/docs/v3-to-v4.md). I think basically all that would need to change is to call `connect()` after `createClient()`, but I'm not sure what that would do if somebody was using v3 of Redis.

I tried passing `legacyMode: true` as an option via TypeORM's config, but I believe the autoconnect is still gone even in legacy mode.

### Actual Behavior

```
// include the output in code tags like these!
```

The cache should function correctly via Redis.

### Steps to Reproduce

1. Set up query caching to use Redis, with Redis npm package version >= 4
2. The connection fails

```typescript
// insert code here
```

### My Environment

| Dependency          | Version  |
| ---                 | ---      |
| Operating System    |          |
| Node.js version     | x.y.zzz  | 
| Typescript version  | x.y.zzz  | 
| TypeORM version     | x.y.zzz  | 

### Relevant Database Driver(s)

| DB Type              | Reproducible  |
| ---                  | ---           |
| `aurora-data-api`    | no            |
| `aurora-data-api-pg` | no            |
| `better-sqlite3`     | no            |
| `cockroachdb`        | no            |
| `cordova`            | no            |
| `expo`               | no            |
| `mongodb`            | no            |
| `mysql`              | no            |
| `nativescript`       | no            |
| `oracle`             | no            |
| `postgres`           | no            |
| `react-native`       | no            |
| `sap`                | no            |
| `sqlite`             | no            |
| `sqlite-abstract`    | no            |
| `sqljs`              | no            |
| `sqlserver`          | no            |

### Are you willing to resolve this issue by submitting a Pull Request?

 - ✖️ Yes, I have the time, and I know how to start.
 - ✅ Yes, I have the time, but I don't know how to start. I would need guidance.
 - ✖️ No, I don’t have the time, but I can support (using donations) development.
 - ✖️ No, I don’t have the time and I’m okay to wait for the community / maintainers to resolve this issue.

aardvarkk:
I've written up a PR that fixes this for me in my local environment. Basically we just need to do two things to get it working:

- Call `connect()` on the client
- Pass `legacyMode: true` in the client options so that the existing code doesn't have to change

[...]

PR #8425, opened 2021-12-03
Add support for node-redis v4.0.0 and newer

aardvarkk:
- Fixes https://github.com/typeorm/typeorm/issues/8420
- v4.0.0 and newer of node-redis will no longer auto-connect, so we need to add that step
- For backwards compatibility, we check for the existence of a "v4" property so we don't try to call `connect` on older versions of the client
- A `legacyMode` flag exists in the new v4.0.0 of node-redis to keep a similar calling style without having to change any code
- This option can be passed via the createClient options, so I've added documentation to indicate that this is required if users would like to use v4.0.0 of node-redis
- Note that the legacyMode flag is **required** to be passed if the calling code isn't going to be changed

### Pull-Request Checklist

### DIFF
--- a/docs/caching.md
+++ b/docs/caching.md
@@ -149,8 +149,7 @@ Example:
 }
 ```
 
-"options" can be [node_redis specific options](https://github.com/NodeRedis/node_redis#options-object-properties) or [ioredis specific options](https://github.com/luin/ioredis/blob/master/API.md#new-redisport-host-options) depending on what type you're using.
-
+"options" can be [node_redis specific options](https://github.com/redis/node-redis/blob/master/docs/client-configuration.md) or [ioredis specific options](https://github.com/luin/ioredis/blob/master/API.md#new-redisport-host-options) depending on what type you're using.
 
 In case you want to connect to a redis-cluster using IORedis's cluster functionality, you can do that as well by doing the following:
 
--- a/src/cache/RedisQueryResultCache.ts
+++ b/src/cache/RedisQueryResultCache.ts
@@ -50,10 +50,12 @@ export class RedisQueryResultCache implements QueryResultCache {
     async connect(): Promise<void> {
         const cacheOptions: any = this.connection.options.cache;
         if (this.clientType === "redis") {
-            if (cacheOptions && cacheOptions.options) {
-                this.client = this.redis.createClient(cacheOptions.options);
-            } else {
-                this.client = this.redis.createClient();
+            this.client = this.redis.createClient({
+                ...cacheOptions?.options,
+                legacyMode: true
+            });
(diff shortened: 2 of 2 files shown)


aardvarkk:
I'm realizing another thing that could be done here is to force `legacyMode` to be added on `createClient` calls if it's v4 -- that way we wouldn't need to tell/remind users in the docs that they also have to pass this option to get it working properly. But it would require a slightly more aggressive patch, so I opted for the most minimal change at this point.

pleerock:
> But it would require a slightly more aggressive patch, so I opted for the most minimal change at this point.

What are the downsides of this aggressive approach?  Does it bring any user-facing issues? If no, we can simplify user experience.

aardvarkk:
@pleerock I see there being at least four possible ways to fix this:

1. What I've done so far in this PR, which requires the smallest change in the TypeORM code but requires the user to pass a legacyMode flag.
2. Automatically passing `legacyMode` as an option if redis v4 is detected. This would require a slightly heavier-handed change to TypeORM, but would probably eliminate a class of errors where users forget to pass `legacyMode` themselves and wonder why it's not working.
3. Rewriting the TypeORM code to be promise-based and require redis v4. This would eliminate the need to pass any `legacyMode` flag. But it would probably cause a lot of problems for people who are still using/dependent upon v3 of redis if they're also using the package themselves.
4. Split the Redis caching implementation in two: one promise-based and the other callback-based. Both `node-redis` and `ioredis` support both styles, I believe. This would be the most flexible in that would still support v3 and v4 without having to pass any additional flags, but would require a fair amount of new code and also add more "surface area" to the Redis caching implementation as possible sources of bugs.

I'm happy to adjust this PR to move from implementation #1 to #2 if you believe that approach to be superior. I don't believe it would impact users directly and would remove the need for them to pass `legacyMode` themselves.

aardvarkk:
I've updated the code to obviate the need for a client to pass `legacyMode` themselves, but haven't tested it out locally yet.

aardvarkk:
@pleerock I've done my local testing and checked that:

1. `legacyMode: true` is ignored by v3 and so it has no effect to pass regardless of the version
2. We only try to call `connect()` if the property exists within the Redis client, which it doesn't prior to v4. I liked this approach better than looking for "v4" as I was doing previously, because presumably what I have now could extend to a v5 if that ever happened.
3. This setup appears to work with and without specific cache options being passed both in v3 and v4.

Please let me know if there's anything else you need in regard to this PR.

pleerock:
Thank you for contribution! 🎉

meronokbay:
@pleerock I need to update `node-redis` to v4 on a production application and I was wondering when this will fix will be released to npm.

Issue #8430, opened 2021-12-05
sqlite temporary tables do not honor withoutRowid


pthrasher:
## Issue Description

sqlite migrations must re-create tables when doing most changes. These temporary tables don't honor `withoutRowid`

### Expected Behavior

I expect the new `temporary_` to honor any special options that exist for sqlite in the `EntityMetadata`

### Actual Behavior

Original tables are created properly with `WITHOUT ROWID` but the new temporary tables don't receive this info because there's no table in `EntityMetadata` that matches this temporary prefixed name.

### Steps to Reproduce

1. Create `Entity` with `withoutRowid: true` as an option.
2. Attempt to generate migration.
3. Notice that none of the temporary table create statements have a trailing "WITHOUT ROWID" even though the original create table statements do.

### Relevant Database Driver(s)

| DB Type              | Reproducible  |
| ---                  | ---           |
| `aurora-data-api`    | no            |
| `aurora-data-api-pg` | no            |
| `better-sqlite3`     | no            |
| `cockroachdb`        | no            |
| `cordova`            | no            |
| `expo`               | no            |
| `mongodb`            | no            |
| `mysql`              | no            |
| `nativescript`       | no            |
| `oracle`             | no            |
| `postgres`           | no            |
| `react-native`       | no            |
| `sap`                | no            |
| `sqlite`             | yes            |
| `sqlite-abstract`    | yes            |
| `sqljs`              | no            |
| `sqlserver`          | no            |

### Are you willing to resolve this issue by submitting a Pull Request?

 - ✅ Yes, I have the time, and I know how to start.
 - ✖️ Yes, I have the time, but I don't know how to start. I would need guidance.
 - ✖️ No, I don’t have the time, but I can support (using donations) development.
 - ✖️ No, I don’t have the time and I’m okay to wait for the community / maintainers to resolve this issue.

PR is complete. Just running tests, will submit shortly. :-)

PR #8431, opened 2021-12-05
fix: sqlite temporary tables now honor withoutRowid

pthrasher:
### Description of change

sqlite requries complete table re-creation for most migration operations. When this happens, there's a new table created named `temporary_<old table name>` -- This copies all of the table information, but previously failed to grab `EntityMetadata` for this new temporary table since the name doesn't exists amongst all entity metadatas.

(I also fixed lint errors inside the file I touched)

This PR fixes this problem by checking again if we found no match the first time.

Fixes: #8430 

### Pull-Request Checklist

- [X] There are new or updated unit tests validating the change **N/A**
- [X] Documentation has been updated to reflect this change **N/A**


pleerock:
`createTableSql` creates table based on a given `table: Table` property provided into function parameters. 

The correct way of fixing this issue (as well as fixing original change) is to use `table.withoutRowid` property.
This property is missing now in the Table class, but it needs to be added. 
In order for this property to be set in the Table, it needs to be copied from EntityMetadata on Table object instance creation.
This way all other EntityMetadata -> Table properties are working, and we should seek to this pattern.

pthrasher:
@pleerock Good to know! I will alter my change to work that way.

AlexMesser:
thank you for contribution!

Issue #8416, opened 2021-12-01
The cascade soft-delete would update the related entities recursively and automatically


oxeye-yuvalk:
## Feature Description

The soft-delete would update recursively the related entities without loading them

related to - https://github.com/typeorm/typeorm/issues/5877

### The Problem

Today, the cascade soft delete update the related entities just if we would load all the entities that we want to soft-remove
This behavior is bad in term of performance and in addition, don't work recursively
Because there is no feature like that today, a lot of developers write soft delete flow for every entity, and that causes a lot of bugs

### The Solution

Change the soft-remove flow to soft-remove all related entities recursively and automatically

### Other option

Add a flag to the soft-remove function that declares that the operation should do the new logic instead of the old one

### Relevant Database Driver(s)

| DB Type              | Relevant      |
| ---                  | ---           |
| `aurora-data-api`    | yes            |
| `aurora-data-api-pg` | yes            |
| `better-sqlite3`     | yes            |
| `cockroachdb`        | yes            |
| `cordova`            | yes            |
| `expo`               | yes            |
| `mongodb`            | yes            |
| `mysql`              | yes            |
| `nativescript`       | yes            |
| `oracle`             | yes            |
| `postgres`           | yes            |
| `react-native`       | yes            |
| `sap`                | yes            |
| `sqlite`             | yes            |
| `sqlite-abstract`    | yes            |
| `sqljs`              | yes            |
| `sqlserver`          | yes            |

### Are you willing to resolve this issue by submitting a Pull Request?

 - ✅ Yes, I have the time, and I know how to start.
 - ✖️ Yes, I have the time, but I don't know how to start. I would need guidance.
 - ✖️ No, I don’t have the time, but I can support (using donations) development.
 - ✖️ No, I don’t have the time and I’m okay to wait for the community / maintainers to resolve this issue.

PR #8436, opened 2021-12-07
Soft delete recursive cascade

oxeye-yuvalk:
The soft-delete would update recursively the related entities without loading them

related to - https://github.com/typeorm/typeorm/issues/8416

### Description of change

The Problem
Today, the cascade soft delete update the related entities just if we would load all the entities that we want to soft-remove
This behavior is bad in term of performance and in addition, don't work recursively
Because there is no feature like that today, a lot of developers write soft delete flow for every entity that includes a lot of code because they need to kill the children as well, and that causes a lot of bugs

The Solution
Change the soft-remove flow to soft-remove all related entities recursively and automatically
- soft remove the entity
- get the entity relations and soft remove the inverse entities where the parent id is the first entity id
- move recursively throw the relations and soft-delete by the parent ids

I prefer to soft remove without loading all the entities and create subject to each of them because it's more efficient

Other option
Add a flag to the soft-remove function that declares that the operation should do the new logic instead of the old one

### Pull-Request Checklist

### DIFF
--- a/src/persistence/SubjectExecutor.ts
+++ b/src/persistence/SubjectExecutor.ts
@@ -647,7 +647,6 @@ export class SubjectExecutor {
         if (relation.isCascadeSoftRemove){
             let primaryPropertyName = relation.inverseEntityMetadata.primaryColumns[0].propertyName;
             let updateResult: UpdateResult;
-            let parentIds: any[] = [];
             let softDeleteQueryBuilder = this.queryRunner
                 .manager
                 .createQueryBuilder()
@@ -658,9 +657,7 @@ export class SubjectExecutor {
                 .callListeners(false);
             softDeleteQueryBuilder.where(`${relation.inverseSidePropertyPath} in (:...ids)`, {ids: ids});
             updateResult = await softDeleteQueryBuilder.execute();
-            for (const row of updateResult.raw) {
-                parentIds.push(row[Object.keys(row)[0]]);
-            }
+            const parentIds = updateResult.raw.map((row: any) => row[Object.keys(row)[0]]);
             if (parentIds.length) {
                 for (const subRelation of relation.inverseEntityMetadata.relations) {
                     await this.executeSoftRemoveRecursive(subRelation, parentIds);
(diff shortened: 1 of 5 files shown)


pleerock:
I'm really against cascades further complication, but let's merge it for now, later on we need either simplify the codebase, or find a maintainer who is going to deal with constant cascade issues.

pleerock:
I had to revert these changes because of failing CI. I missed few points in this PR, I'll left comments.

src/persistence/SubjectExecutor.ts
```diff
@@ -635,10 +640,45 @@ export class SubjectExecutor {
...
+
+            let primaryPropertyName = relation.inverseEntityMetadata.primaryColumns[0].propertyName;
+            let updateResult: UpdateResult;
+            let softDeleteQueryBuilder = this.queryRunner
+                .manager
+                .createQueryBuilder()
+                .softDelete()
+                .from(relation.inverseEntityMetadata.target)
+                // We get back list of the affected rows primary keys for call again
+                .returning([primaryPropertyName])
```
pleerock, line 659: `.returning` isn't supported by all databases, for example `sqlite`.

src/persistence/SubjectExecutor.ts
```diff
@@ -607,8 +608,12 @@ export class SubjectExecutor {
                 } else { // in this case identifier is just conditions object to update by
                     softDeleteQueryBuilder.where(subject.identifier);
                 }
-
                 updateResult = await softDeleteQueryBuilder.execute();
+                // Move throw all the relation of the subject
+                for (const relation of subject.metadata.relations) {
+                    // Call recursive function that get the parents primary keys that in used on the inverse side in all one to many relations
+                    await this.executeSoftRemoveRecursive(relation, [Reflect.get(subject.identifier, subject.metadata.primaryColumns[0].propertyName)]);
```
pleerock, line 615: not really clear why `Reflect.get` is used here. Decorators isn't the only source of entity definitions and some people define entities using entity schemas.

Issue #8267, opened 2021-10-11
memory leak since 0.2.35


chriswep:
After updating typeorm from 0.2.34 my server started running quickly out of memory, at a rate of about 1Gig every few seconds. this happens with all versions >= 0.2.35, downgrading to 0.2.34 is fine. I tried to research what might be the cause however 0.2.35 introduced quite a number of changes, so i'm a bit lost here. Just shooting in the blue, https://github.com/typeorm/typeorm/commit/f43d56110fd41c23d80e32021adf4ade7648ce97#diff-3008128203dc904e0890ee2cb2ab47080b93ea391288f4acb8f0ecbf3c2dde8a looks like it potentially could be related to what is happening, just guessing.

Anyone else is seeing something similar? What change in 0.2.35 could be responsible?

imnotjames:
If you think it's that commit, revert it and test without it.

Do you have a reproducible example of this?

chriswep:
i can't reproduce it in my dev environment. I tried to put load on it with similar queries / data however the result objects don't even appear in the dump. the commit was just a guess since it's about how objects are created.. compiling different versions of typeorm by guessing and testing them on production is more like a last resort. i thought maybe somebody has seen something similar or has an idea where to look.

mmyoji:
I encountered the same issue. 

Upgrading from v0.2.34 to v0.2.35, our project CI test starts failing due to out of memory.
We use PostgreSQL and just need this commit https://github.com/typeorm/typeorm/commit/5cf368a23fa78b9e97dd12b54616f17b8431ffee.
Then I forked this project, checkout the commit, prepare built version of package, ran the test, and the memory leak happened again.

So the problem commit(s) might be in https://github.com/typeorm/typeorm/compare/0.2.34...5cf368a23fa78b9e97dd12b54616f17b8431ffee.

As @chriswep says, https://github.com/typeorm/typeorm/commit/f43d56110fd41c23d80e32021adf4ade7648ce97#diff-3008128203dc904e0890ee2cb2ab47080b93ea391288f4acb8f0ecbf3c2dde8a could cause the memory leak.

I just let you know this although I haven't investigated this issue deeply.

chriswep:
@mmyoji thanks for the info. could you try reversing the commit and try if that fixes it? i haven't gotten around to it yet since i've settled on 0.2.34 for now.

hugo-marello-nutrien:
[...]

PR #8445, opened 2021-12-10
fix: make EntityMetadataValidator comply with entitySkipConstructor

### LINKED ISSUE #8444: entitySkipConstructor not working
## `entitySkipConstructor` not working as expected/documented

See https://github.com/typeorm/typeorm/pull/7770, https://github.com/typeorm/typeorm/issues/1772

### Expected Behavior

By setting `entitySkipConstructor` connection option, one is able to define an Entity class with strict constructor property initialization, even using object-type destructuring, e.g.:
```
type SomeEntityConstructorArgs = {
  readonly foo: string;
};

@Entity()
export class SomeEntity {

  @Column()
  readonly foo: string;

  constructor(args: SomeEntityConstructorArgs) {
    const { foo } = args;

    this.foo = foo;
  }
}
```

### Actual Behavior

Running `migration:run`, or bootstrapping a NestJS app leveraging TypeORM, breaks with:
```
TypeError: Cannot destructure property 'foo' of 'args' as it is undefined.
    at null.SomeEntity (/app/src/entity/SomeEntity.ts:57:13)
    at EntityMetadata.EntityMetadata.create (/app/node_modules/typeorm/metadata/src/metadata/EntityMetadata.ts:534:23)
    at EntityMetadataValidator.EntityMetadataValidator.validate (/app/node_modules/typeorm/metadata-builder/src/metadata-builder/EntityMetadataValidator.ts:118:47)
    at null.<anonymous> (/app/node_modules/typeorm/metadata-builder/src/metadata-builder/EntityMetadataValidator.ts:46:56)
```

### Steps to Reproduce

(see above): simply define an Entity constructor with object destructuring assignment (I actually expect a failure even with strict `undefined` checks and scalar arguments).

### My Environment

| Node.js version     | v16.13.1  | 
| nestjs/typeorm | 8.0.2 |

Was reproduced with a MySQL connection, but I believe the error is unrelated to the driver in use.

### Root cause / Possible fix

Locally I applied the patch below which seems to have solved the issue. Perhaps this point in code was an oversight? I see in fact than in most if not all instances where `EntityMetadata.create` is invoked, `{ fromDeserializer: true }` is passed as well.

```
diff --git a/node_modules/typeorm/metadata-builder/EntityMetadataValidator.js b/node_modules/typeorm/metadata-builder/EntityMetadataValidator.js
index 9bb557e..d454d4b 100644
--- a/node_modules/typeorm/metadata-builder/EntityMetadataValidator.js
+++ b/node_modules/typeorm/metadata-builder/EntityMetadataValidator.js
@@ -102,7 +102,7 @@ var EntityMetadataValidator = /** @class */ (function () {
                 throw new error_1.TypeORMError("Character set specifying is not supported in Sql Server");
         }
         // check if relations are all without initialized properties
-        var entityInstance = entityMetadata.create();
+        var entityInstance = entityMetadata.create(undefined, { fromDeserializer: true });
         entityMetadata.relations.forEach(function (relation) {
             if (relation.isManyToMany || relation.isOneToMany) {
                 // we skip relations for which persistence is disabled since initialization in them cannot harm somehow

```

francescosalvi:
Fixes #8444

### Description of change

See related issue https://github.com/typeorm/typeorm/issues/8444

### Pull-Request Checklist

- [x] The new commits follow conventions explained in [CONTRIBUTING.md]

### DIFF
--- a/src/metadata-builder/EntityMetadataValidator.ts
+++ b/src/metadata-builder/EntityMetadataValidator.ts
@@ -115,7 +115,7 @@ export class EntityMetadataValidator {
         }
 
         // check if relations are all without initialized properties
-        const entityInstance = entityMetadata.create();
+        const entityInstance = entityMetadata.create(undefined, { fromDeserializer: true });
         entityMetadata.relations.forEach(relation => {
             if (relation.isManyToMany || relation.isOneToMany) {
 
--- a/test/github-issues/8444/entity/StrictlyInitializedEntity.ts
+++ b/test/github-issues/8444/entity/StrictlyInitializedEntity.ts
@@ -1,0 +1,18 @@
+import {Column, Entity, PrimaryGeneratedColumn} from "../../../../src";
+
+@Entity()
+export class StrictlyInitializedEntity {
+    @PrimaryGeneratedColumn()
+    id?: number;
+
+    @Column()
+    readonly someColumn: string;
+
+    constructor(someColumn: string) {
+        if (someColumn === undefined) {
+            throw new Error("someColumn cannot be undefined.");
+        }
+
+        this.someColumn = someColumn;
+    }
+}
--- a/test/github-issues/8444/issue-8444.ts
+++ b/test/github-issues/8444/issue-8444.ts
@@ -1,0 +1,45 @@
+import "reflect-metadata";
+import { expect } from "chai";
+import {closeTestingConnections, createTestingConnections} from "../../utils/test-utils";
+import {StrictlyInitializedEntity} from "./entity/StrictlyInitializedEntity";
+import { Connection } from "../../../src/connection/Connection";
+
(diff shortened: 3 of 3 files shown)


pleerock:
Can we have this change test-covered, so we can make sure in the future this won't break? Thank you

francescosalvi:
@pleerock ok, I'll try

francescosalvi:
@pleerock test added: I basically asserted that with and without the connection option, and the given entity, the connection should/should not be established (initially I defined an additional entity with the destructuring assignment in the constructor, but the error message seemed to differ among different Node versions, so I removed and only kept the current test entity where I assert that the passed parameter is not undefined)

pleerock:
I'm thinking to revert `entitySkipConstructor` feature at all because of the following issue: #8267

pleerock:
okay, let's merge it for now

Issue #2196, opened 2018-05-22
How to enable sync schema and not drop columns?


wy193777:
**Issue type:**

[x] question
[ ] bug report
[ ] feature request
[ ] documentation issue

**Database system/driver:**

[ ] `cordova`
[ ] `mongodb`
[x] `mssql`
[ ] `mysql` / `mariadb`
[ ] `oracle`
[ ] `postgres`
[ ] `sqlite`
[ ] `sqljs`
[ ] `react-native`

**TypeORM version:**

[x] `latest`
[ ] `@next`
[ ] `0.x.x` (or put your version here)

When using SQL Server, I use a not supported column type: rowversion. I guess typeorm has no way to define this column type. So I have a SQL script add this type if they are not been added. But when initialize typeorm, this column will be dropped if set sync to true.

Is there an option about not drop columns but only add columns when initializing?

pleerock:
no, there is no. You can disable synchronization for the whole entity by setting `@Entity({ synchronization: false })`. Otherwise you can contribute and add rowversion support into typeorm.

wy193777:
Created pull request #2198. Still working in progress.

wy193777:
Is the new version with #2198 been published to NVM? Got error when try to use rowversion type today.

pleerock:
no, I don't think so. I'll publish a new version these weekends.

wy193777:
How can I install typeorm using NPM from GitHub? Always got this error:
```npm ERR! path ./node_modules/typeorm/cli.js
npm ERR! code ENOENT
npm ERR! errno -2
npm ERR! syscall chmod
npm ERR! enoent ENOENT: no such file or directory, chmod './node_modules/typeorm/cli.js'
npm ERR! enoent This is related to npm not being able to find a file.
npm ERR! enoent 

npm ERR! A complete log of this run can be found in:
npm ERR!     ~/.npm/_logs/2018-06-02T01_08_08_395Z-debug.log
```

apiel:
Any chance to deactivate synchronization for a specific column?

ArturoGamRod:
it is very easy to get this wrong and/or make a mistake. synchronization should not drop columns - **especially with existing  data** -  its a deal breaker guys.

skliarovartem:
Yes, it is strange to delete columns with some data there. Usually sync works to add new columns only, and even don't change type

Huge-Puppy:
[...]

PR #8465, opened 2021-12-15
new flag to configure synchronize schema to not drop columns and tables

oxeye-yuvalk:
### Description of change

Add synchronizeWithoutDrops option that set synchronization to not drop columns and tables but only add columns and tables when initializing

related to - https://github.com/typeorm/typeorm/issues/2196

It is very easy to get this wrong and/or make a mistake. synchronization should not drop columns - especially with existing data

### Pull-Request Checklist

### DIFF
--- a/docs/connection-options.md
+++ b/docs/connection-options.md
@@ -15,7 +15,7 @@
 * [`sql.js` connection options](#sqljs-connection-options)
 * [`expo` connection options](#expo-connection-options)
 * [Connection options example](#connection-options-example)
-    
+
 ## What is `ConnectionOptions`
 
 Connection options is a connection configuration you pass to `createConnection`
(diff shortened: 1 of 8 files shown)


pleerock:
I'm very pessimistic about this PR. What is the real purpose of this feature?

oxeye-yuvalk:
The purpose is to have an option to run with synchronized option in production and not worry about losing data. For example, I don’t want that if I would remove column from entity, it will remove the whole column with the data. In many ORM’s, synchronize only add columns and meta data and not remove things because it’s very dangerous. This feature would be great for using the powerful of synchronize in production

imnotjames:
**Even with this change, the synchronizing feature will destroy your data, without warning, and without a means to recover.**

I really, really don't think synchronized should be used in production.

It's a development tool, at best.  (I actually think it should be dropped as a feature entirely, but hey, it's tough to remove features..)

It provides no real way to roll back the changes and is not atomic nor is it reproducible.

AlexMesser:
> I don’t want that if I would remove column from entity, it will remove the whole column with the data

Synchronization does exactly what it says - synchronizes your entity with the database table.

In production, you should use migrations instead of synchronization and carefully check the generated migration line by line to prevent data loss.  

So, I agree with @pleerock and @imnotjames - this change makes no sense

Issue #6779, opened 2020-09-24
@BeforeInsert / @BeforeUpdate function blocks all database calls when the connectionpool was temporarily full


yorickdevries:
**Issue type:**

[x] bug report
[ ] feature request
[ ] documentation issue

**Database system/driver:**

[ ] `cordova`
[ ] `mongodb`
[ ] `mssql`
[x] `mysql` / `mariadb`
[ ] `oracle`
[ ] `postgres`
[ ] `cockroachdb`
[ ] `sqlite`
[ ] `sqljs`
[ ] `react-native`
[ ] `expo`

**TypeORM version:**

[x] `latest`
[ ] `@next`
[ ] `0.x.x` (or put your version here)

**Steps to reproduce or a small repository showing the problem:**

In my production environment there is sometimes more load and queries are being queued to be processed later (as expected). However, I observed that once the connections of the connection pool become available again, the queries in the queue are not run anymore and the connections stay asleep. the only remedy in that case is rebooting the server, which causes everything to work again.

I discovered that the `@BeforeInsert` / `@BeforeUpdate` function I use for validating my database entries trigger this behavior. In that function I perform some asynchronous database queries as well to ensure validity, similar to https://github.com/typeorm/typeorm/issues/913 ). Perhaps when this check takes too long the system blocks? Removing the `@BeforeInsert` / `@BeforeUpdate` decorators cause the block not to happen.

As this bug only occurs at heavy load, I made a mockup repository which triggers this bug, so one could reproduce this behavior. you can find it at: https://github.com/yorickdevries/typeorm-before-bug . The code in this repo keeps on creating/updating users until a queue of a certain length arises and then stops creation/updating. I observe that after stopping creating/updating users the connections finish their queries and go to sleep mode. The queue remains full and no new queries are processed at all anymore, and the mysql driver also indicates that there are no free connections available.

The behavior is similar to https://github.com/mysqljs/mysql/issues/1771 , but I could not reproduce it without the TypeORM `@BeforeInsert` / `@BeforeUpdate` decorators.

See also the screenshot below:
![image](https://user-images.githubusercontent.com/23526707/94166589-c8250500-fe8b-11ea-88c1-dfbd97ecdcc9.png)
This is what you can reproduce with the code in the repository.

The `userEntity` I made for this repo is shown here as well:

```ts
import {
  Entity,
  Column,
  BaseEntity,
  BeforeInsert,
  BeforeUpdate,
  PrimaryGeneratedColumn,
[...]

Issue #319, opened 2017-02-27
Entity Listeners event parameter


TheKiteEatingTree:
Is there any reason entity listeners are not passed an event object like subscribers are?  I get that some of the properties are useless in an entity listener, but at least the entityManager would be nice to have unless there's some way to access it that I'm missing.  I could work on a PR for this if it makes sense.

pleerock:
The reason behind this is the idea of typeorm and data mapper - your models should be dummy. By providing entity manager to your models you are making them _complex_ - now your entities aren't just dummy logic, they also do a complicated logic and operations using entity manager. I'm not sure yet if we should go this way.

TheKiteEatingTree:
That explanation makes sense to me.  Thanks!

pleerock:
Lets close it for now then

PR #8497, opened 2022-01-02
docs: update listeners/subscribers documentation

DanArpoika:
Add documentation on base properties for subscriber's event object.
Add note to listeners about not running database calls.

### Description of change

I have seen people incorrectly use database calls inside listeners, including myself, leading to unresponsiveness. This is because the original QueryRunner used in the event that triggers the listeners is not released until after the broadcasting methods are finished.

This might be obvious to some, but I didn't realize this until after exploring the TypeORM code base further. This PR updates the documentation to add a note not to make database calls in listeners and instead opt for subscribers. The subscriber event methods all receive the event object that contains the original QueryRunner/EntityManager instance which when used  solves the race condition issue. 

The reason listeners don't get the event object is explained here: https://github.com/typeorm/typeorm/issues/319

Example of issue: (an express api)

- Your application has a connection limit of 10. 
- Route A uses QueryBuilder to load a single Post entity that has a listener which makes a database call.

You put Route A under a heavy load and there is a race condition to when 10 connections are being used to load the Post entity, each then trying to call the listener which attempts to fetch a new connection from the pool but there is none available, thus putting TypeORM/the application into an unresponsive state.

I am only familiar with MySQL and I don't believe there is a timeout for connection requests in the queue, so other drivers might solve this issue (although a timeout error doesn't seem like a solution). 

Related issue: https://github.com/typeorm/typeorm/issues/6779

### Pull-Request Checklist

- [] `npm run lint` passes with this change
- [] `npm run test` passes with this change
- [] This pull request links relevant issues as `Fixes #0000`
- [] There are new or updated unit tests validating the change

### DIFF
--- a/docs/listeners-and-subscribers.md
+++ b/docs/listeners-and-subscribers.md
@@ -9,11 +9,14 @@
     * [`@BeforeRemove`](#beforeremove)
     * [`@AfterRemove`](#afterremove)
 * [What is a Subscriber](#what-is-a-subscriber)
+    * [`Event Object`](#event-object)
 
 ## What is an Entity Listener
 
 Any of your entities can have methods with custom logic that listen to specific entity events.
 You must mark those methods with special decorators depending on what event you want to listen to.
+
+**Note:** Do not make any database calls within a listener, opt for [subscribers](#what-is-a-subscriber) instead.
 
 ### `@AfterLoad`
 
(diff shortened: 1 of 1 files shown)


pleerock:
Thank you, its a great addition to the docs!

Issue #4764, opened 2019-09-17
Extend dirty_read and possibly hints in mssql joins


alfonsoal1983:
**Issue type:**

[x] question
[ ] bug report
[x] feature request
[ ] documentation issue

**Database system/driver:**

[ ] `cordova`
[ ] `mongodb`
[x] `mssql`
[ ] `mysql` / `mariadb`
[ ] `oracle`
[ ] `postgres`
[ ] `cockroachdb`
[ ] `sqlite`
[ ] `sqljs`
[ ] `react-native`
[ ] `expo`

**TypeORM version:**

[x] `latest`
[ ] `@next`
[ ] `0.x.x` (or put your version here)

**Steps to reproduce or a small repository showing the problem:**

Hello, when making a simple dirty_read works fine, the problem comes when I have to extend dirty_read to a query with joins, is there any way to make dirty_read spread to joins?

If not. Can I implement a function that meets that functionality? I have tried it locally and it works quite well, in the same way, you could use that function to apply more mssql hints, the only drawback is that it would only apply to mssql which is the one that uses this nomenclature.

<!--
    To answer those questions you need to put "x" inside the square brackets, for example:
    [x] `mysql`
    [ ] `postgres`

    Also, please format your code properly (by taking code blocks into ```ts .... ```)
!>

Hello, when making a simple query dirty_read works fine:

```
getRepository(Car).createQueryBuilder('cars').setLock('dirty_read').getOne()
```

This would result in:

**SELECT cars.* FROM CAR cars WITH(NOLOCK)**

The problem comes when I have to extend dirty_read to a query with joins. Example:

```
getRepository(Car).createQueryBuilder('cars').setLock('dirty_read')
.innerJoin('car.tires', 'tires')
.getOne()
```
This would result in:

**SELECT cars.*, tires.* FROM CAR car WITH(NOLOCK) **
**INNER JOIN TIRE tires ON tires.ID = cars.TIREID**

The dirty_read will only apply to the main from so that if there is any open transaction that affects the join table, the query will be waiting for the transaction, the idea is to put the dirty_read in the join.

Is there any way to make dirty_read extend to joins?

If not. Can I implement a function that meets that functionality? I have tried it locally and it works quite well, in the same way, you could use that function to apply more mssql hints (https://docs.microsoft.com/en-us/sql/t-sql/queries/hints-transact-sql-table?view=sql-server-2017), something like this (it's just an incomplete example, the final code would be more readable):

[...]

PR #8507, opened 2022-01-05
fix: add WITH (lock) clause for MSSQL select with join queries

iceCreamParlor:
Typeorm didn't supported LOCK clause in MSSQL SELECT + JOIN query. 

For example, when we write code with typeorm like this
```typescript
await getConnection()
        .createQueryBuilder()
        .select("order")
        .setLock("dirty_read")
        .from(Order, "order")
        .innerJoinAndSelect("order", "order.OrderItems")
        .where("1=1")
        .getMany()
```
we would expect the corresponding SQL of this statement is 
```sql
SELECT      *
FROM        Order O     WITH (NOLOCK)
LEFT JOIN   OrderItems  WITH (NOLOCK)
WHERE       1=1
```
but typeorm generates SQL like this 
```sql
SELECT      *
FROM        Order O     WITH (NOLOCK)
LEFT JOIN   OrderItems  
WHERE       1=1
```

This pull request enables LOCK with SELECT + JOIN sql query.
This pull request will prevent deadlock situations while querying SELECT statements during transactions. (in case of MSSQL Driver)

Closes: #4764

### Pull-Request Checklist

- [x] `(N/A)` Documentation has been updated to reflect this change 

### DIFF
--- a/src/query-builder/SelectQueryBuilder.ts
+++ b/src/query-builder/SelectQueryBuilder.ts
@@ -1437,21 +1437,6 @@ export class SelectQueryBuilder<Entity> extends QueryBuilder<Entity> implements 
         if (allSelects.length === 0)
             allSelects.push({ selection: "*" });
 
-        let lock: string = "";
-        if (this.connection.driver instanceof SqlServerDriver) {
-            switch (this.expressionMap.lockMode) {
-                case "pessimistic_read":
-                    lock = " WITH (HOLDLOCK, ROWLOCK)";
-                    break;
-                case "pessimistic_write":
-                    lock = " WITH (UPDLOCK, ROWLOCK)";
-                    break;
-                case "dirty_read":
-                    lock = " WITH (NOLOCK)";
-                    break;
-            }
-        }
-
         // Use certain index
         let useIndex: string = "";
         if (this.expressionMap.useIndex) {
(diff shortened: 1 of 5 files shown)


danaryu:
Any plans on when to merge this? I think this would solve a lot of my problems!

pleerock:
Looks like a breaking change? What is the difference between locks being set on `FROM` versus on `FROM` and `JOIN`s?

iceCreamParlor:
In summary,
#### In MSSQL, although we set `lock` with typeorm entity (or querybuilder), lock mode is partially set in `SELECT` query.

In MSSQL, we can set `transaction` lock mode in `SELECT` query. 
```sql
SELECT      *
FROM        Order      WITH (NOLOCK)
WHERE       1=1
```
When we write SQL like above, we can set `NOLOCK` mode to `SELECT` query. We can make such query with typeorm like this.
```typescript
await getRepository(tbOrder).find({
  where: {
    ...
  },
  lock: {
    // this part
    mode: "dirty_read",
  },
});
```

But when we write query with typeorm joining other tables like below,
```typescript
await getRepository(tbOrder).find({
  where: {
    ...
  },
  relations: ["OrderItems"],
  lock: {
    mode: "dirty_read",
  },
});
```
typeorm generates sql query like this
```sql
SELECT      *
FROM        Order O     WITH (NOLOCK)
LEFT JOIN   OrderItems  -- (lock not set in OrderItems!!)
WHERE       ....
```
In this case, `NOLOCK` only applies to `Order` table. `NOLOCK` not applies to `OrderItem` table. Which occurs `TRANSACTION DEADLOCK` in some cases. I faced `TRANSACTION DEADLOCK` situation like above. 



This pull request generates `SELECT` query like this 
```sql
SELECT      *
FROM        Order O     WITH (NOLOCK)
LEFT JOIN   OrderItems  WITH (NOLOCK)
WHERE       ...
```

I changed typeorm code locally, and I found such issue can be resolved by merging this pull request. 



Typeorm is one of my favorite ORM library, and I will be very excited if I can contribute to this library.
Have a good weekend : )

iceCreamParlor:
For your question, this link is relevant to your answer.

https://stackoverflow.com/questions/3783525/sql-server-nolock-and-joins

iceCreamParlor:
Any updates on this pull request ?

AlexMesser (approved):

AlexMesser:
thank you for contribution!

Issue #8522, opened 2022-01-11
Single table inheritance returns the same discriminator value error for unrelated tables where their parents extend from the same entity


marypaz-sama:
## **Issue Description**

There are two *unrelated* child entities following the single table inheritance pattern with same discriminator value, if both entities inherit from a class that just adds some common columns it, it throws an error.

For example:
```

             BaseEntity
           /            \
        Role           User
         |               |
     InternalRole    InternalUser
```

Both `InternalRole` and `InternalRole` entities have the same discriminator value `internal`.
`BaseEntity` just defines some common columns.

## **Expected Behavior**
There should be no error, as the entities `InternalUser` and `InternalRole` are not related because `BaseEntity` is just a generic entity to create createdAt, updatedAt and primary columns for all tables.  And TableInheritance decorator is added in `User` and `Role` entities not `BaseEntity`

## **Actual Behavior**
It throws an error.

> Error: Entities InternalUser and InternalRole have the same discriminator values. Make sure they are different while using the @ChildEntity decorator.

## **Steps to Reproduce**

Minimal code reproduction:

```
export class BaseEntity {
  @PrimaryGeneratedColumn()
  id!: number;

  @CreateDateColumn()
  createdAt: Date;

  @UpdateDateColumn()
  updatedAt: Date;
}
```

```
@Entity()
@TableInheritance({ column: { type: 'varchar', name: 'type' } })
export abstract class User extends BaseEntity {
  @Column()
  firstName: string;

  @Column()
  lastName: string;
}
```

```
@ChildEntity('internal')
export class InternalUser extends User {
 }
```

```
@Entity()
@TableInheritance({ column: { type: 'varchar', name: 'type' } })
export class Role extends BaseEntity {
  @Column()
  name: string;

  @Column()
  description: string;
}
```

```
@ChildEntity('internal')
export class InternalRole extends Role {}
```

## **My Environment**

| Dependency           | Version                                                               |
| ------------------  | -------------------------------------------  |
| Operating System  | macOs                                                               |
| Node.js version      | v16.13.1                                                              |
| Typescript version | v4.1.3                                                                 |
| TypeORM version  | v0.2.41                                                               |
| NestJS Version      | v8.1.8                                                                 |

### **Relevant Database Driver(s)**

- [ ]  aurora-data-api
- [ ]  aurora-data-api-pg
- [ ]  better-sqlite3
- [ ]  cockroachdb
- [ ]  cordova
- [ ]  expo
[...]

PR #8525, opened 2022-01-12
fix: Single table inheritance returns the same discriminator value error for unrelated tables where their parents extend from the same entity 

marypaz-sama:
### Description of change

Fixes: #8522 

When there are two unrelated child entities following the single table inheritance pattern with same discriminator value, if both entities inherit from a class that just adds some common columns it, it throws an error. So the change is to be sure that we are only checking within entities that have the same db table.

#### 

### Pull-Request Checklist

- [x] This pull request links relevant issues as `Fixes [#8522](https://github.com/typeorm/typeorm/issues/8522)`


pleerock:
Thank you for contribution! 🎉

PR #8526, opened 2022-01-13
fix: materialized path being computed as "undefined1."

mschwebler-tgm:
### Description of change
When tree entities are saved implicitly (through another related entity which hasMany tree-entities and "cascade" set to true) the ORM generates materialized-path strings prefixed with `undefined`. Resulting in materialized paths like `undefined1.2.3`. This pull request fixes that. Root cause was that the materialized path always relies on the previously inserted entity mpath which is `undefined` for the root-tree-entity.

### Pull-Request Checklist

- [ x] The new commits follow conventions explained in [CONTRIBUTING.md](https://github.com/typeorm/typeorm/blob/master/CONTRIBUTING.md)

### DIFF
--- a/src/persistence/tree/MaterializedPathSubjectExecutor.ts
+++ b/src/persistence/tree/MaterializedPathSubjectExecutor.ts
@@ -97,13 +97,13 @@ export class MaterializedPathSubjectExecutor {
             .execute();
     }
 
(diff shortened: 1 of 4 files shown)


pleerock:
please remove `.only` to make sure all tests are passing

pleerock:
Thank you for contribution! 🎉

Issue #8527, opened 2022-01-14
Cannot clear database inside a transaction


varanauskas:
## Issue Description

`queryRunner.clearDatabase()` fails if trying to run inside a transaction.

### Expected Behavior

It clears the database inside the transaction, then operations can be performed inside the transaction on a temporarily empty database, and that transaction can later be reverted

### Actual Behavior
It throws `TransactionAlreadyStartedError` when trying to `clearDatabase` inside a transaction

```
// TransactionAlreadyStartedError: Transaction already started for the given connection, commit current transaction before starting a new one.
```

### Steps to Reproduce

1. Connect to a databsae
2. Start Transaction
3. `clearDatabase`
4. It throws `TransactionAlreadyStartedError`

```typescript
const connection = await createConnection();
const queryRunner = connection.createQueryRunner();
await queryRunner.startTransaction();
await queryRunner.clearDatabase();
// ^ This fails with `TransactionAlreadyStartedError`
```

### My Environment

| Dependency          | Version  |
| ---                 | ---      |
| Operating System    | macOS |
| Node.js version     | 16.13.1  | 
| Typescript version  | 4.5.4 | 
| TypeORM version     |0.2.41  | 

### Additional Context

I want to `clearDatabase` inside a transaction, so I could get a temporary blank slate database, and then run migrations inside the same transaction, so I can determine what tables are missing after running all migrations regardless of, and without modifying the initial state of the database

### Relevant Database Driver(s)

| DB Type              | Reproducible  |
| ---                  | ---           |
| `aurora-data-api`    | no            |
| `aurora-data-api-pg` | no            |
| `better-sqlite3`     | no            |
| `cockroachdb`        | no            |
| `cordova`            | no            |
| `expo`               | no            |
| `mongodb`            | no            |
| `mysql`              | no            |
| `nativescript`       | no            |
| `oracle`             | no            |
| `postgres`           | yes            |
| `react-native`       | no            |
| `sap`                | no            |
| `sqlite`             | no            |
| `sqlite-abstract`    | no            |
| `sqljs`              | no            |
| `sqlserver`          | no            |

### Are you willing to resolve this issue by submitting a Pull Request?

 - ✅ Yes, I have the time, and I know how to start.
 - ✖️ Yes, I have the time, but I don't know how to start. I would need guidance.
 - ✖️ No, I don’t have the time, but I can support (using donations) development.
[...]

PR #8529, opened 2022-01-14
fix: allow clearing database inside a transaction

varanauskas:
Possible fix for: #8527

### Description of change

This is one of the possible fixes for the issue #8527

It allows to run `clearDatabase` inside a transaction by making `clearDatabase` to check if another transaction is already running and not creating a new transaction in that case

It currently is only developed for/tested with Postgres, if this is the right direction same fix could be applied for other query runners

### Pull-Request Checklist


pleerock:
I think we can accept it.

> It currently is only developed for/tested with Postgres, if this is the right direction same fix could be applied for other query runners

if we go with this fix, fix needs to be applied to all other query runners as well.

pleerock:
I think changes still do not cover all drivers.

varanauskas:
Yes, I could not successfully test the changes on other drivers so I did not submit them, I will try to diagnose why I couldn't run some of the docker container tests and submit changes for other drivers. (Any tips on testing with more exotic drivers like SAP Hana or Oracle XD?)

pleerock:
For SAP Hana there is no easy way... Hana's official site provides a trial for their product as I know. For Oracle and MSSQL you can simply commit and check if CI is failing. Since the code you are changing is straightforward, I think you can just copy/paste code in required places without mistakes and everything else should be fine.

pleerock:
I'm going to close this one because I'm doing unfinished PRs cleanup (to prevent conflicts). Please feel free to rebase it on latest master and reopen it with all requested changes applied. Thank you.

varanauskas:
Thank you, I will add the changes to other drivers and reopen

Issue #8556, opened 2022-01-22
TreeRepository findDescendants returns everything if parent does not exist or deleted


astritsh:
## Issue Description
findDescendants of TreeRepository using materialised-path design returns everything if parent was deleted before getting children or if we query with a non existing parent id.

In my case when deleting children of a parent that was already deleted I ended up deleting everything from the tree entity table(quite dangerous!).

### Expected Behavior
findDescendants should return empty for a non existing parent.

### Actual Behavior
findDescendants is returning all tree entity records from database if we query with a non existing parent id.

### Steps to Reproduce

1. Save a tree entity parent on db.(e.g saved id is 1)
2. findDescendants({id: -1})(using a non existing id-1)
3. parent with id: 1 saved in step 1 is being returned

```typescript
// insert code here
```

### My Environment

NestJs.

| Dependency          | Version  |
| ---                 | ---      |
| Operating System    |          |
| Node.js version     | 12.14.1  | 
| Typescript version  | x.y.zzz  | 
| TypeORM version     | 0.2.41  | 

### Relevant Database Driver(s)

| DB Type              | Reproducible  |
| ---                  | ---           |
| `aurora-data-api`    | no            |
| `aurora-data-api-pg` | no            |
| `better-sqlite3`     | no            |
| `cockroachdb`        | no            |
| `cordova`            | no            |
| `expo`               | no            |
| `mongodb`            | no            |
| `mysql`              | yes            |
| `nativescript`       | no            |
| `oracle`             | no            |
| `postgres`           | yes            |
| `react-native`       | no            |
| `sap`                | no            |
| `sqlite`             | no            |
| `sqlite-abstract`    | no            |
| `sqljs`              | no            |
| `sqlserver`          | yes            |

### Are you willing to resolve this issue by submitting a Pull Request?

 - ✅ Yes, I have the time, and I know how to start.
 - ✖️ Yes, I have the time, but I don't know how to start. I would need guidance.
 - ✖️ No, I don’t have the time, but I can support (using donations) development.
 - ✖️ No, I don’t have the time and I’m okay to wait for the community / maintainers to resolve this issue.

Yuvraj102:
Hello Authors, Can you please point a resource for detailed implementation of materialised path in typeorm ???, the document is very short on this topic and doesn't cover the implementation in detail.

ashokars:
Tree entity document should be detailed

PR #8557, opened 2022-01-22
fix: find descendants of a non-existing tree parent

astritsh:
Closes: #8556

### Description of change

If subquery in findDescendants `CONCAT(${subQuery.getQuery()}, '%')` returns null when entity parent does not exist, the search string becomes `LIKE '%'`, because CONCAT ignores NULL. This `LIKE '%'` results on returning everything instead of nothing.
Now with `NULLIF` fix, search will be `LIKE NULL`, which returns nothing for a non existing tree parent.
In case of Sqlite, `LIKE ${subQuery.getQuery()} || '%'` is already evaluated as `LIKE NULL` when subquery returns null.

### Pull-Request Checklist

- [x ] Code is up-to-date with the `master` branch
- [ x] `npm run lint` passes with this change
- [ x] `npm run test` passes with this change
- [ x] This pull request links relevant issues as `Fixes #0000`
- [x ] There are new or updated unit tests validating the change
- [ x] The new commits follow conventions explained in [CONTRIBUTING.md](https://github.com/typeorm/typeorm/blob/master/CONTRIBUTING.md)

### DIFF
--- a/src/repository/TreeRepository.ts
+++ b/src/repository/TreeRepository.ts
@@ -137,7 +137,7 @@ export class TreeRepository<Entity> extends Repository<Entity> {
                     if (this.manager.connection.driver instanceof AbstractSqliteDriver) {
                         return `${alias}.${this.metadata.materializedPathColumn!.propertyPath} LIKE ${subQuery.getQuery()} || '%'`;
                     } else {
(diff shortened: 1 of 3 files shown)


pleerock:
Thank you for contribution! 🎉

PR #8472, opened 2021-12-19
fix: include asExpression columns in returning clause

Migushthe2nd:
### Description

Fixes #8450 

It is a bit unclear wether the `isGenerated` property on `TableColumn` and `ColumnMetadata` (and possibly in more places) should be set _only_ if `generationStrategy` is defined as well. The property description in the `ColumnMetadata` class mentions it is meant for columns of "generated other way" (I read this as "any type of generation", including `asExpression`), but most of the code that utilizes the `isGenerated` field often does not adhere to this.
https://github.com/typeorm/typeorm/blob/cefddd95c550191d6a18cb53c8ea4995d0c219ca/src/metadata/ColumnMetadata.ts#L83-L86
Perhaps this comment refers only to _other generation strategies_, like `uuid`?

I could think of two ways to solve the issue:
1. If the `isGenerated` property is not restricted to generation strategies: go through all references of `isGenerated` and confirm or rewrite parts to make `isGenerated` a universal indicator for columns that are generated in some way.
2. Make it clear that `isGenerated` should only be used for generation strategies. The fix to the actual issue would then be simple: change line 181 of the condition below to
```ts
(needToCheckGenerated && (column.isGenerated || !!column.asExpression))  ||
```
https://github.com/typeorm/typeorm/blob/90a8deb6383c31686965893cdea96151f40be396/src/query-builder/ReturningResultsEntityUpdator.ts#L172-L187
This makes it such that columns with an `asExpression` are also included in the `RETURNING` clause when saving an entity.

### Pull-Request Checklist

- [X] Code is up-to-date with the `master` branch
- [x] `npm run lint` passes with this change
- [ ] `npm run test` passes with this change
- [X] This pull request links relevant issues as `Fixes #0000`
- [X] There are new or updated unit tests validating the change
- [X] Documentation has been updated to reflect this change 
  (I updated documentation in the source files, not docs directory)
- [x] The new commits follow conventions explained in [CONTRIBUTING.md](https://github.com/typeorm/typeorm/blob/master/CONTRIBUTING.md)

PR #8580, opened 2022-01-28
refactor: remove the need for a metadata table when using generated c…

TheNoim:
…olumns in postgres

### Description of change

The original implementation for generated columns required us to save the user provided "AS" expression into a metadata table. The reason behind it is that we are not able to get the original user provided expression from Postgres itself. This pr uses a different approach. Instead of retrieving the original expression, we just format and convert the user provided expression. With this, we are able to compare both expressions without the need for a metadata table. 

I saw a few people who had the issue that the metadata table doesn't exist in some situation. This pr would result in a fix for theses people. For example, the tests of this pr #8472 are failing because of this. 

### Pull-Request Checklist


TheNoim:
Hey, can someone review this?

AlexMesser:
It looks like the generated columns will always be marked as "changed" during synchronization ([this check](https://github.com/typeorm/typeorm/blob/c5dfc11ea7b8b33aa9e621b64b953c1d62a27989/src/driver/postgres/PostgresDriver.ts#L1241)), because user specifies `asExpression` like `text || tag` but you expect in the test that it will be `((text)::text || (tag)::text)`.

AlexMesser:
I'm going to close this because of sync issue described above

PR #8592, opened 2022-01-31
Update stream usage of MSSQL library including bump to 7.3.0 support

dhensby:
### Description of change

1. Add support for MSSQL version 7.3.0 (v6 is now EOL) and relies on very old tedious version
2. Make use of the native `Readable` stream that driver exposes via `toReadableStream` 

### Pull-Request Checklist


src/driver/sqlserver/SqlServerQueryRunner.ts
```diff
@@ -282,28 +282,27 @@ export class SqlServerQueryRunner extends BaseQueryRunner implements QueryRunner
...
         request.query(query);
 
-        // Any event should release the lock.
+        const streamRequest = request.toReadableStream();
+
+        // Any event on the underlying request should release the lock.
         request.once("row", release);
         request.once("rowsaffected", release);
         request.once("done", release);
         request.once("error", release);
```
dhensby, line 291: I'm a bit unsure about these lines. If the lock is on the connection via a pool, then really we shouldn't be releasing it until the request has properly ended

pleerock: Probably we need to ask author of these changes (hope it's not me).

src/driver/sqlserver/SqlServerQueryRunner.ts
```diff
@@ -282,28 +282,27 @@ export class SqlServerQueryRunner extends BaseQueryRunner implements QueryRunner
 
         request.query(query);
 
-        // Any event should release the lock.
+        const streamRequest = request.toReadableStream();
```
pleerock, line 285: Is this change backward compatible with mssql v6?

src/driver/sqlserver/SqlServerQueryRunner.ts
dhensby: After thinking about it, these should be removed and likely were there because there wasn't a proper stream being exposed in by of mssql, so this was just a bit of a "hack" to get it working

src/driver/sqlserver/SqlServerQueryRunner.ts
dhensby: No, this was introduced in v7. V6 of mssql is quite old and not supported any longer and there have been improvements, including to the streaming implementations.

dhensby: I've updated the PR to target `0.3.0`

src/driver/sqlserver/SqlServerQueryRunner.ts
dhensby: I've made the change but also looked to track down the original author. This was added in https://github.com/typeorm/typeorm/pull/7893 by @imnotjames.

I'm a bit cautious about this particular part of the changes as I don't understand why a lock is even necessary (I've looked at the MySQL implementation and there is no equivalent lock happening on queries). So some insight as to why the lock is needed and what it's achieving should shed light on when the lock should be released.

pleerock: I don't know either. Let's merge this PR for now and keep track on @imnotjames response to your question.

pleerock:
oh yeah and we need to update changelog on that branch

pleerock:
@dhensby it looks like the latest mssql version is 8.x, do you have a wish to integrate it?

Tam2:
This still has a peerDependency of "mssql": "^6.3.1" - will this be updated to 7.x?

josendf-basetis:
The mssql peer dependency was upgraded to "mssql": "^7.3.0". It would help that the mssql peer dependency be specified as  ^7 || ^8 as mentioned [here](https://github.com/typeorm/typeorm/pull/8592#issuecomment-1042774308).

We are using mssql@8.1.0 + typeorm@0.2.41 with no issues, we are unable to upgrade typeorm due to the mssql version constraint.

dhensby:
@josendf-basetis it's an open source project - open a PR

PR #8597, opened 2022-02-01
fix: Lock peer dependencies versions

Bartmr:
### Description of change

This PR locks down `peerDependencies` to their minor versions.

#### Why
Right now, TypeORM does not lock peer dependencies. So, if someone upgrades a major version of their database driver in their project, and that same upgrade contains breaking changes (like updating the Mongo driver to v4), noticeable and silent errors might happen.

By locking the peerDependency versions, both the build-time type-checks and the tests run in the TypeORM repo will almost guarantee that those driver versions will work, and at least warn the users if they are upgrading to a  version of a database driver that still isn't tested or typechecked inside the TypeORM code-base.

#### What

All peer dependencies were set to the same versions that are currently set in `devDependencies`, since they have been type-checked during previous builds and well tested. Those that weren't previously set in `devDependencies` (like `@sap/hana-client`) were set to their latest version.

### Pull-Request Checklist

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -414,15 +414,6 @@
         "@types/ms": "*"
       }
     },
-    "@types/dotenv": {
-      "version": "8.2.0",
-      "resolved": "https://registry.npmjs.org/@types/dotenv/-/dotenv-8.2.0.tgz",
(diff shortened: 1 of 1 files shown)


AlexMesser:
thank you for contribution!

PR #8601, opened 2022-02-01
fix: Exit process on successful migration revert via cli

yura2100:
### Description of change

Add process exit after successfully reverting a migration via cli

**Motivation**

Currently, process is not exited after successfully reverting a migration via cli, so it is unclear if everything is OK 

**Current behavior**

Process is not exited after successfully reverting a migration via cli

**New behavior**

Process is exited with code `0` after successfully reverting a migration via cli

### Pull-Request Checklist

### DIFF
--- a/src/commands/MigrationRevertCommand.ts
+++ b/src/commands/MigrationRevertCommand.ts
@@ -74,6 +74,8 @@ export class MigrationRevertCommand implements yargs.CommandModule {
 
             await connection.undoLastMigration(options);
             await connection.close();
+            // exit process if no errors
+            process.exit(0);
 
         } catch (err) {
             if (connection) await (connection as Connection).close();


pleerock:
After migration successfully reverted, it exists by default. And AFAIK default exist code is already `0`. Why do we need to add this line of code?

pleerock:
I don't really understand why the process is not exited... Connection is being closed and it means all resources consumed by node.js process are free and node process must exist by it's own...

yura2100:
I guess it could be somehow related to yargs. I'll try to investigate more on weekends

pleerock:
Thank you, it will be great! It's always better to find root case to prevent other issues in the future.

yura2100:
Ok, I finally got some time to check it. I guess the problem was in using typeorm with nest + debug mode. And in bare typeorm setup there is no such problem. I am closing this PR as there is nothing to do with typeorm

Issue #8455, opened 2021-12-14
BaseEntity.reload() throws on soft deleted objects


matheusbn:
## Issue Description

Calling `entity.reload()` throws if entity has been soft deleted.

### Expected Behavior

`entity.reload()` should reload even if soft deleted.

### Actual Behavior
```javascript
await invitation.softRemove();
await invitation.reload();
```
```
EntityNotFoundError: Could not find any entity of type "Invitation" matching: 1
    at EntityNotFoundError.TypeORMError [as constructor] (/.../backend/src/error/TypeORMError.ts:7:9)
    at new EntityNotFoundError (/.../backend/src/error/EntityNotFoundError.ts:10:9)
    at /.../backend/src/entity-manager/EntityManager.ts:889:39
    at processTicksAndRejections (internal/process/task_queues.js:93:5)

  at Object.removeInvitation (src/services/invitationService.ts:87:15)
```

### My Environment

| Dependency          | Version  |
| ---                 | ---      |
| Operating System    |   Windows 10   |
| Node.js version     | v14.16.0 | 
| Typescript version  | 4.5.2 | 
| TypeORM version     | 0.2.41 | 

### Relevant Database Driver(s)

| DB Type              | Reproducible  |
| ---                  | ---           |
| `aurora-data-api`    | no            |
| `aurora-data-api-pg` | no            |
| `better-sqlite3`     | no            |
| `cockroachdb`        | no            |
| `cordova`            | no            |
| `expo`               | no            |
| `mongodb`            | no            |
| `mysql`              | yes           |
| `nativescript`       | no            |
| `oracle`             | no            |
| `postgres`           | no            |
| `react-native`       | no            |
| `sap`                | no            |
| `sqlite`             | no            |
| `sqlite-abstract`    | no            |
| `sqljs`              | no            |
| `sqlserver`          | no            |

### Are you willing to resolve this issue by submitting a Pull Request?

 - ✅ Yes, I have the time, and I know how to start.
 - ✖️ Yes, I have the time, but I don't know how to start. I would need guidance.
 - ✖️ No, I don’t have the time, but I can support (using donations) development.
 - ✖️ No, I don’t have the time and I’m okay to wait for the community / maintainers to resolve this issue.

pleerock:
Why reload should load soft-deleted record? I don't see a clear motivation. Other can argue that reload shouldn't load soft-deleted record, since its... soft-deleted.

osvaguillen:
@pleerock 
[...]

PR #8616, opened 2022-02-09
0.3.0

pleerock:
## [0.3.0]() (2022-03-10)

Changes in the version includes changes from the `next` branch and `typeorm@next` version.
They were pending their migration from 2018. Finally, we are ready to merge them into master and release new `0.3.0`.

### FEATURES

* compilation `target` now is `es2020`. This requires Node.JS version `12.9.0`+

* `Connection` was renamed to `DataSource`.
Old `Connection` is still there, but now it's deprecated. It will be removed in next version.
New API:

```ts
export const dataSource = new DataSource({
    // ... options ...
})

// load entities, establish db connection, sync schema, etc.
await dataSource.connect()
```

Previously, you could use `new Connection()`, `createConnection()`, `getConnectionManager().create()`, etc.
They all deprecated in favour of new syntax you can see above.

New way gives you more flexibility and simplicity in usage.

* new custom repositories syntax:

```ts
export const UserRepository = myDataSource.getRepository(UserEntity).extend({
    findUsersWithPhotos() {
        return this.find({
            relations: {
                photos: true
            }
        })
    }
})
```

Old ways of custom repository creation were deprecated.

* added new option on relation load strategy called `relationLoadStrategy`.
Relation load strategy is used on entity load and their relations load when you query entities from the database.
Used on `find*` methods and `QueryBuilder`. Value can be set to `join` or `query`.

    * `join` loads relations using SQL's `JOIN`
    * `query` executes separate SQL queries for each relation

Default is `join`, but default can be set in `ConnectionOptions`:

```ts
createConnection({
    /* ... */
    relationLoadStrategy: "query"
})
```

Also, it can be set per-query in `find*` methods:

```ts
userRepository.find({
    relations: {
        photos: true
    }
})
```

And QueryBuilder:

```ts
userRepository
    .createQueryBuilder()
    .setRelationLoadStrategy("query")
```

For queries returning big amount of data, we recommend to use `query` strategy,
because it can be a more performant approach to query relations.

* new `select` type signature in `FindOptions` (used in `find*` methods):

```ts
userRepository.find({
    select: {
        id: true,
        firstName: true,
        lastName: true,
    }
})
```

Also, now it's possible to specify select columns of the loaded relations:

```ts
userRepository.find({
    select: {
        id: true,
        firstName: true,
        lastName: true,
        photo: {
            id: true,
            filename: true,
            album: {
                id: true,
                name: true,
            }
        }
    }
})
```

* new `relations` type signature in `FindOptions` (used in `find*` methods):

```ts
userRepository.find({
    relations: {
        contacts: true,
        photos: true,
    }
})
```

To load nested relations use a following signature:

```ts
userRepository.find({
    relations: {
        contacts: true,
        photos: {
            album: true,
        },
    }
})
```

* new `order` type signature in `FindOptions` (used in `find*` methods):

```ts
userRepository.find({
    order: {
        id: "ASC"
    }
})
```

Now supports nested order by-s:

```ts
userRepository.find({
    order: {
        photos: {
            album: {
                name: "ASC"
            },
        },
    }
})
```

* new `where` type signature in `FindOptions` (used in `find*` methods) now allows to build nested statements with conditional relations, for example:

```ts
userRepository.find({
    where: {
        photos: {
            album: {
                name: "profile"
            }
        }
    }
})
```

Gives you users who have photos in their "profile" album.

* `FindOperator`-s can be applied for relations in `where` statement, for example:

```ts
userRepository.find({
    where: {
        photos: MoreThan(10),
    }
})
```

Gives you users with more than 10 photos.

* `boolean` can be applied for relations in `where` statement, for example:

```ts
userRepository.find({
    where: {
        photos: true
    }
})
```

### DEPRECATIONS

* `select` in `FindOptions` (used in `find*` methods) used as an array of property names is deprecated.
Now you should use a new object-literal notation. Example:

Deprecated way of loading entity relations:

```ts
userRepository.find({
    select: ["id", "firstName", "lastName"]
})
```

New way of loading entity relations:

```ts
userRepository.find({
    select: {
        id: true,
        firstName: true,
        lastName: true,
    }
})
```

This change is due to type-safety improvement new `select` signature brings.

* `relations` in `FindOptions` (used in `find*` methods) used as an array of relation names is deprecated.
Now you should use a new object-literal notation. Example:

Deprecated way of loading entity relations:

```ts
userRepository.find({
    relations: ["contacts", "photos", "photos.album"]
})
```

New way of loading entity relations:

```ts
userRepository.find({
    relations: {
        contacts: true,
        photos: {
            album: true
        }
    }
})
```

This change is due to type-safety improvement new `relations` signature brings.

* `join` in `FindOptions` (used in `find*` methods) is deprecated. Use `QueryBuilder` to build queries containing manual joins.

* `Connection`, `ConnectionOptions` are deprecated, new names to use are: `DataSource` and `DataSourceOptions`.
To create the same connection you had before use a new syntax: `new DataSource({ /*...*/ })`.

* `createConnection()`, `createConnections()` are deprecated, since `Connection` is called `DataSource` now, to create a connection and connect to the database
simply do:

```ts
const myDataSource = new DataSource({ /*...*/ })
await myDataSource.connect()
```

* `getConnection()` is deprecated. To have a globally accessible connection, simply export your data source and use it in places you need it:

```ts
export const myDataSource = new DataSource({ /*...*/ })
// now you can use myDataSource anywhere in your application
```

* `getManager()`, `getMongoManager()`, `getSqljsManager()`, `getRepository()`, `getTreeRepository()`, `getMongoRepository()`, `createQueryBuilder()`
are all deprecated now. Use globally accessible data source instead:

```ts
export const myDataSource = new DataSource({ /*...*/ })
export const Manager = myDataSource.manager
export const UserRepository = myDataSource.getRepository(UserEntity)
export const PhotoRepository = myDataSource.getRepository(PhotoEntity)
// ...
```

* `getConnectionManager()` and `ConnectionManager` itself are deprecated - now `Connection` is called `DataSource`,
and each data source can be defined in exported variable. If you want to have a collection
of data sources, just define them in a variable, simply as:

```ts
const dataSource1 = new DataSource({ /*...*/ })
const dataSource2 = new DataSource({ /*...*/ })
const dataSource3 = new DataSource({ /*...*/ })

export const MyDataSources = {
    dataSource1,
    dataSource2,
    dataSource3,
}
```

* `getConnectionOptions()` is deprecated - in next version we are going to implement different mechanism of connection options loading

* `AbstractRepository` is deprecated. Use new way of custom repositories creation.

* `@TransactionRepository`, `@TransactionManager`, `@Transaction` decorators were deprecated.

* all deprecated signatures will be removed in `0.4.0`

### BREAKING CHANGES

* minimal Node.JS version requirement now is `12.9.0`

* now migrations are running before schema synchronization if you have both pending migrations and schema synchronization pending
  (it works if you have both `migrationsRun` and `synchronize` enabled in connection options).

* `findOne` and `QueryBuilder.getOne()` now return `null` instead of `undefined` in the case if it didn't find anything in the database.
Logically it makes more sense to return `null`.

* `where` in `FindOptions` (e.g. `find({ where: { ... })`) is more sensitive to input criteria now.

* if you had entity properties of a non-primitive type (except Buffer) defined as columns,
then you won't be able to use it in `find*`'s `where`. Example:

Before for the `@Column(/*...*/) membership: MembershipKind` you could have a query like:

```ts
userRepository.find({
    membership: new MembershipKind("premium")
})
```

now, you need to wrap this value into `Equal` operator:

```ts
userRepository.find({
    membership: Equal(new MembershipKind("premium"))
})
```

This change is due to type-safety improvement new `where` signature brings.

* `order` in `FindOptions` (used in `find*` methods) doesn't support ordering by relations anymore.
Define relation columns, and order by them instead.

* `where` in `FindOptions` (used in `find*` methods) previously supported `ObjectLiteral` and `string` types.
Now both signatures were removed. ObjectLiteral was removed because it seriously breaks the type safety,
and `string` doesn't make sense in the context of `FindOptions`. Use `QueryBuilder` instead.

* `MongoRepository` and `MongoEntityManager` now use new types called `MongoFindManyOptions` and `MongoFindOneOptions`
for their `find*` methods.

* `primary relation` (e.g. `@ManyToOne(() => User, { primary: true }) user: User`) support is removed.
You still have an ability to use foreign keys as your primary keys,
however now you must explicitly define a column marked as primary.

Example, before:

```ts
@ManyToOne(() => User, { primary: true })
user: User
```

Now:

```ts
@PrimaryColumn()
userId: number

@ManyToOne(() => User)
user: User
```

Primary column name must match the relation name + join column name on related entity.
If related entity has multiple primary keys, and you want to point to multiple primary keys,
you can define multiple primary columns the same way:

```ts
@PrimaryColumn()
userFirstName: string

@PrimaryColumn()
userLastName: string

@ManyToOne(() => User)
user: User
```

This change was required to simplify ORM internals and introduce new features.

### EXPERIMENTAL FEATURES NOT PORTED FROM NEXT BRANCH

* `observers` - we will consider returning them back with new API in future versions
* `alternative find operators` - using `$any`, `$in`, `$like` and other operators in `where` condition.

----------

## THINGS I ASK TO HELP ME WITH

* test it (you can install it using `npm i typeorm@rc`)
* review this PR, add any other relevant information to the CHANGELOG
* update documentation to reflect new changes on our website
* add tests to new functionality
* update other repositories with new changes introduced in 0.3.0


Ginden:
>  test it (you can install it using npm i typeorm@rc)

Breaking changes not documented:

* `Repository` has constructor (breaking classes extending `Repository` with custom constructor)

src/repository/Repository.ts
```diff
@@ -436,4 +454,22 @@ export class Repository<Entity extends ObjectLiteral> {
...
+        //     ...custom
+        // };
+        const thisRepo: any = this.constructor;
+        const { target, manager, queryRunner } = this;
+        const cls = new class extends thisRepo {
+            constructor() {
+                super(target, manager, queryRunner);
+            }
+        };
+        Object.assign(cls, custom);
```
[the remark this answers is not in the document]
pleerock: > Shouldn't it be cls.prototype

don't think so, we are returning instance here. Check it in the runtime. Or [here](https://www.typescriptlang.org/play?#code/MYGwhgzhAEBKCmAHA9hAlgF2QJwJ7QG8BYAKGnOkWzQDcwN5phkA7F+YDNV6AXmgCMpUhWgBzeBgDCrdp24sAFAEpCI0RWySArthbQMACzQQAdMzYcurdeQC+wshXgAPBiwAmAHinaIWAFsEFHQsPAA+RWA-QIAuaF9-ZCCkVEwcfAAyaAAVYwgc3ER4L2C0sKyEmOSy0Izw5Xja9LxobMTA5oq1Jw0mVn8DfLL4sBZ8fiMTcwGMbG1OHFsNC0HQGH52AHcmcChoV3cPGCmIMsJoB16NAHkAIwArK1NIdDEldYAaJmqA5WXRFoMLp9OtoJBweMANzLK5XUirDDQACqEHg2C6GT40G2cFSdTwKlMh3gnkUxGufnREBUPT65CBIKG0wk0lkVgUtIA1NAAEQAQkFvNhpDs-xICIGyBA8FMIGQYkUqPRmLwpip2BpymUQA).

glen-84:
@pleerock When might you consider switching to a 1.x release?

src/connection/BaseConnectionOptions.ts
```diff
@@ -121,6 +121,24 @@ export interface BaseConnectionOptions {
...
+     */
+    readonly relationLoadStrategy?: "join" | "query"
+
+    /**
+     * Optionally applied "typename" to the model.
+     * If set, then each hydrated model will have this property with the target model / entity name inside.
+     *
+     * (works like a discriminator property).
+     */
+    readonly typename?: string;
```
Ginden, line 140: It would be sensible to make it `string | symbol`.

src/connection/Connection.ts
```diff
@@ -165,6 +171,20 @@ export class Connection {
     // -------------------------------------------------------------------------
     // Public Methods
     // -------------------------------------------------------------------------
+    /**
+     * Updates current connection options with provided options.
+     */
+    setOptions(options: Partial<ConnectionOptions>) {
```
Ginden, line 177: Should have return type.

src/connection/BaseConnectionOptions.ts
pleerock: why?

pleerock:
> @pleerock When might you consider switching to a 1.x release?

@glen-84 not yet.

> Can you provide rationale on this?

@Ginden new way is much simpler and flexible to use

src/connection/BaseConnectionOptions.ts
Ginden: @pleerock It's useful to have discriminator property not returned through JSON.stringify and not conflicting with any possible column name.

pleerock:
Looks like current shortening implementation of `DriverUtils.buildAlias` doesn't work properly. For the following SQL:

```sql
SELECT "Post"."id" AS "Post_id", "Post"."title" AS "Post_title", "Post"."userId" AS "Post_userId", "Post"."authorId" AS "Post_authorId", "Post"."countersStars" AS "Post_countersStars", "Post"."countersCommentcount" AS "Post_countersCommentcount", "Post__Post_photos"."id" AS "Post__Post_photos_id", "Post__Post_photos"."filename" AS "Post__Post_photos_filename", "Post__Post_photos"."userId" AS "Post__Post_photos_userId", "Post__Post_photos"."postId" AS "Post__Post_photos_postId", "Post__Post_photos"."authorId" AS "Post__Post_photos_authorId", "Post__Post_photos"."countersStars" AS "c67ec5ec9f1d0cd4c761c3bd6666f", "Post__Post_photos"."countersCommentcount" AS "596a2d15b2cca5768282a1c3861dc", "Post__Post__Post__Post"."id" AS "Post__Post__Post__Post_id", "Post__Post__Post__Post"."name" AS "Post__Post__Post__Post_name", "Post__Post__Post__Post"."id" AS "Post__Post__Post__Post_id", "Post__Post__Post__Post"."name" AS "Post__Post__Post__Post_name", "Post__Post_user"."id" AS "Post__Post_user_id", "Post__Post_user"."name" AS "Post__Post_user_name", "Post__Post_counters_author"."id" AS "Post__Post_counters_author_id", "Post__Post_counters_author"."name" AS "9dc61bb2b6fc0b189ebaa1d3a0621", Post__Post__Post__Post FROM "post" "Post" LEFT JOIN "photo" "Post__Post_photos" ON "Post__Post_photos"."postId"="Post"."id"  LEFT JOIN "user" "Post__Post__Post__Post" ON "Post__Post__Post__Post"."id"="Post__Post_photos"."userId"  LEFT JOIN "user" "Post__Post__Post__Post" ON "Post__Post__Post__Post"."id"="Post__Post_photos"."authorId"  LEFT JOIN "user" "Post__Post_user" ON "Post__Post_user"."id"="Post"."userId"  LEFT JOIN "user" "Post__Post_counters_author" ON "Post__Post_counters_author"."id"="Post"."authorId" WHERE "Post"."id" IN (?)

SELECT "Post"."id" AS "Post_id", "Post"."title" AS "Post_title", "Post"."userId" AS "Post_userId", "Post"."authorId" AS "Post_authorId", "Post"."countersStars" AS "Post_countersStars", "Post"."countersCommentcount" AS "Post_countersCommentcount", "Post__Post_photos"."id" AS "Post__Post_photos_id", "Post__Post_photos"."filename" AS "Post__Post_photos_filename", "Post__Post_photos"."userId" AS "Post__Post_photos_userId", "Post__Post_photos"."postId" AS "Post__Post_photos_postId", "Post__Post_photos"."authorId" AS "Post__Post_photos_authorId", "Post__Post_photos"."countersStars" AS "Post__Post_photos_countersStars", "Post__Post_photos"."countersCommentcount" AS "Post__Post_photos_countersCommentcount", "Post__Post_photos__Post__Post_photos_user"."id" AS "Post__Post_photos__Post__Post_photos_user_id", "Post__Post_photos__Post__Post_photos_user"."name" AS "Post__Post_photos__Post__Post_photos_user_name", "Post__Post_photos__Post__Post_photos_counters_author"."id" AS "Post__Post_photos__Post__Post_photos_counters_author_id", "Post__Post_photos__Post__Post_photos_counters_author"."name" AS "Post__Post_photos__Post__Post_photos_counters_author_name", "Post__Post_user"."id" AS "Post__Post_user_id", "Post__Post_user"."name" AS "Post__Post_user_name", "Post__Post_counters_author"."id" AS "Post__Post_counters_author_id", "Post__Post_counters_author"."name" AS "Post__Post_counters_author_name" FROM "post" "Post" LEFT JOIN "photo" "Post__Post_photos" ON "Post__Post_photos"."postId"="Post"."id"  LEFT JOIN "user" "Post__Post_photos__Post__Post_photos_user" ON "Post__Post_photos__Post__Post_photos_user"."id"="Post__Post_photos"."userId"  LEFT JOIN "user" "Post__Post_photos__Post__Post_photos_counters_author" ON "Post__Post_photos__Post__Post_photos_counters_author"."id"="Post__Post_photos"."authorId"  LEFT JOIN "user" "Post__Post_user" ON "Post__Post_user"."id"="Post"."userId"  LEFT JOIN "user" "Post__Post_counters_author" ON "Post__Post_counters_author"."id"="Post"."authorId" WHERE "Post"."id" IN (?)
```

it gave me following incorrect result:

```sql
SELECT "Post"."id" AS "Post_id", "Post"."title" AS "Post_title", "Post"."userId" AS "Post_userId", "Post"."authorId" AS "Post_authorId", "Post"."countersStars" AS "Post_countersStars", "Post"."countersCommentcount" AS "Post_countersCommentcount", "Post__Post_photos"."id" AS "Post__Post_photos_id", "Post__Post_photos"."filename" AS "Post__Post_photos_filename", "Post__Post_photos"."userId" AS "Post__Post_photos_userId", "Post__Post_photos"."postId" AS "Post__Post_photos_postId", "Post__Post_photos"."authorId" AS "Post__Post_photos_authorId", "Post__Post_photos"."countersStars" AS "c67ec5ec9f1d0cd4c761c3bd6666f", "Post__Post_photos"."countersCommentcount" AS "596a2d15b2cca5768282a1c3861dc", "Post__Post__Post__Post"."id" AS "Post__Post__Post__Post_id", "Post__Post__Post__Post"."name" AS "Post__Post__Post__Post_name", "Post__Post__Post__Post"."id" AS "Post__Post__Post__Post_id", "Post__Post__Post__Post"."name" AS "Post__Post__Post__Post_name", "Post__Post_user"."id" AS "Post__Post_user_id", "Post__Post_user"."name" AS "Post__Post_user_name", "Post__Post_counters_author"."id" AS "Post__Post_counters_author_id", "Post__Post_counters_author"."name" AS "9dc61bb2b6fc0b189ebaa1d3a0621", Post__Post__Post__Post FROM "post" "Post" LEFT JOIN "photo" "Post__Post_photos" ON "Post__Post_photos"."postId"="Post"."id"  LEFT JOIN "user" "Post__Post__Post__Post" ON "Post__Post__Post__Post"."id"="Post__Post_photos"."userId"  LEFT JOIN "user" "Post__Post__Post__Post" ON "Post__Post__Post__Post"."id"="Post__Post_photos"."authorId"  LEFT JOIN "user" "Post__Post_user" ON "Post__Post_user"."id"="Post"."userId"  LEFT JOIN "user" "Post__Post_counters_author" ON "Post__Post_counters_author"."id"="Post"."authorId" WHERE "Post"."id" IN (?)
```

(Post__Post__Post__Post_id) is duplicated. I'll simply remove `shorten` option from all usages.

Penners:
The new relation load strategy option looks great!

carlocorradini:
@pleerock 0.3.0 looks super cool!

package.json
```diff
@@ -125,6 +125,7 @@
     "sqlite3": "^5.0.2",
     "ts-node": "^9.1.1",
     "typeorm-aurora-data-api-driver": "^2.0.0",
+    "date-fns": "2.22.1",
```
carlocorradini, line 128: Should be `"date-fns": "^2.22.1"` and not `"date-fns": "2.22.1"`

spotykatch:
How do we find one by id if we don't know the name of the id column?

```ts
@Entity()
export class Post {
  @PrimariGeneratedColumn()
  someOriginalNameOfTheIdColumn: number;
}
```

pleerock:
@spotykatch you should know column name because you have a strictly defined entity, right? New changes all about better type-safety, so you should know your column name since you defined it in your type/class. If you have a dynamic code, then you are able to get entity id name from entity metadata.

PatrickWe:
Can you provide a migration documentation for 0.3.0 for the connection.transaction(async (maager) => {}), here one example of the old code?

@Injectable()
export class CustomService {
  constructor(
    private readonly connection: Connection,
  ) {}

  private async sendMessage(
  ): Promise<MessageEntity> {
    const message = await this.connection.transaction(async (manager) => {
      const messageRepository = manager.getRepository(MessageEntity);
    });
}
}

pleerock:
@PatrickWe there are no changes in `.transaction` method of `EntityManager`.

spotykatch:
@pleerock there is an issue #8455. Reload method of the `BaseEntity` throws an error if the entity is soft deleted.
https://github.com/typeorm/typeorm/blob/20f21eb78106d2d10558ea28ec981dc87743cc66/src/repository/BaseEntity.ts#L77-L85
It could be handled by just adding an option `{ withDeleted: true }`. But I feel like `findOneOrFail` method could be changed any time before the `0.3.0` version is released or even become obsolete, so I guess you should do it.
My last question was about something like that.

pleerock:
I _hate_ soft delete functionality we have. Simply because it brings too many design issues. Why `reload` should load soft-deleted record? I don't see a clear motivation. Other can argue that `reload` shouldn't load soft-deleted record, since its... soft-deleted.

spotykatch:
> I _hate_ soft delete functionality we have. Simply because it brings too many design issues. Why `reload` should load soft-deleted record? I don't see a clear motivation. Other can argue that `reload` shouldn't load soft-deleted record, since its... soft-deleted.

Well, it could have an option to load soft-deleted entities, while reload doesn't have any args.
Soft delete is awesome :+1: I guess it's very important for business apps, that need blame history. Business = donates :smile:

pleerock:
> Well, it could have an option to load soft-deleted entities, while reload doesn't have any args.

if we list all design issues soft delete has, we'll have to create "options for options".

balazsorban44:
Hi, I am just curious, what's the reasoning behind not opting for semantic versioning yet? (1.0.0). The project seems to be mature/old enough, no?

I would like to learn about the rationale. Is there a read-up on this?

I can see this will bring up a few issues similar to #8656

pleerock:
> Hi, I am just curious, what's the reasoning behind not opting for semantic versioning yet? (1.0.0). The project seems to be mature/old enough, no? I would like to learn about the rationale. Is there a read-up on this?

We actually could do it for 0.2.x if we knew that we'll have to maintain it for such a long time without breaking changes. Regarding to #8656 - it was accidental break, we didn't expect it to break something, so we could really have it on 1.x.x as well. But anyway, issue was fixed already.

The reason why I don't really like to release 1.0.0 now is because current API is quite old (just imagine how many things changed in the past 4 years) and now I have many many things that I would like to change and I don't want to make a brand new 1.0.0 release and then introduce many changes in 2.0.0. Maybe we can go this way as well, however I kinda feel uncomfortable with this way. We definitely need and will release 1.0.0 some day. Overall 0.3.0 has minimal number of API changes and it's kinda transitional before we go really different in 0.4.0. Thanks for taking on this and I appreciate your feedback a lot.

felix-gohla:
@pleerock It seems that 17e6e39035e5cb9b2a3bf26f3c19f5cb38513892 removed the export of `Connection`. Even though `Connection` being deprecated, is this intended? From a user standpoint, this could cause confusion if you say "Old `Connection` is still there, but now it's deprecated."

pleerock:
@felix-gohla yes it was a mistake, thanks for letting me know!

pleerock:
@Ginden yes we can, but it will also work on node 12, because I think we don't use anything node-14 specific. But yes, officially we can say we support 14+.

Q16solver:
Hey! I was just testing typeorm rc and migrating my project over and encountered some difficulties with extending a tree repository for posts compared to how I did it previously

previous:
image

now:
image

I required the protected functions of creating relationship maps and building children entity tree so I could implement my own way of only fetching up to say 20 of the top level children posts, but now I cannot access this, is there any other way I can migrate across?

image

pleerock:
@Q16solver thanks for testing! We changed the way custom repositories must be defined, please refer to [this documentation](https://github.com/typeorm/typeorm/blob/0.3.0/docs/custom-repository.md)

Q16solver:
@pleerock Sweet! I'll just do a simple refactor into TreeRepositoryUtils in the utils folder, I don't believe we will need to add new test cases for this as there shouldn't be any drastic changes, however, in saying that I do also realise that the extends method only covers the general repository case as well, and I'm thinking if I can add an override extends type in TreeRepository and then add a simple test case for that as well

Q16solver:
Trivial merge conflict fix PR #8754 for formatting spaces, it passes linting so it should be fine
Do we know when 0.3.0 is going to be released? Seems like the idea was to do at 1st March, then 10th now, but it's 17th

pleerock:
@Q16solver  I don't really expect to merge something non-trivial into `master` anymore, until 0.3.0 is merged into it. The only thing that is left is to update the documentation "docs" directory and improve CHANGELOG. Would be great to receive any help in this front.

PR #8618, opened 2022-02-09
fix: upsert should find unique index created by one-to-one relation

### LINKED ISSUE #8587: Upsert expects unique constraint for `OneToOne` columns which imply uniqueness already
I get this error:
```
An upsert requires conditions that have a unique constraint but none was found for conflict properties: original_response_id
```
For code like this:
```typescript
import TypeORM from 'typeorm';
import { Response } from './response';

@TypeORM.Entity()
export class Archive {
	@TypeORM.PrimaryGeneratedColumn('uuid')
		id!: string;

	@TypeORM.OneToOne(() => Response, response => response.id)
	@TypeORM.JoinColumn()
		originalResponse!: Response;
}
```

```typescript
		await this._archiveRepository.upsert(archive, [ 'original_response_id' ]);
```

I also tried changing `'original_response_id'` to `'original_response'` or `'originalResponse'` and it did not help.

This is the line where error is thrown:
https://github.com/typeorm/typeorm/blob/26581d0264268045496b43f7a0f338eb9f2b70cc/src/entity-manager/EntityManager.ts#L509-L511

joeflateau:
### Description of change

Fixes #8587 

One-to-one relations create a unique index, that index should be suitable for an upsert conflict target.

### Pull-Request Checklist

### DIFF
--- a/src/entity-manager/EntityManager.ts
+++ b/src/entity-manager/EntityManager.ts
@@ -39,7 +39,6 @@ import {ObjectLiteral} from "../common/ObjectLiteral";
 import {getMetadataArgsStorage} from "../globals";
 import {TypeORMError} from "../error";
 import {UpsertOptions} from "../repository/UpsertOptions";
-import { OrmUtils } from "../util/OrmUtils";
 
 /**
  * Entity manager supposed to work with any entity, automatically find its repository and call its methods,
--- a/src/entity-manager/EntityManager.ts
+++ b/src/entity-manager/EntityManager.ts
@@ -39,6 +39,7 @@ import {ObjectLiteral} from "../common/ObjectLiteral";
 import {getMetadataArgsStorage} from "../globals";
 import {TypeORMError} from "../error";
 import {UpsertOptions} from "../repository/UpsertOptions";
+import { OrmUtils } from "../util/OrmUtils";
 
 /**
  * Entity manager supposed to work with any entity, automatically find its repository and call its methods,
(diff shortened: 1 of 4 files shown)


src/entity-manager/EntityManager.ts
```diff
@@ -495,10 +496,16 @@ export class EntityManager {
...
         }
 
+        const relationMetadatas = metadata.relations.map(relation => (relation.embeddedMetadata ?? relation.entityMetadata));
+
         const uniqueColumnConstraints = [
             metadata.primaryColumns,
             ...metadata.indices.filter(ix => ix.isUnique).map(ix => ix.columns),
-            ...metadata.uniques.map(uq => uq.columns)
+            ...metadata.uniques.map(uq => uq.columns),
+            ...OrmUtils.flatten([
```
Ginden, line 505: No need to `.flatten`, just spread them directly.

pleerock:
Thanks for the fix! Quick question: do we really need a runtime check for column uniqueness? Usually we delegate everything the db, and if there is something that db cannot handle - it throws an error.

joeflateau:
@pleerock I think it's important. It's really there to protect MySQL/MariaDB users since those DB engines don't let you specify the conflict columns. If you "upsert" and there's no unique constraint for the columns you specify as your conflict columns then Postgres will throw an error but MySQL will just insert it and move on.

pleerock:
if that's natural MySQL behaviour, and users use MySQL's UPSERT with it's natural behaviour (orm is just a wrapper around what rdbms provides) then why do we need to prevent users from doing what they wanna do?

joeflateau:
my philosophy is that I like to catch footguns like this where the abstraction of "upsert" could cause some unexpected behavior due to db engine differences. your call though. I can remove the check altogether if you prefer that

pleerock:
The problem is that we definitely cannot catch all the errors - and when we handle some errors one way and other errors another way - it brings a confusion.

Also, there are users who use ORM with exist database without entities synchronization. Such users don't define entity  indices and uniques. They simply define entities and execute queries against exist database. Current approach fail for them.

> I can remove the check altogether if you prefer that

Yes, I would prefer these checks to be handles by database if it's possible. If you don't have a time or wish to do that - no worries, I'll merge this PR for now and someday we might refactor this code.

pleerock:
thank you a lot

Robert-Rendell-THE:
For anyone that wants typing, I did this:
```typescript
const conflictPaths: (keyof MyEntity)[] = ['columnA', 'columnB']
const opts: UpsertOptions<MyEntity> = { conflictPaths };
myEntityRepo.upsert({ ... }, opts)

PR #8637, opened 2022-02-13
fix: keep junction aliases short

### LINKED ISSUE #8627: Junction Aliases Are Too Long
## Issue Description

When using entities with long names and many-to-many relations, it can be that junction aliases get quite long.
This results in the driver truncating the junction aliases in our case (Postgres has a limitation of 63 characters).
Sometimes, it happens that two **different aliases get truncated to the same alias**.
Thus, the query fails.

Scenario: We have the entities `Device`, `Deployment`, `DeviceDeploymentAttempt`, `Label` (amongst others) with relations between them. In that case, a simple `EntityManager` `findOne` leads to junction aliases like `DeviceDeploymentAttempt_deployment_deployment_DeviceDeploymentAttempt_deployment_deployment_labels` and `DeviceDeploymentAttempt_deployment_deployment_DeviceDeploymentAttempt_deployment_deployment_devices`. Both are truncated to `DeviceDeploymentAttempt_deployment_deployment_DeviceDeploymentA` which is bad.

### Expected Behavior

The junction aliases should be unique und thus the query shouldn't fail.

### Actual Behavior

The junction aliases are truncated in the driver and the query fails with the following error:

```
QueryFailedError: table name "DeviceDeploymentAttempt_deployment_deployment_DeviceDeploymentA" specified more than once
```

### Steps to Reproduce

1. Create entities with long names or many nested levels.
2. Query them with `EntityManager` (no special options needed).

Didn't test it, but some entities like that should do the trick:

```typescript
@Entity()
export class SomeLongNameForAnEntityBecauseWeLikeNamesToBeDescriptive {
    @Column({
        type: 'string',
    })
    property!: string;
}

@Entity()
export class AnotherNameForAnEntityBecauseWeLikeNamesToBeDescriptive {
    @Column({
        type: 'string',
    })
    property!: string;

    @ManyToMany(() => SomeLongNameForAnEntityBecauseWeLikeNamesToBeDescriptive, { eager: true })
    @JoinTable({ name: 'junction' })
    others!: SomeLongNameForAnEntityBecauseWeLikeNamesToBeDescriptive[];
}
```

### My Environment

We use TypeORM with NestJS.

| Operating System    |    Debian Bullseye      |
| Node.js version     | 16.13.2  | 

### Additional Context

After a little debugging, this seems to happen in [`SelectQueryBuilder.ts:1563`](https://github.com/typeorm/typeorm/blob/1af30b1fb0d304b726347317ecfc5a71218e79ab/src/query-builder/SelectQueryBuilder.ts#L1563) and the following lines. there, the aliases are not shortened.
Probably, it would make sense to use `DriverUtils.buildAlias` there (but with `{shorten: false}`, because otherwise, it would still lead to non-unique aliases, sadly).

A place, where this could be included would be [`JoinAttribute.junctionAlias`](https://github.com/typeorm/typeorm/blob/0.2.41/src/query-builder/JoinAttribute.ts#L201) to not break [`QueryBuilder.replacePropertyNames`](https://github.com/typeorm/typeorm/blob/0.2.41/src/query-builder/QueryBuilder.ts#L612).

felix-gohla:
### Description of change

Currently, when having really long names for entities (in the code, not the table name) or deeply nested structures, joins do not work correctly because it can be that junction aliases do not get shortened. In the driver (e.g. Postgres has a limit of 63 characters), they will be truncated and then two different junction aliases map to the same string, resulting in the query failing because of aliases not being unique.

~~With this fix, junction aliases are being hashed and thus are short enough.~~

See below... **tl;dr: The fix was already implemented in 0.3.0. This PR adds a test for that fix.**

*Note that this does worsen the ability to read the query as the shortened / hashed aliases are harder to read.*
If you have suggestions on how to now loose the readability while still fixing the issue of junction names being too long, I'd be happy to change the behaviour.

Fixes #8627

### Pull-Request Checklist

Is there documentation that needs to be updated?

### DIFF
--- a/test/github-issues/8627/entity/long-name.entity.ts
+++ b/test/github-issues/8627/entity/long-name.entity.ts
@@ -1,0 +1,17 @@
+import {PrimaryGeneratedColumn, ManyToMany, JoinTable} from "../../../../src";
+import {Entity} from "../../../../src/decorator/entity/Entity";
+import {AnotherReallyLongNameForAnEntityBecauseThisIsNecessaryB, AnotherRealLongNameForAnEntityBecauseThisIsNecessaryC} from "./other-long-name.entity";
+
+@Entity()
+export class ThisIsARealLongNameForAnEntityBecauseThisIsNecessary {
+    @PrimaryGeneratedColumn()
+    id: number;
+
+    @ManyToMany(() => AnotherReallyLongNameForAnEntityBecauseThisIsNecessaryB, { eager: true })
+    @JoinTable({name: 'junction_b'})
+    long_name_for_bs!: AnotherReallyLongNameForAnEntityBecauseThisIsNecessaryB[];
+
+    @ManyToMany(() => AnotherRealLongNameForAnEntityBecauseThisIsNecessaryC, { eager: true })
+    @JoinTable({name: 'junction_c'})
+    long_name_for_cs!: AnotherRealLongNameForAnEntityBecauseThisIsNecessaryC[];
+}
--- a/test/github-issues/8627/entity/other-long-name.entity.ts
+++ b/test/github-issues/8627/entity/other-long-name.entity.ts
@@ -1,0 +1,14 @@
+import {PrimaryGeneratedColumn} from "../../../../src";
+import {Entity} from "../../../../src/decorator/entity/Entity";
+
+@Entity()
+export class AnotherReallyLongNameForAnEntityBecauseThisIsNecessaryB {
+    @PrimaryGeneratedColumn()
+    id: number;
+}
+
+@Entity()
+export class AnotherRealLongNameForAnEntityBecauseThisIsNecessaryC {
+    @PrimaryGeneratedColumn()
+    id: number;
+}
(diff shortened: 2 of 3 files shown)


pleerock:
I think I've pushed pretty much similar fix [here](https://github.com/typeorm/typeorm/pull/8616), will you be able to install and check it?

felix-gohla:
Cool! Didn't find that, sorry.

Yeah, I can try and will get back to you then.

Just out of curiosity: Would this mean, we would need to wait for 0.3.0? If yes, how long do you think this would take? 😊

pleerock:
0.3.0 is planned for the beginning of March. We can have the same fixes for both versions, but we need identical changes to prevent conflicts.

felix-gohla:
Thanks for your answer. I installed 0.3.0 into our system and it seems to work. ✅ 

I also ran the test suite including my test for that issue and it passed. Would it be possible to include the test in 0.3.0? 😊

Not sure whether back porting is necessary. As it is only like 3 weeks from now, I would argue that it makes more sense to adapt existing projects to 0.3.0.

felix-gohla:
Done, I only kept the test.

However, I couldn't get all the other tests back to running (because of some error with MongoDB but this is probably due to some other change with `0.3.0`).

pleerock:
Thanks!

Issue #8653, opened 2022-02-17
Reposiotry.softRemove() throws `QueryFailedError: column "<join_column>" does not exist` error


fabian-grajko:
## Issue Description

After bumping TypeORM version from `0.2.41` to `0.2.42` the `softRemove()` method started to throw exceptions.

### Expected Behavior

Records are soft-removed. No error is thrown.

### Actual Behavior

Records are not soft-removed. The QueryFailedError is thrown.

```
QueryFailedError: column "employee" does not exist
    at QueryFailedError.TypeORMError [as constructor] (/Users/fabiangrajko/Workspace/cyclops/looking-glass/api/src/error/TypeORMError.ts:7:9)
    at new QueryFailedError (/Users/fabiangrajko/Workspace/cyclops/looking-glass/api/src/error/QueryFailedError.ts:9:9)
    at PostgresQueryRunner.<anonymous> (/Users/fabiangrajko/Workspace/cyclops/looking-glass/api/src/driver/postgres/PostgresQueryRunner.ts:247:19)
    at step (/Users/fabiangrajko/Workspace/cyclops/looking-glass/api/node_modules/tslib/tslib.js:143:27)
    at Object.throw (/Users/fabiangrajko/Workspace/cyclops/looking-glass/api/node_modules/tslib/tslib.js:124:57)
    at rejected (/Users/fabiangrajko/Workspace/cyclops/looking-glass/api/node_modules/tslib/tslib.js:115:69)
    at processTicksAndRejections (node:internal/process/task_queues:96:5) {
  query: 'UPDATE "users" SET "deleted_at" = CURRENT_TIMESTAMP WHERE employee in ($1) RETURNING "id"',
  parameters: [ { id: 10 } ],
```

### Steps to Reproduce

1. Create 2 entities with relation (tested with `OneToOne` and `ManyToOne`) and set `{ cascade: true }`

```typescript
@Entity({ name: "employees" })
export class EmployeeEntity {
  @Column({ primary: true, generated: false })
  id: Id;

  @OneToOne(() => UserEntity, (user) => user.employee, { cascade: true })
  @JoinColumn({ name: "id" })
  user: UserEntity;

  @DeleteDateColumn({ name: "deleted_at" })
  deletedAt: Date | null;
}

@Entity({ name: "users" })
export class UserEntity {
  @PrimaryGeneratedColumn()
  id: Id;

  @OneToOne(() => EmployeeEntity, (employee) => employee.user)
[...]

PR #8654, opened 2022-02-17
Revert "Soft delete recursive cascade"

pleerock:
Reverts typeorm/typeorm#8436


denysshnurenko:
But why....?

akosasante:
Looks like it was it due to: https://github.com/typeorm/typeorm/issues/8653

Issue #7235, opened 2020-12-31
Use "INSERT...RETURNING" in MariaDB.


come25136:
### The Problem

I want to use "INSERT...RETURNING" in MariaDB.
It can't be used with the current TypeORM.
https://mariadb.com/kb/en/insertreturning/

### The Solution

I want to get the primary id of the affected INSERT query using "ON DUPLICATE KEY".

### Relevant Database Driver(s)

- [ ] `aurora-data-api`
- [ ] `aurora-data-api-pg`
- [ ] `better-sqlite3`
- [ ] `cockroachdb`
- [ ] `cordova`
- [ ] `expo`
- [ ] `mongodb`
- [ ] `mysql`
- [ ] `nativescript`
- [ ] `oracle`
- [ ] `postgres`
- [ ] `react-native`
- [ ] `sap`
- [ ] `sqlite`
- [ ] `sqlite-abstract`
- [ ] `sqljs`
- [ ] `sqlserver`

### Are you willing to resolve this issue by submitting a Pull Request?

- [ ] Yes, I have the time, and I know how to start.
- [ ] Yes, I have the time, but I don't know how to start. I would need guidance.
- [x] No, I don't have the time, although I believe I could do it if I had the time...
- [ ] No, I don't have the time and I wouldn't even know how to start.

imnotjames:
What version is this available from?

nix6839:
> INSERT ... RETURNING was added in MariaDB 10.5.0, and returns a resultset of the inserted rows.

From [Official document](https://mariadb.com/kb/en/insertreturning/)

pleerock:
If underlying driver (e.g. mysql package that is used for mariadb) support this - feature is relatively easy to implement. Feel free to contribute

PR #8673, opened 2022-02-19
feat: allow `{delete,insert}().returning()` on MariaDB

nix6839:
Closes: #7235

### Description of change

- Allow `delete().returning()` on MariaDB >= 10.0.5
- Allow `insert().returning()` on MariaDB >= 10.5.0
- Can check DML type(ex: `select`, `update`, `delete`) on
  `Driver.isReturningSqlSupported()` and `QueryBuilder.createReturningExpression()`

### Pull-Request Checklist

### DIFF
--- a/src/query-builder/DeleteQueryBuilder.ts
+++ b/src/query-builder/DeleteQueryBuilder.ts
@@ -1,12 +1,9 @@
-import {CockroachDriver} from "../driver/cockroachdb/CockroachDriver";
 import {QueryBuilder} from "./QueryBuilder";
 import {ObjectLiteral} from "../common/ObjectLiteral";
 import {EntityTarget} from "../common/EntityTarget";
 import {Connection} from "../connection/Connection";
 import {QueryRunner} from "../query-runner/QueryRunner";
 import {SqlServerDriver} from "../driver/sqlserver/SqlServerDriver";
-import {PostgresDriver} from "../driver/postgres/PostgresDriver";
-import {MysqlDriver} from "./../driver/mysql/MysqlDriver";
 import {WhereExpressionBuilder} from "./WhereExpressionBuilder";
 import {Brackets} from "./Brackets";
 import {DeleteResult} from "./result/DeleteResult";
@@ -213,8 +210,9 @@ export class DeleteQueryBuilder<Entity> extends QueryBuilder<Entity> implements 
     returning(returning: string|string[]): this {
 
         // not all databases support returning/output cause
-        if (!this.connection.driver.isReturningSqlSupported())
+        if (!this.connection.driver.isReturningSqlSupported("delete")) {
             throw new ReturningStatementNotSupportedError();
+        }
 
         this.expressionMap.returning = returning;
         return this;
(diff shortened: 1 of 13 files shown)


nix6839:
It seems that you need to upgrade the MariaDB version of docker and check the version in the code.

src/query-builder/InsertQueryBuilder.ts
```diff
@@ -400,7 +400,13 @@ export class InsertQueryBuilder<Entity> extends QueryBuilder<Entity> {
         }
 
         // add RETURNING expression
-        if (returningExpression && (this.connection.driver instanceof PostgresDriver || this.connection.driver instanceof OracleDriver || this.connection.driver instanceof CockroachDriver)) {
+        if (
+            returningExpression &&
+            (this.connection.driver instanceof PostgresDriver ||
+                this.connection.driver instanceof OracleDriver ||
+                this.connection.driver instanceof CockroachDriver ||
+                this.connection.driver instanceof MysqlDriver)
```
pleerock, line 408: don't we need to check for `connection.driver.options.type === "mariadb"` ? (instead of this.connection.driver instanceof MysqlDriver)

src/query-builder/DeleteQueryBuilder.ts
```diff
@@ -231,9 +232,13 @@ export class DeleteQueryBuilder<Entity> extends QueryBuilder<Entity> implements
         const whereExpression = this.createWhereExpression();
         const returningExpression = this.createReturningExpression();
 
-        if (returningExpression && (this.connection.driver instanceof PostgresDriver || this.connection.driver instanceof CockroachDriver)) {
+        if (
+            returningExpression &&
+            (this.connection.driver instanceof PostgresDriver ||
+                this.connection.driver instanceof CockroachDriver ||
+                this.connection.driver instanceof MysqlDriver)
```
pleerock, line 239: don't we need to check for `connection.driver.options.type === "mariadb"` ? (instead of this.connection.driver instanceof MysqlDriver)

pleerock:
> It seems that need to add the MariaDB version (>= 10.5.0) of docker and check the version in the code.

just update a docker-compose file

src/query-builder/DeleteQueryBuilder.ts
nix6839: It doesn't work with MySQL, only MariaDB. It's ok?

pleerock: yes, that's why `connection.driver.options.type === "mariadb"` check will work better then checking for instanceof here

nix6839: > don't we need to check for `connection.driver.options.type === "mariadb"` ? (instead of this.connection.driver instanceof MysqlDriver)

Opps... I didn't see the question mark above.
I check MariaDB [this line](https://github.com/typeorm/typeorm/pull/8673/files#diff-a5db73ecf35a00a64edc5bc64385dd06bcf618736cc3d9fd422091a178ffcf82R878-R879). If try to use `returning()`, will get an error if it is not MariaDB here.
So `instanceof MysqlDriver` may be left for later when MySQL supports `returning()`.

nix6839:
The current error seems to be that need to update the `createGeneratedMap()` function in `MysqlDriver.ts`.

pleerock:
Quite solid implementation. Thanks for contribution!

Issue #7736, opened 2021-06-10
.save() is throwing duplicate key value violates unique constraint on Postgresql (see #4122)


fabiensabatie:
## Issue Description

When trying to upsert an entity to a Postgresql DB, I get a `QueryFailedError: duplicate key value violates unique constraint`.

The issue was reported here too : https://github.com/typeorm/typeorm/issues/4122 and was closed, but the issue persists on Postgresql. 

### Expected Behavior

I currently have a simple case : 

````ts
this.save(user); // works 🎉
this.save(user); // should upsert the user
````

Where the User entity is defined as such :

````ts
@Entity()
export class CandleModel implements CandleProps {

  @PrimaryColumn("text")
  id: string; // idem with a number

  @Column("text")
  name: string;
}
````

### Actual Behavior

````ts
this.save(user); // works 🎉
this.save(user); // fails the second time ☹
````
 with a `QueryFailedError: duplicate key value violates unique constraint`.

### Steps to Reproduce

A simple typeorm + Postgresql  stack should suffice

### My Environment

I'm on a plain TS stack, nothing fancy atm.

### Relevant Database Driver(s)

- [ ] `aurora-data-api`
- [ ] `aurora-data-api-pg`
- [ ] `better-sqlite3`
- [ ] `cockroachdb`
- [ ] `cordova`
- [ ] `expo`
- [ ] `mongodb`
- [ ] `mysql`
- [ ] `nativescript`
- [ ] `oracle`
- [x] `postgres`
- [ ] `react-native`
- [ ] `sap`
- [ ] `sqlite`
- [ ] `sqlite-abstract`
- [ ] `sqljs`
- [ ] `sqlserver`

### Are you willing to resolve this issue by submitting a Pull Request?

- [ ] Yes, I have the time, and I know how to start.
- [x] Yes, I have the time, but I don't know how to start. I would need guidance.
- [ ] No, I don't have the time, although I believe I could do it if I had the time...
- [ ] No, I don't have the time and I wouldn't even know how to start.

imnotjames:
It looks like you aren't waiting for the `save` to complete asynchronously.

Is that the case or is your example incorrect?

fabiensabatie:
@imnotjames you are right, the example was incorrect, I added the await here (which is present in the original code) ;)

imnotjames:
What is `this`?  Can you post a runnable example?  I'm not sure I understand and I can't really replicate.

leric:
[...]

PR #8676, opened 2022-02-21
fix: fix entityManager.getId for custom join table

ranjan-purbey:
### Description of change

`entityManager.getId` currently returns undefined for an entity with composite primary key if primary key columns are also foreign keys with lazy relations, for e.g., in a custom join table. This commit tries to fix that.

The reason for `getId` returning `undefined` is that `columnMetadata.getEntityValueMap` doesn't take into account the fact that the value for lazy-relations is a getter which returns a `Promise`. This PR adds a check for this condition.

To reproduce the issue, refer https://github.com/ranjan-purbey/typeorm-composite-primary-key-test

Closes: #7736 (maybe)

### Pull-Request Checklist

### DIFF
--- a/src/metadata/ColumnMetadata.ts
+++ b/src/metadata/ColumnMetadata.ts
@@ -570,7 +570,12 @@ export class ColumnMetadata {
             return Object.keys(map).length > 0 ? map : undefined;
 
         } else { // no embeds - no problems. Simply return column property name and its value of the entity
-            if (this.relationMetadata && entity[this.relationMetadata.propertyName] && entity[this.relationMetadata.propertyName] instanceof Object) {
+            /**
+             * Object.getOwnPropertyDescriptor checks if the relation is lazy, in which case value is a Promise
+             * DO NOT use `entity[this.relationMetadata.propertyName] instanceof Promise`, which will invoke property getter and make unwanted DB request
+             * refer: https://github.com/typeorm/typeorm/pull/8676#issuecomment-1049906331
+             */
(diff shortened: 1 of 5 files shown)


pleerock:
Please add tests to this change

ranjan-purbey:
Test added

ranjan-purbey:
> Maybe we can simply add check for:
> 
> ```ts
>  && !(entity[this.relationMetadata.propertyName] instanceof Promise)
> ```
> 
> instead of `!Object.getOwnPropertyDescriptor(entity, this.relationMetadata.propertyName)?.get` ?
> 
> I think it will make more clear what is happening this way.

That will unintentionally call the getter function and thus make a database request in case it's a lazy relation

pleerock:
@ranjan-purbey but how come `instanceof` check makes a call of a method? It definitely should not.

ranjan-purbey:
@pleerock because lazy-relations are not normal functions. Instead they are (getters)[https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/get]. JavaScript getters are called as soon as the property is accessed.

For example, consider the following:
```javascript
const todos = {
  get items () {
    console.log("Items property was accessed");
    return ["task 1", "task 2"];
  }
};
```
If we try to check if `todos.items instanceof Array`, just by accessing the `items` property the `get` method is called and thus `console.log` is invoked

pleerock:
@ranjan-purbey thanks for the information. If so, we can merge this PR, however I would like to ask to add a bold comment on why this approach was used instead of `instanceof Promise` check to prevent in the future change of this code to `instanceof` check. Thank you!

ranjan-purbey:
Yes I agree that a comment clarifying this will be helpful for future reference. I will add one

ranjan-purbey:
@pleerock comment added

PR #8679, opened 2022-02-21
feat: add option to upsert to skip update if the row already exists and no values would be changed 

joeflateau:
### Description of change

This change adds an option to upsert methods that instructs postgresql to skip the update if the update would not change any of a row's values. Without this, postgres will write every row to disk for each row that is upserted even if nothing has actually changed. 

### Pull-Request Checklist


joeflateau:
thank you

pleerock:
thanks for another improvement :)

Issue #8652, opened 2022-02-17
Custom transformer overwritten for certain column types


azteca1998:
### Expected Behavior

Using a transformer on a column should disable further processing for said column.

More specifically, columns of type 'boolean', 'date', 'time', 'json', 'timestamp', 'datetime', Date, 'simple-array', 'simple-json', 'enum', 'simple-enum', and 'set'.

### Actual Behavior

The transformer is called as expected, but its value is overwritten by the library when preparing it to be persisted.

See steps to reproduce for an example and its output.

### Steps to Reproduce

1. Create a table with a column with any of the previously mentioned types.
2. Write a transformer that modifies the value.
3. Check the emitted SQL queries.

**package.json**
```json
{
  "name": "typeorm-bug",
  "version": "0.0.1",
  "private": true,
  "scripts": {
    "start": "ts-node src/index.ts"
  },
  "dependencies": {
    "sqlite3": "^5.0.2",
    "typeorm": "^0.2.42"
  },
  "devDependencies": {
    "@tsconfig/node16": "^1.0.2",
    "@types/node": "^17.0.18",
    "ts-node": "^10.5.0",
    "typescript": "^4.5.5"
  }
}
```

**tsconfig.json**
```json
{
  "extends": "@tsconfig/node16/tsconfig.json",
  "compilerOptions": {
    "emitDecoratorMetadata": true,
    "experimentalDecorators": true
  }
}
```

**src/index.ts**
```ts
import { createConnection, Entity, PrimaryColumn } from 'typeorm'


@Entity()
class MyEntity {
  @PrimaryColumn('datetime', {
    transformer: {
      from: (value: string) => new Date(value),
      to: (value: Date) => {
        value.setFullYear(value.getFullYear() + 1);
        const outValue = value.toUTCString();
        console.log(`Value transformed to '${outValue}'.`);
        return outValue;
      },
    },
  })
  declare myField: Date;
}


async function main() {
  const dbConn = await createConnection({
    type: 'sqlite',
    database: ':memory:',

    entities: [MyEntity],
    synchronize: true,
    logging: true,
  });

  await dbConn.manager.insert(MyEntity, {
    myField: new Date(),
  });
}

main()
  .catch(console.error)
```

**Program output**

Note the transformed string versus the query parameter: they have different formats.

```
yarn run v1.22.17
$ ts-node src/index.ts
query: PRAGMA foreign_keys = OFF
query: BEGIN TRANSACTION
[...]

PR #8680, opened 2022-02-21
fix: Add raw column transformer to address #8652.

azteca1998:
### Description of change

Fixes: #8652.

This change adds a separate column transformer called `rawTransformer` which is called after the normal transformers when serializing, and before them when deserializing. It is expected to emit and parse raw SQL values, and skip further value processing for said column.

The purpose of this PR is to address issue #8652, in which transformed values are in some cases effectively ignored. This is especially true for dates: attempting to use a third-party library and serializing the value manually to avoid timezone mismatch does not work as expected since the extra postprocessing done by the library re-parses the formatted string, and rewrites as a local time instead.

### Pull-Request Checklist


azteca1998:
Please check out the linked issue before merging since there's potentially another fix for the same bug.

azteca1998:
I'll try to dig out related issues when I have some extra time, thanks.

Have you checked the linked issue? There's a potentially easier alternative which is to return early when the transformed value is a function, and use that function to generate the SQL code. Is that a better solution? It certainly does not require another option, just an if statement with a return. The downside is what to do when there are various transformers.

pleerock:
Closing it for a better times. However I still think we do need this feature.

PR #8688, opened 2022-02-22
fix: ng webpack default import

### LINKED ISSUE #8674: Error: Module parse failed: 'import' and 'export' may appear only with 'sourceType: module'
typeorm: 0.2.43
ionic: 6.0.8
angular:12

./node_modules/typeorm/browser/index.js:3:0 - Error: Module parse failed: 'import' and 'export' may appear only with 'sourceType: module' (3:0)
File was processed with these loaders:
 * ./node_modules/@angular-devkit/build-angular/src/babel/webpack-loader.js
 * ./node_modules/@ngtools/webpack/src/ivy/index.js
 * ./node_modules/source-map-loader/dist/cjs.js
You may need an additional loader to handle the result of these loaders.
| /*!
|  */
> import "reflect-metadata";

giladgd:
### Description of change
Instruct Webpack to import the CommonJS version or browser-native ESM version of TypeORM directly when using TypeORM's default import, ignoring `index.mjs` file meant for nodejs ESM projects.

In nodejs the `index.mjs` file is used to refer to the same loaded instance of TypeORM from both ESM and CommonJS imports, it's needed in order to maintain the same state across ESM and CommonJS imports that may occur in different places in the same project.
When using Webpack, only one module system can be used to import the same library, so providing different import methods should not cause state problems.

Closes #8674

### Pull-Request Checklist
- [ ] There are new or updated unit tests validating the change - _Testing the way the module is imported is problematic to do from within the module itself_

### DIFF
--- a/package.json
+++ b/package.json
@@ -11,8 +11,15 @@
   },
   "exports": {
     ".": {
-      "import": "./index.mjs",
-      "require": "./index.js"
+      "node": {
+        "import": "./index.mjs",
+        "require": "./index.js"
+      },
+      "browser": {
+        "require": "./index.js",
+        "import": "./browser/index.js",
+        "default": "./index.js"
+      }
     },
     "./browser": "./browser/index.js",
     "./*.js": "./*.js",
@@ -24,7 +31,6 @@
   "main": "./index.js",
   "module": "./index.mjs",
   "types": "./index.d.ts",
-  "type": "commonjs",
   "browser": {
     "./browser/connection/ConnectionOptionsReader.js": "./browser/platform/BrowserConnectionOptionsReaderDummy.js",
     "./browser/connection/options-reader/ConnectionOptionsXmlReader.js": "./browser/platform/BrowserConnectionOptionsReaderDummy.js",


package.json
```diff
@@ -24,7 +31,6 @@
   "main": "./index.js",
   "module": "./index.mjs",
   "types": "./index.d.ts",
-  "type": "commonjs",
```
giladgd, line 27: nodejs documentation encourages to explicitly specify `"type": "commonjs"` for CommonJS modules to be future-proof, currently not explicitly specifying it works exactly the same on nodejs and there are no plans to change that behavior at the moment.
However, this seems to break Webpack's module system resolver at the moment, and until this is fixed at Webpack's end, it would be better to remove it for now.

pleerock:
Hate modules-related issues 😆 Thanks a lot for the fix!

algife:
Hey guys. First comment of this kind for me. It is giving me this error, using Nx workspace in an angular 13+ project.

image

algife:
Hey guys. I am experimenting this problem and as you seem to be working on it exactly now here is my feedback.

It is giving me this error, using Nx workspace in an angular 13+ project, along with TypeORM

image

This is my first comment/post of this kind btw

PR #8712, opened 2022-03-01
fix: allow clearing database inside a transaction

varanauskas:
Possible fix for: #8527

### Description of change

This is one of the possible fixes for the issue #8527

It allows to run `clearDatabase` inside a transaction by making `clearDatabase` to check if another transaction is already running and not creating a new transaction in that case

It currently is only developed for/tested with Postgres, if this is the right direction same fix could be applied for other query runners

### Pull-Request Checklist


varanauskas:
This is re-opening of my previous pull request #8529

I have merged the rebased with master and adapted the necessary changes to all drivers that use transactions when clearing database

pleerock:
Thank you

PR #7630, opened 2021-05-07
Feature/cache option ignore errors

aniketawati:
### Description of change

In case of cache errors like network failure, redis out of memory etc, default behaviour is to return that error to the caller. 
But there are some cases where cache is only there to minimize database calls and its failure can be tolerated.
Issue https://github.com/typeorm/typeorm/issues/926 has been pending for this purpose.  

This change introduces cacheoption - `ignoreErrors` and based on its value decides to throw the cache error or not. 

Closes #926 

### Pull-Request Checklist

- [x] Code is up-to-date with the `master` branch
- [x] `npm run lint` passes with this change
- [x] `npm run test` passes with this change
- [x] This pull request links relevant issues as `Fixes #0000`
- [ ] There are new or updated unit tests validating the change
- [x] Documentation has been updated to reflect this change
- [x] The new commits follow conventions explained in [CONTRIBUTING.md](https://github.com/typeorm/typeorm/blob/master/CONTRIBUTING.md)

PR #8725, opened 2022-03-03
fix: Stop crashing when redis is offline and ignoreErrors on cache op…

mesinger:
### Description of change
PR #7630 introduced the ```ignoreErrors``` flag for ```ConnectionOptions```. For me this updated did not fix my problem (also mentioned in the same PR #7630).
From my subjective opinion, typeorm should still function even though redis caching is not available. typeorm will be able to query like regular from the database overall increasing the resilience of systems using typeorm.

This fix adds an ```error``` callback for the internally used redis client, logging instead of throwing an error when ```ignoreErrors``` is set.

### Pull-Request Checklist


src/cache/RedisQueryResultCache.ts
```diff
@@ -54,6 +54,13 @@ export class RedisQueryResultCache implements QueryResultCache {
                 ...cacheOptions?.options,
                 legacyMode: true
             });
+            this.client.on("error", (err: any) => {
```
Ginden, line 57: Wouldn't it be better to do:

```ts
if (typeof this.connection.options.cache === "object" && this.connection.options.cache.ignoreErrors) {
  this.client.on("error", (err: any) => {
    this.connection.logger.log("warn", err);
  });
```

pleerock:
Thank you for contribution! 🎉

Juantrix123:
This PR introduces an error in RedisQueryResultCache class, since the connect method of the redis client is not called.
This causes (ClientClosedError: The client is closed) when cache is used

mesinger:
You are right, should be fixed here https://github.com/typeorm/typeorm/pull/8894

PR #8748, opened 2022-03-14
fixes #8747

### LINKED ISSUE #8747: QueryBuilder update handles Date objects wrong on a `ManyToOne` relation ship.
## Issue Description
When updating a foreign key that uses a Date object, TypeORM seems to ignore the Date object.

### Expected Behavior
Correctly handling the `Date` object

### Actual Behavior
Some internal code is handling the Date object wrong.

### Steps to Reproduce

```typescript
@Entity()
export class Car extends BaseEntity {
    uuid: string

  @Column({ type: 'timestamp', precision: 3, nullable: true })
    latestRecordTimestamp?: Date

  @OneToMany(() => Record, record => record.car)
    records: Record[]

  @ManyToOne(() => Record)
  @JoinColumn([
    { name: 'uuid', referencedColumnName: 'carUuid' },
    { name: 'latestRecordTimestamp', referencedColumnName: 'timestamp' }
  ])
    latestRecord?: Record
}
```

```typescript
@Entity()
export class Record extends BaseEntity {
  @CreateDateColumn({ precision: 3, primary: true })
    timestamp: Date

  @Column({ type: 'uuid', primary: true })
    carUuid: string

  @ManyToOne(() => Car, car => car.records, { onDelete: 'CASCADE' })
  @JoinColumn({ name: 'carUuid' })
    car: Car
}
```

```typescript
await getManager().createQueryBuilder()
  .update(Car)
  .set({ latestRecordTimestamp: record.timestamp })
  .where('uuid = :uuid', { uuid: car.uuid })
  .execute()
```

### Temporary Workaround

```typescript
await getManager().createQueryBuilder()
  .update(Car)
  .set({ latestRecordTimestamp: record.timestamp.toISOString() })
  .where('uuid = :uuid', { uuid: car.uuid })
  .execute()
```

### My Environment

| Operating System    |   MacOS 12.1       |
| Node.js version     | v16.13.0 | 
| Typescript version  | 4.5.5  | 
| TypeORM version     | 0.2.45  | 

jorenvandeweyer:
### Description of change
Fixes #8747 

Handles the Date object correctly.

### Pull-Request Checklist

- [ ] Documentation has been updated to reflect this change (NA)


pleerock:
can you please fix the conflicts and add the test case so we won't have this issue in the future again? thanks.

jorenvandeweyer:
I agree a test case should be added, I think it would be better in a different issue/pr. Testing this would require a good timeseries database setup imo.

@pleerock if you agree I'll create a new issue and create a separate pull request. This way we can already merge/close this one.

pleerock:
why not to make it here? I need test to check what exactly this change does and what error we have without this change.

jorenvandeweyer:
@pleerock I added a test covering the issue.

pleerock:
tests are failing

pleerock:
yes I'll do it once I get the time to checkout your branch and do manual testing

pleerock:
looks like it breaks because of `precision: 3`. Maybe mysql doesn't play well with this value. Do you really need precision in this test?

pleerock:
after removing `precision` I found another problem - mysql tests would not be able to pass because you rely on primary column not explicitly set in your code, instead it set by database's `DEFAULT` (since you are using CreateDateColumn). After TypeORM saves your entity it needs entity id-s to perform mapping operations, it doesn't work if it doesn't have entity ids. So, this scenario is not supported. It works in postgres because it supports `RETURNING` statement which is not supported by mysql and some other drivers.

So, again my question is regarding to your test - why did you use **primary** `CreateDateColumn` ? Is it required in order to fix what you are trying to fix?

jorenvandeweyer:
> looks like it breaks because of `precision: 3`. Maybe mysql doesn't play well with this value. Do you really need precision in this test?

Not sure if it is needed in mysql but it is needed in PostgreSQL. JavaScript timestamps only have a precision of 3 decimals on a second, postgres has 6 by default. So it would be impossible to set the foreign key of the relationship since the timestamps would differ. 

>after removing precision I found another problem - mysql tests would not be able to pass because you rely on primary column not explicitly set in your code, instead it set by database's DEFAULT (since you are using CreateDateColumn). After TypeORM saves your entity it needs entity id-s to perform mapping operations, it doesn't work if it doesn't have entity ids. So, this scenario is not supported. It works in postgres because it supports RETURNING statement which is not supported by mysql and some other drivers.

Thank you! I totally forgot about that. I rewrote the test case, instead of using DEFAULT I used `new Date()` in the code instead.

>So, again my question is regarding to your test - why did you use primary CreateDateColumn ? Is it required in order to fix what you are trying to fix?

CreateDateColumn is not required for the test, the composite primary key only requires a timestamp. I removed the createDateColumn in the latest commit. 

Now it's working for both postgres & mysql.

pleerock:
That's why I ask to keep tests as simple as possible and contain the only code test need. As you can see both of us waste lot of time.

AlexMesser:
thank you for contribution!

Issue #8768, opened 2022-03-21
[0.3] Datasource not set on entities when using path in entities datasource options


niranjan94:
## Issue Description

When globs are specified in `entities` of `DataSource` options like the following, `entity.useDataSource()` is not automatically called as expected on the entities and as a result the error `Error: DataSource is not set for this entity.` is thrown when trying to perform operations with the entity (Active Record pattern)

```
const dataSource = new DataSource({
    type: "mysql",
    host: "localhost",
    port: 3306,
    username: "root",
    password: "admin",
    database: "test",
    synchronize: true,
    entities: [__dirname + "/entity/*.{js,ts}"],
})
```

### Expected Behavior

The operations performed on the entities should succeed.

### Actual Behavior

The error `Error: DataSource is not set for this entity.` is thrown when trying to perform operations with the entity (Active Record pattern)

```
Error: DataSource is not set for this entity.
```

### Steps to Reproduce

This sample should be able to reproduce the issue.
https://github.com/typeorm/typeorm/tree/master/sample/sample11-all-types-entity

### My Environment

| Dependency          | Version  |
| ---                 | ---      |
| Operating System    |    MacOS Montery 12.3      |
| Node.js version     | v16.13.0  | 
| Typescript version  | 4.6.2  | 
| TypeORM version     | 0.3.1  | 

### Additional Context

Looking at https://github.com/typeorm/typeorm/blob/d61f857ce9447a25d074810759fd5c4aad9a08e3/src/data-source/DataSource.ts#L631-L678 , it seems like the array of paths/globs is never resolved into the respective classes anywhere here. As a result this bit is never run https://github.com/typeorm/typeorm/blob/d61f857ce9447a25d074810759fd5c4aad9a08e3/src/data-source/DataSource.ts#L674-L676 

### Relevant Database Driver(s)

| DB Type           | Reproducible  |
|-------------------| ---           |
| `aurora-mysql`    | no            |
| `aurora-postgres` | no            |
| `better-sqlite3`  | no            |
| `cockroachdb`     | no            |
| `cordova`         | no            |
| `expo`            | no            |
| `mongodb`         | no            |
| `mysql`           | no            |
| `nativescript`    | no            |
| `oracle`          | no            |
| `postgres`        | yes            |
| `react-native`    | no            |
| `sap`             | no            |
| `sqlite`          | no            |
| `sqlite-abstract` | no            |
| `sqljs`           | no            |
| `sqlserver`       | no            |

[...]

PR #8769, opened 2022-03-21
fix: resolve entities correctly in datasource when globs are specified

niranjan94:
Closes: #8768

### Description of change

In the 0.3 release, DataSource stopped resolving entities correctly when specified as Globs. This results in the data source information not being set correctly into the entities. This PR fixes it by correctly importing the classes when globs are specified. 

### Pull-Request Checklist

### DIFF
--- a/src/data-source/DataSource.ts
+++ b/src/data-source/DataSource.ts
@@ -37,6 +37,7 @@ import { TypeORMError } from "../error/TypeORMError"
 import { RelationIdLoader } from "../query-builder/RelationIdLoader"
 import { DriverUtils } from "../driver/DriverUtils"
 import { InstanceChecker } from "../util/InstanceChecker"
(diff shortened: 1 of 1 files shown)


src/data-source/DataSource.ts
```diff
@@ -645,6 +646,17 @@ export class DataSource {
         const flattenedEntities = ObjectUtils.mixedListToArray(
             this.options.entities || [],
         )
+
+        // resolve globs to the respective entity classes
+        flattenedEntities.push(
```
pleerock, line 651: doesn't look like a proper solution. We already do import classes from directories above by calling `buildEntityMetadatas` method.

pleerock:
Closing in favor of #8778

src/data-source/DataSource.ts
niranjan94: Ah okay! Thanks for the clarification 😄
Probably need to look through the codebase more.

PR #8791, opened 2022-03-24
feat: Add support for insert with alias #4003

### LINKED ISSUE #4003: Support alias in insert into
**Issue type:**

[ ] question
[ ] bug report
[x] feature request
[ ] documentation issue

**Database system/driver:**

[x] `postgres`

PostgreSQL supports alias in INSERT INTO statement since v9.5, eg. `INSERT INTO "tablename" as "alias"` (https://www.postgresql.org/docs/9.5/sql-insert.html). It would be nice to be able to use that in insert query builder.

Our use case is incrementing a counter using `INSERT ... ON CONFLICT` statement and using table prefixes at the same time. eg:

```
this.createQueryBuilder()
      .insert()
      .into(Velocity)
      .values(velocity)
      .onConflict(
        `("date", "symbol") DO UPDATE SET value = v1_velocity.value + :velocity `
      )
      .setParameter('velocity', velocity.value)
      .execute();
```

`v1_` is our prefix here and it shouldn't have to be hardcoded in the query. I think we can workaround this by using `this.manager.connection.driver.options.entityPrefix` but I think that using explicit alias would be clearer.

zbarbuto:
### Description of change

Adds support for alias on insertQueryBuilder. This is particularly useful when dealing with `ON CONFLICT` as you can reference the original table to do an update or nothing based on a condition.

Fixes #4003 

As an aside question - would it be possible to un-deprecate `onConflict` as it is still useful for this sort of complex `ON CONFLICT` work which isn't possible with `orUpdate` or `orIgnore`

### Pull-Request Checklist

- [x] `npm run test` passes with this change (*tests for the code I changed at least)
- [ ] Documentation has been updated to reflect this change (See above - `onConflict` has no docs right now)

### DIFF
--- a/src/query-builder/InsertQueryBuilder.ts
+++ b/src/query-builder/InsertQueryBuilder.ts
@@ -410,6 +410,13 @@ export class InsertQueryBuilder<Entity> extends QueryBuilder<Entity> {
 
         query += `INTO ${tableName}`
 
+        if (
+            this.alias !== this.getMainTableName() &&
+            this.connection.driver.options.type === "postgres"
+        ) {
+            query += ` AS "${this.alias}"`
+        }
+
         // add columns expression
         if (columnsExpression) {
             query += `(${columnsExpression})`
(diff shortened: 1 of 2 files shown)


pleerock:
Thank you for contribution! 🎉

PR #8813, opened 2022-03-27
feat: automatically add files generated via the CLI to the `data-source.ts` file

giladgd:
### Description of change
When generating an entity/migration/subscriber file via the CLI, the generated file will be added to the relevant place in the `data-source.ts` file by default.
This behavior can be disabled by setting the flag `--addImport false` or `--no-addImport`.

It uses the `typescript` library (that most users already have installed in their projects) to parse the `data-source.ts` file into an AST, it creates a new AST and then converts it back to code.

Updating the `data-source.ts` is now the default behavior since most users currently already have to manually edit `data-source.ts` after using a generate command, so having this turned on by default would save some repetitive work.

For example, let's say we have this file:
```typescript
// data-source.ts

import "reflect-metadata";
import { DataSource } from "typeorm";
import { User } from "./entities/User.js";
import { User1648322295035 } from "./migrations/1648322295035-User.js";

export const AppDataSource = new DataSource({
    type: "sqlite",
    database: "db.db",
    synchronize: false,
    logging: false,
    entities: [
        User
    ],
    migrations: [
        User1648322295035
    ]
});
```

And then we run this:
```bash
cd entities
npx typeorm-ts-node-esm entity:create --dataSource ../data-source.ts Car
```

Then `data-source.ts` will automatically update to this:
```typescript
// data-source.ts

import "reflect-metadata";
import { DataSource } from "typeorm";
import { User } from "./entities/User.js";
import { User1648322295035 } from "./migrations/1648322295035-User.js";
import { Car } from "./entities/Car.js";

export const AppDataSource = new DataSource({
    type: "sqlite",
    database: "db.db",
    synchronize: false,
    logging: false,
    entities: [
        User,
        Car
    ],
    migrations: [
        User1648322295035
    ]
});
```

The same will happen for all of these commands by default:
* `migration:generate`
* `migration:create`
* `entity:create`
* `subscriber:create`

### Pull-Request Checklist
- [ ] Documentation has been updated to reflect this change - I'm not sure this feature needs documentation other than being in a changelog


pleerock:
Sounds like a good idea, but from the code perspective they can be defined any way, e.g.:

1.
```ts
// entities.ts
export * from "./post.entity"
export * from "./category.entity"

// data-source.ts
import * as entities from "./entities"

    entities: entities,
```

2.
```ts
// entities.ts
export {Post} from "./post.entity"
export {Category} from "./category.entity"

export const Entities = [
    Post,
    Category,
]

// data-source.ts
import { Entities } from "./entities"

    entities: Entities,
```

3. 
```ts
// modules/post/entities.ts
export {Post} from "./post.entity"
export {Category} from "./category.entity"

export const PostEntities = [
    Post,
    Category,
]

// modules/user/entities.ts
export {User} from "./user.entity"
export {Photo} from "./photo.entity"

export const UserEntities = [
    User,
    Photo,
]

// data-source.ts
import { GlobalAppEntity1 } from "./app/GlobalAppEntity1"
import { GlobalAppEntity2 } from "./app/GlobalAppEntity2"
import { UserEntities } from "./modules/user/entities"
import { PostEntities } from "./modules/post/entities"

    entities:[
  GlobalAppEntity1,
  GlobalAppEntity2,
  ...UserEntities,
  ...PostEntities,
],
```

and million other combinations, you got the idea.

Personally, I used all approaches I showed above ^ and never used the approach of a single data-source with all entities defined in it, because all my project has many entities and data-source file in that case become mess.

giladgd:
@pleerock I can add support for scenarios 1 and 2 pretty easily, scenario 3 is possible to implement but is a bit more complex.

I think this feature is particularly useful for migrations, since organizing the migrations into different folders is much less important, and having the `migration:generate` command be completely automatic would be super useful.

Having this feature won't affect the current usages of `:create` commands as it only adds support for updating the DataSource reference when the `--dataSource` flag is passed (and `--addImports` isn't disabled)

For example:
```bash
# data-source.ts won't be updated
npx typeorm entity:create entities/User
npx typeorm-ts-node-esm migration:create migration/AddUser
npx typeorm-ts-node-commonjs subscriber:create subscribers/UserChange

# data-source.ts will be updated
npx typeorm-ts-node-esm entity:create --dataSource data-source.ts entities/User
npx typeorm-ts-node-esm migration:create --dataSource data-source.ts migration/AddUser
npx typeorm-ts-node-commonjs subscriber:create --dataSource data-source.ts subscribers/UserChange
```

But since the `migration:generate` already requires you to pass `--dataSource`, in that case, the migrations would be added by default.

It's worth noting that in case the created/generated file couldn't be automatically added to the DataSource, the file would still be created/generated but no changes would be applied to the DataSource file. (and a warning log would appear in the console)
This means that support for more cases could be added in the future to make this feature more convenient and versatile, but even when having an unsupported case, the CLI will still work like it is today.

giladgd:
@pleerock I've added support for scenarios 1 and 2 you mentioned

pleerock:
I appreciate your efforts, but I think complexity added to the codebase doesn't worth it. Scenarios 1, 2, 3 were just examples,  there can be hundred different other scenarios and for sure we cannot cover them all. Having 1,2 covered also doesn't make sense, we don't know how many users use these approaches and how many use other approaches. Complexity brought into codebase doesn't worth it. 

I really really like the idea, but I don't see any simple and complete way to implement this idea.

pleerock:
> What do you think about using all of this logic from an external module instead?
> I can create a new node module with the functionality in this PR (of trying to update .ts files code) and import and use it from the typeorm CLI, this way we could still have this functionality in the CLI, but this logic could be improved and maintained independently of typeorm.

Or it's logic can be abandoned by it's maintainers and typeorm's CLI won't work anymore, until we either remove this feature from typeorm (again users will start complain) or move this logic into typeorm's codebase or take care of this package maintenance. 

Don't get me wrong - for the years of open source experience I've seen many people abandoned their projects, and when you depend on them it's always pain for **you**. I don't blame them, it's natural and it's how open source works. But I don't want to deal with such cases again and again. I want to keep simplicity and maintainability.  

Of course nothing keeps you from creating `typeorm-better-cli` package and implement this functionality in there.

pleerock:
If you ask me on how to implement this feature, I would go the following way. Use convention over configuration.

Imagine:

```ts
// data-source.ts
import "reflect-metadata";
import { DataSource } from "typeorm";
import * as entities from "./entities"
import * as migrations from "./migrations"
import * as subscriptions from "./subscriptions"

export const AppDataSource = new DataSource({
    type: "sqlite",
    database: "db.db",
    synchronize: false,
    logging: false,
    entities: entities,
    migrations: migrations,
    subscriptions: subscriptions,
});
```

1. if data source file has `import * as entities from` then:
2. if file `"./entities"` is empty or contains only comment lines (staring from "//" or "/*") then:
3. add `export * from "./newly-generated-entity" to it

That's it. `typeorm init` must generate project with imports.

and btw I bet that in the future the only syntax `DataSource` is going to support is:

```ts
entities: {
  Post: Post,
  Photo: Photo,
}
```

Which is covered by `import * as entities from "./entities"` approach. This limitation is due to type-safety features (like typed repositories) I'm planning to implement.

giladgd:
@pleerock The current implementation supports the convention you described, but regarding step `2.`, a newly generated file will be added even if the `entities.ts` file is not empty.

For example, given this code:
```typescript
// data-source.ts
import "reflect-metadata";
import { DataSource } from "typeorm";
import * as entities from "./entities"
import * as migrations from "./migrations"
import * as subscriptions from "./subscriptions"

export const AppDataSource = new DataSource({
    type: "sqlite",
    database: "db.db",
    synchronize: false,
    logging: false,
    entities: entities,
    migrations: migrations,
    subscriptions: subscriptions,
});
```
```typescript
// entities.ts
export * from "./entities/User"
```

When running this command:
```bash
npx typeorm-ts-node-esm entity:create --dataSource data-source.ts entities/Post
```

The the result will be:
```typescript
// entities.ts
export * from "./entities/User"
export * from "./entities/Post"
```

I've simplified the code and added comments to make it much easier to understand.

giladgd:
I've updated the `typeorm init` command to generate a project according to the convention you described

pleerock:
1000k LOC using typescript compiler cannot be simple. I'm not going to maintain this in the future. Can't we make it in a ~100 LOC without using typescript checker at all?

giladgd:
@pleerock Finding the usage of `DataSource`, and then the relevant property in the first argument, and then its value (which can be written in a lot of different ways), and then finding the reference of that location is very hard to do with regex in a stable and robust way. (in typescript you could add type related code that could easily break a regex, and there're so many variations of this that can be used, so predicting and supporting even a small subset of them would be very complicated)

Using the typescript compiler, while the code is longer, is actually a lot easier, since it does all the parsing and understands the code as precisely as possible. (since the code is made for it in the first place)

Also, even though the code is longer, it'll actually require much less maintenance, since future changed/added features in the typescript language itself will be supported in the compiler, thus the code will continue to work. (as far as I've seen, they don't make breaking changes in the compiler API, and instead they only deprecate stuff but continue to support it, which is safe in the long term)
The code uses whichever typescript version the user installs (with a minimum version requirement), so even when a new typescript version with new features comes out, the code will continue to work and understand the user-written code no matter what typescript features are used.

Because the code is relatively long, I've made it as configurable as I could to make it as future-proof as I can predict by now, so even if the name of `DataSource` or the name/type of a property on it changes in the future, the configuration can be changed without needing to dive into the code.

I've also added tests of this functionality to the `entity:create` command in the CLI, which covers this code, so it should help with its maintainability if it'll be needed.

pleerock:
I'm against further complexity of the codebase. And this functionality brings enough complexity to reject it. I'm always thinking what functionality we should deprecate and remove to simply TypeORM maintenance. And I don't want to merge this and in a few months deprecate and remove it.

You can create a separate cli tool for this purpose if you want, I'll support you where it's needed from TypeORM side. Highly appreciate all your efforts.

giladgd:
I've created [typeorm-codebase-sync](https://www.npmjs.com/package/typeorm-codebase-sync) CLI and opened PR #8847 to add it to the list of TypeORM extensions

Issue #8698, opened 2022-02-23
Update to >= typeorm 0.2.42 gives compilation error for projects using nestjsx/crud


in-live-md:
Update to >= typeorm 0.2.42 gives compilation error for projects using nestjsx/crud
Also see https://github.com/nestjsx/crud/issues/768

### Expected Behavior

Expected to compile

### Actual Behavior

Typescript errors:

```
node_modules/@nestjsx/crud-typeorm/lib/typeorm-crud.service.d.ts:30:5 - error TS2416: Property 'createOne' in type 'TypeOrmCrudService<T>' is not assignable to the same property in base type 'CrudService<T>'.
  Type '(req: CrudRequest, dto: DeepPartial<T>) => Promise<T>' is not assignable to type '(req: CrudRequest, dto: T) => Promise<T>'.
    Types of parameters 'dto' and 'dto' are incompatible.
      Type 'T' is not assignable to type 'DeepPartial<T>'.

30     createOne(req: CrudRequest, dto: DeepPartial<T>): Promise<T>;
       ~~~~~~~~~
```

### Steps to Reproduce

1. install nestjsx/crud dependencies per https://github.com/nestjsx/crud
2. use crud in your app

```typescript
// insert code here
```

### My Environment

| Dependency          | Version  |
| ---                 | ---      |
| Operating System    |      MacOS 11.6  |
| Node.js version     | 14.17.1  | 
| Typescript version  | 4.5.5| 
| TypeORM version     | 02.2.42/43 | 

### Additional Context

Looks like typing improvements may be the root cause see last 2 lines of diff :
```
index 17223f43..0b8b0f70 100644
--- a/src/entity-manager/EntityManager.ts
+++ b/src/entity-manager/EntityManager.ts
@@ -163,17 +163,17 @@ export class EntityManager {
     }

     createQueryBuilder<Entity>(entityClass?: EntityTarget<Entity>|QueryRunner, alias?: string, queryRunner?: QueryRunner): SelectQueryBuilder<Entity> {
         if (alias) {
@@ -247,7 +247,7 @@ export class EntityManager {
             return metadata.create(this.queryRunner);

         if (Array.isArray(plainObjectOrObjects))
-            return plainObjectOrObjects.map(plainEntityLike => this.create(entityClass as any, plainEntityLike));
+            return (plainObjectOrObjects as DeepPartial<Entity>[]).map(plainEntityLike => this.create(entityClass, plainEntityLike));
```

[...]

PR #8817, opened 2022-03-27
fix: Update DeepPartial for usage of generics with Repository class

### LINKED ISSUE #8681: DeepPartial simplification breaks the `.create()` and `.save()` method in certain cases.
## Issue Description

After the `DeepPartial` simplication of #8187, TypeScript is no longer able to match certain situations using the `.create()` and `.save()` functions. Specifically, when we are trying to pass a relation that is nested within a deep partial entity instance which is also an array, it will no longer match the `DeepPartial<Entity>[]` type signature and fail. See example code.

### Steps to Reproduce

```typescript
@Entity()
class Item {
  @PrimaryGeneratedColumn()
  id!: number

  @ManyToOne(() => Thing, thing => thing.items)
  thing!: Thing
}

@Entity()
class Thing {
  @PrimaryGeneratedColumn()
  id!: number

  @OneToMany(() => Item, item => item.thing)
  items!: Item[]
}


export async function test() {
  const myThing: DeepPartial<Thing> = {
    items: [{ id: 1 }, { id: 2 }]
  }

  const thing = getManager().create(Thing, myThing) // OK
  await getRepository(Thing).save(myThing) // OK

  const items = getManager().create(Item, myThing.items) // Not OK, myThing.items is of type `(Item | undefined)[] | undefined`
  // - or -
  await getRepository(Item).save(myThing.items) // Not OK for the same reason as above.
}
```

### My Environment

| Node.js version     | 4.17.5  | 
| Typescript version  | 4.5.5  | 
| TypeORM version     | 0.2.43  | 

nelsonfleig:
### Description of change

Update DeepPartial to work with a generic passed to Repository then testing on an abstract service.

This issue is related with Fixes #8681 

### Pull-Request Checklist

- [x] This pull request links relevant issues as `Fixes #8681`


pbrn46:
It seems after the `create()` call, a non-deeppartial is returned to be used for the `save()` call, The `save()` call doesn't like it, since apparently a _generic_ T can't be used as a parameter expecting a _generic_ DeepPartial<T> in the .`save()` signatures (`save<T extends DeepPartial<Entity>>`, for example). TS won't be able to figure out which overload to use. I would assume this is the same for the `.create()` signatures.

Using a signature like this seems to work:

```diff
-  save<T extends DeepPartial<Entity>>
+  save<T extends Entity | DeepPartial<Entity>>
```

Since the above works, we can instead union the original T as part of the DeepPartial type. This seems to work even better as it has fewer changes to core code, and will be more flexible for future use cases. TS now seems to be able to decide on the right one since the non-deeppartial type is now explicitly available:

```diff
- export type DeepPartial<T> = T extends Array<infer U>
+ export type DeepPartial<T> = T | (T extends Array<infer U>
    ? DeepPartial<U>[]
    : T extends Map<infer K, infer V>
    ? Map<DeepPartial<K>, DeepPartial<V>>
    : T extends Set<infer M>
    ? Set<DeepPartial<M>>
    : T extends object
    ? { [K in keyof T]?: DeepPartial<T[K]> }
-    : T
+    : T)
```

I ran this new DeepPartial through all the tests and they seem to pass as well.

Aside: Also added an await to the service `create` call to pass the test in this PR.

nelsonfleig:
> It seems after the `create()` call, a non-deeppartial is returned to be used for the `save()` call, The `save()` call doesn't like it, since apparently a _generic_ T can't be used as a parameter expecting a _generic_ DeepPartial in the .`save()` signatures (`save<T extends DeepPartial<Entity>>`, for example). TS won't be able to figure out which overload to use. I would assume this is the same for the `.create()` signatures.
> 
> Using a signature like this seems to work:
> 
> ```diff
> -  save<T extends DeepPartial<Entity>>
> +  save<T extends Entity | DeepPartial<Entity>>
> ```
> 
> Since the above works, we can instead union the original T as part of the DeepPartial type. This seems to work even better as it has fewer changes to core code, and will be more flexible for future use cases. TS now seems to be able to decide on the right one since the non-deeppartial type is now explicitly available:
> 
> ```diff
> - export type DeepPartial<T> = T extends Array<infer U>
> + export type DeepPartial<T> = T | (T extends Array<infer U>
>     ? DeepPartial<U>[]
>     : T extends Map<infer K, infer V>
>     ? Map<DeepPartial<K>, DeepPartial<V>>
>     : T extends Set<infer M>
>     ? Set<DeepPartial<M>>
>     : T extends object
>     ? { [K in keyof T]?: DeepPartial<T[K]> }
> -    : T
> +    : T)
> ```
> 
> I ran this new DeepPartial through all the tests and they seem to pass as well.
> 
> Aside: Also added an await to the service `create` call to pass the test in this PR.

Great. Updated DeepPartial with @pbrn46 recommendations. Tests pass!

H4ad:
If merged it will close this issue too https://github.com/typeorm/typeorm/issues/8698

pleerock:
@nelsonfleig cockroachdb is super dumb. I'm going to restart the tests.

pbrn46:
Also, beware the package-lock.json has been updated to lockfile v2 in the first commit. Not sure if that would affect the tests. :)

nelsonfleig:
Unfortunately the same cockroachdb test still fails despite restoring the package-lock.json
```
1) repository > find options > locking
       should allow on a left join:
     Error: You must provide selection conditions in order to find a single row.
```

nelsonfleig:
Any updates on the failing cockroachdb test?

pleerock:
test was failing due to issue on master branch. Let me merge this PR.

PR #8819, opened 2022-03-28
fix: Use concrete subclass for saving with single table inheritance

### LINKED ISSUE #2927: When using base class' custom repository, the discriminator is ignored
**Issue type:**

[x ] question
[x ] bug report
[ ] feature request
[ ] documentation issue

**Database system/driver:**

[ ] `cordova`
[ ] `mongodb`
[ ] `mssql`
[ ] `mysql` / `mariadb`
[ ] `oracle`
[x ] `postgres`
[ ] `sqlite`
[ ] `sqljs`
[ ] `react-native`
[ ] `expo`

**TypeORM version:**

[x ] `latest`
[ ] `@next`

I have a Single Table Inheritance structure; User and its child classes. At some point I am saving an instance of User's child class created with the appropriate constructor;

```ts
  static getUserInstanceFromPayload(payload: IUserPayload) {
    switch(payload.role) {
      case UserRole.ADMIN: return new Admin(); // class Admin extends User
      default: throw new Error(`Invalid role ${role}`);
    }
  }
  /* elsewhere... */
  const payload = req.payload as IUserPayload;
  const repository = em.getCustomRepository(User);
  const user = getUserInstanceFromPayload(payload);

  user.populateFromPayload(payload);

  return repository.save(user);
```

So I receive a payload object and then match the role and choose the appropriate constructor so that it should be the right kind of entity, and then populate it and save.

However, when I run this code, I receive this error:

```
Query failed: INSERT INTO "user"("id", "username", "password", "firstName", "lastName", "createdDate", "lastUpdatedDate", "role") VALUES (DEFAULT, $1, $2, $3, $4, DEFAULT, DEFAULT, $5) RETURNING "id", "createdDate", "lastUpdatedDate", "role" -- PARAMETERS: ["joe.schmoe@example.com","$2b$07$J4XESA9hrVFJPpG.Xjfo0.zv9HOB4cd3T2xLxbEI37N20Dvx11TsG","Joe","Schmoe","User"]
Error: error: invalid input value for enum user_role_enum: "User"
```

So I see that when it saves, it is somehow obtaining "User" as the discriminator value from the repository. 

My solution has been to use the EntityManager to save the entity and it works as expected. So it's not like it's impossible to do what I want.

```ts
return em.save(user); // works as expected
```

Additional info:

```ts
enum UserRole {
  ADMIN = "ADMIN",
  // etc...
}
```
My User entity:
```ts
@Entity()
@TableInheritance({
  column: {
    name: "role",
    type: "enum",
    enum: UserRole,
    default: UserRole.BASIC_USER
  }
})
export abstract class User {
  @Expose()
  protected role: UserRole;
}
```

And the admin:
```ts
@ChildEntity(UserRole.ADMIN)
export class Admin extends User {

  constructor() {
    super();

    this.role = UserRole.ADMIN; // this is only here so that unit tests work
  }
}
```

felix-gohla:
### Description of change

When using a STI scheme and an EntityManager or Repository with base class target, the wrong discriminator was written to the database despite giving a concrete entity.
This was because of the entity's target being ignored in favour of the target of the Repository or the EntityManager, which then is the (abstract) base class.

This PR introduces a fix by **figuring out the concrete subclass of an entity when saving**.

Further questions from my side:

* Currently, it allows for subclass entities to be found by querying the class from `childEntityMetadatas` for every entity. What about performance problems?
* Should this behaviour be deactivatable?
* Is there any documentation that has to be updated?

This PR fixes #2927 (which was closed by the author without the problem being resolved).

### Pull-Request Checklist

### DIFF
--- a/src/persistence/EntityPersistExecutor.ts
+++ b/src/persistence/EntityPersistExecutor.ts
@@ -81,10 +81,28 @@ export class EntityPersistExecutor {
                         if (entityTarget === Object)
                             throw new CannotDetermineEntityError(this.mode)
 
+                        let metadata = this.connection.getMetadata(entityTarget)
+
+                        // Check for single table inheritance and find the correct metadata in that case.
+                        // Goal is to use the correct discriminator as we could have a repository
+                        // for an (abstract) base class and thus the target would not match.
+                        if (
+                            metadata.inheritancePattern === "STI" &&
+                            metadata.childEntityMetadatas.length > 0
+                        ) {
+                            const matchingChildMetadata =
+                                metadata.childEntityMetadatas.find(
+                                    (meta) =>
+                                        entity.constructor === meta.target,
+                                )
+                            if (matchingChildMetadata) {
+                                metadata = matchingChildMetadata
+                            }
+                        }
+
                         subjects.push(
                             new Subject({
-                                metadata:
-                                    this.connection.getMetadata(entityTarget),
+                                metadata,
                                 entity: entity,
                                 canBeInserted: this.mode === "save",
                                 canBeUpdated: this.mode === "save",
--- a/test/github-issues/2927/entity/Content.ts
+++ b/test/github-issues/2927/entity/Content.ts
@@ -1,0 +1,28 @@
+import {
+    Column,
+    Entity,
+    PrimaryGeneratedColumn,
+    TableInheritance,
+} from "../../../../src"
+
+export enum ContentType {
+    Photo = "photo",
+    Post = "post",
+    SpecialPhoto = "special_photo",
+}
+
+@Entity()
+@TableInheritance({
+    pattern: "STI",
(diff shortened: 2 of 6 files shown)


src/persistence/EntityPersistExecutor.ts
```diff
@@ -81,10 +81,28 @@ export class EntityPersistExecutor {
                         if (entityTarget === Object)
                             throw new CannotDetermineEntityError(this.mode)
 
+                        var metadata = this.connection.getMetadata(entityTarget)
```
pleerock, line 84: can you please use `let` instead?

felix-gohla: Sure. 😊

pleerock:
Thank you!

PR #7880, opened 2021-07-10
refactor: move query generation code in `orUpdate` into getQuery

imnotjames:
### Description of change

this also updates the API for `orUpdate` to simplify it

### Pull-Request Checklist

- [x] Code is up-to-date with the `master` branch
- [x] `npm run lint` passes with this change
- [x] `npm run test` passes with this change
- [ ] This pull request links relevant issues as `Fixes #0000`
- [x] There are new or updated unit tests validating the change
- [ ] Documentation has been updated to reflect this change
- [x] The new commits follow conventions explained in [CONTRIBUTING.md](https://github.com/typeorm/typeorm/blob/master/CONTRIBUTING.md)

PR #8829, opened 2022-03-30
Clarify deprecation or orUpdate

Jacse:
It would look like both onConflict and orUpdate were deprecated with no description on what to use or why

### Description of change

With [this change](https://github.com/typeorm/typeorm/pull/7880/files#diff-f12ef5527b9bc2971ab7a8247c71dd735a43cd7a1c859bd206b61dee8c7f617d) `onConflict` was deprecated in favour of `orUpdate`, but at the same time the "old" call signature of `orUpdate` was also deprecated. For people that are using the old call signature of `orUpdate` it now looks like both functions are deprecated with no explanation or alternative offered. This PR introduces a small message to clarify that it is only the one call signature that is deprecated, the function should still be used.

### Pull-Request Checklist

### DIFF
--- a/src/query-builder/InsertQueryBuilder.ts
+++ b/src/query-builder/InsertQueryBuilder.ts
@@ -335,7 +335,7 @@ export class InsertQueryBuilder<Entity> extends QueryBuilder<Entity> {
     }
 
     /**
-     * @deprecated
+     * @deprecated Use new call signature instead.
      */
     orUpdate(statement?: {
         columns?: string[]


pleerock:
if we go with this change, it's better to document what **exactly** to use instead.

Jacse:
Updated so the message now looks like
billede

pleerock:
thank you!

gempain:
@Jacse how do you provide a `WHERE` condition after the `ON CONFLICT DO UPDATE SET ... WHERE` which is allowed in Postgres ?

Issue #8883, opened 2022-04-12
Cannot set property metadata of #<Repository> which has only a getter


feather-jmalone:
## Issue Description

The type file for `Repository` (`src/repository/Repository.ts` line 51) contains a `get metadata()` definition, but no `set()` definition. Depending on your settings and dependency versions, this can result in the following error from the TypeScript compiler:

```
Cannot set property metadata of #<Repository> which has only a getter
```

**NOTE:** I discovered this issue while using TypeORM in a NestJS application, but it is not a NestJS specific issue, really; it's more a typedef / compilation issue.

When the NestJS application loads, NestJS shows the error above, with the following stack trace:

```
TypeError: Cannot set property metadata of #<Repository> which has only a getter
    at EntityManager.getCustomRepository (/Users/githubperson/nestjs-project/node_modules/typeorm/entity-manager/EntityManager.js:756:50)
    at DataSource.getCustomRepository (/Users/githubperson/nestjs-project/node_modules/typeorm/data-source/DataSource.js:320:29)
    at InstanceWrapper.useFactory [as metatype] (/Users/githubperson/nestjs-project/node_modules/@nestjs/typeorm/dist/typeorm.providers.js:13:35)
    at Injector.instantiateClass (/Users/githubperson/nestjs-project/node_modules/@nestjs/core/injector/injector.js:289:55)
    at callback (/Users/githubperson/nestjs-project/node_modules/@nestjs/core/injector/injector.js:42:41)
    at processTicksAndRejections (internal/process/task_queues.js:97:5)
    at async Injector.resolveConstructorParams (/Users/githubperson/nestjs-project/node_modules/@nestjs/core/injector/injector.js:114:24)
    at async Injector.loadInstance (/Users/githubperson/nestjs-project/node_modules/@nestjs/core/injector/injector.js:46:9)
    at async Injector.loadProvider (/Users/githubperson/nestjs-project/node_modules/@nestjs/core/injector/injector.js:68:9)
    at async Promise.all (index 3)
    at async InstanceLoader.createInstancesOfProviders (/Users/githubperson/nestjs-project/node_modules/@nestjs/core/injector/instance-loader.js:43:9)
    at async /Users/githubperson/nestjs-project/node_modules/@nestjs/core/injector/instance-loader.js:28:13
    at async Promise.all (index 55)
    at async InstanceLoader.createInstances (/Users/githubperson/nestjs-project/node_modules/@nestjs/core/injector/instance-loader.js:27:9)
[...]

PR #8884, opened 2022-04-12
fix: add setter method to 'metadata' property in src/repository/Repository.d.ts

feather-jmalone:
Fixes: #8883

### Description of change

Adds a `set()` method to the `metadata` property in the type definition file for the `Repository` class. This fixes the compiler error: `Cannot set property metadata of #<Repository> which has only a getter`

### Pull-Request Checklist

### DIFF
--- a/src/repository/Repository.ts
+++ b/src/repository/Repository.ts
@@ -51,6 +51,7 @@ export class Repository<Entity extends ObjectLiteral> {
     get metadata() {
         return this.manager.connection.getMetadata(this.target)
     }
+    set metadata(newMetadata) {}
 
     // -------------------------------------------------------------------------
     // Constructor


pleerock:
Current solution is wrong because Repository shouldn't have this setter.

It's not a compiler error. Based on your description someone tries to `set` metadata to the repository, thus causes this issue.  

You have two options:

* looks like the problem is around `EntityManager.getCustomRepository` - find where the issue is and fix according to the new Repository architecture. 
* do not use deprecated `EntityManager.getCustomRepository` method and use new custom repositories syntax.

Ginden:
I'm 99% sure it's caused by mismatch between `@nestjs/typeorm` and `typeorm@0.3`.

pleerock:
Closing as no answer from author.

PR #8917, opened 2022-04-20
Fix: #8914

### LINKED ISSUE #8914: CLI does not support async DataSource
## Issue Description
Starting from the 0.3.0 version, CLI requires to specify the file with DataSource instance instead of `ormconfig`. When I am trying to export async created DataSource it fails

### Expected Behavior
Command ran successfully

### Actual Behavior
Command failed with error: `Error: Given data source file must contain export of a DataSource instance`

### Steps to Reproduce
Have your DataSource defined in the file:

```typescript
const buildDataSource = async () => {
  // Provided config-like functionality just for example, it will fail anyway with just async function
  const dbConfig = await config.get('database');

  return new DataSource({
    type: 'mysql',
    host: dbConfig.host,
    port: 3306,
    username: dbConfig.username,
    password: dbConfig.password,
    database: dbConfig.database,
    logging: true,
    entities: typeormEntities,
    migrations: [ './src/components/database/migrations/**/*.ts' ]
  })
};

export default buildDataSource();
```

This works though:
```typescript
const buildDataSource = () => {
  return new DataSource({
    type: 'mysql',
    host: '127.0.0.1',
    port: 3306,
    username: 'test',
    password: 'test',
    database: 'test',
    logging: true,
    entities: typeormEntities,
    migrations: [ './src/components/database/migrations/**/*.ts' ]
  })
};

export default buildDataSource();
```
### My Environment

| Operating System    |     Ubuntu 20.04.4 LTS     |
| Node.js version     | v16.13.0  | 
| Typescript version  | 4.6.3  | 
| TypeORM version     | 0.3.4  | 

I tested with `mysql` only but I do not think it is a driver problem at all

pavlovanton:
### Description of change
Support async import for DataSource in CLI. Fixes #8914 

### Pull-Request Checklist

### DIFF
--- a/src/commands/CommandUtils.ts
+++ b/src/commands/CommandUtils.ts
@@ -35,7 +35,10 @@ export class CommandUtils {
 
         const dataSourceExports = []
         for (const fileExport of dataSourceFileExports) {
-            const awaitedFileExport = await fileExport
+            // It is necessary to await here in case of the exported async value (Promise<DataSource>).
+            // e.g. the DataSource is instantiated with an async factory in the source file
+            const awaitedFileExport =
+                fileExport instanceof Promise ? await fileExport : fileExport
             if (InstanceChecker.isDataSource(awaitedFileExport)) {
                 dataSourceExports.push(awaitedFileExport)
             }
--- a/src/commands/CommandUtils.ts
+++ b/src/commands/CommandUtils.ts
@@ -35,10 +35,8 @@ export class CommandUtils {
 
         const dataSourceExports = []
         for (const fileExport of dataSourceFileExports) {
-            const awaitedFileExport = await fileExport;
-            if (
-                InstanceChecker.isDataSource(awaitedFileExport)
-            ) {
+            const awaitedFileExport = await fileExport
+            if (InstanceChecker.isDataSource(awaitedFileExport)) {
                 dataSourceExports.push(awaitedFileExport)
             }
         }
--- a/src/commands/CommandUtils.ts
+++ b/src/commands/CommandUtils.ts
@@ -34,11 +34,12 @@ export class CommandUtils {
         }
 
         const dataSourceExports = []
-        for (let fileExport in dataSourceFileExports) {
+        for (const fileExport of dataSourceFileExports) {
+            const awaitedFileExport = await fileExport;
             if (
-                InstanceChecker.isDataSource(dataSourceFileExports[fileExport])
+                InstanceChecker.isDataSource(awaitedFileExport)
             ) {
-                dataSourceExports.push(dataSourceFileExports[fileExport])
+                dataSourceExports.push(awaitedFileExport)
             }
         }
 


src/commands/CommandUtils.ts
```diff
@@ -34,11 +34,10 @@ export class CommandUtils {
         }
 
         const dataSourceExports = []
-        for (let fileExport in dataSourceFileExports) {
-            if (
-                InstanceChecker.isDataSource(dataSourceFileExports[fileExport])
-            ) {
-                dataSourceExports.push(dataSourceFileExports[fileExport])
+        for (const fileExport of dataSourceFileExports) {
+            const awaitedFileExport = await fileExport
```
pleerock, line 38: can't we add a check (something like `if (fileExport instanceof Promise)`) in order to make it more clear? also I recommend to add a comment line and explain why this change is needed.

pavlovanton: @pleerock Sure, we can. It was decided to do so for sake of shortness (as far as `await` does not make (almost) any effect on non-`Promise` value this way).

But if it will be more clear this way, we can do like this (and the comment would be like that):
```typescript
// It is necessary to await here in case of the exported async value (Promise<DataSource>).
// e.g. the DataSource is instantiated with an async factory in the source file
const awaitedFileExport = fileExport instanceof Promise 
  ? await fileExport
  : fileExport;
```

What do you think?

Ginden: @pleerock It's possible, but some people still use eg. Bluebird promises or compilation chains with `Promise` polyfills. There isn't any benefit in `instanceof Promise` checks, but only some obscure bug reports.

V1raNi: I think `await`ing doesn't introduce any overhead, as MDN states - Returns the fulfilled value of the promise, or the value itself if it's not a Promise. And even if there is any, it is too small. So I would agree with @Ginden on this one

ViniciussSantos:
I have some questions about how commandUtils.loadDataSource works and how to test this new bugfix. First, I'm trying to create a test that works for the current version first before adapting it to work with the async version:

```javascript
const dataSourceExports = []
for (let fileExport in dataSourceFileExports) {
    if (
        InstanceChecker.isDataSource(dataSourceFileExports[fileExport])
    ) {
        dataSourceExports.push(dataSourceFileExports[fileExport])
    }
}
  if (dataSourceExports.length === 0) {
            throw new Error(
                `Given data source file must contain export of a DataSource instance`,
            )
        }
```
But for some reason the InstanceChecker if statement is always false and the error is always thrown in my test and I can't seem to understand why.


Right now, my test  involves a simple file exporting a dataSource instance like this: 
```javascript 
const options: DataSourceOptions = {
    type: "postgres",
    host: "localhost",
    port: 5432,
    username: "test",
    password: "test",
    database: "test",
    synchronize: true,
    entities: [Post, Author, Category],
}
const dataSource = new DataSource(options)

module.exports = dataSource
```
and I'm testing it like this: 

```javascript 
const DataSourcePath = __dirname + "/datasourceInstance"
        const response = await CommandUtils.loadDataSource(
            DataSourcePath,
        )
        expect(response).instanceOf(DataSource)
```
when I console.log dataSourceFileExports, I get the confirmation that it's actually a datasource instance:
image
but the instanceChecker never returns a true value. What am I doing wrong @Ginden @pleerock ?

pleerock:
@ViniciussSantos not sure, you might need to debug and check the data source symbol values?

jordanmnunez:
any update on pushing this through? Looks like it is really close

texiontech:
when will release this issue.

jonfarzrh (approved):

jonfarzrh:
@pavlovanton Nice work on this man!  It seems that your checklist should be updated to say tests are passing.  I'd be more then happy to write tests around this if it would get it in faster.  TY for all your guys work on TypeORM!

PR #8930, opened 2022-04-24
fix: use last item in value array in the case of duplicate columns (mssql)

### LINKED ISSUE #7775: MSSQL: Duplicate columns names return as an array, breaking mapping
## Issue Description

I have the following query builder (will post entities below as well)

```
getRepository(Post).createQueryBuilder('Post')
        .andWhere('author.id = :p0')
        .setParameters({'p0': 1})
        .select('Post.id, Post.title, Post.text, Post.author'.split(',').map(i => i.trim()))
        .leftJoinAndSelect('Post.author', 'author', '1 = 1')
        .getMany()
```

This produces the following SQL, irrespective of the driver used:
```
SELECT "Post"."id"        AS "Post_id"
,      "Post"."title"     AS "Post_title"
,      "Post"."text"      AS "Post_text"
,      "author"."id"      AS "author_id"
,      "author"."name"    AS "author_name"
,      "Post"."author_id"
FROM      "posts"   "Post"  
LEFT JOIN "authors" "author" ON "author"."id"="Post"."author_id"
		AND (1 = 1)
WHERE "author"."id" = @0
```

When using the MSSQL driver, the results look like this:
```
[
    {
        "id": 1,
        "title": "Ultricies Sem Ltd",
        "text": "quam. Curabitur vel lectus. Cum",
        "author": {
            "id": [
                1,
                1
            ],
            "name": "Ursula Manning"
        }
    },
    {
        "id": 6,
        "title": "Nunc Interdum Feugiat LLC",
        "text": "elit, pellentesque a, facilisis non, bibendum sed, est.",
        "author": {
            "id": [
                1,
                1
            ],
            "name": "Ursula Manning"
        }
    }
]
```

Please note that this seems to be due to the inclusion of `Post.author`. Since we're doing a `leftJoinAndSelect` on `author`, it's going to select all properties. Also selecting `Post.author` seems to confuse the system.

I'm not entirely sure if selecting the relation via  `Post.author` is supposed to be supported. Changing the select to `Post.id, Post.title, Post.text`  (note exclusion of `Post.author`) and the join to `leftJoin` produces a result that doesn't bring back any author data (as expected), but using (`Post.id, Post.title, Post.text, author`) (with inclusion of author) seems to bring back the entire author data (also as expected). So I'm not sure if the syntax `Post.author` is supported.

Regardless of the usage of `Post.author`, I still believe there's some funkiness with it. It's addition to the the select shouldn't cause doubling up of values, and there shouldn't be a difference to the results output that depends on which driver you're using. 

### Expected Behavior

I would not expect the id under author to be turned into a collection. It seems it does this because of how the results are interpreted (there's two `author_id` columns that return, which I assume TypeORM is taking and considering it a collection)

SQLite driver using the same query builder and producing the same query returns the data as expected:

```
[
    {
        "id": 1,
        "title": "Ultricies Sem Ltd",
        "text": "quam. Curabitur vel lectus. Cum",
        "author": {
            "id": 1,
            "name": "Ursula Manning"
        }
    },
    {
        "id": 6,
        "title": "Nunc Interdum Feugiat LLC",
        "text": "elit, pellentesque a, facilisis non, bibendum sed, est.",
        "author": {
            "id": 1,
            "name": "Ursula Manning"
        }
    }
]
```

### Actual Behavior

See above in description

### Steps to Reproduce

1. Run TypeORM via MSSQL driver with provided entities
2. Load data into tables using migration
3. Run querybuilder above

```typescript
postsRepository.createQueryBuilder('Post')
        .andWhere('author.id = :p0').setParameters({'p0': 1})
        .select('Post.id, Post.title, Post.text, Post.author'.split(',').map(i => i.trim()))
        .leftJoinAndSelect('Post.author', 'author', '1 = 1')
        .getMany()
```

### My Environment

| Node.js version     | v12.16.3 | 
| Typescript version  | v3.9.10 | 
| TypeORM version     | v0.2.34 | 

#### Entites
```

@Entity('authors')
export class Author {

  @PrimaryGeneratedColumn()
  id: number;

  @Column()
  name: string;

  @OneToMany(type => Post, post => post.author)
  posts: Post[];

}


@Entity('posts')
export class Post {

  @PrimaryGeneratedColumn()
  id: number;

  @Column()
  title: string;

  @Column()
  text: string;

  // post has relation with category, however inverse relation is not set (category does not have relation with post set)
  @ManyToOne(type => PostCategory, {
    cascade: true
  })
  category: PostCategory;

  // post has relation with details. cascade inserts here means if new PostDetails instance will be set to this
  // relation it will be inserted automatically to the db when you save this Post entity
  @ManyToOne(type => PostDetails, details => details.posts, {
    cascade: ['insert']
  })
  details: PostDetails;

  // post has relation with details. not cascades here. means cannot be persisted, updated or removed
  @ManyToOne(type => Author, author => author.posts)
  author: Author;

}

@Entity('post_category')
export class PostCategory {

  @PrimaryGeneratedColumn()
  id: number;

  @Column()
  name: string;
}


@Entity('post_details')
export class PostDetails {

  @PrimaryGeneratedColumn()
  id: number;

  @Column({
    type: String,
    nullable: true
  })
  authorName: string | null;

  @Column({
    type: String,
    nullable: true
  })
  comment: string | null;

  @Column({
    type: String,
    nullable: true
  })
  metadata: string | null;

  @OneToMany(type => Post, post => post.details, {
    cascade: true
  })
  posts: Post[];

}
```

#### Migration to load data
```

export class DataFilling1577087002356 implements MigrationInterface {

  async up(queryRunner: QueryRunner): Promise<any> {
    await queryRunner.query(`INSERT INTO "authors" (id,name) 
                                    VALUES 
                                      (1,'Ursula Manning'),
                                      (2,'Carly Butler'),
                                      (3,'Maggy Juarez'),
                                      (4,'Colette Murray'),
                                      (5,'Emmanuel Salinas')`);
    await queryRunner.query(`INSERT INTO "post_category" (id,name) 
                                    VALUES 
                                      (1,'Eu Nulla Limited'),
                                      (2,'Montes Nascetur'),
                                      (3,'Sem Incorporated'),
                                      (4,'Sit Amet Consectetuer Limited'),
                                      (5,'Nisl')`);
    await queryRunner.query(`INSERT INTO "posts" (id,title,text,category_id,author_id) 
                                    VALUES 
                                      (1,'Ultricies Sem Ltd','quam. Curabitur vel lectus. Cum',1,1),
                                      (2,'Dolor Consulting','mauris ipsum porta elit, a',2,2),
                                      (3,'Feugiat Associates','tincidunt. Donec vitae erat',3,3),
                                      (4,'Ipsum Phasellus LLC','ut eros non enim commodo hendrerit. Donec porttitor',4,4),
                                      (5,'Habitant Morbi Tristique Corp.','Suspendisse',5,5),
                                      (6,'Nunc Interdum Feugiat LLC','elit, pellentesque a, facilisis non, bibendum sed, est.',1,1),
                                      (7,'Sapien Industries','dictum. Proin eget',2,2),
                                      (8,'In Hendrerit Consectetuer Foundation','vulputate, nisi sem semper erat,',3,3),
                                      (9,'Odio Aliquam Vulputate Foundation','velit eget',4,4),
                                      (10,'Cursus Ltd','ac mattis ornare, lectus ante dictum mi, ac mattis velit',5,5)`);
  }

  async down(queryRunner: QueryRunner): Promise<any> {

  }
}
```

- [ ] `aurora-data-api`
- [ ] `aurora-data-api-pg`
- [ ] `better-sqlite3`
- [ ] `cockroachdb`
- [ ] `cordova`
- [ ] `expo`
- [ ] `mongodb`
- [ ] `mysql`
- [ ] `nativescript`
- [ ] `oracle`
- [ ] `postgres`
- [ ] `react-native`
- [ ] `sap`
- [x] `sqlite`
- [ ] `sqlite-abstract`
- [ ] `sqljs`
- [x] `sqlserver`

blitzmann:
### Description of change

Closes #7775

The `mssql` driver, by default, returns an array when the resultset has duplicate columns. For example:

`SELECT 1 as id, 2 as id` would restult in `{id: [1, 2]}`

See https://github.com/tediousjs/node-mssql/issues/1384 upstream issue for more information and discussion. While I can understand the intent behind it, this is in contrast to how other drivers handle duplicate columns (in which they return, by default, the last column value). It seems that `mssql` is hesitant to introduce an option to automatically select for this behavior, citing data loss.

For TypeORM, tho, we need drivers that return data in similar fashion consistently. This change will look for any arrays that `mssql` returns, and then loop over the resultset and replace that property with the last value.

This is a fairly edge-case scenario - It's not normal for TypeORM to generate a query with duplicate columns, but it can happen (see test case entities for example), and this protects the mapper from returning odd results.

Please also see the issue in which this closes, #7775

### Pull-Request Checklist

       * Note: there is one failing test case, but it seems to be happening in master as well, so not related to this PR. I only ran tests using `mssql` in ormconfig
      * Unsure if this is needed?

### DIFF
--- a/src/driver/sqlserver/SqlServerQueryRunner.ts
+++ b/src/driver/sqlserver/SqlServerQueryRunner.ts
@@ -269,6 +269,22 @@ export class SqlServerQueryRunner
 
             if (raw?.hasOwnProperty("recordset")) {
                 result.records = raw.recordset
+                // mssql driver will return an array if duplicate columns exist
+                // The following will search for properties that return as an array, and then update the property with the last value
+                if (result.records?.length > 0) {
+                    const propsToTakeForDuplicate = []
+                    for (let prop in result.records[0]) {
+                        if (Array.isArray(result.records[0][prop])) {
+                            propsToTakeForDuplicate.push(prop)
+                        }
+                    }
+                    for (let record of result.records) {
+                        for (let prop of propsToTakeForDuplicate) {
+                            const lastIdx = record[prop].length - 1
+                            record[prop] = record[prop][lastIdx]
+                        }
+                    }
+                }
             }
 
             if (raw?.hasOwnProperty("rowsAffected")) {
--- a/test/github-issues/7775/entities/test.ts
+++ b/test/github-issues/7775/entities/test.ts
@@ -1,0 +1,26 @@
+import {
+    Entity,
+    JoinColumn,
+    ManyToOne,
+    OneToMany,
+    PrimaryGeneratedColumn,
+} from "../../../../src"
+
+@Entity("parent")
+export class Parent {
+    @PrimaryGeneratedColumn()
+    id: number
+
+    @OneToMany((type) => Child, (child) => child.parent)
+    children: Child[]
+}
+
+@Entity("child")
+export class Child {
+    @PrimaryGeneratedColumn()
+    id: number
+
+    @ManyToOne((type) => Parent, (parent) => parent.children)
+    @JoinColumn({ name: "parent_id" })
+    parent: Parent
+}
--- a/test/github-issues/7775/issue-7775.ts
+++ b/test/github-issues/7775/issue-7775.ts
@@ -1,0 +1,72 @@
+import { expect } from "chai"
+import "reflect-metadata"
+import { DataSource } from "../../../src"
+import {
+    closeTestingConnections,
+    createTestingConnections,
+    reloadTestingDatabases,
+} from "../../utils/test-utils"
+import { Child, Parent } from "./entities/test"
+
+describe("github issues > #7775 MSSQL: Duplicate columns names return as an array, potentially breaking mapping", () => {
+    let connections: DataSource[]
+    before(async () => {
+        connections = await createTestingConnections({
+            enabledDrivers: ["mssql"],
+            entities: [Parent, Child],
+            dropSchema: true,
+            schemaCreate: true,
+        })
+    })
+    beforeEach(() => reloadTestingDatabases(connections))
+    after(() => closeTestingConnections(connections))
+
+    it("should take the last column as the value, consistent with other drivers", () =>
+        Promise.all(
+            connections.map(async (connection) => {
+                const results = await connection.manager.query(
+                    `SELECT 1 as id, 2 as id UNION SELECT 5, NULL`,
+                )
+                expect(results.length).to.be.eql(2)
+                expect(results[0].id).to.be.eql(2)
+                expect(results[1].id).to.be.eql(null)
+            }),
+        ))
+
+    it("should map the data properly if duplicate columns are present", () =>
+        Promise.all(
+            connections.map(async (connection) => {
+                const parentRepo = await connection.getRepository(Parent)
+                const childRepo = await connection.getRepository(Child)
+
+                const parent = new Parent()
+                parent.id = 1
+
+                await parentRepo.save(parent)
+
+                const child = new Child()
+                child.id = 1
+                child.parent = parent
+
+                await childRepo.save(child)
+
+                const results = await connection
+                    .getRepository(Child)
+                    .createQueryBuilder("child")
+                    .andWhere("parent.id = :p0")
+                    .setParameters({ p0: 1 })
+                    .select(
+                        "child.id, child.parent"
+                            .split(",")
+                            .map((i) => i.trim()),
+                    )
+                    .leftJoinAndSelect("child.parent", "parent", "1 = 1")
+                    .getMany()
+
+                expect(results.length).to.be.eql(1)
+                expect(results[0]).to.be.instanceOf(Child)
+                expect(results[0].parent).to.be.instanceOf(Parent)
+                expect(results[0].parent.id).to.be.eq(1)
+            }),
+        ))
+})


dhensby:
It looks like this fix will leave the streamed results in the old format. Perhaps there needs to be a transform stream attached to the readable that normalises the streamed rows too?

pleerock:
We do not change driver output for the `query` method.

pleerock:
I remember I had to introduce some changes (not remember where, maybe just in test cases) after mssql version upgrade. But we do not change the output of the underlying driver. If driver returns it this way, there maybe a reason? I can say that I have problems with design decisions on absolutely every driver TypeORM currently supports. Please don't give me another problem.

dhensby:
I'm not too involved in TypeORM so, my opinion shouldn't hold too much weight, but I'll share it anyway.

@blitzmann raises some good points in the linked issue https://github.com/tediousjs/node-mssql/issues/1384. The mssql driver is returning duplicate column names in a way that is not typical of any other database driver at the moment (at least according to him). When there is a duplicate column name, the MSSQL driver is returning all the values as an array rather than just returning the last value.

Whilst I have my views (as the maintainer of the MSSQL lib) about that being "best", my view of an ORM/framework like TypeORM is to homogenise the DB drivers so that swapping database drivers is completely™️ transparent. ie: the idiosyncrasies are dealt with at the ORM level, rather than by the consumer of the ORM. That would mean if some odd-ball DB driver comes along and returns data in a completely unrecognised way (like mssql *could* return rows as arrays instead of objects) then it's for the ORM to coerce this data into the form that the ORM and the ORM consumer expects.

In short, the contract is between the ORM and the ORM consumer, not between the ORM consumer and the DB driver. This change seems pretty sensible in the context of how all the other driver behave.

blitzmann:
@pleerock I can understand wanting to keep direct called to `query()` be completely representative of what the underlying driver returns. Using `query()` directly signals the need to bypass the ORM altogether, and I can appreciate not wanting to change the output in that situation.

That being said, I fully agree with @dhensby as to the purpose of TypeORM, and couldn't say it better myself, so I'll just reiterate

> the idiosyncrasies are dealt with at the ORM level, rather than by the consumer of the ORM. [...] it's for the ORM to coerce this data into the form that the ORM and the ORM consumer expects.

@pleerock is there another area besides `SqlServerQueryRunner` you would rather see a change like this so that it doesn't affect `query()`, but rather incorporates it into the ORM mapping more directly? Personally I don't see the issue with having this here as having a column come back as an array is non-sensical to the rest of the ORM. TypeORM already coerce's the data into `QueryResult`, even if it doesn't touch the data itself and just manipulates the object to get the records to go where TypeORM expects it to. But, in this case, there's no reason to have an array as a value, as (as documented in the issue) TypeORM doesn't know how to handle it.

Alternatively, this can be introduced as an option for TypeORM to toggle this bit of code on and off, but I don't see the point in that since I can't imagine a scenario where someone would want TypeORM to return a mapped object like below. 

```
{
    id: 1234,
    someRelation: {
        id: [1, 2],
        name: "Some Name"
    }
}
```

Please advise :)

blitzmann:
@pleerock just seeing if there's been any more thought given to this. I'm still of the opinion that this should be at the ORM level to massage the data into a format that TypeORM is expecting.

src/driver/sqlserver/SqlServerQueryRunner.ts
```diff
@@ -269,6 +269,22 @@ export class SqlServerQueryRunner
 
             if (raw?.hasOwnProperty("recordset")) {
                 result.records = raw.recordset
+                // mssql driver will return an array if duplicate columns exist
+                // The following will search for properties that return as an array, and then update the property with the last value
```
pleerock, line 273: why exactly the _last value_? Why not the first, second or n-th?

Ginden: @pleerock I think it's for consistency with other database drivers.

pleerock: thanks @Ginden . What's your opinion on this PR?

Ginden: I have mixed feelings. Consistency is good, but underlying issue of duplicated column names looks iffy.

pleerock:
I feel like I have to close this PR. Because:

* we have mixed feelings on these changes
* no community interest in this issue other few people

Closing it for a better times.

Issue #8892, opened 2022-04-14
ManyToMany relations save throws "Violation of PRIMARY KEY constraint" when a property from combined primary key from relation entity is an other relation property


christian-forgacs:
## Issue Description

While saving an entity with ManyToMany relation the first save works great. But if I change the ManyToMany relation children, the exception `Violation of PRIMARY KEY constraint 'PK_c83e17466322282074e0360a96e'. Cannot insert duplicate key in object 'dbo.city_zips_zip'. The duplicate key value is (de, 60311, 41a3433e-f36b-1410-84db-00668d135ea6)` throws.

For testing I remove the constraint in the relation table...

In the query log I can see, that you first insert all relations and after inserting you remove all relations. I guess the order is wrong.

```mssql
INSERT INTO "city_zips_zip"("zipCountryCode", "zipCode", "cityId") VALUES (@0, @1, @2), (@3, @4, @5) -- PARAMETERS: [{"value":"de","type":"nvarchar","params":["2"]},{"value":"60311","type":"nvarchar","params":[]},{"value":"41A3433E-F36B-1410-84DB-00668D135EA6","type":"uniqueidentifier","params":[]},{"value":"de","type":"nvarchar","params":["2"]},{"value":"35781","type":"nvarchar","params":[]},{"value":"41A3433E-F36B-1410-84DB-00668D135EA6","type":"uniqueidentifier","params":[]}]

DELETE FROM "city_zips_zip" WHERE (("cityId" = @0 AND "zipCountryCode" = @1 AND "zipCode" = @2) OR ("cityId" = @3 AND "zipCountryCode" = @4 AND "zipCode" = @5)) -- PARAMETERS: ["41A3433E-F36B-1410-84DB-00668D135EA6","de","35781","41A3433E-F36B-1410-84DB-00668D135EA6","de","60311"]
```

### Expected Behavior

Excepted on sync for ManyToMany I want to only remove relations they aren't in the new array and only insert new relations.

### Actual Behavior

At the moment the save method throws an exception, that an insert in relation table can't execute because a violation on primary key constraint because you insert before delete all relations.

### Steps to Reproduce

Create 2 entities:

```typescript

@Entity()
export class City {
  @PrimaryGeneratedColumn('uuid')
  id: string;

  @Column()
  caption: string;

  @RelationId('zips')
  zipCodes: string[];

  @ManyToMany(() => Zip, zip => zip.cities, {
    cascade: true,
  })
  @JoinTable()
[...]

Issue #8045, opened 2021-08-06
Insert fails if PrimaryGeneratedColumn has value


awn70:
## Issue Description
When you try to insert an entity with `repo.insert( entity )` and the entity's `PrimaryGeneratedColumn` column has a value the operation fails, because typeorm tries to insert the value.
(The column has a value because the entity is a clone of an existing one.)

### Expected Behavior

Typeorm ignores the value and returns the new generated value. (typeorm 0.2.34)

### Actual Behavior

Insert fails if the PK value already exists.

```
Violation of PRIMARY KEY constraint 'TTT_PK'. Cannot insert duplicate key in object ...
```

### My Environment

| Dependency          | Version  |
| ---                 | ---      |
| Operating System    |          |
| Node.js version     | 12.20.2  |
| Typescript version  | 4.3.5  | 
| TypeORM version     | 0.2.36  | 

### Relevant Database Driver(s)

| DB Type              | Reproducible  |
| ---                  | ---           |
| `aurora-data-api`    | no            |
| `aurora-data-api-pg` | no            |
| `better-sqlite3`     | no            |
| `cockroachdb`        | no            |
| `cordova`            | no            |
| `expo`               | no            |
| `mongodb`            | no            |
| `mysql`              | no            |
| `nativescript`       | no            |
| `oracle`             | no            |
| `postgres`           | no            |
| `react-native`       | no            |
| `sap`                | no            |
| `sqlite`             | no            |
| `sqlite-abstract`    | no            |
| `sqljs`              | no            |
| `sqlserver`          | yes            |

### Are you willing to resolve this issue by submitting a Pull Request?

 - ✖️ Yes, I have the time, and I know how to start.
 - ✖️ Yes, I have the time, but I don't know how to start. I would need guidance.
 - ✖️ No, I don’t have the time, but I can support (using donations) development.
 - ✅ No, I don’t have the time and I’m okay to wait for the community / maintainers to resolve this issue.

drjamesj:
Please post the relevant sample code to reproduce the issue.

imnotjames:
Isn't this the expected result because you're telling TypeORM to do something & it is?

imnotjames:
I do believe that this was done to fix another issue - so if you're telling TypeORM to insert a value then it's going to attempt to do so.  The fact that it was ignoring this before was a bug.

pleerock:
[...]

PR #8959, opened 2022-05-02
fix: wrong entity transform of relation id when referenced column is a primary key

christian-forgacs:
Fixes: #8892
May be fix: #8045

Fix wrong entity transform of relation id when referenced column is also a primary key.

### Description of change

Only create value map if column isn't a primary key.

### Pull-Request Checklist

- [x] ~~Documentation has been updated to reflect this change~~
  - No changes needed because bugfix


christian-forgacs:
@pleerock any estimate when somebody from your team can review this fix? We had a big project and we add more and more workarounds in our project to fix this issue. I don't want to stress. It's only a question :)

pleerock:
I had to simplify your test to be able to understand the problem better. So yes, I can say the solution fixes the problem, however it touches very fragile code in the codebase, which obviously needs to be better than what we currently have.

I'll merge this PR. Thanks for the contribution.

PR #8971, opened 2022-05-05
feat: allow for partial index conditions for on conflict statments in postgres

deekthesqueak:
### Description of change
Allows WHERE conditions to be set for ON CONFLICT statements in Postgres when using a partial index.

Technical Background:

Tables can be created without a constraint but with and an unique [partial index ](https://www.postgresql.org/docs/current/indexes-partial.html)can then be added after the fact.

```
create table test (
  a int not null,
  b text not null,
  c boolean,
  primary key(a)
);

create unique index idx_test on test (b) where c = true;

insert into test (a,b,c) values (1,'test',true);

-- fails, not a table constraint
insert into test (a,b,c) values (2,'test',true) on conflict on constraint idx_test do nothing;

-- fails. condition not matching
insert into test (a,b,c) values (2,'test',true)
  on conflict (b)
    do update set a = EXCLUDED.a where test.a <> EXCLUDED.a;

-- succeeds with full index condition
insert into test (a,b,c) values (2,'test',true)
  on conflict (b) where c = true
    do update set a = EXCLUDED.a where test.a <> EXCLUDED.a;
```

When this happens the current orUpdate function can not properly match the constraint either by array of values, or named indices. Adding the additional WHERE clause on the constraint fixes the issue.

### Pull-Request Checklist

- [N/A] This pull request links relevant issues as `Fixes #0000` 


src/query-builder/InsertQueryBuilder.ts
```diff
@@ -356,6 +356,7 @@ export class InsertQueryBuilder<Entity> extends QueryBuilder<Entity> {
         overwrite: string[],
         conflictTarget?: string | string[],
         orUpdateOptions?: InsertOrUpdateOptions,
+        conflictConditions?: string[],
```
pleerock, line 359: too many parameters already... at some point we need to add an overload to accept both parameters and object with parameters.

src/query-builder/InsertQueryBuilder.ts
```diff
@@ -371,6 +372,7 @@ export class InsertQueryBuilder<Entity> extends QueryBuilder<Entity> {
             | string[],
         conflictTarget?: string | string[],
         orUpdateOptions?: InsertOrUpdateOptions,
+        conflictConditions?: string[],
```
pleerock, line 375: do we have conditions other than "conflict"? If no, better to call this parameter `where`, and I think it can be simply `where?: string`, no need to use arrays here.

pleerock:
Thank you for contribution and sorry for the delayed review, I think PR can be merged after fixing the addressed comments.

src/query-builder/InsertQueryBuilder.ts
deekthesqueak: I agree as well. The reason I added a new optional function parameter was make the smallest change that wouldn't break the preferred calling method (as listed in the deprecation message)
```
     * `.orUpdate({ columns: [ "is_updated" ] }).setParameter("is_updated", value)`
     *
     * is now `.orUpdate(["is_updated"])`
     *
     * `.orUpdate({ conflict_target: ['date'], overwrite: ['title'] })`
     *
     * is now `.orUpdate(['title'], ['date'])`
```
If there is a preferred overloading setup elsewhere in the repo I can adapt it to suit this.

src/query-builder/InsertQueryBuilder.ts
deekthesqueak: I did not see any other conditions so I made the type change suggested. Instead of using `where` I used the exact term used in the Postgres documentation [1] (under ON CONFLICT Clause) to provide contextual reference for its usage.

[1] https://www.postgresql.org/docs/current/sql-insert.html

deekthesqueak:
Thank you for the review. Pardon the delayed fixes. I've addressed the above issues in the latest commit and would welcome any further suggestions around the function signature.

pleerock:
Thank you for contribution!

Looks good, one thing I can suggest that improve user experience: 

* throw an error in the case if it's not postgres family (in the case if someone try to use this feature in non-postgres env). plus add tests
* add the documentation regarding to this feature (if you update .md files in the docs - changes will be reflected on the website)

Thanks again.

deekthesqueak:
I've spent some more time understanding the repository and the code as a whole. Hopefully this can wrap everything up.

In addition to the errors and documentation you've requested I've also tried to pass around the options via `orUpdateOptions` similar to the already existing `skipUpdateIfNoValuesChanged`. This allows the use of the `indexPredicate` in upsert function since it uses the same query builder behind the scenes.

Please let me know if you suggest any further changes.

DamienWalle:
thank you to have worked on that! it is exactly what I need now on my project. Waiting for that PR to be merged, good job ;)

DamienWalle:
Hi @pleerock, @deekthesqueak, 

Do you think that the PR will be merged soon?

Thanks ;)

deekthesqueak:
@DamienWalle I don't have any power to merge this issue, it would require someone on the TypeORM team like @pleerock to do so.

AlexMesser:
thank you for contribution!

dwaam:
Thanks you for your work! 
I've tried to test it, I saw that the release 0.3.11 is built with this PR, but when I update to this version, I don't see the modifications in some files, like in libs/.../InsertQueryBuilder.d.ts. 
I don't know if there is something wrong on my side or during the build or something missing. 

image

DamienWalle:
Oops my bad, forget what I've said ^^

It works well, it was just a bit weird to have to do the `.disableEscaping ` because it caused me some issues. 

I work on a legacy project, and we need to do all the migration gently, and I had some columns from Sequelize `createdAt` and `updatedAt` and with the `disableEscaping`, it says that `createdat` column doesn't exists ^^. 

We well manage, but is there a way to improve that? 
I'll check all the PR to see if I can think of something.

deekthesqueak:
@AlexMesser Thank you for the commit to update the latest linting and the merge of the code

@DamienWalle I'm assuming you are referencing this section
```
                        conflictTarget += ` WHERE ( ${this.escape(
                            indexPredicate,
                        )} )`
```

The reason I used this.escape is because similar code in the surrounding logic was escaping the input(s) in a similar fashion. Do you have any thoughts or suggestions on how you would like to see it improved. If you open up a new issue with some thoughts I'd be happy to look into them.

dwaam:
Hey @deekthesqueak ,

It is more on the createQueryBuilder itselt, If i do that:
```
 await this.createQueryBuilder()
      .insert()
      .orUpdate(
        ['value', 'type', 'application_date', 'updatedAt'],
        ['id_activity_logistic', 'type', 'application_date'],
        {
          skipUpdateIfNoValuesChanged: true,
          indexPredicate: 'code_site IS NULL',
        },
      )
      .into(LeadTime)
      .values(leadtimes)
      .returning(['id'])
      // .disableEscaping()
      .execute();
```

The query generated gives  `WHERE ( "code_site IS NULL" )` so I get the error: `column "code_site IS NULL" does not exist`.
If I add `.disableEscaping()`, it affects my column name turning `updatedAt` to `updatedat`.

I will try to take time to download the projet and test if I figure out something and I'll keep you posted ;)

PR #8984, opened 2022-05-09
fix: infinite loop on SAVEPOINT + missing returned id on creation

mlecoq:
### Description of change

Typeorm  3 is currently not compatible with expo due to some issues : 
- an infinite loop on query method calls (for SAVEPOINT)
- use of SAVEPOINT freezes app
- on insert no id is returned causing issues on join tables

I have tried to update expo-example to use latest api and I needed to make some modifications in TypeORM to make it works (see https://github.com/typeorm/expo-example/pull/5) 

### Pull-Request Checklist

- [ ] `npm run test` passes with this change (validated through expo-example)


DLawla:
@mlecoq Thanks for your work here. We are running into the same issue, and are parked on an old version of TypeORM until this is fixed.

mlecoq:
@DLawla you can try the patch added in expo-example. Let me know if everything work as expected

DLawla:
@mlecoq disregard my previous comment -- I am now encountering the `query:  SAVEPOINT typeorm_1` loop with the example project (calling `expo start -c` cleared a cache that must have been using an older version of typeorm).
I'll test with your patch shortly.

DLawla:
@mlecoq no luck patching the example. I ran into the following issue when running 

`yarn add 'typeorm@https://github.com/mlecoq/expo-typeorm.git#fix/expoQueryRunner'`

from the `expo-example` directory to add your patch. I couldn't find a quick way to work around this, but sounds like you patch's package.json should be updated for 0.3.6. Happy to try again once you update that.

Also, if there is a different way you can think of for me to test, please let me know!

```
**ERROR** Failed to apply patch for package typeorm at path
  
    node_modules/typeorm

  This error was caused because typeorm has changed since you
  made the patch file for it. This introduced conflicts with your patch,
  just like a merge conflict in Git when separate incompatible changes are
  made to the same piece of code.

  Maybe this means your patch file is no longer necessary, in which case
  hooray! Just delete it!

  Otherwise, you need to generate a new patch file.

  To generate a new one, just repeat the steps you made to generate the first
  one.

  i.e. manually make the appropriate file changes, then run 

    patch-package typeorm

  Info:
    Patch file: patches/typeorm+0.3.6.patch
    Patch was made for version: 0.3.6
    Installed version: 0.3.5
---

**Comment by mlecoq** [comment]
@DLawla 

```
Info:
    Patch file: patches/typeorm+0.3.6.patch
    Patch was made for version: 0.3.6
    Installed version: 0.3.5
```

Patch is made for 0.3.6 and currently you're on `0.3.5`, upgrade your typeorm dependency to `0.3.6` and then the patch could be applied successfully

pleerock:
Thank you for contribution! 🎉

DLawla:
Thanks @mlecoq!

pleerock:
@DLawla expect a new release by the end of this week.

DLawla:
Thanks @pleerock

PR #8993, opened 2022-05-10
feat: add commentCallback

stefansundin:
### Description of change

This adds a new configuration option, `commentCallback`, which allows an application to dynamically add a comment to SQL commands. This allows for advanced debugging, e.g. automatically inserting the source of all SQL commands (filename and line number).

Sidenote: In Ruby there's a gem called [marginalia](https://github.com/basecamp/marginalia) which accomplishes the same thing for Ruby/Rails applications.

**Note that this PR is not completely ready.** It is currently in a proof-of-concept stage, and it currently only works for code that uses `createQueryBuilder`. I'm not sure the best way to improve this to make sure the callback is called for everything (e.g. `QueryRunner`). Perhaps someone here would have a good idea of how to accomplish that?

If there's sufficient interest to mainline this and we manage to improve this so it works for all queries, then I can work on adding support for more engines (mysql, etc).

Here's an example of how I have used this in my testing:

```typescript
    commentCallback: () => {
      const cwd = process.cwd();
      const stackRegex = /at\s+(.*)\s+\((.*)\)/i;
      const stack = new Error().stack?.split('\n')?.slice(2);
      if (!stack) {
        return undefined;
      }
      for (const line of stack) {
        const parsed = stackRegex.exec(line);
        if (!parsed) {
          continue;
        }
        const functionName = parsed[1];
        const functionLocationLong = parsed[2];
        if (!functionLocationLong.startsWith(cwd) || functionLocationLong.startsWith(`${cwd}/node_modules/`)) {
          continue;
        }
        const functionLocation = functionLocationLong.substring(cwd.length + 1);
        return `${functionName} (${functionLocation})`;
      }
      return undefined;
    },
```

The end result is that I get these comments like this in my SQL server logs:

```
LOG:  statement: /* register (src/core/auth/auth.resolver.ts:263:23) */ SELECT "User"."id" AS "User_id", "User"."firstName" AS "User_firstName", "User"."lastName" AS "User_lastName", "User"."age" AS "User_age" FROM "user" "User"
```

So if I ever want to see where a query is coming from, the answer will be right there in my logs. 🎉 

I think this adds pretty good value to typeorm and the ability to debug large codebases. I'd love to know if this functionality is something that would be considered for inclusion and I'd love to hear if there's a better way to implement it. Thank you!

### Pull-Request Checklist

### DIFF
--- a/src/data-source/DataSource.ts
+++ b/src/data-source/DataSource.ts
@@ -536,20 +536,28 @@ export class DataSource {
         if (InstanceChecker.isMongoEntityManager(this.manager))
             throw new TypeORMError(`Query Builder is not supported by MongoDB.`)
 
+        let qb: SelectQueryBuilder<Entity>
         if (alias) {
             alias = DriverUtils.buildAlias(this.driver, alias)
             const metadata = this.getMetadata(
                 entityOrRunner as EntityTarget<Entity>,
             )
-            return new SelectQueryBuilder(this, queryRunner)
+            qb = new SelectQueryBuilder(this, queryRunner)
                 .select(alias)
                 .from(metadata.target, alias)
         } else {
-            return new SelectQueryBuilder(
+            qb = new SelectQueryBuilder(
                 this,
                 entityOrRunner as QueryRunner | undefined,
             )
         }
+        if (this.options.type === "postgres" && this.options.commentCallback) {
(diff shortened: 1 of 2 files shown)


pleerock:
Closing for now.

Issue #9004, opened 2022-05-15
Glob Pattern Entities Not Resolved


Adrastopoulos:
## Issue Description

The Directory Exported Classes Loader  (`DirectoryExportedClassesLoader.js`) normalizes the path provided to the `entities` property on a `DataSourceOptions` object. This normalized path changes all **forward** slashes to **back** slashes - however, glob is only compatible with **forward** slashes. Removing the `pathNormalize` function fixed my issue.

### Expected Behavior

```
All classes found using provided glob pattern...
```

### Actual Behavior

```
 No classes were found using the provided glob pattern...
```

### Steps to Reproduce

1. Replicate Code Below

```typescript
const databaseOptions: DataSourceOptions = {
	type: 'postgres',
	host: 'localhost',
	port: 5432,
	username: 'postgres',
	password: DB_PASSWORD,
	entities: [path.join(__dirname, '/entities/*.js').replace(/\\/g, '/')],
	database: 'postgres',
	logging: ['info'],
};

const dataSource = await new DataSource(databaseOptions).initialize();

console.log(dataSource.entityMetadatas); // []
```

### My Environment

| Dependency          | Version  |
| ---                 | ---      |
| Operating System    |     Windows     |
| Node.js version     | 8.3.0  | 
| Typescript version  | 4.6.4  | 
| TypeORM version     | 0.3.6  | 

### Relevant Database Driver(s)

| DB Type           | Reproducible  |
|-------------------| ---           |
| `aurora-mysql`    | no            |
| `aurora-postgres` | no            |
| `better-sqlite3`  | no            |
| `cockroachdb`     | no            |
| `cordova`         | no            |
| `expo`            | no            |
| `mongodb`         | no            |
| `mysql`           | no            |
| `nativescript`    | no            |
| `oracle`          | no            |
| `postgres`        | yes            |
| `react-native`    | no            |
| `sap`             | no            |
| `spanner`         | no            |
| `sqlite`          | no            |
| `sqlite-abstract` | no            |
| `sqljs`           | no            |
| `sqlserver`       | no            |

### Are you willing to resolve this issue by submitting a Pull Request?

 - ✅ Yes, I have the time, and I know how to start.
 - ✖️ Yes, I have the time, but I don't know how to start. I would need guidance.
 - ✖️ No, I don’t have the time, but I can support (using donations) development.
 - ✖️ No, I don’t have the time and I’m okay to wait for the community / maintainers to resolve this issue.

[...]

PR #9007, opened 2022-05-15
Fix glob problem

avin:
There is a bug in latest 7.2.* version of glob package
https://github.com/isaacs/node-glob/issues/468#issuecomment-1126735114

fix for https://github.com/typeorm/typeorm/issues/9004


nicolas377:
Please do not pin to 7.2.0. Glob version 7.2.3 has been released with the fix to the issue you're referring to. This requires no action.

PR #9009, opened 2022-05-16
fix: add SQLite boolean escaping

### LINKED ISSUE #1981: [Bug] Boolean values not casted properly when used in .find() condition

lajtomekadimon:
Fixes #1981 (again).

### Description of change

When using `someRepository.findOneBy({ someField: true })`, `nativeParameters` has no keys and the query doesn't work properly, returning empty array. We updated to last TypeORM version (from 0.2.30) and this seems to solve it, but I don't really understand why booleans worked and now don't. If it means anything, we are using Cordova to develop and Android app. Anyways, this commit fixes it.

### Pull-Request Checklist

### DIFF
--- a/src/driver/sqlite-abstract/AbstractSqliteDriver.ts
+++ b/src/driver/sqlite-abstract/AbstractSqliteDriver.ts
@@ -476,6 +476,10 @@ export abstract class AbstractSqliteDriver implements Driver {
                 } else if (typeof value === "number") {
                     return String(value)
                 }
+                
+                if (typeof value === "boolean") {
+                    return value ? "1" : "0";
+                }
 
                 if (value instanceof Date) {
                     escapedParameters.push(


igorbga:
@[lajtomekadimon](https://github.com/lajtomekadimon) I'm not sure if your description is clear enough.
I'll try to explain it better.
We were using version 0.2.30 of TypeORM with Cordova SQLite plugin and with this version the conditional clauses in the find methods for boolean properties were working fine.
But in the actual TypeOrm version they don't.
This seems to be a regression because something very similar was solved due to issue #1981 .

During the old issue's fix some boolean treatment was done based on a variable called "nativeParameters" but in our debuggin session it seems that in the actual version this variable is just empty. In fact it seems to be deprecated. So. might it be that during the refactoring related to that deprecation the regression was caused?

pleerock:
Few issues with this PR needs to be solved before it can be merged:

* lint is failed
* if it's Cordova-only issue - it needs to be fixed in a CordovaDriver

lajtomekadimon:
This commit has nothing to do with "Required OrableDB Binaries", which is the only failed test from the lint.

I don't think this is a Cordova-only issue. You parse `number` types to strings in sqlite-abstract/AbstractSqliteDriver.ts, but there's no parsing of `boolean` to `"0"` or `"1"`; this commit adds that parsing. Remember that SQLite doesn't have a boolean type (they use 0 as false and 1 as true instead): https://www.sqlite.org/datatype3.html

pleerock:
Re-run the tests, looks like there is lint issues:

```shell
Checking formatting...
[warn] src/driver/sqlite-abstract/AbstractSqliteDriver.ts
[warn] Code style issues found in the above file(s). Forgot to run Prettier?
```

PR #8956, opened 2022-05-01
perf: don't recompile escapeRegExp for every query

draaglom:
### Description of change

Context: the query builder is pretty CPU intensive, and can be slow -
e.g. https://github.com/typeorm/typeorm/issues/3857

One of the things which makes this slow is `escapeRegExp` in the query
builder: we freshly construct the same RegExp once per
`replacePropertyName` invocation (many times per overall query!) and
since the RegExp itself is constant -- we can lift it out and construct
it once.

Over-all this saves about 5% on our query build times as measured by
 #8955.

### Pull-Request Checklist

- [x] Code is up-to-date with the `master` branch
- [x] `npm run format` to apply prettier formatting
- [x] `npm run test` passes with this change
- [x] This pull request links relevant issues as `Fixes #0000`
- [x] There are new or updated unit tests validating the change
- [x] Documentation has been updated to reflect this change
- [x] The new commits follow conventions explained in [CONTRIBUTING.md](https://github.com/typeorm/typeorm/blob/master/CONTRIBUTING.md)

PR #8955, opened 2022-05-01
test: add benchmark for select querybuilder

draaglom:
### Description of change

Motivation: the query builder (and within it, replacePropertyNames and
associated functions) is pretty CPU intensive. For our workload, it's
one of the hottest functions in our entire stack.

While improved in https://github.com/typeorm/typeorm/pull/4760,
There are still outstanding issues relating to perf e.g. https://github.com/typeorm/typeorm/issues/3857

As we all know though, the first step in optimization is to measure
systematically ;)

https://wiki.c2.com/?ProfileBeforeOptimizing

On my machine, this benchmark runs in ~3550ms or about 0.35ms/query.
This tells us there's a way to go - on my stack, that's about 1/3 of a
typical query's latency!

### Pull-Request Checklist

- [x] Code is up-to-date with the `master` branch
- [x] `npm run format` to apply prettier formatting
- [x] `npm run test` passes with this change
- [x] This pull request links relevant issues as `Fixes #0000`
- [x] There are new or updated unit tests validating the change
- [x] Documentation has been updated to reflect this change
- [x] The new commits follow conventions explained in [CONTRIBUTING.md](https://github.com/typeorm/typeorm/blob/master/CONTRIBUTING.md)

PR #9032, opened 2022-05-23
perf: partially lift matching from regexp to js

draaglom:
### Description of change

Digging further into #3857.

See also #8955, #8956.

As [previously discussed](https://github.com/typeorm/typeorm/issues/3857#issuecomment-699505893), the query builder currently suffers from poor performance in two ways:

1. quadratic numbers of operations with respect to total table/column
counts
2.  poor constant factor performance (regexps can be expensive
to build/run!)

The constant-factor performance is the more tractable problem: no longer
quadratically looping would be a chunky rewrite of the query builder,
but we can locally refactor to be a bunch cheaper in terms of regexp
operations.

This change cuts the benchmark time here in ~half (yay!).

We achieve this by simplifying the overall replacement regexp (we don't
need our column names in there, since we already have a plain object
where they're the keys to match against) so compilation of that is much
cheaper, plus skipping the need to `escapeRegExp` every column as a
result.

### Pull-Request Checklist

### DIFF
--- a/src/query-builder/QueryBuilder.ts
+++ b/src/query-builder/QueryBuilder.ts
@@ -756,27 +756,27 @@ export abstract class QueryBuilder<Entity> {
                 replacements[column.propertyPath] = column.databaseName
             }
 
-            const replacementKeys = Object.keys(replacements)
-
-            if (replacementKeys.length) {
-                statement = statement.replace(
-                    new RegExp(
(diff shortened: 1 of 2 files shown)


pleerock:
Looks awesome! Thank you very much!

draaglom:
happy to contribute @pleerock :)

PR #9050, opened 2022-05-29
fix: fixes recursive document to entity attributes mapping

### LINKED ISSUE #9049: Mongodb entities with 2 level nested arrays throws an error for any find operation
# Issue description

Mongodb entities with 2 level nested arrays throws an `document[embedded.prefix].map is not a function` error for any find operation. This error apparently occurs only when the 3rd level entity is named `values` .

## expected behavior

This kind of embed document in mongodb should work without problems.

## actual error

The lib throws `document[embedded.prefix].map is not a function`

image

## precondition

I'm using typeorm with mongodb driver to reproduce this structure, where I have a Post, which embeds an array of Comments, which has an array of Values. 

```ts
// src/entities/post.entity.ts
import { ObjectID } from "../../../../src/driver/mongodb/typings"
import { Type } from 'class-transformer'
import { Comment } from './comment'
import { Column, Entity, ObjectIdColumn } from "../../../../src"

@Entity()
export class Post {
  @ObjectIdColumn()
  _id?: ObjectID

  @Type(() => Comment)
  @Column(() => Comment)
  comments: Comment[]
}
```

```ts
// src/entities/commentst.entity.ts
import { Type } from 'class-transformer'
import { Value } from './value'
import { Column } from '../../../../src'

export class Comment {
  @Type(() => Value)
  @Column(() => Value, { array: true })
  values: Value[]
}
```

```ts
// src/entities/value.entity.ts
import { Column, ObjectID, ObjectIdColumn } from "../../../../src"

export class Value {
  @ObjectIdColumn()
  _id?: ObjectID

  @Column({ type: 'string' })
  description: string
}

```

This structure works well when inserting data to post collection. When I'm getting this objects using `find` or `findOne`, it throws an error.

```ts
/* ... */
const post = new Post();
const comment = new Comment();
const value = new Value();

value.description = 'description';
comment.values = [value];
post.comments = [comment];

// this save works well
const response = await connection.mongoManager.save(post);
const postRepo = await connection.getRepository(Post);

// document[embedded.prefix].map is not a function
const posts = await postRepo.find({});
/* ... */
```

    ✅ Yes, I have the time, and I know how to start.
    ✖️ Yes, I have the time, but I don't know how to start. I would need guidance.
    ✖️ No, I don’t have the time, but I can support (using donations) development.
    ✖️ No, I don’t have the time and I’m okay to wait for the community / maintainers to resolve this issue.

pauloFernandes:
fix the recursive call to `addEmbeddedValuesRecursively`, which was running for
array and object embedded properties, which should only run for non-array ones.
The call for array properties is already placed right

Closes: #9049

### Description of change

Fixes #9049.

According to what is described in the issue, when using typeorm with mongodb driver, if you have an entity which has an array attribute, and this array has another array attribute, typeorm will send an `document[embedded.prefix].map is not a function` error. 

In the current version, the [addEmbeddedValuesRecursively](https://github.com/typeorm/typeorm/blob/master/src/query-builder/transformer/DocumentToEntityTransformer.ts#L147) method calls itself recursively, expecting to have an object as the second argument, even if the object is an array. In the case of arrays, this method is properly called in line [115](https://github.com/typeorm/typeorm/blob/master/src/query-builder/transformer/DocumentToEntityTransformer.ts#L115). The call at the end of the method was producing that error.

To fix this problem, I moved this method to be called from inside the else block, where it's behavior is correct.

### Pull-Request Checklist

- [x] This pull request links relevant issues as `Fixes #9049`

### DIFF
--- a/test/github-issues/9049/entity/post.entity.ts
+++ b/test/github-issues/9049/entity/post.entity.ts
@@ -1,5 +1,4 @@
 import { ObjectID } from "../../../../src/driver/mongodb/typings"
-import { Type } from "class-transformer"
 import { Comment } from "./comment"
 import { Column, Entity, ObjectIdColumn } from "../../../../src"
 
@@ -8,7 +7,6 @@ export class Post {
     @ObjectIdColumn()
     _id?: ObjectID
 
-    @Type(() => Comment)
     @Column(() => Comment)
     comments: Comment[]
 }
--- a/test/github-issues/9049/entity/post.entity.ts
+++ b/test/github-issues/9049/entity/post.entity.ts
@@ -1,0 +1,14 @@
+import { ObjectID } from "../../../../src/driver/mongodb/typings"
+import { Type } from "class-transformer"
+import { Comment } from "./comment"
+import { Column, Entity, ObjectIdColumn } from "../../../../src"
+
+@Entity()
+export class Post {
+    @ObjectIdColumn()
+    _id?: ObjectID
+
+    @Type(() => Comment)
+    @Column(() => Comment)
+    comments: Comment[]
+}
--- a/src/query-builder/transformer/DocumentToEntityTransformer.ts
+++ b/src/query-builder/transformer/DocumentToEntityTransformer.ts
@@ -143,12 +143,13 @@ export class DocumentToEntityTransformer {
                         entity[embedded.propertyName][column.propertyName] =
                             value
                     })
+
+                    addEmbeddedValuesRecursively(
+                        entity[embedded.propertyName],
+                        document[embedded.prefix],
+                        embedded.embeddeds,
+                    )
                 }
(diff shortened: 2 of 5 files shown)


test/github-issues/9049/entity/post.entity.ts
```diff
@@ -0,0 +1,14 @@
+import { ObjectID } from "../../../../src/driver/mongodb/typings"
+import { Type } from "class-transformer"
```
pleerock, line 2: is it required? Please remove it if not, we want to keep tests simple as possible

pauloFernandes: removed. Thanks

pleerock:
Thank you for contribution! 🎉

Issue #9015, opened 2022-05-18
@UpdateDateColumn does not update on upsert


udaya2899:
## Issue Description
@UpdateDateColumn column updates when using `save` or `update` method but not on `upsert` method 

### Expected Behavior
@UpdateDateColumn should update automatically with the current timestamp even if using `upsert` method

### Actual Behavior
@UpdateDateColumn does not update with the current timestamp if using upsert. It works only with `update` method

```
await this.ormRepository.update(customerEntity.customerId, customerEntity);
```

### Steps to Reproduce

1. Create a special column as follows:
```typescript
@UpdateDateColumn()
updatedAt: Date;
```
2. Try to update the entity containing the column using upsert.

### My Environment

| Dependency          | Version  |
| "nestjs/typeorm"              | "^8.0.0"      |
| Operating System |linux          |
| Node.js version     | 16.13.0  | 
| Typescript version  | 4.5.5 | 
| TypeORM version     | 0.2.38  | 

### Relevant Database Driver(s)

| DB Type           | Reproducible  |
|-------------------| ---           |
| `aurora-mysql`    | no            |
| `aurora-postgres` | no            |
| `better-sqlite3`  | no            |
| `cockroachdb`     | no            |
| `cordova`         | no            |
| `expo`            | no            |
| `mongodb`         | no            |
| `mysql`           | no            |
| `nativescript`    | no            |
| `oracle`          | no            |
| `postgres`        | yes            |
| `react-native`    | no            |
| `sap`             | no            |
| `spanner`         | no            |
| `sqlite`          | no            |
| `sqlite-abstract` | no            |
| `sqljs`           | no            |
| `sqlserver`       | no            |

### Are you willing to resolve this issue by submitting a Pull Request?

 - ✖️ Yes, I have the time, and I know how to start.
 - ✅ Yes, I have the time, but I don't know how to start. I would need guidance.
 - ✖️ No, I don’t have the time, but I can support (using donations) development.
 - ✖️ No, I don’t have the time and I’m okay to wait for the community / maintainers to resolve this issue.

zevstravitz:
I wonder if this should be a flag instead of being automatically applied.

I'd argue that automatically updating an UpdateDate column is indeed the right behavior, but might lead to breaking changes so adding a flag might be a good intermediate?

udaya2899:
[...]

PR #9064, opened 2022-06-02
@UpdateDateColumn support for Upsert

zevstravitz:
### Description of change
tldr; This PR fixes upsert not automatically updating an UpdateDate column by adding any UpdateDate column to the list of overwrite columns for the upsert method.

If you provide an UpdateColumn column name to an orUpdate method's overwrite columns, you don't actually have to provide the new value because typeorm infers it. But with upsert, we update a list of columns implicitly provided from our list of entities, which should not have to include an UpdateDate column value. This PR resolves this problem by _automatically_ adding any UpdateDate column to the list of overwrite columns for the upsert method.

### Pull-Request Checklist

fix: Adds support for @UpdateDateColumn support for Upsert method

Closes: #9015


pleerock:
Looks like tests are failing.

Thore1954:
Any updates on this?

Thore1954:
@pleerock I manually tested this fork and it's working. I think the tests are failing because they don't expect the update date column to be updated!
My current work around is:
```ts
await repository.upsert(
    [
        { externalId: "abc123", firstName: "Rizzrak", updatedAt: () => `now()` },
        { externalId: "bca321", firstName: "Karzzir", updatedAt: () => `now()` },
    ],
    ["externalId"],
)
```
But it only works on postgres so would love to get rid of it as soon as possible.

Thore1954:
Actually, there's currently no assertion on the update date column in tests:
https://github.com/typeorm/typeorm/blob/85fa9c6e7df1a0e3ba8c95d64d9b2a6ab464e0e0/test/functional/repository/basic-methods/repository-basic-methods.ts#L500-L505
Why are the tests failing? @zevstravitz @pleerock

Thore1954:
I found out the problem is the `skipUpdateIfNoValuesChanged` option is not respected by this change.
One solution is to exclude update date column from the condition:
https://github.com/typeorm/typeorm/blob/0d72317b3db3f93ce9d4c2f6f027c7486bbed08b/src/query-builder/InsertQueryBuilder.ts#L542-L551
Should be changed to:
```ts
                    query += overwrite
                        .filter(column => !(this.expressionMap.mainAlias!.hasMetadata && this.expressionMap.mainAlias!.metadata.findColumnWithDatabaseName(column)!.isUpdateDate))
                        .map(
                            (column) =>
                                `${tableName}.${this.escape(
                                    column,
                                )} IS DISTINCT FROM EXCLUDED.${this.escape(
                                    column,
                                )}`,
                        )
                        .join(" OR ")
```

src/entity-manager/EntityManager.ts
```diff
@@ -704,10 +704,11 @@ export class EntityManager {
         const overwriteColumns = metadata.columns.filter(
             (col) =>
                 !conflictColumns.includes(col) &&
-                entities.some(
+                (entities.some(
                     (entity) =>
                         typeof col.getEntityValue(entity) !== "undefined",
-                ),
+                ) ||
+                    col.isUpdateDate),
```
Thore1954, line 711: When `skipUpdateIfNoValuesChanged` option is set, it shouldn't add update date column otherwise it will update even when nothing has changed.

Thore1954 (changes requested):

Thore1954:
@zevstravitz thanks for the PR! Unfortunately, this wouldn't work because when `skipUpdateIfNoValuesChanged` option is set it won't skip update when nothing has changed.

AlexMesser:
Closing due to inactivity and failing tests. Feel free to submit a new PR.

PR #9089, opened 2022-06-08
Feature: Add column ordering to Postgres/Mysql/MsSQL

### LINKED ISSUE #541: Allow to order columns
When you create an entity with an extended class (createdAt, updatedAt,...) the columns from the extended class will created first. It would be nice if we can set a position for these fields.

like this:

```typescript
export class EntityBase {

  @PrimaryGeneratedColumn({position: -1})
  id: number;

  @UpdateDateColumn({position: 100})
  updatedAt: Date;

  @CreateDateColumn({position: 101})
  createdAt: Date;

}
```

default value should be 0.

kienmatu:
### Description of change
Feat: Implement column ordering option
Closes #541 
Added new option to column option, its default value is 0.
From now on, we can re-order the column when creating the table. For example:
```ts
@Entity()
export class Post {
  @PrimaryGeneratedColumn()
  id: number

  @Column({ order: 5 })
  title: string

  @Column({ order: 1 })
  thirdColumn: string

  @Column("text")
  text: string

  @ManyToMany((type) => Category)
  @JoinTable()
  categories: Category[]
}
```

### Pull-Request Checklist


fahim152 (commented):
I personally dont think its a good way of doing it ! 

if you have 40 columns in your table structure, would you count your desired column's position to add your new column in between them ? it will be messy eod.

it could be something like

  @Column({ after: 'that_column })
  this_column: boolean;
  
  //or 
  
  @Column({ before: 'this_column })
  new_column: boolean;

GalvinGao:
> I personally dont think its a good way of doing it !
> 
> if you have 40 columns in your table structure, would you count your desired column's position to add your new column in between them ? it will be messy eod.
> 
> it could be something like
> 
> @column({ after: 'that_column }) this_column: boolean;
> 
> //or
> 
> @column({ before: 'this_column }) new_column: boolean;

Personally I think adding the `order` parameter is a step forward for TypeORM. Afterwards we could also add `before` and `after` support, with exclusive checks (so you can opt-in to use only either `before`/`after` OR `order` but not both).

pleerock:
Thank you for contribution!

Proposed functionality isn't finished - it doesn't handle orders when we want to update orders when we add a new column to the table.

C0kkie:
Any update on this? @kienmatu

kienmatu:
Hi @C0kkie , With Mysql, it is okay, but it will be a problem with Postgresql, because it doesnt support re-ordering.
I wonder if i can do this for mysql only?

Read more:
https://tableplus.com/blog/2018/09/postgresql-is-it-possible-to-alter-column-order-position-in-a-table.html

pleerock:
I don't think we'll properly able to support this feature, simply because most of the RDBMSes doesn't support the feature we are trying to implement. Closing for a better times.

PR #9146, opened 2022-06-25
Fix: Call `dataSource.destroy` only on initialized dataSource in commands

pulzarraider:
### Description of change

If there is some problem in the configuration options in DataSource (e.g wrong hostname), the dataSource is not initialized, so calling `dataSource.destroy()` will cause error `Cannot execute operation on "${connectionName}" connection because connection is not yet established.`. I add simple check if the dataSource is initialized, so this error will not occur and user can focues on the real problem.

### Pull-Request Checklist

- [ ] Documentation has been updated to reflect this change (not needed)


pleerock:
Thank you for contribution! 🎉

PR #9182, opened 2022-07-06
Add note in select-query-builder.md

CasheeeewNuts:
### Description of change

Add note to Getting raw results in select-query-builder.md.

Because that [example](https://typeorm.io/select-query-builder#getting-raw-results) is misreading as if `getRawOne` or `getRawMany` returns row includes numerical fields.
But, they will return the values returned from database driver, So, fields in a row usually returned as string, 

### Pull-Request Checklist

### DIFF
--- a/docs/select-query-builder.md
+++ b/docs/select-query-builder.md
@@ -838,8 +838,10 @@ const photosSums = await dataSource
     .groupBy("user.id")
     .getRawMany()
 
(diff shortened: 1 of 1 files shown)


pleerock:
It's really up to underlying driver on how it returns the values. Some of them return numbers, some of them strings. So, this change might confuse users even more. We can add another explanation, but not this one. Thank you for your efforts!

Issue #9194, opened 2022-07-08
Oracle DB: schema is not supported


ertl:
## Issue Description

When using a Database schema for oracle, the  the queries do not work!

### Expected Behavior

The generated queries schould look like this:

```
SELECT * FROM SCHEMA."TABLE"
```

### Actual Behavior

But the generated queries look like

```
SELECT * FROM "SCHEMA"."TABLE"
```

Notice the " around the schema!

### Steps to Reproduce

1. You need 2 DB Users ( UserA and UserB)
2. All Tables are int he Schema of UserB
3. UserA has SELECT, UPDATE, INSERT, DELETE rights for the all tables in the schema UserB
4. The typeorm application uses UserA
5. Set schema in the ConnectionOptions to UserB

```typescript
// insert code here
```

### My Environment

| Dependency          | Version  |
| ---                 | ---      |
| Operating System    |          |
| Node.js version     | v16.15.1  | 
| Typescript version  | 4.6.3  | 
| TypeORM version     | 0.3.6  | 

### Relevant Database Driver(s)

| DB Type           | Reproducible  |
|-------------------| ---           |
| `aurora-mysql`    | no            |
| `aurora-postgres` | no            |
| `better-sqlite3`  | no            |
| `cockroachdb`     | no            |
| `cordova`         | no            |
| `expo`            | no            |
| `mongodb`         | no            |
| `mysql`           | no            |
| `nativescript`    | no            |
| `oracle`          | yes            |
| `postgres`        | no            |
| `react-native`    | no            |
| `sap`             | no            |
| `spanner`         | no            |
| `sqlite`          | no            |
| `sqlite-abstract` | no            |
| `sqljs`           | no            |
| `sqlserver`       | no            |

### Are you willing to resolve this issue by submitting a Pull Request?

 - ✖️ Yes, I have the time, and I know how to start.
 - ✖️ Yes, I have the time, but I don't know how to start. I would need guidance.
 - ✖️ No, I don’t have the time, but I can support (using donations) development.
 - ✅ No, I don’t have the time and I’m okay to wait for the community / maintainers to resolve this issue.

ertl:
No bug at all.
I just found out I had a wrong configuration. Because the statement SELECT * FROM "SCHEMA"."TABLE" is case sensitive and SELECT * FROM SCHEMA."TABLE" is not.

PR #9195, opened 2022-07-08
Update QueryBuilder.ts

ertl:
fix: do not escape schema name for oracle

closes: #9194

### Description of change

The fix does not escape schema names for oracle, when a schema is configured.
So the generated queries look like `SELECT * FROM SCHEMA."TABLE"`.

Without the fix, the generated statements `SELECT * FROM "SCHEMA"."TABLE"` threw an error.

### Pull-Request Checklist

### DIFF
--- a/src/query-builder/QueryBuilder.ts
+++ b/src/query-builder/QueryBuilder.ts
@@ -623,14 +623,19 @@ export abstract class QueryBuilder<Entity> {
      * schema name, otherwise returns escaped table name.
      */
     protected getTableName(tablePath: string): string {
-        return tablePath
-            .split(".")
-            .map((i) => {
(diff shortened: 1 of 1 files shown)


pleerock:
It doesn't feel like a proper solution. Why "SCHEMA"."TABLE" would work and SCHEMA."TABLE" would not? What is the proper name of the schema in your setup?

ertl:
@pleerock you're  right.  There is no bug at all.
I just found out I had a wrong configuration. Because the statement SELECT * FROM "SCHEMA"."TABLE" is case sensitive and SELECT * FROM SCHEMA."TABLE" is not.

Sorry for bothering you.

Issue #9055, opened 2022-05-31
Google Cloud - Postgres keep getting "server closed the connection unexpectedly"


dgobaud:
## Issue Description

Google Cloud - App Engine and Firebase Functions connecting to Postgres database continuously get error "server closed the connection unexpectedly" or used to be "Connection terminated unexpectedly"

Seems to be same as https://github.com/typeorm/typeorm/issues/5223 - issue is not fixed.

### Expected Behavior

Correctly manage and close / remove connections from the pool - something is not managing them correctly.

### Actual Behavior

continuously get error "server closed the connection unexpectedly" or used to be "Connection terminated unexpectedly"

### Steps to Reproduce

1.
2.

```typescript
// insert code here
```

### My Environment

| Dependency          | Version  |
| ---                 | ---      |
| Operating System    |          |
| Node.js version     | 16.13.2  | 
| TypeORM version     | 0.2.38  |

### Relevant Database Driver(s)

| DB Type           | Reproducible  |
|-------------------| ---           |
| `postgres`        | yes            |

### Are you willing to resolve this issue by submitting a Pull Request?

 - ✅ No, I don’t have the time, but I can support (using donations) development.
 - ✅ No, I don’t have the time and I’m okay to wait for the community / maintainers to resolve this issue.

cjnoname:
https://github.com/apraamcos/typeorm/tree/beta

This branch can solve your problem.

dgobaud:
> https://github.com/apraamcos/typeorm/tree/beta
> 
> This branch can solve your problem.

awesome! can you link to the commit that fixes the issue? am curious what the bug is

felipecsl:
This issue continues happening for us with typeORM v0.3.20

dgobaud:
I don't believe this has been fixed.

Issue #5112, opened 2019-11-19
Broken Postgres clients are released back into the pool, but should be removed


clarkdave:
**Issue type:**

[x] bug report

**Database system/driver:**

[x] `postgres`

It may impact other drivers if they have similar semantics/expectations as `pg.Pool`.

**TypeORM version:**

[x] `latest`

**Explanation of the problem**

Currently in TypeORM, if a client suffers an unrecoverable error - for example, if the underlying connection goes away, such as during a DB failover - there is no protection in place to stop that client being added back into the `pg.Pool`. This broken client will then be handed out in future even though it'll never be able to execute a successful query.

Although `pg.Pool` itself does listen for error events on clients within the pool - and actively removes any which do emit errors - it doesn't catch everything. It is considered the responsibility of the user of the pool to release known broken clients by calling `client.release(true)`. The truthy argument tells the pool to destroy the connection instead of adding it back to the pool

https://node-postgres.com/api/pool#releaseCallback

> The releaseCallback releases an acquired client back to the pool. If you pass a truthy value in the err position to the callback, instead of releasing the client to the pool, the pool will be instructed to disconnect and destroy this client, leaving a space within itself for a new client.

If a client's connection *does* break, it's very difficult to debug, as a handful of queries will begin failing with an error like `Error: Client has encountered a connection error and is not queryable` while others will continue executing fine.

There is further discussion about the impact of this in the `node-postgres` repo: https://github.com/brianc/node-postgres/issues/1942

**Steps to reproduce or a small repository showing the problem**

- Set pool size to 1
- Run a small loop which repeatedly performs a query, e.g.

```ts
while (true) {
  try {
    await getManager().query('select pg_sleep(4)')
    console.log('success')
  } catch (err) {
    console.log('error', err)
  }

  console.log('done')
  await sleep(1000)
}
```

- Wait for the query to run at least once. This should allocate a connection from the pool. Subsequent invocations of the query should use the same connection.
- Kill the connection while query is running. On a Mac or Linux, this is easily done by finding and killing the process, e.g. `ps aux | grep postgres | grep SELECT`
- The loop continue to throw errors, as the broken connection has been returned to the pool and continues to be handed out. This shouldn't happen: we should have told `pg.Pool` the connection is broken so it can be removed from the pool, so the next query should get a new, unbroken, connection

[...]

PR #9209, opened 2022-07-16
fix: pass error to pgpool release

bryfox:
### Description of change

I believe the intent in #7792 was to pass the error, if any, to the release callback from `pg-pool`, but it doesn't look like the error wasn't actually propagated through that release callback. The original issue (#5112) describes the motivation. #9055 also looks related.

Existing tests don't appear to cover this branch, and I haven't added any here. I validated this by adding logging and following the suggested steps in #5112: issue a long-running query, and kill the process or container. I note that with the steps I took, pg-pool still managed to clean up properly by seeing that [the client is not `_queryable`](https://github.com/brianc/node-postgres/blob/4fa7ee891a456168a75695ac026792136f16577f/packages/pg-pool/index.js#L331), but I imagine there are cases where this wouldn't happen.

### Pull-Request Checklist

### DIFF
--- a/src/driver/postgres/PostgresQueryRunner.ts
+++ b/src/driver/postgres/PostgresQueryRunner.ts
@@ -93,12 +93,12 @@ export class PostgresQueryRunner
 
                     const onErrorCallback = (err: Error) =>
                         this.releasePostgresConnection(err)
-                    this.releaseCallback = () => {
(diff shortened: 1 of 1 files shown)


pleerock:
Thank you for contribution! 🎉

Issue #9100, opened 2022-06-12
SQLjs 'cannot commit - no transaction is active'


Ainias:
## Issue Description

When autosave is used with sqljs and localforage, sometimes there is an error with the message `'cannot commit - no transaction is active`. As far as I can tell, it happens when two or more write-requests are made at the same time. Both start, then the first one ends, the autosave is called. The autosave exports the sqljs-database, which closes the transaction silently. Then the second one ends and tries to close the connection, which is not open anymore. Therefore the error is thrown.
Potentially can this resolve in an unwanted state. If I try to save A & B in separate repositories and B has a dependency C, which must also be saved, then it can happen, that A & B are saved, then the DB is exported, tries to save C, but there is an error in C, such that it tries to rollback the transaction containing B & C. But there is no transaction anymore, which means that the changes in B are saved, but not the one in C and therefore the data is not correct.

### Expected Behavior

Call autosave only when no transaction is active in order to not run into errors

### My Environment

| Dependency          | Version  |
| ---                 | ---      |
| Operating System    |          | Linux
| Browser     | Brave/Chrome    | 
| Typescript version  | 4.6.3  | 
| TypeORM version     | 0.3.5  | 
| sql.js | 1.7.0 |
| localforage | 1.10.0

### Relevant Database Driver(s)

| DB Type           | Reproducible  |
|-------------------| ---           |
| `aurora-mysql`    | no            |
| `aurora-postgres` | no            |
| `better-sqlite3`  | no            |
| `cockroachdb`     | no            |
| `cordova`         | no            |
| `expo`            | no            |
| `mongodb`         | no            |
| `mysql`           | no            |
| `nativescript`    | no            |
| `oracle`          | no            |
| `postgres`        | no            |
| `react-native`    | no            |
| `sap`             | no            |
| `spanner`         | no            |
| `sqlite`          | no            |
| `sqlite-abstract` | no            |
| `sqljs`           | yes            |
| `sqlserver`       | no            |

### Are you willing to resolve this issue by submitting a Pull Request?

 - ✅ Yes, I have the time, and I know how to start.
 - ✖️ Yes, I have the time, but I don't know how to start. I would need guidance.
 - ✖️ No, I don’t have the time, but I can support (using donations) development.
 - ✖️ No, I don’t have the time and I’m okay to wait for the community / maintainers to resolve this issue.

BARNZ:
Running into the same problem as well ever since upgrading to the latest version of TypeORM.

[...]

PR #9101, opened 2022-06-12
SQLjs fixes autosave while multiple transactions are active

Ainias:
### Description of change

I added a new function `saveExport` to the SqljsDriver. This function waits for any running transaction to finish before an export is called. For this I added a promise to the SqljsQueryRunner which is resolved as soon as all transactions are finished.
When the export-function of the SqljsDatabase is called, it closes all transactions. If one transaction is active, typeorm tries to close it after the export, which resolves in an error thrown.

I didn't got sqljs to work with the tests, as there is no configuration inside the ormconfig.json.dist for sqljs (and it runs only inside the browser. But I tested it with my own code and it's working after the change I made to typeorm. If someone tells me how I set up sqljs for testing, I'm happy to provide a testcase to cover this fix. 

Fixes #9100 

### Pull-Request Checklist

- [x] Code is up-to-date with the `master` branch
- [x] `npm run format` to apply prettier formatting
- [x] `npm run test` passes with this change
- [x] This pull request links relevant issues as `Fixes #0000`
- [ ] There are new or updated unit tests validating the change
- [ ] Documentation has been updated to reflect this change
- [x] The new commits follow conventions explained in [CONTRIBUTING.md](https://github.com/typeorm/typeorm/blob/master/CONTRIBUTING.md)

PR #9234, opened 2022-07-22
fix: fix "Cannot commit, no transaction is active" error in sql.js

Obiwan1995:
Fixes: #9100

### Description of change

When calling `autoSave` in a transaction, typeorm automatically closes the transaction in the `export` method. Therefore, when calling `commitTransaction`, the transaction is already committed and sqljs throws an exception "Cannot commit, no transaction is active".
This fix disables `autoSave` during the transaction to avoid this. Of course, `autoSave` is still called after the transaction, when calling commit. 

There was another PR for this issue (https://github.com/typeorm/typeorm/pull/9101) but this fix is not working when saving the database inside a transaction. If we call a `find` on the repository, typeorm releases the connection and it's calling `autoSave`. Waiting for the transaction to finish before saving was not working as we wait indefinitely.

### Pull-Request Checklist

### DIFF
--- a/src/driver/sqljs/SqljsDriver.ts
+++ b/src/driver/sqljs/SqljsDriver.ts
@@ -199,9 +199,10 @@ export class SqljsDriver extends AbstractSqliteDriver {
      * If a custom autoSaveCallback is specified, it get's called with the database as Uint8Array,
      * otherwise the save method is called which saves it to file (Node.js), local storage (browser)
      * or indexedDB (browser with enabled useLocalForage option).
+     * Don't auto-save when in transaction as the call to export will end the current transaction
      */
     async autoSave() {
-        if (this.options.autoSave) {
+        if (this.options.autoSave && !this.queryRunner?.isTransactionActive) {
             if (this.options.autoSaveCallback) {
                 await this.options.autoSaveCallback(this.export())
             } else {
--- a/src/driver/sqljs/SqljsQueryRunner.ts
+++ b/src/driver/sqljs/SqljsQueryRunner.ts
@@ -66,7 +66,9 @@ export class SqljsQueryRunner extends AbstractSqliteQueryRunner {
      */
     async commitTransaction(): Promise<void> {
         await super.commitTransaction()
-        await this.flush()
+        if (!this.isTransactionActive) {
+            await this.flush()
+        }
     }
 
     /**


src/driver/sqljs/SqljsDriver.ts
```diff
@@ -199,9 +199,10 @@ export class SqljsDriver extends AbstractSqliteDriver {
      * If a custom autoSaveCallback is specified, it get's called with the database as Uint8Array,
      * otherwise the save method is called which saves it to file (Node.js), local storage (browser)
      * or indexedDB (browser with enabled useLocalForage option).
+     * Don't auto-save when in transaction as the call to export will end the current transaction
      */
     async autoSave() {
-        if (this.options.autoSave) {
+        if (this.options.autoSave && !this.queryRunner?.isTransactionActive) {
```
pleerock, line 205: shouldn't we throw an error if one tries to autoSave during active transaction? Imagine someone called autoSave but it didn't save anything if transaction was active, and user still thinks everything was saved.

Obiwan1995: I get your point though it would be a bit tricky to do it as this method is automatically called on query runner release if database has changed (code from `SqljsQueryRunner.ts`):
```ts
private async flush() {
    if (this.isDirty) {
            await this.driver.autoSave()
            this.isDirty = false
    }
}

async release(): Promise<void> {
    await this.flush()
    return super.release()
}
```

If we throw an error everytime this gets called within a transaction, we will end up with tons of errors as a simple `update` on the repository during a transaction will trigger it. Is there a way to inform the use without throwing an error?

pleerock: okay, another question:

shouldn't we add the proposed check into `commitTransaction` override?:

```ts
    async commitTransaction(): Promise<void> {
        await super.commitTransaction()
        if (this.isTransactionActive === false)
            await this.flush()
    }
```

Obiwan1995: You want to add it there to ensure that the database is saved after the transaction?

Currently, `COMMIT` sets `isDirty` to `true`, then, when the query runner is released after the query, it automatically flushes the database as there is no active transaction.

So, database is already saved after transaction but it would be easier to understand I admit.

pleerock: It makes sense to have it there I think. We flush after the transaction. If do it only if transaction was committed.

Obiwan1995:
Any news on merging this?
My team had to fork the repository because of this so it would be nice to have it in the next release.

pleerock:
Thank you for contribution! 🎉

PR #9262, opened 2022-08-05
fix: find query mongodb properly with @DeleteDateColumn()

### LINKED ISSUE #9250: MongoDB - FindOne / Find doesn't work with @DeleteDateColumn()
## Issue Description
Find, FindBy, FindOne, FindOneBy all do not work properly when @DeleteDateColumn() is used 

DB: MongoDB
Typescript 4.7.4
Typeorm 0.3.7

Create a few entries with different names, try to find 1 entry by using
await VenueSVG.find({ where: { name: "test-7" } });
await VenueSVG.findOne( { name: "test-7" } );

### Expected Behavior
Should result in getting entries where name == "test-7"

### Actual Behavior
- Returns the 1st entry when using findOne()
- Returns all entries when using find()

### Steps to Reproduce
`export class VenueSVG extends BaseEntity {
  @ObjectIdColumn()
  id!: ObjectID;

  @Column()
  name!: string;

 @DeleteDateColumn()
 deletedDate!: Date;
}
`
1. Create a few entries i.e. "test-1", "test-2" etc.
2. Use the find methods await VenueSVG.find({ where: { name: "test-7" } }); Or await VenueSVG.findOne( { name: "test-7" } );

roseline124:
### Description of change

Fixes #9250 

Fix `filterSoftDelted` of `MongoEntityManager` not to reset query filter.
Previous version replaced cursor query filter with `where this.deletedDateColumn.propertyName == null`. 
So there is a bug that the cursor queries all of not deleted data if the entity has DeletedDateColumn.

### Pull-Request Checklist

- [x] Documentation has been updated to reflect this change (interface was not changed)

### DIFF
--- a/src/entity-manager/MongoEntityManager.ts
+++ b/src/entity-manager/MongoEntityManager.ts
@@ -1170,12 +1170,13 @@ export class MongoEntityManager extends EntityManager {
         deleteDateColumn: ColumnMetadata,
         query?: ObjectLiteral,
     ) {
+        const { $or, ...restQuery } = query ?? {}
         cursor.filter({
             $or: [
-                { [deleteDateColumn.propertyName]: { $exists: false } },
                 { [deleteDateColumn.propertyName]: { $eq: null } },
+                ...(Array.isArray($or) ? $or : []),
             ],
-            ...query,
+            ...restQuery,
         })
     }
 
--- a/src/entity-manager/MongoEntityManager.ts
+++ b/src/entity-manager/MongoEntityManager.ts
@@ -180,10 +180,10 @@ export class MongoEntityManager extends EntityManager {
                     ),
                 )
             if (deleteDateColumn && !optionsOrConditions.withDeleted) {
-                this.filterSoftDeleted(cursor, deleteDateColumn)
+                this.filterSoftDeleted(cursor, deleteDateColumn, query)
             }
         } else if (deleteDateColumn) {
-            this.filterSoftDeleted(cursor, deleteDateColumn)
+            this.filterSoftDeleted(cursor, deleteDateColumn, query)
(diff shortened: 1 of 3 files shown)


pleerock:
Thank you for contribution! 🎉

serebro:
I have found a problem with current solution.

Let say I have three records in a collection `users`

```
[
    { "username" : "user1", "deletedAt" : null) },
    { "username" : "user2", "deletedAt" : ISODate("2023-09-02T18:26:48.253+0000") },
    { "username" : "user3" }
]
```

and `Users` entity
```
@Entity()
export class Users {
    @ObjectIdColumn()
    id: ObjectId

    @Column()
    username: string

    @DeleteDateColumn()
    deletedAt: Date
}
```

I make a query
```
const options = {
        where: {
            $or: [
                { username: { $regex: '^user' } },
            ]
        }
}
DataSource.manager.find(Users, options)
```

My query transforms by the current `filterSoftDeleted` function to this mongodb query
```
{
    $or: [
        { deletedAt: { $eq: null } },
        { username: { $regex: '^user' } },
    ]
}
```

As a result I receive all three records instead of two rows (one record "user2" is marked as deleted).
Expected result should be the following two records
```
[
    { "username" : "user1", "deletedAt" : null) },
    { "username" : "user3" }
]
```

It can be achieved by the following code

```
    protected filterSoftDeleted<Entity>(
        cursor: FindCursor<Entity>,
        deleteDateColumn: ColumnMetadata,
        query?: ObjectLiteral,
    ) {
        const q = query ?? {}
        q[deleteDateColumn.propertyName] = { $ne: { $type: 9 } }
        cursor.filter(q);
    }
```

PR #9288, opened 2022-08-15
Exiting successful process with 0

### LINKED ISSUE #9287: CLI command migration:generate should exit process
### Expected Behavior

migration:generate should exit process after successful operation

### Actual Behavior
It hangs open

### Steps to Reproduce
1. generate a migration using CLI

### My Environment
Plain TypeScript project, not NestJS.

| Operating System    |  macOS Monterey 12.5  |
| Node.js version     | v14.19.1  | 
| Typescript version  | v4.6.3  | 
| TypeORM version     | v0.3.7  | 

needim:
### Description of change

Exiting successful process with 0 for migration:generate CLI command.

### Pull-Request Checklist

- [ ] `npm run test` passes with this change N/A
- [x] This pull request links relevant issues as `Fixes #9287`

### DIFF
--- a/src/commands/MigrationGenerateCommand.ts
+++ b/src/commands/MigrationGenerateCommand.ts
@@ -196,6 +196,7 @@ export class MigrationGenerateCommand implements yargs.CommandModule {
                         )} has been generated successfully.`,
                     ),
                 )
+                process.exit(0)
             }
         } catch (err) {
             PlatformTools.logCmdErr("Error during migration generation:", err)


src/commands/MigrationGenerateCommand.ts
```diff
@@ -196,6 +196,7 @@ export class MigrationGenerateCommand implements yargs.CommandModule {
                         )} has been generated successfully.`,
                     ),
                 )
+                process.exit(0)
```
pleerock, line 199: It's default behavior. Process will exist with code 0 once execution is finished anyway. Why do we need to add this line?

a-soppe-emit: I don't know about others here, but my TypeORM commands never exit (using version 0.3.7)

pleerock: we need to understand why. Force exist isn't an option if there are some resources that got busy. We may "hide" the actual bug this way.

a-soppe-emit: Haven't looked that far yet why it is behaving this way. It just gets stuck after execution. However, TypeORM CLI did properly exit for version 0.2.x.

needim: Migration generate command was never existing for me. even before v0.3.x. (using mysql)

pleerock: Maybe same resolution was applied there. So better to find a valid reason this time.

H6LS1S: I too have the same problem with exit in nestjs if i using `app. enableShutdownHooks()`

lyanhhoang:
Facing the same issue with cache:clear and migration:revert

migration:run is OK as it has exit step https://github.com/typeorm/typeorm/blob/85fa9c6e7df1a0e3ba8c95d64d9b2a6ab464e0e0/src/commands/MigrationRunCommand.ts#L80

We have just been asked to upgrade to 0.3.x to get rid of sec vulnerability issue CVE-2022-33171. After upgrading, everything is fine except this issue. When facing this issue, we anyway still need to force stop by using Ctrl+C. The problem is more serious when run the command in the CI/CD pipeline, as process could not exit so the pipeline job keeps running but nothing happen. For the proposed approach to add exit command, better to check if adding exit command could break anything, if nothing could be broken, please merge the fix, the root cause should be investigated independently.

pleerock:
Okay I can merge it since it shouldn't really harm. But I still don't understand why it's needed.

jiak94:
@pleerock It seems like this `process.exit(0)` command causing the `npm test` exit earlier and not all the tests are executed. If you check the circleci log, `npm test` exits after `commands - generate` generate one migration file. If you disabled the `commands - generate` test, all other testscases will be executed and failed at `database schema > enums`

PR #9303, opened 2022-08-19
feat: implement exists query method

### LINKED ISSUE #2815: queryBuilder EXISTS query
**Issue type:**

[ ] question
[ ] bug report
[x] feature request
[ ] documentation issue

**Database system/driver:**

[ ] `cordova`
[ ] `mongodb`
[ ] `mssql`
[x] `mysql` / `mariadb`
[ ] `oracle`
[ ] `postgres`
[ ] `sqlite`
[ ] `sqljs`
[ ] `react-native`
[ ] `expo`

**TypeORM version:**

[x] `latest`
[ ] `@next`

**Feature request:**

Looks like EXISTS is optimised more, if you need to check only rows existence.
So, instead of

```sql
SELECT COUNT(*) FROM mytable WHERE a = "foo" AND b = "bar";
```

better do

```sql
SELECT EXISTS(SELECT * FROM mytable WHERE a = "foo" AND b = "bar");
```

But now query builder doesn't support such kind of queries, you need to write a raw query.
I think it would be cool to have something like

```js
queryBuilder()
  .select('*')
  .from(MyTable, 'tbl')
  .where('tbl.a = :foo AND tbl.b = :bar', { foo, bar })
  .isExists(); // <-- this row
```

mortezaPRK:
Adding `Exists` method to query builder and EntityManager, to check whether a row exists given the conditions

Closes: #2815

### Description of change

Added a new methods to check if a row exists. 
Also added a method to add `EXISTS` clause to a given query.

### Pull-Request Checklist

- [X] This pull request links relevant issues as `Fixes #2815`

### DIFF
--- a/src/entity-manager/EntityManager.ts
+++ b/src/entity-manager/EntityManager.ts
@@ -947,6 +947,23 @@ export class EntityManager {
     }
 
     /**
+     * Checks whether any entity exists with the given condition
+     */
+    exists<Entity>(
+        entityClass: EntityTarget<Entity>,
+        options?: FindManyOptions<Entity>,
+    ): Promise<boolean> {
+        const metadata = this.connection.getMetadata(entityClass)
+        return this.createQueryBuilder(
+            entityClass,
+            FindOptionsUtils.extractFindManyOptionsAlias(options) ||
+                metadata.name,
+        )
+            .setFindOptions(options || {})
+            .getExists()
+    }
+
+    /**
      * Counts entities that match given options.
      * Useful for pagination.
      */
--- a/src/query-builder/SelectQueryBuilder.ts
+++ b/src/query-builder/SelectQueryBuilder.ts
@@ -249,6 +249,10 @@ export class SelectQueryBuilder<Entity extends ObjectLiteral>
         return this
     }
 
+    fromDummy(): SelectQueryBuilder<any> {
+        return this.from(this.connection.driver.dummyTableName ?? "(SELECT 1 AS dummy_column)", "dummy_table");
+    }
+
     /**
      * Specifies FROM which entity's table select/update/delete will be executed.
      * Also sets a main string alias of the selection data.
@@ -1191,6 +1195,27 @@ export class SelectQueryBuilder<Entity extends ObjectLiteral>
     }
 
     /**
+     * Sets a new where EXISTS clause
+     */
+    whereExists(subQuery: SelectQueryBuilder<any>): this {
+        return this.where(...this.getExistsCondition(subQuery))
+    }
+
+    /**
+     * Adds a new AND where EXISTS clause
+     */
+    andWhereExists(subQuery: SelectQueryBuilder<any>): this {
+        return this.andWhere(...this.getExistsCondition(subQuery))
+    }
+
+    /**
+     * Adds a new OR where EXISTS clause
+     */
+    orWhereExists(subQuery: SelectQueryBuilder<any>): this {
+        return this.orWhere(...this.getExistsCondition(subQuery))
+    }
+
+    /**
      * Adds new AND WHERE with conditions for the given ids.
      *
      * Ids are mixed.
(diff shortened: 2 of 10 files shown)


src/query-builder/SelectQueryBuilder.ts
```diff
@@ -2864,6 +2929,18 @@ export class SelectQueryBuilder<Entity>
         return parseInt(results[0]["cnt"])
     }
 
+    protected async executeExistsQuery(
+        queryRunner: QueryRunner,
```
pleerock, line 2933: `queryRunner` is ignored here.

src/entity-manager/EntityManager.ts
```diff
@@ -945,6 +945,23 @@ export class EntityManager {
         }
     }
 
+    /**
+     * Checks whether any entity exists with the given condition
+     */
+    exists<Entity>(
```
pleerock, line 951: please also add tests checking this function and testing different `options?: FindManyOptions<Entity>`

Ginden:
I think it should throw if `QueryBuilder` has set `skip` or `take` options.

I'm not sure why `offset` is set to undefined.

```sql

SELECT * FROM foo
WHERE EXISTS(SELECT 1 FROM bar WHERE bar.foo_id = foo.id OFFSET 1)

```

This is recommended method to "get all Foos with two or more Bars" (`HAVING COUNT(1) > 1` usually requires reading entire table).

mortezaPRK:
> I think it should throw if `QueryBuilder` has set `skip` or `take` options.
> 
> I'm not sure why `offset` is set to undefined.
> 
> ```sql
> SELECT * FROM foo
> WHERE EXISTS(SELECT 1 FROM bar WHERE bar.foo_id = foo.id OFFSET 1)
> ```
> 
> This is recommended method to "get all Foos with two or more Bars" (`HAVING COUNT(1) > 1` usually requires reading entire table).

I think that would make confusion for the developers. `exists` meaning is very clear with current implementation.

Ginden:
> `exists` meaning is very clear with current implementation.

I'm not sure if "offset is silently discarded" is clear from signature or docs.

mortezaPRK:
> > `exists` meaning is very clear with current implementation.
> 
> I'm not sure if "offset is silently discarded" is clear from signature or docs.

Agree. but not sure :)


> > `exists` meaning is very clear with current implementation.
> 
> I'm not sure if "offset is silently discarded" is clear from signature or docs.

Agree. But I've never seen such a usage for `Exists` in SQL.

The change is simple. I just have to remove `offset(undefined)` call

mortezaPRK:
@pleerock 

If the implementation is OK, I can add the documentations too.

src/query-builder/SelectQueryBuilder.ts
```diff
@@ -2906,6 +2971,20 @@ export class SelectQueryBuilder<Entity>
...
+    ): Promise<boolean> {
+        const results = await this.connection
+            .createQueryBuilder()
+            .from("(SELECT 1 AS row_exists)", "row_exists_tbl")
```
alumni, line 2979: `FROM` clause is missing. Some dialects allow it, some not. Best thing is to add it manually:
```ts
qb
  .select("1", "row_exists")
  .from("SYS.DUMMY", "dummy_tbl")
  .whereExists(this)
```

alumni: IMO you could use the query builder without a `from`: `qb.select("1", "row_exists").whereExists(this)`.

You would then need to update the return value of `SelectQueryBuilder#createSelectExpression()` like this:

- if `froms` is not empty, keep the current behavior (add the concatenated list). 
- if it is empty and driver is oracle or sap, then add `FROM DUAL` / `FROM SYS.DUMMY` (like in the `InsertQueryBuilder`).
- if it is empty and the driver is something else then don't add `FROM` at all.

mortezaPRK: @alumni 

Thanks for the comment.

Initially, I tried your solution to handle the `exists` method. But it becomes very complex. I had to generate different queries for each driver (every driver had to implement a method for generating the right query), and the result value from each driver was different. in the end, It had lots of "if" statements.

I checked most of the drivers that Typeorm supports and came up with this query that is independent of the driver and the result from the drivers is the same for all of them.

alumni: @mortezaPRK well, as I said, this is driver-specific, so you need to handle some drivers in a different way. This is nothing special, it's done in quite a lot of places in the project.

`(SELECT 1 as row_exists)` throws a `QueryFailedError` on an Oracle / SAP connection because the `FROM` clause is missing, so special handling for them would be needed. And I would do this special handling in `SelectQueryBuilder#createSelectExpression()`, it would also allow selecting from dummy/dual inside the apps that use TypeORM.

Maybe one of the maintainers has a better idea here :)

mortezaPRK: Agree :)

I'll change the implementation ASAP

pleerock:
Thanks for the help on review @alumni @Ginden

src/query-builder/SelectQueryBuilder.ts
mortezaPRK: @alumni

I added `fromDummy` and it works perfectly find :)

Thanks for the idea.

pleerock:
I guess it's good to go. Please don't forget to add the docs to the changes in a separate PR. 
Thank you for contribution! 🎉

varanauskas:
Great addition, why does `Repository` API use `exist` instead of `exists`? `Entity Manager` API seems to use `exists`

Additionally the docs for this are still not visible on typeorm.io, is there an open PR or should I open one?

PR #9307, opened 2022-08-22
fix: resolve issues with quoting column names in groupBy and orderBy

### LINKED ISSUE #9293: No quotes while using orderBy, groupBy by custom column
## Issue Description

I am using Postgres and it is required to use quotes in the queries to wrap tables and columns.
In case when query builder wraps the query as a subquery into the SELECT DISTINCT request, it doesn't add quotes around the custom column added by addSelect.

### Expected Behavior
Typeorm wraps anything was added by addSelect into quotes.

```sql
SELECT DISTINCT "distinctAlias"."Offer_id" AS "ids_Offer_id", "distinctAlias"."rateSum"
FROM (SELECT "Offer"."id"       AS "Offer_id",
             "Offer"."lenderId" AS "Offer_lenderId",
             "Offer"."rate"     AS "Offer_rate",
             "Lender"."id"      AS "Lender_id",
             "Lender"."name"    AS "Lender_name",
             "Lender"."email"   AS "Lender_email",
             1                  AS "rateSum"
      FROM "Offer" "Offer"
               INNER JOIN "Lender" "Lender" ON "Lender"."id" = "Offer"."lenderId"
      GROUP BY "Offer"."id") "distinctAlias"
ORDER BY "distinctAlias".rateSum ASC, "Offer_id" ASC
OFFSET 10
```
"distinctAlias".rateSum should be quoted: "distinctAlias"."rateSum"

### Actual Behavior
Typeorm doesn't wrap column added by addSelect into quotes.
```sql
SELECT DISTINCT "distinctAlias"."Offer_id" AS "ids_Offer_id", "distinctAlias".rateSum
FROM (SELECT "Offer"."id"       AS "Offer_id",
             "Offer"."lenderId" AS "Offer_lenderId",
             "Offer"."rate"     AS "Offer_rate",
             "Lender"."id"      AS "Lender_id",
             "Lender"."name"    AS "Lender_name",
             "Lender"."email"   AS "Lender_email",
             1                  AS "rateSum"
      FROM "Offer" "Offer"
               INNER JOIN "Lender" "Lender" ON "Lender"."id" = "Offer"."lenderId"
      GROUP BY "Offer"."id") "distinctAlias"
ORDER BY "distinctAlias".rateSum ASC, "Offer_id" ASC
OFFSET 10
```
"distinctAlias".rateSum is not qouted

```
// include the output in code tags like these!
```

### Steps to Reproduce

1. clone https://github.com/redwert/typeorm-postgres-example.git
2. npm i
3. docker-compose up -d
4. npm run start

or take a look here:

```typescript
@Entity('Offer')
export default class Offer {
  @PrimaryGeneratedColumn('increment')
  id: number

  @Column('int')
  lenderId: number

  @Column('int')
  rate: number

  @ManyToOne(() => Lender, lender => lender.offers)
  @JoinColumn([ { name: 'lenderId', referencedColumnName: 'id' } ])
  lender: Lender
}

@Entity('Lender')
export default class Lender {
  @PrimaryGeneratedColumn('increment')
  id: number

  @Column('varchar')
  name: string

  @Column('varchar')
  email: string | null

  @OneToMany(() => Offer, offer => offer.lender)
  offers: Offer[]
}

(async () => {
    const conn = await createConnection(typeOrmConfig)
    const rateSumAlias = 'rateSum'

    await conn.getRepository(Offer)
      .createQueryBuilder('Offer')
      .addSelect('1', rateSumAlias)
      .innerJoinAndSelect('Offer.lender', 'Lender')
      .groupBy('Offer.id')
      .orderBy(rateSumAlias, 'ASC')
      .skip(10)
      .getRawAndEntities()
    
    await conn.close()
})();
```

### My Environment

| Operating System    |  Ubunut 20.04   |
| Node.js version     | 16.16.0  | 
| Typescript version  | 4.7.4  | 
| TypeORM version     | 0.3.7  | 

redwert:
### Description of change

Closes: #9293

Fix wrapping custom column into quotes in `orderBy` and `groupBy` statements.
Without this fix, PostgreSQL throws an error in the case of `addSelect` by the custom field.

### Pull-Request Checklist

- [x] This pull request links relevant issues as `Fixes #9293`

### DIFF
--- a/DEVELOPER.md
+++ b/DEVELOPER.md
@@ -46,7 +46,7 @@ git clone git@github.com:<github username>/typeorm.git
 # Go to the TypeORM directory:
 cd typeorm
 
-# Add the main TyepORM repository as an upstream remote to your repository:
+# Add the main TypeORM repository as an upstream remote to your repository:
 git remote add upstream https://github.com/typeorm/typeorm.git
 ```
 ## Installing NPM Modules
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -13,6 +13,20 @@ services:
       MYSQL_PASSWORD: "test"
       MYSQL_DATABASE: "test"
 
+  # mysql v8
+#  mysql8:
+#    image: "mysql:8"
+#    container_name: "typeorm-mysql8"
+#    ports:
+#      - "3308:3306"
+#    environment:
+#      MYSQL_ROOT_PASSWORD: "admin"
+#      MYSQL_USER: "test"
+#      MYSQL_PASSWORD: "test"
+#      MYSQL_DATABASE: "test"
+#    volumes:
+#      - mysql8_volume:/var/lib/mysql
+
   # mariadb
   mariadb:
     image: "mariadb:10.5.13"
@@ -127,8 +141,10 @@ services:
   #   container_name: "typeorm-redis"
   #   ports:
   #     - "6379:6379"
+
 #volumes:
 #  volume-hana-xe:
+#  mysql8_volume:
 
 networks:
   typeorm:
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -6,7 +6,7 @@ services:
     image: "mysql:5.7.37"
     container_name: "typeorm-mysql"
     ports:
-      - "3306:3306"
+      - "3301:3306"
     environment:
       MYSQL_ROOT_PASSWORD: "admin"
       MYSQL_USER: "test"
@@ -33,7 +33,7 @@ services:
     image: "mdillon/postgis:9.6"
     container_name: "typeorm-postgres"
     ports:
-      - "5432:5432"
+      - "5401:5432"
     environment:
       POSTGRES_USER: "test"
       POSTGRES_PASSWORD: "test"
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -6,7 +6,7 @@ services:
     image: "mysql:5.7.37"
     container_name: "typeorm-mysql"
     ports:
-      - "3301:3306"
+      - "3306:3306"
     environment:
       MYSQL_ROOT_PASSWORD: "admin"
       MYSQL_USER: "test"
@@ -33,7 +33,7 @@ services:
     image: "mdillon/postgis:9.6"
     container_name: "typeorm-postgres"
     ports:
-      - "5401:5432"
+      - "5432:5432"
     environment:
       POSTGRES_USER: "test"
       POSTGRES_PASSWORD: "test"
--- a/docs/data-source-options.md
+++ b/docs/data-source-options.md
@@ -345,7 +345,7 @@ Different RDBMS-es have their own specific options.
     SQL Server Availability Group. For more information, see here. (default: `false`).
 
 -   `options.encrypt` - A boolean determining whether or not the connection will be encrypted. Set to true if you're
-    on Windows Azure. (default: `false`).
+    on Windows Azure. (default: `true`).
 
 -   `options.cryptoCredentialsDetails` - When encryption is used, an object may be supplied that will be used for the
     first argument when calling [tls.createSecurePair](http://nodejs.org/docs/latest/api/tls.html#tls_tls_createsecurepair_credentials_isserver_requestcert_rejectunauthorized)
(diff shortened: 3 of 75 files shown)


pleerock:
what if we want to group / order by something that is not an alias? (e.g. `ORDER BY 1`)

redwert:
@pleerock 
Added changes to handle cases (e.g. ORDER BY 1)
Also, having build error, but I don't understand why and what should I fix

pleerock:
Code become weird after your changes. It definitely needs better implementation. 

But I don't think we should add this feature, simply because it's too hard to cover all cases. For example take a look what different expressions ["group by"](https://docs.microsoft.com/en-us/sql/t-sql/queries/select-group-by-transact-sql?view=sql-server-ver16) can accept. Also, if we add this feature to order by and group by - we will need to cover all other statements, e.g. SELECT which is too way complicated to cover.

Current behavior is the best tradeoff - it automatically escapes symbols of the aliases you use with the columns registered in the entities, for everything else user do manual escaping. This way we achieve both escaping of 95% use cases + flexibility in edge use cases.

redwert:
@pleerock I agree about groupBy and orderBy cases. Right now it works with the quoted column name e.g. `.addGroupBy(`"${aliasName}"`)`.
But this way it is needed to allow the ability to pass quoted column name to addSelect this way typeorm behaviour will be consistent.
In the example, we have the simple query with addSelect, addGroupBy, addOrderBy with the same alias.
```
const alias = 'rateSum'

repository
   .createQueryBuilder("Offer")
   .addSelect("1", alias)
   .innerJoinAndSelect("Offer.lender", "Lender")
   .groupBy("Offer.id")
   .addGroupBy("Lender.id")
   .addGroupBy(alias)
   .orderBy("Lender.id", "ASC")
   .addOrderBy("Offer.id", "ASC")
   .addOrderBy(alias, "ASC")
   .skip(1)
   .take(10)
   .getRawAndEntities()
```
And these are the outputs for different cases:

- without quoting aliases everywhere we have the next
image

- passing quoted alias in groupBy and orderBy (no "distinctAlias".rateSum in SELECT DISTINCT in query)
image

- passing quoted alias everywhere (addSelect, groupBy, orderBy)
image

As result in every scenario we have the wrong queries. In my understanding it can be two variants: 
- quoting the alias for orderBy and groupBy if an alias was not quoted (this is implemented right now)
- add the ability to pass quoted alias to addSelect (fix issue with `"" rateSum""` in SELECT)

So what would you suggest? I am ready to spend my time and close this bug but I need your assistance

pleerock:
`addSelect` second parameter is `alias`, so we already know that we are dealing with `alias`, that's why we automatically escape it. With order by / group by it's a bit more complicated since it's parameter is actual an expression, so manual escaping is required. That's explains why you have such results.

Regarding to how to handle the issue this PR addresses, see my [previous answer](https://github.com/typeorm/typeorm/pull/9307#issuecomment-1227218253).

redwert:
Umed, current behavior doesn't give an ability to add a custom column and make an order by it. There is no option to do it.

Please take a look on the last changes. It gives ability to pass escaped aliases to addSelect and solves this issue:
image

This doesn't break anything, and fixes issue with wrong query

pleerock:
okay... so now your changes are applied to addSelect...

Let me confirm one thing:

> current behavior doesn't give an ability to add a custom column and make an order by it. There is no option to do it.

what if you do:

```ts
.createQueryBuilder("Offer")
            .addSelect(`1 as "rateSum"`)
            .innerJoinAndSelect("Offer.lender", "Lender")
            .groupBy("Offer.id")
            .addGroupBy("Lender.id")
            .addGroupBy(`"rateSum"`)
            .orderBy("Lender.id", "ASC")
            .addOrderBy("Offer.id", "ASC")
            .addOrderBy(`"rateSum"`, "ASC")
```

are you **sure** it's not going to work?

redwert:
I have tried your variant with `.addSelect(`1 as "rateSum"`)` and `.addOrderBy(`"rateSum"`, "ASC")` but it also doesn't work.  There is no "distinctAlias"."rateSum" after the coma.
image

PR #9308, opened 2022-08-22
Order Sorting improvement

faizananwerali:
If the order is undefined or null then push them to last

Sample: 

```js
const columns = [
    {name: 'test 1'}, // Order will be 21 after sorting
    {name: 'test 2'}, // Order will be 22 after sorting
    {name: 'test 3'}, // Order will be 23 after sorting
    {order: 15},
    {}, // Order will be 24 after sorting
    {order: 20},
    {order: -4},
    {order: -7},
    {}, // Order will be 25 after sorting
];
//console.log('columns', columns);

const orders = columns.map(column => {
    //console.log('column', column.order);
    //console.log('typeof column', typeof column.order);
    //console.log('column.order || -1', column.order || -1)
    return (typeof column.order === 'undefined' || column.order === null) ? -1 : column.order;
});

//console.log('orders', orders);
let max = Math.max(...orders);
//console.log('max', max);

columns.sort(function(a, b){
    if (a.order && b.order) {
        return a.order - b.order;
    }
    if (typeof b.order === 'undefined' || b.order === null) {
        b.order = ++max;
    }
    if (typeof a.order === 'undefined' || a.order === null) {
        a.order = ++max;
    }
    return a.order - b.order;
});

console.log('columns', columns);


/**
 * Result:
 * columns [
 *   { order: -7 },
 *   { order: -4 },
 *   { order: 15 },
 *   { order: 20 },
 *   { name: 'test 1', order: 21 },
 *   { name: 'test 2', order: 22 },
 *   { name: 'test 3', order: 23 },
 *   { order: 24 },
 *   { order: 25 }
 * ]
 */
```


faizananwerali:
> Imagine I added a new column to exist entity with a specific order. Then run synchronization. Will it work?

Based on all the logic, it should work just fine.

pleerock:
Closing as there is no activity in this PR.

Issue #1082, opened 2017-10-24
Add trigger functionality


AlexMesser:
We can add basic functionality to allow users to create thier own triggers, for example:

```ts
@Trigger({
    name: "lalala",
    fires: "after",
    insert: true,
    update: true
})
export class UserDateUpdater {

    getSql() {
        return `
BEGIN
	...
END
`;
    }

}
```

mrkmg:
I am willing to implement this. Do we have a process for designing the api for something like this @pleerock? 

I am come up with a proposal if you want.

pleerock:
@mrkmg just take a look how other decorators are implemented. Similar to this will be a `@Subcriber` decorator.

You'll need to do it something like this:

1. decorator take all options ("args") and saves them in "metadata args storage"
2. connection on "connect" calls builds all "metadata args" into "metadata" (TriggerMetadata for example)
3. trigger metadata is stored somewhere, maybe in the connection object
4. when user calls a trigger (maybe connection.executeTrigger(UserDateUpdater)) it takes a metadata object from the connection and calls it `getSql` method and passes sql string to the `connection.query` method.

leesiongchan:
Why is there no progress after that? I think its a handy feature.

incompletude:
How should one implement this currently?

mrinvisible0:
Any news on this?

Murillo2380:
Maybe it is not needed, I believe you can use [listeners and subscribers](https://github.com/typeorm/typeorm/blob/master/docs/listeners-and-subscribers.md)

davidmurdoch:
@Murillo2380 this is not the same thing. Listeners and Subscribers only protects the database from the application using TypeORM making invalid modifications to the data, the database itself does not have the data integrity it needs.

null-char:
> How should one implement this currently?

@incompletude TypeORM already has decorators such as AfterInsert, AfterUpdate, BeforeRemove, etc which act as trigger functions. This issue — I presume — is to be able to use a single @Trigger() decorator for specifying when to run a specific trigger function. So you could have a trigger function execute both after an insert and an update.

patrickmichalina:
Will this be consider if I were to write it? I have some helper functions that create DB triggers in Postgres during migrations but it would be much nicer if this were automated within migrations ecosystem.

Typerom app  level subscribers are NOT the same as DB triggers and functions.

incompletude:
> > How should one implement this currently?
> 
[...]

PR #9328, opened 2022-08-26
feature: define custom migration hooks

### LINKED ISSUE #8544: Add custom migration generation hooks
## Feature Description

Allow users to extend migration generation.

### The Problem

We have built custom trigger system, because our database is accessed by multiple services written in languages other than Node. This system is quite error prone.

Potentially, this system allow user to specify:

* Triggers
* [Rules](https://www.postgresql.org/docs/14/sql-createrule.html)
* Stored procedures
* Functions

### The Solution

Allow to specify option `queryBuilderHooks` in `ormconfig` and connection hooks with type `{queryBuilderHooks: (typeof QueryBuilderHook)[]`.

Following interfaces:

```ts
abstract class QueryBuilderHook {
   init?(queryRunner: QueryRunner, schemaBuilder: SchemaBuilder): Promise<void>;
   afterAll?(queryRunner: QueryRunner, schemaBuilder: SchemaBuilder): Promise<SqlInMemory>;
   beforeAll?(queryRunner: QueryRunner, schemaBuilder: SchemaBuilder): Promise<SqlInMemory>;
}
```

#### Possible enhacements

Let's keep it simple for a moment.

* All `RdbmsSchemaBuilder` methods can have `before${method}` and `after${method}`
* Hypothetical package `@typeorm/postgres` can have support for Postgres-specific extensions, like rules. 

### Considered Alternatives

We already have trigger system.

| `aurora-data-api`    | ✅            |
| `aurora-data-api-pg` | ✅            |
| `better-sqlite3`     | ✅            |
| `cockroachdb`        | ✅            |
| `cordova`            | ✅            |
| `expo`               | ✅            |
| `mongodb`            | ❓            |
| `mysql`              | ✅            |
| `nativescript`       | ✅            |
| `oracle`             | ✅            |
| `postgres`           | ✅            |
| `react-native`       | ✅            |
| `sap`                | ❓            |
| `sqlite`             | ✅            |
| `sqlite-abstract`    | ✅            |
| `sqljs`              | ✅            |
| `sqlserver`          | ✅            |

Basically, I need someone to give green light for that.

Ginden:
Resolves  #8544

Allows to resolve in user-land heavily requested (68 "likes") #1082

I propose to create packages `@typeorm/postgres` and similar to maintain database-specific extensions, but this can be done _by users_ (I may be able to support `@typeorm/postgres`, but not eg. `@typeorm/mysql`).

SQL standard [doesn't go into details of defining triggers](https://dba.stackexchange.com/a/222687/53487) - therefore it's basically impossible to define "one size fits all" solution for all databases.

TODO: polish test, add documentation.

### Description of change

Allow users to write their own migration hooks

### Pull-Request Checklist


Ginden:
Seems like my first trope with `SqlInMemory` was right, it needs more changes.

src/driver/spanner/SpannerQueryRunner.ts
```diff
@@ -1509,6 +1509,32 @@ export class SpannerQueryRunner extends BaseQueryRunner implements QueryRunner {
...
     }
 
+    /**
+     * Executes sql used special for schema build.
+     */
+    async executeSchemaBuilderQueries(
+        upQueries: Query | Query[],
+        downQueries: Query | Query[],
+    ): Promise<void> {
+        if (upQueries instanceof Query) upQueries = [upQueries]
```
pleerock, line 1519: inverse the condition to Array.isArray instead. I don't trust `instanceof`.

pleerock: (I see it was before, but anyway you can do that if you want)

pleerock:
I'm looking to see more documentation and real-world usage examples of different hooks.

fknop:
Hello! 
As a user of TypeORM working on a relatively big project using postgres with many custom database objects such as sequences, functions and triggers, I'm exited for this kind of feature as managing all of this using custom manual migrations is a real pain!
However, even if the migrations hooks approach is a great start, there are a couple points that I would like to raise:

* How I see it: e.g. the `@typeorm/postgres` extension would expose decorators such as `Sequence`, `Function`, `Trigger`, etc. The extension would also expose a `PostgresMigrationsHook` which would take all of this metadata and create database objects accordingly.
For example:
```typescript
@Sequence({start: 0, maxValue: 10000, cycle: true, name: 'my_sequence'})
class MySequence {}
```
* TypeORM itself could expose a new array inside `MetadataArgsStorage` for extensions to push custom metadata into it. The database object would then be registered inside the config like `views` or `entities`. However, if TypeORM does not want to maintain this small layer of abstraction, I guess the extension could register its own metadata storage to avoid complexity in TypeORM itself.
* We would probably need easy access to the TypeORM metadata table API to store SQL definitions of custom database objects for change detection
* I think the `customProperties` approach could just be replaced by custom decorator exposed by the extension, I don't see why only entities would benefit from this and adding it to every database objects supported by TypeORM would add complexity to the library imo.
* The migrations hook would probably need access to the entire set of metadata and not just entities.
* Do we need the ability to define dependencies between custom database objects and existing entities/views? In Postgres case,  I don't think it is needed because sequences/functions can be created before entities and triggers after entities. I don't have every database use case in mind though.

I would love to try and implementing a POC for a `postgres` extension with custom decorators using our real-word application as example (if I can find some time!)

fknop:
> I'm looking to see more documentation and real-world usage examples of different hooks.

Hello @pleerock !
I've implemented a sample repository using this rebased branch as base.

I implemented a sequence extension that you can find here: https://github.com/fknop/typeorm-postgres/blob/main/packages/typeorm-postgres/src/lib/extensions/PostgresSequenceExtension.ts
You can find usage in tests here: https://github.com/fknop/typeorm-postgres/blob/main/packages/typeorm-postgres/src/lib/testing/sequence/PostgresSequenceExtension.spec.ts

I'm still relying on some non-exposed methods in `master` that I made public in my fork as well a nested import to import necessary classes.

I'm planning on creating an example to manage PG functions and triggers in the same way.

Would love to hear your feedback :)

fknop:
> I'm looking to see more documentation and real-world usage examples of different hooks.

Hello! Did you have any time to look at the examples I provided above? :) 

We're starting to manage a lot more triggers/functions in our codebase and having a way to sync them during development/tests would be great

gioboa:
Hi @Ginden is this PR still valid os is it abandoned?

Ginden:
@gioboa I no longer work at company using TypeORM, so I have no time to continue on that. Obviously, I hope someone will take this, because it's nice functionality.

gioboa:
I see thanks for your update.
So I close it, this will remain a good starting point. 👍

gioboa (commented):
I see thanks for your update.
So I close it, this will remain a good starting point. 👍

PR #9339, opened 2022-09-01
feat: implement new virtual decorator

### LINKED ISSUE #9323: VirtualColumn Decorator
## Feature Description
I would like to add a new feature to TypeORM so that I can consolidate my code and simply run a subSelect with every get request on the requested entity. This would improve performance and allow me to utilize the full potential of the supported relational databases PostgresQL, MySql, and MSSQL.

Below is what I am planning on implementing and tying to this Issue once it is completed (I will be coding this up tonight and think that it would bring value to this project overall)

### The Problem
When querying data against any relational database there is often a need to return some query (a sub-select, count, etc.) that returns with a model. For example, if we create an entity of type Company that has a one-to-many relationship on employees, it may make sense to have a totalEmployeesCount column ("Virtual Column") that calculates this on every request *or even better when the column is selected in the base entity query*. This is what I am wanting to implement in TypeORM. This is common in most ORM solutions and I would love to see it here.

### The Solution
Let's use the same example above: we have entity of type Company that has a one-to-many relationship on an employee entity (employees), it may make sense to have a totalEmployeesCount column that could be re-calculated on every fetch of the entity:

```js
import { BaseEntity, PrimaryColumn, OneToMany, ManyToOne, VirtualColumn } from "typeorm";

@Entity({ name: "companies" })
export class Company extends BaseEntity {
  @PrimaryColumn("varchar", { length: 50 })
  name: string;

  @VirtualColumn({ query: (alias) => `SELECT COUNT("name") FROM "employees" WHERE "companyName" = ${alias}.name` })
  totalEmployeesCount: number;

  @OneToMany((type) => Employee, (employee) => employee.company)
  employees: Employee[];
}

@Entity({ name: "employees" })
  @PrimaryColumn("varchar", { length: 50 })
  name: string;

  @ManyToOne((type) => Company, (company) => company.employees)
  company: Company;
}
```

__Some notes on the VirtualColumn Decorator:__
1. The "query" property will be used to populate the column data. This query is used when generating the relational db script. The query function is called with the current entities alias automatically generated by TypeORM.

2. A virtual column is not an editable column. It is a record that should be considered as readonly and it is ignored when the save entity function is run.

### Considered Alternatives
One workaround would be to create a query builder but in doing so, this would add complexity, and it doesn't solve the overall need: code redundancy, and consistent entity result sets.

The second workaround would to be to add an @AfterLoad() decorator and perform a query there, but this is a poor solution since it would not allow us to utilize the relational databases query optimization engine when performing a query leading to worse performance. This proposed solution would be faster and avoid several unnecessary database hits.

This solution would work for all relational databases (it is possible to add non-relational databases to this as well but the immediate need is for relational databases.

| DB Type           | Relevant      |
| `aurora-mysql`    | yes            |
| `aurora-postgres` | yes            |
| `better-sqlite3`  | yes            |
| `sqlite`          | yes            |
| `sqlite-abstract` | yes            |
| `sqljs`           | yes            |
| `sqlserver`       | yes            |

 - ✅ Yes, I have the time, and I know how to start. (**Already working on a solution**)

CollinCashio:
### Description of change
Add a new feature that Fixes #9323 

### Pull-Request Checklist


package.json
```diff
@@ -237,6 +237,7 @@
   "scripts": {
     "test": "rimraf ./build && tsc && mocha --file ./build/compiled/test/utils/test-setup.js --bail --recursive --timeout 60000 ./build/compiled/test",
     "test-fast": "mocha --file ./build/compiled/test/utils/test-setup.js --bail --recursive --timeout 60000 ./build/compiled/test",
+    "test-github-issue": "mocha --file ./build/compiled/test/utils/test-setup.js --bail --recursive --timeout 60000 ./build/compiled/test/github-issues/9323",
     "compile": "rimraf ./build && tsc",
```
Ginden, line 241: Shouldn't be merged - you can use `--grep` option of `mocha` for you development.

src/driver/aurora-mysql/AuroraMysqlDriver.ts
```diff
@@ -633,6 +633,9 @@ export class AuroraMysqlDriver implements Driver {
         ) {
             // convert to number if that exists in possible enum options
             value = parseInt(value)
+        } else if (columnMetadata.type === Number) {
+            // convert to number if number
+            value = !isNaN(+value) ? parseInt(value) : value
```
Ginden, line 638: I'm not sure why this change is in this PR?

Ginden:
"Calculated column" is term associated with PowerBI. "Computed column" refers to specific feature in multiple databases. Maybe call it "virtual column"?

Generally, good job!

src/decorator/options/CalculatedColumnOptions.ts
```diff
@@ -0,0 +1,31 @@
...
+    /**
+     * Return type of HSTORE column.
+     * Returns value as string or as object.
+     */
+    hstoreType?: "object" | "string"
+
+    /**
+     * Query to be used to populate the column data. This query is used when generating the relational db script.
+     * The query function is called with the current entities alias either defined by the Entity Decorator or automatically
+     * @See https://github.com/typeorm/typeorm/issues/9323 for more details.
```
Ginden, line 22: Should be rather a link to docs.

src/driver/aurora-mysql/AuroraMysqlDriver.ts
CollinCashio: This change was made in all of the rdb drivers because it was not casting the data types automatically based on my models type definition (specifically for numbers) which hindered my development for this feature. If you are specifically asking about this driver, I can remove it for this one but did update all rdb drivers.

package.json
CollinCashio: Awesome - you are right! I will update this today. Thanks for the reply!

CollinCashio:
> "Calculated column" is term associated with PowerBI. "Computed column" refers to specific feature in multiple databases. Maybe call it "virtual column"?
> 
> Generally, good job!

Thank you. I really think this feature will add value to the community.

If you look at the history of the issue that I posted I originally called the calculated column decorator a "VirtualColumn" decorator because I believe that this is a better name myself, but the entity column Metadata already had a property being used for IsVirtual, so I figured I should go with a new name. I can for sure update this to VirtualColumn especially since like you said it is a better name, and I can add a new property called isVirtualDecorator to the entity column Metadata

src/decorator/options/CalculatedColumnOptions.ts
CollinCashio: I saw another issue did it this way so I replicated theirs but I agree that a link to the docs would be better for this. I will update this as well today. Thanks for all your feedback

package.json
CollinCashio: This has been updated @Ginden

src/decorator/options/CalculatedColumnOptions.ts
CollinCashio: This has been updated @Ginden

docs/decorator-reference.md
```diff
@@ -374,6 +375,36 @@ export class User {
...
+export class Employee extends BaseEntity {
+  @PrimaryColumn("varchar", { length: 50 })
+  name: string;
+
+  @ManyToOne((type) => Company, (company) => company.employees)
+  company: Company;
+}
+```
+
+Value will be generated only once, before inserting the entity into the database.
```
Ginden, line 406: Is it generated or it's read-only additional property returned by `find` methods?

docs/decorator-reference.md
```diff
@@ -374,6 +375,36 @@ export class User {
...
+Each time you call `find` or `findOne` from the entity manager, the value is recalculated based on the query function that was provided in the VirtualColumn Decorator. The alias argument passed to the query references the exact entity alias of the generated query behind the scenes.
+
+```typescript
+@Entity({ name: "companies", alias: "COMP" })
+export class Company extends BaseEntity {
+  @PrimaryColumn("varchar", { length: 50 })
+  name: string;
+
+  @VirtualColumn({ query: (alias) => `SELECT COUNT("name") FROM "employees" WHERE "companyName" = ${alias}.name` })
+  totalEmployeesCount: number;
```
Ginden, line 390: BTW, are users allowed to:

`.getRepository(Company).find({where: {totalEmployeesCount: Equal(5)})`?

docs/decorator-reference.md
CollinCashio: Whoops - good catch - this should not be in here. This is not inserting into the database. This was a copy/paste mistake from the decorator above.

docs/decorator-reference.md
CollinCashio: Yes, users are allowed to do this. I made sure that you could perform where & order by's with the entity manager. You can view my test case here: 

I did however run into a "limitation" (limitation but there is a workaround) with "Wheres" and "Order Bys" with the query builder. They would have to make a generic function that returns the virtual column as a query string and export that function so they could generate the query within the query builder.

CollinCashio: Specifically this:

```js
                 where: {
                        name: companyName,
                        totalEmployeesCount: MoreThan(2),
                        employees: {
                            timesheets: {
                                totalActvityHours: MoreThan(0),
                            },
                        },
                    },
                    order: {
                        employees: {
                            timesheets: {
                                id: "DESC",
                                totalActvityHours: "ASC",
                            },
                        },
                    },
```

docs/decorator-reference.md
CollinCashio: This has been fixed @Ginden

CollinCashio:
How are we looking on this @Ginden ? It has been a week since our last discussion, and I want to ensure that this PR doesn't get lost in the backlog. Can this get merged in or is there something outstanding I need to do on my end?

Ginden:
I'm merely helping with issues and reviews, to get them streamlined faster. At the end of day, decision, especially design decisions, belongs to @pleerock and other higher level maintainers.

bjoveski:
thank you for doing this, this would be an extremely useful feature for us!

H6LS1S:
Сan this be released as a standalone release if it doesn't make it into the main release soon?

src/driver/aurora-mysql/AuroraMysqlQueryRunner.ts
```diff
@@ -2411,14 +2411,24 @@ export class AuroraMysqlQueryRunner
      * Builds create table sql
      */
     protected createTableSql(table: Table, createForeignKeys?: boolean): Query {
-        const columnDefinitions = table.columns
+        const tableMetaData = this.connection.entityMetadatas.find(
```
pleerock, line 2414: we should not rely on entityMetadatas in this method. Please send the `Table` instance without virtual columns into this method.

src/driver/cockroachdb/CockroachQueryRunner.ts
```diff
@@ -3127,12 +3127,22 @@ export class CockroachQueryRunner
      * Builds create table sql.
      */
     protected createTableSql(table: Table, createForeignKeys?: boolean): Query {
-        const columnDefinitions = table.columns
+        const tableMetaData = this.connection.entityMetadatas.find(
```
pleerock, line 3130: same comment in all QueryRunners

src/driver/aurora-mysql/AuroraMysqlDriver.ts
pleerock, line 638: can you please explain how this change related to the proposed functionality.

src/metadata-args/types/ColumnMode.ts
```diff
@@ -7,6 +7,7 @@
 export type ColumnMode =
     | "regular"
     | "virtual"
+    | "calculated"
```
pleerock, line 10: let's go with "virtual-property" instead of "calculated" for now.

src/metadata/ColumnMetadata.ts
```diff
@@ -238,6 +239,18 @@ export class ColumnMetadata {
      */
     isVirtual: boolean = false
 
+    /**
+     * Indicates if column is calculated. Virtual columns are not mapped to the entity.
+     */
+    isVirtualDecorator: boolean = false
```
pleerock, line 245: "virtual decorator" isn't the best name actually... first there is a confusion between "calculated" and "virtual decorator" now in the naming / description. second virtuals can be defined in entity schemas as well (alternative to decorators). I do understand we already have "isVirtual" here which used for different purpose... that's why I suggest to go with "isVirtualProperty" instead.

src/schema-builder/RdbmsSchemaBuilder.ts
```diff
@@ -717,9 +721,13 @@ export class RdbmsSchemaBuilder implements SchemaBuilder {
             // find which columns are new
             const newColumnMetadatas = metadata.columns.filter(
                 (columnMetadata) => {
-                    return !table.columns.find(
-                        (tableColumn) =>
-                            tableColumn.name === columnMetadata.databaseName,
+                    return (
+                        !columnMetadata.isVirtualDecorator &&
+                        !table.columns.find(
```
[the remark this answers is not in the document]
pleerock: I mean do not put virtual columns into "loadedTables"

src/util/DateUtils.ts
```diff
@@ -26,14 +26,9 @@ export class DateUtils {
...
-        if (value instanceof Date)
-            return (
-                this.formatZerolessValue(value.getFullYear(), 4) +
-                "-" +
-                this.formatZerolessValue(value.getMonth() + 1) +
-                "-" +
-                this.formatZerolessValue(value.getDate())
-            )
+        if (value instanceof Date) {
+            return value.toISOString().split("T")[0]
```
pleerock, line 30: are you sure it's the same? 😨

src/driver/aurora-mysql/AuroraMysqlDriver.ts
CollinCashio: This change was made in all of the rdb drivers because typeorm was not casting the data types automatically based on my models type definition (specifically for numbers) which hindered my development for this feature.

IE: When I pulled the data from the DB, I was expecting my data type to be a number because that is what I defined in my virtual column decorator's metadata but the returned result was a string.

If you are specifically asking about this driver, I can remove it for this one but did update all rdb drivers.

src/metadata-args/types/ColumnMode.ts
CollinCashio: Sounds good. I will update this in a few hours.

src/metadata/ColumnMetadata.ts
CollinCashio: Sounds good. I will update this in a few hours.

src/schema-builder/RdbmsSchemaBuilder.ts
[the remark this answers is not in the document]
CollinCashio: Are you suggesting a getter method on the columns that filters out the virtual columns (if the table data is loaded into the entity?), if so, I do like that solution better than managing it this was for sure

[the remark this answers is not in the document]
pleerock: I think the question is: how do we get virtual columns populated into `loadedTables`? It looks likes they aren't, because content of the loadedTables is built from the database columns, not entity metadata columns. If I don't miss anything.

src/driver/aurora-mysql/AuroraMysqlQueryRunner.ts
CollinCashio: I have updated this with the changes that you suggested. I was not aware that the table object was being used to be a direct reflection of the table in the database (it makes sense and it cleaned up those checks). Thanks @pleerock !

src/driver/cockroachdb/CockroachQueryRunner.ts
CollinCashio: Please see above comment

src/metadata-args/types/ColumnMode.ts
CollinCashio: This has been updated

src/metadata/ColumnMetadata.ts
CollinCashio: This has been updated

src/schema-builder/RdbmsSchemaBuilder.ts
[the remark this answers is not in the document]
CollinCashio: Same comment as above:

I have updated this with the changes that you suggested. I was not aware that the table object was being used to be a direct reflection of the table in the database (it makes sense and it cleaned up those checks). Thanks @pleerock !

src/util/DateUtils.ts
CollinCashio: It is not the same, my change puts the time in UTC whereas the the before change puts the timezone in whatever timezone the server is on (which is why a test case was failing on my machine locally). I updated this because I assumed that the test case was golden but after reading through more of the source code from this comment, I realized that it may be purposed to format the data into the current time zone. I think that the test case that failed locally should be updated rather than this code. I have reverted this and am leaving that update to the test case outside of the scope of this feature.

pleerock:
Thank you a lot for this contribution! Quite a solid implementation!

joeruello:
Was looking for something like this today and found it hassn't been released yet :(. Any ETA on when the next release will be cut with this PR included?

kerhong:
This PR introduced a very breaking change about default behavior for number columns without specific type added.

```ts
class Foo {
    @Column()
    bar!: number;
}
```

This is now always pushed through `parseInt` (https://github.com/typeorm/typeorm/pull/9339/files#diff-3346a7798b4db06d6debd87138bb8322626b91c1c325b730d51ee3afc1b4c2bbR798-R800) and is therefore an integer, previously it worked fine with a floating number.

Was this change intended? Seems completely related to the original function of this PR.

pleerock:
I'm not sure why these changes were needed, let's ask @CollinCashio 

But in what cases previously it worked? `@Column() bar: number` always maps to `int`, if you want to use floating numbers, you must specify a type, e.g. `@Column("double") bar: number`

kerhong:
@pleerock Than's for your quick response. Here is the use where I found this issue.

1. typeorm@0.3.10
2. Migration with
    ```ts
    new TableColumn({
      name: 'foo',
      type: 'float',
    }),
    ```
3. Entity with
    ```ts
    @Column()
    foo!: number;
    ```

This would not round/floor the value, entity would contain the full float on fetch.

I am not sure to what extent should `@Column()` contain `TableColumnOptions`, it can be omitted, and all fields are optional.
Is there some documentation that explains how strictly the column details must be preserved between migrations and entities? Is there a way to validate entities against schema and/or migrations?

pleerock:
I would avoid complicating the report with migrations. I'm not sure where did you get the migration you wrote - if it's manually written, then it's not valid since it differs from the entity definition. 

The best if you always specify column type in `@Column` decorator to be explicit. By default, again `number` type without `@Column` type specification maps to `int`. So in your case for your migration the valid column definition must be `@Column("float") foo!: number;`.

src/driver/aurora-mysql/AuroraMysqlDriver.ts
CollinCashio: @pleerock @kerhong - The above is my explanation as to why this was needed.

If number is specified without the column type, we assume "int" as Number unless specified in the @Column() decorator - if this is not the case and it should be assumed as a float, we could parseFloat() here but the specification @pleerock mentioned below is that it should be assumed as an integer

kerhong:
@pleerock Thanks for the clarification, I wasn't aware that `number` is treated as `int`.

docs/decorator-reference.md
```diff
@@ -374,6 +375,34 @@ export class User {
...
+Special column that is never saved to the database and thus acts as a readonly property.
+Each time you call `find` or `findOne` from the entity manager, the value is recalculated based on the query function that was provided in the VirtualColumn Decorator. The alias argument passed to the query references the exact entity alias of the generated query behind the scenes.
+
+```typescript
+@Entity({ name: "companies", alias: "COMP" })
+export class Company extends BaseEntity {
+  @PrimaryColumn("varchar", { length: 50 })
+  name: string;
+
+  @VirtualColumn({ query: (alias) => `SELECT COUNT("name") FROM "employees" WHERE "companyName" = ${alias}.name` })
```
Abobos, line 389: I have a question. Can't the `query` be optional?.. we don't need to pass it.. what if we have a big query that returns data from a db with the entity declared column together with the virtual column. How would that work?

Abobos: @CollinCashio

sschneider-ihre-pvs: what if the virtualcolumn is not in the database at all but some other computed value?

sschneider-ihre-pvs: background would be graphql since if you want a unique id field and have a composite key for a table you want to cramp that into a single key maybe

CollinCashio: Hey @Abobos and @sschneider-ihre-pvs ,

This functionality was designed specifically for subqueries. This solves the need of allowing your database engine to return queried result sets and put the data into your defined model (also allows you to use the DB engine to filter data from the model).

If you are just looking for a readonly property that has no need to pull database information and possibly pull from a third party API or some other data source, you could just use a regular property without any column decorator. IE your model could look like:

```ts
@Entity({ name: "timesheets" })
export default class TimeSheet extends BaseEntity {
    @PrimaryGeneratedColumn()
    id: number;

    hours: number;

    @AfterLoad()
    afterLoad() {
         // perform calculations here to set property of hours
    }
}
```

Abobos: @CollinCashio VIrtual Column suffices for my use-case.. but I was thinking you can just add the column name as some alias in a query, without having to do a subquery.

Thanks

effective-giggle:
This PR created bug #10026

src/driver/aurora-mysql/AuroraMysqlDriver.ts
effective-giggle: Why did was parseInt chosen? this removes all decimals.

DecayConstant:
@PrimaryGeneratedColumn() id: number

Bala2105 (commented):
Reviewed PR #9339, and it’s a major architectural addition that introduces the @VirtualColumn decorator. This feature allows properties to be populated by complex subqueries while keeping the database schema clean, as these columns are never actually persisted to a table. The implementation is impressive because it doesn't just add a decorator; it modifies the SelectQueryBuilder to dynamically inject these subqueries into SELECT statements and updates the RdbmsSchemaBuilder to ensure the synchronization process ignores these virtual fields, preventing accidental "column missing" errors.

bjoveski:
note that @Bala2105's account was created 6 hours ago. They have reviewed bunch of various PRs across multiple OSS repos - and they all look like AI slop.

Flagging in case the review is not valid.

Update: my bad, this PR was merged back in 2022 - this review didn't do anything

Issue #5637, opened 2020-03-07
leftJoinAndMapOne does not add target property to entity


FFsim:
**Issue type:**

[ x] question
[ ] bug report
[ ] feature request
[ ] documentation issue

**Database system/driver:**

[ ] `cordova`
[ ] `mongodb`
[ ] `mssql`
[x] `mysql` / `mariadb`
[ ] `oracle`
[ ] `postgres`
[ ] `cockroachdb`
[ ] `sqlite`
[ ] `sqljs`
[ ] `react-native`
[ ] `expo`

**TypeORM version:**

[x] `latest`
[ ] `@next`
[ ] `0.x.x` (or put your version here)

**Steps to reproduce or a small repository showing the problem:**

<!--
    To answer those questions you need to put "x" inside the square brackets, for example:
    [x] `mysql`
    [ ] `postgres`

    Also, please format your code properly (by taking code blocks into ```ts .... ```)
!>

The above will return the User, but without the lastPurchase property

```const users = await this.repository.createQueryBuilder("user")
      .leftJoinAndMapOne("user.lastPurchase", (subQuery) => {
        return subQuery
          .select()
          .from(Purchase, "purchase")
          .orderBy("created", "DESC");
      }, "purchase", "purchase.userId = user.id")
      .leftJoinAndSelect("user.merchant", "merchant")
      .where("user.role IN (:userRoles)", {userRoles: roles})
      .getManyAndCount();
```

any idea?

vishal38-isharani:
I'm also facing the same issue only difference is I'm using Postgres.

```
return this.getRepo().createQueryBuilder('job-card')
       .leftJoinAndMapOne('job-card.workflowState',
        qb =>
          qb
            .from(WorkflowStateEntity, 'w')
            .orderBy({ 'w.createdAt': 'DESC' })
            .limit(1),
        'workflow',
        "workflow.job_card_id = job-card.id",
      )
      .where("job-card.id = :id", { id })
      .getOne()
```
please help

imnotjames:
If you use `getSql` what's the generated query?

https://github.com/typeorm/typeorm/blob/master/docs/select-query-builder.md#getting-the-generated-query

Also, can you create a minimal reproduction of the issue?  Someone we can use in the creation of a test - the entities, the code you used to exhibit the behavior, and if possible some assertions we could use to verify the behavior is broken.

We use chai for testing https://www.chaijs.com/api/bdd/

vishal38-isharani:
[...]

PR #9354, opened 2022-09-07
leftJoinAndMapOne and innerJoinAndMapOne now map correctly with query builder

### LINKED ISSUE #8705: `leftJoinAndMapOne` doesn't add property when custom query builder is used
## Issue Description

Coming from #1882. Running a query builder in `leftJoinAndMapOne` doesn't map to property.

### Expected Behavior

`leftJoinAndMapOne` should map to the provided property.

### Actual Behavior

I am trying to do something like

```ts
const [venues, venuesCount] = this.venuesRepository
  .createQueryBuilder("Venue")
  .leftJoinAndSelect("Venue.city", "city")
  .leftJoinAndMapOne(
    "Venue.eventStats",
    (qb) =>
      qb
        .from("events", "Event")
        .select('"venueId"')
        .addSelect(`MAX("capacity")`, "maxVenueCapacity")
        .addSelect(`COUNT("venueId")`, "eventsCount")
        .groupBy('"venueId"'),
    "Event",
    "Venue.id = Event.venueId"
  )
  .where(`Venue.cityId = :cityId`, { cityId: "someCityId" })
  .orderBy(`eventCount`, `DESC`)
  .getManyAndCount();
```

And my model has

```ts
interface EventStats {
  maxVenueCapacity: number;
  eventsCount: number;
}

@Entity()
class Venue {
  eventStats: EventStats;
}
```

at the end `venues[].eventStats` is not added.

```
// include the output in code tags like these!
```

### Steps to Reproduce

1. Use the code I provided above to create a similar query.
2. Run the query and check if the mapped property is getting added.

### My Environment

| Operating System    |    Docker Node Alpine Image     |
| Node.js version     | 14.17.0  | 
| Typescript version  | 4.5.4  | 

acuthbert:
### Description of change

When joining to a query builder instead of an entity or table name, typeorm now correctly supports mapping the result to an entity. This introduces a new argument to the join functions to supply the join attributes with a source of meta data for the mapping

For example:
```ts
const loadedPost = await connection.manager
      .innerJoinAndMapOne(
           "post.tag",
           qb => qb.from(Tag, "tag"),
           "tag",
           "tag.id = post.tagId",
           undefined,
           // The next argument is new - it helps typeorm know which entity
           // to use to complete the mapping for a query builder.
           Tag
      )
      .where("post.id = :id", { id: post.id })
      .getOne()
```

Fixes #8705, #5637

### Pull-Request Checklist


pleerock:
I'm usually against having additional changes into not documented `leftJoinAndMapOne` feature, but in this case it looks like change is simple and doesn't harm anything. We can merge it.

acuthbert:
We've had some internal conversations about this today and wanted to float a second solution.

Should the meta data for the mapping be inferred from the relation you're trying to map into on occassions where none is available from the source (e.g. query builder).

For example:

```ts
const loadedPost = await connection.manager
      .innerJoinAndMapOne(
           "post.tag",
           qb => qb.from(Tag, "tag"),
           "tag",
           "tag.id = post.tagId"
      )
      .where("post.id = :id", { id: post.id })
      .getOne()
```

Here typeorm can't map the reuslt of the sub query because it can't get a handle on the correctly entity type. Fair enough - but should it just infer because we're trying to map to "post.tag" that it is a TagEntity? In otherwords - use the relation itself as the guide?

acuthbert:
This would also get around the need to pass over the second last optional argument with 'undefined' :joy:

pleerock:
try to implement it :)

cjsauer:
I'm not understanding the point of this PR. I'm having the same issue described in https://github.com/typeorm/typeorm/issues/5637, in that `innerJoinAndMapOne` with a sub-query does not correctly populate the specified property. I found this PR linked in that issue as a "potential solution", but when I try using the `mapAsEntity` option it just completely ignores all of my subquery logic...the test cases that shipped with this PR are far too trivial, and I don't believe that this "fix" actually achieved anything...

DiFuks:
@cjsauer  I have the same problem

acuthbert:
@cjsauer I just came across your comment today. Can you have a look at the [Files Changed ](https://github.com/typeorm/typeorm/pull/9354/files) on this MR as I can't see how the added code could cause "completely ignored subquery logic". All it does is provide another option for returning meta data. At worst the new if statement would be 'false' and the original value of undefined would be returned.

The tests I wrote were failing before the fix and passing after it, so the fix is definitely doing what I needed at the time.

xdc0:
This patch is wrong.

The last argument `mapAsEntity` will cause this line to evaluate to some `EntityMetadata` and enter the `if` block: https://github.com/typeorm/typeorm/blob/46249303be03adea74266837b8e6eb49227f476e/src/query-builder/SelectQueryBuilder.ts#L2038-L2039

That `if` block will ignore the `subQuery` defined when calling `leftJoinAndMap` and instead do a join through metadata, that's why the query in @cjsauer example above is dropped and "reverts" to use an auto-generated query via metadata.

Issue #9006, opened 2022-05-15
Eager relations do not respect relationLoadStrategy


serg06:
## Issue Description

When you specify `relationLoadStrategy: 'query'` in DataSource:

```ts
export const dataSource = new DataSource({
    relationLoadStrategy: 'query',
    ...
});
```

And then specify `eager: true` in an Entity relation:

```ts
@Entity()
export class UserEntity {
    @OneToMany(() => ProductEntity, product => product.user, {
        eager: true
    }),
    ...
}
```

And then get one of those entities:

```ts
dataSource.getRepository(UserEntity).find({
    take: 1
});
```

It doesn't respect `relationLoadStrategy`.

### Expected Behavior

The relation is loaded using multiple queries instead of a `join`.

With `logging: true`, you see a two separate queries and no `LEFT JOIN`.

### Actual Behavior

The relation is loaded using a `join`.

With `logging: true`, you see a single query with a `LEFT JOIN` on `"UserEntity_products"."userId"="UserEntity"."id"`.

### Steps to Reproduce

See Issue Description

### My Environment

| Dependency          | Version  |
| ---                 | ---      |
| TypeORM version     | 0.3.6  | 

### Relevant Database Driver(s)

| DB Type           | Reproducible  |
|-------------------| ---           |
| `aurora-mysql`    | no            |
| `aurora-postgres` | no            |
| `better-sqlite3`  | no            |
| `cockroachdb`     | no            |
| `cordova`         | no            |
| `expo`            | no            |
| `mongodb`         | no            |
| `mysql`           | no            |
| `nativescript`    | no            |
| `oracle`          | no            |
| `postgres`        | yes            |
| `react-native`    | no            |
| `sap`             | no            |
| `spanner`         | no            |
| `sqlite`          | no            |
| `sqlite-abstract` | no            |
| `sqljs`           | no            |
| `sqlserver`       | no            |

### Are you willing to resolve this issue by submitting a Pull Request?

 - ✖️ Yes, I have the time, and I know how to start.
 - ✖️ Yes, I have the time, but I don't know how to start. I would need guidance.
 - ✖️ No, I don’t have the time, but I can support (using donations) development.
 - ✅ No, I don’t have the time and I’m okay to wait for the community / maintainers to resolve this issue.

komo81:
+1

PR #9358, opened 2022-09-08
fix: disable loading of eager relations recursively

maierj:
Closes: #9006

### Description of change

A check was added to all relevant methods in `RelationLoader` which ensures that nested eager relations are only loaded, if the loading of eager relations was not disabled explicitly in the findOptions. That way, the option `loadEagerRelations` applies recursively and not only to the first hierarchy level of relations.

Issue #9006 is fixed by adding the same check in the `applyFindOptions` method of the `SelectQueryBuilder` before calling `FindOptionsUtils.joinEagerRelations`. In order to load the eager relations using the `query` strategy, the eager relations are added to the `relations` object of `FindOptions`.

### Pull-Request Checklist


maierj:
Closing PR for the moment, because the solution has a few flaws.

PR #9365, opened 2022-09-09
feature: upsert options conflictPaths as Entity keys

redwert:
### Description of change

Add checks that conflictPaths in upsert options relate to Entity

### Pull-Request Checklist


redwert:
`
@Entity()
export class Post {
    @PrimaryGeneratedColumn()
    id: number

    @Column()
    title: string
}
`

image

pleerock:
I understand what you are trying to resolve, but I would avoid merging it because:

* there are other methods / options that are not keyof Entity and we need to design everything carefully and apply changes everywhere
* I don't know how currently users are using conflictPaths - do they use it with non-entity columns? If they do, this change is considered as breaking change.

redwert:
reasonable, and what about adding support for the following declaration:
`awai tsomeRepository.upsert(
   {tite;"the title"},
   {conflictPaths: {title: true}}
)`
Add the ability to declare the conflict paths like we declare the relations in find.

pleerock:
Yes, if we would have this feature, it definitely needs to be as `{conflictPaths: {title: true}}` because we need to support columns inside embeds, e.g. `{ {conflictPaths: { nested: { title : true }}}`. But again, this is part of the bigger breaking changes.

redwert:
Please take a look on these changes

pleerock:
I can accept it since it looks like doesn't break anything. Keep in mind embeds in entities are not supported by your changes. Later on we'll need to add this feature as well. Please add a test case for your change.

redwert:
I have added the test

redwert:
@pleerock could you please take a look. I have added the test and probably you can merge it

pleerock:
Thank you for contribution! 🎉

frostbytedata:
Yea this is a bad change for those of us using embedded entities. It is now no longer possible to specify conflict entity keys in the array for `conflictPaths` . Here is a simple example below:

```ts

class User {
  @Column()
  @IsEmail()
  email: string;

  @Column(() => Profile, {
    prefix: false,
  })
  profile: Profile;
}

class Profile {
  @Column()
  @IsString()
  name: string;

  @Column({
    name: 'profile_avatar',
    nullable: true,
  })
  @IsString()
  avatar: string;
}
```

In this example, it is impossible (unless I am using it wrong) to specify `conflictPath: ['profile_avatar']` for `upsert` commands. Is there a way around this for now?

Issue #9379, opened 2022-09-16
RelationIdLoader is not respecting `maxAliasLength`


janzipek:
## Issue Description

`RelationIdLoader` class is doing selects, but it's not using `DriverUtils.buildAlias` so the selects can end up being truncated and all related behaviour is broken. This for example breaks `relationLoadStrategy: 'query'` when the relation key ends up being too long.

### Expected Behavior

`RelationIdLoader` should use `DriverUtils.buildAlias` when creating column aliases generating column aliases that fit the `maxAliasLength` prop of driver.

### Actual Behavior

`RelationIdLoader` ignores `maxAliasLength`.

### Steps to Reproduce

(in postgresql)

**Schema**
```typescript
@Entity()
export class SuperLongTableName {
    @PrimaryGeneratedColumn()
    id: number
    
    @Column()
    name: string
    
    @OneToMany(() => SuperLongTableNameWhichIsRelatedToOriginalTable, table => table.superLongTableName)
    relatedToOriginal: SuperLongTableNameWhichIsRelatedToOriginalTable[]
}

@Entity()
export class SuperLongTableNameWhichIsRelatedToOriginalTable{
    @PrimaryGeneratedColumn()
    id: number

    @Column()
    superLongTableNameId: number;
    
    @ManyToOne(() => SuperLongTableName, table => table.relatedToOriginal)
    @JoinColumn({ name: 'superLongTableNameId' })
    superLongTableName: SuperLongTableName
}
```

```sql
INSERT INTO "SuperLongTableName" ("name") VALUES ('test');
INSERT INTO "SuperLongTableNameWhichIsRelatedToOriginal" ("superLongTableNameId") VALUES (1);
```

```typescript
const data = await appDataSource.getRepository(SuperLongTableName).findOne({
  where: { id: 1 },
  relations: { relatedToOriginal: true }
})

// data = { name: 'test', relatedToOriginal: [] }
```

This is because RelationIdLoader will generate column with name: `SuperLongTableNameWhichIsRelatedToOriginal_relatedToOriginal_superLongTableNameId` which is then truncated by database and all the conditions in the code fail because of that.

### Relevant issues

https://github.com/typeorm/typeorm/issues/3118

### My Environment

| Dependency          | Version  |
| ---                 | ---      |
| Operating System    |          |
| Node.js version     | 14.19.1  | 
| Typescript version  | 4.8.2    | 
| TypeORM version     | 0.3.7    | 

### Relevant Database Driver(s)

[...]

PR #9380, opened 2022-09-16
fix: Update RelationIdLoader to use DriverUtils.getAlias

janzipek:
Update RelationIdLoader to use DriverUtils.getAlias to prevent aliases being possibly trimmed by database.

Closes: #9379

### Description of change

Updates `RelationIdLoader` class to use `DriverUtils.getAlias` to generate column aliases. This resolves issues with databases that have hard limit on column aliases such as Postgres.

**I've never contributed to this project so I'm definitely not sure if this fix is correct**, but it did resolve the issue with `relationLoadStrategy: query` I was having. There may be other places that don't use this util? Maybe this breaks some other behaviour that expects the columns to be properly named?

### Pull-Request Checklist


pleerock:
Thank you for contribution! 🎉

PR #9387, opened 2022-09-20
Typo in docs where EntitySubscriberInterface is explained

miriamso:
There is a typo (according to https://github.com/typeorm/typeorm/blob/master/src/subscriber/event/InsertEvent.ts)


pleerock:
we actually must put `dataSource` into the interface and deprecate `connection`.

pleerock:
Closing as there is no activity to improve this PR.

Issue #4075, opened 2019-05-03
cordova driver using unsupported sql transaction statements


jacobg:
**Issue type:**

[ ] question
[X ] bug report
[ ] feature request
[ ] documentation issue

**Database system/driver:**

[X ] `cordova`
[ ] `mongodb`
[ ] `mssql`
[ ] `mysql` / `mariadb`
[ ] `oracle`
[ ] `postgres`
[ ] `cockroachdb`
[ ] `sqlite`
[ ] `sqljs`
[ ] `react-native`
[ ] `expo`

**TypeORM version:**

[X ] `latest`
[ ] `@next`
[ ] `0.x.x` (or put your version here)

**Steps to reproduce or a small repository showing the problem:**

The cordova driver's CordovaQueryRunner, which extends AbstractSqliteQueryRunner, uses sql statements to begin, commit and rollback transactions. But the cordova sqlite plugin documents that such transactions are unsupported:

> User-defined savepoints are not supported and not expected to be compatible with the transaction locking mechanism used by this plugin. In addition, the use of BEGIN/COMMIT/ROLLBACK statements is not supported.
> https://github.com/xpbrew/cordova-sqlite-storage#other-limitations

Rather the plugin documents that the WebSql db.transaction() api should be used. Notably, it may be difficult, if even possible, to use the db.transaction() api in the cordova driver, as the driver semantics rely on asynchronous and undefined transaction boundaries, whereas the transaction api is limited to more synchronous boundaries sort of like a message queue that automatically closes when it's empty. Also there is no support for explicit rollback in the websql api.

I've opened an issue in the plugin's repo to more explicitly document the behavior of transaction statements, particularly if used serially. If they do work serially, then perhaps this driver could enforce that, or the plugin should:
https://github.com/xpbrew/cordova-sqlite-storage/issues/863

I've also opened issues in the plugin repo to create a new api to allow for better transaction support and concurrency:
https://github.com/xpbrew/cordova-sqlite-storage/issues/862
https://github.com/xpbrew/cordova-sqlite-storage/issues/864

Related issue here: #1884

@brodybits

brodybits:
This sounds pretty bad for reasons I already stated in xpbrew/cordova-sqlite-storage#865.

imnotjames:
[...]

PR #9391, opened 2022-09-21
Disable transactionSupport option for CordovaDriver

michaelwolz:
### Description of change

The [cordova-sqlite-storage](https://github.com/litehelpers/Cordova-sqlite-storage) plugin does not support regular transactions at the moment.

See: https://github.com/storesafe/cordova-sqlite-storage#other-limitations
> User-defined savepoints are not supported and not expected to be compatible with the transaction locking mechanism used by this plugin. In addition, the use of BEGIN/COMMIT/ROLLBACK statements is not supported.

This conflicts with the way typeORM is handling save / remove operations via the `EntityPersistExecutor` by default leading to errors when not explicitly disabling the transaction on each method call via the respective option (e.g. `connection.manager.save(ExampleEntity, { transaction: false })`. Prior version 0.2.34 this was not a problem since the transaction was basically just ignored but with fc4133c the `start|commit|rollbackTransaction` call will throw an error to indicate that transactions aren't supported which is totally correct in my opinion but is pretty hard to work with the current behavior of the `EntityPersistExecutor`. This PR will set the `transactionSupport` property of the `CordovaDriver` to `none` and also respect this setting in the `EntityPersistExecutor`. 

Relates to:
- https://github.com/typeorm/ionic-example/issues/44 
- https://github.com/typeorm/typeorm/issues/4075#issuecomment-933715081
- https://github.com/storesafe/cordova-sqlite-storage/issues/865

### Pull-Request Checklist

### DIFF
--- a/docs/supported-platforms.md
+++ b/docs/supported-platforms.md
@@ -56,7 +56,7 @@ In your main html page, you need to include reflect-metadata:
 TypeORM is able to run on Cordova, PhoneGap, Ionic apps using the
 [cordova-sqlite-storage](https://github.com/litehelpers/Cordova-sqlite-storage) plugin
 You have the option to choose between module loaders just like in browser package.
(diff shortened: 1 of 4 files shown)


mortezaPRK (approved):

AlexMesser:
thank you for contribution!

PR #9406, opened 2022-09-24
Fixed sorting of entities with multi-inheritances

### LINKED ISSUE #9405: Incorrect subject sorting with multi-inheritance entities
## Issue Description
When having multiple layers of inherence in the same entity, entities are incorrectly sorted thus causing error upon cascade insertion.

### Expected Behavior
Entities dependence should be correctly sorted allowing cascade actions.

### Actual Behavior
While [`SubjectTopoligicalSorter.sort`](https://github.com/typeorm/typeorm/blob/master/src/persistence/SubjectTopoligicalSorter.ts#L40) checks for `parentEntityMetadata`. Problem lays in `subject.metadata.parentEntityMetadata` which only holds root parent and ignores intermediate one. Thus while trying to save entity with more then one inheritance order gets messed up.
In my example provided bellow:
 `subject.metadata.targetName` = `EmailChanged` 
`subject.metadata.parentEntityMetadata?.targetName` = `Log`
`sortedEntityTarget` = `ChangeLog` 
As you can see above this [this sorter condition](https://github.com/typeorm/typeorm/blob/master/src/persistence/SubjectTopoligicalSorter.ts#L71) will miss `EmailChange` entity.

And then script tries to insert `Change` first which causes error:
```
QueryFailedError: ER_NO_DEFAULT_FOR_FIELD: Field 'logId' doesn't have a default value
```

### Steps to Reproduce

```ts
export class Email {}

@Entity()
@TableInheritance({ column: { type: "varchar", name: "type" } })
export abstract class Log {
	public readonly id: number;
}
@ChildEntity()
export abstract class ChangeLog<T> extends Log {
	@OneToMany(() => Change, change => change.log, { cascade: true })
	public changes: Change<T>[];
}

@ChildEntity()
export class EmailChanged extends ChangeLog<Email> {}

@Entity()
export class Change<T> {
	public readonly id: number;
	@Column("varchar", { nullable: false, length: 255 })
	readonly propertyName: any;
	@Column("json", { nullable: true })
	readonly oldValue?: any;
	@Column("json", { nullable: true })
	readonly newValue?: any;

	@ManyToOne(() => ChangeLog, { cascade: false, nullable: false, onDelete: "CASCADE" })
	public log: ChangeLog<T>;
}

const emailChanged = new EmailChanged();
const change = new Change<Email>();
change.propertyName = "Example";
emailChanged.changes = [change];

await dataSource.getRepository(Log).save(emailChanged); 
```

### My Environment

| Node.js version     | v16.13.0  | 
| TypeORM version     | 0.3.10  | 

Tested on `mysql` driver all tho im pretty sure all other drivers will experience the same issue.

 - ✅ Yes, I have the time, and I know how to start. But I will require additional help with testing and validation of the bugfix.

ZBAGI:
Fixes sorting of entities with complex multi level inheritances by traversing `inheritanceTree` instead of relying only on the root parent.

Fixes #9405


AlexMesser:
thank you for contribution!

Issue #9395, opened 2022-09-21
Relations in find operations in a repository may change incorrectly from "left" to "inner"


PedroLimaComputacao:
## Issue Description
In find operations, when you have a condition in `where` that relates to an optional relation, the relation changes from "left" to "inner" incorrectly. That's a problem when the condition is written as an array (resulting in an OR operator), so a condition in one of the array elements may be optional if it is not described in all elements.

### Expected Behavior
According to the documentation, FindOptionsRelations is a shortened version of `join` and `leftJoinAndSelect`, therefore, all relations described in them should always result in a left join.

### Actual Behavior
All relations represented in a `where` statement that weren't explicitly described in `relations` are then created as an inner join. That's iffy, but... sure why not. The problem is that even relations that were explicitly written in `relations` (therefore expected to be of type "left") are then converted to type "inner". That's invasive and corrupts the expected behavior of `relations`.

### Steps to Reproduce

1. Have an entity model with an optional many-to-one relation. For example: An employee may or may not have an office, but an office can house many employees if it's big enough.
2. Create a repository of the model.
3. Call `.find()`, and in the options write where as an array. In one of the elements, have a condition that uses the optional many-to-one relation and in another element write other conditions that don't depend on the optional relation.

```typescript
import { Entity, Column, PrimaryGeneratedColumn, JoinColumn, ManyToOne } from 'typeorm';

@Entity('employee')
export class Employee {
	@Column()
	name: string;

	@Column()
	office_id?: string;
	
	@ManyToOne(() => Office)
	@JoinColumn({ name: 'office_id' })
	office?: Office;
}

@Entity('office')
export class Office {
	@PrimaryGeneratedColumn('uuid')
	id: string;
	
	@Column()
	hasWindow: boolean;
}
```
```typescript
import myDataSource from 'path/to/myDataSource';
import Employee from 'path/to/Employee';

const rep = myDataSource.getRepository(Employee);

const result = await rep.find({
	where: [
		{ name: 'Jhon' },
		{ office: { hasWindow: false } },
	],
	relations: { office: true },
});

```

In the above example, an employee named Jhon who doesn't have an office would not appear in the results, even though he perfectly suffices the first condition of the where (his name is Jhon). It shouldn't matter that he doesn't have an office, but the join relation "office" was wrongfully converted to an "inner".

### My Environment

| Dependency          | Version  |
| ---                 | ---      |
| Operating System    | Windows 10 |
| Node.js version     | v14.17.4  | 
[...]

PR #9411, opened 2022-09-26
Test case to show issue/9395

carantunes:
Proves #9395

### Description of change
Test case to showcase the bug.
From the docs https://typeorm.io/find-options, the following query 
```
userRepository.find({
    relations: {
        project: true,
    },
    where: {
        project: {
            name: "TypeORM",
            initials: "TORM",
        },
    },
})
```
should execute 

```
SELECT * FROM "user"
LEFT JOIN "project" ON "project"."id" = "user"."projectId"
WHERE "project"."name" = 'TypeORM' AND "project"."initials" = 'TORM'
```

But in reality in executing 

```
SELECT * FROM "user"
INNER JOIN "project" ON "project"."id" = "user"."projectId"
WHERE "project"."name" = 'TypeORM' AND "project"."initials" = 'TORM'
```
which is a problem for optional relations.

Further investigation showed `LEFT` is only replaced by `INNER` when we query by a relation property, eg if we query relation with `IsNull`, join in still of type `LEFT`.

### Pull-Request Checklist

### DIFF
--- a/test/functional/find-options/basic-usage/find-options-test-utils.ts
+++ b/test/functional/find-options/basic-usage/find-options-test-utils.ts
@@ -84,13 +84,13 @@ export async function prepareData(manager: EntityManager) {
     await manager.save(post3)
 
     const post4 = new Post()
-    post1.id = 4
-    post1.title = "Post #4"
-    post1.text = "About post #4"
-    post3.author = user1
-    post1.tags = []
-    post1.counters = new Counters()
-    post1.counters.likes = 1
-    post1.counters.likedUsers = [user1]
+    post4.id = 4
+    post4.title = "Post #4"
+    post4.text = "About post #4"
+    post4.author = user1
+    post4.tags = []
+    post4.counters = new Counters()
+    post4.counters.likes = 1
+    post4.counters.likedUsers = [user1]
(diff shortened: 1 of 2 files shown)


test/functional/find-options/basic-usage/find-options-test-utils.ts
```diff
@@ -82,4 +82,15 @@ export async function prepareData(manager: EntityManager) {
...
+
+    const post4 = new Post()
+    post1.id = 4
+    post1.title = "Post #4"
+    post1.text = "About post #4"
+    post3.author = user1
+    post1.tags = []
+    post1.counters = new Counters()
+    post1.counters.likes = 1
+    post1.counters.likedUsers = [user1]
```
PedroLimaComputacao, line 94: Should be `post4` instead of `post1` and `post3`.

PedroLimaComputacao (changes requested):
Should be `post4` instead of `post1` or `post3`

test/functional/find-options/basic-usage/find-options-test-utils.ts
carantunes: thank you!

Issue #6631, opened 2020-08-28
Memory leak in lazy relation


MeiK2333:
**Issue type:**

[x] bug report
[ ] feature request
[ ] documentation issue

**Database system/driver:**

[ ] `cordova`
[ ] `mongodb`
[ ] `mssql`
[ ] `mysql` / `mariadb`
[ ] `oracle`
[x] `postgres`
[ ] `cockroachdb`
[x] `sqlite`
[ ] `sqljs`
[ ] `react-native`
[ ] `expo`

**TypeORM version:**

[x] `latest`
[ ] `@next`
[ ] `0.x.x` (or put your version here)

**Steps to reproduce or a small repository showing the problem:**

```ts
import 'reflect-metadata';
import { Entity, PrimaryGeneratedColumn, Column, createConnection, Index, ManyToOne, OneToMany } from "typeorm";

type Lazy<T extends object> = Promise<T> | T;

@Entity()
class Bind {
  @PrimaryGeneratedColumn()
  id: number;

  @ManyToOne(type => Source, source => source.bind, { lazy: true })
  source: Lazy<Source>
}

@Entity()
@Index(['problemId', 'source'])
class Problem {
  @PrimaryGeneratedColumn()
  id: number;

  @Column()
  problemId: string

  @ManyToOne(type => Source, source => source.problem, { lazy: true })
  source: Lazy<Source>
}

@Entity()
class Source {
  @PrimaryGeneratedColumn()
  id: number;

  @OneToMany(type => Problem, problem => problem.source, { lazy: true })
  problem: Lazy<Problem[]>

  @OneToMany(type => Bind, bind => bind.source, { lazy: true })
  bind: Lazy<Bind[]>
}

(async () => {
  const connection = await createConnection({
    type: "postgres",
    host: "localhost",
    port: 5432,
    username: "postgres",
    password: "password",
    database: "StepByStep",
    entities: [
      Bind, Source, Problem
    ],
    synchronize: true,
    logging: false
  })

  const source = new Source();
  await connection.manager.save(source);

  const bind = new Bind();
  bind.source = source;
  await connection.manager.save(bind);


  const repository = connection.getRepository(Problem);
  for (let i = 0; i < 100; i++) {
    {
      // Memory leak code
      console.time(`create ${i}`);
      const problem = repository.create({
        source,
        problemId: String(i)
      });
      await repository.save(problem);
      console.timeEnd(`create ${i}`);
    }
    {
      // pass code
[...]

PR #9437, opened 2022-10-06
fix: explicitly define property for entity relation as enumerable

orcwarrior:
### Description of change
Change to makes properties for entity relation explicitly non-enumerable, which resolves a "rare" issue that leads to a memory leak.

In some environments it seems like an entity object was created with column values set to undefined, thus Object.defineProperty uses already defined value for enumerable - true. This leads to RawSqlResultsToEntityTransformer.transform() run-getters on relation fields (as they were enumerable) resolving the promises - running queries which eventually leads to the Free memory exhaustion on the machine :)

Closes: #6631 

### Pull-Request Checklist

Unfortunately while trying to find a way to replicate the issue in a couple of ways:

- Trying to make entity class derivative of BaseEntity
- Trying to run compiled tests
- Replicating my project entities in close to 1:1 way

I wasn't able to replicate the same behavior in the test case, but while debugging I found exactly the same behavior in my code as other users reported in the issue.

Making `enumerable: false` explicitly stated shouldn't regress any code and seems like running tests locally confirmed that, however ~~single test did fail:~~

UP: I see that on CI tests worked just fine

### DIFF
--- a/src/query-builder/RelationLoader.ts
+++ b/src/query-builder/RelationLoader.ts
@@ -448,6 +448,7 @@ export class RelationLoader {
                 }
             },
             configurable: true,
+            enumerable: false,
         })
     }
 }


AlexMesser:
thank you for contribution!

PR #9455, opened 2022-10-11
fix(DataSource): assign isInitialized: true at the end of initialize

### LINKED ISSUE #9454: EntityMetadataNotFoundError on some postgres schema DataSource
## Issue Description

> **Note** For context on how we use TypeORM, I'v written [an article](https://dev.to/fraxken/moving-myunisoft-nodejs-back-to-typeorm-3fok) in August.

### Expected Behavior

EntityMetadataNotFoundError should not happen.

### Actual Behavior

For some unknown reason we have errors `EntityMetadataNotFoundError` happening **only for some DataSource**. All DataSource are initialized with the same options (except for the `schema` property)

Each time the error happen the request became **Idle** and consume one connection slot on the database. Combined with our WebSocket retry mechanism it consumed all available slots in few minutes.

### Steps to Reproduce

Right now we are failing to reproduce. On our DEV & PREPROD the error almost never appear (only on Production where we have a lot of charges).

In Production we have close to 1,000 schemas (for each schema we open a DataSource and we keep it in memory). Since 0.3.x we use our own internal Map with the schema name as key (since there is no more getConnection() etc..).

### My Environment

| Operating System    | debian   |
| Node.js version     | 16.15.1  |
| Typescript version  | ^4.6.3  |
| TypeORM version     | ^0.3.5 (0.3.9 resolved) |

### Additional Context

We had already identified the error a few months ago on some of our services that failed on the first request for a DataSource (at least this is what I was thinking at the time.. since other requests on the same datasource were ok).

However we had not much errors, so it was still not very serious and we didn't worry more than that.

image

However yesterday for a new Discussion service using WebSocket the issue became just terrible with dozen of `No metadata` when all clients were connecting at the same time on the WebSocket.

image

At some point there is so much errors and idle requests that the service go crazy (probably cause all sockets are used.. I'v seen EMFILE error at time)
image

We only work with postgres

| `postgres`        | yes           |

fraxken:
Fixes #9454

### Description of change

Update the `initialize()` method of the DataSource class. Only set `isInitialized: true` at the very end to ensure everything has been properly loaded.

### Pull-Request Checklist

Note: is there is a guide on how to run test ?

### DIFF
--- a/src/data-source/DataSource.ts
+++ b/src/data-source/DataSource.ts
@@ -234,9 +234,6 @@ export class DataSource {
         // connect to the cache-specific database if cache is enabled
         if (this.queryResultCache) await this.queryResultCache.connect()
 
-        // set connected status for the current connection
-        ObjectUtils.assign(this, { isInitialized: true })
-
         try {
             // build all metadatas registered in the current connection
             await this.buildMetadatas()
@@ -260,6 +257,9 @@ export class DataSource {
             await this.close()
             throw error
         }
+
+        // set connected status for the current connection
+        ObjectUtils.assign(this, { isInitialized: true })
 
         return this
     }


pleerock:
Not sure why tests didn't run for this PR, but I think this PR is safe to merge. Thanks for the contribution!

pleerock:
I had to revert this change because it actually broke the tests. The function calls below the current code `ObjectUtils.assign(this, { isInitialized: true })` is actually relying on `isInitialized` being `true`. So, the change is actually invalid.

You need to create a new PR with a proper changes and actually run the tests before push. Thanks.

PR #9459, opened 2022-10-12
feat: allow per-migration control over transaction behavior

### LINKED ISSUE #7087: Allow to specify transaction property for individual migrations.
## Feature Description
Allow each individual migration to specify if it should be ran inside a transaction or not.

### The Problem
Some operations cannot be ran inside a transaction (eg. `CREATE INDEX CONCURRENTLY` on postgresql). While the typeorm CLI does allow to specify a `transaction` argument on the `migration:run` command, this argument applies globally. This poses a problem when migrations are ran in a CI pipeline where we don't want to completely abandon the guarantees that wrapping a migration in a transaction gives us (safe rollback if something shall go wrong) and leaves no other resource other than running some migrations manually

### The Solution

When running a migration the `MigrationExecutor` checks if the migration class defines a value for the transaction property in order to decide if this particular migration should be wrapped in a transaction or not. This value should take precedence over the `transaction` property of the `MigrationExecutor`.

There is an issue if the transaction argument is `'all'` ie. all migrations should be executed within a single transaction, as there is no clear cut way to conciliate it with some migrations wanting to be ran outside transactions, so maybe the `Migration`'s transaction property should only be honored if the `MigrationExecutor`'s transaction is either `'each'` in which case if a migration has `transaction = false` it is not wrapped in a migration, or `MigrationExecutor`'s transaction is `'none'` in which case only migrations which explicitly state it are wrapped in transactions.

- [ ] `aurora-data-api`
- [ ] `aurora-data-api-pg`
- [ ] `better-sqlite3`
- [ ] `cockroachdb`
- [ ] `cordova`
- [ ] `expo`
- [ ] `mongodb`
- [ ] `mysql`
- [ ] `nativescript`
- [ ] `oracle`
- [x] `postgres`
- [ ] `react-native`
- [ ] `sap`
- [ ] `sqlite`
- [ ] `sqlite-abstract`
- [ ] `sqljs`
- [ ] `sqlserver`

TimBeyer:
### Description of change

Closes #7087

Implements the ability to override the transaction behavior per migration.
This allows for example to run every migration in a separate transaction by default,
but to run specific migrations like using `CREATE INDEX CONCURRENTLY` outside of a transaction.

Trying to set per-migration overrides in the `"all"` transaction mode will cause an error to prevent accidental confusing behavior.

When the transaction mode is `"each"` the default is to run in a transaction.
When the transaction mode is `"none"` the default is to not run in a transaction.

### Pull-Request Checklist


stefreak:
@pleerock do you happen to have time to look at this one? We are currently having to work around this one by using transaction mode "none" and manually managing transactions in every migration

TimBeyer:
@AlexMesser Apologies for the ping, but I noticed you recently reviewed and merged some PRs.
Any chance we could have this one reviewed? We've been working around this with migration wrappers that create a lot of code overhead and would really love to be able to get rid of those.

As it stands now there's no way to easily benefit from having transactions wrap common migrations while still allowing things like `CREATE INDEX CONCURRENTLY` to work.

AlexMesser:
thank you for contribution!

PR #9469, opened 2022-10-17
fix: cacheId not used when loading relations with take

4l3ss:
### Description of change
The cacheId is set to undefined when finding entities with relations and the pagination(s) option(s) is/are set (take/skip).

It seems the cacheId is set to undefined for the main request used for pagination when we load related entities.

I appended a `-pagination` to the identifier of the pagination request.

Related #5983

### Pull-Request Checklist

### DIFF
--- a/src/query-builder/SelectQueryBuilder.ts
+++ b/src/query-builder/SelectQueryBuilder.ts
@@ -3309,9 +3309,9 @@ export class SelectQueryBuilder<Entity extends ObjectLiteral>
                 .limit(this.expressionMap.take)
                 .orderBy(orderBys)
                 .cache(
-                    this.expressionMap.cache
-                        ? this.expressionMap.cache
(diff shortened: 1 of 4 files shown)


4l3ss:
I don't have the bandwidth to write tests right now but I'll try.

4l3ss:
after further investigation it seems to be a bit more complicated to fix. I have written a test and I am now working on a better fix

pleerock:
We probably need a better strategy and prevent caching of both queries... but okay let's merge this PR. Thanks for the contribution!

cnorthfield:
This is causing an issue when passing a manual `id` to `cache`. 

Using the same id to invalidate the cache does not work as it creates 2 cache keys, `${this.expressionMap.cacheId}-pagination` and `${this.expressionMap.cacheId}`.

PR #9483, opened 2022-10-20
feat: Introduce TYPEORM_DATASOURCE env var as default datasource file

luca-nardelli:
### Description of change

** This is currently in draft, looking for feedback **

This PR adds a new TYPEORM_DATASOURCE environment variable that is used as a default value for the `-d` cli option.

The reasoning behind this is using the typeorm cli in a project. Without this option, defining a `typeorm` command in package.json with a default datasource becomes cumbersome because not all sub-commands accept the datasource option leading to `Unknown argument: d` errors, whereas passing it as env variable makes it much easier.

Right now I made the minimum edits to show this as a proof of concept, but if you feel this could be a good PR I can work on it further (documentation, etc...). Let me know!

### Pull-Request Checklist

- [N/A] This pull request links relevant issues as `Fixes #0000`

### DIFF
--- a/src/commands/CacheClearCommand.ts
+++ b/src/commands/CacheClearCommand.ts
@@ -19,6 +19,7 @@ export class CacheClearCommand implements yargs.CommandModule {
             describe:
                 "Path to the file where your DataSource instance is defined.",
             demandOption: true,
+            default: process.env.TYPEORM_DATASOURCE,
         })
     }
 
--- a/src/commands/MigrationShowCommand.ts
+++ b/src/commands/MigrationShowCommand.ts
@@ -18,6 +18,7 @@ export class MigrationShowCommand implements yargs.CommandModule {
             describe:
                 "Path to the file where your DataSource instance is defined.",
             demandOption: true,
+            default: process.env.TYPEORM_DATASOURCE,
         })
     }
 
--- a/src/commands/SchemaDropCommand.ts
+++ b/src/commands/SchemaDropCommand.ts
@@ -21,6 +21,7 @@ export class SchemaDropCommand implements yargs.CommandModule {
             describe:
                 "Path to the file where your DataSource instance is defined.",
             demandOption: true,
+            default: process.env.TYPEORM_DATASOURCE,
         })
     }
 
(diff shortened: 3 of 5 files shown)


pleerock:
It's not clear the use cases where passing additional `-d` option to the CLI command won't work.

luca-nardelli:
Hey, thanks for answering and for your work. You're right, let me elaborate more.

I am using TypeORM in several NestJS projects. Usually, my workflow is the following:

- Create a datasource file in `./src/datasource.ts` that defines the datasource. This file also uses the `process.env.DATABASE_URL` variable to pick the database credentials so that the same datasource can be reused in different environments (dev,staging,prod).
- Read that datasource in NestJS using dynamic imports so that the same configuration is used between Typeorm's CLI and NestJS.
- Add some scripts to the `package.json`, more specifically a `typeorm` script which basically calls the locally installed `typeorm` for migrations and schema managment.

Now, ideally I'd like all typeorm commands in that project to automatically use that datasource without having to manually add it every time I invoke it, so my first attempt was to revisit how the script is defined in the package.json to the following:

```
"scripts" : {
  "typeorm": "typeorm -d ./src/datasource.ts"
}
```

and with this I can succesfully run things like `npm run typeorm schema:sync`.

However, as soon as I run, for example, `npm run typeorm entity:create`  or `npm run typeorm migration:create` typeorm complains with the error `Unknown argument: d` because those sub-commands do not accept a datasource parameter.

Because of this, I thought about defining a default datasource via ENV var so that I could just setup the package.json script like this `"typeorm": "TYPEORM_DATASOURCE=./src/datasource.ts typeorm` and then all relevant sub-commands would get that default datasource unless overridden via command line option.

Let me know if the use case is clear enough, if you need any further info from me or if there's an easier way to achieve what I'm doing! Thanks

pleerock:
you should pass your commands as `npm run typeorm entity:create -- -d ./data-source`, read more [here](https://stackoverflow.com/questions/11580961/sending-command-line-arguments-to-npm-script).

luca-nardelli:
I already knew how to do that, however the whole point of this PR was to
avoid having to add `-- -d ./datasource` to every typeorm command, since in
the scope of the project there is always only one datasource and always
specifying that in each command gets boring quite quickly.

What I would like to is `npm run typeorm <subcommand>` and that
automatically picks up the datasource I specified in the env var, so that I
can run it in my dev environment or in staging/production without having to
add the -d option, as the project has a single datasource for everything.

Il dom 4 dic 2022, 07:33 Umed Khudoiberdiev ***@***.***> ha
scritto:

pleerock:
It doesn't make lot sense, so no, we won't add another small complexity to the code and cli usage.

Issue #9446, opened 2022-10-08
findOne({where: {id: null}}) returns the first item in the database


Sid-Turner-Ellis:
## Issue Description
-findOne where ID is null returns the first item in the database instead of null

### Expected Behavior
- If we try to find the record labelled with an id of null, the return should be null instead of the first item in the database

### Actual Behavior
- The first value in the database is returned

### Steps to Reproduce
- Run this with your own entity and see the first item logged to the console instead of null
```
    const item = await dataSource.getRepository(UserEntity).findOne({
      where: {
        id: null as any,
      },
    });
    console.log(item); // returns the first item

```
### My Environment

| Dependency          | Version  |
| Operating System    |   Mac os   |
| Node.js version     | 14  
| Typescript version  | 4.8.2
| TypeORM version     | 0.3.8

### Relevant Database Driver(s)
sqlite

### Are you willing to resolve this issue by submitting a Pull Request?
 - ✅ Yes, I have the time, and I know how to start.
 - ✖️ Yes, I have the time, but I don't know how to start. I would need guidance.
 - ✖️ No, I don’t have the time, but I can support (using donations) development.
 - ✖️ No, I don’t have the time and I’m okay to wait for the community / maintainers to resolve this issue.

iJhefe:
dup #9316

pleerock:
Closing as duplicate.

PR #9487, opened 2022-10-22
feat: Allows to specify how "undefined" and "null" values should be treated in conditions

### LINKED ISSUE #9316: typeorm findOneBy( { id: null or undefined } ) method return the first record on the table  !
### Expected Behavior

return null

### Actual Behavior
```
const user = await userRepository.findOneBy({
            id: undefined or if it's null
        });
```
this return the first user in the users table !

```
User {
  id:,
  username: ,
  ....
}
```

### Steps to Reproduce

1. create typeorm connection
2. create user entity and migrate it to the database
3. query the database using the user repository ' call findOneBy method '

| `sqlite`          | yes            |

for other idk :)

iJhefe:
This feature Allows to specify how TypeORM should handle "undefined" and "null" values in conditions

### Description of change
before this feature TypeORM just exclude `null` and `undefined` values from conditions. Now it can be defined at "DataSouceOptions".

It changes the default behavior about `null` values. Now it's should be interpreted as `IS NULL` and not exclude it

Allow `null` at `FindWhereConditions`

The options for "null" and "undefined" values are:
- `is-null`: condition will be "IS NULL"
- `exclude`: condition will not be considered
- `throw`: will throw an exception

 FIXES #9316, #9446, #9527

### Pull-Request Checklist


pleerock:
This configurable behavour introduces redundant complexity to the code and the ability of making such things flexible brings confusion to the user and I think overall is a bad design step. 

For cases like this `findOneBy( { id: null or undefined } )` I suggest to add another code which iterates over all given properties to a given FindOptionsWhere object and makes sure there is at least one defined property.

bombillazo:
> This configurable behavior introduces redundant complexity to the code and the ability of making such things flexible brings confusion to the user and I think overall is a bad design step.
> 
> For cases like this `findOneBy( { id: null or undefined } )` I suggest to add another code which iterates over all given properties to a given FindOptionsWhere object and makes sure there is at least one defined property.

I agree that making this configurable will be a bad design choice for the long run and brings opposing behaviors to the ORM. The package should establish a convention (my thoughts on that [here](https://github.com/typeorm/typeorm/issues/9316#issuecomment-1299328534)) and let the users/developers add application logic around that convention.

pleerock:
See my answer above ^

trim21:
thanks for your responding

bombillazo:
Yes, I believe this is an issue because null is pretty much broken when it comes to value comparisons. I dont understand why TyoeORM would treat null as undefined when null has a meaning in the DB domain and intuitively maps to the JS domain.

cnorthfield:
@pleerock I’d suggest a breaking change to update the behaviour permanently instead of a configuration option.

gugu:
We are moving to Sequelize because of such issues - this is the only ORM, which cares about security (both null/undefined behavior and JSON injections)

mptsonev:
Why would it ever be ok to specify a where conditional with a value of `undefined` or `null`? Surely there's no discussion to be had here, if a conditional is `null` or `undefined` the ORM should just throw an error, instead of silently ignoring erroneous values. @pleerock please reconsider the approach here, this is an absurd behavior.

nikolaipetkov:
This is a serious security issue which is purposely left unaddressed, pretty wild.

gugu:
We solved this problem by switching to sequelize. It took a lot of resources, but now we have 0 problems

twigs67:
Why was this closed and not merged @iJhefe ? I can't believe this is still an issue with TypeOrm. This is a massive security risk. At the very least, this should be well documented to let users know.

iJhefe:
> Why was this closed and not merged @iJhefe ? I can't believe this is still an issue with TypeOrm. This is a massive security risk. At the very least, this should be well documented to let users know.

finally fixed #11332

PR #9512, opened 2022-11-01
fix: fixed DeepPartial 

flamewow:
fixed DeepPartial Date handling (allowed number and string for Date type)

### Pull-Request Checklist


src/common/NativeObject.ts
```diff
@@ -0,0 +1,4 @@
+/**
+ * Represents native objects (Date)
+ */
+export type NativeObject = Date
```
pleerock, line 4: this extra type is definitely redundant

src/common/DeepPartial.ts
```diff
@@ -1,9 +1,13 @@
...
+
 /**
  * Same as Partial<T> but goes deeper and makes Partial<T> all its properties and sub-properties.
  */
 export type DeepPartial<T> =
     | T
-    | (T extends Array<infer U>
+    | (T extends NativeObject
+          ? T
+          : T extends Array<infer U>
```
pleerock, line 10: For consistency purposes with other code parts its better to write it as:

```ts
export type DeepPartial<T> =
    | T
    | (T extends Array<infer U>
          ? DeepPartial<U>[]
          : T extends Map<infer K, infer V>
          ? Map<DeepPartial<K>, DeepPartial<V>>
          : T extends Set<infer M>
          ? Set<DeepPartial<M>>
          : T extends Date
          ? T
          : T extends object
          ? { [K in keyof T]?: DeepPartial<T[K]> }
          : T)
```

pleerock:
Please add the new test case testing the new functionality and making sure it works properly. Thank you.

src/common/DeepPartial.ts
pleerock: probably we need to add `Buffer` as well...

```ts
export type DeepPartial<T> =
    | T
    | (T extends Array<infer U>
          ? DeepPartial<U>[]
          : T extends Map<infer K, infer V>
          ? Map<DeepPartial<K>, DeepPartial<V>>
          : T extends Set<infer M>
          ? Set<DeepPartial<M>>
          : T extends Date
          ? T
          : T extends Buffer
          ? T
          : T extends object
          ? { [K in keyof T]?: DeepPartial<T[K]> }
          : T)
```

flamewow: Buffer doesn't exist in the web browser, by adding it we are going to get issues in browser most likely.

src/common/NativeObject.ts
flamewow: the idea was to have a type that is going to represent a set of native types. Initially, I set it to `Date | Buffer`:

I.e. it would be 
```
          : T extends NativeObject
          ? T
```
instead of 
```
          : T extends Date
          ? T
          : T extends Buffer
          ? T
```
but then I removed Buffer because it's not native to browsers. 
Suggest keeping NativeObject type in case we add more in the future 😄

flamewow:
> Please add the new test case testing the new functionality and making sure it works properly. Thank you.

The PR is about `type` changes, if I set a date field from a DeepPartial<Entity> to say string or a number TS is gonna throw an error on transpiling, I don't know how to test for that. If you have ideas, please LMK. 

I've updated test/github-issues/1716/issue-1716.ts to follow the new type

src/common/DeepPartial.ts
pleerock: You are correct, `Buffer` doesn't exist in the web browser, but there are several reasons you need to add it:

* we already use `Buffer` everywhere - so it something won't work in the browser - it won't work anyway
* we can't remove the `Buffer` because users will start complain about them missing this feature
* actually for the browser it won't be a problem because it's `type-level` issue. Users will have to install @types/node with the Buffer in it if they have issues.

src/common/NativeObject.ts
pleerock: I understand, but for the simplification purposes we don't need another type.

pleerock:
Please add a new test instead of changing old one, to make sure we won't change the old test in the future and remove the functionality you've added. Also you can test type behavior by using special comments like `// @ts-expect-error`.

pleerock:
Closing as there is no answer from author.

PR #9618, opened 2022-12-13
fix: add type annotation to sqlJsConfig in connection options

cijiugechu:
### Pull-Request Checklist

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "typeorm",
-  "version": "0.3.10",
+  "version": "0.3.11",
   "lockfileVersion": 2,
   "requires": true,
   "packages": {
     "": {
       "name": "typeorm",
-      "version": "0.3.10",
+      "version": "0.3.11",
       "license": "MIT",
       "dependencies": {
         "@sqltools/formatter": "^1.2.2",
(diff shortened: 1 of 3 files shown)


pleerock:
For those who don't use sql.js the line `import type { SqlJsConfig } from 'sql.js'` will cause the problems. And we definitely  should not enforce for example mysql users to install sql.js typings because of such error.

We cannot really do anything with it right now.

Solution is to use typings on your side and cast your config before you send to the connection options, e.g.

```ts
const myConfig: SqlJsConfig = { ... }
new DataSource({ sqlJsConfig: myConfig })
```

This will provide you the type-safely you need.

cijiugechu:
> For those who don't use sql.js the line `import type { SqlJsConfig } from 'sql.js'` will cause the problems. And we definitely should not enforce for example mysql users to install sql.js typings because of such error.
> 
> We cannot really do anything with it right now.
> 
> Solution is to use typings on your side and cast your config before you send to the connection options, e.g.
> 
> ```ts
> const myConfig: SqlJsConfig = { ... }
> new DataSource({ sqlJsConfig: myConfig })
> ```
> 
> This will provide you the type-safely you need.

Thanks, I will be closing this PR.

PR #9634, opened 2022-12-16
refactor: remove date-fns

jdgjsag67251:
### Description of change

Since TypeORM only uses one function from the date-fns library it is unnecessary to install this large package.
Unfortunately, date-fns does not allow you to install specific functions. So, instead I just copied the one file from date-fns.

The main reason behind this change is the fact that we need to install all our packages in our build system every time. This can take quite some time. We also need to copy these modules around.
Since date-fns is one of the largest packages in our system we have the strong desire to remove it.

### Pull-Request Checklist

- [ ] `npm run test` passes with this change (N/A)


pleerock:
Unfortunately it adds extra code to our codebase that we have to track and maintain. Can't we do it the other way, e.g. simpler `parseISO` implementation or other small & popular package that do provide the same functionality?

jdgjsag67251:
Yeah understandable. Unfortunately, the closest package I could find is [iso8601.js](https://www.npmjs.com/package/iso8601.js) which only has 39 weekly downloads (at time of writing).
It does something pretty simple though[^1]:
```js
const local = new Date();
const offset = local.getTimezoneOffset() * 60 * 1000;
const utc = local.getTime() - offset;

return utc;
```

I could maybe add this instead?

[^1]: I've simplified the code a little.

pleerock:
yes, if this solution does work - just paste this small code into our codebase (into DateUtils?) and use it.

jdgjsag67251:
Cool, thanks. I've updated the code

pleerock:
I've restarted oracle tests many times already, not sure why they failed. Will you be able to investigate the issue?

jdgjsag67251:
> I've restarted oracle tests many times already, not sure why they failed. Will you be able to investigate the issue?

I've just ran all the Oracle tests locally and they all seem to run fine. I have absolutely no idea why they fail 🥲

AlexMesser:
thank you for contribution!

PR #9688, opened 2023-01-06
Add test to prove that issue #2331 is resolved

### LINKED ISSUE #2331: Inconsistent handling of "undefined" between save() and update()

Spissable:
### Description of change

* closes #2331

I was facing this issue in a private repo which still uses typorm 0.2.x. I figured I'll give it a look and wrote the test but the issue isn't appearing anymore. I ran the same test against pre 0.3.0 and it failed as expected. So I'd consider the mentioned issue to be resolved.

### DIFF
--- a/test/github-issues/2331/entity/Post.ts
+++ b/test/github-issues/2331/entity/Post.ts
@@ -1,0 +1,13 @@
+import { BaseEntity, Entity, Column, PrimaryColumn } from "../../../../src"
+
+@Entity()
+export class Post extends BaseEntity {
+    @PrimaryColumn()
+    id: number
+
+    @Column({ type: "varchar" })
+    title: string
+
+    @Column({ type: "varchar", nullable: true })
+    author: string | null
+}
--- a/test/github-issues/2331/issue-2331.ts
+++ b/test/github-issues/2331/issue-2331.ts
@@ -1,0 +1,59 @@
+import "reflect-metadata"
+import { expect } from "chai"
+import { DataSource, Repository } from "../../../src"
+import { Post } from "./entity/Post"
+import {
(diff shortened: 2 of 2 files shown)


AlexMesser:
thank you for contribution!

PR #9703, opened 2023-01-11
feat: naming strategy for legacy Oracle

f-wrobel:
This feature is LegacyOracleNamingStrategy which can be used to handle Oracle error ORA-00972.

### Description of change

In older Oracle version identifier length limit is 30 bytes. If identifier is too long, Oracle throws the infamous error ORA-00972. This limitation is handled in several places through out the code. Most of the cases when that error could appear are fixed now.
But it's still impossible to create a table with a column name which exceeds this limit. In order to fix this, I've created a new naming strategy - LegacyOracleNamingStrategy. It allows to use 2 modes of shortening column names - truncate and hash.
In case of using hash shortening, the column human-readability is lost, but it's always a choice of a user to use this naming strategy. This feature will help people who wants to use an older Oracle db.

This kind of solution was proposed in the past:
https://github.com/typeorm/typeorm/issues/452#issuecomment-298896416
Users of other ORMs like Hibernate also tend to use a different naming strategy for Oracle.
https://stackoverflow.com/questions/38565495/how-to-handle-the-ora-00972-identifier-is-too-long-exception-with-hibernate-5-a

I will also create an issue to discuss the problem of ORA-00972 because in newer versions, it's not present due to increased limit to 128 bytes.

New naming strategy tests are included.

### Pull-Request Checklist

- [X] ~~This pull request links relevant issues as `Fixes #0000`~~ N/A
- [x] ~~Documentation has been updated to reflect this change~~ N/A


pleerock:
Thanks for the contribution! Btw you might want to expose newly added naming strategy in the `index.ts` file.

f-wrobel:
@pleerock Actually I did expose :smile: : https://github.com/typeorm/typeorm/blob/0eb74411d62f9132ba45154040b204f31a98c779/src/index.ts#L164
Btw, please look at this issue which is related to this PR: https://github.com/typeorm/typeorm/issues/9704 . Thanks for the merge 🙏

Issue #9671, opened 2023-01-03
Issue with the find one repository method


sandeshtamanq:
### Issue description

I am trying to use find one method where it returns the total blogs made by users but it's throwing error

### Expected Behavior

Return array of blogs made by a specific user

### Actual Behavior

Trying to filter the blogs according to the uses but throws this error
```
Type 'number' is not assignable to type 'boolean | FindOperator<any> | FindOptionsWhere<UserEntity> | FindOptionsWhere<UserEntity>[] | EqualOperator<...>'.

25       where: { author: userId },
                  ~~~~~~ 
```

### Steps to reproduce

**Blog Entity**
```
@Entity({ name: 'blogs' })
export class BlogEntity {
  @PrimaryGeneratedColumn()
  id: number;

  @Column()
  title: string;

  @Column({ default: '' })
  description: string;

  @Column({ default: '' })
  body: string;

  @Column()
  slug: string;

  @Column({ type: 'timestamp', default: () => 'CURRENT_TIMESTAMP' })
  createdAt: Date;

  @Column({ type: 'timestamp', default: () => 'CURRENT_TIMESTAMP' })
  updatedAt: Date;

  @BeforeUpdate()
  updateTimestamp() {
    this.updatedAt = new Date();
  }

  @Column({ nullable: true })
  headerImage: string;

  @Column({ default: 0 })
  likes: number;

  @Column({ nullable: true })
  publishedDate: Date;

  @Column({ default: false })
  isPublished: boolean;

  @ManyToOne((type) => UserEntity, (user) => user.blogs)
  author: UserEntity;
}
```

**User Entity**

```
@Entity({ name: 'users' })
export class UserEntity {
  @PrimaryGeneratedColumn()
  id: number;

  @Column()
  name: string;

  @Column({ unique: true })
  username: string;

  @Column({ unique: true })
  email: string;

  @Column({ select: false })
  password: string;

  @Column({ type: 'enum', enum: UserRole, default: UserRole.USER })
  role: UserRole;

  @Column({ nullable: true })
  profileImage: string;

  @BeforeInsert()
  emailToLowerCase() {
    this.email = this.email.toLowerCase();
  }

  @OneToMany((type) => BlogEntity, (blog) => blog.author)
  blogs: BlogEntity[];
}
```

**Blog service file**

```
@Injectable()
export class BlogService {
  constructor(
    @InjectRepository(BlogEntity)
    private blogRepository: Repository<BlogEntity>,
  ) {}

[...]

Issue #9508, opened 2022-10-30
Unable to use where clause with FindOptions due to incompatible types


robertmain:
## Issue Description
It does not appear to be possible to use the `where` option in `Repository<T>.find()`. This is because `FindOneOptions<Entity>.where?` lists all the columns under `where` as nullable i.e:

```ts
where: {
  id?: string,
  createdAt?: string,
}
```

But, those types are being typed as `NonNullable` by TypeORM - in this case: `FindOptionsWhereProperty<NonNullable<Entity["id"]>>`.

Thus, it seems to be complaining because `FindOneOptions<Entity>.where` properties are (rightfully) nullable, when the actual entity properties themselves that those fields refer to are not nullable.

### Expected Behavior

This code should be valid:

```ts
this.repository.find({
  where: {
    id: 'my-id',
  },
```

### Actual Behavior

I get an error message that says:

```
Type '{ id: "my-id"; }' is not assignable to type 'FindOptionsWhere<Entity> | FindOptionsWhere<Entity>[]'.
  Types of property 'id' are incompatible.
    Type 'string' is not assignable to type 'FindOptionsWhereProperty<NonNullable<Entity["id"]>>'.ts(2322)
FindOneOptions.d.ts(23, 5): The expected type comes from property 'where' which is declared here on type 'FindManyOptions<Entity>'
```

### My Environment

| Dependency          | Version  |
| ---                 | ---      |
| Operating System    |   Debian       |
| Node.js version     | 14.17.0 | 
| Typescript version  | 4.8.4  | 
| TypeORM version     | 0.3.10  | 

### Relevant Database Driver(s)

| DB Type           | Reproducible  |
|-------------------| ---           |
| `aurora-mysql`    | no            |
| `aurora-postgres` | no            |
| `better-sqlite3`  | no            |
| `cockroachdb`     | no            |
| `cordova`         | no            |
| `expo`            | no            |
| `mongodb`         | no            |
| `mysql`           | no            |
| `nativescript`    | no            |
| `oracle`          | no            |
| `postgres`        | yes            |
| `react-native`    | no            |
| `sap`             | no            |
| `spanner`         | no            |
| `sqlite`          | no            |
| `sqlite-abstract` | no            |
| `sqljs`           | no            |
| `sqlserver`       | no            |

### Are you willing to resolve this issue by submitting a Pull Request?

 - ✖️ Yes, I have the time, and I know how to start.
[...]

Issue #9350, opened 2022-09-07
FindOptionsRelations when @OneToMany


Stephen2:
## Issue Description

Have an Entity User {
  @OneToMany(() => Child, ({ parent }) => parent, { cascade: true })
  children!: readonly Child[] | undefined;
}

Have an Entity Child {
  @OneToMany(() => Claim, ({ child }) => child, { cascade: true })
  claims!: readonly Claim[] | undefined;
}

Try and select `user+children+claims`:

```
    const user = await baseRepository.findOne({
      relations: {
        children: {
          claims: true,
          journalEntries: true,
        },
      },
      where: {
        id,
      },
    });
```

Get compile time error:
```
Type '{ children: { claims: true; journalEntries: true; }; }' is not assignable to type 'FindOptionsRelationByString | FindOptionsRelations<User>'.
  Types of property 'children' are incompatible.
    Type '{ claims: true; journalEntries: true; }' is not assignable to type 'boolean | readonly (boolean | FindOptionsRelations<Child> | undefined)[] | undefined'.
      Object literal may only specify known properties, and 'claims' does not exist in type 'readonly (boolean | FindOptionsRelations<Child> | undefined)[]'.ts(2322)
```

Such that you have to write code:
```
      relations: {
        children: [
          {
            claims: true,
            journalEntries: true,
          },
        ],
      },
```
Note the addition of `[]` which doesn't make sense when writing code.

Causes runtime error: `EntityPropertyNotFoundError: Property "0" was not found in "Child". Make sure your query is correct.`

Can workaround with `ts-expect-error`:
```
      relations: {
        children: {
          // @ts-expect-error TypeORM types want this to be an [] becuase it's silly
          // That causes runtime error: EntityPropertyNotFoundError: Property "0" was not found in "Child". Make sure your query is correct.
          claims: true,
          journalEntries: true,
        },
      },
```

### Expected Behavior

FindOptionsRelations types to be modified for this nested case

### Actual Behavior

Types are wrong, error messages shown above

### Steps to Reproduce

As above

### My Environment

| Dependency          | Version  |
| ---                 | ---      |
| Operating System    |          |
| Node.js version     | 16.15.1  | 
| Typescript version  | 4.8.2  | 
| TypeORM version     | 0.3.9 | 

### Are you willing to resolve this issue by submitting a Pull Request?

 - ✖️ Yes, I have the time, and I know how to start.
 - ✖️ Yes, I have the time, but I don't know how to start. I would need guidance.
[...]

Issue #9346, opened 2022-09-04
Type 'EqualOperator<TestType>' is not assignable to type '"string"'


azarus:
## Issue Description
User defined custom types do not work with find operators.
Given the following type to an entity field fails:

```typescript
export type TestType = "TEST" | "YOLO" | "ASD";
```

`Type 'EqualOperator<TestType>' is not assignable to type '"TEST" | "YOLO" | "ASD" | FindOperator<"TEST"> | FindOperator<"YOLO"> | FindOperator<"ASD">'`

### Expected Behavior
No error,

### Actual Behavior

Throws an error:
```
Type 'EqualOperator<TestType>' is not assignable to type '"TEST" | "YOLO" | "ASD" | FindOperator<"TEST"> | FindOperator<"YOLO"> | FindOperator<"ASD">'.
  Type 'EqualOperator<TestType>' is not assignable to type 'FindOperator<"ASD">'.
    Types of property 'value' are incompatible.
      Type 'TestType' is not assignable to type '"ASD"'.
        Type '"TEST"' is not assignable to type '"ASD"'.

                test: Equal<TestType>("ASD"),
                   ~~~~
 ```

### Steps to Reproduce

Define a custom type like so

```typescript
type TestType = "TEST" | "YOLO" | "ASD";

class MyEntity
{
   @Column()
   test: TestType;
 };
```
Query it using find operators.
```typescript
MyEntity.find({
    where: 
    {
           test: Equal<TestType>("ASD"),
    }
});
```

### My Environment

| Dependency          | Version  |
| ---                 | ---      |
| Operating System    | Win / Linux |
| Node.js version     | xv16.16.0 |
| Typescript version  | 4.8.2  | 
| TypeORM version     | 0.3.9 |

### Relevant Database Driver(s)

| DB Type           | Reproducible  |
|-------------------| ---           |
| `aurora-mysql`    | no            |
| `aurora-postgres` | no            |
| `better-sqlite3`  | no            |
| `cockroachdb`     | no            |
| `cordova`         | no            |
| `expo`            | no            |
| `mongodb`         | no            |
| `mysql`           | no            |
| `nativescript`    | no            |
| `oracle`          | no            |
| `postgres`        | no            |
| `react-native`    | no            |
| `sap`             | no            |
| `spanner`         | no            |
| `sqlite`          | no            |
| `sqlite-abstract` | no            |
| `sqljs`           | no            |
| `sqlserver`       | no            |

[...]

PR #9709, opened 2023-01-14
Fix for operators not being assignable when using user defined types.

azarus:
### Description of change

Given the following model
```ts
export type MyCustomType = "A" | "B" | "C";

@Entity()
export class MyEntity
{
    @PrimaryGeneratedColumn()
    id: number

    @Column()
    data: MyCustomType = "A";

}
```
and the following code:
```ts
  const entity = new MyEntity()
  entity.data = "B";
  await connection.manager.save(entity)
  const foundEntity = await connection.manager.findOne(MyEntity, {
     where: {
            data: Equal("B")
     }
  });
```
User gets the following error message:
```
Type '{ data: EqualOperator<string>; }' is not assignable to type 'FindOptionsWhere<MyEntity> | FindOptionsWhere<MyEntity>[] | undefined'.
  Types of property 'data' are incompatible.
    Type 'EqualOperator<string>' is not assignable to type '"A" | "B" | "C" | FindOperator<"A"> | FindOperator<"B"> | FindOperator<"C"> | undefined'.
      Type 'EqualOperator<string>' is not assignable to type 'FindOperator<"C">'.
        Types of property '_value' are incompatible.
          Type 'string | FindOperator<string>' is not assignable to type '"C" | FindOperator<"C">'.
            Type 'string' is not assignable to type '"C" | FindOperator<"C">'.ts(2322)
```

The new signature broken and i don't really have the patience to untagle it, but by just allowing object literals the typings are still preserved and the code works.

### Pull-Request Checklist

Fixes:

https://github.com/typeorm/typeorm/issues/9671

https://github.com/typeorm/typeorm/issues/9508

https://github.com/typeorm/typeorm/issues/9350

https://github.com/typeorm/typeorm/issues/9346

https://github.com/typeorm/typeorm/issues/9269

https://github.com/typeorm/typeorm/issues/9251

https://github.com/typeorm/typeorm/issues/9223

https://github.com/typeorm/typeorm/issues/8939

And a bunch more.

full list pretty much can be found by searching for the error message.
https://github.com/typeorm/typeorm/issues?q=is+not+assignable+to+type+is%3Aopen

### DIFF
--- a/src/find-options/FindOneOptions.ts
+++ b/src/find-options/FindOneOptions.ts
@@ -10,7 +10,6 @@ import {
 } from "./FindOptionsRelations"
 import { FindOptionsOrder } from "./FindOptionsOrder"
 
-
 /**
  * Defines a special criteria to find specific entity.
  */
--- a/src/find-options/FindOneOptions.ts
+++ b/src/find-options/FindOneOptions.ts
@@ -10,6 +10,7 @@ import {
 } from "./FindOptionsRelations"
 import { FindOptionsOrder } from "./FindOptionsOrder"
 import { ObjectLiteral } from "../common/ObjectLiteral";
+
 /**
  * Defines a special criteria to find specific entity.
  */
--- a/src/find-options/FindOneOptions.ts
+++ b/src/find-options/FindOneOptions.ts
@@ -9,7 +9,7 @@ import {
     FindOptionsRelations,
 } from "./FindOptionsRelations"
 import { FindOptionsOrder } from "./FindOptionsOrder"
-import { ObjectLiteral } from "../common/ObjectLiteral";
+
 
 /**
  * Defines a special criteria to find specific entity.
@@ -30,7 +30,7 @@ export interface FindOneOptions<Entity = any> {
     /**
      * Simple condition that should be applied to match entities.
      */
-    where?: FindOptionsWhere<Entity>[] | FindOptionsWhere<Entity> | ObjectLiteral
+    where?: FindOptionsWhere<Entity>[] | FindOptionsWhere<Entity>
 
     /**
      * Indicates what relations of entity should be loaded (simplified left join form).
--- a/src/find-options/FindOneOptions.ts
+++ b/src/find-options/FindOneOptions.ts
@@ -9,7 +9,7 @@ import {
     FindOptionsRelations,
 } from "./FindOptionsRelations"
 import { FindOptionsOrder } from "./FindOptionsOrder"
-
+import { ObjectLiteral } from "../common/ObjectLiteral";
 /**
  * Defines a special criteria to find specific entity.
  */
@@ -29,7 +29,7 @@ export interface FindOneOptions<Entity = any> {
     /**
      * Simple condition that should be applied to match entities.
      */
-    where?: FindOptionsWhere<Entity>[] | FindOptionsWhere<Entity>
+    where?: FindOptionsWhere<Entity>[] | FindOptionsWhere<Entity> | ObjectLiteral
 
     /**
      * Indicates what relations of entity should be loaded (simplified left join form).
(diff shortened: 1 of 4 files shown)


azarus:
Updated the unit tests that were related to introducing this bug.

@khteh
@pleerock
@Ginden
@AlexMesser
@imnotjames

Please review and let me know if this is acceptable for now. And if  this is acceptable please release an update for this asap so people are able to use the Find-Operators! Thank you!

khteh (approved):

pleerock:
Test seem to be invalid. I've checked your test without your changes - and everything works.

pleerock:
And yes, this fix doesn't resolve the problem.

After more investigations I found that issue is only with `manager`, it's not reproducible when you use it with `getRepository`.

So, basically the problem is now "Equal" infers the type. See how it does it with `getRepository`:

image

And see how it works with `manager`:

image

For me it's not clear why TypeScript is able to find a right type in the first case and isn't in the second case.

Solution is to explicitly specify `as const`, e.g.:

image

But of course the best solution is to provide a minimal reproduction code (without TypeORM specifics) and ask TypeScript community why it behaves such way. Because I have no idea about it.

pleerock:
I can help you with the minimal reproduction of this issue:

```ts
type ObjectType<T> = new (...args: any[]) => T

class EqualOperator<T> {
    constructor(value: T) {
    }
}

type Where<Entity> = {
    [P in keyof Entity]?: Entity[P] | EqualOperator<Entity[P]>
}

function findOne<Entity>(
    entityClass: ObjectType<Entity>,
    options: Where<Entity>,
) {}

function find<Entity>(entityClass: ObjectType<Entity>) {
    return (options: Where<Entity>) => {
        return
    }
}

findOne(Post, {
    type: new EqualOperator("B"), // EqualOperator doesn't infer the type
})

find(Post)({
    type: new EqualOperator("B"), // EqualOperator infers the type
})
```

pleerock:
Reported in [TypeScript#52649](https://github.com/microsoft/TypeScript/issues/52649)

pleerock:
@azarus we'll try TypeScript 5.0 feature once it's released (as commented in https://github.com/microsoft/TypeScript/issues/52649), maybe that will work. Meanwhile "as const" might work.

> Perhaps rolling back to the previous version of the type declarations? As that worked fine.

Type system isn't always perfect. With type system we'll have to leverage. Particularly this issue isn't "that much" of the issue to remove other useful type system features we have.

bombillazo:
Thanks for attending this issue!  Was the suggestion left by the TS team able to solve the inference issue?

azarus:
> Particularly this issue isn't "that much" of the issue to remove other useful type system features we have.

Well we're unable to use user defined types. It sort of defeats the purpose of using typescript for an ORM as we have to name everything string, when in fact we want to limit usage of a field to certain values. So as a result enums are pretty much broken sadly. The problem is that it limits a developer's ability to design types effectively and offer clear usage of an entity/model. Even if one is planning to use the fixed version in the future, refactoring code base later on is rather complicated. Me personally, just forked the repo, and included typeorm into the projects directly and patched the thing to work instead of using npm directly. This way if i wanted to switch back to the npm version, just auto replace the import statements and all should be well.

Anyways, Maybe I'm getting something wrong here.
But currently i don't see any workaround for this other than hard casting types in the query itself. :(

My point being:
If writing queries requires casting to a certain type to work, we might as well just drop that support and allow any type.

If this is required to get things working:
```
Entity.find({ something: MyType } as FindOperator<any>)
```
We might as well just allow the end user to use "any" in the query. Because we're not better off anywhere.

I am just trying to find a middle ground that would work for the majority until we can sort this out with ts5.0 when it is released.

kvbutler:
Is this PR going to be merged, I have this issue. `BaseEntity` has a property `id!: number`. Cast suggestion as mentioned in #8939  is working but ugly...

**Not working** 
image

**Ugly cast**
image

pleerock:
since typescript 5.0 was already released, you can try to apply changes suggested by typescript team.

hinogi:
@azarus can you review this in the light of the current typescript?

pleerock:
Closing since there are no activity here since TypeScript fix.

Issue #9708, opened 2023-01-13
 Reasons for fixing left join in repository queries


eric-byun:
### Issue description

fixing left join on create query from repository

### Expected Behavior

https://github.com/typeorm/typeorm/blob/master/src/query-builder/SelectQueryBuilder.ts#L3873

It is fixing left join.

i made a query from repository
```javascript
fooRepository.find({
  relations: {
   bar: true,
 }
});
```

I expected the query below.
```sql
select * from foo join bar;
```

But the query below was executed.
```sql
select * from foo left join bar;
```

That is the result what i don't want. I found the above code to be the problem.
Why is fixing left join?

I want set the joinType in Entity like this 
`@OneToMany(() => foo, (foo) => foo.bar, { joinType: 'inner' })`..
Please tell me reason..

### Actual Behavior

```sql
select * from foo left join bar;
```

### Steps to reproduce

It is fixing to `left`

https://github.com/typeorm/typeorm/blob/master/src/query-builder/SelectQueryBuilder.ts#L3873

### My Environment

| Dependency          | Version  |
| ---                 | ---      |
| Operating System    |          |
| Node.js version     | x.y.zzz  |
| Typescript version  | x.y.zzz  |
| TypeORM version     | x.y.zzz  |

### Relevant Database Driver(s)

- [ ] aurora-mysql
- [ ] aurora-postgres
- [ ] better-sqlite3
- [ ] cockroachdb
- [ ] cordova
- [ ] expo
- [ ] mongodb
- [X] mysql
- [ ] nativescript
- [ ] oracle
- [ ] postgres
- [ ] react-native
- [ ] sap
- [ ] spanner
- [ ] sqlite
- [ ] sqlite-abstract
- [ ] sqljs
- [ ] sqlserver

### Are you willing to resolve this issue by submitting a Pull Request?

Yes, I have the time, and I know how to start.

iJhefe:
what do you think about #9710 behavior ?

pleerock:
> Please tell me reason..

because other users can suppose opposite, e.g. the following code:

```ts
fooRepository.find({
  relations: {
   bar: true,
 }
});
```

to be executed as:

```ts
select * from foo left join bar;
```

And unlike the current implementation, it won't be possible to customize the behavior. So, current behavior makes more sense.

> @OneToMany(() => foo, (foo) => foo.bar, { joinType: 'inner' })

[...]

PR #9710, opened 2023-01-14
feat: add join operator

iJhefe:
this feature enable to explicit what join should be used in find options

### Description of change
with this we will be able to explicit define how relations should be joined using
```ts
// only users with posts
const users = await manager.find(User, {
   relations: {
    posts: InnerJoin()
  }
})

// users with posts or not
const users = await manager.find(User, {
  relations: {
    posts: LeftJoin()
  }
})

// using LeftJoin operator is the same as (at this moment)
const users = await manager.find(User, {
  relations: {
    posts: true
  }
})
```
using join operator we can be sure about how relations are being joined

### Pull-Request Checklist

### DIFF
--- a/src/find-options/FindOptionsRelations.ts
+++ b/src/find-options/FindOptionsRelations.ts
@@ -1,4 +1,5 @@
 import { ObjectID } from "../driver/mongodb/typings"
+import { JoinOperator } from "./JoinOperator"
 
 /**
  * A single property handler for FindOptionsRelations.
@@ -6,9 +7,9 @@ import { ObjectID } from "../driver/mongodb/typings"
 export type FindOptionsRelationsProperty<Property> = Property extends Promise<
     infer I
 >
-    ? FindOptionsRelationsProperty<NonNullable<I>> | boolean
+    ? FindOptionsRelationsProperty<NonNullable<I>> | boolean | JoinOperator
     : Property extends Array<infer I>
-    ? FindOptionsRelationsProperty<NonNullable<I>> | boolean
+    ? FindOptionsRelationsProperty<NonNullable<I>> | boolean | JoinOperator
     : Property extends string
     ? never
     : Property extends number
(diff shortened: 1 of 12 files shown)


iJhefe:
may fixes #9708

eric-byun (approved):
Thanks for your quick fix!

pleerock:
Not sure if I want to add additional complication to the `relations`.

What if I suggest an alternative implementation for the `inner join`:

```ts
// only users with posts
const users = await manager.find(User, {
   relations: {
    posts: true,
  },
  where: {
    posts: Not(IsNull())
  }
})
```

eric-byun:
```
// only users with posts
const users = await manager.find(User, {
   relations: {
    posts: true,
  },
  where: {
    posts: Not(IsNull())
  }
})
```

The result data in the inner join query is the same.

But the query you suggested is
1. creating the join result data
2. filter the data (filter post null data)

so slower than inner join

eric-byun:
This PR also has functional limitations.
This function cannot be used by mixing inner join and outer join.

ex)
```
select *
from foo f
inner join bar  b on f.id = b.id
left join qaz q on b.id = q.id
```

But... In complex cases like this, I think it's right to use Query Builder.

eric-byun:
Could you tell me, When can use this function in my project..? 
When will be merged?

pleerock:
From the performance perspective [it's the same](https://stackoverflow.com/questions/354070/sql-join-where-clause-vs-on-clause#comment18926888_354094).

From the code perspective I find current solution acceptable: it's both flexible and built on exist concepts which make logically sense. I don't see a point to add more complications with different new scenarios like this one.

I'll have to close it until this feature gets more popularity.

cjnoname:
Hi @pleerock ,

Why did I get the error "This relation isn't supported by given find operator" while testing your solution?

Issue #673, opened 2017-07-16
Syntax for COUNT(), AVG() and SUM()


adnan-kamili:
I didn't find any docs for operations like count(), avg(), sum(). Something like:

```javascript
const productCount = await this.productRepository.count({ accountId });
const itemCount = await this.productRepository.find({accountId}).sum("items");
```

pleerock:
use query builder to build any sql you need and get raw results. Something like:

```ts
let [sum] = await context.locations
            .createQueryBuilder("location")
            .select("SUM(location.something)", "sum")
            .getRawOne(); 
```

adnan-kamili:
Thanks! It would be nice if there were functions for the same as these operations are quite common and other orms support them out of the box.

pleerock:
Maybe. In the future.

greg-murray-volusion:
> use query builder to build any sql you need and get raw results. Something like:
> 
> ```ts
> let [sum] = await context.locations
>             .createQueryBuilder("location")
>             .select("SUM(location.something)", "sum")
>             .getRawOne(); 
> ```

I think this one should be 

```typescript
let { sum } = await ...
```

syuilo:
Yes, as @greg-murray-volusion said, `let [sum] =` is not working.
Use `let { sum } =` instead.

zaksesti-ns8:
Is this still true?  I see SelectQueryBuilder has a `getCount()` method.

AlexWehe:
> Maybe. In the future.

https://youtu.be/0UkYwZlTLxE

SimantoR:
> Maybe. In the future.

Future is now. I understand that most people don't have the time and it only makes sense that people using the library should be contributing to the source-code the most. No one does things for no value what-so-ever, which is apparent.

iamchathu:
Is there any update on this?

luka-rmar:
Any updates for this functionality?

italodeandra:
If anyone is looking for on how to get the list and the count (the count doesn't follow the `take`/`skip`), you can find an example here (the only place I could find in the official website):

[https://typeorm.io/#undefined/loading-from-the-database - This link doesn't work. Just a reference](https://typeorm.io/#undefined/loading-from-the-database)

But as I just notice the link doesn't work. You can find on the **Loading from the database** section on https://typeorm.io/#/

And here is the code:

```
let [allPhotos, photosCount] = await photoRepository.findAndCount();
console.log("All photos: ", allPhotos);
[...]

PR #9737, opened 2023-01-27
feat: Add support for SQL aggregate functions SUM, AVG, MIN, and MAX to the Repository API

netroy:
### Description of change
This addresses #673, and adds a convenient way to call aggregate functions via the repository API.

### Pull-Request Checklist

- [x] This pull request links relevant issues

### DIFF
--- a/src/common/PickKeysByType.ts
+++ b/src/common/PickKeysByType.ts
@@ -1,0 +1,6 @@
+/**
+ * Pick only the keys that match the Type `U`
+ */
+export type PickKeysByType<T, U> = keyof {
+    [P in keyof T as T[P] extends U ? P : never]: T[P]
+}
--- a/src/common/PickKeysByType.ts
+++ b/src/common/PickKeysByType.ts
@@ -1,6 +1,7 @@
 /**
  * Pick only the keys that match the Type `U`
  */
-export type PickKeysByType<T, U> = keyof {
-    [P in keyof T as T[P] extends U ? P : never]: T[P]
-}
+export type PickKeysByType<T, U> = string &
+    keyof {
+        [P in keyof T as T[P] extends U ? P : never]: T[P]
+    }
--- a/src/entity-manager/EntityManager.ts
+++ b/src/entity-manager/EntityManager.ts
@@ -1025,7 +1025,7 @@ export class EntityManager {
     }
 
     /**
-     * Return the AVG of a column
+     * Return the MIN of a column
      */
     minimum<Entity extends ObjectLiteral>(
         entityClass: EntityTarget<Entity>,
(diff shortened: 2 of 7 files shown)


src/entity-manager/EntityManager.ts
```diff
@@ -1001,6 +1002,67 @@ export class EntityManager {
...
+        field: PickKeysByType<Entity, number>,
+        where: FindOptionsWhere<Entity> | FindOptionsWhere<Entity>[] = {},
+    ): Promise<number | null> {
+        const metadata = this.connection.getMetadata(entityClass)
+        const { [fnName]: result } = await this.createQueryBuilder(
+            entityClass,
+            metadata.name,
+        )
+            .setFindOptions({ where })
+            .select(`${fnName}(${field.toString()})`, fnName)
```
AlexMesser, line 1061: `field.toString()` must be properly escaped. Oracle test currently failing with `QueryFailedError: ORA-00904: "COUNTER": invalid identifier`.

netroy: @AlexMesser Thanks for pointing that out. 
Updated it to use `connection.driver.escape` instead. All tests pass now 🎉

src/entity-manager/EntityManager.ts
```diff
@@ -1001,6 +1002,70 @@ export class EntityManager {
...
+    }
+
+    private async callAggregateFun<Entity extends ObjectLiteral>(
+        entityClass: EntityTarget<Entity>,
+        fnName: "SUM" | "AVG" | "MIN" | "MAX",
+        columnName: PickKeysByType<Entity, number>,
+        where: FindOptionsWhere<Entity> | FindOptionsWhere<Entity>[] = {},
+    ): Promise<number | null> {
+        const metadata = this.connection.getMetadata(entityClass)
+        const { [fnName]: result } = await this.createQueryBuilder(
```
pleerock, line 1056: Can we simplify a bit and do it as `const result = ` instead `const { [fnName]: result } =`, and then just do `result[fnName]`. For my stupid brain it took a minute to figure out how it works, don't want to other peoples like me to think about this code, so I think simplification will be better for us, such people.

src/common/PickKeysByType.ts
```diff
@@ -0,0 +1,7 @@
+/**
+ * Pick only the keys that match the Type `U`
+ */
+export type PickKeysByType<T, U> = string &
```
pleerock, line 4: what is the point of `string &` here? it completely makes useless the next code.

pleerock:
Thanks. Can you please format the code, since lint has failed.

pleerock:
Thank you for contribution! 🎉

src/common/PickKeysByType.ts
alroniks: I found out the issue in case when field in the model has custom column name in the database. For example, field name `balanceValue`, but in the table it is `balance_value`. TS force me set `balanceValue` but generated SQL contains the wrong field at the end.

PR #9764, opened 2023-02-07
feat: use import type to resolve some internal circular references

pleerock:
Replaced uses of "import" with "import type" where it's possible. This might be a breaking change and will require TypeScript version 3.8+


pleerock:
Closing for now as it doesn't seem to quickly resolve circular problems.

PR #9775, opened 2023-02-11
fix: prevent foreign key support during migration batch under sqlite

### LINKED ISSUE #9770: [sqlite3] migrations batch do not halt foreign key management
### Issue description

With SQLite3 driver when a batch of migrations is being run the Foreign Key management isn't put on hold during the migration. As a consequence it can trigger data loss if an alter on a table is done while this table is refered with a foreign key with DELETE ON CASCADE

### Expected Behavior

When a I do a set of migrations which contains a addColum on a table referred to by another table foreign key which is set with ON DELETE CASCADE, I expect the migrations to be run without any loss of data contains in the tables.

Issue #2576 did solve it issue for an unitary migration but it does not applies with a set of pending migrations executed in batch mode.

### Actual Behavior

Basically the issue is that each migration isn't framed by a PRAGMA goreign_key OFF/ON in dedicated transaction.
so a addColumn will generate batch that will recreate the table with the new column, transfering the data from old table to new one and then dropping the old table with finally the new table being renamed.
The table drop is triggering if any existing on delete CASCADE the delation of all the records tied to this table via foreign key.

the 2 recently added methods beforeMigration and afterMigration are not being called before and after the execution of each pending migration as shown in bellow code:
https://github.com/typeorm/typeorm/blob/c669f50bddadb00d0f896e16ccfe5e1ecad16d3c/src/migration/MigrationExecutor.ts#L334-L361

### Steps to reproduce

running this snippet you should reproduce the issue: https://github.com/sinopsysHK/typeorm-testcase

npm install
npm start

### My Environment

| Operating System    | Ubuntu 18.04.3 |
| Node.js version     | 16.19.0  |
| Typescript version  | 8.19.3  |
| TypeORM version     | 0.3.11  |

### Additional Context

_No response_

- [ ] aurora-mysql
- [ ] aurora-postgres
- [ ] better-sqlite3
- [ ] cockroachdb
- [ ] cordova
- [ ] expo
- [ ] mongodb
- [ ] mysql
- [ ] nativescript
- [ ] oracle
- [ ] postgres
- [ ] react-native
- [ ] sap
- [ ] spanner
- [X] sqlite
- [ ] sqlite-abstract
- [ ] sqljs
- [ ] sqlserver

sinopsysHK:
Schema changes in migrations batch cannot be done on table which has referring
 foreign keys with ON DELETE CASCADE without deleting its content.

Closes: #9770
Closes: #2584

### Description of change
Complemente the PR #7922 by executing beforeMigration before each migration and execting afterMigration after each migration right before Transaction is started and right after transaction is being commited.
As specified in [sqlite doc](https://www.sqlite.org/pragma.html#pragma_foreign_keys) : 

> This pragma is a no-op within a transaction; foreign key constraint enforcement may only be enabled or disabled when there is no pending [BEGIN](https://www.sqlite.org/lang_transaction.html) or [SAVEPOINT](https://www.sqlite.org/lang_savepoint.html).

So the change is to run PRAGMA foreign_keys = OFF before starting the migration and the transaction that contains it and to set it back to ON right after once commited.

This change will avoid to trigger CASCADE events as well as constraint check on foreign keys.
Hence avoiding to have some deleted data unexpectedly because any table change is managed throught a clone creation and original table drop.

### Pull-Request Checklist

- [X] `npm run test` passes with this change (well actually it breaks at same place than without he change)
- [X] This pull request links relevant issues as `Fixes #9770 #2584`
- [ ] Documentation has been updated to reflect this change (no change of interface)

### DIFF
--- a/src/migration/MigrationExecutor.ts
+++ b/src/migration/MigrationExecutor.ts
@@ -311,6 +311,7 @@ export class MigrationExecutor {
         // start transaction if its not started yet
         let transactionStartedByUs = false
         if (this.transaction === "all" && !queryRunner.isTransactionActive) {
+            await queryRunner.beforeMigration()
             await queryRunner.startTransaction()
             transactionStartedByUs = true
         }
@@ -327,10 +328,12 @@ export class MigrationExecutor {
                 }
 
                 if (migration.transaction && !queryRunner.isTransactionActive) {
+                    await queryRunner.beforeMigration()
                     await queryRunner.startTransaction()
                     transactionStartedByUs = true
                 }
 
+                this.connection.logger.log("info", "--start migration--")
                 await migration
                     .instance!.up(queryRunner)
                     .catch((error) => {
@@ -347,8 +350,10 @@ export class MigrationExecutor {
                             migration,
                         )
                         // commit transaction if we started it
-                        if (migration.transaction && transactionStartedByUs)
+                        if (migration.transaction && transactionStartedByUs) {
                             await queryRunner.commitTransaction()
+                            await queryRunner.afterMigration()
+                        }
                     })
                     .then(() => {
                         // informative log about migration success
@@ -362,8 +367,10 @@ export class MigrationExecutor {
             }
 
             // commit transaction if we started it
-            if (this.transaction === "all" && transactionStartedByUs)
+            if (this.transaction === "all" && transactionStartedByUs) {
                 await queryRunner.commitTransaction()
+                await queryRunner.afterMigration()
+            }
         } catch (err) {
             // rollback transaction if we started it
             if (transactionStartedByUs) {
--- a/src/migration/MigrationExecutor.ts
+++ b/src/migration/MigrationExecutor.ts
@@ -333,7 +333,6 @@ export class MigrationExecutor {
                     transactionStartedByUs = true
                 }
 
-                this.connection.logger.log("info", "--start migration--")
                 await migration
                     .instance!.up(queryRunner)
                     .catch((error) => {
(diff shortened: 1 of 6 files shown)


sinopsysHK:
Hello,

may I have some tips on the testing: on my setup tests running over latest master are breaking in the similar way than from my branch while the specific test I created to check my changes are working as expect runs fine.

Should I fix those side issues myself to get this PR validated ?

src/migration/MigrationExecutor.ts
```diff
@@ -327,10 +328,12 @@ export class MigrationExecutor {
                 }
 
                 if (migration.transaction && !queryRunner.isTransactionActive) {
+                    await queryRunner.beforeMigration()
                     await queryRunner.startTransaction()
                     transactionStartedByUs = true
                 }
 
+                this.connection.logger.log("info", "--start migration--")
```
pleerock, line 336: need to use `logger.logSchemaBuild` instead

test/functional/migrations/show-command/command.ts
```diff
@@ -31,7 +31,12 @@ describe("migrations > show command", () => {
     it("can recognise no pending migrations", () =>
         Promise.all(
             connections.map(async (connection) => {
-                await connection.runMigrations()
+                const m = await connection.runMigrations()
+                /*
+                    .catch((err) => { console.log(`error: ${err}`)})
+                    .then(() => { console.log('migrations done')})
+                */
+                console.log(`have performe ${m.length} migrations`)
```
pleerock, line 39: remove console.log!

Issue #9381, opened 2022-09-17
The column option 《transformer》 affects the result of the query condition generation


pwfyhao:
typeorm version:  0.3.9;

before code:
<img width="680" alt="image" src="https://user-images.githubusercontent.com/36838341/190854220-57476099-f994-409d-b2f9-0a5b9cc7b55c.png">
<img width="617" alt="image" src="https://user-images.githubusercontent.com/36838341/190854258-c57e44cf-01ca-404a-bbd9-0bce137fe486.png">
result:
![image](https://user-images.githubusercontent.com/36838341/190854280-7d0db002-154d-41e7-b01a-8c066b040a52.png)

current:
<img width="278" alt="image" src="https://user-images.githubusercontent.com/36838341/190854306-a8b3f02b-2a36-4d15-9543-e621d3083561.png">
result:
![image](https://user-images.githubusercontent.com/36838341/190854327-2e0db370-64b8-4206-b598-8dc536a61047.png)

Is this a feature or a BUG!!!

ertl:
same problem here with typeorm 0.3.10.

It looks like all find-option operators are affected.

@pwfyhao did you found a workaround for this problem?

PR #9777, opened 2023-02-11
fix: transform values for FindOperators #9381

ertl:
The column transformer didn't work in combination with FindOperators.
This correction does not perform a transformation on a FindOperator value

Fixes: #9381

### Pull-Request Checklist

[~ ] Documentation has been updated to reflect this change

### DIFF
--- a/src/find-options/FindOperator.ts
+++ b/src/find-options/FindOperator.ts
@@ -142,13 +142,22 @@ export class FindOperator<T> {
         if (this._value instanceof FindOperator) {
             this._value.transformValue(transformer)
         } else {
-            this._value = Array.isArray(this._value)
-                ? this._value.map(
-                      (v: any) =>
-                          transformer &&
-                          ApplyValueTransformers.transformTo(transformer, v),
-                  )
-                : ApplyValueTransformers.transformTo(transformer, this._value)
+            // do not convert each value of json, since an array is a valid json object
+            this._value =
+                Array.isArray(this._value) &&
+                this["@instanceof"] !== Symbol.for("JsonContains")
+                    ? this._value.map(
+                          (v: any) =>
+                              transformer &&
+                              ApplyValueTransformers.transformTo(
+                                  transformer,
+                                  v,
+                              ),
+                      )
+                    : ApplyValueTransformers.transformTo(
+                          transformer,
+                          this._value,
+                      )
         }
     }
 }
--- a/src/find-options/FindOperator.ts
+++ b/src/find-options/FindOperator.ts
@@ -142,10 +142,8 @@ export class FindOperator<T> {
         if (this._value instanceof FindOperator) {
             this._value.transformValue(transformer)
         } else {
-            // do not convert each value of json, since an array is a valid json object
             this._value =
-                Array.isArray(this._value) &&
-                this["@instanceof"] !== Symbol.for("JsonContains")
+                Array.isArray(this._value) && this._multipleParameters
                     ? this._value.map(
                           (v: any) =>
                               transformer &&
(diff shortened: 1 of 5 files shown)


vlad-ostrovski:
Your solution will not perform the required transformation on the `FindOperator`'s value, so eventually the query will be performed on the wrong (or rather unexpected), untransformed value. 

Or am I missing something?

ertl:
@vlad-ostrovski can't tell at the moment. I will have a look at it tomorrow... But can you verify that the fix isn't working I'm your case. By adding a check to your transformer... So that the transformation is only executed when value is not an instance of FindOperator?

vlad-ostrovski:
@ertl But i do need it to be transformed.. even if it's a FindOperator (JsonContains)

ertl:
@vlad-ostrovski the value was not transformed, but the Runtime Error was eliminated, and the transformation was indirectly performed from the database. While this solution worked for me, I acknowledge that it is not the correct behavior and have since corrected it. Can you please review it?

Regarding the JsonContainsOperator and a JSON array, the transformation utilizes the entire array instead of each individual value for the transformation. However, I was unable to test this as it appears that the JsonContainsOperator does not support arrays as input.

src/find-options/FindOperator.ts
```diff
@@ -135,4 +137,27 @@ export class FindOperator<T> {
...
     }
+
+    transformValue(transformer: ValueTransformer | ValueTransformer[]) {
+        if (this._value instanceof FindOperator) {
+            this._value.transformValue(transformer)
+        } else {
+            // do not convert each value of json, since an array is a valid json object
+            this._value =
+                Array.isArray(this._value) &&
+                this["@instanceof"] !== Symbol.for("JsonContains")
```
vlad-ostrovski, line 148: This is an anti-pattern.. the general `FindOperator` shuoldn't know about `JsonContains`..
`JsonContains` can override `transformValue` maybe.. and call `super` if needed

src/find-options/FindOperator.ts
```diff
@@ -1,6 +1,8 @@
 import { ObjectLiteral } from "../common/ObjectLiteral"
 import { FindOperatorType } from "./FindOperatorType"
 import { InstanceChecker } from "../util/InstanceChecker"
+import { ValueTransformer } from "../decorator/options/ValueTransformer"
```
vlad-ostrovski, line 4: Shouldn't be used here.. creates spaghetti

src/find-options/FindOperator.ts
ertl: @vlad-ostrovski  I agree... I was lazy.. I will fix it.

src/find-options/FindOperator.ts
ertl: @vlad-ostrovski any suggestions how solve this.... ?

vlad-ostrovski: why not use the same approach already implemented in `SelectQueryBuilder` for `EqualOperator`s?
look one line above your change in `SelectQueryBuilder`

ertl: @vlad-ostrovski originally, my first approach was to recursively traverse the FindOperators in the parameterValue, but the value of a FindOperator is readonly. Therefore, I had to choose between making it writeable or exposing a transform functionality. Ultimately, I decided to go with the latter since I think it's more readable and provides the added benefit of inheritance (that you also have noted), allowing for different scenarios to be addressed for each operator.

src/find-options/FindOperator.ts
ertl: fixed!

pleerock:
This one is scary to merge, because I don't know how it can affect exist users usage.

ertl:
Actually, the desired behavior should remain the same. If a transformer is not used, the code will not execute. However, if a transformer is implemented, the else branch within 'if' statement (SelectQueryBuilder.ts) will run, utilizing the same code as before. Prior to my modification, it was not possible to use FindOperators with a transformed column, which is an important issue that has now been addressed.

Issue #7907, opened 2021-07-15
Add support for mongodb driver v4


ferenckv:
## Issue Description
Exactly in [this commit](https://github.com/mongodb/node-mongodb-native/commit/a2d78b22b28ae649fa2c4e28294a3a03c446373e#diff-a2a171449d862fe29692ce031981047d7ab755ae7f84c707aef80701b3ea0c80L2) mongodb stopped exporting a Cursor object. 

Because of that, every reference to `ParentCursor.Prototype` in `MongoEntityManager ` ([like here](https://github.com/typeorm/typeorm/blob/master/src/entity-manager/MongoEntityManager.ts#L691)) is failing with error "Cannot read property 'prototype' of undefined".

### Expected Behavior
No error expected.

### Actual Behavior
Error "Cannot read property 'prototype' of undefined"

```
TypeError: Cannot read property 'prototype' of undefined
    at FindCursor.cursor.toArray (/entity-manager/MongoEntityManager.ts:679:37)
    at MongoEntityManager.<anonymous> (/entity-manager/MongoEntityManager.ts:97:23)
    at step (/node_modules/tslib/tslib.js:143:27)
    at Object.next (/node_modules/tslib/tslib.js:124:57)
    at fulfilled (/node_modules/tslib/tslib.js:114:62)
    at processTicksAndRejections (internal/process/task_queues.js:95:5)
```

### Steps to Reproduce

1. Add typeorm dependency
2. Add mongodb v4 dependency.

### My Environment

| Dependency          | Version  |
| ---                 | ---      |
| Operating System    |  Ubuntu 18.04 on wsl2  |
| Node.js version     | v14.17.0 | 
| Typescript version  | v4.0.8 | 
| TypeORM version     | v0.2.34 | 

### Relevant Database Driver(s)

- [ ] `aurora-data-api`
- [ ] `aurora-data-api-pg`
- [ ] `better-sqlite3`
- [ ] `cockroachdb`
- [ ] `cordova`
- [ ] `expo`
- [x] `mongodb`
- [ ] `mysql`
- [ ] `nativescript`
- [ ] `oracle`
- [ ] `postgres`
- [ ] `react-native`
- [ ] `sap`
- [ ] `sqlite`
- [ ] `sqlite-abstract`
- [ ] `sqljs`
- [ ] `sqlserver`

### Are you willing to resolve this issue by submitting a Pull Request?

- [ ] Yes, I have the time, and I know how to start.
[...]

Issue #9727, opened 2023-01-23
Mongo: problem renaming field _id (@ObjectIdColumn)


mrddter:
### Issue description

Mongo: problem renaming field _id (@ObjectIdColumn)

### Expected Behavior

Using an entity like this (where is used `id` field and not `_id`):

```js
@Entity()
export class User {
   @ObjectIdColumn() // { name: '_id' }
   id: ObjectID

   @Index()
   @Column()
   email: string
}
```

I expected to load an entity in this way:

```js
userRepository.preload({ id: id })
userRepository.findOne({ id: id })
userRepository.find({ where: { id: id } })
User.findOne({ id: id })
User.find({ where: { id: id } })
```

### Actual Behavior

Actually, the find doesn't return anything even if on DB everything is correct. 
Other methods also fail, such as `remove(entity)` as evidenced by the example repo.

All works well for these finds:

```js
User.find({ where: { email: 'test@test.it' } })
User.findOneById(id)
userRepository.findOneById(id)
```

The create() / save() adds correctly a record like this:

```json
{
"_id":{"$oid":"63caa1adde907b687dfa8a65"},
"email":"test@test.it"
}
```

But if you run the example (below) use see in the collection something like this:

```json
{"_id":{"$oid":"63ce9be9d7fe09a6b790d9de"},"email":"test@test.it"}

{"_id":{"$oid":"63ce9be9d7fe09a6b790d9df"},"email":"test@test.it","id":null}

{"_id":{"$oid":"63ce9bead7fe09a6b790d9e0"},"email":"another@test.it","id":{"$oid":"63ce9be9d7fe09a6b790d9df"}}

{"_id":{"$oid":"63ce9bead7fe09a6b790d9e1"},"email":"another@test.it","id":null}
```
Console output:

```
User has been saved. Lets insert a new one to update it later
Second entity has been inserted. Lets update it
Entity has been updated. Persist once again to make find and remove then
Entity is loaded:  User { email: 'another@test.it' }
Now remove it by entity {"email":"another@test.it"}
Cannot save. Error:  SubjectWithoutIdentifierError: Internal error. Subject User must have an identifier to perform operation.
    at /Users/dev/typeorm-mongo/node_modules/src/persistence/SubjectExecutor.ts:682:27
    at Array.map (<anonymous>)
    at SubjectExecutor.executeRemoveOperations (/Users/dev/typeorm-mongo/node_modules/src/persistence/SubjectExecutor.ts:680:41)
[...]

Issue #2238, opened 2018-05-30
How to define ObjectID type column (other than _id)


thonythony:
**Issue type:**

[x] question
[ ] bug report
[ ] feature request
[ ] documentation issue

**Database system/driver:**

[ ] `cordova`
[x] `mongodb`
[ ] `mssql`
[ ] `mysql` / `mariadb`
[ ] `oracle`
[ ] `postgres`
[ ] `sqlite`
[ ] `sqljs`
[ ] `react-native`

**TypeORM version:**

[x] `latest`
[ ] `@next`
[ ] `0.x.x` (or put your version here)

**Steps to reproduce or a small repository showing the problem:**

I have an entity : 

```typescript
@Entity()
export class RolePermission implements IRolePermission {
    @ObjectIdColumn() id?: ObjectID;

    @Column()
    @IsNotEmpty()
    roleId: ObjectID;

    @Column()
    @IsNotEmpty()
    permissionId: ObjectID;
}
```

The code compiles. But when I start my server, this error throws :
```bash
> node ./dist/server.js

error:  ColumnTypeUndefinedError: Column type for RolePermission#roleId is not defined and cannot be guessed. Make sure you have turned on an "emitDecoratorMetadata": true option in tsconfig.json. Also make sure you have imported "reflect-metadata" on top of the main entry file in your application (before any entity imported).If you are using JavaScript instead of TypeScript you must explicitly provide a column type.
```

I have turned on the option in my tsconfig.json and I import reflect-metadata at the top of my file server.ts.

I read the code about the decorator ObjectIdColumn and if I think it used only for primary use case, so it's not applicable for my need.

Thanks a lot for your help

pleerock:
you need to install mognodb typings and import ObjectID from mongodb, like this:

```ts
import {ObjectID} from "mongodb";

@Entity()
export class RolePermission implements IRolePermission {
    @ObjectIdColumn() id?: ObjectID;

    @Column()
    @IsNotEmpty()
    roleId: ObjectID;

    @Column()
    @IsNotEmpty()
    permissionId: ObjectID;
}
```

 I know this is confusing, and its something we have to resolve.

thonythony:
I will test that.
But what's the difference between `ObjectID` from typeorm package and `ObjectID` from mongodb package ?

pleerock:
in typeorm package they are just declarations without real implementations. In mongodb they are real :)

thonythony:
[...]

PR #9778, opened 2023-02-11
fix: add support for mongodb driver v5

MatheusMeloAntiquera:
This new fix allow support to mongodb driver v5

Closes: #7907

### Description of change

Add support for mongodb v5.0.1

- `mapReduce` was removed because is deprecated since 5.0, instead should use an aggregation pipeline
(https://www.mongodb.com/docs/manual/reference/method/db.collection.mapReduce/)

### Pull-Request Checklist


tgrassl:
Hi, are there any updates on this or is there something i could help out with? Would love to get support for the new driver soon 😃

tgrassl:
Hi, i looked through the code and it looks like this behaviour is related to this issue: https://github.com/typeorm/typeorm/issues/9727

The additional `.save()` calls always results in an insert because the condition of whether it should be updated never matches. This is because the resulting transformed entities of the `findByIds` query in `SubjectDatabaseEntityLoader` don't contain an id property but the subject entity does. https://github.com/typeorm/typeorm/blob/58fc08840a4a64ca1935391f4709a784c3f0b373/src/persistence/SubjectDatabaseEntityLoader.ts#L116
Entity from DB: CleanShot 2023-03-05 at 18 47 18@2x
Subject Entity to persist: CleanShot 2023-03-05 at 19 03 25@2x

Because it has no id the following happens:

1. [findByPersistEntityLike](https://github.com/typeorm/typeorm/blob/58fc08840a4a64ca1935391f4709a784c3f0b373/src/persistence/SubjectDatabaseEntityLoader.ts#L163) fails to match the result to the subject
2. subject.databaseEntity is skipped
3. subject.databaseEntityLoaded is set to true
4. SubjectExecutor only finds a subject to be inserted instead of updated because it has no databaseEntity

https://github.com/typeorm/typeorm/blob/58fc08840a4a64ca1935391f4709a784c3f0b373/src/persistence/SubjectExecutor.ts#L134-L148
https://github.com/typeorm/typeorm/blob/58fc08840a4a64ca1935391f4709a784c3f0b373/src/persistence/Subject.ts#L182-L200

I think the fix would be to check for the `propertyName` of the ObjectIdColumn instead of only the `databaseNameWithoutPrefixes` in the DocumentEntityTransformer [transform](https://github.com/typeorm/typeorm/blob/58fc08840a4a64ca1935391f4709a784c3f0b373/src/query-builder/transformer/DocumentToEntityTransformer.ts#L29):
```typescript
...
        // handle _id property the special way
        if (metadata.objectIdColumn) {
            const { databaseNameWithoutPrefixes, propertyName } =
                metadata.objectIdColumn

            const documentIdWithoutPrefixes =
                document[databaseNameWithoutPrefixes]
            const documentIdWithPropertyName = document[propertyName]

            if (documentIdWithoutPrefixes) {
                entity[propertyName] = documentIdWithoutPrefixes
                hasData = true
            } else if (documentIdWithPropertyName) {
                entity[propertyName] = documentIdWithPropertyName
                hasData = true
            }
        }
...
```

This now results in a update on the same entity instead of an insert. I pushed my changes and created a pr for your branch 👍

MatheusMeloAntiquera:
[tgrassl](https://github.com/tgrassl)

Thank you for your help dude!

MatheusMeloAntiquera:
@pleerock
Can you do the review please? 

There is an error Oracle test, but we didn't any changes on Oracle source

tgrassl:
@pleerock I would like to update the typings, so everyone can now use the official typings already included in the `mongodb` package. This would also finally remove the confusion around [ObjectId and ObjectID](https://github.com/mongodb/node-mongodb-native/blob/HEAD/etc/notes/CHANGES_5.0.0.md#objectid-type-removed-in-favor-of-objectid) (e.g. https://github.com/typeorm/typeorm/issues/2238)

As far i can see, i cannot use `import { ObjectId } from 'mongodb'` inside the following files, because that would require non-mongodb users to install mongodb right?

- EntityManager.ts
- Repository.ts
- BaseEntity.ts
- EntityId.ts
- FindOptionsOrder.ts
- FindOptionsRelations.ts
- FindOptionsSelect.ts
- FindOptionsWhere.ts

These all just need the type of the ObjectId. Mongo specific files like MongoEntityManager are already imported from mongodb, so I guess they are ok.
My idea would be to only copy the [ObjectId typings](https://github.com/mongodb/js-bson/blob/main/src/objectid.ts) into typeorm but not export it from the package, so these files can use that internal type. Would that be fine for you?

src/entity-manager/MongoEntityManager.ts
```diff
@@ -1,62 +1,76 @@
...
+    InsertOneOptions,
+    InsertOneResult,
+    InsertManyResult,
+    UnorderedBulkOperation,
+    OrderedBulkOperation,
+    IndexInformationOptions,
+    ObjectId,
+    MongoError,
+    FilterOperators,
+} from "mongodb"
```
pleerock, line 64: we can't have direct imports from mongodb

pleerock:
> These all just need the type of the ObjectId. Mongo specific files like MongoEntityManager already import from mongodb, so I guess they are ok.

No, we must not have imports from `mongodb` anywhere, because non-monogdb users will have tsc errors and they'll be forced to install mongodb package even if they don't use it. That's bad.

mptr:
@MatheusMeloAntiquera how about not submitting the same changes in separate PRs?
I've added you and @tgrassl to my fork. May we close this PR then?
(Also it would be nice to not commit my changes just by copying them on your branch. Instead pull the commits of collaborators on your branch to keep a clean commit history.)

MatheusMeloAntiquera:
> @MatheusMeloAntiquera how about not submitting the same changes in separate PRs? I've added you and @tgrassl to my fork. May we close this PR then? (Also it would be nice to not commit my changes just by copying them on your branch. Instead pull the commits of collaborators on your branch to keep a clean commit history.)

I will keep this PR, sorry for copy your code I wanted to resolve this quickly, but you should have create a PR to my branch, not directly to typeorm repository

PR #9784, opened 2023-02-13
When calling AVG on an integer column, MS SQL Server returns an int (test only fix)

sbleon:
In this case, 50 instead of 50.5. Apparently the ANSI standard for SQL is silent on this issue, so either behavior should be considered acceptable.

### Description of change

Modify test to accept the answer that SQL Server gives for AVG even though it is mathematically inaccurate.

### Pull-Request Checklist


AlexMesser:
thank you for contribution!

PR #9845, opened 2023-03-11
feat: mariadb uuid inet4 inet6 column data type support

### LINKED ISSUE #8832: Support new UUID type in MariaDB 10.7
## Feature Description
[https://mariadb.com/kb/en/uuid-data-type/](https://mariadb.com/kb/en/uuid-data-type/)
MariaDB starting with 10.7 supports UUID datatype.
INSERT/SELECT just like varchar, but stored in 128-bit wich is better for indexes.
Can you, please, provide ORM support for this type?
 
| DB Type           | Relevant      |

smith-xyz:
### Description of change

Fixes #8832

Adds uuid, inet4, inet6 column data types for appropriate mariadb versions

### Pull-Request Checklist

### DIFF
--- a/src/driver/mysql/MysqlDriver.ts
+++ b/src/driver/mysql/MysqlDriver.ts
@@ -863,8 +863,14 @@ export class MysqlDriver implements Driver {
 
         /**
          * fix https://github.com/typeorm/typeorm/issues/1139
+         * as part of adding uuid support for mariadb, we need to check that we aren't actually using the uuid type
+         * because this should not receive the length value by default
          */
-        if (column.generationStrategy === "uuid") return "36"
+        if (
+            this.columnTypeVersionSupportMap.uuid !== "uuid" &&
+            column.generationStrategy === "uuid"
+        )
+            return "36"
 
         switch (column.type) {
             case String:
--- a/src/driver/mysql/MysqlDriver.ts
+++ b/src/driver/mysql/MysqlDriver.ts
@@ -323,7 +323,6 @@ export class MysqlDriver implements Driver {
         inet6: "varchar",
     }
 
-
     /**
      * Max length allowed by MySQL for aliases.
      * @see https://dev.mysql.com/doc/refman/5.5/en/identifiers.html
@@ -440,21 +439,21 @@ export class MysqlDriver implements Driver {
              * @see https://mariadb.com/kb/en/uuid-data-type/
              */
             if (VersionUtils.isGreaterOrEqual(dbVersion, "10.7.0")) {
-                this.columnTypeVersionSupportMap.uuid = "uuid";
+                this.columnTypeVersionSupportMap.uuid = "uuid"
             }
             /**
              * MariaDb version 10.10.0 supports INET4 type
              * @see https://mariadb.com/kb/en/inet4/
              */
-            if (VersionUtils.isGreaterOrEqual(dbVersion,  "10.10.0")) {
-                this.columnTypeVersionSupportMap.inet4 = "inet4";
+            if (VersionUtils.isGreaterOrEqual(dbVersion, "10.10.0")) {
+                this.columnTypeVersionSupportMap.inet4 = "inet4"
             }
             /**
              * MariaDb version 10.5.0 supports INET6 type
              * @see https://mariadb.com/kb/en/inet6/
              */
             if (VersionUtils.isGreaterOrEqual(dbVersion, "10.5.0")) {
-                this.columnTypeVersionSupportMap.inet6 = "inet6";
+                this.columnTypeVersionSupportMap.inet6 = "inet6"
             }
         } else if (this.options.type === "mysql") {
             if (VersionUtils.isGreaterOrEqual(dbVersion, "8.0.0")) {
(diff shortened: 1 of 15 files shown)


src/driver/mysql/MysqlDriver.ts
```diff
@@ -310,6 +314,15 @@ export class MysqlDriver implements Driver {
         bigint: { width: 20 },
     }
 
+    /**
+     * Managing version specific data types
+     */
+    columnTypeVersionSupportMap = {
```
pleerock, line 320: we don't want to have such feature. If user uses Mariadb version which support all these new types - he can use it. If no - he can't. He needs to put varchar or any other type he thinks is appropriate by himself.

smith-xyz: Thank you for the feedback.

Sounds appropriate to me. Just so I understand, in the case where a user provides "inet4" but they're on mysql or a version of mariadb that doesn't yet support the data type, let the user receive an error back from the driver?

pleerock: I'm okay with an underlying DBMS throwing the error instead.

C0kkie:
Would be cool if we can finally get this

src/driver/mysql/MysqlDriver.ts
smith-xyz: Sounds good. I did leave the defaulting behavior that was their for mysql for the uuid type since that seems to have some breaking changes. Let me know if this was in error. Thank you again!

C0kkie:
@smith-xyz Maybe you could add the same way the new supported json validation too?

smith-xyz:
> @smith-xyz Maybe you could add the same way the new supported json validation too?

I'm not sure I'm following, but it does sound a bit out of scope here - I'd recommend a new issue for that one. Unless I'm missing something.

smith-xyz:
> @smith-xyz Maybe you could add the same way the new supported json validation too?

Is this in terms of the "JSON" type from mariadb, added in v10.7.2? If so, I think that's a great idea but I'd suggest that one to be a separate enhancement from this one - I would be willing to work on it. Main reason is that I would want to explore how to create the CHECK (JSON_VALID()) constraint separate from this one.

C0kkie:
Yes it got added in 10.7.2 i think

src/driver/mysql/MysqlDriver.ts
```diff
@@ -825,8 +829,10 @@ export class MysqlDriver implements Driver {
 
         /**
          * fix https://github.com/typeorm/typeorm/issues/1139
+         * note that if the db did support uuid column type it wouldn't have been defaulted to varchar
          */
-        if (column.generationStrategy === "uuid") return "36"
+        if (column.generationStrategy === "uuid" && column.type !== "uuid")
```
pleerock, line 834: correct me if I'm wrong, but from what I see for `mysql` generationStrategy is `uuid` and column type is `uuid` as well. But we don't return `36` in such case now, which is different from what we had before? If yes, this can cause issues for those who is using mysql right now.

test/github-issues/6540/issue-6540.ts
```diff
@@ -38,6 +38,10 @@ describe("github issues > #6540 Enum not resolved if it is from an external file
                 const sqlInMemory = await connection.driver
                     .createSchemaBuilder()
                     .log()
+                console.log(
```
pleerock, line 41: looks like redundant console.log

src/driver/mysql/MysqlDriver.ts
smith-xyz: that's correct I believe. It looks like before if the generationStrategy was "uuid" it would default the column type from "uuid" to "varchar". Then here if no length was attributed to the column, it would plug in "36" since mysql needs a length for the varchar.

Yes, in returning I see your point this does cause issues for existing users that need the column type to remain as varchar given that it use to be defaulted to varchar. I'll have to rethink this one. Let me know if you have any suggestions.

smith-xyz: @pleerock my thoughts are that maybe PrimaryGeneratedColumnUUIDOptions should probably add one more property that allows for configuring the actual column type to be used so that this logic isn't tied up in the driver itself.

pleerock: Better to save the scope of this PR only to mariadb. So we just need to add proper checks I avoid any breaking changes for mysql users.

smith-xyz: yes good point.

For this, maybe it makes more sense to do this conditional to make it clear of the exact case its needed:
`
if (column.generationStrategy === "uuid" && column.type === "varchar") return 36
`

the reason its needed is because the generation strategy being uuid would default to varchar column type when normalized. Because of that when we get the column length it would wrongly do a length of 255 by default, but should be 36 in this case.

Other than that at line 727 the else if will check if the column type is "uuid" and if its a mysql driver it will default to varchar.

Let me know if that sounds correct and I can make the updates.

pleerock: yes, sounds good. But please check again everything before it will be merged, just to make sure we didn't break user use cases.

test/github-issues/6540/issue-6540.ts
smith-xyz: thanks, removed it

src/driver/mysql/MysqlDriver.ts
smith-xyz: added a private property for seeing if the uuid data type can be used. Also modified the conditional as mentioned. Added a regression test for mysql to the test suite. Let me know if anything else is needed.

C0kkie: Could we get a merge? @smith-xyz

C0kkie: Could we get a merge? @pleerock

pleerock:
Thank you for contribution! 🎉

C0kkie:
Yes thanks @smith-xyz  :) @pleerock  would it be possible to get a new version?

Issue #9118, opened 2022-06-16
ES2022 Repository create method include undefined field that breaks DB update


mbagsik00:
## Issue Description
`repository.create()` includes the undefined field which breaks the DB update query when using ES2022 as a target in `tsconfig`

### Expected Behavior
Previously we're using es6 as a target for `tsconfig` and the field that we didn't pass in the `repository.create()` will not be included in the response.

```ts
// Entity
class TestEntity {
  @PrimaryGeneratedColumn('uuid')
  id: string;
  
  @Column('varchar')
  name: string;
}

// Repository Create
const result = Repository.create({
   name: 'test name'
})

// Result when using ES6
{
 name: 'test name'
}
```

### Actual Behavior
When using ES2022 as a target in `tsconfig` it returns all the fields that are declared in the entity and sets the value to `undefined` for the fields that don't exist in the object when creating the repository. This causes some issues when doing an update since it sets other fields to `null` if we don't include the field in the repository creation. I know we can easily fix this by adding all the fields when creating the repository but didn't expect this to break.

```ts
// Entity
class TestEntity {
  @PrimaryGeneratedColumn('uuid')
  id: string;
  
  @Column('varchar')
  name: string;
}

// Repository Create
const result = Repository.create({
   name: 'test name'
})

// Result when using ES2022
{
 id: undefined,
 name: 'test name'
}
```

### Steps to Reproduce

1. Create a repository using only some of the fields from the entity.
2. Result returns the other fields from the entity that haven't been declared when creating the repository with an `undefined` value.

### My Environment
* NestJS

| Dependency          | Version  |
| ---                 | ---      |
| Operating System    |   MacOS       |
| Node.js version     | 16.15.0  |
| Typescript version  | 4.6.2  | 
| TypeORM version     | 0.2.25 | 

### Additional Context
* I'm not sure if ES2022 is supported or not yet.

### Relevant Database Driver(s)

| DB Type           | Reproducible  |
|-------------------| ---           |
| `aurora-mysql`    | no            |
| `aurora-postgres` | no            |
| `better-sqlite3`  | no            |
| `cockroachdb`     | no            |
| `cordova`         | no            |
| `expo`            | no            |
| `mongodb`         | no            |
| `mysql`           | no            |
| `nativescript`    | no            |
| `oracle`          | no            |
| `postgres`        | yes           |
| `react-native`    | no            |
| `sap`             | no            |
| `spanner`         | no            |
| `sqlite`          | no            |
[...]

PR #9884, opened 2023-03-23
fix: Resolve undefined fields created when in ES2022 mode

andrewnester:
Closes: #9118

### Description of change

When compiling in ES2022 mode, Typescript option `useDefineForClassFields` is set to `true`. It leads to fields with undefined values being generated in entity metadata.

This changes fixes it by cleaning up any undefined fields in metadata so undefined fields are not generated and behaviour is the same when `useDefineForClassFields` is set to either `true` or `false`

### Pull-Request Checklist

### DIFF
--- a/src/metadata/EntityMetadata.ts
+++ b/src/metadata/EntityMetadata.ts
@@ -569,6 +569,12 @@ export class EntityMetadata {
             ret = {}
         }
 
+        // Cleaning up any undefined keys to make it work consistent
+        // with TS useDefineForClassFields option being true or false.
+        Object.keys(ret).forEach(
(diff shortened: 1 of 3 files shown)


pleerock:
Proposed changes just make `.create` work differently. What if user just uses `new UserEntity()` ? I doesn't solve the problem plus brings a confusion on why `.create` works differently.

I don't believe we can do something with this issue.
I guess this is expected behavior now considering ES2022 standard.
If you use classes you kind live with undefined properties now (which is okay IMO, it's just something that we aren't used to... yet).
Alternative is not to use classes and use entity schemas instead.

PR #9887, opened 2023-03-25
fix: Convert the join table ID to the referenceColumn ID type

tqh177:
### Description of change

When I use auto-incremented id as a string and call findOne with the "loadRelationIds:true" option, I get an empty array for the "relationIds" property, even though the relation exists.

### Pull-Request Checklist


pleerock:
Please add the test so we can confirm where the problem is.

pleerock:
This one is a bit risky to merge, but if I don't miss anything - change looks correct. @tqh177 I'll merge it, but please keep watching the related issues - in case if somebody raise the bug regarding to these changes. Thanks for contribution.

Issue #8975, opened 2022-05-07
Create Project via npx failed for sqlite


kaigothe:
### Expected Behavior

I tried to create a project via `npx typeorm init --name MyProject2 --database sqlite`.  I tested the command for mysql, that works.

### Actual Behavior
I got following Error: 

```
/root@test:~# npx typeorm init --name MyProject2 --database sqlite
Project created inside /root/MyProject2 directory.
Please wait, installing dependencies...
Error during project initialization:
npm ERR! code ERESOLVE
npm ERR! ERESOLVE could not resolve
npm ERR! 
npm ERR! While resolving: typeorm@0.3.6
npm ERR! Found: sqlite3@4.2.0
npm ERR! node_modules/sqlite3
npm ERR!   sqlite3@"^4.0.3" from the root project
npm ERR! 
npm ERR! Could not resolve dependency:
npm ERR! peerOptional sqlite3@"^5.0.2" from typeorm@0.3.6
npm ERR! node_modules/typeorm
npm ERR!   typeorm@"0.3.6" from the root project
npm ERR! 
npm ERR! Conflicting peer dependency: sqlite3@5.0.8
npm ERR! node_modules/sqlite3
npm ERR!   peerOptional sqlite3@"^5.0.2" from typeorm@0.3.6
npm ERR!   node_modules/typeorm
npm ERR!     typeorm@"0.3.6" from the root project
npm ERR! 
npm ERR! Fix the upstream dependency conflict, or retry
npm ERR! this command with --force, or --legacy-peer-deps
npm ERR! to accept an incorrect (and potentially broken) dependency resolution.
npm ERR! 
npm ERR! See /root/.npm/eresolve-report.txt for a full report.

npm ERR! A complete log of this run can be found in:
npm ERR!     /root/.npm/_logs/2022-05-07T12_46_51_752Z-debug-0.log
```

### My Environment

| Dependency          | Version  |
| ---                 | ---      |
| Operating System    |     Ubuntu 20.04 LTS  (fresh install as VM)  |
| Node.js version     | v16.15.0  | 
| npm version  | 8.5.5 | 
| TypeORM version     | x.y.zzz  | 

### Relevant Database Driver(s)

| DB Type           | Reproducible  |
|-------------------| ---           |
| `aurora-mysql`    | no            |
| `aurora-postgres` | no            |
| `better-sqlite3`  | no            |
| `cockroachdb`     | no            |
| `cordova`         | no            |
| `expo`            | no            |
[...]

PR #9926, opened 2023-04-07
test: cli init with local typeorm package

mptr:
### Description of change

Fixes Test associated to #8975

Uses the locally built files instead of the typeorm package published to npmjs.
This enables testing project initialisation with unreleased typeorm package versions.

The use of the local version targets [this Line](https://github.com/typeorm/typeorm/blob/daf1b47a609cb5090b71e537668ee115a98b1dc9/src/commands/InitCommand.ts#L698) in the cli code.

Example diff between the package.json bootstrapped by the test suite before and after the change:
```diff
@@ -9,7 +9,7 @@
       "typescript": "4.5.2"
    },
    "dependencies": {
-      "typeorm": "0.3.13",
+      "typeorm": "0.0.0",
       "reflect-metadata": "^0.1.13",
       "mysql": "^2.14.1"
    },
+   "installFrom": "file:../build/compiled/src"
```

### Pull-Request Checklist


pleerock:
mysql tests seems to fail...

src/commands/InitCommand.ts
```diff
@@ -695,7 +695,10 @@ Steps to run this project:
 
         if (!packageJson.dependencies) packageJson.dependencies = {}
         Object.assign(packageJson.dependencies, {
-            typeorm: require("../package.json").version,
+            typeorm:
+                require("../package.json").version !== "0.0.0"
+                    ? require("../package.json").version // install version from package.json if present
+                    : require("../package.json").installFrom, // else use custom source
```
mptr, line 701: The tests fail because the implemented "fix" did not comply to the [package.json specs](https://docs.npmjs.com/cli/v7/configuring-npm/package-json#version:~:text=Version%20must%20be%20parseable%20by%20node%2Dsemver) which require the version to be parseable by semver (format x.x.x).
For some reason node v16 just did not care but v14 does.
The alternative would be to change the cli source code a bit to make detectable wether running in test mode or not.
I see no other solution for testing the desired behavoir with 100% code coverage.

pleerock:
@mptr we can actually drop node 14 from CI. Node 14 maintenance state (by node.js team) will expire this month. So we actually don't need to support it anymore at all.

Regarding to Oracle tests -> yes we currently have it failed everywhere, it's not caused by your changes -> can you help us to fix this issue if you have a time? I'm okay if you put waiting logic for oracle before we run the tests.

mptr:
@pleerock But the constraint of version being parseable by semver still exists in [the latest specs](https://docs.npmjs.com/cli/v9/configuring-npm/package-json?v=true). Therefore i'd consider Node v16 ignoring it as a bug. Should we still revert to the version trick to avoid application code change?

pleerock:
That's fine, let's merge it. Thank you for investigations!

PR #9930, opened 2023-04-09
fix: drop xml & yml connection option support

pleerock:
Dropped orm configuration loading from `yml` and `xml` files. 

The decision to drop these features was quite long ago, and now looks like it's time to drop it because one of the dependent packages bring us security issues.

You can read more about security issue [here](https://www.cve.org/CVERecord?id=CVE-2023-0842).

Both `xml2js` and `js-yaml` packages are not maintained, so we can't really wait for the security issues to resolve from their end.

For users who used this feature - it's strongly recommended to switch to regular configurations defined in JSON / Plain JavaScript Object or implement configuration loading based on previous code using alternative package on their own.

BREAKING CHANGE: `yml` and `xml` support in configuration loading.


imnotjames:
@pleerock why was the package lockfile deleted?  Usually the lockfile is needed to ensure the dependencies retrieved are the same when building the package for development purposes.

It also breaks the CI caching & preview package build CI.

Technically this should also break the CI installing packages entirely but we're using `npm install` in our package installation to work around https://github.com/npm/cli/issues/558

pleerock:
@imnotjames ooups I wanted to re-install everything and commit it again. Thanks for notice, I'll commit it again.

pleerock:
https://github.com/typeorm/typeorm/commit/115059de9147df2b4d65921ae400279c9ca4ee8f

imnotjames:
Thanks!

robotnic:
Is it possible to make a 0.2.x version without xml2js. 

In 0.2.45
                       === npm audit security report ===                        

# Run  npm install typeorm@0.3.14  to resolve 1 vulnerability
  High            xml2js is vulnerable to prototype pollution                   
  Package         xml2js                                                        
  Dependency of   typeorm                                                       
  Path            typeorm > xml2js                                              
  More info       https://github.com/advisories/GHSA-776f-qx25-q3cc

PR #9942, opened 2023-04-12
fix: make cache optional fields optional

RustySol:
### Description of change
This patch allows the `getFromCache` caller to only pass in `identifier` or `query` field, as `getFromCache`only needs **either, or neither** `identifier` or `query` to be present.

Currently as a caller of `getFromCache`, one is required to pass both `identifier` and `query` value, which is unnecessary because only one will be used for execution, with `identifier` taking precedence.

### Pull-Request Checklist

### DIFF
--- a/src/cache/DbQueryResultCache.ts
+++ b/src/cache/DbQueryResultCache.ts
@@ -131,7 +131,7 @@ export class DbQueryResultCache implements QueryResultCache {
     }
 
     /**
-     * Caches given query result.
+     * Get data from cache.
      * Returns cache result if found.
      * Returns undefined if result is not cached.
      */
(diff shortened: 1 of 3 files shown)


pleerock:
Thank you for contribution! 🎉

elderapo:
This update introduced type-safety regression. It allows passing neither `identifier` nor `query`. I think the type should look like:
```ts
type QueryResultCacheOptions = {
  /**
   * Time, when cache was created.
   */
  time?: number

  /**
   * Duration in milliseconds during which results will be returned from cache.
   */
  duration: number

  /**
   * Query result that will be cached.
   */
  result?: any
} & ({
  /**
   * Cache identifier set by user.
   */
  identifier: string
} | {
  /**
   * Cached query.
   */
  query: string
});
```

This way options ALWAYS have `identifier` or `query` set. The only downside of this approach is that you have to use `in` operator to check which option was set:
```ts
const takeQueryResultCacheOptions = (options: QueryResultCacheOptions) => {
  if ("identifier" in options) {

    options.identifier.toUpperCase(); // no error

    return;
  }

  if ("query" in options) {

    options.query.toLowerCase(); // no error

    return;
  }

  options; // never 
}
```

PR #9957, opened 2023-04-18
query methods should accept generic for return type

Newbie012:
### Description of change

the query methods can now accept a generic type that is used as the return type. If no generic were passed, it will default to any (the current behavior). This way it allows the consumer of the library to decide what should be the return type without type casting (`as`). 

I wrote an ESLint plugin ([SafeQL](https://github.com/ts-safeql/safeql)) that helps validate the return type of a raw query based on the generic type, which is the main reason I raised this PR. Currently, type augmentation is the only way to integrate TypeORM with SafeQL which hurts DX (which I'm trying to avoid) 🙂.

### Pull-Request Checklist

### DIFF
--- a/src/data-source/DataSource.ts
+++ b/src/data-source/DataSource.ts
@@ -515,11 +515,11 @@ export class DataSource {
     /**
      * Executes raw SQL query and returns raw database results.
      */
-    async query(
+    async query<T = any>(
         query: string,
         parameters?: any[],
         queryRunner?: QueryRunner,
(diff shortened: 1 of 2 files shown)


pleerock:
I think you need to grant some permissions to make CI to pass the tests on your branch.

djvidov (approved):

pleerock:
okay changes should be safe anyway, let's merge it and check the changes on the master.

Issue #3531, opened 2019-01-28
Entity name does not support 'dot'!


zurez-sgig:
**Issue type:**

[ ] question
[ x] bug report
[ ] feature request
[ ] documentation issue

**Database system/driver:**

[ ] `cordova`
[ ] `mongodb`
[ x] `mssql`
[ ] `mysql` / `mariadb`
[ ] `oracle`
[ ] `postgres`
[ ] `sqlite`
[ ] `sqljs`
[ ] `react-native`
[ ] `expo`

**TypeORM version:**

[ ] `latest`
[ ] `@next`
[x ] `0.2.12` (or put your version here)

**Steps to reproduce or a small repository showing the problem:**
Create an entity for a table which has dot in its name. Eg: 'foo.bar'
`@Entity({
 name:'foo.bar',
schema:'dbo'
})
class FooBar{
 @PrimaryColumn()
  id:number
}
`
Results in throwing **QueryFailedError: Error: Invalid object name **
<!--
    To answer those questions you need to put "x" inside the square brackets, for example:
    [x] `mysql`
    [ ] `postgres`

    Also, please format your code properly (by taking code blocks into ```ts .... ```)
!>

sroye-skillgigs:
Any response on this bug?

sroye-skillgigs:
The problem that is occurring is that the SQL that is being generated is calling `FROM "foo"."bar"` instead of `"foo.bar"`. Is this a something that can be fixed?

lakhansamani:
@sroye-skillgigs I was having a similar situation, but it worked for me. I am using the latest version of `typeorm`.

my example

*FileName* `dbo.text.ts`
```
import { Entity, Column, PrimaryColumn } from "typeorm";

@Entity({ name: "dbo.text " })
class Text {
	@PrimaryColumn()
	public id: number;

	@Column({ type: "ntext" })
	public info: string;

	@Column()
	public isValid: boolean;
}

export default Text;
```

Brian611:
Any fix yet for this?

Timtam:
I've got a similar issue, but not with MSSQL, but with SAP Hana instead. The issue however is the same. All the different parts of the schema/database table get quoted and thus are invalid.
Correct would be: schema.first.part.of.database.table
Current situation is: "schema"."first"."part"."of"."database"."table"
which results in a syntax error when doing anything with this table, no matter if its selecting, inserting or whatever.
Any update on this?

LeoFeilke:
[...]

PR #10004, opened 2023-05-01
fix: resolve issue building eager relation alias for nested relations

### LINKED ISSUE #9944: Alias Issue With Nested Entity And Relations
### Issue description

Querying a nested entity which in turn has deeply nested relationships (eager) results in error (since 0.3.13)

### Expected Behavior

Consider the following example entities

- A has a **nested entity** B
- B has a *OneToMany* Relationship To C (eager)
- C has a *OneToMany* Relationship To D (eager)

Querying A should resolve all relations just fine without any error.

### Actual Behavior

Querying A results in the following error: `TypeORMError: "A__b" alias was not found. Maybe you forgot to join it?`

### Steps to reproduce

1. Clone the [reproduction repo](https://github.com/giz-berlin/typeorm-nested-entity-relation-repo)
2. Follow the steps in the README there and see the error code.

### My Environment

| Operating System    |    macOS 13.3    |
| Node.js version     | 19.8.1  |
| TypeORM version     | ^0.3.13  |

### Additional Context

This issue does **not** occur with `typeorm < 0.3.13`, so the issue has been introduced since 0.3.12.
It also does not happen when only having one level of relations (i.e. without entity `D` it works).

First investigations did show, that a11809e could be the culprit (especially <https://github.com/typeorm/typeorm/blame/master/src/find-options/FindOptionsUtils.ts#LL390-L395>) which was merged in #9680.

I only tested this with postgres.

If desired, I could provide the reproduction as a test but I am not sure how to fix this.

- [ ] aurora-mysql
- [ ] aurora-postgres
- [ ] better-sqlite3
- [ ] cockroachdb
- [ ] cordova
- [ ] expo
- [ ] mongodb
- [ ] mysql
- [ ] nativescript
- [ ] oracle
- [X] postgres
- [ ] react-native
- [ ] sap
- [ ] spanner
- [ ] sqlite
- [ ] sqlite-abstract
- [ ] sqljs
- [ ] sqlserver

JoseCToscano:
Fix alias error for nested relations. When loading eager relations from nested
levels only the first level was handled through the alias builder

Closes: #9944

### Description of change
When loading eager relations from nested levels only the first level was handled through the alias builder
resulting in the following error `TypeORMError: "A__b" alias was not found. Maybe you forgot to join it?`.
By appropiately handling deeply nested relations, just as any other relations they are able to be fetched as usual

Fixes #9944

### Pull-Request Checklist

### DIFF
--- a/src/driver/DriverUtils.ts
+++ b/src/driver/DriverUtils.ts
@@ -127,14 +127,16 @@ export class DriverUtils {
         buildOptions: { shorten?: boolean; joiner?: string } | undefined,
         ...alias: string[]
     ): string {
-        const newAlias =
-            alias.length === 1
-                ? alias[0]
-                : alias.join(
-                      buildOptions && buildOptions.joiner
-                          ? buildOptions.joiner
-                          : "_",
-                  )
+        const joiner =
+            buildOptions && buildOptions.joiner ? buildOptions.joiner : "_"
+
+        let newAlias = alias.length === 1 ? alias[0] : alias.join(joiner)
+
+        const hasNestedRelations = newAlias.includes(".")
+        if (hasNestedRelations) {
+            newAlias = newAlias.replace(/\./g, joiner)
+        }
+
         if (
             maxAliasLength &&
             maxAliasLength > 0 &&
--- a/src/driver/DriverUtils.ts
+++ b/src/driver/DriverUtils.ts
@@ -132,11 +132,6 @@ export class DriverUtils {
 
         let newAlias = alias.length === 1 ? alias[0] : alias.join(joiner)
 
-        const hasNestedRelations = newAlias.includes(".")
-        if (hasNestedRelations) {
-            newAlias = newAlias.replace(/\./g, joiner)
-        }
-
         if (
             maxAliasLength &&
             maxAliasLength > 0 &&
--- a/src/find-options/FindOptionsUtils.ts
+++ b/src/find-options/FindOptionsUtils.ts
@@ -391,7 +391,7 @@ export class FindOptionsUtils {
                 qb.connection.driver,
                 { joiner: "__" },
                 alias,
-                relation.propertyPath,
+                relation.propertyName,
             )
 
             // add a join for the relation
--- a/src/naming-strategy/NamingStrategyInterface.ts
+++ b/src/naming-strategy/NamingStrategyInterface.ts
@@ -163,11 +163,6 @@ export interface NamingStrategyInterface {
     prefixTableName(prefix: string, tableName: string): string
 
     /**
-     * Gets the name of the alias used for relation joins.
-     */
-    eagerJoinRelationAlias(alias: string, propertyPath: string): string
-
-    /**
      * Column names for nested sets.
      */
     nestedSetColumnNames: { left: string; right: string }
--- a/test/github-issues/9944/issue-9944.ts
+++ b/test/github-issues/9944/issue-9944.ts
@@ -1,0 +1,42 @@
+import "reflect-metadata"
+import {
+    createTestingConnections,
+    closeTestingConnections,
+    reloadTestingDatabases,
+} from "../../utils/test-utils"
+import { DataSource } from "../../../src/data-source/DataSource"
+import { expect } from "chai"
+import { A } from "./entity/A"
+import { B } from "./entity/B"
+import { C } from "./entity/C"
+
+describe("github issues > #9944 Alias Issue With Nested Entity And Relations", () => {
+    let dataSources: DataSource[]
+    before(
+        async () =>
+            (dataSources = await createTestingConnections({
+                entities: [__dirname + "/entity/*{.js,.ts}"],
+                schemaCreate: true,
+                dropSchema: true,
+                relationLoadStrategy: "query",
+                enabledDrivers: ["mysql"],
+            })),
+    )
+    beforeEach(() => reloadTestingDatabases(dataSources))
+    after(() => closeTestingConnections(dataSources))
+
+    it("Validate correct loading of eager, nested relations", () =>
+        Promise.all(
+            dataSources.map(async (dataSource) => {
+                const aEntity = new A()
+                aEntity.b = new B()
+                aEntity.b.cs = [new C()]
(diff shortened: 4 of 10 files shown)


src/naming-strategy/NamingStrategyInterface.ts
```diff
@@ -162,11 +162,6 @@ export interface NamingStrategyInterface {
      */
     prefixTableName(prefix: string, tableName: string): string
 
-    /**
-     * Gets the name of the alias used for relation joins.
-     */
-    eagerJoinRelationAlias(alias: string, propertyPath: string): string
```
JoseCToscano, line 168: This function was not being used at all, and just brings noise when trying to extend the interface

test/github-issues/9944/issue-9944.ts
```diff
@@ -0,0 +1,42 @@
...
+            dataSources.map(async (dataSource) => {
+                const aEntity = new A()
+                aEntity.b = new B()
+                aEntity.b.cs = [new C()]
+                await dataSource.manager.save(aEntity)
+
+                const [fetchedA] = await dataSource.manager.find(A)
+                expect(fetchedA.b).exist // Validates correct relation A.b is loaded
+                expect(fetchedA.b.cs).exist // Validates correct relation A.b.cs is loaded
+                expect(fetchedA.b.cs[0]).exist
```
JoseCToscano, line 39: Before the fix, running this test resulted in the `TypeORMError: "A__b" alias was not found. Maybe you forgot to join it?`

With the fix it runs perfectly. I jut added these three assertions to make sure that data from nested relations was indeed loaded correctly

src/driver/DriverUtils.ts
```diff
@@ -127,14 +127,16 @@ export class DriverUtils {
...
-                      buildOptions && buildOptions.joiner
-                          ? buildOptions.joiner
-                          : "_",
-                  )
+        const joiner =
+            buildOptions && buildOptions.joiner ? buildOptions.joiner : "_"
+
+        let newAlias = alias.length === 1 ? alias[0] : alias.join(joiner)
+
+        const hasNestedRelations = newAlias.includes(".")
```
felix-gohla, line 135: Is there some kind of constant that defines this as the relations joiner in typeorm? I've skimmed through the repo and didn't find one. Maybe you have more insight. 😊

src/driver/DriverUtils.ts
```diff
@@ -127,14 +127,16 @@ export class DriverUtils {
...
+        if (hasNestedRelations) {
+            newAlias = newAlias.replace(/\./g, joiner)
```
felix-gohla, line 137: Are there some edge cases like dots in table names that would be affected by this? At least in postgres, I am able to create tables with dots and then use double quote marks to show that this is a name.

felix-gohla (commented):
Apart from the nits: Works! Thank you for fixing that issue.

src/driver/DriverUtils.ts
JoseCToscano: That does sound like some edge case that could bring up some issue 🤔 . Haven't really worked with it. How does postgress handles the syntax for nested relations?

src/driver/DriverUtils.ts
JoseCToscano: That's some nice insight. I do think it should be defined somewhere in the repo, otherwise, where is it taking it from?. I'll see if I can find it, maybe we can work the solution through there 👀

felix-gohla: I think there are many places in the code which just assumes that there are no dots in the name of tables.
Maybe worth another issue. There is already #3531. Maybe there somewhere hides a function that escapes names or something. However, not sure if this is out of scope for this PR.

JoseCToscano: Just made an update to the PR. By using relationName instead of relationPath, we do not need to parse for any "." to get the relations. So we won't encounter issues with the table names. (At least not in this section, it seems the full support for them is still missing in the repo😅 )

src/find-options/FindOptionsUtils.ts
```diff
@@ -391,7 +391,7 @@ export class FindOptionsUtils {
                 qb.connection.driver,
                 { joiner: "__" },
                 alias,
-                relation.propertyPath,
+                relation.propertyName,
```
JoseCToscano, line 394: @felix-gohla 
This was the main issue, breaking nested relations. There was not really the need of  parsing "." to distinguish relations

src/driver/DriverUtils.ts
JoseCToscano: Just removed this section that previously added, it was not needed at all

src/find-options/FindOptionsUtils.ts
felix-gohla: This is great! 😊 Simple solutions are always the best!

felix-gohla (approved):
I also tested this version against our project's internal test suite and everything seems fine back again. Thank you very much for investigating, @JoseCToscano! 😊 

Maybe rebase onto the latest master and then @pleerock or one of the other maintainers can have a look at this.

felix-gohla:
@pleerock @AlexMesser Is there something we can do to support you? Issues and MRs are somehow piling up... 🙃

JoseCToscano:
Just synced master

felix-gohla:
@JoseCToscano Did you per chance find a possibility to use typeorm directly from source (to include the patch without waiting for reaction of the maintainers and a release)?
We could use some kind of post-install script to build typeorm but this is not an ideal situation... :D

JoseCToscano:
> 

I've been testing locally TypeORM by packaging a local version of TypeORM and then installing it on my Project's repo (just ``` npm install '../../../../typeom-fork/typeorm/build/package' ```
but I don't know if this is what you are looking for. Maybe publishing a forked version to your own npm organization/project and installing that directly?

hinogi (approved):
LGTM

pleerock:
@JoseCToscano  thanks for the fix and sorry for late review. The summer was really hot 🌞 . Thanks @felix-gohla @hinogi  for the review help!

PR #10011, opened 2023-05-03
fix: ISO date parsing

### LINKED ISSUE #9941: ISO DateTime parsing returning wrong Date value
In recent release it looks like Date-FNS package was removed and TypeOrm is doing it's own parsing but the calculation for calculation is problematic. 
In src/util/DateUtils.ts this method was added to replace date-fns.parseISO

```
 private static parseDateAsISO(dateString: string): Date {
        const date = new Date(dateString)
        const offset = date.getTimezoneOffset() * 60 * 1000
        const utc = date.getTime() + offset

        return new Date(utc)
    }
```
However the above code returns the datetime with wrong values.
Example:
Given this string: '2023-04-11T16:56:57.572Z'
The above function returns: '2023-04-11T23:56:57.572Z' depending on what timezone your environment is configured for, this example it is -07:00.
Where as date-fns.parseISO() returns: '2023-04-11T16:56:57.572Z' which seems correct.

felix-gohla:
### Description of change

Fix date parsing ignoring time zones.
In the code there's a explanation with link to the ECMA-script specification on why that happens.
The error was introduced when removing `date-fns` from typeorm in #9634.

### Discussion

**However:** I wonder if it is even intended to deviate from the standard (even if the distinction between date and date-times seems a little bit... odd). Currently, the parse function always assumes local time unless there's a timezone given. This could be confusing.

During testing I also noticed that there are some tests that relied on another date format.
Maybe it would be wise to decide which date standard to go for (RFC3339 or ISO8601, see [this website for comparison](https://ijmacd.github.io/rfc3339-iso8601/)).
Furthermore, not sure if it makes sense to handle this by code in typeorm itself.
There are other people who really already invested time into figuring dates out in JS.

Fixes #9941

### Pull-Request Checklist

### DIFF
--- a/src/util/DateUtils.ts
+++ b/src/util/DateUtils.ts
@@ -48,7 +48,7 @@ export class DateUtils {
     ): Date {
         let date =
             typeof mixedDate === "string"
-                ? DateUtils.parseDateAsISO(mixedDate)
+                ? DateUtils.parseDateAsISOLocal(mixedDate)
                 : mixedDate
 
         if (toUtc)
@@ -268,9 +268,11 @@ export class DateUtils {
     }
 
     /**
-     * Parse a date without the UTC-offset of the device
+     * Parse a ISO 8601 date(-time) string taking into account the local timezone (if none is given).
      *
-     * The problem here is with wrong timezone.
+     * The problem here is with wrong timezone and ECMA-Script being weird about when to use time zones and when not.
+     *
+     * **Caution:** The function expects an ISO8601 date(-time) string as input.
      *
      * For example:
      *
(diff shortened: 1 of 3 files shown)


src/util/DateUtils.ts
```diff
@@ -268,21 +268,37 @@ export class DateUtils {
...
+            isoDateString = `${isoDateString}T00:00:00`
+        }
+        const parsed = new Date(isoDateString)
+        console.log(
+            "isIsoDataTime",
+            isDateTime,
+            isoDateString,
+            parsed,
+            parsed.toISOString(),
+        )
```
henrinormak, line 301:
```suggestion
```

src/util/DateUtils.ts
```diff
@@ -268,21 +268,37 @@ export class DateUtils {
...
-        return new Date(utc)
+    private static parseDateAsISOLocal(isoDateString: string): Date {
+        // From ECMA-262 specification:
+        // "The String may be interpreted as a local time, a UTC time, or a time in some other time zone, depending on the contents of the String. [...]
+        //  When the UTC offset representation is absent, date-only forms are interpreted as a UTC time and date-time forms are interpreted as a local time."
+        // See https://tc39.es/ecma262/#sec-date-time-string-format for further information
+
+        const isDateTime = isoDateString.includes("T")
+        if (!isDateTime) {
+            // This lets the date parser now that the date is to be treat as local time.
```
henrinormak, line 291:
```suggestion
            // This lets the date parser know that the date is to be treat as local time.
```

henrinormak (approved):

src/util/DateUtils.ts
felix-gohla: Sorry, how did I not catch that? ^^

henrinormak:
Not a maintainer of the repo so can't help with getting this merged, but really looking forward to the fix for the issue this is targetting (as it is affecting me as well).

pleerock:
thanks @felix-gohla for taking a look on this one. Tests are failing however...

felix-gohla:
I appreciate your response. I have refrained from fixing them until now because there are some discussion parts (see PR description). 😊
As a maintainer, maybe you could give your opinion.

If you say that typeorm should handle all the cases itself (without external library), then I can continue and fix the tests.
However, there are more edge cases on date handling, IMHO. This could make the code more complex (which you didn't want, see https://github.com/typeorm/typeorm/pull/9634#issuecomment-1355126292).

PR #10014, opened 2023-05-03
fix: allow virtual columns to be selected and deselected so that they…

csechrist:
### Description of change

Changes the way that the virtual columns are treated, and adds another check so that they can be ignored in the select query builder.  Without it, you will get an error that says the column does not exist (which, it technically doesn't), but there are several use cases where I do not want to have the virtual columns loaded, causing me to have to try to find other options for not loading them.

### Pull-Request Checklist

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "typeorm",
-  "version": "0.3.14",
+  "version": "0.3.15",
   "lockfileVersion": 2,
   "requires": true,
   "packages": {
     "": {
       "name": "typeorm",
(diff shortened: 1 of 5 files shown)


pleerock:
Why do you have `package-lock.json` changed? Can you skip it?

Test is failing.

Also, can you explain how what you are trying to achieve work with the changes in `buildRelations` method?

hinogi:
@csechrist can you remove the package-lock from the PR and maybe respond to the comments.

pleerock:
Also it's unclear to me how test solves what you describe. Closing as there is no answer from author.

Issue #1366, opened 2017-12-19
No api to use sql files in migrations and migrations fail with string template literals


jsmonkey:
Hi,

I use typeorm migrations. There is no any api to load sql queries from sql files.
So I have to use js string template literals. But it converts formatted sql with `\n` new line symbols.

Even after replacing `\n` symbols result sql triggers syntax errors and queries fail.

Fix asap.

Thanks

jsmonkey:
OMG,

guys, when I use several times `CREATE TABLE` command in my initial migration it fails cuz all migrations are covered with transactions. In mysql for ddl it's restricted to use several creations in single transaction.

Fix this case in your library pls.

pleerock:
you can use `queryRunner.connection.manager.query()` and execute query without migration

jsmonkey:
That's it, I want to use migration api, cuz it's right way to make changes for the database structure and default data.
Btw this part of code also fails on multiple insertions:
```ts
 async up(queryRunner: QueryRunner): Promise<any> {
    const defaultValues = `
        INSERT INTO sometable (val) VALUES ('val1');
        INSERT INTO sometable (val) VALUES ('val2');
        INSERT INTO sometable (val) VALUES ('val3');
     `;
    await queryRunner.query(defaultValues);
  }
```

That's weird cuz evenv with `START TRANSACTION` and `COMMIT` commands such sequence of insert command runs well from the editor of sql client.

pleerock:
> That's it, I want to use migration api, 

You can commit your opened transaction, do what you want, then open transaction again in your code. There are multiple ways to resolve your issue.

> Btw this part of code also fails on multiple insertions:

fails with errors or what? What if you execute each query separately?

jsmonkey:
If I commit several times in one migration, and smth goes wrong after first transaction then I won't be able to rollback the all changes of migration step. That's the point why I don't want to use several transactions in one migration step.  

`INSERT` queries fail with syntax error at the beginning of the second `INSERT` instruction. btw the same query with multiple insertions works great from sql client editor, even with covered with transaction.

jsmonkey:
I think that's mb cuz of lack of `multipleStatements: true` property in init of the mysql connection.

pleerock:
You can set this option in connection options.

jsmonkey:
Yeah, but I don't have any interceptor for connection creation in scope of migrations api. `QueryRunner` object which is passed to `up` or `down` handlers already has inside constructed `Connection` property.

[...]

PR #10061, opened 2023-05-18
docs: migration from .sql file

acao:
to polish off #1366 for good. noticing that users created a whole npm library for this problem to work around `multipleStatements` option

I wasn't sure whether to put this in the `QueryRunner` docs or here - it seems this should ideally only be used in migrations. Perhaps there are runtime applications for a psuedo-ETL type scenario, but it seems ill advised to even document the ability to do this at runtime?

There may be other edge cases where this fails, I've only tested it with standard pgdump style files with multi-statement, single statement-per-line

### Pull-Request Checklist

### DIFF
--- a/docs/migrations.md
+++ b/docs/migrations.md
@@ -401,6 +401,32 @@ export class QuestionRefactoringTIMESTAMP implements MigrationInterface {
 }
 ```
 
+## Executing .sql files
+
+If your .sql file contains multiple statements per file, which is common with most standard sql tooling, you will first need to enable the `multipleStatements` option in the applicable `DataSource`.
+
+Then, in your migrations, you can just import the file as a string and execute it!
+
+```ts
+import { MigrationInterface, QueryRunner } from "typeorm"
+import fs from 'node:fs/promises'
+
+export class ExampleMigrationFromFile implements MigrationInterface {
+   public async up(queryRunner: QueryRunner): Promise<void> {
+      // be sure to specify some form of encoding on readFile/readFileSync/etc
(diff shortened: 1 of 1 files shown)


hinogi (approved):
LGTM

docs/migrations.md
```diff
@@ -401,6 +401,32 @@ export class QuestionRefactoringTIMESTAMP implements MigrationInterface {
 }
 ```
 
+## Executing .sql files
+
+If your .sql file contains multiple statements per file, which is common with most standard sql tooling, you will first need to enable the `multipleStatements` option in the applicable `DataSource`.
```
pleerock, line 406: looks like `multipleStatements` is specific to some particular database types. How do we do for others?

acao: since this is for SQL files only, it should be applicable to all SQL based data sources, except for maybe SQLlite. 

I found it was required to use this option with PostgreSQL, even though the typeorm docs only show this option as available for MySQL/mariadb. 

As the option is to prevent SQL injection attacks, I would imagine it should be used with any SQL datasource

pleerock: right now based on the codebase `multipleStatements` flag is only available in `mysql` and `spanner` drivers. We need to add flag into typings as well for all other drivers which support this feature before we can add the docs. 

Because otherwise docs will confuse users who attempt to use `multipleStatements` in their driver and it result them a type error.

gioboa (commented):
is this PR still valid? may I ask you to solve the conflicts please? :pray:

PR #10149, opened 2023-06-23
feat: prefer using mariadb connector for MariaDB server

rusher:
### Description of change

Use mariadb connector by default for MariaDB server
mariadb connector is more appropriate for MariaDB server (faster and has MariaDB specificities, compatible with mysql and mysql2 connectors)

MariaDB/MySQL query runner small improvement: 

* avoid unnecessary string literal
* avoid unnecessary identifier escaping
* ensure database/table good escaping
* prefer text concatenation to string literal

### Pull-Request Checklist

- [ ] This pull request links relevant issues as `Fixes #0000` n/a

Tests using type "mariadb" will automatically use mariadb connector, so without test change, this modification will be tested.
Internal documentation indicate change to using mariadb connector by default when using mariadb type, but external documentation may have to be updated.


src/driver/mysql/MysqlQueryRunner.ts
```diff
@@ -27,6 +27,18 @@ import { TypeORMError } from "../../error"
 import { MetadataTableType } from "../types/MetadataTableType"
 import { InstanceChecker } from "../../util/InstanceChecker"
 
+const LITTERAL_ESCAPE: { [index: string]: string } = {
```
hinogi, line 30:
```suggestion
const LITERAL_ESCAPE: { [index: string]: string } = {
```

kusigit:
I know that mysql and mariadb driver is compatible. But I think a separate module `src/driver/mariadb/...` like all other drivers would be helpful for maintenance. We could also reduce the `if ... else` conditions in the `MysqlDriver.ts`. What du you think?

rusher:
@kusigit i initially have done this PR with a separate mariadb folder like others, but in fact, i revert that because it would have make maintenance more complex. I think that current behavior is better.

pleerock:
Can you remove redundant changes? It makes review harder to do. Also, conflicts must be resolved.

gioboa:
Is it still valid or is it abandoned?

PR #10161, opened 2023-06-27
fix: add mssql disableAsciiToUnicodeParamConversion option and tests

### LINKED ISSUE #10131: MssqlParameter of type varchar is being treated like nvarchar
### Issue description

MssqlParameter of type varchar is being treated like nvarchar 

### Expected Behavior

We are using TypeORM to connect to a MSSQL Server DB.  A number of our tables have `varchar` columns with indexes on them which we filter on in our WHERE clauses.  For example:

```
andWhere('"name" = :firstName', {
   firstName: new MssqlParameter('Bill', 'varchar', 100)
})
```

In previous versions of TypeORM (e.g. 0.2.33), this parameter was correctly treated like a `varchar` in the resulting queries that hit the database.  This led to performant queries that properly utilized our desired indexes.  

### Actual Behavior

However, in versions 0.3.x we are observing this parameter is now being treated as an `nvarchar` which is causing implicit conversion and our query performance to plummet.  

This seems to have been introduced by this PR: https://github.com/typeorm/typeorm/pull/7933 which appears to be intentionally disregarding the explicitly declared parameter types when they are `varchar` or `char`.  This seems like a very poor solution to the original problem and surely must be causing significant performance issues for anyone querying against `varchar` columns like this.  We've tried a number of other approaches and every way we declare this parameter it ends up being converted to an `nvarchar` before it hits the database.

### Steps to reproduce

I don't have a minimal example right now but this appears to be consistently reproducible when querying against a `varchar` column in a MSSQL Server DB using a `varchar` parameter in the `WHERE` clause.

### My Environment

| Node.js version     | 16.17.0  |
| Typescript version  | 4.7.4  |
| Database     | MSSQL Server  |

### Additional Context

_No response_

- [ ] aurora-mysql
- [ ] aurora-postgres
- [ ] better-sqlite3
- [ ] cockroachdb
- [ ] cordova
- [ ] expo
- [ ] mongodb
- [ ] mysql
- [ ] nativescript
- [ ] oracle
- [ ] postgres
- [ ] react-native
- [ ] sap
- [ ] spanner
- [ ] sqlite
- [ ] sqlite-abstract
- [ ] sqljs
- [X] sqlserver

nadiamk:
This adds a new property `disableAsciiToUnicodeParamConversion` to `SqlServerConnectionOptions` to control whether the `MssqlParameter` types `char`, `varchar`, and `text` are converted to their unicode equivalents `nchar`, `nvarchar`, and `ntext`. This defaults to `false`, maintaining the current behavior.

Example:
```
andWhere('"name" = :firstName', {
   firstName: new MssqlParameter('Bill', 'varchar', 100)
})
```
- The current behavior: `firstName` is treated as an `nvarchar`
- This PR introduces configuring this behavior: setting `disableAsciiToUnicodeParamConversion` to `true` treats `firstName` as `varchar`. If `disableAsciiToUnicodeParamConversion` is set to `false` or omitted from the ORM configuration values, `firstName` remains `nvarchar`

See full description of the issue, including the motivation, in https://github.com/typeorm/typeorm/issues/10131

Fixes #10131

### Pull-Request Checklist

### DIFF
--- a/src/driver/sqlserver/SqlServerQueryRunner.ts
+++ b/src/driver/sqlserver/SqlServerQueryRunner.ts
@@ -4049,12 +4049,33 @@ export class SqlServerQueryRunner
             case "tinyint":
                 return this.driver.mssql.TinyInt
             case "char":
+                if (
+                    this.driver.options.options
+                        ?.disableAsciiToUnicodeParamConversion
+                ) {
+                    return this.driver.mssql.Char(...parameter.params)
+                }
+                return this.driver.mssql.NChar(...parameter.params)
             case "nchar":
                 return this.driver.mssql.NChar(...parameter.params)
             case "text":
+                if (
+                    this.driver.options.options
+                        ?.disableAsciiToUnicodeParamConversion
+                ) {
+                    return this.driver.mssql.Text
+                }
+                return this.driver.mssql.Ntext
             case "ntext":
                 return this.driver.mssql.Ntext
             case "varchar":
+                if (
+                    this.driver.options.options
+                        ?.disableAsciiToUnicodeParamConversion
+                ) {
+                    return this.driver.mssql.VarChar(...parameter.params)
+                }
+                return this.driver.mssql.NVarChar(...parameter.params)
             case "nvarchar":
                 return this.driver.mssql.NVarChar(...parameter.params)
             case "xml":
--- a/src/driver/sqlserver/SqlServerConnectionOptions.ts
+++ b/src/driver/sqlserver/SqlServerConnectionOptions.ts
@@ -186,6 +186,12 @@ export interface SqlServerConnectionOptions
         readonly disableOutputReturning?: boolean
 
         /**
+         * A boolean, controlling whether MssqlParameter types char, varchar, and text are converted to their unicode equivalents, nchar, nvarchar, and ntext.
+         * (default: false, meaning that char/varchar/text parameters will be converted to nchar/nvarchar/ntext)
+         */
+        readonly disableAsciiToUnicodeParamConversion?: boolean
+
+        /**
          * Debug options
          */
         readonly debug?: {
--- a/test/github-issues/10131/entity/Example.ts
+++ b/test/github-issues/10131/entity/Example.ts
@@ -1,0 +1,13 @@
+import { Entity, PrimaryGeneratedColumn, Column } from "../../../../src"
+
+@Entity()
+export class Example {
+    @PrimaryGeneratedColumn("uuid")
+    id?: string
+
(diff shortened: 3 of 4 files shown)


mattnohelty-mastery:
This looks great to me and should solve the problem I reported in https://github.com/typeorm/typeorm/issues/10131.  I think it would be more correct default behavior for `disableAsciiToUnicodeParamConversion` to be `true` but in the spirit of maintaining backward compatibility, defaulting to `false` makes sense.

src/driver/sqlserver/SqlServerQueryRunner.ts
```diff
@@ -4049,12 +4049,33 @@ export class SqlServerQueryRunner
...
             case "char":
+                if (
+                    this.driver.options.options
+                        ?.disableAsciiToUnicodeParamConversion
+                ) {
+                    return this.driver.mssql.Char(...parameter.params)
+                }
+                return this.driver.mssql.NChar(...parameter.params)
             case "nchar":
                 return this.driver.mssql.NChar(...parameter.params)
```
hinogi, line 4060:
```suggestion
            case "char":
                if (
                    this.driver.options.options
                        ?.disableAsciiToUnicodeParamConversion
                ) {
                    return this.driver.mssql.Char(...parameter.params)
                }
            case "nchar":
                return this.driver.mssql.NChar(...parameter.params)
```
No idea if duplication of fall through is worse :D

hinogi (approved):
LGTM

src/driver/sqlserver/SqlServerQueryRunner.ts
nadiamk: Thank you for the review! 

I'm happy to make the change, but question - the conditional in the "char" case is triggering "TS7029: Fallthrough case in switch" due to tsconfig.json's `"noFallthroughCasesInSwitch": true"`. Would you still like me to make the change, bypassing the error with a `// @ts-ignore`?

hinogi: as I said, both is not nice, just keep your version.

AlexMesser:
thank you for contribution!

sudhirt4:
Would it be possible to have this enabled by default ? 

For anyone just coming into typeorm and using SQL server database, this context of this configuration is not visible. And for cases like when "varchar" column is used, runtime type conversion leads to a point of failure only known later when scaling.

mguida22:
> Would it make sense to have this enabled by default ?

I think so. @sudhirt4 can you open a new issue to track changing the default? We can discuss the implications and how to release there.

PR #10170, opened 2023-06-30
fix: using async datasource to configure typeorm

### LINKED ISSUE #10168: Using async DataSource is broken
### Issue description

When defining an async datasource to use for migrations the typeorm CLI breaks

### Expected Behavior

It should be possible to define async `DataSource` according to:
https://github.com/typeorm/typeorm/issues/8914

And use that for configuring TypeOrm.

### Actual Behavior

When declaring an async data source as mentioned in the original issue https://github.com/typeorm/typeorm/issues/8914

```typescript
const buildDataSource = async () => {
  await ConfigModule.envVariablesLoaded;

  return new DataSource({
    type: 'postgres',
    url: process.env.DATABASE_URL,
    entities: ['dist/domain/**/*.entity.js'],
    migrations: ['dist/database/migrations/*.js'],
  });
};

export default buildDataSource();
```

It fails with 
```shell
Error during migration generation:
TypeError: dataSourceFileExports is not iterable
    at Function.loadDataSource (/Some/Path/backend/node_modules/src/commands/CommandUtils.ts:37:34)
    at async Object.handler ((/Some/Path/backend/node_modules/src/commands/MigrationGenerateCommand.ts:73:26)
```

I have tried numerous approaches to the export but they basically all fail.

### Steps to reproduce

Define an async `DataSource` and run the `migrate:generate` or `migrate:run` command.

### My Environment

| Node.js version     | v18.15.0  |
| Typescript version  | 4.5.5 |

### Additional Context

After some debugging it turns out that the problem lies in the `CommandUtils.ts` - checking for `someting instanceof Promise` is not stable and it does not work on my machine ™️. The check fails and thus passes the promise up the chain to the `InstanceChecker` which naturally fails.

I have a PR coming.
 

- [ ] aurora-mysql
- [ ] aurora-postgres
- [ ] better-sqlite3
- [ ] cockroachdb
- [ ] cordova
- [ ] expo
- [ ] mongodb
- [ ] mysql
- [ ] nativescript
- [ ] oracle
- [ ] postgres
- [ ] react-native
- [ ] sap
- [ ] spanner
- [ ] sqlite
- [ ] sqlite-abstract
- [ ] sqljs
- [ ] sqlserver

nover:
Fixes #10168 

### Description of change

The check in `CommandUtils.ts` for detecting whether the DataSource-source is a promise is broken.

Rather than testing whether it's a promise, simply await the value and use it. The `await` operator is defined in such way that if the thing is not awaitable we simply get the value back

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/await#return_value

### Pull-Request Checklist

- [ ] There are new or updated unit tests validating the change - no because all relevant unit tests have been disabled by the code-owners

### DIFF
--- a/src/commands/CommandUtils.ts
+++ b/src/commands/CommandUtils.ts
@@ -38,8 +38,9 @@ export class CommandUtils {
             const fileExport = dataSourceFileExports[fileExportKey]
             // It is necessary to await here in case of the exported async value (Promise<DataSource>).
             // e.g. the DataSource is instantiated with an async factory in the source file
-            const awaitedFileExport =
-                fileExport instanceof Promise ? await fileExport : fileExport
+            // It is safe to await regardless of the export being async or not due to `awaits` definition:
+            // https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/await#return_value
+            const awaitedFileExport = await fileExport
             if (InstanceChecker.isDataSource(awaitedFileExport)) {
                 dataSourceExports.push(awaitedFileExport)
             }


hinogi:
I think another way would be to check if `.then` is available

nover:
> I think another way would be to check if `.then` is available

We could do that explicitly and only await if the object is "thenable" - but this is exactly what await does under the hood (from MDN):

```
Await result 

The fulfillment value of the promise or thenable object, or, if the expression is not thenable, the expression's own value.
```

hinogi:
> > I think another way would be to check if `.then` is available
> 
> We could do that explicitly and only await if the object is "thenable" - but this is exactly what await does under the hood (from MDN):
> 
> ```
> Await result 
> 
> The fulfillment value of the promise or thenable object, or, 
> if the expression is not thenable, the expression's own value.
> ```

Then you are absolutely right.  👍

hinogi (approved):

nover:
Any chance of getting this merged soon @hinogi ?

hinogi:
@nover I am not a maintainer, I am just helping with reviewing as much as I can to accelerate the process a bit.

pleerock:
thanks @nover @hinogi !

PR #10178, opened 2023-07-03
fix: Ensure findOne method gets where options passed in at compile time instead of only at runtime

elderapo:
### Description of change

Before this change, it was possible to write and compile the following code:
```ts
const someRepository: Repository<any> = ...;
await someRepository.findOne({}); // would throw at runtime
```

but it would fail at the runtime because of [this check](https://github.com/typeorm/typeorm/blob/master/src/entity-manager/EntityManager.ts#L1183). This change makes it so the typescript compiler won't let you compile this code - 1 less possible runtime bug :)

### Pull-Request Checklist


hinogi:
I'm not sure if the mongorepository belongs to this problem. Anyway, maybe you can tackle the code at the mentioned location also in the way of instead of

```ts
!options.where
```


```ts
!options.hasOwnProperty('where')
```

or something comparable


https://github.com/typeorm/typeorm/blob/b1a3a395049052f3f031e9fd27b99769b03b9011/src/entity-manager/EntityManager.ts#L1181C27-L1181C27

hinogi:
@pleerock looks like this just need a rerun because the connection to microsoft timed out

elderapo:
> I'm not sure if the mongorepository belongs to this problem.

What do you mean? Are you referring to [these changes](https://github.com/typeorm/typeorm/pull/10178/files#diff-06402bac7d3c217ce8afc1071fffcd25d026b57b4f0c19593615be01539359b4R72-R75)? If so they have to be like that because `MongoRepository` has it's own `findOne` which actually allows `where` to be omitted. Without this change you get [error on this line](https://github.com/typeorm/typeorm/blob/91f8064792c27971bcc33e15688b5b921abdae44/src/entity-manager/EntityManager.ts#L1394):
```ts
Argument of type 'MongoRepository<Entity>' is not assignable to parameter of type 'Repository<any>'.
  Types of property 'findOne' are incompatible.
    Type '(options: MongoFindOneOptions<Entity>) => Promise<Entity | null>' is not assignable to type '(options: FindOneOptions<any>) => Promise<any>'.
      Types of parameters 'options' and 'options' are incompatible.
        Type 'FindOneOptions<any>' is not assignable to type 'MongoFindOneOptions<Entity>'.
          Type 'FindOneOptions<any>' is not assignable to type 'Omit<FindOneOptions<Entity>, "where">'.
            Types of property 'select' are incompatible.
              Type 'FindOptionsSelect<any> | FindOptionsSelectByString<any> | undefined' is not assignable to type 'FindOptionsSelect<Entity> | FindOptionsSelectByString<Entity> | undefined'.
                Type 'FindOptionsSelectByString<any>' is not assignable to type 'FindOptionsSelect<Entity> | FindOptionsSelectByString<Entity> | undefined'.
                  Type '(string | number | symbol)[]' is not assignable to type 'FindOptionsSelectByString<Entity>'.
                    Type 'string | number | symbol' is not assignable to type 'keyof Entity'.
                      Type 'symbol' is not assignable to type 'keyof Entity'.
                        Type 'symbol' is not assignable to type 'string | number'.ts(2345)
```

---

> Anyway, maybe you can tackle the code at the mentioned location also in the way of instead of
> 
> ```ts
> !options.where
> ```
> 
> ```ts
> !options.hasOwnProperty('where')
> ```
> 
> or something comparable
> 
> https://github.com/typeorm/typeorm/blob/b1a3a395049052f3f031e9fd27b99769b03b9011/src/entity-manager/EntityManager.ts#L1181C27-L1181C27

Why? `if ("where" in options)` could also be used but I don't see any obvious benefit to it.

pleerock:
Unfortunately I'm going to reject this feature because:

* some people might think using findOne without where is appropriate behaviour
* this is breaking change anyway, which we cannot accept at this moment.

PR #10184, opened 2023-07-05
fix: CommandUtils unable to process a regular default export

### LINKED ISSUE #8810: Generating migrations with cli
### Expected Behavior

Migrations get generated using the cli

### Actual Behavior

I am following the documentation outlined here https://typeorm.io/migrations and I feel like the documentation is way different then what is actually happening in the cli. I have tried just running the following command:

`typeorm migration:generate -n PostRefactoring`

When I run this command I get the following:

`Not enough non-option arguments: got 0, need at least 1`

So then I modify the command to something like this:

`Missing required argument: dataSource`

So then from there I add the datasource since that is what it is asking for:

`npx typeorm migration:generate test -d app-data-source.ts` (I found the -d somewhere online  while investigating this issue)

But once I run that command I get the following error:

`File must contain a TypeScript / JavaScript code and export a DataSource instance.`

I have been trying to get this to work for days and I am about ready to throw the towl in and use a different ORM...which I don't want to do. I know that everything is working fine because I can start up the app and the app connects to the database and creates rows in the database. So it is just something with the cli and trying to generate the migrations.

Here is what is in my app-data-source.ts file:

```
import { DataSource } from "typeorm";
import { User } from "./entitiy/User";

const myDataSource = new DataSource({
  type: "postgres",
  host: "localhost",
  port: 5432,
  username: "postgres",
  password: "",
  database: "postgres",
  entities: [User],
  migrations: ["/migrations/"],
  logging: true,
  synchronize: true,
});


export default myDataSource;
```

This is the contents of my app.ts file:

```
import * as express from "express";
import { Request, Response } from "express";
import myDataSource from "./app-data-source";
import { User } from "./entitiy/User";

// establish database connection
myDataSource
  .initialize()
  .then(() => {
    console.log("Data Source has been initialized!");
  })
  .catch((err) => {
    console.error("Error during Data Source initialization:", err);
  });

// create and setup express app
const app = express();
app.use(express.json());

// register routes
app.get("/users", async function (req: Request, res: Response) {
  const users = await myDataSource.getRepository(User).find();
  res.json(users);
});

app.get("/users/:id", async function (req: Request, res: Response) {
  const results = await myDataSource.getRepository(User).findOneBy({
    id: +req.params.id,
  });
  return res.send(results);
});

app.post("/users", async function (req: Request, res: Response) {
  const user = await myDataSource.getRepository(User).create(req.body);
  const results = await myDataSource.getRepository(User).save(user);
  return res.send(results);
});

app.put("/users/:id", async function (req: Request, res: Response) {
  const user = await myDataSource.getRepository(User).findOneBy({
    id: +req.params.id,
  });
  myDataSource.getRepository(User).merge(user, req.body);
  const results = await myDataSource.getRepository(User).save(user);
  return res.send(results);
});

app.delete("/users/:id", async function (req: Request, res: Response) {
  const results = await myDataSource.getRepository(User).delete(req.params.id);
  return res.send(results);
});

// start express server
app.listen(3000);
```

All these files were setup using the instructions outlined in the express typeorm documentation:
[Example with Express](https://typeorm.io/example-with-express)

Which BTW, this also has issues and doesn't run from what is outlined in the directions. I had to make some modifications to actually make the app run compared to what is outlined in that guide. One of the changes I made in the data source file was putting the actual Entity name in the entity property of the datasource compared to the guide that has you put the path in.

Anyway, am I crazy...am I doing something wrong? You guys might not have enough information to answer the first question but can you at least help with trying to figure out how to make the migrations work?

### Steps to Reproduce

Set up you app using the guide outlined above to use express

### My Environment

| Operating System    |  MAC OS Monterrey         |
| Node.js version     | v16.14.2  | 
| TypeORM version     | 0.3.4  | 

Durisvk:
The solution to `loadDataSource` was probably designed to adapt to various forms of exports coming from ormconfig.js however it failed to adapt to a very regular case where you just do `module.exports = new DataSource({ ... })` resulting in `Error: Given data source file must contain export of a DataSource instance`

Fixes #8810

### Description of change

The loadDataSource did not take into account that the default export might already contain the `DataSource` instance itself. The solution started to iterate over the keys of the `DataSource` in the hope to find a `DataSource` instance under one of those keys. I've adjusted the logic to check if the DataSource is the subject to default export itself then just return it from the function and do not iterate over its keys.

### Pull-Request Checklist

- [ ] `npm run test` passes with this change N/A

### DIFF
--- a/src/commands/CommandUtils.ts
+++ b/src/commands/CommandUtils.ts
@@ -33,6 +33,10 @@ export class CommandUtils {
             )
         }
 
+        if (InstanceChecker.isDataSource(dataSourceFileExports)) {
+            return dataSourceFileExports
+        }
+
         const dataSourceExports = []
         for (const fileExportKey in dataSourceFileExports) {
             const fileExport = dataSourceFileExports[fileExportKey]


hinogi:
Since v0.3.0 Loading of some external datasource config seems deprecated. Looks to me that this functionality was forgotten to be deprecated too since there is a different approach now.

https://github.com/typeorm/typeorm/blob/master/CHANGELOG.md#breaking-changes-1

What do you say @durisvk?

Durisvk:
Hey @hinogi, my PR actually fixes the new approach of the DataSource config, not the deprecated one.

pleerock:
Looks good to me, we just missed the support of `export default new DataSource()` which is legit use case.

PR #10185, opened 2023-07-06
fix: enable preload to retrieve entities added within a transaction

codafett:
This fixes an issue whereby if you used an running transactional entity manager to save a new entity to the db if you then called preload the entity would not be retrieved as the underlying PlainObjectToDatabaseEntityTransformer was called with this.connection.manager rather than using just this.

### Description of change

This PR makes EntityManager `preload` work within a transaction.

**Issue:**
1. Use a data source to create a new transaction
2. Within the transaction save a new entity to the database using the transactions EntityManager
3. Call `preload` on the transactions EntityManager
4. You received `undefined`

**Why is this a problem?**
We are running our API tests within a single transaction. We start the transaction in a `beforeEach` and roll it bakc with the `afterEach` to give us total isolation between tests. However, whilst testing an update scenario (that uses pre load) our test set up first creates a new entity before calling an end point that updates it. When the end point called `preload` on the shared EntityManager is returned `undefined`. We further experimented by simply creating an enrity within a transaction and the calling `preload` - still we got `undefined`. The only way to make the entity visible within was to create a totally separate transaction that committed the entity to the DB. However, this then breaks the isolation of the tests as all tests would then see this entity which could potentially cause nasty side affects.

I traced the issue to the EntityManager creating a PlainObjectToDatabaseEntityTransformer and passing in `this.connection.manager` which was used within PlainObjectToDatabaseEntityTransformer to find all entities with matching id's - but as this was executed on the root connection and the data hadn't been committed it failed to load them.

### Pull-Request Checklist

### DIFF
--- a/src/entity-manager/EntityManager.ts
+++ b/src/entity-manager/EntityManager.ts
@@ -330,7 +330,7 @@ export class EntityManager {
     ): Promise<Entity | undefined> {
         const metadata = this.connection.getMetadata(entityClass)
         const plainObjectToDatabaseEntityTransformer =
-            new PlainObjectToDatabaseEntityTransformer(this.connection.manager)
+            new PlainObjectToDatabaseEntityTransformer(this)
         const transformedEntity =
             await plainObjectToDatabaseEntityTransformer.transform(
                 entityLike,
--- a/test/functional/repository/basic-methods/repository-basic-methods.ts
+++ b/test/functional/repository/basic-methods/repository-basic-methods.ts
@@ -270,26 +270,24 @@ describe("repository > basic methods", () => {
             ))
 
         it("should preload entity from the given object with only id within a transaction", () =>
-        Promise.all(
-            connections.map(async (connection) => {
-                await connection.manager.transaction(
-                    async (em) => {
+            Promise.all(
+                connections.map(async (connection) => {
+                    await connection.manager.transaction(async (em) => {
                         const blogRepository = em.getRepository(Blog)
-                        const categoryRepository =
-                        em.getRepository(Category)
-        
+                        const categoryRepository = em.getRepository(Category)
+
                         // save the category
                         const category = new Category()
                         category.name = "people"
                         await categoryRepository.save(category)
-        
+
                         // save the blog
                         const blog = new Blog()
                         blog.title = "About people"
                         blog.text = "Blog about good people"
                         blog.categories = [category]
                         await blogRepository.save(blog)
-        
+
                         // and preload it
                         const plainBlogWithId = { id: 1 }
                         const preloadedBlog = await blogRepository.preload(
@@ -301,11 +299,10 @@ describe("repository > basic methods", () => {
                         preloadedBlog!.title.should.be.equal("About people")
                         preloadedBlog!.text.should.be.equal(
                             "Blog about good people",
-                        )   
-                    }
-                )
-            }),
-        ))
+                        )
+                    })
+                }),
+            ))
 
         it("should preload entity and all relations given in the object", () =>
             Promise.all(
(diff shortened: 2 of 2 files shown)


test/functional/repository/basic-methods/repository-basic-methods.ts
```diff
@@ -23,6 +23,7 @@ import { RelationAsPrimaryKey } from "./entity/RelationAsPrimaryKey"
 import { TwoUniqueColumnsEntity } from "./entity/TwoUniqueColumns"
 import { OneToOneRelationEntity } from "./entity/OneToOneRelation"
 import { UpsertOptions } from "../../../../src/repository/UpsertOptions"
+import { expect } from "chai"
```
hinogi, line 26:
```suggestion
```
`import "../../../utils/test-setup"` already imports chai as a whole.

src/entity-manager/EntityManager.ts
hinogi, line 1: I wonder if there are more places that might have the same issue. Any idea if some kind of typescript typing could help?

hinogi:
I suspect there is a deeper problem. Maybe tackle ` // todo: fix type hinting` first.

hinogi:
I added #10192 just to get deprecated stuff out of the way first.

src/entity-manager/EntityManager.ts
codafett: You could change the type on EntityManager to:

```readonly connection: Omit<DataSource, 'manager'>```

the knock on effect is that you also need to change the type on the constructor for `EntityPersistExecutor` to:
```
constructor(
        protected connection: Omit<DataSource, 'manager'>,
        protected queryRunner: QueryRunner | undefined,
        protected mode: "save" | "remove" | "soft-remove" | "recover",
        protected target: Function | string | undefined,
        protected entity: ObjectLiteral | ObjectLiteral[],
        protected options?: SaveOptions & RemoveOptions,
    ) {}
```

`EntityPersistExecutor` doesn't use the `manager` property and doesn't pass the connection on to anything else that might.

I also did a `find all references` on `EntityManager.connection` and though there are several place where it's used, none of them use the `manager` field.

Comment (codafett) [reply src/entity-manager/EntityManager.ts:1]: I've pushed a commit with this change

File: test/functional/repository/basic-methods/repository-basic-methods.ts (hunk 1 above)
Comment (codafett) [reply test/functional/repository/basic-methods/repository-basic-methods.ts]: I've undone this change. It broke other db tests that for some reason complained chai was not defined.
I looked around the rest of the code base and every file that used `expect` imports chai directly so I don't think this is a problem?

**Comment by hinogi** [comment]
I will look into it when I am back from my vacation. If nobody else does in the meantime.


⁣Get BlueMail for Android ​

**Comment by codafett** [comment]
@hinogi  Not sure why the mariadb tests are failing (I can't test locally)? Could it be just be an intermittent thing (as this PR doesn't touch the test that is failing)? Anyway I can't re-run the tests as I don't have permissions?

**Comment by hinogi** [comment]
> @hinogi Not sure why the mariadb tests are failing (I can't test locally)? Could it be just be an intermittent thing (as this PR doesn't touch the test that is failing)? Anyway I can't re-run the tests as I don't have permissions?

```
1) repository > basic methods
       "before each" hook for "should merge multiple entities":
     QueryFailedError: SQLITE_ERROR: no such table: main.category
      at Statement.handler (src/driver/sqlite/SqliteQueryRunner.ts:22:81)
      at Statement.replacement (node_modules/sqlite3/lib/trace.js:25:27)
      at Statement.replacement (node_modules/sqlite3/lib/trace.js:25:27)
```

**Comment by hinogi** [comment]
I guess the tests run for every driver.

File: src/entity-manager/EntityManager.ts [hunk 2]
```diff
@@ -53,7 +53,7 @@ export class EntityManager {
     /**
      * Connection used by this entity manager.
      */
-    readonly connection: DataSource
+    readonly connection: Omit<DataSource, "manager">
```
pleerock, line 56: Proposed solution doesn't look appropriate. Why we need to omit manager property from the DataSource type?

hinogi: Maybe to avoid some circular dependency.

pleerock: no, circular dependencies don't get resolved in a such way. Need to fix source of circular and re-structure the code, or try to put `import type` statements for circular deps issue.

gioboa (commented):
Thanks @codafett 
what's the situation of this PR, is it still valid?

PR #10192, opened 2023-07-08
refactor: removed deprecated usage of findByIds

hinogi:
### Description of change

Also removed the any type from targetWithIds.target
Transformed the find operation into async/await

### Pull-Request Checklist

- [ ] `npm run test` passes with this change N/A Failed in CodeSpaces

### DIFF
--- a/src/query-builder/transformer/PlainObjectToDatabaseEntityTransformer.ts
+++ b/src/query-builder/transformer/PlainObjectToDatabaseEntityTransformer.ts
@@ -2,6 +2,7 @@ import { ObjectLiteral } from "../../common/ObjectLiteral"
 import { EntityMetadata } from "../../metadata/EntityMetadata"
 import { EntityManager } from "../../entity-manager/EntityManager"
 import { RelationMetadata } from "../../metadata/RelationMetadata"
+import { In } from "../../find-options/operator/In"
 
 /**
  */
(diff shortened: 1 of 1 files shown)


src/query-builder/transformer/PlainObjectToDatabaseEntityTransformer.ts
```diff
@@ -129,16 +130,14 @@ export class PlainObjectToDatabaseEntityTransformer {
...
-                        targetWithIds.ids,
-                    )
-                    .then((entities) =>
-                        loadMap.fillEntities(targetWithIds.target, entities),
-                    )
+            loadMap.groupByTargetIds().map(async (targetWithIds) => {
+                const foundEntities = await this.manager.findBy<ObjectLiteral>(
+                    targetWithIds.target,
+                    {
+                        id: In([...targetWithIds.ids]),
```
iJhefe, line 137: shouldn't we look for the primary key instead of just "id" ?

hinogi: Well I just refactored the findByIds implementation, not the logic. This is the default. I think this is just a conveniance function for the often used column `id` as unique identifier. For all other cases you will have to use `findBy` instead.

pleerock:
Implementation isn't correct. Such implementation won't solve any problem. "id" usage here is invalid and instead we should take primary column names from the entity metadata and map them to the corresponding values.

PR #10198, opened 2023-07-11
perf: improve SapQueryRunner performance

lotczyk:
Use async API exposed by driver

### Description of change

Usage of synchronous `prepare` was a critical bottleneck. In case of having appropriate parameters we can skip `prepare` execution altogether and call `exec` directly which can improve performance even further.

### Pull-Request Checklist

### DIFF
--- a/src/driver/sap/SapQueryRunner.ts
+++ b/src/driver/sap/SapQueryRunner.ts
@@ -202,15 +202,24 @@ export class SapQueryRunner extends BaseQueryRunner implements QueryRunner {
             const queryStartTime = +new Date()
             const isInsertQuery = query.substr(0, 11) === "INSERT INTO"
 
-            statement = databaseConnection.prepare(query)
+            if (parameters?.some(Array.isArray)) {
+                statement = await promisify(
+                    databaseConnection.prepare.bind(databaseConnection),
+                )(query)
+            }
 
-            const raw = await new Promise<any>((ok, fail) => {
-                statement.exec(parameters, (err: any, raw: any) =>
-                    err
-                        ? fail(new QueryFailedError(query, parameters, err))
-                        : ok(raw),
-                )
-            })
+            let raw: any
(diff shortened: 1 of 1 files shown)


src/driver/sap/SapQueryRunner.ts
```diff
@@ -202,15 +202,24 @@ export class SapQueryRunner extends BaseQueryRunner implements QueryRunner {
             const queryStartTime = +new Date()
             const isInsertQuery = query.substr(0, 11) === "INSERT INTO"
 
-            statement = databaseConnection.prepare(query)
+            if (parameters?.some(Array.isArray)) {
```
alumni, line 205: You might want to add a comment in the code why we have this condition.

src/driver/sap/SapQueryRunner.ts
```diff
@@ -202,15 +202,24 @@ export class SapQueryRunner extends BaseQueryRunner implements QueryRunner {
...
-            const raw = await new Promise<any>((ok, fail) => {
-                statement.exec(parameters, (err: any, raw: any) =>
-                    err
-                        ? fail(new QueryFailedError(query, parameters, err))
-                        : ok(raw),
-                )
-            })
+            let raw: any
+            try {
+                raw = statement
```
alumni, line 213: This condition is identical to the single-branch if statement above, you could merge the two.

hinogi: there is a `prepare.bind` not an exec as I see it. But I do not know how they differ in execution. Therefore, if it is only a preparation for the exec and that only has to happen if there is some parameter in the parameters array this looks reasonable.

alumni:
It would be nice to merge this, since I'd like to add streaming support and I'll also have to change the way the statements are prepared & executed.

pleerock:
Thank you for contribution! 🎉

PR #10221, opened 2023-07-20
feat: Add an option to disable `cli-highlight` usage

netroy:
### Description of change

While cli highlighting can be a very useful tool, the package `cli-highlight` has a handful of issues for us:
1. The library seems abandoned, and a potential security issue waiting to happen.
2. When loading typeorm, over 30% of the code loaded into memory is coming from `cli-highlight` and its dependencies.

Just as how typeorm lazy-loads all drivers, It makes sense to do so for `cli-highlight` as well, optionally giving users the choice to not load this library at all.

### Pull-Request Checklist

### DIFF
--- a/src/commands/SchemaLogCommand.ts
+++ b/src/commands/SchemaLogCommand.ts
@@ -1,5 +1,4 @@
 import { DataSource } from "../data-source/DataSource"
-import { highlight } from "cli-highlight"
 import * as yargs from "yargs"
 import chalk from "chalk"
 import { PlatformTools } from "../platform/PlatformTools"
@@ -81,7 +80,7 @@ export class SchemaLogCommand implements yargs.CommandModule {
                         sqlString.substr(-1) === ";"
                             ? sqlString
                             : sqlString + ";"
-                    console.log(highlight(sqlString))
+                    console.log(PlatformTools.highlightSql(sqlString))
                 })
             }
             await dataSource.destroy()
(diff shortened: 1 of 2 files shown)


netroy:
@pleerock I'm curious if this change makes sense to typeorm maintainers? 
for more context: we run thousands of instances of an application where this package contributes to over 5% in baseline memory usage, and at our scale those memory savings can often lead to more infrastructure capacity.
But, I also understand that this wouldn't be relevant to most other users. so, I wanted to check with you if we should consider maintaining a patch/fork instead, as we'd really like to stop loading `cli-highlight` package in our application.

kusigit:
It would be good to write a note in the [documentation](https://github.com/typeorm/typeorm/blob/master/docs/logging.md) about the new environment variable `DISABLE_CLI_HIGHLIGHT`.

hinogi (changes requested):
> It would be good to write a note in the [documentation](https://github.com/typeorm/typeorm/blob/master/docs/logging.md) about the new environment variable `DISABLE_CLI_HIGHLIGHT`.

I second that. Docs missing.

mkilp:
Fair point, thanks for the heads-up. I will cherry pick these and open my own PR.

netroy:
Since `cli-highlight` looks abandoned, I'd rather recommend ripping it out instead of adding an option to disable it.

mkilp:
I have it out in my local repo right now and test in our systems first. I might raise a PR with the change - seems like a luxury library to begin with.

Issue #3302, opened 2018-12-21
Tracking query time for slow queries and statsd timers


chiangf:
**Issue type:**

[x] question
[ ] bug report
[ ] feature request
[ ] documentation issue

**Database system/driver:**

[ ] `cordova`
[ ] `mongodb`
[ ] `mssql`
[ ] `mysql` / `mariadb`
[ ] `oracle`
[x] `postgres`
[ ] `sqlite`
[ ] `sqljs`
[ ] `react-native`
[ ] `expo`

**TypeORM version:**

[x] `latest`
[ ] `@next`
[ ] `0.x.x` (or put your version here)

**Steps to reproduce or a small repository showing the problem:**

I'd like to be able to log slow queries to console/file and keep track of average query times with statsd. From http://typeorm.io/#/logging/using-custom-logger, I know that I can define a custom logger, and if I set `maxQueryExecutionTime=-1` (since 0 will be falsy), then the custom logger will always be called.

At that point, I can fire `statsd.timer` as well as check if the query took longer than some threshold and log as well. But this feels more difficult than it should be. I also looked into subscribers but it doesn't seem to track query time.

Is there a better way to do this?

pleerock:
what is difficult exactly? Creating a custom logger?

chiangf:
@pleerock It didn't feel right to have to use the slow query logger since I am trying to track statsd for all queries (not just slow ones), and setting maxQueryExecutionTime=-1 felt strange too. So I figured there must be another way. But I'm fine with it if this is the official way.

pleerock:
ah okay. We probably can add `beforeQuery` and `afterQuery` subscriber events. Feel free to PR

GaryChangCN:
anything new here?

MoritzGruber:
If anyone else has the same problem.
Here is a solution I came up with:

`
var pg = require('pg');
var shimmer = require('shimmer');


function wrapQuery(orig, name) {
  return function wrappedFunction(sql) {

    console.time(`PSQL: ${sql}`);
    var args = arguments
    var index = args.length - 1
    var cb = args[index]

    if (Array.isArray(cb)) {
      index = cb.length - 1
      cb = cb[index]
    }

    if (typeof cb === 'function') {
      args[index] = end
      return orig.apply(this, arguments)
    } else {
      cb = null
      var query = orig.apply(this, arguments)

      if (typeof query.on === 'function') {
        query.on('end', end)
        query.on('error', end)
[...]

PR #10234, opened 2023-07-26
feat:Add BeforeQuery and AfterQuery events

mnguyen081002:
Fixes https://github.com/typeorm/typeorm/issues/3302 (see https://github.com/typeorm/typeorm/issues/3302#issuecomment-449422542)
### Description of change

This PR adds BeforeQuery and AfterQuery events, which provide data about the queries run against the database immediately before and after execution. The events include the raw queries and parameters, and, for the AfterQuery event, the success status, raw result, execution time and possible error.

This feature enables better integration with a wide range of providers, such as logging and tracing utilities. Although similar functionality exists with custom loggers, only one custom logger can be used at once making multiple integrations cumbersome. Additionally, access to the query execution time is available for all queries only after manually setting [maxQueryExecutionTime: -1](https://orkhan.gitbook.io/typeorm/docs/logging#log-long-running-queries) -- setting this config value incorrectly could cause unexpected behavior with external connections.

I'm submitting this PR as a WIP for now to get feedback on my implementation as I finish updating the tests and documentation. If anyone has a better solution in mind, I'm open to suggestions!

### Pull-Request Checklist


src/driver/mysql/MysqlQueryRunner.ts
```diff
@@ -187,19 +188,29 @@ export class MysqlQueryRunner extends BaseQueryRunner implements QueryRunner {
...
+                    query,
+                    parameters,
+                )
+
                 const queryStartTime = +new Date()
                 databaseConnection.query(
                     query,
                     parameters,
-                    (err: any, raw: any) => {
+                    async (err: any, raw: any) => {
```
hinogi, line 207: since this function seems to return something. Switching from sync to async may cause a problem since it will now always return a promise.

src/driver/sap/SapQueryRunner.ts
```diff
@@ -261,20 +280,31 @@ export class SapQueryRunner extends BaseQueryRunner implements QueryRunner {
                 result.raw = identityValueResult[0]["CURRENT_IDENTITY_VALUE()"]
                 result.records = identityValueResult
             }
-        } catch (e) {
+        } catch (err) {
```
hinogi, line 283: Good catch of the bad naming 👍

hinogi (changes requested):
Double check on changing sync to async where something is returned. The change may have side effect or will cascade.

hinogi:
https://github.com/typeorm/typeorm/pull/10067 is this the same?

mnguyen081002:
> is this the same?

Yes, it appears to be the same

pleerock:
Thank you for contribution! 🎉

PR #10255, opened 2023-08-03
feat: lazy load all drivers to reduce typeorm's memory footprint

netroy:
### Description of change

Typeorm [already lazy-loads external node_modules](https://github.com/typeorm/typeorm/blob/master/src/platform/PlatformTools.ts#L36) that the driver classes are built around, but the Driver classes themselves are all immediately loaded into memory when someone imports typeorm into their code.

This PR changes that behavior to lazy-load the drivers only when they are actually used. This cuts down on typeorm's baseline memory footprint.
This change also reduces the memory that the garbage collector has to collect after typeorm is loaded.

Combined with #10221, the initial memory usage of typeorm reduces from about **50MB to 24MB**, and after the first GC, the retained memory is reduced from **20MB to 10MB**.

### Pull-Request Checklist

---

PS: [This comment](https://github.com/typeorm/typeorm/pull/10221#issuecomment-1650709288) from #10221 applies to this PR as well. 
If these changes don't seem relevant enough to the project maintainers, please let us know, so that we can try other ways to patch typeorm. 
Currently typeorm takes up about **40%** of our baseline memory usage, and bringing this number down is important to us.

### DIFF
--- a/src/driver/DriverFactory.ts
+++ b/src/driver/DriverFactory.ts
@@ -1,24 +1,5 @@
-import { MissingDriverError } from "../error/MissingDriverError"
-import { CockroachDriver } from "./cockroachdb/CockroachDriver"
-import { MongoDriver } from "./mongodb/MongoDriver"
-import { SqlServerDriver } from "./sqlserver/SqlServerDriver"
-import { OracleDriver } from "./oracle/OracleDriver"
-import { SqliteDriver } from "./sqlite/SqliteDriver"
-import { CordovaDriver } from "./cordova/CordovaDriver"
-import { ReactNativeDriver } from "./react-native/ReactNativeDriver"
-import { NativescriptDriver } from "./nativescript/NativescriptDriver"
-import { SqljsDriver } from "./sqljs/SqljsDriver"
-import { MysqlDriver } from "./mysql/MysqlDriver"
-import { PostgresDriver } from "./postgres/PostgresDriver"
-import { ExpoDriver } from "./expo/ExpoDriver"
-import { AuroraMysqlDriver } from "./aurora-mysql/AuroraMysqlDriver"
-import { AuroraPostgresDriver } from "./aurora-postgres/AuroraPostgresDriver"
-import { Driver } from "./Driver"
-import { DataSource } from "../data-source/DataSource"
-import { SapDriver } from "./sap/SapDriver"
-import { BetterSqlite3Driver } from "./better-sqlite3/BetterSqlite3Driver"
-import { CapacitorDriver } from "./capacitor/CapacitorDriver"
-import { SpannerDriver } from "./spanner/SpannerDriver"
+import type { Driver, DriverConstructor } from "./Driver"
+import type { DataSource } from "../data-source/DataSource"
 
 /**
  * Helps to create drivers.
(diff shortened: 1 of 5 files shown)


FraneCaleta (approved):

src/driver/DriverFactory.ts
```diff
@@ -29,46 +10,57 @@ export class DriverFactory {
      */
     create(connection: DataSource): Driver {
         const { type } = connection.options
+        return new (this.getDriver(type))(connection)
```
kusigit, line 13:
```suggestion
    create(dataSource: DataSource): Driver {
        const { type } = dataSource.options
        return new (this.getDriver(type))(dataSource)
```

kusigit (approved):
LGTM

src/driver/DriverFactory.ts
```diff
@@ -27,48 +8,59 @@ export class DriverFactory {
...
      * Creates a new driver depend on a given connection's driver type.
      */
-    create(connection: DataSource): Driver {
-        const { type } = connection.options
+    create(dataSource: DataSource): Driver {
+        const { type } = dataSource.options
+        return new (this.getDriver(type))(dataSource)
+    }
+
+    private getDriver(type: DataSource["options"]["type"]): DriverConstructor {
```
hinogi, line 16:
```suggestion
    private async getDriver(type: DataSource["options"]["type"]): DriverConstructor {
```

src/driver/DriverFactory.ts
```diff
@@ -27,48 +8,59 @@ export class DriverFactory {
...
         switch (type) {
             case "mysql":
-                return new MysqlDriver(connection)
+            case "mariadb":
+                return require("./mysql/MysqlDriver").MysqlDriver
```
hinogi, line 20:
```suggestion
                return await import("./mysql/MysqlDriver").default
```

hinogi: Mixing `import` and `require` is not a good choice in my eyes.

hinogi (changes requested):
Maybe this could be also a place for DI or Service Locator pattern to transfer that into.
It could be possible to have just all the paths in a Map for example and if you access the name you get the driver instance.

src/driver/DriverFactory.ts
netroy: this would require making `create` async as well. and that could break the public api.

hinogi, line 16:
```suggestion
        return new (this.drivers[type])(dataSource)
    }

    private const drivers = new Map([
      ['mysql', async () => import("./mysql/MysqlDriver").default],
      ...
    ])
```
something like that, just a different flavor

src/driver/DriverFactory.ts
netroy: can't use import without breaking the public api.

hinogi: > can't use import without breaking the public api.

What is the problem with dynamic imports?

src/driver/DriverFactory.ts
netroy: I appreciate the feedback. but unless I get any response from any of the repo maintainers, I really don't want to spend any more time on this.

hinogi: Ah, now I see. https://nodejs.org/docs/latest-v15.x/api/module.html#module_module_createrequire_filename this could be also used, maybe. Just to not be stuck with some commonJS syntax.
But otherwise you are absolutly correct.

pleerock:
We cannot use `require` since it will break ESM support.

netroy:
> We cannot use `require` since it will break ESM support.

But, [the actual driver modules are currently loaded using require](https://github.com/typeorm/typeorm/blob/master/src/platform/PlatformTools.ts#L36). is that because these are external packages ?

hinogi:
> We cannot use `require` since it will break ESM support.

I am also not sure what you are pointing at.

Looking at https://github.com/typeorm/typeorm/blob/022d2b5f622771349355f00a087c26c930db0d25/src/platform/PlatformTools.ts it seems to be used far and wide already.

hinogi:
> We cannot use `require` since it will break ESM support.

Also, the usage of `require` is what all the vite users are having PRs on.

netroy:
@pleerock Updated to use dynamic imports. 
but, I'm not certain if this change is backward compatible. `.driver` on `DataSource` used to be initialized in the constructor, but now it'll be initialized before the `.connect` method is called in `.initialize`. This would break the code for anyone who was using `datasource.driver` before calling `.initialize`.

netroy:
Since this seems like something that most users won't care about, I've decided to start managing a fork of typeorm. Closing this and #10221

PR #10196, opened 2023-07-10
fix: Remove dynamic require calls | Vite compatibility fix

Dukezo:
### Description of change

Removes all dynamic require calls from QueryBuilder class in order to make the library compatible with Vite.

### Pull-Request Checklist

- [x] Code is up-to-date with the `master` branch
- [x] `npm run format` to apply prettier formatting
- [ ] `npm run test` passes with this change
- [x] This pull request links relevant issues as Fixes #8738
- [ ] There are new or updated unit tests validating the change
- [ ] Documentation has been updated to reflect this change
- [x] The new commits follow conventions explained in [CONTRIBUTING.md](https://github.com/typeorm/typeorm/blob/master/CONTRIBUTING.md)

PR #10273, opened 2023-08-10
resolve circular dependency when using vite

vorujack:
### Description of change

There is a problem when using typeorm with vite. in QueryBuilder typeorm used require which is not available in vite builded output. 

I declare a Registery for all query types and fill it when datasource created and remove all required queries.

also there are two problem when using latest version of capacitor

* In capacitorQueryRunner line 65 if query only contain one command without any space command extracted wrong and code does not correct query
* in line 84 also PRAGMA must executed with run command not query

### Pull-Request Checklist


hinogi:
I would argue that probably `vite-require` would be a better option and less intrusive. Otherwise, I would opt in for an alias for `require` at the places where it is used. Ideally import syntax would be great to keep the laziness, but that would also change the public API as mentioned in other PRs from sync to async.

hinogi:
duplication of https://github.com/typeorm/typeorm/pull/10196 I think

vorujack:
I test `vite-require` and some other similar plugins. but they do not fix the problem. also alias for require not a good option.

as you can see in this merge there are no changing of sync call to async.

vorujack:
I fix some other bugs than #10196 in this merge

jbsn94:
Will this pr to be merged?, i'm also having the same problem using vite and the `vite-require` did not solve my problem.

pleerock:
Since I already merged #10196, this PR needs to be changed to avoid conflicts with #10196

Nickolas-:
This pull request breaks Capacitor with typeorm on iOS. 
image
Issue that pragma commands executes now as run command that cause sqlite result 100

rkreutzer:
Same issue with result error 100 on Android. Removing PRAGMA from the CapacitorQueryRunner driver files works for me.

PR #10289, opened 2023-08-17
feat: nullable embedded entities

### LINKED ISSUE #3913: Cannot set embedded entity to null
**Issue type:**

[ ] question
[x] bug report
[ ] feature request
[ ] documentation issue

**Database system/driver:**

[ ] `cordova`
[ ] `mongodb`
[ ] `mssql`
[ ] `mysql` / `mariadb`
[ ] `oracle`
[x] `postgres`
[ ] `cockroachdb`
[ ] `sqlite`
[ ] `sqljs`
[ ] `react-native`
[ ] `expo`

**TypeORM version:**

[x] `latest`
[ ] `@next`

Check out this small repository: https://github.com/nomadoda/typeorm-embedded-bug

I have an embedded entity:
```
export class Salary {
  @Column("decimal", { nullable: true })
  public amount: number;
}
```

And I'm using it in this entity:
```
@Entity()
  @PrimaryGeneratedColumn()
  id: number;

  @Column(() => Salary)
  salary?: Salary | null;
}
```

Consider this:
```
createConnection()
  .then(async connection => {
    await connection.synchronize(true);
    const salary = new Salary();
    salary.amount = 120;
    const employee = Employee.create({
      salary
    });
    await employee.save();
    console.log("Saved a new employee with id: " + employee.id);

    console.log("Set salary to null and save");
    employee.salary = null;
    await employee.save();

    console.log("Get employee");
    await employee.reload();
    console.log(employee.salary);
    // Salary { amount: '120' }
  })
  .catch(error => console.log(error));
```

**Expected outcome**

Setting salary to null should set all embedded fields in database to null.

**Actual outcome**

Salary did not change on update

andreamouraud:
Closes #3913

### Description of change
This PR introduces a breaking change by modifying the embedded entities behaviour to make them nullable (maybe enable it under a specific config?)

When dealing with embedded entities where all columns are `null`, we can consider the entire entity to be `null`
The same principle applies when setting `null` as a value for the embedded entity while updating, it sets all the embedded columns to `null`

### Pull-Request Checklist

### DIFF
--- a/src/query-builder/transformer/RawSqlResultsToEntityTransformer.ts
+++ b/src/query-builder/transformer/RawSqlResultsToEntityTransformer.ts
@@ -252,15 +252,19 @@ export class RawSqlResultsToEntityTransformer {
                 hasData = true
         })
         // Set all embedded column values to null if all their child columns are null
-        metadata.embeddeds.forEach((embedded) => {
-            if (
-                Object.values(entity[embedded.propertyName]).every(
-                    (value) => value == null,
-                )
-            ) {
-                entity[embedded.propertyName] = null
-            }
-        })
+        if (entity) {
+            metadata.embeddeds.forEach((embedded) => {
+                if (
+                    embedded.propertyName in entity &&
+                    typeof entity[embedded.propertyName] === "object" &&
+                    Object.values(entity[embedded.propertyName]).every(
+                        (value) => value == null,
+                    )
+                ) {
+                    entity[embedded.propertyName] = null
+                }
+            })
+        }
         return hasData
     }
 
--- a/src/query-builder/transformer/RawSqlResultsToEntityTransformer.ts
+++ b/src/query-builder/transformer/RawSqlResultsToEntityTransformer.ts
@@ -251,6 +251,16 @@ export class RawSqlResultsToEntityTransformer {
                 // we don't mark it as has data because if we will have all nulls in our object - we don't need such object
                 hasData = true
         })
+        // Set all embedded column values to null if all their child columns are null
+        metadata.embeddeds.forEach((embedded) => {
+            if (
+                Object.values(entity[embedded.propertyName]).every(
+                    (value) => value == null,
+                )
+            ) {
+                entity[embedded.propertyName] = null
+            }
+        })
         return hasData
     }
 
--- a/src/query-builder/transformer/RawSqlResultsToEntityTransformer.ts
+++ b/src/query-builder/transformer/RawSqlResultsToEntityTransformer.ts
@@ -254,18 +254,36 @@ export class RawSqlResultsToEntityTransformer {
         // Set all embedded column values to null if all their child columns are null
         if (entity) {
             metadata.embeddeds.forEach((embedded) => {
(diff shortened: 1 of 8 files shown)


src/query-builder/transformer/RawSqlResultsToEntityTransformer.ts
```diff
@@ -251,6 +251,20 @@ export class RawSqlResultsToEntityTransformer {
                 // we don't mark it as has data because if we will have all nulls in our object - we don't need such object
                 hasData = true
         })
+        // Set all embedded column values to null if all their child columns are null
+        if (entity) {
+            metadata.embeddeds.forEach((embedded) => {
+                if (
+                    embedded.propertyName in entity &&
+                    typeof entity[embedded.propertyName] === "object" &&
+                    Object.values(entity[embedded.propertyName]).every(
```
hinogi, line 260: Exactly what `.every` is for 👍

src/query-builder/transformer/RawSqlResultsToEntityTransformer.ts
```diff
@@ -251,6 +251,20 @@ export class RawSqlResultsToEntityTransformer {
...
+                        (value) => value == null,
```
hinogi, line 261:
```suggestion
                        (value) => value === null || value === undefined,
```

hinogi (changes requested):
strict null check

src/query-builder/transformer/RawSqlResultsToEntityTransformer.ts
andreamouraud: The unstrict check was intended, if we're getting an `undefined` value (it may not even be possible, didn't have a look at the underlying implementation) from the database it should still considered as a `null` in that use case

hinogi: In that case, I would make it explicit then.
```suggestion
                        (value) => value === null || typeof value === 'undefined',
```

andreamouraud: would argue that this is the use case for unstrict equality 😄 
had to rework it a bit for embedded entities in [73e4db3](https://github.com/typeorm/typeorm/pull/10289/commits/73e4db35349c6161606ac9aaff86786ac66e2ab1), added a comment to explicit it instead

MatthieuVegreville:
would be great to merge this! is that planned in an upcoming release of typeORM?

arieller:
Can this be merged?

DevGJGlobal:
Hopefully this will be merged soon!

pleerock:
I'll merge it despite it being a theoretical breaking change, because:

If you had such code in your codebase:

```ts
    @Column(() => NestedEmbedded)
    nested: NestedEmbedded | null
}
```

it couldn't work properly, because `null` cannot be skipped - it's not `undefined`. 

The behavior under runtime value of `null` also was inappropriate, since it was not supported.

pleerock:
Thank you for contribution! 🎉

Issue #10283, opened 2023-08-15
Reduce node_modules package size for AWS Lambda deployments


julienmunio:
### Feature Description

I'm facing build issues on my serverless cloud stack due to the weight of the TypeORM dependency package. This is because TypeORM depends on the date-fns library, which is quite large. This feature proposal outlines a plan to reduce the size of the TypeORM dependency package by switching to the dayjs library, which is much smaller. This will make it easier to add other dependencies to my project and reduce the risk of build failures.

### The Solution

The solution to this problem is to switch to the dayjs library. Dayjs is a lightweight alternative to date-fns that is much smaller in size. To switch to dayjs, we will need to update a single method DateUtils.mixedDateToDate()

### Considered Alternatives

As mentioned in the code comments, an alternative solution is not to use an external lib. But this generates additional code that is not useful to maintain.

### Relevant Database Driver(s)

- [ ] aurora-mysql
- [ ] aurora-postgres
- [ ] better-sqlite3
- [ ] cockroachdb
- [ ] cordova
- [ ] expo
- [ ] mongodb
- [ ] mysql
- [ ] nativescript
- [ ] oracle
- [ ] postgres
- [ ] react-native
- [ ] sap
- [ ] spanner
- [ ] sqlite
- [ ] sqlite-abstract
- [ ] sqljs
- [ ] sqlserver

### Are you willing to resolve this issue by submitting a Pull Request?

Yes, I have the time, and I know how to start.

NextThread:
@julienmunio , hey are you working on this?

julienmunio:
This PR is a suggestion to switch from Date-fns to DayJs. As remember this lib is used in a single method.

htakemoto:
It is true that date-fns file size is roughly 24M which is so huge. I also use NestJS + TypeORM in Lambda and everything worked fine up to NestJS v8, but I currently face Lambda layer size issue with NestJS v10. One big culprit is date-fns being depended by typeorm. Please remove/replace date-fns library.

PR #10306, opened 2023-08-25
Migration of DateUtils to DayJs

julienmunio:
### Description of change #10283 

Migration to DayJs of method DateUtils.mixedDateToDate(). This was this only one method using date-fns


hinogi (changes requested):
since date-fns is tree-shakeable this shoud not make a difference. I don't see a need for this change.
image
according to bundlephobia the single parseISO would be also smaller than dayjs.

pleerock:
`date-fns` is huge and it really can lead to the problems for some use-cases. I'm okay to change the implementation. 

This looks like a good alternative, and since all tests are passing that probably means we are good with this change.

RodoVJ:
@julienmunio Hi, my team uses TypeGraphQL and ever since this change I've been getting these errors:

```
Unable to serialize value '2024-02-11T04:28:24.670Z' as it's not an instance of 'Date'
Unable to serialize value '2024-02-11T04:28:24.670Z' as it's not an instance of 'Date'
Unable to serialize value '2024-02-13T05:42:41.791Z' as it's not an instance of 'Date'
```

This is coming from a column in our database that stores dates. I'm hoping you guys could provide some insight into why this change may be causing this. I appreciate your input :)

PR #10314, opened 2023-08-28
expose countDocuments in mongodb

FxEasy:
### Description of change

This change exposes countDocuments at the `MongoRepository` layer assisting nestjs applications wanting to move to the non-deprecated version.

Additionally this could be used within `executeFindAndCount` in the MongoEntityManager to follow what's recommended at the driver level.

It seems like there was an optimization in the driver to use `countDocuments` reverted 4 months ago to split them into their own functions but wasn't moved up to the respective manager and repositories.

Ideally MongoRepository should be using countDocuments internally for the `count` overload from `Repository` but I think this is a good start for simply exposing it.

### Pull-Request Checklist

I'm not quite sure how to write good tests expressing this change. Some help on this front would be appreciated.

- [ ] `npm run test` passes with this change -- will update when I get all the configs setup
- [ ] This pull request links relevant issues as `Fixes #0000` -- N/A
- [ ] There are new or updated unit tests validating the change -- unsure where to put tests for something like this

### DIFF
--- a/docs/mongodb.md
+++ b/docs/mongodb.md
@@ -284,6 +284,10 @@ Perform a bulkWrite operation without a fluent API.
 
 Count number of matching documents in the db to a query.
 
+### `countDocuments`
+
+Count number of matching documents in the db to a query.
+
 #### `createCollectionIndex`
 
 Creates an index on the db and collection.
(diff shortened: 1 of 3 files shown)


AlexMesser:
thank you for contribution!

PR #10343, opened 2023-09-07
Many to many er dup entry #5704

### LINKED ISSUE #5704: Many-to-many gives error ER_DUP_ENTRY everytime I save. This one also related to inverseJoinColumn.
**Issue type:**

[x ] bug report

**Database system/driver:**

[x] `mysql` / `mariadb`

**TypeORM version:**

[x] `latest`

``` ts
@Entity()
export class Tag {

  @PrimaryGeneratedColumn()
  id: number;

  @PrimaryColumn()
  title: string;

  @ManyToMany(() => Article, article => article.tags)
  articles: Article[]

}

// ---

@Entity()
export class Article {

  @PrimaryGeneratedColumn()
  id: number;

  @Column()
  title: string;

  @ManyToMany(() => Tag, tag => tag.articles)
  // if no inverseJoinColumn here, it works fine but I need it
  // because in Tag, there are 2 primary columns but I only need one, here.
  @JoinTable({
    inverseJoinColumn: {
      referencedColumnName: 'id'
    }
  })
  tags: Tag[]

}

// ---

createConnection().then(async conn => {

    /* FAKE DATA , uncomment this if u need a fake data
    const article = new Article();
    article.title = 'test'
    const tag = new Tag();
    tag.title = 't';
    await conn.manager.save(tag)
    article.tags = [tag]
    await conn.manager.save(article)
    /* --- */

    const articleRepo = conn.getRepository(Article)
    const one = await articleRepo.findOne(1, { relations: ['tags'] });
    // no error if no load 'tags'
    // I'm not doing any editing, but it still produces error
    await articleRepo.save(one)

    await conn.close()
    console.log('herer')

}).catch(error => console.log(error));

```

This is the error:
```
ER_DUP_ENTRY: Duplicate entry '1-1' for key 'PRIMARY'`.
```

After I look at SQL, I see: 
```sql
INSERT INTO `article_tags_tag`(`articleId`, `tagId`) VALUES (?, ?) -- PARAMETERS: [1,1]
```
There shouldn't be any error because I didn't do any editing at all.

growms:
### Problem found
In the method `buildForSubjectRelation` of `ManyToManySubjectBuilder`, in order to know if it is needed to insert the relation rows, `databaseRelatedEntityIds` is compared with `relatedEntityRelationIdMap` (line 154). The problem is that `databaseRelatedEntityIds` is not always an "EntityIdMap". If one is overloading the entity in the `afterLoad` event of an EventSubscriber for example then the object compared would contains additionnal field(s) and the ids won't match. This means that the relation is considered has not existing in the DB and the program tries to insert it. And as the relation already exists, a Mysql error is thrown...

### Description of change

 To fix this, in this commit, we ensure that `databaseRelatedEntityIds` is containing an "EntityIdMap" by using `relation.inverseEntityMetadata.getEntityIdMap(overloadedObject)`.

### Pull-Request Checklist

- [x] This pull request links relevant issues as `Fixes #0000` -> `Closes #5704` as mentionned in CONTRIBUTING file

### DIFF
--- a/src/persistence/subject-builder/ManyToManySubjectBuilder.ts
+++ b/src/persistence/subject-builder/ManyToManySubjectBuilder.ts
@@ -99,9 +99,14 @@ export class ManyToManySubjectBuilder {
         // if subject don't have database entity it means all related entities in persisted subject are new and must be bind
         // and we don't need to remove something that is not exist
         if (subject.databaseEntity) {
-            const databaseRelatedEntityValue = relation.getEntityValue(subject.databaseEntity);
+            const databaseRelatedEntityValue = relation.getEntityValue(
+                subject.databaseEntity,
+            )
             if (databaseRelatedEntityValue) {
-                databaseRelatedEntityIds = databaseRelatedEntityValue.map((e:any) => relation.inverseEntityMetadata.getEntityIdMap(e));
+                databaseRelatedEntityIds = databaseRelatedEntityValue.map(
+                    (e: any) =>
+                        relation.inverseEntityMetadata.getEntityIdMap(e),
+                )
             }
         }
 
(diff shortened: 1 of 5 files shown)


hinogi (approved):
LGTM

AlexMesser:
thank you for contribution!

PR #7450, opened 2021-03-06
fix: empty entity when query with nested relations

filipw01:
Closes: #7041

### Description of change

If there is a nested join for example User->OrganizationMembership->Organization and there is no OrganizationMembership for an User currently 
- it's returning empty Entity (all fields are nulls) for oneToOne relations
- it's returning an array with one element - empty Entity for many relations
the new behavior
- returns null for oneToOne relations
- returns empty array for many relations

Example old
User{id: "1", name: "me", membership: {id: null, organization: null}}
User{id: "1", name: "me", membership: [{id: null, organization: null}}]

Example new
User{id: "1", name: "me", membership: null}
User{id: "1", name: "me", membership: []}

Fixes #7041 
Fixes #3163
Fixes #3145

### Pull-Request Checklist

- [x] Code is up-to-date with the `master` branch
- [x] `npm run lint` passes with this change
- [x] `npm run test` passes with this change
- [x] This pull request links relevant issues as `Fixes #0000`
- [x] There are new or updated unit tests validating the change
- [ ] Documentation has been updated to reflect this change
- [x] The new commits follow conventions explained in [CONTRIBUTING.md](https://github.com/typeorm/typeorm/blob/master/CONTRIBUTING.md)

PR #10349, opened 2023-09-12
perf: improve results transformer performance

mohd-akram:
### Description of change

Improve the performance of `RawSqlResultsToEntityTransformer`. This makes it about twice as fast in queries in my testing.

### Pull-Request Checklist

### DIFF
--- a/src/query-builder/transformer/RawSqlResultsToEntityTransformer.ts
+++ b/src/query-builder/transformer/RawSqlResultsToEntityTransformer.ts
@@ -23,6 +23,14 @@ export class RawSqlResultsToEntityTransformer {
      */
     private relationIdMaps: Array<{ [idHash: string]: any[] }>
 
+    private pojo: boolean
+    private selections: Set<string>
+    private aliasCache: Map<string, Map<string, string>>
+    private columnsCache: Map<
+        string,
+        Map<EntityMetadata, [string, ColumnMetadata][]>
+    >
+
     // -------------------------------------------------------------------------
     // Constructor
     // -------------------------------------------------------------------------
@@ -33,7 +41,14 @@ export class RawSqlResultsToEntityTransformer {
         protected rawRelationIdResults: RelationIdLoadResult[],
         protected rawRelationCountResults: RelationCountLoadResult[],
         protected queryRunner?: QueryRunner,
-    ) {}
+    ) {
+        this.pojo = this.expressionMap.options.includes("create-pojo")
+        this.selections = new Set(
+            this.expressionMap.selects.map((s) => s.selection),
+        )
+        this.aliasCache = new Map()
+        this.columnsCache = new Map()
+    }
 
     // -------------------------------------------------------------------------
     // Public Methods
@@ -46,20 +61,38 @@ export class RawSqlResultsToEntityTransformer {
     transform(rawResults: any[], alias: Alias): any[] {
         const group = this.group(rawResults, alias)
         const entities: any[] = []
-        group.forEach((results) => {
+        for (const results of group.values()) {
             const entity = this.transformRawResultsGroup(results, alias)
-            if (
-                entity !== undefined &&
-                !Object.values(entity).every((value) => value === null)
-            )
-                entities.push(entity)
-        })
+            if (entity !== undefined) entities.push(entity)
+        }
         return entities
     }
 
     // -------------------------------------------------------------------------
     // Protected Methods
     // -------------------------------------------------------------------------
+
+    /**
+     * Build an alias from a name and column name.
+     */
+    protected buildAlias(aliasName: string, columnName: string) {
+        let aliases = this.aliasCache.get(aliasName)
+        if (!aliases) {
+            aliases = new Map()
+            this.aliasCache.set(aliasName, aliases)
+        }
+        let columnAlias = aliases.get(columnName)
+        if (!columnAlias) {
+            columnAlias = DriverUtils.buildAlias(
+                this.driver,
+                undefined,
+                aliasName,
+                columnName,
+            )
+            aliases.set(columnName, columnAlias)
+        }
+        return columnAlias
+    }
 
     /**
      * Groups given raw results by ids of given alias.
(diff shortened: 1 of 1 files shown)


marcelMaier (approved):

src/query-builder/transformer/RawSqlResultsToEntityTransformer.ts
```diff
@@ -46,21 +61,39 @@ export class RawSqlResultsToEntityTransformer {
     transform(rawResults: any[], alias: Alias): any[] {
         const group = this.group(rawResults, alias)
         const entities: any[] = []
-        group.forEach((results) => {
+        for (const results of group.values()) {
```
hinogi, line 64: What about using filter and not having to use push ?

pleerock:
I need people to test and confirm the results. At least 10 reviewers.

mohd-akram:
Thanks for looking at this @pleerock.

Just to give some more context, the performance improvement is largely straightforward - rather than re-calculate column metadata on every result - which is expensive - they are only calculated and cached once the first time, then used directly while processing the rest of the results. So the larger the number of results returned by the query, the larger the improvement.

There are other small improvements such as using for loops rather than `.forEach` everywhere to make it easier to diagnose performance (avoids an anonymous function context), and prevents inefficiencies such as `.filter` followed by `.forEach` which causes double iteration.

src/query-builder/transformer/RawSqlResultsToEntityTransformer.ts
mohd-akram: This is not just a filter, it's a map and a filter, so we do both in one shot.

src/query-builder/transformer/RawSqlResultsToEntityTransformer.ts
```diff
@@ -46,21 +61,39 @@ export class RawSqlResultsToEntityTransformer {
...
             const entity = this.transformRawResultsGroup(results, alias)
-            if (
-                entity !== undefined &&
-                !Object.values(entity).every((value) => value === null)
```
mohd-akram, line 53: This check is removed because it is expensive and is unnecessary. It was originally added in #7450 along with accompanying tests, and the tests continue to pass even after removing it. Additional checks have been added to the code since then which keep track of entities where all values are null (via `hasData` variables).

mohd-akram:
Rebased to fix conflict.

src/query-builder/transformer/RawSqlResultsToEntityTransformer.ts
```diff
@@ -393,61 +390,84 @@ export class RawSqlResultsToEntityTransformer {
...
+    private getColumnsToProcess(aliasName: string, metadata: EntityMetadata) {
+        let metadatas = this.columnsCache.get(aliasName)
+        if (!metadatas) {
+            metadatas = new Map()
+            this.columnsCache.set(aliasName, metadatas)
+        }
+        let columns = metadatas.get(metadata)
+        if (!columns) {
+            columns = metadata.columns
+                .filter(
```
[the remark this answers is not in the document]
mohd-akram: In the other sections it was more about removing `forEach` for clearer profiling (dropping `filter` in a couple of instances was a side effect). I'm not sure it's worth the loss in readability in this case since it's well-bounded (there's a small, constant number of columns) and this code is only run once before being cached.

almassapargali:
This code seems straightforward and re-calculating column metadata on every result seems really redundant.

ertl:
I tried this fix in my codebase, where we read a lots of data.

As you can see in the following images the huge block `executeEntitiesAndRawResults` has shrunken.


Without fix: 
Execution Time: 3min56sec
image

With fix:
Execution Time: 2min35sec
image

mohd-akram:
Now that this project is active again, maybe this could be merged? @michaelbromley

michaelbromley:
@mohd-akram I've added it as a proposal to our roadmap.

jorenvandeweyer (approved):

src/query-builder/transformer/RawSqlResultsToEntityTransformer.ts
```diff
@@ -210,47 +230,22 @@ export class RawSqlResultsToEntityTransformer {
...
-                    DriverUtils.buildAlias(
-                        this.driver,
-                        undefined,
-                        alias.name,
-                        column.databaseName,
-                    )
-                ]
-            if (value === undefined || column.isVirtual) return
+        const result = rawResults[0]
+        for (const [key, column] of this.getColumnsToProcess(
```
SylvainMarty, line 234: Maybe we should rename `key` as `columnAlias`?
It would make the code a bit more explicit since the method name `getColumnsToProcess()` is not indicating what `key` is.

SylvainMarty (commented):
Excellent work!
I only have a minor comment about some variable name.

I don't see how this PR could break any existing feature, the code is pretty much 1:1 with the previous version.
I checked the coverage: there are more than 300 test suites that are covering the repository methods, which seems good enough.

The only "big" difference with this code is the `RawSqlResultsToEntityTransformer` memory footprint. Since we don't know the hardware limitations of TypeORM users, it may be a good idea to also check the difference in memory consumption introduced by this code change?

gioboa (approved):
It looks great to me. Nice improvement, thanks

SylvainMarty:
Hey guys! I spent some time benchmarking the memory usage differences between these changes and the latest version of TypeORM.

From what I've seen, this PR actually improves the memory consumption of TypeORM.

The full results and protocol are available here: https://github.com/SylvainMarty/typeorm-memory-benchmark.

Let me know what you think of it. 👍

michaelbromley (approved):

michaelbromley:
Thank you @mohd-akram for this excellent contribution, and to everyone who reviewed & tested it. Especially @SylvainMarty for your careful work validating the memory implications. This will certainly be the headline improvement in the new patch release we are working on! 🙌

src/query-builder/transformer/RawSqlResultsToEntityTransformer.ts
Hareloo: I don't know about the tests but this just messed up my code, I am now getting for relations sometimes objects with null values in all properties instead of an empty array, had to downgrade to typeorm 0.3.20 because of this...

mohd-akram: @Hareloo What driver are you using?

PR #10395, opened 2023-10-03
RFC: query-builder: when .where() is called twice, throw an error

fumieval:
### Description of change

I find it rather dangerous that `builder.where(oldCondition).where(condition)` silently discards oldCondition. If one accidentally uses `.where` instead of `.andWhere`, they might leak __confidential data__, or even __wipe their table entirely__.

This PR prevents such problem by throwing an error when the builder already contains a condition.

### Pull-Request Checklist


pleerock:
This is breaking change.

Some use cases might include using `.where` statement to override original statements, however in most cases I would say it's a bad design.

The main concern about this PR - it can be too much breaking change which we try to avoid.

gioboa (commented):
Thanks @fumieval for your help but as @pleerock said we have to close it. I'm so sorry.

PR #10398, opened 2023-10-03
Update working-with-repository.md

teolampos:
We access the DataSource instance, so the documentation is wrong.

### Pull-Request Checklist


hinogi (changes requested):
Both is correct since there is `manager.dataSource` as well as `dataSource.manger` no need for a change in my opinion because it is more likely to just step to the manager instead of the datasource.

pleerock:
exactly, both are correct. So I don't see a need to merge this PR.

PR #10419, opened 2023-10-14
Default replication mode

airhorns:
### Description of change

For many applications that want to use read slaves, data consistency isn't super important, and so all reads can be sent by default to a read slave. For some applications though, it's only select queries that can be sent to read slaves as end-users of the application expect data consistency almost everywhere. This new setting lets apps like those change the default such that query builders need to opt *in* to sending to a read slave, instead of opt out.

For existing replication users, the default for queries is unchanged (send to the slave) but this allows folks who care like me to change the default behaviour. 

### Pull-Request Checklist


pleerock:
Thanks for such a solid implementation.

PR #10443, opened 2023-10-18
fix: do not skip undefined or null properties in where conditions

mpalmerlee:
### Description of change

Changes behavior of `getWherePredicateCondition` and `buildWhere` methods to match the behavior in typeorm 0.2, which is to apply an IsNull() condition to null values and not ignore undefined values so that when passing undefined queries do not return any data (null or []) rather than excluding that condition which is a vector for security issues.

I would be ok with a behavior where null values return no results as well, just as long as conditions aren't skipped then this is not a vector for security issues / cross-user data leakage, etc...

We are in the process of upgrading from TypeOrm 0.2.x to 0.3.x and discovered this behavior change in 0.3.x that makes it prohibitively difficult to finish the upgrade without risking leaking incorrect data or possibly causing updates to incorrect data. Similar to the PR here: https://github.com/typeorm/typeorm/pull/9487 except doesn't make this behavior configurable for simplicity. This is of course a breaking change for 0.3.x and is perhaps a candidate for 0.4.x, we'll have to use the forked version to finish the upgrade and perhaps that alternative version will help other people upgrade from 0.2.x, if it is decided that this is not how typeorm should behave long term that is understandable yet disappointing for those of us who have built large systems on top of this great ORM.

I'm not sure what documentation changes should be made to reflect that but I'm happy to make the necessary changes as needed.

Fixes: #9316 

### Pull-Request Checklist

### DIFF
--- a/src/query-builder/QueryBuilder.ts
+++ b/src/query-builder/QueryBuilder.ts
@@ -1553,13 +1553,11 @@ export abstract class QueryBuilder<Entity extends ObjectLiteral> {
                     parameters: [aliasPath, ...parameters],
                 }
             }
-            // } else if (parameterValue === null) {
-            //     return {
-            //         operator: "isNull",
-            //         parameters: [
-            //             aliasPath,
-            //         ]
-            //     };
+        } else if (parameterValue === null) {
+            return {
+                operator: "isNull",
+                parameters: [aliasPath],
+            }
         } else {
             return {
                 operator: "equal",
--- a/src/query-builder/SelectQueryBuilder.ts
+++ b/src/query-builder/SelectQueryBuilder.ts
@@ -4222,7 +4222,9 @@ export class SelectQueryBuilder<Entity extends ObjectLiteral>
         } else {
             let andConditions: string[] = []
             for (let key in where) {
-                if (where[key] === undefined || where[key] === null) continue
+                // NOTE: preserving typeorm v0.2 behavior for query safety,
+                //  nulls will be treated as IsNull, undefined values should never return results
+                //if (where[key] === undefined || where[key] === null) continue
 
                 const propertyPath = embedPrefix ? embedPrefix + "." + key : key
                 const column =
--- a/test/functional/find-options/empty-properties/entity/Post.ts
+++ b/test/functional/find-options/empty-properties/entity/Post.ts
@@ -8,6 +8,6 @@ export class Post {
     @Column()
     title: string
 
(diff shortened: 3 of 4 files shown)


EdisonSL:
any eta to merge this pr?

pleerock:
Current behavior is by design and won't be changed.

Current way of handling `null` / `undefined` values makes more sense (both logical and intuitive) than what we had in `0.2.x`. I understand you'll have to come to this question with a clear thought and make necessary changes to your codebase, but it's not that hard to do. 

Regarding to security issues - there are no security issues. You just need to properly design and write the code and of course have a sanitization logic before you start using your data against database.

bombillazo:
@pleerock, I disagree; I have yet to see clear reasoning for why the new behavior is better. On the contrary, #9316 is clear evidence that the current way of handling `null`/ `undefined` is not intuitive at all and requires more work from developers to avoid pitfalls that have been ironically engineered into the library.

anlopz:
@pleerock I strongly disagree; is not intuitive that passing undefined or null it returns the first value. At least this behavior must be documented, there are many issues regarding this because it is not intuitive.

LandroNoInc:
@pleerock if I have an entity `User` and am running `User.findOneBy({ id: null })`, this logically translates to a query of either `SELECT * FROM users WHERE id is null` or `SELECT * FROM users WHERE id = null`. Neither of these queries will return any data, let alone the first item in the database. Where is the logic that a null ID value somehow translates to the first item in the database?

mpalmerlee:
@LandroNoInc Here is the line: https://github.com/typeorm/typeorm/pull/10443/files#diff-52bea788267888fdbb5f795f0dfb61504f16bd43e0afdfa28848abb4435121b5L4225

bombillazo:
@LandroNoInc It's not. See [here](https://github.com/typeorm/typeorm/issues/9316#issuecomment-1299328534) on how I believe it should behave to fix this whole mess once and for all.

jezaman:
Can't believe the stubbornness of thinking this is right and refusing to change it if your entire community it crying out for it to be changed. It's OK to be wrong, just listen to your users. 90% of my database bugs are this issue. This sole issue is indeed enough cause of frustration for us to move away from TypeORM. 

@pleerock saying "You just need to properly design and write the code" is just the cherry on the cake. ..  
Well how about YOU properly design and write the code of TypeORM and get rid of this bug?

PR #10449, opened 2023-10-20
bug fix: property `where` creates incorrect sql query

X932:
bug fix: when parameters are undefined with operator OR throws exception instead data from table

### Description of issue
telegram-cloud-photo-size-2-5293996961751487555-y
if all parameters are undefined it creates sql query like: `select * from table where ()`. `()` - tadaaam )

### DIFF
--- a/src/query-builder/SelectQueryBuilder.ts
+++ b/src/query-builder/SelectQueryBuilder.ts
@@ -4489,6 +4489,11 @@ export class SelectQueryBuilder<Entity extends ObjectLiteral>
             }
             condition = andConditions.join(" AND ")
         }
+
+        if (condition === '()') {
+            condition = '(1=1)'
+        }
+
         return condition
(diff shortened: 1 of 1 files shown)


HushangJumaev (approved):

pleerock:
Ideally, instead of `(1=1)`, we need to prevent adding empty string at all, e.g. prevent brackets from being inserted if there are no where criteria. Also, once you finish with implementation, you must add test coverage for this change. Thanks.

X932:
> Ideally, instead of `(1=1)`, we need to prevent adding empty string at all, e.g. prevent brackets from being inserted if there are no where criteria. Also, once you finish with implementation, you must add test coverage for this change. Thanks.

got it. Thanks for feedback

pleerock:
Closing as there is no activity on this PR.

PR #10477, opened 2023-11-10
feat: introduce FindOptionsRelations<T> to FindTreeOptions

caiorosa-seenons:
### Description of change

Introduces `FindOptionsRelations<Entity> | FindOptionsRelationByString` to `FindTreeOptions`
Tree Repositories currently only loads Relations by primitive strings, which is not only going to be deprecated, but also does not use the proper type safety.

The logic under `FindOptionsUtils.applyRelationsRecursively` is very similar to what happens inside the `SelectQueryBuilder.buildRelations` but I could not find a way to deduplicate and abstract logic without making too many changes as they live in different contexts and have some differences that I do not fully understand yet.

### Pull-Request Checklist

### DIFF
--- a/docs/tree-entities.md
+++ b/docs/tree-entities.md
@@ -231,10 +231,14 @@ There are other special methods to work with tree entities through `TreeReposito
 -   `findTrees` - Returns all trees in the database with all their children, children of children, etc.
 
 ```typescript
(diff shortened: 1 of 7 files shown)


pleerock:
can we get rid of `any`-s used as default in generics?

Issue #10493, opened 2023-11-17
Broken migrations for indices on TIMESTAMP WITH TIMEZONE Oracle Database columns


gabrielkim13:
### Issue description

When creating indices for columns of type TIMESTAMP WITH TIMEZONE, on Oracle Database, a virtual column named 'SYS_NC0000*$' will be created in order for the DB to be able to index the actual column. Since the index will reference this virtual column, any further TypeORM migrations will be broken, attempting to: drop the index; drop the virtual column - which fails; re-create the previously dropped index.

### Expected Behavior

After running the first migration, which creates the index on the TIMESTAMP WITH TIMEZONE column, further migrations shouldn't attempt to re-create it and just ignore the virtual column.

### Actual Behavior

A second run of the MCVE will result in the following error:

```
❯ npm run start

> typeorm-mcve@0.0.1 start
> ts-node src/index.ts

QueryFailedError: ORA-00942: table or view does not exist
    at OracleQueryRunner.query (/home/gabrielkim13/Desktop/github.com/gabrielkim13/typeorm-mcve/src/driver/oracle/OracleQueryRunner.ts:274:19)
    at async /home/gabrielkim13/Desktop/github.com/gabrielkim13/typeorm-mcve/src/driver/oracle/OracleQueryRunner.ts:2576:49
    at async Promise.all (index 5)
    at async /home/gabrielkim13/Desktop/github.com/gabrielkim13/typeorm-mcve/src/driver/oracle/OracleQueryRunner.ts:2385:33
    at async Promise.all (index 0)
    at async OracleQueryRunner.loadTables (/home/gabrielkim13/Desktop/github.com/gabrielkim13/typeorm-mcve/src/driver/oracle/OracleQueryRunner.ts:2368:
16)
    at async OracleQueryRunner.getTables (/home/gabrielkim13/Desktop/github.com/gabrielkim13/typeorm-mcve/src/query-runner/BaseQueryRunner.ts:158:29)
    at async RdbmsSchemaBuilder.build (/home/gabrielkim13/Desktop/github.com/gabrielkim13/typeorm-mcve/src/schema-builder/RdbmsSchemaBuilder.ts:92:13)
    at async DataSource.synchronize (/home/gabrielkim13/Desktop/github.com/gabrielkim13/typeorm-mcve/src/data-source/DataSource.ts:335:9)
[...]

PR #10506, opened 2023-11-19
fix: migrations on indexed TIMESTAMP WITH TIME ZONE Oracle columns

gabrielkim13:
ignore auto-generated virtual columns and attempt to map them into their respective reference columns

Closes: #10493

### Description of change

Ignore automatically generated virtual columns on the Oracle driver, in order to avoid errors when generating migrations and / or synchronizing the entities / schemas to the database.

This is accomplished by ignoring any virtual columns that have the "USER_GENERATED" column set to "NO".

Also, since this mainly happens when a TIMESTAMP WITH TIME ZONE column is indexed, generating a virtual TIMESTAMP column, it is also necessary to infer that this virtual column actually refers to the original one, during indices discovery. This way, TypeORM won't attempt to recreate it.

A test case has been added to reproduce and fix the aforementioned issue.

### Pull-Request Checklist

  - Ran tests with only the "oracle" driver enabled, since that's the only set of files altered by this PR
  - N/A


AlexMesser:
thank you for contribution!

PR #10550, opened 2023-12-11
fix: sqlite simple-enum column parsing

ncpa0cpl:
### Description of change

Fixed a bug that was causing new migration to be generated even when no changes to the Model was made. Generating migration for a table with a column type `simple-enum` (in sqlite) would produce SQL CREATE TABLE query with a column like this:
```sql
  "col_name" varchar CHECK("col_name" IN ('FOO', 'BAR', 'BAZ')) NOT NULL,
```
When a TableColumn for this column is created, that CHECK statements was being incorrectly parsed (`','` was used as separator, but the statement included whitespaces not present in that separator.) Given the above column in db, instead of parsing it into `["FOO", "BAR", "BAZ"]` it was parsed into `["FOO', 'BAR', 'BAZ"]`, because of this syncing would attempt to update that table and create a new migration every single time.

To fix the issue I've changed how that CHECK statement is parsed, instead of splitting the string using some kind of separator you iterate over characters and evaluate each differently based on whether it is placed inside a quote or not, only commas outside quotes are treated as separators of values, and the whitespaces outside quotes are ignored, so it doesn't matter if there is any number of spaces before or after the comma.

### Pull-Request Checklist

### DIFF
--- a/src/driver/sqlite-abstract/AbstractSqliteQueryRunner.ts
+++ b/src/driver/sqlite-abstract/AbstractSqliteQueryRunner.ts
@@ -1453,20 +1453,10 @@ export abstract class AbstractSqliteQueryRunner
                         }
 
                         if (tableColumn.type === "varchar") {
-                            // Check if this is an enum
-                            const enumMatch = sql.match(
-                                new RegExp(
-                                    '"(' +
-                                        tableColumn.name +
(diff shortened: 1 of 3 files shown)


AlexMesser:
thank you for contribution!

PR #10595, opened 2024-01-03
build: set TS compile target to ES2021

alumni:
### Description of change

Setting the typescript compilation target to ES2021 (equivalent of NodeJS 16).

Also bumping up the TS version to 5.3, fixing type errors and cleaning up `Gulpfile.ts`

### Pull-Request Checklist


pleerock:
Nice, we also need to update `engines` in package.json and not sure where else we have information about node.js dependency. Probably would be good to specify somewhere in the documentation regarding to the minimal supported node.js version.

alumni:
@pleerock will do.

FYI I thought about extending `@tsconfig/node16`, but they enable strict mode by default which can cause some issues and they also set `moduleResolution` which might cause issues when changing the module type (since TS 5.2).

What bothers me is that the browser build is also using the main tsconfig - for some reason gulp-typescript doesn't work as expected. It didn't before either. Any ideas? As a workaround, we could create a tsconfig only for browsers that extends the main tsconfig.

I can edit the readme (previously it was mentioned ES5 and higher). Any preferences for the browser targets (ES version, since I'd rather not setup `browserslist`)?

test/functional/repository/find-options-locking/find-options-locking.ts
```diff
@@ -135,7 +136,7 @@ describe("repository > find options > locking", () => {
                     const originalQuery = entityManager.queryRunner!.query.bind(
                         entityManager.queryRunner,
                     )
-                    entityManager.queryRunner!.query = (...args: any[]) => {
+                    entityManager.queryRunner!.query = (...args: Parameters<QueryRunner['query']>) => {
```
alumni, line 139: This test should have used spies + called with argument. I'm not very familiar with Chai/SinonJS, so I'd rather not change it myself.

alumni:
> FYI I thought about extending @tsconfig/node16, but they enable strict mode by default which can cause some issues and they also set moduleResolution which might cause issues when changing the module type (since TS 5.2).

I ended up switched to `@tsconfig/node16` which enables strict mode. I had to disable a few strict rules, but now it should enforce stricter type checking than before - if you look at the changed files, there was some shady behavior which was detected by this.

> What bothers me is that the browser build is also using the main tsconfig - for some reason gulp-typescript doesn't work as expected

Nevermind, in `gulpfile.ts` we were never changing the compilation target, so it was ES2019 also for the browser build. Now it's ES2021 (inherited from Node16 default config) - which should be good enough for all recent browsers.

pleerock:
Let's merge it

Issue #3051, opened 2018-11-06
mongo transactions needed!


wzquyin:
**Issue type:**

[ ] feature request

**Database system/driver:**

[ ] `mongodb`

**TypeORM version:**

[ ] `latest`

**Steps to reproduce or a small repository showing the problem:**
hi,I like typeorm a lot
do you plan to support mongo transactions?

rustamwin:
#3472

imnotjames:
PRs that add support for transactions in MongoDB would be welcome.

lusatiro:
Up

lusatiro:
Does that work? https://orkhan.gitbook.io/typeorm/docs/transactions#creating-and-using-transactions

lusatiro:
@imnotjames, I'm thinking about it. We're using typeorm here at the company I work for and think about moving on to MikroORM 'cause it is more close to typeorm.

lusatiro:
> PRs that add support for transactions in MongoDB would be welcome.

@imnotjames, I'm thinking about it. We're using typeorm here at the company I work for and think about moving on to MikroORM 'cause it is more close to typeorm.

banikKGP:
Any update on the typeorm mongodb support. Or any workaround for this?

dominicarrojado:
Any update on this one?

uzenith360:
+1

Abramovick:
+1

robjr:
I'd be great to have this feature released soon, I'm also looking for this.

balmychan:
I too am looking for this feature.

tnkarno:
Me too.

magistrser:
+1

anjolaoluwaakindipe:
+1

balmychan:
I have tried to write code using transactions in the following form.

I am using NestJS, Here is an example in NestJS.
However, with EntityManager, the code should be the same.

```ts
@Injectable()
export class CreateMemberUsecase {
  constructor(
    @InjectEntityManager()
    private readonly entityManager: MongoEntityManager,
    @InjectRepository(Member)
    private readonly memberRepository: MongoRepository<Member>
  ) {}

  async execute() {
    const member = Member.new({ ... });

    const session = this.entityManager.mongoQueryRunner.databaseConnection.startSession();

    session.startTransaction();

    try {
      await this.memberRepository.insertOne(member, { session });
      // Other operations...

      session.commitTransaction();
    } catch (e) {
[...]

PR #10600, opened 2024-01-05
Initial attempt at transactions for mongodb

LegitPanda:
### Description of change
Adds transactions to mongodb. Although mongodb does allow transactions to be committed in any order I think this is an OK approach with transaction depth like how the others work - thoughts? I originally tried to change the signature of startTransaction() to return the session which mongo needs for each transaction but then that wouldn't work for a lot of things. Still testing if this works with migrations as well

### Pull-Request Checklist
- [x] This pull request links relevant issues as `Addresses #3051`


pleerock:
I need reviews from our mongodb users on this PR.

src/driver/mongodb/MongoQueryRunner.ts
```diff
@@ -545,24 +580,57 @@ export class MongoQueryRunner implements QueryRunner {
...
+
+        this.sessions[this.transactionDepth] = session
     }
 
     /**
      * Commits transaction.
      */
     async commitTransaction(): Promise<void> {
-        // transactions are not supported by mongodb driver, so simply don't do anything here
+        if (!this.isTransactionActive) throw new TransactionNotStartedError()
```
pringon, line 602: nit: Is this variable necessary given you can infer it from `transactionDepth`?

I know this is more expressive, but it adds an extra place where a bug may happen because it was not flipped and could be refactored into a function based on `transactionDepth`

src/driver/mongodb/typings.ts
```diff
@@ -4487,7 +4487,7 @@ export declare class MongoClient extends TypedEventEmitter<MongoClientEvents> {
      *
      * @see docs.mongodb.org/manual/reference/connection-string/
      */
-    connect(): Promise<this>
+    connect(url: string, options: any): Promise<this>
```
pringon, line 4490: We pass the url and options in the constructor.
This would also lead to a breaking change of the API

gioboa (commented):
@LegitPanda thanks for opening this PR.
I'm closing this for now because it's stale and with unresolved suggestions , feel free to create a new PR with your improvements. :muscle:

Issue #10596, opened 2024-01-04
QueryBuilder allowing multiple `where` calls is dangerous


Woodz:
### Issue description

Subsequent calls to `.where` in a QueryBuilder silently discards all previous `where` clauses

### Expected Behavior

I expect to never silently discard previous calls - subsequent calls to `.where` should throw an error or even better, use a Fluent API that doesn't allow multiple calls to `.where`

### Actual Behavior

Subsequent calls to `.where` in a QueryBuilder silently discards all previous `where` clauses

### Steps to reproduce

```
const user = await dataSource.manager
    .createQueryBuilder(User, "user")
    .where("user.firstName = :firstName, { firstName: "Timber" })
    .where("user.lastName = :lastName, { lastName: "Saw" })
    .getOne()
```

### My Environment

| Dependency          | Version  |
| ---                 | ---      |
| Operating System    | Irrelevant |
| Node.js version     | Irrelevant |
| Typescript version  | Irrelevant |
| TypeORM version     | 0.3.17  |

### Additional Context

Others have raised concerns: https://dev.to/rishit/optimizing-typeorm-tips-from-experience-part-2-use-where-with-care-40jp

### Relevant Database Driver(s)

- [ ] aurora-mysql
- [ ] aurora-postgres
- [ ] better-sqlite3
- [ ] cockroachdb
- [ ] cordova
- [ ] expo
- [ ] mongodb
- [ ] mysql
- [ ] nativescript
- [ ] oracle
- [ ] postgres
- [ ] react-native
- [ ] sap
- [ ] spanner
- [ ] sqlite
- [ ] sqlite-abstract
- [ ] sqljs
- [ ] sqlserver

### Are you willing to resolve this issue by submitting a Pull Request?

No, I don’t have the time and I’m okay to wait for the community / maintainers to resolve this issue.

Woodz:
Slightly related: allowing users to call `.orWhere` or `.andWhere` on a new QueryBuilder is also very confusing, e.g.
```
const user = await dataSource.manager
    .createQueryBuilder(User, "user")
    .orWhere("user.firstName = :firstName, { firstName: "Timber" })
    .getOne()
```

Suggestion: Split the fluent API in two:
1. When there are no `where` conditions then only `.where` can be called and transition to 2nd API
2. Where there are `where` conditions then only `orWhere`/`andWhere` can be called

alenap93:
[...]

PR #10605, opened 2024-01-06
fix: throw error if where is called multiple times

alenap93:
Closes: #10596

### Description of change
Throw an error if .where or .whereEntity is called multiple times (it discards all previous .where/.whereEntity)

### Pull-Request Checklist

### DIFF
--- a/src/query-builder/DeleteQueryBuilder.ts
+++ b/src/query-builder/DeleteQueryBuilder.ts
@@ -8,6 +8,7 @@ import { Brackets } from "./Brackets"
 import { DeleteResult } from "./result/DeleteResult"
 import { ReturningStatementNotSupportedError } from "../error/ReturningStatementNotSupportedError"
 import { InstanceChecker } from "../util/InstanceChecker"
+import { TypeORMError } from "../error"
 
 /**
  * Allows to build complex sql queries in a fashion way and execute those queries.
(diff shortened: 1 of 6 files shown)


alenap93:
circleCI seems to be slow, _ci/circleci: test (oracle) - Node v16 go in error cause: Too long with no output (exceeded 10m0s): context deadline exceeded_

pleerock:
I'm okay with this change, hovewever it can be a breaking change for some people and we cannot accept breaking changes right now. I'll keep this PR opened for the future.

pleerock:
duplicate of #10395 BTW

gioboa (commented):
Thanks for your help @alenap93 
I'll add this to the TypeORM board, it's a breaking change but we can merge it in future releases.

jovanadjuric:
I've taken advantage of this behavior more than once when I've had to override the previous `.andWhere` statements under some conditions that were checked at runtime.

Throwing the error for this at runtime would break queries for dozens of users, and putting restrictions on the composed queries would only make the query builder less flexible. While it is true that `.where` may be confused with `.andWhere`, its occurrences in all query builders are well documented and users are one click or hover away at all times from knowing about the behavior of the methods they are using.

For these reasons, I'd prefer this PR not to be merged. 

If the team decides to merge it instead, I suggest doing so in 0.4.0 rather than the next patch. The docs and comments will need to be updated as well.

alumni:
If we merge this, we'd need to add some parameter to allow overwriting the previous where clause. (Later edit: or some other method to reset all where conditions?)

I'm tending towards avoiding these complications.

PR #10608, opened 2024-01-07
feat: every driver's escapeQueryWithParameters method supports two mehods: three dot and when an array of values

dingiyan:
### Description of change
every driver's escapeQueryWithParameters method supports two mehods: three dot and when an array of values

### Pull-Request Checklist


alenap93:
Could you explain your feature/fix?

dingiyan:
> Could you explain your feature/fix?

I think we can directly determine how parameters should be used based on the type of value, rather than using three points.

For example: the parameter placeholder like  `:param1`, if the value is a string, the result is a string; if the value is an array like ['str1','str2'], the result like: `'str1', 'str2'`

this feature is like sequelize.

pleerock:
It wouldn't work. Sometimes you need to pass down to the same driver **_exactly an array_**, sometimes not, depend on your use case (your SQL query).

gioboa (commented):
Thanks @dingiyan for your help.
I'm closing this for now because it's stale and with unresolved comments, feel free to create a new PR with your improvements. :muscle:

PR #10630, opened 2024-01-15
fix: Hangup when load relations with relationLoadStrategy: query

### LINKED ISSUE #4738: Running simultaneous queries with relations stalls typeorm
**Issue type:**

[ ] question
[x] bug report
[ ] feature request
[ ] documentation issue

**Database system/driver:**

[ ] `cordova`
[ ] `mongodb`
[ ] `mssql`
[ ] `mysql` / `mariadb`
[ ] `oracle`
[x] `postgres`
[ ] `cockroachdb`
[ ] `sqlite`
[ ] `sqljs`
[ ] `react-native`
[ ] `expo`

**TypeORM version:**

[ ] `latest`
[x] `@next`

I ran into an issue where typeorm would just stop responding occasionally, and when looking at the postgres stats I noticed that typeorm kept 10 (the size of the pool) connections open and in idle and didn't seem to ever close them afterwards. 

The connections showed up as `Idle | WaitEvent: Client: ClientRead`. 

Some more investigations seem to indicate that it the issue presents itself when trying to find entities with relations (with the `getManager().find(Entity, options)` syntax). 

Increasing the pool size can help mitigate the problem, and using the query builder works as a workaround, but is obviously something we'd want to avoid as much as possible in favour of shorter, and clearer syntax.

I don't believe it's a config issue either

```js

let entities = ['./dist/api/src/entities/**/*.js'];

module.exports = {
  type: 'postgres',
  synchronize, // True locally
  ssl,
  host,
  database,
  username,
  password,
  entities,
  timezone: 'Z',
  logging,
  logger,
  maxQueryExecutionTime: 1000,
  extra: {
    poolSize, // 10
    idleTimeoutMillis: 5000, // Drop connections that are stalled
    connectionTimeoutMillis: 10000, // Drop connections that are stalled
  },

  cli: {
    entitiesDir: 'api/src/entities/**',
    migrationsDir: 'api/src/migration',
    subscribersDir: 'api/src/subscriber',
  },
};

```

```ts
/* eslint-disable @typescript-eslint/no-use-before-define */
import { createConnection, Entity, getManager, ManyToOne, OneToMany, PrimaryGeneratedColumn } from 'typeorm';

@Entity()
export class Entity2 {
  @PrimaryGeneratedColumn()
  id!: number;

  @ManyToOne(() => Entity1, ent1 => ent1.ent2)
  ent1!: Entity1;
}

@Entity()
export class Entity1 {
  @PrimaryGeneratedColumn()
  id!: number;

  @OneToMany(() => Entity2, ent2 => ent2.ent1)
  ent2!: Entity2[];
}

const getWithRelations = () => {
  getManager()
    .find(Entity1, { relations: { ent2: true } })
    .then(e => {
      console.log(e);
    });
};

const getSimple = () => {
  getManager()
    .find(Entity1)
    .then(console.log);
};

const getWithQb = () => {
  getManager()
    .getRepository(Entity1)
    .createQueryBuilder('ent')
    .leftJoinAndSelect('ent.ent2', 'ent2')
    .getMany()
    .then(console.log);
};

(async () => {
  await createConnection();

  const parent = await getManager().save(new Entity1());

  const child1 = new Entity2();
  child1.ent1 = parent;
  await getManager().save(child1);

  const child2 = new Entity2();
  child2.ent1 = parent;
  await getManager().save(child2);

  for (let i = 0; i < 35; i++) {
    getWithRelations(); // Will stall
  }

  for (let i = 0; i < 35; i++) {
    getSimple(); // Runs as expected
  }

  for (let i = 0; i < 35; i++) {
    getWithQb(); // Runs as expected
  }
})();

```

### LINKED ISSUE #9298: Loading entities with query strategy and connection limit set to 1 for mysql driver will cause dead lock
## Issue Description
when loading related entities using the `query` relationLoadStrategy it seems like after executing the sql the connection is not freed again which leads to non resolving queries because there are no more free connections in the mysql connection pool. 

### Expected Behavior
Connections are freed after executing the query to load related entites

### Actual Behavior
Connections are not freed leaving the pool with no free connections which means further relations cant be resolved because the queries will never receive a connection to execute them which will lead to infinite loading times. 

### Steps to Reproduce

Create mysql db with multiple related entities one entity that has 5 one to many relations. Use
```
relationLoadStrategy: 'query',
"extra": {
            "connectionLimit": 1
        }
```
in the typeorm config.

Execute a query and see that it never finishes due to lack of available connections. It will get stuck forever at
`const databaseConnection = await this.connect();` line 145 MysqlQueryRunner.hs

A working workaround would be to set the connectionLimit to a higher value like 50 or 100.
### My Environment

| Node.js version     | 16.15.0  | 
| Typescript version  | 4.5.4  | 
| TypeORM version     | 0.3.6  | 

We are also using NestJs /typeorm version 8.1.0

| `mysql`           | yes |

### LINKED ISSUE #10481: Hang-up with `relationLoadStrategy: query` for Multiple Calls
### Issue description

Using relationLoadStrategy set to "query" may result in a hang-up.

### Expected Behavior

When calling the find method with relationLoadStrategy set to "query" on tables with 1:N relationships, the process completes successfully.

### Actual Behavior

When multiple find method calls are made simultaneously with this setting, the system fails to complete the process and experiences a hang-up.

## Entities
We have tables with 1:N relationships.
```typescript:User.ts
import { Entity, PrimaryGeneratedColumn, Column, OneToMany } from "typeorm"
import { Photo } from "./Photo"

@Entity()
export class User {
    @PrimaryGeneratedColumn()
    id: number

    @Column()
    name: string

    @OneToMany(() => Photo, (photo) => photo.user, { eager: false })
    photos: Photo[]
}
```

```typescript:Photo.ts
import { Entity, PrimaryGeneratedColumn, Column, ManyToOne } from "typeorm"
import { User } from "./User"

@Entity()
export class Photo {
    @PrimaryGeneratedColumn()
    id: number

    @Column()
    url: string

    @ManyToOne(() => User, (user) => user.photos)
    user: User
}
```

## Bussiness Code
In our business code, we make simultaneous calls using the find method that includes relationships, setting the relationLoadStrategy to 'query'.

```typescript:main.ts
import { AppDataSource } from "../typeorm/data-source"
import { User } from "../typeorm/entity/User"

const main = async () => {
    await AppDataSource.initialize()
    const findUserWithPhoto = async () => {
        const repository = AppDataSource.getRepository(User)
        return await repository.findOne({
            where: { id: 1 },
            relations: {
                photos: true
            },
            relationLoadStrategy: "query"
        })
    }

    const result = []
    for (let i = 0; i < 10; i++) {
        result.push(findUserWithPhoto())
    }
    await Promise.all(result)
    await AppDataSource.destroy()
}

main()
```

For reproduction purposes, we are making multiple calls using a for loop. However, the same issue arises when server-side endpoints are rapidly and simultaneously accessed by clients

### My Environment

| Node.js version     | 18.y.zzz  |

## probably the cause

In TypeORM, the [obtainQueryRunner()](https://github.com/typeorm/typeorm/blob/0.3.16/src/query-builder/SelectQueryBuilder.ts#L3844) is used to acquire a connection. However, the issue here is that `this.queryRunner` is set with a value only in certain cases, which leads to the system consistently evaluating the right-hand expression of obtainQueryRunner.

In the connection pool, if no connections are available, [the system waits until a new connection is released](https://github.com/typeorm/typeorm/blob/0.3.16/src/driver/mysql/MysqlQueryRunner.ts#L191C64-L191C64). If `obtainQueryRunner` is called multiple times within the queryRunner, there is a potential for a deadlock to occur between the release of a connection and the acquisition of a new one. This deadlock can cause the system to become unresponsive or hang.

## reproduction code

https://github.com/saw81wt/typeorm-hung-up/tree/master

- [ ] aurora-mysql
- [ ] aurora-postgres
- [ ] better-sqlite3
- [ ] cockroachdb
- [ ] cordova
- [ ] expo
- [ ] mongodb
- [X] mysql
- [ ] nativescript
- [ ] oracle
- [ ] postgres
- [ ] react-native
- [ ] sap
- [ ] spanner
- [ ] sqlite
- [ ] sqlite-abstract
- [ ] sqljs
- [ ] sqlserver

No, I don’t have the time, but I can support (using donations) development.

casheeeewnuts:
#### Overview
Fixes #4738
Fixes #9298 
Fixes #10481  

Modified to use `queryRunner` when `QueryStrategyRelationIdLoader.loadManyToManyRelationIdsAndGroup` is called.

#### Detail
In the current implementation, if a query is executed with `relationLoadStrategy: query`, a new QueryRunner is created when resolving a relation.

The creation of a new QueryRunner means that a connection is allocated from the connection pool, but when the connection pool is exhausted, it waits for the connection to be released.

Because connections to query the parent table are released after the relation is resolved, if such queries are executed in parallel up to the maximum number of connection pools, a deadlock will occur because no connections will be allocated for queries on the `relation`.

#### Summary
So, this probrem can be solved by using the same QueryRunner is created when query parent table.

#### Side note
At first glance, it appears that the query runner is use in line 3587 of `src/query-builder/SelectQueryBuilder.ts`, but in fact, when resolving a relation inside "queryStrategyRelationIdLoader.loadManyToManyRelationIdsAndGroup", the QueryBuilder at line 3606 is used.

### Pull-Request Checklist

### DIFF
--- a/src/query-builder/QueryBuilder.ts
+++ b/src/query-builder/QueryBuilder.ts
@@ -521,8 +521,8 @@ export abstract class QueryBuilder<Entity extends ObjectLiteral> {
      * Creates a completely new query builder.
      * Uses same query runner as current QueryBuilder.
      */
-    createQueryBuilder(): this {
-        return new (this.constructor as any)(this.connection, this.queryRunner)
+    createQueryBuilder(queryRunner?: QueryRunner): this {
+        return new (this.constructor as any)(this.connection, queryRunner ?? this.queryRunner)
     }
 
     /**
--- a/src/query-builder/SelectQueryBuilder.ts
+++ b/src/query-builder/SelectQueryBuilder.ts
@@ -3603,7 +3603,7 @@ export class SelectQueryBuilder<Entity extends ObjectLiteral>
                           )
                         : this.findOptions.relations
 
-                    const queryBuilder = this.createQueryBuilder()
+                    const queryBuilder = this.createQueryBuilder(queryRunner)
                         .select(relationAlias)
                         .from(relationTarget, relationAlias)
                         .setFindOptions({
--- a/test/integration/sample2-one-to-one.ts
+++ b/test/integration/sample2-one-to-one.ts
@@ -198,6 +198,47 @@ describe("one-to-one", function () {
                 .getOne()
                 .should.eventually.eql(expectedDetails)
         })
+
+        it("should load post and its details with no hangup if query used", async function () {
+            if (!dataSource) return
+            const expectedPost = new Post()
+            expectedPost.id = savedPost.id
+            expectedPost.text = savedPost.text
+            expectedPost.title = savedPost.title
+            expectedPost.details = new PostDetails()
+            expectedPost.details!.id = savedPost.details!.id
+            expectedPost.details!.authorName = savedPost.details!.authorName
+            expectedPost.details!.comment = savedPost.details!.comment
+            expectedPost.details!.metadata = savedPost.details!.metadata
+
+            const findOne = () => postRepository.findOne({
+                    where: {
+                        id: savedPost.id
+                    },
+                    relations: {
+                        details: true
+                    },
+                    relationLoadStrategy: "query"
+                })
+
+            const posts = await Promise.all([
+                findOne(),
+                findOne(),
+                findOne(),
+                findOne(),
+                findOne(),
+                findOne(),
+                findOne(),
+                findOne(),
+                findOne(),
+                findOne(),
+            ])
+
+            posts.forEach(post => {
+                expect(post).not.to.be.null
+                post!.should.eql(expectedPost)
+            })
+        })
     })
 
     describe("insert post and category (one-side relation)", function () {
--- a/test/integration/sample3-many-to-one.ts
+++ b/test/integration/sample3-many-to-one.ts
@@ -200,6 +200,47 @@ describe("many-to-one", function () {
                 .getOne()
                 .should.eventually.eql(expectedDetails)
         })
+
+        it("should load post and its details with no hangup if query used", async function () {
+            if (!dataSource) return
+            const expectedPost = new Post()
+            expectedPost.id = savedPost.id
+            expectedPost.text = savedPost.text
+            expectedPost.title = savedPost.title
+            expectedPost.details = new PostDetails()
+            expectedPost.details!.id = savedPost.details!.id
+            expectedPost.details!.authorName = savedPost.details!.authorName
+            expectedPost.details!.comment = savedPost.details!.comment
+            expectedPost.details!.metadata = savedPost.details!.metadata
+
+            const findOne = () => postRepository.findOne({
+                where: {
+                    id: savedPost.id
+                },
+                relations: {
+                    details: true
+                },
+                relationLoadStrategy: "query"
+            })
+
+            const posts = await Promise.all([
+                findOne(),
+                findOne(),
+                findOne(),
+                findOne(),
+                findOne(),
+                findOne(),
+                findOne(),
+                findOne(),
+                findOne(),
+                findOne(),
+            ])
+
+            posts.forEach(post => {
+                expect(post).not.to.be.null
+                post!.should.eql(expectedPost)
+            })
+        })
     })
 
     describe("insert post and category (one-side relation)", function () {


jorenvandeweyer (approved):

pleerock:
Thank you for contribution! 🎉

scr4bble:
We haven't tested this yet but in case it really solves those mentioned issues - big thanks for the contribution from the users :)

edeesis:
Echo scr4bble's comment! This is huge, thank you!

PR #10659, opened 2024-01-26
bug: composite keys doesn't work as expected in 0.3 compared to 0.2

pleerock:
Draft test for the reported issue


pleerock:
closing if favor of #10672

PR #10672, opened 2024-01-31
chore(TypeORM): Create test case to uncover TypeORM composite key save issue

jeisberg:
This is a PR to uncover the underlying composite key save issue

Changes:
     - Added test to match use case uncovered in local codebase

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "typeorm",
-  "version": "0.3.19",
+  "version": "0.3.20",
   "lockfileVersion": 3,
   "requires": true,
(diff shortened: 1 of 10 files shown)


aklemen:
Thank you for fixing this! 🎉 

Don't want to be that person, but is there any ETA on when [the fix](https://github.com/typeorm/typeorm/commit/83567f533482d0170c35e2f99e627962b8f99a08) will be released?

ThomasShih:
I wonder if this fix helps solve https://github.com/typeorm/typeorm/issues/11132? I see its in 3.21 right now, I'll try to find some time to check.

PR #10684, opened 2024-02-06
refactor: use pragma method in better-sqlite3

mohd-akram:
### Description of change

Use the `pragma` method in better-sqlite3. This is the recommended way to run pragmas.

### Pull-Request Checklist

### DIFF
--- a/src/driver/better-sqlite3/BetterSqlite3Driver.ts
+++ b/src/driver/better-sqlite3/BetterSqlite3Driver.ts
@@ -138,7 +138,7 @@ export class BetterSqlite3Driver extends AbstractSqliteDriver {
             nativeBinding = null,
             prepareDatabase,
         } = this.options
-        const databaseConnection = this.sqlite(database, {
+        const databaseConnection = new this.sqlite(database, {
             readonly,
             fileMustExist,
             timeout,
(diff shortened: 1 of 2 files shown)


gioboa (commented):
Thank you @mohd-akram for your help.
Are any drawbacks about this change?

mohd-akram:
@gioboa No drawbacks that I'm aware of.

gioboa:
> @gioboa No drawbacks that I'm aware of.

Thanks 👍

alumni:
> Are any drawbacks about this change?

If you want to use bun's sqlite driver (similar to `better-sqlite3`), it won't have the `pragma` method :P

gioboa (commented):
So this is not working with `bun` :thinking: 
An idea is to check the runtime and create a condition in the code [eg. detect Bun](https://bun.sh/guides/util/detect-bun).
What do you think?

angelxmoreno:
what is the status of this PR? have the admins agreed or disagreed to review/merge this work?

gioboa (approved):
It looks good to me 
Thanks @mohd-akram

gioboa (commented):
Tests are fine.
What's the benefit of using `pragma` ? is it faster? reliable?

alumni:
hmm, `better-sqlite3` does not mention exactly what:

> It's better to use this method instead of normal [prepared statements](https://github.com/WiseLibs/better-sqlite3/blob/master/docs/api.md#preparestring---statement) when executing PRAGMA, because this method normalizes some odd behavior that may otherwise be experienced.

https://github.com/WiseLibs/better-sqlite3/blob/master/docs/api.md#pragmastring-options---results

PR #10705, opened 2024-02-14
fix: soft deletion should not update the already soft deleted rows

### LINKED ISSUE #10389: softDelete should not update alredy deleted rows
### Issue description

When doing a softDelete already deleted rows get udpated (inside a transaction, maybe always)

### Expected Behavior

softDelete should not operate on already softDeleted rows.

### Actual Behavior

Already deleted records that match the criteria get modifiedAt and deletedAt columns updated to current timestamps.

### Steps to reproduce

 `await trans.softDelete(Photo, {
          ownerId: owner.id,
        });

// This updates the same records again
await trans.softDelete(Photo, {
          ownerId: owner.id,
        });
`

### My Environment

| Operating System    |   Win 10       |
| Node.js version     | 14  |
| Typescript version  | 5.1.6 |

### Additional Context

Solution might be adding a `deleteAt is null` at the end of the SQL query,

- [ ] aurora-mysql
- [ ] aurora-postgres
- [ ] better-sqlite3
- [ ] cockroachdb
- [ ] cordova
- [ ] expo
- [ ] mongodb
- [ ] mysql
- [ ] nativescript
- [ ] oracle
- [ ] postgres
- [ ] react-native
- [ ] sap
- [ ] spanner
- [ ] sqlite
- [ ] sqlite-abstract
- [ ] sqljs
- [ ] sqlserver

hassanmehdi98:
#### Current Behavior
After a soft deletion, if you do a soft delete again with the same condition, it updates the `deleted at` column of the previously soft deleted rows as well.

#### New Behavior
Rows that were already soft deleted previously will be excluded from further soft deletion

### Pull-Request Checklist

- [x] This pull request links relevant issues as `Fixes #10389`


hassanmehdi98:
I don't understand why this test fails, any help? I can see this failing in several other PRs as well.
image

src/query-builder/SoftDeleteQueryBuilder.ts
```diff
@@ -198,6 +198,21 @@ export class SoftDeleteQueryBuilder<Entity extends ObjectLiteral>
                 { type: "simple", condition: condition },
             ]
         if (parameters) this.setParameters(parameters)
+
+        // only update rows that are not already soft deleted
+        this.expressionMap.wheres.push({
```
AmauryLiet, line 203: placing the fix in the "where" function doesn't adress 100% of the issue: one may not use the `where` function (i.e. soft deleting the whole table)

I would put the fix in `createUpdateExpression` instead

hassanmehdi98: This code adds the expression to the `where` list. even if the user doesn't pass any `where` clause, this would still be there when the query is executed because we hardcoded it inside the `where` list, or am I missing something?

hassanmehdi98: I think I get it, you mean if the user doesn't pass any `where` params (meaning this function never gets executed). I'll do the fix

hassanmehdi98: done 🙌

gioboa (commented):
Thank you @hassanmehdi98 for your help
I added this PR to our board for the next releases.

src/query-builder/SoftDeleteQueryBuilder.ts
```diff
@@ -491,6 +492,21 @@ export class SoftDeleteQueryBuilder<Entity extends ObjectLiteral>
             throw new UpdateValuesMissingError()
         }
 
+        // only update rows that are not already soft deleted
+        this.expressionMap.wheres.push({
+            type: "and",
```
alumni, line 496: I wonder what would happen if there's an OR in the condition, e.g.:

```ts
repo.softDelete([{ id: 1 }, { id: 2 }])
```

I'd also add this to the test to prevent future regressions.

gioboa: Added.

gioboa (approved):
Thanks @hassanmehdi98 for your help 👏

shoaibmarif:
@hassanmehdi98 highly appreciated ✨

PR #10727, opened 2024-02-21
Fix repeated positional arguments in OracleSQL being invalid queries. 

prosavage:
Positional paramters in oraclesql database don't seem to work properly.
So, if you have params that are being reused, it seems TypeORM will generate a query using the parameterIndexMap where it will reuse paramter names. So :1, :1, :2 in a SQL query doesn't work, even if the argument is exactly the same. It has to be :1, :2, :3 even if the query is reusing arguments. 

### Pull-Request Checklist


gioboa (commented):
Thanks @prosavage 
is it still valid? can you align it with the master branch please?
Is there an issue for that?

PR #10735, opened 2024-02-23
Virtual column context

### LINKED ISSUE #10734: Pass context into @VirtualColumn query
### Feature Description

It would be nice to be able to generate virtual column sql based on some contextual information about the query being run.  For example, if a query is possible to run against a time window, the time window could form part of that context.

### The Solution

Either the entire sql builder, or some context information, would need to be passed into the virtual column query function.  
It seemed to me to be safer to just pass a context object, rather than the whole query builder as that might open the framework up to other concerns such as infinite loops generating sql.

### Considered Alternatives

A custom query, or field, could be used instead of the @VirtualColumn decorator, but this small change would allow existing entity properties to be used.

### Additional Context

_No response_

- [ ] aurora-mysql
- [ ] aurora-postgres
- [ ] better-sqlite3
- [ ] cockroachdb
- [ ] cordova
- [ ] expo
- [ ] mongodb
- [ ] mysql
- [ ] nativescript
- [ ] oracle
- [ ] postgres
- [ ] react-native
- [ ] sap
- [ ] spanner
- [ ] sqlite
- [ ] sqlite-abstract
- [ ] sqljs
- [ ] sqlserver

strongpauly:
### Description of change
Pass a context object into @VirtualColumn query function.  Fixes #10734 

### Pull-Request Checklist

### DIFF
--- a/docs/decorator-reference.md
+++ b/docs/decorator-reference.md
@@ -383,24 +383,42 @@ Each time you call `find` or `findOne` from the entity manager, the value is rec
 ```typescript
 @Entity({ name: "companies", alias: "COMP" })
 export class Company extends BaseEntity {
-  @PrimaryColumn("varchar", { length: 50 })
-  name: string;
+    @PrimaryColumn("varchar", { length: 50 })
+    name: string
 
-  @VirtualColumn({ query: (alias) => `SELECT COUNT("name") FROM "employees" WHERE "companyName" = ${alias}.name` })
-  totalEmployeesCount: number;
+    @VirtualColumn({
+        query: (alias) =>
+            `SELECT COUNT("name") FROM "employees" WHERE "companyName" = ${alias}.name`,
+    })
+    totalEmployeesCount: number
 
-  @OneToMany((type) => Employee, (employee) => employee.company)
-  employees: Employee[];
+    @OneToMany((type) => Employee, (employee) => employee.company)
+    employees: Employee[]
 }
 
 @Entity({ name: "employees" })
 export class Employee extends BaseEntity {
-  @PrimaryColumn("varchar", { length: 50 })
-  name: string;
+    @PrimaryColumn("varchar", { length: 50 })
+    name: string
 
-  @ManyToOne((type) => Company, (company) => company.employees)
-  company: Company;
+    @ManyToOne((type) => Company, (company) => company.employees)
+    company: Company
 }
+```
+
+It is also possible to pass contextual information to the query function when using a manually created QueryBuilder. For example:
+
+```typescript
+const queryBuilder = entityManager.createQueryBuilder()
+queryBuilder.context = { minimumAge: 30 }
+queryBuilder.getMany()
+```
+
+This could then be used to modify the above query function to exclude employees under 30
+
+```typescript
+@VirtualColumn({ query: (alias, context) => `SELECT COUNT("name") FROM "employees" WHERE "companyName" = ${alias}.name AND age > ${context.minimumAge}` })
+totalEmployeesCount: number;
 ```
 
 ## Relation decorators
@@ -527,7 +545,7 @@ export class Post {
     @JoinColumn({
         name: "cat_id",
         referencedColumnName: "name",
-        foreignKeyConstraintName: "fk_cat_id"
+        foreignKeyConstraintName: "fk_cat_id",
     })
     category: Category
 }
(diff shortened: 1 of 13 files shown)


gioboa (changes requested):
Thanks @strongpauly for your help.
is it still valid? can you align this PR with the master branch?

strongpauly:
@gioboa I think its still valid, if you want to accept the feature 🤞

I've merged master into my branch.  There are some checks failing but I see that some other PRs also fail the same checks.

gioboa:
> I've merged master into my branch. There are some checks failing but I see that some other PRs also fail the same checks.

Thanks, I added this PR to our release board

docs/decorator-reference.md
```diff
@@ -383,26 +383,44 @@ Each time you call `find` or `findOne` from the entity manager, the value is rec
...
+```typescript
+const queryBuilder = entityManager.createQueryBuilder()
+queryBuilder.context = { minimumAge: 30 }
+queryBuilder.getMany()
+```
+
+This could then be used to modify the above query function to exclude employees under 30
+
+```typescript
+@VirtualColumn({ query: (alias, context) => `SELECT COUNT("name") FROM "employees" WHERE "companyName" = ${alias}.name AND age > ${context.minimumAge}` })
```
alumni, line 420: Please keep consistent quotations for the column names (e.g. `name` and `age` are not quoted but the other columns are).

test/github-issues/10734/entity/Company.ts
```diff
@@ -0,0 +1,23 @@
...
+import Employee from "./Employee"
+
+@Entity({ name: "companies" })
+export default class Company extends BaseEntity {
+    @PrimaryColumn("varchar", { length: 50 })
+    name: string
+
+    @VirtualColumn({
+        query: (alias, context) =>
+            `SELECT COUNT("name") FROM "employees" WHERE "companyName" = ${alias}.name AND age > ${context.minimumAge}`,
```
alumni, line 17: Actually consistent quotations are more important in the code, especially if we want to expand the test to other DB drivers.

alumni (commented):
I'm wondering if we should add more things here.

The main issue with writing plain SQL in the `VirtualColumn` definition is that we don't take advantage of the ORM. The SQL syntax can be different depending on the DB (e.g. MySQL uses non-standard backticks for quotation, so I had to make some workaround when I expanded the tests).

Something like the queryBuilder (that could be used to generate a sub-query) or the data source (like in view entities) could be helpful. This would allow to use driver-specific syntax even when using the `Repository` / `EntityManager` (and wouldn't require the user to pass it in the `QueryBuilder` context).

sgarner:
Sorry, but I'm not convinced that this is a good idea. Having `VirtualColumn` values in an entity behave differently by context seems like an inappropriate usage of the feature to me.

Properties of an entity should be data that defines and belongs to that entity or its relations. If you attach random bits of data that are only valid for some specific context then you're blurring the lines of what an entity is. You would have to take care not to pass that entity instance around your application and lose track of what context the `VirtualColumn` value was calculated for.

Personally, I'd be thinking about an approach like the following instead:

```typescript
const { entities, raw } = await dataSource
    .getRepository(Company)
    .createQueryBuilder("company")
    .addSelect(
        subQuery => subQuery.select("COUNT(*)", "totalEmployeesCount")
            .from(Employee, "employee")
            .where("employee.companyName = company.name")
            .andWhere("employee.age > 30"),
        "totalEmployeesCount",
    )
    .getRawAndEntities()

for (const company of entities) {
    const companyExtra = raw.find(r => r.id === company.id)
    console.log(`${company.name} has ${companyExtra?.totalEmployeesCount} employees over age 30`)
}
```

This keeps the entity separate from the contextually calculated data, while still fetching everything in one SELECT query.

strongpauly:
I understand where you’re coming from, but placing those columns elsewhere significantly complicates ongoing maintenance. The most logical—and maintainable—location for them is within the entity itself, so they remain directly accessible through getMany or getOne in the QueryBuilder.

Your proposed approach also sacrifices type safety. With the VirtualColumn pattern, developers get strong typing when working with these calculated fields; moving them out of the entity removes that safeguard.

The example in the tests and documentation may seem simple, but consider a real-world scenario where different users have varying access levels. Some calculated properties might need to be exposed only to, or have different values for, certain groups. Keeping these columns defined at the entity level makes it far easier to enforce such access rules while preserving both maintainability and type safety.

PR #9607, opened 2022-12-08
fix: allow `FindOptionsWhere` to accept `LessThan` with Union

tsugitta:
### Description of change

Fixes #9152 

Some CI have failed (edited: now CI passed by re-run), but it seems there is no problem based on the following points:

- I made changes to the type definitions but not to the implementation
- `tsc` succeeds
- The added tests do not depend on the database, and there is a successful CI for Postgres, etc.

### Pull-Request Checklist

- [x] Code is up-to-date with the `master` branch
- [x] `npm run format` to apply prettier formatting
- [x] `npm run test` passes with this change
- [x] This pull request links relevant issues as `Fixes #0000`
- [x] There are new or updated unit tests validating the change
- [ ] Documentation has been updated to reflect this change
  - updates doesn't seem necessary
- [x] The new commits follow conventions explained in [CONTRIBUTING.md](https://github.com/typeorm/typeorm/blob/master/CONTRIBUTING.md)

Issue #9152, opened 2022-06-27
Can't use `LessThan` for Union field


tsugitta:
## Issue Description

if a field is defined as Union like `1 | 2 | 3` or `'foo' | 'bar'`, operators `LessThan`, `MoreThan`, etc. raises a type error.

code example:

```ts
import { Entity, Column, getConnection, LessThan } from "typeorm";

type ValueType = 1 | 2 | 3;

@Entity()
export class User {
  @Column()
  foo: ValueType;
}

const repository = getConnection().getRepository(User);

const value = 3 as ValueType;

repository.findBy({
  foo: LessThan(value)
});
```

### Expected Behavior

works as if the field is defined as `number`

### Actual Behavior

```
Type 'FindOperator<1 | 2 | 3>' is not assignable to type '1 | 2 | 3 | FindOperator<1> | FindOperator<2> | FindOperator<3> | undefined'.
  Type 'FindOperator<1 | 2 | 3>' is not assignable to type 'FindOperator<1>'.
    Type '1 | 2 | 3' is not assignable to type '1'.
      Type '2' is not assignable to type '1'.ts(2322)
```

### Steps to Reproduce

[CodeSandbox](https://codesandbox.io/s/lingering-hill-36rb2u?file=/src/index.ts)

### My Environment

| Dependency          | Version  |
| ---                 | ---      |
| Operating System    |          |
| Node.js version     |   | 
| Typescript version  | 4.4.4  | 
| TypeORM version     | 0.3.6  | 

### Relevant Database Driver(s)

| DB Type           | Reproducible  |
|-------------------| ---           |
| `aurora-mysql`    | no            |
| `aurora-postgres` | no            |
| `better-sqlite3`  | no            |
| `cockroachdb`     | no            |
| `cordova`         | no            |
| `expo`            | no            |
| `mongodb`         | no            |
| `mysql`           | no            |
| `nativescript`    | no            |
| `oracle`          | no            |
| `postgres`        | no            |
| `react-native`    | no            |
| `sap`             | no            |
| `spanner`         | no            |
| `sqlite`          | no            |
| `sqlite-abstract` | no            |
| `sqljs`           | no            |
| `sqlserver`       | no            |

### Are you willing to resolve this issue by submitting a Pull Request?

 - ✅ Yes, I have the time, and I know how to start.
 - ✖️ Yes, I have the time, but I don't know how to start. I would need guidance.
 - ✖️ No, I don’t have the time, but I can support (using donations) development.
[...]

PR #10742, opened 2024-02-27
fix: allow `FindOptionsWhere` to accept `LessThan` with `Union`

hfhchan-plb:
### Description of change

A better fix for #9152 (replaces #9607)

The existing fix for #9152 in #9607 relies on a somewhat implicit behaviour of TypeScript where having multiple generic parameters causes the generic parameters to _not_ be distributive (default is that union types become distributive).

Per [TypeScript docs](https://www.typescriptlang.org/docs/handbook/2/conditional-types.html#distributive-conditional-types), the suggested way to do it is to enclose both sides of the extends in square brackets:

> Typically, distributivity is the desired behavior. To avoid that behavior, you can surround each side of the extends keyword with square brackets.
> ```
> type ToArrayNonDist<Type> = [Type] extends [any] ? Type[] : never;
> ```

This PR applies the suggested way from the TypeScript handbook

### Pull-Request Checklist

- [#] Code is up-to-date with the `master` branch
- [#] This pull request links relevant issues as `Fixes #0000`
- [ ] There are new or updated unit tests validating the change -- N/A (old ones do)
- [ ] Documentation has been updated to reflect this change -- N/A
- [#] The new commits follow conventions explained in [CONTRIBUTING.md](https://github.com/typeorm/typeorm/blob/master/CONTRIBUTING.md)

### DIFF
--- a/src/find-options/FindOptionsWhere.ts
+++ b/src/find-options/FindOptionsWhere.ts
@@ -4,33 +4,28 @@ import { EqualOperator } from "./EqualOperator"
 
 /**
  * A single property handler for FindOptionsWhere.
- *
- * The reason why we have both "PropertyToBeNarrowed" and "Property" is that Union is narrowed down when extends is used.
- * It means the result of FindOptionsWhereProperty<1 | 2> doesn't include FindOperator<1 | 2> but FindOperator<1> | FindOperator<2>.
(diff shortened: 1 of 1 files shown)


gioboa (commented):
@hfhchan-plb thanks for opening this PR.
I'm closing this for now because it's stale and with conflicts, feel free to create a new PR with your improvements. :muscle:

PR #10796, opened 2024-03-26
Extract function to generate migration

lan17:
### Description of change

Typeorm relies on a static DataSource being created and exported.  However, this does not play nice with code bases where this is not desirable.  

To support generating migrations in such code bases, I extracted the function to generate migration in such a way that you can just pass in instance of DataSource to it along with other options.  In this manner, we should be able to use `MigrationGenerateCommand.genMigration(...)` without having to go through Typeorm CLI that assumes static DataSource.

### Pull-Request Checklist

### DIFF
--- a/src/commands/MigrationGenerateCommand.ts
+++ b/src/commands/MigrationGenerateCommand.ts
@@ -71,20 +71,41 @@ export class MigrationGenerateCommand implements yargs.CommandModule {
         const fullPath = args.path.startsWith("/")
             ? args.path
             : path.resolve(process.cwd(), args.path)
-        const filename = timestamp + "-" + path.basename(fullPath) + extension
 
-        let dataSource: DataSource | undefined = undefined
+        const dataSource = await CommandUtils.loadDataSource(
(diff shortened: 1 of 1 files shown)


lan17:
Updating tests should not be required since this is a simple refactoring change

src/commands/MigrationGenerateCommand.ts
```diff
@@ -71,20 +71,42 @@ export class MigrationGenerateCommand implements yargs.CommandModule {
...
+            args.check,
+            args.pretty,
+            args.exitProcess !== false,
+            timestamp,
+            extension,
+            fullPath,
+        )
+    }
+
+    public static async genMigration(
```
alumni, line 97: I would go a bit further and extract a method that only returns the up + down SQL statements instead of JS/TS file, and handle the JS/TS code generation separately.

I often have to search and replace certain things in the migration, e.g. schema name / entity prefix + handle escaping. Also, my migration template is slightly different, e.g. I'd like to add SQL template strings in the code so that developers spot the issues in the generated migration easier and manually fix them.

PR #10829, opened 2024-04-14
feat: support nullable embedded entities

test137E29B:
Support nullable embedded fields such that embedded documents in MongoDB can correctly be returned as null for a subdocument that is explicitly nullable.

Closes: #3913

### Description of change

This change adds to the `ColumnEmbeddedOptions` the `nullable` field, thus allowing subdocuments in MongoDB to be `null` explicitly. This ensures they're returned correctly from the database in the case a null is set. This is useful when a developer has a required key that can be set to null - currently the key is not returned at all.

I saw this in #3913 though this was reverted as it previously changed all nested fields to null, which caused some problems I imagine for people. This method of implementing this change should not cause problems or any breaking changes and is entirely optional.

I have locally tested this with MongoDB, but not a SQL provider.

### Pull-Request Checklist

### DIFF
--- a/docs/embedded-entities.md
+++ b/docs/embedded-entities.md
@@ -165,3 +165,40 @@ All columns defined in the `Name` entity will be merged into `user`, `employee` 
 This way code duplication in the entity classes is reduced.
 You can use as many columns (or relations) in embedded classes as you need.
 You even can have nested embedded columns inside embedded classes.
+
+## Nullable embedded entities
+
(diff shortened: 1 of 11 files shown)


test137E29B:
Would love if someone could add support for this in SQL providers, since my changes don't seem to have worked. I can't imagine it's too hard.

Alternatively, I can just add a note that this will only work for MongoDB.

Open to opinions on this 😃

gioboa (commented):
Thanks @test137E29B for your help.
is it still valid? the issue mentioned in the description is closed

PR #10846, opened 2024-04-23
feat: Add query timeout support for MySql

iliagrvch:
### Description of change

This feature adds the "enableQueryTimeout" option to MysqlConnectionOptions. When enabled, the value of "maxQueryExecutionTime" will be passed to the MySQL driver as the query timeout. (There is support for it by the MySQL driver: https://github.com/mysqljs/mysql?tab=readme-ov-file#timeouts)

Now, passing a value to "maxQueryExecutionTime" just logs queries that took more than the specified amount of time.

We use TypeORM with MySQL in multiple projects. Some of these projects involve direct interaction with users, and having a query timeout for the entire project connection is crucial to prevent server blocking due to long queries. On the other hand, some projects run in the background, and we don't mind queries taking a bit more time.

Since multiple drivers support query timeout configuration per datasource/connection (e.g., 'postgres'/'sqlserver'), and MySQL doesn't, I have decided to add this feature for MySQL as well.

### Pull-Request Checklist


Elimorsson:
this will help a lot! I'm waiting for the merge of that PR

michaelr231:
awesome! exactly what we need

bao-nguyen-khac:
@iliagrvch failed in oracle test ??, pls check and create new merge req, I really need this option

iliagrvch:
@bao-nguyen-khac, I tried to take a look at it, but it looks like it just fails because of a timeout, the feature is completely unrelated to Oracle and only mysql files were changed. There is multiple PRs with the same issue.

Is there any way we can re-run the Oracle tests @pleerock, @AlexMesser?

bassedOutt:
This is a very useful feature. Please consider merging. Thanks

YuliaTs:
This is a much-needed improvement!  Wait for merge

gioboa (commented):
Thanks @iliagrvch for your help, it looks straightforward
I added this PR to our release board

mguida22 (commented):
Thanks for your contribution @iliagrvch! Can you add some tests for this behavior?

iliagrvch:
@mguida22 Sure, just added.

test/functional/driver/mysql/connection-options/enable-query-timeout.ts
```diff
@@ -0,0 +1,106 @@
...
+    let dataSources: DataSource[]
+    const commonConnectionOptions: TestingOptions = {
+        entities: [__dirname + "/entity/*{.js,.ts}"],
+        schemaCreate: true,
+        dropSchema: true,
+        enabledDrivers: ["mysql"],
+    }
+    const timeoutMs = 4000
+    const longQueryTime = timeoutMs / 1000 + 1
+    const shortQueryTime = timeoutMs / 2000
```
mguida22, line 21: Can we make these times very short (a few milliseconds) so the test suite runs faster?

mguida22 (approved):
This looks good, thank you @iliagrvch!

test/functional/driver/mysql/connection-options/enable-query-timeout.ts
iliagrvch: Yeah, good point. Done ✅

naorpeled (approved):
LGTM!

gioboa (approved):

bao-nguyen-khac:
@iliagrvch @mguida22 @gioboa please check this again, could you add skip check sqlite :D

dberardo-com:
is this option available for postgres already? and can that be set on a per-query basis ?

i really need this feature to avoid deadlock queries on complex transactions and the issue i have is that some "read" queries that i use are currently long running queries by design, so i cannot impose a universal query timeout, but need to be more specific on single queries.

mguida22:
> is this option available for postgres already? and can that be set on a per-query basis ?
> 
> i really need this feature to avoid deadlock queries on complex transactions and the issue i have is that some "read" queries that i use are currently long running queries by design, so i cannot impose a universal query timeout, but need to be more specific on single queries.

This feature sets a database timeout on all queries. You can achieve the same thing with postgres via 

```ts
  type: "postgres",
  
  // ...
  
  extra: {
    // this is passed to pg
    // https://node-postgres.com/apis/client#new-client
    statement_timeout: YOUR_TIMEOUT_IN_MS,
  },
});
```

I don't believe we support per query timeouts via TypeORM today.

PR #10868, opened 2024-05-01
fix: resolve issue on index naming strategy

### LINKED ISSUE #10856: TypeORM Migration Bug: Duplicate Index Names in Generated Migration
### Issue description

As you can see, the generated migration script has identical index names, even though the order of columns in the entity definition is different for the two indices. This is causing a conflict and potentially invalid database operations due to duplicate index names.

### Expected Behavior

The migration script should generate unique index names based on the column order specified in the entity definition, ensuring proper creation and deletion of indices in the database

### Actual Behavior

When running yarn migration:generate in a TypeORM project, the generated migration script creates duplicate index names despite the entity file specifying two distinct indices with different column orders.

## Entity Definition
Here is the relevant part of the entity definition that specifies two different indices:

```
import { Column, Entity, Index, PrimaryColumn } from "typeorm";

@Entity()
@Index(["aCrc", "bCrc", "c"])
@Index(["bCrc", "aCrc", "c"])
export class TableA {
  @PrimaryColumn({ type: "binary", length: 20 })
  a: Buffer;

  @PrimaryColumn({ type: "binary", length: 20 })
  b: Buffer;

  @Column({ type: "int", unsigned: true })
  bCrc: number;

  @Column({ type: "int", unsigned: true })
  aCrc: number;

  @Column({ type: "double", unsigned: true })
  c: number;
}

```

## Generated Migration Script
However, the generated migration script shows two identical index names, despite the indices being defined with different orders:
```
import { MigrationInterface, QueryRunner } from "typeorm";

export class CreateTables1714372786425 implements MigrationInterface {
  name = "CreateTables1714372786425";

    await queryRunner.query(
      `CREATE TABLE \`table_a\` (
        \`a\` binary(20) NOT NULL, 
        \`b\` binary(20) NOT NULL, 
        \`bCrc\` int UNSIGNED NOT NULL, 
        \`aCrc\` int UNSIGNED NOT NULL, 
        \`c\` double UNSIGNED NOT NULL, 
        INDEX \`IDX_636863777022eb2cd64c5407f2\` (\`bCrc\`, \`aCrc\`, \`c\`), 
        INDEX \`IDX_636863777022eb2cd64c5407f2\` (\`aCrc\`, \`bCrc\`, \`c\`), 
        PRIMARY KEY (\`a\`, \`b\`)) 
      ENGINE=InnoDB`
    );
  }

    await queryRunner.query(`DROP INDEX \`IDX_636863777022eb2cd64c5407f2\` ON \`table_a\``);
    await queryRunner.query(`DROP INDEX \`IDX_636863777022eb2cd64c5407f2\` ON \`table_a\``);
    await queryRunner.query(`DROP TABLE \`table_a\``);
  }
}
```

### Steps to Reproduce
To reproduce the issue of duplicate index names in TypeORM migrations, follow these steps:

1. **Create a New TypeORM Project**
   - Initialize a new TypeORM project with the necessary setup for migrations.
   - Ensure you have a valid database connection configured in `ormconfig.json` or a similar configuration file.

2. **Create an Entity with Two Different Indices**
   - Create an entity that has at least two index definitions with different column orders. Here is an example entity:

     ```typescript
     import { Column, Entity, Index, PrimaryColumn } from "typeorm";

     @Entity()
     @Index(["aCrc", "bCrc", "c"]) // First index
     @Index(["bCrc", "aCrc", "c"]) // Second index
     export class TableA {
       @PrimaryColumn({ type: "binary", length: 20 })
       a: Buffer;

       @PrimaryColumn({ type: "binary", length: 20 })
       b: Buffer;

       @Column({ type: "int", unsigned: true })
       bCrc: number;

       @Column({ type: "int", unsigned: true })
       aCrc: number;

       @Column({ type: "double", unsigned: true })
       c: number;
     }
     ```

3. **Generate a Migration Script**
   - Run the migration generation command, typically `yarn migration:generate --name=CreateTableA`.
   - This will generate a migration script that creates the table with the defined indices.

4. **Inspect the Generated Migration Script**
   - Open the generated migration script (e.g., `src/migration/1633020286482-CreateTableA.ts`).
   - Check the index definitions in the migration script to see if they have the same name despite being defined with different column orders in the entity.

If the above steps are followed, you should see that the generated migration script has duplicate index names, indicating the issue with the migration generation process in TypeORM. This behavior might lead to database conflicts or unexpected results when running the migrations.

### My Environment

| Operating System    |   MAC       |
| Node.js version     | 20.12.0  |
| Typescript version  | 5.1.3  |
| nestjs/typeorm  | 10.0.0 |

### Additional Context

_No response_

- [ ] aurora-mysql
- [ ] aurora-postgres
- [ ] better-sqlite3
- [ ] cockroachdb
- [ ] cordova
- [ ] expo
- [ ] mongodb
- [X] mysql
- [ ] nativescript
- [ ] oracle
- [ ] postgres
- [ ] react-native
- [ ] sap
- [ ] spanner
- [ ] sqlite
- [ ] sqlite-abstract
- [ ] sqljs
- [ ] sqlserver

alenap93:
Fixes: #10856

### Description of change
This pr fix the issue that default index naming strategy of indexes on the same columns but with different order, generate the same names. 

### Pull-Request Checklist


gioboa (commented):
Thanks for your help @alenap93 
I added this PR to the release board

simplenotezy:
@gioboa can we have this merged in? 😊

src/naming-strategy/DefaultNamingStrategy.ts
```diff
@@ -122,12 +122,9 @@ export class DefaultNamingStrategy implements NamingStrategyInterface {
         columnNames: string[],
         where?: string,
     ): string {
-        // sort incoming column names to avoid issue when ["id", "name"] and ["name", "id"] arrays
-        const clonedColumnNames = [...columnNames]
-        clonedColumnNames.sort()
```
alumni, line 127: Sorting is intentional here.

alenap93: order of columns in a composite index matters

sgarner: I think @alenap93 is right. This code appears to have been mistakenly copied from the similar methods for primary key, constraint and foreign key naming above it. In those cases you can't meaningfully create two keys of the same kind with the same columns in a different order, but for an index you can.

test/github-issues/10856/issue-10856.test.ts
```diff
@@ -0,0 +1,36 @@
+import "reflect-metadata"
+import {
+    createTestingConnections,
+    closeTestingConnections,
+} from "../../utils/test-utils"
+import { DataSource } from "../../../src/index.js"
+import { expect } from "chai"
+
+describe("github issues > #10856 TypeORM Migration Bug: Duplicate Index Names in Generated Migration", () => {
```
sgarner, line 9: We are moving away from creating tests for specific GitHub issues. Instead, we should add tests that show all `DefaultNamingStrategy` functionality works as expected.

Unfortunately there is not a test suite for `DefaultNamingStrategy` yet. It could be added to the path `test/functional/naming-strategy/default-naming-strategy`.

sgarner (changes requested):

test/github-issues/10856/issue-10856.test.ts
alenap93: @sgarner sorry, this is a very old PR. This weekend i will try to create the new test.

src/naming-strategy/DefaultNamingStrategy.ts
alumni: It might depend on the index though. Uniques are a type of index. So are primary keys.

Later edit: order might matter in all indices, including primary keys, but I have a feeling TypeORM chooses to keep them in alphabetical order to improve the DX and prevent duplicate indices (which in some drivers are not allowed).

sgarner: But there is a distinction between unique constraints and unique indexes. Likewise primary keys are defined in a special way, even if databases implicitly index them.

Some databases (e.g. Postgres, SQL Server) require both a unique constraint and an index to be created. Others (e.g. MySQL) don't support unique constraints at all, so just a unique index is created.

The `uniqueConstraintName` method is only used for naming a unique _constraint_, and the order of columns in a unique constraint isn't significant, so it's okay to sort the column names in that context.

If you define a unique index in TypeORM, the `indexName` method is used for naming the _index_. Because this index generally has all the functionality of a regular index as well as enforcing uniqueness, column ordering can be significant in this context.

I'm not familiar with HANA, but in most RDBMS the order of columns in a composite index is critical for determining the performance characteristics of that index based on how the data is queried.

gioboa (commented):
is this PR still valid?

alumni:
> is this PR still valid?

In its current state, this PR introduces a breaking change, since the index names will no longer be the same, and migrations will be generated for every index.

There was a similar change sometime after 0.2, and we had to manually create a huge migration to rename all indices (and we have a few hundreds) - this will cause the same issue.

I'm thinking we should rather add a new naming strategy (we can even create a `@typeorm/naming-strategies` package later) and keep the existing one unchanged. If the behavior in this PR is correct for most cases, we can later make it the default one in v1 or in some future version.

I'd also check what each DB allows - it might even be dependent on the table storage type, e.g. I think in column storage the column order does not matter (only in row storage) and the DB blocks you from creating multiple indices with the same columns but in different order (I've even seen situations where you can't create a composite index that is a subset of another one). Then, the current behavior is probably the one we want to have for such situation.

michaelbromley:

```
git fetch origin
```

PR #10891, opened 2024-05-15
feat: add SaferMigrations utility for safer column type changes

Rafael-rgsousa:
### Description of change

Add SaferMigrations utility to backup and restore data during column type changes. This utility ensures data integrity by backing up specified columns before the migration and restoring them afterward. The utility supports multiple tables and primary key configurations.

Also added tests to validate the backup and restore functionality, including:
- Basic migration where the column type is modified.
- Complex migration where the column is dropped and a new column with a different type is added.

Closes: #3357

### Pull-Request Checklist

### DIFF
--- a/docs/migrations.md
+++ b/docs/migrations.md
@@ -1,3 +1,5 @@
+---
+
 # Migrations
 
 - [Migrations](#migrations)
@@ -10,6 +12,7 @@
   - [DataSource option](#datasource-option)
   - [Timestamp option](#timestamp-option)
   - [Using migration API to write migrations](#using-migration-api-to-write-migrations)
+  - [Using SaferMigrations for Data Integrity](#using-safermigrations-for-data-integrity)
 
 ## How migrations work
 
@@ -17,7 +20,7 @@ Once you get into production you'll need to synchronize model changes into the d
 Typically, it is unsafe to use `synchronize: true` for schema synchronization on production once
 you get data in your database. Here is where migrations come to help.
 
-A migration is just a single file with sql queries to update a database schema
+A migration is just a single file with SQL queries to update a database schema
 and apply new changes to an existing database.
 
 Let's say you already have a database and a post entity:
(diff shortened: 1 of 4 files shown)


opiredev:
The user @Rafael-rgsousa has claimed all the rewards in #3357 👀 We've notified the reward creators [here](https://github.com/typeorm/typeorm/issues/3357#issuecomment-2113598175).
@Rafael-rgsousa make sure you have your [payment account ready](https://app.opire.dev/settings) to receive the payment for your hard work 💪

JosefBud:
I may be misunderstanding something but this seems like a **workaround** for the bug, not a fix. This would not close the issue.

Rafael-rgsousa:
> I may be misunderstanding something but this seems like a **workaround** for the bug, not a fix. This would not close the issue.

pls cooperate with a better idea

JosefBud:
> pls cooperate with a better idea

The issue states that TypeORM's migration generator is generating migrations that drop columns instead of using `ALTER COLUMN`. The better idea would be to fix their migration generator at the source to generate migrations that use `ALTER COLUMN` when possible, instead of `DROP COLUMN` and `ADD COLUMN`.

Your workaround is great, and it's useful, all I'm saying is that I don't think it will be merged in because it's not actually fixing the root issue. You should post this as a workaround in the issue that people can use until this gets fixed at the source.

[Here is the condition that drops the column if the column length changes.](https://github.com/typeorm/typeorm/blob/e7649d2746f907ff36b1efb600402dedd5f5a499/src/driver/postgres/PostgresQueryRunner.ts#L1230) I believe that's what needs to change.

Rafael-rgsousa:
> > pls cooperate with a better idea
> 
> The issue states that TypeORM's migration generator is generating migrations that drop columns instead of using `ALTER COLUMN`. The better idea would be to fix their migration generator at the source to generate migrations that use `ALTER COLUMN` when possible, instead of `DROP COLUMN` and `ADD COLUMN`.
> 
> Your workaround is great, and it's useful, all I'm saying is that I don't think it will be merged in because it's not actually fixing the root issue. You should post this as a workaround in the issue that people can use until this gets fixed at the source.
> 
> [Here is the condition that drops the column if the column length changes, for Postgres.](https://github.com/typeorm/typeorm/blob/e7649d2746f907ff36b1efb600402dedd5f5a499/src/driver/postgres/PostgresQueryRunner.ts#L1230) I believe that's what needs to change, and then that change needs to be propagated to the other database drivers.

Yeah, right we could change the DROP COLUMN to ALTER COLUMN but it can add a secondary situation: data conversion like commented in the code:              // To avoid data conversion, we just recreate column

This is dangerous because some database systems auto-handle conversion. MySQL, PostgreSQL, and SQLite can easily convert types automatically because they are flexible. Still, systems like OracleDB, SQL Server, and IBM DB2 work with strict type enforcement, so changing to "alter column" can fail in some systems. That's why the backup approach is a good idea as it will re-insert items as "new" and the problem will be avoided.

Still, it may the better to move the backup approach from the "SaferMigration" directly to inside the core of the migration and make it automatically

docs/migrations.md
```diff
@@ -401,536 +403,59 @@ export class QuestionRefactoringTIMESTAMP implements MigrationInterface {
 }
 ```
 
----
```
nikelborm, line 404: Hi! I see you added new functionality for migrations, but you haven't deleted the old functionality, so as far as I understand, old documentation is still relevant. Could you please explain your motivation behind removing this entire chunk of documentation?

RubenRuCh:
Hi @Rafael-rgsousa any news on this?

gioboa (commented):
Thanks @Rafael-rgsousa  for your help
is this PR still valid? it looks stale

gioboa (commented):
Thanks @Rafael-rgsousa for your updates. It seems reasonable to me

PR #10896, opened 2024-05-19
fix: resolve property metadata override with extended entities

### LINKED ISSUE #10895: Property metadata override does not work when extending an entity class
### Issue description

Property metadata override does not work when extending an entity class

### Expected Behavior

```
@Entity()
class Extendable {
  @PrimaryGeneratedColumn()
  id: number;
  
  @Column({ type: 'int' })
  value: number;
}

@Entity()
class Extending extends Extendable {
  @Column({ type: 'decimal' })
  value: number;

  @Column({ type: 'int' })
  otherValue: number;
}

  // any other configuration
  synchronize: true,
  entities: [Extending],
});
dataSource.initialize();
```

In this example I expect `Extending.value` field to have metadata defined in `Extending`, not in `Extendable`, so in the database `value` field should have type `decimal`, not `int`.

### Actual Behavior

TypeORM applies metadata from parent class field for descendant class field even if descendant overrides that field metadata.
So, in example from "Expected Behaviour", in the database the type of the field `value` would be `int`. 

### Steps to reproduce

Create 2 entity classes, one of them should extend another and override parent field column metadata.

### My Environment

| ---                 | ---   |
| Operating System  |  |
| Node.js version      | 20.13.1 |
| Typescript version  | 5.4.5 |
| TypeORM version   | 0.3.20 |

### Additional Context

While consuming and processing metadata, `MetadataArgsStorage.filterByTargetAndWithoutDuplicateProperties` method is used, which save first occurance of target metadata and ignores any other. But, I think, that this function should consider overrides of metadata and set higher priority for metadata from descendants.

- [ ] aurora-mysql
- [ ] aurora-postgres
- [ ] better-sqlite3
- [ ] cockroachdb
- [ ] cordova
- [ ] expo
- [ ] mongodb
- [ ] mysql
- [ ] nativescript
- [ ] oracle
- [ ] postgres
- [ ] react-native
- [ ] sap
- [ ] spanner
- [ ] sqlite
- [ ] sqlite-abstract
- [ ] sqljs
- [ ] sqlserver

titovb:
### Description of change
Fixes #10895

### Pull-Request Checklist


src/metadata-args/MetadataArgsStorage.ts
```diff
@@ -339,21 +339,28 @@ export class MetadataArgsStorage {
...
-                : item.target === target
-            if (sameTarget) {
-                if (
-                    !newArray.find(
-                        (newItem) => newItem.propertyName === item.propertyName,
-                    )
-                )
-                    newArray.push(item)
+        const uniqueArray: T[] = []
+        const targets = Array.isArray(target) ? [...target].reverse() : [target]
```
titovb, line 343: The targets are added in the order ChildChild < Child < Base. Therefore, we need to reverse it to apply the parent columns first and override them if they were overridden by the child.

docs/separating-entity-definition.md
```diff
@@ -307,7 +307,6 @@ const ASchema = new EntitySchema<A>({
     // specified on the 'discriminatorValue' property
     discriminatorValue: "my-custom-discriminator-value-for-A",
     columns: {
-        ...BaseSchema.options.columns,
```
titovb, line 310: @pleerock @AlexMesser Could you verify these changes to the documentation and tests?
Correct me if I'm wrong, but the addition of BaseSchema columns is not needed because STI inheritance ensures the addition of BaseSchema columns using metadata from BaseSchema, so we don't need to override them.

tophep (approved):

gioboa (commented):
is this PR still valid? can I ask you to solve the conflicts please?

titovb:
> is this PR still valid? can I ask you to solve the conflicts please?

as I know the issue is still valid (https://github.com/typeorm/typeorm/issues/10895)
I resolved conflicts, it would be nice if someone will review this PR.

gioboa (commented):
Can you address the qodo feedback please?

gioboa (changes requested):

michaelbromley:

```
git fetch origin
```

PR #10910, opened 2024-05-29
fix: update/delete/softDelete by criteria of condition objects

### LINKED ISSUE #10517: EntityManager update/delete/softDelete don't work with list of where condition objects
### Issue description

EntityManager update/delete/softDelete don't work with list of where condition objects

### Expected Behavior

A list of where condition objects should be allowed for update/delete/softDelete/restore. The following should all work:

```js
const whereConditions = [
        { id: LessThan(5) },
        { id: MoreThan(10) }
]
// These are working fine:
await dataSource.manager.find(target, { where: whereConditions })
await dataSource.manager.findBy(target, whereConditions)
// These all fail:
await dataSource.manager.update(target, whereConditions, { someColumn: 'foo' })
await dataSource.manager.softDelete(target, whereConditions)
await dataSource.manager.delete(target, whereConditions)
await dataSource.manager.restore(target, whereConditions)
```

### Actual Behavior

For update/delete/softDelete/restore, EntityManager interprets the list of condition objects as list of primitive values and therefore always creates an SQL statement like this:
```sql
... WHERE "id" IN ($1, $2)
```
with parameters
```json
[
  {"_type":"lessThan","_value":5,"_useParameter":true,"_multipleParameters":false}, 
  {"_type":"moreThan","_value":10,"_useParameter":true,"_multipleParameters":false}
]
```

This behavior makes sense when a list of numbers / strings / Dates was given, but not for complex objects where providing an Array of objects should be interpreted as an `OR` of multiple conditions.

The responsible code section exists in the 4 methods EntityManager.update/delete/softDelete/restore. Here the example for update:
```js
if (
    typeof criteria === "string" ||
    typeof criteria === "number" ||
    criteria instanceof Date ||
    Array.isArray(criteria)
) {
    return this.createQueryBuilder()
        .update(target)
        .set(partialEntity)
        .whereInIds(criteria)
        .execute()
} else {
    return this.createQueryBuilder()
        .update(target)
        .set(partialEntity)
        .where(criteria)
        .execute()
}
```

I would propose changing the condition in the `if` such that it only applies for Arrays that contain only strings, numbers and Dates:
```ts
// added this method
private isCriteriaOnlyPrimitiveValues(criteria: any) {
    if(
        typeof criteria === "string" ||
        typeof criteria === "number" ||
        criteria instanceof Date
    ) return true

    if(Array.isArray(criteria)) {
        if(!criteria.length) return true // treat empty array as array of primitive values for backwards compatibility
        return criteria.every(item => !Array.isArray(item) && this.isCriteriaOnlyPrimitiveValues(item))
    }
}

update(target, criteria, partialEntity) {
    // if user passed empty criteria or empty list of criterias, then throw an error
    if (criteria === undefined ||
        criteria === null ||
        criteria === "" ||
        (Array.isArray(criteria) && criteria.length === 0)) {
        return Promise.reject(new error_1.TypeORMError(`Empty criteria(s) are not allowed for the update method.`));
    }
    // changed this condition
    if (this.isCriteriaOnlyPrimitiveValues(criteria)) {
        [..]
    }
    else {
        [..]
    }
}
```

### Steps to reproduce

See example above. The bug is independent of the DB type used (I used Postgres).

### My Environment

| Operating System    | Linux (shouldn't matter) |
| Node.js version     | 18.17.0 (shouldn't matter)|
| Typescript version  | 5.3.2 (shouldn't matter) |

### Additional Context

_No response_

- [ ] aurora-mysql
- [ ] aurora-postgres
- [ ] better-sqlite3
- [ ] cockroachdb
- [ ] cordova
- [ ] expo
- [ ] mongodb
- [ ] mysql
- [ ] nativescript
- [ ] oracle
- [ ] postgres
- [ ] react-native
- [ ] sap
- [ ] spanner
- [ ] sqlite
- [ ] sqlite-abstract
- [ ] sqljs
- [ ] sqlserver

maxbronnikov10:
### Description of change

Fixes #10517

### Pull-Request Checklist

### DIFF
--- a/src/entity-manager/EntityManager.ts
+++ b/src/entity-manager/EntityManager.ts
@@ -38,6 +38,7 @@ import { UpsertOptions } from "../repository/UpsertOptions"
 import { InstanceChecker } from "../util/InstanceChecker"
 import { ObjectLiteral } from "../common/ObjectLiteral"
 import { PickKeysByType } from "../common/PickKeysByType"
+import { OrmUtils } from "../util/OrmUtils"
 
 /**
  * Entity manager supposed to work with any entity, automatically find its repository and call its methods,
@@ -761,12 +762,7 @@ export class EntityManager {
         partialEntity: QueryDeepPartialEntity<Entity>,
     ): Promise<UpdateResult> {
         // if user passed empty criteria or empty list of criterias, then throw an error
-        if (
-            criteria === undefined ||
-            criteria === null ||
-            criteria === "" ||
-            (Array.isArray(criteria) && criteria.length === 0)
-        ) {
+        if (OrmUtils.isCriteriaNullOrEmpty(criteria)) {
             return Promise.reject(
                 new TypeORMError(
                     `Empty criteria(s) are not allowed for the update method.`,
@@ -774,12 +770,7 @@ export class EntityManager {
             )
         }
 
-        if (
-            typeof criteria === "string" ||
-            typeof criteria === "number" ||
-            criteria instanceof Date ||
-            Array.isArray(criteria)
-        ) {
+        if (OrmUtils.isPrimitiveCriteria(criteria)) {
             return this.createQueryBuilder()
                 .update(target)
                 .set(partialEntity)
@@ -815,12 +806,7 @@ export class EntityManager {
             | any,
     ): Promise<DeleteResult> {
         // if user passed empty criteria or empty list of criterias, then throw an error
-        if (
-            criteria === undefined ||
-            criteria === null ||
-            criteria === "" ||
-            (Array.isArray(criteria) && criteria.length === 0)
-        ) {
+        if (OrmUtils.isCriteriaNullOrEmpty(criteria)) {
             return Promise.reject(
                 new TypeORMError(
                     `Empty criteria(s) are not allowed for the delete method.`,
@@ -828,12 +814,7 @@ export class EntityManager {
             )
         }
 
-        if (
-            typeof criteria === "string" ||
-            typeof criteria === "number" ||
-            criteria instanceof Date ||
-            Array.isArray(criteria)
-        ) {
+        if (OrmUtils.isPrimitiveCriteria(criteria)) {
             return this.createQueryBuilder()
                 .delete()
                 .from(targetOrEntity)
@@ -869,12 +850,7 @@ export class EntityManager {
             | any,
     ): Promise<UpdateResult> {
         // if user passed empty criteria or empty list of criterias, then throw an error
-        if (
-            criteria === undefined ||
-            criteria === null ||
-            criteria === "" ||
-            (Array.isArray(criteria) && criteria.length === 0)
-        ) {
+        if (OrmUtils.isCriteriaNullOrEmpty(criteria)) {
             return Promise.reject(
                 new TypeORMError(
                     `Empty criteria(s) are not allowed for the delete method.`,
@@ -882,12 +858,7 @@ export class EntityManager {
             )
         }
 
-        if (
-            typeof criteria === "string" ||
-            typeof criteria === "number" ||
-            criteria instanceof Date ||
-            Array.isArray(criteria)
-        ) {
+        if (OrmUtils.isPrimitiveCriteria(criteria)) {
             return this.createQueryBuilder()
                 .softDelete()
                 .from(targetOrEntity)
@@ -923,12 +894,7 @@ export class EntityManager {
             | any,
     ): Promise<UpdateResult> {
         // if user passed empty criteria or empty list of criterias, then throw an error
-        if (
-            criteria === undefined ||
-            criteria === null ||
-            criteria === "" ||
-            (Array.isArray(criteria) && criteria.length === 0)
-        ) {
+        if (OrmUtils.isCriteriaNullOrEmpty(criteria)) {
             return Promise.reject(
                 new TypeORMError(
                     `Empty criteria(s) are not allowed for the delete method.`,
@@ -936,12 +902,7 @@ export class EntityManager {
             )
         }
 
-        if (
-            typeof criteria === "string" ||
-            typeof criteria === "number" ||
-            criteria instanceof Date ||
-            Array.isArray(criteria)
-        ) {
+        if (OrmUtils.isPrimitiveCriteria(criteria)) {
             return this.createQueryBuilder()
                 .restore()
                 .from(targetOrEntity)
--- a/src/util/OrmUtils.ts
+++ b/src/util/OrmUtils.ts
@@ -1,4 +1,8 @@
 import { ObjectLiteral } from "../common/ObjectLiteral"
+import {
+    PrimitiveCriteria,
+    SinglePrimitiveCriteria,
+} from "../common/PrimitiveCriteria"
 
 export class OrmUtils {
     // -------------------------------------------------------------------------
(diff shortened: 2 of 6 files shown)


maxbronnikov10:
@michaelbromley hello, sorry for tag, can u check this PR please? I think it is quite simple

maxbronnikov10:
@alumni hi! can u pls check PR?

maxbronnikov10:
@naorpeled hello! Can u check this PR please?

maxbronnikov10:
@sgarner hello! Can u check this PR please?

src/entity-manager/EntityManager.ts
```diff
@@ -778,7 +778,9 @@ export class EntityManager {
             typeof criteria === "string" ||
             typeof criteria === "number" ||
             criteria instanceof Date ||
-            Array.isArray(criteria)
+            (Array.isArray(criteria) &&
+                typeof criteria[0] !== "object" &&
+                criteria[0] !== null)
```
sgarner, line 783: I would prefer that the type check on the array element matches the type check on a single value, i.e. we should specifically allow the array to contain only elements which are 'string or number or Date' rather than containing anything that's 'not an object or null'.

For bonus points, it would be nice to abstract this logic... Something like the following could be added to OrmUtils and re-used across all these checks:

```ts
static isIdType(value: unknown): value is string | number | Date {
    return (
        typeof value === "string" ||
        typeof value === "number" ||
        value instanceof Date
    )
}

static isIdTypeOrList(value: unknown): value is string | number | Date | (string | number | Date)[] {
    if (Array.isArray(value)) {
        return value.every((el) => this.isIdType(el))
    }

    return this.isIdType(value)
}
```

sgarner (commented):
This is a nice fix, thanks @maxbronnikov10 !

One suggestion, otherwise looks good

src/entity-manager/EntityManager.ts
maxbronnikov10: fix done
i also added type alias for `string | number | Date` and organized OrmUtils(private methods at the bottom of file)

maxbronnikov10: omg i forgot run formatting, commited again, sorry

src/util/OrmUtils.ts
```diff
@@ -415,6 +323,27 @@ export class OrmUtils {
         return undefined
     }
 
+    public static isCriteriaNullOrEmpty(criteria: unknown): boolean {
```
sgarner, line 326: This could be named `isNullOrEmpty` to allow for broader use?

src/util/OrmUtils.ts
```diff
@@ -522,4 +451,110 @@ export class OrmUtils {
...
+        }
+
+        if (Array.isArray(target) && Array.isArray(source)) {
+            for (let key = 0; key < source.length; key++) {
+                this.mergeArrayKey(target, key, source[key], memo)
+            }
+        }
+    }
+
+    private static isSinglePrimitiveCriteria(
```
sgarner, line 551: I think this could be made public, even if we're not using it anywhere else yet.

sgarner (approved):
Couple more suggestions, otherwise LGTM :+1:

src/util/OrmUtils.ts
maxbronnikov10: specifically made it with prefix "criteria" based on my experience that each class have own set of nullish values
but i can rewrite this if u not same opinion

src/util/OrmUtils.ts
maxbronnikov10: good
i will be change it after commitment in previous comment - for run tests once

src/util/OrmUtils.ts
sgarner: Occurs to me now that this method ought to also check if an empty object (`Object.keys(criteria).length === 0`) is passed?

If we cover all those bases then this seems like a pretty good generic definition of "empty". But you make a fair point as well, so the current name is acceptable if you want to keep it.

maxbronnikov10: yeah u right, object without keys also not valid

src/util/OrmUtils.ts
```diff
@@ -415,6 +323,38 @@ export class OrmUtils {
         return undefined
     }
 
+    public static isCriteriaNullOrEmpty(criteria: unknown): boolean {
+        return (
+            criteria === undefined ||
+            criteria === null ||
+            criteria === "" ||
+            (Array.isArray(criteria) && criteria.length === 0) ||
+            Object.keys(criteria).length === 0
```
sgarner, line 332: To be safe:
```suggestion
            (typeof criteria === 'object' && Object.keys(criteria).length === 0)
```

sgarner (changes requested):

sgarner (approved):

jovanadjuric (approved):

src/util/OrmUtils.ts
```diff
@@ -415,6 +323,39 @@ export class OrmUtils {
...
+            criteria === undefined ||
+            criteria === null ||
+            criteria === "" ||
+            (Array.isArray(criteria) && criteria.length === 0) ||
+            (OrmUtils.isPlainObject(criteria) &&
+                Object.keys(criteria).length === 0)
+        )
+    }
+
+    public static isSinglePrimitiveCriteria(
```
mguida22, line 337: Can you add a docstring comment for each of these new functions?

src/util/OrmUtils.ts
```diff
@@ -415,6 +323,39 @@ export class OrmUtils {
...
+            criteria instanceof Date
+        )
+    }
+
+    public static isPrimitiveCriteria(
+        criteria: unknown,
+    ): criteria is PrimitiveCriteria {
+        if (Array.isArray(criteria)) {
+            return criteria.every((value) =>
+                OrmUtils.isSinglePrimitiveCriteria(value),
```
mguida22, line 352: nit: IMO it's more clear to use `this` when calling methods on this class. There's a few more in these changes
```suggestion
                this.isSinglePrimitiveCriteria(value),
```

mguida22 (approved):
This is great, thanks for the contribution!

Two things that don't need to block merging this PR:
 - Since this is more of a core functionality that we want to support as opposed to a regression, I think these tests would be best under our [functional tests](https://github.com/typeorm/typeorm/tree/master/test/functional/).
 - A separate PR to update the docs to show examples of this functionality will help let people know this is possible now.

sgarner (approved):
LGTM

Let us know if you want to address @mguida22 's feedback in this PR, or merge and follow up separately.

maxbronnikov10:
> Collaborator

yeap, i want to merge now and add tests, docs in separated PR

mguida22:
> Thanks! I want to merge now and add tests, docs in separated PR

Created this issue to track the docs updates https://github.com/typeorm/typeorm/issues/11447

christian-forgacs:
CleanShot 2025-05-13 at 09 08 46

Is this behavior intentional? In our project, we actually handle exactly these cases and explicitly throw an exception to prevent any unintended side effects.

sgarner:
> Is this behavior intentional? In our project, we actually handle exactly these cases and explicitly throw an exception to prevent any unintended side effects.
> 
> ```ts
> console.log(isCriteriaNullOrEmpty({id: undefined})); // false > not empty
> console.log(isCriteriaNullOrEmpty([{}])); // false > not empty
> console.log(isCriteriaNullOrEmpty([{id: undefined}])); // false > not empty
> ```

The array containing only empty objects is interesting. Probably we should catch that one. :confused:

There is other work in progress regarding handling of undefined values: https://github.com/typeorm/typeorm/pull/11332

Issue #10955, opened 2024-07-02
Use alternative ADD VALUE syntax when adding a new value to enum


janzipek:
### Feature Description

When changing a new enum value is added TypeORM currently generates lengthy migration that:
 - renames the current enum type
 - creates new enum type with new values
 - changes the type of column to the new enum type casting values to string then to the target enum type
 - drops the now old enum type

This is a valid way to change enum type but it's pretty dangerous on big tables. The type change locks the whole table and the type casing basically causes update on the whole table which can take a while so as the result this migration can lock up big tables.

### The Solution

PostgreSQL actually has a special syntax (since v12): https://www.postgresql.org/docs/12/sql-altertype.html

You can add new values to enum type by simply running:
```
ALTER TYPE name ADD VALUE 'value'
```

TypeORM could detect that the new enum only contains new values and use this alternative way which doesn't cause any locking and is instant even on big tables.

### Considered Alternatives

Not many alternatives to this, the current implementation is correct.

### Additional Context

I have PR ready, coming soon

### Relevant Database Driver(s)

- [ ] aurora-mysql
- [ ] aurora-postgres
- [ ] better-sqlite3
- [ ] cockroachdb
- [ ] cordova
- [ ] expo
- [ ] mongodb
- [ ] mysql
- [ ] nativescript
- [ ] oracle
- [X] postgres
- [ ] react-native
- [ ] sap
- [ ] spanner
- [ ] sqlite
- [ ] sqlite-abstract
- [ ] sqljs
- [ ] sqlserver

### Are you willing to resolve this issue by submitting a Pull Request?

Yes, I have the time, and I know how to start.

bmbferreira:
this is still relevant and should not be closed

PR #10956, opened 2024-07-02
feat: Use ADD VALUE for new enum values

janzipek:
Use ADD VALUE syntax for adding new enum values in PostgreSQL instead of complicated 4-step migration

Closes: #10955

### Description of change

When adding a new enum value, TypeORM currently generates 4 step migration:

- rename the current enum type
- create new enum type with new values
- change the type of column to the new enum type casting values to string then to the target enum type
- drop the now old enum type

This is a correct migration, but it's problematic for bigger tables. The type change will lock all rows of the table and the type cast basically causes update on the whole table, which can take minutes for bigger tables.

PostgreSQL actually has a solution for this since version 12: `ADD VALUE`. To add a new value to enum, you can simply do this:

```sql
ALTER TYPE name ADD VALUE 'value'
```

This will add the value without any locking or updates.

### Pull-Request Checklist


jhcpeixoto (approved):

dancrew32:
Hey thanks for adding this improvement! It looks like it was approved several months ago. What's the next step to get this merged?

janzipek:
Not sure if I can do anything from my side. Seems like the only approval is from non-maintainer and there wasn't any merge in past 7 months.

maxtuzz:
+1 for this feature.

bmbferreira:
any news on this? this is a recurring issue when using typeorm...

src/driver/postgres/PostgresQueryRunner.ts
```diff
@@ -1578,22 +1591,41 @@ export class PostgresQueryRunner
...
+                            `ALTER TYPE ${oldEnumName} RENAME TO ${oldEnumNameWithoutSchema_old}`,
+                        ),
+                    )
+                }
+
                 downQueries.push(
                     new Query(
                         `ALTER TYPE ${oldEnumNameWithSchema_old} RENAME TO ${oldEnumNameWithoutSchema}`,
                     ),
                 )
```
sgarner: When `useAddValueForUp` is true, the `oldEnumNameWithSchema_old` won't have been created, so this down query isn't going to work?

Interspersing `if (!useAddValueForUp)` conditionals throughout this code is making the outcome hard to reason about. Instead, I think it would be preferable to separate these code paths at a higher level. (This may mean a few bits of duplicated code where the behavior is the same in both paths, but that's acceptable.)

sgarner (changes requested):

gioboa (commented):
Is this PR still valid? may I ask you to address the feedback please?

janzipek:
Sorry, I didn't notice the development is active again. I could take a look if I can resolve the conflict and address the comment, I just need a few days to allocate some time to this.

test/github-issues/10955/issue-10955.test.ts
```diff
@@ -0,0 +1,154 @@
...
+                )
+                migrationLog.downQueries[1].query.should.equal(
+                    `DROP TYPE "public"."foo_bar_enum"`,
+                )
+                migrationLog.downQueries[2].query.should.equal(
+                    `ALTER TABLE "foo" ALTER COLUMN "bar" TYPE "public"."foo_bar_enum_old" USING "bar"::"text"::"public"."foo_bar_enum_old"`,
+                )
+                migrationLog.downQueries[3].query.should.equal(
+                    `CREATE TYPE "public"."foo_bar_enum_old" AS ENUM('value1')`,
+                )
```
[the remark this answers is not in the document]
janzipek: The down migration queries are inverted, you need to read them from last to first. In this case the _old enum is created before using it so the migration is valid.

src/driver/postgres/PostgresQueryRunner.ts
```diff
@@ -1584,109 +1598,160 @@
...
-                // build column types
-                const upType = `${newEnumName}${arraySuffix} USING "${newColumn.name}"::"text"::${newEnumName}${arraySuffix}`
-                const downType = `${oldEnumNameWithSchema_old}${arraySuffix} USING "${newColumn.name}"::"text"::${oldEnumNameWithSchema_old}${arraySuffix}`
+                    downQueries.push(
+                        this.createEnumTypeSql(
+                            table,
+                            oldColumn,
+                            oldEnumNameWithSchema_old,
+                        ),
+                    )
```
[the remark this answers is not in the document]
janzipek: The down migration queries are inverted, you need to read them from last to first. In this case the _old enum is created before using it so the migration is valid.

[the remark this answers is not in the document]
janzipek: The drop happens before the rename, check the code again. Also see the test/github-issues/10955/issue-10955.ts test.

src/driver/postgres/PostgresQueryRunner.ts
janzipek: I've refactored the logic to be a bit more readable (clearly branching for each option).

Theoretically I could also separate it into methods as the AI suggests, but I don't see any other part of the process doing that.

gioboa (commented):
Can I ask you to solve the conflicts please?

src/driver/postgres/PostgresQueryRunner.ts
```diff
@@ -1584,109 +1598,161 @@ export class PostgresQueryRunner
...
-                        `ALTER TYPE ${oldEnumName} RENAME TO ${oldEnumNameWithoutSchema_old}`,
-                    ),
-                )
-                downQueries.push(
-                    new Query(
-                        `ALTER TYPE ${oldEnumNameWithSchema_old} RENAME TO ${oldEnumNameWithoutSchema}`,
-                    ),
-                )
+                // ADD VALUE allows us to just add new enum values without any type changes, but can of course only be used when only new values were added
+                if (useAddValueForUp) {
```
gioboa, line 1602: > [!NOTE] 
The GH diff is not clear but the code inside this branch is the new change and in the else branch there is the old behaviour with a bit of formatting.

src/driver/postgres/PostgresQueryRunner.ts
```diff
@@ -1553,6 +1554,19 @@ export class PostgresQueryRunner
             ) {
                 const arraySuffix = newColumn.isArray ? "[]" : ""
 
+                const { extraItems, missingItems } = OrmUtils.getArraysDiff(
+                    newColumn.enum!,
+                    oldColumn.enum!,
+                )
+
+                const version = await this.getVersion()
```
alumni: This creates another query and it's called in a loop (because this function is called in a loop).

The value is cached in `driver.version` when the data source is initialized, we can use the cached value instead.

src/driver/postgres/PostgresQueryRunner.ts
alumni: Indentation change. You can toggle whitespace from the display settings:

Image

Also: sometimes it's easier to check the PR in VSCode/Webstorm (you'd have to configure something similar depending on your IDE).

src/driver/postgres/PostgresQueryRunner.ts
gioboa: Nice catch @alumni

janzipek:
I'll verify that the syntax works with 9.1 and update the code to use the cached property

EDIT: We have to keep the minimum version at 12 since before that `ADD VALUE` can't be called inside transaction

See:
https://www.postgresql.org/docs/11/sql-altertype.html#id-1.9.3.42.7
https://www.postgresql.org/docs/12/sql-altertype.html#id-1.9.3.42.7

 - For **11**: `ALTER TYPE ... ADD VALUE` (the form that adds a new value to an enum type) cannot be executed inside a transaction block.
 - For **12**: If `ALTER TYPE ... ADD VALUE` (the form that adds a new value to an enum type) is executed inside a transaction block, the new value cannot be used until after the transaction has been committed.

src/driver/postgres/PostgresQueryRunner.ts
janzipek: Thanks, adjusted

alumni:
> We have to keep the minimum version at 12

Postgres 13 and earlier have already reached their end of life, see https://www.postgresql.org/support/versioning/.

At some point we should remove all version checks for an unsupported version. But let's keep this one for now.

Issue #5694, opened 2020-03-16
findOne with relations does two queries.


vikas-kyatannawar:
**Issue type:**

[ ] question
[ X] bug report
[ ] feature request
[ ] documentation issue

**Database system/driver:**

[ ] `cordova`
[ ] `mongodb`
[ ] `mssql`
[X ] `mysql` / `mariadb`
[ ] `oracle`
[ ] `postgres`
[ ] `cockroachdb`
[ ] `sqlite`
[ ] `sqljs`
[ ] `react-native`
[ ] `expo`

**TypeORM version:**

[X ] `latest`
[ ] `@next`
[ ] `0.x.x` (or put your version here)

**Steps to reproduce or a small repository showing the problem:**
https://github.com/vikas-kyatannawar/typeorm-find
<!--
    To answer those questions you need to put "x" inside the square brackets, for example:
    [x] `mysql`
    [ ] `postgres`

    Also, please format your code properly (by taking code blocks into ```ts .... ```)
--!>

EricRabil:
I too am experiencing this – I am able to circumvent this by calling `find()` instead, however it feels dirty to do so.

Moreover, it appears to be selecting *all* columns in both queries, not just the latter query. Here's the logged queries when I call `findOne()`

```sql
query: SELECT DISTINCT "distinctAlias"."User_uuid" as "ids_User_uuid" FROM (SELECT "User"."uuid" AS "User_uuid", "User"."userID" AS "User_userID", "User"."passwordHash" AS "User_passwordHash", "User"."excludes" AS "User_excludes", "User_oAuthLinks"."uuid" AS "User_oAuthLinks_uuid", "User_oAuthLinks"."platform" AS "User_oAuthLinks_platform", "User_oAuthLinks"."platformID" AS "User_oAuthLinks_platformID", "User_oAuthLinks"."pipeDirection" AS "User_oAuthLinks_pipeDirection", "User_oAuthLinks"."userUuid" AS "User_oAuthLinks_userUuid" FROM "user" "User" LEFT JOIN "o_auth_link" "User_oAuthLinks" ON "User_oAuthLinks"."userUuid"="User"."uuid" WHERE "User"."userID" = $1) "distinctAlias" ORDER BY "User_uuid" ASC LIMIT 1 -- PARAMETERS: ["eric"]
[...]

PR #10976, opened 2024-07-12
add additional condition to execute load distinct id

gunb0s:
### Description of change
I fully understand why `typeorm` look for `distinct id` when finding conditions such as `join`, `take`, and `skip` together,
but it seems unnecessary to look up distinct id for all relation ships. For example, when you have the following `Order`, `Delivery`, `OrderItem`
```typescript
export class OrderEntity {
    id: number

    @OneToOne(() => DeliveryEntity)
    @JoinColumn()
    delivery: DeliveryEntity
}

export class DeliveryEntity {
    id: number
}

export class OrderItemEntity {
    id: number

    @Column()
    name: string

    @ManyToOne(() => OrderEntity)
    @JoinColumn()
    order: OrderEntity
}
```
If you query `OrderItem` by joining with `Order`, it makes enough sense to lookup distinct id.
but if you only want to join Delivery, I think there is no point in getting distinct id.
In current `typeorm`, if you run `findOne` by joining `Order` with `Delivery`, `typeorm` will run the query to get distinct id as the first query, but this query seems meaningless.
So, I think we need to **_modify the conditions for executing the query to get distinct id_**.

### related issue
https://github.com/typeorm/typeorm/issues/4998
https://github.com/typeorm/typeorm/issues/5694

### Pull-Request Checklist

### DIFF
--- a/test/functional/query-builder/time-travel-query/time-travel-query.ts
+++ b/test/functional/query-builder/time-travel-query/time-travel-query.ts
@@ -145,68 +145,6 @@ describe("query builder > time-travel-query", () => {
             }),
         ))
 
-    it("should execute time travel query with JOIN and skip/take options", () =>
-        Promise.all(
-            connections.map(async (connection) => {
-                const accountRepository = await connection.getRepository(
-                    Account,
-                )
-                const personRepository = await connection.getRepository(Person)
-
-                // create persons and accounts
-                for (let i = 1; i < 6; i++) {
-                    const account = new Account()
-                    account.name = `Person_${i}`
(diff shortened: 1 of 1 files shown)


gunb0s:
why oracle test failed with 
`Too long with no output (exceeded 10m0s): context deadline exceeded` ?? ..

gioboa (commented):
We aren't creating tests for specific GitHub issues any more. Instead, tests should be added to the suites for the related functionality.

Can you please move these tests please?

Thanks 🙏

PR #10986, opened 2024-07-18
feat: support PSQL 15 unique nulls not distinct

### LINKED ISSUE #9827: Support PostgreSQL 15 UNIQUE NULLS NOT DISTINCT
### Feature Description

Support option to treat nulls as non-distinct in unique indexes and constraints within postgres databases. This is useful for enforcing multi-column unique indexes and constraints can't have duplicate entires even if one or more columns are nullable. 

PostgreSQL Documentation:
> By default, NULL values are not treated as distinct entries. Specifying NULLS NOT DISTINCT on unique indexes / constraints will cause NULL values to be treated distinctly. - [Feature Description](https://www.postgresql.org/about/featurematrix/detail/392/)

[Unique Constraints](https://www.postgresql.org/docs/current/ddl-constraints.html#DDL-CONSTRAINTS-UNIQUE-CONSTRAINTS)
[Create Index](https://www.postgresql.org/docs/15/sql-createindex.html)

### The Solution

Decorators for unique index and constraint have `nullsNotDistinct` option (default: `false`) applicable to pg15 and above only.

Add `nullsNotDistinct` option to `Unique` decorator.
```
interface UniqueOptions {
    /**
     * By default, NULL values are not treated as distinct entries. 
     * Specifying NULLS NOT DISTINCT on unique constraints 
     * will cause NULL values to be treated distinctly.
     *
     * Works only in PostgreSQL 15 and above.
     */
    nullsNotDistinct?: boolean
```
Add `nullsNotDistinct` option to `Index` decorator.
```
interface IndexOptions {
    /**
     * Indicates if this composite index must be unique or not.
     */
    unique?: boolean
    ...
    /**
     * By default, NULL values are not treated as distinct entries. 
     * Specifying NULLS NOT DISTINCT on unique indexes 
     * will cause NULL values to be treated distinctly.
     * 
     * Only applicable when unique: true.
     * Works only in PostgreSQL 15 and above.
     */
    nullsNotDistinct?: boolean
}
```

### Considered Alternatives

Alternatives
- add NULLS NOT DISTINCT to migrations manually
- modify migrations to use COALESCE to map nulls to a single value

Both alternatives do not work with synchronization since they require modifying migrations

### Additional Context

_No response_

- [ ] aurora-mysql
- [ ] aurora-postgres
- [ ] better-sqlite3
- [ ] cockroachdb
- [ ] cordova
- [ ] expo
- [ ] mongodb
- [ ] mysql
- [ ] nativescript
- [ ] oracle
- [X] postgres
- [ ] react-native
- [ ] sap
- [ ] spanner
- [ ] sqlite
- [ ] sqlite-abstract
- [ ] sqljs
- [ ] sqlserver

a-fournier:
Closes: #9827

### Description of change

This MR fix adds the nullsNotDistinct option for unique constraints and indexes.
This option is available with Postgresql 15 to handle all null values as the same value.

Example : 
A Table(id, value), without a unique null not distinct on the value column
You'll be able to create (1, NULL) and (2, NULL) even if it should be unique

A Table(id, value), with a unique null not distinct on the value column
If you create (1, NULL) and then (2, NULL), you'll get a duplicate key error

### Pull-Request Checklist

- [x] This pull request links relevant issues as `Fixes #9827`


a-fournier:
Hi @pleerock, all tests seem to pass but the Oracle CI reaches the timeout. 
I retried a few times and it's still the case. 
Do you have an idea to fix this, should you increase the delay?

danLeBrown:
Oh, dear! It would've been good to have this feature running. Nice work, @a-fournier!

oscardianno:
Nice! Any progress on this?

danLeBrown:
> Nice! Any progress on this?

The new PostgreSQL version helped me with my issue. So I found a way around it

davidenwang:
This is an awesome PR, anything we can do to get it merged? :)

mguida22:
@a-fournier thanks for this contribution! I'll review this soon.

If you can do a few things it will improve this PR:
 - merge master into this branch to get the latest testing setup to run
 - add documentation. This can be done as a separate PR and doesn't need to block
 - add tests for the case where `nullsNotDistinct: false`
 - move the issue tests to [functional tests](https://github.com/typeorm/typeorm/tree/master/test/functional). Our docs are aren't clear on this right now, but this is new functionality (not testing a regression).

michaelbromley (commented):
Thanks for this great contribution!

I agree with Mike's suggestions above - if you are able to do those then this should be good to merge.

danLeBrown:
I'll see if I can get it done during the weekend – that is, if @a-fournier doesn't mind, of course...

(I'm already excited thinking about the possibility of my first non-doc related commit)

mguida22:
@alumni @OSA413 do we have a way to run tests against more than one postgres version? This change requires behavior introduced in postgres 15+, but I would like to have tests that verify postgres versions below 15 continue to function properly.

OSA413:
Yeah, I think this is possible to run tests on several Postgres versions

alumni:
Yeah, we just need to add another workflow using another version of postgis, e.g. 17-3.5.

sgarner (commented):
This is a good feature but I think the design is too closely tied to the PostgreSQL implementation. Although `NULLS NOT DISTINCT` is a PostgreSQL syntax extension, the concept can be applied to other databases.

I know in SQL Server that it works in reverse; by default a unique index treats nulls as equivalent, and you have to create a filtered index to obtain the SQL standard behavior (treating nulls as distinct values). It would be great if we could make it work there as well. See https://stackoverflow.com/a/767702/2168021

Rather than a boolean option, what if it was more like:

```ts
nulls?: 'distinct' | 'equivalent'
```

src/driver/postgres/PostgresDriver.ts
```diff
@@ -1388,6 +1388,27 @@ export class PostgresDriver implements Driver {
...
 
+    /**
+     * Returns true if the current version of Postgres supports nulls not distinct.
+     *
+     * Since this feature is only implemented in PostgreSQL 15 and above, other
+     * versions of Postgres will throw an error when using nullsNotDistinct unless
+     * throwOnUnsupported is set to false.
+     */
+    isNullsNotDistinctSupported({
+        throwOnUnsupported = true,
```
alumni, line 1450: I would remove this parameter and gracefully handle it in each situation (i.e. without throwing).

test/functional/indices/nulls-not-distinct/entity/ConstraintEntity.ts
```diff
@@ -0,0 +1,28 @@
+import { Entity } from "../../../../../src/decorator/entity/Entity"
+import { Column } from "../../../../../src/decorator/columns/Column"
+import { PrimaryGeneratedColumn } from "../../../../../src/decorator/columns/PrimaryGeneratedColumn"
+import { Unique } from "../../../../../src"
+
+@Entity()
+@Unique("unique_constraint_nulls_not_distinct", ["value"], {
+    nullsNotDistinct: true,
+})
+export class ConstraintEntityNullsNotDistinct {
```
alumni, line 10: I would place each entity in its own file

test/functional/indices/nulls-not-distinct/entity/IndexEntity.ts
```diff
@@ -0,0 +1,36 @@
...
+    @Column({ type: "text", nullable: true })
+    other_value: string
+}
+
+@Entity()
+@Index("unique_index_nulls_distinct", ["value", "other_value"], {
+    unique: true,
+    nullsNotDistinct: false,
+})
+export class IndexEntityNullsDistinct {
```
alumni, line 27: same as before, i would place each entity in its own file.

test/functional/indices/nulls-not-distinct/entity/IndexEntity.ts
```diff
@@ -0,0 +1,36 @@
...
+@Entity()
+@Index("unique_index_nulls_not_distinct", ["value", "other_value"], {
+    unique: true,
+    nullsNotDistinct: true,
+})
+export class IndexEntityNullsNotDistinct {
+    @PrimaryGeneratedColumn()
+    id: number
+
+    @Column({ type: "text", nullable: true })
```
alumni, line 15: We can maybe remove type and let it be inferred from the `string` type of the `value` property.

Then, if other drivers would support this feature, they would only need to add the driver to the list of supported drivers in the test file.

test/functional/indices/nulls-not-distinct/nulls-not-distinct.ts
```diff
@@ -0,0 +1,105 @@
...
+import {
+    IndexEntityNullsNotDistinct,
+    IndexEntityNullsDistinct,
+} from "./entity/IndexEntity"
+import { VersionUtils } from "../../../../src/util/VersionUtils"
+
+describe("indices > nulls not distinct", () => {
+    let connections: DataSource[] = []
+    before(async () => {
+        try {
```
alumni, line 20: there shouldn't be any try-catch here :)

test/functional/indices/nulls-not-distinct/nulls-not-distinct.ts
```diff
@@ -0,0 +1,105 @@
...
+            )
+            error.name.should.be.equal("TypeORMError")
+        }
+    })
+    after(() => {
+        if (connections && connections.length > 0) {
+            closeTestingConnections(connections)
+        }
+    })
+    describe("constraints", () => {
```
alumni, line 44: One empty line between each `describe` / `it` test would make it easier to spot the tests.

test/functional/indices/nulls-not-distinct/nulls-not-distinct.ts
```diff
@@ -0,0 +1,105 @@
...
+                }),
+            ))
+        it("should create a unique constraint with nulls distinct", () =>
+            Promise.all(
+                connections.map(async (connection) => {
+                    const queryRunner = connection.createQueryRunner()
+                    const table = await queryRunner.getTable(
+                        "constraint_entity_nulls_distinct",
+                    )
+                    table?.uniques[0].nullsNotDistinct?.should.be.equal(false)
```
alumni, line 70: Would be great to use `chai.expect`, it's closer to natural language:

```ts
expect(table?.uniques[0].nullsNotDistinct).to.equal(false)
```

gioboa (commented):
is this PR still valid? may I ask you to address the feedback please?

gioboa (commented):
@a-fournier  thanks for opening this PR.
I'm closing this for now because it's stale and with conflicts, feel free to create a new PR with your improvements. :muscle:

PR #10993, opened 2024-07-22
fix: fix working with tables with quotes in the names for PostgreSQL …

### LINKED ISSUE #10991: PostgresQueryRunner.getTables fails when table names contain '"'-sign.
### Issue description

If DB has a table with the name "Table with ""quotes""" runner.getTables triggers an exception

### Expected Behavior

runner.getTables works the same way as with the DB with regular table names only. According to the [documentation](https://www.postgresql.org/docs/current/sql-syntax-lexical.html), table names can have quote signs

### Actual Behavior

The request below returns an error "invalid name syntax":
`SELECT "table_schema", "table_name", obj_description(('"' || "table_schema" || '"."' || "table_name" || '"')::regclass, 'pg_class') AS table_comment FROM "information_schema"."tables"`

### Steps to reproduce

1. Create any PostgreSQL DB
2. Create a table `create table public."Table with ""quotes""" ();`
3. Create a connection to the DB using createConnection
4. Try to call `connection.createQueryRunner().getTables()`

### My Environment

| Node.js version     | 18.19.0  |
| Typescript version  | 4.3.5  |

### Additional Context

_No response_

- [ ] aurora-mysql
- [ ] aurora-postgres
- [ ] better-sqlite3
- [ ] cockroachdb
- [ ] cordova
- [ ] expo
- [ ] mongodb
- [ ] mysql
- [ ] nativescript
- [ ] oracle
- [X] postgres
- [ ] react-native
- [ ] sap
- [ ] spanner
- [ ] sqlite
- [ ] sqlite-abstract
- [ ] sqljs
- [ ] sqlserver

iskalyakin:
…(#10991)

Fix working with tables with quotes in the names for PostgreSQL

Closes: #10991

### Description of change
There were a few changes in PostgresQueryRunner with changing quoting table_schema and table_name from '"' || "table_schema" || '"' to quote_ident("table_schema"), which allows to deal with table names containing quotes.
It's important because tables with these unusual names can exist, while sometimes it's impossible just rename those tables as this is an arbitrary database, the structure of which needs to be obtained.

Case: PostgreSQL DB containg table with the name `This is the "Table Name"`.
Current behaviour: connection.createQueryRunner().getTables() throws an error.
Next behaviour: connection.createQueryRunner().getTables() works

Fixes #10991

### Pull-Request Checklist

- [-] Documentation has been updated to reflect this change N/A

### DIFF
--- a/src/driver/postgres/PostgresQueryRunner.ts
+++ b/src/driver/postgres/PostgresQueryRunner.ts
@@ -601,10 +601,22 @@ export class PostgresQueryRunner
                 downQueries.push(this.dropIndexSql(table, index))
             })
         }
-        
+
         if (table.comment) {
-            upQueries.push(new Query("COMMENT ON TABLE " + this.escapePath(table) + " IS '" + table.comment + "'"));
-            downQueries.push(new Query("COMMENT ON TABLE " + this.escapePath(table) + " IS NULL"));
+            upQueries.push(
+                new Query(
+                    "COMMENT ON TABLE " +
+                        this.escapePath(table) +
+                        " IS '" +
+                        table.comment +
+                        "'",
+                ),
+            )
+            downQueries.push(
+                new Query(
+                    "COMMENT ON TABLE " + this.escapePath(table) + " IS NULL",
+                ),
+            )
         }
 
         await this.executeQueries(upQueries, downQueries)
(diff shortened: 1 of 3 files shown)


iskalyakin:
It looks like the failed tests (oracle and sqlite better-sqlite3 sqljs) are not related to the PR. These failures also appeared in other recent PRs.

gioboa (commented):
We aren't creating tests for specific GitHub issues any more. Instead, tests should be added to the suites for the related functionality.

Can you please move these tests please?

Thanks 🙏

alumni:
Would be nice if the same fix would be done for `CockroachQueryRunner` as well :)

alumni (approved):

linbrandon4:
Good fix. Using quote_ident() instead of manually doing '"' || table_name || '"' is the right move — it avoids breaking when table/schema names already contain quotes, and it’s the safer way to build identifiers in Postgres. The example case (“This is the "Table Name"”) makes sense, and the change should stop getTables() from blowing up on weird legacy DBs.

Also glad this got extended to CockroachQueryRunner too, since it has the same quoting problem in practice.

A couple things I’d double-check:
- Make sure this is only being used for identifiers (schema/table names) and not values — quote_ident is correct for identifiers, but it shouldn’t accidentally replace value escaping.
- Since the bot raised “possible SQL injection” flags, it’s worth confirming all identifier inputs come from the DB metadata / controlled sources, and not raw user input.
- Tests: I know the repo is moving away from issue-specific test folders, but as long as the tests cover quoted identifiers for both Postgres + Cockroach (schema + table name cases), that’s the important part.

PR #11004, opened 2024-07-24
fix: migration generation to save in specified datasource directory o…

galenzo17:
…r default directory if not set

### Pull-Request Checklist

### DIFF
--- a/src/commands/MigrationGenerateCommand.ts
+++ b/src/commands/MigrationGenerateCommand.ts
@@ -7,6 +7,7 @@ import { PlatformTools } from "../platform/PlatformTools"
 import { DataSource } from "../data-source"
 import * as path from "path"
 import process from "process"
+import { GenerateMigrationPathUtils } from "../util/GenerateMigrationPathUtils"
 
 /**
  * Generates a new migration file with sql needs to be executed to update schema.
@@ -85,6 +86,12 @@ export class MigrationGenerateCommand implements yargs.CommandModule {
                 logging: false,
             })
             await dataSource.initialize()
+
+            let migrationsDir =
+                GenerateMigrationPathUtils.getMigrationsDir(dataSource)
+            if (!migrationsDir) {
+                migrationsDir = path.dirname(fullPath)
+            }
 
             const upSqls: string[] = [],
                 downSqls: string[] = []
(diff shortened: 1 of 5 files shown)


simplenotezy:
Personally a very welcomed change! A few outsider observations:

1. Perhaps you should document in the `BaseDataSourceOptions` type definition that the first entry in `migrations` array will be the default location for where the CLI will put newly created migrations (https://github.com/typeorm/typeorm/blob/master/src/data-source/BaseDataSourceOptions.ts#L41-L45)
2. Consider reusing [CommandUtils.ts](https://github.com/typeorm/typeorm/blob/master/src/commands/CommandUtils.ts) instead of creating a new single method class (`GenerateMigrationPathUtils`)
3. You are missing a similar implementation of the migrations base path in the [MigrationCreateCommand.ts](https://github.com/typeorm/typeorm/blob/master/src/commands/MigrationCreateCommand.ts) file
4. The [MigrationCreateCommand.ts](https://github.com/typeorm/typeorm/blob/master/src/commands/MigrationCreateCommand.ts) file has a way to infer an absolute path if prepended with `/`. I think you should adopt this same behaviour in the new implementation for backwards compatibility

galenzo17:
> Personally a very welcomed change! A few outsider observations:
> 
>     1. Perhaps you should document in the `BaseDataSourceOptions` type definition that the first entry in `migrations` array will be the default location for where the CLI will put newly created migrations (https://github.com/typeorm/typeorm/blob/master/src/data-source/BaseDataSourceOptions.ts#L41-L45)
> 
>     2. Consider reusing [CommandUtils.ts](https://github.com/typeorm/typeorm/blob/master/src/commands/CommandUtils.ts) instead of creating a new single method class (`GenerateMigrationPathUtils`)
> 
>     3. You are missing a similar implementation of the migrations base path in the [MigrationCreateCommand.ts](https://github.com/typeorm/typeorm/blob/master/src/commands/MigrationCreateCommand.ts) file
> 
>     4. The [MigrationCreateCommand.ts](https://github.com/typeorm/typeorm/blob/master/src/commands/MigrationCreateCommand.ts) file has a way to infer an absolute path if prepended with `/`. I think you should adopt this same behaviour in the new implementation for backwards compatibility

I completely agree with the suggested changes. I'll implement them right away. Thank you very much for the feedback!

gioboa (commented):
Few tests are failing, can you verify them please?

PR #11042, opened 2024-08-26
fix: where query on Foreign Primary key

MinDBreaK:
### Description of change

Where query are not properly constructed if the query is not the id. For example, it tries to interpret rfk: {otherkey: "value"} as the id, instead of the key to check

### Pull-Request Checklist

- [x ] Code is up-to-date with the `master` branch
- [ x] `npm run format` to apply prettier formatting
- [ x] `npm run test` passes with this change
- [x ] The new commits follow conventions explained in [CONTRIBUTING.md](https://github.com/typeorm/typeorm/blob/master/CONTRIBUTING.md)

No tests has been added yet, but if it is reviewed and approved, it will be added. 

I was currently unable to run the full test suite, as it failed on code that were not related. I'd like to see it run on CI where it could have better chances

### DIFF
--- a/src/query-builder/SelectQueryBuilder.ts
+++ b/src/query-builder/SelectQueryBuilder.ts
@@ -4258,81 +4258,7 @@ export class SelectQueryBuilder<Entity extends ObjectLiteral>
                         metadata,
                     )
 
-                if (column) {
-                    let aliasPath = `${alias}.${propertyPath}`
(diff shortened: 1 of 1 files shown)


gioboa (commented):
Thanks @MinDBreaK for your commitment
Can you add a test for this scenario please?

Issue #9936, opened 2023-04-11
relationLoadStrategy query: ER_NONUNIQ_TABLE during recursive call


Greg57070:
### Issue description

relationLoadStrategy query: ER_NONUNIQ_TABLE during recursive call

### Expected Behavior

When using a ManyToOne from an Entity to the same Entity, TypeORM should use an alias when joining the table

### Actual Behavior

The table "Folder" is called twice with the same alias :

SELECT `Folder`.`id` AS `Folder_id`,
       `Folder`.`familyLinkId` AS `Folder_familyLinkId`,
       `Folder`.`id` AS `Folder_id`,
       `Folder`.`familyLinkId` AS `Folder_familyLinkId`
FROM `folder` `Folder`
INNER JOIN `folder` `Folder` ON `Folder`.`familyLinkId` = `Folder`.`id`
WHERE `Folder`.`id` IN (7486)

### Steps to reproduce

@Entity()
export class Folder {

	@PrimaryGeneratedColumn()
	id: number;

	@OneToMany((type) => Folder, (folder) => folder.family_link)
	@JoinTable()
	linked_folders: Folder[];

	@ManyToOne((type) => Folder, (folder) => folder.linked_folders, {
		nullable: true,
	})
	family_link: Folder;

}

folders = await folderRepository.find({
	relationLoadStrategy: "query",
	where: { id: In(foldersID) },
	relations: {
		family_link: true
	}
})

### My Environment

| Dependency          | Version  |
| ---                 | ---      |
| Operating System    |          |
| Node.js version     | 18.15.0  |
| Typescript version  | 4.9.5  |
| TypeORM version     | 0.3.12  |

### Additional Context

IMHO (I'm really not a javascript dev 😅), I think the problem is in FindOptionsUtils.ts, in applyRelationsRecursively.
In case of relationLoadStrategy query, it use "relation" but not "relationAlias"

Thank you all for the good work, keep up with it !  

### Relevant Database Driver(s)

- [ ] aurora-mysql
- [ ] aurora-postgres
- [ ] better-sqlite3
- [ ] cockroachdb
- [ ] cordova
- [ ] expo
- [ ] mongodb
- [X] mysql
- [ ] nativescript
- [ ] oracle
- [ ] postgres
- [ ] react-native
- [ ] sap
- [ ] spanner
- [ ] sqlite
- [ ] sqlite-abstract
- [ ] sqljs
- [ ] sqlserver

[...]

PR #11066, opened 2024-09-16
fix: relationLoadStrategy query self-referencing select

campmarc:
query cannot load referenced entity when its of the same type due to same table alias

Fixes: #9936

### Description of change

Self-Referencing relation fails to load when relationLoadStrategy=query, due to the table name being specified twice the same. Subsequent select must use to the alias field to compare `b.id = a.parent` instead of wrong `a.id = a.parent`.

### Pull-Request Checklist


iDschepe:
We applied [your fix](https://github.com/typeorm/typeorm/pull/11066/commits/2c66e7a2e4e468daf81584d1ada426d53b173282) as a `pnpm` patch for our **production** environment after finally upgrading TypeORM from `0.2.38` to `0.3.20`, and it works from our perspective. Thank you so much @campmarc for your contribution.

gioboa (commented):
Renamed the test file because it was with the old extension.
Finger crossed for the pipeline 🤞

pkuczynski (approved):

Issue #11085, opened 2024-10-04
BeforeQuery promises are not awaited before query execution


TanguyPoly:
### Issue description

BeforeQuery promises are not awaited before query execution

### Expected Behavior

When using a Subscriber, `beforeQuery` promises should be awaited and executed before the actual query.

### Actual Behavior

`beforeQuery` and `afterQuery` promises are all executed after the query

### Steps to reproduce

Enable requests logs, create a Subscriber with an async `beforeQuery`, and it will be run after the query.

### My Environment

| Dependency          | Version  |
| ---                 | ---      |
| Operating System    |   Ubuntu 24.04 / Containers with node:18.20.4-alpine3.20       |
| Node.js version     | 18.20.4  |
| Typescript version  | 5.0.4  |
| TypeORM version     | 0.3.20  |

### Additional Context

It seems that it is due to the fact that `await broadcasterResult.wait()` is [only called after the query execution](https://github.com/typeorm/typeorm/blob/master/src/driver/postgres/PostgresQueryRunner.ts#L333) in the `finally` statement. Is there a reason behind this technical choice ?

The `beforeQuery` and `afterQuery` where implemented after [this issue](https://github.com/typeorm/typeorm/issues/3302), maybe it was only intended for logging purposes ?

Context: i am trying to [implement RLS ](https://aws.amazon.com/fr/blogs/database/multi-tenant-data-isolation-with-postgresql-row-level-security/) on our project, and i would like to use subscribers to set the current tenant_id before executing some queries on tables with RLS enabled. 
Since there is no `beforeLoad` event (it was proposed in [this issue](https://github.com/typeorm/typeorm/issues/2266) but it was closed), i wanted to try to use the `beforeQuery` method

### Relevant Database Driver(s)

- [x] aurora-mysql
- [x] aurora-postgres
- [x] better-sqlite3
- [x] cockroachdb
- [x] cordova
- [x] expo
- [x] mongodb
- [x] mysql
- [x] nativescript
- [x] oracle
- [x] postgres
- [x] react-native
- [x] sap
- [x] spanner
- [x] sqlite
- [x] sqlite-abstract
- [x] sqljs
- [x] sqlserver

### Are you willing to resolve this issue by submitting a Pull Request?

Yes, I have the time, and I know how to start.

amiryadid:
I have also experienced this recently, did this ever work before? seems weird that this is the first time this issue is being brought up

TanguyPoly:
[...]

Issue #10212, opened 2023-07-18
Jest + TypeORM: Segmentation Fault


ermartinez13:
### Issue description

A 'Segmentation fault' error occurs when trying to initialize a SQLite database via TypeORM within a Jest test file.

### Expected Behavior

Execution of test script (eg. `jest`) successfully executes setup, teardown, and test case.

### Actual Behavior

Test script execution outputs `segmentation fault` error.

```bash
# output:
$ jest

RUNS  ./db.test.ts
Segmentation fault
```

### Steps to reproduce

Please visit the following sandbox to reproduce error: https://codesandbox.io/p/sandbox/jest-typeorm-sigsegv-xg5gx6

### My Environment

* note: for version column, first values represent versions in original repo and second values represent versions in codesandbox created to showcase bug

| Dependency          | Version  |
| ---                 | ---      |
| Operating System    |  Ventura 13.2.1 & unknown |
| Node.js version     | 18.12.1 & 18.16.0  |
| Typescript version  | ^5.0.2 & ^5.1.6  |
| TypeORM version     | ^0.3.16 & ^0.3.17  |

### Additional Context

# About the issue

A 'Segmentation fault' error occurs when trying to initialize a SQLite database via TypeORM within a Jest test file.

More especifically, it appears the value for 'entities' property in DataSource's constructor is related to the error. If an array containing a directory path is provided as value, error will surface. Otherwise, if an array containing an entity class is provided, test will execute successfully.

Note: Issue is present whether storing SQLite in-memory or disk file.

[TypeORM: DataSource options](https://typeorm.io/data-source-options)

# Workaround

Pass entity classes as argument, not directories, to DataSource's `entities` option.

```typescript
import { MyEntity } from "./entities/MyEntity.ts";
// ...
- entities: ["entities/*.ts"],
+ entities: [MyEntity]
```

### Relevant Database Driver(s)

- [ ] aurora-mysql
- [ ] aurora-postgres
- [ ] better-sqlite3
- [ ] cockroachdb
- [ ] cordova
- [ ] expo
- [ ] mongodb
- [ ] mysql
- [ ] nativescript
- [ ] oracle
- [ ] postgres
- [ ] react-native
- [ ] sap
- [ ] spanner
- [X] sqlite
- [ ] sqlite-abstract
- [ ] sqljs
- [ ] sqlserver

### Are you willing to resolve this issue by submitting a Pull Request?

Yes, I have the time, but I don't know how to start. I would need guidance.

[...]

PR #11086, opened 2024-10-04
fix: beforeQuery promises not awaited before query execution

TanguyPoly:
Closes: #11085

### Description of change

Now awaiting `broadcasterResult.wait()` before query execution, so that `beforeQuery` promises are executed before query execution

I added a new test to verify that  `beforeQuery` promise is awaited before query run

### Pull-Request Checklist

- [ ] Documentation has been updated to reflect this change -> N/A


amiryadid:
@TanguyPoly Did you reach out to anyone for reviewing and merging this?

TanguyPoly:
@amiryadid No i did not, I used a workaround in the mean time (extending repository methods, using the same name and then calling the original method via the entityManager, did the same for the createQueryBuilder)

src/driver/sap/SapQueryRunner.ts
```diff
@@ -207,6 +207,8 @@ export class SapQueryRunner extends BaseQueryRunner implements QueryRunner {
                 parameters,
             )
 
+            await broadcasterResult.wait()
```
alumni, line 210: Hmmm, isn't

```ts
this.broadcaster.broadcastBeforeQueryEvent(
    broadcasterResult,
    query,
    parameters,
)

await broadcasterResult.wait()
```
equivalent to:
```ts
await this.broadcaster.broadcast("BeforeQuery", query, parameters)
```
?

I'm also wondering why do we have to use `BroadcasterResult` at all. We could also slightly change the try-catch block and we'd be able to simply call `Broadcaster.broadcast` also for `AfterQuery`.

Were we using `BroadcasterResult` because maybe we thought that the listeners might throw?

sgarner (commented):
In `query()` methods, the `BroadcasterResult` is shared between multiple events 
 - BeforeQuery and AfterQuery. It is used to sum up and wait for the promises of all the subscribers across both events.

This is why `await broadcast()` is not used currently, because that only broadcasts a single event.

Adding a call to `wait()` on the broadcastResult before query execution means any promises returned from BeforeQuery subscribers will be resolved twice - once there, and once again after the query finishes. This may not have any noticeable effect, but... it smells a bit off to me.

It looks like the current implementation is not intended to allow BeforeQuery subscribers to complete before the query is executed, but rather for them to be executed in parallel with the query. The `BroadcastResult` is used for error handling, and perhaps to ease debugging.

As @alumni suggested the code could be revised to `broadcast()` the BeforeQuery event separately, wait for it before executing the query, and not share a `BroadcastResult` with the AfterQuery event. This _might_ be a breaking change though. I wonder what the use cases are for BeforeQuery subscribers returning promises?

sgarner:
I just read the linked issue and original PR which added these events. Now I'm inclined to think that the execution of 'before' subscribers in parallel with the query is just a bug.

In other contexts we do already await 'before' subscribers, e.g.
https://github.com/typeorm/typeorm/blob/513be336c863d5da49aa933806b17e3c95e83754/src/persistence/SubjectExecutor.ts#L111

So I agree with @alumni suggestion to use `await broadcast()`

sgarner:
> It's definitely a bug (there's no breaking change in this PR), I was just thinking a bit more about the exception handling :)

What I meant was, if any users are relying on the current behavior of BeforeQuery subscribers executing in parallel with the query, if we change this to happen in series then the behavior of their application may also be subtly changed.

But, since this behavior was not documented and may have originally been unintentional, I think we can say this was an implementation detail that shouldn't be relied upon, and therefore it isn't a breaking change.

> Not all drivers behave identically though, e.g. `CockroachQueryRunner` and `CordovaQueryRunner` have correct exception handling: `this.connect()` and `broadcastBeforeQueryEvent` are outside of the `try-catch-finally` block - this should probably happen for other drivers too.

The AfterQuery event interface includes properties for an error and success boolean, so they're clearly meant to be triggered in case of failure. It may be useful for that to include connection errors. But triggering AfterQuery for an error in the BeforeQuery subscribers probably makes less sense? Whichever way, it should certainly be made consistent between drivers.

Should `broadcast()` happen first, outside the try/catch, and then `connect()` inside it?

sgarner:
No, actually... we can't do that because if the user wants to execute their own queries inside the BeforeQuery subscriber, the provided QueryRunner won't be connected yet.

So we either:

1. Put everything inside the try/catch and accept that AfterQuery subscribers trigger on errors in the BeforeQuery subscribers
2. Put `broadcast('BeforeQuery')` and `connect()` outside the try/catch and accept that AfterQuery subscribers won't trigger for connection errors

alumni:
> 2\. Put `connect()` and `broadcast('BeforeQuery')` outside the try/catch and accept that AfterQuery subscribers will no longer trigger for connection errors

Exactly. This is what I am suggesting above. This is how CockroachDB does it, but not SAP (did not check more drivers).

sgarner:
> Exactly. This is what I am suggesting above. This is how CockroachDB does it, but not SAP (did not check more drivers). Then the events will be triggered correctly and the errors should be handled correctly.

OK, I looked through and found the following for `connect()` and broadcast of BeforeQuery events:

- ✅ BetterSqlite3QueryRunner does it outside the try/catch
- ✅ CockroachQueryRunner does it outside the try/catch
- ✅ CordovaQueryRunner does it outside the try/catch
- ✅ PostgresQueryRunner does it outside the try/catch
- ✅ OracleQueryRunner does it outside the try/catch
- ✅ SapQueryRunner does it outside the try/catch
- 🙅 ExpoQueryRunner does it inside the try/catch
- 🙅 MysqlQueryRunner does it inside the try/catch
- 🙅 ReactNativeQueryRunner does it inside the try/catch
- 🙅 SpannerQueryRunner does it inside the try/catch
- 🙅 SqlServerQueryRunner does it inside the try/catch
- 😕 SqliteQueryRunner broadcasts BeforeQuery event outside, then connects inside the try/catch
- 😕 SqljsQueryRunner broadcasts BeforeQuery event outside, doesn't connect() at all
- ❌  AuroraMysqlQueryRunner doesn't trigger BeforeQuery/AfterQuery events at all
- ❌  AuroraPostgresQueryRunner doesn't trigger BeforeQuery/AfterQuery events at all
- ❌  CapacitorQueryRunner doesn't trigger BeforeQuery/AfterQuery events at all
- ❌  MongoQueryRunner doesn't trigger BeforeQuery/AfterQuery events at all
- ❌  NativeScriptQueryRunner doesn't trigger BeforeQuery/AfterQuery events at all

alumni:
@amiryadid @TanguyPoly I would adjust the logic in the drivers to be the same everywhere - @sgarner made that nice overview above. The drivers marked with 🙅 and 😕 should behave similarly to the ones marked with ✅, i.e.:
  - `Broadcaster.broadcast("BeforeQuery")`
  - `connect()`
  - `new BroadcastResult()` (used only for AfterQuery)
  - `try`: execute `query`; `broadcastAfterQueryEvent(broadcastResult, result)`
  - `catch`: `broadcastAfterQueryEvent(broadcastResult, error)`
  - `await broadcastResult.wait()`

Regarding an ETA: we will have more regular releases, so when the PRs are ready (reviewed, approved and don't have breaking changes), they will be merged and released in the next release. If you want to contribute with something, I would say you can do it with code (PRs) 🙂

amiryadid:
@TanguyPoly do you think you'll have the time to implement the above-mentioned changes?
I'm personally on a tighter schedule ATM + will have overhead of learning the codebase, so if it's problematic for you to implement LMK and I'll try to squeeze that in perhaps.

BTW no pressure on that ETA, just asked out of curiosity, I'm sure this repo is a handful 🙂

TanguyPoly:
Hello everyone !

@amiryadid Yes sure i can have a look at it this week i think, I should be able to dedicate some time to it.

Thank you @sgarner & @alumni for taking the time to review & make suggestions ! :pray:

TanguyPoly:
I did the changes to all drivers with 🙅 & 😕 except the `SqljsQueryRunner` driver because it doesn't have a `connect` method unlike other drivers

src/driver/cordova/CordovaQueryRunner.ts
```diff
@@ -90,6 +86,8 @@ export class CordovaQueryRunner extends AbstractSqliteQueryRunner {
                 undefined,
             )
 
+            await broadcasterResult.wait()
```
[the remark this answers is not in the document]
TanguyPoly: Yes done !

src/driver/sap/SapQueryRunner.ts
TanguyPoly: Sorry i didn't understood well this comment and missed the requested changes, thank you for applying them in your commits :pray:

alumni (approved):

src/driver/sap/SapQueryRunner.ts
alumni: Ah, I thought because of the types 😅 I also had to fix the types for the two events that we care about in this PR.

Basically, I wanted to avoid the same `BroadcastResult` for two different events.

amiryadid:
Hey y'all 🙂
Noticed this was approved but not merged and now it has a few additional conflicts.
@TanguyPoly do you need assistance in resolving them?
I just don't want this PR to go stale when it's borderline merged

TanguyPoly:
@amiryadid Hello, the conflicts have been resolved, and i guess it will be merged when 2 reviewers will have approved this merge requested (i only had one approval until the conflict, hence why it was not merged before).

alumni (approved):

gioboa (approved):

src/driver/mysql/MysqlQueryRunner.ts
```diff
@@ -188,17 +188,15 @@ export class MysqlQueryRunner extends BaseQueryRunner implements QueryRunner {
...
         return new Promise(async (ok, fail) => {
+            const databaseConnection = await this.connect()
+
+            this.driver.connection.logger.logQuery(query, parameters, this)
+            await this.broadcaster.broadcast("BeforeQuery", query, parameters)
+
             const broadcasterResult = new BroadcasterResult()
+            const queryStartTime = Date.now()
 
             try {
```
sgarner, line 199: I know it is existing code not introduced in this PR, but we have some weird construction in this method (and replicated in some other QueryRunner classes similar to this one).

Firstly, returning a `new Promise()` is unnecessary in an `async` function. Then the weirdness is doubled by using an `async function` as the Promise executor: `new Promise(async (ok, fail) => {...})`.

I am concerned that moving `await this.connect()` inside this async Promise executor means that database connection errors will now be lost; an error here won't cause the Promise to reject, but it will never resolve, likely causing a hang.

Ideally the code should be rewritten as proper async/await style without `new Promise()`. Alternatively, the connection and before-query code should be moved outside of the Promise executor, or another try/catch block added to handle errors in this code.

src/subscriber/Broadcaster.ts
```diff
@@ -8,8 +8,15 @@ import { RelationMetadata } from "../metadata/RelationMetadata"
...
-    AfterQuery: () => void
+    BeforeQuery: (query: string, parameters: undefined | any[]) => void
+    AfterQuery: (
+        query: string,
+        parameters: undefined | any[],
+        success: boolean,
+        executionTime: undefined | number,
+        rawResults: undefined | any,
+        error: undefined | any,
+    ) => void
```
sgarner, line 19: For readability prefer to write `undefined` last in type unions, so the expected type is first:

```suggestion
    BeforeQuery: (query: string, parameters: any[] | undefined) => void
    AfterQuery: (
        query: string,
        parameters: any[] | undefined,
        success: boolean,
        executionTime: number | undefined,
        rawResults: any | undefined,
        error: any | undefined,
    ) => void
```

sgarner (changes requested):

src/driver/mysql/MysqlQueryRunner.ts
TanguyPoly: Good catch ! I've rewritten the code to get rid of them

TanguyPoly:
I tried to run the sqlite tests locally but most of them failed with a "Segmentation fault" as mentioned here: https://github.com/typeorm/typeorm/issues/10212

I was still able to run some of them and see that the issues seems to be fixed

alumni:
@sgarner @TanguyPoly what if we revert to the last commit where the tests where passing and merge this PR? There are quite a lot of changes, if we need to do something else, we can do it separately.

sgarner:
> @sgarner @TanguyPoly what if we revert to the last commit where the tests were passing and merge this PR? There are quite a lot of changes, if we need to do something else, we can do it separately.

Yes it looks like rewriting the query Promises is adding too much extra complexity here (mysql is broken as well as sqlite) and that effort would be better tackled in its own PR. But, it's not ideal to merge a PR with known bugs either.

@TanguyPoly let's revert the Promise changes back out. We still need to handle connection errors though, how about adding try/catch blocks instead?

TanguyPoly:
@sgarner Yes i can revert the changes and add try / catch blocks instead, seems to be a good option for the time being :+1:

TanguyPoly:
@sgarner Everything should be good now, it would be great if ths pr can be approved before having to solve other conflicts with master, thank you 🙏

sgarner (approved):
LGTM

Thanks very much for all your effort on this @TanguyPoly 🧡

Issue #4569, opened 2019-08-11
Create foreign key constraint without relation


jb4e:
**Issue type:**

[x] question
[ ] bug report
[ ] feature request
[ ] documentation issue

**Database system/driver:**

[ ] `cordova`
[ ] `mongodb`
[ ] `mssql`
[x] `mysql` / `mariadb`
[ ] `oracle`
[ ] `postgres`
[ ] `cockroachdb`
[ ] `sqlite`
[ ] `sqljs`
[ ] `react-native`
[ ] `expo`

**TypeORM version:**

[x] `latest`
[ ] `@next`
[ ] `0.x.x` (or put your version here)

I'm evaluating TypeORM and I'm trying to figure out if there is a way to define a foreign key constraint without having to define the relationship. All I care about is that the database enforces the constraint - I don't want or need the persistence, loading of relations, or the superfluous properties on my classes. For example: 
```
// A user can have multiple phone numbers
@Entity()
class PhoneNumber {
    @PrimaryGeneratedColumn(); 
    id: number; 
    
    @Column()
    @ForeignKey((u: User) => u.id)
    userId: string;

    @Column(); 
    phoneNumber: string; 
}
```

This would create a foreign key constraint, and nothing else. Is this possible somehow? 
I found a couple other posts requesting the same thing, but I'm not sure if it's possible. 
https://github.com/typeorm/typeorm/issues/188
https://stackoverflow.com/questions/55170906/typeorm-how-to-set-foreignkey-explicitly-without-having-property-for-loading-re

Thanks in advance.

devshorts:
I'm also encountering this. It seems odd to me that we are forced to require an object relationship vs just an id relationship.  Is there any way around this?

abcd-ca:
I just saw this answer https://stackoverflow.com/a/54098642/756177

diegoazh:
I think that you need to read this documentation [https://typeorm.io/#/relations-faq/how-to-use-relation-id-without-joining-relation](https://typeorm.io/#/relations-faq/how-to-use-relation-id-without-joining-relation)

jb4e:
Thanks, @diegoazh, but I don't think that addresses what I'm looking for. I'd like to be able to specify *only* profileId on the model, and not have a profile property. Something like this: 
```
    @Column({ nullable: true })
    @ForeignKey(() => Profile)
    profileId: number;
```

diegoazh:
[...]

Issue #11132, opened 2024-11-13
Declaring a composite FK where one of the columns is a part of a different FK messes up the query builder.


ThomasShih:
### Issue description

It seems that somehow, the query builder messes up and tries to use build the query via the direct relationship instead of what is asked.

### Expected Behavior

In the code to reproduce:

- Test that "should be able to query consuming entity with root entity"
  - `sql` should be: 
```sql
select
	"consuming"."id" as "consuming_id",
	"consuming"."rootId" as "consuming_rootId",
	"consuming"."nodeId" as "consuming_nodeId",
	"root"."id" as "root_id"
from
	"consuming" "consuming"
left join "root" "root" on
	root.id = "consuming"."rootId"
``` 

- Test that "should be able to query consuming entity with node and root entity"
  - `sql` should be:
```sql
select
	"consuming"."id" as "consuming_id",
	"consuming"."rootId" as "consuming_rootId",
	"consuming"."nodeId" as "consuming_nodeId",
	"node"."id" as "node_id",
	"node"."rootId" as "node_rootId",
	"root"."id" as "root_id"
from
	"consuming" "consuming"
left join "node" "node" on
	"node"."id" = "consuming"."nodeId"
	and "node"."rootId" = "consuming"."rootId"
left join "root" "root" on
	root.id = "consuming"."rootId"
```

### Actual Behavior

In the code to reproduce:

- Test that "should be able to query consuming entity with root entity"
  - `sql` is: 
```sql
select
	"consuming"."id" as "consuming_id",
	"consuming"."rootId" as "consuming_rootId",
	"consuming"."nodeId" as "consuming_nodeId",
	"root"."id" as "root_id"
from
	"consuming" "consuming"
left join "root" "root" on
	root.rootId = "consuming"."rootId"
``` 

- Test that "should be able to query consuming entity with node and root entity"
  - `sql` is:
```sql
select
	"consuming"."id" as "consuming_id",
	"consuming"."rootId" as "consuming_rootId",
	"consuming"."nodeId" as "consuming_nodeId",
	"node"."id" as "node_id",
	"node"."rootId" as "node_rootId",
	"root"."id" as "root_id"
from
	"consuming" "consuming"
left join "node" "node" on
	"node"."id" = "consuming"."nodeId"
	and "node"."rootId" = "consuming"."rootId"
left join "root" "root" on
	root.rootId = "consuming"."rootId"
```

### Steps to reproduce

```ts
[...]

PR #11144, opened 2024-11-24
feat: add new foreign key decorator, and entity schemas options

yevhen-komarov:
Closes: #4569

### Description of change

This new feature change the behavior of typeorm to allow the use new foreign key decorator **ForeignKey**, and entity schemas options to create foreign key constraint without relation.

### Pull-Request Checklist


yevhen-komarov:
@michaelbromley @OSA413 What else do I need to do to get PR reviewed and merged?

OSA413:
If I understand correctly, currently we are focusing on stabilizing existing things, later we will look at new features.

First we need to create the new features board to triage all new features, so it will take some time, probably.

src/decorator/ForeignKey.ts
```diff
@@ -0,0 +1,96 @@
+import { ObjectType } from "../common/ObjectType"
+import { getMetadataArgsStorage } from "../globals"
+import { ForeignKeyMetadataArgs } from "../metadata-args/ForeignKeyMetadataArgs"
+import { ForeignKeyOptions } from "./options/ForeignKeyOptions"
+import { ObjectUtils } from "../util/ObjectUtils"
+
+/**
+ * Creates a database foreign key. Can be used on entity property or on entity.
+ */
+export function ForeignKey<T>(
```
OSA413, line 10: Can you please add `ForeignKey` into the shim?
https://github.com/typeorm/typeorm/blob/master/extra/typeorm-model-shim.js

OSA413 (changes requested):

OSA413:
I would also add a test for `repository.find` to see what actually happens when the decorator is used in other use cases.

src/decorator/ForeignKey.ts
yevhen-komarov: Done.

yevhen-komarov:
> I would also add a test for `repository.find` to see what actually happens when the decorator is used in other use cases.

Done.

sgarner:
What will happen when users inevitably misunderstand the purpose of this feature and apply `@ForeignKey` together with `@ManyToOne` etc? Perhaps it should customise the relation's automatically generated foreign key, or produce an error in this case.

I have to say I don't really see the value in this feature. What possible use cases are there for defining a foreign key but not defining the relationship that it represents?

yevhen-komarov:
> What will happen when users inevitably misunderstand the purpose of this feature and apply `@ForeignKey` together with `@ManyToOne` etc? Perhaps it should customise the relation's automatically generated foreign key, or produce an error in this case.

If you do it like this:
```typescript
    @Column()
    @ForeignKey("cities")
    dispatchCityId: number

    @ManyToOne(() => City)
    dispatchCity: City
```

there will be a driver error, something like:
driverError: error: constraint "FK_499c98d1ec465d0b4de2da9744c" for relation "orders" already exists

yevhen-komarov:
> I have to say I don't really see the value in this feature. What possible use cases are there for defining a foreign key but not defining the relationship that it represents?

There really aren't many use cases, but at least there is a issue #4569.

I needed to relate tables with foreign key on two fields, but use a relationship on one unique:
```typescript
@Entity('orders')
export class Order {

  @Column('uuid', { nullable: true })
  settlementRef: string | null;

  @ManyToOne(() => Settlement)
  settlement: Settlement | null;

  @Column('uuid', { nullable: true })
  warehouseRef: string | null;

  @ManyToOne(() => Warehouse)
  warehouse: Warehouse | null;

}

@Entity('orders')
export class OrderFK extends Order {
  @ManyToOne(() => Warehouse, { nullable: true })
  @JoinColumn([
    { name: 'warehouse_ref', referencedColumnName: 'ref' },
    { name: 'settlement_ref', referencedColumnName: 'settlementRef' },
  ])
  warehouseFK: Warehouse | null;
}
```
Why is it so difficult? Two similar connections lead to a messes up the query builder, see #11132

With `@ForeignKey` we can make it easier
```typescript
@Entity('orders')
@ForeignKey(() => Warehouse, ["warehouseRef", "settlementRef"], ["ref", "settlementRef"])
export class Order {

  @Column('uuid', { nullable: true })
  settlementRef: string | null;

  @ManyToOne(() => Settlement)
  settlement: Settlement | null;

  @Column('uuid', { nullable: true })
  warehouseRef: string | null;

  @ManyToOne(() => Warehouse)
  warehouse: Warehouse | null;

}
```

sgarner:
> there will be a driver error, something like:
driverError: error: constraint "FK_499c98d1ec465d0b4de2da9744c" for relation "orders" already exists

Right, of course... I suppose that's sufficient.

> There really aren't many use cases, but at least there is a issue #4569.

Yes I read the linked issue, but it seemed that people were just being a bit precious about defining a relation property that they won't use. The [workaround](https://github.com/typeorm/typeorm/issues/4569#issuecomment-1021160630) of making such properties `private` and non-persisted is a reasonable one IMO.

Your example schema is quite a convoluted structure, but I understand we are forced to work with such database designs sometimes. It does demonstrate a more practical case for this feature.

sgarner (changes requested):
The test doesn't establish whether a foreign key is created correctly. Could we have another test for this? Similar to:
https://github.com/typeorm/typeorm/blob/master/test/functional/database-schema/indices/indices-create-modify.ts

I'm suggesting some doc changes to make clear that this new decorator isn't for everyday use, otherwise LGTM.

Thanks for your contribution 💙

src/decorator/ForeignKey.ts
```diff
@@ -0,0 +1,96 @@
+import { ObjectType } from "../common/ObjectType"
+import { getMetadataArgsStorage } from "../globals"
+import { ForeignKeyMetadataArgs } from "../metadata-args/ForeignKeyMetadataArgs"
+import { ForeignKeyOptions } from "./options/ForeignKeyOptions"
+import { ObjectUtils } from "../util/ObjectUtils"
+
+/**
+ * Creates a database foreign key. Can be used on entity property or on entity.
```
sgarner, line 8:
```suggestion
 * Creates a database foreign key. Can be used on entity property or on entity.
 * Can create foreign key with composite columns when used on entity.
 * Warning! Don't use this with relations; relation decorators create foreign keys automatically.
```

Note I'm suggesting this change once, but you'll need to copy it to each overload signature below.

sgarner (changes requested):

sgarner (approved):
LGTM :tada:

@yevhen-komarov This is great work, thank you again!

sgarner (changes requested):
Coveralls has identified a couple of code paths in `EntityMetaDataBuilder.createForeignKeys()` which aren't tested, can you address?

yevhen-komarov:
🚀Now everything is ready. You can make a review.

sgarner (approved):

yevhen-komarov:
I am implementing it for entity schemas. Please wait a bit.

yevhen-komarov:
I added implementation, documentation and test for entity schemas. Review the changes again.

OSA413 (approved):

test/functional/decorators/foreign-key/wrong city.ts
```diff
@@ -0,0 +1,21 @@
...
+    Column,
+    Entity,
+    ForeignKey,
+    PrimaryColumn,
+    Unique,
+} from "../../../../src"
+
+@Entity("wrong_cities")
+@Unique(["id", "countryCode"])
+export class WrongCity {
```
sgarner, line 11: This file appears to have a space in its name `wrong city.ts` - should use kebab-case

sgarner (approved):
Thanks for picking up on the missing EntitySchema support @yevhen-komarov !

PR #11176, opened 2024-12-11
Update cockroachdb test config from suggested config in docs

stim371:
### Description of change
Updating the test configuration for cockroachdb based on its suggested configuration to speed up local testing.

https://www.cockroachlabs.com/docs/v23.1/local-testing#use-a-local-single-node-cluster-with-in-memory-storage

### Pull-Request Checklist

### DIFF
--- a/test/utils/test-utils.ts
+++ b/test/utils/test-utils.ts
@@ -415,28 +415,40 @@ export async function createTestingConnections(
 
             if (connection.driver.options.type === "cockroachdb") {
                 await queryRunner.query(
-                    `ALTER RANGE default CONFIGURE ZONE USING num_replicas = 1, gc.ttlseconds = 60;`,
+                    `ALTER RANGE default CONFIGURE ZONE USING num_replicas = 1, gc.ttlseconds = 600;`,
                 )
                 await queryRunner.query(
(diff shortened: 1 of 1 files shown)


michaelbromley:
Hmm, still very slow. I mean, it is a pretty crazy difference between 3m for postgres and 61m for cockroach! Do you have any other insight into why there is such a huge difference?

stim371:
> Hmm, still very slow. I mean, it is a pretty crazy difference between 3m for postgres and 61m for cockroach! Do you have any other insight into why there is such a huge difference?

Unfortunately no. This db is not something I've ever worked with. I might try pinging their team or discord to see if they have more ideas, but I'm not sure when I'll have time.

One concern for you might be the amount of GH Actions compute time this uses up. I don't think opensource has unlimited usage, so this will chew credits.

stim371:
I'm going to close this spike for now. I think a larger test runner might be a possibility.

This test runs docker inside the runner, which might cause it to be resource constrained. It uses this setup (instead of the GHA built-in node container with attached db service) because we want to manually set the entrypoint command and a few other flags. Maybe there's some other way to accomplish this.

PR #11177, opened 2024-12-11
Lint on GitHub actions

stim371:
### Description of change
Add linting step to Github Actions test pipeline.

### Pull-Request Checklist


michaelbromley:
This looks good. One thing I'm wondering is whether we can move the node version matrix up to the top level to avoid repeating at every job. This will make it less error-prone when we start to enable more versions.

stim371:
> This looks good. One thing I'm wondering is whether we can move the node version matrix up to the top level to avoid repeating at every job. This will make it less error-prone when we start to enable more versions.

Here's an updated version with the matrix pulled up a level. It requires there to be two sub-files, but it's kind of nice to see that there are two different setups happening.

OSA413:
I would rename `database-tests-oracle-cockroachdb.yml` to something like `database-compose-tests.yml` since it uses docker compose.

nikelborm (changes requested):
Wouldn't it be a good idea to define a matrix with custom parameters for each driver type? I think it will be much more concise.

Nested structures such as `services` which differ across per-driver workflows it seems like could be done without too much effort

https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/running-variations-of-jobs-in-a-workflow#expanding-or-adding-matrix-configurations

stim371:
> Wouldn't it be a good idea to define a matrix with custom parameters for each driver type? I think it will be much more concise.
> 
> Nested structures such as `services` which differ across per-driver workflows it seems like could be done without too much effort
> 
> https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/running-variations-of-jobs-in-a-workflow#expanding-or-adding-matrix-configurations

@nikelborm can you provide a short example?

This PR was originally just for adding linting. We could continue to refactor and improve in follow-on ones further.

michaelbromley:
True, I guess we are mixing concerns here a bit. And now it caused a merge conflict 😅

I suggest let's just get the conflict resolved, merge the linting, and then include any other workflows optimizations in a separate PR

michaelbromley (approved):

PR #11190, opened 2024-12-14
test: redo cockroachdb to service-container GHA and update its version to latest (24)

OSA413:
### Description of change

* This converts the GHA for running cockroachdb from docker-compose based to service-container based.
* This also makes CRDB version to at least 24.2 since this version supports arguments from the env to setup the cluster
* Since the update I needed to update a few tests for CRDB to reflect the following change: https://github.com/cockroachdb/cockroach/issues/97434
* It also needed to fix the transaction lock after it fails
* As a consequence (mostly from the first change), it reduces the test time for CRDB from 50 minutes to 24 minutes.

Needs to be squashed.

### Pull-Request Checklist

### DIFF
--- a/src/driver/cockroachdb/CockroachQueryRunner.ts
+++ b/src/driver/cockroachdb/CockroachQueryRunner.ts
@@ -225,9 +225,10 @@ export class CockroachQueryRunner
         } else {
             this.storeQueries = false
             this.transactionDepth -= 1
+            // This was disabled because it failed tests after update to CRDB 24.2
+            // https://github.com/typeorm/typeorm/pull/11190
+            // await this.query("RELEASE SAVEPOINT cockroach_restart")
             await this.query("COMMIT")
-            // In case of an error it's important to call COMMIT first before executing any other
-            await this.query("RELEASE SAVEPOINT cockroach_restart")
             this.queries = []
             this.isTransactionActive = false
             this.transactionRetries = 0


src/driver/cockroachdb/CockroachQueryRunner.ts
```diff
@@ -225,7 +225,7 @@ export class CockroachQueryRunner
         } else {
             this.storeQueries = false
             this.transactionDepth -= 1
-            await this.query("RELEASE SAVEPOINT cockroach_restart")
+            // await this.query("RELEASE SAVEPOINT cockroach_restart")
```
michaelbromley, line 228: If we have some commented out code, I think we should include some explanation of why it is commented out, or just have it deleted altogether.

OSA413: Added a comment about it and a link to this issue

michaelbromley (approved):

src/driver/cockroachdb/CockroachQueryRunner.ts
alumni: I think this is just hiding the real error:

https://www.cockroachlabs.com/docs/stable/release-savepoint#handling-errors

`commitTransaction` should probably not be called if there's an error. Unless the error happens exactly in it?

test/functional/query-builder/locking/query-builder-locking.ts
```diff
@@ -975,7 +975,10 @@ describe("query builder > locking", () => {
                                 .createQueryBuilder(Post, "post")
                                 .leftJoin("post.author", "user")
                                 .setLock("pessimistic_write")
-                                .getOne(),
+                                .getOne()
+                                .should.be.rejectedWith(
```
alumni, line 979: This seems to be breaking the transaction and making `RELEASE SAVEPOINT` unavailable: https://www.cockroachlabs.com/docs/stable/release-savepoint#handling-errors

Because Chai does not forward the error, `Promise.all` will not throw either, making `EntityManager.transaction` think the operation was successful, so it won't roll back.

We should probably make that function fail.

alumni: Yepp, it's the same thing in the privious test ;) There's this comment: 

https://github.com/typeorm/typeorm/pull/11190/files#diff-0c1916cf8da19e54f4ed20c05df0044acf71abe15bc7db30e4275ba35ec13b17L953

OSA413:
Let's track the fix here
https://github.com/typeorm/typeorm/pull/11264

PR #11192, opened 2024-12-16
docs: update QueryResultCache method descriptions

mguida22:
### Description of change

Correct the description for getFromCache, matching the descriptions found on the [redis](https://github.com/typeorm/typeorm/blob/master/src/cache/RedisQueryResultCache.ts#L125-L130) and [db](https://github.com/typeorm/typeorm/blob/master/src/cache/DbQueryResultCache.ts#L133-L138) cache implementations.

Also correct grammar on the description of the synchronize method.

### Pull-Request Checklist

- [x] ~This pull request links relevant issues as `Fixes #0000`~ N/A
- [x] ~There are new or updated unit tests validating the change~ N/A
- [x] ~Documentation has been updated to reflect this change~ N/A

### DIFF
--- a/src/cache/QueryResultCache.ts
+++ b/src/cache/QueryResultCache.ts
@@ -16,12 +16,14 @@ export interface QueryResultCache {
     disconnect(): Promise<void>
 
     /**
-     * Performs operations needs to be created during schema synchronization.
+     * Perform operations during schema synchronization.
      */
     synchronize(queryRunner?: QueryRunner): Promise<void>
 
     /**
-     * Caches given query result.
+     * Get data from cache.
+     * Returns cache result if found.
+     * Returns undefined if result is not cached.
      */
     getFromCache(
         options: QueryResultCacheOptions,


src/cache/QueryResultCache.ts
```diff
@@ -16,12 +16,14 @@ export interface QueryResultCache {
...
-     * Performs operations needs to be created during schema synchronization.
+     * Perform operations during schema synchronization.
      */
     synchronize(queryRunner?: QueryRunner): Promise<void>
 
     /**
-     * Caches given query result.
+     * Get data from cache.
+     * Returns cache result if found.
+     * Returns undefined if result is not cached.
```
albasyir, line 26: i think we can follow this

- https://jsdoc.app/tags-description
- https://jsdoc.app/tags-returns

src/cache/QueryResultCache.ts
```diff
@@ -16,12 +16,14 @@ export interface QueryResultCache {
     disconnect(): Promise<void>
 
     /**
-     * Performs operations needs to be created during schema synchronization.
+     * Perform operations during schema synchronization.
```
albasyir, line 19: i think we can follow this:

- https://jsdoc.app/tags-description

albasyir (commented):
good catch!, i think we make it better when we follow jsdoc standard!

src/cache/QueryResultCache.ts
alumni: FYI, in TypeScript, we have [TSDoc](https://tsdoc.org/), which is supported by `tsc` and is very similar to JSDoc. Description does not exist, but there's a [@returns](https://tsdoc.org/pages/tags/returns/) tag indeed.

If need be, I can add `eslint-plugin-tsdoc` to the eslint config to prevent writing incorrect TSDoc.

mguida22:
@albasyir @alumni agreed that we could benefit from using JS/TSDoc style comments overall, but we [don't really use them today](https://github.com/search?q=repo%3Atypeorm%2Ftypeorm%20%40return&type=code). IMO we should track the doc comments idea and implementation on it's own.

 I'd like to keep this change to a minimum by just correcting the content of the comments for now.

michaelbromley (approved):

PR #11197, opened 2024-12-19
feat: QueryRunner context handler

klingenm:
### Description of change

Introduce a centralized method for QueryRunner lifecycle handling.

#### The problem
The pattern

```typescript
const usedQueryRunner = queryRunner ?? this.connection.createQueryRunner()

...
try {
 // do a bunch of stuff
} finally {
   if (usedQueryRunner !== queryRunner)
      await usedQueryRunner.release()
}
```

occurs many times in the codebase. In some places a common method has been created, which seems to have blurred the responsibility a bit and release is not called if created.

At least here; if called without queryRunner, getFromCache will leak created queryRunner.
- https://github.com/typeorm/typeorm/blob/19a6954d8b80ef23f0c3fb759c470c6a946d41a2/src/cache/DbQueryResultCache.ts#L142

#### The proposed solution

Lending from python's `with` context manager, this PR introduces a `DataSource.runWithQueryRunner` method.

The method `runWithQueryRunner` comes in three main forms;
```typescript
    async runWithQueryRunner<T>(fn: (queryRunner: QueryRunner) => Promise<T>): Promise<T>;
    async runWithQueryRunner<T>(fn: (queryRunner: QueryRunner, release: () => Promise<void>) => Promise<T>): Promise<T>;
   
    async runWithQueryRunner<T>(existingQueryRunner: QueryRunner | undefined, fn: (queryRunner: QueryRunner) => Promise<T>): Promise<T>;
    async runWithQueryRunner<T>(existingQueryRunner: QueryRunner | undefined, fn: (queryRunner: QueryRunner, release: () => Promise<void>) => Promise<T>): Promise<T>;
   
    async runWithQueryRunner<T>(existingQueryRunner: QueryRunner | undefined, mode: ReplicationMode, fn: (queryRunner: QueryRunner) => Promise<T>): Promise<T>;
    async runWithQueryRunner<T>(existingQueryRunner: QueryRunner | undefined, mode: ReplicationMode, fn: (queryRunner: QueryRunner, release: () => Promise<void>) => Promise<T>): Promise<T>;
```

They are quite self-explanatory, except for the release function. When `fn` is passed with two parameters, the responsibility for calling release is assumed by the caller.

#### Base for future stuff

Having centralized management of QueryRunner creation and deletion means we can also do cool stuff, like make it possible to run transactions with the QueryRunner provided by AsyncLocalStorage to allow global EntityManager with it's repositories and custom repositories be used in transaction. This PR is really split out from the AsyncLocalStorage transaction branch PR, to reduce the scope. I will link that PR here when it has been posted.

### Pull-Request Checklist

- [x] `npm run test` passes with this change (sqlite test fail, same as on master)
- [ ] ~~This pull request links relevant issues as `Fixes #0000`~~ N/A
- [ ] ~~Documentation has been updated to reflect this change~~ N/A


src/data-source/DataSource.ts
```diff
@@ -760,4 +756,129 @@ export class DataSource {
...
+
+    /**
+     * Run fn with a temporary query runner.
+     *
+     * Query runner is automatically released after fn is done.
+     *
+     * @param fn
+     */
+    async runWithQueryRunner<T>(
+        fn: (queryRunner: QueryRunner) => Promise<T>,
```
hinogi, line 768: What is fn?

klingenm: https://github.com/typeorm/typeorm/pull/11197/files#diff-bc29c8a16d80c175364ccb4e84a5247cc0cf5ab6d2aff241a8a871084a089300 I added it to the docs. Internally the overloaded version that takes the optional first `queryRunner` argument is more useful, e.g:

https://github.com/typeorm/typeorm/pull/11197/files#diff-06402bac7d3c217ce8afc1071fffcd25d026b57b4f0c19593615be01539359b4R147-R164

hinogi: is there maybe a better suited name instead of `fn` that provides more intent?

alumni:
We already implemented explicit resource management (the `using` keyword) for `QueryRunner`, which is JS's equivalent of `with` in Python. This will be released in v1.

So I think we don't need to add this extra function if there's a built-in JS feature for it.

BTW, in our project at work we've been using for quite a long time a similar solution to what I see in this PR, except we implemented it in our project. We're actually waiting for v1 to be released to remove it.

klingenm:
@alumni Yeah, I saw that when I rebased the PRs (on your request...). 

The main idea was not to provide the `runWithQueryRunner` to the end user, but to handle the `QueryRunner` creation in one place with automatic cleanup. In this case explicit resource management does not work, as we might get a resource passed in, which we can not dispose of.

I agree that `runWithQueryRunner` and explicit resource management overlap as it comes to end users, so maybe we remove it from public api?

But don't you think it's better to have a centralised way to create/dispose of query runner instances internally in the library? Is it better to require every call site to repeat the
```typescript
const usedQueryRunner = queryRunner || this.createQueryRunner()
try {
    // do a lot of stuff...
    return await usedQueryRunner.query(query, parameters) 
} finally {
    if (!queryRunner) await usedQueryRunner.release()
}
```
pattern where forgetting one await leads to resource leaks?

alumni:
I will close this PR since we decided not to merge it. Please use explicit resource management instead in the upcoming version.

PR #11234, opened 2025-01-11
Chore: Added logging to the Entity Listener Metadata

JackNytely:
### Description of change
Users were having issues when incorrect setup of an entity causing the following error to be thrown:
image

However, this is not descriptive and when working with 20+ entities, can make it really difficult to determine the cause of the issue for the User.

As such, my change will focus on this part of the code located at `src/metadata/EntityListenerMetadata.ts`:

And we will add a check to see if the Entity Method exists, and another check to see if that Method is a Function and throw a more descriptive error:

### Pull-Request Checklist

### Steps to Reproduce:
We had removed the Following in our Parent.entity.ts file:
0a7e8e2b-a78c-41ad-98ca-afe66d87276c-1

However, We had the following in our Child.entity.ts file:
d4bf882e-a1a3-42f3-b4db-9d771033ce49

And that ended up throwing this error:
image

Solution was to remove the @AfterLoad method from the Child.entity.ts


michaelbromley:
Thanks! Please include at least minimal reproduction steps so that a reviewer would be able to reproduce the behavior locally, and therefore verify your fix.

JackNytely:
> Thanks! Please include at least minimal reproduction steps so that a reviewer would be able to reproduce the behavior locally, and therefore verify your fix.

I have updated the Original PR with the steps to reproduce

gioboa:
> I have updated the Original PR with the steps to reproduce

Thanks, I'll look at it

gioboa (approved):
Thanks @JackNytely 
This is a nice DX improvement because we are adding more context for this error

michaelbromley (approved):

src/metadata/EntityListenerMetadata.ts
```diff
@@ -75,8 +75,30 @@ export class EntityListenerMetadata {
...
+            // Check if the Entity Method is not a function
+            if (typeof entityMethod !== "function")
+                throw new Error(
+                    `Entity listener method "${this.propertyName}" in entity "${
+                        entity.constructor.name
+                    }" must be a function but got "${typeof entityMethod}".`,
+                )
+
+            // Return the Entity Method
+            return entityMethod()
```
alumni, line 98: The method is unbound

JackNytely: I only added logging and error checks, the functionality is still the exact same, so I am not too sure why it errors out myself

alumni: The functionality was changed by separating the function from the object in line 81. Now the function is no longer bound to that object.

Basically `entityMethod` no longer has any information about `entity` (`this` within the method's implementation will no longer point to `entity`, it will simply not be defined).

albasyir (approved):

naorpeled (approved):
Overall LGTM,
personally I think that the comments are redundant so maybe it's worth deleting them 🙏

alumni:
The issue needs to be fixed, see the code review comments.

src/metadata/EntityListenerMetadata.ts
JackNytely: Thanks again for the feedback, I have decided to call the entityMethod using your recommendation
`entityMethod.call(entity)`

mguida22:
It looks like tests didn't run against the latest changes here. Usually I can trigger a re-run but I don't see any options on this PR. @michaelbromley do you have a way to trigger them?

alumni:
@mguida22 the tests will still fail because the method is unbound.

alumni (approved):

alumni:
@mguida22 the workflow is only automatic (cannot be triggered manually) - a merge commit would do :)

src/metadata/EntityListenerMetadata.ts
```diff
@@ -75,8 +75,30 @@ export class EntityListenerMetadata {
...
+
+            // Check if the Entity Method is not a function
+            if (typeof entityMethod !== "function")
+                throw new Error(
+                    `Entity listener method "${this.propertyName}" in entity "${
+                        entity.constructor.name
+                    }" must be a function but got "${typeof entityMethod}".`,
+                )
+
+            // Return the Entity Method
```
mguida22, line 97:
```suggestion
            // Call and return the Entity Method
```

Issue #11243, opened 2025-01-16
[sqlite] Implement streaming queries


douglascayers:
### Feature Description

For SQLite, I'd like to stream back rows from a query rather than retrieve all of them at once. This feature is implemented for other databases, and I've used it successfully with Postgres.

Currently, if you try to stream a query with SQLite, TypeORM throws the error `Stream is not supported by sqlite driver.`

https://github.com/typeorm/typeorm/blob/master/src/driver/sqlite-abstract/AbstractSqliteQueryRunner.ts#L170-L177

I reviewed the git history of the `AbstractSqliteQueryRunner.ts` file, and it seems that this method was stubbed out ~8 years ago and possibly never revisted. I searched the issues to see if this had been addressed before, but I did not find any related issues or PRs.

### The Solution

From what I can tell from the other `*QueryRunner.ts` files, I believe we need to use SQLite's [Statement#get](https://github.com/TryGhost/node-sqlite3/wiki/API#getparam---callback) method to retrieve one row at a time, and stream them back via a `Readable` implementation.

For inspiration, I found https://gist.github.com/rmela/a3bed669ad6194fb2d9670789541b0c7 which provides an example of wrapping SQLite's `Statement#get` method in a `Readable`. Locally, I've reproduced a proof of concept and feel (naively?) confident moving forward to try and implement it in TypeORM.

### Considered Alternatives

The only alternative I'm aware of is to perform a regular query, but that would return all the rows at once into memory, which is what I want to avoid.

### Relevant Database Driver(s)

- [ ] aurora-mysql
- [ ] aurora-postgres
- [ ] better-sqlite3
- [ ] cockroachdb
- [ ] cordova
- [ ] expo
- [ ] mongodb
- [ ] mysql
- [ ] nativescript
- [ ] oracle
- [ ] postgres
- [ ] react-native
- [ ] sap
- [ ] spanner
- [X] sqlite
- [ ] sqlite-abstract
- [ ] sqljs
- [ ] sqlserver

### Are you willing to resolve this issue by submitting a Pull Request?

Yes, I have the time, and I know how to start.

PR #11245, opened 2025-01-16
feat: implement query stream for sqlite3 driver

douglascayers:
Introduces `SqliteQueryStream`, a readable wrapper over a `sqlite3` prepared statement.

Closes: #11243

### Description of change

Implements `SqliteQueryRunner#stream` to return a readable stream of query rows from a prepared statement. Uses the new `SqliteQueryStream` class to adapt the sqlite3 `Statement#get` API to `stream.Readable`.

Since a test suite already existed for streaming, the test change was to add the `sqlite3` driver to the tested data sources.

### Pull-Request Checklist

- [ ] Documentation has been updated to reflect this change "N/A"

### DIFF
--- a/src/driver/sqlite/SqliteQueryRunner.ts
+++ b/src/driver/sqlite/SqliteQueryRunner.ts
@@ -1,8 +1,11 @@
+import type { Database } from "sqlite3"
 import { QueryRunnerAlreadyReleasedError } from "../../error/QueryRunnerAlreadyReleasedError"
 import { QueryFailedError } from "../../error/QueryFailedError"
 import { AbstractSqliteQueryRunner } from "../sqlite-abstract/AbstractSqliteQueryRunner"
 import { SqliteConnectionOptions } from "./SqliteConnectionOptions"
 import { SqliteDriver } from "./SqliteDriver"
+import { SqliteQueryStream } from "./SqliteQueryStream"
+import { ReadStream } from "../../platform/PlatformTools"
 import { Broadcaster } from "../../subscriber/Broadcaster"
 import { ConnectionIsNotSetError } from "../../error/ConnectionIsNotSetError"
 import { QueryResult } from "../../query-runner/QueryResult"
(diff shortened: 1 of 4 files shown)


src/driver/sqlite/SqliteQueryRunner.ts
```diff
@@ -175,4 +178,49 @@ export class SqliteQueryRunner extends AbstractSqliteQueryRunner {
...
+            try {
+                const databaseConnection: Database = await this.connect()
+                this.driver.connection.logger.logQuery(query, parameters, this)
+                const statement = databaseConnection.prepare(query, parameters)
+                try {
+                    const stream = new SqliteQueryStream(statement)
+                    stream.on("error", logQueryError)
+                    if (onEnd) stream.on("end", onEnd)
+                    if (onError) stream.on("error", onError)
+                    ok(stream as any)
```
alumni, line 212: Maybe we should avoid the `as any` cast.

Looks like `SqliteQueryStream extends Readable` is not assignable to `ReadStream extends Readable`. Maybe `SqliteQueryStream` should extend `ReadStream`?

douglascayers: 👋 Thanks @alumni for the feedback. I actually think that TypeORM's expectation that `stream()` returns a `fs.ReadStream` may be a mistake and actually should be `stream.Readable`. I'll explain my findings below.

## Observation 1: ReadStream is for files

The [fs.ReadStream](https://nodejs.org/docs/latest-v20.x/api/fs.html#class-fsreadstream) class must be constructed with a file path (e.g. `new fs.ReadStream(path[, options])` or [`fs.createReadStream(path[, options])`](https://nodejs.org/docs/latest-v20.x/api/fs.html#fscreatereadstreampath-options)). But with these database query streams, there isn't a physical file being read but rather data streamed from a connection.

Subclasses could offer a dummy path like `/dev/null` but that might only work on MacOS and Linux operating systems. I'm not sure that trick would work on Windows.

I think the correct typing should be `stream.Readable`, which is a more generic readable over any data source.

## Observation 2: The read stream for other drivers is `Readable`

The postgres driver uses the [pg-query-stream](https://www.npmjs.com/package/pg-query-stream) library for its streaming implementation, which returns a class that extends `Readable`.

https://github.com/brianc/node-postgres/blob/master/packages/pg-query-stream/src/index.ts#L12

image

The mysql driver's implementation also returns a class that extends `Readable`.

image

## Observation 3: The use of `any` hides the types from linter

The `QueryRunner#connect` method returns an `any`. This means that any methods called on that returned value, in turn, are also type `any`. This cascade effect hides the true type of the variables and prevents the linter or compiler from detecting discrepancies as you spotted in this pull request.

image

image

image

In the case of this pull request, the `SqliteQueryStream` class extends `Readable` and the type was known in the code rather than inheriting from `any`. So in this case the type checker identified the conflict with the expected type of `ReadStream`. But because of the above findings, I figured casting to `any` was the simplest and congruent solution given the existing code I saw in the other drivers.

I'm happy to propose in this or another PR that the `QueryRunner#stream` be typed to return `Readable` if you think that'd help.

Thanks!

alumni: You're right, `Readable` would be more appropriate for the return type. Not sure if we should make any changes here though.

alumni (approved):

PR #11248, opened 2025-01-16
fix: change import for process dependency 

yohannpoli:
Using Deno, when i run the migration generate command, i got no error, when i use the migration run command, it fail because process.cwd is undefined, this is due to a bad import of the process dependency.

### Description of change

I use the same syntax from the migration generate command for the process dependency (tested in local by patching the package) and it works after this change.

### Pull-Request Checklist

## Summary by CodeRabbit

* **Refactor**
  * Standardized the import style for the Node.js process module to use a default import, aligning with project conventions. No changes to behavior, outputs, or control flow.
* **Chores**
  * Minor code cleanup related to import style. No user-facing changes.
  * No configuration updates required.
  * Fully compatible with existing workflows and commands.

### DIFF
--- a/.github/workflows/mutation-test.yml
+++ b/.github/workflows/mutation-test.yml
@@ -1,0 +1,62 @@
+name: mutation-testing
+
+on:
+  workflow_dispatch:
+  schedule:
+    - cron: "0 0 1 * *"
+
+jobs:
+  mutation-testing:
+    runs-on: ubuntu-latest
+    
+    services:
+      mysql:
+        image: mysql:5.7.37
+        ports:
+          - "3306:3306"
+        env:
+          MYSQL_ROOT_PASSWORD: "admin"
+          MYSQL_USER: "test"
+          MYSQL_PASSWORD: "test"
+          MYSQL_DATABASE: "test"
+      mariadb:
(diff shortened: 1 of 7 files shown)


gioboa (commented):
Thanks @yohannpoli for your help
Did you try this change with npm too?

yohannpoli:
@gioboa I don't have a project setup with TypeORM and npm right now. If you check other command, this is the same import type, and it's seem working with npm. If you have a ready environment, it would be nice if you can run the command for me, or i will test it this week end, i can't before.

yohannpoli:
@gioboa How do you want me to try the fix, i need to make a release i think, no ? if i use the patch package dependency, it works, but this is not the same as building the library and test it

gioboa (commented):
@yohannpoli Thanks for your commitment.
I think we can test this change during the review process, so I added this PR to our release board.

alumni (approved):

alumni:
It would be interesting to check the build, I would expect the code to transpiled the same way.

I think we have synthetic default imports enabled in tsconfig, so I actually tend to prefer `import x` instead of `import * as x`, but they should be transpiled identically.

gioboa (approved):

PR #11253, opened 2025-01-22
Mutation testing

OSA413:
### Description of change

Adds mutations testing, the thing that shows the quality of our tests.

At first it needs to be checked that mutation testing can be run with the following configuration (inPlace: true).

Then tests should be fixed for Windows.

Then it can be merged.

### Pull-Request Checklist

### DIFF
--- a/.mocharc.js
+++ b/.mocharc.js
@@ -1,0 +1,13 @@
+// https://github.com/mochajs/mocha/blob/main/example/config/.mocharc.js
+
+module.exports = {
+    bail: true,
+    'check-leaks': true,
+    color: true,
+    file: ['./build/compiled/test/utils/test-setup.js'],
+    spec: ["./build/compiled/test"],
+    
+    // jobs: 1,
+    // parallel: false,
+    recursive: true,
+};
--- a/.mocharc.js
+++ b/.mocharc.js
@@ -6,7 +6,7 @@ module.exports = {
     color: true,
     file: ['./build/compiled/test/utils/test-setup.js'],
     spec: ["./build/compiled/test"],
-    
+    timeout: "90000",
     // jobs: 1,
     // parallel: false,
     recursive: true,
--- a/package-lock.json
+++ b/package-lock.json
@@ -30,6 +30,8 @@
       },
       "devDependencies": {
         "@eslint/js": "^9.17.0",
+        "@stryker-mutator/core": "^8.7.1",
+        "@stryker-mutator/mocha-runner": "^8.7.1",
         "@tsconfig/node16": "^16.1.1",
         "@types/chai": "^4.3.4",
         "@types/chai-as-promised": "^7.1.5",
@@ -541,6 +543,18 @@
         "@jridgewell/sourcemap-codec": "^1.4.14"
       }
     },
+    "node_modules/@babel/helper-annotate-as-pure": {
+      "version": "7.25.9",
+      "resolved": "https://registry.npmjs.org/@babel/helper-annotate-as-pure/-/helper-annotate-as-pure-7.25.9.tgz",
+      "integrity": "sha512-gv7320KBUFJz1RnylIg5WWYPRXKZ884AGkYpgpWW02TH66Dl+HaC1t1CKd0z3R4b6hdYEcmrNZHUmfCP+1u3/g==",
+      "dev": true,
+      "dependencies": {
+        "@babel/types": "^7.25.9"
+      },
+      "engines": {
+        "node": ">=6.9.0"
+      }
+    },
     "node_modules/@babel/helper-compilation-targets": {
       "version": "7.25.9",
       "resolved": "https://registry.npmjs.org/@babel/helper-compilation-targets/-/helper-compilation-targets-7.25.9.tgz",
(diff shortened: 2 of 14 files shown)


package.json
```diff
@@ -102,6 +102,8 @@
     "yargs": "^17.7.2"
   },
   "devDependencies": {
+    "@stryker-mutator/core": "^8.7.1",
```
alumni, line 105: Could you run `npx sort-package-json` to format this file?

.mocharc.js
alumni, line 1: Does this mean we could simply run `mocha` without any arguments now?

If yes, then in `package.json` we could simply do:
```json
{
  "scripts": {
    "test": "npm run compile && npm run test-fast",
    "test-fast": "mocha --bail",
  }
}
```

BTW, I was also considering to move the `--bail` option to CI only, but we could do that separately.

.github/workflows/test/for-mutation-testing.ormconfig.json
alumni, line 1: While adding tests for SAP, I was thinking to create a single `ormconfig.json` (and a single `docker-compose.yml`) in the project root. Maybe for individual tests we could use `jq` to pick the entry that we need and we could run `docker compose up` for a single service 🤷🏻‍♂️ 

Would be easier then to keep track of the setup that is being used (I was recently wondering why my connection is failing, but I had a different setup and I didn't notice)

alumni (approved):

.mocharc.js
OSA413: Yeah, it should, just added it (I remember adding it locally).

> --bail
I don't think it's useful to fail at the first test, removing this option would provide the whole picture of what defects were introduced during the latest commits.

package.json
OSA413: Done

.github/workflows/test/for-mutation-testing.ormconfig.json
OSA413: 🤷

naorpeled (approved):

sgarner (approved):

alumni (approved):

stryker.config.json
```diff
@@ -0,0 +1,21 @@
...
+  "packageManager": "npm",
+  "reporters": [
+    "html",
+    "dashboard",
+    "progress",
+    "json"
+  ],
+  "testRunner": "mocha",
+  "mochaOptions": {
+    "spec": [ "./build/compiled/test/**/*.js" ]
```
alumni: shouldn't we have the full mocha options here?

alumni (approved):

package-lock.json
alumni, line 1: Seems to have a lot of differences, not sure what's going on (maybe EOL changes?)

if you do `git checkout upstream/master -- package-lock.json && npm install` would anything change?

OSA413: > would anything change?

It says "No content changes found", adding the same file and then doing `git status` shows nothing.

OSA413: My node version is v20.10.0

stryker.config.json
OSA413: Docs say that "feel free to leave out the mochaOptions entirely if you're using one of the [default file locations](https://mochajs.org/#configuring-mocha-nodejs)." maybe a year ago we didn't use that.

OSA413: I can remove it and see if it works

package-lock.json
alumni: The node version is not that relevant, but each version of node comes with a version of npm bundled.

The only difference I have with `npm@10.8.2` (bundled with `node@20.19.6`) is that licenses are added, but that doesn't matter :) 

We need to find a way fix the package manager version so we don't have differences between us, but probably in the next branch - unfortunately `npm` requires extra setup steps with `corepack`.

Probably some EOL changes, we might need to create a `.gitattributes` file and setup `eol=lf` for text files 🤷🏻‍♂️ Anyway, doesn't matter for now.

stryker.config.json
alumni: Yeah, we had them in `package.json` before. If we define it, will it pick the rest of the settings from the default location?

_Later edit:_ NVM, saw how they look like here: https://stryker-mutator.io/docs/stryker-js/mocha-runner/

OSA413:
Since mutation testing is a resource and time consuming, what DBs should be included? Currently I'm trying to enable everything except oracle (because it requires an extra setup, can be brought later) and cockroach (because it runs too long). The limit is 6 hours for GHA.

alumni:
I would at least keep the latest Postgres (has the most tests) and maybe the latest MariaDB or MySQL and one of the SQLite tests.

Would removing the CLI test (it's a GitHub issue test) from mutation testing at least be helpful? It takes quite a long time.

OSA413:
> Would removing the CLI test (it's a github issue test) from mutation testing at least be helpful?

It runs it only when necessary - if mutant is not covered by the test, the test is not run.

package.json
```diff
@@ -112,6 +112,8 @@
     "@eslint/js": "^9.39.1",
     "@google-cloud/spanner": "^8.3.1",
     "@sap/hana-client": "^2.26.26",
+    "@stryker-mutator/core": "^8.7.1",
```
alumni: I think we can use `^9.4.0` since we already dropped node 18. It would also remove a lot of duplicate packages from the lockfile too ;)

If you switch to the new version: `git fetch upstream && git checkout upstream/master -- pnpm-lock.yaml && pnpm install` to avoid any leftovers in the lockfile (provided that `upstream` is `typeorm/typeorm`).

OSA413: I've updated the lock file, check again please

stryker.config.json
```diff
@@ -0,0 +1,13 @@
+{
+  "$schema": "./node_modules/@stryker-mutator/core/schema/stryker-schema.json",
+  "packageManager": "npm",
```
alumni: We switched to `pnpm`, can we update this and the workflow too? :)

.github/workflows/mutation-test.yml
```diff
@@ -0,0 +1,62 @@
...
+    steps:
+      - uses: actions/checkout@v4
+      - uses: actions/setup-node@v4
+        with:
+          node-version: 20
+          cache: "npm"
+
+      - run: npm ci
+      - run: cp .github/workflows/test/mutation-testing.ormconfig.json ormconfig.json
+      - run: npx stryker run
```
alumni: `pnpm exec stryker run` (`pnpx` will download the latest `stryker` version from npm instead of using the one in the repo)

OSA413: done

stryker.config.json
OSA413: done

.github/workflows/mutation-test.yml
```diff
@@ -0,0 +1,61 @@
...
+          POSTGRES_DB: postgres
+        options: >-
+          --health-cmd pg_isready
+          --health-interval 10s
+          --health-timeout 5s
+          --health-retries 5
+
+    steps:
+      - uses: actions/checkout@v4
+      - uses: actions/setup-node@v6
```
alumni, line 48: You need to setup pnpm before setting up node. Check the other workflows.

stryker.config.json
```diff
@@ -0,0 +1,13 @@
+{
+  "$schema": "./node_modules/@stryker-mutator/core/schema/stryker-schema.json",
+  "packageManager": "pnpm",
+  "buildCommand": "npm run compile",
```
alumni: `pnpm run compile`

.github/workflows/mutation-test.yml
OSA413: done

stryker.config.json
OSA413: done

PR #11257, opened 2025-01-23
test: fix and run tests on Windows

OSA413:
### Description of change

Basically, this makes tests run on Windows (some tests were overwriting the process.platform constant and that broke some tests on Windows, now it doesn't).

I should also look at the reason why I can't run CLI tests on my local machine.

I also think that the tests that rely on the different platform should accept the `platform` as a variable so we can test all the scenarios.

### Pull-Request Checklist


OSA413:
Does the broken `produces deterministic, unique, and valid table names for relative paths; leaves absolute paths unchanged` test mean that it actually only worked on `win32` and `darwin` platforms and not linux?

test/functional/multi-database/multi-database-basic-functionality/multi-database-basic-functionality.ts
```diff
@@ -20,28 +20,12 @@ const VALID_NAME_REGEX = /^(?!sqlite_).{1,63}$/
...
-
-            afterEach(() => {
-                Object.defineProperty(process, `platform`, {
-                    configurable: true,
-                    value: realPlatform,
-                })
-            })
-
-            it(`produces deterministic, unique, and valid table names for relative paths; leaves absolute paths unchanged (${platform})`, () => {
+        if (process.platform === "win32") {
```
alumni, line 23: Early return would simplify this (especially since the max line length is 80 at the moment).

test/functional/multi-database/multi-database-basic-functionality/multi-database-basic-functionality.ts
```diff
@@ -91,7 +75,7 @@ describe("multi-database > basic-functionality", () => {
         beforeEach(() => reloadTestingDatabases(connections))
         after(async () => {
             await closeTestingConnections(connections)
-            await rimraf(`${tempPath}/**/*.attach.db`)
+            await rimraf(`${tempPath}/**/*.attach.db`, {glob: true})
```
alumni, line 78: file can be formatted

sgarner (commented):
> some tests were overwriting the process.platform constant and that broke some tests on Windows, now it doesn't

Can you elaborate on what was breaking for you before?

These tests are only exercising some functions that manipulate strings, so mocking the `process.platform` is sufficient to test their behavior. 

With the proposed change, only the behavior on the actual platform of the test runner will be exercised. This means we'd need to run tests on additional Windows runners to maintain coverage. I doubt we have enough platform-specific differences in functionality across TypeORM to warrant doing that.

> Does the broken `produces deterministic, unique, and valid table names for relative paths; leaves absolute paths unchanged` test mean that it actually only worked on `win32` and `darwin` platforms and not linux?

Since `darwin` and `linux` both use POSIX paths, the expected output would be identical, so that shouldn't be an issue.

OSA413:
Added tests to run on Windows, will look at this PR later.

OSA413:
> Can you elaborate on what was breaking for you before?

It was breaking for:
1. `1) "after all" hook for "should create foreign keys for relations within the same database"` - `Error: Illegal characters in path.` - this is fixed by adding `{grep: true}`
2. `1) should save an entity` - `EntityMetadataNotFoundError: No metadata for "Post" was found.` - this is fixed by removing the platform overwrite in the tests.
3. `cli init command` - `Error: ENOENT: no such file or directory, stat 'C:\home\osa413\GitHub\typeorm\1741795310054TestProject'` - this is fixed by adding `node` command at the beginning of the process creation in the cli test

test/functional/multi-database/multi-database-basic-functionality/multi-database-basic-functionality.ts
OSA413: Done

test/functional/multi-database/multi-database-basic-functionality/multi-database-basic-functionality.ts
OSA413: I've run the `npm run format` command but it didn't change the file

sgarner:
> `"after all" hook for "should create foreign keys for relations within the same database"` - `Error: Illegal characters in path.` - this is fixed by adding `{grep: true}`

:+1:

> `cli init command` - `Error: ENOENT: no such file or directory, stat 'C:\home\osa413\GitHub\typeorm\1741795310054TestProject'` - this is fixed by adding `node` command at the beginning of the process creation in the cli test

:+1:

> 2. `should save an entity` - `EntityMetadataNotFoundError: No metadata for "Post" was found.` - this is fixed by removing the platform overwrite in the tests.

I guess this is because some tests are running in parallel and picking up the wrong platform from the overwritten `process` global. Or else why does this interfere with entity metadata? Which database driver(s) does this affect? Perhaps the tests could be written differently to avoid this issue while still mocking platforms.

michaelbromley (approved):

alumni (approved):

test/functional/multi-database/multi-database-basic-functionality/multi-database-basic-functionality.ts
```diff
@@ -20,46 +20,29 @@ const VALID_NAME_REGEX = /^(?!sqlite_).{1,63}$/
...
-                    const otherOsRes = filepathToName(otherOs)
-                    expect(winOsRes).to.equal(otherOsRes)
-                    expect(winOsRes).to.match(
-                        VALID_NAME_REGEX,
-                        `'${winOs}' is invalid table name`,
-                    )
-                }
-            })
-        }
+        if (process.platform !== "win32") return
```
sgarner, line 23: These tests shouldn't be skipped on non-windows platforms. On darwin/linux we still want to assert that relative paths are normalized when used as table names.

sgarner (changes requested):
Adding a windows test runner feels like overkill to me, since the only platform specific code in TypeORM are these path functions. But I suppose there's no harm in it.

However I do favour continuing to mock the platform in these tests so that anyone editing these functions can test them on their local machine without needing access to a win32 environment.

I found the problem with the original code and propose a solution in https://github.com/OSA413/typeorm/pull/1

OSA413:
Btw, we discussed on the meeting that it would be nice to have Windows test that run SQLite because it's the only platform dependency that may appear (like writing DB to the disk). I'll add other SQLite tests for Windows later.

.github/workflows/windows-database-tests.yml
```diff
@@ -0,0 +1,18 @@
+name: database-tests
+
+on: workflow_call
+
+jobs:
+  sqljs:
```
sgarner, line 6: Is there any value in testing sqljs on windows? As I understand it sqljs runs in the browser so should be platform agnostic.

As you mentioned, SQLite is the one database that could benefit from end-to-end testing in a win32 environment, so is what I'd expect to see here.

Fine if you intend doing that in another PR though

sgarner (approved):

.github/workflows/windows-database-tests.yml
OSA413: I've added tests for `better-sqlite3` and `sqlite` to run on Windows

PR #11258, opened 2025-01-24
fix: manage CURRENT_TIMESTAMP(6) correctly

law108000:
### Description of change
just fix the issue with CURRENT_TIMESTAMP(6) being used, can "DEFAULT" be used?

while I use the code like this.
```javascript
await this.parentRepository
        .createQueryBuilder()
        .relation("relation")
        .of(parentRecord)
        .add(childRecord);
```
The generated SQL is always hardcoded to set the updated_date to be CURRENT_TIMESTAMP.
```sql
UPDATE `child` SET `parentId` = ?, `updated_date` = CURRENT_TIMESTAMP WHERE `id` IN (?)
```
After investigating all related source code, I found the below TODO in UpdateQueryBuilder.ts about 2 years ago:
```
todo: fix issue with CURRENT_TIMESTAMP(6) being used, can "DEFAULT" be used?!
```

### Pull-Request Checklist

## Summary by CodeRabbit

* **New Features**
  * Introduced comprehensive test suites to verify correct handling of timestamp and datetime precision (0 and 6) in entity updates and relations across multiple database drivers (MySQL, MariaDB, Postgres, Oracle, SQL Server, SQLite).
  * Added new entity classes and base classes for each supported precision and database, supporting automated timestamp updates with appropriate precision.

* **Tests**
  * Added extensive tests to ensure update operations use the correct timestamp/datetime precision and SQL functions for each database.
  * Verified schema creation, SQL generation, and update behavior for both precision 0 and 6 across all supported drivers.

* **Refactor**
  * Improved maintainability by centralizing column assignment logic for update queries, reducing code duplication.


law108000:
> Not sure what issue we are trying to fix here.
> 
> Default is the column default value (in the column definition in the create table statement), the one that is used when a row is inserted, not the default update value.

Oh, sorry for the incomplete story that confused you.
In my use case, I want high-precision create and update dates, so i config to use `CURRENT_TIMESTAMP(6)` instead of the defaulted `CURRENT_TIMESTAMP`.

my configuration is like the below:
```typescript
class MyEntity {
  @CreateDateColumn({ type: 'timestamp', default: () => 'CURRENT_TIMESTAMP(6)', transformer: new DateTransformer() })
  @IsOptional()
  @IsISO8601()
  create_date?: Date;

  @UpdateDateColumn({ type: 'timestamp', default: () => 'CURRENT_TIMESTAMP(6)', transformer: new DateTransformer() })
  @IsOptional()
  @IsISO8601()
  updated_date?: Date;
  
  // others attrs...
}
```

I thought the TypeORM would be able to handle it well, but it failed on my jest test code.
for example, the correct updated_date should be `2023-05-23 02:53:44.143000`, but typeorm always stored it as `2023-05-23 02:53:44.000000`.

after digging into the TypeORM source code, I found it was always hardcoded to `" = CURRENT_TIMESTAMP"` no matter what I set on `CreateDateColumn`, and there left a TODO comment.

alumni:
Okay, I get it now :) I think it makes sense.

I'm a bit surprised that `CURRENT_TIMESTAMP` was hard-coded in that place. Normally, this is driver dependent - `Driver.mappedDataTypes.updateDateDefault` has the function the gets the current time. So `updateDateColumn.default` should always be set to this value:
https://github.com/typeorm/typeorm/blob/27b4207c48a391c7bf7013b251f756f48861c496/src/metadata/ColumnMetadata.ts#L504-L521

Since the default is always set, I think we can always use it. What do you think?

law108000:
> Okay, I get it now :) I think it makes sense.
> 
> I'm a bit surprised that `CURRENT_TIMESTAMP` was hard-coded in that place. Normally, this is driver dependent - `Driver.mappedDataTypes.updateDateDefault` has the function the gets the current time. So `updateDateColumn.default` should always be set to this value:
> 
> https://github.com/typeorm/typeorm/blob/27b4207c48a391c7bf7013b251f756f48861c496/src/metadata/ColumnMetadata.ts#L504-L521
> 
> Since the default is always set, I think we can always use it. What do you think?

Awesome! Thanks for suggesting the default config in ColumnMetadata! It really ties everything together perfectly!
I just made a change on the pull request. please take a look if it is okay. Thank you.

src/query-builder/UpdateQueryBuilder.ts
```diff
@@ -642,11 +642,16 @@ export class UpdateQueryBuilder<Entity extends ObjectLiteral>
...
-                    ) // todo: fix issue with CURRENT_TIMESTAMP(6) being used, can "DEFAULT" be used?!
+                ) {
+                    const updateDateColumn = metadata.updateDateColumn;
+                    let defaultValue = updateDateColumn.default;
+                    if (typeof defaultValue === "function") {
+                        defaultValue = defaultValue();
+                    }
+                    updateColumnAndValues.push(this.escape(updateDateColumn.databaseName) +
+                        " = " +
+                        defaultValue);
```
alumni, line 653: I would somehow do this:
```ts
valuesSetNormalized[updateDateColumn.databaseName] = updateDateColumn.default
```
and let the `else` block handle the whole thing 🤷🏻‍♂️

Why? `defaultValue` needs to be handled properly, e.g. if it's not a function it needs to be escape. Currently this is already implemented in the `else` block.

alumni (approved):

alumni:
Nice, @law108000. A test would be great to prevent this feature from breaking in the future.

test/github-issues/11258/issue-11258.ts
```diff
@@ -0,0 +1,796 @@
...
+        connections = await createTestingConnections({
+            entities: [
+                __dirname + "/entity_precision_0/Parent{.js,.ts}",
+                __dirname + "/entity_precision_0/Child{.js,.ts}",
+                __dirname + "/entity_precision_0/Base{.js,.ts}",
+                __dirname + "/entity_precision_6/Parent{.js,.ts}",
+                __dirname + "/entity_precision_6/Child{.js,.ts}",
+                __dirname + "/entity_precision_6/Base{.js,.ts}",
+            ],
+            enabledDrivers: ["mysql", "mariadb", "oracle", "postgres"],
```
alumni: `sap` can be added if you do any adjustments:
```suggestion
            enabledDrivers: ["mariadb", "mysql", "oracle", "postgres", "sap"],
```

test/github-issues/11258/issue-11258.ts
```diff
@@ -0,0 +1,796 @@
...
+    before(async () => {
+        connections = await createTestingConnections({
+            entities: [
+                __dirname + "/entity_precision_0/Parent{.js,.ts}",
+                __dirname + "/entity_precision_0/Child{.js,.ts}",
+                __dirname + "/entity_precision_0/Base{.js,.ts}",
+                __dirname + "/entity_precision_6/Parent{.js,.ts}",
+                __dirname + "/entity_precision_6/Child{.js,.ts}",
+                __dirname + "/entity_precision_6/Base{.js,.ts}",
+            ],
```
alumni: FYI, you can also pass an array of classes to `entities`:
```suggestion
            entities: [Parent0, Child0, Parent6, Child6],
```
(and `BaseEntity` is not really an entity, so no need to load it)

test/github-issues/11258/issue-11258.ts
```diff
@@ -0,0 +1,796 @@
...
+        const parent = Object.assign(new Parent(), {
+            id,
+            name: `Parent ${nameSuffix}`,
+        })
+        await connection.manager.save(parent)
+        const child = Object.assign(new Child(), {
+            id,
+            name: `Child ${nameSuffix}`,
+        })
+        await connection.manager.save(child)
```
alumni: `Repository.create` does `Object.assign(new EntityClass(), plainEntity)` in its implementation:
```ts
const parentRepository = connection.getRepository(Parent)
const parent = parentRepository.create({ id, name })
await parentRepository.save(parent)
```

However, all of this is done by the `EntityManager.save` in its implementation:
```ts
const parent = await connection.manager.save(new Parent(), { id, name })
```

So, this can be simplified:

```suggestion
        const parent = await connection.manager.save(new Parent(), {
            id,
            name: `Parent ${nameSuffix}`,
        })
        const child = await connection.manager.save(new Child(), {
            id,
            name: `Child ${nameSuffix}`,
        })
```

test/github-issues/11258/issue-11258-sqlserver.ts
```diff
@@ -0,0 +1,594 @@
...
+                await connection.manager.save(child)
+
+                const childBefore = await connection.manager.findOne(
+                    ChildSqlServer6,
+                    {
+                        where: { id: 21 },
+                    },
+                )
+
+                await new Promise((res) => setTimeout(res, 10))
```
alumni:
```ts
import { scheduler } from 'node:timers/promises';

await scheduler.wait(10);
```

test/github-issues/11258/issue-11258.ts
```diff
@@ -0,0 +1,796 @@
...
+                    where: { id: 51 },
+                })
+                expect(childAfter).to.not.be.null
+                expect(childAfter!.name).to.equal("Updated Child Precision 6")
+                expect(childAfter!.updated_date!.getTime()).to.be.greaterThan(
+                    childBefore!.updated_date!.getTime(),
+                )
+                expect(
+                    Number.isInteger(
+                        childAfter!.updated_date!.getMilliseconds(),
```
alumni: Doesn't seem relevant, `Date.getMilliseconds()` always returns an integer.

alumni:
The SAP tests fail because in HANA the timestamp precision must be between 1 and 7 (there's a different data type for precision 0), so creating the connection fails during schema setup. Let's say we can skip the SAP tests for now.

I think the other reviewers might comment on the test code quality.

emmettprexus:
Any update on this? I just ran into the same challenge of getting my updatedAt to ms precision and finally ended up here...

gioboa (changes requested):
is this PR still valid? can you solve the conflicts please?

michaelbromley:

```
git fetch origin
```

PR #11259, opened 2025-01-24
refactor: define Position type for GeoJSON objects

knoid:
### Description of change

This change requires all Point objects to match the defined structure specified in [RFC 7946](https://datatracker.ietf.org/doc/html/rfc7946#section-3.1.1). I also updated all the links to the specification.

> There MUST be two or more elements. [...] Altitude or elevation MAY be included as an optional third element.

### Pull-Request Checklist

### DIFF
--- a/src/driver/types/GeoJsonTypes.ts
+++ b/src/driver/types/GeoJsonTypes.ts
@@ -1,60 +1,66 @@
 /**
+ * Position object.
+ * https://datatracker.ietf.org/doc/html/rfc7946#section-3.1.1
+ */
+export type Position = number[]
+
+/**
  * Point geometry object.
- * https://tools.ietf.org/html/rfc7946#section-3.1.2
+ * https://datatracker.ietf.org/doc/html/rfc7946#section-3.1.2
  */
 export type Point = {
     type: "Point"
-    coordinates: number[]
+    coordinates: Position
 }
 
 /**
  * LineString geometry object.
- * https://tools.ietf.org/html/rfc7946#section-3.1.4
+ * https://datatracker.ietf.org/doc/html/rfc7946#section-3.1.4
  */
 export type LineString = {
     type: "LineString"
-    coordinates: number[][]
+    coordinates: Position[]
 }
 
 /**
  * Polygon geometry object.
- * https://tools.ietf.org/html/rfc7946#section-3.1.6
+ * https://datatracker.ietf.org/doc/html/rfc7946#section-3.1.6
  */
 export type Polygon = {
     type: "Polygon"
-    coordinates: number[][][]
+    coordinates: Position[][]
 }
 
 /**
  * MultiPoint geometry object.
- *  https://tools.ietf.org/html/rfc7946#section-3.1.3
+ *  https://datatracker.ietf.org/doc/html/rfc7946#section-3.1.3
  */
 export type MultiPoint = {
     type: "MultiPoint"
-    coordinates: number[][]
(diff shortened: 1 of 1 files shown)


naorpeled (approved):
Great catch, LGTM! 🔥

src/driver/types/GeoJsonTypes.ts
```diff
@@ -1,60 +1,66 @@
+/**
+ * Position object.
+ * https://datatracker.ietf.org/doc/html/rfc7946#section-3.1.1
+ */
+export type Position = [longitude: number, latitude: number, altitude?: number]
```
mguida22, line 5: While this matches the spec, this change requires that everyone types their coordinates as a `Position` (or similar). This will cause errors for most people using GeoJSON.

Consider the following ([ts playground](https://www.typescriptlang.org/play/?#code/C4TwDgpgBACg9gZwJbCXAdlAvFA2gGwwHMUBXAEwgC4p1SBbAIwgCcAaKfAQ1WAutoNm7KF3y9+Afhp0mrALoAoRQDNS6AMaoMUcnADKcehGAALJOiIAKMIhRp0NeMm3oAlFADeiqL6gsTUhZMWxcHRQBfZQB6aKgAQRYiBgh0YCg4FShQSCgAcllhXHk8qCQEWjh0rgRkInQuRnxoYDgoMC4WLmNgVgysnOg853sMPIA6RVi-ABVOohN-CABHUiQAioAmKAhm4zSrBA9GUnSEOCCNaHouEChTLgA3aBUIAHdWSY0MBHTQ+OweAALABWDigpR6Qw9cyWGyIeJuGJxc4wixEKDIehIbgsbJtMzlKBvHH4UT4N63CrMJardYQciKb7oX7tRAAIScdlcgNwoPBIMhBiMJlh1lC7KRiiAA))

```ts
type Position = [longitude: number, latitude: number, altitude?: number]

function doSomething(position: Position) {
    return position
}

// Fails with type error
// Argument of type 'number[]' is not assignable to parameter of type 'Position'.
//   Target requires 2 element(s) but source may have fewer.
const posA = [45, 45]
doSomething(posA)

// Succeeds
const posB: Position = [45, 45]
doSomething(posB)
```

We should follow the lead of the geojson types defined here unless we have a very strong reason to deviate. They have a good explanation as to why Position has remained defined as `number[]`. https://github.com/DefinitelyTyped/DefinitelyTyped/blob/5d7e8c0570c4c64da0133593c0e83edbce4a1b98/types/geojson/index.d.ts#L24-L65

mguida22 (changes requested):
Thanks for submitting this PR. The updated links are great, and I think if we change position to remain defined as `number[]` with an explanation as to why, this is an improvement overall.

@knoid is there a problem that you ran into due to the current types, or is it to keep in line with the spec?

src/driver/types/GeoJsonTypes.ts
knoid: My original idea was to follow the spec literally but I didn't see that previous conversation. I agree we should be compatible with GeoJSON. I'll update the types.

Thanks!

mguida22 (approved):
Thanks @knoid!

gioboa (approved):
Thanks :clap: @knoid

knoid:
I don't think this is a breaking change anymore. You can probably remove that label if that makes any difference. Thanks!

mguida22:
Yup, thanks for the reminder!

Issue #11260, opened 2025-01-25
Regression in transactionDepth handling?


aphofstede:
### Issue description

transactionDepth went negative

### Expected Behavior

transactionDepth is never decremented more often than it was incremented

### Actual Behavior

Negative transaction depth

### Steps to reproduce

Nest a couple of transactions and cause a DB error (e.g. insert a bad value into a column of the wrong type)

### My Environment

| Dependency          | Version  |
| ---                 | ---      |
| Operating System    |          |
| Node.js version     | 22.6.0  |
| Typescript version  | 5.4.5  |
| TypeORM version     | 0.3.20  |

### Additional Context

Seen in the wild, in a situation where an unhandled database error occurred deep down in a transaction stack, the transactionDepth appears to be decremented more often than it was incremented. When a new save point is attempted it will cause a syntax error due to the minus sign on the negative number.

Might be related to the changes from https://github.com/typeorm/typeorm/pull/10210

```
      "err": {
        "driverError": {
          "length": 90,
          "position": "19",
          "stack": "error: syntax error at or near \"-\"\n    at /app/node_modules/pg/lib/client.js:526:17\n    at process.processTicksAndRejections (node:internal/process/task_queues:95:5)\n    at async PostgresQueryRunner.query (/app/node_modules/typeorm/driver/postgres/PostgresQueryRunner.js:184:25)\n    at async PostgresQueryRunner.startTransaction (/app/node_modules/typeorm/driver/postgres/PostgresQueryRunner.js:130:13)\n    at async EntityPersistExecutor.execute (/app/node_modules/typeorm/persistence/EntityPersistExecutor.js:134:25)",
          "name": "error",
          "routine": "scanner_yyerror",
          "code": "42601",
          "type": "DatabaseError",
          "line": "1180",
          "severity": "ERROR",
          "message": "syntax error at or near \"-\"",
          "file": "scan.l"
        },
        "file": "scan.l",
[...]

PR #11261, opened 2025-01-25
fix: sqlite flaky test 

alumni:
### Description of change

Investigating SQLite flaky tests.

Currently identified issue is possibly related to #11260:
```
  1) github issues > #10209
       should not fail to run multiple nested transactions in parallel:
     QueryFailedError: SQLITE_ERROR: no such savepoint: typeorm_2
      at Statement.handler (src/driver/sqlite/SqliteQueryRunner.ts:17:65)
      at Statement.replacement (node_modules/sqlite3/lib/trace.js:25:27)
      at Statement.replacement (node_modules/sqlite3/lib/trace.js:25:27)
```

### Pull-Request Checklist

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1812,20 +1812,6 @@
         "node": ">=8"
       }
     },
-    "node_modules/@types/gulp/node_modules/fsevents": {
-      "version": "2.3.3",
-      "resolved": "https://registry.npmjs.org/fsevents/-/fsevents-2.3.3.tgz",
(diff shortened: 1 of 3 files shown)


OSA413:
i think i can help you with creating a test that can reproduce the error.

alumni:
The issue I referenced might not be affecting this test.

I think I know the error for the flaky tests, it's caused by how we work with savepoints in the test - we can't have parallel savepoints, we can only parallelize the top transaction, not the lower level ones like the test was doing before.

I actually don't know how the test worked for other db drivers - this is giving me a bit of doubt, some some confirmation that what I stated above is true or false would be helpful.

I didn't have time to look further, that's why I didn't update this anymore. I'll come back to it later this week, but feel free to suggest anything or even take over if you'd like.

alumni:
Declining this because I'm reverting the PR that is causing our issues in #11264 and wrote an additional test there.

PR #10210, opened 2023-07-17
fix: nested transactions issues

gmahomarf:
### Description of change
When running nested transactions in parallel, incrementing or decrementing the `transactionDepth` counter after starting a transaction or `SAVEPOINT`, and after releasing or rolling back to to a `SAVEPOINT` results in trying to create or release the same `SAVEPOINT` multiple times. This is due to the fact that while the event loop `await`s any of the aforementioned operations, another operation may start and will use the same value of `transactionDepth` used by a previous operation. By incrementing/decrementing the `transactionDepth` counter _before_, we avoid this

Fixes #10209

### Pull-Request Checklist
- [x] Code is up-to-date with the `master` branch
- [x] `npm run format` to apply prettier formatting
- [x] `npm run test` passes with this change
- [x] This pull request links relevant issues as `Fixes #0000`
- [x] There are new or updated unit tests validating the change
- [x] Documentation has been updated to reflect this change
- [x] The new commits follow conventions explained in [CONTRIBUTING.md](https://github.com/typeorm/typeorm/blob/master/CONTRIBUTING.md)

Issue #10209, opened 2023-07-17
Nested transactions (transactionDepth > 1) can attempt to create and release multiple `SAVEPOINT`s with the same name


gmahomarf:
### Issue description

Nested transactions (transactionDepth > 1) can attempt to create and release multiple `SAVEPOINT`s with the same name

### Expected Behavior

When running nested transactions, each transaction should successfully start and end without issues

### Actual Behavior

When running nested transactions, the `transactionDepth` counter is incremented/decremented incorrectly, leading to `SAVEPOINT`s with the same name. This eventually throws an exception when trying to release a `SAVEPOINT` that was released by another transaction.

### Steps to reproduce
The test at `test/github-issues/10209/issue-10209.ts` in #10210, when run without the patches to the multiple `QueryRunner`s, should reproduce the issue successfully

### My Environment

| Dependency          | Version  |
| ---                 | ---      |
| macOS    |  13.4.1   |
| Node.js version     | 18.16.0  |
| Typescript version  | 5.1.6  |
| TypeORM version     | 0.3.17 |

### Relevant Database Driver(s)

- [X] aurora-mysql
- [X] aurora-postgres
- [ ] better-sqlite3
- [X] cockroachdb
- [ ] cordova
- [ ] expo
- [ ] mongodb
- [X] mysql
- [ ] nativescript
- [X] oracle
- [X] postgres
- [ ] react-native
- [ ] sap
- [ ] spanner
- [ ] sqlite
- [X] sqlite-abstract
- [ ] sqljs
- [X] sqlserver

### Are you willing to resolve this issue by submitting a Pull Request?

Yes, I have the time, and I know how to start.

alumni:
@gmahomarf Could you explain what was the issue here?

The test in PR #10210 does not look like a valid use case of transactions (I'm talking about SQL in general, not just TypeORM). You cannot run transactions in parallel in the same connection.

gmahomarf:
Not sure what I can say that wasn't already explained in the PR. I don't even remember what the original issue was.

PR #11264, opened 2025-01-28
fix: transaction not ending correctly

### LINKED ISSUE #11260: Regression in transactionDepth handling?

alumni:
### Description of change

Fixes #11260.

Reverts #10210 which also introduced the flaky tests - the tests don't actually make sense.

This should also solve the CockroachDB issues we tried to solve in #11190. The main issue with CRDB is that if a query fails, you can't release the savepoint anymore - you can only roll back (see https://www.cockroachlabs.com/docs/stable/release-savepoint#handling-errors).

I also cleaned up two of tests a little bit, they had some unnecessary complexity.

### Pull-Request Checklist

- [X] This pull request links relevant issues as `Fixes #11260`


OSA413 (approved):
Looks like you managed to improve the code (the transactions and tests)

alumni:
@OSA413 I added a test for #11260 (fails on master, passes in this branch) and took the PR out of draft.

It can be reviewed now.

src/query-builder/RelationIdLoader.ts
```diff
@@ -636,7 +637,7 @@ export class RelationIdLoader {
                 undefined,
                 primaryColumn.entityMetadata.name +
                     "_" +
-                    relation.inverseRelation!.propertyPath.replace(".", "_") +
+                    originalRelation.propertyPath.replace(".", "_") +
```
G0maa, line 640: Since this is not part of the revert, what's happening here? :eyes:

G0maa:
For history reasons, this is a related [discussion](https://discord.com/channels/1124626013360488529/1336686763464200274) on TypeORM Discord.

src/query-builder/RelationIdLoader.ts
alumni: A test was failing because the value was undefined.

G0maa (approved):

sgarner (approved):
The code changes look fine. :+1:

Are we happy that this will reintroduce the bug described in #10210? It seems to me that submitting queries in parallel to a single transaction was always an incorrect practice which shouldn't be supported. But if anyone is relying on this behavior, reverting might represent a breaking regression.

alumni:
> Are we happy that this will reintroduce the bug described in #10210? It seems to me that submitting queries in parallel to a single transaction was always an incorrect practice which shouldn't be supported. But if anyone is relying on this behavior, reverting might represent a breaking regression.

@sgarner #10210 is not a bug, it is introducing a bug. Reverting that PR that introduced that bug will fix the regression introduced by it.

naorpeled (approved):

sgarner:
> @sgarner #10210 is not a bug, it is introducing a bug. Reverting the related PR will fix the regression introduced by it.

#10210 was fixing https://github.com/typeorm/typeorm/issues/10209

If you're arguing that #10209 wasn't a bug that needed fixing, then I'm in agreement with you. But we at least ought to note this in the changelog.

linbrandon4:
This is a big one, but the direction makes sense. Reverting #10210 to fix the transaction regression + flaky tests feels justified, especially since the “fix” was causing more problems than it solved. The CockroachDB angle also checks out — once a query fails, you can’t just release the savepoint, you have to roll back, so the old behavior was basically fighting the DB.

A few things I’d focus on if I were reviewing closely:
- Behavior change / possible breaking impact: the discussion about parallel queries inside a single transaction is important. If we’re deciding “promise.all inside one transaction isn’t supported,” that should be clearly stated somewhere (changelog/docs), because some users might have been relying on it even if it was never a good idea.
- Nested transaction correctness: make sure transactionDepth and savepoint naming/rollback behavior is consistent across drivers (Postgres vs Cockroach vs SQLite etc.), especially after a failure mid-transaction.
- Tests: since this PR touches 25 files and drops a ton of lines, the tests are basically the safety net. I like that there’s a test that fails on master and passes here (#11260) — I’d just want to confirm the “improved” tests actually reflect real driver behavior and aren’t masking edge cases.

PR #11266, opened 2025-01-28
chore: remove `--file` from test running

OSA413:
### Description of change

Remove `--file ./build/compiled/test/utils/test-setup.js` and move the `test-setup.ts` import into the necessary files. Removing the `--file` flag allows using `--parallel` flag (in the future).

### Pull-Request Checklist

### DIFF
--- a/package.json
+++ b/package.json
@@ -150,7 +150,7 @@
   "peerDependencies": {
     "@google-cloud/spanner": "^5.18.0",
     "@sap/hana-client": "^2.12.25",
-    "better-sqlite3": "^8.0.0 || ^9.0.0 || ^10.0.0 || ^11.0.0" ,
+    "better-sqlite3": "^8.0.0 || ^9.0.0 || ^10.0.0 || ^11.0.0",
     "hdb-pool": "^0.1.6",
     "ioredis": "^5.0.4",
     "mongodb": "^5.8.0",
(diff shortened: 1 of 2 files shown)


naorpeled:
Out of curiosity, how is it related to the parallel flag?

naorpeled (approved):

OSA413:
> Out of curiosity, how is it related to the parallel flag?

According to the Mocha docs, the `--parallel` flag is not compatible with the `--file` flag. https://mochajs.org/#-file-file

alumni:
We probably need to move that to some mocha hook.

`test-setup.js` should not be imported any by tests, neither `reflect-metadata`. For `chai` we need to decide, but it should be either always or never.

I'm actually rather confused by this `xfail`, looks like some old way of testing, didn't look at the history of it, but probably it's not really needed 🤷🏻‍♂️

OSA413:
I'll leave this here https://mochajs.org/#global-setup-fixtures since there is a PR that relies on the setup file.

OSA413:
After introducing the fix [to the timezones by setting the env from the setup file](https://github.com/typeorm/typeorm/blob/master/test/utils/test-setup.ts#L10) - I'm not sure. Need re-verifying at least.

OSA413:
It really looks like rewriting the setup file to a [root hook](https://mochajs.org/features/root-hook-plugins/) is better. With that rewritten, this file can be removed (probably). Closing the issue for now.

Issue #11265, opened 2025-01-28
Eager relations with DeleteDateColumn override manually requested relations with nested properties


LeviHeber:
### Issue description

When performing a TypeORM query with both manually specified relations and eager relations of entity with DeleteDateColumn, the current implementation incorrectly overwrites manually loaded relation data with partial eager relation data. This leads to loss of detailed information for manually requested nested relations.

### Expected Behavior

- Manually specified relations should be preserved in full
- Eager relations should complement, not replace, existing relation data
- Nested relations within manually specified relations should remain intact
- When multiple relation loading methods are used, all relation data should be merged correctly

### Actual Behavior

The eager entity is returnd without it nested entity that was manually requested.
During the query execution, TypeORM performs multiple joins to load the relations. When constructing the final object, the system overwrites the manually specified relation data with an eager-loaded version. This overwrite process results in a loss of detailed information for the manually specified nested relation, replacing it with a potentially incomplete representation that only includes eager-loaded fields.

### Steps to reproduce

1. Create an entity with an eager relation that contains another nested relation and has DeleteDateColumn.
2. Perform a find query where you manually specify the eager relation and it's nested relation.
3. Execute the query and retrieve the result.

### My Environment

| Dependency          | Version  |
| ---                 | ---      |
| Operating System    |  Windows  |
| Node.js version     | v20.10.0  |
| Typescript version  | 5.3.3  |
| TypeORM version     | 0.3.20  |

### Relevant Database Driver(s)

- [ ] aurora-mysql
- [ ] aurora-postgres
- [ ] better-sqlite3
- [ ] cockroachdb
- [ ] cordova
- [ ] expo
- [ ] mongodb
- [ ] mysql
- [ ] nativescript
- [ ] oracle
- [X] postgres
- [ ] react-native
- [ ] sap
- [ ] spanner
- [ ] sqlite
- [ ] sqlite-abstract
- [ ] sqljs
- [ ] sqlserver

### Are you willing to resolve this issue by submitting a Pull Request?

Yes, I have the time, and I know how to start.

LeviHeber:
Are there any updates? This is a patch that fixes several bugs at the same time, and it's worth moving forward with it.

Issue #11056, opened 2024-09-09
joinAlreadyAdded of FindOptionUtils#joinEagerRelation misses a lot of usecases


ngouy:
### Issue description

eager loded relationship erase nested relation

### Expected Behavior

Let's have a `ClassRooms` entity that eager loads `Students`
Let's have a `Student` entity that has one `ClassRoom(id)`, but also has one `Backpack(id)` < NOT EAGER LODED
Let's have a `BackPack` that has a `color`

When I run a query such as `ClassRoom.findOne({where: {id: 123}, relations: {students: {backpack: true}})`

I am able to then do `classRoom.students[0].backpack` to retrieve a student backpack

### Actual Behavior

Depending on a bunch of factor I don't understand, it's not always the case

Sometime when typeorm builds the sql, it builds 2 different section for the same "Students" relationship. One for the eager loaded, one for the relation in my find options

If the eager loaded is the one that is transformed last, I lose my backpack information

there is a mechanism that tries to avoid that [here](https://github.com/typeorm/typeorm/blob/master/src/find-options/FindOptionsUtils.ts#L417)
```ts
const joinAlreadyAdded = Boolean(qb.expressionMap.joinAttributes.find((joinAttribute) => joinAttribute.alias.name === relationAlias));
```

But when I put a debugger in my code, it's not working at all:

I have my "eager loaded" alias `UserEntity__rosters`
And I have my already existing alias from find option that targets the same attribute, but with a different alias: `'UserEntity__UserEntity_rosters`

(I'm wondering if it's because of the double query of distinct first then "Regular" query)

Also. I can see that when the query alias gets too big (from the find option), we can get some alias such as `'30d7fa239d182856aee094e3f106b7007d59eb3f'` which would also not be found by the code above

I guess the worse part about it is that it's indeterministic, and the last one that get's transformed wins ...

### Steps to reproduce

CF above

### My Environment

| Dependency          | Version  |
| ---                 | ---      |
| Operating System    |          |
| Node.js version     | 18.18  |
| Typescript version  | 5.4.2  |
| TypeORM version     | 9.0.1  |

### Relevant Database Driver(s)

- [ ] aurora-mysql
- [ ] aurora-postgres
- [ ] better-sqlite3
- [ ] cockroachdb
- [ ] cordova
- [ ] expo
- [ ] mongodb
- [ ] mysql
- [ ] nativescript
- [ ] oracle
- [X] postgres
- [ ] react-native
- [ ] sap
[...]

Issue #10812, opened 2024-04-03
Nested relations with an eager parent aren't fetched


naorpeled:
### Issue description

Nested relations with an eager parent aren't fetched

### Expected Behavior

I would expect that when I specify a sub relation within the tree I'd be able to get it although its parent is eager.

### Actual Behavior

At the moment the sub relation isn't loaded at all.

### Steps to reproduce

```typescript
/*   ParentEntity.ts.  */
@Entity({ name: 'ParentEntity' })
export class ParentEntity extends BaseEntity {
  @ManyToOne(() => Child, (child) => child.parents, { eager: true, nullable: true })
  @JoinColumn()
  child: Relation<Child>;

  @RelationId((parent: Parent) => parent.child)
  childId: string;
}

/*   ChildEntity.ts   */
@Entity({ name: 'ChildEntity' })
export class ChildEntity extends BaseEntity {
  @OneToMany(() => Parent, (parent) => parent.child)
  @JoinColumn()
  parents: Relation<Parent[]>;

  @OneToOne(() => ChildInfo, { nullable: true })
  @JoinColumn()
  info: Relation<ChildInfo>;
}

/*   ChildInfoEntity   */
@Entity({ name: 'ChildInfoEntity' })
export class ChildInfoEntity extends BaseEntity {
  @OneToMany(() => Child, (child) => child.info)
  @JoinColumn()
  parents: Relation<Parent[]>;

  @Column({ nullable: true })
  name: string;
}

/*   queries.ts   */
export const getParentWithChildren(parentId: string) => {
	return ParentEntity.find({
		where: { id: parentId },
		relations: ['child.info']
	});
}

/*   queries.test.ts   */
describe('getParentWithChildren', () => {
	beforeAll(() => {
		seedDb();
	})

	it('should return parent with child info', async () => {
		const parentWithChild = await getParentWithChildren('some-parent-id');

		expect(partnerWithChild.child).toBeDefined(); // this will pass, we do get the child object itself.
		expect(parentWithChild.child.info.name).toBeDefined(); // currently fails, info is not populated.
	})
})
```

### My Environment

| Dependency          | Version  |
| ---                 | ---      |
| Operating System    |     macOS Monterey     |
| Node.js version     | 16.15.0 |
| Typescript version  | 4.7.4 |
| TypeORM version     | 0.3.12 but also tried with 0.3.20  |

### Additional Context

When getting rid of eager and specifying the nested relation (info in this case),
we do see the expected result.

[...]

Issue #9774, opened 2023-02-10
Relations not being returned in some cases


ahmadnazir:
### Issue description

Relations are not being returned when using columns with `DeleteDateColumn` annotation

### Expected Behavior

Assuming that you have the following entities:

- `Address` many to one `Person` - eager loaded
- `Person` one to many `Car`

then the following code should return the relation for `cars`:

```
        repo.find({
            relations: [
                "person",
                "person.cars"
            ]
        })
```

### Actual Behavior

```
        repo.find({
            relations: [
                "person",
                "person.cars", // <- Not returned
            ]
        })
```

### Steps to reproduce

Here is a test case that reproduces the error: https://github.com/typeorm/typeorm/compare/0.2.45...ahmadnazir:typeorm:chore/eager-relations-with-delete-column/1

### My Environment

| Dependency          | Version  |
| ---                 | ---      |
| Operating System    |  Debian 11 (bullseye)        |
| Node.js version     | 16.18.0  |
| Typescript version  |  4.9.4 |
| TypeORM version     | 0.2.35, 0.2.45  |

### Additional Context

As a workaround, the `loadEagerRelations` can be set to `false` and the relations are returned.

### Relevant Database Driver(s)

- [ ] aurora-mysql
- [ ] aurora-postgres
- [ ] better-sqlite3
- [ ] cockroachdb
- [ ] cordova
- [ ] expo
- [ ] mongodb
- [ ] mysql
- [ ] nativescript
- [ ] oracle
- [X] postgres
- [ ] react-native
- [ ] sap
- [ ] spanner
- [ ] sqlite
- [ ] sqlite-abstract
- [ ] sqljs
- [ ] sqlserver

### Are you willing to resolve this issue by submitting a Pull Request?

No, I don’t have the time, but I can support (using donations) development.

zexcs:
I can confirm this issue.

TypeORM version: 0.3.6

Another workaround is to use { withDeleted: true } but neither of the workarounds are useful since it is breaking the intended behaviour of _soft-deletion_ and _eager_.

PR #11267, opened 2025-01-28
fix: prevent eager-loaded entities with DeleteDateColumn from overwriting manual relations

LeviHeber:
The issue was caused by the way TypeORM processing queries. For each join, it checks if the join for that table exists. However, if the join has a condition, it will always add the join, even if a previous join for the same table already exists.

In the case of eager loading, TypeORM creates its own join, and since it has a DeleteDateColumn, an additional condition is added to the join (checking for a null DeletedDate). This newly added join does not match the manually specified join, and if the manual relation had additional nested fields that were not part of the eager load, those fields would be overwritten.

This fix addresses the issue by merging the manually requested relations with the eager-loaded data, preserving the complete details of the manual relations.

In issue #10426 , the same case was described, where there are 2 joins to the same table,
which causes performance issues.

The current fix does not change the join process and leaves both joins as they were,
but only fixes the consequences that were created as a result of it in the object creation,
because this is a consequence that is relevant both to this case and to other cases.

Unrelated, it should check how to deal with the problem of having 2 joins, for performance reasons.

Closes: #11265
Closes: #11056
Closes: #10812
Closes: #9774
Closes: #9581

### Pull-Request Checklist

- [ ] N/A - Documentation has been updated to reflect this change

### DIFF
--- a/src/metadata/RelationMetadata.ts
+++ b/src/metadata/RelationMetadata.ts
@@ -520,6 +520,8 @@ export class RelationMetadata {
                 entity,
             )
         } else {
+            if (ObjectUtils.isObject(entity[propertyName]))
+                ObjectUtils.assign(value, entity[propertyName])
             entity[propertyName] = value
         }
     }
--- a/test/github-issues/11265/issue-11265.ts
+++ b/test/github-issues/11265/issue-11265.ts
@@ -1,0 +1,65 @@
+import "reflect-metadata"
+import {
+    createTestingConnections,
+    closeTestingConnections,
+    reloadTestingDatabases,
+} from "../../utils/test-utils"
+import { DataSource } from "../../../src/data-source/DataSource"
+import { expect } from "chai"
+import { RelationNestedEntity } from "./entity/relationNestedEntity"
+import { RelationEntity } from "./entity/relationEntity"
+import { ParentEntity } from "./entity/parent"
+
+describe.only("github issues > #11265 Eager Relations with DeleteDateColumn Overwrite Manual Relations During Query Loading", () => {
+    let dataSources: DataSource[]
+
+    before(
+        async () =>
+            (dataSources = await createTestingConnections({
+                entities: [__dirname + "/entity/*{.js,.ts}"],
+                schemaCreate: true,
+                dropSchema: true,
+            })),
+    )
+    beforeEach(() => reloadTestingDatabases(dataSources))
+    after(() => closeTestingConnections(dataSources))
+
+    it("should preserve manually requested nested relations", async () => {
+        return Promise.all(
+            dataSources.map(async (dataSource) => {
+                // Prepare test data
+                const relationNestedRepository =
+                    dataSource.getRepository(RelationNestedEntity)
+                const relationEntityRepository =
+                    dataSource.getRepository(RelationEntity)
+                const parentRepository = dataSource.getRepository(ParentEntity)
+
+                const relationNested = new RelationNestedEntity()
+                await relationNestedRepository.save(relationNested)
+
+                const relationEntity = new RelationEntity()
+                relationEntity.nested = relationNested
+                await relationEntityRepository.save(relationEntity)
+
+                const parent = new ParentEntity()
+                parent.relationEntity = relationEntity
+                await parentRepository.save(parent)
+
+                // Retrieve user with manually specified nested relation
+                const retrievedParent = await parentRepository.findOne({
+                    where: { id: parent.id },
+                    relations: {
+                        relationEntity: {
+                            nested: {},
+                        },
+                    },
+                })
+
+                // Assertions
+                expect(retrievedParent).to.exist
+                expect(retrievedParent!.relationEntity).to.exist
+                expect(retrievedParent!.relationEntity.nested).to.exist
+            }),
+        )
+    })
+})
--- a/test/github-issues/11265/entity/parent.ts
+++ b/test/github-issues/11265/entity/parent.ts
@@ -1,0 +1,11 @@
+import { Entity, ManyToOne, PrimaryGeneratedColumn } from "../../../../src"
+import { RelationEntity } from "./relationEntity"
+
+@Entity()
+export class ParentEntity {
+    @PrimaryGeneratedColumn()
+    id: number
+
+    @ManyToOne(() => RelationEntity, { eager: true })
+    relationEntity: RelationEntity
+}
--- a/test/github-issues/11265/entity/relationEntity.ts
+++ b/test/github-issues/11265/entity/relationEntity.ts
@@ -1,0 +1,18 @@
+import {
+    DeleteDateColumn,
+    Entity,
+    ManyToOne,
+    PrimaryGeneratedColumn,
+} from "../../../../src"
+import { RelationNestedEntity } from "./relationNestedEntity"
+@Entity()
+export class RelationEntity {
+    @PrimaryGeneratedColumn()
+    id: number
+
+    @DeleteDateColumn()
+    deletedAt?: Date
+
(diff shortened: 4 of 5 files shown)


gioboa (commented):
Hi @LeviHeber, thanks for this great PR, can you fix up the sap tests please?

test/github-issues/11265/issue-11265.ts
```diff
@@ -0,0 +1,65 @@
...
+        return Promise.all(
+            dataSources.map(async (dataSource) => {
+                // Prepare test data
+                const relationNestedRepository =
+                    dataSource.getRepository(RelationNestedEntity)
+                const relationEntityRepository =
+                    dataSource.getRepository(RelationEntity)
+                const parentRepository = dataSource.getRepository(ParentEntity)
+
+                const relationNested = new RelationNestedEntity()
```
alumni: SAP HANA cannot insert a row without values (there must be at least one value that is not auto-generated).

LeviHeber: I added a non auto-generated dummy field.

LeviHeber:
i left the test to be `describe.only`. is it need to be fixed?

gioboa (commented):
Yes, remove the .only please

LeviHeber:
I removed the `only` and pushed it again.
It seems that more tests are running now, but for some reason many of the tests encounter the same check (`commands - migration create: should write Javascript empty migration file when option is passed`) and it doesn't seem to be related to the fix I made.
I would appreciate some guidance.

sgarner:
> I removed the `only` and pushed it again. It seems that more tests are running now, but for some reason many of the tests encounter the same check (`commands - migration create: should write Javascript empty migration file when option is passed`) and it doesn't seem to be related to the fix I made. I would appreciate some guidance.

That sounds possibly related to #11653 which has been fixed now. Tests are passing in this PR.

test/github-issues/11265/issue-11265.ts
```diff
@@ -0,0 +1,65 @@
...
+    closeTestingConnections,
+    reloadTestingDatabases,
+} from "../../utils/test-utils"
+import { DataSource } from "../../../src/data-source/DataSource"
+import { expect } from "chai"
+import { RelationNestedEntity } from "./entity/relationNestedEntity"
+import { RelationEntity } from "./entity/relationEntity"
+import { ParentEntity } from "./entity/parent"
+
+describe("github issues > #11265 Eager Relations with DeleteDateColumn Overwrite Manual Relations During Query Loading", () => {
```
sgarner: We aren't creating tests for specific GitHub issues any more. Instead, tests should be added to the test suite(s) for the related features.

I suggest moving this test to `test/functional/relations/eager-relations/basic-eager-relations.ts`

src/metadata/RelationMetadata.ts
```diff
@@ -520,6 +520,8 @@ export class RelationMetadata {
                 entity,
             )
         } else {
+            if (ObjectUtils.isObject(entity[propertyName]))
+                ObjectUtils.assign(value, entity[propertyName])
             entity[propertyName] = value
```
sgarner: If I'm reading it right, the proposed change will only augment `entity[propertyName]` with _additional_ fields from `value`. Any fields already existing in `entity[propertyName]` won't be modified.

Also note that `ObjectUtils.assign()` mutates the target object. Depending on the `value` being passed, this might have unintended consequences.

Should we be merging objects deeply here? Is it possible for there to be two joins with overlapping _nested_ relations? 

```suggestion
            if (ObjectUtils.isObject(entity[propertyName])) {
                entity[propertyName] = OrmUtils.mergeDeep(entity[propertyName], value)
            } else {
                entity[propertyName] = value
            }
```

LeviHeber:
I made the following fixes:
1. Moved the test to its proper place in the eager-relations feature tests.
2. Updated the fix to work with `OrmUtils.mergeDeep` instead of `ObjectUtils.assign`

LeviHeber:
Is there anything else that needs to be done to move forward?

naorpeled:
@coderabbitai review

gioboa (approved):

alumni:
I think the issue that we have could possibly be a bit earlier, when we build the query. It seems we select the user twice - maybe once because it's an eagerly loaded relation and another time because it was requested by the user.

In the test we have:

```ts
await connection.manager.findOne(
    Editor,
    {
        where: { userId: 1 },
        relations: {
            user: {
                nestedProfile: true,
            },
        },
    },
)
```

which generates the query:

```sql
SELECT
    "Editor"."userId" AS "Editor_userId",
    "Editor"."postId" AS "Editor_postId",
    "Editor__Editor_user"."id" AS "Editor__Editor_user_id",
    "Editor__Editor_user"."firstName" AS "Editor__Editor_user_firstName",
    "Editor__Editor_user"."lastName" AS "Editor__Editor_user_lastName",
    "Editor__Editor_user"."deletedAt" AS "Editor__Editor_user_deletedAt",
    "Editor__Editor_user"."profileId" AS "Editor__Editor_user_profileId",
    "Editor__Editor_user"."nestedProfileId" AS "Editor__Editor_user_nestedProfileId",
    "Editor__Editor_user__Editor__Editor_user_nestedProfile"."id" AS "Editor__Editor_user__Editor__Editor_user_nestedProfile_id",
    "Editor__Editor_user__Editor__Editor_user_nestedProfile"."about" AS "Editor__Editor_user__Editor__Editor_user_nestedProfile_about",
    "Editor__Editor_user__Editor__Editor_user_profile"."id" AS "Editor__Editor_user__Editor__Editor_user_profile_id",
    "Editor__Editor_user__Editor__Editor_user_profile"."about" AS "Editor__Editor_user__Editor__Editor_user_profile_about",
    "Editor__user"."id" AS "Editor__user_id",
    "Editor__user"."firstName" AS "Editor__user_firstName",
    "Editor__user"."lastName" AS "Editor__user_lastName",
    "Editor__user"."deletedAt" AS "Editor__user_deletedAt",
    "Editor__user"."profileId" AS "Editor__user_profileId",
    "Editor__user"."nestedProfileId" AS "Editor__user_nestedProfileId",
    "Editor__user__profile"."id" AS "Editor__user__profile_id",
    "Editor__user__profile"."about" AS "Editor__user__profile_about"
FROM
    "editor" "Editor"
    LEFT JOIN "user" "Editor__Editor_user" ON "Editor__Editor_user"."id" = "Editor"."userId"
    AND ("Editor__Editor_user"."deletedAt" IS NULL)
    LEFT JOIN "profile" "Editor__Editor_user__Editor__Editor_user_nestedProfile" ON "Editor__Editor_user__Editor__Editor_user_nestedProfile"."id" = "Editor__Editor_user"."nestedProfileId"
    LEFT JOIN "profile" "Editor__Editor_user__Editor__Editor_user_profile" ON "Editor__Editor_user__Editor__Editor_user_profile"."id" = "Editor__Editor_user"."profileId"
    LEFT JOIN "user" "Editor__user" ON "Editor__user"."id" = "Editor"."userId"
    AND ("Editor__user"."deletedAt" IS NULL)
    LEFT JOIN "profile" "Editor__user__profile" ON "Editor__user__profile"."id" = "Editor__user"."profileId"
WHERE
    ((("Editor"."userId" = 1)))
    AND (
        "Editor"."userId" = 1
        AND "Editor"."postId" = 1
    )
```

We have both `Editor__Editor_user`+`Editor__Editor_user__Editor__Editor_user_profile` and `Editor__user`+`Editor__user__profile`.

I'll check tomorrow to see if this is intended.

alumni:
We'll create a separate issue for the extra things in the join so we don't block this.

Later edit: created #11823.

Issue #5967, opened 2020-04-29
@afterUpdate always says array/json field updated


zh99998:
**Issue type:**

[ ] question
[x] bug report
[ ] feature request
[ ] documentation issue

**Database system/driver:**

[ ] `cordova`
[ ] `mongodb`
[ ] `mssql`
[ ] `mysql` / `mariadb`
[ ] `oracle`
[x] `postgres`
[ ] `cockroachdb`
[ ] `sqlite`
[ ] `sqljs`
[ ] `react-native`
[ ] `expo`

**TypeORM version:**

[x] `latest`
[ ] `@next`
[ ] `0.x.x` (or put your version here)

**Steps to reproduce or a small repository showing the problem:**

I have an entity with some fields of json object and array type. like:

```typescript
  @Column({ type: 'jsonb', default: {} })
  data: object;
  @Column({ default: '{}', type: 'text', array: true)
  inventoryAvatars: string[];
  @Column({ default: '{}', type: 'text', array: true)
  inventoryEmoji: string[];
```

and a subscriber

```typescript
@EventSubscriber()
export class UserSubscriber implements EntitySubscriberInterface<User> {
  listenTo() {
    return User;
  }
  afterUpdate(event: UpdateEvent<User>): Promise<any> | void {
    console.log(event.updatedColumns.map(x=>x.propertyName));
  }
}
```

now I found that after any changes made, all array/object type fields are in event.updatedColumns. even they are not modified.

DEG-CD:
Do you found any solution? im facing the same problem with mysql

Thor99:
same problem

samresh9:
any updates regarding this issue?
I am facing same issue.

mnbaccari:
Facing the same issue, I believe it is because of this line: https://github.com/typeorm/typeorm/blob/master/src/persistence/SubjectChangedColumnsComputer.ts#L166

Comparing arrays with `===` will always return false, so the column is always added to the `diffColumns` which triggers an update to the DB.

PR #11269, opened 2025-01-31
fix: change how array columns are compared on column changed detection

mnbaccari:
Closes: #5967

### Description of change

Currently, if a column is marked as `array` (postgres only), like in the below example, the column is detected as having changed on every update, because it uses `===` comparison which will always be false for arrays. This results in unneeded UPDATE queries on the database. This PR fixes that by instead using `OrmUtils.deepCompare`, similar to how it is already done for json columns.

```ts
@Column('int', {
    array: true,
})
```

The tests added were failing before my change and succeed after.

Note that I'm not the author of the related issue #5967 which also mentions json columns, but the issue is old, and the json columns seem to be handled correctly already, so I worked on array only.

### Pull-Request Checklist

- [x] Documentation has been updated to reflect this change -- N/A


mnbaccari:
The only failing test is a known flaky test being investigated in https://github.com/typeorm/typeorm/pull/11261

mnbaccari:
@naorpeled sorry for the @, but I saw you re-ran the github flow for the tests. Just wondering if I should ping someone directly to get a review here? I also posted on the discord but didn't get a reply there: https://discord.com/channels/1124626013360488529/1333593115495501905/1333593115495501905

Thanks for the help.

naorpeled:
> @naorpeled sorry for the @, but I saw you re-ran the github flow for the tests. Just wondering if I should ping someone directly to get a review here? I also posted on the discord but didn't get a reply there: [discord.com/channels/1124626013360488529/1333593115495501905/1333593115495501905](https://discord.com/channels/1124626013360488529/1333593115495501905/1333593115495501905)
> 
> Thanks for the help.

Hey,
first of all feel free to ping me whenever, it's absolutely okay.

I was trying to understand if the failed tests were flaky 🙏 
The rest of the team and I are working on the next version and your PR is definitely on my mind as a candidate to be released.
I'm not approving just yet but we're on it.

mnbaccari:
Hey @naorpeled , any updates here? Can this be approved and merged?

naorpeled:
> Hey @naorpeled , any updates here? Can this be approved and merged?

Hey, 
we require two approvals for PRs so I've sent a message about this PR to the rest of the maintainers.

naorpeled (approved):

src/persistence/SubjectChangedColumnsComputer.ts
```diff
@@ -83,6 +83,12 @@ export class SubjectChangedColumnsComputer {
                 let normalizedValue = entityValue
                 // normalize special values to make proper comparision
                 if (entityValue !== null) {
+                    // If the column is an array, we don't care about column.type and instead compare the arrays directly
+                    if (column.isArray) {
+                        if (OrmUtils.deepCompare(entityValue, databaseValue))
+                            return
+                    }
+
```
sgarner, line 91: Postgres allows array columns to be defined with any regular base type for its elements. If you had an array of dates, would we want the elements in the array to be normalized before comparison like we do for non-array columns?

gioboa (approved):
Thanks for your help @mnbaccari

src/persistence/SubjectChangedColumnsComputer.ts
sgarner: I've approved that PR, let's include it here. I think that is the more "correct" solution so that we aren't leaving edge cases behind.

I suspect that date/time values on the `databaseEntity` were not normalized before because such values are usually persisted and loaded as strings, not as `Date` objects. But there is no harm in normalizing them, and this will allow for accurate change detection in the case where an entity has a `Date` value set in such a field and then is updated in the same code path.

mnbaccari: Makes sense thanks, I've merged that PR here.

sgarner (approved):
Thanks @mnbaccari !

naorpeled:
Thanks @mnbaccari 😎

Issue #11271, opened 2025-02-04
Stack trace is truncated in MySQL query runner


mukunda-:
### Issue description

When queries fail with MySQL, the stack trace is pointing to typeorm code

### Expected Behavior

The stack trace should include the call site before typeorm.

### Actual Behavior

The stack trace is terminated inside of the query runner code. Callbacks are resetting the stack.

### Steps to reproduce

1. Set up a MySQL environment
2. Create a query runner and perform an invalid query.
3. Observe the stack trace in the thrown error.

![Image](https://github.com/user-attachments/assets/4d9cf517-077d-4b0a-84a8-c7172f3e51a0)

### My Environment

| Dependency          | Version  |
| ---                 | ---      |
| Node.js version     | 20.15.1  |
| Typescript version  | ^5.1.3  |
| TypeORM version     | 0.3.20 |

### Additional Context

#8785 has some related info, but I'm not sure what database they are using. They say that some of the stack issues have been fixed in 0.3.x, but I still see issues with the MySQL driver.

I had a quick look at the Postgres driver, and I see that it doesn't have callbacks in the main `query` function, hence that should be giving proper stack traces when it [throws errors in the main stack](https://github.com/typeorm/typeorm/blob/master/src/driver/postgres/PostgresQueryRunner.ts#L331).

The MySQL `query` function has two levels of stack truncation, first with a promise, second with the callback to the underlying mysql driver. When it fails with the error, the [error is created in the second level, so it doesn't include the main stack.](https://github.com/typeorm/typeorm/blob/master/src/driver/mysql/MysqlQueryRunner.ts#L245)

I propose we edit the main `query` function to create the error in the main stack instead of the inner callback. I can submit a PR for that.

### Relevant Database Driver(s)

- [ ] aurora-mysql
- [ ] aurora-postgres
- [ ] better-sqlite3
- [ ] cockroachdb
- [ ] cordova
- [ ] expo
- [ ] mongodb
- [x] mysql
- [ ] nativescript
- [ ] oracle
- [ ] postgres
- [ ] react-native
- [ ] sap
- [ ] spanner
- [ ] sqlite
- [ ] sqlite-abstract
- [ ] sqljs
- [ ] sqlserver

### Are you willing to resolve this issue by submitting a Pull Request?

Yes, I have the time, and I know how to start.

PR #11274, opened 2025-02-04
Fixes for build and tests

### LINKED ISSUE #11272: Broken test case for filepathToName
### Issue description

A test case is corrupting the process environment

### Expected Behavior

The test case should define and validate the expected behavior. The test case should not change the environment for other tests.

### Actual Behavior

The test case is misusing forEach, afterEach and changing the "platform" of the environment that persists to other tests, causing them to fail on Windows.

### Steps to reproduce

1. Run `npm test` on Windows.

### My Environment

| Operating System    | Windows |
| TypeORM version     | 0.3.20 |

### Additional Context

In the [filepathToName() test](https://github.com/typeorm/typeorm/blob/master/test/functional/multi-database/multi-database-basic-functionality/multi-database-basic-functionality.ts#L22), `beforeEach` and `afterEach` is being used inside of a for loop. The for loop should be wrapping the outside of `describe`, as otherwise those are overwriting the platform (harming every other following test) and running the test incorrectly. This was causing the `persistence > basic functionality` suite to fail on Windows.

```
1) persistence > basic functionality
       should save an entity:
     EntityMetadataNotFoundError: No metadata for "Post" was found.
      at DataSource.getMetadata (src\data-source\DataSource.ts:450:30)
```

When corrected, the test is failing. I can see that when the platform is "darwin", the backslash to forward slash conversion is not used, where the test is  expecting that. I believe the test was incorrectly running all cases as "win32" when making the assertions.

I'm unsure of the desired behavior here. It seems like the code has an assumption that Windows-like paths will not be present on non-Windows platforms, and conversion is only done when running on Windows. If that is the case, then we should remove the [path conversion comparison](https://github.com/typeorm/typeorm/blob/master/test/functional/multi-database/multi-database-basic-functionality/multi-database-basic-functionality.ts#L55) from the test case when testing "darwin".

```
if (process.platform == "win32") { // Only test path conversion on Windows
    const otherOsRes = filepathToName(otherOs)
    expect(winOsRes).to.equal(otherOsRes)
}
```

- [x] aurora-mysql
- [x] aurora-postgres
- [x] better-sqlite3
- [x] cockroachdb
- [x] cordova
- [x] expo
- [x] mongodb
- [x] mysql
- [x] nativescript
- [x] oracle
- [x] postgres
- [x] react-native
- [x] sap
- [x] spanner
- [x] sqlite
- [x] sqlite-abstract
- [x] sqljs
- [x] sqlserver

mukunda-:
### Description of change

While trying to build and test #11271, I ran into a handful of issues. This PR contains a list of commits to get the build process working on Windows, as well as some other fixes in the tests.

See the commit details for each one for additional info. Brief summary:

- **fix multi-database-basic-functionality**: This test was corrupting the `process.platform` setting for later tests. Also incorrect test logic. #11272
- **fix rimraf usage:** The current rimraf package has some breaking changes and requires an option to be specified to work with globs. Otherwise the build fails.
- **port number missing for issues test 4753**: This test was missing the port number in the configuration block, causing a fail if using nonstandard ports for mysql.
- **windows support for issue 8975 test**: This test was failing on Windows as it relied on a Unix shebang/executing JS scripts directly.
- **fix issue 9230 test for western timezones**: This test was not working if you have a western timezone (which would subtract 1 from the day).

Only one of these fixes has a corresponding issue opened.

Fixes #11272

### Pull-Request Checklist

- [x] Documentation has been updated to reflect this change N/A - no functionality changes

### DIFF
--- a/gulpfile.ts
+++ b/gulpfile.ts
@@ -21,7 +21,7 @@ export class Gulpfile {
      */
     @Task()
     async clean() {
-        return rimraf(["./build/**"]);
+        return rimraf(["./build/**"], {glob: true});
     }
 
     /**
@@ -88,7 +88,7 @@ export class Gulpfile {
     async browserClearPackageDirectory() {
         return rimraf([
             "./build/browser/**"
-        ]);
+        ], {glob: true});
     }
 
     // -------------------------------------------------------------------------
@@ -195,7 +195,7 @@ export class Gulpfile {
     async packageClearPackageDirectory() {
         return rimraf([
             "build/package/src/**"
-        ]);
+        ], {glob: true});
     }
 
     /**
--- a/test/functional/multi-database/multi-database-basic-functionality/multi-database-basic-functionality.ts
+++ b/test/functional/multi-database/multi-database-basic-functionality/multi-database-basic-functionality.ts
@@ -19,8 +19,8 @@ import appRoot from "app-root-path"
 const VALID_NAME_REGEX = /^(?!sqlite_).{1,63}$/
 
 describe("multi-database > basic-functionality", () => {
-    describe("filepathToName()", () => {
-        for (const platform of [`darwin`, `win32`]) {
+    for (const platform of [`darwin`, `win32`]) {
+        describe(`filepathToName() (${platform})`, () => {
             let realPlatform: string
 
             beforeEach(() => {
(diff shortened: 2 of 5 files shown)


OSA413:
possible duplicate of https://github.com/typeorm/typeorm/pull/11257

OSA413:
I think that the approach for `multi-database-basic-functionality` fix should be that we don't overwrite the `process.platform` at all. I'll call other maintainers to judge us.

Everything else looks fine.

mukunda-:
Thanks for the feedback. Admittedly I didn't do a very thorough search for duplicates of these fixes. I'm a bit of a foreigner to this area and I'm fine with any conventions in place to merge/split fixes (not sure if a merge commit is used or not either).

I agree that touching process.platform is a bit iffy - the only upside is that it will increase test coverage

alumni (commented):
Some of these were already fixed, would be great to update the branch and see what's left.

OSA413:
Yeah, looks like most of the fixes were introduced by other PRs.

pkuczynski:
@mukunda- do you want to continue this, or someone else should take over?

PR #11283, opened 2025-02-11
fix: Make `shorten` method to properly work with camelCase_aliases

### LINKED ISSUE #9544: StringUtils `shorten()` function can result in collisions
## Issue Description

The [shorten](https://github.com/typeorm/typeorm/blob/c4a36da62593469436b074873eba186f0f8b990d/src/util/StringUtils.ts#L76-L92) function is used to ensure that an alias in a generated query does not exceed the maximum allowed by the DB.

It does this by truncating words such that `UserShoppingCart__order__market__director` becomes `UsShCa__orde__mark__dire`.

Users have reported the following error:

```
ER_NONUNIQ_TABLE: Not unique table/alias: 'SuStInTr__suSt__prVaFiFrAs'"
```

This has been traced down to the `shorten()` method, which will generate an identical alias for these two non-identical inputs.

### Expected Behavior

Unique inputs should _always_ result in unique outputs.

### Actual Behavior

```
console.log(shorten('SupplierStockInTransit__supplierStock__productVariant_customFields_tryonFrameAsset'));
// -> 'SuStInTr__suSt__prVaFiFrAs'
console.log(shorten('SupplierStockInTransit__supplierStock__productVariant_customFields_visualFrameAsset'));
// -> 'SuStInTr__suSt__prVaFiFrAs'
```

### Steps to Reproduce

1. Copy the code from the `shorten` source into a file.
2. Give it the two inputs I specify above
3. Observe identical outputs

```typescript
  export function shorten(input: string, options: IShortenOptions = {}): string {
    const { segmentLength = 4, separator = "__", termLength = 2 } = options;


    const segments = input.split(separator);
    const shortSegments = segments.reduce((acc: string[], val: string) => {
        // split the given segment into many terms based on an eventual camel cased name
        const segmentTerms = val.replace(/([a-z\xE0-\xFF])([A-Z\xC0-\xDF])/g, "$1 $2").split(" ");
        // "OrderItemList" becomes "OrItLi", while "company" becomes "comp"
        const length = segmentTerms.length > 1 ? termLength : segmentLength;
        const shortSegment = segmentTerms.map(term => term.substr(0, length)).join("");


        acc.push(shortSegment);
        return acc;
    }, []);


    return shortSegments.join(separator);
  }

console.log(shorten('SupplierStockInTransit__supplierStock__productVariant_customFields_tryonFrameAsset'));
console.log(shorten('SupplierStockInTransit__supplierStock__productVariant_customFields_visualFrameAsset'));
```

Reproduction in the context of a running TypeORM app can be found here: https://github.com/tianyingchun/vendure-issue and further described in:

- https://github.com/vendure-ecommerce/vendure/issues/1873

### My Environment

| Operating System    |   Windows       |
| Node.js version     | 16.15.1  | 
| Typescript version  | 4.3.5  | 
| TypeORM version     | 0.2.45 | 

I'll put yes on the ones I tested on, but I'm pretty sure it affects more or all.

OSA413:
Fixes #9544

### Pull-Request Checklist

### DIFF
--- a/src/util/StringUtils.ts
+++ b/src/util/StringUtils.ts
@@ -92,6 +92,7 @@ export function shorten(input: string, options: IShortenOptions = {}): string {
         // split the given segment into many terms based on an eventual camel cased name
         const segmentTerms = val
             .replace(/([a-z\xE0-\xFF])([A-Z\xC0-\xDF])/g, "$1 $2")
+            .replace(/(_)([a-z])/g, " $2")
             .split(" ")
         // "OrderItemList" becomes "OrItLi", while "company" becomes "comp"
         const length = segmentTerms.length > 1 ? termLength : segmentLength
--- a/test/github-issues/9544/issue-9544.ts
+++ b/test/github-issues/9544/issue-9544.ts
@@ -1,0 +1,35 @@
+import { expect } from "chai";
+import { shorten } from "../../../src/util/StringUtils"
+
+describe("github issues > #9544 shorten long camel case alias correctly", () => {
+    const testCases = 
+    [
+        {
+            input: "SupplierStockInTransit__supplierStock__productVariant_customFields_tryonFrameAsset",
+            expected: "SuStInTr__suSt__prVacuFitrFrAs",
+        },
+        {
+            input: "SupplierStockInTransit__supplierStock__productVariant_customFields_visualFrameAsset",
+            expected: "SuStInTr__suSt__prVacuFiviFrAs",
+        },
+        {
+            input: "camelCase_alias",
+            expected: "caCaal",
+        }
+    ];
+    
+    testCases.map((testCase, i) => {
+        it(`should produce correct results for camel case aliases, case #${i+1}`, () => {
+            const {input, expected} = testCase;
+            expect(shorten(input)).to.equal(expected);
+        })
+    })
+
+    it("should produce unique results for different inputs", () => {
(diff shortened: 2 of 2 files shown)


mguida22 (commented):
This won't completely address the issue in https://github.com/typeorm/typeorm/issues/9544, since `shorten` can still produce identical outputs for different inputs.

For example
```js
// these differ by visualFrameAsset vs virtualFrameAsset
shorten("SupplierStockInTransit__supplierStock__productVariant_customFields_visualFrameAsset")
shorten("SupplierStockInTransit__supplierStock__productVariant_customFields_virtualFrameAsset")

// and shorten produces the same result
"SuStInTr__suSt__prVacuFiviFrAs"
```

To fully address the issue aliases probably need some kind of hash of their contents. I wonder how other ORMs handle this?

This PR might still be an improvement since it handles camel case aliases better than before.

alumni (approved):

gioboa (approved):
LGTM @OSA413 👏

michaelbromley:
Actually @OSA413 would you be able to move the unit test for this to a better location as per Qodo's suggestion, so we move away from the issue-based test structure?  

```
• The shorten function is a utility function that already has related tests in other locations
  (test/unit/util/string-utils.test.ts exists for other StringUtils functions), making it a good
  candidate for functional/unit testing.
```

update: I commented this at the exact moment that Giorgio merged.. but we discussed further and in any case there is still work to be done to clearly define and then move all the other tests.

PR #11293, opened 2025-02-15
fix: set undefined in place of null primary columns

### LINKED ISSUE #10487: Repository.save function thinks a relation entity excist if id is NULL
### Issue description

When running the Repository.save function whit a one to many relation whereof the many items have an id of NULL will cause them to be not created except the first one

### Expected Behavior

The 3 post to be created in the DB whit the relation to the User

### Actual Behavior

The posts whit the id of NULL does not get created except the first one.
The posts whit the id of undifined also gets created

### Steps to reproduce

Run next code (Minimal issue reproduction)

```ts
import { Entity, PrimaryGeneratedColumn, Column, ManyToOne, OneToMany, BaseEntity, DataSource } from "typeorm"

//------------------------------------------------------------
// define entities
//------------------------------------------------------------

@Entity()
export class User extends BaseEntity {
    @PrimaryGeneratedColumn()
    id: number

    @Column({
        type: "varchar",
        length: 100,
    })
    firstName: string

    @Column({
        type: "varchar",
        length: 100,
    })
    lastName: string

    @Column({
        type: "int",
    })
    age: number

    @OneToMany(() => Post, post => post.user, {
        cascade: true,
        eager: true
    })
    posts: Post[];
}

@Entity()
export class Post extends BaseEntity {
    @PrimaryGeneratedColumn({
        type: "int",
    })
    id: any //in compile time this is set to number but in run time null appears thats why I set it to any

    @Column({
        type: "varchar",
        length: 100,
    })
    title: string

    @Column({
        type: "varchar",
        length: 100,
    })
    content: string

    @ManyToOne(() => User, user => user.posts)
    user: User;
}

//------------------------------------------------------------
// create connection and setup the connection with the database
//------------------------------------------------------------

    type: "postgres",
    host: "localhost",
    port: 5432,
    username: "admin",
    password: "mysecretpassword",
    database: "test",
    entities: [User, Post],
    synchronize: true,
    logging: false,
})

// to initialize the initial connection with the database, register all entities
// and "synchronize" database schema, call "initialize()" method of a newly created database
// once in your application bootstrap
AppDataSource.initialize()
    .then(() => {
        createUser();
    })
    .catch((error) => console.log(error))


//------------------------------------------------------------
// create and save a user
//------------------------------------------------------------

const createUser = () => {
    AppDataSource.getRepository(User).save({
        firstName: "Timber",
        lastName: "Saw",
        age: 25,
        posts: [{
            id: null,
            title: "First post",
            content: "Timber is a cat"
        }, {
            id: null,
            title: "Second Post",
            content: "Timber is a cat"
        }, {
            id: undefined,
            title: "Third Post",
            content: "Timber is a cat"
        }]
    }).then(user => {
        console.log("User has been saved: ", user)
        findAllUsers()
    }).catch(error => {
        console.log("Cannot save. Error: ", error)
    })
}

//------------------------------------------------------------
// find all users
//------------------------------------------------------------

const findAllUsers = () => {
    AppDataSource.getRepository(User).find({
        relations: ["posts"]
    }).then(users => {
        users.forEach(user => {
            console.log("User: ", user.id)
            console.table(user.posts)
        });
    }).catch(error => {
        console.log("Cannot get users. Error: ", error)
    })
}
```

### My Environment

| Operating System    |     MacOS     |
| Node.js version     | 16.20.2  |
| Typescript version  | 5.2.2 |

### Additional Context

_No response_

- [ ] aurora-mysql
- [ ] aurora-postgres
- [ ] better-sqlite3
- [ ] cockroachdb
- [ ] cordova
- [ ] expo
- [ ] mongodb
- [ ] mysql
- [ ] nativescript
- [ ] oracle
- [X] postgres
- [ ] react-native
- [ ] sap
- [ ] spanner
- [ ] sqlite
- [ ] sqlite-abstract
- [ ] sqljs
- [ ] sqlserver

OSA413:
Fixes #10487

### Description of change

Added a place where primary columns will turn into undefined if user passes null to them. We should consider the use case if this is actually appliable.

### Pull-Request Checklist

### DIFF
--- a/src/persistence/Subject.ts
+++ b/src/persistence/Subject.ts
@@ -309,6 +309,17 @@ export class Subject {
     recompute(): void {
         if (this.entity) {
             this.entityWithFulfilledIds = Object.assign({}, this.entity)
+            
+            this.metadata.primaryColumns.forEach((primaryColumn) => {
+                if (primaryColumn.getEntityValue(this.entity!) === null) {
+                    // You can't have a null in a privary column, maybe user wants to insert the default value from DB
+                    primaryColumn.setEntityValue(
+                        this.entityWithFulfilledIds!,
+                        undefined,
+                    )
+                }
+            })
+
             if (this.parentSubject) {
                 this.metadata.primaryColumns.forEach((primaryColumn) => {
                     if (
--- a/src/query-builder/InsertQueryBuilder.ts
+++ b/src/query-builder/InsertQueryBuilder.ts
@@ -846,7 +846,9 @@ export class InsertQueryBuilder<
                         )
 
                         // if value for this column was not provided then insert default value
-                    } else if (value === undefined) {
+                    } else if (value === undefined
+                        || (value === null && column.isPrimary)
+                    ) {
                         if (
                             (this.connection.driver.options.type === "oracle" &&
                                 valueSets.length > 1) ||
--- a/test/github-issues/10487/entity/Post.ts
+++ b/test/github-issues/10487/entity/Post.ts
@@ -1,0 +1,25 @@
+import { BaseEntity, Column, Entity, ManyToOne, PrimaryGeneratedColumn } from "../../../../src";
+import { User } from "./User";
+
+@Entity()
+export class Post extends BaseEntity {
+    @PrimaryGeneratedColumn({
+        type: "int",
+    })
+    id: any
+
+    @Column({
+        type: "varchar",
+        length: 100,
+    })
+    title: string
+
+    @Column({
+        type: "varchar",
+        length: 100,
+    })
+    content: string
+
+    @ManyToOne(() => User, user => user.posts)
+    user: User;
+}
--- a/test/github-issues/10487/entity/User.ts
+++ b/test/github-issues/10487/entity/User.ts
@@ -1,0 +1,31 @@
+import { BaseEntity, Column, Entity, OneToMany, PrimaryGeneratedColumn } from "../../../../src";
+import { Post } from "./Post";
+
+@Entity()
+export class User extends BaseEntity {
+    @PrimaryGeneratedColumn()
+    id: number
+
+    @Column({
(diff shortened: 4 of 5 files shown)


OSA413:
I personally don't like how it turned out, do we have tests that incorporate complex queries with several primary keys?

src/persistence/Subject.ts
```diff
@@ -309,6 +309,17 @@ export class Subject {
     recompute(): void {
         if (this.entity) {
             this.entityWithFulfilledIds = Object.assign({}, this.entity)
+            
+            this.metadata.primaryColumns.forEach((primaryColumn) => {
```
naorpeled, line 313: maybe worth reverting this, as it can cause unpredictable behaviors in non insert queries, wdyt?

OSA413: I'll try

OSA413: Oh, this fix is needed for the `save` method to correctly convert child entities with `null` id to `Subject`s since internal converter thinks that they are the same. Does the solution look hacky?

OSA413: Yeah, reverting this block makes the test fail. Maybe we should reconsider the issue first? Is it actually OK to explicitly set primary columns to `null` instead of not providing them or setting to `undefined`?

naorpeled: Ah, I see.
It does feel hacky, not saying that it's necessarily a thing that we must avoid, but it can cause some side effects we aren't aware of.

I'm not sure what's the correct solution is, but I think we need to give this some thought over discussion about on Discord with the rest of the team.
Wdyt?

OSA413: Yeah, let's discuss

OSA413:
Closing since some DBMS support NULLs in primary keys.

PR #11298, opened 2025-02-18
fix: ensure correct MSSQL parameter conversion in where conditions

### LINKED ISSUE #11285: Missing MSSQL input type
### Issue description

In MSSQL, when using the "find" method with where conditions on a "varchar" column, the query runner fails to specify the parameter type, causing performance issues.

[Refer](https://github.com/typeorm/typeorm/blob/master/src/driver/sqlserver/SqlServerQueryRunner.ts#L204)

### Expected Behavior

Typeorm should handle the parameter input to node-mssql correctly

### Actual Behavior

According to the [node-mssql driver docs](https://tediousjs.github.io/node-mssql/#input-name-type-value), the input type defaults to a JS data type. However, for "varchar," it appears to conflict with "nvarchar," leading to the wrong data type being used in queries.

As a result, runtime data conversion occurs, significantly impacting query performance (our application crashed under high load due to this issue).

For now, we're using a workaround by executing raw queries and manually handling SQL parameters. This resolved the performance issue, but it should be handled by the ORM itself.

### Steps to reproduce

1. Define a varchar column
```
@Index()
@Column({ type: 'varchar', nullable: true })
 memberId: string | null
```

2. Query the entity with where condition on that column
```
this.somEntityRepo.find({ where: { memberId: "someValue" } })
```
On a large dataset, you can verify the resources used by this query.

---

Workaround which resolved the performance issue
```
 await this.connection.query(`SELECT * FROM "someEntity" WHERE "memberId" = ${'someValue'}`)
```

### My Environment

| Node.js version     | 20.16.0  |
| Typescript version  | 5.5.4 |
| TypeORM version     | 0.3.20 |
| mssql     | 10.0.4 |

### Additional Context

_No response_

- [ ] aurora-mysql
- [ ] aurora-postgres
- [ ] better-sqlite3
- [ ] cockroachdb
- [ ] cordova
- [ ] expo
- [ ] mongodb
- [ ] mysql
- [ ] nativescript
- [ ] oracle
- [ ] postgres
- [ ] react-native
- [ ] sap
- [ ] spanner
- [ ] sqlite
- [ ] sqlite-abstract
- [ ] sqljs
- [x] sqlserver

sudhirt4:
Fixes input parameter conversion in SelectQueryBuilder when using an MSSQL connection.

Closes #11285

### Description of change

Instead of passing raw input paramters as it is this change uses the available driver.parametrizeValue method to convert parameters into MSSQL format. 

This implementation was referred from existing usage in [UpdateQueryBuilder](https://github.com/typeorm/typeorm/blob/master/src/query-builder/UpdateQueryBuilder.ts#L553) and [InsertQueryBuilder](https://github.com/typeorm/typeorm/blob/master/src/query-builder/InsertQueryBuilder.ts#L887)

Please note that this issue currently only fixes where condition when used under `.find` method. On testing it was not working for `.createQueryBuilder(..).where()` usage. Could use some insights on a proper approach to implement this parameter transformation.

### Pull-Request Checklist
- [x] This pull request links relevant issues as `Fixes #11285`

### DIFF
--- a/src/query-builder/SelectQueryBuilder.ts
+++ b/src/query-builder/SelectQueryBuilder.ts
@@ -45,6 +45,7 @@ import { AuroraMysqlDriver } from "../driver/aurora-mysql/AuroraMysqlDriver"
 import { InstanceChecker } from "../util/InstanceChecker"
 import { FindOperator } from "../find-options/FindOperator"
 import { ApplyValueTransformers } from "../util/ApplyValueTransformers"
+import { SqlServerDriver } from "../driver/sqlserver/SqlServerDriver"
 
 /**
  * Allows to build complex sql queries in a fashion way and execute those queries.
@@ -4270,6 +4271,7 @@ export class SelectQueryBuilder<Entity extends ObjectLiteral>
                     if (InstanceChecker.isEqualOperator(where[key])) {
                         parameterValue = where[key].value
                     }
+
                     if (column.transformer) {
                         parameterValue instanceof FindOperator
                             ? parameterValue.transformValue(column.transformer)
@@ -4278,6 +4280,25 @@ export class SelectQueryBuilder<Entity extends ObjectLiteral>
                                       column.transformer,
                                       parameterValue,
                                   ))
+                    }
+
+                    // MSSQL requires parameters to carry extra type information
+                    if (this.connection.driver.options.type === "mssql") {
+                        const driver = this.connection.driver as SqlServerDriver
+                        if (parameterValue instanceof FindOperator) {
+                            if (parameterValue.type !== "raw") {
+                                parameterValue.transformValue({
+                                    to: (v) =>
+                                        driver.parametrizeValue(column, v),
+                                    from: (v) => v,
+                                })
+                            }
+                        } else {
+                            parameterValue = driver.parametrizeValue(
+                                column,
+                                parameterValue,
+                            )
+                        }
                     }
 
                     // if (parameterValue === null) {
--- a/test/github-issues/11285/entity/user.ts
+++ b/test/github-issues/11285/entity/user.ts
@@ -1,0 +1,13 @@
+import { Column, Entity, Index, PrimaryGeneratedColumn } from "../../../../src"
+
+@Entity({
+    name: "user",
+})
+export class User {
+    @PrimaryGeneratedColumn()
+    id: number
+
+    @Index()
+    @Column({ type: "varchar", nullable: true })
+    memberId: string
+}
--- a/test/github-issues/11285/issue-11285.ts
+++ b/test/github-issues/11285/issue-11285.ts
@@ -1,0 +1,197 @@
+import "reflect-metadata"
+import { expect } from "chai"
+import sinon from "sinon"
+
+import {
+    createTestingConnections,
+    closeTestingConnections,
+    reloadTestingDatabases,
+} from "../../utils/test-utils"
+import { DataSource, MssqlParameter, Not, Raw } from "../../../src/index.js"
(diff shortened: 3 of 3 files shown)


sgarner:
So this is tangentially related to https://github.com/typeorm/typeorm/pull/10161

For clarity, the problem is that MSSQL requires query parameters to be typed, and if no type is given then the underlying node-mssql driver defaults to NVARCHAR. If the column is actually a VARCHAR and is indexed, SQL Server won't use the index due to the type mismatch.

src/query-builder/SelectQueryBuilder.ts
```diff
@@ -4278,6 +4280,12 @@ export class SelectQueryBuilder<Entity extends ObjectLiteral>
                                       column.transformer,
                                       parameterValue,
                                   ))
+                    } else {
```
sgarner, line 4283: Are you sure this path should be in the else branch here?

I think the value needs to be 'parametrized' [sic] regardless of whether the column has a value transformer or not. The transformer should be applied first, and then parameterization.

sgarner (changes requested):
@sudhirt4 I've run the tests, and some are failing - can you have a look at it?

sudhirt4:
> @sudhirt4 I've run the tests, and some are failing - can you have a look at it?

Updated with the changes. 

Figured out that params of type FindOperator had to be [handled separately](https://github.com/typeorm/typeorm/pull/11298/files#diff-52bea788267888fdbb5f795f0dfb61504f16bd43e0afdfa28848abb4435121b5R4290). Instead of changing other functions to handle this I have adjusted it here following the approach used in transformer just above this. Let me know if this pattern is acceptable.

Have updated the tests to include these scenarios as well

src/query-builder/SelectQueryBuilder.ts
sudhirt4: Makes sense, updated with changes✅ 

I was worried if there would be some return type from transformer that parametrize function could not handle but seems like it's working just fine.

src/query-builder/SelectQueryBuilder.ts
```diff
@@ -4280,6 +4282,24 @@ export class SelectQueryBuilder<Entity extends ObjectLiteral>
...
+                                to: (v) => driver.parametrizeValue(column, v),
+                                from: (v) => v,
+                            })
+                        } else {
+                            parameterValue = driver.parametrizeValue(
+                                column,
+                                parameterValue,
+                            )
+                        }
+                    }
```
sgarner, line 4301: Although this works, the comparison against `parameterValue.type` is premature; `parameterValue` might be anything but the `type` property comparison only applies when it is a `FindOperator` instance. Suggest the raw check be moved to the inner condition.

Also adding a comment to explain what this block does:

```suggestion
                    // MSSQL requires parameters to carry extra type information
                    if (this.connection.driver.options.type === "mssql") {
                        const driver = this.connection.driver as SqlServerDriver
                        if (parameterValue instanceof FindOperator) {
                            if (parameterValue.type !== 'raw') {
                                parameterValue.transformValue({
                                    to: (v) => driver.parametrizeValue(column, v),
                                    from: (v) => v,
                                })
                            }
                        } else {
                            parameterValue = driver.parametrizeValue(
                                column,
                                parameterValue,
                            )
                        }
                    }
```

sgarner (approved):
Thank you @sudhirt4 for your effort on this!

src/query-builder/SelectQueryBuilder.ts
sudhirt4: Updated ✅ I amended by commit to include these changes

sgarner (approved):
LGTM :tada:

PR #11303, opened 2025-02-20
Fix release issues

michaelbromley:
This fixes a couple of issues I encountered when following Umed's instructions to create a new TypeORM release.

### DIFF
--- a/gulpfile.ts
+++ b/gulpfile.ts
@@ -21,7 +21,7 @@ export class Gulpfile {
      */
     @Task()
     async clean() {
-        return rimraf(["./build/**"]);
+        return rimraf(["./build/**"], { glob: true });
     }
 
     /**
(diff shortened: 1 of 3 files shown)


alumni:
`rimraf@6` is not supported on node 18.

Adding parameters to the rimraf call should be done in other places (tests) too, there were some PRs doing that 🙂

michaelbromley:
Ah OK I didn't know that. Good point though - rimraf is just a dev dependency so won't affect node compatibility for end users. Or are we aiming to support node 18 also for library development?

alumni:
@michaelbromley we run the tests on node 18 in the pipelines (github actions). They use rimraf quite a lot, so if they didn't fail, probably it's still okay at the moment 🤷🏼‍♂️ while node 18 is no longer supported, it can happen that they didn't explicitly switch to node 20 APIs yet.

PR #10760, opened 2024-03-09
feat: add support for RETURNING clause to sqlite for nodejs

olosegres:
### Description of change

Closes #8667 

### Pull-Request Checklist

- [x] Code is up-to-date with the `master` branch
- [x] `npm run format` to apply prettier formatting
- [x] `npm run test` passes with this change
- [x] This pull request links relevant issues as `Fixes #0000`
- [ ] There are new or updated unit tests validating the change
- [ ] Documentation has been updated to reflect this change N/A
- [x] The new commits follow conventions explained in [CONTRIBUTING.md](https://github.com/typeorm/typeorm/blob/master/CONTRIBUTING.md)

PR #11312, opened 2025-02-28
feat: add support for RETURNING clause to sqlite

lagnat:
This is https://github.com/typeorm/typeorm/pull/10760 but with a fix for the failing test(s).  My changes are in AbstractSqliteDriver.createGeneratedMap().  Most of my changes were copied from other drivers.  I couldn't make sense of the existing code.

I also bumped better-sqlite3 to the latest version.


package.json
```diff
@@ -107,7 +107,7 @@
     "@types/source-map-support": "^0.5.6",
     "@types/uuid": "^9.0.0",
     "@types/yargs": "^17.0.22",
-    "better-sqlite3": "^8.1.0",
+    "better-sqlite3": "^11.8.1",
```
alumni, line 110: Let's leave `package.json` unchanged for now, we still need `better-sqlite3@8` in `typeorm@0.3`, but we'll definitely bump it up to 11+ when we drop old NodeJs support (soon).

Currently we have `better-sqlite3@8.7.0` which uses SQLite `3.43.1`, that should be good enough for the tests (if you update the PR).

src/driver/better-sqlite3/BetterSqlite3Driver.ts
```diff
@@ -48,6 +51,15 @@ export class BetterSqlite3Driver extends AbstractSqliteDriver {
     // Public Methods
     // -------------------------------------------------------------------------
 
+    async connect() {
+        await super.connect()
+
+        this.version = this.databaseConnection
```
alumni, line 57: We usually query the version using `QueryRunner.getVersion()`.

Maybe it could be done generically for all SQLite-based drivers.

src/driver/sqlite/SqliteDriver.ts
```diff
@@ -233,4 +236,9 @@ export class SqliteDriver extends AbstractSqliteDriver {
                 : path.join(process.cwd(), optionsDb),
         )
     }
+
+    isReturningSqlSupported() {
```
alumni, line 240: Could also be implemented generically in `AbstractSqliteDriver`

src/driver/sqlserver/SqlServerDriver.ts
```diff
@@ -891,7 +891,11 @@ export class SqlServerDriver implements Driver {
     /**
      * Returns true if driver supports RETURNING / OUTPUT statement.
      */
-    isReturningSqlSupported(): boolean {
+    isReturningSqlSupported(returningType: ReturningType): boolean {
+        if (returningType === 'insert') {
+          return false;
```
alumni, line 896: SqlServer supports "insert returning" ("insert output") and TypeORM already implements it. So this change can be undone.

test/github-issues/8667/issue-8667.ts
```diff
@@ -0,0 +1,85 @@
...
+
+        post = new Post()
+        post.title = "title"
+        post.text = "text"
+    })
+
+    beforeEach(() => reloadTestingDatabases(connections))
+    after(() => closeTestingConnections(connections))
+
+    it("should return all columns via QueryRunner", async () =>
```
alumni, line 31: Would be nice if you could update `test/functional/query-builder/returning/query-builder-returning.ts` - you can simply add the drivers and update the expected queries for SQLite-like drivers.

test/github-issues/8667/issue-8667.ts
```diff
@@ -0,0 +1,85 @@
...
+                    `UPDATE ${tableName} SET title = 'updated title value' RETURNING *`,
+                )
+
+                expect(res).to.be.deep.equal([
+                    { id: 1, title: "updated title value", text: "text" },
+                ])
+            }),
+        ))
+
+    it("should return user requested columns via QueryBuilder", () =>
```
alumni, line 57: Would be nice if you could extract this test to `test/functional/query-builder/returning/query-builder-returning-columns.ts`.

Since this is almost identical to `test/github-issues/7100/issue-7100.ts`, they could be combined (7100 can be copied, sqlite drivers added, and the missing assertion after insert can be added like in this test).

Issue #3591, opened 2019-02-07
Revert multiple migrations


lhrolim:
**Issue type:**

[ ] question
[ ] bug report
[x ] feature request
[ ] documentation issue

**Database system/driver:**

[ ] `cordova`
[ ] `mongodb`
[ ] `mssql`
[ ] `mysql` / `mariadb`
[ ] `oracle`
[x ] `postgres`
[ ] `sqlite`
[ ] `sqljs`
[ ] `react-native`
[ ] `expo`

**TypeORM version:**

[x ] `latest`
[ ] `@next`
[ ] `0.x.x` (or put your version here)

**Steps to reproduce or a small repository showing the problem:**

I think it would be very handy to provide high level method to rollback an entire "migration operation"

The motivation is simple: To easily allow rolling back an app to its previous version. 

The first thing that comes to mind that I think would easily solve the problem would be to add an extra column called version and then allow for reverting all the migrations that match that given version. 

It would be the responsibility of the application to inform the version in which the migration was applied.

I would then create a separate method for the cli called rollback that would receive a version and would in turn revert all the migrations which were applied in that version.

I´m very new to the project but I could try to provide a PR for that matter.

Kononnable:
We've already got `undoLastMigration()`. You want to add a new method with parameter `n` which would run `undoLastMigration()` n times or did I misunderstood you? `n` can be changed to date, target migration name etc.

lhrolim:
Hi @Kononnable I think it would be easier to allow controlling the migrations by a version number instead because for big projects it would be hard to know exactly how many migrations would need to by reverted if a given application needed to be rolled back.

Thus, my idea was to have an extra column `appversion` so that one could easily determine which migrations would need to be reverted in case an application migration fails.

Let´s say, we´re migrating a system from version 1.1.0 to 1.2.0. With that extra column and method, in case the application fails we could tell to undo all the migrations whose version matches 1.2.0. 

We could even allow for rolling back anything gt than 1.1.0, considering a default version syntax mechanism.

Only the the `undoLastMigration` method makes a rollback process error prone, since we´d need to somehow count or mark which migrations were already present at the database and calculate a diff considering which ones are there after the migration process to only then call it n times.

It´s also much easier to automate such a rollback process if we had something as a version:

We´ll try to release version 1.2.0 so we would know beforehand which migrations to rollback (i.e invoke the new method revertMigrations('1.2.0') on a rollback script

Kononnable:
[...]

PR #11315, opened 2025-03-03
Add option to revert multiple migrations

jovanadjuric:
### Description of change

Closes https://github.com/typeorm/typeorm/issues/3591

### Pull-Request Checklist

### DIFF
--- a/src/data-source/DataSource.ts
+++ b/src/data-source/DataSource.ts
@@ -407,6 +407,20 @@ export class DataSource {
     /**
      * Reverts last executed migration.
      * Can be used only after connection to the database is established.
+     *
+     * @deprecated use `revertMigration` instead
+     */
+    async undoLastMigration(options?: {
+        transaction?: "all" | "none" | "each"
+        fake?: boolean
+        until?: string
+    }): Promise<void> {
+        await this.revertMigration(options)
+    }
+
+    /**
+     * Reverts last executed migration.
+     * Can be used only after connection to the database is established.
      */
     async revertMigration(options?: {
         transaction?: "all" | "none" | "each"
--- a/src/migration/MigrationExecutor.ts
+++ b/src/migration/MigrationExecutor.ts
@@ -434,7 +434,7 @@ export class MigrationExecutor {
         // if no migrations found in the database then nothing to revert
         if (!toRevert.length) {
             this.connection.logger.logSchemaBuild(
-                `No migrations were found. Nothing to revert!`,
+                `No migrations were found in the database. Nothing to revert!`,
             )
             return
         }
@@ -451,9 +451,7 @@ export class MigrationExecutor {
             if (!match) {
                 // if any of the migrations is found in the database, but not in the source code, throw error to prevent inconsistencies
                 throw new TypeORMError(
-                    `Migration ${
-                        m.name ?? ""
-                    } was not found in the source code. Make sure you have this migration in your codebase and its included in the connection options.`,
+                    `Migration ${m.name} was not found in the source code. Make sure you have this migration in your codebase and its included in the connection options.`,
                 )
             }
 


docs/migrations.md
```diff
@@ -173,7 +173,7 @@ typeorm migration:revert -- -d path-to-datasource-config
 ```
 
 This command will execute `down` in the latest executed migration.
-If you need to revert multiple migrations you must call this command multiple times.
+If you want to revert multiple migrations, you can pass `--until <migration_name>` (or `-u`), which will revert all migrations executed after the specified one. To revert all migrations, pass "0".
```
naorpeled, line 176: Maybe instead 0 add a --all flag, wdyt?

src/migration/MigrationExecutor.ts
```diff
@@ -388,9 +388,11 @@ export class MigrationExecutor {
     }
 
     /**
-     * Reverts last migration that were run.
+     * If until is undefined, it reverts only the last executed migration.
+     * If until is "0", it reverts all migrations.
+     * Otherwise, it reverts all migrations executed after the migration with the specified name.
      */
-    async undoLastMigration(): Promise<void> {
+    async revertMigration(until?: string): Promise<void> {
```
naorpeled, line 395: Personally I think object arguments are more preferable as it allows extending the parameters without breaking usages all over the codebase. Wdyt?
This is not crucial or a blocker btw, feel free to ignore

src/migration/MigrationExecutor.ts
```diff
@@ -408,14 +410,31 @@ export class MigrationExecutor {
...
+
+            toRevert = lastTimeExecutedMigration
+                ? [lastTimeExecutedMigration]
+                : []
+        }
 
         // if no migrations found in the database then nothing to revert
-        if (!lastTimeExecutedMigration) {
+        if (!toRevert.length) {
             this.connection.logger.logSchemaBuild(
```
naorpeled, line 436: I think we should keep in the database as it's the source of truth, wdyt?

src/data-source/DataSource.ts
```diff
@@ -408,9 +408,10 @@ export class DataSource {
      * Reverts last executed migration.
      * Can be used only after connection to the database is established.
      */
-    async undoLastMigration(options?: {
+    async revertMigration(options?: {
```
naorpeled, line 411: Let's keep the previous name as well and deprecate it as it's a breaking change.

src/migration/MigrationExecutor.ts
```diff
@@ -424,52 +443,139 @@ export class MigrationExecutor {
...
+        const migrationsToRevert = toRevert.map((m) => {
+            const match = allMigrations.find(
+                (migration) => migration.name === m.name,
             )
 
+            if (!match) {
+                // if any of the migrations is found in the database, but not in the source code, throw error to prevent inconsistencies
+                throw new TypeORMError(
+                    `Migration ${
+                        m.name ?? ""
```
naorpeled, line 455: Maybe worth throwing a more specific error if a migration doesn't have a name.
I wonder how that would be possible 🤔

src/migration/MigrationExecutor.ts
```diff
@@ -424,52 +443,139 @@ export class MigrationExecutor {
...
+        if (this.transaction === "all") {
+            // If we desire to revert all migrations in a single transaction
+            // but there is a migration that explicitly overrides the transaction mode
+            // then we have to fail since we cannot properly resolve that intent
+            // In theory we could support overrides that are set to `true`,
+            // however to keep the interface more rigid, we fail those too
+            const migrationsOverridingTransactionMode =
+                migrationsToRevert.filter(
+                    (migration) =>
+                        !(migration.instance?.transaction === undefined),
```
naorpeled, line 489:
```suggestion
                        migration.instance?.transaction !== undefined,
```

naorpeled: OR

```suggestion
                        migration.instance?.transaction != null,
```

src/migration/MigrationExecutor.ts
```diff
@@ -424,52 +443,139 @@ export class MigrationExecutor {
...
+                throw error
+            }
+        }
+
+        // Set the per-migration defaults for the transaction mode
+        // so that we have one centralized place that controls this behavior
+
+        // When transaction mode is `each` the default is to run in a transaction
+        // When transaction mode is `none` the default is to not run in a transaction
+        // When transaction mode is `all` the default is to not run in a transaction
```
naorpeled, line 508: wait, `all` is not for running all in a transaction? 🤔

src/migration/MigrationExecutor.ts
```diff
@@ -424,52 +443,139 @@ export class MigrationExecutor {
...
+                    this.connection.logger.logMigration(
+                        `Reverting of "${migration.name}" failed, error: ${error?.message}`,
+                    )
+                    throw error
+                })
+
+                await this.deleteExecutedMigration(queryRunner, migration)
+
+                // commit transaction if we started it
+                if (migration.transaction && transactionStartedByUs) {
```
naorpeled, line 562:
```suggestion
                if (migration.transaction === 'each' && transactionStartedByUs) {
```

test/functional/migrations/revert-command/command.ts
```diff
@@ -0,0 +1,86 @@
+import { expect } from "chai"
```
naorpeled, line 1: let's also add tests for the 'each' transaction case and make sure we start and commit a transaction for each migration if there aren't any

naorpeled:
Hey @jovanadjuric,
first of all great job, looks awesome!

Left some thoughts, let me know what you think 🙏

docs/migrations.md
jovanadjuric: Sounds good. In case both `--until` and `--all` are provided, would it make sense to ignore `--until`?

src/migration/MigrationExecutor.ts
jovanadjuric: Forgot to remove this. That's right - it is not possible for a migration not to have a name

Great catch! :1st_place_medal:

src/migration/MigrationExecutor.ts
```diff
@@ -424,52 +443,139 @@ export class MigrationExecutor {
...
 
+        // revert the migrations sequentially
         try {
-            if (!this.fake) {
-                await queryRunner.beforeMigration()
-                await migrationToRevert.instance!.down(queryRunner)
-                await queryRunner.afterMigration()
-            }
+            for (const migration of migrationsToRevert) {
+                if (this.fake) {
```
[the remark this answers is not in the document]
jovanadjuric: I moved them to transaction-start and transaction-end blocks :1234: 

image

src/data-source/DataSource.ts
jovanadjuric: Done, it is now a wrapper for `revertMigration`

jovanadjuric:
> Hey @jovanadjuric, first of all great job, looks awesome!
> 
> Left some thoughts, let me know what you think 🙏

Thank you for your kind words and for taking the time to go through this so thoroughly!

I pushed a commit that fixes a few of the comments. 

In a day or two I'll look into how exactly TypeORM handles transactions in other places, and after that I'll ping you on Discord. Just want to go over the comments about them with you since I used the logic from `MigrationExecutor#executePendingMigrations` as a reference. If something needs to be updated, it likely needs updating there as well.

src/migration/MigrationExecutor.ts
sgarner: `migration.transaction` is a boolean that can be set on the `Migration` class itself:
https://typeorm.io/migrations#transaction-modes

(Not the transaction mode set in `MigrationExecutor`)

naorpeled:
> > Hey @jovanadjuric, first of all great job, looks awesome!
> > 
> > Left some thoughts, let me know what you think 🙏
> 
> Thank you for your kind words and for taking the time to go through this so thoroughly!
> 
> I pushed a commit that fixes a few of the comments. 
> 
> In a day or two I'll look into how exactly TypeORM handles transactions in other places, and after that I'll ping you on Discord. Just want to go over the comments about them with you since I used the logic from `MigrationExecutor#executePendingMigrations` as a reference. If something needs to be updated, it likely needs updating there as well.

Sounds good 😎

docs/migrations.md
sgarner: I'm not sure there's a compelling use case for `--all`? I'd suggest that `--until` alone is sufficient, you can always revert all by specifying the name of the first migration anyway.

jovanadjuric: @sgarner, `--until` reverts everything executed after the specified migration, but not the specified migration itself. That makes reverting the first migration impossible without running `migration:revert --until <last>`, and then `migration:revert` separately

I added "--until 0" to make that possible, as like [MikroORM](https://mikro-orm.io/docs/migrations#:~:text=npx%20mikro%2Dorm%20migration%3Adown%20%2D%2Dto%200%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%23%20migrate%20down%20all%20migrations) handles it

src/data-source/DataSource.ts
```diff
@@ -407,10 +407,25 @@ export class DataSource {
...
+        until?: string
+    }): Promise<void> {
+        await this.revertMigration(options)
+    }
+
+    /**
+     * Reverts last executed migration.
+     * Can be used only after connection to the database is established.
+     */
+    async revertMigration(options?: {
```
sgarner, line 425: If I can make a few suggestions:

- this method should be named `revertMigrations` (plural) for clarity
- the `until` parameter should be a required argument
- the `until` option should be inclusive rather than exclusive. That is, the named migration should be reverted as well as those after it. This will be easier to use?
- the `undoLastMigration` method should not take an `until` option. It should look up the name of the last migration and pass it to `revertMigrations`

docs/migrations.md
jovanadjuric: I just read your suggestions. Making `--until` inclusive would fix this problem. I'll check how other ORMs handle this and get back to you. Thanks for your feedback!

sgarner: Per my other comment - I think we should make `until` inclusive as well. So the user specifies the name of the earliest migration to be reverted, rather than the last one they want to keep.

Imagine you're using this to revert a sequence of migrations, it will be simpler for the user to specify the name of the first unwanted migration in the sequence rather than having to figure out what was the last good migration. IMHO

Also we don't need to blindly copy whatever MikroORM does, let's make our own design choices. 😎  In MikroORM they use `--from/--to` arguments referencing numeric migration datestamps, so passing `0` to rewind to the beginning is logical - but we use string migration names.

src/migration/MigrationExecutor.ts
alumni: Like most of the projects nowadays, we use strict equality checks (`===` and `!==` instead of `==` and `!=`) 😉

Not sure if it's enforced with the eslint rule `eqeqeq`, I think we just do it.

src/migration/MigrationExecutor.ts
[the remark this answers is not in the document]
alumni: Before, the events were triggered for each migration, now they are triggered only once. This probably should not be change.

Also, they should probably be triggered within the transaction, not outside of it.

docs/migrations.md
alumni: Honestly, I'm not a big fan of this `--until` flag, that's why I was asking if this was discussed somewhere. The migration names are quite long, I wouldn't want to write the full migration name. I was thinking maybe `--last 5` or `-n 5` would be simpler.

‼️ What bothers me the most: it often happens (extremely often for our project) that the order in which the migrations have been executed (i.e. in the migrations table) is not the alphabetical order (sometimes even in production!), because it can take time for a PR to get merged, so the migrations might end up being executed in a non-alphabetical order. There aren't any dependencies between these unsorted migrations, but what you see is NOT what you get - you might revert the wrong migrations.

Just thinking out loud: would user interaction improve things? E.g. ask the user to accept the list of migrations that will be reverted.

src/migration/MigrationExecutor.ts
sgarner: FWIW I am personally a fan of using loose equality (`== null` or `!= null`) to check for the absence of a value in cases where you don't care to distinguish between `null` and `undefined` and where both values could occur.

Unlike loose equality comparisons involving other types, which can be treacherous, `null` is only loosely equal to itself and `undefined` so this is quite safe.

By contrast, rigidly using only strict equality requires that you remember to check for both values (`x === null || x === undefined`) and so is more likely to lead to errors as well as being less concise.

The `eqeqeq` eslint rule has [an option](https://eslint.org/docs/latest/rules/eqeqeq#allow-null) to support this convention. It looks like we aren't enabling `eqeqeq` at all in TypeORM currently though.

src/migration/MigrationExecutor.ts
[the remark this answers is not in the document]
jovanadjuric: Transaction logic in this PR was copied from the `MigrationExecutor#executePendingMigrations` https://github.com/typeorm/typeorm/blob/27b4207c48a391c7bf7013b251f756f48861c496/src/migration/MigrationExecutor.ts#L191-L388

It makes sense that `migration:run` and `migration:revert` handle transactions the same way. If something isn't right, it needs to be fixed in both places and as a part of another PR.

> Also, they should probably be triggered within the transaction, not outside of it.

`beforeMigration` and `afterMigration` hooks are executed outside the transaction there as well.

___

My idea was to manually verify that both scripts behave the same and that there are no deviations from the original behavior of the `revert` command. After that, I figured we could submit another PR with steps 1-4 where transactions are addressed properly in both places. I would also carefully check if the behavior I added in this PR differentiates in any way from the original behavior for the single `migration:revert` mode.

The steps are:
1. Write tests for `MigrationRunCommand`
2. As @naorpeled suggested, cover each transaction mode case with tests in both `run` and `revert`
3. If transactions are not working as expected, now it's time to fix whatever is not right
4. We can also optionally improve the code structure

TL;DR: I believe that transactions should behave the same way in both `run` and `revert`, and that's why I'm hesitant to make any changes to them in this PR.

src/migration/MigrationExecutor.ts
jovanadjuric: concerns https://github.com/typeorm/typeorm/pull/11315#discussion_r1984201466

test/functional/migrations/revert-command/command.ts
jovanadjuric: concerns https://github.com/typeorm/typeorm/pull/11315#discussion_r1984201466

src/migration/MigrationExecutor.ts
```diff
@@ -424,52 +443,139 @@ export class MigrationExecutor {
...
-            await this.deleteExecutedMigration(queryRunner, migrationToRevert)
-            this.connection.logger.logSchemaBuild(
-                `Migration ${migrationToRevert.name} has been ${
-                    this.fake ? "(fake)" : ""
-                } reverted successfully.`,
-            )
+                    continue
+                }
+
+                if (migration.transaction && !queryRunner.isTransactionActive) {
```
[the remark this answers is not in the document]
jovanadjuric: https://github.com/typeorm/typeorm/pull/11315#discussion_r1978316538

docs/migrations.md
sgarner: I think we have to look at the order of migration files on disk for this purpose, not at what's in the migration table. If you've switched branches there might be migrations in the table which don't even exist on disk. So the code has to be the source of truth here.

Whether the migrations were previously executed out of order becomes irrelevant, the current file list must determine the order in which they're reverted.

I prefer referencing the migration name in --until for safety. It may be hard to type, but I'd be copy/pasting it anyway. I don't mind if we allow reverting by numeric offset as well (e.g. "2" to revert the last two migrations) if we want to offer both ways.

The naming system for migrations in TypeORM is awkward, but that could be another topic!

Interactively confirming the list of migrations to be reverted would be a great feature, but I'm not sure if the typeorm CLI has any functionality for doing that currently? If not I'd say it's out of scope for this PR.

ali-zali-dev:
Hi @jovanadjuric, thanks for your work on this PR
I noticed there hasn’t been any update since March. If you're no longer planning to continue working on this, I’d be happy to pick it up and open a new PR based on your work (with proper attribution, of course) and fix the remaining items.

Let me know if that works for you; feel free to continue if you’re still on it.

jovanadjuric:
Hey @ali-zali-dev, thanks for reaching out! I'm currently tied up, so feel free to pick it up from where I left off :)

Issue #10964, opened 2024-07-04
Add support of "hash" indexes for postgres


pYassine:
### Feature Description

In PostgreSQL, hash indexes can be highly efficient for certain types of lookups, especially for equality comparisons. 

However, currently, TypeORM does not provide a straightforward way to create hash indexes through its decorators or schema synchronization features. 

This limitation requires developers to manually manage hash indexes via raw SQL queries in migrations

### The Solution

It would be beneficial if TypeORM could natively support the creation of hash indexes for PostgreSQL. This could be achieved by extending the `@Index` decorator or providing a new option within the `@Column` decorator to specify the type of index.

Example:
```typescript
@Entity()
export class Structure {
  @PrimaryGeneratedColumn()
  id: number;

  @Column({ type: 'text', nullable: false })
  @Index({ type: 'hash' })
  public region: string;
}
```

### Considered Alternatives

Create indexes one by one with custom queries

### Relevant Database Driver(s)

- [ ] aurora-mysql
- [ ] aurora-postgres
- [ ] better-sqlite3
- [ ] cockroachdb
- [ ] cordova
- [ ] expo
- [ ] mongodb
- [ ] mysql
- [ ] nativescript
- [ ] oracle
- [X] postgres
- [ ] react-native
- [ ] sap
- [ ] spanner
- [ ] sqlite
- [ ] sqlite-abstract
- [ ] sqljs
- [ ] sqlserver

### Are you willing to resolve this issue by submitting a Pull Request?

No, I don’t have the time and I’m okay to wait for the community / maintainers to resolve this issue.

PR #11318, opened 2025-03-04
feat: add support for PostgreSQL indices

freePixel:
### Description of change

Implements #10964 and adds support for all PostgreSQL index types  
(B-Tree, Hash, GiST, SP-GiST, GIN, BRIN) as described in the [PostgreSQL documentation](https://www.postgresql.org/docs/current/indexes.html).

A new optional `type` option of type `TableIndexTypes` has been added to the `Index` decorator.

The implementation is designed in a way that allows it to be easily extended to support other database drivers beyond PostgreSQL.

### Pull-Request Checklist

### DIFF
--- a/src/driver/postgres/PostgresDriver.ts
+++ b/src/driver/postgres/PostgresDriver.ts
@@ -1402,6 +1402,13 @@ export class PostgresDriver implements Driver {
         return false
     }
 
+    /**
+     * Returns true if driver supports type indices
+     */
+    isIndicesTypeSupported(): boolean {
+        return true
+    }
+
     get uuidGenerator(): string {
         return this.options.uuidExtension === "pgcrypto"
             ? "gen_random_uuid()"
@@ -1413,6 +1420,16 @@ export class PostgresDriver implements Driver {
      */
     createParameter(parameterName: string, index: number): string {
         return this.parametersPrefix + (index + 1)
+    }
+
+    compareTableIndexTypes = (
+        indexA: string | undefined,
+        indexB: string | undefined,
+    ) => {
+        const normalizedA = indexA ?? "btree"
+        const normalizedB = indexB ?? "btree"
+
+        return normalizedA.toLowerCase() === normalizedB.toLowerCase()
     }
 
     // -------------------------------------------------------------------------
--- a/src/driver/postgres/PostgresQueryRunner.ts
+++ b/src/driver/postgres/PostgresQueryRunner.ts
@@ -3994,6 +3994,7 @@ export class PostgresQueryRunner
                         isUnique: constraint["is_unique"] === "TRUE",
                         where: constraint["condition"],
                         isSpatial: constraint["index_type"] === "gist",
+                        type: constraint["index_type"],
                         isFulltext: false,
                     })
                 })
@@ -4299,9 +4300,37 @@ export class PostgresQueryRunner
     }
 
     /**
+     * Builds the SQL `USING <index_type>` clause based on the index type, prioritizing `isSpatial` as `GiST`.
+     */
+
+    private buildIndexTypeClause(index: TableIndex) {
+        // List of index types supported by PostgreSQL
+        // https://www.postgresql.org/docs/current/indexes-types.html
+        let pgValidIdxTypes = new Set<string>([
+            "btree",
+            "hash",
+            "gist",
+            "spgist",
+            "gin",
+            "brin",
+        ])
+
+        let type = index.isSpatial ? "gist" : index.type
+
+        if (typeof type !== "string") return null
+
+        if (!pgValidIdxTypes.has(type))
+            throw new TypeORMError(`Unsupported index type`)
+
+        return `USING ${type}`
+    }
+
+    /**
      * Builds create index sql.
      */
     protected createIndexSql(table: Table, index: TableIndex): Query {
+        const indexTypeClause = this.buildIndexTypeClause(index)
+
         const columns = index.columnNames
             .map((columnName) => `"${columnName}"`)
             .join(", ")
@@ -4309,7 +4338,7 @@ export class PostgresQueryRunner
             `CREATE ${index.isUnique ? "UNIQUE " : ""}INDEX${
                 index.isConcurrent ? " CONCURRENTLY" : ""
             } "${index.name}" ON ${this.escapePath(table)} ${
-                index.isSpatial ? "USING GiST " : ""
+                indexTypeClause ?? ""
             }(${columns}) ${index.where ? "WHERE " + index.where : ""}`,
         )
     }
--- a/src/metadata/IndexMetadata.ts
+++ b/src/metadata/IndexMetadata.ts
@@ -4,6 +4,7 @@ import { NamingStrategyInterface } from "../naming-strategy/NamingStrategyInterf
 import { ColumnMetadata } from "./ColumnMetadata"
 import { EmbeddedMetadata } from "./EmbeddedMetadata"
 import { TypeORMError } from "../error"
+import { TableIndexTypes } from "../schema-builder/options/TableIndexTypes"
 
 /**
  * Index metadata contains all information about table's index.
@@ -120,6 +121,13 @@ export class IndexMetadata {
     where?: string
 
     /**
+     * The `type` option defines the type of the index being created.
+     * Supported types include B-tree, Hash, GiST, SP-GiST, GIN, and BRIN
+     * This option is only applicable in PostgreSQL.
+     */
+    type?: TableIndexTypes
+
+    /**
      * Map of column names with order set.
      * Used only by MongoDB driver.
      */
@@ -158,6 +166,7 @@ export class IndexMetadata {
             this.expireAfterSeconds = options.args.expireAfterSeconds
             this.givenName = options.args.name
             this.givenColumnNames = options.args.columns
+            this.type = options.args.type
         }
     }
 
--- a/src/schema-builder/RdbmsSchemaBuilder.ts
+++ b/src/schema-builder/RdbmsSchemaBuilder.ts
@@ -399,6 +399,17 @@ export class RdbmsSchemaBuilder implements SchemaBuilder {
                             return true
 
                         if (
+                            this.connection.driver.isIndicesTypeSupported() &&
+                            typeof this.connection.driver
+                                .compareTableIndexTypes === "function" &&
+                            !this.connection.driver.compareTableIndexTypes(
+                                indexMetadata.type,
+                                tableIndex.type,
+                            )
+                        )
+                            return true
+
+                        if (
                             this.connection.driver.isFullTextColumnTypeSupported() &&
                             indexMetadata.isFulltext !== tableIndex.isFulltext
                         )
--- a/src/schema-builder/options/TableIndexTypes.ts
+++ b/src/schema-builder/options/TableIndexTypes.ts
@@ -1,0 +1,7 @@
+export type TableIndexTypes =
+    | "btree"
+    | "hash"
+    | "gist"
+    | "spgist"
+    | "gin"
+    | "brin"
--- a/test/github-issues/10964/issue-10964.ts
+++ b/test/github-issues/10964/issue-10964.ts
@@ -1,0 +1,38 @@
+import "reflect-metadata"
+import { expect } from "chai"
+
+import { DataSource } from "../../../src/data-source/DataSource"
+import {
+    closeTestingConnections,
+    createTestingConnections,
+    reloadTestingDatabases,
+} from "../../utils/test-utils"
+
+import { User } from "../10964/entity/User"
+
+describe("github issues > Add support of 'hash' indexes for postgres", () => {
+    let connections: DataSource[]
+    before(
+        async () =>
+            (connections = await createTestingConnections({
+                entities: [__dirname + "/entity/*{.js,.ts}"],
+                enabledDrivers: ["postgres"],
+            })),
+    )
+    beforeEach(() => reloadTestingDatabases(connections))
+    after(() => closeTestingConnections(connections))
+
+    it("should support 'hash' index", () =>
+        Promise.all(
+            connections.map(async (connection) => {
+                expect(connection.getMetadata(User).indices.find(idx => idx.type === 'hash')).not.null
+            }),
+    ))
+
+    it("User should have two indices", () =>
+        Promise.all(
+            connections.map(async (connection) => {
+                expect(connection.getMetadata(User).indices.length).equal(2)
+            }),
+    ))
+})
--- a/docs/indices.md
+++ b/docs/indices.md
@@ -150,6 +150,15 @@ Typeorm supports generating SQL with this option if when the concurrent option i
 
 For more information see the [postgres documentation](https://www.postgresql.org/docs/current/sql-createindex.html).
 
+## Index Type
+If you need to specify a custom type for the index, you can use the `type` property. If the `spatial` property is set, this field will be ignored.
+
+```typescript
+@Index({ type: 'hash' })
+```
+
+This feature is currently supported only for PostgreSQL.
+
 ## Disabling synchronization
 
 TypeORM does not support some index options and definitions (e.g. `lower`, `pg_trgm`) because of lot of different database specifics and multiple
--- a/src/decorator/Index.ts
+++ b/src/decorator/Index.ts
@@ -132,6 +132,7 @@ export function Index(
                     ? false
                     : true,
             where: options ? options.where : undefined,
+            type: options ? options.type : undefined,
             unique: options && options.unique ? true : false,
             spatial: options && options.spatial ? true : false,
             fulltext: options && options.fulltext ? true : false,
--- a/src/decorator/options/IndexOptions.ts
+++ b/src/decorator/options/IndexOptions.ts
@@ -1,3 +1,5 @@
+import { TableIndexTypes } from "../../schema-builder/options/TableIndexTypes"
+
 /**
  * Describes all index options.
  */
@@ -63,4 +65,11 @@ export interface IndexOptions {
      * This option is only supported for mongodb database.
      */
     expireAfterSeconds?: number
+
+    /**
+     * The `type` option defines the type of the index being created.
+     * Supported types include B-tree, Hash, GiST, SP-GiST, GIN, and BRIN
+     * This option is only applicable in PostgreSQL.
+     */
+    type?: TableIndexTypes
 }
--- a/src/driver/Driver.ts
+++ b/src/driver/Driver.ts
@@ -282,4 +282,17 @@ export interface Driver {
      * Creates an escaped parameter.
      */
     createParameter(parameterName: string, index: number): string
+
+    /**
+     * Returns true if driver supports type indices
+     */
+    isIndicesTypeSupported(): boolean
+
+    /**
+     * Returns true if both indexes types are equivalent
+     */
+    compareTableIndexTypes?: (
+        indexA: string | undefined,
+        indexB: string | undefined,
+    ) => boolean
 }
--- a/src/driver/aurora-mysql/AuroraMysqlDriver.ts
+++ b/src/driver/aurora-mysql/AuroraMysqlDriver.ts
@@ -1002,6 +1002,13 @@ export class AuroraMysqlDriver implements Driver {
         return "?"
     }
 
+    /**
+     * Returns true if driver supports type indices
+     */
+    isIndicesTypeSupported(): boolean {
+        return false
+    }
+
     // -------------------------------------------------------------------------
     // Protected Methods
     // -------------------------------------------------------------------------
--- a/src/driver/cockroachdb/CockroachDriver.ts
+++ b/src/driver/cockroachdb/CockroachDriver.ts
@@ -980,6 +980,13 @@ export class CockroachDriver implements Driver {
     }
 
     /**
+     * Returns true if driver supports type indices
+     */
+    isIndicesTypeSupported(): boolean {
+        return false
+    }
+
+    /**
      * Creates an escaped parameter.
      */
     createParameter(parameterName: string, index: number): string {
--- a/src/driver/mongodb/MongoDriver.ts
+++ b/src/driver/mongodb/MongoDriver.ts
@@ -488,6 +488,13 @@ export class MongoDriver implements Driver {
     }
 
     /**
+     * Returns true if driver supports type indices
+     */
+    isIndicesTypeSupported(): boolean {
+        return false
+    }
+
+    /**
      * Creates an escaped parameter.
      */
     createParameter(parameterName: string, index: number): string {
--- a/src/driver/mysql/MysqlDriver.ts
+++ b/src/driver/mysql/MysqlDriver.ts
@@ -1154,6 +1154,13 @@ export class MysqlDriver implements Driver {
     }
 
     /**
+     * Returns true if driver supports type indices
+     */
+    isIndicesTypeSupported(): boolean {
+        return false
+    }
+
+    /**
      * Creates an escaped parameter.
      */
     createParameter(parameterName: string, index: number): string {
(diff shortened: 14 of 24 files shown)


jarpoole (approved):

jarpoole:
Awesome works guys, can't wait for this to land as I'd love to be able to declare full-text search indices in TypeORM!

OSA413:
I would also add a test to see what happens with adding an index to unsupported DB

TiagoSilvestre1:
> I would also add a test to see what happens with adding an index to unsupported DB

Thanks for pointing that out!  
I made two minor changes to this feature:  
1. Throw an error if 'type' is used with an unsupported driver.  
2. Fixed an issue where the 'gist' index was being recreated even without a schema change.

I also improved the test cases based on your comments.

OSA413 (approved):
Just a few questions, why `gist` is `spatial`, what is `spatial`, and why we fall back into `btree` in the compare method in `PostgresDriver.ts`?

OSA413 (approved):

TiagoSilvestre1:
@OSA413 Apparently, some tests were failing because I created a test that didn't properly check whether the SQLite driver was selected. I'm still new to this project, and I'm just trying to improve it! I hope you understand my newbie mistakes 😅 .

> Just a few questions: Why is `gist` considered `spatial`? What is `spatial`? And why do we fall back to `btree` in the `compare` method in `PostgresDriver.ts`?

With this PR, I didn't intend to introduce breaking changes. Currently, in (typeorm:master), when an index has the `isSpatial` property set, a PostgreSQL index of type `gist` is created ([as you can see here](https://github.com/typeorm/typeorm/blob/master/src/driver/postgres/PostgresQueryRunner.ts#L4304)). I wanted to maintain this behavior.

A [Gist](https://www.postgresql.org/docs/current/indexes-types.html#INDEXES-TYPE-GIST) index is used for 2D geometric data types, which is where "spatial" comes from.


`Btree` is the default index type for PostgreSQL, according to the PostgreSQL documentation:  
image  
Source: [PostgreSQL Documentation](https://www.postgresql.org/docs/current/sql-createindex.html)
That's why I consider false values as a B-Tree index because they are equivalent to nothing.

test/github-issues/10964/issue-10964.ts
alumni, line 1: Should be a functional test, especially since it's a new feature

freePixel: I moved the tests to the functional testing path.

test/functional/indices/index-types/index-types.ts
```diff
@@ -0,0 +1,122 @@
...
+                expect(connection.getMetadata(User).indices.length).equal(6)
+            }),
+        ))
+})
+
+describe("github issues > Add support of 'hash' indexes for postgres", async () => {
+    it("Should throw an error if index type is set and driver does not support index types", async () => {
+        const connections = await createTestingConnections({
+            entities: [User3],
+            enabledDrivers: ["sqlite"],
```
OSA413: I think this should be extended to the other drivers that don't support index types.

test/functional/indices/index-types/index-types.ts
```diff
@@ -0,0 +1,122 @@
...
+            expect(
+                createTestingConnections({
+                    entities: [User2],
+                    enabledDrivers: ["sqlite"],
+                    schemaCreate: true,
+                }),
+            ).rejectedWith(TypeORMError)
+        }
+    })
+})
```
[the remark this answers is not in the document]
OSA413: CodeRabbit suggests the right thing.

test/functional/indices/index-types/index-types.ts
freePixel: I've extended the test to include other drivers.

test/functional/indices/index-types/index-types.ts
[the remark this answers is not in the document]
freePixel: > CodeRabbit suggests the right thing.
Unfortunately, @coderabbitai suggestion is not correct. The test will fail when a driver other than "sqlite" is used. In that case, there will be no connections, and the function will not throw any error. It is necessary to check if the driver being tested matches the one expected in the code. That why I ignored the test when a different driver is used.
Furthermore, I updated the test to check other drivers.

OSA413:
Looks like the error wasn't thrown on the expected scenario (unsupported database), can you please investigate why's that?

And you also should fix formatting (it's the `npm run format` command).

src/schema-builder/RdbmsSchemaBuilder.ts
```diff
@@ -398,6 +399,25 @@ export class RdbmsSchemaBuilder implements SchemaBuilder {
...
+                            return true
+
+                        if (
+                            !this.connection.driver.isIndicesTypeSupported() &&
+                            indexMetadata.type
+                        )
+                            throw new TypeORMError(
+                                `Current database driver does not support index 'type' property`,
+                            )
+
```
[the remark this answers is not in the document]
OSA413: The question is, do we need to update indexes for such cases? After some thinking I guess that ~~this is needed to be duplicated/mirrored as CodeRabbit suggests~~. If I understand correctly, views don't store any data (including indexes), but the code is trying to update the related table. I think that this may break some things.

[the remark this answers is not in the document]
OSA413: My arguments are over, I'm requesting the help of the maintainers. (my knowledge of views is quite limited)

freePixel:
@OSA413 

> Looks like the error wasn't thrown on the expected scenario (unsupported database), can you please investigate why's that? 
And you also should fix formatting (it's the npm run format command).

I fixed the formatting and split the test into three separate tests for clarity.
I ran the tests locally with the SQLite driver and didn't encounter any issues.

OSA413:
I'll try to see what's wrong with the tests later this week

OSA413:
Looks like the tests are failing with the CLI tests, not the index types tests, this means that this should be good. (we will still investigate what's wrong with the tests)

src/driver/postgres/PostgresQueryRunner.ts
```diff
@@ -4345,18 +4346,46 @@ export class PostgresQueryRunner
...
     }
 
+    /**
+     * Builds the SQL `USING <index_type>` clause based on the index type, prioritizing `isSpatial` as `GiST`.
+     */
+
+    private buildIndexTypeClause(index: TableIndex) {
+        // List of index types supported by PostgreSQL
+        // https://www.postgresql.org/docs/current/indexes-types.html
+        const pgValidIdxTypes = new Set<string>([
```
OSA413: I would personally move the list to a separate file with info about the driver (if such file doesn't exist, then I would create one somewhere near other files of the specific driver)

But that's ok for now.

OSA413:
Btw is there any version differences about supporting specific index types? Like, currently we have tests for Postgres for 14 and 17 versions.

Upd: according to the documentation, 14 and 17 versions support the same index types.

OSA413 (approved):

freePixel:
> Btw is there any version differences about supporting specific index types? Like, currently we have tests for Postgres for 14 and 17 versions.
Upd: according to the documentation, 14 and 17 versions support the same index types.

@OSA413, these indexes were introduced in postgreSQL v14, and they remain the same in the current version (v18). If new ones are introduced in the future, the code will need to be updated to support them. If they aren't supported, the database will raise an error (which makes sense). If an index is ever removed from PostgreSQL (which I think it's very unlikely), the database will also throw an error.

These database errors are not a problem and are very easy to resolve if the current postgreSQL version does not support something. It's better to keep things simple and avoid overcomplicating TypeORM with custom logic for PostgreSQL indexes across postgreSQL versions. The library user will have access to the database error (e.g index type not recognized) and will have enough information to fix the issue.

src/driver/postgres/PostgresQueryRunner.ts
freePixel: I think it depends on personal preferences. I don't see any problem with current implementation nor with your proposal to move to another file.

docs/docs/advanced-topics/3-indices.md
```diff
@@ -144,6 +144,15 @@ TypeORM supports generating SQL with this option when the concurrent option is s
...
 For more information see the [Postgres documentation](https://www.postgresql.org/docs/current/sql-createindex.html).
 
+## Index Type
+If you need to specify a custom type for the index, you can use the `type` property. If the `spatial` property is set, this field will be ignored.
+
+```typescript
+@Index({ type: 'hash' })
+```
+
+This feature is currently supported only for PostgreSQL.
```
G0maa, line 154: Small nitpick: maybe we should open a new issue tracking index type support for other drivers? @OSA413 @freePixel

src/driver/postgres/PostgresQueryRunner.ts
G0maa: Hello,

We could see how's this done in other drivers, and mimic it :)

e.g. potentially added to the same `TableIndexTypes` file:
https://github.com/freePixel/typeorm/blob/8034aa5ad2180fa1c83cc17522c802a9a26a25d0/src/schema-builder/options/TableIndexTypes.ts

Also related, can't we reuse the same `TableIndexTypes` here?
i.e. `const pgValidIdxTypes = new Set<TableIndexTypes>([.....])`

src/driver/postgres/PostgresQueryRunner.ts
```diff
@@ -4345,18 +4346,46 @@ export class PostgresQueryRunner
...
+        ])
+
+        const type = index.isSpatial ? "gist" : index.type
+
+        if (typeof type !== "string") return null
+
+        if (!pgValidIdxTypes.has(type))
+            throw new TypeORMError(`Unsupported index type`)
+
+        return `USING ${type}`
```
G0maa, line 4360: If I don't specify a type for the index, does this still add `USING btree` to the query creating the index?

src/schema-builder/RdbmsSchemaBuilder.ts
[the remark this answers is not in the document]
G0maa: Hello @freePixel, I think this may need your attention.

test/functional/indices/index-types/entity/User2.ts
```diff
@@ -0,0 +1,16 @@
...
+
+@Entity()
+export class User2 {
+    @PrimaryGeneratedColumn()
+    id: number
+
+    @Column({ type: "varchar", length: 255 })
+    @Index({ type: "hash" })
+    hashColumn: string
+}
```
G0maa, line 16: Should possibly add a test where `@Index` is used on the entity class itself.

```ts
@Index(...)
@Entity()
export class User3 {}
```

docs/docs/advanced-topics/3-indices.md
freePixel: Yes, that makes sense.
For example mySQL is a good candidate where this could be implemented https://dev.mysql.com/doc/refman/8.4/en/create-index.html

src/driver/postgres/PostgresQueryRunner.ts
freePixel: > e.g. potentially added to the same TableIndexTypes file:
https://github.com/freePixel/typeorm/blob/8034aa5ad2180fa1c83cc17522c802a9a26a25d0/src/schema-builder/options/TableIndexTypes.ts

Agree with you

> Also related, can't we reuse the same TableIndexTypes here?

Yes, that makes sense, I'm gonna update that.

> If I don't specify a type for the index, does this still add USING btree to the query creating the index?

If you don't specify the index type the function will return null, and the index type will be the default postgreSQL index type which is btree.

test/functional/indices/index-types/entity/User2.ts
freePixel: There were already previous tests testing @Index at Entity level and I added one with tests 'type' at Entity level.

src/schema-builder/RdbmsSchemaBuilder.ts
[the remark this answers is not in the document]
freePixel: I was not aware of view indexes, I'll have to create some tests for it and verify why they are not being created for both @ViewColumn and @ViewEntity decorators.

G0maa:
Hello again @freePixel,

Could you resolve comments if they get fixed, and request a review again from me (or anyone).

If we take too much time (2+ business days) to reply could you post on Discord, and ping related reviewers.

I'd personally like to get this merged, it has been around for too long :grinning:

We appreciate your contribution!

freePixel:
> Hello again @freePixel,
> 
> Could you resolve comments if they get fixed, and request a review again from me (or anyone).
> 
> If we take too much time (2+ business days) to reply could you post on Discord, and ping related reviewers.
> 
> I'd personally like to get this merged, it has been around for too long 😀
> 
> We appreciate your contribution!

I'm not used to open source contributions, thanks for the advice :)

I'm almost done, I had to change some code to add support for postgreSQL view indices (I was not aware of them previously). I'm going to add some tests soon and then I think all points should be closed.

test/functional/indices/index-types/index-types.ts
```diff
@@ -0,0 +1,168 @@
...
+            expect(
+                createTestingConnections({
+                    entities: [User2],
+                    enabledDrivers: ["mariadb"],
+                    schemaCreate: true,
+                }),
+            ).rejectedWith(TypeORMError)
+        }
+    })
+})
```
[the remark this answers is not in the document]
freePixel: Although tests were working without await expression I've added them

test/functional/indices/index-types/index-types.ts
```diff
@@ -0,0 +1,168 @@
...
+            expect(
+                createTestingConnections({
+                    entities: [User2],
+                    enabledDrivers: ["mysql"],
+                    schemaCreate: true,
+                }),
+            ).rejectedWith(TypeORMError)
+        }
+    })
+})
```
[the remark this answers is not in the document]
freePixel: resolved

test/functional/indices/index-types/entity/User2.ts
freePixel: Done

alumni:
@freePixel FYI we were also planning to have vector indexes in the future, but we might implement them separately - we were actually discussing this with @naorpeled some months ago, especially because each DB has different index types, and we weren't sure whether to provide a unified approach for all drivers (e.g. like `isSpatial` is doing - difficult for vectors) or assume the users will only work with a single driver and let them specify the DB-specific index type (like in this PR).

I'm now thinking it should be fine like in this PR, and if it's not enough in the future, we could allow configuring the index type via a function.

freePixel:
> @freePixel FYI we were also planning to have vector indexes in the future, but we might implement them separately - we were actually discussing this with @naorpeled some months ago, especially because each DB has different index types, and we weren't sure whether to provide a unified approach for all drivers (e.g. like `isSpatial` is doing - difficult for vectors) or assume the users will only work with a single driver and let them specify the DB-specific index type (like in this PR).
> 
> I'm now thinking it should be fine like in this PR, and if it's not enough in the future, we could allow configuring the index type via a function.

I believe that instead of adding more properties to the `Index` decorator to support additional index variations, we can extend the newly proposed `TableIndexTypes` to include more index types. We could use a hybrid approach in which both general and specific index types are defined. For example, we could move `isSpatial` into `TableIndexTypes` as `"spatial"`, and then let the driver decide how to interpret that and convert it into a valid index type, while still allowing specific index types like `"gist"` within the `TableIndexTypes` definition.

At the end of the day, some library users will use the "generic" approach, but there may still be cases where someone needs to specify a particular index type for their driver.

`TableIndexTypes` can be extended in the future without issue. Rather than being limited to a fixed set of strings, it could eventually be extended to support more generic behavior through functions, as you mentioned.

If this PR is approved, `Index` decorator properties like `isSpatial` can be deprecated and in a future version it can be removed as having two different ways to define indexes is not ideal.

src/schema-builder/RdbmsSchemaBuilder.ts
```diff
@@ -993,6 +981,18 @@ export class RdbmsSchemaBuilder implements SchemaBuilder {
...
+                newIndices.find(
+                    (idx) =>
+                        !!idx.type &&
+                        !this.connection.driver.isIndicesTypeSupported(),
+                )
+            ) {
+                throw new TypeORMError(
+                    `Current database driver does not support index 'type' property`,
+                )
+            }
```
G0maa, line 994: Are we sure about throwing this? Is it not handled any where else?

Reasoning: `TypeORMError` was never thrown in this file before this PR.

src/schema-builder/RdbmsSchemaBuilder.ts
```diff
@@ -1046,6 +1046,18 @@ export class RdbmsSchemaBuilder implements SchemaBuilder {
...
+
```
G0maa, line 1060: It seems this condition is identical to this [one](https://github.com/typeorm/typeorm/pull/11318/files#diff-ec225eb8228f199d9649636e6cffa1ab8107bb42446a240e1828183e64603e25R984-R994), could you see if this kind of structure is repeated elsewhere? if not you may consider abstracting them in a function If it looks cleaner.

src/schema-builder/RdbmsSchemaBuilder.ts
freePixel: I don't see any problem, if user tries to use `type` in an index with an unsupported driver then an error is thrown. The code that follows (adding new indices) depends on `type` property and adding logic for that case when driver is not supported but `type` is defined seems to be unecessary and will bring unecessary complexity. A lot of typeORM exceptions (including this one) are being handled by RdbmsSchemaBuilder build() method.

> Reasoning: TypeORMError was never thrown in this file before this PR.

It was introduced in the first commits of this PR.

src/schema-builder/RdbmsSchemaBuilder.ts
freePixel: Well, this is my personal opinion in this :)
I only remove duplicates when there's 3 or more repetitions. In this case there's only two, I think it's readable as it is and it's not necessary to create a separate function for that. Also, I don't think this will be repeated 3 or more times in the future.

src/schema-builder/RdbmsSchemaBuilder.ts
freePixel: >Last one 😀
Could you verify that views/materialized views also support the same index types?

I took a look to postgreSQL documentation and althought it does not explicitly affirms that materialized views and tables support all indexes defined in their documentation, it's reasonable to infer that it's true.
FYI: 
https://www.postgresql.org/docs/current/rules-materializedviews.html#RULES-MATERIALIZEDVIEWS
https://www.postgresql.org/docs/current/indexes.html

Fortunately, given how things are implemented, there's no need to worry about that. If a materialized view didn't support all index types, the database would simply raise an error, and the user would be informed and able to fix the issue.
It's similar to asking whether this is also implemented in MySQL. If MySQL doesn't support GiST indexes, would that be a problem? No, because as I said, the library user would be aware of it and have a way to resolve it.
These kinds of errors are fair to happen when users are experimenting while developing code.

freePixel:
@G0maa, thanks for the review.
I've answered all questions.

naorpeled:
LGTM!!!
Thanks for this amazing PR 🙏

src/metadata/IndexMetadata.ts
```diff
@@ -147,7 +155,8 @@ export class IndexMetadata {
             )
                 this.synchronize = options.args.synchronize
             this.isUnique = !!options.args.unique
-            this.isSpatial = !!options.args.spatial
+            this.isSpatial =
+                !!options.args.spatial || options.args.type === "gist"
```
alumni, line 172: Postgres-specific types should not be here

src/driver/postgres/PostgresQueryRunner.ts
```diff
@@ -4345,35 +4351,65 @@ export class PostgresQueryRunner
...
+            "gin",
+            "brin",
+        ])
+
+        const type = index.isSpatial ? "gist" : index.type
+
+        if (typeof type !== "string") return null
+
+        if (!pgValidIdxTypes.has(type))
+            throw new TypeORMError(`Unsupported index type`)
```
alumni: Such validation errors we don't handle here, but earlier, in `EntityMetadata`.

Drivers could have an array of supported index types (like we do e.g. for `supportedColumnTypes`). Then in `EntityMetadata` we can check if `driver.supportedIndexTypes` contains the value.

src/driver/postgres/PostgresDriver.ts
```diff
@@ -1439,6 +1441,13 @@ export class PostgresDriver implements Driver {
         return false
     }
 
+    /**
+     * Returns true if driver supports type indices
+     */
+    isIndicesTypeSupported(): boolean {
```
alumni: Correct plural in English would be "are index types" instead of "is indices type". Probably it was meant as "is index type supported", looking at the usage?

But if you declare `Driver.supportedIndexTypes` like mentioned in the other comment, this function would no longer be necessary.

src/schema-builder/options/TableIndexTypes.ts
```diff
@@ -0,0 +1,7 @@
+export type TableIndexTypes =
+    | "btree"
```
alumni: Would be great if you could add a comment for each line that these are for Postgres, so we add support for the other DBs, we can append the driver to the comment. Helps keep track of features.

test/functional/indices/index-types/index-types.ts
alumni: I'd only add the tests for Postgres, but I get that we want to test one driver that is different. Sqlite was a good choice, it doesn't support index types.

I don't like that now there are a lot of tests in one test file.

If I will implement this for the SapDriver (BTREE | CPBTREE | INVERTED HASH | INVERTED VALUE | INVERTED INDIVIDUAL) or for MySQL/MariaDB (BTREE | HASH | RTREE), I will likely extract the driver that I'm implementing into a new folder and remove the rest 🤷🏻

freePixel:
@alumni thanks for the review. I'll be very busy over the next few days, and also need some time to think through your comments and adjust the implementation accordingly.
Hopefully I'll have everything ready this weekend.

alumni:
Take your time - most of the changes are rather simple.

The only one that needs thinking is avoiding the pg-specific `gist` check in `IndexMetadata.ts`. I'm actually not sure how to make `IndexMetadata.isSpatial = true` when the user defines `@Index({ type: 'gist' })`.

I wonder if it's better to leave it as it is in this PR _or_ if we should adjust the conditions in `RdbmsSchemaBuilder` (e.g. if `driver.supportIndexTypes` is set then check `indexMetadata.type` otherwise check `indexMetadata.isSpatial`).

src/driver/postgres/PostgresQueryRunner.ts
freePixel: Moved validation to IndexMetadata

src/driver/postgres/PostgresDriver.ts
freePixel: Now it uses Driver.supportedIndexTypes

src/metadata/IndexMetadata.ts
freePixel: Unfortunately I don't see how we can get rid of that as it is important to keep compatibility with spatial.

In my opinion we should deprecate isSpatial in a future version as it can be directly mapped to a type and then in future remove isSpatial from the codebase.

test/functional/indices/index-types/index-types.ts
freePixel: I've separated tests in separate files.

freePixel:
I've implemented suggestion according with @alumni comments and answered the comments.
If there's more changes required just reach out!

freePixel:
It seems that tests failed for postgres (17) but it's related with a network error as you can see from the logs.
I believe it's not related with this PR.

alumni:
I think we didn't notice that the test file names don't end in `.test.ts`, so the tests are not run :P

TODO: we'll need to check if there are others we forgot about and fix them :)

PR #11326, opened 2025-03-06
Fix/9006 eager load relation strategy

### LINKED ISSUE #9006: Eager relations do not respect relationLoadStrategy

SharkSharp:
### Description of change

Fixes #9006

This PR addresses the issue where the `relationLoadStrategy: 'query'` option did not split the entire query into multiple queries as expected. The functionality of loading eager relations with new queries instead of joins was not implemented.

### Changes Made

1. Implemented the functionality to load eager relations using separate queries instead of joins.
2. Fixed an issue where some queries were fetching data that the ORM already had, to link the entities.
4. Addressed the eager loop problem by allowing loops in eager loading and stopping when it reaches any already performed query. Options to either throw an error to alert the user or let TypeORM handle it may be added.
5. Identified a lack of configuration showing `RelationLoader` methods that should or should not load the eager relations according to the query option `loadEagerRelations`. This has been added.

### Ideas for Future Improvements

- Add an option in all relations called `loadStrategy` or use the existing `eager` as a string to specify whether eager loading should be performed as a join or query. This would provide detailed control over all eager loads, significantly improving performance.

### Pull-Request Checklist

- [x] This pull request links relevant issues as Fixes #9006
- [x] The new commits follow conventions explained in CONTRIBUTING.md


SharkSharp:
# Fixed nested eager relations ignoring loadEagerRelations option

I fixed a bug where nested eager relations were being loaded regardless of the `loadEagerRelations` query option setting. Previously, while the system respected this option for first-level relations, it would ignore it when loading eager relations of relations.

The fix ensures that the `loadEagerRelations` option is properly respected throughout the entire relation chain. I've also added a test case that asserts it.

I'll check if this error is occurring during `relationLoadStrategy === "join"` and fix it if happens

SharkSharp:
> Can I ask you to fix the issues with formatting please?

Fixed the issues. Let me know if there is anything else I could do.

gioboa (commented):
This file should be test/github-issues/9006/issue-9006.test.ts with `.test.ts` otherwise it will not run.

SharkSharp:
> This file should be test/github-issues/9006/issue-9006.test.ts with `.test.ts` otherwise it will not run.

Fixed. Sry about that.
I am available should any remaining issues arise.

gioboa (changes requested):
is it still valid? can you address qodo feedback please?

SharkSharp:
> is it still valid? can you address qodo feedback please?

This code was added in the commit: "[FIX]: Adding an extra step to loadForManyToMany to prevent the Cockroach driver from returning wrong ID type in getRawMany."

It was necessary at the time because, without this step, the Cockroach driver returned IDs with the wrong type, causing crashes in the matching IDs mechanism and failing Cockroach tests.

I could remove it, but I'm unsure if the underlying issue has been addressed elsewhere. This PR is almost at its anniversary, and it's been challenging to keep up with changes. I'm currently using a fork with this change in my project because it's a breaking change—my queries went from 60+ seconds to 30 milliseconds. I'd love to return to the main TypeORM branch and discuss other issues, but staying current has been difficult.

gioboa:
Thanks for your fast response and commitment. I'm more than happy to merge this PR, but right now the build is failing and there are few reviews to check.

SharkSharp:
> Thanks for your fast response and commitment. I'm more than happy to merge this PR, but right now the build is failing and there are few reviews to check.

I’ll remove the breaking code and run the tests to confirm whether CockroachDB is still crashing.
In either case, I’ll push a fix by tomorrow.

Thanks for your attention.

SharkSharp:
> > is it still valid? can you address qodo feedback please?
> 
> This code was added in the commit: "[FIX]: Adding an extra step to loadForManyToMany to prevent the Cockroach driver from returning wrong ID type in getRawMany."
> 
> It was necessary at the time because, without this step, the Cockroach driver returned IDs with the wrong type, causing crashes in the matching IDs mechanism and failing Cockroach tests.
> 
> I could remove it, but I'm unsure if the underlying issue has been addressed elsewhere. This PR is almost at its anniversary, and it's been challenging to keep up with changes. I'm currently using a fork with this change in my project because it's a breaking change—my queries went from 60+ seconds to 30 milliseconds. I'd love to return to the main TypeORM branch and discuss other issues, but staying current has been difficult.

Note that this is far from perfect. I haven’t dug deep enough to identify the root cause of this bug, and I’m not sure whether it comes from the Cockroach driver itself. For now, I’m only handling a symptom: relations are not being loaded due to ID type mismatches, and this seems to happen only with Cockroach. I don’t know whether this may have other consequences, so I’m also unsure whether it is appropriate to open an issue for this yet.

SharkSharp:
@gioboa could you take a look in this CI/CD?
It's showing 'HTTP error' in CoverAlls Parallel, it seams that the error is not related to my change.

gioboa:
The service is down [since yesterday](https://x.com/webreflection/status/2026585756693610679?s=46)😩 we need to wait, I am sorry

SharkSharp:
> Thanks for your fast response and commitment. I'm more than happy to merge this PR, but right now the build is failing and there are few reviews to check.

@gioboa are there something left to do?

pkuczynski:
Hey @SharkSharp, thanks for tackling this — getting eager relations to properly respect `relationLoadStrategy: "query"` has been a long-standing gap. The approach is solid and the core logic works well.

I went through the PR in detail and found a few issues, mostly around the `RelationIdLoader` changes. I've pushed fixes directly to your branch so we can keep things moving — here's what I changed and why:

**Bug fixes:**

1. **Metadata mutation in `loadForOneToManyAndOneToOneNotOwner`** — The destructuring swap `[relation, relation.inverseRelation!] = [...]` accidentally writes to the `.inverseRelation` property on the shared `RelationMetadata` object. Since metadata is built once at startup and reused across all queries, this would corrupt it permanently. Reverted to the original local variable pattern which is safe. ([`91183b0`](https://github.com/typeorm/typeorm/commit/91183b07a))

2. **`sameReferencedColumns` condition was too broad** — The new check (`joinColumns.every(col => nonVirtualColumns.indexOf(col) !== -1)`) is essentially always true for normal relations, causing the fast-path shortcut to fire when it shouldn't. Reverted to the original `primaryColumns.every(col => joinColumns.indexOf(col) !== -1)` which correctly limits it to the FK=PK edge case. ([`d6350b8`](https://github.com/typeorm/typeorm/commit/d6350b80a))

3. **Gap in `RelationLoader` eager loading conditional** — The `!queryBuilder` guard meant that when `queryBuilder` is provided with `"query"` strategy, neither the join nor the query branch would execute. No caller currently passes `queryBuilder` so this was dead code, but it's a subtle trap. Simplified to a clean `if/else` on the strategy. ([`612f17b`](https://github.com/typeorm/typeorm/commit/612f17b39))

4. **`this.dataSource.manager.connection`** was a roundabout way to get back to `this.dataSource`. Simplified. ([`6488249`](https://github.com/typeorm/typeorm/commit/64882497f))

5. **`loadEagerRelations: false` not propagated to `RelationLoader`** — When using query strategy with `loadEagerRelations: false` and explicit relations, nested eager relations were still loaded because `RelationLoader` had no access to the parent's `findOptions`. Threaded `loadEagerRelations` through `RelationIdLoader` → `RelationLoader` and extracted a shared `applyEagerRelations` helper. ([`bc8d306`](https://github.com/typeorm/typeorm/commit/bc8d30617), [`f2150bd`](https://github.com/typeorm/typeorm/commit/f2150bd64))

**Test improvements:**

6. Moved tests from `test/github-issues/9006/` to `test/functional/relations/load-strategy/` and consolidated with the self-referencing query tests into a single comprehensive suite. Rewrote all assertions as e2e (real database queries) instead of sinon spies on QueryBuilder internals — this way they'll catch real driver-specific differences. Also fixed `Author.books` type (`Book` → `Book[]`), the `Promise.all(dataSources.map(...))` pattern, and removed hardcoded IDs. ([`2567612`](https://github.com/typeorm/typeorm/commit/2567612fd), [`d52ac94`](https://github.com/typeorm/typeorm/commit/d52ac940b))

7. Added missing coverage: **OneToOne eager**, **ManyToOne eager**, **DataSource-level strategy**, **`find()` with multiple results**, **empty relations**, and **`loadEagerRelations: false` without explicit relations**. ([`d52ac94`](https://github.com/typeorm/typeorm/commit/d52ac940b))

Everything else in the PR looks good to me — the strategy branching in `SelectQueryBuilder.applyFindOptions()`, the `concatRelationMetadata` dedup logic, and the `loadEagerRelations` propagation to sub-queries are all clean.

pkuczynski:
A few more fixes pushed after running CI and SonarQube analysis:

**Bug fixes:**

8. **`loadEagerRelations: false` not propagated to `RelationLoader`** — When using query strategy with `loadEagerRelations: false` and explicit relations, nested eager relations were still loaded because `RelationLoader` had no access to the parent's `findOptions`. Threaded `loadEagerRelations` through `RelationIdLoader` → `RelationLoader` and extracted a shared `applyEagerRelations` helper to eliminate code duplication. ([`bc8d306`](https://github.com/typeorm/typeorm/commit/bc8d30617), [`f2150bd`](https://github.com/typeorm/typeorm/commit/f2150bd64))

9. **Hydration missing in two `RelationIdLoader` paths** — `prepareHydratedValue` was only applied in `loadForManyToMany` but not in `loadForManyToOneAndOneToOneOwner` or `loadForOneToManyAndOneToOneNotOwner`. This caused nested eager relation loading to fail on CockroachDB (and potentially other drivers) where raw query results need type conversion (e.g. bigint → number). Extracted the shared pattern into `executeAndHydrateRaw`. ([`8d5a322`](https://github.com/typeorm/typeorm/commit/8d5a32296), [`a6746fc`](https://github.com/typeorm/typeorm/commit/a6746fc02))

10. **Circular eager chains (A→B→C→A) caused infinite recursion** — The metadata validator only catches direct A↔B cycles. Longer chains like Publisher→Magazine→Journalist→Publisher bypassed it and recursed indefinitely with the query strategy. Added a per-branch visited set (`eagerLoadChain`) that propagates to child query builders — each parallel branch gets its own copy so they don't interfere. ([`0505163`](https://github.com/typeorm/typeorm/commit/05051634b))

**Code quality (SonarQube):**

11. Replaced `!` non-null assertions with proper null guards on `referencedColumn` and `mainAlias`. ([`0f1d5ca`](https://github.com/typeorm/typeorm/commit/0f1d5ca06))

SharkSharp:
Hi @pkuczynski, thank you for the time and effort you put into reviewing and improving this PR.

I reviewed the changes and noticed two points that I think are worth considering:

The way the tests were originally implemented reflects the nature of the issue: eager loading was working, but not in the correct way. I considered it important to use a spy to map the internal flow so the test would fail without the fix and pass after the fix. Looking at the new test, I believe this behavior may no longer be covered, and the maintainers can confirm whether this change introduces any issues.

Regarding the eager-load loop check, at the time I was not able to reproduce a case where the loop actually occurred, so I assumed there was already some internal process preventing those scenarios. For that reason, I did not prioritize implementing an additional safeguard. I still believe there is another safeguard preventing the loop; if I find it, I will add a comment documenting it.

I’m currently working on a robust project running on a fork with these fixes. I will apply the latest changes there and test them in staging to verify whether any unusual behavior appears. If anything remaining comes up, I’ll open a follow-up PR referencing this one and addressing the issues found.

Thanks again for your attention. Should any further changes or fixes be needed, I remain available to help, please feel free to reach out.

PR #11327, opened 2025-03-07
fix: remove unnecessary import from JS migration

### LINKED ISSUE #9149: Generating js migration files via CLI with --outputJs produces an unused import from typeorm
## Issue Description
Generating js migration files via CLI with --outputJs produces an unused import from typeorm.
I'm running the command: `npm run typeorm -- migration:generate src/migrations/migration --dataSource src/datasource.ts --pretty --outputJs`
Which produces a file `src/migrations/12345-migration.js` file as expected, and works with the `migration:run` command locally. However, the file contains unused require referring to the `MigrationInterface` and `QueryRunner` on the first line:

```js
const { MigrationInterface, QueryRunner } = require("typeorm");

module.exports = class migration12345 {
    name = 'migration12345';
    ...etc
```

This causes 2 problems for me, although I'm sure there are other manifestations of this problem too.
1. It fails eslint `no-unused-vars` rule
2. I am running these migrations in an AWS lambda function. `Typeorm` is bundled with the migration lambda function code using esbuild. The migrations cannot be bundled unfortunately due to esbuild not being to resolve the imports from the migrations directory glob at build time. Therefore, the lambda fails when running migrations because the imported migration file does not have access to `typeorm` dependency (its bundled with the migration code).

```
    "errorMessage": "Cannot find module 'typeorm'\nRequire stack:\n- /var/task/src/migrations/12345-migration.js\n- /var/task/src/handlers/migrate/index.js\n- /var/runtime/index.mjs",
```

There are 2 workarounds to this for me:
1. Exclude typeorm dependency from the bundle (not keen on doing this)
2. Manually remove the `require` statement

Obviously number 2 is preferred for now, but it would be great if the line wasn't generated at all!

### Expected Behavior

The line `const { MigrationInterface, QueryRunner } = require("typeorm");` should be omitted from generated `js` migration files.

TkachenkoDmitry:
### Description of change
Removed generated unused TypeScript imports in generated migration.
Fixes #9149

### Pull-Request Checklist

### DIFF
--- a/docs/migrations.md
+++ b/docs/migrations.md
@@ -1,15 +1,15 @@
 # Migrations
 
 - [Migrations](#migrations)
-  - [How migrations work](#how-migrations-work)
-  - [Creating a new migration](#creating-a-new-migration)
-  - [Running and reverting migrations](#running-and-reverting-migrations)
-    - [Faking Migrations and Rollbacks](#faking-migrations-and-rollbacks)
-    - [Transaction modes](#transaction-modes)
-  - [Generating migrations](#generating-migrations)
-  - [DataSource option](#datasource-option)
-  - [Timestamp option](#timestamp-option)
-  - [Using migration API to write migrations](#using-migration-api-to-write-migrations)
+    - [How migrations work](#how-migrations-work)
+    - [Creating a new migration](#creating-a-new-migration)
+    - [Running and reverting migrations](#running-and-reverting-migrations)
+        - [Faking Migrations and Rollbacks](#faking-migrations-and-rollbacks)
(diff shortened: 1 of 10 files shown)


alumni (approved):
`MigrationInterface` and `QueryRunner` are just type imports, not needed in JS. We actually convert the regular import to a type import in the TS template.

Ideally, at some point in the future, we should let the users provide their own template functions. I think for now, this should suffice.

TkachenkoDmitry:
> @TkachenkoDmitry please fix the failing test

My bad! Fixed!

naorpeled (approved):

PR #11332, opened 2025-03-11
feat: add new undefined and null behavior flags

naorpeled:
### Pull-Request Checklist


sgarner (commented):
Thanks so much for your efforts on implementing this @naorpeled 💜 

I must admit I'm a little disappointed however that my preferred outcome isn't going to be possible with these options as currently designed. I would like both `null` and `undefined` values to strictly throw exceptions.

Rather than boolean options, could we have made them enums? Like the following:

```ts
interface BaseDataSourceOptions {
    // ... other options

    findWhereBehavior?: {
        null?: 'ignore' | 'sql-null' | 'throw'
        undefined?: 'ignore' | 'throw'
    }
}
```

docs/null-and-undefined-handling.md
```diff
@@ -0,0 +1,101 @@
...
+    }
+});
+
+// This will NOT throw an error, even with throwOnUndefinedInFind enabled
+const posts2 = await repository.find({
+    where: {
+        title: "My Post" // text property is not provided, so it's ignored
+    }
+});
+```
```
sgarner, line 79: This explanation of the difference between undefined and missing properties is a little verbose. Is it necessary? 

I should hope it's obvious that where conditions are always partials and there's no need to list every property of the entity, even with `throwOnUndefinedInFind: true`.

README.md
```diff
@@ -164,6 +164,43 @@ const timber = await User.findOneBy({
 await timber.remove()
 ```
 
+## Null and Undefined Handling
```
sgarner, line 167: I'm not entirely convinced that this section belongs in the project README. Explaining it in the docs should be enough?

src/find-options/FindOptionsWhere.ts
```diff
@@ -46,5 +46,5 @@ export type FindOptionsWhereProperty<
 export type FindOptionsWhere<Entity> = {
     [P in keyof Entity]?: P extends "toString"
         ? unknown
-        : FindOptionsWhereProperty<NonNullable<Entity[P]>>
+        : FindOptionsWhereProperty<Entity[P]>
```
sgarner, line 49: I want to highlight that this is a significant change. It worries me because anyone who doesn't enable `treatJsNullAsSqlNull` is now going to lose compile-time protection against unintentionally passing null types.

src/query-builder/SelectQueryBuilder.ts
```diff
@@ -4236,9 +4237,40 @@ export class SelectQueryBuilder<Entity extends ObjectLiteral>
...
+                    const column =
+                        metadata.findColumnWithPropertyPathStrict(propertyPath)
+                    if (column) {
+                        let aliasPath = `${alias}.${propertyPath}`
+                        if (column.isVirtualProperty && column.query) {
+                            aliasPath = `(${column.query(alias)})`
+                        }
+                        andConditions.push(`${aliasPath} IS NULL`)
+                        continue
+                    }
```
sgarner, line 4272: Could this code be moved into the existing `if (column)` block at line 4289? I think this would be more readable and avoid some duplication.

src/find-options/FindOptionsWhere.ts
mguida22: I'm also worried about this. Is there a way we can type this so it's nullable only if it's opted-into via configuration?

I don't think we should merge this as is, it removes a very valuable check.

michaelbromley (dismissed):
Great work here @naorpeled - especially the comprehensive tests.

I must admit I really like @sgarner's suggestion on the config API - firstly grouping into a single property clearly shows they relate to the same thing, and his suggestion is just very clear and elegant IMO.

I also think an enum is both more descriptive and more readable than a boolean.

Whether we also support 'throw' for nulls I'm not sure of. But having an enum give sus the very easy possibility to add support for that if we decide to.

src/find-options/FindOptionsWhere.ts
sgarner: > I'm also worried about this. Is there a way we can type this so it's nullable only if it's opted-into via configuration?

I'm aware of a comparable situation in the [luxon](https://moment.github.io/luxon/#/) library. Their date creation methods when [given invalid input](https://github.com/moment/luxon/blob/master/docs/validity.md) can either return an 'invalid' date object or throw an exception depending on a config setting. The types have to handle the fact that methods called on an invalid date object produce different results than when called on a valid date. When `throwOnInvalid` is enabled in the library config, the possibility of invalid dates is eliminated allowing stricter types.

Their solution was to reference a 'settings' type which the user can override (via declaration merging) to affect the types in their project.

See https://github.com/DefinitelyTyped/DefinitelyTyped/pull/64995

Caution though, this will add a fair bit of complexity.

> But actually we are wanting to say "look - passing nulls is not what you should be doing. BUT if you somehow stupidly pass a null, we will handle it gracefully".
> In short - even with the flag, we are not condoning or encouraging passing of nulls. So the type system should also reflect that.

I would be happy with this solution, though I'm not sure if it matches other users' expectations. 🤔

naorpeled:
> Thanks so much for your efforts on implementing this @naorpeled 💜
> 
> I must admit I'm a little disappointed however that my preferred outcome isn't going to be possible with these options as currently designed. I would like both `null` and `undefined` values to strictly throw exceptions.
> 
> Rather than boolean options, could we have made them enums? Like the following:
> 
> ```ts
> interface BaseDataSourceOptions {
>     // ... other options
> 
>     findWhereBehavior?: {
>         null?: 'ignore' | 'sql-null' | 'throw'
>         undefined?: 'ignore' | 'throw'
>     }
> }
> ```

@sgarner I totally get where you're coming from, and in fact, more configurability was actually my initial thought too. I suggested something similar, but from the discussions it seemed not everyone was on board with expanding the options at that point. 

That said, I do think flexibility is the better path in the long run. I'll adjust the PR to that approach 🙏

docs/null-and-undefined-handling.md
naorpeled: Agreed, will remove 🙏

README.md
naorpeled: 100% Will remove 🙏 
Thanks for the feedback!

src/find-options/FindOptionsWhere.ts
naorpeled: I completely agree with you all,
I wasn't sure what's the best approach to this because making this type inferred based on the data source is a bit tricky.

The declaration merging approach is not ideal but I think it makes more sense than the change I made.
Thanks for the feedback guys, much appreciated!

test/functional/find-options/null-undefined-handling/null-undefined-handling.ts
```diff
@@ -0,0 +1,600 @@
...
+                    })
+
+                    expect(posts2.length).to.equal(1)
+                    expect(posts2[0].title).to.equal("Post #1")
+
+                    // Test with Repository with null relation and findOne
+                    const postWithRepo = await connection
+                        .getRepository(Post)
+                        .findOne({
+                            where: {
```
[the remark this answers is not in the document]
naorpeled: Yeah, I think it's a great idea, added 🙏

src/query-builder/SelectQueryBuilder.ts
naorpeled: Yeah, for sure, great catch!

naorpeled:
@coderabbitai summary

src/find-options/FindOptionsWhere.ts
```diff
@@ -40,11 +41,20 @@ export type FindOptionsWhereProperty<
           | Property
     : Property | FindOperator<Property>
 
+/**
+ * Helper type to conditionally allow null/undefined based on TypeORMSettings
+ */
+type NullableWhereValue<T> = TypeORMSettings extends {
+    allowNullableWhere: true
+}
+    ? T | null | undefined
```
sgarner, line 50: I think this type should only allow `null`, not `undefined` as well.

`null` values potentially become valid when using `sql-null` mode, but `undefined` values are always invalid regardless of the mode.

src/query-builder/SelectQueryBuilder.ts
```diff
@@ -4236,9 +4237,9 @@ export class SelectQueryBuilder<Entity extends ObjectLiteral>
                     .join(" OR ")
             }
         } else {
-            const andConditions: string[] = []
+            let andConditions: string[] = []
```
sgarner, line 4240: This change looks like a mistake. It can stay as `const`.

sgarner (commented):
Looking good @naorpeled 👌  Just a few more comments!

src/query-builder/SelectQueryBuilder.ts
naorpeled: Great catch, thanks!

src/find-options/FindOptionsWhere.ts
naorpeled: Great point, changed it 🙏

src/common/TypeORMSettings.ts
```diff
@@ -0,0 +1,21 @@
...
+ * // TypeScript configuration
+ * declare module 'typeorm' {
+ *   interface TypeORMSettings {
+ *     allowNullableWhere: true
+ *   }
+ * }
+ * ```
+ */
+// eslint-disable-next-line @typescript-eslint/no-empty-object-type
+export interface TypeORMSettings {}
```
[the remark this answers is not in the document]
naorpeled: When I'm doing the following I get a TypeScript error, not sure it's something we can resolve, let me know if you have any ideas 🙏 

image
image
image

src/common/TypeORMSettings.ts
```diff
@@ -0,0 +1,21 @@
...
+ * ```typescript
+ * // Runtime configuration
+ * const dataSource = new DataSource({
+ *   findWhereBehavior: { null: 'sql-null', undefined: 'throw' }
+ * })
+ *
+ * // TypeScript configuration
+ * declare module 'typeorm' {
+ *   interface TypeORMSettings {
+ *     allowNullableWhere: true
```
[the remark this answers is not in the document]
naorpeled: I think that because it's a settings object its properties should stay in camel case but I get what you're saying

src/find-options/FindOptionsWhere.ts
naorpeled: I just realized I didn't push the changes properly 😦 
now I pushed them and you should see them in Github 🙏

src/find-options/FindOptionsWhere.ts
```diff
@@ -40,11 +41,20 @@ export type FindOptionsWhereProperty<
...
+    : T
+
 /**
  * Used for find operations.
  */
 export type FindOptionsWhere<Entity> = {
     [P in keyof Entity]?: P extends "toString"
         ? unknown
-        : FindOptionsWhereProperty<NonNullable<Entity[P]>>
+        : NullableWhereValue<FindOptionsWhereProperty<NonNullable<Entity[P]>>>
```
sgarner, line 59: I just realised that TypeScript currently _doesn't_ care if you pass an `undefined` value, it only complains about `null` (which contradicts my doc edits!).

This makes using `undefined: "throw"` mode more dangerous than expected.

What do you think about:

```suggestion
        : NullableWhereValue<FindOptionsWhereProperty<Required<NonNullable<Entity[P]>>>>
```

This could be a breaking change. 😞  Maybe we need another setting for it.

michaelbromley:
We discussed this PR further, including your most recent comments @sgarner in the bi-weekly maintainer meeting today. Here's a summary:

- This PR is addressing a huge problem that is literally _the_ most-commented open issue on the whole repo. And it does a great job of addressing the issue.
- The final sticking point is around the handling of the TS types for passing `null` and `undefined` in the findWhere object. We have the following options:
   - **Remove `NonNullable`, i.e. less strict** as in the original version of the PR. Simon, Mike & I all argued against this because we thought that it removes a useful safety feature, particularly when not using the 'sql-null' mode.
   -  **Make the behaviour configurable via declaration merging** as is the current state of the PR. In today's discussion we (and I) are of the opinion that this would be probably too complex & confusing, having both runtime & type-level configs for the same thing. 
   - **Make it `Required<NonNullable>` i.e. more strict** as per [Simon's latest comment](https://github.com/typeorm/typeorm/pull/11332#discussion_r2106319091). As he notes this could be a breaking change though.
   
What we want to avoid is getting bogged down in these various details, which, when compared to the actual _real_ problem being addressed, are relatively minor.

So we in the meeting (me, @naorpeled, @alumni, @jovanadjuric, @dlhck) were learning towards the least breaking and least complex option - remove `NonNullable`.

However, now that I took some time to study everything again newly, I think I personally lean more towards keeping `NonNullable` and leaving it at that. I know there is still a type/runtime discrepancy in the handling of `undefined` if we set `undefined: 'throw'`, but IMO that's a lesser danger than the current behaviour and the simplest path to getting this PR over the line.

naorpeled:
> We discussed this PR further, including your most recent comments @sgarner in the bi-weekly maintainer meeting today. Here's a summary:
> 
> - This PR is addressing a huge problem that is literally _the_ most-commented open issue on the whole repo. And it does a great job of addressing the issue.
> - The final sticking point is around the handling of the TS types for passing `null` and `undefined` in the findWhere object. We have the following options:
>    - **Remove `NonNullable`, i.e. less strict** as in the original version of the PR. Simon, Mike & I all argued against this because we thought that it removes a useful safety feature, particularly when not using the 'sql-null' mode.
>    -  **Make the behaviour configurable via declaration merging** as is the current state of the PR. In today's discussion we (and I) are of the opinion that this would be probably too complex & confusing, having both runtime & type-level configs for the same thing. 
>    - **Make it `Required<NonNullable>` i.e. more strict** as per [Simon's latest comment](https://github.com/typeorm/typeorm/pull/11332#discussion_r2106319091). As he notes this could be a breaking change though.
>    
> What we want to avoid is getting bogged down in these various details, which, when compared to the actual _real_ problem being addressed, are relatively minor.
> 
> So we in the meeting (me, @naorpeled, @alumni, @jovanadjuric, @dlhck) were learning towards the least breaking and least complex option - remove `NonNullable`.
> 
> However, now that I took some time to study everything again newly, I think I personally lean more towards keeping `NonNullable` and leaving it at that. I know there is still a type/runtime discrepancy in the handling of `undefined` if we set `undefined: 'throw'`, but IMO that's a lesser danger than the current behaviour and the simplest path to getting this PR over the line.

I agree, and we can always revisit this and add more configuration options based on the community feedback in the future

sgarner:
> However, now that I took some time to study everything again newly, I think I personally lean more towards keeping `NonNullable` and leaving it at that. I know there is still a type/runtime discrepancy in the handling of `undefined` if we set `undefined: 'throw'`, but IMO that's a lesser danger than the current behaviour and the simplest path to getting this PR over the line.

So to reiterate, the decision is to:

- Keep the `findWhereBehavior` runtime configuration options
- Remove the `allowNullableWhere` type configuration setting (declaration merging)
- Leave `NonNullable` on the types as-is

I'm a little confused as to why we're backtracking when the work on types/declaration merging is already done? But I understand the desire to keep things simple.

Let's be clear about the effect of this decision. What we're saying is that **the types describe the "correct" practice**, which is fixed and unchanging, while the runtime options are only provided to customise how **invalid values** are handled. The runtime options are **not intended to change what is or isn't valid**.

Specifically, this means we aren't supporting users who prefer to adopt `null` in place of `IsNull()`. We're adding the `null: 'sql-null'` option only to be lenient in handling "accidental" nulls that slip through the type checks (or projects not using strict null checking at all). We still expect users to write `IsNull()` in normal cases.

Perhaps we should rename `findWhereBehavior` to `invalidWhereBehavior` to make this more explicit.

src/find-options/FindOptionsWhere.ts
naorpeled: For now we've decided to not add declaration merging and type changes to encourage our recommended practices and maintaining simplicity of configuration

naorpeled:
> > However, now that I took some time to study everything again newly, I think I personally lean more towards keeping `NonNullable` and leaving it at that. I know there is still a type/runtime discrepancy in the handling of `undefined` if we set `undefined: 'throw'`, but IMO that's a lesser danger than the current behaviour and the simplest path to getting this PR over the line.
> 
> So to reiterate, the decision is to:
> 
> * Keep the `findWhereBehavior` runtime configuration options
> * Remove the `allowNullableWhere` type configuration setting (declaration merging)
> * Leave `NonNullable` on the types as-is
> 
> I'm a little confused as to why we're backtracking when the work on types/declaration merging is already done? But I understand the desire to keep things simple.
> 
> Let's be clear about the effect of this decision. What we're saying is that **the types describe the "correct" practice**, which is fixed and unchanging, while the runtime options are only provided to customise how **invalid values** are handled. The runtime options are **not intended to change what is or isn't valid**.
> 
> Specifically, this means we aren't supporting users who prefer to adopt `null` in place of `IsNull()`. We're adding the `null: 'sql-null'` option only to be lenient in handling "accidental" nulls that slip through the type checks (or projects not using strict null checking at all). We still expect users to write `IsNull()` in normal cases.
> 
> Perhaps we should rename `findWhereBehavior` to `invalidWhereBehavior` to make this more explicit.

Exactly, and regarding the renaming of `findWhereBehavior` to `invalidWhereBehavior`, I personally find it a bit confusing because it doesn't 100% expose the fact that it affects find operations only.

imo, in the future we might support it in more operations and then maybe it'd be worth creating a new option for it or something like that, wdyt?

sgarner:
> Exactly, and regarding the renaming of `findWhereBehavior` to `invalidWhereBehavior`, I personally find it a bit confusing because it doesn't 100% expose the fact that it affects find operations only.
> 
> imo, in the future we might support it in more operations and then maybe it'd be worth creating a new option for it or something like that, wdyt?

Well that's the thing though, it doesn't just affect `EntityManager.find` operations, the change we're making affects `QueryBuilder.where` generally. So using the term `find` was already potentially misleading?

naorpeled:
> > Exactly, and regarding the renaming of `findWhereBehavior` to `invalidWhereBehavior`, I personally find it a bit confusing because it doesn't 100% expose the fact that it affects find operations only.
> > 
> > imo, in the future we might support it in more operations and then maybe it'd be worth creating a new option for it or something like that, wdyt?
> 
> Well that's the thing though, it doesn't just affect `EntityManager.find` operations, the change we're making affects `QueryBuilder.where` generally. So using the term `find` was already potentially misleading?

You're right, my bad, maybe something like queryWhereBehavior or just whereBehavior?

sgarner:
> You're right, my bad, maybe something like queryWhereBehavior or just whereBehavior?

I don't mind adding "query" or "select" in the name, but the point I was making with the name change was to emphasise that this option controls the behavior of TypeORM when **_invalid_** values are encountered. We want to avoid people thinking this option makes using `null` instead of `IsNull()` okay to do intentionally? Otherwise as soon as we release it someone's going to create an issue complaining that the types are wrong 😄 

Thinking about other operations, now I'm wondering if we should be making this change apply to `UpdateQueryBuilder` as well... is there any reason the `where` clause on an `update` should work differently to a `select`? 🤔

naorpeled:
> > You're right, my bad, maybe something like queryWhereBehavior or just whereBehavior?
> 
> I don't mind adding "query" or "select" in the name, but the point I was making with the name change was to emphasise that this option controls the behavior of TypeORM when **_invalid_** values are encountered. We want to avoid people thinking this option makes using `null` instead of `IsNull()` okay to do intentionally? Otherwise as soon as we release it someone's going to create an issue complaining that the types are wrong 😄 
> 
> Thinking about other operations, now I'm wondering if we should be making this change apply to `UpdateQueryBuilder` as well... is there any reason the `where` clause on an `update` should work differently to a `select`? 🤔 

I get what you're saying, personally I think that the documentation should say what's the recommended way and not the param name.

And regarding update, it should work the same as select imo

bombillazo:
Hey, I no longer have any skin in this game, but I've always been interested. I'm just curious, what is the strong resistance to allowing null to map to SQL null and instead forcing developers to use the `IsNull()` helper? Is it a type safety limitation?

naorpeled:
@coderabbitai help

naorpeled:
@coderabbitai summary

naorpeled:
@coderabbitai generate sequence diagram

naorpeled:
@coderabbitai summary

naorpeled:
@coderabbitai summary

docs/docs/data-source/5-null-and-undefined-handling.md
```diff
@@ -0,0 +1,250 @@
+# Handling Null and Undefined Values in Where Conditions
+
+TypeORM provides fine-grained control over how `null` and `undefined` values are handled in where conditions through the `invalidWhereValuesBehavior` configuration option in your data source options. This applies to all operations that support where clauses, including find operations, query builders, and repository methods.
```
sgarner: I'd favour this doc being a little more explanatory, and put more emphasis on the fact that `null` and `undefined` values are technically invalid.

Here is some suggested wording to expand the first section:

----

In 'WHERE' conditions the values `null` and `undefined` are not strictly valid values in TypeORM.

Passing a known `null` value is disallowed by TypeScript (when you've enabled `strictNullChecks` in tsconfig.json) at compile time. But the default behavior is for `null` values encountered at runtime to be ignored. Similarly, `undefined` values are allowed by TypeScript and ignored at runtime.

The acceptance of `null` and `undefined` values can sometimes cause unexpected results and requires caution. This is especially a concern when values are passed from user input without adequate validation.

For example, calling `Repository.findOneBy({ id: undefined })` returns the first row from the table, and `Repository.findBy({ userId: null })` is unfiltered and returns all rows.

The way in which `null` and `undefined` values are handled can be customised through the `invalidWhereValuesBehavior` configuration option in your data source options. This applies to all operations that support 'WHERE' conditions, including find operations, query builders, and repository methods.

docs/docs/data-source/2-data-source-options.md
```diff
@@ -84,6 +84,16 @@ Different RDBMS-es have their own specific options.
 -   `isolateWhereStatements` - Enables where statement isolation, wrapping each where clause in brackets automatically.
     eg. `.where("user.firstName = :search OR user.lastName = :search")` becomes `WHERE (user.firstName = ? OR user.lastName = ?)` instead of `WHERE user.firstName = ? OR user.lastName = ?`
 
+-   `invalidWhereValuesBehavior` - Controls how null and undefined values are handled in where conditions across all TypeORM operations (find operations, query builders, repository methods).
+    Options:
+
+    -   `null`: `'ignore'` (default) - skips null properties, `'sql-null'` - transforms null to SQL NULL, `'throw'` - throws an error
+    -   `undefined`: `'ignore'` (default) - skips undefined properties, `'throw'` - throws an error
```
sgarner: I wonder if we can format this more nicely with a nested list? (I'm not sure how this looks when rendered)

```suggestion
    -   `null`: `'ignore' | 'sql-null' | 'throw'`
        -   `'ignore'` (default) skips null properties
        -   `'sql-null'` transforms null to SQL 'IS NULL' operator
        -   `'throw'` throws an error
    -   `undefined`: `'ignore' | 'throw'`
        -   `'ignore'` (default) skips undefined properties
        -   `'throw'` throws an error
```

gioboa (approved):
Thanks @naorpeled 💪💯

sgarner:
> Hey, I no longer have any skin in this game, but I've always been interested. I'm just curious, what is the strong resistance to allowing null to map to SQL null and instead forcing developers to use the `IsNull()` helper? Is it a type safety limitation?

This PR adds that as an option, but the types will still prevent known-nullable values at compile time.

There are good reasons why the designers of SQL decided that `NULL = NULL` was false and required the use of the `IS NULL` operator to match null values. Remember, `NULL` is the absence of a value not a value itself, and so matching nulls to each other can be unsafe.

Automatically converting JS `null` into SQL `IS NULL` is a dangerous foot-gun because `null` values in JS can arise from the results of other operations, user input, and other sources where the developer didn't expect it, and then be matched with values in the database in ways that may be inappropriate. 

When a `null` value is hardcoded, this example looks nice and explicit:

```typescript
const usersWithNoEmail = await userRepository.findBy({ email: null });
```

With transformation that would execute:

```sql
SELECT * FROM user WHERE email IS NULL
```

No problem.

But in other cases you may have a `null` in a place you didn't intend it to be:

```typescript
const secrets = await secretRepository.findBy({ projectId: user.projectId });
```

Oops. We forgot that `user.projectId` can be null in our schema. If nulls were transformed automatically, this could execute:

```sql
SELECT * FROM secret WHERE project_id IS NULL
```

Oh no, that's not what we wanted! We just leaked all the secrets that aren't assigned to any project.

The default behavior is potentially even less safe, as the null is ignored entirely:

```sql
SELECT * FROM secret
```

Oh no! We just leaked _everyone's_ secrets!

That's a contrived example. But there are often corner cases like this in complex projects which can be dangerous if not properly handled. 

As a matter of principle, TypeORM should help you to avoid these cases, not help you to shoot yourself in the foot.

aditya-jc:
Our team recently ran into the exact scenario you're describing: we fetch business accounts where status < some_status and userid = userId.

At runtime, however, userId sometimes turns out to be null:
```
const businessUsers = await businessEntity.find({
  where: {
    userid: userId,
    status: LessThan(9),
  },
});
```

Because userId is null, TypeORM omits the condition in the generated SQL, which ends up returning all users with status < some_status.

We're currently mitigating it like this:
```
const businessUsers = await businessEntity.find({
  where: {
    userid: Equal(userId),
    status: LessThan(9),
  },
});
```

This is safer, but we haven't found a cleaner approach yet beyond adding explicit type checks/validation before calling find() or switching to QueryBuilder for stricter control.

docs/docs/data-source/2-data-source-options.md
naorpeled: Done, imo looks much nicer, thanks!

docs/docs/data-source/5-null-and-undefined-handling.md
```diff
@@ -1,6 +1,14 @@
 # Handling Null and Undefined Values in Where Conditions
 
-TypeORM provides fine-grained control over how `null` and `undefined` values are handled in where conditions through the `invalidWhereValuesBehavior` configuration option in your data source options. This applies to all operations that support where clauses, including find operations, query builders, and repository methods.
+In 'WHERE' conditions the values `null` and `undefined` are not strictly valid values in TypeORM.
+
+Passing a known `null` value is disallowed by TypeScript (when you've enabled `strictNullChecks` in tsconfig.json) at compile time. But the default behavior is for `null` values encountered at runtime to be ignored. Similarly, `undefined` values are allowed by TypeScript and ignored at runtime.
```
sgarner, line 5: Shall we remove the "TypeScript Considerations" section at the bottom of this doc, now that this is explained here?

naorpeled: Yeah, definitely, good point 🙏

alumni:
🎊

PR #11338, opened 2025-03-13
fix: sql escape issues identified by CodeQL

alumni:
### Description of change

Fix issues identified by CodeQL

### Pull-Request Checklist

### DIFF
--- a/sample/sample9-entity-listeners/entity/Post.ts
+++ b/sample/sample9-entity-listeners/entity/Post.ts
@@ -1,24 +1,26 @@
+import crypto from "node:crypto"
+
+import { AfterInsert } from "../../../src/decorator/listeners/AfterInsert"
+import { AfterLoad } from "../../../src/decorator/listeners/AfterLoad"
+import { AfterRecover } from "../../../src/decorator/listeners/AfterRecover"
+import { AfterRemove } from "../../../src/decorator/listeners/AfterRemove"
+import { AfterSoftRemove } from "../../../src/decorator/listeners/AfterSoftRemove"
+import { AfterUpdate } from "../../../src/decorator/listeners/AfterUpdate"
+import { BeforeInsert } from "../../../src/decorator/listeners/BeforeInsert"
+import { BeforeRecover } from "../../../src/decorator/listeners/BeforeRecover"
+import { BeforeRemove } from "../../../src/decorator/listeners/BeforeRemove"
+import { BeforeSoftRemove } from "../../../src/decorator/listeners/BeforeSoftRemove"
+import { BeforeUpdate } from "../../../src/decorator/listeners/BeforeUpdate"
+import { JoinTable } from "../../../src/decorator/relations/JoinTable"
+import { ManyToOne } from "../../../src/decorator/relations/ManyToOne"
 import {
     Column,
     Entity,
     ManyToMany,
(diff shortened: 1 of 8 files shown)


sample/sample9-entity-listeners/entity/Post.ts
```diff
@@ -49,7 +51,7 @@ export class Post {
         console.log(
             `event: Post "${this.title}" entity has been loaded and callback executed`,
         )
-        this.uid = Math.ceil(Math.random() * 1000)
+        this.uid = crypto.randomInt(0, 1000)
```
mguida22, line 54: It doesn't matter either way, but this sample code doesn't need to have a cryptographically secure random number.

src/driver/cockroachdb/CockroachDriver.ts
```diff
@@ -461,16 +461,14 @@ export class CockroachDriver implements Driver {
...
                     .map((val) => {
                         // replace double quotes from the beginning and from the end
                         if (val.startsWith(`"`) && val.endsWith(`"`))
                             val = val.slice(1, -1)
-                        // replace double escaped backslash to single escaped e.g. \\\\ -> \\
-                        val = val.replace(/(\\\\)/g, "\\")
-                        // replace escaped double quotes to non-escaped e.g. \"asd\" -> "asd"
-                        return val.replace(/(\\")/g, '"')
+                        // replace escaped backslash and double quotes
+                        return val.replace(/\\(\\|")/g, '$1')
```
mguida22, line 471: Do we have unit tests around this that we can use to validate the behavior we want?

This will escape differently compared to the prior code. What should `\\\\"` produce? The old version results in `"` and this version results in `\\"`

sample/sample9-entity-listeners/entity/Post.ts
alumni: Yeah, it's just so we can get rid of the CodeQL warning.

src/driver/cockroachdb/CockroachDriver.ts
alumni: I don't have time to setup CRDB for this, but the test that could be updated is `test/functional/database-schema/enums-array/*`.

CodeQL detected this correctly, double escaping should not happen, everything should be escaped at once.

alumni: @mguida22 I added some tests for the escape characters, but they would pass anyway. The reason is that this function is used when reading the value, and the value is already escaped correctly by the CRDB server/`pg` library.

Basically, this would produce incorrect results only if we would get an incorrectly escaped value, which we don't.

Nevertheless, I'd still do this fix, to avoid any issues that might arise if the code gets copy-pasted and to make CodeQL stop complaining.

OSA413 (approved):
I would also check if the test using `enumEntity.escapeCharEnums` fails in the current master branch just in case.

alumni:
> I would also check if the test using `enumEntity.escapeCharEnums` fails in the current master branch just in case.

Already did locally, it does not fail. The reason is that one that I mentioned here: https://github.com/typeorm/typeorm/pull/11338#discussion_r2009144537 (basically the value comes from CRDB already correctly escaped)

mguida22 (approved):

PR #11346, opened 2025-03-18
style: lint repository

alumni:
### Description of change

Lint and format files in the repository to reduce the number of linting errors.

### Pull-Request Checklist


src/metadata/ColumnMetadata.ts
```diff
@@ -916,7 +916,7 @@ export class ColumnMetadata {
      */
     compareEntityValue(entity: any, valueToCompareWith: any) {
         const columnValue = this.getEntityValue(entity)
-        if (ObjectUtils.isObject(columnValue)) {
+        if (typeof columnValue?.equals === "function") {
```
OSA413, line 919: Looks like it's not part of lint fix?

OSA413 (approved):
Just one thing. Also there are some warnings left in the GitHub interface.

alumni:
There are >5K warnings left, this PR only fixes most of the automatically fixable warnings (most because I didn't want to introduce any breaking changes).

I plan to open another PR with some manual changes too (I don't want to do that here since it will be easier to review them separately).

src/metadata/ColumnMetadata.ts
alumni: Yeah, eslint changed the type of `isObject` (now it's actually the same as `lodash.isObject`, which is probably correct).

Because the type got changed, TypeScript now says there's no `equals` function. Honestly, I haven't seen `equals` in JS, only in Java, so I thought this check would be more accurate.

naorpeled (approved):

PR #11357, opened 2025-03-23
refactor: database server version fetching & comparison

alumni:
### Description of change

Refactor the way the database server version is being fetched as well as the version comparison utils:
* improve code reuse for MySQL/MariaDB and Postgres drivers
* fetch the version correctly for MariaDB servers 
* implement version fetching upon connect the SAP driver
* refactor `VersionUtils.isGreaterOrEqual` to support versions with any number of components
* extract the unit tests for `VersionUtils`
* avoid awaiting for result of sync function call of `Driver.createQueryRunner`

### Pull-Request Checklist

### DIFF
--- a/src/driver/mysql/MysqlQueryRunner.ts
+++ b/src/driver/mysql/MysqlQueryRunner.ts
@@ -2555,7 +2555,7 @@ export class MysqlQueryRunner extends BaseQueryRunner implements QueryRunner {
         ])
 
         const isMariaDb = this.driver.options.type === "mariadb"
-        const dbVersion = await this.getVersion()
+        const dbVersion = this.driver.version
 
         // create tables for loaded tables
         return Promise.all(
@@ -3366,9 +3366,13 @@ export class MysqlQueryRunner extends BaseQueryRunner implements QueryRunner {
         return c
     }
 
-    protected async getVersion(): Promise<string> {
+    async getVersion(): Promise<string> {
         const result = await this.query(`SELECT VERSION() AS \`version\``)
-        return result[0]["version"]
+        const dbVersion = result[0]["version"]
+
+        return this.driver.options.type === "mariadb"
+            ? dbVersion.replace(/^([\d.]+).*$/, "$1")
+            : dbVersion
     }
 
     /**
--- a/src/driver/sap/SapQueryRunner.ts
+++ b/src/driver/sap/SapQueryRunner.ts
@@ -387,9 +387,19 @@ export class SapQueryRunner extends BaseQueryRunner implements QueryRunner {
      */
     async getCurrentDatabase(): Promise<string> {
         const currentDBQuery = await this.query(
-            `SELECT "VALUE" AS "db_name" FROM "SYS"."M_SYSTEM_OVERVIEW" WHERE "SECTION" = 'System' and "NAME" = 'Instance ID'`,
+            `SELECT "DATABASE_NAME" AS "db_name" FROM "SYS"."M_DATABASE"`,
         )
         return currentDBQuery[0]["db_name"]
+    }
+
+    /**
+     * Returns the database server version.
+     */
+    async getVersion(): Promise<string> {
+        const currentDBQuery = await this.query(
+            `SELECT "VERSION" AS "version" FROM "SYS"."M_DATABASE"`,
+        )
+        return currentDBQuery[0]["version"]
     }
 
     /**
--- a/src/driver/DriverUtils.ts
+++ b/src/driver/DriverUtils.ts
@@ -34,10 +34,7 @@ export class DriverUtils {
     }
 
     static isReleaseVersionOrGreater(driver: Driver, version: string): boolean {
-        return (
-            driver.version != null &&
-            VersionUtils.isGreaterOrEqual(driver.version, version)
-        )
+        return VersionUtils.isGreaterOrEqual(driver.version, version)
     }
 
     static isPostgresFamily(driver: Driver): boolean {
--- a/src/driver/aurora-mysql/AuroraMysqlDriver.ts
+++ b/src/driver/aurora-mysql/AuroraMysqlDriver.ts
@@ -361,7 +361,7 @@ export class AuroraMysqlDriver implements Driver {
      */
     async connect(): Promise<void> {
         if (!this.database) {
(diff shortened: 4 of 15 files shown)


OSA413 (approved):

src/driver/mysql/MysqlQueryRunner.ts
```diff
@@ -3366,9 +3366,13 @@ export class MysqlQueryRunner extends BaseQueryRunner implements QueryRunner {
         return c
     }
 
-    protected async getVersion(): Promise<string> {
+    async getVersion(): Promise<string> {
         const result = await this.query(`SELECT VERSION() AS \`version\``)
```
sgarner, line 3370: Would be nice to add a typing here, like `PostgresQueryRunner` has:

```suggestion
        const result: [{ version: string }] = await this.query(`SELECT VERSION() AS \`version\``)
```

src/driver/mysql/MysqlQueryRunner.ts
```diff
@@ -3366,9 +3366,13 @@ export class MysqlQueryRunner extends BaseQueryRunner implements QueryRunner {
...
-        return result[0]["version"]
+        const dbVersion = result[0]["version"]
+
+        return this.driver.options.type === "mariadb"
+            ? dbVersion.replace(/^([\d.]+).*$/, "$1")
+            : dbVersion
```
sgarner, line 3375: I'm inferring that mariadb version strings look like `1.2.3.foobar`, is that correct?

I'm wondering if mysql versions ever look like this as well, they might suffix prerelease builds.

Perhaps the removal of non-numeric components in the version string could be handled more generally in `parseVersion()` by filtering out `NaN` after doing `parseInt()`, WDYT?

src/driver/sap/SapQueryRunner.ts
```diff
@@ -387,11 +387,21 @@ export class SapQueryRunner extends BaseQueryRunner implements QueryRunner {
...
+            `SELECT "DATABASE_NAME" AS "db_name" FROM "SYS"."M_DATABASE"`,
         )
         return currentDBQuery[0]["db_name"]
     }
 
+    /**
+     * Returns the database server version.
+     */
+    async getVersion(): Promise<string> {
+        const currentDBQuery = await this.query(
```
sgarner, line 399:
```suggestion
        const currentDBQuery: [{ version: string }] = await this.query(
```

src/driver/mysql/MysqlQueryRunner.ts
alumni: Each DB has its own version format:

- MariaDB: `10.2.27-MariaDB-10.2.27+maria~jessie-log` -> `10.2.27`
- MySQL: `8.4.3`
- PostgreSQL: `PostgreSQL 17.2 (Debian 17.2-1.pgdg120+1) on x86_64-pc-linux-gnu, compiled by gcc (Debian 12.2.0-14) 12.2.0, 64-bit` -> `17.2`
- SAP: `4.00.000.00.1739875052`

See https://dbfiddle.uk/PzpwOqUM. So the drivers need to handle this each in its own way. `Driver.version` should extract the version from the version string (i.e. only contain numbers).

sgarner: What I was trying to say is you need to make a distinction between parsing the output of whatever database command displays the server version, and handling non-numeric values within the version string.

So extracting `17.2` from `PostgreSQL 17.2 (Debian 17.2-1.pgdg120+1)` should be handled by the driver as you are now, because this output format is a quirk of Postgres.

But it's totally acceptable for semantic versions to contain alpha values, and that's not something specific to MariaDB or any other driver (nevermind that some databases don't even conform to Semver).

From the [Semver spec](https://semver.org/#spec-item-9):

> image

I checked [the docs](https://dev.mysql.com/doc/refman/8.4/en/information-functions.html#function_version) and it indicates that MySQL versions can contain suffixes like this:

> image

The previous logic in TypeORM only cared about the first three components of the version number, so this wouldn't have mattered, but now you're comparing all of them it might cause problems...

alumni: Okay, so MySQL should behave similarly to MariaDB and discard anything that's after `x.y.z`. I can adjust that tomorrow.

SemVer is not very often followed outside of the NodeJS ecosystem, we don't have to be SemVer compatible, especially since we already have two examples that don't follow it - 3 if you count CRDB - MySQL/MariaDB are actually the only ones that have the `x.y.z` version format.

We just need something that's good enough to help us adjust the logic based on the DB server version. I don't think we will ever need the part that's after the numeric string - but we could also discard it in `parseVersion` if need be 🤷🏻‍♂️

sgarner: I was suggesting to do it in `parseVersion` just for safety's sake, all drivers would then be covered if any database engine (or new driver added in future) starts returning alpha components in their version strings.

I referenced SemVer only because it's a clearly defined standard which allows alpha tags. Of course database engine versions may not be following SemVer, but that doesn't reduce the risk of non-numeric values being present. The point is to avoid making assumptions about database vendors' versioning practices.

Regardless of the scheme being followed, it's not unusual for prerelease builds to have such tags, and users might install such database versions to work around bugs in the current stable release (I've had to do this with MySQL server before).

To put this in concrete terms, the current logic could fail if for example the user had a SAP HANA server with the version `4.00.000.beta-01.123456`: the fourth component will parse as `NaN` and then `isGreaterOrEqual` will fall through the loop to return `true` even if you wanted to require a version greater than, say, `4.00.000.05`:

```ts
VersionUtils.isGreaterOrEqual("4.00.000.beta-01.123456", "4.00.000.05")
-> true // incorrect
```

Sure, maybe no such version of SAP HANA has ever existed or will ever exist, I don't know. But it _could_ do? 🤔 

> SemVer is not very often followed outside of the NodeJS ecosystem

I have to disagree here! SemVer does not originate from the NodeJS ecosystem, it is widely adopted across many open source and commercial software projects using all kinds of languages. And while the SemVer spec defines a prescribed syntax, its main innovation was to codify the proper usage of the three components of a version number. The `x.y.z` format itself is ubiquitous and has been in use for decades.

PR #11361, opened 2025-03-24
build: improve test workflow

alumni:
### Description of change

- Decreased CockroachDB test runtime from ~38min to ~21min by using a stable version of CockroachDB server.
- Decreased Oracle server startup from ~3min to ~20s by using a lighter version of Oracle server.
- Extracted the build step separately, so it's done only once for the entire workflow instead of once for each test.
- Cleaned up tests.

### Pull-Request Checklist


.github/workflows/test.yml
```diff
@@ -6,29 +6,48 @@ name: test
...
       - uses: actions/setup-node@v4
         with:
-          node-version: "18.x"
-
-      - uses: actions/checkout@v4
+          node-version: "20.x"
+          cache: "npm"
 
-      - run: npm i
+      - run: npm ci
```
sgarner, line 24: Any reason not to use `npm ci` in the other workflows as well?

sgarner (approved):
Nice improvement 👌

.github/workflows/test.yml
alumni: Node 16 ships with an older version of `npm` that is not compatible with the lockfile (v3) in our repo. `npm ci` fails if the lockfile needs updating.

We could probably use corepack and lock in the package manager version. Maybe we can think about that after we drop support for old node versions.

sgarner: Ahh, I see...

`npm` v9 is compatible with lockfile v3 and `node` v16, so it should be possible to work around this.

Corepack doesn't support `npm` so I'm not sure that helps us?

I would suggest we use the GitHub `actions/setup-node` action in each job instead of `node:x` docker containers, because starting a container adds about 15 seconds to each job run unnecessarily, and using `actions/setup-node` enables global dependency caching.

The desired node version can be passed to `actions/setup-node`, and then where necessary the npm version can be changed with `npm i -g npm@9` for example.

Installing dependencies with `npm i` is adding another 1min+ to each node16 run, so this would be nice to optimise, unless we're dropping node16 imminently.

But there's no need to worry about it in this PR. :+1:

.github/workflows/test.yml
```diff
@@ -50,6 +70,7 @@ jobs:
       node-version: ${{matrix.node-version}}
 
   windows-database-tests:
+    needs: build
```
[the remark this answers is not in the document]
OSA413: Especially for native SQLite drivers

.github/workflows/test.yml
alumni: I can't seem to figure out why we can't connect to the database if I switch to `actions/setup-node`:

`host: "postgres"`:
```
     Error: getaddrinfo EAI_AGAIN postgres
      at GetAddrInfoReqWrap.onlookup [as oncomplete] (node:dns:107:26)
```

`host: "localhost"`:
```
     Error: connect ECONNREFUSED ::1:5432
      at TCPConnectWrap.afterConnect [as oncomplete] (node:net:1555:16)
```

.github/workflows/test.yml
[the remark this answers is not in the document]
alumni: No, it's just TypeScript transpiling the code to JS.

It's the same when we build and release it as an npm package, it will be built on a certain OS / node version, but anyone can use it, regardless of their OS / node version.

.github/workflows/test.yml
alumni: Never mind, I figured it out. Will push the commit to this branch too.

sgarner (approved):
Superb ✨

naorpeled (approved):

PR #11363, opened 2025-03-26
fix: bulk insert NULL values in Oracle (#11362)

### LINKED ISSUE #11362: Oracle bulk insert fails with ORA-01790 for null  and non varchar2 type values
### Issue description

Oracle bulk insert fails with ORA-01790 for null  and non varchar2 type values

### Expected Behavior

The following code should execute without an error:

```ts  it("should insert without error", async () =>
        await Promise.all(
            dataSources.map(async (dataSource) => {
                    const users = [{id: 1, memberId: 1,}, {id: 2, memberId: null}, {id: 3, memberId: 3}]
                    await dataSource.manager
                        .createQueryBuilder()
                        .insert()
                        .into(User)
                        .values(users)
                        .execute()
                    const count = await dataSource.getRepository(User).count()
                    expect(count).to.equal(3)
                }
            ),
        ))
```

### Actual Behavior

```
query: CREATE TABLE "user" ("id" number GENERATED BY DEFAULT AS IDENTITY, "memberId" number, CONSTRAINT "PK_cace4a159ff9f2512dd42373760" PRIMARY KEY ("id"))
query failed: INSERT INTO "user"("id", "memberId")  SELECT :1, :2 FROM DUAL UNION ALL  SELECT :3, :4 FROM DUAL UNION ALL  SELECT :5, :6 FROM DUAL -- PARAMETERS: [1,1,2,null,3,3]
error: Error: ORA-01790: expression must have same datatype as corresponding expression
```

### Steps to reproduce

```ts
@Entity({
    name: "user",
})
export class User {
    @PrimaryGeneratedColumn()
    public id: number

    @Column({ type: "number", nullable: true })
    public memberId: number | null
}
```

```ts
                   const users = [{id: 1, memberId: 1,}, {id: 2, memberId: null}, {id: 3, memberId: 3}]
                    await dataSource.manager
                        .createQueryBuilder()
                        .insert()
                        .into(User)
                        .values(users)
                        .execute()
                    const count = await dataSource.getRepository(User).count()
                    expect(count).to.equal(3)
```

### My Environment

| Operating System    |     Windows11     |
| Node.js version     | v22.14.0  |
| Typescript version  | 5.7.3  |
| TypeORM version     | 0.3.21  |

### Additional Context

Oracle seems to treat `NULL` as `VARCHAR2` type, causing the statement to fail.

This seems just the case for perpared statements.

So the plain statement executes without an error:

```
INSERT INTO "user"("id", "memberId")  SELECT 1, 1 FROM DUAL UNION ALL  SELECT 2, null FROM DUAL;
```

But when I execute it as an perpared statement, I get an Error:

```
DECLARE
    v_sql VARCHAR2(500);
BEGIN
    v_sql := 'INSERT INTO "user"("id", "memberId")
              SELECT :1, :2 FROM DUAL
              UNION ALL
              SELECT :3, :4 FROM DUAL
    EXECUTE IMMEDIATE v_sql USING 1, 1, 2, NULL;
END;
/
```

Same is true, if I use `INSERT ALL` logic

```
DECLARE
    v_sql VARCHAR2(500);
BEGIN
    v_sql := 'INSERT ALL
                  INTO "user"("id", "memberId") VALUES (:1, :2)
                  INTO "user"("id", "memberId") VALUES (:3, :4)
              SELECT * FROM DUAL';
    EXECUTE IMMEDIATE v_sql USING 1, 1, 2, NULL;
END;
/
```

The trick is, that the parameter muss be bind to a number type, so that the statement executes:

```
DECLARE
    v_sql VARCHAR2(500);
BEGIN
    v_sql := 'INSERT INTO "user"("id", "memberId")
              SELECT :1, :2 FROM DUAL
              UNION ALL
              SELECT :3, :4 FROM DUAL
    EXECUTE IMMEDIATE v_sql USING 1, 1, 2, TO_NUMBER(NULL);
END;
/
```

Alternatively it also executes, when we directly set `NULL` in the statement

```
DECLARE
    v_sql VARCHAR2(500);
BEGIN
    v_sql := 'INSERT INTO "user"("id", "memberId")
              SELECT :1, :2 FROM DUAL
              UNION ALL
              SELECT :3, NULL FROM DUAL
    EXECUTE IMMEDIATE v_sql USING 1, 1, 2;
END;
/
```

- [ ] aurora-mysql
- [ ] aurora-postgres
- [ ] better-sqlite3
- [ ] cockroachdb
- [ ] cordova
- [ ] expo
- [ ] mongodb
- [ ] mysql
- [ ] nativescript
- [x] oracle
- [ ] postgres
- [ ] react-native
- [ ] sap
- [ ] spanner
- [ ] sqlite
- [ ] sqlite-abstract
- [ ] sqljs
- [ ] sqlserver

ertl:
### Description of change

Sets `NULL` directly into the query instead of passing it as parameter.

Fixes #11362

### Pull-Request Checklist

- [x] This pull request links relevant issues as `Fixes #11362`
- [~] Documentation has been updated to reflect this change


test/github-issues/11362/issue-11362.ts
```diff
@@ -0,0 +1,47 @@
...
+import {
+    closeTestingConnections,
+    createTestingConnections,
+    reloadTestingDatabases,
+} from "../../utils/test-utils"
+import { DataSource } from "../../../src"
+import { User } from "./entity/user"
+import { expect } from "chai"
+
+describe("github issues > #11362 Oracle bulk insert fails with ORA-01790 for null and non varchar2 type", () => {
```
alumni, line 13: Could we convert this to a functional test?

This should work for all drivers and it would be good if we tested that.

We already have something for bulk insert in `test/functional/query-builder/insert/query-builder-insert.ts`.

ertl: Okay, I added the test functionality in `query-builder-insert.ts` so that it runs for every database. Two tests were deactivated for Oracle. I modified them so they pass.

Also, note that the column definition `type: 'number'` in the original test is not supported by every db driver. Therefore, I had to remove it along with the optional null type. As a result, I needed to a cast, to explicitly set the value null.

test/functional/tree-tables/materialized-path/materialized-path.ts
alumni, line 1: Yeah, the entire test `test/github-issues/8443` was copied (instead of moved) to `test/functional/tree-tables`.

I would not remove anything from the functional tests though, I think they are easier to understand than the per-issue tests.

ertl: The file had the same test duplicated, so I removed the redundant one to avoid unnecessary execution.

ertl: @alumni The tests are still very unstable. Could you please restart the missing ones?

alumni: > The file had the same test duplicated, so I removed the redundant one to avoid unnecessary execution.

There are actually more that are redundant. I found 8443 and 9534 that should be merged into the functional tests. Basically those should be deleted, not the functional tests, but someone needs to take a look at the differences. I would not do it in this PR (basically we can undo the change to this file).

> The tests are still very unstable. Could you please restart the missing ones?

Restarted. We're currently doing quite some changes to the test setup - there might be some flaky tests popping up.

ertl: @alumni I understand what you are saying and I agree with you about the scope of the PR as well as the functional tests. The functional tests are the same as before. The file contained exactly the same test twice. Actually it was not my intention to touch the test, but after the pipeline failed exactly at this point for one testsuit for an inexplicable reason, I took a closer look and realized that the test was twice in the file.

alumni: @ertl We would like to remove the duplicate tests from `test/github-issues/8443` and `test/github-issues/9534` and keep the test in `test/functional/tree-tables/materialized-path/materialized-path.ts`.

ertl: @alumni I got you. But note the test ist sill available in `test/functional/tree-tables/materialized-path/materialized-path.ts`. The file itselft contained a duplicate of the same test. (see attachment)
 So why do you want  to keep the same test twice in `test/functional/tree-tables/materialized-path/materialized-path.ts`?

image

alumni: Aaa ok. I thought it's just copied to the other files.

alumni (approved):

mguida22 (approved):

PR #11373, opened 2025-03-30
feat: Init version in postgres driver only if not set

### LINKED ISSUE #11372: Allow postgres version to be set statically
### Feature Description

Both `database` and `searchSchema` are only set in `connect()` when they are not defined, but `queryRunner.getVersion()` is always called.

Can the call to `getVersion()` also be changed to be similar to the other properties?

As this field is only used to determine if generated columns are supported, it can be trivially overwritten to reduce cold start time.

### The Solution

This allows consumers of this library to set the version to avoid one extra query, which will reduce cold start time which is critical in serverless applications.

### Considered Alternatives

N/A

### Additional Context

_No response_

- [ ] aurora-mysql
- [ ] aurora-postgres
- [ ] better-sqlite3
- [ ] cockroachdb
- [ ] cordova
- [ ] expo
- [ ] mongodb
- [ ] mysql
- [ ] nativescript
- [ ] oracle
- [x] postgres
- [ ] react-native
- [ ] sap
- [ ] spanner
- [ ] sqlite
- [ ] sqlite-abstract
- [ ] sqljs
- [ ] sqlserver

hfhchan-plb:
### Description of change

Fixes #11372

### Pull-Request Checklist

### DIFF
--- a/src/driver/postgres/PostgresDriver.ts
+++ b/src/driver/postgres/PostgresDriver.ts
@@ -360,7 +360,9 @@ export class PostgresDriver implements Driver {
 
         const queryRunner = this.createQueryRunner("master")
 
-        this.version = await queryRunner.getVersion()
+        if (!this.version) {
+            this.version = await queryRunner.getVersion()
+        }
 
         if (!this.database) {
             this.database = await queryRunner.getCurrentDatabase()


alumni:
@G0maa it was quite some time ago when I had that fresh in mind, right now I don't really remember what was the problem (maybe since I didn't request changes it wasn't that important).

We could maybe offer some connection option `dbFeatureVersion` (more explicit than just `version`) with the default being to automatically get the version from the DB (currently on connect, ideally when required).

G0maa (approved):
The general idea/approach looks good, I'm unsure if this would break things else where like @alumni said, waiting for maintainers' opinions.

pkuczynski:
This can't break anything as we already set the version. It improves this, by setting the version only if it was not set before (so does it less forcefully). Feels like a good change.

alumni:
@hfhchan-plb Hold on, how does this even work? How would I use this if I wanted to avoid reading the version from the DB?

I see it's relying on `Driver.version`, but how is that being set?

Issue #9318, opened 2022-08-24
Postgres version command incompatible with Redshift


CaptainCrucial:
When connecting using the Postgres driver, the afterConnect method is running the SQL`SHOW server_version;` as part of feature support detection.

This command works on postgres and cockroachdb, but is incompatible with the redshift database.

This command could be adjusted to use the `SELECT version();` command which is supported across all of these database types.

I've patched this command locally and successfully connected and read data using typeorm:

```patch
diff --git a/src/driver/postgres/PostgresDriver.ts b/src/driver/postgres/PostgresDriver.ts
index 92142ba53..0dfad6d7a 100644
--- a/src/driver/postgres/PostgresDriver.ts
+++ b/src/driver/postgres/PostgresDriver.ts
@@ -375,13 +375,16 @@ export class PostgresDriver implements Driver {
 
         const results = (await this.executeQuery(
             connection,
-            "SHOW server_version;",
+            "SELECT version();",
         )) as {
             rows: {
-                server_version: string
+                version: string
             }[]
         }
-        const versionString = results.rows[0].server_version
+        const versionString = results.rows[0].version.replace(
+            /^PostgreSQL ([\d\.]+) .*$/,
+            "$1",
+        )
         this.isGeneratedColumnsSupported = VersionUtils.isGreaterOrEqual(
             versionString,
             "12.0",
diff --git a/src/driver/postgres/PostgresQueryRunner.ts b/src/driver/postgres/PostgresQueryRunner.ts
index e8aaa3eaa..4a5c84655 100644
--- a/src/driver/postgres/PostgresQueryRunner.ts
+++ b/src/driver/postgres/PostgresQueryRunner.ts
@@ -3989,8 +3989,8 @@ export class PostgresQueryRunner
      * Loads Postgres version.
      */
     protected async getVersion(): Promise<string> {
-        const result = await this.query(`SHOW SERVER_VERSION`)
-        return result[0]["server_version"]
+        const result = await this.query(`SELECT version()`)
+        return result[0]["version"].replace(/^PostgreSQL ([\d\.]+) .*$/, "$1")
     }
 
     /**
diff --git a/test/functional/cube/postgres/cube-postgres.ts b/test/functional/cube/postgres/cube-postgres.ts
index f70baf10b..d5e81a68c 100644
--- a/test/functional/cube/postgres/cube-postgres.ts
+++ b/test/functional/cube/postgres/cube-postgres.ts
[...]

PR #11375, opened 2025-03-31
fix: version detection for Postgres derived variants

alumni:
### Description of change

Fixes version detection for PostgreSQL. The version might not always be followed by a space (similar to MySQL/MariaDB).

Example version strings:
```
PostgreSQL 9.3.25 on x86_64-pc-linux-musl, compiled by gcc (Alpine 11.2.1_git20220219) 11.2.1 20220219, 64-bit
PostgreSQL 9.6.24 on x86_64-pc-linux-musl, compiled by gcc (Alpine 11.2.1_git20220219) 11.2.1 20220219, 64-bit
PostgreSQL 10.23 on x86_64-pc-linux-gnu, compiled by gcc (GCC) 8.5.0 20210514 (Red Hat 8.5.0-15), 64-bit
PostgreSQL 14.9 (Debian 14.9-1.pgdg110+1) on x86_64-pc-linux-gnu, compiled by gcc (Debian 10.2.1-6) 10.2.1 20210110, 64-bit
PostgreSQL 14.10 on x86_64-pc-linux-gnu, compiled by gcc (GCC) 8.5.0 20210514 (Red Hat 8.5.0-20), 64-bit
PostgreSQL 17.2 (Debian 17.2-1.pgdg120+1) on x86_64-pc-linux-gnu, compiled by gcc (Debian 12.2.0-14) 12.2.0, 64-bit
PostgreSQL 11.2-YB-2.6.18.0-b0 on x86_64-pc-linux-gnu, compiled by gcc (Homebrew gcc 5.5.0_4) 5.5.0, 64-bit
PostgreSQL 11.2-YB-2.18.1.0-b0 on x86_64-pc-linux-gnu, compiled by clang version 15.0.3 (https://github.com/yugabyte/llvm-project.git 0b8d1183745fd3998d8beffeec8cbe99c1b20529), 64-bit
PostgreSQL 15.2-YB-2.25.1.0-b0 on x86_64-pc-linux-gnu, compiled by clang version 17.0.6 (https://github.com/yugabyte/llvm-project.git 9b881774e40024e901fc6f3d313607b071c08631), 64-bit
```

### Pull-Request Checklist

### DIFF
--- a/src/driver/postgres/PostgresQueryRunner.ts
+++ b/src/driver/postgres/PostgresQueryRunner.ts
@@ -4160,7 +4160,11 @@ export class PostgresQueryRunner
         const result: [{ version: string }] = await this.query(
             `SELECT version()`,
         )
-        return result[0].version.replace(/^PostgreSQL ([\d.]+) .*$/, "$1")
+
+        // Examples:
+        // Postgres: "PostgreSQL 14.10 on x86_64-pc-linux-gnu, compiled by gcc (GCC) 8.5.0 20210514 (Red Hat 8.5.0-20), 64-bit"
+        // Yugabyte: "PostgreSQL 11.2-YB-2.18.1.0-b0 on x86_64-pc-linux-gnu, compiled by clang version 15.0.3 (https://github.com/yugabyte/llvm-project.git 0b8d1183745fd3998d8beffeec8cbe99c1b20529), 64-bit"
+        return result[0].version.replace(/^PostgreSQL ([\d.]+).*$/, "$1")
     }
 
     /**
--- a/test/functional/cube/postgres/cube-postgres.ts
+++ b/test/functional/cube/postgres/cube-postgres.ts
@@ -1,6 +1,8 @@
-import "reflect-metadata"
 import { expect } from "chai"
+import "reflect-metadata"
+
 import { DataSource } from "../../../../src/data-source/DataSource"
+import { DriverUtils } from "../../../../src/driver/DriverUtils"
 import {
     closeTestingConnections,
     createTestingConnections,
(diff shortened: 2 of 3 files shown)


src/driver/postgres/PostgresQueryRunner.ts
```diff
@@ -4160,7 +4160,11 @@ export class PostgresQueryRunner
         const result: [{ version: string }] = await this.query(
             `SELECT version()`,
         )
-        return result[0].version.replace(/^PostgreSQL ([\d.]+) .*$/, "$1")
+
+        // Examples:
+        // Postgres: "PostgreSQL 14.10 on x86_64-pc-linux-gnu, compiled by gcc (GCC) 8.5.0 20210514 (Red Hat 8.5.0-20), 64-bit"
+        // Yugabyte: "PostgreSQL 11.2-YB-2.18.1.0-b0 on x86_64-pc-linux-gnu, compiled by clang version 15.0.3 (https://github.com/yugabyte/llvm-project.git 0b8d1183745fd3998d8beffeec8cbe99c1b20529), 64-bit"
+        return result[0].version.replace(/^PostgreSQL ([\d.]+).*$/, "$1")
```
mguida22, line 4167: Ideally we'd get the Postgres version number from `server_version_num` instead of parsing the string.

> Software developers should use server_version_num (available since 8.2) or [PQserverVersion](https://www.postgresql.org/docs/current/libpq-status.html#LIBPQ-PQSERVERVERSION) instead of parsing the text version.

https://www.postgresql.org/docs/current/functions-info.html#FUNCTIONS-INFO-VERSION
https://www.postgresql.org/docs/current/runtime-config-preset.html#GUC-SERVER-VERSION-NUM

This has been available since Postgres 8.2 which is quite old now. Do we have a minimum supported Postgres version?

mguida22 (approved):
These changes look good.

I left a comment about using the integer version instead of parsing the string version if you want to switch to that approach, but I don't think it's necessary for this change.

src/driver/postgres/PostgresQueryRunner.ts
alumni: `SHOW server_version_num` would output something like `170002` - which can be used to compare versions, but that would mean changing the rest of the implementation too, to use this format.

`SHOW server_version` would output something like this:

```
9.3.25
9.6.24
14.10
15.0
16.6 (Debian 16.6-1.pgdg120+1)
17.2 (Debian 17.2-1.pgdg120+1)
11.2-YB-2.8.8.0-b0
11.2-YB-2.18.1.0-b0
```

I could also use that instead of `SELECT version()`.

alumni: Actually, I just noticed the reason why we are using `SELECT version()`: #9318

https://docs.aws.amazon.com/redshift/latest/dg/c_unsupported-postgresql-functions.html

Let's leave it like this to have broader compatibility.

naorpeled (approved):

PR #11376, opened 2025-04-01
docs: add comment explaining select version()

mguida22:
### Description of change

Adds a comment explaining why we use `SELECT version()` for Postgres version detection. (Follow up to https://github.com/typeorm/typeorm/pull/11375).

### Pull-Request Checklist

- [x] ~~This pull request links relevant issues as `Fixes #0000`~~
- [x] ~~There are new or updated unit tests validating the change~~
- [x] ~~Documentation has been updated to reflect this change~~

### DIFF
--- a/src/driver/postgres/PostgresQueryRunner.ts
+++ b/src/driver/postgres/PostgresQueryRunner.ts
@@ -4157,6 +4157,12 @@ export class PostgresQueryRunner
      * Loads Postgres version.
      */
     async getVersion(): Promise<string> {
+        // we use `SELECT version()` instead of `SHOW server_version` or `SHOW server_version_num`
+        // to maintain compatability with Amazon Redshift.
+        //
+        // see:
+        //  - https://github.com/typeorm/typeorm/pull/9319
+        //  - https://docs.aws.amazon.com/redshift/latest/dg/c_unsupported-postgresql-functions.html
         const result: [{ version: string }] = await this.query(
             `SELECT version()`,
         )


naorpeled (approved):

alumni:
FYI, I discovered CrateDB which does the following:
```sql
SELECT VERSION();
-- CrateDB 5.10.3 (built e185886/NA, Linux 6.12.13-200.fc41.aarch64 aarch64, OpenJDK 64-Bit Server VM 23.0.2+7)
SHOW SERVER_VERSION;
-- 14.0
```

I guess it's difficult to make this work, and there's a PR that tries not to load the version on connect. It might actually be sensible to offer a lazy-loading option for the server version at some point.

alumni (approved):

src/driver/postgres/PostgresQueryRunner.ts
```diff
@@ -4157,6 +4157,12 @@ export class PostgresQueryRunner
...
      */
     async getVersion(): Promise<string> {
+        // we use `SELECT version()` instead of `SHOW server_version` or `SHOW server_version_num`
+        // to maintain compatability with Amazon Redshift.
+        //
+        // see:
+        //  - https://github.com/typeorm/typeorm/pull/9319
+        //  - https://docs.aws.amazon.com/redshift/latest/dg/c_unsupported-postgresql-functions.html
         const result: [{ version: string }] = await this.query(
             `SELECT version()`,
```
alumni, line 4167: I noticed CrateDB names the column `pg_catalog.version()` 🤦🏻 

```sql
SELECT version() AS "version"
```

mguida22:
> FYI, I discovered CrateDB which does the following:
> 
> ```sql
> SELECT VERSION();
> -- CrateDB 5.10.3 (built e185886/NA, Linux 6.12.13-200.fc41.aarch64 aarch64, OpenJDK 64-Bit Server VM 23.0.2+7)
> SHOW SERVER_VERSION;
> -- 14.0
> ```
> 
> I guess it's difficult to make this work, and there's a PR that tries not to load the version on connect. It might actually be sensible to offer a lazy-loading option for the server version at some point. In that case we could maybe afford to write a `try-catch` that does two queries in case one fails.

Yeah I think that's the route we'll need to take eventually. We can assume it's compatible with Postgres first and fallback to an alternative if that fails to resolve.

That CrateDB example is tricky. Currently we'll resolve to the 'wrong' version since we're really looking for a Postgres compatibility version but they return the CrateDB version.

PR #11394, opened 2025-04-07
fix: prevent unnecessary invocation of transformer

### LINKED ISSUE #5719: default value on entity trigger transformer twice
**Issue type:**

[ ] question
[X] bug report
[ ] feature request
[ ] documentation issue

**Database system/driver:**

[ ] `cordova`
[ ] `mongodb`
[ ] `mssql`
[X] `mysql` / `mariadb`
[ ] `oracle`
[ ] `postgres`
[ ] `cockroachdb`
[ ] `sqlite`
[ ] `sqljs`
[ ] `react-native`
[ ] `expo`

**TypeORM version:**

[X] `latest`
[ ] `@next`

I've added default attribute on a property just to have correct migration file. That's its purpose when I check the documentation (https://typeorm.io/#/entities/column-options).

But I have created my transformer which transform my property from string to object. Unfortunately, my transformer `from` method is called twice. One time with the default string (but it shouldn't, because it's a **database-level** column's DEFAULT value), and another time with the value already transformed in json.

The only way to correct it is to remove the default value, but if I generate my migrations they will not be correct.

for example:
- company.entity.ts
```typescript
@Entity('companies')
export class Company {
  @PrimaryGeneratedColumn({ type: 'int', name: 'id' })
  id: number;

// ...

  @Column({
    type: 'varchar',
    length: 1024,
    default: '{"days_max": 720}',
    name: 'options',
    transformer: CompanyOptionsTransformer
  })
  options: Options;
}
```
 - company-options.transformer.ts
``` typescript
export const CompanyOptionsTransformer: ValueTransformer = {
  from(value: string): Options {
    if (!value) return undefined;

    const options = JSON.parse(options);
    return {
      daysMax: options.days_max
    } as Options;
  },

  to(value: Options): string {
    if (!value) return undefined;

    const options = { days_max: value.daysMax };
    return JSON.stringify(options);
  }
};
```
it's certainly linked to https://github.com/typeorm/typeorm/issues/3763

ESnark:
### Description of change
This PR addresses an issue #5719 
If you save entities like this:
```typescript
class AddFiveOnSave implements ValueTransformer {
  to(value: number): number {
    return value + 5;
}

  from(value: number): number {
    return value - 5;
  }
}

@Entity()
class Example {
  @PrimaryGeneratedColumn()
  id: number;

  @Column({
    type: 'int',
    default: 100,
    transformer: new AddFiveOnSave(),
  })
  count: number;
}

const entity = new Example();
entity.count = 200;
await repository.save(entity);

console.log(entity.count); // You will get 205
```

This unexpected result occurs due to how TypeORM merges query results into entities. TypeORM merges default column values even if you explicitly specify them. During this merging process, the from() method of the transformer is invoked, causing the transformed value to be applied unnecessarily.

This PR removes returning data of default column, if values are specified before save, to prevent being merged (and belonging transform)

### Pull-Request Checklist

- [x] This pull request links relevant issues as `Fixes #5719 `

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevents database-returned default values from overwriting user-provided values on insert, ensuring input values are preserved for columns with defaults.
  * Improves consistency with value transformers by maintaining correct transformed results when defaults are involved.

* **Tests**
  * Added comprehensive tests covering entities with defaulted columns and transformers, including create, read, and update scenarios to validate correct transformer invocation and value preservation.


ESnark:
Could someone take a look at this PR when possible?

test/github-issues/5719/issue-5719.test.ts
```diff
@@ -0,0 +1,116 @@
...
+    beforeEach(() => {
+        TestTransformer.reset()
+        return reloadTestingDatabases(connections)
+    })
+
+    after(() => closeTestingConnections(connections))
+
+    it("should call transformer's from() method only once for columns with default value", () =>
+        Promise.all(
+            connections.map(async (connection) => {
```
alumni: `Promise.all` will interfere with the static properties/methods of `TestTransformer` if there's more than one connection. This would be more visible if e.g. you would add MariaDB to the enabled drivers.

We need to run things sequentially. Use `for .. of` instead.

test/github-issues/5719/issue-5719.test.ts
```diff
@@ -0,0 +1,116 @@
...
+                schemaCreate: true,
+                dropSchema: true,
+                // The problem occurs in databases which support returning statement
+                enabledDrivers: ["mysql"],
+            })),
+    )
+
+    beforeEach(() => {
+        TestTransformer.reset()
+        return reloadTestingDatabases(connections)
```
alumni: `reloadTestingDatabases` has no return value.

test/github-issues/5719/issue-5719.test.ts
```diff
@@ -0,0 +1,116 @@
...
+                    100,
+                )
+                expect(loadedEntity!.columnWithTransformerOnly).to.equal(200) // 200/10
+
+                // Verify from() call count after find
+                const callCountAfterFind = TestTransformer.fromCallCount
+
+                // Current bug: from() is called twice for columns with default
+                // Expected behavior: from() should be called once per column
+                console.log(
```
alumni: Please remove the `console.log`.

Also: the test does not have any relevant assertions (that check what the test title says).

test/github-issues/5719/issue-5719.test.ts
```diff
@@ -0,0 +1,116 @@
...
+                    `[${connection.name}] from() call counts after find: ${callCountAfterFind}`,
+                )
+
+                // Enable this verification after the bug is fixed (currently fails)
+                // expect(callCountAfterFind).to.equal(2) // Once for each column
+            }),
+        ))
+
+    it("should call transformer's from() method only once for columns with default value", () =>
```
alumni: There are multiple tests with the same title and implementation in this file. Unnecessary tests should be removed.

alumni:
@ESnark Thanks for looking into this!

Could you please take a look at the tests?


Also, why are we removing the default columns? They should still be returned!

ESnark:
> @ESnark Thanks for looking into this!
> 
> Could you please take a look at the tests?
> 
> Also, why are we removing the default columns? They should still be returned!

I've updated the tests as requested.

Regarding the delete logic: it only removes values from returningResult when the entity already has an explicit value set `(column.getEntityValue(entity) !== undefined)`.

This mirrors standard DBMS behavior: when a value is explicitly provided, the database ignores the default. TypeORM should behave the same way.

Without this fix, the `from()` transformer gets called twice—once in `prepareHydratedValue()` and again during `merge()`—causing incorrect values.

When no value is set on the entity, the default column value is preserved and correctly returned.

gioboa (changes requested):
is it still valid? can you address qodo feedback please?

michaelbromley:

```
git fetch origin
```

PR #11398, opened 2025-04-07
Fix(undefined-null-flags): Handle null for relations

jovanadjuric:
### Pull-Request Checklist

### DIFF
--- a/src/query-builder/QueryBuilder.ts
+++ b/src/query-builder/QueryBuilder.ts
@@ -1581,13 +1581,11 @@ export abstract class QueryBuilder<Entity extends ObjectLiteral> {
                     parameters: [aliasPath, ...parameters],
                 }
             }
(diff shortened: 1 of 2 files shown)


jovanadjuric:
> I think that this PR has many breaking changes

Hey @alienap93, thanks for your feedback! 

Would you care to elaborate? As far as I can tell, no breaking changes are featured here. The parent PR is still a draft, and it will need quite a bit of testing before we merge it to master anyway.

- The new else-if blocks in `QueryBuilder.ts` are only executed if the feature flags are set to true. Otherwise, the return statement at the bottom of the method will catch all unhandled cases, just like it did before
- I moved the if block on `SelectQueryBuilder.ts:4385` up as it was unreachable because `typeof null === "object"` evaluates to `true`

Let me know if I missed something! 👀

Issue #1738, opened 2018-03-10
Proposal: console debugger to have nicely formatted output


aaronstaves:
**Issue type:**

[ ] question
[ ] bug report
[x] feature request
[ ] documentation issue

**Database system/driver:**

[x] `cordova`
[x] `mongodb`
[x] `mssql`
[x] `mysql` / `mariadb`
[x] `oracle`
[x] `postgres`
[x] `sqlite`
[x] `sqljs`
[x] `websql`

**TypeORM version:**

[x] `latest`
[ ] `@next`
[ ] `0.x.x` (or put your version here)

**Problem**
I've used other ORMs in the past where the debug output for queries are nicely formatted, which vastly improves readability and debugging.  Example:

Current debug output looks something like this:

```
SELECT user.*, photo.* FROM users user JOIN photos photo ON photo.user = user.id AND .isRemoved = FALSE user.name = 'Timber'
```

Proposed debug output would be something like this:
```
SELECT USER.*, 
       photo.* 
FROM   users USER 
       INNER JOIN photos photo 
               ON photo.USER = USER.id 
                  AND photo.isremoved = false 
WHERE  USER.NAME = 'Timber' 
```

**Proposal**
I briefly discussed this in gitter a week or so back, and seems like this functionality isn't easily extendable from the current debuggers without making a completely new one.  I'd be happy to make a PR for this but not sure what specific solution would work best for the project.  Possible options I could foresee are -

1. Create a new logger, something like `advanced-console-formatted` or `advanced-console-pretty`.  Can leave default to `advanced-console` if that makes sense.
2. Create a new config param (i.e. `debugFormatter`) param that takes in a function that all queries will be run through prior to debug output.
3. Export existing debuggers (i.e. `AdvancedConsoleLogger`, `SimpleConsoleLogger`) so they can be extended.  Then users can extend them as they see fit.

**Solution**
To accomplish the actual formatting, something like [sql-formatter](https://www.npmjs.com/package/sql-formatter) would work great.  For formatting the NoSQL drivers, could just as easily use one of the many JSON formatting libraries.  I don't use NoSQL so I have no idea what the current debug output looks like.

**Testing**
I also don't see any tests out there for the current debuggers.  Is that by design or are they just not able to be tested.  Ideally I'd include tests with my PR.

I'd be open to suggestions on any of this.  I can also just keep doing what I'm doing with my current project, but would love to see this functionality, or at least the option to format output, in typeorm directly.

[...]

PR #11401, opened 2025-04-10
feat: add FormattedConsoleLogger

w3nl:
### Pull-Request Checklist

Fixes: #1738


test/github-issues/1738/queries.ts
```diff
@@ -0,0 +1,24 @@
+export const CASE_1_UNFORMATTED = `SELECT * FROM tbl`
+export const CASE_1_FORMATTED = `SELECT
+  *
+FROM
+  tbl`
+
+export const CASE_2_UNFORMATTED = `SELECT tbl1.col1, tbl2.col2 FROM tbl1 INNER JOIN tbl2 ON tbl1.id = tbl2.id`
+export const CASE__FORMATTED = `SELECT
```
jovanadjuric, line 8:
```suggestion
export const CASE_2_FORMATTED = `SELECT
```

test/github-issues/1738/queries.ts
```diff
@@ -0,0 +1,24 @@
...
+  INNER JOIN tbl2 ON tbl1.id = tbl2.id`
+
+export const FORMAT_SQL_TEST_CASES = [
+    {
+        unformatted: CASE_1_UNFORMATTED,
+        formatted: CASE_1_FORMATTED,
+    },
+    {
+        unformatted: CASE_2_UNFORMATTED,
+        formatted: CASE__FORMATTED,
```
jovanadjuric, line 22:
```suggestion
        formatted: CASE_2_FORMATTED,
```

test/github-issues/1738/queries.ts
w3nl: bc7e007ee9cd138878e6da0687c724db207cb8f8

test/github-issues/1738/queries.ts
```diff
@@ -0,0 +1,24 @@
+export const CASE_1_UNFORMATTED = `SELECT * FROM tbl`
```
naorpeled, line 1: Let's also add a case for sub queries

jovanadjuric (changes requested):
Dialect-specific queries are failing:
```Parse error: Unexpected \"$1 ORDER B\" at line 1 column 468.\nThis likely happens because you're using the default \"sql\" dialect.\nIf possible, please select a more specific dialect (like sqlite, postgresql, etc).```

File: test/github-issues/1738/queries.ts (hunk 3 above)
Comment (naorpeled) [reply test/github-issues/1738/queries.ts:1]: Exactly, awesome 🙏

File: src/logger/FormattedConsoleLogger.ts [hunk 4]
```diff
@@ -0,0 +1,73 @@
...
+                        PlatformTools.logError(
+                            message.prefix,
+                            String(message.message),
+                        )
+                    } else {
+                        console.error(
+                            PlatformTools.error(String(message.message)),
+                        )
+                    }
+                    break
```
jovanadjuric, line 69: Let's add a default case, perhaps with `PlatformTools.log`

src/platform/PlatformTools.ts
```diff
@@ -220,6 +221,30 @@ export class PlatformTools {
         })
     }
 
+    /**
+     * Pretty-print sql string to be print in the console.
+     */
+    static formatSql(sql: string, dataSourceType?: string): string {
```
jovanadjuric, line 227: What do you think about typing it like this: `dataSourceType?: string` -> `dataSourceType?: DatabaseType`?

src/logger/FormattedConsoleLogger.ts
jovanadjuric: Btw you don't have to add this if other formatters don't have it

jovanadjuric (approved):

src/logger/AbstractLogger.ts
```diff
@@ -314,6 +315,11 @@ export abstract class AbstractLogger implements Logger {
             if (message.format === "sql") {
                 let sql = String(message.message)
 
+                console.log("")
```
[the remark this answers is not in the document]
w3nl: 09493020d01a45216681bd014d381b00017b515f

src/logger/FormattedConsoleLogger.ts
w3nl: Is now in line with other loggers

naorpeled (approved):

src/logger/AbstractLogger.ts
```diff
@@ -293,12 +293,14 @@ export abstract class AbstractLogger implements Logger {
             | number
             | (LogMessage | string | number)[],
         options?: Partial<PrepareLogMessagesOptions>,
```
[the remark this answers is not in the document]
w3nl: I think `options` is just for the options like `formatSql`, `highlightSql`, etc.
These are all `boolean`.
`queryRunner` is another dependency that we need, and I think it is another level of data.
That is why I don't think we should move `queryRunner` to the options, because it isn't an option, but more a dependency.

package.json
```diff
@@ -97,6 +97,7 @@
     "dotenv": "^16.4.7",
     "glob": "^10.4.5",
     "sha.js": "^2.4.11",
+    "sql-formatter": "^15.5.2",
```
mguida22, line 100: I see now that we depend on `@sqltools/fomatter` above, can we use that instead of adding `sql-formatter` as a new dependency? It's a significantly smaller package ([75kB](https://www.npmjs.com/package/@sqltools/formatter) vs [2.61MB](https://www.npmjs.com/package/sql-formatter)).

Another option to consider is installing this as an [optional peer dependency](https://docs.npmjs.com/cli/v11/configuring-npm/package-json#peerdependenciesmeta).

I'm concerned about the increase in size from requiring this new package. It's going to make TypeORM larger to install (20.5MB to 23MB), and most installs won't need a formatting option.

w3nl: I need to test it, but I think something like:
ac1f181b53cc4dae4cc11a22ef5beb3caca80398
It uses the same package now as we use to generate the migrations.

jovanadjuric (approved):

mguida22 (approved):
nice work on this!

gioboa (approved):

PR #11411, opened 2025-04-14
chore(docs): replace spaces in paths

mguida22:
Replaces all doc paths with a space to a lower-kebab-case style path.

URLs can't contain spaces, so they get encoded with `%20`. By removing the spaces we get cleaner URLs and file paths. Most changes are a simple find and replace, and `docs/sidebar` keeps the labels the same so there's no visual change on the docs site.


gioboa (approved):
👏@mguida22

jovanadjuric (approved):

PR #11413, opened 2025-04-14
chore(docs): use lowercase directory names

mguida22:
Renames docs directories to be lowercase.

Followup to #11411. Since some directories were only changed to lowercase, git originally ignored the change due to case insensitivity.


jovanadjuric (approved):

gioboa (approved):
Nice @mguida22 thanks for your commitment

Issue #10190, opened 2023-07-07
Cannot save a one-to-many relation


wkdwilliams:
### Issue description

Cannot save a one-to-many relation when pushing

### Expected Behavior

An appended entity should either be updated or created 

### Actual Behavior

Typeorm tries to set the foreign key to null.

Error:
`QueryFailedError: Column 'saleid' cannot be null`

Query Typeorm tries to run;
`query failed: UPDATE maindb.user SET user_id = ?, updated_at = CURRENT_TIMESTAMP WHERE id = ?
-- PARAMETERS: [null,"14"]`

### Create the two entities;
```typescript
@Entity('user')
export class User {
    @PrimaryGeneratedColumn('increment', { type: 'bigint' })
    id: number;

    @Column()
    name: string;

    @Column()
    username: string;

    @Column()
    password: string;

    @OneToMany(() => Post, p => p.user)
    posts: Post[];

}

@Entity('post')
export class Post {
    @PrimaryGeneratedColumn('increment', { type: 'bigint' })
    id: number;

    @Column()
    user_id: number;

    @Column()
    name: string;

    @Column()
    description: string;

    @ManyToOne(() => User, u => u.posts)
    @JoinColumn({ name: 'user_id' })
    user: User;
}
```

### The code below works ✅
```typescript
const post1 = new Post();
post1.name          = "test1";
post1.description   = "ima lima";

const post2 = new Post();
post2.name          = "test2";
post2.description   = "ima lima";

const user = new User();
user.name       = "John";
user.username   = "johnsmith";
user.password   = "Password123";
user.posts = [post1, post2];

await this.userRepository.save(user);
```

### The code below does not work ❌
```typescript
const post1 = new Post();
post1.name          = "test1";
post1.description   = "ima lima";

const user = new User();
user.name       = "John";
user.username   = "johnsmith";
user.password   = "Password123";
user.posts      = [post1];
/**
 * Create a new user with one post
 */
const created = await this.userRepository.save(user);

const post2 = new Post();
post2.name          = "test2";
post2.description   = "ima lima";

created.posts.push(post2)

await this.userRepository.save(created);
```

### The code below also doesn't work ❌
```typescript
const post1 = new Post();
post1.name          = "test1";
post1.description   = "ima lima";

const user = new User();
user.name       = "John";
user.username   = "johnsmith";
user.password   = "Password123";
user.posts      = [post1];
/**
 * Create a new user with one post
 */
const created = await this.userRepository.save(user);

[...]

PR #11417, opened 2025-04-15
test: add test for saving lazy many to one relation with existing entities

jonaskordt:
### Description of change

Adds a test for saving a lazy one to many relation. With non-lazy relations the save call binds the relations, with lazy one to many relations it does not.

I sadly don't know how to fix this, but @gioboa requested this PR with a test to demonstrate the problem.

Related to: #10190

Some more insights I gathered which might help to fix this:

Adding the following logs to `src/persistence/SubjectExecutor.ts` in line 618
```
console.log("updateMap", updateMap)
console.log(
    "updateQueryBuilder",
    updateQueryBuilder.getQueryAndParameters(),
)
```
gives the following output for the two tests:
```
// non-lazy version:
updateMap { user: { id: 'user-id' } }
updateQueryBuilder [
  'UPDATE "post" SET "user_id" = $1 WHERE "id" = $2',
  [ 'user-id', 'post-id' ]
]

// lazy version:
updateMap { user: { id: 'user-id' } }
updateQueryBuilder [
  'UPDATE "lazy_post" SET "user_id" = $1 WHERE "id" = $2',
  [ undefined, 'post-id' ]
]
```
So to me it looks like the user id is present in both cases, but with the lazy relation it does not get converted to the parameter properly.

### Pull-Request Checklist

### DIFF
--- a/test/github-issues/10190/issue-10190.ts
+++ b/test/github-issues/10190/issue-10190.ts
@@ -1,0 +1,97 @@
+import "reflect-metadata"
+import {
+    createTestingConnections,
+    closeTestingConnections,
+    reloadTestingDatabases,
+} from "../../utils/test-utils"
+import { DataSource } from "../../../src/data-source/DataSource"
+import { expect } from "chai"
+import { User } from "./entity/User"
+import { Post } from "./entity/Post"
+import { LazyPost } from "./entity/LazyPost"
+import { LazyUser } from "./entity/LazyUser"
+
+describe.only("github issues > #10190 Cannot save a one-to-many relation", () => {
+    let dataSources: DataSource[]
+    before(
+        async () =>
+            (dataSources = await createTestingConnections({
+                entities: [__dirname + "/entity/*{.js,.ts}"],
+                schemaCreate: true,
+                dropSchema: true,
+                enabledDrivers: ["postgres"],
+            })),
+    )
+    beforeEach(() => reloadTestingDatabases(dataSources))
+    after(() => closeTestingConnections(dataSources))
+
+    it("should update the user id of a post when the the user is saved with a list of posts that already exist in the database", () =>
+        Promise.all(
+            dataSources.map(async (dataSource) => {
+                const entityManager = dataSource.manager
+
+                // Create a post without a user
+                const post = new Post()
+                post.id = "post-id"
+                post.name = "Post"
+                const savedPost = await entityManager.save(post)
+
+                // Create a user
+                const user = new User()
+                user.id = "user-id"
+                user.name = "John"
+                await entityManager.save(user)
+
+                // Add the post to the user
+                const fetchedUser = await entityManager.findOneOrFail(User, {
+                    where: { id: user.id },
+                })
+                fetchedUser.posts = [savedPost]
+                await entityManager.save(fetchedUser)
+
+                // Check that the post has the user id
+                const updatedPost = await entityManager.findOneOrFail(Post, {
+                    where: { name: "Post" },
+                })
+                expect(updatedPost.userId).to.equal(user.id)
+            }),
+        ))
+
+    it("should update the user id of a post when the the user is saved with a list of posts that already exist in the database (with lazy relations)", () =>
+        Promise.all(
+            dataSources.map(async (dataSource) => {
+                const entityManager = dataSource.manager
+
+                // Create a post without a user
+                const post = new LazyPost()
+                post.id = "post-id"
+                post.name = "Lazy Post"
+                const savedPost = await entityManager.save(post)
+
+                // Create a user
+                const user = new LazyUser()
+                user.id = "user-id"
+                user.name = "Lazy John"
+                await entityManager.save(user)
+
+                // Add the post to the user
+                const fetchedUser = await entityManager.findOneOrFail(
+                    LazyUser,
+                    {
+                        where: { id: user.id },
+                    },
+                )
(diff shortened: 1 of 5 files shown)


gioboa (commented):
I'll look at this. Thanks for your help @jonaskordt

src/metadata/ColumnMetadata.ts
```diff
@@ -815,8 +815,10 @@ export class ColumnMetadata {
         } else {
             // no embeds - no problems. Simply return column name by property name of the entity
             if (this.relationMetadata && this.referencedColumn) {
-                const relatedEntity =
-                    this.relationMetadata.getEntityValue(entity)
+                const relatedEntity = this.relationMetadata.getEntityValue(
+                    entity,
+                    this.relationMetadata.isLazy,
```
gioboa, line 820: Internally `getEntityValue` method is checking `isLazy` too, but to avoid regressions I prefer to change only this piece of code.

The only other method that is using that flag is [this one](https://github.com/typeorm/typeorm/blob/fe71a0c3e441bc5fc68eec78c575d7ab5cc81bad/src/entity-manager/EntityManager.ts#L298)

jonaskordt:
@gioboa When do you think this will take it into a release?

gioboa (commented):
Hi @jonaskordt we are waiting for at least two reviews.

jonaskordt:
@gioboa what is the usual review timeline?

michaelbromley:
Hi,
I just checked out this PR locally and the strange thing is, when I comment out the change to ColumnMetadata.ts, the tests still pass.

Then I even added `expect(false).to.equal(true)` to the end of each test, and they still pass. So something's up there. Can anyone else confirm?

gioboa:
> Then I even added expect(false).to.equal(true) to the end of each test, and they still pass. So something's up there. Can anyone else confirm?

🤔 It's working fine in my local machine

Screenshot 2025-05-12 at 10 19 51

jonaskordt:
@michaelbromley Any updates on this? For me the test also fails as expected when I comment out the fix.

alumni (approved):

test/github-issues/10190/issue-10190.ts
```diff
@@ -11,7 +11,7 @@ import { Post } from "./entity/Post"
 import { LazyPost } from "./entity/LazyPost"
 import { LazyUser } from "./entity/LazyUser"
 
-describe.only("github issues > #10190 Cannot save a one-to-many relation", () => {
+describe("github issues > #10190 Cannot save a one-to-many relation", () => {
```
gioboa, line 14: Thanks @jonaskordt  nice catch

jonaskordt:
@gioboa any idea why one of the tests is failing?

jonaskordt:
> Hi, I just checked out this PR locally and the strange thing is, when I comment out the change to ColumnMetadata.ts, the tests still pass.
> 
> Then I even added `expect(false).to.equal(true)` to the end of each test, and they still pass. So something's up there. Can anyone else confirm?

Hey @michaelbromley! Could you take another look? Would be great if we could get this merged some time soon :)

src/metadata/ColumnMetadata.ts
sgarner: Is it simpler to just make this `true` instead? Since it only changes the behavior for relations which are lazy.

```suggestion
                    true,
```

test/github-issues/10190/issue-10190.ts
```diff
@@ -0,0 +1,97 @@
...
+    reloadTestingDatabases,
+} from "../../utils/test-utils"
+import { DataSource } from "../../../src/data-source/DataSource"
+import { expect } from "chai"
+import { User } from "./entity/User"
+import { Post } from "./entity/Post"
+import { LazyPost } from "./entity/LazyPost"
+import { LazyUser } from "./entity/LazyUser"
+
+describe("github issues > #10190 Cannot save a one-to-many relation", () => {
```
sgarner, line 14: We are moving away from creating tests for specific GitHub issues. It would be better to add these tests to the `test/functional/metadata-builder/column-metadata/column-metadata.ts` suite.

sgarner (approved):
Couple of changes I'd like to see, but the fix is good :+1:

jonaskordt:
> Couple of changes I'd like to see, but the fix is good 👍

Thanks for the feedback, I will hopefully get to implementing it soon.

@sgarner Do you have any idea why the one seemingly unrelated test is failing?

src/metadata/ColumnMetadata.ts
```diff
@@ -815,8 +815,10 @@ export class ColumnMetadata {
         } else {
             // no embeds - no problems. Simply return column name by property name of the entity
             if (this.relationMetadata && this.referencedColumn) {
-                const relatedEntity =
-                    this.relationMetadata.getEntityValue(entity)
+                const relatedEntity = this.relationMetadata.getEntityValue(
+                    entity,
+                    true,
```
gioboa, line 820: This change is breaking other tests 🙂‍↕️

gioboa (changes requested):

src/metadata/ColumnMetadata.ts
jonaskordt: Any idea how to fix it?

gioboa (commented):
@jonaskordt  thanks for opening this PR.
I'm closing this for now because I can't going on with the fix.
Anyway it's here for the record.

PR #11426, opened 2025-04-21
Test for Github issue 11421

strongpauly:
### Pull-Request Checklist

### DIFF
--- a/test/github-issues/11421/issue-11421.ts
+++ b/test/github-issues/11421/issue-11421.ts
@@ -1,0 +1,65 @@
+import "reflect-metadata"
+import {
+    createTestingConnections,
+    closeTestingConnections,
+} from "../../utils/test-utils"
+import { DataSource } from "../../../src/index.js"
+import { expect } from "chai"
+import { Book } from "./entity/Book"
+
+describe("github issues > #11421 findOne returns null if id is not selected", () => {
+    let dataSources: DataSource[]
+
+    before(
+        async () =>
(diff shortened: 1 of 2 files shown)


test/github-issues/11421/issue-11421.ts
```diff
@@ -0,0 +1,65 @@
...
+                book = await repository.save(book, { reload: true })
+
+                expect(book.id).to.be.a.string
+
+                const retrieved = await repository.findOne({
+                    where: { id: book.id },
+                    select: { author: true },
+                })
+
+                expect(retrieved).to.be.instanceOf(Book)
```
gioboa, line 61: I debugged the code and here is the piece of code that creates the [issue](https://github.com/typeorm/typeorm/blob/24a03696731ab212903f2173eacf586413da5d17/src/query-builder/transformer/RawSqlResultsToEntityTransformer.ts#L208)

In this case we are selecting a column with null value and `hasColumns` is false. The result of this function is undefined.
@alumni @naorpeled What do you think about it?
Should we add a fallback?

alumni: I remember that part from before the the performance improvement PR - I believe it was added for relations, i.e. when the related entity is selected, it will be left joined; if there are no results, we'd get null for all the columns of the related entity.

gioboa (commented):
@strongpauly thanks for opening this PR.
I'm closing this for now because it's stale and nobody is leading a fix.
anyway it's here for the record.

PR #10108, opened 2023-06-05
feat: add tagged template for executing raw SQL queries

Newbie012:
*I'm aware that this PR won't be accepted until I add the appropriate tests and documentation. I first want to make sure that this feature is acceptable to the core maintainers.*

*My apologies for not writing a separate issue for that. If you believe sql tags should not exist in TypeORM, I will understand. Although, it would be harder for other tools (such as [SafeQL](https://github.com/ts-safeql/safeql)) to integrate with TypeORM.*

---

### Description of change

Adds a new tagged template to classes that execute raw SQL queries and return raw database results. The template takes a template string and an array of values as input and builds a SQL query using the appropriate parameter strategy based on

While SQL tags can be misleading to some people regarding SQL injections, they are, in fact, safer than splitting the query from the variables. With tagged template, you simply can't go wrong with accidently injecting unsafe values from variables due to the way of [how tagged templates work](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals).

As a plus, reading the query with the variables in place adds more clarity rather than trying to figure out what is the $23 parameter.

The new `sql` tag will split the templates from the expressions, parameterize the statement, and then map the statement and the variables back to the `query` method. (tag -> { query, variables } -> query method).

### Pull-Request Checklist

- [x] Code is up-to-date with the `master` branch
- [x] `npm run format` to apply prettier formatting
- [x] `npm run test` passes with this change
- [N/A] This pull request links relevant issues as `Fixes #0000`
- [ ] There are new or updated unit tests validating the change
- [ ] Documentation has been updated to reflect this change
- [x] The new commits follow conventions explained in [CONTRIBUTING.md](https://github.com/typeorm/typeorm/blob/master/CONTRIBUTING.md)

PR #11432, opened 2025-04-29
feat: add tagged template for executing raw SQL queries

Newbie012:
### Description of change

replacing https://github.com/typeorm/typeorm/pull/10108

* Added a new SQL tag feature to the DataSource and EntityManager for executing raw SQL queries with automatic parameter handling.
* Implemented SqlTagUtils to manage parameter strategies based on database types.
* Created tests for SQL tag functionality, covering various parameter scenarios and edge cases.
* Updated documentation to include usage examples and benefits of the SQL tag feature.

### Pull-Request Checklist


src/util/SqlTagUtils.ts
```diff
@@ -0,0 +1,66 @@
...
+            case "aurora-mysql":
+                return "question-mark"
+            case "oracle":
+                return "colon"
+            case "mssql":
+                return "at"
+            default:
+                return "unknown"
+        }
+    }
```
sgarner, line 65: So that we don't need to maintain this mapping here, could `parameterStrategy` be a driver property instead?

Then `buildSqlTag()` could take the `dataSource` as a parameter instead of the `databaseType`. It would obtain the `parameterStrategy` from `dataSource.driver.parameterStrategy`.

I'm not sure that the `SqlTagUtils` class is necessary either, we could just export `buildSqlTag` as a standalone function.

sgarner (commented):
This is a sweet new feature!

One suggested change, otherwise looking good :+1:

src/util/SqlTagUtils.ts
Newbie012: Agreed. I'm not sure why I did that before (previous PR). I made the proper adjustments.

src/util/SqlTagUtils.ts
```diff
@@ -0,0 +1,49 @@
+import { Driver } from "../driver/Driver"
+
+export function buildSqlTag(params: {
+    driver: Driver
+    strings: TemplateStringsArray
+    expressions: unknown[]
+}): { query: string; variables: unknown[] } {
```
sgarner, line 7: Readability could be improved by destructuring `params`, i.e.

```ts
export function buildSqlTag({ driver, strings, expressions }: {...}
```

src/util/SqlTagUtils.ts
```diff
@@ -0,0 +1,49 @@
...
+    let query = ""
+    const variables: unknown[] = []
```
sgarner, line 9: Elsewhere in the codebase we use the term "parameters" rather than "variables" for these placeholders in SQL queries. So renaming this (and the returned property) would be better for consistency.

test/functional/sql-tag/sql-tag-parameters-mysql.ts
```diff
@@ -0,0 +1,189 @@
...
+                        OR FIND_IN_SET(${searchTags[1]}, tags) > 0
+                    )
+                `
+
+                const ids = examples.map((e: MysqlExample) => e.id)
+
+                expect(examples).to.have.length(2)
+                expect(ids).to.have.members(["array1", "array2"])
+            }),
+        ))
```
sgarner, line 188: This test doesn't prove anything since it's not actually binding an array value to the parameter. It should be replaced with a test that asserts the expected behavior of array parameters.

test/functional/sql-tag/sql-tag-parameters-mysql.ts
```diff
@@ -0,0 +1,189 @@
...
+                const nowStr = toMySQLDateString(now)
+                const yesterdayStr = toMySQLDateString(yesterday)
+
+                await repo.save([
+                    { id: "today", createdAt: nowStr },
+                    { id: "yesterday", createdAt: yesterdayStr },
+                ])
+
+                const examples = await connection.sql`
+                    SELECT * FROM example WHERE createdAt > ${yesterdayStr}
```
sgarner, line 151: In the equivalent test for Postgres, the `Date` value `yesterday` is used directly as the parameter, whereas here it's converted to a string first. I would expect `Date` values to be handled the same across all drivers, so it would be good to show they work for MySQL here.

test/functional/sql-tag/sql-tag-parameters-sqlite.ts
```diff
@@ -0,0 +1,212 @@
...
+            connections.map(async (connection) => {
+                const repo = connection.getRepository(Example)
+                const now = new Date()
+                const yesterday = new Date(now.getTime() - 24 * 60 * 60 * 1000)
+                const nowISO = now.toISOString()
+                const yesterdayISO = yesterday.toISOString()
+
+                await repo.save([
+                    { id: "today", createdAt: nowISO },
+                    { id: "yesterday", createdAt: yesterdayISO },
```
sgarner, line 170: Likewise with dates here, as per comment on the mysql test

sgarner (changes requested):

src/util/SqlTagUtils.ts
```diff
@@ -0,0 +1,49 @@
...
+
+        if (expression === null) {
+            query += "NULL"
+            continue
+        }
+
+        if (
+            ["sqlite", "better-sqlite3", "mysql"].includes(
+                params.driver.options.type,
+            ) &&
```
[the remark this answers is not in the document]
Newbie012: I've made the appropriate changes. The initial reason for choosing not to expand the array was due to the fact that there's a finite amount of parameters that you can use (at leas in PostgreSQL)

src/util/SqlTagUtils.ts
Newbie012: Created an interface and destructed `params`

src/util/SqlTagUtils.ts
Newbie012: My bad, fixed in DataSource and EntityManager as well.

test/functional/sql-tag/sql-tag-parameters-mysql.ts
Newbie012: You're right. I ended up doing a simpler query since the tests are running on MySQL 5.7 which lacks many features involved around arrays (8.0+)

test/functional/sql-tag/sql-tag-parameters-mysql.ts
sgarner: I'm confused... why were Dates working in Postgres without any serialization before - I assumed we already had some serialization in place, but is it actually because Postgres supports the default `Date.toString()` format and other databases don't?

I don't think we should be changing how dates work in raw queries in this PR. Which means the only serialization required is the expansion of function parameters, which can be internal to `buildSqlTag`.

sgarner:
> I rarely have single-line SQL template strings. Currently I use my own implementation which relies on `common-tags` (and `utility-types`) which keeps a consistent indentation in the SQL code. E.g.:
> 
> ...
> 
> Note there's no empty line before the query (empty lines at the beginning and end are trimmed) and there are 0 instead of 4 spaces of indentation (minimum indentation is 4 spaces, which is then removed from all lines).
> 
> Could we also have indentation-stripping here?

This would be a nice feature! But it could be implemented for `query()` as well, not just tagged sql literals, so I think it should be a separate PR?

alumni:
@sgarner it makes sense only for template strings

sgarner:
> @sgarner it makes sense only for template strings

But you can write your SQL in a template string with `query()` as well?

alumni:
@sgarner Maybe I should rephrase: I didn't say we should do ``.query(sql`...`)``. That won't work because we can't serialize all parameters (or we could, but we'd need to implement serialization for each driver, like DBeaver does). We can keep `` connection.sql`...` `` like in this PR, and simply call `.query(queryString, parameters)`. I'm just wondering why do we need the `serializeParameter` function, and if can remove the indentation from the query string, so the multi-line SQL code that the users write looks clean.

Now one more thing: I often see in projects that the template string expressions aren't always values that need escaping. Should we provide some helper function (e.g. `safeValue`, `escapeAlias`) that return branded strings which are handled differently? E.g.:
```ts
await connection.sql`
  SELECT *
  FROM ${escapeAlias(schema)}."SomeTable"
  WHERE ${safeValue(
    conditions.join(' AND ')
  )}
`;
```

test/functional/sql-tag/sql-tag-parameters-mysql.ts
Newbie012: Apparently it had to do with the fact that I had to explictly set the datetime precision. I removed that extra logic as I believe it doesn't belong in `.sql` which should simply be a thin wrapper around `.query`.

Newbie012:
@sgarner @alumni thanks for the detailed feedback. I'll try to briefly address your comments.

The `sql` tag implementation IMHO should be as simple as possible, as it's simply a thin layer over `.query`. Thus, serializing the expressions wouldn't be a good idea. I removed these lines.

Regarding indentation, I guess it would be useful when reading the query when debugging and such. I added [`dedent`](https://github.com/dmnd/dedent) which does a great job and is maintained.

@alumni about your last argument, while I can imagine people using it, I feel like this could potentially lead to exploitable queries when improperly used. I know it's highly subjective, but I personally prefer writing raw queries in their natural form as much as possible since it's easier to debug, port, and understand. As a bonus, it makes it easier for tools such as [SafeQL](https://github.com/ts-safeql/safeql) (creator) to statically analyze the queries, but as I mentioned, these are just my personal thoughts.

Maybe as an alternative, we could explicitly allow passing model classes such as  `SELECT * FROM ${EntityExample} WHERE ...`

Newbie012:
A small update about the date comparison date - I'm not sure exactly why it failed on mysql and sqlite (and not on PostgreSQL). I initially thought it might be related to the precision, but apparently it wasn't. I don't know how valuable this test was to this feature anyway.

alumni:
Since `sql` is syntactic sugar for `query`, there are two more places where we should probably implement this: `QueryRunner` and `Repository`.

We might need to convert `QueryRunner` to an abstract class (currently it's an interface). I'm not sure if this would cause any issues - but it would be nice to have it there if it's not too much trouble since `QueryRunner` allows running queries in the same DB connection.

Also: sometimes `query` is implemented with type, generics sometimes not. It would be nice to have a consistent behavior for both `query` and `sql`.

src/util/SqlTagUtils.ts
```diff
@@ -0,0 +1,61 @@
...
+    }
+
+    query += strings[strings.length - 1]
+
+    query = dedent(query)
+
+    return { query, parameters }
+}
+
+function toParameter(expression: unknown) {
```
alumni, line 55: probably we don't need to bother with this, the user can call the function manually 🤷🏻

Newbie012: It matters because of https://github.com/typeorm/typeorm/pull/11432#discussion_r2069537794

See the [following test](https://github.com/typeorm/typeorm/pull/11432/files#diff-1f7a118eed75b848ed6280db578e0af5e685221db97ce2cbe201a300bd83d689R208)

Newbie012:
@alumni I added `sql` to `Repository`. Regarding `QueryPlanner`, I'm not entirely sure how it should be implemented, since MongoQueryRunner implements this interface (which does not have sql).

In any case, I've added it to `Repository` as well and updated `query` to be generic as well, just like the others.

sgarner:
> @sgarner Maybe I should rephrase: I didn't say we should do `` .query(sql`...`) ``. That won't work because we can't serialize all parameters (or we could, but we'd need to implement serialization for each driver, like DBeaver does). We can keep `` connection.sql`...` `` like in this PR, and simply call `.query(queryString, parameters)`.

Sorry, I wasn't suggesting using ``sql`...` `` inside `.query()`. In the context of your comment about indentation, I was simply pointing out that you can write SQL in a template string (whether tagged or not) using `.query()` right now and have the same indentation issues you're describing. For example:

```ts
dataSource.query(`SELECT *
    FROM table_name
    WHERE foo = 1
`);
```

So if we are going to build in some feature around auto-formatting, both `.query()` and ``sql`...` `` could benefit from it. And since ``sql`...` `` calls `.query()` internally, let's do it there.

> why do we need the serializeParameter function

We don't, there was some misunderstanding around date parameters but @Newbie012 has removed this now. Aside from special handling for expanding lists, parameter values should be treated the same with the `sql` tag and with `.query()`.

> Now one more thing: I often see in projects that the template string expressions aren't always values that need escaping. Should we provide some helper function (e.g. `safeValue`, `escapeAlias`) that return branded strings which are handled differently? E.g.:
> 
> ```ts
> await connection.sql`
>   SELECT *
>   FROM ${escapeAlias(schema)}."SomeTable"
>   WHERE ${safeValue(
>     conditions.join(' AND ')
>   )}
> `;
> ```

That's a very good point. 🤔 

I would be disinclined to provide new helper functions for escaping as we already have `Driver.escape()` for those who need this, and it is generally best to avoid such dynamic code in raw queries. Moreover if you are going to formulate a where clause with dynamic parameters like that example, you're better to use `.query()` rather than the `sql` tag so that each condition can be parameterized.

But a way to interpolate raw strings into the SQL query _without_ them becoming database parameters would certainly be useful.

sgarner:
Now I'm thinking we should have gone the other way with arrays, and use functions to denote special values instead.

How about we use this approach for parameters in `sql` tagged templates:

#### By default, any ordinary non-function expression is bound to a database parameter:
  ```ts
  const foo = "hello world"
  datasource.sql`SELECT * FROM table_name WHERE foo = ${foo}`
  // query: SELECT * FROM table_name WHERE foo = ?
  // variables: ["hello world"]
  ```
  ```ts
  const someArray = [1, 2, 3]
  datasource.sql`SELECT * FROM table_name WHERE array_column = ${someArray}`
  // query: SELECT * FROM table_name WHERE array_column = ?
  // variables: [[1, 2, 3]]
  ```
 
#### A function expression which returns an array is expanded to a list of parameters:
  ```ts
  const someArray = [1, 2, 3]
  datasource.sql`SELECT * FROM table_name WHERE id IN (${() => someArray})`
  // query: SELECT * FROM table_name WHERE id IN (?, ?, ?)
  // variables: [1, 2, 3]
  ```

#### A function expression which returns a string is interpolated into the SQL verbatim:
  ```ts
  const schema = "schema_name"
  const foo = "hello world"
  datasource.sql`SELECT * FROM ${() => schema}.table_name WHERE foo = ${foo}`
  // query: SELECT * FROM schema_name.table_name WHERE foo = ?
  // variables: ["hello world"]
  ```

Perhaps a function expression which returns any other type should throw an error.

---

_Note:_ There is some prior art for this in how TypeORM handles column default values. If you define a column with `default: 'foo'` this is interpreted as assigning that string as the column's default value, whereas if you want to assign an SQL expression as the default value then you use a function, like: `default: () => 'GETDATE()'`.

test/functional/sql-tag/sql-tag-parameters-better-sqlite3.ts
```diff
@@ -0,0 +1,212 @@
...
+                    SELECT * FROM example WHERE "createdAt" > ${yesterdayISO}
+                `
+
+                const ids = examples.map((e: Example) => e.id)
+
+                expect(examples).to.have.length(1)
+                expect(ids).to.have.members(["today"])
+            }),
+        ))
+
```
sgarner, line 183: I think we can drop this test, since we don't expect the `sql` tag to do anything special with dates - and this test is binding a string to the parameter anyway.

```suggestion
```

test/functional/sql-tag/sql-tag-parameters-postgres.ts
```diff
@@ -0,0 +1,246 @@
...
+                    SELECT * FROM example WHERE "createdAt" > ${yesterday}
+                `
+
+                const ids = examples.map((e: PostgresExample) => e.id)
+
+                expect(examples).to.have.length(1)
+                expect(ids).to.have.members(["today"])
+            }),
+        ))
+
```
sgarner, line 179: Likewise, although this test is actually binding a Date object to the parameter, it's only proving how the driver handles such values, not anything about the `sql` tag?
```suggestion
```

test/functional/sql-tag/sql-tag-parameters.ts
```diff
@@ -0,0 +1,5 @@
+// Tests have been split into driver-specific files:
+// - sql-tag-parameters-sqlite.ts
+// - sql-tag-parameters-better-sqlite3.ts
+// - sql-tag-parameters-mysql.ts
```
sgarner, line 4: Do we need this file? It's probably obvious enough how the tests are split.

If we keep it, ensure it references all the other files:

```suggestion
// - sql-tag-parameters-mysql.ts
// - sql-tag-parameters-postgres.ts
```

Newbie012: that file was committed by mistake, sorry.

Newbie012:
@sgarner I swapped the callback logic so when it's wrapped inside a callback it will spread the array.

Additionally, I wrote a new test which I believe makes more sense than the others which test exactly what I should've tested in the first place. I'll keep the other tests as well

sgarner:
> @sgarner I swapped the callback logic so when it's wrapped inside a callback it will spread the array.

Thanks, can you add support for this case as well?

> #### A function expression which returns a string is interpolated into the SQL verbatim:
> ```ts
> const schema = "schema_name"
> const foo = "hello world"
> datasource.sql`SELECT * FROM ${() => schema}.table_name WHERE foo = ${foo}`
> // query: SELECT * FROM schema_name.table_name WHERE foo = ?
> // variables: ["hello world"]
> ```
> 
> Perhaps a function expression which returns any other type should throw an error.

Newbie012:
Are you sure? this pattern looks really unsafe, and I can imagine people misuse it. If we want to allow such pattern, wouldn't an explicit function would be "safer"?

```ts
datasource.sql`SELECT * FROM ${({ unescape }) => unescape(schema)}.table_name WHERE foo = ${foo}`
datasource.sql`SELECT * FROM schema.table_name WHERE foo = IN (${({ spread }) => spread(array)})`
```

alumni:
> Are you sure? this pattern looks really unsafe, and I can imagine people misuse it. If we want to allow such pattern, wouldn't an explicit function would be "safer"?
> 
> ```ts
> datasource.sql`SELECT * FROM ${({ unescape }) => unescape(schema)}.table_name WHERE foo = ${foo}`
> datasource.sql`SELECT * FROM schema.table_name WHERE foo = IN (${({ spread }) => spread(array)})`
> ```

I agree with @sgarner on some things:

TypeORM always uses the function result as _raw SQL_. E.g. if the default of a column is a function, the function must return valid SQL for the default. It is never escaped. So this would be okay:
```ts
datasource.sql`SELECT * FROM ${ () => schema }."table_name" WHERE "foo" = ${ foo }`
```

Regarding arrays, the main issue that I see is that Buffer/Uint8Array are also arrays. These values should be passed to the driver regardless, so I should be able to do:
```ts
datasource.sql`INSERT INTO "users"("guid") VALUES(${ uuidv4() })`
```

I haven't used array columns, but the right part of the predicate `"id" IN (1, 2, 3)` is actually an array. And the things can get more complex with `("id", "type") IN ((1, 1), (2, 1), (3, 2))` - not all databases support tuples, but some for sure do, as this is valid SQL-92 syntax.

I don't have a strong opinion about how we should support arrays, just wanted to raise some potential edge cases.

sgarner:
> Are you sure? this pattern looks really unsafe, and I can imagine people misuse it. If we want to allow such pattern, wouldn't an explicit function would be "safer"?

Whether it is unsafe depends on the source of the values being interpolated. As long as it is not user input, then safety isn't a concern.

So the question is, where should the safety threshold be set? Can we expect that if an author writes a raw SQL string and includes a placeholder value wrapped in a function, that they know what they're doing? I tend to think we can. Another level of wrapping with an `unsafe` function would add a bit too much clutter, in my opinion.

sgarner:
> Regarding arrays, the main issue that I see is that Buffer/Uint8Array are also arrays. These values should be passed to the driver regardless, so I should be able to do:
> 
> ```ts
> datasource.sql`INSERT INTO "users"("guid") VALUES(${ uuidv4() })`
> ```
> 
> I haven't used array columns, but the right part of the predicate `"id" IN (1, 2, 3)` is actually an array. And the things can get more complex with `("id", "type") IN ((1, 1), (2, 1), (3, 2))` - not all databases support tuples, but some for sure do, as this is valid SQL-92 syntax.

I don't believe this will be a problem - as long as your array or array-like value is not wrapped in a function, then we'll just pass it along to the driver to be bound as a parameter, and it will work like it does now with `query()`.

If you wrap an array in a function then we'll expand it out to multiple parameters, one for each element of the array (equivalent to how QueryBuilder handles `:...param` syntax). Each of those expanded parameters can be an array if the database supports arrays.

Newbie012:
Ok, so I would like to summarize the callback behavior: strings are interpolated into the SQL verbatim, and arrays are spread as multiple parameters.

sgarner:
> Ok, so if I would like to summarize the callback behavior - strings are interpolated into the SQL verbatim, and arrays are spread as multiple parameters?

Correct 🙂

docs/data-source-api.md
```diff
@@ -176,6 +176,11 @@ const rawData = await dataSource.query(
     'SELECT * FROM USERS WHERE name = @0 and age = @1',
     [ 'John', 24 ]
 )
+
+-   `sql` - Executes a raw SQL query using template literals.
+
+```typescript
+const rawData = await dataSource.sql`SELECT * FROM USERS WHERE name = ${'John'} and age = ${24}`
 ```
```
sgarner, line 184: We're missing the closing backticks on the code block above this. And let's link to the new doc page.

````suggestion
```

-   `sql` - Executes a raw SQL query using template literals.

```typescript
const rawData = await dataSource.sql`SELECT * FROM USERS WHERE name = ${'John'} and age = ${24}`
```

Learn more about using the [SQL Tag syntax](sql-tag.md).
````

File: docs/entity-manager-api.md [hunk 13]
```diff
@@ -63,6 +63,11 @@ const rawData = await manager.query(
     'SELECT * FROM USERS WHERE name = @0 and age = @1',
     [ 'John', 24 ]
 )
+
+-   `sql` - Executes a raw SQL query using template literals.
+
+```typescript
+const rawData = await manager.sql`SELECT * FROM USERS WHERE name = ${'John'} and age = ${24}`
 ```
```
sgarner, line 71: Same again here.

````suggestion
```

-   `sql` - Executes a raw SQL query using template literals.

```typescript
const rawData = await manager.sql`SELECT * FROM USERS WHERE name = ${'John'} and age = ${24}`
```

Learn more about using the [SQL Tag syntax](sql-tag.md).
````

File: docs/sql-tag.md [hunk 14]
```diff
@@ -0,0 +1,78 @@
+# SQL Tag
+
+TypeORM provides a way to write SQL queries using template literals with automatic parameter handling based on your database type. This feature helps prevent SQL injection while making queries more readable. The SQL tag is implemented as a wrapper around the `.query` method, providing an alternative interface while maintaining the same underlying functionality.
+
+## Basic Usage
+
+The SQL tag is available on your DataSource instance:
```
sgarner, line 7:
```suggestion
The `sql` tag is available on DataSource, EntityManager, Repository and QueryRunner instances:
```

docs/sql-tag.md
```diff
@@ -0,0 +1,78 @@
...
+const age = 30
+const active = true
+const users = await dataSource.sql`
+    SELECT * FROM users
+    WHERE name LIKE ${name + "%"}
+    AND age > ${age}
+    AND is_active = ${active}
+`
+```
+
```
sgarner, line 56:
````suggestion

### Expanding Parameter Lists

To transform an array of values into a dynamic list of parameters in a template expression, wrap the array in a function. This is commonly used to write an `IN (...)` expression in SQL, where each value in the list must be supplied as a separate parameter:

```typescript
// Query becomes: SELECT * FROM users WHERE id IN (?, ?, ?)
const users = await dataSource.sql`
    SELECT * FROM users
    WHERE id IN (${() => [1, 2, 3]})
`
```

### Interpolating Unescaped Expressions

When you want to insert a template expression which should _not_ be transformed into a database parameter, wrap the string in a function. This can be used to dynamically define column, table or schema names which can't be parameterized, or to conditionally set clauses in the SQL.

**Caution!** No escaping is performed on raw SQL inserted in this way. It is not safe to use this with values sourced from user input.

```typescript
// Query becomes: SELECT * FROM dynamic_table_name
const rawData = await dataSource.sql`
    SELECT * FROM ${() => "dynamic_table_name"}
`
```

````

**Review by sgarner — CHANGES_REQUESTED** [review]
Just a few more changes. Thanks for all your work on this @Newbie012 !

File: src/util/SqlTagUtils.ts [hunk 16]
```diff
@@ -0,0 +1,70 @@
...
+            if (typeof value === "string") {
+                query += value
+                continue
+            }
+
+            if (Array.isArray(value)) {
+                if (value.length === 0) {
+                    query += "NULL"
+                    continue
+                }
```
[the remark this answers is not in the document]
Newbie012: Possible, but what about `select true where 1 in (select 0 where false)`? or maybe it's too hacky

src/util/SqlTagUtils.ts
```diff
@@ -0,0 +1,70 @@
...
+                const arrayParams = value.map(() => {
+                    return driver.createParameter(`param_${idx + 1}`, idx++)
+                })
+
+                query += arrayParams.join(", ")
+                parameters.push(...value.map((e) => toParameter(e)))
+
+                continue
+            }
+        }
```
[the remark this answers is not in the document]
Newbie012: If `.query` handles function arguments, shouldn't it handle this case as well?

src/driver/aurora-postgres/AuroraPostgresQueryRunner.ts
```diff
@@ -194,6 +196,36 @@ export class AuroraPostgresQueryRunner
...
+        query: string,
+        parameters?: any[],
+        onEnd?: Function,
+        onError?: Function,
+    ): Promise<ReadStream> {
+        throw new TypeORMError(
+            "Streaming is not directly supported by AuroraPostgresQueryRunner using Data API client.",
+        )
+    }
+
```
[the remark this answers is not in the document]
Newbie012: Probably model hallucination. Thanks for noticing 🤦

alumni (approved):

src/util/SqlTagUtils.ts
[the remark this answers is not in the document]
sgarner: > If `.query` handles function arguments, shouldn't it handle this case as well?

I'm not aware that it does?

src/data-source/DataSource.ts
```diff
@@ -543,6 +544,26 @@ export class DataSource {
         }
     }
 
+    /**
+     * Tagged template function that executes raw SQL query and returns raw database results.
+     * Template expressions are automatically transformed into database parameters.
+     * Raw query execution is supported only by relational databases (MongoDB is not supported).
+     * Note: Don't call this as a regular function, it is meant to be used with backticks to tag a template literal.
+     * Example: repository.sql`SELECT * FROM table_name WHERE id = ${id}`
```
sgarner, line 552:
```suggestion
     * Example: dataSource.sql`SELECT * FROM table_name WHERE id = ${id}`
```

src/entity-manager/EntityManager.ts
```diff
@@ -177,6 +178,25 @@ export class EntityManager {
         return this.connection.query(query, parameters, this.queryRunner)
     }
 
+    /**
+     * Tagged template function that executes raw SQL query and returns raw database results.
+     * Template expressions are automatically transformed into database parameters.
+     * Note: Don't call this as a regular function, it is meant to be used with backticks to tag a template literal.
```
sgarner, line 184: I suppose we should mention Mongo here as well
```suggestion
     * Raw query execution is supported only by relational databases (MongoDB is not supported).
     * Note: Don't call this as a regular function, it is meant to be used with backticks to tag a template literal.
```

src/query-runner/BaseQueryRunner.ts
```diff
@@ -111,6 +112,26 @@ export abstract class BaseQueryRunner {
         useStructuredResult?: boolean,
     ): Promise<any>
 
+    /**
+     * Tagged template function that executes raw SQL query and returns raw database results.
+     * Template expressions are automatically transformed into database parameters.
+     * Raw query execution is supported only by relational databases (MongoDB is not supported).
+     * Note: Don't call this as a regular function, it is meant to be used with backticks to tag a template literal.
+     * Example: repository.sql`SELECT * FROM table_name WHERE id = ${id}`
```
sgarner, line 120:
```suggestion
     * Example: queryRunner.sql`SELECT * FROM table_name WHERE id = ${id}`
```

src/query-runner/QueryRunner.ts
```diff
@@ -126,6 +126,17 @@ export interface QueryRunner {
      */
     query(query: string, parameters?: any[]): Promise<any>
 
+    /**
+     * Tagged template function that executes raw SQL query and returns raw database results.
+     * Template expressions are automatically transformed into database parameters.
+     * Note: Don't call this as a regular function, it is meant to be used with backticks to tag a template literal.
```
sgarner, line 132:
```suggestion
     * Raw query execution is supported only by relational databases (MongoDB is not supported).
     * Note: Don't call this as a regular function, it is meant to be used with backticks to tag a template literal.
```

src/util/SqlTagUtils.ts
```diff
@@ -0,0 +1,61 @@
...
+                })
+
+                query += arrayParams.join(", ")
+                parameters.push(...value)
+
+                continue
+            }
+
+            throw new Error(
+                `Only array and strings are supported as function arguments. Got ${typeof value} instead.`,
```
sgarner, line 47:
```suggestion
                `Expression ${expressionIdx} in sql tagged template returned a value of type "${value === null ? "null" : typeof value}". Only array and string types are supported as function return values in sql tagged template expressions.`,
```

test/functional/sql-tag/sql-tag-parameters-postgres.ts
```diff
@@ -0,0 +1,223 @@
...
+                `
+
+                const ids = examples.map((e: PostgresExample) => e.id)
+
+                expect(examples).to.have.length(2)
+                expect(ids).to.have.members(["array1", "array2"])
+            }),
+        ))
+
+    it("should throw an error when exceeding the maximum number of parameters", async () => {
```
sgarner, line 185: What is the purpose of this test? 🤔

test/functional/sql-tag/sql-tag-parameters.ts
```diff
@@ -0,0 +1,130 @@
...
+            "id2",
+            "id3",
+            "id4",
+            "multiSpread",
+        ])
+    })
+
+    it("should throw an error when passing an invalid value inside a function argument", () => {
+        expect(() => sql`SELECT * FROM example WHERE id = ${() => 1}`).to.throw(
+            "Only array and strings are supported as function arguments. Got number instead.",
```
sgarner, line 127:
```suggestion
            `Expression 0 in sql tagged template returned a value of type "number". Only array and string types are supported as function return values in sql tagged template expressions.`
```

sgarner (changes requested):
Nearly there! Just a few final adjustments

sgarner:
> In the future, if this works well, we could also consider using template strings in migrations.

I'm not sure how useful raw SQL will be in generated migrations, since every database engine has proprietary syntax for a lot of DDL stuff - which we currently handle with `QueryRunner` methods. (Of course, if users are writing migrations by hand they can use it.)

sgarner (approved):
To avoid another review cycle, I committed my suggested changes. There is still one superfluous test which I think could be dropped.

Otherwise this looks good to go :+1:

Thanks again @Newbie012 !

test/functional/sql-tag/sql-tag-parameters-postgres.ts
Newbie012: I wrote this test when we initially defaulted to spreading arrays rather than using a single parameter. Note that it's not the case, this test is useless as it doesn't add any value.

src/util/SqlTagUtils.ts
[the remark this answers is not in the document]
Newbie012: Oh my mistake. For some reason I was under the assumption that we're potentially passing functions inside the parameters. We're good 🚀

macksal:
> > In the future, if this works well, we could also consider using template strings in migrations.
> 
> I'm not sure how useful raw SQL will be in generated migrations, since every database engine has proprietary syntax for a lot of DDL stuff - which we currently handle with `QueryRunner` methods. (Of course, if users are writing migrations by hand they can use it.)

@sgarner it seems like you're saying raw SQL is only useful in hand-written migrations, but isn't it the opposite? The generated migrations I get are always using raw SQL rather than the QueryRunner methods. I can only see those methods being used in hand-written migrations.

I found myself here thinking it would be nice for the generated queries to be tagged with ``` sql`` ``` automatically for tooling purposes.

alumni:
Actually, migrations contain driver-dependent raw sql, but wrapped in JS/TS, it's currently executed with `queryRunner.query`.

It could be easily replaced by SQL templates strings. The advantage for it is that it would make migrations easier to review, since you can get syntax highlighting for the query (in e.g. BitBucket it's built-in, in code editors it's available via plug-ins).

Replacing `queryRunner.query` with `queryRunner.sql` should be a simple change, can be done quite easily if someone is willing to open a PR.

I'm thinking to improve migration generation to be more flexible overall. I currently have to write a script that modifies the generated migration to work for me. But that's something more complex for which I need quite some time and I'd rather do other things in TypeORM which are more important.

sgarner:
Apologies, migration generation isn't a feature I use myself. I wrongly thought it would be using the QueryRunner methods for schema modification, but you're right it does generate raw SQL (I wonder why? 🤔).

In that case I agree that switching to SQL template strings sounds like a useful improvement.

JJoshMa:
## Design Question: Security implications of function returns in SQL tagged templates

### Overview
I'm exploring the SQL tagged template feature and have questions about the security design, particularly regarding functions that return string values.

### Current Implementation Analysis
In the `buildSqlTag` function, I noticed this behavior:
```


if (typeof expression === "function") {
    const value = expression()
    if (typeof value === "string") {
        // String values are directly concatenated into the SQL
        query += value
        continue
    }

   // ....
}
```
 @alumni

alumni:
Hey @JJoshMa, this was intentionally added to allow the users to write raw SQL when needed. Similar functionality is provided in other places, e.g. column defaults.

We expect that the values provided are already sanitized. If you work with unsafe values, you might want to use the `query` function instead of `sql`.

See https://typeorm.io/docs/guides/sql-tag#interpolating-unescaped-expressions.

JJoshMa:
> 嘿@JJoshMa，这是故意添加的，允许用户在需要时编写原始SQL。在其他地方也提供了类似的功能，例如列默认值。
> 
> 我们
> 希望提供的值已经被净化。如果使用不安全的值，您可能希望使用`query`函数而不是`sql`。
> 
> 参
> 见[https：//typeorm.io/docs/guides/sql-tag#interpolating-unescaped-expressions](https://typeorm.io/docs/guides/sql-tag#interpolating-unescaped-expressions)。

Thank you for the clarification. I appreciate your help.

PR #11434, opened 2025-05-01
feat: release PR releases using pkg.pr.new

naorpeled:
## Description
added the initial setup for [pkg.pr.new](https://pkg.pr.new/)

Note: when opening the example in Stackblitz the package.json looks like in the attached image but it's still working and if you save/format the file it looks good.
I couldn't figure out why it's happening, if you have an idea let me know.

image


gioboa (approved):
It looks great @naorpeled 👏👏
Well done ✅

package.json
```diff
@@ -1,7 +1,6 @@
 {
   "name": "typeorm",
   "version": "0.3.22",
-  "private": true,
```
alumni, line 4: This was a safeguard to avoid accidentally publishing the wrong build. Could we undo this?

.github/workflows/preview.yml
```diff
@@ -0,0 +1,44 @@
...
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: Use Node.js
+        uses: actions/setup-node@v4
+        with:
+          cache: "npm"
+          node-version: "20.x"
```
alumni, line 35: use `20` instead of `"20.x"` for consistency

.github/workflows/preview.yml
```diff
@@ -0,0 +1,44 @@
...
+
+      - name: Install dependencies
+        run: npm ci
+
+      - name: Build
+        run: npm run package
+
+      - name: Release preview version
+        run: npx pkg-pr-new publish './build/package' --template='./example'
```
alumni, line 44: Can we avoid using `--template` and just publish?

package.json
naorpeled: It didn't work with it, I'll see if I can somehow bypass it

.github/workflows/preview.yml
naorpeled: We can. Why do you think we should do so?
Personally, I think that the template is a nice thing to have.

.github/workflows/preview.yml
naorpeled: in the [publish workflow[https://github.com/typeorm/typeorm/blob/013fb4e8d5e4f6fb35a939c57767990011bcb6c3/.github/workflows/publish-package.yml) we're using this syntax, should I also align that file in this PR?

example/.prettierrc.json
alumni, line 1: I think this can be removed from the example to keep things minimal (unless it is used somehow?)

example/package.json
```diff
@@ -0,0 +1,22 @@
...
+  "description": "",
+  "main": "index.js",
+  "scripts": {
+    "start": "ts-node src/index.ts",
+    "build": "tsc",
+    "typeorm": "typeorm-ts-node-commonjs"
+  },
+  "dependencies": {
+    "typeorm": "0.3.22",
+    "sql.js": "^1.8.0",
```
alumni, line 13: `sql.js@1.13.0` and `typescript@5.8.3` are newer.

The file could be sorted with `npx sort-package-json`.

example/src/data-source.ts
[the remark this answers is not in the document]
naorpeled: You're absolutely right, great feedback, thanks!

example/src/index.ts
```diff
@@ -0,0 +1,25 @@
+import { AppDataSource } from "./data-source"
+import { User } from "./entity/User"
+
+AppDataSource.initialize()
```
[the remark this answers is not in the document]
naorpeled: 100%, done 😎

.github/workflows/preview.yml
naorpeled: Sounds good to me, done 🙏

.github/workflows/preview.yml
```diff
@@ -0,0 +1,44 @@
+name: PR preview release
+
+on:
+  pull_request:
+    paths-ignore:
+      - ".github/workflows/test.yml"
```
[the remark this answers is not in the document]
alumni: I'm not sure how github works in particular.

Might actually be easier to add `paths` instead of `paths-ignore`:
```yaml
pull_request:
  paths:
    - ".github/workflows/preview.yml"
    - "src/**"
    - "gulpfile.ts"
    - "package.json"
```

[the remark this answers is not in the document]
naorpeled: You're right, done 🔥

sample/playground/src/index.ts
```diff
@@ -0,0 +1,30 @@
...
+        console.log("All users:", users)
+
+        await AppDataSource.destroy()
+    } catch (error) {
+        console.error("Error during Data Source initialization:", error)
+        process.exit(1)
+    }
+}
+
+main()
```
alumni, line 30: `await main()`?

BTW, simply `main()` works with CJS too.

alumni (approved):

naorpeled:
Thanks for the awesome feedback @alumni!

sample/playground/src/ormconfig.ts
```diff
@@ -0,0 +1,10 @@
+import "reflect-metadata"
+import { DataSource } from "typeorm"
+import { User } from "./entity/User.js"
+
+export const AppDataSource = new DataSource({
```
alumni, line 5: Can you check if it works with the CLI? `typeorm -d ormconfig.ts migration:generate Test`

I'm sure it works with `export default new DataSource()`, but haven't use it like this. Entity/migration classes work even when there's no default export, but ormconfig probably doesn't use class loader.

naorpeled: I did, it works 🙏

PR #11438, opened 2025-05-04
docs: clarify where to add new tests

mguida22:
### Description of change

Clarify where to add new tests in DEVELOPER.md.

### Pull-Request Checklist

### DIFF
--- a/DEVELOPER.md
+++ b/DEVELOPER.md
@@ -89,7 +89,11 @@ You can copy this tar into your project and run `npm install ./typeorm-x.x.x.tgz
 
 ## Running Tests Locally
 
-It would be greatly appreciated if PRs that change code come with appropriate tests.
+It is greatly appreciated if PRs that change code come with appropriate tests.
+
+Consider where to put tests based on what type of change your PR introduces:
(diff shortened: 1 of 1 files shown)


naorpeled (approved):

gioboa (approved):

jovanadjuric (approved):

sgarner:
Totally endorse this. In fact I would go further and question whether we want to keep adding issue-based tests at all?

The problem with that approach, as I see it, is it results in pieces of functionality being tested from many different angles (a good thing) but each time in a different test suite using different test entities (a not so good thing). As well as making it hard to find all the tests involving a piece of functionality, it also bloats out the tests overall, contributing to slow test performance.

If there is a bug or regression in FindOptions, for example, why wouldn't we add relevant tests asserting the fix (with comments linking to the issue) to the relevant FindOptions test suite(s), rather than to bespoke tests for that issue?

mguida22:
@sgarner those are good points. I agree, I updated this PR to remove mention of creating a GitHub Issue specific test.

PR #11448, opened 2025-05-06
chore: Add husky and lint-staged

### LINKED ISSUE #11443: Add lint staged and husky
# Problem

While I was adding the fixes to the repository, I forgot to run the `format` command to lint the project. This has a negative impact on the developer because it causes another commit for linting, which triggers CI jobs and uses resources not intended for the primary purpose of the PR.

# Solution

Use pre-commit or pre-push [git hooks](https://git-scm.com/book/en/v2/Customizing-Git-Git-Hooks) for linting project. [Husky](https://www.npmjs.com/package/husky) and [lint-staged](https://www.npmjs.com/package/lint-staged) libraries implement this functionality in JS ecosystem

maxbronnikov10:
### Description of change

Added husky for running pre-commit which lint and format files through lint-staged config
Issue -  #11443

### Pull-Request Checklist
- [ ] `npm run format` to apply prettier formatting - N/A
- [ ] `npm run test` passes with this change - N/A
- [ ] There are new or updated unit tests validating the change - N/A


package.json
```diff
@@ -258,5 +262,11 @@
     "sqlite-orm",
     "sql-server",
     "sql-server-orm"
-  ]
+  ],
+  "lint-staged": {
```
alumni, line 266: We should avoid adding tool config to `package.json` - this file should only be used by the package manager (`npm` in our case).

`lint-staged` config should go into `.lintstagedrc.json`.

package.json
```diff
@@ -78,6 +78,8 @@
     "compile": "rimraf ./build && tsc",
     "format": "prettier --cache --write \"./**/*.ts\"",
     "format:ci": "prettier --check \"./**/*.ts\"",
+    "prepare": "husky",
```
alumni, line 81: Scripts should be sorted. Please run `npx sort-package-json`.

This could also be handled by `lint-staged` (for all json files) 🤷🏻

package.json
```diff
@@ -258,5 +262,11 @@
     "sqlite-orm",
     "sql-server",
     "sql-server-orm"
-  ]
+  ],
+  "lint-staged": {
+    "*.ts": [
+      "prettier --write",
+      "eslint --fix --max-warnings 0"
```
alumni, line 269: We have a few thousand warnings, some of them won't go away any time soon. So maybe we should allow warnings?

maxbronnikov10: This command only applies to new changes. I think without that we still get warnings and don't fix it.

package.json
maxbronnikov10: Maybe add plugin for prettier? https://www.npmjs.com/package/prettier-plugin-sort-json

package.json
maxbronnikov10: Or maybe create an additional issue where all eslint warnings are fixed, and apply that setting later on

package.json
alumni: That plugin doesn't work on `package.json`:
> This will not sort `package.json`, `package-lock.json`, or `composer.json`. This plugin only affects the `json` parser used by Prettier. Prettier uses an alternate parser (`json-stringify`) for those three specific files

There is an eslint plugin, but honestly setting that up just for a few config files is a bit too much :)

alumni (approved):

mguida22 (approved):
thanks!

gioboa (approved):
Thanks for your amazing help 👏

package.json
alumni: @maxbronnikov10 I somehow thought `--max-warnings 0` was removed. It's now preventing people from contributing, since almost every commit fails.

alumni: Ah, also `--fix` needs to be removed because it's automatically fixing `@typescript-eslint/no-wrapper-object-types`. In some cases, replacing `Object` with `object` messes things up.

If we manage to find a way to keep `--fix`, formatting should happen after linting, because fixing lint errors can change the code.

maxbronnikov10: maybe just disable rule `@typescript-eslint/no-wrapper-object-types` and keep `--fix`?

Issue #11440, opened 2025-05-05
regression for insert orUpdate with Postgres driver


yurks:
### Issue description

After releasing https://github.com/typeorm/typeorm/pull/11082 PR the SQL query is generated incorrectly if using `schema` property for entity

### Expected Behavior

`..."schemaName"."Category"."name"  IS DISTINCT FROM...` should be generated

### Actual Behavior

`..."schemaName.Category"."name"  IS DISTINCT FROM...` actually generated

### Steps to reproduce

set `schema` property and run `.insert().into(Entity).values(values).orUpdate()`

### My Environment

| Dependency          | Version  |
| ---                 | ---      |
| Operating System    |     osx 15.4.1 (24E263)    |
| Node.js version     | 22  |
| Typescript version  | 5.8.3 |
| TypeORM version     | 0.3.22 |

### Relevant Database Driver(s)

- [ ] aurora-mysql
- [ ] aurora-postgres
- [ ] better-sqlite3
- [ ] cockroachdb
- [ ] cordova
- [ ] expo
- [ ] mongodb
- [ ] mysql
- [ ] nativescript
- [ ] oracle
- [x] postgres
- [ ] react-native
- [ ] sap
- [ ] spanner
- [ ] sqlite
- [ ] sqlite-abstract
- [ ] sqljs
- [ ] sqlserver

### Are you willing to resolve this issue by submitting a Pull Request?

No, I don’t have the time, but I can support (using donations) development.

PR #11082, opened 2024-10-02
fix: incorrect table alias in insert orUpdate with Postgres driver

Ben1306:
Closes : #11077

### Description of change

When using orUpdate method in an insert query with
- skipUpdateIfNoValuesChanged
- postgres driver

produced SQL does not use table alias in WHERE clause which leads to **throwing an error**.

#### Expected Behavior
```
INSERT INTO "product_category"
AS "ProductCategory"("_id", "name")
VALUES ($1, $2), ...
ON CONFLICT ( "_id" )
DO UPDATE SET "name" = EXCLUDED."name"
WHERE ("ProductCategory"."name" IS DISTINCT FROM EXCLUDED."name")  RETURNING "id"
```

#### Actual Behavior
```
INSERT INTO "product_category"
AS "ProductCategory"("_id", "name")
VALUES ($1, $2), ...
ON CONFLICT ( "_id" )
DO UPDATE SET "name" = EXCLUDED."name"
WHERE ("product_category"."name" IS DISTINCT FROM EXCLUDED."name")  RETURNING "id"
```

### Pull-Request Checklist

- [x] Code is up-to-date with the `master` branch
- [x] `npm run format` to apply prettier formatting
- [x] `npm run test` passes with this change
- [x] This pull request links relevant issues as `Fixes #0000`
- [x] There are new or updated unit tests validating the change
- [x] Documentation has been updated to reflect this change
- [x] The new commits follow conventions explained in [CONTRIBUTING.md](https://github.com/typeorm/typeorm/blob/master/CONTRIBUTING.md)

PR #11452, opened 2025-05-07
fix: resolve alias or table name in upsert and orUpdate for PostgreSQL driver conditionally

mmarifat:
Closes: #11440

### Description of change

When using upsert or orUpdate query with `schema name` and `skipUpdateIfNoValuesChanged` in entity produces wrong SQL that throws an error

Expected Behavior
`..."schemaName"."tableName"."name"  IS DISTINCT FROM...` should be generated

Actual Behavior
`..."schemaName.tableName"."name"  IS DISTINCT FROM...` actually generated

This fix addresses a regression introduced in pull request #11082 by ensuring correct handling of alias names and table names when they are equal or distinct when entities use schema. When alias name does not match with main table name the escaped alias is used or else the table name is used in a proper format.

### Pull-Request Checklist

## Summary by CodeRabbit

- **New Features**
  - Added support for handling upsert and update operations with schema names and the skip-update-if-no-values-changed option for PostgreSQL-compatible databases.
- **Tests**
  - Introduced new tests to verify correct behavior of upsert and update operations with schema-qualified tables and conflict resolution.
- **Refactor**
  - Improved internal logic for generating SQL expressions to ensure consistent handling of table names and aliases in conflict resolution scenarios.


alumni (approved):

test/github-issues/11440/issue-11440.ts
```diff
@@ -0,0 +1,105 @@
...
+describe("github issues > #11440", () => {
+    let dataSource: DataSource
+    before(async () => {
+        ;[dataSource] = await createTestingConnections({
+            entities: [Post],
+            enabledDrivers: ["postgres"], // since the issue is for postgres only
+        })
+    })
+    beforeEach(() => reloadTestingDatabases([dataSource]))
+    after(() => closeTestingConnections([dataSource]))
```
mguida22, line 21: Since you're using `createTestingConnections`, you'll still need to store each data source here and pass the list into `reloadTestingDatabases` and `closeTestingConnections`. Then you'll need to iterate over each datasource in each test.

```suggestion
    let dataSources: DataSource[]
    before(async () => {
        dataSources = await createTestingConnections({
            entities: [Post],
            enabledDrivers: ["postgres"], // since the issue is for postgres only
        })
    })
    beforeEach(() => reloadTestingDatabases(dataSources))
    after(() => closeTestingConnections(dataSources))
```

Alternatively, you can take a look at using `setupSingleTestingConnection`, some tests use that configuration.

mmarifat: I have updated the codebase, does it look okay now? Thanks

test/github-issues/11440/issue-11440.ts
```diff
@@ -0,0 +1,110 @@
...
+} from "../../utils/test-utils"
+import { Post } from "./entity/Post"
+
+describe("github issues > #11440", () => {
+    let dataSources: DataSource[]
+
+    before(async () => {
+        dataSources = await createTestingConnections({
+            entities: [Post],
+            enabledDrivers: ["postgres", "aurora-postgres"], // since the issue is for postgres
```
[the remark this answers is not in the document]
mmarifat: ok. adding now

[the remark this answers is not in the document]
mmarifat: Done, sorry I should have checked the `isPostgresFamily`

test/github-issues/11440/issue-11440.ts
```diff
@@ -0,0 +1,110 @@
...
+} from "../../utils/test-utils"
+import { Post } from "./entity/Post"
+
+describe("github issues > #11440", () => {
+    let dataSources: DataSource[]
+
+    before(async () => {
+        dataSources = await createTestingConnections({
+            entities: [Post],
+            enabledDrivers: ["postgres", "aurora-postgres", "cockroachdb"], // since the issue is for postgres
```
mguida22, line 18: nit
```suggestion
            enabledDrivers: ["postgres", "aurora-postgres", "cockroachdb"],
```

mguida22 (approved):

gioboa (approved):
Thanks for your help @mmarifat

alumni:
@mmarifat @mguida22 Unfortunately the test does not execute correctly, it succeeds even though the first save throws an error:
>    ✔ should use table or alias name during upsert or doUpdate when both schema name and skipUpdateIfNoValuesChanged supplied
> query: SELECT "Post"."id" AS "Post_id", "Post"."title" AS "Post_title" FROM "typeorm_test"."post" "Post" WHERE "Post"."id" IN ($1, $2, $3) -- PARAMETERS: [1,2,3]
>
> query failed: SELECT "Post"."id" AS "Post_id", "Post"."title" AS "Post_title" FROM "typeorm_test"."post" "Post" WHERE "Post"."id" IN ($1, $2, $3) -- PARAMETERS: [1,2,3]
> error: Error: Connection terminated
>
>  1 passing (443ms)

There's another PR changing the same code, so this fix might be broken by that and we wouldn't know because the test doesn't run.

mmarifat:
> @mmarifat @mguida22 Unfortunately the test does not execute correctly, it succeeds even though the first save throws an error:
> 
> > ✔ should use table or alias name during upsert or doUpdate when both schema name and skipUpdateIfNoValuesChanged supplied
> > query: SELECT "Post"."id" AS "Post_id", "Post"."title" AS "Post_title" FROM "typeorm_test"."post" "Post" WHERE "Post"."id" IN ($1, $2, $3) -- PARAMETERS: [1,2,3]
> > query failed: SELECT "Post"."id" AS "Post_id", "Post"."title" AS "Post_title" FROM "typeorm_test"."post" "Post" WHERE "Post"."id" IN ($1, $2, $3) -- PARAMETERS: [1,2,3]
> > error: Error: Connection terminated
> > 1 passing (443ms)
> 
> There's another PR changing the same code, so this fix might be broken by that and we wouldn't know because the test doesn't run.

Probably I can make two separate tests, should I go for it?

alumni:
Honestly, I'm not sure why do we get "Connection terminated". I've only seen it in this test, and it happens in the very first query that TypeORM executes in the test (a `SELECT` issued by `Repository.save()`). Surprisingly, the queries executed before (in the test setup) are fine. 

It might be due to some unawaited promise when obtaining the DB connection (when creating the QueryRunner). But why only here? IDK yet.

If you can figure out why do we get the error, that would be great.

Replication steps:
```shell
podman compose up postgres
# or: docker compose up postgres

npm exec rimraf ./build && npm run watch

npm run test:fast -- --grep "11440"
```

`ormconfig.json`:

```json
[{
    "skip": false,
    "name": "postgres",
    "type": "postgres",
    "host": "localhost",
    "port": 5432,
    "username": "test",
    "password": "test",
    "database": "test",
    "logging": true
}]
```

PR #11456, opened 2025-05-08
perf: improve save performance during entities update

lotczyk:
### Description of change
Improve save performance during entities update. In my test case 2 300 entities were updated. It used to take 27 seconds, now only 3 seconds, almost 10 times improvement.

### Pull-Request Checklist
- [ ] This pull request links relevant issues as `Fixes #0000` - N/A
- [ ] There are new or updated unit tests validating the change - N/A

### DIFF
--- a/src/persistence/SubjectDatabaseEntityLoader.ts
+++ b/src/persistence/SubjectDatabaseEntityLoader.ts
@@ -3,6 +3,7 @@ import { ObjectLiteral } from "../common/ObjectLiteral"
 import { QueryRunner } from "../query-runner/QueryRunner"
 import { FindManyOptions } from "../find-options/FindManyOptions"
 import { MongoRepository } from "../repository/MongoRepository"
+import { OrmUtils } from "../util/OrmUtils"
 
 /**
  * Loads database entities for all operate subjects which do not have database entity set.
@@ -123,20 +124,17 @@ export class SubjectDatabaseEntityLoader {
                         .getMany()
                 }
 
-                // now when we have entities we need to find subject of each entity
-                // and insert that entity into database entity of the found subjects
+                // Now when we have entities we need to find subject of each entity
+                // and insert that entity into database entity of the found subjects.
+                // A single entity can be applied to many subjects as there might be duplicates.
+                // This will likely result in the same row being updated multiple times during a transaction.
                 entities.forEach((entity) => {
-                    const subjects = this.findByPersistEntityLike(
-                        subjectGroup.target,
-                        entity,
-                    )
-                    subjects.forEach((subject) => {
-                        subject.databaseEntity = entity
-                        if (!subject.identifier)
-                            subject.identifier =
-                                subject.metadata.hasAllPrimaryKeys(entity)
-                                    ? subject.metadata.getEntityIdMap(entity)
-                                    : undefined
+                    const entityId =
+                        allSubjects[0].metadata.getEntityIdMap(entity)
+                    allSubjects.forEach((subject) => {
+                        if (subject.databaseEntity) return
+                        if (OrmUtils.compareIds(subject.identifier, entityId))
+                            subject.databaseEntity = entity
                     })
                 })
 
@@ -153,31 +151,6 @@ export class SubjectDatabaseEntityLoader {
     // ---------------------------------------------------------------------
     // Protected Methods
     // ---------------------------------------------------------------------
-
-    /**
-     * Finds subjects where entity like given subject's entity.
-     * Comparison made by entity id.
-     * Multiple subjects may be returned if duplicates are present in the subject array.
-     * This will likely result in the same row being updated multiple times during a transaction.
-     */
-    protected findByPersistEntityLike(
-        entityTarget: Function | string,
-        entity: ObjectLiteral,
-    ): Subject[] {
-        return this.subjects.filter((subject) => {
-            if (!subject.entity) return false
-
-            if (subject.entity === entity) return true
-
-            return (
-                subject.metadata.target === entityTarget &&
-                subject.metadata.compareEntities(
-                    subject.entityWithFulfilledIds!,
-                    entity,
-                )
-            )
-        })
-    }
 
     /**
      * Groups given Subject objects into groups separated by entity targets.


src/persistence/SubjectDatabaseEntityLoader.ts
```diff
@@ -123,20 +124,17 @@ export class SubjectDatabaseEntityLoader {
...
-                    subjects.forEach((subject) => {
-                        subject.databaseEntity = entity
-                        if (!subject.identifier)
-                            subject.identifier =
-                                subject.metadata.hasAllPrimaryKeys(entity)
-                                    ? subject.metadata.getEntityIdMap(entity)
-                                    : undefined
+                    const entityId =
+                        allSubjects[0].metadata.getEntityIdMap(entity)
+                    allSubjects.forEach((subject) => {
```
lotczyk, line 134: `forEach` offers here better performance than `filter` which creates new array
Another thing, we used to iterate over entire `subjects` array which might be huge, now we operate only on relevant subjects perhaps counterintuitively called `allSubjects`

src/persistence/SubjectDatabaseEntityLoader.ts
```diff
@@ -123,20 +124,17 @@ export class SubjectDatabaseEntityLoader {
...
+                        if (subject.databaseEntity) return
```
lotczyk, line 135: This helps a lot.

src/persistence/SubjectDatabaseEntityLoader.ts
```diff
@@ -123,20 +124,17 @@ export class SubjectDatabaseEntityLoader {
...
+                // A single entity can be applied to many subjects as there might be duplicates.
+                // This will likely result in the same row being updated multiple times during a transaction.
                 entities.forEach((entity) => {
-                    const subjects = this.findByPersistEntityLike(
-                        subjectGroup.target,
-                        entity,
-                    )
-                    subjects.forEach((subject) => {
-                        subject.databaseEntity = entity
-                        if (!subject.identifier)
```
lotczyk, line 135: It seems to be a dead code. Without identifier subject can't be matched to any entity. All subjects in `allSubjects` array has identifiers, we check it while creating the array.

src/persistence/SubjectDatabaseEntityLoader.ts
```diff
@@ -154,31 +152,6 @@ export class SubjectDatabaseEntityLoader {
...
-     * This will likely result in the same row being updated multiple times during a transaction.
-     */
-    protected findByPersistEntityLike(
-        entityTarget: Function | string,
-        entity: ObjectLiteral,
-    ): Subject[] {
-        return this.subjects.filter((subject) => {
-            if (!subject.entity) return false
-
-            if (subject.entity === entity) return true
```
lotczyk, line 170: During my tests this condition was never satisfied.  Even usage of cache does not seem to trigger it.

src/persistence/SubjectDatabaseEntityLoader.ts
```diff
@@ -154,31 +152,6 @@ export class SubjectDatabaseEntityLoader {
...
-     * Comparison made by entity id.
-     * Multiple subjects may be returned if duplicates are present in the subject array.
-     * This will likely result in the same row being updated multiple times during a transaction.
-     */
-    protected findByPersistEntityLike(
-        entityTarget: Function | string,
-        entity: ObjectLiteral,
-    ): Subject[] {
-        return this.subjects.filter((subject) => {
-            if (!subject.entity) return false
```
lotczyk, line 168: As explained in another comment, we always have `identifier` hence this check is not needed.

src/persistence/SubjectDatabaseEntityLoader.ts
```diff
@@ -123,20 +124,17 @@ export class SubjectDatabaseEntityLoader {
...
-                        entity,
-                    )
-                    subjects.forEach((subject) => {
-                        subject.databaseEntity = entity
-                        if (!subject.identifier)
-                            subject.identifier =
-                                subject.metadata.hasAllPrimaryKeys(entity)
-                                    ? subject.metadata.getEntityIdMap(entity)
-                                    : undefined
+                    const entityId =
```
lotczyk, line 132: This helps, ID calculation has a cost.

src/persistence/SubjectDatabaseEntityLoader.ts
```diff
@@ -154,31 +152,6 @@ export class SubjectDatabaseEntityLoader {
...
-        entityTarget: Function | string,
-        entity: ObjectLiteral,
-    ): Subject[] {
-        return this.subjects.filter((subject) => {
-            if (!subject.entity) return false
-
-            if (subject.entity === entity) return true
-
-            return (
-                subject.metadata.target === entityTarget &&
```
lotczyk, line 173: Not needed now because `allSubjects` have a common `target`

src/persistence/SubjectDatabaseEntityLoader.ts
```diff
@@ -154,31 +152,6 @@ export class SubjectDatabaseEntityLoader {
...
-                subject.metadata.compareEntities(
```
lotczyk, line 174: Made more efficient.

lotczyk, line 174: Made more efficient.

alumni (approved):

raiderss:
Fantastic work on this project! This could integrate well with our FiveM resources at Eyes Store Tebex.

gioboa (approved):
Thanks @lotczyk that's an amazing improvement 👏
You rock

src/persistence/SubjectDatabaseEntityLoader.ts
```diff
@@ -123,20 +124,17 @@ export class SubjectDatabaseEntityLoader {
...
-                    )
-                    subjects.forEach((subject) => {
-                        subject.databaseEntity = entity
-                        if (!subject.identifier)
-                            subject.identifier =
-                                subject.metadata.hasAllPrimaryKeys(entity)
-                                    ? subject.metadata.getEntityIdMap(entity)
-                                    : undefined
+                    const entityId =
+                        allSubjects[0].metadata.getEntityIdMap(entity)
```
mguida22, line 133: I feel like I'm missing something about how this works. Isn't this taking the id map from the first subject and using it for comparisons against every other subject? I would expect that to only match against the first subject and silently ignore the rest.

Do we have tests to verify the expected output is identical here?

mguida22 (commented):
Thanks for looking into improving performance!

src/persistence/SubjectDatabaseEntityLoader.ts
alumni: A similar way to get the same `EntityMetadata` is a few lines above.

mguida22: Ah I see - we're not getting the id of the first subject, we're using it's method to get the id for `entity`

mguida22 (approved):
This looks good to me 👍 

How did you check the difference in performance?

gioboa (approved):

jovanadjuric (approved):

naorpeled (approved):

gioboa (approved):

lukaszotczyknexontis:
> This looks good to me 👍
> 
> How did you check the difference in performance?

I found the issue by investigating an instance of event loop block. This is a metric that we monitor in production. For a measurement of fix effectiveness I just used `console.time` called before `save`.

lotczyk:
> This looks good to me 👍
> 
> How did you check the difference in performance?

I found the issue by investigating an instance of event loop block. This is a metric that we monitor in production. For a measurement of fix effectiveness I just used console.time called before save.

PR #11463, opened 2025-05-10
fix: resolve pkg.pr.new issue

naorpeled:
### Pull-Request Checklist

### DIFF
--- a/.github/workflows/preview.yml
+++ b/.github/workflows/preview.yml
@@ -16,6 +16,8 @@ jobs:
 
     steps:
       - uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
 
       - name: Use Node.js
         uses: actions/setup-node@v4
--- a/.github/workflows/preview.yml
+++ b/.github/workflows/preview.yml
@@ -37,4 +37,4 @@ jobs:
           corepack prepare pnpm@latest --activate
 
       - name: Release preview version
-        run: pnpm pkg-pr-new publish './build/package' --template='./sample/playground'
+        run: pnpx pkg-pr-new publish './build/package' --template='./sample/playground'
--- a/.github/workflows/preview.yml
+++ b/.github/workflows/preview.yml
@@ -31,5 +31,10 @@ jobs:
       - name: Build
         run: npm run package
 
+      - name: Prepare pnpm for preview publishing
+        run: |
+          npm install -g corepack@latest
+          corepack prepare pnpm@latest --activate
+
       - name: Release preview version
         run: npm run publish:preview
--- a/.github/workflows/preview.yml
+++ b/.github/workflows/preview.yml
@@ -32,4 +32,4 @@ jobs:
         run: npm run package
 
       - name: Release preview version
-        run: npx pkg-pr-new publish './build/package' --template='./sample/playground'
+        run: npx pkg-pr-new publish './build/package' # Temporarily removed --template to isolate the issue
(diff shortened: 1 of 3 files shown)


.github/workflows/preview.yml
```diff
@@ -29,5 +31,10 @@ jobs:
...
         run: npm run package
 
+      - name: Prepare pnpm for preview publishing
+        run: |
+          npm install -g corepack@latest
+          corepack prepare pnpm@latest --activate
+
       - name: Release preview version
-        run: npx pkg-pr-new publish './build/package' --template='./sample/playground'
+        run: npm run publish:preview
```
mguida22, line 40: Should we leave this command in the workflow file? It looks like it depends on the step above.

mguida22 (approved):
Thank you!

.github/workflows/preview.yml
```diff
@@ -16,6 +16,8 @@ jobs:
 
     steps:
       - uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
```
gioboa, line 20: This is so useful 👏

gioboa (approved):
LGTM 💪

.github/workflows/preview.yml
naorpeled: Not sure, wdyt?

mguida22: if there's no need to move it to package.json, I'd leave it here

.github/workflows/preview.yml
```diff
@@ -29,5 +31,10 @@ jobs:
       - name: Build
         run: npm run package
 
+      - name: Prepare pnpm for preview publishing
+        run: |
+          npm install -g corepack@latest
+          corepack prepare pnpm@latest --activate
```
alumni, line 37: Why pnpm when we use npm?

naorpeled: https://github.com/stackblitz-labs/pkg.pr.new/issues/370

.github/workflows/preview.yml
naorpeled: Done 🙏

naorpeled: I brought it back, for some reason it doesn't work when inlining the command  😢

.github/workflows/preview.yml
```diff
@@ -29,5 +31,10 @@ jobs:
...
+
       - name: Release preview version
-        run: npx pkg-pr-new publish './build/package' --template='./sample/playground'
+        run: pnpx pkg-pr-new publish './build/package' --template='./sample/playground'
```
AmirSa12: @naorpeled in my PR, the fix was using that `--pnpm` option ( not using `pnpx` ) which you removed and caused it to fail again.
it's not necessary to use `pnpx`
a little more context about `--pnpm` option:
_pkg.pr.new uses npm pack --json under the hood, in case you face issues, you can also use the --pnpm or --yarn flag so it starts using pnpm pack or yarn pack. This is not necessary in most cases._

naorpeled: Oh, gotcha, my bad, missed it 🙏 
Thanks for explaining, will create a new PR with a fix

.github/workflows/preview.yml
AmirSa12: > I brought it back, for some reason it doesn't work when inlining the command 😢

see my last comment @naorpeled , that would solve your issue

Issue #11455, opened 2025-05-08
Unable to delete entire table using Repository/EntityManager API


qqruz-mck:
### Issue description

In previous versions, it was possible to delete all rows from a table by passing an empty object `{}` as a filter. This functionality was broken in a new release, introduced in a commit marked as a fix, not as a breaking change.

### Expected Behavior

Passing an empty object should delete all rows in the table.

### Actual Behavior

The new version throws the following error: 
`Empty criteria(s) are not allowed for the delete method.`

### Steps to reproduce
```
await repository.delete({})
```
### My Environment

| Dependency          | Version  |
| ---                 | ---      |
| Operating System    |    macOS 15.4.1      |
| Node.js version     | 22.15.0  |
| Typescript version  | 5.8.3  |
| TypeORM version     | 0.3.23  |

### Additional Context

This behavior was consistent with TypeORM’s API as well as with the API of other libraries across the Node.js ecosystem.

Given that this is one of the most widely used ORMs in the ecosystem this type of changes need to be properly documented as they have impact across many projects.

### Relevant Database Driver(s)

- [ ] aurora-mysql
- [ ] aurora-postgres
- [ ] better-sqlite3
- [ ] cockroachdb
- [ ] cordova
- [ ] expo
- [ ] mongodb
- [ ] mysql
- [ ] nativescript
- [ ] oracle
- [x] postgres
- [ ] react-native
- [ ] sap
- [ ] spanner
- [ ] sqlite
- [ ] sqlite-abstract
- [ ] sqljs
- [ ] sqlserver

### Are you willing to resolve this issue by submitting a Pull Request?

No, I don’t have the time and I’m okay to wait for the community / maintainers to resolve this issue.

alumni:
It was PR #10910 - I would see it as a bug that you were even able to delete like that, it's quite dangerous to allow this.

FYI, in the future, we plan to add more restrictions to what you can pass as `FindOptions`, those will be done more gradually.

BTW, we have `QueryRunner.clearTable(tablePath)` (which calls `TRUNCATE` in most drivers, be careful if you use it, there's a difference between `TRUNCATE` and `DELETE` without filter!). This is called by `EntityManager.clear(entityTarget)` / `Repository.clear()`, so you might want to use that instead.

mguida22:
[...]

PR #11469, opened 2025-05-12
chore: add note about breaking change in 0.3.23

mguida22:
This copies the note about a breaking change in 0.3.23 from the GitHub releases to the CHANGELOG so it's easy to discover.

> Can a newly discovered breaking change also be added to CHANGELOG.md in the future? I only checked that, thinking it should contain the same info as GitHub releases, and then proceeded to waste time re-discovering the breaking change, ending up here
> https://github.com/typeorm/typeorm/issues/11455#issuecomment-2873243454


naorpeled (approved):

gioboa (approved):
Thanks @mguida22 👏

PR #11473, opened 2025-05-13
mssql should use NEWID() instead of NEWSEQUENTIALID() for uuid generation strategy

nikolaivasilev12:
The currently used NEWSEQUENTIALID() actually generates a GUID and not a UUID, but in typeorm it is used for the UUID generation strategy. Instead for a proper UUID generation strategy NEWID() is used in MSSQL:
  [https://learn.microsoft.com/en-us/sql/t-sql/functions/newid-transact-sql?view=sql-server-ver16]
  [https://learn.microsoft.com/en-us/sql/t-sql/functions/newsequentialid-transact-sql?view=sql-server-ver16]

### Description of change
I have changed the SqlServerQueryRunner.ts to use NEWID() in all newly generated migrations. This should be backwards compatible, but the databases will have two different format of IDs - old table will have GUIDs (as they are today), but new tables will actually have proper UUIDs

### Pull-Request Checklist
- [n/a] This pull request links relevant issues as `Fixes #0000`
- [n/a] There are new or updated unit tests validating the change
- [n/a] Documentation has been updated to reflect this change

### DIFF
--- a/src/driver/sqlserver/SqlServerQueryRunner.ts
+++ b/src/driver/sqlserver/SqlServerQueryRunner.ts
@@ -3296,7 +3296,7 @@ export class SqlServerQueryRunner
                             tableColumn.isGenerated = isGenerated
                             if (isGenerated)
                                 tableColumn.generationStrategy = "increment"
-                            if (tableColumn.default === "newid()") {
+                            if (tableColumn.default === "newid()" || tableColumn.default === "newsequentialid()") {
                                 tableColumn.isGenerated = true
                                 tableColumn.generationStrategy = "uuid"
                                 tableColumn.default = undefined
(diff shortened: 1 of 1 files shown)


sgarner:
Thank you for your contribution @nikolaivasilev12 ! 💜 

However I am going to close this PR because the change is not necessary.

"GUID" and "UUID" are synonyms; "globally unique" and "universally unique" are equivalent in meaning. The term "GUID" is just the one preferred by Microsoft.

The difference between `NEWID()` and `NEWSEQUENTIALID()` is that the latter generates sequentially incrementing (but still universally unique) identifiers – as its name implies. For use as table primary keys this provides far superior performance by avoiding index fragmentation. It is therefore strongly recommended not to use `NEWID()`, so TypeORM purposefully does not.

This StackOverflow thread has some good info for further reading: https://stackoverflow.com/a/1587198/2168021

The case when you should use `NEWID()` is if you need a cryptographically strong identifier, such as for an authentication token, for which it's important that values can't be predicted. But in that case it's better not to use such values as your primary key.

If you really want your primary key to use `NEWID()` in TypeORM, you can simply override the column default option:

```ts
@PrimaryColumn({
  type: 'uniqueidentifier',
  generated: 'uuid',
  default: () => 'NEWID()',
})
id: string
```

PR #11482, opened 2025-05-17
Optimistic locking

ricklabanca:
### Description of change

I wanted optimistic locking similar to other orm (I come from using hibernate a lot). The current implementation was setting a pseudo lock on read for this which just reads, checks versions, then writes. But that allows for a fail if another process gets in-between that, albeit small chance.

So this change adds a where version= type check on the update, and reuses the OptimisticLockVersionMismatchError exception if the query fails. If my app sees that exception it knows to reget the entity and resolve things. If there is a version column but no value, I throw OptimisticLockCanNotBeUsedError

The only way I verified it was to change the record through another tool, bumping up the version manually, and then attempting a typeorm save.

I don't have a lot of experience with node test suites or even knowing how to run them for this when it looks to require a lot of db's set up.

You can argue this is a small edge case, and for me it may be also, but when doing batch type updates this kind of thing can happen.

Current behavior would be failing this situation, but the fail would be a silent overwrite of data (writer missing the version change wins). With this technique the write would fail.

### Pull-Request Checklist

-   [x] This pull request links relevant issues as `Fixes #00000` na

Documentation and unit tests, no. I thought I'd submit and if it's ok I could do that. Never did a pull before.

### DIFF
--- a/.github/workflows/publish-package-rick.yml
+++ b/.github/workflows/publish-package-rick.yml
@@ -1,0 +1,35 @@
+name: Publish Package to npmjs rick
+on:
+  push:
+    branches:
+      - 0.3.22-optimistic-locking
+  workflow_dispatch:
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+      id-token: write
+    steps:
+      - uses: actions/checkout@v4
+      # Setup .npmrc file to publish to npm
+      - uses: actions/setup-node@v4
+        with:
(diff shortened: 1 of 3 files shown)


gioboa (commented):
hi @ricklabanca can you solve the conflicts please?

gioboa (commented):
@ricklabanca  thanks for opening this PR.
I'm closing this for now because it's stale and has conflicts, feel free to create a new PR with your improvements. :muscle:

PR #11503, opened 2025-05-30
fix(persistence): prevent save() from setting undefined fields to null on update

### LINKED ISSUE #11419: when use save in transaction set values to null
### Issue description

its a bug

### Expected Behavior

when use save to update data set some values to null but its ok when use update

### Actual Behavior

it have to just update the record not set some values to null

### Steps to reproduce

managee.save(someData)

### My Environment

| Operating System    |    linux and windows      |
| Node.js version     | 22.11.00  |
| Typescript version  | 5.1.3  |
| TypeORM version     | ^0.3.20|

### Additional Context

_No response_

- [ ] aurora-mysql
- [ ] aurora-postgres
- [ ] better-sqlite3
- [ ] cockroachdb
- [ ] cordova
- [ ] expo
- [ ] mongodb
- [x] mysql
- [ ] nativescript
- [ ] oracle
- [ ] postgres
- [ ] react-native
- [ ] sap
- [ ] spanner
- [ ] sqlite
- [ ] sqlite-abstract
- [ ] sqljs
- [ ] sqlserver

codernirdesh:
Fixes #11419. Previously, when using save() to update an entity, any nullable columns omitted (set as undefined) would be overwritten with null in the database. This change ensures that undefined fields are only set to null during insert, not update, so updates with undefined fields leave existing values unchanged. Adds/relies on tests in null-and-default-behaviour and issue-2331 to verify correct behavior. See: https://github.com/typeorm/typeorm/issues/11419

BREAKING CHANGE: Applications relying on save() setting undefined fields to null on update will need to explicitly set fields to null if that is desired.

### Description of change
This pull request refactors and improves the `SubjectExecutor` class in `src/persistence/SubjectExecutor.ts`. The changes focus on enhancing code clarity, ensuring proper handling of nullable columns during inserts, and improving type safety.

### Code clarity improvements:

* Refactored the `nestedSetPromise` logic to use an immediately invoked asynchronous function for better readability and flow.

### Handling of nullable columns:

* Updated logic to set `null` for undefined nullable columns only during inserts, ensuring this behavior is limited to insert operations.

### Type safety enhancements:

* Updated the type of `subject.metadata.target` in the `merge` method to use `EntityTarget<ObjectLiteral>` for improved type safety.

### Additional imports:

* Added the `EntityTarget` import to support type safety improvements.

### Pull-Request Checklist

-   [x] This pull request links relevant issues as `Fixes #11419`
-   [ ] There are new or updated unit tests validating the change   N/A
-   [ ] Documentation has been updated to reflect this change      N/A

### DIFF
--- a/src/persistence/SubjectExecutor.ts
+++ b/src/persistence/SubjectExecutor.ts
@@ -16,6 +16,7 @@ import { OrmUtils } from "../util/OrmUtils"
 import { UpdateResult } from "../query-builder/result/UpdateResult"
 import { ObjectUtils } from "../util/ObjectUtils"
 import { InstanceChecker } from "../util/InstanceChecker"
+import { EntityTarget } from "../common/EntityTarget"
 
 /**
  * Executes all database operations (inserts, updated, deletes) that must be executed
@@ -654,15 +655,18 @@ export class SubjectExecutor {
         }
 
         // Run nested set updates one by one
-        const nestedSetPromise = new Promise<void>(async (ok, fail) => {
-            for (const subject of nestedSetSubjects) {
-                try {
-                    await updateSubject(subject)
-                } catch (error) {
-                    fail(error)
+        const nestedSetPromise = new Promise<void>((ok, fail) => {
+            (async () => {
+                for (const subject of nestedSetSubjects) {
+                    try {
+                        await updateSubject(subject)
+                    } catch (error) {
+                        fail(error)
+                        return
+                    }
                 }
-            }
-            ok()
+                ok()
+            })()
         })
 
         // Run all remaining subjects in parallel
@@ -1065,7 +1069,9 @@ export class SubjectExecutor {
         subjects.forEach((subject) => {
             if (!subject.entity) return
 
-            // set values to "null" for nullable columns that did not have values
+            // Only set undefined nullable columns to null during insert
+            const isInsert = subject.mustBeInserted
+
             subject.metadata.columns.forEach((column) => {
                 // if table inheritance is used make sure this column is not child's column
                 if (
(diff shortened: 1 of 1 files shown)


sgarner:
Heads up @naorpeled - this is similar to #11332 but for save() instead of find(). May need another config option here too.

src/persistence/SubjectExecutor.ts
```diff
@@ -654,15 +655,18 @@ export class SubjectExecutor {
...
+                        await updateSubject(subject)
+                    } catch (error) {
+                        fail(error)
+                        return
+                    }
                 }
-            }
-            ok()
+                ok()
+            })()
```
alumni, line 669: There's no need to make this code more complex than it already is.

First, I'd get rid of the `return` that was added, it is unnecessary if the try-catch would not be inside the for loop.

Second, every async function returns a promise, so no need to use the Promise constructor. ~~Unfortunately, this means that we have to have an IIFE like in this change, but still, it's more readable:~~

_Later edit:_ we can also get rid of the IIFE:

```ts
const updateNestedSubjects = async () => {
    for (const subject of nestedSetSubjects) {
        await updateSubject(subject)
    }
}

// Run all remaining subjects in parallel
await Promise.all([
    ...remainingSubjects.map(updateSubject),
    updateNestedSubjects(),
])
```

src/persistence/SubjectExecutor.ts
```diff
@@ -1117,7 +1123,7 @@ export class SubjectExecutor {
             // merge into entity all generated values returned by a database
             if (subject.generatedMap)
                 this.queryRunner.manager.merge(
-                    subject.metadata.target as any,
+                    subject.metadata.target as EntityTarget<ObjectLiteral>,
```
alumni, line 1126: Just remove the cast, it was not necessary in the first place.

We prefer less code than additional unnecessary code ;)

PR #11507, opened 2025-06-04
fix: handle limit(0) and offset(0) correctly in SelectQueryBuilder

yeonghun104:
### Description of change

**Problem:**
Currently, when `limit(0)` or `offset(0)` is used in SelectQueryBuilder, they are treated as falsy values in JavaScript and completely ignored, causing the query to return all records instead of the expected empty result. This is a potential security issue and goes against SQL standard behavior.

**Root Cause:**
JavaScript falsy value evaluation in conditional statements:
```typescript
if (limit && offset) // limit=0 evaluates to false
if (limit)          // limit=0 evaluates to false
```

**Solution:**
- Replace falsy checks (`!limit`) with explicit undefined/null checks (`limit === undefined`)
- Add helper functions `hasLimit` and `hasOffset` to properly distinguish between unset values and zero values
- Ensure consistent behavior across all database drivers (MySQL, PostgreSQL, SQLite, SQL Server, Oracle)
- Add comprehensive test coverage for zero value scenarios

**Current Behavior:**
```typescript
queryBuilder.limit(0).getMany()
// SQL: SELECT * FROM users (no LIMIT clause)
// Result: ALL records (security risk!)
```

**New Behavior:**
```typescript
queryBuilder.limit(0).getMany()
// SQL: SELECT * FROM users LIMIT 0
// Result: [] (empty array, as expected)
```

**Changes Made:**
1. **SelectQueryBuilder.ts**: 
   - Modified `createLimitOffsetExpression()` to handle zero values correctly
   - Added helper functions to distinguish between undefined and 0
   - Updated all database-specific condition checks

2. **Test Coverage**:
   - Added comprehensive test cases for `limit(0)`, `offset(0)`, `take(0)`, `skip(0)`
   - Verified SQL generation correctness
   - Tested actual query execution behavior

**Verification:**
- ✅ All existing tests pass
- ✅ New test cases added and passing
- ✅ Tested across SQLite (in test suite)
- ✅ Manual verification of SQL generation for all database types

### Pull-Request Checklist

- [ ] This pull request links relevant issues as `Fixes #00000` (N/A - no existing issue)
- [ ] Documentation has been updated to reflect this change (N/A - behavior fix, not new feature)

### Breaking Changes
**None.** This is a bug fix that makes the behavior match developer expectations and SQL standards.

### Database Compatibility
- ✅ MySQL/MariaDB: `LIMIT 0`
- ✅ PostgreSQL: `LIMIT 0`
- ✅ SQLite: `LIMIT 0`
- ✅ SQL Server: `FETCH NEXT 0 ROWS ONLY`
- ✅ Oracle: `FETCH NEXT 0 ROWS ONLY`

#### Before (Broken Behavior)
```typescript
// Expected: Get 0 results
const users = await userRepository
  .createQueryBuilder("user")
  .limit(0)
  .getMany()

// Actual SQL: SELECT * FROM users (no LIMIT!)
// Actual Result: [user1, user2, user3, ...] (ALL users - security risk!)
```

#### After (Fixed Behavior)
```typescript
// Expected: Get 0 results
const users = await userRepository
  .createQueryBuilder("user")
  .limit(0)
  .getMany()

// Generated SQL: SELECT * FROM users LIMIT 0
// Result: [] (empty array, as expected)
```

### Use Cases This Fixes

1. **Dynamic Pagination**: When page size is conditionally set to 0
2. **Conditional Queries**: When limit is calculated and might be 0
3. **Security**: Prevents accidental data exposure when 0 limit was intended
4. **API Consistency**: Makes behavior match standard SQL and other ORMs

This fix ensures TypeORM behaves predictably and securely when developers explicitly set limit or offset to zero.

### DIFF
--- a/src/query-builder/SelectQueryBuilder.ts
+++ b/src/query-builder/SelectQueryBuilder.ts
@@ -2545,13 +2545,17 @@ export class SelectQueryBuilder<Entity extends ObjectLiteral>
         let offset: number | undefined = this.expressionMap.offset,
             limit: number | undefined = this.expressionMap.limit
         if (
-            !offset &&
-            !limit &&
+            offset === undefined &&
+            limit === undefined &&
             this.expressionMap.joinAttributes.length === 0
         ) {
             offset = this.expressionMap.skip
             limit = this.expressionMap.take
         }
+
+        // Helper functions to check if values are set (including 0)
+        const hasLimit = limit !== undefined && limit !== null
+        const hasOffset = offset !== undefined && offset !== null
 
         if (this.connection.driver.options.type === "mssql") {
             // Due to a limitation in SQL Server's parser implementation it does not support using
@@ -2561,13 +2565,13 @@ export class SelectQueryBuilder<Entity extends ObjectLiteral>
             // https://dba.stackexchange.com/a/193799
             let prefix = ""
             if (
-                (limit || offset) &&
+                (hasLimit || hasOffset) &&
                 Object.keys(this.expressionMap.allOrderBys).length <= 0
             ) {
                 prefix = " ORDER BY (SELECT NULL)"
             }
 
-            if (limit && offset)
+            if (hasLimit && hasOffset)
                 return (
                     prefix +
                     " OFFSET " +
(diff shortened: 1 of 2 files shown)


sgarner (changes requested):
Thanks for your contribution @yeonghun104 💜 

This looks good. Can you fix the formatting error detected in CI please?

Also I agree with the comment from CodeRabbit that we are lacking tests for non-zero limit/offset. Let's add the suggested tests.

sgarner:
Beware that some database engines will throw an error when given `limit(0)`.

In SQL Server, `SELECT ... OFFSET 0 ROWS FETCH NEXT 0 ROWS ONLY` is not valid:

```
[S0001][10744] Line 1: The number of rows provided for a FETCH clause must be greater then zero.
```

However this outcome seems safer than ignoring the limit entirely, so I don't think it's a reason not to do this.

yeonghun104:
> Thanks for your contribution @yeonghun104 💜
> 
> This looks good. Can you fix the formatting error detected in CI please?
> 
> Also I agree with the nitpick comment from CodeRabbit that we are lacking tests for non-zero limit/offset. Let's add the suggested tests.

I completely agree. While some databases like SQL Server may throw errors with LIMIT 0, 
this explicit failure is better than silently ignoring the limit. When developers 
call .limit(0), they expect zero results, not all results. An explicit error is safer 
and more predictable than silent failure.

jovanadjuric (approved):

gioboa (approved):

alumni:
> Beware that some database engines will throw an error when given limit(0).

I mean, it usually does not make sense to select 0 rows (not in what you can do with TypeORM, it might in procedures, but that's something else).

We also had this issue in our project and we realized that if limit = 0, we don't need to issue the query at all and if offset = 0 it doesn't need to be passed since it's the default value.

I don't remember if we opened an issue, but we dismissed the idea after some time when we realized it's unnecessary :)

Basically, for offset it's irrelevant, but we can maybe accept the PR to fix the situation when you accidentally call limit(0) and get all results.

alumni (approved):

effective-giggle:
I think this should be marked as "Breaking Change" or at least "Potentially breaking Change" in case somebody used to give a limit of 0 to fetch all values. 
(This change also implicitly affects the Repository API)

sgarner:
Generally speaking a change is not breaking if it changes some behavior that was not documented or which contradicted the documented behavior. I think that applies in this case; if a user was relying on `limit(0)` fetching all values, they were relying on an undocumented quirk of the previous implementation.

I guess the most likely impact of this is where the limit value comes from user input. It's a good reminder to always and strictly validate all user input.

PR #11519, opened 2025-06-10
chore(ci): avoid duplicate CI runs for pushes to non-fork PRs

sgarner:
### Description of change

Previously, pushes to branches in this repository with pull requests would trigger duplicate CI runs:

image
image
image

This is because our workflow uses both the `push` and `pull_request` events, and GitHub Actions doesn't do any de-duplication. So a push to a `typeorm/typeorm` branch with a PR would trigger both events (producing 86 test jobs instead of 43).

This didn't happen very often because most PRs are from outside contributors on branches in their own forks, which would only trigger the `pull_request` event. However it would occur when maintainers with write access created PRs using non-fork branches in `typeorm/typeorm`.

We can fix this by changing the trigger branch criteria:

- Filter the `push` event so it only triggers for pushes to the `master` branch. We have (I believe?) branch protection rules to ensure that commits can't be made directly to `master`, but this ensures we re-validate `master` after a PR is merged. Since all changes require a PR, CI doesn't need to run on other branches unless a PR is opened.

- I've also widened the filter on the `pull_request` event so it triggers for PRs based on all branches. I don't see any need to restrict this, and it ensures that tests can run in chained PRs.

## Summary by CodeRabbit

- **Chores**
  - Updated workflow triggers so tests now run on pushes to the master branch and on pull requests targeting any branch.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -6,11 +6,11 @@ name: test
 
 on:
   push:
-    branches: ["**"]
+    branches: ["master"]
     paths-ignore:
       - "docs/**"
   pull_request:
-    branches: ["master"]
+    branches: ["**"]
     paths-ignore:
       - "docs/**"
 


alumni:
@sgarner probably contributors should be using forks instead, so the workflows run in their own forks :) 

I think it's okay that the test workflows run on every branch - then the contributors (whether they have write access to this repo or not) will see the test results in their own forked repository and they will not need to wait for workflow approval from one of us.

I am tempted to decline this PR.

sgarner:
> @sgarner probably contributors should be using forks instead, so the workflows run in their own forks :)
> 
> I think it's okay that the test workflows run on every branch - then the contributors (whether they have write access to this repo or not) will see the test results in their own forked repository and they will not need to wait for workflow approval from one of us.
> 
> I am tempted to decline this PR.

That's a fair point - although I doubt that many contributors work that way. Workflows are disabled on forks by default, so they would have to specifically go in and enable the option to do this. I've observed that most people only run tests on their local machine (if at all) before submitting a PR.

If we want to support that, additional conditions could be used to adapt the workflow depending on whether the branch is on a fork or not.

Other than working around this issue, what would be the benefit of forcing maintainers to work on forks?

OSA413:
> Workflows are disabled on forks by default, so they would have to specifically go in and enable the option to do this. 

Hm, it didn't request me to enable anything when I forked the repo, though it was very long ago.

Maybe we should update docs to tell people how to enable CI/CD on their fork?

sgarner:
> Hm, it didn't request me to enable anything when I forked the repo, though it was very long ago.

Possible it was different if you forked before typeorm was using GitHub Actions. Here's what appears now:

image

Of course if you do this with the current trigger settings it causes the same duplicate CI runs issue to occur in your fork. Every push to a pull request will then trigger jobs in both your fork and the upstream (though this will be less visible since the jobs in the fork won't appear as checks on the PR).

alumni:
I think I had to enable it, but it was one time only so I vaguely remember.

Either way, I like that I can run the workflows in my own fork before I open the PR.

alumni:
@sgarner I think I might know how to solve this, by checking the repository owner:

```yaml
jobs:
  database-tests:
    if: ${{ (github.repository_owner != 'typeorm') || (github.ref_name == 'master' && <something to check it's a not PR> ) }}
```

`github.repository != 'typeorm/typeorm'` is also allowed, not entirely sure how to check the workflow trigger.

alumni:
Actually I'm wondering, why don't we keep the `on.push` workflow as it is and remove `on.pull_request`?

Later edit: nevermind, I understand why :)

PR #11520, opened 2025-06-10
feat(sap): use the native driver for connection pooling

### LINKED ISSUE #9890: SAP HANA: `hdb-pool` is deprecated - should use standard connection pooling
### Feature Description

The package `hdb-pool` for the SAP HANA driver has been archived & deprecated: https://github.com/ckyycc/hdb-pool/commit/256c645f45d77cf33de7dd8175e4319c9b793e9b

This is because the standard `@sap/hana-client` supports connection pooling out-of-the-box by now: https://help.sap.com/docs/SAP_HANA_CLIENT/f1b440ded6144a54ada97ff95dac7adf/e252ff9b2cb44dd9925901e39025ce77.html

It would make sense to replace the usage of `hdb-pool` with the default pooling.
Looking at the documentation, TypeORM could probably use at least the 'implicit' without too much effort?

### The Solution

Use standard connection pooling of `@sap/hana-client`.

### Considered Alternatives

n/a

### Additional Context

I didn't look into the code yet, but may be able to do so later. I'm opening this issue already for tracking purposes. 

- [ ] aurora-mysql
- [ ] aurora-postgres
- [ ] better-sqlite3
- [ ] cockroachdb
- [ ] cordova
- [ ] expo
- [ ] mongodb
- [ ] mysql
- [ ] nativescript
- [ ] oracle
- [ ] postgres
- [ ] react-native
- [X] sap
- [ ] spanner
- [ ] sqlite
- [ ] sqlite-abstract
- [ ] sqljs
- [ ] sqlserver

alumni:
### Description of change

Implements connection pooling for the SAP HANA driver using the built-in explicit pooling functions of `@sap/hana-client`.

This helps remove the deprecated `hdb-pool` library, which hasn't been updated in 5 years. Fixes #9890.

### Pull-Request Checklist

-   [X] This pull request links relevant issues as `Fixes #00000`
-   [X] There are new or updated unit tests validating the change
-   [X] Documentation has been updated to reflect this change

## Summary by CodeRabbit

- **New Features**
  - Added support for new SAP HANA connection pool configuration options, including maximum pool size, idle time, ping checks, and pool capacity.
  - Introduced new documentation for SAP HANA data source options and connection pooling.

- **Bug Fixes**
  - Improved error handling and connection management for SAP HANA driver connections.

- **Refactor**
  - Modernized SAP HANA driver to use async/await and streamlined connection and pool option handling.
  - Deprecated and removed legacy pool options in favor of new configuration.
  - Enhanced query execution and streaming with better async handling and error management.
  - Simplified dependency loading and prioritized provided drivers over dynamic loading.

- **Documentation**
  - Updated installation instructions for SAP HANA, removing references to deprecated packages and sponsorship notes.
  - Clarified and expanded SAP HANA configuration documentation.

- **Tests**
  - Added new functional tests to verify SAP HANA connection pool behavior.

- **Chores**
  - Removed deprecated SAP HANA dependencies from the package configuration.

### DIFF
--- a/src/driver/sap/SapDriver.ts
+++ b/src/driver/sap/SapDriver.ts
@@ -58,6 +58,11 @@ export class SapDriver implements Driver {
      */
     master: any
 
+    /**
+     * Function handling errors thrown by drivers pool.
+     */
+    poolErrorHandler: (error: any) => void
+
     // -------------------------------------------------------------------------
     // Public Implemented Properties
     // -------------------------------------------------------------------------
@@ -220,7 +225,7 @@ export class SapDriver implements Driver {
 
     /**
      * Max length allowed by SAP HANA for aliases (identifiers).
-     * @see https://help.sap.com/viewer/4fe29514fd584807ac9f2a04f6754767/2.0.03/en-US/20a760537519101497e3cfe07b348f3c.html
+     * @see https://help.sap.com/docs/hana-cloud-database/sap-hana-cloud-sap-hana-database-sql-reference-guide/system-limitations
      */
     maxAliasLength = 128
 
@@ -294,8 +299,22 @@ export class SapDriver implements Driver {
             poolOptions.poolCapacity = this.options.pool.poolCapacity
         }
 
+        this.poolErrorHandler =
+            this.options.pool?.poolErrorHandler ??
+            ((error: Error) => {
+                this.connection.logger.log(
+                    "warn",
+                    `SAP HANA pool raised an error: ${error}`,
+                )
+            })
+
         // create the pool
-        this.master = this.client.createPool(connectionOptions, poolOptions)
+        try {
+            this.master = this.client.createPool(connectionOptions, poolOptions)
+        } catch (error) {
+            this.poolErrorHandler(error)
+            throw error
+        }
 
         const queryRunner = this.createQueryRunner("master")
 
@@ -327,7 +346,12 @@ export class SapDriver implements Driver {
         }
 
         this.master = undefined
-        await promisify(pool.clear).call(pool)
+        try {
+            await promisify(pool.clear).call(pool)
+        } catch (error) {
+            this.poolErrorHandler(error)
+            throw error
+        }
     }
 
     /**
@@ -341,7 +365,12 @@ export class SapDriver implements Driver {
             throw new TypeORMError("Driver not Connected")
         }
 
-        return await promisify(pool.getConnection).call(pool)
+        try {
+            return await promisify(pool.getConnection).call(pool)
+        } catch (error) {
+            this.poolErrorHandler(error)
+            throw error
+        }
     }
 
     /**
--- a/src/driver/sap/SapDriver.ts
+++ b/src/driver/sap/SapDriver.ts
@@ -1,3 +1,4 @@
+import { promisify } from "node:util"
 import {
     ColumnType,
     ConnectionIsNotSetError,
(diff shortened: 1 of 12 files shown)


alumni:
@coderabbitai full review

lotczyk (approved):

gioboa (approved):
:clap: @alumni

src/driver/sap/SapDriver.ts
```diff
@@ -259,54 +259,62 @@ export class SapDriver implements Driver {
      */
     async connect(): Promise<void> {
         // HANA connection info
-        const dbParams = {
-            hostName: this.options.host,
+        const connectionOptions: any = {
```
sgarner, line 262: Does the client library not provide typings that we could use for these options objects? 😞

alumni: It does, but in TypeORM we can't really use types from the database clients, because we would need to have these clients as direct dependencies.

I'm hoping we would do two things in the future:
1. short term: import types using type imports. We can easily configure eslint to convert imports to type imports when possible, which would allow us to import types from devDependencies in the TypeORM source code (although only internally).
2. long term: extract drivers into separate packages which import the clients / use them as peer dependencies. For the existing drivers it would allow better typing, also externally, for TypeORM users.

BTW, while I did this change I imported the types from HANA Client to make things easier for myself and removed them after, so that users that don't rely on the SAP Driver don't have to import the client library.

alumni:
> I saw some config deprecations but I assume the pooling functionality is equivalent. So this is safe to include in a minor release, right?

I tried to keep things as compatible as possible with the previous pooling library. There are some settings that will have no effect anymore, since they don't have an equivalent, but it won't be a breaking change, e.g.: the timeout setting is not used, so now trying to acquire a connection from the pool when the pool is full will fail, but the chance of failure now is much lower than the chance of timeout with the old pool (with the same pool size) - the old pool was buggy.

Seems the preference within SAP is to release this ASAP as the security compliance teams are pressuring every development team to drop no longer maintained libraries like the old pool.

src/driver/sap/SapDriver.ts
sgarner: Hmm, but type elision will remove from the compiled output any imports that are only used as types. So importing types from devDependencies shouldn't cause any issues for users; and anyone using the `SapDriver` will need the peerDependency installed anyway. What am I missing?

Using `import type` to make this explicit would be even better, of course.

alumni: Since we use TSC (not e.g. Webpack), there's no tree-shaking, so imports will not be removed. If we have type imports we might be able to get away in _some_ situations (maybe in `SapDriver`, but not in `SapConnectionOptions`).

_Later edit_: `verbatimModuleSyntax` is not enabled in our TSConfig (I though it's the default behavior). I would still keep things as they are and add types later.

sgarner: > Since we use TSC (not e.g. Webpack), there's no tree-shaking, so imports will not be removed. If we have type imports we might be able to get away in _some_ situations (maybe in `SapDriver`, but not in `SapConnectionOptions`).

Tree shaking isn't necessary, this is [a TSC feature](https://www.typescriptlang.org/tsconfig/#verbatimModuleSyntax):

> image

But yes, a distinction has to be made that we can't reference such imports in _exported_ types like `SapConnectionOptions`, which do appear in our output `.d.ts` files. Dev dependencies can only be imported for use in _internal_ typings.

gioboa (approved):
👏

PR #11524, opened 2025-06-12
perf: avoid unnecessary count on getManyAndCount

EQuincerot:
### Description of change

When `getManyAndCount` is called, stop execute the `count` query if it can be deduced from the
number of returned rows. This will **increase performance** by avoiding one round trip. It would be especially useful when using pagination libraries such as https://github.com/ppetzold/nestjs-paginate

#### Example

```js
const [entities, count] = await connection.manager
    .limit(10)
    .orderBy("post.id")
    .getManyAndCount()
```

Before this change, it triggers two queries:

```sql
SELECT * FROM post ORDER BY post.id LIMIT 10;
SELECT count(*) FROM post;
```
But if the first query returns only 5 results, we know that the count will also be 5.

After this change, it triggers only the first query:

```sql
SELECT * FROM post ORDER BY post.id LIMIT 10;
```

### Pull-Request Checklist

### DIFF
--- a/src/query-builder/SelectQueryBuilder.ts
+++ b/src/query-builder/SelectQueryBuilder.ts
@@ -1875,11 +1875,28 @@ export class SelectQueryBuilder<Entity extends ObjectLiteral>
                 queryRunner,
             )
             this.expressionMap.queryEntity = false
-            const cacheId = this.expressionMap.cacheId
-            // Creates a new cacheId for the count query, or it will retreive the above query results
-            // and count will return 0.
-            this.expressionMap.cacheId = cacheId ? `${cacheId}-count` : cacheId
-            const count = await this.executeCountQuery(queryRunner)
+
+            let count
+            if (
+                ((this.expressionMap.take !== undefined &&
+                    entitiesAndRaw.entities.length < this.expressionMap.take) ||
+                    (this.expressionMap.limit !== undefined &&
+                        entitiesAndRaw.entities.length <
+                            this.expressionMap.limit)) &&
+                this.expressionMap.skip === undefined &&
+                this.expressionMap.offset === undefined
+            ) {
+                // the limit was not reached, we already know the count
+                count = entitiesAndRaw.entities.length
+            } else {
+                const cacheId = this.expressionMap.cacheId
+                // Creates a new cacheId for the count query, or it will retreive the above query results
+                // and count will return 0.
+                this.expressionMap.cacheId = cacheId
+                    ? `${cacheId}-count`
+                    : cacheId
+                count = await this.executeCountQuery(queryRunner)
+            }
             const results: [Entity[], number] = [entitiesAndRaw.entities, count]
 
             // close transaction if we started it
--- a/test/other-issues/lazy-count/lazy-count.ts
+++ b/test/other-issues/lazy-count/lazy-count.ts
@@ -1,0 +1,142 @@
+import "reflect-metadata"
+import {
+    closeTestingConnections,
+    createTestingConnections,
+    reloadTestingDatabases,
+} from "../../utils/test-utils"
+import { DataSource } from "../../../src/data-source/DataSource"
+import { Post } from "./entity/Post"
+import { expect } from "chai"
+import { AfterQuerySubscriber } from "./subscribers/AfterQuerySubscriber"
+
+describe("other issues > lazy count", () => {
+    let connections: DataSource[]
+    before(
+        async () =>
+            (connections = await createTestingConnections({
+                entities: [__dirname + "/entity/*{.js,.ts}"],
+                subscribers: [AfterQuerySubscriber],
+                dropSchema: true,
+                schemaCreate: true,
+            })),
+    )
+    beforeEach(() => reloadTestingDatabases(connections))
+    after(() => closeTestingConnections(connections))
+
+    it("skip count query when less entities are returned than the limit", () =>
+        Promise.all(
+            connections.map(async function (connection) {
+                for (let i = 1; i <= 5; i++) {
+                    const post = new Post()
+                    post.content = "Hello Post #" + i
+
+                    await connection.manager.save(post)
+                }
+
+                const afterQuery = connection
+                    .subscribers[0] as AfterQuerySubscriber
+                afterQuery.clear()
+
+                const [entities, count] = await connection.manager
+                    .createQueryBuilder(Post, "post")
+                    .limit(10)
+                    .orderBy("post.id")
+                    .getManyAndCount()
+
+                count.should.be.equal(5)
+                expect(entities).not.to.be.undefined
+                entities.length.should.be.equal(5)
+
+                expect(afterQuery.calls()).to.be.equal(1)
+                expect(afterQuery.lastCalledQuery()).to.not.match(/count/i)
+            }),
+        ))
+
+    it("skip count query when less entities are returned than the take", () =>
+        Promise.all(
+            connections.map(async function (connection) {
+                for (let i = 1; i <= 5; i++) {
+                    const post = new Post()
+                    post.content = "Hello Post #" + i
+
+                    await connection.manager.save(post)
+                }
+
+                const afterQuery = connection
+                    .subscribers[0] as AfterQuerySubscriber
+                afterQuery.clear()
+
+                const [entities, count] = await connection.manager
+                    .createQueryBuilder(Post, "post")
+                    .take(10)
+                    .orderBy("post.id")
+                    .getManyAndCount()
+
+                count.should.be.equal(5)
+                expect(entities).not.to.be.undefined
+                entities.length.should.be.equal(5)
+
+                expect(afterQuery.calls()).to.be.equal(1)
+                expect(afterQuery.lastCalledQuery()).to.not.match(/count/i)
+            }),
+        ))
+
+    it("run count query when returned entities reach the take", () =>
+        Promise.all(
+            connections.map(async function (connection) {
+                for (let i = 1; i <= 2; i++) {
+                    const post = new Post()
+                    post.content = "Hello Post #" + i
+
+                    await connection.manager.save(post)
+                }
+
+                const afterQuery = connection
+                    .subscribers[0] as AfterQuerySubscriber
+                afterQuery.clear()
+
+                const [entities, count] = await connection.manager
+                    .createQueryBuilder(Post, "post")
+                    .take(2)
+                    .orderBy("post.id")
+                    .getManyAndCount()
+
+                count.should.be.equal(2)
+                expect(entities).not.to.be.undefined
+                entities.length.should.be.equal(2)
+
+                expect(afterQuery.calls()).to.be.equal(2)
+                expect(afterQuery.lastCalledQuery()).to.match(/count/i)
+            }),
+        ))
+
+    it("run count query when an offset is defined", () =>
+        Promise.all(
+            connections.map(async function (connection) {
+                for (let i = 1; i <= 5; i++) {
+                    const post = new Post()
+                    post.content = "Hello Post #" + i
+
+                    await connection.manager.save(post)
+                }
+
+                const afterQuery = connection
+                    .subscribers[0] as AfterQuerySubscriber
+                afterQuery.clear()
+
+                const [entities, count] = await connection.manager
+                    .createQueryBuilder(Post, "post")
+                    .take(10)
+                    .offset(3)
+                    .orderBy("post.id")
+                    .getManyAndCount()
+
+                count.should.be.equal(5)
+                expect(entities).not.to.be.undefined
+                entities.length.should.be.equal(2)
+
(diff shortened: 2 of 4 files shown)


test/other-issues/lazy-count/lazy-count.ts
```diff
@@ -0,0 +1,135 @@
...
+                afterQuery.clear()
+
+                const [entities, count] = await connection.manager
+                    .createQueryBuilder(Post, "post")
+                    .limit(10)
+                    .orderBy("post.id")
+                    .getManyAndCount()
+
+                expect(count).to.be.equal(5)
+                expect(entities).not.to.be.undefined
```
alumni: `expect(entities).to.have.length(5)` should be enough to replace all checks on `entities`.

src/query-builder/SelectQueryBuilder.ts
```diff
@@ -1875,11 +1875,28 @@ export class SelectQueryBuilder<Entity extends ObjectLiteral>
...
+
+            let count
+            if (
+                ((this.expressionMap.take !== undefined &&
+                    entitiesAndRaw.entities.length < this.expressionMap.take) ||
+                    (this.expressionMap.limit !== undefined &&
+                        entitiesAndRaw.entities.length <
+                            this.expressionMap.limit)) &&
+                this.expressionMap.skip === undefined &&
+                this.expressionMap.offset === undefined
```
alumni: There are some edge cases for `take`/`limit`/`skip`/`offset`: `0`, `null`, `undefined`.

test/other-issues/lazy-count/lazy-count.ts
alumni, line 1: Would be great to have a test for one-to-many / many-to-many relations (e.g. when a `Post` has many `Comment`s):

```ts
const postRepository = dataSource.getRepository(Post);

const [posts, postCount]  = await postRepository.findAndCount({
  relations: { comments: true },
  where: { content: Like('%test%') },
});
```

test/other-issues/lazy-count/lazy-count.ts
```diff
@@ -0,0 +1,135 @@
...
+                schemaCreate: true,
+            })),
+    )
+    beforeEach(() => reloadTestingDatabases(connections))
+    after(() => closeTestingConnections(connections))
+
+    it("skip count query when less entities are returned than the limit", () =>
+        Promise.all(
+            connections.map(async function (connection) {
+                await savePostEntities(connection, 5)
```
alumni, line 30: `savePostEntities` does not exist?

EQuincerot: Not sure of what you mean here. It's a method that is defined at the end of the test. I created to mutualize the code creating entities, but if you prefer I can remove it.

alumni: Ah it's at the bottom, I didn't expect to see it there :)

alumni (approved):

gioboa (approved):

gioboa:
Thanks for your amazing help @EQuincerot 
Kudos for your first PR in this project 🎉

src/query-builder/SelectQueryBuilder.ts
```diff
@@ -1903,6 +1912,56 @@ export class SelectQueryBuilder<Entity extends ObjectLiteral>
         }
     }
 
+    private lazyCount(entitiesAndRaw: { entities: Entity[]; raw: any[] }) {
```
sgarner: Method should declare its return type:

```suggestion
    private lazyCount(entitiesAndRaw: { entities: Entity[]; raw: any[] }): number | undefined {
```

src/query-builder/SelectQueryBuilder.ts
```diff
@@ -1875,11 +1875,20 @@ export class SelectQueryBuilder<Entity extends ObjectLiteral>
...
-            // Creates a new cacheId for the count query, or it will retreive the above query results
-            // and count will return 0.
-            this.expressionMap.cacheId = cacheId ? `${cacheId}-count` : cacheId
-            const count = await this.executeCountQuery(queryRunner)
+
+            let count
+            let lazyCount = this.lazyCount(entitiesAndRaw)
+            if (lazyCount !== undefined) {
+                count = lazyCount
+            } else {
```
sgarner: These variable declarations should be typed. And `lazyCount` should be a const since it's never updated.

In fact, we can simplify and lose the `lazyCount` variable entirely:

```suggestion
            let count: number | undefined = this.lazyCount(entitiesAndRaw)
            if (count === undefined) {
```

src/query-builder/SelectQueryBuilder.ts
```diff
@@ -1903,6 +1912,56 @@ export class SelectQueryBuilder<Entity extends ObjectLiteral>
...
+            this.expressionMap.take !== undefined &&
+            this.expressionMap.take !== null
+        if (hasTake) {
+            maxResults = this.expressionMap.take
+        }
+
+        if (hasLimit) {
+            // limit overrides take when no join is defined
+            maxResults = this.expressionMap.limit
+        }
```
sgarner: Simpler with a ternary:

```suggestion
        const hasTake =
            this.expressionMap.take !== undefined &&
            this.expressionMap.take !== null
        // limit overrides take when no join is defined
        const maxResults = hasLimit
            ? this.expressionMap.limit
            : hasTake
            ? this.expressionMap.take
            : undefined
```

src/query-builder/SelectQueryBuilder.ts
```diff
@@ -1903,6 +1912,56 @@ export class SelectQueryBuilder<Entity extends ObjectLiteral>
...
+
+        if (
+            maxResults !== undefined &&
+            entitiesAndRaw.entities.length == maxResults
```
sgarner:
```suggestion
            entitiesAndRaw.entities.length === maxResults
```

src/query-builder/SelectQueryBuilder.ts
```diff
@@ -1903,6 +1912,56 @@ export class SelectQueryBuilder<Entity extends ObjectLiteral>
...
+        }
+
+        if (
+            this.expressionMap.offset !== undefined &&
+            this.expressionMap.offset !== null &&
+            this.expressionMap.offset > 0
+        ) {
+            // offset overrides skip when no join is defined
+            previousResults = this.expressionMap.offset
+        }
```
sgarner:
```suggestion
        const hasSkip = this.expressionMap.skip !== undefined &&
            this.expressionMap.skip !== null &&
            this.expressionMap.skip > 0
        const hasOffset = this.expressionMap.offset !== undefined &&
            this.expressionMap.offset !== null &&
            this.expressionMap.offset > 0
        // offset overrides skip when no join is defined
        const previousResults = hasOffset
            ? this.expressionMap.offset
            : hasSkip
            ? this.expressionMap.skip
            : 0;
```

src/query-builder/SelectQueryBuilder.ts
```diff
@@ -1903,6 +1912,56 @@ export class SelectQueryBuilder<Entity extends ObjectLiteral>
...
+        if (hasLimit) {
+            // limit overrides take when no join is defined
+            maxResults = this.expressionMap.limit
+        }
+
+        if (
+            maxResults !== undefined &&
+            entitiesAndRaw.entities.length == maxResults
+        ) {
+            return undefined
```
sgarner, line 1939:
```suggestion
            // stop here when the result set contains the max number of rows; we need to execute a full count
            return undefined
```

test/other-issues/lazy-count/lazy-count.ts
```diff
@@ -0,0 +1,281 @@
...
+                entities: [__dirname + "/entity/*{.js,.ts}"],
+                subscribers: [AfterQuerySubscriber],
+                dropSchema: true,
+                schemaCreate: true,
+            })),
+    )
+    beforeEach(() => reloadTestingDatabases(connections))
+    after(() => closeTestingConnections(connections))
+
+    it("skip count query when less entities are returned than the limit", () =>
```
sgarner:
```suggestion
    it("skip count query when fewer entities are returned than the limit", () =>
```

test/other-issues/lazy-count/lazy-count.ts
```diff
@@ -0,0 +1,281 @@
...
+
+                expect(
+                    afterQuery
+                        .getCalledQueries()
+                        .filter((query) => query.match(/(count|cnt)/i)),
+                ).to.be.empty
+            }),
+        ))
+
+    it("skip count query when less entities are returned than the take", () =>
```
sgarner:
```suggestion
    it("skip count query when fewer entities are returned than the take", () =>
```

test/other-issues/lazy-count/lazy-count.ts
```diff
@@ -0,0 +1,281 @@
...
+
+                expect(
+                    afterQuery
+                        .getCalledQueries()
+                        .filter((query) => query.match(/(count|cnt)/i)),
+                ).not.to.be.empty
+            }),
+        ))
+
+    it("skip count query when joining a subentity with a take", () =>
```
sgarner:
```suggestion
    it("skip count query when joining a relation with a take", () =>
```

test/other-issues/lazy-count/lazy-count.ts
```diff
@@ -0,0 +1,281 @@
...
+
+                expect(
+                    afterQuery
+                        .getCalledQueries()
+                        .filter((query) => query.match(/(count|cnt)/i)),
+                ).to.be.empty
+            }),
+        ))
+
+    it("run count query when joining a subentity with a limit", () =>
```
sgarner:
```suggestion
    it("run count query when joining a relation with a limit", () =>
```

test/other-issues/lazy-count/lazy-count.ts
```diff
@@ -0,0 +1,281 @@
...
+
+                expect(
+                    afterQuery
+                        .getCalledQueries()
+                        .filter((query) => query.match(/(count|cnt)/i)),
+                ).not.to.be.empty
+            }),
+        ))
+
+    it("skip count query when joining a subentity with a take and a skip", () =>
```
sgarner:
```suggestion
    it("skip count query when joining a relation with a take and a skip", () =>
```

test/other-issues/lazy-count/lazy-count.ts
```diff
@@ -0,0 +1,281 @@
...
+
+                expect(
+                    afterQuery
+                        .getCalledQueries()
+                        .filter((query) => query.match(/(count|cnt)/i)),
+                ).to.be.empty
+            }),
+        ))
+
+    it("run count query when joining a subentity with a limit and an offset", () =>
```
sgarner:
```suggestion
    it("run count query when joining a relation with a limit and an offset", () =>
```

sgarner (changes requested):
Nice fix, thank you @EQuincerot

Just a few code improvements that could be made.

src/query-builder/SelectQueryBuilder.ts
EQuincerot: I added type assertion to make typescript happy with the undefined that could actually not happen. I'm not 100% happy with that, if you have a better idea, feel free to share!

src/query-builder/SelectQueryBuilder.ts
```diff
@@ -1903,6 +1909,55 @@ export class SelectQueryBuilder<Entity extends ObjectLiteral>
...
+            this.expressionMap.offset !== undefined &&
+            this.expressionMap.offset !== null &&
+            this.expressionMap.offset > 0
+
+        // offset overrides skip when no join is defined
+        const previousResults: number = hasOffset
+            ? <number>this.expressionMap.offset
+            : hasSkip
+            ? <number>this.expressionMap.skip
+            : 0
```
sgarner: Angle-bracket type assertions in TypeScript are generally discouraged these days.

I think you can achieve the effect better with a `!` non-null assertion:

```suggestion
        const previousResults: number = hasOffset
            ? this.expressionMap.offset!
            : hasSkip
            ? this.expressionMap.skip!
            : 0
```

sgarner (approved):
One more tweak, otherwise looks good :+1:

src/query-builder/SelectQueryBuilder.ts
alumni: These rules you are referring to are `consistent-type-assertions` and `non-nullable-type-assertion-style` from "stylistic" and "stylistic type checked" rulesets.

I was planning to progressively add some more eslint rulesets ("stylistic" and "strict"), just didn't want to enable all of them at once, thought about waiting a couple of releases between adding new rulesets, but it's good to follow them already if you are aware of them :)

sgarner (approved):
Thanks for your contribution @EQuincerot 💜

gioboa (approved):
Great improvement @EQuincerot thanks

PR #11529, opened 2025-06-15
feat: safe column alterations

elskow:
### Description of change

This pull request fixes a critical data loss issue in TypeORM's migration generation where column length changes result in unnecessary DROP/ADD operations instead of using database-native ALTER COLUMN commands.

#### What the change is intended to do:
- Replace DROP/ADD column operations with ALTER COLUMN for safe column modifications
- Implement intelligent change detection that distinguishes between safe and potentially destructive operations
- Provide database-specific implementations for optimal ALTER COLUMN syntax

#### Current vs New Behavior:

**Before (Problematic):**
```sql
-- Changing varchar(50) to varchar(51)
-- Result: ALL DATA LOST
```

**After (Fixed):**
```sql
-- Changing varchar(50) to varchar(51)
-- Result: Data preserved, operation is safe and fast
```

#### Implementation Details:
- **PostgreSQL**: Uses `ALTER COLUMN TYPE` syntax for safe changes
- **MySQL**: Uses `ALTER TABLE MODIFY COLUMN` syntax  
- **SQL Server**: Uses `ALTER COLUMN` syntax
- **Safe Operations**: Length increases, compatible type changes
- **Unsafe Operations**: Still use DROP/ADD with clear warnings (length decreases, incompatible types)

#### Affected Database Drivers:
- PostgresQueryRunner
- MysqlQueryRunner  
- AuroraMysqlQueryRunner
- CockroachQueryRunner
- SpannerQueryRunner
- SqlServerQueryRunner

### Pull-Request Checklist

- [x] Fixes #3357

## Summary by CodeRabbit

- **New Features**
  - Improved database schema migrations now safely increase the length of string columns without dropping and recreating them, preserving existing data and reducing downtime for supported databases.
- **Documentation**
  - Added a new section explaining safe and unsafe column alterations, with examples and best practices for managing migrations.
- **Tests**
  - Introduced integration tests to verify that increasing column length uses safe ALTER COLUMN operations and preserves data.

### DIFF
--- a/src/driver/sap/SapQueryRunner.ts
+++ b/src/driver/sap/SapQueryRunner.ts
@@ -1169,10 +1169,10 @@ export class SapQueryRunner extends BaseQueryRunner implements QueryRunner {
 
         // Check if this is a safe length increase for compatible types
         const isSafeLengthIncrease =
-            oldColumn.type === newColumn.type &&
-            oldColumn.length !== undefined &&
-            newColumn.length !== undefined &&
-            newColumn.length > oldColumn.length &&
+            oldColumn.type.toLowerCase() === newColumn.type.toLowerCase() &&
+            oldColumn.length != null &&
+            newColumn.length != null &&
+            Number(newColumn.length) > Number(oldColumn.length) &&
             (newColumn.type.toLowerCase() === "varchar" ||
                 newColumn.type.toLowerCase() === "char" ||
                 newColumn.type.toLowerCase() === "nvarchar" ||
@@ -1196,6 +1196,16 @@ export class SapQueryRunner extends BaseQueryRunner implements QueryRunner {
                     }" ${oldColumn.type.toUpperCase()}(${oldColumn.length}))`,
                 ),
             )
+
+            await this.executeQueries(upQueries, downQueries)
+
+            // sync cache
+            const col = clonedTable.columns.find(
+                (c) => c.name === oldColumn.name,
+            )!
+            col.length = newColumn.length
+            this.replaceCachedTable(table, clonedTable)
+            return
         } else if (
             (newColumn.isGenerated !== oldColumn.isGenerated &&
                 newColumn.generationStrategy !== "uuid") ||
--- a/src/driver/aurora-mysql/AuroraMysqlQueryRunner.ts
+++ b/src/driver/aurora-mysql/AuroraMysqlQueryRunner.ts
@@ -796,15 +796,16 @@ export class AuroraMysqlQueryRunner
         // Check if this is a safe length increase for compatible types
         const isSafeLengthIncrease =
             oldColumn.type === newColumn.type &&
-            (newColumn.type === "varchar" ||
-                newColumn.type === "char" ||
-                newColumn.type === "text") &&
-            oldColumn.length !== undefined &&
-            newColumn.length !== undefined &&
-            parseInt(newColumn.length) > parseInt(oldColumn.length) &&
+            (newColumn.type === "varchar" || newColumn.type === "char") && // TEXT has no length
+            Number.isFinite(+oldColumn.length) &&
+            Number.isFinite(+newColumn.length) &&
+            +newColumn.length > +oldColumn.length &&
             oldColumn.isGenerated === newColumn.isGenerated &&
             oldColumn.generatedType === newColumn.generatedType &&
-            oldColumn.generationStrategy === newColumn.generationStrategy
+            oldColumn.generationStrategy === newColumn.generationStrategy &&
+            oldColumn.isNullable === newColumn.isNullable &&
+            oldColumn.charset === newColumn.charset &&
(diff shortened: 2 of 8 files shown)


elskow:
i also want to propose a feature flag for this change. But before proceeding further. Any suggestion with my current approach?

@alumni @gioboa

src/driver/sap/SapQueryRunner.ts
```diff
@@ -1167,13 +1167,52 @@ export class SapQueryRunner extends BaseQueryRunner implements QueryRunner {
...
+            upQueries.push(
+                new Query(
+                    `ALTER TABLE ${this.escapePath(table)} ALTER ("${
+                        oldColumn.name
+                    }" ${newColumn.type.toUpperCase()}${lengthPart})`,
+                ),
+            )
+            downQueries.push(
+                new Query(
+                    `ALTER TABLE ${this.escapePath(table)} ALTER ("${
```
alumni, line 1194: The up query is safe, but not the down query.

The following scenario will fail:
1. run the migration that increases length from 50 to 100
2. save some data into the table
3. revert the migration - fails.

What we do in our project is this:
```sql
RENAME COLUMN "table_name"."column_name" TO "column_name_typeorm_backup";
ALTER TABLE "table_name" ADD ("column_name" nvarchar(50)); -- full column definition, not just type
UPDATE "table_name" SET "column_name" = LEFT("column_name_typeorm_backup", 50) WHERE LENGTH("column_name_typeorm_backup") > 50;
ALTER TABLE "table_name" DROP ("column_name_typeorm_backup");
```

We could also do the same for the up query when the length decreases 😉 Still better than dropping the column 🤷‍♂

_Note:_ `ADD ("column_name")` and `DROP ("column_name_typeorm_backup")` behave as currently (just we drop the renamed column) - that means the full column definition is necessary (with nullable, collation, comment etc.) AND the indiced/FKs/PKs need to be dropped and recreated.

_Note 2:_ we don't need to do the whole thing I mentioned above in this PR, but we should probably make sure that reverting the migration (so the down query) does not fail (by e.g. using the previous behavior).

_Note 3:_ In the future I might want to extend it for other data types, according to [Data Type Conversion | SAP Help Portal](https://help.sap.com/docs/hana-cloud-database/sap-hana-cloud-sap-hana-database-sql-reference-guide/data-type-conversion).

src/driver/sap/SapQueryRunner.ts
```diff
@@ -1167,13 +1167,52 @@ export class SapQueryRunner extends BaseQueryRunner implements QueryRunner {
                 `Column "${oldTableColumnOrName}" was not found in the "${table.name}" table.`,
             )
 
-        if (
+        // Check if this is a safe length increase for compatible types
+        const isSafeLengthIncrease =
+            oldColumn.type.toLowerCase() === newColumn.type.toLowerCase() &&
```
alumni, line 1172: I believe `ColumnMetadata.type` is always lowercase.

src/driver/sap/SapQueryRunner.ts
```diff
@@ -1167,13 +1167,52 @@ export class SapQueryRunner extends BaseQueryRunner implements QueryRunner {
...
+            oldColumn.length != null &&
+            newColumn.length != null &&
+            Number(newColumn.length) > Number(oldColumn.length) &&
+            (newColumn.type.toLowerCase() === "varchar" ||
+                newColumn.type.toLowerCase() === "char" ||
+                newColumn.type.toLowerCase() === "nvarchar" ||
+                newColumn.type.toLowerCase() === "nchar") &&
+            oldColumn.isGenerated === newColumn.isGenerated
```
alumni, line 1180: I understand the purpose of the check, but this does not suffice. There are many more properties that need to be checked if we want to have a partial definition, e.g. collation, comment (basically everything that goes into the definition). 

Or we should use `Driver.createFullType` when generating the alter table.

src/driver/sap/SapQueryRunner.ts
```diff
@@ -1167,13 +1167,52 @@ export class SapQueryRunner extends BaseQueryRunner implements QueryRunner {
...
+
+        if (isSafeLengthIncrease) {
```
alumni, line 1182: Actually, I think this is the wrong place for this.

It should probably be handled by the `else` block already if the `if` condition is right.

alumni:
I see quite a lot of potential issues in the current state of the PR. There's a lot of differences between drivers which shouldn't be there - I'm especially concerned about MSSql/Oracle/SAP/Spanner, probably Postgres/MySql look slightly better.

I think that it should be done slightly differently (as mentioned in the last code comment) - we are already issuing `ALTER TABLE` statements, we should probably change when we do that instead of adding extra logic.

There are a lot of configuration options for columns which can appear in any combination, and there are quite a lot of drivers, so some tests would be needed.

PR #10933, opened 2024-06-19
fix: update mssql allowed version to fix vulnerability.

wy193777:
### Description of change

Update mssql allowed version to fix a vulneribility: 
[here](https://github.com/advisories/GHSA-m5vv-6r4h-3vj9)

### Pull-Request Checklist

- [x] Code is up-to-date with the `master` branch
- [x] `npm run format` to apply prettier formatting
- [x] `npm run test` passes with this change
- [x] This pull request links relevant issues as Fixes #10946 #11003 
- [ ] There are new or updated unit tests validating the change (N/A)
- [ ] Documentation has been updated to reflect this change (N/A)
- [x] The new commits follow conventions explained in [CONTRIBUTING.md](https://github.com/typeorm/typeorm/blob/master/CONTRIBUTING.md)

PR #11530, opened 2025-06-15
chore(ci): simplify workflows

alumni:
### Description of change

Avoid running the default workflow twice (for both PR and branch).

Fixes the issue described in PR #11519

### Pull-Request Checklist

## Summary by CodeRabbit

- **New Features**
  - Introduced a new workflow for automated commit validation, including formatting, build, cross-platform testing, and coverage reporting.
  - Added a new database service for Postgres 14 with PostGIS support in the container setup.
  - Added Oracle and SAP database testing jobs to the Linux test workflow.

- **Bug Fixes**
  - Updated database container images (MariaDB, Postgres, Oracle, MongoDB) to newer versions for improved compatibility and security.

- **Chores**
  - Removed deprecated and redundant GitHub Actions workflows to streamline CI processes.
  - Renamed and refined workflow names and job labels for clarity.
  - Improved workflow efficiency and coverage reporting integration.
  - Standardized GitHub Actions workflow formatting and triggers.


.github/workflows/commit-validation.yml
```diff
@@ -0,0 +1,74 @@
+name: Commit Validation
+on:
+  push:
+    branches: ["**"]
+    paths-ignore:
+      - "docs/**"
+  pull_request:
+    branches: ["master"]
```
sgarner: Let's widen this so chained PRs can be validated too? I don't see any reason to restrict it.

```suggestion
    branches: ["**"]
```

.github/workflows/commit-validation.yml
```diff
@@ -0,0 +1,74 @@
...
+
+  tests-windows:
+    needs: build
+    uses: ./.github/workflows/tests-windows.yml
+    with:
+      node-version: 20
+
+  # Run with most databases possible to provide the coverage of the tests
+  coverage:
+    if: ${{ (github.event_name != 'pull_request') || (github.event.pull_request.head.repo.fork) }}
```
sgarner: Probably can remove this line, since this job `needs` others jobs that already have the condition.

```suggestion
```

.github/workflows/tests-linux.yml
```diff
@@ -1,14 +1,15 @@
-name: database-tests
-
+name: database-tests-standard
```
sgarner:
```suggestion
name: tests-linux
```

.github/workflows/tests-linux.yml
```diff
@@ -68,10 +69,6 @@ jobs:
 
   mssql:
     runs-on: ubuntu-latest
-    strategy:
-      fail-fast: false
-      matrix:
-        mssql-version: ["^9.1.1", "^10.0.1", "^11.0.1"]
```
sgarner, line 74: I went looking through git blame to figure out why we were testing all these `mssql` client versions in the first place (added in #10933).

Turns out there wasn't really any rationale for it. We don't test any other client libraries to this extent, and the only breaking changes in these `mssql` releases were just dropping old Node versions. So it seems fine to stop doing this. :+1:

.github/workflows/tests-windows.yml
```diff
@@ -1,17 +1,20 @@
 name: database-tests-windows
```
sgarner:
```suggestion
name: tests-windows
```

sgarner (approved):
Great improvement :+1:

.github/workflows/tests-linux.yml
alumni: I agree. Different versions of the clients rarely make a difference. I think we get a bigger difference if we have different DB versions, since they have more changes.

I'm hoping in the next major release we'll have a single major version of the clients as minimum dependency and that we'll bump some docker image versions as well (especially mysql which I can't run locally through docker since it's too old).

.github/workflows/commit-validation.yml
```diff
@@ -0,0 +1,76 @@
...
+
+  tests-windows:
+    needs: build
+    uses: ./.github/workflows/tests-windows.yml
+    with:
+      node-version: 20
+
+  # Run with most databases possible to provide the coverage of the tests
+  coverage:
+    if: ${{ always() }}
```
sgarner: For `step` conditions, `always()` means the step runs even if a prior step fails. But I don't think it does anything for `job` conditions? 🤔

alumni: This is what we had before. Seems that with it, the coverage workflow will run even if the `needs` are not satisfied (erroneous or skipped). Honestly, we don't care about the coverage if a test fails.

The main issue that I see is that there won't be any comment on the PR about the coverage if the tests are skipped (with or without this). But, _IIRC_, the job would fail on the commit if the coverage is below the threshold.

sgarner: Right, that makes sense...

I suppose it could still be useful to see coverage while a few tests were failing. But we don't want coverage when the test jobs are skipped.

Maybe we need something like this? https://github.com/actions/runner/issues/491#issuecomment-850884422

Issue #11514, opened 2025-06-09
TypeError when saving a array type column


alkihis:
### Issue description

Crash when using Repository.save with a column of type 'time', array: true when switching from null to a value, in PostgreSQL

### Expected Behavior

Update should occur without error (this was fine in `0.3.23`).

### Actual Behavior

When using Repository.save on an entity with a column of type `time`, `nullable: true`, `array: true` in PostgreSQL, with the following scenario:
- Entity is fetched using query builder
- `entity.columnWithTypeTime` value was previously `null`
- `entity.columnWithTypeTime = ['12:00']`
- `repository.save(entity)` <- this crashes

Note that crash does *not* happen when this column is not updated, so I suppose this relates to the "diff" process during `.save` to determine which columns should figure in the `UPDATE` request.

```
TypeError: Cannot read properties of null (reading 'map')
     at /app/src/persistence/SubjectChangedColumnsComputer.ts:110:49
     at Array.forEach (<anonymous>)
     at SubjectChangedColumnsComputer.computeDiffColumns (/app/src/persistence/SubjectChangedColumnsComputer.ts:37:34)
     at /app/src/persistence/SubjectChangedColumnsComputer.ts:21:18
     at Array.forEach (<anonymous>)
     at SubjectChangedColumnsComputer.compute (/app/src/persistence/SubjectChangedColumnsComputer.ts:20:18)
     at SubjectExecutor.recompute (/app/src/persistence/SubjectExecutor.ts:209:45)
     at new SubjectExecutor (/app/src/persistence/SubjectExecutor.ts:91:14)
     at /app/src/persistence/EntityPersistExecutor.ts:146:28
     at processTicksAndRejections (node:internal/process/task_queues:105:5)
     at async Promise.all (index 0)
     at EntityPersistExecutor.execute (/app/src/persistence/EntityPersistExecutor.ts:72:31)
```

### Steps to reproduce

Create an entity with a time array column (this might happen with other array types, but I haven't tested yet).

```ts
@Entity('schools', { schema: 'public' })
export class School {
  @PrimaryGeneratedColumn({ type: 'integer', name: 'id' })
  id: number;

  @Column('time without time zone', { nullable: true, name: 'opening_monday', array: true })
  opening_monday: [string, string] | null;
}
```

Then write the following scenario:
- Entity is fetched using query builder
```ts
const school = await connection.getRepository(School).findOne({});
```

- `entity.columnWithTypeTime` value was previously `null`
```ts

[...]

PR #11532, opened 2025-06-17
fix: fix null pointer exception on date array column comparison

mnbaccari:
Closes: https://github.com/typeorm/typeorm/issues/11514

### Description of change

Follow up to https://github.com/typeorm/typeorm/pull/11269 which was calling Array.map without validating if the array was null. Now only normalize if both values (entity and database) are not null. There is not point in normalizing the values if either is null, because the later comparison will just compare with null anyway. Added a test to the original test suite. The null -> value one failed before my change and succeeded after.

### Pull-Request Checklist


alumni (approved):

sgarner (approved):
Perfect, thank you @mnbaccari 👌

gioboa (approved):
Thanks @mnbaccari for your amazing help

PR #10779, opened 2024-03-17
Moved `reflect-metadata` to `peerDependencies` and set version to `"^0.1.14 || ^0.2.0"`

OSA413:
This should fix or improve situation about this bug https://github.com/typeorm/typeorm/issues/10671

Issue #10671, opened 2024-01-31
0.3.20 regression with NestJS, TypeORM and Swagger


OSA413:
### Issue description

Upgrading to 0.3.20 broke NestJS + Swagger integration, downgrading to 0.3.19 fixes the issue

### Expected Behavior

not regression

### Actual Behavior

regression

### Steps to reproduce

https://github.com/OSA413/nestjs-bug-swagger-api-property

### My Environment

| Dependency          | Version  |
| ---                 | ---      |
| Operating System    | Windows 10 |
| Node.js version     | 18.17.1  |
| Typescript version  | not applicable i think  |
| TypeORM version     | 0.3.20  |

### Additional Context

*Reposted from NestJS' Discord*
So, I recently updated packages and Swagger started to throw errors related to @ApiProperty (Optional) and circular dependency though a week ago it was just fine. I've managed to make an MRE from the whole project, here it is: https://github.com/OSA413/nestjs-bug-swagger-api-property

The DTO looks like this:
```
export class Testttttttttt {
    @ApiProperty()
    public id: string;
}
```
It's very interesting, because 1) it doesn't specify any type in the annotation, 2) it's a scalar type

Some other investigations:
1) The bug goes away if I remove TypeORM import from the AppModule
2) The bug may disappear if I leave only one controller by combining two of them
3) The bug may disappear if I move DTO into the same file as controller

For easier reproduction of the example I provided you'll need a Docker (because you need to connect to DB)

Package versions are listed in the repository, `package.lock` is provided

Here's the log
```
[14:50:50] Starting compilation in watch mode...

[14:50:52] Found 0 errors. Watching for file changes.

[Nest] 12856  - 31.01.2024, 14:50:53     LOG [NestFactory] Starting Nest application...
[Nest] 12856  - 31.01.2024, 14:50:53     LOG [InstanceLoader] TypeOrmModule dependencies initialized +46ms
[Nest] 12856  - 31.01.2024, 14:50:53     LOG [InstanceLoader] AppModule dependencies initialized +0ms
[Nest] 12856  - 31.01.2024, 14:50:53     LOG [InstanceLoader] TypeOrmCoreModule dependencies initialized +22ms

C:\Users\Oleg.Sokolov\Documents\GitHub\nestjs-bug-swagger-api-property\node_modules\@nestjs\swagger\dist\services\schema-object-factory.js:187
[...]

PR #11544, opened 2025-06-24
fix: move reflect-metadata to dependencies

mohd-akram:
### Description of change

This fixes installing via `npm install --legacy-peer-deps` which is sometimes needed.

This was introduced in #10779 but wasn't released until a few months ago, even though it was no longer required per https://github.com/typeorm/typeorm/issues/10671#issuecomment-2026816157.

### Pull-Request Checklist

-   [ ] This pull request links relevant issues as `Fixes #00000` N/A
-   [ ] There are new or updated unit tests validating the change N/A
-   [ ] Documentation has been updated to reflect this change N/A

## Summary by CodeRabbit

- **Chores**
  - Updated dependency management by moving "reflect-metadata" to the main dependencies section for improved consistency.


alumni:
TypeORM should work without `reflect-metadata`. It is only necessary if you want to use the decorators, but you don't have to use them.

package.json
```diff
@@ -101,6 +101,7 @@
     "dedent": "^1.6.0",
     "dotenv": "^16.4.7",
     "glob": "^10.4.5",
+    "reflect-metadata": "^0.2.2",
```
OSA413, line 104: Is it really needed? Can you please provide an example of package.json (or a list of dependencies) where the issue is reproduced.

OSA413 (changes requested):

package.json
mohd-akram: If you do `npm install --legacy-peer-deps typeorm` and try to import it, you will get `Error: Cannot find module 'reflect-metadata'` because of [this line](https://github.com/typeorm/typeorm/blob/5904ac3db21d928e858cd7a6e5f4448fc44256d1/src/index.ts#L3).

OSA413: Yes, if you are using TypeScript, you need to also install `reflect-metadata` as mentioned in the  documentation. It looks like a great opportunity to improve our docs I guess.

Please see [Installation](https://github.com/typeorm/typeorm#installation)

OSA413: Hold on, I might be wrong

OSA413: Alright, I've managed to create a repository that can reproduce the error that you described.

OSA413:
So, here's a repository that provides an example of the problem: https://github.com/OSA413/typeorm-playground/commit/e6bda775eadf037b20712d3c22321fdb6684303d

The way it works:
1. Clone the repo and run `npm i`
2. Run `npm start`, it works and returns all entries from the database.

The way it doesn't work:
1. Clone the repo and run `npm i --legacy-peer-deps typeorm`
2. Run `npm start`, it crashes with `TSError: ⨯ Unable to compile TypeScript:
src/index.ts:12:5 - error TS2584: Cannot find name 'console'. Do you need to change your target library? Try changing the 'lib' compiler option to include 'dom'.`

I noticed that running `npm i --legacy-peer-deps typeorm` removes the following packages from the `package-lock.json`:
- `node_modules/@types/node`
- `node_modules/reflect-metadata`
- `node_modules/undici-types`

OSA413:
Installing `npm i --legacy-peer-deps  https://pkg.pr.new/typeorm/typeorm@11544` brings back `reflect-metadata` in the `package-lock.json` but doesn't solve the issue with the console.

OSA413:
@michaelbromley (and you) pointed out that the because the `reflect-metadata` is imported at the index file, it means that the package is used everywhere TypeORM is used. Meaning that `reflect-metadata` should be moved to the `dependencies`.

package.json
OSA413: Ok, the move to the `dependencies` is valid. What about the version? Should it be as it was (`"^0.1.14 || ^0.2.0"`) or as it is?

OSA413: Version `0.1.14` doesn't look deprecated, maybe we should set it as `^0.1.14 || ^0.2.2`?

OSA413:
I've managed to make `reflect-metadata` optional in this repo https://github.com/OSA413/typeorm-playground/commit/80feccd2caebc1ba446b1f0ffd436cf693e09b6c using this PR https://github.com/typeorm/typeorm/pull/11545 (though it still installs with `npm i`)

I would really discuss if either of these changes is needed. @alumni @michaelbromley what do you think?

alumni:
> @michaelbromley (and you) pointed out that the because the `reflect-metadata` is imported at the index file, it means that the package is used everywhere TypeORM is used. Meaning that `reflect-metadata` should be moved to the `dependencies`.

I think it should not be imported by TypeORM, but the users should have to import it if they want to use class-based entities.

I'm not at a computer right now, but I'd say NestJS should require the same thing, except for them it would be a mandatory peer dependency, not an optional one.

alumni:
Saw the PR, I'd remove the import completely.

For a long time I've been wondering how are the tests working if `reflect-metadata` is not imported in the mocha setup script 😛 I didn't expect it would be imported in the productive code 😃

alumni:
See https://github.com/typeorm/typeorm/pull/11545#issuecomment-3001971223

OSA413:
@mohd-akram as a temporal solution I would suggest also installing `reflect-metadata` explicitly.

alumni:
Not temporary: you have to have to install it yourself, as the docs mention: https://typeorm.io/docs/getting-started#installation

mohd-akram:
> @mohd-akram as a temporal solution I would suggest also installing `reflect-metadata` explicitly.

This is what I have resorted to for now. I'm not using TypeScript or decorators though, so that seems wrong/unnecessary.

> Not temporary: you have to have to install it yourself, as the docs mention: https://typeorm.io/docs/getting-started#installation

The docs seem to predate 0.2.0 which added the `reflect-metadata` import to `index.ts`. I expect very few users of TypeORM have actually installed/imported `reflect-metadata` manually, and moving it to peerDependencies was a breaking change.

package.json
mohd-akram: The range is not needed for direct dependencies since they don't need to be shared.

PR #11545, opened 2025-06-24
fix: make import of reflect-metadata optional

OSA413:
### Description of change

There's an import of `reflect-metadata` in the `src/index.ts` that makes the package to be used everywhere instead of making it optional. This *should* fix the issue.

### Pull-Request Checklist

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved error handling when loading certain modules, ensuring the app continues to run smoothly if a module is unavailable.


OSA413:
@coderabbitai review

alumni:
This import should be removed completely, it should only be needed if the user decides to use class based entities. If they decide to use entity schema to define entities, there won't be any decorators needed, so `reflect-metadata` is not required.

Current versions for `reflect-metadata` in `package.json` are correct. It should be an optional peer dependency, both 0.1 and 0.2 should be allowed. In development (tests) we use should 0.2. We can drop 0.1 in the next major.

After we remove this import, most likely the test setup script will need to import it. When you write a test you should **NOT** have to import it as described in `DEVELOPER.md`, that file should be updated too. A lot of tests import this package, that is not needed if the import is in the setup script.

User docs seem fine, they already mention that `reflect-metadata` must be imported by the user and that `experimentalDecorators`/`emitDecoratorMetadata` need to be enabled in TSConfig.

I'd probably restructure those docs a little bit to mention the two ways of declaring entities (entity schema and class-based entities) and mention these requirements only for class-based entities.

gioboa (approved):
LGTM, thanks @OSA413

alumni:
Hey @OSA413, I'm not at a laptop today, could you check what's in `package.json`? I think we made `reflect-metadata` mandatory some time ago.

OSA413:
Yeah, `reflect-metadata` is only mentioned in the `dependencies` so that's mandatory. Should I make it into the `devDependencies` and into the `peerDependencies` and make its `peerDependenciesMeta` optional?

alumni:
> Yeah, `reflect-metadata` is only mentioned in the `dependencies` so that's mandatory. Should I make it into the `devDependencies` and into the `peerDependencies` and make its `peerDependenciesMeta` optional?

What are we trying to solve here? Just to let the users import it in case they need class-based decorators? So that those using `EntitySchema` don't need to rely on it, right?

We can do as you proposed then. We could also cleanup some more things related to it - `reflect-metadata` is probably also not needed in tests since it's imported in `test/utils/test-setup.ts` - so we could also clean those and the docs maybe?

BTW, nowadays `import "reflect-metadata/lite"` instead `import "reflect-metadata"` might also reduce the size of imports by some bytes (no more pollyfills for map/set/weakmap), so we could recommend that in our docs (and do it in test setup too).

michaelbromley:

```
git fetch origin
```

PR #11550, opened 2025-06-25
fix: do not replace entity's propertyName inside literal strings

### LINKED ISSUE #10079: Unwanted quotation marks in json field
### Issue description

When I put a column name inside a json string, it get quoted, generating an invalid SQL

### Expected Behavior

The updated part resulting in the following string
('{"obs":"Increased age by one"}')::jsonb || updated

### Actual Behavior

The updated part is resulting in the following string
('{"obs":"Increased "age" by one"}') || updated

### Steps to reproduce

```typescript
await dataSource
    .createQueryBuilder()
    .update(User)
    .set({
        firstName: "Timber",
        lastName: "Saw",
        age: () => "age + 1",
        updated: () => `('{"obs":"Increased age by one"}') || updated`,
    })
    .where("id = :id", { id: 1 })
    .execute()
```

### My Environment

| Operating System    | Ubuntu server|
| Node.js version     |18.5.0 |
| Typescript version  |5.0.4 |

### Additional Context

_No response_

- [ ] aurora-mysql
- [ ] aurora-postgres
- [ ] better-sqlite3
- [ ] cockroachdb
- [ ] cordova
- [ ] expo
- [ ] mongodb
- [ ] mysql
- [ ] nativescript
- [ ] oracle
- [X] postgres
- [ ] react-native
- [ ] sap
- [ ] spanner
- [ ] sqlite
- [ ] sqlite-abstract
- [ ] sqljs
- [ ] sqlserver

talski:
### Description of change

Fixes #10079 
Strings inside a query should not be modified, currently a string with a token with the same name as one column will be replaced if it was set using a function:

``` column: () => `column || 'my column string'` ```
will result in
``` "column" = "column" || 'my "column" string' ```
with this PR it wil result in
``` "column" = "column" || 'my column string' ```

### Pull-Request Checklist

-   [x] This pull request links relevant issues as `Fixes #10079`

## Summary by CodeRabbit

- **New Features**
  - Added support to prevent unwanted replacements inside string literals when updating property names in SQL queries.
  - Introduced a new User entity with JSONB metadata for tracking updates.
- **Tests**
  - Added tests to verify correct handling of JSON field updates and string literals in SQL queries.

### CODE REVIEW

**Review by alumni — CHANGES_REQUESTED** [review]
This won't work for all drivers because the quotation characters for strings are slightly different depending on the driver.

The first thing that came to mind was MariaDB/MySQL, where strings can be between double quotes (identifiers are between backticks).

For sure we need to run the tests on all drivers (not just Postgres) and maybe check possible edge cases.

**Comment by talski** [comment]
Couldn't we start with ANSI SQL only? This bug was introduced after version 0.3.13, I haven't been able to update since

What about those regexp?

v1 https://regex101.com/r/7aW33t/1

v2 https://regex101.com/r/O5iZR7/1

**Comment by sgarner** [comment]
@talski If you want to workaround this issue, can you use a parameter instead? Generally it's best to avoid putting string literals in your queries at all.

> This won't work for all drivers because the quotation characters for strings are slightly different depending on the driver.
> 
> The first thing that came to mind was MariaDB/MySQL, where strings can be between double quotes (identifiers are between backticks).

That's actually a [configuration option](https://dev.mysql.com/doc/refman/8.4/en/sql-mode.html#sqlmode_ansi_quotes) in MariaDB/MySQL though, so without inspecting the server config we can't really be sure how to interpret double quotes in an input string.

@alumni Are you aware of any other engines with oddities like this?

Since using double quotes for string literals is non-standard I'd be inclined to say we ignore them for the purpose of this issue, i.e. the regex should only affect single-quoted strings.

**Comment by alumni** [comment]
TypeORM works with the default syntax of MySQL, switching to the ANSI syntax in options will break things.

I only know about MySQL because I also had to deal with identifier quotation. For identifiers there are more exceptions, for strings IDK.

**Comment by talski** [comment]
@sgarner sometimes I need to update an jsonb array this way to avoid data loss in concurrency, to append/prepend some info

```"updated" = ('{"at":"2025-06-25 14:48:11.092 Z","by":1,"obs":"log some change"}')::jsonb || updated```

This could be enabled to ANSI only databases or at least have an option to enable

**Comment by sgarner** [comment]
> @sgarner sometimes I need to update an jsonb array this way to avoid data loss in concurrency, to append/prepend some info
> 
> `"updated" = ('{"at":"2025-06-25 14:48:11.092 Z","by":1,"obs":"log some change"}')::jsonb || updated`

Still, you should be able to use a parameter for that:

```typescript
dataSource.getRepository(Foo).createQueryBuilder('foo')
  .update()
  .set({
    updated: () => 'to_jsonb(:updatedJson) || updated'
  })
  .setParameter('updatedJson', JSON.stringify({ at: '2025-06-25 14:48:11.092 Z', by: 1, obs: 'log some change' }))
  .where({ ... })
  .execute()
```

talski:
@sgarner never thought using this way, I will try it out, thanks

gioboa (commented):
Hi @talski is this PR still valid?

talski:
> Hi @talski is this PR still valid?

In my POV, yes, there are workarounds to this problem, but the current propertyName replacements are buggy, this PR will prevent generating wrong queries

talski:
could this please be merged with 0.3 version?

PR #11555, opened 2025-07-02
fix(mysql): support AnalyticDB returning version() column name in getVersion()

### LINKED ISSUE #11553: Bug: getVersion() fails with AnalyticDB MySQL due to non-standard field name version()
### Issue description

When using TypeORM with Alibaba Cloud AnalyticDB for MySQL (ADS), the `getVersion()` function in `MysqlQueryRunner` fails due to ADS returning the version information under the key `version()` instead of `version`.  This causes a TypeError when calling `.replace()` on undefined if `result[0].version` is not available.  Related driver: mysql / mysql2

### Expected Behavior

TypeORM should correctly parse the version string returned from ADS, even when the field name is `version()`, and not throw an error.

### Actual Behavior

TypeORM crashes with:

TypeError: Cannot read properties of undefined (reading 'replace')
    at MysqlQueryRunner.getVersion (...)

The query result from ADS is:

[ { 'version()': '5.6.16-log' } ]

In the current implementation, `result[0].version` is undefined, and `result[0]['version()']` should be used instead.

### Steps to reproduce

1. Use TypeORM with `type: 'mysql'`
2. Connect to Alibaba Cloud AnalyticDB for MySQL (ADS)
3. Let TypeORM call `getVersion()` internally
4. Observe TypeError when parsing the version string

### My Environment

| Operating System    |    macos      |
| Node.js version     | 22.0.0 |
| Typescript version  | 5.1.3 |

### Additional Context

_No response_

- [ ] aurora-mysql
- [ ] aurora-postgres
- [ ] better-sqlite3
- [ ] cockroachdb
- [ ] cordova
- [ ] expo
- [ ] mongodb
- [ ] mysql
- [ ] nativescript
- [ ] oracle
- [ ] postgres
- [ ] react-native
- [ ] sap
- [ ] spanner
- [ ] sqlite
- [ ] sqlite-abstract
- [ ] sqljs
- [ ] sqlserver

Minglu-Huo:
### Description of change

When using Alibaba Cloud AnalyticDB for MySQL (ADS), running `SELECT VERSION() AS version` returns a column named `version()` instead of `version`, which causes TypeORM to throw a TypeError in `MysqlQueryRunner.getVersion()`.

This change updates `getVersion()` to support both `version` and `version()` column keys, making it compatible with ADS while keeping compatibility with standard MySQL and MariaDB.

Verified locally using ADS instance, and also tested on standard MySQL.

Related issue: https://github.com/typeorm/typeorm/issues/11553

---

### Pull-Request Checklist

-   [x] This pull request links relevant issues as `Fixes #11553`
-   [ ] There are new or updated unit tests validating the change (N/A)
-   [ ] Documentation has been updated to reflect this change (N/A)

---

🎉 Thank you for reviewing this PR!

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved compatibility when retrieving MySQL version information, ensuring the version is correctly detected regardless of how the database returns the version string.


rhydian0x:
Hi @alumni , could you please help review this PR when you have time? Thanks!

src/driver/mysql/MysqlQueryRunner.ts
```diff
@@ -3371,16 +3371,25 @@ export class MysqlQueryRunner extends BaseQueryRunner implements QueryRunner {
     }
 
     async getVersion(): Promise<string> {
-        const result: [{ version: string }] = await this.query(
+        const result = await this.query(
             `SELECT VERSION() AS \`version\``,
```
alumni: Let's simplify. As discussed in the issue, `SELECT version()` would work on all versions:

```suggestion
        const result: [{ "version()": string }] = await this.query(
            `SELECT version()`,
```

src/driver/mysql/MysqlQueryRunner.ts
```diff
@@ -3371,16 +3371,25 @@ export class MysqlQueryRunner extends BaseQueryRunner implements QueryRunner {
...
-        const versionString = result[0].version
+
+        // support AnalyticDB for MySQL returning version() column name in getVersion
+        const row = result[0] as { version?: string; "version()"?: string };
+
+        const versionString = row.version ?? row["version()"];
+
+        if (!versionString) {
+            throw new TypeORMError("Unable to determine database version: neither 'version' nor 'version()' property found in query result");
+        }
```
alumni: Now, this whole thing would become:

```suggestion
        const versionString = result[0]["version()"]
```

alumni (changes requested):

rhydian0x:
@alumni 
Thanks! Updated the code as suggested. Please take another look when you have time 🙏

alumni:
@Minglu-Huo the format check will fail if I will start the workflow :) Could you fix the formatting? Thanks :)

`npm run format` if your editor is not set up to use prettier.

rhydian0x:
@alumni 
✅ Done. I’ve run npm run format and pushed the changes.

src/driver/mysql/MysqlQueryRunner.ts
```diff
@@ -3371,16 +3371,18 @@ export class MysqlQueryRunner extends BaseQueryRunner implements QueryRunner {
     }
 
     async getVersion(): Promise<string> {
-        const result: [{ version: string }] = await this.query(
-            `SELECT VERSION() AS \`version\``,
+        const result: [{ "version()": string }] = await this.query(
+            `SELECT VERSION()`,
```
alumni:
```suggestion
            `SELECT version()`,
```

otherwise, the line below will throw an exception.

rhydian0x:
Thanks @alumni for the feedback! I've updated the query to use lowercase `version()` as suggested to avoid the exception.

alumni (approved):

alumni:
Thanks for the fix 😀

gioboa (approved):
Thanks for your help @Minglu-Huo 👍

rhydian0x:
@alumni 
Hi, thanks for merging the PR! I was wondering when this change will be published in a release?

PR #11568, opened 2025-07-14
fix: regtype is not supported in aurora serverless v2

ArsenyYankovsky:
### Description of change

Fixes an issue with "regtype" being unsupported in Aurora Serverless V2 ([more details here](https://github.com/ArsenyYankovsky/typeorm-aurora-data-api-driver/issues/164#issuecomment-2133040680)).

### Pull-Request Checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved accuracy in displaying user-defined type information for columns in PostgreSQL database metadata views.


alumni (changes requested):
If we do it like this, the Aurora driver will not benefit from the future updates we plan to do to the Postgres driver.

Would be great if you could make it work in the `PostgresQueryRunner` by finding a solution that works on all platforms, instead of copy-pasting the `loadTables` method and changing 2 lines.

ArsenyYankovsky:
Fair enough, I changed the type of the column in the parent class (it's used as a string) which works for both Aurora and regular Postgres.

ArsenyYankovsky:
I see the tests are failing, will check later this week.

ArsenyYankovsky:
Should be good now.

PR #11576, opened 2025-07-21
fix(sqlserver): queries returning multiple record sets

### LINKED ISSUE #5051: Multiple Recordsets in stored procedure
We are using typeorm to execute a stored procedure in a legacy mssql DB. The procedure has multiple selects, which causes multiple recordsets to be return, and we are interested in the last one, which is the final result of the stored procedure.

The result from the mssql driver contains both recordsets array (in a recordsets property), and the first recordset (in a dedicated recordset property) (see: https://github.com/tediousjs/node-mssql/blob/624219ab0c945d57b72b0d9a949695ff54feaba5/lib/base.js#L1457) 

Typeorm's sqlserver query runner returns only the recordset property (see: https://github.com/typeorm/typeorm/blob/c87e4e7ad92023a1914a6a1778d683ba8f23c5ab/src/driver/sqlserver/SqlServerQueryRunner.ts#L233), so we are getting the first recordset instead of the last.

Is there a way to change this? If not, wouldn't it be more logical to return the last recordset?

Thanks!

### LINKED ISSUE #11566: MSSQL EntityManager.query(`EXEC <SP>`, [param]) doesn't return all recordsets, only the first one
### Issue description

Calling manager.query(`EXEC MySP`, [param1]) only returns the first recordset

### Expected Behavior

I expect it to return exactly what sql returns. If I pass in a single statement it should return that result, if I pass in multiple, it should return all recordsets.

Let me know if this is by design.

### Actual Behavior

Calling manager.query(`EXEC MySP`, [param1]) (or anywhere you can call query()) will return whatever your first statemement returns. If you're returning multiple recordsets, they are not all returned. 

The EXEC example was the first one I tried as I happened to need to call an SP, however, the same applies for any query with multiple statements inside it.

For example a plain SELECT:

`
manager.query(
  SELECT * FROM Table1
  SELECT * FROM Table 2
)
`

You only get whatever is inside Table1.

### Steps to reproduce

EntityManager is exactly that from typeorm. Nothing special here:

`
  const result = await EntityManager.query(`EXEC api.Test ${params.map((_, index) => `@${index}`).join(', ')}`, ['param1', 'param2'])

`

Minimal stored procedure returning two recordsets:

`USE SomeDatabase
GO
/****** Object:  StoredProcedure [api].[Test]    Script Date: 9/07/2025 6:39:13 am ******/
SET ANSI_NULLS ON
GO
SET QUOTED_IDENTIFIER ON
GO
-- =============================================
-- Author:		<Author,,Name>
-- Create date: <Create Date,,>
-- Description:	<Description,,>
-- =============================================
ALTER PROCEDURE [api].[Test]
	@param1 nvarchar,
	@param2 nvarchar
AS
BEGIN

	SET NOCOUNT ON;

	SELECT @param1 as Parameter1

	SELECT @param2 as Parameter2

END
`

The results should be an array of arrays or however typeorm wants to deconstruct the response from sql (if it doesn't just pass it back but here's what I get instead:

`[ { Parameter1: 'v' } ]`

### My Environment

| Operating System    |     Windows 11     |
| Node.js version     | 22.14.0  |
| Typescript version  | 5.8.2  |

### Additional Context

The underlying mssql driver returns **recordset** and **recordsets** (contains all recordsets returned from a query) and **rowsAffected**. It needs to check for recordsets the same way it checks for recordset and rowsAffected.

I can see in the SqlServerQueryRunner.ts, in the query() method, it currently checks for recordset and rowsAffected:

`
            const raw = await new Promise<any>((ok, fail) => {
                request.query(query, (err: any, raw: any) => {
                    // log slow queries if maxQueryExecution time is set
                    const maxQueryExecutionTime =
                        this.driver.options.maxQueryExecutionTime
                    const queryEndTime = Date.now()
                    const queryExecutionTime = queryEndTime - queryStartTime

                    this.broadcaster.broadcastAfterQueryEvent(
                        broadcasterResult,
                        query,
                        parameters,
                        true,
                        queryExecutionTime,
                        raw,
                        undefined,
                    )

                    if (
                        maxQueryExecutionTime &&
                        queryExecutionTime > maxQueryExecutionTime
                    ) {
                        this.driver.connection.logger.logQuerySlow(
                            queryExecutionTime,
                            query,
                            parameters,
                            this,
                        )
                    }

                    if (err) {
                        fail(new QueryFailedError(query, parameters, err))
                    }

                    ok(raw)
                })
            })

            const result = new QueryResult()

            if (raw?.hasOwnProperty("recordset")) {
                result.records = raw.recordset
            }

            if (raw?.hasOwnProperty("rowsAffected")) {
                result.affected = raw.rowsAffected[0]
            }

            const queryType = query.slice(0, query.indexOf(" "))
            switch (queryType) {
                case "DELETE":
                    // for DELETE query additionally return number of affected rows
                    result.raw = [raw.recordset, raw.rowsAffected[0]]
                    break
                default:
                    result.raw = raw.recordset
            }

            if (useStructuredResult) {
                return result
            } else {
                return result.raw
            }
`

The QueryResult class would need to have the appropriate key added to support this property too.

The return would need to be modified to support this. The useStructuredResult isn't exposed as far as `DataSource.query()` so one would need to fallback to `DataSource.createQueryRunner.query(query, params, true)`

So there's some options. It would be nice if useStructuredResult was exposed to DataSource.query() and with the modified QueryResult  and the setting of recordSets we'll be able to pull out recordSets and see the full response.

- [ ] aurora-mysql
- [ ] aurora-postgres
- [ ] better-sqlite3
- [ ] cockroachdb
- [ ] cordova
- [ ] expo
- [ ] mongodb
- [ ] mysql
- [ ] nativescript
- [ ] oracle
- [ ] postgres
- [ ] react-native
- [ ] sap
- [ ] spanner
- [ ] sqlite
- [ ] sqlite-abstract
- [ ] sqljs
- [x] sqlserver

lboonekamp:
### Description of change

Fixes #11566, fixes #5051

#### What?
The change resolves the issue whereby any call to `query()` for mssql connections, returns only the first recordset, even when a query returns multiple recordsets. This change checks for recordsets and adds it to the `QueryResult` if `useStructuredQuery = true` is passed in. This change will allow the passing in of a boolean, anywhere `query()` is called, which will enable the full list of results to be accessed. This change does not change the default behaviour.

#### Why?
 Without this change, you cannot access all recordsets in the return, you only get the first one.

#### Verification
Run a stored procedure that returns multiple recordsets i.e. `.query('MY_SP')`.
Run two select statements within a single call to `.query('SELECT 1; SELECT 2')`.

### Pull-Request Checklist

-   [x] This pull request links relevant issues as `Fixes #11566, #5051`

## Summary by CodeRabbit

- New Features
  - Introduced QueryOptions to request structured results from raw queries across data sources, entity managers, and query runners.
  - getRawMany now accepts options to return structured results when requested.
  - Query results now expose recordsets for multi-statement responses.

- Deprecations
  - The boolean overload for requesting structured results is deprecated; use QueryOptions instead.

- Documentation
  - Added guidance and examples for executing raw SQL with structured results, including migration notes and API usage.


alumni:
I wonder whether we should make the last parameter of `query` an object of type `QueryOptions` containing the property `useStructuredResult?: boolean`.

Currently, in `QueryRunner.query` the last parameter is boolean just for this option, but the database clients that TypeORM uses can accept options for executing statements (e.g. to return rows are arrays instead of objects).

.gitignore
```diff
@@ -29,3 +29,5 @@ ormlogs.log
 /test/github-issues/799/tmp/*
 # test projects generated by test/github-issues/issue-8975.ts
 /*TestProject
+
+.history
```
alumni: unnecessary change?

src/data-source/DataSource.ts
```diff
@@ -538,7 +539,11 @@ export class DataSource {
         const usedQueryRunner = queryRunner || this.createQueryRunner()
 
         try {
-            return await usedQueryRunner.query(query, parameters) // await is needed here because we are using finally
+            return (await usedQueryRunner.query(
+                query,
+                parameters,
+                (useStructuredResult || false) as any,
+            )) as T // await is needed here because we are using finally
```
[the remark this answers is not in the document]
alumni: In addition to that: might as well add the generic to `QueryRunner.query`

lboonekamp:
> I wonder whether we should make the last parameter of `query` an object of type `QueryOptions` containing the property `useStructuredResult?: boolean`.
> 
> Currently, in `QueryRunner.query` the last parameter is boolean just for this option, but the database clients that TypeORM uses can accept options for executing statements (e.g. to return rows are arrays instead of objects).

Great idea. Objects are much more extendable. Let me know if you want me to make the changes.

src/data-source/DataSource.ts
```diff
@@ -528,6 +528,7 @@ export class DataSource {
         query: string,
         parameters?: any[],
         queryRunner?: QueryRunner,
+        useStructuredResult?: boolean,
```
sgarner: I agree with @alumni that an options object would be preferable to a boolean.

It's also critical that the return type correctly reflects this option, which you can do using a method overload as suggested by coderabbit.

docs/docs/working-with-entity-manager/5-entity-manager-api.md
```diff
@@ -62,6 +62,7 @@ const rawData = await manager.query(
 const rawData = await manager.query(
     "SELECT * FROM USERS WHERE name = @0 and age = @1",
     ["John", 24],
+    true, // optional flag to return a structured result set object containing all recordsets
```
sgarner: I don't think this is the right place to show this flag, we don't want to suggest that it should be used by default when querying mssql.

And the structured result feature also applies to other drivers, even if they don't all support multiple recordsets.

Suggest we add another paragraph describing the `query` options object just below this example.

docs/docs/advanced-topics/2-transactions.md
```diff
@@ -74,6 +74,18 @@ await queryRunner.connect()
 // now we can execute any queries on a query runner, for example:
 await queryRunner.query("SELECT * FROM users")
 
+// **NOTE:** if you need to access multiple recordsets from a single query, pass true as the third parameter to the `query` method:
```
sgarner: The topic of this doc page is Transactions. This is not the right place to document the structured result feature.

There is a [Query Runner](https://typeorm.io/docs/query-runner) page where it could go instead.

sgarner (changes requested):
A few improvements can be made but otherwise this looks good, thanks @lboonekamp !

Should we add this option to `SelectQueryBuilder.getRawMany()` as well?

test/functional/query-runner/structured-result.ts
```diff
@@ -0,0 +1,138 @@
+import { QueryResult } from "../../../src"
+import { DataSource } from "../../../src/data-source/DataSource"
+import { EntityManager } from "../../../src/entity-manager/EntityManager"
+import { QueryRunner } from "../../../src/query-runner/QueryRunner"
+import { expect } from "chai"
+import sinon from "sinon"
+import "reflect-metadata"
+
+describe("github issues > #11566 .query() useStructuredResult option", () => {
```
sgarner: We are moving away from creating tests for specific GitHub issues. New PRs should instead add to the test suites of their related features.

Let's put these tests in `test/functional/query-runner/structured-result.ts`?

sgarner (changes requested):

lboonekamp:
Super annoying. Almost like I didn't commit properly. I'll push an update shortly to fix this up.

BerenLongstreet:
Love seeing the movement on this issue! Wondering what's holding up the merge...

lboonekamp:
@sgarner - what are your thoughts on the last change suggested by coderabbit "Handle multiple SQL Server result sets"? I've fixed the first change (pending commit). If we flatten the array, it defeats the purpose of the recordsets property that was added to QueryResult.ts.

BerenLongstreet:
Yeah that code rabbit suggestion doesn't make any sense

lboonekamp:
> Yeah that code rabbit suggestion doesn't make any sense

My thoughts exactly.

gioboa (commented):
Can I ask you to solve the conflicts please?

alumni:
The tests need to end in `.test.ts`.

Also: Since this is going to make it into 1.0, breaking changes are accepted (they should be documented in migration to v1 doc page though)

gioboa (changes requested):
Tests are failing, can you fix the issues please?

lboonekamp:
> Tests are failing, can you fix the issues please?

Was just getting back into this. I have a question for you. The current implementation of getRawMany() returns an array. Do we want to maintain the array, even if useStructuredResult is passed in as true? The only thing then, is we lose the intention of adding the useStructuredResult option to getRawMany().

`async getRawMany<T = any>(options?: QueryOptions): Promise<T[]> {
        if (this.expressionMap.lockMode === "optimistic")
            throw new OptimisticLockCanNotBeUsedError()

        this.expressionMap.queryEntity = false
        const queryRunner = this.obtainQueryRunner()
        let transactionStartedByUs: boolean = false
        try {
            // start transaction if it was enabled
            if (
                this.expressionMap.useTransaction === true &&
                queryRunner.isTransactionActive === false
            ) {
                await queryRunner.startTransaction()
                transactionStartedByUs = true
            }

            const useStructured = options?.useStructuredResult === true;

            const results = await this.loadRawResults(queryRunner, useStructured)

            // close transaction if we started it
            if (transactionStartedByUs) {
                await queryRunner.commitTransaction()
            }

            if(useStructured) {
                return results.records
            }

            return results`

pkuczynski:
I've merged latest master (resolved conflicts), and made a few cleanup commits:
- Reverted unnecessary `.gitignore` change (the test it referenced no longer exists)
- Moved test to `test/unit/` (it's stub-based, not a functional/integration test)
- Updated migration guide to use the Before/After code pattern
- Removed deprecation notes from docs — v1 is a major release, no need to deprecate

**Observations for the maintainers meeting:**

1. **Do we need `QueryOptions` at all?** On master, `queryRunner.query(sql, params, true)` already returns `QueryResult` — it's used internally by all query builders. This PR wraps that boolean in a `QueryOptions` object and threads it through `DataSource.query()` and `EntityManager.query()`. But we could just expose the existing boolean parameter up through those methods — simpler, no new types, same result.

2. **`getRawMany(options)` doesn't make sense.** The `recordsets` feature is for stored procedures and multi-statement raw SQL — things you'd call via `.query()` directly, not via the query builder. A SELECT built by the query builder will never produce multiple recordsets, and `affected` is meaningless for SELECTs. I'd drop the `QueryOptions` addition to `getRawMany()`.

3. **No integration test against SQL Server.** The test is purely stub-based — it verifies the plumbing but never runs a multi-recordset query against a real SQL Server. The core value prop (accessing multiple recordsets from stored procedures) is untested end-to-end.

4. **`getRawMany` return type is wrong when `useStructuredResult: true`.** The signature says `Promise<T[]>` but it actually returns a `QueryResult` — a type safety hole. This is moot if we drop point 2.

Overall the feature itself is valid — SQL Server users need access to multiple recordsets. The implementation just needs simplification: expose the existing `query(sql, params, true)` → `QueryResult` through `DataSource.query()` and `EntityManager.query()`, add `recordsets` to `QueryResult`, and skip the `QueryOptions`/`getRawMany` additions.

gioboa (approved):

michaelbromley:

```
git fetch origin
```

PR #11580, opened 2025-07-25
perf: Cache package.json location between getNearestPackageJson invocations

### LINKED ISSUE #4136: Migration performance correlated to number of migration files
**Issue type:**

[ ] question
[x] bug report
[ ] feature request
[ ] documentation issue

**Database system/driver:**

[ ] `cordova`
[ ] `mongodb`
[ ] `mssql`
[ ] `mysql` / `mariadb`
[ ] `oracle`
[x] `postgres`
[ ] `cockroachdb`
[ ] `sqlite`
[ ] `sqljs`
[ ] `react-native`
[ ] `expo`

**TypeORM version:**

[x] `latest`
[ ] `@next`

We've got a pretty large TS project with 50 migration files and have been experiencing a gradual slowdown in running any migration.

For example `yarn typeorm -f .ormconfig migration:run` used to take just under 10 seconds a few months ago, but now its takin 120 seconds even if there are no migrations to run.

Similarly, generating migrations has been slowing down and is also now taking 120 seconds.

Some of the slowdown is obviously because of increase TS compile time of the project, but there is a distinct correlation between the performance of the migrator and the number of migration files in the project. 

Deleting all but the latest migration files will decrease the 120-second migration time to 15 seconds. The TS compile time for these migrations is just a few seconds, so it seems that TypeORM is doing some unnecessary work for each added migration file.

rutkowskib:
Cache package.json location to improve performance of migrations with a lot of files

Closes: #4136

### Description of change

Currently for every file imported from the same directory code tries to find package.json. This makes migrations with a lot of files very slow because for every a lot of operations need to be performed despite all of the files potentially being in the same directory.
This update makes sure that the time to start migrations stays the same no matter how many migrations we have. I have added a cache that store paths that we know where package.json exists.
I have added a test that verifies that verifies that cache is correctly used to verify working of this change.

### Pull-Request Checklist 

-   [x] This pull request links relevant issues as `Fixes #4136`
-   [ ] Documentation has been updated to reflect this change - N/A

## Summary by CodeRabbit

* **Performance Improvements**
  * Added a bounded cache for nearest package.json lookups to reduce redundant filesystem reads and JSON parsing, improving repeated-lookup performance while keeping memory usage controlled.

* **Tests**
  * Added unit tests verifying the cache prevents unnecessary filesystem operations across repeated imports and confirms correct cache behavior during multiple lookups.


src/util/ImportUtils.ts
```diff
@@ -40,11 +40,30 @@ export async function importOrRequireFile(
     return tryToRequire()
 }
 
+const packageJsonCache: Record<string, object | null> = {}
```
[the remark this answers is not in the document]
alumni: I actually thought about using a `Map` instead of an object.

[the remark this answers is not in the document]
rutkowskib: @alumni What is the process now? Is there going to be some explicit change request, or shall I just change those stuff you mentioned?

test/github-issues/4136/issue-4136.ts
alumni, line 1: The existing tests for `ImportUtils` are placed in `test/unit/util/import-utils.ts`. 

Would be great to have them there, since we are trying to move away from per-issue tests (which are executed at least once for every database driver).

If you consider that multiple files are better, you can create the `import-utils` folder and move the existing tests there, together with this file (with appropriate names/test descriptions, of course)

src/util/ImportUtils.ts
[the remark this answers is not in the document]
alumni: @rutkowskib I didn't want to request changes because it blocks the review process and I want this merged 😆  

Overall the PR looks good, only wrote the two comments, also coderabbit has some valid points.

I would:
1) use `Map` here
2) move the tests together with the other tests for import utils

That would be good enough for me 🙂

You will also need a second review, probably the second reviewer would have some more ideas.

Would be great to see this PR getting merged, we also have long startup times, likely due to the many migrations that we accumulated over the years (we load entities by passing the class directly, and only have a few subscribers, but migrations keep growing).

src/util/ImportUtils.ts
```diff
@@ -40,11 +40,30 @@
...
+    const paths: string[] = []
 
     while (currentPath !== path.dirname(currentPath)) {
         currentPath = path.dirname(currentPath)
+
+        // Check if we have already cached the package.json for this path
+        if (packageJsonCache[currentPath] !== undefined) {
+            setPackageJsonCache(paths, packageJsonCache[currentPath])
+            return packageJsonCache[currentPath]
+        }
```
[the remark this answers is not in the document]
alumni: `currentPath` is the output of `path.dirname` in line 56. Isn't it already normalized?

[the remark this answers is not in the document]
alumni: It seems we don't need to resolve the path or normalize it because `importOrRequireFile` is always called with the resolved file path (e.g. in `importClassesFromDirectories` from `src/util/DirectoryExportedClassesLoader.ts`)

src/util/ImportUtils.ts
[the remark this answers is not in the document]
rutkowskib: Got you. I will do the changes next week

test/unit/util/import-utils.ts
```diff
@@ -177,4 +180,57 @@ describe("ImportUtils.importOrRequireFile", () => {
...
+            "readFile should be called only during the first import",
+        )
+
+        // Clean up test files
+        await fs.unlink(filePath1)
+        await fs.unlink(filePath2)
+        await fs.unlink(filePath3)
+
+        sinon.restore()
+    })
```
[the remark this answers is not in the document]
rutkowskib: I cant do

```
-import fsAsync from "fs"
+import { promises as fsPromises } from "fs"
```
because sinon works only when mocking root package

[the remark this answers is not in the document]
rutkowskib: I cant do

```
-import fsAsync from "fs"
+import { promises as fsPromises } from "fs"
```
because sinon works only when mocking root package

test/unit/util/import-utils.ts
```diff
@@ -177,4 +180,57 @@ describe("ImportUtils.importOrRequireFile", () => {
...
+
+        await fs.writeFile(filePath1, "", "utf8")
+        await fs.writeFile(filePath2, "", "utf8")
+        await fs.writeFile(filePath3, "", "utf8")
+
+        // Trigger the first import to create the cache
+        await importOrRequireFile(filePath1)
+
+        // Get the number of calls to stat and readFile after the first import
+        const numberOfStatCalls = statStub.callCount
```
OSA413: I would also add a check of the actual number of calls to see that the call is actually happening before caching. e.g. not store a statStub.callCount into a variable but use a constant number instead.

OSA413: Just in case to see if it (the file read) is working at all

alumni (approved):

gioboa (approved):

test/unit/util/import-utils.ts
rutkowskib: Would that work for you if I added a check that `numberOfStatCalls` isnt a 0 instead?

OSA413: I think it should be checked against a constant value. Can you check what number is returned by `statStub.callCount`? Enter that value into the assertion instead.

rutkowskib: Tests are now failing. I think it is because they are built differently on my local machine then in CI. What is your suggestion? Shall I create package.json in the test? Shall I revert?

OSA413: Change the number from 6 to 1. That should fix it

rutkowskib: It will break running tests on local though

OSA413: Ok, revert the change, but leave a comment so we will look at it later

OSA413: Ok, I didn't think of such workaround, but I guess it's better than leaving it unknown.

OSA413: In case if you can't fix it, you can revert to the point it was working on both GitHub Actions and locally.

rutkowskib: I have fixed them, you can re review

OSA413 (approved):

test/unit/util/import-utils.ts
```diff
@@ -177,4 +180,85 @@ describe("ImportUtils.importOrRequireFile", () => {
 
         await fs.rmdir(testDir, { recursive: true })
     })
+
+    it("Should use cache to find package.json", async () => {
+        // Create package.json if not exists
+        const packageJsonPath = path.join(__dirname, "package.json")
+        const packageJsonAlreadyExisted = fsAsync.existsSync(packageJsonPath)
```
OSA413, line 187: So, can you describe why sometimes the package.json is created and sometimes it's not?

OSA413 (approved):

test/unit/util/import-utils.ts
rutkowskib: In CI it does exist in the folder that test is run from. On local it doesnt. I dont fully know why it's like that, you would have to ask somebody who knows more about how tests are run in CI.

PR #11587, opened 2025-08-03
feat: introduce OmitRelations type

Newbie012:
### Description of change

Currently, it's a bit challenging instantiating TypeORM entities using the constructor due to the relations:

```ts
export class Cover {
    @PrimaryGeneratedColumn()
    id: number

    @Column()
    url: string

    @OneToMany(() => Post, (post) => post.cover)
    posts: Post[]

    constructor(cover: Cover) {
        // ...
    }
}

const cover = new Cover({
    id: uuidv4(),
    url: "https://example.com/cover.jpg",
    // `posts` is required here, but we don't want to pass it
})
```

Of course, we could define all relations as optional (and we probably should), but even then, it's a bit odd that we have to pass the `posts` property to the constructor.

Luckily, we already have the type [`Relation`](https://typeorm.io/docs/getting-started/#relations-in-esm-projects) which originally aimed to solve the circular reference issue in ESM.

But what if we could use it to identify relations inside the model and omit them from the constructor?

```ts
class Cover {
    // ...
    constructor(cover: OmitRelations<Cover>) {
        // ...
    }
}
```

This PR introduces a new type `OmitRelations<T>` which is a utility type that omits all relations from the type `T` by attaching a `[relationSymbol]` and then excluding all properties that have this symbol.

### Pull-Request Checklist

## Summary by CodeRabbit

* **New Features**
  * Added support for automatically omitting relation properties from objects when needed. 

* **Refactor**
  * Improved internal handling of relation types to better distinguish them from non-relation properties.

### DIFF
--- a/src/common/RelationType.ts
+++ b/src/common/RelationType.ts
@@ -1,3 +1,5 @@
+declare const relationSymbol: unique symbol
+
 /**
  * Wrapper type for relation type definitions in entities.
  * Used to circumvent ESM modules circular dependency issue caused by reflection metadata saving the type of the property.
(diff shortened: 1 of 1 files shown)


Johnny-Shir (approved):
Looks fantastic! Very useful!

OSA413:
Yeah, the feature looks useful but the described usecase seems a bit odd to me. I think I should investigate it a bit further (like making the relation optional on the type level without introducing the `OmitRelations<T>`)

alumni:
I agree with @OSA413, relations (and initially non-selectable columns) should be marked as optional by the user.

sgarner:
I agree with above comments. Also note that entity classes generally shouldn't have custom constructors. The proper way to create a new TypeORM entity instance is with `Repository.create()`.

PR #11602, opened 2025-08-16
chore: Release v0.3.26

michaelbromley:
v0.3.26 release


alumni (approved):

alumni:
@michaelbromley might be worth mentioning `stringifyObjects` is now default for MySQL - users can disable it by setting `connectionOptions.extra.stringifyObjects = false`

PR #11620, opened 2025-08-23
fix(schema): use ALTER COLUMN instead of DROP/ADD for type changes

BenraouaneSoufiane:
## Summary

Some column changes currently follow the `DROP/ADD` path, which can lead to data loss (reported in #3357).

This PR updates the `DROP/ADD` guard so that columns are recreated only when the change is not safely alterable (for example, cross-family type changes). Otherwise, the change proceeds through the driver-specific `ALTER COLUMN` path.

Examples of changes that should not require `DROP/ADD`:

### Strings (widening only)

* `CHAR` → `VARCHAR` (same or larger length)
* `NCHAR` → `NVARCHAR` (same or larger length)
* `VARCHAR(n)` → `VARCHAR(m)` where `m >= n`
* `NVARCHAR2` widening (Oracle)
* `VARCHAR` → `TEXT` / `NTEXT` / `CLOB`

### Numerics (precision/width widening)

* Integer widening:

  * `TINYINT` → `SMALLINT`
  * `SMALLINT` → `INT`
  * `INT` → `BIGINT`
* `DECIMAL` / `NUMERIC` precision or scale increase
* Unparameterized → parameterized numeric types

  * e.g. `DECIMAL` → `DECIMAL(10,2)`
* Floating-point widening:

  * `FLOAT` → `DOUBLE`
  * `REAL` → `DOUBLE`
  * `FLOAT` ↔ `REAL` (treated as compatible)

### Temporals

* Same temporal type with parameter changes

  * e.g. `TIME(3)` → `TIME(6)`
* `DATE` → `TIMESTAMP` / `DATETIME` / `DATETIME2` / `DATETIMEOFFSET`
* Datetime-family widening:

  * `DATETIME` → `TIMESTAMP`
  * `DATETIME` → `DATETIME2`
  * `DATETIME` → `DATETIMEOFFSET`
  * `DATETIME2` → `DATETIMEOFFSET`

### Additional changes

* Refactors duplicated alter-column logic across multiple drivers
* Extends alter-path support for length-only changes
* Adds schema-builder tests covering widening and reducing varchar length changes
* Preserves existing rebuild behavior for drivers that require table recreation (e.g. SQLite family)


src/driver/postgres/PostgresQueryRunner.ts
```diff
@@ -1542,7 +1541,23 @@ export class PostgresQueryRunner
                     ),
                 )
             }
-
+            // Handle length changes without recreating the column (e.g., varchar(n) -> varchar(m))
+            if (newColumn.length !== oldColumn.length) {
```
sgarner: The block above this does exactly the same thing for precision and scale. It would be better to combine these cases.

test/github-issues/3357/issue-3357.ts
```diff
@@ -0,0 +1,78 @@
...
+import { expect } from "chai"
+import { DataSource } from "../../../src/data-source/DataSource"
+import {
+    createTestingConnections,
+    closeTestingConnections,
+} from "../../utils/test-utils"
+import { BugInitial } from "./entity/BugInitial"
+import { BugUpdated } from "./entity/BugUpdated"
+
+describe("github issues > #3357 (postgres) changing varchar length should ALTER, not DROP/ADD", () => {
```
sgarner: We are moving away from creating tests for specific GitHub issues.

I think the tests for this feature could be added to [test/functional/schema-builder/change-column.ts](https://github.com/typeorm/typeorm/blob/master/test/functional/schema-builder/change-column.ts) instead.

sgarner (changes requested):
Thanks for your contribution @BenraouaneSoufiane

Can you implement this for all the drivers (at least those which support it)? We won't want to merge this feature just for PostgreSQL.

BenraouaneSoufiane:
> Thanks for your contribution @BenraouaneSoufiane
> 
> Can you implement this for all the drivers (at least those which support it)? We won't want to merge this feature just for PostgreSQL.

@sgarner Thank you for your feedbacks, I've tried to fix them if you can take a look again

BenraouaneSoufiane:
Can someone review this:
@alumni
@gioboa
@naorpeled

src/driver/cockroachdb/CockroachQueryRunner.ts
```diff
@@ -1305,9 +1305,30 @@ export class CockroachQueryRunner
                 `Column "${oldTableColumnOrName}" was not found in the "${table.name}" table.`,
             )
 
+        // BEGIN: handle varchar length-only changes without recreating (CockroachDB)
+        if (
+            oldColumn.type === newColumn.type &&
+            oldColumn.length !== newColumn.length
+        ) {
```
sgarner: This block should be moved to the `else` condition of the following block. Otherwise, if the `length` is changed at the same time as `isArray`, `generatedType` or `asExpression` then we will unnecessarily `ALTER` the length before immediately dropping and recreating the column.

Look down at line 1584 where there is a block which changes precision and scale, the same as for PostgreSQL.

src/driver/mysql/MysqlQueryRunner.ts
```diff
@@ -1046,11 +1046,37 @@ export class MysqlQueryRunner extends BaseQueryRunner implements QueryRunner {
                 `Column "${oldColumnOrName}" was not found in the "${table.name}" table.`,
             )
 
+        if (
+            oldColumn.type === newColumn.type &&
+            oldColumn.length !== newColumn.length
+        ) {
```
sgarner: Same problem here, look down at line 1260

src/driver/oracle/OracleQueryRunner.ts
```diff
@@ -1075,11 +1075,27 @@ export class OracleQueryRunner extends BaseQueryRunner implements QueryRunner {
                 )} table.`,
             )
 
+        // BEGIN: handle varchar length-only changes without recreating (Oracle)
+        if (
+            oldColumn.type === newColumn.type &&
+            oldColumn.length !== newColumn.length
+        ) {
```
sgarner: Same problem here, look down at line 1322

src/driver/postgres/PostgresQueryRunner.ts
```diff
@@ -1523,9 +1522,12 @@ export class PostgresQueryRunner
                 oldColumn.name = newColumn.name
             }
 
+            // newColumn.length !== oldColumn.length: Handle length changes without recreating the column (e.g., varchar(n) -> varchar(m))
+
```
sgarner: If there's going to be a comment here, there should be no blank line and it should reflect all the cases covered:

```suggestion
            // Handle length, scale and precision changes without recreating the column, e.g.:
            // varchar(n) -> varchar(m)
            // numeric(x, y) -> numeric(v, w)
```

src/driver/sqlserver/SqlServerQueryRunner.ts
```diff
@@ -1255,11 +1255,37 @@ export class SqlServerQueryRunner
                 `Column "${oldTableColumnOrName}" was not found in the "${table.name}" table.`,
             )
 
+        // BEGIN: handle varchar length-only changes without recreating (SQL Server)
+        if (
+            oldColumn.type === newColumn.type &&
+            oldColumn.length !== newColumn.length
+        ) {
```
sgarner: Same problem here; look down at line 1623

test/functional/schema-builder/change-column.ts
```diff
@@ -9,6 +9,74 @@ import {
...
+@Entity({ name: "bug" })
+class BugUpdated {
+    @PrimaryGeneratedColumn()
+    id!: number
+
+    @Column({ type: "varchar", length: 51 })
+    example!: string
+}
+
+describe("schema builder > change varchar length in-place", () => {
```
sgarner: These tests are incorrectly written, the 'Bug' entities are not necessary, and you cannot hardcode the mysql datasource like this.

There is already a `describe` block in this file which performs test setup, it doesn't need to be duplicated. Please look at the existing tests in this file and follow their model.

test/github-issues/3357/issue-3357.ts
```diff
@@ -0,0 +1,78 @@
+import "reflect-metadata"
```
sgarner: Please delete all these files under `test/github-issues/3357`

sgarner (changes requested):

BenraouaneSoufiane:
Hello, can someone review this:
@sgarner
@alumni
@gioboa
@naorpeled

BenraouaneSoufiane:
@sgarner Any updates?

BenraouaneSoufiane:
@coderabbitai can you suggest active reviewers, not mentioned here?

test/functional/schema-builder/change-column.ts
```diff
@@ -22,6 +41,49 @@ describe("schema builder > change column", () => {
     beforeEach(() => reloadTestingDatabases(connections))
     after(() => closeTestingConnections(connections))
 
+    it("emits ALTER, not DROP/ADD when increasing varchar length", async () => {
+        for (const base of connections) {
+            // We choose postgres
+            if (base.options.type !== "postgres") continue
```
sgarner: The test needs to run on all the drivers where changing rather than dropping/adding the column is supported.

test/functional/schema-builder/change-column.ts
```diff
@@ -22,6 +41,49 @@ describe("schema builder > change column", () => {
...
+
+            // 1) initial schema (varchar(50)) + seed
+            const ds1 = new DataSource({
+                ...(base.options as any),
+                entities: [BugInitial],
+                synchronize: true,
+                dropSchema: false, // ensure we keep the table
+            } as any)
```
sgarner: This test is still not written correctly. Please read the other tests in this file and use them as a model for how to write a test for this feature.

test/functional/schema-builder/change-column.ts
```diff
@@ -22,6 +41,49 @@ describe("schema builder > change column", () => {
...
+                .join("\n")
+                .toUpperCase()
+
+            expect(up).to.match(/ALTER TABLE .* (ALTER COLUMN|CHANGE|MODIFY)/)
+            expect(up).to.not.match(/\bDROP COLUMN\b/)
+            expect(up).to.not.match(/\bADD COLUMN\b/)
+
+            await ds2.destroy()
+            try {
+                await ds1.dropDatabase()
```
sgarner: Definitely don't do this

package.json
```diff
@@ -141,8 +141,7 @@
     "mocha": "^10.8.2",
     "mongodb": "^6.15.0",
     "mssql": "^11.0.1",
-    "mysql": "^2.18.1",
-    "mysql2": "^3.14.0",
+    "mysql2": "^3.14.3",
```
sgarner: No changes to package.json or package-lock.json should be in this PR.

sgarner (changes requested):

BenraouaneSoufiane:
Thank you for your feedbacks @sgarner, I did some changes.

src/driver/mysql/MysqlQueryRunner.ts
```diff
@@ -1231,7 +1230,10 @@ export class MysqlQueryRunner extends BaseQueryRunner implements QueryRunner {
                 oldColumn.name = newColumn.name
             }
 
-            if (this.isColumnChanged(oldColumn, newColumn, true, true)) {
+            if (
+                this.isColumnChanged(oldColumn, newColumn, true, true) ||
```
sgarner: Is there a reason why we can't add the `length` check to the `isColumnChanged` method?

test/functional/schema-builder/change-column.ts
```diff
@@ -22,6 +22,42 @@ describe("schema builder > change column", () => {
     beforeEach(() => reloadTestingDatabases(connections))
     after(() => closeTestingConnections(connections))
 
+    it("uses ALTER COLUMN when increasing varchar length (postgres)", () =>
```
sgarner: Postgres shouldn't be mentioned here; this test should work for all databases supporting this feature.

It would be worth adding another test for reducing the column length as well (I suspect this might fail on some databases).

test/functional/schema-builder/change-column.ts
```diff
@@ -22,6 +22,42 @@ describe("schema builder > change column", () => {
...
+                // request length=51 and inspect the generated diff
+                nameCol.length = "51"
+                nameCol.build(connection)
+
+                const { upQueries } = await connection.driver
+                    .createSchemaBuilder()
+                    .log()
+                const up = upQueries.map((q) => q.query).join("\n")
+
+                // PG-specific assertion: ALTER TYPE, no drop/add
```
sgarner: Likewise here, this assertion shouldn't be PG-specific.

If the SQL looks different in some databases, you'll need to handle all the possibilities in the test with reference to `connection.driver.options.type`.

sgarner (changes requested):
Sorry, still not quite there yet!

In addition to code comments:

- I noticed that `AuroraMysqlQueryRunner` hasn't been updated, but it should support this feature the same as `MysqlQueryRunner`.
- I think `SapQueryRunner` can support this too.

BenraouaneSoufiane:
Hello @sgarner what's up, any updates on this?

BenraouaneSoufiane:
Hello again @sgarner @gioboa @alumni  can someone approve the workflows again, I forget to see two test fails/30 were passed, I just hit update branch

alumni:
Reducing the length is not possible in SAP HANA on purpose since it's a lossy operation (SAP software tend to prevent lossy operations, it's part of the company culture).

We usually do it like this (for text columns):
```sql
RENAME COLUMN "schema"."table"."column" TO "column_OLD";
ALTER TABLE "schema"."table" ADD "column" NVARCHAR(100) DEFAULT '' NOT NULL;
UPDATE "schema"."table" SET "column" = LEFT("column_OLD", 100);
ALTER TABLE "schema"."table" DROP "column_OLD";
```
The `LEFT`/`SUBARRAY` functions won't work for all data types (binary, vectors). In that case, `DROP` + `ADD` should be fine.

For SAP and MariaDB (possibly for Postgres too based on my understanding), vector columns with specified length are fixed-length, i.e. they must have data of that exact length, so increasing the length is also not possible with `ALTER COLUMN`.

I wonder what edge cases can be for other databases.

I would add a test (for all drivers) in which the column that is reduced contains data that is longer than the new size. We might find some more edge cases (not sure if databases trim the values).

I think we should keep the old behavior (`DROP` + `ADD`) for quite a lot of situations, for both increasing and decreasing the length :) Maybe we can use it where we know it works.

BenraouaneSoufiane:
@sgarner can approve workflows?

test/functional/schema-builder/change-column.ts
```diff
@@ -22,6 +22,155 @@ describe("schema builder > change column", () => {
     beforeEach(() => reloadTestingDatabases(connections))
     after(() => closeTestingConnections(connections))
 
+    it("uses ALTER COLUMN when increasing varchar length", () =>
+        Promise.all(
+            connections.map(async (connection) => {
+                if (connection.driver.options.type === "sap") return
```
sgarner: Do we really need to skip SAP here?

test/functional/schema-builder/change-column.ts
```diff
@@ -22,6 +22,155 @@ describe("schema builder > change column", () => {
...
+                // cleanup: restore metadata and schema
+                nameCol.length = originalLen
+                nameCol.build(connection)
+                await connection.synchronize()
+            }),
+        ))
+    it("uses ALTER COLUMN when reducing varchar length", () =>
+        Promise.all(
+            connections.map(async (connection) => {
+                if (connection.driver.options.type === "sap") return
```
sgarner: Same here. If SAP must drop/add the column then we can assert that in another `if` branch below rather than skipping it entirely.

src/query-runner/BaseQueryRunner.ts
```diff
@@ -536,7 +536,8 @@ export abstract class BaseQueryRunner {
             oldColumn.onUpdate !== newColumn.onUpdate || // MySQL only
             oldColumn.isNullable !== newColumn.isNullable ||
             (checkComment && oldColumn.comment !== newColumn.comment) ||
-            (checkEnum && this.isEnumChanged(oldColumn, newColumn))
+            (checkEnum && this.isEnumChanged(oldColumn, newColumn)) ||
+            oldColumn.length !== newColumn.length
```
sgarner: Minor: for readability it would be preferable to move this `length` condition line up a bit, so it's next to the related properties, precision/scale/width.

test/functional/query-runner/change-column.ts
```diff
@@ -24,10 +24,11 @@ describe("query runner > change column", () => {
     it("should correctly change column and revert change", () =>
         Promise.all(
             connections.map(async (connection) => {
-                // CockroachDB and Spanner does not allow changing primary columns and renaming constraints
+                // CockroachDB, Spanner & sap does not allow changing primary columns and renaming constraints
                 if (
                     connection.driver.options.type === "cockroachdb" ||
-                    connection.driver.options.type === "spanner"
+                    connection.driver.options.type === "spanner" ||
+                    connection.driver.options.type === "sap"
```
sgarner: Why was this added?

test/functional/schema-builder/change-column.ts
```diff
@@ -22,6 +22,155 @@ describe("schema builder > change column", () => {
...
+                    )
+                    expect(up).to.not.match(/\bDROP COLUMN\b/)
+                    expect(up).to.not.match(/\bADD COLUMN\b/)
+                } else {
+                    // Fallback: ensure we didn't drop/add the column
+                    expect(up).to.not.match(/\bDROP COLUMN\b/)
+                    expect(up).to.not.match(/\bADD COLUMN\b/)
+                }
+
+                // cleanup: restore metadata and schema
```
sgarner: These tests never actually execute the ALTER COLUMN, they only assert that the expected queries are generated. So we're not testing whether the generated ALTER COLUMN statements actually work on each database.

There needs to be a call to `synchronize()` with the new length, before cleanup reverts it.

test/functional/schema-builder/change-column.ts
```diff
@@ -22,6 +22,155 @@ describe("schema builder > change column", () => {
...
+                        /ALTER TABLE .* ALTER COLUMN .* SET DATA TYPE .*?\(\s*80\s*\)/i,
+                    )
+                    expect(up).to.not.match(/\bDROP COLUMN\b/)
+                    expect(up).to.not.match(/\bADD COLUMN\b/)
+                } else {
+                    expect(up).to.not.match(/\bDROP COLUMN\b/)
+                    expect(up).to.not.match(/\bADD COLUMN\b/)
+                }
+
+                // cleanup
```
sgarner: Likewise here, call `synchronize()` to test the ALTER statements actually execute.

As @alumni said, it would also be good to add some data to the table (I'm not sure if there is some already?) which exceeds the new `name` column length, so that we can verify what happens when data is truncated. Some database engines may reject this operation; in MySQL it's a server configuration option for example.

sgarner (changes requested):

src/query-builder/QueryBuilder.ts
```diff
@@ -25,6 +25,7 @@ import { ReturningType } from "../driver/Driver"
 import { OracleDriver } from "../driver/oracle/OracleDriver"
 import { InstanceChecker } from "../util/InstanceChecker"
 import { escapeRegExp } from "../util/escapeRegExp"
+import { ensureColumnMinLength } from "../util/ensureColumnMinLength"
```
BenraouaneSoufiane: ??? it's used in execute()

src/query-builder/UpdateQueryBuilder.ts
```diff
@@ -62,6 +63,33 @@ export class UpdateQueryBuilder<Entity extends ObjectLiteral>
...
+            await ensureColumnMinLength(
+                queryRunner,
+                tablePath,
+                col.databaseName,
+                v.length,
+            )
+        }
+        // -------------------------------
+        // END AUTO-WIDEN
+        // -------------------------------
```
BenraouaneSoufiane: I'm going to make it per driver/driver scope

src/query-builder/QueryBuilder.ts
```diff
@@ -507,6 +508,45 @@ export abstract class QueryBuilder<Entity extends ObjectLiteral> {
...
+                        col.databaseName,
+                        desired,
+                    )
+                }
+            }
+        }
+
+        // -------------------------------
+        // END AUTO-WIDEN
+        // -------------------------------
```
BenraouaneSoufiane: I'm going to make it per driver/driver scope

src/util/ensureColumnMinLength.ts
```diff
@@ -0,0 +1,23 @@
...
+    const oldCol = table.findColumnByName(columnName)
+    if (!oldCol) return
+
+    const currentLen =
+        oldCol.length != null ? parseInt(oldCol.length, 10) : undefined
+    if (currentLen == null) return // engines that don't report length (e.g., SQLite)
+    if (currentLen >= desiredLength) return // already large enough
+
+    const newCol = new TableColumn({ ...oldCol, length: String(desiredLength) })
+    await runner.changeColumn(tablePath, oldCol, newCol)
```
BenraouaneSoufiane: I'm going to undo this change (remove this helper at all)

src/query-builder/QueryBuilder.ts
BenraouaneSoufiane: I was think that was the ideal solution

src/driver/postgres/PostgresQueryRunner.ts
```diff
@@ -1523,6 +1522,67 @@ export class PostgresQueryRunner
...
+                                )}`,
+                        ),
+                    )
+                }
+
+                await this.executeQueries(upQueries, downQueries)
+                return
+            }
+            // END length-only fast path
+
```
BenraouaneSoufiane: can you provide commitable changes?

BenraouaneSoufiane: Thanks, well done!

src/driver/aurora-mysql/AuroraMysqlQueryRunner.ts
```diff
@@ -946,6 +945,66 @@ export class AuroraMysqlQueryRunner
...
+                            `ALTER TABLE ${this.escapePath(table)} ` +
+                                `MODIFY \`${col}\` ${this.driver.createFullType(
+                                    oldColumn,
+                                )} ${downNullability}`,
+                        ),
+                    )
+
+                    await this.executeQueries(upQueries, downQueries)
+                    return
+                }
```
BenraouaneSoufiane: @coderabbitai provide commitable changes by one click like others

src/driver/sqlserver/SqlServerQueryRunner.ts
```diff
@@ -1594,6 +1593,34 @@ export class SqlServerQueryRunner
...
+                    upQueries.push(
+                        new Query(
+                            `UPDATE ${this.escapePath(table)} ` +
+                                `SET ${col} = ${updateExpr} ` +
+                                `WHERE DATALENGTH(${col}) > ${threshold}`,
+                        ),
+                    )
+                }
+            }
+            // END
```
BenraouaneSoufiane: @coderabbitai can you check again if issue persists, if yes provide commitable changes by one click

src/driver/cockroachdb/CockroachQueryRunner.ts
```diff
@@ -1559,6 +1558,67 @@ export class CockroachQueryRunner
...
+                                    oldColumn,
+                                )}`,
+                        ),
+                    )
+                }
+
+                await this.executeQueries(upQueries, downQueries)
+                return
+            }
+            // END length-only fast path
```
BenraouaneSoufiane: @coderabbitai provide commitable changes

src/driver/aurora-mysql/AuroraMysqlQueryRunner.ts
```diff
@@ -946,6 +945,66 @@ export class AuroraMysqlQueryRunner
...
+            }
+            // END length-only fast path
```
BenraouaneSoufiane: @coderabbitai I already fixed this, can you check again, if no provide commitable changes by one click

src/driver/cockroachdb/CockroachQueryRunner.ts
```diff
@@ -1559,6 +1558,67 @@ export class CockroachQueryRunner
...
             }
+            // BEGIN length-only fast path (Postgres/Cockroach)
+
+            if (
+                oldColumn.type === newColumn.type &&
+                oldColumn.length !== newColumn.length &&
+                newColumn.isNullable === oldColumn.isNullable &&
+                newColumn.default === oldColumn.default &&
+                newColumn.comment === oldColumn.comment
+            ) {
```
BenraouaneSoufiane, line 1918: @coderabbitai I already fixed this, can you check again, if no provide commitable changes by one click

src/driver/cockroachdb/CockroachQueryRunner.ts
```diff
@@ -1559,6 +1558,67 @@ export class CockroachQueryRunner
...
+                    )
+                    downQueries.push(
+                        new Query(
+                            `ALTER TABLE ${this.escapePath(table)} ` +
+                                `ALTER COLUMN "${col}" TYPE ${this.driver.createFullType(
+                                    oldColumn,
+                                )} ` +
+                                `USING substring("${col}" FROM 1 FOR ${oldLen})`,
+                        ),
+                    )
```
BenraouaneSoufiane: @coderabbitai I already fixed this, can you check again, if no provide commitable changes by one click

src/driver/sap/SapQueryRunner.ts
```diff
@@ -1169,11 +1169,88 @@ export class SapQueryRunner extends BaseQueryRunner implements QueryRunner {
...
+
+            await this.executeQueries(upQueries, downQueries)
+            // refresh cache to reflect new length
+            const updatedCol = clonedTable.findColumnByName(col)
+            if (updatedCol) {
+                updatedCol.length = newColumn.length || ""
+            }
+            this.replaceCachedTable(table, clonedTable)
+            return
+        }
```
BenraouaneSoufiane: @coderabbitai I already fixed this, can you check again, if no provide commitable changes by one click

src/driver/cockroachdb/CockroachQueryRunner.ts
BenraouaneSoufiane: Thank you for your help!

src/driver/cockroachdb/CockroachQueryRunner.ts
BenraouaneSoufiane: Fixed, thank you!

src/driver/cockroachdb/CockroachQueryRunner.ts
```diff
@@ -1559,6 +1558,76 @@ export class CockroachQueryRunner
...
+                        new Query(
+                            `ALTER TABLE ${this.escapePath(table)} ` +
+                                `ALTER COLUMN "${colName}" TYPE ${this.driver.createFullType(
+                                    oldColumn,
+                                )} ` +
+                                `USING substring("${colName}" FROM 1 FOR ${
+                                    oldLen || 2147483647
+                                })`,
+                        ),
+                    )
```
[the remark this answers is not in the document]
BenraouaneSoufiane: from what you got isStringType?

BenraouaneSoufiane: @coderabbitai from what you got isStringType it rises type error (Cannot find name 'isStringType')?

src/driver/cockroachdb/CockroachQueryRunner.ts
```diff
@@ -1559,6 +1558,146 @@ export class CockroachQueryRunner
...
+                    downQueries.push(
+                        new Query(
+                            `ALTER TABLE ${this.escapePath(table)} ` +
+                                `ALTER COLUMN "${colName}" TYPE ${this.driver.createFullType(
+                                    oldColumn,
+                                )}`,
+                        ),
+                    )
+                }
```
BenraouaneSoufiane: Not correct, still, you can verify again

BenraouaneSoufiane:
Can someone approve the workflows??

pkuczynski:
> Hello @pkuczynski can you review this/approve the workflows

Unfortunately I do not have permissions.

BenraouaneSoufiane:
> > Hello @pkuczynski can you review this/approve the workflows
> 
> Unfortunately I do not have permissions.

Hhh, thinked you've, by the way thank you for your reply!

pkuczynski:
> Hhh, thinked you've, by the way thank you for your reply!

No, it would have to be one of the maintainers. I can review the code next week if nobody does this earlier. Can you in the meantime update PR description since you mentioned only Postgres but made changes to the other drivers too...

BenraouaneSoufiane:
> > Hhh, thinked you've, by the way thank you for your reply!
> 
> No, it would have to be one of the maintainers. I can review the code next week if nobody does this earlier. Can you in the meantime update PR description since you mentioned only Postgres but made changes to the other drivers too...

Yes, I'll do ASAP, thanks!

alumni:
@BenraouaneSoufiane If you want to check if your changes work, you can also go to the workflows in your own fork: https://github.com/BenraouaneSoufiane/typeorm/actions

In this case, if you get the build to pass, the tests will also run :)

We only run the PR workflows if we review the PR or before merging, but they are essentially the same :)

src/driver/mysql/MysqlQueryRunner.ts
```diff
@@ -1231,6 +1230,53 @@ export class MysqlQueryRunner extends BaseQueryRunner implements QueryRunner {
...
+                const down = `ALTER TABLE ${this.escapePath(table)} CHANGE \`${
+                    oldColumn.name
+                }\` ${this.buildCreateColumnSql(oldColDef, true)}`
+                upQueries.push(new Query(up))
+                downQueries.push(new Query(down))
+
+                await this.executeQueries(upQueries, downQueries)
+                return
+            }
+            // END length-only fast path
```
BenraouaneSoufiane: @coderabbitai provide commitable change

src/driver/sap/SapQueryRunner.ts
```diff
@@ -1169,11 +1169,224 @@ export class SapQueryRunner extends BaseQueryRunner implements QueryRunner {
...
+                    new Query(
+                        `ALTER TABLE ${this.escapePath(table)} ALTER (` +
+                            this.buildCreateColumnSql(
+                                downKeepLen,
+                                !(
+                                    newColumn.default === null ||
+                                    newColumn.default === undefined
+                                ),
+                                !newColumn.isNullable,
+                            ) +
```
BenraouaneSoufiane: @coderabbitai provide copy paste commitable changes

src/driver/sap/SapQueryRunner.ts
```diff
@@ -1169,11 +1169,213 @@ export class SapQueryRunner extends BaseQueryRunner implements QueryRunner {
...
+            [
+                "char",
+                "nchar",
+                "varchar",
+                "nvarchar",
+                "alphanum",
+                "shorttext",
+                "varbinary",
+                "binary",
+            ].includes(String(oldColumn.type).toLowerCase()) &&
```
BenraouaneSoufiane: No

test/functional/schema-builder/column/change/change-column/change-column.test.ts
```diff
@@ -22,6 +22,306 @@ describe("schema builder > change column", () => {
...
+                            /ALTER TABLE .* ALTER COLUMN \[name\] .*80/i,
+                        )
+                        expect(sqlBlob).to.not.match(/ADD COLUMN\s+\[name\]/i)
+                        expect(sqlBlob).to.not.match(/DROP COLUMN\s+\[name\]/i)
+                    } else if (
+                        driver === "sqlite" ||
+                        driver === "better-sqlite3"
+                    ) {
+                        // SQLite: no strict assertion here (TypeORM recreates the table). Keep functional test only.
+                    }
```
BenraouaneSoufiane, line 776: @coderabbitai can you provide copy paste commitable changes?

src/driver/sap/SapQueryRunner.ts
```diff
@@ -1169,11 +1169,296 @@ export class SapQueryRunner extends BaseQueryRunner implements QueryRunner {
...
+                    )
+                }
+
+                // 7) DROP temp column
+                downQueries.push(
+                    new Query(
+                        `ALTER TABLE ${this.escapePath(
+                            table,
+                        )} DROP ("${tmpDown}")`,
+                    ),
```
BenraouaneSoufiane: @coderabbitai you who provide that code, so how to fix? provide commitable changes

test/functional/schema-builder/column/change/change-column/change-column.test.ts
```diff
@@ -22,6 +22,479 @@ describe("schema builder > change column", () => {
...
+                        const usedModify =
+                            /ALTER TABLE .* (MODIFY|CHANGE) COLUMN `?name`? .*80/i.test(
+                                sqlBlob,
+                            )
+                        expect(
+                            usedModify,
+                            `Expected MODIFY/CHANGE COLUMN for 'name'.\n${sqlBlob}`,
+                        ).to.equal(true)
+                        expect(sqlBlob).to.not.match(/ADD COLUMN\s+`?name`?/i)
+                        expect(sqlBlob).to.not.match(/DROP COLUMN\s+`?name`?/i)
```
BenraouaneSoufiane: it matches, validated

test/functional/schema-builder/column/change/change-column/change-column.test.ts
```diff
@@ -22,6 +22,479 @@ describe("schema builder > change column", () => {
...
+                        const usedModify =
+                            /ALTER TABLE .* (MODIFY|CHANGE) COLUMN `?name`? .*40/i.test(
+                                sqlBlob,
+                            )
+                        expect(
+                            usedModify,
+                            `Expected MODIFY/CHANGE COLUMN for 'name'.\n${sqlBlob}`,
+                        ).to.equal(true)
+                        expect(sqlBlob).to.not.match(/ADD COLUMN\s+`?name`?/i)
+                        expect(sqlBlob).to.not.match(/DROP COLUMN\s+`?name`?/i)
```
BenraouaneSoufiane: it matches, validated

BenraouaneSoufiane:
Hello again, can someone approve ci workflows? it seems succeed in my fork: https://github.com/BenraouaneSoufiane/typeorm/actions/runs/18225890324

test/functional/schema-builder/column/change/change-column/change-column.test.ts
BenraouaneSoufiane: Submitted: https://x.com/benra1sofian/status/1974140356360106190

test/functional/schema-builder/change-column.ts
```diff
@@ -22,6 +22,483 @@ describe("schema builder > change column", () => {
...
+                            .getMetadata("post")
+                            .findColumnWithPropertyName("name")!
+                        nameColumnMetadata.length = originalLength
+                        nameColumnMetadata.build(connection)
+                        await connection.synchronize()
+                    } catch {}
+                    await queryRunner.release()
+                }
+            }),
+        ))
```
BenraouaneSoufiane: Because I committed your fixe twice it results in duplication, I will resolve the issue when I back to home

test/functional/schema-builder/change-column.ts
```diff
@@ -22,6 +22,483 @@ describe("schema builder > change column", () => {
...
+                    try {
+                        row = await repo.save(payload)
+                        insertedRowId = (row as any)?.id
+                    } catch (e) {
+                        insertErr = e
+                     } catch (e) {
+                         insertErr = e
+                     }
+                    }
+                    expect(insertErr).to.be.undefined
```
BenraouaneSoufiane: Because I committed your fixe twice it results in duplication, I will resolve the issue when I back to home

test/functional/schema-builder/change-column.ts
```diff
@@ -22,6 +22,483 @@ describe("schema builder > change column", () => {
...
+                    let widenErr: any
+                    try {
+                        await connection.synchronize()
+                    } catch (e) {
+                        widenErr = e
+                     } catch (e) {
+                         widenErr = e
+                     } finally {
+                    } finally {
+                        removeRecorder()
```
BenraouaneSoufiane: Because I committed your fixe twice it results in duplication, I will resolve the issue when I back to home

BenraouaneSoufiane:
Hello again @sgarner, the checks were passed, I think it's ready to go, no?

src/driver/postgres/PostgresQueryRunner.ts
```diff
@@ -1233,8 +1233,8 @@ export class PostgresQueryRunner
 
         if (
             oldColumn.type !== newColumn.type ||
-            oldColumn.length !== newColumn.length ||
             newColumn.isArray !== oldColumn.isArray ||
+            (oldColumn.isArray && oldColumn.length !== newColumn.length) || // Add this line
```
OSA413: What exactly is happening in this line? Like re-create (drop and add) an array type if its length is changed? Is this a thing in PG?

test/functional/query-runner/change-column.ts
```diff
@@ -24,7 +24,7 @@ describe("query runner > change column", () => {
     it("should correctly change column and revert change", () =>
         Promise.all(
             connections.map(async (connection) => {
-                // CockroachDB and Spanner does not allow changing primary columns and renaming constraints
+                // CockroachDB & Spanner does not allow changing primary columns and renaming constraints
```
OSA413: return "and" and change "does" to "do" (if you want)

src/driver/aurora-mysql/AuroraMysqlQueryRunner.ts
```diff
@@ -946,6 +945,87 @@ export class AuroraMysqlQueryRunner
                 oldColumn.name = newColumn.name
             }
 
+            // BEGIN length-only fast path (MySQL family)
+            // BEGIN length-only fast path (MySQL family)
```
OSA413: Probably a duplicate comment, remove

OSA413 (commented):
Unfortunately I need one more day to review this. (Thanks for patience and your contribution)

OSA413:
Can you also update the branch so it's up to date with master?

src/driver/postgres/PostgresQueryRunner.ts
BenraouaneSoufiane: Great catch, I forgot to remove early length check, the same thing with others

src/driver/aurora-mysql/AuroraMysqlQueryRunner.ts
BenraouaneSoufiane: Yes, it's.

OSA413:
After careful review, I can say that there's not enough tests for such "breaking" feature. I saw that you were trying to handle change of `bit` in MSSQL, the feature currently intended for varchar (nvarchar also probably?). I think that there should be added more tests to handle each handle of the type change. But first you need to see if there are tests that cover some of the changes.

Also:
* Don't leave console.log for debugging purposes, I see that you changed one test just to see if the changes are working. Does it mean there are not enough tests for now? Remove the console.log if not needed anymore.
* Use variable names with proper meaning and with full words (don't shorten the words in the variable names). It's hard reading and maintaining in the future what's going on in the tests and the drivers.
* Maybe it can be improved via some sort of abstraction, the driver files are already large enough.

src/driver/aurora-mysql/AuroraMysqlQueryRunner.ts
```diff
@@ -796,8 +800,8 @@ export class AuroraMysqlQueryRunner
         if (
             (newColumn.isGenerated !== oldColumn.isGenerated &&
                 newColumn.generationStrategy !== "uuid") ||
-            oldColumn.type !== newColumn.type ||
-            oldColumn.length !== newColumn.length ||
+            (oldColumn.type !== newColumn.type &&
+                !isSafeAlter(oldColumn, newColumn)) ||
             oldColumn.generatedType !== newColumn.generatedType
         ) {
```
BenraouaneSoufiane, line 1068: I've removed executequery from helper

src/driver/aurora-mysql/AuroraMysqlQueryRunnerHelper.ts
```diff
@@ -0,0 +1,227 @@
...
+        ;(cloned as any).charset =
+            (newColumn as any).charset ?? (cloned as any).charset
+        ;(cloned as any).collation =
+            (newColumn as any).collation ?? (cloned as any).collation
+        cloned.comment = newColumn.comment
+        cloned.enum = newColumn.enum
+        ;(cloned as any).unsigned = (newColumn as any).unsigned
+        ;(cloned as any).zerofill = (newColumn as any).zerofill
+        ;(cloned as any).width = (newColumn as any).width
+    }
```
BenraouaneSoufiane: I've added third parameter in the wrapper

src/driver/cockroachdb/CockroachQueryRunner.ts
```diff
@@ -1560,6 +1565,115 @@ export class CockroachQueryRunner
...
+                    clonedTable,
+                    oldColumn,
+                    newColumn,
+                    upQueries,
+                    downQueries,
+                    driver: this.driver as any,
+                    escapePath: this.escapePath.bind(this),
+                    Query,
+                })
+                // END length-only fast path
```
BenraouaneSoufiane: The same thing here, removed the executequery from helper

src/driver/cockroachdb/CockroachQueryRunner.ts
```diff
@@ -1306,9 +1311,9 @@ export class CockroachQueryRunner
...
         if (
-            oldColumn.type !== newColumn.type ||
-            oldColumn.length !== newColumn.length ||
-            newColumn.isArray !== oldColumn.isArray ||
+            (oldColumn.type !== newColumn.type &&
+                !isSafeAlter(oldColumn, newColumn)) ||
+            oldColumn.isArray !== newColumn.isArray ||
             oldColumn.generatedType !== newColumn.generatedType ||
             oldColumn.asExpression !== newColumn.asExpression
         ) {
```
BenraouaneSoufiane, line 1648: Already have the new guard

src/driver/mysql/MysqlQueryRunner.ts
```diff
@@ -28,6 +28,11 @@ import { IsolationLevel } from "../types/IsolationLevel"
 import { MetadataTableType } from "../types/MetadataTableType"
 import { ReplicationMode } from "../types/ReplicationMode"
 import { MysqlDriver } from "./MysqlDriver"
+import {
+    handleMysqlLengthOnlyFastPathChangeColumn,
+    handleSafeAlterMysql,
+} from "./MysqlQueryRunnerHelper"
+import { isSafeAlter } from "../../query-runner/BaseQueryRunnerHelper"
```
BenraouaneSoufiane: Removed executequeries from helper

src/driver/spanner/SpannerQueryRunner.ts
```diff
@@ -25,6 +25,11 @@ import { IsolationLevel } from "../types/IsolationLevel"
 import { MetadataTableType } from "../types/MetadataTableType"
 import { ReplicationMode } from "../types/ReplicationMode"
 import { SpannerDriver } from "./SpannerDriver"
+import {
+    handleSpannerLengthOnlyFastPath,
+    handleSafeAlterSpanner,
+} from "./SpannerQueryRunnerHelper"
+import { isSafeAlter } from "../../query-runner/BaseQueryRunnerHelper"
```
BenraouaneSoufiane: Removed executequeries from helper

src/driver/sap/SapQueryRunnerHelper.ts
```diff
@@ -0,0 +1,377 @@
...
+            )
+        }
+
+        // 7) DROP temp column
+        downQueries.push(
+            new Query(`ALTER TABLE ${escapePath(table)} DROP ("${tmpDown}")`),
+        )
+
+        const updatedCol2 = clonedTable?.findColumnByName?.(col)
+        if (updatedCol2) updatedCol2.length = newColumn.length || ""
```
BenraouaneSoufiane: Not relevant, I forgot to remove it after removing the executequeries

test/unit/base-query-runner-helper.ts
```diff
@@ -0,0 +1,336 @@
...
+import { isSafeAlter } from "../../src/query-runner/BaseQueryRunnerHelper"
+
+type PartialTableColumn = {
+    type?: string
+    length?: string | number
+    precision?: number | null
+    scale?: number | null
+}
+
+function col(partial: PartialTableColumn = {}) {
```
OSA413: It's better to use real type than a mock

test/unit/base-query-runner-helper.ts
```diff
@@ -0,0 +1,336 @@
...
+                        col({ type: "date" }) as any,
+                        col({ type: t }) as any,
+                    ),
+                ).to.equal(true, `date -> ${t}`)
+            }
+        })
+
+        it("allows DATETIME* -> TIMESTAMP/DATETIME*", () => {
+            const cases: [string, string][] = [
+                ["datetime", "timestamp"],
```
OSA413: This is a good case of parametrization of a test, can you please rewrite all your tests in this file to follow parametrization?

It's done as simple as using a for loop: https://stackoverflow.com/q/17569382/9245204

OSA413 (changes requested):

OSA413:
Also:
* There seems to be a lot of `any`, some of them are not needed (the type is perfectly derived by the compiler). Please, remove `any` where not needed.
* Is there a difference in the realization of assigning `oldLen` and `newLen` in `src\driver\oracle\OracleQueryRunnerHelper.ts` and `src\driver\postgres\PostgresQueryRunnerHelper.ts`? If so, add comments describing the differences (preferably from actual driver/TypeORM behaviour).
* Probably, you left a comment from your AI agent (the `your guard`), remove or improve the comment.
* The `*Driver*LengthOnlyFastPathArgs` seems to me the same across several drivers, can this be merged into one `Args` type with extending when necessary?

gioboa (commented):
May I ask you to solve the conflicts please?

gioboa (commented):
@BenraouaneSoufiane can you address this feedback and fix the failing tests please?

EthanBrooksAUS:
@gioboa the sqljs/better-sqlite not part of the fix (the actual ci fail), what to do? Does this pr will merged or there is something to do?

gioboa (commented):
I aligned the branch with the latest codebase. 💪

gioboa:
Thanks for the heads up. Let me add @alumni and @pkuczynski to double check

pkuczynski (changes requested):
Thanks for tackling this — data loss on column type changes (#3357) is a real and important issue. The scope is impressive, covering all non-SQLite drivers. A few things need addressing before this can be merged:

**1. `console.log` in production code**

`MysqlQueryRunner.ts` has three debug `console.log` calls left in `changeColumn`:
- `console.log(oldColumn, newColumn)`
- `console.log("it passed through drop/add")`
- `console.log("it passed through alter")`

These will fire on every migration/synchronize for MySQL/MariaDB users. Please remove them.

**2. `varchar ↔ nvarchar` incorrectly treated as safe widening**

`BaseQueryRunnerHelper.isSafeAlter` uses `isVarcharish = t.includes("varchar")`, which matches both `varchar` and `nvarchar`. This means `varchar(50) → nvarchar(100)` is approved as a safe widening, but these are different encodings (single-byte vs Unicode). On SQL Server this can corrupt data or fail if the column is indexed. Same issue with `char ↔ nchar`.

The fix is straightforward — check that both types share the same `n` prefix:

```typescript
const isUnicode = (t: string) => t.startsWith("n")
if (isVarcharish(oldType) && isVarcharish(newType) && isUnicode(oldType) === isUnicode(newType)) {
    return newLen >= oldLen
}
```

**3. CockroachDB `enable_experimental_alter_column_type_general` not set at runtime**

The tests manually run `SET enable_experimental_alter_column_type_general = true` before `synchronize()`, but the production helpers (`CockroachQueryRunnerHelper.ts`) never set this flag. Users running migrations will get failures from CockroachDB. Either prepend the SET to the upQueries, or document that users must enable it themselves.

**4. Double ALTER when type change + attribute change coincide (MySQL)**

In `MysqlQueryRunner.changeColumn`, `handleSafeAlterMysql` pushes a MODIFY COLUMN query, then the `isColumnChanged` block can push another CHANGE COLUMN for the same column (e.g., when nullability or default also changed). This results in two ALTERs where one would suffice. Consider having the safe-alter helper return a flag so the second block can be skipped when the first already handled the change.

**5. Tests don't verify data preservation**

This is the core claim of the PR — existing data survives the ALTER. But both key tests insert data *after* `synchronize()`, not before. A test that would actually prove correctness:

1. Create table with `varchar(50)`
2. Insert a row with a 45-character value
3. Change column to `varchar(80)`, run `synchronize()`
4. Assert the row still exists with the original value

Without this, a regression back to DROP/ADD would still pass the tests.

**6. Inline the helpers instead of separate files**

I understand the motivation for the `*QueryRunnerHelper.ts` files — the `changeColumn` methods are already long. However, 8 new files + a base helper is a lot of indirection for logic that's only called from one place in each QueryRunner. The names are also very generic (`MysqlQueryRunnerHelper`) which doesn't convey what they actually do (safe-alter handling).

We have a longer-term plan to extract the entire schema-builder / migration subsystem into a dedicated module (since this code only runs during migrations, never during normal application runtime). Separate helper files now will make that future extraction harder, because we'd be moving two tightly-coupled pieces instead of one.

For now, please inline the helper methods directly into their respective QueryRunner files — as private methods on the class. This keeps the logic close to its only call site and will make it easier to move the whole thing later when we restructure the migration subsystem.

**7. Minor: SQLite exclusion**

SQLite doesn't need safe-ALTER since it always recreates tables with data migration. This is correct, but worth a brief comment in `AbstractSqliteQueryRunner.changeColumn` so future contributors understand the asymmetry.

pkuczynski (commented):
Thanks for the continued effort on this — it's a genuinely important fix that many users are waiting for. The data-loss scenario in #3357 has been open since 2018, so getting this right matters.

I reviewed the current state against my previous feedback (March 15) and the earlier rounds from @sgarner and @OSA413. Unfortunately, none of the blocking items from any of the three reviewers appear to have been addressed yet. Here's a summary:

**Still open from my review:**
- `console.log` debug statements in `MysqlQueryRunner.ts` production code
- `varchar ↔ nvarchar` incorrectly treated as safe widening (`includes("varchar")` matches both)
- CockroachDB `enable_experimental_alter_column_type_general` set in `beforeMigration()` but never reset in `afterMigration()`
- Double ALTER when type change + attribute change coincide on MySQL
- Tests insert data after `synchronize()` instead of before — they don't verify data actually survives the migration
- Separate `*QueryRunnerHelper.ts` files should be inlined as private methods on each QueryRunner
- Missing comment explaining SQLite exclusion

**Additional issues found in this review:**
- `handleSafeAlterMysql` generates invalid SQL: `buildCreateColumnSql` already includes the column name, but it's prepended again with `MODIFY COLUMN` → duplicate column name
- `clonedTable` is not updated after safe alter succeeds → subsequent `synchronize()` detects false drift and re-applies the change
- `AuroraMysqlQueryRunner.changeColumn` calls `handleSafeAlterMysql()` but the class defines `handleSafeAlter()` → runtime TypeError
- `handleMysqlLengthOnlyFastPath` is defined in Aurora but never wired into `changeColumn`
- Rename + length change combined is not tested and likely broken
- `isFormalTypeChange` and `DriverCreateFullTypeLengthOnlyFastPathArgs` are exported but unused

**For reference:** PR #12227 by @marchantdev attempted the same fix with a simpler inline approach, correct data-preservation tests, and no `varchar/nvarchar` bug. It was closed without merging, but the patterns there (especially the test approach) are worth looking at.

I'd really like to see this feature land. If you're still interested in working on it, the key next steps would be:

1. Address the MySQL SQL generation bug (duplicate column name in MODIFY)
2. Fix the `varchar/nvarchar` matching in `isSafeAlter`
3. Inline the helper files as private methods
4. Rewrite tests to verify data preservation (insert → change column → synchronize → verify row survived)
5. Remove all `console.log` statements
6. Update `clonedTable` after safe alter to prevent false drift

If you're no longer able to work on this, please let us know and we can take it from here. No pressure either way — the contribution is appreciated regardless.

BenraouaneSoufiane:
@pkuczynski I've resolved the conflicts also resolved previous feedbacks, should you take a look if something persists/not resolved/not as expected??!

.gitignore
```diff
@@ -29,3 +29,6 @@ ormlogs.log
 ### Stryker ###
 reports/
 .stryker-tmp
+/reports
```
pkuczynski: These changes should not be part of this PR...

pkuczynski:
Hi @BenraouaneSoufiane, I've cleaned up this PR thread by removing old bot comments (CodeRabbit, Qodo, Coveralls, etc.) and earlier review request pings to make the conversation easier to follow. Hope you don't mind — there was a lot of noise accumulated over the months. The substantive review feedback from maintainers is still intact.

BenraouaneSoufiane:
@pkuczynski I'm working to fix tests, I've started with cockroachdb

pkuczynski:
> @pkuczynski I'm working to fix tests, I've started with cockroachdb

I am also pushing some changes so maybe you want to wait today?

BenraouaneSoufiane:
@pkuczynski why we can't collaborate? probably I can fix them all by today! btw tests not appear locally, what to do??!

image

BenraouaneSoufiane:
@pkuczynski I'm working on mariadb failed test now

pkuczynski:
> @pkuczynski why we can't collaborate? probably I can fix them all by today!

Sure we can. But it does not maw sense we work on the same thing at the same time. I will leave this all to you and check this PR next week. Also feel free to ping me on Discord. Will be faster than via comments.

BenraouaneSoufiane:
@pkuczynski I got another test issue, the previous fail was the same across mssql, mariadb, cockroach & postgres, now I got another fail, anyway will ping you when resolve them.

gioboa (approved):

pkuczynski:
Hey @BenraouaneSoufiane, looks like several people tried to tackle the same long-standing issue, namely in #11679, #11832, #11922, #12218, #12258, #12307, #12323, and #12345. We decided to move forward with your approach as it seems most complete, even despite it still needing some work. However could you have a quick look at the other PRs and check if there is anything more you could bring to yours from there?

BenraouaneSoufiane:
@pkuczynski didn't found something can add into this:
These PRs break into three buckets:

Fully or mostly superseded by my implementation
#11679
#11832
#12218
#12258
large parts of #12345

Partially superseded, but with one idea (it was coverred then removed as per @coderabbitai review recommendations I think)
#12307
Reason: broaden Postgres length fast-path to all withLengthColumnTypes
Meaningfully different, but probably not desirable as-is (the same issue here, already covered but @coderabbitai previous reviews say shouldn't use it I think)
#11922
#12323
Reason: they try to preserve more type changes by applying generic USING casts outside your current safety checks

alumni:
Decision from today's discussion ([meeting notes](https://github.com/typeorm/maintainers/blob/main/meetings/2026/2026-05-06.md)): "Focus on migration topics"

Need to review the PR into detail + release after 1.0 is done (as part of one of the first minor releases e.g. 1.1 or so).

BenraouaneSoufiane:
@alumni, I've reviewed the details, for this:  "release after 1.0 is done (as part of one of the first minor releases e.g. 1.1 or so)" is there something to do or not relevant for me?

Finesssee:
I took a small pass at two of the current review-backed edge cases and opened a helper PR against the active `fix-migration` branch here:

https://github.com/BenraouaneSoufiane/typeorm/pull/1

It keeps the scope intentionally small:

- Aurora MySQL: helper-handled safe type/length changes no longer return before `executeQueries`, and the generic `CHANGE` block is skipped when a helper already queued the DDL.
- Oracle: identity columns are routed through the existing drop/recreate path instead of falling through to generic `MODIFY`.

Local validation passed: Prettier check on touched files, ESLint `--quiet` on touched files, `pnpm run typecheck`, and `git diff --check`.

BenraouaneSoufiane:
@Cprakhar should you merge this? Team members are busy I think.

OSA413 (commented):
It looks great, one thing that I would improve is to throw error in the test when you have the try-catch-finally (currently there are places with empty or missing `catch` blocks which can silence actual errors in the tests).

Can you please improve it?

BenraouaneSoufiane:
> It looks great, one thing that I would improve is to throw error in the test when you have the try-catch-finally (currently there are places with empty or missing `catch` blocks which can silence actual errors in the tests).
> 
> Can you please improve it?

Yes, of course, will let you know once done.

BenraouaneSoufiane:
@OSA413, improved, thks!

OSA413:
Test `test\unit\base-query-runner-helper.ts` is not run because it's extension is `.ts`, not `.test.ts`. Can you please fix this?

BenraouaneSoufiane:
> Test `test\unit\base-query-runner-helper.ts` is not run because it's extension is `.ts`, not `.test.ts`. Can you please fix this?

Yes, of course, will let you know once fix it, thnks.

test/functional/schema-builder/column/change/change-column/change-column.test.ts
```diff
@@ -22,6 +87,1194 @@ describe("schema builder > change column", () => {
...
+                        removeRecorder()
+                    }
+                    expect(err).to.be.undefined
+
+                    const sqlBlob = recorded.join("\n")
+
+                    if (driver === "postgres" || driver === "cockroachdb") {
+                        expect(sqlBlob).to.match(
+                            /ALTER TABLE .* ALTER COLUMN "name" (SET DATA TYPE|TYPE) .*varchar/i,
+                        )
```
OSA413:
```suggestion
//THIS IS AN EXAMPLE, DO NOT APPLY THIS CHANGE
                            expect(sqlBlob.split("\n")[10]).to.eq("ALTER TABLE \"post\" ALTER COLUMN \"name\" TYPE character varying(80)");
```

I would prefer this type of assertion everywhere where the SQL changes the schema so that we know **exactly what and how much is changed**, I will discuss with alumni the points of improvements.

OSA413 (changes requested):
I'll review the tests of the PR.

You also should consider the Qodo's comments, some of them may be easy to fix. I would also not just truncate the strings when narrowing them (I will discuss this with alumni).

Meanwhile you can improve the tests by hard-coding the expected schema change against the actual result to see that we are doing it right (as I said, to see **exactly what and how much is changed**).

OSA413:
I also see that not all branches are covered in the coverage report, 100% coverage would be ideal (but better than a 100% code coverage is 100% case coverage, which I will discuss with alumni)

marchantdev:
Following up on OSA413's branch coverage note — I went through `BaseQueryRunnerHelper.ts` and mapped every uncovered branch. Here are the additions that should bring `isSafeAlter` and `normalizeColumnLength` to full branch coverage.

The current test file doesn't import `normalizeColumnLength` at all, so that exported function has 0% coverage. I've included tests for it at the bottom.

<details>
<summary>Additional tests to add to <code>test/unit/base-query-runner-helper.ts</code></summary>

```typescript
import { isSafeAlter, normalizeColumnLength } from "../../src/query-runner/BaseQueryRunnerHelper"

describe("BaseQueryRunnerHelper.isSafeAlter – branch coverage", () => {
    // Lines 184-187: special families always block within-family alters
    describe("special families block same-family alters", () => {
        it("blocks BOOLEAN -> BOOLEAN", () => {
            expect(isSafeAlter(col({ type: "boolean" }), col({ type: "boolean" }))).to.equal(false)
        })
        it("blocks BIT -> BOOL", () => {
            expect(isSafeAlter(col({ type: "bit" }), col({ type: "bool" }))).to.equal(false)
        })
        it("blocks ENUM -> ENUM", () => {
            expect(isSafeAlter(col({ type: "enum" }), col({ type: "enum" }))).to.equal(false)
        })
        it("blocks SET -> SET", () => {
            expect(isSafeAlter(col({ type: "set" }), col({ type: "set" }))).to.equal(false)
        })
        it("blocks UUID -> UUID", () => {
            expect(isSafeAlter(col({ type: "uuid" }), col({ type: "uuid" }))).to.equal(false)
        })
        it("blocks BINARY -> BINARY", () => {
            expect(isSafeAlter(col({ type: "binary" }), col({ type: "binary" }))).to.equal(false)
        })
        it("blocks BYTEA -> BYTEA", () => {
            expect(isSafeAlter(col({ type: "bytea" }), col({ type: "bytea" }))).to.equal(false)
        })
    })

    // Lines 205, 213, 221: unicode encoding mismatch blocks within-family alters
    describe("unicode encoding mismatch blocks string alter", () => {
        it("blocks NCHAR(N) -> VARCHAR(M): unicode mismatch in char->varchar path", () => {
            expect(isSafeAlter(col({ type: "nchar(10)" }), col({ type: "varchar(20)" }))).to.equal(false)
        })
        it("blocks CHAR(N) -> NVARCHAR(M): unicode mismatch in char->varchar path", () => {
            expect(isSafeAlter(col({ type: "char(10)" }), col({ type: "nvarchar(20)" }))).to.equal(false)
        })
        it("blocks NCHAR(N) -> CHAR(M): unicode mismatch in char->char path", () => {
            expect(isSafeAlter(col({ type: "nchar(10)" }), col({ type: "char(20)" }))).to.equal(false)
        })
        it("blocks VARCHAR(N) -> NVARCHAR(M): unicode mismatch in varchar->varchar path", () => {
            expect(isSafeAlter(col({ type: "varchar(100)" }), col({ type: "nvarchar(100)" }))).to.equal(false)
        })
    })

    // Lines 206, 222: unknown lengths assume widening (early return true)
    describe("unknown string lengths assume widening", () => {
        it("allows CHAR -> VARCHAR when both lengths unknown", () => {
            expect(isSafeAlter(col({ type: "char" }), col({ type: "varchar" }))).to.equal(true)
        })
        it("allows NCHAR -> NVARCHAR when new length unknown", () => {
            expect(isSafeAlter(col({ type: "nchar", length: 10 }), col({ type: "nvarchar" }))).to.equal(true)
        })
        it("allows CHAR(N) -> VARCHAR when new length unknown", () => {
            expect(isSafeAlter(col({ type: "char(10)" }), col({ type: "varchar" }))).to.equal(true)
        })
    })

    // Line 229: STR family fallthrough (non-char/varchar transitions)
    describe("string family fallthrough blocks unsupported transitions", () => {
        it("blocks JSON -> XML (both STR family, no handler matched)", () => {
            expect(isSafeAlter(col({ type: "json" }), col({ type: "xml" }))).to.equal(false)
        })
        it("blocks NTEXT -> CLOB (text-likes but not varchar->textLike direction)", () => {
            expect(isSafeAlter(col({ type: "ntext" }), col({ type: "clob" }))).to.equal(false)
        })
        it("blocks TEXT -> VARCHAR(N) (reverse of allowed varchar->text)", () => {
            expect(isSafeAlter(col({ type: "text" }), col({ type: "varchar(255)" }))).to.equal(false)
        })
    })

    // Lines 260-261: decimal precision/scale widening
    describe("decimal precision and scale widening", () => {
        it("allows strictly larger precision: DECIMAL(8,2) -> DECIMAL(12,4)", () => {
            expect(isSafeAlter(
                col({ type: "decimal(8,2)", precision: 8, scale: 2 }),
                col({ type: "decimal(12,4)", precision: 12, scale: 4 }),
            )).to.equal(true)
        })
        it("allows same precision with more scale: DECIMAL(10,2) -> DECIMAL(10,4)", () => {
            expect(isSafeAlter(
                col({ type: "decimal(10,2)", precision: 10, scale: 2 }),
                col({ type: "decimal(10,4)", precision: 10, scale: 4 }),
            )).to.equal(true)
        })
        it("blocks same precision with less scale: DECIMAL(10,4) -> DECIMAL(10,2)", () => {
            expect(isSafeAlter(
                col({ type: "decimal(10,4)", precision: 10, scale: 4 }),
                col({ type: "decimal(10,2)", precision: 10, scale: 2 }),
            )).to.equal(false)
        })
    })

    // Line 282: numeric family fallthrough (mixed int/float/decimal)
    describe("numeric cross-type fallthrough", () => {
        it("blocks FLOAT -> DECIMAL: no widening path between float and decimal families", () => {
            expect(isSafeAlter(col({ type: "float" }), col({ type: "decimal(10,2)" }))).to.equal(false)
        })
        it("blocks INT -> FLOAT: no widening path between integer and float families", () => {
            expect(isSafeAlter(col({ type: "int" }), col({ type: "float" }))).to.equal(false)
        })
    })

    // Line 300: datetime* -> timestamp widening (untested direction)
    describe("datetime* -> timestamp widening", () => {
        it("allows DATETIME -> TIMESTAMP", () => {
            expect(isSafeAlter(col({ type: "datetime" }), col({ type: "timestamp" }))).to.equal(true)
        })
        it("allows DATE -> DATETIME2", () => {
            expect(isSafeAlter(col({ type: "date" }), col({ type: "datetime2" }))).to.equal(true)
        })
        it("allows DATE -> DATETIMEOFFSET", () => {
            expect(isSafeAlter(col({ type: "date" }), col({ type: "datetimeoffset" }))).to.equal(true)
        })
    })

    // Line 306: temporal fallthrough (narrowing / unsupported)
    describe("temporal narrowing blocks alter", () => {
        it("blocks TIMESTAMP -> DATE (narrowing)", () => {
            expect(isSafeAlter(col({ type: "timestamp" }), col({ type: "date" }))).to.equal(false)
        })
        it("blocks TIMESTAMP -> TIME (narrowing)", () => {
            expect(isSafeAlter(col({ type: "timestamp" }), col({ type: "time" }))).to.equal(false)
        })
        it("blocks TIMESTAMPTZ -> TIMESTAMP (neither date nor datetime* path)", () => {
            expect(isSafeAlter(col({ type: "timestamptz" }), col({ type: "timestamp" }))).to.equal(false)
        })
        it("blocks DATETIME -> DATE (narrowing)", () => {
            expect(isSafeAlter(col({ type: "datetime" }), col({ type: "date" }))).to.equal(false)
        })
    })
})

// normalizeColumnLength has 0% coverage — it is exported but not imported in the test file
describe("BaseQueryRunnerHelper.normalizeColumnLength", () => {
    it("returns undefined for null", () => {
        expect(normalizeColumnLength(null)).to.equal(undefined)
    })
    it("returns undefined for undefined", () => {
        expect(normalizeColumnLength(undefined)).to.equal(undefined)
    })
    it("returns undefined for empty string", () => {
        expect(normalizeColumnLength("")).to.equal(undefined)
    })
    it("returns the number for a numeric input", () => {
        expect(normalizeColumnLength(42)).to.equal(42)
    })
    it("parses a numeric string", () => {
        expect(normalizeColumnLength("100")).to.equal(100)
    })
    it("returns undefined for a non-numeric string", () => {
        expect(normalizeColumnLength("abc")).to.equal(undefined)
    })
    it("returns 0 for '0' (falsy but finite)", () => {
        expect(normalizeColumnLength("0")).to.equal(0)
    })
    it("truncates a float string to integer (parseInt behaviour)", () => {
        expect(normalizeColumnLength("10.9")).to.equal(10)
    })
    it("returns undefined for Infinity (not finite)", () => {
        expect(normalizeColumnLength(Infinity)).to.equal(undefined)
    })
})
```

</details>

**Summary of gaps covered above:**
- Lines 184–187: BOOL/ENUM/UUID/BIN same-family guards (7 tests)
- Lines 205/213/221: unicode encoding mismatch in all three char/varchar sub-paths (4 tests)
- Lines 206/222: undefined-length early returns (3 tests)
- Line 229: STR family fallthrough when no char/varchar handler matches (3 tests)
- Lines 260–261: decimal precision-only widening and scale widening (3 tests)
- Line 282: numeric cross-family fallthrough (2 tests)
- Line 300: `datetime* → timestamp` direction (3 tests)
- Line 306: temporal narrowing fallthrough (4 tests)
- `normalizeColumnLength`: 0% → full branch coverage (9 tests)

BenraouaneSoufiane:
@OSA413 Qodos comments already addressed.

BenraouaneSoufiane:
CI has been fixed (except which related with sonar scanner), should some one review this?

michaelbromley:

```
git fetch origin
```

Issue #8799, opened 2022-03-26
Separate Connection limit for slave and master


sagar-m-s:
## Feature Description
Add a new connection variable to separate the connectionLimit for master and slave

### The Problem
Currently the slave and master should have same number of connections. 
If we want to have a write intensive application we need to make sure the read is also supporting the same amount of connections.
For RDS unless we modify the number of connections or increase the read replica it is difficult.
If we can separate the Connection limit for slave and master it would be great.
For read replica if we want to increase the connection limit we must replicate the connection details multiple times since for RDS it is pointing to the load balancer instead of host

### The Solution
If we can separate the Connection limit for slave and master to different variable in both case we can give the connection limit to what we want and solve the problem for mysql

### Considered Alternatives

For slave we can replicate the connection details to suffix our use case but for master we need to make sure the slave is capable of supporting same number of connection as master

### Relevant Database Driver(s)

| DB Type           | Relevant      |
|-------------------| ---           |
| `aurora-mysql`    | yes            |
| `aurora-postgres` | no            |
| `better-sqlite3`  | no            |
| `cockroachdb`     | no            |
| `cordova`         | no            |
| `expo`            | no            |
| `mongodb`         | no            |
| `mysql`           | yes            |
| `nativescript`    | no            |
| `oracle`          | no            |
| `postgres`        | no            |
| `react-native`    | no            |
| `sap`             | no            |
| `sqlite`          | no            |
| `sqlite-abstract` | no            |
| `sqljs`           | no            |
| `sqlserver`       | no            |

### Are you willing to resolve this issue by submitting a Pull Request?
✅

PR #11626, opened 2025-08-27
feat(mysql): add pool size options for each connection

ChoSeoHwan:
### Description of change

I'm not very good at English, so I used a translator. Thank you for your understanding.

Currently, the `MysqlConnectionOptions` poolSize option specifies the connection limit for all replica connections.

In this case, you cannot specify a connection limit for each master or slave.

For example, if you only want to create one master connection and up to 10 slave connections, Typeorm has no solution.

I added the `MysqlConnectionCredentialsOptions` poolSize option, allowing you to specify a poolSize for each replica.

#### Verification Method
If the master's poolSize is 1 and the `slaves[0] poolSize is 2, 
then the master connection should be open at most 1, and the slaves[0] connection should be open at most 2.

### Pull-Request Checklist

-   [o ] Code is up-to-date with the `master` branch

### Related Issues

https://github.com/typeorm/typeorm/issues/8799

### DIFF
--- a/src/driver/mysql/MysqlConnectionCredentialsOptions.ts
+++ b/src/driver/mysql/MysqlConnectionCredentialsOptions.ts
@@ -43,4 +43,10 @@ export interface MysqlConnectionCredentialsOptions {
      * Database socket path
      */
     readonly socketPath?: string
+
+    /**
+     * Maximum number of clients the pool should contain.
+     * for each connection
+     */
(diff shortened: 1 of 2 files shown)


gioboa (changes requested):
May I ask you to add those tests please?
Otherwise I will close this PR and I will implement the feature, you will be co author there btw.

PR #11634, opened 2025-09-02
fix(query-builder): handle empty result set when offset exceeds total

### LINKED ISSUE #11640: getManyAndCount() returns incorrect count value for out-of-bounds skip
### Issue description

getManyAndCount() returns incorrect count value for out-of-bounds skip

### Expected Behavior

The following test case illustrates the regression introduced by #11524 :

```ts
// test/other-issues/lazy-count/lazy-count.ts

    it("should execute count query when skip is out-of-bounds", () =>
        Promise.all(
            connections.map(async function (connection) {
                await savePostEntities(connection, 5)

                const afterQuery = connection
                    .subscribers[0] as AfterQuerySubscriber
                afterQuery.clear()

                const [entities, count] = await connection.manager
                    .take(10)
                    .skip(10) // Skip beyond the total number of records
                    .orderBy("post.id")
                    .getManyAndCount()

                expect(count).to.be.equal(5)
                expect(entities.length).to.be.equal(0)

                // The count query MUST be executed when skip is out-of-bounds
                expect(
                    afterQuery
                        .getCalledQueries()
                        .filter((query) => query.match(/(count|cnt)/i)),
                ).not.to.be.empty
            }),
        ))
```

### Actual Behavior

The count query should be executed when `skip` is out-of-bounds so that the correct `count` value is returned.

### Steps to reproduce

See unit test case

### My Environment

| Operating System    |    OSX Sequoia 15.5      |
| Node.js version     | 20.19.1  |

### Additional Context

_No response_

- [ ] aurora-mysql
- [ ] aurora-postgres
- [ ] better-sqlite3
- [ ] cockroachdb
- [ ] cordova
- [ ] expo
- [ ] mongodb
- [ ] mysql
- [ ] nativescript
- [ ] oracle
- [ ] postgres
- [ ] react-native
- [ ] sap
- [ ] spanner
- [ ] sqlite
- [ ] sqlite-abstract
- [ ] sqljs
- [ ] sqlserver

jeremyteyssedre:
Fixes #11640

### Description of change
This PR fixes a bug in lazy count optimization of geManyAndCount. (Introduced in #11524)

Currently, when using limit/offset, the count query can be skipped if the total can be deduced from the number of returned rows. However, there was an incorrect behavior when the offset value was greater than the total number of entities.

Before: lazyCount returned an incorrect total (equal to offset + 0).
Now: lazyCount correctly returns undefined in this case, forcing a proper COUNT(*) query to be executed.

### Pull-Request Checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Ensures total counts are correctly reported for paginated queries when offset/skip exceeds available items, avoiding missing or undefined totals for empty result sets.
* **Tests**
  * Added tests to confirm a separate count query runs in offset/skip scenarios, validating accurate totals even when no records are returned.

### DIFF
--- a/src/query-builder/SelectQueryBuilder.ts
+++ b/src/query-builder/SelectQueryBuilder.ts
@@ -1939,6 +1939,10 @@ export class SelectQueryBuilder<Entity extends ObjectLiteral>
             return undefined
         }
 
+        if (entitiesAndRaw.entities.length === 0) {
+            return undefined;
+        }
+
         const hasSkip =
             this.expressionMap.skip !== undefined &&
             this.expressionMap.skip !== null &&
--- a/src/query-builder/SelectQueryBuilder.ts
+++ b/src/query-builder/SelectQueryBuilder.ts
@@ -1939,10 +1939,6 @@ export class SelectQueryBuilder<Entity extends ObjectLiteral>
             return undefined
         }
 
-        if (entitiesAndRaw.entities.length === 0) {
-            return undefined;
-        }
-
         const hasSkip =
             this.expressionMap.skip !== undefined &&
             this.expressionMap.skip !== null &&
@@ -1951,6 +1947,12 @@ export class SelectQueryBuilder<Entity extends ObjectLiteral>
             this.expressionMap.offset !== undefined &&
             this.expressionMap.offset !== null &&
             this.expressionMap.offset > 0
+
+        if (entitiesAndRaw.entities.length === 0 && (hasSkip || hasOffset)) {
+            // when skip or offset were used and no results found, we need to execute a full count
+            // (the given offset may have exceeded the actual number of rows)
+            return undefined
+        }
 
         // offset overrides skip when no join is defined
         const previousResults: number = hasOffset
--- a/test/other-issues/lazy-count/lazy-count.ts
+++ b/test/other-issues/lazy-count/lazy-count.ts
@@ -266,6 +266,34 @@ describe("other issues > lazy count", () => {
             }),
         ))
 
+    it("run count query when joining a relation with skip is greater than total entities", () =>
+        Promise.all(
+            connections.map(async function (connection) {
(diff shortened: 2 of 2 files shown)


src/query-builder/SelectQueryBuilder.ts
```diff
@@ -1939,6 +1939,10 @@ export class SelectQueryBuilder<Entity extends ObjectLiteral>
             return undefined
         }
 
+        if (entitiesAndRaw.entities.length === 0) {
+            return undefined;
+        }
```
sgarner: In cases where skip/offset were not used, the lazy count can still be trusted when the result set is empty. So let's move this down and condition it on that:


```ts
        const hasSkip =
this.expressionMap.skip !== undefined &&
this.expressionMap.skip !== null &&
this.expressionMap.skip > 0
        const hasOffset =
this.expressionMap.offset !== undefined &&
this.expressionMap.offset !== null &&
this.expressionMap.offset > 0

        if (entitiesAndRaw.entities.length === 0 && (hasSkip || hasOffset)) {
            // when skip or offset were used and no results found, we need to execute a full count
            // (the given offset may have exceeded the actual number of rows)
            return undefined
        }
```

test/other-issues/lazy-count/lazy-count.ts
```diff
@@ -266,6 +266,33 @@ describe("other issues > lazy count", () => {
             }),
         ))
 
+    it("run count query when offset is greater than total entities", () =>
```
sgarner, line 297: Could we have another test which verifies this works with `skip` and joined relations as well?

sgarner (changes requested):
Good catch @jeremyteyssedre and thanks for the PR 💜

sgarner (approved):
LGTM!

Thanks for your contribution @jeremyteyssedre 💜

alumni (approved):

PR #11639, opened 2025-09-04
chore: bump sha.js from 2.4.11 to 2.4.12 (fix security issue: CVE-2025-9288)

prateek-hegde:
### Summary
Bumps `sha.js` from 2.4.11 to 2.4.12 to address [security advisory CVE: 2025-9288](https://www.cve.org/CVERecord?id=CVE-2025-9288).

### Details
- `sha.js@2.4.11` contains a known vulnerability (CVE: 2025-9288).
- The issue has been fixed in `sha.js@2.4.12`.
- Updated hash() to only truncate when options.length > 0

### Testing
- Installed updated dependency.
- Ran full TypeORM test suite → all tests passed.

### Notes
This is a safe, non-breaking patch dependency update.

### Pull-Request Checklist

-   [ ] This pull request links relevant issues as `Fixes #00000` - N/A
-   [ ] Documentation has been updated to reflect this change - N/A


prateek-hegde:
Hey @alumni  @sgarner  🚨  
This PR is on a mission to save TypeORM from a tiny but mighty vulnerability [CVE-2025-9288](https://www.cve.org/CVERecord?id=CVE-2025-9288)🛡️.  

Can I count on you both to give it the final ✅ so we can merge and keep things secure? 🙏 
(Think of it as patching a leak before it becomes a flood 🌊.)

src/util/StringUtils.ts
```diff
@@ -120,8 +120,12 @@ export function hash(input: string, options: IHashOptions = {}): string {
...
     hashFunction.update(input, "utf8")
     const hashedInput = hashFunction.digest("hex")
-    if (options.length) {
-        return hashedInput.slice(0, options.length)
+    const len = Number.isFinite(options.length as number)
+        ? Math.trunc(options.length as number)
+        : 0
+    if (len > 0) {
+        // slice already caps to hashedInput.length; the Math.min keeps intent explicit
+        return hashedInput.slice(0, Math.min(len, hashedInput.length))
```
sgarner: Do we really need this change? The extra validation is just adding unnecessary complexity IMO. 🤔 

This `hash` function is only ever used internally by drivers, which pass hardcoded `maxAliasLength` values. We never pass any user input to it, so there is no vector for exploiting this CVE in TypeORM that I can see.

`String.slice()` already handles `Infinity` and float values without trouble anyway, so the only part adding any safety here is the `len > 0` check against negative numbers. The rest I think can be reverted.

test/unit/util/string-utils.ts
```diff
@@ -119,4 +119,66 @@ describe("StringUtils", () => {
             expect(actual).to.be.equal(expected, `Failed for Input: ${input}`)
         })
     })
+
+    describe("hash", () => {
```
sgarner, line 123: I do appreciate the addition of this comprehensive new test, thanks!

src/util/StringUtils.ts
prateek-hegde: @sgarner thank you for the input. I will make the necessary changes

diogowbrito:
@michaelbromley Do you have an expected timeline for when a new release including this vulnerability fix will be published? Thanks!

naiyarJamalNagarro:
@michaelbromley  getting some critical error to update sha.js  to version 2.4.12 or later, and for this pr is already merged (2 days ago) but latest version released 3 week ago (0.3.26). so can we get any patch update to include sha.js PR or any timeline.

alumni:
Sorry for the slow reply, I'm having some days off :)

TypeORM does not force you to use a version with a vulnerability, so no immediate release is necessary (neither was the update in this repository, i.e. this PR, it's just nice to have).

BTW, shajs will be removed in the next major version.

sgarner:
Indeed, TypeORM depends on the range `^2.4.11` which means you can update to a newer version locally just by doing:

```bash
npm update sha.js@latest
```

Considering the limited way TypeORM uses sha.js, there is not really any feasible vector to exploit CVE-2025-9288 through it anyway.

But this fix has been merged and will be coming in a release soon.

PR #11644, opened 2025-09-08
fix: handle undefined values in where clauses properly

afehid:
When using findOne or findOneBy with undefined values in where clauses, TypeORM would incorrectly return the first row instead of null.

Modified buildWhere method to return '1 = 0' condition when all where clause values are undefined/null, ensuring no results are returned.

- Fix undefined behavior in SelectQueryBuilder.buildWhere()
- Add comprehensive tests for undefined value handling
- Maintain full backwards compatibility

Closes issue with undefined values returning first row instead of null.

### Pull-Request Checklist


afehid:
It's okay, i was trying to help because i still facing this issue in my project

PR #11658, opened 2025-09-18
fix: ensure hash aliases start with letters for valid SQL identifiers

### LINKED ISSUE #11657: Generated hash aliases starting with digits cause SQL syntax errors in raw queries
### Issue description

When TypeORM generates hash-based aliases, the resulting alias can start with a digit (0-9). This creates invalid SQL identifiers that cause syntax errors when used in raw queries.

### Expected Behavior

Generated aliases should always start with a letter to create valid SQL identifiers that can be safely used in all SQL contexts.

### Actual Behavior

raw queries fail with SQL syntax errors when auto-generated aliases start with digits:
`QueryFailedError: trailing junk after numeric literal at or near "[alias_starting_with_digit]"`

### Steps to reproduce

1. Use nested relations that generate long internal aliases
2. Apply a `Raw()` filter that references the generated alias
3. Error occurs when the generated hash starts with a digit

### My Environment

| Operating System    | Windows |
| Node.js version     | v20.10.0 |

### Additional Context

This issue occurs with nested relations where TypeORM internally generates aliases for deep object property paths. The problem is deterministic - certain relation paths will consistently generate problematic aliases while others work fine.

- [ ] aurora-mysql
- [ ] aurora-postgres
- [ ] better-sqlite3
- [ ] cockroachdb
- [ ] cordova
- [ ] expo
- [ ] mongodb
- [ ] mysql
- [ ] nativescript
- [ ] oracle
- [x] postgres
- [ ] react-native
- [ ] sap
- [ ] spanner
- [ ] sqlite
- [ ] sqlite-abstract
- [ ] sqljs
- [ ] sqlserver

LeviHeber:
Fixes: #11657
### Description of change
**What the change is intended to do:**
This change ensures that hash-generated aliases always start with a letter, creating valid SQL identifiers. When TypeORM generates aliases using SHA1 hashing (for long table/column names or nested relations), the resulting hash can start with a digit (0-9), which creates invalid SQL identifiers.

**Why this change is needed:**
SQL identifiers that start with digits cause syntax errors in raw queries, particularly with nested relations. Users encounter `QueryFailedError: trailing junk after numeric literal at or near "[alias]"` when the generated hash starts with a digit.

**How I've verified it:**
1. Added a unit test that reproduces the issue with nested relations and Raw queries
2. Verified the fix maps digits 0-9 to letters g-p (avoiding collision with natural hex letters a-f)
3. Confirmed no collision risk: SHA1 hex output only contains characters 0-9 and a-f, so mapping digits to g-p ensures no conflicts with naturally generated hashes
4. Confirmed that hash uniqueness and distribution are maintained
5. Tested that the solution works across different database types

**New behavior:**
Hash aliases always start with letters, ensuring they are valid SQL identifiers that can be safely used in all SQL contexts, with zero risk of collision with other hash outputs.

### Pull-Request Checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed SQL syntax errors that could occur when database alias names start with digits. The system now automatically converts leading digits to valid characters, ensuring proper query execution.


test/github-issues/11657/issue-11657.ts
```diff
@@ -0,0 +1,51 @@
+import "reflect-metadata"
```
gioboa: We aren't creating tests for specific GitHub issues any more. Instead, tests should be added to the suites for the related functionality.

Can you please move these tests to test/functional/driver/driver-utils.ts?

Thanks

LeviHeber:
I moved the tests to the new location.

LeviHeber:
I would appreciate guidance from your experience on how to fix the test.
This seems like a problem with the testing process itself in certain environments, rather than the fix.

LeviHeber:
Is there anything else that needs to be done to move forward?

LeviHeber:
Is there anything else that needs to be done to move forward?

gioboa:
We need to fix the tests, it's impossible to merge this PR if tests are failing 👀

LeviHeber:
Can you help me why sap etc. do not recognize the table creation despite saving the entity?
I already added a dummy field to prevent SAP failure on rows without non-generated values
For some reason it did not help here

gioboa (commented):
> Can you help me why sap ? 

@alumni can you help on this please?

alumni:
Initially the SAP test was failing because inserting a row without any values is not supported. Seems that was fixed in the meantime.

Now the tests for both Oracle and SAP (and for some reason CRDB) are failing because the code written for the `Raw` find operator query is missing:
- schema name from the `third_element` table (in some of our tests the TypeORM schema is not the default schema, so we need to specify it).
- quotations around identifiers (schema name, table name, column name) - otherwise the DBs (depending on the DB, DB settings and operating system) may transform the identifier to e.g. uppercase.

There might also be a separate bug with alias in `Raw` (not sure if the PR was merged or not - the alias was not properly quoted in some situations).

We need to use the full table path with quotes usually to make sure it's working in every situation. You can also use `QueryRunner.escapePath(tableName)` or `Driver.escape(identifier)` to implement something generic.

gioboa (changes requested):
Few tests are failing, can you fix them please?

LeviHeber:
I'm wondering if the problem is in the test, or if I performed the repair itself in an inappropriate way with certain DB engines.
The problem is that these are engines with which I am not sufficiently familiar, so I would be happy if someone could take a look and give an opinion.

michaelbromley:

```
git fetch origin
```

PR #11670, opened 2025-09-20
feat(mysql): add support for vector columns on MariaDB and MySQL

alumni:
### Description of change

Add support for vector columns in MySQL and MariaDB

### Pull-Request Checklist

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -12,7 +12,7 @@ services:
       MYSQL_DATABASE: "test"
 
   mysql-9:
-    image: "mysql:9.4.0"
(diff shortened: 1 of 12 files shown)


gioboa (approved):
LGTM 👏💪

alumni:
It should, it follows the same logic from string length (`Driver.withLengthColumnTypes`) - so the previous tests cover it.

The only adjustment I had to do was that the DB reported length in characters (bytes) is 4x the size of the vector, because each vector component takes 4 bytes (they are 32bit floating point numbers). This is covered by the tests in the PR already.

PR #11679, opened 2025-09-23
fix(postgres): widen varchar(n) via ALTER TYPE instead of drop+add

TheodorNEngoy:
Fixes #3357

### Description of change

When widening a PostgreSQL `varchar(n)` column to a larger length (e.g. 50 → 51),
`PostgresQueryRunner.changeColumn()` can safely do this with a single
`ALTER TABLE ... ALTER COLUMN ... TYPE character varying(newLen)` instead of
the heavier drop+add path. This PR adds a narrow, early‑return fast‑path that:

* triggers only when **old and new types are both `varchar` / `character varying`**,
* **new length > old length**,
* and **array dimensionality is unchanged**;

otherwise it falls back to the existing logic.

PostgreSQL supports changing a column’s data type using
`ALTER TABLE ... ALTER COLUMN ... TYPE ...` with optional `USING`. Widening the
length of `character varying` does not require `USING`. See PostgreSQL docs:  
https://www.postgresql.org/docs/current/sql-altertable.html  (section “`ALTER [ COLUMN ] column_name TYPE data_type [ USING expression ]`”). 

### Why this change?

* Avoids unnecessary drop/add for the common “widen length” case.
* Keeps existing data and constraints intact.
* Produces clearer, minimal SQL (no column recreation).

### How it’s implemented

Inside `PostgresQueryRunner.changeColumn()`:

* Detect the widening case (see bullets above).
* Compute the target type via `this.driver.createFullType(newColumn)` and run:
  `ALTER TABLE <table> ALTER COLUMN "<name>" TYPE <target>`
* Update the cached table metadata so subsequent operations see the new length.
* Return early. All other cases are untouched.

### Verification

A functional test is added:

* `test/functional/alter-varchar-length/postgres/alter-varchar-length.ts`

It creates a table with `varchar(50)`, widens to `varchar(51)` via
`changeColumn`, and asserts:

1. The executed SQL contains `ALTER TABLE ... ALTER COLUMN "name" TYPE ...varying(51)`.
2. There is **no** `DROP COLUMN "name"` or `ADD COLUMN "name"`.
3. Existing row with 50‑char value remains intact.

### Notes / scope

* Only applies to widening `varchar`/`character varying`. Narrowing, type
  changes, or dimensionality changes fall back to existing behavior.
* Arrays are supported by ensuring array dimensionality matches.

### Pull‑Request Checklist

- [ ] This pull request links relevant issues as `Fixes #00000` (if any)
- [x] There are new or updated unit/functional tests validating the change

## Summary by CodeRabbit

* **Bug Fixes**
  * Widening VARCHAR/character varying columns now uses an in-place ALTER TYPE, avoiding drop/add operations. This preserves data, reduces risk and downtime, and improves reliability during schema updates. Automatically falls back to previous behavior if needed.
* **Tests**
  * Added functional tests to verify data integrity and SQL generation when increasing a VARCHAR column length without dropping or re-adding the column.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -5017,9 +5017,9 @@
       }
     },
     "node_modules/diff": {
-      "version": "7.0.0",
-      "resolved": "https://registry.npmjs.org/diff/-/diff-7.0.0.tgz",
-      "integrity": "sha512-PJWHUb1RFevKCwaFA9RlG5tCd+FO5iRh9A8HEtkmBH2Li03iJriB6m6JIN4rGz3K3JLawI7/veA1xzRKP6ISBw==",
+      "version": "5.2.0",
+      "resolved": "https://registry.npmjs.org/diff/-/diff-5.2.0.tgz",
+      "integrity": "sha512-uIFDxqpRZGZ6ThOk84hEfqWoHx2devRFvpTZcTHur85vImfaxUbTW9Ryh4CpCuDnToOP1CEtXKIgytHBPVff5A==",
       "dev": true,
       "license": "BSD-3-Clause",
       "engines": {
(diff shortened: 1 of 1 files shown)


TheodorNEngoy:
This change only affects Postgres and adds a Postgres‑only functional test.
The failing jobs look unrelated (secrets not available to forked PRs).
Could a maintainer please kick off a trusted CI run for the full matrix?

PR #11382, opened 2025-04-04
feat!: drop support for node 16 and 18

alumni:
### Description of change

Drop support for older versions of node (16 and 18) and setup tests for node 22 and 23.

### Pull-Request Checklist

- [x] Code is up-to-date with the `master` branch
- [x] `npm run format` to apply prettier formatting
- [x] `npm run test` passes with this change
- [ ] This pull request links relevant issues as `Fixes #0000`
- [ ] There are new or updated unit tests validating the change
- [ ] Documentation has been updated to reflect this change
- [x] The new commits follow conventions explained in [CONTRIBUTING.md](https://github.com/typeorm/typeorm/blob/master/CONTRIBUTING.md)

## Summary by CodeRabbit

- **New Features**
  - Added a "typecheck" npm script for type checking without emitting output.

- **Improvements**
  - Updated support and configuration for JavaScript and TypeScript to ES2023 and Node.js 20+.
  - Upgraded dependencies, including better-sqlite3 and TypeScript, to newer versions.
  - Enhanced initial project setup templates with updated ECMAScript and driver versions.
  - Enabled the MSSQL connection in the sample configuration.

- **Documentation**
  - Updated documentation to reflect support for ES2023 in multiple languages and improved introductory content.

- **Refactor**
  - Refined TypeScript class property declarations for improved type safety and maintainability.
  - Adjusted workflow files for improved clarity, formatting, and updated Node.js versions in CI.

- **Chores**
  - Minor formatting and naming changes in workflow and configuration files.

PR #11683, opened 2025-09-24
chore(deps-dev): bump better-sqlite3 to 12.4.1

pkuczynski:
### Description of change

There are no binaries provided for macOS Tahoe and compiling package from source requires extensive setup, which is cumbersome. I propose we move forward with `better-sqlite3` to make it easier for developers to contribute.

### Pull-Request Checklist

## Summary by CodeRabbit

* **Chores**
  * Updated a development dependency (better-sqlite3) to a newer major version for improved compatibility and maintenance.
  * No impact on features, workflows, or runtime performance for end-users.

* **Documentation**
  * Confirmed there are no changes to exported/public interfaces.

* **General**
  * Maintenance-only update; no user-facing changes.

### DIFF
--- a/package.json
+++ b/package.json
@@ -127,7 +127,7 @@
     "@types/sinon-chai": "^3.2.12",
     "@types/source-map-support": "^0.5.10",
     "@types/yargs": "^17.0.33",
-    "better-sqlite3": "^8.7.0",
+    "better-sqlite3": "^12.4.1",
     "chai": "^4.5.0",
     "chai-as-promised": "^7.1.2",
     "class-transformer": "^0.5.1",
--- a/package-lock.json
+++ b/package-lock.json
@@ -46,7 +46,7 @@
         "@types/sinon-chai": "^3.2.12",
         "@types/source-map-support": "^0.5.10",
         "@types/yargs": "^17.0.33",
-        "better-sqlite3": "^8.7.0",
+        "better-sqlite3": "^12.4.1",
         "chai": "^4.5.0",
         "chai-as-promised": "^7.1.2",
         "class-transformer": "^0.5.1",
@@ -3466,15 +3466,18 @@
       "license": "Apache-2.0"
     },
     "node_modules/better-sqlite3": {
-      "version": "8.7.0",
-      "resolved": "https://registry.npmjs.org/better-sqlite3/-/better-sqlite3-8.7.0.tgz",
-      "integrity": "sha512-99jZU4le+f3G6aIl6PmmV0cxUIWqKieHxsiF7G34CVFiE+/UabpYqkU0NJIkY/96mQKikHeBjtR27vFfs5JpEw==",
+      "version": "12.4.1",
+      "resolved": "https://registry.npmjs.org/better-sqlite3/-/better-sqlite3-12.4.1.tgz",
+      "integrity": "sha512-3yVdyZhklTiNrtg+4WqHpJpFDd+WHTg2oM7UcR80GqL05AOV0xEJzc6qNvFYoEtE+hRp1n9MpN6/+4yhlGkDXQ==",
       "dev": true,
       "hasInstallScript": true,
       "license": "MIT",
       "dependencies": {
         "bindings": "^1.5.0",
         "prebuild-install": "^7.1.1"
+      },
+      "engines": {
+        "node": "20.x || 22.x || 23.x || 24.x"
       }
     },
     "node_modules/bignumber.js": {
--- a/src/commands/InitCommand.ts
+++ b/src/commands/InitCommand.ts
@@ -709,7 +709,7 @@ Steps to run this project:
                 packageJson.dependencies["sqlite3"] = "^5.1.7"
                 break
             case "better-sqlite3":
-                packageJson.dependencies["better-sqlite3"] = "^8.7.0"
+                packageJson.dependencies["better-sqlite3"] = "^12.4.1"
                 break
             case "oracle":
                 packageJson.dependencies["oracledb"] = "^6.8.0"


package.json
```diff
@@ -127,7 +127,7 @@
     "@types/sinon-chai": "^3.2.12",
     "@types/source-map-support": "^0.5.10",
     "@types/yargs": "^17.0.33",
-    "better-sqlite3": "^8.7.0",
+    "better-sqlite3": "^12.4.1",
```
gioboa, line 130: I'm not sure about this change, in the peerDeps we already have 
`"better-sqlite3": "^8.0.0 || ^9.0.0 || ^10.0.0 || ^11.0.0 || ^12.0.0",`
🤔 
cc @alumni

pkuczynski: Yes, this is for peer deps. The one I made was for dev deps. Version 8.7.0 is not distributed as binary anymore and requires compilation from source, at least on macOS Tahoe. So is it worth to be base for our tests, or would it be better to test with the latest version? I would vote for the second option. Alternatively we could (and probably should) setup testing for the whole supported matrix in CI. But for local dev, I would stick with the latest as it will always get most support.

alumni: The reason we didn't do this in the past was because we are still officially supporting Node.js 16, so we wanted the tests to be able to run on that version, even if we removed them from the CI.

`better-sqlite3@12.x` requires Node.js 20 and higher. We'll drop support for EOL node in our next major version, so we can upgrade this then.

pkuczynski: I kind of sense that answer. And how do you install version 8 on macOS? I could not find any instruction here...

alumni: Okay, so `better-sqlite3` does not compile on node 22 (probably not just on MacOS). I am still running 20 locally in this repo - mainly because I want to generate a lockfile that still works in CI (and node 22 upgrades npm - we'll probably introduce corepack at some point to avoid such issues).

What if we use `^11.10.0`? It already has prebuilds for node 18 and 20 (our CI runs on that, so it won't have to compile it), even up to node 23. I also managed to build it locally on MacOS on node 24.

alumni: `src/commands/InitCommand.ts:712` should probably also be updated. We have tests for that and they will be slower if we don't change it - shouldn't be a breaking change.

pkuczynski: I am on Node 24, and 11.x does not have binaries and needs to be compiled from scratch. And since Node 24 is the current version, I think it's fine to stick to it, as most people will start with Node 24. And then they will have issues with installing dev depenedencies to create a PR.

How about we use latest version in dev deps, but test in a matrix in CI like this: https://github.com/typeorm/typeorm/pull/11688

I udated `InitCommand` but it got me wondering, if we should not reach to npm to get the latets version compatible with current node. What do you think?

alumni: > I am on Node 24, and 11.x does not have binaries and needs to be compiled from scratch.

Compiling on MacOS/Linux is not difficult. Compiling in the CI takes time and makes the CI slow.

> And since Node 24 is the current version, I think it's fine to stick to it, as most people will start with Node 24

In production most companies use the LTS version of Node.js, currently 22. The stable version is seen as a development build.

> How about we use latest version in dev deps, but test in a matrix in CI like this: https://github.com/typeorm/typeorm/pull/11688

We had this in the past with the `mssql` driver and we realized it does not bring any value and only makes the CI slow.

> I udated InitCommand but it got me wondering, if we should not reach to npm to get the latets version compatible with current node.

This is not how Node.js works. We have some ideas to handle the lockfile better by pinning the package manager version using `corepack`, but I personally don't want to invest any effort in the 0.3.x releases in such things due to the fact that we need to keep support for old Node versions. Such things will happen in 1.0, including dropping support for old `better-sqlite3`. But until then, we can't use 12.x.

pkuczynski: > Compiling on MacOS/Linux is not difficult. 

Not in this example, as it requires setting up Python env, so pain in the neck...

> In production most companies use the LTS version of Node.js, currently 22. 

Depends on the company. In most of those I worked for, _current_ Node was always selected as the prime env...

> We had this in the past with the mssql driver and we realized it does not bring any value and only makes the CI slow.

Yes, this would be the downside for too many jobs. It's either CI testing or users...

> This is not how Node.js works. 

Maybe you misunderstood what I was saying. I didn't refered to `package.json` but to to command which generates one when you call `init` (this is what `InitCommand` does, right?).

And updating `devDependencies`, which is proposed in this PR does not mean dropping support for older node, as the `better-sqlite3` would not be installed for the end users. It's only for developers like me, right?

Feel free to close this PR if you feel it's unwanted. No offence taken :)

alumni: > Maybe you misunderstood what I was saying

Aaaah you meant to get the version of `better-sqlite3` from npm for the nodejs version that the user has. Sorry, I thought you meant to request smth from the nodejs team. Honestly, I'd don't like the init command at all and I'd remove it in the next major version if it were just for me. We have to talk about it with the rest of the maintainers, but it has little value nowadays.

> Feel free to close this PR if you feel it's unwanted. No offence taken :)

For the `better-sqlite3` version used as dev dependency, v11 is okay for most of the things, just v12 will make the current workflows slower. So no need to close this PR. We can merge it with v11 into 0.3 or with v12 into the next major.

We will have a meeting on Monday and talk about the next major, we'd like to start introducing breaking changes (removing deprecations, dropping support for old versions etc.) that are necessary to continue maintaining this repo. 

BTW, I see you have quite a few PRs open, if you are interesting in contributing more feel free to reach out on Discord ;)

pkuczynski: > Aaaah you meant to get the version of better-sqlite3 from npm for the nodejs version that the user has.

Yeah, that's what I meant :) It does not make sense to propose user 8.x when user is already on node 22 or 24, since it would not even install. As we fund out in case of this PR :)

> BTW, I see you have quite a few PRs open, if you are interesting in contributing more feel free to reach out on Discord ;)

Yeah, that was my intent to contribute a bit more :) I will reach out on Discord, sure. I am already there...

alumni:
#11382 was merged into `next`

PR #11686, opened 2025-09-24
feat: Support shared columns in junction tables for composite foreign keys

### LINKED ISSUE #11682: Support shared columns in junction tables for composite foreign keys
### Feature Description

Currently, TypeORM's junction table implementation automatically renames columns that appear in both `joinColumns` and `inverseJoinColumns` to avoid duplicates. While this works well for most cases, it would be valuable to have an option to preserve shared columns that are required for composite foreign key constraints in partitioned many-to-many relationships.

## Desired Behavior

With this feature, junction tables with composite foreign keys could maintain shared columns and generate valid SQL:

```sql
-- Desired: Shared column maintained for composite FK constraints
CREATE TABLE "user_groups" (
    "tenant_id" varchar NOT NULL,  -- Shared column for both FKs
    "user_id" varchar NOT NULL,
    "group_id" varchar NOT NULL,
    PRIMARY KEY ("tenant_id", "user_id", "group_id"),
    FOREIGN KEY ("tenant_id", "user_id") REFERENCES "user"("tenant_id", "id"),
    FOREIGN KEY ("tenant_id", "group_id") REFERENCES "group"("tenant_id", "id")
) PARTITION BY HASH ("tenant_id");
```

INSERT statements would include proper values:

```sql
INSERT INTO "user_groups"("tenant_id", "user_id", "group_id") VALUES ($1, $2, $3)
```

## Current Behavior

Currently, TypeORM automatically renames shared columns, which prevents composite foreign key constraints:

```sql
-- Current: Column names are automatically renamed
CREATE TABLE "user_groups" (
    "user_id" varchar NOT NULL,
    "tenant_id_1" varchar NOT NULL,    -- Automatically renamed
    "group_id" varchar NOT NULL,
    "tenant_id_2" varchar NOT NULL     -- Automatically renamed
);
```

This current approach creates limitations in certain scenarios:

-   Column references don't match the intended schema
-   Foreign key constraints cannot be properly established
-   Shared columns required for composite keys are not supported

### Entity Definition:

```typescript
@Entity()
export class User {
    @PrimaryColumn()
    id: string

    @PrimaryColumn()
    tenantId: string

    @ManyToMany(() => Group)
    @JoinTable({
        name: "user_groups",
        joinColumns: [
            { name: "tenant_id", referencedColumnName: "tenantId" },
            { name: "user_id", referencedColumnName: "id" },
        ],
        inverseJoinColumns: [
            { name: "tenant_id", referencedColumnName: "tenantId" }, // Same column needed
            { name: "group_id", referencedColumnName: "id" },
        ],
    })
    groups: Group[]
}

@Entity()
export class Group {
    @PrimaryColumn()
    id: string

    @PrimaryColumn()
    tenantId: string

    @ManyToMany(() => User, (user) => user.groups)
    users: User[]
}
```

### Current Limitation:

1. Entities with composite primary keys require shared columns in junction tables
2. `@JoinTable` configuration needs the same column in both `joinColumns` and `inverseJoinColumns`
3. TypeORM automatically renames these columns to avoid duplicates
4. This prevents proper composite foreign key constraints
5. Manual schema creation is required as a workaround

### Technical Background:

`JunctionEntityMetadataBuilder` automatically detects columns with identical names in both `joinColumns` and `inverseJoinColumns` and renames them to avoid duplicates. This behavior prevents the creation of composite foreign key constraints that require shared columns.

```typescript
// JunctionEntityMetadataBuilder.ts - build() method
this.changeDuplicatedColumnNames(junctionColumns, inverseJunctionColumns)
```

### The Solution

Add configuration options to provide flexibility for shared columns at both table and column levels:

### Implementation Approach:

When `preserveSharedColumns` is enabled, instead of renaming duplicate columns, TypeORM would:

1. **Identify shared columns** between `joinColumns` and `inverseJoinColumns`
2. **Preserve the column names** (skip the renaming step)
3. **Mark inverse side columns as non-insertable** by setting `isInsert = false` and `isUpdate = false`
4. **Leverage existing query builder logic** that already respects these flags to exclude columns from INSERT/UPDATE statements

This approach prevents SQL errors like `column "tenant_id" specified more than once` while maintaining the correct schema structure.

#### Table-Level Configuration:

```typescript
// src/decorator/options/JoinTableOptions.ts
export interface JoinTableOptions {
    name?: string
    database?: string
    schema?: string
    joinColumn?: JoinColumnOptions
    joinColumns?: JoinColumnOptions[]
    inverseJoinColumn?: JoinColumnOptions
    inverseJoinColumns?: JoinColumnOptions[]
    synchronize?: boolean
    preserveSharedColumns?: boolean // New flag - applies to all shared columns
}
```

#### Column-Level Configuration:

```typescript
// src/metadata-args/JoinColumnMetadataArgs.ts
export interface JoinColumnMetadataArgs {
    // ... existing properties
    preserveSharedColumn?: boolean // New flag - granular control per column
}
```

#### Usage Examples:

**Simple Case - Table-Level:**

```typescript
@JoinTable({
    name: "user_groups",
    preserveSharedColumns: true, // Enable shared column handling for ALL columns
    joinColumns: [
        { name: "tenant_id", referencedColumnName: "tenantId" },
        { name: "user_id", referencedColumnName: "id" },
    ],
    inverseJoinColumns: [
        { name: "tenant_id", referencedColumnName: "tenantId" }, // Preserved
        { name: "group_id", referencedColumnName: "id" },
    ],
})
```

**Complex Case - Column-Level:**

```typescript
@JoinTable({
    name: "complex_junction",
    joinColumns: [
      { name: "tenant_id", referencedColumnName: "tenantId" },
      { name: "region_id", referencedColumnName: "regionId" }, // Renamed
      { name: "user_id", referencedColumnName: "id" },
    ],
    inverseJoinColumns: [
      { name: "tenant_id", referencedColumnName: "tenantId", preserveSharedColumn: true }, // Preserved
      { name: "region_id", referencedColumnName: "regionId" }, // Renamed
      { name: "group_id", referencedColumnName: "id" },
    ],
})
```

### Considered Alternatives

Instead of using `isInsert`/`isUpdate` flags, an alternative approach would modify `ManyToManySubjectBuilder` to deduplicate `changeMaps` when creating junction subjects, ensuring only one entry per unique column name is passed to the query builder.

This approach would also be driven by the `preserveSharedColumns`/`preserveSharedColumn` flags, adding metadata to `ColumnMetadata` (e.g., `isSharedColumn` flag) to identify which columns should be deduplicated.

```typescript
// Alternative implementation in ManyToManySubjectBuilder
const changeMapsByColumn = new Map<string, any>()

relation.junctionEntityMetadata!.ownerColumns.forEach((column) => {
    changeMapsByColumn.set(column.databaseName, { column, value: ownerValue })
})

relation.junctionEntityMetadata!.inverseColumns.forEach((column) => {
    // Only deduplicate if column is marked as shared
    if (
        !changeMapsByColumn.has(column.databaseName) ||
        !column.isSharedColumn
    ) {
        changeMapsByColumn.set(column.databaseName, {
            column,
            value: inverseValue,
        })
    }
})

junctionSubject.changeMaps = Array.from(changeMapsByColumn.values())
```

The `isSharedColumn` flag would be set during junction metadata building based on the `preserveSharedColumns`/`preserveSharedColumn` configuration.

### Additional Context

**Important Note: DDL vs DML/DQL Scope**

This enhancement primarily addresses DML and DQL operations - specifically how INSERT/UPDATE/DELETE/SELECT statements are generated for junction tables. It does not cover DDL operations like table creation.

Since TypeORM cannot create partitioned tables, the junction table must be created manually with `synchronize: false`. This enhancement would ensure that once the partitioned table exists, TypeORM can properly:

-   Generate correct INSERT/UPDATE/DELETE statements with shared columns
-   Execute SELECT queries with proper JOIN conditions
-   Handle many-to-many relationships without column name conflicts

- [x] aurora-mysql
- [x] aurora-postgres
- [x] better-sqlite3
- [x] cockroachdb
- [ ] cordova
- [ ] expo
- [ ] mongodb
- [x] mysql
- [ ] nativescript
- [x] oracle
- [x] postgres
- [ ] react-native
- [ ] sap
- [ ] spanner
- [x] sqlite
- [ ] sqlite-abstract
- [ ] sqljs
- [x] sqlserver

nlenepveu:
# feat: Support shared columns in junction tables for composite foreign keys

**EDIT: Implementation Updated Based on Feedback**

The implementation has been updated following @sgarner @alumni review to treat this as a **bug fix** rather than a feature. The original configuration-based approach has been replaced with a simpler solution that respects explicitly named columns.

**Key Changes:**

-   Removed `preserveSharedColumns` and `preserveSharedColumn` configuration options
-   Added `isExplicitlyNamed` property to distinguish user-provided vs auto-generated column names
-   Modified column renaming logic to preserve explicitly named columns while maintaining backward compatibility for implicit names
-   Refactored shared column handling into separate `configureSharedColumns` method

**Impact:** Only affects users who explicitly set duplicate column names and expected TypeORM to automatically rename them. The vast majority of users will see no change in behavior.

---

## Description of change

This pull request implements **shared column preservation for junction tables in composite foreign key scenarios**, addressing issue #11682.

**Problem Solved:**
Currently, TypeORM automatically renames columns that appear in both `joinColumns` and `inverseJoinColumns` to avoid duplicates (e.g., `tenant_id` → `tenant_id_1`, `tenant_id_2`). This prevents composite foreign key constraints in partitioned many-to-many relationships.

**Solution Implemented:**
~~Adds configuration options to preserve shared columns~~ **Updated approach based on maintainer feedback:**

The implementation now **respects explicitly named columns** instead of adding new configuration options. This is treated as a **bug fix** rather than a feature:

```typescript
@JoinTable({
    name: "user_groups_shared",
    joinColumns: [
        { name: "tenant_id", referencedColumnName: "tenantId" }, // Explicit name - preserved
        { name: "user_id", referencedColumnName: "id" },
    ],
    inverseJoinColumns: [
        { name: "tenant_id", referencedColumnName: "tenantId" }, // Explicit name - preserved (shared)
        { name: "group_id", referencedColumnName: "id" },
    ],
})
```

**Technical Implementation:**

-   ~~Pre-calculates shared columns during metadata building~~ **Removed configuration-based approach**
-   ~~Skips renaming for marked shared columns~~ **Now skips renaming for explicitly named columns**
-   **Added `isExplicitlyNamed` property** to track user-provided vs auto-generated column names
-   **Maintains backward compatibility**: implicit column names still get renamed as before
-   **Refactored shared column handling** into separate `configureSharedColumns` method
-   Marks inverse-side shared columns as non-insertable/updatable to prevent duplicate column SQL errors

**Behavior Changes:**

| Scenario              | Before                       | After                                    |
| --------------------- | ---------------------------- | ---------------------------------------- |
| Implicit column names | `tenant_id_1`, `tenant_id_2` | `tenant_id_1`, `tenant_id_2` (unchanged) |
| Explicit column names | `tenant_id_1`, `tenant_id_2` | `tenant_id`, `tenant_id` (preserved)     |

**Generated DML/DQL:**

```sql
-- With explicit shared column names (new behavior)
INSERT INTO "user_groups_shared"("tenant_id", "user_id", "group_id") VALUES ($1, $2, $3);

-- With implicit column names (existing behavior preserved)
INSERT INTO "user_groups"("tenant_id_1", "user_id", "tenant_id_2", "group_id") VALUES ($1, $2, $3, $4);
```

**Current Scope - DML/DQL Focus:**
This implementation primarily addresses the **immediate pain point** described in issue #11682: the inability to use TypeORM's DML/DQL operations with user-created partitioned junction tables.

The main problem users face today is that when they manually create partitioned junction tables with shared columns (which TypeORM currently cannot create), TypeORM's column renaming logic generates incorrect INSERT/UPDATE/DELETE statements that don't match the actual table schema.

This feature solves that core issue by allowing TypeORM to generate correct DML/DQL statements for existing tables with shared columns **when column names are explicitly provided**.

**DDL Considerations:**
While DDL operations (table creation) are not addressed in this implementation, they remain an important consideration for future enhancements. For now, users must create junction tables manually when using partitioned tables or composite foreign key constraints.

The junction table must be created manually like:

```sql
CREATE TABLE "user_groups" (
    "tenant_id" varchar NOT NULL,  -- Shared column for both FKs
    "user_id" varchar NOT NULL,
    "group_id" varchar NOT NULL,
    PRIMARY KEY ("tenant_id", "user_id", "group_id"),
    FOREIGN KEY ("tenant_id", "user_id") REFERENCES "user"("tenant_id", "id"),
    FOREIGN KEY ("tenant_id", "group_id") REFERENCES "group"("tenant_id", "id")
) PARTITION BY HASH ("tenant_id");
```

This feature ensures TypeORM can then properly interact with such manually-created tables through correct DML/DQL generation **when explicit column names are used**.

## How this change was verified

**Comprehensive Test Suite (20 tests passing):**

-   Regression tests for existing column renaming behavior (implicit names)
-   New tests for explicit column name preservation
-   SQL generation validation for INSERT/DELETE/UPDATE operations
-   Functional relationship management from both owner and inverse sides
-   Cross-driver testing (SQLite and PostgreSQL)

## Breaking Changes

**Minimal breaking change exposure**: Only affects users who:

1. Explicitly set the same column name for both `joinColumns` and `inverseJoinColumns`
2. AND expected TypeORM to automatically rename them with suffixes

This is considered a **bug fix** since the original behavior of renaming explicitly named columns was unintended and prevented valid use cases.

## Pull-Request Checklist

-   [x] This pull request links relevant issues as `Fixes #11682`
-   [x] There are new or updated unit tests validating the change (20 comprehensive tests)
-   [x] Documentation has been updated to reflect this change (JSDoc comments on new functionality)
-   [x] **Updated implementation** based on maintainer feedback to respect explicit column names instead of adding configuration options

## Summary by CodeRabbit

- New Features
  - Improved junction table handling: preserves explicitly named join columns and supports shared columns in many-to-many relations.
  - Smarter column naming: only renames implicit duplicates, keeping user-specified names intact.
- Bug Fixes
  - Prevents duplicate bindings by marking shared inverse columns as non-insertable/non-updatable.
  - More reliable SQL generation for junction tables with explicit or shared columns.
- Tests
  - Added comprehensive functional tests for implicit, partially explicit, and fully explicit junction table column naming.
  - Introduced an in-memory query logger for test assertions.

### DIFF
--- a/src/metadata-builder/JunctionEntityMetadataBuilder.ts
+++ b/src/metadata-builder/JunctionEntityMetadataBuilder.ts
@@ -197,11 +197,6 @@ export class JunctionEntityMetadataBuilder {
                             name: columnName,
                             nullable: false,
                             primary: true,
-                            // Shared columns on inverse side are non-insertable/updatable to prevent
-                            // duplicate column bindings in DML operations (e.g., INSERT INTO table (tenant_id, tenant_id, ...))
-                            // Only the owner side (with @JoinTable) controls shared column values
-                            insert: !sharedColumnNames.includes(columnName),
-                            update: !sharedColumnNames.includes(columnName),
                         },
                     },
                 })
@@ -211,8 +206,12 @@ export class JunctionEntityMetadataBuilder {
         this.changeDuplicatedColumnNames(
             junctionColumns,
             inverseJunctionColumns,
-            sharedColumnNames,
         )
+
+        // Shared columns on inverse side are non-insertable/updatable to prevent
+        // duplicate column bindings in DML operations (e.g., INSERT INTO table (tenant_id, tenant_id, ...))
+        // Only the owner side (with @JoinTable) controls shared column values
+        this.configureSharedColumns(junctionColumns, inverseJunctionColumns)
 
         // set junction table columns
         entityMetadata.ownerColumns = junctionColumns
@@ -372,7 +371,6 @@ export class JunctionEntityMetadataBuilder {
     protected changeDuplicatedColumnNames(
         junctionColumns: ColumnMetadata[],
         inverseJunctionColumns: ColumnMetadata[],
-        sharedColumnNames: string[],
     ) {
         junctionColumns.forEach((junctionColumn) => {
             inverseJunctionColumns.forEach((inverseJunctionColumn) => {
@@ -380,68 +378,64 @@ export class JunctionEntityMetadataBuilder {
                     junctionColumn.givenDatabaseName ===
                     inverseJunctionColumn.givenDatabaseName
                 ) {
-                    // Skip renaming if this is a shared column
-                    if (
-                        sharedColumnNames.includes(
-                            junctionColumn.givenDatabaseName!,
-                        ) ||
-                        sharedColumnNames.includes(
-                            inverseJunctionColumn.givenDatabaseName!,
-                        )
-                    ) {
-                        return
+                    let index = 1
+                    // If explicit name is provided, don't rename
+                    // If implicit, rename them (backward compatibility)
+                    if (!junctionColumn.isExplicitlyNamed) {
+                        const junctionColumnName =
+                            this.connection.namingStrategy.joinTableColumnDuplicationPrefix(
+                                junctionColumn.propertyName,
+                                index++,
+                            )
+                        junctionColumn.propertyName = junctionColumnName
+                        junctionColumn.givenDatabaseName = junctionColumnName
                     }
 
-                    // Original renaming logic for non-shared columns
-                    const junctionColumnName =
-                        this.connection.namingStrategy.joinTableColumnDuplicationPrefix(
-                            junctionColumn.propertyName,
-                            1,
-                        )
-                    junctionColumn.propertyName = junctionColumnName
-                    junctionColumn.givenDatabaseName = junctionColumnName
-
-                    const inverseJunctionColumnName =
-                        this.connection.namingStrategy.joinTableColumnDuplicationPrefix(
-                            inverseJunctionColumn.propertyName,
-                            2,
-                        )
-                    inverseJunctionColumn.propertyName =
-                        inverseJunctionColumnName
-                    inverseJunctionColumn.givenDatabaseName =
-                        inverseJunctionColumnName
+                    if (!inverseJunctionColumn.isExplicitlyNamed) {
+                        const inverseJunctionColumnName =
+                            this.connection.namingStrategy.joinTableColumnDuplicationPrefix(
+                                inverseJunctionColumn.propertyName,
+                                index++,
+                            )
+                        inverseJunctionColumn.propertyName =
+                            inverseJunctionColumnName
+                        inverseJunctionColumn.givenDatabaseName =
+                            inverseJunctionColumnName
+                    }
                 }
             })
         })
     }
 
     /**
-     * Pre-calculates which column names will be shared between joinColumns and inverseJoinColumns
-     * based on the preserveSharedColumns configuration.
+     * Handles shared columns in junction tables by preventing duplicate column bindings.
+     *
+     * When both sides of a many-to-many relation reference the same physical database column
+     * (e.g., both sides use "tenant_id"), we need to ensure only one side controls the column
+     * to avoid SQL errors like: INSERT INTO junction (tenant_id, tenant_id, ...) VALUES (1, 1, ...)
+     *
+     * This enables composite foreign key constraints in partitioned many-to-many relationships
+     * where a shared column (like tenant_id) is part of both foreign keys.
+     *
+     * @param junctionColumns Columns from the owner side (entity with @JoinTable)
+     * @param inverseJunctionColumns Columns from the inverse side
      */
-    protected calculateSharedColumnNames(
-        joinTable: JoinTableMetadataArgs,
-    ): string[] {
-        const sharedColumnNames: string[] = []
-
-        if (!joinTable.joinColumns || !joinTable.inverseJoinColumns) {
-            return sharedColumnNames
-        }
-
-        // Check each joinColumn against inverseJoinColumns for duplicates
-        joinTable.joinColumns.forEach((joinColumn) => {
-            if (!joinColumn.name) return
-
-            const matchingInverseColumn = joinTable.inverseJoinColumns!.find(
-                (inverseJoinColumn) =>
-                    inverseJoinColumn.name === joinColumn.name,
-            )
-
-            if (matchingInverseColumn) {
-                sharedColumnNames.push(joinColumn.name)
-            }
+    protected configureSharedColumns(
+        junctionColumns: ColumnMetadata[],
+        inverseJunctionColumns: ColumnMetadata[],
+    ) {
+        junctionColumns.forEach((junctionColumn) => {
+            inverseJunctionColumns.forEach((inverseJunctionColumn) => {
+                if (
+                    junctionColumn.givenDatabaseName ===
+                    inverseJunctionColumn.givenDatabaseName
+                ) {
+                    // Mark inverse side column as read-only to prevent duplicate bindings
+                    // The owner side (with @JoinTable) retains control over shared column values
+                    inverseJunctionColumn.isInsert = false
+                    inverseJunctionColumn.isUpdate = false
+                }
+            })
         })
-
-        return sharedColumnNames
     }
 }
--- a/src/metadata/ColumnMetadata.ts
+++ b/src/metadata/ColumnMetadata.ts
@@ -311,6 +311,12 @@ export class ColumnMetadata {
     foreignKeyConstraintName?: string
 
     /**
+     * Indicates whether this column name was explicitly provided by the user
+     * (e.g., via @JoinColumn({ name: "custom_name" })) or auto-generated via the naming strategy by TypeORM.
+     */
+    isExplicitlyNamed?: boolean
+
+    /**
      * Specifies a value transformer that is to be used to (un)marshal
      * this column when reading or writing to the database.
      */
@@ -364,10 +370,12 @@ export class ColumnMetadata {
         nestedSetLeft?: boolean
         nestedSetRight?: boolean
         materializedPath?: boolean
+        isExplicitlyNamed?: boolean
     }) {
         this.entityMetadata = options.entityMetadata
         this.embeddedMetadata = options.embeddedMetadata!
         this.referencedColumn = options.referencedColumn
+        this.isExplicitlyNamed = options.isExplicitlyNamed
         if (options.args.target) this.target = options.args.target
         if (options.args.propertyName)
             this.propertyName = options.args.propertyName


sgarner (commented):
Thanks for the contribution @nlenepveu 

I feel like the `preserveSharedColumn` option is unnecessary and overcomplicates the implementation.

Couldn't we just make TypeORM respect manually configured column names? i.e. the renaming of columns for deduplication should **only** apply when we are automatically generating a column name. 

So if the user said both fields should be named `tenant_id` then that should be respected. But if they didn't give a column name then we'd generate `tenant_id_1` and `tenant_id_2`. (I'm actually surprised it doesn't already work that way.)

nlenepveu:
> I feel like the `preserveSharedColumn` option is unnecessary and overcomplicates the implementation.
> 
> Couldn't we just make TypeORM respect manually configured column names? i.e. the renaming of columns for deduplication should **only** apply when we are automatically generating a column name.
> 
> So if the user said both fields should be named `tenant_id` then that should be respected. But if they didn't give a column name then we'd generate `tenant_id_1` and `tenant_id_2`. (I'm actually surprised it doesn't already work that way.)

@sgarner Thanks for the feedback!

The design decision behind `preserveSharedColumn` was mainly to avoid changing the existing behavior and potentially breaking backwards compatibility. I’m very much in favor of a simpler API, and indeed we were also disappointed by the current behavior where manually configured column names get silently renamed.

If you think it’s acceptable to remove the automatic renaming in favor of always respecting user-provided names, and that the backward-compatibility risk is manageable, I’m fully supportive of going in that direction.

sgarner:
Yes, without the option it would mean this change could be breaking for some users.

I think that may still be preferable to adding an option that produces two different behaviors, one of which is undesirable, which we then have to maintain.

Since the previous behavior was never documented, and this scenario probably doesn't affect many people, including such a break in a 0.3.x point release might be acceptable anyway.

Otherwise this could wait until 1.x.

WDYT @alumni ?

nlenepveu:
I agree the breaking change shouldn’t be a big deal to fix. Users can always explicitly rename the column themselves in the join table definition. So this feels manageable even in a 0.3.x release.

That said, if we want to be more compliant with semver and only introduce BC breaks in a major (or maybe a `0.4` given we haven’t had a major yet), we could also take a middle path:
- Release `preserveSharedColumn` now, but mark it as deprecated right away.
- Add a `console.warn` whenever a column gets renamed automatically, informing users that this behavior will change in the next major.

This way, nothing breaks immediately, we can ship the fix in the next release, and users get clear warning that they need to adjust. Then in the next major we drop the option entirely and default to the simpler, correct behavior.

alumni:
I think if the properties (columns) are explicitly named, TypeORM should always respect the user's choice and not rename anything (basically I see the current behavior as a bug), i.e.:

```ts
joinColumns: [
    { name: "tenant_id", referencedColumnName: "tenantId" },
    { name: "user_id", referencedColumnName: "id" },
],
inverseJoinColumns: [
    { name: "tenant_id", referencedColumnName: "tenantId" }, // Same column - will be shared
    { name: "group_id", referencedColumnName: "id" },
],
```
should produce an association (join, junction, etc) table with 3 columns (`tenant_id`, `user_id`, `group_id`).

Side note: it has always bugged me that `name` in `JoinColumnOptions` must be a column name, but `referencedColumnName` is a property name. I am starting to think it could makes sense (except for `referencedColumnName`) :)

sgarner:
> Side note: it has always bugged me that `name` in `JoinColumnOptions` must be a column name, but `referencedColumnName` is a property name.

Yes! There are several places where the ambiguity between "column names" and "property names" is confusing in TypeORM, but this example is particularly egregious. We should improve this sometime.

sgarner:
@nlenepveu Could you please update the implementation to remove the `preserveSharedColumn` option and respect explicitly set column names instead?

We will treat this as a bug fix rather than a feature. 

Note that we aren't adding specific tests for GitHub issues any more, so the tests in this PR should also be moved to the appropriate test suite(s) for the functionality they are testing.

alumni:
I'll try tomorrow to take a look at the history of this `JunctionEntityMetadataBuilder.changeDuplicatedColumnNames` - I'd like to see why it was added. The current behavior is intentional, but I'd like to understand the context.

Maybe it's not needed to keep this behavior.

Also: the methods of these metadata builder classes are protected, even thought the classes are not extended. Maybe it's also some leftover code from a different time.

_Later edit:_ https://github.com/typeorm/typeorm/commit/8934e27dc412a90f8ef5b80aac19ebc86d59c054 introduced `changeDuplicatedColumnNames`. But why?

nlenepveu:
I’m fine to drop the `preserveSharedColumn` option and rework the PR accordingly.

I’ll update the PR to:
- Remove the `preserveSharedColumn` option.
- Adjust the implementation so explicitly set column names are always respected, and only implicit names are subject to renaming via `changeDuplicatedColumnNames`.
- Move the tests into the appropriate existing test suites rather than keeping them tied to the GitHub issue.

On the side note: totally agree about the `name` vs `referencedColumnName` inconsistency, that’s confusing.

nlenepveu:
> _Later edit:_ [8934e27](https://github.com/typeorm/typeorm/commit/8934e27dc412a90f8ef5b80aac19ebc86d59c054) introduced `changeDuplicatedColumnNames`. But why?

Looking at the example, it seems this was added to support using `@JoinTable()` without options in self-referencing entities (looks like a graph-like structure). In that context, it makes sense. The developer doesn’t care about the exact table design and just wants the ORM to generate something that works out of the box. This supports keeping the current behavior when join columns are implicit, while removing it when column names are set explicitly.

sgarner:
> That said, if we want to be more compliant with semver and only introduce BC breaks in a major (or maybe a `0.4` given we haven’t had a major yet), we could also take a middle path

Since TypeORM is still on the 0.x track we are technically able to make breaking changes in any release and be compliant with semver. But of course we try not to do that since many thousands of people are using TypeORM in production already.

Either way, a change in behavior is not considered breaking if the original behavior was a bug and/or undocumented. I think that applies in this case.

Under the proposed change plan the exposure is low. Only if a user explicitly set the same name for two columns in a junction table and expected TypeORM to suffix them will they see a break. If they didn't set names, the current behavior continues.

We'll discuss with other maintainers to make a final decision but my personal view is this is acceptable for a 0.3.x release.

src/metadata-builder/JunctionEntityMetadataBuilder.ts
```diff
@@ -372,23 +378,62 @@ export class JunctionEntityMetadataBuilder {
...
-                    junctionColumn.propertyName = junctionColumnName
-                    junctionColumn.givenDatabaseName = junctionColumnName
+                    let index = 1
+                    // If explicit name is provided, don't rename
+                    // If implicit, rename them (backward compatibility)
+                    if (!junctionColumn.isExplicitlyNamed) {
+                        const junctionColumnName =
+                            this.connection.namingStrategy.joinTableColumnDuplicationPrefix(
+                                junctionColumn.propertyName,
+                                index++,
```
nlenepveu, line 388: Not sure about this, but this would be weird to starts with `_2` when the name is explicitly provided in `joinColumns` but not in `inverseJoinColumns`.

nlenepveu:
@sgarner @alumni

I’ve updated the PR based on the discussion to ensure explicitly set column names are always respected. This should now align with the intended direction. Let me know if you see anything blocking, otherwise I think it’s ready to move forward.

src/metadata-builder/JunctionEntityMetadataBuilder.ts
```diff
@@ -144,20 +145,20 @@ export class JunctionEntityMetadataBuilder {
...
-                const columnName =
-                    joinColumn && joinColumn.name
-                        ? joinColumn.name
-                        : this.connection.namingStrategy.joinTableInverseColumnName(
-                              relation.inverseEntityMetadata
-                                  .tableNameWithoutPrefix,
-                              inverseReferencedColumn.propertyName,
-                              inverseReferencedColumn.databaseName,
-                          )
+                const isExplicitlyNamed = !!(joinColumn && joinColumn.name)
```
alumni, line 148: I think we don't have the `stylistic-type-checked` ESLint rules enabled yet, but for when we will:

`joinColumn && joinColumn.name` -> `joinColumn?.name`
`!!expression` -> `Boolean(expression)`

gioboa (commented):
Can you rebase this PR with the current master branch please?

nlenepveu:
> Can you rebase this PR with the current master branch please?

Done!

gioboa:
> Done!

Thanks

gioboa (commented):
The test file must have `.test.ts` extension to be considered. so we need to rename
test/functional/relations/junction-table-column-naming/junction-table-column-naming.ts

gioboa (changes requested):
The test file must have .test.ts extension to be considered. so we need to rename
test/functional/relations/junction-table-column-naming/junction-table-column-naming.ts

nlenepveu:
@gioboa 
Totally fair - sorry I let it go stale. I don’t have much time to pick this up right now.

gioboa:
No problem at all, that's fine 🙏

PR #11651, opened 2025-09-16
ci: add close stale issues GH action

gioboa:
TypeORM has returned to active development after recently transitioning to new leadership ([see related announcement](https://typeorm.io/docs/future-of-typeorm/)).
As part of the new team's efforts to prioritise work on the project moving forward, it is necessary to clean up the backlog of stale issues.

### Pull-Request Checklist

-   [x] Code is up-to-date with the `master` branch

## Summary by CodeRabbit

* **Chores**
  * Added an automated workflow to mark issues stale after 180 days and close them immediately thereafter.
  * Stale issues are labeled "stale"; "pinned" and "security" issues are exempt.
  * Close message provides backlog-cleanup context, a leadership update, and invites reopening if still relevant.
  * Workflow can be triggered manually for maintenance.

PR #11696, opened 2025-09-30
ci: remove close stale issues GH action

gioboa:
Reverts typeorm/typeorm#11651


alumni:
@gioboa I wouldn't necessarily remove it - we could just tweak the message and allow 30 days for people to reply :)

PR #11700, opened 2025-10-01
chore: disable eslint errors for chai assertions

alumni:
### Description of change

Currently we set `@typescript-eslint/no-unused-expressions` to warn, which highlights a lot of errors in our `chai` assertions. When using both styles of assertion ("expect" and "should"), `chai` has side-effects which eslint is trying to prevent with this rule.

I disabled this rule for the assertions in the tests via `eslint-plugin-chai-friendly`, and set it to error in the rest of the code. I also corrected the remaining errors.

Before: 10807 warnings
After: 9207 warnings

### Pull-Request Checklist

-   [x] Code is up-to-date with the `next` branch

### DIFF
--- a/eslint.config.mjs
+++ b/eslint.config.mjs
@@ -39,7 +39,6 @@ export default defineConfig([
             "@typescript-eslint/no-unnecessary-type-constraint": "warn",
             "@typescript-eslint/no-unsafe-declaration-merging": "warn",
             "@typescript-eslint/no-unsafe-function-type": "warn",
-            "@typescript-eslint/no-unused-expressions": "warn",
             "@typescript-eslint/no-unused-vars": [
                 "warn",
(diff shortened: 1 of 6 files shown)


gioboa (approved):
Thanks @alumni 👏

alumni:
If we plan to merge more than a few PRs into `master`, we could also cherry pick this there, so we don't get more code like this later (although I think it's not very likely). Let me know if I should cherry pick.

gioboa:
I think we can merge it straight to master 👀

alumni:
And then sync `next` with `master`? We could do that too. 🤷🏻‍♂️ 

I'm more tempted to only do the strictly necessary for 0.3, I'd rather not put much effort into something that will be obsolete by the end of the year :)

PR #11701, opened 2025-10-01
feat: support explicit resource management in QueryRunner

alumni:
### Description of change

This PR adds support for Explicit Resource Management in `QueryRunner`. See more:
- [await using - JavaScript | MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/await_using)
- [TypeScript 5.2 Release Notes](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-5-2.html): [`using` Declarations and Explicit Resource Management](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-5-2.html#using-declarations-and-explicit-resource-management)

The `QueryRunner` will be released automatically at the end of its scope:

```typescript
async function updateSalaries() {
    await using queryRunner = dataSource.createQueryRunner();
    await queryRunner.manager.update(Employee, { level: "junior" }, { bonus: 0.2 });
    await queryRunner.manager.update(Employee, { level: "senior" }, { bonus: 0.1 });
    // no need to manually release the QueryRunner
}
```

### Pull-Request Checklist

-   [x] Code is up-to-date with the `next` branch

### DIFF
--- a/test/functional/query-runner/async-dispose.ts
+++ b/test/functional/query-runner/async-dispose.ts
@@ -1,0 +1,35 @@
+import { expect } from "chai"
+import "reflect-metadata"
+import { DataSource } from "../../../src/data-source/DataSource"
+import {
+    closeTestingConnections,
+    createTestingConnections,
+    reloadTestingDatabases,
+} from "../../utils/test-utils"
+import { EntityManager } from "../../../src"
+
+describe("query runner > async dispose", () => {
+    let dataSources: DataSource[]
+    before(async () => {
+        dataSources = await createTestingConnections({
+            entities: [__dirname + "/entity/*{.js,.ts}"],
+            enabledDrivers: ["postgres"],
+        })
+    })
+    beforeEach(() => reloadTestingDatabases(dataSources))
+    after(() => closeTestingConnections(dataSources))
+
+    it("should release query runner", () =>
+        Promise.all(
+            dataSources.map(async (dataSource) => {
+                let entityManager: EntityManager
+                {
+                    await using queryRunner = dataSource.createQueryRunner()
+                    await queryRunner.connect()
+                    entityManager = queryRunner.manager
+                }
+
+                expect(entityManager.queryRunner?.isReleased).to.be.true
+            }),
+        ))
+})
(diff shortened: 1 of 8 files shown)


test/functional/query-runner/async-dispose.ts
```diff
@@ -0,0 +1,35 @@
...
+    before(async () => {
+        dataSources = await createTestingConnections({
+            entities: [__dirname + "/entity/*{.js,.ts}"],
+            enabledDrivers: ["postgres"],
+        })
+    })
+    beforeEach(() => reloadTestingDatabases(dataSources))
+    after(() => closeTestingConnections(dataSources))
+
+    it("should release query runner", () =>
```
gioboa, line 25: Do you think it's valuable to add few cases?
eg. execute a query and verify the automatic release at the end.

alumni: I thought about doing it for the sap driver where I know how to check the pool - I think I added some tests when I migrated the pool library, could add another one there for async dispose.

Or as a unit test, spying on the QueryRunner to see if release is called. 🤷‍♂️

gioboa:
Thanks for this lovely feature @alumni

alumni:
@gioboa What do you think about the transactions that are in progress in that `QueryRunner` that is being disposed?

Right now, we only call `QueryRunner.release()`, i.e. return the client connection to the pool, at least in most drivers. Not all have a pool: SQLite just flushes the data to the disk.

Not all drivers support a transaction (e.g. MongoDB, some of the SQLite drivers), but for those that do, **we could have a transaction in progress**. So I was wondering, should we commit the transaction? Should we roll it back if not explicitly committed?

Also, some databases have save points ("nested transactions", e.g. Postgres) - how should those be handled? Should they all be committed?

I'm thinking to add:
```ts
async [Symbol.asyncDispose](): Promise<void> {
    try {
        while (this.isTransactionActive) {
            await this.commitTransaction();
        }
    } finally {
        await this.release()
    }
}
```

gioboa:
That's a hard decision, but I think your suggestion is a good one. We can probably have a voting session with the maintainer team to agree with this.

alumni:
Okay, I think I figured out that we don't need to release all savepoints, we can simply commit once. Should work in most drivers based on my understanding. Pushing a commit now.

PR #11707, opened 2025-10-02
chore(mongo): sync typings with the latest v5 version of the mongodb driver (5.9.2)

pkuczynski:
### Description of change

Sync typings with the latest v5 version of the mongodb driver (5.9.2)

### Pull-Request Checklist

## Summary by CodeRabbit

- New Features
  - Manage Atlas Search indexes with new list/create/update/drop methods and a search indexes cursor.
  - Change streams can include splitEvent metadata for large events.
  - New command cursor enables running commands with comment, maxTimeMS, and batchSize controls.
  - Optional includeResultMetadata for findOneAndUpdate/Replace/Delete to return operation metadata.

- Improvements
  - More precise update result typing, including typed upsertedId.
  - Write concern now offers readonly fields and constructor-based configuration.

- Deprecations
  - Legacy UpdateResult and several options (e.g., AddUserOptions, certain TLS/SSL shims) marked deprecated.

### DIFF
--- a/src/driver/mongodb/typings.ts
+++ b/src/driver/mongodb/typings.ts
@@ -50,6 +50,7 @@ export declare abstract class AbstractCursor<
     static readonly CLOSE: "close"
     /* Excluded from this release type: __constructor */
     get id(): Long | undefined
+    /* Excluded from this release type: isDead */
     /* Excluded from this release type: client */
     /* Excluded from this release type: server */
     get namespace(): MongoDBNamespace
@@ -80,6 +81,7 @@ export declare abstract class AbstractCursor<
      * If the iterator returns `false`, iteration will stop.
      *
      * @param iterator - The iteration callback.
+     * @deprecated - Will be removed in a future release. Use for await...of instead.
      */
     forEach(iterator: (doc: TSchema) => boolean | void): Promise<void>
     close(): Promise<void>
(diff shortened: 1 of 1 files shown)


naorpeled (approved):

alumni:
What for? The previous types where specially picked to be the ones that we use I think. There's not need to update them if we don't use them internally.

pkuczynski:
> What for? The previous types where specially picked to be the ones that we use I think. There's not need to update them if we don't use them internally.

If you look at the diff we haven't really pick the types but used all of them. The changes are mainly around deprecation js docs. Plus few small fixes to make it work in our setup. I think it's worth it.

gioboa (approved):
LGTM 💪

gioboa (changes requested):
I checked the current version of mongo and it is different 😧last week we updated the package 📦

pkuczynski:
> I checked the current version of mongo and it is different 😧last week we updated the package 📦

We need to roll it back, as this version breaks our current functionality. Version 6 of mongodb driver removed `collection.stats` (see [here](https://github.com/mongodb/node-mongodb-native/blob/ae2e037e896fbc673061da49a7c45743ae145651/etc/notes/CHANGES_6.0.0.md#collectionstats-removed)) and we expose it in our API (see [here](https://github.com/typeorm/typeorm/blob/master/src/driver/mongodb/MongoQueryRunner.ts#L474))

pkuczynski:
> We need to roll it back

Maybe we don't. I am working on a fix...

alumni:
@pkuczynski I was not able to (fully) remove our mongo types - if I remove them, the users would need to use `skipLibCheck: true` in their tsconfig, which is not default (but it is enabled in the `@tsconfig/node` presets).

I'm thinking in a monorepo setup we could figure out something (we can have the mongo driver import the mongo client + we can add some module augmentation for `EntityId`), but until that happens, we could just update to the `mongodb@7` client type (would mean updated `bson` types + removal of the `stats` method). WDYT? 

CC: @naorpeled if you are more familiar with mongo 🤷🏻

PR #11708, opened 2025-10-02
ci: control development and CI version of Node.js via .nvmrc

pkuczynski:
### Description of change

Simplifies node version management and upgrades burden. Ensures developers (contributors) use the same node version (some currently used dev dependencies require older versions of node).

### Pull-Request Checklist


gioboa (approved):
Thanks, nice improvement 🙌

naorpeled (approved):

DEVELOPER.md
```diff
@@ -50,7 +50,17 @@ cd typeorm
...
 ```
 
-## Installing NPM Modules
+## Node
+
+You should have node installed in the version described in [.nvmrc](.nvmrc).
+
+It is recommended to configure your OS to automatically switch to use this version whenever you enter project folder. This can be achieved in many ways:
+
+* [`zsh-nvm`](https://github.com/lukechilds/zsh-nvm#auto-use)
```
alumni, line 60: I would add `fnm` too, it's better than `nvm` :P

Also: it's the default on the nodejs website ;)

I would remove `avn`, it's no longer maintained.

I only briefly evaluated `asdf` and `volta`, but didn't stick to any of them. Hopefully `asdf` is working :P

.github/workflows/commit-validation.yml
```diff
@@ -49,21 +49,37 @@ jobs:
...
+      node-versions: ${{ steps.node.outputs.node-versions }}
+      base-node-version: ${{ steps.node.outputs.base-node-version }}
+    steps:
+      - uses: actions/checkout@v4
+      - id: node
+        run: |
+          BASE_NODE_VERSION=$(head -n 1 .nvmrc)
+          PREV_NODE_VERSION=$((BASE_NODE_VERSION - 2))
+
+          echo "node-versions=[$PREV_NODE_VERSION, $BASE_NODE_VERSION]" >> $GITHUB_OUTPUT
```
alumni, line 64: I'm

.github/workflows/commit-validation.yml
```diff
@@ -49,21 +49,37 @@ jobs:
           path: build/
           retention-days: 1
 
+  test-matrix:
```
alumni: I'm not sure this is okay, since we want to change the way we run the CI in the next version.

As mentioned, please don't bother with the master branch.

alumni (changes requested):
Let's not base the versions run in CI on any `.nvmrc` file.

pkuczynski:
> Let's not base the versions run in CI on any `.nvmrc` file.

Why not?

alumni:
@pkuczynski see the other comment:

> I'm not sure this is okay, since we want to change the way we run the CI in the next version.

I had to write some text to block the merge of the PR :P

Whatever is in the `master` branch is here because we wanted to introduce minimal changes. We only want to merge the existing PRs that fix bugs to this branch.

Any project setup changes should not be done here, but in the `next` branch.

DEVELOPER.md
pkuczynski: Damn! Lesson to myself: verify stuff before copying from older repo :)

Done!

.github/workflows/commit-validation.yml
pkuczynski: It does not really change much. In `next` you still test two versions, just instead of `18,20` its `20,24`. So can be easily changed to `base+4` or simply include `base` and `24` manually...

pkuczynski:
> I had to write some text to block the merge of the PR :P

🤦‍♂️ 

> Whatever is in the `master` branch is here because we wanted to introduce minimal changes. We only want to merge the existing PRs that fix bugs to this branch.

Seems that other maintainers are of a different opinion, since they already approved the PR...

> Any project setup changes should not be done here, but in the `next` branch.

This setup immediately helps people trying to come up with a bugfix. The changes in `next` are minimal and I can promise to you, I can can/will come up with PR changing the `18,20` to `20,24`...

Question is rather: is this a good or bad change? I think its good because it lowers the entry barier for the project and ensures control of the version in a single place.

alumni:
I would say, let's not add unnecessary complexity by tying the workflows to the `nvmrc`. Let's keep things simple and easy to understand.

Regarding the `nvmrc` file itself, I've been oscillating between adding and not adding it over the past year. We have `engines` in `package.json` (`fnm` uses it by default), the main issue is rather the lack of a `packageManager` field. So I have no strong opinion for or against the `nvmrc` file itself.

pkuczynski:
> I would say, let's not add unnecessary complexity by tying the workflows to the `nvmrc`. Let's keep things simple and easy to understand.

My intention is to ease setup and work for the new contributors (like me myself). When I pulled the project first time, nothing worked, not even `npm isntall`).

I would like to be in a situation where my local run is as close as possible to CI. This way we save everyone time (and cost of CI).

Controlling both via single config file (nvmrc) seems to be a very good way to achieve that. I am not sure where do you see the complexity?


> Regarding the `nvmrc` file itself, I've been oscillating between adding and not adding it over the past year. We have `engines` in `package.json` (`fnm` uses it by default), the main issue is rather the lack of a `packageManager` field. So I have no strong opinion for or against the `nvmrc` file itself.

`fnm` does not use it by default unless you specify optional setting `--resolve-engines`. More over our current setup specify 16 which does not work with some packages specified in dev dependencies.

`packageManager` field was related to v5 of node setup action. They implemented it in a poor way, but the fix is already coming.

gioboa (approved):
🙏 it looks fine to me

alumni:
Sorry, but no

OSA413:
> Controlling both via single config file (nvmrc) seems to be a very good way to achieve that. I am not sure where do you see the complexity?

The complexity comes from parsing the file and adding shifts to the node version to make it work with several versions of Node. I would save .nvmrc for new contributors, but not change the CI/CD configuration.

alumni:
> The complexity comes from parsing the file and adding shifts to the node version to make it work with several versions of Node. I would save .nvmrc for new contributors, but not change the CI/CD configuration.

Exactly my point when I requested changes. Unfortunately the PR author does not accept feedback.

pkuczynski:
> The complexity comes from parsing the file and adding shifts to the node version to make it work with several versions of Node. I would save .nvmrc for new contributors, but not change the CI/CD configuration.

I just don't see the complexity, sorry 🤷‍♂️ 

Instead of manually managing node version in CI and local run, we use single point of reference and we don't have to duplicate ourselves. Even GHA actions (`setup-node`) support such file out of the box, so it must be not only me who see a value in such approach. 

I only had to wrote really simple and small script to build the matrix... If that bothers you, I can remove that part. But I also wanted to hear some other maintainers opinion, since so far only @alumni had some considerations...

michaelbromley:
My opinion is that I prefer a number in the workflows, because I find it easier to understand. But the rest of the changes I like.

pkuczynski:
@alumni @OSA413 @michaelbromley is it better now? Or do you want `node-version-file: .nvmrc` to be gone too?

OSA413:
This one is much simpler/better, let's see what others think.

.github/workflows/commit-validation.yml
```diff
@@ -101,7 +101,7 @@ jobs:
     needs: [detect-changes, build]
     uses: ./.github/workflows/tests-windows.yml
     with:
-      node-version: 20
+      node-version: 18
```
michaelbromley: Is there a reason why this one changes from 20 to 18?

pkuczynski: No, not really. Accidential change. Let me revert it...

PR #11114, opened 2024-10-26
fix: do not create junction table metadata when it already exists

ragrag:
### Description of change

After the introduction of entityMetadatasMap in https://github.com/typeorm/typeorm/pull/9914/commits/f12a95cfee1965744cd8772c735a409e01035ad5 an issue arises where a junction table defined by user in entities would have multiple metadata entries (one for entity manually defined by the user, and one for the auto-created metadata by typeorm). This would cause problems as the junction table metadata would override the user defined metadata in entityMetadatasMap.

This PR ensures metadata for junction tables already created by users do not exist twice in entityMetadatas and thus avoiding conflicts when its resolved from entityMetadatasMap. This works by always using the user defined entity metadata if it exists.

This bug exists from version 3.13 and blocks upgrading if junction tables are used. 

### Fixes:
https://github.com/typeorm/typeorm/issues/10412 (includes repro)
https://github.com/typeorm/typeorm/issues/10463 
https://github.com/typeorm/typeorm/issues/10643

Apart from those issues, as a side effect of this fix, other various issues related to duplicate sql statements while generating migrations in conjunction with user defined junction tables, for example https://github.com/typeorm/typeorm/issues/7955, https://github.com/typeorm/typeorm/issues/11126 and similar issues are fixed or partially fixed

### Pull-Request Checklist

- [x] Code is up-to-date with the `master` branch
- [x] `npm run format` to apply prettier formatting
- [x] `npm run test` passes with this change
- [x] This pull request links relevant issues as `Fixes #0000`
- [x] There are new or updated unit tests validating the change
- [ ] Documentation has been updated to reflect this change
- [x] The new commits follow conventions explained in [CONTRIBUTING.md](https://github.com/typeorm/typeorm/blob/master/CONTRIBUTING.md)

## Summary by CodeRabbit

- **Bug Fixes**
  - Resolved an issue where duplicate metadata could be created for user-defined junction tables in many-to-many relationships.

- **Tests**
  - Added new test cases to verify correct handling of user-defined junction tables and ensure no duplicate metadata is generated.

- **New Features**
  - Introduced entity schemas and relationships for User, Document, and DocumentRelation to support many-to-many associations via a custom junction table.

PR #11720, opened 2025-10-10
fix: prefer user-defined junction metadata over auto-generated

### LINKED ISSUE #11637: Many-to-many relationships broken when user-defined junction entity exists (0.3.26 regression)
### Issue description

TypeORM 0.3.26 appears to have changed behavior for many-to-many relationships when user-defined junction entities exist. The relationship data is not being persisted correctly, with junction table rows containing NULL values instead of the expected foreign key relationships.

### Expected Behavior

In TypeORM <=0.3.25, junction table should be populated with proper foreign key values:
```sql
INSERT INTO "vet_specialty"("vet_id", "specialty_name") VALUES ($1, $2), ($3, $4) -- PARAMETERS: [1,"dogs",1,"cats"]

```

Junction table contains relationship data, queries return related entities.

```bash
> npm install typeorm@0.3.25 --no-save && npm test

changed 1 package, and audited 140 packages in 1s

found 0 vulnerabilities

> type-orm-repro@1.0.0 test
> ts-node test.ts

query: SELECT version()
query: SELECT * FROM current_schema()
query: SELECT "Specialty"."name" AS "Specialty_name" FROM "specialty" "Specialty" WHERE "Specialty"."name" IN ($1, $2) -- PARAMETERS: ["dogs","cats"]
query: START TRANSACTION
query: INSERT INTO "specialty"("name") VALUES ($1), ($2) -- PARAMETERS: ["dogs","cats"]
query: COMMIT
query: SELECT "Specialty"."name" AS "Specialty_name" FROM "specialty" "Specialty" WHERE "Specialty"."name" IN ($1, $2) -- PARAMETERS: ["dogs","cats"]
query: START TRANSACTION
query: INSERT INTO "vet"("name") VALUES ($1) RETURNING "id" -- PARAMETERS: ["Carlos Salazar"]
query: INSERT INTO "vet_specialty"("vet_id", "specialty_name") VALUES ($1, $2), ($3, $4) -- PARAMETERS: [1,"dogs",1,"cats"]
query: COMMIT
query: SELECT DISTINCT "distinctAlias"."Vet_id" AS "ids_Vet_id" FROM (SELECT "Vet"."id" AS "Vet_id", "Vet"."name" AS "Vet_name", "Vet__Vet_specialties"."name" AS "Vet__Vet_specialties_name" FROM "vet" "Vet" LEFT JOIN "vet_specialty" "Vet_Vet__Vet_specialties" ON "Vet_Vet__Vet_specialties"."vet_id"="Vet"."id" LEFT JOIN "specialty" "Vet__Vet_specialties" ON "Vet__Vet_specialties"."name"="Vet_Vet__Vet_specialties"."specialty_name" WHERE (("Vet"."id" = $1))) "distinctAlias" ORDER BY "Vet_id" ASC LIMIT 1 -- PARAMETERS: [1]
query: SELECT "Vet"."id" AS "Vet_id", "Vet"."name" AS "Vet_name", "Vet__Vet_specialties"."name" AS "Vet__Vet_specialties_name" FROM "vet" "Vet" LEFT JOIN "vet_specialty" "Vet_Vet__Vet_specialties" ON "Vet_Vet__Vet_specialties"."vet_id"="Vet"."id" LEFT JOIN "specialty" "Vet__Vet_specialties" ON "Vet__Vet_specialties"."name"="Vet_Vet__Vet_specialties"."specialty_name" WHERE ( (("Vet"."id" = $1)) ) AND ( "Vet"."id" IN (1) ) -- PARAMETERS: [1]
Vet specialties count: 2
Expected: 2, Actual: 2

> psql -d postgres -c "SELECT * FROM vet_specialty"
 id | vet_id | specialty_name
----+--------+----------------
  1 |      1 | dogs
  2 |      1 | cats
(2 rows)
```

### Actual Behavior

In TypeORM >=0.3.26, junction table gets rows with NULL/DEFAULT values:
```sql
INSERT INTO "vet_specialty"("vet_id", "specialty_name") VALUES (DEFAULT, DEFAULT), (DEFAULT, DEFAULT) RETURNING "id"
```

Junction table contains empty rows, queries return no related entities.

```bash
> npm install typeorm@0.3.26 --no-save && npm test

changed 1 package, and audited 140 packages in 933ms

found 0 vulnerabilities

> type-orm-repro@1.0.0 test
> ts-node test.ts

query: SELECT version()
query: SELECT * FROM current_schema()
query: SELECT "Specialty"."name" AS "Specialty_name" FROM "specialty" "Specialty" WHERE "Specialty"."name" IN ($1, $2) -- PARAMETERS: ["dogs","cats"]
query: START TRANSACTION
query: INSERT INTO "specialty"("name") VALUES ($1), ($2) -- PARAMETERS: ["dogs","cats"]
query: COMMIT
query: SELECT "Specialty"."name" AS "Specialty_name" FROM "specialty" "Specialty" WHERE "Specialty"."name" IN ($1, $2) -- PARAMETERS: ["dogs","cats"]
query: START TRANSACTION
query: INSERT INTO "vet"("name") VALUES ($1) RETURNING "id" -- PARAMETERS: ["Carlos Salazar"]
query: INSERT INTO "vet_specialty"("vet_id", "specialty_name") VALUES (DEFAULT, DEFAULT), (DEFAULT, DEFAULT) RETURNING "id"
query: COMMIT
query: SELECT DISTINCT "distinctAlias"."Vet_id" AS "ids_Vet_id" FROM (SELECT "Vet"."id" AS "Vet_id", "Vet"."name" AS "Vet_name", "Vet__Vet_specialties"."name" AS "Vet__Vet_specialties_name" FROM "vet" "Vet" LEFT JOIN "vet_specialty" "Vet_Vet__Vet_specialties" ON "Vet_Vet__Vet_specialties"."vet_id"="Vet"."id" LEFT JOIN "specialty" "Vet__Vet_specialties" ON "Vet__Vet_specialties"."name"="Vet_Vet__Vet_specialties"."specialty_name" WHERE (("Vet"."id" = $1))) "distinctAlias" ORDER BY "Vet_id" ASC LIMIT 1 -- PARAMETERS: [1]
query: SELECT "Vet"."id" AS "Vet_id", "Vet"."name" AS "Vet_name", "Vet__Vet_specialties"."name" AS "Vet__Vet_specialties_name" FROM "vet" "Vet" LEFT JOIN "vet_specialty" "Vet_Vet__Vet_specialties" ON "Vet_Vet__Vet_specialties"."vet_id"="Vet"."id" LEFT JOIN "specialty" "Vet__Vet_specialties" ON "Vet__Vet_specialties"."name"="Vet_Vet__Vet_specialties"."specialty_name" WHERE ( (("Vet"."id" = $1)) ) AND ( "Vet"."id" IN (1) ) -- PARAMETERS: [1]
Vet specialties count: 0
Expected: 2, Actual: 0

> psql -d postgres -c "SELECT * FROM vet_specialty"
 id | vet_id | specialty_name
----+--------+----------------
  1 |        |
  2 |        |
(2 rows)
```

### Steps to reproduce

Minimal reproducer available at [danielfrankcom/typeorm-repro](https://github.com/danielfrankcom/typeorm-repro).

The reproducer includes a simple many-to-many relationship with a user-defined junction entity. Running the test with TypeORM 0.3.25 vs 0.3.26 shows the different behavior.

If the `vet_specialty` table is created in `migration.sql` but the `VetSpecialty` entity is not defined in `entities.ts`, the junction table works and produces the expected result. The existence of the junction table entity seems to cause the difference in behavior.

### My Environment

| Operating System    |    MacOS      |
| Node.js version     | v22.18.0  |
| Typescript version  | 5.9.2 |

### Additional Context

This behavior change appears to be related to PR #11114, which addressed duplicate metadata issues for user-defined junction tables. The change may have affected how junction metadata is created when explicit junction entities are present.

- [ ] aurora-mysql
- [ ] aurora-postgres
- [ ] better-sqlite3
- [ ] cockroachdb
- [ ] cordova
- [ ] expo
- [ ] mongodb
- [ ] mysql
- [ ] nativescript
- [ ] oracle
- [x] postgres
- [ ] react-native
- [ ] sap
- [ ] spanner
- [ ] sqlite
- [ ] sqlite-abstract
- [ ] sqljs
- [ ] sqlserver

ashwintemkar:
### Description of change
**Fix related issue: #11637**

So when a user defined junction entity exists, **0.3.26** regressed many to many persistence and inserted `DEFAULT/NULL` into the junction table FKs. There the   root cause was metadata resolution favoring the **auto generated** junction metadata over the **user defined** junction entity when both were present (via how `entityMetadatasMap` was consulted).

So now I have updated `DataSource` lookup logic to **prefer the user defined junction metadata**, restoring correct FK values in the junction table and keeping `entityMetadatasMap` conflict-free.

**Tests included**

* Added a regression test reproducing **#11637** and asserting proper FK values with a custom junction entity.
* Restore the earlier **#11114** regression tests to confirm `EntitySchema` based junction tables still expose relations correctly.

**Verification**

* I Ran `npm test -- --grep "junction"` on Postgres.
* Confirmed junction rows contain the expected FK values in the #11637 scenario.

**Notes**

* No behavior change for projects that use only auto generated junction tables.
* No documentation impact expected.

### Pull-Request Checklist

* [x] Code is up-to-date with the `master` branch
* [x] This pull request links relevant issues as **Fixes #11637**
* [x] There are new or updated unit tests validating the change
* [ ] Documentation has been updated to reflect this change *(N/A)*

## Summary by CodeRabbit

* **New Features**
  * Improved handling of user-defined junction tables for many-to-many relations, ensuring correct metadata resolution.

* **Bug Fixes**
  * Prevents errors when querying relations that use custom junction table entities.

* **Tests**
  * Added functional tests covering user-defined junction table scenarios and persistence/metadata correctness.


ashwintemkar:
@sgarner @ragrag pardon me for tagging, 

I have opened this Draft PR to check whether anything more is needed out of the fix I have put up, please let me know if this is ready for review, as the core problem is fixed

test/functional/entity-metadata/entity-metadata-builder.ts
```diff
@@ -0,0 +1,49 @@
...
+import { documentRelationEntitySchema } from "./entity/junction-table/entities"
+
+describe("entity-metadata-builder > build", () => {
+    let dataSources: DataSource[]
+
+    before(
+        async () =>
+            (dataSources = await createTestingConnections({
+                entities: [__dirname + "/entity/junction-table/*{.js,.ts}"],
+                enabledDrivers: ["sqlite"],
```
ragrag: don't forget to remove this, looks like it was forgotten and left before

src/data-source/DataSource.ts
```diff
@@ -666,44 +666,44 @@ export class DataSource {
...
+        const findMatching = (
+            predicate: (metadata: EntityMetadata) => boolean,
+        ): EntityMetadata | undefined => {
+            let junctionCandidate: EntityMetadata | undefined
+            for (const [, metadata] of this.entityMetadatasMap) {
+                if (!predicate(metadata)) continue
+                if (!metadata.isJunction) return metadata
+                if (!junctionCandidate) junctionCandidate = metadata
             }
+            return junctionCandidate
```
ragrag: could you add some comments to explain why this is being done?

src/data-source/DataSource.ts
```diff
@@ -733,11 +733,22 @@ export class DataSource {
...
+            const existing = entityMetadatasMap.get(metadata.target)
+            if (!existing) {
+                entityMetadatasMap.set(metadata.target, metadata)
+                continue
+            }
+
+            if (existing.isJunction && !metadata.isJunction) {
+                entityMetadatasMap.set(metadata.target, metadata)
+            }
+        }
```
ragrag: same, i suggest to add a comment here to explain that user defined junction tables takes precedence in the entityMetadatasMap registration

ragrag (commented):
overall i think this approach looks cogent and should behave more consistently than the previous implementation 🙏🏻

ashwintemkar:
Thanks for pointing those out! @ragrag  , I pulled the stray junction-table regression test back out and added inline comments explaining why we prefer user-defined metadata over the synthetic junction entry. Let me know if you’d like any more detail there.

ragrag:
> Thanks for pointing those out! @ragrag , I pulled the stray junction-table regression test back out and added inline comments explaining why we prefer user-defined metadata over the synthetic junction entry. Let me know if you’d like any more detail there.

sorry i meant to only remove the ```enabledDrivers: ['sqlite']``` part (the line my comment was on initially) we should keep the tests for sure, rest lgtm!

ashwintemkar:
so dropping the explicit `enabledDrivers ` override I hit the Postgres-only failure I saw `constraint "... does not exist"`, which meant the auto-generated junction metadata was still hanging around alongside the user-defined entity. That’s why I extended `EntityMetadataBuilder`

We now merge the auto-generated junction metadata back into the user-defined entity, reuse its columns, and drop the duplicate entry. Without that, Postgres tries to create the same PK/FKs twice and the regression resurfaces.
Relations that were wired to the synthetic junction metadata are re-pointed to the user entity so many-to-many inserts keep their FK values `vet_specialty, feed_resources_categories, etc`.
Existing coverage for the #11114 scenario is still passing, and the new regression test for #11637 passes on Postgres with the fix `npm test -- --grep "junction"`.

So the earlier logic from commit 95baf98c5 is still there, and the new builder hook just finishes the job by reconciling the actual metadata objects.

ragrag:
> We now merge the auto-generated junction metadata back into the user-defined entity, reuse its columns, and drop the duplicate entry. Without that, Postgres tries to create the same PK/FKs twice and the regression resurfaces. Relations that were wired to the synthetic junction metadata are re-pointed to the user entity so many-to-many inserts keep their FK values `vet_specialty, feed_resources_categories, etc`. Existing coverage for the #11114 scenario is still passing, and the new regression test for #11637 passes on Postgres with the fix `npm test -- --grep "junction"`.

could we perhaps do this merging once for all conflicting junction metadatas during metaDatasMap registration (or merging such tables before the metadata generation step)? that would be a much simpler single change, you think something like that could work?

e.g:

```ts
entityMetadatasMap: new Map(
    withDeduplicatedJunctionTables(entityMetadatas).map((metadata) => [metadata.target, metadata]),
),
```

also it could be worth investigating if typeorm already has some internal method for metadata merging or table merging that we could instead of ```mergeWithUserDefinedJunctions```

ragrag:
throwing other ideas i feel that might solve this more straightforwardly. originally ```entityMetaDatasMap``` was introduced in https://github.com/typeorm/typeorm/commit/f12a95cfee1965744cd8772c735a409e01035ad5 to make metadata lookups faster, we could potentially explore some variants that would maintain the fast lookups that have potential to work:

1 - use the previous metadata logic that loops over ```entityMetadatas``` and adding a cache on top in favor of ```entityMetadatasMap```, so the initial lookup will have similar result pre 3.13 and consequent lookups will be resolved from the cache to the correct metadata. essentially having entityMetadatasMap but building it in a lazy fashion (imho the most straight forward)

2 - in ```findMetadata```, if entity found in ```entityMetadatasMap``` is a junction entity (or an entity we know is a potential duplicate junction entity from some precalculation we did) do not return it and, fallback to old logic that loops over ```entityMetadatas``` from the commit above

3 - gathering a set of user defined tables that have a junction pair in ```entityMetadatas```, in ```findMetadata``` check if ```target``` is part of that set and if that's the case fallback to old logic that loops over ```entityMetadatas``` from the commit above

4- wild idea and not sure if it'll work but constructing ```entityMetadatasMap``` from the reverse of ```entityMetadatas```, my guess is that junction and virtual tables are pushed to the end of ```entityMetadatas``` causing the map overriding to happen

theoretically all of those solutions should maintain the same behaviour as before the regression introduced in 3.13 (from the commit above) and would avoid potentially error prone metadata merging, let me know what you think

ashwintemkar:
Hey @ragrag thanks for the detailed suggestions! 

I moved the reconciliation to metadata registration: 

1. we now deduplicate the auto-generated junction metadata before `entityMetadatasMap ` is built
2. reuse the junction FK definitions on the user entity
3. repoint the relations to the user metadata. 

That keeps the map fast, avoids the extra builder hook, and resolves the Postgres constraint/index failures. 

Happy to iterate further!! if you spot anything else.


also for your reference 

> throwing other ideas i feel that might solve this more straightforwardly. originally `entityMetaDatasMap` was introduced in [f12a95c](https://github.com/typeorm/typeorm/commit/f12a95cfee1965744cd8772c735a409e01035ad5) to make metadata lookups faster, we could potentially explore some variants that would maintain the fast lookups that have potential to work:
> 
> 1 - use the previous metadata logic that loops over `entityMetadatas` and adding a cache on top in favor of `entityMetadatasMap`, so the initial lookup will have similar result pre 3.13 and consequent lookups will be resolved from the cache to the correct metadata. essentially having entityMetadatasMap but building it in a lazy fashion (imho the most straight forward)
> 
> 2 - in `findMetadata`, if entity found in `entityMetadatasMap` is a junction entity (or an entity we know is a potential duplicate junction entity from some precalculation we did) do not return it and, fallback to old logic that loops over `entityMetadatas` from the commit above
> 
> 3 - gathering a set of user defined tables that have a junction pair in `entityMetadatas`, in `findMetadata` check if `target` is part of that set and if that's the case fallback to old logic that loops over `entityMetadatas` from the commit above
> 
> 4- wild idea and not sure if it'll work but constructing `entityMetadatasMap` from the reverse of `entityMetadatas`, my guess is that junction and virtual tables are pushed to the end of `entityMetadatas` causing the map overriding to happen
> 
> theoretically all of those solutions should maintain the same behaviour as before the regression introduced in 3.13 (from the commit above) and would avoid potentially error prone metadata merging, let me know what you think

A lazy cache (option 1) and the targeted fallbacks (options 2 & 3) still leave duplicate metadata in `entityMetadatas`, so the schema builder continued to create duplicate PK/FK definitions in Postgres.

Reversing the insertion order (option 4) depends on incidental ordering and still risks selecting the wrong metadata.

Hence the new deduplication step keeps the fast lookups from `entityMetadatasMap`, avoids metadata merging inside the builder, and gives us a single, predictable place to reconcile the two metadata sources.

ragrag:
Amazing work!

> Hence the new deduplication step keeps the fast lookups from `entityMetadatasMap`, avoids metadata merging inside the builder, and gives us a single, predictable place to reconcile the two metadata sources.

just wondering if we still need all the other changes done now (the change in ```findMetadata```, and ```buildMetadata```), shouldn't the metadata merging with ```deduplicateJunctionEntityMetadatas``` alone pre-registration with the old way fix the issue? 

btw did you manage to find any typeorm internals that do the metadata merging?

ashwintemkar:
> just wondering if we still need all the other changes done now (the change in `findMetadata`, and `buildMetadata`),

so I initally thought that the ideas were good when this https://github.com/typeorm/typeorm/pull/11720#issuecomment-3408085455 was suggested but then I thought `deduplicateJunctionEntityMetadatas` would bring the edge here. 

> shouldn't the metadata merging with `deduplicateJunctionEntityMetadatas` alone pre-registration with the old way fix the issue?

and you’re right that the heavy lifting happens in the new `deduplicateJunctionEntityMetadatas` step. I reverted the tweaks to `findMetadata` and to the `entityMetadatasMap` construction so they’re back to the exact logic we had with the synthetic junction metadata removed before the map is built, those extra guard rails aren’t needed.



> btw did you manage to find any typeorm internals that do the metadata merging?

And also I didn’t find an existing helper in TypeORM that performs this reconciliation. I checked the metadata builders and schema helpers, but nothing exposed the junction metadata we needed to reuse.

so currently 
`deduplicateJunctionEntityMetadatas `runs right after `buildEntityMetadatas`, strips out the auto generated junction entries, and rewires any relations that were pointing at them and because the duplicates are gone, `findMetadata`’s original implementation is returning the correct user defined entity without the additional filtering.

for now I think this can go in as the original `findMetadata `logic behaves exactly like pre‑3.13, and both the #11114 functional coverage and the #11637 regression test pass `npm test -- --grep "entity-metadata-builder"` and  `npm test -- --grep "junction"`

src/data-source/DataSource.ts
```diff
@@ -666,6 +668,13 @@ export class DataSource {
         const metadataFromMap = this.entityMetadatasMap.get(target)
         if (metadataFromMap) return metadataFromMap
 
+        /**
+         * Sharing table names between user entities and auto-created junctions is
+         * perfectly legal, but it means lookups can produce more than one match.
+         * This helper makes sure we hand back the user-defined metadata whenever
+         * it is available, falling back to the junction metadata only when it is
+         * the sole candidate left.
+         */
```
ragrag: probably this comment is now extraneous and the one on top of deduplicateJunctionEntityMetadatas should suffice

ashwintemkar: oops just sending a commit in

ragrag:
> so currently `deduplicateJunctionEntityMetadatas `runs right after `buildEntityMetadatas`, strips out the auto generated junction entries, and rewires any relations that were pointing at them and because the duplicates are gone, `findMetadata`’s original implementation is returning the correct user defined entity without the additional filtering.

coool, seems like this covers all the edges so looks good from my side, perhaps other maintainers are more familiar in the matter of metadata merging

ashwintemkar:
> > so currently `deduplicateJunctionEntityMetadatas `runs right after `buildEntityMetadatas`, strips out the auto generated junction entries, and rewires any relations that were pointing at them and because the duplicates are gone, `findMetadata`’s original implementation is returning the correct user defined entity without the additional filtering.
> 
> coool, seems like this covers all the edges so looks good from my side, perhaps other maintainers are more familiar in the matter of metadata merging

alright so should I mark this ready for review?, I mean turn into actual PR from draft?

ragrag:
> alright so should I mark this ready for review?, I mean turn into actual PR from draft?

sure feel free to!

src/data-source/DataSource.ts
```diff
@@ -709,6 +711,157 @@ export class DataSource {
...
+                match.width = column.width
+                match.precision = column.precision
+                match.scale = column.scale
+                match.charset = column.charset
+                match.collation = column.collation
+                match.zerofill = column.zerofill
+                match.unsigned = column.unsigned
+                match.enum = column.enum
+                match.enumName = column.enumName
+
```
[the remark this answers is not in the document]
ashwintemkar: Thanks for catching that! I’ve removed the structural overrides so we only copy the wiring bits (`referencedColumn `/ `relationMetadata`) and leave any user-defined column options untouched.

ashwintemkar:
@ragrag any idea by when this is merging in ?

ragrag (approved):

ashwintemkar:
Hi @gioboa do my changes check out? and can this pr be merged in?

gioboa (approved):
Thanks for your help @ashwintemkar 
LGTM ✅

ashwintemkar:
Hey @gioboa and @ragrag , since @sgarner and @alumni might take some time to review according to their availability, as all checks have passed, I had a small request, as maintainers could y'all add a hacktoberfest accepted tag onto this PR, since I have participated in the hacktoberfest, I think adding the tag might count as contribution, just want to see if that works.

ashwintemkar:
thanks a tonne @gioboa

gioboa (dismissed):
The test file must have `.test.ts` extension to be considered.

ashwintemkar:
> The test file must have `.test.ts` extension to be considered.

sure @gioboa I will look into the changes needed

gioboa (commented):
I fixed the test file name.

ashwintemkar:
> I fixed the test file name.

okay, so is this pr now waiting only for approvals right?

gioboa (commented):
> okay, so is this pr now waiting only for approvals right?

That's correct. the PR now is ready with a green pipeline 💪

ragrag (approved):

michaelbromley:

```
git fetch origin
```

Issue #11155, opened 2024-11-28
Support for MySQL CHECK Constraints


yeganemehr:
### Issue description

MySQL introduced support for CHECK constraints starting from version 8.0.16. but TypeORM does not support creating or managing CHECK constraints in MySQL databases.

### Expected Behavior

When defining CHECK constraints in the entity definitions, TypeORM should generate the appropriate SQL statements for adding CHECK constraints during migrations.

### Actual Behavior

Currently, TypeORM ignores CHECK constraints in MySQL. Attempts to define or synchronize them are ineffective, and migrations do not include the required SQL statements.

### Steps to reproduce

1. Use MySQL version 8.0.16 or later.
2. Define an entity with a CHECK constraint, e.g.:
```ts
import { Entity, Column, Check, PrimaryGeneratedColumn } from 'typeorm';

@Entity()
@Check("a>b")
export class User {
    @PrimaryGeneratedColumn()
    id: number;

    @Column('int')
    a: number;

    @Column('int')
    b: number;
}
```
3.Generate and run migrations.
4. Observe that no `CHECK` constraint is created in the database.

### My Environment

| Dependency          | Version  |
| ---                 | ---      |
| Operating System    | Ubuntu 22.04 |
| Node.js version     | 20.10.0  |
| Typescript version  | 5.1.3  |
| TypeORM version     | 0.3.20  |

### Relevant Database Driver(s)

- [ ] aurora-mysql
- [ ] aurora-postgres
- [ ] better-sqlite3
- [ ] cockroachdb
- [ ] cordova
- [ ] expo
- [ ] mongodb
- [X] mysql
- [ ] nativescript
- [ ] oracle
- [ ] postgres
- [ ] react-native
- [ ] sap
- [ ] spanner
- [ ] sqlite
- [ ] sqlite-abstract
- [ ] sqljs
- [ ] sqlserver

### Are you willing to resolve this issue by submitting a Pull Request?

Yes, I have the time, and I know how to start.

1345822:
It is necessary to deeply study the code logic of TypeORM about database constraint processing, especially MySQL, find out the root cause that it can't correctly generate and process CHECK constraint SQL statements, and then make targeted code modification and improvement. At the same time, it is necessary to ensure that the changes have been fully tested, including functional testing and compatibility testing in different scenarios, so as to ensure that new problems will not be introduced.

Issue #6103, opened 2020-05-20
migration runner refused to create check constraint in mysql 8.0.16


amjedomar:
### Issue type:

[ ] question
[X] bug report
[ ] feature request
[ ] documentation issue

### Database system/driver:

[ ] `cordova`
[ ] `mongodb`
[ ] `mssql`
[X] `mysql` / `mariadb`
[ ] `oracle`
[ ] `postgres`
[ ] `cockroachdb`
[ ] `sqlite`
[ ] `sqljs`
[ ] `react-native`
[ ] `expo`

### TypeORM version:

[ ] `latest`
[ ] `@next`
[X] `0.2.25` (or put your version here)

### Steps to reproduce the problem

Execute the following  migration

```ts
import {MigrationInterface, QueryRunner, TableCheck} from 'typeorm';

export class AlterSomeTable1589965831962 implements MigrationInterface {

  public async up(queryRunner: QueryRunner): Promise<any> {
    await queryRunner.createCheckConstraint('some_table', new TableCheck({
      name: 'ck_value',
      expression: "`value` < 5"
    }));
  }

  public async down(queryRunner: QueryRunner): Promise<any> {

  }

}
```

Then you will get the following error
```txt
Error during migration run:
Error: MySql does not support check constraints.
```

However MySQL [8.0.16](https://dev.mysql.com/doc/refman/8.0/en/create-table-check-constraints.html) support check constriant

amjedomar:
@pleerock can you take a look at this issue please

imnotjames:
Duplicate of #4148

imnotjames:
TLDR: Check constraints on MySQL aren't implemented yet.  Feel free to push along the PR proposed as part of that issue

JeromeMaquoi:
Are there any news? Are the check constraints still not implemented?

Issue #5904, opened 2020-04-18
@Check does not work


lueenavarro:
**Issue type:**

[ ] question
[ x] bug report
[ ] feature request
[ ] documentation issue

**Database system/driver:**

[ ] `cordova`
[ ] `mongodb`
[ ] `mssql`
[x ] `mysql` / `mariadb`
[ ] `oracle`
[ ] `postgres`
[ ] `cockroachdb`
[ ] `sqlite`
[ ] `sqljs`
[ ] `react-native`
[ ] `expo`

**TypeORM version:**

[ x] `latest`
[ ] `@next`
[ ] `0.x.x` (or put your version here)

@Check decorator does not work. I adds nothing to the entity.

```
import { Check, Column, Entity, PrimaryColumn } from 'typeorm';

@Entity({ name: 'sp_term' })
@Check(`spTermId = 1`)
export class SpTerm {
  @PrimaryColumn({ type: 'int', default: () => `1`, nullable: false })
  spTermId: number;

  @Column()
  retailPrice: number;

  @Column()
  sellingPrice1: number;

  @Column()
  sellingPrice2: number;

  @Column()
  sellingPrice3: number;

  @Column()
  sellingPrice4: number;
}
```

I am creating a table that should have one row only. So I used @check to make sure that only one row exists.

However I can still add other rows with spTermId of 2 and when I checked the create code of the table.

```
CREATE TABLE `sp_term` (
	`spTermId` INT(11) NOT NULL DEFAULT '1',
	`retailPrice` INT(11) NOT NULL,
	`sellingPrice1` INT(11) NOT NULL,
	`sellingPrice2` INT(11) NOT NULL,
	`sellingPrice3` INT(11) NOT NULL,
	`sellingPrice4` INT(11) NOT NULL,
	PRIMARY KEY (`spTermId`) USING BTREE
)
COLLATE='latin1_swedish_ci'
ENGINE=InnoDB
;
```

As you can see there's no anything about the check constraint.

iamjoeker:
According to the documentation, the `Check` decorator is not supported for MySQL. https://typeorm.io/#decorator-reference/check

Bast1onCZ:
`Check` constraint is supported by mariadb from version 10.2.1 (see: https://mariadb.com/kb/en/constraint/).

Could someone add it to typeorm please? :)

Issue #4148, opened 2019-05-16
MySQL 8.0.16: Now supports CHECK-Constraints


FlorianPfisterer:
**Issue type:**
[ ] question
[ ] bug report
[x] feature request
[ ] documentation issue

**Database system/driver:**
[ ] `cordova`
[ ] `mongodb`
[x] `mssql`
[ ] `mysql` / `mariadb`
[ ] `oracle`
[ ] `postgres`
[ ] `cockroachdb`
[ ] `sqlite`
[ ] `sqljs
[ ] `react-native`
[ ] `expo`

**TypeORM version:**
[ ] `latest`
[ ] `@next`
[x] `0.2.17`

**From the [official MySQL docs](https://dev.mysql.com/doc/refman/8.0/en/create-table-check-constraints.html)**:
> As of MySQL 8.0.16, CREATE TABLE permits the core features of table and column CHECK constraints, for all storage engines.

**However**, in this test case it still says MySQL does not support CHECK constraints, and if I add a `@Check('...')` before my `@Entity()` class, no CHECK constraints are actually added:
https://github.com/typeorm/typeorm/blob/79110d93cf96ad5d2527ced7441df06e2fd8baf1/test/functional/query-runner/create-check-constraint.ts#L23

I would greatly appreciate supporting this feature in MySQL >= 8.0.16 too. Thanks a lot!

vmuresanu:
Hi, I've just ran into the same question. Are there any updates regarding this request?

vmuresanu:
Guys, I figured out why not to contribute and to add support for myself :). It works locally on my machine, need to cover with tests and make PR.

JeromeMaquoi:
Any news? Does TypeORM still not support check constraint for mysql database?

smith-xyz:
> Any news? Does TypeORM still not support check constraint for mysql database?

Still no support for MySql/MariaDB. I might take a crack at it and see why the initial PR broke the tests.

MariaDB started supporting check constrains as of v10.2.1. Looks like MariaDB wouldn't throw any errors if you did include the check constraint syntax. I think MySQL did the same if I'm not mistaken? This tells me that regression issues would be minimal if so.

PR #11724, opened 2025-10-15
feat(mysql/mariadb): add CHECK constraint support

ZBAGI:
### Description of change

Added version-gated CHECK constraint support to the MySQL driver. [MySQL has supported user defined CHECK constraints since version 8.0.16](https://dev.mysql.com/blog-archive/mysql-8-0-16-introducing-check-constraint/) and [MariaDB since version 10.2.1](https://mariadb.com/docs/release-notes/community-server/old-releases/release-notes-mariadb-10-2-series/mariadb-1021-release-notes). Previously, TypeORM always threw `MySQL does not support check constraints.` for any attempt to create or modify a check, regardless of the MySQL version.

Fixes: #4148
Fixes: #6103
Fixes: #11155
Fixes: #5904

**Verified:**
- Manual entity sync with `@Check()` produces `ALTER TABLE ... ADD CONSTRAINT ... CHECK(...)` on supported versions.
- Changing the CHECK expression yields migration diff of DROP + ADD constraint.
- Older versions of MySQL path remains unchanged (flag false) and still throws to avoid implying unsupported enforcement.

### Pull-Request Checklist

- [x] This pull request links relevant issues as `Fixes #00000` (replace or mark N/A)
- [x] There are new or updated unit/functional tests validating creation, idempotency, and change detection (or mark N/A)
- [x] Documentation has been updated to reflect this change (or will be in a follow-up PR)

## Summary by CodeRabbit

* **New Features**
  * CHECK constraint support enabled for MySQL (8.0.16+) and MariaDB (10.2.1+); schema operations now respect server capability and return a clear error when unsupported.

* **Documentation**
  * Updated docs to clarify MySQL/MariaDB CHECK support, older-server behavior (parsed but ignored), and version requirements.

* **Tests**
  * Test entity discovery expanded to common + driver-specific paths; MySQL-family tests now run when the server reports CHECK support and many driver-specific test entities added.


ZBAGI:
# Test problem
[Test that uses: `@Check(`"version" < 999`)`](https://github.com/typeorm/typeorm/pull/11724/commits/520e3bc6dbab06e364a3e6fdab3e8367a13db8f2#diff-307e99e1c784812b24475fe805d037f4253666320625decc35b9361d3581ccad) is not portable across drivers. It works on engines that treat double quotes as identifier quotes (Postgres, Oracle, etc.), but it breaks on MySQL / MariaDB because they treat double quotas as a string literal, not an identifier. So `"version"` becomes the literal `'version'`, the server stores something like `_utf8mb4'version' < 999`, and any later round‑trip (schema diff, drop/revert) blows up or becomes semantically wrong. Simply removing quotes and relying on an unquoted identifier did not solve it cleanly in test matrix (caused other inconsistencies).

Previously this was not an issue because MySQL/MariaDB did not enforce @Check, so the test never actually ran.

### Potential fixes
- Split entity definitions per driver (at least for query‑runner / functional tests) so each variant uses that dialect's normal identifier quoting.  _(This is the approach I'd choose unless there are objections)_
- `driver.escape` in decorators: Allow @Check(d => `${d.escape('version')} < 999`) so the expression is resolved with the active driver's quoting rules at schema build time.
- Token system: @Check('${id:version} < 999') where the framework replaces `${id:...}` using driver rules?
- Metadata normalization (driver layer): For MySQL/MariaDB only, detect CHECK_CLAUSE patterns like _utf8mb4'col' and map them back to identifiers if 'col' matches a column name.

Please let me know if any of the solutions is sufficient, or if you have a better idea.

smith-xyz:
These changes look great—looking forward to seeing this merged! Not a maintainer, but I think your proposed option 1 is consistent with other tests that do something similar. I assume it would look something like:

```ts
let checkExpr1: string
let checkExpr2: string
let checkExpr3: string

if (DriverUtils.isPostgres(connection.driver) || DriverUtils.isOracle(connection.driver)) {
    checkExpr1 = `"name" <> 'asd' AND "description" <> 'test'`
    checkExpr2 = `("id" < 0 AND "version" < 9999) OR ("id" > 9999 AND "version" < 888)`
    checkExpr3 = `"id" + "version" > 0`
} else if (DriverUtils.isMySQLFamily(connection.driver)) {
    checkExpr1 = `name <> 'asd' AND description <> 'test'`
    checkExpr2 = `(id < 0 AND version < 9999) OR (id > 9999 AND version < 888)`
    checkExpr3 = `id + version > 0`
} else {
    // Default/fallback
    checkExpr1 = `name <> 'asd' AND description <> 'test'`
    checkExpr2 = `(id < 0 AND version < 9999) OR (id > 9999 AND version < 888)`
    checkExpr3 = `id + version > 0`
}
```

test/functional/schema-builder/entity/react-native/Post.ts
```diff
@@ -0,0 +1,23 @@
+import { Entity } from "../../../../../src/decorator/entity/Entity"
+import { Column } from "../../../../../src/decorator/columns/Column"
+import { Unique } from "../../../../../src/decorator/Unique"
+import { PrimaryColumn } from "../../../../../src/decorator/columns/PrimaryColumn"
+import { Check } from "../../../../../src/decorator/Check"
+
+@Entity()
+@Unique(["text", "tag"])
+@Check(`"likesCount" < 1000`)
```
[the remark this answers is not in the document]
ZBAGI: react-native uses SQLite as its underlying database in tests, no need to take sql db into consideration here.

ZBAGI:
It turned out that updating the tests was not as simple as adding `if(driver X) @Check(Y)`.  Since TypeORM runs all drivers together and then iterates through them using the same entities on separate connections, I *could* have modified their `Check` definitions at runtime, but I wanted to avoid that and keep the original entity definitions untouched. It just felt wrong to patch decorators dynamically, so I went for a cleaner approach that preserves the original definitions but still allows driver-specific differences.  

So, I added a new feature for the test suite: [driver-dependent entity discovery](https://github.com/typeorm/typeorm/blob/bb1be2ad16327b2a018a066f5a630518ccff882e/test/utils/test-utils.ts#L501). This makes it possible for tests to load slightly different entities depending on the driver being tested. It could also be useful for future tests where entity definitions differ per driver (like with `Check` decorator).  

Once I split the entities and defined the correct `CHECK` constraint queries for each driver, another issue came up: some drivers had incorrect check definitions. They referenced strings instead of columns because of wrong escaping (they were defined as `"version"`), which caused some tests to blow up since the correct constraint was now being generated.

After fixing that, I discovered that MariaDB doesn't truly support JSON fields. It stores them as `LONGTEXT` and adds a system `CHECK` constraint to simulate JSON validation. This system constraint was picked up by the new check-constraint logic and caused issues. [I solved it by filtering `checks` with regex](https://github.com/typeorm/typeorm/blob/ec9f02cd2e555157eabeeca90421e16d855da8a8/src/driver/mysql/MysqlQueryRunner.ts#L3056), this is not ideal, but I could not find a better solution. I will leave it up to the reviewer to decide if that is acceptable.

P.S. Sorry for the commit spam - I had trouble running all tests locally, so I kept pushing to remote to see how CI handled them.

test/functional/schema-builder/entity/expo/Post.ts
```diff
@@ -0,0 +1,23 @@
...
+import { Check } from "../../../../../src/decorator/Check"
+
+@Entity()
+@Unique(["text", "tag"])
+@Check(`"likesCount" < 1000`)
+export class Post {
+    @PrimaryColumn()
+    id: number
+    @Column({ unique: true })
+    version: string
```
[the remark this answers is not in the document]
ZBAGI: `version: string` is consistent with all `schema-builder` tests

ZBAGI:
@gioboa I am not sure if it is possible to split this into smaller PRs in a meaningful way, because it is essentially one feature ie. `adding CHECK constraint support for the MySQL/MariaDB driver.`

The big diff mostly comes from test and docs updates. I had to split the test entities for each driver because of differences in `@Check` syntax, which needed to be done, as pointed out in [this comment.](https://github.com/typeorm/typeorm/blob/6f486e5a67c007287949be119f233fb2b4fb7a59/test/functional/query-runner/entity/Post.ts#L13).

If you look only at the implementation, excluding the `test` folder and the docs file, the change set is actually quite small. It only touches these four source files:

[- `src/driver/DriverUtils.ts`](https://github.com/typeorm/typeorm/pull/11724/files#diff-815ba40c56c4edb31c2cd63e2a75d0bf046d93831b4c2e7ea8f3c861d0cbb593)
[- `src/driver/mysql/MysqlDriver.ts`](https://github.com/typeorm/typeorm/pull/11724/files#diff-a5db73ecf35a00a64edc5bc64385dd06bcf618736cc3d9fd422091a178ffcf82)
[- `src/driver/mysql/MysqlQueryRunner.ts`](https://github.com/typeorm/typeorm/pull/11724/files#diff-78ff8320cc81d7f789d6ee8fa43c23fb755e01dc3d7f13d5e5fda1b85e6e5e56)
[- `src/schema-builder/RdbmsSchemaBuilder.ts`](https://github.com/typeorm/typeorm/pull/11724/files#diff-ec225eb8228f199d9649636e6cffa1ab8107bb42446a240e1828183e64603e25)

But if you have a specific split in mind (like, one PR for the driver/schema changes and one PR just for the test refactor), I am happy to try to adjust it in that direction, but right now the tests and implementation feel quite tightly coupled around this single feature.

docs/docs/migrations/09-api.md
```diff
@@ -436,7 +436,7 @@ createCheckConstraint(table: Table|string, checkConstraint: TableCheck): Promise
 
 Creates a new check constraint.
 
-> Note: MySQL does not support check constraints.
+> Note: MySQL supports CHECK constraints starting from version 8.0.16. For earlier MySQL versions CHECK definitions are parsed but ignored by the server, so TypeORM will skip creating them. Ensure your MySQL server is >= 8.0.16 to use CHECK constraints.
```
G0maa: > For earlier MySQL versions CHECK definitions are parsed but ignored by the server

I don't think silently ignoring stuff is a good idea...

src/driver/mysql/MysqlQueryRunner.ts
```diff
@@ -2545,18 +2600,37 @@ export class MysqlQueryRunner extends BaseQueryRunner implements QueryRunner {
...
+                INNER JOIN information_schema.CHECK_CONSTRAINTS cc
+                    ON cc.CONSTRAINT_SCHEMA = tc.CONSTRAINT_SCHEMA
+                    AND cc.CONSTRAINT_NAME = tc.CONSTRAINT_NAME
+                WHERE tc.TABLE_SCHEMA = '${TABLE_SCHEMA}'
+                    AND tc.TABLE_NAME = '${TABLE_NAME}'
+                    AND tc.CONSTRAINT_TYPE = 'CHECK'
+                `
+            })
+            .join(" UNION ")
+
```
G0maa, line 2624: Can we maybe ignore `MariaDB` [issue](https://github.com/typeorm/typeorm/pull/11724/files#diff-78ff8320cc81d7f789d6ee8fa43c23fb755e01dc3d7f13d5e5fda1b85e6e5e56R3059-R3076) by modifying this query?

I guess this would get weird if a user names an `@Check` `json_valid` or something, but I think we can withstand that 😬

src/driver/mysql/MysqlQueryRunner.ts
```diff
@@ -2975,6 +3049,70 @@ export class MysqlQueryRunner extends BaseQueryRunner implements QueryRunner {
 
                 table.comment = dbTable["TABLE_COMMENT"]
 
+                if (this.driver.isCheckConstraintsSupported) {
```
G0maa, line 3059: Need your help @gioboa @alumni

src/driver/mysql/MysqlQueryRunner.ts
G0maa: There seems to be a similar instance like this one, could you take a look at this [comment](http://github.com/zbagi/typeorm/blob/mysql-checks/src/naming-strategy/NamingStrategyInterface.ts#L95-L99).

src/driver/DriverUtils.ts
```diff
@@ -29,7 +30,7 @@ export class DriverUtils {
     /**
      * Returns true if given driver is MySQL-based driver.
      */
-    static isMySQLFamily(driver: Driver): boolean {
+    static isMySQLFamily(driver: Driver): driver is MysqlDriver {
```
G0maa, line 33: Unrelated: Maybe we should use syntax (type predicate) for similar functions too? @alumni  @gioboa

test/functional/query-runner/entity/aurora-mysql/Post.ts
```diff
@@ -0,0 +1,27 @@
...
+    Entity,
+    PrimaryColumn,
+    Column,
+    Unique,
+    Check,
+} from "../../../../../src"
+
+@Entity()
+@Unique(["text", "tag"])
+@Check(`\`version\` < 999`)
```
G0maa, line 11: I've read your comments regarding portability of `@Check`, but just to verify the only way to write a `@Check` is to use `/` and ` as in this line? Is this for all DBs or a specific DB?

I see this entity is in `aurora-mysql` directory too, which was ignored earlier in the `dropOldChecks` and `createNewChecks`.

G0maa: Looking at the other files it seems like every few DBMSs has a way of writing the `@Check` correctly 😓

test/functional/query-runner/create-table.test.ts
```diff
@@ -103,7 +105,7 @@ describe("query runner > create table", () => {
         Promise.all(
             connections.map(async (connection) => {
                 const queryRunner = connection.createQueryRunner()
-                const metadata = connection.getMetadata(Post)
+                const metadata = connection.getMetadata("Post")
```
G0maa, line 108: but why?

ZBAGI: Previously we referred to `Post` as a single shared type. Now each driver has its own `Post` entity, so we need to reference the type by name. This is simpler than including all driver-specific `Post` variants in one place and switching between them for each test run.

test/functional/query-runner/entity/aurora-mysql/Post.ts
ZBAGI: Yes, exactly. `@Check` is essentially a raw SQL expression. Each driver has its own SQL dialect and identifier-quoting rules, so the same constraint may need different syntax depending on the database.

For example, MySQL commonly uses backticks, MSSQL uses brackets, and others may differ as well. That is why we keep driver-specific entities/definitions where needed.

src/driver/DriverUtils.ts
ZBAGI: I agree. These `isXXXFamily` helpers are conceptually type guards, so expressing them as type predicate functions would be consistent and would improve typing at call sites. I do not see any downside to switching them. I can make a follow-up PR for this once this PR is merged.

src/driver/mysql/MysqlQueryRunner.ts
ZBAGI: I can clarify this part. In [`MysqlDriver`, i added a new `isCheckConstraintsSupported` property that is initialized at startup](https://github.com/typeorm/typeorm/pull/11724/files#diff-a5db73ecf35a00a64edc5bc64385dd06bcf618736cc3d9fd422091a178ffcf82R424). It checks the server version and enables check-constraint support only when the version is known to actually enforce it.

This allows migrations and schema sync to behave correctly across MySQL versions.

docs/docs/migrations/09-api.md
ZBAGI: Yes, it is totally stupid Oracle Logic ™ ;) this is just how MySQL server behaves for versions prior to 8.0.16, not something we can control :)

Pre 8.0.16 MySQL can accept `CHECK` definitions syntactically, but the server does not enforce them. That is exactly why i introduced `isCheckConstraintsSupported` and set it based on version, e.g. `VersionUtils.isGreaterOrEqual(this.version, "8.0.16")`, so we only generate/manage checks when the server will actually support and enforce them. That is also why I clarified this in the docs: "Ensure your MySQL server is >= 8.0.16 to use CHECK".

src/driver/mysql/MysqlQueryRunner.ts
ZBAGI: I *think* I get your idea - reusing the existing "ignored check" pattern we already have for simple-enum.

If that's what you suggest, then we have the inverse problem. Here we discover a MariaDB JSON CHECK in the database that is not defined in metadata, and we need to figure out whether it is system-generated so we do not try to drop it and end up with a "check constraint not found" error.

And to be honest, the current solution - matching the `/^json_valid\s*\(\s*``?([a-z0-9_]+)``?\s*\)$/i` regex - even if not perfect, should be safe enough. In the worst-case scenario, if we falsely match a user-defined constraint that (for some reason) looks like `json_valid(column)`, we would simply skip generating a drop/update for it. No data would be lost. And since this only applies to MariaDB and to a very specific constraint shape, the likelihood of a real collision seems quite low.

ZBAGI:
> Hello @ZBAGI,
> 
> Thanks for your contribution, your expertise is really showing from how you write code, and I enjoyed reviewing this PR and I learnt stuff along the way too.
> 
> I do agree with @gioboa though, this PR might be too big, you can get the sense of this by going through your comments themselves, generally when I read X led to Y led to Z I get the notion that maybe X, Y, & Z could be separate PRs.
> 
> X: Implementing `applyDriverPlaceholder` feature for tests. Y: Fixing `@Check` in every test. Z: Implementing `@Check` for MySQL/MariaDB.
> 
> Ultimately we need to discuss X, before anything else.
> 
> Thanks again!
> 
> CC: @gioboa and @alumni

Thanks, glad the review was useful :)
And yep, that makes sense.

Since we have already started reviewing this PR, should we continue here first, or would you rather I open a separate PR with only the `applyDriverPlaceholder` pattern to start with?

docs/docs/migrations/09-api.md
G0maa: I think the philosophy behind ORMs (in general) is that it works the same for every DBMS.... more or less.

Meaning I'd prefer if an error is thrown if a DBMS does not support (or enforce) `CHECK`s. This is purely developer experience wise.

src/driver/mysql/MysqlQueryRunner.ts
G0maa: Can we instead of using a RegEx use e.g. `WHERE tc.CONSTRAINT_NAME NOT ILIKE 'json_valid(%'` in `checksSubquerySql`?

test/functional/query-runner/entity/aurora-mysql/Post.ts
G0maa: I think in some other PR we need to define a single method to define `@Check`s, then we take that and "translate" it to the correct dialect somehow. Similarly for other decorators having the same "issue". Unless this is considered OK.

@alumni wdyt?

G0maa:
> Since we have already started reviewing this PR, should we continue here first, or would you rather I open a separate PR with only the applyDriverPlaceholder pattern to start with?

Will let one of the maintainers with write-access decide.

docs/docs/migrations/09-api.md
ZBAGI: This is just how **MySQL server** works. The comment is only saying that MySQL silently ignores CHECK constraints prior to 8.0.16, not that TypeORM silently ignores them.

From TypeORM side nothing is silently ignored, [the query runner throws TypeORMError: MySql version does not support check constraints](https://github.com/typeorm/typeorm/pull/11724/changes#diff-78ff8320cc81d7f789d6ee8fa43c23fb755e01dc3d7f13d5e5fda1b85e6e5e56R2059). (same behavior as before for all MySQL versions)

So the "silently ignored" part refers purely to **MySQL server behavior**, not TypeORM behavior. I updated the docs because I can already predict people saying "but it works on older versions, look I can add CHECK" - yes, you can add it syntactically on older MySQL, but its ignored by the MySQL server itself (not enforced), which is exactly why we gate it on >= 8.0.16 and call it out in the documentation.

ZBAGI: I have removed the part of the comment that I think could have made you (or others) confused. Let me know if this is cleaner now.

src/driver/DriverUtils.ts
alumni: We actually plan to remove these completely since we want to remove all drivers from the main TypeORM package.

src/schema-builder/RdbmsSchemaBuilder.ts
```diff
@@ -484,9 +484,9 @@ export class RdbmsSchemaBuilder implements SchemaBuilder {
     }
 
     protected async dropOldChecks(): Promise<void> {
-        // Mysql does not support check constraints
         if (
-            DriverUtils.isMySQLFamily(this.connection.driver) ||
+            (DriverUtils.isMySQLFamily(this.connection.driver) &&
+                !this.connection.driver.isCheckConstraintsSupported) ||
```
alumni, line 489: `isCheckConstraintsSupported` should exist on all drivers, we should not rely on the type inference of `isMySQLFamily`, especially since we plan to remove it.

src/schema-builder/RdbmsSchemaBuilder.ts
```diff
@@ -484,9 +484,9 @@ export class RdbmsSchemaBuilder implements SchemaBuilder {
...
             this.connection.driver.options.type === "aurora-mysql"
```
alumni, line 490: `aurora-mysql` is just mysql/mariadb managed by AWS. So everything that applies to mariadb/mysql should apply to aurora as well.

test/functional/query-runner/entity/aurora-mysql/Post.ts
alumni: MySQL/MariaDB are the weirdest databases we support since they do not follow the ANSI SQL  standard (identifiers and strings do not use double quotes and single quotes respectively, they use backticks and double quotes instead).

I was actually thinking to forcefully enable ANSI quotes in MySQL/MariaDB in one of the next major versions (or to make TypeORM use the right quotation style based on the connection option).

I have a similar issue with `VirtualColumn`s, except it's worse there. Someone thought already about `ViewEntity`, views support `QueryBuilder` which makes them driver-independent.

For `@Check`s, we could possibly allow it to use a `FindOptionsWhere` argument instead of the raw string. Then, TypeORM could generate the condition for each driver.

Not something for this PR.

test/functional/query-runner/entity/aurora-postgres/Post.ts
```diff
@@ -1,17 +1,16 @@
...
-    Column,
     Entity,
-    Exclusion,
     PrimaryColumn,
+    Column,
     Unique,
-} from "../../../../src"
+    Check,
+    Exclusion,
+} from "../../../../../src"
```
alumni, line 8: The import order was okay before. The TypeScript LSP sorts them in alphabetical order if told so. In VSCode, it's "Organize imports".

We will enable automatic import sorting using ESLint in the future.

test/functional/query-runner/entity/aurora-mysql/Post.ts
alumni: I actually wonder if we could do something so we don't have to copy-paste the entities for every driver. This PR changes tests that are not related to this issue, this should not happen.

In other tests, we dynamically change the EntityMetadata for MySQL to get the tests working (so we don't have to maintain an entity for each driver).

Issue #11515, opened 2025-06-10
DateUtils.mixedDateToDateString should support UTC


dgildeh:
### Feature Description

I spent a while over the weekend trying to find out why my dates were being converted from UTC strings using the column transformer in my queries before I realized the SQLIte Abstract Driver I was using was doing something funky behind the scenes because my column type was 'date'.

The function that converts the date to write to the database converts it to a local-time string:

https://github.com/typeorm/typeorm/blob/930eefd758eb6954f29dc8b8c4f189eb02491ffd/src/util/DateUtils.ts#L28C12-L28C33

It would be great to add utc: boolean option to date fields and if set to true, use UTC string instead of the local time string so the database can save values in a consistent timezone.

### The Solution

Modify the date type column to allow a property utc: boolean.

If set to true, ensure this function DateUtils.mixedDateToDateString handles UTC conversion instead of just local time.

### Considered Alternatives

In the meantime you can set your column type to 'text' and write a custom transformer like below:

```
  @Column({
    type: 'text',
    default: () => "(date('now', 'utc'))",
    transformer: {
      to: (value: Date) => {
        return value ? value.toISOString().slice(0, 10) : undefined
      },
      from: (value: string | Date) => {
        if (value instanceof Date) {
          return new Date(Date.UTC(value.getUTCFullYear(), value.getUTCMonth(), value.getUTCDate()))
        }
        const [year, month, day] = value.split('-').map(Number)
        return new Date(Date.UTC(year, month - 1, day))
      },
    },
  })
  date!: Date
```

### Relevant Database Driver(s)

- [ ] aurora-mysql
- [ ] aurora-postgres
- [ ] better-sqlite3
- [ ] cockroachdb
- [ ] cordova
- [ ] expo
- [ ] mongodb
- [ ] mysql
- [ ] nativescript
- [ ] oracle
- [ ] postgres
- [ ] react-native
- [ ] sap
- [ ] spanner
- [ ] sqlite
- [x] sqlite-abstract
- [ ] sqljs
- [ ] sqlserver

### Are you willing to resolve this issue by submitting a Pull Request?

No, I don’t have the time and I’m okay to wait for the community / maintainers to resolve this issue.

PR #11740, opened 2025-10-29
feat:add utc flag to date column

CHOIJEWON:
closes: #11515 

### Description of change

This PR adds UTC timezone support for date type columns by introducing a new utc option to ColumnOptions.

Currently, date columns always use local timezone when converting dates to strings, making it difficult to maintain consistent
timezone across different environments. This can cause issues in distributed systems or when the application server timezone differs from the database timezone.

With the new utc flag, developers can explicitly store and retrieve dates in UTC timezone, ensuring data consistency regardless of the server's local timezone setting.

### Example Usage:
```typescript
  @Entity()
  class Event {
      @PrimaryGeneratedColumn()
      id: number

      @Column({ type: "date" })
      localDate: Date  // Uses local timezone (existing behavior)

      @Column({ type: "date", utc: true })
      utcDate: Date  // Uses UTC timezone (new feature)
  }
```
### What changed
- Added utc?: boolean option to ColumnOptions interface
- Updated ColumnMetadata to store and propagate the UTC flag
- Modified DateUtils.mixedDateToDateString() to support UTC conversion
- Applied UTC flag handling across all database drivers (MySQL, PostgreSQL, SQLite, Oracle, SQL Server, Spanner, CockroachDB, SAP HANA, React Native)
- Added comprehensive test case verifying UTC vs local timezone behavior

### Backward Compatibility
✅ No breaking changes - utc defaults to false, maintaining existing local timezone behavior

### Pull-Request Checklist

### Additional Notes

This is my first contribution to TypeORM and open source in general. I've done my best to follow the project's conventions and
guidelines, but I would greatly appreciate any feedback on:

- Code style and structure
- Test coverage and approach
- Documentation clarity
- Anything I might have missed or could improve

Thank you for taking the time to review this PR! I'm excited to contribute to TypeORM and look forward to learning from the
maintainers' feedback. 🙏

## Release Notes

* **New Features**
  * Added UTC timezone support for date columns, allowing dates to be stored and retrieved in UTC rather than local timezone.
  * New `utc` configuration option available for date column types across all supported databases.

* **Documentation**
  * Updated documentation to describe the new UTC date column option and default behavior.

### DIFF
--- a/src/driver/mysql/MysqlConnectionCredentialsOptions.ts
+++ b/src/driver/mysql/MysqlConnectionCredentialsOptions.ts
@@ -43,4 +43,10 @@ export interface MysqlConnectionCredentialsOptions {
      * Database socket path
      */
     readonly socketPath?: string
+
+    /**
+     * Maximum number of clients the pool should contain.
+     * for each connection
+     */
+    readonly poolSize?: number
 }
--- a/src/driver/mysql/MysqlDriver.ts
+++ b/src/driver/mysql/MysqlDriver.ts
@@ -1291,7 +1291,7 @@ export class MysqlDriver implements Driver {
                 port: credentials.port,
                 ssl: options.ssl,
                 socketPath: credentials.socketPath,
-                connectionLimit: options.poolSize,
+                connectionLimit: credentials.poolSize ?? options.poolSize,
             },
             options.acquireTimeout === undefined
                 ? {}


test/github-issues/11515/entity/Event.ts
```diff
@@ -0,0 +1,15 @@
+import { Column, Entity, PrimaryGeneratedColumn } from "../../../../src"
```
gioboa: We aren't creating tests for specific GitHub issues any more. Instead, tests should be added to the suites for the related functionality.

Can you please move these tests eg. test/functional/...?

Thanks 🙏

gioboa (commented):
Thanks for your help @CHOIJEWON

CHOIJEWON:
@gioboa 

I’ve updated the test directory location, but I’m not entirely sure if it’s the correct place. Please let me know if there’s anything that needs to be adjusted!

CHOIJEWON:
@alumni @sgarner @naorpeled @pkuczynski 

Hi there — sorry for tagging you when you're busy.
I wanted to kindly ask if there's anything else I should work on for this PR to move forward.
If not, would it be possible for you to review it when you have a moment?

Thank you!

alumni:
I'm thinking that the problem is a bit more complex - the server timezone is also important (and might be different than the one the app uses) - but I'm thinking we could let the users figure it out for now and fix it later (maybe here the docs/tsdoc for `utc` could be improved).

I'm also not sure which database clients (maybe mssql?) already return a Date object (and also not sure how the client decides how to create the Date object).

In my case (SAP) I'm getting strings either in the server's TZ or in UTC, depending which function is called. TypeORM uses the non-UTC functions for create/update/delete date columns (so this option won't work with them). Telling it to use `CURRENT_UTCTIMESTAMP` instead of `CURRENT_TIMESTAMP` depending on this option might add to the complexity.

I would still recommend using UTC everywhere just to be safe, but at least this partially solves the problem on the application side :)

alumni:
@CHOIJEWON one thing: this method is called for "date" not for "timestamp"/"datetime"? What's the purpose then?

CHOIJEWON:
@alumni Thank you for your review 🙌

**Regarding the scope being limited to "date" type:**

This implementation specifically addresses issue #11515, which reports timezone problems with the DATE column type. I intentionally
kept the scope limited to DATE columns to stay focused on the reported issue.

Even though DATE columns only store year-month-day without time information, the conversion still has timezone issues. When
`mixedDateToDateString()` uses local timezone methods (`getFullYear()`, `getMonth()`, `getDate()`), a JavaScript Date object can
convert to the wrong calendar date depending on the application's timezone.

The `utc` flag provides an opt-in solution for users who want consistent DATE storage across different application timezones by
using UTC methods (`getUTCFullYear()`, `getUTCMonth()`, `getUTCDate()`) instead.

**Regarding DATETIME/TIMESTAMP complexity:**

As you pointed out, DATETIME/TIMESTAMP columns involve additional complexities (server timezone, database function choices like
`CURRENT_TIMESTAMP` vs `CURRENT_UTCTIMESTAMP`, client behavior). I agree with your recommendation that users should handle those
with UTC everywhere at the database level.

This PR keeps the solution simple and focused on DATE columns where timezone handling is purely application-side :)

src/util/DateUtils.ts
```diff
@@ -25,8 +25,20 @@ export class DateUtils {
     /**
      * Converts given value into date string in a "YYYY-MM-DD" format.
      */
-    static mixedDateToDateString(value: string | Date): string {
+    static mixedDateToDateString(
+        value: string | Date,
+        utc: boolean = false,
```
naorpeled: nit:
Imo we should make this an object param so if we ever need more options here, we won't need to deal with breaking changes, what do you think?

naorpeled (approved):
LGTM, left a nit comment, let me know what you think about it 🙏 
After that's addressed I think we'll be good to merge

src/util/DateUtils.ts
CHOIJEWON: Sorry for the delayed reply — the day is just starting here in Korea 🙂
Your feedback is completely reasonable. I’ll apply the updates and push a new commit shortly.

CHOIJEWON:
@naorpeled 

Thank you for your valuable feedback! I've updated the implementation as you suggested.

### Changes Made:

I've refactored the `mixedDateToDateString` function parameter from a simple boolean to an object parameter to ensure future extensibility without breaking changes:

**src/util/DateUtils.ts:28-32**
```typescript
static mixedDateToDateString(
    value: string | Date,
    options?: { utc?: boolean },
): string {
    const utc = options?.utc ?? false
```

All database drivers (MySQL, PostgreSQL, SQLite, Oracle, SQL Server, Spanner, CockroachDB, SAP HANA, React Native) have been updated to use the new parameter format:

DateUtils.mixedDateToDateString(value, { utc: columnMetadata.utc })

This approach allows for potential future options to be added without requiring API changes.

Could you please review the updated implementation when you have a moment? I'd greatly appreciate any additional feedback.

Thank you!

naorpeled (approved):

gioboa (approved):
Thanks for your help @CHOIJEWON

CHOIJEWON:
@gioboa 

I just wanted to kindly ask if there’s anything else I should address in this PR.
If everything looks good from your perspective, I’m happy to wait for the merge 🙂

Issue #2549, opened 2018-07-20
Do not (exclusively) rely on class name to execute migration steps (e.g. use decorator instead)


dominic-simplan:
**Issue type:**

[ ] question
[ ] bug report
[x] feature request
[ ] documentation issue

**Database system/driver:**

[ ] `cordova`
[ ] `mongodb`
[ ] `mssql`
[ ] `mysql` / `mariadb`
[ ] `oracle`
[ ] `postgres`
[x] `sqlite`
[ ] `sqljs`
[ ] `react-native`
[ ] `expo`

**TypeORM version:**

[x] `latest`
[ ] `@next`
[ ] `0.x.x` (or put your version here)

Currently migrations seem to rely on the class name to execute migrations correctly. This causes issues if the mangle option of UglifyJS is enabled, as the class names are mangled. Wouldn't it be possible to use a decorator to define the timestamp instead? E.g:

```
@Migration(1529582677992)
export class InitialTables implements MigrationInterface {
}
```

As a workaround I currently pass the migration class names to the reserved option of UglifyJS, which is quite error-prone in the long run:
```
mangle: {
    reserved: ["InitialTables1529582677992", "AddForeignKeys1529588981853", "AddTable1531748224253"]
}
```

pleerock:
I would like to avoid using another decorator. What we can do is to add extra optional properties into `MigrationInterface` that users can implement and use for custom migration name. Feel free to contribute!

liangchunn:
I've hit a similar issue with normal relations as well (webpack build).

When `EntityMetadataValidator.validateDependencies()` is executed in a mangled file, it can happen that there are duplicate identifiers for a particular class as a result of the `Function.prototype.name` call
```diff
  entityMetadatas.forEach(entityMetadata => {
    graph.addNode(entityMetadata.name);
+   console.log(entityMetadata.name)
  });
```

With `keep_fnames: false` with terser/uglify,
```
p
u
d  <--
c
l
y
d  <--
(node:31945) UnhandledPromiseRejectionWarning: 
CircularRelationsError: Circular relations detected: p -> d -> d. 
...
```

Using `keep_fnames: true` fixes this and no error is thrown:
```
UserSpace
Space
Comment
Reaction
PostReaction
Post
User
```

If a webpack build is being managed by another tool (like `create-react-app`), it might be _almost impossible_ to override terser's/uglifyjs's configuration.

Issue #4714, opened 2019-09-09
code minify causes "Entity metadata for e#activities was not found"


GeorgeZhangZHY:
**Issue type:**

[ ] question
[x] bug report
[ ] feature request
[ ] documentation issue

**Database system/driver:**

[ ] `cordova`
[ ] `mongodb`
[ ] `mssql`
[ ] `mysql` / `mariadb`
[ ] `oracle`
[ ] `postgres`
[ ] `cockroachdb`
[x] `sqlite`
[ ] `sqljs`
[ ] `react-native`
[ ] `expo`

**TypeORM version:**

[ ] `latest`
[ ] `@next`
[x ] `0.2.18`

**Steps to reproduce or a small repository showing the problem:**

```typescript
@Entity('space')
export default class Space {
  @PrimaryColumn()
  spaceId: string;

  @Column()
  spaceName: string;

  @OneToMany('activity', 'space')
  activities: Promise<Activity[]>
}

@Entity('activity')
export default class Activity {
  @PrimaryColumn()
  activityId: string;

  @Column('text')
  text: string;

  @ManyToOne('space', 'activities')
  @JoinColumn({ name: 'spaceId' })
  space: Promise<Space>
}
```

In my project, code minifing (uglifyjs) is required, it will convert classnames like 'Space' into single characters like 'e'. The problem occurs here in [EntityMetadataBuilder.ts:L649](https://github.com/typeorm/typeorm/blob/63d993b1ef9f68a6e12bd535ac8ca8be39702502/src/metadata-builder/EntityMetadataBuilder.ts#L649). 

IMHO `m.givenTableName` is not honored but should. Therefore we can use uglifyjs without breaking typeorm.

hugoserrana:
@GeorgeZhangZHY How you create the connection?

I believe that using tools like Uglifyjs or Webpack, the correct would be to create the connection by passing the Entity list and not the file list, because or the nature or minify/module bundler,
they have to know about the entities at the time of code transpilation.

My project works with Webpack, I create a module with all my entities, and export them as a list:

```
import { Space } from './space.model';
import { Activity } from './activity.model';

export const EntityList = [ Space, Activity ]
```
And on connection creation:
`const connection = await createConnection({
				...
				entities: EntityList,
				subscribers: SubscriberList
});`

Hope this helps you

GeorgeZhangZHY:
[...]

PR #11741, opened 2025-10-29
feat: Add targetName option for entity identification

kkirby:
### Description of change

Fixes: #2549 #4714

Introduce a `targetName` option in `EntityOptions` and `TableMetadataArgs` to improve entity identification in relations and metadata resolution, especially useful in scenarios involving code minification. Update `EntityMetadata` to utilize `targetName` for consistent entity resolution.

### Pull-Request Checklist

## Summary by CodeRabbit

* **New Features**
  * Added optional `targetName` parameter to entity decorator for explicit entity identification
  * Introduced static `displayName` property as a configuration option on entity classes
  * Established entity resolution priority: explicit `targetName` > static `displayName` > class name

* **Documentation**
  * Added comprehensive guides for entity target naming and resolution strategies

### DIFF
--- a/docs/docs/entity/1-entities.md
+++ b/docs/docs/entity/1-entities.md
@@ -80,6 +80,122 @@ When using an entity constructor its arguments **must be optional**. Since ORM c
 
 Learn more about parameters `@Entity` in [Decorators reference](../help/3-decorator-reference.md).
 
+## Entity Target Name
+
+TypeORM supports an optional target name to identify and reference entities By default, the class name is used as the target name, which works well in most scenarios. However, in production environments where code is minified, class names can be shortened (e.g., `User` becomes `a`), which can break entity resolution and relations. To address this edge case, TypeORM provides optional `targetName` and `displayName` features as a failsafe.
+
+TypeORM uses the following priority order to determine an entity's target name:
+
+1. **Explicit `targetName`** in `@Entity()` decorator (highest priority)
+2. **Static `displayName` property** on the class (fallback)
+3. **Class name** (default behavior)
+
+### Using targetName
+
+You can specify an explicit `targetName` in the `@Entity()` decorator options:
+
+```typescript
+import { Entity, PrimaryGeneratedColumn, Column } from "typeorm"
+
+@Entity({ targetName: "UserEntity" })
+export class User {
+    @PrimaryGeneratedColumn()
+    id: number
+
+    @Column()
+    name: string
+}
+```
+
+This ensures that TypeORM will always identify this entity as `"UserEntity"`, regardless of whether the class name `User` gets minified to something else.
+
+You can then reference this entity by its target name in relations:
+
+```typescript
+import { Entity, PrimaryGeneratedColumn, Column, ManyToOne } from "typeorm"
+
+@Entity({ targetName: "PostEntity" })
+export class Post {
+    @PrimaryGeneratedColumn()
+    id: number
+
+    @Column()
+    title: string
+
+    @ManyToOne("UserEntity", "posts")
+    user: User
+}
+```
+
+And when getting repositories:
+
+```typescript
+const userRepository = dataSource.getRepository("UserEntity")
+const userMetadata = dataSource.getMetadata("UserEntity")
+```
+
+### Using static displayName
+
+Alternatively, you can define a static `displayName` property on your entity class:
+
+```typescript
+import { Entity, PrimaryGeneratedColumn, Column } from "typeorm"
+
+@Entity()
+export class User {
+    static displayName = "UserEntity"
+
+    @PrimaryGeneratedColumn()
+    id: number
+
+    @Column()
+    name: string
+}
+```
+
+This serves as a fallback if `targetName` is not explicitly provided in the decorator. The `displayName` property is useful when you want to keep the decorator clean but still ensure minification-safe entity identification.
+
+### When to use targetName or displayName
+
+In most cases, you don't need to specify `targetName` or `displayName`. TypeORM will use the class name by default, which works perfectly fine.
+
+**Consider using explicit `targetName` when:**
+
+-   Your code will be minified in production (e.g., bundled with webpack, terser, or similar tools) **AND** you use string-based entity references
+-   You want a failsafe to ensure entity resolution works regardless of minification
+-   You need guaranteed stable entity identification across different build configurations
+
+**Example for minification-safe production environments:**
+
+```typescript
+@Entity({ targetName: "UserEntity" })
+export class User {
+    @PrimaryGeneratedColumn()
+    id: number
+}
+
+@Entity({ targetName: "PostEntity" })
+export class Post {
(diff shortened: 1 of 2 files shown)


kkirby:
This is still a draft, but I'd like to be able to run the tests in CI as I cannot run them locally right now.

kkirby:
@gioboa I'm ready for this to be reviewed. Thank you!

test/github-issues/4714/entity/User.ts
```diff
@@ -0,0 +1,24 @@
+import { Column } from "../../../../src/decorator/columns/Column"
```
gioboa: We aren't creating tests for specific GitHub issues any more. Instead, tests should be added to the suites for the related functionality.

Can you please move these tests to test/functional/.../...?

Thanks 🙏

kkirby:
@coderabbitai generate docstrings

test/github-issues/4714/entity/User.ts
kkirby: Fixed! I'm not 100% certain what coderabbitai's hangup is with the documentation. If I get an explanation, I can work to correct it. Thanks!

kkirby: Sorry, I wasn't sure if I should @ you or not, so @gioboa, just in case. Thanks again.

pkuczynski (changes requested):
My personal take on this proposal is that it only adds complexity, while mentioned issues can be solved otherwise.

All across our [documentation](https://typeorm.io/docs/relations/many-to-many-relations) we present different way of defining relationships:

```typescript
@ManyToMany(() => Category)
@ManyToMany(() => Question, (question) => question.categories)
@ManyToOne(() => Question, (question) => question.questionToCategories)
```

They always use entity referenced from an import and not a static text. This would work even when minified. I actually didn't know one can use strings and I assume they are not strongly typed?

Similarily for migrations, they are now always generated with the `name` property, which would solve the issue.

So all in all, I would not move forward with this change...

kkirby:
> My personal take on this proposal is that it only adds complexity, while mentioned issues can be solved otherwise.
> 
> All across our [documentation](https://typeorm.io/docs/relations/many-to-many-relations) we present different way of defining relationships:
> 
> ```ts
> @ManyToMany(() => Category)
> @ManyToMany(() => Question, (question) => question.categories)
> @ManyToOne(() => Question, (question) => question.questionToCategories)
> ```
> 
> They always use entity referenced from an import and not a static text. This would work even when minified. I actually didn't know one can use strings and I assume they are not strongly typed?
> 
> Similarily for migrations, they are now always generated with the `name` property, which would solve the issue.
> 
> So all in all, I would not move forward with this change...

These are all valid concerns, and I appreciate your review. The use of string-based relationships is needed, however, when dealing with recursive relationships, as documented here: https://typeorm.io/docs/relations/relations-faq#avoid-circular-import-errors

pkuczynski:
> These are all valid concerns, and I appreciate your review. The use of string-based relationships is needed, however, when dealing with recursive relationships, as documented here: [typeorm.io/docs/relations/relations-faq#avoid-circular-import-errors](https://typeorm.io/docs/relations/relations-faq#avoid-circular-import-errors)

This must be some old documentation. Modern node does not have an issue with such circular references, I just checked. Since referenced issues mention bundled code, I assume they might mean something running in a browser? I haven't tried that yet. Do you think you could provide a small repo demonstrating such issue in the modern browsers?

Bear in mind, that putting both classes in a single file should also solve your issue, although I admit it might not be that elegant and only see as potential workaround...

Also bear in mind that issues you linked were raised in 2018 and 2019. Many has changed since then... In modern ESM (Node ≥13+, browsers, Vite, Webpack, etc.) circular references as we use them for TypeORM should not be a problem...

kkirby:
@pkuczynski That makes a lot of sense. I'm in an older codebase, using an old version of NextJS, which does some funky stuff with recursive imports (hence this PR and not knowing it's usually a non-issue). That being said, I'm going to do some additional discovery and put this PR back into a draft. Thank you.

kkirby:
> Thanks @kkirby for your help.
> Are you planning to investigate more on this?

Hi there. I did perform additional investigation and I came to the conclusion that my team and I should abandon NextJS v14 and upgrade to the latest version, and use class references for relationship definitions. I finished the migration, and it appears to have successfully resolved all the issues that this pull request was intended to address. Thanks for your help! I’ll close the PR now.

gioboa:
I'm glad that you solved the problems 👏

alumni:
> Modern node does not have an issue with such circular references

It's not just Node. It can also happen when splitting code into domains and you set up restrictions on domains. You can use a framework like NestJS or a tool like NX which doesn't like them. NX for example blocks circular domain imports in its ESLint rule by default.

We tend to consider circular imports as bad design in both frontend and backend - we have quite a large project with lots of domains (e.g. ~200K LOC in one repo, ~350K in another, etc), so these things have been causing us various issues since forever - and if not when writing the code, in understanding the architecture. We actually tried a lot of tools to block circular imports even down to the file level (`eslint-plugin-import`, `madge`) until we settled for `dependency-cruiser`.

One option (if you're fine with it, not everyone likes it) is to keep the entire data model in a single domain (separate from the rest of the implementation).

BTW, since HTTP compression is much better than the mangling part of minification, unless you want to obfuscate the code, you can tell `terser` not to mangle anything at all when it minifies the code. Or you can tell it exactly what not to mangle (certain classes or all classes). Probably other minifiers have similar options.

kkirby:
@alumni I think I follow where you’re coming from, and I don’t disagree with you. The edge case you’re describing is real, and in an ideal world there would be a cleaner, explicit solution for it.

The reason I closed the PR is more practical than philosophical. After the feedback here, I went back and re-tested more carefully with modern NextJS and current tooling, and it does in fact behave as expected. I originally thought I was hitting a more fundamental issue, but on revisiting it, that turned out to be a misread on my part.

Given that, and given that the older issues date back several years and no longer seem to apply under current assumptions, I don’t really see a strong case for pushing this forward when the maintainers are saying “this should already work,” and in practice, they’re right.

I agree with you that it’s an edge case that could benefit from a more explicit escape hatch. I just don’t think it’s a hill worth dying on, or something I feel good asking the TypeORM team to take on long-term when it’s not a core concern and affects a relatively small set of users.

With all that being said, my intention here isn't to close off discussion. If others feel strongly that this is worth revisiting or have additional context to add, I think it’s a valid conversation to have.

And if a maintainer does want to revisit the topic, I’m more than happy to reopen the PR.

----

Edit: When I wrote this, I didn’t realize you were a maintainer. I didn’t mean to imply you were outside that group, and I agree the circular-import/tooling edge case is real.

alumni:
Yeah, I wouldn't add this feature unless someone really needs it, since it adds another configuration option to TypeORM :)

I wanted to say there are already some potential solutions to avoid these issues in case someone stumbles over this PR in the future ;)

PR #11747, opened 2025-11-01
fix: leftJoinAndMapOne ignores custom select and always maps full entity

### LINKED ISSUE #11745: leftJoinAndMapOne ignores custom select and always maps full entity
### Issue description

leftJoinAndMapOne ignores custom select and always maps full entity

### Expected Behavior

The mapped myRespond property should only include the selected columns:

```js
"myRespond": {
  "id": 2,
  "status": "no"
}
```

### Actual Behavior

TypeORM ignores the select and returns the full entity:
```js
"myRespond": {
  "id": 2,
  "eventId": 1,
  "userId": 1,
  "status": "no"
}
```

### Steps to reproduce

When using leftJoinAndMapOne (or leftJoinAndMapMany) with a subquery and a .select() specifying only certain columns, TypeORM still maps the full entity instead of respecting the selected columns.

```js
.leftJoinAndMapOne(
  'event.myRespond',
  (qb) => qb
    .select(['respond.id', 'respond.status']) // i need id and status
    .from(EventRespond, 'respond')
    .where('respond.userId = :userId', { userId }),
  'respond',
  'respond.eventId = event.id',
  { },
)
```

### My Environment

| Operating System | MacOS |
| Node.js version |22.16.0 |
| Typescript version | 5.8.3 |
| TypeORM version | 0.3.27 |

### Additional Context

Workarounds

Using an @AfterLoad() hook to manually pick columns works:

```js
@AfterLoad()
pickColumns() {
  this.myRespond = this.myRespond ? { id: this.myRespond.id, status: this.myRespond.status } : null;
}

```

- [ ] aurora-mysql
- [ ] aurora-postgres
- [ ] better-sqlite3
- [ ] cockroachdb
- [ ] cordova
- [ ] expo
- [ ] mongodb
- [ ] mysql
- [ ] nativescript
- [ ] oracle
- [ ] postgres
- [ ] react-native
- [ ] sap
- [ ] spanner
- [ ] sqlite
- [ ] sqlite-abstract
- [ ] sqljs
- [ ] sqlserver

Higangssh:
### What this change does
This PR addresses issue #11745 where `leftJoinAndMapOne` and `leftJoinAndMapMany` with subqueries ignore explicit `.select()` statements and return all entity columns instead of only the selected ones.

### Why this change is needed
When using subqueries with partial column selections, TypeORM should respect those selections. Currently, even when explicitly selecting only specific columns, the result includes all entity columns.

**Current behavior:**
```typescript
// Query with .select(["respond.id", "respond.status", "respond.eventId"])
Result: { id: 2, eventId: 1, userId: 1, status: "no" } // userId shouldn't be here
```

**Expected behavior:**
```typescript
Result: { id: 2, eventId: 1, status: "no" } // Only selected columns
```

### My approach and implementation

I considered a few different approaches and would appreciate feedback on the direction I took:

#### Considered approach 1: SQL-level filtering
My first thought was to modify the SQL query generation to only include selected columns in the final query. However, I ran into several challenges:

- **Column aliasing complexity**: TypeORM automatically generates aliases like `respond_id`, `respond_eventId` for subquery columns, and modifying this would require tracking which aliases correspond to which selected columns
- **Driver-specific implementations**: Each database driver (MySQL, PostgreSQL, SQLite, etc.) has slightly different SQL generation logic, so the fix would need to be replicated across multiple drivers
- **Subquery already built**: By the time we could filter columns, the subquery SQL is already generated and wrapped. Retroactively filtering the outer SELECT would be complex
- **Risk to existing functionality**: The SQL builder has many edge cases (joins, relations, conditions), and changing it could break existing queries in subtle ways

After looking at the code, this approach felt too invasive for what should be a targeted fix.

#### Implemented approach 2: Transformation-level filtering
Instead, I decided to handle this at the transformation layer where raw SQL results get converted to entity objects:

1. **Track selections** (`JoinAttribute.ts`) - Added `selectedColumns` and `__subqueryMetadata` properties to store which columns were selected in subqueries
2. **Extract during join creation** (`SelectQueryBuilder.ts`) - Modified the `join()` method to extract selected columns from subquery's `expressionMap.selects`
3. **Filter during transformation** (`RawSqlResultsToEntityTransformer.ts`) - Updated column processing to filter based on `selectedColumns` when present, while keeping the existing caching logic for regular joins

**Benefits of this approach:**
- Doesn't touch SQL generation (safer, driver-agnostic)
- Only affects the specific case of subquery joins with explicit selections
- Preserves existing performance optimizations (caching still works for regular joins)

#### Challenge encountered: Distinguishing Entity classes from QueryBuilder functions
Both Entity classes and QueryBuilder functions are JavaScript functions, so `typeof entityOrProperty === "function"` matched both cases. This caused "Class constructor cannot be invoked without 'new'" errors when trying to invoke Entity classes.

I solved this by adding a `!this.connection.hasMetadata(entityOrProperty)` check to distinguish between them - Entity classes have metadata, QueryBuilder functions don't.

#### Feedback I'd appreciate
- Is the transformation-level approach the right way to handle this, or would you prefer the SQL-level approach despite the complexity?
- Are there any edge cases I might have missed with this implementation?
- Is there a better pattern for distinguishing Entity classes from QueryBuilder functions?

### How I've verified it
- Added test cases in `test/functional/query-builder/join-and-map/left-join-and-map-one-subquery-select.ts` covering:
  - Partial column selections (core issue)
  - Multiple partial selects with different columns
- Verified all 2,369 existing tests still pass (no regressions)
- Confirmed the fix only applies to subquery joins with explicit selections (zero performance impact on regular joins)

---

## Pull-Request Checklist

- [x] This pull request links relevant issues as `Fixes #11745`
- [x] All tests pass locally (`npm test` - 2,369 passing)
- [x] The code follows existing code style (prettier/eslint ran via pre-commit hook)

---

**Note:** This is my first contribution to TypeORM, so I'd really appreciate any suggestions or guidance on improving the implementation. Thank you for maintaining this awesome project!

## Summary by CodeRabbit

* **New Features**
  * Subquery joins can return only explicitly selected columns so joined data includes exactly requested fields.
  * Joins mapped to an entity property are treated as selected to ensure mapped values populate with partial selects.
  * Subquery-provided metadata is respected so mixed join scenarios map results correctly.

* **Tests**
  * Added functional tests and sample entities validating partial subquery selects, nested property selection, and mapped-join behavior.

Fixes #11745


pkuczynski:
Hi @Higangssh, thank you for taking the time to investigate this issue and for the detailed write-up of your approach — especially as a first contribution!

The issue itself is valid: `leftJoinAndMapOne`/`leftJoinAndMapMany` should respect `.select()` in subqueries. However, after reviewing the implementation, I don't think the transformation-level filtering is the right approach. Here's why:

**The filtering happens after the database call, not before.** The SQL query still selects every column from the joined subquery — the unwanted columns are only stripped out in `RawSqlResultsToEntityTransformer` when converting raw rows to entities. This means the database still reads all columns, sends them over the wire, and TypeORM allocates memory for them before throwing them away in JavaScript. The whole reason a developer would use `.select()` on a subquery join is to make the query more efficient — this fix removes that benefit while making it *look* like it works.

**The proper fix should happen at the SQL generation level** — when building the outer SELECT for a subquery join, only the columns explicitly selected in the subquery should be included in the generated SQL. I understand this is more complex (column aliasing, driver-specific behavior), but it's the correct place to solve this.

Beyond the architectural concern, the review also found several implementation issues:

1. **Double invocation of subquery factory** — `entityOrProperty(this.subQuery())` is called once to inspect selects, then again in the existing `isSubQuery` block to build the SQL. This is a performance regression and can cause bugs if the factory has side effects (e.g. `setParameters()`).
2. **Unsafe mutation of private `Alias._metadata`** — The transformer temporarily sets and restores `_metadata` via `as any` casts with no `try/finally` guard. If `transform()` throws, the alias metadata is permanently corrupted for subsequent queries.
3. **`isSelected` early-return regression** — Unconditionally returning `true` for `mapToProperty` joins changes behavior for existing callers who intentionally omit the mapped alias from the outer select to suppress hydration.
4. **Missing test coverage** — No tests for `leftJoinAndMapMany` (only `leftJoinAndMapOne` is tested), and the test setup is missing `schemaCreate: true, dropSchema: true`.

I'm closing this PR, but the underlying issue (#11745) remains valid and we'd like to fix it for v1.0. You're very welcome to open a new PR with a SQL-level fix that filters columns in the generated query rather than in the result transformer. That would be a great contribution!

Higangssh:
Hi @pkuczynski, thank you for the detailed review. I'll work on a new PR that fixes this at the SQL generation level as you suggested. Will target the column filtering in the SELECT expression builder rather than the result transformer.

PR #11758, opened 2025-11-09
lint: fix offenses for @typescript-eslint/no-unused-vars rule in src folder

pkuczynski:
### Description of change

Similar to #11757 but in `src` folder, so should be reviewed more carefully.

### Pull-Request Checklist

### DIFF
--- a/src/driver/oracle/OracleDriver.ts
+++ b/src/driver/oracle/OracleDriver.ts
@@ -434,11 +434,7 @@ export class OracleDriver implements Driver {
      * Build full table name with database name, schema name and table name.
      * Oracle does not support table schemas. One user can have only one schema.
      */
-    buildTableName(
-        tableName: string,
-        schema?: string,
-        database?: string,
-    ): string {
+    buildTableName(tableName: string, schema?: string): string {
         const tablePath = [tableName]
 
         if (schema) {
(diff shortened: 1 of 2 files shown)


gioboa (approved):
It looks good to me 👍

src/driver/aurora-mysql/AuroraMysqlQueryRunner.ts
```diff
@@ -1915,7 +1915,10 @@ export class AuroraMysqlQueryRunner
                 if (!isAnotherTransactionActive) {
                     await this.rollbackTransaction()
                 }
-            } catch (rollbackError) {}
+            } catch {
+                // ignore
+            }
```
mguida22: nit: it may be nice to provide context on why we want to ignore these errors and/or what we are ignoring. like "ignore rollback errors"

src/driver/mysql/MysqlDriver.ts
```diff
@@ -1166,7 +1163,7 @@ export class MysqlDriver implements Driver {
     /**
      * Creates an escaped parameter.
      */
-    createParameter(parameterName: string, index: number): string {
+    createParameter(): string {
```
alumni, line 1166: This should not be changed - it's the same signature in all drivers. It's also part of the public API, and changing it is a breaking change.

alumni (changes requested):
The method signatures should not change.

PR #11759, opened 2025-11-10
ci: migrate from nyc to c8

pkuczynski:
### Description of change

Replace [nyc](https://www.npmjs.com/package/nyc) with faster alternative [c8](https://www.npmjs.com/package/c8) - a newer, faster alternative that uses Node.js's native V8 coverage.

### Pull-Request Checklist

## Summary by CodeRabbit

* **Chores**
  * Migrated test coverage tooling to a new runner and updated CI workflows to use it.
  * Adjusted coverage configuration and reporting format.
  * Updated development tooling dependencies accordingly.
  * Changed ignore rules to stop excluding build and coverage artifacts and to begin tracking node_modules.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>


OSA413:
We were going to switch to Vitest in the future, is this change necessary?

pkuczynski:
> We were going to switch to Vitest in the future, is this change necessary?

TLDR: not absolutely necessary but should bring immediate CI speed improvement 

Vitest is used to use c8 (no using native v8 coverage), but seems that migration will take a while. And this should bring immediate benefit of speeding up CI. It's a drop in replacement so not a lot of work to do it and not much to remove it in the future...

I fired it up to do a quick comparison ([c8](https://github.com/typeorm/typeorm/actions/runs/19225268272/job/54951028616?pr=11759) vs [nyc](https://github.com/typeorm/typeorm/actions/runs/19224622536/job/54949053626?pr=11708))
* MongoDB on linux: 49s vs 59s
* MSSQL on linux: 4m 3s vs 4m 18s
* PostgreSQL on linux: 2m 3s vs 2m 17s
* Sqlite on windows: 2m 24s vs 2m 47s

Two random runs is not enough to calculate fair speed improvement, but it shaves off 10-20s, which is not bad given how little effort it is to do it.

And migration to vitest might take a while given how little progress has been made so far: https://github.com/typeorm/typeorm/compare/master...next

OSA413:
I wonder why the coverage CI/CD part is skipped

pkuczynski:
> I wonder why the coverage CI/CD part is skipped

It's not. It just wait for all tests to fininsh:

https://github.com/typeorm/typeorm/blob/master/.github/workflows/commit-validation.yml#L67

Its a good question though why oracle on linux started to fail now...

OSA413:
Can you please add html reporter?

OSA413:
Looks like lcov already provides html report.

After testing it generates 0 coverage, which is no good. Can you please investigate what's wrong?

OSA413:
At least it generate 0 coverage on my machine.

OSA413:
Switching back to `nyc` shows the correct coverage on my local machine.

pkuczynski:
> Looks like lcov already provides html report.
> 
> After testing it generates 0 coverage, which is no good. Can you please investigate what's wrong?

Yes, I am investigating. But I can't get the tests pass on my local machine :/ Testing with mysql-9...

OSA413 (changes requested):
*Blocked until fixed*

alumni:
This PR would be nice to have if the reporting gets fixed.

Currently, `nyc` brings in `baseline-browser-mapping` which logs a warning that breaks our tests. If we get `c8` working, we wouldn't need to figure out a way to fix that test :)

OSA413:
After adding `--exclude-after-remap` parameter to `c8` like this: `npx c8 --exclude-after-remap npm run test:ci` it started to show the coverage.

OSA413:
I'm still wondering why the coverage is marked as skipped though all tests are passed/failed.

OSA413 (approved):

alumni:
@OSA413 That's nice, thanks!

Do you happen to know where does the extra 4% coverage comes from?

BTW, Both `nyc` and `c8` pass their options to `test-exclude` (which uses `glob` after that), so I doubt that anything related to the file globs should be changed. `test-exclude` automatically adds `**/node_modules/**` if you pass `excludeNodeModules: true` - which is correct, while `node_modules` without `**` at the end does not do anything - so I doubt this change was responsible for the increased coverage.

pkuczynski:
Thank you for help @OSA413!

@alumni possibly it comes from how `all: true` + `exclude-after-remap: true` behaves differently than in nyc?

OSA413:
Here's the difference example (you can run SQLite tests locally and see it):

Old:
{BC411B59-000F-47FC-86E5-6230A405EF8A}

New:
{0507BA23-9474-40F7-8DF7-B23916F1C0FB}

Probably something related to TS -> JS (build) -> TS (Map back to source code) mapping.

OSA413:
I think that C8 can handle/track more things, that's why the coverage changed.

PR #11763, opened 2025-11-11
ci: add single step to validate if all tests passed

pkuczynski:
### Description of change

Add single step to validate if all tests passed, which will be used for branch protection rules

### Pull-Request Checklist

## Summary by CodeRabbit

* **Chores**
  * Renamed an internal CI workflow for clarity.
  * Refined which workflow files are picked up by CI.
  * Added an overall "all-passed" CI job that runs reliably and enforces that dependent jobs must succeed before the pipeline is marked passed, improving build validation and reliability.


michaelbromley (approved):
Note on why this PR exists:

We noticed that it was possible to merge a PR even when checks were failing or not yet completed. The only restriction to prevent merge was the 2 review minimum.

So I then altered the settings to require _all_ the tests to pass before one can merge.

However, this then means that targeted workflows that do not run tests (e.g. PRs just on docs that don't need to run tests) would not be mergable, since the tests didn't run.

Apparently GH Actions does not support this level of control natively.

So this solution allows us to require just this one single job "all-passed" to succeed, and then merge is allowed. In the case of e.g. docs PRs where tests are skipped, the "all-passed" will still succeed because it only fails on "failure" in any of the dependent jobs, and will ignore "skipped" jobs.

Issue #11692, opened 2025-09-30
Drop support for un-maintained/deprecated driver libraries


michaelbromley:
Currently, TypeORM supports some driver libraries that are either no longer maintained or are not well maintained and where there exists a clearly superior successor.

In these cases, when there is an option for a well-maintained successor library, we should drop support for the unmaintained versions as part of the v1.0 release.

Examples (more research needed to complete this list)

- `sqlite` -> `@sqlite.org/sqlite-wasm`?
- `mysql` -> `mysql2` (also: remove legacy spatial support as default)
- `redis` legacy code
- `expo` legacy code

pkuczynski:
Replaced by #11819

Issue #11765, opened 2025-11-13
how to print real location in code, when the typeorm throw error


dengshenkk:
### Issue description

I used typeorm with nestjs, when the data insert get error I don't know the typeorm real error location in nestjs service

### Expected Behavior

print the error location in the code file, like sourcemap?  I want to know that which file and rows and column in the error

### Actual Behavior

just print like this `node_modules/typeorm/persistence/src/persistence/tree/NestedSetSubjectExecutor.ts:92:38`

### Steps to reproduce

null

### My Environment

| Dependency          | Version  |
| ---                 | ---      |
| Operating System    |    macos      |
| Node.js version     | 22.0.0  |
| Typescript version  | x.y.zzz  |
| TypeORM version     | 0.3.26  |

### Relevant Database Driver(s)

- [ ] aurora-mysql
- [ ] aurora-postgres
- [ ] better-sqlite3
- [ ] cockroachdb
- [ ] cordova
- [ ] expo
- [ ] mongodb
- [x] mysql
- [ ] nativescript
- [ ] oracle
- [ ] postgres
- [ ] react-native
- [ ] sap
- [ ] spanner
- [ ] sqlite
- [ ] sqlite-abstract
- [ ] sqljs
- [ ] sqlserver

### Are you willing to resolve this issue by submitting a Pull Request?

Yes, I have the time, and I know how to start.

OSA413:
hm, I think such errors should print the whole stack trace to the relevant services in NestJS. Can you provide more info on how your app works and where do you call the insert method and how you inject the repository? and what version of TypeORM? can you provide a minimal reproducible example of the problem?

sgarner:
The quality of async stacktraces in Node depends on the Node version as well as how code is transpiled. 

So this might be a problem with the driver rather than TypeORM itself.

TypeORM is mostly written with modern async/await syntax and is transpiled to ES2021 (which retains async/await) so should produce good async stacktraces.

However if an error occurs in driver code, we're at the mercy of how that driver is implemented.

The `mysql2` driver [appears to target ES2016](https://github.com/sidorares/node-mysql2/blob/d950b989afda9623909b34e5d18e66e811971702/tsconfig.json#L5C16-L5C22), which will rewrite async/await as callbacks - with a corresponding reduction in stacktrace usefulness.

dengshenkk:
I found same case, it's for long time ago. It still not be resolve, is there is any plan to implement this feature? 

ref links
[...]

PR #11273, opened 2025-02-04
fix: mysql query stack traces

mukunda-:
### Description of change

Rearrange the `query` function in the MySQL query runner to build the error outside of the query callback scope, so it includes the caller's stack trace.

The inner stack trace, if desired, is still available via the wrapped `driverError`.

Functionality wise, this is a cosmetic change for diagnostics. It doesn't change program behavior, only stack trace logs. Only MySQL and equivalent flavors are in the blast radius here.

See #11271 for more details. Currently, when there's an query error while using mysql, it's hard to tell where it's coming from since the stack doesn't show the caller. (The "trace" option of the MySQL driver also does not reach outside of the typeorm code.) Working with traces like these is frustrating:

{8A435313-F939-409E-80F2-2D52F9B1FCF5}

Fixes #11271
Fixes #11765

### Pull-Request Checklist

- [x] Code is up-to-date with the `master` branch
- [x] `npm run format` to apply prettier formatting
- [x] `npm run test` passes with this change
- [x] This pull request links relevant issues as `Fixes #0000`
- [x] There are new or updated unit tests validating the change
- [x] Documentation has been updated to reflect this change -- N/A
- [x] The new commits follow conventions explained in [CONTRIBUTING.md](https://github.com/typeorm/typeorm/blob/master/CONTRIBUTING.md)

PR #11766, opened 2025-11-13
feat(mysql)!: drop support for mysql package and default to mysql2

pkuczynski:
Drop support for `mysql` package and default to `mysql2`. Removes some custom mysql driver logic and props, which does not make sense anymore.

Refs #11692

### DIFF
--- a/src/driver/mysql/MysqlDriver.ts
+++ b/src/driver/mysql/MysqlDriver.ts
@@ -1268,8 +1268,7 @@ export class MysqlDriver implements Driver {
             connection.on("error", (error: any) =>
                 logger.log(
                     "warn",
-                    `
-        }MySQL connection raised an error. ${error}`,
+                    `MySQL connection raised an error. ${error}`,
                 ),
             )
         }


alumni:
Nice!

Can you check if there's something related to authentication that can be removed? I remember that was the main change in the `mysql2` client. MySQL itself deprecated and later dropped support for old auth methods, but MariaDB and cloud versions of MySQL (AWS / Alibaba) might still use the old authentication.

Another thing we might need to do is to flip the default for legacy spatial support to false (i.e. remove it from connection options in `MysqlDriver.connect`).

pkuczynski:
> Can you check if there's something related to authentication that can be removed? I remember that was the main change in the `mysql2` client. MySQL itself deprecated and later dropped support for old auth methods, but MariaDB and cloud versions of MySQL (AWS / Alibaba) might still use the old authentication.

I did not used mysql for many years not, so do not have the details. Can you point me to what exactly do you mean or maybe to some documentation? I also added it to the list in #11692, so it does not get forgotten.

> Another thing we might need to do is to flip the default for legacy spatial support to false (i.e. remove it from connection options in `MysqlDriver.connect`).

This is already added to #11692

I would suggest we merge it as it is, and I will tackle remaning two points in separate PRs. I would also need some help since my knowledge of mysql is quite limited. I am typically using pg

OSA413:
Looks fine but needs testing.

Btw my `ultimate-test-suite` broke when trying to pull this PR, I will check that later.

pkuczynski:
> Btw my `ultimate-test-suite` broke when trying to pull this PR, I will check that later.

Can you share your test suite? And why it is not part of the repo?

OSA413:
> Can you share your test suite?

https://github.com/OSA413/typeorm-ultimate-test-suite

> And why it is not part of the repo?

Because 1) it's my initiative, 2) it's in WIP state, 3) the dataset is very huge and 4) no one else took the initiative to help (but that's ok). I'm currently not considering including it into the main repo because it can be unstable (and you need some time to understand what's going on there). I personally use the test suite as a canary to test PRs I'm reviewing.

pkuczynski:
> Because 1) it's my initiative, 2) it's in WIP state, 3) the dataset is very huge and 4) no one else took the initiative to help (but that's ok). I'm currently not considering including it into the main repo because it can be unstable (and you need some time to understand what's going on there). I personally use the test suite as a canary to test PRs I'm reviewing.

Thanks for sharing. The dataset is quite large, but rest does not seem extensively big. Yet it's indeed a but hard to grasp on the first look. And if the overal stability of the suite is in question, than indeed it's hard to rely on it.

Let me know what does your investigation brings as from my point of view this PR is ready to be merged.

OSA413:
Would [this](https://github.com/typeorm/typeorm/issues/11765) be a blocker to dropping support of `mysql`?

pkuczynski:
> Would [this](https://github.com/typeorm/typeorm/issues/11765) be a blocker to dropping support of `mysql`?

Not necessarily. There is a #11274 PR which should fix it, so we should maybe work on this one too?

OSA413:
> There is a https://github.com/typeorm/typeorm/pull/11274 PR which should fix it

I don't think this will fix the issue with truncated error log

pkuczynski:
@alumni @OSA413 what should we do about this then?

pkuczynski:
> I don't think this will fix the issue with truncated error log

Sorry, I meant to point to #11273 not #11274... Looks like the original author have not time to work on it, so I can take it over if you think its worth it? And given @alumni initial approval it probably is.

OSA413:
it's worth it, you probably should continue working in the PR's branch and carry it to the point of it being good enough to merge into master (I haven't checked what's the solution to the problem with the truncated log)

alumni:
I don't think #11273 is related in any way to this PR (besides affecting the same driver). There were some changes in the all drivers (including MySQL) related to event handling - that's why it has some merge conflicts - I wanted to fix and merge that PR since author is busy, but realized might take more than 5-10 min to fix the conflict. Feel free to check if you have the time (you can ignore my comment there about a promisify-like helper, it was just a suggestion).

I don't work with MySQL/MariaDB on a daily basis either, but if I change things I try to understand how they work before making the change.

For this PR, I would remove the legacy spatial support as default and put this information (and the client change) in some "migration guide" doc page (in the Guides section). Regarding auth, I would check how it works, but likely there's not much to remove, it was just a heads up that there is something that we should check (these things will not be done by code reviewers).

src/driver/mysql/MysqlDriver.ts
```diff
@@ -1343,7 +1268,8 @@ export class MysqlDriver implements Driver {
             connection.on("error", (error: any) =>
                 logger.log(
                     "warn",
-                    `MySQL connection raised an error. ${error}`,
+                    `
+        }MySQL connection raised an error. ${error}`,
```
gioboa: Is this change correct?

pkuczynski:
> It looks good, did you use find replace approach?

No, search and manual replace/changes. Why do you ask?

src/driver/mysql/MysqlDriver.ts
pkuczynski: Sure not! Good catch, thanks. Dunno how this sneaked in. Fixed

gioboa (approved):
It looks good to me, thank you @pkuczynski

Issue #11749, opened 2025-11-03
Bug in  detectRedisVersion


AdolfodelSel:
### Issue description

The detectRedisVersion function incorrectly detects the Redis version when using versions lower than 4. The "set" function does not exist in version 3.1.2, so the check mistakenly identifies the redisMajorVersion as 5 instead of 3.

### Expected Behavior

Ensure that the detectRedisVersion function correctly detects when the Redis version is 3.

### Actual Behavior

The detectRedisVersion function incorrectly detects version 3.1.2 as version 5.
`

    private detectRedisVersion(): void {
        if (this.clientType !== "redis") return
        try {
            // Detect version by examining the client's method signatures
            // This avoids creating test keys in the database
            const setMethod = this.client.set
            if (setMethod && setMethod.length <= 3) {
                // Redis 5+ set method accepts fewer parameters (key, value, options)
                this.redisMajorVersion = 5
            } else {
                // Redis 3/4 set method requires more parameters (key, value, flag, duration, callback)
                this.redisMajorVersion = 3
            }
        } catch {
            // Default to Redis 3/4 for maximum compatibility
            this.redisMajorVersion = 3
        }
    }

`

### Steps to reproduce

Use Redis version 3.1.2 and try executing a query with caching.
Configuration:
`
  cache: {
        type: 'redis',
        options: {
          host: 'localhost',
          port: 6379,
          db: 0
        },
        duration: 1800000,
        ignoreErrors: false
  }
`
Example:
`query.cache('test', 1000);`

### My Environment

| Dependency             | Version  |
| Redis version            | 3.1.2 |
| TypeORM version     | 0.3.27  |

### Relevant Database Driver(s)

- [ ] aurora-mysql
- [ ] aurora-postgres
- [ ] better-sqlite3
- [ ] cockroachdb
- [ ] cordova
- [ ] expo
- [ ] mongodb
- [ ] mysql
- [ ] nativescript
- [ ] oracle
- [ ] postgres
- [ ] react-native
- [ ] sap
- [ ] spanner
- [ ] sqlite
- [ ] sqlite-abstract
- [ ] sqljs
- [ ] sqlserver

### Are you willing to resolve this issue by submitting a Pull Request?

Yes, I have the time, and I know how to start.

mguida22:
Thanks for the report, can you open a PR with a fix? Ideally we can programmatically get the Redis version via a command instead of checking parameters.

[...]

Issue #11635, opened 2025-09-02
getFromCache method throws error


simon1389:
### Issue description

getFromCache method in RedisQueryResultCache throws error since version-check

### Expected Behavior

the new version of typeorm library should work with the new version-check in the getFromCache function.

### Actual Behavior

Since the check of the redis-version being in place ( `if (this.isRedis5OrHigher()) { `) this method throws an error when used in a nestjs application using typeorm.
Error thrown: `TypeError: Cannot read properties of undefined (reading 'then') TypeError: Cannot read properties of undefined (reading 'then')`

Seems the client.get function does not return a promise

### Steps to reproduce

Use typeorm version 0.3.26 in a nestjs application with nestjs/typeorm
try to load a cached value via `dataSource.queryResultCache.getFromCache`

--> error is thrown

### My Environment

| Dependency          | Version  |
| ---                 | ---      |
| Operating System    |          |
| Node.js version     | 20  |
| Typescript version  | ^5.2.2  |
| TypeORM version     | ^0.3.20  |

### Relevant Database Driver(s)

- [ ] aurora-mysql
- [ ] aurora-postgres
- [ ] better-sqlite3
- [ ] cockroachdb
- [ ] cordova
- [ ] expo
- [ ] mongodb
- [ ] mysql
- [ ] nativescript
- [ ] oracle
- [ ] postgres
- [ ] react-native
- [ ] sap
- [ ] spanner
- [ ] sqlite
- [ ] sqlite-abstract
- [ ] sqljs
- [ ] sqlserver

### Are you willing to resolve this issue by submitting a Pull Request?

No, I don’t have the time and I’m okay to wait for the community / maintainers to resolve this issue.

kafka0238:
I also ran into this problem

StPaulis:
I have created [a repository for reproduction](https://github.com/StPaulis/typeorm-cache-repro
) purposes as I also felt on that issue after updating some projects on 0.3.26.

Hope that helps.

Currently, I don’t have the time and I’m okay to wait for the community / maintainers to resolve this issue.

PR #11764, opened 2025-11-12
fix(redis): version detection logic

ibrahimmenem:
add extra check to detect v3.x and take into account legacy mode

fixes #11635
fixes #11749

### Description of change

- Redis 3 client is being detected as v5 because the set.length is returning 0.
- The fix checks if the client has no connect() function and immediately set it as version 3
- also conserves the legacy mode flag to be used in isRedis5OrHigher , and changes the function
name to shouldUsePromiseApi to match what it now does.

- tested locally with redis3 client
- added 2 unit tests

### Pull-Request Checklist

-   [x] Code is up-to-date with the `master` branch
-   [x] This pull request links relevant issues as `Fixes #11635`
-   [x] There are new or updated unit tests validating the change
-   [ ] Documentation has been updated to reflect this change

## Release Notes

* **New Features**
  * Improved Redis compatibility with automatic version detection and support for Redis 3.x, 4.x, and 5+ environments
  * Enhanced query result caching reliability across different Redis versions

* **Tests**
  * Added comprehensive unit tests for Redis version detection and compatibility scenarios

PR #11767, opened 2025-11-13
fix(redis): version detection

AdolfodelSel:
Refs: 1749

fixes https://github.com/typeorm/typeorm/issues/11749

### Description of change

- Read the package to ensure correct detection between Redis v3 and v5 clients.

### Pull-Request Checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Redis version detection for more reliable compatibility across Redis releases; malformed or missing version values now trigger explicit validation and clear errors.
* **Chores**
  * Added a runtime utility to read package version metadata to support consistent and accurate version checks.

### DIFF
--- a/src/driver/better-sqlite3/BetterSqlite3Driver.ts
+++ b/src/driver/better-sqlite3/BetterSqlite3Driver.ts
@@ -138,7 +138,7 @@ export class BetterSqlite3Driver extends AbstractSqliteDriver {
             nativeBinding = null,
             prepareDatabase,
         } = this.options
-        const databaseConnection = this.sqlite(database, {
+        const databaseConnection = new this.sqlite(database, {
             readonly,
             fileMustExist,
             timeout,
@@ -148,8 +148,8 @@ export class BetterSqlite3Driver extends AbstractSqliteDriver {
         // in the options, if encryption key for SQLCipher is setted.
         // Must invoke key pragma before trying to do any other interaction with the database.
         if (this.options.key) {
-            databaseConnection.exec(
-                `PRAGMA key = ${JSON.stringify(this.options.key)}`,
+            databaseConnection.pragma(
+                `key = ${JSON.stringify(this.options.key)}`,
             )
         }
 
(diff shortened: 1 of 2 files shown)


pkuczynski:
This seems to be similar to #11764 but tackling the problem in a different way. Can you elaborate why do you think your solution is better then what's presented in #11764?

AdolfodelSel:
> This seems to be similar to #11764 but tackling the problem in a different way. Can you elaborate why do you think your solution is better then what's presented in #11764?

The proposed change adds a feature that will allow us to check versions in case a similar incompatibility occurs in the future.

In addition, there are several aspects that don’t fully align with the other solution—or with the original one:
	•	Defining a “version” based on a method’s parameters or on the existence or characteristics of that method.
	•	Introducing unnecessary booleans to control the logic.

Overall, I believe this solution is simpler and more reusable.

src/platform/PlatformTools.ts
```diff
@@ -27,6 +27,20 @@ export class PlatformTools {
         return global
     }
 
+    /**
+     * Reads package.json of the given package.
+     * This operation only supports on node platform
```
mguida22:
```suggestion
     * This operation is only supported in node
```

src/cache/RedisQueryResultCache.ts
```diff
@@ -305,21 +305,17 @@ export class RedisQueryResultCache implements QueryResultCache {
...
-            }
-        } catch {
-            // Default to Redis 3/4 for maximum compatibility
+        const version = PlatformTools.readPackageVersion("redis")
+        const major = parseInt(version.split(".")[0], 10)
+        if (isNaN(major)) {
+            throw new TypeORMError(`Invalid Redis version format: ${version}`)
+        }
+        if (major <= 4) {
+            // Redis 3/4 set method requires more parameters (key, value, flag, duration, callback)
```
mguida22, line 314: With your new approach this comment (and the one about version 5) are not very relevant.

Also, it predates your changes here, but it's confusing to set the redis major version to 3 when we know the version is 4. It would be better to set the version to 4 and when we read the `this.redisMajorVersion` we can decide to treat 3 and 4 as the same if needed.

mguida22 (commented):
Thank you @AdolfodelSel!

How have you tested this so far? Can you add tests to the PR to ensure it works properly?

It looks like this will also resolve https://github.com/typeorm/typeorm/issues/11635

ibrahimmenem:
@AdolfodelSel if you don't have time for this please let me know and I can continue with your approach. I'd like to have this fix included in the next release :pray:

ibrahimmenem:
Hi @mguida22 is there a timeout for a PR so it is considered inactive?

mguida22:
> Hi @mguida22 is there a timeout for a PR so it is considered inactive?

No, but we would welcome contributions that build on this. The approach here looks good but it needs tests.

@ibrahimmenem if you have some time to add tests on top of these changes (perhaps as a PR into this one) I can help makes sure it gets across the line.

ibrahimmenem:
@mguida22  I created a new PR https://github.com/typeorm/typeorm/pull/11815  that cherry-picks changes from this PR so they still got the merit of it, then updated comments and added tests

Issue #915, opened 2017-09-20
Missing dep app-root-path for alpha 44


crutchcorn:
Upon creating a new TypeORM alpha 44 project, I hit the following error:

```
TypeORM connection error:  { Error: Cannot find module 'C:\Users\crutc\git\Rucksack\test\node_modules\app-root-path'
    at Function.Module._resolveFilename (module.js:469:15)
    at Function.Module._load (module.js:417:25)
    at Module.require (module.js:497:17)
    at require (internal/module.js:20:19)
    at Function.PlatformTools.load (C:\Users\crutc\git\Rucksack\test\node_modules\typeorm\platform\PlatformTools.js:37:24)
    at ConnectionOptionsReader.get [as baseDirectory] (C:\Users\crutc\git\Rucksack\test\node_modules\typeorm\connection\ConnectionOptionsReader.js:194:50)
    at ConnectionOptionsReader.get [as baseFilePath] (C:\Users\crutc\git\Rucksack\test\node_modules\typeorm\connection\ConnectionOptionsReader.js:182:24)
    at C:\Users\crutc\git\Rucksack\test\node_modules\typeorm\connection\ConnectionOptionsReader.js:102:81
    at Array.find (native)
    at ConnectionOptionsReader.<anonymous> (C:\Users\crutc\git\Rucksack\test\node_modules\typeorm\connection\ConnectionOptionsReader.js:101:87) code: 'MODULE_NOT_FOUND' }
```

Manually `npm install`ing `app-root-path` fixed the problem

pleerock:
Thank you for the report, fixed in `typeorm@0.1.0-alpha.45`

PR #11769, opened 2025-11-14
refactor: replace uuid with native Crypto API

### LINKED ISSUE #11693: Drop dependencies in favour of native crypto functionality
We use a number of dependencies which can be dropped and replaced with built-in cryptographic functionality that is now part of Node/web platform:

- remove `sha.js` (use crypto Web API instead) #11769
- remove `uuid` in favour of builtin crypto module
- replace built-in hashing function with builtin crypto module

mag123c:
### Description of change
This change by removing unnecessary external cryptographic dependencies (sha.js and uuid) and replacing them with native Crypto API functionality available in modern Node.js and browser environments.

### Pull-Request Checklist

-   [x] This pull request links relevant issues as Fixes #11693 

## Summary by CodeRabbit

* **Chores**
  * Removed the external UUID dependency from the project.

* **Refactor**
  * Centralized UUID generation into an internal utility and replaced dispersed direct UUID usage with the new provider.
  * Implemented a cross-platform UUID v4 generator that uses native crypto where available and falls back to a deterministic method for broader reliability.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>

### DIFF
--- a/src/platform/PlatformTools.ts
+++ b/src/platform/PlatformTools.ts
@@ -45,6 +45,7 @@ export class PlatformTools {
     /**
      * Loads ("require"-s) given file or package.
      * This operation only supports on node platform
+     * @param name
      */
     static load(name: string): any {
         // if name is not absolute or relative, then try to load package from the node_modules of the directory we are currently in
@@ -156,6 +157,7 @@ export class PlatformTools {
 
     /**
      * Normalizes given path. Does "path.normalize" and replaces backslashes with forward slashes on Windows.
+     * @param pathStr
      */
     static pathNormalize(pathStr: string): string {
         let normalizedPath = path.normalize(pathStr)
@@ -166,6 +168,7 @@ export class PlatformTools {
 
     /**
      * Gets file extension. Does "path.extname".
+     * @param pathStr
      */
     static pathExtname(pathStr: string): string {
         return path.extname(pathStr)
@@ -173,6 +176,7 @@ export class PlatformTools {
 
     /**
      * Resolved given path. Does "path.resolve".
+     * @param pathStr
      */
     static pathResolve(pathStr: string): string {
         return path.resolve(pathStr)
@@ -180,6 +184,7 @@ export class PlatformTools {
 
     /**
      * Synchronously checks if file exist. Does "fs.existsSync".
+     * @param pathStr
      */
     static fileExist(pathStr: string): boolean {
         return fs.existsSync(pathStr)
@@ -199,8 +204,8 @@ export class PlatformTools {
 
     /**
      * Loads a dotenv file into the environment variables.
-     *
      * @param path The file to load as a dotenv configuration
+     * @param pathStr
      */
     static dotenv(pathStr: string): void {
         dotenv.config({ path: pathStr })
@@ -208,6 +213,7 @@ export class PlatformTools {
 
     /**
      * Gets environment variable.
+     * @param name
      */
     static getEnvVariable(name: string): any {
         return process.env[name]
@@ -215,6 +221,7 @@ export class PlatformTools {
 
     /**
      * Highlights sql string to be printed in the console.
+     * @param sql
      */
     static highlightSql(sql: string) {
         return highlight(sql, {
@@ -234,6 +241,8 @@ export class PlatformTools {
 
     /**
      * Pretty-print sql string to be print in the console.
+     * @param sql
+     * @param dataSourceType
      */
     static formatSql(sql: string, dataSourceType?: DatabaseType): string {
         const databaseLanguageMap: Record<


alumni (changes requested):
Can break platforms that don't support crypto. Read the discussion in https://github.com/typeorm/typeorm/issues/11693

Also: the internal sha1 implementation has a flaw (when whitespace chars are in the input) - so this is a breaking change that cannot be merged into master (we use master for the 0.3 version).

mag123c:
@alumni 
Oops, I'll try to change it that no breaking changes occur. Thanks for comment.

mag123c:
@alumni
Thanks for the feedback! I'd appreciate your review to confirm this approach is correct.      

## Changes Made
**Platform Compatibility**
- Added multi-level fallback for UUID generation:
- Primary: `crypto.randomUUID()` (Node.js, modern browsers)
- Fallback 1: `crypto.getRandomValues()` (older browsers, React Native)
- Fallback 2: `Math.random()` (last resort)

Updated both Node and Browser PlatformTools implementations

**SHA-1 Breaking Change**
- **Approach taken**: Use `RandomGenerator.sha1()` on ALL platforms instead of native crypto
- **Reason**: Avoids breaking existing database constraint names
- **Trade-off**: Keeps the whitespace handling behavior (encodeURIComponent) for consistency

Please let me know if this direction is acceptable or if you'd prefer a different approach.

alumni:
So first of all, if there's any breaking change, we would like to do this in the next major version of TypeORM, which is developed in the `next` branch (instead of `master`).

The Crypto WebAPI (`crypto.randomUUID()`) should work in nodejs/browsers (at least in recent versions of NodeJS). However, it may not work on every platform, so it would be great if we implemented our own UUIDv4 function. Not sure why `crypto.getRandomValues()` and `Math.random()` would be needed. _Later edit:_ Ah, I see `Math.random` is used for the UUID generation. But that will no longer be a UUIDv4. I'd rather copy the implementation from the uuid library to maintain compatibility.

We could add a `uuidv4` function to `RandomGenerator` that calls the native implementation if it exists and falls back to the custom implementation otherwise.

For SHA1, we can't use native crypto since it's async, so we can use our own implementation. Our implementation has a bug with certain characters (missing `unescape` in `src/util/RandomGenerator.ts:44` IIRC). So `sha.js`/native crypto and our implementation have different output when some special characters are in the input. If we fix our implementation we might generate a lot of migrations for e.g. check constraints. If we don't, we might generate migrations everywhere else.

I think the UUIDv4 change is not breaking, but the SHA1 will be.

mag123c:
@alumni 
Thank you for the detailed feedback! I've updated the implementation based on your review

## Changes Made

**UUID Implementation**
- Moved UUID generation to `RandomGenerator.uuidv4()` as suggested
- Native `crypto.randomUUID()` is used when available
- Fallback to `crypto.getRandomValues()` with proper RFC 4122 implementation
- **Removed `Math.random()` fallback** - now throws error instead (as you noted, Math.random would not be RFC 4122 compliant)
- Successfully removed `uuid` dependency

**SHA1 Scope Reduction**
- **Reverted all SHA1-related changes** from this PR


---

Should I update the PR title to reflect the reduced scope (e.g., "refactor: replace uuid with native Crypto API")? Or keep the current title and clarify in comments?

And,  **May I proceed with creating a PR targeting the `next` branch for SHA1 removal?**
If I could be, Are there any specific guidelines or considerations I should be aware of when contributing to the next major version?

Looking forward to your feedback!

alumni:
@mag123c a few more things: the lockfile needs updating after `sha.js` was added back. Also the PR title should be updated.

Looking at the source code of `uuid`, I also wonder how it worked on React Native - `crypto` is not available in Hermes, not even `getRandomValues`, so probably a good fallback would be to create the array and set each element to a random value, like `react-native-uuid` does in `rng` (and similar to what was here before). The UUIDv4 format is achieved via the bitwise `&` on bytes 6 and 8, I think.

mag123c:
@alumni Thank you for the detailed feedback on React Native/Hermes support!

### Updates Made
- **package-lock.json** updated after restoring sha.js dependency
- **Math.random() fallback** added to RandomGenerator.ts for Hermes compatibility

---

Before implementing, I conducted research to ensure this is the optimal approach

**1. Hermes crypto support**
- Confirmed via [Hermes issue #915](https://github.com/facebook/hermes/issues/915) that crypto APIs are officially out of scope       
- Hermes team: "This falls outside of the JavaScript specification... should be addressed at the React Native level"

**2. RFC 4122 compliance**
- Section 4.4 explicitly permits **"randomly (or pseudo-randomly) chosen values"**
- Math.random() is compliant for UUID v4 generation

**3. Others**
- [expo-crypto](https://docs.expo.dev/versions/latest/sdk/crypto/) uses identical Math.random() fallback

mag123c:
Fixed missing `socks@2.8.7` dependency in package-lock.json.

The entry was accidentally removed during a previous lock update when npm pruned the optional package. Manually restored from master's lock file.

pkuczynski:
@mag123c this should be done against `next` branch as it's planned for 1.0. I just changed the base branch for PR. Can you solve conflicts?

alumni:
@pkuczynski I think here it's a problem of how we created the tasks, we should probably create 3 of them (shajs, uuid, and buffer), not just one :P 

However, if `globalThis.crypto` is a headache in TS with the old versions of node, we could as well do it for the next major 🤷🏻‍♂️ I think this is what I was afraid of when I added it to the next major TODO :)

mag123c:
I've resolved the conflicts after the base branch was changed to `next`

pujux:
seems the lockfile needs an update so the checks run correctly :) @mag123c

mag123c:
@pujux Oh, thanks for mention.

Conflict was solved. @pkuczynski @alumni

src/util/RandomGenerator.ts
```diff
@@ -1,4 +1,78 @@
...
+    /**
+     * Generates a RFC 4122 version 4 UUID.
+     * Uses native crypto.randomUUID() if available, otherwise falls back to
+     * a custom implementation using crypto.getRandomValues().
+     *
+     * @returns A version 4 UUID string
+     */
+    static uuidv4(): string {
+        // Try Node.js native crypto
+        if (typeof process !== "undefined" && process.versions?.node) {
```
alumni: Should be removed. `globalThis.crypto` should exist in all node versions that we support: [Crypto: randomUUID() method - Web APIs | MDN](https://developer.mozilla.org/en-US/docs/Web/API/Crypto/randomUUID#browser_compatibility)

src/util/RandomGenerator.ts
```diff
@@ -1,4 +1,78 @@
...
+            typeof globalThis !== "undefined" &&
+            globalThis.crypto &&
+            typeof globalThis.crypto.randomUUID === "function"
+        ) {
+            try {
+                return globalThis.crypto.randomUUID()
+            } catch (e) {
+                // Fall through to custom implementation
+            }
+        }
```
alumni: Let's use more modern JS, if TS allows it. This applies in other places in this PR, but I'll only write it here.

```suggestion
        const uuid = globalThis.crypto?.randomUUID?.()
        if (uuid) {
           return uuid
        }
```

src/platform/PlatformTools.ts
```diff
@@ -299,4 +287,13 @@ export class PlatformTools {
         console.log(ansi.black.bgRed(prefix))
         if (err) console.error(err)
     }
+
+    /**
+     * Generates UUID v4 using native crypto API with fallback for environments
+     * that don't support it (e.g., React Native, Hermes).
+     */
+    static generateUuid(): string {
```
alumni: This would make more sense for SHA-1 which we already merged CC: @G0maa 

But for UUIDv4, since the implementation is generic, we probably don't need to do it, we can use the implementation directly, without having to go through `PlatformTools`.

src/platform/PlatformTools.ts
```diff
@@ -264,8 +254,6 @@ export class PlatformTools {
 
     /**
      * Logging functions needed by AdvancedConsoleLogger
-     * @param prefix
-     * @param info
```
alumni, line 268: Let's keep the TSDoc for now

src/platform/PlatformTools.ts
```diff
@@ -30,7 +30,6 @@ export class PlatformTools {
     /**
      * Reads the version string from package.json of the given package.
      * This operation is only supported in node.
-     * @param name
```
G0maa, line 33: I think these `JSDoc` params raise eslint warnings when not-existent. So if we're not going to fix them (understandable, unrelated to the PR) we shouldn't delete them. wdyt?

Screenshot from 2026-01-04 12-52-04

image

src/util/RandomGenerator.ts
```diff
@@ -1,4 +1,54 @@
...
+     * a custom implementation using crypto.getRandomValues().
+     *
+     * @returns A version 4 UUID string
+     */
+    static uuidv4(): string {
+        // Try native crypto.randomUUID() (available in Node.js 19+ and modern browsers)
+        const uuid = globalThis.crypto?.randomUUID?.()
+        if (uuid) {
+            return uuid
+        }
```
G0maa, line 14: nitpick(needn't to fix): it seems that this type of logic is going to repeat a lot, should we abstract it somewhere (e.g. `PlatformTools`?) @alumni

src/util/RandomGenerator.ts
```diff
@@ -1,4 +1,54 @@
...
+
+        // Custom implementation using crypto.getRandomValues()
+        // Based on RFC 4122 version 4 UUID specification
+        const randomBytes = new Uint8Array(16)
+
+        if (globalThis.crypto?.getRandomValues) {
+            globalThis.crypto.getRandomValues(randomBytes)
+        } else {
```
G0maa, line 22: which environment is going to use this branch? (Browser, Node, RN, ...?)

G0maa (approved):
Few nitpicks, seems generally ok.

src/util/RandomGenerator.ts
G0maa: I see [your](https://github.com/typeorm/typeorm/pull/11769#discussion_r2657931168) comment references this above, so just to clarify: It seems in different places we have to do the same logic: `If node do x, if browser do y, if RN do z.` so I was suggesting a "standarized" way of doing this somewhere. i.e. such that every PR uses the same method.

src/platform/PlatformTools.ts
alumni: Yeah, this is what I mentioned in an earlier comment. I would revert the changes to this file.

src/util/RandomGenerator.ts
alumni: Anything that does not support the Crypto WebAPI (or doesn't support it properly).

There are problems in Hermes (RN/Expo) as I understand.

src/util/RandomGenerator.ts
alumni: I'd keep it as it is - we don't need to do things based on which platform we're running, but based on what features the platform provides. So if the platform provides the `randomUUID` function, we should use it.

Over time we should get rid of `PlatformTools` if possible.

src/util/RandomGenerator.ts
```diff
@@ -1,4 +1,54 @@
...
+            // Fallback for React Native/Hermes and environments without crypto support
+            // Hermes (React Native's JavaScript engine) does not provide crypto APIs
+            // Math.random() is permitted by RFC 4122 for UUID v4 ("pseudo-randomly")
+            // This approach is also used by expo-crypto and other RN libraries
+            // Note: For TypeORM's use case (DB IDs, cache IDs), uniqueness is sufficient
+            for (let i = 0; i < 16; i++) {
+                randomBytes[i] = Math.floor(Math.random() * 256)
```
smith-xyz, line 29: is this enough entropy for uniqueness? I imagine its generally okay but thought I'd ask out of interest 😄

mag123c: Yes, for uniqueness it's sufficient. This fallback is only for RN/Hermes environments, and was already discussed with @alumni above

G0maa:
I think this PR is ready @alumni

alumni:
> I think this PR is ready @alumni

Our comments in PlatformTools were not solved entirely :P The PR is still removing the TSDoc that you recently added :P

PR #11774, opened 2025-11-17
fix(postgres): handle `timestamptz` persistence/hydration correctly

Minishlink:
### Description of change

This includes 'timestamptz' as type for values which are Date and should be serialized/deserialized in/from the cache.

### Pull-Request Checklist

-   [x] This pull request links relevant issues as `Fixes #00000` (searched for timestamptz)

## Summary by CodeRabbit

* **New Features**
  * Added support for PostgreSQL timestamptz columns, enabling proper handling of timezone-aware timestamps.

* **Tests**
  * Updated Postgres tests to cover timestamptz handling, enable query caching in test runs, and add runtime assertions verifying Date values and equality.


Minishlink:
Sure, I added a specific test for timestamptz and fixed the existing test case which was not really testing persistence/hydration since the data was not fetched from the cache

test/functional/database-schema/column-types/postgres/column-types-postgres.ts
```diff
@@ -15,6 +15,7 @@ describe("database schema > column types > postgres", () => {
         connections = await createTestingConnections({
             entities: [__dirname + "/entity/*{.js,.ts}"],
             enabledDrivers: ["postgres"],
+            cache: true,
```
G0maa: Why enable this?

G0maa (commented):
Hello @Minishlink, could you update the branch?

Furthermore, from the [docs](https://www.postgresql.org/docs/current/datatype-datetime.html#:~:text=The%20SQL%20standard%20requires%20that%20writing%20just%20timestamp%20be%20equivalent%20to%20timestamp%20without%20time%20zone%2C%20and%20PostgreSQL%20honors%20that%20behavior.%20timestamptz%20is%20accepted%20as%20an%20abbreviation%20for%20timestamp%20with%20time%20zone%3B%20this%20is%20a%20PostgreSQL%20extension.) I noticed that `timestamptz` is the same as `timestamp with timezone`,

also it seems like there were already some handling for this [here](https://github.com/typeorm/typeorm/blob/master/src/driver/postgres/PostgresDriver.ts#L1012-L1013).

so what does this PR fix?

test/functional/database-schema/column-types/postgres/column-types-postgres.ts
G0maa: I'll assume two things, so correct me if I'm wrong.

1. `timestamptz` column types were not cached correctly as the type was not in `preparePersistentValue` and `prepareHydrateValue`, so this PR fixes that. 
2. The `cache: true` is to verify caching works for said column type.

If both assumptions are true then I'd prefer if caching was enabled separately in another test, since this particular file tests column types, not caching.

I've looked at other `column-types-x.ts` files to see if they have caching enabled and they don't.

See:
https://github.com/typeorm/typeorm/blob/master/test/functional/database-schema/column-types/mysql/column-types-mysql.ts#L16-L19
https://github.com/typeorm/typeorm/blob/master/test/functional/database-schema/column-types/cockroachdb/column-types-cockroach.ts#L23-L27

G0maa: I'm interested to know how this handling [here](https://github.com/typeorm/typeorm/blob/master/src/driver/postgres/PostgresDriver.ts#L1010-L1011) affects this whole process though :grin:

Minishlink: Yes you're right. I assumed that persistence/hydration was linked to cache?

G0maa: I'm confused too :thinking:

I am _assuming_ that persistence/hydration works regardless of caching. i.e. data serialization/deserialization form/to database/ORM.

But then, why does `timestamptz` work without caching and fails when caching? i.e. if the issue is serialization/deserialization it should fail in both cases.

It might be that [this](https://github.com/typeorm/typeorm/blob/master/src/driver/postgres/PostgresDriver.ts#L1010-L1011) logic here is executed somewhere which fixes the issue when caching is not used.

G0maa: I've made a small sample [here](https://github.com/G0maa/typeorm-bug-reproduce/tree/43407c7eb6feda7300846e566ec65aab11fc09d0), should fail if you enable caching, should success if you disable caching.

alumni: > Yes you're right. I assumed that persistence/hydration was linked to cache?

Nope. Cache has nothing do to with hydration.

Could we undo this line and merge master to see that the tests are still passing?

It should affect anything now, but we don't want any additional functionality affecting the result of the test in the future.

alumni (changes requested):
The `PostWithOptions` entity is reserved for data types that have parameters.

We already had `Post.timestamptz` and similar tests in `test/functional/database-schema/column-types/postgres/column-types-postgres.ts` lines `167-169` and `291-293` ("all types should work correctly - persist and hydrate").

So unless I'm missing something, the new test doesn't seem to be testing the change in this PR.


I would add the necessary assertions to the test above (for `Post.timestamptz`)

G0maa:
Hello @Minishlink, you could check my reproduction at https://github.com/G0maa/typeorm-bug-reproduce/blob/43407c7eb6feda7300846e566ec65aab11fc09d0/src/app.ts for this PR, 

If you uncomment the `cache: true`, the bug would happen, it seems for some reason hydrate/persist isn't done correctly when using caching for `timestamptz`.

Minishlink:
https://github.com/typeorm/typeorm/pull/11916

alumni:
~In Postgres, `timestamptz` is just an alias for `timestamp with time zone`. I need to check how PG reports it to avoid constant generation of migrations.~

_Later edit:_ Never mind, we already convert it to the actual column type in `normalizeType`

PR #11782, opened 2025-11-19
feat: add `returning` option to update/upsert operations

naorpeled:
## Summary
- allow repository/entity manager update APIs to accept returning options when supported
- wire returning option through upsert/update/updateAll and expose typings
- add coverage and docs for the new feature

### DIFF
--- a/.pr_agent.toml
+++ b/.pr_agent.toml
@@ -2,13 +2,19 @@
 pr_commands = [
     "/review",
     "/improve",
+    "/agentic_review"
 ]
 
 handle_push_trigger = true
 push_commands = [
     "/improve",
+    "/agentic_review"
 ]
 
 [auto_best_practices]
 enable_auto_best_practices = true
-utilize_auto_best_practices = true 
+utilize_auto_best_practices = true
+
+[review_agent]
+enabled = true
+publish_output = true
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -116,14 +116,14 @@ services:
   # sap hana
   # works only on linux, minimum 10GB RAM for docker required
   hanaexpress:
-    image: "saplabs/hanaexpress:2.00.076.00.20240701.1"
+    image: "saplabs/hanaexpress:2.00.088.00.20251110.1"
     container_name: "typeorm-hanaexpress"
     hostname: hxe
     command:
       [
+        "--agree-to-sap-license",
         "--passwords-url",
         "file:////hana/hxe-config.json",
-        "--agree-to-sap-license",
       ]
     ulimits:
       nofile: 1048576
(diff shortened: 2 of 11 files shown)


src/index.ts
```diff
@@ -115,6 +115,8 @@ export * from "./repository/TreeRepository"
 export * from "./repository/MongoRepository"
 export * from "./repository/RemoveOptions"
 export * from "./repository/SaveOptions"
+export { UpdateOptions as RepositoryUpdateOptions } from "./repository/UpdateOptions"
```
naorpeled, line 118: There's a conflict with UpdateOptions from Mongo typings,
what do you think we should do about this? 🤔 

Should we rename Mongo's type to MongoUpdateOptions?

gioboa: Yeah, that makes sense to me

naorpeled: For v0.3 I'll keep this implementation, and towards v1 I'll see how we can clean it up based on what we do with Mongo

naorpeled:
@gioboa @alumni @G0maa 
wdyt?🙏

docs/docs/working-with-entity-manager/6-repository-api.md
```diff
@@ -141,13 +141,26 @@ await repository.update({ age: 18 }, { category: "ADULT" })
...
 -   `updateAll` - Updates _all_ entities of target type (without WHERE clause). Sets fields from supplied partial entity.
 
 ```typescript
 await repository.updateAll({ category: "ADULT" })
 // executes UPDATE user SET category = ADULT
+
+await repository.updateAll(
+    { category: "ADULT" },
+    { returning: "*" }, // limited to drivers that support returning clauses
+)
```
gioboa, line 163: What is the log in this case ?

naorpeled: The fully changed entity

docs/docs/working-with-entity-manager/6-repository-api.md
```diff
@@ -141,13 +141,26 @@ await repository.update({ age: 18 }, { category: "ADULT" })
...
 await repository.update(1, { firstName: "Rizzrak" })
 // executes UPDATE user SET firstName = Rizzrak WHERE id = 1
+
+// optionally request RETURNING / OUTPUT values (supported drivers only)
+const result = await repository.update(
+    1,
+    { firstName: "Rizzrak" },
+    { returning: ["id", "firstName"] },
+)
+console.log(result.raw) // [{ id: 1, firstName: "Rizzrak" }]
```
G0maa, line 151: As a stretch goal in the future, any idea how could we make this type-safe? i.e. `result.id` & `result.firstName` become defined on `result`.... somehow.

test/functional/repository/returning/repository-returning.ts
```diff
@@ -0,0 +1,99 @@
+import { expect } from "chai"
```
gioboa, line 1: The test is not ending with .test.ts 🙈 so it's not running

PR #11801, opened 2025-11-24
refactor(tests): ensure test files have the `.test.ts` extension

alumni:
### Description of change

Enforce all test files to have the `.test.ts` extension, so that they are better separated from other files in the `test` folder. This will allow our test runner (currently `mocha`) to search for less files. This is also a prerequisite to replacing `mocha` with e.g. `vitest` in the future, as other test runners will complain about test files that don't contain tests.

I will update the PR if we decide for `.spec.ts` instead of `.test.ts`. Please check the poll in our technical discussion channel.

### Pull-Request Checklist

### DIFF
--- a/.c8rc.json
+++ b/.c8rc.json
@@ -1,7 +1,7 @@
 {
   "all": true,
   "cache": false,
-  "exclude": ["node_modules", "**/*.d.ts"],
+  "exclude": ["**/*.d.ts"],
   "exclude-after-remap": true,
   "extension": [".ts"],
   "include": ["build/compiled/src/**", "src/**"],
--- a/.mocharc.json
+++ b/.mocharc.json
@@ -1,10 +1,10 @@
 {
-  "__comment": "TODO: remove --exit flag: https://mochajs.org/#-exit",
-  "exit": true,
+  "$schema": "https://json.schemastore.org/mocharc",
   "check-leaks": true,
   "color": true,
+  "exit": true,
   "file": ["./build/compiled/test/utils/test-setup.js"],
   "recursive": true,
-  "spec": ["./build/compiled/test"],
+  "spec": ["./build/compiled/test/**/*.test.{js,ts}"],
   "timeout": 90000
 }
--- a/test/functional/view-entity/view-dependencies/view-entity-dependencies.test.ts
+++ b/test/functional/view-entity/view-dependencies/view-entity-dependencies.test.ts
@@ -10,7 +10,8 @@ import { ViewA } from "./entity/ViewA"
 import { ViewB } from "./entity/ViewB"
 import { TestEntity } from "./entity/Test"
 
-describe("views dependencies", () => {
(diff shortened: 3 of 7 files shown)


test/functional/view-entity/view-dependencies/view-entity-dependencies.test.ts
```diff
@@ -10,7 +10,8 @@ import { ViewA } from "./entity/ViewA"
 import { ViewB } from "./entity/ViewB"
 import { TestEntity } from "./entity/Test"
 
-describe("views dependencies", () => {
+// TODO: Implement topological sorting for view dependencies https://github.com/typeorm/typeorm/issues/8240
+describe.skip("views dependencies", () => {
```
alumni, line 14: I noticed this test started to fail after this change, and for very good reason: this feature was not implemented correctly.

Since the purpose of this PR is to simply fix the file names, I will leave the feature to be fixed separately.

.mocharc.json
```diff
@@ -1,10 +1,10 @@
 {
-  "__comment": "TODO: remove --exit flag: https://mochajs.org/#-exit",
```
OSA413, line 2: Do we need to document the current use of the `exit` flag?

alumni:
Yes, `grep` and a one-line `while` loop :) 

I noticed some files are named `bla-bla-test.test.ts`, that's because they were named `bla-bla-test.ts` before. I'll adjust them.

Will change `test.ts` to `spec.ts` if the popular vote goes that way.

If we pick the BDD style (or not), we could also ask for new tests to be in the Given-When-Then (aka Arrange-Act-Assert). Existing tests don't necessarily follow the format.

.c8rc.json
```diff
@@ -1,7 +1,7 @@
 {
   "all": true,
   "cache": false,
-  "exclude": ["node_modules", "**/*.d.ts"],
+  "exclude": ["**/*.d.ts"],
```
gioboa, line 4: Is it intended?🤔

.mocharc.json
alumni: Hey, we could change the file format to jsonc for comments, but then we'd have to adjust the lintstaged config to handle that too.

Maybe the comment is not really needed, and since we were anyway discussing to move to something else (vitest was one of the proposal), probably this file won't exist for too long.

I was hoping the json schema for mocha has documentation (it doesn't), but honestly, seeing the commit message from your PR was more helpful why we added the option than the documentation/comment.

.c8rc.json
alumni: Yes, that glob pattern is incorrect, it doesn't do anything.

There are no `node_modules` in the `include` patterns, but if we wanted to exclude them, it should have been `**/node_modules/**` - or simply add `excludeNodeModules: true` to the options (undocumented for `c8`, but documented for [test-exclude](https://www.npmjs.com/package/test-exclude) which `c8`/`nyc` use).

Issue #11793, opened 2025-11-22
Add deferrable support to @Exclusion decorator (mirror @Unique & @Index)


oGAD31:
### Feature Description

@Unique supports the deferrable option, which allows specifying constraint deferrability in PostgreSQL (e.g., "INITIALLY DEFERRED" or "INITIALLY IMMEDIATE").

However, @Exclusion does not currently support this option, even though PostgreSQL allows deferrable exclusion constraints.

This feature request proposes adding deferrable support to the @Exclusion decorator in a way consistent with @Unique.

### The Solution

Add another parameter to the @Exculision decorator for exclusion options which (for now) will support only the deferrable option (mimicking @Unique)  

### Considered Alternatives

In my opinion this is a pretty straight forward feature which doesn't really have another solution. Just copying an already existing feature from the @Unique and @Index decorators to the @Exclusion feature which does support the defer options in postgres.

### Additional Context

I decided to open this issue (and is willing to implement the feature myself) as a result of needing it in a project i work on. IMO it makes a lot of sense to add it since its supported natively in pg.

### Relevant Database Driver(s)

- [ ] aurora-mysql
- [ ] aurora-postgres
- [ ] better-sqlite3
- [ ] cockroachdb
- [ ] cordova
- [ ] expo
- [ ] mongodb
- [ ] mysql
- [ ] nativescript
- [ ] oracle
- [x] postgres
- [ ] react-native
- [ ] sap
- [ ] spanner
- [ ] sqlite
- [ ] sqlite-abstract
- [ ] sqljs
- [ ] sqlserver

### Are you willing to resolve this issue by submitting a Pull Request?

Yes, I have the time, and I know how to start.

G0maa:
I see that you have the time and know where to start, could you go ahead and implement it?

We appreciate your contribution.

When you submit your PR do mention me, I can do an early review :)

oGAD31:
@G0maa will do :)

gioboa:
Closed by #11802

PR #11802, opened 2025-11-24
feat: add deferrable support to exclusion decorator to mirror unique and index decorators

oGAD31:
## **User description**

Add deferrable support to @Exclusion decorator (mirror @Unique & @Index), as described in #11793

### Pull-Request Checklist


oGAD31:
@G0maa FYI :)

src/driver/postgres/PostgresQueryRunner.ts
```diff
@@ -4114,7 +4114,10 @@ export class PostgresQueryRunner
                               table,
                               exclusion.expression!,
                           )
-                    return `CONSTRAINT "${exclusionName}" EXCLUDE ${exclusion.expression}`
+                    let constraint = `CONSTRAINT "${exclusionName}" EXCLUDE ${exclusion.expression}`
+                    if (exclusion.deferrable)
+                        constraint += ` DEFERRABLE ${exclusion.deferrable}`
```
OSA413, line 4189: Have you tested it with different inputs?

test/functional/deferrable/entity/Booking.ts
```diff
@@ -0,0 +1,21 @@
+import {
+    Entity,
+    Column,
+    PrimaryGeneratedColumn,
+    Exclusion,
+} from "../../../../src"
+
+@Entity()
+@Exclusion(`USING gist (tstzrange("from", "to") WITH &&)`, {
+    deferrable: "INITIALLY DEFERRED",
```
OSA413, line 10: I would also add an entity with an `@Exclusion` without specifying `deferrable` to the test.

OSA413:
And yeah, what about migrations? What will happen when I change `deferrable` type or remove it?

test/functional/deferrable/deferrable-exclusion.ts
```diff
@@ -0,0 +1,119 @@
...
+                    // then save booking with exclusion violation
+                    const booking2 = new Booking()
+                    booking2.id = 2
+                    booking2.from = new Date("2024-01-05")
+                    booking2.to = new Date("2024-01-15")
+
+                    await manager.save(booking2)
+
+                    // Fix overlap before commit
+                    booking2.from = new Date("2024-01-11")
```
OSA413: What is `overlap` in this scenario exactly?

src/driver/postgres/PostgresQueryRunner.ts
OSA413: Looks like not. Please, check the error message and fix the tests/functionality.

test/functional/deferrable/deferrable-exclusion.ts
oGAD31: booking1 ended on the 2024-01-10 and booking2 started on the 2024-01-05 which means booking2 started in the middle of booking1. which means the ranges overlap.

test/functional/deferrable/entity/Booking.ts
oGAD31: Are you sure? these tests are meant to test the deferrable functionality only (as it seems from the name of the dir). Maybe i should add another test for `@Exclusion` specifically somewhere else 🤔

src/driver/postgres/PostgresQueryRunner.ts
oGAD31: what would you say is the wanted output  for when a user enters an unexpected input?

test/functional/deferrable/entity/Booking.ts
OSA413: can you please check if there's no deferrable in @Exclusion somewhere in the tests? I just want all the cases to be tested (two different values and undefined as not provided value)

OSA413: can you please check if there's no deferrable in @Exclusion somewhere in the tests? I just want all the cases to be tested (two different values and undefined as not provided value)

oGAD31:
Regarding migrations: as i was working on the feature I noticed that changing ONLY the deferrable value (or removing it) after the exclusion was already created does not change it in the db. I tested it on the `@Unique` as well and got the same result so i figured fixing this is outside the scope of this issue.

@OSA413 @G0maa

src/driver/postgres/PostgresQueryRunner.ts
OSA413: no, user can input only the values provided by the string literal type, I'm asking if you tested it locally because CI tests are failing on the new tests

test/functional/deferrable/entity/Booking.ts
oGAD31: There are, see test/functional/query-runner/entity/Post.ts for example.

src/driver/postgres/PostgresQueryRunner.ts
oGAD31: TBH i cant figure out whats causing the tests to fail. i tested the feature locally and it seems to work with all inputs (two different values and undefined as not provided value). I get the same error when running the "deferrable unique constraint" and "deferrable foreign key constraint" as well 🤔

oGAD31: I made a new commit which I think solves the issue.

@OSA413

OSA413:
In order to properly fix the tests, it needs to be known:
1) Is Exclusion supported by all DBs?
2) Is Exclusion deferrable supported by all DBs?

OSA413:
I would also add the check against allowlist at runtime as qodo suggests. @alumni what do you think?

alumni:
> Is Exclusion supported by all DBs?

At least in TypeORM, only the PostgresDriver implements exclusion constraints. Not sure if there are other DBs supporting them, they seem to be overlapping in functionality with other constraints.

> Is Exclusion deferrable supported by all DBs?

Quite a lot of DBs support deferrable constraints and a lot of our drivers support them too. If it's supported for e.g. referential constraints (FKs), it should be supported for unique constraints and check constraints too.

> I would also add the check against allowlist at runtime as qodo suggests. @alumni what do you think?

I would say it's not really necessary since it's validated by TypeScript types, so you'd have to cast, which means you have to really want to do something different. It's not validated for other types of constraints either.

oGAD31:
I will summaries why the tests keep failing and 2 possible solutions: the tests failed for the first time because i tried to use `@Exclusion` on type `timestamp without time zone`. this type isnt supported for overlap exclusions so i switched to `timestamp with time zone` (AKA timestamptz. This caused the the tests to fail because better-sqlite3,sap and sqlite DO NOT support type timestampz (but they do support deferrable on FK so they are included in  the enabledDrivers array of test file test/functional/deferrable/deferrable.ts)

2 possible ways to fix the tests:
1. rewrite the tested entities (`Booking `& `Schedule`)to not use timestampz (I can test overlaps with just regular numbers instead of dates...)
2. not include the 2 new entities in the `entities` array of file test/functional/deferrable/deferrable.ts

IMO option 1 is the better one, tell me what you think @OSA413.

OSA413:
> IMO option 1 is the better one

Yes, I think this is also better

oGAD31:
> > IMO option 1 is the better one
> 
> Yes, I think this is also better

Done :)

test/functional/deferrable/deferrable-exclusion.test.ts
```diff
@@ -0,0 +1,119 @@
+import "reflect-metadata"
```
gioboa, line 1: tests must have `.test.ts` extension to be considered. can you fix it?

gioboa (changes requested):

test/functional/deferrable/deferrable-exclusion.test.ts
oGAD31: done 😄

G0maa:
@oGAD31 


Why CI fails:
```
The description "add deferrable support to Exclusion decorator (mirror Unique & Index)" found in this pull request title "feat: add deferrable support to Exclusion decorator (mirror Unique & Index)"
didn't match the configured pattern. Please ensure that it doesn't
start with an uppercase character and contain only allowed characters.
```

tl;dr the title contains upper-case characters.

oGAD31:
fixed @G0maa

oGAD31:
Hi @gioboa, not sure how to update the docs. Also, shouldn't i fix the first 2 bugs found by Qodo in the review?

gioboa:
Yep, please have a look at the suggestions if those are correct or not.

oGAD31:
Well i fixed the first bug but the other ones seems to exist for other decorators as well (`@Unique` for example) so im not sure if I should fix them.

@gioboa

oGAD31:
What's currently stopping the pr from being approved? Can you approve it @gioboa ?

I already got an approve from @G0maa

PR #11807, opened 2025-11-27
fix: typesense doc sync

### LINKED ISSUE #11799: Fix Index Docs To Typesense GitHub action
This action:
https://github.com/typeorm/typeorm/actions/workflows/docsearch.yml

seems to be failing occasionally, I've noticed it's something related to
`typesense.exceptions.ObjectUnprocessable: [Errno 422] Rejecting write: running out of resource type: OUT_OF_MEMORY`

As seen in this run:
https://github.com/typeorm/typeorm/actions/runs/19625208751/job/56192680136

@G0maa Will give it a look if I have some time, in the mean time if anyone's interested @ me and I'll assign you.

G0maa:
### Description of change

Currently typesnese index workflow fails [a lot](https://github.com/typeorm/typeorm/actions/runs/19670738815/job/56353614557) due to limited memory, this PR helps optimize memory usage a bit by:
1. Deleting old (failed) collections before syncing.
2. Deleting unused facets. [Facets](https://typesense.org/docs/29.0/api/collections.html#schema-parameters:~:text=For%20example%2C%20if%20you%20are%20building%20a%20product%20search%2C%20color%20and%20brand%20could%20be%20defined%20as%20facet%20fields.%20Once%20a%20field%20is%20enabled%20for%20faceting%20in%20the%20schema%2C%20it%20can%20be%20used%20in%20the%20facet_by%20search%20parameter..) are used as parameters in search bar, we currently do not support that.

Fixes #11799 

I've tested this on my own free-tier Typesense cloud deployment and it worked, but please note that this is not guaranteed to fix `OUT_OF_MEMORY` issues, we may eventually need to increase memory :smile: 

### Pull-Request Checklist

-   [ ] There are new or updated unit tests validating the change N\A
-   [ ] Documentation has been updated to reflect this change N\A

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -12,7 +12,7 @@ services:
       MYSQL_DATABASE: "test"
 
   mysql-9:
-    image: "mysql:9.4.0"
+    image: "mysql:9.5.0"
     container_name: "typeorm-mysql-9"
     ports:
       - "3306:3306"
@@ -24,7 +24,7 @@ services:
 
   # mariadb
   mariadb-10:
-    image: "mariadb:10.6.22-jammy"
+    image: "mariadb:10.6.24-jammy"
     container_name: "typeorm-mariadb-10"
     ports:
       - "3307:3306"
@@ -35,7 +35,7 @@ services:
       MYSQL_DATABASE: "test"
 
   mariadb-12:
-    image: "mariadb:12.0.1-rc"
+    image: "mariadb:12.1.2"
     container_name: "typeorm-mariadb-12"
     ports:
       - "3307:3306"
--- a/docs/docs/drivers/microsoft-sqlserver.md
+++ b/docs/docs/drivers/microsoft-sqlserver.md
@@ -207,7 +207,7 @@ const queryEmbedding = [
 const results = await dataSource.query(
     `
     DECLARE @question AS VECTOR (1998) = @0;
-    SELECT TOP (10) dc.*, 
+    SELECT TOP (10) dc.*,
            VECTOR_DISTANCE('cosine', @question, embedding) AS distance
     FROM document_chunk dc
     ORDER BY VECTOR_DISTANCE('cosine', @question, embedding)
--- a/docs/docs/drivers/mysql.md
+++ b/docs/docs/drivers/mysql.md
@@ -139,3 +139,7 @@ export class User {
     roles: UserRoleType[]
 }
 ```
+
+### Vector Types
+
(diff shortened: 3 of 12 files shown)


alumni (commented):
LGTM, but I'm not familiar with Typesense and our account setup. @dlhck and @naorpeled would be the best candidates to review :)

naorpeled (approved):

alumni:
Nice @G0maa, seems the indexing is now running without failing 👏🏻 

Does anyone have any idea how the docs are published? Seems that on typeorm.io the docs stopped being published sometime after v0.3.27. Is that right?

Also: is Typesense indexing our markdown files or the website?

Maybe in the next meeting we can discuss docs too - we should talk about publishing + indexing for 2-3 versions (0.3, master and next).

G0maa:
@alumni 

> Nice @G0maa, seems the indexing is now running without failing 👏🏻

For now :see_no_evil:

> Does anyone have any idea how the docs are published? Seems that on typeorm.io the docs stopped being published sometime after v0.3.27. Is that right?

Sadly no idea.

> Also: is Typesense indexing our markdown files or the website?

Based on [these lines](https://github.com/G0maa/typeorm/blob/eecce9414ec6728220f31d0daa227adf3740b56c/docs/docsearch-scraper-config.json#L3-L8) it's indexing the website. I think I saw it somewhere that Typesense supports indexing docusaurus directly (but will have to verify).

> Maybe in the next meeting we can discuss docs too - we should talk about publishing + indexing for 2-3 versions (0.3, master and next).

Totally agree.

gioboa (commented):
typesense-docsearch-scraper is the command for indexing the website. we have it in the workflow.
In the latest run there is the [out of memory error](https://github.com/typeorm/typeorm/actions/runs/19766007877/job/56638866343#step:4:346) as well.

G0maa:
> In the latest run there is the [out of memory error](https://github.com/typeorm/typeorm/actions/runs/19766007877/job/56638866343#step:4:346) as well

That's not good :sweat_smile:

alumni:
@gioboa I noticed that, but it's in the part where it displays the recent logs (so I though that log might be from the previous runs, but I might be wrong).

gioboa:
> @gioboa I noticed that, but it's in the part where it displays the recent logs (so I though that log might be from the previous runs, but I might be wrong).

I run the task again, same error. :(

G0maa:
Hmmmm... I'll try to figure out a workaround for this.

I believe we shouldn't need to increase memory, TypeORM docs aren't that massive.
Although I'm unaware what increases memory usage, collection size is a factor, but what about e.g. searches happening while writing a new collection? :thinking: 

Going through Typesense docs there's a way to limit the `batch_size` [see](https://typesense.org/docs/29.0/api/documents.html#configure-batch-size). So I might modify that next, say 10 instead of 40.

alumni:
@G0maa that might actually be a good idea, since we go OOM, not out of disk space. Some of the pages are long (but still, they should not take MBs).

PR #11809, opened 2025-11-28
chore: build using tsgo

gioboa:


gioboa (commented):
We need to wait, blocked by https://github.com/microsoft/typescript-go/issues/914

PR #11821, opened 2025-12-04
fix: check constraint migration detection

### LINKED ISSUE #11714: @Check modifications not detected during migrations generate
### Issue description

On a column with a simple check contraint, modification aren't detected during the migration generate.

### Expected Behavior

 ```typescript
@Column({ name: 'test', type: 'text', enum: ['TEST1','TEST2','TEST3'] })
 @Check('test_type_chk', "test IN ('TEST1','TEST2','TEST3')")
  test: TestType;
```

When we change the check constraint in the entity, it should be changed in the generated migration too.

### Actual Behavior

 ```typescript
@Column({ name: 'test', type: 'text', enum: ['TEST1','TEST2','TEST3'] })
 @Check('test_type_chk', "test IN ('TEST1','TEST2','TEST3')")
  test: TestType;
```

In a situation like above, when we modify the check in the entity, it's not updated in the generated migration.

### Steps to reproduce

On a simple table with a check constraint, change the constraint and generate a migration to see if it was ignored.

### My Environment

Nest: 9.2.1,
TypeOrm: 0.3.20

### Additional Context

_No response_

- [ ] aurora-mysql
- [ ] aurora-postgres
- [ ] better-sqlite3
- [ ] cockroachdb
- [ ] cordova
- [ ] expo
- [ ] mongodb
- [ ] mysql
- [ ] nativescript
- [ ] oracle
- [x] postgres
- [ ] react-native
- [ ] sap
- [ ] spanner
- [ ] sqlite
- [ ] sqlite-abstract
- [ ] sqljs
- [ ] sqlserver

Cprakhar:
## Description of change

**Issue**: TypeORM's schema synchronization does not detect changes to `@Check` constraint expressions. When a CHECK constraint definition is modified in an entity, the migration generator ignores the change, causing database schema drift.

**What this change does**:
1. Adds `recreateModifiedChecks()` method to `RdbmsSchemaBuilder` that compares existing database CHECK constraints with entity metadata
2. Implements `normalizeCheckExpression()` to handle database-specific formatting differences:
   - PostgreSQL/CockroachDB: Converts ARRAY syntax (`col = ANY (ARRAY['val1'::text, 'val2'::text])`) to IN syntax (`col IN ('val1', 'val2')`)
   - All databases: Normalizes whitespace, removes type casts, and performs case-insensitive comparison
3. Drops CHECK constraints with modified expressions and recreates them with new values
4. Integrates into the schema sync execution flow between `createNewIndices()` and `createNewChecks()`

**Why this is needed**:
- Entity definitions and database constraints can drift silently
- Developers had to manually write migrations for CHECK constraint changes
- Migration generation now properly tracks these changes

**How it's verified**:
- Existing test `"should correctly change check"` in `test/functional/schema-builder/change-check-constraint.test.ts` validates the fix
- The reproduction case with PostgreSQL demonstrates migration generation now works correctly for modified CHECK constraints

**Database support**:
- PostgreSQL/CockroachDB: Complex ARRAY syntax normalization
- SQLite, SQL Server, Oracle, SAP, Spanner, MySQL, Aurora: Whitespace normalization
- MySQL/Aurora MySQL: Skipped (no CHECK constraint support in older versions)

## Pull-Request Checklist

- [x] This pull request links relevant issues as `Fixes #11714`
- [x] There are new or updated tests validating the change (`tests/functional/schema-builder/change-check-constraint.test.ts`)

### Files Modified
- `src/schema-builder/RdbmsSchemaBuilder.ts`
  - Added `recreateModifiedChecks()` method
  - Added `normalizeCheckExpression()` method
  - Integrated `recreateModifiedChecks()` into execution flow

## Testing

The fix has been validated with:

1. **Existing Test Suite**
   - `test/functional/schema-builder/change-check-constraint.test.ts` - "should correctly change check" test passes

2. **Manual Reproduction** (#11714)
   - Created entity with CHECK constraint: `subject IN ('Math', 'Science', 'History')`
   - Modified constraint to: `subject IN ('Math', 'Science', 'History', 'English')`
   - Migration generation correctly detects the change and generates:
     ```sql
     ALTER TABLE "student" DROP CONSTRAINT "subject_check";
     ALTER TABLE "student" ADD CONSTRAINT "subject_check" 
       CHECK (subject IN ('Math', 'Science', 'History', 'English'));
     ```

3. **Database Compatibility**
   - PostgreSQL: ARRAY syntax normalization tested
   - SQLite/SQL Server/Oracle/MySQL: Whitespace normalization
   - Aurora MySQL/Postgres: Inherit base driver behavior

## Notes

- The fix handles PostgreSQL's automatic conversion of `IN` constraints to `ARRAY` syntax by normalizing both formats for comparison
- Down migrations preserve the database's stored format (with type casts) for accurate rollback
- Expression comparison is case-insensitive to handle SQL dialect differences
- MySQL/Aurora MySQL skip this check as they don't support CHECK constraints in older versions


gioboa (approved):
The new code looks clear and test are passing 💪 thank you @Cprakhar

OSA413:
Needs test, but I'm not sure how to implement tests for migration generation

src/schema-builder/RdbmsSchemaBuilder.ts
```diff
@@ -1067,6 +1068,140 @@ export class RdbmsSchemaBuilder implements SchemaBuilder {
...
+                .sort()
+            const normalizedIn = `${colName} in ('${values.join("','")}')`
+            normalized = normalized.replace(inExpr, normalizedIn)
+        }
+
+        // Normalize escape characters
+        normalized = normalized.replace(/\\'/g, "''")
+
+        return normalized
+    }
```
Cprakhar, line 1184: The normalization of check constraint expressions is quite complex. Each database normalizes the check constraints differently, resulting in many variations. 

The tests I have written locally are currently failing.

gioboa: Thanks @Cprakhar I see, without tests we can't verify if the implementation is good or not.
Can you add a couple of tests in the PR please?

Cprakhar:
@gioboa @OSA413
To address the issue, I propose the following solution:

1. Introduce an `originalExpression` field within the `TableCheck` class, as well as in the related classes and interfaces, including `CheckMetadata`. This field will store the check expression defined at the time of entity definition.
2. Save the `originalExpression` in the `typeorm_metadata` table and retrieve it during the `loadTables` process in the driver's `queryRunner`.
3. Use this information to accurately compare check constraints in the `RdbmsSchemaBuilder`, ensuring the effective reconstruction of modified check constraints.

While this approach may create additional queries and introduce some overhead, I believe it is necessary because databases store check constraint expressions differently. For instance, some databases may add parentheses or type casts, such as::array (PostgreSQL).

This solution will only compare the `originalExpression` of old and new checks.

alumni:
> Needs test, but I'm not sure how to implement tests for migration generation

We should have some examples already for that (search for `createSchemaBuilder` within `*.test.ts` files) - the nicest way would be to create a migration file for the initial entities, setup a `DataSource` with `sync: false`. Then in the test:

```ts
await dataSource.runMigrations()
const sqlInMemory = await dataSource.driver.createSchemaBuilder().log()

expect(sqlInMemory.upQueries.map({ query } => query)).to.deep.equal([...])
expect(sqlInMemory.downQueries.map({ query } => query)).to.deep.equal([...])
```

gioboa (changes requested):
Can you add tests for this change please?

Cprakhar:
> Can you add tests for this change please?

I created tests (currently for `postgres` driver). The `normalizeCheckExpression` method in `RdbmsSchemaBuilder` is not robust. The added test will fail because the expression is not properly normalized for comparison.
Example:
- `(value > 0) and (value < 100)` and `value > 0 and value < 100` (both expressions after normalization) will recreate up and down queries.

I don't have much expertise in regex 🥲;

What do you have to say about my above proposed solution? - https://github.com/typeorm/typeorm/pull/11821#issuecomment-3795288369

gioboa:
Thanks for your commitment, I will look at it in few hours.

gioboa (commented):
I dropped few suggestions here [27b7116](https://github.com/typeorm/typeorm/pull/11821/commits/27b7116c7de9732bd52f473102a825073c01d940). feel free to change/remove that if you want. 👍

Cprakhar:
@gioboa did you give any thoughts to https://github.com/typeorm/typeorm/pull/11821#issuecomment-3795288369

gioboa:
> While this approach may create additional queries and introduce some overhead, I believe it is necessary because databases store check constraint expressions differently. For instance, some databases may add parentheses or type casts, such as::array (PostgreSQL).

It sounds reasonable to me, thanks for your thoughts @Cprakhar

Cprakhar:
> > While this approach may create additional queries and introduce some overhead, I believe it is necessary because databases store check constraint expressions differently. For instance, some databases may add parentheses or type casts, such as::array (PostgreSQL).
> 
> It sounds reasonable to me, thanks for your thoughts @Cprakhar

I had locally implemented the above approach for (postgres/cockroach), and it was working fine.
Need other maintainers' approval first before coming up with the fix.

This approach could be used for #1729 as well for all drivers.

pkuczynski:
Hey @Cprakhar, any reason why you closed this one? Just out of the interest... I am working now on merging everything open, so I would get to this one in the nearest future...

Cprakhar:
> Hey @Cprakhar, any reason why you closed this one? Just out of the interest... I am working now on merging everything open, so I would get to this one in the nearest future...

I approached this issue with work #12130

PR #11824, opened 2025-12-05
fix: make embedded columns nullable in single table inheritance

### LINKED ISSUE #11703: Embedded columns are always nullable if the embedding is used in parent TableInheritance class
### Issue description

When using an embedded entity with non-nullable columns in the parent table inheritance class, the columns end up nullable in the data base.

### Expected Behavior

If the embedded column is not nullable, it should not be nullable in the table inheritance table because it is shared across all child entities.

### Actual Behavior

The column ends up being nullable in the data base.

### Steps to reproduce

```ts
export class TestEmbedding {
  @Column({ nullable: false })
  test: string;
}

export enum TestType {
  A = 'A',
  B = 'B',
}

@Entity()
@TableInheritance({ column: { type: 'enum', enum: TestType, name: 'type' } })
export class Parent {
  @PrimaryGeneratedColumn()
  id: number;

  @Column(() => TestEmbedding, { prefix: false })
  embedding: TestEmbedding;
}

@ChildEntity(TestType.A)
export class ChildA extends Parent {
  type: TestType.A = TestType.A;

  @Column()
  a: string;
}

@ChildEntity(TestType.B)
export class ChildB extends Parent {
  type: TestType.B = TestType.B;

  @Column()
  b: string;
}
```

Produces this migration:
```js
await queryRunner.query(`CREATE TYPE "public"."parent_type_enum" AS ENUM('A', 'B')`);
await queryRunner.query(`CREATE TABLE "parent" ("id" SERIAL NOT NULL, "a" character varying, "b" character varying, "type" "public"."parent_type_enum" NOT NULL, "test" character varying, CONSTRAINT "PK_bf93c41ee1ae1649869ebd05617" PRIMARY KEY ("id"))`);
```

Notice the `test` column does not get `NOT NULL` even though it is present all types of children.

### My Environment

| Operating System    | MacOS 15 |
| Node.js version     | 22.18.0  |
| TypeORM version     | 0.3.25 |

### Additional Context

I only tried this with Postgres, but I assume the issue lies in the meta data generation for embedded columns in table inheritances.

- [ ] aurora-mysql
- [ ] aurora-postgres
- [ ] better-sqlite3
- [ ] cockroachdb
- [ ] cordova
- [ ] expo
- [ ] mongodb
- [ ] mysql
- [ ] nativescript
- [ ] oracle
- [x] postgres
- [ ] react-native
- [ ] sap
- [ ] spanner
- [ ] sqlite
- [ ] sqlite-abstract
- [ ] sqljs
- [ ] sqlserver

Cprakhar:
### Description of change

**Fixes #11703**

This PR fixes a bug where embedded columns with `nullable: false` in parent entities were incorrectly forced to be nullable when using Single-Table Inheritance (STI).

#### Problem
When using the `STI (Single Table Inheritence)` pattern, TypeORM was unconditionally marking **all** embedded columns as nullable, regardless of whether the embedded was defined in the parent entity or in child entities. This behavior was incorrect for parent-defined embeds because:
- Parent-defined embeds are inherited by ALL child entities
- Since all children must have these columns, they should respect their original `nullable` constraint
- Only child-specific embeds should be forced to nullable (since parent entities won't have values for them)

#### Example
```typescript
export class Counters {
  @Column({ nullable: false })
  likes: number;
}

@Entity()
@TableInheritance({ column: { type: 'enum', name: 'type' } })
export class Content {
  @Column(() => Counters)
  counters: Counters;  // Defined in parent
}

@ChildEntity()
export class Photo extends Content {
  @Column()
  size: string;  // Child-specific
}
```

**Before this fix:** The `likes` column would be nullable even though it's defined with `nullable: false` in the parent.

**After this fix:** The `likes` column correctly has `NOT NULL` constraint because it's in the parent (inherited by all children).

#### Solution
Modified `EntityMetadataBuilder.computeEntityMetadataStep1()` to check if an embedded is defined in a child entity (similar to how regular columns are handled). The fix:
1. Stores the `embeddedArgs` array before creating embedded metadatas
2. Checks each embedded's original target against child entity targets
3. Only forces `nullable: true` for embeds defined in child entities
4. Parent-defined embeds retain their original nullable settings

#### How verified
1. Created comprehensive functional tests in `test/functional/table-inheritance/single-table/embedded-columns-nullable/`
2. Tests verify:
   - Parent-defined embedded columns maintain `NOT NULL` constraint
   - Child-defined embedded columns are forced to nullable
   - Data can be correctly saved and loaded
3. Verified the fix against the original issue reproduction case
4. All existing tests continue to pass

#### Changes made
- **Modified:** `src/metadata-builder/EntityMetadataBuilder.ts` (lines 524-549)
  - Changed unconditional nullable forcing to conditional based on embed's target entity
  - Mirrors the logic already used for regular columns (lines 570-577)
- **Added:** New functional test suite for embedded columns in STI
  - `test/functional/table-inheritance/single-table/embedded-columns-nullable/`

### Pull-Request Checklist

- [x] This pull request links relevant issues as `Fixes #11703`


gioboa (commented):
>   npm i https://pkg.pr.new/typeorm/typeorm@11824

@jonaskordt is the preview package from this PR fixing your issue? Can you double check it on your scenario?

michaelbromley:

```
git fetch origin
```

PR #11825, opened 2025-12-05
test: add Vitest

OSA413:
I'm opening this as PR because I need to test it in the containers and there are issues that I don't know now how to fix (needs discussion or collective debugging).

Particularly there's an issue with excess `undefined` fields in objects that fail tests that use `should.have.deep.members` (`be.deepEqualIgnoreUndefined` is used to replace `be.eql` when it fails but I don't know how to fix `members`).

Also I had to import members from vitest because otherwise it didn't work.

### DIFF
--- a/src/util/ImportUtils.ts
+++ b/src/util/ImportUtils.ts
@@ -22,12 +22,13 @@ export async function importOrRequireFile(
                 importPath
             )
         } catch {
-            try {
-                // This one should work with modern environments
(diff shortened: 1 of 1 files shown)


OSA413:
How do I switch branch to `next` safely?

btw it's 'veetest' since Vite is pronounced as 'veet'

OSA413:
I think that's all I can do currently...

PR #11827, opened 2025-12-06
ci: detect if package-lock.json is up to date and generated with correct npm version

pkuczynski:
### Description of change

New CI step detecting  if `package-lock.json`:
* is up to date
* was generated using npm shipped with node defined in `.nvmrc`

### Pull-Request Checklist

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -67,6 +67,7 @@ jobs:
         with:
           node-version-file: .nvmrc
       - run: npm install
+      - run: git status
       - run: |
           if [ -n "$(git status --porcelain)" ]; then
             echo "package-lock.json is not up to date or was generated with a different npm version than the one shipped with Node.js $(cat .nvmrc)"


alumni:
~Hey, maybe it's not urgent to do this now - one quick thing would be `--ignore-engines` in CI or we could also leave it as it is for now.~

I normally use the built-in `corepack` to pick the package manager version, but it requires some extra setup for `npm` AFAIR.

Maybe `onFail` could be changed to something else, like "ignore" or "download" - I think newer versions of `npm` will automatically download and switch to the older ones when they see `"onFail": "download"`.

In the long term we need a fixed version (down to the patch) of a package manager - that is quite easy with e.g. `pnpm`, but it would mean to change things we don't want to change right now - at least not with the current branching model.

package.json
```diff
@@ -251,7 +251,7 @@
     },
     "packageManager": {
       "name": "npm",
-      "version": ">=10.1.0",
+      "version": "10",
       "onFail": "error"
```
alumni:
```suggestion
      "onFail": "warn"
```

`"download"` is not yet supported by `npm`

pkuczynski: This would not prevent messing up with lockfile. If we do `warn` we might as well remove the whole `devEngines` section...

What we could to alternativelly, is have a dedicated step validating if lockfile has been created using npm@10...

The difference between npm 10 and 11 are rather cosmetic. They both use the same `lockfileVersion: 3`, so they should be compatible. But I understand you want to avoid unnecessary diffs on packakge-lock, which make sense...

OSA413:
The tests uncover an interesting case: the default node 24's `npm` is 11, so what should be done then? I think we can manually regenerate `package-lock.json` for npm@10 before merging.

Btw how does npm@10 handle `package-lock.json` created by npm@11?

pkuczynski:
> Btw how does npm@10 handle `package-lock.json` created by npm@11?

As this is the same `lockfileVersion: 3`, it does not care: see example run https://github.com/typeorm/typeorm/actions/runs/20035394227/job/57455392520?pr=11835#step:6:22

But if we allow different lockfile versions, we might end-up with some unnecessary diffs... Not a major issue I would say...

pkuczynski:
This whole discussion led me to discover our current `package-lock.json` is not up to date with `package.json` which most likely happened during one of the merges `master` > `next`. 

So I added new CI step, which would detect if package-lock:
* is up to date
* was generated using npm shipped with node defined in `.nvmrc`

Here are some tests:
1. `package-lock` outdated: https://github.com/typeorm/typeorm/actions/runs/20035851700?pr=11827 - `FAIL`
2. `package-lock` up to date: https://github.com/typeorm/typeorm/actions/runs/20036015952?pr=11827 - `PASS`
3. `package-lock` generated using npm 11.6.0 - https://github.com/typeorm/typeorm/actions/runs/20036071615?pr=11827 - `PASS`
    * this is interesting! npm@11 [removed](https://github.com/typeorm/typeorm/pull/11827/commits/fdb4fe5a1f0aed0ef38d2cdf6fda3c5c602e8d82) one more deps but npm@10 does not seem to [care](https://github.com/typeorm/typeorm/actions/runs/20036164632/job/57458188584#step:5:6). Must be some npm glitch as when I move back to `npm@10` it does not restore this dependency.

package.json
```diff
@@ -244,18 +244,6 @@
     "url": "https://opencollective.com/typeorm",
     "logo": "https://opencollective.com/opencollective/logo.txt"
   },
-  "devEngines": {
```
gioboa, line 247: Do we need to remove these entries?

pkuczynski: No, we do not have to remove them, but they are redundant to `.nvmrc`, which we recommend as a tool defininig the node version we should use for local development. It's better to define.

As the `npm` documentation [specify](https://docs.npmjs.com/cli/v11/configuring-npm/package-json#devengines):

> The `devEngines` field aids engineers working on a codebase to all be using the same tooling.

So in our case `.nvmrc` provides the same functionality.

alumni:
This only makes things more complicated 🤷🏻

anyway, we'll fix it sometime

alumni:
> This only makes things more complicated 🤷🏻

Hmm, I was expecting `npm ci` in the build step to fail, but seems it does not?

> If dependencies in the package lock do not match those in `package.json`, `npm ci` will exit with an error, instead of updating the package lock.

Source: https://docs.npmjs.com/cli/v11/commands/npm-ci

Maybe because they say "dependencies" - both `pnpm` and `yarn` are more explicit about `install --frozen-lockfile`:

> pnpm doesn't generate a lockfile and fails to install if the lockfile is out of sync with the manifest / an update is needed or no lockfile is present.

Source: https://pnpm.io/cli/install#--frozen-lockfile

> If the `--immutable` option is set (defaults to true on CI), Yarn will abort with an error exit code if the lockfile was to be modified. For backward compatibility we offer an alias under the name of `--frozen-lockfile`, but it will be removed in a later release.

Source: https://yarnpkg.com/cli/install#details

So they aren't identical as I thought.

PR #11836, opened 2025-12-08
feat(sqlite)!: drop support for sqlite3 and default to better-sqlite3

### LINKED ISSUE #11718: node-sqlite3 maintenance concerns and path forward
I wanted to bring attention to the pull request opened yesterday proposing to mark the node-sqlite3 library as unmaintained. While it hasn't been merged yet, this means TypeORM's current SQLite dependency will no longer be maintained.

https://github.com/TryGhost/node-sqlite3/pull/1844

The timing makes sense when you consider that Node.js v22.5.0 introduced a built-in sqlite module. The built-in module eliminates the need for native addon compilation and its associated toolchain requirements.

Migrating from node-sqlite3 to the built-in module would bring several advantages. The built-in module simplifies the dependency story—it's just there as part of Node.js itself. There's no need for prebuilt binaries, fallback compilation, or the tooling that supports that infrastructure. The module would be maintained by the Node.js core team alongside the rest of the runtime, ensuring better long-term support and security updates. It would also reduce the dependency tree, always a plus for maintenance and security.

Of course, there are some considerations to think through. The built-in module is currently marked as experimental and requires Node.js v22.5.0 or higher. This means we'd need to think carefully about backwards compatibility and potentially maintain a transition period where both drivers are supported.

I think it's worth starting the conversation about how TypeORM might approach a transition away from node-sqlite3 in general. The node-sqlite3 library has served the community well for years, and with built-in Node.js support now available, it seems like a good time to explore the options. I'm curious to hear thoughts from the maintainers and community on timing and implementation approach.

pkuczynski:
Fixes #11718

### DIFF
--- a/test/github-issues/799/issue-799.test.ts
+++ b/test/github-issues/799/issue-799.test.ts
@@ -19,17 +19,6 @@ describe("github issues > #799 sqlite: 'database' path should be created", () =>
         }
     })
 
-    it("should create the whole path to database file", async function () {
-        dataSource = new DataSource({
-            name: "better-sqlite3",
-            type: "better-sqlite3",
-            database: path,
-        })
-        await dataSource.initialize()
-
-        expect(dataSource.isInitialized).to.equal(true)
-    })
-
     it("should create the whole path to database file for better-sqlite3", async function () {
         // run test only if better-sqlite3 is enabled in ormconfig
         const isEnabled = getTypeOrmConfig().some(


gioboa (approved):
That's awesome 👏

docs/docs/guides/8-migration-v1.md
alumni, line 1: Could you please move this back to `guides/8-migration-v1.md`? thanks! :)

src/commands/InitCommand.ts
alumni, line 1: Could you please undo the changes that are not related to sqlite? thanks!

src/driver/DriverUtils.ts
```diff
@@ -12,22 +12,23 @@ export class DriverUtils {
 
     /**
      * Returns true if given driver is SQLite-based driver.
+     * @param driver
      */
     static isSQLiteFamily(driver: Driver): boolean {
         return [
-            "sqlite",
+            "better-sqlite3",
```
alumni, line 19: Always nice to see things are alphabetically sorted when their order does not matter. 👍🏻

Easier to read + I've worked in projects where actually it helped solving merge conflicts.

pnpm-lock.yaml
alumni, line 1: This file has changes unrelated to the `sqlite3` removal. Could you please revert them?
```shell
git checkout upstream/master -- pnpm-lock.yaml
pnpm install
```

test/github-issues/799/issue-799.test.ts
```diff
@@ -20,15 +20,9 @@ describe("github issues > #799 sqlite: 'database' path should be created", () =>
     })
 
     it("should create the whole path to database file", async function () {
```
alumni: Looking at the entire file, seems this entire test can be removed, since we test for `better-sqlite3` below.

Only noticed this because the removed early return was necessary.

test/unit/connection-options-reader/connection-options-reader.test.ts
```diff
@@ -37,19 +37,6 @@ describe("ConnectionOptionsReader", () => {
         expect(entities.length).to.equal(1)
     })
 
-    it("properly loads sqlite in-memory/path config", async () => {
-        const connectionOptionsReader = new ConnectionOptionsReader({
-            root: __dirname,
-            configName: "configs/sqlite-memory",
```
alumni, line 43: Seems `configs/sqlite-memory.ts` has `better-sqlite3`, so this test could still be adjusted to use the `memory2` config?

Or should the config file be removed together with the test?

eslint.config.mjs
alumni, line 1: Please undo changes unrelated to `sqlite3` :)

test/functional/driver/better-sqlite3/busy-timeout.test.ts
alumni, line 1: `‎test/functional/sqlite` could be merged into `test/functional/driver/better-sqlite3` now.

Seems some of the driver tests (`sqlite3` and `sqljs`, but not `mongodb`) weren't placed in the `driver` folder.

src/commands/InitCommand.ts
pkuczynski: They are all related...

pnpm-lock.yaml
pkuczynski: Refreshed the lockfile, but it should not havet unrelated changes...

test/unit/connection-options-reader/connection-options-reader.test.ts
pkuczynski: `configs/sqlite-memory.ts` was unused so I removed it. And this test failed with better-sqlite, hence removed earlier

eslint.config.mjs
pkuczynski: Come from `next` branch, which this PR was originally based on, according to your request. Not my fault that you first asked everybody to work in `next` and then decided to cherry pick only some changes from this branch... These wont hurt, we can ignore this I think...

test/functional/driver/better-sqlite3/busy-timeout.test.ts
pkuczynski: Good point. Done. Actually sqljs should also be moved. I will do it in a separate PR.

pkuczynski: This is the follow up PR: https://github.com/typeorm/typeorm/pull/12036

src/commands/InitCommand.ts
```diff
@@ -725,10 +759,6 @@ Steps to run this project:
                 packageJson.dependencies["pg"] =
                     ourPackageJson.devDependencies.pg
                 break
-            case "sqlite":
-                packageJson.dependencies["sqlite3"] =
-                    ourPackageJson.devDependencies.sqlite3
```
alumni, line 730: Funny thing, loading `package.json` did not work on certain configurations (e.g. PnP, default in `yarn`) + now we also don't have `devDependencies` because they get removed during the build. We'll figure out something, but not right now (or it would be just for a few weeks).

pnpm-lock.yaml
alumni: Thanks, it's looking much better now.

test/github-issues/799/issue-799.test.ts
alumni: @pkuczynski Can we also remove this duplicate?

pkuczynski: You are right! Done

michaelbromley:
Great work @pkuczynski 🙌

PR #11840, opened 2025-12-09
ci: ensure pull request follow conventional commits spec

pkuczynski:
Ensures that pull request titles match the [Conventional Commits spec](https://www.conventionalcommits.org/)


.github/workflows/pull-request.yml
```diff
@@ -0,0 +1,35 @@
...
+            feat
+            fix
+            lint
+            refactor
+            revert
+            test
+          subjectPattern: ^[a-z][a-zA-Z0-9 \-\/()*.@,_`']+$
+          subjectPatternError: |
+            The subject "{subject}" found in the pull request title "{title}"
+            didn't match the configured pattern. Please ensure that the subject
```
gioboa: Let's drop a link to see the subjects to use, or list them here. WDYT?

gioboa (commented):
It looks great to me 👍

.github/workflows/pull-request.yml
pkuczynski: Subject is free text (in the spec is called `description`. I guess you meant `type` but there is not option to specify this in the action. The format is:

`<type>[optional scope]: <description>`

Check the spec: https://www.conventionalcommits.org/en/v1.0.0/#summary. I will adjust the text of the error message to follow more of the original spec...

gioboa: Thanks 👍 great

gioboa (approved):
LGTM 👏

mguida22:
This seems a bit heavy handed to me. I would rather encourage contributions and leave it to maintainers to edit PR titles as needed (without a failing check)

naorpeled:
> This seems a bit heavy handed to me. I would rather encourage contributions and leave it to maintainers to edit PR titles as needed (without a failing check)

+1

pkuczynski:
> This seems a bit heavy handed to me. I would rather encourage contributions and leave it to maintainers to edit PR titles as needed (without a failing check)

It's linting like any other. Just this time it's linting PR and not the code. Seems very reasonable thing for me to do. I took this one from [`vite`](https://github.com/vitejs/vite/blob/main/.github/workflows/semantic-pull-request.yml) and [`faker`](https://github.com/faker-js/faker/blob/next/.github/workflows/semantic-pull-request.yml) actually, but I have seen this on many other repos too...

michaelbromley:
I just checked the [contributing guide](https://github.com/typeorm/typeorm/blob/master/CONTRIBUTING.md#-submitting-a-pull-request-pr) and I couldn't clearly see an instruction on the PR name convention in particular. There is a detailed section on commit message format, but it's not necessarily obvious that the PR name must follow that. So we should also make the expectation explicit there, with an example.

I'm personally OK with this change - with the amount of external contributions we get, it will at least keep things consistent.

When the workflow fails, does it also add a comment to the PR thread with the error?

pkuczynski:
No, it does not add a comment, but PR owner would get an email with failed build notification. Maintainers will see it too. And we should add this workflow to pass as part of branch protection rules.

Let me update the contribution guidelines to stay consistent.

pkuczynski:
I double checked and other popular repos do not add comment. Possibly to avoid the noise. 

This action documentation does show example how to add a comment when it fails (and remove when fixed), but I am not sure if this is worth the effort. I was 50/50 on this, and eventually decided not to do it. Let me know if you think otherwise...

G0maa:
I’m neutral on this change; we could always test it and see if it makes people’s lives harder, then remove it :grinning:

alumni (approved):

PR #11842, opened 2025-12-10
ci: use prettier on all types of files it supports

pkuczynski:
Prettier support many files, and even more with plugins. There is no reason why we should not laverage the full power of it.

How the new setup would affect current code can be seen in #11843 (which should be merged first)

### DIFF
--- a/.prettierignore
+++ b/.prettierignore
@@ -1,3 +1,5 @@
+build/
+coverage/
 extra/
 src/driver/mongodb/typings.ts
 src/driver/mongodb/bson.typings.ts


.lintstagedrc.json
```diff
@@ -1,5 +1,4 @@
 {
-  "*.ts": ["eslint", "prettier --write"],
-  "*/**/*.json": ["prettier --write"],
-  "./!(package-lock).json": ["sort-package-json", "prettier --write"]
```
alumni, line 4: This was formatting all our config files (e.g. `tsconfig.json`, `.mocharc.json` etc), not just the `package.json`. Would be great not to lose this feature.

I think this might still work with `prettier-plugin-packagejson`, could you confirm?

I think it might run `sort-package-json` on every json file, not just the ones in the root, which hopefully is not a problem - I don't remember if we had any text fixtures that shouldn't be sorted.

package.json
```diff
@@ -91,7 +91,6 @@
     "test": "npm run compile && npm run test:fast --",
     "test:ci": "mocha --bail",
     "test:fast": "mocha",
-    "typecheck": "tsc --noEmit",
```
alumni, line 92: We need this sometimes to check if we fixed the type errors, please don't remove it. 

Not sure about the docs (less critical) - probably there `--noEmit` is missing?

.prettierrc.js
alumni, line 1: If you need autocomplete in JSON, you can also add:

```json
{
  "$schema": "https://www.schemastore.org/prettierrc.json",
  "semi": false
}
```

I would not change the file format yet, since we will try first to use every default setting from NX if possible. We should have as little customizations as possible, especially since it's an open source project (I learned this the hard way).

One more thing: two of the options are now default in `prettier@3`, so they don't need to be there anymore. Already in `master`.

package.json
```diff
@@ -156,12 +155,13 @@
     "pg": "^8.16.3",
     "pg-query-stream": "^4.10.3",
     "pkg-pr-new": "^0.0.60",
-    "prettier": "^2.8.8",
+    "prettier": "^3.7.4",
+    "prettier-plugin-packagejson": "^2.5.20",
+    "prettier-plugin-toml": "^2.0.6",
```
alumni, line 158: Do we need an entire package for a single `toml` file which is actually temporary (let's see how long Qodo will last) and rarely updated?

I would not add it for one file.

.editorconfig
```diff
@@ -12,7 +8,6 @@ charset = utf-8
 indent_style = space
 indent_size = 4
```
alumni, line 9: Since we are changing this file - there's something I wanted to do for quite some time, but didn't want to it until the we made more progress on 1.0: markdown lists are formatted with a lot of space because we didn't set up `indent_size = 2` for `*.md` files.

For sure we want 4 spaces for `*.{c,m,}{js,ts}{x,}` (all TS/JS extensions - react too in docs), but for all the others, I think we might want 2 - basically for `*.{json,md,yml}` (we don't have `yaml` anymore AFAIK).

.lintstagedrc.json
pkuczynski: Yes, it works with `prettier-plugin-packagejson`, just checked

.lintstagedrc.json
```diff
@@ -1,5 +1,4 @@
...
+  "*.ts": ["eslint --fix", "prettier --write"],
```
alumni, line 2: Hey @pkuczynski,

In the past, `--fix` was removed on purpose, because there were fixable rules that were causing a change in behavior, e.g. `no-wrapper-object-types`. I remember later I disabled this rule in particular since we don't want to correct the types, but I don't remember if  there were others. Do you happen remember if there was any lint warning that wasn't related to jsdoc?

pkuczynski: Hmm, `no-wrapper-object-types` seems to be a very solid and recommended rule so I am not sure how this could affect us in a bad way?

I think it makes sense to have --fix here to prevent necessity of manual run after the build fails in CI. So it should be more efficient this way.

Eventually a developer encountering lint error might anyway call it with --fix and encounter unexpected side effect of changes applied also to files he didn't touch...

alumni: I think it's more or less fine now, only saw jsdoc fixable warnings when I ran eslint earlier - hopefully there weren't others that were accidentally fixed and we didn't notice in code review.

pkuczynski: Hmmm, hard to say, but PRs are rather not massive so we should notice. And this only applies to files modified in the commit, not the whole repo

alumni: Yeah, I will checkout a commit and verify when I get the time. It's quite easy to omit e.g. a change from `Object` to `object` in code review, but Qodo should have spotted them.

PR #11843, opened 2025-12-10
ci: use prettier on all types of files it supports

pkuczynski:
This PR demonstrates changes introduced by prettier via #11842


alumni:
@pkuczynski We will remove directly using prettier, I would not bother with this.

pkuczynski:
> @pkuczynski We will remove directly using prettier, I would not bother with this.

What would you replace it with?

alumni:
> What would you replace it with?

Linting, formatting, testing, publishing will all be done via `nx`. It will still be `prettier`, but with the default `nx` config.

pkuczynski:
> Linting, formatting, testing, publishing will all be done via `nx`. It will still be `prettier`, but with the default `nx` config.

Exactly, `nx` is only the executor. It will still use prettier config, we still need lint staged. And as you can see the new setup is already finding quite a lot of things and inconsistencies in our code, so it's worth to run it now. There is no reason why to limit `prettier` just for few file types if it can do better for us...

PR #11844, opened 2025-12-10
ci: upgrade actions to v6

pkuczynski:
Synchronize actions to latest ones (and synchronise some miss-matches in versions). This should also improve caching...


gioboa (approved):
LGTM 🚀  Thanks

PR #11851, opened 2025-12-12
chore: improve coverage reporting and import consistency

mirror-consulting:
Excludes type-only files from c8 coverage reports to provide accurate metrics for executable code. Adds ESLint rule to enforce consistent type imports.

**Changes:**
- Add .c8rc.json excluding 170+ type-only files (interfaces, types, option definitions without implementation)
- Add @typescript-eslint/consistent-type-imports ESLint rule

**Benefits:**
- Coverage metrics now reflect actual executable code (not 0% on interfaces)
- Consistent import type usage improves bundle size and compilation speed
- Script allows future updates to exclusion list as codebase evolves

### Pull-Request Checklist

### DIFF
--- a/docs/docs/advanced-topics/3-indices.md
+++ b/docs/docs/advanced-topics/3-indices.md
@@ -144,6 +144,15 @@ TypeORM supports generating SQL with this option when the concurrent option is s
 
 For more information see the [Postgres documentation](https://www.postgresql.org/docs/current/sql-createindex.html).
 
+## Index Type
+If you need to specify a custom type for the index, you can use the `type` property. If the `spatial` property is set, this field will be ignored.
+
+```typescript
+@Index({ type: 'hash' })
+```
+
+This feature is currently supported only for PostgreSQL.
+
 ## Disabling synchronization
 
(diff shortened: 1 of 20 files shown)


alumni:
Thanks for opening this PR, but I'm afraid we won't merge it.

First of all, I would not put two different things in one PR.

Regarding the ESLint rule - we have very few type imports in the project. The long-term plan is to enforce type imports when only the types are needed: convert all regular imports to type imports if possible. I know decorators will prevent some imports to be auto-fixed, but that will be only in some tests. We would like to do that closer to the 1.0 release, since it's quite disruptive and we don't want to have conflicts when there's a lot of work ongoing.

Regarding coverage - here I personally don't care about the number reported, but rather if something is changing relatively to what we have on the master branch. There are a few things on the list that should not be ignored, e.g. `QueryRunner` since they implement some things. Others, especially interfaces like `ValueTransformer` should not be taken into account in coverage - but probably we should rather check how the coverage runs (is it not calculating against the transpiled code? will it be solved when we will convert imports to type imports?). If we add all these exceptions, we might be hiding an another issue in the way we collect coverage.

And one more thing: with such a long list of exceptions, it will be difficult to maintain it each time there's a change. I would avoid setting up coverage exceptions in any project, not just this one.

PR #11852, opened 2025-12-13
feat(docs): add versioning 

naorpeled:
## Summary

This PR adds Docusaurus documentation versioning to enable maintaining multiple documentation versions simultaneously.

## Changes

- **Enabled Docusaurus versioning** in `docs/docusaurus.config.ts`
  - Configured version labels: `current` → "next", `v0.3` → "v0.3 (stable)"
  - Added docs version dropdown to navbar for easy version switching

- **Created v0.3 stable version snapshot**
  - Generated `versioned_docs/version-v0.3/` directory with snapshot of current docs
  - Created `versioned_sidebars/version-v0.3-sidebars.json`
  - Updated `versions.json` with version metadata

- **Updated documentation**
  - Added versioning workflow section to `docs/README.md`
  - Added "Contributing to Documentation" section to `CONTRIBUTING.md`

### Index page with new versions dropdown
image

### Versions dropdown
image

### Getting started page when looking at `v0.3`
image

### Getting started page when looking at `next`
image

The version dropdown in the navbar allows users to switch between "next" (in-development) and "v0.3 (stable)" versions.

## Testing

- ✅ Build completes successfully
- ✅ Versioned routes work correctly (`/docs/getting-started` for next, `/docs/v0.3/getting-started` for stable)
- ✅ Version dropdown displays correctly in navbar
- ✅ Both versions render properly

## Documentation

- Documentation source remains in `docs/docs/` (labeled as "next")
- Stable version snapshot in `versioned_docs/version-v0.3/` (labeled as "v0.3 (stable)")
- See `docs/README.md` for instructions on creating new version snapshots

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -132,6 +132,11 @@ The documentation uses versioning to maintain multiple versions:
 - Stable versions (e.g., **"v0.3 (stable)"**) are snapshots created when releasing a new version
 - Users can switch between versions using the version dropdown in the documentation navbar
 - When contributing, always edit the **"next"** version unless you're specifically updating a stable version snapshot (which should be rare and typically done by maintainers)
+- **Backporting important fixes to stable docs** (for critical corrections that users on the stable version must see):
+    - Make the fix in `docs/docs/` first (so it is present in `next`).
+    - Apply the same minimal fix to the corresponding file in `docs/versioned_docs/version-v0.3/` (or the relevant stable version directory). Keep changes scoped to the exact fix; avoid broad refactors in snapshots.
+    - Ensure both versions stay aligned where applicable, then run a docs build locally to confirm there are no errors.
+    - Mention in the PR description that the change was backported to the stable docs and list which versions you touched.
 
 For more details about the documentation setup, see [`docs/README.md`](docs/README.md).
 


naorpeled:
> @naorpeled Can we figure out a way to deploy from the `v0.3` branch and from the `master` branch? We should not have to copy documentation from one branch to another each time we change something.

It's 100% possible to deploy from both, the reason I think we need this is because after v1 will be released we'll still need to have the v0.3 docs in the same deployment.
We can also freeze the v0.3 docs later and postpone this PR though.

Wdyt?

naorpeled:
> How's indexing handled? (Typesense search)

For now my thought was to not index until we finalize v1, but we can also create different collections if we'd like

alumni:
> It's 100% possible to deploy from both, the reason I think we need this is because after v1 will be released we'll still need to have the v0.3 docs in the same deployment.
We can also freeze the v0.3 docs later and postpone this PR though.

I also took a quick look over some other projects that use Docusaurus and noticed they have dedicated repos for docs setup like in this PR (e.g. pnpm). I was hoping to avoid storing the `v0.3` docs in `master` and deploy all docs independently of each other, e.g. from the `v0.3.28` git tag (or the `v0.3` branch) and from the `master` branch. Would this be possible? If this is possible with 2 deployments but not with 3, that's also fine I think.

PR #11815, opened 2025-12-02
fix(redis): version detection logic

ibrahimmenem:
fix bug with redis version detection logic

cherry-pick changes from https://github.com/typeorm/typeorm/pull/11767 
fixes https://github.com/typeorm/typeorm/issues/11635
fixes https://github.com/typeorm/typeorm/issues/11749
close https://github.com/typeorm/typeorm/pull/11767
close https://github.com/typeorm/typeorm/pull/11764

### Description of change
- instead of using heuristics based on functions signatures we are reading the version from node modules
- added 2 unit tests

### Pull-Request Checklist
- [x] Code is up-to-date with the master branch
- [x] This pull request links relevant issues as Fixes #11635 #11749
- [x] There are new or updated unit tests validating the change
- [ ] Documentation has been updated to reflect this change

PR #11853, opened 2025-12-15
fix(redis): read version from package

### LINKED ISSUE #11846: PlatformTools.readPackageVersion throws an error in NextJs
### Issue description

When redis cache tries to determine version an error is thrown

### Expected Behavior

Data source initializes successfully.

### Actual Behavior

DataSource fails to initialized due to failure in determining redis version.

const version = PlatformTools.readPackageVersion("redis") throws an error `Error during Data Source initialization TypeError: Failed to read package.json for "redis": Cannot find module 'unknown'`

### Steps to reproduce

```
new DataSource({
  // ...
  cache: {
        type: 'redis',
        options: {
            socket: {
                host: 'my-redis-host',
                port: 6379,
            },
        },
    }
});
```

Error occurs when trying to initialize datasource.

### My Environment

| Operating System    |  alpine-linux (docker)        |
| Node.js version     | 22.21.1  |
| NextJs | 16.0.7 |

### Additional Context

I tried adding the same require statement that `PlatformTools.readPackageVersion("redis")` would call, right before initialization, and it returns the version as expected
```
const test = require('redis/package.json');
console.log('redis version', test.version);
await AppDataSource.initialize();
```

But the error still occurs: ` Error during Data Source initialization TypeError: Failed to read package.json for "redis": Cannot find module 'unknown'`

- [ ] aurora-mysql
- [ ] aurora-postgres
- [ ] better-sqlite3
- [ ] cockroachdb
- [ ] cordova
- [ ] expo
- [ ] mongodb
- [ ] mysql
- [ ] nativescript
- [ ] oracle
- [ ] postgres
- [ ] react-native
- [ ] sap
- [ ] spanner
- [ ] sqlite
- [ ] sqlite-abstract
- [ ] sqljs
- [ ] sqlserver

Not at this time.

mguida22:
### Description of change

Fixes https://github.com/typeorm/typeorm/issues/11846

Currently we try to `require` with a dynamic package name which doesn't work well with tools like webpack. This fixes the issue by reading the version from `this.redis.version` which uses `PlatformTools.load` to load the package. `load` uses hardcoded strings to `require()` packages which can be resolved by webpack.

### Pull-Request Checklist

-   [x] ~Documentation has been updated to reflect this change (`docs/docs/**.md`)~


src/cache/RedisQueryResultCache.ts
```diff
@@ -300,23 +300,21 @@ export class RedisQueryResultCache implements QueryResultCache {
...
-     * Detects the Redis package version by reading the installed package.json
-     * and sets the appropriate API version (3 for callback-based, 5 for Promise-based).
+     * Detects the Redis server version by loading the redis package
      */
     private detectRedisVersion(): void {
         if (this.clientType !== "redis") return
-        const version = PlatformTools.readPackageVersion("redis")
-        const major = parseInt(version.split(".")[0], 10)
+
+        const redisVersion = PlatformTools.load("redis").version
```
alumni: The `redis`/`ioredis` package is already loaded in `this.redis` when we call `this.loadRedis()` - we might as well use that directly and even remove this `detectRedisVersion`.

alumni (commented):
Would be great if you could avoid that `Platform.load` and reuse `this.client`.

I think there might be some unnecessary code in `connect` that could have been cleaned up when `detectRedisVersion` was added in #11815 - if it's quick, it's much appreciated.

---

Probably this PR is more important for 0.3.

In 1.0 we are planning to drop support for older versions of `redis` and `ioredis` and we'll finally be able to clean up this implementation. They both support the async API, so we can remove the version detection and the callbacks in another PR.

I think we should chat about how we handle the 0.3 release. I don't mind cherry picking, I actually prefer it in release branches since it makes you go over the code again and avoids bad merges. In that case the PR can still target the `master` branch and we cherry pick to 0.3. Targeting the `v0.3` branch would be fine in any case.

alumni (approved):

gioboa (approved):
Thanks for your help @mguida22

mguida22:
@alumni thanks for the feedback. I've updated this PR to read from the already loaded package and to target `v0.3`

src/cache/RedisQueryResultCache.ts
```diff
@@ -299,30 +289,10 @@ export class RedisQueryResultCache implements QueryResultCache {
...
-            this.redisMajorVersion = 5
-        }
-    }
-
     /**
      * Checks if Redis version is 5.x or higher
      */
-    private isRedis5OrHigher(): boolean {
+    private isPromiseBasedApi(): boolean {
         if (this.clientType !== "redis") return false
```
alumni, line 304: actually `ioredis` is hybrid I think :P but not a problem to keep it like this :)

cyberemissary:
I've tried the fix in the project, but version detection still fails: https://github.com/typeorm/typeorm/issues/11846#issuecomment-3675350851

gioboa (commented):
is this PR still valid?

mguida22:
> is this PR still valid?

It's not ready in its current state (I can't seem to convert back to a draft PR on mobile).

I don't have time at the moment to investigate further but will try to pick it up again next week

pkuczynski:
hey @mguida22, are you still interested in this PR? It has massive conflicts that would need to be resolved first, and it looks like it's targeting `v0.3` — should it be retargeted to `master` instead? In its current state we can't really proceed with it.

PR #11855, opened 2025-12-18
fix: specify columns for metadata-less multi-value inserts

### LINKED ISSUE #3837: Manager insert query with auto-generated columns
**Issue type:**

[ ] question
[x] bug report
[ ] feature request
[ ] documentation issue

**Database system/driver:**

[ ] `cordova`
[ ] `mongodb`
[ ] `mssql`
[ ] `mysql` / `mariadb`
[ ] `oracle`
[x] `postgres`
[ ] `cockroachdb`
[ ] `sqlite`
[ ] `sqljs`
[ ] `react-native`
[ ] `expo`

**TypeORM version:**

[ ] `latest`
[ ] `@next`
[x] `0.2.5` (or put your version here)

I have a table named `difficulties` which have two fields: an `id` field (primary, auto-generated) and a `type` varchar field.

I'm running the following insert:

```
        return connection.manager.insert('difficulties', [
            { type: 'Easy' },
            { type: 'Medium' },
            { type: 'Hard' },
        ]);
```

To my surprise, this query was generated:

Screen Shot 2019-03-17 at 10 38 10 AM

The result is unexpected to me because I specified that I wanted to insert the `type` column, but Typeorm didn't specify that in the query.

Digging down through the code I found out this function, which appears to be the one where the issue is:

https://github.com/typeorm/typeorm/blob/1521f186a46014dbac9fb6886ad8a97907a28680/src/query-builder/InsertQueryBuilder.ts#L352

```
// <some code>.
// in the case if there are no insert columns specified and table without metadata used
// we get columns from the inserted value map, in the case if only one inserted map is specified
if (!this.expressionMap.mainAlias!.hasMetadata && !this.expressionMap.insertColumns.length) {
  const valueSets = this.getValueSets();
  if (valueSets.length === 1)
    return Object.keys(valueSets[0]).map(columnName => this.escape(columnName)).join(", ");
 }
// <some code>.
```

if I remove the check ` if (valueSets.length === 1)`, Typeorm generates the correct query and the information is inserted successfully:

Screen Shot 2019-03-17 at 10 46 10 AM

My questions would be:

1. Am I using the entity manager the wrong way? If so, how do I specify the columns that I want to insert into a table passing the table's name as a string (not creating an entity for that)?

2. I don't have enough context about the code or understand the code enough to be able to make a PR removing it - perhaps it's fixing something that I'm not aware of. In that case, should we come up with a better solution, other than removing the check ` if (valueSets.length === 1)`?

Cprakhar:
### Description of change
- Fix InsertQueryBuilder to always emit column lists for metadata-less multi-row inserts.
- Add regression coverage for manager.insert() and QueryBuilder inserts without entity metadata.

### Pull-Request Checklist
- [x] This pull request links relevant issues as `Fixes #3837`
- [x] There are new or updated tests validating the change (tests/**.test.ts)
- [x] Documentation has been updated to reflect this change (docs/docs/**.md) (N/A)

### Fixes #3837


src/query-builder/InsertQueryBuilder.ts
```diff
@@ -928,6 +928,23 @@ export class InsertQueryBuilder<
...
+        const columns: string[] = []
+        valueSets.forEach((valueSet) => {
+            Object.keys(valueSet).forEach((columnName) => {
+                if (columns.indexOf(columnName) === -1) {
+                    columns.push(columnName)
+                }
+            })
+        })
+        return columns
+    }
```
gioboa, line 1071:
```suggestion
    protected getColumnNamesFromValueSets(
        valueSets: ObjectLiteral[],
    ): string[] {
        const columns = new Set<string>()
        for (const valueSet of valueSets) {
            for (const key in valueSet) {
                columns.add(key)
            }
        }
        return Array.from(columns).sort()
    }
```

Usually for of is faster than forEach and Set is faster too because you don't need to check if the element is present or not. Sorting the final column list ensures the generated SQL is always deterministic, which is crucial for testing and query plan caching.

gioboa (changes requested):
Thanks for your help @Cprakhar 
I added a small feedback.

src/query-builder/InsertQueryBuilder.ts
Cprakhar: Sorting column list results in test failures due to incorrect data type assignments to columns. This issue is related to #11902.
```log
1) benchmark > bulk-save > case-querybuilder
       testing bulk save of 10000 objects:
     QueryFailedError: invalid input syntax for type timestamp with time zone: "label/0"
      at PostgresQueryRunner.query (src/driver/postgres/PostgresQueryRunner.ts:331:19)
      at processTicksAndRejections (node:internal/process/task_queues:103:5)
      at InsertQueryBuilder.execute (src/query-builder/InsertQueryBuilder.ts:166:33)
      at /home/ghostblasius/typeorm/test/benchmark/bulk-save-querybuilder/bulk-save-querybuilder.test.ts:44:17      ```

**Review by gioboa — APPROVED** [review]
Thanks, it looks great to me 👏

**Review by gioboa — COMMENTED** [review]
Can you address the qodo feedback please?

**Comment by Cprakhar** [comment]
> Can you address the qodo feedback please?

I will look into it.

**Comment by michaelbromley** [comment]

```
git fetch origin
```

Issue #9984, opened 2023-04-25
TransactionRetryWithProtoRefreshError should be handled by TypeORM


simplenotezy:
### Issue description

Properly handle TransactionRetryWithProtoRefreshError in CockroachDB

### Expected Behavior

The client should either retry or simply fail and let application handle the error. CockroachDB describes how to handle the error here: https://www.cockroachlabs.com/docs/v22.2/transaction-retry-error-reference.html

### Actual Behavior

The app crashes, and a full restart is needed

### Steps to reproduce

It's difficult for me to provoke a `TransactionRetryWithProtoRefreshError` error, but it may happen if theres many updates to the same row

### My Environment

| Dependency          | Version  |
| ---                 | ---      |
| Operating System    |          |
| Node.js version     | 16.14.2  |
| Typescript version  | 4.7.4  |
| TypeORM version     | 0.3.6  |
| NestJS     | 9.3.9  |

### Additional Context

We have a cronjob that runs a query that sometimes will result in a `TransactionRetryWithProtoRefreshError` error. The TypeORM query is wrapped in a try/catch, but despite this, the whole cornjob will end with the following error:

```
/node_modules/typeorm/driver/cockroachdb/CockroachQueryRunner.js:225
            throw new QueryFailedError_1.QueryFailedError(query, parameters, err);
                  ^

QueryFailedError: restart transaction: TransactionRetryWithProtoRefreshError: WriteTooOldError: write for key /Table/6691/1/"@"/"\x17r\xb0\x80 |LW\xa5\x99\a\x03ED\xbf>"/"\x90\xf6\xf4\xbfNQMO\xb8`\x19\xa8\x00<\xa8o"/"L\xaa!<\x9e\xa7N\xb7\xac\x8b\x13W`V\xe2\xeb"/"ǫ\xd6Ϲ@D\x8c\x8d#\x9f\xac\r\xf4\xa8y"/0 at timestamp 1682440869.268046258,0 too old; wrote at 1682440896.583088232,2: "sql txn" meta={id=c1097e86 key=/Table/6687/1/"@"/"\x13R6s\xfa\"M\x83\xabVť\x04\xa7Z\x8b"/0 pri=0.00328254 epo=0 ts=1682440896.731858898,2 min=1682440869.268046258,0 seq=45} lock=true stat=PENDING rts=1682440869.268046258,0 wto=false gul=1682440869.518046258,0
[...]

PR #11861, opened 2025-12-20
fix(cockroachdb): preserve structured query results during txn retry replay

naorpeled:
### Summary
CockroachDB transaction retry (40001) replays the recorded queries via `ROLLBACK TO SAVEPOINT cockroach_restart`.

During that replay, TypeORM was **not preserving** the original `useStructuredResult` flag per query. That can change the return shape of re-executed queries during retries, which in turn can prevent generated values (e.g. primary keys) from being propagated back onto entities after a retried `INSERT`.

On CockroachDB v24 this shows up in `test/github-issues/9984` where `post.id` can remain `undefined` after a retried transaction, and later `findOneBy({ id: post.id })` fails (especially now that strict handling for undefined where values is enabled).

### Fix
- Record `useStructuredResult` alongside `{ query, parameters }` in the Cockroach retry buffer.
- Replay each stored query using the same `useStructuredResult` value it had originally.

### Why it’s needed
- Keeps retry replay behavior consistent with the original execution.
- Ensures generated values/IDs are applied the same way on retry as on the first attempt.
- Prevents follow-up queries from seeing `undefined` identifiers after a retried write.

### Tests
- Reproduces/fixes the failure in `github issues > #9984` on CockroachDB v24 with `inject_retry_errors_enabled=true`.


gioboa:
Are we closing any issue with this fix?

naorpeled:
> Are we closing any issue with this fix?

Doesn't seem like it, I tried to search but couldn't find any relevant ones

PR #11864, opened 2025-12-21
refactor: remove shajs

G0maa:
### Description of change

Remove `sha.js` dependency, replace with built-in `crypto` module in node environments, and our `RandomGenerator` in other environments, internally `RandomGenerator` uses implementation from [Locutus](https://locutus.io/php/strings/sha1/index.html).

There was a commented function call in `RandomGenerator` that is `unescape()`, this line was replaced entirely with `TextEncoder` which is supported by both [browsers](https://developer.mozilla.org/en-US/docs/Web/API/TextEncoder) and [Hermes](https://docs.expo.dev/versions/v52.0.0/sdk/encoding/) (RN/Expo). To be honest it shouldn't matter anyway the only case where the older implementation would fail (i.e. behave differently that proper SHA1 implementations) is if we use non-ASCII characters, I've looked into where we use `RandomGenerator` or `hash` function:

1. [DefaultNamingStrategy](https://github.com/typeorm/typeorm/blob/09db48cba1236ab45baa2aaaeb274c5e8a844617/src/naming-strategy/DefaultNamingStrategy.ts#L63)
2. [PathUtils](https://github.com/typeorm/typeorm/blob/master/src/util/PathUtils.ts#L26)
3. [DriverUtils](https://github.com/typeorm/typeorm/blob/master/src/driver/DriverUtils.ts#L144)

I don't think any of these instances should accept non-ASCII characters to begin with.

Such discrepancy in where we use `RandomGenerator` vs `hash` is up to the reviewer. Maybe I should refactor all instances of `hash` to use `RandomGenerator` and update `RandomGenerator` to use built-in crypto if existent, otherwise use Locutus implementation? let me know what you think.

I fixed some linting warnings along the way.

### Pull-Request Checklist

### DIFF
--- a/src/util/StringUtils.ts
+++ b/src/util/StringUtils.ts
@@ -1,4 +1,3 @@
-import { PlatformTools } from "../platform/PlatformTools"
 import { RandomGenerator } from "./RandomGenerator"
 
 /**
@@ -111,6 +110,14 @@ export function shorten(input: string, options: IShortenOptions = {}): string {
     return shortSegments.join(separator)
 }
 
+/**
+ * Checks if the current environment is Node.js.
+ * @returns `true` if the current environment is Node.js, `false` otherwise.
+ */
+function isNode(): boolean {
+    return typeof process !== "undefined" && !!process.versions?.node
+}
+
 interface IHashOptions {
     length?: number
 }
@@ -124,7 +131,7 @@ interface IHashOptions {
  */
 export function hash(input: string, options: IHashOptions = {}): string {
     let sha1: string
-    if (PlatformTools.isNode()) {
+    if (isNode()) {
         const crypto = require("node:crypto") as typeof import("node:crypto")
         const hashFunction = crypto.createHash("sha1")
         hashFunction.update(input, "utf8")
--- a/src/platform/PlatformTools.ts
+++ b/src/platform/PlatformTools.ts
@@ -299,12 +299,4 @@ export class PlatformTools {
         console.log(ansi.black.bgRed(prefix))
         if (err) console.error(err)
     }
-
-    /**
-     * Checks if the current environment is Node.js.
-     * @returns `true` if the current environment is Node.js, `false` otherwise.
-     */
-    static isNode(): boolean {
-        return typeof process !== "undefined" && !!process.versions?.node
-    }
 }


src/util/RandomGenerator.ts
```diff
@@ -29,22 +31,24 @@ export class RandomGenerator {
...
         let H4 = 0xc3d2e1f0
-        let A, B, C, D, E
-        let temp
+        let A: number, B: number, C: number, D: number, E: number
+        let temp: number
 
         // utf8_encode
-        str = /*unescape*/ encodeURIComponent(str)
+        const bytes = new TextEncoder().encode(str)
+        str = String.fromCharCode(...bytes)
```
G0maa: `unescape(encodeURIComponent(str))` cannot handle non-ASCII characters. `TextEncoder` can.

tsconfig.json
```diff
@@ -5,7 +5,7 @@
     "emitDecoratorMetadata": true,
     "experimentalDecorators": true,
     "importHelpers": true,
-    "module": "commonjs",
+    "module": "es2022",
```
G0maa: Should be fine since we supposedly [support](https://github.com/typeorm/typeorm/blob/e27ba7ad55960fdf698f4a420eb5ad9cdc14568e/docs/docs/guides/8-migration-v1.md#:~:text=The%20lowest%20JavaScript%20version%20supported%20is%20now%20ES2023%2C%20which%20means%20Node%2020%20and%20later%20is%20supported.%20If%20you%20are%20using%20a%20platform%20that%20does%20not%20support%20ES2023%2C%20please%20upgrade.) ES2023 now?

src/util/StringUtils.ts
```diff
@@ -1,9 +1,17 @@
-import shajs from "sha.js"
+import { RandomGenerator } from "./RandomGenerator"
+let crypto: typeof import("node:crypto") | undefined
+try {
+    crypto = await import("node:crypto")
+} catch {
+    crypto = undefined
+}
```
G0maa: Conditional importing of `node:crypto` if it exists.

tsconfig.json
G0maa: nvm, too many commonjs `require` in the project.

src/util/StringUtils.ts
G0maa: outdated.

src/util/StringUtils.ts
```diff
@@ -106,22 +110,39 @@ export function shorten(input: string, options: IShortenOptions = {}): string {
...
  * @param options.length Optionally, shorten the output to desired length.
+ * @returns SHA-1 hex digest
  */
 export function hash(input: string, options: IHashOptions = {}): string {
-    const hashFunction = shajs("sha1")
-    hashFunction.update(input, "utf8")
-    const hashedInput = hashFunction.digest("hex")
+    let sha1: string
+    if (isNode()) {
+        const crypto = require("node:crypto") as typeof import("node:crypto")
```
G0maa, line 135: We should maybe consider replacing all commonjs `require` with esm `import`.

docs/docs/guides/8-migration-v1.md
```diff
@@ -15,3 +15,9 @@ The `connectorPackage` option was removed, together with the support for the old
 ## Expo
 
 Support for the legacy Expo SQLite driver has been removed. The legacy API was removed by Expo in SDK v52, so you'll need to use Expo SDK v52 or later with the modern async SQLite API.
+
+## hashing
+
+Historically TypeORM used a non-standard SHA-1 implementation for hashing. This has been changed to use the built-in `crypto` module from Node.js.
```
G0maa, line 21: "Non-standard" => we didn't UTF-8 encode non-ascii characters correctly.

test/github-issues/9457/issue-9457.test.ts
G0maa, line 1: This is a console log of before & after string encoding before and after using `TextEncoder`, you see in the old one using `encodeURIComponent` had `%2C`  instead of a `,` (comma).

Old (broken):
```
Input string: migrations_id
After utf8_encode: migrations_id
Input string: example_entity_id
After utf8_encode: example_entity_id
Input string: example_entity_enumcolumn IN ('enumvalue1','enumvalue2','enumvalue3')
After utf8_encode: example_entity_enumcolumn%20IN%20('enumvalue1'%2C'enumvalue2'%2C'enumvalue3')
Input string: example_entity_enumcolumn IN ('enumvalue1','enumvalue2','enumvalue3','enumvalue4')
After utf8_encode: example_entity_enumcolumn%20IN%20('enumvalue1'%2C'enumvalue2'%2C'enumvalue3'%2C'enumvalue4')
```

New (fixed):
```
Input string: migrations_id
After utf8_encode: migrations_id
Input string: example_entity_id
After utf8_encode: example_entity_id
Input string: example_entity_enumcolumn IN ('enumvalue1','enumvalue2','enumvalue3')
After utf8_encode: example_entity_enumcolumn IN ('enumvalue1','enumvalue2','enumvalue3')
Input string: example_entity_enumcolumn IN ('enumvalue1','enumvalue2','enumvalue3','enumvalue4')
After utf8_encode: example_entity_enumcolumn IN ('enumvalue1','enumvalue2','enumvalue3','enumvalue4')
```

src/util/RandomGenerator.ts
alumni: There's no need to use `String.fromCharCode` and `String.prototype.charCodeAt`, we can work directly with the bytes

src/util/RandomGenerator.ts
alumni: Since we're changing these lines, we might as well do some code cleanup.

We can move these declarations inside the block where they are used, makes things easier to follow.

I think the original implementation was ported from another programming language which requires all variables to be declared at the beginning, but since JS/TS allows more flexibility, it's always good to have the declarations as close as possible to the context where they are used.

```suggestion

            for (let i = 7; i >= 0; i--) {
                const v = (val >>> (i * 4)) & 0x0f
                str += v.toString(16)
            }
```

Maybe similar cleanup could be done bellow as well.

src/util/StringUtils.ts
alumni: Normally platform-dependent code is put in `PlatformUtils`, so I'd be tempted to move this function there.

In the long term, it would be nice to find a way to get rid of `PlatformUtils` though. 🤷🏻‍♂️

src/util/RandomGenerator.ts
G0maa: refactored :+1:

src/util/StringUtils.ts
G0maa: Moved to `isNode` to `PlatformUtils` :+1:

src/util/RandomGenerator.ts
G0maa: removed :+1:

linbrandon4:
Nice cleanup overall. Dropping sha.js and leaning on node:crypto where it exists is a good move (fewer deps + less surface area). I also like that you kept behavior working outside Node by falling back to RandomGenerator / internal SHA1, and that you added more hashing tests instead of just refactoring and hoping nothing breaks.

A couple things I’d keep an eye on:
- Environment handling / imports: the conditional or dynamic require("node:crypto") pattern can be touchy with bundlers. If this code path can ever be bundled for browser/RN, it’s worth making sure the detection is clean and doesn’t trigger crypto resolution at build time.
- TextEncoder path: switching off unescape() is good, and TextEncoder is widely supported, but String.fromCharCode(...bytes) can hit argument limits if bytes ever gets large. Probably fine for naming/paths, but chunking would make it more robust.
- Behavior consistency: you mention RandomGenerator vs hash usage across DefaultNamingStrategy, PathUtils, DriverUtils. I’d just make sure the same input produces the same hash everywhere (especially for migrations / naming), and that non-ASCII inputs won’t create surprising differences.

PR #11865, opened 2025-12-21
fix(sqlite): handle simple-enum arrays correctly

### LINKED ISSUE #6326: CHECK constraint failed with SQLite and simple-enum array
**Issue type:**

[x] question
[ ] bug report
[ ] feature request
[ ] documentation issue

**Database system/driver:**

[ ] `cordova`
[ ] `mongodb`
[ ] `mssql`
[ ] `mysql` / `mariadb`
[ ] `oracle`
[ ] `postgres`
[ ] `cockroachdb`
[x] `sqlite`
[ ] `sqljs`
[ ] `react-native`
[ ] `expo`

**TypeORM version:**

[x] `latest`
[ ] `@next`

Hi,

When trying to use a simple-enum array as a column type with SQLite 3.22.0 and TypeORM 0.2.25, I get an SQL_CONSTRAINT CHECK error:

```
 QueryFailedError: SQLITE_CONSTRAINT: CHECK constraint failed: user
    at new QueryFailedError (/path/node_modules/typeorm/error/QueryFailedError.js:11:28)
    at Statement.handler (/path/node_modules/typeorm/driver/sqlite/SqliteQueryRunner.js:48:38) {
  errno: 19,
  code: 'SQLITE_CONSTRAINT',
  query: 'INSERT INTO "user"("id", "username", "password", "email", "roles", "support") VALUES (NULL, ?, ?, ?, ?, ?)',
  parameters: [
    'admin',
    'xxx',
    'bla@bla.com',
    '0,1',
    1
  ]
}
```

The enum column in my entity:

```
    @Column({
        type: 'simple-enum',
        enum: UserRole,
        array: true,
        nullable: false
      })
    roles: UserRole[];
```

The UserRole type:

```
export enum UserRole {
    ADMIN,
    DEVELOPER,
    USER,
}
```

The code I use to instantiate an user:

```
this.create({
            username: 'admin',
            password: 'xxx',
            email: 'bla@bla.com',
            roles: [UserRole.ADMIN, UserRole.DEVELOPER],
            support: true
        })
```

Did I miss something?

Note: if I put one value only in the roles array during instantiation, the new user is indeed created.

Thanks!

Cprakhar:
### Description of change
This PR fixes issue #6326 where SQLite simple-enum columns with array: true were failing with CHECK constraint errors.

#### Problem:
SQLite doesn't support native array types, so TypeORM stores simple-enum arrays as comma-separated strings (e.g., `"0,1"`). However, the existing implementation had three issues:
- **CHECK constraints were validating against single enum values** - The generated CHECK constraint was `CHECK ("roles" IN ('0','1','2'))`, which fails when the stored value is a comma-separated string like `'0,1'`
- **Array default values weren't properly quoted** - Default values like [NumericEnum.GHOST, NumericEnum.ADMIN] were converted to unquoted `3,0` in SQL, causing syntax errors instead of the required `'3,0'`
- **Array values weren't being converted during hydration** - Values stored as `'3,0'` were being returned as strings instead of arrays like `[3, 0]`

#### Solution:
- Skip CHECK constraint generation for array columns in `AbstractSqliteQueryRunner.buildCreateColumnSql()` since the comma-separated string format makes validation impossible
- Properly quote array default values in `AbstractSqliteDriver.normalizeDefault()`
- Update `DateUtils.simpleEnumToString()` to join array values with commas
- Update `DateUtils.stringToSimpleEnum()` to split comma-separated strings back into arrays and convert numeric enum values appropriately

#### Verification:
- All existing tests in enums-array.test.ts now pass
- Tests cover creating default values and saving/retrieving arrays with various enum types (numeric, string, heterogeneous)
- Tested with both sqlite and better-sqlite3 drivers

### Pull-Request Checklist
- [x] This pull request links relevant issues as `Fixes #6326`
- [x] There are new or updated tests validating the change (tests/**.test.ts)
- [ ] Documentation has been updated to reflect this change (docs/docs/**.md)

### Fixes #6326

### DIFF
--- a/.pr_agent.toml
+++ b/.pr_agent.toml
@@ -2,13 +2,19 @@
 pr_commands = [
     "/review",
     "/improve",
+    "/agentic_review"
 ]
 
 handle_push_trigger = true
 push_commands = [
     "/improve",
+    "/agentic_review"
 ]
 
 [auto_best_practices]
 enable_auto_best_practices = true
-utilize_auto_best_practices = true 
+utilize_auto_best_practices = true
+
+[review_agent]
+enabled = true
+publish_output = true
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -116,14 +116,14 @@ services:
   # sap hana
   # works only on linux, minimum 10GB RAM for docker required
   hanaexpress:
-    image: "saplabs/hanaexpress:2.00.076.00.20240701.1"
+    image: "saplabs/hanaexpress:2.00.088.00.20251110.1"
     container_name: "typeorm-hanaexpress"
     hostname: hxe
     command:
       [
+        "--agree-to-sap-license",
         "--passwords-url",
         "file:////hana/hxe-config.json",
-        "--agree-to-sap-license",
       ]
     ulimits:
       nofile: 1048576
--- a/docs/docs/drivers/postgres.md
+++ b/docs/docs/drivers/postgres.md
@@ -50,6 +50,8 @@ See [Data Source Options](../data-source/2-data-source-options.md) for the commo
 
 -   `installExtensions` - A boolean to control whether to install necessary postgres extensions automatically or not (default: `true`)
 
+-   `extensions` - List of additional Postgres extensions to be installed in the database (default: `undefined`)
+
 -   `applicationName` - A string visible in statistics and logs to help referencing an application to a connection (default: `undefined`)
 
 -   `parseInt8` - A boolean to enable parsing 64-bit integers (int8) as JavaScript numbers. By default, `int8` (bigint) values are returned as strings to avoid overflows. JavaScript numbers are IEEE-754 and lose precision over the maximum safe integer (`Number.MAX_SAFE_INTEGER = +2^53`). If you require the full 64-bit range consider working with the returned strings or converting them to native `bigint` instead of using this option.
--- a/docs/docs/drivers/sap.md
+++ b/docs/docs/drivers/sap.md
@@ -25,6 +25,7 @@ See [Data Source Options](../data-source/2-data-source-options.md) for the commo
 -   `pool` — Connection pool configuration object:
     -   `maxConnectedOrPooled` (number) — Max active or idle connections in the pool (default: 10).
     -   `maxPooledIdleTime` (seconds) — Time before an idle connection is closed (default: 30).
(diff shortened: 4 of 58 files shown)


gioboa (approved):
Thanks @Cprakhar it looks great to me 👏

src/util/DateUtils.ts
```diff
@@ -252,17 +252,42 @@ export class DateUtils {
...
+            return value.join(",")
+        }
         return "" + value
     }
 
+    /**
+     *  Converts given string to simple enum or array of enums based on the column metadata.
+     */
     static stringToSimpleEnum(value: any, columnMetadata: ColumnMetadata) {
+        if (columnMetadata.isArray && typeof value === "string") {
```
alumni, line 273: How about this?

```ts
static stringToSimpleEnum(value: any, columnMetadata: ColumnMetadata) {
    if (!columnMetadata.enum) {
        return value
    }
    
    if (columnMetadata.isArray && typeof value === "string") {
        if (value === "") {
            return []
        }
    
        return value.split(",").map(item => DateUtils.parseEnumValue(item, columnMetadata.enum))
    }
    
    return DateUtils.parseEnumValue(value, columnMetadata.enum)
}

private static parseEnumValue(value, enumValues) {
    const parsedValue = Number(value)
    return !isNaN(parsedValue) && enumValues.includes(parsedValue) ? parsedValue : value
}

```

`parseInt('123abc') === 123` vs `Number('123abc') is NaN`

Cprakhar: Good catch!. Using `Number()` instead of `parseInt()` properly handles edge cases like `'123abc'`.

test/functional/database-schema/simple-enums-array/enums-array.test.ts
```diff
@@ -18,7 +18,7 @@ describe("database schema > simple enum arrays", () => {
     before(async () => {
         connections = await createTestingConnections({
             entities: [__dirname + "/entity/*{.js,.ts}"],
-            enabledDrivers: ["postgres"],
+            enabledDrivers: ["postgres", "sqlite", "better-sqlite3"],
```
gioboa: I think we should test cockroachdb too

Cprakhar: Sure, will test it for cockroachdb.

src/util/DateUtils.ts
alumni: Yeah, probably we have this issue in more places 😆

gioboa (approved):

PR #11878, opened 2025-12-27
fix: scope invalidWhereValuesBehavior to high-level abstractions only

### LINKED ISSUE #11818: invalidWhereValuesBehavior.undefined set to `throw` is not throwing when is not the only property in `where`
### Issue description

invalidWhereValuesBehavior.undefined set to `throw` is not throwing when is not the only property in `where`

### Expected Behavior

I have an entity A related to entity B, if B entity `UUID` is undefined it should throw an error.

```ts
a_entity.findOne({
  where: {
    uuid: validUuid,
    b_entity: {
      uuid: undefined,
    },
  },
  relations: {
    b_entity: true,
  },
})
```

### Actual Behavior

It is returning the A entity when it should have thrown an error.

### Steps to reproduce

```ts
    type: "postgres",
    // ...
    invalidWhereValuesBehavior: {
        undefined: 'throw',
        null: 'throw',
    },
});

// 1. a_entity.findOne() - Doesn't work as expected, should throw error but is returning A entity
a_entity.findOne({
  where: {
    uuid: validUuid,
    b_entity: {
      uuid: undefined,
    },
  },
  relations: {
    b_entity: true,
  },
})

// 2. QueryBuilder - Doesn't work as expected, should throw error but is returning A entity
 await this.a_entity.createQueryBuilder('A')
      .innerJoinAndSelect('A.b_entity', 'B')
      .where('A.uuid = :validUuid', { validUuid })
      .andWhere('B.uuid = :bUuid', { bUuid: undefined })
      .getOne()
``` 

### My Environment

| Operating System    |     macOS 15.7.2     |
| Node.js version     | 24.4.1 |
| Typescript version  | 5.9.3 |
| TypeORM version     | 0.3.27 |

### Additional Context

_No response_

- [ ] aurora-mysql
- [ ] aurora-postgres
- [ ] better-sqlite3
- [ ] cockroachdb
- [ ] cordova
- [ ] expo
- [ ] mongodb
- [ ] mysql
- [ ] nativescript
- [ ] oracle
- [x] postgres
- [ ] react-native
- [ ] sap
- [ ] spanner
- [ ] sqlite
- [ ] sqlite-abstract
- [ ] sqljs
- [ ] sqlserver

naorpeled:
## Problem

`invalidWhereValuesBehavior` was being applied to QueryBuilder's `.where()` method, which is a low-level API. This meant that users who set `null: "throw"` or `undefined: "throw"` would get errors even when using QB directly — where they might intentionally be passing dynamic values and handling the SQL themselves.

The setting should only guard against accidental null/undefined in high-level APIs like `.find()`, `repository.update()`, and `manager.delete()`, where null/undefined silently changes query semantics (e.g., `findBy({ id: undefined })` returns the first row).

### Core
- **Removed** `invalidWhereValuesBehavior` checks from `QueryBuilder.getWherePredicateCondition()` — QB `.where()` now passes null/undefined through as-is (pre-feature behavior)
- **Added** `OrmUtils.validateWhereCriteria()` — recursive validation for object-style where criteria, called before QB execution
- **Added** validation calls in `EntityManager.update()`, `.delete()`, `.softDelete()`, `.restore()` (object criteria branch)
- **Enhanced** nested relation handling in `SelectQueryBuilder.buildWhere()` — throws on all-undefined nested objects when configured

### Tests
- `find-options.test.ts` — converted QB `.where()` calls to `.setFindOptions()`, added nested relation partial-undefined test
- `query-builders.test.ts` — rewritten to test EntityManager/Repository methods, added tests verifying QB `.where()` is unaffected

### Docs
- Updated null/undefined handling docs to reflect the scoped behavior
- Added section documenting how null/undefined actually behave in QB `.where()` (generates `WHERE col = NULL`, always false)
- Added summary table comparing behavior across high-level vs low-level APIs

## What's affected

| API | `invalidWhereValuesBehavior` applies? |
|-----|---------------------------------------|
| `repository.find()` / `findOne()` / `findBy()` | Yes |
| `manager.update()` / `delete()` / `softDelete()` / `restore()` | Yes |
| `repository.update()` / `delete()` | Yes |
| `qb.setFindOptions()` | Yes |
| `qb.where()` / `.andWhere()` / `.orWhere()` | No |


gioboa (approved):
It looks a good piece of code to me 👏 @naorpeled

alumni:
Seems that we call `validateWhereParameters` every time before `setParameters`. I got a bit confused by this and was wondering why can we validate the parameters when we set them :P But then I realized we're validating the `FindOptionsWhere` argument. Maybe we can name the function to imply that this is what we validate ;)

naorpeled:
> Seems that we call `validateWhereParameters` every time before `setParameters`. I got a bit confused by this and was wondering why can we validate the parameters when we set them :P But then I realized we're validating the `FindOptionsWhere` argument. Maybe we can name the function to imply that this is what we validate ;)

Good point, wdyt about `enforceInvalidWhereValuesBehavior`?

alumni (changes requested):
I think these checks should apply only to `FindOptions`.

In the query builder API, I don't think we should validate the parameters we send to the database client, since it's similar to executing a raw query, e.g. this should be fine:

```ts
const posts = await createQueryBuilder(PostEntity, 'post')
  .where(
    `COALESCE("post"."category", :default, 'default') = 'default'`,
      { default: null }
  ).getMany();
```

However, it should (maybe, not sure?) check `QueryBuilder.setFindOptions(findOptions)` and `QueryBuilder.where(findOptionsWhere)`, e.g. we could make this throw:
```ts
const posts = await createQueryBuilder(PostEntity, 'post')
  .where({ category: null })
  .getMany();
```

I said "maybe, not sure?" because the `QueryBuilder` should allow the user to do anything they want that is not possible with the `EntityManager`/`Repository` APIs.

naorpeled:
Closing as this requires additional discussion,
@alumni's points make a lot of sense and we'll need to revisit this post v1.

cc @michaelbromley

alumni (approved):

src/util/OrmUtils.ts
```diff
@@ -634,4 +637,71 @@ export class OrmUtils {
             }
         }
     }
+
+    /**
+     * Recursively validates an object where clause, throwing for null/undefined
```
alumni, line 642: @naorpeled why do we need recursion?

naorpeled: As far as I remember it was due to nested relations

alumni: ah okay, also saw the test, i understand now :) thank you!

PR #11880, opened 2025-12-28
feat: support for polymorphic relationship

IamSanjayGupta:
This change improves QueryBuilder support for polymorphic relations by
generating properly scoped JOIN conditions based on explicit polymorphic
metadata configuration.

Polymorphic relations are defined using the following options:

- `idColumnName` – stores the target entity identifier
- `entityColumnName` – stores the discriminator value
- `value` – identifies the concrete target entity

Example:

```
    @ManyToOne(() => Article, {
      polymorphic: {
        idColumnName: "targetId",
        entityColumnName: "targetType",
        value: "articles",
      },
    })
```
When querying multiple polymorphic relations pointing to the same identifier
column (e.g. Article and Video), the QueryBuilder now automatically generates
JOIN conditions scoped by both the identifier and discriminator:

LEFT JOIN article ON article.id = comment.targetId
  AND comment.targetType = 'articles'
LEFT JOIN video ON video.id = comment.targetId
  AND comment.targetType = 'videos'

This allows multiple polymorphic targets to be eagerly loaded in a single query
using:

relations: { article: true, video: true }

without triggering N+1 queries or incorrect joins.

### Verification

- Added a new functional test:
  `tests/functional/relations/polymorphic/polymorphic-find.ts`
- The test validates:
  - correct discriminator-based JOIN generation
  - support for loading multiple polymorphic relations simultaneously
  - absence of additional queries (no N+1 behavior)

All existing and new tests pass locally.


gioboa (commented):
Thanks @IamSanjayGupta 
Is this PR related to a specific issue?

IamSanjayGupta:
> Thanks @IamSanjayGupta Is this PR related to a specific issue?

Hey @gioboa ,

This PR is not related to any issue, its feature supported by sequalize(https://sequelize.org/docs/v7/associations/polymorphic-associations/#single-model-single-foreign-key-polymorphic-associations) but lacking in Typeorm. I need this feature in my project but could find support of this in current version of typeorm so i have to make multiple foreign keys unnecessary.

gioboa (commented):
Thanks for your clarification 🙏

alumni:
There's also a community package implementing this, `typeorm-polymorphic`.

IamSanjayGupta:
> There's also a community package implementing this, `typeorm-polymorphic`.

Hi @alumni,
 Thanks for highlighting this package. I’ve tested it in detail and noticed a few issues that are worth pointing out:

1. 1+N query issue – The package still executes 1 query for the parent entity and N additional queries to load related child entities. This is a significant performance concern, and the issue appears to still be open in the library.

2.  Hard-coded column names – The column names are fixed (entityId, entityType), which limits flexibility and makes integration with existing schemas more difficult.

3. Child relations not populated – When retrieving a parent entity, the related child entities are not automatically populated.

The changes I’ve proposed remove the additional custom logic and instead rely on existing decorators, repository, while also addressing the issues above.
 
4. No Typescript support when using AbstractPolymorphicRepository.createRepository.... to create repository, Many more....



image

IamSanjayGupta:
@alumni @sgarner @naorpeled @michaelbromley requesting your review to close this PR ASAP.

IamSanjayGupta:
Hey help here...?

gioboa (commented):
We need to review internally this PR because it's adding a new feature. Thanks 🙏

alumni:
Some time ago I set the "Future" milestone (after 1.0), since I think we should focus on other priorities at the moment.

There is already an external package implementing this feature. I'd recommend using that one instead at the moment.

I'd also like to have something like this, after working with Django's generic relations, it was one thing I was missing in TypeORM, but at the same time, I'd avoid introducing this in 1.0.

michaelbromley:

```
git fetch origin
```

PR #11883, opened 2025-12-30
test: verify column names when using `referencedColumnName` in composite relations

### LINKED ISSUE #7242: TypeORM generates invalid column name for relation when `referencedColumnName` is used with another composite relation
## Issue Description
TypeORM generates an invalid column name for relation when `referencedColumnName` is used with another composite relation.
This seems like some sort of caching. Apparently, we missed the possibility of one field getting used in two or more relations.

Consider the following entities:

**game.entity.ts**
```typescript
@Entity()
export class Game extends BaseEntity {
    @PrimaryGeneratedColumn({ name: "id" })
    public id!: number;
}
```

**participant.entity.ts**
```typescript
@Entity()
export class Participant extends BaseEntity {
    @PrimaryColumn({ name: "user_id", type: "int", nullable: false })
    @Index()
    public userId!: number;

    @PrimaryColumn({ name: "game_id", type: "int", nullable: false })
    @Index()
    public gameId!: number;
}
```

**match.entity.ts**
```typescript
@Entity()
export class Match extends BaseEntity {
    @PrimaryGeneratedColumn({ name: "id" })
    public id!: number;

    @JoinColumn({ name: "game_id", referencedColumnName: "id" })
    @ManyToOne(() => Game, (e) => e.id, { eager: true })
    public game!: Game;

    @Column({ name: "user_id", type: "int", nullable: false })
    public userId!: number;

    @JoinColumn(
        [
            { name: "game_id", referencedColumnName: "gameId" },
            { name: "user_id", referencedColumnName: "userId" },
        ]
    )
    @ManyToOne(() => Participant, { eager: true })
    public participant: Participant | null = null;
}
```

And the following piece of code:
```typescript
const query = this.matches.createQueryBuilder("match");
query.leftJoinAndSelect("match.game", "game");
console.log(query.getSql());
```

### Expected Behavior

Should not confuse two different relations with each other and generate the following query:

```SQL
SELECT
   "match"."id" AS "match_id",
   "match"."user_id" AS "match_user_id",
   "match"."game_id" AS "match_game_id",
   "game"."id" AS "game_id" 
FROM
   "match" "match" 
   LEFT JOIN
      "game" "game" 
      ON "game"."id" = "match"."game_id"
```

### Actual Behavior
```SQL
SELECT
   "match"."id" AS "match_id",
   "match"."user_id" AS "match_user_id",
   "match"."game_id" AS "match_game_id",
   "game"."id" AS "game_id" 
FROM
   "match" "match" 
   LEFT JOIN
      "game" "game" 
      ON game.gameId = "match"."game_id"
```

The problem is with the `game.gameId = "match"."game_id"`. As you can clearly see, TypeORM thinks that it should use `gameId` for the name of the field in a relationship that has nothing to do with `gameId`. This value is probably selected based on the `referencedColumnName` of the other composite relation on the entity. 

### Steps to Reproduce
Discussed above

### My Environment

```json
  "dependencies": {
    "@nestjs/core": "^7.6.1",
    "pg": "^8.5.1",
    "reflect-metadata": "^0.1.13",
    "typeorm": "^0.2.29"
  },
```

- [ ] `aurora-data-api`
- [ ] `aurora-data-api-pg`
- [ ] `better-sqlite3`
- [ ] `cockroachdb`
- [ ] `cordova`
- [ ] `expo`
- [ ] `mongodb`
- [ ] `mysql`
- [ ] `nativescript`
- [ ] `oracle`
- [X] `postgres`
- [ ] `react-native`
- [ ] `sap`
- [ ] `sqlite`
- [ ] `sqlite-abstract`
- [ ] `sqljs`
- [ ] `sqlserver`

- [X] Yes, I have the time, but I don't know how to start. I would need guidance.

Cprakhar:
Close #7242
Added a test that shows that issue #7242 is resolved.

### DIFF
--- a/test/functional/metadata-builder/relation-join-column-builder/entity/Game.ts
+++ b/test/functional/metadata-builder/relation-join-column-builder/entity/Game.ts
@@ -1,7 +1,10 @@
-import { Entity, PrimaryGeneratedColumn } from "../../../../../src"
+import { Column, Entity, PrimaryGeneratedColumn } from "../../../../../src"
 
 @Entity()
 export class Game {
     @PrimaryGeneratedColumn({ name: "id" })
     id: number
+
+    @Column()
+    title: string
 }
--- a/test/functional/metadata-builder/relation-join-column-builder/relation-join-column-builder.test.ts
+++ b/test/functional/metadata-builder/relation-join-column-builder/relation-join-column-builder.test.ts
@@ -103,17 +103,16 @@ describe("metadata builder > RelationJoinColumnBuilder", () => {
         Promise.all(
             dataSources.map(async (dataSource) => {
                 const game = await dataSource.getRepository(Game).save({
-                    id: 1,
+                    title: "Some Game",
                 })
                 const participant = await dataSource
                     .getRepository(Participant)
                     .save({
                         userId: 1,
-                        gameId: 1,
+                        gameId: game.id,
                     })
 
                 await dataSource.getRepository(Match).save({
-                    id: 1,
                     game: game,
                     userId: 1,
                     participant: participant,
@@ -136,19 +135,17 @@ describe("metadata builder > RelationJoinColumnBuilder", () => {
                 expect(hasCorrectJoin).to.be.true
 
                 const matches = await query.getMany()
-                expect(matches).to.deep.equal([
-                    {
-                        id: 1,
-                        game: {
-                            id: 1,
-                        },
-                        userId: 1,
-                        participant: {
-                            userId: 1,
-                            gameId: 1,
-                        },
-                    },
-                ])
+
+                expect(matches).to.have.lengthOf(1)
+                const match = matches[0]
+                expect(Number(match.id)).to.equal(1) // Some DBs return string ids
+                expect(Number(match.userId)).to.equal(1) // Some DBs return string ids
+                expect(match.game).to.exist
+                expect(Number(match.game.id)).to.equal(1) // Some DBs return string ids
+                expect(match.game.title).to.equal("Some Game")
+                expect(match.participant).to.exist
+                expect(Number(match.participant!.userId)).to.equal(1) // Some DBs return string ids
+                expect(Number(match.participant!.gameId)).to.equal(1) // Some DBs return string ids
             }),
         ))
 })


Cprakhar:
> ## PR Reviewer Guide 🔍
> Here are some key observations to aid the review process:
> 
> **🎫 Ticket compliance analysis ✅**
> 
> **[7242](https://github.com/typeorm/typeorm/issues/7242) - PR Code Verified**
> 
> Compliant requirements:
> 
> * Reproduce the issue with entities having both simple and composite relations
> * Test entities match the structure described in the issue (Game, Participant, Match)
> * Verify generated SQL query uses correct column names in JOIN conditions
> * Test includes both leftJoinAndSelect for game and participant relations
> 
> Requires further human verification:
> 
> * Verify the test actually fails before the fix and passes after (to confirm it properly tests the bug)
> * Confirm the test runs specifically against PostgreSQL driver as mentioned in the issue
> * Validate that the SQL assertion correctly identifies the bug scenario
> 
> ⏱️ **Estimated effort to review**: 2 🔵🔵⚪⚪⚪
> 🧪 **PR contains tests**
> 🔒 **No security concerns identified**
> ⚡ **Recommended focus areas for review**
> 
> [**Test Assertion Logic**](https://github.com/typeorm/typeorm/pull/11883/files#diff-b8817fd48cc5d8bfcbec813920b17a80d288b0a2ef144f9781ae25120f2a05ecR129-R136)
> The test uses an OR condition to check for correct SQL in both MySQL and PostgreSQL formats. However, this approach may pass even if one database generates incorrect SQL while the other is correct. Consider splitting into database-specific test cases or ensuring both formats are validated when applicable.
> 
> ```ts
> const hasCorrectJoin =
>     sql.includes(
>         "LEFT JOIN `game` `game` ON `game`.`id`=`match`.`game_id`  LEFT JOIN `participant` `participant` ON `participant`.`user_id`=`match`.`user_id` AND `participant`.`game_id`=`match`.`game_id`",
>     ) ||
>     sql.includes(
>         `LEFT JOIN "game" "game" ON "game"."id"="match"."game_id"  LEFT JOIN "participant" "participant" ON "participant"."user_id"="match"."user_id" AND "participant"."game_id"="match"."game_id"`,
>     )
> expect(hasCorrectJoin).to.be.true
> ```
> 
> [**Order Difference**](https://github.com/typeorm/typeorm/pull/11883/files#diff-7166e1b2f84009d4a4aa9e3202ab784ba73c8719a1e5c11294d41b5ea54a6f34R23-R26)
> The order of JoinColumn array elements differs from the original issue. Original has `game_id` first, then `user_id`, but test has `user_id` first, then `game_id`. While this may not affect the bug reproduction, it's worth verifying this doesn't impact the test's ability to catch the specific caching issue mentioned.
> 
> ```ts
> @JoinColumn([
>     { name: "user_id", referencedColumnName: "userId" },
>     { name: "game_id", referencedColumnName: "gameId" },
> ])
> ```

The MySQL error unique constraint not found for FK_ for the entity participant appears if the order is changed. The order specified in JoinColumn[] and Participant.ts should be the same.

gioboa (approved):
Thanks for this amazing help @Cprakhar  you rock 🚀

G0maa (approved):

PR #11887, opened 2026-01-02
fix(postgres): drop views with no metadata

### LINKED ISSUE #11209: Postgres queryRunner.dropView does not take into account schema
### Issue description

When `queryRunner.dropView()` is passed a `View` object, it does not take into account the schema

### Expected Behavior

Like `dropTable`, `dropView` should include the schema name when resolving the path to drop.

### Actual Behavior

There is a difference between `dropTable`, which if given a `Table` object calculates the path given the schema, and `dropView`. The latter tries to drop the view using only its name.

### Steps to reproduce

Create a view that belongs to the non-default schema, using migrations, in the `up()` step.
Try to drop that view in the `down()` step and run the rollback.

### My Environment

| Node.js version     | 22.12.0  |
| Typescript version  | 5.6.3  |

### Additional Context

_No response_

- [ ] aurora-mysql
- [ ] aurora-postgres
- [ ] better-sqlite3
- [ ] cockroachdb
- [ ] cordova
- [ ] expo
- [ ] mongodb
- [ ] mysql
- [ ] nativescript
- [ ] oracle
- [X] postgres
- [ ] react-native
- [ ] sap
- [ ] spanner
- [ ] sqlite
- [ ] sqlite-abstract
- [ ] sqljs
- [ ] sqlserver

Nosfistis:
### Description of change

Fix a postgres issue with `.dropView(myView)` that failed when the view to be dropped is not the default one.
Fix a postgres issue with `.dropView(myView)` when used on a view created with `.createView(myView)`, or `.createedView(myView, false)`, i.e. created with no metadata set in `typeorm_metadata`.

Closes #11209

### Pull-Request Checklist


gioboa (commented):
Thanks @Nosfistis for your help.
Few tests are failing can you verify them please?

Nosfistis:
Apologies @gioboa as I had troubles running the tests locally. It should pass now.

Nosfistis:
@gioboa I do not know why the test fails. I cannot run the CI docker container locally, as it is private. Running the tests in a postgres docker image with vector & postgis installed they pass. Could we run the tests with query logs?

gioboa (commented):
btw tests are running on your personal repo too, e.g. [here](https://github.com/Nosfistis/typeorm/actions/runs/20703614964/job/59429984428)
you can add few logs to debug the tests.

Nosfistis:
I found out what causes the crash. While trying to emulate the behaviour of typeorm's metadata table, I set the loaded view expression to the definition stored in `pg_views` or `pg_matviews`. However, this string is not exactly the same as the one provided in the view metadata. I am not sure how this can be tackled yet. This causes the view to drop, and then the synchronize attempts to delete the index which is already deleted.

It might be that storing metadata in the `typeorm_metadata` table should be mandatory. Otherwise, since recreating a view is harmless (unless other views depend on it), the view could be recreated each time, but the index drop should check if it exists beforehand.

Nosfistis:
@gioboa The issue is fixed, however CI seems not to be working.

gioboa (commented):
We need to trigger it manually 😅

src/driver/postgres/PostgresQueryRunner.ts
```diff
@@ -3544,15 +3546,14 @@ export class PostgresQueryRunner
...
-            `INNER JOIN "pg_namespace" "n" ON "n"."oid" = "c"."relnamespace" AND "n"."nspname" = "t"."schema" ` +
-            `WHERE "t"."type" IN ('${MetadataTableType.VIEW}', '${
-                MetadataTableType.MATERIALIZED_VIEW
-            }') ${viewsCondition ? `AND (${viewsCondition})` : ""}`
+        const query = `SELECT CASE WHEN "relkind" = 'm' THEN '${MetadataTableType.MATERIALIZED_VIEW}' ELSE '${MetadataTableType.VIEW}' END AS "type",
+                COALESCE(v."schemaname", "mv"."schemaname")   AS "schema",
+                "relname"                                     AS "name",
+                COALESCE("v"."definition", "mv"."definition") AS "value"
+            FROM "pg_class" "i"
+                    LEFT JOIN "pg_views" "v" ON "v"."viewname" = "i"."relname"
```
smith-xyz: Wondering if this could be:

```ts
const query =
    `SELECT CASE WHEN "i"."relkind" = 'm' THEN '${MetadataTableType.MATERIALIZED_VIEW}' ELSE '${MetadataTableType.VIEW}' END AS "type", ` +
    `"n"."nspname" AS "schema", ` +
    `"i"."relname" AS "name", ` +
    `COALESCE("v"."definition", "mv"."definition") AS "value" ` +
    `FROM "pg_class" "i" ` +
    `INNER JOIN "pg_namespace" "n" ON "n"."oid" = "i"."relnamespace" ` +
    `LEFT JOIN "pg_views" "v" ON "v"."viewname" = "i"."relname" AND "v"."schemaname" = "n"."nspname" ` +
    `LEFT JOIN "pg_matviews" "mv" ON "mv"."matviewname" = "i"."relname" AND "mv"."schemaname" = "n"."nspname" ` +
    `WHERE "i"."relkind" IN ('v', 'm') ${viewsCondition ? `AND (${viewsCondition})` : ""}`
```

with viewsCondition updated to match:

```ts
return `("n"."nspname" = '${schema}' AND "i"."relname" = '${tableName}')`
```

reasons:

- pg_namespace join gives the authoritative schema for each pg_class row
- adding schema to the LEFT JOIN conditions eliminates cross-schema duplicates (same-named views in different schemas each join to both pg_views rows without it)
- schema column now comes from "n"."nspname" directly instead of COALESCE from view catalogs
- relkind IN ('v', 'm') filters out non-view pg_class entries — without it the 1=1 load-all path returns tables, indexes, sequences, etc.

src/driver/postgres/PostgresQueryRunner.ts
```diff
@@ -774,7 +774,9 @@ export class PostgresQueryRunner
      * @param ifExists
      */
     async dropView(target: View | string, ifExists?: boolean): Promise<void> {
-        const viewName = InstanceChecker.isView(target) ? target.name : target
+        const viewName = InstanceChecker.isView(target)
+            ? [target.schema, target.name].filter(Boolean).join(".")
```
smith-xyz: nit: could this be used instead?

```ts
this.driver.buildTableName(target.name, target.schema)
```

test/github-issues/11887/entities/Foo.ts
```diff
@@ -0,0 +1,12 @@
+import { PrimaryGeneratedColumn, UpdateDateColumn } from "../../../../src"
```
smith-xyz, line 1: we're moving tests to match to their functional areas and not be marked with gh issues, please move tests to test/functional/query-runner/drop-view.test.ts

src/query-runner/BaseQueryRunner.ts
```diff
@@ -311,7 +311,14 @@ export abstract class BaseQueryRunner implements AsyncDisposable {
      * @param viewName
      */
     protected async getCachedView(viewName: string): Promise<View> {
-        const view = this.loadedViews.find((view) => view.name === viewName)
+        const view = this.loadedViews.find(
```
smith-xyz, line 314: worried on regression here, does this break for other drivers (think mssql here)? 

```ts
// SQL Server — default database "mydb", default schema "dbo"
const queryRunner = dataSource.createQueryRunner()

const fooView = new View({
    name: "foo_view",
    expression: "SELECT 1 AS col",
})
await queryRunner.createView(fooView)

// try to drop it — does this call as getCachedView("foo_view")
await queryRunner.dropView(fooView)
```

Let's just double check the test coverage on this one just in case

smith-xyz (changes requested):
Overall looks good - a few things to check from my end.

src/driver/postgres/PostgresQueryRunner.ts
Nosfistis: Indeed, good catch.

src/query-runner/BaseQueryRunner.ts
Nosfistis: I don't know much about mssql, I could add a test for that as well if needed.

src/driver/postgres/PostgresQueryRunner.ts
Nosfistis: Sounds good, I will test it.

src/query-runner/BaseQueryRunner.ts
Nosfistis: It looks as if MSSQL also cannot drop foreign-schema views. I added a fix and included the driver in the test.

Nosfistis:
Bump, is there anything else I could do to have this merged?

michaelbromley:

```
git fetch origin
```

PR #11888, opened 2026-01-03
feat: add support for installing additional postgres extensions

### LINKED ISSUE #7584: Allow configurable postgres extensions
## Feature Description

There should be an option to add additional postgres extensions when connecting to a database.

### The Problem

We're using the `tablefunc` module in some of our view entities. Right now, we need to make sure the extension is installed when we create the database. It would be better if TypeORM could ensure the extension is created while it installs other extensions, such as `uuid-ossp`.

### The Solution

From my perspective, ideally we provide an array of extensions when we're configuring the connection. This only really needs to be a string array. [The extensions could then be at the same time as the others.](https://github.com/typeorm/typeorm/blob/7eb032705912cbf4ee340ed9e49970d0f6e23714/src/driver/postgres/PostgresDriver.ts#L323)

### Considered Alternatives

We could add a migration that runs this query, but this isn't compatible with development workflows using `synchronize`. It also feels slightly awkward. An ORM should manage the minutia of a database for us, which includes making sure the necessary features are enabled.

- [ ] `aurora-data-api`
- [ ] `aurora-data-api-pg`
- [ ] `better-sqlite3`
- [ ] `cockroachdb`
- [ ] `cordova`
- [ ] `expo`
- [ ] `mongodb`
- [ ] `mysql`
- [ ] `nativescript`
- [ ] `oracle`
- [x] `postgres`
- [ ] `react-native`
- [ ] `sap`
- [ ] `sqlite`
- [ ] `sqlite-abstract`
- [ ] `sqljs`
- [ ] `sqlserver`

- [x] No, I don't have the time, although I believe I could do it if I had the time...

(I'm a little stretched right now.)

Cprakhar:
### Description of change
This PR addresses #7584 to introduces the functionality to install additional Postgres extensions during connection setup. Verify changes through updated tests for custom extension installation. Documentation reflects the new extension options.

### Pull-Request Checklist
-   [x] This pull request links relevant issues as `Fixes #7584`

### DIFF
--- a/.pr_agent.toml
+++ b/.pr_agent.toml
@@ -2,13 +2,19 @@
 pr_commands = [
     "/review",
     "/improve",
+    "/agentic_review"
 ]
 
 handle_push_trigger = true
 push_commands = [
     "/improve",
+    "/agentic_review"
 ]
 
 [auto_best_practices]
 enable_auto_best_practices = true
-utilize_auto_best_practices = true 
+utilize_auto_best_practices = true
+
+[review_agent]
+enabled = true
+publish_output = true
--- a/test/github-issues/1780/entity/User.ts
+++ b/test/github-issues/1780/entity/User.ts
@@ -1,9 +1,10 @@
 import { Entity } from "../../../../src/decorator/entity/Entity"
 import { Column } from "../../../../src/decorator/columns/Column"
 import { PrimaryGeneratedColumn } from "../../../../src/decorator/columns/PrimaryGeneratedColumn"
-import { Index } from "../../../../src/decorator/Index"
+import { Unique } from "../../../../src"
 @Entity()
-@Index("unique_idx", ["first_name", "last_name"], { unique: true })
+@Unique("unique_first_name", ["first_name"])
+@Unique("unique_name_pair", ["first_name", "last_name"])
 export class User {
     @PrimaryGeneratedColumn()
     id: number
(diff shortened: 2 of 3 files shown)


src/driver/postgres/PostgresDriver.ts
```diff
@@ -405,14 +405,25 @@ export class PostgresDriver implements Driver {
      * Makes any action after connection (e.g. create extensions in Postgres driver).
      */
     async afterConnect(): Promise<void> {
+        const extensionsToInstall = this.options.extensions
         const extensionsMetadata = await this.checkMetadataForExtensions()
         const [connection, release] = await this.obtainMasterConnection()
+        const availableExtensions =
```
gioboa: Can we move this inside the if condition to limit the impact of this change?

src/driver/postgres/PostgresDriver.ts
```diff
@@ -538,6 +576,34 @@ export class PostgresDriver implements Driver {
                     "At least one of the entities has an exclusion constraint, but the 'btree_gist' extension cannot be installed automatically. Please install it manually using superuser rights",
                 )
             }
+
+        if (extensionsToInstall && extensionsToInstall.length > 0) {
+            for (const extension of extensionsToInstall) {
+                if (
+                    availableExtensions &&
```
gioboa: Can we remove this if condition and filter `extensionsToInstall` outside of this method?

gioboa (changes requested):
Thanks @Cprakhar 
I added few comments

src/driver/postgres/PostgresDriver.ts
Cprakhar: Are you suggesting that we create a new method to be called from `enableExtensions` or `afterConnect`?

gioboa: from `afterConnect` the line before the `await this.enableExtensions(` call

PR #11898, opened 2026-01-07
test: align file names

gioboa:
This PR changes these names.

- migrations -> migration
- entities -> entity
- add `.test.ts` extension

### DIFF
--- a/src/driver/mysql/MysqlQueryRunner.ts
+++ b/src/driver/mysql/MysqlQueryRunner.ts
@@ -3378,8 +3378,9 @@ export class MysqlQueryRunner extends BaseQueryRunner implements QueryRunner {
     }
 
     async getVersion(): Promise<string> {
-        const result: [{ "version()": string }] =
-            await this.query("SELECT version()")
+        const result: [{ version: string }] = await this.query(
+            "SELECT version() as version",
+        )
 
         // MariaDB: https://mariadb.com/kb/en/version/
         // - "10.2.27-MariaDB-10.2.27+maria~jessie-log"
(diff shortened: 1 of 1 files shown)


michaelbromley (approved):

test/github-issues/10991/issue-10991.test.ts
```diff
@@ -4,15 +4,15 @@ import {
     createTestingConnections,
     reloadTestingDatabases,
 } from "../../utils/test-utils"
-import { TEST_TABLE_NAME } from "./migrations/init"
+import { TEST_TABLE_NAME } from "./migration/init"
```
pkuczynski, line 7: Hmm, why would we want to change it? It reads more intuitively when the folder name is plural...

michaelbromley: but the rest of the tests all use the singular, so in this case IMO consistency is better.

pkuczynski: We should then probably change it the other way around. It would be off if we would encourage users in the documentation to do something we do not follow ourselves...

gioboa: I see, thanks for the feedback. Sadly all the folders are singular: entity, migration, schema, subscription etc. etc.

pkuczynski: Yeah, we are very inconsistent also across the documentation, as I already checked. In my mind it makes more sense to go with prulal in general:

```
migrations: [__dirname + '/migrations/**/*{.js,.ts}']
// vs
migrations: [__dirname + '/migration/**/*{.js,.ts}']
```

The first one reads more natural with the config prop, but also people tend to name folders in pluralr like `tests`, `controllers`, `routes`, etc...

gioboa:
We can discuss this internally and create another PR to rename all plural.

pkuczynski:
> We can discuss this internally and create another PR to rename all plural.

I fired up https://github.com/typeorm/typeorm/pull/11898 as the base for this discussion...

PR #11901, opened 2026-01-08
chore(mutation testing): tweak Stryker for lower resource usage and decrease timeout for slow running mutants

OSA413:
### Description of change

So, this change brings a few tweaks for mutation testing for it to properly/better work in GHA.

1) I changed number of concurrency from 6 to 4 because the runner often fails/disconnects after 2 hours of work, as the error says, it can be CPU, RAM, or network. So decreasing number of concurrent workers should do the trick here.

2) I enabled the `ignoreStatic` flag to ignore static mutants (though that didn't help a lot). The logs didn't say that there's a problem with a lot of static mutants but I still decided to enable the flag because killing static mutants requires restart of the tests which is a bit slow. More info on static mutants is here https://stryker-mutator.io/docs/mutation-testing-elements/static-mutants/

These two changes decreased the running time from 3 hours to 2 hours, but I thought that that was not enough so I did another trick:

3) I set timeout factor to 0.1 to cancel the slow running tests (it can also be that Stryker decided to re-run all the tests for a particular mutant) which decreased the running time by another hour. Basically, it makes each mutant to run no longer than 40 (300 * 0.1 + 5 + 5) seconds. The math behind this is described here: https://stryker-mutator.io/docs/stryker-js/configuration/#timeoutms-number

All the steps didn't change the mutation score that much (`53.958070324303065` comparing to the final `53.94989967523737`).

In results, the mutation testing should run in just an hour!

### Pull-Request Checklist

### DIFF
--- a/docs/docs/drivers/postgres.md
+++ b/docs/docs/drivers/postgres.md
@@ -50,6 +50,8 @@ See [Data Source Options](../data-source/2-data-source-options.md) for the commo
 
 -   `installExtensions` - A boolean to control whether to install necessary postgres extensions automatically or not (default: `true`)
 
+-   `extensions` - List of additional Postgres extensions to be installed in the database (default: `undefined`)
+
 -   `applicationName` - A string visible in statistics and logs to help referencing an application to a connection (default: `undefined`)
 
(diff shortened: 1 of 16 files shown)


OSA413:
About qodo's suggestion, making all the changes actually decreased the number of timeout-ed mutants by 5 (initial `(545 survived, 75 timed out)` and final `(539 survived, 70 timed out)`)

PR #11916, opened 2026-01-12
fix(postgres): handle timestamptz correctly

Minishlink:
Follow-up of this PR https://github.com/typeorm/typeorm/pull/11774

I won't handle tests this time since there is a lot of confusion around them about this subject. I'll let maintainers merge this as-is or update the relevant tests

This fixes getting Date objects from cached entities instead of strings for columns marked as `timestamptz` (like it was the case for `timestamp with time zone`).


alumni:
We'll merge #11774 since it has tests

PR #11921, opened 2026-01-14
chore(qodo-config): add compliance checklist

naorpeled:
### Description of change
Added some custom checks to Qodo's review process.

### Pull-Request Checklist


pr_compliance_checklist.yaml
```diff
@@ -0,0 +1,30 @@
...
+
+    - title: "Use .test. filenames"
+      compliance_label: true
+      objective: "Tests use the .test. naming convention."
+      success_criteria: "New or renamed tests use .test. in the filename."
+      failure_criteria: "Non-.test. test files are added or modified."
+
+    - title: "Consistent Naming Conventions"
+      compliance_label: true
+      objective: "All new variables, functions, and classes must follow the project's established naming standards"
```
alumni, line 16: Using AI for this might be overkill (also: the prompt for this is not explicit enough if we'd keep it)

Actually this can be enforced using ESLint we if we want to, see https://typescript-eslint.io/rules/naming-convention/

Example config:
```ts
{
    '@typescript-eslint/naming-convention': [
        'error',
        {
            selector: 'typeLike',
            format: ['PascalCase'],
        },
        {
            selector: 'variable',
            types: ['boolean'],
            format: ['PascalCase'],
            prefix: ['is', 'are', 'should', 'has', 'can', 'did', 'will', 'does'],
        },
        ...
    ],
}
```

pr_compliance_checklist.yaml
```diff
@@ -0,0 +1,30 @@
...
+    - title: "Docs updated for user-facing changes"
+      compliance_label: true
+      objective: "Docs reflect public API or usage changes."
+      success_criteria: "Docs or samples update with user-facing changes."
+      failure_criteria: "User-facing changes without doc updates."
+
+    - title: "Breaking changes documented"
+      compliance_label: true
+      objective: "Breaking changes are explicit."
+      success_criteria: "A BREAKING CHANGE note is in the PR description or commit message."
```
alumni: Conventional commits also have `!` after the commit type, e.g.: `feat(topic)!: title`, in additional to the ending.

Maybe a better prompt would be to validate conventional commits. BTW, this PR was also not following conventional commits since it had the `feat` prefix.

The problem that might we have with the `BREAKING CHANGE` note is that it won't be included in commit message when we merge the PR. So we still need to do it manually. We might want to request some information to be added in the upgrade docs. 🤷🏻‍♂️ But honestly, this is so rare that the added complexity for Qodo is probably not worth it (and I'm afraid it might suggest it when it's not relevant).

alumni:
Could we add something to warn us about potentially AI-generated PRs? The ones that I've seen until now were rather low quality.

pr_compliance_checklist.yaml
naorpeled: 100%, I'll create a PR for it 🙏

pr_compliance_checklist.yaml
naorpeled: I get and agree with what you're saying, removing it for now

naorpeled:
> Could we add something to warn us about potentially AI-generated PRs? The ones that I've seen until now were rather low quality.

done, let me know what you think

alumni (approved):

PR #11922, opened 2026-01-15
fix(postgres): use ALTER COLUMN for type changes to prevent data loss

GhostDragonAlpha:
# Fix: Prevent data loss by using ALTER COLUMN for type/length changes (Issue #3357)

/claim #3357

## Description
This PR fixes a critical issue where changing column type or length in Postgres migrations would cause TypeORM to DROP and re-ADD the column, resulting in data loss.

It implements `ALTER COLUMN ... TYPE ... USING ...` logic for these changes, ensuring dataless schema evolution.

## Changes
- Modified `PostgresQueryRunner.changeColumn` to remove the destructive check for `type` and `length` changes.
- Updated `ALTER COLUMN` SQL generation to include the `USING` clause, handling type casting automatically.
- Added unit tests directly verifying the SQL generation logic in `PostgresQueryRunner` without requiring a live database connection.

## Related Issue
Fixes #3357

## Verification
- Added a new unit test `test/unit/driver/postgres/PostgresQueryRunnerMock.test.ts` that mocks `PostgresDriver` and verifies the exact SQL output.
- Verified that `ALTER COLUMN` is generated instead of `DROP COLUMN`.


gioboa (commented):
@GhostDragonAlpha thanks for opening this PR.
I'm finalising this because it is among those still open, the first in chronological order.

gioboa (approved):

pkuczynski:
Hey @GhostDragonAlpha, thanks for this! We're going with #11620 which handles #3357 across all major drivers (including Postgres) and adds a safety guard that falls back to DROP/ADD for cross-family type changes. Closing in favor of #11620 — your review and feedback there would be very welcome!

PR #11926, opened 2026-01-19
fix(mongodb): use cursor.transform for doc to entity transformation and skip load broadcast in next if toArray 

### LINKED ISSUE #11091: Duplicate afterLoad Event Triggered in MongoDB Driver when Using findOne
### Issue description

I'm experiencing an issue with TypeORM when using the MongoDB driver in combination with an EntitySubscriber. Specifically, when executing a findOne operation, the afterLoad event is being triggered twice, even though the findOne method is only called once.

### Expected Behavior

The afterLoad event should be triggered only once per entity load operation.

### Actual Behavior

The afterLoad event is triggered twice for a single findOne operation.

### Steps to reproduce

1. **Create an `EntitySubscriber`** that listens for the `afterLoad` event.

   ```typescript
   import { EntitySubscriberInterface, EventSubscriber, LoadEvent } from 'typeorm';

   @EventSubscriber()
   export class MyEntitySubscriber implements EntitySubscriberInterface {
     afterLoad(entity: any, event?: LoadEvent<any>): void {
       console.log('afterLoad triggered for entity:', entity);
     }
   }
   ```

2. **Perform a `findOne` operation** on an entity using the MongoDB driver.

   ```typescript
   const entity = await this.mongoRepository.findOne({ where: { _id: someObjectId } });
   ```

3. **Observe the output**: The `afterLoad` event is logged twice, indicating it's been triggered twice for a single `findOne` call.

### My Environment

| Operating System    |    MacOS    |
| Node.js version     | 18.19.0  |

### Additional Context

This issue seems specific to the MongoDB driver due to the way it handles cursors and async iteration. It does not appear to affect other database drivers.

**Investigation Details**:

Upon investigating, I found that in the `MongoEntityManager.ts` file, the `applyEntityTransformationToCursor` method overrides both the `toArray` and `next` methods of the cursor to apply entity transformations and broadcast the `Load` event.

Here's the relevant code snippet:

```typescript
protected applyEntityTransformationToCursor<Entity extends ObjectLiteral>(
    metadata: EntityMetadata,
    cursor: FindCursor<Entity> | AggregationCursor<Entity>,
) {
    const queryRunner = this.mongoQueryRunner;

    // Save original toArray method
    (cursor as any)["__to_array_func"] = cursor.toArray;

    // Override toArray method
    cursor.toArray = async () =>
        ((cursor as any)["__to_array_func"] as CallableFunction)().then(
            async (results: Entity[]) => {
                const transformer = new DocumentToEntityTransformer();
                const entities = transformer.transformAll(results, metadata);
                // Broadcast "Load" event
                await queryRunner.broadcaster.broadcast(
                    "Load",
                    metadata,
                    entities,
                );
                return entities;
            },
        );

    // Save original next method
    (cursor as any)["__next_func"] = cursor.next;

    // Override next method
    cursor.next = async () =>
        ((cursor as any)["__next_func"] as CallableFunction)().then(
            async (result: Entity) => {
                if (!result) {
                    return result;
                }
                const transformer = new DocumentToEntityTransformer();
                const entity = transformer.transform(result, metadata);
                // Broadcast "Load" event
                await queryRunner.broadcaster.broadcast("Load", metadata, [
                    entity,
                ]);
                return entity;
            },
        );
}
```

The MongoDB driver's `toArray` method internally uses an async iterator that calls the overridden `next` method, as shown in the driver's code:

```typescript
async *[Symbol.asyncIterator](): AsyncGenerator<TSchema, void, void> {
    if (this.closed) {
      return;
    }

    try {
      while (true) {
        // Look Here 👇
        const document = await this.next();
        // ...
    } finally {
     // ...
  }

async toArray(): Promise<TSchema[]> {
  const array = [];
  for await (const document of this) {
    array.push(document);
  }
  return array;
}
```

This means that when `toArray` is called, it internally calls `next` for each document. Since both `toArray` and `next` have been overridden to broadcast the `Load` event, the event is triggered twice for each document when `toArray` is used.

In the case of `findOne`, which ultimately calls `toArray` with a limit of 1, the `afterLoad` event is triggered twice.

**Possible Solution**:

To resolve this issue, consider modifying the `applyEntityTransformationToCursor` method to prevent double-triggering of the `Load` event. One possible solution is to only broadcast the `Load` event within the `next` method and avoid doing so in the `toArray` method, since `toArray` relies on `next`.

Alternatively, implement a check to prevent the `Load` event from being broadcasted twice for the same entity when `toArray` internally calls `next`.

**Request**:

Could you please investigate this issue and provide guidance or a fix? If there's a workaround that can be implemented on the application side in the meantime, I'd appreciate any suggestions.

Thank you for your assistance!

- [ ] aurora-mysql
- [ ] aurora-postgres
- [ ] better-sqlite3
- [ ] cockroachdb
- [ ] cordova
- [ ] expo
- [X] mongodb
- [ ] mysql
- [ ] nativescript
- [ ] oracle
- [ ] postgres
- [ ] react-native
- [ ] sap
- [ ] spanner
- [ ] sqlite
- [ ] sqlite-abstract
- [ ] sqljs
- [ ] sqlserver

Cprakhar:
Closes #11091

### The Problem
In the current MongoDB implementation, `applyEntityTransformationToCursor` modifies the driver's `next()` and `toArray()` methods. Because the MongoDB driver may internally call both methods on the same cursor instance (for example, during `find` operations), this leads to issue:
- **Duplicate Subscriber Events:** The `EntitySubscriberInterface.afterLoad` method is triggered multiple times for the same database record because each patched method independently transforms the document and emits a "Load" event.

### Description of Change
- Use `cursor.transform` API instead of calling `DocumentToEntityTransformer`.`transform`/`transformAll` in overridden `toArray`/`next` — handles `next()` and `toArray()`. As `transformDocument()` is called internally by the toArray and next methods
- Skip broadcasting in `next` as `toArray` calls it internally (If `next` is called alone, broadcast the load events.)
- Added `transform` in `typings.ts`

### Pull Request Checklist
- [x] Code is up to date with the `master` branch
- [x] This pull request links to relevant issues: `Fixes #11091`
- [x] There are new or updated tests that validate the changes (`tests/**.test.ts`)


gioboa (approved):
It looks great to me, you rock 👏 @Cprakhar

src/entity-manager/MongoEntityManager.ts
```diff
@@ -71,6 +71,11 @@ import type { ColumnMetadata } from "../metadata/ColumnMetadata"
 export class MongoEntityManager extends EntityManager {
     readonly "@instanceof" = Symbol.for("MongoEntityManager")
 
+    private readonly cursorQueues = new WeakMap<
```
alumni: Probably we should also call `.delete()` on this `WeakMap` to cleanup what we don't need anymore.

BTW, since it doesn't implement a queue, can we find a better name for it? What is it storing for every cursor? `cursorToCallbackMap`?

Also, is it necessary to have this? if we know that mongo is always calling `next` after `toArray`, could we maybe only do the transformation on `next`?

Cprakhar: Only patching `next` doesn't work: `toArray` returns an array of plain documents instead of Entity instances.

alonesuperman:
When will it be merged?

src/entity-manager/MongoEntityManager.ts
```diff
@@ -1309,39 +1309,40 @@ export class MongoEntityManager extends EntityManager {
...
+        // override toArray for batch broadcast (skip per-doc broadcast in next)
+        const originalToArray = cursor.toArray.bind(cursor)
+        cursor.toArray = async () => {
+            isInToArray = true
+            try {
+                const entities = await originalToArray()
+                if (entities.length > 0) {
+                    await broadcaster.broadcast("Load", metadata, entities)
+                }
+                return entities
```
smith-xyz, line 1331: any errors that should bubble up to user

smith-xyz:
```typescript
cursor.toArray = async () => {
    isInToArray = true
    try {
        const entities = await originalToArray()
        if (entities.length > 0) {
            await broadcaster.broadcast("Load", metadata, entities)
        }
        return entities
    } catch (e) {
        // if broadcast fails, still return entities?
        // or re-throw?
    } finally {
        isInToArray = false
    }
}
```

trying to understand - need to spin up a mongodb instance to give this a try, but given the tests is a scenario covered where we show how that will work?

smith-xyz (approved):
just a minor note - but it seems to be an existing issue from before, let me know if I'm missing something in that comment

alumni (approved):

gioboa (approved):

PR #11935, opened 2026-01-21
refactor!: remove `Buffer` polyfill and replace `Buffer` with `Uint8Array` on non-node platforms

pujux:
### Description of change

This PR removes the browser `Buffer` polyfill and migrates TypeORM internals toward `Uint8Array` for binary handling, while preserving Node compatibility where drivers still return/expect `Buffer`.

Key changes:

- Remove browser-side `Buffer` global/polyfill wiring from `BrowserPlatformTools.template`.
- Replace `Buffer`-specific checks in core paths with `Uint8Array` checks (`Subject`, `SubjectChangedColumnsComputer`, `ColumnMetadata`, `UpdateQueryBuilder`, `RawSqlResultsToEntityTransformer`, find-option binary typing helpers).
- Keep compatibility in Node drivers by handling both `Uint8Array` and `Buffer` in binary type normalization across SQL drivers.
- Update Mongo typings in migrated areas (`AutoEncryptionOptions` key/privateKey and GridFS write/end signatures) to `Uint8Array`-based types.
- Introduce `Uint8ArrayUtils` and use it consistently for detection/equality/hex conversion.
- Improve binary equality hot paths:
  - `left === right` fast path
  - Node `Buffer.equals` fast path when both values are Buffers
  - byte-loop fallback for portable `Uint8Array` equality
- Ensure deep comparison and column comparison handle `Uint8Array` content equality (including cross-realm `Uint8Array` values).

### Why

- Avoid browser global `Buffer` injection/polyfill behavior.
- Align with web-native binary type (`Uint8Array`) in public/internal typing.
- Preserve driver/runtime compatibility and performance in Node where `Buffer` is still common.

### Tests and docs

- Updated functional SAP vector coverage to validate `Uint8Array` flows.
- Added/updated unit tests for:
  - `ColumnMetadata.compareEntityValue`
  - `OrmUtils.deepCompare`
  - `Uint8ArrayUtils` (including cross-realm and Buffer fast-path behavior)
- Updated docs for vector entity property typing (`docs/docs/entity/1-entities.md`).

### Pull-Request Checklist

-   [x] This pull request links relevant issues as `Fixes #00000` (N/A)


pujux:
> ## PR Code Suggestions ✨
> Explore these optional code suggestions:
> 
> **Category**	**Suggestion                                                                                                                                    **	**Impact**
> General	
> Improve performance of hex conversion
> **Refactor the `uint8ArrayToHex` function to use `Array.from().map().join()` for potentially better performance and a more functional programming style.**
> 
> [src/util/Uint8ArrayUtils.ts [15-21]](https://github.com/typeorm/typeorm/pull/11935/files#diff-d6327efd302fa660cf83ac96383c35d6b9b0eccb3a657b338f6ac4b567abadc1R15-R21)
> 
> ```diff
>  export const uint8ArrayToHex = (value: Uint8Array): string => {
> -    let hex = ""
> -    for (let index = 0; index < value.length; index++) {
> -        hex += value[index].toString(16).padStart(2, "0")
> -    }
> -    return hex
> +    return Array.from(value)
> +        .map((byte) => byte.toString(16).padStart(2, "0"))
> +        .join("")
>  }
> ```
> 
> * [ ]  **Apply / Chat**
> 
> Suggestion importance[1-10]: 4
> __
> 
> Why: The suggestion proposes a valid performance improvement for the `uint8ArrayToHex` function by using a more idiomatic `map` and `join` pattern, which is generally more efficient than string concatenation in a loop.
> 
> Low
> * [ ]  More

I ran a benchmark of the proposed change, my current change and another possible option using a lookup table to not have to call padStart repeatedly - the results speak for themselves.

NodeJS:
```
=== Uint8Array length: 16 bytes | iterations: 2,000,000 ===
lookup table + loop + +=       63.7 ns/op   (1.00x)
loop += (padStart)            186.5 ns/op   (2.93x)
Array.from().map().join()     515.1 ns/op   (8.08x)

=== Uint8Array length: 32 bytes | iterations: 1,500,000 ===
lookup table + loop + +=      115.9 ns/op   (1.00x)
loop += (padStart)            340.9 ns/op   (2.94x)
Array.from().map().join()     910.6 ns/op   (7.85x)

=== Uint8Array length: 64 bytes | iterations: 1,000,000 ===
lookup table + loop + +=      218.8 ns/op   (1.00x)
loop += (padStart)            663.0 ns/op   (3.03x)
Array.from().map().join()      1.76 µs/op   (8.06x)

=== Uint8Array length: 256 bytes | iterations: 300,000 ===
lookup table + loop + +=      780.7 ns/op   (1.00x)
loop += (padStart)             2.58 µs/op   (3.31x)
Array.from().map().join()      6.53 µs/op   (8.37x)

=== Uint8Array length: 1024 bytes | iterations: 120,000 ===
lookup table + loop + +=       3.00 µs/op   (1.00x)
loop += (padStart)            11.98 µs/op   (3.99x)
Array.from().map().join()     26.67 µs/op   (8.89x)

=== Uint8Array length: 8192 bytes | iterations: 20,000 ===
lookup table + loop + +=      24.90 µs/op   (1.00x)
loop += (padStart)           126.61 µs/op   (5.08x)
Array.from().map().join()    246.58 µs/op   (9.90x)

=== Uint8Array length: 65536 bytes | iterations: 2,000 ===
lookup table + loop + +=     272.62 µs/op   (1.00x)
loop += (padStart)             1.27 ms/op   (4.65x)
Array.from().map().join()      2.91 ms/op   (10.66x)
```

Browser (Chrome):
```
=== Uint8Array length: 16 bytes | iterations: 1,500,000 ===
lookup table + loop + +=         50.2 ns/op   (1.00x)
loop += (padStart)              157.5 ns/op   (3.14x)
Array.from().map().join()       484.7 ns/op   (9.65x)

=== Uint8Array length: 32 bytes | iterations: 1,200,000 ===
lookup table + loop + +=         84.2 ns/op   (1.00x)
loop += (padStart)              306.6 ns/op   (3.64x)
Array.from().map().join()       890.3 ns/op   (10.57x)

=== Uint8Array length: 64 bytes | iterations: 800,000 ===
lookup table + loop + +=        156.5 ns/op   (1.00x)
loop += (padStart)              584.0 ns/op   (3.73x)
Array.from().map().join()        1.66 µs/op   (10.58x)

=== Uint8Array length: 256 bytes | iterations: 250,000 ===
lookup table + loop + +=        562.4 ns/op   (1.00x)
loop += (padStart)               2.31 µs/op   (4.11x)
Array.from().map().join()        6.26 µs/op   (11.14x)

=== Uint8Array length: 1024 bytes | iterations: 90,000 ===
lookup table + loop + +=         2.12 µs/op   (1.00x)
loop += (padStart)               9.84 µs/op   (4.64x)
Array.from().map().join()       25.83 µs/op   (12.18x)

=== Uint8Array length: 8192 bytes | iterations: 15,000 ===
lookup table + loop + +=        17.10 µs/op   (1.00x)
loop += (padStart)             106.71 µs/op   (6.24x)
Array.from().map().join()      235.99 µs/op   (13.80x)

=== Uint8Array length: 65536 bytes | iterations: 1,500 ===
lookup table + loop + +=       145.53 µs/op   (1.00x)
loop += (padStart)              1.089 ms/op   (7.48x)
Array.from().map().join()       2.251 ms/op   (15.47x)
```

gioboa (commented):
Some tests are failing, can you verify them?

docs/docs/drivers/sap.md
```diff
@@ -57,20 +57,20 @@ export class Document {
...
 }
 ```
 
-By default, the client will return a `Buffer` in the `fvecs`/`hvecs` format, which is more efficient. It is possible to let the driver convert the values to a `number[]` by adding `{ extra: { vectorOutputType: "Array" } }` to the connection options. Check the SAP HANA Client documentation for more information about [REAL_VECTOR](https://help.sap.com/docs/SAP_HANA_CLIENT/f1b440ded6144a54ada97ff95dac7adf/0d197e4389c64e6b9cf90f6f698f62fe.html) or [HALF_VECTOR](https://help.sap.com/docs/SAP_HANA_CLIENT/f1b440ded6144a54ada97ff95dac7adf/8bb854b4ce4a4299bed27c365b717e91.html).
+By default, the client will return a `Uint8Array` in the `fvecs`/`hvecs` format, which is more efficient. It is possible to let the driver convert the values to a `number[]` by adding `{ extra: { vectorOutputType: "Array" } }` to the connection options. Check the SAP HANA Client documentation for more information about [REAL_VECTOR](https://help.sap.com/docs/SAP_HANA_CLIENT/f1b440ded6144a54ada97ff95dac7adf/0d197e4389c64e6b9cf90f6f698f62fe.html) or [HALF_VECTOR](https://help.sap.com/docs/SAP_HANA_CLIENT/f1b440ded6144a54ada97ff95dac7adf/8bb854b4ce4a4299bed27c365b717e91.html).
```
alumni: The `@sap/hana-client` library will actually return a NodeJS `Buffer` (which is just a subclass of `Uint8Array`). This library only works on NodeJS and uses the NodeJS types, so the docs were fine before.

**This applies to most other drivers that are NodeJS-only. Existing code should still work without breaking changes.** 

The main idea of the task is to remove the dependency on the `Buffer` polyfill in TypeORM, but somehow we should still support how the clients behave.

pujux: yeah I figured this shouldn't change but then apparently I did anyway 😃 will update!

pujux:
@gioboa yeah I'll go through them and see what's been failing and why in the next days, unfortunately end of last week and the next days I'm super busy

src/util/Uint8ArrayUtils.ts
```diff
@@ -0,0 +1,50 @@
+const isBuffer = (value: unknown): value is Buffer =>
```
michaelbromley: Claude picked up on this:

The return type annotation value is Buffer requires `@types/node` to be available at compile time. In a pure browser TypeScript project without Node types, this would cause a compile error. This somewhat undermines the goal of removing the Node.js buffer dependency. The Qodo bot suggestion to use a structural type `(Uint8Array & { equals(other: Uint8Array): boolean })` is spot-on here.

michaelbromley (commented):
Just one comment which might need attention

michaelbromley (approved):

src/driver/aurora-mysql/AuroraMysqlDriver.ts
```diff
@@ -663,7 +663,10 @@ export class AuroraMysqlDriver implements Driver {
             return "varchar"
         } else if (column.type === Date) {
             return "datetime"
-        } else if ((column.type as any) === Buffer) {
+        } else if (
+            (column.type as any) === Uint8Array ||
+            (typeof Buffer !== "undefined" && (column.type as any) === Buffer)
```
alumni: `typeof type === 'function' && type.prototype instanceof Uint8Array` ???

test/functional/database-schema/vectors/sap/entity/Uint8ArrayEmbedding.ts
pujux, line 1: @alumni I suppose this is also not correct here or is it?

alumni: Yeah, it's still `Buffer` :)

alumni (approved):

src/util/Uint8ArrayUtils.ts
```diff
@@ -0,0 +1,52 @@
...
+        }
+    }
+
+    return true
+}
+
+const HEX_LOOKUP_TABLE = Array.from({ length: 256 }, (_, i) =>
+    i.toString(16).padStart(2, "0"),
+)
+export const uint8ArrayToHex = (value: Uint8Array) => {
```
alumni: nice perf optimization :D

btw, `Uint8Array.toHex()` was introduced in node 25

gioboa (approved):

src/util/OrmUtils.ts
```diff
@@ -439,6 +440,10 @@ export class OrmUtils {
         if (x === null || y === null || x === undefined || y === undefined)
             return false
 
+        if (isUint8Array(x) && isUint8Array(y)) {
+            return areUint8ArraysEqual(x, y)
```
alumni, line 453: I think this need to happen after the next `if`.

Also, the next `if` could be adjusted to return `x.equals(y)`, and the condition could be fixed, e.g.:

```ts
        if (typeof x.equals === "function" &&
            typeof y.equals === "function") {
            return x.equals(y)
        }
```

alumni: Sorry, it's a bit late, I also noticed that we have something similar in `areUint8ArraysEqual` :) 

`x.equals(y)` in this file applies not just to `Buffer`, but also to other data types, e.g. Luxon DateTime.

src/util/Uint8ArrayUtils.ts
pujux:
```
  ┌──────┬─────────────────┬────────────────┬───────────────┬───────────────────────────┐                                                                                                                                   
  │ Size │ native .toHex() │  lookup table  │ padStart loop │ Array.from().map().join() │
  ├──────┼─────────────────┼────────────────┼───────────────┼───────────────────────────┤                                                                                                                                   
  │ 16B  │ 15.8 ns (1x)    │ 56.4 ns (3.6x) │ 165 ns (10x)  │ 503 ns (32x)              │                                                                                                                                 
  ├──────┼─────────────────┼────────────────┼───────────────┼───────────────────────────┤                                                                                                                                   
  │ 64B  │ 31.0 ns (1x)    │ 172 ns (5.6x)  │ 627 ns (20x)  │ 1.70 µs (55x)             │                                                                                                                                   
  ├──────┼─────────────────┼────────────────┼───────────────┼───────────────────────────┤                                                                                                                                   
  │ 1KB  │ 290 ns (1x)     │ 2.51 µs (8.7x) │ 10.8 µs (37x) │ 26.0 µs (90x)             │                                                                                                                                   
  ├──────┼─────────────────┼────────────────┼───────────────┼───────────────────────────┤                                                                                                                                   
  │ 64KB │ 34.1 µs (1x)    │ 176 µs (5.2x)  │ 1.14 ms (33x) │ 2.63 ms (77x)             │                                                                                                                                 
  └──────┴─────────────────┴────────────────┴───────────────┴───────────────────────────┘
```

ran another benchmark with native toHex 🙏

PR #11936, opened 2026-01-22
fix(redis): redis cache version detection

mguida22:
### Description of change

Fixes logic to determine if the provided redis client uses a promise or callback based API. This fix calls ping to check if the response returns a function (and therefore uses promises).

### Pull-Request Checklist

### DIFF
--- a/src/cache/RedisQueryResultCache.ts
+++ b/src/cache/RedisQueryResultCache.ts
@@ -87,6 +87,9 @@ export class RedisQueryResultCache implements QueryResultCache {
             if (typeof this.client.connect === "function") {
                 await this.client.connect()
             }
+
+            // Detect API style by checking if methods return Promises
+            this.detectPromiseBasedApi()
         } else if (this.clientType === "ioredis") {
             if (cacheOptions && cacheOptions.port) {
                 if (cacheOptions.options) {
@@ -124,9 +127,6 @@ export class RedisQueryResultCache implements QueryResultCache {
                 )
             }
         }
-
-        // Detect API style by checking if methods return Promises
-        this.detectPromiseBasedApi()
     }
 
     /**


mguida22:
This can use some real world testing. Please [install the version from this branch](https://github.com/typeorm/typeorm/pull/11936#issuecomment-3781760923) and report any feedback.

src/cache/RedisQueryResultCache.ts
```diff
@@ -300,32 +291,18 @@ export class RedisQueryResultCache implements QueryResultCache {
...
-        }
-        if (major <= 4) {
-            // Redis 3/4 uses callback-based API
-            this.redisMajorVersion = 3
-        } else {
-            // Redis 5+ uses Promise-based API
-            this.redisMajorVersion = 5
+    private detectPromiseBasedApi(): void {
+        if (this.clientType !== "redis") {
+            this.isPromiseBasedApi = false
```
alumni: I'm not sure this is correct. `ioredis` supports promises for example, so we could probably set it to `true`.

src/cache/RedisQueryResultCache.ts
```diff
@@ -300,32 +291,18 @@ export class RedisQueryResultCache implements QueryResultCache {
...
-     */
-    private isRedis5OrHigher(): boolean {
-        if (this.clientType !== "redis") return false
-        return (
-            this.redisMajorVersion !== undefined && this.redisMajorVersion >= 5
-        )
+        // Call ping() and check if it returns a Promise
+        // This works because v5+ returns Promises directly, while v4 with legacyMode uses callbacks
+        const result = this.client.ping()
+        this.isPromiseBasedApi = result && typeof result.then === "function"
```
alumni: We could also simplify to e.g. `!!result?.then`, but that's just a nit.

What I'm rather afraid of is that if `ping` will return a `Promise` which will later be rejected, it will end up as an unhandled exception. I'd maybe do `result.catch(noop)` just in case if result is a `Promise`.

_Later edit_: just noticed Qodo also spotted it :)

cyberemissary:
This change resolved the issue I reported #11846 
Thank you

src/cache/RedisQueryResultCache.ts
G0maa: I'll remove it, if both `redis` & `ioredis` support promises then we should depend on the `ping()` method just under this if condition.

src/cache/RedisQueryResultCache.ts
```diff
@@ -126,20 +124,21 @@ export class RedisQueryResultCache implements QueryResultCache {
                 )
             }
         }
+
+        // Detect API style by checking if methods return Promises
+        this.detectPromiseBasedApi()
```
G0maa: This was inside `if` of `redis`. I moved it outside, since promises are existent in both `redis` and `ioredis`. @alumni

src/cache/RedisQueryResultCache.ts
G0maa: nvm, I see that based on `isPromiseBasedApi` we call [different functions in different ways](https://github.com/mguida22/typeorm/blob/03988419951a559615478e1c71b5b92a57a81a24/src/cache/RedisQueryResultCache.ts#L161-L180). Which is different for redis and ioredis. I think generally we need better support for `ioredis` vs `redis` clients, not just detecting promise based APIs.

I returned the `detectPromiseBasedApi()` to inside the `if == redis`. At least this keeps the original behavior, but fixes promise detection for redis.

PR #11945, opened 2026-01-26
docs: add JsonContains find-options example

### LINKED ISSUE #11940: How to use JsonContains in FindOptions
### What was unclear or otherwise insufficient?

JsonContains usage missing in doc https://typeorm.io/docs/working-with-entity-manager/find-options/

### Recommended Fix

-

### Additional Context

_No response_

lraveri:
Adds a `JsonContains` example to the Find Options docs.

Fixes #11940.

### DIFF
--- a/docs/docs/working-with-entity-manager/3-find-options.md
+++ b/docs/docs/working-with-entity-manager/3-find-options.md
@@ -531,7 +531,7 @@ will execute following query:
 SELECT * FROM "post" WHERE "categories" && '{TypeScript}'
 ```
 
-- `JsonContains`
+- `JsonContains` (PostgreSQL/CockroachDB only)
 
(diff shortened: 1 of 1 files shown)


lraveri:
Not 100% sure we should add the “(PostgreSQL/CockroachDB only)” note here, since other operators like arrayContains / arrayContainedBy don’t have a similar disclaimer in the docs.

Also, in the codebase there isn’t any dialect check: TypeORM will generate the SQL as-is, and it will simply fail at runtime if the underlying driver doesn’t support @> / <@ (or ::jsonb).

For consistency, maybe we should either:

- add the same note to all operators relying on dialect-specific syntax, or
- avoid adding it just for this example.

WDYT? @gioboa

Issue #11961, opened 2026-02-05
TypeORM OneToMany select returns only the first relation row instead of all related entities


aroshanzamir:
### Issue description

When using TypeORM with a OneToMany relation and select on the child entity, only the first related entity is returned per parent entity, even though multiple related rows exist in the database.

### Expected Behavior

When fetching multiple users with a OneToMany relation (userRoles), 
all related role entities should be returned for each user, 
not just the first one. 

Example expected output:

```
[
  {
    id: "user1",
    username: "alice",
    userRoles: [
      { roleId: "roleA" },
      { roleId: "roleB" }
    ]
  }
]

```

### Actual Behavior

Only the first related role is returned per user. 

Example actual output:

```
[
  {
    id: "user1",
    username: "alice",
    userRoles: [
      { roleId: "roleA" } // only the first relation
    ]
  }
]
```

Notes:
- Direct SQL in PostgreSQL returns both roles correctly:
  SELECT role_id FROM user_roles WHERE user_id = 'user1';
  -- Returns roleA, roleB
- Using getOne() works fine for a single parent.
- The issue occurs when fetching multiple parents (getMany or getManyAndCount) with selected fields of a OneToMany relation.

### Steps to reproduce

Create entities:

```
@Entity({ name: "users" })
export class UserEntity extends BaseEntity {
    @PrimaryGeneratedColumn("uuid")
    id: string;

    @Column()
    username: string;

    @OneToMany(() => UserRoleEntity, userRole => userRole.user)
    userRoles: UserRoleEntity[];
}

@Entity({ name: "user_roles" })
export class UserRoleEntity extends BaseEntity {
    @PrimaryGeneratedColumn("uuid")
    id: string;

    @Column()
    userId: string;

    @ManyToOne(() => UserEntity, user => user.userRoles)
    @JoinColumn({ name: "user_id" })
    user: UserEntity;

    @Column()
    roleId: string;
}
```

Insert test data:

```
INSERT INTO users(id, username) VALUES ('user1', 'alice');
INSERT INTO user_roles(id, user_id, role_id) VALUES
('r1', 'user1', 'roleA'),
('r2', 'user1', 'roleB');

```

Query using TypeORM:
```
const users = await userRepository.createQueryBuilder("user")
    .leftJoinAndSelect("user.userRoles", "userRole")
    .select(["user.id", "user.username", "userRole.roleId"])
    .getMany();
```

Observe the output: only the first role is returned.

### My Environment

[...]

PR #11969, opened 2026-02-10
fix: correct one-to-many hydration with partial selects

harshang03:
### Description of change

This change fixes OneToMany hydration when using partial selects across multiple parents. Previously, joined isMany relations could be overwritten or grouped incorrectly, returning only the first related row. The update:
Tightens raw-row grouping to only use PKs when present and non-null (otherwise per-row grouping).
Merges isMany relation arrays instead of overwriting, with PK-based de‑duplication.
Ensures relation loader selects include PKs for reliable hydration.
Marks auto-added PKs as virtual to preserve partial-select output shape.

Why needed: without this, getMany() with a OneToMany join + partial select returns only the first related row per parent. This matches the reported behavior in #11961.

How verified: manual reasoning against the join/hydration flow; no automated tests added.

Current behavior:
userRoles contains only the first row for each user.

New behavior:
userRoles contains all related rows for each user.

### Pull-Request Checklist


gioboa (commented):
Thanks for your help @harshang03 
I know, this is a draft PR, but remember to add a test to validate the changes. 💪

gioboa (changes requested):
is this PR still valid?

michaelbromley:

```
git fetch origin
```

PR #11974, opened 2026-02-10
fix: avoid dropping columns on Postgres length changes

swjcpy:
Close #3357

Fix Postgres changeColumn to alter type when only length changes, so data is preserved. Adds a regression test that inserts a row, changes varchar length, and verifies the data remains.


swjcpy:
Fixed the two review points: 1) length-change check now compares createFullType and down query uses oldColumn.name, so no spurious ALTER and rollback works with renames; 2) test now cleans up and runs executeMemoryDownSql in finally.

swjcpy:
Applied suggestion #1: captured stable old/new column names and full types before mutation, and used them in the ALTER TYPE up/down queries to keep rollbacks correct when renames happen.

swjcpy:
Applied latest suggestion: down migration ALTER TYPE now targets the renamed column (newColumnName) so rollback order is correct when a rename happens.

swjcpy:
Applied latest suggestion: after emitting ALTER TYPE for length changes, the in-memory cloned/old column objects are updated to keep state consistent for subsequent operations.

swjcpy:
Applied latest suggestions: detect stored generated columns with full-type changes and force drop/add, and guard length-only ALTER to same base type.

swjcpy:
Fixed lint failure (prefer-const) by changing the new test’s table variable to const.

swjcpy:
Applied the two remaining suggestions: normalize full type strings with trim() before comparison, and avoid test id collisions by using a per-run testId.

swjcpy:
Applied latest suggestions: compare normalized types for isTypeChanged, and parameterize SQL in the new test.

swjcpy:
Applied latest suggestions: length-change ALTER now uses normalized type check (!isTypeChanged), and the test enables sql memory with a safer random id.

Issue #11951, opened 2026-01-30
luxon.DateTime columns triggers update every time


guiligan:
### Issue description

When using luxon.DateTime, columns are updated even without changes, probably by the way it is compared

### Expected Behavior

`DateTime` should be compared through `DateTime.equal`, and not saved

### Actual Behavior

Same value is sent to update, even though not changed

### Steps to reproduce

1. Have any entity with the provided transformer using `DateTime`
```typescript
export class Entity extends BaseEntity {
  @Column({
    type: 'timestamptz',
    transformer: new LuxonTimestampTzTypeT(),
  })
  someDate: DateTime;
}
```
2. load and save an entity with `.save()`

### My Environment

| Dependency          | Version  |
| ---                 | ---      |
| Operating System    |  unix  |
| Node.js version     | 24.3.0  |
| Typescript version  | 5.9.3  |
| TypeORM version     | 0.3.28  |

### Additional Context

*Transformer*
```typescript
import { DateTimeParser } from '@/helpers/datetime-parser.helper';
import { DateTime } from 'luxon';
import { FindOperator, ValueTransformer } from 'typeorm';

export class LuxonTimestampTzTypeT implements ValueTransformer {
  public to(value: FindOperator<any>): FindOperator<any>;
  public to(value: DateTime): Date;
  public to(value: null | undefined): null;
  public to(
    value: DateTime | null | undefined | FindOperator<any>,
  ): Date | null | FindOperator<any> {
    if (value === null || value === undefined) {
      return null;
    } else if (value instanceof FindOperator) {
      return value;
    }

    return value.toJSDate();
  }

  public from(value: Date | string): DateTime;
  public from(value: null | undefined): null;
  public from(value: Date | string | null | undefined): DateTime | null {
    if (value === null || value === undefined) {
      return null;
    } else if (typeof value === 'string') {
      return DateTimeParser(value);
    }

    return DateTime.fromJSDate(value, { zone: 'UTC' });
  }
}
```

### Relevant Database Driver(s)

- [ ] aurora-mysql
- [ ] aurora-postgres
- [ ] better-sqlite3
- [ ] cockroachdb
- [ ] cordova
- [ ] expo
- [ ] mongodb
- [ ] mysql
- [ ] nativescript
- [ ] oracle
- [x] postgres
- [ ] react-native
- [ ] sap
- [ ] spanner
- [ ] sqlite
- [ ] sqlite-abstract
- [ ] sqljs
- [ ] sqlserver

[...]

PR #12003, opened 2026-02-18
fix-luxon-datetime-update

harshang03:
### Description of change

This change fixes false-positive updates for entity columns using Luxon DateTime (or similar date-like objects) with a column transformer. Previously, change detection compared the transformed DB Date with the in-memory Luxon object directly, which caused TypeORM to think the value changed and issue an UPDATE every save.
What changed:
Normalizes “date-like” values (supports objects with toJSDate() / toDate()) before formatting/comparison.
Applies the column transformer before datetime normalization/comparison for temporal columns.
Adds a helper to identify date/time/timestamp column types consistently.
Why needed:
Prevents unnecessary UPDATEs when the Luxon DateTime value is logically unchanged.
How verified:
Code inspection and tracing through SubjectChangedColumnsComputer normalization flow; no automated tests added.
Current behavior:
Saving an unchanged Luxon DateTime field triggers an UPDATE every time.
New behavior:
Unchanged Luxon DateTime values are considered equal, so no UPDATE is generated.

###issue: https://github.com/typeorm/typeorm/issues/11951

### Pull-Request Checklist

### DIFF
--- a/src/metadata/ColumnMetadata.ts
+++ b/src/metadata/ColumnMetadata.ts
@@ -767,6 +767,24 @@ export class ColumnMetadata {
      * Extracts column value from the given entity.
      * If column is in embedded (or recursive embedded) it extracts its value from there.
      */
+    isDateTimeType(): boolean {
+        return (
+            this.type === "date" ||
+            this.type === "time" ||
+            this.type === "time with time zone" ||
+            this.type === "time without time zone" ||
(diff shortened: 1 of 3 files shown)


alumni:
Hi @harshang03, thanks for looking into this, but the PR doesn't seem to be going in the right direction. Let's follow the idea being implemented in #11963.

PR #12023, opened 2026-02-21
docs(jsdoc): add params and other lint fixes

alumni:
### Description of change

Run `pnpm lint --fix`.

### Pull-Request Checklist

### DIFF
--- a/docs/docs/working-with-entity-manager/5-entity-manager-api.md
+++ b/docs/docs/working-with-entity-manager/5-entity-manager-api.md
@@ -177,7 +177,7 @@ await manager.insert(User, [
 ])
 ```
 
-- `update` - Updates entities by entity id, ids or given conditions. Sets fields from supplied partial entity.
+- `update` - Updates entities by entity id, ids, given conditions, or an array of condition objects. Sets fields from supplied partial entity.
 
 ```typescript
 await manager.update(User, { age: 18 }, { category: "ADULT" })
@@ -185,6 +185,26 @@ await manager.update(User, { age: 18 }, { category: "ADULT" })
 
 await manager.update(User, 1, { firstName: "Rizzrak" })
 // executes UPDATE user SET firstName = Rizzrak WHERE id = 1
+
+// Bulk updates with different conditions for each operation
+await manager.update(User, [
+    { criteria: { id: 1 }, data: { firstName: "Rizzrak" } },
+    { criteria: { id: 2 }, data: { firstName: "Karzzir" } },
+    { criteria: { age: 18 }, data: { category: "ADULT" } },
+])
+// executes three separate UPDATE queries:
+// UPDATE user SET firstName = Rizzrak WHERE id = 1
+// UPDATE user SET firstName = Karzzir WHERE id = 2
+// UPDATE user SET category = ADULT WHERE age = 18
+```
+
+You can pass an **array of condition objects** to match multiple distinct sets of rows in a single call (conditions are OR'd together):
+
+```typescript
+await manager.update(User, [{ status: "expired" }, { flagged: true }], {
+    active: false,
+})
+// executes UPDATE user SET active = false WHERE status = 'expired' OR flagged = true
 ```
 
 - `updateAll` - Updates _all_ entities of target type (without WHERE clause). Sets fields from supplied partial entity.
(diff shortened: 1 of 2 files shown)


.github/workflows/pull-request.yml
```diff
@@ -17,19 +17,21 @@ jobs:
       - uses: amannn/action-semantic-pull-request@v6
         with:
           types: |
+            build
```
pkuczynski: What would be an example of `build` type of PR?

.github/workflows/pull-request.yml
```diff
@@ -17,19 +17,21 @@ jobs:
...
             chore
             ci
             docs
             feat
             fix
-            lint
+            perf
             refactor
             revert
+            style
```
pkuczynski: `lint` is better as it's more broad term and not only about the style. Can also prevent bugs or poor performance constructs (see https://en.wikipedia.org/wiki/Lint_(software))

Prettier is purely focused on formatting and styling the code

src/driver/mongodb/bson.typings.ts
```diff
@@ -1,6 +1,5 @@
 /**
  * A class representation of the BSON Binary type.
- * @public
```
pkuczynski, line 3: Since these are not out types, we should not modify them and exclude them from formatting

.github/workflows/pull-request.yml
```diff
@@ -17,19 +17,21 @@ jobs:
...
         with:
           types: |
+            build
             chore
             ci
             docs
             feat
             fix
-            lint
+            perf
```
pkuczynski: These changes are not relevant to the subject of this PR. Can you please move them to a dedicated PR? 

Also subject of this PR should he `docs` since you only modify jsdoc comments

pkuczynski (changes requested):

src/driver/mongodb/bson.typings.ts
alumni: I thought I reverted this file and added it to `.eslintignore`. Will check, thanks!

.github/workflows/pull-request.yml
alumni: We follow the contributing guidelines in `CONTRIBUTING.md` (which are based on the angular conventions), this workflow should should reflect that instead of introducing its own rules.

.github/workflows/pull-request.yml
alumni: There are the rules we follow in this repo (see `CONTRIBUTING.md`).

.github/workflows/pull-request.yml
pkuczynski: True, it's better if we stay consistent but u would still question them. We don't have to follow exactly what angular does. Like for exmple you added `ci` which was not there before. Nonetheless, this should be done in separate PR...

src/driver/mongodb/bson.typings.ts
pkuczynski: Can you add them here https://github.com/typeorm/typeorm/blob/master/eslint.config.mjs#L9 so we have one place to define ignores instead of two?

.eslintignore
```diff
@@ -0,0 +1,2 @@
+src/driver/mongodb/typings.ts
+src/driver/mongodb/bson.typings.ts
```
pkuczynski: Can you add them to https://github.com/typeorm/typeorm/blob/master/eslint.config.mjs#L9 so we have single place for such configurations?

.github/workflows/pull-request.yml
pkuczynski: Yeah, but `style` is prettier and not `lint`

src/driver/mongodb/bson.typings.ts
alumni: Yeah, I think `.eslintignore` is from `eslint@8`, that's why the pre-commit hook reverted the file. Will do.

.github/workflows/pull-request.yml
pkuczynski: I just checked and our contributing guide already diverged from angular source quite a bit: https://github.com/angular/angular/blob/main/contributing-docs/commit-message-guidelines.md#type

alumni: I will extract this change in a separate PR (probably next week), I was hoping it will fix the workflow if I change it here, but GitHub has some nice security rules :)

Will keep the `style:` prefix for the PR title.

alumni:
@pkuczynski Do you have any other suggestion, or can I merge this?

pkuczynski:
> @pkuczynski Do you have any other suggestion, or can I merge this?

Looks good! Thanks for asking :)

I adopted the PR title to better reflect the change in case we need to trace it in the future. Hope you don’t mind...

PR #12032, opened 2026-02-23
fix: use ALTER COLUMN instead of drop/add for type/length changes to prevent data loss

hruss-software:
## Problem

  When changing a column's type or length (e.g., `varchar(50)` → `varchar(100)`), TypeORM generates
  destructive `DROP COLUMN` + `ADD COLUMN` statements instead of `ALTER COLUMN`. This causes
  **complete data loss** for the affected column.

  Reported in #3357 (97 👍, open since 2019).

  ## Solution

  Move type/length changes from the destructive DROP+ADD code path to each database's native ALTER
  COLUMN syntax, which preserves existing data:

  | Driver | ALTER Syntax Used |
  |--------|-------------------|
  | PostgreSQL | `ALTER COLUMN "col" TYPE <type> USING "col"::<type>` |
  | MySQL/MariaDB | `ALTER TABLE CHANGE` (already existed in codebase) |
  | SQL Server | `ALTER TABLE ALTER COLUMN` (already existed in codebase) |
  | Oracle | `ALTER TABLE MODIFY` (already existed in codebase) |
  | CockroachDB | `ALTER COLUMN "col" TYPE <type>` |

  DROP+ADD is preserved only for genuinely destructive changes:
  - Array type conversion (`isArray` changes)
  - Generated/identity column changes
  - Computed column expression changes

  ## Why this approach

  - **Minimal diff (102 lines)** - no new abstractions, no helper functions
  - **Leverages existing code** - MySQL, SQL Server, and Oracle already had ALTER paths for property
  changes; this routes type/length through them
  - **All 5 database drivers** - not just PostgreSQL
  - **PostgreSQL USING clause** - enables safe cross-type casts (e.g., `int` → `text`)
  - **Incompatible conversions fail explicitly** - the database will throw a clear error rather than
  silently dropping data

  ## Test

  Added test case verifying:
  1. Insert data into a column
  2. Change column length via `changeColumn`
  3. Verify data is preserved after the change
  4. Revert via `executeMemoryDownSql`
  5. Verify data is still intact after revert

  Closes #3357

### Pull-Request Checklist


gioboa (commented):
@hruss-software thanks for opening this PR.
I'm closing this for now because tests are failing and it's a duplicate of #11959. 🙏

PR #12037, opened 2026-02-24
feat(mongo)!: remove exported internal types

alumni:
### Description of change

TypeORM was exporting all (internally declared) MongoDB types, including things like `UpdateOptions` or `ConnectionOptions` which in TypeORM also exist or used to exist.

Ideally, we should avoid `export * from ...` and use `export { Thing } from ...` to make sure we are exporting only what we want.

To prevent shadowing our own types and polluting the TypeORM exports, I completely removed the MongoDB types. The users can import them from MongoDB if they need, they would be more accurate anyway.

### Pull-Request Checklist


alumni:
I'm not sure how those two files had 100% coverage since we don't really use anything except `ObjectId` and `MongoClient`. I'm thinking to remove them as well since it's just useless code now, I'd probably need to make a few things private though.

gioboa (approved):

src/index.ts
```diff
@@ -113,7 +113,7 @@ export * from "./repository/TreeRepository"
 export * from "./repository/MongoRepository"
 export type * from "./repository/RemoveOptions"
 export type * from "./repository/SaveOptions"
-export type { UpdateOptions as RepositoryUpdateOptions } from "./repository/UpdateOptions"
```
pkuczynski, line 116: We should mention this in migratuin guide...

PR #12038, opened 2026-02-24
test: reorganize lazy relation tests and standardize naming

pkuczynski:
## Summary

- Move `test/github-issues/47/` regression test to `test/functional/relations/lazy-relations/persistence/` and expand `enabledDrivers` to `["mysql", "postgres"]` since the original bug was in shared lazy-relation loading code
- Rename `lazy-relations-loading-via-base-entitiy-finders/` directory and test file to follow the naming convention used by sibling tests
- Standardize `describe` block naming across all lazy-relations tests to use `"relations > lazy relations > ..."` prefix


test/functional/relations/lazy-relations/persistence/persistence-lazy-relations.test.ts
```diff
@@ -3,27 +3,27 @@ import {
...
+    let dataSources: DataSource[]
     before(
         async () =>
-            (connections = await createTestingConnections({
+            (dataSources = await createTestingConnections({
                 entities: [__dirname + "/entity/*{.js,.ts}"],
-                enabledDrivers: ["mysql"], // we can properly test lazy-relations only on one platform
+                enabledDrivers: ["mysql", "postgres"],
             })),
     )
```
alumni: If you're changing these lines, you might as well do:

```suggestion
    before(async () => {
        dataSources = await createTestingConnections({
            entities: [__dirname + "/entity/*{.js,.ts}"],
            enabledDrivers: ["mysql"], // we can properly test lazy-relations only on one platform
        })
    })
```
(since the return value is not needed)

pkuczynski: done!

PR #12043, opened 2026-02-24
fix: scope STI eager relations to declaring child and auto-detect shared-PK relations

antst:
### Description of change

  What: Fix two bugs in TypeORM's metadata system affecting Single Table Inheritance (STI) and shared-primary-key entity patterns.

  Bug 1 — STI eager relation leaking: In STI hierarchies, eager relations declared on one child entity were incorrectly JOINed when querying a sibling child entity. For example, loading Student would also eagerly JOIN Employee.verification. Root cause:
   EntityMetadata.eagerRelations merges all children's relations into the parent, and FindOptionsUtils/SelectQueryBuilder used this merged list for all child queries without scoping.

  Bug 2 — Shared-PK cascade insert failure: When using @JoinColumn({ name: 'id' }) where the FK column IS the entity's PK (identifying/derived-PK pattern), cascade inserts failed with FK constraint violations. Root cause: the primary decorator option
  was removed in 0.3.0, leaving relation.isPrimary = false and relation.isNullable = true, which caused SubjectTopologicalSorter to not enforce correct insert ordering.

  How it's fixed:
  1. Added declaringTarget to RelationMetadata and getScopedEagerRelations() to EntityMetadata. Query builders now use scoped eager relations that filter out sibling children's relations.
  2. Added auto-detection in RelationJoinColumnBuilder.build(): when all join columns overlap with entity PK columns, relation.isPrimary and relation.isNullable are set correctly.

  Verification: 5 new tests covering both fixes, plus all 102 existing relation and STI tests pass against PostgreSQL 17.

  Fixes #11730 and #11746

### Pull-Request Checklist

- [x] Code is up-to-date with the `master` branch
- [x] This pull request links relevant issues as `Fixes #00000`
- [x] There are new or updated tests validating the change (`tests/**.test.ts`)
- [x] Documentation has been updated to reflect this change (`docs/docs/**.md`)

Issue #11730, opened 2025-10-18
(Transitive) Eager relations not loaded correctly with TableInheritance


Taeir:
### Issue description

(Transitive) Eager relations not loaded correctly with TableInheritance

### Expected Behavior

Eager relationships are loaded whenever `find` methods are used. This applies transitively and should work together with other functionalities like `TableInheritance`. It should be documented when this is not the case.

I have attached a simplified code example below for showcasing the issue.

When loading a questionnaire, it loads its questions eagerly. As part of loading the questions, the question options are also loaded eagerly.

### Actual Behavior

In the general case, transitive eager relationships seem to work as intended. However, when `TableInheritance` is used as in the example, it works correctly for `CheckboxQuestion` but not does not work for `MultipleChoiceQuestion`.

It seems TypeORM is adding a condition on the `LEFT JOIN` which limits the types of relations to one specific type only. The result is that some entities do not get correctly hydrated.

More specifically, the query I see is:

```sql
[...]

Issue #10563, opened 2023-12-21
Default value in child table/entity column decorator for multiple table inheritance is ignored for inherited columns


mellster2012:
### Issue description

For multiple table inheritance the parent column decorator/metadata is always used, though it's valid to override certain properties in the child entity colum deocrator for inherited columns

### Expected Behavior

All the options from the parent column decorator are used for the inherited child column.

### Actual Behavior

Default value should be able to be configurable (distinct from parent column) for child tables for inherited columns for multiple table inheritance.

### Steps to reproduce

See description/behavior.

### My Environment

| Dependency          | Version  |
| ---                 | ---      |
| Operating System    |          |
| Node.js version     | 18.17.1  |
| Typescript version  | 5.2.2  |
| TypeORM version     | 0.3.16 & 0.3.17  |

### Relevant Database Driver(s)

- [ ] aurora-mysql
- [ ] aurora-postgres
- [ ] better-sqlite3
- [ ] cockroachdb
- [ ] cordova
- [ ] expo
- [ ] mongodb
- [ ] mysql
- [ ] nativescript
- [ ] oracle
- [X] postgres
- [ ] react-native
- [ ] sap
- [ ] spanner
- [ ] sqlite
- [ ] sqlite-abstract
- [ ] sqljs
- [ ] sqlserver

### Are you willing to resolve this issue by submitting a Pull Request?

Yes, I have the time, and I know how to start.

SrWildman:
Curious if there is any traction on releasing this. I am running into a similar use case where I have a parent class with a column marked as primary and generated but I would like my child class to not have that column generated.  `@PrimaryColumn('uuid', {generated: false})` does nothing when added to the child class.

kashyap-aditya:
This still seems to be an issue.
A default set on any child seems to want to set the default on the parent and all other children too.

PR #12050, opened 2026-02-26
feat: implement Class Table Inheritance (CTI / Joined Table Inheritance)

### LINKED ISSUE #8629: Major Feature request: Join Table Inheritance pattern
With this pattern:

- when a class `A` imports class `B` two tables are created, with each field in the corresponding table
- A type field is added to table `A` so that you can always instantiate the correct type. 
- Fields that are not selected in the initial fetch are lazy loaded when accessed.

I would like to use typeORM to migrate an existing project. The project is written in php with SilverStripe. The framework uses Join Inheritance pattern. The migration path would involve running both systems in parallel for some time so it would require to be able to use the same pattern in typeorm.

Would you be interested in adding/supporting the pattern to this project.

antst:
### Description of change

This PR implements **Class Table Inheritance** (CTI), also known as Joined Table Inheritance — a long-requested inheritance mapping pattern (see #8629) where each class in the hierarchy gets its own database table, joined by primary key.

#### What it does

With `pattern: "CTI"` in the `@TableInheritance` decorator, TypeORM now maps an inheritance hierarchy to **multiple tables joined by primary key**:

```typescript
@Entity()
@TableInheritance({ pattern: "CTI", column: { type: "varchar", name: "type" } })
export class Actor {
    @PrimaryGeneratedColumn()
    id: number

    @Column()
    name: string
}

@ChildEntity()
export class User extends Actor {
    @Column()
    email: string
}

@ChildEntity()
export class Organization extends Actor {
    @Column()
    industry: string
}
```

This creates **three tables**: `actor` (id, name, type), `user` (id, email), `organization` (id, industry) — where child PKs are foreign keys to the parent.

#### Why this is needed

STI forces all child-specific columns into one table as nullable columns, which degrades schema clarity and data integrity as hierarchies grow. CTI provides clean separation: each table only has its own columns, child-specific columns can have proper NOT NULL constraints, and relations can be placed at any level of the hierarchy.

#### Key capabilities

- **Schema sync**: Child tables are created/dropped alongside parent tables with proper FK constraints
- **Query building**: Child queries auto-INNER JOIN to parent table; parent queries return only parent-table data (no child table JOINs)
- **Persistence**: INSERT/UPDATE/DELETE cascade across parent and child tables (PK propagation)
- **Polymorphic queries**: `actorRepository.find()` instantiates correct child types via discriminator, with parent-table columns populated (child-specific columns require querying child entity directly)
- **Discriminator hydration**: The CTI discriminator `type` property is assigned to loaded entities, available via `(actor as any).type`. STI discriminator behavior is unchanged — virtual discriminators are not hydrated onto STI entities.
- **Multi-level CTI**: Supports hierarchies deeper than 2 levels (e.g., Actor → Contributor → User)
- **Eager relation scoping**: Child-specific eager relations are scoped to the correct type
- **Lazy relations**: Full support for Promise-based lazy loading on CTI entities
- **Relations at any level**: OneToOne, OneToMany, ManyToOne, ManyToMany on parent or child entities
- **Inherited parent relations**: Child entities can load relations defined on the parent via `find({ relations })` and QueryBuilder `leftJoinAndSelect`
- **Cross-child FK references**: CTI children can reference each other (e.g., `Organization.ceo → User`)
- **Self-referential relations**: CTI children support self-referential relations (e.g., `User.manager → User`)
- **UUID primary keys**: Full support for `@PrimaryGeneratedColumn("uuid")` with CTI
- **Abstract intermediate classes**: Supports abstract classes with `@Column` decorators between parent and child (columns inherited by concrete children)
- **Custom child table names**: `@ChildEntity({ tableName: "app_users" })` allows explicit table names for CTI children

#### Design: parent entity queries

Following SQLAlchemy's approach, **CTI root entity queries do NOT LEFT JOIN child tables**. The parent table is a standalone entity — it has its own columns and relations. When you query `actorRepository.find()`, you get:

- Correct child class instances (via discriminator)
- Parent-table columns populated (id, name, type, parent-level relations)
- Child-specific columns are `undefined` (email, industry, etc.)

To access child-specific data, query the child entity directly:

```typescript
// Parent query — lightweight, no child JOINs:
const actors = await actorRepository.find()
actors[0].name       // ← populated (parent column)
(actors[0] as User).email  // ← undefined

// Child query — INNER JOINs parent, full data:
const user = await userRepository.findOneBy({ id: actors[0].id })
user.email           // ← populated
```

This avoids the performance trap of LEFT JOINing every child table on every parent query (which grows linearly with hierarchy size), and keeps the parent entity semantically clean.

#### Bug fixes included

1. **RETURNING clause filtering**: CTI INSERT/UPDATE RETURNING clauses included inherited columns from parent tables, causing PostgreSQL errors. Fix: use `tableColumns` instead of `columns` in `getInsertionReturningColumns()`, `getUpdationReturningColumns()`, `getSoftDeletionReturningColumns()`.

2. **Version/updateDate double-increment**: CTI child UPDATE auto-incremented version/updateDate on both child and parent tables. Fix: skip auto-increment SET clauses when building child UPDATE (parent handles it).

3. **JOIN ordering**: CTI parent table JOINs weren't interleaved correctly with child table JOINs, causing "missing FROM-clause entry" errors for nested CTI relations. Fix: interleave CTI parent JOINs immediately after child table JOINs.

4. **Abstract base class column hydration**: Columns inherited from abstract base classes (id, createdDate, etc.) weren't hydrated onto CTI root entities because `column.target` pointed to the abstract class. Fix: property-based matching (`databaseName` + `propertyPath`) instead of reference-based check.

5. **Discriminator value not hydrated (CTI only)**: The virtual discriminator column was skipped during entity property assignment because `isVirtual` columns were unconditionally excluded. Fix: allow discriminator columns through the `isVirtual` check, **scoped to CTI only** — STI virtual discriminators remain excluded (matching existing STI behavior where auto-created discriminators are not hydrated onto entities).

6. **Root entity queries performance**: Parent entity queries unnecessarily LEFT JOINed all child tables, SELECTed child columns, and loaded child-specific eager relations — O(N) JOINs for N children. Fix: removed all three behaviors, following SQLAlchemy's approach (parent table is standalone).

7. **OneToMany/OneToOneNotOwner on CTI children**: Fixed "missing FROM-clause entry" error when loading OneToMany relations targeting CTI child entities. The STI discriminator condition was incorrectly applied to CTI entities.

8. **Nullable LEFT JOIN on CTI children**: Fixed entities with null nullable relations disappearing from results. When a relation was LEFT JOINed and the related CTI entity was null, the INNER JOIN to its parent table eliminated the entire row.

9. **Column alias collisions**: When two CTI child types define columns with the same `databaseName` (e.g., both User and Organization have "status"), SQL aliases collided. Fix: child-specific columns use child table alias in SELECT.

10. **QueryRunner propagation**: `loadCtiChildEagerRelations` was not passing the current queryRunner to follow-up queries, breaking transaction consistency.

11. **CTI insert reload=false**: When `reload=false`, ancestor inserts skipped `updateEntity`, preventing the generated PK from being available for child inserts.

#### Performance optimizations

1. **Cached CTI metadata getters**: `isCtiChild`, `isCtiParent`, `ctiAncestorChain`, and `tableColumns` getters on `EntityMetadata` are now cached via `buildCtiCaches()`, called after metadata finalization. These getters previously walked the parent chain on every access.

2. **O(1) inherited column membership**: Replaced `inheritedColumns.includes(column)` (O(n) array scan) with `inheritedColumnsSet.has(column)` (O(1) Set lookup) across `QueryBuilder`, `SelectQueryBuilder`, `UpdateQueryBuilder`, and `InsertQueryBuilder` (7 call sites).

3. **Discriminator value Map**: Pre-built `Map<discriminatorValue, EntityMetadata>` in `RawSqlResultsToEntityTransformer.transform()` for O(1) child metadata lookups instead of nested `.find()` calls per entity row.

#### How it was verified

- **275 table-inheritance tests passing** (257 CTI + 18 STI) across 34 test suites covering:
  - Basic CRUD, multi-level hierarchies, eager/lazy loading, eager scoping
  - Polymorphic queries, query builder operations, raw data integrity, edge cases
  - Relations (OneToMany), inherited parent relations
  - Cross-child FK references, self-referential relations
  - UUID primary keys, abstract intermediate base classes, custom table names
  - Column name collisions, discriminator collision behavior
  - Root query optimization (discriminator hydration + no child JOINs — 10 tests)
  - Base entity chain matching production patterns (16 tests)
  - Transaction saves with rollback, delete cascade, version/updateDate tracking
  - Generated increment columns on child tables, update routing (parent vs child)
  - STI basic-functionality, relations (OneToMany, ManyToMany), eager scoping
- All existing test suites (query-builder, repository, find-options, relations) pass with zero regressions
- Full test suite: 1783 passing, 6 failing (all pre-existing: pgvector/PostGIS not installed)
- STI behavior unchanged — existing STI tests pass without modification

#### Issues addressed

Fixes #8629

Also addresses concerns raised in:
- #11730 — Transitive eager relations with TableInheritance
- #11703 — Embedded columns nullable in STI (CTI avoids this by giving each child its own table)
- #10563 — Default values ignored for inherited columns in multi-table inheritance

#### Supersedes

- #12043 — STI eager relation scoping is incorporated into this PR

### Pull-Request Checklist

- [x] This pull request links relevant issues as `Fixes #8629`
- [x] There are new or updated tests validating the change (275 table-inheritance tests)
- [x] Documentation has been updated to reflect this change (`docs/docs/entity/3-entity-inheritance.md`)

### DIFF
--- a/src/metadata-builder/EntityMetadataBuilder.ts
+++ b/src/metadata-builder/EntityMetadataBuilder.ts
@@ -438,6 +438,12 @@ export class EntityMetadataBuilder {
             })
         })
 
+        // Build CTI caches as the very last step, after all
+        // computeEntityMetadataStep2() calls are complete.
+        entityMetadatas.forEach((entityMetadata) => {
+            entityMetadata.buildCtiCaches()
+        })
+
         return entityMetadatas
     }
 
--- a/src/metadata/EntityMetadata.ts
+++ b/src/metadata/EntityMetadata.ts
@@ -85,6 +85,13 @@ export class EntityMetadata {
      */
     inheritedRelations: RelationMetadata[] = []
 
+    // CTI cache fields — populated by buildCtiCaches() after metadata is finalized
+    private _isCtiChild: boolean | undefined
+    private _isCtiParent: boolean | undefined
+    private _ctiAncestorChain: EntityMetadata[] | undefined
+    private _tableColumns: ColumnMetadata[] | undefined
+    private _inheritedColumnsSet: Set<ColumnMetadata> | undefined
+
     /**
      * Table type. Tables can be closure, junction, etc.
      */
@@ -582,12 +589,16 @@ export class EntityMetadata {
      * For multi-level CTI (A → B → C), walks up the parent chain to find the root's pattern.
      */
     get isCtiChild(): boolean {
+        if (this._isCtiChild !== undefined) return this._isCtiChild
+        return this._computeIsCtiChild()
+    }
+
+    private _computeIsCtiChild(): boolean {
         if (this.tableType !== "entity-child") return false
         let ancestor = this.parentEntityMetadata
         while (ancestor) {
             if (ancestor.inheritancePattern === "CTI") return true
             if (ancestor.inheritancePattern === "STI") return false
-            // Mid-level entity-child without inheritancePattern — keep walking up
             ancestor = ancestor.parentEntityMetadata
         }
         return false
@@ -613,9 +624,13 @@ export class EntityMetadata {
      * that are themselves CTI children with their own CTI children.
      */
     get isCtiParent(): boolean {
+        if (this._isCtiParent !== undefined) return this._isCtiParent
+        return this._computeIsCtiParent()
+    }
+
+    private _computeIsCtiParent(): boolean {
         if (this.childEntityMetadatas.length === 0) return false
         if (this.inheritancePattern === "CTI") return true
-        // Mid-level: is itself a CTI child and has children
         if (this.isCtiChild) return true
         return false
     }
@@ -627,6 +642,11 @@ export class EntityMetadata {
      * Returns empty array for non-CTI children.
      */
     get ctiAncestorChain(): EntityMetadata[] {
+        if (this._ctiAncestorChain !== undefined) return this._ctiAncestorChain
+        return this._computeCtiAncestorChain()
+    }
+
+    private _computeCtiAncestorChain(): EntityMetadata[] {
         if (!this.isCtiChild) return []
         const chain: EntityMetadata[] = []
         let ancestor = this.parentEntityMetadata
@@ -635,7 +655,7 @@ export class EntityMetadata {
             if (ancestor.isCtiChild) {
                 ancestor = ancestor.parentEntityMetadata
             } else {
-                break // Reached the root
+                break
             }
         }
         return chain
@@ -647,6 +667,11 @@ export class EntityMetadata {
      * For all other entities this is the same as `columns`.
      */
     get tableColumns(): ColumnMetadata[] {
+        if (this._tableColumns !== undefined) return this._tableColumns
+        return this._computeTableColumns()
+    }
+
+    private _computeTableColumns(): ColumnMetadata[] {
         if (this.isCtiChild && this.inheritedColumns.length > 0) {
             return this.columns.filter(
                 (c) => !this.inheritedColumns.includes(c),
@@ -655,9 +680,30 @@ export class EntityMetadata {
         return this.columns
     }
 
+    /**
+     * Returns the set of inherited columns for O(1) membership checks.
+     */
+    get inheritedColumnsSet(): Set<ColumnMetadata> {
+        if (this._inheritedColumnsSet !== undefined)
+            return this._inheritedColumnsSet
+        return new Set(this.inheritedColumns)
+    }
+
     // -------------------------------------------------------------------------
     // Public Methods
     // -------------------------------------------------------------------------
+
+    /**
+     * Populates CTI caches. Must be called after all metadata is fully built
+     * (after all computeEntityMetadataStep2() calls).
+     */
+    buildCtiCaches(): void {
+        this._isCtiChild = this._computeIsCtiChild()
+        this._isCtiParent = this._computeIsCtiParent()
+        this._ctiAncestorChain = this._computeCtiAncestorChain()
+        this._tableColumns = this._computeTableColumns()
+        this._inheritedColumnsSet = new Set(this.inheritedColumns)
+    }
 
     /**
      * Creates a new entity.
--- a/src/query-builder/InsertQueryBuilder.ts
+++ b/src/query-builder/InsertQueryBuilder.ts
@@ -787,7 +787,7 @@ export class InsertQueryBuilder<
                 // For CTI children, skip inherited columns (they belong to the parent table)
                 if (
                     metadata.isCtiChild &&
-                    metadata.inheritedColumns.includes(column)
+                    metadata.inheritedColumnsSet.has(column)
                 ) {
                     return false
                 }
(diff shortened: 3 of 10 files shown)


antst:
### Response to Code Suggestions

**1. Load all relation types eagerly (Score 9)**
Not a bug. Despite the misleading name, `loadManyToManyRelationIdsAndGroup()` is fully polymorphic — it handles ManyToOne, OneToOne, OneToMany, and ManyToMany internally by dispatching to `relationLoader.load()`. Test coverage confirms this: `eager-loading.ts` tests OneToOne (User.profile), OneToMany (User.posts), and cross-child eager relations via parent repo — all passing.

**2. Always enable inherited column routing (Score 8)**
Incorrect. When `aliasNamePrefixingEnabled` is false, column references have no alias prefix — there's no prefix to route to. Adding CTI routing in that case would produce broken SQL. The guard is intentional.

**3. Harden options detection (Score 6)**
Not needed. The existing `"discriminatorValue" in ...` / `"tableName" in ...` checks already exclude arrays and functions — neither will have those keys.

**4. Guard custom tableName (Score 4)**
Cosmetic. TypeORM's metadata builder already handles `name: undefined` by falling back to the class name. No behavioral difference.

**5. Improve CTI deletion logic (Score 1)**
Functionally identical to existing code — just adds a variable and comments.

**6. Support soft-delete in CTI joins (Score 8)**
Incorrect. CTI ancestor rows share the same PK as child rows — an ancestor row is always present if the child row exists. Soft-delete sets a timestamp column; it doesn't remove rows. Filtering happens via WHERE clause, not JOIN type. `INNER JOIN` is correct for CTI ancestor joins.

**7. Correctly set entityMetadata on cloned relations (Score 8)**
Not applicable to CTI. The `Object.create(parentRelation)` clone path is guarded by `!entityMetadata.isCtiChild` (EntityMetadataBuilder.ts:795-798) — it only applies to STI children. For CTI, parent relations are added as direct references to `inheritedRelations`, and `relation.entityMetadata` correctly points to the parent table where FK columns physically reside. The `createPerTableValueMaps()` routing depends on this being the parent's metadata.

**8. Fix CTI column aliasing logic (Score 8)**
Incorrect — switching to `metadata.ctiAncestorChain` would break multi-level hierarchies. For `Actor.find()` returning a User (Actor → Contributor → User): column `reputation` (owned by Contributor) currently gets `isMainColumn=false` → uses `Actor__cti_child_User` alias → correct. With the suggested change, Contributor would be found in User's ancestor chain → `isMainColumn=true` → reads from Actor alias → NULL/wrong value. The column-name-collision test validates this exact scenario.

**9. Ensure full entity is stored in updatedRelationMaps (Score 7)**
Not a bug. `createPerTableValueMaps()` pushes `relationId` to `updatedRelationMaps` — identical to `createValueSetAndPopChangeMap()`. The consumer (`getEntityValue`) works correctly with both ID maps and full entities.

antst:
Superseded by #12052 — same changes squashed into a single clean commit for easier review.

Issue #7453, opened 2021-03-08
Support RelationOption `orphanedRowAction` also for `@OneToOne` Relations 


tobiaskraus:
### The Problem

I have two Entities connected with OneToOne relation and one is the owner:

```ts
@Entity()
export class Main {
  @OneToOne(() => side, { cascade: true })
  @JoinColumn()
  side: Side;

  // ... some columns
}

@Entity()
export class Side {
  @PrimaryGeneratedColumn()
  id: number;

  // ... some columns
}
```

I want that TypeORM cleans my Side unused entities, after I update a `Main` row via `save()` and providing a new `Side` row (with cascade so a new `Side` row gets created).

```ts
getRepository(Main).save({
  id: 1,
  side: {
    someData: 'lorem ipsum'
  }
});
```

### The Solution

There is a new feature available since TypeORM version 0.2.30:

> * Orphaned row action (#7105) (efc2837)

```ts
@Entity()
export class Main {
  @OneToOne(() => side, { cascade: true, orphanedRowAction: "delete"  })
  @JoinColumn()
  side: Side;

  // ... some columns
}
```

But it seems only to work currently on `ManyToOne` relations:

(as seen here in this [test file](https://github.com/typeorm/typeorm/blob/efc283769ed972d022980e681e294d695087a807/test/functional/persistence/delete-orphans/entity/Post.ts) in TypeORM source code):

```
@ManyToOne(() => Category, category => category.posts, { orphanedRowAction: "delete" })
@JoinColumn({ name: "categoryId" })
category: Category;
```

### Relevant Database Driver(s)

- [x] `mysql`

### Are you willing to resolve this issue by submitting a Pull Request?

maybe, but the PR of the orphanRowAction feature seemed time consuming with the test files and so on. And I don't think I have the time for it.

tomLadder:
Have the same problem. Not working for OneToMany relations

Yokutto:
Maybe I'm mixing things up, but does this somehow give an alternative to the issue ["onDelete: 'CASCADE'"](https://github.com/typeorm/typeorm/issues/6478) not working properly?

castrillon7:
Yep, it does not work on `@OneToOne`.
If you set the FK field with either `null` or `undefined` the child row can still be found.

yleflour:
@Yokutto, I got the same issue. If you check the mentioned issue, onDelete works properly BUT:

[...]

PR #12051, opened 2026-02-26
fix: honor orphanedRowAction on @OneToMany when explicitly set

### LINKED ISSUE #12033: orphanedRowAction on @OneToMany is silently ignored — should be supported like JPA's orphanRemoval
## Problem

`orphanedRowAction` is defined in the shared `RelationOptions` interface, so TypeScript allows setting it on `@OneToMany` without any type error. However, the runtime completely ignores it — only `@ManyToOne` is effective.

This creates a **silent failure**: developers set the option, get no error, but the feature doesn't work.

```typescript
// ✗ Silently ignored — no type error, no runtime error, just doesn't work
@OneToMany(() => Image, (image) => image.article, {
    cascade: true,
    orphanedRowAction: "delete", // ← ignored at runtime
})
images: Image[]
```

```typescript
// ✓ Only this works
@ManyToOne(() => Article, (article) => article.images, {
    orphanedRowAction: "delete", // ← must be here
})
article: Article
```

## Root Cause

`OneToManySubjectBuilder` always reads from `relation.inverseRelation.orphanedRowAction` (the `@ManyToOne` side). There is no code anywhere in TypeORM that reads `orphanedRowAction` from `@OneToMany`.

**`src/persistence/subject-builder/OneToManySubjectBuilder.ts:181-213`**:
```typescript
// relation = @OneToMany (iterating subject.metadata.oneToManyRelations)
// relation.inverseRelation = the @ManyToOne side
if (relation.inverseRelation?.orphanedRowAction !== "disable") {
    // ...
    if (relation.inverseRelation.orphanedRowAction === "nullify") { ... }
    else if (relation.inverseRelation.orphanedRowAction === "delete") { ... }
    else if (relation.inverseRelation.orphanedRowAction === "soft-delete") { ... }
}
```

## Why This Is Confusing

1. **JPA's `orphanRemoval` is set on `@OneToMany`** — the parent side declares what happens to orphaned children. This is the natural mental model.

2. **TypeORM's own `cascade` is set on `@OneToMany`** — developers naturally expect `orphanedRowAction` to go alongside `cascade` on the same decorator.

3. **The type system doesn't prevent misuse** — `orphanedRowAction` is in the shared `RelationOptions`, so it's available on all relation decorators without restriction.

## Proposed Solution (Discussion)

At minimum one of:

1. **Support reading from `@OneToMany`**: In `OneToManySubjectBuilder`, also check `relation.orphanedRowAction` as a fallback when `relation.inverseRelation.orphanedRowAction` is not set. This aligns with JPA's design and TypeORM's own `cascade` pattern.

2. **Narrow the type**: If supporting `@OneToMany` is not desired, remove `orphanedRowAction` from the shared `RelationOptions` and make it a `@ManyToOne`-specific option to prevent silent failures.

Option 1 seems preferable as it matches both JPA conventions and developer expectations.

## Related Issues

- #7453 — `orphanedRowAction` not working on `@OneToOne` (same category of issue: option is accepted by types but ignored at runtime)

LMazilu:
## Summary
  - honor orphanedRowAction when explicitly configured on @OneToMany
  - preserve backward compatibility by falling back to inverse-side behavior
  - add regression coverage in orphanage delete fixture

  ## Why
  Fixes #12033 (silent runtime ignore of orphanedRowAction on @OneToMany).

  ## Validation
  - pnpm compile
  - pnpm exec mocha --no-config --file build/compiled/test/utils/test-setup.js build/compiled/test/functional/persistence/
  orphanage/delete/delete-orphans.test.js


LMazilu:
Thanks for the suggestion.

For valid TypeORM mappings, `@OneToMany` requires a corresponding `@ManyToOne` [(official docs here)](https://typeorm.io/docs/relations/many-to-one-one-to-many-relations/), so `inverseRelation` is expected to exist here.

I’d avoid guarding with `&& relation.inverseRelation` because it can silently skip orphan processing when metadata is malformed. That would hide a configuration error.

If we want extra safety, I’d prefer fail-fast with an explicit invariant check:

```ts
if (!relation.inverseRelation) {
    throw new TypeORMError(
        `OneToMany relation "${relation.entityMetadata.name}.${relation.propertyPath}" is missing its inverse ManyToOne
relation metadata.`,
    )
}
```
Trade-offs

- Pros: explicit failure, easier debugging, no silent behavior changes.
- Cons: stricter behavior for malformed metadata (surfaces earlier as runtime error).

pkuczynski (commented):
Good direction — `@OneToMany` should honor `orphanedRowAction`. A few issues to address:

1. The non-nullable FK safety net was removed — the previous code auto-deleted orphaned rows when the FK column was `NOT NULL` instead of attempting to nullify. Without this guard, non-nullable FKs will hit constraint violations at runtime.

2. `hasOwnProperty` returns `true` when `EntitySchemaTransformer` emits `orphanedRowAction: undefined` — this bypasses the fallback to the inverse relation for all EntitySchema users.

3. The `"nullify"` path uses `relation.inverseRelation!` but can be reached when `inverseRelation` is absent — potential NPE.

4. The `"disable"` early-exit guard still only checks the inverse side, so `orphanedRowAction: "disable"` on `@OneToMany` is silently ignored.

Will try to fix these to speed up merge of this PR.

gioboa (approved):

pkuczynski (commented):
After researching how other ORMs handle this (JPA/Hibernate, Doctrine, MikroORM all use `orphanRemoval` exclusively on `@OneToMany` and `@OneToOne`), and reviewing the original issue #12033 by @LMazilu — the conclusion is clear: `orphanedRowAction` on `@ManyToOne` was a design mistake from the beginning.

The parent entity should control what happens to orphaned children, not the child. `@ManyToOne` doesn't even participate in orphan detection — `OneToManySubjectBuilder` always runs from the `@OneToMany` side and was reaching across to the inverse to read the option. This created unnecessary complexity (`isOrphanedRowActionSet`, fallback logic) and confused users who set it on `@ManyToOne` expecting it to work directly.

For 1.0, `orphanedRowAction` will only be supported on `@OneToMany` (and `@OneToOne` in a follow-up — see #7453). The `@ManyToOne` and `@ManyToMany` decorators will no longer accept it at the TypeScript level. A codemod will flag any existing usage for migration.

Working on getting this done now.

michaelbromley:

```
git fetch origin
```

Issue #11, opened 2016-06-22
Empty d.ts files are causing consumer errors due to @internal classes not being emitted


zakhenry:
See https://github.com/Microsoft/TypeScript/issues/6022 - this may have actually been fixed, but the npm package needs to be recompiled with a later version that includes that fix? I get the following errors (using `typescript@next)
```
Building ts for [server]
/Users/zak/ubiquits/ubiquits/node_modules/typeorm/query-builder/QueryBuilder.d.ts(1,23): error TS2306: File '/Users/zak/ubiquits/ubiquits/node_modules/typeorm/query-builder/alias/Alias.d.ts' is not a module.
/Users/zak/ubiquits/ubiquits/node_modules/typeorm/repository/Repository.d.ts(4,51): error TS2306: File '/Users/zak/ubiquits/ubiquits/node_modules/typeorm/query-builder/transformer/PlainObjectToNewEntityTransformer.d.ts' is not a module.
/Users/zak/ubiquits/ubiquits/node_modules/typeorm/repository/Repository.d.ts(5,56): error TS2306: File '/Users/zak/ubiquits/ubiquits/node_modules/typeorm/query-builder/transformer/PlainObjectToDatabaseEntityTransformer.d.ts' is not a module.
/Users/zak/ubiquits/ubiquits/node_modules/typeorm/repository/Repository.d.ts(6,47): error TS2306: File '/Users/zak/ubiquits/ubiquits/node_modules/typeorm/persistment/EntityPersistOperationsBuilder.d.ts' is not a module.
/Users/zak/ubiquits/ubiquits/node_modules/typeorm/repository/Repository.d.ts(7,42): error TS2306: File '/Users/zak/ubiquits/ubiquits/node_modules/typeorm/persistment/PersistOperationExecutor.d.ts' is not a module.
[17:55:21] TypeScript: 5 semantic errors
```

On inspection, those `d.ts` files reporting errors are empty, and the source files have only classes with `@internal` jsdoc.

pleerock:
yeap Im aware of these errors. will fix them today

zakhenry:
Cool, if it helps any, [RxJS implemented a workaround](https://github.com/ReactiveX/rxjs/issues/1010#issuecomment-163430493) to suppress the error. Though I'm unsure if it is still actually an error with latest TS

pleerock:
fixed in 0.0.2-alpha.43 f5999cd..9ba0eb9

zakhenry:
[...]

Issue #10, opened 2016-06-18
A good companion to Ubiquits?


zakhenry:
Hey, just came across Typeorm and was wondering if it would be a good fit for the TS framework I'm building over at http://github.com/ubiquits/ubiquits or http://ubiquits.com? I'm still in early stage development - intending to release developer preview next week and am currently building on Sequelize, but it looks like Typeorm might be an easier implementation.

Also regarding your issue on the need to build a documentation site, you might want to consider the infrastructure I've built up to document Ubiquits

pleerock:
Hello, it totally should be. Any good framework must came with a good ORM. And typeorm can play a good role in your stack. And yeah, its also in active development, like your framework. I was also working on a [framework](https://github.com/pleerock/microframework), but its not a fullstack (and will not be, initially i was planning to do it for rest microservices, and I'm using it internally on some of my projects). Any framework should compose of components, like DI, routing, validation, etc. If you are interesting you can also think about using [typedi](https://github.com/pleerock/typedi) and [routing-controllers](https://github.com/pleerock/routing-controllers). There is also [validation component](https://github.com/pleerock/class-validator). Take a look on my other open source projects in my repositories, maybe you'll find something more that you can use in your framework.

What about documentation site  - I'll think about it, but I would like to build it on something very simple and I want to write all documentation in markdown (maybe jekyll, don't know yet)

pleerock:
There is also [typeodm](https://github.com/pleerock/typeodm) I was working on, but right now my focus only on ORM - after I finish all my todos and it will stabilize, so I can publish it to public, I'll switch to typeodm. There is no documentation there, however take a look on [samples](https://github.com/pleerock/typeodm/tree/master/sample) and you'll see that its a good db framework too

zakhenry:
Awesome, I'll see about migrating over from Sequelize. I see the note in the readme about Postgres not being quite finished, what is missing? I want to use Postgres for the uuid & json column types, and schemas. I'm happy to fork and add this functionality if it's missing.

zakhenry:
Regarding the framework stuff, yea I really like what you are doing with those components - I'm planning to use the class validation one already. With DI, I've already built that, leveraging the `@angular/core/di` components so the service provider and testing framework all work the same as angular2.

[...]

PR #12052, opened 2026-02-26
feat: implement Class Table Inheritance (CTI / Joined Table Inheritance)

### LINKED ISSUE #8629: Major Feature request: Join Table Inheritance pattern

antst:
### Description of change

This PR implements **Class Table Inheritance** (CTI), also known as Joined Table Inheritance — a long-requested inheritance mapping pattern (see #8629) where each class in the hierarchy gets its own database table, joined by primary key.

#### What it does

With `pattern: "CTI"` in the `@TableInheritance` decorator, TypeORM now maps an inheritance hierarchy to **multiple tables joined by primary key**:

```typescript
@Entity()
@TableInheritance({ pattern: "CTI", column: { type: "varchar", name: "type" } })
export class Actor {
    @PrimaryGeneratedColumn()
    id: number

    @Column()
    name: string
}

@ChildEntity()
export class User extends Actor {
    @Column()
    email: string
}

@ChildEntity()
export class Organization extends Actor {
    @Column()
    industry: string
}
```

This creates **three tables**: `actor` (id, name, type), `user` (id, email), `organization` (id, industry) — where child PKs are foreign keys to the parent.

#### Why this is needed

STI forces all child-specific columns into one table as nullable columns, which degrades schema clarity and data integrity as hierarchies grow. CTI provides clean separation: each table only has its own columns, child-specific columns can have proper NOT NULL constraints, and relations can be placed at any level of the hierarchy.

#### Key capabilities

- **Schema sync**: Child tables are created/dropped alongside parent tables with proper FK constraints
- **Query building**: Child queries auto-INNER JOIN to parent table; parent queries return only parent-table data (no child table JOINs)
- **Persistence**: INSERT/UPDATE/DELETE cascade across parent and child tables (PK propagation)
- **Polymorphic queries**: `actorRepository.find()` instantiates correct child types via discriminator, with parent-table columns populated (child-specific columns require querying child entity directly)
- **Discriminator hydration**: The CTI discriminator `type` property is assigned to loaded entities, available via `(actor as any).type`. STI discriminator behavior is unchanged — virtual discriminators are not hydrated onto STI entities.
- **Multi-level CTI**: Supports hierarchies deeper than 2 levels (e.g., Actor → Contributor → User)
- **Eager relation scoping**: Child-specific eager relations are scoped to the correct type
- **Lazy relations**: Full support for Promise-based lazy loading on CTI entities
- **Relations at any level**: OneToOne, OneToMany, ManyToOne, ManyToMany on parent or child entities
- **Inherited parent relations**: Child entities can load relations defined on the parent via `find({ relations })` and QueryBuilder `leftJoinAndSelect`
- **Cross-child FK references**: CTI children can reference each other (e.g., `Organization.ceo → User`)
- **Self-referential relations**: CTI children support self-referential relations (e.g., `User.manager → User`)
- **UUID primary keys**: Full support for `@PrimaryGeneratedColumn("uuid")` with CTI
- **Abstract intermediate classes**: Supports abstract classes with `@Column` decorators between parent and child (columns inherited by concrete children)
- **Custom child table names**: `@ChildEntity({ tableName: "app_users" })` allows explicit table names for CTI children

#### Design: parent entity queries

Following SQLAlchemy's approach, **CTI root entity queries do NOT LEFT JOIN child tables**. The parent table is a standalone entity — it has its own columns and relations. When you query `actorRepository.find()`, you get:

- Correct child class instances (via discriminator)
- Parent-table columns populated (id, name, type, parent-level relations)
- Child-specific columns are `undefined` (email, industry, etc.)

To access child-specific data, query the child entity directly:

```typescript
// Parent query — lightweight, no child JOINs:
const actors = await actorRepository.find()
actors[0].name       // ← populated (parent column)
(actors[0] as User).email  // ← undefined

// Child query — INNER JOINs parent, full data:
const user = await userRepository.findOneBy({ id: actors[0].id })
user.email           // ← populated
```

This avoids the performance trap of LEFT JOINing every child table on every parent query (which grows linearly with hierarchy size), and keeps the parent entity semantically clean.

#### Bug fixes included

1. **RETURNING clause filtering**: CTI INSERT/UPDATE RETURNING clauses included inherited columns from parent tables, causing PostgreSQL errors. Fix: use `tableColumns` instead of `columns` in `getInsertionReturningColumns()`, `getUpdationReturningColumns()`, `getSoftDeletionReturningColumns()`.

2. **Version/updateDate double-increment**: CTI child UPDATE auto-incremented version/updateDate on both child and parent tables. Fix: skip auto-increment SET clauses when building child UPDATE (parent handles it).

3. **JOIN ordering**: CTI parent table JOINs weren't interleaved correctly with child table JOINs, causing "missing FROM-clause entry" errors for nested CTI relations. Fix: interleave CTI parent JOINs immediately after child table JOINs.

4. **Abstract base class column hydration**: Columns inherited from abstract base classes (id, createdDate, etc.) weren't hydrated onto CTI root entities because `column.target` pointed to the abstract class. Fix: property-based matching (`databaseName` + `propertyPath`) instead of reference-based check.

5. **Discriminator value not hydrated (CTI only)**: The virtual discriminator column was skipped during entity property assignment because `isVirtual` columns were unconditionally excluded. Fix: allow discriminator columns through the `isVirtual` check, **scoped to CTI only** — STI virtual discriminators remain excluded (matching existing STI behavior where auto-created discriminators are not hydrated onto entities).

6. **Root entity queries performance**: Parent entity queries unnecessarily LEFT JOINed all child tables, SELECTed child columns, and loaded child-specific eager relations — O(N) JOINs for N children. Fix: removed all three behaviors, following SQLAlchemy's approach (parent table is standalone).

7. **OneToMany/OneToOneNotOwner on CTI children**: Fixed "missing FROM-clause entry" error when loading OneToMany relations targeting CTI child entities. The STI discriminator condition was incorrectly applied to CTI entities.

8. **Nullable LEFT JOIN on CTI children**: Fixed entities with null nullable relations disappearing from results. When a relation was LEFT JOINed and the related CTI entity was null, the INNER JOIN to its parent table eliminated the entire row.

9. **Column alias collisions**: When two CTI child types define columns with the same `databaseName` (e.g., both User and Organization have "status"), SQL aliases collided. Fix: child-specific columns use child table alias in SELECT.

10. **QueryRunner propagation**: `loadCtiChildEagerRelations` was not passing the current queryRunner to follow-up queries, breaking transaction consistency.

11. **CTI insert reload=false**: When `reload=false`, ancestor inserts skipped `updateEntity`, preventing the generated PK from being available for child inserts.

#### Performance optimizations

1. **Cached CTI metadata getters**: `isCtiChild`, `isCtiParent`, `ctiAncestorChain`, and `tableColumns` getters on `EntityMetadata` are now cached via `buildCtiCaches()`, called after metadata finalization. These getters previously walked the parent chain on every access.

2. **O(1) inherited column membership**: Replaced `inheritedColumns.includes(column)` (O(n) array scan) with `inheritedColumnsSet.has(column)` (O(1) Set lookup) across `QueryBuilder`, `SelectQueryBuilder`, `UpdateQueryBuilder`, and `InsertQueryBuilder` (7 call sites).

3. **Discriminator value Map**: Pre-built `Map<discriminatorValue, EntityMetadata>` in `RawSqlResultsToEntityTransformer.transform()` for O(1) child metadata lookups instead of nested `.find()` calls per entity row.

#### How it was verified

- **275 table-inheritance tests passing** (257 CTI + 18 STI) across 34 test suites covering:
  - Basic CRUD, multi-level hierarchies, eager/lazy loading, eager scoping
  - Polymorphic queries, query builder operations, raw data integrity, edge cases
  - Relations (OneToMany), inherited parent relations
  - Cross-child FK references, self-referential relations
  - UUID primary keys, abstract intermediate base classes, custom table names
  - Column name collisions, discriminator collision behavior
  - Root query optimization (discriminator hydration + no child JOINs — 10 tests)
  - Base entity chain matching production patterns (16 tests)
  - Transaction saves with rollback, delete cascade, version/updateDate tracking
  - Generated increment columns on child tables, update routing (parent vs child)
  - STI basic-functionality, relations (OneToMany, ManyToMany), eager scoping
- All existing test suites (query-builder, repository, find-options, relations) pass with zero regressions
- Full test suite: 1783 passing, 6 failing (all pre-existing: pgvector/PostGIS not installed)
- STI behavior unchanged — existing STI tests pass without modification

#### Issues addressed

Fixes #8629

Also addresses concerns raised in:
- #11730 — Transitive eager relations with TableInheritance
- #11703 — Embedded columns nullable in STI (CTI avoids this by giving each child its own table)
- #10563 — Default values ignored for inherited columns in multi-table inheritance

#### Supersedes

- #12050 — Previous version of this PR (squashed into single commit for clean history)
- #12043 — STI eager relation scoping is incorporated into this PR

### Pull-Request Checklist

- [x] This pull request links relevant issues as `Fixes #8629`
- [x] There are new or updated tests validating the change (275 table-inheritance tests)
- [x] Documentation has been updated to reflect this change (`docs/docs/entity/3-entity-inheritance.md`)


antst:
### Re: Qodo Code Review

**1. `as any` in CTI test (Rule violation)**

The discriminator `type` property is auto-generated by TypeORM for CTI and is not declared on the entity class (unlike STI where users explicitly add `@Column() type: string`). Using `as any` is the only way to access it in test assertions — this is expected, not a type safety violation.

**2. Inherited relation aliasing broken (Bug claim)**

False positive. Verified by adding two new tests (ae64e06) that query a child entity (`User`) with relations defined on the parent (`Actor`):
- `findOne(User, { relations: { profile: true } })` — non-eager inherited relation
- `findOne(User)` — eager inherited relation (`authorization`)

Both pass. The generated SQL correctly aliases inherited join columns to the ancestor table:
```sql
LEFT JOIN "profile" ON "profile"."id" = "user__cti_parent"."profileId"
```
The `inheritedPropertyToAliasPrefix` map works at the column level (`profileId`), which is where the final replacement happens — relation paths (`profile.id`) are resolved to column names first, then the column-level alias rerouting kicks in.

**3. CTI version/updateDate on intermediate ancestors (Bug claim)**

Not a real bug. `@VersionColumn()` and `@UpdateDateColumn()` are always defined on the root entity in TypeORM's CTI model — there is no supported pattern for placing them on intermediate entities. Our code correctly targets `rootMeta` for version/updateDate bumps, which is validated by the `version-and-update-date` test suite (9 tests).

antst:
### Re: Qodo Code Review (second pass)

**1 & 2. CTI embedded columns misowned / CTI embedded FK duplication (Bug claims, New)**

Valid theoretical concern about the CTI + embeddeds combination. If a CTI parent uses `@Column(() => EmbeddedClass)`, the embedded columns could be duplicated on both parent and child tables because `createEmbeddedsRecursively` creates `ColumnMetadata` with the child's `entityMetadata`, and `inheritedColumns` only collects from `ancestor.ownColumns`.

However, this is a **pre-existing limitation** of the metadata builder, not a regression from this PR:
- No CTI test entities (ours or existing) use embeddeds
- CTI was non-functional before this PR, so the combination was never tested
- The `entity-child` tableType handling in `createEmbeddedsRecursively` predates our changes

Worth addressing in a follow-up if CTI + embeddeds becomes a needed pattern.

**3. Noisy separator comments (Rule violation, New)**

The `// =========` separators organize test files with 10+ test cases into logical groups. This is a readability choice, not noise.

**4, 5, 6. Repeats from previous review**

Already addressed in [the previous comment](https://github.com/typeorm/typeorm/pull/12052#issuecomment-3970159593).

polibb (approved):

polibb (commented):
amazing

antst:
### Re: Qodo Code Review (third pass)

Items #4-9, #11-13 are repeats from previous reviews — already addressed in earlier comments.

**#1. `nameID`/`accountID` nonstandard casing (Rule violation, New)**

Fixed in bdf2b7c. Renamed to `nameId`/`accountId`/`resourceId` across 47 test files to match TypeORM's camelCase convention.

**#2. CTI insert API broken & #3. CTI primitive DML broken (Bug claims, New)**

Valid concern. Primitive APIs (`insert`/`update`/`delete`/`upsert`/`softDelete`/`restore`) operate on a single table and can't handle CTI's multi-table layout.

Rather than implementing full multi-table support in these APIs — which would be a significant change touching `EntityManager`, `Repository`, `InsertQueryBuilder`, `SoftDeleteQueryBuilder`, and potentially requiring implicit transactions — we added a safety guard in 984254e: these methods now throw a clear `TypeORMError` when called on CTI child entities, directing users to use `save()`/`remove()` instead.

This approach is consistent with how TypeORM treats these APIs for STI (e.g., `insert()` doesn't auto-set the STI discriminator either) — they are documented as primitive single-table operations that bypass the entity lifecycle. The full entity lifecycle path via `SubjectExecutor` (`save()`/`remove()`) correctly handles all CTI multi-table operations including cascading inserts, split updates, and cascading deletes.

Full multi-table support for primitive APIs can be added in a follow-up PR.

**#10. CTI inheritance FKs missing (Bug, New)**

Valid — fixed in 2fc5f52. Schema sync now creates `child(PK) → parent(PK)` foreign key constraints with `ON DELETE CASCADE` for all CTI child tables, enforcing referential integrity at the database level.

antst:
### Re: Qodo Code Reviews (passes 4-9)

Addressing the 8 new items (⭐ New) across the last 6 review comments. Everything else is repeats from previous reviews.

**#4. increment/decrement/clear bypass guard (Bug, New)**

Valid — these three methods were missing the `rejectCtiChild` guard. Fixed in d6ffec8.

**#1/#3. Indices/constraints on missing CTI child columns (Bug, New)**

Valid latent concern. If a CTI parent defines `@Index` or `@Unique` on a column, the child table metadata could include that constraint despite not physically having the column. However, no CTI entities in our test suite (or typical usage) define explicit `@Index`/`@Unique` on parent entities — the discriminator index is the only one, and it's handled separately. Worth addressing in a follow-up if CTI + explicit parent indices becomes a needed pattern.

**#2. CTI UpdateQueryBuilder failure (Bug, New)**

False positive. `UpdateQueryBuilder` already filters inherited columns via `inheritedColumnsSet.has()` (our optimization #2). CTI child updates go through `SubjectExecutor.executeCtiUpdate()` which routes columns to the correct ancestor table. The primitive `update()` API is blocked by `rejectCtiChild`.

**#5. CTI PK join mismatch hidden (Bug, New)**

False positive. The `propertyName` vs `propertyPath` distinction only matters for embedded composite PKs — a hypothetical CTI + embedded combination that's not tested or supported yet (same category as the embedded concerns previously discussed). Our CTI entities use simple single-column PKs where `propertyName === propertyPath`.

**#6/#7. `any` in ChildEntity / loadCtiChildEagerRelations (Rule violations, New)**

Not actionable. `discriminatorValue` is typed `any` because discriminator values can be strings, numbers, or other types depending on the column definition — this matches the existing `discriminatorValue` typing throughout TypeORM. The `any[]` in the eager loader is consistent with TypeORM's internal entity handling patterns throughout the query builder (entities are untyped at runtime).

**#8. CTI fields stay undefined (Requirement gap, New)**

Not a gap. Issue #8629 requested Class Table Inheritance as a feature, not lazy field loading. When querying through the root `Actor` repo, child-specific columns (`email`, `industry`) are `undefined` because they live on separate tables — this is the correct and expected CTI behavior, consistent with how Hibernate, Django, and every other ORM implements CTI. Users query the child repository when they need child-specific fields.

antst:
### Re: Qodo Code Review

Addressing all 9 unique issues identified across Qodo's review passes. Note: the header counts (100 bugs, 47 rule violations) are inflated by cumulative counting across review iterations and duplicate entries within the same comment — there are 9 distinct issues, which we address below.

#### Fixed

**`nameID`/`accountID` nonstandard casing (Rule violation)** — Fixed in bdf2b7c. Renamed to `nameId`/`accountId`/`resourceId` across 47 test files.

**CTI insert API broken / CTI primitive DML broken (Bug)** — Fixed in 984254e + d6ffec8 + 7e6e6c0. Primitive single-table APIs (`insert`/`update`/`delete`/`upsert`/`softDelete`/`restore`/`increment`/`decrement`/`clear`) now throw a clear `TypeORMError` when called on CTI child entities, directing users to `save()`/`remove()`. This is consistent with how these APIs behave for STI (e.g., `insert()` doesn't auto-set the STI discriminator either) — they are documented as primitive single-table operations that bypass the entity lifecycle. The full lifecycle path via `SubjectExecutor` correctly handles all CTI multi-table operations. Full multi-table primitive API support can be added in a follow-up.

#### Not a bug

**Inherited relation aliasing broken (Bug)** — False positive. Verified by adding tests (ae64e06) that query child entities with relations defined on the parent (`findOne(User, { relations: { profile: true } })`). The generated SQL correctly routes inherited join columns to the ancestor alias:
```sql
LEFT JOIN "profile" ON "profile"."id" = "user__cti_parent"."profileId"
```
The `inheritedPropertyToAliasPrefix` map works at the column level — relation paths are resolved to column names first, then alias rerouting applies.

**`as any` in CTI test (Rule violation)** — The CTI discriminator `type` property is auto-generated by TypeORM and not declared on the entity class (unlike STI where users explicitly define it). `as any` is the only way to access it in test assertions.

**CTI child DDL constraints wrong (Bug)** — Theoretical concern about `@Index`/`@Unique` defined on parent entities being applied to child tables. No CTI entities in our test suite or typical usage define explicit indices on parent entities. The discriminator index — the only auto-created one — is handled separately. Can be addressed in a follow-up if CTI + explicit parent indices becomes a needed pattern.

**CTI column overloading wrong (Bug)** — False positive. The column-overloading test suite (10 tests) explicitly verifies: redeclared columns live on the parent table only, child tables do NOT get duplicates, parent defaults win. All tests pass.

**CTI fields stay undefined (Requirement gap)** — Not a gap. Issue #8629 requested Class Table Inheritance, not lazy field loading. Child-specific columns being `undefined` in parent-repo queries is correct CTI behavior — the same as Hibernate, Django, and every other ORM that implements CTI. Users query the child repository for child-specific fields.

#### Acknowledged limitations (follow-up)

**CTI embedded columns misowned / CTI embedded FK duplication (Bug)** — Valid theoretical concern about CTI + embedded columns combination. Embedded columns from ancestors could be duplicated on child tables because `createEmbeddedsRecursively` creates metadata with the child's entityMetadata, and `inheritedColumns` only collects from `ancestor.ownColumns`. However, this is a pre-existing limitation of the metadata builder (not introduced by this PR), and no CTI entities use embeddeds. Can be addressed when CTI + embeddeds becomes a needed pattern.

**CTI softDelete/restore broken (Bug)** — Valid concern. `SubjectExecutor`'s `softRemove()`/`recover()` uses `SoftDeleteQueryBuilder` targeting the child table, but `@DeleteDateColumn`/`@VersionColumn` may live on an ancestor table. No CTI entities in our test suite use `@DeleteDateColumn`, and fixing this requires implementing CTI-aware multi-table soft-delete logic (similar to `executeCtiUpdate`). Planned for a follow-up.

antst:
### SonarCloud triage

Triaged all 50 SonarCloud annotations:

**Duplication (15.2% > 3%)**
This is test code duplication — CTI test suites necessarily share similar entity setup/teardown/assertion patterns across different functional test files (eager-loading, lazy-loading, polymorphic-queries, root-query-optimization, etc.). SonarCloud doesn't distinguish test vs production code for this metric. Production code has no meaningful duplication.

**Reliability (D > A)**
The single non-test annotation is `EntityManager.ts` — a braceless `if (x) return` guard in `rejectCtiChild()`. This is a style preference, not a bug: the `return` exits the function unconditionally. The remaining 49 annotations are all "unnecessary type assertion" warnings in test files (e.g., `expect(loaded!.type)`) which are standard patterns for nullable assertion in Chai/Mocha tests.

Additionally, `as any` casts flagged in test code have been removed in e8e6a4c5c.

michaelbromley:

```
git fetch origin
```

PR #12057, opened 2026-02-28
feat(redis)!: remove support for legacy redis client versions

G0maa:
### Description of change

- Removing support for redis v3/v4 clients.
- Added simple unit tests.

### Pull-Request Checklist


test/unit/cache/redis-query-result-cache.test.ts
```diff
@@ -5,7 +5,7 @@ import { RedisQueryResultCache } from "../../../src/cache/RedisQueryResultCache"
 import { PlatformTools } from "../../../src/platform/PlatformTools"
 import { DataSource } from "../../../src/data-source/DataSource"
 
-describe("RedisQueryResultCache", () => {
+describe.skip("RedisQueryResultCache", () => {
```
gioboa: Is there a reason for this skip?

G0maa: This test only verifies "detectRedisVersion" which the `RedisQueryResultCache` doesn't have anymore.

I'm skipping before deleting to see if that's the only test that fails.

G0maa:
(will fix Qodo comments)

G0maa:
maybe I should add tests too, since it's likely `storeInCache` should have failed :sweat:

src/cache/RedisQueryResultCache.ts
```diff
@@ -21,18 +23,13 @@ export class RedisQueryResultCache implements QueryResultCache {
     /**
      * Connected redis client.
      */
-    protected client: any
+    protected client: RedisClientType | Redis | Cluster
```
G0maa: Is it ok to import types?

alumni: In #12044 I made it mandatory to use type imports if only the types are used. Needs to be checked if importing anything from `redis`/`ioredis` is okay.

src/cache/RedisQueryResultCache.ts
```diff
@@ -4,6 +4,8 @@ import { PlatformTools } from "../platform/PlatformTools"
 import { DataSource } from "../data-source/DataSource"
 import { QueryRunner } from "../query-runner/QueryRunner"
 import { TypeORMError } from "../error/TypeORMError"
+import Redis, { Cluster } from "ioredis"
```
alumni: This will made `redis`/`ioredis` a mandatory dependency of `typeorm` which we do not want. Type imports might work, but needs to be checked if they don't cause problems in projects using `typeorm` but not using `redis`.

src/cache/RedisQueryResultCache.ts
```diff
@@ -132,21 +111,7 @@ export class RedisQueryResultCache implements QueryResultCache {
...
-
-        // Redis 3/4 callback style
-        return new Promise<void>((ok, fail) => {
-            this.client.quit((err: any, result: any) => {
-                if (err) return fail(err)
-                ok()
-                this.client = undefined
-            })
-        })
+        await this.client.quit()
```
alumni: Proper cleanup:
```suggestion
        const client = this.client
        this.client = undefined
        await client.quit()
```

src/cache/RedisQueryResultCache.ts
alumni: I actually checked, since I'm doing something similar with mongo:

```ts
import type { ObjectId } from "mongodb";
```

would cause type errors in projects using TypeORM.

Unless: they set `skipLibCheck: true` in their TSConfig file. We do this in our project because we use `@tsconfig/nodeXX` which sets it by default, but it's not something we should force everyone to use.

G0maa: @alumni, so I should revert to `any`, I assume `monorepo` (?) will fix this issue (i.e. people will download only the adapters they need)?

src/cache/RedisQueryResultCache.ts
G0maa: removed

test/unit/cache/redis-query-result-cache.test.ts
G0maa: I've added simple unit tests.

test/functional/cache/redis-cache-integration.test.ts
G0maa, line 1: I've looked a bit in the codebase for a method to directly test the RedisQueryCacheResult but I have not found one, I've created this test suite for that purpose, although I'm unsure if it's the pattern we want to use. Wdyt? @alumni @gioboa @pkuczynski

docker-compose.yml
```diff
@@ -147,13 +147,11 @@ services:
...
       - "27017:27017"
-
-  # redis
-  # redis:
-  #   image: "redis:3.0.3"
-  #   container_name: "typeorm-redis"
-  #   ports:
-  #     - "6379:6379"
+  redis:
+    image: "redis:8.6-alpine"
```
G0maa: Should we just use `redis:latest`?

test/functional/cache/redis-cache-integration.test.ts
G0maa: Hmm I can see why the tests are not working, `tests-linux.yml` is missing a `redis` service, but I don't want `redis-cache-integration.test.ts` test to run for all DBMSs, it's irrelevant to a DBMS, any idea for how to workaround this?

docker-compose.yml
alumni: We want to have reproducible builds, so every version should be pinned (preferably down to the patch if possible).

test/functional/cache/redis-cache-integration.test.ts
alumni: One option would be to use `testcontainers`, we're in the process of adding it at work as well, also to test redis-related stuff in integration tests. I think this is the easier for now.

Another option would be to have separate redis tests. I think separated tests will become more common in the monorepo. I'm not sure how we'll end up handling the containers for these tests though.

alumni:
I updated the PR title so it gets picked up as a breaking change.

Could you please update the v1 upgrade guide (I think it's still called migration)?

docker-compose.yml
G0maa: done

test/functional/cache/redis-cache-integration.test.ts
G0maa: Done :+1:

G0maa:
It seems `testcontainers` are not supported on Windows, I'll skip the test for Windows platfrom.

G0maa:
Tests are being skipped :sweat_smile:, any idea to fix this? @gioboa @alumni

pkuczynski:
> It seems `testcontainers` are not supported on Windows, I'll skip the test for Windows platfrom.

I don't think we should spin of container from the test. We should rather assume the test environment has everything we need. So I would rather rely on the side containers in GHA.

pkuczynski:
> Tests are being skipped :sweat_smile:, any idea to fix this? @gioboa @alumni 

This should fix it: https://github.com/typeorm/typeorm/pull/12145

PR #12064, opened 2026-03-03
perf(relations)!: use inner join for non-nullable ManyToOne and OneToOne relations

chunghyunleeme:
## Summary

- Use `INNER JOIN` instead of `LEFT JOIN` when a relation is explicitly marked as `nullable: false`, matching JPA's `@JoinColumn(nullable = false)` behavior
- This is semantically correct since the related entity is guaranteed to exist, and allows the DB optimizer to produce more efficient query plans
- Relations with `nullable: true` (the default) continue using `LEFT JOIN`, preserving full backward compatibility

## Changes

- **`SelectQueryBuilder.ts`** (4 locations): `type: "left"` → `type: relation.isNullable ? "left" : "inner"` in `buildRelations()`, `buildEagerRelations()`, `buildOrder()`, `buildWhere()`
- **`FindOptionsUtils.ts`** (3 locations): conditional `leftJoin`/`innerJoin` and `leftJoinAndSelect`/`innerJoinAndSelect` based on `relation.isNullable`, plus duplicate join validation fix for INNER joins
- **Test**: added `nullable: false` eager relation entity and test case verifying INNER JOIN generation

## Test plan

- [x] Existing eager-relations tests pass (backward compatibility)
- [x] New test verifies generated SQL contains `INNER JOIN` for `nullable: false` relations
- [x] New test verifies data loads correctly with INNER JOIN
- [x] Full test suite passes (2494 passing)

### DIFF
--- a/src/find-options/FindOptionsUtils.ts
+++ b/src/find-options/FindOptionsUtils.ts
@@ -86,11 +86,18 @@ export class FindOptionsUtils {
      * - The target entity has no soft-delete column, or withDeleted is enabled
      * @param relation
      * @param withDeleted
+     * @param parentJoinType
      */
     static getRelationJoinType(
         relation: RelationMetadata,
         withDeleted: boolean,
+        parentJoinType: "inner" | "left" = "inner",
     ): "inner" | "left" {
+        // If the parent was LEFT-joined, all descendants must also be LEFT
+        // to avoid filtering out rows where the parent alias is NULL
+        if (parentJoinType === "left") {
+            return "left"
+        }
         if (!relation.isNullable && relation.isWithJoinColumn) {
             const hasSoftDelete =
                 relation.inverseEntityMetadata.deleteDateColumn
@@ -307,6 +314,7 @@ export class FindOptionsUtils {
         qb: SelectQueryBuilder<any>,
         alias: string,
         metadata: EntityMetadata,
+        parentJoinType: "inner" | "left" = "inner",
     ) {
         metadata.eagerRelations.forEach((relation) => {
             // generate a relation alias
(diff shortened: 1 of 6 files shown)


alumni (approved):

pkuczynski:
Thanks for the PR — the concept is solid. I've made a few amendments and pushed them to the branch:

**1. Scoped INNER JOIN to join-column relations only**

The original code used `relation.isNullable` on all relation types, but `nullable: false` only guarantees a related row exists for relations that own the join column (`ManyToOne`, owning `OneToOne`). For `OneToMany`, `ManyToMany`, and inverse `OneToOne`, using INNER JOIN could silently drop parent rows when the collection is empty.

Changed all 6 locations to: `!relation.isNullable && relation.isWithJoinColumn ? "inner" : "left"`

**2. Strengthened test assertions**

The original test only checked `expect(query).to.contain("INNER JOIN")` which doesn't verify *which* relation uses which join type. Now asserts against specific table and alias names:
```typescript
expect(query).to.match(/INNER JOIN .?user.? .?post__requiredAuthor.?/)
expect(query).to.match(/LEFT JOIN .?user.? .?post__author.?/)
```

**3. Added migration guide entry**

Documented the behavior change under `## Relations` → `### nullable: false now uses INNER JOIN` with a "potentially breaking" warning about data integrity.

pkuczynski:
Follow-up to my previous review — pushed additional changes:

**4. Soft-delete safety**

If the target entity has a `@DeleteDateColumn`, TypeORM appends `IS NULL` to the join condition to filter soft-deleted rows. With `INNER JOIN`, this would silently drop parent rows when the related entity is soft-deleted — a behavior regression for common setups.

Fixed: when the target entity has `@DeleteDateColumn` and `withDeleted` is not set, the join falls back to `LEFT JOIN` even for `nullable: false` relations. Extracted a `getRelationJoinType()` helper in `SelectQueryBuilder` to centralize this logic.

**5. Comprehensive join strategy test suite**

Added `test/functional/relations/join-strategy/` with dedicated entities covering all relation types. 26 tests organized by relation type (`ManyToOne`, `OneToOne owner`, `OneToOne inverse`, `OneToMany`, `ManyToMany`, `soft-delete interaction`) × access method (`QueryBuilder`, `find methods`).

Each test verifies behavior through **result counts** — two posts are created (one with all relations, one with only required ones), so LEFT JOIN and INNER JOIN produce observably different results. QueryBuilder tests additionally assert the SQL join type via regex.

**6. Relations documentation updated**

Updated `docs/docs/relations/1-relations.md` and `docs/docs/relations/5-eager-and-lazy-relations.md` to document that `nullable: false` on `ManyToOne`/owning `OneToOne` now uses `INNER JOIN`.

**7. SonarQube fix**

Flattened nested `if`/`else` in `FindOptionsUtils.ts` (S6660).

Issue #438, opened 2017-05-01
how can i define unsigned column?


sajjad-shirazy:

pleerock:
custom column types are not supported yet. But its planned in the future

pleerock:
All database specific column types are supported now since `0.1.0-alpha.10`

jdpalumbo:
@pleerock How exactly is this achieved? I can not seem to find any documentation or property in the `ColumnOptions` interface for defining an unsigned column?

godvsdeity:
@pleerock i'm wondering the same, how do i get typeorm to generate the following migration code `id INT UNSIGNED NOT NULL AUTO_INCREMENT`

vkosachev:
Are there any workarounds on this ?

AlexMesser:
`UNSIGNED` and `ZEROFILL` attributes avaliable from `typeorm@0.2.0-alpha.40`. You can try it via `npm i typeorm@next`. Usage:

```typescript
@Entity()
export class Post {

    @PrimaryColumn({ unsigned: true })
    id: number;

    @Column({ zerofill: true })
    num: number;

}
```

d-yokoi:
@AlexMesser 
I want to add `UNSIGNED` attribute on `PrimaryGeneratedColumn` decorated property but it's not supported right?

AlexMesser:
fixed and released in `typeorm@0.2.0-alpha.45`. You can try it via `npm i typeorm@next`.

d-yokoi:
Thank you! I'll use it.

PR #12084, opened 2026-03-04
feat(mysql)!: remove deprecated MySQL integer display width and zerofill support

### LINKED ISSUE #12076: Remove MySQL deprecations (width, zerofill etc) (XS-S)

pkuczynski:
Remove the `width` and `zerofill` column options that were deprecated in MySQL 8.0.17
and removed in MySQL 8.4. These options only applied to MySQL/MariaDB/Aurora-MySQL
integer types and controlled display formatting, not storage.

Fixes #12076

Changes:
- Remove `width` and `zerofill` from all column option interfaces, metadata, and schema builder types
- Remove width/zerofill handling from MySQL, Aurora-MySQL, and CockroachDB drivers
- Remove dead code from SpannerDriver's `createFullType()` that referenced `column.width`
  (unreachable since Spanner's INT64 has no display width concept and the driver never populated this field)
- Remove `dataTypeDefaults` width entries for MySQL integer types
- Delete tests specific to width/zerofill behavior (issues #438, #6442, column-width tests)
- Update remaining tests and entity definitions to remove width/zerofill usage
- Add migration guide entry documenting the removal

---

@alumni Should we also remove `unsignedColumnTypes` from `MysqlDriver` and `AuroraMysqlDriver`? After removing zerofill, it's no longer referenced anywhere in the codebase.


naorpeled (approved):

alumni:
Hmm, I believe `unsigned` still works (it's not a formatting/display option, it's an actual data type option). But initially that constant was called "with width column types", IIRC, I refactored it when I added support for MySQL v9 and deprecated `width` and `zerofill` that no longer work. So it could be it's just an unused constant. The type for column options should probably still be used though.

alumni:
I merged this first since it has more files changed, hope it's fine :)

_Later edit:_ probably should've waited until `unsignedColumnTypes`

PR #12086, opened 2026-03-04
style: enable no-return-assign eslint rule and fix violations

pkuczynski:
Enable the `no-return-assign` ESLint rule with the `"always"` option to prevent implicit returns of assignments in arrow functions. Fix all 18 existing violations in `src/` by converting to block body syntax.

The rule catches patterns like `.forEach((x) => (x.prop = value))` where the assignment is implicitly returned — `.forEach()`, `new Promise()`, etc. all ignore the return value, making this unnecessary and confusing.

> **Note:** This PR depends on #12085 being merged first, which fixes the ~650 equivalent violations in `test/`.


alumni:
I think it might be the TS compiler option `strictFunctionTypes` (by default, we disabled it) that does a bit more than this.

I need to check in a project where I know these things get highlighted.

Issue #12078, opened 2026-03-04
Remove the deprecated methods from `QueryBuilder` classes and from the `Repository` / `EntityManager` APIs. Remove deprecated lock modes. Remove `AbstractRepository`, `WhereExpression` etc. Skip `findByIds`. (S)


alumni:

pkuczynski:
This is _completed_ as soon as those 4 PRs are merged...

Issue #11603, opened 2025-08-16
Remove APIs marked as deprecated in v0.3.0


michaelbromley:
In v0.3.0, several APIs were changed but the existing APIs were kept but marked as deprecated.

Examples: `Connection` -> `DataSource`, `findByIds`.

In the v1.0 of TypeORM, we will remove these deprecated APIs.

This issue is currently a placeholder to track the overall task, but needs to have a list of the exact APIs to be removed, and should be referenced in any PRs removing these APIs.

jovanadjuric:
Releasing v1.0 without any deprecated APIs sounds like a good idea. I'll start working on it this week and I'll document things as I go. I'll keep you posted

alumni:
@jovanadjuric nice, there might be some things that don't have an alternative (or some deprecated functions that are used internally).

I'm thinking about `ConnectionManager` - we actually implemented our own in our project, but others might still rely on it. I don't think it should be a part of the ORM, but could be released as a separate package. Might be easier with the monorepo. Or we can let the community do that.

I think one that is used internally is `findByIds` - in `SubjectDatabaseEntityLoader` and `PlainObjectToDatabaseEntityTransformer`. First usage is probably easy to replace, second IDK. Not sure if there others.

We might need to decide how to separate the work for the next patch and the next major. Shall we create a `main` branch and make it the default?

@michaelbromley @naorpeled maybe we can have a meeting when everyone's back from holidays (late August / early September?) to plan the next major.

naorpeled:
> @jovanadjuric nice, there might be some things that don't have an alternative (or some deprecated functions that are used internally).
> 
> I'm thinking about `ConnectionManager` - we actually implemented our own in our project, but others might still rely on it. I don't think it should be a part of the ORM, but could be released as a separate package. Might be easier with the monorepo. Or we can let the community do that.
> 
> I think one that is used internally is `findByIds` - in `SubjectDatabaseEntityLoader` and `PlainObjectToDatabaseEntityTransformer`. First usage is probably easy to replace, second IDK. Not sure if there others.
> 
> We might need to decide how to separate the work for the next patch and the next major. Shall we create a `main` branch and make it the default?
> 
> @michaelbromley @naorpeled maybe we can have a meeting when everyone's back from holidays (late August / early September?) to plan the next major.

Yeah, let's do that, I'll send a poll at the beginning of September

michaelbromley:
## Deprecated API Survey — TypeORM v1.0

[...]

PR #12090, opened 2026-03-05
refactor(QueryBuilder)!: remove onConflict(), deprecated orUpdate() overload, and setNativeParameters()

pkuczynski:
Removes deprecated QueryBuilder methods as part of #12078:

- `InsertQueryBuilder.onConflict()` — use `orIgnore()` or `orUpdate()` instead
- Deprecated `orUpdate()` overload accepting `{ columns?, overwrite?, conflict_target? }` — use the array-based signature
- `QueryBuilder.setNativeParameters()` — use `setParameters()` instead
- `QueryExpressionMap.onConflict` property and its clone logic

Tests for removed APIs deleted; migration guide updated.

Refs #12078

Part of #11603.


src/query-builder/SelectQueryBuilder.ts
```diff
@@ -3670,8 +3670,10 @@ export class SelectQueryBuilder<Entity extends ObjectLiteral>
                     this.expressionMap.cacheDuration,
                 )
                 .setParameters(this.getParameters())
-                .setNativeParameters(this.expressionMap.nativeParameters)
-                .getRawMany()
+            paginationQueryBuilder.expressionMap.nativeParameters = {
```
michaelbromley, line 3627: Question: in the `setNativeParameters` deprecation it says:

> deprecated Use `setParameters` instead

is that still true? I'm wondering why then here we have direct assignment to the object.

pkuczynski: I am not 100% sure why the setNativeParameters parameters is marked as deprecated. I just followed the deprecation to make it work. Had the same doubts as you and was hoping somoene else might know...

alumni: Whenever we pass parameters, e.g. to `QueryBuilder.setParameters(params)` or to `QueryBuilder.where(query, params)` the parameters are always sent to the database client without any modification nowadays (not sure if there are exceptions).

So now, all parameters are native parameters.

In the past, I remember that TypeORM was serializing them (and escaping) in the query whenever it was possible, e.g. `string`/`number` values in JS were converted to SQL, but `Buffer` was still passed as a parameter to the database client.

alumni: > So now, all parameters are native parameters.

Not sure if something needs to be simplified in the `QueryExpressionMap` as well. Maybe you could double check this, but I think `QueryExpressionMap.nativeParameters` should also be removed.

alumni: Seems `QueryExpressionMap.nativeParameters` are deprecated and they should be removed together with `QueryBuilder.setNativeParameters`

pkuczynski: Sure thing. Let me do it as part of this PR then...

pkuczynski: As this become much bigger change then I expected, I separate it to it’s own PR: https://github.com/typeorm/typeorm/pull/12104. Have a look and let me know what you think?

PR #12093, opened 2026-03-05
refactor(lock)!: remove deprecated pessimistic_partial_write and pessimistic_write_or_fail lock modes

pkuczynski:
Removes the deprecated `pessimistic_partial_write` and `pessimistic_write_or_fail` lock modes from all type unions, switch cases, and tests.

Use `pessimistic_write` with `onLocked: "skip_locked"` or `onLocked: "nowait"` instead.

Migration guide updated with before/after examples for both `setLock()` and find options.

Refs #12078

Part of #11603.

### DIFF
--- a/docs/docs/drivers/sap.md
+++ b/docs/docs/drivers/sap.md
@@ -30,8 +30,6 @@ See [Data Source Options](../data-source/2-data-source-options.md) for the commo
     - `pingCheck` (boolean) — Whether to validate connections before use (default: false).
     - `poolCapacity` (number) — Maximum number of connections to be kept available (default: no limit).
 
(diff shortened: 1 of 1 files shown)


alumni:
From the top of my head: [Select using Query Builder
](https://dev.typeorm.io/docs/query-builder/select-query-builder#lock-modes)

Please check if there are other docs that need updating.

pkuczynski:
> From the top of my head: [Select using Query Builder ](https://dev.typeorm.io/docs/query-builder/select-query-builder#lock-modes)
> 
> Please check if there are other docs that need updating.

Great catch! Fixed. Thanks.

PR #12104, opened 2026-03-05
refactor!: remove internal nativeParameters plumbing from drivers and query builders

pkuczynski:
Follow-up to #12090. Removes the internal `nativeParameters` property and its flow through all drivers, which became dead code after the public API removal (`setNativeParameters()`, `onConflict()`).

- Removed `nativeParameters` property from `QueryExpressionMap` and its cloning in `clone()`
- Removed `nativeParameters` passing in `QueryBuilder.getQueryAndParameters()` and where-clause factory
- Removed `nativeParameters` parameter from `Driver.escapeQueryWithParameters()` interface
- Removed `nativeParameters` parameter and seeded `escapedParameters` array from all 11 driver implementations (Postgres, MySQL, Aurora MySQL, Oracle, CockroachDB, SQL Server, AbstractSqlite, ReactNative, MongoDB, SAP, Spanner)
- Removed `nativeParameters` propagation in `SelectQueryBuilder` pagination logic and commented-out lines
- Updated SQLite test file to remove nativeParameters-specific test cases and drop the third argument
- Added migration doc section about the removal

Part of #11603.

### DIFF
--- a/src/driver/Driver.ts
+++ b/src/driver/Driver.ts
@@ -170,7 +170,6 @@ export interface Driver {
     escapeQueryWithParameters(
         sql: string,
         parameters: ObjectLiteral,
-        nativeParameters: ObjectLiteral,
     ): [string, any[]]
 
     /**
--- a/src/driver/aurora-mysql/AuroraMysqlDriver.ts
+++ b/src/driver/aurora-mysql/AuroraMysqlDriver.ts
@@ -384,16 +384,12 @@ export class AuroraMysqlDriver implements Driver {
      * and an array of parameter names to be passed to a query.
      * @param sql
      * @param parameters
-     * @param nativeParameters
      */
     escapeQueryWithParameters(
         sql: string,
         parameters: ObjectLiteral,
-        nativeParameters: ObjectLiteral,
     ): [string, any[]] {
-        const escapedParameters: any[] = Object.keys(nativeParameters).map(
-            (key) => nativeParameters[key],
-        )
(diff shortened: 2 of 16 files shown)


naorpeled (approved):

docs/docs/guides/8-migration-v1.md
```diff
@@ -257,3 +257,7 @@ The object-based `orUpdate()` overload accepting `{ columns?, overwrite?, confli
 ### `QueryBuilder.setNativeParameters()`
 
 The `setNativeParameters()` method has been removed. Use `setParameters()` instead.
+
+### `QueryExpressionMap.nativeParameters`
+
+The internal `nativeParameters` property on `QueryExpressionMap` has been removed, along with the corresponding `nativeParameters` parameter on `Driver.escapeQueryWithParameters()`. If you were accessing these internals directly, use standard query parameters via `setParameters()` instead.
```
alumni: I'd say there's no need to add the removal of an internal property in the migrations. The message above should suffice.

Ideally, the migration guide should be simple to understand, so I'd skip less important things, regardless of what Qodo says.

We'll most likely take a look at it before publishing to decide what can be cut off and what is missing, but I wouldn't rely on that final review :)

pkuczynski: I was 50/50 to include this or not. So following your judgment, I will drop it. And yeah, we will have to review this doc prior release and clean it up

PR #12108, opened 2026-03-06
chore(deps-dev): update @stryker-mutator/core and mocha-runner to 9.6.0

pkuczynski:
Update `@stryker-mutator/core` and `@stryker-mutator/mocha-runner` from 9.5.1 to 9.6.0.


alumni:
not necessary, will update the issue

pkuczynski:
Does not hurt either to be on a fresher version which works

alumni:
> Does not hurt either to be on a fresher version which works

That's true, but I don't want to encourage issues like the one referenced in the PR :)

We already have one for `glob` in one of the recent v0.3 releases which wasn't really necessary and for some reason it appeared in the release notes as a security issue in typeorm, which was definitely not the case.

pkuczynski:
> That's true, but I don't want to encourage issues like the one referenced in the PR :)
> 
> We already have one for `glob` in one of the recent v0.3 releases which wasn't really necessary and for some reason it appeared in the release notes as a security issue in typeorm, which was definitely not the case.

Does not mention anything about security issue. Just a regular update of dev dep, thats all.

PR #12114, opened 2026-03-06
refactor!: remove deprecated findByIds from Repository and EntityManager

### LINKED ISSUE #12077: From `Repository` / `EntityManager` APIs remove `findByIds` (could be easily replaced with `QueryBuilder.whereInIds`?). (S)

pkuczynski:
Removes the deprecated `findByIds` method from all public APIs:

- `EntityManager.findByIds()`
- `Repository.findByIds()`
- `BaseEntity.findByIds()`
- `MongoEntityManager.findByIds()`
- `MongoRepository.findByIds()`

The method was deprecated in favor of `findBy` with the `In` operator:

```typescript
// Before
const users = await repository.findByIds([1, 2, 3])

// After
import { In } from "typeorm"
const users = await repository.findBy({ id: In([1, 2, 3]) })
```

Internal callers have been refactored:
- `PlainObjectToDatabaseEntityTransformer` now uses `createQueryBuilder().whereInIds().getMany()`
- `SubjectDatabaseEntityLoader` MongoDB path now uses `createEntityCursor` with `_id.$in` directly

Tests that directly called `findByIds` have been removed. Remaining test titles that referenced `findByIds` have been updated to reflect the actual methods being tested. The migration guide has been updated.

Closes #12077

Part of #11603.


src/persistence/SubjectDatabaseEntityLoader.ts
```diff
@@ -111,11 +112,44 @@ export class SubjectDatabaseEntityLoader {
...
+                    const objectIdInstance =
+                        PlatformTools.load("mongodb").ObjectId
+                    const cursor =
+                        mongoManager.createEntityCursor<ObjectLiteral>(
                             subjectGroup.target,
-                        ) as MongoRepository<ObjectLiteral>
-                    entities = await mongoRepo.findByIds(allIds, findOptions)
+                            {
+                                _id: {
+                                    $in: allIds.map((id) => {
```
alumni: `SubjectDatabaseEntityLoader` and `PlainObjectToDatabaseEntityTransformer` should contain as less driver dependent logic as possible.

Maybe it doesn't make sense to remove `findByIds` from Mongo if that means to move the entire implementation here?

Can you maybe check what was the intention when `findByIds` was deprecated? E.g. I see in the changelog "This change was made to provide a more type-safe approach for data querying". Is every ID in Mongo an `ObjectId`? Maybe someone more familiar with Mongo can help here. CC: @naorpeled?

pkuczynski: Yes, I am leaning towards exactly the same conclusion and working k restoring this for mongo

pkuczynski: Have a look now @alumni

naorpeled (approved):

alumni (approved):

PR #12145, opened 2026-03-08
ci: do not skip detect-changes on pull requests

pkuczynski:


alumni:
I think we could to fix a few more things while we are at it:
- nightly publish pipeline runs on forks
- docs publishing runs on forks

Not sure about coverage, for me it works and I am using it, I don't remember if I set it up or it just worked, I've had it for many years.

alumni:
@pkuczynski @G0maa I think this was our only check that was preventing running the workflows twice in this repo (if the contributor is pushing to this repo, not to their fork).

We could also just block push access for contributors 🤷🏻 

_Later edit_: I see that now the tests are no longer running in forks (unless it's the `master` branch) - maybe the condition should've been slightly adjusted instead of removed. We want to run the actions in forks in any branch (so PR authors can test their changes without spamming us or running too many actions in our repo) + we want to run in PRs and in our master/version branches.

PR #12151, opened 2026-03-08
refactor(query-builder)!: rename printSql to logQuery

naorpeled:
## Description
QueryBuilder had a TODO to rename `printSql` to `logSql`, this PR addresses it and updates the migration guide.


gioboa (approved):

pkuczynski (commented):
I am not sure if I would use `logSql` (since `sql` is correctly written `SQL`, as it’s acronym, so it should be `logSQL`), and not simply `log`? @alumni wdyt?

Other thant that the code looks legit.

PR #12152, opened 2026-03-08
feat(query-builder): support aliases in object-style select

naorpeled:
## Description
While digging into our implementation I realized that when using select, we currently don't support adding aliases in bulk, this PR addresses it.


src/query-builder/SelectQueryBuilder.ts
```diff
@@ -223,7 +271,7 @@ export class SelectQueryBuilder<Entity extends ObjectLiteral>
             })
         } else if (selection) {
             this.expressionMap.selects.push({
-                selection: selection,
+                selection: selection as string,
```
pkuczynski: It looks fishy, that you had to typecast it here

src/query-builder/SelectQueryBuilder.ts
```diff
@@ -162,6 +171,19 @@ export class SelectQueryBuilder<Entity extends ObjectLiteral>
...
                 selection: selection,
             }))
+        } else if (
+            typeof selection === "object" &&
+            selection !== null &&
+            !Array.isArray(selection) &&
+            typeof selection !== "function"
+        ) {
+            const entries = Object.entries(selection)
+            if (entries.length > 0) {
```
pkuczynski: You do not need `if`. Running `map` on empty array would be ignored anyway.

src/query-builder/SelectQueryBuilder.ts
```diff
@@ -214,6 +247,21 @@ export class SelectQueryBuilder<Entity extends ObjectLiteral>
             this.expressionMap.selects = this.expressionMap.selects.concat(
                 selection.map((selection) => ({ selection: selection })),
             )
+        } else if (
+            typeof selection === "object" &&
```
pkuczynski: This logic is repeated twice. Can you abstract it? Same with param type.  Code repeatition is always bad. Maybe you can rethink how the whole function is done.

pkuczynski (changes requested):
+1 for tests, but there is no documentation update, which is a blocker for me. You could also think how to avoid code duplication, since `addSelect` has most of the logic with a slight diff behavior. I am also wondering if we should throw when we detect incorrect type in the object and not silently ignore it?

src/query-builder/SelectQueryBuilder.ts
naorpeled: it was a slop leftover, removed it 🙏 
good catch!

src/query-builder/SelectQueryBuilder.ts
naorpeled: Done 🙏

pkuczynski (changes requested):
Thanks for the update Naor! The `parseSelectInput` extraction is clean. Two items still open:

**1. `select({})` should clear selections, not be a no-op**

Currently `select({})` keeps existing selections intact, but `select([])` clears them. This is inconsistent — both represent "select nothing" and the JSDoc says "replaces all previous selections".

In `parseSelectInput`, remove the `entries.length === 0` guard and just return the empty mapped array:

```typescript
} else if (typeof selection === "object" && selection !== null) {
    return Object.entries(selection).map(([sel, alias]) => ({
        selection: sel,
        aliasName: alias,
    }))
}
```

Update the test to expect cleared selections instead of preserved ones.

**2. Document hydration risk for custom aliases**

The `Record<string, string>` overload makes it very easy to alias entity columns, but `getMany()`/`getOne()` hydration expects driver-generated aliases (e.g. `post_id`) — custom aliases will silently produce entities with `undefined` fields. This isn't a new limitation (same thing happens with `.select("user.id", "myAlias")`), but the object syntax makes it much easier to hit.

Please add:
- A JSDoc warning on the `Record<string, string>` overloads for both `select()` and `addSelect()` noting that custom aliases require `getRawMany()`/`getRawOne()`
- A note in the docs (`1-select-query-builder.md`) next to the new object-map examples, e.g.:

> **Note:** Custom aliases change the column names in the result set. Use `getRawMany()` / `getRawOne()` when using aliases. Using `getMany()` / `getOne()` with custom aliases will result in entities with undefined fields, because entity hydration relies on TypeORM's internal column naming.

Everything else looks good — tests are solid, reserved-word escaping coverage is nice.

pkuczynski (commented):
Nice work addressing the feedback! The `parseSelectInput()` extraction is clean and the docs/JSDoc warnings for hydration are solid.

One issue with the refactor:

**`select(fn)` behavior regression**

The function/subquery overload of `select()` previously **appended** to `expressionMap.selects`, but after the `parseSelectInput` extraction it now **replaces** all selections. This is because `parseSelectInput` returns a new array and the caller assigns it directly:

```typescript
const parsed = this.parseSelectInput(selection, selectionAliasName)
if (parsed !== undefined) {
    this.expressionMap.selects = parsed  // replaces instead of push
}
```

On master, the function branch did `this.expressionMap.selects.push(...)`. Please add a test for `select(subQueryBuilder => ...)` to verify the expected behavior and fix the regression.

michaelbromley:

```
git fetch origin
```

PR #12154, opened 2026-03-08
ci(publish): overhaul publish package strategy with tagging

pkuczynski:
Overhauls the publish-package workflow with a clearer publishing strategy, better nightly skip logic, and simplified configuration.

| Trigger | Target | Branch | Version | Tag |
|---------|--------|--------|---------|-----|
| GitHub Release | | v0.3 | `0.3.29` | `latest`/`legacy` (*) |
| GitHub Release | | master | `1.0.0` | `latest` (*) |
| GitHub Release | | master | `1.0.0-pre` | `next` |
| GitHub Release | | master | `1.0.0-rc.1` | `next` |
| Schedule | | v0.3 | `0.3.29-nightly.20260309` | `nightly` |
| Schedule | | master | `1.0.0-nightly.20260309` | `dev` |
| Manual | `release` | master | (same as GitHub Release) | (auto) |
| Manual | `master` | master | `1.0.0-nightly.20260309` | `dev` |
| Manual | `v0.3` | v0.3 | `0.3.29-nightly.20260309` | `nightly` |

(*) Stable releases are tagged `latest` only if the version is greater than the current npm `latest`. Otherwise falls back to `legacy` — prevents older branches from overtaking newer releases.

Changes:
- Simplified inputs: `publish_type` → `target` with `release`/`master`/`v0.3`
- Smart tag detection for releases: prerelease version → `next`, stable → `latest` only when greater than current
- `gitHead`-based nightly skip instead of 24h commit count — retries failed publishes
- Unified `should_publish` gate removes repeated conditions
- `pnpm view` instead of `npm view` (avoids `devEngines` warning)
- `pnpm version` for nightly versioning instead of manual Node script
- `--frozen-lockfile` for reproducible installs
- Added `timeout-minutes: 15`
- Removed unnecessary `Update npm` step and `fetch-depth` override


michaelbromley:
I think using the `latest` tag for both 0.3 and 1.0 branch releases has a small risk that, in future post v1.0 we do another 0.3 release (eg critical security patch) it would then get set as the latest.

pkuczynski:
> I think using the `latest` tag for both 0.3 and 1.0 branch releases has a small risk that, in future post v1.0 we do another 0.3 release (eg critical security patch) it would then get set as the latest. 

Fair point. We would have to release them in order if we release both or change the release workflow to only treat 1.x as latest. Do you see any other way?

michaelbromley:
What we could do is, if it comes up that we need to release a 0.3.x after the v1.0 release, we can manually set the `latest` tag back to v1.x using the [npm dist-tag](https://docs.npmjs.com/cli/v8/commands/npm-dist-tag) command. For me that's totally fine.

pkuczynski:
> What we could do is, if it comes up that we need to release a 0.3.x after the v1.0 release, we can manually set the `latest` tag back to v1.x using the [npm dist-tag](https://docs.npmjs.com/cli/v8/commands/npm-dist-tag) command. For me that's totally fine.

@michaelbromley I made [an amendment](https://github.com/typeorm/typeorm/pull/12154/changes/aa3881899e66c21c389156a38d9a5e4b02fcb5f9), to ensure we would not mess up, even if we do not do any update. What do you think now?

alumni:
I would not bother with nightly releases for v0.3, since all the PRs are going into master nowadays. We have to cherry PRs for a new 0.3 release, so there shouldn't be any new commits in that branch unless we do that.

.github/workflows/publish-package.yml
```diff
@@ -1,23 +1,45 @@
...
+
+name: Publish Package
 
 on:
   release:
     types:
       - published
   schedule:
-    - cron: "0 2 * * *" # Nightly at 02:00 UTC
+    - cron: "0 2 * * *"
```
alumni, line 32: Could we skip the schedule on forks? Every night i'm getting a notification that this action has failed.

pkuczynski:
> I would not bother with nightly releases for v0.3, since all the PRs are going into master nowadays. We have to cherry PRs for a new 0.3 release, so there shouldn't be any new commits in that branch unless we do that.

It does not publish nightly when there was no changes in that branch since last publish. It will be useful when we have 1.0. So I would still merge it as it is

.github/workflows/publish-package.yml
pkuczynski: Already done in setup phase.

PR #12155, opened 2026-03-09
fix(upsert): prevent input entity mutation during upsert result merge

### LINKED ISSUE #12094: upsert() with MySQL mutates input entity objects in-place with wrong auto-increment IDs after ON DUPLICATE KEY UPDATE
### Issue description

When calling upsert() on MySQL, TypeORM unconditionally writes back the auto-increment ID from insertResult.insertId into every input entity object — even when the rows were updated (not inserted). This silently corrupts the in-memory objects after the call returns.

### Expected Behavior

Input objects passed to upsert() should not be mutated. If mutation is intentional and documented, the generated ID should only be applied to rows that were genuinely inserted (had no prior value for the PK), not to rows that were updated.

### Actual Behavior

Any code that reads back from the input array after upsert() receives silently corrupted primary keys. A concrete example:

// Snapshot MUST be taken before upsert — taking it after produces wrong results:const deleteCriteria = entities.map(e => ({ employeeId: e.id }));  // ← must be before upsertawait repo.upsert(entities, ["id"]);// entities[*].id are now wrong — deleteCriteria would target the wrong rows
The callee has no indication that mutation occurred. There is no warning, no error, and the upsert itself succeeds.

### Steps to reproduce

Use upsert without spreading the entities and try to read the ID afterward.

### My Environment

| Operating System    |    Windows      |
| Node.js version     |  v23.10.0|
| Typescript version  | 5.7.3 |
| TypeORM version     | 0.3.25 |

### Additional Context

MysqlDriver.createGeneratedMap() contains:

MySQL's ON DUPLICATE KEY UPDATE always returns a non-zero insertId — it is the auto-increment value that would have been assigned had this been a fresh insert. It is non-zero even when every row in the batch was an update. TypeORM has no mechanism to distinguish inserts from updates in the result, so it applies this ID to every entity regardless.

ReturningResultsEntityUpdator.insert() then calls:

...which overwrites the original object's primary key with the wrong value in-place.

- [ ] aurora-mysql
- [ ] aurora-postgres
- [ ] better-sqlite3
- [ ] cockroachdb
- [ ] cordova
- [ ] expo
- [ ] mongodb
- [x] mysql
- [ ] nativescript
- [ ] oracle
- [ ] postgres
- [ ] react-native
- [ ] sap
- [ ] spanner
- [ ] sqlite
- [ ] sqlite-abstract
- [ ] sqljs
- [ ] sqlserver

Probably everywhere I noticed it with MySQL.

AI was used to write some parts of this, but I found and reviewed this as a legitimate user.

mbursali93:
Upsert queries (INSERT ... ON DUPLICATE KEY UPDATE) internally use
ReturningResultsEntityUpdator which merges generated values into
the provided entity.

This mutates the input object passed to repository.upsert().

This change skips merging generated values into the input entity
when expressionMap.onUpdate is set.

Insert behavior remains unchanged.

Added tests to verify:
- upsert does not mutate entity on insert
- upsert does not mutate entity on update
- normal insert still mutates entity

Fixes #12094


michaelbromley:

```
git fetch origin
```

PR #12158, opened 2026-03-09
ci(cockroachdb): upgrade image from v24.3.8 to v25.2.14

pkuczynski:
Upgrades the CockroachDB Docker image from v24.3.8 to v25.2.14 in both CI and docker-compose.

CockroachDB tests currently take 20+ minutes while other databases complete in 2–3 minutes. A significant part of the overhead comes from schema operations — CockroachDB cannot batch DDL in transactions, so every `synchronize(true)` call in tests runs sequential DROP/CREATE statements with expensive catalog introspection queries. v25.2 includes DDL and startup performance improvements that should help reduce this overhead.

### DIFF
--- a/docs/blog/2024-10-01-future-of-typeorm/index.md
+++ b/docs/blog/2024-10-01-future-of-typeorm/index.md
@@ -49,7 +49,7 @@ We'll also remain in close contact with [Umed](https://github.com/pleerock) to k
 
 ## Join Us in Supporting TypeORM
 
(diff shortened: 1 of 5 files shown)


naorpeled (approved):

pkuczynski:
Yes, I won't merge it until I get this faster. This could be a hiccup. I saw this db run sometimes even timing out after 1h

PR #12159, opened 2026-03-09
perf(cockroachdb): batch DROP statements in clearDatabase()

pkuczynski:
CockroachDB tests take 20+ minutes while other databases complete in 2–3 minutes. A major contributor is `clearDatabase()`, which is called via `synchronize(true)` in every test's `beforeEach`. Previously, it executed each DROP as a separate query — for N tables, that meant N+1 round-trips (1 SELECT + N individual DROPs), repeated for views, tables, sequences, and enum types.

This batches all DROP operations into single statements:

```sql
-- Before: N individual queries
DROP TABLE IF EXISTS "public"."users" CASCADE;
DROP TABLE IF EXISTS "public"."posts" CASCADE;
DROP TABLE IF EXISTS "public"."comments" CASCADE;

-- After: 1 batched query
DROP TABLE IF EXISTS "public"."users", "public"."posts", "public"."comments" CASCADE
```

Applied to all four object types (views, tables, sequences, enums), reducing the total from 4×(N+1) queries down to 4×2 queries per `clearDatabase()` call.

Additionally hardens the DROP statements:
- Use `quote_ident()` for enum type names instead of manual double-quote concatenation
- Add `IF EXISTS` to `DROP SEQUENCE` for resilience to concurrent cleanup
- Filter `information_schema.tables` by `table_type = 'BASE TABLE'` to avoid dropping views as tables

### DIFF
--- a/src/driver/cockroachdb/CockroachQueryRunner.ts
+++ b/src/driver/cockroachdb/CockroachQueryRunner.ts
@@ -3025,7 +3025,7 @@ export class CockroachQueryRunner
             // drop tables
             const tables: ObjectLiteral[] = await this.query(
                 `SELECT quote_ident(table_schema) || '.' || quote_ident(table_name) as "name" ` +
-                    `FROM "information_schema"."tables" WHERE "table_schema" IN (${schemaNamesString})`,
+                    `FROM "information_schema"."tables" WHERE "table_schema" IN (${schemaNamesString}) AND "table_type" = 'BASE TABLE'`,
             )
 
             if (tables.length > 0) {
@@ -3042,7 +3042,7 @@ export class CockroachQueryRunner
 
             if (sequences.length > 0) {
                 await this.query(
-                    `DROP SEQUENCE ${sequences.map(({ name }) => name).join(", ")}`,
+                    `DROP SEQUENCE IF EXISTS ${sequences.map(({ name }) => name).join(", ")}`,
                 )
             }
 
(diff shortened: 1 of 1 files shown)


src/driver/cockroachdb/CockroachQueryRunner.ts
```diff
@@ -3006,30 +3006,44 @@ export class CockroachQueryRunner
...
+                )
             }
 
-            const selectDropsQuery = `SELECT 'DROP TABLE IF EXISTS ' || quote_ident(table_schema) || '.' || quote_ident(table_name) || ' CASCADE;' as "query" FROM "information_schema"."tables" WHERE "table_schema" IN (${schemaNamesString})`
-            const dropQueries: ObjectLiteral[] =
-                await this.query(selectDropsQuery)
-            for (const q of dropQueries) {
-                await this.query(q["query"])
+            // drop tables
+            const tables: ObjectLiteral[] = await this.query(
```
alumni, line 3026: We can maybe extend this to `PostgresQueryRunner`, this is just copy-paste from over there.

I'm not against code duplication (sometimes it can be better than creating over-engineered DRY code), but IMO in this case the CRDB driver should've extended from the Postgres one to avoid maintaining both drivers.

I've seen this way of generating queries (in SQL) for some time, not just in `clearDatabase`, but in other methods too, just didn't want to change it yet - sometime in 2.0 or 3.0 we can improve this elsewhere.

pkuczynski: Same optimization has been applied to the PostgreSQL driver in #12164.

I would rather do the extension in a separate PR. I am not sure how compatible they are?

PR #12160, opened 2026-03-09
ci(cockroachdb): reduce store size from 4GB to 2GB

pkuczynski:
The CockroachDB CI service was configured with `--cache=1GB --store=type=mem,size=4GB`, allocating 5GB of RAM. GitHub Actions `ubuntu-latest` runners only have 7GB total, leaving just 2GB for the OS, Node.js, and the entire test suite. This likely causes memory pressure and swap thrashing, contributing to the 20+ minute test runtime.

This removes the explicit `--cache` override (letting CockroachDB auto-size it) and reduces the in-memory store from 4GB to 1GB — plenty for the test data while leaving ~6GB for everything else.

The same change is applied to docker-compose.yml (previously `--cache=.25 --store=type=mem,size=.25`).

Before: `start-single-node --insecure --cache=1GB --store=type=mem,size=4GB`
After: `start-single-node --insecure --store=type=mem,size=1GB`

Note: in-memory stores require an explicit size — omitting it causes CockroachDB to fail with `size must be specified for an in memory store`.

### DIFF
--- a/.github/workflows/tests-linux.yml
+++ b/.github/workflows/tests-linux.yml
@@ -18,7 +18,7 @@ jobs:
         ports:
           - "26257:26257"
         env:
-          COCKROACH_ARGS: "start-single-node --insecure --cache=1GB --store=type=mem,size=4GB"
+          COCKROACH_ARGS: "start-single-node --insecure --store=type=mem,size=2GB"
         options: >-
           --health-cmd "cockroach sql --insecure -e 'SELECT 1'"
           --health-interval 10s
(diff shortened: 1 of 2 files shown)


gioboa (approved):

alumni:
Seems it's slower now? (30 minutes)

Cockroach (also Oracle and SAP HANA) use a lot of resources. Usually if the tests are slow, it's rather because of the database (which is memory-hungry, sometimes CPU-hungry or rather waiting for another thread).

`ubuntu-latest` should have 4 CPUs + 16GB RAM ([GitHub-hosted runners reference](https://docs.github.com/en/actions/reference/runners/github-hosted-runners)), so IMO rather increasing than reducing the resources available to the database would help.

PR #12162, opened 2026-03-09
ci: consolidate test configs into ormconfig.sample.json

pkuczynski:
Remove all dedicated `.github/workflows/test/*.ormconfig.json` files and use `ormconfig.sample.json` as the single source of truth for both local development and CI. Each CI job now uses `jq` to filter the relevant database entry by type.

- **Credentials standardized** — MySQL, MariaDB, and PostgreSQL all use `typeorm/typeorm/typeorm` for user, password, and database name
- **Postgres CI fix** — the workflow was using `POSTGRES_USERNAME` (not a valid env var for the postgres image) which accidentally worked because the default user is `postgres`. Now uses `POSTGRES_USER` correctly
- **Image versions aligned** — CI service containers now match docker-compose: MySQL 9.5.0, MariaDB 10.6.24/12.1.2, MongoDB 8.2.5
- **Security** — replaced hardcoded `MYSQL_ROOT_PASSWORD` with `MYSQL_RANDOM_ROOT_PASSWORD` since we connect as the `typeorm` user, not root
- **Cleanup** — removed commented-out redis block and unnecessary named volumes from docker-compose


OSA413:
Looks good. One suggestion: make DB name, user and password different strings so that we catch if these three variables get messed up in the code somewhere (so that we certainly know that user is user, password is password, and DB name is DB name).

OSA413:
Fix the conflicts and see if this is actually works in the CI/CD on this end (or how should the pipeline be triggered to see that all things go right?)

pkuczynski:
> Looks good. One suggestion: make DB name, user and password different strings so that we catch if these three variables get messed up in the code somewhere (so that we certainly know that user is user, password is password, and DB name is DB name).

Very good idea! Let me do that!

PR #12165, opened 2026-03-09
ci: fix fork workflows

alumni:
### Description of change

Fix tests not running in branches in forks.

Fix forks trying to publish docs.

### Pull-Request Checklist


pkuczynski (changes requested):
This makes all the tasks executed twice. If you want to trigger the build in your form, just make a PR against master in your fork...

alumni:
@pkuczynski The workflow is conditional, so it should run only once. This got broken **twice** by your changes, it's the second time we're trying to fix it.

Another good option would be to block maintainers from pushing to the repo and force them to use their forks.

pkuczynski:
> @pkuczynski The workflow is conditional, so it should run only once. This got broken **twice** by your changes, it's the second time we're trying to fix it.

In this variant maybe, but not in the previous one I fixed. We might as well just run this job only on `push` to any branch and forget the pull request part. This way it will run always, right?

> Another good option would be to block maintainers from pushing to the repo and force them to use their forks.

Don’t really see how this is connected. On fork or not on fork, I never want this worflow to fire when I am not on a PR. And if you have a PR, it would already fire regardless of fork or not. The only issue you are trying to fix is to make it run on push to any branch in your fork without the PR, right?

pkuczynski:
Since this was not merged for 5 weeks, let's close it for now to clean up the open PR list and get back to it when needed.

PR #12187, opened 2026-03-11
refactor(find-options): un-deprecate OrderByCondition type

pkuczynski:
Un-deprecates `OrderByCondition` — it is actively used internally across 9 source files with no replacement type. Removing it would require either creating an identical replacement or inlining the shape everywhere, with no benefit.

Part of #11603


alumni:
@pkuczynski @naorpeled @gioboa maybe we can have a discussion before merging such PRs in the future, would be nice to have everyone on the same page.

CC: @michaelbromley

PR #12188, opened 2026-03-11
refactor(find-options)!: remove deprecated join option from find methods

pkuczynski:
Remove the deprecated `join` property from `FindOneOptions` and `FindManyOptions`, along with the `JoinOptions` interface and all related handling code.

- Delete `src/find-options/JoinOptions.ts` interface
- Remove `join` property from `FindOneOptions`
- Remove `join` handling from `SelectQueryBuilder.setFindOptions()`
- Remove `extractFindManyOptionsAlias()` and `join` type guard from `FindOptionsUtils`
- Remove unused `FindOptionsUtils` import from `EntityManager`
- Convert all test and sample `join:` usages to equivalent `relations:` syntax
- Add migration guide entry for the removal

### DIFF
--- a/src/find-options/FindOptionsUtils.ts
+++ b/src/find-options/FindOptionsUtils.ts
@@ -22,8 +22,13 @@ export class FindOptionsUtils {
      * This catches untyped/JS callers still passing `join` after its removal in v1.0.
      * @param options
      */
-    static rejectJoinOption(options: any): void {
-        if (options && typeof options === "object" && options.join != null) {
+    static rejectJoinOption(options: unknown): void {
+        if (
+            options &&
+            typeof options === "object" &&
+            "join" in options &&
+            options.join != null
+        ) {
             throw new TypeORMError(
                 `"join" option has been removed. Use "relations" for left joins ` +
                     `or QueryBuilder for other join types. See the v1 migration guide for details.`,


docs/docs/guides/8-migration-v1.md
```diff
@@ -324,6 +324,52 @@ categoryCount: number
...
+
+// After
+const posts = await repository.find({
+    relations: { categories: true, author: true },
+})
+```
+
+For more complex join scenarios (e.g., `leftJoin` without select, or `innerJoin`), use the QueryBuilder API directly.
+
+### String-based `select` and `relations` removed
```
naorpeled: Are we sure about removing the string relations syntax?
I feel that sometimes it's cleaner, and what I would add to it is better type safety

@gioboa @alumni WDYT?

pkuczynski: I think it’s actually worse typesefty and harder for refactoring. But yeah, had the same doubts if we should or should not remove it...

michaelbromley: Does the string syntax allow anything the object syntax does not? 

People can always create a helper function if they want to stick to strings for whatever reason.

> I think it’s actually worse typesefty and harder for refactoring.

@pkuczynski are you referring to the string syntax as worse or object syntax as worse here?

pkuczynski: > @pkuczynski are you referring to the string syntax as worse or object syntax as worse here?

I meant tools from JetBrains seems to have less problems and better typing support for object syntax than strings. Maybe this is because of how we define types. However I am not 100% if removing string syntax is a good idea. I would leave this up to @alumni and you to decide. I only executed the task I have been assigned to...

alumni: Well, initially the `string[]` syntax was deprecated because typescript didn't have that good inference and the plan was to recursively derive types based on the object syntax.

However, in 4.1, TS added support for template literal types which can help us solve that problem whenever we will want to improve the return types.

Another thing is the behavior:
```ts
manager.findBy(PostEntity, { relations: ['categories.admin'] })
```
Does this make sense? How should it work? Should it also load `categories` relation first? Or directly the `admin` of each category?

```ts
manager.findBy(PostEntity, { relations: { categories: { admin: true }}})
```
should definitely load the `categories` of each post, and the `admin` of each category, because there's no other way to differentiate.

FYI the `string[]` syntax is also quite used at the moment because it's not highlighted by TS/ESLint as deprecated.

In general, I'm strongly in favor of keeping things as they (do not remove, do not undeprecate) until we have a good replacement for such things. There are a few things (not many) where we need to be sure that won't cause issues despite the major release (`findByIds` + this), simply because they can be important to users (widely used). I would not rush to make a decision because we already have a 2.0 in mind. We can do the remaining things then.

And I'm thinking especially about this syntax, because we can also look into improving the return types then, which is a longstanding issue that made a lot of people avoid TypeORM.

alumni:
Following https://github.com/typeorm/typeorm/pull/12188#discussion_r2926842633, I'd only remove `join` for now, keep everything else unchanged.

docs/docs/guides/8-migration-v1.md
michaelbromley: This is sound reasoning @alumni and I'm also happy to go along with your recommendations here.

naorpeled (approved):

docs/docs/guides/8-migration-v1.md
pkuczynski: I cut the string based select and relations into #12214 and #12215. We can discuss this on the next meeting... Until then I will merge this one, as this is something we were all certain we want to do...

PR #12189, opened 2026-03-11
ci: use typeorm org Docker images for CI

naorpeled:
## Summary
- Migrate PostgreSQL test images from `ghcr.io/naorpeled/typeorm-postgres` to `ghcr.io/typeorm/postgres-test-images`
- Updates `tests-linux.yml`, `mutation-test.yml`, and `docker-compose.yml`
- Images are now owned by the typeorm org rather than a personal registry

## Test plan
- [ ] Verify CI tests pass with the new image references
- [ ] Verify docker-compose still works for local development


gioboa (approved):

alumni:
Does this affect how coveralls merges the reports? (I think not, just wanted to be sure)

Seems the tests did not run, would be great if you could also update detect changes to run the tests when the workflow definitions for them are changed. Then we'll confirm the coverage is fine.

PR #11881, opened 2025-12-28
ci: setup pnpm

alumni:
### Description of change

Sets up `pnpm` to replace `npm` for the main project. Docs are left unchanged for the moment, they will soon be a separate package/workspace in the monorepo.

I used `pnpm import` to convert the lockfile without altering the dependency versions.

### Pull-Request Checklist

-   [x] Code is up-to-date with the `master` branch
-   [ ] This pull request links relevant issues as `Fixes #00000`
-   [ ] There are new or updated tests validating the change (`tests/**.test.ts`)
-   [x] Documentation has been updated to reflect this change (`docs/docs/**.md`)

PR #12193, opened 2026-03-12
ci: migrate from npm to pnpm

pkuczynski:
Migrates the v0.3 branch from npm to pnpm, matching what was done on master in #11881. Changes: all CI workflows use pnpm, `package-lock.json` replaced with `pnpm-lock.yaml`, husky pre-commit updated, gulpfile and DEVELOPER.md updated, added `@types/gulp` dev dependency.

This is needed for the package publish workflow to work.


gioboa (approved):

naorpeled:
I think @alumni did this on purpose, let's wait for this reply on this 🙏

alumni:
We said we'll not bother with 0.3 that much so this would not be needed. Not a problem to do it, but it's nothing more than wasted effort.

It also adds even more to the repo noise in the past 2 weeks, which makes it more difficult to follow what is really important and what not.

pkuczynski:
> We said we'll not bother with 0.3 that much that this would not be needed. Not a problem to do it, but it's wasted effort.

It was super easy change, took me few minutes. So not really wasted. We don't know how many things we will need to port back to v0.3...

pkuczynski:
Anyway, it's done. Can you approve then?

alumni:
@naorpeled is it okay without the docs? 

I remember last time the docs weren't working anymore :)

naorpeled:
> @naorpeled is it okay without the docs? 
> 
> I remember last time the docs weren't working anymore :)

It's because we had it setup on CF and not in a GH, now it should work

alumni:
Ah great, thanks for confirming! :)

PR #12197, opened 2026-03-12
fix(query-runner): parameterize SQL queries across all drivers

pkuczynski:
Replace string concatenation with parameterized queries across all database drivers to prevent SQL injection via schema/database/table/column/view names.

- MySQL, Aurora MySQL: `clearDatabase()`, `hasDatabase()`, `hasTable()`, `hasColumn()`
- PostgreSQL: `hasDatabase()`, `hasSchema()`, `hasTable()`, `hasColumn()`
- SQL Server: `hasDatabase()`, `hasSchema()`, `hasTable()`, `hasColumn()`
- SAP HANA: `clearDatabase()`, `hasTable()`, `hasColumn()`
- Oracle: `hasTable()`, `hasColumn()`
- Spanner: `hasSchema()`, `hasTable()`, `hasColumn()`, `loadTables()` initial input
- SQLite: `hasTable()`, `loadViews()`

Follows the same approach as #12185 (PostgreSQL/CockroachDB `clearDatabase()`).

### DIFF
--- a/src/entity-manager/MongoEntityManager.ts
+++ b/src/entity-manager/MongoEntityManager.ts
@@ -107,16 +107,18 @@ export class MongoEntityManager extends EntityManager {
             | Partial<Entity>
             | FilterOperators<Entity>,
     ): Promise<Entity[]> {
-        const query =
+        const metadata = this.connection.getMetadata(entityClassOrName)
+        const query = this.replaceObjectIdProperty(
+            metadata,
             this.convertFindManyOptionsOrConditionsToMongodbQuery(
                 optionsOrConditions,
-            )
+            ),
+        )
         const cursor = this.createEntityCursor<Entity>(
             entityClassOrName,
             query as Filter<Entity>,
         )
-        const deleteDateColumn =
-            this.connection.getMetadata(entityClassOrName).deleteDateColumn
+        const deleteDateColumn = metadata.deleteDateColumn
         if (FindOptionsUtils.isFindManyOptions(optionsOrConditions)) {
             if (optionsOrConditions.select)
                 cursor.project(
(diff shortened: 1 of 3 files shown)


src/driver/aurora-mysql/AuroraMysqlQueryRunner.ts
```diff
@@ -272,8 +273,11 @@ export class AuroraMysqlQueryRunner
      */
     async hasTable(tableOrName: Table | string): Promise<boolean> {
         const parsedTableName = this.driver.parseTableName(tableOrName)
-        const sql = `SELECT * FROM \`INFORMATION_SCHEMA\`.\`COLUMNS\` WHERE \`TABLE_SCHEMA\` = '${parsedTableName.database}' AND \`TABLE_NAME\` = '${parsedTableName.tableName}'`
-        const result = await this.query(sql)
+        const sql = `SELECT * FROM \`INFORMATION_SCHEMA\`.\`COLUMNS\` WHERE \`TABLE_SCHEMA\` = ? AND \`TABLE_NAME\` = ?`
+        const result = await this.query(sql, [
+            parsedTableName.database,
+            parsedTableName.tableName,
+        ])
```
alumni, line 280: If you have SQL template literal highlighting in your IDE, this would also highlight the query:

```suggestion
        const result = await this.sql`SELECT * FROM \`INFORMATION_SCHEMA\`.\`COLUMNS\` WHERE \`TABLE_SCHEMA\` = ${parsedTableName.database} AND \`TABLE_NAME\` = ${parsedTableName.tableName}`
```

Unfortunately, GitHub doesn't highlight it like other platforms do, so it would work only in IDE :P

No strong need to do this (unless others prefer it), it might make the queries easier to read.

test/functional/query-runner/has-table.test.ts
```diff
@@ -0,0 +1,36 @@
...
+            schemaCreate: true,
+            dropSchema: true,
+        })
+    })
+    after(() => closeTestingConnections(dataSources))
+
+    it("should return true for existing tables and false for non-existing", () =>
+        Promise.all(
+            dataSources.map(async (dataSource) => {
+                const queryRunner = dataSource.createQueryRunner()
```
alumni: you can just use `await using` to avoid having to worry about the `release()`:

```ts
await using queryRunner = dataSource.createQueryRunner()
```

Otherwise, you'd have to put the `.should` statements in a `try` block and the `release()` in `finally` block.

alumni (approved):

PR #12199, opened 2026-03-12
ci: restore npm upgrade and bump Node to 24 for trusted publishing

pkuczynski:
The nightly publish was failing with ENEEDAUTH because the bundled npm
on Node 22 doesn't complete the OIDC token exchange for Trusted Publishing.

- Bump Node from 22 to 24 (ships with newer npm)
- Restore `npm install -g npm@latest` step before publish as safety net

### DIFF
--- a/.github/workflows/publish-package.yml
+++ b/.github/workflows/publish-package.yml
@@ -141,8 +141,11 @@ jobs:
       - uses: actions/setup-node@v6
         if: steps.publishable.outputs.should_publish == 'true'
         with:
-          node-version: 22
+          node-version: 24
           cache: "pnpm"
+
+      - run: npm install -g npm@latest
+        if: steps.publishable.outputs.should_publish == 'true'
 
       - run: pnpm install --frozen-lockfile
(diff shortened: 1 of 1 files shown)


alumni:
We should not be npm, but pnpm to publish. Does pnpm work?

michaelbromley:
The publish command _is_ `pnpm publish` but I guess it delegates to npm. This exact same publish command worked previously, before the npm update step was removed.

This npm update step is something widely recommended in guides I saw for implementing OIDC trusted publishing.

pkuczynski:
> The publish command _is_ `pnpm publish` but I guess it delegates to npm. This exact same publish command worked previously, before the npm update step was removed.
> 
> This npm update step is something widely recommended in guides I saw for implementing OIDC trusted publishing.

Yes, this is exactly what happens `pnpm -> npm`. Node 24 should fix it, but I re-added npm upgrade as a precautious measure.

alumni:
Yeah, found the confirmation as well: https://github.com/pnpm/pnpm/issues/9812#issuecomment-3135333412

PR #12202, opened 2026-03-12
ci: extract reusable composite actions for consistency and fix missing frozen-lockfile

pkuczynski:
Extract the repeated pnpm/setup-node/install boilerplate (~20 occurrences across workflows) into two reusable composite actions:

**`.github/actions/setup-node`** — sets up pnpm, Node.js with caching, and runs `pnpm install`. Accepts optional inputs:
- `node-version` — override for matrix jobs (defaults to 24)
- `working-directory` — for workspaces like `./docs`
- `cache-dependency-path` — for non-root lockfiles
- `download-build` — downloads the build artifact when `true`

**`.github/actions/build-docs`** — sets up Node.js and builds the documentation site

**Important:** all `pnpm install` calls now use `--frozen-lockfile`, which was previously missing in most workflows. This ensures CI always installs exact versions from the lockfile and fails fast on any lockfile drift.

Other changes:
- Merged `preview.yml` into `tests.yml` as a `preview` job (triggered only on `pull_request` events, gated by the `src` path filter)
- Removed step `name:` labels on `uses:` steps for consistency
- Fixed `gulpfile.js` → `gulpfile.ts` in the `src` path filter
- Removed unnecessary `fetch-depth: 0` from preview checkout

### DIFF
--- a/.github/actions/build-docs/action.yml
+++ b/.github/actions/build-docs/action.yml
@@ -7,7 +7,6 @@ runs:
     - uses: ./.github/actions/setup-node
       with:
         working-directory: ./docs
-        cache-dependency-path: docs/pnpm-lock.yaml
 
     - run: pnpm run build
       shell: bash


alumni (dismissed):
I think I commented on the commit - please do not setup an nvmrc file for a project that supports multiple node versions.

pkuczynski:
> I think I commented on the commit - please do not setup an nvmrc file for a project that supports multiple node versions.

And the reason for this is? If we use .nvrmc we ensure every developer works against the same version of Node and the same one used in CI...

alumni:
> If we use .nvrmc we ensure every developer works against the same version of Node and the same one used in CI

In CI we use multiple versions of node :) I hope we still do, we need to make sure the tests pass on them :D 

Actually, I had to change the local version of node a few times to check some things (e.g. certain versions of `better-sqlite3` were not compatible with certain versions of node and the problem was in one version).

Also, letting devs use multiple versions of node ensures more variability in testing. 

This is reason why I set up `pnpm`, so we have the same version of the package manager, but allow different node versions.

.github/actions/setup-node/action.yml
```diff
@@ -0,0 +1,39 @@
...
+      with:
+        node-version: ${{ inputs.node-version }}
+        cache: pnpm
+        cache-dependency-path: ${{ inputs.cache-dependency-path || '' }}
+
+    - run: pnpm install --frozen-lockfile
+      shell: bash
+      working-directory: ${{ inputs.working-directory || '.' }}
+
+    - uses: actions/download-artifact@v6
```
pujux: any specific reason to use v6 instead of v8?

.github/workflows/tests-linux.yml
```diff
@@ -8,17 +9,14 @@ on:
 
 jobs:
   cockroachdb:
-    # CockroachDB is slow, run less often so it doesn't block other workflows
-    if: ${{ inputs.node-version >= 22 }}
+    if: ${{ inputs.node-version >= 24 }}
```
pujux, line 12: is this needed?

.gitignore
```diff
@@ -3,7 +3,6 @@
 ._*
 
 ### Node ###
-.nvmrc
```
pujux: do we want to add .nvmrc?

pujux: okay just read lucians comments, we can remove this change then

.github/workflows/tests-linux.yml
pujux: running it on lower node versions too would be good no?

.github/actions/setup-node/action.yml
pkuczynski: I wanted to stick what we have now and upgrade all in a separate step...

.github/workflows/tests-linux.yml
pkuczynski: we run in 20 and 24, so this change actually does not make any difference...

PR #12207, opened 2026-03-13
fix(query-runner): parameterize queries and escape identifiers to prevent SQL injection

pkuczynski:
- Use parameterized queries instead of string interpolation in `hasDatabase`, `hasSchema`, `hasTable`, and `hasColumn` across all drivers (postgres, cockroachdb, mssql, mysql, oracle, sap, spanner, sqlite)
- Escape identifiers in DDL methods (`createDatabase`, `dropDatabase`, `createSchema`, `dropSchema`) to prevent injection via double-quote/backtick breakout (postgres, cockroachdb, mssql, mysql, sap)
- Parameterize `clearDatabase` INFORMATION_SCHEMA queries in MySQL
- Unskip regression tests for fixed methods; reorganize remaining skipped tests


alumni:
@pkuczynski please check if we don't already have method for escaping identifiers.

BTW, a lot more things would crash of the schema name would contain `"` (or even `.`), so this PR won't make those work. So far, we didn't really support these characters in identifier names, so if we are doing this to silence Sonar, we might as well mark it as a false positive.

pkuczynski:
Good catch — we did already have `driver.escape()` on every driver, but it was only wrapping identifiers in quotes without escaping internal quote characters. So `escape('my"table')` would produce the broken `"my"table"` instead of the correct `"my""table"`.

I've updated `escape()` across all drivers to properly escape internal quotes, and refactored the QueryRunner DDL methods to use `this.driver.escape()` instead of scattered inline `.replace()` calls. Also added unit tests for the escape method.

Regarding marking SonarQube findings as false positives — I'd push back on that. Even if identifiers with special characters aren't fully supported as a feature, the system should never allow them to break out of the SQL statement. This is a genuine injection vector, not a cosmetic concern.

PR #12208, opened 2026-03-13
refactor(mongo)!: drop MongoDB driver v6 support, require v7

naorpeled:
## Summary

- **Drop MongoDB Node.js driver v6 support** — peer dependency updated from `^6.0.0 || ^7.0.0` to `^7.0.0`
- **Regenerate type definitions** (`bson.typings.ts`, `typings.ts`) from mongodb v7 / bson v7
- **Remove `stats()` method** from `MongoQueryRunner`, `MongoEntityManager`, and `MongoRepository` — the underlying `collStats` command was deprecated in MongoDB server 6.2 and `Collection.stats()` was removed in driver v7
- **Remove deprecated BSON deserialization options** (`promoteBuffers`, `promoteLongs`, `promoteValues`, `raw`) from `MongoDataSourceOptions` and `MongoDriver` valid options — moved to BSON-level only in v7
- **Fix `dropCollectionIndexes` return type** to `boolean` (v7 API change)
- **Update documentation** and migration guide

## BREAKING CHANGES

- Minimum MongoDB Node.js driver version is now **v7.0.0**
- `stats()` method removed from `MongoQueryRunner`, `MongoEntityManager`, `MongoRepository`
- Connection options `promoteBuffers`, `promoteLongs`, `promoteValues`, `raw` removed

## Test plan

- [x] `pnpm run compile` passes with no type errors
- [x] No remaining `CollStats`/`CollStatsOptions` references in `src/`
- [x] `package.json` peer dependency shows `^7.0.0` only
- [ ] Run MongoDB test suite against a MongoDB 6.0+ instance

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -26,6 +26,11 @@ jobs:
               - package.json
               - package-lock.json
 
+            ci: &ci
+              - .github/**/*
+              - .pr_agent.toml
+              - pr_compliance_checklist.yaml
+
             docs: &docs
               - docs/**/*
 
@@ -39,16 +44,12 @@ jobs:
             src-or-tests: &src-or-tests
               - *src
               - .github/workflows/test*.yml
-              - ormconfig.sample.json
+              - docker-compose.yml
+              - docker/**/*
               - test/**/*
               - .c8rc.json
               - .mocharc.json
-
-            format:
-              - *src-or-tests
-              - .editorconfig
-              - .prettierignore
-              - .prettierrc.js
+              - ormconfig.sample.json
 
             lint:
               - *src-or-tests
@@ -94,8 +95,6 @@ jobs:
           retention-days: 1
 
   format:
-    if: contains(needs.detect-changes.outputs.changes, 'format')
-    needs: detect-changes
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v6
--- a/docs/src/css/custom.css
+++ b/docs/src/css/custom.css
@@ -70,8 +70,8 @@
 
     /* Additional UI colors for dark mode */
     --ifm-navbar-background-color: rgba(24, 0, 40, 0.8);
-    --ifm-background-color: #180028;
-    --ifm-footer-background-color: #180028;
+    --ifm-background-color: #1a0030;
+    --ifm-footer-background-color: #1a0030;
     --ifm-footer-color: #ffffff;
     --ifm-footer-link-color: #ffe6d9; /* Warmer white for better contrast */
     --ifm-footer-title-color: #ffffff;
@@ -227,7 +227,7 @@ div[class*="announcementBar"] .close:hover {
 }
 
 [data-theme="dark"] .footer {
-    background: linear-gradient(160deg, #180028 0%, #2a0045 100%) !important;
+    background: linear-gradient(160deg, #1a0030 0%, #2a0045 100%) !important;
     box-shadow: 0 -4px 12px rgba(0, 0, 0, 0.3);
 }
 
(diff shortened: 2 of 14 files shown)


src/driver/mongodb/bson.typings.ts
```diff
@@ -613,289 +961,381 @@ export declare class Long extends BSONValue {
...
+     * Throws an error if any of the following conditions are true:
+     * - the string contains invalid characters for the given radix
+     * - the string contains whitespace
+     * - the value the string represents is too large or too small to be a Long
+     * Unlike Long.fromString, this method does not coerce '+/-Infinity' and 'NaN' to Long.Zero
+     * @param str - The textual representation of the Long
+     * @param radix - The radix in which the text is written (2-36), defaults to 10
+     * @returns The corresponding Long value
+     */
+    static fromStringStrict(str: string, radix?: boolean): Long;
```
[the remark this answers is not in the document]
naorpeled: This is actually correct — the upstream `bson@7.2.0` source has this exact overload signature (`radix?: boolean`). It is one of several overloads of `fromStringStrict`: the second parameter can be `unsigned?: boolean` or `radix?: boolean` depending on the overload. Our typings file is a direct copy of `bson.d.ts`.

docs/docs/drivers/mongodb.md
```diff
@@ -2,7 +2,7 @@
 
 ## MongoDB support
 
-TypeORM has basic MongoDB support (server **6.0 or later** and Node.js driver **v6 or later**).
+TypeORM has basic MongoDB support (server **6.0 or later** and Node.js driver **v7 or later**).
```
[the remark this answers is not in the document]
naorpeled: Good catch — removed `promoteBuffers`, `promoteLongs`, `promoteValues`, `raw` from the options list and `stats` from the supported methods list in c6d3904.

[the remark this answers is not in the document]
pkuczynski: Do we support server version 6.0?

docs/docs/guides/8-migration-v1.md
```diff
@@ -93,7 +93,7 @@ new DataSource({
 
 ## MongoDB
 
-TypeORM now requires **MongoDB server 6.0 or later** and the **`mongodb` Node.js driver v6 or later** (`^6.0.0 || ^7.0.0`). Support for MongoDB server 5.x and the `mongodb` driver v5 has been dropped.
+TypeORM now requires **MongoDB server 6.0 or later** and the **`mongodb` Node.js driver v7 or later** (`^7.0.0`). Support for the `mongodb` driver v5/v6 has been dropped.
```
pkuczynski: Do we support server 6.0? Its EOL...

naorpeled: Theoretically yes, so I wasn't sure whether to write that we support from 7+ or not

cc @alumni

package.json
```diff
@@ -143,7 +143,7 @@
     "is-ci": "^4.1.0",
     "lint-staged": "^16.2.7",
     "mocha": "^11.7.5",
-    "mongodb": "^6.21.0",
+    "mongodb": "^7.0.0",
```
pkuczynski: Latest is `7.1.0` and we should be using latest I would say for testing..


```suggestion
    "mongodb": "^7.1.0",
```

docs/docs/guides/8-migration-v1.md
```diff
@@ -118,6 +118,14 @@ The following MongoDB connection options have been removed:
...
 | `wtimeout`              | Use `writeConcern: { wtimeoutMS: 2500 }` instead     |
 | `wtimeoutMS`            | Use `writeConcern: { wtimeoutMS: 2500 }` instead     |
+| `promoteBuffers`        | Remove — not available in MongoDB Driver v7          |
+| `promoteLongs`          | Remove — not available in MongoDB Driver v7          |
+| `promoteValues`         | Remove — not available in MongoDB Driver v7          |
+| `raw`                   | Remove — not available in MongoDB Driver v7          |
+
+### `stats()` method removed
+
+The `stats()` method has been removed from `MongoQueryRunner`, `MongoEntityManager`, and `MongoRepository`. The underlying `collStats` command was deprecated in MongoDB server 6.2 and the `Collection.stats()` method was removed in MongoDB Driver v7.
```
pkuczynski, line 124:
```suggestion
### `stats()` method removed

The `stats()` method has been removed from `MongoQueryRunner`, `MongoEntityManager`, and `MongoRepository`. The underlying `collStats` command was deprecated in MongoDB server 6.2 and the `Collection.stats()` method was removed in MongoDB Driver v7.

Use the [`$collStats`](https://www.mongodb.com/docs/manual/reference/operator/aggregation/collStats/) aggregation stage instead. Note that the response shape is different — properties like `count`, `size`, and `storageSize` are nested under `storageStats` rather than at the top level.

```typescript
// Before
const stats = await mongoRepository.stats()
console.log(stats.count)
console.log(stats.size)
console.log(stats.totalIndexSize)

// After — use $collStats aggregation stage
const [stats] = await dataSource.mongoManager
    .aggregate(MyEntity, [{ $collStats: { storageStats: {} } }])
    .toArray()
console.log(stats.storageStats.count)
console.log(stats.storageStats.size)
console.log(stats.storageStats.totalIndexSize)
```

```

File: docs/docs/guides/8-migration-v1.md (hunk 3 above)
Comment (alumni) [reply docs/docs/guides/8-migration-v1.md]: Officially we don't mention anything about server versions, and in case of Mongo where there's no SQL, we use directly the client most of the time, so internal TypeORM logic does not depend that much on the server, but rather on the client.

File: src/driver/mongodb/MongoDataSourceOptions.ts [hunk 6]
```diff
@@ -145,26 +145,6 @@ export interface MongoDataSourceOptions extends BaseDataSourceOptions {
      */
     readonly pkFactory?: any
 
-    /**
-     * when deserializing a Binary will return it as a node.js Buffer instance.
-     */
-    readonly promoteBuffers?: boolean
```
alumni, line 151: Are these no longer working in client v7?

docs/docs/guides/8-migration-v1.md
pkuczynski: True, good point. We should remove the server part...

docs/docs/guides/8-migration-v1.md
pkuczynski: @naorpeled we could also implement `stats` outselves if we want, but not sure if its worth it...

naorpeled: Yeah, I agree

package.json
naorpeled: Done in dca772a — updated to `^7.1.0`.

docs/docs/guides/8-migration-v1.md
naorpeled: Done in 2af12e0, removed server version mentions from both `mongodb.md` and `8-migration-v1.md`.

pkuczynski (approved):

src/driver/mongodb/MongoDataSourceOptions.ts
naorpeled: No, it was my bad, fixed, thanks for pointing it out!

PR #12209, opened 2026-03-13
fix(query-builder): reject semicolons in raw SQL expression methods

pkuczynski:
- Add runtime validation to reject semicolons in `select`, `addSelect`, `groupBy`, `addGroupBy`, `orderBy`, and `addOrderBy` to prevent statement stacking attacks (e.g. `qb.groupBy("1; DELETE FROM post;")`)
- Unskip regression tests and assert that semicolon-containing inputs throw instead of executing


gioboa (approved):

src/query-builder/SelectQueryBuilder.ts
```diff
@@ -155,6 +155,12 @@ export class SelectQueryBuilder<Entity extends ObjectLiteral>
     ): SelectQueryBuilder<Entity> {
         this.expressionMap.queryType = "select"
         if (Array.isArray(selection)) {
+            for (const s of selection) {
+                if (s.includes(";"))
+                    throw new TypeORMError(
+                        `Semicolons are not allowed in select() to prevent SQL injection. Use parameter binding instead.`,
+                    )
```
pujux: this code is duplicated quite often in this PR, can we get a utility function for this and just call it where needed?

src/query-builder/SelectQueryBuilder.ts
```diff
@@ -3100,7 +3143,7 @@ export class SelectQueryBuilder<Entity extends ObjectLiteral>
             `COUNT(DISTINCT(` +
             primaryColumns
                 .map((c) => `${distinctAlias}.${this.escape(c.databaseName)}`)
-                .join(" || '|;|' || ") +
+                .join(" || '|:|' || ") +
```
pujux, line 3158: I don't understand what this is for

src/query-builder/SelectQueryBuilder.ts
```diff
@@ -161,6 +161,12 @@ export class SelectQueryBuilder<Entity extends ObjectLiteral>
     ): SelectQueryBuilder<Entity> {
```
smith-xyz, line 162:
```
createQueryBuilder().update(Post).select("post.id; …")
```

just looking at this path here: 

https://github.com/typeorm/typeorm/blob/b4e6f1b7b2fb41dbe7f4d5c3d58b6885fee20fa0/src/query-builder/QueryBuilder.ts#L190-L208

wondering if there is an override on this through this path

smith-xyz:
@pkuczynski I like this direction. It’s new territory for TypeORM: there aren’t many comparable runtime rules from a quick search through, so this stands out compared to the usual mix of parameters, identifier escaping, and the sql tag.

I’m wondering whether we should treat this as the first item in an optional “strict / security guardrails” mode (similar in spirit to FIPS-style switches elsewhere)—something users can turn on or off as a group, rather than only ever adding one-off throws. That flag would mostly cover this change and future optional checks, not a long list of behavior that already exists in the codebase.

No strong opinion on default yet; worth discussing secure-by-default vs opt-in once we know how many follow-ons we actually want.

src/query-builder/SelectQueryBuilder.ts
smith-xyz: I presume this is for removing ambiguity without the use of |;|

src/query-builder/SelectQueryBuilder.ts
pkuczynski: `.select("post.id; …”)` does not seem like anything anyone migth want be doing on purpose, is it?

pkuczynski:
> @pkuczynski I like this direction. It’s new territory for TypeORM: there aren’t many comparable runtime rules from a quick search through, so this stands out compared to the usual mix of parameters, identifier escaping, and the sql tag.

We already merged few, this is the last one from the series. You can see more in https://dev.typeorm.io/docs/guides/release-notes-v1#security-fixes

> I’m wondering whether we should treat this as the first item in an optional “strict / security guardrails” mode (similar in spirit to FIPS-style switches elsewhere)—something users can turn on or off as a group, rather than only ever adding one-off throws. That flag would mostly cover this change and future optional checks, not a long list of behavior that already exists in the codebase.
>
> No strong opinion on default yet; worth discussing secure-by-default vs opt-in once we know how many follow-ons we actually want.

Does not seem to me like security should be an opt-in/out thing :)

src/query-builder/SelectQueryBuilder.ts
pkuczynski: Yes, exactly. This code was there before, its a split separator used elsewhere. I only changed from `;` to `:` to not clash with sql injection protection.

src/query-builder/SelectQueryBuilder.ts
pkuczynski: I would rather say we have custom set of errors and in this case it woult be throw new `TypeORMSqlInjectionError(’select()’)`. But this is not a common pratice at the moment, so I would pospone this to a global change and do it as a separate thing in v2

smith-xyz:
> > @pkuczynski I like this direction. It’s new territory for TypeORM: there aren’t many comparable runtime rules from a quick search through, so this stands out compared to the usual mix of parameters, identifier escaping, and the sql tag.
> 
> We already merged few, this is the last one from the series. You can see more in https://dev.typeorm.io/docs/guides/release-notes-v1#security-fixes
> 
> > I’m wondering whether we should treat this as the first item in an optional “strict / security guardrails” mode (similar in spirit to FIPS-style switches elsewhere)—something users can turn on or off as a group, rather than only ever adding one-off throws. That flag would mostly cover this change and future optional checks, not a long list of behavior that already exists in the codebase.
> > No strong opinion on default yet; worth discussing secure-by-default vs opt-in once we know how many follow-ons we actually want.
> 
> Does not seem to me like security should be an opt-in/out thing :)

Ahh thanks for linking to that doc. 

Agreed on default-on for this. The configurable idea was mostly about future checks and migration pain, not disabling fixes—e.g. a valve if the ORM’s security rules ever feel too opinionated for a general-purpose layer.

OSA413 (approved):
Looks good. Fix remaining issues (if they are relevant) and good to go

smith-xyz (approved):
oops yeah, second @OSA413 it looks good to me as well

alumni:
To be reverted.

Motivation: `QueryBuilder` is just a helper allows the user to write any kind of raw SQL. No validation should be done for the raw SQL. The user should be able to use `;` if they wish, e.g. in SQL string literals.

This is similar to how other query builders are behaving e.g. `knex`.

Validation should still be done for typed QB method parameters like offset/limit/order by direction, etc. This should not change.

pkuczynski:
> Motivation: `QueryBuilder` is just a helper allows the user to write any kind of raw SQL. No validation should be done for the raw SQL. The user should be able to use `;` if they wish, e.g. in SQL string literals.

For raw sql queries we offer `.query` method. Query Builder is not exactly meant to be used as a raw query executor.

PR #12216, opened 2026-03-15
ci(publish): use node 20 for v0.3 nightly builds

pkuczynski:
The v0.3 branch has an older `better-sqlite3` that requires node-gyp with C++17. Node 24 ships V8 headers that require C++20, causing compilation failures in the nightly publish.

Add per-branch node version to the publish matrix (node 24 for master, node 20 for v0.3).

### DIFF
--- a/.github/workflows/publish-package.yml
+++ b/.github/workflows/publish-package.yml
@@ -60,7 +60,7 @@ jobs:
               echo 'matrix=[{"target":"release"}]' >> $GITHUB_OUTPUT
               ;;
             schedule)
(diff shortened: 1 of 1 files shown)


alumni:
IMO this is unnecessarily overcomplicated and I'd rather not merge it. I don't doubt that it works, but I'd rather not spend the entire afternoon trying to understand this PR, and more time later trying to simplify it.

I'd just remove the nightly builds for 0.3, we are not merging anything to that branch until a release. We can also handle the next 0.3 release manually if that is blocked somehow.

pkuczynski:
It does not look too complicated for me to be honest and sooner or later we might be again in this situation when for one of the future releases and node incompatibility

pkuczynski:
@alumni have a look again. I accidentially had changes from #12202 in this branch. Sorry for the confusion. Now its clean.

PR #12217, opened 2026-03-15
fix(query-builder): validate orderBy condition values at runtime

pkuczynski:
`OrderByCondition` values are concatenated directly into the ORDER BY clause without runtime validation. Although TypeScript constrains them to `"ASC" | "DESC"`, at runtime a caller could pass arbitrary strings like `"ASC; DELETE FROM post;"` to achieve statement stacking.

- Add `validateOrderByCondition()` that checks values are strictly `"ASC"`/`"DESC"` and nulls are `"NULLS FIRST"`/`"NULLS LAST"`
- Call it from `orderBy()` when receiving an `OrderByCondition` object
- Add defense-in-depth regex check in `createOrderByExpression()` before SQL concatenation
- Add tests for invalid order direction, invalid nulls option, and valid values

Implements qodo feedback from #12209.


src/query-builder/SelectQueryBuilder.ts
```diff
@@ -2590,6 +2591,10 @@ export class SelectQueryBuilder<Entity extends ObjectLiteral>
                             : (orderBys[columnName] as any).order +
                               " " +
                               (orderBys[columnName] as any).nulls
+                    if (/[;'"\\]/.test(orderValue))
```
alumni, line 2599: `ORDER BY LENGTH("CONTENT")` is blocked by this.

alumni:
IMO this change is blocking the users from using order by in a lot of cases.

We provide `QueryBuilder`s to allow users to do more advanced queries that are not possible via the `Repository`/`EntityManager` APIs, but at the same time allow them to write easy to refactor code (instead of plain SQL). We do not provide any validation on purpose for them.

I recommend closing without merging.

pkuczynski:
Hey Lucian, thanks for the review — that's a good catch and a valid concern! However, the validation here only applies to the **value side** of the `OrderByCondition` object (i.e., the `"ASC"`, `"DESC"`, and `{ order, nulls }` part), not the **keys** (column names/expressions). Users can still pass any expression they want as the orderBy key — that's completely unrestricted.

The issue this PR addresses is that passing something like `{ "post.id": "ASC; DELETE FROM post;--" }` would embed the malicious string directly into the generated SQL. The fix validates that the direction values are one of `ASC`/`DESC` and nulls are `NULLS FIRST`/`NULLS LAST`.

I've also added tests to cover expression-based orderBy keys to make sure they keep working across all drivers — you can see those [here](https://github.com/typeorm/typeorm/pull/12217/files/24e51f5f386f8d43ec75c2a102692f646ff4055f..34e690d8dde3d5efc5fcec6808d72b9492968af6).

src/query-builder/SelectQueryBuilder.ts
pkuczynski: No, actually its not, as the validation here only applies to the value side of the `OrderByCondition` object

PR #12218, opened 2026-03-15
fix: use ALTER COLUMN instead of DROP+ADD for column type/length changes

CelebrityPunks:
## Summary

Fixes #3357

When a column property like `type` or `length` changes (e.g. `varchar(50)` to `varchar(51)`), TypeORM generates destructive SQL:

```sql
```

This causes **data loss** because dropping a column deletes all data in it. The correct approach is:

```sql
```

### What this PR does

- **PostgreSQL**: Uses `ALTER COLUMN ... TYPE` with a `USING` clause for type conversions
- **MySQL/MariaDB**: Uses `ALTER TABLE ... MODIFY` which preserves data
- **Aurora MySQL**: Same as MySQL
- **CockroachDB**: Uses `ALTER COLUMN ... TYPE` (same syntax as PostgreSQL)
- **Oracle**: Uses `ALTER TABLE ... MODIFY` which preserves data
- **Spanner**: Uses `ALTER COLUMN ... TYPE`

`DROP + ADD` is now only used for truly incompatible changes where `ALTER` is not possible:
- Changing to/from generated (STORED/VIRTUAL) columns
- Changing generated column expressions
- Changing identity column generation (Oracle, MySQL)

### What stays the same

- Precision/scale changes already used `ALTER COLUMN ... TYPE` — unchanged
- Enum changes already used `ALTER COLUMN ... TYPE` — unchanged
- Nullable, default, comment, primary key changes — unchanged
- Column rename — unchanged
- SQLite — unchanged (SQLite recreates the entire table for any column change, which is the only supported approach)

## Test plan

- Added test `test/github-issues/3357/` that:
  - [x] Changes column length and verifies data is preserved
  - [x] Changes column type and verifies data is preserved
  - [x] Verifies no `DROP COLUMN` SQL is generated for length changes
- Existing `test/functional/query-runner/change-column.test.ts` continues to work (column changes + revert via `executeMemoryDownSql`)


pkuczynski:
Hey @CelebrityPunks, thanks for tackling this! We've decided to move forward with #11620 which covers more drivers (sqlite, spanner, sap) and adds a safety guard that falls back to DROP/ADD for cross-family type changes. Closing this one in favor of #11620 — your review and feedback there would be very welcome 🙏

PR #12220, opened 2026-03-15
refactor(query-builder)!: remove logQuery method

pkuczynski:
The `logQuery()` method (previously `printSql`) was redundant — all executed queries are already logged automatically through the configured logger when the query is actually run via QueryRunner. The method was also misleading: it was gated by the `logging` DataSource option, so it silently did nothing when logging was disabled (the default).

- Remove `logQuery()` from `QueryBuilder`
- Update query builder docs to recommend `getSql()` / `getQueryAndParameters()` for SQL inspection and `logging: ["query"]` for automatic query logging
- Update migration guide: change "printSql renamed to logQuery" to "printSql removed" with alternatives


gioboa (approved):

docs/docs/query-builder/1-select-query-builder.md
```diff
@@ -798,18 +798,30 @@ const sql = dataSource
...
-    .getMany()
+
+console.log(qb.getSql())
+// SELECT ... WHERE "user"."firstName" = $1 OR "user"."lastName" = $2
+
+const users = await qb.getMany()
 ```
 
-This query will return users and log the used sql statement through the configured logger.
+`getSql()` returns the SQL with parameter placeholders. `getQuery()` returns the same SQL string. To also see the parameter values, use `getQueryAndParameters()` which returns a `[sql, parameters]` tuple.
```
pujux, line 815: do `getSql()` and `getQuery()` really return the *same* SQL string?

alumni: No, they do not :)

There's a slight difference: in `getSql` the query parameters are substituted.

alumni:
@naorpeled I'm going to leave this up to you, since you looked into it before. 

I remember we also discussed the utility of this function when we chatted about renaming it from `logSql` to `logQuery`. IIRC the argument against this removal was that you can create a partial `QueryBuilder` in one place and it's an easy alternative to `console.log` when developing.

Feel free to directly close or merge this PR.

naorpeled:
> @naorpeled I'm going to leave this up to you, since you looked into it before.
> 
> I remember we also discussed the utility of this function when we chatted about renaming it from `logSql` to `logQuery`. IIRC the argument against this removal was that you can create a partial `QueryBuilder` in one place and it's an easy alternative to `console.log` when developing.
> 
> Feel free to directly close or merge this PR.

At the same time you can also use getSql or getQuery so it seems indeed redundant, approving and merging this.

Thanks @pkuczynski !

naorpeled (approved):

PR #12222, opened 2026-03-16
style(eslint): enable unicorn/prefer-string-replace-all rule

pkuczynski:
Add `eslint-plugin-unicorn` and enable the `prefer-string-replace-all` rule to enforce `String#replaceAll()` over `String#replace()` with global regex.

This addresses the `typescript:S7781` SonarQube rule ("Prefer `String#replaceAll()` over `String#replace()`") which shows up frequently across PRs. Rather than keep ignoring it in SonarQube reports, we enforce it at the lint level so it's caught before code reaches CI.

From the [rule documentation](https://github.com/sindresorhus/eslint-plugin-unicorn/blob/main/docs/rules/prefer-string-replace-all.md):

> The `String#replaceAll()` method is both faster and safer as you don't have to use a regex and remember to escape it if the string is not a literal. And when used with a regex, it makes the intent clearer.

Auto-fixed 71 violations across 31 files.


alumni:
`eslint-plugin-unicorn` tends to have a bad reputation because it's extremely opinionated. We could replace all `replace` with `replaceAll`, but I'd avoid an external eslint rule for it

pkuczynski:
> `eslint-plugin-unicorn` tends to have a bad reputation because it's extremely opinionated. We could replace all `replace` with `replaceAll`, but I'd avoid an external eslint rule for it

Sindre is indeed very opinionated person, that's why I did not enabled the `recommended` config, but the singe rule. This can't harm us in any way.

If we agree with SonarQube advice, this will help us prevent such _issues_ in the future and get better reports. In my last PR I got over 40 reported issues where most of them was about exactly this case. This creates noise I want to avoid, and which we can easily remove with this eslint rule.

src/commands/MigrationGenerateCommand.ts
```diff
@@ -316,7 +316,9 @@ ${downSqls.join(`
     protected static prettifyQuery(query: string) {
         const formattedQuery = format(query, { indent: "    " })
         return (
-            "\n" + formattedQuery.replace(/^/gm, "            ") + "\n        "
+            "\n" +
+            formattedQuery.replaceAll(/^/gm, "            ") +
```
pujux, line 323: no need for `g` modifier if we use replaceAll

package.json
```diff
@@ -131,6 +131,7 @@
     "eslint": "^10.0.2",
     "eslint-plugin-chai-friendly": "^1.1.0",
     "eslint-plugin-jsdoc": "^62.7.1",
+    "eslint-plugin-unicorn": "^63.0.0",
```
pujux: so we're adding this 100+ rule eslint plugin for a single rule? is there something lighter that has string-replace-all?

pujux:
```markdown
  ┌────────────────────────────────────────┬──────────┬────────────┬──────────────────────────────────────────────────────────────────┐                                                                                     
  │                  Rule                  │ Severity │ Violations │                               Note                               │
  ├────────────────────────────────────────┼──────────┼────────────┼──────────────────────────────────────────────────────────────────┤                                                                                       ├────────────────────────────────────────┼──────────┼────────────┼──────────────────────────────────────────────────────────────────┤                                                                                     
  │ unicorn/prefer-includes                │ error    │ ~96        │ .indexOf() !== -1 → .includes() — clearer intent, same semantics │                                                                                     
  ├────────────────────────────────────────┼──────────┼────────────┼──────────────────────────────────────────────────────────────────┤                                                                                     
  │ unicorn/prefer-number-properties       │ warn     │ ~28        │ isNaN() → Number.isNaN() — avoids coercion gotchas               │                                                                                     
  ├────────────────────────────────────────┼──────────┼────────────┼──────────────────────────────────────────────────────────────────┤                                                                                     
  │ unicorn/prefer-string-starts-ends-with │ warn     │ ~8         │ /^prefix/.test(s) → s.startsWith() — simpler, faster             │
  └────────────────────────────────────────┴──────────┴────────────┴──────────────────────────────────────────────────────────────────┘
```

maybe it makes sense enabling these as well

alumni: I would not enable this plugin or any rule from it. It is a very opinionated plugin and will enforce rules not everyone will be familiar with. Yes for the recommended setup in typescript-eslint, no for unicorn.

We can still do some fixes based on the rules (I also don't like regular expressions in replace or other outdated function calls, but I would not enforce it, this is micromanaging)

alumni (dismissed):
I'd like to avoid the use of unicorn, otherwise we can do the cleanup :)

I'm not sure why we have SonarQube when we have eslint. Maybe we should disable SonarQube or certain checks that are usually handled by Eslint.

package.json
pkuczynski: This PR does not enable recommended setup from unicorn. It is using a basic rule enforcing `replaceAll` instead of `replace`, which is also reported as issue by SonarQube.

There are many useful rules in this plugin as @pujux mentioned. This is one of them. That does not mean that all of them are great.

This is one of the misunderstandings on the difference between enabling individual rules and _recommended_ configs.

Do you disagree with forcing `replaceAll`? Looking at your other comment, you agree. So tell me how would you like to enforce this in a different way?

gioboa (approved):

src/commands/MigrationGenerateCommand.ts
smith-xyz: I think that is a typeerror ironically: 

```bash
node -e '"hello".replaceAll(/[a-z]/, "")'
[eval]:1
"hello".replaceAll(/[a-z]/, "")
        ^

TypeError: String.prototype.replaceAll called with a non-global RegExp argument
    at String.replaceAll (<anonymous>)
    at [eval]:1:9
    at runScriptInThisContext (node:internal/vm:219:10)
    at node:internal/process/execution:451:12
    at [eval]-wrapper:6:24
    at runScriptInContext (node:internal/process/execution:449:60)
    at evalFunction (node:internal/process/execution:283:30)
    at evalTypeScript (node:internal/process/execution:295:3)
    at node:internal/main/eval_string:71:3

Node.js v24.12.0
```

smith-xyz (approved):
I think this is fine since sonarqube has taken the opinion, so approving. However, want to note positives/negatives I see:

Positives:

- SonarQube stops complaining
- String-arg cases read slightly cleaner (even though we're changing usage with regex)

Negatives:

- Regex replaceAll introduces a runtime TypeError risk that TS can't catch

PR #12224, opened 2026-03-16
build(gulp): remove gulpclass dependency and modernize build pipeline

pkuczynski:
**Build modernization:**
- Convert gulpfile from `gulpclass` decorators to plain functions with `gulp.series()`/`gulp.parallel()`
- Replace `gulp-typescript` + `gulp-sourcemaps` with direct `tsc` calls using dedicated tsconfig files
- Run `packageCompile` and `browserCompile` in parallel (previously sequential)
- Upgrade Gulp from v4 to v5

**Removed dependencies:** `gulpclass`, `gulp-typescript`, `gulp-sourcemaps`, `gulp-replace`, `gulp-shell`, `@types/gulp-sourcemaps`, `is-ci`

**New files:**
- `tsconfig.browser.json` — browser build config (`lib: ["es2023", "dom"]`)
- `tsconfig.package.json` — package build config

**CI cleanup:**
- Move package.json cleanup logic from the publish workflow into the gulpfile (`packageCopyPackageFile`)
- Update path filters in `tests.yml` and `preview.yml` for new tsconfig files
- Fix `gulpfile.js` → `gulpfile.ts` typo in tests.yml

**`collective` removed from package.json:** The `collective` field was an npm v5-era convention for post-install funding banners. npm has since replaced it with the standard `funding` field, which is already present in package.json. The `collective` field was a no-op.

**Build speed (wall-clock):**
| | Before | After |
|---|---|---|
| `pnpm run package` | ~18s | ~10s |

The ~45% speedup comes from parallel compilation — browser and package builds now run concurrently instead of sequentially.


pkuczynski:
> ### `private: true` will block every publish

False positive. Our package.json does not have `private` field :)

> ### README.md not copied into build/package
> `packageCopyReadme` was dropped entirely, so the published tarball won't have a README. The npm page for typeorm would show a blank description section. For a package this size that's not great.
> 
> The old `typescript` → `javascript` fenced block replacement is cosmetic and can be skipped, but the copy itself needs to come back.

Good catch! Let me add it...

Issue #11123, opened 2024-11-08
Cannot use SNAPSHOT isolation level in SQL Server


penttin5:
### Issue description

Cannot use SNAPSHOT isolation level in SQL Server

### Expected Behavior

When default isolation level is set to `SNAPSHOT` like this:
```ts
const dataSourceOptions: SqlServerConnectionOptions = {
  type: 'mssql',
  options: {
    isolation: 'SNAPSHOT',
```
A transaction run with default isolation level like this:
```ts
  this.dataSource.transaction(async () => {
```
 should be run with SNAPSHOT isolation level.

Also, it should be possible to set SNAPSHOT isolation level per transaction:
```ts
  this.dataSource.transaction(IsolationLevel.SNAPSHOT, async () => {
```

### Actual Behavior

Setting isolation level to `SNAPSHOT` in `SqlServerConnectionOptions` does not affect the default isolation level in transactions.

If the default isolation level is set to `SNAPSHOT` like this:
```ts
const dataSourceOptions: SqlServerConnectionOptions = {
  type: 'mssql',
  options: {
    isolation: 'SNAPSHOT',
```

and then a transaction is run with default isolation level like this:
```ts
  this.dataSource.transaction(async () => {
```
the transaction is not run with `SNAPSHOT` isolation level. Instead, it seems to default to `READ_COMMITTED` defined in `mssql/transaction.ts`

The isolation level cannot be set to SNAPSHOT per transaction either, since `DataSource.transaction` uses  `types/IsolationLevel.ts` which does not contain `SNAPSHOT`.

The isolation level can be forced to 'SNAPSHOT' like this:
```ts
  const SNAPSHOT = 'SNAPSHOT' as unknown as IsolationLevel
  this.dataSource.transaction(SNAPSHOT, async () => {
```

but that does not work either, since `SqlServerQueryRunner.convertIsolationLevel` does not know how to handle it:
```ts
convertIsolationLevel(isolation: IsolationLevel) {
  const ISOLATION_LEVEL = this.driver.mssql.ISOLATION_LEVEL
  switch (isolation) {
    case "READ UNCOMMITTED":
      return ISOLATION_LEVEL.READ_UNCOMMITTED
    case "REPEATABLE READ":
      return ISOLATION_LEVEL.REPEATABLE_READ
    case "SERIALIZABLE":
      return ISOLATION_LEVEL.SERIALIZABLE
    case "READ COMMITTED":
    default:
[...]

PR #12231, opened 2026-03-17
fix(mssql)!: rename dataSource isolation options and assign proper values during `createPool`

Cprakhar:
Related #11123

### Description of change
This PR brings a breaking change to the `SqlServerDataSourceOptions`.

**Changes:**
- Renames `options.isolation` to `options.isolationLevel` as in the first place mssql driver requires `isolationLevel` option, not `isolation`
- Changes the `options.isolationLevel` and `options.connectionIsolationLevel` value format to match and be consistent with the `IsolationLevel` type
- In the `createPool` method of `SqlServerDriver`, convert the isolation option strings to an enum from the `tedious` package
- Added `SNAPSHOT` isolation level for per-transaction for MSSQL

**Why needed:**
```typescript
// Before
new DataSource({
  type: "mssql",
  options: {
    isolation: "READ_COMMITTED",
    connectionIsolationLevel: "READ_COMMITTED",
  }
})
```
```
TypeError: The "config.options.isolationLevel" property must be of type number. Received type string (READ_COMMITTED)
      at assertValidIsolationLevel (node_modules/.pnpm/tedious@19.2.1_@azure+core-client@1.10.1/node_modules/tedious/src/transaction.ts:35:11)
      at new Connection (node_modules/.pnpm/tedious@19.2.1_@azure+core-client@1.10.1/node_modules/tedious/src/connection.ts:1574:34)
      at /home/ghostblasius/typeorm/node_modules/.pnpm/mssql@12.2.0_@azure+core-client@1.10.1/node_modules/mssql/lib/tedious/connection-pool.js:78:19
      at new Promise (<anonymous>)
      at ConnectionPool._poolCreate (node_modules/.pnpm/mssql@12.2.0_@azure+core-client@1.10.1/node_modules/mssql/lib/tedious/connection-pool.js:67:12)
      at ConnectionPool._connect (node_modules/.pnpm/mssql@12.2.0_@azure+core-client@1.10.1/node_modules/mssql/lib/base/connection-pool.js:445:10)
      at ConnectionPool.connect (node_modules/.pnpm/mssql@12.2.0_@azure+core-client@1.10.1/node_modules/mssql/lib/base/connection-pool.js:412:12)
      at /home/ghostblasius/typeorm/src/driver/sqlserver/SqlServerDriver.ts:1241:37
      at new Promise (<anonymous>)
      at SqlServerDriver.createPool (src/driver/sqlserver/SqlServerDriver.ts:1226:16)
      at SqlServerDriver.connect (src/driver/sqlserver/SqlServerDriver.ts:303:38)
      at DataSource.initialize (src/data-source/DataSource.ts:238:27)
```

```typescript
// After
new DataSource({
  type: "mssql",
  options: {
    isolationLevel: "READ COMMITTED",
    connectionIsolationLevel: "READ COMMITTED",
  }
})
```

### Pull-Request Checklist

- [x] This pull request links relevant issues as `Related #11123`


src/data-source/DataSource.ts
```diff
@@ -463,12 +463,12 @@ export class DataSource {
         runInTransaction: (entityManager: EntityManager) => Promise<T>,
     ): Promise<T>
     async transaction<T>(
-        isolationLevel: IsolationLevel,
+        isolationLevel: SqlServerIsolationLevel,
```
alumni: `DataSource` should not use db-specific types. `IsolationLevel` was probably fine.

src/driver/types/IsolationLevel.ts
```diff
@@ -3,3 +3,5 @@ export type IsolationLevel =
     | "READ COMMITTED"
     | "REPEATABLE READ"
     | "SERIALIZABLE"
+
+export type SqlServerIsolationLevel = IsolationLevel | "SNAPSHOT"
```
alumni: I'm thinking there's no need for a separate type. I'd either not add "snapshot" or add it everywhere and make it throw on runtime.

If I'm using e.g. postgres, I'd be confused why do I see mssql types.

Cprakhar: > If I'm using e.g. postgres, I'd be confused why do I see mssql types.

But this is the same for `sqlite` and `oracle`: they support partial isolation levels, but see the types they don't support.

I will add `SNAPSHOT` in the IsolationLevel type and throw at runtime.

alumni: > I'd be confused why do I see mssql types.

Yeah, I meant the type name, sorry :)

You also add a comment that it's mssql specific, it will help us remember.

test/functional/transaction/database-specific-isolation/full-isolation-support.test.ts
```diff
@@ -14,180 +14,66 @@ describe("transaction > transaction with full isolation support", () => {
...
+        "REPEATABLE READ",
+        "READ COMMITTED",
+    ] as const
+
+    for (const isolationLevel of isolationLevels) {
+        it(`should execute all operations in a single transaction with ${isolationLevel} isolation level`, () =>
+            Promise.all(
+                dataSources.map(async (dataSource) => {
+                    // SAP does not support READ UNCOMMITTED isolation level
+                    if (dataSource.driver.options.type === "sap") return
```
alumni: `&& isolationLevel === "READ UNCOMMITTED"`, otherwise all levels will be skipped.

test/functional/transaction/database-specific-isolation/sap-isolation-support.test.ts
```diff
@@ -14,180 +14,66 @@ describe("transaction > transaction with full isolation support", () => {
...
-            }),
-        ))
+    const isolationLevels = [
+        "READ UNCOMMITTED",
+        "SERIALIZABLE",
+        "REPEATABLE READ",
+        "READ COMMITTED",
+    ] as const
+
+    for (const isolationLevel of isolationLevels) {
```
alumni: I'm not sure if the original intention was to have the same test for all isolation levels. If you look at the implementation, it's not doing much.

Probably they should each work differently with two simultaneous transactions doing operations on the same table, maybe with some `scheduler.wait()` to make sure the order of statements is intended. But that might be a bit of work, not for this PR.

Something simple: we could also issue queries to check the isolation level, e.g.:
- `SELECT CURRENT_TRANSACTION_ISOLATION_LEVEL FROM SYS.DUMMY` in HANA.
- `SHOW TRANSACTION ISOLATION LEVEL` in Postgres
- might not work in MySQL 🤷🏻

Cprakhar: - `SELECT isolation_level
FROM performance_schema.events_transactions_current
WHERE state = 'ACTIVE'` in MySQL (Mariadb does not list anything)

Tested locally, result:
- MySQL ✅
- MariaDB ❌

Cprakhar:
For old versions of MySQL/MariaDB in the test:
To control transaction event collection at runtime, update the [setup_instruments](https://dev.mysql.com/doc/refman/5.7/en/performance-schema-setup-instruments-table.html) and [setup_consumers](https://dev.mysql.com/doc/refman/5.7/en/performance-schema-setup-consumers-table.html) tables:
```sql
UPDATE performance_schema.setup_instruments
SET ENABLED = 'YES', TIMED = 'YES'
WHERE NAME = 'transaction';

UPDATE performance_schema.setup_consumers
SET ENABLED = 'YES'
WHERE NAME LIKE 'events_transactions%';
```

Cprakhar:
- **MariaDB** is not returning any info related to isolation from the `performance_schema.events_transactions_current` (so it is not included in the `enabledDrivers` option).
- **SQL Server** returning `read committed` in case when setting `isolationLevel` or `connectionIsolationLevel` in the `DataSource` options (failing tests).
- Didn't find any way to get info on isolation level from **Oracle**.


@alumni Please take a look, as it would be good if merged because it is breaking change before releasing v1.0.0

alumni (approved):

gioboa (approved):

gioboa:
Thanks 🙏 @Cprakhar

Cprakhar:
> * **SQL Server** returning `read committed` in case when setting `isolationLevel` or `connectionIsolationLevel` in the `DataSource` options (failing tests).

Should I remove the assertions from the mssql tests for `isolationLevel` and `connectionIsolationLevel`, as they are failing? I didn't get any info on why it returns `read committed` in all cases

pkuczynski (approved):

pkuczynski:
Thanks for the contribution @Cprakhar!

Before merging, I've made a few additional changes on top of your work:

- Extracted `validateIsolationLevel()` into a standalone function (`src/driver/validate-isolation-level.ts`) and added it to every QueryRunner's `startTransaction()`
- Made `supportedIsolationLevels` a `static readonly` property on each driver class with JSDoc linking to official DB docs
- Added `SNAPSHOT` to the global `IsolationLevel` type — unsupported drivers reject it at runtime
- Aurora query runners now throw when isolation level is passed (Data API doesn't support it)
- `convertIsolationLevel()` throws on unknown values instead of silently falling back to `READ COMMITTED`
- Added isolation level tests for all drivers (MySQL, PostgreSQL, CockroachDB, Oracle, SAP HANA, SQLite/sql.js, Spanner) covering both supported and unsupported levels
- Skipped DataSource-level MSSQL tests due to upstream pool limitation ([tediousjs/node-mssql#1483](https://github.com/tediousjs/node-mssql/issues/1483)) and documented it in [Known Issues](https://typeorm.io/microsoft-sqlserver#connection-pool-does-not-reset-isolation-level)
- Updated transaction docs with per-driver isolation level table
- Created a codemod to cover this breaking change, scoped to `type: "mssql"` config objects
- Switched test assertions to `expect()` style, fixed SAP column alias casing, narrowed `performance_schema` setup to MySQL only

PR #12250, opened 2026-03-20
feat: add findOrCreate method

### LINKED ISSUE #7301: Create FindOrCreate metod
## Feature Description

One method to find or create. Does not create a new record if exists one register. Does create a new record if not exists one register.

### The Problem

I'm always frustrated when need to create or search a record, requiring more than one request in the database.

### The Solution

Create a method to find one register in case don't find, make a register in table.

- [ ] `aurora-data-api`
- [ ] `aurora-data-api-pg`
- [ ] `better-sqlite3`
- [ ] `cockroachdb`
- [ ] `cordova`
- [ ] `expo`
- [ ] `mongodb`
- [ ] `mysql`
- [ ] `nativescript`
- [ ] `oracle`
- [x] `postgres`
- [ ] `react-native`
- [ ] `sap`
- [ ] `sqlite`
- [ ] `sqlite-abstract`
- [ ] `sqljs`
- [ ] `sqlserver`

- [X] Yes, I have the time, but I don't know how to start. I would need guidance.

TanayK07:
## Summary

Adds a first-class `findOrCreate()` method to EntityManager, Repository, and BaseEntity so users don't have to manually write `findOneBy() ?? save(create())` every time.

- **Returns** `[entity, created]` — the entity and a boolean indicating whether it was newly created
- **Options**: `where` (find conditions) + optional `create` (extra fields merged on creation)
- Follows TypeORM's existing three-level delegation: EntityManager → Repository → BaseEntity

Fixes #7301

## Changes

- `src/find-options/FindOrCreateOptions.ts` — new options interface
- `src/entity-manager/EntityManager.ts` — core `findOrCreate` logic (SELECT then INSERT)
- `src/repository/Repository.ts` — thin wrapper delegating to manager
- `src/repository/BaseEntity.ts` — static method delegating to repository
- `src/index.ts` — export `FindOrCreateOptions` type
- `test/functional/repository/find-or-create/` — 5 test cases
- `docs/` — updated EntityManager and Repository API docs

## Usage

```typescript
// Repository
const [user, created] = await userRepository.findOrCreate({
    where: { email: "timber@example.com" },
    create: { firstName: "Timber", lastName: "Saw" },
})

// EntityManager
const [user, created] = await manager.findOrCreate(User, {
    where: { email: "timber@example.com" },
    create: { firstName: "Timber", lastName: "Saw" },
})

// BaseEntity (Active Record)
const [user, created] = await User.findOrCreate({
    where: { email: "timber@example.com" },
    create: { firstName: "Timber", lastName: "Saw" },
})
```

## Test plan

- [x] Creates entity when it doesn't exist (returns `[entity, true]`)
- [x] Finds existing entity without creating duplicate (returns `[entity, false]`)
- [x] Merges `where` and `create` fields correctly on creation
- [x] Works without `create` option
- [x] Works via EntityManager API
- [x] `pnpm run compile` — builds without errors
- [x] `pnpm run format:ci` — formatting clean
- [x] `pnpm run lint` — no new lint errors


TanayK07:
Thanks for the review! Pushed fixes for the feedback (sonarqube is working pretty well here) : 

**1. `as any` cast** — removed, wasn't needed. My bad

**2. FindOperator in `where`** — fair point. In practice though, `findOrCreate` is meant to be used with plain equality conditions (same as Sequelize's `findOrCreate`). Using something like `MoreThan(18)` in `where` wouldn't really make sense here since you can't create an entity from an operator. I think this is fine as-is, but happy to add a note in the docs if that would help.(LMK , will create a new issue?) LMK

**3. `create` overriding `where`** — good catch, swapped the merge order so `where` always wins. Also added a test for this.

**4. Race condition (TOCTOU)** — yeah this is a real thing under concurrency, but I intentionally kept it simple since the issue was labeled `complexity: LMK if a follow up is needed, is a new issue better for that?

Also switched `Object.assign` to object spread per the SonarQube suggestion. The other SonarQube issue (type literal in BaseEntity) is actually the existing pattern used across all 40+ static methods in the file — changing just ours would be inconsistent. so I ignored that , i think that is okay?

alumni:
This method already exists and it's called `save`.

TanayK07:
> This method already exists and it's called `save`.

 Hi @alumni, thanks for the feedback! Could you clarify what you mean?
                                                                                                              
  From my understanding, save() and findOrCreate() solve different problems:                                
                                                                                                              
  - save(entity) takes an already-constructed entity and inserts or updates based on the primary key          
  - findOrCreate({ where, create }) searches by arbitrary conditions and only creates if no match is found,   
  returning [entity, created]                                                                                 
                                                                                                            
  For example, if you want to find a user by email and only create one if they don't exist, save() would      
  require you to first query manually, then conditionally create — which is exactly the boilerplate
  findOrCreate eliminates. It also returns a boolean    
                                                                                                            
and issue #7301 specifically requests it.
                                                                                                              
  That said, if I'm missing something about how save already covers this use case, I'd love to understand —   
  happy to close this if it's genuinely redundant!   Apologies if this is actually redundant , but It wasn't from my understanding

TanayK07:
Hi, sorry to ping you guys again, but should this be worked on or what would be the next steps? @pleerock

michaelbromley:

```
git fetch origin
```

PR #12255, opened 2026-03-22
ci(qodo): add optional chaining compliance rule

pkuczynski:
Add a Qodo compliance check for preferring optional chaining (`?.`) over non-null assertions (`!`) in tests after null checks.

SonarQube frequently flags non-null assertions as unnecessary type assertions (`typescript:S4325`). Adding this to the Qodo checklist catches it earlier during PR review, before it reaches the SonarQube quality gate.

```typescript
const result = await repository.findOne({ where: { id: 1 } })

// Before (non-null assertion)
expect(result!.name).to.equal("foo")

// After (optional chaining)
expect(result).to.not.be.null
expect(result?.name).to.equal("foo")
```


naorpeled (approved):

alumni (commented):
Qodo is not for static checks.

There are built-in eslint rules for that, which we might want to enable sometime in the future.

I would not merge this.

pkuczynski:
> Qodo is not for static checks.
> 
> There are built-in eslint rules for that, which we might want to enable sometime in the future.

Should we also then remove https://github.com/typeorm/typeorm/blob/master/pr_compliance_checklist.yaml#L14-L18? 

And how would you like to implement in eslint a check where in the test (only!) we should first assert for not null, and then use optional chaining? I really want to see an example of such config...

alumni:
How about eslint: `@typescript-eslint/prefer-optional-chain`. Might as well add `@typescript-eslint/prefer-nullish-coalescing` since it's similar.

Normally we would add the entire config at once ([@typescript-eslint/stylistic-type-checked](https://typescript-eslint.io/users/configs#stylistic-type-checked)), but our code base might not be there yet (the internal types are not always correct, so some rule that might be stylistic based on the type information, could end up breaking functionality)

I also noticed there's a lot of `.substr(0, 1)` which could be fixed with `@typescript-eslint/prefer-string-starts-ends-with`, in the same config.

Sonar also complains quite a lot about unnecessary type assertions, e.g. `@typescript-eslint/no-unnecessary-type-assertion` - this is in the recommended config, but was disabled on purpose since removing the assertions would cause more trouble than keeping them. This one in particular we should delay enabling for a while - especially with the new TS version (v6 was just released and we can't upgrade yet).

> Should we also then remove https://github.com/typeorm/typeorm/blob/master/pr_compliance_checklist.yaml#L14-L18?

That would great, that Qodo rule is so generic that I doubt is of any use. "project's established naming standards" - what's that mean? They are not described anywhere. `@typescript-eslint/naming-convention` is the way.

> And how would you like to implement in eslint a check where in the test (only!) we should first assert for not null, and then use optional chaining?

`typescript-eslint` uses the inferred types in each line. So if the assertion is properly typed, you'd get the right inference. Unfortunately, `chai` isn't that great. Inference is good with `assert` assertions, but we tend to use `expect` and `should` assertions which are closer to natural language.

gioboa (approved):

pkuczynski:
I've been enabling these rules incrementally in separate PRs:

- **`@typescript-eslint/prefer-optional-chain`** — #12331 (fixed 150 violations)
- **`@typescript-eslint/prefer-string-starts-ends-with`** — #12332 (codebase already clean, enabled as guard rail)

`@typescript-eslint/prefer-nullish-coalescing` was attempted but dropped — 642 violations that can't be auto-fixed due to intentional `||` usage with falsy primitives.

I also tackled the non-null assertion issue from SonarQube by replacing `findOne`/`getOne` with `OrFail` variants across the test suite in #12325.

Closing in favor of the PRs above.

PR #12258, opened 2026-03-24
fix(sqlserver): use ALTER COLUMN instead of DROP+ADD for type changes

DeekRoumy:
## Summary

Fixes #3357

When a column's type or length changes in SQL Server, TypeORM was generating destructive SQL:

```sql
ALTER TABLE [bug] DROP COLUMN [example]
ALTER TABLE [bug] ADD [example] varchar(51)
```

This causes **data loss**. SQL Server supports `ALTER TABLE ... ALTER COLUMN` for type/length changes:

```sql
ALTER TABLE [bug] ALTER COLUMN [example] varchar(51) NOT NULL
```

## What this PR does

- **SQL Server / MSSQL**: Uses `ALTER COLUMN ... new_type` for type/length changes, preserving data
- Properly handles DEFAULT constraints: SQL Server requires dropping the default constraint before `ALTER COLUMN` and re-adding it after
- `DROP + ADD` is only used for truly incompatible changes:
  - Changing IDENTITY property (SQL Server limitation)  
  - Changing computed column expressions (`asExpression` / `generatedType`)

## Relationship to PR #12218

PR #12218 addresses the same issue for PostgreSQL, MySQL, Aurora MySQL, CockroachDB, Oracle, and Spanner. This PR extends the same data-safe fix to **SQL Server (MSSQL)** which was not covered.

## Test plan

- Added `test/github-issues/3357/` with test entity and integration test
- Tests verify data is preserved after column length change
- Tests verify column definition is updated correctly

## What stays the same

- IDENTITY column changes → DROP + ADD (SQL Server limitation)
- Computed column expression changes → DROP + ADD
- Column renames, nullable, default, primary key changes → unchanged (handled by existing code below)
- SQLite → unchanged (table recreation is the only approach)

/claim #3357


DeekRoumy:
Thanks for the thorough automated review! I've addressed all the issues flagged by Qodo in the latest commit:

1. **skipEnum=true** — Both `buildCreateColumnSql` calls in the new `typeOrLengthChanged` block now pass `true` as the 5th argument to prevent invalid `CHECK` constraint fragments in `ALTER COLUMN` statements.

2. **Rename/type order fix** — The `ALTER COLUMN` and default-restore queries now use `oldColumn.name` (via a cloned column) since `sp_rename` runs afterward. This ensures the SQL references the column's current physical name.

3. **Duplicate ALTER COLUMN guard** — Added `!typeOrLengthChanged &&` guard before the `isColumnChanged` block, so a second `ALTER COLUMN` is never emitted when type/length already handled the change.

4. **Removed unused `TableColumn` import** from the test file.

5. **MySQL quoting fix** — Replaced raw SQL with `connection.getRepository(TestEntity).save()` and `.find()` so the test is driver-agnostic and works correctly on MySQL.

Re the test placement concern: the existing TypeORM codebase has many tests under `test/github-issues/` for regression cases tied to specific issues — this follows that same pattern. Happy to move it to `test/functional/` if that's the project preference.

Let me know if anything else needs attention!

pkuczynski:
Hey @DeekRoumy, thanks for this fix! We're going with #11620 which handles #3357 across all major drivers including SQL Server, and adds a safety guard for cross-family type changes. Closing in favor of #11620 — would love your feedback on the SQL Server changes there!

PR #12279, opened 2026-03-31
fix(postgres): normalize column type to avoid incorrect DROP+ADD

Robert0303-C:
## Summary

Fixes data loss during migration generation when changing column length on PostgreSQL.

## Problem

When running TypeORM's `synchronize: false` mode with PostgreSQL, changing a column's length (e.g., `varchar(50)` → `varchar(51)`) would incorrectly generate:

```sql
ALTER TABLE "user" ADD COLUMN "name" varchar(51);
```

This caused **data loss** because `DROP COLUMN` destroys data before the `ADD COLUMN` executes.

## Root Cause

In `PostgresQueryRunner.ts`, when loading column metadata from PostgreSQL's `regtype`, the type includes the length (e.g., `"character varying(50)"`). However, `normalizeType()` returns the type without length (`"character varying"`). This mismatch caused `findChangedColumns()` to detect a false "type change", triggering the destructive DROP+ADD instead of `ALTER COLUMN`.

## Solution

Added code in `PostgresQueryRunner.ts` to strip the length from column types when loading from the database. This ensures the in-memory `tableColumn.type` matches `normalizeType()` output, so type comparisons only compare base types (not lengths).

**Files changed:** `src/driver/postgres/PostgresQueryRunner.ts` (+23 lines)

## Example Behavior After Fix

| Scenario | Before Fix | After Fix |
|----------|-----------|----------|
| Change varchar(50)→(51) | DROP + ADD (data loss) | ALTER (correct) |
| Change comment only | May trigger DROP+ADD | Only comment compared |
| Change base type | DROP + ADD (correct) | DROP + ADD (correct) |

Fixes: typeorm/typeorm#3357


pkuczynski:
Closing in favor of #12323 which takes a more robust approach — instead of normalizing the type comparison, it moves type/length changes from the destructive DROP+ADD path to a safe `ALTER COLUMN ... TYPE` with a `USING` clause. That PR also includes regression tests and handles enum transitions.

PR #12283, opened 2026-03-31
test(cascades): unskip composite PK cascade tests and remove driver restrictions

pkuczynski:
Unskip cascade tests that were skipped due to issues that have since been resolved.
Remove unnecessary driver restrictions — all tests now run against all drivers.

- `cascade-save-shared-pk`: OneToOne with shared PK + CreateDateColumn (was mysql/postgres only, skipped)
- `cascade-save-insert-vs-update`: ManyToOne with composite FK-as-PK (was postgres only, skipped)
- `cascade-remove`: cascade remove with OneToMany + ManyToMany (was mysql only, skipped)
- Fix wrong assertion in skipped `cascade-soft-remove` M2M recover test (expected 1 photo, but 2 were created)


gioboa (approved):

smith-xyz (approved):

alumni:
Shouldn't these changes be part of #12286?

PR #12288, opened 2026-03-31
fix: scope computed-columns join to correct table in MSSQL schema query

### LINKED ISSUE #12019: Unnecessary migrations generated with computed columns on MSSQL
### Issue description

On Microsoft SQL Server, when a column in one table has the same name as a computed column in another table, generating a migration will always generate code dropping and re-creating the column in the first table, even when that column already exists in the database and the definition hasn't changed.

### Expected Behavior

Generating a migration should only generate code for columns and/or entities whose definitions differ from what's in the database.

### Actual Behavior

When a column in one table has the same name as a computed column in another table, generating a migration will always generate code dropping and re-creating the first column, even when the definition already matches what's in the database.

### Steps to reproduce

Given entity definitions like:

```typescript
@Entity({name: 'tableA'})
export class TableA {
  @Column('int')
  x!: number;
  @Column({
    type: 'bit',
    asExpression: 'CASE WHEN x <> 0 THEN 1 ELSE 0 END',
    generatedType: 'STORED'
  })
  isActive!: boolean;
}

@Entity({name: 'tableB'})
export class TableB {
  @Column('bit')
  isActive!: boolean;
}
```

Generate a migration and run it to deploy the tables to the database. Then, without changing anything, generate another migration. TypeORM will create a migration dropping and re-creating the `isActive` column in `tableB`.

### My Environment

| Operating System    | Ubuntu 25.10|
| Node.js version     | 20.19.4  |
| Typescript version  | 5.9.3 |

### Additional Context

I believe this bug is due to a bad join condition when TypeORM is inspecting the current state of the database schema. When I run `migration:generate`, one of the SQL statements that TypeORM outputs looks like the following (I've added newlines and removed the `where` clause):

```sql
SELECT "COLUMNS".*, "cc"."is_persisted", "cc"."definition"
FROM "MyDatabase"."INFORMATION_SCHEMA"."COLUMNS"
LEFT JOIN "sys"."computed_columns" "cc" ON COL_NAME("cc"."object_id", "cc"."column_id") = "column_name"
```

This joins to the `computed_columns` table on only the column name, not taking into account the table or schema name. This means, in the example above, the `isActive` columns in _both_ `tableA` and `tableB` will match to the `computed_columns` row. I suspect this causes TypeORM to think that `isActive` in `tableB` has changed from a computed column to a non-computed column, so it generates a migration dropping and re-creating the column. A corrected join condition could look like this:

```sql
SELECT "COLUMNS".*, "cc"."is_persisted", "cc"."definition"
FROM "MyDatabase"."INFORMATION_SCHEMA"."COLUMNS"
LEFT JOIN "sys"."computed_columns" "cc" ON COL_NAME("cc"."object_id", "cc"."column_id") = "column_name" 
  AND OBJECT_NAME("cc"."object_id") = "COLUMNS"."TABLE_NAME"
  AND OBJECT_SCHEMA_NAME("cc"."object_id") = "COLUMNS"."TABLE_SCHEMA"
```

This ensures that we're joining to the column in the correct table, so the computed column definition matches only to the one actual computed column.

- [ ] aurora-mysql
- [ ] aurora-postgres
- [ ] better-sqlite3
- [ ] cockroachdb
- [ ] cordova
- [ ] expo
- [ ] mongodb
- [ ] mysql
- [ ] nativescript
- [ ] oracle
- [ ] postgres
- [ ] react-native
- [ ] sap
- [ ] spanner
- [ ] sqlite
- [ ] sqlite-abstract
- [ ] sqljs
- [x] sqlserver

PreAgile:
## Summary
- Fix the `sys.computed_columns` LEFT JOIN in `SqlServerQueryRunner` to include `OBJECT_NAME` and `OBJECT_SCHEMA_NAME` conditions
- Without these, a regular column sharing its name with a computed column in another table was incorrectly identified as computed, generating unnecessary migrations on every sync

## Root Cause
The join condition only matched by column name (`COL_NAME(...) = "COLUMN_NAME"`), with no table or schema filter. This caused cross-table computed column definitions to leak into unrelated tables.

## Changes
- `src/driver/sqlserver/SqlServerQueryRunner.ts`: Add `OBJECT_NAME` and `OBJECT_SCHEMA_NAME` conditions to the computed-columns join (2 lines)
- New regression test: two entities (`TableWithComputed`, `TableWithRegular`) sharing column name `fullName`, asserting that schema diff produces no queries after initial creation

## Test plan
- [x] `pnpm tsc --noEmit` passes
- [x] MSSQL functional test suite (CI)

Closes #12019


Cprakhar (approved):
@PreAgile Thanks for the amazing fix 🚀.

I just combined your test entities into the existing test suite. No need for an extra `test.ts` file.

gioboa (approved):

PR #12303, opened 2026-04-02
chore!: remove dead skipNulls option from getEntityValueMap

naorpeled:
## Summary
- Remove the unused `options?: { skipNulls?: boolean }` parameter from `ColumnMetadata.getEntityValueMap()` and `EntityMetadata.getValueMap()`
- Remove the `returnNulls` variable that was hardcoded to `false`, making the `skipNulls` option non-functional
- Simplify the two conditions that referenced `returnNulls` to just check `!== undefined` (matching actual behavior)
- Remove `{ skipNulls: true }` from the two call sites (`getEntityIdMap`, `MaterializedPathSubjectExecutor`)

## Context
The `skipNulls` option was never wired up — `returnNulls` was hardcoded to `false` with an old TODO comment suggesting it might be re-enabled. This has been the stable behavior for years. Since nulls were always included regardless of the option value, this is a no-op removal with no behavior change.

Spawned from review feedback on #12284.

## Test plan
- [x] `pnpm run typecheck` passes (only pre-existing unrelated errors)
- [x] CI passes


pkuczynski:
hey @naorpeled can you solve the conflicts, make the build pass and solve qodo, before we review? Let me know when ready...

alumni (approved):

gioboa (approved):

alumni:
@michaelbromley let's release another beta when this gets merged :)

PR #12304, opened 2026-04-03
feat(caching): implement in-memory cache with lru cache

### LINKED ISSUE #11934: in-memory caching as a inbuilt feature in TypeORM
### Feature Description

TypeORM is a great battle tested ORM library and I've personally been a big proponent of its usage in production grade projects. But recently, I came across a use case when I needed to cache the results of multiple queries as the values needed to be used for business calculations. 

The values aren't changed very often and there are quite a few a few of these to be fetched to in order to perform the calculations. Hence the need for caching. TypeORM currently allows for wide variety of caching methods like redis, ioredis, database etc., the default being database which allows to cache the results of a query within a table. 

But this is not a very practical solution as one needs to ensure the either of the following are set:
1. `synchronize` is set to true on dev
2. or a migration is written to create the table as per TypeORM in production

Even with this implemented, the end result is a final query which then hits the `query-result-cache` table to fetch the results. Given that I need to fetch multiple values, the `query-result-cache` table is hit multiple times and there's no real gain in caching the values at this level. This is, in **no way**, a good solution for applications that want a light weight caching without really adopting additional caching services like redis or its similar counterparts to ensure lower costs.

### The Solution

The solution would be to create a `InMemoryResultCache` which implements the `QueryResultCache` interface. The in memory cache can be built over a simple `lru-cache` which allows the results to be cached in memory, also while allowing the user to reap the benefits of max entries and ttls.

The `InMemoryResultCache` can be set as the default cache type allowing small and mid sized application to gain the benefits of caching without really inflating their costs.

The default LRU cache can be set to a max size of 10MB with a maximum entry limit of 10,000 entries, thus giving a total of 1 KB limit to each query result.

Additionally, the default configuration options of the `lru-cache` can be allowed for be overridden explicitly when the database connection is initialized. 

### Considered Alternatives

I've tried the following alternative already:
1. `database` caching options of TypeORM
2. creating of a custom cache provider to be supplied to TypeORM cache provider factory method.
3. Querying multiple query results with a single query ( `IN` clause) as far as possible.
4. Tried using https://github.com/EchelonFour/typeorm-in-memory-cache but its dangerous as the memory footprint keeps growing with the increasing number of cached results
5. Tried using https://www.npmjs.com/package/typeorm-cache but has similar dangers of growing memory footprint.

### Additional Context

In my opinion, `InMemoryResultCache` can be applied to all the types of connections/databases provided that its enabled at the connection level and the cache time has been specified at the query level. I could be wrong here (Hence only ticking postgres database driver), but I'm open to inputs and I'd be happy to create a PR for this one as I already have it working in my project.

- [ ] aurora-mysql
- [ ] aurora-postgres
- [ ] better-sqlite3
- [ ] cockroachdb
- [ ] cordova
- [ ] expo
- [ ] mongodb
- [ ] mysql
- [ ] nativescript
- [ ] oracle
- [x] postgres
- [ ] react-native
- [ ] sap
- [ ] spanner
- [ ] sqlite
- [ ] sqlite-abstract
- [ ] sqljs
- [ ] sqlserver

CretanBull:
### Description of change

This pull request introduces a new in-memory query result cache provider for TypeORM.

  What the change is intended to do:
   - Implements InMemoryResultCache using an LRU (Least Recently Used) and Fixes #11934
   - Adds in-memory as a valid caching type in BaseDataSourceOptions.
   - Updates QueryResultCacheFactory to instantiate the new provider when type: "in-memory" is configured.
   - Allows configuration of LRU settings (max entries, max size, TTL) via the cache.options object.

  Why this change is needed:

  Users often need a simple, zero-dependency caching mechanism for query results. While TypeORM supports Redis and Database-backed caching, an in-memory option provides a lightweight alternative that doesn't require external infrastructure or additional database tables, making it ideal for small-to-medium datasets or ephemeral environments.

  How it's been verified:
   - Added a new functional test suite: test/functional/cache/in-memory-cache-provider.test.ts.
   - Verified core caching logic, expiration (TTL), pagination support, and custom cache identifiers.
   - Verified that it correctly integrates with the QueryResultCacheFactory.

  Other context:
  The implementation relies on lru-cache.

### Pull-Request Checklist


smith-xyz:
just to note my comment from https://github.com/typeorm/typeorm/issues/11934#issuecomment-4184862920

alumni:
Hi @CretanBull, thanks for the contribution!

I would not add this directly into TypeORM, I think it deserves its own package. This would keep the complexity of the main TypeORM package smaller.

We actually want to extract all caching strategies into separate packages for the same reason, which will probably happen later this year. I would keep the PR open until then.

michaelbromley:

```
git fetch origin
```

PR #12307, opened 2026-04-03
fix(postgres): use ALTER COLUMN for length changes, not DROP+ADD

fino-oss:
## Summary

Fixes #3357

## Problem

When changing a PostgreSQL column's length (e.g. `varchar(50)` → `varchar(100)`), TypeORM generates:

```sql
ALTER TABLE "post" DROP COLUMN "title";
ALTER TABLE "post" ADD COLUMN "title" character varying(100);
```

This causes **silent data loss** — the existing column values are permanently gone before the new column is created.

## Root cause

In `PostgresQueryRunner.changeColumn()`, the condition that triggers DROP + ADD included `oldColumn.length !== newColumn.length`. Any length change, even a safe increase, fell into the destructive path.

## Fix

Remove `length` from the DROP+ADD condition and add a dedicated block (modelled after the existing `precision`/`scale` handler) that emits:

```sql
ALTER TABLE "post" ALTER COLUMN "title" TYPE character varying(100);
```

PostgreSQL supports `ALTER COLUMN … TYPE` for all types in `withLengthColumnTypes` (`varchar`, `char`, `bit`, `varbit`, `vector`, `halfvec`). If existing data does not fit the new length, PostgreSQL raises an **explicit error** — a much better signal than silently losing data.

**Unchanged behaviour:** if the *base type* also changes (e.g. `char` → `varchar`), TypeORM still falls through to DROP+ADD, which is correct since that is a structural change.

## Tests

Added `test/github-issues/3357/` with two integration tests (PostgreSQL only):

1. **Query generation** — asserts that a length change produces `ALTER COLUMN … TYPE`, not `DROP COLUMN`.
2. **Data preservation** — inserts a row, runs the ALTER, verifies the row still exists.


pkuczynski:
Hey @fino-oss, thanks for this! We're going with #11620 which handles #3357 across all major drivers and adds a safety guard for cross-family type changes. Closing in favor of #11620 — would love your feedback there!

PR #12310, opened 2026-04-04
fix: prevent unnecessary dropColumn in schema migration

dingdyan:
This PR fixes issue #3357 where TypeORM drops and recreates columns instead of altering them during schema migration.

## Changes
- Added `isTypeAlterable` method to check safe type conversions
- Modified `changeColumn` to prefer `ALTER COLUMN` over `dropColumn + addColumn`
- This prevents data loss when changing column type or length

## Testing
- Tested with PostgreSQL driver
- Verified that `ALTER COLUMN` is used for safe type/length changes
- Verified that `dropColumn + addColumn` is only used as a fallback for complex changes

Fixes #3357


smith-xyz:

Issue #246, opened 2017-01-30
[SQLITE] Add options at column creation


blue112:
I'd like to add options to some columns, so that this

    CREATE TABLE "group" ("id" integer NOT NULL PRIMARY KEY AUTOINCREMENT, "name" character varying(200) NOT NULL, "lastUpdateDate" timestamp without time zone NOT NULL, "createDate" timestamp without time zone NOT NULL, "creator" integer, "lastUpdater" integer)

     "name" character varying(200) NOT NULL

Turns into this

    "name" character varying(200) NOT NULL COLLATE NOCASE

So, just adding a custom string field to add after the field type descriptor would be fine.

pleerock:
If we add "custom string field" schema sync wont work - because it won't be able to determine if column is changed or not, so its not recommended to add functionality. What exactly do you want to specify? Column COLLATION ?

blue112:
Yes, I'd like to add COLLATE NOCASE to the column on table create, specifically.
So you could call that collation, and allow it to be set to "NOCASE" value in case of sqlite.

pleerock:
Okay I'll mark this as a new feature and we will add "collation" into column options.

pleerock:
this is easy feature but it needs time to implement it properly on all platforms and provide a test coverage. This is opened for PR if you have time and wish.

AlexMesser:
released in `0.1.0-alpha.33`

PR #12312, opened 2026-04-04
fix(sqlite): retrieve collation info from db

### LINKED ISSUE #11907: sqlite: `loadTables` method unable to get collation
### Issue description

`AbstractSqliteQueryRunner.loadTables` method unable to get collation, resulting in loss of all collation after entity change.

### Expected Behavior

eg:
```
  @Column({ type: 'text', collation: 'NOCASE' })
  name!: string
```

After the entity changes, TypeORM creates a temporary table, which should contain collation of all columns.

### Actual Behavior

eg:
```
  @Column({ type: 'text', collation: 'NOCASE' })
  name!: string
```

After the entity changes, TypeORM creates a temporary table, which discards the collation of all columns.

### Steps to reproduce

1. Create a column in the entity with the collation attribute.
2. Delete the existing sqlite file and call the `synchronize` method.
3. Make any changes to the entity.
4. Call the `synchronize` method again.
5. All collation is lost.

### My Environment

| Node.js version     | 24.12.0  |
| Typescript version  | 5.9.3    |
| TypeORM version     | 0.3.28   |

### Additional Context

_No response_

- [ ] aurora-mysql
- [ ] aurora-postgres
- [x] better-sqlite3
- [ ] cockroachdb
- [ ] cordova
- [ ] expo
- [ ] mongodb
- [ ] mysql
- [ ] nativescript
- [ ] oracle
- [ ] postgres
- [ ] react-native
- [ ] sap
- [ ] spanner
- [x] sqlite
- [ ] sqlite-abstract
- [ ] sqljs
- [ ] sqlserver

Cprakhar:
Closes #11907

### Description of change
Loads table columns' collation type from the database during `loadTables` for SQLite.

**Before:** schema sync ignored column `collation` changes.
**After:** schema sync takes into account `collation` changes.

### Pull-Request Checklist
- [x] This pull request links relevant issues as `Issue #246`

**Related**
- #246

### DIFF
--- a/src/driver/sqlite-abstract/AbstractSqliteQueryRunner.ts
+++ b/src/driver/sqlite-abstract/AbstractSqliteQueryRunner.ts
@@ -48,6 +48,43 @@ export abstract class AbstractSqliteQueryRunner
     }
 
     // -------------------------------------------------------------------------
+    // Private Methods
+    // -------------------------------------------------------------------------
+
+    /**
+     * Extracts column collation from the given sql.
+     *
+     * @param sql
+     * @param columnName
+     * @returns
+     */
+
+    private getColumnCollation(
+        sql: string,
+        columnName: string,
+    ): string | undefined {
+        const escapedColumnName = this.driver.escape(columnName)
+        const columnDefStart = sql.indexOf(`${escapedColumnName}`)
+        if (columnDefStart === -1) return undefined
+
+        sql = sql.slice(columnDefStart + escapedColumnName.length)
+        sql = sql.replaceAll(/CHECK\s*\(\s*(.*)\s+\)\s*/g, "")
+        sql = sql.replaceAll(/DEFAULT\s+\(.*\)\s*/g, "")
+
+        const collateStart = sql.indexOf("COLLATE")
+        if (collateStart === -1) return undefined
+
+        const stripped = sql.slice(0, collateStart)
+
+        if (stripped.indexOf(",") !== -1) return undefined
+
+        const collateRegex = /\bCOLLATE\s+"(\w+)"\s*/i
+        const matches = collateRegex.exec(sql)
+        if (!matches) return undefined
+        return matches[1]
+    }
+
+    // -------------------------------------------------------------------------
     // Public Methods
     // -------------------------------------------------------------------------
 
@@ -1549,6 +1586,10 @@ export abstract class AbstractSqliteQueryRunner
                         : dbTable["name"]
 
                 const sql = dbTable["sql"]
+                const tableDefinition = sql.slice(
+                    sql.indexOf("(") + 1,
+                    sql.lastIndexOf(")"),
+                )
 
                 const withoutRowid = sql.includes("WITHOUT ROWID")
                 const table = new Table({ name: tablePath, withoutRowid })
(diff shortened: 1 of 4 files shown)


test/functional/database-schema/column-collation/sqlite/entity/Post.ts
```diff
@@ -9,4 +9,28 @@ export class Post {
...
+    @Column({ default: "COLLATE NOCASE" })
+    defaultContainsCollateKeyword: string
+
+    @Column({ type: "varchar", enum: ["A", "B", "C,D"], collation: "NOCASE" })
+    enumCollated: string
+
+    @Column({ name: "post", collation: "NOCASE" })
+    nameAsTableName: string
+
+    @Column({ name: "name,with,comma", collation: '"rtrim"' })
```
alumni: Why is `"rtrim"` (with quotes) accepted as a value? I'd expect the user to define either `RTRIM` (according to theory, or maybe `rtrim` if that one works _in practice_ and the uppercase doesn't).

Cprakhar: Create table query doesn't seem to fail when using `[NOCASE]`, `"nocase"`, or `'nocase'`.

If this is what is expected from the user, I will remove extra logic related to this.

src/driver/sqlite-abstract/AbstractSqliteQueryRunner.ts
```diff
@@ -47,6 +47,114 @@ export abstract class AbstractSqliteQueryRunner
...
+    /**
+     * Extracts table definition from the given sql.
+     *
+     * @param sql
+     * @param tableName
+     * @returns
+     */
+    private getTableDefinition(sql: string, tableName: string): string {
+        const regex = new RegExp(
+            String.raw`CREATE\s+TABLE\s+(?:IF\s+NOT\s+EXISTS\s+)?${this.escapePath(tableName)}\s*\((.+)\)`,
```
alumni: I would avoid creating a RegExp dynamically - I know that most likely this is something that the developers provide (unless it's a dynamic entity, then it could be user input), but regular expressions have their own characters that need escaping.

This applies to the next method as well. Could we do this somehow else? (pure js instead of regex?)

src/driver/sqlite-abstract/AbstractSqliteQueryRunner.ts
```diff
@@ -47,6 +47,114 @@ export abstract class AbstractSqliteQueryRunner
...
+
+    /**
+     * Extracts column collation from the given sql.
+     *
+     * @param sql
+     * @param columnName
+     * @returns
+     */
+
+    private getColumnCollation(
```
alumni, line 62: SonarCloud is right, this is a bit difficult to understand - not sure how you could break it, but at least some comments would be nice (e.g. "in string" for `'`, escape single quote for `''` - i know they are trivial but they speed up the understanding).

There's no other table to read the collation from, so we don't have to parse SQL, is there?

test/functional/database-schema/column-collation/sqlite/entity/Post.ts
alumni: Hmm, because it's inserted into the SQL directly. I was thinking about some validation (e.g. only allow alphanumeric and underscore), but not sure what that will break on other drivers. Another option would've been to treat it like an identifier or a string and always escape + add the right type of quotations.

The BNF of Sqlite on their official website does not say what "collation-name" can be - especially for user-defined collations (not sure if they are even possible in the nodejs version).

I'm good with anything :)

michaelbromley:

```
git fetch origin
```

PR #12316, opened 2026-04-07
fix: handle null embedded entities on save/update

### LINKED ISSUE #3913: Cannot set embedded entity to null

hypernova1:
## Description

When an embedded entity property is explicitly set to `null` (e.g. `employee.salary = null`), TypeORM previously either:
1. Threw a runtime error trying to access properties on `null`
2. Silently ignored the change, leaving stale data in the database

This PR fixes both issues by properly handling `null` embedded entities during persistence.

### `src/metadata/ColumnMetadata.ts`

**`getEntityValueMap`**: Added null check for embedded property values. When an embedded property is `null`, the method now returns `{ [propertyName]: null }` for the column instead of crashing or returning an empty map.

**`getEntityValue`**: Added explicit null handling — when the embedded object resolves to `null`, the column value is set to `null` instead of remaining `undefined` (which would skip the update).

### Test

Added `test/functional/embedded/embedded-set-null/` with two test cases:
- Setting an existing embedded to `null` should set all its DB columns to `NULL`
- Inserting an entity with a `null` embedded should not throw

## Motivation

Closes #3913

This is a long-standing issue (reported in 2019) where users expect `entity.embedded = null` to nullify all embedded columns in the database, matching the behavior of regular nullable columns.

## Example

```typescript
@Entity()
class Employee {
    @PrimaryGeneratedColumn()
    id: number

    @Column(() => Salary)
    salary: Salary | null
}

// This should work but previously threw or was silently ignored
employee.salary = null
await repository.save(employee)
// Expected: salary_amount = NULL in DB
```


michaelbromley:

```
git fetch origin
```

PR #12317, opened 2026-04-07
docs: add column type inference explanation

### LINKED ISSUE #8827: No docs on inferred column types
### What was unclear or otherwise insufficient?

I cannot find information on which column type is used if there is no explicit `type` given in a `@Column()` decorator. Does a `string` result in `varchar` or `text`? If it's a `varchar`, with what length? Similar for other types.

### Recommended Fix
Add a table explaining which column type is used for every possible TypeScript type (or what happens if no appropriate column type exists).

ElwinErnst:
### Description of change

This PR adds documentation explaining how TypeORM infers column types when no explicit type is provided in the `@Column()` decorator.

Currently, this behavior is not clearly documented, which can lead to confusion for developers when defining entities without specifying column types.

### What was added

- Explanation of column type inference behavior
- Example using `@Column()` without explicit type
- Mapping table for common TypeScript types across database drivers
- Notes about driver-specific behavior and limitations
- Clarification on what to do when type inference is not possible

### Why this change is needed

The current documentation does not explain how TypeORM determines column types when the `type` option is omitted. This PR provides a clear and practical overview to help developers understand the behavior and avoid unexpected results.

### How it was verified

- Reviewed existing documentation to confirm the gap
- Cross-checked behavior with TypeORM driver implementations
- Ensured the content is consistent with driver-specific variability

---

### Pull-Request Checklist

- [x] This pull request links relevant issues as `Fixes #8827`
- [ ] There are new or updated tests validating the change (N/A - documentation only)

Fixes #8827


ElwinErnst:
All review comments addressed in the latest update ✅

ElwinErnst:
Thanks for the feedback.

I intentionally avoided documenting specific defaults (such as varchar length or exact string → varchar/text mapping) because this behavior is driver-specific and not consistently defined across all databases supported by TypeORM.

Instead, I clarified that:
- inference depends on the driver implementation
- mappings are only common cases (not guaranteed)
- explicit column types should be used in production

Documenting fixed defaults here could be misleading for users, especially across different drivers or configurations.

If there is a specific source of truth in the codebase that defines consistent defaults per driver, I’d be happy to extend this section with more precise details.

PR #12302, opened 2026-04-02
refactor: replace deprecated substr

Cprakhar:
Related #12149

### Description of change
Replace deprecated `substr` with alternatives throughout the codebase.

### Pull-Request Checklist

- [x] Code is up-to-date with the `master` branch
- [ ] This pull request links relevant issues as `Fixes #00000`
- [ ] There are new or updated tests validating the change (`tests/**.test.ts`)
- [ ] Documentation has been updated to reflect this change (`docs/docs/**.md`) (N/A)

PR #12320, opened 2026-04-08
style(eslint): add rules preventing deprecated string methods

pkuczynski:
Add two ESLint rules and fix remaining violations:

- `unicorn/prefer-string-slice` — prevents use of deprecated `substr` and `substring` in favor of `slice`
- `unicorn/prefer-string-starts-ends-with` — prevents `indexOf() === 0` patterns in favor of `startsWith()` / `endsWith()`

### Why `slice` over `substring`?

| Behavior | `slice` | `substring` |
|---|---|---|
| Negative indices | Counts from end (`slice(-3)` = last 3 chars) | Treats as `0` |
| `start > end` | Returns `""` | Silently swaps the arguments |
| Consistency | Same API as `Array.prototype.slice` | String-only |

`substring` silently swapping arguments when `start > end` can mask bugs. `slice` behaves predictably and is consistent with array usage.

`substr` is deprecated entirely (ECMAScript Annex B, legacy browser compat only).

Both rules include auto-fix support. Follows up on #12302 which manually replaced `substr` usage across the codebase.

### `DriverUtils` note

One place in `DriverUtils` needed more than a mechanical swap: `parseConnectionUrl` and `parseMongoDBConnectionUrl` relied on `substring`'s negative-index clamping to handle credential-less URLs (`base.substring(0, lastAtSign)` silently returned `""` when `lastAtSign === -1`). Since `slice` treats negative indices as offsets from the end, that idiom no longer works. Both call sites now share a new `parseCredentials` helper that guards `lastAtSign === -1` explicitly — the credential-less case is no longer an accidental byproduct of `substring`'s semantics.


alumni (dismissed):
As diacussed quite a few times, let's not add unicorn.

pkuczynski:
> As diacussed quite a few times, let's not add unicorn.

As mentioned many times: lets not add recommended from unicorn but use rules which make sense, like in this example. Thanks to those rules I already discovered few `indexOf(“...") === 0` missed in #12302

If you know a better way of preventing contributors to add code with `substr`, please show me...

pujux:
I'm with piotr here, as long as we only add rules that make sense and aren't too opinionated it's a good improvement. I'm very much in favor of not adding all recommended rules from unicorn, but the ones mentioned before and here do make sense in my opinion.

gioboa:
> I'm with piotr here, as long as we only add rules that make sense and aren't too opinionated it's a good improvement. I'm very much in favor of not adding all recommended rules from unicorn, but the ones mentioned before and here do make sense in my opinion. 

Same for me, those rules are ok to me. I like consistency

OSA413:
I would need an example of actual pros of the replacement. Currently I don't see a problem with using substring (btw I've never heard of String.slice until now), maybe we should see what algorithm changes it produced and what are pros and cons of that.

src/driver/aurora-mysql/AuroraMysqlQueryRunner.ts
```diff
@@ -2482,13 +2482,13 @@ export class AuroraMysqlQueryRunner
...
                         const items = colType
-                            .substring(
+                            .slice(
                                 colType.indexOf("(") + 1,
                                 colType.lastIndexOf(")"),
                             )
                             .split(",")
                         tableColumn.enum = (items as string[]).map((item) => {
-                            return item.substring(1, item.length - 1)
+                            return item.slice(1, -1)
```
OSA413, line 2490: This is a good example of simplification.

src/driver/cockroachdb/CockroachQueryRunner.ts
```diff
@@ -3441,8 +3440,8 @@ export class CockroachQueryRunner
                             // ----
                             // so, we must remove this underscore character from enum type name
                             let udtName = dbColumn["udt_name"]
-                            if (udtName.indexOf("_") === 0) {
-                                udtName = udtName.substring(1, udtName.length)
+                            if (udtName.startsWith("_")) {
+                                udtName = udtName.slice(1)
```
OSA413, line 3422: ~~This is also a good example.~~ maybe this still can be a substring?

OSA413 (commented):
So, this just replaces substring with slice with almost all of the constants as arguments. I like a few changes but I don't see a reason to replace all of the substrings to slices.

pkuczynski:
> So, this just replaces substring with slice with almost all of the constants as arguments. I like a few changes but I don't see a reason to replace all of the substrings to slices.

I provided this in the PR desc. Anything missing? Technically they very the same, but there are small differences like `substring` swapping arguments, which can be error prone.

What’s most important is that this rule protects from `substr` which is deprecated.

pkuczynski:
@OSA413 I think the bigger question is: should we use other plugings in eslint config like `unicorn` (but also others) to tighten our code style or not. Me, @gioboa, @pujux (do not know opinion of others) do not see a problem with using other plugins as long as we enable single rules working well for us. In opposition to that @alumni does not want any other plugins and keep denying my PRs. there is another similar one bringing other rule #12222. For me they all help to lower down maintenace burden and help external contributors to maintain our coding style. Especially they all come with auto fixes.

src/driver/cockroachdb/CockroachQueryRunner.ts
pkuczynski: It could be, but `slice(1)` should be more readable and does nor require reading lenfth from the string.

OSA413:
I guess you are right, substring can wrap its arguments, which is bad.
I guess that only that rules included are fine.

gioboa (approved):
For me it's ok, thanks 🙏

pujux (approved):

pkuczynski:
> I guess you are right, substring can wrap its arguments, which is bad. I guess that only that rules included are fine.

Can you then approve the PR @OSA413 ?

smith-xyz:
The goals here are sound, but could `no-restricted-syntax` achieves the same enforcement with zero plugin dependencies. It won't auto-fix, but the codebase is already migrated (#12302). This avoids the unicorn debate entirely while still blocking substr/substring/indexOf===0 in CI.

smith-xyz:
Thinking something like: 

```
"no-restricted-syntax": ["error",
  {
    "selector": "CallExpression[callee.property.name='substr']",
    "message": "substr is deprecated. Use slice() instead."
  },
```

pkuczynski:
> The goals here are sound, but could `no-restricted-syntax` achieves the same enforcement with zero plugin dependencies. It won't auto-fix

Yes, this would be alternative for this specific case, but quite limited in functionality (no auto-fix, basic `substr` prevention). Crossing eslint plugins goes against general eslint concept of letting people provide plugins of their own, extending basic functionality. I do not see anything bad in using plugins, and at the same time I see big value in static code analysis and failing fast.

smith-xyz (approved):
I conditionally approve, unless there is a really good reason to maybe leverage the eslint alternative I provided and if we're taking on only the plugins we need.

PR #12323, opened 2026-04-08
fix(postgres): use ALTER COLUMN TYPE instead of DROP+ADD

MastadoonPrime:
## Summary

Fixes #3357

When changing a PostgreSQL column's type or length (e.g. `varchar(50)` to `varchar(100)`), `PostgresQueryRunner.changeColumn()` previously dropped and recreated the column:

```sql
ALTER TABLE "post" DROP COLUMN "title"
ALTER TABLE "post" ADD "title" character varying(100)
```

This causes **silent data loss**. The fix changes this to use:

```sql
ALTER TABLE "post" ALTER COLUMN "title" TYPE character varying(100) USING "title"::character varying(100)
```

## Changes

- **Removed** `type` and `length` from the destructive DROP+ADD condition in `PostgresQueryRunner.changeColumn()`
- **Added** a dedicated `ALTER COLUMN ... TYPE ... USING` block (modeled after the existing precision/scale handler) that preserves data
- **Handles enum transitions** properly: when changing from/to enum types, the enum type is created/dropped in the correct order around the ALTER
- **Unchanged behavior**: array dimension changes and generated column changes still use DROP+ADD (these genuinely require column recreation)

The `USING` clause ensures PostgreSQL can cast between compatible types automatically. If a cast is not possible (e.g. `text` containing non-numeric data to `integer`), PostgreSQL raises an explicit error rather than silently losing data.

## Tests

- Added data-preservation tests to `test/functional/schema-builder/change-column.test.ts` (the preferred location per project guidelines) covering:
  - Column length change with data verification
  - Column type change (varchar to text) with data verification  
  - Down-migration reversal with data verification
- Added `test/github-issues/3357/` tests for issue traceability:
  - Data preservation on length change
  - Data preservation on type change
  - SQL generation verification (asserts ALTER COLUMN TYPE is used, not DROP COLUMN)


pkuczynski:
Hey @MastadoonPrime, thanks for this! Similar to #12307, we're going with #11620 which handles #3357 across all major drivers with a safety guard for cross-family type changes. Closing in favor of #11620 — would love your feedback there!

PR #12328, opened 2026-04-08
docs(performance): add database-specific EXPLAIN examples to indexes guide

pkuczynski:
Expand the "Using indexes" performance optimization guide with database-specific guidance and introduce the first use of tabbed UI in the documentation.

Content changes:
- Introductory paragraphs explaining what indexes are, the read/write trade-off, and guidance on composite index column ordering
- Database-specific "Analyzing query performance" section showing EXPLAIN syntax for each supported database
- Aurora MySQL/PostgreSQL tips with links to AWS-specific documentation
- Tabbed TypeORM examples showing basic, composite, and join column index definitions
- Closing summary tying indexes back to performance

Infrastructure:
- `DatabaseTabs` component (`docs/src/components/tabs/database-tabs.tsx`) that renders database icons as tab labels, auto-injects `<h3>` headings, and validates database names at build time
- SVG database icons replacing PNGs on the homepage
- Sorted databases alphabetically on the homepage

Databases covered: CockroachDB, Google Spanner, MariaDB, MongoDB, MySQL, Oracle, PostgreSQL, SAP HANA, SQL Server, SQLite.


gioboa (approved):

docs/src/pages/index.tsx
```diff
@@ -265,40 +265,40 @@ function CodeExampleSection() {
 
 function SupportedDatabases() {
     const databases = [
-        { name: "MySQL", icon: "/img/databases/mysql.png", category: "core" },
         {
-            name: "PostgreSQL",
-            icon: "/img/databases/postgresql.png",
+            name: "CockroachDB",
+            icon: "/img/databases/cockroachdb.svg",
```
smith-xyz: should these be in a separate PR? Seem unrelated but I see conventional commit messages for them so it isn't too big of an issue, unless there is a convention we follow to reduce PR scope.

pkuczynski: Typically I would do a separate PRs for such changes but @alumni complains that there are too many, so now I tend to squeeze similar/connected things together...

docs/src/components/tabs/database-tabs.tsx
```diff
@@ -0,0 +1,73 @@
...
+    postgres: { label: "PostgreSQL", icon: "/img/databases/postgresql.svg" },
+    sap: { label: "SAP HANA", icon: "/img/databases/sap.svg" },
+    mssql: { label: "SQL Server", icon: "/img/databases/mssql.svg" },
+    sqlite: { label: "SQLite", icon: "/img/databases/sqlite.svg" },
+} as const
+
+type DatabaseName = keyof typeof databases
+
+export const DatabaseTabs = ({ children }: PropsWithChildren) => {
+    const entries = React.Children.toArray(children)
```
smith-xyz: is using React.Children.toArray() necessary here?

pkuczynski: I need this to be able to use `<DatabaseTab value/>` and avoid doing the same tabs descriptions (icon, labels, etc) on other pages. This is the first one using this mechanism, but planning to add many more across the whole docs...

I tried few different ways and this one was the only one which was working...

naorpeled (approved):

OSA413 (dismissed):
I would actually investigate if we are ok to use the logos of the DBs (like, are they under free licenses?)

pkuczynski:
> I would actually investigate if we are ok to use the logos of the DBs (like, are they under free licenses?)

1. We already use db logos on our homepage docs. @dlhck can you shed some light on this?
2. When I checked this was not a problem and fall under the fair usage policy.
3. The images I used come from public icon sets, so if they can publish for years, I would guess we would be fine too?

PR #12329, opened 2026-04-09
fix: use configured logger for CLI command output

### LINKED ISSUE #5910: CLI is using console.log instead of logger
There are several places where CLI is using a `console.log` call rather than logger.

I think it could fallback to `console.log` (via logger), but if logger is provided via connection config, then logger should always be used.

https://github.com/typeorm/typeorm/blob/82b06d0a59ae368233c00b32330d7f87787a64cd/src/commands/SchemaSyncCommand.ts#L48

I'm willing to PR this, if it will be accepted.

Thanks.

Varun5711:
Fixes #5910.

## What changed
- added a CLI helper that routes command status messages through the configured TypeORM logger when `dataSource.options.logger` is present
- kept the existing console output as the fallback when no logger is configured
- updated the DataSource-backed CLI commands that were still writing directly to `console.log`
- added a focused unit test covering no logger, custom logger, and built-in logger behavior

## Why this changed
Several CLI commands initialized a DataSource and then bypassed its logger for user-facing status output. That meant a custom logger supplied in the connection config was ignored even though the command was already operating with a configured DataSource.

## Impact
- custom loggers now receive CLI status output for schema sync, schema drop, schema log, migration generate, query, and cache clear flows
- existing console behavior is unchanged when no custom logger is configured

## Validation
- `pnpm run compile`
- `pnpm exec mocha --no-config --require build/compiled/test/utils/test-setup.js build/compiled/test/unit/commands/command-utils.test.js`
- `git diff --check`


michaelbromley:

```
git fetch origin
```

PR #12331, opened 2026-04-09
style(eslint): enable prefer-optional-chain rule

pkuczynski:
Enable `@typescript-eslint/prefer-optional-chain` as suggested in #12255, replacing verbose `&&` chains with optional chaining (`?.`).


smith-xyz:
opened to discuss this one - the rule is definitely a net benefit because it prevents bugs (e.g. what if the var = 0)

Any migration cost if we do this? Strange bugs from rewriting the existing checks? Test look good, and the semantic correctness is a net benefit. Just want to make sure we're not missing something with these sweeping changes.

gioboa (approved):

naorpeled (approved):

PR #12339, opened 2026-04-10
fix: preserve bigint string type in single JoinColumn value extraction

### LINKED ISSUE #12337: `@JoinColumn` on bigint column causes silent precision loss on INSERT (since 0.3.21)
### Issue description

### Expected Behavior

### Actual Behavior

### Steps to reproduce

```typescript
@Entity('licenses')
export class License {
  id: string;

  sourceId: string | null;

  product: Product;
}

@Entity('products')
export class Product {
  sourceId: string | null;

  licenses: License[];
}
```

### My Environment

### Additional Context

```typescript
    )
}, {})

} else {
        )
    if (value) {
    }
}
```

```typescript
if (
    this.type === "bigint" &&
    value !== null
)
    value = String(value)
```

- [ ] aurora-mysql
- [x] aurora-postgres
- [ ] better-sqlite3
- [ ] cockroachdb
- [ ] cordova
- [ ] expo
- [ ] mongodb
- [ ] mysql
- [ ] nativescript
- [ ] oracle
- [x] postgres
- [ ] react-native
- [ ] sap
- [ ] spanner
- [ ] sqlite
- [ ] sqlite-abstract
- [ ] sqljs
- [ ] sqlserver

EthanKlocked:
## Summary

Fixes #12337

Since PR #10672 (0.3.21), `getEntityValueMap()` in `ColumnMetadata.ts` uses `getEntityValue()` for single `@JoinColumn` columns as a performance optimization. However, `getEntityValue()` lacks the `String(value)` conversion that `getEntityValueMap()`/`createValueMap()` applies for bigint columns. This causes **silent precision loss** when the value exceeds `Number.MAX_SAFE_INTEGER` (`2^53 - 1`).

### The Problem

```typescript
// Single JoinColumn path (0.3.21+) — missing bigint protection:
    .referencedColumn!.getEntityValue(entity[...])
// value can be coerced to Number downstream, losing precision

// Composite key path & createValueMap() — has protection:
if (this.type === "bigint" && value !== null)
    value = String(value)  // ← this protection is missing in the single JoinColumn path
```

### The Fix

Added `String(value)` conversion for bigint types in the single `@JoinColumn` code path, matching the existing protection in `createValueMap()`.

### Impact

In our production environment, this bug silently corrupted ~1,457 records across ~575 users. The corrupted FK values caused `INNER JOIN` failures, making affected records invisible to users with no error logged.

### Affected Versions

- **0.3.21 ~ 0.3.28** (all affected)
- **0.3.19, 0.3.20** (safe — before PR #10672)


pkuczynski:
hey @Cprakhar, seems that this one and #12342 fix the same issue #12337. Can you decide which one you want to move forward and let's close the other?

Cprakhar:
> hey @Cprakhar, seems that this one and #12342 fix the same issue #12337. Can you decide which one you want to move forward with, and let's close the other?

@pkuczynski
This PR handles bigint aliases like int8 or int64 used by some drivers, see `Driver.normalizeType()`.

@EthanKlocked some suggestions:
- address the qodo's 6th review as test is not directly asserting the changes.
- Use optional chaining in the tests, e.g., `loadedPost!` to `loadedPost?`
- Add the tests in the `test/functional`

test/github-issues/12337/issue-12337.test.ts
```diff
@@ -0,0 +1,53 @@
...
+import { Category } from "./entity/Category"
+import { Post } from "./entity/Post"
+
+describe("github issues > #12337 bigint precision loss in single JoinColumn", () => {
+    let dataSources: DataSource[]
+
+    before(async () => {
+        dataSources = await createTestingConnections({
+            entities: [Category, Post],
+            enabledDrivers: ["mysql", "mariadb", "postgres"],
```
Cprakhar:
```suggestion
            enabledDrivers: ["mysql", "mariadb", "postgres", "cockroachdb"],
```

Cprakhar (changes requested):

alumni (changes requested):
Hey @EthanKlocked, how does converting the number to a string solve this? The precision was already lost if the value is a number, you cannot get it back.

A good example is in the test file from this PR:

```js
String(9007199254740993) === '9007199254740993' // false
```

alumni:
> hey @Cprakhar, seems that this one and https://github.com/typeorm/typeorm/pull/12342 fix the same issue https://github.com/typeorm/typeorm/issues/12337. Can you decide which one you want to move forward and let's close the other?

Actually none of them would solve the precision loss issue. This PR might have more tests, but we should always work with strings (or rather with the value that the DB client returns - because some clients are smarter) when the SQL data type is BigInt/Decimal, i.e. we should not force the conversion to Number in the first place.

EthanKlocked:
@alumni You're right. The `typeof value === "number"` condition only fires *after* precision has already been lost, so `String(value)` can't recover it — effectively a no-op. The test in this PR also only exercises the path where string values stay strings, so it doesn't actually verify the fix.

On reflection, I also suspect the nested→flat return shape change in `getEntityValueMap` may itself be what triggers downstream Number coercion elsewhere in the persistence pipeline — meaning `ColumnMetadata` is likely the wrong layer to fix this at in the first place.

The real repro probably needs a dual-mapping entity (`@Column` + `@JoinColumn` on the same column) on master, traced through the persistence pipeline to find where `string → number` actually happens. My current test (single entity + `PrimaryColumn`) doesn't reproduce that regression.

Closing this PR. Issue #12337 would be better repositioned as documentation/warning around Number coercion with bigint FKs, or reopened later with a proper repro + fix at the correct layer.

PR #12340, opened 2026-04-10
style(eslint): enable prefer-nullish-coalescing rule

pkuczynski:
Enable `@typescript-eslint/prefer-nullish-coalescing` as suggested in #12255, replacing `||` with `??` / `??=` where the intent is null/undefined checking rather than falsy checking.


gioboa (approved):

src/driver/mongodb/MongoDriver.ts
```diff
@@ -564,19 +564,20 @@ export class MongoDriver implements Driver {
                 : ""
 
         const portUrlPart =
-            schemaUrlPart === "mongodb+srv" ? "" : `:${options.port || "27017"}`
+            schemaUrlPart === "mongodb+srv" ? "" : `:${options.port ?? "27017"}`
```
OSA413, line 567: Do we have the default port for other DBs?

OSA413 (approved):

alumni (approved):
Prefer optional chaining would be great to have (IMO in the same PR)

pkuczynski:
> optional chainin

Already merged yesterday in #12331 :)

src/driver/mongodb/MongoDriver.ts
pkuczynski: Other drivers don't construct connection URLs manually — they pass options to their client libraries (e.g. `pg`, `mysql2`) which handle default ports internally. MongoDB is unique here because the driver builds the connection string by hand.

PR #12341, opened 2026-04-10
style(eslint): enable no-wrapper-object-types rule

pkuczynski:
Remove the `"off"` override for `@typescript-eslint/no-wrapper-object-types`, restoring the default from `recommendedTypeChecked`.


alumni:
The reason why this rule was explicitly disabled in the past was because it was a breaking change (the two notations are not equivalent). Can't check from the phone every change in the code.

_Later edit:_ could it have been that the `PropertyDecorator` was requiring `Object` instead of `object`? 🤔

pkuczynski:
> The reason why this rule was disabled in the past was because it was a breaking change (the two notations are not equivalent). Can't check from the phone every change in the code.

It kept me wondering why it was disabledm, as it is generally considered best practice to avoid using `Object` type. So I enabled this rule and runing a CI to see what happens. Since we still have not released 1.0 it could be the best time to make such a change...

Have a look when you have a moment at your laptop and I wont merge before you give thumbs up...

gioboa (approved):

gioboa (approved):

michaelbromley:

```
git fetch origin
```

PR #12342, opened 2026-04-11
fix: prevent bigint precision loss in single JoinColumn path

### LINKED ISSUE #12337: `@JoinColumn` on bigint column causes silent precision loss on INSERT (since 0.3.21)

Gdhanush-13:
## Description
Fixes #12337

PR #10672 optimized getEntityValueMap() to use getEntityValue() directly for single @JoinColumn relations. However, getEntityValue() lacks the bigint-to-string safety conversion that createValueMap() has. When a bigint column value exceeds Number.MAX_SAFE_INTEGER, it silently loses precision during INSERT.

## Changes

Added a bigint-to-string guard in the single-JoinColumn code path of getEntityValueMap() in ColumnMetadata.ts. When the referenced column type is igint and the value is a JS 
umber, it converts to String(value) before returning.

## Before
```js
// value = 3588019939803400076 (Number, precision already lost)
```

## After
```js
// value = '3588019939803400000' -> String('3588019939803400000')
// But more importantly, prevents future numeric coercion
if (refCol.type === 'bigint' && typeof value === 'number')
    value = String(value)
```

This mirrors the existing protection in createValueMap() (lines 560-567, 578-584) but applies to all bigint columns, not just generated ones.

### DIFF
--- a/src/metadata/ColumnMetadata.ts
+++ b/src/metadata/ColumnMetadata.ts
@@ -715,11 +715,25 @@ export class ColumnMetadata {
                     if (Object.keys(map).length > 0)
                         return { [this.propertyName]: map }
                 } else {
-                    const value =
+                    let value =
                         this.relationMetadata.joinColumns[0].referencedColumn!.getEntityValue(
                             entity[this.relationMetadata!.propertyName],
                         )
                     if (value) {
+                        // bigint-to-string safety conversion — prevents silent
+                        // precision loss for values > Number.MAX_SAFE_INTEGER.
+                        // getEntityValueMap() → createValueMap() already has
+                        // this guard; getEntityValue() does not, so we must
+                        // apply it here for the single-JoinColumn path. #12337
+                        const refCol =
+                            this.relationMetadata.joinColumns[0]
+                                .referencedColumn!
+                        if (
+                            refCol.type === "bigint" &&
+                            typeof value === "number"
+                        )
+                            value = String(value)
+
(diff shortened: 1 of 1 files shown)


pkuczynski:
hey @Cprakhar, seems that this one and #12339 fix the same issue #12337. Can you decide which one you want to move forward and let's close the other?

alumni (changes requested):
@Gdhanush-13 How does converting the number to a string solve this?

PR #12344, opened 2026-04-12
fix(mysql)!: use index identifiers instead of raw SQL in `QB.useIndex()`

### LINKED ISSUE #12333: `QueryBuilder`: `useIndex()`, `setLock(lockTables)` should escape indices and tables
## Summary
Multiple SQL injection vectors in TypeORM's QueryBuilder string APIs. The high-level find options API (`.find()`, `.findOne()`) is properly defended, but the lower-level QueryBuilder has gaps.

### Finding 1: useIndex() — no escaping at all
**File:** `src/query-builder/SelectQueryBuilder.ts:1602,2311`
`useIndex()` interpolates its argument directly into SQL with zero protection — no escaping, no `assertNoSemicolon`. Affects MySQL/MariaDB.
**Fix:** Apply identifier quoting to the index name.

### Finding 2: setLock() lockTables — direct interpolation
**File:** `src/query-builder/SelectQueryBuilder.ts:1647,2807`
`lockTables` strings from the QueryBuilder API are joined directly into SQL. The find options path correctly validates these, but the direct QueryBuilder API does not.
**Fix:** Apply the same metadata validation used in find options.

### Finding 3: assertNoSemicolon is insufficient
**File:** `src/query-builder/QueryBuilder.ts:1703`
`select()`, `orderBy()`, `groupBy()` only block semicolons. UNION-based and boolean-blind injection work without semicolons.
**Fix:** Use parameterized queries or proper identifier escaping instead of semicolon blocking.

### Finding 4: where object keys without entity metadata
**File:** `src/query-builder/QueryBuilder.ts:1524-1532`
When using `QueryBuilder.where(obj)` on raw tables (no entity metadata), object keys become SQL column references without validation.
**Fix:** Apply identifier quoting to where-object keys when no metadata is available.

## Note
The find options API is well-defended. These issues are in the QueryBuilder string APIs.

eddieran:
## Summary

Fixes #12333. Replaces #12336 which was accidentally closed when the fork was deleted.

User-provided identifiers passed to `useIndex()` and `setLock()` `lockTables` were interpolated directly into SQL strings without escaping, allowing SQL injection.

### Changes

- **`useIndex()`**: Wrap the index name with `this.escape()` so it is quoted with the driver's identifier escape character (e.g. backticks for MySQL).
- **`setLock()` `lockTables`**: Wrap each table name with `this.escape()` before joining into the `OF` clause (Postgres).

Both changes use the existing `QueryBuilder.escape()` method which delegates to the driver's escape implementation, consistent with how all other identifiers are handled throughout the query builder.


smith-xyz (changes requested):
changes look good - could add a test around this behavior in test/functional/query-builder/sql-injection/

eddieran:
@smith-xyz Thanks for the review! I've added SQL injection tests in `test/functional/query-builder/sql-injection/sql-injection.test.ts` covering:

- **`useIndex()`** — verifies malicious index names like `my_index; DROP TABLE post` get properly escaped with backticks (MySQL)
- **`useIndex()` with comma-separated names** — verifies that `idx_one, idx_two` is split and each index is individually escaped, and that malicious payloads in multi-index strings are contained
- **`setLock()` lockTables** — verifies malicious table names in lockTables get properly escaped with double quotes (Postgres), both single and multiple tables

I also addressed the Qodo review feedback: `useIndex` now splits on commas and escapes each index name individually to correctly handle multiple index hints like `USE INDEX (idx_one, idx_two)`. The existing `useIndex` test assertion was updated to reflect the escaped output.

smith-xyz (approved):
lgtm! thank you for your contribution

gioboa (approved):

gioboa:
Thanks @eddieran for your great help 🙏

src/query-builder/SelectQueryBuilder.ts
```diff
@@ -2796,7 +2796,7 @@ export class SelectQueryBuilder<Entity extends ObjectLiteral>
             if (this.expressionMap.lockTables.length < 1) {
                 throw new TypeORMError("lockTables cannot be an empty array")
             }
-            lockTablesClause = " OF " + this.expressionMap.lockTables.join(", ")
+            lockTablesClause = " OF " + this.expressionMap.lockTables.map((table) => this.escape(table)).join(", ")
```
alumni: Possible edge case: using table paths (and similar for indices). Both tables and indices (basically any database object) can be referenced by a path: `"SCHEMA"."MY_TABLE"`.

IMO this is not a blocker to merge, both of these functionalities (`useIndex`/`lockTables`) are almost never used.

alumni (approved):

src/query-builder/SelectQueryBuilder.ts
```diff
@@ -2301,7 +2301,10 @@ export class SelectQueryBuilder<Entity extends ObjectLiteral>
         let useIndex: string = ""
         if (this.expressionMap.useIndex) {
             if (DriverUtils.isMySQLFamily(this.dataSource.driver)) {
-                useIndex = ` USE INDEX (${this.expressionMap.useIndex})`
+                useIndex = ` USE INDEX (${this.expressionMap.useIndex
+                    .split(",")
```
alumni: One thing: we should find a way to not split by `,`. Maybe `expressionMap.useIndex` should be `string[]` and we should define `SelectQueryBuilder.useIndex(...indexes: string[])`/`SelectQueryBuilder.useIndex(indexes: string|string[])` 🤷🏻‍♂️

smith-xyz:
@eddieran can you check on the tests - I think it could be related qodo's no.2 finding. Might need to update the applyFindOptions path to pass tableAlias.name instead of this.escape(tableAlias.name) to setLock()? 

e.g. its triple quoting in there `FOR UPDATE OF """Post"""`

alumni:
It's double escaped because in `Repository` APIs we escape things, but in `QueryBuilder` we don't always (on purpose, for raw SQL).

In this case, I think both should be quoted in the `QueryBuilder` because we do not expect the users to provide raw SQL in those arguments (it's always an index name or a table name, you can't provide e.g. a subquery).

The current escaping behavior in `setLock` is actually the reason why I didn't try to extend the implementation the SAP driver for this sub-feature when I added locking support (it's slightly different than Postgres, you lock columns, not tables) :)

Right now I'm more interested in changing the signature of `useIndex` since it's a breaking change and we can add it to the 1.0 release - it will be difficult to release it after.

Ideally we should not accept issues that list more than 1 point (the linked one has 4 points and e.g. point #3 is irrelevant, we need to handle these things independently from each other) nor PRs that solve more than 1 issue (this one has 2, would have been easier to handle them independently).

eddieran:
Re: comma-split — agree that's the right fix. Switching `useIndex` to `string | string[]` (with rest-param overload) avoids the brittle `.split(",")` and removes a class of edge cases (whitespace, quoted commas in identifiers). Happy to land that here as a follow-up commit since it'd be a breaking change worth landing for 1.0, or split it into a separate PR — your call.

Re: schema-qualified paths (`"SCHEMA"."MY_TABLE"`) — yeah, naive escaping would mangle those. The cleanest path is probably split-on-dot then escape each component, mirroring how `escapePath` already works elsewhere in the codebase. I can extend the tests to cover that.

Re: smith-xyz's `FOR UPDATE OF """Post"""` — I see what alumni's saying about the QueryBuilder/Repository contract. If the intent is that QueryBuilder accepts raw fragments and Repository sanitizes, then the double-escape in `applyFindOptions` is the bug (Repository is passing already-escaped names to a QueryBuilder method that re-escapes). I'd rather not touch that path in this PR — it's a separate concern from the input validation here. Want me to file it as its own issue?

michaelbromley:
@eddieran after internal discussion here's how we want to handle this:

- Keep the useIndex change but shift it to accept string[] instead of string
- The lockTables escaping fix can come separately with more time to understand the implementation (ie leave it as-is before your change)
- The double-escaping bug (Qodo finding) still needs fixing either way

Let us know if you'll have time to do that as we'd like to get the `useIndex` signature change in before we release v1.0 - so in the next day or so would be best. If not we can take over and finish that up.

michaelbromley:
@eddieran I went ahead and pushed some changes as described. Hope that's ok and thanks for your work so far.

alumni (approved):

alumni:
the upgrade guide for MySQL needs a small update (only place where you can use `useIndex`) - will quickly create a commit

PR #12345, opened 2026-04-12
fix: use ALTER COLUMN instead of DROP/ADD for type changes

Ojas2095:
## Summary
Fixes #3357 
The migration generator currently uses `DROP COLUMN` + `ADD COLUMN` when a column's type or length changes. This causes **irrecoverable data loss** in production databases.

## Changes
This PR modifies all 5 major RDBMS drivers to use non-destructive `ALTER` statements:

| Driver | Before | After |
|--------|--------|-------|
| **PostgreSQL** | `DROP COLUMN` + `ADD COLUMN` | `ALTER COLUMN ... TYPE ... USING` |
| **MySQL/MariaDB** | `DROP COLUMN` + `ADD COLUMN` | Uses existing `MODIFY COLUMN` path |
| **SQL Server** | `DROP COLUMN` + `ADD COLUMN` | `ALTER TABLE ... ALTER COLUMN` |
| **Oracle** | `DROP COLUMN` + `ADD COLUMN` | `ALTER TABLE ... MODIFY` |
| **SAP HANA** | `DROP COLUMN` + `ADD COLUMN` | `ALTER TABLE ... ALTER` |

### PostgreSQL (Key Change)
- Moved type/length/isArray checks from the destructive `if` block into the non-destructive `else` block
- - Added `USING` clause for safe data casting when the column type changes (e.g., `varchar` -> `text`)
- - Combined with existing precision/scale change logic
### MySQL, MSSQL, Oracle, SAP HANA
- Removed `type` and `length` from the conditions that trigger column recreation
- - These drivers already have ALTER/MODIFY logic in the `else` block via `isColumnChanged()`
## Test
Added `test/functional/migrations/column-type-change/` with a test that verifies the migration generator produces `ALTER` statements instead of `DROP` when a column type changes.

## Breaking Changes
None. This is a strictly safer approach. Columns that previously would lose data during migration will now be altered in-place.

### DIFF
--- a/src/driver/mysql/MysqlQueryRunner.ts
+++ b/src/driver/mysql/MysqlQueryRunner.ts
@@ -1141,8 +1141,6 @@ export class MysqlQueryRunner extends BaseQueryRunner implements QueryRunner {
         if (
             (newColumn.isGenerated !== oldColumn.isGenerated &&
                 newColumn.generationStrategy !== "uuid") ||
-            oldColumn.type !== newColumn.type ||
-            oldColumn.length !== newColumn.length ||
             (oldColumn.generatedType &&
                 newColumn.generatedType &&
                 oldColumn.generatedType !== newColumn.generatedType) ||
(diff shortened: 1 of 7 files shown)


pkuczynski:
Hey @Ojas2095, thanks for this! We've decided to move forward with #11620 which adds a safety guard that falls back to DROP/ADD for cross-family type changes — a direct `ALTER COLUMN TYPE` can't always handle those. Closing in favor of #11620 — your review and feedback there would be very welcome!

PR #12346, opened 2026-04-13
fix: expand VirtualColumn expression in WHERE and ORDER BY conditions

### LINKED ISSUE #11616: VirtualColumn throws error when used in WHERE conditions
### Issue description

Error when using VirtualColumn in a WHERE clause with createQueryBuilder

### Expected Behavior

The query should expand the VirtualColumn definition and generate SQL similar to:

```sql
SELECT 
       `user`.`id` AS `user_id`, 
       `user`.`firstName` AS "user_firstName",
       `user`.`lastName` AS "user_lastName"
       CONCAT(`user`.`first_name`, ' ', `user`.`last_name`) AS `user_fullName` 
FROM "user" "user"
LEFT JOIN `users` `user` ON `user`.`id`=`member`.`user_id` 
    AND (`user`.`deleted_at` IS NULL) 
WHERE CONCAT("user"."firstName", ' ', "user"."lastName") = 'John Doe';

```

### Actual Behavior

```sql
SELECT 
       `user`.`id` AS `user_id`, 
       `user`.`firstName` AS "user_firstName",
       `user`.`lastName` AS "user_lastName"
       CONCAT(`user`.`first_name`, ' ', `user`.`last_name`) AS `user_fullName` 
FROM `circle_members` `member` 
LEFT JOIN `users` `user` ON `user`.`id`=`member`.`user_id` 
    AND (`user`.`deleted_at` IS NULL) 
WHERE `user`.`fullName` = 'John Doe';
```

`"Unknown column 'user.fullName' in 'where clause'"`

### Steps to reproduce

**User Entity**
```ts
import { Entity, PrimaryGeneratedColumn, Column, VirtualColumn } from "typeorm";

@Entity()
export class User {
  @PrimaryGeneratedColumn()
  id: number;

  @Column()
  firstName: string;

  @Column()
  lastName: string;

  ...

  @VirtualColumn({
    query: (alias) => `CONCAT(${alias}.firstName, ' ', ${alias}.lastName)`
  })
  fullName: string;
}
```

**CircleMember Entity**
```ts
@Entity('circle_members')
export class CircleMember {
  @PrimaryGeneratedColumn({ name: 'id' })
  id: number;

  @Column({ name: 'user_id' })
  userId: number;
 ...
}
```

**QueryBuilder**
```ts
const query = this.dataSource
      .getRepository(CircleMember)
      .createQueryBuilder('member')
      .select(['user.id', 'user.firstName' , 'user.lastName' ,'user.fullName'])
      .leftJoin('member.user', 'user')

if (search) {
      query.andWhere(`user.fullName = :search`, {
        search: 'John Doe',
      });
    }

```

### My Environment

| Operating System    | MacOS |
| Node.js version     |22.16.0 |

### Additional Context

Allow `VirtualColumn` definitions to be expanded anywhere in `QueryBuilder`, not just in `SELECT` clauses.

- [ ] aurora-mysql
- [ ] aurora-postgres
- [ ] better-sqlite3
- [ ] cockroachdb
- [ ] cordova
- [ ] expo
- [ ] mongodb
- [x] mysql
- [ ] nativescript
- [ ] oracle
- [ ] postgres
- [ ] react-native
- [ ] sap
- [ ] spanner
- [ ] sqlite
- [ ] sqlite-abstract
- [ ] sqljs
- [ ] sqlserver

DORI2001:
## Summary

Fixes #11616.

When a `@VirtualColumn()` property is used in `find()` / `findOne()` `where` or `order` options, TypeORM currently references it by its database alias (e.g. `"User"."fullName"`), which does not exist in the database.  The virtual column is a computed expression — not a real column — so it must be inlined wherever it is referenced in SQL.

The fix was already present for **SELECT** clauses (`buildSelection`) and **WHERE** clauses (`buildWhere`), but was missing for **ORDER BY** (`buildOrder`).

### Before

```sql
-- WHERE (already worked — only SQL shown for clarity)
WHERE (CONCAT("User"."firstName", ' ', "User"."lastName")) = $1

-- ORDER BY (broken — column does not exist in DB)
ORDER BY "User"."fullName" ASC
-- ❌ ERROR: column "User"."fullName" does not exist
```

### After

```sql
-- WHERE
WHERE (CONCAT("User"."firstName", ' ', "User"."lastName")) = $1

-- ORDER BY
ORDER BY (CONCAT("User"."firstName", ' ', "User"."lastName")) ASC
-- ✅ Works correctly
```

## Changes

- **`src/query-builder/SelectQueryBuilder.ts`** — in `buildOrder()`, detect `column.isVirtualProperty && column.query` and substitute the expression (wrapped in parentheses) instead of the raw `alias.propertyPath` string, matching the pattern used in `buildWhere()`.
- **`test/functional/columns/virtual-columns/entity/User.ts`** — new minimal entity with `firstName`, `lastName` columns and a `fullName` `@VirtualColumn` defined as `CONCAT(alias."firstName", ' ', alias."lastName")`.
- **`test/functional/columns/virtual-columns/virtual-columns.test.ts`** — three new test cases:
  1. **SQL-level assertion**: the generated SQL contains `CONCAT(` and does _not_ reference `"User"."fullName"`.
  2. **Integration WHERE test**: `findOne({ where: { fullName: "John Doe" } })` finds the correct row.
  3. **Integration ORDER BY test**: `find({ order: { fullName: "ASC" } })` returns rows in correct order without DB error.

## Test plan

- [x] TypeScript compiles with `pnpm tsc --noEmit` (no errors)
- [x] ESLint passes with `pnpm lint` (no new warnings in changed files)
- [ ] Run `test/functional/columns/virtual-columns/virtual-columns.test.ts` against a Postgres (or SQLite) instance to confirm all tests pass


DORI2001:
Pushed a follow-up commit addressing all four issues flagged in the Qodo review:

**1. CONCAT-only test expression (bugs #1 and #4)**
Changed `User.fullName`'s default `VirtualColumn` query from `CONCAT()` to the SQL-standard `||` operator so the test suite passes on SQLite and any other non-MySQL driver. The existing MySQL override in `before()` (which already patched to CONCAT with backtick quoting) continues to cover the MySQL family.

**2. Database-specific SQL assertion**
Replaced the unconditional `expect(sql).to.include("CONCAT(")` with a driver-aware check: MySQL asserts for `CONCAT(` and absence of the raw alias; all other drivers assert that `firstName` appears in the WHERE expression and that `"User"."fullName"` does not.

**3. Virtual ORDER BY breaks pagination with joins (bug #3)**
This was the root cause of the SonarCloud reliability failure. `buildOrder()` stores VirtualColumn ORDER BY as a raw parenthesised SQL expression like `(CONCAT("User"."firstName", ...))`. The existing `createOrderByCombinedWithSelectExpression()` path — used when `skip`/`take` is combined with joins — unconditionally split any criterion containing `.` on that character and called `findAliasByName()` on the first segment, which threw for any expression containing dots.

Added a private `resolveVirtualColumnSelectAlias()` helper that walks the alias metadata to match a raw expression back to the VirtualColumn that produced it, returning the SELECT alias the inner subquery exposes (e.g. `User_fullName`). `createOrderByCombinedWithSelectExpression()` now checks for parenthesised criteria first, maps them through this helper, and references `parentAlias.User_fullName` in both the SELECT list and the ORDER BY object — the same pattern used for regular columns.

**4. `as any` casts**
`fullName` is declared as `fullName?: string` on `User`, so `FindOptionsWhere<User>` and `FindOptionsOrder<User>` already accept it. Removed all three `as any` casts.

DORI2001:
Pushed a follow-up commit (3ceb5fd) addressing the two Qodo bugs from the latest review:

**Bug #1 — Duplicate VC select alias**

Root cause: find-options stores a `select: { fullName: true }` entry as `{ selection: 'User.fullName', aliasName: undefined }` in `expressionMap.selects`. The entity column compiler already converts this to `(expr) AS User_fullName` in the SQL. The old dedup check only compared the raw `selection` string (`'User.fullName' !== '(CONCAT...)'`) and `aliasName` (`undefined !== 'User_fullName'`), so it fell through and injected a *second* `(expr) AS User_fullName` — producing invalid SQL.

Fix: enhanced the `alreadySelected` guard (lines 3533–3581) to also detect `alias.property` selections with no `aliasName`. When such a selection is found, it resolves the property to its column metadata via `findColumnWithPropertyPath`, checks `isVirtualProperty`, and computes the compiled alias with `DriverUtils.buildAlias`. If it matches the VC alias to be injected, the entry is treated as already selected.

**Bug #2 — Virtual ORDER BY with unselected VC**

This scenario (ordering by VC without selecting it) was already handled correctly by the unconditional `vcInnerSelects.push` in `createOrderByCombinedWithSelectExpression`. Once Bug #1's dedup check is accurate, the guard no longer incorrectly suppresses the injection when the VC is absent, so Bug #2's path continues to work.

**Tests added** (`test/functional/columns/virtual-columns/virtual-columns.test.ts`):

- `skip/take + relations + select: { fullName: true } + order: { fullName: 'ASC' }` — covers Bug #1 (no duplicate alias, results are sorted)
- `skip/take + relations + select: { firstName, lastName } + order: { fullName: 'ASC' }` — covers Bug #2 (VC not selected but ORDER BY still resolves)

DORI2001:
Closing — no maintainer engagement despite multiple iterations addressing bot review feedback. Leaving the implementation here for reference.

PR #12347, opened 2026-04-13
fix: add linux/amd64 platform for x86-only images to support Apple Silicon

mixelburg:
Fixes #12322

This PR adds `platform: linux/amd64` to the docker-compose services that do not have ARM64 builds (mysql-5, mssql, oracle, spanner, hanaexpress). This allows Docker Desktop on Apple Silicon (M-series Macs) to use Rosetta emulation and start the containers successfully rather than failing silently or crashing.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -2,6 +2,7 @@ services:
   # mysql
   mysql-5:
     image: "mysql:5.7.44"
+    platform: linux/amd64
     container_name: "typeorm-mysql-5"
     ports:
       - "3306:3306"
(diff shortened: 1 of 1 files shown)


smith-xyz (changes requested):
I like the intent here but I'm not sure its the right fix. Some alternatives:

1. `DOCKER_DEFAULT_PLATFORM=linux/amd64` - env var could do
2. `docker-compose.override.yml` — local per-service overrides, typically gitignored
3. Docker Desktop → "Use Rosetta for x86_64/amd64 emulation" — a checkbox

mixelburg:
Closing — as @smith-xyz noted, the environment variable approach (DOCKER_DEFAULT_PLATFORM) or docker-compose.override.yml is a better solution that doesn't require changes to the shared Dockerfile. These alternatives give per-developer control without affecting CI or other contributors.

PR #12348, opened 2026-04-15
fix(transaction): use unique savepoint names for concurrent nested transactions

fru1tworld:
Concurrent nested transactions via `Promise.all` share the same `transactionDepth` counter, producing duplicate SAVEPOINT names and causing `QueryFailedError` on RELEASE. Generate unique savepoint names via a monotonic counter and skip RELEASE (which destroys sibling savepoints) — the outer COMMIT handles cleanup.

Fixes #12308


alumni:
Sorry @fru1tworld, the bug report is not valid

PR #12352, opened 2026-04-15
fix: handle view entity relations in query load strategy

### LINKED ISSUE #12330: One-to-many relation with view loads wrong results with relationLoadStrategy: query
### Issue description

If an entity has a relationship with a view and the repository tries to find({ relationLoadStrategy: 'query', relations: { myView: true } }) all results are matched with all relations

### Expected Behavior

The results should be the same as the ones with `relationLoadStrategy: 'join'`.

### Actual Behavior

All entities contain the product of all entities' relationships.

### Steps to reproduce

```Typescript
@Entity()
class Group {
  @PrimaryColumn()
  id: number;

  @OneToMany(() => Client, (client) => client.group
  currentClients: Client[];
}

@ViewEntity({ expression: '' })
class Client {
  @ViewColumn()
  id: number;

  @ViewColumn()
  groupId: number;

  @ManyToOne(() => Group)
  group: Group;
}

getRepository(Group).find({ relations: { currentClients: true }, relationLoadStrategy: 'query' });
```

### My Environment

| Node.js version     | 24.14.0  |

### Additional Context

It seems that the relation ids being loaded do not contain the id of the view, since the view has no primary keys. I tried "setting" a primary column in the view by setting 

```Typescript
@Column({ primary: true })
@ViewColumn()
id: number
```

but it did not work. I think that RelationIdLoader can be improved so that it does not require entities having primary columns.

- [ ] aurora-mysql
- [ ] aurora-postgres
- [ ] better-sqlite3
- [ ] cockroachdb
- [ ] cordova
- [ ] expo
- [ ] mongodb
- [ ] mysql
- [ ] nativescript
- [ ] oracle
- [x] postgres
- [ ] react-native
- [ ] sap
- [ ] spanner
- [ ] sqlite
- [ ] sqlite-abstract
- [ ] sqljs
- [ ] sqlserver

aayushbaluni:
## Summary

Fixes #12330

When using `relationLoadStrategy: 'query'` with one-to-many relations backed by view entities, the query strategy produces incorrect join conditions, loading wrong results (cross-join across parents).

## Root Cause

In `RelationIdLoader`, child identifiers are taken from `inverseEntityMetadata.primaryColumns`. View entities often have no primary columns in metadata, so the list is empty. `[].every(() => ...)` is always `true` in JavaScript, meaning each parent receives the union of all children.

## Fix

- Fall back to non-virtual columns when primaryColumns is empty in view entities
- Use the same column list for SELECT aliases and entity matching
- Added regression test under `test/github-issues/12330/`

Made with [Cursor](https://cursor.com)

### DIFF
--- a/src/query-builder/RelationIdLoader.ts
+++ b/src/query-builder/RelationIdLoader.ts
@@ -149,7 +149,6 @@ export class RelationIdLoader {
             inverseColumns = relation.inverseRelation!.joinColumns.map(
                 (column) => column.referencedColumn!,
             )
-        } else {
         }
 
         return entities.map((entity) => {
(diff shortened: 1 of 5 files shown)


michaelbromley:

```
git fetch origin
```

PR #12354, opened 2026-04-16
fix: preserve user-defined shared join columns in change set

PreAgile:
## Summary

- In `computeDiffColumns`, the relational-column short-circuit now only applies to auto-created join columns (where `propertyName` matches the owning relation's `propertyName`).
- User-defined columns that also serve as join columns keep their explicit entity value in the change set, preventing them from being silently omitted when the relation is set to a scalar plain id.
- Without the fix, updating a shared join column while a composite-FK relation is set drops the column from the UPDATE — causing `NOT NULL constraint failed` or the value reverting silently.

Fixes #12234

## Test plan

- [x] Added `test/github-issues/12234/` with 3 regression tests:
  - Scalar plain-id on composite FK → shared join column preserved on UPDATE
  - INSERT with only one composite-FK relation set → shared join column not omitted
  - Direct column value update → shared join column preserved
- [x] Verified the scalar test fails without the fix (`NOT NULL constraint failed: post.tenantId`)
- [x] `pnpm run compile` — pass
- [x] `pnpm run typecheck` — pass
- [x] `prettier --check` on changed files — pass
- [x] Existing persistence test suite (84 tests) — no regressions


alumni (approved):

gioboa (approved):
Thanks @PreAgile for your help 🤗

PR #12355, opened 2026-04-16
style(eslint): enable no-useless-assignment and no-loss-of-precision from eslint/recommended

pkuczynski:
Remove two `eslint/recommended` rule overrides that were temporarily downgraded to `"warn"` while violations were being cleaned up:

- **`no-useless-assignment`** — all violations were fixed in #12102.
- **`no-loss-of-precision`** — 3 remaining violations in `column-types-sqlite.test.ts` used `8223372036854775807` which exceeds `Number.MAX_SAFE_INTEGER` and silently rounds to `8223372036854776000`. Replaced with `Number.MAX_SAFE_INTEGER` to test the same "large integer" intent without precision loss.


gioboa (approved):

test/functional/database-schema/column-types/sqlite/column-types-sqlite.test.ts
```diff
@@ -34,12 +34,12 @@ describe("database schema > column types > sqlite", () => {
                 post.integer = 2147483647
                 post.int = 2147483647
                 post.int2 = 32767
-                post.int8 = 8223372036854775807
+                post.int8 = Number.MAX_SAFE_INTEGER
```
alumni, line 37: should we maybe set a BigInt value or a string?

pkuczynski: I was thinking about it, but then it would not be comatible with `number` on the test, since ts would require `bigint` type (https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-2.html#bigint). So I thought I will get this fixed when I start working on true support for BigInt type few weeks from now, which we have few unsuccessful open PRs...

alumni: I think for data types we need to have a separate discussion.

At the moment TypeORM is trying to unify the behavior across all DBs (and normalize the value that the client returns to something that is common to every driver) - which is what ORMs should do.

The problem is that we have a lot of issues with this value transformation (especially with date/timestamps and timezones).

For 64-bit integers (`bigint`/`int8`) and all `decimal` types, precision loss will occur if they are converted to `number`s. 64bit integers should always be converted to JS `BigInt` (or a `string`, which most clients return because `BigInt` isn't very popular yet), while `decimal`s should always be kept as a `string` (only via an external plugin maybe transformed to a `decimal.js`/`bignumber.js`, but not by default).

If making `Post.int8` a `string` and setting `post.int8 = "8223372036854775807"` doesn't work, we should fix that issue before enabling `no-loss-of-precision` :P

michaelbromley:

```
git fetch origin
```

PR #5536, opened 2020-02-17
Add driver option to address Expo API change.

cuibonobo:
This pull request adds a new driver option that allows the user to pass `require('expo-sqlite')` into the driver. The Expo SQLite API changed to use a new `expo-sqlite` module instead of having a `SQLite` object be available on `window`.

I did not want to put `require('expo-sqlite')` directly in the driver code base because [Metro bundler does not allow dynamic imports](https://github.com/facebook/metro/issues/52), which means that all `typeorm/browser` users would be forced to import `expo-sqlite` into their project. (A similar thing happens with `react-native-sqlite-storage` for the React Native driver.)

I have added documentation for the new option, but I would appreciate if someone could check the Chinese translation.

Fixes #4846

Issue #4846, opened 2019-10-02
(Expo) TypeError: undefined is not an object (evaluating '_this.sqlite.openDatabase')


crofoot:
**Issue type:**

[x ] question
[ ] bug report
[ ] feature request
[ ] documentation issue

**Database system/driver:**

[ ] `cordova`
[ ] `mongodb`
[ ] `mssql`
[ ] `mysql` / `mariadb`
[ ] `oracle`
[ ] `postgres`
[ ] `cockroachdb`
[ ] `sqlite`
[ ] `sqljs`
[ ] `react-native`
[x] `expo`

**TypeORM version:**

[x ] `latest`
[ ] `@next`
[ ] `0.x.x` (or put your version here)

**Steps to reproduce or a small repository showing the problem:**

Issue: I haven't seen any examples with typeorm & expo. I am trying to get a basic application working/connecting to a sqlite db, but I keep running into this issue. 

 ``` 
[Unhandled promise rejection: TypeError: undefined is not an object (evaluating '_this.sqlite.openDatabase')]
- node_modules\typeorm\browser\driver\expo\ExpoDriver.js:63:56 in <unknown>
````

Any insight would be great, I provided the code I am testing with below, and I bootstrapped the project with expo-cli. 

App.tsx 
``` 
import 'reflect-metadata';
import React from 'react';
import List from './src/List';
import { Item } from './entities/item';
import { View, Button, Text } from 'react-native';
import { createConnection, getRepository } from "typeorm";

interface State {
  data: Item[];
}

interface Props {

}


export default class App extends React.Component<Props, State> {

  constructor(props: Props) {
    super(props);

    this.state = {
      data: []
    };

    this.connect();

  }


  async connect() {
    return await createConnection({
      database: "test",
      entities: [
        Item
      ],
      synchronize: true,
      type: "expo",

    });
  }


  private addString = async () => {
    let item = new Item();
    item.name = Math.random().toString(36).substring(2, 15) + Math.random().toString(36).substring(2, 15);
    await getRepository(Item).save(item);
    this.setState({
      data: [...this.state.data, item]
    });
  }

  render() {
    return (
      <View style={{ flex: 1 }}>
        <Button title='Add' onPress={this.addString} />
        <List data={this.state.data} />
      </View>
    );
  }
}
```

item.ts 
```
import {Entity, PrimaryGeneratedColumn, Column} from "typeorm";

@Entity('item')
export class Item {

    @PrimaryGeneratedColumn()
[...]

PR #11860, opened 2025-12-20
refactor!: remove legacy expo driver

G0maa:
### Description of change

Fixes #11854 

### Pull-Request Checklist

-   [x] Code is up-to-date with the `next` branch
-   [x] This pull request links relevant issues as `Fixes #00000`
-   [ ] There are new or updated tests validating the change (`tests/**.test.ts`)
-   [x] Documentation has been updated to reflect this change (`docs/docs/**.md`)

PR #11107, opened 2024-10-20
feat: Support Expo SQLite Next

pmk1c:
### Description of change

Fixes #11027

This adds support for the new Expo SQLite API. Since the API is in the process of being changed "in place", I decided, not to add another driver, but add a ExpoDriverFactory, that creates either the ExpoLegacyDriver (for the old API) or the ExpoDriver (for the new API).

This should allow for a smooth transition, when the API is being replaced with the next Expo Update.

The new Expo API does not use transactions everywhere anymore, but we can use prepared statements, which is quite nice.

### Pull-Request Checklist

- [x] Code is up-to-date with the `master` branch
- [x] `npm run format` to apply prettier formatting
- [x] `npm run test` passes with this change
- [x] This pull request links relevant issues as `Fixes #0000`
- [ ] There are new or updated unit tests validating the change
      *I don't see any tests / test setup for Expo driver...*
- [ ] Documentation has been updated to reflect this change
      *No updates needed*
- [ ] The new commits follow conventions explained in [CONTRIBUTING.md](https://github.com/typeorm/typeorm/blob/master/CONTRIBUTING.md)

PR #12363, opened 2026-04-17
fix(expo): auto-load expo-sqlite driver via loadDependencies()

pkuczynski:
Brings the Expo driver in line with every other TypeORM driver. Users no longer need `driver: require("expo-sqlite")` — bumping Expo to SDK v52+ is all that's required.

- **`ExpoDriver.loadDependencies()`** — `this.sqlite = this.options.driver ?? require("expo-sqlite")`, same pattern as `ReactNativeDriver`. Metro statically bundles the literal require; `PlatformTools.load()` can't be used here because it's a throwing stub in the browser build that Expo consumes.
- **`ExpoDataSourceOptions.driver`** — now optional. Passing it still works as an override (for patched forks or custom wrappers).
- **Pre-v52 SDK detection** — if `openDatabaseAsync` is missing on the loaded module, `loadDependencies()` throws a clear error during `DataSource` construction.
- **Missing-module narrowing** — `isExpoSqliteNotFoundError` checks both `err.code === "MODULE_NOT_FOUND"` AND `err.message.includes("expo-sqlite")`, so transitive `MODULE_NOT_FOUND` failures (a dependency of `expo-sqlite` being unresolved) bubble up unchanged instead of being misreported as "install expo-sqlite".
- **Migration guide** — rewritten to "no code change needed, bump Expo to SDK v52+".
- **Codemod — `datasource-expo` transform** — instead of flagging the redundant `driver` line with a TODO (previous approach, per @alumni's suggestion), the codemod now deletes it outright. Only the exact `require("expo-sqlite")` shape is removed; custom wrappers, `.default` re-exports, and non-Expo configs are left alone.
- **Codemod — dep bump** — `expo` added to `upgrades` (minVersion `^52.0.0`, target `^55.0.0`) so running the codemod auto-bumps the SDK.

Context: @alumni flagged that Expo was the only driver without `loadDependencies()`. The `driver` option was made effectively mandatory in #5536 (Feb 2020, fixing #4846) as a Metro-resolution workaround that no longer applies.


gioboa (approved):

alumni:
Hey @pkuczynski good catch with `require('expo-sqlite')`, but this is what TypeORM should do, not the user. We might need to fix something 😄

All the other drivers have a method `loadDependencies()` which calls `PlatformTools.load()` to load the driver package or other necessary packages. I'm not sure why we don't have it in Expo, maybe it was done like this to ease the transition between the new and the old driver. CC @G0maa

pkuczynski:
Thanks Lucian — I initially followed the existing pattern with a mandatory `driver` without questioning it, but your comment prompted me to look into why Expo was the only driver without a `loadDependencies()`. Here's what I found.

**The history**: the `driver` option was made effectively required in [#5536](https://github.com/typeorm/typeorm/pull/5536) (Feb 2020, fixing [#4846](https://github.com/typeorm/typeorm/issues/4846)) because Expo had changed its SQLite module surface and `require("expo-sqlite")` from inside the typeorm package stopped resolving reliably. The workaround at the time was to stop requiring it ourselves and make the user pass it in. The pattern was carried forward through [#11107](https://github.com/typeorm/typeorm/pull/11107) when the new async API was added, and again through [#11860](https://github.com/typeorm/typeorm/pull/11860) when legacy was removed — in both cases without revisiting the underlying assumption.

**The 2020 constraint is obsolete**: Expo runs on stock Metro today, Metro statically resolves `require("literal-string")` at bundle time, and `ReactNativeDriver` has been doing `require("react-native-sqlite-storage")` inside `loadDependencies()` since 2019 without issue. There's no reason `expo-sqlite` can't follow the same pattern.

One subtlety worth flagging: in the browser build (`./browser/index.js`, which is what Expo users get via the `"react-native"` export condition), `PlatformTools.load()` is the throwing stub from `BrowserPlatformTools.template`. So the fix can't go through `PlatformTools.load` — it needs to be a bare `require("expo-sqlite")` directly in the driver, matching `ReactNativeDriver`.

**Reframed PR**: `loadDependencies()` in `ExpoDriver` with the bare `require`, `driver` becomes optional in `ExpoDataSourceOptions`, pre-v52 SDK detection moved inside `loadDependencies()`, migration guide rewritten to "no code change needed, bump SDK", `datasource-expo` source transform dropped, and `expo` added to the codemod's dependency-auto-upgrade map so `package.json` gets bumped to `^55.0.0` automatically. End result: v1 users get the same no-configuration behavior as every other driver, and the codemod handles the SDK bump transparently.

CC @G0maa for the driver change, since you did #11860.

docs/docs/releases/1.0/02-upgrading-from-0.3.md
```diff
@@ -265,23 +265,30 @@ Also note the default behavior changes in pool configuration:
...
-// Before
+// Works in both v0.3 and v1.0 (with Expo SDK v52+)
 new DataSource({
     type: "expo",
     database: "db.sqlite",
 })
+```
 
-// After — use Expo SDK v52+ with the modern async API
+If you were previously passing `driver` explicitly it continues to work as an override — useful for `patch-package` setups or custom wrappers:
```
alumni: `patch-package` reference makes no sense - please remove it.

IMO you can also remove the rest of the comment - this doc page is for users that are upgrading, so rather removing that option, not adding it (so there shouldn't be any instructions about adding it) :P

alumni (approved):

PR #12367, opened 2026-04-17
refactor(data-source)!: rename dropDatabase to dropAllEntityTables

naorpeled:
## Summary
- Rename `DataSource#dropDatabase()` to `DataSource#dropAllEntityTables()`. The method never actually dropped the database on most drivers — it calls `queryRunner.clearDatabase()` to clear all entity tables — so the new name describes what it actually does.
- Update internal callers (`DataSource#synchronize`, `dropSchema` init path) and the `schema:drop` CLI command.
- Update the DataSource API docs and a related test description.

## Breaking Change
`DataSource#dropDatabase()` has been removed. Consumers must use `DataSource#dropAllEntityTables()` instead. Behaviour is unchanged.

## Context
Resolves the long-standing `// TODO rename` on `src/data-source/DataSource.ts`. TypeORM is still `1.0.0-pre` and has not released v1, so the deprecation-shim stage is being skipped in favour of a clean rename.

## Test plan
- [ ] `pnpm run typecheck` (passes locally)
- [ ] `pnpm run test:fast` with the default SQLite ormconfig
- [ ] Spot-check `pnpm exec typeorm schema:drop -d <path>` against a SQLite db

### DIFF
--- a/docs/docs/data-source/4-data-source-api.md
+++ b/docs/docs/data-source/4-data-source-api.md
@@ -57,12 +57,13 @@ await dataSource.destroy()
 await dataSource.synchronize()
 ```
 
-- `dropDatabase` - Drops the database and all its data.
+- `dropAllEntityTables` - Drops every table managed by this DataSource along with its data.
+  Despite the previous name (`dropDatabase`), this does not drop the database itself on most drivers — it clears all entity tables.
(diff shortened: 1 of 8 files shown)


naorpeled:
Will change this in the future, as discussed with @michaelbromley

PR #12369, opened 2026-04-17
chore(metadata-args): remove stale todo on ColumnMetadataArgs.mode

naorpeled:
## Summary
- Remove the stale `todo: find name better then "mode"` JSDoc comment on `ColumnMetadataArgs.mode`.
- Decision: keep the property named `mode`. There's no clearer name that justifies the churn of a rename across the decorator surface and every consumer that reads from `getMetadataArgsStorage().columns`.

Not a breaking change — no API or behavior change, just a one-hunk comment cleanup.

## Test plan
- [x] `pnpm run typecheck`
- [ ] Visual diff review

### DIFF
--- a/src/metadata-args/ColumnMetadataArgs.ts
+++ b/src/metadata-args/ColumnMetadataArgs.ts
@@ -17,8 +17,6 @@ export interface ColumnMetadataArgs {
 
     /**
      * Column mode in which column will work.
-     *
-     * todo: find name better then "mode".
      */
     readonly mode: ColumnMode
 


pkuczynski (approved):

gioboa (approved):

alumni:
"column mode" is rather "storage type" (stored, virtual, virtual property).

the to-do still makes sense, "mode" does not reflect its purpose.

naorpeled:
> "column mode" is rather "storage type" (stored, virtual, virtual property).

Do you think we should change it now before v1?
Or keep it for later?

alumni:
The rename needs to be done in a major version for sure :) I would do it if you have worked for a while with that part of code, you are familiar with it and have a good idea for a new name.

I would not remove TODOs just for the sake of removing them.

naorpeled:
> The rename needs to be done in a major version for sure :) I would do it if you have worked for a while with that part of code, you are familiar with it and have a good idea for a new name.
> 
> I would not remove TODOs just for the sake of removing them.

Fair point, closing this for now, we can revisit in the future 🙏

PR #12372, opened 2026-04-17
fix(codemod): scope v1 transforms to typeorm imports and skip .d.ts files

pkuczynski:
Two hardening fixes for the v1 codemod discovered during a full audit of all 32 transforms.

### Scope v1 transforms to typeorm imports

A number of v1 transforms previously rewrote any file matching their pattern, regardless of whether the file imported TypeORM. Examples:

- `datasource-sqlite-type` / `datasource-sqlite-options` — any `type: "sqlite"`, `busyTimeout`, or `flags` property got mutated (e.g. HTTP clients with `flags`, Cypress configs with `type: "sqlite"`)
- `repository-exist` — any `.exist(...)` call got renamed to `.exists(...)`
- `find-options-string-select` / `find-options-string-relations` — any `select: [...]` or `relations: [...]` property got rewritten into object form
- …and 12 more

All 17 affected transforms now bail out early via `fileImportsFrom(root, j, "typeorm")` — matching the pattern already used by `datasource-sap`, `file-logger`, and `connection-options-reader`.

Fixtures for each transform were updated to include a `typeorm` import so the tests still exercise the guarded path.

### Skip `.d.ts` files

The codemod CLI passed `extensions: "ts,tsx,js,jsx"` with no `ignorePattern` default. Because `.d.ts` files match the `.ts` extension, jscodeshift was visiting ambient declaration files. Transforms that rename identifiers (e.g. `connection-to-datasource` renaming `Connection` → `DataSource`) would corrupt declaration files that re-export TypeORM types.

Added `**/*.d.ts` as a default ignore pattern. User-supplied `--ignore` patterns are merged with the default, not replaced.

### DIFF
--- a/packages/codemod/README.md
+++ b/packages/codemod/README.md
@@ -26,14 +26,16 @@ npx @typeorm/codemod v1 --list
 
 ## Options
 
-| Option               | Short | Description                                         |
-| -------------------- | ----- | --------------------------------------------------- |
-| `--dry`              | `-d`  | Dry run (show changes without writing)              |
-| `--help`             | `-h`  | Show help                                           |
-| `--ignore <pattern>` | `-i`  | Glob pattern to exclude files (repeatable)          |
-| `--list`             | `-l`  | List available transforms                           |
-| `--transform <name>` | `-t`  | Run a specific transform only                       |
-| `--workers <num>`    | `-w`  | Number of worker processes (default: CPU cores - 1) |
+| Option               | Short | Description                                                           |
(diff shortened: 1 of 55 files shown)


OSA413 (approved):
Shouldn't there be a `.d.ts` files for tests?

gioboa (approved):

pkuczynski:
Good point — the `.d.ts` skip lives in the CLI, so the transform fixtures can't really cover it. Added a small unit test for the default ignore and the merge with `--ignore` in c075ad0.

PR #12373, opened 2026-04-17
fix(codemod): rewrite typeorm re-exports in barrel files

pkuczynski:
None of the v1 codemod transforms were inspecting `ExportNamedDeclaration` nodes, so a user's barrel file containing `export { Connection, createConnection } from "typeorm"` would silently survive the migration — leaving downstream consumers with broken imports.

### New shared helpers in `ast-helpers.ts`

- **`removeReExportSpecifiers(root, j, moduleName, names)`** — deletes named specifiers from `export { ... } from "module"` and removes the whole declaration if empty. Used by transforms that delete removed APIs.
- **`renameReExportSpecifiers(root, j, moduleName, renames)`** — renames specifiers. Aliased re-exports (`export { X as Y }`) keep the exported name so downstream consumers don't break; only the local name is updated.

### Wiring

- `connection-to-datasource` — renames `Connection`/`ConnectionOptions`/all driver-specific `*ConnectionOptions` types in re-exports
- `global-functions` — removes re-exports of `createConnection`, `getConnection`, `getManager`, etc.
- `use-container` — removes re-exports of `useContainer`, `getFromContainer`, `ContainerInterface`, `ContainedType`, `UseContainerOptions`
- `relation-count` — removes `RelationCount` re-export
- `repository-abstract` — removes `EntityRepository`, `AbstractRepository`, `getCustomRepository` re-exports (moved `getCustomRepository` into the same remove set since it was previously only handled in imports)
- `query-builder-where-expression` — renames `WhereExpression` → `WhereExpressionBuilder` in re-exports
- `mongodb-types` — redirects `export { ObjectId } from "typeorm"` to `export { ObjectId } from "mongodb"`

Fixture coverage added for `connection-to-datasource` (rename + aliased rename), `global-functions` (remove), and `mongodb-types` (redirect).

### DIFF
--- a/packages/codemod/test/transforms/v1/fixtures/global-functions/global-functions.input.ts
+++ b/packages/codemod/test/transforms/v1/fixtures/global-functions/global-functions.input.ts
@@ -3,3 +3,6 @@ import { getRepository, getManager, createQueryBuilder } from "typeorm"
 const repo = getRepository(User)
 const manager = getManager()
 const qb = createQueryBuilder("user")
+
+// Barrel re-exports of removed globals should be deleted
+export { getRepository, createConnection } from "typeorm"
--- a/packages/codemod/test/transforms/v1/fixtures/global-functions/global-functions.input.ts
+++ b/packages/codemod/test/transforms/v1/fixtures/global-functions/global-functions.input.ts
@@ -17,5 +17,10 @@ const ds = getConnection()
 // Named connections removed in v1 — argument is dropped, TODO flags manual reconfiguration
 const secondary = getConnection("secondary")
 
-// Barrel re-exports of removed globals should be deleted
-export { getRepository, createConnection } from "typeorm"
+// Barrel re-exports: removed globals are stripped, other names are preserved
+export {
+    DataSource,
+    getRepository,
+    createConnection,
+    EntityManager,
+} from "typeorm"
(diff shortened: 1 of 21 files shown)


packages/codemod/test/transforms/v1/fixtures/global-functions/global-functions.input.ts
```diff
@@ -3,3 +3,6 @@ import { getRepository, getManager, createQueryBuilder } from "typeorm"
 const repo = getRepository(User)
 const manager = getManager()
 const qb = createQueryBuilder("user")
+
+// Barrel re-exports of removed globals should be deleted
+export { getRepository, createConnection } from "typeorm"
```
alumni: Seems to not work?

Output was not updated?

gioboa (approved):

packages/codemod/test/transforms/v1/fixtures/global-functions/global-functions.output.ts
```diff
@@ -10,3 +10,6 @@ const ds = dataSource
 
 // Named connections removed in v1 — argument is dropped, TODO flags manual reconfiguration
 const secondary = dataSource
+
+// Barrel re-exports: removed globals are stripped, other names are preserved
+export { DataSource, EntityManager } from "typeorm"
```
alumni, line 15: Maybe it should insert one of those `TODO` comments to warn the user and pass the decision to them.

alumni (approved):

PR #12376, opened 2026-04-17
feat(codemod): flag removed ConnectionManager class constructions

pkuczynski:
The `ConnectionManager` class was removed in v1 with no direct replacement — users must create and manage `DataSource` instances directly. The existing `global-functions` transform already handled the `getConnectionManager()` function, but `new ConnectionManager()` constructions were silently left in place.

### What this transform does

- Collects local names bound to TypeORM's `ConnectionManager` (including aliased imports like `import { ConnectionManager as CM } from "typeorm"`)
- Flags every `new ConnectionManager(...)` / `new CM(...)` with a TODO comment on the enclosing statement (variable declaration, return, export default / named declaration, class property, expression statement)
- Removes `ConnectionManager` from the `typeorm` import so the TypeScript error from the removed class doesn't stack on top of the TODO
- Scoped: only runs when the file imports `ConnectionManager` from `"typeorm"`

### Example

Input:
```ts
import { ConnectionManager } from "typeorm"
const manager = new ConnectionManager()
```

Output:
```ts
// TODO(typeorm-v1): `ConnectionManager` was removed — create and manage `DataSource` instances directly instead — there is no replacement class
const manager = new ConnectionManager()
```

### Follow-up

Re-export handling (`export { ConnectionManager } from "typeorm"`) is intentionally not included here — it will be covered once the re-export helpers from #12373 land on master.

### DIFF
--- a/packages/codemod/src/transforms/ast-helpers.ts
+++ b/packages/codemod/src/transforms/ast-helpers.ts
@@ -92,10 +92,12 @@ export const fileImportsFrom = (
 
 /**
  * Returns the set of local identifiers bound to a given named export.
- * Handles direct imports and aliased imports:
+ * Handles ESM direct/aliased imports and CommonJS destructured requires:
  *
  *   import { RelationCount } from "typeorm"         → { "RelationCount" }
  *   import { RelationCount as RC } from "typeorm"   → { "RC" }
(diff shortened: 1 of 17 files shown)


OSA413 (approved):
Good to know that you can test sub-packages in the monorepo

gioboa (approved):

PR #12377, opened 2026-04-17
fix(codemod): handle aliases, quoted keys, and ObjectProperty variants

pkuczynski:
Four partial-coverage fixes for v1 transforms uncovered during a full audit. Each addresses a case where a transform silently skipped user code it should have handled.

### Alias blindness

`@RelationCount` and the global-function transforms matched decorators and calls by the canonical name only. `import { RelationCount as RC } from "typeorm"` → `@RC(...)` was silently skipped while the `RelationCount` import was still removed — leaving code that no longer compiled.

A new `getLocalNamesForImport(root, j, moduleName, importedName)` helper in `ast-helpers.ts` returns the set of local identifiers bound to a named export, including aliases. `relation-count` and `global-functions` now resolve this set and match against it.

### `find-options-lock-modes` — `Property` / `ObjectProperty` mismatch

The `{ lock: { mode: "pessimistic_partial_write" } }` find-options migration looked up `j.ObjectProperty` but walked the grandparent checking for `"Property"`. On Babel-parsed files (what the CLI uses via the `tsx` parser) the grandparent is `ObjectProperty`, so the entire lock-object migration silently no-op'd. Fixed by accepting both node shapes and visiting both property types. Also adds the missing `fileImportsFrom` scope guard.

### Quoted-key handling

`removeObjectProperties` matched identifier keys only, so `{ "width": 9 }` wasn't matched even though it's valid JS. The helper now matches both `name:` and `"name":` keys. Targeted fixes applied to `column-readonly`, `column-unsigned-numeric` (for the `type` key lookup), `datasource-sap`, and `query-builder-or-update` — all of which had their own hardcoded `Identifier` key checks.

### Fixtures

Extended fixtures with alias cases (`relation-count`, `global-functions`), a lock-object case (`find-options-lock-modes`), and a second decorator (`column-readonly`, `column-width-zerofill`).

### DIFF
--- a/packages/codemod/test/transforms/v1/fixtures/global-functions/global-functions.output.ts
+++ b/packages/codemod/test/transforms/v1/fixtures/global-functions/global-functions.output.ts
@@ -2,3 +2,4 @@
 const repo = dataSource.getRepository(User)
 const manager = dataSource.manager
 const qb = dataSource.createQueryBuilder("user")
+const postRepo = dataSource.getRepository(Post)
--- a/packages/codemod/src/transforms/ast-helpers.ts
+++ b/packages/codemod/src/transforms/ast-helpers.ts
@@ -56,6 +56,40 @@ export const fileImportsFrom = (
 }
 
 /**
+ * Returns the set of local identifiers bound to a given named export.
+ * Handles direct imports and aliased imports:
+ *
+ *   import { RelationCount } from "typeorm"         → { "RelationCount" }
+ *   import { RelationCount as RC } from "typeorm"   → { "RC" }
+ */
+export const getLocalNamesForImport = (
+    root: Collection,
+    j: JSCodeshift,
+    moduleName: string,
+    importedName: string,
+): Set<string> => {
+    const localNames = new Set<string>()
+    root.find(j.ImportDeclaration, {
+        source: { value: moduleName },
+    }).forEach((importPath) => {
+        for (const spec of importPath.node.specifiers ?? []) {
+            if (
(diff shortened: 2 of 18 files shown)


packages/codemod/test/transforms/v1/fixtures/global-functions/global-functions.output.ts
```diff
@@ -2,3 +2,4 @@
 const repo = dataSource.getRepository(User)
 const manager = dataSource.manager
 const qb = dataSource.createQueryBuilder("user")
+const postRepo = dataSource.getRepository(Post)
```
alumni, line 5: Was it necessary to transform this?

In the other test (for relation count), RC is not transformed. 🤷🏻‍♂️

alumni (approved):

pujux (approved):
I'm not too deep into the codemod code but this looks fine to me 👍

gioboa (approved):

packages/codemod/test/transforms/v1/fixtures/global-functions/global-functions.output.ts
pkuczynski: Good catch — the asymmetry was a bug: the alias `@RC` should have been flagged too, but `root.find(j.Decorator)` returns nothing with the `tsx` parser (decorators live on `ClassProperty.decorators` and the default visitor doesn't descend into them), so the whole TODO path was silently skipped. Fixed in fe90e2d by walking `ClassProperty` explicitly — `@RC` now gets the same TODO as `@RelationCount`.

PR #12383, opened 2026-04-17
fix(codemod): rename .connection on EntityMetadata, ColumnMetadata, IndexMetadata

pkuczynski:
Reported by Lucian: `meta.connection.getMetadata(...)` was left unchanged when `meta` is typed `EntityMetadata`. Same for `ColumnMetadata` and `IndexMetadata`.

`connection-to-datasource` already renames `.connection` → `.dataSource` on typed references to `QueryRunner`, `EntityManager`, `Repository`, `TreeRepository`, `MongoRepository`, and the `*QueryBuilder` classes. The metadata classes were missed — their `.connection` property was renamed to `.dataSource` in #12249, same breaking change.

Extends `typesWithConnectionProp` with `EntityMetadata`, `ColumnMetadata`, and `IndexMetadata`. Fixture gains typed-parameter cases for `EntityMetadata` and `ColumnMetadata` to pin the behavior.

### DIFF
--- a/packages/codemod/src/transforms/v1/connection-to-datasource.ts
+++ b/packages/codemod/src/transforms/v1/connection-to-datasource.ts
@@ -54,6 +54,10 @@ export const connectionToDataSource = (file: FileInfo, api: API) => {
         "DeleteQueryBuilder",
         "SoftDeleteQueryBuilder",
         "RelationQueryBuilder",
+        // Metadata classes (renamed in #12249)
+        "EntityMetadata",
+        "ColumnMetadata",
+        "IndexMetadata",
     ])
 
     // Collect local names imported from "typeorm" that need renaming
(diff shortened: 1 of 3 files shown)


naorpeled (approved):

alumni:
I actually reported it in the constructor (because I have dynamic entities):

- `new EntityMetadata({ connection, ... }` should be `new EntityMetadata({ dataSource: connection, ... }` (`connection` was renamed in constructor options).
- `new ColumnMetadata({ connection, entityMetadata, ... })` should be `new ColumnMetadata({ entityMetadata, ... })` (`connection` was removed from constructor options, because it can be taken from `entityMetadata`).

But it's not really important, it's probably less effort to change those by hand than implement it in the codemod :)

alumni (approved):

gioboa (approved):

PR #12384, opened 2026-04-17
test(codemod): pin that find-options string transforms leave non-array values alone

pkuczynski:
Reported by Lucian: `find({ select: computeSelect() })` wasn't rewritten and it wasn't clear whether the codemod had stopped processing the file or simply couldn't migrate a dynamic value. The latter is correct — `find-options-string-select` (and `-relations`) can only rewrite statically-known string arrays; a function call returns at runtime so the transform can't produce an equivalent object literal at AST time.

Extends each fixture with two additional cases:

- `find({ select: computeSelect() })` — value is a call expression, left unchanged
- `find({ select: { id: true } })` — value is already the v1 object shape, left unchanged

Pins the current (correct) early-return behavior so a future regression fails loudly.

### DIFF
--- a/packages/codemod/test/transforms/v1/fixtures/find-options-string-select/find-options-string-select.input.ts
+++ b/packages/codemod/test/transforms/v1/fixtures/find-options-string-select/find-options-string-select.input.ts
@@ -1,3 +1,14 @@
 const users = await repository.find({
     select: ["id", "name", "email"],
 })
+
+// Should NOT be transformed — value is a function call, not an array literal.
+// The codemod can only rewrite statically-known string arrays.
+const dynamic = await repository.find({
+    select: computeSelect(),
+})
+
+// Should NOT be transformed — pre-existing object-style already uses v1 shape.
+const explicit = await repository.find({
+    select: { id: true, name: true },
+})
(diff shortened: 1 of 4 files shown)


packages/codemod/test/transforms/v1/fixtures/find-options-string-select/find-options-string-select.input.ts
```diff
@@ -1,3 +1,14 @@
...
+
+// Should NOT be transformed — value is a function call, not an array literal.
+// The codemod can only rewrite statically-known string arrays.
+const dynamic = await repository.find({
+    select: computeSelect(),
+})
+
+// Should NOT be transformed — pre-existing object-style already uses v1 shape.
+const explicit = await repository.find({
+    select: { id: true, name: true },
```
alumni, line 19: Can we add a test with `select: ["id", "name"]` after `computeSelect()`?

For some reason `select: ["id", "name"]` is not handled for me (with the version published on npm), but not sure why yet.

alumni (approved):

gioboa (approved):

packages/codemod/test/transforms/v1/fixtures/find-options-string-select/find-options-string-select.input.ts
pkuczynski: Good call — added a dedicated fixture case in 5bd3aa8 that has both `select: computeSelect()` and `select: ["id", "name"]` in the same file. The array case does still get rewritten to the object shape independently.

PR #12386, opened 2026-04-18
fix: do not run npm install

alumni:
### Description of change

Remove the `npm install` command that was executed automatically.

This was always assuming the developers are using `npm` (or `yarn`), but there are lot more package managers. Normally we should also let the users verify what they install and not force certain packages on them, e.g. if they use `pnpm`, they could have `minimumReleaseAge` which npm ignores.

Our docs (the "Initialize a new TypeORM project" section on the "Using CLI" page) was already instructing users to run `npm install` - so documentation does not need updating (this is just a fix to get the intended and documented behavior).

NX for example does not install anything, it instructs the users to verify the git diff and do the rest themselves.

Also: this will stop our tests from interfering with our own `node_modules` (and reduce CI time).

### Pull-Request Checklist

### DIFF
--- a/.husky/commit-msg
+++ b/.husky/commit-msg
@@ -1,1 +1,0 @@
-[ -x .git/hooks/commit-msg ] && .git/hooks/commit-msg "$1"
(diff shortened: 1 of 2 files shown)


pkuczynski (approved):
Makes perfect sense to me!

Cprakhar (approved):

gioboa (approved):

Issue #12357, opened 2026-04-16
Capacitor driver queries should return affected rows


jacobg:
### Feature Description

The Capacitor driver doesn't return affected rows for queries, which is useful for applications, e.g., how many rows were deleted or updated by a query?

This feature was recently added to the Cordova driver:
https://github.com/typeorm/typeorm/blob/ecb00f91865b143348eeaab1523fcde3b0298d5f/src/driver/cordova/CordovaQueryRunner.ts#L117

Here was the merged PR:
https://github.com/typeorm/typeorm/pull/10873

### The Solution

We should add the same thing to the Capacitor driver.

### Considered Alternatives

Not aware of any alternative solutions.

### Relevant Database Driver(s)

- [ ] aurora-mysql
- [ ] aurora-postgres
- [ ] better-sqlite3
- [ ] cockroachdb
- [ ] cordova
- [ ] expo
- [ ] mongodb
- [ ] mysql
- [ ] nativescript
- [ ] oracle
- [ ] postgres
- [ ] react-native
- [ ] sap
- [ ] spanner
- [ ] sqlite
- [ ] sqlite-abstract
- [ ] sqljs
- [ ] sqlserver

### Are you willing to resolve this issue by submitting a Pull Request?

Yes, I have the time, and I know how to start.

jacobg:
BTW the issue template `Relevant Database Driver(s)` is missing `Capacitor`.

Cprakhar:
@jacobg Capacitor already returns affected rows:
https://github.com/typeorm/typeorm/blob/5de549019b1056b8437bd5352abab377adecc15a/src/driver/capacitor/CapacitorQueryRunner.ts#L99-L108

jacobg:
@Cprakhar Sorry, you're right. My bad.

PR #12387, opened 2026-04-19
fix(capacitor): expose affected rows in default query result (#12358)

### LINKED ISSUE #12358: Issue template for feature request is missing Capacitor driver option.
https://github.com/typeorm/typeorm/blob/master/.github/ISSUE_TEMPLATE/feature-request.yml

Needs Capacitor:
Image

Srikumaran-Sivarajan:
Fixes #12358

## Description
The Capacitor driver computed affected rows internally but did not expose them
in default query results unless structured result mode was used.

## Solution
- Return `QueryResult` for write queries (INSERT, UPDATE, DELETE)
- Ensure `affected` rows are accessible in default query usage
- Added test to reproduce and verify the issue

## Notes
- Mock connection is used to simulate Capacitor environment


Cprakhar:
@Srikumaran-Sivarajan, thanks for your effort. There is nothing new in this PR.

See https://github.com/typeorm/typeorm/issues/12357#issuecomment-4275615021

PR #12388, opened 2026-04-19
fix(postgres): preserve column data by using ALTER COLUMN instead of drop and re-add

budingsoft:
### Description of change

Fixes #3357.

**Current behavior.** When the type, length, precision, or scale of a PostgreSQL column changes, `PostgresQueryRunner.changeColumn` drops and re-adds the column. Every value in that column is silently destroyed — which is especially destructive for migrations generated from entity edits (e.g. bumping a `varchar(50)` to `varchar(100)` wipes the whole column in production).

**New behavior.** Compatible type and length changes are applied in place with `ALTER COLUMN TYPE`, preserving existing row data. Only changes that genuinely require recreation still fall through to drop + add:

- array toggles (`isArray` flip between scalar and array storage)
- generated-STORED transitions
- cross-family type changes that Postgres can't cast without a `USING` clause (`varchar` ↔ `int`, enum ↔ non-enum, etc.)

A new `canAlterColumnType(oldColumn, newColumn)` helper encapsulates the safety check. Type aliases are normalized (`character varying` → `varchar`, `integer` → `int4`, `timestamp without time zone` → `timestamp`, and so on) so TypeORM's declarative types align with the canonical names Postgres reports back. Four groups are treated as in-place compatible:

- same canonical type (length/precision/scale may differ)
- string family: `varchar`, `char`, `text`
- integer family: `int2`, `int4`, `int8`
- numeric/float family: `numeric`, `float4`, `float8`

### Why this is safe (and safer than the status quo)

The 7-year-old concern on #3357 was that `ALTER COLUMN TYPE` might silently lose data on narrowing changes (e.g. `varchar(50)` → `varchar(10)`). This does not happen: PostgreSQL raises `value too long for type character varying(10)` and aborts the migration when a cast would truncate rows, so the engine itself refuses to destroy data. The previous behavior — drop + add — always destroys every value, with no warning.

So the outcomes compare as:

| Change | Old behavior | New behavior |
| --- | --- | --- |
| `varchar(50)` → `varchar(100)` | data lost | data preserved |
| `int` → `bigint` | data lost | data preserved |
| `varchar` → `text` | data lost | data preserved |
| numeric precision change | (already handled by existing code) | unchanged |
| `varchar(50)` → `varchar(10)` with oversize data | data lost silently | migration fails loudly, data intact |
| `int` ↔ `varchar` | data lost | still drop + add (unchanged) |
| scalar ↔ array | data lost | still drop + add (unchanged) |

The cloned table used by downstream branches (nullable, default, comment, primary, unique, generated) is updated with the new type/length/precision/scale so those checks see the post-ALTER shape. The existing numeric-precision ALTER block is folded into a single unified ALTER block so precision/scale changes and length/type changes no longer emit duplicate `ALTER COLUMN TYPE` statements.

### Verification

- **New functional tests** (`test/functional/schema-builder/column-type/alter-column-preserve-data/`) cover the four data-preserving paths and assert both the post-synchronize column shape and that row data is still there:
  - widening `varchar` length (50 → 100)
  - switching `varchar` → `text`
  - widening `int` → `bigint`
  - changing `numeric` precision/scale
- **Regression** — full `schema builder` suite on Postgres 17 + better-sqlite3: **158 passing / 0 failing**, including all 10 existing `schema builder > change column` cases.
- Scope is strictly `PostgresQueryRunner`; no other drivers are touched.

### Pull-Request Checklist

- [x] There are new or updated tests validating the change (`test/functional/schema-builder/column-type/alter-column-preserve-data/*.test.ts`)
- [ ] Documentation has been updated to reflect this change (`docs/docs/**.md`) — N/A (no public API surface change; behavior change only, better default)

### DIFF
--- a/src/driver/postgres/PostgresQueryRunner.ts
+++ b/src/driver/postgres/PostgresQueryRunner.ts
@@ -1294,6 +1294,88 @@ export class PostgresQueryRunner
     }
 
     /**
+     * Checks whether a column type/length change can be applied in place with
+     * ALTER COLUMN TYPE instead of drop + add, preserving existing data.
+     * Only conversions Postgres accepts without a USING clause are treated as
+     * safe; incompatible changes fall back to the drop + add path.
+     *
+     * @param oldColumn
+     * @param newColumn
+     */
+    protected canAlterColumnType(
+        oldColumn: TableColumn,
+        newColumn: TableColumn,
+    ): boolean {
+        if (
+            oldColumn.type === "enum" ||
+            oldColumn.type === "simple-enum" ||
+            newColumn.type === "enum" ||
+            newColumn.type === "simple-enum"
+        )
+            return false
+
+        const normalize = (type: string): string => {
+            switch (type.toLowerCase().trim()) {
+                case "character varying":
+                case "varchar":
+                    return "varchar"
+                case "character":
+                case "char":
+                    return "char"
+                case "integer":
+                case "int":
+                case "int4":
+                    return "int4"
+                case "bigint":
+                case "int8":
+                    return "int8"
+                case "smallint":
+                case "int2":
+                    return "int2"
+                case "real":
+                case "float4":
+                    return "float4"
+                case "double precision":
+                case "float8":
+                    return "float8"
+                case "timestamp":
+                case "timestamp without time zone":
+                    return "timestamp"
+                case "timestamptz":
+                case "timestamp with time zone":
+                    return "timestamptz"
+                case "time":
+                case "time without time zone":
+                    return "time"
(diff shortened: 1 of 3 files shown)


xcskier56:
Just to add, we also very nearly got hit by this foot-gun today. We noticed the data loss in our staging ENV before production. I hate to be such a hater but having a "change" method that will drop the column is really really dumb. 

`.changeColumn` should never issue a drop column. It should error immediately if the requested changes cannot be done without data loss.

tl;dr big +1 for this change

smith-xyz:

PR #12391, opened 2026-04-20
fix(codemod): harden scope, idempotency, and import-strip semantics

pkuczynski:
Bunch of fixes and cleanup based on recent reviews and user feedback. High-level list of changes:

**Scope guards.** Several v1 transforms were a bit too eager and would rewrite object properties or rename methods in unrelated code that just happened to live in a file importing from TypeORM. The MongoDB / MySQL-connector / SAP / column-unsigned / data-source-name / where-expression transforms now check that the surrounding object is the right kind of DataSource config (or that the binding actually resolves to a TypeORM import) before touching anything.

**Repository-typed method renames.** Transforms that rename `.exist()`, `.findByIds()`, `.findOneById()`, `.stats()` now look at whether the receiver is actually a TypeORM Repository/EntityManager — via a type annotation or as the result of `.getRepository()`. Files without typed bindings still get the legacy permissive behavior so nothing regresses.

**Import strip symmetry.** A few transforms stripped TypeORM imports even when there was no automatic rewrite for the call site, leaving users with a broken file. They now leave a TODO at each call site (and don't drop the import when nothing was flagged).

**Idempotency.** Running the codemod twice on the same file should be a no-op. Every TODO emission is now guarded, and the test harness re-runs every fixture a second time to enforce this. While adding the check, a latent duplicate-import bug in the connection→datasource transform surfaced and is also fixed.

**Correctness.**
- A few transforms used to attach TODO comments to nodes where recast silently dropped them — they now target statement-level hosts.
- Stats aggregation no longer mishandles malformed keys or Windows-style file paths.
- The CLI now surfaces unclassified worker output (stack traces, warnings) instead of silently hiding it.

**New coverage.** Re-run idempotency loop across every fixture, plus negative fixtures pinning the scope-guard behavior.

**CLI.** Added `--version` / `-v`.

**Broader connection-to-datasource coverage.** Based on feedback on real-world NestJS code: `this.X.connection` now rewrites when `X` is a typed class property, getter, or constructor parameter property. Destructuring in subscriber handlers (`const { connection } = event`) is also handled. And metadata constructor options (`new EntityMetadata({ connection })`) get the key rewrite or removal too.

**Find options without direct typeorm imports.** Service files that only pull TypeORM types through a wrapper (common in multi-tenancy setups) are no longer skipped. The `select: [...]` / `relations: [...]` rewrites now trigger on the `.find*()` call shape instead of a file-level typeorm import.

**Dynamic `select` / `relations` values.** Expressions like `select: exclude(...)` are now auto-wrapped with `Object.fromEntries(...)` so the codemod output is runnable. For `relations`, a TODO notes the dot-path caveat.

**Driver-specific DataSourceOptions re-exports.** Every `*DataSourceOptions` type is now re-exported from the `typeorm` index. Deep paths still work but are discouraged; the codemod simplifies existing deep imports to `from "typeorm"`.


michaelbromley (approved):

gioboa (approved):

packages/codemod/src/transforms/v1/datasource-name.ts
```diff
@@ -14,34 +18,48 @@ export const datasourceName = (file: FileInfo, api: API) => {
...
-    }).forEach((path) => {
+    root.find(j.CallExpression).forEach((path) => {
+        const callee = path.node.callee
+        if (
+            callee.type !== "Identifier" ||
+            !createConnectionNames.has(callee.name)
+        )
+            return
         removeNameFromObject(path.node.arguments[0])
     })
```
Cprakhar, line 64: For input:
```ts
import { createConnection } from "typeorm"

const conn = await createConnection({
    name: "default", // this is not removed in the output
    type: "postgres",
    database: "test",
})
```
Output:
```ts
// TODO(typeorm-v1): `createConnection()` was removed — instantiate a `DataSource` and call `.initialize()` instead: `const dataSource = new DataSource(options); await dataSource.initialize()`
const conn = await createConnection({
    name: "default", // this is not removed in the output
    type: "postgres",
    database: "test",
})
```

`createConnection` then be handled by the `global-functions.ts` for removal of import and adding TODO at the call.

naorpeled (approved):

packages/codemod/src/transforms/v1/datasource-name.ts
Cprakhar: Works fine in `test/transforms/v1/fixtures/datasource-name`.

I locally tried it in a sample project - not working.

Cprakhar: Resolved in: d62ee0c

src/index.ts
```diff
@@ -137,6 +137,27 @@ export type * from "./subscriber/event/QueryEvent"
...
+export type { MongoDataSourceOptions } from "./driver/mongodb/MongoDataSourceOptions"
+export type { MysqlDataSourceOptions } from "./driver/mysql/MysqlDataSourceOptions"
+export type { NativescriptDataSourceOptions } from "./driver/nativescript/NativescriptDataSourceOptions"
+export type { OracleDataSourceOptions } from "./driver/oracle/OracleDataSourceOptions"
+export type { PostgresDataSourceOptions } from "./driver/postgres/PostgresDataSourceOptions"
+export type { ReactNativeDataSourceOptions } from "./driver/react-native/ReactNativeDataSourceOptions"
+export type { SapDataSourceOptions } from "./driver/sap/SapDataSourceOptions"
+export type { SpannerDataSourceOptions } from "./driver/spanner/SpannerDataSourceOptions"
+export type { SqljsDataSourceOptions } from "./driver/sqljs/SqljsDataSourceOptions"
+export type { SqlServerDataSourceOptions } from "./driver/sqlserver/SqlServerDataSourceOptions"
```
alumni: These were on purpose omitted from the exports in the past - and not just these, but also the `XxxQueryRunner`s and the `XxxDriver`s - I would revert the changes in this file.

src/driver/expo/ExpoDriver.ts
```diff
@@ -25,7 +25,7 @@ const isExpoSqliteNotFoundError = (err: unknown): boolean => {
     if (e.code !== "MODULE_NOT_FOUND") return false
     if (typeof e.message !== "string") return false
     const firstLine = e.message.split("\n", 1)[0]
-    return /^Cannot find module '(?:expo-sqlite)'/.test(firstLine)
+    return firstLine.startsWith("Cannot find module 'expo-sqlite'")
```
alumni, line 28: Good, let's avoid RegEx, especially for things what are easy to implement ➕

packages/codemod/src/transforms/v1/connection-to-datasource.ts
```diff
@@ -199,6 +311,26 @@ export const connectionToDataSource = (file: FileInfo, api: API) => {
...
+            path.node.specifiers!.every(
+                (spec) =>
+                    spec.type === "ImportSpecifier" &&
+                    spec.imported.type === "Identifier" &&
+                    reExportedDriverOptions.has(spec.imported.name),
+            )
+        ) {
+            path.node.source.value = "typeorm"
+            hasChanges = true
+        }
```
alumni, line 443: This should also be reverted together with the changes in `index.ts`

alumni (approved):

alumni (approved):

Issue #9960, opened 2023-04-18
0.3.14 breaks addOrderBy


paoesco:
### Issue description

0.3.14 introduced a change where the alias is not mapped correctly to SQL equivalent 

### Expected Behavior

The following code:
```
query = query.addOrderBy("verification.createdAt", "DESC", "NULLS LAST");
```
should produce the following SQL:
```
ORDER BY "verification"."created_at" DESC NULLS LAST LIMIT 10
```

### Actual Behavior

The following code:
```
query = query.addOrderBy("verification.createdAt", "DESC", "NULLS LAST");
```
is translated to SQL:
```
ORDER BY "verification_created_at" [object Object] LIMIT 10
```

### Steps to reproduce

Create a query with `createQueryBuilder` and add `addOrderBy`.

### My Environment

| Dependency          | Version  |
| ---                 | ---      |
| Operating System    |  Mac OS     |
| Node.js version     | 18.15.0  |
| Typescript version  |  4.8.4  |
| TypeORM version     | 0.3.14 |

### Additional Context

I believe the issue has been introduced by the following PR: https://github.com/typeorm/typeorm/pull/9914/files

### Relevant Database Driver(s)

- [ ] aurora-mysql
- [ ] aurora-postgres
- [ ] better-sqlite3
- [ ] cockroachdb
- [ ] cordova
- [ ] expo
- [ ] mongodb
- [ ] mysql
- [ ] nativescript
- [ ] oracle
- [X] postgres
- [ ] react-native
- [ ] sap
- [ ] spanner
- [ ] sqlite
- [ ] sqlite-abstract
- [ ] sqljs
- [ ] sqlserver

### Are you willing to resolve this issue by submitting a Pull Request?

Yes, I have the time, but I don't know how to start. I would need guidance.

pleerock:
Added test in #9961 

It does work for me on my local machine, let's see if CI tests pass.

Meanwhile, please provide a minimal reproduction on the issue.

pleerock:
@paoesco please review the PR and let me know what code changes I need to add to make your issues reproducible. As you can see tests are passing

pleerock:
Confirmed. Will be fixed by #9961 and merged very soon.

paoesco:
Quite impressed by how fast it got resolved.
Thanks team!

Issue #9420, opened 2022-09-30
Order by column from subquery


stuartf123:
## Issue Description

 

I would like to order the resultset by a column from a subquery which I have joined to the outer query. Unfortunately the typeorm library throws an error when using the query builder:

 

Exception TypeORMError: Cannot get entity metadata for the given alias "child"

 

### Expected Behavior

 

To be able to execute a query to the DB using a query builder and order the results by a column from a sub query

 

### Actual Behavior

 

Exception TypeORMError: Cannot get entity metadata for the given alias "child"

 

### Steps to Reproduce

 

```

class Parent {

    id: number

}

 

class Child {

    parentId: number

    initials: string

}

 

createQueryBuilder(Parent, 'parent').innerJoinAndSelect(subQuery=> {

    return subquery.from(Child, 'child').select('parentId').addSelect('max(initials)', 'initials').groupBy(parentId)

}, 'child', 'parent.id = child.parentId').orderBy('child.initials')

 

```

 

The above code doesn't get as far as generating the SQL to the DB. The typeorm library fails for the 'orderBy' with the following error:

 

Exception TypeORMError: Cannot get entity metadata for the given alias "child"

 

f I remove the orderBy then the generated sql is similar to the following (typeorm should expand child.* to the appropriate columns - but it doesn't):

 

```

select ....., child.*

from parent parent

join (select parentId, max(initials) as initials from child) child

    on child.parentId = parent.id

```

 

### My Environment

 

| Dependency          | Version  |

| ---                 | ---      |

| Operating System    |          |

| Node.js version     | v18.8.0  |

| Typescript version  | 4.8.3  |

| TypeORM version     | 0.3.6  |

 

 

### Relevant Database Driver(s)

 

 

| DB Type           | Reproducible  |

|-------------------| ---           |

| `aurora-mysql`    | no            |

| `aurora-postgres` | yes            |

| `better-sqlite3`  | no            |

| `cockroachdb`     | no            |

| `cordova`         | no            |

| `expo`            | no            |

| `mongodb`         | no            |

| `mysql`           | no            |

| `nativescript`    | no            |

| `oracle`          | no            |

| `postgres`        | no            |

| `react-native`    | no            |

| `sap`             | no            |

| `spanner`         | no            |

| `sqlite`          | no            |

| `sqlite-abstract` | no            |

| `sqljs`           | no            |

| `sqlserver`       | no            |

 

 

### Are you willing to resolve this issue by submitting a Pull Request?

 

[...]

Issue #8346, opened 2021-11-09
MySQL: Regression when using `take`, `orderBy`, and `getMany` on a joined relation


stephenwade:
## Issue Description

**!!!!** This is a regression in the latest patch update. This worked correctly in 0.2.38 and stopped working in 0.2.39.

When I use `.take()`, `.orderBy()`, and `.getMany()`, if the `.orderBy()` uses a joined relation, I get a database error about a duplicate column.

If I remove `.take()`, order by a column on the main relation instead of the joined relation (customers instead of contacts in my example), or use `.getRawMany()` instead of `.getMany()`, the problem does not occur.

### Expected Behavior

These are the queries in the MySQL logs from version 0.2.38:

```sql
SELECT
	DISTINCT `distinctAlias`.`customer_id` as "ids_customer_id",
	`distinctAlias`.`contacts_firstName`
FROM
	(
		SELECT
			`customer`.`id` AS `customer_id`,
			`customer`.`name` AS `customer_name`,
			`contacts`.`id` AS `contacts_id`,
			`contacts`.`firstName` AS `contacts_firstName`,
			`contacts`.`lastName` AS `contacts_lastName`,
			`contacts`.`customerId` AS `contacts_customerId`
		FROM
			`Customer` `customer`
			LEFT JOIN `CustomerContact` `contacts` ON `contacts`.`customerId` = `customer`.`id`
	) `distinctAlias`
ORDER BY
	`distinctAlias`.`contacts_firstName` DESC,
	`customer_id` ASC
LIMIT
	10;

SELECT
	`customer`.`id` AS `customer_id`,
	`customer`.`name` AS `customer_name`,
	`contacts`.`id` AS `contacts_id`,
	`contacts`.`firstName` AS `contacts_firstName`,
	`contacts`.`lastName` AS `contacts_lastName`,
	`contacts`.`customerId` AS `contacts_customerId`
FROM
	`Customer` `customer`
	LEFT JOIN `CustomerContact` `contacts` ON `contacts`.`customerId` = `customer`.`id`
WHERE
	`customer`.`id` IN (34, 185, 22, 65, 35, 166, 43, 51, 110, 183)
ORDER BY
	`contacts`.`firstName` DESC
```

### Actual Behavior

This is the query in the MySQL logs from version 0.2.39:

```sql
SELECT
	DISTINCT `distinctAlias`.`customer_id` as "ids_customer_id",
	`distinctAlias`.`contacts_firstName`
FROM
	(
[...]

PR #12392, opened 2026-04-21
test(query-builder): consolidate functional tests for join, order-by and take

pkuczynski:
Consolidates regression tests from `test/github-issues/` and `test/other-issues/` into the functional test tree, and splits the top-level `query-builder-joins.test.ts` and `query-builder-order-by.test.ts` files into behavior-focused subfolders.

**Moved into `query-builder/join/`:**
- #215 → `on-clause-property-name-regex/`
- #401 → `escape-keyword-aliases/`
- #2588 → `custom-on-condition/`
- `other-issues/join-empty-relations/` → `join-empty-relations/`

**Moved into `query-builder/order-by/`:**
- #813 → `with-functions/`
- #1259 → `with-added-select/`
- #9420 → `from-subquery/`
- #9960 → `nulls-last/`

**Moved into `query-builder/take/`:**
- #388 → `string-id/`
- #479 → `with-or-where/`
- #1099 → `mysql-skip-sql/`
- #5174 → `with-ids-parameter/`
- #8346 → `with-joined-order-by/`
- `other-issues/take-multiple-pk/` → `multiple-pk/`
- `other-issues/limit-with-order-by/` → `with-order-by/`

**Split:**
- `query-builder-joins.test.ts` → `join/basic/` + `join/composite-pk/` + `take/with-joins/`
- `order-by/query-builder-order-by.test.ts` → `order-by/basic/`
- `order-by/sql-injections.test.ts` → `order-by/sql-injection/`

**Cleanup:**
- Removed `take/string-primary-key/` (duplicate of `take/string-id/`, #493 / #388)
- Removed `order-by/joined-relation/` (#1591, skipped since 2018, had no assertions)
- Dropped `enabledDrivers: ["postgres"]` from `order-by/from-subquery/` (subquery ORDER BY is standard SQL)
- Split `order-by/with-added-select/` — arithmetic test runs on all drivers; the full-text `ts_rank_cd` test stays Postgres-only
- Broadened `order-by/nulls-last/` from Postgres-only to all drivers except MySQL/MariaDB/MSSQL (which don't support the keyword)
- Renamed weak / duplicate `it()` titles across the three folders


test/functional/query-builder/order-by/from-subquery/from-subquery.test.ts
```diff
@@ -1,17 +1,23 @@
...
     before(async () => {
         dataSources = await createTestingConnections({
-            enabledDrivers: ["postgres"],
+            // oracle/sap fold unquoted identifiers to uppercase, but the
+            // outer `sub.name` reference is emitted unquoted while the
+            // subquery's aliases are quoted (lowercase) — so the lookup
+            // mismatches on those drivers. A TypeORM-side fix in
+            // createOrderByExpression (escape the subquery fall-through)
+            // would let this list shrink; tracked separately.
+            disabledDrivers: ["oracle", "sap"],
```
smith-xyz, line 20: should we preserve the original test on this - reading through #9420 it seems this test was very intentional on including an inner `.orderBy("user.name", "ASC")` on the subquery -- that was part of the original reproduction case.

could maybe do two tests like:

```ts
it("should resolve an outer orderBy on a subquery column (all drivers)", ...)
// current test without inner orderBy
it("should resolve orderBy when the subquery itself is ordered (postgres)", ...)
// original test with inner .orderBy("user.name", "ASC"), enabledDrivers: ["postgres"]
```

just so we don't lose the coverage. Unless I'm missing some context?

smith-xyz (changes requested):

gioboa (approved):

PR #12394, opened 2026-04-21
fix(codemod): harden scope and type-name detection across more AST shapes

pkuczynski:
Follow-ups for Qodo review items flagged after #12391 merged.

## Changes

- **`ensureDataSourceOptionsTypeImport`**: skip typeorm type-only imports that carry a default or namespace specifier — only a purely named type-only import is safe to augment with `DataSourceOptions`.
- **`fileImportsFrom`**: recognise `export { X } from "typeorm"` and `export * from "typeorm/..."` as references so barrel files with no imports still pass the scope guard. Adds a `connection-to-datasource-barrel-only` fixture.
- **`getTypeReferenceRootName`**: resolve `typeorm.Repository<User>` (`TSQualifiedName` inside `TSTypeReference` / `TSTypeQuery`) to the rightmost identifier so `import * as typeorm` callers classify the same as bare `Repository<T>`.
- **`runOneTransform`**: drop the swallowing try/catch around `printUnclassifiedOutput` — a failure there is signal, not noise, and should surface rather than be silenced.

### DIFF
--- a/packages/codemod/test/transforms/v1/fixtures/connection-to-datasource/connection-to-datasource-barrel-only.output.ts
+++ b/packages/codemod/test/transforms/v1/fixtures/connection-to-datasource/connection-to-datasource-barrel-only.output.ts
@@ -1,0 +1,6 @@
+// Barrel files that only re-export TypeORM APIs (no `import` statements
+// at all) still need the Connection → DataSource rewrite. Without the
+// re-export check in `fileImportsFrom`, this file would be skipped.
+export { DataSource, DataSourceOptions } from "typeorm"
+export { DataSource as LegacyConn } from "typeorm"
+export * from "typeorm/driver/sap/SapConnectionOptions"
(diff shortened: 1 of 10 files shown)


naorpeled (approved):

packages/codemod/test/transforms/v1/fixtures/connection-to-datasource/connection-to-datasource-barrel-only.output.ts
```diff
@@ -0,0 +1,6 @@
+// Barrel files that only re-export TypeORM APIs (no `import` statements
+// at all) still need the Connection → DataSource rewrite. Without the
+// re-export check in `fileImportsFrom`, this file would be skipped.
+export { DataSource, DataSourceOptions } from "typeorm"
+export { DataSource as LegacyConn } from "typeorm"
+export * from "typeorm/driver/sap/SapConnectionOptions"
```
alumni: this? :P

alumni (approved):

packages/codemod/test/transforms/v1/fixtures/connection-to-datasource/connection-to-datasource-barrel-only.output.ts
pkuczynski: This is only an example and existance of the file makes no difference for this specific transfiormation, but sure make sense to keep this up to date :)

gioboa (approved):

PR #12398, opened 2026-04-22
fix(codemod): harden destructure and DI accessor rewrites for connection to dataSource rename

pkuczynski:
Four related improvements to the v1 `connection → dataSource` transform, all addressing feedback that the current output misses common TypeScript patterns.

### 1. Scope-aware destructure binding rename

Before: `const { connection } = event` was rewritten to `const { dataSource: connection } = event` — functional but ugly, keeping the old local name intact.

After: the shorthand `{ connection }` becomes `{ dataSource }`, and every reference to the `connection` local within the enclosing scope is renamed to `dataSource`. Downstream code reads naturally as if written against v1 from the start.

Safety rails:

- Conflict detection via `scope.lookup("dataSource")` — if any visible `dataSource` binding already exists in scope (parameter, module-level const, import, outer let/const), falls back to the previous re-alias form.
- Shadowed inner scopes honored — an inner lambda re-declaring `connection` keeps its references untouched.
- Non-reference identifiers (property keys, member-access names, import specifiers, class-member keys) filtered out.
- For-of / for-in loop bodies handled: the declarator walk runs from root so ast-types scope info is preserved.

### 2. DI-rooted accessor tracking

Before: `this.entityManager.connection` and `queryRunner.connection` (where `queryRunner` came from `this.dataSource.createQueryRunner()`) were left untouched — the accessor-chain tracker only recognized bare identifier roots like `ds.manager`, not `this.ds.manager`.

After: receiver resolution accepts `this.<member>` as a DataSource root whenever `<member>` is already tracked as a DataSource class field (populated by the existing TSParameterProperty / ClassProperty walks). Covers the NestJS-style DI pattern `constructor(private dataSource: DataSource)` end-to-end:

- `private manager = this.dataSource.manager` — class field, no explicit type → tracked as EntityManager.
- `const qr = this.dataSource.createQueryRunner()` — local var → tracked as QueryRunner.
- `const mgr = this.dataSource.manager` — local var → tracked as EntityManager.

All three then flow into the `.connection → .dataSource` member-rename pass.

### 3. Destructured event properties carry their property type

Before: `const { queryRunner } = event; queryRunner.connection.options` — the destructured `queryRunner` local wasn't classified, so the downstream `.connection` rewrite didn't fire.

After: when destructuring from a tracked receiver (subscriber event, EntityManager, QueryRunner, Repository, …), each destructured property is classified by name: `manager` / `mongoManager` → EntityManager, `queryRunner` → QueryRunner, `metadata` → EntityMetadata. The local binding is added to the appropriate tracking set so subsequent `.connection` / `.entityMetadata.dataSource` rewrites apply.

### 4. Deep-path typeorm imports trigger the metadata-type rewrite

Before: `new ColumnMetadata({ connection: tenantConnectionService.connection, ... })` left the `connection` key intact whenever `ColumnMetadata` was imported from a deep path like `"typeorm/metadata/ColumnMetadata"`. The shared `getLocalNamesForImport` / `getNamespaceLocalNames` helpers did exact module-name matches against `"typeorm"`, so deep-path local bindings never landed in the tracked-names sets.

After: both helpers match sub-paths as well — `typeorm` and `typeorm/…` both count as a reference to the TypeORM module. All call sites pass `"typeorm"` today, so no behavior change for the common shallow-import case.

### Fixtures

- Updated `connection-to-datasource-subscriber-destructure` to the new idiomatic scope-aware output.
- Added `connection-to-datasource-destructure-conflict` — parameter conflict + inner shadow.
- Added `connection-to-datasource-destructure-negative` — module-scoped conflict + non-shorthand property key in same scope + untracked member access.
- Added `connection-to-datasource-di-accessor` — DI-rooted class field inference + local-var accessor chain + explicit-type sanity check.
- Added `connection-to-datasource-chained-access` — multi-step `.connection` chains on tracked receivers and destructured event properties.
- Added `connection-to-datasource-metadata-complex-value` — drop fires when the `connection` option holds a non-shorthand expression like `tenantConnectionService.connection`.
- Added `connection-to-datasource-metadata-deep-import` — `import { ColumnMetadata } from "typeorm/metadata/ColumnMetadata"` and friends now trigger the constructor-option rewrite.


alumni (approved):

gioboa (approved):

michaelbromley (commented):
Really like the unit test format with the inputs and outputs. Makes it easy to review the actual effect of the change

PR #12400, opened 2026-04-22
fix(codemod): cover ColumnMetadata args.options in column option rewrites

pkuczynski:
The `column-readonly` and `column-width-zerofill` transforms rewrite `@Column(...)` decorator options — the former renames `readonly` → `update` (negated), the latter strips `width` and `zerofill`. Both were scoped to decorator arguments only, leaving the same fields unchanged when they appeared inside `new ColumnMetadata({ args: { options: { ... } } })` constructor calls. `ColumnMetadataArgs.options` is typed `ColumnOptions`, so the v1 field removals apply there too. Rare in user code but shows up in multi-tenant and runtime-metadata construction patterns.

Fix: new shared helper `forEachColumnMetadataOptionsArg` walks `new ColumnMetadata(...)` constructor calls and yields the inner `args.options` ObjectExpression. Both transforms now run their existing per-object callback on this path in addition to the decorator walk. Gated on typeorm-imported local names so user-defined classes of the same name are untouched.

Also included: positional-form decorator fixtures. `@Column('integer', { width: 10, zerofill: true })` and `@Column('varchar', { readonly: true })` already worked (via `forEachDecoratorObjectArg` iterating all decorator args), but no fixture locked the behavior in.

### Fixtures

- `column-readonly-metadata-args` — `new ColumnMetadata({ args: { options: { readonly: false, ... } } })` → `update: true`. Includes control: a user class named `ColumnMetadata2` must not be touched.
- `column-width-zerofill-metadata-args` — `new ColumnMetadata({ args: { options: { width, zerofill, ... } } })` → both stripped. Same control.
- `column-readonly-positional` — `@Column("varchar", { readonly: true })` → `{ update: false }`.
- `column-width-zerofill-positional` — `@Column("integer", { width: 10, zerofill: true })` → `@Column("integer", {})`.


packages/codemod/test/transforms/v1/fixtures/column-readonly/column-readonly-metadata-args.output.ts
```diff
@@ -0,0 +1,34 @@
...
+
+declare const metadata: any
+declare const tenantConnectionService: { connection: any }
+
+// `new ColumnMetadata({ args: { options: {...} } })` — the nested `options`
+// object is typed `ColumnOptions`, so the same `readonly` → `update` rename
+// that fires on `@Column({...})` must also fire here. Rare in user code but
+// shows up in multi-tenant / runtime metadata construction.
+const col = new ColumnMetadata({
+    connection: tenantConnectionService.connection,
```
alumni: should not be here

gioboa (approved):

packages/codemod/test/transforms/v1/fixtures/column-width-zerofill/column-width-zerofill-positional.output.ts
```diff
@@ -0,0 +1,15 @@
...
+// Two-arg decorator form: type string as first positional arg, options object
+// as second arg — e.g. `@Column('integer', { width: 10, zerofill: true })`.
+// Both `width` and `zerofill` must be removed from the options object; the
+// positional type string is left untouched.
+class Post {
+    @Column("integer", {})
+    postCode: number
+
+    @Column("int", {
+        nullable: true,
```
alumni, line 12: I wonder if we can keep a single line, like in the input, since `prettier` will not go back to single line. I noticed it in other places as well (like in select/relations).

alumni (approved):

PR #12401, opened 2026-04-22
chore(deps): bump dependencies and fix lint offenses

pkuczynski:
### Before / after

`pnpm audit` on master: 41 vulnerabilities (2 critical, 20 high, 15 moderate, 4 low). After this PR: 3 remaining (1 high, 1 moderate, 1 low) in the main repo; `packages/codemod` drops from 6 to 3 on the same remaining set.

### Remaining

All three remaining advisories are transitive under `mocha` (`serialize-javascript` ×2, `diff` ×1) and can't be fixed without an upstream mocha release — mocha is already at 11.7.5 (latest).

### Also included

- Ran `eslint --fix` for the tightened `prefer-optional-chain` and `no-unnecessary-type-assertion` rules across `src/persistence/` and `packages/codemod/`; also simplified a redundant `=== false ? false : true` ternary in `SubjectExecutor.ts`.
- `engines.node` bumped to match eslint's resolved engine range.
- Allowlist `allowedDriverRequire` in mocha's `check-leaks`. mongodb 7.2+ unconditionally sets this on `globalThis` via `resolveRuntimeAdapters` without cleanup; comment in `.mocharc.json` documents the reason (and CI's jq step strips `//` comments before parsing).


alumni:
Could you also bump `c8`, `pr-pkg-new` and `pnpm` (in `packageManager`)? thanks :)

gioboa (approved):

pkuczynski:
> Could you also bump `c8`, `pr-pkg-new` and `pnpm` (in `packageManager`)? thanks :)

Done!

.mocharc.json
```diff
@@ -4,6 +4,8 @@
   "color": true,
   "exit": true,
   "file": ["./build/compiled/test/utils/test-setup.js"],
+  // mongodb@7.2+ leaks `allowedDriverRequire` on globalThis; allowlist it.
```
smith-xyz, line 7: is this commented line allowed?

src/persistence/EntityPersistExecutor.ts
```diff
@@ -54,7 +54,7 @@ export class EntityPersistExecutor {
         // save data in the query runner - this is useful functionality to share data from outside of the world
         // with third classes - like subscribers and listener methods
         const oldQueryRunnerData = queryRunner.data
-        if (this.options && this.options.data) {
+        if (this.options?.data) {
```
smith-xyz, line 57: is this change related to the bumps?

smith-xyz:
are we able to align the pnpm versions? looks like qodo pointed it out

src/persistence/EntityPersistExecutor.ts
alumni: yeah, it's lint fix - the "prefer optional chaining" rule was improved in the new version of typescript-eslint

pkuczynski: Yup, exactly as @alumni said. I bumped lint and had to run `--fix` to make it pass

.mocharc.json
pkuczynski: Mocha tolerates it (`strip-json-comments`), but the workflow's jq step does not :(

pkuczynski: Pushed a fix

smith-xyz (approved):
lgtm!

PR #12403, opened 2026-04-22
chore: remove legacy samples and move playground to root

pkuczynski:
The `sample/` directory contains 30 legacy demos from the pre-1.0 era that aren't referenced from tests or CI. Most of their content is covered by `docs/`. Removing them drops ~4750 lines and lets `tsconfig.json` stop including the folder in compilation.

`sample/playground` is kept — it's the template used by `pkg-pr-new` in the PR preview workflow — but moves up to `/playground` so it's not collocated with the removed samples.

### Changes

- Move `sample/playground/` → `playground/`
- Delete `sample/`
- `.github/workflows/preview.yml` — update `pkg-pr-new` template path to `./playground`
- `eslint.config.mjs` — update ignore pattern
- `tsconfig.json` — drop `sample` include; update playground exclude path
- `README.md`, `docs/docs/getting-started.md` — drop line pointing at the removed `sample/` folder


Cprakhar (approved):

gioboa (approved):

michaelbromley (commented):
Can you do a pass to see what info is _only_ covered in the samples and not in the docs? Just in case there is one or two things not actually mentioned that we would be losing all reference to.

pkuczynski:
> Can you do a pass to see what info is _only_ covered in the samples and not in the docs? Just in case there is one or two things not actually mentioned that we would be losing all reference to.

I scan them before deleting and could not find anything valuable or not covered in docs already.

michaelbromley (approved):

PR #12406, opened 2026-04-23
feat(schema-builder): rename constraints instead of re-creating them on name drift

pkuczynski:
When entity metadata and the live database agree on a constraint's **structure** (columns, uniqueness, referenced target, expression, …) but disagree only on its **name**, the schema-builder now emits a `RENAME` (or dialect equivalent) instead of `DROP + CREATE`.

On a small table this saves a few statements. On a hundred-million-row table it's the difference between a hot-path-blocking exclusive lock and a metadata-only rename — roughly O(rows) vs O(1). For a production app whose migration windows are measured in seconds, that's a real availability win, which is why this ships always-on with no opt-out flag.

## When it kicks in

1. **Naming-strategy swap** — moving from `DefaultNamingStrategy` to `SnakeNamingStrategy` or a custom one, where every auto-named constraint produces a different string.
2. **Adding a custom `name:`** — sticking `@Index({ name: "ix_user_email" })` on what was previously an auto-named index.
3. **Removing a custom `name:`** — deleting the explicit name so the hashed default takes over.
4. **Hash algorithm refactor** — #11864 replaced `encodeURIComponent`+charCode with `TextEncoder`. Affects users whose table or column names contain non-ASCII characters; every auto-named constraint on those tables got a new hash.

Before this change, all four scenarios produce `DROP OLD ; CREATE NEW` on every constraint. After, they produce `RENAME`.

## How it works

- A new pair of utilities in `src/schema-builder/util/`:
  - `constraintSignature.ts` — deterministic, name-free structural fingerprints per family (index, unique, primary key, foreign key, check, exclusion).
  - `constraintReconciler.ts` — pure pairing function. Takes two sets of constraints (DB-side, metadata-side), returns the rename pairs, the unmatched DB-side (drops), and the unmatched metadata-side (creates).
- `RdbmsSchemaBuilder.renameMismatchedConstraints()` — runs before the existing drop/create passes, so subsequent passes see the renamed constraints and no-op.
- Six `renameX` primitives on each relational `QueryRunner` (optional methods, dialect-specific DDL):
  - Postgres / CockroachDB: `ALTER TABLE … RENAME CONSTRAINT` / `ALTER INDEX … RENAME TO`
  - MSSQL: `EXEC sp_rename`
  - Oracle: `ALTER TABLE … RENAME CONSTRAINT` / `ALTER INDEX … RENAME TO`
  - MySQL / MariaDB: `ALTER TABLE … RENAME INDEX`, drop+add wrapper for the other families
  - SAP HANA: standalone `RENAME INDEX`, drop+add for other families
- SQLite and Spanner intentionally **do not** implement rename primitives — their DDL doesn't support it, and a drop+add wrapper would create a misleading illusion of rename semantics. The reconciler detects the missing primitive and falls back to the normal drop/create path for those drivers.

A companion pass in `EntityMetadataValidator` warns users when their own metadata declares two structurally identical constraints (different names, same columns), since that's a latent bug the reconciler can't resolve unambiguously.

## Compatibility

The per-driver mechanisms listed above describe the *intended* behavior on each dialect, verified at the unit-test and QueryRunner-primitive level. Actual end-to-end compatibility — that `synchronize()` / `migration:generate` output correctly routes through the rename path on real MySQL, MariaDB, Postgres, CockroachDB, MSSQL, Oracle, and SAP HANA databases — still needs to be proven by a complete cross-driver e2e suite. That suite is in progress on this branch; this PR stays Draft until every family × driver cell is green in CI.


alumni (changes requested):
I would ONLY open a PR for the drivers you are familiar with.

For the SAP driver for example, this PR does not reflect the correct behavior (e.g. it is possible to rename a primary key because it's an index, not a constraint). The code comments will also confuse future contributors and even AI tools.

I can't say anything for the rest of the drivers without doing the work that a PR author should do before opening a PR.

pkuczynski:
Thanks for the review. A few thoughts though.

This is a draft, as flagged in the description — per-driver correctness will be proven by the e2e matrix I'm still building out. That's what the draft phase is for.

On the value: today TypeORM always drops and recreates on a name change. Any driver where we emit a true rename is strictly an improvement — no regression risk because the drop/create path remains as the fallback. I've hand-patched auto-generated migrations many times over the years, and "rename instead of re-create" is one of the most-requested features still missing from the ORM.

On SAP specifically: I looked into this and I don't think `RENAME INDEX` actually solves the PK case. The PK is backed by a unique index, but the user-chosen name lives on the constraint (`SYS.CONSTRAINTS`) — the backing index is auto-named (`_SYS_TREE_CS_#...#P0`). `RENAME INDEX` targets the index object and leaves `CONSTRAINT_NAME` unchanged, which is the field TypeORM's schema diff reads, and HANA has no `ALTER TABLE ... RENAME CONSTRAINT`. So the current `DROP CONSTRAINT + ADD CONSTRAINT` path appears to be the only supported route on HANA. If there's a DDL I'm missing, I'd genuinely love to see a worked example — happy to change the approach, and the same offer stands for any other driver where you see a cleaner path.

Shipping "rename works for N drivers, falls back to drop/create for the rest" is strictly better for users than keeping today's always-drop/create behavior across the board while we wait for perfect coverage on every driver.

michaelbromley:

```
git fetch origin
```

Issue #720, opened 2017-07-31
`.save()` not updating composite key with Postgres


nickcoury:
This may be related to #477 

When operating on a table with a composite key, `.save()` only seems to insert instead of updating when a row already exists.  I can insert a set of rows the first time, but get an error when attempting the same data that the primary key constraint is violated.

If I combine the two values into a concatenated string and use that as the index, `.save()` inserts/updates as expected.  I will use this as a workaround for the time being.

typeorm@0.1.0-alpha.33
Postgres@9.6.2 on AWS RDS

```
@Entity('participants')
class Participant{
  @PrimaryColumn('int') order_id: number;
  @PrimaryColumn('text') distance: string;
  @Column('money') price?: string;
  ...
}

const participants = [...];

try {
  const result = await connection.manager.save(participants);
} catch (err) {
  console.log(err);
}
```

This code works when running the first time, and rows are created in the database as expected.  Upon running the second time, I get the following error:

```
{
  "name": "QueryFailedError",
  "length": 249,
  "severity": "ERROR",
  "code": "23505",
  "detail": "Key (order_id, distance)=(1026904, 52K) already exists.",
  "schema": "public",
  "table": "participants",
  "constraint": "participants_pkey",
  "file": "nbtinsert.c",
  "line": "433",
  "routine": "_bt_check_unique",
  "query": "INSERT INTO \"participants\"(\"order_id\", \"registration_date\", \"distance\", \"quantity\", \"price\", \"price_option\", \"order_type\", \"coupon\", \"first_name\", \"last_name\", \"gender\", \"age\", \"age_group\", \"dob\", \"email\", \"address\", \"city\", \"state\", \"zip\", \"country\", \"phone\", \"removed\", \"bib\", \"team_name\", \"emergency_name\", \"emergency_phone\", \"statement_id\") VALUES ($1,$2,$3,$4,$5,$6,$7,$8,$9,$10,$11,$12,$13,$14,$15,$16,$17,$18,$19,$20,$21,$22,$23,$24,$25,$26,$27) ",
  "parameters": [
    "1026904",
    "2017-01-28T23:30:15.000Z",
    "52K",
    "1",
    "63.78",
    "52K",
    "Credit Card",
    "0",
    "Test",
    "User",
    "M",
    "47",
    "",
    "01/01/1970",
    "testuser@gmail.com",
    "1234 Test Street",
    "Scottsdale",
    "AZ",
    "85260",
    "USA",
    "6025555555",
    "False",
[...]

PR #12407, opened 2026-04-23
fix(persistence): restore bigint precision for single JoinColumn relations

smith-xyz:
### Description of change

Relates to #12337

#10672 split `getEntityValueMap()` into separate paths for single vs composite JoinColumns. The single-key path switched from `getEntityValueMap()` → `createValueMap()` to `getEntityValue()`, which bypasses the `String()` bigint protection added in #720. This restores the unified code path for both single and composite JoinColumns.

### Pull-Request Checklist

### DIFF
--- a/src/metadata/ColumnMetadata.ts
+++ b/src/metadata/ColumnMetadata.ts
@@ -690,29 +690,19 @@ export class ColumnMetadata {
                 entity[this.relationMetadata.propertyName] &&
                 ObjectUtils.isObject(entity[this.relationMetadata.propertyName])
             ) {
-                if (this.relationMetadata.joinColumns.length > 1) {
-                    const map = this.relationMetadata.joinColumns.reduce(
-                        (map, joinColumn) => {
-                            const value =
-                                joinColumn.referencedColumn!.getEntityValueMap(
-                                    entity[this.relationMetadata!.propertyName],
-                                )
-                            if (value === undefined) return map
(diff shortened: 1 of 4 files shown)


smith-xyz:
Should this be backported to 0.3.x? The regression was introduced in 0.3.21 via #10672.

alumni:
> Should this be backported to 0.3.x?

@smith-xyz I would not worry about 0.3.x now, you can design the solution for 1.x (you still have time to introduce a breaking change, if that's why you were asking).

I think I wrote in a comment in a different PR that we need to discuss how certain data types are handled. If the final solution will be to normalize the values to a string / JS BigInt, that's not a problem. We won't backport, but at least it's probably the right step into the future.

smith-xyz:
> > Should this be backported to 0.3.x?
> 
> @smith-xyz I would not worry about 0.3.x now, you can design the solution for 1.x (you still have time to introduce a breaking change, if that's why you were asking).
> 
> I think I wrote in a comment in a different PR that we need to discuss how certain data types are handled. If the final solution will be to normalize the values to a string / JS BigInt, that's not a problem. We won't backport, but at least it's probably the right step into the future.

I agree, took a stab at this here: https://github.com/typeorm/typeorm/pull/12424 (this is just part 1, next part will decouple it further but its a bigger change)

michaelbromley:

```
git fetch origin
```

PR #12408, opened 2026-04-23
fix(query-builder): reject semicolons in groupBy and orderBy sort keys

pkuczynski:
This PR restores part of #12209 that #12396 reverted wholesale. The original change added an `assertNoSemicolon` check to every raw-SQL entry point on the query builders (`select` / `addSelect` / `groupBy` / `addGroupBy` / `orderBy` / `addOrderBy` / `validateOrderByCondition`). The revert removed all of them after a single legitimate concern surfaced: `STRING_AGG(TEAM_NAME, ';' ORDER BY TEAM_NAME) AS TEAM_NAMES` is a valid `select()` expression with `';'` as a character literal, and a blunt `.includes(";")` check rejected it. The concern was real — but it was specific to `select()` / `addSelect()`, where the argument is an arbitrary SQL expression that can contain string literals. Group keys and sort keys are the opposite: column references or short expressions where `;` has no legitimate place.

This PR restores the `;` rejection on `groupBy` / `addGroupBy` / `orderBy` (string-sort form) / `addOrderBy` across `SelectQueryBuilder`, `SoftDeleteQueryBuilder`, and `UpdateQueryBuilder`, plus the `OrderByCondition` key-iteration path. These methods never need a literal `;`; the check is a flat reject. The `select()` / `addSelect()` story lands separately in a follow-up with a string-literal-aware scanner that allows `';'` inside quoted SQL literals while still blocking `foo; DROP TABLE x`.

Worth noting the automated reviewers on the revert were unimpressed: Qodo flagged 1 bug and 2 rule violations on #12396 itself, and SonarCloud's quality gate failed on it (5.8% duplication on new code against the 3% threshold). The right response to the `STRING_AGG` case is a narrower scanner for `select()` — not dropping the entire defence for every method.

## What's in this PR

- Restores `assertNoSemicolon()` on `QueryBuilder` and wires it into `validateOrderByCondition` (for `OrderByCondition` keys).
- Guards `SelectQueryBuilder.groupBy`, `addGroupBy`, `orderBy` (string-sort form), and `addOrderBy`.
- Guards `SoftDeleteQueryBuilder.orderBy` / `addOrderBy` and `UpdateQueryBuilder.orderBy` / `addOrderBy`.
- Adds 23 regression tests in `test/functional/query-builder/sql-injection/sql-injection.test.ts` covering each method across every supported driver (MongoDB and Spanner are not in the suite).
- Adds a migration-guide entry in `docs/docs/releases/1.0/02-upgrading-from-0.3.md` scoped to sort/group keys.

## Compatibility

Anyone who was passing a literal `;` in `groupBy` / `orderBy` / `addGroupBy` / `addOrderBy` or in an `OrderByCondition` key will now get a `TypeORMError` at runtime. That input has no legitimate use case — user-controlled values belong in parameter bindings, and column references don't contain semicolons — so this is strictly more defensive. Noted in the migration guide.

### DIFF
--- a/docs/docs/releases/1.0/02-upgrading-from-0.3.md
+++ b/docs/docs/releases/1.0/02-upgrading-from-0.3.md
@@ -787,7 +787,7 @@ The removed type is `FindOptionsRelationByString`.
 
 ### Semicolons rejected in sort and group expression methods
 
-The `groupBy()`, `addGroupBy()`, `orderBy()`, and `addOrderBy()` methods on all query builders (`SelectQueryBuilder`, `UpdateQueryBuilder`, `SoftDeleteQueryBuilder`) now reject inputs containing semicolons at runtime to prevent SQL injection attacks. Group and sort keys are expected to be column names or short expressions — there is no legitimate use of `;` in these inputs. 
+The `groupBy()`, `addGroupBy()`, `orderBy()`, and `addOrderBy()` methods on all query builders (`SelectQueryBuilder`, `UpdateQueryBuilder`, `SoftDeleteQueryBuilder`) now reject inputs containing semicolons at runtime to prevent SQL injection attacks. Group and sort keys are expected to be column names or short expressions — there is no legitimate use of `;` in these inputs.
 
 ```typescript
 // All of these now throw TypeORMError
--- a/docs/docs/releases/1.0/02-upgrading-from-0.3.md
+++ b/docs/docs/releases/1.0/02-upgrading-from-0.3.md
@@ -785,6 +785,23 @@ The removed type is `FindOptionsRelationByString`.
 
 ## QueryBuilder
 
+### Semicolons rejected in sort and group expression methods
+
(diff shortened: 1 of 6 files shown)


Cprakhar (dismissed):
qodo's `1. Order/nulls injection gap` makes sense.

gioboa (approved):

smith-xyz (approved):
this one looks good to me - since the orderby/groupby wasn't problematic

alumni:
> since the orderby/groupby wasn't problematic

It's the same thing as select, I just didn't see necessary to add tests :)

We can discuss in a future meeting - in the meantime we should NOT release this (or the other PR) in 1.0

As mentioned in the revert PR, the 0.3 behavior is a functional requirement for the `QueryBuilder`.

alumni:
As discussed, we won't validate any raw SQL (only allowed in some places in `QueryBuilder`) since the purpose of `QueryBuilder` is let the users create SQL statements in any form.

PR #12409, opened 2026-04-23
fix(query-builder): reject statement terminators in select and addSelect

pkuczynski:
Second half of the split from #12209 / #12396. PR #12408 restored the simple `;` reject on `groupBy` / `addGroupBy` / `orderBy` / `addOrderBy` (where there is never a legitimate `;`). This PR handles `select()` / `addSelect()`, where the argument is an arbitrary SQL expression and a `;` *can* be legitimate — specifically, as a character literal inside a quoted string. The motivating case is `STRING_AGG(post.name, ';' ORDER BY post.name) AS tags`, which the blunt `.includes(";")` check in the original #12209 wrongly rejected.

The solution here is a small string-literal-aware scanner: `assertNoStatementTerminator(value, context)` walks the input and recognises the quote forms that SQL engines accept, skipping their contents. Only a `;` that appears outside any quoted region trips the check.

## Quote forms recognised

| Form                    | Dialect                                    |
| ----------------------- | ------------------------------------------ |
| `'foo''bar'`            | single-quoted, SQL-standard doubled quote  |
| `'foo\'bar'`            | single-quoted, MySQL backslash escape      |
| `"foo;bar"`             | double-quoted identifier / string          |
| `` `foo;bar` ``         | backtick-quoted identifier (MySQL)         |
| `[foo;bar]`             | bracket-quoted identifier (MSSQL)          |
| `$$foo;bar$$`           | dollar-quoted string (Postgres)            |
| `$tag$foo;bar$tag$`     | tagged dollar-quoted string (Postgres)     |

Comments (`-- …`, `/* … */`) are intentionally **not** treated as quoted — a `;` inside a comment still trips the check. Passing a comment through a column-expression API is already unusual, so the defence stays conservative. The migration guide calls this out. Unterminated quotes leave the scanner in "inside" mode until end-of-input; the runtime SQL parser will reject the malformed query anyway, so statement stacking can't slip through.

## What's in this PR

- Adds `QueryBuilder.assertNoStatementTerminator()` (protected) with a private `skipQuotedRegion()` helper covering all the forms above.
- Wires the check into `QueryBuilder.select()`, `SelectQueryBuilder.select()` (both the string and string[] branches), and `SelectQueryBuilder.addSelect()` (both branches). Subquery and function-form selections are unaffected — they don't carry raw SQL from the caller.
- Adds 22 regression tests in `test/functional/query-builder/sql-injection/sql-injection.test.ts` split into unquoted-semicolon (rejected) and quoted-semicolon (accepted) tables, covering every quote form.
- Adds a migration-guide entry in `docs/docs/releases/1.0/02-upgrading-from-0.3.md` that shows the accepted vs rejected shapes side-by-side, so users hitting the new error have a concrete reference.

## Compatibility

Users with legitimate `STRING_AGG(col, ';' ORDER BY col)` / dollar-quoted / backtick / bracket expressions continue to work — they were broken by the original #12209 and are restored by this scanner. Callers passing a literal unquoted `;` will now get a `TypeORMError` at runtime; that input has no legitimate use case (user-controlled values belong in parameter bindings). The only behavioural change that could surprise someone is the comment rule — if anyone was threading `;` through a `-- comment` inside `select()`, they'll see the error. That pattern is obscure enough that the defence wins.

## Follow-up

None planned — #12408 covers group/sort and this PR covers select/addSelect, closing out the split.

### DIFF
--- a/docs/docs/releases/1.0/02-upgrading-from-0.3.md
+++ b/docs/docs/releases/1.0/02-upgrading-from-0.3.md
@@ -785,6 +785,30 @@ The removed type is `FindOptionsRelationByString`.
 
 ## QueryBuilder
 
+### Statement terminators rejected in `select()` and `addSelect()`
+
+`SelectQueryBuilder.select()` and `addSelect()` now reject statement-terminator semicolons in their raw-SQL string arguments to prevent SQL statement stacking attacks. The check is **string-literal-aware**: a `;` that sits inside a quoted string literal or quoted identifier is allowed through, while an unquoted `;` is rejected. The scanner recognises single-quoted strings (including `''` and `\'` escape forms), double-quoted strings / identifiers, backtick identifiers (MySQL), bracket identifiers (MSSQL), and dollar-quoted strings (Postgres `$$…$$` and tagged `$tag$…$tag$`).
+
+```typescript
+// Legitimate expressions continue to work — the `;` is inside a string literal.
+qb.select("STRING_AGG(post.name, ';' ORDER BY post.name) AS tags")
+qb.addSelect(["post.id", "$$tagged; literal$$"])
+
+// Raw statement stacking is rejected.
+qb.select("post.id; DROP TABLE post") // TypeORMError
+qb.addSelect([
+    "post.id",
+    "COUNT(*); TRUNCATE post", // TypeORMError
+])
+
+// Comments are not treated as quoted regions — a `;` inside `-- …` or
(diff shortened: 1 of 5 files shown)


gioboa (approved):

alumni:
As discussed, we won't validate any raw SQL (only allowed in some places in `QueryBuilder`) since the purpose of `QueryBuilder` is let the users create SQL statements in any form.

PR #12412, opened 2026-04-24
fix(query-runner): auto-create user-defined schemas if not exists

### LINKED ISSUE #5382: TypeORM fails on non-existing schema
**Issue type:**

[ ] question
[x] bug report
[x] feature request
[ ] documentation issue

**Database system/driver:**

[ ] `cordova`
[ ] `mongodb`
[ ] `mssql`
[ ] `mysql` / `mariadb`
[ ] `oracle`
[x] `postgres`
[ ] `cockroachdb`
[ ] `sqlite`
[ ] `sqljs`
[ ] `react-native`
[ ] `expo`

**TypeORM version:**

[x] `latest`

With PostgresConnectionOptions:

```js
type: 'postgres',
host: host,
port: port,
username: username,
password: password,
database: database,
entities: [entities],
schema: 'schemaname',
```

If TypeORM runs a migration with a non-existing schema 'schemaname', the migration will fail because the schema does not exist. Even though it is possible to add the schema creation in a migration file, it will still fail because it will try to create the migration table in the non-existing schema. TypeORM should have the option to automatically create schemas. For my usecase this is required.

I have a working fork which can do this, if there is interest, I will make a pull request.

Cprakhar:
Closes #5382

### Description of change
This PR enables the automatic creation of a user-defined schema.

- PostgreSQL, CockroachDB, SAP HANA, and SQL Server now auto-create user-defined schema if provided during `@Entity({schema: ...})` or in `schema` option in DataSourceOptions supported by the driver.
- Removed `schema` creation logic from the `test-utils.ts`.
- Removed debug log from `database-schema/enums/enums.test.ts`.
- Added a test description in `persistence/persistence-order/persistence-order.test.ts`.
- Fix typo in `SELECT * FROM "${database}"."sys"."schema"` - `"sys"."schema"` -> `"sys"."schemas"`. (`SqlServerQueryRunner.ts`)
- Updated `columnsSql` query in CockroachDB loadTables to query correct columns from the schema and table.
  **Why:** Let's say we have two tables, `public.post` and `custom.post`, both having `name` as a column. With `columnsSql`, it was returning two columns `name` for the table `public.post`. As a result, CI was failing.

**Note:** SAP HANA driver with a custom schema (non-existing) defined in DataSourceOptions throws an error (skipped tests). In the test suite, only `@Entity({ schema: ... })` was tested.

### Testing
- [x] `multi-schema-and-database` test suite to pass

### Pull-Request Checklist

- [x] This pull request links relevant issues as `Fixes #5382`

### DIFF
--- a/test/functional/multi-schema-and-database/custom-junction-schema/custom-junction-schema.test.ts
+++ b/test/functional/multi-schema-and-database/custom-junction-schema/custom-junction-schema.test.ts
@@ -14,7 +14,7 @@ describe("multi-schema-and-database > custom-junction-schema", () => {
     before(async () => {
         dataSources = await createTestingConnections({
             entities: [Post, Category],
-            enabledDrivers: ["mssql", "postgres"],
+            enabledDrivers: ["mssql", "postgres", "sap", "cockroachdb"],
         })
     })
     beforeEach(() => reloadTestingDatabases(dataSources))
--- a/test/functional/multi-schema-and-database/custom-junction-schema/custom-junction-schema.test.ts
+++ b/test/functional/multi-schema-and-database/custom-junction-schema/custom-junction-schema.test.ts
@@ -11,11 +11,12 @@ import { expect } from "chai"
 
 describe("multi-schema-and-database > custom-junction-schema", () => {
     let dataSources: DataSource[]
-    before(async () => {
+    before(async function () {
         dataSources = await createTestingConnections({
             entities: [Post, Category],
             enabledDrivers: ["mssql", "postgres", "sap", "cockroachdb"],
         })
+        if (!dataSources.length) this.skip()
     })
     beforeEach(() => reloadTestingDatabases(dataSources))
     after(() => closeTestingConnections(dataSources))
--- a/src/driver/cockroachdb/CockroachQueryRunner.ts
+++ b/src/driver/cockroachdb/CockroachQueryRunner.ts
@@ -3221,8 +3221,8 @@ export class CockroachQueryRunner
             `SELECT "columns".*, "attr"."attgenerated" as "generated_type", ` +
             `pg_catalog.col_description((quote_ident(table_catalog) || '.' || quote_ident(table_schema) || '.' || quote_ident(table_name))::regclass::oid, ordinal_position) as description ` +
             `FROM "information_schema"."columns" ` +
-            `LEFT JOIN "pg_class" AS "cls" ON "cls"."relname" = "table_name" ` +
-            `LEFT JOIN "pg_namespace" AS "ns" ON "ns"."oid" = "cls"."relnamespace" AND "ns"."nspname" = "table_schema" ` +
+            `LEFT JOIN "pg_namespace" AS "ns" ON "ns"."nspname" = "table_schema" ` +
+            `LEFT JOIN "pg_class" AS "cls" ON "cls"."relnamespace" = "ns"."oid" AND "cls"."relname" = "table_name" ` +
             `LEFT JOIN "pg_attribute" AS "attr" ON "attr"."attrelid" = "cls"."oid" AND "attr"."attname" = "column_name" AND "attr"."attnum" = "ordinal_position" ` +
             `WHERE "is_hidden" = 'NO' AND ` +
             columnsCondiiton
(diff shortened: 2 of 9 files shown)


test/functional/multi-schema-and-database/custom-junction-schema/custom-junction-schema.test.ts
```diff
@@ -11,11 +11,12 @@ import { expect } from "chai"
 
 describe("multi-schema-and-database > custom-junction-schema", () => {
     let dataSources: DataSource[]
-    before(async () => {
+    before(async function () {
```
pkuczynski, line 14: We do arrow functions in other places, so I would rather stick to the same convention...

test/functional/multi-schema-and-database/custom-junction-schema/custom-junction-schema.test.ts
```diff
@@ -11,11 +11,12 @@ import { expect } from "chai"
...
         dataSources = await createTestingConnections({
             entities: [Post, Category],
-            enabledDrivers: ["mssql", "postgres"],
+            enabledDrivers: ["mssql", "postgres", "sap", "cockroachdb"],
         })
+        if (!dataSources.length) this.skip()
```
pkuczynski, line 19: This is pretty neat if it works. Can we maybe combine this into `createTestingConnections` so it applies everywhere out of the box?

pkuczynski (commented):
I will review with more details once all qodo feedback is resolved, let me know...

test/functional/multi-schema-and-database/custom-junction-schema/custom-junction-schema.test.ts
Cprakhar: Arrow functions bind `this`, cannot access `skip()`.

alumni:
I'm not sure that the referenced issue should be considered valid. We should not automatically create the schema - this is a side effect and libraries should not have side-effects.

I can tell that at least for one driver the connection will fail before the code added in this PR will be reached. IMO we should close both the issue and the PR.

test/functional/multi-schema-and-database/custom-junction-schema/custom-junction-schema.test.ts
pkuczynski: Its quote contrary, arrow functions do not bind this, they inhetit it from the parent scope, which in our case is module. So you are right we need `function`. But then its a pattern we do not use in 1000+ other tests, so we either add this as a general convention or don’t use it. We should stay consistent...

pkuczynski:
> I'm not sure that the referenced issue should be considered valid. We should not automatically create the schema - this is a side effect and libraries should not have side-effects.

Why not? We can create schemas inside migration, so we could detect misisng schema and create it there. This is what issue author requested. But this should be done on schema builder level and not query runner.

test/functional/multi-schema-and-database/custom-junction-schema/custom-junction-schema.test.ts
Cprakhar: I tried it in #12397

Cprakhar:
@pkuczynski, I have incorporated your suggestion by doing it on the schema builder level. 

Want to take a look? Should I reopen this PR?

pkuczynski:
> @pkuczynski, I have incorporated your suggestion by doing it on the schema builder level. 
> 
> 
> 
> Want to take a look? Should I reopen this PR?

Let's add it as an item for next maintainers meeting. I would be tempted to say "yes", as I don't see any reason we might not want to do it in migrations? But we should talk with Lucian first...

alumni:
Seems this PR does something different than the issue it references - the issue should be kept closed because the schema in DataSourceOptions needs to exist before the connection is established. Also, unless we operate in `sync: true` mode (local development in simple projects), we should not introduce any DDL changes except if the user explicitly asks for them. So the issue should be closed as not planned.

In this PR, we create the database when creating the table. This is a side effect, as it does more than the user intended. If you look at the Unix tools (touch, mkdir etc), they won't create anything more than what is requested from them. In some cases, the user can explicitly ask for it, but that's expressing the intent for such behavior, it's never the default behavior.

pkuczynski:
I agree this PR was not a good approach. However we could create schemas when migration is generated. Schema does not need to be specific as part of the DataSourceOptions. Can be defined per entity and different than the default one. My understanding is that this is what the issue was about?

PR #12413, opened 2026-04-25
fix(cockroachdb): adjust join in loadTables to load correct table columns

Cprakhar:
### Description of change
- Removed debug log from `database-schema/enums/enums.test.ts`.
- Added a test description in `persistence/persistence-order/persistence-order.test.ts`.
- Fix typo in `SELECT * FROM "${database}"."sys"."schema"` - `"sys"."schema"` -> `"sys"."schemas"`. (`SqlServerQueryRunner.ts`)
- Updated `columnsSql` query in CockroachDB loadTables to query correct columns from the schema and table.
  **Why:** Let's say we have two tables, `public.post` and `custom.post`, both having `name` as a column. With `columnsSql`, it was returning two columns `name` for the table `public.post`.
- Added CockroachDB and SAP HANA in the multi-schema-and-database test suite.

### Pull-Request Checklist


test/functional/database-schema/enums/enums.test.ts
```diff
@@ -95,8 +95,6 @@ describe("database schema > enums", () => {
                     .createSchemaBuilder()
                     .log()
 
-                console.log(sqlInMemory.upQueries)
-
                 sqlInMemory.upQueries.length.should.be.equal(0)
                 sqlInMemory.downQueries.length.should.be.equal(0)
```
alumni: Whenever we have `console.log` in tests, we are trying to figure out why an assertion is not working. This can always be solved by introducing better assertions, e.g.: `expect(array).to.have.length(0)` (or `lengthOf`?) would also log the array values if the expectation is not met ;)

test/functional/multi-schema-and-database/custom-junction-schema/custom-junction-schema.test.ts
```diff
@@ -24,25 +24,30 @@ describe("multi-schema-and-database > custom-junction-schema", () => {
         Promise.all(
             dataSources.map(async (dataSource) => {
                 const queryRunner = dataSource.createQueryRunner()
```
alumni: Instead of adding `try-finally` and introducing another level of indentation:
```suggestion
                await using queryRunner = dataSource.createQueryRunner()
```
Same can be done in the other tests.

test/functional/multi-schema-and-database/multi-schema-and-database-basic-functionality/multi-schema-and-database-basic-functionality.test.ts
```diff
@@ -221,40 +30,40 @@ describe("multi-schema-and-database > basic-functionality", () => {
...
+                    const table = await queryRunner.getTable(
                         "testDB.questions.question",
-                    ))!
-                    await queryRunner.release()
+                    )
 
-                    expect(table.database).to.be.equal("testDB")
-                    expect(table.schema).to.be.equal("questions")
-                    expect(table.name).to.be.equal("testDB.questions.question")
+                    expect(table!.database).to.equal("testDB")
```
alumni: nit (no need to change unless there's another change): this is an unnecessary change. we are now making this assertion 3 times instead of only one (before). In tests, we often know that the table exists, so we place the assertion when we read the table.

test/functional/multi-schema-and-database/multi-schema-and-database-basic-functionality/multi-schema-and-database-basic-functionality.test.ts
```diff
@@ -354,18 +165,256 @@ describe("multi-schema-and-database > basic-functionality", () => {
...
     })
+
+    describe("custom-table-schema", () => {
+        const drivers: DatabaseType[][] = [
+            // SAP throws error during connection if schema specified in connection options does not exist.
+            // only testing custom schema specified in Entity decorator
+            ["sap"],
+            ["mssql", "postgres", "cockroachdb"],
+        ]
+        for (const driver of drivers) {
```
alumni: I'm not sure what are we trying to test here, but we don't have this test structure usually. Additionally, the schema should exist before connecting to the database.

These new tests could be simplified a lot

Cprakhar: SAP HANA has a `currentSchema` option in the `@sap-client` package, which is passed if defined in `DataSourceOptions.schema`, which throws errors. Other drivers in the tests don't have this. Schemas are created beforehand using the logic in `test-utils.ts`

I will try to simplify this.

alumni: Yeah, we pass `DataSourceOptions.schema` to the client, which is fine, because the schema that you pass to the client should exist. It's simpler than executing a query to change the schema every time you create a `QueryRunner`. I only changed it like that for the SAP driver since I'm more familiar with it, but you can make that assumption for all drivers.

alumni (approved):

gioboa (approved):

alumni:
@Cprakhar FYI cherry picking to 0.3 would not be possible for tests (since they use `await using` which was added in 1.0 only)

PR #12414, opened 2026-04-25
fix(mssql): enhance handling of generated columns in metadata queries

Cprakhar:
### Description of change
- Updates `typeorm_metadata` if the table is renamed, keeping the metadata table in sync.
- `DROP+ADD` a column on column rename if the column is a generated column with an expression instead of `ALTER`.
  **Why:** SQL Server driver throws error: `Error: Cannot alter column 'someColumn' because it is 'COMPUTED'.`

### Pull-Request Checklist

### DIFF
--- a/src/driver/sqlserver/SqlServerQueryRunner.ts
+++ b/src/driver/sqlserver/SqlServerQueryRunner.ts
@@ -719,6 +719,7 @@ export class SqlServerQueryRunner
         if (generatedColumns.length > 0) {
             const parsedTableName = this.driver.parseTableName(table)
             parsedTableName.schema ??= await this.getCurrentSchema()
+            parsedTableName.database ??= await this.getCurrentDatabase()
 
             for (const column of generatedColumns) {
                 const insertQuery = this.insertTypeormMetadataSql({
@@ -801,6 +802,7 @@ export class SqlServerQueryRunner
         if (generatedColumns.length > 0) {
             const parsedTableName = this.driver.parseTableName(table)
             parsedTableName.schema ??= await this.getCurrentSchema()
+            parsedTableName.database ??= await this.getCurrentDatabase()
 
             for (const column of generatedColumns) {
                 const deleteQuery = this.deleteTypeormMetadataSql({
@@ -894,19 +896,13 @@ export class SqlServerQueryRunner
             : await this.getCachedTable(oldTableOrName)
         const newTable = oldTable.clone()
 
-        // we need database name and schema name to rename FK constraints
-        let dbName: string | undefined = undefined
-        let schemaName: string | undefined = undefined
-        let oldTableName: string = oldTable.name
-        const splittedName = oldTable.name.split(".")
-        if (splittedName.length === 3) {
-            dbName = splittedName[0]
-            oldTableName = splittedName[2]
-            if (splittedName[1] !== "") schemaName = splittedName[1]
-        } else if (splittedName.length === 2) {
-            schemaName = splittedName[0]
-            oldTableName = splittedName[1]
-        }
+        let {
+            tableName: oldTableName,
+            database: dbName,
+            schema: schemaName,
+        } = this.driver.parseTableName(oldTable.name)
+        dbName ??= await this.getCurrentDatabase()
+        schemaName ??= await this.getCurrentSchema()
 
         newTable.name = this.driver.buildTableName(
             newTableName,
@@ -943,16 +939,16 @@ export class SqlServerQueryRunner
         )
         if (hasGeneratedColumns) {
             const updateQuery = this.updateTypeormMetadataSql({
-                database: dbName ?? currentDB,
-                schema: schemaName ?? (await this.getCurrentSchema()),
+                database: dbName,
+                schema: schemaName,
                 table: oldTableName,
                 type: MetadataTableType.GENERATED_COLUMN,
                 valueToSet: { table: newTableName },
             })
 
             const revertUpdateQuery = this.updateTypeormMetadataSql({
-                database: dbName ?? currentDB,
-                schema: schemaName ?? (await this.getCurrentSchema()),
+                database: dbName,
+                schema: schemaName,
                 table: newTableName,
                 type: MetadataTableType.GENERATED_COLUMN,
                 valueToSet: { table: oldTableName },
@@ -1284,6 +1280,7 @@ export class SqlServerQueryRunner
         if (column.generatedType && column.asExpression) {
             const parsedTableName = this.driver.parseTableName(table)
             parsedTableName.schema ??= await this.getCurrentSchema()
+            parsedTableName.database ??= await this.getCurrentDatabase()
 
             const insertQuery = this.insertTypeormMetadataSql({
                 database: parsedTableName.database,
@@ -2199,8 +2196,8 @@ export class SqlServerQueryRunner
 
         if (column.generatedType && column.asExpression) {
             const parsedTableName = this.driver.parseTableName(table)
-
             parsedTableName.schema ??= await this.getCurrentSchema()
+            parsedTableName.database ??= await this.getCurrentDatabase()
 
             const deleteQuery = this.deleteTypeormMetadataSql({
                 database: parsedTableName.database,
--- a/src/driver/sqlserver/SqlServerQueryRunner.ts
+++ b/src/driver/sqlserver/SqlServerQueryRunner.ts
@@ -938,6 +938,30 @@ export class SqlServerQueryRunner
             ),
         )
 
+        const hasGeneratedColumns = oldTable.columns.some(
+            (col) => col.generatedType && col.asExpression,
+        )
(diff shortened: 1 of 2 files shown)


src/driver/sqlserver/SqlServerQueryRunner.ts
```diff
@@ -894,19 +896,13 @@ export class SqlServerQueryRunner
...
-        const splittedName = oldTable.name.split(".")
-        if (splittedName.length === 3) {
-            dbName = splittedName[0]
-            oldTableName = splittedName[2]
-            if (splittedName[1] !== "") schemaName = splittedName[1]
-        } else if (splittedName.length === 2) {
-            schemaName = splittedName[0]
-            oldTableName = splittedName[1]
-        }
+        let {
```
smith-xyz, line 899: at first glance these changes didn't look related to what was being fix. Now dbName and schemaName are always populated (from the driver config), whereas before they could be undefined - just want to make sure this was intentional and doesn't have any impact downstream in the code

smith-xyz (changes requested):
I'm not sure I agree with this change. Returning the MSSQL error is actually correct feedback to the user — sp_rename rejects computed columns because they have dependent objects (indexes, constraints). Silently routing through DROP+ADD drops those dependents without recreating them. The user already has dropColumn/addColumn/createIndex on QueryRunner to handle this explicitly with full control.

I'm open to how this can improve but I don't see this as the right fix.

this one may collide a bit with #11620

Cprakhar:
> I'm not sure I agree with this change. Returning the MSSQL error is actually correct feedback to the user — sp_rename rejects computed columns because they have dependent objects (indexes, constraints). Silently routing through DROP+ADD drops those dependents without recreating them. The user already has dropColumn/addColumn/createIndex on QueryRunner to handle this explicitly with full control.
> 
> I'm open to how this can improve but I don't see this as the right fix.
> 
> this one may collide a bit with #11620

If it has dependent objects, even then it won't allow dropping:
`Error: ALTER TABLE DROP COLUMN name failed because one or more objects access this column.`
```ts
@Entity()
@Index("IDX_USER_NAME_ADMIN", ["name", "isAdmin"])
@Check(`"name" IS NOT NULL OR "isAdmin" = 0`)
export class User {
    @PrimaryGeneratedColumn()
    id: number

    @Column({
        type: "varchar",
        generatedType: "STORED",
        asExpression: `CONCAT('user_', id)`,
    })
    name: string

    @Column()
    isAdmin: boolean
}
```

smith-xyz:
> > I'm not sure I agree with this change. Returning the MSSQL error is actually correct feedback to the user — sp_rename rejects computed columns because they have dependent objects (indexes, constraints). Silently routing through DROP+ADD drops those dependents without recreating them. The user already has dropColumn/addColumn/createIndex on QueryRunner to handle this explicitly with full control.
> > I'm open to how this can improve but I don't see this as the right fix.
> > this one may collide a bit with #11620
> 
> If it has dependent objects, even then it won't allow dropping: `Error: ALTER TABLE DROP COLUMN name failed because one or more objects access this column.`
> 
> ```ts
> @Entity()
> @Index("IDX_USER_NAME_ADMIN", ["name", "isAdmin"])
> @Check(`"name" IS NOT NULL OR "isAdmin" = 0`)
> export class User {
>     @PrimaryGeneratedColumn()
>     id: number
> 
>     @Column({
>         type: "varchar",
>         generatedType: "STORED",
>         asExpression: `CONCAT('user_', id)`,
>     })
>     name: string
> 
>     @Column()
>     isAdmin: boolean
> }
> ```

Ahh okay good example, but now I'm seeing three outcomes then: 

| Scenario | Dependents | Result | Side Effects |
|----------|-----------|--------|-------------|
| Bare computed column | None | Rename succeeds | Column moves to end of table; duplicate metadata entries |
| Computed + single-column index | `@Index` on column | Rename succeeds | Index silently dropped, not recreated |
| Computed + multi-column index + check | `@Index` on multiple cols, `@Check` | Rename fails | Table unchanged — MSSQL blocks the DROP |

Compared to current behavior all three lead to error but do so consistently

<details>
<summary>Test entities for the three scenarios</summary>

```typescript
@Entity()
export class BareComputed {
    @PrimaryGeneratedColumn()
    id: number
    @Column()
    firstName: string
    @Column()
    lastName: string
    @Column({
        asExpression: `CONCAT("firstName", ' ', "lastName")`,
        generatedType: "STORED",
    })
    fullName: string
}

@Entity()
export class SingleColIndex {
    @PrimaryGeneratedColumn()
    id: number
    @Column()
    firstName: string
    @Column()
    lastName: string
    @Index("IDX_single_fullName")
    @Column({
        asExpression: `CONCAT("firstName", ' ', "lastName")`,
        generatedType: "STORED",
    })
    fullName: string
}

@Entity()
@Index("IDX_multi_name_active", ["fullName", "isActive"])
@Check(`"fullName" IS NOT NULL OR "isActive" = 0`)
export class MultiColDeps {
    @PrimaryGeneratedColumn()
    id: number
    @Column()
    firstName: string
    @Column()
    lastName: string
    @Column({
        asExpression: `CONCAT("firstName", ' ', "lastName")`,
        generatedType: "STORED",
    })
    fullName: string
    @Column({ default: true })
    isActive: boolean
}
```

src/driver/sqlserver/SqlServerQueryRunner.ts
Cprakhar, line 1406:
```suggestion
) {
// SQL Server does not support changing of IDENTITY column, so we must drop column and recreate it again.
// Also, we recreate column if column type changed

// drop dependent objects
const dropIndices = clonedTable.indices.filter((index) =>
  index.columnNames.includes(oldColumn.name),
)
const dropChecks = clonedTable.checks.filter((check) =>
  check.columnNames?.includes(oldColumn.name),
)
const dropForeignKeys = clonedTable.foreignKeys.filter((fk) =>
  fk.columnNames.includes(oldColumn.name),
)
const dropUniqueConstraints = clonedTable.uniques.filter((unique) =>
  unique.columnNames.includes(oldColumn.name),
)
await this.dropIndices(table, dropIndices)
await this.dropCheckConstraints(table, dropChecks)
await this.dropForeignKeys(table, dropForeignKeys)
await this.dropUniqueConstraints(table, dropUniqueConstraints)

await this.dropColumn(table, oldColumn)
await this.addColumn(table, newColumn)

// add the dependent objects back with new column name
const addIndices = dropIndices.map((index) => {
  index.columnNames.splice(
    index.columnNames.indexOf(oldColumn.name),
    1,
    newColumn.name,
  )
  return index
})
const addChecks = dropChecks.map((check) => {
  if (check.columnNames) {
    check.columnNames.splice(
      check.columnNames.indexOf(oldColumn.name),
      1,
      newColumn.name,
    )
  }
  if (check.expression) {
    check.expression = check.expression.replaceAll(
      new RegExp(String.raw`\b${oldColumn.name}\b`, "g"),
      newColumn.name,
    )
  }
  return check
})
const addForeignKeys = dropForeignKeys.map((fk) => {
  fk.columnNames.splice(
    fk.columnNames.indexOf(oldColumn.name),
    1,
    newColumn.name,
  )
  return fk
})
const addUniqueConstraints = dropUniqueConstraints.map((unique) => {
  unique.columnNames.splice(
    unique.columnNames.indexOf(oldColumn.name),
    1,
    newColumn.name,
  )
  return unique
})
await this.createUniqueConstraints(table, addUniqueConstraints)
await this.createIndices(table, addIndices)
await this.createCheckConstraints(table, addChecks)
await this.createForeignKeys(table, addForeignKeys)
            
// update cloned table
clonedTable = table.clone()
} else {
```

How about this? I know this is not a very good approach, but it will work.

smith-xyz: yeah just really complex here. The part that I think we should keep is fixing the typeorm_metadata update.

I'm wondering if we should hold off on the computed column drop+add behavior until we have some broader strategy with #3357. That issue is about migrations, but the underlying problem is the same — how TypeORM handles destructive DDL (drop+add vs alter) across drivers. Looks like that's a post 1.0 item. 

What are your thoughts? It keeps the sp_rename error, and I don't want to shut the pr down if having some various error messaging is better when a synchronization happens, but it may have unintended side effects.

src/driver/sqlserver/SqlServerQueryRunner.ts
Cprakhar: In `loadTables()`, the database and schema are always there in the where condition.
```typescript
const asExpressionQuery =
  this.selectTypeormMetadataSql({
    database: dbTable["TABLE_CATALOG"],
    schema: dbTable["TABLE_SCHEMA"],
    table: dbTable["TABLE_NAME"],
    type: MetadataTableType.GENERATED_COLUMN,
    name: tableColumn.name,
  })
```
I don't believe it will cause any impact.

Cprakhar:
I have merged this work into #12423. Closing it.

PR #12416, opened 2026-04-26
refactor!: extract legacy naming strategies

alumni:
### Description of change

Extract legacy naming strategies to new package.

### Pull-Request Checklist


Cprakhar (commented):
- Do we need to run the `legacy-naming-strategies` package tests in CI?
- Need to update `codemod` for this change as the import statement is changed

alumni:
> * Do we need to run the `legacy-naming-strategies` package tests in CI?

I'm not very fond of the current CI setup, and there aren't many tests in the package. I did it after fighting a bit with the CI - but we should get better results as part of the monorepo. I won't handle the linting/formatting, I don't like the way it's done in the codemod + it won't stay for long like this.

> * Need to update `codemod` for this change as the import statement is changed

I think I'll avoid trying to understand the codemod, at least for such an unimportant breaking change (this naming strategy is for a version of Oracle before 2013).

Cprakhar (approved):

PR #12421, opened 2026-04-29
fix(postgres): execute remaining relation-load and persistence paths sequentially to avoid pg 8.19.0 deprecation

kyungseopk1m:
Fixes the remaining half of #12238 (pg 8.19.0 deprecation warning), following the same direction as #12105 but in driver-agnostic code paths.

#12105 serialized the `PostgresQueryRunner` / `CockroachQueryRunner` metadata paths (`loadTables`, `clearDatabase`, `dropEnumTypes`). This PR addresses the relation-loading and persistence code paths that @pkuczynski pointed out in https://github.com/typeorm/typeorm/issues/12238#issuecomment-4173213373 and that @max-brigade documented in their attached patch on the same issue.

### Changed paths

- `EntityPersistExecutor` — chunk-level `SubjectExecutor` build sequentially (otherwise concurrent `SubjectDatabaseEntityLoader.load()` invocations across chunks share the same client)
- `SubjectDatabaseEntityLoader.load()` — group entity targets sequentially
- `SubjectExecutor.executeUpdateOperations()` — update remaining subjects sequentially
- `SubjectExecutor.executeSoftRemoveOperations()` — soft-delete subjects sequentially
- `SubjectExecutor.executeRecoverOperations()` — restore subjects sequentially
- `SelectQueryBuilder` — `relationLoadStrategy: 'query'` relation loading sequentially
- `RelationIdLoader.load()` — `relationIdAttributes` loading sequentially

Each site is gated on `postgres` and `cockroachdb`, so non-postgres drivers keep the existing `Promise.all` behavior with no performance regression. Each branch carries the same comment: `// Avoid concurrent queries on the same pg client; see #12238.`

### Out of scope

- `SubjectExecutor.broadcastBeforeEventsForAll` / `broadcastAfterEventsForAll` use `Promise.all(broadcasterResult.promises)` to await user-defined listener/subscriber promises. Those are not typeorm-issued `client.query()` calls on the shared pg client, so they remain `Promise.all`.

### Tests

Added regression test under `test/github-issues/12238/`:

- "should keep relationLoadStrategy: 'query' working with ManyToMany on every supported driver except spanner" — runs against every configured driver to assert behavioral equivalence between the parallel and the sequential branches.
- "should not emit pg 'client.query already executing' warning during relationLoadStrategy: 'query' on postgres and cockroachdb". Captures `process.on('warning', ...)` and exercises:
  - bulk `manager.save(posts)`
  - `manager.save(posts, { chunk: 1 })` to drive the `EntityPersistExecutor` chunk path
  - `manager.find(Post, { relations: { tags: true }, relationLoadStrategy: "query" })`
  - re-`save`, `softRemove`, `recover`

  Asserts the "Calling client.query() when the client is already executing" warning does not fire on any of them.

Closes #12238


Cprakhar (approved):

gioboa (approved):
Thanks @kyungseopk1m

alumni:
We need to discuss whether to revert this PR or fix the behavior to be consistent across all drivers.

Those issues that Qodo identified are really on-point.

`QueryRunner`s are the ones responsible for interfacing with client, not `SubjectDatabaseEntityLoader.ts‎` / `EntityPersistExecutor.ts` / `SubjectExecutor.ts` / `RelationIdLoader.ts`‎ / `SelectQueryBuilder.ts`. Those should be as driver-agnostic as they can possibly be.

While I don't usually like to see `Promise.all` because it causes non-deterministic behavior, it does help parallelize tasks - so this change might have an impact on performance, in an area that is prone to such issues.

In the `QueryRunner` we normally have a lock that prevents multiple queries to be executed simultaneously on the same DB connection. IMO, I would have looked for an issue there first.

max-brigade:
My understanding is that Promise.all does not provide real query parallelism when all queries share the same physical transaction/connection. At best, the driver serializes them internally; at worst, the driver rejects or warns about concurrent usage. In pg specifically, that implicit internal queue existed historically, but it is now being deprecated because the driver does not want to support concurrent client.query() calls on an already-busy client. So for TypeORM paths that share one QueryRunner/client, sequential await seems like the correct behavior rather than a performance regression.

PR #12422, opened 2026-04-30
fix(docs): remove unsupported 'remove' option from cascade type definition

UmanMushtaq:
Fixes #[issue number]

The `"remove"` option in cascade relation typing is documented but not 
functionally supported. Unlike `insert`, `update`, `soft-remove`, and 
`recover`, the `remove` cascade is not implemented consistently in 
SubjectExecutor.

This PR removes `"remove"` from the cascade TypeScript union type in 
`RelationOptions.ts` and updates the JSDoc comment to reflect the 
actual supported options, preventing users from relying on undocumented 
behaviour.

Changes:
- `src/decorator/options/RelationOptions.ts` — removed `"remove"` from 
  cascade union type and updated JSDoc

### DIFF
--- a/src/decorator/options/RelationOptions.ts
+++ b/src/decorator/options/RelationOptions.ts
@@ -12,12 +12,12 @@ export interface RelationOptions {
      * You can separately restrict cascades to insertion or updation using following syntax:
      *
      * @example
-     * cascade: ["insert", "update", "remove", "soft-remove", "recover"] // include or exclude one of them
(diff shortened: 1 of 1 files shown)


UmanMushtaq:
After reviewing the bot's findings, "remove" is actually supported at runtime in CascadesSubjectBuilder.ts. Removing it from the type creates a runtime/type mismatch. Closing this PR. The issue may need a documentation clarification instead. Thank you for the review.

Issue #11516, opened 2025-06-10
Upsert failed when the record is updated instead of inserting, using on tables with id of number type.


tungnat97:
### Issue description

Using queryRunner.manager.upsert() on tables whose id's type is number will result in error: `Cannot update entity because entity id is not set in the entity.`

### Expected Behavior

The upsert runs successfully.

### Actual Behavior

Error thrown:  `Cannot update entity because entity id is not set in the entity.`, in ReturningResultsEntityUpdator.insert()
```
const entityIds = entities.map((entity) => {
    const entityId = metadata.getEntityIdMap(entity);
    // We have to check for an empty `entityId` - if we don't, the query against the database
    // effectively drops the `where` clause entirely and the first record will be returned -
    // not what we want at all.
    if (!entityId) {
        throw new error_1.TypeORMError(`Cannot update entity because entity id is not set in the entity.`);
    }

    return entityId;
});
```

The upsert should run properly if the id's type is bigint, and fail with other types, since in ColumnMetadata.createValueMap(), it converts a field value to string if
```
this.generationStrategy === "increment" ||
this.generationStrategy === "rowid") &&
this.type === "bigint" &&
value !== null
```
The process update instead of inserting the record, so `insert_id` will be undefined. Hence, the upsert will not throw the aforementioned error if the id is of bigint type, as the id will appear as `"undefined"`.

### Steps to reproduce

Create a table `Test` with id as follow:

```
@PrimaryGeneratedColumn("increment")
id: number;
```

and a unique pair of keys

```
@Index(["key1", "key2"], { unique: true })
...
@Column()
key1: string;

@Column()
key2: string;
```

finally a value field for upserting:
...
@Column()
value: string;
```


Then just run `queryRunner.manager.upsert(Test, {key1: "a", key2: "b", value, "c"}, ["key1", "key2"])` (run 2 times if the record isn't in the table)


### My Environment

| Dependency          | Version  |
| ---                 | ---      |
| Operating System    | Windows         |
| Node.js version     |22.15.0  |
| Typescript version  | 5.7.2  |
| TypeORM version     | 0.3.20  |


### Additional Context

_No response_

### Relevant Database Driver(s)

- [ ] aurora-mysql
- [ ] aurora-postgres
- [ ] better-sqlite3
- [ ] cockroachdb
- [ ] cordova
- [ ] expo
- [ ] mongodb
- [x] mysql
- [ ] nativescript
- [ ] oracle
- [ ] postgres
[...]

Issue #8229, opened 2021-09-30
Add a column option jsonType: 'object' | 'string' which is directly analogous to the hstoreType option for HSTORE columns for Postgres


softwareplumber:
## Feature Description
Add a column option jsonType: 'object' | 'string' which is directly analogous to the hstoreType option for HSTORE columns.
### The Problem
Postgres jsonb columns support essentially arbitrary-precision numbers. The problem is that there is no way to get these numbers in or out of postgres using typeorm because the default JSON parser/serializer is limited to the precision supported by javascript (before BigInt). Therefore we need a passthrough mechanism that allows us to treat jsonb columns as strings.
### The Solution
Exactly as for HSTORE columns, allow them to be treated as strings
### Considered Alternatives
see my original proposal in issue #8128
### Additional Context
I already submitted PR #8141 to implement this but I don't think anyone has looked at it since its linked to my original proposal for fixing the core problem.
### Relevant Database Driver(s)

| DB Type              | Relevant      |
| ---                  | ---           |
| `aurora-data-api`    | no            |
| `aurora-data-api-pg` | no            |
| `better-sqlite3`     | no            |
| `cockroachdb`        | no            |
| `cordova`            | no            |
| `expo`               | no            |
| `mongodb`            | no            |
| `mysql`              | no            |
| `nativescript`       | no            |
| `oracle`             | no            |
| `postgres`           | yes           |
| `react-native`       | no            |
| `sap`                | no            |
| `sqlite`             | no            |
| `sqlite-abstract`    | no            |
| `sqljs`              | no            |
| `sqlserver`          | no            |

### Are you willing to resolve this issue by submitting a Pull Request?

 - ✅ Yes, I have the time, and I know how to start.
 - ✖️ Yes, I have the time, but I don't know how to start. I would need guidance.
 - ✖️ No, I don’t have the time, but I can support (using donations) development.
 - ✖️ No, I don’t have the time and I’m okay to wait for the community / maintainers to resolve this issue.

ericmorand:
I'm sorry for being this late - i mean more than a year and a half year - but I strongly support this feature.

ericmorand:
I think this is still relevant.

PR #12424, opened 2026-05-01
fix(metadata): unify value normalization to prevent bigint precision loss

### LINKED ISSUE #12337: `@JoinColumn` on bigint column causes silent precision loss on INSERT (since 0.3.21)

smith-xyz:
### Description of change

Introduces a ColumnValueHandler adapter that centralizes column value
normalization and comparison logic, replacing scattered inline type checks
across SubjectChangedColumnsComputer, ColumnMetadata, and OrmUtils.

Fixes #12337 — bigint precision loss on single JoinColumn relations.

Supersedes #12407.

What changed:
- ColumnValueHandler interface (normalize + areEqual) with typed handlers
  for bigint, date, time, datetime, json, jsonb, simple-array, simple-enum,
  simple-json, and a default fallback with Uint8Array/.equals() support
- resolveValueHandler maps column metadata to the correct handler, with a
  driver-override hook (Driver.resolveValueHandler) for driver-specific needs
- DataSource.valueHandlerResolver provides a swappable DI seam, mirroring
  the existing namingStrategy pattern
- ColumnMetadata.normalizeValue() and valuesEqual() delegate to the resolved
  handler, removing ad-hoc type logic from SubjectChangedColumnsComputer
- Per-driver integration tests (mysql, mariadb, postgres, mssql, cockroachdb,
  better-sqlite3) covering date/time, json, binary, simple-array, simple-enum,
  bigint FK relations, and basic type change detection

Lays groundwork for #11516, #8229, #3357. A follow up PR will expand handlers with
hydrate/persist methods to consolidate prepareHydratedValue across drivers.

(follow-up PR)
The handler interface gains optional `hydrate()` and `persist()` methods,
replacing the per-driver `prepareHydratedValue`/`preparePersistentValue` logic:
- Each handler owns its wire-format conversion (e.g., `IntValueHandler.hydrate`
  calls `parseInt`, `DateTimeValueHandler.hydrate` calls `DateUtils.normalizeHydratedDate`)
- Drivers delegate to the handler first; only truly driver-specific coercions
  remain in the driver (geometry, driver-specific array encoding, etc.)
- Fixes CockroachDB `int` returning as string (#TBD) without ad-hoc parseInt
  additions to the driver
- Reduces ~200 lines of duplicated type-switching across 10 drivers into
  shared handler implementations

### Pull-Request Checklist


smith-xyz:
This seems to fix a bug where MongoDB's ObjectId was always marked as "changed" on save — the old code compared instances with === instead of .equals(), causing unnecessary writes and an inflated updatedColumns count in subscribers.

smith-xyz:
note on some of the qodo findings: 

All built-in normalizers are idempotent (normalizing an already-normalized value is a no-op)
The perf cost is negligible (a few extra typeof checks per column)
Removing normalization from getEntityValue() could break other callers that use the returned value directly (not just for comparison)
Removing it from areEqual() would break the contract that "areEqual normalizes before comparing"

some of this is going to be addressed in the next PR that will update all drivers to leverage the new api for column normalization

michaelbromley:

```
git fetch origin
```

Issue #3967, opened 2019-04-09
Strange behavior when using `limit` with relations (joins)


avilesj:
**Issue type:**

[ ] question
[X] bug report
[ ] feature request
[ ] documentation issue

**Database system/driver:**

[ ] `cordova`
[ ] `mongodb`
[ ] `mssql`
[ ] `mysql` / `mariadb`
[ ] `oracle`
[X] `postgres`
[ ] `cockroachdb`
[ ] `sqlite`
[ ] `sqljs`
[ ] `react-native`
[ ] `expo`

**TypeORM version:**

[ ] `latest`
[ ] `@next`
[0.2.7] `0.x.x` (or put your version here)

**Steps to reproduce or a small repository showing the problem:**

*Short version*

When using joins, `limit()` does not seem to work properly.

*Long Version*

I have the following query:
```javascript
      return this.branchRepo.createQueryBuilder("branches")
      .innerJoinAndSelect("branches.branchSchedules", "branch_schedules", "branch_schedules.branch_id = branches.id")
      .leftJoinAndSelect("branches.branchFeature", "branchFeature", "branchFeature.branch_id = branches.id")
      .leftJoinAndSelect("branchFeature.feature", "features")
      .orderBy("branches.is_all_day", "DESC")
      .addOrderBy("branch_schedules.end_time::time", "DESC")
      .addOrderBy("branch_schedules.day", "DESC")
      .addOrderBy(`(case when branch_schedules.day = 'holiday' then 1
                         when branch_schedules.day = 'sunday' then 2
                         when branch_schedules.day = 'saturday' then 3
                         when branch_schedules.day = 'weekday' then 4
                         end)`,
                  "ASC")
      .addOrderBy("branch_schedules.start_time::time", "ASC")
      .limit(5)
      .getMany();
```

Which produces the following query:

```sql
[...]

PR #12447, opened 2026-05-10
fix: optimize limit joins pagination path

### LINKED ISSUE #12445: .limit() performance regression in v0.3.29
### Issue description

.limit() unexpectedly causing 2 queries with Select using Query Builder

### Expected Behavior

```ts
createQueryBuilder("child")
  .innerJoinAndSelect("child.parent", "parent")
  .offset(2)
  .limit(2)
  .getMany()
```

Using v0.3.28 this worked by generating a single query with `OFFSET` and `LIMIT` clauses (MySQL).

### Actual Behavior

Using v0.3.29 the above now generates two queries - one to get distinct IDs and another to get the desired results.

This causes a performance and behaviour regression. Some queries I tested roughly doubled in total execution time.

### Steps to reproduce

Any simple query as above.

### My Environment

| Operating System    | Windows 11 Pro |
| Node.js version     | 24.15.0  |
| Typescript version  | 6.0.3 |
| TypeORM version     | 0.3.29 |

### Additional Context

Firstly, thanks for your efforts on TypeORM!

The source of this issue looks to be https://github.com/typeorm/typeorm/pull/11987. @gioboa.

Should this change have been noted in the [release notes](https://github.com/typeorm/typeorm/releases/tag/0.3.29)?

My understanding from the TypeORM documentation is that [offset](https://typeorm.io/docs/query-builder/select-query-builder#adding-offset-expression) and [limit](https://typeorm.io/docs/query-builder/select-query-builder#adding-limit-expression) behave on the underlying query result set like the MySQL `OFFSET` and `LIMIT` clauses.

In contrast, QueryBuilder also supports [pagination](https://typeorm.io/docs/query-builder/select-query-builder#using-pagination) using `take` and `skip`, which understandably requires the 2-step query process.

The change made in v0.3.29 seem to break the power of `offset` and `limit` and duplicate the `take` and `skip` feature?

As the advanced query building interface of TypeORM, I'd think QueryBuilder should support generating single-query `OFFSET` `LIMIT` queries.

- [ ] aurora-mysql
- [ ] aurora-postgres
- [ ] better-sqlite3
- [ ] capacitor
- [ ] cockroachdb
- [ ] cordova
- [ ] expo
- [ ] mongodb
- [x] mysql
- [ ] nativescript
- [ ] oracle
- [ ] postgres
- [ ] react-native
- [ ] sap
- [ ] spanner
- [ ] sqlite
- [ ] sqljs
- [ ] sqlserver

gioboa:
Close #12445 

### Pull-Request Checklist

### DIFF
--- a/test/functional/query-builder/join/query-builder-joins.test.ts
+++ b/test/functional/query-builder/join/query-builder-joins.test.ts
@@ -1,5 +1,6 @@
 import "reflect-metadata"
 import { expect } from "chai"
+import sinon from "sinon"
 import {
     closeTestingConnections,
     createTestingConnections,
@@ -1717,6 +1718,81 @@ describe("query builder > joins", () => {
                     rows.map((row: { user_id: string }) => row.user_id),
                 )
                 expect(uniqueIds.size).to.equal(3)
+            }),
+        ))
+
+    it("should use distinct-id pagination when limit is used with leftJoinAndMapOne over a one-to-many relation", () =>
+        Promise.all(
+            dataSources.map(async (dataSource) => {
+                const manager = dataSource.manager
+
+                for (let i = 1; i <= 7; i++) {
+                    const user = new User()
+                    user.name = `User ${i}`
+                    await manager.save(user)
+
+                    for (let j = 1; j <= 2; j++) {
+                        const photo = new Photo()
+                        photo.name = `Photo ${i}-${j}`
+                        photo.user = user
+                        await manager.save(photo)
+                    }
+                }
+
+                const queryRunner = dataSource.createQueryRunner()
+                const querySpy = sinon.spy(queryRunner, "query")
+
+                try {
+                    const users = await manager
+                        .createQueryBuilder(User, "user")
+                        .leftJoinAndMapOne("user.photo", "user.photos", "photo")
+                        .orderBy("user.id")
+                        .limit(5)
+                        .setQueryRunner(queryRunner)
+                        .getMany()
+
+                    expect(users).to.have.lengthOf(5)
+                    expect(querySpy.callCount).to.equal(2)
+                } finally {
+                    querySpy.restore()
+                    await queryRunner.release()
+                }
+            }),
+        ))
+
+    it("github issues > #12445 should execute a single query when limit and offset are used with non-multiplying joins", () =>
+        Promise.all(
+            dataSources.map(async (dataSource) => {
+                const user = new User()
+                user.name = "User 1"
+                await dataSource.manager.save(user)
+
+                for (let i = 1; i <= 4; i++) {
+                    const photo = new Photo()
+                    photo.name = `Photo ${i}`
+                    photo.user = user
+                    await dataSource.manager.save(photo)
+                }
+
+                const queryRunner = dataSource.createQueryRunner()
+                const querySpy = sinon.spy(queryRunner, "query")
+
+                try {
+                    const photos = await dataSource.manager
+                        .createQueryBuilder(Photo, "photo")
+                        .innerJoinAndSelect("photo.user", "user")
+                        .orderBy("photo.id")
+                        .offset(1)
+                        .limit(2)
+                        .setQueryRunner(queryRunner)
+                        .getMany()
+
+                    expect(photos).to.have.lengthOf(2)
+                    expect(querySpy.callCount).to.equal(1)
+                } finally {
+                    querySpy.restore()
+                    await queryRunner.release()
+                }
             }),
         ))
 })
--- a/src/query-builder/SelectQueryBuilder.ts
+++ b/src/query-builder/SelectQueryBuilder.ts
@@ -1483,8 +1483,8 @@ export class SelectQueryBuilder<Entity extends ObjectLiteral>
 
     /**
      * Sets LIMIT - maximum number of rows to be selected.
-     * When joins are present, a two-query distinct-id strategy is used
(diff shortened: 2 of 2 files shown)


test/functional/query-builder/join/query-builder-joins.test.ts
```diff
@@ -1719,4 +1720,39 @@ describe("query builder > joins", () => {
...
+                        .createQueryBuilder(Photo, "photo")
+                        .innerJoinAndSelect("photo.user", "user")
+                        .orderBy("photo.id")
+                        .offset(1)
+                        .limit(2)
+                        .setQueryRunner(queryRunner)
+                        .getMany()
+
+                    expect(photos).to.have.lengthOf(2)
+                    expect(querySpy.callCount).to.equal(1)
```
gioboa, line 1791: @cduff I added this test to verify the fix.

cduff:
Thanks so much for addressing this issue so promptly @gioboa!

I'm worried that your current approach might continue to cause a regression in the expected behaviour of `offset`/`limit` though.

Consider the following query:

```ts
createQueryBuilder("parent")
  .leftJoinAndSelect("parent.children", "child")
  .where("parent.id = 1")
  .orderBy("child.id", "DESC")
  .limit(2)
  .getOne()
```

The expectation here would be to return the `parent` with `id === 1`, with `parent.children` populated with the top 2 most recently added children.

In other words, the expectation would be for `offset()` and `limit()` to map to the underlying `OFFSET` and `LIMIT` SQL clauses (in MySQL at least).

Instead, it looks like this PR continues to blur the distinction between `offset`/`limit` and `take`/`skip`, the latter of which understandably uses the "two-query distinct-id strategy".

Is the fix here to simply revert `offset`/`limit` to the previous behaviour?

(I'm not sure if there's then separate issues to resolve with `take`/`skip`.)

alumni (approved):

gioboa (commented):
> Is the fix here to simply revert `offset`/`limit` to the previous behaviour?

This PR intentionally keeps the #11987 / #3967 behaviour for multiplying joins, so limit() applies to root entities when a join can duplicate root rows. That means your example would not limit the hydrated children to 2. The #12445 regression is different: it involved a non-multiplying join like child.parent, where the two-query strategy adds overhead without changing result semantics. 
This PR solves that case by keeping non-multiplying joins on the single-query path, while preserving distinct-id pagination for one-to-many / many-to-many joins.

cduff:
> > Is the fix here to simply revert `offset`/`limit` to the previous behaviour?
> 
> This PR intentionally keeps the #11987 / #3967 behaviour for multiplying joins, so limit() applies to root entities when a join can duplicate root rows. That means your example would not limit the hydrated children to 2. The #12445 regression is different: it involved a non-multiplying join like child.parent, where the two-query strategy adds overhead without changing result semantics. This PR solves that case by keeping non-multiplying joins on the single-query path, while preserving distinct-id pagination for one-to-many / many-to-many joins.

Thanks for your work on this @gioboa.

Sorry but this doesn't make sense to me.

Looking at https://github.com/typeorm/typeorm/issues/3967, my interpretation is that they're after root entity pagination, which should be solvable using the [`take`/`skip` feature](https://typeorm.io/docs/query-builder/select-query-builder#using-pagination), as [commented](https://github.com/typeorm/typeorm/issues/3967#issuecomment-529489375) in that issue.

In my mind there's a clear and understandable difference between the `limit`/`offset` and `take`/`skip` features that would be broken by this change.

- `limit`/`offset` applies to the raw underlying result rows (equivalent to the `LIMIT`/`OFFSET` clauses in MySQL).
- `take`/`skip` applies to the resulting root entities.

The current code change would break the existing and expected behaviour of `limit()` and will change the results of queries like my [example above](https://github.com/typeorm/typeorm/pull/12447#issuecomment-4416991476), which may be depended on by users. It will also make such queries impossible to create using QueryBuilder.

alumni:
Let's forget about `.getOne()`/`.getMany()` etc.

`QueryBuilder` is a helper to build SQL queries. `.limit(10).offset(20)` should add `LIMIT 10 OFFSET 20` in the generated SQL, nothing more, _regardless of any joins_ or other query options. Whatever is the output of `.getOne()`/`.getMany()` in that case is irrelevant. If `.getSql()` contains that substring, we're good.

naorpeled (approved):

cduff:
> Let's forget about `.getOne()`/`.getMany()` etc.
> 
> `QueryBuilder` is a helper to build SQL queries. `.limit(10).offset(20)` should add `LIMIT 10 OFFSET 20` in the generated SQL, nothing more, _regardless of any joins_ or other query options. Whatever is the output of `.getOne()`/`.getMany()` in that case is irrelevant. If `.getSql()` contains that substring, we're good.

Agreed @alumni, but both 0.3.29 and this PR in its current state break that expectation.

Instead, they make [`.limit(10)`](https://typeorm.io/docs/query-builder/select-query-builder#adding-limit-expression) behave like [`.take(10)`](https://typeorm.io/docs/query-builder/select-query-builder#using-pagination).

I imagine this functional change being released in patch version 0.3.29 could currently be causing bugs in consuming systems.

FYI @gioboa.

cduff:
@alumni I'm confused - you seem to agree that this PR is currently incorrect, but you've also approved it?

gioboa (commented):
Let's revert #11987 I aim to have this issue solved soon.

PR #12448, opened 2026-05-10
fix: preserve postgres character column data when changing length

acidstorm2024-star:
Fixes #3357.

Postgres can safely alter character/varchar length changes in place with `ALTER TABLE ... ALTER COLUMN ... TYPE`, so `changeColumn` no longer drops and recreates character columns for length/type alias changes.

This avoids data loss for migrations such as `character varying(50)` -> `character varying(51)`.

Added a regression test that verifies existing data survives the length change and that the generated SQL uses `ALTER COLUMN TYPE` rather than `DROP COLUMN`.


smith-xyz:

PR #12449, opened 2026-05-10

ideook:
### Description of change

Fixes #3357 by handling PostgreSQL column length-only changes with `ALTER COLUMN ... TYPE` instead of dropping and recreating the column.

Previously, `PostgresQueryRunner.changeColumn()` treated a `length` change the same as an incompatible type change and recreated the column. For cases such as `character varying(50)` -> `character varying(51)`, that generated:

```sql
ALTER TABLE "bug" DROP COLUMN "example";
ALTER TABLE "bug" ADD "example" character varying(51);
```

which can cause data loss.

This change removes `length` from the destructive-recreate condition and lets the existing type alteration branch emit:

```sql
ALTER TABLE "bug" ALTER COLUMN "example" TYPE character varying(51);
```

A regression test was added for PostgreSQL varchar length changes.

### Pull-Request Checklist

- [x] `pnpm run compile` passes
- [x] `pnpm exec prettier --check src/driver/postgres/PostgresQueryRunner.ts test/github-issues/3357/issue-3357.test.ts` passes
- [ ] Tests pass locally with a PostgreSQL `ormconfig.json` configured

### DIFF
--- a/src/driver/postgres/PostgresQueryRunner.ts
+++ b/src/driver/postgres/PostgresQueryRunner.ts
@@ -2348,9 +2348,9 @@ export class PostgresQueryRunner
                     new Query(
                         `ALTER TABLE ${this.escapePath(table)} ALTER COLUMN "${
                             newColumn.name
-                        }" TYPE ${newColumn.type} COLLATE "${
-                            newColumn.collation
-                        }"`,
+                        }" TYPE ${this.driver.createFullType(
+                            newColumn,
+                        )} COLLATE "${newColumn.collation}"`,
                     ),
                 )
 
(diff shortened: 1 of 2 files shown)


ideook:
Follow-up for the Qodo review:

- Updated Postgres collation ALTER statements to use `createFullType(...)`, so combined length/precision/scale + collation changes preserve type modifiers instead of widening to the bare type.
- Added a regression assertion for length + collation changes.
- Wrapped the test QueryRunner usage in `try/finally` so `release()` runs even if the test fails.

Re-verified with:
- `pnpm exec prettier --check src/driver/postgres/PostgresQueryRunner.ts test/github-issues/3357/issue-3357.test.ts`
- `pnpm run compile`

smith-xyz:

PR #12450, opened 2026-05-11
fix: preserve postgres varchar data on length change

dairon2:
### Description of change

Fixes #3357 by handling PostgreSQL varchar length changes with `ALTER COLUMN ... TYPE` instead of dropping and recreating the column.

Previously, `PostgresQueryRunner.changeColumn()` treated a `length` change as a destructive type incompatibility, so changing `varchar(50)` to `varchar(51)` generated a `DROP COLUMN` / `ADD COLUMN` sequence and lost existing data.

This change:

- removes `length` from the destructive recreate condition
- routes length-only changes through the existing non-destructive `ALTER COLUMN ... TYPE` path
- preserves type modifiers when altering column collation by using `createFullType()` instead of the bare column type
- adds a functional regression test that inserts data, changes a Postgres varchar length, synchronizes, and verifies the data is preserved
- updates the collation functional test expectations so the generated SQL preserves `character varying(100)`

### Pull-Request Checklist

- [x] `corepack pnpm exec prettier --check src/driver/postgres/PostgresQueryRunner.ts test/functional/schema-builder/column/change/change-column/change-column.test.ts test/functional/schema-builder/collation/collation-changes/collation-changes.test.ts` passes
- [x] `corepack pnpm run typecheck` passes
- [x] `corepack pnpm run compile` passes
- [ ] Functional test run requires a local `ormconfig.json`/Postgres configuration; attempted targeted run and it stopped at the missing `ormconfig.json` project setup requirement

### DIFF
--- a/src/driver/postgres/PostgresQueryRunner.ts
+++ b/src/driver/postgres/PostgresQueryRunner.ts
@@ -1328,7 +1328,6 @@ export class PostgresQueryRunner
 
         if (
             oldColumn.type !== newColumn.type ||
-            oldColumn.length !== newColumn.length ||
             newColumn.isArray !== oldColumn.isArray ||
             (!oldColumn.generatedType &&
                 newColumn.generatedType === "STORED") ||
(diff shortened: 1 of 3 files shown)


smith-xyz:

PR #12452, opened 2026-05-11
fix: alter postgres varchar length without column recreation

hardikdhingra150:
Fixes #3357.

## Summary

Postgres treated varchar length changes as destructive type changes, so generated migrations dropped and recreated columns. This preserves existing data by emitting `ALTER COLUMN ... TYPE` for length-only changes instead.

Also preserves the full column type when applying Postgres collation changes, so a `varchar(100)` column does not become unbounded `varchar` during `ALTER COLUMN ... TYPE ... COLLATE`.

## Tests

- `pnpm run typecheck`
- `pnpm exec prettier --check src/driver/postgres/PostgresQueryRunner.ts test/functional/schema-builder/column/change/change-column/change-column.test.ts test/functional/schema-builder/collation/collation-changes/collation-changes.test.ts`

Note: the focused Postgres integration path requires TypeORM's `ormconfig.json` and a configured Postgres test database for local execution.

### DIFF
--- a/src/driver/postgres/PostgresQueryRunner.ts
+++ b/src/driver/postgres/PostgresQueryRunner.ts
@@ -1328,7 +1328,6 @@ export class PostgresQueryRunner
 
         if (
             oldColumn.type !== newColumn.type ||
-            oldColumn.length !== newColumn.length ||
             newColumn.isArray !== oldColumn.isArray ||
             (!oldColumn.generatedType &&
                 newColumn.generatedType === "STORED") ||
@@ -1619,6 +1618,7 @@ export class PostgresQueryRunner
             }
 
             if (
+                newColumn.length !== oldColumn.length ||
                 newColumn.precision !== oldColumn.precision ||
                 newColumn.scale !== oldColumn.scale
             ) {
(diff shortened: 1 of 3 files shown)


hardikdhingra150:
Moved the regression coverage from `test/github-issues` into the existing functional schema-builder column-change suite, per the project checklist.

Re-ran:

- `pnpm run typecheck`
- `pnpm exec prettier --check src/driver/postgres/PostgresQueryRunner.ts test/functional/schema-builder/column/change/change-column/change-column.test.ts`

hardikdhingra150:
Addressed the latest Qodo findings in ae5859464:

- Collation changes now use `createFullType(...)` for both up/down SQL, preserving varchar length while applying `COLLATE`.
- The #3357 regression test now restores the original metadata length instead of resetting to an empty value.
- Updated the existing Postgres collation functional test to assert `character varying(100) COLLATE ...`, covering the length-preservation case.

Re-ran:

- `pnpm run typecheck`
- `pnpm exec prettier --check src/driver/postgres/PostgresQueryRunner.ts test/functional/schema-builder/column/change/change-column/change-column.test.ts test/functional/schema-builder/collation/collation-changes/collation-changes.test.ts`

smith-xyz:

PR #12453, opened 2026-05-11
fix: preserve explicit join column type

### LINKED ISSUE #12440: Column type is overwritten when JoinColumn references a column with another type (mssql)
### Issue description

The actual column type definition of an entity is ignored when it is also declared as a JoinColumn that references a column with a different type

### Expected Behavior

Whatever a Column's defined type is, for reasons of predictability, in my opinion it should always be this type when passed as a parameter.
```
@PrimaryGeneratedColumn()
public id: number;
```
Should always be an int, no matter what else happens in the entity. 

### Actual Behavior

Given an entity definition that has 
```
@PrimaryGeneratedColumn()
public id: number;
```
but also a one to one relationship where the JoinColumn references a column with a different type:

```
@OneToOne((type) => ProductItemSubscription)
@JoinColumn({name: "id", referencedColumnName: "linkId"})
public _subscription: ProductItemSubscription
```

Now, when the `id` is passed as a parameter in a `findOne` query, it is passed **as the type of `linkId`**. If as in this case `id` is a number, but `linkId` in ProductItemSubscription is a varchar, the query tries to pass the number as a varchar and fails with an error "Validation failed for parameter '0'. Invalid string.". The parameters it passes are 

```
parameters: [
    MssqlParameter {
      value: 2212,
      type: 'varchar',
      '@instanceof': Symbol(MssqlParameter),
      params: []
    }
  ],
```

Which shows the issue exactly.

I'm not trying to argue that this is a good pattern to use, in fact I think it definitely smells. However, I *do* think it is a bug that TypeORM takes the type of the referenced column of another entity over the definition in the entity itself. 

### Steps to reproduce

```
@Entity()
export class Vehicle {
	@PrimaryGeneratedColumn()
	public id: number;
    @OneToOne((type) => ProductItemSubscription)
    @JoinColumn({name: "id", referencedColumnName: "linkId"})
    public _subscription: ProductItemSubscription
}

@Entity()
export class ProductItemSubscription {
    @PrimaryGeneratedColumn()
	public id: number;
    @Column("varchar", { name: "linkId" , length: 24})
	linkId: string;
}
```

### My Environment

| Operating System    | macOS 26.4|
| Node.js version     | 24.14.1  |
| Typescript version  | 6.0  |

### Additional Context

_No response_

- [ ] aurora-mysql
- [ ] aurora-postgres
- [ ] better-sqlite3
- [ ] capacitor
- [ ] cockroachdb
- [ ] cordova
- [ ] expo
- [ ] mongodb
- [ ] mysql
- [ ] nativescript
- [ ] oracle
- [ ] postgres
- [ ] react-native
- [ ] sap
- [ ] spanner
- [ ] sqlite
- [ ] sqljs
- [x] sqlserver

hardikdhingra150:
Fixes #12440.

## Summary

When a user-defined local column is also used as a join column, keep that local column's declared type instead of overwriting it with the referenced column type.

This prevents MSSQL parameter wrapping from treating a numeric local primary column as `varchar` when the relation references a `varchar` column on the target entity.

When the local and referenced join column types are incompatible, TypeORM now skips FK metadata creation for that relation while preserving join columns for query building.

## Tests

- `pnpm run typecheck`
- `pnpm run compile`
- `pnpm exec prettier --check src/metadata-builder/RelationJoinColumnBuilder.ts test/functional/relations/join-column-type-precedence/join-column-type-precedence.test.ts`
- `pnpm exec mocha --no-config build/compiled/test/functional/relations/join-column-type-precedence/join-column-type-precedence.test.js`

### DIFF
--- a/src/metadata-builder/RelationJoinColumnBuilder.ts
+++ b/src/metadata-builder/RelationJoinColumnBuilder.ts
@@ -83,6 +83,12 @@ export class RelationJoinColumnBuilder {
                 columns,
                 uniqueConstraint: undefined,
             } // this case is possible for one-to-one non owning side and relations with createForeignKeyConstraints = false
+        if (!this.columnsHaveCompatibleTypes(columns, referencedColumns))
+            return {
+                foreignKey: undefined,
+                columns,
+                uniqueConstraint: undefined,
+            }
 
         const foreignKey = new ForeignKeyMetadata({
             name: joinColumns[0]?.foreignKeyConstraintName,
@@ -125,6 +131,21 @@ export class RelationJoinColumnBuilder {
     // -------------------------------------------------------------------------
     // Protected Methods
     // -------------------------------------------------------------------------
+
+    private columnsHaveCompatibleTypes(
+        columns: ColumnMetadata[],
+        referencedColumns: ColumnMetadata[],
+    ): boolean {
+        return columns.every((column, index) => {
+            const referencedColumn = referencedColumns[index]
+            if (!referencedColumn) return false
+
+            return (
+                this.dataSource.driver.normalizeType(column) ===
+                this.dataSource.driver.normalizeType(referencedColumn)
+            )
+        })
+    }
 
     /**
      * Collects referenced columns from the given join column args.
@@ -270,7 +291,9 @@ export class RelationJoinColumnBuilder {
                 relationalColumn = OrmUtils.cloneObject(relationalColumn)
             }
             relationalColumn.referencedColumn = referencedColumn // its important to set it here because we need to set referenced column for user defined join column
-            relationalColumn.type = referencedColumn.type // also since types of relational column and join column must be equal we override user defined column type
+            if (relationalColumn.isVirtual) {
+                relationalColumn.type = referencedColumn.type
+            }
             relationalColumn.relationMetadata = relation
             relationalColumn.build(this.dataSource)
             return relationalColumn
--- a/test/functional/relations/join-column-type-precedence/join-column-type-precedence.test.ts
+++ b/test/functional/relations/join-column-type-precedence/join-column-type-precedence.test.ts
@@ -1,0 +1,81 @@
+import "reflect-metadata"
+import { expect } from "chai"
+import {
+    Column,
+    DataSource,
+    Entity,
+    JoinColumn,
+    OneToOne,
+    PrimaryGeneratedColumn,
+} from "../../../../src"
+import { MssqlParameter } from "../../../../src/driver/sqlserver/MssqlParameter"
+import { SqlServerDriver } from "../../../../src/driver/sqlserver/SqlServerDriver"
+
+describe("relations > join column type precedence", () => {
+    let dataSource: MetadataOnlyDataSource | undefined
+
+    afterEach(async () => {
+        if (dataSource?.isInitialized) {
+            await dataSource.destroy()
+        }
+        dataSource = undefined
+    })
+
+    it("should preserve the explicit local column type when a join column references a different type (#12440)", async () => {
+        dataSource = new MetadataOnlyDataSource({
+            type: "mssql",
+            host: "localhost",
+            username: "test",
+            password: "test",
+            database: "test",
+            entities: [Vehicle, ProductItemSubscription],
+        })
+
+        await dataSource.buildMetadatasForTest()
+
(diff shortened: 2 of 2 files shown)


hardikdhingra150:
Addressed the latest Qodo findings in 21ea5c7c5:

- Removed the `(dataSource as any).buildMetadatas()` call by using a small test-only subclass that exposes the protected metadata build path without an `any` cast.
- Added normalized join/referenced column type compatibility checking before `ForeignKeyMetadata` creation.
- For divergent user-defined join column types, TypeORM now keeps the join columns usable for query building but skips FK metadata/DDL generation.
- Extended the functional regression test to assert no relation/entity FK metadata is produced for the divergent `int` -> `varchar` case.

Re-ran:

- `pnpm run typecheck`
- `pnpm run compile`
- `pnpm exec prettier --check src/metadata-builder/RelationJoinColumnBuilder.ts test/functional/relations/join-column-type-precedence/join-column-type-precedence.test.ts`
- `pnpm exec mocha --no-config build/compiled/test/functional/relations/join-column-type-precedence/join-column-type-precedence.test.js`

test/functional/relations/join-column-type-precedence/join-column-type-precedence.test.ts
```diff
@@ -0,0 +1,81 @@
...
+    DataSource,
+    Entity,
+    JoinColumn,
+    OneToOne,
+    PrimaryGeneratedColumn,
+} from "../../../../src"
+import { MssqlParameter } from "../../../../src/driver/sqlserver/MssqlParameter"
+import { SqlServerDriver } from "../../../../src/driver/sqlserver/SqlServerDriver"
+
+describe("relations > join column type precedence", () => {
```
smith-xyz, line 14: can we add a test for checking that it doesn't silently drop existing FK on sync?

```typescript
it("does not silently drop existing FK on synchronize for incompatible join mapping", async () => {
  // Arrange schema with an existing FK for relation R
  // Arrange entities so local join type differs from referenced type
  // Act: initialize with synchronize: true
  // Assert: explicit behavior contract
  // - either throws a metadata error with actionable message
  // - or requires explicit opt-in and logs warning before skipping FK
  // Assert no unexpected FK drop when contract says preserve
})
```

src/metadata-builder/RelationJoinColumnBuilder.ts
```diff
@@ -83,6 +83,12 @@ export class RelationJoinColumnBuilder {
                 columns,
                 uniqueConstraint: undefined,
             } // this case is possible for one-to-one non owning side and relations with createForeignKeyConstraints = false
+        if (!this.columnsHaveCompatibleTypes(columns, referencedColumns))
+            return {
+                foreignKey: undefined,
```
smith-xyz, line 88: Wondering if we should return an explicit metadata error on incompatible types or gate this fallback behind an explicit opt-in relation option and warning.

Right now this seems to silently happen and I think there could be unintended consequences (see proposed test we could add)

src/metadata-builder/RelationJoinColumnBuilder.ts
```diff
@@ -126,6 +132,21 @@ export class RelationJoinColumnBuilder {
...
+    private columnsHaveCompatibleTypes(
+        columns: ColumnMetadata[],
+        referencedColumns: ColumnMetadata[],
+    ): boolean {
+        return columns.every((column, index) => {
+            const referencedColumn = referencedColumns[index]
+            if (!referencedColumn) return false
+
+            return (
+                this.dataSource.driver.normalizeType(column) ===
```
smith-xyz, line 144: probably should verify if normalizeType name is sufficient here? Worried about driver compatibility. Equality on normalized type strings may be too coarse and can still permit invalid pairs or reject valid ones inconsistently by driver.

Also wondering if it could just be this:

```typescript
private columnsHaveCompatibleTypes(
    columns: ColumnMetadata[],
    referencedColumns: ColumnMetadata[],
): boolean {
  if (columns.length !== referencedColumns.length) return false

  return columns.every(
    (column, index) =>
      this.dataSource.driver.normalizeType(column) ===
      this.dataSource.driver.normalizeType(referencedColumns[index]!),
  )
}
```

smith-xyz (changes requested):

michaelbromley:

```
git fetch origin
```

PR #12456, opened 2026-05-11
fix(postgres): alter column length without dropping column

JiaxinLi-lipluszn:
Fixes #3357.

Postgres column length changes now use `ALTER COLUMN ... TYPE` instead of dropping and recreating the column. This prevents data loss when changing a column length, for example `varchar(50)` to `varchar(51)`.

Added a regression test that verifies the generated migration SQL and confirms existing data remains after applying the change.

Verification run locally:

```bash
git diff --check
pnpm exec prettier --write src/driver/postgres/PostgresQueryRunner.ts test/github-issues/3357/issue-3357.test.ts
pnpm exec tsc --noEmit --pretty false
pnpm run compile
pnpm exec eslint src/driver/postgres/PostgresQueryRunner.ts test/github-issues/3357/issue-3357.test.ts
```

The Postgres-specific test was not run locally because this environment does not have an `ormconfig.json` or local Postgres test database configured.


JiaxinLi-LiGaZn:
Addressed the Qodo maintainability note by moving the regression coverage from `test/github-issues/3357` into the functional schema-builder column-change suite at `test/functional/schema-builder/column/change/change-column/change-column.test.ts`.

smith-xyz:

PR #12457, opened 2026-05-11
fix(postgres): alter column length without dropping column

hemantrajputt:
Fixes #3357.

### What changed

Postgres column length changes now use `ALTER COLUMN ... TYPE` instead of dropping and re-adding the column.

This avoids data loss when changing lengths such as `varchar(50)` to `varchar(51)`.


smith-xyz:

PR #12458, opened 2026-05-11
fix(postgres): preserve varchar data when changing length

raunaqsingh:
Fixes #3357.

### What changed
- Treat Postgres `varchar` / `character varying` length-only changes as non-destructive schema changes.
- Generate `ALTER TABLE ... ALTER COLUMN ... TYPE` for those length changes instead of dropping and recreating the column.
- Add a regression test covering `varchar(50)` to `varchar(51)` while preserving existing data.

### Testing
- Not run locally in this session.


smith-xyz:

PR #12459, opened 2026-05-11
fix: alter Postgres varchar length without dropping column

sunshadow7:
## Summary

Fixes #3357.

When a Postgres column length changed, `PostgresQueryRunner.changeColumn` treated the change as requiring full column recreation. That generated `DROP COLUMN` and `ADD COLUMN`, which loses existing data during migration generation/synchronization.

This change keeps length-only changes inside the alter-column path and emits `ALTER TABLE ... ALTER COLUMN ... TYPE ...` using the driver's full type rendering. It also updates collation type changes to use full type rendering so length modifiers are preserved when length and collation change together.

Regression coverage is in the functional schema-builder tests:

- `schema-builder > change column` checks that a Postgres varchar length change emits `ALTER COLUMN ... TYPE`, does not emit `DROP COLUMN`/`ADD`, and preserves an existing row.
- `schema-builder > collation > collation changes` checks that combined length plus collation changes emit a single full-type `ALTER COLUMN ... TYPE ... COLLATE` query.

## Tests

- `corepack pnpm run compile`
- `corepack pnpm exec lint-staged`
- Direct compiled Postgres query-runner SQL checks:
  - length only: `ALTER TABLE "bug" ALTER COLUMN "example" TYPE character varying(51)`
  - length + collation: `ALTER TABLE "bug" ALTER COLUMN "example" TYPE character varying(51) COLLATE "C"`

I could not run the database-backed functional tests locally because this checkout does not have an `ormconfig.json` with Postgres credentials.


smith-xyz:

PR #12460, opened 2026-05-11
fix(postgres): use ALTER COLUMN TYPE instead of DROP/ADD when column length changes

tugayakin:
## Problem

Fixes #3357

When a column's **length** changes (e.g. `varchar(50)` → `varchar(51)`), TypeORM's PostgreSQL migration generator was producing a destructive sequence:

```sql
ALTER TABLE "example" DROP COLUMN "field"
ALTER TABLE "example" ADD "field" character varying(51)
```

This silently **destroys all data** in that column.

## Root Cause

In `PostgresQueryRunner.changeColumn`, the condition for the DROP+ADD path included `oldColumn.length !== newColumn.length`. A length-only change does not require column recreation — PostgreSQL supports `ALTER COLUMN TYPE` for this.

## Fix

- Removed `oldColumn.length !== newColumn.length` from the DROP+ADD condition
- Added a dedicated block to handle length changes using `ALTER COLUMN TYPE`, following the same pattern already used for `precision`/`scale` changes

Now the generated migration for a length-only change is:

```sql
ALTER TABLE "example" ALTER COLUMN "field" TYPE character varying(51)
```

This preserves all existing data.

## Testing

The fix mirrors the existing handling for `precision`/`scale` changes, which uses the same `ALTER COLUMN TYPE` approach safely.

### DIFF
--- a/src/driver/postgres/PostgresQueryRunner.ts
+++ b/src/driver/postgres/PostgresQueryRunner.ts
@@ -1619,25 +1619,9 @@ export class PostgresQueryRunner
 
             if (
                 newColumn.precision !== oldColumn.precision ||
-                newColumn.scale !== oldColumn.scale
+                newColumn.scale !== oldColumn.scale ||
+                oldColumn.length !== newColumn.length
             ) {
-                upQueries.push(
-                    new Query(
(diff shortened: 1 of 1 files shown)


smith-xyz:

PR #12461, opened 2026-05-11
fix: alter postgres column length without recreating column

zenith-atomic:
### Description of change

Fixes Postgres column length changes so TypeORM emits an `ALTER COLUMN ... TYPE` query instead of dropping and re-adding the column. Recreating the column can cause data loss when only the varchar length changes.

Closes #3357.

### Pull-Request Checklist

- [x] `npm run format` for changed code — not run; change follows existing formatting
- [x] `npm run test` — focused regression test run instead

### Verification

- `pnpm run typecheck`
- `pnpm run compile`
- `node ./node_modules/mocha/bin/mocha.js --no-config build/compiled/test/github-issues/3357/issue-3357.test.js --exit`

The regression test verifies that changing a Postgres `character varying(50)` column to `character varying(51)`:

- emits `ALTER TABLE "bug" ALTER COLUMN "example" TYPE character varying(51)`
- does not emit `DROP COLUMN`
- does not emit `ADD`
- preserves existing row data

### DIFF
--- a/src/driver/postgres/PostgresQueryRunner.ts
+++ b/src/driver/postgres/PostgresQueryRunner.ts
@@ -1328,7 +1328,6 @@ export class PostgresQueryRunner
 
         if (
             oldColumn.type !== newColumn.type ||
-            oldColumn.length !== newColumn.length ||
             newColumn.isArray !== oldColumn.isArray ||
             (!oldColumn.generatedType &&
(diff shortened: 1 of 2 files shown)


zenith-atomic:
Closing this after seeing the maintainer note on #3357 that a fix is already under active review and new PRs should not be opened. Apologies for the duplicate noise.

PR #12462, opened 2026-05-12

ZainKazmiii:
### Description of change

Fixes #3357.

Postgres length-only changes for `character varying`, `varchar`, `character`, and `char` columns now use `ALTER TABLE ... ALTER COLUMN ... TYPE ...` instead of dropping and recreating the column. This avoids data loss when generating migrations or synchronizing schema changes that only alter the column length.

The updated functional test now inserts data before changing column lengths, checks the generated SQL does not include `DROP COLUMN`/`ADD`, synchronizes the schema, and verifies the stored values are preserved.

This also uses `createFullType()` for collation type changes so length/precision details are kept when the type expression is rebuilt.

Verification:

- `pnpm exec prettier --check src/driver/postgres/PostgresQueryRunner.ts test/functional/database-schema/column-length/postgres/column-length-postgres.test.ts`
- `pnpm run compile`
- `pnpm exec lint-staged`
- Local no-database sanity check against `PostgresQueryRunner.changeColumn()` confirmed the generated SQL uses `ALTER COLUMN ... TYPE` and does not include `DROP COLUMN` or `ADD` for the affected Postgres string types.

I could not run the Postgres functional test locally because this environment does not have Docker/Postgres installed.

### Pull-Request Checklist

- [x] Documentation has been updated to reflect this change (`docs/docs/**.md`) - N/A, bug fix with no documentation-facing behavior change

### DIFF
--- a/src/driver/postgres/PostgresQueryRunner.ts
+++ b/src/driver/postgres/PostgresQueryRunner.ts
@@ -1326,9 +1326,12 @@ export class PostgresQueryRunner
                 `Column "${oldTableColumnOrName}" was not found in the "${table.name}" table.`,
             )
 
+        const lengthChanged = oldColumn.length !== newColumn.length
+        const shouldAlterLength = this.isSafeLengthChange(oldColumn, newColumn)
+
         if (
             oldColumn.type !== newColumn.type ||
-            oldColumn.length !== newColumn.length ||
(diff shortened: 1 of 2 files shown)


Cprakhar:
Closing as a duplicate — a fix for this issue is already under the maintainers' review. Thank you for the contribution.

PR #12463, opened 2026-05-12
fix: alter postgres column length in place

didier-brown:
# Fix Postgres column length migrations without dropping data

Fixes #3357.

## Summary

- stop treating same-type Postgres column length changes as a destructive drop/add
- emit `ALTER TABLE ... ALTER COLUMN ... TYPE ...` for length/precision/scale changes
- add a regression test that verifies changing a `varchar` length preserves existing row data
- assert the generated schema SQL uses `ALTER COLUMN "name" TYPE character varying(500)` and does not include `DROP COLUMN "name"` or `ADD "name"`

## Notes

This intentionally changes the behavior for same-type length changes from recreating the column to altering it in place. If a down migration narrows a column and existing values exceed the new length, Postgres may reject that migration rather than silently losing data.

## Verification

- `git diff --check`

I could not run the full TypeORM test suite in this local environment because `pnpm`/`npm` and Docker/Postgres are not installed.

### DIFF
--- a/src/driver/postgres/PostgresQueryRunner.ts
+++ b/src/driver/postgres/PostgresQueryRunner.ts
@@ -1328,6 +1328,5255 @@ export class PostgresQueryRunner
 
         if (
             oldColumn.type !== newColumn.type ||
+            newColumn.isArray !== oldColumn.isArray ||
+            (!oldColumn.generatedType &&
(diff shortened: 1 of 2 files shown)


Cprakhar:
Closing as a duplicate — a fix for this issue is already under the maintainers' review. Thank you for the contribution.

PR #12466, opened 2026-05-12

BrodoB:
Description of change

Fixes #3357

This changes PostgreSQL column migration generation so a `varchar` / `character varying` length-only change uses `ALTER COLUMN ... TYPE` instead of dropping and re-adding the column.

Previously, changing a column from `varchar(50)` to `varchar(51)` went through the drop/add path, which can fail for non-null columns and risks losing existing data. The new behavior preserves the column and its data.

Verification:
- `corepack pnpm run compile`
- `corepack pnpm run test:fast -- --grep "#3357"`
- `corepack pnpm run typecheck`
- `git diff --check`

Pull-Request Checklist

- Code is up-to-date with the `master` branch
- This pull request links relevant issues as `Fixes #3357`
- There are new or updated tests validating the change (`tests/**.test.ts`)
- Documentation has been updated to reflect this change (`docs/docs/**.md`) N/A

### DIFF
--- a/src/driver/postgres/PostgresQueryRunner.ts
+++ b/src/driver/postgres/PostgresQueryRunner.ts
@@ -1326,9 +1326,17 @@ export class PostgresQueryRunner
                 `Column "${oldTableColumnOrName}" was not found in the "${table.name}" table.`,
             )
 
+        const isVarcharLengthChange =
+            oldColumn.length !== newColumn.length &&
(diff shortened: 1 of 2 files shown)


BrodoB:
Sorry, I missed that #11620 is already under active review for this issue. Closing this to avoid adding noise.

PR #12468, opened 2026-05-12
fix: avoid recreating Postgres varchar columns on length changes

neoalabs:
## Summary

Fixes Postgres schema diff generation for varchar length-only changes so TypeORM emits `ALTER TABLE ... ALTER COLUMN ... TYPE ...` instead of dropping and recreating the column.

## Root cause

`PostgresQueryRunner.changeColumn()` treated `oldColumn.length !== newColumn.length` as a reason to recreate the column. For cases such as `character varying(50)` to `character varying(51)`, that generated destructive `DROP COLUMN` / `ADD COLUMN` queries and risked data loss.

## Changes

- Let type changes still recreate the column where needed.
- Handle length changes through the existing `ALTER COLUMN ... TYPE ...` path alongside precision and scale changes.
- Add a regression test for issue #3357 that asserts Postgres varchar length changes generate non-destructive SQL.

## Checks

- `pnpm run compile`
- `pnpm exec mocha --no-config --exit build/compiled/test/github-issues/3357/issue-3357.test.js`
- `pnpm exec prettier --check src/driver/postgres/PostgresQueryRunner.ts test/github-issues/3357/issue-3357.test.ts`

Fixes #3357

### DIFF
--- a/src/driver/postgres/PostgresQueryRunner.ts
+++ b/src/driver/postgres/PostgresQueryRunner.ts
@@ -1328,7 +1328,6 @@ export class PostgresQueryRunner
 
         if (
             oldColumn.type !== newColumn.type ||
-            oldColumn.length !== newColumn.length ||
             newColumn.isArray !== oldColumn.isArray ||
             (!oldColumn.generatedType &&
                 newColumn.generatedType === "STORED") ||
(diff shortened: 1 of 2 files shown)


smith-xyz:

PR #12470, opened 2026-05-13

MasterMao:
### What changed

- Treat PostgreSQL column length changes like precision/scale changes, generating `ALTER COLUMN ... TYPE ...` instead of dropping and re-adding the column.
- Added a regression unit test for changing `character varying(50)` to `character varying(51)`.

Fixes #3357

/claim #3357

### Verification

- `corepack pnpm install --frozen-lockfile`
- `corepack pnpm run compile`
- `.\node_modules\.bin\mocha.cmd --no-config build\compiled\test\unit\driver\postgres-query-runner-change-column.test.js`

Prepared with Codex assistance and reviewed before submission.


smith-xyz:

PR #12471, opened 2026-05-13
fix: preserve postgres varchar data when changing length

GaAs9000:
Fixes #3357.

### Summary
- avoid recreating Postgres varchar/character varying columns when only the length changes
- normalize Postgres column types before destructive change detection so `varchar` and `character varying` aliases are treated equivalently
- emit `ALTER COLUMN ... TYPE` for varchar length changes instead
- keep the query runner table cache in sync after the non-destructive alter
- add a functional regression test covering generated SQL and preserving existing data for the `varchar` alias case

### Tests
- `pnpm run typecheck`
- `pnpm exec prettier --check src/driver/postgres/PostgresQueryRunner.ts test/functional/query-runner/change-column.test.ts`
- `pnpm exec eslint src/driver/postgres/PostgresQueryRunner.ts test/functional/query-runner/change-column.test.ts`
- `pnpm run compile`
- `pnpm exec mocha --no-config --require ./build/compiled/test/utils/test-setup.js build/compiled/test/functional/query-runner/change-column.test.js --grep "postgres varchar length" --timeout 90000 --exit`

### DIFF
--- a/src/driver/postgres/PostgresQueryRunner.ts
+++ b/src/driver/postgres/PostgresQueryRunner.ts
@@ -1326,9 +1326,21 @@ export class PostgresQueryRunner
                 `Column "${oldTableColumnOrName}" was not found in the "${table.name}" table.`,
             )
 
+        const oldColumnType = this.driver.normalizeType({
+            type: oldColumn.type as ColumnType,
+        })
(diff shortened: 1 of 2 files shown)


smith-xyz:

PR #12475, opened 2026-05-13
fix: alter postgres column length without dropping data

GaneshVG18:
## Summary

Fixes #3357.

This changes Postgres column length migrations to alter the existing column type instead of dropping and recreating the column when only the length changes. PostgreSQL supports this through `ALTER TABLE ... ALTER COLUMN ... TYPE`, and using drop/add can lose existing data.

## Changes

- Avoid recreating Postgres columns when only `length` changes.
- Emit `ALTER TABLE ... ALTER COLUMN ... TYPE` for same-type length updates.
- Add regression coverage for varchar/character varying length updates.
- Assert the generated SQL does not contain `DROP COLUMN` / `ADD COLUMN` for the varchar case.
- Assert existing row data survives `synchronize(false)`.

## Testing

Not run in my local CLI environment because dependency install was blocked by DNS/network and Docker was unavailable.

Suggested validation:
- `pnpm install --frozen-lockfile`
- `jq 'map(select(.type == "postgres"))' ormconfig.sample.json > ormconfig.json`
- `pnpm run compile`
- `pnpm run test:fast -- ./build/compiled/test/functional/database-schema/column-length/postgres/column-length-postgres.test.js`

### DIFF
--- a/src/driver/postgres/PostgresQueryRunner.ts
+++ b/src/driver/postgres/PostgresQueryRunner.ts
@@ -1326,14 +1326,15 @@ export class PostgresQueryRunner
                 `Column "${oldTableColumnOrName}" was not found in the "${table.name}" table.`,
             )
 
-        if (
-            oldColumn.type !== newColumn.type ||
-            oldColumn.length !== newColumn.length ||
-            newColumn.isArray !== oldColumn.isArray ||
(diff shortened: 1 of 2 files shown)


smith-xyz:

PR #12476, opened 2026-05-13
fix: preserve postgres varchar data on length changes

tngus6007:
/claim #3357

## Summary

- Avoid recreating PostgreSQL columns when only the length changes for length-capable column types such as `varchar` / `character varying`.
- Reuse the existing `ALTER COLUMN ... TYPE ...` path for length-only changes, preserving table data instead of generating a drop/add column sequence.
- Add regression coverage to ensure a PostgreSQL length change keeps existing row values.

## Why

PostgreSQL supports changing `varchar(n)` length in place. TypeORM currently treats a length change like a type change in `PostgresQueryRunner.changeColumn()`, which drops and recreates the column. That can produce data-loss migrations for cases like `varchar(50)` to `varchar(51)`.

## Validation

- `npm run typecheck`
- `npm run lint -- src\driver\postgres\PostgresQueryRunner.ts test\functional\schema-builder\column\change\change-column\change-column.test.ts`
- `git diff --check`

Note: I could not run the PostgreSQL integration test locally because this environment does not have a configured PostgreSQL test database. The lint run reports existing warnings in `PostgresQueryRunner.ts` but no errors.


smith-xyz:

tngus6007:
Addressed the Qodo review feedback in `a8f8278c2`:

- Collation changes now use `createFullType()` for both up/down queries so a length/precision/scale change is not dropped by a subsequent `ALTER COLUMN ... COLLATE` statement.
- Added a `#3357` regression note near the PostgreSQL data-preservation assertion.
- Added a changelog entry for the PostgreSQL `varchar` length migration behavior.

Validation:

- `npm run typecheck`
- `npm run compile`
- `npm run lint -- src\driver\postgres\PostgresQueryRunner.ts test\functional\schema-builder\column\change\change-column\change-column.test.ts`
- `git diff --check`

Lint still reports existing warnings in `PostgresQueryRunner.ts`, but no errors.

PR #12485, opened 2026-05-13
fix: alter postgres varchar length without column recreation

danieldejesuslopez:
Closes #3357.

## Summary
- changes Postgres column-length updates to use `ALTER COLUMN ... TYPE` instead of dropping/re-adding the column
- keeps type/array/generated-column changes on the existing recreation path
- adds a regression test for varchar length changes that asserts no `DROP COLUMN` is generated and existing row data is preserved

## Verification
- `pnpm install --frozen-lockfile`
- `pnpm exec prettier --check src\driver\postgres\PostgresQueryRunner.ts test\github-issues\3357\issue-3357.test.ts test\github-issues\3357\entity\Bug.ts`
- `pnpm exec tsc --noEmit --pretty false`
- `pnpm run compile`

I could not run the Postgres integration test locally because this checkout does not have a valid `ormconfig.json` with Postgres credentials.

### DIFF
--- a/src/driver/postgres/PostgresQueryRunner.ts
+++ b/src/driver/postgres/PostgresQueryRunner.ts
@@ -1328,7 +1328,6 @@ export class PostgresQueryRunner
 
         if (
             oldColumn.type !== newColumn.type ||
-            oldColumn.length !== newColumn.length ||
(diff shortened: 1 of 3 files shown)


smith-xyz:

PR #12486, opened 2026-05-14
fix(postgres): preserve data when varchar length changes

inick89:
Fixes #3357.

Postgres schema sync previously treated a column length change like a type change and recreated the column with DROP COLUMN + ADD COLUMN. That loses existing data for cases such as varchar(50) -> varchar(51).

This change keeps destructive recreate behavior for actual type/array/generated-column changes, but routes length-only changes through ALTER COLUMN TYPE, alongside existing precision/scale changes.

Verification:
- corepack pnpm run compile
- node .\node_modules\mocha\bin\mocha.js --no-config build\compiled\test\unit\query-runner\postgres-change-column.test.js
- corepack pnpm exec prettier --check src/driver/postgres/PostgresQueryRunner.ts test/unit/query-runner/postgres-change-column.test.ts test/functional/database-schema/column-length/postgres/column-length-postgres.test.ts

I also added a Postgres functional regression test for data preservation. I could not run that DB-backed test locally because this machine does not have a Postgres server on localhost:5432.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@
 - **cli:** init command reading package.json from two folders up ([#11789](https://github.com/typeorm/typeorm/issues/11789)) ([dd55218](https://github.com/typeorm/typeorm/commit/dd55218648eb449937e22e1e7c88182db0048f1d))
 - **deps:** upgrade glob to fix CVE-2025-64756 ([#11784](https://github.com/typeorm/typeorm/issues/11784)) ([dc74f53](https://github.com/typeorm/typeorm/commit/dc74f5374ef5ec83d53045e4bca99cb9ff7d49d4))
 - **mongodb:** add missing `findBy` method to MongoEntityManager ([#11814](https://github.com/typeorm/typeorm/issues/11814)) ([38715bb](https://github.com/typeorm/typeorm/commit/38715bbd4169cae2910aac035cd2b05bddbaec5c))
+- **postgres:** avoid dropping columns when changing varchar length ([#3357](https://github.com/typeorm/typeorm/issues/3357))
 - **redis:** version detection logic ([#11815](https://github.com/typeorm/typeorm/issues/11815)) ([6f486e5](https://github.com/typeorm/typeorm/commit/6f486e5a67c007287949be119f233fb2b4fb7a59))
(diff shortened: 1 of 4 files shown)


inick89:
Follow-up review fixes pushed through `646c1c3`:

- kept the collation ALTER path on `createFullType(...)` so varchar length is preserved when length and collation change together
- handled collation removal with `COLLATE pg_catalog."default"` instead of emitting `"undefined"`
- added unit coverage for varchar length changes, length+collation changes, and collation removal
- added the `#3357` functional regression reference and a CHANGELOG entry
- removed the redundant `synchronize(true)` from the functional regression test; the suite `beforeEach` already reloads the DB from reset metadata

Latest local verification:

- `corepack pnpm run compile`
- `node .\node_modules\mocha\bin\mocha.js --no-config build\compiled\test\unit\query-runner\postgres-change-column.test.js` (3 passing)
- `corepack pnpm exec prettier --check src/driver/postgres/PostgresQueryRunner.ts test/unit/query-runner/postgres-change-column.test.ts test/functional/database-schema/column-length/postgres/column-length-postgres.test.ts CHANGELOG.md`
- `git diff --check`

The DB-backed Postgres functional test is added but I could not run it locally because this environment does not have a running Postgres service on `localhost:5432`.

Asti1982:
I checked the latest follow-up fixes and prepared one small local cleanup for the remaining advisory about brittle SQL assertions in `test/unit/query-runner/postgres-change-column.test.ts`.

The change keeps your implementation intact and only makes the new unit tests assert behavior/invariants instead of exact full SQL arrays:

- no `DROP COLUMN "example"`
- no `ADD "example"`
- required `ALTER COLUMN ... TYPE ...` statements are still present

Local verification on top of your PR head passed:

```text
corepack pnpm run compile
node .\node_modules\mocha\bin\mocha.js --no-config build\compiled\test\unit\query-runner\postgres-change-column.test.js
corepack pnpm exec prettier --check test/unit/query-runner/postgres-change-column.test.ts
git diff --check
```

Result: compile passed, targeted unit test `3 passing`, prettier passed, diff check passed.

I cannot push to your fork/branch without your explicit permission. If you want the patch, I can either share the tiny diff here or, if you prefer, you can grant temporary collaborator access to `Asti1982` on the fork branch and I can push the cleanup directly.

smith-xyz:

PR #12491, opened 2026-05-14

helh1723-lang:
## Summary

Fixes #3357.

Postgres varchar length changes are now emitted as ALTER COLUMN TYPE instead of DROP COLUMN followed by ADD COLUMN, avoiding data loss during schema synchronization / migration generation.

## Changes

- Detect length-only column changes in PostgresQueryRunner.changeColumn
- Generate reversible ALTER COLUMN TYPE statements for length-only changes
- Add a regression test for issue #3357 ensuring DROP/ADD is not generated

## Testing

- git diff --check
- Full test run was not possible in this Codex environment because Corepack could not create its pnpm cache directory due local filesystem permissions.

/opire try


smith-xyz:

Issue #4591, opened 2019-08-16
.save() returns columns marked select: false


0radek:
**Issue type:**

[ ] question
[X] bug report
[ ] feature request
[ ] documentation issue

**Database system/driver:**

[ ] `cordova`
[ ] `mongodb`
[ ] `mssql`
[ ] `mysql` / `mariadb`
[ ] `oracle`
[X] `postgres`
[ ] `cockroachdb`
[ ] `sqlite`
[ ] `sqljs`
[ ] `react-native`
[ ] `expo`

**TypeORM version:**

[X] `latest`
[ ] `@next`
[ ] `0.x.x` (or put your version here)

I have a `deletedAt` column on my entities which are `select: false`.  This works just fine for any select queries.

However, when using `.save()`, the returned values includes properties which are not supposed to be selected.  

Is this standard behavior, or should this be considered a bug?

Redanski:
This issue still persists is there a fix planned for future versions?

carlosempd:
Any update about this issue?

ardyfeb:
@carlosempd nothing, nobody seemed to care with this critical bug

kozzztya:
Have the same issue. It would be great to fix it.

KevLehman:
Same here. What I'm doing is to use the rest operator to hide manually the properties. But this shouldn't be the way.
```
const { password, ...user } = await repo.save(u);
```

ablankenship10:
Any updates on this? The workaround above is somewhat tedious and causes typing issues when attempting to return an instance of your entity missing expected fields. I can take a crack at it but unsure on direction. Has it been decided this is infact a bug and save() functionality should be fixed, or should we add a key, say `returning: bool` to `ColumnOptions` and make it an option without changing current functionality?

tkssharma:
```
     const userSupplierData = this.userSupplierTagsRepo.create();
      userSupplierData.organization_id = organization_id;
      userSupplierData.user_id = user_id;
      userSupplierData.supplier_profile = supplier;
      return  await this.userSupplierTagsRepo.save({ ...userSupplierData, ...payload })
```

How can I hide `supplier_profile foreign key Table data` coming in response after it saved
marking selected false won't work when we are doing same

Ojka91:
Another workaround after the save() is to delete de property like this
`await Entity.save();
delete Entity.password;
return Entity;`

Indeed tedious to do...

laurent-brisbois:
In my User entity I have :

```ts
[...]

PR #12493, opened 2026-05-14
fix(persistence): preserve select-false columns in entity subscriber events

### LINKED ISSUE #12473: `select: false` columns are no longer available in event subscriber afterUpdate
### Issue description

Data provided to a `select: false` column update is not present in the `UpdateEvent.entity`

### Expected Behavior

This behaviour has changed in the 0.3.29 release, stemming from [fix: prevent columns with select false from being returned](https://github.com/typeorm/typeorm/pull/11944).

I am unsure whether this is intended behaviour, as the UpdateEvent entity documentation states that it ["Contains the same data that was passed to the updating method"](https://github.com/typeorm/typeorm/blob/master/src/subscriber/event/UpdateEvent.ts#L14).

I think the resolution/expected behaviour here would be either:
1. Do not remove `select: false` columns in this scenario
1. Update the `UpdateEvent.entity` documentation to mention the exclusion of `select: false` columns

### Actual Behavior

Columns marked as `select: false` are removed from `UpdateEvent.entity`.

### Steps to reproduce

1. Create an entity, add a column with the `select: false` option
1. Add an `EventSubscriber` for the entity, add an `afterUpdate` method that logs `event.entity`
1. Execute a repository method on that entity (e.g. `.save()`) and include the `select: false` column.

```ts
// entity
@Entity()
export class MyEntity extends BaseEntity {
  @Column()
  normalCol: string;

  @Column({ select: false })
  selectFalseCol: string;

  // ...other columns
}

// subscriber to register
@EventSubscriber()
export class MyEntitySubscriber implements EntitySubscriberInterface<MyEntity> {
  listenTo() {
    return MyEntity;
  }

  async afterUpdate(event: UpdateEvent<MyEntity>) {
      console.log(event.entity);
  }
}

// call a method that would trigger the subscriber
await datasource.getRepository(MyEntity).save({
  normalCol: "...",
  selectFalseCol: "..."
});
```

Subscriber entity log excludes `selectFalseCol` change
```json
{
  "normalCol": "..."
}
``` 

### My Environment

| Operating System    |  macOS Tahoe 26.5        |
| Node.js version     | 24.11  |
| TypeORM version     | 0.3.29  |

### Additional Context

_No response_

- [ ] aurora-mysql
- [ ] aurora-postgres
- [ ] better-sqlite3
- [ ] capacitor
- [ ] cockroachdb
- [ ] cordova
- [ ] expo
- [ ] mongodb
- [ ] mysql
- [ ] nativescript
- [ ] oracle
- [ ] postgres
- [ ] react-native
- [ ] sap
- [ ] spanner
- [ ] sqlite
- [ ] sqljs
- [x] sqlserver

mbursali93:
## Problem

`select: false` columns are missing from `event.entity` in entity subscriber events (afterInsert, afterUpdate) because `SubjectExecutor` mutates the entity object before broadcasting.

## Root Cause

`SubjectExecutor` manipulates the entity during execution. By the time `broadcastAfterEventsForAll` is called, `select: false` columns have been stripped from the entity.

## Fix

Added `snapshotEntity()` method to `Subject` class that takes a shallow copy of the entity before execution begins. Broadcaster uses `originalEntity ?? entity` to ensure original values are preserved.

## Testing

Added integration tests for `afterInsert` and `afterUpdate` subscriber events verifying that `select: false` columns are present.

Fixes #12473

### DIFF
--- a/src/persistence/Subject.ts
+++ b/src/persistence/Subject.ts
@@ -63,6 +63,12 @@ export class Subject {
     databaseEntity?: ObjectLiteral
 
     /**
+     * Original entity snapshot before execution.
+     * Preserved to pass select:false columns to afterUpdate broadcast event.
+     */
+    originalEntity?: ObjectLiteral
+
+    /**
      * Indicates if database entity was loaded.
      * No matter if it was found or not, it indicates the fact of loading.
      */
@@ -169,6 +175,7 @@ export class Subject {
             this.changeMaps.push(...options.changeMaps)
 
         this.recompute()
+        this.snapshotEntity()
     }
 
     // -------------------------------------------------------------------------
@@ -331,6 +338,16 @@ export class Subject {
             )
         } else if (this.databaseEntity) {
             this.identifier = this.metadata.getEntityIdMap(this.databaseEntity)
+        }
+    }
+
+    /**
+     * Takes a snapshot of the current entity state.
+     * Used to preserve original values (e.g. select:false columns) before execution mutates the entity.
+     */
+    snapshotEntity(): void {
+        if (this.entity && !this.originalEntity) {
+            this.originalEntity = structuredClone(this.entity)
         }
     }
 }
--- a/src/persistence/Subject.ts
+++ b/src/persistence/Subject.ts
@@ -347,7 +347,7 @@ export class Subject {
      */
     snapshotEntity(): void {
         if (this.entity && !this.originalEntity) {
-            this.originalEntity = structuredClone(this.entity)
+            this.originalEntity = { ...this.entity }
         }
     }
 }
--- a/src/persistence/SubjectExecutor.ts
+++ b/src/persistence/SubjectExecutor.ts
@@ -299,7 +299,7 @@ export class SubjectExecutor {
                 this.queryRunner.broadcaster.broadcastAfterInsertEvent(
                     result,
                     subject.metadata,
-                    subject.entity!,
+                    subject.originalEntity!,
                     subject.identifier,
                 ),
             )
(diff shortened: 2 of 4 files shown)


gioboa (commented):
Thanks for your help @mbursali93 
Is this fix working for you @tada5hi @Johoseph ?

tada5hi:
This PR routes only the *subscriber broadcast* (`broadcastAfterInsertEvent`, `broadcastAfterUpdateEvent`) through `subject.originalEntity`. The underlying `delete target[column.propertyName]` in `SubjectExecutor.updateSpecialColumnsInInsertedAndUpdatedEntities` (from #11944) still mutates `subject.entity` itself, which is the same reference `repository.save()` hands back to the caller — and the same reference the caller passed in. So `select: false` columns are still stripped from the post-save return value, just not from the subscriber event anymore.

Minimal repro (no subscriber):

```ts
@Entity()
class User {
    @PrimaryGeneratedColumn('uuid') id: string;
    @Column() name: string;
    @Column({ select: false }) email: string;
}

const entity = repo.create({ name: 'foo', email: 'foo@example.com' });
const saved = await repo.save(entity);

console.log(saved.email);       // undefined on 0.3.29, 'foo@example.com' on 0.3.28
console.log('email' in entity); // false on 0.3.29, true on 0.3.28 (passed-in ref is mutated too)
```

To fully restore the 0.3.28 contract, the snapshot-and-restore needs to apply to `subject.entity` as well — either by restoring stripped properties onto `subject.entity` after `updateSpecialColumnsInInsertedAndUpdatedEntities` runs, or by avoiding the `delete` altogether and computing the "DB-shaped" projection only at the broadcast boundary. The current diff leaves the second symptom in place; further context in #12473.

alumni:
Hi @mbursali93 thanks for the contribution!

We decided instead to revert the PR that introduced the bug, by merging #12501. I think it's the right thing to do considering the original issue - see https://github.com/typeorm/typeorm/issues/4591#issuecomment-4615171349

PR #12495, opened 2026-05-14
feat: add `prepareEntityMetadata` option to DataSource

sleep-written:
**What does this PR do?**

Adds an optional `prepareEntityMetadata` hook to `DataSourceOptions` that is called for each `EntityMetadata` instance after it is built by `ConnectionMetadataBuilder`, but before it is assigned to the data source and validated.

**Motivation**

Currently, the only way to mutate entity metadata on a per-datasource basis is to patch the global `MetadataArgsStorage` via `getMetadataArgsStorage()` before calling `initialize()`. This is a side effect that contaminates all data sources initialized after the patch, requiring a manual restore to avoid leaking state — particularly problematic in test environments running multiple data sources.

**Previous behavior**

```ts
// Only option: mutate global storage and restore manually
const storage = getMetadataArgsStorage();
storage.columns
    .filter(x => x.options.type === 'bit')
    .forEach(x => x.options.type = 'boolean');

await dataSource.initialize();
```

**New behavior**

```ts
    type: 'better-sqlite3',
    // ...
    prepareEntityMetadata(meta) {
        meta.columns
            .filter(c => c.type === 'bit')
            .forEach(c => c.type = 'boolean');
    },
});

await dataSource.initialize();
```

The hook supports `async`/`await`, is completely isolated to the data source instance, and does not touch `MetadataArgsStorage`.

**Tests**

- should call `prepareEntityMetadata` once per entity metadata
- should make mutations visible in the final entity metadata
- should await async `prepareEntityMetadata` before continuing
- should initialize normally when `prepareEntityMetadata` is not provided

**Testing**

To run the tests for this feature specifically:

```shell
pnpm run test -- --grep 'DataSource > prepareEntityMetadata'
```


sleep-written:
Re: 5. Order-dependent early-stop test `🐞 Bug` `☼ Reliability`
Fixed in [2eb55cd](https://github.com/typeorm/typeorm/pull/12495/changes/2eb55cd306f6db3f2f6fe23b5408d861ff6a8d02): replaced entity name check with a call counter to avoid order-dependent assertions.

michaelbromley:

```
git fetch origin
```

PR #12496, opened 2026-05-14

sureshchouksey8:
### Description

Fixes #3357.

Postgres currently treats a column length change as a destructive type change in `changeColumn`, so generated migrations drop and recreate the column. For a safe varchar length change, that risks data loss even though Postgres can alter the column type in place.

This PR moves length changes into the existing `ALTER COLUMN ... TYPE` path for Postgres, alongside precision and scale changes.

### Changes

- Stop recreating Postgres columns when only `length` changes.
- Generate `ALTER TABLE ... ALTER COLUMN ... TYPE ...` for varchar length changes.
- Add a regression test for changing `varchar(50)` to `varchar(51)` without generating `DROP COLUMN` / `ADD COLUMN`.

### Verification

Local checks run:

- `npx --yes pnpm@9.15.4 install --frozen-lockfile`
- `PATH=/Users/ronny/.cache/codex-runtimes/codex-primary-runtime/dependencies/node/bin:$PATH ./node_modules/.bin/tsc --noEmit`
- `npx --yes pnpm@9.15.4 exec prettier --check src/driver/postgres/PostgresQueryRunner.ts test/github-issues/3357/issue-3357.test.ts test/github-issues/3357/entity/Post.ts`
- `npx --yes pnpm@9.15.4 run compile`
- `git diff --check`

I could not run the Postgres integration test locally because this environment does not have Docker or an `ormconfig.json`/Postgres test database configured.


sureshchouksey8:
Addressed the Qodo review feedback in follow-up commits:

- Moved the regression coverage into `test/functional/query-runner/change-column.test.ts` and removed the temporary `test/github-issues/3357` files.
- Updated the Postgres `changeColumn` path so the cloned table cache is kept in sync after `ALTER COLUMN ... TYPE` changes length, precision, or scale.

Local checks run after the update:

- `tsc --noEmit`
- `pnpm exec prettier --check src/driver/postgres/PostgresQueryRunner.ts test/functional/query-runner/change-column.test.ts`
- `pnpm run compile`
- `git diff --check`

smith-xyz:

PR #12498, opened 2026-05-14

neverstopclicking:
Fixes #3357.

This changes the Postgres schema diff path for varchar/character varying length-only changes so migration generation emits `ALTER COLUMN ... TYPE ...` instead of dropping and recreating the column.

The previous drop/add behavior can lose data when a column changes from e.g. `varchar(50)` to `varchar(51)`.

Verification:
- `./node_modules/.bin/prettier --check src/driver/postgres/PostgresQueryRunner.ts test/github-issues/3357/entity/Bug.ts test/github-issues/3357/issue-3357.test.ts`
- `./node_modules/.bin/tsc --noEmit`
- `./node_modules/.bin/eslint src/driver/postgres/PostgresQueryRunner.ts test/github-issues/3357/entity/Bug.ts test/github-issues/3357/issue-3357.test.ts` (warnings only)

Note: the added Postgres regression test requires a local TypeORM `ormconfig.json`/Postgres test setup, which is not configured in this environment.


neverstopclicking:
Addressed the Qodo findings:

- Moved the regression coverage from `test/github-issues/3357` into the existing functional Postgres column-length suite.
- Normalized default comparison for the varchar length-only path so Postgres function defaults that differ only by casing still use `ALTER COLUMN ... TYPE ...` instead of falling back to drop/add.

Re-ran:
- `prettier --check` on changed files
- `tsc --noEmit`
- `eslint` on changed files (warnings only, matching existing file-level warning pattern)

The targeted functional test still requires the repo's local `ormconfig.json`/Postgres test setup, which is not configured in this environment.

smith-xyz:

PR #12499, opened 2026-05-15
fix(query-builder): propagate queryEntity flag in QueryExpressionMap.clone()

hypernova1:
## Problem

`QueryExpressionMap.clone()` does not copy the `queryEntity` field (which defaults to `false`). This causes a `QueryFailedError` when calling `findOne()` with a partial `select` on an entity that has eager relations.

### Reproduction

```ts
// Entity with an eager OneToMany relation
@Entity()
class Post {
  @PrimaryGeneratedColumn() id: number
  @Column() title: string
  @OneToMany(() => Editor, e => e.post, { eager: true }) editors: Editor[]
}

// Throws: column "distinctAlias"."Post_id" does not exist
await dataSource.manager.findOne(Post, {
  where: { id: 1 },
  select: ['title'],
})
```

### Root cause

`findOne()` sets `take: 1` and joins eager relations (adding join attributes). TypeORM detects `take + joinAttributes > 0` and generates a pagination DISTINCT wrapper inside `executeEntitiesAndRawResults()`:

```sql
-- outer wrapper
SELECT DISTINCT "distinctAlias"."Post_id" AS "ids_Post_id"
FROM (
  -- inner: originalQuery = this.clone()
  SELECT "Post"."title" AS "Post_title"  -- Post_id is MISSING
  FROM "post" "Post"
  LEFT JOIN ...
) AS "distinctAlias"
```

The outer query references `distinctAlias.Post_id`, but the inner query (the clone) does not include `Post_id` because `queryEntity` was not copied by `clone()` and defaults to `false`. With `queryEntity = false`, the `nonSelectedPrimaryColumns` auto-include logic in `buildEscapedEntityColumnSelects()` is skipped.

### Fix

Add `map.queryEntity = this.queryEntity` to `QueryExpressionMap.clone()`.

When `getRawAndEntities()` sets `queryEntity = true` before calling `executeEntitiesAndRawResults()`, cloning now correctly propagates `true` to the inner query, causing the PK to be included in the SELECT — enabling the DISTINCT wrapper to resolve the column.

## Changes

- `src/query-builder/QueryExpressionMap.ts` — add `queryEntity` to `clone()`
- `test/functional/relations/eager-relations/basic-eager-relations/basic-eager-relations.test.ts` — regression test


test/functional/relations/eager-relations/basic-eager-relations/basic-eager-relations.test.ts
```diff
@@ -233,9 +230,33 @@ describe("relations > eager relations > basic", () => {
...
+            dataSources.map(async (dataSource) => {
+                await prepareData(dataSource)
+
+                // Regression test for: QueryExpressionMap.clone() did not copy
+                // queryEntity, causing the pagination DISTINCT wrapper's inner
+                // query to omit primary key columns from the SELECT. When eager
+                // relations exist the wrapper references distinctAlias.<Entity>_id
+                // which then could not be found, producing a QueryFailedError.
+                let error: Error | undefined
+                try {
```
smith-xyz, line 253: nit: try/catch here is unnecessary

test/functional/relations/eager-relations/basic-eager-relations/basic-eager-relations.test.ts
```diff
@@ -233,9 +230,33 @@ describe("relations > eager relations > basic", () => {
...
+                    await dataSource.manager.findOne(Post, {
```
smith-xyz, line 254: this only verifies no error was thrown - let's also verify the correct data was returned

smith-xyz (approved):
lgtm! nothing blocking for this change - thanks for fixing this issue

michaelbromley:

```
git fetch origin
```

PR #12501, opened 2026-05-15
fix(persistence): preserve select false columns on the in-memory entity after save()

### LINKED ISSUE #12473: `select: false` columns are no longer available in event subscriber afterUpdate

tada5hi:
### Description of change

`SubjectExecutor.updateSpecialColumnsInInsertedAndUpdatedEntities` deletes every `select: false` column from `subject.entity` after persistence (introduced in #11944). This patch removes that delete and restores the 0.3.28 contract.

### Problem

The post-persist delete added in #11944 has two unintended effects on `Repository.save()`:

1. The entity returned by `save()` no longer carries the value the caller just supplied.
2. The original entity reference passed to `save()` is mutated in place, so application code holding that reference also loses the field.

It also produces an asymmetry: columns with `default !== undefined` (or `@CreateDateColumn`, `@Generated`, ...) survive because `getInsertionReturningColumns()` includes them in `RETURNING` and `subject.generatedMap` is merged back into the entity *after* the delete. Columns without a default are stripped permanently. Same `select: false` decorator option, opposite outcomes.

### Reproduction (no subscriber needed)

```ts
@Entity()
class User {
    id: string

    @Column()
    name: string

    @Column({ select: false })
    email: string
}

const entity = repo.create({ name: "foo", email: "foo@example.com" })
const saved = await repo.save(entity)

console.log(saved.email)        // undefined in 0.3.29, "foo@example.com" in 0.3.28
console.log("email" in entity)  // false in 0.3.29, true in 0.3.28
```

### Why revert rather than narrow the delete

The 0.3.28 contract was "`select: false` controls what comes *out of the DB* on a read; the in-memory entity is untouched by persist." Callers who want the value back from a subsequent fetch use `addSelect`. Restoring that contract for both the returned entity and the subscriber event keeps existing user code working without a new opt-in flag, and removes the default-vs-non-default asymmetry above.

If the maintainers prefer the stricter "`select: false` means hidden everywhere, including post-save" semantics that #11944 introduced, that's a defensible position too — but it's a breaking change that warrants a `0.4.0` rather than a `0.3.x` patch, plus a documented migration path (e.g. a `reload: true` option on `save()` that re-fetches with `addSelect`, or `RETURNING` of `select: false` columns when the caller explicitly opts in).

### Relationship to #12493

#12493 addresses the same root cause (the same delete in `SubjectExecutor`) by snapshot-restoring `event.entity` for subscriber broadcast only. That fixes the symptom in `afterInsert` / `afterUpdate` events but leaves the user-facing `repo.save()` return mutated. By removing the delete entirely, this PR fixes both symptoms and makes the snapshot/restore step in #12493 unnecessary.

### Tests

- The two assertions added to `test/functional/columns/select-control/select-control.test.ts` in #11944 are inverted to match the restored contract: `select: false` columns and embedded `select: false` columns are now expected to remain on the returned entity, and the input reference is asserted to be unmutated.
- New `test/github-issues/12473/` covers the returned entity, the input reference, and both `afterInsert` and `afterUpdate` subscribers.

Closes #12473.
Supersedes #12493.

### Pull-Request Checklist

- [x] Code is up-to-date with the \`master\` branch
- [x] This pull request links relevant issues as Close #12473
- [x] There are new or updated tests validating the change (\`tests/**.test.ts\`)
- [x] Documentation has been updated to reflect this change (\`docs/docs/**.md\`)


tada5hi:
@alumni

alumni (approved):

gioboa (approved):
Thanks @tada5hi for your help

alumni:
@tada5hi thanks for the PR and sorry for the slow reply - yes, `select: false` should not remove the property after saving - for the original issue we might introduce a separate option in the future if requested by users, as mentioned here https://github.com/typeorm/typeorm/issues/4591#issuecomment-4615171349

alumni:
Removed the "backport" label because it was already backported - the fix will be available in both 0.3 and 1.0

PR #12502, opened 2026-05-15
fix(postgres): use ALTER COLUMN TYPE instead of DROP and ADD when only length changes

yi-hialong:
## Summary
Fixes issue where changing a column's length (e.g., archar(50) ? archar(100)) in PostgreSQL would incorrectly trigger a DROP COLUMN + ADD COLUMN operation, causing data loss.

## Problem
The original code checked oldColumn.length !== newColumn.length and would drop and recreate the column whenever the length changed, even if the column type stayed the same. This caused unnecessary data loss and downtime.

## Solution
When only the length changes (type stays the same), PostgreSQL can safely use ALTER COLUMN TYPE without data loss. The fix separates the logic:

1. **Type/array/generated column changes** ? Still use DROP + ADD (data conversion required)
2. **Length-only changes** ? Use ALTER COLUMN ... TYPE (safe for PostgreSQL)

## Fixes
Fixes typeorm/typeorm#3357

## Testing
- Manual testing confirms varchar length changes no longer cause data loss
- Existing tests pass

---
*This PR is submitted for the  bounty on issue #3357*

### DIFF
--- a/src/driver/postgres/PostgresQueryRunner.ts
+++ b/src/driver/postgres/PostgresQueryRunner.ts
@@ -1326,39 +1326,17 @@ export class PostgresQueryRunner
                 `Column "${oldTableColumnOrName}" was not found in the "${table.name}" table.`,
             )
 
-        if (oldColumn.type !== newColumn.type || newColumn.isArray !== oldColumn.isArray ||
+        if (oldColumn.type !== newColumn.type ||
+            newColumn.isArray !== oldColumn.isArray ||
             (!oldColumn.generatedType && newColumn.generatedType === "STORED") ||
             (oldColumn.asExpression !== newColumn.asExpression && newColumn.generatedType === "STORED")
         ) {
-            // Type, array, or generated column changed - need to recreate column
+            // To avoid data conversion, we just recreate column
             await this.dropColumn(table, oldColumn)
             await this.addColumn(table, newColumn)
 
             // update cloned table
             clonedTable = table.clone()
-        } else if (oldColumn.length !== newColumn.length) {
-            // For PostgreSQL, when type stays the same and only length changes,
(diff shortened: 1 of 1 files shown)


yi-hialong:
Thanks for the thorough code review! @qodo-free-for-open-source-projects[bot]

## Fixed Issues

### 1. Bug: Length change short-circuits updates ?
**Fixed**: Removed the separate else if (oldColumn.length !== newColumn.length) branch and merged the length change detection into the existing precision/scale block (line ~1639). Now when length changes alongside other properties (default, nullable, etc.), all changes are applied correctly.

### 2. Inconsistent formatting ?  
**Fixed**: Restored the original comment formatting and removed the new code block that was causing style inconsistencies.

## Code Changes
`diff
- if (oldColumn.type !== newColumn.type || newColumn.isArray !== oldColumn.isArray ||
-     ...
- ) {
-     // Type, array, or generated column changed - need to recreate column
+ if (oldColumn.type !== newColumn.type ||
+     newColumn.isArray !== oldColumn.isArray ||
+     ...
+ ) {
+     // To avoid data conversion, we just recreate column
`

`diff
  if (
      newColumn.precision !== oldColumn.precision ||
-     newColumn.scale !== oldColumn.scale
+     newColumn.scale !== oldColumn.scale ||
+     oldColumn.length !== newColumn.length
  ) {
`

The fix ensures that:
1. Type/array changes ? DROP + ADD (data conversion needed)
2. Length-only OR precision/scale changes ? ALTER COLUMN TYPE (safe in PostgreSQL)
3. Other changes (default, nullable, rename) ? handled in same block

Note: Push is pending due to network issues. Will update once pushed.

smith-xyz:

PR #12505, opened 2026-05-16
fix: alter postgres varchar length without dropping data

falorfrozen-cmd:
## Summary

Fixes #3357.

Postgres currently recreates a column when only the length of a length-aware column changes, for example `varchar(50)` -> `varchar(51)`. That generates a `DROP COLUMN` followed by `ADD COLUMN`, which loses existing data during schema sync or generated migrations.

This change keeps those changes on the regular `ALTER COLUMN ... TYPE ...` path when the column type itself is unchanged and the type supports length.

## Changes

- Detect length-only changes for Postgres column types that support length.
- Use `ALTER TABLE ... ALTER COLUMN ... TYPE ...` instead of dropping/recreating the column.
- Keep the QueryRunner cached table metadata in sync after length-only changes.
- Add functional regression coverage for changing a `varchar` length while preserving existing row data.
- Verify the generated down query restores the previous length without dropping data.

## Testing

- `corepack pnpm run typecheck`
- `corepack pnpm run compile`
- `corepack pnpm exec prettier --check src/driver/postgres/PostgresQueryRunner.ts test/functional/query-runner/change-column.test.ts`
- `corepack pnpm run lint -- src/driver/postgres/PostgresQueryRunner.ts test/functional/query-runner/change-column.test.ts`
- `corepack pnpm exec mocha --no-config --exit --timeout 90000 --file ./build/compiled/test/utils/test-setup.js ./build/compiled/test/functional/query-runner/change-column.test.js`

### DIFF
--- a/src/driver/postgres/PostgresQueryRunner.ts
+++ b/src/driver/postgres/PostgresQueryRunner.ts
@@ -1326,9 +1326,18 @@ export class PostgresQueryRunner
                 `Column "${oldTableColumnOrName}" was not found in the "${table.name}" table.`,
             )
 
+        const lengthChanged = oldColumn.length !== newColumn.length
+        const collationChanged = oldColumn.collation !== newColumn.collation
+        const canAlterLength =
+            lengthChanged &&
+            oldColumn.type === newColumn.type &&
+            this.driver.withLengthColumnTypes.indexOf(
+                newColumn.type as ColumnType,
+            ) !== -1
+
         if (
             oldColumn.type !== newColumn.type ||
-            oldColumn.length !== newColumn.length ||
+            (lengthChanged && !canAlterLength) ||
             newColumn.isArray !== oldColumn.isArray ||
             (!oldColumn.generatedType &&
                 newColumn.generatedType === "STORED") ||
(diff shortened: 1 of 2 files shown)


falorfrozen-cmd:
I reviewed the latest Qodo findings against the current PR head (`4e8826f`) and addressed the actionable items there.

Summary of the current state:

- The collation `ALTER COLUMN ... TYPE ... COLLATE` path now uses `this.driver.createFullType(newColumn)`, so changing length and collation in one `changeColumn()` call no longer drops the varchar typmod.
- The length-only `ALTER TYPE` path is skipped when collation also changes, avoiding the redundant `ALTER TYPE` statement.
- The cloned table cache is now updated for both `length` and `collation` before `replaceCachedTable()`.
- The regression test is in `test/functional/query-runner/change-column.test.ts`; the PR no longer adds a `test/github-issues/3357` file.
- The test clears SQL memory after setup, so `executeMemoryDownSql()` only reverts the `changeColumn()` migration SQL rather than dropping the setup table.
- The test also asserts that only one up and one down `ALTER COLUMN "example" TYPE` query are generated for the combined length + collation change.

Verification run locally:

```text
corepack pnpm run typecheck
corepack pnpm run compile
corepack pnpm exec prettier --check src/driver/postgres/PostgresQueryRunner.ts test/functional/query-runner/change-column.test.ts
corepack pnpm run lint -- src/driver/postgres/PostgresQueryRunner.ts test/functional/query-runner/change-column.test.ts
corepack pnpm exec mocha --no-config --exit --timeout 90000 --file ./build/compiled/test/utils/test-setup.js ./build/compiled/test/functional/query-runner/change-column.test.js
```

The targeted Postgres test passed with `4 passing`; lint reported `0 errors` with existing warnings in the large query runner file.

smith-xyz:

PR #12506, opened 2026-05-16
feat(save): add skipExistingRecordCheck option to SaveOptions

sceccotti89:
## Summary

- Adds a new `skipExistingRecordCheck` option to `SaveOptions`
- When `true`, skips the SELECT query in `EntityPersistExecutor` that checks whether
  each entity already exists, forcing an INSERT for every entity
- Logic lives in `EntityPersistExecutor` (the save orchestrator); `SubjectDatabaseEntityLoader`
  is unchanged

## Motivation

In bulk-insert scenarios where the caller knows none of the entities exist yet, the
SELECT roundtrip that determines INSERT vs UPDATE is unnecessary overhead.

## How it works

When the option is set, `EntityPersistExecutor` marks each subject's
`databaseEntityLoaded = true` (without setting `databaseEntity`). This drives the
existing `Subject` decision logic so `mustBeInserted = true` and `mustBeUpdated = false`.

## Usage

await repository.save(entities, { skipExistingRecordCheck: true })

### DIFF
--- a/docs/docs/working-with-entity-manager/6-repository-api.md
+++ b/docs/docs/working-with-entity-manager/6-repository-api.md
@@ -98,10 +98,15 @@ const user = await repository.preload(partialUser)
   It saves all given entities in a single transaction (in the case of entity, manager is not transactional).
   Also supports partial updating since all undefined properties are skipped.
   Returns the saved entity/entities.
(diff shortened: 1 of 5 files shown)


michaelbromley:

```
git fetch origin
```

PR #12507, opened 2026-05-17
fix: handle schema-qualified closure table joins

TupiC:
### Description of change

- what the change is intended to do
sometimes a join uses a table name with a schema, e.g. public.some_closure_table. That string contains a dot. TypeORM must treat that whole thing as one table name, not as something.something_else where the part before the dot is a query alias. This change fixes that: only dotted strings whose left part is a real join alias count as relation paths.

- why this change is needed
without this, TreeRepository closure tree joins can build that qualified closure table path and TypeORM mistakenly looks for an alias called public. Then it throws "TypeORMError: "public" alias was not found. Maybe you forgot to join it?" before Postgres even runs a query

- how you've verified it
```bash
pnpm run compile && pnpm run test:fast -- --grep "schema-qualified closure tables"
```

- any other context that will help reviewers understand the change
This did not occur on `v0.3.28` but started after upgrading to `v0.3.29`, which is why I opened this PR.

### Pull-Request Checklist

- [x] Code is up-to-date with the `v0.3` branch
- [ ] This pull request links relevant issues as `Fixes #00000` N/A


smith-xyz (commented):
@TupiC can you address the qodo findings - I agree that this was just exception based. Any unwanted side effects for returning undefined now?

TupiC:
@smith-xyz thanks. I just updated the code.. we now only return undefined for schema-qualified metadata table paths when the left side is not an existing query alias. If the left side is an alias we continue through the normal relation-resolution path via findAliasByName. so the previous missing-alias validation behavior is preserved. (also moved the test to the correct existing file)

smith-xyz (approved):

gioboa (approved):
Thanks @TupiC

PR #12521, opened 2026-05-19
fix(sqlserver): clone FindOperator before transforming to prevent FindOptionsWhere mutation

### LINKED ISSUE #11733: Unable to reuse `FindOptionsWhere<T>` instances as values get mutated in place
### Issue description

TypeORM now mutates `FindOptionsWhere<T>` instances in place from raw values to `MssqlParameter` instances, which causes problems when they get passes to value transformers

### Expected Behavior

Global `FindOptionsWhere<T>` instances should be able to be reused across different queries.

### Actual Behavior

We need to construct new `FindOptionsWhere<T>` instances for each query.

### Steps to reproduce

I've opened <https://github.com/typeorm/typeorm/pull/11734> in an attempt to expose the actual behavior.

Running the updated test on my local produces the following output:

```
  github issues > #11285 Missing MSSQL input type
    mssql connection
      1) should convert input parameter with FindOperator with array to MssqlParameter


  0 passing (349ms)
  1 failing

  1) github issues > #11285 Missing MSSQL input type
       mssql connection
         should convert input parameter with FindOperator with array to MssqlParameter:

      AssertionError: expected { memberId: FindOperator{ …(7) } } to deeply equal { memberId: FindOperator{ …(7) } }
      + expected - actual

                 "_objectLiteralParameters": [undefined]
                 "_type": "in"
                 "_useParameter": true
                 "_value": [
      -            {
      -              "@instanceof": Symbol(MssqlParameter)
      -              "params": []
      -              "type": "varchar"
      -              "value": "test-member-id-2"
      -            }
      +            "test-member-id-2"
                 ]
               }
             }
           ]

      at C:\Users\me\dev\typeorm\test\github-issues\11285\issue-11285.ts:185:38
      at processTicksAndRejections (node:internal/process/task_queues:105:5)
      at async Promise.all (index 0)
```

### My Environment

| Operating System    | Windows 11         |
| TypeORM version     | 0.3.27  |

### Additional Context

_No response_

- [ ] aurora-mysql
- [ ] aurora-postgres
- [ ] better-sqlite3
- [ ] cockroachdb
- [ ] cordova
- [ ] expo
- [ ] mongodb
- [ ] mysql
- [ ] nativescript
- [ ] oracle
- [ ] postgres
- [ ] react-native
- [ ] sap
- [ ] spanner
- [ ] sqlite
- [ ] sqlite-abstract
- [ ] sqljs
- [x] sqlserver

kyungseopk1m:
### Description of change

This fixes #11733. Reusing the same `FindOptionsWhere<T>` instance across multiple queries was broken on the SQL Server driver because `SqlServerDriver.parametrizeValues()` called `FindOperator.transformValue()` which mutates the operator's internal value in place. On the second query the value was already wrapped in a `MssqlParameter` and value transformers received the wrapper instead of the raw value.

The same in-place mutation pattern is also used in `ApplyValueTransformers.transformTo()` for any driver with a column-level transformer, so the fix is applied symmetrically.

### Approach

- Added `FindOperator.clone()` that returns a new `FindOperator` with the same configuration, recursively cloning nested `FindOperator` values (including those inside arrays) and copying array values so per-element mutations do not leak back to the caller.
- `SqlServerDriver.parametrizeValues()` and `ApplyValueTransformers.transformTo()` now clone the input operator before calling `transformValue()`, so the caller's instance is never mutated. The `type === "raw"` short-circuit in the SQL Server path is preserved.

### Tests

- `test/functional/find-options/find-options-where-no-mutate/` — five unit checks of `FindOperator.clone()` (new instance with same configuration, mutation isolation, nested recursion, array element copy, and recursive cloning of `FindOperator` values inside arrays) plus four functional cases covering reused `FindOperator` with a column transformer, nested operator reuse, ordinary single-shot queries (no-op), and reuse without any column transformer.
- `test/github-issues/11285/issue-11285.test.ts` — adds an MSSQL-only regression that reuses `Not(In([memberId]))` twice and asserts the original operator's array values are not replaced by `MssqlParameter` instances on subsequent queries.

### Pull-Request Checklist

- [x] Code is up-to-date with the \`master\` branch
- [x] This pull request links relevant issues as \`Closes #11733\`

Closes #11733


src/driver/sqlserver/SqlServerDriver.ts
```diff
@@ -1088,14 +1088,18 @@ export class SqlServerDriver implements Driver {
...
-                    to: (v) => this.parametrizeValues(column, v),
-                    from: (v) => v,
-                })
-            }
+            if (value.type === "raw") return value
+
+            // Clone before mutating so callers that reuse a shared
+            // FindOptionsWhere<T> object across multiple queries see the
+            // original FindOperator unchanged (#11733).
+            const cloned = value.clone()
```
smith-xyz, line 1096: nit: The clone-before-mutate pattern is correct but lives in the wrong layer imo. TransformValue itself should return a new instance rather than mutating in place, so future callers don't need to remember to clone defensively. Fine for this PR, but worth a follow-up to make FindOperator immutable-by-default.

so then its:

```typescript
transformValue(transformer: ValueTransformer | ValueTransformer[]): this {
    const cloned = Object.assign(
        Object.create(Object.getPrototypeOf(this)),
        this,
    ) as this
    if (cloned._value instanceof FindOperator) {
        cloned._value = cloned._value.transformValue(transformer)
    } else {
        cloned._value = Array.isArray(cloned._value) && cloned._multipleParameters
            ? cloned._value.map((v: any) =>
                  ApplyValueTransformers.transformTo(transformer, v),
              )
            : ApplyValueTransformers.transformTo(transformer, cloned._value)
    }
    return cloned
}
```

and simplified call sites:

```typescript
// SqlServerDriver
return value.transformValue({ to: (v) => this.parametrizeValues(column, v), from: (v) => v })
// ApplyValueTransformers
return entityValue.transformValue(transformer)
```

src/find-options/FindOperator.ts
```diff
@@ -158,4 +158,29 @@ export class FindOperator<T> {
...
+     * Returns a shallow copy of this FindOperator so callers can safely
+     * apply mutating operations like `transformValue` without affecting
+     * the original instance. The prototype is preserved so subclasses
+     * such as EqualOperator stay intact. Nested FindOperator children
+     * are cloned recursively, including FindOperator values inside
+     * arrays, and array values are copied so that in-place mutations on
+     * the clone's value do not leak back to the caller.
+     */
+    clone(): this {
+        const cloned = Object.create(Object.getPrototypeOf(this)) as this
```
smith-xyz: nit: Could collapse the Object.create + Object.assign to one line

```typescript
clone(): this {
    const cloned = Object.assign(
        Object.create(Object.getPrototypeOf(this)),
        this,
    ) as this

    if (InstanceChecker.isFindOperator(this._value)) {
        cloned._value = this._value.clone()
    } else if (Array.isArray(this._value)) {
        cloned._value = this._value.map((v: any) =>
            InstanceChecker.isFindOperator(v) ? v.clone() : v,
        )
    }

    return cloned
}
```

smith-xyz (approved):
nothing blocking from my end, thank you for the fix!

kyungseopk1m:
@smith-xyz thanks for the nit. applied in 3300709, kept `as T` since TS doesn't narrow on the array branch. transformValue rework as a follow-up PR after this lands.

gioboa (approved):
Thanks for your help @kyungseopk1m

src/driver/sqlserver/SqlServerDriver.ts
alumni, line 1096: I would wonder why do we even mutate the values. I think this should probably not happen, so cloning should not be needed.

michaelbromley:

```
git fetch origin
```

PR #12509, opened 2026-05-18
fix: alter Postgres varchar length in migrations

xuanan080106:
## Summary

Fixes #3357 by changing Postgres column length-only updates to emit `ALTER COLUMN ... TYPE` instead of dropping and recreating the column.

## Changes

- Stops treating `TableColumn.length` changes as a destructive Postgres column recreation trigger.
- Includes `length` in the existing Postgres `ALTER COLUMN ... TYPE` path used for precision/scale changes.
- Adds a regression test covering `varchar(50)` to `varchar(51)` migration SQL generation without `DROP COLUMN`/`ADD COLUMN`.

## Verification

```bash
git diff --check
# Intended focused test:
pnpm run test:fast -- test/github-issues/3357/issue-3357.test.ts
```

Results:

- `git diff --check` passed.
- Focused runtime test was not run in this environment: `pnpm` is not installed/in PATH, and the TypeORM Postgres test requires a configured Postgres test database.

## Payout

Opire bounty: https://app.opire.dev/issues/01HWJNZ5HQMVG2TCW6XHQQJ3QT

Payout wallet: `RTC29WwMjwcaFeTTQqKaMNmFUFLYz3f`

PR #12522, opened 2026-05-19
fix: alter postgres varchar length with alter column instead of drop and add

coolcam2:
## Summary
Fixes #3357 — when changing a Postgres `varchar` column's length (e.g., 50→51), migration generation previously emitted `DROP COLUMN` + `ADD COLUMN`, causing **data loss**. This PR makes length-only changes go through the `ALTER COLUMN TYPE` path instead.

### `src/driver/postgres/PostgresQueryRunner.ts`
- **Removed** `oldColumn.length !== newColumn.length` from the destructive column-recreation check
- **Added** `newColumn.length !== oldColumn.length` to the `ALTER COLUMN TYPE` condition alongside precision/scale

### Why this is cleaner than #12509
`createFullType()` already includes `length` in the type string. By routing length changes through the existing ALTER COLUMN path (same as precision/scale), we get correct SQL without any new code paths.

## Before
```sql
ALTER TABLE "bug" ADD "example" character varying(51)  -- DATA LOSS!
```

## After
```sql
ALTER TABLE "bug" ALTER COLUMN "example" TYPE character varying(51)  -- Safe!
```

Fixes #3357

### DIFF
--- a/src/driver/postgres/PostgresQueryRunner.ts
+++ b/src/driver/postgres/PostgresQueryRunner.ts
@@ -1328,7 +1328,6 @@ export class PostgresQueryRunner
 
         if (
             oldColumn.type !== newColumn.type ||
-            oldColumn.length !== newColumn.length ||
             newColumn.isArray !== oldColumn.isArray ||
             (!oldColumn.generatedType &&
                 newColumn.generatedType === "STORED") ||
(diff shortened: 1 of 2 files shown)


smith-xyz:

PR #12524, opened 2026-05-20

jspector-fr:
## Summary
- Fixes PostgreSQL `varchar` length changes so migration generation emits `ALTER COLUMN ... TYPE` instead of dropping and recreating the column.
- Keeps existing data when increasing a varchar length, addressing #3357.
- Adds a regression test covering the generated SQL and data preservation path.

## Test plan
- `corepack pnpm install --frozen-lockfile`
- `corepack pnpm exec prettier --write src/driver/postgres/PostgresQueryRunner.ts test/github-issues/3357/issue-3357.test.ts`
- `corepack pnpm exec tsc --noEmit --pretty false`

I could not run the PostgreSQL integration test locally because this environment has no local TypeORM `ormconfig.json`/Postgres test database configured.

/claim #3357


jspector-fr:
Updated the PR to address the compliance feedback: moved the #3357 regression coverage from `test/github-issues/3357` into the functional schema-builder column change suite, while keeping the Postgres-specific assertion that varchar length changes use `ALTER COLUMN ... TYPE` and preserve existing data.

I could not run the full TypeORM test suite in this environment because dependencies are not installed, but `git diff --check` passes.

PR #12531, opened 2026-05-21
fix: reload upsert results by conflict columns

### LINKED ISSUE #11516: Upsert failed when the record is updated instead of inserting, using on tables with id of number type.

puneetdixit200:
### Description of change

Fixes #11516.

For drivers without INSERT RETURNING support, TypeORM reloads inserted rows after insert/upsert so generated/default columns can be merged back into the result. When an upsert hits the update path and the entity does not include its generated primary key, that reload currently fails because it can only build criteria from the primary key.

This falls back to the upsert conflict columns when the primary key is not available, so updated rows can still be reloaded by the same unique columns that matched the conflict. Reloaded rows are now matched back by their reload criteria instead of result array position, so unordered or incomplete reload results cannot merge into the wrong entity.

### Pull-Request Checklist

- [x] Code is up-to-date with the `v0.3` branch
- [x] This pull request links relevant issues as `Fixes #11516`
- [x] There are new or updated tests validating the change (`tests/**/*.test.ts`)
- [x] Documentation has been updated to reflect this change (`docs/docs/**/*.md`) N/A, no public API change

Verification:
- `pnpm run compile`
- `pnpm exec mocha --no-config build/compiled/test/unit/query-builder/returning-results-entity-updator.test.js`
- `cp .github/workflows/test/sqljs.ormconfig.json ormconfig.json && pnpm exec mocha --no-config build/compiled/test/functional/repository/basic-methods/repository-basic-methods.test.js --grep "reload updated rows by conflict columns" && rm -f ormconfig.json`
- `pnpm exec prettier --check src/query-builder/ReturningResultsEntityUpdator.ts test/unit/query-builder/returning-results-entity-updator.test.ts`
- `pnpm exec eslint src/query-builder/ReturningResultsEntityUpdator.ts test/unit/query-builder/returning-results-entity-updator.test.ts` (warnings only, exit 0)
- `git diff --check`


puneetdixit200:
Rechecked the latest head after the follow-up commits. The reload path now maps reloaded rows by criteria instead of result index, rejects missing or duplicate reload matches, reloads generated-id-only conflict rows, normalizes criteria values, handles bigint keys, rejects SQL-expression criteria, and avoids putting criteria values in reload errors. The unit test helpers are typed without new `any` casts.

Local verification:
- `pnpm run compile` with Node 18.20.8
- `pnpm exec mocha --no-config --timeout 90000 --exit build/compiled/test/unit/query-builder/returning-results-entity-updator.test.js` -> 8 passing

puneetdixit200:
Rechecked the fork `Tests` workflow for this branch. The current PR head `446a03b` already has a successful fork run:

- current-head run: https://github.com/puneetdixit200/typeorm/actions/runs/26352770543
- upstream visible `semantic` check is also passing

I also reran the older failed fork run on `530cca4`; attempt 2 completed successfully as well, so the previous `detect-changes` setup failure is no longer present on the branch history.

dikshit-n:
Hi @puneetdixit2000 👋

This PR (reload upsert results by conflict columns — great fix for #11516!) has been open since May with no recent activity.

Is this still on your radar? Would love to help review or test if there's anything blocking it. If you've moved on, happy to take it over with your blessing.

PR #12533, opened 2026-05-22
fix: release queryRunner when migration setup fails

ATOM00blue:
### Description of change

`MigrationExecutor.showMigrations()`, `executePendingMigrations()` and `undoLastMigration()` each created a `QueryRunner` and then ran several `await` calls — `createMigrationsTableIfNotExist()`, `createMetadataTableIfNecessary()` and `loadExecutedMigrations()` — *before* entering their `try`/`finally` block. If any of those setup calls rejected (e.g. the database is unreachable, or creating the migrations/metadata table fails), the runner was never released and leaked from the connection pool. `showMigrations()` had no `try`/`finally` at all, so a failure anywhere in it leaked the runner too.

This PR routes all three methods through the existing `withQueryRunner()` helper, which already creates the runner, wraps the whole body in `try`/`finally`, and releases it only when the executor created it (`executeMigration()`, `getExecutedMigrations()`, `insertMigration()` and `deleteMigration()` already use it). The runner is now always released regardless of where a failure occurs. Transaction handling (start/commit/rollback) is unchanged.

#### Before

```ts
const queryRunner = this.queryRunner ?? this.dataSource.createQueryRunner()
await this.createMigrationsTableIfNotExist(queryRunner) // throws -> runner leaks
// ...
try { /* ... */ } finally { if (!this.queryRunner) await queryRunner.release() }
```

#### After

```ts
return this.withQueryRunner(async (queryRunner) => {
    await this.createMigrationsTableIfNotExist(queryRunner) // throws -> still released
    // ...
})
```

Verified with a new test that stubs `loadExecutedMigrations()` to throw and asserts the created runner is still released exactly once for each of the three methods. The test fails on `master` and passes with this change; the existing migration suites still pass.

### Pull-Request Checklist

- [x] This pull request links relevant issues as `Fixes #12243`
- [x] There are new or updated tests validating the change (`test/github-issues/12243/issue-12243.test.ts`)
- [ ] Documentation has been updated to reflect this change (`docs/docs/**.md`) — N/A, internal bug fix with no API/behavior change

Fixes #12243

### DIFF
--- a/src/migration/MigrationExecutor.ts
+++ b/src/migration/MigrationExecutor.ts
@@ -153,41 +153,38 @@ export class MigrationExecutor {
      * returns true if there are unapplied migrations
      */
     async showMigrations(): Promise<boolean> {
-        let hasUnappliedMigrations = false
-        const queryRunner =
-            this.queryRunner ?? this.dataSource.createQueryRunner()
-        // create migrations table if its not created yet
-        await this.createMigrationsTableIfNotExist(queryRunner)
+        return this.withQueryRunner(async (queryRunner) => {
+            let hasUnappliedMigrations = false
+            // create migrations table if its not created yet
+            await this.createMigrationsTableIfNotExist(queryRunner)
 
-        // get all migrations that are executed and saved in the database
-        const executedMigrations =
-            await this.loadExecutedMigrations(queryRunner)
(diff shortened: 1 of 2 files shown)


michaelbromley:

```
git fetch origin
```

PR #12535, opened 2026-05-22
fix(postgres): use ALTER COLUMN TYPE instead of DROP+ADD when only column length changes

q404365631:
Fixes #3357

## Problem
When a varchar column length was changed, TypeORM would generate DROP COLUMN + ADD COLUMN, causing data loss.

## Solution
1. Removed oldColumn.length check from DROP+ADD condition
2. Added length changes to ALTER condition
3. Fixed COLLATE undefined bug
4. Updated collation test expectations

## Before
DROP COLUMN + ADD COLUMN

## After
ALTER COLUMN TYPE


smith-xyz:
not a real solution

PR #12541, opened 2026-05-24
fix(postgres): use alter column for varchar length changes instead of dropping and adding

wbobbynmworley:
## Problem

When modifying a varchar/nchar column length in TypeORM, the migration generator incorrectly uses DROP+ADD, causing **data loss**.

```sql
-- Before: data loss
ALTER TABLE "user" ADD "name" character varying(100);
-- ALL DATA LOST!
```

## Root Cause

All 6 QueryRunner drivers had `oldColumn.length !== newColumn.length` in the drop+recreate condition, triggering column recreation instead of alteration.

### Postgres / CockroachDB / Spanner
- Remove the early length block that runs before rename
- Add length to the precision/scale ALTER COLUMN block (runs **after** rename, safe)

### MySQL / Oracle / AuroraMySQL
- Remove length from drop+recreate condition 
- Add length check to 
- Uses CHANGE/MODIFY syntax which preserves data

```sql
-- After: data preserved
ALTER TABLE "user" ALTER COLUMN "name" TYPE character varying(100);
```

## Changes

| Driver | Change |
|--------|--------|
| PostgresQueryRunner | Removed early block, merged into precision/scale ALTER |
| CockroachQueryRunner | Same pattern |
| SpannerQueryRunner | Same pattern |
| MysqlQueryRunner | Removed from drop+add, handled via isColumnChanged |
| OracleQueryRunner | Same |
| AuroraMysqlQueryRunner | Same |
| BaseQueryRunner | Added length to isColumnChanged() |

## Qodo Review Feedback Addressed

1. **Rename+length ordering**: Fixed — length change now runs after rename (uses oldColumn.name which is already updated)
2. **Template formatting**: Fixed — uses consistent single-line format

Fixes typeorm/typeorm#3357

### DIFF
--- a/.npmrc
+++ b/.npmrc
@@ -1,1 +1,0 @@
-node-linker=hoisted
--- a/.npmrc
+++ b/.npmrc
@@ -1,0 +1,1 @@
+node-linker=hoisted
--- a/src/driver/aurora-mysql/AuroraMysqlQueryRunner.ts
+++ b/src/driver/aurora-mysql/AuroraMysqlQueryRunner.ts
@@ -889,7 +889,6 @@ export class AuroraMysqlQueryRunner
             (newColumn.isGenerated !== oldColumn.isGenerated &&
                 newColumn.generationStrategy !== "uuid") ||
             oldColumn.type !== newColumn.type ||
-            oldColumn.length !== newColumn.length ||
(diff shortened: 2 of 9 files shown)


michaelbromley:

```
git fetch origin
```

PR #12542, opened 2026-05-24
fix: respect per-migration transaction setting on revert

### LINKED ISSUE #9981: --revert creates a wrapper transaction when using --transaction each
### Issue description

`--transaction each` does not work correctly when reverting a migration.

### Expected Behavior

The `--transaction each` flag should create a transaction for each migration, unless that migration sets `transaction = false` (see https://github.com/typeorm/typeorm/pull/9459). Creating a migration with `transaction = false`, and using a non-transactional statement (like Postgres's `CREATE INDEX CONCURRENTLY`) should work with `--transaction each`.

### Actual Behavior

It works when migrating up.

With `--revert`, it still creates a wrapper transaction:

```shell
**$ yarn -s typeorm migration:revert -d ./src/db/TypeORMDataSource.ts --transaction each
[INFO] 13:50:22 ts-node-dev ver. 1.1.9 (using ts-node ver. 9.1.1, typescript ver. 4.6.4)
query: SELECT * FROM current_schema()
query: CREATE EXTENSION IF NOT EXISTS "uuid-ossp"
query: SELECT version();
query: SELECT * FROM "information_schema"."tables" WHERE "table_schema" = 'public' AND "table_name" = 'migrations'
query: SELECT * FROM "migrations" "migrations" ORDER BY "id" DESC
8 migrations are already loaded in the database.
... is the last executed migration. It was executed on Mon Apr 24 2023 12:35:11 GMT+0000 (GMT).
Now reverting it...
query: START TRANSACTION
query: CREATE INDEX CONCURRENTLY "..." ON ... USING btree ("...")
# ...
error: error: CREATE INDEX CONCURRENTLY cannot run inside a transaction block
query: ROLLBACK
# ...
```

### Steps to reproduce

1. Create a migration.
2. Set `transaction = false`.
3. Add a statement that can't be run in a transaction in the `down` step (like `CREATE INDEX CONCURRENTLY` on Postgres).
4. Run the migration: `typeorm migration:run ... --transaction each`.
5. Try to revert it: `typeorm migration:revert ... --transaction each`.

### My Environment

| Node.js version     | 18.5.0  |
| Typescript version  | 4.6.4  |
| TypeORM version     | 0.3.12  |

### Additional Context

It looks like https://github.com/typeorm/typeorm/blob/68aa573262229898b3396598e2077efd451c1820/src/migration/MigrationExecutor.ts#L451-L455 is the problem here. If we just change that check to `=== "all"`, we won't (I don't think) get individual wrapper transactions for `transaction = true` migrations, so we'd need to copy / abstract out the remaining logic from https://github.com/typeorm/typeorm/pull/9459/files to handle this case correctly.

- [ ] aurora-mysql
- [ ] aurora-postgres
- [ ] better-sqlite3
- [ ] cockroachdb
- [ ] cordova
- [ ] expo
- [ ] mongodb
- [ ] mysql
- [ ] nativescript
- [ ] oracle
- [ ] postgres
- [ ] react-native
- [ ] sap
- [ ] spanner
- [ ] sqlite
- [ ] sqlite-abstract
- [ ] sqljs
- [ ] sqlserver

anipnwr7777:
### Description of change

Fixes #9981: Migration revert now respects `transaction = false`

When reverting migrations with `--transaction each`, migrations that set `transaction = false` will no longer be wrapped in a transaction. This allows non-transactional statements like `CREATE INDEX CONCURRENTLY` to work correctly during revert.

### Pull-Request Checklist
- [X] This pull request links relevant issues as `Fixes #00000`
- [X] There are new or updated tests validating the change (`tests/**.test.ts`)

### DIFF
--- a/src/migration/MigrationExecutor.ts
+++ b/src/migration/MigrationExecutor.ts
@@ -453,10 +453,11 @@ export class MigrationExecutor {
         )
         this.dataSource.logger.logSchemaBuild(`Now reverting it...`)
 
+        // defaults for single migration
         const txModeDefault = {
             each: true,
             none: false,
-            all: false,
+            all: true, // safe to set this to true for a single migration
         }[this.transaction]
 
         let migrationTransactionMode = txModeDefault
--- a/src/migration/MigrationExecutor.ts
+++ b/src/migration/MigrationExecutor.ts
@@ -453,9 +453,20 @@ export class MigrationExecutor {
         )
         this.dataSource.logger.logSchemaBuild(`Now reverting it...`)
 
-        // start transaction if its not started yet
+        const txModeDefault = {
+            each: true,
+            none: false,
+            all: false,
+        }[this.transaction]
+
+        let migrationTransactionMode = txModeDefault
+        if (migrationToRevert.instance?.transaction !== undefined) {
+            migrationTransactionMode = migrationToRevert.instance.transaction
+        }
+
+        // start transaction if its not started yet and the migration requires it
         let transactionStartedByUs = false
-        if (this.transaction !== "none" && !queryRunner.isTransactionActive) {
+        if (migrationTransactionMode && !queryRunner.isTransactionActive) {
             await queryRunner.startTransaction()
             transactionStartedByUs = true
         }
--- a/test/github-issues/9981/issue-9981.test.ts
+++ b/test/github-issues/9981/issue-9981.test.ts
@@ -1,0 +1,41 @@
+import "reflect-metadata"
+import {
+    closeTestingConnections,
+    createTestingConnections,
+    reloadTestingDatabases,
+} from "../../utils/test-utils"
+import type { DataSource } from "../../../src/data-source/DataSource"
+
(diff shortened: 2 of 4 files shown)


anipnwr7777:
Hi @gioboa, could you pls have a look at this PR!

anipnwr7777:
> All-mode revert non-transactional

While solving for this comment from Qodo, I realised that per migration overrides are not allowed when `tranactionMode: "all"` (https://github.com/typeorm/typeorm/blob/master/src/migration/MigrationExecutor.ts#L267) which makes sense in case of multiple migrations.

However this will not make any difference when there is a single migration which is true for `undoLastMigration()`, hence I have allowed overrides even when tranactionMode is "all". Let me know if this needs to be kept in sync with https://github.com/typeorm/typeorm/blob/master/src/migration/MigrationExecutor.ts#L267 - @alumni.

michaelbromley:

```
git fetch origin
```

PR #12543, opened 2026-05-24
fix(postgres): use ALTER COLUMN TYPE instead of drop and re-add on column type changes

Pulkit7070:
## Problem

Fixes #3357

When running `typeorm migration:generate`, if a PostgreSQL column changes its **type**, **length**, or **array** status, TypeORM generates destructive `DROP COLUMN` + `ADD COLUMN` SQL instead of the safe `ALTER COLUMN ... TYPE` equivalent.

**Before (destructive — causes data loss):**
```sql
```

**After (safe — preserves data):**
```sql
ALTER TABLE "bug" ALTER COLUMN "example" TYPE character varying(51) USING "example"::character varying(51)
```

## Root Cause

In `PostgresQueryRunner.changeColumn`, the condition at lines 1329–1343 treated **any** type/length/array change the same as a STORED generated column change and fell back to `dropColumn` + `addColumn`. This is correct for STORED generated columns (PostgreSQL can't alter their expression in-place), but wrong for ordinary type/length changes.

## Fix

The fix splits the condition into two separate branches:

1. **STORED generated columns** (unchanged behaviour): still drop and re-create, because PostgreSQL does not support altering a generated column's expression.

2. **Type/length/array changes** on regular columns: now generate `ALTER COLUMN ... TYPE <newType> USING "<col>"::<newType>`, which:
   - Preserves all existing data
   - Handles compatible type casts (e.g. `varchar(50)` → `varchar(51)`, `integer` → `bigint`)
   - Raises a meaningful runtime error for truly incompatible casts, rather than silently deleting data

The `USING` clause is the standard PostgreSQL mechanism for explicit type conversion during an `ALTER COLUMN TYPE` and is already used elsewhere in this file (e.g. for enum changes).

## Testing

- TypeScript compiles with no errors (`tsc --noEmit`)
- Linting and formatting pass (lint-staged hooks)
- Manual verification: generating a migration after changing a `varchar(50)` column to `varchar(51)` now emits `ALTER COLUMN ... TYPE` instead of `DROP + ADD`

/claim

### DIFF
--- a/src/driver/postgres/PostgresQueryRunner.ts
+++ b/src/driver/postgres/PostgresQueryRunner.ts
@@ -1344,12 +1344,28 @@ export class PostgresQueryRunner
                 oldColumn.length !== newColumn.length ||
                 newColumn.isArray !== oldColumn.isArray
             ) {
-                // Use ALTER COLUMN ... TYPE with a USING cast to avoid data loss.
-                // This is safe for type/length changes (e.g. varchar(50) -> varchar(51),
-                // integer -> bigint, etc.) and preserves existing data.
                 const newType = this.driver.createFullType(newColumn)
                 const oldType = this.driver.createFullType(oldColumn)
                 const columnName = oldColumn.name
+
+                // drop default before type cast to avoid cast errors, matching enum handling
+                if (
+                    oldColumn.default !== null &&
+                    oldColumn.default !== undefined
+                ) {
+                    defaultValueChanged = true
+                    upQueries.push(
+                        new Query(
+                            `ALTER TABLE ${this.escapePath(table)} ALTER COLUMN "${columnName}" DROP DEFAULT`,
+                        ),
+                    )
+                    downQueries.push(
+                        new Query(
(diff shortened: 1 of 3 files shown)


Pulkit7070:
Thanks for the review. Addressed the enum and array issues flagged by Qodo in dedb982.

**Issue 1 (enum type cast broken):** `createFullType()` produces the literal `"enum"` for `enum` / `simple-enum` columns rather than the user-defined PG type name, so the new `ALTER COLUMN ... TYPE enum USING col::enum` is invalid SQL.

**Issue 2 (array cast type invalid):** `createFullType()` appends `" array"` for `isArray` columns, and `::type array` casts between scalar and array forms aren't safe via a plain USING cast.

**Fix:** restrict the new ALTER-based path to ordinary scalar / length changes. Enum involvement (either side) and `isArray` toggling continue to use the original drop+add behavior, same as before this PR, so existing enum migrations and array-toggle paths are unchanged.

Scope of the PR is now narrowly the safe cases it was always intended to cover: `varchar(N)` length changes, `int` to `bigint`, and similar same-shape type widening.

michaelbromley:

```
git fetch origin
```

PR #12527, opened 2026-05-21
fix: alter column type without data loss

cianmag:
## Description

When a column's type or length changes (e.g., `varchar(250)` → `varchar(500)`), the `changeColumns` method in every query runner previously **dropped and recreated the column**, causing **complete data loss** for all existing rows.

This PR fixes the behavior to use `ALTER COLUMN TYPE` (or equivalent per-driver syntax) for type/length changes, preserving existing data. Drop+recreate is now reserved only for truly incompatible changes where no ALTER syntax exists.

## Root Cause

The `changeColumns` method in each query runner had a single condition combining:
- Type/length changes (safe to ALTER)
- Array/generateType/expression changes (require recreation)

Both were handled via `dropColumn()` + `addColumn()`, destroying data even when a non-destructive ALTER was possible.

## Fix

Split the condition into two branches:
1. **Drop+recreate**: Only for `isArray`, `generatedType`, `asExpression` changes
2. **ALTER COLUMN**: For `type` and `length` changes (data-preserving)

## Affected Drivers (all 8)

| Driver | ALTER Syntax Used |
|--------|------------------|
| PostgreSQL | `ALTER COLUMN ... TYPE` |
| MySQL | `CHANGE COLUMN` |
| Aurora MySQL | `CHANGE COLUMN` |
| CockroachDB | `ALTER COLUMN ... TYPE` |
| Oracle | `MODIFY` |
| SQL Server | `ALTER COLUMN` |
| SAP HANA | `ALTER COLUMN` |
| Spanner | `ALTER COLUMN ... TYPE` |

## Related Issues

- Fixes #9046 — varchar length change drops column
- Fixes #9254 — column length specified inline causes recreation  
- Fixes #9255 — varchar vs character varying type mismatch

## Testing

TypeScript compilation passes cleanly (`tsc --noEmit`). The existing `change-column.test.ts` covers the `queryRunner.changeColumn()` API path. The migration generator path (`schema:sync` / `migration:generate`) is tested through the `RdbmsSchemaBuilder` which calls `findChangedColumns` → `changeColumns`.

## Example

Before this fix:
```typescript
@Entity()
class User {
  @Column({ type: "varchar", length: 250 })
  name: string;
}

// Change to length: 500 → generated migration:
// DROP COLUMN "name"
// ADD COLUMN "name" varchar(500)  ← DATA LOST
```

After this fix:
```typescript
// Change to length: 500 → generated migration:
[...]

PR #12544, opened 2026-05-25
fix(postgres): use ALTER COLUMN TYPE for type/length changes with length-decrease guard (#3357)

lekt9:
## What

Fix #3357 — postgres `changeColumn` now uses `ALTER COLUMN TYPE … USING` for type and length changes that do not lose data, and explicitly guards the lossy length-decrease case behind a new dataSource option.

## Why this is different from the other open PRs

PRs #11620, #12527, #12532, #12535, #12541, #12543 all route some subset of type/length changes through `ALTER COLUMN TYPE` and stop there. None of them addresses the review concern @alumni left on #11620 on 2025-09-20:

> I think we should keep the old behavior (DROP + ADD) for quite a lot of situations, for both increasing and decreasing the length :) Maybe we can use it where we know it works… I would add a test (for all drivers) in which the column that is reduced contains data that is longer than the new size.

The competitors' shape silently changes behavior on length-decrease: the `varchar(50) → varchar(10)` migration with a 30-character row now throws a runtime error during the `USING` cast, where it used to silently delete the value. That is a strict regression even if the new behavior is morally better, and it does not match the maintainer's stated preference for preserving the old path "for quite a lot of situations."

This PR's shape:

| Change | Before | After (this PR) |
|---|---|---|
| `varchar(50) → varchar(100)` | DROP+ADD, data lost | `ALTER COLUMN TYPE`, **data preserved** |
| `text → varchar(255)` | DROP+ADD, data lost | `ALTER COLUMN TYPE … USING`, **data preserved** |
| `integer → bigint` | DROP+ADD, data lost | `ALTER COLUMN TYPE … USING`, **data preserved** |
| `varchar(50) → varchar(10)` (default config) | DROP+ADD, data lost | DROP+ADD, data lost, **plus a `logSchemaBuild` warning naming the new opt-in** |
| `varchar(50) → varchar(10)` (`migrationsAllowLossyAlter: true`) | DROP+ADD, data lost | `ALTER COLUMN TYPE … USING`, may truncate or fail per row |
| Array flip, STORED-generated promotion | DROP+ADD | DROP+ADD (unchanged) |

So default behavior is **strictly safer than today** (no rows are silently lost where they were not lost before), explicit opt-in unlocks the modern path, and `@alumni`'s review concern is satisfied because the historical path stays the default for length decreases.

## Scope

Postgres only. The new option `migrationsAllowLossyAlter` lives on `PostgresDataSourceOptions`; other drivers are untouched. A column-level `allowDataLoss` flag was considered for finer granularity but deliberately deferred to a follow-up PR to keep this one reviewable — happy to add it as a follow-up if maintainers prefer per-column control.

## How

- `src/driver/postgres/PostgresDataSourceOptions.ts`: new optional field `migrationsAllowLossyAlter?: boolean` with JSDoc explaining the trade-off.
- `src/driver/postgres/PostgresQueryRunner.ts` `changeColumn`:
  - Pre-classify the change into `isArrayChanged`, `isNewlyStoredGenerated`, `storedExpressionChanged`, `typeChanged`, `lengthChanged`, `lengthDecreases`, `allowLossyAlter`.
  - The original IF triggers DROP+ADD only for array flips and STORED-generated changes (cases postgres truly cannot ALTER in place).
  - A new ELSE-IF triggers DROP+ADD with a `logSchemaBuild` warning for length-decrease without opt-in, preserving historical behavior but explaining it.
  - The ELSE branch now emits `ALTER COLUMN TYPE … USING` at the top of its sequence, before the existing rename / unique / default / nullable handling, mirroring the precision/scale ALTER pattern already at lines ~1620.

## Tests

`test/functional/query-runner/change-column-postgres-safe-alter.test.ts` covers all five scenarios with real INSERT + SELECT round-trip assertions on data preservation, scoped via `enabledDrivers: ["postgres"]`. The length-decrease-without-opt-in case is the one `@alumni` named in the Sept 20 review.

- `pnpm tsc --noEmit` clean
- pre-commit eslint + prettier clean
- functional tests against real postgres assumed to pass; the testing-container postgres in the project's `docker-compose.yml` reproduces the scenarios

## Coordination

Tagging @pleerock (the bounty originator on opire.dev) and @alumni (the Sept 20 reviewer on #11620). Happy to coordinate a merged approach with @BenraouaneSoufiane on #11620 if maintainers want a unified path; this PR is intentionally narrow so it can land as the postgres slice while broader cross-driver work continues.

Refs: https://github.com/typeorm/typeorm/issues/3357, https://github.com/typeorm/typeorm/pull/11620#discussion_r2293247000

### DIFF
--- a/src/driver/postgres/PostgresDataSourceOptions.ts
+++ b/src/driver/postgres/PostgresDataSourceOptions.ts
@@ -106,4 +106,22 @@ export interface PostgresDataSourceOptions
      * List of additional Postgres extensions to be installed in the database.
      */
     readonly extensions?: string[]
+
+    /**
+     * Opt-in: allow schema migrations to issue ALTER COLUMN TYPE (with USING
+     * cast) when a column's length is decreased. The default is `false`, in
+     * which case the legacy DROP+ADD path is used and a warning is logged via
+     * the configured logger's `logSchemaBuild` channel.
+     *
+     * Set this to `true` if you accept that an ALTER COLUMN TYPE with a
+     * shorter length will truncate (or fail, depending on USING cast and the
+     * data) existing rows that exceed the new size.
+     *
+     * Length increases, precision/scale changes, and pure type changes are
+     * always issued via ALTER COLUMN TYPE; this option only affects the
+     * lossy length-decrease case.
+     *
+     * Default: `false`. Postgres only.
+     */
+    readonly migrationsAllowLossyAlter?: boolean
 }
--- a/src/driver/postgres/PostgresQueryRunner.ts
+++ b/src/driver/postgres/PostgresQueryRunner.ts
@@ -1326,22 +1326,105 @@ export class PostgresQueryRunner
                 `Column "${oldTableColumnOrName}" was not found in the "${table.name}" table.`,
             )
 
+        // Fix #3357 — pre-classify the change so length increases, type
+        // changes, and (opt-in) length decreases route through ALTER COLUMN
+        // TYPE instead of the destructive DROP+ADD path. The DROP+ADD path
+        // is preserved verbatim for changes Postgres cannot ALTER in place
+        // (isArray flip, STORED-generated column promotion or expression
+        // change) and for lossy length-decrease cases when the dataSource
+        // has not opted in via migrationsAllowLossyAlter.
+        const typeChanged = oldColumn.type !== newColumn.type
+        const lengthChanged = oldColumn.length !== newColumn.length
+        const isArrayChanged = newColumn.isArray !== oldColumn.isArray
+        const isNewlyStoredGenerated =
+            !oldColumn.generatedType && newColumn.generatedType === "STORED"
+        const storedExpressionChanged =
+            oldColumn.asExpression !== newColumn.asExpression &&
+            newColumn.generatedType === "STORED"
+        const oldLen = Number(oldColumn.length)
+        const newLen = Number(newColumn.length)
+        const lengthDecreases =
+            lengthChanged &&
+            oldColumn.length !== undefined &&
+            oldColumn.length !== "" &&
+            newColumn.length !== undefined &&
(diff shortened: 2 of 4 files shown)


lekt9:
Following up: this is a focused postgres-only take on #3357 that explicitly addresses @alumni's review concern on PR #11620 (length-decrease + overlong-data test, per the Sept 20 thread). Default behavior stays on DROP+ADD with a logged warning; the new `migrationsAllowLossyAlter` opt-in unlocks ALTER COLUMN TYPE for the lossy case. Happy to coordinate with @BenraouaneSoufiane on #11620 if maintainers prefer a unified path. Targeting ready-for-review within 24h.

lekt9:
/claim #3357

Hey, picking up #3357 as a focused postgres-only fix. I've read PR #11620 and the four newer racing PRs, and the gap none of them closes is the exact point alumni flagged on Sept 20: length-decrease safety with overlong data, and a per-driver test that proves it.

My plan:
1. Patch `PostgresQueryRunner.changeColumn` so type-only and length-only changes route through `ALTER COLUMN ... TYPE ... USING` instead of DROP+ADD.
2. Add an opt-in `allowDataLoss` flag on `ColumnOptions` plus a connection-level `migrationsAllowLossyAlter` switch. Length-decrease without the flag keeps the existing DROP+ADD path and logs an explicit warning, so default behavior stays safe.
3. Five functional tests covering varchar length up, varchar length down with overlong data (both flag states), numeric precision, and text-to-varchar narrowing.
4. Postgres scope only for this PR. Other drivers stay on the existing path; follow-up PRs can extend the pattern once this lands.

Targeting ready-for-review within 48 hours. Happy to coordinate with #11620's author if the maintainers want a merged approach instead of competing PRs.

Lewis
github.com/lekt9

michaelbromley:

```
git fetch origin
```

PR #12551, opened 2026-05-28
fix(query-runner): use ALTER/MODIFY column instead of DROP/ADD to prevent data loss

qikezhang:
Currently, when generating migrations or syncing schema, any change to column types, lengths, or array properties triggers a hard `DROP COLUMN` followed by an `ADD COLUMN` in drivers like Postgres, MySQL, CockroachDB, and Spanner to "avoid data conversion".

This approach results in catastrophic data loss on production schemas even for simple alterations (e.g., expanding varchar length or widening integer types).

This PR refactors the `changeColumn` logic across Postgres, MySQL, Spanner, and CockroachDB to bypass column recreation for these standard changes. Instead, it utilizes native and safe `ALTER COLUMN TYPE` / `CHANGE COLUMN` queries to perform in-place column alterations, completely eliminating the data loss risk while successfully passing all `changeColumn` test suites.

Closes #3357

### DIFF
--- a/src/driver/cockroachdb/CockroachQueryRunner.ts
+++ b/src/driver/cockroachdb/CockroachQueryRunner.ts
@@ -1402,13 +1402,10 @@ export class CockroachQueryRunner
             )
 
         if (
-            oldColumn.type !== newColumn.type ||
-            oldColumn.length !== newColumn.length ||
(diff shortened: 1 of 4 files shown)


alumni:
As mentioned in the issue #3357, we do not accept PRs for it at the moment

PR #12554, opened 2026-06-01
fix: correct grammar in AlreadyHasActiveConnectionError message

DucMinhNe:
Fixes a subject-verb agreement error in the `AlreadyHasActiveConnectionError` runtime message.

The thrown message read:

> ...because connection with such name **already exist** and it now has an active connection session.

The subject ("connection with such name") is singular, so it should read **already exists**. User-facing error string only; no behavior change.


Cprakhar (approved):

gioboa (approved):
Thanks @DucMinhNe

PR #12556, opened 2026-06-02
feat(postgres): per-endpoint extra pool configuration in replication mode

### LINKED ISSUE #12555: feat(postgres): allow per-endpoint extra pool configuration in replication mode
## Problem

When using `replication` mode with the Postgres driver, TypeORM passes the same top-level `extra` object to every `pg.Pool` constructor — both master (write) and slaves (read). There is no API to configure pool options differently per endpoint.

This is a real problem with Aurora read replicas: the reader endpoint is behind a proxy that silently kills connections after a fixed lifetime (e.g. after a fail-over or maintenance window). You want to set `maxLifetimeSeconds` **only on the slave pool** so pg-pool rotates its connections proactively, while keeping the writer pool unaffected.

## Current workaround

Because there is no API for this, we were forced to reach into pg-pool's private internals **after** `DataSource.initialize()`:

```ts
/**
 * typeorm shares a single `extra` between the writer and reader pools, so a reader-specific
 * lifetime cannot be set at pool-creation time. pg-pool reads `options.maxLifetimeSeconds`
 * afresh for every client as it connects, so mutating it here applies the lifetime to every
 * connection the reader pool subsequently opens. typeorm also opens one reader connection
 * during initialize() — before this runs — which therefore has no expiry timer, so we
 * additionally evict any already-idle reader connections.
 */
export function applyReaderConnectionMaxLifetime(
  dataSource: DataSource,
  maxLifetimeSeconds: number,
): boolean {
  const driver = dataSource.driver as unknown as { master: Pool; slaves?: Pool[] }
  const readerPool = driver.slaves?.[0]
  if (!readerPool) return false

  readerPool.options.maxLifetimeSeconds = maxLifetimeSeconds
  // evict idle connections that were opened before the lifetime was applied
  const internals = readerPool as unknown as {
    _idle?: { client: PoolClient }[]
    _remove?: (client: PoolClient) => void
  }
  if (Array.isArray(internals._idle) && typeof internals._remove === 'function') {
    for (const item of [...internals._idle]) {
      internals._remove(item.client)
    }
  }
  return true
}
```

This is fragile (it relies on pg-pool private fields `_idle` and `_remove`) and should not be necessary.

## Proposed fix

Add an `extra` field directly to `PostgresConnectionCredentialsOptions`. When present, it is merged **over** the top-level `extra` during pool creation, so each master/slave endpoint can carry its own pool overrides.

### `src/driver/postgres/PostgresConnectionCredentialsOptions.ts`

```ts
/**
 * Extra pg pool options to apply to this specific endpoint only.
 * Merged over the top-level `extra` in the DataSource options.
 * Has no effect outside of `replication` mode.
 */
readonly extra?: Record<string, unknown>
```

### `src/driver/postgres/PostgresDriver.ts` — `createPool`

```ts
const connectionOptions = Object.assign(
    {},
    {
        connectionString: credentials.url,
        host: credentials.host,
        // ... existing fields ...
        max: options.poolSize,
    },
    options.extra ?? {},
    credentials.extra ?? {},   // ← merge per-endpoint overrides on top
)
```

`connect()` requires **no changes** — `slave` and `master` are already the credential objects, so the override flows automatically.

### Usage

```ts
DataSource({
  type: 'postgres',
  extra: { idleTimeoutMillis: 10_000 },          // applies to both pools
  replication: {
    master: { host: 'writer.example.com', ... },
    slaves: [{
      host: 'reader.example.com',
      extra: { maxLifetimeSeconds: 60 },          // reader-only override
      ...
    }],
  },
})
```

## Why not top-level `masterExtraConfiguration` / `slavesExtraConfiguration`?

A top-level pair of fields would also work and be simpler to implement, but embedding `extra` directly in the credentials object is more consistent (it follows the same shape as host/port/ssl) and is more granular (each slave can differ). It also avoids polluting the top-level options namespace.

## Backward compatibility

Fully backward-compatible. If `credentials.extra` is absent (the common case), `credentials.extra ?? {}` is an empty object and the merge is a no-op. All existing behaviour is preserved.

slukes:
## Problem

When using `replication` mode with the Postgres driver, TypeORM passes the same top-level `extra` object to **every** `pg.Pool` constructor — both master and all slaves. There is no way to configure pool options differently per endpoint.

This is a real operational problem with Aurora read replicas and similar setups: the reader endpoint is behind a proxy that can silently drop connections after a fixed lifetime (e.g. during a fail-over or maintenance window). The standard solution is to set `maxLifetimeSeconds` **only** on the slave pool so pg-pool rotates its connections proactively — but TypeORM gives no API to do this without hacks.

## Workaround required today

Without this feature, users are forced to reach into pg-pool's **private internals** post-initialization:

```ts
// Had to do this because typeorm shares a single `extra` between writer and reader pools
const readerPool = (dataSource.driver as any).slaves?.[0]
readerPool.options.maxLifetimeSeconds = 60          // mutate private option
// also evict already-idle connections via _idle / _remove internals
```

Closes #12555.

## Solution

Add an optional `extra` field to `PostgresConnectionCredentialsOptions`. In `createPool`, it is merged **over** the top-level `extra`, so each master/slave endpoint can carry its own pool overrides:

```diff
// src/driver/postgres/PostgresConnectionCredentialsOptions.ts
+    readonly extra?: Record<string, unknown>

// src/driver/postgres/PostgresDriver.ts — createPool()
     const connectionOptions = Object.assign(
         {},
         { /* existing fields */ },
         options.extra ?? {},
+        credentials.extra ?? {},
     )
```

`connect()` requires **no changes** — `slave` and `master` are already the credential objects.

## Usage

```ts
new DataSource({
    type: 'postgres',
    extra: { idleTimeoutMillis: 10_000 },      // applies to both pools
    replication: {
        master: { host: 'writer.example.com', ... },
        slaves: [{
            host: 'reader.example.com',
            extra: { maxLifetimeSeconds: 60 },  // reader-only override
            ...
        }],
    },
})
```

## Backward compatibility

Fully backward-compatible. `credentials.extra ?? {}` is a no-op when the field is absent, so all existing behavior is preserved.

## Changes

- `src/driver/postgres/PostgresConnectionCredentialsOptions.ts` — add `extra?: Record<string, unknown>`
- `src/driver/postgres/PostgresDriver.ts` — one line: merge `credentials.extra` in `createPool`
- Replication test suite — three new integration tests verifying `credentials.extra` merging via `application_name` / `current_setting()`


src/driver/postgres/PostgresDataSourceOptions.ts
```diff
@@ -41,12 +41,24 @@ export interface PostgresDataSourceOptions
         /**
          * Master server used by orm to perform writes.
          */
-        readonly master: PostgresConnectionCredentialsOptions
+        readonly master: PostgresConnectionCredentialsOptions & {
```
Cprakhar: Instead of this, we can add a `readonly extra?: any` field in `PostgresConnectionCredentialsOptions`. Like you proposed in the PR description solution.

Cprakhar (changes requested):

src/driver/postgres/PostgresDataSourceOptions.ts
slukes: Done — added `readonly extra?: Record<string, unknown>` directly to `PostgresConnectionCredentialsOptions` and dropped the `(credentials as { extra?: ... })` cast in `createPool`. The field now has first-class typing and full autocomplete, matching the approach described in the PR description.

slukes:
> **Item 4 — `credentials` type assertion in `createPool`**

Valid — addressed in the latest push. Added `readonly extra?: Record<string, unknown>` directly to `PostgresConnectionCredentialsOptions`, so `createPool` can now use `credentials.extra ?? {}` without any cast.

> **Item 5 — Docs missing `new` keyword**

Valid — fixed in the latest push. The replication example now reads `new DataSource({...})`.

Cprakhar (approved):

slukes:
@Cprakhar Addressed your feedback — added `readonly extra?: any` directly to `PostgresConnectionCredentialsOptions` and removed the type-cast in `createPool`. All CI checks are now green. Could you take another look when you get a chance?

michaelbromley:

```
git fetch origin
```

PR #12557, opened 2026-06-02
feat(mysql): add per-column changeStrategy for safe schema migrations

smith-xyz:
### Description of change

This gives MySQL/MariaDB users per-column control over how type/length changes are applied during synchronize and migration:generate. Instead of always using the destructive DROP+ADD path, users can set changeStrategy on @Column() (or entity schema) to choose:

- auto — uses classification only (no live DB queries): ALTER COLUMN CHANGE for widen/no-change, DROP+ADD for narrow/incompatible, with warnings on risky paths
- alter — always uses CHANGE, warns on narrowings and incompatible changes (DB may reject or truncate)
- drop-add — preserves legacy behavior (default); silent data loss on type/length changes

Partially resolves https://github.com/typeorm/typeorm/issues/3357

Remaining drivers could follow if this approach works. I scoped this to MySQL/MariaDB first rather than a large sweeping change.

### Pull-Request Checklist

NOTE: will follow up with a pr on docs once accepted.


smith-xyz:
I see qodo's point on the sql executing inline with migrations - I think its safe to also just deterministically decide whether its a drop+add vs alter - so adjusted it

in the end, still just gives user's the control on these scenarios and provide them info of what could happen given the sql generated

michaelbromley:

```
git fetch origin
```

PR #12560, opened 2026-06-03
feat(join): evaluate soft-delete condition at query-build time for joins

### LINKED ISSUE #12571: withDeleted() does not work for joined entities when called after join methods
### Issue description

When using `QueryBuilder` with entities that have `@DeleteDateColumn()`, calling `.withDeleted()` after `.leftJoinAndSelect()` (or any other join method) has no effect on the joined entities. The soft-delete condition (`deletedAt IS NULL`) is still added to the `JOIN ON` clause.  This happens because the soft-delete condition is evaluated at **join-creation time** (inside the `join()` method), not at **query-build time**. Since `.withDeleted()` is typically chained after the joins, the `withDeleted` flag is still `false` when the condition is set.  Additionally, there is no way to include soft-deleted rows for a **specific join** without affecting all other joins in the query.

### Expected Behavior

1. `.withDeleted()` should work regardless of call order — placing it before or after `.leftJoinAndSelect()` should produce the same result.
2. There should be a way to include soft-deleted rows on a per-join basis.

### Actual Behavior

```typescript
// This does NOT include soft-deleted conversations — the condition was already baked in
const users = await dataSource
    .createQueryBuilder(User, "user")
    .leftJoinAndSelect("user.conversations", "conversation")
    .withDeleted()  // <-- has no effect on the join above
    .getMany()
```

The generated SQL still contains `conversation.deletedAt IS NULL` in the `JOIN ON` clause, even though `.withDeleted()` was called.

## Steps To Reproduce

1. Create two entities with a relation, where the joined entity has `@DeleteDateColumn()`:

```typescript
@Entity()
export class User {
    @PrimaryGeneratedColumn()
    id: number

    @OneToMany(() => Conversation, (c) => c.user)
    conversations: Conversation[]
}

@Entity()
export class Conversation {
    @PrimaryGeneratedColumn()
    id: number

    @Column()
    title: string

    @DeleteDateColumn()
    deletedAt: Date | null

    @ManyToOne(() => User, (u) => u.conversations)
    user: User
}
```

2. Create some conversations, then soft-delete one of them.
3. Query using `QueryBuilder` with `.withDeleted()` after the join:

```typescript
const users = await dataSource
    .createQueryBuilder(User, "user")
    .leftJoinAndSelect("user.conversations", "conversation")
    .withDeleted()
    .getMany()
```

4. Observe that soft-deleted conversations are **not** included in the results.

### My Environment

| Operating System    | ubuntu  and windows     |
| Node.js version    | 22.14.0            |
| Typescript version | 5.9.3              |
| TypeORM version    | latest master      |

### Additional Context

PR with fix: #12560

The fix involves two changes:

1. **Move soft-delete condition evaluation** from `join()` to `createJoinExpression()` so it is applied at query-build time, when the `withDeleted` flag has its final value.
2. **Add `includeDeleted?: boolean` parameter** to `leftJoin()`, `leftJoinAndSelect()`, `innerJoin()`, `innerJoinAndSelect()` for per-join control of the soft-delete filter:

```typescript
// Include soft-deleted conversations, but still filter soft-deleted messages
const users = await dataSource
    .createQueryBuilder(User, "user")
    .leftJoinAndSelect("user.conversations", "conversation", undefined, undefined, true)
    .leftJoinAndSelect("conversation.messages", "message")
    .getMany()
```

- [ ] aurora-mysql
- [ ] aurora-postgres
- [ ] better-sqlite3
- [ ] capacitor
- [ ] cockroachdb
- [ ] cordova
- [ ] expo
- [ ] mongodb
- [x] mysql
- [ ] nativescript
- [ ] oracle
- [x] postgres
- [ ] react-native
- [ ] sap
- [ ] spanner
- [ ] sqlite
- [ ] sqljs
- [ ] sqlserver

JulioVianaDev:
## Motivation

In some of my projects, I consistently run into issues caused by relying solely on QueryBuilder for more complex queries. One of the main problems is that it becomes impossible to perform LEFT JOINs on entities that use a DeletedDateColumn for soft deletes.

As a result, we often spend a significant amount of time dealing with limitations for which there are currently no mechanisms or annotations that allow us to control this behavior. The available workarounds are usually either writing raw SQL queries or removing the DeletedDateColumn altogether. However, in large enterprise projects, both approaches can be harmful in terms of maintainability, consistency, and long-term scalability.

For scenarios like these, I would like to propose the following solution implemented in the PR.

## Summary

- **Fix 1**: \`withDeleted()\` now works regardless of call order. The soft-delete condition on joins is evaluated at query-build time instead of join-creation time.
- **Fix 2**: Added per-join \`withDeleted\` control via \`leftJoinAndSelectWithDeleted()\`, \`leftJoinWithDeleted()\`, \`innerJoinAndSelectWithDeleted()\`, \`innerJoinWithDeleted()\`.

## Problem

When using QueryBuilder with entities that have \`@DeleteDateColumn()\`, the \`deletedAt IS NULL\` condition was baked into the JOIN ON clause at the time \`.leftJoinAndSelect()\` was called. Since \`.withDeleted()\` is typically chained after the joins, the flag was still \`false\` when the condition was set — making it impossible to include soft-deleted rows from joined entities.

## Changes

- \`src/query-builder/SelectQueryBuilder.ts\`: moved soft-delete condition from \`join()\` to \`createJoinExpression()\`; added 4 new \`*WithDeleted()\` methods
- \`src/query-builder/JoinAttribute.ts\`: added \`withDeleted\` property

## Usage

\`\`\`typescript
// Fix 1: withDeleted() now works in any position
dataSource.createQueryBuilder(User, \"user\")
    .leftJoinAndSelect(\"user.conversations\", \"conversation\")
    .withDeleted()
    .getMany()

// Fix 2: per-join control
dataSource.createQueryBuilder(User, \"user\")
    .leftJoinAndSelectWithDeleted(\"user.conversations\", \"conversation\")
    .leftJoinAndSelect(\"conversation.messages\", \"message\")
    .getMany()
\`\`\`

## Test plan

- [x] 9 new tests covering both fixes
- [x] 2949 total tests passing, 0 failures
- [x] No regressions on existing soft-delete, join strategy, or QueryBuilder tests

Fixes #12571


src/query-builder/SelectQueryBuilder.ts
```diff
@@ -648,6 +648,123 @@ export class SelectQueryBuilder<Entity extends ObjectLiteral>
...
+            alias,
+            condition,
+            parameters,
+            undefined,
+            undefined,
+            undefined,
+            true,
+        )
+        return this
+    }
```
Cprakhar: `WithDeleted()` join variants need to have overload signatures like `leftJoin`, `innerJoinAndSelect`, etc.

Cprakhar (changes requested):

Cprakhar:
Instead of introducing new methods, maybe add an optional param `includeDeleted?: bool`.

JulioVianaDev:
changes fixed for this implementation, i added as parameter inside the leftJoin ,etc

### Before (removed)

```typescript
.leftJoinAndSelectWithDeleted("user.conversations", "conversation")
.leftJoinWithDeleted("user.conversations", "conversation")
.innerJoinAndSelectWithDeleted("user.conversations", "conversation")
.innerJoinWithDeleted("user.conversations", "conversation")
```

### After (new API)

```typescript
.leftJoinAndSelect("user.conversations", "conversation", undefined, undefined, true)
.leftJoin("user.conversations", "conversation", undefined, undefined, true)
.innerJoinAndSelect("user.conversations", "conversation", undefined, undefined, true)
.innerJoin("user.conversations", "conversation", undefined, undefined, true)
```

Cprakhar:
@JulioVianaDev, could I ask you to create an issue for this feature request first? It will be easier for us to track PR and issues later.

Also, remove the redundant comment flagged by qodo.

Later Edit: ~~Also, remove the redundant comment flagged by qodo.~~

JulioVianaDev:
Thanks fow reports @Cprakhar , now i create the pr with this description bug into the issues and vinculate the pr.

Cprakhar:
Thinking on the current impl, it's not easy to use at the moment, to include the deleted columns in the join we are doing:
```ts
.leftJoin("user.conversations", "conversation", undefined, undefined, true) // need to pass 'undefined' to `condition` and `parameters` optional params to access `includeDeleted` param
```

JulioVianaDev:
normally i use keys in all my projects because is more easier than use params position because keys optional is free {withDeleted?:boolean,..otherMethods} but how i dont want change how typeorm interacts, i tried first create one method for this specific, but https://github.com/typeorm/typeorm/pull/12560#issuecomment-4672993971 in this part i think sayed to me make this way instead create one new method, and i have little afraid to change how typeorm internally interacts with the params provided into leftJoinAndSelect methods. i dont have any other idea from this 3 ways

JulioVianaDev:
@Cprakhar after thinking more about i can put in this method to be more easier to use:

## Proposed Solution

Make `withDeleted()` **chainable after any join method** so it applies only to the preceding join:

```typescript
// Per-join: only "conversation" includes soft-deleted rows
const users = await dataSource
    .createQueryBuilder(User, "user")
    .leftJoinAndSelect("user.conversations", "conversation")
    .withDeleted()
    .leftJoinAndSelect("conversation.messages", "message")
    .getMany()
```

### Behavior

| Call position | Effect |
|---|---|
| `.withDeleted()` **before** any join | Global — disables soft-delete filter on all joins (backward compatible) |
| `.leftJoinAndSelect(...).withDeleted()` | Per-join — disables soft-delete filter only on the preceding join |

### Multiple joins example

```typescript
// Both joins include soft-deleted rows
const users = await dataSource
    .createQueryBuilder(User, "user")
    .leftJoinAndSelect("user.conversations", "conversation")
    .withDeleted()
    .leftJoinAndSelect("conversation.messages", "message")
    .withDeleted()
    .getMany()

// Only the second join includes soft-deleted rows
const users = await dataSource
    .createQueryBuilder(User, "user")
    .leftJoinAndSelect("user.conversations", "conversation")
    .leftJoinAndSelect("conversation.messages", "message")
    .withDeleted()
    .getMany()
```

michaelbromley:

```
git fetch origin
```

PR #12564, opened 2026-06-05
fix(migration): preserve data when altering column type/length

pranavs321:
## Description
Fixes a critical data loss bug where TypeORM's migration generation would DROP and CREATE columns instead of using ALTER statements when column properties like type or length changed.

## Problem
- Schema synchronization with column type/length changes triggered DROP COLUMN + ADD COLUMN
- This permanently deleted existing data in affected columns
- Affected all drivers: PostgreSQL, MySQL, Aurora MySQL, CockroachDB, Oracle, SQL Server, Spanner

## Solution
- Column type/length changes now use ALTER TABLE ... ALTER/MODIFY COLUMN
- Data is preserved during migration
- DROP/CREATE only happens for changes that require it (e.g., IDENTITY/generated columns)

## Fixes
- Migration generation drops and creates columns instead of altering resulting in data loss


alumni:
SEE https://github.com/typeorm/typeorm/issues/3357#issuecomment-4638034647

PR #12565, opened 2026-06-06
fix(postgres): use alter column type instead of drop and add for type length isarray changes

Purkulosh:
## What
Changes migration generation to use `ALTER COLUMN ... TYPE` instead of `DROP COLUMN` + `ADD COLUMN` when a column's type, length, or `isArray` changes in PostgreSQL.

## Why
Previously, when changing a column type/length (e.g. `varchar(50)` → `varchar(51)`), TypeORM would generate:
```sql
```
This causes **data loss** because the column is dropped and recreated empty.

The correct behavior is:
```sql
```

## How
- Removed `type`, `length`, and `isArray` from the DROP+ADD condition in `PostgresQueryRunner.changeColumn()`
- Added new `ALTER COLUMN ... TYPE` handling for these cases inside the safe `else` block
- Uses existing `this.driver.createFullType()` which correctly constructs the full type string including length (e.g. `character varying(51)`)

## Example
Before this fix, for `@Column({ length: 50 })` → `@Column({ length: 51 })`:
```sql
ALTER TABLE "bug" DROP COLUMN "example"     -- DATA LOSS!
ALTER TABLE "bug" ADD "example" varchar(51)
```

After this fix:
```sql
ALTER TABLE "bug" ALTER COLUMN "example" TYPE varchar(51)  -- preserves data
```

## Checklist
- [x] Fix compiled successfully (`pnpm run compile`)
- [x] Uses existing `createFullType()` helper for full type construction
- [x] Backward compatible — required column changes still work
- [x] Follows same pattern as existing precision/scale ALTER COLUMN handling

Fixes #3357

### DIFF
--- a/src/driver/postgres/PostgresQueryRunner.ts
+++ b/src/driver/postgres/PostgresQueryRunner.ts
@@ -1621,19 +1621,24 @@ export class PostgresQueryRunner
                 newColumn.isArray !== oldColumn.isArray
             ) {
                 // Use buildEnumName() for enum/simple-enum types since createFullType()
-                // would produce invalid "TYPE enum" / "TYPE simple-enum" SQL
-                const isEnumType =
-                    oldColumn.type === "enum" ||
-                    oldColumn.type === "simple-enum" ||
+                // would produce invalid "TYPE enum" / "TYPE simple-enum" SQL.
+                // Check each side independently: only use buildEnumName for the side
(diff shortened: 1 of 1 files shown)


alumni:
Hi @Purkulosh, as mentioned in the comments for #3357, we will not accept PRs that try to fix that issue.

PR #12566, opened 2026-06-08
fix: export DateUtils from the package entry point

### LINKED ISSUE #3948: Export DateUtils in typeorm module
**Issue type:**

[ ] question
[ ] bug report
[X] feature request
[ ] documentation issue

**Database system/driver:**

[ ] `cordova`
[ ] `mongodb`
[ ] `mssql`
[ ] `mysql` / `mariadb`
[ ] `oracle`
[ ] `postgres`
[ ] `cockroachdb`
[ ] `sqlite`
[ ] `sqljs`
[ ] `react-native`
[ ] `expo`

**TypeORM version:**

[X ] `latest`
[ ] `@next`

It would be very useful to export DateUtils in the typeorm module in order to aid in raw sql queries. Raw queries are particularly useful for me, because I am in the gradual process of moving sql tables into TypeORM, but it's just not feasible (or CI/CD-approach) to do it all in one release of my app. If I should elaborate more, please let me know.

MahinAnowar:
## What

Closes #3948.

`DateUtils` (`src/util/DateUtils.ts`) is used to format/normalize dates for raw SQL queries, but unlike its sibling utilities `InstanceChecker` and `TreeRepositoryUtils` it was never re-exported from the package entry point (`src/index.ts`).

Consumers therefore had to deep-import `typeorm/util/DateUtils`. As the issue reporter notes, that path is no longer reliable now that TypeORM ships **pre-bundled** (`main`/`types` point at `./index.js` / `./index.d.ts`), so internal paths aren't guaranteed stable. A maintainer asked for a PR on the issue.

## Fix

Add one line to `src/index.ts`, mirroring the existing util re-exports:

```ts
export { InstanceChecker } from "./util/InstanceChecker"
export { TreeRepositoryUtils } from "./util/TreeRepositoryUtils"
export { DateUtils } from "./util/DateUtils"
```

`import { DateUtils } from "typeorm"` now works.

## Tests

Added `test/unit/util/date-utils.test.ts` (pure unit test, no DB) asserting that:
- `DateUtils` is exported from the package entry point and is the same reference as the direct module export;
- its static helpers are reachable through the public export (`mixedDateToDateString`).

Verified the test fails to compile on the previous code (`TS2305: Module '"../../../src"' has no exported member 'DateUtils'`) and passes after the fix. `pnpm run compile`, Prettier, and ESLint are all clean.


MahinAnowar:
On the docs flag from the bot — I looked at whether the two sibling util exports are documented, and neither is: `InstanceChecker` and `TreeRepositoryUtils` both get zero hits anywhere under `docs/`. So this PR is consistent with how the existing util re-exports are treated, and adding a docs page just for `DateUtils` would put it ahead of its siblings rather than in line with them.

Happy to write something if you'd rather have all three documented together, but that felt like a separate call for a maintainer to make, not something to fold into a one-line export.

alumni:
> So this PR is consistent with how the existing util re-exports are treated

That's fine - I would not document such things. These are rather internal APIs - even though we export them for ease-of-use, we should encourage people to use them actively, especially `DateUtils`, since we'd like to review the way dates are handled in the future.

PR #12569, opened 2026-06-08
fix(postgres): use ALTER COLUMN TYPE for compatible string type/length changes

TakoVHS:
## Description of change

Fixes #3357.

When a `character varying` / `varchar` / `character` / `char` / `text`
column changes its type or length, TypeORM's PostgreSQL driver currently
drops and re-adds the column:

```typescript
// To avoid data conversion, we just recreate column
await this.dropColumn(table, oldColumn)
await this.addColumn(table, newColumn)
```

This silently destroys all existing row data — a data-loss bug that
affects anyone who uses `synchronize()` or `migration:generate` to
widen a `varchar` length.

## What this PR does

Introduces `isAlterColumnTypeCompatible()`, which returns `true` when:

1. Both types are in the PostgreSQL string-type family, **and**
2. The change is a **widening** (not a narrowing).

When both conditions hold, `changeColumn()` emits:

```sql
ALTER TABLE t ALTER COLUMN c TYPE character varying(N)
```

instead of DROP + ADD. PostgreSQL handles implicit casts within the
string family without a `USING` clause — no data is at risk.

## Narrowing guard (addresses #11620 review concern)

In PR #11620, @alumni noted (2025-09-20):

> I think we should keep the old behavior (DROP + ADD) for quite a lot
> of situations, for both increasing and decreasing the length …
> I would add a test in which the column that is reduced contains data
> that is longer than the new size.

This PR respects that guidance exactly:

| Change | Behaviour |
|---|---|
| `varchar(50) → varchar(100)` | **ALTER COLUMN TYPE** — data preserved ✓ |
| `varchar(N) → text` | **ALTER COLUMN TYPE** — widening to unlimited ✓ |
| `varchar(100) → varchar(50)` | **DROP + ADD** — historical behaviour kept |
| `text → varchar(N)` | **DROP + ADD** — narrowing from unlimited kept |
| Array-membership change | **DROP + ADD** — unchanged |
| STORED generated column | **DROP + ADD** — unchanged |
| Type outside string family | **DROP + ADD** — unchanged |

## What changed

| File | Change |
|---|---|
| `src/driver/postgres/PostgresQueryRunner.ts` | Nested `if` inside the recreate block; new `isAlterColumnTypeCompatible()` method with widening guard |
| `test/…/change-column/change-column.test.ts` | New test: inserts a row, widens `varchar(255)→varchar(512)`, asserts data survived |

## Pull-Request Checklist

- [x] There are new or updated tests validating the change

> Note: This PR targets PostgreSQL only. A parallel effort for
> MySQL/MariaDB is tracked in #12557.

### DIFF
--- a/src/driver/postgres/PostgresQueryRunner.ts
+++ b/src/driver/postgres/PostgresQueryRunner.ts
@@ -1335,12 +1335,46 @@ export class PostgresQueryRunner
             (oldColumn.asExpression !== newColumn.asExpression &&
                 newColumn.generatedType === "STORED")
         ) {
-            // To avoid data conversion, we just recreate column
-            await this.dropColumn(table, oldColumn)
-            await this.addColumn(table, newColumn)
+            if (
+                (oldColumn.type !== newColumn.type ||
+                    oldColumn.length !== newColumn.length) &&
+                newColumn.isArray === oldColumn.isArray &&
+                !newColumn.generatedType &&
+                this.isAlterColumnTypeCompatible(oldColumn, newColumn)
+            ) {
+                // Use ALTER COLUMN … TYPE to preserve existing data.
+                // This avoids the destructive DROP + ADD path for compatible
+                // type-family or length-only changes (e.g. varchar(50) → varchar(100)).
+                upQueries.push(
+                    new Query(
+                        `ALTER TABLE ${this.escapePath(table)} ALTER COLUMN "${
+                            oldColumn.name
+                        }" TYPE ${this.driver.createFullType(newColumn)}`,
+                    ),
+                )
+                downQueries.push(
+                    new Query(
+                        `ALTER TABLE ${this.escapePath(table)} ALTER COLUMN "${
+                            newColumn.name
+                        }" TYPE ${this.driver.createFullType(oldColumn)}`,
+                    ),
+                )
+                // Reflect the new type/length in the in-memory table cache.
+                const clonedColumn = clonedTable.columns.find(
+                    (c) => c.name === oldColumn.name,
+                )
+                if (clonedColumn) {
+                    clonedColumn.type = newColumn.type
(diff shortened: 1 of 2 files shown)


TakoVHS:
### Differentiation from other open PRs on #3357

The bot flagged four other PRs (#12532, #12541, #12543, #12544). Here is how this PR differs:

**The key gap in the other PRs:** @alumni left this review note on #11620 (2025-09-20):

> I think we should keep the old behavior (DROP + ADD) for quite a lot of situations, for both increasing and decreasing the length … I would add a test in which the column that is reduced contains data that is longer than the new size.

PRs #12532, #12541, and #12543 all route length changes through `ALTER COLUMN TYPE` without distinguishing widening from narrowing. That means `varchar(100) → varchar(50)` now raises a runtime error if any row exceeds 50 characters — a different kind of failure where before there was silent deletion. This contradicts the maintainer's stated preference to preserve DROP+ADD for the decreasing-length case.

PR #12544 does address the narrowing concern via a new `migrationsAllowLossyAlter` option, but was labeled `Compliance violation` by the triage bot.

**This PR's shape:**

| Change | Behaviour |
|---|---|
| `varchar(N) → varchar(M)`, M ≥ N | `ALTER COLUMN TYPE` — data preserved ✓ |
| `varchar(N) → text` | `ALTER COLUMN TYPE` — widening to unlimited ✓ |
| `varchar(N) → varchar(M)`, M < N | `DROP + ADD` — historical behaviour kept ✓ |
| `text → varchar(N)` | `DROP + ADD` — narrowing from unlimited kept ✓ |
| Array / generated / out-of-family change | `DROP + ADD` — unchanged ✓ |

No new config options. No `USING` clause (not needed for implicit string-family casts). Conservative scope: PostgreSQL only, string family only, widening only.

TakoVHS (commented):
ok

PR #12576, opened 2026-06-11
fix: preserve Buffer values in normalizeWhereCriteria

### LINKED ISSUE #12574: Buffer and Array parameters are not passed correctly to the DB clients when `invalidWhereValuesBehavior` is enabled
### Issue description

Any Buffer parameter is incorrectly copied and passed to DB client as a plain object instead of a Buffer value.

### Expected Behavior

Buffer values should be passed correctly to the DB client.

### Actual Behavior

The Buffer values are converted to a plain object, which most DB clients don't accept.

### Steps to reproduce

```ts
export class UserEntry {
  @Column()
  guid: Buffer;
}

const guid = Buffer.from(crypto.randomUUID().replaceAll('-', ''), 'hex');

createQueryBuilder()
  .insert()
  .into(UserEntity)
  .values({ guid })
  .execute();
```

### My Environment

| Node.js version     | 24.16.0  |
| Typescript version  | 5.9.x  |
| TypeORM version     | 0.3.30  |

### Additional Context

Introduced by #11878 in `OrmUtils.normalizeWhereCriteria`.

- [ ] aurora-mysql
- [ ] aurora-postgres
- [ ] better-sqlite3
- [ ] capacitor
- [ ] cockroachdb
- [ ] cordova
- [ ] expo
- [ ] mongodb
- [ ] mysql
- [ ] nativescript
- [ ] oracle
- [ ] postgres
- [ ] react-native
- [ ] sap
- [ ] spanner
- [ ] sqlite
- [ ] sqljs
- [ ] sqlserver

handsomesix:
## Summary

Fixes #12574.

Buffer parameters passed to TypeORM queries were being converted to plain objects, causing DB clients to reject them.

## Root Cause

`OrmUtils.normalizeWhereCriteria` recursively processes all object values as nested criteria. Since `typeof Buffer` is `'object'` and Buffer is not an Array, Date, or FindOperator, it was treated as nested criteria and its byte-index properties were copied into a plain object.

## Changes

- Added `!isUint8Array(value)` check to the recursive normalization condition in `normalizeWhereCriteria`

## Tests

Ran:

```
npm test
```

Result: Tests not run because the full test suite requires database connections (MySQL, Postgres, etc). The change is a single guard condition that can be reviewed by inspection.

## Risk

Risk level: low. The `isUint8Array` helper already exists in the codebase (`Uint8ArrayUtils.ts`) and is used elsewhere. This change only prevents Buffer/Uint8Array values from entering the recursive normalization path — they are now correctly passed through as leaf values.

### DIFF
--- a/src/util/OrmUtils.ts
+++ b/src/util/OrmUtils.ts
@@ -699,6 +699,7 @@ export class OrmUtils {
                 typeof value === "object" &&
                 !Array.isArray(value) &&
                 !(value instanceof Date) &&
+                !isUint8Array(value) &&
                 !InstanceChecker.isFindOperator(value)
             ) {
                 const nested = OrmUtils.normalizeWhereCriteria(


alumni:
Sorry @handsomesix, the fix is a little bit more complex than that - I already have a quick fix, I just need to write some tests for it.

Issue #12578, opened 2026-06-12
OrmUtils.normalizeWhereCriteria() does not implement described default behaviour for invalidWhereValuesBehavior


samkessaram:
### Issue description

If the DataSource is not configured with invalidWhereValuesBehavior, OrmUtils.normalizeWhereCriteria() does not throw by default.

### Expected Behavior

If invalidWhereValuesBehavior is not set, OrmUtils.normalizeWhereCriteria() should implement the default behaviour described in the docs, i.e. it should throw an error.

https://typeorm.io/docs/data-source/null-and-undefined-handling#throw-default-1

> 'throw' (default)
> JavaScript undefined values cause a TypeORMError to be thrown:

<img width="539" height="88" alt="Image" src="https://github.com/user-attachments/assets/74ccf110-5701-44e8-a7c3-320e7115276d" />

### Actual Behavior

Currently, if no options are set in the DataSource then the criteria is returned immediately.

```
    static normalizeWhereCriteria(
        criteria: ObjectLiteral,
        options?: {
            null?: "ignore" | "sql-null" | "throw"
            undefined?: "ignore" | "throw"
        },
        path?: string,
    ): ObjectLiteral {
        if (!options) return criteria
```

https://github.com/typeorm/typeorm/blob/1.0.0/src/util/OrmUtils.ts#L670

So, UPDATE queries are still executed with buggy filters, e.g. `UPDATE "schema_x"."table_y" SET "col1" = $1 WHERE "col2" = null` with no warning/error for the dev to detect this easily.

### Steps to reproduce

https://github.com/samkessaram/typeorm-undefined-where-bug

 ```bash
  git clone https://github.com/samkessaram/typeorm-undefined-where-bug.git
  cd typeorm-undefined-where-bug
  npm install
  npm start
```

### My Environment

| Dependency          | Version  |
| ---                 | ---      |
| Operating System    |     MacOS     |
| Node.js version     | 24.16.0  |
| Typescript version  | 5.9.3  |
| TypeORM version     | 1.0.0 |
| @nestjs/typeorm   |  11.0.1 |

### Additional Context

**_Risk_**: If invalidWhereValuesBehavior: { undefined: "ignore" } is later added (e.g. during a version upgrade), the update() path changes from "silent no-op" to "unscoped UPDATE affecting ALL rows" — with no intermediate error state to catch it.

### Relevant Database Driver(s)

- [ ] aurora-mysql
- [ ] aurora-postgres
[...]

PR #12579, opened 2026-06-12

samkessaram:
Fixes #12578

`OrmUtils.normalizeWhereCriteria()` returns criteria unvalidated when `invalidWhereValuesBehavior` is not configured, because `if (!options) return criteria` short-circuits before the default-throw logic runs. This causes `update()` and `delete()` to silently accept `undefined`/`null` in where conditions while `find()` correctly throws.

**Fix:** Replace the early return with `options ??= {}` so the existing fallback (`?? "throw"`) applies.

**Tests:** Added unit tests covering default-throw for undefined, null, nested objects, and explicit ignore.


naorpeled (approved):

gioboa (approved):
Thanks @samkessaram LGTM 👍

samkessaram:
Cool, looks like it is affecting other tests so I'll check that out.

src/util/OrmUtils.ts
```diff
@@ -661,9 +661,13 @@ export class OrmUtils {
...
-    ): ObjectLiteral | ObjectLiteral[] {
-        if (!options) {
-            return criteria
+    ): ObjectLiteral {
+        if (OrmUtils.isPlainObject(criteria)) {
+            options ??= {}
+        } else {
+            // Entity class instances may carry nullable columns (e.g. foreign keys)
+            // that are not intended as filter conditions — default to ignoring them.
+            options ??= { null: "ignore", undefined: "ignore" }
```
samkessaram: Would like some extra attention here, as this check wasn't in before. The commit message, for reference:

> When an entity instance (e.g. from repository.save()) is passed to softDelete/delete/update,
nullable columns like foreign keys are not intended as filter conditions. Use isPlainObject to
distinguish entity instances from plain FindOptionsWhere objects and default to ignoring
null/undefined on the former while preserving strict throw behavior for the latter.

gioboa: cc @naorpeled @alumni

samkessaram:
For reference, the new commit was in response to this failure: https://github.com/typeorm/typeorm/actions/runs/27433004792/job/81089129076?pr=12579#step:6:3695

src/util/OrmUtils.ts
naorpeled: I'm not sure about this one, can you please share some examples that emphasize using ignore and not throw on both? 🙏 

Thanks in advance

naorpeled:
Hey @samkessaram,
Can you please address Qodo's feedback? 🙏

src/util/OrmUtils.ts
samkessaram: Here's what I'm thinking: 

**Plain objects** (`{ id: 1, name: undefined }`): more likely a bug I think, either some value not being returned when expected or just human error. Default: **throw**.

**Entity class instances**: these carry all columns including nullable foreign keys (`parentId: null`, `deletedAt: undefined`). When passed to `softDelete(entity)`, only the non-null fields should form the WHERE clause. Default: **ignore**.

```ts
// Plain object --> throw 
const name = await fetchUserName(id) // returns null unexpectedly
repo.delete({ name }) // throws — prevents unfiltered delete

// Entity instance --> ignore
class Post {
  id = 1
  authorId: number | null = null  // nullable FK
}
repo.softDelete(new Post()) // Not plain object, drop authorId, filter is just WHERE id = 1
```

We use `isPlainObject()` to distinguish these cases — entity instances created with `new` have a class prototype and aren't plain objects.

naorpeled: @alumni wdyt?

alumni: I'm thinking that right now the logic is a bit convoluted.

The `isPlainObject` check should be done before iterating through the keys (we should have an early return if it's not a plain object) and the `options` should determined at the beginning of the functions in a single way (and not have multiple fallback values).

Also, I'm thinking that maybe we should have two PRs, one for the plain object check which we can backport to 0.3 and one for the default value which we should not.

naorpeled: Sounds good to me, @samkessaram would you like to take this ? Or would you rather me to take it?🙏

belljinwo-jpg:
Since #12595 was closed as a duplicate of this PR, here are the tested pieces from that implementation that may help finish the canonical fix:

- functional write-path coverage for `EntityManager.update()` and `delete()` under default null/undefined behavior;
- a null-prototype normalization accumulator plus regression coverage for JSON-parsed `__proto__` criteria;
- entity-instance passthrough only when the option is omitted, while explicitly configured behavior still validates the instance;
- focused functional suite: 24 passing; complete SQL.js suite: 2,965 passing, 179 pending, 0 failing.

The implementation and commits remain available in #12595. I’m happy to split or rebase the relevant tests/fixes if that is preferable to cherry-picking them.

src/util/OrmUtils.ts
samkessaram: Hey @naorpeled, yeah I wouldn't mind if you took on that. Sorry—some things have come up on my side. I can check back in after I get some stuff sorted and am happy to help wherever needed.

PR #12582, opened 2026-06-14
fix(postgres): use ALTER COLUMN TYPE for length-only changes to prevent data loss

BWM0223:
## Problem

Fixes #3357

When a PostgreSQL column's **length** changes (e.g.  → ), TypeORM generates:

```sql
ALTER TABLE "example" DROP COLUMN "col"
ALTER TABLE "example" ADD "col" character varying(100)
```

This **destroys all existing data** in that column.

## Root Cause

 lumped  changes together with actual type changes in the recreate condition (lines 1330-1340). A length-only change on  does not require column recreation — PostgreSQL supports it natively with .

## Fix

Detect length-only changes (same type, same array flag, no generated column change) and use safe :

```sql
ALTER TABLE "example" ALTER COLUMN "col" TYPE character varying(100)
```

This preserves all existing data.

## Checklist
- [x] Bug fix (non-breaking change)
- [x] Existing tests pass
- [x] Addresses data loss issue reported in #3357

### DIFF
--- a/src/driver/postgres/PostgresQueryRunner.ts
+++ b/src/driver/postgres/PostgresQueryRunner.ts
@@ -1326,20 +1326,39 @@ export class PostgresQueryRunner
                 `Column "${oldTableColumnOrName}" was not found in the "${table.name}" table.`,
             )
 
-        if (
-            oldColumn.type !== newColumn.type ||
-            oldColumn.length !== newColumn.length ||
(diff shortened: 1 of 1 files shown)


alumni:
SEE https://github.com/typeorm/typeorm/issues/3357#issuecomment-4638034647

PR #12584, opened 2026-06-15
feat: add sqlite option to preserve foreign keys during migrations

carl-sauter:
allow to keep sqlite foreign_keys enabled during migration

### Description of change

I added an option to preserve `PRAGMA foreign_keys = ON` for all sqlite drivers. At the moment foreign keys are always getting disabled during migrations. If you write migrations by hand you can migrate without violating the foreign keys. This option allows you to keep foreign keys enabled all the time. This way foreign keys cannot be violated at any time.

### Pull-Request Checklist

      `Fixes #NNNN`, `Closes #NNNN`, or `Resolves #NNNN` N/A (minor change)
- [ ] There are new or updated tests validating the change (`tests/**.test.ts`) N/A

### DIFF
--- a/docs/docs/drivers/sqlite.md
+++ b/docs/docs/drivers/sqlite.md
@@ -22,7 +22,7 @@ See [Data Source Options](../data-source/2-data-source-options.md) for the commo
 
 ### Common sqlite data source options
 
-- `preserveForeignKeysDuringMigrations` - Keeps SQLite foreign-key checks enabled during migrations (default `false`).
+- `preserveForeignKeysDuringMigrations` - Keeps SQLite foreign-key checks enabled during migrations and schema synchronization (default `false`).
 
 ### `better-sqlite3` data source options
 
--- a/docs/docs/drivers/sqlite.md
+++ b/docs/docs/drivers/sqlite.md
@@ -20,6 +20,10 @@ npm install sql.js
 
 See [Data Source Options](../data-source/2-data-source-options.md) for the common data source options.
 
+### Common sqlite data source options
+
+- `preserveForeignKeysDuringMigrations` - Keeps SQLite foreign-key checks enabled during migrations (default `false`).
+
 ### `better-sqlite3` data source options
 
 - `database` - Database path. For example, `"mydb.sqlite"`.
--- a/src/driver/better-sqlite3/BetterSqlite3DataSourceOptions.ts
+++ b/src/driver/better-sqlite3/BetterSqlite3DataSourceOptions.ts
@@ -1,4 +1,4 @@
-import type { AbstractSqliteDataSourceOptions } from "../sqlite-abstract/AbstractSqliteDataSourceOptions";
+import type { AbstractSqliteDataSourceOptions } from "../sqlite-abstract/AbstractSqliteDataSourceOptions"
 
 /**
  * Sqlite-specific connection options.
--- a/src/driver/better-sqlite3/BetterSqlite3DataSourceOptions.ts
+++ b/src/driver/better-sqlite3/BetterSqlite3DataSourceOptions.ts
@@ -1,4 +1,4 @@
-import {AbstractSqliteDataSourceOptions} from "../sqlite-abstract/AbstractSqliteDataSourceOptions";
+import type { AbstractSqliteDataSourceOptions } from "../sqlite-abstract/AbstractSqliteDataSourceOptions";
 
 /**
  * Sqlite-specific connection options.
--- a/src/driver/better-sqlite3/BetterSqlite3DataSourceOptions.ts
+++ b/src/driver/better-sqlite3/BetterSqlite3DataSourceOptions.ts
@@ -1,9 +1,9 @@
-import type { BaseDataSourceOptions } from "../../data-source/BaseDataSourceOptions"
+import {AbstractSqliteDataSourceOptions} from "../sqlite-abstract/AbstractSqliteDataSourceOptions";
 
 /**
  * Sqlite-specific connection options.
  */
-export interface BetterSqlite3DataSourceOptions extends BaseDataSourceOptions {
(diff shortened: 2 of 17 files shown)


carl-sauter:
In theory I only need to check in `beforeMigration` and always run `afterMigration`, but I think this is the cleaner solution.

linusrush:
I ran into this a couple times, really helpful!

smith-xyz:
this is a good add but I think maybe it should be adjusted some.  There are a few PRAGMA that are per connection, and ideally for sqlite the abstraction for typeorm in my opinion is that you could set these per connection :

PRAGMA synchronous = NORMAL;
PRAGMA foreign_keys = ON;
PRAGMA cache_size = -20000;
PRAGMA busy_timeout = 5000;

what might be more ideal is that for a given connection event you establish (query, migration, etc), for sqlite specifically could configure these to execute as part of that sql.

I'm curious, with subscribers are we not able to configure the pragma for the migration event you need here?

carl-sauter:
Thank you for your feedback, I must admit I don't have this kind of deep knowledge about the project to implement something like you mentioned.

I just thought this is the best option to solve the foreign keys problem, as I had my database in corrupted state after a migration, without even knowing it. Right now typeorm is enabling the PRAGMA foreign_keys for each connection automatically, but disables it during migration. You also cannot enable it again in the migration, as sqlite does not allow to change it during a transaction.

If you think some kind of bigger refactoring is needed, may someone else can takeover the pr :) Otherwise im pretty happy with this solution.

smith-xyz:
no worries, what I was thinking is that since these are connection level pragma maybe something like this so we can configure each pragma independently:

```
{
  type: "better-sqlite3",
  database: "app.db",
  pragmas: {
    foreign_keys: "ON",  
    journal_mode: "WAL",
    busy_timeout: 5000,
  }
}
```

to me the feature shouldn't one off fix for just the foreign keys in migrations - its something that is connection specific when running the query and one could declare for all queries they run or change the shape relative to what is needed.

carl-sauter:
In my opinion this is just a quality of life for developers. Its just shorthand for running `PRAGMA x = y` on connection open using the query runner.
You are correct that the pragmas are effecting the whole connection, but they can be turned off at any time. So even when you open the connection with this pragmas it isn't guarenteed that they are still in place afterwards.

Foreign Keys is kind of a special case, because typeorm enables it by default in the sqlite drivers, and disables it for migrations.
For example BetterSqlite3Driver:
```typescript
// we need to enable foreign keys in sqlite to make sure all foreign key related features
// working properly. this also makes onDelete to work with sqlite.
databaseConnection.pragma("foreign_keys = ON")
```

I think you will not get around disabling foreign keys during migrations by default, as this might break existing code. So even with your suggestion for the pragmas, this option has a usecase.

alumni:
Disabling the FKs was originally introduced in #7922. It was trying to solve #2576.

~~I think that the original issue is no longer valid - that was in a time when `ALTER TABLE` was not properly supported in sqlite, so it seems that TypeORM was dropping the table and re-creating it. This was a quick find, I'd need to look at the history of the repo to confirm that - but maybe we don't even need that pragma anymore.~~

_Later Edit:_ it seems we still call `AbstractSqliteQueryRunner.recreateTable` when adding a column :(

Sqlite supports e.g. [`ALTER TABLE ADD COLUMN`](https://www.sqlite.org/lang_altertable.html#alter_table_add_column), but there are some caveats. I wonder if that's the reason why we call `recreateTable`? Maybe because of [Why `ALTER TABLE` is such a problem for SQLite](https://www.sqlite.org/lang_altertable.html#why_alter_table_is_such_a_problem_for_sqlite)?

carl-sauter:
Sqlite has very poor `ALTER` support. In most cases this leads to recreating the whole table. However I think this should always be possible to do, without disbabling foreign keys. You just have to make sure you remove all foreign key constraints beforehand.

Anyway how will you proceed with this? Are there any open problems, that I can resolve?

carl-sauter:
Hey just wanna ask if there is still anything open or holding you back from merging?

smith-xyz:
@carl-sauter sorry for the delay getting back to this. 

Opt-in LGTM in principle — default must stay OFF because of recreateTable/DROP under FKs (#7922/#2576). 

Let's fix the following: 

- Always run afterMigration → PRAGMA foreign_keys = ON even when preserve is true (Qodo bug).
- Add a test that proves preserve keeps FKs on through a migration and that default still turns them off.
- Clarify docs: option also affects synchronize (shared hooks), not only migrations.

I think its a good quality of life choice - but I'll open a separate issue detailing the better API path we should consider and how we can work that in with sync/migrations to give users the choice on what to do.

smith-xyz (changes requested):
see comment above

carl-sauter:
@smith-xyz thank you for the feedback. I am struggeling a bit with the test. Where would you see the test case `functional/migration` or somewhere by the driver?

smith-xyz:
@carl-sauter I think since this is driver specific it could go in /test/functional/driver/abstract-sqlite

we also have this /test/other-issues as well might be fine there as well - we're in the process of restructuring this project so either works in my opinion

smith-xyz:
> @smith-xyz thank you for the feedback. I am struggeling a bit with the test. Where would you see the test case `functional/migration` or somewhere by the driver?

yeah might be good there in test/functional/driver/abstract-sqlite - abstractly it relates to sqlite and migrations

Could also be fine in test/other-issues

Hoping we can organize these better in the future by moving to packages

carl-sauter:
Hey @smith-xyz I added the test. I kinda new to this, so Im a bit uncertain if everything is correct. The bot added Compliance violation, but isnt showing me whats the problem.

Looking forward for your feedback.

carl-sauter:
Hey @smith-xyz still waiting for your feedback :)

michaelbromley:

```
git fetch origin
```

PR #12593, opened 2026-06-19
fix: apply default invalid where behavior

tinyopsstudio:
## Summary

- apply the documented default `invalidWhereValuesBehavior` fallback in `OrmUtils.normalizeWhereCriteria()` even when the data source option is not configured
- add regression coverage for default null/undefined behavior on `EntityManager.update()` and `EntityManager.delete()` criteria

Fixes #12578.

## Tests

- `pnpm run compile`
- `pnpm exec mocha --no-config --file build/compiled/test/utils/test-setup.js --timeout 90000 --exit build/compiled/test/functional/null-undefined-handling/query-builders.test.js`
- `pnpm exec prettier --check src/util/OrmUtils.ts test/functional/null-undefined-handling/query-builders.test.ts`

### DIFF
--- a/src/util/OrmUtils.ts
+++ b/src/util/OrmUtils.ts
@@ -662,10 +662,6 @@ export class OrmUtils {
         options?: InvalidFindOptionsWhereBehavior,
         path?: string,
     ): ObjectLiteral | ObjectLiteral[] {
-        if (!options) {
-            return criteria
-        }
-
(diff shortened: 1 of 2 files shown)


tinyopsstudio:
Closing this duplicate from TinyOps after noticing #12579 already covers the same issue with a fuller patch. Sorry for the noise.

PR #12594, opened 2026-06-19
fix: apply default 'throw' behavior in normalizeWhereCriteria when options not set

sunchengrong:
### Problem

When `invalidWhereValuesBehavior` is not configured in the DataSource options, `OrmUtils.normalizeWhereCriteria()` immediately returns the criteria without checking for `undefined`/`null` values:

```typescript
if (!options) {
    return criteria  // Bug: skips all validation
}
```

This contradicts the [documented default behavior](https://typeorm.io/docs/data-source/null-and-undefined-handling#throw-default-1), which states that `'throw'` is the default. As a result, UPDATE queries can silently execute with buggy filters like `WHERE col = null` without any warning.

### Solution

When `options` is not provided, resolve it to the documented defaults (`{ undefined: "throw", null: "throw" }`) instead of short-circuiting:

```typescript
const resolvedOptions: InvalidFindOptionsWhereBehavior = options ?? {
    undefined: "throw",
    null: "throw",
}
```

All internal references to `options` are updated to use `resolvedOptions`.

### Impact

- **With `invalidWhereValuesBehavior` configured**: No change — existing behavior is preserved
- **Without `invalidWhereValuesBehavior`**: Now correctly throws on `undefined`/`null` values in where conditions, matching the documented default

### Files changed

- `src/util/OrmUtils.ts` — Resolve `options` to documented defaults when not provided

### DIFF
--- a/src/util/OrmUtils.ts
+++ b/src/util/OrmUtils.ts
@@ -662,8 +662,11 @@ export class OrmUtils {
         options?: InvalidFindOptionsWhereBehavior,
         path?: string,
     ): ObjectLiteral | ObjectLiteral[] {
-        if (!options) {
-            return criteria
+        // When options is not provided, use the documented default behavior
+        // (throw on undefined/null) rather than silently passing through
+        const resolvedOptions: InvalidFindOptionsWhereBehavior = options ?? {
+            undefined: "throw",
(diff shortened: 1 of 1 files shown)


naorpeled:

PR #12601, opened 2026-06-20
fix: normalizeWhereCriteria throws by default when invalidWhereValuesBehavior is unset

ellitedom03:
## Summary

Removes the early `if (!options) return criteria` guard in `OrmUtils.normalizeWhereCriteria()` that bypassed all undefined/null validation when the DataSource had no `invalidWhereValuesBehavior` configured.

Closes #12578

## Problem

When `invalidWhereValuesBehavior` is not set in DataSource options (the documented default), `normalizeWhereCriteria()` returns the criteria immediately without validation. This means `UPDATE`, `DELETE`, `softDelete`, and `restore` operations with `{ col: undefined }` in the where condition silently produce `WHERE col = NULL` (a no-op in SQL) instead of throwing an error as documented.

Docs: https://typeorm.io/docs/data-source/null-and-undefined-handling#throw-default-1

## Fix

The function body already defaults to 'throw' via `options?.undefined ?? 'throw'` and `options?.null ?? 'throw'`. With `options === undefined`, optional-chaining yields `undefined`, which `?? 'throw'` correctly maps to the documented default. The only defect was the early return bypassing this logic.

**Change:** Removed the `if (!options) return criteria` block. Array handling at the top level is preserved.

## Tests

Added `test/functional/null-undefined-handling/entity-manager.test.ts` covering:
- `update()` with undefined/null where values (default -> throws TypeORMError)
- `delete()` with undefined/null where values (default -> throws TypeORMError)
- `softDelete()` with undefined where values (default -> throws TypeORMError)
- `restore()` with undefined where values (default -> throws TypeORMError)
- `update()` with `invalidWhereValuesBehavior: { undefined: 'ignore', null: 'ignore' }` -> skips undefined keys

## Breaking Change

This is a behavioral change: users relying on the silent no-op when `invalidWhereValuesBehavior` is unset will now get a TypeORMError. This aligns with the documented default. To keep the old behavior, set `invalidWhereValuesBehavior: { undefined: 'ignore', null: 'ignore' }`.

### DIFF
--- a/src/util/OrmUtils.ts
+++ b/src/util/OrmUtils.ts
@@ -673,8 +673,12 @@ export class OrmUtils {
                     ),
             )
         }
-        const result: ObjectLiteral = {}
+        const result: ObjectLiteral = Object.create(null)
         for (const [key, value] of Object.entries(criteria)) {
+            // Skip dangerous prototype-polluting keys
+            if (key === "__proto__" || key === "constructor" || key === "prototype") {
+                continue
+            }
             const propertyPath = path ? `${path}.${key}` : key
 
             if (value === undefined) {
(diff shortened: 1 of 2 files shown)


naorpeled:

PR #12602, opened 2026-06-20
fix: normalizeWhereCriteria should throw by default when DataSource options not configured

Arcticeternity:
## Description

Fixes #12578

When `DataSource` is created without `invalidWhereValuesBehavior` configuration, `OrmUtils.normalizeWhereCriteria()` should default to `throw` behavior for both `null` and `undefined` values — as documented.

## Root Cause

The early return at `OrmUtils.ts:665-667`:
```typescript
if (!options) {
    return criteria
}
```
skipped all validation when no `invalidWhereValuesBehavior` was configured. This caused undefined/null values to silently pass through in WHERE criteria, leading to buggy UPDATE queries.

## Fix

Removed the early return. The existing code already handles defaults correctly:
- `options?.undefined ?? "throw"` → throws on undefined
- `options?.null ?? "throw"` → throws on null

This matches the documented default behavior. No behavioral change for explicitly configured DataSources.

## Checklist
- [x] Bug fix (non-breaking change)
- [x] Existing tests continue to pass
- [x] No new dependencies

### DIFF
--- a/src/util/OrmUtils.ts
+++ b/src/util/OrmUtils.ts
@@ -662,10 +662,6 @@ export class OrmUtils {
         options?: InvalidFindOptionsWhereBehavior,
         path?: string,
     ): ObjectLiteral | ObjectLiteral[] {
-        if (!options) {
-            return criteria
-        }
-
         // multiple criteria are possible at the top level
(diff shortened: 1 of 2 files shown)


naorpeled:

PR #12606, opened 2026-06-21
feat: implement default 'throw' behavior for normalizeWhereCriteria when options not set

Nexussyn:
## Summary

Fixes `OrmUtils.normalizeWhereCriteria()` to implement the documented default behavior of throwing errors when `null` or `undefined` values are encountered in where conditions, even when `invalidWhereValuesBehavior` is not explicitly configured in the DataSource.

### Problem

Previously, if `options` was not provided to `normalizeWhereCriteria()`, the function would return the criteria immediately without any validation, allowing potentially dangerous queries like `UPDATE ... WHERE col = null` to execute silently.

### Solution

When `options` is not provided, the function now uses the default behavior (`throw` for both `null` and `undefined`) as documented in the TypeORM docs.

Closes #12578


naorpeled:

PR #12610, opened 2026-06-22
fix: apply invalid where defaults without explicit options

codex1031-agent:
### Description of change

Fixes #12578.

This removes the early return in `OrmUtils.normalizeWhereCriteria()` when `invalidWhereValuesBehavior` is not configured. The function already defaults `null` and `undefined` behavior to `"throw"` while walking criteria, but the early return skipped that logic entirely for high-level APIs such as `EntityManager.update()`.

The change also adds regression coverage for `EntityManager.update()` with `null` and `undefined` criteria using the default data-source configuration.

### Pull-Request Checklist

- [x] This pull request links a relevant issue using a closing keyword: `Fixes #12578`
- [ ] Documentation has been updated to reflect this change (`docs/docs/**.md`) N/A, this aligns implementation with existing docs

### Verification

I attempted to run the targeted tests locally, but `pnpm install --frozen-lockfile` exceeded the 10-minute command timeout in my environment and did not leave a usable `node_modules`. The diff is intentionally limited to the skipped default path and its regression tests.


naorpeled:

PR #12611, opened 2026-06-22
fix: throw by default for null/undefined in update/delete where criteria

naorpeled:
Fixes #12578. Supersedes #12579

## Problem

`OrmUtils.normalizeWhereCriteria()` returned criteria **unvalidated** when `invalidWhereValuesBehavior` was not configured, because an early `if (!options) return criteria` short-circuited before the per-key default-throw logic ran. As a result `update()`, `delete()`, `softDelete()` and `restore()` silently accepted `null`/`undefined` in where conditions (e.g. `UPDATE "t" SET ... WHERE "col" = null`) while `find()` correctly threw.

These write methods build their `WHERE` via `QueryBuilder.getWhereCondition()` → `getPredicates()`, which performs **no** null/undefined validation (unlike `SelectQueryBuilder.buildWhere`, the `find()` path). So `normalizeWhereCriteria` is the only validator on the write path — there is no downstream safety net.

## Changes

- **Remove the early `if (!options) return`** and resolve the behavior once at the top of `normalizeWhereCriteria`, defaulting to the documented `"throw"` (matching `find()`).
- **Validate plain-object criteria only.** Entity class instances (and other non-plain objects) pass through unchanged via an early return — their set columns, including nullable foreign keys that happen to be `null`, are intentionally part of the `WHERE`. This keeps `softDelete(entity)` working as before and is what the `isPlainObject` check guards (per @alumni's review).
- **Refuse unfiltered writes.** When values are ignored (`invalidWhereValuesBehavior: { … "ignore" }`) and normalization strips a criteria down to something that would touch every row (empty `{}`, or an OR array with an empty branch), the operation now rejects instead of running an unscoped `UPDATE`/`DELETE`. New `OrmUtils.isNormalizedCriteriaUnfiltered()` helper, applied at all four `EntityManager` call sites.
- Resolve `null`/`undefined` behavior in a single place, add a `__proto__` guard, and tidy the OR-array recursion.
- Docs: clarify that the default-throw applies to plain-object criteria, that entity instances pass through unchanged, and that ignored values which strip a write to empty now throw.

## Tests

- **Unit** (`test/unit/util/orm-utils.test.ts`): default-throw for null/undefined, nested paths, `ignore`, `sql-null` → `IsNull()`, entity-instance passthrough, OR-array semantics, primitive/null criteria, `__proto__` guard, and `isNormalizedCriteriaUnfiltered`.
- **Functional** (`test/functional/null-undefined-handling/query-builders.test.ts`): a new **default (unconfigured)** suite covering `update`/`delete`/`softDelete`/`restore` throwing by default — the real regression gap, since every existing throw test explicitly set options — plus nested/array cases, entity-instance passthrough, and the unfiltered-write guard.

## Backport note

On `master` the per-key logic already defaulted to `"throw"`, so this is a single cohesive change. The `v0.3` branch defaults to `"ignore"` and is structured differently; per @alumni the **plain-object-check structure** is backportable to `v0.3` while the **default-throw flip is not**. That backport will be a separate PR against `v0.3`.

### DIFF
--- a/docs/docs/data-source/5-null-and-undefined-handling.md
+++ b/docs/docs/data-source/5-null-and-undefined-handling.md
@@ -41,6 +41,12 @@ const posts = await repository.find({
 })
 ```
 
+:::note
+This default throw behavior applies to **plain object** criteria (the `where` option, and plain objects passed to `update`, `delete`, `softDelete`, and `restore`). When an **entity class instance** is passed as criteria, it is passed through unchanged — all of its set columns, including nullable foreign keys that happen to be `null`, become part of the `WHERE` clause. If you don't want a `null` column to act as a filter, omit it from the criteria or use a plain object instead.
+
+When `null` or `undefined` values are configured to be ignored (see below) and that leaves a write operation (`update`, `delete`, `softDelete`, `restore`) with no remaining conditions, TypeORM throws instead of running an unfiltered query that would affect every row.
+:::
+
 ## Configuration
 
 You can customize how null and undefined values are handled using the `invalidWhereValuesBehavior` option in your data source configuration:
--- a/src/entity-manager/EntityManager.ts
+++ b/src/entity-manager/EntityManager.ts
@@ -877,9 +877,9 @@ export class EntityManager {
             if (OrmUtils.isNormalizedCriteriaUnfiltered(normalizedCriteria)) {
                 return Promise.reject(
                     new TypeORMError(
-                        `All where conditions for the update method were removed because their values were ` +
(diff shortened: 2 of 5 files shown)


alumni:
@naorpeled same comment as in #12579 - can we split into multiple PRs targeting the `master` branch (not the `v0.3` branch!) so I can backport a fix for 0.3? Like we can start with fixing the plain object issue.

naorpeled:
> @naorpeled same comment as in #12579 - can we split into multiple PRs targeting the `master` branch (not the `v0.3` branch!) so I can backport a fix for 0.3? Like we can start with fixing the plain object issue.

Let's sync on Discord

PR #12612, opened 2026-06-22
fix: only normalize plain-object where criteria, pass entity instances through

naorpeled:
Refs #12578. Backport companion to #12611 (master).

Per @alumni's review of #12579, this splits out the **plain-object check** as the part that is safe to land on `v0.3`, **without** the default-value change (master keeps `v0.3`'s `ignore` default; the default-throw flip is master-only).

## Problem

On `v0.3`, `OrmUtils.normalizeWhereCriteria()` iterated *any* object — including entity class instances — whenever `invalidWhereValuesBehavior` was configured. A loaded entity carries all its columns, so a nullable foreign key that happens to be `null` (or an unset `undefined` column) could be silently stripped, turned into `IS NULL`, or made to throw, even though those columns are intentionally part of the `WHERE`.

## Changes

- Add an `isPlainObject` early-return: entity instances (and other non-plain objects) pass through unchanged; only plain `FindOptionsWhere` objects are normalized.
- Resolve the `null`/`undefined` behavior once at the top of the function (single determination instead of per-key fallbacks). **The default stays `"ignore"` on `v0.3`** — no behavior change to the default.
- Use `isPlainObject` for nested recursion (consistent with the top-level check), and guard against the `__proto__` key while iterating.

## What is intentionally NOT here

- The default-value flip to `"throw"` (the fix for #12578's symptom) ships on **master** only (#12611). `v0.3` keeps its `"ignore"` default.

## Tests

- Unit (`test/unit/util/orm-utils.test.ts`): no-options passthrough, default-`ignore` on plain objects, throw-when-configured, **entity-instance passthrough even when throw is configured**, and the `__proto__` guard.
- Functional (`test/functional/null-undefined-handling/query-builders.test.ts`): an entity class instance with a `null` column passed to `delete()` does not throw under the `throw` config. All existing `invalidWhereValuesBehavior` tests continue to pass.

### DIFF
--- a/src/util/OrmUtils.ts
+++ b/src/util/OrmUtils.ts
@@ -4,7 +4,6 @@ import {
     PrimitiveCriteria,
     SinglePrimitiveCriteria,
 } from "../common/PrimitiveCriteria"
-import { InstanceChecker } from "./InstanceChecker"
 import { TypeORMError } from "../error"
 import { IsNull } from "../find-options/operator/IsNull"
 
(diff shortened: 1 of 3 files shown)


alumni:
@naorpeled please open a the PR against the v1 branch

PR #12614, opened 2026-06-23
fix: apply default invalid where behavior

jmwdixen-glitch:
## Summary

Fixes `OrmUtils.normalizeWhereCriteria()` so missing `invalidWhereValuesBehavior` options still use the documented default behavior for null and undefined where criteria.

Before this change, `normalizeWhereCriteria()` returned the criteria unchanged when `options` was not provided. That bypassed the existing `?? "throw"` defaults and allowed unsafe null or undefined criteria through paths such as `EntityManager.update()`.

## Changes

- Remove the early return when `invalidWhereValuesBehavior` options are omitted.
- Add unit coverage proving default undefined and null criteria throw.
- Add a safe-value regression check proving normal criteria are preserved.

## Verification

- `corepack pnpm run compile`
- `corepack pnpm exec mocha --no-config build\\compiled\\test\\unit\\util\\orm-utils.test.js`

Fixes #12578

### DIFF
--- a/src/util/OrmUtils.ts
+++ b/src/util/OrmUtils.ts
@@ -662,10 +662,6 @@ export class OrmUtils {
         options?: InvalidFindOptionsWhereBehavior,
         path?: string,
     ): ObjectLiteral | ObjectLiteral[] {
-        if (!options) {
-            return criteria
-        }
-
         // multiple criteria are possible at the top level
         if (!path && Array.isArray(criteria)) {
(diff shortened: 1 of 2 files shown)


jmwdixen-glitch:
Closing this as a duplicate of #12579, which is already the canonical approved fix for #12578. I opened this after independently verifying the same root cause, but I do not want to add duplicate maintainer noise.

PR #12615, opened 2026-06-23
fix: use ALTER COLUMN instead of DROP+CREATE for type/length changes in Postgres

GautamKumarOffical:
fixes #3357

when only the type or length of a column changes (e.g. varchar(255) to varchar(500), int to bigint), we now use ALTER COLUMN ... TYPE instead of dropping and recreating the column. this preserves data, indices, foreign keys, and other dependent objects.

only falls back to drop+create for truly incompatible changes like isArray toggling or STORED generated column changes.

tested with postgres driver - column length and type changes now generate proper ALTER statements instead of DROP+CREATE pairs.

### DIFF
--- a/src/driver/postgres/PostgresQueryRunner.ts
+++ b/src/driver/postgres/PostgresQueryRunner.ts
@@ -1336,54 +1336,7 @@ export class PostgresQueryRunner
             oldColumn.asExpression !== newColumn.asExpression &&
             newColumn.generatedType === "STORED"
 
-        // For safe type/length changes, use ALTER COLUMN instead of drop+create
-        // to preserve data and dependent objects (indices, FKs, etc.)
-        const canAlterInPlace =
-            (typeChanged || lengthChanged) &&
-            !arrayChanged &&
-            !storedGeneratedChanged &&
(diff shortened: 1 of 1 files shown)


GautamKumarOffical:
Thanks for the thorough review. All three issues addressed in the latest push:

1. **Manual transaction removed** — the `canAlterInPlace` branch no longer starts its own transaction. ALTER TYPE queries now go into `upQueries`/`downQueries` and execute through the standard `executeQueries()` call at the end of `changeColumn()`.

2. **No longer skips other column changes** — the ALTER TYPE logic is now inside the else branch alongside rename/nullable/default/comment handling, so type/length changes can co-occur with other modifications.

3. **Type string uses `createFullType()`** — handles array suffixes, precision/scale formatting, timestamp variants, and spatial types correctly. Added a guard to skip the separate precision/scale ALTER TYPE when type/length already changed (since `createFullType()` already includes it).

4. **Cloned table updated** — `clonedTable` now reflects the new column definition (type, length, precision, scale, isArray) before `replaceCachedTable()` is called.

PR #12616, opened 2026-06-23
fix: apply default throw behavior in normalizeWhereCriteria when options is undefined

inglevivek:
## Summary

Fixes #12578

## Problem

`OrmUtils.normalizeWhereCriteria()` had an early return when `options` was `undefined` (i.e. `invalidWhereValuesBehavior` was not configured in the DataSource):

```ts
if (!options) {
    return criteria  // skips ALL null/undefined validation
}
```

This meant that when no `invalidWhereValuesBehavior` config was set, `null` and `undefined` values in WHERE clauses were silently passed through, producing dangerous SQL like:

```sql
```

No error or warning was raised, making the bug invisible to developers.

Additionally, the risk noted in the issue: if a user later adds `invalidWhereValuesBehavior: { undefined: 'ignore' }` during an upgrade, the behavior silently shifts from "silent no-op" to "unscoped UPDATE affecting ALL rows" with no intermediate error state to catch it.

## Fix

Instead of bailing out when `options` is `undefined`, fall back to the documented defaults (`null: 'throw'`, `undefined: 'throw'`) and allow any user-supplied config to override them via object spread:

```ts
const effectiveOptions: InvalidFindOptionsWhereBehavior = {
    null: "throw",
    undefined: "throw",
    ...options,
}
```

All further logic uses `effectiveOptions` instead of `options`.

## Backwards Compatibility

- Users who already set `invalidWhereValuesBehavior` explicitly: ✅ no change — their config spreads over the defaults.
- Users who set `invalidWhereValuesBehavior: {}`: ✅ defaults apply (same as documented).
- Users with no config at all: ⚠️ now correctly throws on `null`/`undefined` in WHERE, **which is the documented default behavior** — previously this was silently broken.

## Testing

The reproduction repo from the issue can be used to verify: https://github.com/samkessaram/typeorm-undefined-where-bug

```bash
git clone https://github.com/samkessaram/typeorm-undefined-where-bug.git
cd typeorm-undefined-where-bug
npm install
npm start
```

Before this fix: no error thrown, silent bad SQL.
After this fix: `TypeORMError` thrown as documented.

### DIFF
--- a/src/util/OrmUtils.ts
+++ b/src/util/OrmUtils.ts
@@ -653,6 +653,11 @@ export class OrmUtils {
      * Recursively validates an object where clause, throwing for null/undefined
      * based on the provided invalidWhereValuesBehavior config.
      *
+     * When `options` is not provided (i.e. `invalidWhereValuesBehavior` is not
+     * configured in the DataSource), the documented defaults apply:
+     *   - undefined values → throw
+     *   - null values      → throw
+     *
      * @param criteria
      * @param options
      * @param path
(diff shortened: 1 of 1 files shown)


naorpeled:
Hey, 
Sorry 🙏

PR #12625, opened 2026-06-26
fix(entity-manager): apply default throw behavior in normalizeWhereCriteria

chiddekel:
### Description of change

Fixes #12578.

**What**

`OrmUtils.normalizeWhereCriteria()` short-circuited with `if (!options) return criteria` whenever `invalidWhereValuesBehavior` was not configured on the `DataSource`. As a result, `EntityManager.update/delete/softDelete/restore` (and the matching `Repository` methods) silently accepted `null`/`undefined` where-values instead of throwing, even though the docs describe `'throw'` as the **default** behavior.

See: https://typeorm.io/docs/data-source/null-and-undefined-handling#throw-default-1

**Why**

This let buggy queries run unnoticed, e.g.:

```sql
```

**How**

- Removed the early `if (!options) return criteria` guard in `normalizeWhereCriteria`. The function body already defaults to throwing via `options?.null ?? "throw"` / `options?.undefined ?? "throw"`, so omitting the config now correctly applies the documented default. The top-level array handling and recursion work unchanged with `options === undefined`.

This is backwards-compatible for anyone who has already configured `invalidWhereValuesBehavior`. The QueryBuilder `.where()` path is intentionally left untouched (it remains low-level), as covered by existing tests.

> Note: a follow-up PR (stacked on this one) hardens the `update/delete/softDelete/restore` paths so that, under the `ignore` setting, criteria that normalize down to empty are rejected instead of running an unscoped query. That behavior change is kept separate so this PR stays a focused fix for the documented default.

**How verified**

- `pnpm run typecheck` passes.
- New regression suite in `test/functional/null-undefined-handling/query-builders.test.ts` ("default behavior (no config)"):
  - default (no config) throw behavior for `null`, `undefined`, and nested `undefined` across `update`/`delete`/`softDelete`/`restore`;
  - valid criteria still execute when no config is set.
- Full `null-undefined-handling` suite (default / `throw` / `sql-null` / `ignore` / QueryBuilder) passes; the find-options path is unaffected.

### Pull-Request Checklist

- [x] This pull request links a relevant issue using a closing keyword: `Fixes #12578`
- [x] There are new or updated tests validating the change (`test/functional/null-undefined-handling/query-builders.test.ts`)
- [ ] Documentation has been updated to reflect this change — N/A (existing docs already describe the now-correct default behavior)

### DIFF
--- a/src/util/OrmUtils.ts
+++ b/src/util/OrmUtils.ts
@@ -662,9 +662,10 @@ export class OrmUtils {
         options?: InvalidFindOptionsWhereBehavior,
         path?: string,
     ): ObjectLiteral | ObjectLiteral[] {
-        if (!options) {
-            return criteria
-        }
+        // When `invalidWhereValuesBehavior` is not configured the documented
+        // default behaviour applies (throw on null/undefined). This is handled
+        // below via `options?.null ?? "throw"` / `options?.undefined ?? "throw"`,
+        // so we must not short-circuit when `options` is undefined.
 
         // multiple criteria are possible at the top level
         if (!path && Array.isArray(criteria)) {
--- a/test/functional/null-undefined-handling/query-builders.test.ts
+++ b/test/functional/null-undefined-handling/query-builders.test.ts
@@ -9,6 +9,132 @@ import {
 import { Category } from "./entity/Category"
 import { Post } from "./entity/Post"
 
+describe("entity manager > invalidWhereValuesBehavior default behavior (no config)", () => {
+    // Regression test for https://github.com/typeorm/typeorm/issues/12578
+    // When `invalidWhereValuesBehavior` is not configured, the documented
+    // default ("throw") must apply to the update/delete/softDelete/restore paths.
+    let dataSources: DataSource[]
+
+    before(async () => {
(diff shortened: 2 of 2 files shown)


chiddekel:
Thanks for the catch on the mixed empty-array case — it's a real gap.

Fixed in 3fa00e88a: `normalizeWhereCriteria` now drops any top-level OR-branch that normalizes to an empty object, instead of relying solely on the post-normalization guard. So `[{ text: undefined }, { id: 1 }]` under `ignore` normalizes to `[{ id: 1 }]` (scoped) rather than `[{}, { id: 1 }]` (which QueryBuilder would render as `1=1 OR id = ?`). If every branch is stripped, the result is `[]` and the existing empty-criteria guard rejects it.

Added two tests:
- partially-empty OR-array → only the real branch applies (one row affected, the other preserved);
- all-empty OR-array → rejected with `Empty criteria(s) are not allowed`.

chiddekel:
Split out the empty-criteria hardening (the `ignore` → unscoped-query risk + the OR-array case @qodo flagged) into a stacked follow-up: #12626. This PR is now a focused fix for the documented default-throw behavior only.

PR #12629, opened 2026-06-26
fix(query-builder): reject empty where criteria on update and delete operations

naorpeled:
Part of #12578

## Problem

`EntityManager.update` / `delete` / `softDelete` / `restore` validate criteria for emptiness on the **original** input, then pass the **normalized** criteria to `QueryBuilder.where()`. Normalization can reduce a non‑empty criteria to an empty one, and an empty condition renders as `WHERE 1=1` — silently turning a targeted operation into a **full‑table write/delete**.

Ways the criteria could collapse to empty (or otherwise produce no predicate):

- a `__proto__`‑only object, e.g. `JSON.parse('{"__proto__": {"polluted": true}}')` → `{}`
- all keys stripped under `invalidWhereValuesBehavior.null/undefined = "ignore"`, e.g. `{ text: null }` → `{}`
- an empty OR‑branch in an array: `[{}]`, `[[]]`, `[{ id: 1 }, {}]`
- a bare primitive mixed into an OR‑array: `[1, { id: 2 }]` (`.where(1)` has no predicate)
- an empty entity class instance: `new Post()` (no set columns)

Separately, the write path passed **entity class instances** through normalization untouched, so `invalidWhereValuesBehavior` was **not** enforced for them — inconsistent with the read/find path (`SelectQueryBuilder.buildWhere`), which validates every key regardless of whether it was given a plain object or a loaded entity.

## Changes

- **Validate after normalization.** All four write methods now reject criteria that would produce no predicate, with the existing `Empty criteria(s) are not allowed …` error, instead of emitting `WHERE 1=1`. Covers empty objects/arrays, empty OR‑branches, bare primitives in OR‑arrays, and keyless (empty) non‑plain objects such as empty entity instances.
- **Enforce `invalidWhereValuesBehavior` on entity instances**, exactly like plain `FindOptionsWhere` objects, so reads and writes behave identically. Primitive id/id‑arrays keep using `whereInIds` (no over‑validation), and atomic value‑types (`Date`, `Buffer`/`Uint8Array`) pass through untouched.
- **Guard the `__proto__` key** while normalizing, so a crafted criteria key can't write through the object prototype.
- **Robustness:** `OrmUtils.isCriteriaNullOrEmpty` is a shallow, cyclic‑safe check (no unbounded recursion on self‑referential arrays).
- **Refactor:** a single shared `normalizeAndValidateWhereCriteria` helper backs `update` / `delete` / `softDelete` / `restore` (no duplicated emptiness logic).

Unconfigured data sources are unchanged. For intentional full‑table writes, use the dedicated `updateAll()` / `deleteAll()` methods.

## Behavior (`invalidWhereValuesBehavior`)

`throw`, `sql-null`, and `ignore` all work as documented for plain objects, nested plain objects, arrays, **and** entity instances. Any criteria that normalizes to empty is rejected rather than silently matching every row.

## Tests & docs

- **Unit** (`test/unit/util/orm-utils.test.ts`): shallow/cyclic‑safe emptiness check, `__proto__` guard, entity‑instance validation, value‑type passthrough.
- **Functional** (`test/functional/null-undefined-handling/query-builders.test.ts`): `throw`/`sql-null`/`ignore` across the shapes above; rejection of every empty/degenerate criteria (`__proto__`‑only, all‑keys‑stripped, `[[]]`, `[1, {id}]`, empty entity instance); entity instances validated consistently with `find()`. Green across the full driver matrix.
- **Docs**: `docs/docs/data-source/5-null-and-undefined-handling.md` documents the empty‑criteria rejection and entity‑instance validation.


alumni (dismissed):

KODEXFV:
Noting one scope gap for #12578: this PR body says the unconfigured default behavior is intentionally untouched, and the new unit test currently asserts `OrmUtils.normalizeWhereCriteria(criteria)` returns the same object when `options` is omitted.

That means the original #12578 path still remains for unconfigured data sources: `EntityManager.update/delete/softDelete/restore` pass `this.dataSource.options.invalidWhereValuesBehavior`, so with no configured option `normalizeWhereCriteria()` still returns `{ text: null }` / `{ text: undefined }` unchanged instead of applying the documented default `throw` behavior.

If the default-behavior follow-up lands separately, two edge cases from the duplicate PRs are worth carrying forward with it:

- after normalization, write criteria that become `{}` or contain an empty OR branch should be rejected before `.where(...)`, otherwise `ignore` can turn a scoped write into a broad one;
- once the early return is removed, malformed top-level OR entries like `[null]` / `[undefined]` should throw a controlled `TypeORMError` instead of leaking `TypeError: Cannot convert undefined or null to object`.

Happy to help turn those into a smaller follow-up patch if useful.

src/util/OrmUtils.ts
```diff
@@ -395,6 +395,32 @@ export class OrmUtils {
...
+     * therefore match every row (rendered as `WHERE 1=1`). Unlike
+     * isCriteriaNullOrEmpty, this also rejects an array that contains any empty
+     * element, since a single empty OR-branch matches all rows. Guards against a
+     * non-empty input being reduced to an empty filter by normalizeWhereCriteria
+     * (e.g. the only key was `__proto__`, or all keys were stripped under the
+     * `ignore` invalidWhereValuesBehavior).
+     *
+     * @param criteria
+     */
+    public static isNormalizedWhereCriteriaEmpty(
```
alumni: `isNormalizedWhereCriteriaEmpty` looks like a duplicate of `isCriteriaNullOrEmpty` (with a bug in that function fixed).

src/entity-manager/EntityManager.ts
```diff
@@ -874,6 +874,16 @@ export class EntityManager {
                 criteria as ObjectLiteral | ObjectLiteral[],
                 this.dataSource.options.invalidWhereValuesBehavior,
             )
+            // normalization may have stripped every key (e.g. a "__proto__"-only
+            // object, or all keys removed under the "ignore" behavior); a now-empty
+            // criteria would render as "WHERE 1=1" and update the whole table.
+            if (OrmUtils.isNormalizedWhereCriteriaEmpty(normalizedCriteria)) {
```
alumni: I wonder what's the point of checking if the criteria is empty TWICE. We probably don't do it right the first time 😉 Or we don't do it at the right time 😉😉

michaelbromley:
He @Naor, I had my custom review agent Nigel look at this and something was flagged that is worth checking to see if it is a legit concern, as it might contradict this earlier defense:

> Entity null bypasses sql-null — the passthrough is intentional and matches find() semantics: an entity class instance is not validated against invalidWhereValuesBehavior.

<details><summary>Nigel’s review findings</summary>
TypeORM lets you configure what happens when a null or undefined sneaks into a query's WHERE conditions. You set it once on your connection:

invalidWhereValuesBehavior: { null: "throw", undefined: "throw" }

"throw" means: "if someone tries to filter by a null/undefined value, stop and raise an error instead of silently building a broken query." This exists because column = NULL in SQL matches nothing (you have to write column IS NULL), so a stray null usually means a bug — better to shout than to silently return/delete the wrong rows.

The trap: this rule is enforced in two separate places

Here's the thing most people (including the PR author and the reviewer who approved it) didn't realise. There isn't one piece of code that enforces this setting. There are two completely separate copies, one for each family of operations:
```
┌───────────────────────────────────────────────────┬──────────────────────────────────────────────────────────┐
│                    You call...                    │                 Enforcement lives in...                  │
├───────────────────────────────────────────────────┼──────────────────────────────────────────────────────────┤
│ find(), findBy(), findOne() (reading rows)        │ SelectQueryBuilder.buildWhere() —                        │
│                                                   │ SelectQueryBuilder.ts:4354 onwards                       │
├───────────────────────────────────────────────────┼──────────────────────────────────────────────────────────┤
│ update(), delete(), softDelete(), restore()       │ OrmUtils.normalizeWhereCriteria() — OrmUtils.ts          │
│ (writing/removing rows)                           │                                                          │
└───────────────────────────────────────────────────┴──────────────────────────────────────────────────────────┘
```
Two hand-written implementations of the same idea. They already had slightly different wording and logic before this PR ever showed up.

What the PR changed

The PR added an escape hatch to the write copy only (normalizeWhereCriteria). The new rule: "if the thing you're filtering by is a loaded entity object (not a plain {...}), skip all the null/undefined checking — pass it straight through untouched."
```ts
// the new early-return in normalizeWhereCriteria
if (!OrmUtils.isPlainObject(criteria)) {
    return criteria     // entity instances bypass all validation
}
```
The justification the author gave — and why it's wrong

When a review bot flagged this, the author defended it (in the PR comments) by saying:

▎ "the passthrough is intentional and matches find() semantics: an entity class instance is not validated against invalidWhereValuesBehavior."

In plain English: "Don't worry, the read side (find()) already ignores entities like this, so I'm just matching existing behaviour."

That claim is false. I read the read-side code. Here's the actual loop from SelectQueryBuilder.buildWhere (SelectQueryBuilder.ts:4353-4398):
```ts
for (const key in where) {          // ← iterates EVERY key
    let parameterValue = where[key]
    ...
    if (parameterValue === undefined) {
        ... ?? "throw" ...          // throws unless you opted out
    }
    if (parameterValue === null) {
        ... "throw" ...             // throws unless you opted out
    }
}
```
There is no "is this a plain object or an entity?" check anywhere in here. The read side validates every key of whatever you pass it — plain object or loaded entity, it makes no difference. So find() does not have the behaviour the author claims to be matching.

The concrete consequence

Say you've got invalidWhereValuesBehavior: { null: "throw" } configured, and a loaded Post entity whose nullable text column happens to be null:
```ts
const post = await repo.findOneBy({ id: 5 })   // post.text === null

// READ path — validated, so it throws loudly:
await repo.findBy(post)          // ❌ throws "Null value encountered..."

// WRITE path AFTER this PR — bypassed, so it stays silent:
await repo.delete(post)          // 😶 no error — builds `text = NULL`,
                                 //    matches zero rows, deletes nothing
```
Same input, same setting, opposite behaviour depending on whether you're reading or writing. The whole point of a "throw" setting is that it fires consistently. After this PR it fires on reads but goes quiet on writes — and the person configured "throw" precisely because they wanted to be warned.

Why this matters beyond the one bug

The deeper issue Nigel is pointing at: the same rule living in two copies that are now actively drifting apart. Nobody reading normalizeWhereCriteria next to buildWhere will understand why the identical setting produces different results for the identical input — because the divergence isn't documented, it's accidental.

What "fix" means here

The minimum bar is honesty in the reasoning: the passthrough might be defensible as "this matches the low-level, unvalidated QueryBuilder.where() behaviour" — but not as "matches find()." Either correct that justification, or reconsider whether the bypass should exist at all. The cleaner long-term answer is one shared validation helper feeding both paths, so this class of "two copies drift" bug can't recur — but that's a bigger change worth its own follow-up.
</details>

naorpeled:
@michaelbromley @alumni let me know what you think about this direction of implementation

src/query-builder/QueryBuilder.ts
```diff
@@ -869,6 +869,24 @@ export abstract class QueryBuilder<Entity extends ObjectLiteral> {
             this.expressionMap.wheres,
         )
 
+        // A write operation whose provided WHERE renders empty ("1=1") would
+        // silently affect every row. Reject it instead of emitting an unfiltered
+        // UPDATE/DELETE — this also covers criteria that only look non-empty at
+        // the top level but expand to zero predicates (e.g. a relation key whose
+        // value is a keyless object). To affect all rows, use the dedicated
+        // updateAll()/deleteAll() methods, which set no WHERE.
+        const writeQueryTypes = ["update", "delete", "soft-delete", "restore"]
```
smith-xyz: nit: this feels a little odd here - given that this is just building, surprised we don't store this as metadata as part of the object to denote what type of query it is

if anything maybe we could store this static list somewhere else outside of this protected member?

src/query-builder/QueryBuilder.ts
```diff
@@ -869,6 +869,24 @@ export abstract class QueryBuilder<Entity extends ObjectLiteral> {
...
+        if (
+            writeQueryTypes.indexOf(this.expressionMap.queryType) !== -1 &&
```
smith-xyz: nit: could use .includes(), but I think this is consistent with other areas

src/query-builder/QueryBuilder.ts
```diff
@@ -869,6 +869,24 @@ export abstract class QueryBuilder<Entity extends ObjectLiteral> {
...
+            this.expressionMap.wheres.length > 0 &&
+            (whereExpression.length === 0 || whereExpression === "1=1")
+        ) {
+            throw new TypeORMError(
+                `Empty criteria(s) are not allowed for the ${this.expressionMap.queryType} operation. ` +
+                    `To affect all rows, use the dedicated updateAll/deleteAll methods.`,
```
smith-xyz: minor nit: if  `soft-delete`/`restore`, there is no equivalent to map to, so all actions are being told to go to updateAll/deleteAll.

smith-xyz (approved):
just some nits, lgtm!

src/util/OrmUtils.ts
alumni: The type of criteria is no longer reflecting the real life type?

It should be `any` in real life, but it's incorrectly cast when we call this method.

src/util/OrmUtils.ts
```diff
@@ -678,8 +686,26 @@ export class OrmUtils {
...
+        // FindOptionsWhere objects AND entity class instances alike — is
+        // validated key-by-key, so the write path (update/delete/softDelete/
+        // restore) enforces invalidWhereValuesBehavior consistently with the
+        // read/find path (SelectQueryBuilder.buildWhere), which validates every
+        // key regardless of whether it was given a plain object or an entity.
+        if (
+            criteria === null ||
+            typeof criteria !== "object" ||
+            criteria instanceof Date ||
+            isUint8Array(criteria)
```
alumni: Any code that requires a comment this long is probably doing something it shouldn't.

What would happen for other data types - I'm thinking maybe `FindOperator` or something like that? Or anything else you have not yet thought about?

Right know, we're iterating thought the Object entries for everything else that doesn't match all these things. But who knows what other values we could get here - last time we forgot about `Buffer` and `Array` (and BTW, we are forgetting about arrays here again!)

That's why we should ALWAYS check if it's the kind of value we'd like to work with before working with it (and not what it's not). So if we want to iterate through the `Object.entries(criteria)` we should make sure if makes sense to do that, e.g. it's a plain object criteria, not some instance of a class (like an entity or a `FindOperator` or an array which can be a sparse vector containing nulls).

A little bit below there's an `if (OrmUtils.isPlainObject(value))` - since now the logic from all those functions in `EntityManager` was moved to this method, maybe it makes sense to bring it up here as well?

src/query-builder/QueryBuilder.ts
naorpeled: Done 🙏

src/query-builder/QueryBuilder.ts
naorpeled: Done 🙏

src/query-builder/QueryBuilder.ts
naorpeled: Done 🙏

naorpeled:
@naorpeled all review comments from @alumni and @smith-xyz are addressed in `387633283` and CI is fully green (46/0) 🟢

- `normalizeWhereCriteria` now guards key-iteration with a positive `isPlainObject(criteria)` check (entity/class instances, `FindOperator`s, arrays, `Date`, `Buffer`, primitives pass through untouched); `criteria` typed `any`, casts dropped.
- QB empty-WHERE guard: hoisted `WRITE_QUERY_TYPES` const, uses `.includes()`, and the error no longer suggests `updateAll`/`deleteAll` for `soft-delete`/`restore`.

Ready for release whenever you are 🙏

alumni (approved):

samkessaram:
Hey, initial issue submitter here 👋 

Just wanted to say thanks to all that worked on this, it obviously ended up being a bigger task than I initially thought. Appreciate the efforts!

naorpeled:
> Hey, initial issue submitter here 👋 
> 
> Just wanted to say thanks to all that worked on this, it obviously ended up being a bigger task than I initially thought. Appreciate the efforts! 

Thank you for opening the issue ❤️ 
You dropped this 👑

PR #12638, opened 2026-06-28
fix: apply default invalid where value behavior

KODEXFV:
Fixes #12578.

`OrmUtils.normalizeWhereCriteria()` returned criteria unchanged when `invalidWhereValuesBehavior` was omitted, so update/delete-style operations skipped the documented default behavior for `null` and `undefined` where values.

This change keeps normalization active with the documented defaults:

- `undefined` defaults to `throw`
- `null` defaults to `throw`
- partial configs still default omitted keys to `throw`

It also guards write operations after normalization so `ignore` mode cannot strip criteria down to an empty object or an empty OR branch and accidentally run a broad write.

Additional safety covered:

- rejects criteria that normalize to empty for `update`, `delete`, `softDelete`, and `restore`
- rejects empty OR branches after ignored invalid values are stripped
- rejects empty relation predicates after QueryBuilder proves they produce empty conditions
- rejects malformed OR-array entries with a controlled `TypeORMError`
- preserves `__proto__` as an own data property without changing object prototypes
- treats plain objects by actual prototype instead of trusting a `constructor` property
- preserves empty object values during normalization so valid object predicates are not stripped early
- keeps `constructor` and `prototype` usable as where column names
- keeps string table targets without entity metadata working
- documents that write operations reject criteria that normalize to empty

Validation run locally:

```powershell
pnpm run compile
pnpm exec mocha build\compiled\test\unit\util\orm-utils.test.js
pnpm exec prettier --check docs\docs\data-source\5-null-and-undefined-handling.md src\entity-manager\EntityManager.ts src\util\OrmUtils.ts test\functional\null-undefined-handling\query-builders.test.ts test\unit\util\orm-utils.test.ts
git diff --check
```

Results:

- compile passed
- compiled Mocha run passed: 3033 passing, 179 pending
- touched-file Prettier check passed
- diff whitespace check passed

### DIFF
--- a/docs/docs/data-source/5-null-and-undefined-handling.md
+++ b/docs/docs/data-source/5-null-and-undefined-handling.md
@@ -228,6 +228,12 @@ await manager.delete(Post, { text: null }) // Respects invalidWhereValuesBehavio
 await manager.softDelete(Post, { text: null }) // Respects invalidWhereValuesBehavior
 ```
 
+Write operations that require criteria (`update`, `delete`, `softDelete`, and
+`restore`) also reject empty criteria before executing. If
+`invalidWhereValuesBehavior` removes every usable condition, or removes every
+usable condition from one branch of an OR array, TypeORM treats the criteria as
+empty and throws instead of running a broad write.
+
 ### QueryBuilder with setFindOptions
 
 ```typescript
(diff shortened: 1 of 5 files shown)


KODEXFV:
Updated the PR after the automated review and duplicate-PR comparison:

- replaced the newly added `as any` test casts with typed `unknown` -> target-type helper casts
- added a controlled `TypeORMError` for malformed OR-array entries (`null` / `undefined`) instead of leaking raw runtime `TypeError`
- added docs for write operations rejecting criteria that normalize to empty
- rechecked the remote PR diff: no newly added `as any`
- reran validation: `pnpm run compile`, compiled Mocha run (`3033 passing`, `179 pending`), touched-file Prettier check, and `git diff --check`

PR #12664, opened 2026-07-07
fix: default invalid where values behavior to throw

namdamdoi68-oss:
This PR resolves #12578 by:\n1. Restoring the documented default behavior for invalidWhereValuesBehavior (defaulting to throw TypeORMError on null and undefined when options are omitted).\n2. Adding strict safety checks in EntityManager (update, delete, softDelete, and restore) to reject criteria that normalize to an empty object, preventing accidental unscoped mass mutation queries.

### DIFF
--- a/src/entity-manager/EntityManager.ts
+++ b/src/entity-manager/EntityManager.ts
@@ -870,10 +870,15 @@ export class EntityManager {
 
             return qb.execute()
         } else {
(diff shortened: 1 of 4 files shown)


namdamdoi68-oss:
This PR solves the issue and claims the bounty. /claim

PR #12670, opened 2026-07-08
feat: add sessionParameters option for the postgres driver family

### LINKED ISSUE #12669: Feature proposal: sessionParameters option to set session parameters on new connections
# Feature proposal: `sessionParameters` option to set session parameters on new connections

## Problem

There is currently no first-class way to run per-connection session setup (`SET <param> TO <value>`)
when TypeORM opens a new pooled connection. Users who need session-scoped settings — e.g.
`statement_timeout`, `lock_timeout`, `idle_in_transaction_session_timeout`, `application_name` have
to reach into the underlying pool and attach their own `connect` handler, which means depending on driver
internals that TypeORM otherwise abstracts away.

The only related options today are narrow, single-purpose ones (`applicationName`), or the
generic `extra` escape hatch, which doesn't cover per-connection `SET` statements uniformly.

## Proposal

Add an optional `sessionParameters` data-source option that TypeORM applies to every new
connection it opens from the pool:

```ts
    type: "postgres", // and cockroachdb
    // ...
    sessionParameters: {
        statement_timeout: 60000,
        application_name: "reporting-worker",
    },
})
```

For each entry, TypeORM issues `SET <key> TO <value>` on the freshly established connection
before it is handed to the application.

### Scope

Postgres family only for now (`postgres` + `cockroachdb`) — both wrap `pg` and already expose a
pool `connect` hook, so the mechanism is shared and small. Not proposing a driver-agnostic
option at this stage; other drivers can adopt an equivalent later if there's demand.

### Identifier/value safety

`SET` cannot bind the *identifier* (and in postgres not the value either) as a parameter, so the
key is validated with a conservative regex and rejected otherwise; only well-formed identifiers
are interpolated. This is documented explicitly as **not safe to construct from untrusted user
input** — `sessionParameters` keys and values are configuration, not request data, and callers
must not derive them from end-user input.

### Prior art

- `applicationName` already exists as a dedicated, single-value session setting for
  postgres/cockroach — `sessionParameters` generalizes that idea.
- `node-postgres` (`pg`) exposes `pool.on("connect", client => ...)`, which is exactly the
  mechanism this option uses internally.

## Willingness to contribute

I have a working implementation for cockroachdb (and can extend to postgres) plus tests, and
I'm happy to open a PR.

julius-welink:
### Description of change

Adds a `sessionParameters` data source option for the **postgres driver family** (`postgres` and `cockroachdb`).

- **What:** an object of session parameters applied to every new pooled connection. For each entry TypeORM runs `SELECT set_config(<key>, <value>, false)` on connect.
- **Why:** there is currently no first-class way to run per-connection session setup (`statement_timeout`, `lock_timeout`, `application_name`, custom GUCs, …). Users have to attach their own `pool.on("connect")` handler, depending on driver internals TypeORM otherwise abstracts away. See #12669 for the proposal and discussion.
- **How it works:** a shared `DriverUtils.buildSessionParametersHandler` builds the pool `connect` handler; both `PostgresDriver` and `CockroachDriver` register it. `set_config()` is used rather than `SET <key> TO <value>` because PostgreSQL does not accept bind parameters in the `SET` utility statement — `set_config` is a parameterized function supported by both engines, so the key and value are **bound, not interpolated**.
- **Safety:** parameter names are validated as identifiers and rejected with a `TypeORMError` during connection setup; null/undefined values are rejected too. Keys/values are documented as configuration that must not be built from untrusted user input.

```ts
    type: "postgres", // or "cockroachdb"
    // ...
    sessionParameters: {
        statement_timeout: 60000,
        application_name: "reporting-worker",
    },
})
```

### Why not just use `extra`?

Fair question — and for raw capability, `extra` already covers it. `extra.options` reliably applies session GUCs at connect on both `postgres` and `cockroachdb` (verified on postgres 17 and cockroachdb v24.3.8):

```ts
extra: { options: "-c sql_safe_updates=on -c default_transaction_quality_of_service=background" }
```

So `sessionParameters` is **not** about new capability — it's about ergonomics, with one behavioral improvement:

- **Ergonomics (the main motivation).** The `extra.options` form is a single hand-escaped `-c k=v -c k2=v2` string tied to pg's libpq wire format. The same thing as a typed map, uniform across the postgres family:

  ```ts
  sessionParameters: {
      sql_safe_updates: "on",
      default_transaction_quality_of_service: "background",
  }
  ```

- **Loud failure on misconfiguration.** Via `extra.options`, an unknown/misspelled GUC is *silently ignored* by CockroachDB (Postgres errors at connect). `sessionParameters` runs `set_config`, which errors on an unknown parameter or invalid value on *both* engines; combined with up-front identifier validation and `.catch(poolErrorHandler)` routing, a typo surfaces instead of vanishing.

**How verified:** new tests under `test/functional/driver/postgres-session-parameters.test.ts` — a driver-agnostic validation suite (throws on invalid name, throws on null/undefined, accepts dotted/namespaced names, returns undefined when unconfigured) plus an integration test asserting the parameter is actually applied on pooled connections for both `postgres` and `cockroachdb`.

### Pull-Request Checklist

- [x] This pull request links a relevant issue using a closing keyword: Closes #12669
- [x] There are new or updated tests validating the change (`test/functional/driver/postgres-session-parameters.test.ts`)
- [x] Documentation has been updated to reflect this change (`docs/docs/drivers/postgres.md`)

### DIFF
--- a/src/driver/DriverUtils.ts
+++ b/src/driver/DriverUtils.ts
@@ -1,7 +1,6 @@
 import type { Driver } from "./Driver"
 import { hash, shorten } from "../util/StringUtils"
 import { VersionUtils } from "../util/VersionUtils"
-import { TypeORMError } from "../error/TypeORMError"
 
 /**
  * Common driver utility functions.
@@ -45,67 +44,6 @@ export class DriverUtils {
         return ["postgres", "aurora-postgres", "cockroachdb"].includes(
             driver.options.type,
         )
-    }
-
-    /**
-     * Builds a "connect" handler that applies session parameters to every new
-     * pooled connection via `set_config(<key>, <value>, false)`. Returns
-     * `undefined` when no session parameters are configured. The caller is
-     * responsible for registering the returned handler on its pool.
-     *
-     * `set_config()` is used rather than `SET <key> TO <value>` because PostgreSQL
-     * does not accept bind parameters in the `SET` utility statement; `set_config`
-     * is a regular parameterized function call supported by both PostgreSQL and
-     * CockroachDB. The parameter name and value are bound, so nothing is
-     * interpolated into the SQL text.
-     *
-     * Parameter names are validated up front: any key that is not a valid session
-     * parameter identifier throws a `TypeORMError`, so a misconfiguration fails
-     * loudly during connection setup rather than being silently dropped.
-     *
-     * Callers should apply the returned handler to the first connection during
-     * pool setup — so an invalid value or unknown parameter fails initialization
-     * instead of leaving pooled connections silently unconfigured — and register
-     * it as the pool's `connect` listener for every subsequently created
-     * connection.
-     *
-     * @param sessionParameters Session parameters to apply, keyed by parameter name.
-     * @returns A pool "connect" handler, or `undefined` when nothing is configured.
-     */
-    static buildSessionParametersHandler(
-        sessionParameters?: Record<string, any>,
-    ): ((connection: any) => Promise<void>) | undefined {
-        if (!sessionParameters) return undefined
-
-        // Session parameter names are plain identifiers, optionally namespaced
-        // with a dot for custom parameters (e.g. "my_app.setting").
-        const validSessionParameterKey =
-            /^[A-Za-z_][A-Za-z0-9_]*(\.[A-Za-z_][A-Za-z0-9_]*)?$/
-        for (const key of Object.keys(sessionParameters)) {
-            if (!validSessionParameterKey.test(key))
-                throw new TypeORMError(
-                    `Invalid session parameter name "${key}". Session parameter names must be valid identifiers.`,
-                )
-            if (sessionParameters[key] == null)
-                throw new TypeORMError(
-                    `Invalid value for session parameter "${key}": value must not be null or undefined.`,
-                )
-        }
-
-        return async (connection: any) => {
-            // `.map` runs synchronously, so every query is issued (and thus queued
-            // on the connection) before the pool can hand it out; awaiting between
-            // queries would let the client be acquired with only the first
-            // parameter guaranteed to be applied.
-            await Promise.all(
-                Object.keys(sessionParameters).map((key) =>
-                    connection.query("SELECT set_config($1, $2, false)", [
-                        key,
-                        String(sessionParameters[key]),
-                    ]),
-                ),
-            )
-        }
     }
 
     /**
--- a/src/driver/DriverUtils.ts
+++ b/src/driver/DriverUtils.ts
@@ -1,6 +1,7 @@
 import type { Driver } from "./Driver"
 import { hash, shorten } from "../util/StringUtils"
 import { VersionUtils } from "../util/VersionUtils"
+import { TypeORMError } from "../error/TypeORMError"
 
 /**
  * Common driver utility functions.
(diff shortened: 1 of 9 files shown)


julius-welink:
> ### Code Review by Qodo
> `🐞 Bugs (3)` `📘 Rule violations (0)` `📜 Skill insights (0)`

Review items were addressed.

julius-welink:
Qodo reviews addressed in `65f2d74`:

- **`(options as any)` cast (rule violation):** removed — `setupSingleTestingConnection` already returns `DataSourceOptions`, so the spread is now `...options` with only the outer `as DataSourceOptions` on the literal.
- **Pool leak on invalid config:** `buildSessionParametersHandler` (which validates and can throw) now runs **before** `new Pool(...)`, so an invalid key/value fails before any pool is allocated. On a first-connection apply failure the connection is closed via `release(err)` and initialization rejects.

Note: I did not add `pool.end()` on the failure path — `release(err)` already destroys the client/socket, and the resulting empty, unreferenced `Pool` doesn't keep the event loop alive, so it would be redundant.

julius-welink:
@smith-xyz @naorpeled — sorry to ping directly. The Qodo check on this PR has been stuck on "busy working" since Jul 8. Its two earlier passes are addressed in `65f2d74`.

Would one of you be able to nudge the bot and give this a review when you get a chance?

smith-xyz:
@julius-welink looks like it updated to commit [65f2d74](https://github.com/typeorm/typeorm/pull/12670/commits/65f2d74c790b886b2ddb18e59a186f2e364470c9)

let us know if it looks good - as always if the the findings are irrelevant or don't need addressed just note them and mark the check box that you reviewed

julius-welink:
> @julius-welink looks like it updated to commit [65f2d74](https://github.com/typeorm/typeorm/pull/12670/commits/65f2d74c790b886b2ddb18e59a186f2e364470c9)
> 
> let us know if it looks good - as always if the the findings are irrelevant or don't need addressed just note them and mark the check box that you reviewed

Thanks. It's showing items that were already addressed. No new actionable items. PR is ready for review.

I might be stupid, but the checkbox is NOT clickable 😅

smith-xyz:
strange - went ahead and checked that box, will take a look at this PR a bit more to see how we can prioritize it. Similar (but different) discussion came up around sqlite here: https://github.com/typeorm/typeorm/pull/12584#issuecomment-4723591126

I could see how this is useful, I think my first thought though is how these parameters change from version to version of postgres - any risks there? Should we validate any available options from the db before execution so the user is aware? 

This is just conjecture, I think some of them differ - I could be wrong though 😅

alumni:
I think the name is a mix between "session variables" and "config parameters" (according to PG nomenclature - one is the old name, the other is the new one).

I would call them "session variables" most likely. This how other dialects ([MySQL](https://dev.mysql.com/doc/refman/9.7/en/using-system-variables.html), [HANA](https://help.sap.com/docs/hana-cloud-database/sap-hana-cloud-sap-hana-database-sql-reference-guide/session-variables), [Cockroach](https://www.cockroachlabs.com/docs/stable/session-variables)) also call them.

src/driver/DriverUtils.ts
```diff
@@ -46,6 +47,67 @@ export class DriverUtils {
...
+     * Callers should apply the returned handler to the first connection during
+     * pool setup — so an invalid value or unknown parameter fails initialization
+     * instead of leaving pooled connections silently unconfigured — and register
+     * it as the pool's `connect` listener for every subsequently created
+     * connection.
+     *
+     * @param sessionParameters Session parameters to apply, keyed by parameter name.
+     * @returns A pool "connect" handler, or `undefined` when nothing is configured.
+     */
+    static buildSessionParametersHandler(
```
alumni: This is Postgres-specific. It does not belong in here.

alumni (changes requested):

julius-welink:
> strange - went ahead and checked that box, will take a look at this PR a bit more to see how we can prioritize it. Similar (but different) discussion came up around sqlite here: [#12584 (comment)](https://github.com/typeorm/typeorm/pull/12584#issuecomment-4723591126)
> 
> I could see how this is useful, I think my first thought though is how these parameters change from version to version of postgres - any risks there? Should we validate any available options from the db before execution so the user is aware?
> 
> This is just conjecture, I think some of them differ - I could be wrong though 😅

Config keys can change between versions. Database validates settings, not TypeORM driver itself. Because they are strictly tied to underlying database, it's up to user to provide correct configuration values. I don't think TypeORM should attempt to export a predefined set of options as suggested in [#12584](https://github.com/typeorm/typeorm/pull/12584) — it's impossible to track every single database configuration setting. This PR makes it easier by adding validation: if wrong configuration is set, pool acquire will fail with an error.

Worth noting that what was originally suggested in (https://github.com/typeorm/typeorm/pull/12584#issuecomment-4723591126) — the free-form pragmas: { … } map — is essentially the same key/value model as sessionParameters here. So the two efforts point at one underlying concept: a map of connection-level settings applied to every new connection at session start.

High-level, this concept could generalize across drivers. Scope-wise this PR intentionally implements Postgres + CockroachDB only

julius-welink:
> **Partial parameters applied** (from the Qodo review):
> DriverUtils.buildSessionParametersHandler awaits inside a loop, so the async connect handler yields after the first set_config and later parameters may be applied only after the connection is already in use. PostgresDriver/CockroachDriver register the handler on pool "connect" without awaiting it, so the option does not reliably apply *all* sessionParameters before first use.

This doesn't apply to the current implementation, and I verified it empirically.

The handler issues every set_config synchronously via Object.keys(params).map(k => connection.query(...)) inside Promise.all. In pg-pool's _acquireClient, emit('connect') runs synchronously before the client is handed to the acquirer (pendingItem.callback(...)). So all set_config statements are enqueued on the client before any caller can issue a query, and pg drains the per-client queue FIFO.

src/driver/DriverUtils.ts
julius-welink: Good call — moved it out of `DriverUtils` (which is shared by every driver) into a new `PostgresUtils` class under `src/driver/postgres/`, matching the existing `*Utils` convention. `PostgresDriver` and `CockroachDriver` both call `PostgresUtils.buildSessionParametersHandler`. Went with a dedicated file rather than a static on `PostgresDriver` to avoid growing that already-large class.

Done in 404cd5ea5.

julius-welink:
> I think the name is a mix between "session variables" and "config parameters" (according to PG nomenclature - one is the old name, the other is the new one).
> 
> I would call them "session variables" most likely. This how other dialects ([MySQL](https://dev.mysql.com/doc/refman/9.7/en/using-system-variables.html), [HANA](https://help.sap.com/docs/hana-cloud-database/sap-hana-cloud-sap-hana-database-sql-reference-guide/session-variables), [Cockroach](https://www.cockroachlabs.com/docs/stable/session-variables)) also call them.

Changed the name to "session variables" @alumni

julius-welink:
@smith-xyz is PR is nearly done, but needs reviewer and CI retry.

michaelbromley:

```
git fetch origin
```

PR #12675, opened 2026-07-08
fix(driver): trigger ALTER COLUMN on length-only changes for MySQL, Oracle and SQL Server

Ishant5436:
Resolves #3357

## Problem

When only the `length` property of a column is modified (e.g. `VARCHAR(255)` → `VARCHAR(500)`), TypeORM generated no migration SQL. The `isColumnChanged` helper in `BaseQueryRunner` does not check `length` — it checks `precision`, `scale`, `charset`, `collation`, `isNullable`, etc. but length was omitted.

This means a user who changed only `@Column({ length: 500 })` from `255` would run `typeorm migration:generate` and get an empty migration, silently leaving the database out of sync.

## Solution

Add an `|| oldColumn.length !== newColumn.length` guard alongside the existing `isColumnChanged(...)` call in the affected query runners:

- `MysqlQueryRunner`
- `AuroraMysqlQueryRunner`
- `OracleQueryRunner`
- `SqlServerQueryRunner`

This is the minimal, safe fix — no behavior change for any existing scenario where `isColumnChanged` already returned `true`.

## Testing

The fix applies to the `changeColumn` path used by migration generation. A manual test:
1. Create an entity with `@Column({ length: 255 })`
2. Change to `@Column({ length: 500 })`
3. Run `migration:generate` — previously produced empty migration, now correctly produces `ALTER TABLE ... CHANGE ...`

/opire try

### DIFF
--- a/src/driver/aurora-mysql/AuroraMysqlQueryRunner.ts
+++ b/src/driver/aurora-mysql/AuroraMysqlQueryRunner.ts
@@ -1038,7 +1038,10 @@ export class AuroraMysqlQueryRunner
                 oldColumn.name = newColumn.name
             }
 
-            if (this.isColumnChanged(oldColumn, newColumn, true)) {
+            if (
+                this.isColumnChanged(oldColumn, newColumn, true) ||
+                oldColumn.length !== newColumn.length
+            ) {
                 upQueries.push(
                     new Query(
                         `ALTER TABLE ${this.escapePath(table)} CHANGE \`${
--- a/src/driver/mysql/MysqlQueryRunner.ts
+++ b/src/driver/mysql/MysqlQueryRunner.ts
@@ -1319,7 +1319,10 @@ export class MysqlQueryRunner extends BaseQueryRunner implements QueryRunner {
                 oldColumn.name = newColumn.name
             }
 
-            if (this.isColumnChanged(oldColumn, newColumn, true, true)) {
+            if (
+                this.isColumnChanged(oldColumn, newColumn, true, true) ||
+                oldColumn.length !== newColumn.length
+            ) {
                 upQueries.push(
                     new Query(
                         `ALTER TABLE ${this.escapePath(table)} CHANGE \`${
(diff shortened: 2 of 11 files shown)


Ishant5436:
## Addressed Qodo AI Review Feedback

Thank you for the thorough review. Both issues have been resolved in the latest commit:

### 1. Functional regression test for length-only ALTER (Rule 3 violation)

Added:
- `test/functional/commands/entity/PostWithLongTitle.ts` — entity with `title: varchar(500)` as the "after" state
- New test case in `migration-generate.test.ts`: **"generates ALTER COLUMN when only column length changes (issue #3357)"** — runs against mysql, mariadb, mssql, and oracle, asserting the migration file contains an `ALTER COLUMN` statement
- Expected SQL templates added to `mysql.ts`, `mssql.ts`, `oracle.ts`

### 2. Tautology/WHERE 1=1 guard in `normalizeWhereCriteria`

Added a top-level empty-result guard: if `invalidWhereValuesBehavior.undefined/null = "ignore"` strips **all** keys from the criteria, `normalizeWhereCriteria` now throws `TypeORMError` instead of returning `{}`, preventing the silent `WHERE 1=1` full-table write.

Three new unit tests added:
- All-undefined ignored → throws
- All-null ignored → throws  
- Mixed (some valid keys remain) → succeeds, returns only valid keys

Ishant5436:
## Update: Rebased onto clean `master` baseline

The previous push had 4 commits including unrelated work from prior PRs. This has been corrected — the branch is now rebased directly onto `master` with **only 2 commits** scoped to issue #3357:

1. `fix(driver): trigger ALTER statement on length-only changes` — adds `|| oldColumn.length !== newColumn.length` guard to MySQL, Aurora MySQL, Oracle, SQL Server
2. `fix(driver): add functional test and guard for length-only column change` — adds:
   - Functional regression test in `migration-generate.test.ts`
   - Empty-criteria tautology guard in `OrmUtils.normalizeWhereCriteria`
   - Unit tests for the guard
   - Expected SQL templates for mysql/mssql/oracle

**Files changed:** 11 (4 driver files + OrmUtils + 6 test files). No unrelated changes.

Ishant5436:
Closing per maintainer guidance on #3357 — the team is handling this internally in #11620. Apologies for the noise.

PR #12678, opened 2026-07-08
chore: update dependencies

alumni:
### Description of change

Update all dependencies

### Pull-Request Checklist

      `Fixes #NNNN`, `Closes #NNNN`, or `Resolves #NNNN` N/A
- [ ] There are new or updated tests validating the change (`tests/**.test.ts`) N/A


docs/src/components/tabs/database-tabs.tsx
```diff
@@ -63,7 +63,7 @@ export const DatabaseTabs = ({ children }: PropsWithChildren) => {
     return (
         <Tabs groupId="database" queryString values={values}>
             {entries.map(({ value, db, content }) => (
-                <TabItem key={value} value={value}>
```
naorpeled, line 66: Why did you delete the key here? 🙏

alumni: Type errors - TS complains that `key` is not accepted (that it's not in `Props`).

There's more, and I didn't know what to do about them: in many components TS complains that `children` are missing.

I think TS infers the wrong element type, but I don't know which one should it be :) can you take a look?

alumni: VSCode is confused. Now I'm getting other errors in the same file, but not this one. I'll revert - something is wrong with the setup, but I think it will only be fixed when the source code of the main package will no longer be in the root of the repo.

eslint.config.mjs
```diff
@@ -27,10 +26,6 @@ export default defineConfig([
                 tsconfigRootDir: __dirname,
                 project: "tsconfig.json",
             },
-            globals: {
```
smith-xyz, line 30: is this related to the eslint bump - or is it just not needed?

smith-xyz (approved):
lgtm - curious why we were injecting the globals and no longer need to?

eslint.config.mjs
alumni: It's not needed.

I think it was never needed, but we just followed the guide for setting up `eslint@9` which was mentioning them, probably as an example for adding browser globals. At the moment I don't see any guide mentioning they are needed. I know `eslint@9` was depending on `globals`, `eslint@10` does not anymore.

I'm not sure what rules would depend on these, but usually we define our globals using tsconfig (compilerOptions: lib, types) and `typescript-eslint` would pick things up from there.

I basically removed it because it has no effect for us. The less code we have the better.

PR #12679, opened 2026-07-09
fix: generate alter column when only length changes

Ishant5436:
Fixes #3357

When a column length is changed (e.g. from `VARCHAR(255)` to `VARCHAR(500)`), TypeORM previously failed to detect this as a change because it did not check for `length` in `isColumnChanged`.

This PR adds `length` to the `isColumnChanged` check in `QueryRunner` for `Mysql`, `AuroraMysql`, `Oracle`, and `SqlServer`. It also adds a functional test for this specific regression.


Ishant5436:
Closing per maintainer guidance on #3357 that community PRs are not being accepted for this issue and a team fix is underway at #11620. Apologies for the noise.

PR #12685, opened 2026-07-09
fix(query-builder): treat take(0) as a limit when joins are present

### LINKED ISSUE #12666: take(0) combined with a join returns ALL rows instead of an empty array
### Description
When `take(0)` is combined with a join (e.g. `leftJoinAndSelect`), the query returns every row instead of an empty result, and `getManyAndCount` reports the full count.

### Root cause
`src/query-builder/SelectQueryBuilder.ts`: the distinct-id paginated-join branch is guarded by a truthiness check -- `if ((this.expressionMap.skip || this.expressionMap.take) && this.expressionMap.joinAttributes.length > 0)`. With `take(0)`, `expressionMap.take === 0` is falsy, so the branch is skipped. Execution falls to the plain path, which builds SQL via `createLimitOffsetExpression` -- but that only copies `skip`/`take` into `offset`/`limit` when `joinAttributes.length === 0`. With a join present, `limit` stays `undefined` -> no `LIMIT` clause -> all rows.

### Steps to reproduce
```ts
const [rows, count] = await ds.createQueryBuilder(Post, "post")
  .leftJoinAndSelect("post.tags", "tags")
  .take(0)
  .getManyAndCount();
// observed: rows.length === (all rows), count === (all)
// expected: rows.length === 0
```
The no-join path correctly emits `LIMIT 0` and returns `[]` (asserted by existing tests); the join case is untested and broken. `skip(0)` with a join is NOT affected (skip 0 is a semantic no-op).

### Suggested fix
Use `!= null` checks (mirroring the `hasLimit`/`hasOffset` helpers already used in `createLimitOffsetExpression`/`lazyCount`) instead of the truthiness check, so `take === 0` triggers the paginated path.

Develop-KIM:
### Description of change

`take(0)` combined with a join (e.g. `leftJoinAndSelect`) returned every row instead of an empty result, and `getManyAndCount` returned all rows as the page.

The join pagination branch in `SelectQueryBuilder` is guarded by `(this.expressionMap.skip || this.expressionMap.take)`. With `take(0)`, `take` is `0` (falsy), so the branch is skipped and execution falls through to the plain path — which only copies `skip`/`take` into `limit`/`offset` when there are no joins. With a join present no `LIMIT` is emitted, so all rows come back. The no-join path already emits `LIMIT 0` and returns `[]`.

The fix detects `take`/`skip` the same way the existing `lazyCount` helper in this file does: `take` is set when it is not `undefined`/`null` (so `0` counts as a limit), and `skip` is set only when it is `> 0` (offset `0` is a no-op, so `skip(0)` keeps returning all rows).

### How I verified

Added `test/github-issues/12666`. On `master` the two `take(0)` assertions fail (3 rows instead of 0); with the fix all pass, and `skip(0)` with a join still returns every row. Full sqlite suite passes (3031 passing, 0 failing).

Fixes #12666

_AI-assisted: drafted with Claude, reviewed and verified by me._

### Pull-Request Checklist

- [x] This pull request links a relevant issue using a closing keyword: `Fixes #12666`
- [ ] Documentation has been updated to reflect this change (`docs/docs/**.md`) — N/A, bug fix with no documented behavior change

### DIFF
--- a/src/query-builder/SelectQueryBuilder.ts
+++ b/src/query-builder/SelectQueryBuilder.ts
@@ -3488,8 +3488,15 @@ export class SelectQueryBuilder<Entity extends ObjectLiteral>
         // where we make two queries to find the data we need
         // first query find ids in skip and take range
         // and second query loads the actual data in given ids range
+        // a take of 0 is a meaningful limit (return no rows), so it must be
+        // detected explicitly rather than through truthiness; a skip of 0 is a
+        // no-op offset, so it is treated as unset (mirrors the checks in lazyCount)
         if (
-            (this.expressionMap.skip || this.expressionMap.take) &&
+            ((this.expressionMap.take !== undefined &&
+                this.expressionMap.take !== null) ||
+                (this.expressionMap.skip !== undefined &&
+                    this.expressionMap.skip !== null &&
+                    this.expressionMap.skip > 0)) &&
             this.expressionMap.joinAttributes.length > 0
         ) {
             // we are skipping order by here because its not working in subqueries anyway
--- a/test/github-issues/12666/entity/Category.ts
+++ b/test/github-issues/12666/entity/Category.ts
@@ -1,0 +1,10 @@
+import { Entity, PrimaryGeneratedColumn, Column } from "../../../../src"
+
+@Entity()
+export class Category {
+    @PrimaryGeneratedColumn()
+    id: number
+
+    @Column()
+    name: string
+}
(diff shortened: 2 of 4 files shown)


Develop-KIM:
Kept the regression test under `test/github-issues/12666/` since that's where issue-specific tests usually go in this repo. Can move it under `test/functional` if you'd prefer.

Develop-KIM:
Reopened as #12763 with the same commit — I'd deleted my fork when I closed this, so GitHub wouldn't let me reopen it here.

PR #12687, opened 2026-07-09
fix: honor take(0) when joins are present

### LINKED ISSUE #12666: take(0) combined with a join returns ALL rows instead of an empty array

voladoradepapantla-netizen:
Fixes #12666

Use a nullish check for skip/take in the joined pagination branch so take(0) still routes through the distinct-id path and returns an empty result. Includes a regression test for left joins.

### DIFF
--- a/src/query-builder/SelectQueryBuilder.ts
+++ b/src/query-builder/SelectQueryBuilder.ts
@@ -2995,8 +2995,7 @@ export class SelectQueryBuilder<Entity extends ObjectLiteral>
         ) {
             selectionPath = (
                 this.dataSource.driver as
-                    | AbstractSqliteDriver
-                    | ReactNativeDriver
+                    AbstractSqliteDriver | ReactNativeDriver
             ).wrapWithJsonFunction(selectionPath, column, false)
         }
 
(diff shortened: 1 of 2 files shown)


michaelbromley:

```
git fetch origin
```

PR #12690, opened 2026-07-10
fix(entity-manager): default invalidWhereValuesBehavior to throw on the write path

naorpeled:
Closes #12578. 

## Problem

`invalidWhereValuesBehavior`'s documented default is **throw**, and the read/find path already honours it — `SelectQueryBuilder.buildWhere` resolves each sub-option with `?? "throw"`. The **write path** did not: `OrmUtils.normalizeWhereCriteria` short-circuited with `if (!options) return criteria` when the option was unset, so `update` / `delete` / `softDelete` / `restore` with a plain-object criteria containing `null`/`undefined` ran silently (`col = NULL`) instead of throwing. That's the bug reported in #12578 (repro: https://github.com/samkessaram/typeorm-undefined-where-bug).

## Change

Default the options inside `normalizeWhereCriteria` (`options ??= {}`), so the existing per-key `options?.null ?? "throw"` / `?.undefined ?? "throw"` fallbacks apply. The write path now matches the read path:

- plain object `{ text: null }` / `{ text: undefined }` → **throws** by default
- array (OR) `[{ text: null }, { id: 1 }]` → **throws** by default
- nested plain `{ category: { name: null } }` → **throws** by default
- entity instances / `FindOperator` / `Date` / `Buffer` / primitives → **pass through** unchanged (the `isPlainObject` guard from #12629 runs before the key loop) — so `delete(entityInstance)` and the tree-repository soft-delete paths are unaffected
- configured `ignore` / `sql-null` → unchanged

No read-path change (already correct); the #12629 empty-criteria (`WHERE 1=1`) guards stay.

## Notes

- This is intended for **v1 / master** only. The legacy `v0` line keeps the lenient write default (it carries only #12629's empty-criteria fix).
- Deep validation of entity-instance criteria (defaulting entity nulls to `ignore` / strip) needs entity metadata and is intentionally left as a follow-up.

## Tests

- **Functional** — new `invalidWhereValuesBehavior default (unconfigured)` block: `delete`/`update`/`softDelete` throw on null/undefined by default; array-criteria null throws; entity-instance passthrough (no throw, deletes its row); read/write parity (`findBy` also throws).
- **Unit** — `normalizeWhereCriteria` now throws by default with no options (and with empty `{}` options); entity/`Date`/`Uint8Array`/primitive passthrough retained.
- Full suite audited locally on better-sqlite3 (persistence, query-builder, repository, tree-tables, relations/cascades/transaction/columns/embedded) — no pre-existing unconfigured plain-object null/undefined write relied on the old silent behaviour.

### DIFF
--- a/src/data-source/BaseDataSourceOptions.ts
+++ b/src/data-source/BaseDataSourceOptions.ts
@@ -219,7 +219,8 @@ export interface BaseDataSourceOptions {
     readonly isolateWhereStatements?: boolean
 
     /**
-     * Controls how null and undefined values are handled in find operations.
+     * Controls how null/undefined values in where criteria are handled by find
+     * and write methods (update/delete/softDelete/restore). Defaults to "throw".
      */
     readonly invalidWhereValuesBehavior?: InvalidFindOptionsWhereBehavior
 }
--- a/src/util/OrmUtils.ts
+++ b/src/util/OrmUtils.ts
@@ -656,10 +656,12 @@ export class OrmUtils {
     /**
      * Applies invalidWhereValuesBehavior to a plain-object where criteria: for
      * each own key a null/undefined value is thrown on, skipped, or converted
-     * to IsNull() per the configured behavior. Only plain FindOptionsWhere
-     * objects are normalized; anything else (entity/class instances,
-     * FindOperators, arrays, Date, Buffer, primitives) — and the criteria when
-     * no behavior is configured — is returned untouched.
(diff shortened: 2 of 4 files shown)


smith-xyz (approved):
lgtm! 

one thing to maybe note is that increment/decrement might bypass this error? Not sure if that was pre-existing or needs adjusted here

alumni (approved):

PR #12691, opened 2026-07-10
feat: add bun-sqlite driver (bun built-in sqlite module)

### LINKED ISSUE #11236: How can support bun native sqlite drive?
### Feature Description

I want to use bun [sqlite](https://bun.sh/docs/api/sqlite)

### The Solution

eg
```ts
    type: "bun-sqlite",
    database: "sqlite.db",
})
```

### Considered Alternatives

use sqlite or better-sqlite, but build as docker alpine image with bun cannot work.

### Additional Context

_No response_

- [ ] aurora-mysql
- [ ] aurora-postgres
- [X] better-sqlite3
- [ ] cockroachdb
- [ ] cordova
- [ ] expo
- [ ] mongodb
- [ ] mysql
- [ ] nativescript
- [ ] oracle
- [ ] postgres
- [ ] react-native
- [ ] sap
- [ ] spanner
- [X] sqlite
- [ ] sqlite-abstract
- [ ] sqljs
- [ ] sqlserver

trashhead:
### Description of change

Adds a new `bun-sqlite` driver that uses Bun's built-in `bun:sqlite` module.

**Why this is needed**

The existing `better-sqlite3` driver requires compiling a native Node.js addon (`better_sqlite3.node`), which fails in environments without build tools (MSVC on Windows, `node-gyp` on Linux/macOS). The Bun runtime ships SQLite as a first-class built-in — no native compilation, no extra dependencies.

**What was added**

- BunSqliteDriver.ts — driver extending `AbstractSqliteDriver`
- BunSqliteQueryRunner.ts — query runner extending `AbstractSqliteQueryRunner`
- BunSqliteDataSourceOptions.ts — `type: "bun-sqlite"` options type
- DriverFactory.ts — registered the new driver
- DriverUtils.ts — added `"bun-sqlite"` to `isSQLiteFamily()` (required so TypeORM uses `NULL` instead of unsupported `DEFAULT` in SQLite INSERT statements)
- DatabaseType.ts — added `"bun-sqlite"` to the union
- DataSourceOptions.ts — added `BunSqliteDataSourceOptions` to the union
- PlatformTools.ts — added `"bun:sqlite"` case to `load()`
- ormconfig.sample.json — added `bun-sqlite` entry with `skip: true` (opt-in; requires Bun runtime)
- DEVELOPER.md — added instructions for running `bun-sqlite` tests

**Key implementation differences from `better-sqlite3`**

| better-sqlite3 | bun:sqlite |
|---|---|
| `stmt.reader` | `stmt.columnNames.length > 0` |
| `db.pragma("key = val")` | `db.prepare("PRAGMA key = val").run()` |
| `db.pragma("key")` → array | `db.prepare("PRAGMA key").all()` |

**How it was verified**

51 tests passing under the Bun runtime, covering: WAL mode, column types, generated columns, views, UUID generation, JSON/JSONB, transactions, query builder, schema builder, SQL tags, bulk saves, and more. See DEVELOPER.md for how to run them.

### Pull-Request Checklist

- [x] This pull request links a relevant issue using a closing keyword: Fixes #11236
- [x] There are new or updated tests validating the change (`test/functional/driver/bun-sqlite/*.test.ts`)
- [x] Documentation has been updated to reflect this change (DEVELOPER.md)


trashhead:
Not sure why qodo still references old code.
How do I trigger a new review?

trashhead:
@qodo-free-for-open-source-projects review

package.json
```diff
@@ -51,6 +51,7 @@
     "./browser/connection/options-reader/ConnectionOptionsYmlReader.js": "./browser/platform/BrowserConnectionOptionsReaderDummy.js",
     "./browser/driver/aurora-data-api/AuroraDataApiDriver.js": "./browser/platform/BrowserDisabledDriversDummy.js",
```
smith-xyz, line 52: do we need a bunsqlite entry on BrowserDisabledDriversDummy.template?

src/driver/bun-sqlite/BunSqliteQueryRunner.ts
```diff
@@ -0,0 +1,222 @@
...
+
+        try {
+            const result = new QueryResult()
+
+            // bun:sqlite has no stmt.reader — use columnNames as indicator instead.
+            if (stmt.columnNames.length > 0) {
+                const raw = stmt
+                    .all(...normalizedParameters)
+                    .map((row: Record<string, unknown>) =>
+                        Object.fromEntries(
```
smith-xyz, line 125: isn't a stopper but this probably carries overhead on blob queries, not sure if we can do something like this to just skip when no blobs: 

```ts
const raw = stmt.all(...normalizedParameters)
if (raw.length > 0) {
    const blobKeys = Object.keys(raw[0]).filter(
        (k) => raw[0][k] instanceof Uint8Array,
    )
    if (blobKeys.length > 0) {
        for (const row of raw) {
            for (const k of blobKeys) {
                if (row[k] instanceof Uint8Array) {
                    (row as any)[k] = Buffer.from(row[k] as Uint8Array)
                }
            }
        }
    }
}
```

src/driver/bun-sqlite/BunSqliteDriver.ts
```diff
@@ -0,0 +1,202 @@
...
+     */
+    protected async attachDatabases() {
+        for (const { attachHandle, attachFilepathAbsolute } of Object.values(
+            this.attachedDatabases,
+        )) {
+            await this.createDatabaseDirectory(
+                path.dirname(attachFilepathAbsolute),
+            )
+            await this.dataSource.query(
+                `ATTACH "${attachFilepathAbsolute}" AS "${attachHandle}"`,
```
smith-xyz, line 189: we probably need to safe guard this better from sqli - not a blocker, you're matching what better sqlite3 already does

test/utils/test-setup.ts
```diff
@@ -2,13 +2,13 @@ import "chai/register-should"
 import "source-map-support/register"
 import "reflect-metadata"
 
-import chai from "chai"
+import { should, use } from "chai"
```
smith-xyz, line 5: is this for Bun compatibility?

trashhead: Yes.
When running TypeORM's test suite with Bun as the JavaScript runtime, using a default import of chai causes a runtime crash:

TypeScript compiles import chai from "chai" into const chai_1 = require("chai") and then accesses chai_1.default. In Node.js this works because its CJS interop sets .default to the module exports. Bun's CJS interop does not — chai_1.default is undefined, so calling [chai_1.default.should()] throws immediately before any tests run:
TypeError: undefined is not an object (evaluating 'chai_1.default.should')]

The fix is to use named imports instead

src/driver/bun-sqlite/BunSqliteQueryRunner.ts
trashhead: Optimized blob detections

package.json
trashhead: Implemented

michaelbromley:

```
git fetch origin
```

PR #12692, opened 2026-07-10
fix(entity-manager): validate where criteria in increment/decrement

naorpeled:
## Problem

`EntityManager.increment()` / `decrement()` (and the `Repository` methods that delegate to them) built their `UPDATE` directly:

```ts
this.createQueryBuilder(entityClass, "entity")
    .update(entityClass)
    .set(values)
    .where(conditions)   // conditions went straight to the low-level QB .where()
    .execute()
```

This bypassed `normalizeAndValidateWhereCriteria` — the guard that `update` / `delete` / `softDelete` / `restore` use. Two consequences:

- **Empty criteria ran an unfiltered write.** `increment(User, {}, "points", 1)` rendered `UPDATE user SET points = points + 1` with no `WHERE` — every row. This is the same `WHERE 1=1` mass-update class of bug the empty-criteria guard was added to prevent. (Intentional all-row updates already have `updateAll`.)
- **`invalidWhereValuesBehavior` was ignored** for `null`/`undefined` values in the criteria, unlike the other write methods.

Flagged in review of #12690.

## Change

Route `conditions` through `normalizeAndValidateWhereCriteria` in both `increment` and `decrement`, mirroring `update` exactly (including the `isPrimitive` → `whereInIds` branch). They now reject empty criteria and honor `invalidWhereValuesBehavior`, consistent with the other write methods.

No change to the QueryBuilder layer; low-level `.where()` remains intentionally exempt.

## Tests

- `repository-increment` / `repository-decrement`: empty `{}` criteria throws `Empty criteria(s)` and leaves all rows untouched; with `invalidWhereValuesBehavior: { null: "throw" }` configured, a `null` in the criteria throws `Null value encountered`.
- Full increment/decrement suites pass (18/18 locally on better-sqlite3).

## Note

This is a behavior change: an empty criteria that previously ran silently now throws — the intended, safer default (consistent with #12629 for `update`/`delete`).

### DIFF
--- a/src/entity-manager/EntityManager.ts
+++ b/src/entity-manager/EntityManager.ts
@@ -1470,8 +1470,8 @@ export class EntityManager {
 
     /**
      * Shared implementation of {@link increment} and {@link decrement}: builds a
-     * `column = column +/- value` UPDATE for the rows matched by the conditions,
-     * validating the criteria the same way the other write methods do.
+     * `column = column +/- value` UPDATE and delegates execution to {@link update},
+     * so the criteria handling stays aligned with the other write methods.
      */
     protected incrementOrDecrementBy<Entity extends ObjectLiteral>(
         operation: "increment" | "decrement",
@@ -1503,19 +1503,7 @@ export class EntityManager {
                     value,
             )
 
-        const { criteria: whereCriteria, isPrimitive } =
-            this.normalizeAndValidateWhereCriteria(conditions, operation)
-
-        const qb = this.createQueryBuilder<Entity>(entityClass as any, "entity")
-            .update(entityClass)
-            .set(values)
-        if (isPrimitive) {
-            qb.whereInIds(whereCriteria)
-        } else {
-            qb.where(whereCriteria)
-        }
-
-        return qb.execute()
+        return this.update(entityClass, conditions, values)
     }
 
     /**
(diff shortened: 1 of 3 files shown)


src/entity-manager/EntityManager.ts
```diff
@@ -1488,22 +1490,32 @@ export class EntityManager {
...
+        const { criteria: whereCriteria, isPrimitive } =
+            this.normalizeAndValidateWhereCriteria(conditions, operation)
+
+        const qb = this.createQueryBuilder<Entity>(entityClass as any, "entity")
             .update(entityClass)
             .set(values)
-            .where(conditions)
-            .execute()
+        if (isPrimitive) {
+            qb.whereInIds(whereCriteria)
```
alumni: This line would never be reached.

However, I'm thinking we could call `EntityManager.update` which would already handle the validation part.

alumni:
@naorpeled I'd simply make `increment`/`decrement` call `update`, since in reality they're nothing more than a specialized update (one column is updated in a certain way).

This would ease maintaining them in general in the future, since we wouldn't have to think about them anymore.

naorpeled:
@copilot address the feedback by @alumni

alumni (approved):

PR #12693, opened 2026-07-12
refactor(postgres): use promise returning query overload

Samuron:
Hello! I was monkey-patching typeorm to work with Bun.SQL and found out that some part of driver are using callback style query overload. It still is converted to promise, so I think this is a nice little change and also simplifies my monkey patch along the way. I tried to run the tests to best of my ability, seems that they are fine

### DIFF
--- a/src/driver/postgres/PostgresDriver.ts
+++ b/src/driver/postgres/PostgresDriver.ts
@@ -1833,11 +1833,7 @@ export class PostgresDriver implements Driver {
     protected executeQuery(connection: any, query: string) {
         this.dataSource.logger.logQuery(query)
 
-        return new Promise((ok, fail) => {
-            connection.query(query, (err: any, result: any) =>
-                err ? fail(err) : ok(result),
-            )
-        })
+        return connection.query(query)
     }
 
     /**
--- a/src/driver/postgres/PostgresDriver.ts
+++ b/src/driver/postgres/PostgresDriver.ts
@@ -430,7 +430,7 @@ export class PostgresDriver implements Driver {
      * Makes any action after connection (e.g. create extensions in Postgres driver).
      */
     async afterConnect(): Promise<void> {
-        const [connection, release] = await this.obtainMasterConnection()
+        const connection = await this.obtainMasterConnection()
 
         const installExtensions =
             this.options.installExtensions === undefined ||
@@ -458,7 +458,7 @@ export class PostgresDriver implements Driver {
             "12.0",
         )
 
-        await release()
+        connection.release()
     }
 
     protected async getAvailableExtensions(connection: any) {
@@ -1401,20 +1401,12 @@ export class PostgresDriver implements Driver {
      * Used for replication.
      * If replication is not setup then returns default connection's database connection.
      */
-    async obtainMasterConnection(): Promise<[any, Function]> {
+    obtainMasterConnection(): Promise<any> {
         if (!this.master) {
             throw new TypeORMError("Driver not Connected")
         }
 
-        return new Promise((ok, fail) => {
-            this.master.connect((err: any, connection: any, release: any) => {
-                if (err) {
-                    fail(err)
-                } else {
-                    ok([connection, release])
-                }
-            })
-        })
+        return this.master.connect()
     }
 
     /**
@@ -1422,24 +1414,14 @@ export class PostgresDriver implements Driver {
      * Used for replication.
      * If replication is not setup then returns master (default) connection's database connection.
      */
-    async obtainSlaveConnection(): Promise<[any, Function]> {
+    obtainSlaveConnection(): Promise<any> {
         if (!this.slaves.length) {
             return this.obtainMasterConnection()
         }
 
         const random = Math.floor(Math.random() * this.slaves.length)
 
-        return new Promise((ok, fail) => {
-            this.slaves[random].connect(
-                (err: any, connection: any, release: any) => {
-                    if (err) {
-                        fail(err)
-                    } else {
-                        ok([connection, release])
-                    }
-                },
-            )
-        })
+        return this.slaves[random].connect()
     }
 
     /**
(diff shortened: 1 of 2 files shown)


alumni:
@Samuron Thanks for this!

Did you encounter any issues with `Pool`? We currently use the callback APIs for both `Pool` and `PoolClient`, was wondering if it's necessary to change anything there :) 

Ideally we should move the entire driver to the async APIs in time :)

Samuron:
@alumni I deployed a couple of services with the version that didn't handle the callback in query case, so at least running migrations. synchronize schema for e2e tests and running regular queries via string, query builder and object api (like save or find) worked without needing callback. I only encountered the problem with the service that needed to install the uuid-ossp extension under the hood for now. I can check the driver for remaining cases with callback and update the PR if needed
Edit: I was not able to find any other usages of callback style queries in Postgres driver folder, but I didn't check outside of it

Samuron:
I also changed the pool connection to be promise based, claude run the tests and told me that they are ok, I have no idea what is going on there TBH.

But what is nice that it seems as it can even be a performance improvement:

1. minus one promise allocation from obtainConnection being async function
2. minus second promise  allocation from new Promise((ok,fail)) =>
3. minus one closure allocation (ok, fail) =>
4. minus one closure allocation from inside connect (err, connection, release) =>
5. minus one array allocation [connection, release]
6. plus one promise allocation from pg itself

Still, pg can be worse under the hood for the promise based method, so I am not sure about net benefit

Samuron:
## Allocation micro-benchmark: callback-wrapper vs promise-style `obtainMasterConnection`

> **Note:** This benchmark and analysis were produced by **Claude** (Anthropic's Claude Code, model Opus 4.8).

### Context

The Postgres driver's connection-acquisition path (`obtainMasterConnection` /
`obtainSlaveConnection`, plus `createPool` / `closePool`) wrapped `pg`'s
**callback** overload in a hand-rolled `new Promise`. Switching to `pg`'s native
**promise** overload removes, per acquisition:

- the wrapper `Promise`
- the `(ok, fail) => {}` executor closure
- the `(err, connection, release) => {}` connect callback closure
- the `[connection, release]` return-tuple array

### Method

A standalone micro-benchmark models a minimal `pg` Pool with **both** overloads
as `pg` actually behaves — callback form hands back a per-checkout `release`
closure; promise form allocates its own internal promise and exposes `release`
as a prototype method. Heap is measured GC-bracketed with `--expose-gc`, and GC
events during the measured loop are counted to confirm the deltas are real
allocated bytes rather than GC noise.

- **Runtime:** Node.js **v26.5.0**
- **(A) Retained heap/call:** result kept alive (what stays on the heap while the connection is held)
- **(B) Total alloc/call:** result dropped, includes transient garbage; `gc during loop = 0` in both cases

### Results

| Metric | old (callback wrapper) | new (promise direct) | reduction |
|---|---|---|---|
| Retained heap / call | 222.6 B | 152.4 B | **−70 B (−32%)** |
| Total alloc / call   | 659.0 B | 181.3 B | **−478 B (−72%)** |

- **Retained (−70 B/call):** old keeps the extra `[connection, release]` array + settled wrapper `Promise` + per-checkout release closure alive.
- **Total (−478 B/call):** old additionally churns the executor and connect-callback closures as short-lived garbage on every acquisition.

Stable across repeated runs (~70 B retained, ~478 B total), with `0` GC events
during the total-allocation loop, so the figures are uncontaminated.

### Caveats

- Absolute byte figures are **V8/Node-version specific** (object header sizes, hidden-class layout).
- The Pool is a **faithful model**, not real `pg` — treat these as the *shape and magnitude* of the win, not exact production numbers. The direction and relative reduction are robust.

<details>
<summary>Reproducible benchmark script (<code>node --expose-gc alloc-bench.mjs</code>)</summary>

```js
// Allocation micro-benchmark: old callback-wrapper vs new promise-style
// obtainMasterConnection path from TypeORM's PostgresDriver.
//
// Run with:  node --expose-gc alloc-bench.mjs
//
// It models a minimal `pg` Pool close to reality:
//   - callback form  pool.connect(cb) -> cb(null, client, release)
//        pg allocates a per-checkout `release` closure and hands it to cb.
//   - promise  form  pool.connect()   -> Promise<client>
//        pg allocates its own internal promise; `client.release` is a
//        prototype method (no per-checkout closure).
//
// This mirrors what each code path actually allocates on top of the driver.

import { PerformanceObserver } from "node:perf_hooks"

if (typeof global.gc !== "function") {
    console.error("Run with --expose-gc")
    process.exit(1)
}

// ---- minimal pg client with release on the prototype (promise form) ----
class PgClient {
    constructor(id) {
        this.id = id
        this.query = null // shape padding, like a real client field
    }
    release(_err) {
        /* return to pool - no-op in bench */
    }
}

// ---- mock Pool supporting BOTH the callback and promise overloads ----
let clientId = 0
class MockPool {
    // callback form: allocates a fresh per-checkout release closure (like pg)
    connectCb(cb) {
        const client = new PgClient(++clientId)
        const release = (err) => client.release(err) // per-checkout closure
        cb(null, client, release)
    }
    // promise form: pg builds an internal promise, release is prototype method
    async connectPromise() {
        return new PgClient(++clientId)
    }
}

// ---- OLD: hand-rolled Promise wrapping the callback overload ----
// new Promise + executor closure + connect callback closure + [c, r] array
function oldObtainMaster(pool) {
        pool.connectCb((err, connection, release) => {
            if (err) {
                fail(err)
            } else {
                ok([connection, release])
            }
        })
    })
}

// ---- NEW: return the driver promise directly ----
function newObtainMaster(pool) {
    return pool.connectPromise()
}

// ---------------------------------------------------------------------------

function gc2() {
    global.gc()
    global.gc()
}

// (A) RETAINED heap per call: hold every produced result alive, gc-bracketed.
async function measureRetained(label, run, N) {
    gc2()
    const retain = new Array(N)
    const before = process.memoryUsage().heapUsed
    for (let i = 0; i < N; i++) retain[i] = await run()
    const after = process.memoryUsage().heapUsed
    // keep `retain` alive across the measurement
    if (retain[N - 1] === undefined && retain[0] === undefined) console.log("")
    gc2()
    return { label, perCall: (after - before) / N, N }
}

// (B) TOTAL alloc per call incl. transient garbage: no retention, count GCs
//     that fire during the loop so we know if the delta is contaminated.
async function measureTotal(label, run, N) {
    let gcCount = 0
    const obs = new PerformanceObserver((list) => {
        gcCount += list.getEntries().length
    })
    obs.observe({ entryTypes: ["gc"] })

    gc2()
    const before = process.memoryUsage().heapUsed
    for (let i = 0; i < N; i++) {
        await run() // result dropped -> becomes garbage
    }
    const after = process.memoryUsage().heapUsed
    obs.disconnect()

    return { label, perCall: (after - before) / N, gcCount, N }
}

async function warmup(run, N) {
    for (let i = 0; i < N; i++) await run()
}

async function main() {
    const pool = new MockPool()
    const oldRun = () => oldObtainMaster(pool)
    const newRun = () => newObtainMaster(pool)

    const N_RET = 200_000
    const N_TOT = 20_000 // small enough to limit auto-GC during the loop

    // warm up JIT for both paths
    await warmup(oldRun, 50_000)
    await warmup(newRun, 50_000)

    console.log(`node ${process.version}\n`)

    console.log("== (A) retained heap per call (result kept alive) ==")
    const retOld = await measureRetained("old (callback wrapper)", oldRun, N_RET)
    const retNew = await measureRetained("new (promise direct)", newRun, N_RET)
    for (const r of [retOld, retNew])
        console.log(`  ${r.label.padEnd(26)} ${r.perCall.toFixed(1)} B/call`)
    console.log(
        `  -> retained delta: ${(retOld.perCall - retNew.perCall).toFixed(1)} B/call fewer with new\n`,
    )

    console.log(
        "== (B) total alloc per call incl. transient garbage (result dropped) ==",
    )
    const totOld = await measureTotal("old (callback wrapper)", oldRun, N_TOT)
    const totNew = await measureTotal("new (promise direct)", newRun, N_TOT)
    for (const r of [totOld, totNew])
        console.log(
            `  ${r.label.padEnd(26)} ${r.perCall.toFixed(1)} B/call   (gc during loop: ${r.gcCount})`,
        )
    console.log(
        `  -> total delta: ${(totOld.perCall - totNew.perCall).toFixed(1)} B/call fewer with new`,
    )
}

main()
```

</details>

*Benchmark authored and run by Claude (Anthropic).*

michaelbromley:

```
git fetch origin
```

PR #12704, opened 2026-07-17
fix(data-source): prevent connection leaks in DataSource.initialize()

### LINKED ISSUE #12705: DataSource.initialize() leaks connections on concurrent calls and on cache connection failure
### Issue description

`DataSource.initialize()` leaks database connections when called concurrently, and when the query result cache fails to connect

### Expected Behavior

1. Calling `initialize()` while a previous `initialize()` call is still in flight should not open a second driver connection/pool. It should either reject (like calling `initialize()` on an already-initialized DataSource does) or coalesce with the in-flight attempt.
2. If `queryResultCache.connect()` fails during `initialize()`, the driver connection that was opened just before it should be closed before the error propagates, leaving the DataSource fully un-initialized.

### Actual Behavior

1. **Concurrent `initialize()` calls both connect the driver.** The `isInitialized` guard is only assigned *after* `driver.connect()` resolves, so two overlapping calls both pass the check and each open a driver connection/pool. Both calls resolve successfully, `destroy()` only closes the pool the driver currently references, and the other pool leaks.

2. **A cache connection failure leaks the driver connection.** In `initialize()`, `queryResultCache.connect()` runs after `driver.connect()` but *before* the cleanup `try/catch` begins. If the cache fails to connect (e.g. Redis is unreachable), `initialize()` rejects with the driver still connected — and it cannot be released afterwards, because `destroy()` throws `CannotExecuteNotConnectedError` while `isInitialized` is still `false`.

### Steps to reproduce

Both leaks reproduce with plain TypeORM and any driver; `better-sqlite3` shown for a dependency-free example, but the same happens with pooled drivers like `postgres` (where the leak means orphaned server connections).

**1. Concurrent initialization:**

```ts
import { DataSource } from "typeorm"

    type: "better-sqlite3",
    database: ":memory:",
})

// e.g. two request handlers racing to lazily initialize the same DataSource
await Promise.all([dataSource.initialize(), dataSource.initialize()])
// driver.connect() has now run twice; both calls resolved successfully.
// With a pooled driver (postgres/mysql), one pool is orphaned and its
// connections are never released by destroy().
```

**2. Cache connection failure:**

```ts
import { DataSource } from "typeorm"

    type: "postgres",
    host: "localhost",
    username: "test",
    password: "test",
    database: "test",
    cache: {
        type: "redis",
        options: { socket: { host: "unreachable-host", connectTimeout: 1000 } },
    },
})

try {
    await dataSource.initialize() // rejects: Redis is unreachable
} catch (error) {
    // the postgres pool opened by driver.connect() is still open here
    await dataSource.destroy() // throws CannotExecuteNotConnectedError
}
// no way to release the leaked pool; the process keeps the sockets open
```

### My Environment

### Additional Context

The root cause is the ordering inside `DataSource.initialize()` (`src/data-source/DataSource.ts`): the `isInitialized` flag is only set after `driver.connect()` and `queryResultCache.connect()` have both succeeded, and the cleanup `try/catch` only covers the steps *after* the flag assignment. Everything between the initial guard and that point is unprotected against both re-entry and failure.

All drivers — the bug is in driver-agnostic `DataSource` code (reproduced with `better-sqlite3` and `postgres`).

Yes, I have the time, and I know how to start — PR incoming.

IbrahimHafez1:
Fixes #12705

### Description of change

`DataSource.initialize()` could leak database connections in two ways:

1. **Concurrent calls both connected the driver.** The `isInitialized` guard is only set *after* `driver.connect()` resolves, so two overlapping `initialize()` calls both passed the check and each opened a driver connection/pool — one of them leaked.
2. **A cache connection failure leaked the driver connection.** `queryResultCache.connect()` runs after the driver connects but *before* the existing cleanup `try/catch` began, so if the cache failed to connect, `initialize()` rejected with the driver still connected and no way to release it.

**New behavior:**

- A private `isInitializing` flag makes an overlapping `initialize()` call reject with `CannotConnectAlreadyConnectedError` — the same error already thrown when calling `initialize()` on an initialized DataSource, so no new API surface. The flag is cleared in a `finally`, so a *failed* initialize can still be retried, as before.
- The entire post-connect sequence is now covered by the cleanup `try/catch`. On failure **before** `isInitialized` is set (a cache connect failure), the driver is disconnected directly — `destroy()` cannot be used on that path because it throws `CannotExecuteNotConnectedError` when the DataSource is not initialized. A best-effort `queryResultCache.disconnect()` runs first (errors ignored) to release providers that partially connected before throwing, without masking the original error. On failure **after** `isInitialized` is set, behavior is unchanged: `destroy()` is called and the error re-thrown.

**Behavioral note for reviewers:** code that previously raced two `initialize()` calls and appeared to work (while silently double-connecting the driver) will now get a rejection on the second call. Rejecting is consistent with the existing double-initialize semantics; if you'd rather coalesce (return the in-flight promise so concurrent callers all resolve together), I'm happy to switch — rejecting seemed like the more conservative default.

**Verification:** two new regression tests in `test/functional/data-source/data-source.test.ts`, both using `better-sqlite3` in-memory so they need no external services, and both failing against `master`:

- *when cache initialization fails → should disconnect the database driver* — asserts the driver and the cache provider are disconnected and `isInitialized` stays `false`.
- *while initialization is in progress → should reject another initialization attempt* — gates `driver.connect()` to hold the first call mid-flight, asserts the second call rejects with `CannotConnectAlreadyConnectedError` and that `connect()` was only invoked once.

### Pull-Request Checklist

      `Fixes #NNNN`, `Closes #NNNN`, or `Resolves #NNNN` — Fixes #12705
- [x] Documentation has been updated to reflect this change (`docs/docs/**.md`) — noted in the DataSource docs that `initialize()` rejects concurrent or repeated calls and that a failed initialization can be retried

### DIFF
--- a/docs/docs/data-source/1-data-source.md
+++ b/docs/docs/data-source/1-data-source.md
@@ -8,6 +8,10 @@ establishes the initial database connection or connection pool depending on the 
 
 To establish the initial connection/connection pool, you must call the `initialize` method of your `DataSource` instance.
 
+`initialize` must only be called once: calling it again while a previous call is still in progress, or after it has
+succeeded, rejects with a `CannotConnectAlreadyConnectedError`. If initialization fails, the partially established
+connection is cleaned up and `initialize` can safely be called again.
+
 Disconnection (closing all connections in the pool) occurs when the `destroy` method is called.
 
 Generally, you call the `initialize` method of the `DataSource` instance on the application bootstrap,
--- a/src/data-source/DataSource.ts
+++ b/src/data-source/DataSource.ts
@@ -128,6 +128,8 @@ export class DataSource {
 
     readonly relationIdLoader: RelationIdLoader
 
+    private isInitializing = false
+
     // -------------------------------------------------------------------------
     // Constructor
     // -------------------------------------------------------------------------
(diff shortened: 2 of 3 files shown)


IbrahimHafez1:
Addressed the review feedback in 9c06c0c:

- **Cleanup masks original error** — fixed: the cleanup `driver.disconnect()` is now guarded so its failure cannot replace the original initialization error.
- **Test cleanup leaks driver** — fixed: the concurrency test's `finally` now falls back to a best-effort `driver.disconnect()` when the DataSource never reached the initialized state.
- **Tests missing issue reference** — added `#12705` reference comments to both new regression tests.
- **initialize() now rejects concurrently** — documented in `docs/docs/data-source/1-data-source.md`: `initialize()` rejects concurrent/repeated calls with `CannotConnectAlreadyConnectedError`, and a failed initialization can be retried.

On **"Connect failure still leaks"**: that path (`driver.connect()` itself failing after partially creating resources) is pre-existing behavior this PR doesn't change — the old code had the identical gap. Cleaning up partial state from a failed `connect()` is only safely done inside each driver's `connect()`, where the driver knows what was actually created; a blanket `disconnect()` from `DataSource` is not safe across all 15+ drivers. Happy to tackle the sqlite-family case in a follow-up PR if maintainers want it.

IbrahimHafez1:
Rebased onto current master; CI is re-running. This closes the connection leak when `initialize()` fails partway through. @alumni would you have a moment to take a look when convenient? Happy to narrow the scope if the lifecycle change needs more discussion.

michaelbromley:

```
git fetch origin
```

Issue #11823, opened 2025-12-05
Eager relations are joined twice


alumni:
When an eager relation is selected manually, we try to load the relation twice.

```
Editor.user: User (eager)
User.profile: Profile (eager)
User.nestedProfile: Profile (non-eager)
```

In the test `test/functional/relations/eager-relations/basic-eager-relations/basic-eager-relations.test.ts` we have:

```ts
await connection.manager.findOne(
    Editor,
    {
        where: { userId: 1 },
        relations: {
            user: {
                nestedProfile: true,
            },
        },
    },
)
```

which generates the query:

```sql
SELECT
    "Editor"."userId" AS "Editor_userId",
    "Editor"."postId" AS "Editor_postId",
    "Editor__Editor_user"."id" AS "Editor__Editor_user_id",
    "Editor__Editor_user"."firstName" AS "Editor__Editor_user_firstName",
    "Editor__Editor_user"."lastName" AS "Editor__Editor_user_lastName",
    "Editor__Editor_user"."deletedAt" AS "Editor__Editor_user_deletedAt",
    "Editor__Editor_user"."profileId" AS "Editor__Editor_user_profileId",
    "Editor__Editor_user"."nestedProfileId" AS "Editor__Editor_user_nestedProfileId",
    "Editor__Editor_user__Editor__Editor_user_nestedProfile"."id" AS "Editor__Editor_user__Editor__Editor_user_nestedProfile_id",
    "Editor__Editor_user__Editor__Editor_user_nestedProfile"."about" AS "Editor__Editor_user__Editor__Editor_user_nestedProfile_about",
    "Editor__Editor_user__Editor__Editor_user_profile"."id" AS "Editor__Editor_user__Editor__Editor_user_profile_id",
    "Editor__Editor_user__Editor__Editor_user_profile"."about" AS "Editor__Editor_user__Editor__Editor_user_profile_about",
    "Editor__user"."id" AS "Editor__user_id",
    "Editor__user"."firstName" AS "Editor__user_firstName",
    "Editor__user"."lastName" AS "Editor__user_lastName",
    "Editor__user"."deletedAt" AS "Editor__user_deletedAt",
    "Editor__user"."profileId" AS "Editor__user_profileId",
    "Editor__user"."nestedProfileId" AS "Editor__user_nestedProfileId",
    "Editor__user__profile"."id" AS "Editor__user__profile_id",
    "Editor__user__profile"."about" AS "Editor__user__profile_about"
FROM
[...]

PR #12715, opened 2026-07-18
fix: keep constructor-preset relation values on first hydration

### LINKED ISSUE #12683: Relation hydration silently dropped when entity constructor pre-initializes the relation property with a class instance (regression in 1.0 via #11267)
## Description

Since TypeORM 1.0, `RelationMetadata.setEntityValue()` merges the loaded relation
entity into an already-set relation property instead of assigning it
(introduced in #11267 to prevent eager-loaded relations from overwriting
manually requested nested relations):

```js
// RelationMetadata.setEntityValue
if (ObjectUtils.isObject(entity[propertyName])) {
    OrmUtils.mergeDeep(entity[propertyName], value);
} else {
    entity[propertyName] = value;
}
```

However `OrmUtils.merge` only merges **plain** objects (`isPlainObject`
guard, see #7774):

```js
static merge(target, source, memo = new Map()) {
    if (OrmUtils.isPlainObject(target) && OrmUtils.isPlainObject(source)) { ... }
    ...
}
```

Loaded relation values are always **class instances** (entity instances), and
when the target property already holds a class instance too, `mergeDeep` is a
silent no-op: the freshly hydrated relation entity is discarded and the stale
pre-set instance stays on the parent entity.

This breaks any entity whose constructor pre-initializes a relation property,
e.g.:

```ts
@Entity()
class Post {
    constructor() {
        this.author = new Author(); // sensible default for "create" flows
    }

    @ManyToOne(() => Author)
    @JoinColumn({ name: 'authorId' })
    author: Author;
}

// typeorm creates entities via the constructor during hydration (unless
// entitySkipConstructor is set), so `author` is already an Author instance...
const post = await repo.findOne({ where: {}, relations: { author: true } });
post.author; // ...and stays the EMPTY Author() from the constructor.
             // All hydrated Author columns are silently dropped.
```

On 0.3.x this worked: `setEntityValue` assigned unconditionally.

## Expected Behavior

`relations: { author: true }` returns the hydrated Author entity, as in 0.3.x.

## Actual Behavior

`post.author` is the empty instance created in the constructor; every column
of the loaded relation row is silently discarded (the SELECT contains the
columns; only the entity transformation drops them).

## Discussion / possible fixes

Note the constraint: the #11265 scenario currently *relies* on the no-op —
when the same to-one relation is hydrated twice (manual nested join first,
eager join second), the silent no-op is what preserves the deeper first
value. So a naive "assign when target is a class instance" would fix this
issue but re-open #11265 for that ordering.

Possible directions:

1. Distinguish "pre-existing value was set by this hydration run" from
   "pre-existing value came from the user/constructor". E.g. the transformer
   marks relation properties it has already written (a `Symbol` on the
   entity, or a per-run `Set<entity, propertyPath>`); first write assigns
   (restoring 0.3.x semantics for constructor presets), subsequent writes
   merge/no-op (keeping #11265 behavior).

2. Make the merge class-instance aware: when target and value are instances
   of the same entity, merge column values (and only overwrite relation
   properties that are `undefined` on the target). Heavier, touches
   `OrmUtils`.

Option 1 seems minimal and semantically precise. Happy to open a PR with a
regression test if you agree with a direction.

vishnujayvel:
# fix: preserve constructor-preset relation values on first hydration

Fixes #12683.

## Root cause (credit: @bredar's analysis in the issue)

`RelationMetadata.setEntityValue()` merges a freshly hydrated relation
into an already-set relation property instead of assigning it
(introduced in #11267 to keep a deep manually-requested nested relation
from being overwritten by a shallower eager auto-join for the same
relation):

```js
// RelationMetadata.setEntityValue
if (ObjectUtils.isObject(entity[propertyName])) {
    OrmUtils.mergeDeep(entity[propertyName], value)
} else {
    entity[propertyName] = value
}
```

However `OrmUtils.merge` only merges **plain objects** (`isPlainObject`
guard, see #7774). Loaded relation values are always entity **class
instances**, so when the target property already holds a class instance
too — e.g. an entity constructor pre-initializes it, a common pattern
for "create" flows:

```ts
@Entity()
class Post {
    constructor() {
        this.author = new Author() // sensible default for "create" flows
    }

    @ManyToOne(() => Author)
    @JoinColumn({ name: "authorId" })
    author: Author
}
```

...the merge is a **silent no-op**: the freshly hydrated `Author` row is
discarded and the stale, empty `Author` instance created in the
constructor stays on the parent entity. `post.author` ends up empty even
though the SELECT contains all the joined columns — only the entity
transformation step drops them.

## Fix

The constraint identified in the issue: the #11265/#11823 fix currently
*relies* on the no-op — when the same to-one relation is hydrated twice
in one query (e.g. a manually requested nested relation joined first,
then the same relation re-joined because it's also `eager: true`), the
no-op is what preserves the deeper, first-loaded value instead of letting
the shallower eager join clobber it.

This PR implements option 1 from the issue's discussion: distinguish
*"this is the first time this relation is hydrated for this entity in
this query run"* from *"this relation was already hydrated earlier in
this run"*.

`RawSqlResultsToEntityTransformer` now tracks, per query execution, which
`(entity, relation.propertyPath)` pairs have already been hydrated via a
join (`hydratedJoinRelations: WeakMap<entity, Set<propertyPath>>`).
`RelationMetadata.setEntityValue()` gains an optional `forceOverwrite`
parameter (default `false`, so every other call site is unaffected):

- **First** hydration of a relation on an entity within a run →
  `forceOverwrite = true` → always assigns, overwriting whatever was on
  the property before (a constructor default, `undefined`, whatever).
  This restores 0.3.x behavior and fixes #12683.
- **Subsequent** hydration of the *same* relation on the *same* entity
  within the *same* run → `forceOverwrite = false` → keeps the existing
  merge/no-op behavior, so the deeper first-loaded value from a manually
  requested nested relation is not clobbered by a later shallow eager
  auto-join. This preserves the #11265/#11823 fix.

The change is scoped to `RawSqlResultsToEntityTransformer`'s join
hydration path only; every other caller of `RelationMetadata.setEntityValue`
(`SubjectExecutor`, `PlainObjectToNewEntityTransformer`,
`SelectQueryBuilder`'s query-strategy relation loader, etc.) keeps its
existing behavior since `forceOverwrite` defaults to `false`.

## Test plan

Added `test/github-issues/12683/` (sqlite/better-sqlite3, in-memory,
follows the repo's functional test conventions):

- `Post` entity whose constructor pre-initializes `author = new Author()`.
- `should hydrate the requested relation even though the constructor
  pre-initializes it with an empty instance` — reproduces the reported
  bug: `findOneOrFail(Post, { relations: { author: true } })` now returns
  the fully hydrated `Author` (previously returned the empty
  constructor-created instance).
- `should not populate the relation when it was not requested` — sanity
  check that we only overwrite a relation we actually hydrate.

Regression coverage for the scenario the fix must **not** break — the
`#11265`/`#11823` double-hydration protection — verified against the
existing suite:

```
test/functional/relations/eager-relations/basic-eager-relations/basic-eager-relations.test.ts
  ✔ should load all eager relations when object is loaded
  ✔ should not load eager relations when query builder is used
  ✔ should preserve manually requested nested relations with DeleteDateColumn
  ✔ should not join eager relations twice when explicitly specified with DeleteDateColumn entity (issue #11823)
```

Revert-proof check: stashed the two source-file changes (kept the new
test), rebuilt, and confirmed the new test fails with
`AssertionError: expected undefined to equal 1` — i.e. `post.author.id`
stays `undefined`, showing the relation hydration really was silently
dropped. Restored the fix and reran; both new tests plus the full
`relations > eager relations` suite pass.

Ran the full `test/functional/relations/**` suite locally against
`better-sqlite3`: **3040 passing**, 175 pending (skipped for
non-configured drivers), **1 failing** — `RedisQueryResultCache
Integration` — pre-existing/unrelated: it times out because no local
Redis server is available in this environment, not caused by this
change.

Ran `eslint` and `prettier --check` on the touched files: no new errors
or warnings introduced (the pre-existing JSDoc/`no-explicit-any`
warnings on `RelationMetadata.ts` and
`RawSqlResultsToEntityTransformer.ts` are unchanged from `master`).


vishnujayvel:
Addressed both review findings:

- **Test location**: moved the regression coverage from `test/github-issues/12683/` to `test/functional/relations/constructor-preset-relation-hydration/` per the functional-suite preference (046ec1d).
- **Hardcoded IDs**: `prepareData()` now returns the saved entities, and both tests query/assert with the generated ids instead of assuming they start at 1.

Also synced the branch with `master` (9e9a754) and retitled the PR to conventional-commit format for the semantic check. Focused suite passes locally (2 passing); wider better-sqlite3 mocha run is green apart from two pre-existing, unrelated failures (Redis cache integration without a local Redis, and a `NO_COLOR` warning in the cli init test).

vishnujayvel:
Both automated review findings are addressed (moved the regression test into test/functional, and the test now asserts against the saved entities' generated ids instead of hardcoded 1s), and all checks are green. Let me know if a rebase against master would help get this reviewed.

bredar:
We retested with TypeORM 1.1.1 and confirmed that #12683 still reproduces. Our application currently needs a local compatibility patch to hydrate constructor-initialized relations correctly.

Is there anything still needed to move this PR forward? We’re happy to help validate the fix against our SQL Server application.

michaelbromley:

```
git fetch origin
```

PR #12718, opened 2026-07-19
fix(query-builder): keep getManyAndCount count accurate for paginated joins

### LINKED ISSUE #11744: Query Builder getManyAndCount() returns incorrect data and count when used in conjunction with pagination + join + a condition on the joined column
### Issue description

query builder getManyAndCount() returns incorrect results in typeorm 0.3.26+

### Expected Behavior

getManyAndCount() should return the same count as getCount()

### Actual Behavior

The count will be inaccurate and the page may be truncated

### Steps to reproduce

This bug requires a couple conditions for it to appear:
- Use query builder with pagination (take) and a join to another table.
- At least one of the returned entities on the page should have 0 related entries in the joined table.
- At least one of the returned entities on the page should have 2+ related entries in the joined table.

Minimal example

```
@Entity()
export class Sport {
  @PrimaryGeneratedColumn()
  id!: number;
  
  @Column()
  name?: string;

  @OneToMany(
    () => Player,
    (player) => player.sport,
  )
  players!: Players[];
}

@Entity()
export class Player {
  @PrimaryGeneratedColumn()
  id!: number;
  
  @Column()
  name?: string;

  @Column()
  sportId?: number;

  @ManyToOne(() => Sport, (sport) => sport.players)
  @JoinColumn({ referencedColumnName: 'id', name: 'sportId' })
  sport!: Sport;
}
```
Sports table entites:
| id | name |
| ---   | ---  |
| 1 | soccer |
| 2 | tennis |

Players table entities:
| id | name | sportId |
| ---   | ---  | --- |
| 1 | Alice | 1 |
| 2 | Bob | 1 |

query:
```
const results = await this.sportsRepo.createQueryBuilder('sports')
      .leftJoinAndSelect('sports.players', 'players')
      .addOrderBy('players.name', 'DESC')
      .take(2);
return results[1];
```

On typeorm 0.3.26+, this returns 1, which is incorrect.
On typeorm 0.3.25, this returns 2, which is correct.

### My Environment

| Node.js version     | 20.x  |
| Typescript version  |  5.x  |
| TypeORM version     | 0.3.26+  |

This bug is not present on typeorm 0.3.25

### Additional Context

_No response_

- [ ] aurora-mysql
- [ ] aurora-postgres
- [ ] better-sqlite3
- [ ] cockroachdb
- [ ] cordova
- [ ] expo
- [ ] mongodb
- [x] mysql
- [ ] nativescript
- [ ] oracle
- [ ] postgres
- [ ] react-native
- [ ] sap
- [ ] spanner
- [ ] sqlite
- [ ] sqlite-abstract
- [ ] sqljs
- [ ] sqlserver

kyungseopk1m:
### Description of change

`getManyAndCount()` returns a wrong count since 0.3.26 when a query combines `take`, a join, and an order by a joined column.

The lazy count added in #11524 infers the total from the number of loaded entities. For this kind of query the entities are loaded through the distinct-ids pagination, which keeps the ordered column in its `SELECT DISTINCT` list. A root row that has several joined rows then produces several distinct tuples, each one taking a `LIMIT` slot, so the page comes back with fewer roots than actually match and the inferred count ends up too low.

Example (two sports, one with two players and one with none):

```ts
const [rows, count] = await dataSource
    .createQueryBuilder(Sport, "sports")
    .leftJoinAndSelect("sports.players", "players")
    .addOrderBy("players.name", "DESC")
    .take(2)
    .getManyAndCount()
// count is 1 on 0.3.26+, was 2 on 0.3.25
```

The fix only infers the count from the page when every order by is a plain column of the main alias, whose value is fixed per primary key and cannot split the `DISTINCT`. A joined column, a computed or select-alias expression, or an embedded path now runs a real count query instead. Ordering by the main alias keeps the lazy path, so the optimization from #11524 is unchanged for the common case (still covered by the existing lazy-count tests).

This does not change the loaded page itself: `getMany()` can still return fewer roots than `take` for a to-many join ordered by the joined column. That is an older, separate pagination behaviour and is left out of scope here; this change only makes the count agree with `getCount()`.

Verified with a new test under `test/functional/query-builder/getManyAndCount-join-pagination/` that covers ordering by the joined column directly, through a select alias, and through a computed alias, plus the existing lazy-count suite.

### Pull-Request Checklist

- [x] This pull request links a relevant issue using a closing keyword: `Closes #11744`

### DIFF
--- a/src/query-builder/SelectQueryBuilder.ts
+++ b/src/query-builder/SelectQueryBuilder.ts
@@ -1920,13 +1920,29 @@ export class SelectQueryBuilder<Entity extends ObjectLiteral>
         const hasLimit =
             this.expressionMap.limit !== undefined &&
             this.expressionMap.limit !== null
-        if (this.expressionMap.joinAttributes.length > 0 && hasLimit) {
-            return undefined
-        }
-
         const hasTake =
             this.expressionMap.take !== undefined &&
             this.expressionMap.take !== null
+        const hasJoin = this.expressionMap.joinAttributes.length > 0
+
+        // A raw LIMIT on a joined query limits joined rows rather than root
+        // entities, so the loaded page size can never be trusted as the total.
+        if (hasJoin && hasLimit) {
+            return undefined
+        }
+
+        // `take` on a joined query loads entities through a separate distinct-ids
+        // subquery (see executeEntitiesAndRawResults). Any ordering value that is
+        // not functionally determined by the primary key is kept in the
+        // subquery's DISTINCT list, so a single root entity can occupy several
+        // LIMIT slots and the page is truncated below the number of matching
+        // roots — making the page size under-report the total. Only keep the
+        // lazy shortcut when every ordering is a plain main-alias column;
+        // otherwise fall back to a real count query.
+        // https://github.com/typeorm/typeorm/issues/11744
+        if (hasJoin && hasTake && !this.orderByIsPrimaryKeySafe()) {
+            return undefined
+        }
 
         // limit overrides take when no join is defined
         const maxResults = hasLimit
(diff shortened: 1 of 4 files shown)


michaelbromley:

```
git fetch origin
```

PR #12724, opened 2026-07-21
fix(driver): broadcast query events in capacitor and nativescript drivers

kyungseopk1m:
### Description

The Capacitor and NativeScript query runners never emit the `beforeQuery` / `afterQuery` subscriber events. Subscribers registered through these two drivers are therefore never invoked, unlike every other SQLite-family driver (better-sqlite3, cordova, expo, sqljs, react-native), all of which broadcast these events from their `query()` method.

This has been the case since both drivers were added; they never contained the broadcast calls.

### Fix

Add the `beforeQuery` / `afterQuery` broadcasts to each driver's `query()`, mirroring the existing sibling drivers.

- **Capacitor** already awaits the DB call inside an async method, so the added broadcasts follow the same shape as the other async runners.
- **NativeScript** wraps a callback-based native call, so the DB result is only available inside the callback. The `afterQuery` broadcast and `await broadcasterResult.wait()` are therefore placed inside the callback, right before resolving or rejecting, so that async subscribers settle before the query promise does. This is done on both the success and failure paths. The failure branch also gains a `return` that was missing: without it the handler fell through after `fail()` and produced a second, spurious success result (previously harmless only because resolving an already-rejected promise is a no-op, but it would now emit an incorrect success event).

### Tests

Added `test/unit/driver/query-broadcast-events.test.ts` covering Capacitor and NativeScript, success and failure, each asserting that an async `afterQuery` subscriber has fully settled before `query()` resolves/rejects. With the previous (broken) NativeScript ordering the success/failure cases resolve before the subscriber settles; with the fix all four pass. The rest of the unit suite passes unchanged.

These two drivers target mobile runtimes that are not exercised in CI, so verification is at the unit level against the broadcaster contract.

### DIFF
--- a/src/driver/capacitor/CapacitorQueryRunner.ts
+++ b/src/driver/capacitor/CapacitorQueryRunner.ts
@@ -3,6 +3,7 @@ import { QueryFailedError } from "../../error/QueryFailedError"
 import { QueryRunnerAlreadyReleasedError } from "../../error/QueryRunnerAlreadyReleasedError"
 import { QueryResult } from "../../query-runner/QueryResult"
 import { Broadcaster } from "../../subscriber/Broadcaster"
+import { BroadcasterResult } from "../../subscriber/BroadcasterResult"
 import { AbstractSqliteQueryRunner } from "../sqlite-abstract/AbstractSqliteQueryRunner"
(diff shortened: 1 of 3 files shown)


michaelbromley:

```
git fetch origin
```

PR #12647, opened 2026-07-01
fix: remove require() calls that break bundlers

alumni:
### Description of change

Removes `require()` calls that break bundlers.

Fixes #12559

### Pull-Request Checklist

- [x] Code is up-to-date with the `master` branch
- [x] This pull request links a relevant issue using a closing keyword:
      `Fixes #NNNN`, `Closes #NNNN`, or `Resolves #NNNN`
- [ ] There are new or updated tests validating the change (`tests/**.test.ts`)
- [x] Documentation has been updated to reflect this change (`docs/docs/**.md`)

> 🛡️ **Security fixes:** Do not submit security fixes as public PRs — the diff exposes the vulnerability. Use [GitHub Security Advisories](https://github.com/typeorm/typeorm/security/advisories/new) instead.

PR #12729, opened 2026-07-22
fix: use static require calls in PlatformTools.load for bundler compatibility

### LINKED ISSUE #12721: webpack bundling broken in v1.1
### Issue description

Recent changes break webpack bundles for pg. 

Update in comment below: I think it is broken for all build types.

### Expected Behavior

Changes made in this commit: (specifically the removed require statements) break webpack bundles https://github.com/typeorm/typeorm/pull/12647/changes

Previously, pg would be required by this file, but now that the import name is dynamic, the bundler does not know what to include so the pg package is missing in the compiled file, prompting error:

"Postgres package has not been found installed. Please run "npm install pg"." 

Tried manually importing pg using `import 'pg';`  but to no avail.

Does these recent changes mean that webpack is no longer supported by typeorm?

### Actual Behavior

Error on runtime

Postgres package has not been found installed. Please run "npm install pg".

### Steps to reproduce

Install typeom + pg
Build using webpack, here is a sample config

```js
module.exports = function (options) {
  const { plugins, ...config } = options;

  return {
    ...config,
    externals: [],
    plugins: [
      ...plugins,
      new webpack.optimize.LimitChunkCountPlugin({ maxChunks: 1 }),
    ],
  };
};
```

### My Environment

| Operating System    |   linux       |
| Node.js version     | 22.22.3  |
| Typescript version  | 6.0.3  |
| TypeORM version     | 1.1.1 |

### Additional Context

_No response_

- [ ] aurora-mysql
- [ ] aurora-postgres
- [ ] better-sqlite3
- [ ] capacitor
- [ ] cockroachdb
- [ ] cordova
- [ ] expo
- [ ] mongodb
- [ ] mysql
- [ ] nativescript
- [ ] oracle
- [ ] postgres
- [ ] react-native
- [ ] sap
- [ ] spanner
- [ ] sqlite
- [ ] sqljs
- [ ] sqlserver

h4sht:
## Problem

PR #12647 replaced explicit `require()` calls in `PlatformTools.load()` with a dynamic `require(name)`, which is invisible to static analyzers used by bundlers (Webpack, Vite, esbuild, etc.).

This causes errors like:
```
Postgres package has not been found installed. Please run "npm install pg".
```

...because the bundler does not know to include the database driver dependency in the bundle.

Fixes #12721

## Root Cause

Before PR #12647, `PlatformTools.load()` used explicit `require("pg")`, `require("mysql2")` calls that bundlers could statically analyze. After the PR, it was changed to `require(name)` where `name` is a runtime variable — invisible to static analysis.

When the bundler can't see which modules are needed, it excludes them from the bundle. The runtime check `if (!KNOWN_MODULES.includes(name))` validates the name but doesn't help the bundler.

## Solution

Replaced the dynamic `require(name)` with an explicit `switch/case` containing static `require()` calls for each known module. Each case has a literal string argument to `require()`, which bundlers can analyze.

The catch-all fallback preserves the global-install resolution path (when typeorm is installed globally but drivers are in the project).

### DIFF
--- a/src/platform/PlatformTools.ts
+++ b/src/platform/PlatformTools.ts
@@ -38,57 +38,62 @@ export class PlatformTools {
      * @returns the module
      */
     static load(name: string): any {
-        const KNOWN_MODULES = [
-            // AWS Aurora Data API (PostgreSQL/MySQL)
-            "typeorm-aurora-data-api-driver",
-            // better-sqlite3
-            "better-sqlite3",
-            // Expo
-            "expo-sqlite",
-            // Google Cloud Spanner
-            "@google-cloud/spanner",
-            // Microsoft SQL Server
-            "mssql",
-            // MongoDB
-            "mongodb",
-            // MySQL / MariaDB
-            "mysql2",
-            // Oracle
-            "oracledb",
-            // PostgreSQL
-            "pg",
-            "pg-native",
-            "pg-query-stream",
-            // React Native
-            "react-native-sqlite-storage",
-            // SAP HANA
-            "@sap/hana-client",
-            "@sap/hana-client/extension/Stream",
-            // sql.js
-            "sql.js",
-            // redis
-            "redis",
-            "ioredis",
-        ]
-
-        if (!KNOWN_MODULES.includes(name)) {
-            throw new TypeError(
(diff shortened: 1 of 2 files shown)


michaelbromley:

```
git fetch origin
```

PR #12730, opened 2026-07-22
fix: throw error when primitive value passed for relation in findBy

### LINKED ISSUE #12712: findBy with a primitive under a relation key silently returns all rows (no WHERE clause)
### Issue description

### Expected Behavior

### Actual Behavior

### Steps to reproduce

```js
require("reflect-metadata")

    name: "Author",
    columns: {
        name: { type: String },
    },
})
    name: "Post",
    columns: {
        title: { type: String },
    },
})

async function main() {
    const ds = new DataSource({
        type: "better-sqlite3",
        database: ":memory:",
        synchronize: true,
        entities: [Post, Author],
    })
    await ds.initialize()

    await posts.save([
    ])

    await ds.destroy()
}
main()
```

### Root cause

### My Environment

h4sht:
## Problem

When a primitive value (e.g. `1` or a string) is passed for a relation property in `findBy()` or `findOneBy()`, the `buildWhere()` method in `SelectQueryBuilder` silently ignores the condition. The JOIN is created but no WHERE filter is applied, causing the query to return **ALL rows** instead of filtering by the relation.

This is a silent data-corruption bug — particularly dangerous for multi-tenant applications where it may return records belonging to different users.

Fixes #12712

### Example

```ts
// Before: silently returns ALL posts (no filter applied)
await postRepository.findBy({ user: 1 })

// After: throws a clear error
// Error: Invalid value for relation property 'Post.user' in where condition.
// Expected an object with the relation's primary key(s) (e.g. { id: 1 }),
// but received a value of type 'number'.

// Correct usage:
await postRepository.findBy({ user: { id: 1 } })
```

## Changes

- `SelectQueryBuilder.buildWhere()`: Added a check after the null-handling block that throws a `TypeORMError` when a non-object (primitive) value is passed for a relation property
- Added functional test in `repository-find-options-relations.test.ts`

### DIFF
--- a/src/query-builder/SelectQueryBuilder.ts
+++ b/src/query-builder/SelectQueryBuilder.ts
@@ -4464,6 +4464,17 @@ export class SelectQueryBuilder<Entity extends ObjectLiteral>
                         continue
                     }
 
+                    // primitive values like 1 or "str" are not valid for
+                    // relation where conditions -- the user likely meant
+                    // { id: value }
+                    if (typeof where[key] !== "object") {
+                        throw new TypeORMError(
+                            `Invalid value for relation property '${alias}.${key}' in where condition. ` +
+                            `Expected an object with the relation's primary key(s) (e.g. { id: 1 }), ` +
+                            `but received a value of type '${typeof where[key]}'.`,
+                        )
+                    }
+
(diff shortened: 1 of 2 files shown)


michaelbromley:

```
git fetch origin
```

PR #12738, opened 2026-07-26
fix: preserve data when changing column length (3357)

luckyseoul:
## Summary
Fixes #3357 — length-only column changes previously forced `DROP COLUMN` + `ADD COLUMN`, destroying existing values (e.g. `varchar(50)` → `varchar(51)`).

## Changes
1. **Postgres / CockroachDB / Spanner** — length-only changes use `ALTER COLUMN … TYPE` (or equivalent) instead of drop+add.
2. **MySQL / Aurora MySQL / Oracle** — length is no longer a drop+add trigger; length-only updates go through `CHANGE` / `MODIFY`.
3. **`BaseQueryRunner.isColumnChanged`** — now treats `length` as a change so MySQL/Aurora/Oracle actually emit SQL (otherwise length-only would silent no-op after (2)).
4. **Regression test** — `test/github-issues/3357/issue-3357.ts` (Postgres path).

## Verified locally
- Postgres 16: `ALTER TABLE … ALTER COLUMN "example" TYPE character varying(51)` — data preserved
- MySQL 8: `ALTER TABLE … CHANGE \`example\` \`example\` varchar(51)` — data preserved
- Unit: `isColumnChanged` detects length-only vs same-length

## Related
- Opire bounty ~$590 on issue #3357


jumtalexalitch:
1. The main problem was that changing only the length of a column could cause TypeORM to drop the old column and create a new one. This could make the existing data disappear.


2. For PostgreSQL, CockroachDB, and Spanner, the fix is to change the column type directly using ALTER COLUMN ... TYPE instead of dropping and recreating the column.


3. For MySQL, Aurora MySQL, and Oracle, length-only changes should use CHANGE or MODIFY. This keeps the existing column and its data.


4. BaseQueryRunner.isColumnChanged() was also updated so that a change in the column length is properly detected.


5. A regression test was added for issue #3357 to make sure that changing something like varchar(50) to varchar(51) does not delete the existing data.



In short: the fix makes TypeORM modify the existing column instead of deleting and recreating it whenever only the column length has changed.

michaelbromley:

```
git fetch origin
```

PR #12739, opened 2026-07-27
Fix: Migration generation drops and creates columns instead of altering - data loss risk

EnderChest-YT:
## Summary

Fixes TypeORM issue #3357: "Migration generation drops and creates columns instead of altering resulting in data loss"

## Problem

The `changeColumn` method in `PostgresQueryRunner.ts` was dropping and recreating columns whenever the length changed, causing data loss. This happened even for safe length increases on string types.

## Solution

Added a condition to detect when only the column length changes and the change is safe (string types with length increase). In these cases, use `ALTER COLUMN TYPE` instead of `DROP COLUMN` + `ADD COLUMN`.

### Changes

- Modified `changeColumn` in `src/driver/postgres/PostgresQueryRunner.ts`
- Added check for safe length increases on string types (varchar, char, text, etc.)
- Uses `ALTER TABLE ... ALTER COLUMN ... TYPE` for safe changes
- Falls back to drop+add for unsafe changes (length decrease, type changes, etc.)

## Testing

- TypeScript compilation passes (`tsc --noEmit`)
- No existing tests modified — this is a targeted fix for the migration generation behavior

## Related

- Opire bounty: https://app.opire.dev/issues/01HWJNZ5HQMVG2TCW6XHQQJ3QT ($590)
- TypeORM issue: #3357


michaelbromley:

```
git fetch origin
```

PR #12742, opened 2026-07-29
fix(driver): ignore a length carried by types that accept none

### LINKED ISSUE #12741: PostgresDriver.createFullType appends a length to types that do not take one (uuid(36))
## Issue description

`PostgresDriver.createFullType()` appends `(length)` for **any** column that has `length` set, without checking whether the type accepts a length. The driver already declares which types do - `withLengthColumnTypes` - and `uuid` is (correctly) not among them. So `@Column({ type: "uuid", length: 36 })` renders as `uuid(36)`, which is not a valid PostgreSQL type.

`CockroachDriver` has the same shape. `MysqlDriver`, `OracleDriver` and `SqlServerDriver` are unaffected, because they route through `getColumnLength()`, whose type switch returns `""` for types that take no length.

## Expected Behavior

A `length` on a type that does not support one is ignored, so the rendered type stays `uuid` - consistent with the other drivers, and with the driver's own `withLengthColumnTypes` declaration.

## Actual Behavior

The rendered type is `uuid(36)`. Schema comparison then treats the column as permanently changed, and generated/synchronized DDL emits `ALTER TABLE ... TYPE uuid(36)`, which fails to parse.

## Steps to reproduce

Smallest form, straight against the driver:

```ts
const dataSource = new DataSource({ type: "postgres", /* ... */ });

dataSource.driver.createFullType({ type: "uuid" });               // "uuid"        ✔
dataSource.driver.createFullType({ type: "uuid", length: "36" }); // "uuid(36)"    ✘
```

Via an entity:

```ts
@Entity()
export class Example {
    id: string;

    @Column({ type: "uuid", length: 36 })
    ownerId: string;
}
```

`synchronize` / `migration:generate` then produces `uuid(36)` DDL.

## Root cause

`driver/postgres/PostgresDriver.ts` - `createFullType()` reads `column.length` directly:

```ts
let type = column.type;
if (column.length) {
    type += "(" + column.length + ")";
}
```

and `getColumnLength()` is a passthrough with no type awareness:

```ts
getColumnLength(column): string {
    return column.length ? column.length.toString() : "";
}
```

Compare the drivers:

| Driver | `createFullType` uses | Guarded? |
|---|---|---|
| mysql | `getColumnLength()` (type switch → `""` for unsupported) | yes |
| oracle | `getColumnLength()` | yes |
| sqlserver | `getColumnLength()` | yes |
| **postgres** | `column.length` directly | **no** |
| **cockroachdb** | `column.length` directly | **no** |

## Suggested fix

Gate the append on the list the driver already maintains:

```ts
if (column.length && this.withLengthColumnTypes.indexOf(column.type as ColumnType) !== -1) {
    type += "(" + column.length + ")";
}
```

or give `getColumnLength()` the same type awareness the mysql driver's has and use it, which would fix cockroachdb by the same change.

## Why it matters beyond a stray option

It removes the only way to express a uuid column width portably. MySQL has no uuid type, so `type: "uuid"` normalizes to `varchar`, and `MysqlDriver.getColumnLength` returns `36` **only** when `generationStrategy === "uuid"` - a plain (non-generated) uuid column falls through to the generic varchar default of **255**. Setting `length: 36` on the entity is the natural fix for that, and it works on MySQL - but the same entity then renders `uuid(36)` on PostgreSQL, so a codebase targeting both dialects cannot use it.

## My Environment

| Dependency | Version |
|---|---|
| Operating System | macOS (Darwin 25.5.0), also reproduced in Linux CI |
| Node.js | 24.15.0 |
| typeorm | 1.1.0 |
| Database | PostgreSQL 18 |

tada5hi:
Closes #12741

## What

A `length` on a column type that accepts none reaches validation, DDL rendering, and schema comparison. All three paths now ignore it consistently.

Every schema-backed driver consults its normalized `withLengthColumnTypes` allowlist before honoring a declared length. This keeps one shared entity portable across dialects: `{ type: "uuid", length: 36 }` becomes native `uuid` on PostgreSQL and MariaDB, while MySQL uses `varchar(36)`. Driver defaults such as generated UUID widths remain intact.

Schema comparison now uses `getColumnLength()` instead of raw metadata in PostgreSQL, CockroachDB, Aurora MySQL, SQLite, and React Native, closing the remaining perpetual-drift paths. PostgreSQL and CockroachDB also match direct `TableColumn` type names case-insensitively.

The public column-option docs now state that an unsupported length is ignored.

## Tests

- `getColumnLength` and `createFullType` unit matrix across PostgreSQL/Aurora, CockroachDB, MySQL/Aurora, Oracle, SQL Server, SAP, Spanner, SQLite, and React Native
- metadata validation across every constructible driver family
- functional synchronize-plus-schema-log coverage across the database CI drivers, including exact UUID type/length assertions for PostgreSQL, MySQL, and MariaDB

Local verification: 254 unit tests and 16 entity-metadata-validator/schema tests pass; `tsc --noEmit`, targeted Prettier, and ESLint complete with zero errors. Live PostgreSQL 18, MySQL 9, MariaDB 10.11, and sql.js runs all report zero follow-up schema queries.


tada5hi:
@alumni is there a timeline, when and if you intend to release this fix?

michaelbromley:

```
git fetch origin
```

PR #12743, opened 2026-07-29
fix(cli): include the database driver in projects generated by init

tada5hi:
### Description of change

`typeorm init --database <db>` currently generates a project that is missing its database driver, so the project fails on first run with the very error it tells you to fix:

```
DriverPackageNotInstalledError: Postgres package has not been found installed. Please run "npm install pg".
```

This affects every `--database` choice, not just postgres.

**Root cause**

`InitCommand.appendPackageJson` pins the generated project's driver version by reading it out of TypeORM's own `package.json`:

```ts
packageJson.dependencies["pg"] = ourPackageJson.devDependencies.pg
```

`copyPackageFile` in `gulpfile.ts` prunes the published `devDependencies`, and the allow-list only covers `@types/node`, `ts-node` and `typescript`. In the published package every driver lookup therefore resolves to `undefined`, and `JSON.stringify` silently drops keys whose value is `undefined`, so the dependency never makes it into the generated `package.json`.

`delete pkg.devDependencies` was introduced in #12224; #12281 restored the three tooling entries but not the drivers the init command also reads. This PR completes that allow-list.

**Current behaviour** with `npx typeorm init --name MyProject --database postgres` on `typeorm@1.1.0`:

```json
"dependencies": {
   "reflect-metadata": "^0.2.2",
   "typeorm": "1.1.0"
}
```

**New behaviour:**

```json
"dependencies": {
   "reflect-metadata": "^0.2.2",
   "typeorm": "1.1.0",
   "pg": "^8.22.0"
}
```

**Verification**

Built the package (`pnpm run package`) and ran the built CLI for every `--database` choice, asserting the generated `package.json`:

| `--database` | emitted dependency |
| --- | --- |
| postgres | `pg@^8.22.0` |
| mysql | `mysql2@^3.22.6` |
| mariadb | `mysql2@^3.22.6` |
| better-sqlite3 | `better-sqlite3@^12.11.1` |
| mssql | `mssql@^12.7.0` |
| oracle | `oracledb@^7.0.0` |
| mongodb | `mongodb@^7.5.0` |
| cockroachdb | `pg@^8.22.0` |
| spanner | `@google-cloud/spanner@^8.8.1` |

All nine emitted nothing before this change. I also installed and started a generated `better-sqlite3` project against the local build and confirmed `DriverPackageNotInstalledError` is gone.

**On tests:** the defect is in the packaging step (`gulpfile.ts`), which the mocha suite does not build or import. The suite runs against `build/compiled`, whereas this code path only exists in the published `build/package` layout. I did not find a way to cover it that wasn't more fragile than the fix. Happy to add something if you have a preferred approach.

A follow-up worth considering: the allow-list and `InitCommand`'s driver lookups are maintained by hand in two places, which is what let this drift twice. Deriving one from the other would close the gap for good. I left it out here to keep the fix minimal, but I'm glad to do it in this PR if you'd prefer.

### Pull-Request Checklist

- [ ] This pull request links a relevant issue using a closing keyword: reported in #12721, but that issue also covers a separate webpack/esbuild bundling problem that this PR does **not** address, so I have not used a closing keyword. Happy to split the issue if you prefer.
- [ ] There are new or updated tests validating the change (`tests/**.test.ts`): see note above
- [x] Documentation has been updated to reflect this change (`docs/docs/**.md`): N/A, no user-facing API change


michaelbromley:

```
git fetch origin
```

PR #12744, opened 2026-07-29
feat(mysql): use ALTER MODIFY instead of drop/add for safe column changes

parastejpal987-cmyk:
ran into this while debugging a migration that wiped a varchar column when i changed its length. turns out changeColumn() in the mysql driver does a full drop + add instead of just modifying the column in place. obviously that nukes all the data in that column which is... not great.

the fix is pretty straightforward — added a safeChangeColumn() helper that uses ALTER TABLE ... MODIFY COLUMN instead of dropping and recreating. mysql supports in-place column modification natively so theres no reason to destroy data for simple type/length/nullability tweaks.

**what changed:**
- `src/driver/mysql/MysqlQueryRunner.ts` — new `safeChangeColumn()` method, hooked into the existing `changeColumn()` flow for non-rename cases
- `tsconfig.json` — added explicit types array (was getting a build error on @types/mime)

both up and down queries are generated so migrations are still fully reversible.

compiled and tested locally, everything passes.
Closes #XXXX


michaelbromley:

```
git fetch origin
```

PR #12749, opened 2026-08-02
fix(postgres): read `coord_dimension` when loading spatial columns

### LINKED ISSUE #12747: PostgreSQL driver doesnt read Z and M Geometry suffixes from coord_dimension
### Issue description

When loading tables form DB and using PostGIS, PostgreSQL driver doesnt take into account coord_dimension column of geometry_columns table. Therefore when using specific geometry spatialFeatureType like `GeometryZ` TypeORM generates migration to this type over and over again.

### Expected Behavior

Translate coord_dimension and type columns of geometry columns table into correct `spatialFeatureType`
ie
|coord_dimension|type|spatialFeatureType|
|---|---|---|
|2|GEOMETRY|Geometry|
|3|GEOMETRY|GeometryZ|
|3|GEOMETRYM|GeometryM|
|4|GEOMETRY|GeometryZM|

### Actual Behavior

Only type field is taken into account, so if I create a table with `GeometryZ` column, generate migration and apply it, all following generated migrations will still have this change included, despite columns being already GeometryZ:
```ts
    await queryRunner.query(`ALTER TABLE "test" ALTER COLUMN "geometry" TYPE geometry(geometryz,5514)`);
}

    await queryRunner.query(`ALTER TABLE "test" ALTER COLUMN "geometry" TYPE geometry(GEOMETRY,5514)`);
}
```

### Steps to reproduce

* create entity with GeometryZ column
  `@Column({ type: "geometry", spatialFeatureType: "geometryz", srid: 5514 })`
* generate migration
* run migrations
* generate migration again
* see typeorm trying to ALTER COLUMN geometry again to GeometryZ, despite it being already GeometryZ

### My Environment

| Operating System    |   Debian GNU/Linux 12 (bookworm)|
| Node.js version     |v22.16.0  |
| Typescript version  | 5.8.3 |
| TypeORM version     | 1.1.0  |

### Additional Context

_No response_

- [ ] aurora-mysql
- [ ] aurora-postgres
- [ ] better-sqlite3
- [ ] capacitor
- [ ] cockroachdb
- [ ] cordova
- [ ] expo
- [ ] mongodb
- [ ] mysql
- [ ] nativescript
- [ ] oracle
- [x] postgres
- [ ] react-native
- [ ] sap
- [ ] spanner
- [ ] sqlite
- [ ] sqljs
- [ ] sqlserver

samuelmbabhazi:
### Description of change

Fixes #12747

When loading tables, the Postgres driver reads `srid` and `type` from the PostGIS `geometry_columns` view but ignores `coord_dimension`. PostGIS folds the Z dimension into `coord_dimension` and only keeps the M suffix in `type`, so a `geometry(PointZ)` column comes back as feature type `POINT`. The loaded table then never matches metadata declaring `PointZ`, and `migration:generate` produces the same ALTER over and over.

Verified against PostGIS 3.6, where the views report:

| declared column | view | `coord_dimension` | `type` |
| --- | --- | --- | --- |
| `geometry(Geometry)` | geometry_columns | 2 | `GEOMETRY` |
| `geometry(GeometryZ)` | geometry_columns | 3 | `GEOMETRY` |
| `geometry(GeometryM)` | geometry_columns | 3 | `GEOMETRYM` |
| `geometry(GeometryZM)` | geometry_columns | 4 | `GEOMETRY` |
| `geography(PointZ)` | geography_columns | 3 | `PointZ` |

The query now also selects `coord_dimension`, and the loader appends the `Z` or `ZM` suffix when the dimension calls for it and the suffix is not already present. The guard keeps `geography_columns` working unchanged, since that view already reports the full suffix in `type`.

`SpatialColumnOptions.spatialFeatureType` previously only accepted the plain GeoJSON type names, so a dimensional column could not even be declared without a cast. The type now also accepts the `Z`, `M` and `ZM` suffixed forms, and the decorator reference documentation mentions them.

The regression test declares geometry `PointZ`, `PointM` and `PointZM` columns plus a geography `PointZ` column, then asserts both that the loaded table reports the dimensional feature types and that the schema builder reports no pending changes, which is the loop from the issue. Reverting the source changes makes both tests fail. The full spatial test suite passes against the postgres-14 PostGIS image from docker-compose.

### CockroachDB

As requested in review, the same change is applied to the CockroachDB driver, with the same block shape so the two can be extracted together later. Probed on CockroachDB v25.2, the views match PostGIS except for one case:

| declared column | view | `coord_dimension` | PostGIS `type` | CockroachDB `type` |
| --- | --- | --- | --- | --- |
| `geometry(PointZ)` | geometry_columns | 3 | `POINT` | `POINT` |
| `geometry(PointM)` | geometry_columns | 3 | `POINTM` | `POINT` |
| `geometry(PointZM)` | geometry_columns | 4 | `POINT` | `POINT` |
| `geometry(GeometryZM)` | geometry_columns | 4 | `GEOMETRY` | `GEOMETRY` |
| `geography(PointZ)` | geography_columns | 3 | `PointZ` | `PointZ` |
| `geography(PointZM)` | geography_columns | 4 | `PointZM` | `PointZM` |

CockroachDB drops the M suffix from `geometry_columns`, which makes `PointM` indistinguishable from `PointZ` there. The driver already loads `crdb_sql_type` for every column (`GEOMETRY(POINTM,4326)`), so the declared feature type is taken from it when present and the `coord_dimension` logic stays as the shared fallback. The only other difference is that `coord_dimension` comes back as a bigint string, parsed the same way the existing code parses `srid`.

The CockroachDB spatial suite gains the same two tests (dimensional feature types loaded, no pending schema changes), and reverting the driver change makes exactly those two fail. Both spatial suites pass: 26 tests across postgres, cockroachdb and mysql.

### Pull-Request Checklist

### DIFF
--- a/src/driver/postgres/PostgresQueryRunner.ts
+++ b/src/driver/postgres/PostgresQueryRunner.ts
@@ -3951,7 +3951,7 @@ export class PostgresQueryRunner
                                 const sql =
                                     `SELECT * FROM (` +
                                     `SELECT "f_table_schema" "table_schema", "f_table_name" "table_name", ` +
-                                    `"f_${tableColumn.type}_column" "column_name", "srid", "type" ` +
+                                    `"f_${tableColumn.type}_column" "column_name", "srid", "type", "coord_dimension" ` +
                                     `FROM "${tableColumn.type}_columns"` +
                                     `) AS _ ` +
                                     `WHERE "column_name" = '${dbColumn["column_name"]}' AND ` +
@@ -3962,8 +3962,30 @@ export class PostgresQueryRunner
                                     await this.query(sql)
 
                                 if (results.length > 0) {
-                                    tableColumn.spatialFeatureType =
+                                    // "geometry_columns" folds the Z dimension into
+                                    // "coord_dimension" and only keeps the M suffix in
+                                    // "type", while "geography_columns" reports the
+                                    // full suffix in "type", so the suffix is only
+                                    // appended when it is missing.
+                                    let spatialFeatureType: string =
                                         results[0].type
+                                    const upperType =
+                                        spatialFeatureType.toUpperCase()
+                                    if (
+                                        results[0].coord_dimension === 3 &&
+                                        !upperType.endsWith("Z") &&
+                                        !upperType.endsWith("M")
+                                    ) {
+                                        spatialFeatureType += "Z"
+                                    } else if (
+                                        results[0].coord_dimension === 4 &&
+                                        !upperType.endsWith("Z") &&
+                                        !upperType.endsWith("M")
+                                    ) {
+                                        spatialFeatureType += "ZM"
+                                    }
+                                    tableColumn.spatialFeatureType =
+                                        spatialFeatureType
                                     tableColumn.srid = results[0].srid
                                 }
                             }
--- a/docs/docs/help/3-decorator-reference.md
+++ b/docs/docs/help/3-decorator-reference.md
@@ -165,7 +165,7 @@ export class User {
 - `hstoreType: "object"|"string"` - Return type of `HSTORE` column. Returns value as string or as object. Used only in [Postgres](https://www.postgresql.org/docs/9.6/static/hstore.html).
 - `array: boolean` - Used for postgres and cockroachdb column types which can be array (for example int[]).
 - `transformer: ValueTransformer|ValueTransformer[]` - Specifies a value transformer (or array of value transformers) that is to be used to (un)marshal this column when reading or writing to the database. In case of an array, the value transformers will be applied in the natural order from entityValue to databaseValue, and in reverse order from databaseValue to entityValue.
-- `spatialFeatureType: string` - Optional feature type (`Point`, `Polygon`, `LineString`, `Geometry`) used as a constraint on a spatial column. If not specified, it will behave as though `Geometry` was provided. Used only in PostgreSQL and CockroachDB.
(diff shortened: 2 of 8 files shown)


src/driver/postgres/PostgresQueryRunner.ts
```diff
@@ -3951,7 +3951,7 @@ export class PostgresQueryRunner
                                 const sql =
                                     `SELECT * FROM (` +
                                     `SELECT "f_table_schema" "table_schema", "f_table_name" "table_name", ` +
-                                    `"f_${tableColumn.type}_column" "column_name", "srid", "type" ` +
+                                    `"f_${tableColumn.type}_column" "column_name", "srid", "type", "coord_dimension" ` +
```
alumni, line 3954: Could we extend this to CockroachDB? It should be more-or-less compatible with PostGIS, they also support Z/M dimensions: https://docs.cockroachlabs.com/docs/stable/point

Over time, we'd like to extract the common code (like Sqlite), so it's good to have as little variance as possible.

samuelmbabhazi: Done in e4593cce. CockroachDB exposes the same views, so the block is the same shape as the Postgres one, with two differences I could not avoid after probing v25.2 (table in the PR description): `coord_dimension` comes back as a bigint string, parsed like the existing `srid`, and CockroachDB drops the M suffix from `geometry_columns`, which makes `PointM` indistinguishable from `PointZ` there. Since the driver already loads `crdb_sql_type` for every column, the declared feature type is taken from it when present and the `coord_dimension` logic stays as the shared fallback. The CockroachDB spatial suite gains the same two tests, and both spatial suites pass.

alumni (approved):

gioboa (approved):
Thanks @samuelmbabhazi 👏

PR #12750, opened 2026-08-02
perf: use performance.now() for query execution timing

### LINKED ISSUE #12731: Replace Date.now() with performance.now() in query() method
Replace `Date.now()` with `performance.now()` when measuring query execution time in `QueryRunner.query()`. `performance.now()` provides a monotonic, higher-resolution timer that's better suited for measuring elapsed durations.

pri12ya871:
### Description of change

`QueryRunner.query()` measured elapsed execution time with `Date.now()`. That is a wall-clock source, so an NTP correction or a manual clock change mid-query can make the measurement jump backwards, producing a negative or inflated duration. `performance.now()` is monotonic and higher resolution, which is what an elapsed-time measurement wants.

**Current behavior:** `queryExecutionTime` is derived from two `Date.now()` readings, at whole-millisecond resolution, and is not monotonic.

**New behavior:** the same measurement uses `performance.now()`. The value feeding the `maxQueryExecutionTime` comparison and `logQuerySlow` is now monotonic and sub-millisecond.

This replaces the 30 timing call sites across the 13 driver query runners.

`Date.now()` is deliberately left in place wherever the value is a real timestamp rather than an elapsed measurement — swapping those would be a bug, since `performance.now()` is relative to process start rather than the epoch:

- `src/cache/DbQueryResultCache.ts`
- `src/cache/RedisQueryResultCache.ts`
- `src/commands/CommandUtils.ts`
- `src/query-builder/SelectQueryBuilder.ts`

**How I verified it:** `pnpm run compile` passes, `eslint` reports no new errors on the changed files, and `prettier --check` is clean. No import is needed — `performance` has been a Node global since v16 and this package requires `^20.19.0 || ^22.13.0 || >=24.11.0`.

### One decision worth your call

`Date.now()` returns an integer; `performance.now()` returns a float. That value is interpolated straight into the slow-query log in `FileLogger`:

```ts
`[SLOW QUERY: ${message.additionalInfo?.time} ms]: ${message.message}`
```

so slow-query lines change from `[SLOW QUERY: 1234 ms]` to e.g. `[SLOW QUERY: 1234.5678901 ms]`. The `maxQueryExecutionTime` comparison stays correct; only the rendered text changes.

I left it unrounded because the issue asks for a higher-resolution timer and rounding to whole milliseconds would discard exactly that. Rounding at the log boundary instead would touch `AbstractLogger` and affect every logger implementation, which felt broader than this issue's scope. Happy to add rounding at either the measurement or the log boundary if you would prefer to keep the existing log format — just say which.

### Pull-Request Checklist

- [x] This pull request links a relevant issue using a closing keyword: Closes #12731
- [ ] There are new or updated tests validating the change (`tests/**.test.ts`) — N/A, see below
- [ ] Documentation has been updated to reflect this change (`docs/docs/**.md`) — N/A, no user-facing API change

On tests: this is a like-for-like swap of the timing source with no behavioral branch to assert, and the driver query paths need live database connections, so `test:fast` does not reach them. If you would like coverage, the meaningful place is a logger-level assertion on the slow-query path — glad to add one if you want it in this PR.

### DIFF
--- a/src/platform/PlatformTools.ts
+++ b/src/platform/PlatformTools.ts
@@ -31,22 +31,6 @@ export class PlatformTools {
     }
 
     /**
-     * @returns a timestamp in milliseconds suitable for measuring elapsed time.
-     *
-     * Prefers the monotonic, higher-resolution `performance.now()` where the
-     * runtime provides it, and falls back to `Date.now()` where it does not.
-     * Only ever use the returned value to compute a duration: its time origin
-     * is not the Unix epoch.
-     */
-    static performanceNow(): number {
-        const globalVariable = PlatformTools.getGlobalVariable()
-        if (typeof globalVariable?.performance?.now === "function") {
-            return globalVariable.performance.now()
-        }
-        return Date.now()
-    }
-
-    /**
      * Loads ("require"-s) given file or package.
      * This operation is only supported on the NodeJS platform
      *
--- a/src/platform/PlatformTools.ts
+++ b/src/platform/PlatformTools.ts
@@ -31,6 +31,22 @@ export class PlatformTools {
     }
 
     /**
+     * @returns a timestamp in milliseconds suitable for measuring elapsed time.
+     *
+     * Prefers the monotonic, higher-resolution `performance.now()` where the
+     * runtime provides it, and falls back to `Date.now()` where it does not.
+     * Only ever use the returned value to compute a duration: its time origin
+     * is not the Unix epoch.
+     */
+    static performanceNow(): number {
+        const globalVariable = PlatformTools.getGlobalVariable()
+        if (typeof globalVariable?.performance?.now === "function") {
+            return globalVariable.performance.now()
+        }
+        return Date.now()
+    }
+
+    /**
      * Loads ("require"-s) given file or package.
      * This operation is only supported on the NodeJS platform
      *
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,26 @@
+## [1.1.1](https://github.com/typeorm/typeorm/compare/1.1.0...1.1.1) (2026-09-01)
+
+### Bug Fixes
+
+- **better-sqlite3:** await query event subscribers and report failed queries ([#12752](https://github.com/typeorm/typeorm/issues/12752)) ([8cf2258](https://github.com/typeorm/typeorm/commit/8cf22582d32f9b33ff2f2f5ab5c5f98832ab0d84))
+- **cache:** release query runners created by clear and getFromCache ([#12797](https://github.com/typeorm/typeorm/issues/12797)) ([3883d76](https://github.com/typeorm/typeorm/commit/3883d764aa597985b0d93c9db0af502090fbdaea))
+- evaluate function defaults before enum/string normalization ([#12508](https://github.com/typeorm/typeorm/issues/12508)) ([c95b3bd](https://github.com/typeorm/typeorm/commit/c95b3bd8504c68c80122e1e9362f3fe569bf0143))
+- export `DateUtils` from the package entry point ([#12566](https://github.com/typeorm/typeorm/issues/12566)) ([9fed5b4](https://github.com/typeorm/typeorm/commit/9fed5b41502d05e0affd8ebd8c5da8a5b7728da7))
+- export `RelationMetadata` ([#12789](https://github.com/typeorm/typeorm/issues/12789)) ([9694a4a](https://github.com/typeorm/typeorm/commit/9694a4ad4796be29fdb30e189e6757b5ed3ed614))
+- guard metadata access in InsertQueryBuilder for raw-table upsert ([#12702](https://github.com/typeorm/typeorm/issues/12702)) ([281bfba](https://github.com/typeorm/typeorm/commit/281bfba42a5813b106bcf5ba79bc7f7251bd1e73))
+- **persistence:** keep many-to-many junction rows when the relation is not loaded ([#12711](https://github.com/typeorm/typeorm/issues/12711)) ([04ff4da](https://github.com/typeorm/typeorm/commit/04ff4daedcf60fa4ffd0d5d33bbafaac1a9bbc96))
+- **postgres:** read `coord_dimension` when loading spatial columns ([#12749](https://github.com/typeorm/typeorm/issues/12749)) ([1b2df90](https://github.com/typeorm/typeorm/commit/1b2df90bead65a3437f6ecf2c7a21c160b319872))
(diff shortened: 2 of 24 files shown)


pri12ya871:
Both findings were correct — fixed in a74a442a.

**1. Missing `performance.now` fallback.** This was the more serious of the two and I had reasoned about it wrongly: I checked `engines.node` and concluded `performance` was safe as a global, but that field describes the build environment, not the runtimes the Cordova / React Native / Expo / NativeScript / sql.js drivers actually execute in. Unguarded, this would have thrown before the query ran — turning a timing improvement into broken database access.

Added `PlatformTools.performanceNow()`, which prefers `performance.now()` where the runtime exposes it and falls back to `Date.now()` otherwise, and routed all 30 call sites through it. It goes via the existing `getGlobalVariable()` accessor to stay consistent with how this codebase already handles platform globals.

Two details worth flagging, since they were not in the review and would each have been silent failures:

- `BrowserPlatformTools.template` replaces `PlatformTools` in browser builds (per the `browser` mapping in `package.json`), so the helper is added there too. Adding it only to `PlatformTools.ts` would have left the browser bundle without it — precisely the targets this finding is about.
- 7 of the 13 query runners referenced `PlatformTools` only through `import type { ReadStream }`, which erases at runtime. Those now carry an explicit value import; reusing the type-only import would have compiled cleanly and failed in production.

**2. Falsy start-time check.** Correct, and I introduced it. `SqlServerQueryRunner` declares `queryStartTime: number | undefined` and guarded the error path with a truthiness check. `Date.now()` could never return `0`, so the guard was previously safe by accident; `performance.now()` can, which would have silently suppressed `queryExecutionTime` for after-query broadcasting and slow-query logging. Now an explicit `!== undefined` check.

`pnpm run compile` passes, eslint reports no errors on the changed files, and prettier is clean.

The open question from the original description still stands for a maintainer: `performance.now()` returns a float, and `FileLogger` interpolates it straight into `[SLOW QUERY: ${time} ms]`, so slow-query lines gain decimals. I left it unrounded because the issue asks for higher resolution, but happy to round at the measurement or the log boundary if you would rather preserve the existing output.

src/platform/PlatformTools.ts
```diff
@@ -30,6 +30,22 @@ export class PlatformTools {
...
 
+    /**
+     * @returns a timestamp in milliseconds suitable for measuring elapsed time.
+     *
+     * Prefers the monotonic, higher-resolution `performance.now()` where the
+     * runtime provides it, and falls back to `Date.now()` where it does not.
+     * Only ever use the returned value to compute a duration: its time origin
+     * is not the Unix epoch.
+     */
+    static performanceNow(): number {
```
alumni: `PlatformTools` (and `BrowserPlatformTools`) are for those utility functions that are different across platforms. Since the implementation is the same for both NodeJS and browsers, this functions could be moved maybe to `DateUtils`.

src/platform/PlatformTools.ts
```diff
@@ -30,6 +30,22 @@ export class PlatformTools {
...
+        const globalVariable = PlatformTools.getGlobalVariable()
```
alumni: I think `globalThis` can safely be used everywhere. `PlatformTools.getGlobalVariable()` is probably not necessary anymore.

```ts
globalThis.performance?.now() ?? Date.now()
```

alumni:
Hi @pri12ya871, thanks for the PR, only wrote a small comment.

Heads up for reviewers: there might be other PRs changing lines with `Date.now()`, check before merge.

Cprakhar:
@pri12ya871 You can round-off the time at logger level - only few changes.

src/platform/PlatformTools.ts
pri12ya871: Moved — thanks, that's the right call. `performanceNow()` is gone from both `PlatformTools.ts` and `BrowserPlatformTools.template` (both files are byte-identical to `master` again) and now lives in `DateUtils`, which the 13 query runners import instead. `DateUtils` already has unit coverage in `test/unit/util/date-utils.test.ts`, so it seems a natural home for it.

Done in a511c242.

src/platform/PlatformTools.ts
pri12ya871: Agreed, and applied — the helper is now just:

```ts
static performanceNow(): number {
    return globalThis.performance?.now() ?? Date.now()
}
```

`PlatformTools.getGlobalVariable()` is no longer involved. As a side effect this also settles the bot's "clock source may mismatch" note above: with no per-call resolution left, there is nothing that can change between the start and end readings.

pri12ya871:
Thanks both — pushed an update addressing all of this.

@Cprakhar rounding is now done at the logger level as you suggested (c3cdafa1). A single `Math.round()` in `AbstractLogger.logQuerySlow` covers both `additionalInfo.time` and the execution-time message, so every logger implementation keeps its existing output and `[SLOW QUERY: 1234 ms]` renders exactly as it did before. The `maxQueryExecutionTime` comparison and the `afterQuery` event keep the unrounded value, so no resolution is lost where it actually matters. If you would rather keep some sub-millisecond detail in the log, `Number(time.toFixed(2))` is a one-word change — just say so.

@alumni both of your comments are addressed in a511c242: the helper is out of `PlatformTools`/`BrowserPlatformTools.template` (both are byte-identical to `master` again) and into `DateUtils` as `globalThis.performance?.now() ?? Date.now()`. I have left the threads open for you to resolve.

I also added `test/unit/logger/abstract-logger.test.ts`, which asserts the rounding without needing a database connection, so it runs under `test:fast`. Locally: `pnpm run compile` is clean, `eslint` reports no new errors on the changed files, and the fast suite passes.

On the `possible-duplicate` label: #12756 covers the same driver call sites. This one predates it and additionally handles the integer-to-float change in the slow-query log format, plus test coverage. No strong feelings about which lands — happy to close mine or rebase on top of it, whichever the maintainers prefer.

pri12ya871:
Rebased onto current `master` (b495ed4). It replayed cleanly with no conflicts and the diff is unchanged — still 16 files, +123/-34.

@alumni that should also tie off your heads-up about other PRs touching these lines: on top of current `master` there are no `Date.now()` calls left in any `queryStartTime`/`queryEndTime` assignment, and the four remaining `Date.now()` uses in `src/` are the epoch timestamps listed in the description, which are intentionally left alone.

Cprakhar (approved):

gioboa (approved):
Thanks @pri12ya871 👏

pri12ya871:
Thanks for approving! Just checking if anything else is needed. If it's all good, could you please merge it when you get a chance?

michaelbromley:

```
git fetch origin
```

PR #12740, opened 2026-07-27
fix: use ALTER TABLE for column length changes instead of drop and recreate

qwenbona:
## Problem

When a column's `length` property changes (e.g., `varchar(255)` → `varchar(500)`), TypeORM drops and recreates the column, **destroying all existing data** in that column.

Closes #3357

## Solution

Modified all QueryRunner implementations to detect length-only changes and use `ALTER TABLE ... MODIFY/ALTER COLUMN` instead of dropping and recreating the column. This preserves existing data.

## Changes

- `src/driver/mysql/MysqlQueryRunner.ts` — Use `ALTER TABLE MODIFY` for length changes
- `src/driver/aurora-mysql/AuroraMysqlQueryRunner.ts` — Same
- `src/driver/postgres/PostgresQueryRunner.ts` — Use `ALTER TABLE ALTER COLUMN TYPE` for length changes
- `src/driver/cockroachdb/CockroachQueryRunner.ts` — Same
- `src/driver/oracle/OracleQueryRunner.ts` — Use `ALTER TABLE MODIFY` for length changes
- `src/driver/spanner/SpannerQueryRunner.ts` — Use `ALTER TABLE ALTER COLUMN` for length changes
- `test/functional/github-issues/3357/` — Regression test

## Testing

Added regression test that:
1. Creates a table with `varchar(255)` column
2. Inserts data
3. Changes length to `varchar(500)`
4. Verifies data is preserved after sync

PR #12717, opened 2026-07-19
fix: use ALTER COLUMN TYPE instead of DROP ADD for column length changes

elevasyncsolutions-jpg:
## Summary
Fixes #3357 — migration generation drops and recreates columns when only the length changes, causing data loss.

## Problem
When a column length changes (e.g., varchar(50) → varchar(51)), every driver unconditionally performs DROP COLUMN + ADD COLUMN, destroying all existing data.

## Fix
For length-only changes, generate ALTER COLUMN TYPE instead of DROP+ADD.

## Changes
- **Postgres/CockroachDB/Spanner**: Removed length from destructive DROP+ADD; added safe ALTER COLUMN TYPE handler
- **MySQL/AuroraMySQL/Oracle/SAP/SQL Server**: Removed length from DROP+ADD; handled via BaseQueryRunner.isColumnChanged()
- **BaseQueryRunner**: Added length comparison to isColumnChanged()
- **Tests**: Added column-length-safe functional test

Closes #3357

PR #12751, opened 2026-08-03
fix(postgres): use alter column type for length-only changes

zhangzhen9434:
## Summary

Fixes typeorm/typeorm#3357 — migration generation drops and recreates columns when only the length changes, causing data loss.

## ✅ Code Review Status

- **Qodo (AI code review)**: ✅ Approved on both passes
  - Pass 1: Found collation edge case (fixed)
  - Pass 2: "There are currently no active Qodo findings on this PR. The changes look aligned with the reported regression."
- **semantic (commit title)**: ✅ Approved
- **issue-link (PR ↔ issue)**: ✅ Approved

## Changes

**PostgresQueryRunner.ts**:
- Remove the length comparison from the destructive `DROP+ADD` trigger in `changeColumn`
- Add `newColumn.length !== oldColumn.length` to the existing precision/scale handler, so length-only changes generate the safe `ALTER COLUMN ... TYPE` query instead of `DROP+ADD`
- **Edge case fix** (caught by Qodo): use `createFullType(newColumn)` in the collation block so the length modifier is preserved when both length and collation change in the same migration

**test/functional/schema-builder/column/change/column-length-safe-alter/** (new):
- Functional test that asserts no `DROP COLUMN` / `ADD COLUMN` is emitted for length-only changes
- Asserts that `ALTER COLUMN "name" TYPE` is generated instead

## Why a focused Postgres-only PR?

Several other PRs (#12532, #12541, #12543, #12544, #12657, #12717, #12738, #12739, #12740, #12744) are attempting multi-driver fixes. This PR is intentionally scoped to **Postgres only** so that:

1. The Postgres fix can land independently if/when the broader PR stalls
2. The diff is minimal and easy to review
3. PostgreSQL is the most widely affected (referenced in the original bug report)

The same pattern can be applied to other drivers in follow-up PRs.

## Edge case fix (per Qodo)

When both `length` and `collation` change in the same migration, the collation block previously used `newColumn.type` (which drops the `(length)` modifier), overwriting the length set by the earlier `ALTER COLUMN ... TYPE` block. This regression was introduced by removing the length comparison from the `DROP+ADD` trigger. The fix uses `this.driver.createFullType(newColumn)` so the length modifier is preserved.

## References

- Opire bounty: $590 — https://app.opire.dev/issues/01HWJNZ5HQMVG2TCW6XHQQJ3QT
- Issue: #3357
- Qodo review: https://github.com/typeorm/typeorm/pull/12751#issuecomment-5161976693

## Test Plan

- [x] Qodo automated code review passed
- [x] functional test `column-length-safe-alter.test.ts` covers the regression
- [ ] Run `npm test -- --grep "column length safe alter"`
- [ ] Manual verify with a migration that changes `varchar(50)` to `varchar(51)`:
  - Before: `DROP COLUMN` + `ADD COLUMN` (data loss)
  - After: `ALTER COLUMN "x" TYPE varchar(51)` (preserves data)


zhangzhen9434:
@qodo-free-for-open-source-projects[bot] /review

I've addressed the collation edge case in commit ae9f64c:
- `PostgresQueryRunner.ts:2348-2371` now uses `createFullType(newColumn)` in both up and down queries for the collation block, preserving the length/precision/scale/array modifiers

Also added a functional test:
- `test/functional/schema-builder/column/change/column-length-safe-alter/` asserts no `DROP COLUMN` / `ADD COLUMN` is generated for length-only changes

Re-review would be appreciated.

PR #12754, opened 2026-08-04
fix: use ALTER COLUMN instead of drop/create for type and length changes

parastejpal987-cmyk:
Closes #3357 

### What changed?
Currently, if you change the `length`, `type`, or `isArray` property of an existing column, TypeORM's migration generator creates a `DROP COLUMN` and `ADD COLUMN` query. This causes silent and catastrophic data loss in production. 

This PR updates the `changeColumn` implementation for Postgres, CockroachDB, and Spanner. Instead of dropping the column, it now correctly generates an `ALTER TABLE ... ALTER COLUMN ... TYPE ...` query for these modifications to preserve data. We fallback to the old drop/recreate behavior *only* if the `generatedType` expression changes (since Postgres still requires a hard recreate for STORED generated columns).

### Testing
- Verified boolean conditions correctly bypass drop/recreate for `type`/`length` updates and correctly push `ALTER COLUMN` queries in the drivers.
- Ran formatting and checks.


michaelbromley:

```
git fetch origin
```

PR #12759, opened 2026-08-06
fix(sqljs): measure query execution time after the statement runs

### LINKED ISSUE #12758: sqljs measures query execution time before the statement runs
### Issue description

On the sqljs driver, query execution time is measured before the statement runs, so `maxQueryExecutionTime` never fires and the `afterQuery` event carries a wrong `executionTime`.

### Expected Behavior

`executionTime` covers the time the query actually took, and a query slower than `maxQueryExecutionTime` reaches `logQuerySlow`.

Every other query runner takes `queryEndTime` after the execution call has returned. Postgres measures after `await databaseConnection.query(...)`, better-sqlite3 after `stmt.all()` / `stmt.run()`, expo after `await statement.executeAsync()`, react-native and nativescript inside the completion callback, spanner after `await executor.run(...)`.

### Actual Behavior

`SqljsQueryRunner.query()` takes `queryEndTime` right after `prepare()` and `bind()`. In sql.js the statement runs while `step()` is iterated, and that loop sits below the measurement, so only the preparation is timed.

```ts
const queryStartTime = Date.now()
    statement = databaseConnection.prepare(query)
    statement.bind(parameters)
const queryEndTime = Date.now()          // measurement ends here
    if (slow) logQuerySlow(...)          // and the check runs here
    while (statement.step()) { ... }     // the query actually runs here
    broadcastAfterQueryEvent(..., queryExecutionTime, ...)
```

A query that takes 393 ms of wall clock reports 3 ms, and never reaches `logQuerySlow`:

```
sqljs
  afterQuery executionTime   3
  actual elapsed           393
  logQuerySlow             not called

better-sqlite3 (same query, same options)
  logQuerySlow             150
  afterQuery executionTime 150
  actual elapsed           150
```

### Steps to reproduce

```js
const { DataSource } = require('typeorm')

// all of the work happens inside step(), so preparation and execution are separable
const slow = `WITH RECURSIVE counter(x) AS (
    SELECT 1 UNION ALL SELECT x + 1 FROM counter WHERE x < 2000000
) SELECT count(*) AS total FROM counter`

class Recorder {
    logQuery() {}
    logQueryError() {}
    logSchemaBuild() {}
    logMigration() {}
    log() {}
    logQuerySlow(time) {
        console.log('  logQuerySlow', time)
    }
}

class Sub {
    afterQuery(event) {
        console.log('  afterQuery executionTime', event.executionTime)
    }
}

async function run(type, extra) {
        type,
        maxQueryExecutionTime: 50,
        logger: new Recorder(),
        entities: [],
        ...extra,
    })
    await dataSource.initialize()
    dataSource.subscribers.push(new Sub())

    console.log(type)
    const started = Date.now()
    await dataSource.query(slow)
    console.log('  actual elapsed', Date.now() - started)

    await dataSource.destroy()
}

run('sqljs').then(() => run('better-sqlite3', { database: ':memory:' }))
```

### My Environment

| Operating System    | macOS 15.7.7 |
| Node.js version     | 22.16.0  |
| Typescript version  | 5.9.3    |
| TypeORM version     | master (04ff4dae) |
| sql.js version      | 1.14.1   |

### Additional Context

Two things break because of this.

1. `maxQueryExecutionTime` is effectively inert on sqljs. Preparation is sub-millisecond, so the threshold is never crossed no matter how slow the query is.
2. Anything consuming `AfterQueryEvent.executionTime` for instrumentation gets a value that has nothing to do with the query.

Issue #10180 covered a different case, mysql2 queueing the query after `queryStartTime` was taken, and was closed by the stale bot rather than resolved. This one is the opposite end of the same measurement, and is specific to sqljs.

PR #12750 changes the clock source on these two lines from `Date.now()` to `PlatformTools.performanceNow()`. It leaves the measurement point where it is, so it does not affect this.

kyungseopk1m:
### Description of change

Fixes #12758.

`SqljsQueryRunner.query()` took `queryEndTime` right after `prepare()` and `bind()`. In sql.js the statement runs as `step()` is iterated, and that loop sits below the measurement, so only the preparation was timed.

Two things follow from that. `maxQueryExecutionTime` is inert on sqljs, because preparation is sub-millisecond and never crosses the threshold no matter how slow the query is. And `AfterQueryEvent.executionTime` carries a number that has nothing to do with the query, which matters for anything using it to instrument.

The fix moves the row collection above the measurement. That is the order every other query runner already uses: postgres measures after `await databaseConnection.query(...)`, better-sqlite3 after `stmt.all()` / `stmt.run()`, expo after `await statement.executeAsync()`, react-native and nativescript inside the completion callback, spanner after `await executor.run(...)`. I read all eleven runners that take a `queryEndTime`, and sqljs was the only one measuring before execution, so the change is limited to this file.

`queryStartTime` is left alone. Every driver starts the clock before preparation, so the start point already matches the convention.

### Before and after

Same query, `maxQueryExecutionTime: 50`, better-sqlite3 as the control. Repro script is in the issue.

```
before
  sqljs            executionTime   4   elapsed 437   logQuerySlow not called
  better-sqlite3   executionTime 158   elapsed 158   logQuerySlow 158

after
  sqljs            executionTime 423   elapsed 424   logQuerySlow 423
  better-sqlite3   executionTime 149   elapsed 149   logQuerySlow 149
```

### Side effect worth naming

If `step()` throws, the slow-query log is now skipped and the failure goes to `logQueryError` and the failed `afterQuery` event. Previously a slow `prepare()` could log the query as slow even though it never ran. The other runners do not log a slow query on a failed execution either, so this brings sqljs in line rather than changing behaviour independently.

### Tests

`test/functional/driver/sqljs/query-execution-time.test.ts` plus a subscriber that records `executionTime`. Both assertions fail on master:

```
1) should measure the time spent executing the query, not just preparing it:
   AssertionError: expected 3 to be at least 33.666666666666664
2) should report a slow query once it exceeds maxQueryExecutionTime:
   AssertionError: expected [] to have a length of 1 but got +0
```

The timing assertion compares against the caller's own measurement rather than a fixed millisecond threshold, so machine speed cancels out. A fixed threshold would let a slow runner pass the test on preparation time alone. The bound is deliberately loose: the caller's window is wider than the measured one, since it also covers the `BeforeQuery` broadcast before `queryStartTime` and the subscriber settling after `queryEndTime`, and time spent there should not be able to fail a correct measurement.

### Note on #12750

#12750 rewrites these same two lines, swapping `Date.now()` for `PlatformTools.performanceNow()` across every driver. It changes the clock source and not the measurement point, so it does not fix this and this does not fix that. Whichever lands second will need a small rebase on these two lines. Happy to be the one to rebase.

### DIFF
--- a/test/functional/driver/sqljs/query-execution-time.test.ts
+++ b/test/functional/driver/sqljs/query-execution-time.test.ts
@@ -50,7 +50,9 @@ describe("sqljs driver > query execution time", () => {
             dataSources.map(async (dataSource) => {
                 const subscriber = dataSource
                     .subscribers[0] as QueryExecutionTimeSubscriber
+                const logger = dataSource.options.logger as RecordingLogger
                 subscriber.clear()
+                logger.slowQueries.length = 0
 
                 const startedAt = Date.now()
                 await dataSource.query(slowQuery)
@@ -70,20 +72,11 @@ describe("sqljs driver > query execution time", () => {
                 // left loose so that time spent outside the measured section
                 // cannot fail a correct measurement.
                 expect(executionTimes[0]).to.be.at.least(elapsed / 3)
-            }),
-        ))
 
-    it("should report a slow query once it exceeds maxQueryExecutionTime", () =>
-        Promise.all(
-            dataSources.map(async (dataSource) => {
-                const logger = dataSource.options.logger as RecordingLogger
-                logger.slowQueries.length = 0
-
-                await dataSource.query(slowQuery)
-
+                // the same measurement decides whether the query is slow
                 expect(logger.slowQueries).to.have.lengthOf(1)
                 expect(logger.slowQueries[0].query).to.equal(slowQuery)
-                expect(logger.slowQueries[0].time).to.be.greaterThan(1)
+                expect(logger.slowQueries[0].time).to.equal(executionTimes[0])
             }),
         ))
 })
--- a/test/functional/driver/sqljs/query-execution-time.test.ts
+++ b/test/functional/driver/sqljs/query-execution-time.test.ts
@@ -39,7 +39,8 @@ describe("sqljs driver > query execution time", () => {
             entities: [],
             enabledDrivers: ["sqljs"],
             subscribers: [QueryExecutionTimeSubscriber],
-            driverSpecific: { maxQueryExecutionTime: 1 },
+            // above the few milliseconds prepare() takes on its own
+            driverSpecific: { maxQueryExecutionTime: 50 },
             createLogger: () => new RecordingLogger(),
         })
     })
--- a/test/functional/driver/sqljs/query-execution-time.test.ts
+++ b/test/functional/driver/sqljs/query-execution-time.test.ts
@@ -1,0 +1,89 @@
+import "reflect-metadata"
+import { expect } from "chai"
+import type { DataSource } from "../../../../src/data-source/DataSource"
+import type { Logger } from "../../../../src/logger/Logger"
+import {
+    closeTestingConnections,
+    createTestingConnections,
+} from "../../../utils/test-utils"
+import { QueryExecutionTimeSubscriber } from "./subscribers/QueryExecutionTimeSubscriber"
+
+/**
+ * sql.js runs the statement while `step()` is iterated, not while it is
+ * prepared. A query whose work lives entirely in that loop therefore separates
+ * preparation time from execution time.
+ */
+const slowQuery = `WITH RECURSIVE counter(x) AS (
+    SELECT 1 UNION ALL SELECT x + 1 FROM counter WHERE x < 500000
+) SELECT count(*) AS total FROM counter`
+
+class RecordingLogger implements Logger {
+    readonly slowQueries: { time: number; query: string }[] = []
+
+    logQuery(): void {}
+    logQueryError(): void {}
+    logSchemaBuild(): void {}
+    logMigration(): void {}
+    log(): void {}
+
+    logQuerySlow(time: number, query: string): void {
+        this.slowQueries.push({ time, query })
+    }
+}
+
+describe("sqljs driver > query execution time", () => {
+    let dataSources: DataSource[]
+
+    before(async () => {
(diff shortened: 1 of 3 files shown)


kyungseopk1m:
Done in 85caa09cd.

test/functional/driver/sqljs/query-execution-time.test.ts
```diff
@@ -0,0 +1,82 @@
...
+
+describe("sqljs driver > query execution time", () => {
+    let dataSources: DataSource[]
+
+    before(async () => {
+        dataSources = await createTestingConnections({
+            entities: [],
+            enabledDrivers: ["sqljs"],
+            subscribers: [QueryExecutionTimeSubscriber],
+            driverSpecific: { maxQueryExecutionTime: 1 },
```
official-burak: Prepare was 3-4ms in #12758. A 1ms threshold fires on that, so this does not prove `step()` is inside the timer. Use something above prepare noise (50).

official-burak (changes requested):
The `step()` move is right and matches the other runners.

The test does not lock it in. `maxQueryExecutionTime: 1` is below the 3-4ms prepare time from the issue, so `logQuerySlow` can fire on master and the last three asserts would still pass after a revert. `elapsed / 3` is the only check that would fail.

Raise the threshold above prepare noise (50ms, same as the issue) so a timer that still stops after `bind()` cannot trip it.

test/functional/driver/sqljs/query-execution-time.test.ts
kyungseopk1m: Thanks for catching that. You're right, the threshold was below prepare time so the slow-query asserts passed either way. Raised it to 50ms in 3fa990ec1. Reverting the step() move now fails on those asserts too, not just the timing one.

alumni (approved):

Cprakhar (approved):
Thanks @kyungseopk1m 🙏

official-burak (approved):
The 50ms threshold is above prepare noise, so the slow-query asserts now fail if step() is still outside the timer. Looks good.

PR #12761, opened 2026-08-06
fix(postgres): use ALTER COLUMN TYPE instead of DROP/ADD for type and length changes

parastejpal987-cmyk:
So I ran into this issue today where TypeORM was generating migrations that dropped and recreated columns just because I changed a varchar length. That obviously wipes out all the data in the column which is pretty bad.

The problem is in `PostgresQueryRunner.changeColumn()` — when either the type or length is different between old and new column, it just does a `DROP COLUMN` + `ADD COLUMN` with the comment "to avoid data conversion". But Postgres can handle most of these cases natively with `ALTER COLUMN ... TYPE ... USING`, which keeps all the existing data intact.

I split the condition into two cases:

1. Changes that genuinely need a drop/recreate — toggling `isArray`, or adding a `STORED` generated expression. These are structurally incompatible so the old behaviour is correct there.

2. Plain type or length changes — for these I now emit `ALTER TABLE t ALTER COLUMN c TYPE new_type USING c::new_type` instead. Both the up and down queries are generated so rollbacks work fine too.

Before this change, bumping a `varchar(100)` to `varchar(200)` would silently delete every value in that column. After the fix it just alters the type in place and nothing is lost.

Fixes #3357


michaelbromley:

```
git fetch origin
```

PR #12762, opened 2026-08-06
fix(schema-builder): drop dependent indices and constraints for removed columns.

Delorne007:
### Description of change
- **What the change is intended to do:** Automatically drops dependent indices, unique constraints, and check constraints linked to database columns right before those columns are removed during schema synchronization.
- **Why this change is needed:** Prevents database query execution errors or failures that happen when a schema synchronization tries to drop a column that still has active constraints or indices pointing to it.
- **How you've verified it:** Inspected the code structure inside `RdbmsSchemaBuilder.ts`, verified the sequential execution order of schema operations, and reviewed a clean diff showing 58 target additions.


michaelbromley:

```
git fetch origin
```

PR #11137, opened 2024-11-18
fix(query-builder): wrap inner joins under left joins correctly

harm-less:
### Description of change

Fixes https://github.com/typeorm/typeorm/issues/3080. After 6 years of knowing this issue exists, it's finally no more!

This [StackOverflow](https://stackoverflow.com/questions/28590173/sql-inner-joining-to-left-joined-table) still explain the case fairly well. Simply put, if the nested `INNER JOIN` returns 0 rows, the entire query will by definition return 0 results even if the `FROM` table or the `LEFT JOIN `had results. So the `ON` condition of the parent `LEFT JOIN` wraps the `INNER JOIN` in order to prevent this from happening.

We are running this on a Postgres DB.

This is what the PR will do to queries. I took an advanced example from our application.

Before (current TypeORM version):
```sql
inner join "public"."publication_links" "pubLink" on
    "pubLink"."id" = "variant"."publicationId"
    and ("pubLink"."deletedAt" is null)
inner join "public"."merchants" "merchant" on
    "merchant"."id" = "pubLink"."retailerId"
    and ("merchant"."deletedAt" is null)
inner join "public"."merchant_channels" "merchantChannel" on
    "merchantChannel"."retailerId" = "merchant"."id"
    and ("merchantChannel"."isActive" = $1)
inner join "public"."channel_data" "channelData" on
    "channelData"."id" = "merchantChannel"."channelId"
    and (("channelData"."slug" = $2)
        and "channelData"."deletedAt" is null)
left join "public"."affiliate_metadata" "affiliateMeta" on
    "affiliateMeta"."retailerId" = "merchant"."id"
    and ("affiliateMeta"."affiliateId" = $3)
left join "public"."ab_test_variants" "abTestVariant" on
    "abTestVariant"."publicationVariantsId" = "variant"."id"
left join "public"."ab_tests" "abTest" on
    "abTest"."id" = "abTestVariant"."publicationAbTestsId"
    and ("abTest"."deletedAt" is null)
left join "public"."ab_test_channels" "abTestChannel" on
    "abTestChannel"."publicationAbTestsId" = "abTest"."id"
left join "public"."channel_references" "channelRef" on
    "channelRef"."id" = "abTestChannel"."channelsId"
    and (("channelRef"."slug" = $4)
        and "channelRef"."deletedAt" is null)
left join "public"."variant_channels" "variantChannel" on
    "variantChannel"."publicationVariantId" = "variant"."id"
    and ("variantChannel"."isActive" = $5)
[...]

PR #12765, opened 2026-08-07
fix(mssql): place table lock hint inside nested join parentheses

### LINKED ISSUE #12764: MSSQL: lock table hint emitted outside nested join parentheses, producing invalid SQL (1.1.0 regression)
### Issue description

mssql + TypeORM 1.1.0: a lock combined with a two-level relation join emits WITH (NOLOCK) outside the nested join's parentheses, so the query fails with Incorrect syntax near the keyword 'with'. 1.1.0 regression.

### Expected Behavior

A query with a relation chain of depth >= 2 and `setLock()` should produce valid T-SQL,
with the table hint attached to each table reference.

### Actual Behavior

The hint is appended after the closing `)` of the nested join, where a table hint is not
allowed, and the inner table loses its hint entirely. SQL Server fails with:

```
QueryFailedError: Error: Incorrect syntax near the keyword 'with'. If this statement is a
common table expression, an xmlnamespaces clause or a change tracking context clause, the
previous statement must be terminated with a semicolon.
```

### Steps to reproduce

Entities (three distinct entities, plain many-to-one — nothing special about them):

```ts
@Entity() class Country { @PrimaryColumn() id: number }

@Entity() class Profile {
  @PrimaryColumn() id: number
  @Column() countryId: number
  @ManyToOne(() => Country) @JoinColumn({ name: 'countryId' }) country: Country
}

@Entity() class User {
  @PrimaryColumn() id: number
  @Column() profileId: number
  @ManyToOne(() => Profile) @JoinColumn({ name: 'profileId' }) profile: Profile
}
```

Depth 1 — **valid**:

```ts
repo.createQueryBuilder('u').leftJoinAndSelect('u.profile', 'p').setLock('dirty_read')
```
```sql
FROM "user" "u" WITH (NOLOCK) LEFT JOIN "profile" "p" WITH (NOLOCK) ON "p"."id"="u"."profileId"
```

Depth 2 — **invalid**:

```ts
repo.createQueryBuilder('u')
  .leftJoinAndSelect('u.profile', 'p')
  .leftJoinAndSelect('p.country', 'c')
  .setLock('dirty_read')
```
```sql
FROM "user" "u" WITH (NOLOCK) LEFT JOIN ("profile" "p" LEFT JOIN "country" "c" WITH (NOLOCK) ON "c"."id"="p"."countryId") WITH (NOLOCK) ON "p"."id"="u"."profileId"
--                            ^ hint missing on "p"                                                                    ^ hint illegal here
```

Also reproduces through find options, so it is not query-builder specific:

```ts
repo.find({ relations: { profile: { country: true } }, lock: { mode: 'dirty_read' } })
```

And with the other two lock modes — `pessimistic_read` → `WITH (HOLDLOCK, ROWLOCK)`,
`pessimistic_write` → `WITH (UPDLOCK, ROWLOCK)` — in the same wrong position.

Depth 2 *without* a lock is valid, so the parenthesised nesting itself is fine; only the
hint placement is wrong.

### My Environment

| Dependency | Version |
| --- | --- |
| operating system | Linux |
| typeorm | 1.1.0 (works on 1.0.0 and 0.3.29) |
| driver | mssql |

### Additional Context

Regression appeared between 1.0.0 and 1.1.0: on 1.0.0 and 0.3.29 the same code emits every
join flat, with no parentheses, so the hint is always attached to a table reference.

- [ ] aurora-mysql
- [ ] aurora-postgres
- [ ] better-sqlite3
- [ ] capacitor
- [ ] cockroachdb
- [ ] cordova
- [ ] expo
- [ ] mongodb
- [x] mysql
- [ ] nativescript
- [ ] oracle
- [ ] postgres
- [ ] react-native
- [ ] sap
- [ ] spanner
- [ ] sqlite
- [ ] sqljs
- [x] sqlserver

lazerg:
Fixes #12764

### Description of change

On SQL Server, a relation chain two levels deep combined with `setLock()` produced invalid T-SQL. The table hint landed after the closing parenthesis of the nested join, where SQL Server does not allow one, and the inner table was left without a hint at all:

```sql
LEFT JOIN ("profile" "p" LEFT JOIN "country" "c" WITH (NOLOCK) ON ...) WITH (NOLOCK) ON ...
```

The query then failed with `Incorrect syntax near the keyword 'with'`. Nesting the joins in parentheses came in with #11137, but `buildJoinClause` still appended the hint after `postfix`, so once a join had children the hint drifted outside the parentheses instead of staying on its own table reference. Moving the call in front of `childJoins` puts the hint straight after the alias it belongs to, which is where the flat (depth 1) case already put it:

```sql
LEFT JOIN ("profile" "p" WITH (NOLOCK) LEFT JOIN "country" "c" WITH (NOLOCK) ON ...) ON ...
```

Verified with a new test in `test/functional/query-builder/locking`, which fails on `master` and passes here. The full suite runs green against SQL Server 2025.

### Pull-Request Checklist

      `Fixes #12764`
- [ ] Documentation has been updated to reflect this change (`docs/docs/**.md`) N/A, bug fix with no documented behavior change

### DIFF
--- a/src/query-builder/SelectQueryBuilder.ts
+++ b/src/query-builder/SelectQueryBuilder.ts
@@ -2576,9 +2576,9 @@ export class SelectQueryBuilder<Entity extends ObjectLiteral>
             tableName +
             " " +
             this.escape(alias) +
+            this.createTableLockExpression() +
             childJoins +
             postfix +
-            this.createTableLockExpression() +
             (condition ? " ON " + condition : "")
         )
     }
--- a/test/functional/query-builder/locking/query-builder-locking.test.ts
+++ b/test/functional/query-builder/locking/query-builder-locking.test.ts
@@ -246,6 +246,28 @@ describe("query builder > locking", () => {
         }
     })
 
+    // https://github.com/typeorm/typeorm/issues/12764
+    it("should attach dirty read lock statement to every table of a nested join", () => {
+        for (const dataSource of dataSources) {
+            if (!(dataSource.driver.options.type === "mssql")) {
+                return
+            }
+
+            const sql = dataSource
+                .createQueryBuilder(Post, "post")
+                .innerJoinAndSelect("post.categories", "categories")
+                .innerJoinAndSelect("categories.images", "images")
+                .setLock("dirty_read")
+                .getSql()
+
+            expect(sql).to.contain(
+                'INNER JOIN ("category" "categories" WITH (NOLOCK)',
+            )
+            expect(sql).to.contain('INNER JOIN "image" "images" WITH (NOLOCK)')
+            expect(sql).not.to.contain(") WITH (NOLOCK)")
+        }
+    })
+
     it("should not attach pessimistic write lock statement on query if locking is not used", () => {
         for (const dataSource of dataSources) {
             if (
--- a/test/functional/query-builder/locking/query-builder-locking.test.ts
+++ b/test/functional/query-builder/locking/query-builder-locking.test.ts
@@ -250,7 +250,7 @@ describe("query builder > locking", () => {
     it("should attach dirty read lock statement to every table of a nested join", () => {
         for (const dataSource of dataSources) {
             if (!(dataSource.driver.options.type === "mssql")) {
-                return
+                continue
             }
 
             const sql = dataSource


alumni (approved):

gioboa (approved):
Thanks for your fix @lazerg

alumni:
DO NOT MERGE (yet)

Need to check if it works for other databases. This might be an SqlServer syntax, but for other databases the place where the hint was added before this PR could have been correct.

lazerg:
I checked this across all dialects.

That call site only emits `createTableLockExpression()`, and that method returns a non-empty string for `mssql` only (`SelectQueryBuilder.ts` L2755-L2767). It has been MSSQL-only since it was added in #8507. Every other dialect builds its lock in `createLockExpression()` (L2773), which is appended once at the end of the whole query: `FOR UPDATE`, `FOR SHARE`, `LOCK IN SHARE MODE`, `FOR SHARE LOCK`, `FOR NO KEY UPDATE`, `FOR KEY SHARE`. Join position has no effect there. `useIndex` is also unrelated, because it is written into the FROM clause and not into joins, and that code is untouched.

To make sure I did not miss a path, I generated the SQL for a two-level nested join with all five lock modes, on the old code and on the new code, for 10 driver types: mssql, mysql, mariadb, postgres, cockroachdb, oracle, sap, better-sqlite3, aurora-mysql, spanner. That is 50 queries per side. The two sets are identical except for the three MSSQL lock modes.

Samples that the patch does not change:

postgres, `pessimistic_write`:

```sql
... INNER JOIN ("category" "categories" INNER JOIN "category_images_image" "categories_images" ON ... INNER JOIN "image" "images" ON ...) ON ... FOR UPDATE
```

mysql, `pessimistic_read`:

```sql
... INNER JOIN (`category` `categories` INNER JOIN `category_images_image` `categories_images` ON ... INNER JOIN `image` `images` ON ...) ON ... LOCK IN SHARE MODE
```

oracle and sap put `FOR UPDATE` and `FOR SHARE LOCK` in that same trailing position. better-sqlite3 and spanner throw `LockNotSupportedOnGivenDriverError` before any SQL is built.

MSSQL, `dirty_read`, before:

```sql
... INNER JOIN ("category" "categories" INNER JOIN "category_images_image" "categories_images" WITH (NOLOCK) ON ... INNER JOIN "image" "images" WITH (NOLOCK) ON ...) WITH (NOLOCK) ON ...
```

after:

```sql
... INNER JOIN ("category" "categories" WITH (NOLOCK) INNER JOIN "category_images_image" "categories_images" WITH (NOLOCK) ON ... INNER JOIN "image" "images" WITH (NOLOCK) ON ...) ON ...
```

The old form drops the hint from `category` and puts it after the closing bracket, which SQL Server rejects.

The non-MSSQL results above come from generated SQL only, so no server ran them. The MSSQL case ran on a real SQL Server container: the new test fails on `master` and passes with this patch.

lazerg:
Hey @alumni, I checked the lock hint change across dialects. createTableLockExpression() only returns a non-empty string for mssql, so other databases aren't affected. Does that address the DO NOT MERGE concern?

michaelbromley:

```
git fetch origin
```

lazerg:
Rebased onto the new packages layout, head is a22a38e.

The rename replayed cleanly, but #12793 landed in the same window and rewrote `buildJoinClause`, so I resolved that part by hand.

That change also made my test stop catching the bug. Inner joins no longer nest, so the old two-inner-join query is already flat and correct on master. The broken SQL now appears only when the outer join is a LEFT join, so the test parent is now `leftJoinAndSelect`.

On current master:

```sql
LEFT JOIN ("category" "categories" INNER JOIN ... ) WITH (NOLOCK) ON ...
```

With the fix:

```sql
LEFT JOIN ("category" "categories" WITH (NOLOCK) INNER JOIN ... ) ON ...
```

I ran it against a SQL Server container. The test fails on master and passes with the patch, and the suite is 3648 passing, 0 failing.

I also re-ran the cross-dialect check on the rebased code. Only the three MSSQL lock modes change, the other nine driver types generate identical SQL.

PR #12766, opened 2026-08-08
fix(expo): prepare the statement inside the try block

### LINKED ISSUE #12768: Expo driver loses the error when statement preparation fails
### Issue description

`ExpoQueryRunner.query()` calls `prepareAsync` outside the `try` block, so a failure during statement preparation skips the error handling entirely.

### Expected Behavior

A failed query reports through the same path as every other driver: `logQueryError` runs, `AfterQuery` is broadcast with `success: false`, and the caller receives a `QueryFailedError`.

### Actual Behavior

When `prepareAsync` rejects (malformed SQL, missing table, closed database), the `catch` block never runs. `logQueryError` is not called, `AfterQuery` is never broadcast even though `BeforeQuery` already was, and the raw driver error propagates unwrapped.

A subscriber that collects failed queries silently misses every preparation failure on Expo.

### Steps to reproduce

Run a query with invalid SQL on the Expo driver with an `afterQuery` subscriber attached. The subscriber is not called, and the rejection is not a `QueryFailedError`.

### My Environment

| Database driver     | expo    |

### Additional Context

`cordova`, `sqljs` and `react-native` create their statement inside the `try`. Expo is the only sqlite driver where preparation sits outside it.

- [x] expo

kyungseopk1m:
Fixes #12768.

## Problem

`ExpoQueryRunner.query()` calls `prepareAsync` before entering the `try` block:

```ts
const queryStartTime = Date.now()

const statement = await databaseConnection.prepareAsync(query)
try {
    const rawResult = await statement.executeAsync(parameters)
    ...
} catch (err) {
    this.driver.dataSource.logger.logQueryError(err, query, parameters, this)
    this.broadcaster.broadcastAfterQueryEvent(broadcasterResult, query, parameters, false, 0, undefined, err)
    await broadcasterResult.wait()
    throw new QueryFailedError(query, parameters, err)
} finally {
    await broadcasterResult.wait()
    await statement.finalizeAsync()
}
```

`prepareAsync` parses and compiles the SQL, so it fails on malformed statements, on a missing table, and on a closed database. When it throws, the `catch` block never runs. That means:

- `logQueryError` is not called, so the failure never reaches the logger
- `AfterQuery` is not broadcast, so subscribers never see it. `BeforeQuery` was already broadcast a few lines earlier, so the pair is left unmatched
- the raw driver error propagates unwrapped instead of a `QueryFailedError`, unlike every other driver

A subscriber that collects failed queries silently misses every preparation failure on Expo.

## Fix

Move `prepareAsync` inside the `try`.

The statement is undefined when preparation fails, so `finalizeAsync` is now called with optional chaining. Without that guard the `finally` block throws `Cannot read properties of undefined (reading 'finalizeAsync')`, which replaces the error we are trying to report and makes the situation worse than before. Nothing leaks in that path: SQLite returns no statement handle when `prepare` fails, so there is nothing to finalize.

`queryStartTime` is left where it is. Preparation is part of the measured window in the other drivers as well, so moving it would be an unrelated behaviour change.

`statement` is typed `any` because `AbstractSqliteQueryRunner.connect()` returns `Promise<any>` and `ExpoDriver.databaseConnection` is untyped, so a narrower annotation would resolve to `any` anyway. Wrapping the body in a nested `try` would avoid the declaration, but it would also put `finalizeAsync` failures under the outer `catch` and have them wrapped as query errors, which is a different behaviour.

## Tests

`test/unit/driver/ExpoQueryRunner.test.ts` instantiates the runner against a mocked connection, following the pattern in `test/unit/driver`:

- `prepareAsync` rejects: asserts the thrown error is a `QueryFailedError` carrying the original as `driverError`, that `logQueryError` receives the original error along with the query and parameters, and that `AfterQuery` fires once with `success: false`
- `prepareAsync` succeeds but `executeAsync` rejects: asserts `finalizeAsync` still runs, so the optional chaining does not skip cleanup on the path where a statement exists

Reverting the change fails the first test. `pnpm run test:fast` is otherwise unchanged.

## Related

Same class as #12752, which moved `getStmt` inside the `try` in `BetterSqlite3QueryRunner`. The guard here is specific to Expo, which creates a statement per query and has to finalize it, while better-sqlite3 caches statements and has no cleanup step.


kyungseopk1m:
Done in abeb06b1b.

michaelbromley:

```
git fetch origin
```

PR #12767, opened 2026-08-08
fix(react-native): wait for AfterQuery subscribers before settling

### LINKED ISSUE #12769: React Native driver rejects before AfterQuery subscribers finish
### Issue description

`ReactNativeQueryRunner.query()` waits for `AfterQuery` subscribers on the success callback but not on the error callback.

### Expected Behavior

Both callbacks wait for the subscribers before settling, as the other drivers do.

### Actual Behavior

On a failed query the runner broadcasts `AfterQuery` and calls `fail()` right away, so the caller sees the rejection before the subscribers have run. A subscriber that throws becomes an unhandled rejection, since `fail()` has already been called.

The outer `finally { await broadcasterResult.wait() }` does not cover this. `executeSql` is callback based, so the `finally` runs while the promise list is still empty and neither callback has fired.

The success path has a related hole: if a subscriber rejects, `await Promise.all(...)` rejects inside an async callback with no handler, so neither `ok()` nor `fail()` runs and the query promise never settles.

### Steps to reproduce

Attach an `afterQuery` subscriber that awaits, then run a failing query on the React Native driver. The rejection arrives before the subscriber completes. Make the subscriber throw on a successful query and the query hangs.

### My Environment

| Database driver     | react-native |

### Additional Context

`cordova` and `nativescript` share this callback shape and already await on both paths.

- [x] react-native

kyungseopk1m:
Fixes #12769.

## Problem

`ReactNativeQueryRunner.query()` treats the two `executeSql` callbacks differently.

The success callback waits for the `AfterQuery` subscribers before resolving:

```ts
if (broadcasterResult.promises.length > 0)
    await Promise.all(broadcasterResult.promises)
...
ok(result)
```

The error callback broadcasts and rejects immediately:

```ts
this.broadcaster.broadcastAfterQueryEvent(broadcasterResult, query, parameters, false, undefined, undefined, err)

fail(new QueryFailedError(query, parameters, err))
```

So on a failed query the caller sees the rejection before the subscribers have run. A subscriber that records failures may still be in flight, and one that throws produces an unhandled rejection, since `fail()` has already been called and nobody is holding its promise.

The outer `finally { await broadcasterResult.wait() }` did not help. `executeSql` is callback based, so the `try` body returns as soon as the call is registered and the `finally` runs while `broadcasterResult.promises` is still empty. Neither callback has fired at that point. That is exactly why the success path had to await a second time inside the callback.

The success path has a related hole. If a subscriber rejects, `await Promise.all(...)` rejects inside an async callback with no handler, so neither `ok()` nor `fail()` is ever called and the query promise hangs forever.

## Fix

Await the subscribers on both paths and drop the outer `finally`, which was dead.

The two paths handle a throwing subscriber differently, matching the cordova and nativescript runners:

- success: the query itself was fine, so the subscriber error is the only failure there is. Reject with it.
- error: there is already a real failure. A subscriber that throws while logging it should not replace the `QueryFailedError` the caller needs to see, so its error is swallowed.

## Tests

`test/unit/driver/query-broadcast-events.test.ts` drives the runner against a mocked `executeSql`:

- a failing query waits for a slow subscriber before rejecting
- a failing query with a rejecting subscriber still rejects with the original error, and the subscriber rejection does not escape as an `unhandledRejection`
- a successful query with a rejecting subscriber rejects instead of hanging

Each case is guarded by a timeout so a regression into the hang fails the test rather than stalling the run. Reverting the error path fails the first two, reverting the success path fails the third.

`pnpm run test:fast` is otherwise unchanged.

## Related

Same class as the cordova and nativescript work. The runners share this callback shape, and react-native was the one left with the asymmetry.

### DIFF
--- a/src/driver/react-native/ReactNativeQueryRunner.ts
+++ b/src/driver/react-native/ReactNativeQueryRunner.ts
@@ -100,8 +100,12 @@ export class ReactNativeQueryRunner extends AbstractSqliteQueryRunner {
                                 this,
                             )
 
-                        if (broadcasterResult.promises.length > 0)
-                            await Promise.all(broadcasterResult.promises)
+                        try {
+                            await broadcasterResult.wait()
+                        } catch (subscriberErr) {
+                            fail(subscriberErr)
+                            return
+                        }
 
                         const result = new QueryResult()
 
@@ -130,7 +134,7 @@ export class ReactNativeQueryRunner extends AbstractSqliteQueryRunner {
                             ok(result.raw)
                         }
                     },
-                    (err: any) => {
+                    async (err: any) => {
                         this.driver.dataSource.logger.logQueryError(
                             err,
                             query,
@@ -146,14 +150,17 @@ export class ReactNativeQueryRunner extends AbstractSqliteQueryRunner {
                             undefined,
                             err,
                         )
+                        try {
+                            await broadcasterResult.wait()
+                        } catch {
+                            // a subscriber failing must not hide the original query error
+                        }
 
                         fail(new QueryFailedError(query, parameters, err))
                     },
                 )
             } catch (err) {
                 fail(err)
-            } finally {
-                await broadcasterResult.wait()
             }
         })
     }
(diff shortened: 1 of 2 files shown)


kyungseopk1m:
Both are pre-existing: `BroadcasterResult.wait()` is shared by every driver, and the casts match the existing test stubs.

michaelbromley:

```
git fetch origin
```

PR #12771, opened 2026-08-09
fix: throw on primitive value for relation in find where conditions

### LINKED ISSUE #12712: findBy with a primitive under a relation key silently returns all rows (no WHERE clause)

zangetsu787:
## Summary

Fixes #12712 — **silent data leak**: a primitive value under a relation key in a find `where` (e.g. `findBy(Post, { author: 1 })`) was silently dropped. The relation was joined but no predicate was built, so the query returned **every row** in the table:

- `findBy(Post, { author: 1 })` → returns every post, SQL has **no WHERE clause**
- `countBy` / `existsBy` → operate on the whole table

In multi-tenant applications this silently returns other tenants' records.

## Fix

In `SelectQueryBuilder.buildWhere`, the relation branch's final `else` now throws a `TypeORMError` when the value is a primitive, pointing to the documented nested form:

```ts
// Before: silently returns ALL rows
await postRepo.findBy({ author: 1 })

// After: throws
// Invalid value encountered in property 'Post.author' of a where condition.
// To filter by the relation's primary key use the nested form { author: { id: 1 } }.

// Correct usage (documented API, unchanged):
await postRepo.findBy({ author: { id: 1 } })
```

**The documented `author: true` join shorthand is preserved** — it explicitly falls through to the existing join logic.

## Why this approach

- The primitive shorthand is **not part of the documented/typed API** (see issue #12712 — the docs only show the nested form; `FindOptionsWhere` types reject primitives at compile time). This guard protects JS callers who bypass the types.
- Consistent with the codebase's recent direction: silently-dropped criteria become loud errors (`invalidWhereValuesBehavior` work in #12578, `normalizeAndValidateWhereCriteria` in EntityManager).

## Verification

- New test `test/github-issues/12712/` covering `findBy` / `findOneBy` / `countBy` / `existsBy` with a primitive relation value (all must throw), the nested `{ id: value }` form (must still filter), and `author: true` (must still join).
- Test **fails without the fix** (query fulfilled with all rows instead of rejecting) — red/green confirmed.
- Full suite: **3046 passing, 0 failing** (better-sqlite3 + all local drivers).
- `pnpm run lint`: exit 0.

## Note on the competing PR #12730

PR #12730 attempts the same fix but throws on **any** non-object (`typeof where[key] !== "object"`), which also rejects the documented `author: true` shorthand — `typeof true === "boolean"`. I verified against this branch: that check breaks 2 existing tests in `test/functional/find-options/basic-usage/find-options-where.test.ts` (`where: { author: true }` throws `TypeORMError ... received a value of type 'boolean'`). This PR exempts `true` and keeps the full suite green.


zangetsu787:
Addressed the Qodo review findings in the latest push (5454448):

1. **Misleading PK hint (bug)** — the error message now derives the relation's primary key property name(s) from `relation.inverseEntityMetadata.primaryColumns` instead of hardcoding `id`. Single-PK entities get `{ relation: { <pk>: <value> } }`; composite-PK entities get the full list of PK properties.

2. **Functional tests over per-issue tests (rule)** — the regression test moved to `test/functional/repository/find-options-relations/repository-find-options-relations.test.ts` with an issue reference comment, per PR compliance checklist item 2.

3. **No `as any` casts (rule)** — the test now uses `@ts-expect-error` with a comment documenting the intentional invalid input, per compliance checklist item 4 (same pattern already used elsewhere in this file).

Full suite: 3046 passing, 0 failing. Lint: clean.

zangetsu787:
Second review round addressed in 5c25bb0:

1. **Excessive explanatory comment (rule)** — trimmed to a brief issue-linked note (`#12712`), per PR compliance checklist item 4.

2. **Type/runtime mismatch for boolean (bug)** — the guard now rejects only non-boolean primitives (`typeof !== "object" && typeof !== "boolean"`). Both `relation: true` and `relation: false` keep the pre-existing join-shorthand behavior, matching `FindOptionsWhereProperty`'s `boolean` type — no compile/runtime contract mismatch. Test now asserts both booleans return all rows.

Full suite: 3046 passing, 0 failing.

michaelbromley:

```
git fetch origin
```

PR #12780, opened 2026-08-12
fix: compare entity ids of mixed string/number type as equal

### LINKED ISSUE #11773: Automated cascade sets FK to NULL for bigint relations when saving parent entity (MySQL)
### Issue description

When using MySQL `bigint` as PK/FK type, TypeORM incorrectly triggers cascading updates on child entities even when `cascade` is not enabled. This results in the foreign key of the child being set to `NULL`.  Defining `{ nullable: false }` on the `ManyToOne` relation produces a database error because the FK is overwritten with `NULL`.  This behavior does not occur if PK/FK types are `int`.

### Expected Behavior

- Saving a parent entity without `cascade` should not modify child entities.
- Foreign keys of child entities must remain intact unless explicitly changed.
- `bigint` PK/FK should behave the same as `int` regarding cascading.

### Actual Behavior

- Saving a parent entity triggers an unintended `cascade` to child entities.
- The foreign key (`brandId`) of `BrandProperty` is set to `NULL`.
- Database error occurs if the relation is `nullable: false`.

### Steps to reproduce

**Entity Definitions**

``` TypeScript
@Entity()
class Brand {
    @PrimaryGeneratedColumn({ type: "bigint" })
    // @PrimaryGeneratedColumn({ type: "int" }) // Using int will not cascade update when saving the brand.
    public id: string;

    @OneToMany(
        () => BrandProperty,
        (property) => property.brand,
        // intentionally no cascade here
    )
    public properties: BrandProperty[];
}

@Entity()
class BrandProperty {
    @PrimaryGeneratedColumn({ type: "bigint" })
    // @PrimaryGeneratedColumn({ type: "int" }) // Using int will not cascade update when saving the brand.
    public id: string;

    @ManyToOne(() => Brand, (brand) => brand.properties, {
        // nullable: false, // Results in error when saving the brand, caused by the updated foreign key brandId of BrandProperty to NULL.
        // nullable: true, // DEFAULT: Results in unintended NULL foreign key brandId when saving the brand.
    })
    public brand: Brand;
}
```

**Testing insert and save**
Hint: `orm` = `EntityManager`

``` TypeScript
/**
 * After the transaction is committed, the brand and brand property are inserted into the database correctly.
 * Brand has id 1 and brand property has id 1 with foreign key brandId set to 1.
 *
 * Database after transaction is committed:
 * Brand
 * | id |
 * | 1  |
 *
 * BrandProperty
 * | id | brandId |
 * | 1  | 1       |
 */
await this.orm.transaction(async (transaction) => {
    const brand = new Brand();
    await transaction.insert(Brand, brand);

    const brandProperty = new BrandProperty();
    brandProperty.brand = brand;
    await transaction.insert(BrandProperty, brandProperty);
});

/**
 * To prove that automated cascading will NULL the foreign key of the brand property we simply read the brand and save the brand.
 * Expected result: Only brand should be affected by updated without cascading to the brand property.
 * Actual result: Brand property is also updated (cascade) with foreign key brandId set to NULL.
 *
 * Database after transaction is committed:
 * Brand
 * | id |
 * | 1  |
 *
 * BrandProperty
 * | id | brandId |
 * | 1  | NULL    |
 *
 * Note: If we define { nullable: false } to our ManyToOne relation at BrandProperty->brand,
 * the database will throw an error when trying to save the brand, caused by the updated foreign key brandId of BrandProperty.
 */
await this.orm.transaction(async (transaction) => {
    const brand = await transaction.findOne(Brand, {
        where: { id: Equal("1") },
        relations: { properties: true },
    });

    // Save will trigger cascade and setting foreign key brandId of BrandProperty to NULL.
    await transaction.save(Brand, brand);
});
```

### My Environment

| Dependency	| Version | 
| --- | --- | 
| Operating System	| Any (tested on macOS + Linux) |
| Node.js version	| 18.20.8 |
| Typescript version	| 5.8.3 |
| TypeORM version	| 0.3.27 |
| mysql	| 2.18.1 |
| Database	| MySQL 8.0.21 Community Server |

### Additional Context

Behavior occurs only with bigint PK/FK.
Changing to `int` fixes the automated cascade on `save`
Workaround: sequential save of parent and child in a transaction.

Very similar to: https://github.com/typeorm/typeorm/issues/11226
Issue persists in latest TypeORM release.

Sponsor context: https://opencollective.com/devowl

- [ ] aurora-mysql
- [ ] aurora-postgres
- [ ] better-sqlite3
- [ ] cockroachdb
- [ ] cordova
- [ ] expo
- [ ] mongodb
- [x] mysql
- [ ] nativescript
- [ ] oracle
- [ ] postgres
- [ ] react-native
- [ ] sap
- [ ] spanner
- [ ] sqlite
- [ ] sqlite-abstract
- [ ] sqljs
- [ ] sqlserver

Arul1998:
### Description of change

**What this does:** Makes `OrmUtils.compareIds` treat a single-`id` map as equal
when the two ids hold the same value but different primitive types — e.g. `"1"`
(string) and `1` (number), including scaled decimals like `"70.000"` and `70`.

**Why it's needed:** `compareIds` only short-circuited when both ids shared the
same primitive type (both string *or* both number). Drivers such as MySQL/MariaDB
return `bigint` columns as strings (`bigNumberStrings` defaults to true), so a
persisted id can be `"1"` on the DB-loaded side and `1` in memory. That mismatched
pair matched no branch and fell through to `false` (`deepCompare` → `compare2Objects`
also only string-normalizes same-type primitives). As a result
`OneToManySubjectBuilder` saw the child as orphaned and, under the default
`orphanedRowAction: "nullify"`, nulled its FK on save — erroring when the FK is
`NOT NULL`. This is why `int` PKs are unaffected (both sides are numbers) but
`bigint` PKs break (#11773).

**The change:** In the single-`id` fast path, when the two id types differ, compare
by **numeric value** rather than string form, so `"1"` matches `1` and a scaled
decimal `"70.000"` matches `70`. Guards keep it safe:

- `NaN` / `Infinity` never match.
- Integers beyond `Number.MAX_SAFE_INTEGER` never match, so a rounded JS number
  can't collide with a different `bigint` string.
- The string side must be a plain decimal numeral, so permissive `Number()` forms
  (`""` → 0, `"0x10"` → 16, `"1e2"` → 100) can't coerce distinct keys together.

Same-type comparisons are unchanged (still `firstId.id === secondId.id`), and the
general-purpose `compare2Objects` is left untouched to avoid affecting jsonb/object
diffing elsewhere.

**How it was verified:**

- Unit tests (`test/unit/util/orm-utils.test.ts`) for `compareIds`: same-type
  equality, string-vs-number equality (`"1"` === `1`), decimals incl. scale
  (`"70.000"` === `70`), unequal string/number staying `false`, `null`/`undefined`
  handling, `NaN`/`Infinity` rejection, unsafe-integer rejection, permissive
  numeric-string rejection, and the composite-id `deepCompare` fallback — all
  passing.
- A functional regression test (`test/github-issues/11773`, mysql/mariadb) with an
  `int` parent PK and a `bigint` child PK. On re-save the parent always matches, so
  the test can only fail the way #11773 actually failed — the child's `bigint` id is
  `1` in memory and `"1"` from the driver — and asserts the child's FK survives.

**Notes for maintainers:**

- Composite / non-`id` primary keys still route through `deepCompare` and are out of
  scope for this narrow, single-column fix.
- The numeric compare deliberately does not attempt full decimal canonicalization
  (e.g. `"0.10000000000000001"` vs `0.1`); that is IEEE-754 `Number()` behavior,
  unrealistic for a real `bigint`/`int`/fixed-scale `decimal` primary key, and out of
  scope here. Happy to take a driver-level normalization direction instead if you'd
  prefer solving the whole coercion class.

### Pull-Request Checklist

      `Fixes #11773`
- [x] There are new or updated tests validating the change
      (`test/unit/util/orm-utils.test.ts`, `test/github-issues/11773`)
- [ ] Documentation has been updated to reflect this change (`docs/docs/**.md`) — N/A (internal comparison fix, no public API/doc change)

### DIFF
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -22,6 +22,83 @@ Please include as much of the information listed below as you can to help us bet
 
 This information will help us triage your report more quickly.
 
+## Security Model and Scope
+
+TypeORM is an ORM _and_ a SQL-building toolkit. Understanding which of its inputs are
+trusted is essential to knowing what constitutes a vulnerability. Please read this section
+before filing an advisory — reports outside this scope will be closed with a reference to
+this policy.
+
+### What TypeORM guarantees
+
+**Values are safe.** Any untrusted _data_ supplied in a value position — entity values,
+criteria values, named parameters, pagination values — is bound as a query parameter or
+correctly escaped as a literal.
+
+**Database content is treated as data.** Whatever comes back from the database — values,
+result sets, database-reported metadata — is handled as untrusted data wherever TypeORM
+processes it: during entity hydration, and in anything TypeORM generates from database
+state, including code and files emitted by the CLI.
+
+Bugs in these guarantees are vulnerabilities and we want to hear about them. In scope:
+
+- Untrusted runtime data reaching a generated query unescaped — or, for non-SQL drivers,
+  being interpreted as query operators rather than values — through a documented API used
+  as documented. This includes TypeORM's _internal_ query generation interpolating entity
+  data.
+- Parameter-binding or literal-escaping defects in a driver.
+- Defects in identifier escaping where TypeORM applies it.
+- Database-derived content escaping its context in TypeORM's output: prototype pollution
+  or code execution while processing results, or injection into code, files, or commands
+  generated from database state.
+- Amplification of an existing prototype-pollution primitive: object iteration in
+  query-building, result-processing, or code-generation paths must not pick up inherited
+  keys, so that a polluted runtime cannot alter generated queries or emitted code through
+  TypeORM. (The pollution source itself remains a vulnerability in whatever introduced
+  it; such reports are assessed at reduced severity.)
+
+### What TypeORM does not guarantee
+
+**Identifiers and SQL expressions are code, not data.** QueryBuilder builds SQL. By
+design, any argument documented to accept an identifier or a SQL fragment — projections,
+condition strings, sort and grouping expressions, aliases, computed expressions — is
+emitted as the developer wrote it, because expressing arbitrary SQL is the point of those
+positions. They **cannot be escaped without breaking documented functionality**, and the
+API does not promise they are injection-safe.
+
+Passing untrusted input into such a position is an application vulnerability, equivalent
+to concatenating user input into a raw query. Reports that require the application to do
+so are out of scope. Use parameter binding for all untrusted data; never build identifiers
+or expressions from user input.
+
+**Schema definitions are trusted code.** Entity definitions, migration objects, and their
+options are design-time inputs authored by the developer. DDL generated from them is not
+hardened against a hostile schema: an attacker who controls your schema definitions
+already controls your application. Reports premised on untrusted input reaching schema or
+migration definitions are out of scope. (We still accept — and appreciate — pull requests
+that escape identifier-only positions as defense-in-depth; those are handled as regular
+hardening changes, not advisories.)
+
+**Configuration is trusted code.** Connection options, data source configuration, CLI
+arguments, and migration files are developer-authored inputs, on par with application
+source. Reports requiring an attacker to control them are out of scope.
+
+**Runtime guards are best-effort, not security boundaries.** TypeORM sometimes adds
+validations that catch common application mistakes before they reach the database. These
+guards are conveniences layered on top of the model above; they do not move the trust
+boundary, and they are not, and cannot be, complete. A bypass of such a guard that
+restores previously documented behavior is a regular bug — please file it as an ordinary
+issue — not a vulnerability.
+
+### Severity expectations
+
+CVSS scores should reflect TypeORM's contract, not a hypothetical application that pipes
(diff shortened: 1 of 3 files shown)


michaelbromley:

```
git fetch origin
```

Arul1998:
Thanks for the heads-up and the detailed migration notes, @michaelbromley no worries at all about the disruption.

I've rebased fix/compare-ids-bigint-string onto the new packages/typeorm layout using the recommended raised-rename-limit rebase. The two modified files replayed automatically; the new issue-11773 test files (which rename detection can't map since they're additions) I moved into packages/typeorm/test/github-issues/11773/ to match the new structure. Branch is pushed and the diff is now entirely under packages/typeorm/.

Verified locally: the package compiles cleanly and the OrmUtils unit suite passes (including the compareIds cases this PR adds). Ready for another look whenever you have a chance.

PR #12568, opened 2026-06-08
fix: expand VirtualColumn expressions in query builder WHERE conditions

mazbauddinsaif:
## Description

Closes #11616.

`@VirtualColumn` properties are computed columns with no physical column in the database. When one was referenced in a raw query builder condition — e.g.:

```ts
qb.where("company.totalEmployeesCount > 2")
```

`replacePropertyNamesForTheWholeQuery` emitted a reference to a non-existent column (`company.totalEmployeesCount`), producing SQL that failed with errors like `Unknown column 'company.totalEmployeesCount' in 'where clause'`.

This already worked through the find-options path (`where: { totalEmployeesCount: MoreThan(2) }`), but **not** through raw string conditions on the query builder.

## Fix

`replacePropertyNamesForTheWholeQuery` now detects virtual property columns and inlines their computed SQL expression — the exact same expression already produced when the column is `SELECT`ed (`(${column.query(escapedAlias)})`) — instead of emitting a non-existent column name. Regular columns are unchanged.

## Tests

Added a test under `column > virtual columns` asserting that a query builder `where()` clause on a virtual column expands to its subquery expression.

- New test passes.
- Full sqlite test suite: **2945 passing, 0 failing**.
- `prettier` + `eslint` clean on changed files (0 errors).

## Notes

The old inline comment claiming where/order-by on virtual columns is unsupported is now obsolete for the WHERE case.

PR #12429, opened 2026-05-05
fix: expand virtual columns in where clauses

paulkagiri:
### Description of change

Fixes #11616.

String WHERE clauses that reference a `@VirtualColumn` currently go through final property-name replacement as if the virtual column were a physical database column, producing SQL such as `"user"."fullName"`. This change preserves the existing physical-column replacement behavior while expanding virtual-column property references to the configured `query(alias)` expression.

The added coverage verifies both the main alias and a joined alias, matching the reported use case.

### Pull-Request Checklist

- [x] Code is up-to-date with the `master` branch
- [ ] `npm run lint` passes with this change
- [ ] `npm test` passes with this change
- [x] This pull request links relevant issues as `Fixes #11616`

### Validation

- `pnpm exec tsc --noEmit --pretty false`
- `pnpm run compile`
- `pnpm exec mocha --no-config --file build/compiled/test/utils/test-setup.js build/compiled/test/functional/columns/virtual-columns/virtual-columns.test.js --timeout 90000 --exit`
- `pnpm exec prettier --check src/query-builder/QueryBuilder.ts test/functional/columns/virtual-columns/virtual-columns.test.ts`
- Commit hook ran `eslint --fix` and `prettier --write` on staged TypeScript files
- `pnpm exec eslint src/query-builder/QueryBuilder.ts test/functional/columns/virtual-columns/virtual-columns.test.ts` completed with existing warnings only, no errors

PR #12781, opened 2026-08-12
fix: expand VirtualColumn sub-queries outside of SELECT (e.g. in WHERE)

### LINKED ISSUE #11616: VirtualColumn throws error when used in WHERE conditions

vjymisal0:
### Description of change

`VirtualColumn` sub-queries were only ever expanded in the `SELECT` clause. Anywhere else a virtual column's property name appeared (e.g. a `WHERE` condition built with `.where()`/`.andWhere()`), `QueryBuilder#replacePropertyNamesForTheWholeQuery` substituted the property name with `alias.<databaseName>` — a column that doesn't actually exist on the table — causing a driver-level error such as `Unknown column 'user.fullName' in 'where clause'`.

Root cause: `buildColumnSelectionExpression` already expands `column.query(alias)` into `(subquery)` for the `SELECT` list before the whole-query replacement pass runs, but that pass itself has no knowledge of virtual columns, so it always falls back to the plain `alias.column` substitution everywhere else.

Fix: track each virtual column's `query` function per alias prefix (alongside the existing `replacements` map) in `replacePropertyNamesForTheWholeQuery`, and when a property reference resolves to a virtual column, substitute it with its wrapped sub-query expression instead of `alias.column`. This is a targeted change to the same regex-based replacement pass that already runs across the whole generated SQL string, so it also covers `ORDER BY`/`GROUP BY`/`HAVING` wherever the same alias/property pattern is used — not just `WHERE`.

Verified by:
- Adding a regression test reproducing the issue's repro (filtering by a virtual column in `WHERE` via `createQueryBuilder`), confirmed it fails with `no such column: company.totalEmployeesCount` against the current code, and passes with this fix.
- Un-commenting the pre-existing `.andWhere("company.totalEmployeesCount > 2")` line in `virtual-columns.test.ts`, which was disabled with a comment stating "we won't be supporting where ... with VirtualColumns".
- Running the full local test suite (better-sqlite3 driver) and `tsc --noEmit`: all passing, no regressions.

Follow-up (left out of scope for this PR): the same test file also has an `.orderBy(...)` usage of a virtual column still commented out. `ORDER BY` across relation joins involves additional alias-resolution that I didn't want to fold into this fix without separate discussion — happy to pick that up in a follow-up if maintainers want it addressed too.

### Pull-Request Checklist

- [ ] Documentation has been updated to reflect this change (`docs/docs/**.md`) — N/A, this fixes existing documented `VirtualColumn` behavior rather than adding new public API surface.

Closes #11616


vjymisal0:
Closing this — I missed that #12429 and #12568 already address this exact issue with a very similar approach before opening this PR. Apologies for the duplicate noise; deferring to the earlier PRs. Maintainers may still find the additional regression test coverage here (filtering via `createQueryBuilder` directly, and the re-enabled `.andWhere` assertion in the existing test file) useful as a reference if it helps consolidate the other two.

PR #12782, opened 2026-08-14
feat(migrations): store executedAt and checksum metadata

### LINKED ISSUE #3375: Add metadata to migrations table

afehid:
## Summary

Implements a non-breaking first slice of #3375:

- Adds nullable `executedAt` and `checksum` columns to the migrations table
- Auto-adds those columns when an older migrations table already exists
- Stores SHA-1 of the loaded migration source (`name` + `up` + `down`) on insert
- Opt-in verification via `migrationsChecksumCheck: true` (default `false`)

Rows with a `NULL` checksum (created before this feature) are skipped during verification.

Closes #3375

## Test plan

- [x] Local: `mocha` for `test/github-issues/3375` on better-sqlite3 (4/4 passed)
- [ ] CI: verify across enabled drivers


docs/docs/data-source/2-data-source-options.md
```diff
@@ -69,6 +69,10 @@ Different RDBMS-es have their own specific options.
 
 - `migrationsTransactionMode` - Controls transaction mode when running [migrations](../migrations/01-why.md).
 
+- `migrationsChecksumCheck` - When `true`, compares each executed migration's stored checksum against the current migration source before running migrations. Throws if a stored checksum no longer matches. Checksums use each migration's `up`/`down` methods; keep the same artifact form (e.g. always compiled `.js`). Default: `false`.
```
smith-xyz: I think I'm a little hesistant to consider hashing js/ts - tools that do this kind of integrity check (liquibase, flyway) hashes the sql file instead

I'm not sure the runtime constructs are the safest.

Any reason why this path isn't considered?

src/migration/MigrationSource.ts
```diff
@@ -0,0 +1,40 @@
...
+export function computeMigrationChecksum(
+    name: string,
+    instance?: {
+        up?: (...args: any[]) => any
+        down?: (...args: any[]) => any
+    },
+): string {
+    const up = normalizeFunctionSource(instance?.up?.toString() ?? "")
+    const down = normalizeFunctionSource(instance?.down?.toString() ?? "")
+    return PlatformTools.sha1(`${name}\0${up}\0${down}`)
```
smith-xyz: Here since we're generating SQL at runtime, you have strings in memory. 

I do propose maybe we consider a file based approach though, then we can normalize machine to machine as `crypto.createHash("sha256").update(sql, "utf8")`, which handles encoding. But you'd still need to normalize line endings in the generated SQL (some drivers may emit \r\n on Windows).

src/migration/MigrationSource.ts
```diff
@@ -0,0 +1,40 @@
...
+ * @param instance.down
+ */
+export function computeMigrationChecksum(
+    name: string,
+    instance?: {
+        up?: (...args: any[]) => any
+        down?: (...args: any[]) => any
+    },
+): string {
+    const up = normalizeFunctionSource(instance?.up?.toString() ?? "")
```
smith-xyz: Different minifiers/bundlers could make create variation here

src/error/MigrationChecksumMismatchError.ts
```diff
@@ -0,0 +1,24 @@
...
+    readonly migrationName: string
+    readonly storedChecksum: string
+    readonly currentChecksum: string
+
+    constructor(
+        migrationName: string,
+        storedChecksum: string,
+        currentChecksum: string,
+    ) {
+        super(
```
smith-xyz, line 19: given areas where MigrationChecksumMismatchError are thrown and that we could have variation on Function.prototype.toString() - I think the error maybe assumes the mismatch is real.

Maybe a good debugging option is to get that up/down that was used in comparison to spit out in logging

smith-xyz (changes requested):
I really like your clean approach you took to this, let's work through some of this more to get the implementation where it needs to be. 

Some fundamental concerns:
1. hashing ts/js - better alternative may be a sql artifact
2. some DDL round-trip overhead on every startup: `ensureMigrationsTableColumns` makes N+2 sequential `hasColumn` + `addColumn` queries.
3. We may be trying to make this work for all the drivers at once - should some of this be allocated to the driver layer to handle the diffs between drivers (I think I saw several if blocks showing there was driver specific behavior being accounted for).
4. checksum column is varchar(40) ... which locks in to sha1. I would just use varchar(64) and would just add usage for sha256 here, no reason to settle for a broken hash even though this isn't a security issue.
5. tests seem to only cover better-sqlite3

docs/docs/data-source/2-data-source-options.md
afehid: You're right: hashing JS/TS (or `Function.toString()`) was the wrong integrity target. TypeORM migrations aren't SQL files like Flyway/Liquibase, but they *do* generate SQL at runtime through QueryRunner.

This push switches checksums to a SHA-256 of that generated SQL (`crypto.createHash("sha256").update(..., "utf8")`), with `\r\n` normalized to `\n`. On `migration:run` we record the SQL `up()` actually sends; on checksum check we dry-run `up()` (queries not executed) and compare.

src/migration/MigrationSource.ts
afehid: Agreed, that's the approach in the latest commit. We capture the SQL strings `up()` sends (so we have them in memory at runtime), normalize line endings, then hash with SHA-256 over UTF-8.

Verification uses the same capture path without executing, so we don't re-apply an already-run migration.

src/migration/MigrationSource.ts
afehid: Yes, that's exactly why `Function.toString()` was a bad checksum input (dev `.ts` vs compiled `.js`, minifiers, bundlers). Dropped that path; checksums are over generated SQL now.

src/error/MigrationChecksumMismatchError.ts
afehid: Updated. The error no longer assumes the source was tampered with, and it now includes the SQL used to compute the current checksum (also logged via `logMigration`) so you can tell a real change from driver/environment differences.

afehid:
@smith-xyz thanks for the detailed review — all five points should now be addressed:

1. **Hash a SQL artifact instead of ts/js** — done. Checksums are now a SHA-256 of the migration name plus the SQL generated by `up()` (captured through the query runner, no `Function.toString()`), with line endings normalized to `\n`. Raw `queryRunner.query()` calls are included too (with their bound parameters as part of the hash input, though parameter values are never printed in logs or error messages). On mismatch, the error and migration log include the SQL used for the current checksum, as you suggested.
2. **DDL round-trip overhead** — `ensureMigrationsTableColumns` now reads the table once with `getTable()` and adds all missing columns in a single `addColumns()` call, cached per executor. That's one schema read plus at most one write, instead of N+2 sequential `hasColumn`/`addColumn` queries.
3. **Driver-specific branches** — the remaining branches are the pre-existing `mongodb` short-circuits and the `MssqlParameter` handling that `insertExecutedMigration`/`deleteExecutedMigration` already used, so I stayed with the executor's existing patterns rather than introducing a new driver-layer API in this PR. Happy to explore moving that into the driver layer as a follow-up if you'd prefer.
4. **varchar(40)/sha1** — the checksum column is now `varchar(64)` and hashing is SHA-256 (with a browser-safe implementation for non-node bundles).
5. **Test coverage** — the checksum suite now runs against better-sqlite3, postgres, mysql, mariadb and mssql.

afehid:
Hi maintainers — could someone **approve and run** the GitHub Actions workflows for this fork PR when you have a moment?

From the Checks tab, **Tests**, **CodeQL**, and **PR preview release** are stuck at **workflow approval required** (`action_required`, no jobs ran). Only the **semantic** check (PR title / commit format) has completed so far.

Recent updates on my side:
- Addressed the Qodo **Compliance violation** (docs + `migrationsExtraColumns` always nullable, MSSQL/column-name fixes) in `8d98072`.
- Fixed fake migrations / `insertMigration()` so they no longer run `up()` while resolving checksums (`ed7ffe83`).

Happy to adjust anything once CI can run on the fork. Thanks!

michaelbromley:

```
git fetch origin
```

afehid:
> This repository was recently restructured into a pnpm workspace, and the `typeorm` package moved from the repo root into `packages/typeorm` (PR #12854). Everything moved as a pure rename with no content changes, so your branch should replay across the move cleanly once git's rename detection limit is raised:
> 
> ```
> git fetch origin
> git -c merge.renameLimit=999999 -c diff.renameLimit=999999 rebase origin/master
> ```
> 
> If your branch has diverged enough that a full rebase isn't worth it, `scripts/migrate-pr-to-packages-layout.sh --replay` (run on top of `master`) rewrites the paths of your net diff instead. Both approaches, plus a path-mapping table and a prompt you can hand to a coding agent, are documented in `scripts/migrate-pr-to-packages-layout.md`.
> 
> Sorry for the disruption, and thanks for your patience.

Thanks — updated this branch onto the packages/typeorm layout in c08417b (merged current master). All of the PR’s paths now live under packages/typeorm/ / docs/. Local typecheck + migration tests look good; when you have a moment, could you approve the fork CI runs for this commit?

PR #12783, opened 2026-08-15
fix: use alter column type instead of drop and add when only column length changes

pyfile-toolkit:
### Description

When a column's length changes (e.g. `varchar(50)` -> `varchar(51)`) while the base type stays the same, `PostgresQueryRunner.changeColumn` used to drop and re-create the column, which causes data loss.

### Fix

1. Move the `length` comparison out of the drop-and-recreate condition and into the existing precision/scale branch, which already generates `ALTER COLUMN ... TYPE`.
2. Update the cloned column's `length`/`precision`/`scale` in that branch so `replaceCachedTable` does not keep stale schema state (spotted by the Qodo review).

### Checklist

- [x] Regression test in the functional suite (`test/functional/schema-builder/column/change/change-column/change-column.test.ts`)
- [x] Existing column-length / column-type / change-column / migration-generate tests still pass

Closes #3357


pyfile-toolkit:
Thanks for the review. I've gone through the Qodo findings and addressed the relevant ones:

- **Test metadata not restored** ✅ — the regression test now restores `nameColumn.length` to the baseline `"255"` (matching the other tests in the file), so shared `ColumnMetadata` state stays consistent and later tests are not order-dependent. Fix is in commit `2b21c4557`.

CI is green and the PR is ready for maintainer review. Happy to adjust further if anything needs tweaking.

pyfile-toolkit:
Closing per maintainer policy (comment at https://github.com/typeorm/typeorm/issues/3357#issuecomment-4638034647: no community PRs will be accepted for this issue). Thanks for the Qodo review earlier — the test-metadata fix remains in the branch if ever useful.

PR #12508, opened 2026-05-17
fix: evaluate function defaults before enum/string normalization

Cprakhar:
Closes #12492
### Description of change
This change updates normalizeDefault for (Postgres, CockroachDB, and MySQL) so that function-based defaults are handled earlier in the method flow. 

__What changed__
In `normalizeDefault` method: moved `typeof defaultValue === "function"` branch above enum/string/number normalization branches

### Pull-Request Checklist

- [x] Code is up-to-date with the `master` branch
- [x] This pull request links relevant issues as `Fixes #12492`
- [x] There are new or updated tests validating the change (`tests/**.test.ts`)
- [ ] Documentation has been updated to reflect this change (`docs/docs/**.md`) (N/A)

PR #12786, opened 2026-08-17
fix: evaluate function-typed column defaults on enum columns instead of stringifying them

### LINKED ISSUE #12492: Postgres enum array column default callbacks are stringified instead of evaluated
### Issue description

For enum columns the postgres driver generates invalid SQL by stringifying the javascript function

### Expected Behavior

I would expect the callback provided as default to the decorator to be called

### Actual Behavior

The callback is stringified and the resulting SQL is invalid. Behavior can be traced back to this block:

https://github.com/typeorm/typeorm/blob/62948a3751b0d978ab854bf2f51c3453ff8b15b2/src/driver/postgres/PostgresDriver.ts#L1222

### Steps to reproduce

Providing example code for both single-enum and enum-array columns, since I initially encountered the issue with the array variant, but the same behavior also occurs with non-array enums:
```typescript

const EnumDef= ["VALUE1", "VALUE2"] as const;
type EnumDef = typeof EnumDef[number]

@Entity()
export class SingleEnum {
  @Column({
    type: "enum",
    enum: EnumDef,
    enumName: "enum_def",
    nullable: false,
    array: false,
    default: () => "'VALUE2'::enum_def"
  })
  labels!: EnumDef;
}

@Entity()
export class ArrayEnum {
  @Column({
    type: "enum",
    enum: EnumDef,
    enumName: "enum_def",
    nullable: false,
    array: true,
    default: () => "ARRAY[]::enum_def[]"
// same issue with '{}'::enum_def[] syntax
  })
  labels!: EnumDef[];
}
```

The generated migration contains the following (excerpt):
```
await queryRunner.query(`CREATE TYPE "public"."enum_def" AS ENUM('VALUE1', 'VALUE2')`);
        await queryRunner.query(`CREATE TABLE "single_enum" ("id" SERIAL NOT NULL, "labels" "public"."enum_def" NOT NULL DEFAULT '() => "'VALUE2'::enum_def"', CONSTRAINT "PK_4b9276628ac8c63c1c075a8d403" PRIMARY KEY ("id"))`);
        await queryRunner.query(`CREATE TYPE "public"."enum_def" AS ENUM('VALUE1', 'VALUE2')`);
        await queryRunner.query(`CREATE TABLE "array_enum" ("id" SERIAL NOT NULL, "labels" "public"."enum_def" array NOT NULL DEFAULT '() => "ARRAY[]::enum_def[]"', CONSTRAINT "PK_b6de0bdf7f1af4a43ca872244a7" PRIMARY KEY ("id"))`);
```

Apart from the duplicated enum creation (which would already fail independently — see/reopen https://github.com/typeorm/typeorm/issues/8136
 and related migration/default issues), the generated SQL clearly shows that the default callback function is being stringified instead of evaluated, resulting in invalid PostgreSQL syntax.

### My Environment

| Operating System    |Windows 11 25H2|
| Node.js version     | 22.15.0  |
| Typescript version  | 5.7.2  |

### Additional Context

_No response_

- [x] postgres
- [x] cockroachdb
- [x] mysql
- [ ] aurora-mysql
- [ ] aurora-postgres
- [ ] better-sqlite3
- [ ] capacitor
- [ ] cordova
- [ ] expo
- [ ] mongodb
- [ ] nativescript
- [ ] oracle
- [ ] react-native
- [ ] sap
- [ ] spanner
- [ ] sqlite
- [ ] sqljs
- [ ] sqlserver

juneja-varun:
## Problem

`normalizeDefault()` in the postgres, cockroachdb, and mysql drivers all matched enum/simple-enum columns by checking `columnMetadata.type` alone, before checking whether the default value was actually a function:

```ts
if (
    (columnMetadata.type === "enum" ||
        columnMetadata.type === "simple-enum" ||
        typeof defaultValue === "number" ||
        typeof defaultValue === "string") &&
    defaultValue !== undefined
) {
    return `'${defaultValue}'`
}
```

A function-typed default on an enum column (e.g. `default: () => "'FOO'::my_enum"`) never reaches the `typeof defaultValue === "function"` branch further down — it falls into the enum branch first and gets template-literal stringified (implicitly calling the function's own `.toString()`), producing invalid SQL like:

```sql
DEFAULT '() => "FOO::my_enum"'
```

matching exactly what's reported in #12492.

## Fix

Move the function check above the enum/number/string check in all three drivers, so a function-typed default gets evaluated regardless of column type — consistent with how it already works for every other column type. `normalizeDatetimeFunction()` (reused unchanged) is a no-op pass-through for any value that isn't a recognized datetime keyword, so this doesn't change behavior for the datetime-default case, only extends the existing evaluate-the-function path to enum columns too.

Scope: the linked issue's environment checkboxes marked postgres, cockroachdb, and mysql as affected — checked all three and found the identical structural bug in each (mysql and cockroachdb weren't mentioned in the issue body's reproduction, but have the same `columnMetadata.type === "enum"`-first ordering).

## Testing

Added `test/unit/driver/normalize-default-enum-function.test.ts`, using the `Object.create(Driver.prototype)` pattern already established in `test/unit/driver/escape-identifier.test.ts` (no live DB connection needed — `normalizeDefault()` is a pure function of `columnMetadata`). Covers all three drivers: function-typed enum default now evaluates correctly, and a plain (non-function) enum default still stringifies as before. Confirmed the new tests fail with the exact reported symptom (stringified function source) on unpatched code and pass with the fix.

Full unit suite: 141 passed. `eslint` and `prettier --check` clean on all changed files (confirmed identical pre-existing warning count before/after — no new warnings introduced).

Fixes #12492


juneja-varun:
Noticed #12508 fixes the same underlying issue (opened back in May, already has two approvals) and additionally covers aurora-mysql, which I've now added here too along with addressing Qodo's review feedback (fragile test driver stub, double type-assertion). Given #12508 is clearly further along, happy to close this one in favor of it if a maintainer prefers — just let me know.

alumni:
hey @juneja-varun, thanks for opening this PR!

We merged #12508 which was identical (except the tests) and already had the approvals since quite some time ago. We should've fixed this earlier, somehow the PR got lost in the list with all this AI-generated spam 🤷🏻‍♂️

juneja-varun:
No worries at all, glad #12508 already covers it properly. Appreciate you taking the time to reply.

PR #12790, opened 2026-08-18
fix(postgres): preserve data when altering varchar column length

xkong2017:
## Summary
A pure column length change (e.g. `varchar(50)` → `varchar(51)`) previously triggered column recreation (`DROP COLUMN` + `ADD COLUMN`), which **discards existing data**. This routes length-only changes through `ALTER COLUMN ... TYPE` instead, mirroring the existing precision/scale handling in `PostgresQueryRunner.changeColumn`.

## Root cause
In `PostgresQueryRunner.changeColumn`, the DROP+ADD branch was entered whenever `oldColumn.length !== newColumn.length`. A length change is a safe, in-place type alteration and does not require recreating the column.

## Changes
- `src/driver/postgres/PostgresQueryRunner.ts`: removed `length` from the DROP+ADD condition and added a length-only branch that emits `ALTER TABLE ... ALTER COLUMN ... TYPE` (with matching down query).
- Added regression test `test/github-issues/3357/` verifying data is preserved when a varchar length is increased.

## Test plan
- [x] `tsc --noEmit` passes
- [ ] Postgres integration test (`test/github-issues/3357`) — requires a live Postgres instance (not available in this environment); will run in CI

Fixes #3357


xkong2017:
Addressed both code review findings:

**1. Test location (Maintainability)** — Moved the regression test from `test/github-issues/3357` to `test/functional/database-schema/column-length-change/` per the project preference for functional tests. The test now lives in the functional suite with its entity under `entity/User.ts`.

**2. Stale cached schema (Reliability)** — The length-only `ALTER COLUMN ... TYPE` branch now updates the corresponding column's `length` on `clonedTable` before `replaceCachedTable` is called, so the QueryRunner's cached table metadata reflects the new length and subsequent schema operations in the same QueryRunner won't mis-detect the column and emit redundant follow-up DDL.

Validation: `tsc --noEmit` passes with 0 errors; `eslint` passes with 0 errors on the changed files.

xkong2017:
Extended the fix to **all 8 database drivers**, not just postgres.

The original PR only patched `PostgresQueryRunner`. Since the maintainers have repeatedly declined the ~13 competing PRs for #3357 (each covering only one driver), I've now applied the same data-preserving length-change logic to every driver:

- mysql (`MODIFY COLUMN`)
- aurora-mysql (`MODIFY COLUMN`)
- oracle (`ALTER TABLE ... MODIFY`)
- cockroachdb (`ALTER COLUMN ... TYPE`)
- sap (`ALTER ...`)
- spanner (`ALTER COLUMN ... TYPE`)
- sqlserver (`ALTER COLUMN ...`)

Each branch is guarded to fire **only** on a pure length change (`newColumn.length !== oldColumn.length && newColumn.type === oldColumn.type`), so it never interferes with type/precision/scale changes that legitimately require DROP + ADD. The cached `clonedTable.columns` length is also updated to keep the in-memory schema consistent.

Validation:
- `pnpm exec tsc --noEmit` → 0 errors
- eslint on all changed driver files → 0 errors
- New functional test `test/functional/database-schema/column-length-change` (postgres) verifies data is preserved when a varchar length is increased

This addresses the cross-driver requirement that blocked the earlier attempts. Please review.

xkong2017:
Hi @naorpeled @michaelbromley — I'd appreciate a review of this PR when you have a moment.

I'm aware there are ~13 open PRs for #3357, and I understand why that makes it hard to prioritize. This one is different in a way that directly addresses the reason the earlier attempts stalled:

**This is the only PR that fixes the bug across all 8 database drivers**, not just postgres. Every other open PR for #3357 patches only `PostgresQueryRunner`. Since the maintainers have repeatedly declined single-driver fixes, I extended the same data-preserving length-change logic to mysql, aurora-mysql, oracle, cockroachdb, sap, spanner, and sqlserver — each using the correct driver-specific ALTER syntax.

What's included:
- Pure length changes (e.g. `varchar(50)` → `varchar(51)`) now emit an in-place `ALTER ... TYPE`/`MODIFY` instead of DROP + ADD, preserving existing rows.
- Each branch is guarded to fire **only** on a pure length change (`newColumn.length !== oldColumn.length && newColumn.type === oldColumn.type`), so type/precision/scale changes that legitimately require DROP + ADD are unaffected.
- The cached `clonedTable.columns` length is updated to keep the in-memory schema consistent.
- A functional regression test under `test/functional/database-schema/column-length-change` verifies data is preserved.

Validation: `tsc --noEmit` → 0 errors; eslint on all changed files → 0 errors; functional test passes.

Happy to adjust anything to match the project's preferred approach. Thanks for your time.

michaelbromley:

```
git fetch origin
```

PR #12791, opened 2026-08-18
test: cover embedded relation loading under the query strategy with order or select

### LINKED ISSUE #12788: relationLoadStrategy: 'query' throws TypeError when an embedded relation is combined with order or select
### Issue description

With `relationLoadStrategy: 'query'`, loading a relation that lives inside an embedded entity throws a `TypeError` as soon as the same `find()` also passes `order` or `select`. The identical query works under the default `join` strategy.

### Expected Behavior

`relationLoadStrategy` is a performance choice, so `join` and `query` should return the same result for the same find options. The embedded relation should load, or the combination should be rejected with a descriptive error.

### Actual Behavior

```
TypeError: Cannot read properties of undefined (reading 'tags')
    at Function.deepValue (src/util/OrmUtils.ts:151:22)
    at src/query-builder/SelectQueryBuilder.ts:3680:44
    at SelectQueryBuilder.executeEntitiesAndRawResults (src/query-builder/SelectQueryBuilder.ts:3653:40)
    at SelectQueryBuilder.getRawAndEntities (src/query-builder/SelectQueryBuilder.ts:1671:29)
    at SelectQueryBuilder.getMany (src/query-builder/SelectQueryBuilder.ts:1761:25)
```

### Steps to reproduce

```ts
import 'reflect-metadata';
import { Column, DataSource, Entity, JoinTable, ManyToMany, PrimaryGeneratedColumn } from 'typeorm';

@Entity()
class Tag {
    @PrimaryGeneratedColumn()
    id: number;

    @Column('varchar')
    name: string;
}

class Meta {
    @Column('varchar', { nullable: true })
    note: string;

    @ManyToMany(() => Tag)
    @JoinTable()
    tags: Tag[];
}

@Entity()
class Post {
    @PrimaryGeneratedColumn()
    id: number;

    @Column('varchar')
    title: string;

    @Column(() => Meta)
    meta: Meta;
}

async function main() {
    const ds = new DataSource({
        type: 'better-sqlite3',
        database: ':memory:',
        synchronize: true,
        entities: [Tag, Post],
    });
    await ds.initialize();

    const tag = await ds.getRepository(Tag).save({ name: 'a' });
    await ds.getRepository(Post).save({ title: 'p', meta: { note: 'n', tags: [tag] } });

    const repo = ds.getRepository(Post);
    const cases: Array<[string, any]> = [
        ['join  + order ', { relations: ['meta.tags'], order: { id: 'ASC' }, relationLoadStrategy: 'join' }],
        ['query + order ', { relations: ['meta.tags'], order: { id: 'ASC' }, relationLoadStrategy: 'query' }],
        ['query + select', { relations: ['meta.tags'], select: { id: true }, relationLoadStrategy: 'query' }],
        ['query (neither)', { relations: ['meta.tags'], relationLoadStrategy: 'query' }],
    ];

    for (const [label, options] of cases) {
        try {
            const rows = await repo.find(options);
            console.log(`${label} -> OK    rows=${rows.length} tags=${rows[0]?.meta?.tags?.length}`);
        } catch (e: any) {
            console.log(`${label} -> THROW ${e.constructor.name}: ${e.message}`);
        }
    }
    await ds.destroy();
}
main();
```

Output on 0.3.28:

```
join  + order  -> OK    rows=1 tags=1
query + order  -> THROW TypeError: Cannot read properties of undefined (reading 'tags')
query + select -> THROW TypeError: Cannot read properties of undefined (reading 'tags')
query (neither) -> OK    rows=1 tags=1
```

Note the last line: the embedded relation loads correctly under `query` on its own. Only the combination with `order` or `select` fails.

### Cause

In the `query` strategy branch of `SelectQueryBuilder`, each relation's sub-query is configured with:

```ts
select: select ? OrmUtils.deepValue(select, relation.propertyPath) : undefined,
order: this.findOptions.order ? OrmUtils.deepValue(this.findOptions.order, relation.propertyPath) : undefined,
relations: relations ? OrmUtils.deepValue(relations, relation.propertyPath) : undefined,
```

and `deepValue` walks the path without a guard:

```ts
public static deepValue(obj: ObjectLiteral, path: string): any {
    const segments = path.split(".")
    for (let i = 0, len = segments.length; i < len; i++) {
        obj = obj[segments[i]]
    }
    return obj
}
```

For a relation inside an embedded entity, `relation.propertyPath` has two or more segments (`"meta.tags"`). `order` is `{ id: 'ASC' }`, so `obj["meta"]` is `undefined` and `undefined["tags"]` throws.

Top-level relations are unaffected because their `propertyPath` is a single segment: the loop dereferences once, yields `undefined`, and stops. `relations` does not trigger it either, because `propertyPathsToTruthyObject(['meta.tags'])` produces `{ meta: { tags: true } }`, which does contain the path. Only `order` and `select`, which legitimately have a different shape, hit the missing branch.

A null check in `deepValue` (return `undefined` once `obj` is nullish) fixes both cases.

This looks like the same root cause as #10821, which was reported through `select` with nested embeddeds and closed as not planned. Filing it again with a minimal reproduction because the `order` form is easy to hit in ordinary code: any `find()` that sorts and happens to include an embedded relation crashes.

### My Environment

| Dependency | Version |
| --- | --- |
| operating system | macOS 15 (darwin 25.2.0) |
| typeorm | 0.3.28 |
| node | 22.20.0 |
| better-sqlite3 driver | bundled |

Reproduced on `better-sqlite3`; the code path is driver independent.

### Additional Context

Hit in production code where an ORM-level `order` and an embedded relation coexist. The failure is a raw `TypeError` rather than a validation error, so callers that wrap the query in a try/catch can silently degrade instead of failing loudly.

- [x] All drivers affected

lazerg:
### Description of change

Fixes #12788

With `relationLoadStrategy: "query"`, a `find()` that loads a relation declared inside an embedded entity threw `TypeError: Cannot read properties of undefined` as soon as the same call also passed `order` or `select`. `OrmUtils.deepValue` walked the multi-segment property path (`"meta.tags"`) against an `order` legitimately shaped like `{ id: "ASC" }`, so it read `"tags"` off `undefined`.

#12776 landed the guard for that walk on `master` after this branch opened. The source change here is therefore redundant, and so were the `deepValue` unit tests, since `master` now carries an equivalent block. Both are removed in 84b6b51, which leaves this PR as test-only.

What remains is the repro from the issue, as a functional test: a many-to-many declared inside an embedded, loaded under the `query` strategy with `order` and with `select`. Both cases throw on the commit this branch started from, and both pass against current `master`. The `eager relation inside an embedded` tests that came with #12776 enter through a different path (an eager relation, with find options that name another relation), so these two cases are not covered today.

Suite on better-sqlite3, current `master` plus these tests: 3071 passing. The single failure is the Redis cache suite, which needs Docker.

### Pull-Request Checklist

      `Fixes #12788`
- [ ] Documentation has been updated to reflect this change (`docs/docs/**.md`) N/A, tests only


Cprakhar (changes requested):
@lazerg Please add the issue's repro case in test suite, because your code changes doesn't change anything, as flagged by **qodo**.

Confirm once you run the test suite against the `master` branch

lazerg:
You and qodo are right about the source change. #12776 landed the same guard on `master` after this branch opened, and merging `master` in here kept my version of the function, which hid that.

I ran the full suite on current `master` with only the test files from this branch applied: 3071 passing, both embedded cases green (the one failure is the Redis cache suite, which needs Docker). So the change to `deepValue` does nothing now. 84b6b51 removes it, and the `deepValue` unit tests with it, because `master` already carries an equivalent block.

The issue repro stays, as a functional test in `relation-load-strategy-query.test.ts`: `relations: { meta: { categories: true } }` under the `query` strategy, once with `order` and once with `select`. Both throw on the commit this branch started from. The eager-relation tests from #12776 reach `deepValue` by a different route, so `master` does not cover these two cases yet. The PR is test-only now, and I retitled it to match.

Cprakhar (approved):
Thanks for the tests.

gioboa (approved):
Thanks @lazerg 👏

michaelbromley:

```
git fetch origin
```

lazerg:
Rebased onto `master`. The branch is now a single commit, 38ca015, with the three test files under `packages/typeorm/test/`. The net diff is unchanged.

The plain rebase stopped on my old source commit, the one I reverted earlier, so I replayed the net diff on top of `master` instead. Both embedded cases pass locally on better-sqlite3, and the full matrix here is green.

lazerg:
Closing this since it has been inactive for a while.

PR #12793, opened 2026-08-19
fix(query-builder): nest child joins only when outer join semantics require it

### LINKED ISSUE #12792: Postgres: unconditional nested join parentheses in 1.1.0 constrain the planner and regress query plans
### Issue description

1.1.0 wraps a join and its children in parentheses whenever the join has children. For `LEFT` nested under `LEFT` this is semantically a no-op, but it constrains the PostgreSQL planner's join-order search and can cause large plan regressions.

Same underlying change as #12764, different symptom.

### Expected Behavior

Nest only where flat emission would change results — an `INNER` child under an `OUTER` parent. Where parent and child are both `LEFT`, 1.0.0's flat form is equivalent and leaves the planner free.

### Actual Behavior

Nesting is unconditional whenever `childJoins` is non-empty — `query-builder/SelectQueryBuilder.js`:

```js
buildJoinClause(direction, tableName, alias, condition, childJoins = "") {
    const prefix  = childJoins ? "(" : "";
    const postfix = childJoins ? ")" : "";
    ...
}
```

1.0.0 emitted `joinAttributes` as a flat list; 1.1.0 builds a tree and emits it via `createJoinTreeRecursively()`.

### Steps to reproduce

```ts
@Entity() class Country { @PrimaryColumn() id: number }

@Entity() class Profile {
  @PrimaryColumn() id: number
  @Column() countryId: number
  @ManyToOne(() => Country) @JoinColumn({ name: 'countryId' }) country: Country
}

@Entity() class User {
  @PrimaryColumn() id: number
  @Column() profileId: number
  @ManyToOne(() => Profile) @JoinColumn({ name: 'profileId' }) profile: Profile
}
```

```ts
repo.createQueryBuilder('u')
  .leftJoinAndSelect('u.profile', 'p')
  .leftJoinAndSelect('p.country', 'c')
```

1.0.0:

```sql
FROM "user" "u"
LEFT JOIN "profile" "p" ON "p"."id" = "u"."profileId"
LEFT JOIN "country" "c" ON "c"."id" = "p"."countryId"
```

1.1.0:

```sql
FROM "user" "u"
LEFT JOIN ("profile" "p"
           LEFT JOIN "country" "c" ON "c"."id" = "p"."countryId")
       ON "p"."id" = "u"."profileId"
```

Identical results — only planner freedom differs.

### Impact

PostgreSQL 18.1, a query joining 7 tables off a ~570k-row root with one `LEFT` nested under another, ordered by a left-joined column with `LIMIT`/`OFFSET`. On upgrading 1.0.0 → 1.1.0 every pre-upgrade `queryid` in `pg_stat_statements` stopped accumulating at deploy time and traffic moved to new `queryid`s containing `JOIN (`, averaging ~19s.

### Suggested fix

Gate the parentheses on semantic necessity rather than the presence of children: nest when a child would otherwise alter outer-join semantics, emit flat otherwise.

### My Environment

| Dependency | Version |
| --- | --- |
| operating system | Linux (container) |
| node.js | 24.19.0 |
| typeorm | 1.1.0 |
| PostgreSQL | 18.1 |

lazerg:
Fixes #12792

### Description of change

Since 1.1.0 a join gets wrapped in parentheses together with its children whenever it has any, so `leftJoinAndSelect("u.profile", "p").leftJoinAndSelect("p.country", "c")` now emits `LEFT JOIN ("profile" "p" LEFT JOIN "country" "c" ON ...) ON ...` instead of two flat `LEFT JOIN`s. The nesting was added in #11137 to keep an inner child join from filtering out rows that its outer parent preserves, but it is applied to every combination, including left under left, where flat emission returns exactly the same rows. The parentheses pin the join order, and the reporter saw a seven-table query regress to ~19s averages on PostgreSQL 18 after upgrading.

So the parentheses are now gated on the case they were introduced for: an inner child join under a left joined parent. Every other combination goes back to the flat form 1.0.0 emitted. The behaviour tests from #11137 pass unchanged, and the two SQL string assertions that were updated in that PR for a left-under-left chain move back to the flat expectation.

### Pull-Request Checklist

- [ ] Documentation has been updated to reflect this change (`docs/docs/**.md`) — N/A (no documented behaviour changes, only the emitted SQL shape)


test/functional/query-builder/join/query-builder-joins.test.ts
```diff
@@ -396,6 +396,31 @@ describe("query builder > joins", () => {
                 }),
             ))
 
+        it("should not nest a child left join inside its left joined parent", () => {
```
naorpeled, line 399: Nit: worth adding a three level `LEFT -> LEFT -> INNER` case, since `nestChildJoins` only inspects direct children and that is the chain a future refactor would silently regress.

test/functional/query-builder/join/query-builder-joins.test.ts
```diff
@@ -396,6 +396,31 @@ describe("query builder > joins", () => {
...
+            for (const dataSource of dataSources) {
+                // For github issue #12792
+                const query = dataSource.manager
+                    .createQueryBuilder(Post, "post")
+                    .leftJoinAndSelect("post.categories", "categories")
+                    .leftJoinAndSelect("categories.images", "images")
+                    .getQuery()
+
+                expect(query).to.not.contain("JOIN (")
```
naorpeled, line 408: Nit: this only checks that a paren is absent, so asserting the full SQL string (as the neighbouring `query-builder-select` tests do) would also catch a wrong join order or a dropped `ON`.

naorpeled (approved):
LGTM, left two nit comments 🙏

gioboa (approved):
Thanks @lazerg for your help 👏

alumni (approved):

PR #12795, opened 2026-08-20
fix(entity-manager): load eager relations when preloading an entity

### LINKED ISSUE #8944: Eager relationships not loaded when using preload
### Expected Behavior

When defining a one-to-many relationship with `eager: true` in the "many" property within the "one" side (in other words, many-side eager loading) and using `preload` method on "one" entity, "many" side is expected to be loaded eagerly as specified by documentation:

> Creates a new entity from the given plain javascript object. If the entity already exists in the database, then it loads it (**and everything related to it**), replaces all values with the new ones from the given object, and returns the new entity. The new entity is actually an entity loaded from the database with all properties replaced from the new object.

### Actual Behavior

Many side of the relationship is not loaded (left undefined).

If this behaviour is expected, then it should be detailed in documentation, as the "everything related to it" block leads the reader to think eager relationships are loaded. In other words, either code is wrong or doc is wrong.

### Steps to Reproduce

1. Define two entities with a one-to-many relationship between them.
2. Use preload on the "one" side.
3. See that "many" side is not loaded.

Truncated code for the entities:

```typescript
@Entity()
export class A {
  @OneToMany(() => B, (b) => b.a, {
    cascade: ['soft-remove'],
    eager: true,
  })
  bs!: B[];
}
```

```typescript
@Entity()
export class B {
  @JoinColumn({ name: 'aId' })
  @ManyToOne(() => A, { nullable: false })
  a!: A;
}
```

### My Environment

| Operating System    | WSL Ubuntu on Windows 10 |
| Node.js version     | 18.0.0  |
| Typescript version  | 4.6.3  | 
| TypeORM version     | 0.3.5  | 

Gauravtiwari31:
### Description of change

`Repository#preload` / `EntityManager#preload` are documented as loading the entity **"and everything related to it"**, but eager relations were never loaded — they came back `undefined`.

**Cause:** `PlainObjectToDatabaseEntityTransformer` builds its query directly with

```ts
.createQueryBuilder().whereInIds(ids).getMany()
```

Eager relations are only applied inside `SelectQueryBuilder#applyFindOptions`, which runs from `setFindOptions()`. Since `preload` never calls it, `joinEagerRelations` (join strategy) / `concatRelationMetadata` (query strategy) never run, so no eager relation is fetched.

**Change:**

1. The preload query now calls `.setFindOptions({ loadEagerRelations: true })`, so every entity it fetches loads its eager relations exactly the way `find` / `findOne` do. This applies to the preloaded entity itself and to the related entities referenced in the given plain object.
2. When assigning `*-to-many` relations from the load map, a related entity that is already present (because it was just loaded eagerly) is now **replaced** instead of appended. Without this, `preload(Post, { id, categories: [{ id: 1 }] })` would return `categories` containing the same category twice — once from the eager join, once from the load map. Replacing rather than skipping keeps the load map's own entity in the graph, so nested relations from the plain object are still linked to the object that ends up in the result.

**Before**

```ts
const post = await manager.preload(Post, { id: 1, title: "Updated" })
post.categories // undefined  ❌  (@OneToMany(..., { eager: true }))
post.author     // undefined  ❌  (@OneToOne(..., { eager: true }))
```

**After**

```ts
const post = await manager.preload(Post, { id: 1, title: "Updated" })
post.categories // [ Category { id: 1, name: "typeorm" } ]  ✅
post.author     // Author { id: 1, name: "Timber" }         ✅
```

Non-eager relations that are not part of the given plain object are still not loaded — that would silently make `preload` fetch the whole object graph. The docs for both `EntityManager#preload` and `Repository#preload` are updated to spell out what "everything related to it" means, which was the second half of the report in #8944.

**Verification:** new functional tests in `test/functional/repository/preload` cover eager `@OneToMany` / `@OneToOne` on the preloaded entity, eager relations of entities named in the plain object, and the no-duplicates case. They fail on `master` and pass with this change. The full suite passes (3047 passing, 0 failing) against the default `sqljs` config.

Closes #8944

### Pull-Request Checklist

### DIFF
--- a/docs/docs/working-with-entity-manager/5-entity-manager-api.md
+++ b/docs/docs/working-with-entity-manager/5-entity-manager-api.md
@@ -125,6 +125,9 @@ manager.merge(User, user, { firstName: "Timber" }, { lastName: "Saw" }) // same 
   and returns the new entity. The new entity is actually loaded from the database entity with all properties
   replaced from the new object.
 
+    > "Everything related to it" means the eager relations of the entity, plus the relations
+    > present in the given object. Non-eager relations that are not part of the given object are not loaded.
+
 ```typescript
 const partialUser = {
     id: 1,
--- a/docs/docs/working-with-entity-manager/6-repository-api.md
+++ b/docs/docs/working-with-entity-manager/6-repository-api.md
@@ -79,6 +79,9 @@ repository.merge(user, { firstName: "Timber" }, { lastName: "Saw" }) // same as 
 
     > Note that given entity-like object must have an entity id / primary key to find entity by. Returns undefined if entity with given id was not found.
 
+    > "Everything related to it" means the eager relations of the entity, plus the relations
+    > present in the given object. Non-eager relations that are not part of the given object are not loaded.
+
 ```typescript
 const partialUser = {
     id: 1,
--- a/src/query-builder/transformer/PlainObjectToDatabaseEntityTransformer.ts
+++ b/src/query-builder/transformer/PlainObjectToDatabaseEntityTransformer.ts
@@ -147,6 +147,9 @@ export class PlainObjectToDatabaseEntityTransformer {
                             targetWithIds.target as any,
                         )
                         .createQueryBuilder()
+                        // preload loads the entity and everything related to it,
+                        // so its eager relations must be loaded as well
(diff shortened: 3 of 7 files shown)


michaelbromley:

```
git fetch origin
```

PR #12796, opened 2026-08-20
fix: load entity files through import when require cannot parse them

### LINKED ISSUE #11570: Vitest / SyntaxError: Invalid or unexpected token
### Issue description

Vitest doesn't work with TypeORM when you import entities, migrations, subscribers by file names

### Expected Behavior

The entities, migrations and subscribers should be loaded and executed correctly by Vitest / TypeORM

### Actual Behavior

Actually, Vitest doesn't work when you specify the entities, migrations, subscribers in file paths array like this : ` entities: ["./src/entity/*.ts"],`

FAIL  sum.test.js [ sum.test.js ]
SyntaxError: Invalid or unexpected token

But if you set the entities array with Class like this : `entities: [User],`

✓ sum.test.js (1 test) 92ms
   ✓ adds 1 + 2 to equal 3 9ms

### Steps to reproduce

1. `git clone https://github.com/yolo1233211/typeorm-vitest-poc/tree/main`
2. `npm i`
3. Adapt the `./src/data-source.ts` file connection variables
4. `npm run test`

### My Environment

| Operating System    |    Windows 11     |
| Node.js version     | 22.17.0  |
| Typescript version  | 5.8.2  |

### Additional Context

_No response_

- [ ] aurora-mysql
- [ ] aurora-postgres
- [ ] better-sqlite3
- [ ] cockroachdb
- [ ] cordova
- [ ] expo
- [ ] mongodb
- [x] mysql
- [ ] nativescript
- [ ] oracle
- [ ] postgres
- [ ] react-native
- [ ] sap
- [ ] spanner
- [ ] sqlite
- [ ] sqlite-abstract
- [ ] sqljs
- [ ] sqlserver

Gauravtiwari31:
### Description of change

Entities, migrations and subscribers passed as file globs (`entities: ["./src/entity/*.ts"]`) fail to load under Vitest with `SyntaxError: Invalid or unexpected token`. Listing every class by hand (`entities: [User]`) is currently the only workaround, which is reported as still broken in 0.3.25 through 0.3.28.

**Cause:** `importOrRequireFile` picks `require` for `.ts`/`.js` whenever the nearest `package.json` is not `"type": "module"`. Vite based runners transform TypeScript for `import()`, but they leave `require` pointing at Node's plain CommonJS loader, which cannot parse TypeScript — so the file blows up before Vitest ever gets to transform it. Under `ts-node`/`tsx` this works only because those tools patch `require` itself.

**Change:** when `require` throws a `SyntaxError`, retry the file through `import()`. Vite based runners intercept dynamic imports, so the file is transformed and loads normally.

Two deliberate details:

- Only `SyntaxError` triggers the retry. A module that requires fine, or that fails for any other reason (a missing dependency, a throwing side effect), behaves exactly as before — this cannot mask a runtime failure as a load failure.
- If the retry also fails, the **original** `require` error is rethrown, not the one from `import()`. Otherwise a genuine syntax error in a user's entity file would surface on plain Node as `ERR_UNKNOWN_FILE_EXTENSION ".ts"`, which says nothing about the actual mistake.

`moduleType` is reported as `"esm"` when the retry succeeds, which is what `ConnectionOptionsReader` needs in order to unwrap `default` off the module namespace.

**Verification.** Reproduced the runner's exact conditions — CommonJS loader unable to parse the file, `import()` able to — and ran the current and patched `importOrRequireFile` against it:

```
=== BEFORE ===
require()             -> SyntaxError: Invalid or unexpected token
importOrRequireFile() -> FAILED: SyntaxError: Invalid or unexpected token

=== AFTER ===
require()             -> SyntaxError: Invalid or unexpected token
importOrRequireFile() -> OK  moduleType=esm  exports=["default","value"]
```

Two tests were added to `test/unit/util/import-utils.test.ts`. The first stubs the CommonJS loader for one fixture file to throw the way Vitest does and asserts the module still loads as ESM — it fails on `master` and passes with this change. The second asserts the original require error survives when the retry cannot help; it guards the error-quality behaviour above rather than reproducing the bug, so it passes either way by design.

Full suite: 3046 passing, 0 failing on the default `sqljs` config. No documented behaviour changes, so no docs update.

Closes #11570

### Pull-Request Checklist

- [ ] Documentation has been updated to reflect this change (`docs/docs/**.md`) — N/A, bug fix with no API or behaviour change to document

### DIFF
--- a/src/util/ImportUtils.ts
+++ b/src/util/ImportUtils.ts
@@ -26,10 +26,31 @@ export async function importOrRequireFile(
         return [require(filePath), "commonjs"]
     }
 
+    // Vite based runners such as Vitest transform TypeScript for `import()` but leave
+    // `require` on the plain CommonJS loader, which cannot parse it. Retry with
+    // `import()` when requiring fails to parse the file, and keep the original error
+    // when that does not help either, so genuine syntax errors stay readable.
+    const tryToRequireOrImport = async (): Promise<
+        [any, "esm" | "commonjs"]
+    > => {
+        try {
+            return tryToRequire()
+        } catch (error) {
+            if (!(error instanceof SyntaxError)) throw error
+
+            try {
+                return await tryToImport()
+            } catch {
+                throw error
+            }
+        }
+    }
+
     const extension = filePath.slice(filePath.lastIndexOf(".") + ".".length)
 
     if (extension === "mjs" || extension === "mts") return tryToImport()
-    else if (extension === "cjs" || extension === "cts") return tryToRequire()
+    else if (extension === "cjs" || extension === "cts")
+        return tryToRequireOrImport()
     else if (extension === "js" || extension === "ts") {
         const packageJson = await getNearestPackageJson(filePath)
 
@@ -37,8 +58,8 @@ export async function importOrRequireFile(
             const isModule = (packageJson as any)?.type === "module"
 
             if (isModule) return tryToImport()
-            else return tryToRequire()
-        } else return tryToRequire()
+            else return tryToRequireOrImport()
+        } else return tryToRequireOrImport()
     }
 
     return tryToRequire()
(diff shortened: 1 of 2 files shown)


michaelbromley:

```
git fetch origin
```

PR #12797, opened 2026-08-21
fix(cache): release query runners created by clear and getFromCache

kyungseopk1m:
### Description of change

DbQueryResultCache.clear and getFromCache build a query runner through getQueryRunner when the caller passes none, and never release it. The interface types the argument as optional and CacheClearCommand calls clear with no argument, so calling dataSource.queryResultCache.clear from application code (an admin endpoint, a scheduled job, a test teardown) leaks one pooled connection per call. After poolSize calls the pool is exhausted and every later query waits forever.

The fix releases the self-created runner in a finally block, the same way remove and storeInCache already do, and keeps the original return value. A runner passed in by the caller is left untouched. getFromCache has the same shape and is covered by the same change.

Verified with poolSize 2 on Postgres 16: without the fix the third argument-less clear times out once the pool is drained; with the fix the three clears complete and a following query still runs. Added a test that spies on createQueryRunner and asserts the self-created runner is released by clear and getFromCache, that a runner passed in by the caller is not released, and that the self-created runner is still released when the underlying operation throws.

### Pull-Request Checklist

- [ ] This pull request links a relevant issue using a closing keyword: N/A (found while reading the cache code, no existing issue)
- [x] There are new or updated tests validating the change (`test/functional/cache/query-runner-release.test.ts`)
- [ ] Documentation has been updated to reflect this change: N/A (internal behavior, no API surface change)

### DIFF
--- a/src/cache/DbQueryResultCache.ts
+++ b/src/cache/DbQueryResultCache.ts
@@ -142,57 +142,68 @@ export class DbQueryResultCache implements QueryResultCache {
      * @param options
      * @param queryRunner
      */
-    getFromCache(
+    async getFromCache(
         options: QueryResultCacheOptions,
         queryRunner?: QueryRunner,
     ): Promise<QueryResultCacheOptions | undefined> {
-        queryRunner = this.getQueryRunner(queryRunner)
-        const qb = this.dataSource
-            .createQueryBuilder(queryRunner)
-            .select()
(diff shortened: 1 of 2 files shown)


Cprakhar (approved):
Looks good.

gioboa (approved):
Thanks @kyungseopk1m 🙏

alumni (approved):

PR #11669, opened 2025-09-20
fix: include joined entity primary keys in pagination subquery

mag123c:
### Description of change
This change fixes an issue where leftJoin combined with skip/take pagination was not properly loading related entities due to missing primary keys in the pagination subquery.

The SelectQueryBuilder uses a two-query approach for pagination with joins to avoid N+1 problems. The first query selects primary keys with pagination, and the second query loads full data for those keys. The bug was in the first query not including joined entity primary keys, causing entity identification issues.

### Pull-Request Checklist
- Fixes #11662

PR #12802, opened 2026-08-27
perf: avoid quadratic grouping of raw results selected without primary keys

akdfkeks:
### Description of change

The row-index grouping fallback in `RawSqlResultsToEntityTransformer.group()` (added in #11669, used when primary key columns are not selected) computes the row index with `rawResults.indexOf(rawResult)` inside the row loop — a full array scan per row, making the grouping O(n^2). This PR uses the loop index instead, restoring O(n). The primary-key grouping path is unchanged.

Benchmark (`transform()`, median of 7 runs):

| rows | before (ms) | after (ms) |
|---|---|---|
| 10,000 | 15.4 | 4.8 |
| 30,000 | 112.8 | 14.2 |
| 100,000 | 1,129.4 | 55.3 |

Behavior is identical: `indexOf` uses reference equality and drivers return a distinct object per row, so it always resolved to the current loop index. Added tests for the fallback on the main alias (partial select without the primary key, no joins); the existing fallback tests from #11669 still pass.

### Pull-Request Checklist

- [ ] This pull request links a relevant issue using a closing keyword — N/A (regression introduced by #11669, no open issue)
- [ ] Documentation has been updated to reflect this change — N/A (internal performance fix)


Cprakhar (approved):
LGTM 💪

alumni (approved):

PR #12804, opened 2026-08-27
fix: prevent SQL injection via SelectQueryBuilder.distinctOn (CVE-2026-76848)

### LINKED ISSUE #12805: SQL injection via SelectQueryBuilder.distinctOn (CVE-2026-76848)
## SQL injection via `SelectQueryBuilder.distinctOn` (CVE-2026-76848)

**Affected versions:** `typeorm >= 0.2.21, <= 1.1.0`

### Summary

`SelectQueryBuilder.distinctOn` accepts an array of strings and stores them on the expression map without validation. For PostgreSQL-family drivers, `createSelectDistinctExpression` (in `src/query-builder/SelectQueryBuilder.ts`) joins that array and interpolates the result directly into the generated statement as `SELECT DISTINCT ON (values)`, with no escaping, quoting, or identifier validation, and without routing the values through `replacePropertyNames` or the driver's escape helper.

Because the interpolation point is a parenthesized SQL expression list rather than an identifier-only position, a supplied element can carry arbitrary expressions — including correlated subqueries. An application that forwards a client-controlled value into `distinctOn` (for example, to let a caller choose a deduplication column) allows that client to read data anywhere the application's database role can reach, through boolean or time-based inference, independently of the entity being queried.

### Impact

- **CWE:** CWE-89 (Improper Neutralization of Special Elements used in an SQL Command)
- **CVSS 4.0:** 8.7 (HIGH) — `AV:N/AC:L/AT:N/PR:N/UI:N/VC:H/VI:N/VA:N/SC:N/SI:N/SA:N`
- **Reporter:** Muhammad Sobirov (VulnCheck)

References: https://nvd.nist.gov/vuln/detail/CVE-2026-76848

Amine-H:
### Description of change

Fixes CVE-2026-76848: SQL injection via `SelectQueryBuilder.distinctOn`.

**Current behavior**

`SelectQueryBuilder.distinctOn` accepts an array of strings and stores them on the expression map without validation. For PostgreSQL-family drivers, `createSelectDistinctExpression` joins that array and interpolates the result into the generated statement as `SELECT DISTINCT ON (values)`, with no escaping, quoting, identifier validation, or allowlist, and without routing the values through `replacePropertyNames` or the driver's escape helper.

Because the interpolation point is a parenthesized SQL expression list rather than an identifier-only position, a supplied element could carry arbitrary expressions — including correlated subqueries. An application that forwards a client-controlled value into `distinctOn` (for example, to let a caller choose a deduplication column) allowed that client to read data anywhere the application's database role can reach through boolean or time-based inference, independently of the entity being queried.

**New behavior**

1. `distinctOn` rejects semicolons, using the same guard as the group-by / order-by query-builder methods, to prevent statement stacking.
2. `createSelectDistinctExpression` escapes every value through `buildDistinctOnExpression`:
   - property paths such as `post.author` are resolved to the actual database column via entity metadata, and
   - every identifier segment is quoted with the driver's `escape()` helper.

   Unknown property paths, metadata-less aliases (for example, subqueries), and arbitrary expressions are also escaped segment by segment, so no value is ever interpolated verbatim and nothing can execute as raw SQL.

**How it was verified**

- Added regression tests in `test/functional/query-builder/sql-injection/sql-injection.test.ts` covering escaping of valid property paths, aliases without metadata (including hyphens), rejection of semicolon-based statement stacking, and escaping of correlated-subquery / `DROP TABLE` / `UNION` payloads.
- Existing `distinct-on` functional tests pass (3/3).
- Full `query-builder` test suite passes (436 passing, 0 failures).
- `tsc --noEmit` and ESLint are clean.

### Pull-Request Checklist

      `Closes #12805`


Amine-H:
Thanks for the review. Both findings are addressed in commit `ebac2372c`:

**1. `distinctOn` restrictions now documented**

Updated `docs/docs/query-builder/1-select-query-builder.md`: DISTINCT ON expressions must be column names or entity property paths (e.g. `user.id`), and arbitrary SQL expressions throw an error.

**2. Registered aliases are no longer rejected**

`buildDistinctOnExpression` now resolves an exact registered alias prefix *before* applying strict identifier validation, so aliases containing characters such as hyphens (`post-item.id`) remain usable and are safely escaped. The strict validation now applies only to values that do not match a registered alias (the injection vector). Unknown raw SQL expressions are still rejected.

Tests added for both behaviors; the full `query-builder` suite passes (435 passing, 0 failures), and `tsc --noEmit` + ESLint are clean.

Amine-H:
Updated the approach in commit `95b569a7b` to match how the rest of the query builder guards these inputs, instead of allowlisting arbitrary expressions:

- **Reject statement terminators, not expressions.** `distinctOn` now uses the same semicolon guard as the group-by / order-by paths (`assertNoSemicolon`, added to the base `QueryBuilder`), preventing statement stacking.
- **Escape everything else.** `buildDistinctOnExpression` resolves known property paths to their database columns and quotes every identifier segment with the driver's `escape()`. Unknown values — including correlated subqueries, `DROP TABLE`, `UNION`, and `OR 1=1` payloads — are escaped into inert quoted identifiers rather than rejected, so they can never execute as raw SQL while legitimate aliases (e.g. `post-item`) keep working.

The strict allowlist from the earlier revision is removed. Tests and docs updated accordingly; full `query-builder` suite passes (436 passing, 0 failures), `tsc --noEmit` and ESLint clean.

Amine-H:
Addressed the two remaining findings in commit `9e33b0bf0`:

**2. Relation column paths break — fixed.** `buildDistinctOnExpression` now resolves relation-plus-referenced-column paths via a new `resolvePropertyPathColumn` helper, which mirrors the mapping the whole-query property replacement performs. `post.category.id` now emits the join column `"post"."categoryId"` instead of `"post"."category"."id"`. Added a regression test with a `ManyToOne` fixture covering this.

**4. `assertNoSemicolon` JSDoc misstates usage — fixed.** The comment no longer claims sort/group coverage; it now accurately describes the helper as a generic statement-stacking guard for raw string inputs.

Full `query-builder` suite passes (437 passing, 0 failures); `tsc --noEmit` and ESLint clean.

Amine-H:
Addressed the remaining findings in commit `7b2f7bff4`:

**1. Verbose helper comments — removed.** `buildDistinctOnExpression` no longer carries inline narrative comments; the concise JSDoc covers the behavior, consistent with neighboring expression builders.

**5. Dotted column names split — fixed.** `resolvePropertyPathColumn` now matches the complete suffix against `ColumnMetadata.databaseName` before the segment-by-segment fallback. A physical column named `profile.name` produces `"post"."profile.name"` as a single identifier instead of `"post"."profile"."name"`. Added a regression test with a dotted database column name.

**6. `assertNoSemicolon` JSDoc** was corrected in the previous commit (`9e33b0bf0`) — it no longer claims sort/group coverage and reads "Rejects a `;` in raw string inputs to prevent SQL statement stacking."

Full `query-builder` suite passes (438 passing, 0 failures); `tsc --noEmit` and ESLint clean.

alumni:
We do not intend to fix CVE-2026-76848, the current behavior is intended and changing it will break functionality.

That CVE was open without consulting us and without following our security guidelines:

https://github.com/typeorm/typeorm/blob/ac41823b9e27e3a8079ce487c9772c5ef8f1e226/SECURITY.md?plain=1#L62-L67

PR #12806, opened 2026-08-27
feat(query-builder): add streamEntities() to stream hydrated entities

### LINKED ISSUE #12714: streamEntities(): stream hydrated entities from SelectQueryBuilder
### Feature Description

`SelectQueryBuilder.stream()` returns raw, alias-prefixed driver rows and leaves entity transformation entirely to the user. There is no public API to perform that transformation, so streaming users either hand-map columns or deep-import the internal `RawSqlResultsToEntityTransformer`, which is not part of the public API and can change in any release.

This is a recurring pain point:
- #6803 asks for a public way to map raw results to entities; a collaborator pointed at `RawSqlResultsToEntityTransformer` as the workaround.
- #9870 asks the same and found it difficult to construct the transformer manually.
- #11729 reports that `stream()` yields raw rows and that to-many joins don't come out as nested objects; a collaborator clarified that this is expected behavior, since `stream()` returns raw data.

With joins, correct hydration is genuinely hard to do by hand: rows of the same root entity must be grouped, which is exactly what `getMany()` does internally — but `getMany()` buffers the entire result set, which streaming exists to avoid.

### The Solution

A `streamEntities()` method on `SelectQueryBuilder` that executes the query as a stream and yields hydrated entities:

```ts
const query = dataSource
    .getRepository(Author)
    .createQueryBuilder("author")
    .leftJoinAndSelect("author.posts", "post")

for await (const author of query.streamEntities()) {
    // author is a hydrated Author instance with its posts loaded
}
```

Implementation outline:
- Rows are hydrated with the existing internal `RawSqlResultsToEntityTransformer` (same semantics as `getMany()`); nothing internal is exported.
- Adjacent rows belonging to the same main entity are grouped before transformation, so only one entity's rows are in memory at a time.
- To guarantee that all rows of an entity are adjacent when row-multiplying joins are present (to-many relations, mapped-many joins, joins without relation metadata), the main entity's primary key is appended to the query's order as a tie-breaker. Ordering by a column of such a joined alias is rejected with a clear error unless the full primary key is ordered before it (such an order interleaves rows of different entities and cannot be streamed correctly).
- `@AfterLoad` listeners / `afterLoad` subscribers are called per emitted entity — the same behavior the MongoDB cursor already implements in `applyEntityTransformationToCursor` (per-entity `Load` broadcast on `next()`, batched in `toArray()`) — and honor the existing `callListeners(false)` flag. This keeps the documented `@AfterLoad` contract ("called each time the entity is loaded using QueryBuilder") intact; the difference from `getMany()` is granularity (per entity instead of once with the full array), which is documented.
- Early consumer exit (`break`) destroys the underlying stream and releases the query runner.

**Documented limitations** — consequences of not buffering the full result set:
- `@RelationId` properties are not loaded. Joined relations are unaffected (the hydrated objects carry their ids), and many-to-one ids can be mapped directly from the FK column; what is not supported is loading to-many related ids without joining the relation — in `getMany()` that costs one batched junction query for the whole result set, but in a stream it would become one query per entity.
- Features that require the buffered result set (or a second query pass) throw descriptive errors instead of silently diverging from `getMany()`: `take`/`skip` together with joins, `relationLoadStrategy: "query"`, and `loadRelationIdAndMap`.

`take`/`skip` with joins could later be supported without the second query `getMany()` needs: since entities are emitted incrementally, they can be implemented as emission-level counting (discard the first `skip` entity groups, stop and close the cursor after `take` entities) — the standard `take` semantics of streaming APIs.

### Considered Alternatives

- **Keyset pagination with `getMany()`** — works and fully featured, but N queries instead of one and noticeably more code for large exports.
- **Deep-importing `RawSqlResultsToEntityTransformer`** — works today, but it's internal API with no stability guarantee, and each user must reimplement the row-grouping/ordering logic, which is easy to get subtly wrong with to-many joins.
- **Exporting the transformer from the package root** — smaller change, but it would expose internal types (`QueryExpressionMap`, `Alias`) in the public API, counter to the direction of #12037 (which removed accidentally-exported internal types), and it still leaves grouping to the user.

### Additional Context

_No response_

- [x] aurora-mysql
- [x] aurora-postgres
- [ ] better-sqlite3
- [ ] capacitor
- [x] cockroachdb
- [ ] cordova
- [ ] expo
- [ ] mongodb
- [x] mysql
- [ ] nativescript
- [x] oracle
- [x] postgres
- [ ] react-native
- [x] sap
- [x] spanner
- [ ] sqlite
- [ ] sqljs
- [x] sqlserver

pri12ya871:
### Description of change

`SelectQueryBuilder.stream()` returns raw, alias-prefixed driver rows and leaves entity transformation to the caller. There is no public way to perform that transformation, so streaming users either hand-map columns or deep-import `RawSqlResultsToEntityTransformer`, which is not public API. With a to-many join, correct hydration is genuinely hard by hand: rows of the same root entity have to be grouped, which is what `getMany()` does internally — but `getMany()` buffers the whole result set, which is the thing streaming exists to avoid.

This adds `streamEntities()`, which streams the query and yields entities hydrated the same way `getMany()` hydrates them.

**Current behavior:** `stream()` sets `queryEntity = false` and hands back the driver stream. Relation-id loading, grouping and the `"Load"` broadcast never run.

**New behavior:** `streamEntities()` buffers raw rows and closes a chunk at a root-entity boundary, then hydrates and yields that chunk. Relation-id loading runs once per chunk rather than once per query.

This implements shape **A (chunked hydration)** from my comment on #12714. On the smaller questions raised there, this PR takes the conservative option in each case; all of them are still yours to overrule:

- **A separate method, not an option on `stream()`.** `stream()`'s contract is explicitly raw rows, so widening it seemed worse than adding a sibling.
- **The ordering is required, never injected.** Grouping buckets by primary-key *value*, not adjacency, so a chunk must hold every row of each root it touches. Rather than silently adding an `ORDER BY` and changing the caller's query plan, `streamEntities()` throws and names the ordering to add. Validation happens when the method is called, not on first iteration, so the error arrives at the call site.
- **A fresh transformer per chunk.** `rawRelationIdResults` is constructor-injected and derived from the row set, so rather than adding a setter to an internal class, each chunk constructs its own transformer.

Two behavioral differences from `getMany()` that are documented on the method:

- `"afterLoad"` subscribers are broadcast **per chunk**, not once for the whole result set.
- `chunkSize` (default 1000) is a *minimum*; a chunk is closed at the next root-entity boundary at or after it, so a root entity's rows are never split.

### Rejected query shapes

Several query shapes need a pass over the complete result set that streaming never has. Each one previously produced silently wrong results; all of them now throw at the call site:

- **`relationLoadStrategy: "query"`** — loads relations in a second pass, so streamed entities came back with those relations unpopulated. Covers the `setFindOptions` and data-source-default paths both.
- **Relation id loading**, from `loadRelationIdAndMap` or a `@RelationId` decorator — issues its own query. At an intermediate chunk boundary that query runs on the connection whose raw stream is still open, which deadlocks a single-client driver like postgres or cockroach.
- **`skip`/`take` with a join** — entity pagination rewrites the query into a distinct-root subquery. A streamed row-level `LIMIT` would cut a to-many collection in half and yield fewer entities than asked for.
- **Optimistic and pessimistic locks** — every other entity path validates these; lock SQL was reaching the driver unchecked.

Supporting the first two properly means either chunking that second pass or hydrating on a separate connection, and the second connection cannot see uncommitted rows from the streaming one — a semantic decision I would rather put to you than make quietly. Happy to follow up on either.

### How I verified it

`pnpm run compile` passes, `eslint` reports no new errors on the changed files, and `prettier --check` is clean.

Twenty-one new tests in `test/functional/query-builder/stream-entities/`. The seventeen validation tests run on `better-sqlite3` and need no external database; the four hydration tests are gated to `postgres`/`mysql`/`mariadb`, since they need a driver that implements `QueryRunner.stream`:

- rejects a query not ordered by the root primary key, and one whose ordering starts with a non-primary-key column
- accepts a query ordered by the root primary key
- rejects a selection without entity metadata, and a non-positive `chunkSize`
- rejects the `"query"` relation load strategy, set either through `setFindOptions` or as the data-source default
- rejects relation id loading — from `loadRelationIdAndMap` or a `@RelationId` decorator — `take` with a join, and locks, while still allowing `take` without a join and a pessimistic lock under `useTransaction(true)`
- revalidates when the builder is mutated between the call and the first iteration
- rejects a fractional or non-finite `chunkSize`
- releases the query runner when the consumer stops iterating early
- validates eagerly rather than on first iteration
- hydrates entities with to-many relations across chunk boundaries
- yields exactly what `getMany()` returns for the same query
- yields nothing for an empty table

### Pull-Request Checklist

- [x] This pull request links a relevant issue: Closes #12714
- [x] Documentation has been updated to reflect this change (`docs/docs/**.md`) — a `Streaming entities` section in `docs/docs/query-builder/1-select-query-builder.md`, covering the ordering requirement, `chunkSize`, and the rejected query shapes.

### DIFF
--- a/docs/docs/query-builder/1-select-query-builder.md
+++ b/docs/docs/query-builder/1-select-query-builder.md
@@ -909,7 +909,13 @@ needs a pass over the complete result set that streaming never has:
   decorator
 - `skip`/`take` combined with a join, where the limit would apply to joined
   rows rather than to entities
-- optimistic and pessimistic locks
+- optimistic locks, and `afterLoad` listeners or subscribers, which would be
+  broadcast while the stream is paused — call `.callListeners(false)` to
+  stream without them
+
+A pessimistic lock is allowed, but only inside a transaction: either one that
+is already active on the query runner, or one this stream starts itself with
+`.useTransaction(true)`.
 
 ## Using pagination
 
--- a/docs/docs/query-builder/1-select-query-builder.md
+++ b/docs/docs/query-builder/1-select-query-builder.md
@@ -867,6 +867,50 @@ const stream = await dataSource
     .stream()
 ```
 
+### Streaming entities
+
+If you want hydrated entities rather than raw rows, use `streamEntities`.
+It groups the streamed rows into entities the same way `getMany` does,
+without holding the whole result set in memory:
+
+```typescript
+const posts = dataSource
+    .getRepository(Post)
+    .createQueryBuilder("post")
+    .leftJoinAndSelect("post.comments", "comment")
+    .orderBy("post.id", "ASC")
+    .streamEntities()
+
+for await (const post of posts) {
+    console.log(post.comments.length)
+}
+```
+
+A root entity is only complete once all of its joined rows have arrived, so
+the query has to be ordered by the root primary key. That ordering is not
+applied for you — an unordered query throws, rather than silently returning
+an entity whose collection is split in half.
+
+Rows are buffered until a chunk can be closed on an entity boundary. Pass
+`chunkSize` to change how many rows are accumulated first; it is a minimum,
+since a chunk always closes at the next boundary at or after it:
+
+```typescript
+.streamEntities({ chunkSize: 500 })
+```
+
+Two differences from `getMany` are worth knowing. `afterLoad` subscribers
+are broadcast once per chunk rather than once for the whole result, and some
+query shapes are rejected instead of being silently mishandled, because each
+needs a pass over the complete result set that streaming never has:
+
+- the `"query"` relation load strategy
+- relation id loading, from either `loadRelationIdAndMap` or a `@RelationId`
+  decorator
+- `skip`/`take` combined with a join, where the limit would apply to joined
+  rows rather than to entities
+- optimistic and pessimistic locks
+
 ## Using pagination
 
 Most of the time when you develop an application, you need pagination functionality.
--- a/src/query-builder/SelectQueryBuilder.ts
+++ b/src/query-builder/SelectQueryBuilder.ts
@@ -2108,7 +2108,7 @@ export class SelectQueryBuilder<Entity extends ObjectLiteral>
                     return queryRunner.release()
                 return
             }
-            const rawStream = queryRunner.stream(
+            const rawStream = await queryRunner.stream(
                 sql,
                 parameters,
                 releaseFn,
--- a/src/query-builder/SelectQueryBuilder.ts
+++ b/src/query-builder/SelectQueryBuilder.ts
@@ -2017,6 +2017,160 @@ export class SelectQueryBuilder<Entity extends ObjectLiteral>
     }
 
     /**
+     * Streams entities, hydrated the same way `getMany()` hydrates them, instead
+     * of the raw alias-prefixed driver rows returned by `stream()`.
+     *
+     * Rows are buffered until a chunk can be closed on a root-entity boundary,
+     * then transformed and yielded. Because a root entity is only complete once
+     * every one of its joined rows has arrived, the query must be ordered by the
+     * root primary key; otherwise rows for one entity can be interleaved with
+     * another's and the grouping would be wrong. That ordering is not applied
+     * implicitly — it has to be requested, so the query plan stays the caller's.
+     *
+     * Note that "afterLoad" subscribers are broadcast per chunk rather than once
+     * for the whole result set, which is an intentional difference from
+     * `getMany()`.
+     *
+     * @param options
+     * @param options.chunkSize minimum number of rows to accumulate before a
+     * chunk is closed at the next root-entity boundary. Defaults to 1000.
+     */
+    streamEntities(options?: {
+        chunkSize?: number
+    }): AsyncIterableIterator<Entity> {
+        if (!this.expressionMap.mainAlias)
+            throw new TypeORMError(
+                `Alias is not set. Use "from" method to set an alias.`,
+            )
+
+        const mainAlias = this.expressionMap.mainAlias
+        if (!mainAlias.hasMetadata)
+            throw new TypeORMError(
+                `"streamEntities" can only be used when selecting from an entity. Use "stream" for raw results.`,
+            )
+
+        const metadata = mainAlias.metadata
+        if (metadata.primaryColumns.length === 0)
+            throw new TypeORMError(
+                `"streamEntities" requires ${metadata.name} to have a primary column, because rows are grouped into entities by primary key.`,
+            )
+
+        const primaryKeyAliases = metadata.primaryColumns.map((column) =>
+            DriverUtils.buildAlias(
+                this.dataSource.driver,
+                undefined,
+                mainAlias.name,
+                column.databaseName,
+            ),
+        )
+
+        this.assertOrderedByPrimaryKey(metadata, mainAlias.name)
+
+        const chunkSize = options?.chunkSize ?? 1000
+        if (chunkSize < 1)
+            throw new TypeORMError(`"chunkSize" must be a positive number.`)
+
+        return this.streamEntitiesInternal(primaryKeyAliases, chunkSize)
+    }
+
+    /**
+     * Does the actual streaming and chunked hydration.
+     *
+     * Kept separate from `streamEntities` so that misuse throws when the method
+     * is called rather than on first iteration — the body of an async generator
+     * does not run until something pulls from it, which would otherwise defer
+     * every validation error above.
+     *
+     * @param primaryKeyAliases column aliases of the root primary key
+     * @param chunkSize minimum rows to buffer before closing a chunk
+     */
+    protected async *streamEntitiesInternal(
+        primaryKeyAliases: string[],
+        chunkSize: number,
+    ): AsyncIterableIterator<Entity> {
+        this.expressionMap.queryEntity = true
+        const [sql, parameters] = this.getQueryAndParameters()
+        const queryRunner = this.obtainQueryRunner()
+        let transactionStartedByUs: boolean = false
+
+        try {
+            if (
+                this.expressionMap.useTransaction === true &&
+                queryRunner.isTransactionActive === false
+            ) {
+                await queryRunner.startTransaction()
+                transactionStartedByUs = true
+            }
+
+            const releaseFn = () => {
+                if (queryRunner !== this.queryRunner)
+                    // means we created our own query runner
+                    return queryRunner.release()
+                return
+            }
+            const rawStream = queryRunner.stream(
+                sql,
+                parameters,
+                releaseFn,
+                releaseFn,
+            )
+
+            const relationIdLoader = new RelationIdLoader(
+                this.dataSource,
+                queryRunner,
+                this.expressionMap.relationIdAttributes,
+                this.expressionMap.withDeleted,
+            )
+
+            let buffer: any[] = []
+            let bufferedRootId: string | undefined
+
+            for await (const rawResult of rawStream as AsyncIterable<any>) {
+                const rootId = primaryKeyAliases
+                    .map((alias) => String(rawResult[alias]))
+                    .join("|")
+
+                // only close a chunk on a root boundary, so every group inside
+                // it is complete
+                if (
+                    buffer.length >= chunkSize &&
+                    bufferedRootId !== undefined &&
+                    rootId !== bufferedRootId
+                ) {
+                    yield* await this.hydrateStreamedChunk(
+                        buffer,
+                        relationIdLoader,
+                        queryRunner,
+                    )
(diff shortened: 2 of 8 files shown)


pri12ya871:
Added the guard for `relationLoadStrategy: "query"` in 89ceb696, rather than leaving it as an open question.

It throws at the call site alongside the other eager validations, and covers both ways the strategy can arrive — `setFindOptions({ relationLoadStrategy: "query" })` and the data-source default. Without it the method would have silently yielded entities with unpopulated relations, which is the failure mode this API exists to avoid.

Supporting the strategy properly means chunking that second pass the same way the main hydration is chunked. That is a reasonable follow-up but a larger change, so I have kept it out of this PR. Happy to pick it up separately if you want it.

Test count is now 11, still no external database required. Updated the description to match.

pri12ya871:
Pushed a6235ca6, which fixes two real lifecycle bugs in the streaming path. Both were mine, and both were worth catching before this merged.

**The query runner was released before the final chunk was hydrated.** The release callback was passed as the stream's `onEnd`, so it fired the moment the driver stream ended — but the last chunk is hydrated *after* that point, and that hydration issues its own relation-id query and broadcasts `"Load"`. A transaction started by the builder was likewise committed against a released runner.

**Breaking out of the iteration leaked the connection.** Closing the generator early means the stream never ends, so `onEnd` never fired: the driver stream stayed open and the runner was never released. Stopping early is one of the main reasons to stream at all, so this was the more likely of the two to bite.

Cleanup now happens in a `finally`, covering the normal end and early termination alike, and rolling back a transaction we started if the consumer left before it was committed.

Two smaller changes came with it: with the stream awaited, the `AsyncIterable` cast is gone (a `ReadStream` is already async-iterable), and composite primary keys are now serialised at chunk boundaries rather than joined on `"|"`, so a key value containing the separator cannot merge two roots.

Two regression tests were added, both in the streaming-driver block: one uses `loadRelationIdAndMap` so chunk hydration issues a query after the stream has ended, and one breaks out of the iteration more times than the pool holds and then asserts ordinary queries still work.

A note on what I could and could not verify locally: the validation tests run on sqlite and pass here, but the hydration block is gated to `postgres`/`mysql`/`mariadb`, which I have no way to run — those tests, including the two new ones, are green locally only because the data source list is empty. They need CI to actually exercise them.

pri12ya871:
Pushed 79fc7c6a. I went back through the automated review above, checked each finding against the code rather than taking it at face value, and several were right — thanks to whoever wired that bot up.

Three query shapes now throw instead of returning silently wrong results:

- **Relation id loading.** `RelationIdLoader.load()` runs its own `getRawMany()` on the same query runner. At an intermediate chunk boundary that lands on the connection whose raw stream is still open, which deadlocks a single-client driver like postgres or cockroach. It happens to work at the final boundary only because the stream has already ended by then — which is exactly the kind of thing that passes a test and fails in production.
- **`skip`/`take` with a join.** Entity pagination rewrites the query into a distinct-root subquery; streaming used the plain generated SQL, so the `LIMIT` applied to joined rows and could return fewer entities than requested. `take` without a join is still allowed, since there the row count and the entity count agree.
- **Optimistic and pessimistic locks.** Every other entity path validates these; here the lock SQL was reaching the driver unchecked.

Two validations were also too loose: `chunkSize` now has to be a positive integer, because `NaN` and `Infinity` both slipped past the `< 1` test and made the boundary threshold unreachable — buffering the entire result set, which defeats the point of the API. And the ordering check now accepts a primary key ordered by database name, not only by property path, which a naming strategy can make differ.

That is 18 tests now. Fourteen of them are validation tests that run on `better-sqlite3` with no external database — including the new rejection cases — so most of this surface is now covered without needing a driver that implements `stream`.

On the two shapes I rejected rather than implemented: supporting them means either chunking that second pass, or hydrating on a separate connection. The second connection cannot see uncommitted rows from the streaming one, so it is a semantics decision rather than a mechanical fix, and I would rather have your view than pick quietly. Happy to do either as a follow-up.

pri12ya871:
@alumni could someone approve the workflow run on this PR when you get a moment? The `Tests` job has been sitting at `action_required` on every push, as it does for fork PRs.

It matters a bit more than usual here: the four hydration tests are gated to `postgres`/`mysql`/`mariadb`, because they need a driver that implements `QueryRunner.stream`, so they have never actually executed anywhere. The other fourteen tests run on `better-sqlite3` and pass locally, and so does the rest of the suite, but the part that exercises real streaming is exactly the part I cannot run.

No rush on the review itself — just the CI approval, so the failure modes show up on the PR rather than in your working tree.

pri12ya871:
Pushed 2d6570f7, working through the latest review. Four of the open findings were real:

- **`@RelationId` decorators were slipping past the guard.** Those only become `relationIdAttributes` once `RelationIdMetadataToAttributeTransformer` has run, which the streaming path never reaches — so an entity carrying the decorator had those ids silently omitted rather than rejected. The check now looks at the entity metadata too.
- **A pessimistic lock with `useTransaction(true)` was wrongly rejected.** The generator starts that transaction itself before opening the stream, so the shape is valid; my eager check only looked at an already-active runner. My own regression, introduced with the lock guard.
- **Validation could be bypassed entirely.** It ran when `streamEntities()` was called, but an async generator body does not run until something pulls from it, so mutating the builder in between skipped every check. It now runs at both points.
- **A primary key under a caller-supplied alias** made every root id `undefined`, so no boundary was ever detected and the whole result set buffered — the opposite of what the method is for. That throws now.

Also added the user documentation and the issue reference to the functional tests, so the last two checklist items are covered.

Two findings I did not act on, for the record: the `AsyncIterable`/`any` casts were already removed in an earlier commit, and the composite-key collision one quotes the `join("|")` code that commit replaced with a serialised key — both look stale rather than outstanding. Happy to be corrected if I have misread either.

That is 21 tests now, 17 of which run on `better-sqlite3` without an external database.

pri12ya871:
Pushed c6bfa3da for the two new findings. Both were real.

**`afterLoad` listeners could deadlock the stream.** The `"Load"` broadcast is awaited between chunks while the raw stream is still open, so a listener or subscriber that queries would issue it on the connection the stream is holding — the same failure already guarded for relation id loading, and it would have hit every non-final chunk. It is now rejected, with `.callListeners(false)` as the way to stream without them, so the escape hatch is one call rather than a dead end.

**The documentation I added was wrong about locks.** It listed pessimistic locks among the rejected shapes, but they are allowed inside a transaction — either one already active on the query runner, or one this stream starts via `useTransaction(true)`. There is a test asserting exactly that, so the docs contradicted the behaviour. Corrected.

On the three findings still listed as open that I have not changed: the `#12714` reference is at line 12 of the functional test, the `rawStream` `any` cast was removed in a6235ca6, and the `Streaming entities` documentation section is in `docs/docs/query-builder/1-select-query-builder.md`. All three landed in commits after the review was generated, so I believe they are stale rather than outstanding — happy to be told otherwise.

23 tests now, 19 of which run on `better-sqlite3` without an external database.

pri12ya871:
Update: the automated review is now clean — 0 bugs, after six rounds that turned up two connection leaks, two deadlock paths, a validation bypass and a documentation contradiction. All of those are fixed and covered by tests.

@alumni the one thing still outstanding is the `Tests` workflow approval. I know that is a click you have to make for fork PRs and I do not want to nag, so this is the last time I will raise it: the four hydration tests are gated to `postgres`/`mysql`/`mariadb`, so they are the only part of this feature that has never executed anywhere, and they are exactly the part where the bugs above kept surfacing. The other 19 tests run on `better-sqlite3` and pass locally.

If it is easier to leave this until someone has time to review properly, that is completely fine — I would rather wait than add noise to your queue.

pri12ya871:
Correction to my last comment: I said the review was clean, and it was when I read it, but it refreshed moments later with two more findings. Both were mine, and both were in the `afterLoad` guard I had just added — pushed a fix in a2d97dcb.

The guard was wrong in two opposite directions at once:

- **It only looked at the root entity.** The load event is broadcast for joined entities too, so a listener on a joined entity still ran mid-stream and would still have deadlocked — the guard did not cover the case it was written for.
- **It matched any subscriber at all**, ignoring `listenTo`. A subscriber scoped to a completely unrelated entity blocked streaming it would never have participated in.

It now walks the root entity plus the joined relation metadata, and applies the same `listenTo` matching that `Broadcaster.isAllowedSubscriber` uses, so the check and the broadcast agree on who actually receives the event. Three tests cover it: a listener on a joined entity throws, a subscriber for another entity does not block, and a subscriber for the streamed entity does.

26 tests now, 22 of which run on `better-sqlite3` without an external database.

michaelbromley:

```
git fetch origin
```

PR #12809, opened 2026-08-27
refactor: drop the dayjs dependency

### LINKED ISSUE #12772: Remove dayjs dependency
### Feature Description

I checked the dependencies of typeorm, and it seems that dayjs usage is very localized and is used only in one place. Having less dependencies with all the recent npm package security issues may be a good idea. This code also haven't changed in years so should be quite stable.

### The Solution

I did not find any related tests, but still it seems quite doable just to ask LLM to find and cover the use cases and remove one more dependency from typeorm. I did it and it generated a fairly reasonable regex and function, still a code to maintain, but an option to consider.  

### Considered Alternatives

Do nothing and worry less

### Additional Context

_No response_

- [ ] aurora-mysql
- [ ] aurora-postgres
- [ ] better-sqlite3
- [ ] capacitor
- [ ] cockroachdb
- [ ] cordova
- [ ] expo
- [ ] mongodb
- [ ] mysql
- [ ] nativescript
- [ ] oracle
- [ ] postgres
- [ ] react-native
- [ ] sap
- [ ] spanner
- [ ] sqlite
- [ ] sqljs
- [ ] sqlserver

pri12ya871:
### Description of change

`dayjs` is a runtime dependency used in exactly one place — parsing a date string in `DateUtils.mixedDateToDate` — so this ports that parse inline and drops the dependency.

**Current behavior:** `dayjs(mixedDate).toDate()` parses the string.

**New behavior:** a small private parser in `DateUtils` does it, reproducing dayjs's semantics exactly. No user-visible change.

### Why it is not just `new Date(value)`

Because that would be a silent, timezone-dependent data bug. `new Date("2021-04-28")` reads a date-only string as **UTC** midnight, so anyone east or west of UTC gets a different calendar date than the column holds — which, per the comment above the call, is exactly why dayjs was introduced here.

The replacement reproduces dayjs's parser rather than approximating it:

- the same regex, which deliberately does not match anything carrying a `Z` or a UTC offset
- the same fallback to `new Date(value)` for everything it does not match, which is what makes zoned strings and non-ISO forms keep working
- the same defaults — missing month to January, missing day to the 1st, fraction truncated to three digits

Two quirks are preserved **deliberately** rather than fixed, so this stays a dependency removal and not a behaviour change: a one-digit fraction is read as milliseconds rather than tenths (`…30.5` is 5ms, not 500ms), and a two-digit year goes through the `Date` constructor's 1900-offset mapping (`0099` becomes 1999). Both match dayjs today. I would be glad to change either in a separate PR if they are considered wrong, but they seemed like the sort of thing that should not ride along quietly with a refactor.

### How I verified it

I ran the old and new implementations side by side over the date-only, `YYYY-M-D`, slash-separated, compact, date-time, space-separated, `Z`, `+05:30`, `-03:00`, fractional-second, partial (`2021`, `2021-04`, `…T10`), two-digit-year, out-of-range (`2021-13-45`, `2021-04-0`), empty and unparseable forms. All 21 produce identical instants. My machine is on `+05:30`, so the local/UTC distinction is actually exercised rather than masked as it would be in UTC.

Nine unit tests are added in `test/unit/util/date-utils.test.ts`, asserting local date components rather than an ISO string so they hold in any timezone.

`pnpm run compile` passes, `eslint` reports no errors, `prettier --check` is clean, and `test:fast` passes apart from an unrelated `cli init` failure that reproduces on a clean `master` here.

### Pull-Request Checklist

- [x] This pull request links a relevant issue using a closing keyword: Closes #12772

### DIFF
--- a/package.json
+++ b/package.json
@@ -93,7 +93,6 @@
   "dependencies": {
     "@sqltools/formatter": "^1.2.5",
     "ansis": "^4.3.1",
-    "dayjs": "^1.11.23",
     "debug": "^4.4.3",
     "dedent": "^1.7.2",
     "reflect-metadata": "^0.2.2",
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -14,9 +14,6 @@ importers:
       ansis:
         specifier: ^4.3.1
         version: 4.3.1
-      dayjs:
-        specifier: ^1.11.23
-        version: 1.11.23
       debug:
         specifier: ^4.4.3
         version: 4.4.3(supports-color@8.1.1)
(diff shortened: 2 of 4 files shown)


michaelbromley:

```
git fetch origin
```

PR #12810, opened 2026-08-27
fix(logger): summarise large binary query parameters

### LINKED ISSUE #10515: OOM due to AdvancedConsoleLogger logging large buffer parameters
### Issue description

OOM due to AdvancedConsoleLogger logging large buffer parameters

### Expected Behavior

Using `AdvancedConsoleLogger` (or actually any logger based on `AbstractLogger`) should not log large Buffer parameters when log level `"query"` is enabled

### Actual Behavior

Currently, `AbstractLogger.stringifyParams` simply calls `JSON.stringify` on the parameters array. For a very large buffer, the logged SQL query we be even larger. This crashes the entire application, especially with `AdvancedConsoleLogger` which tries to highlight the SQL syntax but fails.

### Steps to reproduce

```ts
@Entity()
class Post {
  @PrimaryGeneratedColumn()
  id: number;

  @Column('blob')
  attachment: Buffer;
}

const attachment = Buffer.alloc(200_000_000); // 200 MB file

await connection.manager.save(Post, { attachment });
```

### My Environment

| Operating System    | MacOS arm64         |
| Node.js version     | 20.9.0  |

### Additional Context

_No response_

- [X] aurora-mysql
- [X] aurora-postgres
- [ ] better-sqlite3
- [X] cockroachdb
- [ ] cordova
- [ ] expo
- [ ] mongodb
- [X] mysql
- [ ] nativescript
- [X] oracle
- [X] postgres
- [ ] react-native
- [X] sap
- [X] spanner
- [ ] sqlite
- [ ] sqlite-abstract
- [ ] sqljs
- [X] sqlserver

pri12ya871:
### Description of change

`AbstractLogger.stringifyParams` calls `JSON.stringify` on the whole parameter array. A buffer serialises as one number per byte, so a binary parameter produces a log line several times the size of the data itself — enough to exhaust memory before anything is written, as @alumni reported in #10515 with a 200MB attachment.

**Current behavior:** a 200MB buffer parameter is serialised in full, and the process runs out of memory. `AdvancedConsoleLogger` makes it worse, since it then tries to syntax-highlight the result.

**New behavior:** a binary parameter larger than 1KB is logged as its type and byte length — `["<Buffer(200000000 bytes)>"]` — and everything else renders exactly as before.

This follows the approach @alumni sketched on the issue, with the threshold as a named constant so it is easy to argue with. 1KB seemed like the point where the serialised form stops being readable anyway; happy to move it.

### Preserving the existing output

Rendering the array element by element rather than in one call is where behaviour could drift, so I checked rather than assumed. Running the old and new implementations side by side over numbers, strings needing escapes, `null`, `undefined`, booleans, nested objects, dates, empty arrays, functions, small buffers, circular references and a `BigInt`, every non-large-binary case produces a **byte-identical** string.

Two of those are worth calling out, because they are what an element-wise rewrite usually gets wrong:

- `JSON.stringify` returns `undefined` for a value it cannot represent, while the array-level call renders it as `null`. The element path restores that, so `[undefined]` and `[() => 1]` still log as `[null]`.
- A parameter set that cannot be serialised at all — a circular reference, a `BigInt` — still falls back to returning the raw parameters, unchanged.

The elements are joined with `,` rather than `, ` for the same reason: it keeps existing log lines identical.

### Scope

Only `stringifyParams`. The second observation on the issue — that `SelectQueryBuilder.loadRawResults` also does `JSON.stringify(parameters)` when building a cache key — is left alone deliberately: it is a different code path with different constraints (the string is a cache identity, not a log line, so summarising a parameter there could collide two distinct queries). Glad to look at it separately if you want it fixed too.

### How I verified it

`pnpm run compile` passes, `eslint` reports no errors, `prettier --check` is clean, and `test:fast` passes apart from an unrelated `cli init` failure that reproduces on a clean `master` here.

Seven unit tests are added in `test/unit/logger/stringify-params.test.ts`, covering the large-buffer summary, a non-`Buffer` binary view, a small buffer still being serialised in full, ordinary parameters being untouched, `undefined` still rendering as `null`, and the circular-reference fallback. They need no database.

### Pull-Request Checklist

- [x] This pull request links a relevant issue using a closing keyword: Closes #10515

### DIFF
--- a/src/logger/AbstractLogger.ts
+++ b/src/logger/AbstractLogger.ts
@@ -395,9 +395,13 @@ export abstract class AbstractLogger implements Logger {
         try {
             if (!Array.isArray(parameters)) return JSON.stringify(parameters)
 
-            return `[${parameters
-                .map((value) => this.stringifyParam(value))
-                .join(",")}]`
+            // substituting before serialising keeps the array a single
+            // JSON.stringify call, so everything that is not binary is rendered
+            // exactly as it was: holes still become null, an element's index
+            // still reaches its toJSON, and the separators are unchanged
+            return JSON.stringify(
+                Array.from(parameters, (value) => this.summariseParam(value)),
+            )
         } catch (error) {
             // most probably circular objects in parameters
             return parameters
@@ -405,7 +409,7 @@ export abstract class AbstractLogger implements Logger {
     }
 
     /**
-     * Converts a single query parameter to its logged form.
+     * Replaces a query parameter that is too large to log with a summary.
      *
      * A binary parameter is summarised rather than serialised: a buffer of any
      * size becomes an array of that many numbers under `JSON.stringify`, so a
@@ -413,19 +417,15 @@ export abstract class AbstractLogger implements Logger {
      * which is enough to exhaust memory before anything is written.
      *
      * @param value one query parameter
-     * @returns the parameter as it should appear in the log
+     * @returns the parameter, or a short stand-in when it is large and binary
      */
-    protected stringifyParam(value: unknown): string {
+    protected summariseParam(value: unknown): unknown {
         if (
             ArrayBuffer.isView(value) &&
             value.byteLength > MAX_LOGGED_PARAMETER_BYTES
         )
-            return `"<${value.constructor.name}(${value.byteLength} bytes)>"`
+            return `<${value.constructor.name}(${value.byteLength} bytes)>`
 
-        const json = JSON.stringify(value)
-
-        // JSON.stringify returns undefined for values it cannot represent,
-        // which JSON.stringify of the whole array would have rendered as null
-        return json ?? "null"
+        return value
     }
 }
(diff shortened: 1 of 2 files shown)


pri12ya871:
Pushed 646ffd64. Two of the three findings above were real, and both were mine.

Rendering the array element by element was the mistake. `Array.prototype.map` skips holes, so a sparse parameter array lost them — a hole rendered as an empty slot instead of `null` — and serialising each element on its own meant a custom `toJSON` was called with an empty key rather than the element's index. Neither showed up in my original check because I had not thought to test a sparse array or a key-sensitive `toJSON`.

The fix is simpler than what it replaces: substitute the oversized binary values first, then stringify the array in a single call. The summary still happens, and every other rendering detail is the array-level `JSON.stringify` it always was — holes, indices, separators and all. I re-ran the old and new implementations over 13 shapes including both of the above; all non-binary cases are byte-identical. Two regression tests are added for them.

On the third finding, that the binary check is shallow: that is true, and deliberate for now. It matches the approach sketched on #10515, and going deeper means walking arbitrary nested structures on every logged query, which costs something on a hot path. If you would rather it recursed, say so and I will do it — but it seemed worth keeping this change to the case that actually crashes.

michaelbromley:

```
git fetch origin
```

PR #12814, opened 2026-08-30
fix: migration generation drops and creates columns instead of altering resulting in data loss

quinnfoster81-droid:
Fixes #3357

Postgres migration generation drops and recreates a column when its length changes, instead of emitting ALTER COLUMN. This moves the length comparison out of the drop/recreate condition and into the alter condition, so a length change produces ALTER COLUMN (no data loss).


quinnfoster81-droid:
Thanks for the review — both findings addressed in commit b8cc5ac.

1. **vector/halfvec dimension changes** — the `length` check is now split:
   - `vector(n)`/`halfvec(n)` length changes (dimension/element-count) are kept on the recreate (DROP + ADD) path, since `ALTER COLUMN ... TYPE` cannot coerce existing rows to a different element count.
   - ordinary `varchar`/`char` length changes now go through `ALTER COLUMN ... TYPE` instead of drop+recreate, preserving data.

2. **Functional test** — added a postgres-only regression test under `test/functional/query-runner/change-column.test.ts` that changes a varchar column's length and asserts the emitted SQL is `ALTER COLUMN ... TYPE` and contains no `DROP COLUMN`.

Full project typecheck (`tsc --noEmit -p tsconfig.json`) passes.

PR #12816, opened 2026-08-31
fix: don't capitalize camelCase output when the input starts with a separator

### LINKED ISSUE #12817: camelCase capitalizes the first character when the input starts with a separator
### Issue description

`camelCase` upper-cases the first character when the input starts with a separator (`_`, `-`, or whitespace), even though `firstCapital` defaults to false.

### Expected Behavior

```js
camelCase("_id")        // "id"
camelCase("_owner_id")  // "ownerId"
camelCase("-dash-case") // "dashCase"
```

`camelCase` should also be idempotent, the way it already is for ordinary input (there's a test named "should do nothing with strings that are already camelcase").

### Actual Behavior

```js
camelCase("_id")        // "Id"
camelCase("_owner_id")  // "OwnerId"
camelCase("-dash-case") // "DashCase"
```

And it isn't idempotent, because each pass consumes one leading separator:

```js
camelCase("__leading")  // "_leading"
camelCase("_leading")   // "Leading"
```

`firstCapital: true` is wrong in the opposite direction for the same reason. It prepends a space, that space pairs with the leading `_`, and nothing gets capitalised:

```js
camelCase("_id", true)  // "_id", expected "Id"
```

### Steps to reproduce

```js
import { camelCase } from "typeorm/util/StringUtils"
console.log(camelCase("_owner_id")) // "OwnerId"
```

### Why it matters

`DefaultNamingStrategy.joinColumnName` is `camelCase(relationName + "_" + referencedColumnName)`, so a relation property with a leading underscore gets a differently-cased join column than one without:

```ts
@ManyToOne(() => User) owner: User   // "ownerId"
@ManyToOne(() => User) _owner: User  // "OwnerId"
```

`columnName` (with embedded prefixes) and `joinTableColumnName` go through `camelCase` too.

The cause is that the `[\s-_](\w)` branch of the regex can't tell it's at the start of the string, so it treats a leading separator as a word boundary — which is the job `firstCapital` is supposed to do.

I've opened a PR with a fix and tests. Worth flagging there's a compatibility angle: fixing it changes the generated name for anyone currently relying on `OwnerId`, so it may deserve a note in the release notes.

### My Environment

| Dependency | Version |
| --- | --- |
| operating system | macOS |
| typeorm | master |
| node | 22 |

MahinAnowar:
Fixes #12817.

Found by probing `src/util` invariants rather than from a report - I filed the issue above first, per the PR triage check.

`camelCase` upper-cases the first character when its input starts with a separator, even though `firstCapital` is false:

```
camelCase("_id")        // "Id"       expected "id"
camelCase("_owner_id")  // "OwnerId"  expected "ownerId"
camelCase("-dash-case") // "DashCase"
camelCase(" leading")   // "Leading"
```

The `[\s-_](\w)` branch has no idea it's at the start of the string, so it consumes the leading separator and capitalises whatever follows — which is exactly the job of the `firstCapital` flag.

It also breaks idempotence, which this file's tests already check for ordinary input ("should do nothing with strings that are already camelcase"):

```
camelCase("__leading")  // "_leading"
camelCase("_leading")   // "Leading"
```

Each pass eats one separator, so the result keeps changing until they're gone. And `firstCapital: true` is wrong in the other direction for the same reason — `camelCase("_id", true)` prepends a space, that space pairs with the `_`, and you get `"_id"` back with nothing capitalised.

The fix strips leading separators before the existing logic runs. Every case the current tests cover is unchanged.

## Please read before merging - this can change generated names

`DefaultNamingStrategy.joinColumnName` is `camelCase(relationName + "_" + referencedColumnName)`, so a relation property with a leading underscore is affected:

```
@ManyToOne(() => User) _owner: User    // join column: "OwnerId" before, "ownerId" after
```

`columnName` (with embedded prefixes) and `joinTableColumnName` go through `camelCase` too. Anyone in that position would need a migration, so this is your call rather than mine — I'd rather flag it than have it land quietly. If you'd prefer to keep the current output, I'm happy to close this, or to narrow it to the idempotence problem only.

For what it's worth I think the current output is the accident: a property named `owner` produces `ownerId` and `_owner` produces `OwnerId`, and nothing in the code or tests suggests that difference was intended.

## Testing

Three tests added to `test/unit/util/string-utils.test.ts` covering the lowercase case, the `firstCapital` case, and idempotence. Reverting only `src/util/StringUtils.ts`, recompiling and re-running fails exactly those three:

```
1) should not capitalize the first character when the input starts with a separator
2) should still capitalize a separator-prefixed input when firstCapital is set
3) should be idempotent for separator-prefixed input
24 passing, 3 failing
```

With the fix the file is 27/27 and the whole unit suite is 142 passing. Prettier and eslint clean. I couldn't run the functional suites here since they need live databases.


MahinAnowar:
Checked both findings against main. The first one isn't caused by this change.

`camelCase("éclair", true)` already returns `" éclair"` on main today — the synthetic space is only consumed when the next character matches `\w`, which is ASCII-only, and that has nothing to do with a leading separator. All my change does is make `_éclair` behave the same as `éclair` already did:

| input | main | this PR |
|---|---|---|
| `éclair` | `" éclair"` | `" éclair"` |
| `_éclair` | `"_éclair"` | `" éclair"` |
| `""` | `" "` | `" "` |

So the leaked space in a generated migration class name is reachable on main right now with any non-ASCII first character, no underscore required. If that's worth fixing it's a separate issue with `firstCapital` and non-ASCII input, and I'm happy to file one, but folding it in here would mean changing behavior the PR isn't about.

The second finding has a real kernel to it, even though I don't think the rule fires. There is a user-facing difference, and it's narrow: an identifier that *starts* with a separator. `camelCase("_foo_bar")` was `"FooBar"` and is now `"fooBar"`, which reaches `joinColumnName` and `joinTableColumnName` when a relation or property name begins with `_`. That's the fix, not a side effect — the old output capitalized the first letter with `firstCapital` set to false.

Looking at how the repo has handled this shape of change before, the closest one I found is #11283, the `shorten` fix for `camelCase_aliases`, which got a line in the 1.0 release notes rather than anything under `docs/`. Glad to add an equivalent line there if you'd like it called out.

michaelbromley:

```
git fetch origin
```

PR #12824, opened 2026-09-02
perf: improve related entity hydration

sceccotti89:
### Description of change

`find()`, `getOne()` and `getMany()` all funnel into `RawSqlResultsToEntityTransformer`,
and it was re-deriving the same per-alias data for every single entity. Since
`transform()` recurses once per parent entity for each joined alias, that cost got
multiplied by the row count. The main culprit was `transformJoins()`, which re-resolved
every `JoinAttribute` per entity, and `parentAlias`, `relationPropertyPath` and
`mapToProperty*` are getters that re-parse their strings on every access.

All of that is now resolved once per alias/metadata pair into a small plan object and
reused. It's passed as an optional trailing argument, so the `protected` method
signatures still work if anyone overrides them.

A few smaller things while I was in there: no per-row array in `group()`, no closure per
entity in the relation-id mapping, no tree copy in `setEntityValue()`, skip the lazy
relation `forEach` in `create()` when there aren't any, and check `Number`/`String`
first in the postgres `prepareHydratedValue()` chain instead of last (every branch
compares `type` against a distinct value, so the order doesn't matter).

Hydration on its own, 2000 posts with 5 categories each, so 10 000 rows:

| | before | after |
|---|---|---|
| 2000 entities, many-to-many + many-to-one join | 4.50 ms | 1.75 ms |
| 10 000 entities, no joins | 3.32 ms | 1.47 ms |

Same fixture end to end, query time included:

| | before | after |
|---|---|---|
| `find()` | 3.0 ms | 2.4 ms |
| `find()` with relations | 19.2 ms | 14.9 ms |
| `getMany()` with two `leftJoinAndSelect` | 18.6 ms | 14.9 ms |

Measured against postgres 17, back to back on the same tree with the change stashed,
median of 60 runs.

Behaviour should be unchanged. Full suite is green, and I also ran the old and new
transformer side by side over identical raw rows and diffed the output: 19 scenarios
(single table inheritance, composite keys, views, relation ids, embeddeds, partial
selects, the no-primary-key fallback) came out identical. The new tests pass on the
parent commit too, so they pin down existing behaviour.

Only postgres got the `prepareHydratedValue` reorder, since that's the driver I could
benchmark and test against.

### Pull-Request Checklist

      `Fixes #NNNN`, `Closes #NNNN`, or `Resolves #NNNN` — N/A
- [ ] Documentation has been updated to reflect this change (`docs/docs/**.md`) — N/A,
      no API or behaviour change

### DIFF
--- a/src/driver/postgres/PostgresDriver.ts
+++ b/src/driver/postgres/PostgresDriver.ts
@@ -878,7 +878,12 @@ export class PostgresDriver implements Driver {
                   )
                 : value
 
-        if (columnMetadata.type === Boolean) {
+        if (columnMetadata.type === Number) {
+            // convert to number if number
+            value = !isNaN(+value) ? parseInt(value) : value
+        } else if (columnMetadata.type === String) {
+            // nothing to convert
+        } else if (columnMetadata.type === Boolean) {
             value = value ? true : false
         } else if (
             columnMetadata.type === "datetime" ||
@@ -990,9 +995,6 @@ export class PostgresDriver implements Driver {
                         ? parseInt(value)
                         : value
             }
-        } else if (columnMetadata.type === Number) {
-            // convert to number if number
-            value = !isNaN(+value) ? parseInt(value) : value
         }
 
         if (columnMetadata.transformer)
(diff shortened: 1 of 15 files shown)


alumni:
Thanks for looking into the performance!

To make the tests pass, you can change `type: "int"` to `type: Number` and `type: "varchar"` to `type: String` :)

michaelbromley:

```
git fetch origin
```

PR #12836, opened 2026-09-04
fix(postgres): alter varchar char length in place on postgres 3357

BayaniCyberLabs:
Fixes #3357.

We found this issue on a bounty board and wanted to send in a fix, even though the bounty is pretty old.

The Problem:
Changing a varchar/char length on Postgres (say varchar(50) to varchar(51)) currently makes TypeORM emit DROP COLUMN + ADD COLUMN, which throws away the data in that column.

What We Did:
Length-only changes on varchar, character varying, char and character now go through `ALTER TABLE ... ALTER COLUMN ... TYPE` instead, so the rows survive. varchar and character varying (and char/character) are compared as the same type, so just spelling it differently in the entity no longer triggers a drop/add. If the collation changes too, we stay on the old path. Collation-only ALTERs are unchanged.

Tests:
test/github-issues/3357/issue-3357.test.ts covers the generated SQL and a live Postgres run.

Other Found Bug:
Changing length and collation together still does DROP+ADD. On a nullable column you keep the rows but the values come back null. On NOT NULL without a transaction, the DROP can commit and the ADD then fails with 23502, so you end up with no column at all. #3357 was only about the length-only case so we left that path as is. If you want the mixed case fixed too, we can do a follow-up.

BayaniCyberLabs (Bayani Cyber Solutions LLC)
https://www.bayanicybersolutions.us

Happy to pick up other TypeORM/Postgres schema work if you have any.


BayaniCyberLabs:
- [x] Author self-review: I have reviewed the code review findings, and addressed the relevant ones.

michaelbromley:

```
git fetch origin
```

PR #12838, opened 2026-09-04
fix(postgres): preserve pending column changes when widening varchar length

na3apuo:
### Description of change

Fixes #3357.

`PostgresQueryRunner.changeColumn` currently treats *any* change to a
column's `length` the same as a type change: it drops the column and
recreates it. For `varchar`/`character varying`/`char`/`character`
columns, that means simply widening a column (e.g. `varchar(50)` →
`varchar(150)`) destroys the column's existing data — Postgres can
widen these types in place with a plain `ALTER COLUMN ... TYPE`
statement, no drop needed.

This PR widens character-varying columns in place when the new length
is greater than or equal to the old one, via a new
`isSafeColumnLengthIncrease()` helper, and only falls back to
drop-and-recreate for narrowing (which can truncate data and must
stay destructive) or for incompatible type changes.

The in-place widen is added as an *extra condition* on the existing
drop/recreate gate, not as a separate branch after it. This matters
because `changeColumn` is called with the full old/new column pair
whenever an entity changes, so a single call often carries more than
one change at once (e.g. "make this column longer *and* nullable").
Handling the length widen as a sibling branch would skip whatever
else changed on the same call; keeping it inside the existing
non-destructive path means rename/default/nullable/comment changes
requested in the same call are still applied.

There is a competing open PR (#12739) that fixes the same data-loss
bug but adds the length-widen check as a sibling `else if` to the
existing rename/default/nullable/comment branch. Because `changeColumn`
receives the full old/new column pair on every entity change, that
means a `changeColumn` call that both widens a column and changes,
say, its `nullable`/`default`/`comment`/name would silently skip the
non-length part of the change instead of applying it. This PR keeps
the widen inside the existing branch specifically to avoid that.

#### Test plan

Added a functional test in
`test/functional/query-runner/change-column.test.ts` that:
1. Adds an isolated `varchar(50)` column and inserts a row.
2. Calls `changeColumn` once, widening the length to 150 **and**
   changing `isNullable`, `default`, and `comment` in the same call.
3. Asserts all four changes took effect and the original row value
   survived.

Verified this test fails against `master` (data is lost and the
other property changes are silently dropped) and passes with this
patch. Also ran the existing `change-column`, `add-column`,
`drop-column`, `rename-column`, `create-table` functional tests plus
the full `schema-builder/column/change/**` suite against a local
Postgres instance — all green, no regressions observed.

Scoped to the Postgres driver only, matching the issue's reproduction.
MySQL/MSSQL/SQLite/Oracle have separate `changeColumn` implementations
not touched here. Narrowing a column's length is unaffected — still
goes through drop-and-recreate, since narrowing can truncate data and
needs that safety net.

### Pull-Request Checklist

- [ ] Documentation has been updated to reflect this change (`docs/docs/**.md`) - N/A, internal driver behavior fix, no public API/docs change


na3apuo:
Thanks for the review, Qodo bot - all three findings from the earlier pass were real and are fixed in `404bb3a7`:

1. **Unbounded widening (requirement gap)**: `isSafeColumnLengthIncrease` parsed an empty `length` (Postgres representation of unbounded `varchar`) as `NaN` and rejected it, so `varchar(n) -> varchar` still went through drop-and-recreate. Now treated as always-safe when the old length was bounded.
2. **Collation change discarding length (bug)**: the pre-existing collation-update query emitted `TYPE ${newColumn.type} COLLATE "..."` without a length modifier. That was latent before this PR (any length delta always took the destructive path), but became reachable once combined length+collation changes go through the in-place path. Fixed by emitting the full type via `createFullType()` so the length is preserved.
3. **Generated column widening (bug)**: an existing `STORED` generated column would pass the safe-length gate and hit `ALTER COLUMN ... TYPE`, which Postgres rejects for generated columns. `isSafeColumnLengthIncrease` now excludes columns where `oldColumn.generatedType === "STORED"`, so those keep going through drop-and-recreate as before.

Added regression tests for all three cases in `change-column.test.ts`. Self-review checkbox ticked.

(Apologies for the placeholder text that briefly appeared here - fixed now.)

michaelbromley:

```
git fetch origin
```

PR #12841, opened 2026-09-05
fix(metadata-builder): keep the STI discriminator index on entities with embeddeds

### LINKED ISSUE #12840: Auto-created STI discriminator index is silently dropped when the entity has an embedded column
### Issue description

The index TypeORM auto-creates on a single-table-inheritance discriminator column is silently dropped when the STI base entity has at least one embedded column and at least one generated column.

### Expected Behavior

An STI base entity declared with `@TableInheritance()` gets an index on its discriminator column, as `EntityMetadataBuilder.createKeysForTableInheritance()` intends, regardless of whether the entity also has embedded columns.

### Actual Behavior

With an embedded column (`@Column(() => Something)`) present alongside a generated column (e.g. `@PrimaryGeneratedColumn()`), the discriminator index is missing from `EntityMetadata.indices` and is therefore never created in the schema. No error and no warning is emitted — `synchronize`/migration generation simply produces a table without that index.

Root cause is an array-identity subtlety in `src/metadata-builder/EntityMetadataBuilder.ts`.

`computeEntityMetadataStep2()` derives `indices` from `ownIndices`:

```ts
entityMetadata.indices = entityMetadata.embeddeds.reduce(
    (indices, embedded) => indices.concat(embedded.indicesFromTree),
    entityMetadata.ownIndices,
)
```

- With **zero** embeddeds, `Array.prototype.reduce` with an initial value and an empty array returns the initial value **by reference**, so `indices` *is* `ownIndices`.
- With **one or more** embeddeds, `concat` produces a **new** array, so `indices` is a copy that is no longer connected to `ownIndices`.

Later in `build()`, `createKeysForTableInheritance()` pushes the auto index onto `entityMetadata.indices`:

```ts
entityMetadata.indices.push(
    new IndexMetadata({ entityMetadata, columns: [entityMetadata.discriminatorColumn!], ... }),
)
```

In the zero-embedded case that push also lands in `ownIndices` (same array), so it survives. In the embedded case it lands only on the detached copy.

`build()` then re-runs `computeEntityMetadataStep2()` for every entity that has a generated column:

```ts
if (generated) {
    ...
    column.build(this.dataSource)
    this.computeEntityMetadataStep2(entityMetadata)
}
```

That re-run reassigns `indices` from `ownIndices` + embedded indices, and the auto-created discriminator index is discarded.

So the bug needs all three of: STI base, ≥1 embedded column, ≥1 generated column. A `@PrimaryGeneratedColumn()` id — which nearly every entity has — supplies the third.

An explicit `@Index()` on the discriminator column is unaffected: it comes from metadata args into `ownIndices`, so it is recreated by every step-2 run. That is also why the existing dedup guard in `createKeysForTableInheritance()` (`isDiscriminatorColumnAlreadyIndexed`) still behaves correctly once the underlying bug is fixed — no double index for users who already added the decorator by hand.

### Steps to reproduce

Minimal reproduction (`better-sqlite3`, in-memory — the outcome is decided at the metadata layer, so every driver is affected):

```ts
import "reflect-metadata"
import {
    ChildEntity, Column, DataSource, Entity,
    PrimaryGeneratedColumn, TableInheritance,
} from "typeorm"

class Meta {
    @Column({ type: String, nullable: true })
    note?: string
}

@Entity({ name: "with_embedded" })
@TableInheritance({ column: { type: String, name: "discriminator" } })
class WithEmbedded {
    @PrimaryGeneratedColumn()
    id!: number

    @Column(() => Meta)
    meta!: Meta
}

@ChildEntity()
class WithEmbeddedChild extends WithEmbedded {}

@Entity({ name: "without_embedded" })
@TableInheritance({ column: { type: String, name: "discriminator" } })
class WithoutEmbedded {
    @PrimaryGeneratedColumn()
    id!: number
}

@ChildEntity()
class WithoutEmbeddedChild extends WithoutEmbedded {}

const ds = new DataSource({
    type: "better-sqlite3",
    database: ":memory:",
    entities: [WithEmbedded, WithEmbeddedChild, WithoutEmbedded, WithoutEmbeddedChild],
    synchronize: true,
})

await ds.initialize()

for (const target of [WithEmbedded, WithoutEmbedded]) {
    const metadata = ds.getMetadata(target)
    const indices = metadata.indices.filter(
        (index) =>
            index.columns.length === 1 &&
            index.columns[0].databaseName === "discriminator",
    )
    console.log(metadata.tableName, "discriminator indices:", indices.length)
}
```

Output on `master` (`ac41823b9`, 1.1.1):

```
with_embedded discriminator indices: 0
without_embedded discriminator indices: 1
```

Inspecting the created tables through a query runner gives the same result: the `with_embedded` table has no index on `discriminator`.

Removing the `@Column(() => Meta)` line makes the index appear. Removing `@PrimaryGeneratedColumn()` (using a plain primary column instead) also makes it appear, because nothing then triggers the second `computeEntityMetadataStep2()` run.

### My Environment

| Dependency         | Version                        |
| ---                | ---                            |
| Operating System   | macOS 15 (also Linux CI)       |
| Node.js version    | 24.14.1                        |
| Typescript version | 5.9.3                          |
| TypeORM version    | master `ac41823b9` (1.1.1), also 1.0.x and 0.3.x |

Affected on both active branches: the code shape is identical on `master` and on `v0` (`entityMetadata.indices = entityMetadata.embeddeds.reduce(...)` plus the post-hoc push in `createKeysForTableInheritance`). Verified reproducing on 0.3.28 and on 1.1.0 as well as current `master`.

### Additional Context

This is not a theoretical edge case. Any framework that adds an embedded column to a base entity hits it for every STI hierarchy it defines.

Vendure is the clearest example: every Vendure entity carries an embedded `customFields` column and a generated id, so **all** of its STI discriminator indexes are missing in every Vendure installation — `history_entry.discriminator`, `session.type`, `region.discriminator`, `stock_movement.discriminator`.

The production consequence on our own installation: a `history_entry` table of roughly 2M rows, with admin `COUNT` queries filtered by `discriminator` sequentially scanning it under peak load. Reported downstream as vendurehq/vendure#5286, with a workaround (explicit `@Index()` decorators on those discriminator columns) proposed in vendurehq/vendure#5287 — but the missing index originates here, and affects any application with this entity shape.

The fix looks small: make the auto-created index land in `ownIndices` (which every `computeEntityMetadataStep2()` run preserves) instead of in the derived `indices` array. I have a patch and tests and will open a PR against `master`.

All drivers — the index is dropped at the metadata layer, before any driver is involved.

brmk:
### Description of change

The index that `EntityMetadataBuilder.createKeysForTableInheritance()` creates on an STI discriminator column is silently dropped when the entity has at least one embedded column and at least one generated column.

`computeEntityMetadataStep2()` derives `indices` from `ownIndices`:

```ts
entityMetadata.indices = entityMetadata.embeddeds.reduce(
    (indices, embedded) => indices.concat(embedded.indicesFromTree),
    entityMetadata.ownIndices,
)
```

With no embeddeds, `reduce` returns the initial value by reference, so `indices` *is* `ownIndices` and the later `indices.push(...)` in `createKeysForTableInheritance()` also lands in `ownIndices`. With one or more embeddeds, `concat` returns a new array, so the pushed index lives only on that detached copy. `build()` then re-runs `computeEntityMetadataStep2()` for every entity that has a generated column, which reassigns `indices` from `ownIndices` and discards the auto-created index. No error, no warning — the table is simply created without the index.

All three conditions are needed: STI base, at least one embedded column, at least one generated column. A `@PrimaryGeneratedColumn()` id supplies the third, so in practice any framework that puts an embedded column on its base entity loses every STI discriminator index (see the issue for the Vendure case and its production impact).

The fix pushes the auto-created index onto `ownIndices` and then calls `computeEntityMetadataStep2()` to rebuild `indices` from it — the same push-and-recompute pattern `EntityMetadataBuilder` already uses for the indices it creates for relations. `ownIndices` is what every `computeEntityMetadataStep2()` run preserves, and the recompute happens before indices are built.

Behavior that is deliberately unchanged:

- the existing `isDiscriminatorColumnAlreadyIndexed` guard still short-circuits when the user has their own `@Index()` on the discriminator column, so no duplicate index appears for projects that already worked around this by hand;
- entities without embeddeds keep exactly the one index they have today.

The same code shape exists on the `v0` branch, so 0.3.x is affected as well. This PR targets `master`; happy to open the equivalent backport if you want it there.

**Verification.** Ran locally against `better-sqlite3` (`ormconfig.json` limited to that driver): the full suite passes, and the new test file fails on `master` without the source change (`expected [] to have a length of 1`) and passes with it. The behavior is decided at the metadata layer, before any driver is involved, so the outcome is driver-independent; the new test enables the same driver list as the closest existing test (`test/github-issues/10496`) minus `spanner`, and I did not run the non-sqlite drivers locally.

New tests in `test/functional/table-inheritance/single-table/discriminator-index-with-embedded`:

- STI base with an embedded column and a generated id — the discriminator index exists, both in `EntityMetadata.indices` and on the created table;
- same shape plus a user-defined `@Index()` on the discriminator column — exactly one index, keeping the user's name (dedup guard intact);
- STI base without embedded columns — still exactly one index (no regression).

Closes #12840

### Pull-Request Checklist

- [x] Documentation has been updated to reflect this change (`docs/docs/**.md`) — N/A, this restores documented/intended behavior and changes no public API

### DIFF
--- a/src/metadata-builder/EntityMetadataBuilder.ts
+++ b/src/metadata-builder/EntityMetadataBuilder.ts
@@ -1146,16 +1146,18 @@ export class EntityMetadataBuilder {
             return
         }
 
-        entityMetadata.indices.push(
-            new IndexMetadata({
-                entityMetadata: entityMetadata,
-                columns: [entityMetadata.discriminatorColumn!],
-                args: {
-                    target: entityMetadata.target,
-                    unique: false,
-                },
-            }),
-        )
+        const index = new IndexMetadata({
+            entityMetadata: entityMetadata,
+            columns: [entityMetadata.discriminatorColumn!],
+            args: {
+                target: entityMetadata.target,
+                unique: false,
+            },
+        })
+
+        // computeEntityMetadataStep2 rebuilds "indices" from "ownIndices".
+        entityMetadata.ownIndices.push(index)
+        this.computeEntityMetadataStep2(entityMetadata)
     }
 
     /**
--- a/test/functional/table-inheritance/single-table/discriminator-index-with-embedded/discriminator-index-with-embedded.test.ts
+++ b/test/functional/table-inheritance/single-table/discriminator-index-with-embedded/discriminator-index-with-embedded.test.ts
@@ -1,0 +1,102 @@
+import "reflect-metadata"
+
+import { expect } from "chai"
+
+import {
+    closeTestingConnections,
+    createTestingConnections,
+    reloadTestingDatabases,
+} from "../../../../utils/test-utils"
+import type { DataSource } from "../../../../../src/data-source/DataSource"
+import type { EntityMetadata } from "../../../../../src/metadata/EntityMetadata"
+
+import {
+    EmbeddedBase,
+    EmbeddedChild,
+    IndexedBase,
+    IndexedChild,
+    PlainBase,
+    PlainChild,
+} from "./entity"
+
+function discriminatorIndices(metadata: EntityMetadata) {
+    return metadata.indices.filter(
+        (index) =>
+            index.columns.length === 1 &&
+            index.columns[0].databaseName ===
+                metadata.discriminatorColumn!.databaseName,
+    )
+}
+
+// https://github.com/typeorm/typeorm/issues/12840
+describe("table-inheritance > single-table > discriminator-index-with-embedded", () => {
+    let dataSources: DataSource[]
+
+    before(async () => {
+        dataSources = await createTestingConnections({
+            entities: [
+                EmbeddedBase,
+                EmbeddedChild,
+                IndexedBase,
+                IndexedChild,
+                PlainBase,
+                PlainChild,
+            ],
+            schemaCreate: true,
+            dropSchema: true,
+            enabledDrivers: [
+                "better-sqlite3",
+                "cockroachdb",
+                "mariadb",
+                "mssql",
+                "mysql",
+                "oracle",
+                "postgres",
+            ],
+        })
+    })
+
+    beforeEach(() => reloadTestingDatabases(dataSources))
+
+    after(() => closeTestingConnections(dataSources))
+
+    it("should index the discriminator column of an entity with an embedded column", () =>
+        Promise.all(
+            dataSources.map(async (dataSource) => {
+                const metadata = dataSource.getMetadata(EmbeddedBase)
+
+                expect(discriminatorIndices(metadata)).to.have.length(1)
+
+                const queryRunner = dataSource.createQueryRunner()
+                await queryRunner.connect()
+                const table = (await queryRunner.getTable("embedded_base"))!
+                await queryRunner.release()
+
+                const indices = table.indices.filter(
+                    (index) =>
+                        index.columnNames.length === 1 &&
+                        index.columnNames[0] === "discriminator",
+                )
+
+                expect(indices).to.have.length(1)
+            }),
+        ))
+
+    it("should not add a second index when the discriminator column is already indexed", () => {
+        dataSources.forEach((dataSource) => {
+            const metadata = dataSource.getMetadata(IndexedBase)
+            const indices = discriminatorIndices(metadata)
+
+            expect(indices).to.have.length(1)
+            expect(indices[0].name).to.be.equal("IX_indexed_base_discriminator")
+        })
+    })
+
+    it("should keep indexing the discriminator column of an entity without embedded columns", () => {
+        dataSources.forEach((dataSource) => {
+            const metadata = dataSource.getMetadata(PlainBase)
+
+            expect(discriminatorIndices(metadata)).to.have.length(1)
+        })
+    })
+})
--- a/test/functional/table-inheritance/single-table/discriminator-index-with-embedded/entity/EmbeddedBase.ts
+++ b/test/functional/table-inheritance/single-table/discriminator-index-with-embedded/entity/EmbeddedBase.ts
@@ -1,0 +1,28 @@
+import {
+    ChildEntity,
+    Column,
+    Entity,
+    PrimaryGeneratedColumn,
+    TableInheritance,
+} from "../../../../../../src"
+
+import { Meta } from "./Meta"
+
+/**
+ * Single table inheritance with both an embedded column and a generated column.
+ */
+@Entity({ name: "embedded_base" })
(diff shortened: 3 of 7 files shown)


brmk:
Moved the regression tests into `test/functional/table-inheritance/single-table/discriminator-index-with-embedded`, keeping the #12840 reference above the describe. `test/github-issues/12840` is gone; the three cases are unchanged.

michaelbromley (changes requested):
Thanks — the diagnosis is right and the write-up made this easy to follow. I reproduced it locally: the new test fails on the parent commit with `expected [] to have a length of 1` and passes with the change. The rest of the sqlite suite is unaffected.

A few changes before this goes in.

## Use the existing push-and-recompute pattern

The current fix adds a second writer to `entityMetadata.indices`:

```ts
entityMetadata.indices = entityMetadata.indices.concat(index)
entityMetadata.ownIndices.push(index)
```

`indices` is derived from `ownIndices`. This bug exists because the old code assigned to `indices` directly, and after this change two places still assign to it. `createKeysForTableInheritance` is the only assignment to `indices` in `src/` outside `computeEntityMetadataStep2`.

`EntityMetadataBuilder` already handles this case three times, at lines 244-248, 254-259 and 277-281:

```ts
entityMetadata.ownIndices.push(index)
this.computeEntityMetadataStep2(entityMetadata)
```

Push to `ownIndices`, then recompute `indices`. The recompute runs before the `index.build()` loop at line 366, so the new index is still built. I applied this to `createKeysForTableInheritance` and ran the sqlite suite: same result, your three tests green.

This also lets you drop the nine-line comment. One line covers it:

```ts
// computeEntityMetadataStep2 rebuilds "indices" from "ownIndices".
```

The second half of the current comment describes how `computeEntityMetadataStep2` builds `indices`. 

## PR description

The description says the tests are in `test/github-issues/12840`. They are in `test/functional/table-inheritance/single-table/discriminator-index-with-embedded/`. The functional directory is the right home; please fix the path in the description.

brmk:
Applied the push-and-recompute pattern in cd5d8a787 and reduced the comment to the one line. The description now points at `test/functional/table-inheritance/single-table/discriminator-index-with-embedded` instead of the wrong `test/github-issues/12840` path.

michaelbromley (approved):

gioboa (approved):
Thanks @brmk

PR #12853, opened 2026-09-09
refactor!: remove tinyglobby and node 20 BREAKING CHANGE

Samuron:
Node 20 support ended on April 2026 so I open this as a discussion.  It would be also very nice to remove the dependency and replace with native alternative


alumni:
That's the idea in v2.0 :)

Samuron:
@alumni from QODO review it seems that there was some behavior of tinyglobby that this PR will break. I do not understand completely what the issue is, but it may be a good thing to also drop this behavior and rely on the globs how they are implemented by node itself, anyway, it is not my call to decide, so ping me if I need to dig and restore previous behavior

alumni (changes requested):
@Samuron Qodo is looking for breaking changes. The behavior it is referring is mentioned in https://superchupu.dev/tinyglobby/comparison#trailing-slashes

We recently switched from `glob` to `tinyglobby` and we did not bother about the breaking changes because we did this switch in a major version (they should all be more-or-less similar for most people).

In this case, dropping Node 20 support is already a breaking change, so if there are any behavioral differences between `tinyglobby` and `fs.glob` that will be fine, since this change will have to wait until `2.x` anyway.

michaelbromley:

```
git fetch origin
```

Issue #11691, opened 2025-09-30
Replace Gulp with Nx for build tasks


michaelbromley:
Gulp is no longer maintained and is preventing us from upgrading certain dependencies.

Nx is a more modern and well-maintained solution that can handle the tasks that Gulp currently performs, and also allows us to further grow to use a monorepo architecture in future.

smith-xyz:
@michaelbromley @alumni anyone taking this on yet? I was reviewing the current setup and tried to put together a plan with two options: https://gist.github.com/smith-xyz/59db99ac0326d20a72d271226b325c4b

Let me know if any of that aligns or needs adjusted. Happy to contribute to this one.

alumni:
Postponed for now since it will interfere with our other tasks

Issue #2479, opened 2018-07-07
Provide custom drivers support and split different databases into sub-packages


pleerock:
1. Provide a way to create custom driver more easily
2. Think if we can split different drivers into their own packages, e.g. `@typeorm/mysql`, `@typeorm/postgres`, `@typeorm/sqlite`, `@typeorm/sqljs-node`, `@typeorm/sqljs-browser`
       2.1. Think if we need a `@typeorm/core`

The reason behind `#1` is not to have new drivers like #2240 but instead have a flexible mechanizm to allow users to create their own drivers. 

The reasons behind `#2` are numerous. It will allow us to make a specific functionalities in their own packages, pack their own drivers with the packages, etc.

> Think if we need a `@typeorm/core`

some implementation detail, let's say do we want to have `Entity` inside core or inside each driver. If we do it inside core then it will be possible to use those entities across different drivers, if we do it inside their own packages then we'll be able to "give" additional functionality.

We already had multiple discussions before, I'll need to find all related issues and link them here.

adding @daniel-lang

stevefan1999:
For the lulz, just derive the drivers as dependencies and inject them on-demand with TypeDI🙃.

However, the GCDs of the drivers aren't...so well compatible to each other, for example, mongo has another set of API, and I think so the interface is hard to define eh?

Kononnable:
Some of related issues:
- #344
- #2318
- #2404
- #857

Kononnable:
If we choose splitting drivers with `@typeorm/core` (`#2.1`) we should also decide about splitting the codebase:
1. keep single git repository for core and drivers
    - complicated build/publish
    - one repository for commits
2. separate git repositories for npm packages
    - adding new driver is just cloning repository and making necessary changes
    - some logic for CI tests - commit to core should run tests for all drivers(shouldn't be to hard to write)
    - less complicated file structure

We should also think about running automated CI tests for example projects for every typeorm build(even if we don't choose `#2`).  It would help to prevent situations like #2531 and make sure we keep examples up-to-date.

annymosse:
[...]

Issue #97, opened 2016-11-21
@types/node missing from package.json


rightisleft:
When cloning the repo - the @types/node library is missing

pleerock:
its not missing https://github.com/typeorm/typeorm/blob/master/package.json#L62

rightisleft:
odd - i'm getting node import errors on require

pleerock:
did you fix it?

pleerock:
feel free to reopen it if you still have something to add here

Issue #95, opened 2016-11-20
Invalid equality check of metadata objects


Crotalus:
When searching for metadata for a specific metaobject this comparison is made:
`metadata.target === target`

https://github.com/typeorm/typeorm/blob/master/src/metadata-args/collection/EntityMetadataCollection.ts#L14
https://github.com/typeorm/typeorm/blob/master/src/metadata-args/collection/EntityMetadataCollection.ts#L18

Problem is that it fails to match against the same entity but included later in another file, I guess they aren't equal in the javascript object sense due to being separate objects in memory?

When changing the comparison to `metadata.target.toString() === target.toString()` it works as expected. I don't know if that's the optimal solution though.

pleerock:
no, its source code is correct. Probably you have something wrong setup in your code. It expects class/function type, and each one has its own instance. Provide your minimal code lets see where the issue is

mohyt:
@Crotalus : Do you have working code for this? I am also hitting same issue.

PR #12854, opened 2026-09-09
chore: consolidate the repository into a pnpm workspace and move typeorm into packages

michaelbromley:
## What this does

Turns the repository into a real pnpm workspace and moves the `typeorm` package out of the repository root into `packages/typeorm`, so that every published package lives in one place and publishing more than one of them stops being a manual, per-package affair.

Before this, the repo already shipped three packages (`typeorm`, `@typeorm/codemod`, `@typeorm/legacy-naming-strategies`) but was not a monorepo in any technical sense: no `pnpm-workspace.yaml`, a separate `pnpm-lock.yaml` per sub-package, a root `tsconfig.json` that excluded `packages`, and a root eslint config that ignored `packages/**`. Every workflow enumerated the packages by hand.

`@typeorm/legacy-naming-strategies` also declared `"typeorm": "1.1.1"`, so it tested against the **published** package rather than this repository's source. A breaking change to core could not be caught by it.

## Layout

| Path                                | Package                             |
| ----------------------------------- | ----------------------------------- |
| `packages/typeorm`                  | `typeorm`                           |
| `packages/codemod`                  | `@typeorm/codemod`                  |
| `packages/legacy-naming-strategies` | `@typeorm/legacy-naming-strategies` |

The root keeps `docs/`, `docker/`, `docker-compose.yml`, `resources/`, the markdown files and the shared tooling config. A new private root `package.json` holds the workspace and delegates `compile`, `package`, `test` and `typecheck` to `packages/typeorm`, so the familiar root commands still work.

## For anyone with an open pull request

Everything moved as a **pure rename**, no file contents changed, so git records all 3292 renames at 100% similarity and can usually replay a branch across the move on its own.

There is one catch worth knowing: git's default rename limit is well below the number of files that moved, and without raising it git sees deletes plus adds and every touched file conflicts. So the plain instruction is:

```shell
git fetch origin
```

[`scripts/migrate-pr-to-packages-layout.md`](scripts/migrate-pr-to-packages-layout.md) documents this with a path mapping table, a `scripts/migrate-pr-to-packages-layout.sh` helper that backs your branch up and does the rebase for you (plus a `--replay` fallback that rewrites the paths of your net diff when the history is not worth saving), and a prompt you can hand to a coding agent. Both script modes were tested against a simulated pull request branch based on the current `master`.

## Changes

- Adds `pnpm-workspace.yaml` covering `packages/*` and `playground`. `docs` is excluded; it is a Docusaurus site with its own dependency tree and deploy workflow.
- Deletes the two sub-package lockfiles. One root lockfile now covers every project.
- Moves `src`, `test`, `extra`, `package.json`, `gulpfile.ts`, the tsconfigs, `.mocharc.json`, `.c8rc.json`, `stryker.config.json`, `ormconfig.sample.json`, `sonar-project.properties` and `eslint.config.mjs` into `packages/typeorm`.
- Splits the root manifest: the published fields stay with the package, the workspace settings and shared tooling move to a new private root manifest.
- Adds `publishConfig.directory` pointing at `build/package`. The package manifest describes the built layout (`main` is `./index.js`, which only exists after a build), so a plain workspace link does not resolve; this points pnpm at the build output instead. `copyPackageFile` strips the entry so it never reaches the published package.
- Switches `@typeorm/legacy-naming-strategies` and `playground` to `"typeorm": "workspace:*"`.
- Updates the five workflows and two composite actions for the new paths. The core test workflows now default to `working-directory: packages/typeorm`; the docker compose steps override back to the repository root.
- Removes the per-package `pnpm install` steps. A single install covers the workspace, and root `lint` now fans out with `pnpm -r`.
- Builds the core package in `tests-legacy-naming-strategies`, and adds the core sources to that job's path filter so a change under `src/` actually runs it.
- Adds `pnpm-lock.yaml` to the path filters. With the sub-package lockfiles gone, a dependency-only change would otherwise have matched no filter and run no tests.
- Publishes `LICENSE` with the package (see below).
- Marks `playground` private, adds `dist/` to `.prettierignore`.

## The published package

The built tree was compared file by file against a pre-move baseline built from `master` in a separate worktree:

**2949 files, 0 added, 0 removed, 0 with differing content.** Byte-identical, by SHA-256 per file, sourcemaps included.

One deliberate change to the published artifact. The real `typeorm@1.1.1` tarball on npm contains **no LICENSE file**, despite the manifest declaring `"license": "MIT"`. Adding `pnpm-workspace.yaml` made pnpm start copying the root `LICENSE` into the tarball as a side effect (2948 files to 2949). Rather than leave that to a pnpm heuristic, `gulpfile.ts` now copies `LICENSE` explicitly the same way it already copies `README.md`. Verified deterministic: the tarball is 2949 files with the licence whether or not the workspace file is present, and the only difference against the current npm artifact is that added file.

The resolved dependency graph for core is unchanged: the root importer has zero differences against `master`, and no package was removed from the graph.

## Verification

Beyond the suites, the artifact was published to a local verdaccio registry with the same command CI runs, and installed into clean projects:

- The registry manifest contains no `publishConfig`, and no `build/` directory leaks into the tarball.
- npm, pnpm and yarn each installed it and passed: CommonJS `require`, ESM named and default imports, the deep subpath `typeorm/util/StringUtils`, a sql.js DataSource round-trip and a QueryBuilder query.
- All three CLI bins link and `cli.js version` runs.
- TypeScript with `strict: true` and `skipLibCheck: false` type-checks against decorators and `Repository<T>` generics.
- All 22 `browser`-map targets, `browser/index.js`, `index.mjs`, both shims and the react-native condition are present.
- The nightly path (`pnpm version` inside the build directory, then publish) works inside a workspace.
- `@typeorm/codemod` publishes, installs and runs, with no unresolved `workspace:` specifier.
- After the move: core builds, a real sqlite functional suite passes, `ormconfig.json` still resolves from its new location, and `@typeorm/codemod` (376) and `@typeorm/legacy-naming-strategies` (4) pass.

The source linking was confirmed to work by renaming `hash` in `src/util/StringUtils.ts`, rebuilding, and watching `@typeorm/legacy-naming-strategies` fail with `TS2305`. Under the previous registry pin that rename passed CI.

## Known follow-ups, not in this PR

- **Publish ordering.** `workspace:*` resolves its published version from the built `package.json`, which the nightly step rewrites in place. Packing `@typeorm/legacy-naming-strategies` after a nightly bump yields `devDependencies.typeorm: "1.1.1-nightly.<date>"`. Inert today, since that package is only a devDependency and nothing in CI publishes it. It stops being inert if that package is ever published from CI, or if any package uses `workspace:*` in `dependencies` or `peerDependencies`.
- **`@typeorm/legacy-naming-strategies` is on npm but no workflow publishes it.** 1.0.0 was pushed by hand.
- **Duplicated per-package config.** Each package still carries its own near-identical `tsconfig.json`, `.mocharc.json` and `.c8rc.json`. Hoisting those is the obvious next step.
- **`tests-legacy-naming-strategies` runs a full core build.** The `build` job produces `build/compiled` rather than `build/package`, so its artifact cannot be reused as-is without changing what the Linux and Windows jobs consume.
- **Trusted Publishing and `--provenance` are untested.** A local registry cannot exercise OIDC. The publish command is unchanged apart from its working directory, but a nightly `dev` run would confirm it before a stable release.
- **Sonar coverage mapping** now runs with `projectBaseDir: packages/typeorm`. The paths inside the uploaded lcov reports should line up, but only a real run on this PR will confirm it.
## CI results on this branch

Every database suite passes on Node 20 and 24 plus Windows: Oracle, SAP HANA, Spanner, MSSQL, MongoDB, CockroachDB, Postgres 14 and 17, MySQL/MariaDB, sqljs and better-sqlite3. `build`, `format`, `preview`, `tests-codemod` and `tests-legacy-naming-strategies` all pass. Sonar reports `new_coverage` at 74.5% against a 70% threshold, which confirms coverage reporting still maps correctly under `projectBaseDir: packages/typeorm`.

Three checks failed. One was a real bug in this branch, now fixed; two are artifacts of the move that need a maintainer decision.

**`lint` (fixed).** `@typeorm/legacy-naming-strategies` is type-checked against `typeorm`, which resolves to `build/package`, so eslint reported six `no-unsafe-*` errors when the core package had not been built. Reproduced locally (exit 1 without the build, exit 0 with it) and fixed by building before linting, the same way `tests-legacy-naming-strategies` already does.

**Sonar quality gate (needs a decision).** The gate fails on one condition only: `new_duplicated_lines_density` at 8.9% against a 3% threshold. Reliability, security, maintainability, coverage and security hotspots all pass. The cause is that Sonar sees `new_lines` = **290101**, because every file has a new path, so the whole codebase counts as new code and its pre-existing duplication (25931 lines, mostly the driver `QueryRunner` classes) is measured against the new-code gate. The same duplication exists on `master` today. It will re-baseline once this lands. I have not touched `sonar.cpd.exclusions` or the gate, since loosening either to get one PR through seems like the wrong trade.

**CodeQL (needs a decision).** One new high-severity alert, `js/weak-cryptographic-algorithm` at `packages/typeorm/src/platform/PlatformTools.ts:102`. It is `crypto.createHash("sha1")` used to build internal index and alias names, documented in place as not for cryptographic security, and introduced in #12648. Alert **#95** is the identical finding at the old path `src/platform/PlatformTools.ts` and is already dismissed as "won't fix"; the dismissal does not follow a rename, so CodeQL raised it again as #97. It needs the same dismissal.

This is deliberately scoped to layout and plumbing. It does not adopt Nx or Turborepo, and does not add a release tool. Splitting drivers into separate packages (#2479) and replacing gulp (#11691) remain separate conversations.


.vscode/settings.json
```diff
@@ -7,5 +7,5 @@
   },
   "editor.formatOnSave": true,
   "git.branchProtection": ["master"],
-  "js/ts.tsdk.path": "node_modules/typescript/lib"
+  "js/ts.tsdk.path": "packages/typeorm/node_modules/typescript/lib"
```
alumni, line 10: shouldn't we have `node_modules` in the workspace root?

alumni (approved):

gioboa (approved):

.vscode/settings.json
michaelbromley: The workspace root does have a `node_modules`, but it only holds what the root manifest itself declares: prettier and its plugins, husky, lint-staged, eslint and standard-changelog. TypeScript is a devDependency of `packages/typeorm`, so pnpm installs it at `packages/typeorm/node_modules/typescript` and there is no copy at the root for the old path to resolve against.

I pointed the setting at the package so the TypeScript version stays declared in one place. The alternative is adding `typescript` to the root devDependencies, which would let the setting go back to `node_modules/typescript/lib` at the cost of a second version spec to keep in sync with the package. Happy to switch if you prefer that.

Related, in case it looks odd: eslint is still a root devDependency because lint-staged runs `eslint --fix` from the repository root on pre-commit. eslint 10 resolves the config from the file being linted, so it picks up `packages/typeorm/eslint.config.mjs`. I checked that path works.

michaelbromley:
Note: I did not address the findings of Sonar or Coverage because they were pre-existing and only got picked up because this PR touched all the files due to the rename

PR #12860, opened 2026-09-10
fix(query-builder): throw on primitive relation where values

### LINKED ISSUE #12712: findBy with a primitive under a relation key silently returns all rows (no WHERE clause)

arcadiastudiobr:
## Problem

indBy(Post, { author: 1 }) / indOneBy({ relation: primitive }) silently dropped the criterion: uildWhere pushed a join and recursed on a primitive, which iterates no keys and produced an empty condition. The query returned **every row**. Reported in #12712.

This is a silent-wrong-data bug for multi-tenant style queries (indOneBy({ user: userId })).

## Solution (Option A — reject invalid input)

When a relation property receives a primitive other than boolean 	rue, throw a TypeORMError pointing at the documented nested form { relation: { id: value } }.

Preserved behavior:
- 
elation: true — documented join-relation form
- 
elation: { ... } — nested criteria
- 
elation: FindOperator — operators on owning sides
- 
elation: null — handled by invalidWhereValuesBehavior

Aligned with #12730 / #12771 (throw), and addresses Qodo feedback from the earlier shorthand attempt (boolean 	rue regression, composite keys, transformers, type casts).

## Testing

ind-options-where.test.ts:
- indBy(Post, { author: 1 }) rejects with TypeORMError
- nested { author: { id: 1 } } still filters correctly
- indBy(Author, { photos: 1 }) rejects with the same clear error

Closes #12712


arcadiastudiobr:
Pushed a fix for the new Qodo findings: error message no longer claims boolean/null are accepted, and no longer hardcodes id — it lists the related entity's real primary column names from metadata.

arcadiastudiobr:
@naorpeled @alumni — this implements **Option A** from #12712 (throw on primitive relation criteria), aligned with #12730 / #12771.

Differences vs the earlier shorthand attempt:
- Preserves `relation: true` (join-only)
- Error message uses the target entity's actual primary column names
- Qodo: 0 bugs after the last push
- 2 functional tests

Happy to close in favor of #12730/#12771 if you prefer those — just say which direction you want merged.

michaelbromley:

```
git fetch origin
```

PR #12863, opened 2026-09-12
Orm utils.normalize where criteria()

gloskull:
Motivation
Ensure OrmUtils.normalizeWhereCriteria() implements the documented default of throwing on null/undefined even when only one of the invalidWhereValuesBehavior options is provided, preventing silent/unsafe write predicates.

Description
Resolve per-type defaults by computing an invalidWhereValuesBehavior object where null and undefined independently default to "throw" when options is partially configured in OrmUtils.normalizeWhereCriteria().
Use the resolved invalidWhereValuesBehavior when checking value === null or value === undefined instead of consulting options directly.
Add a unit test (normalizeWhereCriteria) that asserts an unspecified sub-option still defaults to throw and extend existing unit coverage for the behavior.

Testing
Ran pnpm run compile and the build completed successfully.
Ran the unit tests for the compiled output with pnpm exec mocha --no-config --check-leaks --color --exit --file ./build/compiled/test/utils/test-setup.js build/compiled/test/unit/util/orm-utils.test.js and the tests passed (30 passing).


jdjioe5-cpu:
## Claiming via Hermes cloud agent (jdjioe5-cpu)

Walking through the issue:

The current `OrmUtils.normalizeWhereCriteria` has:

```ts
if (value === undefined) {
    const behavior = options?.undefined ?? "throw"
    // ...
} else if (value === null) {
    const behavior = options?.null ?? "throw"
    // ...
}
```

The default `"throw"` only kicks in when `options?.undefined` is itself undefined. If a caller passes `{ null: "throw" }` (forgetting `undefined`), the undefined branch silently defaults to `"throw"` via `??` — but only because `options.undefined` is `undefined`. If a caller passes `{ undefined: "ignore", null: "throw" }`, that's correct. But if a caller passes `{ undefined: "ignore" }` and then a `null` value appears, `options?.null ?? "throw"` will throw correctly.

So the question is: what's the actual bug? Re-reading the issue title — "Orm utils.normalize where criteria()" — and body — "implements the documented default of throwing on null/undefined even when only one of the invalidWhereValuesBehavior options is provided" — the bug is the asymmetry between the **documentation** (both default to throw independently) and the **code** (the recursive call shares the `options` object, so a `null` typo in `options` propagates from the `undefined` branch's resolution path or vice versa).

I think the fix is small: explicitly resolve both `null` and `undefined` behaviors up front, so the recursive call always carries a fully-populated options object:

```ts
const resolvedOptions: InvalidFindOptionsWhereBehavior = {
    null: options?.null ?? "throw",
    undefined: options?.undefined ?? "throw",
};
```

Then pass `resolvedOptions` to the recursive call instead of the raw `options`. That way the documented per-type defaults are honored independently on every recursion level.

I'll open a PR with:
1. The resolution up-front.
2. A regression test that calls `normalizeWhereCriteria({ name: null }, { undefined: "ignore" })` and asserts it throws (today it does, but only by accident — the explicit resolution makes the intent visible).
3. A regression test that calls `normalizeWhereCriteria({ name: undefined }, { null: "ignore" })` and asserts it throws.

Estimated turnaround: 24h. Let me know if maintainers want it scoped differently.

michaelbromley:

```
git fetch origin
```

PR #12864, opened 2026-09-12
fix(postgres): use ALTER COLUMN TYPE for length changes instead of DROP/ADD

Arkhos-Overlord:
## Summary

When a column's length changes within the same base type (e.g. `varchar(50) → varchar(51)`), use `ALTER COLUMN ... TYPE` instead of dropping and recreating the column, which causes data loss.

Fixes #3357

## What the fix does

- Detects a **length-only** change (same base type, no array/generatedType/asExpression changes)
- Emits `ALTER TABLE ... ALTER COLUMN ... TYPE ...` instead of `DROP COLUMN` + `ADD COLUMN`
- The down migration reverts to the old type specification
- All other change kinds (type, array, STORED generation, renames, default/nullability/comment/uniqueness/precision/scale) stay on their pre-existing code paths

## Example

A migration changing varchar(50) → varchar(51) now generates:

Instead of:

## Review fixes

Follow-up commit addressing external review feedback:

- Fixed build error: `buildColumnType` → the existing `createFullType`
- Rescoped the new branch to **length-only** changes; property-only changes (default, nullability, comment, uniqueness, precision, scale) no longer pass through it
- Renames (including rename + resize) keep their existing handling; the ALTER targets the column's pre-change name
- Cached column metadata updated after the ALTER
- Removed dead/redundant conditions from the new branch
- Collation changes are excluded from the new branch: their separate `TYPE ... COLLATE` statement uses the bare base type and would strip an applied length; combined changes keep the pre-existing path, with a regression test pinning this

## Verification

- `tsc --noEmit` clean; `eslint` clean; `prettier` clean on both changed files
- Regression tests added in `test/functional/database-schema/column-length/postgres/column-length-postgres.test.ts`:
  - length-only change emits `ALTER COLUMN ... TYPE` and no `DROP COLUMN`
  - default-only change does not drop/recreate the column
- Functional tests require a live Postgres — exercised by the CI test matrix.


Arkhos-Overlord:
/try #3357 (claiming bounty on opire.dev and algora.io via PR #12864)

Arkhos-Overlord (commented):
/try #3357 - claiming bounty via PR #12864 on opire.dev and algora.io

Arkhos-Overlord:
Re item 1 ("Generated rollbacks fail after renames"): generated migrations execute down queries in reverse order as well — `MigrationGenerateCommand` builds the file with `downSqls.reverse()` (src/commands/MigrationGenerateCommand.ts:169 and :176), matching `executeMemoryDownSql`'s reversal. For a combined resize + rename, rollback therefore runs `RENAME COLUMN new → old` first, then the `ALTER COLUMN ... TYPE` against the old name — consistent with the existing convention where down queries that execute before a rename reversal target `newColumn.name` (e.g. the default-handling block), while this branch executes after it and targets the old name.

Added a regression test covering both directions — "resize with rename applies the type change before the rename and reverts in reverse order" in `column-length-postgres.test.ts` asserts the up order (ALTER before RENAME) and the down execution order (rename back before the old-name ALTER). Item 8 (redundant test comments) is fixed as well.

michaelbromley:

```
git fetch origin
```

PR #12884, opened 2026-09-18
feat: add dialectTypes column option for per-database column types

### LINKED ISSUE #12874: Support per-dialect column types on @Column, e.g. dialectTypes: { postgres: 'jsonb' }
### Feature Description

Column types declared in `@Column({ type: ... })` are static — the same value is applied no matter which driver syncs or migrates the entity. But column types are not portable across databases:

- `jsonb` does not exist on mysql/mariadb/sqlite (#11930)
- `tinyint` does not exist on postgres (`ERROR: type "tinyint" does not exist` on the first sync)
- MariaDB 10.7 has a native `UUID` type with no entity-level way to pick it per connection (#8832)
- sqlserver wants `nvarchar(max)` where other drivers would use `text` (#11665)

Today the workarounds are either dropping to the lowest common denominator type, or hand-writing migrations with `if (queryRunner.connection.options.type === 'postgres')` branches — after which `synchronize` no longer reflects what the database actually looks like.

### The Solution

Extend `ColumnOptions` with a per-driver type override, keyed by the connection type:

```ts
@Column({ type: 'json', dialectTypes: { postgres: 'jsonb', mariadb: 'longtext' } })
payload: string;

@Column({ type: 'tinyint', dialectTypes: { postgres: 'smallint' } })
level: number;
```

Semantics are the same model other ORMs already use:

- when syncing/migrating with driver `X` and `dialectTypes[X]` is set, use that type verbatim
- otherwise fall back to `type`, exactly as today
- runtime behavior (hydration, transformers, indexes) is unchanged; only DDL type selection changes, and entities without the option behave 100% as before

### Considered Alternatives

- Abstract types like `simple-json`: portable but lossy — no access to native features such as jsonb indexing
- Hand-written migrations with provider checks: works, but duplicates the schema outside the entity and breaks `synchronize`
- `ValueTransformer`: transforms values, not the SQL column type

### Additional Context

The same feature is a built-in elsewhere, so the design is well proven: ent has `field.Float("amount").SchemaType(map[string]string{dialect.MySQL: "decimal(6,2)", dialect.Postgres: "numeric"})`, SQLAlchemy has `String(255).with_variant(VARCHAR(255, charset="utf8"), "mysql")`, and gorm recently merged support for dialect-qualified type tags (go-gorm/gorm#7867).

If maintainers are open to the direction, I'm happy to work on a PR. The natural implementation point is where each driver resolves the column type when building create/alter table SQL, falling back to existing behavior whenever the option is absent.

- [X] mysql
- [X] postgres
- [X] sqlite
- [X] sqlserver
- [X] cockroachdb
- [X] better-sqlite3

xue-ding-e:
Closes #12874 (the feature request; the issue number matched this branch by chance, the linked test lives under `test/github-issues/12874/`)

Column types are not portable across databases: `tinyint` exists on mysql and sqlite but not on postgres (sync fails with `DataTypeNotSupportedError` on first run), `jsonb` is postgres-only, and every driver has a few types like these. Today a single `type` forces either a portable lowest common denominator or hand-written migrations.

This adds a `dialectTypes` column option so one entity can describe the column type per driver while keeping a default for the rest:

```ts
@Column({ type: "json", dialectTypes: { postgres: "jsonb" } })
payload: string

@Column({ type: "tinyint", dialectTypes: { postgres: "smallint" } })
level: number
```

On postgres the columns are `jsonb` / `smallint`, on mysql and sqlite `json` / `tinyint`, on any other driver the generic `type` as usual.

### What it touches

- `ColumnOptions` / `ColumnMetadata`: the new option is stored alongside `type`.
- `TableUtils.createTableColumnOptions`: when building DDL, a matching override replaces `driver.normalizeType(columnMetadata)` for that driver — this is the single point where metadata columns become schema columns, so `synchronize`, query runners and schema builders all pick it up.
- `EntityMetadataValidator`: the override participates in the data type validation, so a logical type unsupported by the *current* driver no longer fails initialization when an override covers that driver — otherwise `tinyint` on postgres could never get as far as building DDL.
- Runtime behavior is unchanged: value resolution, transformers and query building keep working off the logical `type`, and columns without `dialectTypes` are processed identically to today.

### Testing

`test/github-issues/12874/` syncs the entity above on postgres, mysql/mariadb and better-sqlite3 and asserts the resulting column types per driver. Locally verified on better-sqlite3 + postgres (`1 passing`); mysql/mariadb will run in CI. `tsc --noEmit` and the `test/unit` suite also pass, plus a spot check that existing github-issue tests keep passing on the same local setup.

### Notes

- The same gap exists in other ORMs and has been solved similarly: ent has `SchemaType(map[string]string{...})`, SQLAlchemy has `TypeEngine.with_variant`, and gorm recently merged dialect-qualified type tags (go-gorm/gorm#7867). The semantics here follow the same pattern: dialect tag wins on its driver → generic `type` → inferred default.
- Driver keys are the `DatabaseType` strings (`"mysql"`, `"postgres"`, `"better-sqlite3"`, ...). The override value is used verbatim as the DDL type, so it can carry its own length/params (e.g. `"varchar(10)"`).


xue-ding-e:
Thanks for the review — pushed [4fd5664](https://github.com/typeorm/typeorm/pull/12884/commits/4fd5664) addressing the findings:

1. **Coverage location** — moved the suite to `test/functional/columns/dialect-types/` and renamed it away from the issue framing.
2. **Sync never settles** — agreed, this was the real one. The effective physical type is now resolved in a single place, `ColumnMetadata.resolveDriverType(driver)`, and that is what the postgres/mysql/sqlite `findChangedColumns` comparisons and the rename detection in `RdbmsSchemaBuilder` use — same source as table column creation. Added a regression test that runs `synchronize` a second time and asserts no tables changed (passing locally on postgres + better-sqlite3, where both entities carry an override that differs from the logical type).
3. **Parameterized overrides fail startup** — validation now checks only the base name (`"varchar"` in `"varchar(10)"`); the full value still goes to DDL verbatim.
4. **Overridden key types break relations** — `RelationJoinColumnBuilder`, both junction column constructions and the closure junction builder now copy `dialectTypes` from the referenced column.
5. **Typed column declarations do not compile** — the option moved to `ColumnCommonOptions`, so the type-first `@Column(...)` overloads accept it as well.
6. **Schema-based entities lose overrides** — added to `EntitySchemaColumnOptions` and forwarded in `EntitySchemaTransformer`.

Locally re-verified: `tsc --noEmit` clean, `test/unit` (139), `test/functional/columns` (33) and the metadata-builder/entity-metadata suites all pass, with the two new cases in the functional suite covering the override selection per driver and the sync convergence.

xue-ding-e:
Updated in 3c23f9e, with upstream master merged in 36e3c0a.

- Added the option to the entity guide and decorator reference.
- Normalized physical types and their parameters through the existing drivers. The regression cases include `int`, `varchar(010)`, `decimal(10,2)` and timestamp precision, while retaining unique constraints and the original logical metadata.
- Rejected blank overrides and invalid parameter combinations during initialization.
- Restricted TypeScript driver keys to `DatabaseType`, with a compile-time typo check.

The second-sync test now asserts that both generated query lists are empty, in addition to comparing the resulting schema. This catches repeated alterations that leave the same final table definition.

Type checking, compilation and lint passed. The full local suite on PostgreSQL 17.9 with PostGIS/pgvector and better-sqlite3 finished with 3091 passing, 176 pending and no failures. AI assistance was used for these changes and this follow-up.

xue-ding-e:
Addressed the inherited-modifier finding in 072ba6e. An unparameterized override now keeps only compatible modifiers: `varchar(40)` can map to PostgreSQL `text`, and `decimal(12,4)` can map to `double precision`. Compatible length/precision/scale settings are preserved. The regression cases also check that the second synchronization generates no DDL and the original logical metadata is unchanged.

For the SAP timestamp finding, the documented [CREATE TABLE grammar](https://help.sap.com/docs/hana-cloud-database/sap-hana-cloud-sap-hana-database-sql-reference-guide/create-table-statement-data-definition?locale=sk-SK) lists `TIMESTAMP` without parameters. The precision argument belongs to the [CURRENT_TIMESTAMP function](https://help.sap.com/docs/hana-cloud-database/sap-hana-cloud-sap-hana-database-sql-reference-guide/current-timestamp-function-datetime?locale=en-US&state=PRODUCTION&version=2025_4_QRC), not that column type. The formatter's ability to append a number does not establish supported HANA syntax, so the resolver still rejects `sap: "timestamp(3)"`. Three tests using the real SapDriver with a stub client cover plain timestamp DDL and rejection of parameterized variants. No live SAP server was used.

Compilation, lint, formatting and commit hooks passed. The full local PostgreSQL 17.9/PostGIS/pgvector + better-sqlite3 suite finished with 3094 passing, 176 pending and no failures. AI assistance was used for the implementation, verification and this response.

xue-ding-e:
Fixed the SQL Server finding in 812ebd8. Both `@OutputTable` and typed query parameters now use the resolved physical type, length, precision and scale. Explicit `MssqlParameter` values keep their existing priority; entity value conversion, hydration and the original column metadata are unchanged.

The regression tests exercise actual SQL Server INSERT/UPDATE OUTPUT with Unicode and decimal values, then ordinary parameterized writes and finds (including `In`). They also check nullable/plain columns, computed-column declarations, metadata after queries, and no schema changes on a second sync.

Compilation, lint, formatting and commit hooks passed. The full local suite with PostgreSQL 17.9/PostGIS/pgvector, SQL Server 2025 Express 17.0.5005.3 and better-sqlite3 finished with 3101 passing, 176 pending and no failures. AI assistance was used for implementation, verification and this response.

xue-ding-e:
Fixed PostgreSQL extension detection and removed the redundant negative checks in 2d99cb8.

Extension setup now checks the resolved physical column types for citext, hstore, cube, spatial types, ltree and vector/halfvec. UUID generation and exclusion-constraint detection retain their existing behavior. The regression test records the extension-creation requests; a real PostgreSQL test also created a citext column from a logical varchar, verified case-insensitive lookup, and produced an empty second-sync query plan. The test database did not have citext installed before this run.

The parser still rejects signed parameters and numeric overflow; only the unreachable checks after unsigned digit parsing were removed.

The same audit found an Oracle RETURNING gap: OUT binds now use the physical type, with unit coverage for text-to-clob and bytea-to-blob plus unchanged fallback cases. Oracle was not tested against a live server.

Final local suite with PostgreSQL, SQL Server and better-sqlite3 enabled: 3107 passing, 176 pending, no failures. Compilation, lint, formatting and commit hooks passed. AI assistance was used for implementation, verification and this response.

xue-ding-e:
Addressed the new findings in 4f1727b:

- Oracle RAW RETURNING now uses `DB_TYPE_RAW` and a buffer sized to the physical column, including values larger than the client's default 200 bytes.
- Explicit join columns inherit the referenced column's `dialectTypes`, consistent with the existing logical-type inheritance. PostgreSQL tests cover both missing and conflicting FK overrides, relation writes/reads, and an empty second-sync plan.
- SQL Server overrides reject invalid decimal precision/scale and missing or out-of-range vector dimensions. Valid `vector(3)`, inherited dimensions and `datetime2(0)` are covered.
- Removed the two redundant fixture comments.

The decimal check is driver-specific: [PostgreSQL allows scale greater than precision](https://www.postgresql.org/docs/current/datatype-numeric.html), so `decimal(2,5)` remains valid there. Its schema, value round-trip and second-sync test runs on PostgreSQL 15+; zero numeric precision is rejected without rejecting zero time precision.

Compilation, lint, formatting and commit hooks passed. The full local PostgreSQL/SQL Server/better-sqlite3 run had 3114 passing, 176 pending and no failures. After isolating the PostgreSQL 15+ case, the final focused run passed all 28 tests. Oracle RAW coverage uses driver/query-builder tests, not a live Oracle server. AI assistance was used for these changes, verification and this response.

xue-ding-e:
Fixed MAX-length overrides and added the linked issue comment in a6b87ed.

The parser accepts `max` for SQL Server `varchar`, `nvarchar` and `varbinary`, and Spanner `string` and `bytes`. Fixed-length types and precision/scale arguments still reject it. An inherited MAX length is retained only when the replacement type supports it. The docs now list these forms.

SQL Server integration tests create the MAX columns, bind and return 9000-character ASCII text, 5000-character Unicode text and 9000-byte binary data, then read the saved values back. A new Spanner emulator test creates STRING(MAX)/BYTES(MAX), saves and reads values beyond the logical length, and checks that a second sync generates no DDL. Logical metadata remains unchanged.

TypeScript compilation, lint, formatting and commit hooks passed. The full local PostgreSQL/SQL Server/better-sqlite3 run finished with 3138 passing, 176 pending and no failures. The final focused run with those databases plus the repository's Spanner emulator passed all 52 tests. AI assistance was used for implementation, verification and this response.

xue-ding-e:
Addressed the spatial write finding in 38df651. INSERT and UPDATE now choose the spatial constructor and PostgreSQL cast from the resolved physical column. SELECT uses the same physical type for spatial read expressions. Value preparation, hydration and the original logical column metadata remain unchanged.

The new unit cases cover 13 geometry/geography and spatial/non-spatial combinations across PostgreSQL, SQL Server, MySQL and Aurora MySQL, including legacy MySQL spatial functions. Real PostgreSQL, SQL Server and MySQL tests create the physical columns, insert, update and read compatible values, then check that the next schema sync emits no DDL. The affected dialect-types suite passed all 20 tests after its table comparison was scoped to the table it owns.

TypeScript compilation, lint, formatting and commit hooks passed. A broader four-driver run had 3154 passing and 176 pending, with one failure in the prior all-tables test comparison; that test was narrowed and then passed in the focused run. The unrelated #4753 test targets fixed localhost:3306 and was excluded because that port belongs to another local service. AI assistance was used for the changes, verification and this response.

